PR #771, opened 2016-07-26
Seo editor improvements

ThomasKeil:
Fixes and improvements:

-  `seopanelTreeRootAction()` now produces the same results for the root document as for the other documents.
- The SeoPanel now shows the reason for it's warnings as a tooltip.
- The edit panel now updates the amount of characters on typing.
- Editing the root document now updates it's data in the grid.


brusch:
Thanks!

PR #774, opened 2016-07-27
Bugfix: Inputfield for traits was missing

ThomasKeil:
The input field for traits in the new Extjs6 surface of class definitions was gone missing.
This fix adds it again.


ThomasKeil:
The second time I add this feature to pimcore ;-)

brusch:
Thanks & Sorry :-)

PR #775, opened 2016-07-29
Moved preAction event after actionData is set

cleggypdc:
Allows the actionData to be accessed in the preAction event.


fashxp:
thank you very much...

PR #776, opened 2016-07-29
Moved setElementState/Status

cleggypdc:
Since moving the state / status out of the fields within a classdefinition, State/Status should be saved after the object version was saved, allowing the workflowmanagement.action.before event and object saving to prevent the transition.


fashxp:
thank you very much.

PR #785, opened 2016-08-02
docs: Fix for example code in "Accessing Classification Store through code"

podarcis:
Minor documentation fix for Classification store example code.


mrudnicki:
thanks

PR #790, opened 2016-08-05
Update AbstractObject.php

mkoprek:
Model from cache are passing by reference it cause problems when you need to use same model twice.

Anyway each model should be separated from others.

### DIFF
--- a/pimcore/models/Object/AbstractObject.php
+++ b/pimcore/models/Object/AbstractObject.php
@@ -236,7 +236,8 @@ class AbstractObject extends Model\Element\AbstractElement
         $cacheKey = "object_" . $id;
 
         try {
-            $object = \Zend_Registry::get($cacheKey);
+            // we need to have new instance of model, in other case it will be a reference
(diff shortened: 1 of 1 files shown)


brusch:
Well, not really, ... that depends on the use case ...

mkoprek:
I'm using object.preUpdate event. I want to compare changes from request with my clean model values. Unfortunately all values from new instance of model are overwritten by pimcore/pimcore/modules/admin/controllers/ObjectController.php:1009, a cause of reference. So there is no any option to see original values, only pure SQL.

Models should represent really data structure in this case that is not possible.

mrudnicki:
@mkoprek if you wanted to see original data, a better solution would be to add some parameter like "originalData" (We will be very happy to see a pull request :) ) or just clone the object by yourself in the code. 
Every change in base classes could cause unpredictable issues, this is the reason why we have to be cautious in every pull request related to crucial elements of the Pimcore core.

PR #793, opened 2016-08-09
Update AbstractObject.php

mkoprek:
Better solution for problem with multiple instances of one model.


brusch:
Probably a different method for obtaining the object is a better approach then adding an argument, but we'll discuss this internally first. If we'd add this parameter it would also necessary do do the same consistently across all models (Asset, Document, ClassDefinition, Staticroute, ...).  
Will keep you updated!

PR #801, opened 2016-08-11
fix webservice import for object-bricks without corresponding class

dpfaffenbauer:
#### Fixes 
Webservice Import for Object-Bricks when corresponding class has not been created yet. This happens when a Class has a brickfield but no Objectbricks are configured for it.


brusch:
Thanks!

PR #804, opened 2016-08-12
thumbnail configuration: use di to get config class

zieglerh:
Use DI for the image and video thumbnail configurations, so that you can overwrite it.

## Additional info  
I included the full class name into di.php to prevent Pimcore\Model\Asset\* from interfering and Pimcore\Model\Asset\Image\Thumbnail\* would be more of a burden than of help.

Thanks for implementing DI 👍

### DIFF
--- a/pimcore/lib/Pimcore/Console/Command/ThumbnailsImageCommand.php
+++ b/pimcore/lib/Pimcore/Console/Command/ThumbnailsImageCommand.php
@@ -98,7 +98,8 @@ class ThumbnailsImageCommand extends AbstractCommand
                 }
 
                 if ($input->getOption("system")) {
-                    $thumbnail = Asset\Image\Thumbnail\Config::getPreviewConfig();
+                    $thumbnail = \Pimcore::getDiContainer()->get('Pimcore\Model\Asset\Image\Thumbnail\Config');
+                    $thumbnail = $thumbnail::getPreviewConfig();
                     if ($input->getOption("force")) {
                         $image->clearThumbnail($thumbnail->getName());
                     }
--- a/pimcore/config/di.php
+++ b/pimcore/config/di.php
@@ -8,4 +8,6 @@ return [
     'Pimcore\Model\Object\*' => DI\object('Pimcore\Model\Object\*'),
 
     \Pimcore\Image\Adapter::class => DI\factory([\Pimcore\Image::class, 'create']),
+	'Pimcore\Model\Asset\Image\Thumbnail\Config' => DI\object('Pimcore\Model\Asset\Image\Thumbnail\Config'),
+	'Pimcore\Model\Asset\Video\Thumbnail\Config' => DI\object('Pimcore\Model\Asset\Video\Thumbnail\Config'),
 ];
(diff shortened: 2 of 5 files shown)


pimcore/lib/Pimcore/Console/Command/ThumbnailsImageCommand.php
```diff
@@ -98,7 +98,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 }
 
                 if ($input->getOption("system")) {
-                    $thumbnail = Asset\Image\Thumbnail\Config::getPreviewConfig();
+                    $thumbnail = \Pimcore::getDiContainer()->get('Pimcore\Model\Asset\Image\Thumbnail\Config');
+                    $thumbnail = $thumbnail::getPreviewConfig();
```
mnapoli: The [`getPreviewConfig()` method is static](https://github.com/pimcore/pimcore/blob/5036920fca970ab124914d55aed9180cf50541dd/pimcore/models/Asset/Video/Thumbnail/Config.php#L101). Using the container means getting back an *instance* of the class: it's not useful in that case since static methods are not executed on instances.

zieglerh: Yes, you are right, but the DI Container will create it only once so it's a singleton.

So, I thought about a better solution and that would be a factory. Since ::getPreviewConfig() is already a configuration factory, I would suggest doing this:
```php
	'Pimcore\Model\Asset\Image\Thumbnail\Config::getPreviewConfig' => DI\factory('Pimcore\Model\Asset\Image\Thumbnail\Config::getPreviewConfig'),
	'Pimcore\Model\Asset\Video\Thumbnail\Config::getPreviewConfig' => DI\factory('Pimcore\Model\Asset\Video\Thumbnail\Config::getPreviewConfig'),
`
Now it would be only matter to define what to use as key. "Pimcore\Model\Asset\Video\Thumbnail\Config::getPreviewConfig" is rather long, but makes sense to use. What would you suggest?

Thanks

brusch:
In my opinion the DI Container should already return the config and not the method to call.

zieglerh:
Like in my comment above. 
Should I post a new pull request or are you going to change this in the near future?

brusch:
I'll implement it as soon as possible :)

pimcore/lib/Pimcore/Console/Command/ThumbnailsImageCommand.php
mnapoli: Since it's a static method call there is no instance to create, so it's not even a singleton.

However your suggestion sounds good (putting the whole config in the container), that would really be a singleton now so it makes sense.

> "Pimcore\Model\Asset\Video\Thumbnail\Config::getPreviewConfig" is rather long, but makes sense to use. What would you suggest?

I'm not working with Pimcore at all so I'm not aware of the naming convention. In my projects I would name it something like `pimcore.assets.video.thumbnail.preview-config` or something like that. I tend to use class names for objects (because there is a real class behind), but I use lower-case IDs for stuff like config, etc. Of course that's just my personal usage, not related to Pimcore.

PR #806, opened 2016-08-16
Removed duplicate code for determining inheritable parent object

podarcis:
This change removed duplicate code for determining whether an object has an inheritable parent object.

I noticed this duplication while trying to override `Object\Concrete::getNextParentForInheritance()` but having not the desired effect as `Object\Service::hasInheritableParentObject()` implemented it again.


brusch:
Thanks!

PR #808, opened 2016-08-17
legacy class name fixes

dpfaffenbauer:
Just some minor legacy classname fixes, but should also be done ;)


weisswurstkanone:
thx

PR #811, opened 2016-08-18
check instanceof type of quantity value

dpfaffenbauer:
Fixes an issue where the type of $this->getUnit() is a string, but its expecting to be Unit.


brusch:
Thanks!

PR #812, opened 2016-08-18
Workflowmanagement - Refresh of formData after the preAction event

cleggypdc:

This small adjustment should make it possible to change the formData entirely within the preAction event. I came across this requirement when migrating a client. We have scenarios where we need to override the state / status of an action based on certain element information.


fashxp:
thank you very much

PR #813, opened 2016-08-19
Sent email enhanced log fix

chix:
Hi,

this is a small fix for the dynamic parameters popup in sent e-mails log. Any asset/document/object parameter is now always marked as deleted. This PR should fix that and also ads a check for classmapping when determinig the pimcore element type.

Before PR:
before

After PR:
after

M.


brusch:
Thanks!

PR #820, opened 2016-08-25
Classification store: Allow setting value zero ("0")

podarcis:
Value "0" evaluated to false in condition if($value) and thus not allow setting/saving this ordinary value.

### DIFF
--- a/pimcore/models/Object/Classificationstore.php
+++ b/pimcore/models/Object/Classificationstore.php
@@ -152,7 +152,10 @@ class Classificationstore extends Model\AbstractModel
     {
         $language  = $this->getLanguage($language);
 
-        if ($value) {
+        // treat value "0" nonempty
+        $nonEmpty = (is_string($value) || is_numeric($value)) && strlen($value) > 0;
+
+        if ($nonEmpty || $value) {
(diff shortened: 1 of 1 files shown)


pimcore/models/Object/Classificationstore.php
```diff
@@ -152,7 +152,10 @@ public function setLocalizedKeyValue($groupId, $keyId, $value, $language = null)
     {
         $language  = $this->getLanguage($language);
 
-        if ($value) {
+        // treat value "0" nonempty
+        $nonEmpty = (is_string($value) || is_numeric($value)) && strlen($value) > 0;
+
+        if ($nonEmpty || $value) {
```
brusch: Shouldn't be `(strlen((string) $data) < 1)` enough here?

podarcis: Well, I haven't tested every possible type and whether those object's __toString() behave this way (returning always an empty string when empty/unset).

I tried to be consistent with the check in https://github.com/pimcore/pimcore/blob/4.3.0/pimcore/models/Object/ClassDefinition/Data.php#L190

Issue #786, opened 2016-08-02
Image Thumbnails: possibility to force resize for any resizing transformation


Yivan:
Actually using Pimcore native thumbnail processor frame and contain doesn't have parameter like "forceResize" (or "doNotScaleUp"), same but inverted logic), like cover.
See: https://github.com/pimcore/pimcore/blob/master/pimcore/models/Asset/Image/Thumbnail/Processor.php#L31

So if original image is smaller, final image is not of the exepected size. I think it is a wanted behaviour (so image is not pixelated), but in many case we need to enforce the size, to be sure the output thumbnail is exactly the size expected (so it won't break some templating for instance).
As improvement, like cover, frame and contain could have an option to force resize. this option could be set in the Pimcore GUI thumbnail processor, and in online PHP call of thumbnail method by parameters.

Best regards.

brusch:
What is the actual use-case for scaling up the image?

brusch:
See also: https://github.com/pimcore/pimcore/issues/783

brusch:
Would also affect: scaleByWidth/Height and any other resizing transformation.

Yivan:
Thanks for your answer.

scaleByWidth/Height already have their force parameters in the core code, just missing the possibility to configure it during inline thumbnail generation (https://github.com/pimcore/pimcore/issues/783).

The actual case where we needed it for instance is, whatever the image size uploaded in pimcore backend, the output image should be of the given width/height (so scaled up if to small). So the output of the thumbnail generator give us an image of **always the same width/height**. Actually this is not possible as the thumbnail generator do not scale up if to small (for non vector graphics).
We managed to get the same effect by extending image asset (see the zip file i just added here https://github.com/pimcore/pimcore/issues/783#issuecomment-240417283), so this improvment could be managed by custom code (but is need https://github.com/pimcore/pimcore/issues/783 to be integrate to core code because it relay on scaleByWidth/Height).

And for homogeneity, why it has been added an option to force scale up on "Contain" but not on Frame or Cover ? I think really such possibility to force scaleup would be a great addition for some user case. For people wanting consistent width/height in output thumbnail generated.

Yivan:
For reference in https://github.com/pimcore/pimcore/pull/824 I asked if i could submit a more global pull request for this feature ?

Purpose of this pull request will be:
- I will try to implement for cover and frame the "doNotScaleUp" parameter
[...]

Issue #783, opened 2016-08-02
Missing "forceResize" parameter for asset inline thumbnail


Yivan:
"forceResize" parameter is missing for asset inline thumbnail generation from PHP code.

Here:
https://github.com/pimcore/pimcore/blob/master/pimcore/models/Asset/Image/Thumbnail/Processor.php#L28 : 

And here:
https://github.com/pimcore/pimcore/blob/master/pimcore/models/Asset/Image/Thumbnail/Processor.php#L29

It should be:
```
        "scaleByWidth" => ["width", "forceResize"],
        "scaleByHeight" => ["height", "forceResize"],
```

Thanks.

brusch:
Makes currently absolutely no sense, since this parameter is currently only used internally and not used by the thumbnail configuration in the user interface. 

![image](https://cloud.githubusercontent.com/assets/142037/17519692/4af77b40-5e4d-11e6-8f62-a9887cdd2ef8.png)

Yivan:
Yes it is not in the GUI (so could be added maybe ?), but the purpose was to be able to use it if extending/DI Asset model.

Here is a working case, where we had to change the core file as asked in this issue so we can use this option when extending image asset class (purpose was to create custom thumbnail parameters to force scale up, to cover the asking improvment https://github.com/pimcore/pimcore/issues/786):

[Image.zip](https://github.com/pimcore/pimcore/files/422706/Image.zip)

For me it doesn't make non sense as it could be used for some case, and it is just a slight modification on core files. I hope it make sense finaly ; )
Thanks.

PR #824, opened 2016-08-29
added missing/correcting processor items parameters

Yivan:
Here you will find some small fixes to thumbnail processor (wrong name of parameters for roundCorner and missing parameters for scaleBy).

Merging this pull request will solve https://github.com/pimcore/pimcore/issues/783 (i have updated some days ago with a real use case with zip exemple, where this missing paramters is needed - extending/DI thumbnail image and using items pipeline).

By the way can i prepare a pull request for https://github.com/pimcore/pimcore/issues/786 ? (i ask here at it is on the same thema).

Thanks.

### DIFF
--- a/pimcore/models/Asset/Image/Thumbnail/Processor.php
+++ b/pimcore/models/Asset/Image/Thumbnail/Processor.php
@@ -26,8 +26,8 @@ class Processor
 {
     protected static $argumentMapping = [
         "resize" => ["width", "height"],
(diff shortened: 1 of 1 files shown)


brusch:
Thanks! 
Naming: fixed in aff4692d1f076d2d28f8ac4968fce4af8f83a514 ($width, $height is correct)
Force parameter: see #786 (should be done in one go and not bit by bit - to be consistent)

PR #826, opened 2016-08-29
Refactor the URL cleanup logic to avoid chained redirects

YKWeyer:
Chaining 301 redirects is bad for user experience (more delay), for SEO, and for performance. We should clean the URL once, instead of redirecting the user at every step of the process.
screen shot 2016-08-29 at 18 44 34

- Creating a `$finalUrl` variable in Frontend router used along the url cleaning process ;
- Remove 3 identical blocks of code and run this code only at the end of the cleaning process, only if the `$finalUrl` is different than the `$originalPath`.


brusch:
Thanks!

PR #831, opened 2016-08-30
Allow empty units in quantity value field

IgorBenko:
This allows for quantity values to be empty. This is important for inheritance - in case the user changes the value in an inherited object or variant, the user can't revert back to inherited state since he cannot set the unit to empty, which is required in order for inheritance to kick back in.

I tested it and didn't notice any unwanted behaviour caused by this change.


brusch:
Thanks!

PR #835, opened 2016-08-31
doNotScaleUp for GUI, inline thumbnail and homogeneous

Yivan:
Here is the pull request for https://github.com/pimcore/pimcore/issues/786

It includes:
- "forceResize" has been changed to "doNotScaleUp" for consistency accross the code. Backward compatibility is not broken as  forceResize was not publically available (doNotScaleUp was publicaly accesible so changing it will have break compatibility).
- Implemented for all resizing methods wich could need this: scaleByWidth, scaleByHeight, cover, contain and frame
- Could be used as inline thumbnail generation: $assetImage->getThumbnail(["frame" => true, "width" => 500, "height" => 300, "doNotScaleUp" => false]);
- GUI Thumbnail on Pimcore backend updated for extjs4 and extjs6 so it can be configured and used from herre
- Bug corrected for cover: in GUI by default "Positioning" is empty, so in the method "cover" of Adapter.php, "center" is set as default if empty (as it is done actually in getByLegacyConfig method of Config object)

Thanks!

### DIFF
--- a/pimcore/lib/Pimcore/Image/Adapter.php
+++ b/pimcore/lib/Pimcore/Image/Adapter.php
@@ -137,9 +137,9 @@ abstract class Adapter
      * @param  $width
      * @return self
      */
-    public function scaleByWidth($width, $forceResize = false)
+    public function scaleByWidth($width, $doNotScaleUp = true)
     {
-        if ($forceResize || $width <= $this->getWidth() || $this->isVectorGraphic()) {
+        if ((!$doNotScaleUp) || $width <= $this->getWidth() || $this->isVectorGraphic()) {
             $height = round(($width / $this->getWidth()) * $this->getHeight(), 0);
             $this->resize(max(1, $width), max(1, $height));
         }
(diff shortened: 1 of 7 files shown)


brusch:
Thanks, but there's one issue with scaleByWidth and scaleByHeight, it actually breaks backward compatibility because the "doNotScaleUp" option wasn't there before, so after after an update the image would be scaled but it wasn't before, so you'd have to tick "doNotScaleUp" on every scaleByWidth/Height transformation after an update. 

So I think it's better to go with the "forceResize" option, which would be backward compatible (same behavior).

Yivan:
Thanks for looking for this PR.

So i will make a new PR with the following parameters name to keep all backward compatibility:

**scaleByWidth/scaleByHeight:** forceResize (as it was before in the code, but used internally)
**cover/contain/frame:** doNotScaleUp (because for cover doNotScaleUp was used and was publicaly available => inline thumbnail and GUI, so renaming it forceResize would break backward compatibility)

Would it be ok like this?
Thanks.

Yivan:
Another solution: I think the only way to call them all forceResize or doNotScaleUp would be an update script running after update. So in this case we could keep my PR as actually ?

brusch:
In my opinion using `forceResize` consistently across all transformations makes the most sense, as it is more self explanatory. 
Any other thoughts? 
So we'd have to write a migration script for cover/contain/frame.

Yivan:
Ok, if forceResize is better for all, i think like you it is more simple to understand than negative logic ; )
I prepare a new PR with only forceResize for all, setted to false by default.
This way, I think we need only to migrate "cover" as only this one use doNotScaleUp for now in the current core version.

Issue #696, opened 2016-06-24
Improvement - Adding Dailymotion video type


Yivan:
Actually, video support (in document or object) is for asset/youtube/vimeo.
Could maybe dailymotion be added ? So all the big 3 videos providers can be managed.

brusch:
PRs are welcome :-)
Alternatively you could use the "embed" editable, which utilizes oembed and works also for dailymotion.

Yivan:
Thanks, ok i will try to make a PR for dailymotion as soon as i can.

PR #836, opened 2016-08-31
Add dailymotion as video source (like youtube or vimeo)

Yivan:
This is a pull request for https://github.com/pimcore/pimcore/issues/696
It adds Dailymotion for video element. Now the 3 main video providers can be choose (YouTube, Vimeo and Dailymotion).

_Side note:
When making this pull request i find some incoherence betweens extjs4 and extjs6 in the videos.js tags. For extjs6 it seems to miss the URL converter to ID but not sure... So i touched nothing : )
Actually if copying youtube/vimeo URL (not ID), it seems to not work (video doesn't preview in the backend). So **I relabeled "ID" instead "URL" or "URL / ID" the input field**, as directly pasting URL seems broken. Maybe it could be abandonned (code to be cleaned) or corrected in another issue ?_

### DIFF
--- a/pimcore/models/Document/Tag/Video.php
+++ b/pimcore/models/Document/Tag/Video.php
@@ -31,7 +31,7 @@ class Video extends Model\Document\Tag
     public $id;
 
     /**
-     * one of asset, youtube, vimeo
+     * one of asset, youtube, vimeo, dailymotion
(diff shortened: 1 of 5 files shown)


brusch:
Thanks!

Issue #734, opened 2016-07-07
Option for choosing maximum selected items for multiselect object data-type


Yivan:
Multiselect data-type for object could have an option in the model tab to limit the maximum of the multiselect field wich can be selected.
Like the field collection/objects relation, where we could set the maximum number of item (max. items).

By the way, it will offer the possibility to use it as a single select (limited to 1 choice), as multiselect ExtJS support well HTML display inside. So it could be fallback for those specific need (for reference see: https://github.com/pimcore/pimcore/issues/732).

PR #837, opened 2016-09-01
Add a maximum selected items for multiselect object field

Yivan:
This is a pull request for https://github.com/pimcore/pimcore/issues/734

This add max. items options for "multiselect" field data type object, like multihref for instance (same position and name in GUI). Like multihref, validation is only done at GUI level (not in checkValidity php method, so direct API call can bypass this).

As a side use case, now multiselect can be used as color selector (as it accept HTML value) for choosing a single color by limit it to 1 max. items (like you did on the e-commerce Pimcore demo but for multiple color choice).

_Side note:
Validating the max items limit was a bit tricky. First i try using combination of "maxSelections", "validateOnChange: true" and "validitychange (combo, isValid, eOpts)" event, but it is not possible to restore previous value or abort change by returning false... so i opted for the "change" event. I had to add a slight timeout so restoring value are visualy done (see code comment)._

_The "maxSelections" config too (line 127 of pimcore\static6\js\pimcore\object\tags\multiselect.js) provide a visual warning on mouseover if more than maximum items is selected (shoud not happen as change event limit it on GUI level, but could be an visuel indicator if API was used to create/update object with more than maxSelections number of items)._

### DIFF
--- a/pimcore/models/Object/ClassDefinition/Data/Multiselect.php
+++ b/pimcore/models/Object/ClassDefinition/Data/Multiselect.php
@@ -47,6 +47,11 @@ class Multiselect extends Model\Object\ClassDefinition\Data
     public $height;
 
     /**
+     * @var int
+     */
+    public $maxItems;
+
+    /**
      * Type for the column to query
      *
      * @var string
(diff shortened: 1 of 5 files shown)


brusch:
Thanks!
Note: added some small fixes in 9a3789d02065c7d4427234953238dec31b559472

Issue #733, opened 2016-07-07
Colorpicker data-type for objects


Yivan:
I know some plugin existed before (i did not find them on packagist or github... maybe i made bad my search work ; )).

But maybe it could be natively integrated to Pimcore, as it is a common field involved in many project (for edition customer want to select color for header, some background, product color, etc...).

It could be a new type of field for object, named "colorpicker". To keep things simple and easy, it could be same like "input" field but with a colorpicker on the right of the field on the object edition. The ExtJS6 colorpicker could be used. When selecting the color it fills the input field (with "#ff0000" for red for instance). We can keep the regex validation of input so administrator can restrict to some color.
For the grid list, no colorpicker (won't be easy to add it on the grid), but user can edit the color in text mode like input.

So it keep thing simple. Adding the alpha layer seems difficult i think, so it could be a later improvment (and in this case i think using input type field as base is not a good idea, but separation of channels on different fields would be better, and having some method like getRGB(), getCMYK,... to use the color in the front end in CSS/HTML tag/style).

brusch:
See: http://docs.sencha.com/extjs/6.0.0-classic/Ext.picker.Color.html 
Needs palette configuration, default color palette is unusable. 

Since this would be also easily possible with a plugin I'd like to open a voting whether this should be a core feature or not.

fashxp:
at first glance I'm more into plugin integration

dpfaffenbauer:
Just take the existing plugin and port it to extjs 6?

Yivan:
Hum yes the extjs colorpicker is only a panel with some color. It is great for some cases (product color, etc.), as it could be setted in gridlist (i think for ecommerce shop). See capture: http://tinyurl.com/jsh8ffd
On the other way something like that could be achieved with multiselect (if we could limit to one, see improvement #734).

Another case, and wich cannot be resolved with the extjs color picker is let the user choosing the color he want (from all existing one). For this we can use a plugin (like http://www.eyecon.ro/colorpicker/ for instance). The simplest would be adding an option (a checkbox) on the model on the input field called "show colorpicker". And it will just make appear the colorpicker when the user edit the input field. So there is no new data-type to do, only GUI call.

This way we could have both classic use case of colorpickers. 
If we don't want introduce new data-type just making multiselect able to limit number of choice and adding "show colorpicker" for input field coudl do the trick.
[...]

PR #839, opened 2016-09-01
Color picker

Yivan:
This is a pull request for https://github.com/pimcore/pimcore/issues/733.

It is very light implementation, based on html5 color input type. No external dependency, just a checkbox for input field "Is color" has been added. If checked, it change the type of input to "color" instead of "text" in the edit layout only.
So it doesn't introduce new datetype, just some small code changes, and provide just a great helper tool for color selection (for HTML5 compatible browser: http://caniuse.com/#feat=input-color , cover nearly 70%, and last chrome/firefox version, the more used by Pimcore administrator).

Sample:
colorpicker

### DIFF
--- a/pimcore/models/Object/ClassDefinition/Data/Input.php
+++ b/pimcore/models/Object/ClassDefinition/Data/Input.php
@@ -1,4 +1,4 @@
-<?php 
+<?php
 /**
  * Pimcore
  *
@@ -47,13 +47,18 @@ class Input extends Model\Object\ClassDefinition\Data
      * @var string
      */
     public $columnType = "varchar";
-    
+
     /**
      * Column length
      *
      * @var integer
      */
     public $columnLength = 255;
+
+    /**
+     * @var bool
+     */
+    public $isColor = false;
 
     /**
      * Type for the generated phpdoc
@@ -145,7 +150,7 @@ class Input extends Model\Object\ClassDefinition\Data
     {
         return $this->getDataFromResource($data, $object, $params);
     }
-    
+
     /**
      * @return integer
      */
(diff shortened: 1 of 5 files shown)


Yivan:
Translation for "is_color" need to be added, i don't know the procedure. Is it on you side ?

Yivan:
By the way, i fixed the maxLength problem: it was only cutted in the save (server side), so enduser was not aware that some part of the text he fill could be lost (until he reopened the object).
Now it prevent typing more caracters than allowed.

brusch:
Thanks for the max-length fix -> merged.

brusch:
*Color* should be a dedicated data-type which can extend the input data-type. Adding an `isColor` flag to the input is a very bad idea and a pretty hacky solution. 

Please have a look at `email`, `time` or `firstname`, they all extend input but represent a dedicated data-type. The *color* data-type should have a proper validation (hex, rgb) and database column (6 digits).

Yivan:
Thanks brush to had a look on my PR.
The purpose of this PR was to not introduce new datatype and be a very lightweight solution, wich is it is true "hacky" ; ) A new data type will be better for data integrity sure.

Is such a new datatype planned ? I think it is better that the core team develop such feature, but if you want i can try a PR for this, you just need to give me expected specifications so i go in the good direction (native HTML5 or JS plugin, database type: 6 digits, no "#" - would have to be text field in this case, ...) ?

brusch:
Currently we don't have this feature on our roadmap. 
The native HTML5 color picker widget sounds like the easiest option, but I'm missing some features like direct hex input, predefined colors, ... so I don't think we should use it at the moment. 
ExtJS offers also a color picker, but again, very limited :( 
So I'm not sure what the best option would be. But you can develop the data-type as a plugin and share it on packagist so you're completely independent and you decide how it should work :)

brusch:
Here is an example plugin that provides a data-type: 
https://github.com/weblizards-gmbh/dynamicdropdown-plugin

Yivan:
Ok thanks brush! Yes native HTML5 depend of browser/OS. On Windows 7/Firefox as you can see there is predefined color, but not hex input. On JS color picker will be more generic/same on different plateform. I will think about a plugin.

Issue #180, opened 2015-07-16
Add contents of text assets to search index


brusch:
Hello,

Uploading asset of type .xls, .pdf, .doc, etc. add the text contained in those asset in the database for search by keyword in Pimcore admin.

CSV asset are not parsed for text, so they can't be searched by their content.
Would be nice like the other one to parse it (i think the process behind is convert it to PDF and after parse it...).

Best regards.

PR #843, opened 2016-09-01
add text asset to search data

Yivan:
This pull request is for https://github.com/pimcore/pimcore/issues/180

It add text content in search data for asset of type "text" having extension txt, csv or xml.

### DIFF
--- a/pimcore/models/Asset/Text.php
+++ b/pimcore/models/Asset/Text.php
@@ -16,7 +16,9 @@
 
 namespace Pimcore\Model\Asset;
 
+use Pimcore\Cache;
 use Pimcore\Model;
+use Pimcore\Logger;
 
 class Text extends Model\Asset
 {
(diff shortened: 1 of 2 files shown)


pimcore/models/Asset/Text.php
```diff
@@ -25,4 +27,22 @@ class Text extends Model\Asset
      * @var string
      */
     public $type = "text";
+
+    public function getText($page = null)
+    {
+        if (preg_match("/\.?(txt|csv|xml)$/", $this->getFilename())) {
+            $cacheKey = "asset_text_text_" . $this->getId() . "_" . ($page ? $page : "all");
+
+            if (!$text = Cache::load($cacheKey)) {
```
brusch: I don't think we need a cache here, reading from the file system should be sufficient. 
I'm also not sure if we need a check for the file extension, if the asset is classified as a text it shouldn't matter which file extension it has, or what is the idea behind that?

brusch:
There's no need for a `getText()` method, the data is actually text, so we can just use the normal `getData()` and make it less complicated. 
We also have to fix encoding issues, if the text tile contains eg. windows-1052 encoded text it would break the search index.

See e3e0b64402a15e16c792e86e16e89264d0e0b498 which should do the job properly.

Yivan:
Thanks brush for your refactoring!

Yes i used getText to make it like for documents (so it is same method name for document and text type). I tried to be the nearer possible of what was coded for document (even if it was not strictly necessary yes).

File extension filtering was just an added check, if by error some asset was of type text but with the wrong extension (ex: with API, by error asset of type text is created for an image, so "my-image.png" is of text type... it will be injected in the search database which is not good). But we can remove it yes if you think such error is no prone to arrive.

Issue #522, opened 2016-03-25
Decimal precision for Quantity Value data type


mmcbride1968:
## Feature description

In the data type, Number, there's the ability to define the Decimal precision of the field. The same feature needs to be added to Quantity Value data type as well.

Yivan:
Hello mmcbride1968, i just made a pull request for your issue: https://github.com/pimcore/pimcore/pull/845
Regards.

BlackbitNeueMedien:
As some of our clients have requests this exact feature, it would be nice to see Yivan's changes merged into Pimcore soon.

fashxp:
already implemented

PR #845, opened 2016-09-01
Options precision and max-min for quantity value

Yivan:
This pull request is for https://github.com/pimcore/pimcore/issues/522

Using quantity value could be improved by having the possibility to definit limits (max/min) and the type (integer, decimal). This way it is like numeric and can be greatly personnalized to different use case.

One of the case is to use it for multi-currency (where currency are units of quantity value), and where the value need to be accurate and so use decimal in database.
Other use case is forcing integer so user can only fill integer (plain quantity), for instance for stock left depending of different parameters (wich will be the 'units' in this case).

Code of this PR is greatly copied from numeric field. I made some extending testing and seems to work great!


brusch:
Unfortunately your PR is now conflicting :( 
Since we'll push the version 5 branch very soon, I'd like to ask you to recreate your PR based on the new branch, once it is available. 

Thank you very much!

Issue #640, opened 2016-06-02
Default module when creating empty document for multisite


Yivan:
When creating an empty document, module is the default module (website for instance).
When the document is created in a sub-site, this is always the default module.
I think it will be better if it is set to the module of the subsite (to be more technical the module associated to the rootsite document).

In multisite environnement, it is not possible to disable creating empty document, so editors can add empty document, by it will by default point to the module website... making it hard to make a default controller/action common for all the module in the website module (if each module have different template, predispatch hook, etc...).

To be more clear, for instance, creating an empty document could bring a page with just a big wysiwyg with the module layout/template. Actually creating an empty document will be "website" module (and default controller/action), but very often each module has his own template/layout (**as we speak of multi-sites**). So it will be better if by default it create a document with {module of the rootSite document}/default/default as module/controller/action.

brusch:
Currently not on our roadmap.

PR #846, opened 2016-09-01
default module for empty page in subsite

Yivan:
This is a pull request for https://github.com/pimcore/pimcore/issues/640

In multisite, when creating an empty page under a subsite, the module of the rootDocument will be used instead of website module. So empty page can be managed by module for multisite. Often each subsite is asociated to on module having his own logic/template. So creating empty page will no more break this logic.

### DIFF
--- a/pimcore/modules/admin/controllers/DocumentController.php
+++ b/pimcore/modules/admin/controllers/DocumentController.php
@@ -161,6 +161,16 @@ class Admin_DocumentController extends \Pimcore\Controller\Action\Admin\Element
                 } elseif ($this->getParam("type") == "page" || $this->getParam("type") == "snippet" || $this->getParam("type") == "email") {
                     $createValues["controller"] = Config::getSystemConfig()->documents->default_controller;
                     $createValues["action"] = Config::getSystemConfig()->documents->default_action;
+                    // we look for sites, so same default module is used than the rootDocument
+                    $sitesList = new Site\Listing();
+                    $sitesObjects = $sitesList->load();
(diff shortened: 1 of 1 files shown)


brusch:
Thanks for you PR, but unfortunately this isn't very generic and also not an expected behavior. Why should the module of the root document of a site automatically be the default for the entire site? Well, it would solve the issue in your case but we have to care about many other use-cases so we have to think more generic.

Yivan:
Yes this change can break compatibility in already multisite deployed. Sure it is a special case, when we use the combinaison domainname/module, but i was thinking it was more common than use same module for multi-domain/sub-site.

As totaly alternative solution:
- there could be an option in the system settings to desactivate "empty page" so GUI will only propose predefined template for enduser (The "Empty page" choice when creating a document won't be showed) ? If you are aggre **i can try to make a PR for this** (by default it will be activated so it don't break backward compatibility).

Here is a screenshot, showing where empty page won't appear if "Desactivate Empty page" is checked in system settings:
no_empty

Yivan:
Hum just thinking... for consistency it should be done for snippet, email, newsletter and pdf2print too ?

Use case of desactivate those "Empty..." is that often wehn we develop a projet we create some established template (page type), but often we are asking us: "And what we do of the Empty case ?" and the answer is often: "We send back a blank page saying: "Not editable"" or we but quickly `<div><?=$this->wysiwyg("text")?></div>` for the view as generic fallback.
I think it is more clean to desactivate the Empty option than provide some patch/fallback if it is not used for some project, as actually it can't be desactivated.

Issue #844, opened 2016-09-01
Delete dialog: show key of element


Yivan:
Actually deleting a document/asset/object bring only a generic popin: "Do you really want to delete this element". It is prone to human error.

As improvement, like deleting a file on windows desktop for instance, it could at least say the element name in the message so mistake can be prevented. It could be: "Do you really want to delete this element: cum-sociis-natoque-penatibus-et-magnis". (so we can keep actual translations)

Here is a screenshot of actual message:
![message](https://cloud.githubusercontent.com/assets/9025839/18178474/bedd5f46-707e-11e6-8a10-5153b22c0b8b.png)

brusch:
See: #847 - needs discussion

PR #847, opened 2016-09-01
Delete dialog: show key of element

Yivan:
This pull request is for https://github.com/pimcore/pimcore/issues/844

It add the key of the element (document/asset/object) in the popin when clic on delete in the main element tree. So enduser can be more confident when he delete element without asking himself if he has done the right click on the right element ; )

Here is an example of the result:
delete-popin


brusch:
We should think about the text in the box (doesn't look so nice) and if this should be solved client-side. 

Pending for the moment ... (we'll discuss this internally first).

Issue #618, opened 2016-05-26
Custom reports: placeholders


Yivan:
I saw the new perspective added in Pimcore 4, so now we can configure wich user can see wich custom reports (using perspective/portlets). This way it is possible to allow custom reports only for specific user it seems.

If we want to go further and allow custom reports data based on some "contextual" parameters (role of the user connected for instance), so user can have reports on "his" data.

A placeholder in the SQL query could open those new possibility.
For instance, we could write for SQL condition something like:
`object_1.name = :role(name)`
So only object linked to the connected user are in the custom reports.

This way custom reports are more "custom" and SQL request can be based on user connected.
If too dificult to make it on the SQL request, maybe as postfilter (grid configuration) is another option.

Yivan:
I got an idea wich seems better than placeholder and more easy to implement. As we have connector "Analytics", "SQL", we could choose "Script" and we could call a PHP script, or class method, that the developper can entirely made. The developper will have to care that this script return the datas in a good format (i think an array of datas, with first line beiing the column label).
This way, he can develop any scripts based on context of logged user and more! In his script he could make more complex SQL request too in PHP using Pimcore dbAdapter and mix it with pure PHP rules/conditions... open many possibilities : )

Yivan:
@weisswurstkanone : i think this issue can be close as now we can make custom report adapter.

PR #849, opened 2016-09-02
custom adapter for custom reports

Yivan:
This is a pull request for https://github.com/pimcore/pimcore/issues/618

Actually custom report can only be SQL request (context independant, so we cannot fetch object of connected user on Pimcore for instance) or Analytics. And extending custom reports seems not possible at the moment.
Purpose is like new reports for newsletter to be able to create custom class adapter for custom reports.
It can be used for parsing CSV, fetch external data on webservices, request database depending connected user on Pimcore, ...
**With this PR administrator can use the powerfull custom reports feature of Pimcore like a generic tool to draw datas coming from various sources.** I think it is a great addition.

Side note:
one translation need to be added: see t("custom_report_adapter_custom"). Some other text are hardcoded for this PR (to be more comprehensive), but should be dynamic.

You can test it with this custom adapter: 
[MyReportAdapter.zip](https://github.com/pimcore/pimcore/files/452361/MyReportAdapter.zip)

Screenshots:

Selector:
custom_new

Configured with results:
custom_with_result

Comprehensive errors for users:
custom_error1

custom_error2

custom_error3

### DIFF
--- a/pimcore/models/Tool/CustomReport/Config.php
+++ b/pimcore/models/Tool/CustomReport/Config.php
@@ -150,7 +150,23 @@ class Config extends Model\AbstractModel
     public static function getAdapter($configuration, $fullConfig = null)
     {
         $type = $configuration->type ? ucfirst($configuration->type) : 'Sql';
-        $adapter = "\\Pimcore\\Model\\Tool\\CustomReport\\Adapter\\{$type}";
+        if ($type === 'Custom') {
+            if (@\class_exists($configuration->className, true)) { // can need autoloader, can send warning and break json response so we use @
+                $adapter = $configuration->className;
+                $adpaterInstance = new $adapter($configuration, $fullConfig);
+                // we make some check for security reason (because className can be choosen at the GUI level)
+                if (get_parent_class($adpaterInstance) !== 'Pimcore\\Model\\Tool\\CustomReport\\Adapter\\AbstractAdapter') {
+                    throw new \Exception("Class" . $configuration->className ." doesn't extend AbstractAdapter");
+                }
+                if ((!method_exists($adpaterInstance, 'getData')) || (!method_exists($adpaterInstance, 'getColumns')) || (!method_exists($adpaterInstance, 'getAvailableOptions'))) {
+                    throw new \Exception("Class" . $configuration->className ." must implement methods getData, getColumns and getAvailableOptions");
+                }
+            } else {
+                throw new \Exception("Class" . $configuration->className ." doesn't exist");
+            }
+        } else {
+            $adapter = "\\Pimcore\\Model\\Tool\\CustomReport\\Adapter\\{$type}";
+        }
 
         return new $adapter($configuration, $fullConfig);
     }
--- a/pimcore/modules/admin/views/scripts/index/index6.php
+++ b/pimcore/modules/admin/views/scripts/index/index6.php
@@ -501,6 +501,7 @@ $scripts = array(
     "pimcore/report/custom/report.js",
     "pimcore/report/custom/definitions/sql.js",
     "pimcore/report/custom/definitions/analytics.js",
+    "pimcore/report/custom/definitions/custom.js",
 
     "pimcore/settings/tagmanagement/panel.js",
     "pimcore/settings/tagmanagement/item.js",
--- a/pimcore/static6/js/pimcore/report/custom/definitions/custom.js
+++ b/pimcore/static6/js/pimcore/report/custom/definitions/custom.js
@@ -1,0 +1,88 @@
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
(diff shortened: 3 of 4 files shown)


pimcore/models/Tool/CustomReport/Config.php
```diff
@@ -150,7 +150,23 @@ public static function getReportsList()
     public static function getAdapter($configuration, $fullConfig = null)
     {
         $type = $configuration->type ? ucfirst($configuration->type) : 'Sql';
-        $adapter = "\\Pimcore\\Model\\Tool\\CustomReport\\Adapter\\{$type}";
+        if ($type === 'Custom') {
+            if (@\class_exists($configuration->className, true)) { // can need autoloader, can send warning and break json response so we use @
+                $adapter = $configuration->className;
```
mrudnicki: typo, should be adapter

pimcore/models/Tool/CustomReport/Config.php
```diff
@@ -150,7 +150,23 @@ public static function getReportsList()
...
+                $adpaterInstance = new $adapter($configuration, $fullConfig);
+                // we make some check for security reason (because className can be choosen at the GUI level)
+                if (get_parent_class($adpaterInstance) !== 'Pimcore\\Model\\Tool\\CustomReport\\Adapter\\AbstractAdapter') {
```
mrudnicki: why not instanceof? You shouldn't check the parent class

pimcore/models/Tool/CustomReport/Config.php
```diff
@@ -150,7 +150,23 @@ public static function getReportsList()
...
+                    throw new \Exception("Class" . $configuration->className ." doesn't extend AbstractAdapter");
+                }
+                if ((!method_exists($adpaterInstance, 'getData')) || (!method_exists($adpaterInstance, 'getColumns')) || (!method_exists($adpaterInstance, 'getAvailableOptions'))) {
```
mrudnicki: \Pimcore\Model\Tool\CustomReport\Adapter\AbstractAdapter::getData is an abstract method if the instance of the class extends that Adapter, in my opinion we shouldn't check if method exists again.

pimcore/models/Tool/CustomReport/Config.php
Yivan: Thanks mrudnicki. You are right we can use instanceof. It allows more felxibility as AbstractAdapter could be any parent of the class. Purpose in my code was to force it to be the direct parent, just to enforce security as giving the user to inject directly the class name to be call from the GUI.
If you think this is not a security problem so instanceof is better, and you can change the code to it yes.

pimcore/models/Tool/CustomReport/Config.php
Yivan: As my previous line comment, it was for security purpose too, but my code is not good, as it will alwyas send true (if class extend AbstractAdapter). In ideal, the custom reports should maybe "implement" an interface than extend the abstract class, but it will need more change in the core code.
So you are right you can remove this check too.

Or if you prefer to keep it, it need to check it exist for the class itself (not including the parent), using reflection could maybe do the trick:
$reflectionMethod = new \ReflectionMethod($adapterInstance, 'getData');
$isNotDeclared = ($reflectionMethod->getDeclaringClass()->getName() !== get_class($adpaterInstance));
This way we are sure methods are implemented. We could have use instead isAbstract() method of ReflectionMethod to ensure it has been implemented too.

pimcore/models/Tool/CustomReport/Config.php
mrudnicki: Thanks, @Yivan, I will change it just after an internal discussion about the sense of that change. 
For me, it's a nice feature but I'm not sure, maybe it would be better idea to make configuration PHP file which is responsible for listing additional adapters. In an array, you can specify additional adapters. For example:

`"Microsoft AX orders" => "\Website\Report\Ax"`

I'm not a big fun of solutions which allow the administration panel users specifying the configuration based on the code of the application.

Yivan: Yes i totally agree it is a lot more secure this way. +1 for me

But in the end the best impementation would be like newsletter so both work the same way. But it need to rewrite a little the current reports adapter (sql, analytics) by implementing interface instead abstracted class. And recode js part too to allow report namespace to have adpater added (for listing adapters added in the GUI). This will be the best solution i think as developper will see consistency coding pratic between newletter/report for adding custom newsletter/report mecanism.

fashxp:
I guess we should not go in that direction. 
A better way of including custom adapters for custom reports is to do it like with newsletters. See #1185 and related [docs](https://www.pimcore.org/docs/latest/Tools_and_Features/Custom_Reports.html)

Yivan:
Great! Yes this is better like you do it, because it keep same logic than newsletters.
Thanks a lot.

Issue #788, opened 2016-08-02
Improvement: link custom reports to new newsletters


Yivan:
I just tried the new newsletter document type. This is a nice addition. But actually, setting the receiver using Pimcore datas is tiedous or could be improved. As it is we can:
- use custom reports to export CSV, and open it and after cut/paste using the CSV List
- use Default object list (end-user/marketing administrator will be lost here with SQL request ; ) )

Maybe, having an adapter using custom reports could make end user management more friendly and global (could use any object and not those with only CRM email field).
For instance it could be:
- the user choose a custom reports and a column of this report as email address (could be improved: he could chose "name column", etc... and those used for placeholder/condition already prepared on the newsletter => customized newsletter). It could be a good base and so there is no more the export CSV/copy/paste procedure to do manually. It could be even improved by using extjs list (grid) to filter the receivers directly from the list grid (like we can do when viewing a custom reports).
It will make a powerfull receivers assignement, and those using the fluid extjs GUI.

I think there is something powerfull to take advantage of having datas on Pimcore and the possibility to use them to drive newsletter receveirs list and newsletter content customisation on those datas (field values, personnas, etc.). Atually most of the time we use Mailchimp by sending data via API from Pimcore to Mailchimp. But in fine graining mailling campaign, it would be so great to be able to use directly Pimcore datas to send newsletter directly from Pimcore (for instance customer who has buy some category of product, etc...).
Creating a "Custom reports adapter" and having at least just two field: "custom report name" and "email colum name" could greatly improve this. Later it could be improved (filtering, placeholder issued from the list that can be reinjected in the newsletter in wysiwyg mode...).

Best regards.

fashxp:
actually this is a good idea. we would be very happy to see an pull-request for this one ;-) 

see also the Extending 'Address Source Adapters' section at https://www.pimcore.org/wiki/pages/viewpage.action?pageId=23101460

Yivan:
Hello fashxp, I just made the pull request here https://github.com/pimcore/pimcore/pull/851
I hope it will be merged with the core!
Thanks.

mrudnicki:
https://github.com/pimcore/pimcore/pull/851

PR #851, opened 2016-09-02
custom report adapter for newsletter

Yivan:
This is a pull request for https://github.com/pimcore/pimcore/issues/788

I think it could be a great addition to Pimcore so custom reports and newsletter can be played together. I know the new newsletter as been thinked as extendable so not to added to the core, but i think this adapter will bring really strength to newsletter if it is added to core. And like asked in https://github.com/pimcore/pimcore/issues/788, flashxp said it is a good idea and so i can make a PR on this.

All columns of the report can be used as placeholder in email (like for Default Object List adapter). So it can be very powerfull to send newsletter based on custom report with some column of it used for placeholder in the email (single email sending).

After choosing the "Report adapter" for newsletter, two combobox are provided. Frist one to choose the custom report to use, and the second (it refresh depending of the choosen custom report) to choose the fieldname use as email.

Here are some screenshot:

newsletter_newtype

newsletter_report

newsletter_fieldname

### DIFF
--- a/pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapter/ReportAdapter.php
+++ b/pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapter/ReportAdapter.php
@@ -1,0 +1,147 @@
+<?php
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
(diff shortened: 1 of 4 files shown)


Yivan:
If merged, there is 3 translation to add:
=> "newsletter_report" (adapter name)
=> t("report")
=> t("email_field_name")
Thanks!

mrudnicki:
Thanks, @Yivan, the feature works pretty cool.
I will add translations.

lohitpereira:
I am unable to get the backend to return data. can someone help me with this please?

dvesh3:
@lohitpereira It is always better to post on https://github.com/pimcore/pimcore/discussions when you are seeking support from the community and sharing ideas. This help others as well when they are searching for similar problems 😉

PR #853, opened 2016-09-04
setTimout of symfony process

ctrl-s:
If setTimeout isn't called, video transcoding runs into the (new) default timeout of 60 seconds. Therefore, increased the timeout to infinity


brusch:
Thanks a lot!

PR #856, opened 2016-09-05
Workflowmanagement bugfix foreach on null value

cleggypdc:
Whoops, my fault on this one, I forgot to check the type being returned


fashxp:
thx

PR #859, opened 2016-09-06
Fix file uploads on alpine linux (musl libc)

amonforstmann:
## Fixes
Fix file uploads on alpine linux (musl libc)

If the filename is empty the upload will not work. Therefore use the iconv result only when it is not empty. 

## Additional info  
This is not an ideal fix but at least file uploads work again.


amonforstmann:
This time with the correct email address...

brusch:
Implemented a fallback solution, see: c44f3d08482b102a7f8658c3beb479a6762cb4bd

PR #860, opened 2016-09-09
Updated usage description

ThomasKeil:
The class definition files are not .psf anymore but .php, the usage description is now reflecting this.


brusch:
Thanks!

PR #862, opened 2016-09-13
check if HttpErrorLog plugin exists

chix:
Hi, this is just a check that the HttpErrorLog plugin exists, like on this line https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Controller/Action/Frontend.php#L595. M.


brusch:
Thanks!

PR #866, opened 2016-09-19
Prevents warning it iptcparse fails

axel-tietje:

This change prevents a warning if `iptcparse` should fail and return `false`.

## Additional info  

I encountered a JPEG image with the value `Photoshop 3.08BIM%` in the IPTC field "APP13". The special characters are an EOT (0x04) and a DLE (0x10).

For some reason this lets iptcparse return `false` and PHP throws a `Warning: Invalid argument supplied for foreach()`.


weisswurstkanone:
thx!

PR #874, opened 2016-09-23
Fixes always true condition

kkarkus:
Fixes always true condition in check path `\Admin_ObjectController::addAction`.
I don't see case when it should be always true.


brusch:
Thanks!

PR #876, opened 2016-09-24
proposed patch to fix issue #421

### LINKED ISSUE #421: Document editables: max name length = 750 only ASCII -> warning if not
lets say we have two areas, one called 'nested' and one called 'text'.

this is view.php for text:
```
<div>
    <?php echo $this->wysiwyg("content_wysiwyg"); ?>
</div>
```

this is view.php for 'nested':
```
<div>
    <?php while($this->block('index')->loop()) { ?>
        <div>
            <?php echo $this->areablock('index_block'); ?>
        </div>
    <?php } ?>
</div>
```

i create a page, and insert the areas like this:
>nested 1
>text 1
>nested 2
>>text 1.1

when i preview the page everything is fine. but when i view the same page without pimcore_preview=true text 1.1 is not displayed. the divs are there so the template gets included, but the content is missing.

i tried to trace the bug and landed deep inside the framework but couldn't find out what's going on.

alesak:
This is proposal how to handle unlimited block nesting with full backward compatibility (hopefully)

Fixes #421 

added lookup for for tags by legacy name and new name with md5 checksum
## Additional info  
Our testing shows no backward compatibility issues so far however if you can see any potential problems please let me know so I have chance to improve this patch

Document editables: max name length = 750

### DIFF
--- a/pimcore/lib/Pimcore/View.php
+++ b/pimcore/lib/Pimcore/View.php
@@ -40,7 +40,19 @@ class View extends \Zend_View
 
         try {
             if ($document instanceof Model\Document\PageSnippet) {
-                $tag = $document->getElement($name);
+                $md5Name = Model\Document\Tag::buildTagNameMd5($type,$name, $document);
+                $tag = $document->getElement($md5Name);
+                if (!($tag instanceof Model\Document\Tag)){
+                    $legacyName = Model\Document\Tag::buildTagName($type,$name, $document);
+                    $tag = $document->getElement($legacyName);
+                    if ($tag instanceof Model\Document\Tag){
+                        $name = $legacyName;
+                    }else{
+                        $name = $md5Name;
+                    }
+                }else{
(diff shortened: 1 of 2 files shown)


alesak:
It looks like this breaks areablock copy & paste functionality in editmode, so it will need more work to be complete. Just if somebody from core team could check my approach and comment if this could be the good way to go...

brusch:
Hmm, I don't think that hashing the name is a good idea, it makes it impossible to reconstruct the origin of an editable, but in some cases this is really very handy and necessary, eg. when migrating content after changing a template or similar needs.

alesak:
Yes you are right, can you suggest another options? We need to be able to do almost unlimited nesting and making the names longer and longer is not an option (even without Mysql varchar limit) as it will be very slow and it will quickly reach any limit, it is already painful when area id is ~700 characters long. Maybe we can keep track of origin somewhere else in separate DB field?

PR #883, opened 2016-09-28
Not used method in DocumentController

mkoprek:

This method is not used anywhere.


mrudnicki:
Thanks @mkoprek

PR #892, opened 2016-09-30
Add `class` configuration to all editables

Cruiser13:
## Additional info  

Adding class to input, textarea and other document editables via JS does not seem to work properly (FF latest, Pimcore 4.3.1) and causes issues with external script interaction.
Might be better to inject it directly?


brusch:
See 0188d7f673575b9edaf9c56375f11f03d119266c
Made it generic for all editables + docs

Cruiser13:
Nice solution, thanks!

PR #912, opened 2016-10-05
web2print - prePdfGeneration event to be able to add config options

Yivan:
This pull-request is to add an event before the pdf generation of web2print documents.
Puporse is to be able to inject dynamicaly options to wkhtml/pdfreactor depending of the document.

**One good usage may be the following one.**

Actually for wkhtml2pdf, you can only set page-size globaly. But some document could be A4 but other have another format (visit cards, small flyer, etc.).
So we could set an heritable property on document container like a property name "page-size" for instance.

Using this event hook, developper could grab the document properties and modify pdf generation option by setting $this->option on the "processor" object passed as argument in the event, and modify options as they want before rendering the PDF.

This way, we could have different format for instance without having to change the web2print global config each time we render some type of document having different page size parameters.

### DIFF
--- a/docs/Development_Documentation/10_Extending_Pimcore/11_Event_API_and_Event_Manager.md
+++ b/docs/Development_Documentation/10_Extending_Pimcore/11_Event_API_and_Event_Manager.md
@@ -122,16 +122,17 @@ attach multiple events
 
 ### Document
 
-| Name                               | Target                   | Parameters                            | Description                                                                                              |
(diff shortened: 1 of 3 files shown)


fashxp:
hi,
could you please add the corresponding documentation changes to your pull request. 
Otherwise your pull request is fine. 

Thank you very much.

Yivan:
Thanks fashxp,
I added documentation changes, and the getter/setter to be able to change options (is a private attribut) when hooking the event.

fashxp:
thx very much.

PR #916, opened 2016-10-05
When loading the Pimcore admin without Google maps API Key there is a error

jorisros:
Fixes # 
When you loading the admin and there is no API Key filled in the settings, there is a error in de javascript console.

* Added a check in the layout template to show only the Google maps initialisation script of there is a key defined.
* Showed a error message box  when you try to enter a geopoint when there is no Key defined

## Additional info  
Added at this point only the english translation


brusch:
Thanks for your PR! 
In terms of consistency, it would be good the have the check also for all the other geo data-types (geobounds, geopolygon) as well as for the ExtJS 3.4 user interface. 
Could you please update your PR accordingly? Btw. please sign the CLA above ;-)

brusch:
Doesn't seem that anything changed: 
https://github.com/pimcore/pimcore/pull/916/files

Thanks for looking into it :) 
(Please also don't forget about the CLA)

jorisros:
The last commit contain the remarks you requested. I added the check also for the extjs3.4 interface and for geobounds and geopolygon

PR #918, opened 2016-10-06
Allow the onlyDirectChildren flag to be set to false from a custom gr…

BlackbitNeueMedien:
In an object's children grid, the "onlyDirectChildren" flag can be set to true by default in [/pimcore/static6/js/pimcore/object/object.js#L279](https://github.com/pimcore/pimcore/blob/7fd56888e21accbe42c684ce9525daeea507daf0/pimcore/static6/js/pimcore/object/object.js#L279). This prevents the flag from being reset to false from a users custom grid config. This change fixes the issue.


brusch:
Thanks!

BlackbitNeueMedien:
On second thought, this might not be correct after all, as it will always overwrite the default value. I guess the indented behavior is to only set the flag if it is included in the reponse object. If so, the assignment should be wrapped in `if (Ext.isDefined(response.onlyDirectChildren)) ...` or something similar.

PR #919, opened 2016-10-06
Grid: Fix for displaying filter condition with operator ">"

podarcis:
The filter condition was shown incorrectly when using the Greater-than-operator. Only ExtJS 6 (static6) is affected.


brusch:
Thanks!

PR #936, opened 2016-10-12
Fix reload document behaviour - href tag

kkarkus:
Currently reload document works only when you drag&drop object into field, but when you change field by search widget (context menu > search) it won't work.


brusch:
Thanks!

PR #941, opened 2016-10-14
do not show frontend http auth in backend

chix:
Hi, imho it's a bit annoying that the frontend http authentication configurable in system settings pops also on documents in backend (editmode and preview). So maybe we could suppress it on frontend requests by admin? Here is a PR if you agree :).


brusch:
Thanks!

PR #956, opened 2016-10-18
Fixed handling of empty-check on field

ThomasKeil:
`empty()` doesn't work on on fields that can store values that trigger this, like `0`, "0" or `false`.

This is especially the case when setting checkboxes to `false`; with activated inheritance the Object Class returns the parent's value:

`if(Object_Abstract::doGetInheritedValues() && $this->getClass()->getFieldDefinition("NAME")->isEmpty($data)) { return $this->getValueFromParent("NAME");}`

The handling of `empty()` seems to be a consistent problem in Pimcore...

### DIFF
--- a/pimcore/models/Object/ClassDefinition/Data.php
+++ b/pimcore/models/Object/ClassDefinition/Data.php
@@ -960,11 +960,7 @@ abstract class Data
      */
     public function isEmpty($data)
     {
-        if (empty($data)) {
-            return true;
-        }
-
-        return false;
+        return empty($data) && !is_numeric($data) && !is_bool($data);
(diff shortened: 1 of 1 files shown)


brusch:
I'm not sure if this should be handled globally - no idea which side-effects this has :( 
So I think it's better to implement isEmpty() in `Pimcore\Model\Object\ClassDefinition\Data\Checkbox` like other data-type do.

ThomasKeil:
OK. Will do :-)

ThomasKeil:
Continued at https://github.com/pimcore/pimcore/pull/957

PR #957, opened 2016-10-18
Fix handling of Checkbox::isEmpty() check

ThomasKeil:
Checkboxes store `true`/`false`, which in case of `false` wrongly triggers php's `empty()` function.

### DIFF
--- a/pimcore/models/Object/ClassDefinition/Data/Checkbox.php
+++ b/pimcore/models/Object/ClassDefinition/Data/Checkbox.php
@@ -275,4 +275,13 @@ class Checkbox extends Model\Object\ClassDefinition\Data
 
         return "IFNULL(" . $key . ", 0) = " . $value . " ";
     }
+    
+    /**
+     * @param Object_Concrete $data
+     * @return bool
+     */
+    public function isEmpty($data)
+    {
+        return empty($data) && !is_bool($data);
+    }
 }


brusch:
Hmm, when `$data` is `false` `isEmpty()`returns `true` -> this will break backward-compatibility when using inheritance. 

Current behavior of checkboxes is that the parent value is inherited when value != `true`. 
What is the actual problem why you want to change the behavior? 
The problem I see is: 
Currently there's no way of resetting the value of a checkbox to `null`, so currently a checkbox has only to possible states: 
- checked, represented by `true`
- unchecked, represented by `false` or `null`

ThomasKeil:
The current behavior is inherited from Data:

```
    public function isEmpty($data) {
        if(empty($data)) {
            return true;
         }
         return false;
    }
```

which is essentially

```
    public function isEmpty($data) {
        return empty($data);
    }
```

since empty() returns a boolean:
> Returns FALSE if var exists and has a non-empty, non-zero value. Otherwise returns TRUE.

The situation is that `isEmpty($data)` returns `true` if the variable is `null` or `false`, and both are valid values. 

In case you uncheck the checkbox or set it programatically to `false`, a `false` is stored, which in turn is thus interpreted as "empty" and the inheritance kicks in in the object class:

```
if(Object_Abstract::doGetInheritedValues() && $this->getClass()->getFieldDefinition("foo")->isEmpty($data)) { return $this->getValueFromParent("foo");}
```

brusch:
Yes, but if i merge the PR now, we'll change the behavior.
Currently values are inherited if the value is `null` or `false`, with this PR, inheritance would be only with `null`, but then again we have the problem that an editor is not able to set the value of a checkbox back to `null`. 

So maybe the only thing that makes sense is to get rid of the `null` state in checkboxes.

PR #963, opened 2016-10-19
throw Exception if password is under 5 characters long

renepenner:
throw Exception if password is under 5 characters long instate of ignore user input

Update User.php

### DIFF
--- a/pimcore/models/User.php
+++ b/pimcore/models/User.php
@@ -118,13 +118,15 @@ class User extends User\UserRole
 
     /**
      * @param string $password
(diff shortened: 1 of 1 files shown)


brusch:
@renepenner Thanks for the fix! Please sign the CLA first, so that we can merge your PR. Thanks!

renepenner:
is signed ... i think ;)

brusch:
Reverted ... reason: With this fix it's not possible to save the user without explicitly setting the password again.

PR #971, opened 2016-10-20
Add query params to store when changing class in search box

manea-eugen:
This code will fix the issue that when changing the class backend search functionality will ignore query parameter, user needs to click on search button in order to actually filter the result based on query

### DIFF
--- a/pimcore/static6/js/pimcore/element/selector/object.js
+++ b/pimcore/static6/js/pimcore/element/selector/object.js
@@ -271,6 +271,7 @@ pimcore.element.selector.object = Class.create(pimcore.element.selector.abstract
         var gridHelper = new pimcore.object.helpers.grid(selectedClass, fields, "/admin/search/search/find", null, true);
         this.store = gridHelper.getStore();
         this.store.setPageSize(pimcore.helpers.grid.getDefaultPageSize());
+        this.applyExtraParamsToStore();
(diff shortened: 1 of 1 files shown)


brusch:
Sorry, can you explain the issue a little bit more in detail? 
Thanks!

manea-eugen:
What i am trying to avoid is this 
Steps: 
1. Giving the scenario i have an object News with key "lorem-ipsum"
screen shot 2016-10-21 at 09 11 15
2. I open the search box i type in "lorem-ipsum" and i press "Search"
screen shot 2016-10-21 at 09 12 36
3. I see a list a list of all objects containing "lorem-ipsum"
4. Now i want to limit my search by class "News" and i select "News" from the class selector
screen shot 2016-10-21 at 09 16 46
5. Now i get list of all "News" objects, but what i would expect is a list of "News"  objects filtered by my original query "lorem-ipsum"
6. To get my actual result i need to press "Search"
7. Now my list contains a list of "News" objects that contains "lorem-ipsum" 
screen shot 2016-10-21 at 09 18 49

I think we would easily fix this by always taking in consideration the query, also because if search box is empty then the pimcore default behavior is to search all 
screen shot 2016-10-21 at 09 22 06

Please let me know if the explanation suffices

brusch:
Thank you very much for your detailed explanation and for the fix.

PR #976, opened 2016-10-21
Nice path request via post

manea-eugen:
I think is safer to use POST instead of default GET because if we have the scenario that an object is linked to many objects inside multi-object field then the request can exceed the maximum of ~ 2000 character url limit 

In my test scenario i got a url with +40000 characters


brusch:
Yep, thanks!

PR #982, opened 2016-10-24
added a description column on document properties

lukas-schnieper:
Properties are often added using the "predefined properties". The Dropdown shows the Description which is not shown in the grid. So you add something and sometimes you are not sure what you added, because it may not have the same name. Therefore i added a new column which is filled with the description from the predefined property, if there is one with the same key.

screenshot from 2016-10-24 11-07-23


brusch:
Implemented it a slightly different way to avoid additional HTTP requests, see #983

PR #990, opened 2016-10-25
ExtJS6 Fix notification icons & window width behaviour

kkarkus:
Notifications except "success" type wont' show icon because in ext6 css there's no icon_notification_* classes, so I added pimcore_icon_success also set window max width and made it closable.


brusch:
Thanks!

PR #997, opened 2016-10-27
Fixes bug that corrupts cache

Sunohara666:

Fixes # 

## Additional info  

This code fixes a nasty bug that corrupts pimcore cached objects.

To reproduce the bug:

- Create class with DateTime field (startdate).
- Clear the cache
- In code do this:
- $object = Object::getById(666);  // just an example
- $dt = $object->getStartdate();  // get get a REFERENCE to the startdate here
- $dt->add(new \DateInterval('P1D'));  // we change the startdate of the object because we chage the reference
- // we never save the changed object
- Now the object with the altered (but not saved!) start date gets cached!

### DIFF
--- a/pimcore/lib/Pimcore/Cache.php
+++ b/pimcore/lib/Pimcore/Cache.php
@@ -342,6 +342,9 @@ class Cache
      */
     public static function save($data, $key, $tags = [], $lifetime = null, $priority = 0, $force = false)
     {
+        if(is_object($data)){
+            $data = clone $data;
+        }
         if (self::getForceImmediateWrite() || $force) {
             if (self::hasWriteLock()) {
                 return;


brusch:
Thanks Roelf!

fashxp:
hmm, this causes another problem in combination with pimcore and zend navigation. 
since in https://github.com/zendframework/zf1/blob/master/library/Zend/View/Helper/Navigation/Breadcrumbs.php#L216 an object comparison is done, the object reference might change due to the clone and therefore the comparison fails. This results in a broken navigation. 

I made some changes, can you please verify that these changes do not break the initial fix.

fashxp:
hmm, ok it seems that the fix doesn't solve the problem at all since clone is not recursive, see also
http://php.net/manual/en/language.oop5.cloning.php
'When an object is cloned, PHP 5 will perform a shallow copy of all of the object's properties. Any properties that are references to other variables will remain references.'

also cloning the object when saving it to the cache (after the change is made) will not have any affect on the problem since the change is already made when the clone is created.

fashxp:
I reverted your change now, we need to find another solution for that issue.
any ideas?

Sunohara666:
make sure the getters on objects dont return references maybe?

brusch:
I think the only way of fixing this issue properly would be by implementing `__clone()` consistently. 
But interesting though that your fix caused issues with the navigation ...

PR #998, opened 2016-10-27
Notifications

kkarkus:
Notifications is a new feature designed for panel users.
User see only notifications assigned to his account. 
Every 10 seconds JS check for new notifications and if there're new it shows badge on icon with number of unread notifications and shows notification popup with possibility to open details, mark as read or just close notification. Feature also includes portlet for dashboard.
User can get notifications from system, e.g.: when workflow status changes, just create new notification instance filled with data and save it.

image
image


brusch:
Since this is a quite huge contribution and has also effects on our general strategy we have to discuss this in detail internally. Thanks for your understanding :)

solverat:
my two cents: Instead of long polling we should use better technologies like node.js or even the browser native WebSocket.

brusch:
+1 for websockets
And maybe it's better to have this implemented as a plugin in the first place.

dpfaffenbauer:
Websocket would need another requirement or at least a running PHP service to handle the request. But I like the idea of a Notification Center. You could combine this with workflows as well

kkarkus:
Moved to plugin: https://github.com/DivanteLtd/pimcore-notifications

brusch:
@kkarkus great - thanks!

PR #1012, opened 2016-11-02
fix broken interval in imagecropper and remove selector and broken im…

solverat:
- go to demo.pimcore.org
- create object with advanced image
- try to open the imagecropper window without choosing an image:
bildschirmfoto 2016-11-02 um 14 21 28

- try to close the imagecropper window: an infinite invalid setInterval starts
bildschirmfoto 2016-11-02 um 14 21 32

this PR fixes both.


brusch:
Thanks!

PR #1034, opened 2016-11-13
Update Editmode.php

jjonas-gcs:
Detection of body tag fails if you have some inline styles, for example:
<body style="height: 100%; width:100% !important; -webkit-text-size-adjust:100%; -ms-text-size-adjust:100%; margin:0; padding:0; background: #ffffff";">
And you need inline styles for html mails...

### DIFF
--- a/pimcore/lib/Pimcore/Controller/Plugin/Frontend/Editmode.php
+++ b/pimcore/lib/Pimcore/Controller/Plugin/Frontend/Editmode.php
@@ -284,9 +284,9 @@ class Editmode extends \Zend_Controller_Plugin_Abstract
             $html = $this->getResponse()->getBody();
 
             if ($html) {
-                $htmlElement = preg_match("/<html[^a-zA-Z]?( [^>]+)?>/", $html);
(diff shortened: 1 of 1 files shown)


brusch:
Can't see any problem with the current regex and your given example: 
`<body style="height: 100%; width:100% !important; -webkit-text-size-adjust:100%; -ms-text-size-adjust:100%; margin:0; padding:0; background: #ffffff";">` 

Check out:  
http://www.phpliveregex.com/p/hRA 
also tested it on our testing environment without any problems.

brusch:
The reason why we're using a regex is because of the `<header>` tag, which would be also a match if using just `strpos()`. 

See also: 
https://github.com/pimcore/pimcore/commit/bba1519c8f22d1c589aa7cdfa60ffdd7fb9a55d9

jjonas-gcs:
My string is:
`body
style="foo:bar;"`
No whitespace after body and before style, only a line break between them This will not be matched by the current regex. My example yesterday was bad, sorry for that,. Maybe we can use this:
<body(\s+[^<>]*)?>

PR #1050, opened 2016-11-21
Avoid warning when passing empty string to md5 function

manea-eugen:
When passing an empty string to getElementsTagsInWysiwyg in Pimcore\Tool\Text, md5 function shows an unnecessary warning, this can be avoided by checking that we are dealing with a string and the string is not empty


brusch:
Thanks for your PR. 
Just made the check a bit shorter :)

manea-eugen:
cool !

PR #1058, opened 2016-11-22
EmailLog / buildArray returns invalid recipients when $data is NULL

cleggypdc:
See here: https://3v4l.org/hrFVT

This causes errors resending mail when CC and Bcc are not set


brusch:
Thanks!

PR #1060, opened 2016-11-22
Fixes MultihrefMetadata phpdoc (return type) generation.

kkarkus:
Right now behaviour is like Multihref (objects, document or assets), but it should return 
```
\Pimcore\Model\Object\Data\ElemenentMetadata[]
```


weisswurstkanone:
thx!

PR #1072, opened 2016-11-23
Removed unnecessary if-condition in Frontend Router

d-gerken:
- Removed a not needed if condition in Frontend Router

## Additional info  
In Line 94 $matchFound is set to false and until line 148 it can't be set to another value. checkForRedirect will be always executed. So the if condition is not necessarily.


brusch:
Thanks!

PR #1074, opened 2016-11-24
ClassDefinition DateTime: Use Carbon for PHPdoc as well

podarcis:
When not using the compat-layer (useZendDate) Date and DateTime hold an instance of Carbon. The auto-generated class files for DateTime still used the PHPdoc for \Pimcore\Date.


brusch:
Thanks!

PR #1080, opened 2016-11-25
Added info on admin and website translations

schuttelaar:

Fixes #


fashxp:
hmm, wouldn't be here a better place for this information - or isn't this information already in there? 
https://github.com/pimcore/pimcore/blob/master/docs/Development_Documentation/06_Multi_Language_i18n/README.md#pimcore-backend-ui-localization
or the sub document https://github.com/pimcore/pimcore/blob/master/docs/Development_Documentation/06_Multi_Language_i18n/07_Admin_Translations.md#admin-translations

schuttelaar:
I have examined the pull request and do think that this is the right place. The text is very specific for some of the intricacies of working with localized fields on objects, and also the most obvious place for a developer to look when running into problems. It does not make sense to look at the page "admin translations" when you get an error "language not found" when saving an object with localized fields.

PR #1087, opened 2016-11-29
Add event admin.document.get.preSendData to PrintPages

Pegasuz:
The workflow component uses the "admin.document.get.preSendData" event to add data about the workflow. Currently the Printpage overrides the getDataByIdAction action and does not add this event.

This PR adds that functionality. It is already tested and worklow component does work on print pages after this change.


cleggypdc:
Looks good to me 👍

fashxp:
thx very much ... changed the title to a more generic one, since this is not just for workflows.

PR #1096, opened 2016-12-02
Symptomatically fixes 1095. In case removing the Alpha-Channel fails,…

### LINKED ISSUE #1095: Imagick fails to set alpha channel when generating treeview-thumbs
`core.ERROR: exception 'ImagickException' with message 'Unable to set image alpha channel' in public/pimcore/lib/Pimcore/Image/Adapter/Imagick.php:186`

Happens, when generating the treeview Image. In console as well as the Ui. The Folder for the thumbnail stays empty. Other Thumbs/ImagePipelines are generated correctly.

It's PHP5.6 from dotdeb. And since this error occurred with the dotdeb package, which has a quite old Imagick extension, we updated Imagick from PECL to Version 3.4.3RC1 with no luck.

Pimcore Version 4.3, but the Imagick adapter didn't change in 4.4.

Seems to happen primarily (don't know if exclusively) when converting PNG to JPEG for the Treeview.

bbak:
… the Image is written with Alpha-Channel.

Fixes #1095 

wrap `Imagick::setImageAlphaChannel();` in a try/catch Block and write a Warning to log if ImagickException is thrown.

**Note:** Just fixes the Symptom (no Image created), not the cause.

### DIFF
--- a/pimcore/lib/Pimcore/Image/Adapter/Imagick.php
+++ b/pimcore/lib/Pimcore/Image/Adapter/Imagick.php
@@ -175,15 +175,19 @@ class Imagick extends Adapter
         if (in_array($format, ["jpeg", "pjpeg", "jpg"]) && $this->isAlphaPossible) {
             // set white background for transparent pixels
             $i->setImageBackgroundColor("#ffffff");
-
-            // Imagick version compatibility
-            $alphaChannel = 11; // This works at least as far back as version 3.1.0~rc1-1
-            if (defined("Imagick::ALPHACHANNEL_REMOVE")) {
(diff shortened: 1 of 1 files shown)


bbak:
The Treeview Thumbs use `/pimcore/static6/img/tree-preview-transparent-background.png` as Background. And a `convert pimcore/static6/img/tree-preview-transparent-background.png -verbose info:` reveals, that this image really has no Alpha-Channel (and there's no bug to work around). 

Therefore there's a better Patch by calling Imagick::getImageAlphaChannel() instead of try/catch. Closing this one.

PR #1097, opened 2016-12-02
Only try to remove AlphaChannel, if Image has an AlphaChannel

### LINKED ISSUE #1095: Imagick fails to set alpha channel when generating treeview-thumbs

bbak:
Fixes #1095

Check if there is an AlphaChannel before trying to remove it. Using `Imagick::getImageAlphaChannel()` which returns (int) 0 in PHP, if the Image has no AlphaChannel. 
This is needed for the Tree-Preview Thumbnails, which use `pimcore/static6/img/tree-preview-transparent-background.png` as Base/Background. 
And this image has no AlphaChannel, which can be checked in the Shell with this command:
`convert pimcore/static6/img/tree-preview-transparent-background.png -verbose info:`

### DIFF
--- a/pimcore/lib/Pimcore/Image/Adapter/Imagick.php
+++ b/pimcore/lib/Pimcore/Image/Adapter/Imagick.php
@@ -176,14 +176,16 @@ class Imagick extends Adapter
             // set white background for transparent pixels
             $i->setImageBackgroundColor("#ffffff");
 
-            // Imagick version compatibility
-            $alphaChannel = 11; // This works at least as far back as version 3.1.0~rc1-1
-            if (defined("Imagick::ALPHACHANNEL_REMOVE")) {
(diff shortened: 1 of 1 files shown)


maff:
Thanks!

PR #1103, opened 2016-12-06
[fix] load geoip.js if Personas use Country as condition

lorextera:
Fixes # 

Fix the problem of missed location when Personas or List of Targets use Country as condition


brusch:
Thanks!

PR #1107, opened 2016-12-07
Use Cache::save() instead of protected method Cache::storeToCache()

xlttj:
This commit fixes a fatal error during cache warming:

```
Fatal error: Call to protected method Pimcore\Cache::storeToCache() from context 'Pimcore\Cache\Tool\Warming' in /pimcore/lib/Pimcore/Cache/Tool/Warming.php
```

This fix uses Cache::save instead of Cache::saveToCache


weisswurstkanone:
thx!

PR #1141, opened 2016-12-20
Refactor Workflow Model and create GUI to configure them

dpfaffenbauer:
This Pull-Requests converts Workflow Model to a Pimcore\Model and adds a GUI for configuration. It also adds a postupdate script to convert old config file to new one. (to be conform with all pimcore config files).

Reason for a GUI to configure Workflows: It is much more convenient and intuitive to handle.

This PR removes "Pimcore\WorkflowManagement\Workflow" and moves it to "Pimcore\Model\Workflow". If someone uses "Pimcore\WorkflowManagement\Workflow", there is now an alias for that, but I think the alias should be removed as soon as possible.


dpfaffenbauer:
here are some screenshots:

bildschirmfoto 2016-12-20 um 17 25 49
bildschirmfoto 2016-12-20 um 17 25 52
bildschirmfoto 2016-12-20 um 17 25 56
bildschirmfoto 2016-12-20 um 17 25 59
bildschirmfoto 2016-12-20 um 17 26 15
bildschirmfoto 2016-12-20 um 17 26 11

fashxp:
Hi Dominik, 
thank you very much for your input. 
The GUI is very cool indeed. On the other hand we think that configuring a workflow is an absolute admin task and also the configuration GUI is quite complex and you really need to know what to do. 
Also it is another GUI part the core team needs to take care of :). 

What would you think about following idea? 
We put the migration of the workflow configuration to a model into the core and the configuration GUI can live in a Pimcore plugin? 

Cheers...

dpfaffenbauer:
I am happy with that. 

But I still think Pimcore makes a great job of having a GUI for almost everything. But I can understand why not. I can live with a plugin as well. Still happy my workflow refactoring gets commited ;)

How should we proceed? Should I create a new PR for only refactoring stuff or do you do it yourself?

fashxp:
hi, if you could create a pull request (or adapt this) it would be great!

thx...

dpfaffenbauer:
see #1169

PR #1143, opened 2016-12-21
Call update video panel in another way - see #1121

### LINKED ISSUE #1121: objects > video datatype in hidden accordion > exception after 30sec
1) create a object class with an accordion and a video data field in a hidden panel
2) create and open object
3) wait for about 30sec 
4) open hidden panel > pimcore crashes with extjs exception

Quinten:
Fixes #1121 

`this.getBody().update(content);`
seemed to mess up the dom. It removed the inner div's of the classes 'x-autocontainer-outerCt' and 'x-autocontainer-innerCt', which caused Ext.js to crash.
If you update the html of the video panel component like this:
`this.component.update(content);`
It will keep the dom elements in place.

I am not sure if this.getBody() returns the right object, but it is certainly related.


fashxp:
thank you very much

fashxp:
got rid of the getBody as a whole... see https://github.com/pimcore/pimcore/commit/09908717a654df1876cba1c811d1a1a00699f5e7

Quinten:
You're welcome ;)

PR #1145, opened 2016-12-21
#1139 Use ReflectionClass to list available actions

### LINKED ISSUE #1139: Improvement: Use ReflectionClass to List Available Actions in Controller Parent Classes
I suggest to use PHPs built-in ReflectionClass to list a controllers available actions in `Admin_MiscController::getAvailableActionsAction` in order to list inherited actions from parent class(es), too. Example:

```
    public function getAvailableActionsAction()
    {
        $actions = [];
        $controller = $this->getParam("controllerName");
        $controllerClass = str_replace("-", " ", $controller);
        $controllerClass = str_replace(" ", "", ucwords($controllerClass, '_'));
        $reflectionClass = "{$controllerClass}Controller";
        $controllerClass = preg_replace_callback("/([_])([a-z])/i", function ($matches) {
            return "/" . strtoupper($matches[2]);
        }, $controllerClass);
        $controllerDir = $this->getControllerDir();
        $controllerFile = $controllerDir . $controllerClass . "Controller.php";

        if (is_file($controllerFile)) {
         require_once $controllerFile;
         $oReflectionClass = new ReflectionClass($reflectionClass);
         $methods = $oReflectionClass->getMethods(ReflectionMethod::IS_PUBLIC);
         $methods = array_filter($methods, function($method) {
          /** @var ReflectionMethod $method */
          return preg_match('/^([a-zA-Z0-9]+)Action$/', $method->getName());
         });
         $actions = array_map(function($method){
          /** @var ReflectionMethod $method */
          return $method->getName();
         }, $methods);
        }

        $this->_helper->json([
            "data" => $actions
        ]);
    }
```

Are there any concerns against that?

xlttj:
Fixes #1139 

This commit uses PHPs built-in ReflectionClass to more reliably list a controllers available actions including inherited methods. Example:

```
// file /lib/Website/Controller/Action.php
namespace Website\Controller;
use Pimcore\Controller\Action\Frontend;

class Action extends Frontend {
    public function init() {}
}

// file /lib/Website/Controller/ContentController.php
namespace Website\Controller;
use Website\Controller\Action;

class ContentController extends Action {
    public function init() { parent::init(); }
    public function fooAction() {}
}

// file /website/controllers/ContentController.php
class ContentController extends Website\Controller\ContentController {
    // some comment: function doesntExistAction is shown if using regex
    public function barAction()
}
```

Current method with regex would list `doesnt-exist` and `bar`, while using ReflectionClass correctly lists `foo` and `bar`.

### DIFF
--- a/pimcore/modules/admin/controllers/MiscController.php
+++ b/pimcore/modules/admin/controllers/MiscController.php
@@ -501,7 +501,10 @@ class Admin_MiscController extends \Pimcore\Controller\Action\Admin
             });
             $actions = array_values(array_map(
                 function (\ReflectionMethod $method) {
-	                return ["name" => preg_replace('/Action$/', '', $method->getName())];
+	                $name = preg_replace('/Action$/', '', $method->getName());
+	                $filter = new \Zend_Filter_Word_CamelCaseToDash();
+	                $name = $filter->filter($name);
+	                return ["name" => $name];
                 }, $methods
             ));
         }
--- a/pimcore/modules/admin/controllers/MiscController.php
+++ b/pimcore/modules/admin/controllers/MiscController.php
@@ -501,7 +501,7 @@ class Admin_MiscController extends \Pimcore\Controller\Action\Admin
             });
             $actions = array_values(array_map(
                 function (\ReflectionMethod $method) {
-                    return ["name" => $method->getName()];
+	                return ["name" => preg_replace('/Action$/', '', $method->getName())];
                 }, $methods
             ));
         }
--- a/pimcore/modules/admin/controllers/MiscController.php
+++ b/pimcore/modules/admin/controllers/MiscController.php
@@ -484,6 +484,7 @@ class Admin_MiscController extends \Pimcore\Controller\Action\Admin
         $controller = $this->getParam("controllerName");
         $controllerClass = str_replace("-", " ", $controller);
         $controllerClass = str_replace(" ", "", ucwords($controllerClass));
+        $reflectionClass = "{$controllerClass}Controller";
         $controllerClass = preg_replace_callback("/([_])([a-z])/i", function ($matches) {
             return "/" . strtoupper($matches[2]);
         }, $controllerClass);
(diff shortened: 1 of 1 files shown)


pimcore/modules/admin/controllers/MiscController.php
```diff
@@ -484,19 +484,26 @@ public function getAvailableActionsAction()
...
+            require_once $controllerFile;
+            $oReflectionClass = new \ReflectionClass($reflectionClass);
+            $methods = $oReflectionClass->getMethods(\ReflectionMethod::IS_PUBLIC);
+            $methods = array_filter(
+                $methods, function (\ReflectionMethod $method) {
+                return preg_match('/^([a-zA-Z0-9]+)Action$/', $method->getName());
+            });
+            $actions = array_values(array_map(
+                function (\ReflectionMethod $method) {
+	                return ["name" => preg_replace('/Action$/', '', $method->getName())];
```
brusch: I think the following functionality is missing here, right? 
`$dat["name"] = strtolower(preg_replace("/[A-Z]/", "-\\0", $match));`

This transforms eg. `fooBar` to `foo-bar` which is the correct format for actions. 
Can you update your PR accordingly? Thanks!

dpfaffenbauer:
I would rather use Zend_Filter_Word_CamelCaseToDash, wouldn't you?. Basically does the same, but is more Zend'ish. (Also available in later Zend Versions)

brusch:
Oh yes, didn't know about this :)

xlttj:
OK, thanks, I'll update the PR.

xlttj:
Done.

brusch:
Thanks a lot!

xlttj:
Thanks for merging :)

brusch:
ok, this wasn't entirely correct, `strotolower()` was missing, but anyway, just fixed it :)

xlttj:
Right, had changed it in my test instance but forgot to copy it back…

PR #1148, opened 2016-12-22
Escaped wildcards in MySQL LIKE Statements.

sebastianMoser:
Wildcards in MySQL LIKE Statements must be escaped. Otherwise wrong classes will be deleted. Unfortunately happened in our PROD environment today.

Eg:
- Delete the class with object Id 6
- The previous LIKE Statement object_collection_%_" . $this->model->getId() . " will also delete the following collection: object_collection_whatever_26
- The _ matches exactly one character in MySQL LIKE Statements. So the _ in the example above will be matched with the %, the non-escaped _ will match the value 2 and the model-Id is 6

### DIFF
--- a/pimcore/models/Object/ClassDefinition/Dao.php
+++ b/pimcore/models/Object/ClassDefinition/Dao.php
@@ -252,29 +252,29 @@ class Dao extends Model\Dao\AbstractDao
         $this->db->delete("objects", $this->db->quoteInto("o_classId = ?", $this->model->getId()));
 
         // remove fieldcollection tables
-        $allTables = $this->db->fetchAll("SHOW TABLES LIKE 'object_collection_%_" . $this->model->getId() . "'");
+        $allTables = $this->db->fetchAll("SHOW TABLES LIKE 'object\_collection\_%\_" . $this->model->getId() . "'");
         foreach ($allTables as $table) {
             $collectionTable = current($table);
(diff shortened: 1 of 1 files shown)


pimcore/models/Object/ClassDefinition/Dao.php
```diff
@@ -252,29 +252,29 @@ public function delete()
...
 
-        $allTables = $this->db->fetchAll("SHOW TABLES LIKE 'object_localized_query_" . $this->model->getId() . "_%'");
+        $allTables = $this->db->fetchAll("SHOW TABLES LIKE 'object\_localized\_query\_" . $this->model->getId() . "_%'");
         foreach ($allTables as $table) {
             $queryTable = current($table);
             $this->db->query("DROP TABLE IF EXISTS `".$queryTable."`");
         }
 
-        $this->db->query("DROP TABLE IF EXISTS object_localized_data_" . $this->model->getId());
+        $this->db->query("DROP TABLE IF EXISTS object\_localized\_data\_" . $this->model->getId());
```
maff: This one shouldn't be needed as it is no LIKE query.

maff:
Thanks for the PR! Could you remove the change on the DROP line where escaping is not necessary?

sebastianMoser:
You are right - I've changed the line.

maff:
Thanks!

PR #1154, opened 2016-12-24
added ratioX and ratioY from image-tag options to image imagecropper

lukas-schnieper:
In class definition i can use crop in advanced images to make sure the selected part of the image has the correct aspect ratio. It would be useful to use this in the image-editable to.

## Usage
echo $this->image("img", [
        "thumbnail" =>  "myimagethumbnail",
        "ratioX" => 370,
        "ratioY" => 208
]);

## Todo if accepted
Update docucemtation
https://www.pimcore.org/docs/latest/Documents/Editables/Image.html


brusch:
Good addition. 
Please add the docs and update your PR. 
Thanks!

PR #1156, opened 2016-12-27
show path of link if no linktext is set instead of not_set message

lukas-schnieper:

Sometimes i use the link editable to be more flexible for links, for example if the target has to be a document or also an external link. But the text part has to be more flexible (for example an wysiwyg or even more). So i don't use the text field of the link. In this case it shows the "Not set" message in the editmode, altough the link is actually set. This is a bit confusing, therefore is suggest to show the path of the link if it is set.


fashxp:
thx very much.

PR #1168, opened 2017-01-08
fix calculated value localized fields getter code

chix:
Hi,

this PR fixes a small namespace bug in calculated value localized fields getter code generator. Maybe a post-update script which rebuilds classes with such calculated value fields should also be added.

Martin


brusch:
Thanks!

PR #1169, opened 2017-01-09
Refactor Workflow Model

dpfaffenbauer:
This PR refactored workflow models to be more Pimcore'ish. The configuration now lives within \Pimcore\Model Namespace and uses standardized pimcore config file.

The GUI is implemented as plugin and available here: https://github.com/dpfaffenbauer/pimcore-WorkflowGUI

See #1141


maff:
Thanks!

PR #1174, opened 2017-01-11
fixes hostUrl for newsletter send command

dpfaffenbauer:
Fixes hostUrl argument in internal-newsletter-document-send command. This argument was ignored until now.

Problem when not using hostUrl: 
1. Multi-Site Newsletter would never be able to display images cause the HostUrl is unknown.
2. When no Main-Domain is configured, images would also not be able to be displayed.


brusch:
Thanks!

Issue #1160, opened 2016-12-28
Adding pimcore.document.tags.wysiwyg.defaultEditorConfig like for object


Yivan:
To easily change global configuration of ckeditor for documents, `pimcore.document.tags.wysiwyg.defaultEditorConfig` could be added and merged during initialization of the wysiwyg for document like for object (see pimcore.object.tags.wysiwyg.defaultEditorConfig, https://github.com/pimcore/pimcore/blob/master/pimcore/static6/js/pimcore/object/tags/wysiwyg.js#L159).

Actually, there seems no way to globally change ckeditor configuration for document. Even doing a CKEDITOR.config in a plugin using pimcoreReady don't work as explained here: https://github.com/pimcore/pimcore/issues/1157

Thanks.

solverat:
## documents
just add a custom config.
```html
<section id="marked-content">
    <?= $this->wysiwyg('specialContent', [
        'height' => 200,
        'customConfig' => '/custom/document-editor-settings.js'
    ]); ?>
</section>
```

add something like this to your config file (in this example, it's called `document-editor-settings.js`):
```js
CKEDITOR.editorConfig = function( config ) {
        editor.config.allowedContent = undefined;
        editor.config.foo = bar;
});
```

## objects

create a plugin, add a js file.

```xml
<pluginJsPaths-extjs6>
     <path>/object-editor-settings.js</path>
</pluginJsPaths-extjs6>
```

add something like this to your config file (in this example, it's called `object-editor-settings.js`) :

```js
CKEDITOR.on('instanceCreated', function( event ) {
    var editor = event.editor;
    editor.on('configLoaded', function() {
        editor.config.allowedContent = undefined;
        editor.config.foo = bar;
    });
});
```
## some notes
- #1157 is not a bug :)
- to allow greater flexibility within the editor configuration, we changed some settings. see https://github.com/pimcore/pimcore/pull/720

Yivan:
Thanks solverat for your detailled explaination.

For object, yes it can be globally managed.
For document, yes we know about custom JS file, but all view scripts need to be updated. We were looking for a global configuration.

I think adding "pimcore.document.tags.wysiwyg.defaultEditorConfig" and add it in the Pimcore documentation (like for object) can be a nice addition to the custom JS file per wysiwyg like you explained.

Yivan:
If it can help, I can make a pull request on this issue. Is it ok if i do one or this is in pending ?

[...]

PR #1178, opened 2017-01-11
added defaultEditorConfig for documents and documentation is updated

Yivan:
This is the pull request for issue https://github.com/pimcore/pimcore/issues/1160.

With this, it is now possible to globally change in just one place document editable wysiwyg configuration.

Thanks!


brusch:
Thanks!

PR #1184, opened 2017-01-12
Obj events

kylepwarren:

Fixes # 

more control over js request, server response. Specifically for Objects [events *for plugins]. 
i.e. increased programmability of object events in plugins

2 Object event modifications:

1) modules/admin/controllers/ObjectController.php: saveAction( checks for member variable 'saveActionExtraJSONdata' and appends it to the response JSON, then unsets it if it exists
2) static6/js/pimcore/object/object.js: save:function( ... now passes rdata[response data] to:
    fireEvent("postSaveObject"

And documentation.
## Additional info  

I needed a way to get more specific result information on the client side. 
e.g. request an object save with js[from browser], then look for certain values in the response.
Previously the response was pretty generic and un-editable.


kylepwarren:
<object>->saveActionExtraJSONdata  probably needs to be prefixed with something so it NEVER collides with an attr name. Is there a convention for that in PIM? like prefix with underscore_?

PR #1190, opened 2017-01-17
Change Objectbrick ClassId to ClassName

dpfaffenbauer:
This PR changes how ObjectBricks save their definition. Currently it looks like this:

```php
<?php 

return Pimcore\Model\Object\Objectbrick\Definition::__set_state(array(
    'classDefinitions' =>
        array(
            0 =>
                array(
                    'classname' => 3, //Here we have the ClassId instead of the ClassName
                    'fieldname' => 'objectBrickField',
                ),
        ),
    'key' => 'ObjectBrickTest',
    'parentClass' => '',
    'layoutDefinitions' => '' //LAYOUT SHOULD BE HERE, BUT HAS BEEN REMOVED FOR READABILITY
));
```

ClassIds are saved within the Database, so they depend to an environment, if we now deploy to a different environment where we have different Ids (which happens a lot), this object brick cant be saved, cause the ID belongs to a different or even non existing class.

The definition should look like this:

```php
<?php 

return Pimcore\Model\Object\Objectbrick\Definition::__set_state(array(
    'classDefinitions' =>
        array(
            0 =>
                array(
                    'classname' => 'AnyClassName', //Now we have the actual name
                    'fieldname' => 'objectBrickField',
                ),
        ),
    'key' => 'ObjectBrickTest',
    'parentClass' => '',
    'layoutDefinitions' => '' //LAYOUT SHOULD BE HERE, BUT HAS BEEN REMOVED FOR READABILITY
));
```

With the actual ClassName, Objectbricks are more easy deployable and detached to an environment.

## Extjs 3
This PR is also backward compatible to Extjs 3.


dpfaffenbauer:
Whats the state of this PR?

brusch:
We're quite busy with Pimcore 5 at the moment - we'll review & test your PR asap.

brusch:
Thanks for your PR. 

Thank you very much for your understanding :)

dpfaffenbauer:
:(

fine with me, it's a very small PR anyway ;)

PR #1199, opened 2017-01-19
Application logger: Fix for unsetting filter "Component"

podarcis:
Once the filter component has been set it couldn't be unset anymore (besides button "reset").


maff:
Thanks!

PR #1208, opened 2017-01-22
Documentation: workflow custom report and workflow GUI

Yivan:
This PR add some documentation to workflow to help users to exploit workflow more easily and all its power
: )

- A section like a tutorial for creation a custom report for workflows.
Having possibility to search elements in the workflow (for monitoring the progress for instance) is not a feature actually. Instead adding some new code, custom reports can beautifully handle workflow monitoring.
This new page is just a helper so user know it is possible to do it this way without any coding.

- A section mentionning the very nice plugin Workflow GUI. Has it has not been merged in the core, i think mentionning it here is a good value.

Finally, those two new pages make Workflow a step higher in feature/possibility for new Pimcore developper discovering Workflow for Pimcore.


fashxp:
thx very much.

PR #1211, opened 2017-01-23
Minor fix: staticroutes filter function uses only scalar values

xlttj:
This fix prevents PHP from warning 'strpos() expects parameter 1 to be string, array given' when filtering. Site ids are stored as an array in config/staticroutes.php since bac7b39


brusch:
Thanks!

PR #1221, opened 2017-01-25
Fixed potential issue with loading tags for an element

BlackbitNeueMedien:
A variable assignment was missing in the getTagsForElement function which can lead to an exception in the compare function.


brusch:
Thanks!

PR #1240, opened 2017-01-31
Fix handling of offset and limit

bbak:
Fix: Sending Newsletter through Report Adapter didn't work due to wrong handling of offset and limit.


brusch:
Thanks!

PR #1242, opened 2017-02-01
Set SQL SECURITY INVOKER on VIEWs

bbak:
This greatly simplifies moving Pimcore Databases around, without (in this case) reducing security, since there's no need anymore to keep the Users in sync - which is even impossible, when IP's/Hostnames change and requires to recreate all VIEWS after a database transfer (or patch the export SQL-file).

Drawback: This is a extension to the SQL Standard.
See [MySQL Docs](https://dev.mysql.com/doc/refman/5.7/en/stored-programs-security.html)

### DIFF
--- a/pimcore/models/Object/ClassDefinition/Dao.php
+++ b/pimcore/models/Object/ClassDefinition/Dao.php
@@ -206,7 +206,7 @@ class Dao extends Model\Dao\AbstractDao
         // create view
         try {
             //$this->db->query('CREATE OR REPLACE VIEW `' . $objectView . '` AS SELECT * FROM `objects` left JOIN `' . $objectTable . '` ON `objects`.`o_id` = `' . $objectTable . '`.`oo_id` WHERE `objects`.`o_classId` = ' . $this->model->getId() . ';');
-            $this->db->query('CREATE OR REPLACE VIEW `' . $objectView . '` AS SELECT * FROM `' . $objectTable . '` JOIN `objects` ON `objects`.`o_id` = `' . $objectTable . '`.`oo_id`;');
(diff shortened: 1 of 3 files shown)


Yivan:
Hello @bbak,
You mean with SQL SECURITY INVOKER, there is no more need to make a `grep -v "50013 DEFINER"` on the sql export before reimporting it in another MySQL environment ? (actually we do this between the different servers stages for deployment).
If so this is great!

bbak:
@Yivan I think so. Since this command tells MySQL that the restrictions of the user accessing the view (= INVOKER) should be applied, it shouldn't matter which user created the view in the first place. We've the same Issue with moving Pimcore Databases between several Stages.

brusch:
@bbak for me this produced the following when using mysqldump: 

```sql 
/*!50013 DEFINER=`root`@`%` SQL SECURITY INVOKER */
```

When importing with a different user i still get the same error: 
`ERROR 1227 (42000) at line 3217: Access denied; you need (at least one of) the SUPER privilege(s) for this operation`

So it seems that this doesn't fix the problem...

bbak:
*arks* Yes, damn. The DEFINER has to exists at least on the target RDBMS. Looks like the only option is to get rid of the DEFINER clause. [mysqldump 5.7 has a option to skip the definer clause](https://dev.mysql.com/doc/refman/5.7/en/mysqlpump.html#option_mysqlpump_skip-definer). I think this will solve our issue @Yivan 

The `SQL SECURITY INVOKER` clause would be beneficial only in the (rare) cases, where 2 different DB-Users would access the VIEWs.

Issue #1244, opened 2017-02-01
Improvement : Add filtering column on Notes & Events panel


Yivan:
In case there is a lot of notes and events for an element, it begin to be hard to find specifical information as there is only ordering but no filter for the grid.

I think it will be valuable to add filtering on the columns of Notes & Events of the element (document, asset, object) panel:
- Type : string filter
- Title: string filter
- Description: string filter
- User: string filter
- Date: date filter

![notes](https://cloud.githubusercontent.com/assets/9025839/22507529/41a8099c-e886-11e6-9f37-677156452e75.png)

Same could be added to the global "Notes & Events" panel (i don't think it is needed to add filtering on "Element" column as a user can go to the Notes & Events of the element to have only the one linked to it).

With this it would be possible to seach for specific type of notes, specific notes from a user, events wich happen between two dates, ... It would make usage more friendly for enduser.

weisswurstkanone:
build 60

PR #1245, opened 2017-02-01
notes user and date search

Yivan:
This pull request is for: https://github.com/pimcore/pimcore/issues/1244
As there is already a global filter for type, title and description, I just added support for this global search for user and date.

I think it is more simple than adding filter grid wich will be make things complicated because of the generic actual common filter.
User can search a specific user or specific date even a month (searching for "2017-01" for instance).


weisswurstkanone:
many thanks!

Issue #1238, opened 2017-01-31
Workflow: "elementPublished" => TRUE don't publish it.


Yivan:
On the last stable Pimcore 4.4.3, configuring a workflow status with `"elementPublished" => TRUE` don't publish the object when he got this status.
(all required fields are previously filled in previous status and have been save without publish).

By the way, i was just asking me what happen if not all required fields are filled when workflow is going to a status asking for publishing the object using `"elementPublished" => TRUE`. Are some warning triggered for the end user ?

Thanks.

Yivan:
As mentionned in the pull request https://github.com/pimcore/pimcore/pull/1246, there is always an issue regarding refreshing the tree after the element has been publishd/unpublished.
Thanks.

fashxp:
should be fixed now. 
for the issue with published/unpublished refresh please make an additional issue.

PR #1246, opened 2017-02-01
Doc workflow elementpublished

Yivan:
This pull request is for : https://github.com/pimcore/pimcore/issues/1238

The status was the previous one (oldStatus) in the comparaison check. This is corrected to take in account the status the element is going to be during the performing action.
Documentation has been updated to be more detailled on this.

I notice that after publishing, the object tree is not updated. As i don't know where it should be handled i didn't change this part, but i think the tree should refresh in case of publishing/unpublishing.
Should a specific issue be opened or can you correct it on the way ?

Thanks.


fashxp:
thx very much

Issue #1128, opened 2016-12-15
Keep usagelog even after sended to Pimcore Team.


Yivan:
Actually everytime the usagelog.log is sended to the Pimcore Team, the file is deleted.
The content of this file is great to retrieve which user make which action if needed.
I think it should be keeped zipped on the server and not being unlink like it is done actually. Please see:
https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Log/Maintenance.php#L140

Actually on a project we wanted to consult this log and it was just flushed and deleted some hours ago... as we wanted to get yesterday information, we lost them : (
The only way to prevent this seems to desactivate the submission for now.

Thanks!

Yivan:
Sorry the line is 104, not 140:
https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Log/Maintenance.php#L104

We can see here that the only way to keep usagelog is to disable usage sending (so `return` line 96 is executed) if we do not want to lost recent usage log (sending is based on the number of line, not on the time like the log cleaning on line 140).

A solution will be: change sending to time based criteria (30 days old for instance) wich involve file rotation (archive of this file). So at least we are sure at each time to be able to check usages 30 days back.
Another solution is simply not delete it (like if sending to pimcore checkbox was not cheked).

PR #1247, opened 2017-02-01
Archiving usagelog.log

Yivan:
This pull request is for: https://github.com/pimcore/pimcore/issues/1128

Instead deleting it directly, the usagelog  is archived for 30 days, so administrator don't loose datas the day the limit size was reached. It will be keep 30 days in case of.
And this way, it is more working like the other logs.


brusch:
Thanks!

Issue #1079, opened 2016-11-24
Improvement for document link (langage): unlink them


Yivan:
Linking document in different langage is a very great new feature, wich helped a lot the editors.
But in some case, we want to remove a link and actully only way is to duplicate document (and so loosing reference anywhere).

Maybe a "Unlink document" option in the menu could be add. It will show the different document linked (like "Open translation"). Choosing one will remove the link.
Here is a try of graphism presentation of this new option:

![link](https://cloud.githubusercontent.com/assets/9025839/20604568/55ee50fa-b267-11e6-9e0d-7526d5cedd55.png)

**Suggestion of documentation improvement on this subject:**
Aside of this, I didn't found in the documentation the way to request linked document using the API. Maybe it could be add. For instance it could be used when generating sitemaps multilingale.

As well, i think it is the case, but it could be add in the documentation that the hreflang meta are automaticaly added to (if zend head helper is injected in the layout template).

msenkpiel:
Yeah... that's a missing feature!!!

Yivan:
I make a bump on this issue as it prevent of using this feature (no possibility to unlink if link has made by error by an editor...).

A pull request was made but got only the Pimcore 4 code : https://github.com/pimcore/pimcore/pull/1248/files
Actually i am not able to provide the pimcore 5 code.

Maybe someone could take back the code of the pull request for Pimcore 4 and add the one for Pimcore 5 so pull request will be accepted.

Thanks.

Yivan:
I just reup this issue, as it lead us for now not using this feature because end user cannot revert his actions without modifing directly database, so it is too risky for now. Or has it been solved on Pimcore 5 (in this case it could be closed).

PR #1248, opened 2017-02-01
Action to unlink document translation

Yivan:
This pull request is for : https://github.com/pimcore/pimcore/issues/1079

It adds the missing unlink feature.

Documentation is updated with this feature (maybe you will want to update the Localization Tool image illustration, it is not in this PR).
I added some code to show how to get the linked translations for documents using API.

And I added a patch, because there was a bug in the current Pimcore release when two document was already linked and if you make the same link (from the other document), it was strangely added in documents_translations table with same id and sourceId. To prevent this i have added `if ($translation->getId() === $sourceId) return;` in the addTranslation method of the DAO.

I hope all is ok in this PR : )
Thanks!


brusch:
Thanks for your PR. 

Thank you very much for your understanding.

Issue #999, opened 2016-10-27
Object brick: Possibility to limit the number of bricks user can add.


Yivan:
This issue comes from #962 for reference.

Purpose is like field collection to be able to limit the number of object brick a user can add.
I think other value than 1 don't make sense (as we don't know in advance wich one will be selected ; ).

So just a simple checkbox could make the job on the class edition for the object brick type field:
"Limit to one brick": yes/no

PR #1249, opened 2017-02-02
Option to limit the number of objectbricks which could be added

Yivan:
This is pull request for : https://github.com/pimcore/pimcore/issues/999

Objectbricks can be limit to one, so only a single objectbrick could be added for the related field in an object.
Purpose of this is for instance, sometimes you model a generic object and want to add some specific fields depending the object type. Object brick is great for it but user could not be limited to the number of them he can add. This PR give the option to prevent the user to be able to add several objectbrick.

A use case could be modeling a generic "product" classe having an objectbrick field called "Type of product".
The generic classe will have price, name, weight, ...
The objectbricks could be: tv, bycicle, book... which will bring additionnal specific field to the product. Only one of those must be added to the product, no more.
Limiting the objectbrick field to one will permit this way of making product model.

### DIFF
--- a/pimcore/config/texts/en.json
+++ b/pimcore/config/texts/en.json
@@ -6279,5 +6279,13 @@
     {
         "term": "site_ids_tooltip",
         "definition": "Provide a comma-seperated list of site IDs"
+    },
+    {
+        "term": "limit_to_one_objectbrick",
+        "definition": "Limit to one objectbrick"
+    },
+    {
+        "term": "only_one_objectbrick_can_be_added",
+        "definition": "Only one objectbrick can be added"
     }
 ]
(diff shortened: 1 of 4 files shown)


fashxp:
shouldn't we make that a bit more generic and add the possibility to limit it to X entries?

Yivan:
I have thinking of it, but i don't see a lot of use case... so I had prefered to make it simple.
But you are right maybe some people could need this for specific use case, and making it number instead of bolean (0/1) is simple, code will be around the same and more generic.
So I will update the PR with a limiting number instead of checkbox.

Yivan:
@fashxp The pull request is updated, using maxItems (exactly like multiselect so code stay homogeneous).
Now user can choose to how many number of objectbricks he want to limit.
Thanks!

weisswurstkanone:
Many thanks!

Issue #1006, opened 2016-10-31
[Improvement] Improve document editable so they keep better the front end style for inline element


Yivan:
Using editable in document on the Pimcore edit tab, the editable area use "div" tag.
`<div id="pimcore_editable_headDescription" class="pimcore_editable pimcore_tag_input" style="min-height: 20px;" contenteditable="true">`

It prevent use it inside inline element (span, p, ...). If you do this:
`<p class="mycss"><?= $this->input("myInput) ?></p>`
The browser (Firefox last version Windows) will render:
```
<p class="myCssStyle"> (javascript goes here ) </p>
<div id="pimcore_editable_headDescription" class="pimcore_editable pimcore_tag_input" style="min-height: 20px;" contenteditable="true">... </div>
```
And we loose the css style applied on p.

In fact, the browser made automatical correction to the HTML, which is ok (a div cannot be on a inline element), but as Pimcore use DIV, sometimes editable loose the styles.

Solution:
1- 
Maybe it could be changed to (changing to span and add display block style):
`<span id="pimcore_editable_headDescription" class="pimcore_editable pimcore_tag_input" style="min-height: 20px; display: block;" contenteditable="true">`

Same could be done for some other editable (textarea, if used inside a p), etc.

2-
Or another solution is to provide a parameter to the editable to set div or span (but i think 1st solution is nicer because it is automatic).

Yivan:
@brusch 
Can i make a pull request on this based on the first solution : changing generated `<div...` to a `<span style=display: block;...` ?

brusch:
@Yivan I guess this change needs a lot of testing, I wouldn't just change it, because this might break a lot of existing installations.

Yivan:
Thanks for your return.
Ok I understand perfectly.

Is there an existing hook/event we could use to change the DIV in SPAN using a backend plugin ? It seems tricky.

Another more proper way is to provide a configuration parameter to define the surrounding tag (by default "div"). So developper on case of P could configuring the surrounding to SPAN tag.
Exemple:
<p class="mycss"><?= $this->input("myInput, array("surroundingTag" => "span")) ?></p>

It will keep backward compatibily and leave the possibility to use inline element.
If you find this solution good, i can make a PR for input tag. Do you think i make it for others editable (i think of: textarea, wysiwyg and image, but maybe others) ?

solverat:
[...]

PR #1250, opened 2017-02-02
Document editable tag name option

Yivan:
This is a pull request for : https://github.com/pimcore/pimcore/issues/1006

It don't break anything as it is implemnted as an option (defaulted to "div" tag for sourrounding container).
Documentation has been updated (it was the long part of this PR ; ) )

With this, it is now possible to do (if forced the style inline for this exemple, but better in a css file for sure):
`<p style="background-color: red; font-size: 50px;"><?= $this->input("myinput", ["tag" => "span"])?></p>`
And the style in editmode won't be broken anymore and user can see as it will render in preview/real mode.

### DIFF
--- a/docs/Development_Documentation/03_Documents/01_Editables/08_Checkbox.md
+++ b/docs/Development_Documentation/03_Documents/01_Editables/08_Checkbox.md
@@ -2,11 +2,12 @@
 
 ## Configuration
 
-| Name     | Type    | Description                                                                        |
-|----------|---------|------------------------------------------------------------------------------------|
-| `reload` | boolean | Set to true to reload the page in editmode after changing the state.               |
-| `label`  | string  | a `<label>` which is added in the editmode                                         |
-| `class`  | string  | A CSS class that is added to the surrounding container of this element in editmode |
+| Name     | Type    | Description                                                                                                    |
+|----------|---------|----------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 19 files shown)


brusch:
There seems to be a problem with the line breaks in alle the *.md files.
I'd also prefer a more meaningful name of the configuration option, `tag` seems to be too generic to me and maybe causes confusion. What about `htmlTag`?

Yivan:
@brush Thanks for you return.

I totally agree, sure I will change the option name  to htmlTag in the code and in the documentation.

For documentation do you prefer for the description I left :
"A tag name that is used instead of the default `div` for the surrounding container of this element in editmode"
Or i change it for:
"An HTML tag name that is used instead of the default `div` for the surrounding container of this element in editmode"
? Or some other sentence ?

As I am goinf to modify all documentation files it is the time to say me if i must change this ; )

---

For the linebreak problem i didn't notice one (i checked 08_Checkbox.md). It seems fine (no warning on IDE - Netbeans - nor Notepad++ which usualy warm a message when there is some linebreak mixing (linux/windows...). It tell it is linux linebreak and all seems ok.
I join a compressed file of my local modified documentation (the one in the pull request):
[01_Editables.zip](https://github.com/pimcore/pimcore/files/771142/01_Editables.zip)
Could you please tell me a little more about this problem so i can correct it please ?

Thanks!

brusch:
Thanks again for your PR. 

Thank you very much for your understanding.

Issue #1086, opened 2016-11-28
New feature: add tags to REST API


Yivan:
In recent Pimcore some methods was added to handle more easily tags
(for instance https://github.com/pimcore/pimcore/pull/868)

A new feature could be adding the possibility to request Pimcore webservice using tags.

This could be:

```

Get List of Tags

    Method: GET
    URL: http://YOUR-DOMAIN/webservice/rest/tags?apikey=[API-KEY]
    Returns: The JSON-encoded list of tags with their ID
```

```
Get Element By TAG

    Method: GET or POST
    URL for GET: http://YOUR-DOMAIN/webservice/rest/elements-list-tag?apikey=[API-KEY]&tags=21,50,87
        ids Parameter for GET: comma-seperated list of tags ids
    URL for POST: http://YOUR-DOMAIN/webservice/rest/element-tag?apikey=API-KEY
        Request body for POST: comma-seperated list of tag ids
    Returns: JSON-encoded success value and list of element ids and type of the element.
```

Best regards.

fashxp:
good idea, pull requests are welcome ;-)

Yivan:
I thinks this issue can be closed as resolved (see https://github.com/pimcore/pimcore/pull/1253).

PR #1253, opened 2017-02-03
Add webservices for tags

Yivan:
This is a pull request for: https://github.com/pimcore/pimcore/issues/1086

I added three services:
- tags: list of tags
- tags for element: list of tags for an element
- elements for tag: list of elements for a tag

### DIFF
--- a/docs/Development_Documentation/14_Web_Services/README.md
+++ b/docs/Development_Documentation/14_Web_Services/README.md
@@ -34,12 +34,22 @@ Unrestricted access is only granted to admin users. For all other users the foll
 * Asset permission:
     * Asset List
     * Asset Count
+    * Tags for Element (element of type Asset)
+    * Elements for Tag (element of type Asset)
 * Document permission
     * Document List
     * Document Count
+    * Tags for Element (element of type Document)
+    * Elements for Tag (element of type Document)
 * Object permission
     * Object List
     * Object Count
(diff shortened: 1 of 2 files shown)


weisswurstkanone:
Can we keep the current naming convention, e.g. assetListAction, objectListAction, ... ?

Yivan:
I made naming like for classes (classesAction) which list classes (so i do tagsAction naming). I thought "list" naming was only for element: document/asset/object.
If we name it listTagAction, maybe classes list action should be rename (but will break backward compatibility).

What do you prefer : you merge and you rename it like you prefer or i do the renaming ?
If i do the renaming i change tagsAction to tagsListAction and what about the 2 others actions naming, is is ok (or shoud i go for elementsTagList and tagsElementList) ?

Thanks.

weisswurstkanone:
Can you do the renaming. After that it should be fine I guess :-)

Yivan:
Sure, I will be pleased to do it.

Just to be certain:
- tagsAction -> tagListAction
- elementsTagAction -> elementsTagListAction
- tagElementsAction -> tagElementsListAction
- I keep classesAction like this

Is it ok ?

weisswurstkanone:
Should be ok

Yivan:
@weisswurstkanone Ok this is done!

weisswurstkanone:
thx!

PR #1254, opened 2017-02-03
Type annotations should not have brackets

bbak:
Issues Warnings in PHP Language Server: https://github.com/felixfbecker/php-language-server


brusch:
thx

PR #1262, opened 2017-02-08
getStyles causes a javascript error

lukas-schnieper:
getStyles is not a function. i replaced it with getStyle which is working.


weisswurstkanone:
many thanks!

PR #1267, opened 2017-02-09
phpdoc of magic method mustn't contain static keyword

fheyer:
phpdoc for magic Listing methods is generated with static keyword
However the phpdoc standard doesn't allow this.

Syntax for @method from phpdoc:
@method [return type] [name]([[type] [parameter]<, ...>]) [<description>]

reference:
https://phpdoc.org/docs/latest/references/phpdoc/tags/method.html

I experienced problems in NetBeans as magic methods generated by Pimcore like
`* @method static \Pimcore\Model\Object\Account\Listing getByAccount_type ($value, $limit = 0)`
annoy the code indexer. The indexer takes `static` as type and `Listing` as method name which appears as many times as we have magic methods. Thus it displays "duplicate method" errors.


brusch:
Thanks!

PR #1269, opened 2017-02-10
when automatically creating a redirect, check if the document is in a…

go4webch:
… site and add the site id to the redirect.

Because, when pimcore creates a redirect for a document within a site, the redirect needs to have the site id otherwise it won't work.


go4webch:
I signed the CLA agreement and added my mail to our github user, does anyone know if the CLAassistant will re-check automatically or do I need to open another PR?

Thanks, Andy

brusch:
No there is no recheck, i guess you have to open the PR again :(

PR #1272, opened 2017-02-10
when automatically creating a redirect, check if the document is in a…

go4webch:
… site and add the site id to the redirect.

Because, when pimcore creates a redirect for a document within a site, the redirect needs to have the site id otherwise it won't work. (Reopen)


brusch:
Thanks!

PR #1273, opened 2017-02-11
Notice error ApplicationLoggerDb

Yivan:
When making a simple log send:

```
$logger = \Pimcore\Log\ApplicationLogger::getInstance("My component", true);
$logger->log("info", "Test message");
```

A notice error is send (undefined index).

This pull request prevent this notice error when using ApplicationLoggerDb.

Thanks.

### DIFF
--- a/pimcore/lib/Pimcore/Log/Handler/ApplicationLoggerDb.php
+++ b/pimcore/lib/Pimcore/Log/Handler/ApplicationLoggerDb.php
@@ -62,7 +62,7 @@ class ApplicationLoggerDb extends AbstractProcessingHandler
             'priority' => strtolower($record["level_name"]),
             'message' => $record["message"],
             'timestamp' => $record["datetime"]->format("Y-m-d H:i:s"),
-            'fileobject' => isset($record["context"]["fileObject"]) ? $record["context"]["fileObject"] : null,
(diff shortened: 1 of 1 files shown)


pimcore/lib/Pimcore/Log/Handler/ApplicationLoggerDb.php
```diff
@@ -62,9 +62,9 @@ public function write(array $record)
             'priority' => strtolower($record["level_name"]),
             'message' => $record["message"],
             'timestamp' => $record["datetime"]->format("Y-m-d H:i:s"),
-            'fileobject' => $record["context"]["fileObject"],
-            'relatedobject' => $record["context"]["relatedObject"],
-            'relatedobjecttype' => $record["context"]["relatedObjectType"],
+            'fileobject' => isset($record["context"]["fileObject"]) ? $record["context"]["fileObject"] : null,
```
manea-eugen: Since the isset checks also if the values is null (which in this context its a perfectly working statement) should we maybe replace "isset($record["context"]["fileObject"])" with "array_key_exists('context', $record) && array_key_exists('fileObject', $record['context'])" just to make it more future-proof, just in case php decides to interpret the isset as falsy instead of null

brusch:
@Yivan Yes, @manea-eugen is right, please change your PR accordingly. 
Thanks!

Yivan:
Thanks manea-eugen, you are totally right yes.
@brush modification is made!

brusch:
Thanks

PR #1274, opened 2017-02-12
Unused variable $sql

manea-eugen:
I think is better to use the sql variable, since we apply trim on $m and we append ';'


brusch:
Thanks!

Issue #1279, opened 2017-02-13
Localized fields at a glance (side-by-side)


msenkpiel:
Add a new option to localizedfields data object definition. So you can view mutliple language fields at a glance side by side. That's helpful to translate text.

brusch:
A generic approach would be to let the user choose the languages and save it somewhere in the local storage.

frithjof:
An overlay as long as you are in the field showing the other language values could be nice

brusch:
Duplicate of #3054

PR #1280, opened 2017-02-13
Localizedfields at a glance

msenkpiel:
# Feature Request 

Issue: https://github.com/pimcore/pimcore/issues/1279

## Feature description 

Added a new option to localizedfields data object definition. Now you can view mutliple language fields at a glance side by side. That's helpful to translate text. 
The Panel will scroll horizontially if the screen smaller than needed space for localized fields. 

feat-glance-class
feat-glance-object


brusch:
I'm afraid that this feature isn't thought quite to the end. 
I completely understand the use-case, but this doesn't make sense for more than 2 languages at the same time - so it's not really scalable. A more generic approach would be to let the user choose the languages and save it somewhere in the local storage. But I think we should make a sophisticated concept for this before we implement anything. 

Thanks for your understanding.

PR #1286, opened 2017-02-14
Added default empty arrays to new variables in UserRole.php

CptnChoas:
Added [] to new variables to prevent an error in pimcore/models/User.php getMergedWebsiteTranslationLanguagesEdit() function.
getWebsiteTranslationLanguagesEdit() will return null without empty array initialization, which will cause:
$this->mergedWebsiteTranslationLanguagesEdit = array_values($this->mergedWebsiteTranslationLanguagesEdit);
to crash.


brusch:
Thanks!

PR #1289, opened 2017-02-14
Add missing PHP Docs Everywhere

dpfaffenbauer:
This PR adds missing PHP Docs everywhere. I think this makes sense for the upcoming pimcore 5 release.

Some other things I encountered doing these PHP Docs:

 - some setter functions return $this, some don't, shouldn't this be consistent?
 - a lot of function return void instead of an expected type. Shouldn't they return null instead?
 - \Pimcore\Tool\Less contains some undefined variables
 - some methods have parameters which aren't used anymore, shouldn't this be cleared?
 - Pimcore uses a lot of \Exception, shouldn't be there Pimcore\Exception which inherits \Exception?
 - a lot @return Statements are wrong, specially in namespace \Pimcore\Model\ClassDefinition\Data 
 - a lot of methods @return null without return null statement.

I also added some //TODO: Messages to find issues again.

I know it's not the right place for discussion, this should only be a hint ;)


kanevbg:
That is great

pimcore/lib/Pimcore/WorkflowManagement/Workflow/Manager.php
```diff
@@ -678,7 +703,7 @@ private function unregisterActionEvents()
 
     /**
      * Returns whether or not an element has a workflow
-     * @param Asset|ConcreteObject|Document $element
+     * @param AbstractElement $element
```
maff: Please keep the previous (more specific) type hint here.

pimcore/models/Asset/Document.php
```diff
@@ -53,6 +56,10 @@ protected function update()
         parent::update();
     }
 
+    /**
+     * @param null $path
+     * @return null
```
maff: This should be ìnt|null` (also applies to `getPageCount`)

pimcore/models/Asset/Listing.php
```diff
@@ -86,17 +85,24 @@ public function setAssets($assets)
...
     }
 
-
     /**
      *
      * Methods for \Zend_Paginator_Adapter_Interface
      */
 
+    /**
+     * @return mixed
```
maff: `int`

pimcore/models/Asset/Listing.php
```diff
@@ -86,17 +85,24 @@ public function setAssets($assets)
...
+     */
     public function count()
     {
         return $this->getTotalCount();
     }
 
+    /**
+     * @param int $offset
+     * @param int $itemCountPerPage
+     * @return mixed
```
maff: This should always return an array of assets.

pimcore/models/Object/ClassDefinition/Data/Classificationstore.php
```diff
@@ -1004,12 +1004,17 @@ public function setLocalized($localized)
...
+    /**
+     * @param $object
+     * @param array $mergedMapping
+     * @return array|boolean
+     */
     public function recursiveGetActiveGroupCollectionMapping($object, $mergedMapping = [])
     {
         if (!$object) {
-            return;
+            return false;
```
maff: Please do not alter the return type - previously it returned `null`, now `false`.

pimcore/models/Object/ClassDefinition/Data/Classificationstore.php
```diff
@@ -1038,14 +1043,14 @@ public function recursiveGetActiveGroupCollectionMapping($object, $mergedMapping
...
+     * @param $object Object\Concrete
      * @param array $activeGroups
-     * @return array
+     * @return array|boolean
      */
     public function recursiveGetActiveGroupsIds($object, $activeGroups = [])
     {
         if (!$object) {
-            return;
+            return false;
```
maff: Same as above.

pimcore/models/Object/ClassDefinition/Data/Classificationstore.php
```diff
@@ -1186,6 +1191,8 @@ public function getAllowedGroupIds()
      */
     public function setAllowedGroupIds($allowedGroupIds)
     {
+        $parts = [];
```
maff: Not sure which implications this has as groups were set to null before, now to an empty array. Please add a `TODO` to check this, but do not alter the behaviour.

pimcore/models/Object/ClassDefinition/Data/Multihref.php
```diff
@@ -394,6 +400,8 @@ public function getDataForGrid($data, $object = null, $params = [])
     public function getVersionPreview($data, $object = null, $params = [])
     {
         if (is_array($data) && count($data) > 0) {
+            $pathes = [];
```
maff: Same as before (although here probably it doesn't change behaviour). Nevertheless I'd leave this as it was and let PHP trigger a warning - make it easier to find. A TODO couldn't hurt though :)

pimcore/models/Object/Classificationstore/GroupConfig.php
```diff
@@ -148,14 +153,11 @@ public function setParentId($parentId)
...
     /**
-     * @param string name
-     * @return void
+     * @param string $name
      */
     public function setName($name)
     {
         $this->name = $name;
-
-        return $this;
```
maff: Please do not alter the method behaviour.

pimcore/models/Object/Classificationstore/GroupConfig/Dao.php
```diff
@@ -70,8 +70,13 @@ public function getByName($name = null)
         }
     }
 
+    /**
+     * @return int
+     */
     public function hasChildren()
     {
+        $amount = 0;
```
maff: Although this should be OK, please do not add something like this in this commit, as it potentially changes behaviour (calls relying on hasChildren returning a `null` or an int). Rather add a TODO, so we can fix this in a later change.

pimcore/models/Tool/Email/Blacklist/Listing/Dao.php
```diff
@@ -45,8 +45,13 @@ public function load()
         return $addresses;
     }
 
+    /**
+     * @return int
+     */
     public function getTotalCount()
     {
+        $amount = 0;
```
maff: Same as before - please do not alter behaviour in this huge commit, rather add a TODO.

pimcore/models/WebsiteSetting/Listing/Dao.php
```diff
@@ -42,9 +42,13 @@ public function load()
         return $settings;
     }
 
-
+    /**
+     * @return int
+     */
     public function getTotalCount()
     {
+        $amount = 0;
```
maff: As before.

pimcore/lib/Pimcore/WorkflowManagement/Workflow/Manager.php
dpfaffenbauer: Can't agree with you on this, since the parameter is declared as AbstractElement $element. So Doctype only allows one of these Asset|ConcreteObject|Document, but method takes AbstractElement anyway.

Your suggestion has a good point, but not actually right, don't you think so?

maff: More specific type hints are possible to get better IDE autocompletion when the actual type is too generic, so something like `AbstractElement|Asset|ConcreteObject|Document` would be OK in my mind.

As this isn't really of use here, I'm fine with changing it to `AbstractElement` only.

maff:
Merged manually in d8d539516d20fc0150b005afbfa7834796c729b3. I reverted a couple of changes to make sure we don't run into too many merge conflicts with this PR (especially cache components and shifting around methods) and squashed your commits so only those files really changed will have this commit in its logs. As reference what I reverted, it's still on the branch [php-doc-fixes-prepare](https://github.com/pimcore/pimcore/compare/php-doc-fixes-prepare) in commit bcf72ca7662795c15c6cc2ab6bfdb376cdc5fe07.

Thanks a lot for the effort of adding all those docblocks :) A little remark: for the future we should try to avoid changing so much at once and better change things gradually (e.g. component based) to make it easier to review changes and to find regressions.

dpfaffenbauer:
I totally understand your worries, I just recognized a lot of small doc block PR's and thought its time to rule them all. Of course its very hard to review such big PR's, and I normally I wouldn't create such a big one, but I thought it would make sense in this case.

Thanks for merging :)

Cheers

PR #1293, opened 2017-02-15
Add less map

Yivan:
This pull request add less source map (to Less_Parser PHP processing only, not lessc binary as i don't use/know it and prefer to left this part to people being able to test it).
Using it make debuging a lot easier, so it will help developper life if this is merged.

A problem with Firefox dev tools using the sources map exists for domain with hyphen. For reference, the issue has been submitted to Firefox issue tracker:
https://bugzilla.mozilla.org/show_bug.cgi?id=1311424
So in case of test/using sources map better to go on Google Chrome for now ; )

Thanks.

### DIFF
--- a/pimcore/lib/Pimcore/Tool/Less.php
+++ b/pimcore/lib/Pimcore/Tool/Less.php
@@ -159,7 +159,15 @@ class Less
         // use php implementation of lessc if it doesn't work
         if (empty($compiledContent)) {
             $parser = new \Less_Parser();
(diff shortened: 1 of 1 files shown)


brusch:
Less support will be removed in v5, so I don't think we should add any more features to it.

brusch:
Thanks anyway :)

Yivan:
I will try Gulp or other tools, it's time for me to evolve ; )
Thanks.

PR #1297, opened 2017-02-17
fixed on date filter for creation and modification date in shared translations

daves-weblab:
The "on" filter for creation and modification date was not working for shared translations due to comparison of timestamps instead of just dates.


weisswurstkanone:
thx!

PR #1298, opened 2017-02-17
object sort brick col removed unknown column bug fix

kylepwarren:

Fixes # 

Found object grid bug when displaying brick columns: if grid is sorted by brick column, then column is removed from grid configuration, grid gives 'unknown column error', because the sort setting is remembered but the [brick]column is no longer in the sql rows/results.

catch that looks for specific error message, clears sort param, and reties gridProxyAction *once - infinite retries prevented, only one retry attempt is made.


brusch:
Hmm, i don't think we should just suppress the error message because then the user expects to get an ordered result, which isn't the case. So I think we should implement a proper info message when trying to sort a non-sortable field.

PR #1299, opened 2017-02-17
added filter for allowed predefined document types

dominikschnieper:
Allowes developers to define whitelist for predefined document types in perspectives.

## Additional info  
use "allowed_predefined_page" => ["My Page", "Special Page"] or "allowed_predefined_email" and so on in perspectives.php inside the customview.


brusch:
Actually we'd prefer a config where the document type isn't part of the config key. 
We also think that we should implement restrictions across the entire system using the same config, which makes it easier to understand and use. 
If you'd like to see this feature in the upcoming version 5, please create an issue for us. 
Thanks again!

PR #1301, opened 2017-02-17
Add windows support for localisation and Carbon

Yivan:
This pull request add localisation support for Windows.
On windows ".utf8" is not supported, so PHP localisation fall in "C", preventing using date or other function depending of it.

It add too the configuration for carbon so something like that will be ready to work natively on the same language than document:
`echo \Carbon\Carbon::now()->subDays(5)->diffForHumans();`

### DIFF
--- a/pimcore/lib/Pimcore/Controller/Action/Frontend.php
+++ b/pimcore/lib/Pimcore/Controller/Action/Frontend.php
@@ -344,10 +344,17 @@ abstract class Frontend extends Action
                 }
             }
 
+            // windows system doesn't support utf8, we add same list without utf8
+            $localeList = array_merge($localeList, str_replace(".utf8", "", $localeList));
+            
             // currently we have to exclude LC_MONETARY from being set, because of issues in combination with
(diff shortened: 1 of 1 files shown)


brusch:
calling the following code automatically can be quite dangerous
```php 
\Carbon\Carbon::setLocale($locale->getLanguage());
\Carbon\CarbonInterval::setLocale($locale->getLanguage());
```

See: 
https://github.com/pimcore/pimcore/commit/3e67d187c85a619374bce33711987d10cc206d26

Yivan:
@brusch I undestand for Carbon, thanks.
What about the other part of this PR (windows locale fix) ? Can i resubmit a PR with just the part for the locallist on windows ? Actually strftime, ... php functions based on locale settings don't work on windows. By setting document en, fr, de or anything else bring "C" locale without thix fix.

brusch:
@Yivan Well, I'm not sure if Pimcore really should automatically set the locale using the wrong encoding (1052 instead of utf8). Imho I think this shouldn't be done by default.

PR #1313, opened 2017-02-23
ClassificationStore: Also get inheritanced values when calling ->getI…

ultramegatom:
ClassificationStore: Also get inheritanced values when calling ->getItems().

@weisswurstkanone please check this pull-request


weisswurstkanone:
Is the result the same when compared to the edit mode? 
Please double-check the lookup mechanism described here
https://www.pimcore.org/docs/latest/Objects/Object_Classes/Data_Types/Classification_Store.html

ultramegatom:
Result should be the same.

The language fallback doesen't really apply in getItems does it? We get the language in the array if the language exists, otherwise its simply not there? Or am i misinterpreting something?

weisswurstkanone:
obsolete

PR #1314, opened 2017-02-24
Added note about Google services for preview

### LINKED ISSUE #1306: Asset document preview tab uses Google service - possible privacy issue?
# Feature Request 
## Please make sure all the following boxes are checked before submitting your feature request - thank you!
- [x] There is no existing issue regarding the same topic

## Feature description 

The asset preview tab in the backend for document uses a Google service. (Url: `https://docs.google.com/viewer?embedded=true&url=` )
As this makes data pass through a 3rd party service of a company whose privacy terms might not be compatible with the providers terms, i think there should be a way to disable this "Feature".
I agree that this is an edge case, as the assets are publicly accessible anyway - still it is non-obvious for users of the system. 
  
[https://github.com/pimcore/pimcore/blob/master/pimcore/modules/admin/views/scripts/asset/get-preview-document.php#L54](https://github.com/pimcore/pimcore/blob/master/pimcore/modules/admin/views/scripts/asset/get-preview-document.php#L54)

If you agree, we should discuss the best way how to handle this issue. A switch in the system config? Just hiding/disabling the preview tab?
I would be happy to work on this and provide a pull request once the best solution is determined.

Chris

christophluehr:
Closes #1306 

Added note about google services.

### DIFF
--- a/docs/Development_Documentation/04_Assets/README.md
+++ b/docs/Development_Documentation/04_Assets/README.md
@@ -4,6 +4,9 @@ Assets are files that can be managed within the Pimcore system which you can org
 are images. Other kinds of common assets are PDF or MS Word documents which people can download from the website.
 Pimcore is able to render preview images for most file types. 
 
+> Please note: The asset preview tab for documents uses Google services if the Browser has no PDF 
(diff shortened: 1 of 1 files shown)


brusch:
Thanks!

PR #1321, opened 2017-03-01
Added optional tags to view cache

alesak:
Added optional tags to view cache to be able to chain cache with object or asset changes


brusch:
I think it would be better to add it to the initial call of the view helper, eg: 
`$this->cache("foo", null, false, ["tag1","tag2"])` 
Would be more consistent in my opinion.

PR #1327, opened 2017-03-04
adding support for plugin-javascript on login page

dominikschnieper:
Added support for plugins to include javascript files on the login page, if they are starting with "login.".


dominikschnieper:
I just found out how to replace the whole login page, so this is not really needed anymore.

PR #1328, opened 2017-03-04
conditionally show a custom layout depending on object values

dominikschnieper:
Conditionally show a custom layout of an object depending on a value. For instance show different default view of unpublished objects.

Steps to implement:
Create a parent class for your object and implement a method like this:

public function isDefaultConditionalLayout(ClassDefinition\CustomLayout $conditionalLayout) {
    if ($conditionalLayout->getName() == "inactive-user") {
        return !$this->getPublished();
    }
    return false;
}


brusch:
I think this should be done by an event to be consistent. 
Please create an issue for us, we're happy to look for a consistent solution for this.

PR #1333, opened 2017-03-07
export group in class definition

manea-eugen:
Allow exporting of group in class definition export


weisswurstkanone:
many thanks!

PR #1336, opened 2017-03-08
classes-rebuild command: Added option for rebuilding only modified classes

podarcis:
If the option --modified-only (or -m) is given, classes-rebuild will only rebuild classes where the generation date of the definition file and the php class file differs. If the generation dates are identical, the rebuild for this class is skipped entirely (no update of definition, php class file or database).

This feature improves working collaboratively on projects under SCM. Before deployment:classes-rebuild was giving the developer a hard time figuring out whether the file changes were only differences resulting from the rebuild process.


podarcis:
This PR seems to be now obsolete with PR #1339 merged (might be the better solution).

weisswurstkanone:
Let's close it for now :-) Many thanks!

PR #1337, opened 2017-03-09
static create asset fix (remote stream)

d-gerken:
If you try to static create an asset from a non-local stream it will just create an empty file. That's because the $tmpFile is unlinked (line 326) right after stream_copy_to_stream in line 321. Later in the getStream method of the Asset class $this->stream is set to null, because a non-local stream is (always) not rewindable.

Fixes # 

1) getStream just sets the stream to null if a LOCAL stream is not rewindable. So the remote stream can be copied to local filesystem in the update method in line 624.

2) Line 307 - 333 is unnecessary, because all these I/O Actions like line 311 are performed twice in the setData method.

- Removed line 307 - 333
- Don't set a non-local stream to null if it's not rewindable
- Changed some array_key_exists calls to isset

### Steps to reproduce:
`$imgUrl = "https://www.arvato.com/content/dam/arvato/images/campaign-material/logimat/Arvato_IT_LogiMAT2017_2880x1800.adaptive.400.jpg";`
`$filename = array_pop(explode("/", $imgUrl));`
`$fp = fopen($imgUrl, 'r');`
`@\Pimcore\Model\Asset\Image::create(1, array('stream' => $fp, 'filename' => $filename));`

### DIFF
--- a/pimcore/models/Asset.php
+++ b/pimcore/models/Asset.php
@@ -1181,8 +1181,8 @@ class Asset extends Element\AbstractElement
      */
     public function getStream()
     {
-        $streamMeta = stream_get_meta_data($this->stream);
         if ($this->stream) {
+            $streamMeta = stream_get_meta_data($this->stream);
             if (!@rewind($this->stream) && $streamMeta['stream_type'] === 'STDIO') {
                 $this->stream = null;
             }
--- a/pimcore/models/Asset.php
+++ b/pimcore/models/Asset.php
@@ -303,7 +303,7 @@ class Asset extends Element\AbstractElement
         // create already the real class for the asset type, this is especially for images, because a system-thumbnail
         // (tree) is generated immediately after creating an image
         $class = "Asset";
-        if (array_key_exists("filename", $data) && array_key_exists("sourcePath", $data) ) {
+        if (isset($data["filename"]) && isset($data["sourcePath"]) ) {
             $mimeType = Mime::detect($data["sourcePath"], $data["filename"]);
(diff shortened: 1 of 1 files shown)


brusch:
You have removed a lot of functionality in this PR. The options `data` and `stream` are totally ignored.

d-gerken:
Sorry, but I don't agree with you. All code I have delete was nonsense. 

Take a closer look at this. In the original code you're checking if it's a stream or data, then create a temporary file, put stream content in there and one line later delete this file. That makes no sense.

All magic happens in this line in the static function create.
$asset->setValues($data); 

If $data['stream'] is set, the method   public function setStream($stream) will be executed and the stream is set. 
If $data['data'] is set, the method public function setData($data) will be executed. 

Apart from this the current code does NOT work with remote streams. Please try my "Steps to reproduce". 

If you have further question do not hesitate to contact me.

brusch:
I'm sorry, you are absolutely right - thank you very much. :)

d-gerken:
You're welcome :)

brusch:
Ok, my first intention was right, it does break backward compatibility, example: 

Before the patch: 
```php 
$image = Asset::create(123, ["data" => $binData]); 
get_class($image) === "Asset\Image"
```

After your patch: 
```php 
$image = Asset::create(123, ["data" => $binData]); 
get_class($image) === "Asset"
```

-> Revered again

d-gerken:
You're right - sorry for that. I'll add the required mimetype detection for correct class creation soon.

PR #1339, opened 2017-03-10
Fixed unnecessary class definition file rebuild during class rebuild command

kkarkus:
During the `php pimcore/cli/console.php deployment:classes-rebuild` always are generated new timestamps in `definition_*` files, in my opinion it a bug, because it first load definition file saves it(why?) and then create/update models and database.


weisswurstkanone:
The only concern I have is that the timestamp in the definition file will not match anymore the one in the generated PHP file. 

/** 
* Generated at: 2017-03-13T09:19:31+01:00
* Inheritance: yes
* Variants: yes
* Changed by: admin (3)

kkarkus:
Well, I understand your point however there wasn't situation when I had to check whether timestamps are correct.

Regenerating timestamps during classes rebuild is a bit annoying because you need to checkout `definition_*` files.

weisswurstkanone:
build 4057

kkarkus:
thank you for merging :+1:

PR #1342, opened 2017-03-14
Quantity value grid display

IgorBenko:
The quantity value fields were displayed like this in the grid: 123 [object Object]
This pull request fixes the issue by implementing the getDataForGrid method.


weisswurstkanone:
many thanks!

build 4060

PR #1349, opened 2017-03-16
Pass the colorspace option to the print document

IgorBenko:
We need this option accessible in our print document for further processing.


brusch:
Thanks!

PR #1360, opened 2017-03-23
Add info about manipulating website settings in PHP

Cruiser13:
The old way using config files ( like described here https://www.pimcore.org/board/viewtopic.php?t=936 ) is no longer valid but still very present in google so I added the current way to the docs.


weisswurstkanone:
thx!

PR #1364, opened 2017-03-29
XLIFF export: Handle selfclosing tags

YKWeyer:

Extra-check on the tag name to handle self-closing tags (do not use `<bpt>`/`<ept>` logic but the `<ph>` instead).

## Additional info  
About the `<ph>` element :
> The <ph> element is used to delimit a sequence of native standalone codes in the translation unit
> http://www.maxprograms.com/articles/xliff.html#Table2

This case happens on WYSIWYG inputs with the `enterMode` parameter to 2.

The modifications handle the following cases:
- `<br>` (HTML 5)
- `<br />` (XHTML syntax, valid in HTML5 as well)
- `<br></br>` (XHTML syntax)


brusch:
Thanks

PR #1367, opened 2017-03-30
Wait 1 sec after gziping

jstsch:
The current maintenance script will wait 5 seconds each iteration regardless of the need for version gzipping. In this case we'll just wait 1 sec after gzipping to reduce total load. Currently the script always sleeps about 50 seconds on many of our sites even if nothing has to be done.


brusch:
Thanks!

Issue #1371, opened 2017-03-30
preUpdate and postUpdate events for properties


sebwebdev:
# Feature Request 

Use case: I require additional preUpdate and postUpdate events for properties in order to hook in and add some validation for the values.

Pull request: [1372](https://github.com/pimcore/pimcore/pull/1372)

PR #1372, opened 2017-03-30
property.preUpdate and property.postUpdate events

sebwebdev:
Use case: I require additional preUpdate and postUpdate events for properties in order to hook in and add some validation for the values.

Issue: [1371](https://github.com/pimcore/pimcore/issues/1371)

### DIFF
--- a/pimcore/models/Property/Dao.php
+++ b/pimcore/models/Property/Dao.php
@@ -71,6 +71,8 @@ class Dao extends Model\Dao\AbstractDao
             "data" => $data
         ];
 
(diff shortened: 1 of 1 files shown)


brusch:
Isn't it better to use the events of asset, object and document?

brusch:
If you have loads of properties this seems quite a big overhead to me. 
Nevertheless, we're not accepting any feature requests for version 4 anymore, since we're publishing the first preview version of 5 within the next couple of hours.

sebwebdev:
I can indeed use the object events and iterate over the properties - which is some overhead in my case. Thus the request. But I can live with this as well.

PR #1411, opened 2017-04-06
Added optional tags to view cache to be able to link cache with objec…

alesak:
Added optional tags to view cache to be able to chain cache with object or asset changes

I have kept option for adding tags in end and stop methods as it is very useful (usually tags are known after the code enclosed in cache, eg: going through list of object)


brusch:
Sorry, we're not accepting any further feature PR's for v4. 
Your PR is also unfortunately for the wrong file (it's in legacy bundle of v5) and the documentation is missing as well.

PR #1434, opened 2017-04-11
Site handling performance improvement

markus-moser:
I installed a fresh copy of the pimcore 5 demo website version.

For example at the portal page of the demo website I noticed in the symfony sql console that some unnecessary/multiple queries regarding sites were fired.

I've not installed a cache backend so the mysql cache is used.

The following query was executed 5 times:
image

This happens because the Site::getByDomain() method is called multiple times with the same domain. Therefore I think it makes sense to store the result of Site::getByDomain() in the runtime cache in order to avoid multiple calls to the backend cache.

Additionally I noticed multiple completely useless calls like this:
image

I added a check to avoid this too into the Staticroutes model.

### DIFF
--- a/pimcore/models/Site.php
+++ b/pimcore/models/Site.php
@@ -16,6 +16,7 @@
 
 namespace Pimcore\Model;
 
+use Pimcore\Cache\Runtime;
 use Pimcore\Logger;
 
 /**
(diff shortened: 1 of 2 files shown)


brusch:
Great, thank you very much!

Issue #1439, opened 2017-04-12
Object grid: add new objectbricks column throws error


Bhavik-kumar:
when I am trying to add any new **object bricks** column in product grid and try to sort by ID, getting 500 Below error.

```
Server threw exception - could not perform action. Please reload the admin interface and try again.

Status: 500 | Internal Server Error
URL: /admin/object/grid-proxy/classId/12/folderId/33?xaction=read&_dc=1491993093467
Params:
-> language: en_GB
-> class: Product



Fatal error:  Uncaught exception 'Zend_Db_Statement_Mysqli_Exception' with message 'Mysqli prepare error: Column 'o_id' in order clause is ambiguous' in /home/pimcore_ecommercedemo/www/vendor/zendframework/zendframework1/library/Zend/Db/Statement/Mysqli.php:77
Stack trace:
#0 /home/pimcore_ecommercedemo/www/vendor/zendframework/zendframework1/library/Zend/Db/Statement.php(115): Zend_Db_Statement_Mysqli->_prepare('SELECT COUNT(*)...')
#1 /home/pimcore_ecommercedemo/www/vendor/zendframework/zendframework1/library/Zend/Db/Adapter/Mysqli.php(388): Zend_Db_Statement->__construct(Object(Zend_Db_Adapter_Mysqli), 'SELECT COUNT(*)...')
#2 /home/pimcore_ecommercedemo/www/vendor/zendframework/zendframework1/library/Zend/Db/Adapter/Abstract.php(479): Zend_Db_Adapter_Mysqli->prepare('SELECT COUNT(*)...')
#3 /home/pimcore_ecommercedemo/www/vendor/zendframework/zendframework1/library/Zend/Db/Adapter/Abstract.php(828): Zend_Db_Adapter_Abstract->query(Object(Zend_Db_Select), Array)
#4 [internal function]: Zend_Db_Adapter_Abstract->fetchOne in /home/pimcore_ecommercedemo/www/vendor/zendframework/zendframework1/library/Zend/Db/Statement/Mysqli.php on line 77
```

Bhavik-kumar:
Hey guys, Any Solution ?

solverat:
I just experienced the same.

## Error
 `Integrity constraint violation: 1052 Column 'o_id' in order clause is ambiguous`

## Query  

```sql
SELECT COUNT(*) AS `totalCount` FROM `object_localized_7_de`  
[...]

PR #1453, opened 2017-04-22
fix ambiguous order in listing, see #1439

solverat:


solverat:
**second check**: this PR is maybe a difficult one. every listing with linked eg. joined object bricks needs an explicit orderkey. Because:

```php
$list->setOrderKey('myBrickColumn');
```
Will lead to an error:
`"Column not found: 1054 Unknown column 'object_localized_1_de.myBrickColumn'"`

After this PR every join needs to inject the table alias like:

```php
//second argument: do not quote argument 1.
$list->setOrderKey('myBrick.myBrickColumn', FALSE);
```

At least we need to add a info to the upgrade notes. Or find another solution...

solverat:
i just updated the PR. Should be save now. It does not require any further query configuration after updating.

brusch:
Thanks!

PR #1456, opened 2017-04-24
QuantityValue getFromWebserviceImport bugfix

markus-moser:
The getFromWebserviceImport method of the QuantityValue data type should not throw an exception if only parts of the field (only value or only unit) are filled. 
It should be a valid counterpart to the getForWebserviceImport method in order to make it possible to transfer the data from one pimcore instance to another one correctly via the REST webservice.

This pull request applies to Pimcore 5 - should I create the same one for Pimcore 4? It should be fixed there too!


fashxp:
yes please

PR #1458, opened 2017-04-25
check REQUEST_URI if set before matching

fheyer:
This avoids notices when run from tests without request.

I experienced a notice with a test that loads `\Pimcore\Config::getWebsiteConfig()` in a utility class.
`\Pimcore\Config::getWebsiteConfig()` calls `\Pimcore\Tool::isFrontentRequestByAdmin()` which produces the notice by accessing `$_SERVER["REQUEST_URI"]` without checking if it is set.

My fix avoids this.

### DIFF
--- a/pimcore/lib/Pimcore/Tool.php
+++ b/pimcore/lib/Pimcore/Tool.php
@@ -319,7 +319,7 @@ class Tool
             $isFrontend = false;
         }
 
-        if ($isFrontend) {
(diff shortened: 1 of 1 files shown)


maff:
Thanks!

maff:
Cherry-picked to master in 7e7b63096c8d16aab20c72777c88aa9d9e852653

PR #1459, opened 2017-04-25
Nonowner fields | bug | extjs

andreas-gruenwald:
Drag&Drop of non owner fields did not work reliably, because find() method did not determine the class name exactly. For instance if there existed "ShopProduct" and "ShopProductMaterial" classes then "ShopProductMaterial" was matching incorrectly.


brusch:
Thanks!

PR #1461, opened 2017-04-25
make object grid cell editing work again in pimcore 5

markus-moser:
Object grid cell editing didn't work anymore because the routing/url structure changed in pimcore 5.


weisswurstkanone:
thx

PR #1472, opened 2017-04-28
MultiRelation path formatter layout bugfix

markus-moser:
Grid layout needs to be refreshed after an update of the grid content occured otherwise strange things will happen if the new content's height is bigger then the original row height.


weisswurstkanone:
thx!

PR #1475, opened 2017-04-29
fix WorkflowController::submitWorkflowTransitionAction json return

dpfaffenbauer:
WorkflowController::submitWorkflowTransitionAction has a second parameter to $this->json, which is a boolean, but json expects it to be a http-header-code, therefore it throws an error with invalid http-code 1. this pr fixes that


fashxp:
thx very much

PR #1476, opened 2017-05-02
Navigation Doku: $this->container

Gerhard-Kanzler:
You need to get the Container from ViewModel before use:
$container = $this->getViewModel()->get('container');


brusch:
This was actually a bug and not an issue in the docs ;) 
I've just fixed it in f3c63b4d45a8bdc6250d167e8d04e38d6b44d0a3 . 

`$this->container` is already used by the DI-container, so we have to use a different variable, I've now choosen `$this->pages` (see also breadcrumbs - that makes it consistent between menus and breadcrumbs). 

`$this->getViewModel()->get('container');` won't work anymore as of build 49. 
Thanks anyway :)

PR #1484, opened 2017-05-04
Unifiy sub-request rendering

maff:
* support `attributes`, `query` and `options` everywhere
* define an interface for `DocumentRenderer` and implement it in new and legacy document renderer


maff:
Wrong merge base

PR #1503, opened 2017-05-08
Allow for resizing of details panel in Notes & Events view.

IgorBenko:
## Fixes Issue #
Details panel in Notes & Events view isn't resizable which makes it hardly usable for one of our use-cases. This pull requst keeps the same default size, but allows for resizing.

I will provide additional Pull Request for Pimcore 5 if this one gets accepted.


weisswurstkanone:
many thanks.

already merged into pimcore-5, no PR needed

PR #1504, opened 2017-05-08
The same file can't be uploaded twice, since the on-change event isn'…

IgorBenko:
The same file can't be uploaded twice, since the on-change event isn't triggered the second time.

To reproduce:
- upload a file
- delete it
- reupload the same file and nothing will happen

I will provide a Pull Request for Pimcore 5 if this one gets accepted.


weisswurstkanone:
many thanks!

weisswurstkanone:
already merged into pimcore-5

IgorBenko:
Thank you!

PR #1509, opened 2017-05-11
System Setting: Hide text edit panel for text assets

IgorBenko:
Problem:
We are working with large XML assets in the Pimcore backend. In some cases the XML Assets are few 100MB big. This of course totally breaks Pimcore GUI and Browser as it is trying to load the whole data into a textarea in the editor. Additionally it will exhaust the PHP memory very quickly and further increasing memory doesn't seem like a good solution just for this case, as opening the XML asset will still take too long because the data is read from the disk. 

Solution:
Checkbox in the Assets section of System Settings to turn of the editor for Text Assets. 

I will of course provide a separate Pull Request for Pimcore 5 if needed.

### DIFF
--- a/pimcore/config/texts/en.json
+++ b/pimcore/config/texts/en.json
@@ -5133,6 +5133,10 @@
         "definition": "Hide Edit Image Tab"
     },
     {
+        "term": "disable_text_edit_panel_tab",
+        "definition": "Disable Edit Tab for Text assets"
+    },
+    {
         "term": "merge_csv",
         "definition": "Import &amp; Merge CSV"
     },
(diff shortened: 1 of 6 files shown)


Yivan:
+1 We got this problem too on production server with sometimes big CSV uploaded in asset.

brusch:
Hey guys! 
Thanks Igor for your PR! I've solved the issue a different way (see #1520) - imho this is a bit better approach and avoids another setting. What do you think?

IgorBenko:
Hey Bernhard,

I find your approach great! I was also thinking in that direction but I thought it is too different to what has been already done with image assets. I would just suggest to make the error more verbose otherwise users won't understand why the preview is not available and that could lead to a lot of confusion. 

Thanks!

PR #1516, opened 2017-05-11
[BugFix] Don't create translations for empty strings

cleggypdc:
Trimmed $messageId's that return empty string translations appear to be saved.


brusch:
Thanks Paul!

PR #1536, opened 2017-05-16
bugfix: localized object relation data lost in field collection

diveshpahuja:
When using object relation field in localized block in field collection, updating data relation in field-collection removes the field-collection object relation data for all other objects of same type.

How to reproduce

- Create a field-collection(fields1) with localized block and add object relation field to localized block
- Create a class e.g. Test and add field collection-field with allowed types = fields1
- Create object(object1) of class Test
- In object1, relate object in localized field collection and save it
- Create object(object2) of class Test
- In object2, relate object in localized field collection and save it
- object1 field-collection data will get lost on object2 field-collection data update


weisswurstkanone:
many thanks!

build 4082

PR #1545, opened 2017-05-18
Skip locked jobs at execution time

YKWeyer:
Skip maintenance jobs that have a lock at execution time, unless we are running in a forced mode.

## Additional info  
In cluster environment, where the database and the `var` folder are shared across instances, running the jobs in parallel on several machines might lead to several issues. 
The current lock process check the state of jobs at initialization time, but assumes this state won't change before actually start processing the job. We should check whether the job is currently locked or not just before executing it.


brusch:
Thanks!

PR #1587, opened 2017-06-09
Exclude Jobs from maintenance script

andreas-gruenwald:
Exclude jobs via -ej option from maintenance script (such as cleaning up the cache).


maff:
Combined in #1588

PR #1588, opened 2017-06-09
Exclude Jobs from maintenance script 

maff:
Combines #1586 and #1587


YKWeyer:
It might be a silly question, but I'm not sure to understand where the `$validOptions` variable is set:
https://github.com/pimcore/pimcore/pull/1588/files#diff-59322edb430567a3d7c288082632df9eR39

maff:
Perfectly valid question :) Indeed, the `$validOptions` is missing due to a merge error and `-ej` doesn't behave as expected - will be fixed shortly.

maff:
Fixed in #1647 (will be published with the next build). The short option to exclude jobs is now `-J` (can be used multiple times), e.g.:

```
$ php pimcore/cli/console.php maintenance -J foo -J bar
```

PR #1560, opened 2017-05-31
[Static Routes] load controller from service

dpfaffenbauer:
## Please make sure your PR complies with all of the following points: 
- [x] Read and accept our [contributing guidelines](/CONTRIBUTING.md) before you submit a PR.
- [x] Features need to be proper documented in `doc/`
- [x] Bugfixes need a short guide how to reproduce them. 
- [x] We're not accepting any feature PR's only for **version 4** anymore, you have to provide a feature PR for both versions. 
- [x] Submit bugfixes for version 4 to the target branch `pimcore4`, version 5 is `master` branch.

**Don't submit a PR if it doesn't comply, it'll be closed without a comment!**
  
  
## Fixes Issue #
fixes #1405 

## Changes in this pull request  
Adds possibility to use a controller as a service in Static Routes.

PR #1593, opened 2017-06-11
in addition to pr #1560 fix staticroutes to append moduleName param

dpfaffenbauer:
Static Routes service controllers are only working with AppBundle services without this pr. This PR appends the moduleName param in the Static Routes grid.


maff:
Thanks!

PR #1598, opened 2017-06-13
Fix HTML Markup for edit template of areabricks

dominikschnieper:
## Fixes Issue
Fix HTML Markup for edit template of areabricks


maff:
Thanks!

PR #1599, opened 2017-06-13
Reduce timestamp columns to reasonable value range - int(11) unsigned

bbak:
## Fixes Issue #
Waste of resources in SQL-Schema, with associated performance penalties.

Reduce all columns that store Un*x timestamps to `int(11) unsigned` datatype. 

## Additional info  
MySQL's `int unsigned` has a max. of 4294967295. This value - assuming we manage the [Y2K38 Problem](https://en.wikipedia.org/wiki/Year_2038_problem) - would resolve to Feb 7th, 2106.
I'm quite confident this will be more than sufficient; and reducing to `int` saves 50% of the storage requirements (8 bytes for BIGINT vs 4 bytes für INT).

See MySQL Docs [here](https://dev.mysql.com/doc/refman/5.7/en/integer-types.html) and [here](https://dev.mysql.com/doc/refman/5.7/en/storage-requirements.html).


brusch:
Thank you very much!

Issue #543, opened 2016-04-11
ExtJS 6: checkbox HTML markup error


aarongerig:
The new checkbox in pimcore 4 (ExtJS 6) has to me a wrong HTML markup.

**Expected behaviour**
```html
<div class="pimcore_editable pimcore_tag_checkbox">
    <input type="checkbox">
    <label for="">Checkbox Label</label>
</div>
```
The label tag should be right next to the input[type="checkbox"] tag. This way you can style a checkbox correctly.

**Actual behaviour**
```html
<div class="pimcore_editable pimcore_tag_checkbox">
    <input type="checkbox">
</div>
<label for="">Checkbox Label</label>
```

brusch:
How did you add the label?
Actually I don't see the possibility to add a label at all. 

Thanks for clarifying.

aarongerig:
<div class="custom-check">
<?= $this->checkbox("myCheckbox") ?>
<label for="myCheckbox">My Checkbox</label>
</div>

Worked perfectly fine until pimcore v3.1.1, that's why I reported this issue.

brusch:
Thanks! 

But the output of  
```php
<?= $this->checkbox("myCheckbox") ?>
``` 
was also wrapped by a `div` in version 3 - so maybe it's a different issue?

aarongerig:
This is what it looked like in version 3:

```html
<div class="pimcore_editable pimcore_tag_checkbox">
    <div class="x-form-check-wrap" id="ext-gen55">
        <input type="checkbox" autocomplete="off" class="x-form-checkbox x-form-field" checked="">
        <label class="x-form-cb-label" for="ext-comp-1021" id="ext-gen56">&nbsp;</label>
    </div>
</div>
```

brusch:
Yes, but the label way generated by pimcore and was not part of your template :) 
So your example would produce the following output: 

```php 
<div class="custom-check">
    <div class="pimcore_editable pimcore_tag_checkbox">
        <div class="x-form-check-wrap" id="ext-gen55">
            <input type="checkbox" autocomplete="off" class="x-form-checkbox x-form-field" checked="">
            <label class="x-form-cb-label" for="ext-comp-1021" id="ext-gen56">&nbsp;</label>
        </div>
    </div>
    <label for="myCheckbox">My Checkbox</label>
</div>
```

aarongerig:
Sure, but why is there no <label> inside the <div> anymore?

brusch:
[...]

PR #1622, opened 2017-06-20
fix checkbox selector and add it to extjs context

solverat:
- Move checkbox tag element to extjs instead of a jquery context
- Add escaped selector function to get elements with dots in id attribute

## Some general information
If you're using a checkbox inside a areabrick, the id gets transformed with some dots and colons in it (thats because of the new naming strategy). jQuery and also extjs will trigger some errors and no checkbox element gets rendered. Because not document tag is using jQuery i decided to move the checkbox tag to a extjs element.

bildschirmfoto 2017-06-20 um 19 02 47

### DIFF
--- a/web/pimcore/static6/js/pimcore/document/tags/checkbox.js
+++ b/web/pimcore/static6/js/pimcore/document/tags/checkbox.js
@@ -10,44 +10,56 @@
  * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
  * @license    http://www.pimcore.org/license     GPLv3 and PEL
  */
-
 pimcore.registerNS("pimcore.document.tags.checkbox");
 pimcore.document.tags.checkbox = Class.create(pimcore.document.tag, {
 
-
     initialize: function(id, name, options, data, inherited) {
+
         this.id = id;
         this.name = name;
+        this.htmlId = this.id + "_editable";
+        this.element = null;
+
         this.setupWrapper();
-        options = this.parseOptions(options);
+
+        var checked = false,
+            domElement = Ext.query("#" + escapeSelector(id)),
+            options = this.parseOptions(options);
 
         if (!data) {
             data = false;
         }
 
-        this.htmlId = id + "_editable";
-        var checked = "";
         if(data) {
-            checked = ' checked="checked"';
+            checked = true;
         }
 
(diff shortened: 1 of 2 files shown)


brusch:
We went away from the ExtJS component since we faced a lot of styling issues after the ExtJS 6 migration, especially in combination with custom styles inherited from the surrounding layout. So I don't think it's a good idea to make a step back in that regard :| 

See also: https://github.com/pimcore/pimcore/commit/d5c5c00447f6bfde85069f64d84c434bf5cc430c#diff-415532c6b838f9821c2a69d7221c2d6e

solverat:
1. that's a pity. what kind of styling issues did you have? all the extjs markup goes into the pimcore_editable element - which is perfectly fine, even if you add some custom layouts around it?
2. with our without extjs, inside a areablock, the checkbox won't work without the escaping query seelctor.

brusch:
2. that's right, I'm already on it :)

brusch:
1. well that's nearly 2 years ago, so I don't know the issues in detail anymore, but as far as I can remember the problems were mainly in combination with text, either before the checkbox or after, eg.: 

```php
<?php while($this->block("foo")->loop()) { ?>
    My Text <?= $this->checkbox("bar") ?>
<?php } ?>
```

solverat:
i just tried that and it works flawless. 

bildschirmfoto 2017-06-21 um 09 41 40

anyway, it's not the flagship solution but at least it's working. Maybe there should be a ticket for some JS clean up, if i activate my ESLint JavaScript linter it gets a multiple heart attack. at least we should remove all the double quote stuff. :)

aarongerig:
See https://github.com/pimcore/pimcore/issues/543

solverat:
i still don't get it. why not using the label option and keep the extjs element (which implements the label automatically if set?

brusch:
As I said, there were multiple problems with the ExtJS component. 
What I still don't get is why the native HTML `<input>` is such a big problem?

solverat:
there is no big problem. :) it just feels a little bit wrong to use jquery in a complex extjs context. before your last commit the `getValue()` method was even nastier since this method gets triggered at least once per second by the extjs framework. every time a jquery selector [gets triggered ](https://github.com/pimcore/pimcore/commit/372b2425e17f3895c7d1f6e065ff1940ff89b21f#diff-3895c34d7f0b573d9ab9cf7f2d479078L50) which could causes performance issues. 

however, thanks for your time. ;)

brusch:
Well, I guess we can use Vanilla JS as well :)

PR #1625, opened 2017-06-21
Composer merge plugin

IgorBenko:
Use case:
- we want to checkout the code from our repository, install the Pimcore and start working
- in the current workflow, the installation will overwrite our own existing composer.json
- with this plugin in place, we can just create a composer.local.json file and the contents of both will be merged on the fly (without actually changing any of the both files)
(this would also reduce the problems during the Pimcore update, although there is already composer.json merging taking place)

If you have a better approach for this use case, please let me know.


maff:
Sounds nice! Can we use the latest released version (`1.4.0`) instead of `dev-master`? The only added feature on master is the `ignore-duplicates` option which is switched off.

IgorBenko:
@maff Yes, absolutely, 1.4.0 is perfectly fine.
Thanks!

maff:
Thanks! FYI the `composer.local.json` is now ignored in the standard `.gitignore`, but feel free to adapt this in your own projects :)

IgorBenko:
@maff OK, thanks a lot for merging :) But just wondering, why this decision to ignore the composer.local.json? I guess if someone has local dependencies, he will want by default to add them to the repository. This is kind of dangerous, as one can quickly forget about adding these files, especially with larger initial commits.

maff:
@IgorBenko that decision was mainly for development/deployment purposes as you can check out the Pimcore repo and add dependencies for testing and development without having local changes. The `.gitignore` most likely needs adaptions per project so I think it's OK to remove the `composer.local.json` line if needed (there's a comment block warning about ignoring the file by default).

IgorBenko:
Thanks for the explanation @maff. Makes sense.

stephenharris:
While this plug-ins a really good addition - isn't putting the configuration in `composer.json` going to make it impossible for users to configure it differently? Suppose I want to set `merge-extra` to `true` then I would update the `composer.json` file, but then when I update Pimcore, my changes are over-written by Pimcore's recursive merge.

I think we should add the plugin, but not the configuration, and thus allow the end user to decide what configuration suits their project. Updating Pimcore then won't replace the user's configurations because it only replaces what it knows about.

Issue #1510, opened 2017-05-11
[Pimcore 5] Performance Optimizations


brusch:
- response time
...
...

brusch:
Memo: 
Some parts are called several times for no reason, eg.: 
https://github.com/pimcore/pimcore/blob/b773b667bbe96cdac6bf5ac89150359234cb0300/pimcore/lib/Pimcore/Service/MvcConfigNormalizer.php#L117-L117 
There should be room for improvement.

PR #1630, opened 2017-06-22
Event handler optimizations

maff:
* Add a `profiler` context to make sure context specific event handlers to not run on profiler requests
* Refine event handler definitions (user context resolver interface everywhere)
* Make sure BlockState, ContentTemplate and FullPageCache listeners run only in frontend context


maff:
Also related to #1510

Issue #1635, opened 2017-06-23
CSS Styles of class layout component are ignored.


geraudi:
# Bug Report
  
### Expected behavior 
When I edit an object, layout's CSS have to be apply.

### Actual behavior  
CSS for layout component are ignored when I edit an object.

### Steps to reproduce  
1. Create or edit a class.
2. Add "background-color: red" in field "CSS style" of a layout Component and Save.
3. Add or edit a new object.
4. The background is still white.
  
### Fix 
In web/pimcore/static6/js/pimcore/object/helpers/edit.js, line 181, replace 
```JavaScript
newConfig = Object.extend(newConfig, xTypeLayoutMapping[l.fieldtype]);
```
by
```JavaScript
newConfig = Object.extend(xTypeLayoutMapping[l.fieldtype], newConfig);
```

weisswurstkanone:
Already fixed. See https://github.com/pimcore/pimcore/pull/1636

PR #1636, opened 2017-06-23
CSS Styles of class layout component are ignored.

geraudi:
newConfig's properties have to overwrite xTypeLayoutMapping's properties to take in account the configuration of class layouts. (web/pimcore/static6/js/pimcore/object/helpers/edit.js)
  
## Fixes Issue #1635


brusch:
Thanks!

PR #1638, opened 2017-06-23
[PIMCORE4] New feature: define di configs inside plugins.

kkarkus:
From now you can define custom di configs inside plugins.

`Pimcore::initPluginsDi` runs before any `Plugin::init` because if you need there di container it must be completely built.

### DIFF
--- a/docs/Development_Documentation/10_Extending_Pimcore/13_Plugin_Developers_Guide/01_Plugin_Anatomy.md
+++ b/docs/Development_Documentation/10_Extending_Pimcore/13_Plugin_Developers_Guide/01_Plugin_Anatomy.md
@@ -19,6 +19,8 @@ The folder structure of a plugin should look as follows:
 plugins
 └───ExtensionExample
     │
+    └───config (this is totally optional)
+    │   |   di.php
     ├───controllers
     │   │   IndexController.php
     │   │   ...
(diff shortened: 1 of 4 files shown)


dpfaffenbauer:
You know there is an Event for that?

'system.di.init' -> gets called before the container gets build

kkarkus:
I will try to explain why event doesn't work in some cases.

# Plugin structure:
ExamplePlugin1
- inside `init` method you want to get "SingletonExampleService" but when you use `\Pimcore::getDiContainer()` it won't build definitions from ExamplePlugin2 because scandir loads dirs in alphabetic order
 
ExamplePlugin2
- here you attach to this event because you want to create singleton object e.g. "SingletonExampleService"


It's simplest case, I hope you understand what I meant to say.

kkarkus:
@brusch @weisswurstkanone could you look at this pull request?

brusch:
@kkarkus we have to discuss this internally first - basically Pimcore 4 is 'support only' so new features only go to the v5 branch.

brusch:
Note: docs missing (please wait until the final decision before updating the PR)

kkarkus:
@brusch ok, I'll wait then.

brusch:
I'm not really happy with the architecture since you're mixing plugin specific code into the `Pimcore` class. 
I think moving `initPluginsDi()` to `\Pimcore\ExtensionManager` makes more sense, and then call it when building the container: `foreach (ExtensionManager::getDiPaths() as $diPath) { ...`

In general: Any other opinions on that? 
Imho this is quite a special use case, which is obsolete with v5 anyway.

kkarkus:
@brusch 
I've just pushed fixed code.
I know v5 is better but I need to use this in v4.

kkarkus:
@brusch docs are updated, please let me know if it's enough for you

brusch:
I'll leave this PR open until it has some more supporters. At the moment it seems you're the only one who needs this particular functionality.

brusch:
Closed due to inactivity / lack of supporters.

PR #1654, opened 2017-06-28
Consider role permissions in price rule overview.

andreas-gruenwald:
Ecommerce Framework bugfix:
the price rule controller now also takes role permissions into account, not only user permissions.

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/Controller/PricingController.php
+++ b/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/Controller/PricingController.php
@@ -37,17 +37,8 @@ class PricingController extends AdminController implements EventedControllerInte
     public function onKernelController(FilterControllerEvent $event)
     {
         // permission check
-        $key = 'bundle_ecommerce_pricing_rules';
-        $access = $this->getUser()->getPermission($key);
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
is this really necessary? Pimcore does the same internally:

```php
//Pimcore\Bundle\AdminBundle\Controller\AdminController:73

if (!$this->getUser() || !$this->getUser()->isAllowed($permission)) {
    $this->get('monolog.logger.security')->error(
        'User {user} attempted to access {permission}, but has no permission to do so', [
            'user'       => $this->getUser()->getName(),
            'permission' => $permission
        ]
    );

    throw new AccessDeniedHttpException('Attempt to access ' . $permission . ', but has no permission to do so.');
}
```

fashxp:
hmm, this shouldn't be necessary ... see https://github.com/pimcore/pimcore/blob/master/pimcore/models/User.php#L318-L318

or do I miss something?

andreas-gruenwald:
Ok, so we  can just replace the part 
`$this->getUser()->getPermission($key)`
by
`$this->getUser()->isAllowed($permission)) { `,
right?

isAllowed() considers roles, while getPermission doesn't seem to?

fashxp:
thats right...

fashxp:
thanks

PR #1655, opened 2017-06-28
[Admin] refactor TokenStorageUserResolver to get rid of Container dependency

dpfaffenbauer:
Is it really necessary to have a Container dependency inside the TokenResolver? Shouldn't this (PR) be enough? I mean, if the security.token_storage service is missing (eg. SecurityBundle is missing as well), something is wrong anyway.


dpfaffenbauer:
There is no benefit for the extra returns, thats more like my coding style ;), when a function should return something, I always return something. But, if your happy with it, I'll remove them

maff:
Thanks :)

PR #1665, opened 2017-06-30
Rewrite mapped pimcore IDs when importing WYSIWYG data in objects via RestImporter

BlackbitNeueMedien:
## Problem encountered
I synchronize objects between two pimcore instances using the RestImporter. The class contains a WYSIWYG field. Images that have been dragged onto the field get removed in the target pimcore instance after import.

## Analysis
The `getFromWebserviceImport` method in `\Pimcore\Model\Object\ClassDefinition\Data\Wysiwyg` does currently not modify the ids in the html text. Because the ids from the source pimcore were not found in the target system, the corresponding image elements were removed in the save process.

## Proposed solution
I have changed the `rewriteIds` method to be able to make use of id mapper from the RestImporter plugin. This way, the `pimcore_id` attributes are replaced with the new asset (or document) ids in the target system, given that they have been synchronized before.

### DIFF
--- a/pimcore/models/Object/ClassDefinition/Data/Wysiwyg.php
+++ b/pimcore/models/Object/ClassDefinition/Data/Wysiwyg.php
@@ -299,9 +299,29 @@ class Wysiwyg extends Model\Object\ClassDefinition\Data
      */
     public function rewriteIds($object, $idMapping, $params = [])
     {
+        $data = $this->getDataFromObjectParam($object, $params);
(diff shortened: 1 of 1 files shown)


brusch:
Hmm, maybe this is better done with an event-based approach :|
The current solution doesn't seem to be very generic.

brusch:
As mentioned already earlier, a event based approach would be more appreciated.

PR #1669, opened 2017-06-30
Homogenize the way we retrieve thumbnails in Image Tag

YKWeyer:
I experienced a weird behaviour with DI on Images and Thumbnails: when defining custom classes in my `di.php` file, the custom Thumbnail class was being used only when a specific thumbnail config was defined in the Tag.

It apparently lies on both image thumbnails not being retrieved the same way. Getting a thumbnail with `new Asset\Image\Thumbnail($image);` seems to ignore the defined overwrite rules (I am not so sure to understand why though).

### Questions/Remarks
- Why is the DI class not being called in this case?
- Deferred option: I let it to its default value, not so sure what would be the best option here

### DIFF
--- a/pimcore/models/Document/Tag/Image.php
+++ b/pimcore/models/Document/Tag/Image.php
@@ -225,7 +225,7 @@ class Image extends Model\Document\Tag
                 $thumbnail = $image->getThumbnail($thumbConfig, $deferred);
             } else {
                 // we're using the thumbnail class only to generate the HTML
-                $thumbnail = new Asset\Image\Thumbnail($image);
+                $thumbnail = $image->getThumbnail();
             }
 
             $attributes = array_merge($this->options, [


brusch:
How does your `di.php` look like? 
Just to avoid misunderstandings: it is not possible to customize/overwrite `\Pimcore\Model\Asset\Image\Thumbnail` :)

YKWeyer:
Hey,

I know, that's why I had to override Image first:
```
return [
    'Pimcore\Model\Asset\Image' => DI\object('Website\Model\Asset\Image'),
];
```

And in `Image.php`:
```
namespace Website\Model\Asset;

class Image extends \Pimcore\Model\Asset\Image
{
    public function getThumbnail($config = null, $deferred = true)
    {
        return new Image\Thumbnail($this, $config, $deferred);
    }
}
```

brusch:
Well, I guess it was just to avoid unnecessary function calls :)

PR #1672, opened 2017-07-03
fix nearest document by path matches

solverat:
some explanation:

1.) https://github.com/pimcore/pimcore/compare/master...solverat:fixNearestDocumentPath?expand=1#diff-1570666f8a7cd8c0e8886d72b4134db1R120

This is weird anyway but i just fixed the passing argument (not the whole request, just the path info)

2.) https://github.com/pimcore/pimcore/compare/master...solverat:fixNearestDocumentPath?expand=1#diff-dd59407f672dfbb4c15d1ec7801489abR148

To reproduce this little (but nasty)  error:

bildschirmfoto 2017-07-03 um 21 09 52

- create a static route "news_detail" (/de/news/news-1)
- create a site (see screenshot)
- open url (/de/news/news-1)
- debug `getNearestDocumentByPath()`
- it returns the ID `1` -> the root page, which is wrong, the nearest page is `/de/` or even `/pimcore5.dev` 

Since this method only checks hardlink parents it will never get to the next closest document.

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
+++ b/pimcore/lib/Pimcore/Bundle/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
@@ -117,7 +117,7 @@ class DocumentFallbackListener extends AbstractFrontendListener implements Event
         // this is only done on the master request as a sub-request's pathInfo is _fragment when
         // rendered via actions helper
         if ($event->isMasterRequest()) {
-            $document = $this->documentService->getNearestDocumentByPath($request);
+            $document = $this->documentService->getNearestDocumentByPath($request->getPathInfo());
             if ($document) {
                 $this->documentResolver->setDocument($request, $document);
             }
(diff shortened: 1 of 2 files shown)


solverat:
mhmm.. there is something more going on here. the second fix triggers the `handleDirectRouteRedirect` and redirects to the just found page instead of mapping.

brusch:
Regarding the first fix: https://github.com/pimcore/pimcore/compare/master...solverat:fixNearestDocumentPath?expand=1#diff-1570666f8a7cd8c0e8886d72b4134db1L120
Shouldn't be necessary as `getNearestDocumentByPath()` extracts the path info automatically: 
https://github.com/pimcore/pimcore/blob/2bb14ffbd193b1137304e09eac4243576dd2c5ab/pimcore/models/Document/Service.php#L523-L523

solverat:
@brusch ah i see. well then this pr is going to be a irrelevant one. :)

solverat:
@brusch since you've closed this PR and problem 2 is a still valid one: should i open an issue or is it already on your radar (I've seen it's on the week 28 project [awaiting triage])? :)

brusch:
#1733

PR #1673, opened 2017-07-03
remove cache boolean casting to allow custom cache tags

solverat:


dpfaffenbauer:
In my opinion: The boolean should stay a boolean and not be checked if string or not. This could lead to confusion. There should be an additional parameter which defines additional Cache identifier tags.

brusch:
Thanks

PR #1679, opened 2017-07-04
[PIMCORE4] Support static files in modules

cleggypdc:
I found that I had to make this adjustment to get /modules/static/* to return any static files in modules.


brusch:
Thanks

Issue #1550, opened 2017-05-23
[ECOMMERCE] Float Issues


brusch:

brusch:
- new projects use only int
- flag for float support

maff:
Implemented in #1701

PR #1701, opened 2017-07-06
[ECOMMERCE] Handle prices as integer value object with a fixed scale 

maff:
Description follows.

See #1550


pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/PricingManager/Action/ProductDiscount.php
```diff
@@ -27,26 +32,34 @@ class ProductDiscount implements IProductDiscount
...
-        $priceinfo->setAmount($amount > 0 ? $amount : 0);
+
+        $amount = Decimal::create($this->amount);
+
+        // TODO use discount()?
+        if ($amount->equals(Decimal::create(0))) {
+            $amount = $priceinfo->getAmount()->mul($this->getPercent() / 100);
+        }
+
+        $amount = $priceinfo->getAmount()->sub($amount);
```
maff: Good catch, fixed.

Issue #1639, opened 2017-06-23
Numeric getData does not return proper value for web2print


Cruiser13:
# Bug Report
### Expected behavior 
Same (correct) output in preview mode and in web2print pdf.
### Actual behavior  
getData() does return different values
#### Code example 
```php
if($this->editmode){
echo $this->numeric("amount1");
}
else{
if($this->numeric("amount1")->getData() === 0){
echo "You did input a zero";
}
}
```

Put a 0 in the number field while editing the document. In preview, you'll get the text "You did input a zero" but it'll not be printed on the pdf.

Preview mode will output a 0 as int(0) while web2print pdf will list a string(1) "0".
Probably connected to https://github.com/pimcore/pimcore/issues/678

Cruiser13:
As a follow-up: The isEmpty() function of the numeric editable returns true for no input and a 0 inside the input. If you use the numeric for currency amounts for example, you can not decide wether you should display a 0,00 $ in the frontend or nothing because the info is missing.

Cruiser13:
I wonder if numeric editable is designed for 0 values at all. If a numeric editable is being used inside a block, the 0s are deleted leaving an empty field on reload.

maff:
Empty and zero are now properly handled (see 49ed7bd00853f5e3a3f3b9d888d61107b41d07cb).

Cruiser13:
@maff this does not solve the issue that 0s dissapear if the document does reload.

maff:
@Cruiser13 should be fixed with 4ecc86ab20dc8516f061190620ab4fa328e6b08b

Cruiser13:
@maff thank you that fixed the issue for me.

PR #1704, opened 2017-07-07
Improvements to numeric data field

maff:
As preparation for #1550 

- [x] Allow to specify both precision and scale for decimal fields `DECIMAL(precision, scale)` - currently only the scale can be set and the precision defaults to 64
- [x] Do not cast decimal results to float, return them as string (or decimal value object?) instead


Yivan:
@maff Would be great to make it for quantity value, as for now using it for monetary (multicurrency) is not possible because it can not be decimal.

Some other people need it, and i think it will be very interresting for quantity value to be able to use it has decimal (so we are sure the quantity is not rounded).
Please see: https://github.com/pimcore/pimcore/issues/522

Just for reference, two pull request where made (including adding min/max too):
https://github.com/pimcore/pimcore/pull/845
https://github.com/pimcore/pimcore/pull/845

Thanks!

Cruiser13:
#1639 could be of interest for this pull request too.

PR #1708, opened 2017-07-07
save tag: fix exception when tag does not yet exist

i-telligence-af:

When saving a tag, only attempt to get idPath of existing tag when tag has an ID (i.e. when the tag already exists). 

## Additional info  
To reproduce bug (before fix):
Try creating a new tag, either through "Tag Configuration" or via the Tags tab on an asset.


brusch:
Thanks!

PR #1715, opened 2017-07-07
[Pimcore 4] Added type restrictions into workflow manager

cleggypdc:
The workflow Manager now checks all element types including object sub types such as variants

Can be easily ported to Pimcore 5 :)


fashxp:
added some additional docs and ported to pimcore 5 - see https://github.com/pimcore/pimcore/issues/1849

thx very much.

Issue #1627, opened 2017-06-21
Adminer Integration: Styling Issues


brusch:
Sometimes the stylesheets/javascript files are not loaded.

brusch:
PR  #1716 didn't fix all problems, still experiencing missing styles.

dpfaffenbauer:
should be fine now?

brusch:
See also #1806

PR #1716, opened 2017-07-09
[#1627] Adminer Integration: Styling Issues

dpfaffenbauer:
Adminer behaves quite weird on first request. It loads all available database but flushes the output (https://github.com/vrana/adminer/blob/master/adminer/drivers/mysql.inc.php#L336 + https://github.com/vrana/adminer/blob/master/adminer/include/functions.inc.php#L1227). Not sure what the idea behind this is. This PR solves the issue and Adminer gets displayed properly.


brusch:
Thanks!

PR #1722, opened 2017-07-12
Fix for the update command: --ignore-maintenance-mode

IgorBenko:
## Fixes Issue #
Update puts the system into maintenance mode, so the internal:update-process command should ignore maintenance mode when being run

Adds --ignore-maintenance-mode flag to the call


brusch:
Thanks!

PR #1731, opened 2017-07-14
No-interact mode support for pimcore:update command

IgorBenko:
When pimcore:update command is executed with -n switch (no-interaction) default the answer to yes


brusch:
Should we better use `-y` as `apt` does? 
`-n` is often used as dry run as well?

IgorBenko:
Hi Bernhard, I find -y just as good. Should I prepare a Pull request for that?

brusch:
Sorry, I forget about that `-n` is Symfony's standard - so everything is fine - thanks :)

Issue #1747, opened 2017-07-20
New Data Type - Input Quantity Value


mgomse:
## Feature description 
Quantity Value Data Type is really great but sometimes there is a need for some kind of merge between Quantity Value and Input Data types. 

I think that the best use case is that:
* we import some attributes from an external system (e.g. CSV file). Let say it looks like this:
    * Attributes Import

AttributeId | Attribute Name
------------ | -------------
1 | Weight
2 | Dimension (H/W/D)
3 | Type

 * Product Import

ProductId | AttributeId | Attribute Value | Attribute Unit 
------------ | ------------- | ------------ | -------------
1 | 1 | 20 | kg
1 | 2 | 2 / 3 / 4 | H/W/D
1 | 3 | Super Product | {no_unit_here}
* we store attributes them into classification store
* we cannot predict the type of attribute so we need to choose one
    * normally we will choose input
    * but we MUST import also units
    * so it seems that Quantity Value is a great type
    * but Value sometimes is not a "Quantity"

And here is my proposition for new type: "Input Quantity Value" - based mainly on Quantity Value, but with Input field for Value (not Number). An example of view (in the Object Classification Store Keys view) is below:

![image](https://user-images.githubusercontent.com/16497506/28413872-76343f8a-6d49-11e7-8e18-fa49cfd86cc8.png)

weisswurstkanone:
Any updates on this one? See my comments in https://github.com/pimcore/pimcore/pull/1748

brusch:
@mgomse  Anything?

mgomse:
@weisswurstkanone @brusch 

Hi, sorry, but somehow I missed emails... I just saw your comment in #1748 so I'll try to do fixes as soon as possible and I'll let you know about progress

weisswurstkanone:
great 👍 please don't hesitate to contact me if there are any questions

mgomse:
FYI: #1748 is updated and I also added a comment with explanations: https://github.com/pimcore/pimcore/pull/1748#issuecomment-326748585

sumit1337:
Hello sir I am facing below issue on saving quantity value  in my pimcore class
Please help me as soon as possible

 Reading... IDW Package.csv
07:18:21 ERROR     [pimcore] Class: Pimcore\Model\DataObject\Package => call to undefined method setLowestLevelUOMImperial

In AbstractModel.php line 266:
                                                                                                
[...]

PR #1748, opened 2017-07-20
New Data Type - Input Quantity Value

mgomse:
## Fixes Issue #
https://github.com/pimcore/pimcore/issues/1747


mgomse:
If you'll like this data type please let me know so I'll upgrade the doc - I think that mainly it will be from the https://github.com/pimcore/pimcore/issues/1747

This is my first PR so sorry for any mistakes and if I can improve for later please let me know 😃

weisswurstkanone:
Hi!

Many thanks! Could you please make the following changes and resubmit the PR?

- icon is missing. If we want to use the same icon just change the line 769 in icons.css to 
.pimcore_icon_quantityValue, .pimcore_icon_inputQuantityValue

-translation files are not needed except en.json

- quantityValue is editable (double-click opens an editor window) in the grid view but inputQuantityValue isn't. Please note that there was an improvement a couple of weeks ag. See https://github.com/pimcore/pimcore/commit/25890a936db77b96382f2f832b922f9b50a073a8
This should probably be incorporated.

- quantityValue allows you to define a default value (in the class editor) which has to be numeric. This doesn't make sense for inputQuantityValue (be aware that there is even a float cast in getDefaultValue).

- maybe (for the future) I also can imagine input validation similar to the input data type as a cool improvement

-InputQuantityValue shared Pimcore\Model\Object\Data\QuantityValue with the QuantityValue data type.

A consequence for example is that:
- phpdoc is wrong (e.g. getValue currently claims that it returns a float)

Thx! Would be cool to have this in the portfolio in the near future :-)

mgomse:
@weisswurstkanone First of all thank you very much for your review here. I just put my changes and below you can find my comments. 

> icon is missing. If we want to use the same icon just change the line 769 in icons.css to
.pimcore_icon_quantityValue, .pimcore_icon_inputQuantityValue

In my opinion, the best will be **add_row.svg** (for me it seems like merging input and quantity value :))

> translation files are not needed except en.json

Done

> quantityValue is editable (double-click opens an editor window) in the grid view but inputQuantityValue isn't. Please note that there was an improvement a couple of weeks ag. See 25890a9
This should probably be incorporated.

Done: 
image

> quantityValue allows you to define a default value (in the class editor) which has to be numeric. This doesn't make sense for inputQuantityValue (be aware that there is even a float cast in getDefaultValue).

I totally agree - this is also done: 
image

> maybe (for the future) I also can imagine input validation similar to the input data type as a cool improvement

👍 for this idea, 😄 I'll do my best to implement it but of course, if somebody will be faster don't hesitate 😄 

> InputQuantityValue shared Pimcore\Model\Object\Data\QuantityValue with the QuantityValue data type. ...

I created `\Pimcore\Model\DataObject\Data\InputQuantityValue` and I changed a lot `\Pimcore\Model\DataObject\ClassDefinition\Data\InputQuantityValue`. I think that it is OK right now but of course please let me know what do you think.

Thank you again and I'm looking for some new comments 😸

weisswurstkanone:
Minor things:
- please add the new datatype to the readme

- phpdoc for things like getFromCsvImport, getFromWebserviceImport, ... ?

- what does this comment mean? 
"... TODO: Refactor - this class is very similar to the parent one so probably we can try to refactor parent and have better results here also ..."

- pimcore.object.tags.inputQuantityValue: I guess valueInvalid could be removed entirely ?

mgomse:
> please add the new datatype to the readme
👍 I'll do it as soon as possible

> phpdoc for things like getFromCsvImport, getFromWebserviceImport, ... ?

👍 I'll update it - it wasn't there because I thought that it will be better to inherit PHP-doc from the parent (QuantityValue). But truly we return different DataObject type

> what does this comment mean?
> "... TODO: Refactor - this class is very similar to the parent one so probably we can try to refactor parent and have better results here also ..."

My idea was that if we will move to the QuantityValue (parent) the method `getNewDataObject()` we won't need to rewrite so many methods. But I need to think more about PHP-docs ... But generally, I don't like code redundancy and right now methods like `getFromWebserviceImport`, `unmarshal` are copy/paste and changed just return type... I believe that it can be done in a better way 😄 

> pimcore.object.tags.inputQuantityValue: I guess valueInvalid could be removed entirely ?

👍 That's right ... sorry for such mistakes... I'll fix it asap.

To sum it all I'll try to find some time soon and I'll let you know about progress. Thank you again very much for the review and sorry for some mistakes. 

If you will have a minute please let me know what do you think about this code redundancy

PR #1751, opened 2017-07-24
Fix "Limit reached" bug in objectbricks.js

kkarkus:
How to reproduce this bug?

1. Add objectbricks field in some class, set limit e.g. to 1.
2. Create new object.
3. Press green (+) button in object brick field.
4. After new element is added, press red (x) button to remove it.
5. Try go to step 3, but you'll get error message "Limited reached".


weisswurstkanone:
thx!

PR #1752, opened 2017-07-24
[PIMCORE4] Fix "Limit reached" bug in objectbricks.js

kkarkus:
Same bug as https://github.com/pimcore/pimcore/pull/1751 but here's PIM4 fix.


weisswurstkanone:
thanks!

PR #1753, opened 2017-07-24
[ROUTING] Add a list of parameters which are mapped to _locale attribute if not set

maff:
Many older projects use parameters like `language` to define a language in the URL. As this is now handled by Symfony's default `_locale` parameter, all existing static routes and code relying on their params needs to be updated. This PR adds a mapping to resolve existing params like `language` to the `_locale` attribute. The mapping is only used if no explicit `_locale` parameter is set. 

Valid params can be set via config:

```yaml
# will map a static route parameter language to _locale
pimcore:
    routing:
        static:
            locale_params:
                - language
```

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/CoreBundle/DependencyInjection/Configuration.php
+++ b/pimcore/lib/Pimcore/Bundle/CoreBundle/DependencyInjection/Configuration.php
@@ -67,6 +67,7 @@ class Configuration implements ConfigurationInterface
         $this->addDocumentsNode($rootNode);
         $this->addModelsNode($rootNode);
 
+        $this->addRoutingNode($rootNode);
         $this->addCacheNode($rootNode);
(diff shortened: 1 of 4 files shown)


dpfaffenbauer:
Shouldn't be stuff like this be done in the LegacyBundle? (https://github.com/pimcore/pimcore4-compatibility-bridge)

maff:
The thing is that this also affects projects which migrate to v5 without the legacy support, but do not want to change all their static routes and locations generating routes from the currently used param to `_locale`. Therefore we need to keep the optional mapping in the core static route router.

brusch:
We should add a hint for this feature to the docs (maybe here doc/Development_Documentation/06_Multi_Language_i18n/README.md and here doc/Development_Documentation/02_MVC/04_Routing_and_URLs/02_Custom_Routes.md) 

Thanks

PR #1757, opened 2017-07-25
Property added to forbidden names for object

mgomse:

## Additional info  

1. Create new class
2. Add a Panel and an Input
3. Name Input "property"
4. Save class definition
4.1. There is no Error in here
5. Try to create an object
5.1. Here we have an error


weisswurstkanone:
thx

Issue #1733, opened 2017-07-14
staticroutes in hardlinks context and direct calls of hardlinks are in wrong context  


solverat:
## direct call of a hardlink document
1. create a document (`/de`) set language to `de`
2. create a hardlink (`/de-at`), link it to the document, enable both checkboxes and set language to `de_AT`
3. open hardlink in frontend `/de-at`: the request locale is still `de` but it should be `de_AT` (i think all the overwritten properties should be available).
4. if you create a subpage in the hardlink context, everything works as expected.
 
## call a staticroute in hardlink context
1. create a document (`/de`) set language to `de`
2. create a hardlink (`/de-at`), link it to the document, enable both checkboxes and set language to `de_AT`
3.  create a staticroute: 
3.1 `pattern: /([a-z-]+)\\/your-slug\\/([A-Za-z0-9._-]+)\\-([0-9]+)/`  
3.2 `reverse: /%_locale/your-slug/%entryName-%entryId`
3.3`variables: _locale,entryName,entryId`
4. call staticroute: `/de-at/your-slug/my-entry-1`
5. since pimcore is calling the `getNearestDocumentByPath()` method, pimcore should get all the properties from the `de-at`hardlink document, instead the root document `1` gets loaded.

solverat:
fixed in 87c44978ac91fd46fb355bff41388eb75a1a4d07. successfully tested with build 82.

maff:
@solverat thanks for testing! The change also fixed a third issue but I didn't find the time to test if it addresses this issue as well.

solverat:
@maff i'm just working on a special customer case and i'm wondering if we could have short discussion about the `getNearestDocumentByPath()` method. Please check this line:

https://github.com/pimcore/pimcore/blob/5483c6f0fa00cfb4026b15ef5ffbfaf9dcee66ca/pimcore/lib/Pimcore/Bundle/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php#L136

In my case we have site structure like:

```yml
- site1 #pimcore site
     - de # a link (!) to a subpage, but with the property "language" for german
     - en # default start page with property "language" for english
- site2 #second pimcore site
     - de
     - en
```
As you can see `site1/de` is a link which does not fit into the `getNearestDocumentByPath` standard. Now i had to implement a specific static route like `site1/de/my-static-route`. The `getNearestDocumentByPath`will return `site1` instead of `site1/de/` which would be the real start of the page where we define the language and so on.

[...]

PR #1767, opened 2017-07-27
Static routes inside a site do not resolve nearest document properly

maff:
When loading a static route inside a site, the nearest document isn't properly resolved as it does not take the site into account.


maff:
Also fixes #1733

PR #1770, opened 2017-07-27
add myclabs/deep-copy to composer.json

solverat:
the "myclabs/deep-copy" package only gets installed in dev mode since the only dependency is in "phpunit/phpunit" - which is not available in `--no-dev` mode.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,6 +65,7 @@
     "linfo/linfo": "~3",
     "monolog/monolog": "~1",
     "mpratt/embera": "~1",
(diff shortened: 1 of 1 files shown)


maff:
Good catch! PHPUnit actualls specifies the dependency as `~1.3` which currently results in `1.6.1` being installed. By specifiying `~1.3.0` it would downgrade to `1.3.1`. Could you update the constraint to `~1.3`?

solverat:
true story. done!

maff:
thx!

PR #1771, opened 2017-07-28
Composer file path

solverat:
## tltr;
- `composerUpdate()`: only update the real pimcore `composer.json`
- `installData()`: move `composer.json` to the real pimcore directory
- `composerDumpAutoload()`: `dump-autoload` only in real pimcore directory

## About this PR
This PR will force pimcore to use the real pimcore `composer.json` file, since the `PIMCORE_PROJECT_ROOT` could lead to a different path and a thereby to a different `composer.json`.

### DIFF
--- a/pimcore/config/constants.php
+++ b/pimcore/config/constants.php
@@ -30,6 +30,10 @@ if (!defined('PIMCORE_COMPOSER_PATH')) {
     define('PIMCORE_COMPOSER_PATH', PIMCORE_PROJECT_ROOT . '/vendor');
 }
 
+if (!defined('PIMCORE_COMPOSER_FILE_PATH')) {
+    define('PIMCORE_COMPOSER_FILE_PATH', realpath(__DIR__ . '/../../'));
+}
+
 if (!defined('PIMCORE_APP_ROOT')) {
     define('PIMCORE_APP_ROOT', PIMCORE_PROJECT_ROOT . '/app');
 }
--- a/pimcore/lib/Pimcore/Update.php
+++ b/pimcore/lib/Pimcore/Update.php
@@ -446,7 +446,7 @@ class Update
 
         try {
             $composerPath = \Pimcore\Tool\Console::getExecutable('composer');
-            $process = new Process($composerPath . ' dumpautoload -d ' . PIMCORE_PROJECT_ROOT);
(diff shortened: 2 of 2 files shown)


brusch:
Thx

brusch:
@solverat Hmm, I'm just thinking, ... does this really makes sense? 
There's also `PIMCORE_COMPOSER_PATH` which is defined as: 

```php
define('PIMCORE_COMPOSER_PATH', PIMCORE_PROJECT_ROOT . '/vendor');
```

Per default, the `/vendor` folder is always within the same directory as the `composer.json`, so why should the `composer.json` be at a different location? 
Wouldn't it be better to change it as follows: 

```php
define('PIMCORE_COMPOSER_FILE_PATH', PIMCORE_PROJECT_ROOT));
define('PIMCORE_COMPOSER_PATH', PIMCORE_COMPOSER_FILE_PATH. '/vendor');
```

In this case the default value for `PIMCORE_COMPOSER_FILE_PATH` would be the project root (which makes sense in most cases), but if it get's overridden, also `PIMCORE_COMPOSER_PATH` changed accordingly.

brusch:
@solverat see https://github.com/pimcore/pimcore/commit/aaed2faee7bb47031556624bfc685ea7303e7426

solverat:
@brusch thx but now we have the same issue again. If you're going to install pimcore as a dependency, the `composer.json` is inside the vendor while the PIMCORE_PROJECT_ROOT is on top. For example:

```php
/var/www/src /* this is our project root */
/var/www/src/composer.json /* => our project composer file */

/var/www/src/vendor/pimcore/pimcore/ /* => this is the pimcore folder */
/var/www/src/vendor/pimcore/pimcore/composer.json /* => this is the pimcore composer file -> that's the one which should get updated. */
````
pimcore should only try to get its own composer.json, regardless of the PIMCORE_PROJECT_ROOT (which could be a complete different location).

brusch:
Well, that depends on the point of view ... you'd have to override both `PIMCORE_COMPOSER_FILE_PATH` and `PIMCORE_COMPOSER_PATH` in that case.

Nevermind, I'll leave it as it is for now.

solverat:
Ah! I'm an idiot! Sorry - you're totally right, thanks! :)

PR #1773, opened 2017-07-28
add PIMCORE_PROJECT_ROOT to internal php commands and processes

### LINKED ISSUE #1681: implement --pimcore_project_root argument
The last step to allow a PIMCORE installation from vendor is a dynamic `PIMCORE_PROJECT_ROOT` env. I also played a little bit with the dotenv component - which is awesome, much cleaner and more state of the art than all those constants - but it doesn't solves the real problem. :) It always leads to the same conclusion: Like the `environment var` the startup process also needs the `project root path` to get pimcore running.  

## Solution 1
Add this wonderful piece of code to your `/bin/console` file. (Ok, It's more a piece of a messy crime but i couldn't found a better solution since this must happen in a very early state)

```php
foreach($argv as $i => $arg) {
    if(strpos($arg, '--pimcore_project_root') !== FALSE) {
        $argOpts = explode('=', $arg);
        if(isset($argOpts[1]) && !defined('PIMCORE_PROJECT_ROOT')) {
            define('PIMCORE_PROJECT_ROOT', $argOpts[1]);
            unset($_SERVER['argv'][$i]);
        }
        break;
    }
}
```
Now it's possible to call the symfony console like this:

```bash
 php bin/console cache:clear --no-warmup --pimcore_project_root=/var/www
```

if you add this argument to the [assetInstaller](https://github.com/pimcore/pimcore/blob/6544bc7081ceafcf14222cb7027126a44cb754b2/pimcore/lib/Pimcore/Tool/AssetsInstaller.php#L95) (also [here](https://github.com/pimcore/pimcore/blob/65ca10f177b6da3fdffbc1300e05e61bf58b4880/pimcore/lib/Pimcore/Web2Print/Processor.php#L72) and [here](https://github.com/pimcore/pimcore/blob/65ca10f177b6da3fdffbc1300e05e61bf58b4880/pimcore/lib/Pimcore/Tool/Console.php#L252)), all the internal commands run smoothly no matter where PIMCORE is coming from.

```php
 $builder = new ProcessBuilder([
     'assets:install',
     'web',
     '--env=' . $this->kernel->getEnvironment(),
     '--pimcore_project_root=' . PIMCORE_PROJECT_ROOT
 ]);
```
## Solution 2

Not that pretty: Set the`PIMCORE_PROJECT_ROOT` as a export argument. Of course this must be implemented to all the PIMCORE internal command executions like i told above.

```bash
 PIMCORE_PROJECT_ROOT=/var/www php bin/console cache:clear --no-warmup
```

solverat:
This PR will resolve #1681 without any "hacking". Adding `PIMCORE_PROJECT_ROOT` to the pimcore processes and commands will solve conflicts when overriding the `PIMCORE_PROJECT_ROOT` constant since this path gets lost after execution.

### DIFF
--- a/pimcore/lib/Pimcore/Tool/AssetsInstaller.php
+++ b/pimcore/lib/Pimcore/Tool/AssetsInstaller.php
@@ -96,6 +96,7 @@ class AssetsInstaller
         ]);
 
         $builder
+            ->addEnvironmentVariables(['PIMCORE_PROJECT_ROOT' => PIMCORE_PROJECT_ROOT])
             ->setWorkingDirectory(PIMCORE_PROJECT_ROOT)
             ->setPrefix('bin/console');
 
--- a/pimcore/lib/Pimcore/Tool/Console.php
+++ b/pimcore/lib/Pimcore/Tool/Console.php
@@ -246,7 +246,7 @@ class Console
     {
         $phpCli = self::getPhpCli();
 
-        $cmd = $phpCli . ' ' . $script;
+        $cmd = 'PIMCORE_PROJECT_ROOT=' . PIMCORE_PROJECT_ROOT . ' ' . $phpCli . ' ' . $script;
 
         if (Config::getEnvironment()) {
             $cmd .= ' --env=' . Config::getEnvironment();
--- a/pimcore/lib/Pimcore/Web2Print/Processor.php
+++ b/pimcore/lib/Pimcore/Web2Print/Processor.php
@@ -72,7 +72,9 @@ abstract class Processor
             $args[] = '--env=' . $env;
         }
 
-        $cmd = Tool\Console::getPhpCli() . ' ' . realpath(PIMCORE_PROJECT_ROOT . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'console'). ' web2print:pdf-creation ' . implode(' ', $args);
+        $phpCli = Tool\Console::getPhpCli();
+        $cmd = 'PIMCORE_PROJECT_ROOT=' . PIMCORE_PROJECT_ROOT . ' ' . $phpCli . ' ';
+        $cmd .= realpath(PIMCORE_PROJECT_ROOT . DIRECTORY_SEPARATOR . 'bin' . DIRECTORY_SEPARATOR . 'console'). ' web2print:pdf-creation ' . implode(' ', $args);
 
         Logger::info($cmd);
 


brusch:
I'm afraid that this isn't complete, what about: 
https://github.com/pimcore/pimcore/blob/fa8298c6caabc3496d17a3b25107d4ad1e10bc9d/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/NewsletterController.php#L323
There might be some other remaining places :| 

Why does Dotenv doesn't solve the problem. 
Assuming that `.env` has to be located at `PIMCORE_PROJECT_ROOT`, it should find the right `.env` when eg. calling `PIMCORE_PROJECT_ROOT . '/bin/console' , shouldn't it? 

I also would prefer the Dotenv solution if possible. 
It's on our todo list to use the Dotenv component for certain things like to populate the `PIMCORE_*` constants.

solverat:
hey @brusch, thanks for your feedback.

## Current Solution
It **is** complete.`runPhpScriptInBackground` triggers the`buildPhpScriptCmd` method which will include the project root path. So this commit will work.

## dotenv Solution
I think it can be done, here's some input. I think at the end we only need one single and new constant (`PIMCORE_PROJECT_ROOT`).

### .env file in document root
```env
//.env example
PIMCORE_PROJECT_ROOT=/var/www
PIMCORE_COMPOSER_PATH=${PIMCORE_PROJECT_ROOT}/vendor
PIMCORE_COMPOSER_FILE_PATH=${PIMCORE_PROJECT_ROOT}
PIMCORE_APP_ROOT=${PIMCORE_PROJECT_ROOT}/app
PIMCORE_WEB_ROOT=${PIMCORE_PROJECT_ROOT}/web
PIMCORE_PATH=${PIMCORE_PROJECT_ROOT}/pimcore
PIMCORE_PRIVATE_VAR=${PIMCORE_PROJECT_ROOT}/var
PIMCORE_PUBLIC_VAR=${PIMCORE_WEB_ROOT}/var
PIMCORE_CLASS_DIRECTORY=${PIMCORE_PRIVATE_VAR}/classes
...
```

### /web/app.php
This **must** happen in the `app.php` file, it's the only one which is always in the right project root context.
```php
<?php
//add this to define the ultimate, real and true document root.
define('PIMCORE_DOCUMENT_ROOT', realpath(__DIR__ . '/..'));

require_once PIMCORE_DOCUMENT_ROOT . '/pimcore/config/bootstrap.php';

```

### /bin/console
This **must** happen in the `bin/console` file, it's the only one which is always in the right project root context.
```php
<?php
//add this to define the ultimate, real and true document root.
define('PIMCORE_DOCUMENT_ROOT', realpath(__DIR__ . '/..'));

//add a custom one?
require_once PIMCORE_DOCUMENT_ROOT . '/pimcore/config/bootstrap.php';

```
### /pimcore/config/bootstrap.php

```php
<?php

use Doctrine\Common\Annotations\AnnotationReader;
use Doctrine\Common\Annotations\AnnotationRegistry;
use Symfony\Component\Dotenv\Dotenv;

error_reporting(E_ALL & ~E_NOTICE & ~E_STRICT);

 // deprecated or set all constants from .env?
require_once __DIR__ . '/constants.php';

// deprecated: autoloader already loaded in bootstrap.
//require_once __DIR__ . '/autoload.php'; 

/** @var $loader \Composer\Autoload\ClassLoader */
$loader = require PIMCORE_DOCUMENT_ROOT . '/vendor/autoload.php';

$dotenv = new Dotenv();
$dotenv->load(PIMCORE_DOCUMENT_ROOT .'/.env'); //environments like .env.dev?

$loader->addPsr4('Pimcore\\Model\\Object\\', getenv('PIMCORE_CLASS_DIRECTORY') . '/Object');

//[...]
// that's it. 
// https://www.youtube.com/watch?v=sKZN115n6MI
```
## Some Thoughts
- remove the `PIMCORE_PATH` constant.  Pimcore defines the right pimcore folder, so the decision [already has been made. ](https://github.com/pimcore/pimcore/blob/master/web/app.php#L18):)
- load bootstrap in app and console?

brusch:
Thanks for your thoughts! 
I assume `PIMCORE_DOCUMENT_ROOT ` should be  `PIMCORE_PROJECT_ROOT` in your code examples, right?

solverat:
well actually i was thinking it has to be a new one because`PIMCORE_DOCUMENT_ROOT` is a constant and not a env variable and it maybe could get lost. But it's also available in every call (app.php/console) so it also could be `PIMCORE_PROJECT_ROOT`, right.

solverat:
hey @brusch, since it's been a while - could you give us some feedback about the status quo of this pr? thanks!

brusch:
See also #1795

maff:
@solverat Could you provide a minimum project setup with Pimcore installed via vendor? Would be awesome if we'd have something to test against.

solverat:
hey @maff, here we go: [pimcore_vendor.zip](https://github.com/pimcore/pimcore/files/1280752/pimcore_vendor.zip)

That's our pimcore kickstarter project. But since we can't set the PIMCORE_PROJECT_ROOT, it won't work out of the box.

In our docker-compose.yml for example, we simply added:

```yaml
    environment:
      - PIMCORE_PROJECT_ROOT=/var/www/public
      - PIMCORE_ENVIRONMENT=dev
```
With that, everything works fine locally.

Otherwise you need to add this (which we have to do in every hosting environment): 

```cmd
$ cd /var/www/public
$ PIMCORE_PROJECT_ROOT=/var/www/public bin/console list --env=staging //same with other commands
```

As you can see in this PR, you still can't:
- update pimcore using the web installer
- activate extensions without any errors (assets symlinking)
- use pimcore functions which relies on php scripts (getting triggered internally).

maff:
@solverat thanks for sharing your environment. Could you have a look at the changes done on the `dotenv-constant-bootstrap` branch and validate that against your setup? Basically that sets the initial constant `PIMCORE_PROJECT_ROOT` from entry points such as `app.php` (as you suggested for `PIMCORE_DOCUMENT_ROOT`) and resolves other constants based on that value. Plus, it adds support to define Pimcore's constants as env vars and loading them via dotenv. That should make adding the project root to every command obsolete.

solverat:
hey @maff! i just tried it - and it works!! 🎉 
i couldn't check the web updater, but i think that will work also.

thanks!

solverat:
also: within the introduction of dotenv: do we still need the `constants.php` file?

maff:
@solverat thanks for testing :) See #1961 which will be merged soon.

Although DotEnv and env vars are now used to populate Pimcore's constants and we need to keep the custom `constants.php` file in `/app` for BC. There may also be use cases where you need to resolve constants via code.

maff:
Closed in favor of #1961.

Issue #1750, opened 2017-07-24
[PIMCORE5] Pimcore Bundle Registering in AppKernel


solverat:
## Feature
Register a Pimcore Bundle in AppKernel as a custom one without colliding.

## Possible Solution
1. Change Bundle Order [here](https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Kernel.php#L151): Switch L148 with L151 and skip activated bundles in the later pimcore bundle iteration.

2. Grey out the bundle in the extension manager, set it automatically to active and disallow any interaction with it:
![bildschirmfoto 2017-07-24 um 08 52 31](https://user-images.githubusercontent.com/700119/28511421-78895912-704d-11e7-8818-25c79e337502.png)

If you also think this is a good idea we could provide a PR for that. :)

dpfaffenbauer:
The basic idea is to allow other Bundles to have a dependency on Pimcore-Bundles. With the proposed solution, it's not possible anymore to deactivate this dependency within the extension-manager and therefore break the application.

As @solverat already proposed, the extension-manager needs to check if the bundle has already been added to the kernel manually (inside the AppKernel) and disable the GUI for actions (enable, disable, update, install, etc.). 

It also needs to check if the bundle has already been added to the Kernel and doesn't enable it twice. This could happen if someone installs the Pimcore-Bundle first, and then installs another bundle, which enables the Pimcore-Bundle again.

Our basic idea is to have dachcom\members-bundle as dependency for CoreShop. So, if someone installs Members-Bundle, activates it in the Extension-Manager, installs CoreShop after that, and runs CoreShops install routine, the Bundle would be added twice and breaks Pimcore. 

The other side around would cause a similar Problem. Install CoreShop first (which already activates MembersBundle) and activate MembersBundle in the ExtensionManager.

Another possible idea would be to not display manually added bundles at all.

PR #1778, opened 2017-07-31
Installer improvements

### LINKED ISSUE #1420: [Installer] add way to integrate additional repositories - like demo-ecommerce
important aspects
- download additional installation profiles from github repositories into /install-profiles/xxx
- list additional install-profiles in install.php
- add possibility to run composer update in order to add additional dependencies if necessary.

maff:
Resolves #1420 

* reusable installer can be used from `install.php` and a CLI command `bin/install`
* install profiles can be resolved from `vendor/` and from `install-profiles/` and define what is installed through a `manifest.yml`
* CLI command adds a switch to symlink profile files instead of copying them (useful for dev environments)
* install data (credendials, ...) for CLI installer can be set via env variables

Example CLI usage:

    $ PIMCORE_ENVIRONMENT=dev bin/install --symlink --profile demo-ecommerce --admin-username admin --admin-password admin --mysql-username homestead --mysql-password secret --mysql-database pimcore5 --ignore-existing-config --no-interaction


dpfaffenbauer:
This is amazing! Thanks very much, makes it very easy for automated installs with custom "bootstrap" design.

One question: What do think about installing pimcore-bundles with install-profiles as well? I mean, installing the demo-ecommerce profile is not enough, Somebody (or something) still needs to enable/install the corresponding Pimcore-Bundle as well.

brusch:
Well, it would be more *enabling* bundles than installing - because installing should be done using composer?

dpfaffenbauer:
With installing I mean calling the Pimcore\Extension\Bundle\Installer\InstallerInterface::install method of the PimcoreBundle.

brusch:
I guess we could add another config option `bundles` to the manifest, yes.

dpfaffenbauer:
related to that, I created a issue for dependent symfony-bundles, which will be a case for bigger plugins (like CoreShop). And I am sure, CoreShop will not be the only plugin which will depend on certain symfony-bundles. (https://github.com/pimcore/pimcore/issues/1780)

maff:
@dpfaffenbauer Good idea! We'll merge this change to have the basic installer infrastructure in place and will have a look at the dependency resolving logic (also inside extension manager) in a later step. I'll still need to test all the implications of calling a/multiple bundle installers simultaneously, especially when coming from "outside" (a Pimcore kernel being booted later) as in this situation. 

Before bringing this to the installer level I think we should sort out #1750 and #1780 for already installed instances to see how this can be integrated here.

For the installer, this would add:

* Installer checks if a bundle is available (installed via composer) before installing a profile (easy)
* Installer applies the same logic as the extension manager when auto-enabling and installing a bundle + handling of dependent bundles

PR #1779, opened 2017-07-31
[ADMIN][SESSION] Drop multiple session handling / handle admin session in same session as frontend

### LINKED ISSUE #1671: [SESSION] Frontend session/admin session overwrite each other
If a frontend session is opened in the admin interface, the cookie for the frontend session will be set to the admin session ID, resulting in `pimcore_admin_sid` and `PHPSESSID` being the same and overwriting each other.

* Activate the ecommerce framework
* `/admin/ecommerceframework/config/js-config` will open the frontend session, resulting in both session IDs being the same and frontend session data being lost

maff:
Resolves #1671


maff:
Follow-up:

* 838f69d1df045925cf8ec1e5859ed17738e87563
* 72061dac903191619f5c948a3487091d230935a8

PR #1781, opened 2017-07-31
Generate parameters.yml file only if it doesn't exist.

IgorBenko:
We use a deployment system where we checkout the project (including parameters.yml) and then run Pimcore Installation and load data from Fixtures. This PR makes sure that the parameters.yml doesn't get overwritten or recreated with different 'secret' parameter which would stage a VCS change on the server after deployment.


maff:
Thanks!

PR #1791, opened 2017-08-04
[PIMCORE 4] Web2print events

ckogler-elements:
Added events (document.print.processor.modifyProcessingOptions + document.print.processor.modifyConfig) to modify the processing options of the Web2Print adapter.


brusch:
@ckogler-elements can you provide a PR also for Pimcore 5? Would be great :)

PR #1797, opened 2017-08-07
Ecommerce | Pricing rule optimisations

andreas-gruenwald:
- Improved serialization of gift item objects in pricing rules.
- Allow assignment of object variants in pricing rules (gift items)


fashxp:
thx very much

PR #1806, opened 2017-08-11
[Adminer] Issue with FastCGI (duplicate-headers)

dpfaffenbauer:
This issue fixes local Problems (MAMP) I have with Adminer. MAMP uses fcgi, and fcgi throws an exception when multiple of the same HTTP Headers come. The issue is, that adminer sets HTTP-Headers using the header() function, and symfony sets them via the response object. Both of them set the Content-Type header and fcgi throws an error therefore. 

PS: It seems this fixes styling issues as well, but I am not sure, would be nice if somebody could confirm that.

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/External/AdminerController.php
+++ b/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/External/AdminerController.php
@@ -42,6 +42,8 @@ namespace Pimcore\Bundle\AdminBundle\Controller\Admin\External {
                 return $this->redirect('/admin/external_adminer/adminer?username=' . $conf->username . '&db=' . $conf->dbname);
             }
 
+            $response = new Response();
+
             chdir($this->adminerHome . 'adminer');
 
             ob_start();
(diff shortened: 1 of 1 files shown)


pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/External/AdminerController.php
```diff
@@ -86,12 +87,10 @@ public function proxyAction(Request $request)
                         $content .= file_get_contents($this->adminerHome . 'designs/konya/adminer.css');
                         $content .= file_get_contents(PIMCORE_WEB_ROOT . '/pimcore/static6/css/adminer-modifications.css');
                     }
-
-                    $response->setContent($content);
                 }
             }
 
-            return $response;
+            return new Response($content, 200, array_merge($headers, $this->getAdminerSetHeaders()));
```
maff: Wouldn't `$response->headers->set($header, $value)` do the same as the `replace` argument defaults to true? Would be nicer than falling back to the array syntax IMHO. Maybe we could add an additional method which applies the headers from `getAdminerSetHeaders()` to a response object and use that method from both actions.

brusch:
Finally :)

dpfaffenbauer:
@maff I've updated the PR. There is now a mergeAdminerHeader function which takes the Response Object as parameter, and sets the headers.

@brusch Yeah, it's a must have feature for me, it so convenient to access the database directly within pimcore.

maff:
Thanks! :)

PR #1807, opened 2017-08-11
[AdminBundle] fix translations export Content-Type

dpfaffenbauer:
Content-Type is set with a ending ':', therefore fastcgi reports duplicate-header exception.


maff:
Thx!

PR #1817, opened 2017-08-14
Mark non-static methods as static

d-gerken:
## Fixes Issue #
1. IDEs aren't able to autocomplete static getBy-calls on an object. 
2. If you extend a class and a getBy method call should be overridden, the IDE will throw an error like 'Cannot make non-static method ... static"

- marked non-static method calls in doc-block as static methods.

## Additional info  
All 'getBy' calls are realized by "__callStatic" in Pimcore\Model\Object\Concrete. So methods should be marked as static.


brusch:
Thanks!

PR #1825, opened 2017-08-17
Only allow full inheritance of blocks - not single fields in localized fields

ckogler-elements:
Only allow full inheritance of blocks - not single fields in localized fields


weisswurstkanone:
build 4097

Issue #1845, opened 2017-08-21
Translate key with empty translation returns empty value, not key anymore


andreas-gruenwald:
# Bug Report
When Pimcore creates new translations automatically, the texts are deleted as soon as a refresh occurs.
 It used to work for the previous Pimcore versions, and the bug probably relates to one of the latest updates.

### Steps to reproduce  
1) Add a new translation key in a view with $this->t("my-translation-key")
2) Press refresh => a translation is created.
3) Press refresh again => the result of $this->t("my-translation-key") is an empty string.

solverat:
i can confirm this. since this commit (https://github.com/pimcore/pimcore/commit/ea53b3e470d68aa5365665a6cf94bd4ffe50798f), 
backend translations are empty which leads to empty / not working area bricks...

solverat:
btw. nothing gets deleted, it's "just" a empty translation key return.

brusch:
@maff 

Adding 

```php
if ($id && !$translated) {
    return $id;
}
```

At the end of 
`Translator::checkForEmptyTranslation` 
should fix it.

brusch:
So befor the last line `return $translated;` ;-)

fashxp:
I'll have a look at it today

andreas-gruenwald:
I tested it, it works.

Issue #1837, opened 2017-08-18
[Pimcore 5] Area Brick names are not shown in toolbar and buttons


brusch:
![image](https://user-images.githubusercontent.com/142037/29454594-0912a172-840e-11e7-954f-9a717440b2bb.png)
![image](https://user-images.githubusercontent.com/142037/29454602-11505b0e-840e-11e7-8941-c82819b3b3ca.png)

maff:
Seems to be caused by https://github.com/pimcore/pimcore/issues/1845#issuecomment-324101148.

PR #1855, opened 2017-08-23
fix [Pimcore 5] Area Brick names are not shown in toolbar and buttons…

js-aw:
#1837 #1845


brusch:
Thanks, but the fallback shouldn't be used for admin translations, since they can have completely different locales.

fashxp:
I'll have a look at it today

fashxp:
fixed it in a bit different way ... please have a look and verify

js-aw:
thx!

PR #1856, opened 2017-08-23
Update Translator.php

brusch:
#1855 #1837 #1845


brusch:
@fashxp this should fix the referenced issues

fashxp:
fixed it a bit differently

PR #1862, opened 2017-08-24
redirect method returns a Redirect response so the previewAction must…

IgorBenko:
The redirect method returns a Redirect response so the previewAction must also return it.


fashxp:
thx

PR #1867, opened 2017-08-25
[Twig] change form Twig_Simple to Twig_ functions

dpfaffenbauer:
since Pimcore uses Twig 2.*, we shouldn't use deprecated Twig functions. (https://twig.symfony.com/doc/1.x/deprecated.html)

In fact, all of the Twig_Simple functions are just aliases for Twig_*.

PS: The deprecation notes are very confusing. As of Twig 1.* the Twig_Function, Twig_Test etc, are deprecated, but as of Twig 2.*, the Twig_SimpleFunction, Twig_SimpleTests are deprecated.


maff:
Thanks! Rebased and merged manually in b8de4b7ece2fffbec81a863187f0e768aa91c7c2.

PR #1870, opened 2017-08-27
convert invalid menu id prefix backslash to dash.

solverat:
instead of `<a id="pimcore\navigation\renderer\menu-23">` => `<a id="pimcore-navigation-renderer-menu-23">`


brusch:
Makes sense ;-) 
Thanks!

PR #1872, opened 2017-08-28
[AdminBundle] fix qtip config for documents with site

dpfaffenbauer:
raised a warning cause variable $site was undefined on line 1079.


brusch:
Thx

Issue #1818, opened 2017-08-14
Invalid values stored into local storage key "pimcore_opentabs"


deltafactory:
When restoring tabs, I received a failed XHR when the requested ID was cast to 0. The usage log contained this:
```
2|Pimcore\Bundle\AdminBundle\Controller\Admin\AssetController::getDataByIdAction|pimcore_admin_asset_getdatabyid|[]|{"id":"0","type":"image"}
```

Tracing the source of the error, I found `pimcore.helpers.openMemorizedTabs()` and the call to localStorage. The `pimcore_opentabs` key contained this, among other correct values: "asset_undefined_image"

I do not have steps to recreate at this time. Checking for valid a ID during the call to `openAsset()` before making the XHR call could prevent the resulting error.

IgorBenko:
This happens when the asset/object/document was removed while its tab was still opened in the UI. So it happens when the UI doesn't know anything about removal of the element, but the tab was left open. Issuing `localStorage.clear();` resolves this temporarily.

deltafactory:
At what point are the tabs written to localStorage? I would think it saved a valid ID that no longer existed, not an invalid value. In any case, it seems there's a few possible ways to resolve this:
1. Filter out entries that are "undefined" when saving tabs to localStorage.
2. Use the UI's event model to announce the deletion of the asset, if/when done in the same browser tab. Changes via the PHP API are still a factor, so ...
3. Gracefully handle ID:0 and any ID that returns an empty record/object. Close the affected tab or populate with a message indicating it's no longer available.

At some point I would be happy to contribute pull requests instead of just reporting bugs. :)

brusch:
IDs are written to localStorage at the point they are opened, therefore option 1 is obsolete, because at the time they are added to the local storage they are valid, so filtering makes no sense at all. 
Option 2 is not better either: User A opens Object X User B deletes Object X, User A reloads the admin UI -> same problem

3. would be the only possible way of handling this issue. 
An option `ignoreNotFoundError` for `pimcore.helpers.openAsset/openDocument/openObject()` similar to `ignoreForHistory` would be a quick win imho.

IgorBenko:
@brusch Your suggestion sounds great!

deltafactory:
If I can recreate the condition that resulted in the undefined ID, I will submit a separate ticket.

weisswurstkanone:
build #100

PR #1873, opened 2017-08-28
[AdminBundle] better handling of Object and Object\Folder not found

dpfaffenbauer:
This PR adds some better handling for Objects and Object\Folders which don't exist anymore. Actually, for Folders this should never happen, but since Objects are remembered in local storage, this causes an exception very often. Therefore I created this PR to solve this issue. It checks instanceof Object (or Folder), if good, same behaviour as before, if not: Respond with success: false and a translateable message to the client "element_not_found". The client checks for the success-property and shows a message-box if not found. Cause of this behaviour, the remembered tab gets removed as well.


weisswurstkanone:
I guess this one can be closed in favor of https://github.com/pimcore/pimcore/issues/1818

dpfaffenbauer:
I guess so as well 👍

weisswurstkanone:
Many thanks!!! ;-)

PR #1878, opened 2017-08-30
[Twig] add pimcore_site function

dpfaffenbauer:
In Multi-Site context, its often necessary to get the site to get its root-document and for example include snippets from its root document.


pimcore/lib/Pimcore/Twig/Extension/PimcoreObjectExtension.php
```diff
@@ -28,6 +29,7 @@ public function getFunctions()
         // simple object access functions in case documents/assets/objects need to be loaded directly in the template
         return [
             new \Twig_SimpleFunction('pimcore_document', [Document::class, 'getById']),
+            new \Twig_SimpleFunction('pimcore_site', [Site::class, 'getById']),
```
maff: As you brought it up: could you change this to `Twig_Function` please? ;)

dpfaffenbauer:
Yeah, I will ;), but since the other PR wasn't merged yet, I kept this with the deprecated function.

PR #1899, opened 2017-09-03
Improve classes-rebuild command

jbq:
Do not dump definition file for object bricks and field collections when
running bin/console classes-rebuild, mimic what is done for classes

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/CoreBundle/Command/ClassesRebuildCommand.php
+++ b/pimcore/lib/Pimcore/Bundle/CoreBundle/Command/ClassesRebuildCommand.php
@@ -96,7 +96,7 @@ class ClassesRebuildCommand extends AbstractCommand
                 $output->writeln($brickDefinition->getKey() . ' saved');
             }
 
-            $brickDefinition->save();
+            $brickDefinition->save(false);
         }
 
         if ($output->isVerbose()) {
@@ -110,7 +110,7 @@ class ClassesRebuildCommand extends AbstractCommand
                 $output->writeln($fc->getKey() . ' saved');
             }
 
-            $fc->save();
+            $fc->save(false);
         }
     }
 }
--- a/pimcore/models/DataObject/Fieldcollection/Definition.php
+++ b/pimcore/models/DataObject/Fieldcollection/Definition.php
@@ -247,7 +247,7 @@ class Definition extends Model\AbstractModel
     /**
      * @throws \Exception
      */
-    public function save()
+    public function save($saveDefinitionFile = true)
     {
         if (!$this->getKey()) {
             throw new \Exception('A field-collection needs a key to be saved!');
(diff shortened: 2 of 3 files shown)


jbq:
@brusch can you please review?  Thanks!

brusch:
@jbq the definition of classes and bricks/field collections always have to be 100% in sync, so it makes definitely sense to save them as well after all classes were saved. 
Any particular problem with that in your case?

jbq:
I think you're misunderstanding this pull request.  Of course bricks and field collections must be saved when the class is saved, observe the default parameter of `public function save($saveDefinitionFile = true)` it is true by default.

Here we are talking about the `classes-rebuild` administrative command, that is not suppose to overwrite the definition files, but make sure the database tables are in sync with the definition files.  Thus we passe `save(false)` in this case, and in this case only.  It was done for classes but not for object bricks and field collections.

Can you please reopen?  Thanks!

dpfaffenbauer:
But what's the point? When you rebuild the classes, the db gets synced and class definitions get written the same as they have been before?

jbq:
When using classes-rebuild he class definitions are not written again, it is the purpose of the `$saveDefinitionFile` parameter set to `false`.  The purpose is to alter classes in a development environment, put `var/classes` into version control, and finally run the `classes-rebuild` in production.  It works great!  This patch is a slight fix to avoid objet bricks and field collections definition files to be written again to disk because it causes new changes detected in the version control system.

brusch:
I think this 'fix' can cause more problems than it helps ... what if a field collection or an object brick changed? What if the new class definition is incompatible with the existing field collections or object bricks? 
I understand your intention, but this fix covers only your particular use case and is not generic enough. 

A generic approach would be to offer an additional flag, to disable the rebuild for field collections and object bricks. But imho changes on the filesystem should be reviewed anyway, it's up to you to update or revert it in the VCS. 

Other opinions are as always welcome :)

fashxp:
maybe provide an additional command that updates database structure only?

dpfaffenbauer:
@fashxp not sure if thats a good idea, I mean, rebuild-classes syncs the db with the class-definition files, which is fine. (+ it re-creates classes) adding a second command which only syncs database could cause confusion. I'd rather stick with a parameter to restrict the deployment command to only sync db.

jbq:
@brusch you introduced the `classes-rebuild` feature in january 2016, and a commit from Kamil Karkus was merged on march 10th 2017 to introduce the extra `$saveDefinitionFile` parameter, and set it to `false` during `classes-rebuild`.  See https://github.com/pimcore/pimcore/commit/5744a56f915be1329d2cbfc0ed66cf5dafae7522

I'm not discussing the approach, just pointing out that there is currently a half-baked hack in place, in the sense that only for class definition there is no writes on disk, but not for field collections and object bricks.

jbq:
It worries me to see this issue closed, without any consensus.  Would you mind reopening and let more time for discussion?  Thanks :-)

PR #1907, opened 2017-09-05
[Twig] add pimcore_site function

dpfaffenbauer:
Kinda messed up the merge for PR #1878, so here is a new one.

In Multi-Site context, its often necessary to get the site to get its root-document and for example include snippets from its root document.


maff:
thx :)

PR #1920, opened 2017-09-06
Update 11_Event_API_and_Event_Manager.md

daniel-travix:
Since Commit: 57044e5 it has to be 
- { name: kernel.event_listener, event: pimcore.dataobject.preUpdate, method: onPreUpdate }
instead of
- { name: kernel.event_listener, event: pimcore.object.preUpdate, method: onPreUpdate }


weisswurstkanone:
thx!

PR #1921, opened 2017-09-07
Rename AdminBundle/Resources/views/Admin/Object/

daniel-travix:
**Rename:** 

> pimcore/lib/Pimcore/Bundle/AdminBundle/Resources/views/Admin/Object/ 

**to**

> pimcore/lib/Pimcore/Bundle/AdminBundle/Resources/views/Admin/DataObject/

**to fix**

rename


weisswurstkanone:
many thanks! build 106

PR #1929, opened 2017-09-08
Namespace consolidation: Move classes and define services in Symfony 3.3 FQCN format 

### LINKED ISSUE #1629: Consolidation of PHP Namespaces, Service-Naming/Visibility

maff:
Resolves #1629 

* Move classes to better namespaces as defined in #1629. Especially, this gets rid of the `Pimcore\Service` namespace and moves classes to more appropriate locations. A mapping autoloader aliases classes to their new names and emits a `E_USER_DEPRECATED` warning of one of the old class names is used.
* Change Pimcore service definitions to Symfony 3.3+ fully qualified class name format where applicable and use autowiring where possible. This allows implementations to use autowiring for most of Pimcore's core services. For backwards compatibility, aliases map the previous service IDs to the new FQCN names.
* Make services private where applicable.
* Define admin controllers as services to make use of autowiring and action injection. A few quick wins in admin controllers were updated to use autowiring as example (e.g. extension manager, renderlet), the rest can be migrated gradually when needed.


maff:
@dpfaffenbauer @solverat (and anybody else already using Pimcore5): could you check this against your implementations if this introduces errors (missing aliases, private services expected being public, ...) to existing installations/third-party code?

dpfaffenbauer:
@maff sure

@aarongerig could you check with your existing projects as well?

solverat:
### DocumentTagExtension
**error:** `Parent definition "pimcore.twig.extension.document_tag" does not exist.`

Replaced  
`parent: '@pimcore.twig.extension.document_tag'` 
with
`parent: 'Pimcore\Twig\Extension\DocumentTagExtension'`

### AbstractFrontendListener

 **error:** `Missing pimcore context resolver. Is the listener properly configured?`

1. Enabled autowiring
```yml
services:
    _defaults:
        autowire: true
        autoconfigure: true
        public: false
```
2. Replaced `AbstractFrontendListener` with the`PimcoreContextAwareTrait` trait.

maff:
@solverat thx for the feedback!

1. DocumentTagExtension: added aliases for twig extensions - now both versions should work.
2. ~~`AbstractFrontendListener`~~ now implements ~~`PimcoreContextResolverAwareInterface`~~ again, so the resolver should automatically be set even without autowiring enabled.

solverat:
well done. looks good to me!

dpfaffenbauer:
First test with CoreShop looks good. UnitTests are running thru. Only thing that broke it:

 - Pimcore\Service\Locale, instance of Pimcore\Localization\Locale given
 
but that seems fine, since Pimcore 5 is not released yet, I wouldn't consider it a break.

maff:
@dpfaffenbauer thx for the feedback. Actually that should be handled by the LegacyMapping autoloader - is that autoloader active in your test env?

dpfaffenbauer:
@maff LegacyMapping::load doesn't get triggered for Pimcore\Service\Locale, I am not sure though, but I think its because I load it as dependency into another service and the constructor expects Service\Locale but gets Localization\Locale.

maff:
@dpfaffenbauer seems to be an autoloader timing issue. The type hint does not trigger autoloading and although the instance can be created successfully and is an instance of the aliased class, the type hint does not match. That's now handled by adding aliases directly instead when the class should be autoloaded.

dpfaffenbauer:
Just tried to update to latest version using the Build in Updater, following issue occurred:
 
***Unable to replace alias "cmf_routing.route_provider" with actual definition "pimcore.routing.dynamic.route_provider".***
Now: Pimcore\Routing\DynamicRouteProvider

Just posting this here for other people who have the same issue.

PR #1946, opened 2017-09-12
Add title to search / selector window

BlackbitNeueMedien:
Before this change the search modal has no title set.


weisswurstkanone:
Hi, which one?
Also see https://github.com/pimcore/pimcore/commit/e2425231a6bd6c438d63335c0a4102f545d06a53

BlackbitNeueMedien:
Ok, should be solved in e242523

Issue #1934, opened 2017-09-10
Install Pimcore5: check requirements fatal error


Yivan:
In the process of installing Pimcore 5, on the install process, using the "check requirement" button display a fatal error at the bottom of the popin, please see:

![pim5-install-error](https://user-images.githubusercontent.com/9025839/30250220-c2e97d98-964a-11e7-81bd-381f367a84c8.png)

It seems it try to connect as root, why ?

Config provided is this one (user and bdd have been created and we checked by connection on phpMyAdmin, credential are working fine with all privileges for this user):

![install-config](https://user-images.githubusercontent.com/9025839/30250254-5c41c45a-964b-11e7-8471-27f23892f100.png)

Yivan:
Note, this is the last Pimcore5 build:
Version: 5.0.0-BETA1
Build: 105

maff:
Fixed in #1953

PR #1953, opened 2017-09-12
[INSTALLER] Web installer improvements/bug fixes

### LINKED ISSUE #1933: Improvement: install Pimcore5
We installed Pimcore 5 on several environment (linux and windows) and all seems ok for installation.
I would like just to suggest imporvement in the installation process.

Just after unzipping the files, and going to http://www.mydomain.ext (thinking it will display the install process...), we got PHP errors:

```
Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in /.../vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php on line 47
Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 65536 bytes) in /.../pimcore/lib/Pimcore/Event/SystemEvents.php on line 17
```

So we go to: https://www.pimcore.org/docs/5.0.0/Getting_Started/Installation.html and see /install.php sould be call.
Running  http://www.mydomain.ext/install.php  work ok!

Improvement suggestion:
If it is a first installation http://www.mydomain.ext should not print fatal error (not good image... we where thinking Pimcore5 is buggy...), but must automatically redirect to http://www.mydomain.ext/install.php.
It will make more easy installation for people and prevent they think it is buggy.

Thanks.

maff:
Resolves #1933 and #1934

* Make installer available via `/install` instead of `/install.php`. This also allows us to use routes other than `/` in installer (e.g. the new check route for additional actions)
* Add the requirements check action as dedicated action on the install controller and don't reuse the controller from PimcoreAdminBundle. Makes both controllers smaller and independent from each other. The install requirements check is now available via `/install/check` and will not load the not yet initialized pimcore environment which resulted in errors (#1934). The admin check is now available via `/admin/install/check` and does not need any special authentication or DB connection loading logic.
* Redirect to `/install` if pimcore is not installed instead of running into errors (#1933)


Yivan:
Great one! Thanks a lot.

Issue #1952, opened 2017-09-12
[Newsletter] Consider loading AddressSourceAdapter from Container


dpfaffenbauer:
Currently Pimcore still loads AddressSourceAdapter from Namespace ```\\Pimcore\\Document\\Newsletter\\AddressSourceAdapter\\```. This should be done using the container.

https://github.com/pimcore/pimcore/blob/1da1bd126caca24cf12e696889aea7da0055ba4c/pimcore/lib/Pimcore/Bundle/CoreBundle/Command/InternalNewsletterDocumentSendCommand.php#L66

If you wish, I could create a PR for that, just wanted to discuss this with you before.

fashxp:
👍 for that one. Maybe similar to https://github.com/pimcore/advanced-object-search/blob/master/src/Service.php#L85 ? 

Could you provide a PR and also add a note here https://github.com/pimcore/pimcore/blob/master/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/02_V4_to_V5.md? 

Thank you very much...

fashxp:
related to https://github.com/pimcore/pimcore/issues/1397

fashxp:
related to https://github.com/pimcore/pimcore/issues/1399

dpfaffenbauer:
#1399 can be done the same way actually. but should be in a separate PR.

fashxp:
yes, bit it is still related :-)

dpfaffenbauer:
solved :)

I'll create a PR for #1399 as well.

fashxp:
Thx!!

PR #1957, opened 2017-09-13
[NewsletterAdapter] change to use configuration and container

dpfaffenbauer:
This relates to issue #1952

It now uses a configuration to register source adapters and a service-locator to retrieve them in controller/command.

```yml
pimcore:
  documents:
    newsletter:
      adapters:
         identifier: service-name
```

Also the AddressSourceAdapterInterface has been changed: renamed __construct to configure to be able to inject services.

### DIFF
--- a/doc/Development_Documentation/03_Documents/02_Document_Types/05_Newsletter_Documents.md
+++ b/doc/Development_Documentation/03_Documents/02_Document_Types/05_Newsletter_Documents.md
@@ -48,14 +48,21 @@ the namespace `pimcore.document.newsletters.addressSourceAdapters`, named like t
 - PHP Class: This class is the server side implementation of the adapter. It is responsible for retrieving and preparing 
 the email addresses. It needs to implement the interface `AddressSourceAdapterInterface`. As sample see
  [csvList](https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapter/CsvList.php).
-- Register your Adapter Source as Service
-- Add your Adapter Source to the configuration:
+- Register your Adapter Source Factory as Service. If you are using a very simple Adapter Source, you can use the DefaultFactory
+  ```yml
+  app.document.newsletter.factory.myAdapter:
+      class: Pimcore\Document\Newsletter\DefaultAddressSourceAdapterFactory
+      arguments:
+        - 'App\Document\Newsletter\AddressSourceAdapter\MyAdapter'
+  ```
+- If you are using a more complex Source Adapter, you can create your own Factory by implementing the interface Pimcore\Document\Newsletter\Factory\AddressSourceAdapterFactoryInterface
+- Add your Adapter Factory to the configuration:
 
 ```yml
 pimcore:
     newsletter:
         adapters:
-            myAdapter: App\Newsletter\Adapter\MyAdapter
+            myAdapter: app.document.newsletter.factory.myAdapter
 
 ````
 
--- a/doc/Development_Documentation/03_Documents/02_Document_Types/05_Newsletter_Documents.md
+++ b/doc/Development_Documentation/03_Documents/02_Document_Types/05_Newsletter_Documents.md
@@ -55,7 +55,7 @@ the email addresses. It needs to implement the interface `AddressSourceAdapterIn
       arguments:
         - 'App\Document\Newsletter\AddressSourceAdapter\MyAdapter'
   ```
-- If you are using a more complex Source Adapter, you can create your own Factory by implementing the interface Pimcore\Document\Newsletter\Factory\AddressSourceAdapterFactoryInterface
+- If you are using a more complex Source Adapter, you can create your own Factory by implementing the interface Pimcore\Document\Newsletter\AddressSourceAdapterFactoryInterface
 - Add your Adapter Factory to the configuration:
 
 ```yml
--- a/doc/Development_Documentation/03_Documents/02_Document_Types/05_Newsletter_Documents.md
+++ b/doc/Development_Documentation/03_Documents/02_Document_Types/05_Newsletter_Documents.md
@@ -46,9 +46,17 @@ It is easily possible to implement custom address source adapter. Following file
 the namespace `pimcore.document.newsletters.addressSourceAdapters`, named like the adapter (e.g. `pimcore.document.newsletters.addressSourceAdapters.myAdapter`)
  and implement the methods `initialize`, `getName`, `getLayout` and `getValues`. As sample see [csvList](https://github.com/pimcore/pimcore/blob/master/pimcore/static6/js/pimcore/document/newsletters/addressSourceAdapters/csvList.js)
 - PHP Class: This class is the server side implementation of the adapter. It is responsible for retrieving and preparing 
-the email addresses. It has to be located in the namespace `Pimcore\Document\Newsletter\AddressSourceAdapter`, named like
-the adapter (e.g. `MyAdapter`) and implement the interface `AddressSourceAdapterInterface`. As sample see
- [csvList](https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapter/CsvList.php). 
+the email addresses. It needs to implement the interface `AddressSourceAdapterInterface`. As sample see
+ [csvList](https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapter/CsvList.php).
+- Add your Adapter Source to the configuration:
+
+```yml
+pimcore:
+    newsletter:
+        adapters:
+            myAdapter: App\Newsletter\Adapter\MyAdapter
+
+````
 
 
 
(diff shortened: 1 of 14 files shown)


dpfaffenbauer:
Oh, just saw: PimcoreCoreExtension changed a bit more than intented cause of automated Code Formatting from PHP-Storm, hope you don't mind?

dpfaffenbauer:
Still not quite happy with the proposed solution. Actually, the ```configure``` function should be inside a factory and prepare the SourceAdapter, the SourceAdapter then gets passed around. so it wouldn't be stateful as it is now.

But, to do that, we would need to introduce a new AddressSourceFactory and add a factory for each type. What do you think about that?

maff:
@dpfaffenbauer thanks for the PR :) I'm afraid the stateful solution imposes some unforseeable risks like state not being reinitialized on configure (e.g. https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapter/DefaultAdapter.php#L66). Although this could be handled by implementations (e.g. resetting state on `configure()`) I think a solution building a new adapter is more favorable. I'd suggest:

* Build adapters through a factory as you suggested
* Implement a configurable default factory which can be configured with a class name to build and just passes the parameters is receives to the adapter constructor (maybe something like [AgentFactory](https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/OrderManager/Order/AgentFactory.php)). Register 3 factory services using this factory class and update the config mapping to point to the factory service instead of the adapter service (which will be removed). More complex adapter implementations can be built by custom factories. 
* Build a locator as your did for the adapters, but containing the factories this time.

WDYT?

dpfaffenbauer:
@maff yeah, thought so too 👍 

Implemented using Factory Pattern now.

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/02_V4_to_V5.md
```diff
@@ -296,3 +296,23 @@ $this->get(\Pimcore\Cache\Runtime::class) // service container
...
+      class: Pimcore\Document\Newsletter\DefaultAddressSourceAdapterFactory
+      arguments:
+        - 'App\Document\Newsletter\AddressSourceAdapter\MyAdapter'
+  ```
+- If you are using a more complex Source Adapter, you can create your own Factory by implementing the interface Pimcore\Document\Newsletter\Factory\AddressSourceAdapterFactoryInterface
+- Add your Adapter Factory to the configuration:
+
+```yml
+pimcore:
+    newsletter:
```
maff: It's currently `pimcore.documents.newsletter` but for me `pimcore.newsletter` sounds better/more concise. Could you update the config + extension accordingly?

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/02_V4_to_V5.md
```diff
@@ -296,3 +296,23 @@ $this->get(\Pimcore\Cache\Runtime::class) // service container
...
+        adapters:
```
maff: `source_adapters` maybe?

pimcore/lib/Pimcore/Bundle/CoreBundle/Command/InternalNewsletterDocumentSendCommand.php
```diff
@@ -63,12 +64,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $addressSourceAdapterName = $data['addressSourceAdapterName'];
         $adapterParams = $data['adapterParams'];
 
-        $adapterClass = '\\Pimcore\\Document\\Newsletter\\AddressSourceAdapter\\' . ucfirst($addressSourceAdapterName);
+        $serviceLocator = $this->getContainer()->get('pimcore.newsletter.address_source_adapter.factories');
+
+        if (!$serviceLocator->has($addressSourceAdapterName)) {
+            //TODO: Shouldn't this throw an exception?
```
maff: I would definitely throw an exception here. Previously a class not found exception was thrown, now we just can use a custom one.

pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/config/pimcore/config.yml
```diff
@@ -150,6 +150,11 @@ pimcore:
             prefixes:
                 - \Pimcore\Model\Document\Tag\
                 - \Document_Tag_
+        newsletter:
+            adapters:
+              defaultAdapter: pimcore.document.newsletter.factory.default
```
maff: indented by 4 please :)

pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapterFactoryInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+interface AddressSourceAdapterFactoryInterface
+{
+    /**
+     * Configures and creates the AddressSourceAdapterInterface
+     *
+     * @param $params
+     * @return AddressSourceAdapterInterface
+     */
+    public function configure($params);
```
maff: I'd prefer a name which indicates that an instance is built here (create, build, ...). `configure()` sounds more like "take an existing instance and add configuration".

dpfaffenbauer:
@maff all requested changes done

pimcore/lib/Pimcore/Bundle/CoreBundle/Command/InternalNewsletterDocumentSendCommand.php
```diff
@@ -16,11 +16,13 @@
 
 use Pimcore\Console\AbstractCommand;
 use Pimcore\Document\Newsletter\AddressSourceAdapterInterface;
+use Pimcore\Document\Newsletter\AddressSourceAdapterFactoryInterface;
 use Pimcore\Logger;
 use Pimcore\Model;
 use Pimcore\Tool\Newsletter;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Intl\Exception\RuntimeException;
```
maff: This was accidentally i suppose?

doc/Development_Documentation/03_Documents/02_Document_Types/05_Newsletter_Documents.md
```diff
@@ -46,9 +46,25 @@ It is easily possible to implement custom address source adapter. Following file
...
+the email addresses. It needs to implement the interface `AddressSourceAdapterInterface`. As sample see
+ [csvList](https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapter/CsvList.php).
+- Register your Adapter Source Factory as Service. If you are using a very simple Adapter Source, you can use the DefaultFactory
+  ```yml
+  app.document.newsletter.factory.myAdapter:
+      class: Pimcore\Document\Newsletter\DefaultAddressSourceAdapterFactory
+      arguments:
+        - 'App\Document\Newsletter\AddressSourceAdapter\MyAdapter'
+  ```
+- If you are using a more complex Source Adapter, you can create your own Factory by implementing the interface Pimcore\Document\Newsletter\Factory\AddressSourceAdapterFactoryInterface
```
maff: `Pimcore\Document\Newsletter\AddressSourceAdapterFactoryInterface`

pimcore/lib/Pimcore/Bundle/CoreBundle/Command/InternalNewsletterDocumentSendCommand.php
dpfaffenbauer: no, thats the exception thrown in line 71. Better idea for a different Exception Type? Maybe ClassNotFound or InvalidArgument?

maff: `\RuntimeException` or `\InvalidArgumentException` but in any case none of Symfony's intl component.

dpfaffenbauer:
alright, resolved again :)

maff:
Thanks a lot :)

dpfaffenbauer:
awesome :) 

🍻

dpfaffenbauer:
@maff Just encountered an issuse in this line: https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Bundle/CoreBundle/DependencyInjection/PimcoreCoreExtension.php#L405

Since the service may not be available at the point of time the service-locator gets defined, this will not work as expected. Solution: Just remove the exception and it should run into runtime issues when service is not available, what do you think?

maff:
@dpfaffenbauer you're right, that check is too early. However, Symfony checks for invalid references in `CheckExceptionOnInvalidReferenceBehaviorPass` so we can just remove that check and rely on Symfony's handling :) Would you provide a PR?

dpfaffenbauer:
alright, PR is coming.

Could you do me another favor pls? Could someone please review PR 1880 before stable release?

Issue #1399, opened 2017-04-05
[CustomReports] add new way to inject custom Data Source Adapters


brusch:
_From @fashxp on March 28, 2017 13:18_

see https://github.com/pimcore/pimcore-5/blob/symfony/doc/Development_Documentation/08_Tools_and_Features/29_Custom_Reports.md#custom-data-source-adapters

_Copied from original issue: pimcore/pimcore-5-issues#62_

maff:
Implemented in #1960

PR #1960, opened 2017-09-13
[CustomReportAdapter] make them available via service-locator

dpfaffenbauer:
related to #1399 

Adds the possibility to load CustomReport Adapters via configuration:

```yml
pimcore:
    custom_report:
        adapters:
            identifier: service_name
```

This PR also creates a Factory class For ReportAddressSourceAdapterFactory to take advantage of DI for Newsletter Address Sources as well.

pimcore/models/Tool/CustomReport/Config.php::156 (getAdapter) is now deprecated, but still works for maybe BC reasons?

There is also a small change in ExtJS for CustomReports, they used to have a body-padding of 10px inside of Pimcore, which is very unusual. Nothing inside the main-tab-panel has a padding of 10px.

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Reports/CustomReportController.php
+++ b/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Reports/CustomReportController.php
@@ -187,7 +187,7 @@ class CustomReportController extends ReportsControllerBase implements EventedCon
         $result = [];
 
         try {
-            $adapter = $this->getReportAdapter($configuration->type, $configuration);
+            $adapter = CustomReport\Config::getAdapter($configuration);
             $columns = $adapter->getColumns($configuration);
             if (!is_array($columns)) {
                 $columns = [];
@@ -273,7 +273,7 @@ class CustomReportController extends ReportsControllerBase implements EventedCon
         $config = CustomReport\Config::getByName($request->get('name'));
         $configuration = $config->getDataSourceConfig();
 
-        $adapter = $this->getReportAdapter($configuration->type, $configuration, $config);
+        $adapter = CustomReport\Config::getAdapter($configuration, $config);
 
         $result = $adapter->getData($filters, $sort, $dir, $offset, $limit, null, $drillDownFilters, $config);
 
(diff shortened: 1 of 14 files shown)


pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Reports/CustomReportController.php
```diff
@@ -187,7 +187,7 @@ public function columnConfigAction(Request $request)
         $result = [];
 
         try {
-            $adapter = CustomReport\Config::getAdapter($configuration);
+            $adapter = $this->getReportAdapter($configuration->type, $configuration);
```
maff: I think we should stick with `CustomReport\Config::getAdapter()` here to make sure the fallback logic for an empty type still works and remove the `getReportAdapter()` method from this controller. We can move the logic in the static method to a custom service at some point but for now i'd just stick to the static method.

dpfaffenbauer: I dont think thats an good idea, the static method uses directly the container to build the adapter, which is a big no-go for me. I just kept it for BC and marked it deprecated.

we could introduce a new factory to create factories :D, but thats kinda weird isn't it? I mean, basically the controller should be responsible to take care about the fallback, shouldn't it? if somebody else uses the service-locator, he should be responsible for fallbacks then. also: configurations should always be complete, otherwise you are producing inconsistent data.

maff: The thing is we can't assume that configurations are complete. I completely agree with your points (basically every `\Pimcore::getContainer()` call is "wrong") and I think that that the loading and fallback should be encapsulated somewhere, but that's something which can be done later. The adapter loading should be consistent - I'd rather stick to the static method for now than duplicating the code in the controller.

For now let's just stick to the static method to have the same logic to load adapters throughout the system and improve that at a later point.

dpfaffenbauer: Alright, at your risk ;) but you owe me a beer for that 🍺  😉 

Will do that later.

maff: Deal ;)

dpfaffenbauer:
pushed requested changes

maff:
Thanks! 🚢

Issue #1882, opened 2017-08-30
Adding possibilty to use GA Pimcore reporting without inject the tag in frontend.


Yivan:
Google analytics can be configured in the marketing settings.

We wanted to add ourselves the GA tags in the front-end website (because of some custom handling of it), but having possibility to have the GA reporting associated in documents in the Pimcore back-end.
Actually, this is not possible.

Solution.
Adding a checkbox to disable injection, like showed on this image:

![ga-inject](https://user-images.githubusercontent.com/9025839/29890194-10ffce60-8dc7-11e7-96d5-1e40336f62e9.png)

This way, developpers can handle themselve the inject of the GA analytics code in the website if needed.

brusch:
I'm just thinking if this really needs another configuration or if this should be better done in the code...
I'd prefer the latter option. 

In the code you can easily deactivate the code injection using the following code: 

Pimcore 4: 
```php
$front = \Zend_Controller_Front::getInstance();
$front->unregisterPlugin('\Pimcore\Controller\Plugin\Analytics');
```

Pimcore 5: 
```php
$gaListener = $container->get('pimcore.event_listener.frontend.google_analytics_code');
$gaListener->disable();
```

brusch:
I'll close this issue for now. 
If anyone else would love to see this feature in Pimcore please react with a +1 :)

Yivan:
Thanks @brusch for your answer, it is a perfect solution. No need to make it UI for us.
Maybe this could be added to the documentation page in marketing settings ? So other people wanting to disable it can read how to do it from the documentation.

PR #1990, opened 2017-09-18
[Pimcore 5] Adding Marketing settings to documentation

Yivan:
This is the pull request for Pimcore 5 for https://github.com/pimcore/pimcore/issues/1882.
As there is no Marketing settings page in documentation I added one here, with a note about desactivate automatic Google Analytics code injection as it can be needed sometimes (to be able to layer some custom debugging code for instance wich need to inject custom code).

This page can be enriched later if neccessary for other marketing settings (they are self explanotory in the GUI for now).

### DIFF
--- a/doc/Development_Documentation/18_Tools_and_Features/28_Marketing_Settings.md
+++ b/doc/Development_Documentation/18_Tools_and_Features/28_Marketing_Settings.md
@@ -10,11 +10,9 @@ The `Marketing Settings` give you the possibility to configure marketing-specifi
 Google Analytics code is automaticaly injected during rendering the page.
 This behaviour can be disabled using:
 
-```
+```php
 <?php
-// either
+// fetch the listener through container or (better) inject it as dependency into your code
 $gaListener = $container->get(\Pimcore\Bundle\CoreBundle\EventListener\Frontend\GoogleAnalyticsCodeListener::class);
-// or
-$gaListener = $container->get('Pimcore\\Bundle\\CoreBundle\\EventListener\\Frontend\\GoogleAnalyticsCodeListener');
 $gaListener->disable();
 ```
--- a/doc/Development_Documentation/18_Tools_and_Features/28_Marketing_Settings.md
+++ b/doc/Development_Documentation/18_Tools_and_Features/28_Marketing_Settings.md
@@ -11,6 +11,10 @@ Google Analytics code is automaticaly injected during rendering the page.
 This behaviour can be disabled using:
 
 ```
-$gaListener = $container->get('pimcore.event_listener.frontend.google_analytics_code');
+<?php
+// either
+$gaListener = $container->get(\Pimcore\Bundle\CoreBundle\EventListener\Frontend\GoogleAnalyticsCodeListener::class);
+// or
+$gaListener = $container->get('Pimcore\\Bundle\\CoreBundle\\EventListener\\Frontend\\GoogleAnalyticsCodeListener');
 $gaListener->disable();
 ```
--- a/doc/Development_Documentation/18_Tools_and_Features/28_Marketing_Settings.md
+++ b/doc/Development_Documentation/18_Tools_and_Features/28_Marketing_Settings.md
@@ -1,0 +1,16 @@
+# Marketing Settings
+
+The `Marketing Settings` give you the possibility to configure marketing-specific settings, which are:
+- Google Analytics
+- Google Search Console
+- Google Tag Manager
+
+
+### Google Analytics
+Google Analytics code is automaticaly injected during rendering the page.
+This behaviour can be disabled using:
+
+```
+$gaListener = $container->get('pimcore.event_listener.frontend.google_analytics_code');
+$gaListener->disable();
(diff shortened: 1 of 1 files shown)


doc/Development_Documentation/18_Tools_and_Features/28_Marketing_Settings.md
```diff
@@ -0,0 +1,16 @@
...
+- Google Search Console
+- Google Tag Manager
+
+
+### Google Analytics
+Google Analytics code is automaticaly injected during rendering the page.
+This behaviour can be disabled using:
+
+```
+$gaListener = $container->get('pimcore.event_listener.frontend.google_analytics_code');
```
maff: The service name was changed to the fully qualified class name in #1929 (see https://github.com/pimcore/pimcore/pull/1929/files#diff-6e09d984014c81decdca0f5a1811cadcR106). Although it is aliases to the previous name and the snippet still works, please reference the fully qualified name in the documentation (`Pimcore\Bundle\CoreBundle\EventListener\Frontend\GoogleAnalyticsCodeListener`).

Yivan:
You mean it become:
`$gaListener = $container->get(Pimcore\Bundle\CoreBundle\EventListener\Frontend\GoogleAnalyticsCodeListener);`
?

Has documentation been updated ? (see https://www.pimcore.org/docs/5.0.0/Extending_Pimcore/Event_API_and_Event_Manager.html) as it refer to `pimcore.event_listener.xxx`.
And the link on this documentation to frontend event link to https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Event/FrontendEvents.php wich doesn't expose `pimcore.event_listener.frontend.google_analytics_code` in a constant. This is normal ?

By the way, it seems the documentation help to migrate from Pimcore 4 to 5, doesn't speak at all about event... (https://www.pimcore.org/docs/5.0.0/Installation_and_Upgrade/Updating_Pimcore/Upgrade_from_4_to_5/index.html)
Many of our project rely on something like that:
```
\Pimcore::getEventManager()->attach("object.postAdd", function (\Zend_EventManager_Event $e) {
    $object = $e->getTarget();
    $object->getId();
    // ...
});
```
Is it handled by the compatility bridge bundle ?
Can we do the same thing with Pimcore 5 ? Could be more convenient to attach event directly from the code logic (and sometimes we attach event only on some logic/dynamic condition which cannot be handled by yaml config i think) than using config yaml files sometimes, and it will provide a way to migrate easy as we just have to refactor light code instead of migrate all event to yaml file. I think the equivalent will be:
```
$listener = new AcmeListener();
$dispatcher->addListener('acme.foo.action', array($listener, 'onFooAction'));
```
Maybe it could be added on the migrate documentation ? (And in the Event documentation. Even if it is said on Symfony documentation, it will be better to have some code snippet directly on Pimcore documentation to help transition from 4 to 5).

Sorry if i ask many question, but as i never used Symfony 3 event before, i think this is question than people comming from Pimcore 4 will have.

maff:
A string:

```php
<?php
// either
$container->get(\Pimcore\Bundle\CoreBundle\EventListener\Frontend\GoogleAnalyticsCodeListener::class);

// or
$container->get('\\Pimcore\\Bundle\\CoreBundle\\EventListener\\Frontend\\GoogleAnalyticsCodeListener');
```

The [FrontendEvents](https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Event/FrontendEvents.php) class does not contain a reference to the Google Analytics listener as it runs on the Symfony core event `kernel.response` (see https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/config/event_listeners.yml#L116).

Regarding updating of events: there is a compatibility layer to Pimcore 4 events, but you should use the new event names and API wherever possible (upgrade notes: https://github.com/pimcore/pimcore/blob/master/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/02_V4_to_V5.md#events).

You're right, you can add events through `EventDispatcher`s API at anytime, but the DI container way is the recommended one, so we won't document that option explicitely (additionally, it depends on the project to find the right timing to add your subscriber). To not handle an event, you can add the check to the event handling method and abort processing.

I updated the documentation to reflect the new event handler service IDs in 49810935184537d10bd1eabad77f2ad4d7822c37.

Yivan:
@maff thanks for taking time to explain.

**I updated the code of the pull-request** (keeping both call possible, by string or class def, more is better ; ))

I missed the documentation page migration talking about event, sorry. Other people could be prone to the same because it is splitted in two different main chapter on the documentation, see this capture please:

double-chapter

Maybe could be grouped so people don't miss one of them ? For instance, move "Migrate for Compatibily Bridge" and "Migrate to Symfony Stack" as children of "V4 to V5" ? (just an idea, maybe there is better arrangement!)

maff:
Thanks! FYI: I changed the example to only include the `::class` version as it results to the same as the string notation but is less error-prone.

PR #1991, opened 2017-09-18
[Pimcore 4] Adding Marketing settings to documentation

Yivan:
This is the pull request for Pimcore 4 for #1882.
As there is no Marketing settings page in documentation I added one here, with a note about desactivate automatic Google Analytics code injection as it can be needed sometimes (to be able to layer some custom debugging code for instance wich need to inject custom code).

This page can be enriched later if neccessary for other marketing settings (they are self explanotory in the GUI for now).


maff:
Thanks!

PR #1994, opened 2017-09-19
Custom Templating Helper Link fix

jremmurd:
Link to example file for creating custom templating helpers was not reachable, now updated to correct url.

From 
https://github.com/pimcore/pimcore/blob/master/install-profiles/demo-basic/src/AppBundle/Templating/Example.php
to
https://github.com/pimcore/pimcore/blob/master/install-profiles/demo-basic/src/AppBundle/Templating/Helper/Example.php


maff:
Thanks!

PR #1999, opened 2017-09-20
[TRANSLATOR] Fix valid language check on adding new message for admin domain

### LINKED ISSUE #1842: [Translations] Admin translations 'automatic adding of new translations' does not work if admin-locale is not a valid frontend language
due to following line 
https://github.com/pimcore/pimcore/blob/7c34745a1d45b94aa0aff76f2829927e9ad40507/pimcore/lib/Pimcore/Translation/Translator.php#L238-L238

when for example website languages are: `en_GB`, `en_US`, `de_DE`
backend user language is: `en`

maff:
Checks messages for admin domain against list of admin languages. Fixes #1842


fashxp:
nice fix!

PR #2015, opened 2017-09-22
Hardcoded mapping in lib/Pimcore/Model/AbstractModel.php

### LINKED ISSUE #2000: Hardcoded mapping in lib/Pimcore/Model/AbstractModel.php
We would like to create our custom DataObject Loader in our custom bundle. But it looks like following hardcoded piece of mapping does block us. 

```
// we have 2 static mappings for objects for performance reasons
if ($this instanceof DataObject\Concrete) {
    $dao = 'Pimcore\Model\DataObject\Concrete\Dao';
} elseif ($this instanceof DataObject\Listing\Concrete) {
    $dao = 'Pimcore\Model\DataObject\Listing\Concrete\Dao';
}
```

It looks like there's no significant performance effect from this solution. But it brakes the Pimcore's flexibility. I'm just curious if you can fix that. Or if we can prepare a PR which will fix that. 

Thank you in advance. 

https://github.com/pimcore/pimcore/blob/99dc0f204e09a83369e02a4bab573dc55970bbd3/pimcore/lib/Pimcore/Model/AbstractModel.php#L96

zhartaunik:
fixed issue #2000 

You had flexibility issue. Architecturally low level models should not care about performance. It's mostly task for server cache. I removed static mapping for several models for further easy overriding these models.

### DIFF
--- a/pimcore/lib/Pimcore/Model/AbstractModel.php
+++ b/pimcore/lib/Pimcore/Model/AbstractModel.php
@@ -93,12 +93,7 @@ abstract class AbstractModel
             self::$daoClassMap = include(PIMCORE_PATH . '/config/dao-classmap.php');
         }
 
-        // we have 2 static mappings for objects for performance reasons
-        if ($this instanceof DataObject\Concrete) {
-            $dao = 'Pimcore\Model\DataObject\Concrete\Dao';
-        } elseif ($this instanceof DataObject\Listing\Concrete) {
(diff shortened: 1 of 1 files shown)


zhartaunik:
I've signed your Contributor License Agreement

brusch:
fixes #2000

maff:
Thanks!

PR #2021, opened 2017-09-25
[CACHE][REDIS] Fix redis cache with LUA not clearing invalidated tags from zc:tags set

### LINKED ISSUE #2009: Redis cache doInvalidateTags() fails to remove tags from zc:tags using LUA
Tag names in the Redis set SET_TAGS (zc:tags) are non-prefixed. But the LUA implementation of doInvalidateTags() gets tag names prefixed ($pTags which is $tags after calling the preprocessTagIds() method). Therefore KEYS in the LUA script are prefixed, and the following command must fail:

https://github.com/pimcore/pimcore/blob/53fe103062833ba3da1cdca9607a5534997651f3/pimcore/lib/Pimcore/Cache/Pool/Redis.php#L494

The non-LUA implementation does it right:
https://github.com/pimcore/pimcore/blob/53fe103062833ba3da1cdca9607a5534997651f3/pimcore/lib/Pimcore/Cache/Pool/Redis.php#L522

maff:
Fixes #2009

### DIFF
--- a/pimcore/lib/Pimcore/Cache/Pool/Redis.php
+++ b/pimcore/lib/Pimcore/Cache/Pool/Redis.php
@@ -42,7 +42,7 @@ class Redis extends AbstractCacheItemPool implements PurgeableCacheItemPoolInter
     const COMPRESS_PREFIX = ":\x1f\x8b";
 
     const LUA_SAVE_SH1 = '1617c9fb2bda7d790bb1aaa320c1099d81825e64';
-    const LUA_CLEAN_SH1 = '42ab2fe548aee5ff540123687a2c39a38b54e4a2';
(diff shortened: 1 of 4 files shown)


maff:
@gadget60 could you check this change against your setup and verify that it fixes your issue?

gadget60:
@maff  I think you must not reuse the variable name "i" for the inner loop - it is already used by the main loop...

gadget60:
@maff don't know why travis build now fails, looks good to me now!

maff:
Travis build is another issue already fixed on master. Thanks for testing!

maff:
@gadget60 merged and added a test for this scenario.

PR #2025, opened 2017-09-26
Object editor: Fix for rendering complex layouts containing invisible items

podarcis:
This is a fixup for d23b6115. There was a missing check for getRecursiveLayout(), which returns false for invisible items. This broke the object editor ("cannot set property 'afterrender' on undefined") for classes with invisible fields in complex layouts.


fashxp:
hmm, doesn't fix the issue for me in advanced-demos category objects

fashxp:
fixed with #2038

PR #2045, opened 2017-10-01
Add missing print.css in demo-basic-twig installation profile

blankse:
In the layout of the demo-basic-twig installation profile is a print.css included:
https://github.com/pimcore/pimcore/blob/master/install-profiles/demo-basic-twig/app/Resources/views/layout.html.twig#L18

But this file is missing. A add the one from the demo-basic installation profile.


fashxp:
thx very much.

PR #1992, opened 2017-09-18
[Pimcore 4] Inheritance can send some undefined errors.

Yivan:
Inheritance can send some undefined errors.
This pull request prevent them for Pimcore 4.

Maybe Pimcore 5 need to be patched as well.

Thanks

PR #2052, opened 2017-10-02
[Pimcore 5] Inheritance can send some undefined errors.

Yivan:
This is the PR containing same fix than https://github.com/pimcore/pimcore/pull/1992 for Pimcore 5.


weisswurstkanone:
Excellent. Many thanks!

PR #2055, opened 2017-10-02
Add possibility to use property "leftNavHide" in twig basic demo.

jentscke:
Add possibility to use the predefined property "leftNavHide" within twig basic demo.


fashxp:
thx very much.

PR #2056, opened 2017-10-03
[Pimcore 4] Update for webservices, cast array to object

avdb:

Cast array to object for wysiwyg

## Additional info  
While testing with the api we discovered that the text from wysiwyg element isn't saved. $wsElement->value is an array and not an object.

## Steps to reproduce
- Create document with webservice api
- Expected result: text of wysiwyg is filled with "hoplaaa"
- Actual result: wysiwyg is empty

Data to test with:
`{
    "path": "\/",
    "userModification": 2,
    "childs": null,
    "controller": "Blog",
    "action": "default",
    "template": null,
    "elements": [
      {
        "type": "wysiwyg",
        "value": {
          "text": "hoplaaa",
          "options": null,
          "name": "richtexteditor",
          "realName": null,
          "inherited": false
        },
        "name": "richtexteditor"
      }
    ],
    "parentId": 1,
    "type": "snippet",
    "key": "test-api-user",
    "index": 2,
    "published": true,
    "userOwner": 2,
    "properties": null
}`

### DIFF
--- a/pimcore/models/Document/Tag/Wysiwyg.php
+++ b/pimcore/models/Document/Tag/Wysiwyg.php
@@ -116,6 +116,9 @@ class Wysiwyg extends Model\Document\Tag
     public function getFromWebserviceImport($wsElement, $document = null, $params = [], $idMapper = null)
     {
         $data = $wsElement->value;
+        if (is_array($data)) {
+            $data =  (object) $data;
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
many thanks!

build 4105

PR #2075, opened 2017-10-06
Update Klarna PaymentProvider because of fatal errors

davidhoeck:
1. The static variables $baseUri and $contentType are not accessible, they aren't even existing. See https://developers.klarna.com/sdk-references/kco_v2_php_4_0_0/class-Klarna_Checkout_Order.html
2. Better idea would be to set the endpoint via the static variable of the Klarna_Checkout_Connector class.

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/PaymentManager/Payment/Klarna.php
+++ b/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/PaymentManager/Payment/Klarna.php
@@ -56,9 +56,9 @@ class Klarna implements IPayment
 
         // set endpoint depending on mode
         if ('live' === $options['mode']) {
-            $this->endpoint = 'https://checkout.klarna.com/checkout/orders';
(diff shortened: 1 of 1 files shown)


fashxp:
thx very much.

davidhoeck:
@fashxp There is also something missing in the configuration. The data array for the ```$order->create($data)``` also needs an array of cart items at ```$data['cart']['items']``` . 
See here: https://developers.klarna.com/en/at/kco-v2/checkout/2-embed-the-checkout 
Because without them, the whole implementation is not working.

fashxp:
Hmm, the config is passed from outside and there we add the items ... since it may be implementation dependent which name, reference, price and tax rates need to be set. 
But should we add a fallback in the initPayment method when `$config['cart']['items']` is not set?

davidhoeck:
Ah ok ... when it is passed from outside it is working of course! A fallback would not be a bad idea...
I would just keep it like that and add the part with the ```$config['cart']['items']``` to the docs! 👍

fashxp:
would be cool, if you could provide an additional pull request ;-) . thank you very much.

PR #2076, opened 2017-10-08
Update outdated pimcore.org links

blankse:
This PR updated all outdated pimcore.org links in the repository.
Some were directed to a 404 page.


fashxp:
thx very much

PR #2084, opened 2017-10-10
Allow classes and id's in CKEditor

Gerhard-Kanzler:
Allow classes and id's in CKEditor on every tag (documents and objects)


brusch:
Needs further investigation how this change affects copy / paste from word.

brusch:
See #2499 and #2674

Issue #2077, opened 2017-10-09
Select datatype - Columnlength


ckogler-elements:
# Feature Request 
The current Select Datatype stores the data as VARCHAR(190) by default. If you have a lot of select fields you get an error because the mysql "Row size too large" error will be thrown. 

Currently we have to change the row format... so that off-page storing is possible...
It would be cool if we can define the "Columnlength" in the data-type settings because most of the time the 190 Chars won't be required.

cleggypdc:
Its been a while since this update, but we're doing an upgrade on some legacy things and I found that the update script on this build is missing the following alterations to field collections and brick definitions:

```
$collectionList = new \Pimcore\Model\Object\Fieldcollection\Definition\Listing();
        $fieldCollections = $collectionList->load();
        /** @var  $collectionDef \Pimcore\Model\Object\Fieldcollection\Definition */
        foreach ($fieldCollections as $collectionDef) {
            $doSave = false;
            foreach ($collectionDef->getFieldDefinitions() as $fieldDef) {
                if($fieldDef instanceof \Pimcore\Model\Object\ClassDefinition\Data\Select) {
                    $fieldDef->setQueryColumnType("varchar");
                    $fieldDef->setColumnType("varchar");
                    $doSave = true;
                }
            }
            if($doSave){
                $collectionDef->save();
            }
        }
        
        $brickList = new \Pimcore\Model\Object\Objectbrick\Definition\Listing();
        $bricks = $brickList->load();
        /** @var  $brickDef \Pimcore\Model\Object\Objectbrick\Definition */
        foreach ($bricks as $brickDef) {
            $doSave = false;
            foreach ($brickDef->getFieldDefinitions() as $fieldDef) {
                if($fieldDef instanceof \Pimcore\Model\Object\ClassDefinition\Data\Select) {
                    $fieldDef->setQueryColumnType("varchar");
                    $fieldDef->setColumnType("varchar");
                    $doSave = true;
                }
            }
            if($doSave){
                $brickDef->save();
            }
        }
`

Hopefully this will help anyone else in the future

PR #2086, opened 2017-10-10
Select datatype - Columnlength

ckogler-elements:
Made columnlength configurable
https://github.com/pimcore/pimcore/issues/2077


pimcore/models/Asset/Listing/Dao.php
```diff
@@ -111,6 +111,8 @@ public function getTotalCount()
         $select = $this->getQuery([new Expression('COUNT(*)')]);
         $select->reset(QueryBuilder::LIMIT_COUNT);
         $select->reset(QueryBuilder::LIMIT_OFFSET);
+        $select->reset(QueryBuilder::ORDER);
```
weisswurstkanone: how is this related to the whole task ?

ckogler-elements: It was just a test if you review the code :p. 

The line is not really related to this Pull request. I think i had just selected the this branch when i committed the issue - sorry . 
Anyway - this should be merged in the core too.

weisswurstkanone:
please resubmit the PR as there are conflicts.

weisswurstkanone:
thx

dpfaffenbauer:
This PR is actually kind of a BC break. Older Class-Definitions (prio to this change) can't now be used anymore. Since old definition file still have the queryColumnType and columnType like this:
```
"queryColumnType": "varchar(190)",
"columnType": "varchar(190)",
```

Pimcore now twice adds the length type:

```
ALTER TABLE `object_query_3` ADD COLUMN `orderLanguage` varchar(190)(190) NULL;
```

This does not only affect select types, it affects all subtypes of it. For example:

 - Language
 - Gender

see here for more: https://travis-ci.org/coreshop/CoreShop/jobs/298138353

In order to fix that, you should probably check if the columnType already has a bracket in it.

#2085 does the same for Pimcore 4.

weisswurstkanone:
Hi,

this issue is taken into account by the update post-update script.
Only point here: the build number is wrong.
Fixed this via:
https://github.com/pimcore/pimcore/commit/2673a33900359eebd289bf0e7bbbb74a35a9ca4e

dpfaffenbauer:
Hi,

But what about plugins/bundles installing classes? They now need to support two versions, the old one and the new one.

ckogler-elements:
The old definition format is now supported too.

dpfaffenbauer:
awesome, cheers (Y) :beers:

PR #2088, opened 2017-10-10
[DOCS] Update documentation for using a editable in twig

jorisros:
I updated some of the documentation about how to use a editable in twig.


maff:
Thank you very much!

PR #2095, opened 2017-10-11
[NewsletterAdapter, CustomReportAdapter] add service locator facade

### LINKED ISSUE #2094: Custom reports: You have requested a non-existent service "pimcore.custom_report.adapter.factories"
I am getting a really weird problem with the Pimcore 5.0.1 (build 133). The 'pimcore.custom_report.adapter.factories' Service can't be retrieved from the container, which kind of makes sense since all service_locator services have their public flag changed to false (in ServiceLocatorTagPass), so they can not be retrieved from the container by just using 'get' on the container. If I remove the 'container.service_locator' tag from the service, then the service is found, but then I get 'Function name get a string' error, since container now tries to call the Class as function. 

Weird is that this seems to function perfectly fine in Pimcore Advanced Demo so I am now out of ideas :)

This is my debug output for this service:
```
$ bin/console debug:container pimcore.custom_report.adapter.factories

Information for Service "pimcore.custom_report.adapter.factories"
=================================================================

 ---------------- ------------------------------------------------------ 
  Option           Value                                                 
 ---------------- ------------------------------------------------------ 
  Service ID       pimcore.custom_report.adapter.factories               
  Class            Symfony\Component\DependencyInjection\ServiceLocator  
  Tags             container.service_locator                             
  Public           no                                                    
  Synthetic        no                                                    
  Lazy             no                                                    
  Shared           yes                                                   
  Abstract         no                                                    
  Autowired        yes                                                   
  Autoconfigured   yes                                                   
 ---------------- ------------------------------------------------------ 
```

I did clear the cache multiple times and rebuilt the DI Container, but nothing seems to work.

Any tip would be highly appreciated.

dpfaffenbauer:
This PR adds a Facade Service for Newsletter Adapter and Custom Report Adapter. This can then be extended to handle services internally somehow different. This also fixes #2094

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
+++ b/pimcore/lib/Pimcore/Bundle/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
@@ -100,8 +100,8 @@ class PimcoreCoreExtension extends Extension implements PrependExtensionInterfac
         $this->configureCache($container, $loader, $config);
         $this->configureTranslations($container, $config['translations']);
         $this->configurePasswordEncoders($container, $config);
-        $this->configureAdapterFactories($container, $config['newsletter']['source_adapters'], 'pimcore.newsletter.address_source_adapter.factories', 'Newsletter Address Source Adapter Factory');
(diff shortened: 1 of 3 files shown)


IgorBenko:
@dpfaffenbauer Thanks!

maff:
See #2094

Issue #1997, opened 2017-09-20
Missing Backend Translations in Safari


ascheider:
Pimcore Platform (5.0.0-BETA1|121)
Backend is not translated in Safari. In Chrome it works fine.
![image](https://user-images.githubusercontent.com/6016485/30635549-0acf78b0-9df3-11e7-9ce4-6e30051d2469.png)
![image](https://user-images.githubusercontent.com/6016485/30635574-1b2714d4-9df3-11e7-9fa1-1f1f4b3feb0c.png)

camico:
In fact the same happens in older Chrome versions...e.g. 46 (okay, no need to be supported, but maybe to give a hint)

maff:
Thanks for reporting, fixed in #2013.

PR #2108, opened 2017-10-13
[PIMCORE4] - Admin translations

jorisros:
In Pimcore 4 are the translations not loaded anymore, this is fixed in pimcore5. See related issue: #1997

## Fixes Issue #
#1997


weisswurstkanone:
many thanks!

PR #2118, opened 2017-10-17
setFrom() in Pimcore\Mail does not work with array of mail addresses

jremmurd:
The `setFrom($email, $name = null)` function of `\Pimcore\Mail` is not working if an array gets passed as `$email` parameter, which is not compatible with the `addFrom()` function of the extended class `\Swift_Mime_SimpleMessage`. This produces an error in line 543 `Warning: preg_match() expects parameter 2 to be string, array given`.

**Reproduce:**
```
$froms = ["test@test.test","asdf@asdf.asdf"];

$mail = new Mail();
$mail->setDocument($myDocument);
$mail->setTo($myMail);

foreach ($froms as $from) {
     $mail->addFrom($from);
}
```

### DIFF
--- a/pimcore/lib/Pimcore/Mail.php
+++ b/pimcore/lib/Pimcore/Mail.php
@@ -531,22 +531,29 @@ class Mail extends \Swift_Message
     /**
      * resets all from headers before setting the new one.
      *
-     * @param string $email
+     * @param string|array $addresses
      * @param null $name
      *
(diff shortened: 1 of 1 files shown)


maff:
As the pwnscriptum check isn't needed for Swiftmailer (as Swiftmailer handles that by itself), we should completely remove the check from Pimcore\Mail. See #2184.

PR #2127, opened 2017-10-18
Track voucher codes in purchase

andreas-gruenwald:
Voucher codes (coupon) is not lost any more when displaying the tracking code on the purchase page.


maff:
Thanks!

PR #2140, opened 2017-10-20
Fixing undefined index -1 error

mbolka:
## Fixes Issue #
Undefined index -1 warning when creating new folder by path.
On the line 751 variable $pathsArray is initialized as empty array.
On the first loop iteration (line 762) PHP tires to get value from element with index -1 what causes problems with tests.
Checking if key in array exists.

### DIFF
--- a/pimcore/models/Element/Service.php
+++ b/pimcore/models/Element/Service.php
@@ -760,7 +760,7 @@ class Service extends Model\AbstractModel
         if (!($foundElement = $type::getByPath($sanitizedPath))) {
             foreach ($parts as $part) {
                 $pathPart = '';
-                if (array_key_exists(count($pathsArray) - 1, $pathsArray)) {
+                if (count($pathsArray) > 0 ) {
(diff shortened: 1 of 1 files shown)


pimcore/models/Element/Service.php
```diff
@@ -759,7 +759,11 @@ public static function createFolderByPath($path, $options = [])
 
         if (!($foundElement = $type::getByPath($sanitizedPath))) {
             foreach ($parts as $part) {
-                $pathsArray[] = $pathsArray[count($pathsArray) - 1] . '/' . self::getValidKey($part, $type);
+                $pathPart = '';
```
mgomse: @mbolka IMHO this can be a little better, don't you think?:
```php
$pathPart = $pathsArray[count($pathsArray) - 1] ?? '';
$pathsArray[] = $pathPart . '/' . self::getValidKey($part, $type);
```

brusch:
Thanks!

PR #2141, opened 2017-10-20
AbstractCart - add getter for ignoreReadonly

ckogler-elements:
Added a getter for ignoreReadonly so we can access the value from a extended class. 
E.g to automatically cance a active payment


fashxp:
thx

PR #2144, opened 2017-10-21
[PIMCORE4] Fix for the update command: --ignore-maintenance-mode

kkarkus:
Update puts the system into maintenance mode, so the internal:update-process command should ignore maintenance mode when being run.


kkarkus:
@weisswurstkanone can you please check this?

fashxp:
thx very much.

PR #2146, opened 2017-10-23
fix Zend_Date method usages

AlternateIf:
remove zend_date method usages and replace them with DateTime methods

@fashxp  @maff

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/PricingManager/Condition/DateRange.php
+++ b/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/PricingManager/Condition/DateRange.php
@@ -36,7 +36,7 @@ class DateRange implements IDateRange
     public function check(IEnvironment $environment)
     {
         if ($this->getStarting() && $this->getEnding()) {
-            return $this->getStarting()->isEarlier(time()) && $this->getEnding()->isLater(time());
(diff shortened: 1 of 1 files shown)


pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/PricingManager/Condition/DateRange.php
```diff
@@ -103,10 +103,10 @@ public function fromJSON($string)
     {
         $json = json_decode($string);
 
-        $starting = new \DateTime(strtotime($json->starting));
+        $starting = \DateTime::createFromFormat('Y-m-d\TH:i:s', $json->starting);
         $starting->setTime(0, 0, 0);
 
-        $ending = new \DateTime(strtotime($json->ending));
+        $ending = \DateTime::createFromFormat('Y-m-d\TH:i:s', $json->ending);
```
fashxp: hmm, toJSON() needs to form a value that can be read by fromJSON() ... there we have timestamp vs. datestring

AlternateIf: the difference in types probably comes from the extsjs datefield that is configured in https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/Resources/public/js/pricing/config/item.js#L659

fashxp:
thx

PR #2162, opened 2017-10-25
Fix doubled setting options in twig example

Cruiser13:


maff:
Thanks! For the record: if adding select stores with labels it should be in the format (add an array with 2 elements):

```twig
{% set selectStore = selectStore|merge([[i, i ~ ' Items']]) %}
```

PR #2163, opened 2017-10-25
Context indices of elements inside fieldcollections aren't updated

IgorBenko:
Problem: Context indices of elements inside fieldcollections aren't updated on adding/deleting/moving field collection entries 

Fix: this updates the indices after any add/delete/move operation


weisswurstkanone:
thx!

PR #2165, opened 2017-10-25
Allow to override system translations using JSON files

Hagith:

Allow to override system translations by providing custom `%locale%.json` files from project's bundle(s).

### DIFF
--- a/doc/Development_Documentation/06_Multi_Language_i18n/README.md
+++ b/doc/Development_Documentation/06_Multi_Language_i18n/README.md
@@ -49,4 +49,16 @@ English is maintained by the core team. In addition to that, everybody can join 
  [Pimcore translation project](http://www.pimcore.org/en/community/translations) to add system translations in additional
  languages. With every Pimcore release, newly added translations are added to the Pimcore installation package.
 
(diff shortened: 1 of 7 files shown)


brusch:
I don't think we need an extra method for configuring a custom translations source, because you should be able to use Symfony's standard way of loading translations, see: 
https://symfony.com/doc/current/translation.html#translation-resource-file-names-and-locations 

So with eg `app/Resources/translations/admin.en.yml` you should be able to add/override admin/system translations in Pimcore

Hagith:
I already tried to override system translations using `app/Resources/translations/admin.en.yml` and `src/AppBundle/Resources/translations/admin.en.yml`.
This way I was only able to provide translations for the keys, which are not already translated by Pimcore.

Maybe I'm missing something, but it looks that translations provided by Pimcore have higher priority than any translations provided from `app` or `bundle`.

This is why I extended `Translator` to merge custom translations on catalogue creation level.

brusch:
Ok, thanks for your feedback! 
I guess we need to change the priorities then, ... they should be (for domain `admin`): 
1. pimcore stock translations
2. other bundles
3. pimcore admin translations

PR #2171, opened 2017-10-27
Multiselect didn't return definition without options provider present

michaelclaesson:
## Fixes Issue
When you're using a multiselect in a classification store the definition of the class does not get returned if no options provider is present since the return statement lies within the option provider conditional check.

Moved return statement outside the conditional scope.


weisswurstkanone:
thx!

PR #2174, opened 2017-10-30
Allow to depend on dev packages

jbq:
  
Update composer.json to allow depending on dev packages for local development.  It is not possible to set those options in `composer.local.json` unfortunately, only in the root `composer.json` file.

## Additional info  

Without this option set it is not possible to depend on development version of any package.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,8 @@
     "optimize-autoloader": true,
     "sort-packages": true
   },
+  "minimum-stability": "dev",
+  "prefer-stable": true,
   "repositories": {
     "vrana-jush": {
       "info1": "External dependency of vrana/adminer which isn't included or referenced by adminer, see also https://sourceforge.net/p/adminer/bugs-and-features/468/ ",


brusch:
Hmm, I don't think this should be the default setting.
If you need to use dev-packages for your projects you can add those configurations manually, they won't get overwritten by updates :) 
Any other opinions?

dpfaffenbauer:
I am with @brusch at this one. Doesn't seem to be a good idea to depend on dev packages by default. As @brusch said: you can do it yourself and it won't get overwritten.

jbq:
This means everytime I upgrade Pimcore I need to copy the latest `composer.json` in my project and integrate this two-line change to allow depending on dev packages for my project.  A bit cumbersome.

dpfaffenbauer:
@jbq no, not at all. Pimcore merges the composer json and will not touch any of your custom entries.

jbq:
@dpfaffenbauer alas `composer` refuses to honor those options if they are put in `composer.local.json`, they are only read when in the root `composer.json` file.

dpfaffenbauer:
@jbq right, I am aware of that. Still, I don't think its a good idea to use dev packages per default. If you need it in your project, thats right. Pimcore respects your composer.json and will never overwrite any of your settings on updates.

PR #2182, opened 2017-10-31
Fix for calling undefined method

kkirsz:
This PR resolves following problem while opening object: Attempted to call an undefined method named "getParam" of class "Symfony\Component\EventDispatcher\GenericEvent".
workflow-custom-layouts


maff:
Thanks!

PR #2187, opened 2017-10-31
fixed usage of legacy-class

AlternateIf:
fixed usage of legacy-class


AlternateIf:
see #2186

@fashxp

fashxp:
thx

PR #2190, opened 2017-11-02
Updating mime types for use with Safari

Kzumueller:
While the previous values were completely correct, I suggest these new types since Safari on MacOS 10 and up will simply append the file extensions '.ps' for .ai or '.psd' for .psb, regardless of an already present extension, apparently based solely on the mime type.
Proper browsers and Microsoft products are not affected and handle both alternatives in the same way.


brusch:
Thanks!

PR #2206, opened 2017-11-07
add gift item to cart only if available

andreas-gruenwald:
Gift items should only be added to a cart or a order if they are still available based on the information of the availability system.


maff:
Thanks!

PR #2213, opened 2017-11-07
add Doctrine filter for DeepCopy

dpfaffenbauer:
This PR makes it possible to clone Pimcore Objects which relate to Doctrine Objects (Collections). Otherwise cloning entities is impossible, since it throws a "container not clone-able" exception.


maff:
Thanks!

PR #2225, opened 2017-11-13
Frontend object data validation

IgorBenko:
This pull request makes it possible to validate object data in the frontend before saving. It is also possible to cancel saving if the data is not valid by throwing exceptions in javascript code.

There are two exceptions that can be thrown:

* pimcore.error.ValidationException
* pimcore.error.ActionCancelledException

The only difference between above exceptions is in how the error is displayed:

* ValidationException displays error alert (the same as ValidationException coming from the backend)
* ActionCancelledException displays instead just a non intrusive notification in the bottom right corner.

Example preSaveObject function in plugin:
```
preSaveObject: function () {
    var userAnswer = confirm("Are you sure you want to change the name?");
    if (userAnswer == false) {
        throw new pimcore.error.ActionCancelledException('Cancelled by user');
    }
}
```

I will update the docs if/when this pull request gets approved. The same mechanism could probably also be applied to documents and assets if that might be of interest to you.


IgorBenko:
Hi @maff , sorry for nagging but I just wanted to check with you if there is any interest from your side for this implementation and do you see this beeing added anytime soon? Otherwise I must look for an alternative solution for our client. Backend validation unfortunately doesn't help us in our case.
If you have any suggestion on improving/extending my Pull Request do let me know and I will do my best.

maff:
Hey @IgorBenko, looks good to us - If you could also provide docs we'll be happy to merge :)

IgorBenko:
Hi @maff, thanks a lot for your prompt reply :) 👍 
I will try to add the docs to this pull request ASAP.

IgorBenko:
Documentation added :-)

maff:
Thanks a lot! :)

PR #2229, opened 2017-11-13
Added getter for update host

davidhoeck:
Because the updateHost variables has private access, it is causing an error in the RunUpdateScriptCommand


weisswurstkanone:
thx

PR #2241, opened 2017-11-15
Update Select.php

ollegz:
This will fix (in pimcore 4.6.3) so that you don't fetch columnLength (that's already in columnType) -> varchar(255)(190).

Reproduce by saving a fieldcollection with select data type through `->save();`

Fatal error: Uncaught PDOException: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '(190) NULL' at line 1 in /opt/pim/www/vendor/zendframework/zendframework1/library/Zend/Db/Statement/Pdo.php on line 235

Zend_Db_Statement_Exception: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '(190) NULL' at line 1, query was: ALTER TABLE `object_collection_DeviceBatteryTimes_1` CHANGE COLUMN `Standard` `Standard` varchar(255)(190) NULL; in /opt/pim/www/vendor/zendframework/zendframework1/library/Zend/Db/Statement/Pdo.php on line 235

### DIFF
--- a/pimcore/models/Object/ClassDefinition/Data/Select.php
+++ b/pimcore/models/Object/ClassDefinition/Data/Select.php
@@ -102,7 +102,7 @@ class Select extends Model\Object\ClassDefinition\Data
      */
     public function getColumnType()
     {
-        return $this->columnType . "(" . $this->getColumnLength() . ")";
+        return $this->columnType;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
thats not the right solution, have a look at this: they did that for Pimcore 5 but not for 4:

https://github.com/pimcore/pimcore/commit/878d8e62f0476e9ab5691d6e732bfd0f099b6566

fashxp:
thx very much

PR #2247, opened 2017-11-16
Ogone payment provider integration:

andreas-gruenwald:
- New payment provider class for ogone
   - New PaymentProvider brick and update for the PaymentInfo field collections.
   - Add additional section within the E-Commerce documentation.


fashxp:
thx very much

PR #2261, opened 2017-11-23
Option to prevent Pimcore from changing PHP error handling settings 

bbak:
## Fixes Issue #
Pimcore (more or less) unconditionally changes error_reporting, error_log and log_error settings. This prevents the use of a (unified) remote logging Service - at least through settings in php.ini.  

Introduced a new (boolean) Constant PIMCORE_ALLOW_PHP_ERROR_LOG_OVERRIDE that allows to prevent Pimcore from changing these PHP settings.


bbak:
Hi, any Problems with this PR? We need that option, since we're passing all logs to a Remote Log Service (Loggly, Papertrail, LogEntries, Logsene, there are many of them). Most of them utilize the syslog protocol. It's quite easy to configure Monolog to use Syslog, so the Application-Level Logs are covered. And of course we want PHP to log through Syslog also, which requires error_log to be set to "syslog".

dpfaffenbauer:
@bbak can't you configure your remote log system to watch for any kind of log? as well as pimcore's php log? Logstash for example can do that.

brusch:
I think it would be better to have 2 constants, one for a custom error level and one for a custom `error_log` value. 

eg. 
`PIMCORE_PHP_ERROR_REPORTING`
`PIMCORE_PHP_ERROR_LOG`

With that you are even able to change the value using DotEnv. 
Other opinions?

bbak:
@dpfaffenbauer Sure, most of them can parse log files. In the end, many of them pass the parsing result to syslog - so there's a unnecessary overhead here. Additionally, when parsing a file, you have to (often manually) re-establish correct Error-Levels, etc. ... And there's IMHO no technical reason that Pimcore changes these settings, especially with a lack of an opt-out.

@brush I'd be fine with that. Gimme a minute or two to update my PR...

bbak:
@brusch Now ok?

bbak:
Thanks!

PR #2266, opened 2017-11-23
PIMCORE4 MultihrefMetdata columnbool

cleggypdc:
This is a slightly different implementation of the boolean field on the multihrefmetadata component.

It forces a whole column of hrefs to be limited to one selection only. Similar to Magento 1's image selection component (that's why i had to implement it)

columnbool

Currently only frontend, it probably needs some backend validation, I'm happy to PR too just let me know.

This is reletively easy to port to Pimcore 5, if you think its a good feature let me know and I will port.

### DIFF
--- a/pimcore/config/texts/en.json
+++ b/pimcore/config/texts/en.json
@@ -4593,6 +4593,10 @@
         "definition": "Bool"
     },
     {
+        "term": "objectsMetadata_type_columnbool",
+        "definition": "Column Bool"
+    },
+    {
         "term": "objectsMetadata_type_select",
         "definition": "Select"
     },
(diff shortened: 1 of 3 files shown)


weisswurstkanone:
Hi, what about the translation for "objectsMetadata_type_columnbool" ?

cleggypdc:
whats the best way to provide the translation, just in the lang file?

weisswurstkanone:
Translations have to be added to the \pimcore\config\texts\en.json file

weisswurstkanone:
Also a PR for pimcore 5 would be nice

weisswurstkanone:
Hi, looks good. Do you have a PR for Pimcore 5 as well ?

cleggypdc:
@weisswurstkanone sorry for the delay! still in my backlog to do ;)

weisswurstkanone:
many thanks!

PR #2278, opened 2017-11-28
PostAdd Event for notes

ckogler-elements:
Added element.note.postAdd Event


ckogler-elements:
Hmmm... something strange happen - going to create a new pull request

PR #2282, opened 2017-11-29
Corrected a mistake in the code example.

thmspl:
This commit corrects a little mistake in the pimcore docs.


maff:
Thanks for the PR :) Actually this is no mistake in the code example, but a bug how the editmode widget is built. Fixed in #2286.

thmspl:
Ah okey. Cause it didn't work for me if I try it like it's described in the examples.

PR #2301, opened 2017-12-01
fix xliffImport header status

solverat:
funny, apparently no one has ever noticed that the xliffImport actually never worked in p5. The second parameter returns `false` which will lead to a `The HTTP status code "0" is not valid.` error message.


weisswurstkanone:
many thanks!

Issue #2236, opened 2017-11-14
No debug.log


Yivan:
During an update from Pimcore i got some error, but there where no file `debug.log` on the server at all.

![no-debug](https://user-images.githubusercontent.com/9025839/32794568-80d2590a-c969-11e7-985e-ca586bee8676.png)

I searched for `debug.log` all over the server (linux find command). Searching for the file `php.log` nothing too.
Nothing in /var/log of the system, nor the var/log of Pimcore.
Is monolog replacing it by some other file or is php error logging disabled ?
I found nothing about it on: https://github.com/pimcore/pimcore/blob/c75105230152bd263d43f8b0da9ae2a7c77579ea/doc/Development_Documentation/19_Development_Tools_and_Details/07_Logging.md
dev.log or prod.log doesn't contains the errors (i searched for the touch problem in the dev.log, but nothing).

Is there some special things to do to activate it ? If yes, it could be documented ?
Thanks.

(aside: maybe the update popin need some CSS modification so message doesn't overflow to the right of the popin, like on my screen capture above)

Yivan:
I should add the phpinfo (from the pimcore menu):

![error_level](https://user-images.githubusercontent.com/9025839/32795309-962d7c4c-c96b-11e7-88b1-1d31d412f94e.png)

All seems ok at this level, but the file `php.log` doesn't exist after the install error like i explained.

davidhoeck:
@Yivan The NGINX or Apache User is not able to write your `debug.log`

Please try the following commands in your project directory depending on your server setup: 

Apache: `sudo chown -R apache:apache . `
Nginx: `sudo chown -R www-data:www-data . `

Yivan:
Thanks @davidhoeck but I think PHP should be able to write debug.log as dev.log and prod.log in var/log haven't any writing problem, and both of those are writen by www-data too.
With the same permission configuration I haven't any problem with Pimcore 4.

fashxp:
which pimcore version are you using?

Yivan:
@fashxp I used the last stable Pimcore 5 version downloaded from the website.

fashxp:
in pimcore 5 there is no debug.log anymore. it is - depending on the environment - dev.log or prod.log ... so it is just the text that is misleading.

fashxp:
created a new issue for that: https://github.com/pimcore/pimcore/issues/2288

Yivan:
@fashxp I understand better thanks.

[...]

PR #2302, opened 2017-12-02
Add install/update log file information

Yivan:
@maff @fashxp This is the PR for https://github.com/pimcore/pimcore/issues/2236

It adds the installer logs file directory in error message for install/update.

Thanks

### DIFF
--- a/pimcore/lib/Pimcore/Update.php
+++ b/pimcore/lib/Pimcore/Update.php
@@ -352,7 +352,7 @@ class Update
             } catch (\Exception $e) {
                 Logger::error($e);
                 $outputMessage .= 'EXCEPTION: ' . $e->getMessage();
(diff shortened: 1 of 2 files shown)


dpfaffenbauer:
IMO, thats not 100% correct. When updating and something fails, I'll get the message to check installer-logs, which is not true, since I am not installing anything, I am updating Pimcore.

Yivan:
But if you see my original issue https://github.com/pimcore/pimcore/issues/2236, which is an update process,  i didn't get the touch warning in neither dev or prod log, nor in any log files of /var/log. So i think update rely on the same installer too and log are in /var/installer ?
@maff do you confirm this ?

maff:
Sorry, seems I misunderstood #2236 - `var/installer/logs` is **only** used when installing Pimcore, not for updates.

Yivan:
Ha ok! So yes this PR is no more good. Thanks.

Issue #1831, opened 2017-08-17
Combining setCondition and addConditionParameter can break SQL request.


Yivan:
Using object API to request list, can lead to malformed SQL request if setCondition is used and after addConditionParameter add some conditions.
The problem only occurs when the resulting list is empty (all the combined criterea send no result).

Here is a sample of the code:

```
$appList = new App\Listing();
if (isset($params["category"]) && $params["category"]) $appList->setCondition("category = ?", $params["category"]);
if (isset($params["isFree"])) {
	if ($params["isFree"] == "0") $appList->addConditionParam("isFree = ? OR isFree IS NULL", 0);
	if ($params["isFree"] == "1") $appList->addConditionParam("isFree = ?", 1);
}
if (isset($params["order"]) && $params["order"] == "averageRating") {
	$appList->setOrderKey("averageRating");
	$appList->setOrder("desc");
} else {
	$appList->setOrderKey("lastUpdate");
	$appList->setOrder("desc");
}
$appList->load();
```

Work ok when their is results. But as soon $appList->load() will have 0 objects, an error is sent:

```
`SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens, query was: SELECT object_18.o_id as o_id, `object_18`.`o_type` FROM `object_18` WHERE ((category = ? AND isFree = ? AND object_18.o_type IN ('object','folder')) AND object_18.o_published = 1) ORDER BY `lastUpdate` DESC"`

file: "...\www\vendor\zendframework\zendframework1\library\Zend\Db\Statement\Pdo.php"
```

For now the solution is to replace in the code above the first `setCondition` by `addConditionParam`, and it works great even with empty resulting list.
But i think it should be corrected at the API level, as setCondition and addConditionParam could be used in combination, as indicated in the documentation:

https://www.pimcore.org/docs/latest/Objects/Working_with_PHP_API.html
```
$entries->setCondition("name LIKE ?", ["%bernie%"]); // use prepared statements! Mysqli only supports ? placeholders
// or
$entries->setCondition("name LIKE :name", ["name" => "%bernie%"]); // With PDO_Mysql you can use named parameters
// to add param to the condition
$entries->addConditionParam("city = ?", "New York", "AND"); // concatenator can be AND or OR
```

Yivan:
Test was done on Pimcore 4 last stable build.

Yivan:
[...]

PR #2303, opened 2017-12-02
Add doc information that setCondition and addConditionParam cannot be used together

Yivan:
This pull request is for https://github.com/pimcore/pimcore/issues/1831

It add information to the documentation that setCondition and addConditionParam cannot be used together.
(as explained in the issue it lead to error/strange result)


brusch:
Thanks!

Issue #2234, opened 2017-11-14
Pimcore update permission problem


Yivan:
When doing an update from stable 5.0.1 to stable 5.0.2, we encoutered several permission problems.

**Process to reproduce:**
- Deploy all project file from a base Pimcore 5.0.1 by sFTP for instance (with a sftp user : myuser being from the group www-data, classic sftp configuration).
- Ensure all file are 664 or 775 (so www-data can write all files)
- Run `chown -R www-data:www-data web/var var pimcore app/config` as recommended in the Pimcore documentation

**Problem:**
Respecting https://pimcore.com/docs/5.0.x/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/File_Permissions.html isn't sufficient. Udpate touch somes files in other directory too:

- web/pimcore
- bin
- composer.json

So a good command would be :
chown -R www-data:www-data web/var var pimcore app/config web/pimcore bin composer.json
This is needed because of the touch() function here: https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Update.php#L304
It need owner of the file to be the same of the php process.

Even with the chown given above, the composer update failed because of
\vendor\ocramius\package-versions\src\PackageVersions\Installer.php, line 127:
`chmod($installPath, 0664);`
![chmod](https://user-images.githubusercontent.com/9025839/32793739-0e8e1638-c967-11e7-8f7d-4427337989fd.png)
It seems to chmod other directory than those chowned above.

**So finally i ended up to:**
`chown -R www-data:www-data *`
**Is it the good way ? Maybe the permission page could be updated to give us the way to configure it.**
(i really lost some times trying to figure out, and finally using the global chown, even i i am not fan of it, because no other solution seems to work...)

Actually `chown -R www-data:www-data web/var var pimcore app/config` isn't sufficient.
This one could be added too documentation too on the permission page:
`chmod ug+x bin/*`
As well, running console not in root, but as user or better php user, for instance:
`su -l www-data -s /bin/bash` do the trick.

Just for reference for user having permission problem, the symfony recommendation:
https://symfony.com/doc/current/setup/file_permissions.html

PR #2304, opened 2017-12-02
doc permissions improvement

Yivan:
This pull request add informations to the documentation to help setting permissions.

It is intended to resolve issue https://github.com/pimcore/pimcore/issues/2234, so the process can be more easy for new users.
(it doesn't resolve the Ocramius/PackageVersions touch problem which lead to the necessity to make a `chown -R YOURUSER:YOURGROUP *`, so setting directory by directory write permissions don't work for now).


brusch:
Thanks!

Issue #2258, opened 2017-11-21
[Improvement] Abilities to filter object list with events and custom condition.


Yivan:
This impovement is for object listing in Pimcore backend / GUI.

Sometimes we need to remove some objects from listing depending other parameters than the role/user defined (using role/user workspaces).
In such case, we need to hook an event to be able to modify the list returned. Proper way is to modify the condition request before the SQL request is sended to the database. This way pagination is not broken (if we remove item after, pagination will be no more good).

Actually, object list can be filtered like this, if i am not wrong:

- search list:
SEARCH_LIST_BEFORE_FILTER_PREPARE (event: pimcore.admin.search.list.beforeFilterPrepare')
This one is ok, as we can inject custom condition params in the requestParams.
See: https://github.com/pimcore/pimcore/blob/5483c6f0fa00cfb4026b15ef5ffbfaf9dcee66ca/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Searchadmin/SearchController.php#L54

- tree list:
We got only OBJECT_TREE_GET_CHILDREN_BY_ID_PRE_SEND_DATA ( event: pimcore.admin.dataobject.treeGetChildsById.preSendData)
This one is not ok, because we can only filter after the SQL request has been done, so removing element will break pagination.
**Like for the search list, it would be great to add an event to be able to inject condition parameters in the SQL request.**
Actually, the only way to achieve proper object removing in the tree is to use sql rule filter on the custom views (which is well handled before make the SQL request: https://github.com/pimcore/pimcore/blob/57cc6d0a39602f140549348ac90efe0d26aaa899/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/DataObjectController.php#L83)

- object list:
Actually there is no way to hook, **so maybe we could add an event here as first line of the action, so just before this line**:
https://github.com/pimcore/pimcore/blob/5483c6f0fa00cfb4026b15ef5ffbfaf9dcee66ca/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/DataObjectController.php#L1525
This event will take the request as parameter so we can modify it, and inject specific condition parameter in it.

To resume, i think by adding 2 new events, one for the tree list and one for the object list, developpers can easily make consistent list on all Pimcore backend according to custom conditions.

Use case:
Many, but one of them often asked will be the user can only see the object he created. With those events it will be possible by adding a condition owner_id = user_id.
[...]

Issue #1838, opened 2017-08-19
addConditionParam method for list don't encapsulate the query fragment


Yivan:
Actually using addConditionParam method for object listing dont put parenthesis for the sql fragment.
I think the usual way is it is automaticaly encapsulated by parenthesis because:
- ZF1 work like this when you add a ->where in a chaining select
- If not parenthesis is added, the 3rd parameter (the concatenator AND/OR) is less predictable

Exemple:

```
$appList = new App\Listing();
if (isset($params["category"]) && $params["category"]) $appList->addConditionParam("category = ?", $params["category"]);
if (isset($params["isFree"])) {
	if ($params["isFree"] == "0") $appList->addConditionParam("isFree = ? OR isFree IS NULL", 0);
	if ($params["isFree"] == "1") $appList->addConditionParam("isFree = ?", 1);
}
if (isset($params["order"]) && $params["order"] == "averageRating") {
	$appList->setOrderKey("averageRating");
	$appList->setOrder("desc");
} else {
	$appList->setOrderKey("lastUpdate");
	$appList->setOrder("desc");
}
$appList->load();
echo $appList->getQuery()->__toString();
```

Will output:
`SELECT object_18.o_id as o_id, `object_18`.`o_type` FROM `object_18` WHERE ((category = ?  AND isFree = ? OR isFree IS NULL  AND  object_18.o_type IN ('object','folder')) AND object_18.o_published = 1) ORDER BY `lastUpdate` DESC`

As it can be seen the **"isFree = ? OR isFree"** fragment is not encapsulated by parenthesis.

Proposed solution:
Maybe here https://github.com/pimcore/pimcore/blob/pimcore4/pimcore/lib/Pimcore/Model/Listing/AbstractListing.php#L209 we could add:
**$key = "(".$key.")";**

Yivan:
Can I made a pull request with the proposed modification so the addConditionParam is properly encapsulated ?

_But maybe it could break existing Pimcore because some code could rely on being not encapsulated...
If you think it will break backward compatibility, maybe we should add a bolean parameter to `addConditionParam`. It could be named `$encapsulate` and be false by default (so it won't break compatibility). Setting it true will encapsulate like this: `$key = "(".$key.")";`_

fashxp:
how was the behavior in pimcore4?

Yivan:
This is the behaviour on Pimcore 4.

For instance, making:
```
$appList->addConditionParam("category = ?", 17)
[...]

Issue #1618, opened 2017-06-20
Improvement: extending list and traits for list


Yivan:
Like object, objectList could be extended the same way and having the possibility to use traits on them too:

![list-object](https://user-images.githubusercontent.com/9025839/27331786-37c02a44-55bf-11e7-9a5e-728f8c1219b8.png)

Fill those two new fields will generate the classList file (`listing.php` in `website\var\classes\Object\...`) based on this.

This way we could have true inheritence available if needed (sometimes DI is not sufficient).

fashxp:
why not use overriding models functionality for that? 
https://pimcore.com/docs/5.x/Development_Documentation/Extending_Pimcore/Overriding_Models.html

Yivan:
@fashxp yes for sure, we can overide the model.
But using override is less powerfull as it is only active at run time (using static method like `::getList()`).
But if you use the classic:
`$objectList = new Pimcore\Model\DataObject\MyClass\Listing();`
You don't have the possibility to have any extend/traits.

Actually, the only way is like you said overide the model list and call the new class which override: `new AppBundle\Model\DataObject\MyClass\Listing();`
Problem is if it is an existing project it needs code refactoring : ( and even for new project developpers generally prefer to call directly the pimcore class name (not the overiding one).

To resume, having those two new field would allow a proper extending directly in the model objectList generated by Pimcore (like it is for object).

If i am not wrong, I think this is a really valuable addition and this issue should be reopen please.
And i think it is pretty straightforward to implement (two new extjs fields for the class, no db change, and adding the value of those field in the generated list class, exactly like it is done for object class).

If you are ok to implement it, I can prepare a pull request if you want.

Thanks a lot.

fashxp:
yes, that would be great!

PR #2306, opened 2017-12-03
New events for filtering

Yivan:
This pull request is for https://github.com/pimcore/pimcore/issues/2258

With this one, alter listing of object, document and asset is more convenient and possible globally.

I should note:
- ASSET_LIST_BEFORE_LIST_LOAD, DOCUMENT_LIST_BEFORE_LIST_LOAD and OBJECT_LIST_BEFORE_LIST_LOAD are the three main and important new events.
- I added {element type}_LIST_BEFORE_FILTER_PREPARE and {element type}_LIST_AFTER_LIST_LOAD for the listing in panels (not in the trees) so it works like the search events.
- Search event have been modified/fixed because `$allParams['context']` is always empty it seems... so it has been replaced by `'context' => $allParams` instead of `'context' => $allParams['context']`.
- In some places `->isAllowed('list')` has been added, so even without using the event hooks, developper can be confident in isAllowed method (actually it was not reliable as for instance proxy grid make direct BDD call). It will slow down a little the request, but ensure security when custom permission reliying on extending/overload isAllowed method are done.

On special note for tags:
- TagController has been modified to hook to he global listing event (when doing tag propagation).
For this one i was not sure because it rely on the "view" permission, and all the hook are more focused on the "list" permission. If you think we should not add the global hooks here you can remove them.
But i found a security breach: there is no control of permission during the action `doBatchAssignmentAction`. So user can change tags whatever his permission. I think there should be somewhere a isAllowed check. Maybe view permission ? The documentation https://pimcore.com/docs/5.0.x/Development_Documentation/Administration_of_Pimcore/Users_and_Roles.html don't give any clues which permission allow user to assign tags... maybe it should be added (list, view or publish ?).

To finish, i added to the event page of the documentation in the exemple section, the way to properly modifying listing globaly. I think this pull request bring real useful events which offer now the possiblity to have custom rules to control which elements are displayed based on those rules. Use case are multiple.
I added in the documentation, that it is better to double the custom permission rules with extending isAllowed method so it is safe from end to end (and not just only based on hook/events).
If you think this PR bring an important feature, maybe it should be more detailled and ported to a dedicated pages or to the permission pages ?


Yivan:
Just forgot this point: it seems that export from the grid list don't check permission with thisAllowed... is it directly BDD permission checking ? Or there is no permissions check ?

dpfaffenbauer:
IMO, I think your PR is insufficient, you shouldn't filter in the Controller, you should filter it directly within the Listing Class to be globally available and not only from certain Controller calls. Therefore, I would change it and fire the event within the Listing Class.

Yivan:
In fact, to be usable this PR need the following issue dependies to be resolved i think:
- https://github.com/pimcore/pimcore/issues/1831: because we should be able to use addConditionParam to add some custom rules as setCondition has already been setted before the hook. So this issue need to be resolved so we can easily add condition in the OBJECT_LIST_BEFORE_LIST_LOAD event hook for instance.
- https://github.com/pimcore/pimcore/issues/1838: this one too, for conveniance, because using addConditionParam could break the condition if developper forget to encapsulate himself the addConditionParam. .i.e actually he must always add parenthesis for safety like: `$list->addConditionParam("(param= ?)", $value)`

Even with an event fired from the Listing Class, this two issues need to be resolved, so developper can easily add conditions.

I totally agree an event directly on the Listing Class model will be the best! I was hesitating about it because i don't know if Pimcore team would agree for such deep nested event.
Problem could be that it will impact all list load, even frontend, so in the event hook the developper had to be precotionous about the context. For instance, he can rely on the module of the request and limit only `if ($module === "admin")` (we cannot rely totally on the user connected because it will modify frontend request if he his connected in the same time in Pimcore). This context can not be send to the event i think.
Will `$this->container->get('request_stack')` be callable in the event hook ?

~~If this issue is implemented https://github.com/pimcore/pimcore/issues/1618, we don't need any more events and we could directly overload the "load" method of list object : )) (even the setCondition, etc... so a lot more possibility). Actually using DI won't work because extends is only static (using getList method).~~
=> EDIT: in fact it won't allow to filter globally because we should be able to extend  DataObject\Listing  : (( (see for instance: https://github.com/Yivan/pimcore/blob/8306362c941b8b01eae6a6ee0a94df4ddb1fdd7b/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/DataObjectController.php#L80)
So like you say a hook is needed and DI/extending can not make the work.

pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/AssetController.php
```diff
@@ -180,6 +182,14 @@ public function treeGetChildsByIdAction(Request $request)
             $childsList->setOrderKey("FIELD(assets.type, 'folder') DESC, assets.filename ASC", false);
 
             \Pimcore\Model\Element\Service::addTreeFilterJoins($cv, $childsList);
+
+            $beforeListLoadEvent = new GenericEvent($this, [
+                'list' => $childsList,
+                'context' => $allParams
+            ]);
+            \Pimcore::getEventDispatcher()->dispatch(AdminEvents::ASSET_LIST_BEFORE_LIST_LOAD, $beforeListLoadEvent);
```
kkarkus: Maybe injecting directly event dispatcher could safe time in a future? There won't be need to refactor this.

brusch: @Yivan can you please update your PR to use DI for the event dispatcher? Afterwards we're happy to merge your PR. 

Thanks!

weisswurstkanone:
Any updates on this ?

Yivan:
@kkarkus Sorry, what to you mean by directly injecting it ? Actually all events seems to be this way (for instance: https://github.com/pimcore/pimcore/blob/e0a341a0d08febda134fe0f90a7d07ce9f72fed1/pimcore/models/Asset/Video.php#L130)
Could you give me some exemples please ?

@dpfaffenbauer I really like the idea you proposed to hook directly the Listing model.
Pimcore team (@brusch @weisswurstkanone) what do you think about it ? It seems you are ok for my first try directly on controllers but i can try to change it for Listing model instead if you think it is better.
Or maybe hooks for Listing model could be other new hooks added to those ones (i can make a separate PR for this one in this case) ?

fashxp:
Ad injecting directly: 
Since the controllers are all auto registered and auto wired, it should be sufficient to add the event dispatcher as action argument (`EventDispatcherInterface $dispatcher`). So you can avoid the static method call. In other places like the models this would need extensive refactoring of Pimcore, that's why there the static method call is used. 

Ad hooking directly into listing models:
We don't like the idea so much and wouldn't merge it - at least not for now. We see too much potential problems with it since such hooks may cause unexpected behavior (up to unwanted data loss) and we think that the use cases between listing in Pimcore Admin and using object listings are quite different ones - also the available context information like logged in user etc. can be completely different.
 
If you have such special filter requirements for object listings why not add an additional custom layer with all your business logic and use that one in your application?

Yivan:
@maff @weisswurstkanone @fashxp   
I just updated this PR, resolving conflict because of new adminJson and getAdminUser methods which should now be used.
I added DI for event dispactcher like asked.

**I hope all is ok now and can be merged : )**
Thanks.

p.s: @fashxp thanks for the details for DI. To answer to your question "_If you have such special filter requirements for object listings why not add an additional custom layer with all your business logic and use that one in your application?_" : only the Pimcore back-end GUI is going to be used, there is not dedicated application for this project. So we need to be able to filter directly object in Pimcore back-end grid/tree. Developping a all new back-end listing panel will be a lot more complexe and we will loose the  great native tree/grid functionnality (sharing, operator, export, etc...). So just changing filter is the faster method i think and the best for enduser because he can use all the powerfull feature of Pimcore GUI.
Or is it another way than events to overload/overide those listing ? Extending DataObject\Listing directly to overload methods don't seems possible actually as only custom Pimcore object are extendable using DI, not the core model class themselves. And override admin controller action isn't possible too i think ?

Yivan:
Side bug:
Maybe a separated issue should be open, because it seems some others event dispatch (not related to this PR) don't use DI.
For instance here https://github.com/pimcore/pimcore/blob/14787874ce20f690ae144bfb7d3e5835f97a38eb/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/DataObjectHelperController.php#L1559
It use the static method, but the eventDispatcher is injected in the action... MAybe there are other core files like this one which need to be revised.

fashxp:
@Yivan : but with these events in the admin controller you should be perfectly able to apply your custom filters? Or did I miss something? With the quote I was referring to the idea of adding an event directly to the listing classes - what we think can be really dangerous.

Yivan:
@fashxp Yes with those events in admin controller, it will be possible to filter elements listing easily to add some custom rules which are not possible with user permissions. It fill totally my need.
No you don't miss something, all is ok. I understand the risk if we had directly added events to model.

So this PR is ready to be merged as it is now : )

Maybe a separated issue need to be opened for the tag issue if i was right, because there can be a security problem (please see the paragraph `On special note for tags` in my very first post in this PR):
- it seems there is no permission control during tag propagation to objects
- tagging object seems linked to the "view" permission, is it normal ? Maybe it should be added that tag use this permission to the permission page of the documentation.

PR #2307, opened 2017-12-04
[IMPORT] Refactor import to be wired via DI

maff:
* Resolvers are services implementing ResolverInterface which are tagged
  with `pimcore.import.resolver` and an id
* Operators and values are run-time objects, but built via factories
  which can inject dependencies on demand. Registering their factories
  is done via `pimcore.import.operator_factory` and
  `pimcore.import.value_factory` and an id attribute
* The import service holds service locators for resolvers, operator and
  value factories and uses those to build the needed objects


pimcore/models/DataObject/ImportResolver/Filename.php
```diff
@@ -88,33 +87,29 @@ public function resolve($parentId, $rowData)
                 $object->setKey($objectKey);
             }
         } else {
-            $this->getAlternativeObject($intendedPath, $parent, $className);
+            // TODO is an assignment missing here?
+            $this->getAlternativeObject($prefix, $intendedPath, $parent, $className);
```
weisswurstkanone: $object = ...

pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/DataObjectHelperController.php
```diff
@@ -183,7 +186,7 @@ public function importExportConfigAction(Request $request)
             $gridConfig = GridConfig::getById($gridConfigId);
             $user = $this->getUser();
             if ($gridConfig && $gridConfig->getOwnerId() != $user->getId()) {
-                $sharedGridConfigs = $this->getSharedGridColumnConfigs($this->getUser(), $gridConfig->getClassId());
+                $sharedGridConfigs = $this->getSharedGridColumnConfigs($this->getUser(), $gridConfig->getClassId()); // TODO parameter searchType is missing
```
weisswurstkanone: wanted behavior ... we wanna offer all configs (i.e. grid, variants, childs, ...), not just the grid config

pimcore/models/DataObject/ImportResolver/Fullpath.php
```diff
@@ -19,47 +19,45 @@
...
-     */
-    public function resolve($parentId, $rowData)
+    public function resolve(\stdClass $config, int $parentId, array $rowData)
     {
-        $fullpath = $rowData[$this->getIdColumn()];
-        $object = DataObject::getByPath($fullpath);
-        if (!$object && $this->createOnDemand) {
-            $keyParts = explode('/', $fullpath);
+        $createOnDemand = $config->resolverSettings->createOnDemand;
+        $createParents  = $config->resolverSettings->createParents; // TODO can this be removed?
```
weisswurstkanone: yes

PR #2308, opened 2017-12-04
Fix documentation

LosHawlos:
Small fix for the documentation:

In a controller action the exception need to be thrown - not returned.


fashxp:
thx very much

PR #2316, opened 2017-12-06
fix for caching Staticroutes objects which doesn't exist

deltatag:
### Example

```php
$route = Staticroute::getById(123); // does not exist -> $route is null
$route = Staticroute::getById(123); // $route is empty object as it is saved in runtime cache
``` 

### Solution

This pull request provides a fix.


fashxp:
thx very much

PR #2327, opened 2017-12-08
Modifing unit abbreviation length

mbolka:
Currently in Pimcore database in table **quantityvalue_units** column **abbreviation** has length of 10 chars.

Maximum length of unit in standard Etim classification is 15 characters, so importing etim classification ends up with incorrect, cut units.
(All units can be found here: https://www.etim-international.com/downloads/category/8-model-releases)
  
## Fixes Issue #
Too short abbreviation column length.
Change of length of **abbreviation** column from 10 characters to 20 characters.

### DIFF
--- a/update-scripts/154/postupdate.php
+++ b/update-scripts/154/postupdate.php
@@ -1,0 +1,8 @@
+<?php
+
+$db = \Pimcore\Db::get();
+$db->query('
+ALTER TABLE `quantityvalue_units` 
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
This has to be considered in install.sql as well, right ?

Also note that the next build number is 155 (update folder has to be renamed).

Could you please provide a new PR ?

weisswurstkanone:
build 161

Issue #2346, opened 2017-12-14
Environment test prevent logging in the backend.


Yivan:
By using the default test environment, i cannot log in the Pimcore backend.

Information:
- Debian 9, Pimcore 5.0.2 stable
- I got no error message (authentification failed), it just bring back to the login screen without any message.
- By looking in the log i got:
`security.DEBUG: Access denied, the user is not fully authenticated; redirecting to authentication entry point. {"exception":"[object] (Symfony\\Component\\Security\\Core\\Exception\\AccessDeniedException(code: 403): Access Denied. at .../vendor/symfony/symfony/src/Symfony/Component/Security/Http/Firewall/AccessListener.php:70)"} []`
- session directory has all the permissions to be written
- deleted manually /var/cache/test directory and clearing the Pimcore cache/cache_tags table change nothing

Actual solution:
If i comment the session handler
```
    session:
        storage_id: session.storage.mock_file
```
in pimcore\lib\Pimcore\Bundle\CoreBundle\Resources\config\pimcore\config_test.yml
i can log in the backend.

Why there is a problem with the session.storage.mock_file ? Is it a bug ? I don't find any information on this on the Pimcore documentation.

maff:
Hi,

the test environment isn't meant for logging into the admin via browser but only for testing (unit/integration tests). Therefore the session is a mock session which doesn't actually start a session and won't work with browsers. See https://github.com/symfony/http-foundation/blob/master/Session/Storage/MockFileSessionStorage.php.

Yivan:
@maff Thanks for the ultra fast answer!
I undestand better now.

Is it documented somewhere ? If not I think you should explain this here that "test" is some special environement presseted: https://pimcore.com/docs/5.0.x/Development_Documentation/Deployment/Multi_Environment.html
Will help people comming from Pimcore 4 where test env could be used like any other environnement whereuser intend to log.

So we need to create a "testing" environment for instance.
For reference, when trying to set a new environment (by copying de dev environment files), some problems occurs => https://github.com/pimcore/pimcore/issues/2265

maff:
No not explicitely in Pimcore as it follows the Symfony convention (see https://github.com/symfony/symfony-standard/blob/3.4/app/config/config_test.yml). It's also the environment which is used to run Pimcore's test suite on Travis.

Issue #2265, opened 2017-11-23
[Pimcore 5.0.2 Stable] Unable set a PIMCORE_ENVIRONMENT which is different than default ones


Yivan:
Downlading the last stable on the web site and running installation bring an error:

![error1b](https://user-images.githubusercontent.com/9025839/33187041-1e147ce4-d08f-11e7-83bd-465e6527a6a2.png)

Environnement is ok (i was able to install previous Pimcore 5 versions). Check requirement is ok (some warning, only Composer is missing).

Here the detail (if i refresh the installation page i got this page):

![error2](https://user-images.githubusercontent.com/9025839/33177547-3e212cde-d063-11e7-8e7e-cfdc4e01838f.png)

Just specificity of my environment:
Our ENV name is `development` (and not `dev`) so before install I copy config_dev.yml as config_development.yml. I don't know if it could be related.

Yivan:
I tested by renaming our dev environement to `dev` and install working ok.

So this bug appears only when PIMCORE_ENVIRONMENT is different than default ones.
I changed the GitHub issue title.

Yivan:
In fact the same problem occurs if you create a new environment in existing installation of Pimcore 5.0.2 stable.
I will update the title of the issue.

I managed to make it work a little further (bypassing the monolog error), by creating a file named "development.yml" (by copying dev.yml) here too:
dev-phileass.fr\pimcore\lib\Pimcore\Bundle\CoreBundle\Resources\config\pimcore
It seems this directory contains environment file and should be mentioned here for people setting new environments:
https://pimcore.com/docs/5.0.x/Development_Documentation/Deployment/Multi_Environment.html

I get now another error : (

```
 (2/2) FileLoaderLoadException
There is no extension able to load the configuration for "web_profiler" (in .../pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/config/pimcore/config_dev.yml). Looked for namespace "web_profiler", found "framework", "security", "twig", "monolog", "swiftmailer", "doctrine", "sensio_framework_extra", "cmf_routing", "pimcore", "pimcore_admin", "app" in .../pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/config/pimcore/config_dev.yml (which is being imported from ".../pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/config/pimcore/config_development.yml").
in FileLoader.php (line 177)
```

```
(1/2) InvalidArgumentException

[...]

PR #2353, opened 2017-12-15
Doc for setting up a new environment name

Yivan:
This PR is to give clues how to configure a new environment name.

Even if it follow symfony standard, i think it is worth mentionning this in the documentation to help people (and those comming from Pimcore 4 can get trapped in the test environment which has a special signification in Symfony).

I state too that some environment file need to be created in `pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/config/pimcore/` which seems is not documented.

This PR was done because of those issues relating some difficulties to set a new environment:
https://github.com/pimcore/pimcore/issues/2265
https://github.com/pimcore/pimcore/issues/2346

### DIFF
--- a/doc/Development_Documentation/21_Deployment/03_Multi_Environment.md
+++ b/doc/Development_Documentation/21_Deployment/03_Multi_Environment.md
@@ -74,7 +74,7 @@ use Symfony\Bundle\WebProfilerBundle\WebProfilerBundle;
 				$collection->addBundle(
 					new SensioGeneratorBundle(),
 					80, // priority
-					['staging'] // will use the staging config files of the bundle
+					['staging'] // the bundle will only be loaded in the staging environment
 				);
 				// PimcoreGeneratorBundle depends on SensioGeneratorBundle
 				$collection->addBundle(
--- a/doc/Development_Documentation/21_Deployment/03_Multi_Environment.md
+++ b/doc/Development_Documentation/21_Deployment/03_Multi_Environment.md
@@ -40,15 +40,79 @@ You can need to configure a new environment name different than the existing one
 > **Note:** The default `test` environment should only be used for running test suite (like Travis).
 > It use the session.storage.mock_file which fakes sessions (as consequence, the administrator cannot log in Pimcore for instance)
 
-To configure a new environment, you need to manually create a YAML config file for the project in two places (you can start by copying config_dev.yml for instance in each of those):
+To configure a new environment, you need to manually create a YAML config file for the project in:
 
 ```
 app/config/pimcore/
-pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/config/pimcore/
 ```
 
-If you use some specific bundles, you need to activate them in `app\AppKernel.php`.
-For instance for an new environment called `staging` (copied from `dev` environment YAML config) you should add in the method `registerBundlesToCollection`:
+As a starting point, you can inspire your new YAML config file from the existing YAML config files for dev and prod environment in `app/config/pimcore/` and `pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/config/pimcore/`.
+
+For instance, the most minimal config, reproducting prod environment, will be:
+
+```
+imports:
+    - { resource: config.yml }
+
+monolog:
+    handlers:
+        main:
+            type:         fingers_crossed
+            action_level: error
+            buffer_size: 2000
+            handler:      nested
+        nested:
+            type:  stream
+            path:  "%kernel.logs_dir%/%kernel.environment%.log"
+            level: debug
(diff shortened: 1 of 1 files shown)


doc/Development_Documentation/21_Deployment/03_Multi_Environment.md
```diff
@@ -33,16 +33,62 @@ var/config/system.php
...
+You can need to configure a new environment name different than the existing ones, which respect Symfony convention: `dev`, `test` and `prod`.
+
+> **Note:** The default `test` environment should only be used for running test suite (like Travis).
+> It use the session.storage.mock_file which fakes sessions (as consequence, the administrator cannot log in Pimcore for instance)
+
+To configure a new environment, you need to manually create a YAML config file for the project in two places (you can start by copying config_dev.yml for instance in each of those):
+
+```
+app/config/pimcore/
+pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/config/pimcore/
```
kkarkus: @Yivan does it mean I have to create config inside core dir?

mgomse: Changes in the core isn't good idea IMHO. Isn't it possible to do it in some other way?

dpfaffenbauer: Shouldn't be necessary at all, ```app/config/pimcore``` or ```src/AppBundle/Resources/config``` should be totally enough

maff: @Yivan the only necessary file is `app/config/config_<environment>.yml`. Please do not suggest adding/changing core files. IMHO the previous description was fine - I'd just add the note regarding the test environment.

Yivan:
I totally agree, it is strange to have to change core files, but it was needed when i tried (see the monlog error here https://github.com/pimcore/pimcore/issues/2265 please).
I will try again with the last version to see if this problem always occurs.

Maybe there is some special configuration flags to add in the new Yaml file created for the new environment (`app/config/config_<environment>.yml`) so no more monolog error is triggered ?
Just copying `app/config/config_dev.yml` to `app/config/config_mynewenv.yml` don't work.

Maybe a file config_env.yml.sample could be added as a starting point ?

Yivan:
I updated the documentation, so there is no more need to create yaml file in the core bundle.
Instead i added some exemple/hints, so developper can create new environment like dev or prod as starting base.

If all is ok, this can be merged, thanks.

doc/Development_Documentation/21_Deployment/03_Multi_Environment.md
```diff
@@ -33,16 +33,126 @@ var/config/system.php
...
+				new DebugBundle(),
+				new WebProfilerBundle(),
+				new SensioDistributionBundle()
+			], 80);
+			// add generator bundle only if installed
+			if (class_exists('Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle')) {
+				$collection->addBundle(
+					new SensioGeneratorBundle(),
+					80, // priority
+					['dev'] // will use the dev config files of the bundle
```
maff: This means the bundle will only be loaded in the `dev` environment and should be changed to `staging` in your example (comment should be removed or changed).

maff:
@Yivan I restructured the core config files in #2400, so they are now easily includable from a custom env config. This avoids having to replicate all the config for a specific environment (see the comment on the PR). Could you update this PR accordingly?

maff:
@Yivan any update on this?

Yivan:
@maff I will soon update my PR before the end of the week, is it ok ? Thanks.

Yivan:
@maff Code has just been updated, if all is ok you can merge it! Thanks.

doc/Development_Documentation/21_Deployment/03_Multi_Environment.md
```diff
@@ -33,16 +33,64 @@ var/config/system.php
...
+				new DebugBundle(),
+				new WebProfilerBundle(),
+				new SensioDistributionBundle()
+			], 80);
+			// add generator bundle only if installed
+			if (class_exists('Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle')) {
+				$collection->addBundle(
+					new SensioGeneratorBundle(),
+					80, // priority
+					['staging'] // will use the staging config files of the bundle
```
maff: @Yivan could you change this to something like `the bundle will only be loaded in the staging environment`? It does not actually change how config files are handled, just sets a filter when loading the bundle.

Yivan:
@maff change on the comment has been done, all should be ok now.

maff:
Thanks!

PR #2362, opened 2017-12-18
Symfony 3.4

maff:
Upgrade to Symfony 3.4 and remove deprecations.

**NOTE**: there are still some deprecations in admin controllers which will be addressed in a later PR.


dpfaffenbauer:
Doesn't seem to be finished at all. I am getting a lot of deprecation warnings:

- The "Pimcore\Document\Renderer\DocumentRenderer" service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead
- Auto-registration of the command "Pimcore\Bundle\CoreBundle\Command\MysqlToolsCommand" is deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based service discovery instead: 1x
- Auto-registration of the command "Pimcore\Bundle\CoreBundle\Command\InternalVideoConverterCommand" is deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based service discovery instead: 1x
- Auto-registration of the command "Pimcore\Bundle\CoreBundle\Command\InternalUpdateProcessorCommand" is deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based service discovery instead: 1x
- Auto-registration of the command "Pimcore\Bundle\CoreBundle\Command\InternalUnicodeCldrLanguageTerritoryGeneratorCommand" is deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based service discovery instead: 1x

a lot more than that. Now shows that \Pimcore::getContainer is very bad... Pimcore now needs to make services, which are used by getContainer->get() to be public.

You can find more here: https://travis-ci.org/coreshop/CoreShop/jobs/318607090 (forget about the CoreShop warnings, but Pimcore throws a lot of deprecations as well)

maff:
Hey,

thanks for the input - in fact I clicked the merge button too fast and we're still on it. Will be updated gradually on master - next build will be including all 3.4 changes. However, those are just warnings relevant for 4.0 - we'll try to keep them to a minimum, but Pimcore already works with 3.4 and the merged changes.

dpfaffenbauer:
Ok. Regarding Warnings: Right, I know, but still: Keep them as minimal as somebody will need to change that anyway in order to work with symfony 4 once pimcore decides to upgrade.

PR #2371, opened 2017-12-20
fallback translation in combination with caseinsensitive website translations bugfix

markus-moser:
Fallback languages did not work correctly in combination with case insensitive translations enabled.

for example:

Translation key is "myTranslation"

If the translation key is empty but a "mytranslation" TK exists with a translated value in one of the fallback languages the translation key was printed instead of the fallback value.

I'm not sure if my solution is a good one - escpecially the part where the original locale catalogue get's updated but I have no idea how to solve it differently.


fashxp:
thx

PR #2378, opened 2017-12-21
Added "API Key" as a configuration option for PDFreactor

BlackbitNeueMedien:
My PDFreactor instance in set up to require authentication via API keys. I've added a corresponding option to the Web2Print config panel.


fashxp:
thx very much. please only update `pimcore/lib/Pimcore/Bundle/CoreBundle/Resources/translations/en.json` - the other files are updated from poeditor. could you update the pull request?

BlackbitNeueMedien:
Sure. I've reverted the changes to the other files.

maff:
Thanks!

PR #2379, opened 2017-12-21
Update symfony doc urls (now points to 3.4 version).

kkarkus:
Now every reference to symfony docs points to 3.4 version.


fashxp:
thx very much

PR #2388, opened 2017-12-28
 Bugfix for device detector

deltatag:
Setting device type by request (e.g. Pimcore preview feature) was ignored due to existing cookie. So it was never able to change the device type once changed.


brusch:
Thanks!

PR #2389, opened 2017-12-28
Thumbnail-Action: Return NotFound-Response if asset does not exist

podarcis:
An action has to result a response, in this case a 404 fits best.

Response before: LogicException - The controller must return a response (null given). Did you forget to add a return statement somewhere in your controller?
Response after: NotFoundHttpException - Asset with ID '12345' doesn't exist


weisswurstkanone:
many thx

PR #2428, opened 2018-01-10
Use class attribute instead of static url in pimcore.object.tree class

mbolka:
## Fixes Issue #
In method init variable class attribute treeDataUrl is not used (is uses just static url)


maff:
Thanks!

PR #2430, opened 2018-01-10
Pimcore Community Forums added to the resources

mgomse:
Nothing important, but good to have all places in one resource list :smile_cat:


maff:
Thanks!

PR #2432, opened 2018-01-11
Deeplink: Added helper and respect server port

podarcis:
This change consolidates the deeplink generation (helper).

Further more, it now respects the server port so links for Pimcore installations on non-standard ports are correct, e.g.:
https://example.com:81/admin/login/deeplink?object_12345_object


maff:
Thanks!

Issue #2211, opened 2017-11-07
[Objects] Configuration Option for Date/Time Types to Configure DB Datatype `bigint` or `DATE` / `DATETIME`


fashxp:
now, it is always `bigint` - `DATETIME` is for some usecases a better choice ...

PR #2433, opened 2018-01-11
Properly read the Date type of a DataObject as DATE, not DATETIME.

LosHawlos:
When a Date Field of a DataObject ist configured to be stored as DATE in the database, it currently doesnt read back the proper value.
getDataFromResource does the format conversion based on the wrong columnType.
This fixes a bug introduced for feature request #2211 in commit 23f5677a435c6058e3270ee99a94bd1f13432387


weisswurstkanone:
thx

PR #2438, opened 2018-01-11
fix for "get batch jobs" action

kkirsz:
This PR resolves problem with batch assignment. Due to the lack of parentheses in the SQL query, filters are not included when calculating objects for editing.


weisswurstkanone:
many thanks!

Issue #2447, opened 2018-01-15
Interface of object will crash when a 'Objects with Metadata' select field has no data


jorisros:
### Expected behavior 
When you add the field 'object with Metadata' to a object, and you don't put comma separated value, you expect when you open the object a empty dropdown
### Actual behavior  
The interface will crash when you open a object that contains a field  'object with Metadata' and shows the following error.
<img width="1236" alt="screen shot 2018-01-15 at 07 27 07" src="https://user-images.githubusercontent.com/752283/34929891-c0684344-f9c6-11e7-89da-4b78918ede7c.png">

### Steps to reproduce  
- Create a object in classes
- Add a objects with metadata field to the object
- Add a row with with select as type
- Save the object
- Open a object of the above class
- The error will occur (TypeError: this.fieldConfig.columns[i].value is undefined)

weisswurstkanone:
already fixed via https://github.com/pimcore/pimcore/pull/2448

PR #2448, opened 2018-01-15
[#2447] Check if the value exists before splitting

jorisros:
#2447 This fixes the problem when opening a objectsMetadata


weisswurstkanone:
thx

PR #2449, opened 2018-01-15
Update system requirements to meet PHP 7.1

Cruiser13:
Since dependencies will likely need PHP 7.1 often. Updating Pimcore 5.0.2 to 5.1 will fail with PHP 7.0 because of dependencies using 7.1 syntax so we should probably update the docs.

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/01_System_Requirements.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/01_System_Requirements.md
@@ -9,9 +9,9 @@ For production we highly recommend a *nix based system.
 - Nginx
 
 
-### PHP >= 7.0
+### PHP >= 7.1
 Both **mod_php** and **FCGI (FPM)** are supported.  
-Please note that the usage of **PHP 7.1 is strongly recommended** as future versions of Pimcore 5 will require PHP 7.1. 
+
 
 #### Required Settings and Modules & Extensions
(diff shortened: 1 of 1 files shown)


maff:
Do you have an example of dependencies needing 7.1? Composer should take care of installing the latest dependency matching the platform and PHP version.

Cruiser13:
I believe we had issues with Ocramius/PackageVersions but I updated the target system to 7.1 in the meantime so I'm not sure. Composer did throw an error.

brusch:
Is it then a Pimcore issue? Don't think so... :|

maff:
I just tested with a fresh installation: composer is able to install all dependencies on PHP 7.0. When running `composer update` on 7.1 afterwards it's updating `doctrine/inflector` from `v1.2.0` to `v1.3.0` but besides that the packages stay the same. If there is a dependency which does not work on 7.0 we need to fix that as we currently guarantee compatibility with 7.0.

Did you switch back from 7.1 to 7.0 somewhere during the update or migrate between servers? I can imagine running into problems when using a cached Symfony container (`var/cache`) and/or generated files in `vendor/` as generated files on a newer version may generate code targeted to the newer version (e.g. using features which aren't available on 7.0). In that case it should be enough to clear and re-generate those files (`bin/console cache:clear`, `composer update`) to get a working state.

Closing for now as it is no general issue.

PR #2475, opened 2018-01-22
check field editable in completed grid cell edit

solverat:
This PR fixes a nasty js error in grid cell editing. steps: to reproduce on demo:

- open folder with objects
- click "column configuration"
- add a "Operator Trimmer" and any field as child
- save
- click on a cell from the just created operater, a message appears "This element cannot be edited"
- now click on the refresh button on bottom which will trigger an error:

bildschirmfoto 2018-01-22 um 12 42 34

After that you need to reload the whole folder element.

By the way: the `startEdit` method is deprecated in extjs since 5.5:
http://docs.sencha.com/extjs/6.0.2/classic/Ext.grid.plugin.CellEditing.html#method-startEdit


weisswurstkanone:
Thx, I guess we will look at the deprecation issue as soon as a ExtJS update gets relevant again. I am sure there are quite a few more :-)

Issue #2089, opened 2017-10-11
[Install Command] if memory-limit for CLI is set to -1 install fails


fashxp:
because it thinks memory limit is to low...

fabfischer:
I added a fix to address this issue:
https://github.com/pimcore/pimcore/pull/2490

maff:
Closed in #2490.

PR #2490, opened 2018-01-23
fixed: if memory-limit for CLI is set to -1 install fails

fabfischer:
[Install Command] if memory-limit for CLI is set to -1 install fails #2089
https://github.com/pimcore/pimcore/issues/2089

## Fixes Issue 
the install command failed if the php.ini setting for memory_limit was set to '-1'. 
I added a condition to convert the value of the memory_limit only if it is not set to -1

## Additional info  
see issue: https://github.com/pimcore/pimcore/issues/2089


maff:
Thanks!

PR #2495, opened 2018-01-23
Added Event pimcore.mail.preSend event

ckogler-elements:
public function onPreSendMail(ElementEventInterface $e)
    {
        $arguments = $e->getArguments();

        //unset($arguments['mailer']); <- do completely custom sending
        //$arguments['mailer'] = new \Swift_Mailer(...) <- custom implementation
        $e->setArguments($arguments);

        //change mail settings
        $mail = $e->getMail();
        $mail->setFrom('mofified@example.com');
    }


Yivan:
Nice event thanks! Maybe a postSend event could be added so we have all the events which could be need pre and post ?

PR #2501, opened 2018-01-24
Allow change a sorting way of items in object tree

michalfilik:
This pull request provides improvement to sorting items in Object Tree. It allows to choose default sorting way for children of specific folder/object. Items can be sorted manually (by index column - set by drag&drop of elements) or alphabetically (by key column).

pimcore-sorting-objects

## TODO

- [ ] Prepare db-update script:

```sql
ALTER TABLE objects ADD o_childrenSortBy ENUM('key','index')  CHARACTER SET utf8mb4  COLLATE utf8mb4_general_ci  NULL  DEFAULT NULL  AFTER o_className;
ALTER TABLE users_workspaces_object ADD sortBy TINYINT(1)  UNSIGNED  NULL  DEFAULT '0'  AFTER properties;
```


fashxp:
very cool!!! how does it work with the show-in-tree functionality when the tree starts paging?

pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/ElementController.php
```diff
@@ -463,6 +463,7 @@ public function typePathAction(Request $request)
             $data['index'] = $element->getIndex();
         } else {
             $element = DataObject::getById($id);
+            $data['index'] = $element->getIndex();
```
weisswurstkanone: index or idx ?

michalfilik: It's used [here](https://github.com/DivanteLtd/pimcore/blob/f1ecfa4f688540c2fd4f2648794fcf432f461f6d/web/pimcore/static6/js/pimcore/treenodelocator.js#L126). "Index" key is in the same convention like for documents.

weisswurstkanone:
please have a look at my comment here https://github.com/pimcore/pimcore/issues/2530

thx!

weisswurstkanone:
Hi,

will this work without the TODOs mentioned above ?

In addition:
- changes to **install.sql** are probably missing ?
- do we really need sortBy workspace permission anyway ? If yes, do we have all translation keys ?

Could you prepare a update script? I'll modify the build number if necessary.

Thanks!

weisswurstkanone:
One more thing ... adding a column to the objects table will probably break with the data.sql dumps for the various install profiles as they rely on the correct column counts.

Example: 

install-profiles\demo-basic\dump\data.sql

INSERT INTO `objects` VALUES ('3','2','object','lorem-ipsum','/news/','0','1','1368613483','1382958769','0','0','2','news');

michalfilik:
Hi @weisswurstkanone,
1. It will not work without mentioned TODO (especially without first one).
2. Changes to **install.sql** 
- i will provide update.
3. Permissions for workspace sortBy 
- I've prepared them as a full sorting feature, but please decide if they are necessary at all.
4. Install profiles - yes, this will break them. There is 2 ways to fix that:
 - use INSERT INTO with specified columns (without sortBy - default value will be used)
`INSERT INTO table_name (column1, column2, column3, ...)
VALUES (value1, value2, value3, ...);`
- add NULL value to each values of Objects rows

weisswurstkanone:
please have a quick look at
https://github.com/pimcore/pimcore/blob/master/install-profiles/demo-basic/dump/create-dump.php#L67

the script is used to create the dumps. maybe there is an easy way to adapt it (by including the column names)

michalfilik:
@weisswurstkanone I've commited mentioned changes. Please review them and let me know if everything is complete now :)

michalfilik:
@fashxp @weisswurstkanone any news about this PR ?

piotrecky:
I think it is worth adding additional sorting by creationDate and modificationDate.

Issue #1780, opened 2017-07-31
[Pimcore5] pimcore-bundles with dependencies to symfony-bundles


dpfaffenbauer:
Pimcore-Bundles which have dependencies to other bundles need to be activated manually in AppKernel. Therefore, the ExtensionManager isn't enough to install certain Bundles. Are you planning any solution for this problem?

I already thought about following solutions:
 - Symfony Flex, but symfony flex only supports type "symfony-bundle" not "pimcore-bundle", therefore not a real solution at all
 - define bundle dependencies within composer.json like:
```json
"extra": {
    "pimcore": {
      "bundles": [
        "MySuperbBundle\\SuperbBundle"
      ],
      "dependencies": [
          "FOS\\RestBundle\\FOSRestBundle"
      ]
    }
  },
```

OR

```json
"extra": {
    "pimcore": {
      "bundles": [
        "MySuperbBundle\\SuperbBundle": {
            "dependencies": [
                "FOS\\RestBundle\\FOSRestBundle"
            ]
        }
      ]
    }
  },
```

and the ExtensionManager takes care about loading dependent bundles as well. (As well as checking if a bundle has already been registered in Kernel).

This case relates somehow to #1778, cause install-profiles could enable and install Pimcore-Bundles automated as well.

dpfaffenbauer:
Relates to https://github.com/pimcore/pimcore/issues/1750 as well.

dpfaffenbauer:
I played around with symfony/flex a little bit, actually, it really solves this issue. Flex looks for all types with symfony-bundle and auto-generates recipes for them (if there is no recipe in the two repositories). If there is a recipe for a package, it doesn't matter which type it is. So it could be a pimcore-bundle then as well.

If you have a simple dependency on a bundle (and don't need any configuration), flex already registers the bundle for you. Therefore 👍 

Since flex is not gonna be a stable release until end of November (gets released with symfony 4), its currently just gonna be dreams for the future. 

But: one thing to mention is the new symfony 3.3/symfony 4 file structure. Wouldn't it be a good idea for pimcore to adapt to the new structure before stable release? This would make migration to symfony 4 a lot easier.

dpfaffenbauer:
regarding https://github.com/pimcore/pimcore/pull/1778#issuecomment-319677548 (wasn't sure where to put this comment, information kinda gets scattered around, so I'll post this here)

**Flex**
[...]

PR #2517, opened 2018-01-26
Dependency handling for Bundles

maff:
As described in #1780, dependencies between bundles are a common issue which currently needs to be solved manually. As we already have the `BundleCollection`, adding a way to handle dependencies on a simple level is possible without a performance hit coming from dependency resolution. This change:

* Removes the exception which is thrown on the bundle collection when a bundle is added multiple times. Instead, subsequent adds are simply ignored. An already added item can never be changed, so the first item added wins. This also means that an item which is added from the core or from another bundle can be overridden on the collection by adding it manually before it is added from core/dependency. See the new `createBundleCollection` method on the kernel which can be used to add items to the collection early.
* Introduces a new `DependentBundleInterface` a bundle can implement. As soon as the bundle is added to the collection, the `registerDependentBundles()` method is called which allows the bundle to implicitly load dependencies. Order is resolved through the item priority.
* Dependencies of dependencies also work as the `registerDependentBundles()` method is called from the `add()` method on the collection which is always called to add an item. If a bundle adds a dependency, the dependency can again be a `DependentBundleInterface` adding further dependencies (see tests for an example). Circular reference loops are simply avoided by stopping as soon as the item to be added already exists on the collection (see first point).

@dpfaffenbauer want to give this a try for CoreShop's `RegisterBundleHelper` before we merge?

### DIFF
--- a/doc/Development_Documentation/20_Extending_Pimcore/01_Add_Your_Own_Dependencies_and_Packages.md
+++ b/doc/Development_Documentation/20_Extending_Pimcore/01_Add_Your_Own_Dependencies_and_Packages.md
@@ -52,9 +52,7 @@ class AppKernel extends Kernel
 }
 ```
 
- Internally, the `BundleCollection` will be ordered by priority, filtered by environment and returned as plain array in
-`registerBundles`. If you need full control over the registered bundles, you can override `registerBundles` and add your
-customizations on the resulting array.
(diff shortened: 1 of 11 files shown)


dpfaffenbauer:
:) hell yeah, this is amazing :) 👍 🍻 🎉 🍻

dpfaffenbauer:
tell me when your finished, I am down to test that with CoreShop and its bundles.

maff:
The feature is ready so far, I'll merge to master as soon as tests are green.

PR #2531, opened 2018-01-31
Bugfix - CSV-Export loses settings

woellundwoell:
CSV-Export loses settings for data inheritance and separator when number of exported items is more than 20.

**Expected behavior:**
When exporting csv data from grid view inheritance and separator character is considered in all exported data row.

**Actual Behavior:**
When exporting csv data from grid view inheritance and separator character is considered **only in the first 20 exported data rows**. In all other rows default separator is used and no inherited values are exported.


weisswurstkanone:
many thanks!

Issue #1112, opened 2016-12-08
New feature: move asset to a folder


Yivan:
Recently we needed to move some assets from a directory to another one. The actual "Advanced > Search and move" is great if you get only some of them, as you must click them one by one to add them to the move process. Maybe a "Add all" could be added but it will always be limited.

I think adding a move feature directly on the grid will be more powerfull.
For instance, here:
![moveto](https://cloud.githubusercontent.com/assets/9025839/21025486/f135643c-bd88-11e6-9873-419546c201cc.png)

We could add "Move all selected to" (after "Delete selected" for instance) on the right mouse click, and after choosing this option we could choose a directory where to move them (by dragging it for instance).
So the user will benefit of all the powerfull filters. He could not filter by tags/property as the "Search and move", but I think it is a special case. A all day case is more "I want to move all the asset from this directory to this one".  And with this new feature he could just use the List panel, check "Just direct child object", select them all and move them in just on click.

What need to be carefull, is to be sure than "select all" checkbox select even those wich are on the next paging page:
![moveto_all](https://cloud.githubusercontent.com/assets/9025839/21025787/11a8a142-bd8a-11e6-94b4-52ad07325d1f.png)

I hope this idea make sense to help day life of asset editors.
Thanks!

brusch:
I think this topic already well covered by #2532 and #3658
Of course it would be a nice addition to have that as well in the grids, so we'd definitely accept a PR for that 😉

PR #2532, opened 2018-01-31
Allow multiselect in search and move window.

kubaplas:

Enables mutliselect for search and move functionality.

## Additional info  
It seems that it was almost ready - only selModel in selector was not set.
Enabled for objects, assets and documents.


weisswurstkanone:
thanks a lot!

Yivan:
@weisswurstkanone @kubaplas 
Thanks!

Does it solve this one : https://github.com/pimcore/pimcore/issues/1112 ? (i.e. moving multiple assets selected, very useful if enduser need to re-arrange many assets).

weisswurstkanone:
yes and no. I would say that a grid gives better control over which assets you wanna select (grid filters)

PR #2534, opened 2018-01-31
fix for displaying classification store fields in the grid

kkirsz:
This PR resolves problem with displaying classification store fields in the grid.
1) If you have non localized classification store, fields are not displayed because default language is ignored in the method call.
2) If you have localized classification store, fallback languages are not used because fallback language is ignored in the method call.


weisswurstkanone:
The default language acts like a pseudo language which is worth the same as a real language (like en or de). It acts as a "fallback" language only in frontend mode. As said, it is treated as a real language when we are talking about grids or editing object data. Please use the grid column configurator to the select the "default" language setting just as you would do it for other languages.

image

This is probably related to 
https://github.com/pimcore/pimcore/pull/2536

PR #2536, opened 2018-01-31
fix for batch actions related with classification store fields

kkirsz:
It seems that classification store fields are not saved due the lack of the important code.


weisswurstkanone:
may be related to https://github.com/pimcore/pimcore/pull/2534

I don't see any issue here. See attached recording. Select the "default" pseudo language via the grid editor and then do a batch apply ...

[2018-02-01_08-04-01.zip](https://github.com/pimcore/pimcore/files/1684510/2018-02-01_08-04-01.zip)

Maybe I miss something ?

kkirsz:
Thanks for answer. You right. I should select default language. But there is one more issue. I checked it one more time and it only works with simple types (e.g. input). If you use some complex types (e.g. quantity value) changes won't be saved. This solution works with both. Could you verify it?

weisswurstkanone:
Many thanks! Batch update of complex types was definitely missing.

Build 194

PR #2538, opened 2018-02-01
Re-Introduce support for /app/startup.php and remove hardcoded environments

### LINKED ISSUE #2059: Remove hardcoded environments (dev, test)
Follow up to #2041. Currently, the `dev` and `test` environments are hardcoded in multiple locations, e.g. 

* Kernel loading generator bundles only in `dev`
* Kernel bootstrap (debug) and tasks like housekeeping or update having a hardcoded list of non-prod environments

Make these environments configurable somewhere as there might be more than the standard environments (e.g. `staging`).

### LINKED ISSUE #2520: Custom startup script `app/startup.php`
eg. for configuring trusted proxies ...

maff:
Hardcoded environments are now handled in a configurable EnvironmentConfig class which is initialized at boot, but can be overridden/reconfigured during the startup process.

EnvironmentConfig can either be reconfigured using setters or completely replaced by implementing a custom config class. Environments to register in development are still hardcoded on the kernel, but extracted into a method which can be overridden.

* Resolves #2520 
* Resolves #2059

### DIFF
--- a/doc/Development_Documentation/01_Getting_Started/03_Configuration.md
+++ b/doc/Development_Documentation/01_Getting_Started/03_Configuration.md
@@ -99,3 +99,25 @@ control over the used paths here.
 
 In contrast to the other constants, `PIMCORE_PROJECT_ROOT` can not be set via `.env` Pimcore doesn't know where to look
 for a `.env` file at this point.
+
+
+## Adding logic to the startup process
+
+If you need to execute code to influence Pimcore's startup process, you can do so by adding a file in `/app/startup.php`
+which will be automatically included as part of the bootstrap process. Specifically, it will be loaded after all other
+bootstrapping (loading the autoloader, parsing constants, ...) is done, but **before** the kernel is loaded and booted.
+This gives you the possibility to reconfigure environment settings before they are used and to configure the system for
+your needs. Examples:
+
+* Defining the [Trusted Proxies](http://symfony.com/doc/3.4/deployment/proxies.html) configuration on the `Request` object
+* Influencing the default [environment handling](../21_Deployment/03_Multi_Environment.md)
+
+```php
+<?php
+
+// /app/startup.php
+
+use \Symfony\Component\HttpFoundation\Request;
+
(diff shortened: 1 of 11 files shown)


Yivan:
@maff thanks a lot for this addition ! I liked so much this startup.php in Pimcore 4.

Maybe a part of the documentation need to be more explicits, as their is said:
```
Pimcore ships with sensible defaults on which environment to use in which case, e.g. by automatically using the `dev` environment if Pimcore is in debug mode and by automatically enabling the kernel debug flag for the `dev` and `test` environments. If you need to influence that behaviour (e.g. because you have additional environments) you can do so by changing the environment config during the [startup process](../01_Getting_Started/03_Configuration.md). For example, if you want to specify the
 +default environment to use when Pimcore's debug mode is enabled by no environment is explicitely defined:
```

It lead to some questions on the behaviour when activating debug mode:
When switching debug mode on, as it is said it force Pimcore environment to dev so it is bypassing the PIMCORE_ENVIRONMENT variable ? What about the system_xxx.php, as it contains database credentials it can break Pimcore, for instance if we are on prod and activate debug mode, will it use system_dev.php instead system_prod.php and so having wrong database credential ?

Maybe those points could be hinted more precisly on the documentation. I don't think debug mode should change environment (as many configuration can be based on it, even custom code for the project), but only activate debug tools without overriding PIMCORE_ENVIRONMENT.
Recently we got a strange behaviour on our staging server by activating debug mode, it seems some dev credential database where called and we needed to flush the caches manually (as we cannot no more go in admin GUI), i don't know if it is related.

maff:
The default environment with/without enabled debug mode are only set if there is no `PIMCORE_ENVIRONMENT` configured. So

* if you configured an environment somewhere, always that environment will be used, independently from Pimcore's debug mode
* if no environment is configured, Pimcore will use `prod` by default and `dev` if Pimcore's debug mode is active for the Visitor/IP

This PR doesn't change the behaviour of the system (is was as described since Pimcore 5.0), just moves things to a more centralized and configurable location as the environments were hardcoded throughout the system before.

Yivan:
Thanks for the explaination! All is great so.

PR #2539, opened 2018-02-01
Application logger: Apply search form on ENTER key

podarcis:
With this change the search filter form is submitted on ENTER key, too.
E.g. when typing in a search term in "Message". Before the filter could only
be applied by clicking on button "Search".

Ext JS reference: http://docs.sencha.com/extjs/6.0.2/classic/Ext.panel.Panel.html#cfg-defaultButton


weisswurstkanone:
thx!

PR #2542, opened 2018-02-02
Fix semantical error, when saving/updating DataObjects

aarongerig:
# Bugfix

In Version 5.1.2 you can't save / update DataObjects anymore, because of https://github.com/pimcore/pimcore/blob/master/pimcore/models/DataObject/Concrete.php#L146 which generated a Symfony annotation exception.

```log
[Semantical Error] The annotation "@params" in method Pimcore\Model\DataObject\Concrete::update() was never imported. Did you maybe forget to add a "use" statement for this annotation?
```


weisswurstkanone:
thx!

solverat:
same here, downgrading to 5.1.1 is not possible. do you plan a hotfix release for this?

brusch:
@solverat yes, 5.1.3 is on it's way ;) (today afternoon)

PR #2549, opened 2018-02-05
Bugfix cs boolean select

woellundwoell:
### General preparations

- Create a key definition of type "booleanSelect" in the classification store
- Assign the keyDefintion to a group
- Assign the classification store to an class attribute
- Create an object of this class
- Assign the above created group in the classification store attribute
- Assign all possible values ​​(No / Yes / Empty) to the booleanSelect attribute

### Expected behavior
- Correct values are saved
- Display the correct values ​​after reload
  
### Actual Behavior

- Only saves the "Yes value" correctly
- The values ​​"No" and "Empty" are not stored in the database
- Even after manually setting the correct values ​​in the database (-1 for "No", and NULL for "Empty"), the values ​​are not displayed correctly

### DIFF
--- a/pimcore/models/DataObject/ClassDefinition/Data/Classificationstore.php
+++ b/pimcore/models/DataObject/ClassDefinition/Data/Classificationstore.php
@@ -308,8 +308,15 @@ class Classificationstore extends Model\DataObject\ClassDefinition\Data
 
                         $dataFromEditMode = $dataDefinition->getDataFromEditmode($value);
                         $activeGroups[$groupId] = true;
+                        
+                        $forceCreateItem = false;
 
-                        $classificationStore->setLocalizedKeyValue($groupId, $keyId, $dataFromEditMode, $language);
+	                    // CS item must be created in case of "false" or "null" value in booleanSelect too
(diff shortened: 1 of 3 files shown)


weisswurstkanone:
I guess the proper solution would be to replace this magic line
        // treat value "0" nonempty
        $nonEmpty = (is_string($value) || is_numeric($value)) && strlen($value) > 0;

with the isEmpty() method from the respective datatype.

woellundwoell:
OK, but in case of the "empty" value in booleanSelect it will not work because isEmpty(...) returns true for value NULL and the value will not be stored in DB - and not displayed correctly afterwards. 
Should we change the isEmpty() function of the class pimcore/models/DataObject/ClassDefinition/Data/BooleanSelect.php to always return false?

Another question is how to get the needed function in the setLocalizedKeyValue method: It would work like this:
```
$keyConfig = Model\DataObject\Classificationstore\DefinitionCache::get($keyId);
$dataDefinition = Model\DataObject\Classificationstore\Service::getFieldDefinitionFromKeyConfig($keyConfig);
$nonEmpty = ! $dataDefinition->isEmpty($value);
```
but I think it is not a very good coding style ;)

woellundwoell:
Implemented a workaround see #2552

PR #2552, opened 2018-02-05
Implemented workaround for Bug in ClassificationStore booleanSelect

woellundwoell:
## General preparations

- Create a key definition of type "booleanSelect" in the classification store
- Assign the keyDefintion to a group
- Assign the classification store to an class attribute
- Create an object of this class
- Assign the above created group in the classification store attribute
- Assign all possible values ​​(No / Yes / Empty) to the booleanSelect attribute

## Expected behavior

- Correct values are saved
- Display the correct values ​​after reload

## Actual Behavior

- Only saves the "Yes value" correctly
- The values ​​"No" and "Empty" are not stored in the database
- Even after manually setting the correct values ​​in the database (-1 for "No", and NULL for "Empty"), the values ​​are not displayed correctly

## Workaround

As discussed we implemented a workaround for this bug and inserted a TODO for a final fix

### DIFF
--- a/pimcore/models/DataObject/ClassDefinition/Data/BooleanSelect.php
+++ b/pimcore/models/DataObject/ClassDefinition/Data/BooleanSelect.php
@@ -270,8 +270,7 @@ class BooleanSelect extends Model\DataObject\ClassDefinition\Data
      */
     public function isEmpty($data)
     {
-    	// For all 3 states never empty
-        return false;
+        return $data !== true && $data !== false;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pimcore/models/DataObject/ClassDefinition/Data/BooleanSelect.php
```diff
@@ -270,7 +270,8 @@ public function checkValidity($data, $omitMandatoryCheck = false)
      */
     public function isEmpty($data)
     {
-        return $data !== true && $data !== false;
+    	// For all 3 states never empty
+        return false;
```
weisswurstkanone: What is the reason for this change? Will this break the inheritance?

woellundwoell:
The reason is that the booleanSelect needs a data set (in DB) for each state - also for NULL / EMPTY.
If the isEmpty function returns true for the NULL value, the dataset will be deleted. This has the effect that the inheritance does not work and in the GUI  an empty Select is displayed instead of the "empty value".

weisswurstkanone:
thx!

PR #2558, opened 2018-02-06
Fix problem with new inheritance document

cwiecekpiotr:
## How to reproduce the issue ##
Go to you home document. Select "translation"->New document->Using inheritance and select "Language". If you select for example "De" you will get an error:
`Timestamp: Tue Feb 06 2018 11:54:18 GMT+0100 (CET)
URL: /admin/document/translation-determine-parent
Message: Type error: Argument 1 passed to Pimcore\Model\Document\Service\Dao::getTranslations() must be an instance of Pimcore\Model\Document, null given`
  
## Fixes Issue #
If document is a root, if yes then return root, else return parent.
Add condition for getTranslations parameter.


brusch:
Thanks!

Issue #2564, opened 2018-02-06
Missing requierements check - causes a FatalState during installation process 


Apke:
Hi, sry me again.

Thanks for latest merge. Next time i will not use a workaround docker-environment which was build for Symfony and was upgraded to php7 only Problems, but the good thing is, that we can fix an issue ;-)

# Bug Report

### Expected behavior 

- WebInstall
   - on pressing "install" with missing requierements like mysql-5.6.* with "fulltext"-index possibilities, i got an error for that and can re-use the installer
  
### Actual behavior  
  
i belive pcitures would say more then words:

#### Contiditon

![condition](https://user-images.githubusercontent.com/5233027/35883446-8734b242-0b87-11e8-8d17-7430b2f71bf7.png)

#### Error while Installing and the Result

![installer_install-action_without-pre-check-fails](https://user-images.githubusercontent.com/5233027/35883481-a280afec-0b87-11e8-8401-c579d1c2fd6c.png)

_and you don't want to see what happens on index-route_

#### Expected behavior 

![fixed_installer_install-action_without-pre-check-fails](https://user-images.githubusercontent.com/5233027/35883603-0aa5d6ba-0b88-11e8-97f7-a85f534954dd.png)

#### Short description

- no re-used db-checks like on `\Pimcore\Bundle\InstallBundle\Controller\InstallController::checkAction`
- i used the pimcore-stable version and i see, that in latest build some more checks are used, but not all and also is fulltext missing
   - fixed on PR #2565
- The State Producer is a early write of `System.php` which is used as *All Done Marker* but causes a redirecting to a half installed pimcore backend

### Steps to reproduce  

- use mysql version 5.5 without fulltext indexes
- go to installer page and press "install"

## Notes

The Problem is easy to find and happens because of two things, i create a Pull Request with a fast Fix with the biggest win, but personally i liked to do much more...i would extract the single methods and use the full potential of a `Check`/Validation-Model and mix it with a config-folder or array like `Validators`.

And i want to apologize my untypical code-style in this area. i thought that it is better to have a greater doc for multi-functional method. Also that method is not a good thing, i should had split that, but i was not sure where this is all used.

Sorry and Thanks

regards

Nick

PR #2565, opened 2018-02-06
BugFix - Installer - broken-web-installer

Apke:
## Fixes Issue #2564  

 +(add) re-usability to checker-method and convert it into a extractor
 - (fix) broken installer state if db requirements are missing
 - (removed) unneeded duplicates


maff:
@Apke thanks!

PR #2566, opened 2018-02-07
Fix for mass publication of objects.

kkirsz:
## How to reproduce the issue
Create class with at least one mandatory field and then create some objects. Leave mandatory field empty. Go to the grid and select created class. Choose column named "Published", select "Batch Assignment" option and then mark "Published" checkbox. Objects were published although mandatory field was empty.
  
## Fixes Issue #
If object is published then set ```$omitMandatoryCheck``` flag to ```false```.

There is one change in the class ```Pimcore\Bundle\AdminBundle\Controller\Admin\DataObjectHelperController``` (line 2307).


brusch:
Thanks!

PR #2571, opened 2018-02-08
Add note that CMF needs MariaDB

Cruiser13:


Cruiser13:
Hi @brusch you are right here but if you watch recent pimcore publications (for example on youtube), the CMF does play a big role for pimcore. One of our customers did recently setup a new server, based on pimcore system requirements, and once we deployed pimcore (with CMF), we found out that mysql was installed instead of maria db. So I thought a short notice would be nice at this point (not changing the requirements of course).

fashxp:
we are working on supporting mysql with the CMF too so that this restriction is not needed anyway

PR #2573, opened 2018-02-08
fix all params annotations

dpfaffenbauer:
This PR fixes all invalid "@params" annotations


solverat:
@weisswurstkanone do we have another 

Until this gets released, we're struggling with messages like this: "[Semantical Error] The annotation "@params" in method Pimcore\Model\Asset::update() was never imported. Did you maybe forget to add a "use" statement for this annotation?"

Do you plan another patch release or at least a build in near future?

PR #2594, opened 2018-02-14
ECommerce | FilterService | Filter Type| Number Range | upport for {{empty}} placeholder.

andreas-gruenwald:
In case that the keyword {{empty}} is passed a number range filter only has to be applied if there is a precondition.


fashxp:
thx

PR #2598, opened 2018-02-15
Checking if folder has children by searching only one child instead of all children

jakubderdziak:
Added LIMIT to query that should only checks if object has children.

### Steps to reproduce  
1. Create folder in Data Objects -> Home
2. Create 10 subfolders within folder created in 1.
3. Into each of subfolders generate 100 000 objects
4. Refresh page
5. Try to expand folder created in 1.


weisswurstkanone:
thx, maybe it makes sense to check other models as well ...

PR #2602, opened 2018-02-15
Speed up count of children

jakubderdziak:
  
  
First param of Method Dao::getChildAmount() can now be null. This will lead to two thinks:
1. Objects of all counts will be counted (same as setting [DataObject\AbstractObject::OBJECT_TYPE_OBJECT, DataObject\AbstractObject::OBJECT_TYPE_FOLDER, DataObject\AbstractObject::OBJECT_TYPE_VARIANT])
2. It will speed up drastically with large amount of data.

## Additional info  
It speeds up expanding folder in tree, if the folder has a lot of subelements (500k+), but there still place for improvement (expanding folder generates >100 queries!).


weisswurstkanone:
many thanks!

PR #2608, opened 2018-02-16
Unified name of the "client_params" option for ElasticSearch indices

BlackbitNeueMedien:
It was not possible to specify any "client_params" for Elasticsearch, since the option name differed in two pieces of code:

https://github.com/pimcore/pimcore/blob/b6ad9535cb95bc1ee4ae77fc27a9b1460c36f684/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/IndexService/Config/ElasticSearch.php#L89-L99


maff:
@BlackbitNeueMedien good catch, thanks! As changing the option name would be kind of a BC break (the option resolver would throw an exception on every installation configuring `es_client_params` as it doesn't know the option anymore), could you change this PR to use the value of `es_client_params` in `processOptions()` instead of renaming the option?

BlackbitNeueMedien:
Sure. I've updated the PR.

maff:
Thanks!

PR #2611, opened 2018-02-16
pimcore4 Delete Asset Syntax error

thureos:
Steps to reproduce:
Create 1 asset folder, then another one inside the first one, then try to delete the first one.


brusch:
Thanks!

PR #2613, opened 2018-02-16
Several fixes with deleting, copying.

thureos:
Create asset folder with contents
Try to:
1.  Delete it
2.  Copy it


brusch:
Thanks!

Issue #2609, opened 2018-02-16
[Assets] Add download option to tree context menu


weisswurstkanone:

jorisros:
Printscreen how it looks now.

<img width="282" alt="screen shot 2018-02-19 at 09 34 26" src="https://user-images.githubusercontent.com/752283/36368588-3fea97dc-1558-11e8-9642-056f7a1166b0.png">

weisswurstkanone:
Does this work for single files as well ?

jorisros:
No its only working on folders at the moment. Is it needed on singles files to download as zip?

fashxp:
would be nice, yes.

jorisros:
Added the download functionality to a single asset.
<img width="428" alt="screen shot 2018-02-19 at 22 12 38" src="https://user-images.githubusercontent.com/752283/36397451-22ecd68e-15c3-11e8-9c7f-0344e2b2e503.png">

PR #2615, opened 2018-02-19
[ASSETS] Add functionality for download a folder a zip #2609

jorisros:
Suggestion for issue #2609


brusch:
This functionality exists already here: 
https://github.com/pimcore/pimcore/blob/aea60550fae338e73bdb4435a0cc0b7ae370570e/web/pimcore/static6/js/pimcore/asset/folder.js#L313

So we should avoid to duplicate code here by moving the essential parts into a helper function: 
https://github.com/pimcore/pimcore/blob/31f7468561a846c019857ce60b8b4ea68dcbb123/web/pimcore/static6/js/pimcore/helpers.js

jorisros:
Good point, I will create a helper that can handle both of the calls

Issue #2591, opened 2018-02-14
Deployment:classes-rebuild fails after class is deleted


SilberMa:
# Bug Report
## Please make sure all the following boxes are checked before submitting your issue - thank you!
- [x] You're not able to fix the problem yourself and send us a pull request instead of an issue. 
- [x] There's no existing ticket for the same issue

### Expected behavior 

  Running bin/console pimcore:deployment:classes-rebuild finishes without errors and database is cleared.
### Actual behavior  
Before deleting DataObjects files (not in AdminBackend):
ERROR [console] Error thrown while running command “deployment:classes-rebuild”. Message: “Call to a member function save() on null” [“error” => Error { …},“command” => “deployment:classes-rebuild”,“message” => “Call to a member function save() on null”] []

After deleting DataObjects files (not in AdminBackend):
```php
In ClassesRebuildCommand.php line 82:

  [Symfony\Component\Debug\Exception\FatalThrowableError]
  Call to a member function getName() on null
```
### Steps to reproduce  

1. Create DataObject class with some vars
2. Delete definition_***.php
3. run bin/console pimcore:deployment:classes-rebuild

  
 This relates to https://talk.pimcore.org/t/deployment-classes-rebuild-fails-after-class-is-deleted/292

brusch:
This is intentional, we don't want to delete classes automatically for security reasons. 
If you want to delete a class automatically, you'd have to use a script (`$class->delete()`).

dpfaffenbauer:
is it really necessary? couldn't we just add a new flag '-d' where we intentionally delete classes when class definitions don't exist? I often run into this issue on development and quick deploying stuff onto the dev-server. I mean, it is everybodies risk to call rebuild-classes using the -d param the. what do you think?

brusch:
We could - PRs are welcome ;)

PR #2616, opened 2018-02-19
implement delete-classes option for pimcore:deployment:classes-rebuild

dpfaffenbauer:
This PR implements a delete-classes option flag to delete classes that don't have a class-definition file anymore. If the command is run interactively, the user gets a security question if he is sure about deleting classes and in consequence data as well.

This PR is related to #2591


brusch:
Thanks!

PR #2617, opened 2018-02-19
Issue in custom layouts import/export

neha04:
## Fixes Issue #
Custom layouts import and export not working because of routing mismatch.

Import route has been changed from '/import-custom-layout' to '/import-custom-layout-definition'.
Export route has been changed from '/export-custom-layout-definitions' to '/export-custom-layout-definition'.

## Steps to reproduce
- Open a custom layout created for any class.
- Click on Export button. 
- This action is broken due to route mismatch.
- Likewise import action also have the same issue.


brusch:
thanks

PR #2655, opened 2018-03-05
Added correct namespace for a call to the Pimcore\Logger class

BlackbitNeueMedien:
I stumbled upon a call to the Logger class within the old namespace and fixed it in this PR.


weisswurstkanone:
thx!

PR #2659, opened 2018-03-06
Removed Object.php for the linters php -l or parallel-lint

brammittendorff-dd:
- Removed Object.php because when using PHP 7.2 linters are complaining about it because the class name Object is reserved.

```
Checked 2390 files in 9.1 seconds
Syntax error found in 1 file

------------------------------------------------------------
Parse error: ./pimcore/models/Object.php:30
    28|     if (version_compare(PHP_VERSION, '7.2.0', '<')) {
    29|         if (!\Pimcore\Tool::classExists('Pimcore\\Model\\Object')) {
  > 30|             class Object extends \Pimcore\Model\DataObject\AbstractObject
    31|             {
    32|             }
Fatal error: Cannot use 'Object' as class name as it is reserved
ERROR: Job failed: exit code 1
```

## Fixes Issue #

When using `php -l` or:
```
parallel-lint --exclude vendor .
```

We get an error:
```
$ php -l pimcore/models/Object.php 
PHP Fatal error:  Cannot use 'Object' as class name as it is reserved in pimcore/models/Object.php on line 30
Errors parsing pimcore/models/Object.php

$ php -v
PHP 7.2.2-3+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Feb  6 2018 16:11:23) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.2-3+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2018, by Zend Technologies
```

- Removed Object.php because it is just an alias ;-)


brusch:
Would break BC - so we can't remove it as long as we support 7.0+

PR #2668, opened 2018-03-08
Classification Store: Removed erroneous check for localized fields

podarcis:
When calling getLocalizedKeyValue() inherited values were only used if
the store's object had a Localized Fields data component. However, the
classification store itself has the option to be localized.

As a workaround it was possible to add an empty Localized Fields to the
object and the values were inherited as expected. This commit removes
the unnecessary check.


weisswurstkanone:
thx!

PR #2671, opened 2018-03-08
extension - user - workspace-getByPath

Apke:

 + (add) User\Wroskpace::getByPath
 + (add) User\Wroskpace::getById

## Additional info  

Hi pimcore-Team,

Maybe my use-case is a little bit special and i solved it with three wrappers, but i thought that the solution could be useful.

Hopefully i did not something wrong. Please feel Free to decline it, if this PR is a missleading mechanism.

Best Regards and Thanks

Apke

### About the Use-Case

Shortly explained:

i have containers which are used on workspaces. A External Service, needs the User which is behind the Service for validation and a refreshing mechanism.

### DIFF
--- a/pimcore/models/User/Workspace/AbstractWorkspace.php
+++ b/pimcore/models/User/Workspace/AbstractWorkspace.php
@@ -18,6 +18,9 @@
 namespace Pimcore\Model\User\Workspace;
 
 use Pimcore\Model;
+use Pimcore\Logger;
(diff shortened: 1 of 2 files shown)


brusch:
First of all, thank you very much for your PR! 
But as you said already, my opinion is also that this is a very specific use case, therefore I close this PR for the moment, if anyone finds this contribution useful as well, please vote for it 👍

PR #2678, opened 2018-03-09
Mail Bug with "localhost" from Console

Gerhard-Kanzler:
If you send a Mail from console the hostname is always "localhost"


brusch:
Thanks!

PR #2680, opened 2018-03-12
Fixes serp preview with subdomains

Dravenshorst:
Currently when using subsites the Serp preview shows the main domain + path, so this looks very strange (for example): `site.com/site.de/contact` because site.com is the main domain.

Checks for subsites and if the second idPath is equal to `subsites.data.rootId` then change the url to `subsite.data.domain` instead of `window.location.protocol + '//' + window.location.hostname`
This will look like: `site.de/contact` instead of `site.com/site.de/contact`
  
## Fixes Issue #
Strange url's in Serp preview when using subsites.

`web/pimcore/static6/js/pimcore/document/pages/settings.js`

## Additional info  
When there are no subsites or the second idPath is not in subsites then it just uses the normal URL: `window.location.protocol + '//' + window.location.hostname`. As it was before.

### DIFF
--- a/web/pimcore/static6/js/pimcore/document/pages/settings.js
+++ b/web/pimcore/static6/js/pimcore/document/pages/settings.js
@@ -203,8 +203,25 @@ pimcore.document.pages.settings = Class.create(pimcore.document.settings_abstrac
                 return true;
             }.bind(this);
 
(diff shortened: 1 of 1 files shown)


brusch:
Thanks!

PR #2682, opened 2018-03-12
Fixes check on array in settings.js

Dravenshorst:
Fixes check on array. 
It checks on `typeof subSites === 'array'` but should be `object` because typeof would not return array but object.

## Fixes Issue #
This fixes the subdomain show correctly instead of maindomain/subsite.

Changes the check on array must be `typeof subSites === 'object'`

## Additional info  
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#examples


brusch:
Thanks!

PR #2691, opened 2018-03-14
Make layout information available in renderer

andreas-gruenwald:
By passing the layout information, the HTML of the text element is available and can be styled individually, for instance as a message output.


weisswurstkanone:
thx!

weisswurstkanone:
since build 224

PR #2692, opened 2018-03-15
Added empty file upload check in js and AssetController

### LINKED ISSUE #2667: Uploading an empty Asset (=0 Bytes size) fails
Uploading an empty file into Assets fails:

- Via "Add asset" > "Upload files" > Select an empty file: Silently fails
- Via drag'n'drop: Fails with error: {"success":false,"message":"Something went wrong, please check upload_max_filesize and post_max_size in your php.ini and write permissions of \/home\/pimcore_demo_basic\/www\/web\/var"}

I just ran into this while I was trying to upload an empty .txt file. I can't imagine this is intended behaviour, however if it is then at least the error message is misleading.

dvesh3:
## Fixes Issue #
Fixes #2667 
Empty file(0 byte) check added to throw error messages


fashxp:
hi, could you please resolve the conflicts for this pull request so that we can merge it. 
thanks...

dvesh3:
hello..It's done :)

weisswurstkanone:
thanks a lot!

Issue #1392, opened 2017-04-05
[DOCS] AWS Setup


brusch:
_From @brusch on March 22, 2017 15:45_

- archtecture recommendation (EC2, EFS, S3, CloudFront, Aurora, ...) 
- S3 Storage example for assets, ... 
- CloudFront Setup

... see v4 docs 

_Copied from original issue: pimcore/pimcore-5-issues#30_

Yivan:
Just a suggestion: maybe setup for Google Cloud Plateform could be added as it is more and more used (not at the same level than AWS for sure, but many people tend to swith to GCP for simplicity, lower price, Kumbertenes containers, etc.).

Always on arhitecture, I know it is new, but making a page about container/docker implementation could be great. I saw there is a Pimcore Docker available: https://hub.docker.com/r/pimcore/docker-pimcore-demo-standalone/~/dockerfile/
I think it has been done as this for simplicity installation for testing purpose, but it break stateless principles and manage multiple process in one container (normally to be scalable you have to have one process per container).
A new page on documentation for Docker, with multiple container by using services and docker-compose.yml (https://docs.docker.com/get-started/part3/#about-services) could be great. As Pimcore 5 can be installed/manager better from command line/composer it will make it fesable.

brusch:
S3 was done by #2755
[CloudFront](https://pimcore.com/docs/4.6.x/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/Amazon_AWS_Setup/Amazon_AWS_Cloudfront_CDN_Setup.html) and [overview page](https://pimcore.com/docs/4.6.x/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/Amazon_AWS_Setup/index.html) is still missing

PR #2696, opened 2018-03-15
Amazon AWS S3 setup

brammittendorff-dd:
We wanted to use Amazon AWS S3 in a simple way. But in pimcore we have our own asset management instead of the Symfony one. @mjrider and I created a workaround so we can use Amazon AWS S3 with the usage of file streams.

- Fixed file streams issue in Imagick. `stream://`
- Added Amazon AWS S3 setup

# Installation instructions without file lookup caching:

Composer packages:
```
composer require league/flysystem-aws-s3-v3 mjrider/flysystem-factory twistor/flysystem-stream-wrapper
```

Create .env vars:
```
AMAZON_S3_URL="s3://accesstoken:secretkey@region/bucketname/"
PIMCORE_ASSET_DIRECTORY="fly://assetdirectory"
PIMCORE_TEMPORARY_DIRECTORY="fly://tempdirectory"
```

Add a `constants.php` to `app/` with contents:
```
<?php
// Initialize flysystem stream wrapper
use League\Flysystem\Adapter\Local;
use League\Flysystem\Filesystem;
use Twistor\FlysystemStreamWrapper;

$filesystem = \MJRider\FlysystemFactory\create(getenv('AMAZON_S3_URL'));

FlysystemStreamWrapper::register('fly', $filesystem);
$manager = new League\Flysystem\MountManager([
    'fly' => $filesystem
]);

define('STREAMS_API', 1);

if (PHP_SAPI != 'cli') {
    $pimcoreFiles = [
        PIMCORE_ASSET_DIRECTORY . $_SERVER['REQUEST_URI'],
        PIMCORE_TEMPORARY_DIRECTORY . $_SERVER['REQUEST_URI']
    ];

    foreach ($pimcoreFiles as $file) {
        if (!empty($file) && file_exists($file) && ($fileSize = filesize($file))) {
            header("Content-Length: " . $fileSize);

            $contentType = $manager->getMimeType($file);
            header('Content-Type: ' . $contentType);
            if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                $stream = $manager->readStream($file);
                if (!$stream) {
                    stream_copy_to_stream($stream, fopen('php://output', 'wb'));
                }
            }
            exit;
        }
    }
}
```

# Installation instructions with file lookup caching:

Composer packages:
```
composer require league/flysystem-aws-s3-v3 mjrider/flysystem-factory twistor/flysystem-stream-wrapper predis/predis
```

Create .env vars:
```
REDIS_URL="predis-tcp://127.0.0.1/?expire=3600"
AMAZON_S3_URL="s3://accesstoken:secretkey@region/bucketname/"
PIMCORE_ASSET_DIRECTORY="fly://assetdirectory"
PIMCORE_TEMPORARY_DIRECTORY="fly://tempdirectory"
```

Add a `constants.php` to `app/` with contents:
```
<?php
// Initialize flysystem stream wrapper
use League\Flysystem\Adapter\Local;
use League\Flysystem\Filesystem;
use Twistor\FlysystemStreamWrapper;

$cache = getenv('REDIS_URL');
$filesystem = \MJRider\FlysystemFactory\create(getenv('AMAZON_S3_URL'));
if ($cache !== false) {
    $filesystem = \MJRider\FlysystemFactory\cache($cache, $filesystem);
}

FlysystemStreamWrapper::register('fly', $filesystem);
$manager = new League\Flysystem\MountManager([
    'fly' => $filesystem
]);

define('STREAMS_API', 1);

if (PHP_SAPI != 'cli') {
    $pimcoreFiles = [
        PIMCORE_ASSET_DIRECTORY . $_SERVER['REQUEST_URI'],
        PIMCORE_TEMPORARY_DIRECTORY . $_SERVER['REQUEST_URI']
    ];

    foreach ($pimcoreFiles as $file) {
        if (!empty($file) && file_exists($file) && ($fileSize = filesize($file))) {
            header("Content-Length: " . $fileSize);

            $contentType = $manager->getMimeType($file);
            header('Content-Type: ' . $contentType);
            if ($_SERVER['REQUEST_METHOD'] === 'GET') {
                $stream = $manager->readStream($file);
                if (!$stream) {
                    stream_copy_to_stream($stream, fopen('php://output', 'wb'));
                }
            }
            exit;
        }
    }
}
```

# PR

## Fixes Issue #

- Fixed file stream handling
- Added Amazon AWS S3 usage in PR (needs to be documented in the future)

- Fixed file streams

## Additional info  

When using file streams with our test.php you can see the file streams break with the current imagick code:
```
<?php

require 'vendor/autoload.php';

// Initialize flysystem stream wrapper
use League\Flysystem\Adapter\Local;
use League\Flysystem\Filesystem;
use Twistor\FlysystemStreamWrapper;

$filesystem = new Filesystem(new Local('/tmp/pimcoreassets'));

FlysystemStreamWrapper::register('fly', $filesystem);
$manager = new League\Flysystem\MountManager([
    'fly' => $filesystem
]);

$i = new \Imagick();
try {
    $i->readImage('fly://assetdirectory/logo.jpg');
    file_put_contents('fly://assetdirectory/logo-2.jpg', $i->getImageblob());
    $i->writeImage('fly://assetdirectory/logo-3.jpg'); // WILL RETURN AN ERROR
} catch (Exception $e) {
    var_dump($e);
}
```
The Imagick `writeImage` will fail because the file does not exist that's why we need to use:
```
file_put_contents('fly://assetdirectory/logo-2.jpg', $i->getImageblob());
```

### DIFF
--- a/pimcore/lib/Pimcore/Image/Adapter/Imagick.php
+++ b/pimcore/lib/Pimcore/Image/Adapter/Imagick.php
@@ -103,7 +103,7 @@ class Imagick extends Adapter
             }
 
             $imagePathLoad = $imagePath;
-            if (!defined('HHVM_VERSION')) {
+            if (!defined('HHVM_VERSION') && !defined('STREAMS_API')) {
                 $imagePathLoad .= '[0]'; // not supported by HHVM implementation - selects the first layer/page in layered/pages file formats
             }
 
@@ -235,8 +235,8 @@ class Imagick extends Adapter
             $path = PIMCORE_SYSTEM_TEMP_DIRECTORY . '/imagick-tmp-' . uniqid() . '.' . File::getFileExtension($path);
         }
 
-        if (defined('HHVM_VERSION')) {
-            $success = $i->writeImage($path);
+        if (defined('HHVM_VERSION') || defined('STREAMS_API')) {
+            $success = file_put_contents($path, $i->getImageblob());
         } else {
             $success = $i->writeImage($format . ':' . $path);
         }


fashxp:
very nice. please give us some time to review that! thanks.

brusch:
I think using `stream_is_local()` instead of another constant would do the same right? 

Regarding your approach: wouldn't it be better to deliver the assets/tmp-files directly from AWS, either by using CloudFront or directly from S3 (by using an appropriate policy), rather than starting the PHP engine again? 

You could achieve this by using the following events: 
https://github.com/pimcore/pimcore/blob/fa8298c6caabc3496d17a3b25107d4ad1e10bc9d/pimcore/lib/Pimcore/Event/FrontendEvents.php#L32

We had already a how-to based on that concept for Pimcore 4: 
https://pimcore.com/docs/4.6.x/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/Amazon_AWS_Setup/Amazon_AWS_S3_Setup.html

I'm pretty sure this will have a significant impact on performance.

brammittendorff-dd:
The solution of 4.6.x doesn't work on 5.0.x because there is no static getEventManager. So the solution of `pimcore/pimcore/lib/Pimcore/Event/FrontendEvents.php` is of no use for us. It would be better to write directly AWS but then there are more problems popping up when directly writing to AWS because the ASSET_IMAGE_THUMBNAIL is not the only thing we want to write to AWS.

You are right it would have a significant impact on performance. In this PR I showed a relatively easy example to use AWS. We use a load-balancer / CloudFront solution to fix the performance issue. And later we can migrate to a better solution by rewriting the path names but for now, this is the quickest solution we could think of.

We want AWS to handle all files for us. This PR is a fix for **the current Imagick problem** it fixes the writing to a nonexisting file.

```
<?php

require 'vendor/autoload.php';

// Initialize flysystem stream wrapper
use League\Flysystem\Adapter\Local;
use League\Flysystem\Filesystem;
use Twistor\FlysystemStreamWrapper;

$filesystem = new Filesystem(new Local('/tmp/pimcoreassets'));

FlysystemStreamWrapper::register('fly', $filesystem);
$manager = new League\Flysystem\MountManager([
    'fly' => $filesystem
]);

$i = new \Imagick();
try {
    $i->readImage('fly://assetdirectory/logo.jpg');
    file_put_contents('fly://assetdirectory/logo-2.jpg', $i->getImageblob());
    // $i->writeImage('fly://assetdirectory/logo-3.jpg'); // UNCOMMENT WILL RESULT IN A ERROR
} catch (Exception $e) {
    var_dump($e);
}
```

I can imagine you do not like the AWS solution here. But for us, it works and is a quick fix to use AWS. The best fix would be to rewrite the file handling to the Symfony file handling but for now, this is to much work and we will look into this at a later time of period. When we use the Symfony way of file handling we can use the Symfony AWS (https://gist.github.com/maennchen/b075407e1f4f722278f4) way or the Symfony Flysystem(https://github.com/1up-lab/OneupFlysystemBundle) Bundle for our files.

brusch:
@brammittendorff-dd you don't need a static event-manager, you just register an event-listener in your `services.yml` , actually it's even less work to get this up an running ;) 

eg.: 
```yml
services:
    AppBundle\EventListener\FrontendPathListener:
        tags:
            - { name: kernel.event_listener, event: pimcore.frontend.path.asset.image.thumbnail, method: rewritePath }
            - { name: kernel.event_listener, event: pimcore.frontend.path.asset.document.image-thumbnail, method: rewritePath }
            - { name: kernel.event_listener, event: pimcore.frontend.path.asset.video.image-thumbnail, method: rewritePath }
            - { name: kernel.event_listener, event: pimcore.frontend.path.asset.video.thumbnail, method: rewritePath }
```

brusch:
@brammittendorff-dd can you confirm that `stream_is_local()` has the same effect? If so, can you please update your PR? Thanks!

brammittendorff-dd:
@brusch I see your concern, let me check this right now.

brammittendorff-dd:
Unfortunately we get a warning because our defined wrapper for the streams is not recognized by `stream_is_local` (https://3v4l.org/a56Fp):
```
Warning: stream_is_local(): Unable to find the wrapper "fly" - did you forget to enable it when you configured PHP? in /in/a56Fp on line 3
bool(true)
```

brammittendorff-dd:
Unfortunately, it sees our own defined stream wrapper for AWS as a local stream but it isn't. So `stream_is_local()` is not working.

brammittendorff-dd:
Another problem is with your event listeners is we need to have public amazon bucket and we do not want that for our customers (private assets). Second, you can create temporary URLs in Amazon S3 but this requires a lot more than just the event listeners we have.

brusch:
You can expose only certain directories in your bucket using cloudfront or custom policies ... so that shouldn't be an issue - or even better separate the private stuff out into a dedicated private bucket. 
When using the AWS SDK instead of Flysystem the stream is correctly detected as remote stream, maybe this is something that needs to be fixed in Flysystem ...?

frenck:
Really?

How about you guys actually add support for CDN's (instead of silently removing it from the feature set between Pimcore 4 and 5).

How are we supposed to take this into production on a serious level? (multi-server, multi-data center, multi-site)

brammittendorff-dd:
The changes to this PR is just a fix for pimcore? Why are you ranting about Flysystem this is no issue here We can replace this by many others. The problem *fixed* in this PR is a real problem with pimcore why are you declining this?

brusch:
@frenck it wasn't silently removed ... just reference your static resources directly to the CDN endpoint or override on the `head` helpers or one of the existing Symfony packages. When using a CDN with Pimcore resources, just use the events described above. 

@brammittendorff-dd when using a remote file system `stream_is_local()` should return false, this is not the case when using Flysystem, but it is the case when using the AWS SDK, so that's **NOT** a problem of Pimcore and should **NOT** be fixed with another (unprefixed) constant.

frenck:
@brusch First of all, the PR is not about the CDN, but about another issue.

Secondly, the CDN part, sounds easy, please show me the documentation?

brammittendorff-dd:
@brusch This has nothing to do with `stream_is_local()`, this is not even used in the code we changed. The real problem is Imagick cannot write to a file that doesn't exist see next url:
https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Image/Adapter/Imagick.php#L239

The fix should be as this PR stated:
```
$success = file_put_contents($path, $i->getImageblob());
```

Did you even test the script?
```
<?php

require 'vendor/autoload.php';

// Initialize flysystem stream wrapper
use League\Flysystem\Adapter\Local;
use League\Flysystem\Filesystem;
use Twistor\FlysystemStreamWrapper;

$filesystem = new Filesystem(new Local('/tmp/pimcoreassets'));

FlysystemStreamWrapper::register('fly', $filesystem);
$manager = new League\Flysystem\MountManager([
    'fly' => $filesystem
]);

$i = new \Imagick();
try {
    $i->readImage('fly://assetdirectory/logo.jpg');
    file_put_contents('fly://assetdirectory/logo-2.jpg', $i->getImageblob());
    // $i->writeImage('fly://assetdirectory/logo-3.jpg'); // UNCOMMENT WILL RESULT IN A ERROR
} catch (Exception $e) {
    var_dump($e);
}
```

frenck:
Ow, and please tell me how to overcome this line without this patch and using a CDN solution:


https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Image/Adapter/Imagick.php#L239

brusch:
@frenck so why are you then discussing about CDN here? ;) 
See https://github.com/pimcore/pimcore/pull/2696#issuecomment-375301212 
a detailed documentation is currently not available (it's on the roadmap, see #1392), but should be pretty straight forward.

brusch:
It **HAS** something to do with `stream_is_local()` because ... I don't want to introduce a new constant (`STREAMS_API`) for something that could be solved by a native PHP function. 

In detail ... if `stream_is_local('fly://assetdirectory/logo-2.jpg') === false` you'd have the desired result. 
But it isn't, and this is not Pimcore's fault. So if the behavior would be right, as it is with https://github.com/aws/aws-sdk-php (details here: https://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-stream-wrapper.html) then `stream_is_local()` would do the trick in a more elegant way. 

We have implemented this already throughout Pimcore, see:

brammittendorff-dd:
@brusch we can remove the `STREAMS_API` constant right now and the code is still fixed this is also not the issue. I asked you if you rather saw the `STREAMS_API` removed? The problem is still this line of code:
https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Image/Adapter/Imagick.php#L239

Not the constants above!

brusch:
@brammittendorff-dd but what's the point of changing the code without the additional constant?`: 

```php 
        if (defined('HHVM_VERSION')) {
            $success = File::put($path, $i->getImageBlob());
        } else {
            $success = $i->writeImage($format . ':' . $path);
        }
```

This wouldn't fix your problem at all ;) 
Anyway, there's still an issue with remote stream handling, so I've fixed that.

brammittendorff-dd:
Well, thanks @brusch for creating the "Imagick remote stream support" PR and using my code.

PR #2703, opened 2018-03-16
[ECOMMERCE][FILTERSERVICE] Misc FilterService fixes

podarcis:
* Fixed range-condition of Elasticsearch/NumberRange[Selection]
* Allow value zero in from/to of Elasticsearch/NumberRange[Selection]


fashxp:
thx very much

PR #2707, opened 2018-03-20
Consent Datatype

fashxp:
Adds datatype to store consent of users to certain things, like consent to marketing direct mailing etc. 
Consent itself is stored as checkbox in data object, additional meta information (like timestamp and all other things) are stored as a note on the corresponding object. 

Usage: 
- directly in data object: change value of checkbox, when saving object note is added automatically with default values
- in custom controllers use service `Pimcore\DataObject\Consent\Service` and methods `giveConsent` and `revokeConsent` 

In addition to that some refactoring took place in terms of 'datatype supports data inheritance'. Now there is an method `supportsInheritance` in `Pimcore\Model\DataObject\ClassDefinition\Data` that defines if the datatype supports inheritance or not. Default is to `true`, CalculatedValue, FieldCollection and Consent don't support data inheritance.


ckemptner:
Docs: https://pimcore.com/docs/5.x/Development_Documentation/Objects/Object_Classes/Data_Types/Consent.html

PR #2716, opened 2018-03-22
Flush the container of a ObjectBrick on delete of a brick

ThomasKeil:
The ObjectBrick container[ caches the result](https://github.com/pimcore/pimcore/blob/master/pimcore/models/DataObject/Objectbrick.php#L82) of getItems().
This resets the result to the initial value when a brick gets deleted.


ThomasKeil:
I actually did this fix for a local 4.x branch - would you care for a PR for the old branch, too?

brusch:
@ThomasKeil thx, yes, of course we'd accept a PR for the 4.x branch :)

PR #2725, opened 2018-03-26
Import - omit mandatory check if target isn't going to be published

kkarkus:


brusch:
Makes sense - thanks!

Issue #2583, opened 2018-02-09
GA overview - missleading source tooltip


jansarmir:
https://github.com/pimcore/pimcore/blob/02b70e96c27e61c4df66d8d54af82d196997e7d8/web/pimcore/static6/js/pimcore/report/analytics/elementoverview.js#L180

tooltip shows pageviews with % suffix, which is confusing

i suggest something like this: google 587 (13 %)

brusch:
Can you create an PR for that? :) 
Thanks!

PR #2726, opened 2018-03-26
GA overview - missleading source tooltip (#2583)

jansarmir:
PR for https://github.com/pimcore/pimcore/issues/2583


brusch:
Thanks!

PR #2746, opened 2018-04-05
[docs] nginx thumbnail overload protection

bbak:
Utilize nginx rate limiting to protect the server from overload by too many thumbnail generation processes in parallel.

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/02_Nginx_Configuration.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/02_Nginx_Configuration.md
@@ -146,3 +146,44 @@ server {
     }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


brusch:
Merci beaucoup!

bbak:
May I propose, you add something similar to the Apache config example also? We've experienced this to be a major issue: 
If you change a image pipeline of a website in "production" with many assets and reasonable traffic, it's likely that several dozen Imagick Processes try to start converting images. This is likely to overload available memory and so Imagick will start creating files in TEMP. We've experienced 30+ of such files, ranging from roughly 150MB to 500MB, being written in parallel, overloading the I/O-Capacity of the Storage Subsystem, effectively blocking the whole Server due to IO-Wait.

As an alternative: Some kind of locking-mechanism that prevents, say, more than 5 parallel Thumbnail-Generation Processes would solve the issue on Application-Level. Of course that number should be configurable.

PR #2748, opened 2018-04-05
Fixed small Javascript Issues

daves-weblab:
### Batch actions in object grid
When objects in the grid view are edited fast enought ExtJs would batch those changes, which the backend is not compliant with, resulting in an error. The batching process has been disabled for grids created with `pimcore.object.helpers.grid`.

### Multihref Advanced & Objects with Metadata
Both these tags had the issue that ExtJs's garbage collector would collect the cell editors if working long enought with the tag. Usually when adding / removing / sorting elements the garbage collector would destroy the cell editor leading to a javascript exception and disabling the layout rerender, which could only be fixed by refreshing the backend. By destroying and recreating the cell editors each time a cell is being edited, the issue could be fixed.

### Combobox delayed preselect
By recreating the combobox for the cell editor each time, the preselect of the current value is delayed. This behaviour was previously present as well, but only for the first time the combobox was rendered. Was fixed by switching to the local queryMode of the combobox store (only for typeahead stores).

### DIFF
--- a/web/pimcore/static6/js/pimcore/element/abstract.js
+++ b/web/pimcore/static6/js/pimcore/element/abstract.js
@@ -118,12 +118,24 @@ pimcore.element.abstract = Class.create({
     resetChanges: function () {
         this.changeDetectorInitData = {};
 
-        this.tab.setTitle(this.tab.initialConfig.title);
-        this.startChangeDetector();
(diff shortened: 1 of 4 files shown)


weisswurstkanone:
build 231

PR #2761, opened 2018-04-10
[docs] fix nginx thumbnail rate limiting

bbak:
`try_files` directive should use `$uri` instead of `$request_uri` since the latter is not urldecoded.


brusch:
Thanks!

PR #2765, opened 2018-04-10
Bugfix - Thumnail - return correct file system path if no config is passed

ckogler-elements:
Return correct filesystem path if no config was provided.
The getRealFullPath() method doesn't return the Document_Root


brusch:
Thanks!

PR #2767, opened 2018-04-11
Fix newsletter test send fails when address source is empty

### LINKED ISSUE #1344: Newsletter Test-Send fails when address source is empty
We use different custom reports as address souce for newsletter sendouts.
One of these reports did not select any recipient.
When now using this report to start a test send, it fails without showing any feedback to the user.

The ajax call returns an error:
> Fatal error: Uncaught TypeError: Argument 2 passed to Pimcore\Document\Newsletter\SendingParamContainer::__construct() must be of the type array, boolean given, called in /public/pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapter/ReportAdapter.php on line 114 and defined in /public/pimcore/lib/Pimcore/Document/Newsletter/SendingParamContainer.php on line 35
> TypeError: Argument 2 passed to Pimcore\Document\Newsletter\SendingParamContainer::__construct() must be of the type array, boolean given, called in /public/pimcore/lib/Pimcore/Document/Newsletter/AddressSourceAdapter/ReportAdapter.php on line 114 in /public/pimcore/lib/Pimcore/Document/Newsletter/SendingParamContainer.php on line 35

dvesh3:
## Fixes Issue #
Fixes #1344
1. Fixed email field load issue when selecting report in Newsletter sending panel.
2. Added a check for empty email address for test send newsletter

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/NewsletterController.php
+++ b/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/NewsletterController.php
@@ -338,6 +338,11 @@ class NewsletterController extends DocumentControllerBase
         $document = Document\Newsletter::getById($request->get('id'));
         $addressSourceAdapterName = $request->get('addressAdapterName');
         $adapterParams = json_decode($request->get('adapterParams'), true);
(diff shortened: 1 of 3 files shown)


fashxp:
thx

PR #2775, opened 2018-04-13
Added failed add/save/delete events

### LINKED ISSUE #236: Events for object/document/asset save/delete operations failure
Needed for a situation where a plugin should be notified while save/delete operation fail for a reason, to handle plugin related data accordingly.

Simple use case: you log that someone tried to delete object by using preDelete event collecting object data before it is really deleted but than the deletion fails by throwing exception, so you have to clean up but you can not currently because there is no way to get notified about failure in the deletion process of that concrete instance.

dvesh3:
Added events for failure handling:
- DataObjectEvents::POST_ADD_FAILURE
- DataObjectEvents::POST_UPDATE_FAILURE
- DataObjectEvents::POST_DELETE_FAILURE
- AssetEvents::POST_ADD_FAILURE
- AssetEvents::POST_UPDATE_FAILURE
- AssetEvents::POST_DELETE_FAILURE
- DocumentEvents::POST_ADD_FAILURE
- DocumentEvents::POST_UPDATE_FAILURE
- DocumentEvents::POST_DELETE_FAILURE

### DIFF
--- a/pimcore/models/Asset.php
+++ b/pimcore/models/Asset.php
@@ -599,7 +599,7 @@ class Asset extends Element\AbstractElement
             $parent = Asset::getById($this->getParentId());
             if ($parent) {
                 // use the parent's path from the database here (getCurrentFullPath), to ensure the path really exists and does not rely on the path
-                // that is currently in the parent object (in memory), because this might have changed but wasn't not saved
+                // that is currently in the parent asset (in memory), because this might have changed but wasn't not saved
                 $this->setPath(str_replace('//', '/', $parent->getCurrentFullPath() . '/'));
             } else {
                 // parent document doesn't exist anymore, set the parent to to root
@@ -757,7 +757,7 @@ class Asset extends Element\AbstractElement
 
         $this->getDao()->update();
 
-        //set object to registry
+        //set asset to registry
         \Pimcore\Cache\Runtime::set('asset_' . $this->getId(), $this);
         if (get_class($this) == 'Asset' || $typeChanged) {
             // get concrete type of asset
@@ -802,7 +802,7 @@ class Asset extends Element\AbstractElement
         $version = null;
 
         // only create a new version if there is at least 1 allowed
-        // or if saveVersion() was called directly (it's a newer version of the object)
+        // or if saveVersion() was called directly (it's a newer version of the asset)
         if (Config::getSystemConfig()->assets->versions->steps
             || Config::getSystemConfig()->assets->versions->days
             || $setModificationDate) {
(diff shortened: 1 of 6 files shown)


pimcore/models/Asset.php
```diff
@@ -987,50 +992,55 @@ public function delete()
...
+            $this->getDao()->delete();
 
-        // empty object cache
-        $this->clearDependentCache();
+            // empty object cache
+            $this->clearDependentCache();
 
-        //set object to registry
-        \Pimcore\Cache\Runtime::set('asset_' . $this->getId(), null);
+            //set object to registry
```
weisswurstkanone: Misleading doc seems to be there since ages? (nulls it out instead of setting it. Besides that, in this context we are talking about assets ... also see the similar stuff below). Can we correct this together with this PR? :-)

dvesh3: Done :-)

pimcore/models/Asset.php
```diff
@@ -987,50 +992,55 @@ public function delete()
...
-        $this->clearDependentCache();
+            // empty asset cache
+            $this->clearDependentCache();
 
-        //set object to registry
-        \Pimcore\Cache\Runtime::set('asset_' . $this->getId(), null);
+            // clear asset from registry
+            \Pimcore\Cache\Runtime::set('asset_' . $this->getId(), null);
+        } catch (\Exception $e) {
+            \Pimcore::getEventDispatcher()->dispatch(AssetEvents::POST_DELETE_FAILURE, new AssetEvent($this));
```
weisswurstkanone: is there a throw $e missing at the end of the exception handler  ?

fashxp: we definitely still should throw the exceptions.

dvesh3: Done :-) thanks for the catch 👍

pimcore/models/DataObject/AbstractObject.php
```diff
@@ -533,35 +533,39 @@ public function setLocked($o_locked)
...
-        $this->clearDependentCache();
+            // empty object cache
+            $this->clearDependentCache();
 
-        //set object to registry
-        \Pimcore\Cache\Runtime::set('object_' . $this->getId(), null);
+            //clear object from registry
+            \Pimcore\Cache\Runtime::set('object_' . $this->getId(), null);
+        } catch (\Exception $e) {
+            \Pimcore::getEventDispatcher()->dispatch(DataObjectEvents::POST_DELETE_FAILURE, new DataObjectEvent($this));
```
fashxp: some goes for here ... we should throw that exception

dvesh3: Done ;-)

fashxp:
thx

PR #2776, opened 2018-04-13
Set Branch Alias for dev-master

blankse:
I want set a composer require of a bundle for pimcore because of testing.
e. g.
```
"require": {
  "pimcore/pimcore": "^5.2"
}
```

Than i have the problem that the bundle isn't be able to install in a normal pimcore installation.
Because dev-master has no version. With a branch alias it will be work.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -173,6 +173,9 @@
       "merge-extra": false,
       "merge-extra-deep": false,
       "merge-scripts": false
+    },
+    "branch-alias": {
+      "dev-master": "5.2.x-dev"
     }
   }
 }


dpfaffenbauer:
I guess you shouldn't use `pimcore/pimcore`, you should use `pimcore/core-version` right?

blankse:
In `pimcore/core-version` is no code. I need the pimcore code for tests. Some classes extends pimcore classes.

dpfaffenbauer:
thats right, but the Pimcore Core Team doesn't see Pimcore as a dependency, they see it as a project.

blankse:
I don't know how to test extension then. After the first look into the dependents of pimcore/core-version (https://packagist.org/packages/pimcore/core-version/dependents)  I see no tests. Am I the first one who wants to test?

I can set the banch alias in `composer.local.json` for my own of course. Thought that would be better for everyone.

dpfaffenbauer:
this is how i do it: https://github.com/coreshop/CoreShop/blob/master/etc/travis/setup/before_install

brusch:
We'll consider branch aliasing as soon as the Pimcore core can be used as Composer dependency (will be in v6).

PR #2781, opened 2018-04-17
Output cache separation

### LINKED ISSUE #261: Make the permissions over the cache more managable/advanced
Separate the three kind of cache actions by separate permissions
- Output cache
- Cache (everything)
- Temporary files

It may be fine for some persons to have the ability to clear the output cache but access to clearing temporary files or all the caches may be not good idea.

dvesh3:
- Full Page Cache clear option moved out of clear cache menu
- New permission added in user/role for Clear full Page cache
## Additional info  
DB update:
INSERT INTO `users_permission_definitions` (`key`) VALUES ('clear_fullpage_cache');


fashxp:
we also would need to add it in the default setup routine here: 
https://github.com/pimcore/pimcore/blob/master/pimcore/models/Tool/Setup/Dao.php#L126

PR #2785, opened 2018-04-18
[DOC] Update documentation example, use the defaultValue

jorisros:
Updated the code example to explain how to get the default value back from the fielddefinition.


fashxp:
thy very much

PR #2790, opened 2018-04-19
Fixed PHPDoc for type

ThomasKeil:
get() on a Block element in a DataObject doesn't return Data\Block but Data\BlockElement, and to be more specific an array of an array of BlockElements.

It's possible to specify that as FooBar[][], which as notation will be accepted by e.g. PhpStorm:

auswahl_182

For testing this: unfortunately there's apparantly a bug in PhpStorm since the return hint won't suffice:
auswahl_183
auswahl_184

I'll file a bug report with PhpStorm as soon as possible.


fashxp:
thx

ThomasKeil:
Here's the Issue btw: https://youtrack.jetbrains.com/issue/WI-41748

PR #2791, opened 2018-04-20
Bugfix: Cart price modificated twice

ThomasKeil:
The cart price modificators are collected twice by PimcoreEcommerceFrameworkExtension::registerCartManagerConfiguration().
This fixes it.

Currently the read configuration results in an array like this:
auswahl_187
The modificators are read twice, with their names as key and a numeric value.
This results in them being applied twice to the cart price.

This fix to the configuration definition results in omitting the numeric index:
auswahl_186

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/DependencyInjection/Configuration.php
+++ b/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/DependencyInjection/Configuration.php
@@ -301,6 +301,7 @@ class Configuration implements ConfigurationInterface
                                     ->append($this->buildOptionsNode('factory_options'))
                                     ->arrayNode('modificators')
                                         ->info('List price modificators for cart, e.g. for shipping-cost, special discounts, etc. Key is name of modificator.')
+                                        ->useAttributeAsKey('name')
                                         ->prototype('array')
                                             ->children()
                                                 ->scalarNode('class')->isRequired()->end()


fashxp:
thx

ThomasKeil:
I'm kinda hijacking this issue here, but it's very similar.
Something like this is also happening for the checkout steps, or rather the checkoutStepOrder:

auswahl_181

You can't finish a cart like this since there is always "a step more"...
Can you confirm - do you want a PR for this, too? I fixed this for me with an override of the CheckoutManager

fashxp:
that is really strange ... the demo works, doesn't it?

fashxp:
hmm, just double checked in the demo and I didn't notice having a checkout step twice. 
any way to reproduce this? or what would be your patch, does it affect/harm any other behavior?

ThomasKeil:
The demo does indeed work - I'm not quite sure yet if it only occurs if you e.g. just have a single step. I guess this needs further investigation.

I can reproduce this in my system - I'm not quite sure how  to set up the demo to reproduce this since I'm not sure if this unwanted behaviour has prerequisites that are unknown to me.

My fix currently would be to also add another "->useAttributeAsKey('name')" in Configuration.php.
This doesn't harm the outcome of parsing the config as far as I know and produces the desired results.

fashxp:
I tried it also with one step within the demo. 
Let's make a pull request for that and I'll try to investigate a bit further.

PR #2795, opened 2018-04-20
Short hand for getters/setters on Objects 

### LINKED ISSUE #1312: Add a shorthand for "getter" and "setter" on Objects
# Feature Request 
Currently we often need to create getter/setter to get or set values by a field name.

                    $fieldNames = ['field1','field2'];

                    foreach($fieldNames as $field){
                        #currently
                        $value = $o->{"get".ucfirst($field)}();
                        #could be
                        $value = $o->get($field);
                    }

Add a shorthand to get and set Values on objects by there field Name. 

Add the following Code to the AbstractObject.php

    /**
     * @param string $fieldName
     * @return mixed
     */
    public function get($fieldName){
        return $this->{"get".ucfirst($fieldName)}();
    }

    /**
     * @param string $fieldName
     * @param $value
     * @return mixed
     */
    public function set($fieldName,$value){
        return $this->{"set".ucfirst($fieldName)}($value);
    }

harpreet819:
Added getters & setters short hand for Objects. 

Shorthand getter
`$value = $object->get($field);`

Shorthand setter
$language is optional. 
`$object->set($field, $value, $language);`

### DIFF
--- a/pimcore/models/DataObject/AbstractObject.php
+++ b/pimcore/models/DataObject/AbstractObject.php
@@ -1287,4 +1287,26 @@ class AbstractObject extends Model\Element\AbstractElement
     {
         self::$doNotRestoreKeyAndPath = $doNotRestoreKeyAndPath;
     }
+
+    /**
+    * @param string $fieldName
+    * @param null $language
+    * @return mixed
+    */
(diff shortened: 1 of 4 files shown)


fashxp:
thx

PR #2797, opened 2018-04-20
Add object open listener on ID & Path columns in object grid

### LINKED ISSUE #1359: Feature Request: Add double click to open to object grid list ID column
# Feature Request 
This is a feature request for better usability. In Asset Grid Views, we can double click the ID of an asset which does immediately open an asset.

In Object grid lists we do not have this feature, we have to use the right mouse click and select open instead. If you go through a filtered list of a hundred objects and need to get into each of them, the two different clicks needed will be annoying and time consuming.
It'd be great if we could add the double click on ID column (since they are not editable anyway) from the assets to the object grid list.

dvesh3:
Added double click listener on "ID" & "Path" column to open object on grid.


fashxp:
thx very much

PR #2802, opened 2018-04-24
Prevent duplicates in CheckoutStepOrder

ThomasKeil:
Fixes [adding a checkout-step](https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/CheckoutManager/CheckoutManager.php#L137) to $checkoutStepOrder[] more than once.


fashxp:
thx

PR #2807, opened 2018-04-27
Update document path on tree move

### LINKED ISSUE #1170: Reload document if open when it is moved within the document tree
otherwise click on reload button inside edit tab will cause a 'No route, document, custom route or redirect is matching the request: /...' exception. 
see also #1161

dvesh3:
Update document path if open when it is moved within the document tree


fashxp:
thx

PR #2820, opened 2018-05-03
Correctly determine pagination direction when using "show in tree" function

BlackbitNeueMedien:
### Steps to reproduce (without this PR's changes)
1. Create data object with key "aa" in folder x. Copy and paste it until tree items in folder x paginate and the last object on page 1 has a key "aa_*". 
1. Now create object with key "a_". Object gets opened
1. Go to page 1 in tree of folder x
1. click "show in tree" in object "a_"
1. Object does not get found in tree


brusch:
Great - thanks!

PR #2824, opened 2018-05-03
added session id as cookie for wkhtmltoimage so that it can access ol…

frithjof:
…der versions of documents in admin compare

  
## Fixes Issue #
document compare versions with wkhtmltoimage only gets newest version so compare is not working

Added session cookie for wkhtmlimage command so that it can access old versions with your admin session cookie


brusch:
Fixed by #2825

In particular, see: https://github.com/pimcore/pimcore/commit/a9ab8386685dc381fcf4af3828ab6a0c2a5cb92b#diff-a0c9b28b176d2f2f97c6d45a2534e3c5R72

I've refactored a bit more, because it doesn't make sense to parse the session/cookie information out of the query string. The query string method was only used in Pimcore 4 and wasn't migrated properly.

PR #2826, opened 2018-05-07
Fixed asset gets processed as Document

dvesh3:
Added Asset Id empty check for unpublished document exception throw so that asset thumbnail will be generated

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/CoreBundle/EventListener/Frontend/ElementListener.php
+++ b/pimcore/lib/Pimcore/Bundle/CoreBundle/EventListener/Frontend/ElementListener.php
@@ -118,7 +118,7 @@ class ElementListener implements EventSubscriberInterface, LoggerAwareInterface
             $user = $this->userLoader->getUser();
         }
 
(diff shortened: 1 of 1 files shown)


brusch:
I'm afraid that this is not the ultimate solution, because the general question is why the document routing matches and the document is on the request (`contentDocument` attribute), see https://github.com/pimcore/pimcore/blob/2beee36a530aaf7b5ab4a4d7d58a6cc955fe46d2/pimcore/lib/Pimcore/Bundle/CoreBundle/EventListener/Frontend/ElementListener.php#L107

So we first should find out why `/series/batman/batman.jpg` resolves the document `/series/batman`.

dvesh3:
Agree!

So the document routing matches and the document is on the request because of getNearestDocumentByPath service which breaks the path and finds the closet document (in our case it will find closet document to path /series/batman, which is unpublished). So it sets the document in document resolver.
https://github.com/pimcore/pimcore/blob/e1b832c6949a534f1db67ce1c4d91d34dc77d36b/pimcore/lib/Pimcore/Bundle/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php#L157
What do you think is the best way to deal with this issue?

brusch:
I'll close that one for now, as we'll have to go for a different approach, so a new PR makes more sense :)

PR #2829, opened 2018-05-08
Fix: Imagick couldn't load images with colon in path

blankse:
If the path (folder or filename) has a colon (:), Imagick couldn't load the image.

Reason is: Imagick use the colon to specify the type of the image.
https://stackoverflow.com/questions/45878100/handle-colon-in-filename-with-imagemagick


brusch:
Thanks! :)

PR #2831, opened 2018-05-09
Fix for CSV export.

kkirsz:
## Fixes Issue
An error exists when you're exporting into csv file any field of type QuantityValue which is located under classification store. This PR resolves it.

## How to reproduce the issue
Add classification store field with QuantityValue key to class. Try to export objects into csv file. You should get "Internal Server Error".

This PR fixes QuantityValue::getForCsvExport() method.


brusch:
Thanks!

PR #2832, opened 2018-05-10
set options and id before firing preOpen Event

dpfaffenbauer:
Setting ID, Options before preOpen Event, otherwise the event has no sense for objects. Also did it for print documents.


brusch:
Thanks!

PR #2836, opened 2018-05-11
Store page size in column configuration

### LINKED ISSUE #2774: Store page size in saved data object grid configurations
25 items per page is not that much on big screens. Therefore it would be nice if the stored grid column configs also could take the "items per page" setting into account.

dvesh3:
resolves #2774 

Added option to set "Items Per Page" in column configuration
items-per-page


brusch:
Thank you very much!

PR #2850, opened 2018-05-16
ApplicationLogger set wrong relatedObjectType

blankse:
If I use the ApplicationLogger direct without MonoLog, the ApplicationLogger always set "object" as relatedObjectType.

e. g.:

```
namespace TestNamespace;

use Pimcore\Log\ApplicationLogger;
use Pimcore\Model\Asset;

class ExampleClass
{
    /** @var \Pimcore\Log\ApplicationLogger */
    protected $applicationLogger;

    public function __construct(ApplicationLogger $applicationLogger)
    {
        $this->applicationLogger = $applicationLogger;
    }

    public function test()
    {
        $asset = Asset::getByPath('/Test.jpg')
        $this->applicationLogger->info(
            'Test',
            [
                'relatedObject' => $asset,
            ]
    }
}
```

### DIFF
--- a/pimcore/lib/Pimcore/Log/ApplicationLogger.php
+++ b/pimcore/lib/Pimcore/Log/ApplicationLogger.php
@@ -177,10 +177,16 @@ class ApplicationLogger implements LoggerInterface
         }
 
         if ($relatedObject) {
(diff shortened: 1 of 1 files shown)


brusch:
Made it a bit simpler, see 4de1aa725c1ee09022d8bf58a3f39d91fe67f0be

blankse:
@brusch With your change the `$this->relatedObjectType` Attribute from the `setRelatedObject` function is not handled anymore.

PR #2858, opened 2018-05-18
Tadschikistan and Turkmenistan not apear in the Countrylist 

### LINKED ISSUE #2622: Tadschikistan and Turkmenistan not apear in the Countrylist
Hi

The two Countries Tadschikistan and Turkmenistan are not in the Objects Country Field.
Please add this two Countries.

harpreet819:
resolves #2622 

Tadschikistan and Turkmenistan will be shown in country list of objects country field. It uses 3rd party bundle umpirsky/country-list for listing all the countries.

### DIFF
--- a/pimcore/lib/Pimcore/Localization/Locale.php
+++ b/pimcore/lib/Pimcore/Localization/Locale.php
@@ -111,7 +111,12 @@ class Locale
             $locale = $this->findLocale();
         }
 
-        $regions = include(PIMCORE_COMPOSER_PATH . '/umpirsky/country-list/data/' . $locale . '/country.php');
+        if (file_exists(PIMCORE_COMPOSER_PATH . '/umpirsky/country-list/data/' . $locale . '/country.php')) {
+            $regions = include(PIMCORE_COMPOSER_PATH . '/umpirsky/country-list/data/' . $locale . '/country.php');
+
+        } else {
+            $regions = include(PIMCORE_COMPOSER_PATH . '/umpirsky/country-list/data/en/country.php');
+        }
         return $regions;
     }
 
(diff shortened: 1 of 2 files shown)


dpfaffenbauer:
IMO: thats kind of a risky change. until your PR, Pimcore used the intl-extension to get this informations, the intl-extension gets the information from the installed locales on the system. Which is fine -> cause it uses the same information as other software does.

Your change introduces a new dependency and therefore other results. 

Maybe you should check your system setup before trying to introduce this kind of change.

brusch:
@dpfaffenbauer 
The problem is that many Linux distributions have outdated ICU packages, where certain locales/countries are completely missing. But even in the latest ICU version 61.1 Turkmenistan is missing ... so it's not ICU's goal to provide a complete list of countries (that's the CLDR on which ICU is based on).
In this case we're only using the country list nothing else, so I'd say this is pretty safe and better then not having a complete list of countries, but if you have a better idea, let me know :)

dpfaffenbauer:
@brusch I am a bit surprised that the ICU packages are that much outdated. But if thats the case, I am happy with that.

pimcore/lib/Pimcore/Localization/Locale.php
```diff
@@ -111,16 +111,7 @@ public function getDisplayRegions($locale = null)
...
-        $locales = $this->getLocaleList();
-        foreach ($locales as $code) {
-            $regionCode = \Locale::getRegion($code);
-            if ($regionCode) {
-                $regionTranslation = \Locale::getDisplayRegion($code, $locale);
-                $regions[$regionCode] = $regionTranslation;
-            }
-        }
-
+        $regions = include(PIMCORE_COMPOSER_PATH . '/umpirsky/country-list/data/' . $locale . '/country.php');
```
brusch: Unfortunately one thing is missing here, ... we need a check if the `country.php` for the requested locale does exist. if it doesn't we have to fall back on `en`.

PR #2859, opened 2018-05-18
Remove initial <p> on Translation Key WYSIWYG

Gerhard-Kanzler:
translationkeys should never wrappet in a <p> tag by default.


brusch:
Good catch! :) 
Thanks!

PR #2862, opened 2018-05-20
Skip linking glossary terms that match custom routes

elenachoi:
Right now glossary terms skip self-linking on matching documents, which doesn't work if using glossary functionality on dynamic content (no underlying document).


brusch:
Thanks for you PR! 
Can you please sign the CLA, so that we can merge your contribution? 
Thanks!

brusch:
Thanks!

PR #2865, opened 2018-05-22
Admin/Email: Escape email subject and text-body

podarcis:
Fixes:
* Escape mail subject
* Escape text email body


brusch:
Thanks!

PR #2866, opened 2018-05-22
Added phpDoc

ThomasKeil:
Added phpDoc for getOrderParentFolder() to avoid false assumption of returning a static, as defined in AbstractObject::getById()


fashxp:
thx

PR #2875, opened 2018-05-28
Fix: getDataFromObjectParam() for localized CS-Data

woellundwoell:
## Calling the REST API does not return the values ​​of localized classification store attributes.

Reproduce:

- Create classification store
- Create classification store as attribute in class and select option "localized"
- Enter data
- REST call ([your.domain] / webservice / rest / object / id / ...)
- Result for all languages in JSON: `value: null`

### DIFF
--- a/pimcore/models/DataObject/ClassDefinition/Data.php
+++ b/pimcore/models/DataObject/ClassDefinition/Data.php
@@ -1289,7 +1289,7 @@ abstract class Data
                 if (method_exists($object, $getter)) {
                     $groupId = $context['groupId'];
                     $keyId = $context['keyId'];
-                    $language = $context['language'];
+                    $language = $params['language'];
 
                     /** @var $classificationStoreData DataObject\Classificationstore */
                     $classificationStoreData = $object->$getter();


weisswurstkanone:
I believe the proper solution would be to populate the context information right here in
ClassificationStore.php line 425

    public function getForWebserviceExport($object, $params = [])
    {
        /** @var $data DataObject\Classificationstore */
        $data = $this->getDataFromObjectParam($object, $params);

Changing this in Data.php would break other calls I guess.

woellundwoell:
The problem we discovered results from the call from ClassificationStore.php line 471:

`$value = $fd-> getForWebserviceExport ($object, ['context' => $context, 'language' => $language]);`

We do not believe that our proposed solution will break other calls because the fix would only affect the ClassificationStore, and the same access will be done just a few lines below (line 1309).
Even simpler could be the fix (line 1296):

`$data = $classificationStoreData-> getLocalizedKeyValue ($groupId, $keyId, $params ['language'], true, true);`

like line 1309.

weisswurstkanone:
I fear it would break the CSV export as indicated in the screenshot. So maybe we can add it to the $params there or add it to the context for the webservice export ?

image

woellundwoell:
Please see PR #3060

PR #2878, opened 2018-05-29
Support for product search with array of strings

### LINKED ISSUE #2877: Support for product search with array of strings
https://github.com/pimcore/pimcore/blob/f8d731aee312cbcb55d792daeaed9abbca9da193/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/IndexService/ProductList/DefaultMysql.php#L629

It would be nice to have the possibility to search with an "... IN ( ... )" Condition with strings in it.

Currently this would lead to an error when DefaultMysql implementation is used:

```php
$factory = \Pimcore\Bundle\EcommerceFrameworkBundle\Factory::getInstance();
$productList = $factory->getIndexService()->getProductListForCurrentTenant();
$remoteIds = ['A001','2','3'];
$productList->addCondition($remoteIds, 'remoteId');
$products = $productList->load();
```

**Problem**: The values which might be string values won't be surrounded as needed in MySQL. The query would be " ... remoteId IN (A001, 2, 3) ... " which leads to an syntax error.

deltatag:
Suggestion as bugfix for #2877. The change may be reviewed to match all possibilities.

Fixes #2877

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/IndexService/ProductList/DefaultMysql.php
+++ b/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/IndexService/ProductList/DefaultMysql.php
@@ -626,7 +626,7 @@ class DefaultMysql implements IProductList
                     }
 
                     $condition .= is_array($cond)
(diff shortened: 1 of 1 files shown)


fashxp:
thx very much. 
can you please sign the license cla.

PR #2879, opened 2018-05-29
Object Grid Improvements

### LINKED ISSUE #2873: Inherited values should be respected in the grid view by grid column operators
Currently they are ignored. This makes it impossible to use operators for inherited values.

daves-weblab:
This pull requests fixes some minor issues with the object grid. 

First of all the fieldnames are not displayed in brackets in the object grid anymore to keep it compact. They are still displayed in the configuration panel, and for all columns except system columns. Previously columns that were displayed due to configuration in the class definition itself, did not contain the fieldname either. Consistency.

Second inherited values are used in the object grid for operators as well which resolves #2873.

Third the widths for columns could be saved in configurations. Operator columns were ignored and always returned to 200 pixels. They are now saved as well.


weisswurstkanone:
thx

PR #2890, opened 2018-06-01
Allow usage of the DataObject/Service::getUniqueKey method variants

BlackbitNeueMedien:
The list used in this method did not take variants into account when checking if a key is already is use. I've added the variant type to the list as well.


brusch:
Thanks!

PR #2895, opened 2018-06-04
Render type for multiselect types (list or tags)

### LINKED ISSUE #1231: Render MultiSelection fields as ExtJs Formtags
# Feature Request 
Represent MultiSelect fields as tags in ExtJs.

## Feature description 
Currently multiselect fields are represented in lists. This can result in very long lists that are difficult to manage if there are many options (example: 250 colors). An alternative representation (ExtJs Tags; see http://examples.sencha.com/extjs/5.0.0/examples/kitchensink/#form-tag) would be a good solution! 

It would be great, if there would be the option in the multiselection field, whether the output is as a list (=default), or as tags. With the current ExtJs version this shouldn't be too complicated.

dvesh3:
Fixes #1231 

Added Render Type option for multiselect data types:

multiselect_rendertype
- List (default): Show options in list view
multiselect_view_list

- Tags: Show options in form tags view
multiselect_view_tags


brusch:
Thanks!

PR #2899, opened 2018-06-05
[DOCS] Paginator example fixes

putzflorian:
change Zend Framework1 Paginator to Zend Framework2 Paginator

### DIFF
--- a/doc/Development_Documentation/05_Objects/03_Working_with_PHP_API.md
+++ b/doc/Development_Documentation/05_Objects/03_Working_with_PHP_API.md
@@ -350,10 +350,10 @@ public function testAction( Request $request )
 <br />
  
 <!-- pagination start -->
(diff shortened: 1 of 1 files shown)


brusch:
I'm afraid that there are some more legacy examples on that page, like the partial script and the `Zend_Db_Select` examples: 

https://github.com/putzflorian/pimcore/blob/258843439a26294f8aefc00bc0c1f9bf4e67d3fc/doc/Development_Documentation/05_Objects/03_Working_with_PHP_API.md#partial-script-includespagingphp

https://github.com/putzflorian/pimcore/blob/258843439a26294f8aefc00bc0c1f9bf4e67d3fc/doc/Development_Documentation/05_Objects/03_Working_with_PHP_API.md#access-and-modify-internal-object-list-query

Would be nice if you could fix them as well. ☺

Issue #1793, opened 2017-08-07
Backend Branding with Custom Logo


ckemptner:
I suggest to make it possible to define a custom client logo (SVG), which will be displayed additionally to the Pimcore logo in the home screen (if there are no custom reports) and in the upper right corner.

This will give the feeling of a customized product to the user/company and increase the bond of affection.

Yivan:
I think too it would be great to bee able to customise it.
For now, i found Pimcore branding a little invasive for enduser, maybe could be relaxed ; )) or at least provide to add/replace some customer logo yes. As it can be seen, the branding is put in 3 places:

![pimcore](https://user-images.githubusercontent.com/9025839/31076439-68c7ce50-a77a-11e7-9e59-e93017953cc1.png)

brusch:
Just to clarify: we don't want that the Pimcore branding is completely replaced. Pimcore is free and open source and it is really not respectful if someone removes the branding completely and doesn't show some attribution to Pimcore by keeping at least a small logo.

Yivan:
I totally agree with you @brusch, and we promote Pimcore for sure! The idea like ckempter sugest is to display additionnal company logo, or like i sugested, having maybe the possibilty to replace one of the three logo.
For sure you must keep Pimcore branding visible as it will promote Pimcore to people using it and talking from it to other people, etc. But having possibility at least to put a company logo somewhere would be great.

fashxp:
that's the plan

andreas-gruenwald:
Hey guys, is there already an update on that?
I would like to try the feature for a new client.

brusch:
![image](https://user-images.githubusercontent.com/142037/34768872-5e8a7eda-f5fc-11e7-92f1-8c8ea65b7db6.png)

![image](https://user-images.githubusercontent.com/142037/34768770-274707fe-f5fc-11e7-8380-cebc86b8e9d8.png)
![image](https://user-images.githubusercontent.com/142037/34768787-2ded464a-f5fc-11e7-9279-0ec03ed6449a.png)

PR #2904, opened 2018-06-06
Added customization to lostpassword template

asysrieke:
Branding is extended to lostpassword page according to login page

## Additional info  
In 5.1.1 Backend branding was added as a new feature so the login page can be customized but the lostpassword page still looked the same, even with activated branding.
https://github.com/pimcore/pimcore/issues/1793


brusch:
Thanks!

PR #2911, opened 2018-06-08
Add ability to use localized website config values

### LINKED ISSUE #2897: pimcore_website_config language parameter missing in twig helper
In pimcore_website_config twig helper the language parameter is missing:

I think this will be the solution:

`    public function getWebsiteConfig($key = null, $default = null, $language = '')
    {
        return Config::getWebsiteConfigValue($key, $default, $language);
    }`

mittererr:
Update to use localized website config values.
fixes #2897


brusch:
Thanks a lot!

mittererr:
Maybe the comment block should be updated (@param null|mixed $language Language to use). Sorry I forgot that. For the sake of completeness ;-).

fashxp:
done

Issue #2923, opened 2018-06-11
QuantityValue: Apply admin translations


janwalther:
When creating QuantityValue units which have different abbreviations in different languages like pieces (pc (EN), stck (DE)) Pimcore admin translations are not applied.

weisswurstkanone:
already merged

PR #2924, opened 2018-06-11
Translate quantity value units

janwalther:
Solves #2923


weisswurstkanone:
Not sure if this is the best place. Would it be better to translate it via the ts javascript function?

weisswurstkanone:
thx

PR #2927, opened 2018-06-11
Fix a typo in the OGone payment provider

jdreesen:
This patch will assign the `$customerName` to the `customerName` key in the authorized data array instead of adding a random boolean value to the key `0`.


fashxp:
ups, thx!

PR #2928, opened 2018-06-11
[Ecommerce] WirecardSeamless payment provider - updated setting payment state to PENDING

jdreesen:
It seems like this line was copied from below. As it results in `IStatus::STATUS_CANCELLED` in either case, we can get rid of the condition.

The condition would never evaluate to `true` anyway, because the `$response['paymentState']` is already checked above and thus can never contain the corresponding value.

---

**Note**: I'm not entirely sure this is the correct fix. Maybe we shouldn't remove the condition but check `$response['paymentState']` for `PENDING` instead and set the status to `IStatus::STATUS_PENDING` in this case?


fashxp:
I would go for your latter approach, we should set the status to pending, if paymentState is pending.

jdreesen:
Okay, I changed the commit to check if the payment state is `pending` and set the status to `IStatus::STATUS_PENDING` in this case.

fashxp:
thx!

PR #2929, opened 2018-06-11
Include fastcgi_params vs. include fastcgi.conf

janwalther:
Once you include fastcgi_params and later fastcgi.conf. Shouldn't this include the same file? Or how are they different?


fashxp:
thx

PR #2933, opened 2018-06-12
[WIP] [Assets] Focal point on images

### LINKED ISSUE #2509: [Assets] Focal point on images
# Feature Request 
Tweak "Select specific area of image" feature. This is already a great feature, which lets the user select the exact part / outer limit of an image, to be rendered in the frontend.

## Problem description 
It is hard for the user to predict how the image will be rendered in the front end, for example when using different thumbnail settings for various responsive states.
 
## Possible solution
One possible solution would be to add an option to set a focus point (e. g. center of a face) combined with a focal cropping option for the thumbnails (+ crop left or right of focus point). I am open to other suggestions for an improvement.

brusch:
resolves  #2509 

+ possibility to set the focus point on asset images (detail view) 
+ focus point is considered when thumbnail transformation `cover` is used 
+ document editable image: show context menu entry to set the focal point if `cover` thumbnail is used


brusch:
### ToDo
- add context menu entry on all image data types and editables (only opens the asset detail view)
- consider focal point in `cover` thumbnail configuration

PR #2946, opened 2018-06-15
Added "globalParams" for areablock

ckogler-elements:
Added "globalParams" parameter to pass values to all areablocks


brusch:
@piotrecky please provide us a reproducible case.
Btw. i don't think it has something to do with `globalParams` i guess this would happen also when passing this as normal / area specific param.

piotrecky:
When I using 'params' instead of 'globalparams' to inject this object to 3 blocks everything works fine.

imo, outputEditmodeOptions function tdoes not need to work on the front site and should also remove unnecessary parameters from 'globalparams' to json just like with 'params'

brusch:
@piotrecky fixed with #3229

PR #2961, opened 2018-06-19
Remove superfluous cast to boolean

jdreesen:
`$autoRender` is already a boolean as per the method signature. There's no need to cast it again.


fashxp:
thx

PR #2962, opened 2018-06-19
fix: added variant object type to BatchJobs

zavodyanka:
How to reproduce an issue:
`objecttype` `variant` is ignored during `Batch edit all` objects in the grid's column


fashxp:
thx

PR #2970, opened 2018-06-20
Fix position of closing parenthesis

jdreesen:
The parameters `get_class($provider)` and `gettype($vars)` were after the closing parenthesis of the `sprintf()` and thus were applied as the parameters `$code` and `$previous` of the `RuntimeException`, which is obviously wrong.


fashxp:
thx

PR #3018, opened 2018-07-04
Add name property to object textareas

pmuskovac:
Set the name property for textareas to allow easier dom access. Input fields already have it, and this makes it easy to access their content from a custom handler script.

## Additional info  
Initially discussed with Christian Fasching


brusch:
Thanks!

PR #3021, opened 2018-07-04
Added locale interface

ascheider:
Added possibility to decorate locale service.

At the moment it is not possible to decorate locale service because of usage concrete Locale class and not the interface.
So added possibility to decorate the service like:
 AppBundle\DecoratingLocale:
      decorates: Pimcore\Localization\Locale
      # pass the old service as an argument
      arguments: ['@AppBundle\DecoratingLocale.inner']

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/01_Within_V5.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/01_Within_V5.md
@@ -2,6 +2,20 @@
 
 
 ## Version 5.3.0
+
+#### Build 277 (2018-07-09)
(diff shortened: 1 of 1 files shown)


brusch:
Thanks for your PR! 
So far it looks good, but the `LocaleInterface` should have defined all methods required by Pimcore.

ascheider:
@brusch ok, you mean all methods which are implemented in locale?
I will update PR

brusch:
@ascheider yep, otherwise when decorating it, it's not guaranteed that everything still works when they are missing in the interface :)

ascheider:
@brusch ok right. One more question. Should i add declare(strict_types=1); to my interface? Locale.php does not have declare(strict_types=1);

brusch:
@ascheider no, not necessary if `Locale.php` doesn't have it

ascheider:
@brusch ok, thanks.
PR is updated. 
Ping me when there are further issues.

PR #3022, opened 2018-07-04
Grid configuration for Object search view

### LINKED ISSUE #290: Grid-Configuration in Search-View not saved
I choose:
Search > objects > select any custom class (p.e. customer) > click on the right-arrow / kotext-menu next to any field (p.e. "id") > column-configurator (Spaltenkonfigurator) > i make my selection > i close the search-window > reopen and select my class > fields are back to default / no way to save the column-setup

dvesh3:
Resolves #290

## Additional info  
Added grid configuration in Object Search View


weisswurstkanone:
Is there a way to rename and share the config similar to the  "regular grid view"?

image

PR #3034, opened 2018-07-06
Fix "Unknown key for a VALUE_NULL in [from]"

kkarkus:
Filter MultiSelectRelation (`Pimcore\Bundle\EcommerceFrameworkBundle\FilterService\FilterType\ElasticSearch\MultiSelectRelation`) doesn't work in admin panel with elasticsearch configuration. 
It doesn't show any value, instead of that there's msg like this in ajax response:
```
"{"error":{"root_cause":[{"type":"parsing_exception","reason":"Unknown key for a VALUE_NULL in [from].","line":1,"col":34}],"type":"parsing_exception","reason":"Unknown key for a VALUE_NULL in [from].","line":1,"col":34},"status":400}"
```

It sends request to es with `from = null` instead of `from = 0`.


fashxp:
thx

PR #3037, opened 2018-07-08
Fix/ecommerce switch index alias elasticsearch

kkarkus:
You can't mix in one action array `remove` and `add` operations, see https://www.elastic.co/guide/en/elasticsearch/reference/5.6/indices-aliases.html

Right now it gets 400 error from elasticsearch.


fashxp:
thx

PR #3038, opened 2018-07-09
Update query table creation to ignore unique constraints

### LINKED ISSUE #2943: [Objects] unique in combination with inheritance breaks query table
when having a data object class with a unique field and activated inheritance, this breaks the query table
- either you get an unique contraint violation when saving child objects that do not overwrite the data
- or either child or parent object disappear from query table

unique constraint must not be set to query table(s) when inheritance is activated, only to store table(s)

dvesh3:
Resolves #2943

## Additional info  
Unique constraint not set to query table(s), only to store table(s)


fashxp:
how does this behave with existing classes?

brusch:
@dvesh3 We need an update-script saving all classes, or even better only the classes that contain a field with a unique constraint.

PR #3047, opened 2018-07-09
Fixes the problem with the escaping "save" button

cwiecekpiotr:
When performing mass actions on the grid there is a problem with the "save" button when adding many elements, eg to a multihref
  
  

Add autoscroll to Window in gridTabAbstract


brusch:
Thanks!

PR #3049, opened 2018-07-09
Autoreload object view/edit tab on save.

kkarkus:
This PR contains new feature, that auto reloads object tab on save.
Why? Sometimes object contains calculated value fields.

### DIFF
--- a/web/pimcore/static6/js/pimcore/object/object.js
+++ b/web/pimcore/static6/js/pimcore/object/object.js
@@ -729,6 +729,8 @@ pimcore.object.object = Class.create(pimcore.object.abstract, {
 
                                 // for internal use ID.
                                 pimcore.eventDispatcher.fireEvent("postSaveObject", this, task);
+
+                                this.reload(this.layoutId);
                             }
                             else {
                                 pimcore.helpers.showPrettyError(rdata.type, t("error"), t("error_saving_object"),


brusch:
Wouldn't it be better to provide a refresh button next to the calculated value field, instead of reloading the entire object?

brusch:
It could be very annoying if this is the standard behavior ... i know a lot of users pressing Ctrl + S every so often ...with that change this would completely interrupt their work, since the reload also causes the UI to reset to the inital state, eg. when working with tabs, localized fields, things like that ...

kkarkus:
@brusch ok, I understand

brusch:
@kkarkus but we're happy to accept a PR for the reload button next to the calculated field, that refreshes only the value of the calculated field ;-)

kkarkus:
@brusch I'll create PR later.

dpfaffenbauer:
Why not provide the new data as ajax response and have the ability for each editable tag to reload data based on the response?

dpfaffenbauer:
this would allow pimcore to do more dynamic stuff on object-save like: changing data in events and reflect that data immediately in the UI

brusch:
Also a good idea, but would possibly result in a bigger change, but would be worth to evaluate.

PR #3050, opened 2018-07-09
New feature: Added thumbnail download option

adrozdek:
Thumbnail Download option in asset's view:
thumbnail-download

There will be an error message: "No thumbnail selected" if thumbnail was not selected.


brusch:
Good idea! But maybe we should introduce a new option on the thumbnail definition, that causes the thumbnail to be listed in this new download selection. Because there are projects with over a hundred thumbnail definitions, and then this functionality gets quite confusing :) 

What do you think? 

image

adrozdek:
I think it's a good idea :) I will add this checkbox tomorrow or in a few days.

brusch:
I'd also move this box between _Download_ and _Custom Download_

adrozdek:
Okay

adrozdek:
@brusch I pushed these changes. Could you check again? :)

brusch:
Thanks!

PR #3051, opened 2018-07-09
Added option to customize object add dialog box 

### LINKED ISSUE #2772: New Object Modal - custom text
Generally my question is: 
Is it possible to add custom text to the New Object Modal window based on the Class?
image

It would be nice to be able to add custom hint-texts to the Base definition of a Pimcore Class to display within this modal.
If there is no method currently (as I mentioned I couldn't find anything related), here's my feature request:

## Feature description 
I want to be able to add custom text to a base definition of a given class to be displayed within the "Create New Object" modal.
This should open up the ability to e.g. directly add naming conventions for users to remember, or add a reminder to users to make sure a similar object does not already exist, etc.
image
The custom text could be displayed in italic below the input field.

kubaplas:
Resolves #2772

## Additional info  
Keys will show up in Admin Translations for each class, after displaying dialog for the first time:
screenshot_20180709_143500

If it's translated it will replace object add dialog (after ui reload):
screenshot_20180709_143525

### DIFF
--- a/web/pimcore/static6/js/pimcore/object/tree.js
+++ b/web/pimcore/static6/js/pimcore/object/tree.js
@@ -929,7 +929,13 @@ pimcore.object.tree = Class.create({
             dialogText =  t('please_enter_the_name_of_the_new_object');
         }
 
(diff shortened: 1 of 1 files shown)


brusch:
Great - thanks!

PR #3053, opened 2018-07-09
Extended valid order keys

ThomasKeil:
- extended the valid order keys for carts by "creationDateTimestamp" and "modificationDateTimestamp"
- added PhpDoc


fashxp:
thx

PR #3059, opened 2018-07-10
Proposal for the open icon

IgorBenko:
Multiple of our clients were really confused by the open icon in Pimcore (the cursor icon) - to the point that we had to modify Pimcore to make this understandable for their employes.

This Pull Request is a proposal for a new open icon. The icon was very well received by our client and they said that this icon immediately gave the sense of what the button does, contrary to the cursor icon.


brusch:
This is how the icon looks like: 
image

brusch:
Changed icon to: 

image

IgorBenko:
@brusch Thanks! 👍

PR #3060, opened 2018-07-10
Fix: getDataFromObjectParam for localized CS-Data

woellundwoell:
## Calling the REST API does not return the values ​​of localized classification store attributes.

Reproduce:

- Create classification store
- Create classification store as attribute in class and select option "localized"
- Enter data
- REST call ([your.domain] / webservice / rest / object / id / ...)
- Result for all languages in JSON: `value: null`
 
Fix for PR #2875 in file Pimcore\Model\DataObject\ClassDefinition\Data\Classificationstore.php as proposed in https://github.com/pimcore/pimcore/pull/2875#issuecomment-393080149

Sorry for the delayed response!


weisswurstkanone:
thx!

PR #3061, opened 2018-07-10
Feature/object tab refresh data on save

kkarkus:
@brusch According to this PR https://github.com/pimcore/pimcore/pull/3049 
I made some changes. 
Component that listens to `postSaveObject` event (currently only calculatedValue, input and wysiwyg), it sets once again value from `this.object`. 
I tried to make it as easy as it can be, so implementing other compontents should be easy enough.
I'm not able to implement all components right now, I don't have enough time.

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/DataObjectController.php
+++ b/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/DataObjectController.php
@@ -400,8 +400,8 @@ class DataObjectController extends ElementControllerBase implements EventedContr
 
         $objectData['general']['o_locked'] = $object->isLocked();
 
-        ['objectData' => $objectData, 'metaData' => $metaData] = DataObject\Service::getDataForObject($object, $objectFromVersion);
-        $objectData['data'] = $objectData;
+        ['objectData' => $data, 'metaData' => $metaData] = DataObject\Service::getDataForObject($object, $objectFromVersion);
+        $objectData['data'] = $data;
 
         $objectData['metaData'] = $metaData;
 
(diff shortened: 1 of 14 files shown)


dpfaffenbauer:
looks quite nice, what I also would recommend is to decouple the functions ```getDataForObject``` and ```getDataForField``` to a separate Service in order to make them useful for other use-cases as well. I current have these use-cases and had to copy the functions which is not a good practice. The reason I had to copy them was for the embedded classes editable type. (https://github.com/coreshop/CoreShop/blob/master/src/CoreShop/Component/Pimcore/DataObject/EditmodeHelper.php)

kkarkus:
@dpfaffenbauer according to you advice I moved this two methods from controller to `\Pimcore\Model\DataObject\Service`
@brusch is it ok?

kkarkus:
Currently supported components:
- input
- lastname
- firstname
- email
- select
- gender 
- country
- user
- language
- persona
- targetGroup
- quantityValue
- inputQuantityValue

@brusch I'm going to finish this PR until end of the week I think.

brusch:
@kkarkus please wait with further work until we've made a decision, makes more sense ;-)

kkarkus:
@brusch ok, please let me know what's your decision when you are ready

dpfaffenbauer:
I would also not use the postObjectSave event, as every object fires that event. I would introduce a new function called "setData" or something that specifically only gets the data set for the current object it is in.

But as brusch said, this is a huge and dangerous refactor and could have potential side effects. On the other hand, I really would appreciate this, as it opens the doors for other editable types or improvements on current types.

BTW: This would also make the reload function more responsive, as it would not need to reload the whole tab, but just the data.

dpfaffenbauer:
@kkarkus but before changing anything, wait what @brusch has to say

brusch:
See follow-up issue #4678

PR #3069, opened 2018-07-12
Filterable Admin Languages

daves-weblab:
This PR makes the list of admin languages filterable via an event. This can be used to reduce the amount of languages visible in the backend, if not all languages are necessary.


fashxp:
event for that functionality feels somehow wrong. We would prefer a symfony configuration option. see also https://github.com/pimcore/pimcore/issues/3090

PR #3076, opened 2018-07-13
Add a event handler before the logic is called

jorisros:
Resolves #
The problem was that we only had influence on the 'data' array, which was the result of lots of logic, so I added a event handler before the logic is called, so i can have influence from object level.

### DIFF
--- a/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/DataObjectController.php
+++ b/pimcore/lib/Pimcore/Bundle/AdminBundle/Controller/Admin/DataObjectController.php
@@ -340,6 +340,13 @@ class DataObjectController extends ElementControllerBase implements EventedContr
         $objectFromVersion = $latestObject === $object ? false : true;
         $object = $latestObject;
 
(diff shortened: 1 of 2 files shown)


brusch:
@jorisros using the `AdminEvents::OBJECT_GET_PRE_SEND_DATA` event is not an option? 
If not, what is the practical use case for this new event? 

Thanks for clarifying!

jorisros:
@brusch I tried to do 'prefill' of some fields of a object. But I have to do that in the data array. Because $object->setSomeVariable('prefill text') doesn't work because the return of the object is never handled. 
When I do it on the 'data' array I often have to rerun the logic of generating the layout. When I can call it before, I can do a object manipulation and then it will been used in building the layout array.

brusch:
Hmm, probably I'm not getting your point, but prefilling data should be easily possible with the existing event, I can't see any advantage of the additional event. 
Can you please give me a more concrete use-case, ideally with a small code example. 
Thanks!

brusch:
@jorisros any updates on that?

PR #3093, opened 2018-07-17
Fix generated PhpDoc of ObjectBricks

jdreesen:
A PhpDoc with `Set ...` was created for getters.

This also fixes the indentation of the return statement for encrypted fields.


fashxp:
thx

PR #3103, opened 2018-07-19
update Date & Date Time format for Object edit view

### LINKED ISSUE #3075: Improvement : Change in Date Format in Date and Date Time Field..
Improvement 

Change in Date format in Date and Date Time Field : Make Date Format as  dd/mm/yyyy instead of dd/mm/yy as sometimes the selected date does not tell the actual year for 02/04/12 can be 02/04/2012 or 02/04/2112 or 02/04/1912  ..  etc.

date format

If the Date field is editable we can click on the date and check the actual year but if the date or date time field is non editable then there is a confusion in the year..

Also we can give the option of format in specific settings in date and date time fields....

dvesh3:
Resolves #3075


brusch:
Thanks!

PR #3105, opened 2018-07-19
Replace prototype-light.js

### LINKED ISSUE #2974: Replace prototype-light.js
Following on from this PR: https://github.com/pimcore/pimcore/pull/2969

This library does all kinds of nefarious things that can cause end user code to fail that would not fail in a vanilla environment, and will continue to cause problems in the future.

I see a few uses of Object.extend (that could easily be replaced with Object.assign, or $.extend or whatever it is in ExtJS), but not much else code appears to be using it after that.

It doesn't seem like keeping this library around will have a positive impact - is there another reason why it should be kept?

kirjavascript:
Resolves #2974 

## Additional info  
I took the route of replacing usage of prototype.js with native methods if they're available, or with ExtJS equivalents if not.


brusch:
Awesome - thanks!

dpfaffenbauer:
Object.extend doesn't work anymore now right?

brusch:
No, you have to replace it by `Object.assign` or you use `Object.extend = Object.assign;`

brusch:
But we can think about a compatibility layer that does that for BC reasons, ...

dpfaffenbauer:
So this breaks then ;)

Yes, compatibility layer with deprecation note would be good

brusch:
Sure.

kirjavascript:
`Object.extend = Object.assign` would still be potentially problematic in future - unless this is part of the compat layer you are talking about? Would that layer be optional?

Alternatively I guess you could just be talking about user code.

brusch:
@kirjavascript Sorry, the merge was too overhasty. 
I had to revert it since there seem to be some issues, especially with classes and scopes. 

Example: 
```
TypeError: Cannot read property 'bind' of undefined
    at Function.getLayout (properties.js?_dc=285:64)
    at Function.<anonymous> (class.js?_dc=285:67)
    at Function.getLayout (properties.js?_dc=285:38)
    at klass.getLayout (class.js?_dc=285:22)
    at klass.getTabPanel (page.js?_dc=285:72)
    at klass.addTab (page_snippet.js?_dc=285:35)
    at klass.getDataComplete (document.js?_dc=285:44)
    at Object.callback (ext-all-debug.js?_dc=285:8174)
    at constructor.onComplete (ext-all-debug.js?_dc=285:24674)
    at constructor.onStateChange (ext-all-debug.js?_dc=285:24631)
```

kirjavascript:
No worries, I tried to test enough to ensure it was stable but seemingly I have missed some things. I guess this kind of change will be a bit of a bumpy ride. 

Specifically, with bind() I left it alone because it's already on the function prototype so I thought this would be sufficient. Apparently not.

I can address any other issues that are posted and try and ensure my testing is a little more extensive.

brusch:
@kirjavascript you can open a new WIP pull request, where we can track the progress and do testings ... thanks!

kirjavascript:
Ok, sounds good!

PR #3106, opened 2018-07-19
Replace prototype-light.js

### LINKED ISSUE #2974: Replace prototype-light.js

kirjavascript:
Resolves #2974

## Additional info  
Work in progress - additional bugfixes to follow

See [the previous PR](https://github.com/pimcore/pimcore/pull/3105) for more information


kirjavascript:
Having trouble reproducing that, how did you trigger the error?

dpfaffenbauer:
Just for reference: `Object.extend` is important cause of BC. Plugins may use that function and removing it will cause issues.

kirjavascript:
@dpfaffenbauer is a solution for that simply providing an optional backwards compat shim?

(It would literally just be a small variation on the previous prototype-light.js file)

dpfaffenbauer:
@kirjavascript well, but Pimcore needs to require that shim, just providing it doesn't help, it would still be a BC break.

brusch:
@kirjavascript are you able to reproduce it?

kirjavascript:
@brusch yeah, I don't know why it wasn't breaking before - it's actually easy to replicate

in the conversion process I seemingly lost the function context at some point - should be fixed now

brusch:
@kirjavascript Did a lot of testing and it seems it's all done now. Good job!

brusch:
added upgrade notes regarding BC breaks

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/01_Within_V5.md
```diff
@@ -1,6 +1,26 @@
...
+The following list should help you to locate and replace them, if available you can find the replacement in brackets: 
+- `Prototype.*` functions and properties 
+- `Enumerable.*` functions and properties 
+- `$A()` ( `Array.from()` ) 
+- `$w()` 
+- `Object.extend()` ( `Object.assign()` ) 
+- `Object.toQueryString()` ( `Ext.Object.toQueryString` )
+- `Object.clone()` ( `Object.assign({}, object)` )
+- `Object.toHTML()`
+- `Object.is[Element|Array|Hash|Function|String|Number|Undefined]()`
```
dpfaffenbauer: Replacement could partly come from ExtJS.

Ext.isArray, Ext.isFunction, Ext.isString, Ext.isDefined and so on.

brusch: Thanks for the hint, added them in detail

dpfaffenbauer:
I still don't like the BC break, but it is not too bad and it is documented. So there is nothing more to add from my side.

kirjavascript:
Conflicts resolved.

Added alternatives for Array methods into the docs. 

As for the Function prototype methods I'm not able to recreate some of the things in a concise way, so if alternatives cannot be found in Ext we could possibly point to another library that contains them?

This might not be important though, since I imagine these methods are rarely if ever used

brusch:
@kirjavascript I've changed the base of this PR to the `pimcore:v5.4.0-dev` branch, so we can merge it. Unfortunately your fork is outdated and doesn't have the `v5.4.0-dev` branch, so I'm unable to resolve the conflicts, can you help me out once again? 

Thank you very much!

brusch:
Yep, I think thats necessary to resolve the conflicts.

brusch:
@kirjavascript Thanks, I was now able to merge it - so nothing else to do for you.

PR #3107, opened 2018-07-19
add twig extension for pimcore translate helper and update documentation

jms-gsulcer:
Resolves #

Add twig extension for the pimcore translate helper and update the documentation for Shared Translations

### DIFF
--- a/doc/Development_Documentation/06_Multi_Language_i18n/04_Shared_Translations.md
+++ b/doc/Development_Documentation/06_Multi_Language_i18n/04_Shared_Translations.md
@@ -46,6 +46,16 @@ pimcore:
 </div>
 ```
 
+```twig
+<div>
+    <!-- there is a twig extension for the pimcore translate helper -->
+    <address>&copy; {{ pimcore_translate("copyright") }}</address>
+    <a href="/imprint">{{ pimcore_translate("imprint") }}</a>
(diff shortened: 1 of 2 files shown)


brusch:
`{% trans %}Hello %name%{% endtrans %}` is not working?

jms-gsulcer:
No, we have not been able to make that work. Twig expects %name% to be a twig variable, which it isn't, so we get the twig error on the page "Variable %name% does not exist."  We can make it a variable and assign it from controller information, but that seemed extra work.

Also, it seemed simpler for training our template designers to use pimcore_translate since they were already familiar with pimcore 4 helpers and we told them they could generally just add "pimcore_" and it would work, which is mostly true, but wasn't for translate.  This seemed simpler and less verbose to us.

brusch:
How other Twig users are dealing with that, any suggestions? Thanks!

dpfaffenbauer:
```{{ 'blub'|trans }}```

**Comment by dpfaffenbauer** [comment]
that easy ;)

**Comment by jms-gsulcer** [comment]
Wow, thanks, that's simple!  We didn't see it in the documentation anywhere.

**Comment by dpfaffenbauer** [comment]
check twig documentation for useful things as well ;)

**Comment by aarongerig** [comment]
FYI: If you need to add something to admin translation with twig, here's the syntax for it:

```twig
{{ 'blub'|trans({}, 'admin') }}
```

PR #3108, opened 2018-07-19
[Ecommerce] Replace google maps with leaflet on order detail page

### LINKED ISSUE #2995: [Ecommerce] Replace Google Maps Geocoding in Order Detail
see https://github.com/pimcore/pimcore/blob/9cb9858134cbd6c3773baa20fc3f5615d7f92551/pimcore/lib/Pimcore/Bundle/EcommerceFrameworkBundle/Controller/AdminOrderController.php#L185 

Requests won't work anymore without an API key after July, 16 2018

dvesh3:
Resolves #2995


brusch:
Thanks!

PR #3118, opened 2018-07-23
use more open constraint for ocramius/package-versions

dpfaffenbauer:
Use a more open version constraint for ocramius/package-versions. This allows to use newer versions of the package and thus conflict in less version conflicts.


brusch:
Thanks!

PR #3121, opened 2018-07-23
[Updater] Use Composer for Updates

### LINKED ISSUE #3120: [Updater] Use Composer for Updates
- Up / down migrations
- Pimcore as Composer dependency - #2687

brusch:
resolves  #3120


dpfaffenbauer:
Just saw the new File structure, wouldn't it be better to have something like

`lib` -> `src/Pimcore`
`models` -> `src/Pimcore/Models`
`bundles` -> `src/Pimcore/Bundles`

brusch:
@dpfaffenbauer yes, thought about that as well, ... but then i also though i might be useful to structure the code a bit more and reduce the amount of folder levels ...

dpfaffenbauer:
Well, but best practice is to have a `src` Folder ;) So it clearly states where the application code is to be found.

brusch:
Let's see, after Pimcore is a composer library such changes are quite easy, even in future 😉

brusch:
For those who'd like to play around already ... here we go: 
```
composer create-project pimcore/demo-basic my-project dev-master
cd ./my-project
./vendor/bin/pimcore-install
```
(you can ignore the database warnings 😉)

.travis/setup-tests.sh
```diff
@@ -2,11 +2,23 @@
...
+cp .travis/extensions.template.php /var/www/var/config/extensions.php
+cp app/config/parameters.example.yml /var/www/app/config/parameters.yml
+
+cp .travis/composer.local.json /var/www/composer.local.json
 
 # install composer dependencies
-composer install --no-interaction --optimize-autoloader
+
+cd /var/www
+composer install --dev --no-interaction --optimize-autoloader
```
dpfaffenbauer: this never install the right version of Pimcore from the current commit/branch/PR, better use something like this:

```
composer req pimcore/pimcore:dev-$TRAVIS_BRANCH#$TRAVIS_COMMIT
```

lib/Composer.php
```diff
@@ -60,6 +71,17 @@ public static function postUpdate(Event $event)
         $rootPath = self::getRootPath($event);
         self::parametersYmlCheck($rootPath);
         self::zendFrameworkOptimization($rootPath);
+        self::executeMigrationsUp($event);
```
dpfaffenbauer: is it really a good idea to run migrations on composer update? Sometimes there is no database available during this step like when building docker containers with the source-code included. Database would then only be available on runtime of the container.

brusch: If it fails during the composer update nothing happens (except an error message) you can perform the migrations anytime after the upgrade with the following command: 
`./bin/console pimcore:migrations:migrate -s pimcore_core -n` 
Other projects are doing it the same way, I don't see any problems with that so far. 

Btw. you are free to modify your `composer.json`, if you're having problems with that, just remove the event listener and run the scripts manually.

brusch: Our aim is to make the upgrading procedure as simple as possible, but still giving the possibility to customize it for your needs ... I think with the current solution we're fulfilling both topics.

dpfaffenbauer: if it fails during the update, and exit code != 0 happens, which means for docker (or other deployment solutions) that something failed.

Modifying the `composer.json` isn't a real solution, cause it removes other Pimcore tasks from composer as well.

brusch: Only the `parameters.yml` check and the deprecated ZF optimization, ... but anyway :)  
If it makes you feeling more comfortable, we can move the migrations into a separate `post-update-cmd` part - I can live with that ;-)

dpfaffenbauer: And running two commands: 

- ```composer update``` and 
- ```bin/console pimcore:migrations:run```   (just an example)

isn't too bad. In other Symfony projects, you first need to copy migrations files from different Bundles to run them, so having Pimcore Migrations is a huge win anyway.

dpfaffenbauer: @brusch Sounds like a compromise I can live with :)

dpfaffenbauer: @brusch And regarding the other tasks, you don't know today what will change there. Maybe more tasks will come and changing that then will get complicated ;)

brusch: @dpfaffenbauer I think splitting them up makes sense anyway, because the migrations should run after the symfony scripts not prior them.

dpfaffenbauer: alright, happy with that 👍

foobar13372:
@dpfaffenbauer Why is it "best practice to have a src folder"? Please explain! Or a reliable source for this claim? Or is it even a law?

dpfaffenbauer:
@foobar13372 its just what people do. there is no PSR for that, it just "best practice". Also take a look at this: http://php-pds.com/

foobar13372:
@dpfaffenbauer You linked a random website written by some random guy (actually there's not even any author named on this site). Don't treat code as a "law by nature". Code is human-made, it's highly arbitrary. And the code author is free to decide what works best for him.

And obviously it's not "what people do" - as you can see in this very pull request which lead you to this unsourced claim.
It also heavily depends which framework you use. TYPO3, for example, does not use this structure. And it's by far spread wider than Pimcore.

PR #3122, opened 2018-07-23
fix button ordering for video editmode panel

dpfaffenbauer:
Video Edit UI is not consistent with brick editmode UI:

bildschirmfoto 2018-07-23 um 11 32 57

With this PR, it looks like this, so button ordering is the same:

bildschirmfoto 2018-07-23 um 11 33 20


brusch:
Thanks!

PR #3123, opened 2018-07-23
Fix for date filter 'ON' not fetching correct results

### LINKED ISSUE #3071: Date filter is not working on custom report .
### Expected behavior 
 Date filter is working for timestamp also or formatted date
### Actual behavior  
  Custom report is saving date in timestamp and cutsom report sent params in api like (

filter | [{"operator":"eq","value":"06/27/2018","property":"o_creationDate","type":"date"}]
-- | --
)

so it is not working
### Steps to reproduce  
  1) create a test custom report for query (select o_creationDate from objects) 
   2) Apply filter for date 
   3) Save and check for date filter.
   4) It is not working for exact date ie "ON" [filter]
(url
customreport
)
### Code example 
```php
// your code
```

dvesh3:
Resolves #3071


brusch:
Thank you!

PR #3134, opened 2018-07-25
[Application Logger] shows html piece as "text/plain"

### LINKED ISSUE #3133: [Application Logger] shows html piece as "text/plain"
My logged exception via ApplicationLogger is wrapped into a `<pre>` block and don't have a surrounding `<html>` tag.
In this case the exception is interpreted as  `Content-Type = text/plain` and really hard to read.

lgaidzik:
logged exceptions by the ApplicationLogger which are wrapped inside a `<pre>` tag were displayed as `text/plain`

Resolves #3133


brusch:
Thanks!

Issue #2496, opened 2018-01-23
Content Inheritance: Document editable is editable although it is not explicit unlocked


nspacex:
# Bug Report
## Please make sure all the following boxes are checked before submitting your issue - thank you!
- [x] This is not a security issue / vulnerability -> send them to security@pimcore.org instead
- [x] Behavior is reproducible on http://demo.pimcore.org/admin/ (admin/demo) 
or on a clean install of the latest build (out of your project, to avoid problems with custom code, plugins, ...) 
- [x] There's no existing ticket for the same issue
  
### Expected behavior 
Element shoult not be editable until it is unlocked by "right click"
  
### Actual behavior  
When a user clicks on "highlight editable elements", all previously locked elements are unlocked. Editing is possible but the changes will not be saved, even if he clicks "save".  

### Steps to reproduce  
1) Open document with locked and inherited fields (like wysiwyg)
2) Click on "highlight editable elements" (unlocks all fields; click "highlight editable fields" again all grey lock masks disappeared
3) Edit content and save document 
4) Reload document

PR #3137, opened 2018-07-25
Image is possible to add on locked field and Highlight editable elements unlocks locked fields fix

### LINKED ISSUE #3087: Image is possible to add on locked field
It is possible to add an image to an Image Editable when the Editable is locked from the Master Document.

harpreet819:
Resolves #3087 #2496 
Image is not possible to add now on locked fields. Before, video edit button was appearing before clicking overwrite button, now it will show only if user clicks overwrite button (same as other dnd fields like link,multihref). Also highlight editable elements does not unlock previous locked elements.


bundles/AdminBundle/Resources/public/js/pimcore/document/edit.js
```diff
@@ -166,7 +165,7 @@ pimcore.document.edit = Class.create({
                     mask.setStyle("opacity","0.5");
                     mask.setStyle("pointer-events","none");
                 } else {
-                    ed.unmask();
+                    editables[i].setInherited(editables[i].getInherited());
```
brusch: Why have you changed that? 
Now masked items are not unmasked anymore, is this intentional?

harpreet819: Actually ed.unmask() is removing inherited elements grey lock masks as well. So for unmasking items, instead of using ed.mask(), we can use setInherited method() to bring editables back to their state they were before user clicks "highlight editable elements". I tested it for both documents having master document and not having master document.

PR #3141, opened 2018-07-26
Fix Converter file name is only to the first Whitespace

Gerhard-Kanzler:
File in var/tmp
Before:
video-converter-Rosadira

now:
video-converter-Rosadira 2018.mp4


brusch:
Mille Grazie!

PR #3142, opened 2018-07-26
Add expiry date for Tag & Snippet Management

### LINKED ISSUE #2123: [Tag & Snippet Management] add expiry date functionality

dvesh3:
Resolves #2123


pimcore/lib/Pimcore/Bundle/CoreBundle/Command/MaintenanceCommand.php
```diff
@@ -100,6 +100,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $manager->registerJob(Job::fromMethodCall('imageoptimize', '\\Pimcore\\Model\\Asset\\Image\\Thumbnail\\Processor', 'processOptimizeQueue'));
         $manager->registerJob(Job::fromMethodCall('cleanupTmpFiles', '\\Pimcore\\Tool\\Housekeeping', 'cleanupTmpFiles'));
         $manager->registerJob(Job::fromMethodCall('cleanupSymfonyProfilingData', '\\Pimcore\\Tool\\Housekeeping', 'cleanupSymfonyProfilingData'));
+        $manager->registerJob(Job::fromMethodCall('markexpiredtagsdisabled', '\\Pimcore\\Model\\Tool\\Tag\\Config', 'markExpiredTagsAsDisabled'));
```
brusch: I think that's not really necessary, since tags are expired on the fly as well, it's very unlikely that this job ever disables a tag. Or did I get something wrong?

fashxp: what if, frontend is not called for a longer period of time ... then tags are never disabled, which might cause confusion and data inconsistency.

brusch: Well, they are called with every single frontend request ... but we could leave it as it is, it doesn't hurt.

PR #3145, opened 2018-07-27
Pimcore symfony constants

sebwebdev:
The file `app/constants.example.php` suggests, that it is possible to overwrite the default bundle, controller and action by defining the corresponding constants:

- `PIMCORE_SYMFONY_DEFAULT_BUNDLE` (defaults to `'AppBundle'`)
- `PIMCORE_SYMFONY_DEFAULT_CONTROLLER` (defaults to `'default'`)
- `PIMCORE_SYMFONY_DEFAULT_ACTION` (defaults to `'default'`)

In practice this does not work quite as expected. In case of `PIMCORE_SYMFONY_DEFAULT_BUNDLE` there are several hard-coded fallbacks. In case of the other two constants the default values are set via `system.php` upon document creation. This PR aims to resolve these issues by defining defaults in `Bootstrap.php` and thus enabling overwrites via `app/constants.php` or `.env`.

**Steps to reproduce:**
- Set custom values for the aforementioned constants in `app/constants.php`
- Create a new document
- Exception is thrown in Edit tab: `Unable to find template "AcmeAppBundle:Default:default.html.twig”`. This indicates that the configured controller and action are not respected. Also the template path for the bundle is not reset (see Pimcore/Bundle/CoreBundle/EventListener/TemplateControllerListener.php:110)


sebwebdev:
Seems that a major re-sturcuting was just merged. Closing this PR as of now.

Issue #3150, opened 2018-07-30
webservice import: videos in DataObjects are not implemented


LosHawlos:
When importing DataObjects using the webservice REST-API no videos are imported.
This is because the import for video data is currently not implemented and silently ignored.

https://github.com/pimcore/pimcore/blob/db6d1e3aea4673353d1263ea5379421b91fccd22/pimcore/models/DataObject/ClassDefinition/Data/Video.php#L459-L464

In cases of unimplemented Features the webservice should throw an Exception.

I have partially implemented this for Pimcore 5.3 to at least import vimeo videos.

PR #3151, opened 2018-07-30
Partially implement webservice import for DataObjects containing videos

LosHawlos:
This patch for bug #3150 is against the pimcore-5.3.x branch an partially implements importing videos in DataObjects using the webservice api.

I have only implemented the video types that contain a simple ID (vimeo, youtube) and only tested vimeo videos.

Asset videos are not implemented because I dont know how to implement the idMapping.


brusch:
Unfortunately we're not accepting any further PRs for the 5.3.x branch. Please reopen the PR on the master branch.

Thank you very much!

PR #3153, opened 2018-07-30
Add setter for maxWriteToCacheItems

BlackbitNeueMedien:
During a CLI mass import we want to be able to add items to the cache. Currently this fails if there are more than 50 objects / assets related in the currently imported object (hard coded limit in CoreHandler). This PR adds a setter for this field.


brusch:
I wouldn't change the interface, since this isn't necessary at all for the functionality and would potentially break BC.

PR #3157, opened 2018-07-30
Add parent class and traits for listing data object model

Yivan:
Hello,

This PR add parent class and traits for listing data object model.

It resolve https://github.com/pimcore/pimcore/issues/1618 (1 year ago i know, long time to provide this PR ;-) ). But as we are going to Pimcore 5.4 i think it is good to release this feature in the same time.

It should not introduce any BC break and i tested import/export class too.

Documentation has been updated considering this feature will be available in Pimcore 5.4.0.

Thanks.

### DIFF
--- a/models/DataObject/ClassDefinition.php
+++ b/models/DataObject/ClassDefinition.php
@@ -76,9 +76,21 @@ class ClassDefinition extends Model\AbstractModel
     public $parentClass;
 
     /**
+     * Name of the listing parent class if set
+     *
+     * @var string
+     */
+    public $listingParentClass;
+
+    /**
      * @var bool
      */
     public $useTraits;
+
+    /**
+     * @var bool
+     */
+    public $listingUseTraits;
 
     /**
      * @var bool
@@ -437,6 +449,13 @@ class ClassDefinition extends Model\AbstractModel
         }
         File::put($classFile, $cd);
 
+        // create class for object list
+        $extendListingClass = 'DataObject\\Listing\\Concrete';
+        if ($this->getListingParentClass()) {
+            $extendListingClass = $this->getListingParentClass();
+            $extendListingClass = '\\'.ltrim($extendListingClass, '\\');
+        }
+
         // create list class
         $cd = '<?php ';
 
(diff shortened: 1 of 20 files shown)


Yivan:
Here a screenshot:

parent_traits

models/DataObject/ClassDefinition.php
```diff
@@ -450,9 +469,14 @@ public function save($saveDefinitionFile = true)
         $cd .= ' * @method DataObject\\'.ucfirst($this->getName())."[] load()\n";
         $cd .= ' */';
         $cd .= "\n\n";
-        $cd .= 'class Listing extends DataObject\\Listing\\Concrete {';
+        $cd .= 'class Listing extends '.$extendListingClass.' {';
         $cd .= "\n\n";
 
+        if ($this->getListingUseTraits()) {
```
dpfaffenbauer: It says traits (plural), so it should allow more than one trait right? this implementation only allows one trait. I would either change the property name or split the variable on char `;` pr `,` and run it in a loop.

brusch:
@Yivan thank you very much for your PR. 
We'll discuss this PR internally, reason: 
Actually we had already discussions about getting rid of the parent class/trait functionality in favor of the  class overrides (which does in fact offer the same functionality). 

We'll keep you updated. 
Btw. it's always a good idea to contact us prior working on a PR to avoid such situations ;)

dpfaffenbauer:
@brusch that sounds dangerous. Class Overrides are a way more magic than parent-classes. Parent Classes offers your the possibilites to extend from AbstractClasses that implement certain interfaces. Class Overrides don't allow to extend from AbstractClasses.

fashxp:
thx

Yivan:
@fashxp Thanks for merging

@brush yes i know i should have ask if PR is always ok to provide as the related issue where your team said i can provide a PR was one year ago. Next time in such case i will ask again before ; )
Like i explained in the comment issue https://github.com/pimcore/pimcore/issues/1618#issuecomment-373699413, i think extends is different than override, and our developpers use extend/traits a lot because you can simply make a `new DataObject\Myclassname();` and you get the extending method by parent class or traits. With override, you need to work with static method (getList, getById, etc...).
I really like this feature, and removing it will break many of our projects as `new DataObject\Myclassname();` will no more have access to the custom methods.
I really hope you won't remove parent/traits functionnality! (or at least make a survey before)

Thanks a lot.

fashxp:
no worries, we won't ;-)

PR #3160, opened 2018-07-31
Added option to set CSV settings on Import

### LINKED ISSUE #2705: Add option to set CSV settings on import
Sometimes Pimcore can not guess the proper CSV Settings for a data import. All data will be in one row in this case. It'd be great to have the option to set the CSV settings (e.g. delimiter sign) on import if needed.

dvesh3:
Resolves #2705


brusch:
This removes the auto-detection completely, right? 
Maybe we should offer this still as an option and prefill it with the auto-detected values?

dvesh3:
This is will not remove auto-detection completely. On csv upload, auto-detection will prefill csv settings and user can override it. We are basically taking csv settings from js form for import action.

brusch:
Thanks!

PR #3164, opened 2018-08-01
Show unpublished status in relation fields fix

### LINKED ISSUE #1903: Show unpublished status in relation fields
## Usability Improvement
I think it would make sense to be able to see the publish status of linked objects and documents within relation fields, as it would most likely cause the status of the currently opened object to be invalid, even if the currently opened object itself is published.

### Expected behavior 
After selecting an unpublished object in a href or objects relation of another object, it should be displayed in unpublished style (strike-through) within the relation field.

### Actual behavior  
There is no visible difference whether the object is published or unpublished.

publish-status

harpreet819:
Resolves #1903 

Selecting unpublished object or document in relations fields will be displayed in unpublished mode( full path or reference field marked strike-through)


bundles/AdminBundle/Resources/public/js/pimcore/object/tags/href.js
```diff
@@ -349,6 +356,10 @@ pimcore.object.tags.href = Class.create(pimcore.object.tags.abstract, {
         this.data.type = data.type;
         this.data.subtype = data.subtype;
         this.dataChanged = true;
+        this.component.removeCls("strikeThrough");
+        if(data.published === false || data.published === "0") {
```
weisswurstkanone: what is the reason for check against the string "0"

weisswurstkanone:
Hi @harpreet819 , in addition to my comments, could please resolve the conflict?
Many thanks!!!

fashxp:
does not work ... 
invalid css ... see https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Resources/public/css/admin.css#L1879 

with invalid css, items are never visualized as unpublished.
with valud css, items are always visualized as unpublished.

@harpreet819 can you provide a new pull request that fixes these issues?

PR #3180, opened 2018-08-03
Added is_object check, to prevent fatal error on backend search.

ascheider:
Sometimes $data is a string and not object:

object(Carbon\Carbon)#11507 (3) {
  ["date"]=>
  string(26) "2017-03-09 00:00:00.000000"
  ["timezone_type"]=>
  int(3)
  ["timezone"]=>
  string(13) "Europe/Berlin"
}

string(13) "NOT SUPPORTED"

This causes fatal error:
Call to a member function getTimestamp() on string {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Call to a member function getTimestamp() on string at pimcore/models/DataObject/ClassDefinition/Data/Date.php:200)"} []

  
  

Resolves #

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Date.php
+++ b/models/DataObject/ClassDefinition/Data/Date.php
@@ -196,7 +196,7 @@ class Date extends Model\DataObject\ClassDefinition\Data
      */
     public function getDataForGrid($data, $object = null, $params = [])
     {
-        if ($data) {
+        if ($data && is_object($data)) {
             return $data->getTimestamp();
         } else {
             return null;
--- a/models/DataObject/ClassDefinition/Data/Datetime.php
+++ b/models/DataObject/ClassDefinition/Data/Datetime.php
@@ -195,7 +195,7 @@ class Datetime extends Model\DataObject\ClassDefinition\Data
      */
     public function getDataForGrid($data, $object = null, $params = [])
     {
-        if ($data) {
+        if ($data && is_object($data)) {
             return $data->getTimestamp();
         } else {
             return null;


brusch:
Under which conditions `$data` is not an object?

ascheider:
@brusch I don't know why. Some objects in the search have  'NOT SUPPORTED' string instead of Carbon object. This breaks the search.
This message is definitely set in:
models/DataObject/ClassDefinition/Data/Localizedfields.php
```
public function getDataForGrid($data, $object = null, $params = [])
    {
        $result = new \stdClass();
        foreach ($this->getFieldDefinitions() as $fd) {
            $key = $fd->getName();
            $context = $params['context'];
            if ($context && $context['containerType'] = 'objectbrick') {
                $result->$key = 'NOT SUPPORTED';
            } else {
                $result->$key = $object->{'get' . ucfirst($fd->getName())}();
            }
            if (method_exists($fd, 'getDataForGrid')) {
                $result->$key = $fd->getDataForGrid($result->$key, $object, $params);
            }
        }

        return $result;
    }
```

ascheider:
@brusch maybe that is the bug?
if ($context && $context['containerType'] = 'objectbrick') {
should container type be checked with ===?
At the moment 'NOT SUPPORTED' is always assigned if we have context.
So my releaseDate became 'NOT SUPPORTED' instead of get data from else.

brusch:
Does f53c29dad26d32e95c092012af981dadaf393ba6 fix the problem for you?

ascheider:
@brusch yes it does. Thank you.

PR #3183, opened 2018-08-03
Implement webservice import for videos in data objects.

LosHawlos:
This patch for #3150 implements the webservice import for videos in data objects.

I tested this in Pimcore 5.4.0 for youtube, vimeo and dailymotion videos.
Asset based videos are tested when used without an idMapper.

I tried to put in code for the id mapper, but that is untested because I dont know how to trigger the use of an id mapper.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Video.php
+++ b/models/DataObject/ClassDefinition/Data/Video.php
@@ -458,8 +458,66 @@ class Video extends Model\DataObject\ClassDefinition\Data
      */
     public function getFromWebserviceImport($value, $relatedObject = null, $params = [], $idMapper = null)
     {
-
-        // @TODO
+        if (is_string( $value)) {
+            if (! strlen( $value)) {
+                return null;
+            }
+            $data = unserialize( $value);
+            if ($data == FALSE) {
+                throw new \Exception('cannot get object video data from web service import - value cannot be decoded');
+            }
+            if (is_array($data)) {
+                if (isset( $data['type']) && isset( $data['data'])) {
+                    if (in_array( $data['type'], array('youtube','vimeo','dailymotion'))) {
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/Video.php
```diff
@@ -458,8 +458,66 @@ public function getForWebserviceExport($object, $params = [])
...
+                        return $this->getDataFromEditmode($data, $relatedObject, $params);
+                    } else if ($data['type'] === 'asset') {
+                        $video = new DataObject\Data\Video();
+                        $video->setType($data['type']);
+                        $video->setTitle($data['title']);
+                        $video->setDescription($data['description']);
+                        if (is_int($id = $data['data'])) {
+                            if ($idMapper) {
+                                // @TODO check implementation of id mapping
+                                throw new \Exception('cannot get object video data from web service import - id mapping is not implemented');
```
weisswurstkanone: could you please provide an explanation for this ?

LosHawlos: I have no idea how to test the code for the idMapper because i dont know how that works.
Its basically copied from an other element: https://github.com/pimcore/pimcore/blob/ddae6114aebe0f9d704dab0bed427f0e20542800/models/DataObject/ClassDefinition/Data/Link.php#L416

I had multiple options what to do and decided to throw an exception instead of silently doing nothing.
You can remove the exception when you think the id mapping code is correct.

weisswurstkanone: Then from looking at the code it should be fine. Can you please update the PR (remove the exception) ?

LosHawlos: Ok, PR is updated.

weisswurstkanone:
thx!

PR #3188, opened 2018-08-03
Support comma as separation symbol for floating-point numbers

jremmurd:
If the amount gets passes with a comma "," as separation symbol for floating-point numbers in the `initPayment` call, the amount will be returned in the same format, which results in an Exception in the `Decimal::create(..)` call in the `handleResponse` function.

Solution: `Decimal::create(..)`  also accepts "," as separation symbol


fashxp:
the question is, why is it passed with `,` in `initPayment`?

PR #3196, opened 2018-08-08
[Document] Link Editable: add prefix & suffix for text content + possibility to not show text

Gerhard-Kanzler:
textSuffix / textPrefix and hideText Options.

now you can add Icons or something else!


brusch:
Changed `hideText` to `noText` since the text isn't hidden but simply not added to the `<link>` tag. 
I've also splitted up the `if` conditions for prefix and suffix, to avoid warnings if just one of both options is used.

PR #3199, opened 2018-08-08
Navigation: Added `hasVisiblePages()` convenience method

zerlpaMMT:
New function inside the navigation container for checking if children have at least one published/visible child pages

### DIFF
--- a/lib/Navigation/Container.php
+++ b/lib/Navigation/Container.php
@@ -283,6 +283,24 @@ class Container implements \RecursiveIterator, \Countable
     }
 
     /**
+     * Returns true if container contains any visible page
+     *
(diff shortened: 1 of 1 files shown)


brusch:
Thanks!

lib/Navigation/Container.php
```diff
@@ -282,6 +282,24 @@ public function hasPages()
         return count($this->_index) > 0;
     }
 
+    /**
+     * Returns true if container contains any visible page
+     *
+     * @return bool whether cintainer has any visible page
```
cadeyrn: typo: "cintainer". ;-)

brusch: Good catch - thanks! :)

PR #3202, opened 2018-08-09
Added new option - Force redirect with trailing slash

### LINKED ISSUE #3201: Allow trailing Slash in URL
# Bug Report

### Expected behavior 
Allow trailing Slash in URL will not cut a traling slash in the url.

### Actual behavior  
With the option "yes" in "Allow trailing Slash" cuts the trailing slash at the end of the url

### Steps to reproduce  
Activate the option and try to serve an url with traling slash. You will be redirected to the url without traling slash.

# Feature Request 
New option - Force redirect with trailing slash

daniel-travix:
Resolves #3201 

## Additional info  
Added the new option "force" to "Allow trailing slash"

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/system.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/system.js
@@ -781,7 +781,8 @@ pimcore.settings.system = Class.create({
                                 value: this.getValue("documents.allowtrailingslash"),
                                 store: [
                                     ["", t("yes")],
-                                    ["no", t("no")]
+                                    ["no", t("no")],
+                                    ["force", t("force")]
                                 ],
                                 mode: "local",
                                 editable: false,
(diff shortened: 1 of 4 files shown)


brusch:
This PR doesn't really resolve #3201, it just adds a new option. 
Imho we should discuss getting rid of this setting anyway, there's no point to keep it.

daniel-travix:
The drop off these lines solves the problem. 
```
// only allow the original key of a document to be the URL (lowercase/uppercase)	
if ($redirectTargetUrl !== rawurldecode($document->getFullPath())) {
    $redirectTargetUrl = $document->getFullPath();
}
```
Now it is not possible to create create SEO freindly urls with trailing slashes.

brusch:
Dropping those lines will cause additional problems! (see comment - uppercase/lowercase). 
URLs with no trailing slash are as good as URLs with trailing slash regarding SEO, it's just important that only one of both is working and the other redirects to the original one. 

#3201 is fixed by the way.

PR #3210, opened 2018-08-10
bugfix at CustomReport date filter

zerlpaMMT:
without 'FROM_UNIXTIME' the SQL command is incorrect


models/Tool/CustomReport/Adapter/Sql.php
```diff
@@ -191,16 +191,20 @@ protected function getBaseQuery($filters, $fields, $ignoreSelectAndGroupBy = fal
...
-
                             $compMapping = [
                                 'lt' => '<',
                                 'gt' => '>',
                                 'eq' => '='
                             ];
+                            
+                            if ($type == 'date') {
+                                if ($operator == 'eq') {
+                                    $condition[] = $db->quoteIdentifier($filter['property']) . ' BETWEEN FROM_UNIXTIME(' . $db->quote($value) . ') AND FROM_UNIXTIME(' . $db->quote($maxValue) . ')';
```
brusch: If `$operator` is `eq`shouldn't be the condition `=` ? and not `BETWEEN` ?

zerlpaMMT: it was already in the file before my change with the from_unixtime.

weisswurstkanone:
thx!

Issue #2515, opened 2018-01-26
"allowedClassId" for ObjectsMetadata is stored within definition_*.php file


jakubderdziak:
# Bug Report
## Please make sure all the following boxes are checked before submitting your issue - thank you!
- [ ] This is not a security issue / vulnerability -> send them to security@pimcore.org instead
- [ ] Behavior is reproducible on http://demo.pimcore.org/admin/ (admin/demo) 
or on a clean install of the latest build (out of your project, to avoid problems with custom code, plugins, ...) 
- [ ] You're not able to fix the problem yourself and send us a pull request instead of an issue. 
- [ ] There's no existing ticket for the same issue
- [ ] Remove unused sections from below and this bullet points

### Expected behavior 
Definition of ObjectMetadata contains class name or class instead of class ID
  
### Actual behavior  
Definition of ObjectsMetadata if definition_*.php file contains field allowedClassId with ID of the class. Since directory DataObjects should be excluded from git and rebuilded with deployment:classes-rebuild this id may be different for different persons

### Steps to reproduce 
1. Create class Target with one input "targetName"
2. Create class Source with one input "sourceName", and one ObjectMetadata field, with Allowed Class set to Target. Select targetName as visible fields.
3. Remove all files from var/classes/DataObjects and clear "classes" database (to simulate sharing code within git).
4. Run php bin/console deployment:classes-rebuild -c
5. Go to admin panel and try to create sourceName object. You will get error

Thamakil:
This really is a severe bug especially in projects with larger teams and I am wondering why nobody seems to have taken notice of this issue yet!
Maybe the implications of what @jakubderdziak wrote are not clear to some so let me give a real world example.

Imagine a pimcore installation with five defined objects having the IDs 1 to 5.
Two developers - Jim and Bones - are working on that system and both have their own dev instance with their own databases installed on their local dev machines.
They share their code through a GIT repository.

Now Jim creates a new class on his machine - we call it JimsClass.
The class is assigned the ID 6 on Jims dev machine as these are auto increment.
Jim also modifies one of the existing classes - Class4 - to contain a ObjectMetadata reference to JimsClass.

At the same time Bones also creates a new class - BonesClass.
BonesClass is also assigned the ID 6 on his local dev machine.

Now both commit and push and pull their definition_*.php files through GIT.
And both execute a ./bin/console deployment:classes-rebuild -c to get each others classes.
[...]

PR #3218, opened 2018-08-14
getById instead of getByName

zerlpaMMT:
getByName is not working with a classId


brusch:
This is actually right (that the name is used, see #2515 ), but it seems something went wrong during an update on your instance. 
Please try to run the following script manually on your Pimcore instance: 
https://github.com/pimcore/pimcore/blob/7af061a4512f2994dac772812d42c531dc5f7df7/update-scripts/283/postupdate.php 

That should fix the problem.

PR #3219, opened 2018-08-14
Fix: When creating folder by path multiple warnings are logged

blankse:
## Additional info  

Bug can be reproduced with the following command:
```
<?php

namespace AppBundle\Command;

use Pimcore\Console\AbstractCommand;
use Pimcore\Model\Asset\Service;

class TestCommand extends AbstractCommand
{
    protected function configure()
    {
        $this->setName('test');
    }

    {
        $folder = Service::createFolderByPath('/Test/Test2');
    }
}
```
Command:
`php bin/console test`

Output:
```
13:44:05 WARNING   [pimcore] asset with path: /Test/Test2 doesn't exist [] []
13:44:07 WARNING   [pimcore] asset with path: /Test doesn't exist [] []
13:44:07 WARNING   [pimcore] asset with path: /Test/Test2 doesn't exist [] []
```

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -760,53 +760,53 @@ class Service extends Model\AbstractModel
             $sanitizedPath = $sanitizedPath . self::getValidKey($part, $itemType) . '/';
         }
 
-        if (!($foundElement = $type::getByPath($sanitizedPath))) {
-            foreach ($parts as $part) {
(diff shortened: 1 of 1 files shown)


blankse:
For review it is easier to select the "Hide whitespace changes" checkbox. I switched the if/else branches to avoid the indents and make the code more readable.

brusch:
Thanks!

PR #3222, opened 2018-08-14
Fix multiple subtypes & classnames for getElementsForTag

### LINKED ISSUE #3204: [Tags] PDO::quote() expects parameter 1 to be string, array given
# Bug Report

### Expected behavior 
When i use `Tag::getElementsForTag($tag, 'object', ['object'], ['classname'] )` i expect to get a result with only the objects i'm specifying in the 4th parameter (classnames).

### Actual behavior  
An exception is thrown: `PDO::quote() expects parameter 1 to be string, array given`
Same thing happens with parameter 3 (subtypes)

### Steps to reproduce  
1. Create a tag in the back-end
2. In a controller get the tag and try to select the elements for that tag with subtypes & classname parameter

### Code example 
```php
$tag = Tag::getTagsForElement('object', $object->getId());
$relatedObjects = Tag::getElementsForTag($tag, 'object', ['object'], ['ClassName'] );
```

dvesh3:
Resolves #3204

### DIFF
--- a/models/Element/Tag/Dao.php
+++ b/models/Element/Tag/Dao.php
@@ -264,11 +264,17 @@ class Dao extends Model\Dao\AbstractDao
         );
 
         if (! empty($subtypes)) {
-            $select->where($map[$type][2] . ' IN (?)', $subtypes);
+            foreach ($subtypes as $subType) {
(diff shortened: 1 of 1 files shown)


fashxp:
thx

PR #3225, opened 2018-08-15
move mime extension configuration to symfony config in order to make it more extendable

dpfaffenbauer:
I currently have the use case where DWG files are recognized as images and Imagick can't deal with them and Pimcore throws an exception cause of that. To quick fix it, I manually add the entry to the `Mime` class on AppKernel boot, but I thought it be better to have that as a proper symfony config.

I also just deprecated the static variable `$extensionMapping'  and created sort of a BC layer for it in case of someone does the same as I do, not sure how you actually deal with such BC breaks now.


bundles/CoreBundle/PimcoreCoreBundle.php
```diff
@@ -57,6 +58,18 @@ public function getContainerExtension()
...
     }
 
+    public function boot()
+    {
+        parent::boot();
+
+        /**
+         * BC: Remove with Pimcore 6.0.0
+         */
+        if ($this->container->hasParameter('pimcore.mime.extensions')) {
```
brusch: This is only for the case that someone tries to access `Mime::$extensionMapping` directly in project/bundle code, right? 
Do you think that's commonly used? Imho it shouldn't, since this was always intended to be an internal variable.

brusch: Imho we should remove this BC layer and make a note in upgrade notes. 
Can you please update your PR? Thank you very much!

dpfaffenbauer: I don't think anyone directly access the variable, but I wasn't sure to be honest. I will just remove the bc layer and the variable and create a Upgrade note.

brusch: Great, thanks!

dpfaffenbauer: done

PR #3234, opened 2018-08-17
Bugfix: Lost version data on index update

asysrieke:
With 5.2.2 the functionality of sorting tree-objects by key (position in tree) was implemented.
https://github.com/pimcore/pimcore/pull/2501

The implementation leads to a problem when moving objects into a folder. 
All siblings of the moved object get their index value updated to keep the index up to date, regardless if the sibling is only saved as an unpublished version. (DataObjectController::updateIndexesOfObjectSiblings).
The modification timestamp of the object is changed and therefore the link to the version file is lost.
This may lead to dataloss when using unpublished versions

The provided fix saves objects only as versions if there is an unpublished version for that object.
Side effect here is that the index of that particular object is updated only inside the version on the file system and therefore the sorting may not work as expected.
But it's less dangerous than to lose version contents.

## Additional info  
Steps to reproduce:
1) Create folder "foo"
2) Create folder "bar" inside "foo"
3) Create object "foobar" inside "foo"
4) Create object "barfoo" inside "bar"
5) Edit object "barfoo" and save as version
6) Move "foobar" into "bar"

The edited changes from step 5 in "barfoo" are lost.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObjectController.php
@@ -1207,10 +1207,16 @@ class DataObjectController extends ElementControllerBase implements EventedContr
             if (method_exists($sibling, 'setOmitMandatoryCheck')) {
                 $sibling->setOmitMandatoryCheck(true);
             }
-            $sibling
-                ->setIndex($index)
-                ->save();
-            $index++;
+            if($sibling instanceof DataObject\Concrete && $latestVersion = $sibling->getLatestVersion()) {
+                $sibling = $latestVersion->loadData();
+                $sibling->setIndex($index)
+                    ->saveVersion();
+            } else {
+                $sibling
+                    ->setIndex($index)
+                    ->save();
+                $index++;
+            }
         }
     }
 


brusch:
Thanks for your PR! 
As far as I can see, this PR leads to another problem, because when just updating the index in the latest version of the affected data object, the change doesn't go to the database (`objects` table) and therefore it doesn't reflect in the tree, eg. when reloading the parent node. 

I guess what we have to do additionally is updating the index directly in the database table without calling `save()` on the object itself.

brusch:
We're doing the same with documents, see: 
https://github.com/pimcore/pimcore/blob/c62924ca5c3d0c482a71097cb0231aff143208d6/models/Document.php#L617

brusch:
584e38a64c59cacabe5d50849b97d09a15286cc7 should solve all issues regarding versions and index update for objects and documents

brusch:
@asysrieke thanks for coming back on this ... you're absolutely right. 
With 373f41401930a22e6edbdde764d09adc0f0e0a54 this should be fixed as well. 
Thanks again!

asysrieke:
Thanks @brusch, now it works.
While implementing and testing the bugfix I recognized a similar problem elsewhere in the code. When child-paths are updated, e. g. while moving an entire object tree, the modification date is updated too even though the paths are saved directly to the DB without using the save function. 
So unpublished version contents get lost as well.

See: AbstractObject\Dao::updateChildsPaths() 

```php
public function updateChildsPaths($oldPath)
    {
     ...
            //update object child paths
            $this->db->query('update objects set o_path = replace(o_path,' . $this->db->quote($oldPath . '/') . ',' . $this->db->quote($this->model->getRealFullPath() . '/') . "), o_modificationDate = '" . time() . "', o_userModification = '" . $userId . "' where o_path like " . $this->db->quote($oldPath . '/%') . ';');

      ...
            return $objects;
        }
    }
```
My fix would be to simply not update the modificationDate at this point. If desired I would commit another pull request for this topic.

A long term improvement could be to make the modificationDate independent from the version files. This would be a very nice feature ;)

brusch:
@asysrieke thanks again, fixed as well: 00a62aedbee6909e84503df185a9cb9f69fdb080

PR #3240, opened 2018-08-21
Use admin translations for rendering meta column value

BlackbitNeueMedien:
When you have a multihref with metadata field which has a select or multiselect metadata column its values should be translated.

Reproduce:
1. Create multihref with metadata field in a data object class
1. Add meta column "test" of type "select" and add some options
1. Add admin translations with key=option key
1. go to an object and set value -> admin translations are not used

With this PR admin translations get used to display selected option.


brusch:
Thanks!

brusch:
Added the same for data-type Object with Metadata.

PR #3247, opened 2018-08-23
[Bugfix] Use Pimcore constant instead of hard-coded value

sebwebdev:
## Steps to reproduce
* Move the default AppBundle to an own namespace (e.g. AcmeAppBundle) by setting PIMCORE_SYMFONY_DEFAULT_BUNDLE
* Create a document with a default action
* An error will be thrown in the "Edit" tab since Symfony can not find the template


brusch:
Thanks!

PR #3261, opened 2018-08-27
Update DataObjectHelperController.php

mbolka:
The line 649 will be overwritten by line 652, so there is no sense to keep it


brusch:
Thanks!

PR #3293, opened 2018-08-31
[XLIFF export/import] Refactor to make it extendable and reusable

markus-moser:
Currently the XLIFF export/import feature is handled directly in the controller. This makes it impossible to reuse and extend the logic. 

This pull request separates the export/import process into several components:

- **Export data extractor**: extracts the localized field data from the pimcore elements into a neutral representation (AttributeSet). 

- **Translation exporter**: takes the AttributeSet and creates the desired target format/export file (XLIFF by default)

- **Import data extractor**: takes an import file of a given format (XLIFF), parses the data and returns again an AttributeSet representation of the data

- **Importer**: takes the AttributeSet and actually import/save it to the target Pimcore elements

Additionally to this refactoring it adds a configuration option to export only certain data object attributes into the Pimcore configuration tree.

#### Example:
```yaml
pimcore:
    translations:
        data_object:
           translation_extractor:
              attributes:
                  Product:
                     - description
```

Another improvement is that now localized fields in bricks are supported by the export/import feature.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/TranslationController.php
+++ b/bundles/AdminBundle/Controller/Admin/TranslationController.php
@@ -23,6 +23,11 @@ use Pimcore\Model\Document;
 use Pimcore\Model\Element;
 use Pimcore\Model\Translation;
 use Pimcore\Tool;
+use Pimcore\Translation\ExportService\Exporter\ExporterInterface;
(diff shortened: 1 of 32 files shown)


brusch:
Thanks a lot!

PR #3316, opened 2018-09-06
Missing word "not"

IgorBenko:
Added word "not" in the sentence:
That is because the framework does NOT limit you on specific classes or class structures. Literally every class can act as a product or product framework does not limit you on specific classes or class structures.


brusch:
Thanks ;)

PR #3319, opened 2018-09-07
[DOCS] Update urls in documentation pointing to the wrong location 

jorisros:
Update the location of the urls, they where pointing to the old location.


brusch:
Thanks!

PR #3320, opened 2018-09-07
Application Logger: Properly escape log message

podarcis:
Messages with markup (e.g. <?xml version='1.0' encoding='UTF-8' standalone='no' ?>)
break the entire listing and other prepared content can lead to XSS.


weisswurstkanone:
thx

PR #3337, opened 2018-09-12
Fixed variant tab actions based on permissions

### LINKED ISSUE #3335: Variants Rename Permission Issue
# Bug Report
Under user and roles, for a user, I have only given list and view permissions on a particular folder, in Data Object section. 
Still I am able to rename variants that are present in that folder. The rename option is not coming on the  context menu but I can rename it from Variants tab. 

### Expected behaviour 
It should not be renamed from the Variants tab also. 

### Actual behaviour 
Rename option is coming under the Variants tab. 

variantsrenameissue

dvesh3:
resolves #3335


brusch:
Thanks!

PR #3341, opened 2018-09-12
Supply sample configuration for WYSIWYG/CKEditor for filtering out HTML tags in Paste with Word

alexleeelkins:
Provides sample configuration for WYSIWYG/CKEditor to customize what is filtered out when using Paste with Word


alexleeelkins:
Hello, I am not sure what you mean by "apply it to a data type or to an editable". What specifically in the CKeditor/WYSIWYG could be applied to a data type?

Additionally, I added an alternative to modifying directly the stock config.js by creating a plugin that directly modifies the CKeditor config, as the existing documentation for that does not appear to work with pimcore4.

PR #3345, opened 2018-09-13
Custom Reports: Fixed headers exporting CSV ","

cleggypdc:
When you select to include the headers in a custom report, they are exported as CSV (,), while the rest of the report is semi-colon separated.


weisswurstkanone:
thx!

PR #3348, opened 2018-09-14
Fixed localized fields filtering on custom layout change.

ascheider:
Fixed localized fields filtering on layout change via event: https://pimcore.com/docs/5.x/Development_Documentation/Best_Practice/Showing_Custom_Layouts_based_on_Object_Data.html

## Additional info  
How to reproduce:
1. Create custom layout
2. Restrict localized fields via workspace settings
3. Define event like here described: https://pimcore.com/docs/5.x/Development_Documentation/Best_Practice/Showing_Custom_Layouts_based_on_Object_Data.html
4. Change layout programmatically
5. On the first open of object you will see all possible languages in localized fields
6. After reload of object they are restricted to your workspace configuration

The bug appears because there is no localized fields filtering, if layout is changed via event.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObjectController.php
@@ -464,9 +464,6 @@ class DataObjectController extends ElementControllerBase implements EventedContr
                 $objectData['currentLayoutId'] = $currentLayoutId;
             }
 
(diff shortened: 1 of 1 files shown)


ascheider:
@brusch you are right. PR is updated.

brusch:
Thanks!

PR #3350, opened 2018-09-14
[ecommerce][index service] Elasticsearch delete from index bugfix in combination with virtual product ids

markus-moser:
It's not a good idea to delete items from the elastic search based on the osParentId if they are added based on the virtual product ID. Depending on the use cases this could be a big difference and will result in undeleted items in the elasticsearch index.


fashxp:
thx

PR #3371, opened 2018-09-24
Do not disable already disabled redirects

AndrewAteles:
Resolves #
Redirects cache being blown for already delt with entries.
## Additional info  

When running the maintenance job there query that is being used does not consider if the redirect is already disabled. Disabled redirects are disabled again causing the cache to be invalidated and affecting the performance for the web site.

To reproduce:

- Create a redirect with expiration date set to yesterday.
- Open a page on the web site.
- Run the Pimcore maintenance job.
  => Redirect is disabled
  => Cache is gone
- Reload the page (make sure it's not coming from cache, add a query parameter or something).
  => Redirects are saved to cache again.
- Run maintenance job again
 => Redirect is disabled again even though its already disabled
 => BAD Cache is gone
- Reload the page again (make sure it's not coming from cache, add a query parameter or something).
 => BAD: redirects are saved to cache once again.

In our case there are 26000 redirects which greately affects performance.

### DIFF
--- a/models/Redirect.php
+++ b/models/Redirect.php
@@ -316,7 +316,7 @@ class Redirect extends AbstractModel
     public static function maintenanceCleanUp()
     {
         $list = new Redirect\Listing();
-        $list->setCondition('expiry < ' . time() . " AND expiry IS NOT NULL AND expiry != ''");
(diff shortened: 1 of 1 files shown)


brusch:
Makes sense - good catch! 
Thank you very much!

PR #3372, opened 2018-09-25
QuantityValue webservice bugfix

markus-moser:
A check is needed whether the unit actually exists otherwise a problem will result if units get deleted as the unit ID might be still in the database.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/QuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/QuantityValue.php
@@ -426,7 +426,7 @@ class QuantityValue extends Model\DataObject\ClassDefinition\Data
         if ($data instanceof \Pimcore\Model\DataObject\Data\QuantityValue) {
             return [
                 'value' => $data->getValue(),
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Many thanks, maybe we should also think about a proactive cleanup task as part of the maintenance?

markus-moser:
yes, the cleanup task might be a good idea

dpfaffenbauer:
or just use foreign key constraints?

weisswurstkanone:
technically possible, was considering this but wasn't sure if this really performs ...

dpfaffenbauer:
`SET NULL` definitely will work

https://dev.mysql.com/doc/refman/5.6/en/create-table-foreign-keys.html

weisswurstkanone:
maybe open a seperate "improvement" ticket ?

dpfaffenbauer:
of course, either for the foreign key or the cleanup task. Can you?

weisswurstkanone:
done, see #3376

PR #3373, opened 2018-09-25
Generate valid keys with length(key) <= length(key) of database tables

BlackbitNeueMedien:


brusch:
Thanks for your PR. 
Unfortunately it isn't entirely right, since we're supporting utf8 this could lead to other problems, using `mb_substr()` should do it the right way :)

brusch:
I've changed that :)

PR #3374, opened 2018-09-25
[DOCS] Additional config info for Matomo iframe

podarcis:
The default configuration of Matomo prevents the inclusion via IFrame.
See: https://matomo.org/faq/troubleshooting/faq_147/


brusch:
Thanks!

PR #3378, opened 2018-09-25
Updated best practice for custom layouts

cleggypdc:
I found an edge case using the example provided where if you have a custom layout containing an additional layout item, such as a text box, and then you attach a custom renderer to that text box, this example passes null through as the object (and context).

Adding the $object parameter in here catches this edge case :)


brusch:
Thanks a lot!

PR #3382, opened 2018-09-26
[Workflows] data_object_multiple_state marking store fix

markus-moser:
The marking store should not save the subject (like the other marking stores...). Additionally the code was simplified.


brusch:
Can you resolve the conflict please? :) 
Thanks!

PR #3403, opened 2018-10-01
let Pimcore configs silently fail when invalid (eg empty or invalid PHP)

dpfaffenbauer:
When you have a invalid web2print or report config, pimcore fails. This PR makes Pimcore fail silently. 

BTW: Configs should be in symfony, but this is just a temporary fix to make deployments with shared empty files easier.


weisswurstkanone:
thx

PR #3404, opened 2018-10-01
Prevent warning array_keys() expects parameter 1 to be array, null given

faridmovsumov:
Prevent Warning: array_keys() expects parameter 1 to be array, null given


faridmovsumov:
Basically it causing a warning when you call getAvailableActions if there is no actions for current status.

```
$workflowWanager = new Manager($masterConcept, $this->user);
$actions = $workflowWanager->getAvailableActions();
```

fashxp:
thx

PR #3410, opened 2018-10-02
[Workflows] add getTransitionByName helper method to workflow manager

markus-moser:
A helper method to get a workflow transition by name.


fashxp:
thx

PR #3414, opened 2018-10-03
Disable Inline Upload for Document Tags Multihref / Image

daves-weblab:

Gives the possibility to disable the inline upload for the document tags multihref and image via the option `disableInlineUpload`. Docs are updated accordingly.

### Resolves issue with document tag multihref

Currently the inline upload for assets is enable on the document tag multihref even if assets are not allowed and the multihref is restricted to e.g. objects only. The problem lies after actually uploading an asset inline, it is placed in the multihref, which should not be possible. This PR fixes this issue in accordance to the previously mentioned option of disabling the inline upload via the options.


brusch:
Shouldn't the inline upload just be not available if assets are not supported?

daves-weblab:
It does both, first of all the upload is disabled if assets are not supported automatically. 

The option "disableInlineUpload" might be useful for scenarios, when certain users might have only access to specified asset folders which might not be coverable by the option "uploadPath" alone, which would raise exceptions on upload. Thus it could be handy to just disable the inline upload altogether via an option and use the asset tree for asset creation.

weisswurstkanone:
@brusch , pimcore.document.tag line 197 should do the trick

brusch:
The inline drag & drop upload on image editables isn't yet considered, so this still works even if  `disableInlineUpload = true`. 
Also the icon should change if dnd upload is not available. 

Thanks!

PR #3415, opened 2018-10-04
display version publish errors to the user

dpfaffenbauer:
Otherwise the object just reloads and you have no clue that it actually failed.


weisswurstkanone:
thx!

PR #3424, opened 2018-10-05
QuantityValue - add option to configure the width of the unit field

ultramegatom:
QuantityValue - add option to configure the width of the unit field


weisswurstkanone:
thx!

PR #3427, opened 2018-10-05
Update DefaultFactFinder.php

ckogler-elements:
Remove "?" for query building because $this->getSearchUrl() returns a "?" at the end. Would result in "??"


fashxp:
thx

PR #3428, opened 2018-10-06
Add option to disable useIds only 

ckogler-elements:
Add option to disable useIds only parameter (required in special situation where you need e.g. the variant ids in the response)


ckogler-elements:
Would make sense in my opinion

fashxp:
thx

PR #3430, opened 2018-10-07
[Workflow] Don't overwrite symfony registered services

dpfaffenbauer:
this causes issues with other services that already extend it or use it -> for example if you manually define a symfony workflow -> registering that again will cause to lose all of its configuration and therefore fail. even if it is configured properly.

this is quite urgent and has to be merged before 5.5, since this breaks default symfony behaviour in workflows.

### DIFF
--- a/bundles/CoreBundle/DependencyInjection/Compiler/WorkflowPass.php
+++ b/bundles/CoreBundle/DependencyInjection/Compiler/WorkflowPass.php
@@ -38,7 +38,7 @@ class WorkflowPass implements CompilerPassInterface
     {
         $loader = new YamlFileLoader(
             $container,
-            new FileLocator(__DIR__.'..//../Resources/config')
+            new FileLocator(__DIR__.'/../../Resources/config')
         );
 
         $workflowManagerDefinition = $container->getDefinition(Manager::class);
--- a/bundles/CoreBundle/DependencyInjection/Compiler/WorkflowPass.php
+++ b/bundles/CoreBundle/DependencyInjection/Compiler/WorkflowPass.php
@@ -1,0 +1,253 @@
+<?php
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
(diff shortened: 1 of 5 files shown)


dpfaffenbauer:
PS: had to change the PR to move the configuration to a CompilerPass, in the Extension you don't know what other services from other bundles have been loaded -> the CompilerPass knows that, so therefore the pimcore workflow config is added as parameter and picked up again in the CompilerPass. 

I guess there are not unit-tests for Pimcore Workflow right? Are there any other tests somewhere? In a demo or so?

dpfaffenbauer:
ok, just tested with the advanced demo: works

markus-moser:
Hmm, I combined a Symfony workflow and a Pimcore workflow in one project and it worked together. So it's strange why it didn't work in your case whereas it worked in mine? I used the workflow from this bundle together with a Pimcore workflow defined in the AppBundle:
https://github.com/pimcore/translations-provider-interfaces/blob/master/src/Resources/config/workflow.yml

So is it a matter of bundle/configuration priority?

dpfaffenbauer:
Yes it is. Maybe the reason could also be that CoreShop uses the prependExtensionConfig function that causes this to fail.

dpfaffenbauer:
but anyhow: `debug:config FrameworkBundle` shows my workflows but the workflow registry is empty... so clearly the service gets overwritten and removes my workflows. So I guess thats more of a Priority Problem.

markus-moser:
Yes, I think that's the reason. I was not very happy that I needed to add the Symfony service configurations again but without that it only worked when at least one Symfony core workflow was configured as otherwise the XML config file with the Symfony workflow services definitions were not loaded:
https://github.com/symfony/framework-bundle/blob/master/DependencyInjection/FrameworkExtension.php#L502

The solution with the CompilerPass should be a good one. I'll try it out if everything works correctly in my case...

dpfaffenbauer:
@markus-moser had you had the chance to test it?

markus-moser:
No - I will test it within the next 2 days. Pimcore 5.5 will not be released before we checked this pull request ;-)

markus-moser:
In my opinion everything is fine with the pull request. I tested it and everything works for me too.

@fashxp

PR #3431, opened 2018-10-07
Introduce branch alias

dpfaffenbauer:
introduce branch alias to easier install dev versions of pimcore


brusch:
Makes sense, thanks!

PR #3441, opened 2018-10-09
An exception is thrown for users without permission in the GDPR Data Extractor

mcuenez:
Opening the **GDPR Data Extractor** and searching for *any* value will trigger a popup window. This window shows an exception for users without the **emails** permission.

Added a check for user permissions before rendering the *Sent Emails* Panel in the GDPR Data Extractor.
If the permission is not given the user will simply not see the *Sent Emails* tab.

## How to reproduce

1. Configure the GDPR Data Extractor as documented [here](https://pimcore.com/docs/5.x/Development_Documentation/Tools_and_Features/GDPR_Data_Extractor.html).
2. Create a new user without *emails* permission.
3. Log in with the newly created user.
4. Use the GDPR Data Extractor to search for anything.
5. When the *Sent Emails* tab is to be loaded an exception will be thrown in [EmailController.php](https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Controller/Admin/EmailController.php) (which will be shown in a popup window).

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/gdpr/dataproviders/sentMail.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/gdpr/dataproviders/sentMail.js
@@ -20,12 +20,12 @@ pimcore.settings.gdpr.dataproviders.sentMail = Class.create({
     initialize: function (searchParams) {
         this.searchParams = searchParams;
         this.getPanel();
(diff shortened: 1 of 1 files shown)


fashxp:
needs some discussion first ... see #3443

brusch:
Unfortunately this is a bigger topic than just emails - so I close that known in favor of #3443 
Thanks anyway for your PR that - at least - brought up this topic ;)

PR #3442, opened 2018-10-09
[fix] Get class definition in ObjectsMetadata

prugala:
## bug #

In the older version in class definition for field `objectsMetadata` in attribute `allowedClassId` was saved class id instead of class name.

Tested on pimcore 5.23.


brusch:
Maybe it's better to use `is_numeric()` in this case? Just in case the ID ist stored as string?

prugala:
True, I changed it.

PR #3453, opened 2018-10-10
Disable cache write lock after clearTags

j-jonas:

After saving an asset it is not possible to cache something.

  

Resolves #


brusch:
The current behavior of not releasing the lock is intentional. After a single item get's removed from the cache, no more items should be written to the cache to avoid race conditions.

PR #3454, opened 2018-10-11
Remove legacy view script rendering

### LINKED ISSUE #2081: [ECOMMERCE] check for legacy script paths makes debugging very hard
The ecommerce bundle currently checks the legacy views path as fallback. This makes debugging sometimes very hard if there are typos etc. in the template file path. 

Suggested solution: we should do this fallback check only if the legacy bundle is enabled.

```php
// Pimcore\Bundle\EcommerceFrameworkBundle\FilterService\FilterType\AbstractFilterType
try {
            return $this->templatingEngine->render($template, $parameters);
        } catch (\Exception $e) {

            //legacy fallback for view rendering
            $prefix = PIMCORE_PROJECT_ROOT . '/legacy/website/views/scripts';

            return $this->templatingEngine->render($prefix . $template, $parameters);
        }
```

ckogler-elements:
(did't worked properly + not useful any more)

Resolves #2081

### DIFF
--- a/bundles/EcommerceFrameworkBundle/FilterService/FilterType/AbstractFilterType.php
+++ b/bundles/EcommerceFrameworkBundle/FilterService/FilterType/AbstractFilterType.php
@@ -158,14 +158,6 @@ abstract class AbstractFilterType
      */
     protected function render($template, array $parameters = [])
     {
-        try {
-            return $this->templatingEngine->render($template, $parameters);
-        } catch (\Exception $e) {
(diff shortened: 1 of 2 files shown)


markus-moser:
yes this makes debugging very hard - see #2081

ultramegatom:
+ on this - caused me a lot of headaches already

brusch:
Makes sense, thanks a lot!

PR #3455, opened 2018-10-11
[WebDAV fix] Mime detection: check for 0Byte files

myrdd:
Some WebDAV clients create 0B files (zero byte files), e.g. as a mechanism for locking. 
For details, see http://sabre.io/dav/0bytes/.

This PR ensures that 0B files will always have type `application/octet-stream`.

Prior to this fix, when a file was detected e.g. as an image, it would become an `Image`, and tasks like dimension calculation (width/height) or thumbnail creation would fail.

## Additional info  

If I push a 0B file to `dir/img.jpg`, I'll get the following error in the log:

```
[2018-10-11 10:16:41] pimcore.ERROR: ImagickException: insufficient image data in file `/var/www/pimcore/var/tmp/asset-temporary/asset_65_f1683f4c46a775391018314af9e99be4__img.jpg' @ error/jpeg.c/ReadJPEGImage/1039 in /var/www/pimcore/vendor/pimcore/pimcore/lib/Image/Adapter/Imagick.php:117
Stack trace:
#0 /var/www/pimcore/vendor/pimcore/pimcore/lib/Image/Adapter/Imagick.php(117): Imagick->readimage('/var/www/pimcor...')
#1 /var/www/pimcore/vendor/pimcore/pimcore/models/Asset/Image.php(409): Pimcore\Image\Adapter\Imagick->load('/var/www/pimcor...', Array)
#2 /var/www/pimcore/vendor/pimcore/pimcore/models/Asset/Image.php(50): Pimcore\Model\Asset\Image->getDimensions('/var/www/pimcor...', true)
#3 /var/www/pimcore/vendor/pimcore/pimcore/models/Asset.php(501): Pimcore\Model\Asset\Image->update(Array)
#4 /var/www/pimcore/vendor/pimcore/pimcore/models/Asset.php(355): Pimcore\Model\Asset->save()
#5 /var/www/pimcore/vendor/pimcore/pimcore/models/Asset/WebDAV/Folder.php(130): Pimcore\Model\Asset::create(8, Array)
#6 /var/www/pimcore/vendor/sabre/dav/lib/DAV/Server.php(1095): Pimcore\Model\Asset\WebDAV\Folder->createFile('img.jpg', Resource id #29)
#7 /var/www/pimcore/vendor/sabre/dav/lib/DAV/Locks/Plugin.php(247): Sabre\DAV\Server->createFile('dir/img.jpg', Resource id #29)
[...]

PR #3458, opened 2018-10-11
[minor] fixup Image::update()

myrdd:
This is a quick fix for an issue I discovered along with #3455.

If image dimension calculation fails (in #3455 it's due to a 0Byte file), `imageDimensionsCalculated` should be set `false`. (this is a fixup of 37caf9c.)

## Additional info  
I removed a comment. It's obsolete. It was introduced in 37caf9c and referred to the `$this->getFileSize()` check, which was removed (shortly after) in 50dec26.

### DIFF
--- a/models/Asset/Image.php
+++ b/models/Asset/Image.php
@@ -40,8 +40,12 @@ class Image extends Model\Asset
      */
     protected function update($params = [])
     {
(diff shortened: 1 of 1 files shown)


brusch:
Thanks!

PR #3461, opened 2018-10-12
Remove site if document is registered as one

### LINKED ISSUE #3253: Deleting a document which is a site doesn't delete the site entry
***Steps to reproduce***
 - Create a Pimcore Doc
 - Make it a site
 - Delete the Pimcore doc
 - Site is still in the DB.

scrummer:
Resolves #3253 
When a document is registered as site and you remove it from the document tree, the site entry isn't deleted from the DB. This leads into a crash of the webpage.

This PR fixes this problem and checks whether the document is registered as site when you delete it.

### DIFF
--- a/models/Document.php
+++ b/models/Document.php
@@ -779,6 +779,11 @@ class Document extends Element\AbstractElement
             $service = new Document\Service;
             $service->removeTranslation($this);
 
+            // if document is registered as site remove it
+            if (($site = Site::getByRootId($this->getId())) instanceof Site) {
+                $site->delete();
+            }
+
             $this->getDao()->delete();
 
             // clear cache
--- a/models/Document.php
+++ b/models/Document.php
@@ -780,8 +780,12 @@ class Document extends Element\AbstractElement
             $service->removeTranslation($this);
 
             // if document is registered as site remove it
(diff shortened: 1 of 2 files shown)


models/Site.php
```diff
@@ -178,6 +178,16 @@ public static function create($data)
         return $site;
     }
 
+    /**
+     * Remove the site
+     *
+     * @return void
+     */
+    public function delete()
```
brusch: Not necessary, delete is delegated automatically to the DAO

models/Document.php
```diff
@@ -779,6 +779,15 @@ public function delete()
             $service = new Document\Service;
             $service->removeTranslation($this);
 
+            // if document is registered as site remove it
+            try {
+                if (($site = Site::getByRootId($this->getId())) instanceof Site) {
```
brusch: Sites are only possible on pages, so the right place to delete the site would be here: 
https://github.com/pimcore/pimcore/blob/cef6aa4253fa6cde359fe8644e3240aa03ec1d48/models/Document/Page.php#L83

scrummer: Ah yes, I didn't think that far... I will adapt it later

brusch:
fixed already, see 72c7b7ab4f326c0013445523932113d8a2537b0e

PR #3464, opened 2018-10-12
[Workflows] email notification roles fix

markus-moser:
The notifications worked for users only. This fix makes it working for roles too.


fashxp:
thx

PR #3468, opened 2018-10-13
Add objects with meta data date column support

mbolka:
## Add date type info objects metadata and multihref with metadata.
This pull requests adds new type into supported types of meta columns.
image
image
image

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/multihrefMetadata.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/multihrefMetadata.js
@@ -424,7 +424,7 @@ pimcore.object.classes.data.multihrefMetadata = Class.create(pimcore.object.clas
             var types = {
                 number: t("objectsMetadata_type_number"),
                 text: t("objectsMetadata_type_text"),
-                date: t("objectsMetadata_type_date"),
+                date: t("date"),
                 select: t("objectsMetadata_type_select"),
                 bool: t("objectsMetadata_type_bool"),
                 columnbool: t("objectsMetadata_type_columnbool"),
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/multihrefMetadata.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/multihrefMetadata.js
@@ -424,6 +424,7 @@ pimcore.object.classes.data.multihrefMetadata = Class.create(pimcore.object.clas
             var types = {
                 number: t("objectsMetadata_type_number"),
                 text: t("objectsMetadata_type_text"),
+                date: t("objectsMetadata_type_date"),
                 select: t("objectsMetadata_type_select"),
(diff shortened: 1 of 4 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/multihrefMetadata.js
```diff
@@ -424,6 +424,7 @@ pimcore.object.classes.data.multihrefMetadata = Class.create(pimcore.object.clas
             var types = {
                 number: t("objectsMetadata_type_number"),
                 text: t("objectsMetadata_type_text"),
+                date: t("objectsMetadata_type_date"),
```
brusch: The translation key `objectsMetadata_type_date` doesn't exist. 
Please use the `date` translation key instead, which does exist already. 
The other translations keys which are prefixed with `objectsMetadata_type_` are kind of legacy and are just aliased here: https://github.com/pimcore/pimcore/blob/c62924ca5c3d0c482a71097cb0231aff143208d6/bundles/CoreBundle/Resources/translations/aliases.json#L25 so there's nothing wrong by just usinf `date` directly´.

bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/objectsMetadata.js
```diff
@@ -248,6 +248,7 @@ pimcore.object.classes.data.objectsMetadata = Class.create(pimcore.object.classe
             var types = {
                 number: t("objectsMetadata_type_number"),
                 text: t("objectsMetadata_type_text"),
+                date: t("objectsMetadata_type_date"),
```
brusch: Here the same please ;)

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/multihrefMetadata.js
```diff
@@ -114,6 +114,10 @@ pimcore.object.tags.multihrefMetadata = Class.create(pimcore.object.tags.abstrac
                 cellEditor = function() {
                     return new Ext.form.TextField({});
                 };
+            } else if (this.fieldConfig.columns[i].type == "date" && !readOnly) {
+                cellEditor = function() {
+                    return new Ext.form.DateField({});
```
brusch: Imho we should configure a `submitFormat` here, to ensure the format is always the same.

mbolka:
@brusch All fixed. I have some doubts about it - values that are read from DB are displayed properly but are not assigned into Ext.form.data (date picked on first calendar open is a current date).
I have no idea how to fix it.
image
image

brusch:
@mbolka there's still something wrong, when I add new relation items, it looks like that: 

image

brusch:
@mbolka you should also try to solve the preselected date issue from above, otherwise I can't merge it. 
Thanks!

brusch:
Closed due to inactivity. 
We can reopen this PR any time if needed and when the conflicts and missing issues are resolved.

PR #3479, opened 2018-10-16
Atomic writes for generated files

BlackbitNeueMedien:
Pimcore generates PHP files for classes, field collection, object bricks etc. Therefore it generates the code and writes it via file_put_contents() to the file system. The problem is that file_put_contents is not atomic. When another request accesses the file to be written, it can happen that the file does not exist for a short time or is incomplete (access then results in a syntax error most of the times). This is why these files should be written atomically. 
For more information about the problem with file_set_contents (and other PHP write operations) see https://github.com/nette/safe-stream


brusch:
I guess that would break the support for custom stream wrappers, right? 
Eg. when using AWS S3 stream wrapper, or similar libraries.

BlackbitNeueMedien:
You are right. I changed to code to only use nette/safe-stream for generated code files.

BlackbitNeueMedien:
Oh, I actually did not want to include it to assets. 
What do you think to use it in class, brick and field collection definitions? Can you imagine cases where these are not saved locally?

But of course I would agree adding it to best practices.

brusch:
@BlackbitNeueMedien yes, that could be also the case on read-only environments (containers, ...) where it should be still possible to change definitions on the production environment. 

So i guess we should go for the best-practice guide for now - would be nice if you could create a new PR for that. We should keep this existing PR untouched for reference.

PR #3498, opened 2018-10-17
fix sql error in element_workflow_state

BlackIkeEagle:
And also make the schema end up like one would have migrated.

When you migrate from 5.4.x to 5.5.0 you get additional columns which are
described in doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/01_Within_V5/05_Workflow_Management.md
in element_workflow_state. The install.sql script does not properly
reflect the requirement of workflow which should end up as NOT NULL.

Resolves #

## Additional info  

This makes sure the upgraded schema and the clean install schema are more alike


brusch:
Thanks!

Issue #3499, opened 2018-10-17
Voucher: SingleToken generation always displays error


ultramegatom:
If you click 'Assign Configuration' for a 'Single Token Voucher' an error message is always displayed even if it was successful:

![image](https://user-images.githubusercontent.com/7206334/47091648-8a41ef80-d225-11e8-926a-ccb36545eb3f.png)

PR #3500, opened 2018-10-17
fix voucher generation always displaying an error message

ultramegatom:
fixes: https://github.com/pimcore/pimcore/issues/3499

### DIFF
--- a/bundles/EcommerceFrameworkBundle/VoucherService/TokenManager/Pattern.php
+++ b/bundles/EcommerceFrameworkBundle/VoucherService/TokenManager/Pattern.php
@@ -219,13 +219,17 @@ class Pattern extends AbstractTokenManager implements IExportableTokenManager
      * Generates Codes and an according Insert Query, if the MAX_PACKAGE_SIZE
      * may be reached several queries are generated.
      *
-     * @return array|bool
+     * @return bool
      */
     public function insertOrUpdateVoucherSeries()
     {
         $db = \Pimcore\Db::get();
         try {
             $codeSets = $this->generateCodes();
+
+            if($codeSets === false) {
+                return false;
+            }
 
             if (is_array($codeSets)) {
                 foreach ($codeSets as $query) {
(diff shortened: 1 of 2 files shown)


bundles/EcommerceFrameworkBundle/VoucherService/TokenManager/Single.php
```diff
@@ -134,9 +137,13 @@ public function insertOrUpdateVoucherSeries()
                     ON DUPLICATE KEY UPDATE token = ?, length = ?';
 
             $db->query($query, [trim($this->configuration->getToken()), $this->getFinalTokenLength(), $this->getSeriesId(), trim($this->configuration->getToken()), $this->getFinalTokenLength()]);
+
+            return true;
         } catch (\Exception $e) {
-            return ['error' => 'Something went wrong.'];
+            return false;
```
fashxp: I think we at least should log the exception...

bundles/EcommerceFrameworkBundle/VoucherService/TokenManager/Pattern.php
```diff
@@ -235,12 +239,12 @@ public function insertOrUpdateVoucherSeries()
                 $db->query($this->buildInsertQuery($codeSets));
             }
 
-            return $codeSets;
+            return true;
         } catch (\Exception $e) {
-            //            var_dump($e);
-            //            \Pimcore\Log\Simple::log('VoucherSystem', $e);
             return false;
```
fashxp: I think we at least should log the exception...

bundles/EcommerceFrameworkBundle/VoucherService/TokenManager/Single.php
ultramegatom: Well i agree with you - however there is absolutely nothing logged within the whole VoucherService folder. There is a bunch commented log messages here and there and a TODO message: "// TODO - Log Errors".

IMHO the whole 'VoucherService' could use a little polish. I think we should cover that in a seperate issue.

What do you think @fashxp ?

PR #3507, opened 2018-10-18
System thumbnail generation

ckogler-elements:
Only generate system thumbnails if "--system" option is provided an no thumbnails are passed to generate

e.g: php www/bin/console pimcore:thumbnails:image --system


brusch:
Thanks!

PR #3511, opened 2018-10-19
Small update to Cart Manager documentation

NiklasBr:
Removes incorrect statement from documentation. It is unfortunately not obvious because none on our team understand it


brusch:
Thanks!

PR #3519, opened 2018-10-19
boolean select - batch assignment fix

markus-moser:
Boolean select batch assignment does not work (reproducable in the advanced demo) - this fix makes it working.


brusch:
Thanks a lot!

brusch:
Thanks a lot!

PR #3530, opened 2018-10-24
Initial and latest migrations missing as step

MercuryKojo:
#Improvement for documentation
Initial and latest migrations missing as step


brusch:
Thank you very much!

Yivan:
Hello, we upgraded some instance  to 5.4 and above before this fix was released.
Do we must run manually this fix (step 8 of https://pimcore.com/docs/5.x/Development_Documentation/Installation_and_Upgrade/Upgrade_Notes/Within_V5/Update_from_5.x_to_5.4_or_above.html) ?
Even if Pimcore instance are on the last stable (actually v5.5.3) ?
Thanks.

brusch:
@Yivan yes, you definitely should ... Pimcore only executed migrations which haven't been executed before, so you're safe to run the commands as many times as you like ;)

Yivan:
@brusch
Thank you! We will run it on the Pimcore instance we updated so.

As some instances have been upgraded several times (for instance from 5.3.x -> 5.4.1-> 5.4.3 -> latest), do we need to execute manually other migration scripts (which may have been skipped because we don't run the two migration commands), or running those is just enough so all will be ok ?:
```
bin/console pimcore:migrations:execute -s pimcore_core 20180724144005
bin/console pimcore:migrations:migrate -s pimcore_core
```

brusch:
Yes, running those 2 commands on all instances will fix possible missing migrations.

Yivan:
Ok great! Thanks, have a nice day.

PR #3532, opened 2018-10-24
Restore layout state after reloading object

BlackbitNeueMedien:
This PR adds the functionanlity that after clicking "Reload" on an object the layout state gets restored. This means when you have accordions, tab panels etc. and you have edited a field on tab X then this tab will be opened again after reloading the objects.
This even applies to Pimcore's core tab panel ("edit", "properties", "versions" etc.). When you are on versions tab and click reload the versions tab will be opened after the object has been reloaded.


bundles/AdminBundle/Resources/public/js/pimcore/plugin/plugin.js
```diff
@@ -29,6 +29,26 @@ pimcore.plugin.admin = Class.create({
     preOpenObject: function (object, type) {
     },
     postOpenObject: function (object, type) {
+        if (type === 'object') {
+            var uiStates = localStorage.getItem('uiState_'+object.id);
```
dpfaffenbauer: better prefix uiState with `pimcore_`?

BlackbitNeueMedien: Good idea. I have changed the PR.

BlackbitNeueMedien:
Is there any problem with this PR? Perhaps is the code in the wrong place?
Or just not enough time to check it? 

Looking forward for feedback...

brusch:
@BlackbitNeueMedien Yep, the code is definitely in the wrong place, since `pimcore.plugin.admin` is the abstract class for all plugins ;) See https://github.com/pimcore/pimcore/blob/2f2af24a2fbc90718a025fdaa345b5977e715fd5/doc/Development_Documentation/20_Extending_Pimcore/13_Bundle_Developers_Guide/06_Plugin_Backend_UI.md#L16

So in my opinion the right place would be right here: 
https://github.com/pimcore/pimcore/blob/873cb55787241d58be8e6b51af5b19c6fac035d7/bundles/AdminBundle/Resources/public/js/pimcore/object/edit.js#L44
This is the class for the editing panel, where your code actually does belong to. 
Add you methods to this class and call it in the `afterrender` event handler, right after `pimcore.layout.refresh();`.

BlackbitNeueMedien:
Ah thanks. This wrong place has its origin because we developped it as a plugin but I thought it also would make sense to include this behaviour in the core. You agree? (would be nice to know before I move the code)

brusch:
@BlackbitNeueMedien oh sorry, that's not the right place either. I missed the fact that this functionality is only for the reload button :) 
So the right place would be actually here: https://github.com/pimcore/pimcore/blob/5fb48dd0f80f24d7c6d17c71958cb6bca962fd26/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js#L800

The functionality sounds nice to me, so why not :) 
But currently it doesn't work across all layout and data types, right? E.g. tabs are currently not working and probably other layout/data components as well. So if we include that into the core, we'd have to ensure that it works consistently across all types.

brusch:
@BlackbitNeueMedien jfyi: I've just deleted the comments regardings this topic which were on the wrong PR because moving is unfortunately not possible on GitHub ;) 

Looking forward to see this PR updated :)

BlackbitNeueMedien:
When I tested the last time it worked also for tabs (as I wrote also Pimcore's native tabs like version, properties etc. work) but anyway when I moved the code I will test it with some layout configurations.

brusch:
@BlackbitNeueMedien ok, maybe there was something wrong with my configuration. I'll test it again once this PR is updated. Thanks in advance!

brusch:
Awesome! Just tested, works like a charm. Thanks a lot!

weisswurstkanone:
I fear this doesn't work for localized fields in dropdown mode.
- assign "A" to en
- change "B" to de
- save and reload
- language "en" will be selected but value "B" shown => WRONG
- change language to "de" => will be "B" => everything fine
- change language back to "en" => weill be "A" => everthing fine

PR #3547, opened 2018-10-29
Fix cover transformation when focal point too close to edge

dkidd:
If a focal point is set too close to the right edge or the  bottom of an image, the cover thumbnail transformation will crop the  image too much. The resulting thumbnail will be either narrower or  shorter than it's supposed to be. This fix puts a ceiling on the cropX  and cropY values so that cropX won't be too close to the right edge, and  cropY won't be too close to the bottom.


brusch:
Thanks!

Issue #3528, opened 2018-10-24
QuantityValue: Remove (empty) unit


BlackbitDevs:
# Bug Report

### Expected behavior 
When valid units are set for a quantity value field only these should be available for selection. 
  
### Actual behavior  
Beside set valid units there is the "(empty)" option. 

If the "(empty)" unit option nevertheless is useful (I do not think so) an additional "unit mandatory" checkbox would be useful.

fashxp:
Dear community, it's been a while since that issue was updated the last time, so we decided to close it for now. If you're still interested in contributing, feel free to add comments here or contact us before creating a pull request to discuss next steps.

Thanks for your understanding.

PR #3553, opened 2018-10-31
No empty option for units

BlackbitNeueMedien:
Solves #3528


brusch:
I'd prefer an option that gives the choice whether you'd like to allow an empty unit or not, as suggested in #3528  The default behavior should be the same as it is now. 

Could you please implement that as well and update your PR? 
Thanks!

brusch:
Closed due to inactivity. 
If you'd like to update the PR, please let me know after the update ... I'll re-open and review it again 😊

PR #3555, opened 2018-10-31
Package python-uno is referred to by another package

yelmontaser:
Package python-uno is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or is only available from another source
However the following packages replace it: libreoffice-script-provider-python

  
  

Resolves #


brusch:
Thanks!

PR #3556, opened 2018-11-01
Package timeout is not available

yelmontaser:
Package timeout is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it: coreutils

  
  

Resolves #


brusch:
Thanks for your PR! 
The software list doesn't contain the package name (which can vary from distro to distro), it does list the actual program that is used, therefore `timeout` is correct, as well es `pdftotext` and others.

PR #3557, opened 2018-11-01
[WIP] Allow Multiselection of Elements in Tree

ivanmerta:
Allows multiselection of tree elements.
Allow Drag & Drop moving of multiple elements in tree.
Allow or Deny Drag & Drop of multiple elements to data components
- Multihref
- Href
- ...

## Additional info  
Click - open object
Ctrl + click - select multiple individual items
Shift + click - select multiple items in area
Alt + click - select single item and deselect

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/href.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/href.js
@@ -89,8 +89,8 @@ pimcore.object.tags.href = Class.create(pimcore.object.tags.abstract, {
                 },
 
                 onNodeOver: function (target, dd, e, data) {
-
-                    var record = data.records[0];
+				    if(data.records.length>1) return Ext.dd.DropZone.prototype.dropNotAllowed;
+					var record = data.records[0];
                     var data = record.data;
 
                     if (this.dndAllowed(data)) {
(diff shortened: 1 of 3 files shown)


brusch:
Thanks for your PR.
Unfortunately, this PR is by far not complete and not consistent in any way, as this change has affects across the entire system. Anywhere where you can drop a data object, that needs to be considered and handled correctly (error messages, legends, ...). It makes also not much sense that this works only for objects but not for assets and documents.

brusch:
Changed status to WIP

fashxp:
We really like this feature, but as @brusch said this PR is by far not complete. 
If we do it, then we should do it properly, also making it work in assets and documents tree and consider all drop targets. 
An a quick check I found about 120 places that need to be considered (all occurrences of `onNode*`). 
Any contributions are appreciated!!

ivanmerta:
Agree. I started with a small bit I had. There's still a long way to go. Thanks.

brusch:
@ivanmerta any updates on that? :)

ivanmerta:
Sorry, I’ve been too busy lately. Will jump on this during December till end of year :).

brusch:
@ivanmerta ok, thanks for coming back on this! 
We may prioritize that also on our side, so please let us know before you actually start working on that again😉

weisswurstkanone:
created the
https://github.com/pimcore/pimcore/tree/tree-multiselect
branch

which should be used for further development.

See also: 
https://github.com/pimcore/pimcore/issues/3658

PR #3565, opened 2018-11-05
Fixed the checking for validation exceptions

paulovitorbal:
So, it will not throw the notice below.
```Notice: Undefined variable: validationExceptions in /data/projects/kh-pimcore/vendor/pimcore/pimcore/models/DataObject/ClassDefinition/Data/Localizedfields.php on line 1084```

  
  

Resolves #
A notice warning when we are checking for validation exceptions on localized fields.

You can reproduce it by creating a script to import an object that has some localized fields and those fields don't have any errors.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Localizedfields.php
+++ b/models/DataObject/ClassDefinition/Data/Localizedfields.php
@@ -1089,7 +1089,7 @@ class Localizedfields extends Model\DataObject\ClassDefinition\Data
             }
         }
 
-        if (count($validationExceptions)) {
+        if (count($validationExceptions) > 0) {
             $aggregatedExceptions = new Model\Element\ValidationException();
             $aggregatedExceptions->setSubItems($validationExceptions);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: models/DataObject/ClassDefinition/Data/Localizedfields.php [hunk 1]
```diff
@@ -1089,7 +1089,7 @@ public function checkValidity($data, $omitMandatoryCheck = false)
             }
         }
 
-        if ($validationExceptions) {
+        if (count($validationExceptions)) {
```
jorisros: Shouldn't it not be bigger than 0?

paulovitorbal: I improved the conditional to make it easier to understand

brusch: `if (count($validationExceptions))` would have been alright, but anyway ;)

PR #3567, opened 2018-11-06
Fix infinite recursion when get called on object with null param

mbolka:
When a get function is called with null or empty param it won't jump into infinite recursion.
Currently in works like:

```
public function get($fieldName, $language = null) // $fieldName = null $language = null
{
    return $this->{'get'.ucfirst($fieldName)}($language); // is transformed into $this->get($language) - infinite recursion when $language is null;
}

### DIFF
--- a/models/DataObject/AbstractObject.php
+++ b/models/DataObject/AbstractObject.php
@@ -1326,7 +1326,7 @@ class AbstractObject extends Model\Element\AbstractElement
      */
     public function get($fieldName, $language = null)
     {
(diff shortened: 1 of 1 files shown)


brusch:
Maybe better to throw an exception instead? Also in the case the getter doesn't exist?

mbolka:
When a getter does not exists an exception will be raised by __call method.

brusch:
Thanks!

brusch:
Added the same check for `set()`.

PR #3568, opened 2018-11-06
Improve Email Field in Class Editor

blankse:
I created classes with multiple Email Fields with a Symfony Command.

Now I can't edit the class with the class editor. Because there this field has a hardcoded name.

In PHP side the email field is an input field with additionally validation. Nothing else right?


brusch:
This is intentional, because the email field is used in other components so it needs a fixed name. 
If you need additional email fields, just use a normal input field with a validation regex.

PR #3570, opened 2018-11-06
Remove Web-Installer from nginx config

dpfaffenbauer:
Web-Installer got removed -> so also remove it from the docs.


brusch:
Thanks!

PR #3587, opened 2018-11-12
[Pimcore 4] - Re-factored to comply with PHP 7.2 reserved name "Object"

alesak:
support for PHP 7.2 and possibly newer


pimcore/cli/startup.php
```diff
@@ -12,7 +12,7 @@
  * @license    http://www.pimcore.org/license     GPLv3 and PEL
  */
 
-use Pimcore\Model\Object;
+//use Pimcore\Model\Object
```
brusch: We shouldn't use comments, better to just remove the entire line.

pimcore/lib/Pimcore/Placeholder/ConcreteObject.php
```diff
@@ -16,7 +16,7 @@
 
 use Pimcore\Model;
 
-class Object extends AbstractPlaceholder
+class ConcreteObject extends AbstractPlaceholder
```
brusch: Please use `DataObject` as the class name, wich is also used in v5: 
https://github.com/pimcore/pimcore/blob/c62924ca5c3d0c482a71097cb0231aff143208d6/lib/Placeholder/DataObject.php#L19

pimcore/models/Element/AbstractElement.php
```diff
@@ -138,7 +138,11 @@ public function isLocked()
     public function getUserPermissions()
     {
         $elementType = Service::getElementType($this);
-        $vars = get_class_vars("\\Pimcore\\Model\\User\\Workspace\\" . ucfirst($elementType));
+        if ($elementType === 'object'){
```
brusch: We should backport and use `\Pimcore\Model\Element\Service::getBaseClassNameForElement()` from v5 here. 
https://github.com/pimcore/pimcore/blob/856b7649911c8f8e571901e883ec47e8c035163c/models/Element/AbstractElement.php#L174
https://github.com/pimcore/pimcore/blob/8808f7342acca59e56aa34bf9dac47979d11fce1/models/Element/Service.php#L269

dpfaffenbauer:
why don't you do the same as Pimcore did? rename the whole Namespace to DataObject?

alesak:
> why don't you do the same as Pimcore did? rename the whole Namespace to DataObject?

I was trying to keep it as compatible with previous versions as possible, if i change the namespace there is good chance it breaks many existing sites. It won't be too complicated to fix but current way should work without any changes necessary.

brusch:
@alesak so from your side this should be complete if I understood it correctly, right?

brusch:
@alesak because the tests are still failing, maybe you can have another look at this, thanks!

alesak:
> @alesak because the tests are still failing, maybe you can have another look at this, thanks!

@brusch Tests for PHP 5.6 - 7.1 are OK  now, I have added a test for PHP 7.2 but it is failing due to deprecated methods used by Zend1 and I am not sure what would be the best way to handle this. Any suggestions?

brusch:
@alesak thanks! I'm wondering why the deprecation messages are sent to the output, since they should be disabled: https://github.com/pimcore/pimcore/blob/2d306f9e67e140130dcf5443b5b8e8da03e043c1/pimcore/lib/Pimcore.php#L510

Maybe you can further investigate why they appear in the response and how we can suppress them. 
Thanks again!

alesak:
> @alesak thanks! I'm wondering why the deprecation messages are sent to the output, since they should be disabled:
> 
> [pimcore/pimcore/lib/Pimcore.php](https://github.com/pimcore/pimcore/blob/2d306f9e67e140130dcf5443b5b8e8da03e043c1/pimcore/lib/Pimcore.php#L510)
> 
> Line 510 in [2d306f9](/pimcore/pimcore/commit/2d306f9e67e140130dcf5443b5b8e8da03e043c1)
>          error_reporting(E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED); 
> 
> Maybe you can further investigate why they appear in the response and how we can suppress them.
> Thanks again!

All tests looks good, it looks done from my side, now it is on you guys, lets hope there are not many bugs introduced by this update

kachaj7:
In this commit was made a mistake:
https://github.com/WebForPeople/pimcore/commit/2690820ebe50e63b08de439558a8db91af39be39

After that commit the function get_class_vars gets string
\Pimcore\Model\User\Workspace\DataObject but this class doesn't exist and before commit this was
\Pimcore\Model\User\Workspace\ObjectWorkspace.

kachaj7:
Next Exception:
PHP Fatal error:  Uncaught Error: Class '\Pimcore\Model\User\Workspace\Object' not found in /var/www/(...)/htdocs/pimcore/models/User/UserRole/Dao.php:65

$className = "\\Pimcore\\Model\\User\\Workspace\\" . ucfirst($classNameType)
uses \\Pimcore\\Model\\User\\Workspace\\Object but this class doesn't exist and should be
\\Pimcore\\Model\\User\\Workspace\\ObjectWorkspace.

kachaj7:
Can somebody check if this is a bug?

kachaj7:
Really no feedback?

PR #3590, opened 2018-11-12
Fix Double Nested Array in ObjectData Preview Url Generator

solverat:
Use `[]` instead of `[[]]` in link generator options.


brusch:
Mille grazie!

PR #3591, opened 2018-11-12
Fix findByType not available in system settings

dpfaffenbauer:
`findByType` is not available since ExtJs 4 anymore.... it was replaced with `query` in ExtJs 4


brusch:
Thanks!

PR #3596, opened 2018-11-13
Fix sorting variants

bramstroker:
Fix bug when manually sorting objects in the tree. This was not working correctly for variants as the indices of siblings were not updated.


fashxp:
thx

PR #3597, opened 2018-11-13
globalParams

j-jonas:
Merge globalParams to all allowed brick configs.

At the moment it is not possible to set a globalParams like it is described here
https://pimcore.com/docs/5.x/Development_Documentation/Documents/Editables/Areablock/index.html

Without this fix you need to set all globalParams in a array like:
'globalParams' => [[
        'containerClass' => 'mt-5',
]],

  
  

Resolves #


brusch:
Obsolete by 183bc31a44a8b9d56f0715c6ef8ff0a2ca0285ab

PR #3599, opened 2018-11-13
fix doc-blocks for relational types.

dpfaffenbauer:
currently it happens that those types are generated like

```\Pimcore\Model\Asset\image[]```

or it could happen like

```\Pimcore\Model\DataObject\product[]``` which is not correct, since classes are always upper-cased.

  
  

Resolves #


weisswurstkanone:
thx!

PR #3612, opened 2018-11-16
Pimcore 5.5.3 - Pimcore\Model\DataObject\ClassDefinition\Data\Localizedfields#doGetDataForEditMode missing if-Statements

bruenner-elements:
In `doGetDataForEditMode` there were some IF-statements missing that caused code-breaks when  localized-fields of object-parents or bricks were called due to inheritance.


brusch:
Thanks!

PR #3613, opened 2018-11-16
Replace % sign in asset keys

BlackbitNeueMedien:
When uploading an asset or creating a folder with a % sign in its name, the asset is not accessible via its URL. After uploading an asset with name "100%cotton.png", for example, its neither accessible via "example.com/100%cotton.png" nor "example.com/100%25cotton.png" (with the encoded percent sign).
I suggest to replace the % sign in asset keys as in this PR. I suspect that this is somehow related to the rewrite rules for assets, so changing them might also be a solution. But I have no idea what side effect that may have.


brusch:
Thanks!

PR #3623, opened 2018-11-19
add hardlink wrappers for web2print documents

markus-moser:
Currently hardlink wrappers for web2print documents are missing. Therefore hardlinks in combination with web2print documents do not work correctly. In the old pimcore 4 web2print plugin these wrappers where delivered within the plugin...


brusch:
Grazie mille!

PR #3624, opened 2018-11-19
Support style field for objects tag in pricing coditions

jremmurd:
This pull request adds support for setting the `style` property for the objects tag in pricing coditions.

```JS    
new pimcore.bundle.EcommerceFramework.pricing.config.objects(data, {
                    name: "...",
                    title: "...",
                    visibleFields: "path",
                    height: 200,
                    width: 500,
                    style: 'float: left; margin-right: 10px'  // was not supported yet
                    ...
}
```


fashxp:
thx

PR #3629, opened 2018-11-20
Variant disappear issue fixed

### LINKED ISSUE #3602: Variant disappears after moving out of object context
Reproducable on demo-basic:

- create an object 
- add a variant to the object
- move variant to parent folder
- refresh folder
-> Variant is gone - and as far I've tested it, there is no way to get it back (except changing the parent id in database)

Solution: Disallow dropping variants on folders?

psuhane:

Resolves #3602


fashxp:
thx!

PR #3630, opened 2018-11-20
fix multihref-metadata boolean

dpfaffenbauer:
Current solution produces this:

bildschirmfoto 2018-11-20 um 13 30 22

Cause the ending parenthesis is in the wrong place.

with this PR it looks like this:

bildschirmfoto 2018-11-20 um 13 31 13


weisswurstkanone:
thx

PR #3636, opened 2018-11-21
Set group when importing from JSON definition.

IgorBenko:
Issue: When importing class definition from exported JSON file the group is ignored.

Fix: Set the group when importing from JSON definition.


brusch:
Thanks!

PR #3644, opened 2018-11-23
Improve 'Send Test Email'

ultramegatom:
add option to choose a document
add option to add parameters

add from field (prefilled with system settings)

add send test mail button to email and newsletter documents (will prefill the email window with data)

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/EmailController.php
+++ b/bundles/AdminBundle/Controller/Admin/EmailController.php
@@ -356,7 +356,7 @@ class EmailController extends AdminController
 
         if ($request->get('emailType') == 'text') {
             $mail->setBodyText($request->get('content'));
-        } elseif($request->get('emailType') == 'text') {
+        } elseif($request->get('emailType') == 'html') {
             $mail->setBodyHtml($request->get('content'));
         } elseif($request->get('emailType') == 'document') {
             $doc = \Pimcore\Model\Document::getByPath($request->get('documentPath'));
--- a/bundles/AdminBundle/Controller/Admin/EmailController.php
+++ b/bundles/AdminBundle/Controller/Admin/EmailController.php
@@ -346,14 +346,37 @@ class EmailController extends AdminController
         }
 
         $mail = new Mail();
+
+        if($from = $request->get('from')) {
+            $mail->setFrom($from);
+        }
         $mail->addTo($request->get('to'));
         $mail->setSubject($request->get('subject'));
         $mail->setIgnoreDebugMode(true);
 
-        if ($request->get('type') == 'text') {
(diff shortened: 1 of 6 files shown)


ultramegatom:
this is how it looks:

image

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
```diff
@@ -1628,60 +1628,195 @@ pimcore.helpers.searchAndMove = function (parentId, callback, type) {
...
+            {name: 'key', type: 'string'},
+            {name: 'value', type: 'string'}
+        ]
+    });
+
+    var paramGrid = Ext.create('Ext.grid.Panel', {
+        store: paramsStore,
+        columns: [
+            {
+                text: 'key',
```
brusch: we should use a a translation key here: `t("key")`

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
```diff
@@ -1628,60 +1628,195 @@ pimcore.helpers.searchAndMove = function (parentId, callback, type) {
...
+                dataIndex: 'key',
+                editor: new Ext.form.TextField(),
+                width: 200
+            },
+            {
+                text: 'value',
```
brusch: same here: `t("value")`

bundles/AdminBundle/Controller/Admin/EmailController.php
```diff
@@ -346,14 +346,37 @@ public function sendTestEmailAction(Request $request)
...
+        }
         $mail->addTo($request->get('to'));
         $mail->setSubject($request->get('subject'));
         $mail->setIgnoreDebugMode(true);
 
-        if ($request->get('type') == 'text') {
+        if ($request->get('emailType') == 'text') {
             $mail->setBodyText($request->get('content'));
-        } else {
+        } elseif($request->get('emailType') == 'text') {
```
brusch: This should be `} elseif($request->get('emailType') == 'html') {` right?

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
ultramegatom: done

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
ultramegatom: done

bundles/AdminBundle/Controller/Admin/EmailController.php
ultramegatom: good catch :)

done

brusch:
Thanks!

PR #3645, opened 2018-11-23
Changes private attributes to protected ones in TagHandler

strati1991:
When decorating the TagHandler class the method handleBrickActionResult and the attribute $responseStack needs to be reimplemented in the decorating class because these are private in the original TagHandler. 

Changing them to protected would prevent reimplementation.


brusch:
Thanks!

PR #3647, opened 2018-11-23
::newinstance ist now new \Swift_Attachment

Gerhard-Kanzler:
Before you get an Error by sending an Attachment.


brusch:
Thanks!

PR #3648, opened 2018-11-23
add null smtp transport for dev environments where no sendmail nor smtp is available

dpfaffenbauer:
Since the update to SwiftMailer 6, the MailTransport is gone, thus resulting in a lot of issues for developers. Cause sendmail is not an option on most systems (cause its either deactivated or doesn't support TLS). So an SMTP Server has to be used. (Thus means creating a SMTP account for every project, but so it is).

This makes it easier for developers to just don't use any transport at all in dev/staging systems (or possibly also production which doesn't make any sense TBH). The developer can see the email in the backend log then.


dpfaffenbauer:
Alternative is to completely overwrite the swiftmailer configuration in a `dev` or `local` config:

```yml
swiftmailer:
    mailers:
        pimcore_mailer:
            transport: 'null'
        newsletter_mailer:
            transport: 'null'
```

PR #3650, opened 2018-11-23
introduce new delete-info event

dpfaffenbauer:
which allows to determine what elements can be deleted from backend and which don't.

*Cause*
There is a Problem with pre-delete events. They would work in cases where the element doesn't have any childs. When the element has childs, Pimcore deletes those first and then the "selected" element. 

This event looks like that:

*Objects*
http://g.recordit.co/gADtiNMZY7.gif
http://g.recordit.co/6IgWGJ1out.gif

*Documents*
http://g.recordit.co/sYbBgnYYO9.gif

*Assets*
http://g.recordit.co/RgoK2PNjbh.gif

### DIFF
--- a/lib/Event/DataObjectEvents.php
+++ b/lib/Event/DataObjectEvents.php
@@ -67,6 +67,13 @@ final class DataObjectEvents
      */
     const POST_UPDATE_FAILURE = 'pimcore.dataobject.postUpdateFailure';
 
+     /**
+     * @Event("Pimcore\Event\Model\DataObjectDeleteInfoEvent")
+     *
+     * @var string
+     */
+    const DELETE_INFO = 'pimcore.dataobject.deleteInfo';
+
     /**
      * @Event("Pimcore\Event\Model\DataObjectEvent")
      *
--- a/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
+++ b/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
@@ -15,8 +15,17 @@
 namespace Pimcore\Bundle\AdminBundle\Controller\Admin;
 
 use Pimcore\Bundle\AdminBundle\Controller\AdminController;
+use Pimcore\Event\AssetEvents;
+use Pimcore\Event\DataObjectEvents;
+use Pimcore\Event\DocumentEvents;
+use Pimcore\Event\Model\AssetDeleteInfoEvent;
+use Pimcore\Event\Model\DataObjectDeleteInfoEvent;
+use Pimcore\Event\Model\DocumentDeleteInfoEvent;
+use Pimcore\Event\Model\ElementDeleteInfoEventInterface;
 use Pimcore\Logger;
 use Pimcore\Model\Asset;
+use Pimcore\Model\DataObject\AbstractObject;
+use Pimcore\Model\Document;
 use Pimcore\Model\Element\ElementInterface;
 use Pimcore\Model\Element\Service;
 use Symfony\Component\HttpFoundation\JsonResponse;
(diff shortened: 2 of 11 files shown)


lib/Event/DataObjectEvents.php
```diff
@@ -67,6 +67,13 @@
      */
     const POST_UPDATE_FAILURE = 'pimcore.dataobject.postUpdateFailure';
 
+     /**
+     * @Event("Pimcore\Event\Model\DataObjectEvent")
```
weisswurstkanone: maybe change to DataObjectDeleteInfoEvent  ?

dpfaffenbauer: done

dpfaffenbauer:
btw, this is the test-listener for the recordings:

```php
<?php

namespace AppBundle\EventListener;

use Pimcore\Event\Model\ElementDeleteInfoEventInterface;
use Pimcore\Model\Asset;
use Pimcore\Model\DataObject\AbstractObject;
use Pimcore\Model\Document;

class DeleteInfoEventListener
{
    public function deleteInfoEvent(ElementDeleteInfoEventInterface $deleteInfoEvent)
    {
        $element = $deleteInfoEvent->getElement();
        $allowed = false;
        $reason = '';

        if ($element instanceof AbstractObject) {
            $allowed = $element->getId() % 2 === 0;
            $reason = 'Cause your id is even.';
        }
                
        if ($element instanceof Asset) {
            $allowed = $element->getId() % 2 === 1;
            $reason = 'Cause your id is odd.';
        }
                
        if ($element instanceof Document) {
            $allowed = $element->hasChildren();
            $reason = 'Cause you have childs';
        }

        if ($allowed) {
            $deleteInfoEvent->setReason($reason);
            $deleteInfoEvent->setDeletionAllowed(false);
        }
    }
}
```

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -2337,7 +2364,7 @@ public function previewAction(Request $request)
                 }
             }
         } elseif ($linkGenerator = $object->getClass()->getLinkGenerator()) {
-            $url = $linkGenerator->generate($object, ['preview' => true, 'context' => $this]);
+            $url = $linkGenerator->generate($object, [['preview' => true, 'context' => $this]]);
```
fashxp: on purpose?

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -1214,7 +1242,6 @@ protected function updateIndexesOfObjectSiblings(DataObject\AbstractObject $upda
             'o_parentId = ? AND o_id != ?',
             [$updatedObject->getParentId(), $updatedObject->getId()]
         );
-        $list->setObjectTypes([DataObject\AbstractObject::OBJECT_TYPE_OBJECT, DataObject\AbstractObject::OBJECT_TYPE_VARIANT, DataObject\AbstractObject::OBJECT_TYPE_FOLDER]);
```
fashxp: on purpose? 
seems you started from an old tag?

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
dpfaffenbauer: yes, well, I started implementing that from another project and just copied it into my Pimcore Core fork :D, I will fix that later.

dpfaffenbauer: fixed

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
dpfaffenbauer: fixed

brusch:
@dpfaffenbauer Sorry that merging #3671 introduced some conflicts, could you please resolve them? Thanks!

dpfaffenbauer:
done

weisswurstkanone:
thx

Issue #3614, opened 2018-11-16
Improvements for objects and objects with metadata


SEYentl:
# Feature Request 
The feature requests described below were recurring questions from clients of ours concerning the grids that are used for objects, objects with metadata,....

### Objects - select visible fields
It would be nice if objects have the same "visible fields" option in the class definition like objects with metadata. At this point we misuse objects with metadata to have just this functionality. 

![screenshot 2018-11-16 at 10 45 13](https://user-images.githubusercontent.com/40569826/48613867-2c700700-e98d-11e8-9a59-a1a798607c2f.png)

This misuage can bring some additional problems since the structure of "objects" is not the same as the structure of "objects with metadata"

### general remark of visible fields option
At this moment the "visible fields" option is only adding fields. It would be nice if only the selected fields are shown.

For example, if I only select two fields (erpkey and name) I would expect that only those two fields are shown. At this moment this is not the case since the field "id" is always shown.

**Screenshot**:
![screenshot 2018-11-16 at 10 52 54](https://user-images.githubusercontent.com/40569826/48614118-ca63d180-e98d-11e8-9ad2-e06bca831e44.png)

### general remark of column width for objects and objects with metadata
At this moment the default size of columns cannot be defined. This means that the user has to resize these columns manually every single time that they make use of it. It would be nice that the column width could be set for the properties that are selected in the "visible fields"

Additionally it would be handy that in the search functionality of these fields the exact same grid configuration is used. 

### footnote
I hope that the above improvements make sense, as I said these improvements come directly as feedback from clients that use Pimcore on a day-to-day basis. If I'm missing something, please do let me know 😃

jacobsyentl:
@psuhane Exactly that, it would be very nice if this is functionality is implemented for 'object relation' with only one class selected.

I understand that this is not possible for multiple classes.

amarjeet6694:
What should be the UI if option to select **width** for each field is provided alongside **Visible field** section? 

I guess the complete UI for the visible fields needs to be changed then. 
Otherwise, default column width defined in the Class that is selected in the **Allowed Class** can be used.

SEYentl:
@amarjeet6694 Well it can go two ways, either the visible field section gets added here: 
![screenshot 2018-11-22 at 08 13 37](https://user-images.githubusercontent.com/40569826/48887145-8c9ff680-ee2e-11e8-87e4-d28cf373205f.png)

In this case the column to define the width of a field is already in place.

[...]

PR #3653, opened 2018-11-26
[Objects] Rename relational data-types to a more meaningful name

### LINKED ISSUE #3575: [Objects] Rename relational data-types to a more meaningful name
- `Href` -> `ManyToOneRelation`
- `Multihref` -> `ManyToManyRelation`
- `Multihref Advanced` -> `AdvancedManyToManyRelation`

- `Objects` -> `ManyToManyObjectRelation`
- `Objects with Metadata` -> `AdvancedManyToManyObjectRelation`

brusch:
Resolves #3575


dpfaffenbauer:
will there be a bc layer as well?

brusch:
@dpfaffenbauer for PHP class names yes: https://github.com/pimcore/pimcore/pull/3653/files#diff-9927b4f9e0ba22a3373a57d5e406ca99

brusch:
So existing JSON exports and PHP definition files will continue to work. 
If you're extending a JS class of one of the affected data-types for a custom data-type, you'd have to change the class name there.

brusch:
> shouldn't you also introduce a bc layer for the JS files?

Done.

ultramegatom:
shouldn't the renaming also apply to document editables? Otherwise it would be really confusing

dpfaffenbauer:
BC Layer: perfect 👍  Pimcore should be able to run code from 5.5 on 5.6 without changing anything.

ultramegatom:
Whats actually the reason for the 'object-relations'? Is there anything this datatype can do what the normal relation can't? Shares about the same codebase - couldn't it just extend the 'normal relations' and set the restrictions to objects?

brusch:
> shouldn't the renaming also apply to document editables? Otherwise it would be really confusing

@ultramegatom That's actually the plan ;) But this PR will only contain the relevant changes for data objects

brusch:
> Whats actually the reason for the 'object-relations'? Is there anything this datatype can do what the normal relation can't? Shares about the same codebase - couldn't it just extend the 'normal relations' and set the restrictions to objects?

@ultramegatom That has mainly historical reasons. But #3614 requests some additions that gives back this data-type a reason to exist :)

ultramegatom:
👍

PR #3655, opened 2018-11-26
Update ClassController.php

Lexipowder:
Make parent $object available in a provider class if Select is on a fieldcollection


brusch:
Thanks!

PR #3656, opened 2018-11-26
Cleanup orders by modification date, not orderdate

andreas-gruenwald:
The orderdate is set on order creation, so when a user aborts and comes back later, there is a good chance that the cleanup takes place while the payment process is active. In this case it can happen that the payment provider executes the payment, while the Pimcore order becomes aborted. 

The update proposal also includes version notes to improve debugging based on the saved version note.


brusch:
Thanks a lot!

PR #3661, opened 2018-11-27
[Email] Consider multiple recipients and email name annotation 'NAME <name@email.com>'

ultramegatom:
send test email consider multiple recipients and email name annotation `NAME <name@email.com>`


brusch:
We have the same functionality already here: 
https://github.com/pimcore/pimcore/blob/f70577a6366ab57a660750bcd7f599bcc9cbfdd3/lib/Mail.php#L528 

So i guess we should unify that by using the newly introduced function, to avoid duplicate code. 
Can you just quickly fix that? 

Thanks a lot!

ultramegatom:
> We have the same functionality already here:
> [pimcore/lib/Mail.php](https://github.com/pimcore/pimcore/blob/f70577a6366ab57a660750bcd7f599bcc9cbfdd3/lib/Mail.php#L528)
> 
> Line 528 in [f70577a](/pimcore/pimcore/commit/f70577a6366ab57a660750bcd7f599bcc9cbfdd3)
> 
>              if (!empty($fromArray)) { 
> So i guess we should unify that by using the newly introduced function, to avoid duplicate code.
> Can you just quickly fix that?
> 
> Thanks a lot!

yes sure can do.

should we also remove the get(From|To|ReplyTo) methods from the models? e.g. https://github.com/pimcore/pimcore/blob/8b04bd00a02e7f138aaeedf58ba4c5425a4333f1/models/Document/Email.php#L208

And refactor the usages in lib\Email ? As far as i can tell the only usages are there.

ultramegatom:
@brusch updated

I also found something similar in the 'email log': 
https://github.com/pimcore/pimcore/blob/master/models/Tool/Email/Log.php#L661

The Email Addresses are however formated like: ```mail@mail.com (Name)```

Should we refactor this to? I think we would have to change all current 'email logs' in the database within an update script.

brusch:
@ultramegatom yep, that would be definitely nice in my opinion!

ultramegatom:
@brusch i added a migration script. could you test if it works in your setup?

thx

brusch:
@ultramegatom I'm afraid that you last push messed up your PR, it contains also commits from the master branch which were not merged properly, so that they show up as changes of this PR :(

ultramegatom:
@brusch hmm strange. i merged with the master again - should be fine now

brusch:
@ultramegatom that looks better ;)

brusch:
Cherry-picked the log migration into #3743

PR #3668, opened 2018-11-28
fix: don't set brick if brick is empty

### LINKED ISSUE #3667: [Import] "Container could not be resolved" with empty Object brick
https://github.com/pimcore/pimcore/blob/7f5a87b58b146fe48ec5db03104c21e695be0ed3/lib/DataObject/Import/ColumnConfig/Operator/ObjectBrickSetter.php#L84
$brick variable is not necessary initialized, so we must check it before call "process" function.

rbonfils:
Resolves #3667 

## Additional info  
Check if brick is not empty before call "process" function on childs. Because brick is not necessary defined, calling "process" with a target on NULL displaying error on CSV import.


weisswurstkanone:
thx!

PR #3679, opened 2018-11-29
Fix correct spelling of "Referer"

### LINKED ISSUE #3678: Bug: "Referrer" is used instead of "Referer"
According to [RFC 2616][1], the name of the header is "Referer" and not "Referrer".

https://github.com/pimcore/pimcore/blob/4c986d60d890898e466533a21dfd29bde2acb9f6/lib/Targeting/Condition/ReferringSite.php#L59

[1]: https://tools.ietf.org/html/rfc2616#page-140

ihmels:
According to [RFC 2616][1], the header name is "Referer" instead of "Referrer".

[1]: https://tools.ietf.org/html/rfc2616#page-140

  
  

Resolves #3678


brusch:
Good catch - Thanks!

PR #3683, opened 2018-12-01
pump min PHP to 7.2

dpfaffenbauer:
since 7.1 is not supported anymore with tomorrow

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ sudo: false
 language: php
 
 php:
-  - 7.1
   - 7.2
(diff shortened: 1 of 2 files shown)


dpfaffenbauer:
http://php.net/supported-versions.php

brusch:
7.1 is not EOL, security support is provided one more year until Dec 2019. 
So we'll keep 7.1 support for now.

dpfaffenbauer:
True, but that should be reason enough to move to 7.2

PR #3693, opened 2018-12-04
Use readonly attribute for readonly input fields - not disabled

BlackbitNeueMedien:
When you set an input field to be not editable, Pimcore uses ExtJS's disable() method in the input field. This leads to being not able to highlight or copy contents of such readonly fields in Firefox. Anyway readonly is the more approprate attribute for setting something to be not editable.
With this PR readonly attribute is used.

### DIFF
--- a/bundles/AdminBundle/Resources/public/css/ext-modifications.css
+++ b/bundles/AdminBundle/Resources/public/css/ext-modifications.css
@@ -253,7 +253,7 @@ body > .x-mask {
     background-image: url(../js/lib/ext/classic/theme-triton/resources/images/tree/drop-between.gif);
 }
 
-.x-form-item-default.x-item-disabled {
+.x-form-item-default.x-item-disabled, .x-form-item-default.x-form-readonly {
     opacity: 0.7;
 }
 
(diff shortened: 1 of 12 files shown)


brusch:
If we change that, we should change it for all affected data-types, right? 

That would be additionally: 
- Textarea
- Quantity Value
- Input Quantity Value
- Date
- Date Time
- Time
- Select (incl. all descendants) 
- External Image
- Calculated Field

BlackbitNeueMedien:
Tested with normal text input field and tooltip works as expected also with readonly.

Will update the other data types tomorrow...

brusch:
@BlackbitNeueMedien Thanks!

BlackbitNeueMedien:
Ok, I added the data types mentioned by @brusch - only Ext.ux.form.MultiSelect does not support setReadOnly(). Thus I kept disable for multiselect.

brusch:
@BlackbitNeueMedien looks good, thanks a lot!

fashxp:
date time does not support setReadonly ... see #3718

PR #3697, opened 2018-12-04
Refactoring treenodelocator

### LINKED ISSUE #3520: Showing tree (with pagination and subfolders) is not working in Pimcore 5.4.4
# Bug Report
 
  
### Expected behavior 
 There should be no error popup
### Actual behavior  
  It displays error. PFA
showintree

### Steps to reproduce  
https://demo-advanced.pimcore.org/admin/?_dc=1540206620
  1) Search an object id 1511 and click on show in tree icon
### Code example 
```php
// your code
```

# Feature Request

### LINKED ISSUE #3554: "Show in tree" results in JS request loop (in combination with perspectives/views)
This is propably related to #3520 but I have some other strange things happening.

Actually I did some testing with the "Show tree node" because this function did not work in a scenario with multiple object views (e.g. /products, /categories, ...) - just objects of the *first* tree were resolved. However I run into a much more severe bug - here's the boiled-down test setup. 

### Steps to reproduce  

Just set up a clean 5.5.3 install and did following:

- Created a class "Test" with a single property "name".
- In "Objects", created two root folders "tree1" and "tree2.
- Created in objects "/tree2" a bunch of folders (just enought in order to have tree paging)
- Created an object of class "Test" in the last folder (on page 2) - not sure if this matters though

Set up following /app/config/pimcore/customviews.php:
```
<?php

return [
    'views' => [
        [
            'treetype' => 'object',
            'name' => 'view 1',
            'condition' => null,
            'id' => 1,
            'rootfolder' => '/tree1',
            'showroot' => false
        ],
        [
            'treetype' => 'object',
            'name' => 'view 2',
            'condition' => null,
            'id' => 2,
            'rootfolder' => '/tree2',
            'showroot' => false
        ]
    ]
];
```
...and /app/config/pimcore/perspectives.php:
```
<?php

return [
    'default' => [
        'iconCls' => 'pimcore_icon_perspective',
        'elementTree' => [
            [
                'type' => 'documents',
                'position' => 'left',
                'expanded' => false,
                'hidden' => false,
                'sort' => -3
            ],
            [
                'type' => 'assets',
                'position' => 'left',
                'expanded' => false,
                'hidden' => false,
                'sort' => -2
            ],
            [
                'type' => 'objects',
                'position' => 'left',
                'expanded' => false,
                'hidden' => false,
                'sort' => -1
            ]
        ]
    ],
    'custom view' => [
        'elementTree' => [
            [
                'type' => 'customview',
                'id' => 1,
                'position' => 'left'
            ],
            [
                'type' => 'customview',
                'id' => 2,
                'position' => 'right'
            ]
        ]
    ]
];
```
- Reloaded page and switched to perspecitve "custom view"
- Opened the "Test" object
- Navigated back to page 1 in tree pager "view 2"

It should look like this now...
image

Click on "Show in tree" in the "Test" object

I end up in a never ending loop requesting tree-get-childs-by-id:
image

gadget60:
This PR is a complete refacturing of the "Show in Tree" function in Pimcore backend.
Current implementation has many flaws. These are mainly:

- Resolving nodes below tree paging nodes sometimes works, sometimes it does not, see https://github.com/pimcore/pimcore/issues/3520
  It mainly fails when dealing with multiple trees of the same type (custom views), like (objects)/products and (objects)/categories. The following object in demo-advanced for example will be resolved successfully in default view, but not in PIM view: https://demo-advanced.pimcore.org/admin/login/deeplink?object_9155_object 

- Tree search filters are not supported, an active tree search that hides the requested node or one of its ancestors won't be cleared.

- Current implementation is mainly based on try/catches - there is no proper state. This makes it really hard to decide what to do next if one part of the resolving process goes wrong. This leads to strange behaviours like the one described in https://github.com/pimcore/pimcore/issues/3554

- Current implementation blindly iterates over all possible trees. When dealing with custom views you probably have trees with a root shifted to a subfolder/object. This allows us to check if a node is out of scope of a tree before processing it.

Resolves #3520
Resolves #3554

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/treenodelocator.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/treenodelocator.js
@@ -246,7 +246,7 @@ pimcore.treenodelocator = function()
 
             // Reload the tree starting from given node once
             // This solves two issues: All subsequent search filters are reset and we know
-            // that the tree store us up to date:
+            // that the tree store is up to date:
             if (treeState.reloaded == false) {
                 self.reloadTree(node);
                 return;
@@ -397,7 +397,7 @@ pimcore.treenodelocator = function()
                 newPage = (pagingState.minPage + pagingState.maxPage) / 2;
                 self.switchToPage(newPage);
             } else {
-                // Child node was supposed to be on current page, but obiously isn't.
+                // Child node was supposed to be on current page, but obviously isn't.
                 // It may be filtered using a custom view or not being displayed for other
                 // reason - anyway there is nothing more to do here:
                 self.reportProcessPagingFailed();
--- a/bundles/AdminBundle/Resources/public/js/pimcore/treenodelocator.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/treenodelocator.js
@@ -1,377 +1,558 @@
-pimcore.registerNS("pimcore.treenodelocator.x");
+pimcore.registerNS("pimcore.treenodelocator");
 
-pimcore.treenodelocator.showInTree = function (id, elementType, button) {
+pimcore.treenodelocator = function()
+{
 
-    if (button) {
-        button.disable();
-    }
+    /**
+     * Private vars
+     */    
 
-    var callback = function () {
-        if (button) {
-            button.enable();
-        }
-    }
-    pimcore.treenodelocator.startShowInTree(id, elementType, callback, 0);
+    // Holds the state
+    var busy = false;
+    
+    // The states hold current data for each phase of a tree location task.
+    var globalState = null;
+    var treeState = null;
+    var pagingState = null;
+    
+    // Holds the current button.
+    var currentButton = null;
+    
+    // Holds loading indicators type/ids.
+    var loadingIndicators = [];
+    
+    
+    /**
+     * Private functions
+     */
+    var self = {
 
-
-}
-
-pimcore.treenodelocator.startShowInTree = function(id, elementType, callback, treeIdx) {
-    Ext.Ajax.request({
-        url: "/admin/element/type-path",
-        params: {
-            id: id,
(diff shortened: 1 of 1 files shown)


brusch:
@gadget60 Thank you very much for your PR. 

I've just tested your PR and I've seen that filtered trees are still not supported, see: 


Anything still wrong?

brusch:
@gadget60 everything else looks good - well done!

gadget60:
Really? Works in my local install of the ecommerce demo (freshly installed yesterday - just replaced the treenodelocator.js straight in /vendor/pimcore/.... Could you please double check...?
Sorry, can't make such a nice video - but these are the screenshots before/after:

image
image

Having that said, I've just seen that i forgot to disable the passed button on showInTree() start. Should I create another PR or is this messing things up?

Thanks for Prio-Backlog - this is really important to us, as data structure cannot always be defined independant of customer's perception, thus having tree paging is inevitale in large projects  :-)

brusch:
@gadget60 you're done?

gadget60:
Hallo Bernhard

Ehrlich gesagt – ich stehe etwas auf dem Schlauch und kämpfe grand mit dem neuen Pull Request :-)
Muss ich mein geklontes Repo zuerst mergen…?

Gruss
Michael



Von: Bernhard Rusch <notifications@github.com>
Gesendet: Mittwoch, 5. Dezember 2018 15:19
An: pimcore/pimcore <pimcore@noreply.github.com>
Cc: Reufer Michael <M.Reufer@bag.ch>; Mention <mention@noreply.github.com>
Betreff: Re: [pimcore/pimcore] Refactoring treenodelocator (#3697)


@gadget60<https://github.com/gadget60> you're done?

gadget60:
Sorry - yes, I'm done. 
My first PR :-)

PR #3699, opened 2018-12-05
Fix: OwnerAwareField shouldn't save Owner

### LINKED ISSUE #3698: OwnerAwareField shouldn't save Owner
# Bug Report

I have a Link Localizedfield in a class with many languages.

The Dao couldn't load the data, because the serialized string is cut off in the database.

Here is the owner set:
https://github.com/pimcore/pimcore/blob/master/models/DataObject/Localizedfield/Dao.php#L537

I think it should be removed before saving.

### Steps to reproduce 

1. Create class with panel, localizedfield, input and link
test
2. Create data object with filled input and link
3. Save and Publish
3. Reload the tab (Required!)
4. Change the input
5. Save and Publish
6. The owner is filled in the db table
test2

blankse:
Resolves #3698

### DIFF
--- a/models/DataObject/Localizedfield/Dao.php
+++ b/models/DataObject/Localizedfield/Dao.php
@@ -145,17 +145,16 @@ class Dao extends Model\Dao\AbstractDao
 
                     $fd->save($this->model, $childParams);
                 } else {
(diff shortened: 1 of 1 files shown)


brusch:
See #3698

PR #3703, opened 2018-12-05
Correct field name: o_modificationDate instead of o_modification

andreas-gruenwald:
The actual name of the SQL field is "o_modificationDate", not "o_modification".
The bugfix is important so that aborted orders are cleaned up properly.


brusch:
Thanks!

PR #3704, opened 2018-12-05
Cleanup unused order items

ultramegatom:
This will delete orderItems if they are unused.


brusch:
Looks good, thanks a lot!

dpfaffenbauer:
shouldn't you better tackle the issue why this even happens in the first place? I mean, why are you creating objects you don't need?

ultramegatom:
@dpfaffenbauer 

I'll give you an example:

- User has 2 items in his cart
- User procedes to checkout and starts payment (Order is created along with 2 OrderItems)
- User cancles payment
- User removes 1 item from the cart
- User procedes to checkout and starts payment again (We use the same Order - We adopt the references in the Order correctly - however the removed OrderItem-Object will stay)

Its not an issue at all - its more of a cosmetic cleanup

PR #3705, opened 2018-12-05
Sort backend search results by relevance

BlackbitNeueMedien:
Currently backend search results are not sorted for search via menu nor on grid if no columns are set to be used for sorting.
Moreover in https://github.com/pimcore/pimcore/blob/8ef2d6610e438198e11fd381b0172883856213d7/bundles/AdminBundle/Controller/Searchadmin/SearchController.php#L227 general sorting is commented out for performance reasons.

This PR adds sorting by relevance via MySQL's full text search. As the table `search_backend_data`already has a fulltext index on columns data and properties for filtering the same index could be used for sorting without much performance problems. I have tried this sorting on customers with 12.000 and 33.000 objects and search is as fast as before.

### DIFF
--- a/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
+++ b/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
@@ -217,9 +217,16 @@ class SearchController extends AdminController
         $searcherList->setOffset($offset);
         $searcherList->setLimit($limit);
 
-        // do not sort per default, it is VERY SLOW
-        //$searcherList->setOrder("desc");
-        //$searcherList->setOrderKey("modificationdate");
+        $orderQuery = $query;
+        if ($orderQuery === '*') {
+            $orderQuery = '';
+        }
+        $orderQuery = \str_replace(['+','-'.'(',')','"',"'"], '', $orderQuery);
+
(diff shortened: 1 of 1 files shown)


brusch:
@BlackbitNeueMedien oh sorry, that's not the right place either. I missed the fact that this functionality is only for the reload button :) 
So the right place would be actually here: https://github.com/pimcore/pimcore/blob/5fb48dd0f80f24d7c6d17c71958cb6bca962fd26/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js#L800

The functionality sounds nice to me, so why not :) 
But currently it doesn't work across all layout and data types, right? E.g. tabs are currently not working and probably other layout/data components as well. So if we include that into the core, we'd have to ensure that it works consistently across all types.

BlackbitNeueMedien:
Actually it worked for tabs when I tried it last time (as I wrote, it even works for Pimcore's internal tabs like versions, properties etc.) but anyway I will move the code and test it with all layout combinations I can imagine.

brusch:
@BlackbitNeueMedien Thank you very much for your PR. 
I've pushed a slightly different implementation with 3a22e96fba99ac7e76156f9ebb9ca427838db0e1 and also considered it for the quick search.

PR #3706, opened 2018-12-05
Fix: OwnerAwareField shouldn't save Owner

### LINKED ISSUE #3543: Link null after update to 5.5.3
# Bug Report
 
 
### Expected behavior 
  
Working links in objects after update from 5.2.1 (219) to 5.5.3

### Actual behavior  
Links in objects are NULL after an update from 5.2.1 to 5.5.3

Working link in 5.2.1 looks like this:

`O:34:"Pimcore\Model\DataObject\Data\Link":14:{s:4:"text";s:8:"LocalFoo";s:12:"internalType";s:8:"document";s:8:"internal";i:1;s:6:"direct";N;s:8:"linktype";s:8:"internal";s:6:"target";s:6:"_blank";s:10:"parameters";s:0:"";s:6:"anchor";s:0:"";s:5:"title";s:8:"LocalFoo";s:9:"accesskey";s:0:"";s:3:"rel";s:0:"";s:8:"tabindex";s:0:"";s:5:"class";s:0:"";s:10:"attributes";s:0:"";}
  `

in 5.5.3

`O:34:"Pimcore\Model\DataObject\Data\Link":17:{s:7:"�*�text";s:8:"LocalFoo";s:15:"�*�internalType";s:8:"document";s:11:"�*�internal";i:1;s:9:"�*�direct";N;s:11:"�*�linktype";s:8:"internal";s:9:"�*�target";s:6:"_blank";s:13:"�*�parameters";s:0:"";s:9:"�*�anchor";s:0:"";s:8:"�*�title";s:8:"LocalFoo";s:12:"�*�accesskey";s:0:"";s:6:"�*�rel";s:0:"";s:11:"�*�tabindex";s:0:"";s:8:"�*�class";s:0:"";s:13:"�*�attributes";s:0:"";s:9:"�*�_owner";N;s:13:"�*�_fieldname";N;s:12:"�*�_language";N;}`

### LINKED ISSUE #3698: OwnerAwareField shouldn't save Owner

blankse:
Resolves #3698
Resolves #3543

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Block.php
+++ b/models/DataObject/ClassDefinition/Data/Block.php
@@ -210,6 +210,11 @@ class Block extends Model\DataObject\ClassDefinition\Data
                         }
                     }
                     $blockElement = new DataObject\Data\BlockElement($blockElementRaw['name'], $blockElementRaw['type'], $blockElementRaw['data']);
+
+                    if (isset($params['owner'])) {
+                        $blockElement->setOwner($params['owner'], $params['fieldname'], $params['language']);
+                    }
+
                     $items[$elementName] = $blockElement;
                 }
                 $result[] = $items;
--- a/models/DataObject/ClassDefinition/Data/Consent.php
+++ b/models/DataObject/ClassDefinition/Data/Consent.php
@@ -103,12 +103,16 @@ class Consent extends Model\DataObject\ClassDefinition\Data
     public function getDataFromResource($data, $object = null, $params = [])
     {
         if (is_array($data) && $data[$this->getName() . '__consent'] !== null) {
-            $data = new DataObject\Data\Consent($data[$this->getName() . '__consent'], $data[$this->getName() . '__note']);
+            $consent = new DataObject\Data\Consent($data[$this->getName() . '__consent'], $data[$this->getName() . '__note']);
         } else {
-            $data = new DataObject\Data\Consent();
+            $consent = new DataObject\Data\Consent();
         }
 
-        return $data;
+        if (isset($params['owner'])) {
+            $consent->setOwner($params['owner'], $params['fieldname'], $params['language']);
+        }
+
+        return $consent;
     }
 
     /**
--- a/models/DataObject/ClassDefinition/Data/EncryptedField.php
+++ b/models/DataObject/ClassDefinition/Data/EncryptedField.php
@@ -202,7 +202,13 @@ class EncryptedField extends Model\DataObject\ClassDefinition\Data
             $data = $this->decrypt($data, $object, $params);
             $data = $fd->getDataFromResource($data, $object, $params);
 
-            return new Model\DataObject\Data\EncryptedField($this->delegate, $data);
+            $field = new Model\DataObject\Data\EncryptedField($this->delegate, $data);
+
+            if (isset($params['owner'])) {
+                $field->setOwner($params['owner'], $params['fieldname'], $params['language']);
+            }
+
+            return $field;
         }
     }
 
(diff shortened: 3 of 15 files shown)


brusch:
@blankse  please let us know when you're done with testing.
Thanks!

models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -166,14 +166,14 @@ public function getDataForResource($data, $object = null, $params = [])
      * @param null|Model\DataObject\AbstractObject $object
      * @param mixed $params
      *
-     * @return string
+     * @return array|null
```
dpfaffenbauer: not true, should be mixed since it could be anything

models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -115,7 +115,7 @@ class Block extends Model\DataObject\ClassDefinition\Data
     /**
      * @see DataObject\ClassDefinition\Data::getDataForResource
      *
-     * @param string $data
+     * @param array $data
```
dpfaffenbauer: not true, should be mixed since it could be anything

models/DataObject/ClassDefinition/Data.php
```diff
@@ -133,31 +133,34 @@
...
      *
      * @param mixed $data
+     * @param null|\Pimcore\Model\DataObject\AbstractObject $object
+     * @param mixed
      *
      * @return mixed
-     *
-     * abstract public function getDataForResource($data);
      */
+    abstract public function getDataForResource($data, $object = null, $params = []);
```
dpfaffenbauer: I agree with you about that, but this shouldn't be in this PR since it doesn't belong to what to are trying to do

brusch:
@blankse I agree with @dpfaffenbauer : please separate out the things out of scope into a new PR. Thanks again!

dpfaffenbauer:
@blankse same here, but thats legacy stuff Pimcore still carries with it and it obviously needs fixing, but in a coordinated way ;)

models/DataObject/ClassDefinition/Data/Block.php
blankse: PR: #3710

blankse:
@brusch @dpfaffenbauer 

I split off thinks #3709 #3710 
I tested it tomorrow and let you know, when it is ready.

brusch:
@blankse Thanks, that looks very good and complete to me!
Nevertheless ... I'm waiting for your final go. 
@weisswurstkanone could you please also have a look at this? Just to be sure ;) Thanks!

dpfaffenbauer:
I am not sure, but why should this fix the odd characters?

blankse:
@brusch I tested it and works for me.

@dpfaffenbauer The odd characters are only in adminer. In phpmyadmin it looks ok.
The problem here is the owner field. It can be long if the object has many fields and/or many languages. The serialize string can be cut off in the table field and so the unserialize function returns false.

brusch:
@blankse alright, thanks again!

PR #3708, opened 2018-12-05
asset image: xmp data fix when file is not available in filesystem

markus-moser:
Sometimes for dev instances it's usefull to just copy the database but not the assets in the filesystem. Currently the result for images can be that it is not possible to open the asset in the pimcore backend. This fix just returns an empty array instead of a exception and therefore the asset can be opened without a problem.


brusch:
Thanks!

PR #3709, opened 2018-12-05
Add abstract functions for getDataForResource, getDataFromResource an…

blankse:
…d getDataForQueryResource

### DIFF
--- a/lib/Exception/NotImplementedException.php
+++ b/lib/Exception/NotImplementedException.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Exception;
+
+class NotImplementedException extends \Exception
+{
+    /**
+     * Generates: Intentionally not implemented: {$feature}.
+     *
+     * @param string $feature
+     * @param int $code
+     * @param \Exception|null $previous
+     */
+    public function __construct($feature, $code = 0, \Exception $previous = null)
+    {
+        parent::__construct("Intentionally not implemented: {$feature}", $code, $previous);
+    }
+}
--- a/models/DataObject/ClassDefinition/Data.php
+++ b/models/DataObject/ClassDefinition/Data.php
@@ -133,31 +133,34 @@ abstract class Data
      * Returns the the data that should be stored in the resource
      *
      * @param mixed $data
+     * @param null|\Pimcore\Model\DataObject\AbstractObject $object
+     * @param mixed $params
      *
      * @return mixed
-     *
-     * abstract public function getDataForResource($data);
      */
+    abstract public function getDataForResource($data, $object = null, $params = []);
 
     /**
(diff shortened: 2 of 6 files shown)


brusch:
Looks good thanks a lot!

brusch:
Added hint in upgrade notes (possible BC-break).

dpfaffenbauer:
Don't think thats a BC cause Pimcore expects those functions anyway.

dpfaffenbauer:
ok, it does break BC: https://travis-ci.org/coreshop/CoreShop/jobs/464235059

dpfaffenbauer:
to rather not break BC, it should be a interface and types using that should implement that interface.

blankse:
@dpfaffenbauer Yes I see, there a 2 types of ClassDefinition Data.

One with save/load/delete functions and one with getDataForResource/getDataFromResource/getDataForQueryResource.

There it should be 2 interfaces and in the Dao class it can be seperated (Not with method_exists).

But these would be another bc break...

brusch:
Yep, I think the current approach is less evil in terms of BC.

dpfaffenbauer:
true....

But: this should happen with 6.0 instead of 5.6.

@brusch IMO: this should be reverted and kept for 6.0 and also needs discussion of how DataTypes can improve in the future

brusch:
true ... reverted, see #3713
Thanks guys.

blankse:
We could create the interfaces in 5.6 and trigger a deprecated warning if the interfaces are not set. So we have a smooth transition to these interfaces. This bc layer could we remove in 6.0.

I would like create a PR for this. Have you nice names for these 2 interfaces?

dpfaffenbauer:
Yep. That would be good and BC as well ;)

Names... thats a good question.... not sure to be honest. Something Like ```ResourcePersistenceAwareInterface``` and ```QueryResourcePersistenceAwareInterface```

blankse:
I would have put getDataForResource and getDataForQueryResource in one interface. Should these seperated? Which name would be good for the save/load/delete function interface

dpfaffenbauer:
They would need to be in 2 Interfaces since they can be used separately.

```CustomResourcePersistingInterface``` or ```SelfPersistingResourceInterface``` something like that mabye?

**Comment by dpfaffenbauer** [comment]
@brusch can you help out here please? :)

**Comment by brusch** [comment]
`CustomResourcePersistingInterface` sounds good! 
Maybe `StandardResourcePersistingInterface` for the other one?

**Comment by dpfaffenbauer** [comment]
`well, but it is really `standard` and it should be two, since a type can decide if he persists into the store table, or the query table or both.

**Comment by dpfaffenbauer** [comment]
but, thinking about it. I guess it doesn't make any sense for a type to persist in one but not the other table right?

**Comment by brusch** [comment]
Yep, it should definitely persist to both in my opinion. 

So:
```php
interface CustomResourcePersistingInterface {
    public function save($object, $params = []);
    public function load($object, $params = []);
    public function delete($object, $params = []);
}

interface StandardResourcePersistingInterface {
    public function getDataForResource($data, $object = null, $params = []);
    public function getDataFromResource($data, $object = null, $params = []);
    public function getDataForQueryResource($data, $object = null, $params = []);
}
```

blankse:
I created a PR with the 3 interface approach: #3715 

It seems to give mixed data types... In Block it is a filled getDataForResource, getDataFromResource and load, but no save, delete and getDataForQueryResource

brusch:
@blankse ok, thanks so far. also the relation data-types are mixed, they have `getDataForQueryResource` and `save`/`delete`/`load` so yes, i think the 3 interface approach makes most sense. 

Regarding our special friend the block data-type: i guess the only practical solution is to give im also the `CustomResourcePersistingInterface` and implement empty `save()`  and `delete()` methods. 

Therefore we also have to change the if/else logic on save/delete/load to allow both methods in parallel, like here: https://github.com/pimcore/pimcore/blob/40d32939a109319a46db9149d0ebae53c1e2a184/models/DataObject/Concrete/Dao.php#L220

PR #3710, opened 2018-12-05
Fix: PHP Doc Types

blankse:

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Block.php
+++ b/models/DataObject/ClassDefinition/Data/Block.php
@@ -115,7 +115,7 @@ class Block extends Model\DataObject\ClassDefinition\Data
     /**
      * @see DataObject\ClassDefinition\Data::getDataForResource
      *
-     * @param string $data
+     * @param array $data
      * @param null|Model\DataObject\AbstractObject $object
      * @param mixed $params
      *
@@ -166,14 +166,14 @@ class Block extends Model\DataObject\ClassDefinition\Data
      * @param null|Model\DataObject\AbstractObject $object
      * @param mixed $params
      *
-     * @return string
+     * @return array|null
      */
     public function getDataFromResource($data, $object = null, $params = [])
     {
         if ($data) {
             $count = 0;
 
-            $unserializedData = unserialize($data);
+            $unserializedData = Serialize::unserialize($data);
             $result = [];
 
             foreach ($unserializedData as $blockElements) {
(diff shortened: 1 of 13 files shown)


models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -166,14 +166,14 @@ public function getDataForResource($data, $object = null, $params = [])
      * @param null|Model\DataObject\AbstractObject $object
      * @param mixed $params
      *
-     * @return string
+     * @return array|null
```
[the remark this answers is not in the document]
brusch: @blankse yep

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -64,11 +64,7 @@ class Link extends Model\DataObject\ClassDefinition\Data
      */
     public function getDataForResource($data, $object = null, $params = [])
     {
-        if ($data instanceof DataObject\Data\Link and isset($data->object)) {
-            unset($data->object);
-        }
-
-        if ($data) {
+        if ($data instanceof DataObject\Data\Link) {
```
brusch: @blankse I guess that's a leftover of #3706 - that shouldn't change, because it would break things here.

models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -115,7 +115,7 @@ class Block extends Model\DataObject\ClassDefinition\Data
     /**
      * @see DataObject\ClassDefinition\Data::getDataForResource
      *
-     * @param string $data
+     * @param array $data
```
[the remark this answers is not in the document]
dpfaffenbauer: 👍

models/DataObject/ClassDefinition/Data/Link.php
blankse: @brusch The `isset($data->object)` makes no sense. There is neither a public object attribute nor a __isset function.

Also should check if is a Link object, because the method calls afterwards will fail.

What did you think will break?

dpfaffenbauer: There is a setObject and getObject function but no property for that.

blankse: Oh, I see. I will make another PR for this :sweat_smile:

blankse: The setObject and getObject don't use the object attribute anymore (https://github.com/pimcore/pimcore/commit/a67d648e7ef0374e74a9c2b1e2046d92d7e437d8#diff-a67f6fdbaa4e6437319bf7fa07ac1ae4)

brusch:
> I rebased, because of merging #3706

Thanks, just wanted to ask you to do so ;)

PR #3711, opened 2018-12-06
Creates thumbnail if not exists

strati1991:
When merging thumbnails from another branch it is usually really hard to resolve the merge conflicts that occur in the thumbnail configuration file.

When resolving the merge conflict by taking the other branches changes your changes get deleted from the thumbnail configuration. After this it is neither possible to save your changes anymore because the thumbnail doesn't exist in the config and the thumbnailUpdateAction throws a null pointer exception.

When adding the creation of the thumbnail on not existence to the update method this problem is solved and the thumbnail can be saved again.


brusch:
Hmm, but shouldn't you merge the changes from the other branches instead of just wiping your changes? 
Your PR assumes that you have the thumbnail config opened in the admin UI, while performing a merge, and then save your changes again.
Imho that's not really a good practice.

brusch:
To be honest, I'm not really happy with that. Any other opinions?

strati1991:
Is there a reason why the thumbnail configuration is stored as a config file and not in the database?

dpfaffenbauer:
@strati1991 Yes, cause they belong to the project and get deployed as well and are not environment specific.

brusch:
@strati1991 what I don't understand is why you feel merging is so hard ... should be actually quite easy, because the definitions are self-contained ...

brusch:
I'm closing this PR for now, but please feel free to continue the discussion.

PR #3715, opened 2018-12-06
Interfaces for Data Types

blankse:
This PR create interfaces for data types with BC layer


dpfaffenbauer:
re: https://github.com/pimcore/pimcore/pull/3709#issuecomment-444907135

We have following options:
 - Implement an empty save functions, but I am not sure if that will cause other issues, cause Pimcore might not call its "standard" save then
 - Add a separate interfaces eg: `CustomResourcePersistingInterface` and `CustomResourceLoadingInterface`, `CustomResourceDeleteInterface`

brusch:
I post it here as well :) 

@blankse ok, thanks so far. also the relation data-types are mixed, they have `getDataForQueryResource` and `save`/`delete`/`load` so yes, i think the 3 interface approach makes most sense. 

Regarding our special friend the block data-type: i guess the only practical solution is to give im also the `CustomResourcePersistingInterface` and implement empty `save()`  and `delete()` methods. 

Therefore we also have to change the if/else logic on save/delete/load to allow both methods in parallel, like here: https://github.com/pimcore/pimcore/blob/40d32939a109319a46db9149d0ebae53c1e2a184/models/DataObject/Concrete/Dao.php#L220

blankse:
I implemented the interfaces for the Block data type.
Is this the right direction?
Can I implement it so for all types?

brusch:
@blankse let me know when you're done :) Thanks in advance!

brusch:
@blankse are you done?

brusch:
@blankse I think we should move the interface to `DataObject\Data\` can you move them please? 
Thanks!

blankse:
@brusch I test it, than its done.

I think we should move it to `DataObject\ClassDefinition\Data\` or?

brusch:
@blankse yep, sorry, of course `DataObject\ClassDefinition\Data\` :) 
Thanks! Everything else looks ok for me so far. 

We have to keep in mind that this PR also breaks BC in some edge-cases, eg. when a custom data-type extends from a core type but overrides e.g. `delete()` which isn't compatible to the new interface (missing `$params` argument). But I think that's reasonable, but we should put a note on that here: 
https://github.com/pimcore/pimcore/blob/f9a753a8d26f29463555bac264af1d7a99a9bd4d/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/01_Within_V5/README.md#L6

blankse:
@brusch This works good. I have some problems with Nonownerobjects. But this is out of scope. I will handle it here #3728

dpfaffenbauer:
https://travis-ci.org/coreshop/CoreShop/jobs/466090824#L1381
https://travis-ci.org/coreshop/CoreShop/jobs/466090824#L1404
https://travis-ci.org/coreshop/CoreShop/jobs/466090824#L1447

Pimcore's own types raise a deprecation warning?

blankse:
@dpfaffenbauer You are right. I fix it #3734

PR #3716, opened 2018-12-06
Fix PhpDoc parameter type of ICart::removeVoucherToken()

jdreesen:
The implementation of the interface has already the correct type hint in its PhpDoc:
https://github.com/pimcore/pimcore/blob/8b3f023feb8ee06731d2aa0b65dd95d5e30085b2/bundles/EcommerceFrameworkBundle/CartManager/AbstractCart.php#L853


brusch:
Thanks!

PR #3728, opened 2018-12-08
Rename Nonownerobjects to ReverseManyToManyObjectRelation

blankse:
The Objects datatype is renamed #3653 but the Nonownerobjects not.
So I think it's inconsistency.


blankse:
Without this changes the field doesn't work anymore:

No field is selectable because it filter `objects` and not `manyToManyObjectRelation`:
https://github.com/pimcore/pimcore/blob/88c56820b53623d4ba084d7f24a5ad69d6de2000/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/nonownerobjects.js#L107-L124

No data object can be opened with this field, because the setter is not found:
https://github.com/pimcore/pimcore/blob/88c56820b53623d4ba084d7f24a5ad69d6de2000/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L558-L560

blankse:
@brusch Ok I will look into it ;) But I think its a new PR.

blankse:
I restored the getDataForEditmode function before #3614 for NonOwnerManyToManyObjectRelation. So the path will be shown again.

dpfaffenbauer:
Shouldn't it be called `ReverseOwnerManyToMany`? Or `InversedManyToMany`?

brusch:
@dpfaffenbauer I'm also not really happy with the name "non-owner"
I'd prefer `ReverseOwnerManyToMany` or `ForeignOwnerManyToMany`. 
What do you guys think? 

@blankse Thanks!

dpfaffenbauer:
lets keep a similar naming strategy as doctrine does (which the other types no follow as well eg. ManyToMany and OneToMany and so). So yes, 'ReverseManyToMany'?

blankse:
I like the Reverse prefix. `ReverseManyToManyObjectRelation`?

brusch:
Thanks a lot @blankse

PR #3730, opened 2018-12-10
Fix Symfony Deprecations

blankse:
Fix following Deprecations:
`Using a colon after an unquoted mapping key that is not followed by an indication character (i.e. " ", ",", "[", "]", "{", "}") is deprecated since Symfony 3.2 and will throw a ParseException in 4.0`


brusch:
Thanks!

Issue #3737, opened 2018-12-11
[Data Objects] Many-To-Many Object Relation Performance Issues in Admin UI


brusch:
The problem is that the data-type doesn't utilize `getRelationData()` anymore to obtain the relational data, instead all objects are loaded and `getDataForEditmode()` is called. 

See: https://github.com/pimcore/pimcore/blob/cf8186c048650146a825b07ce91f02953282437c/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L554

weisswurstkanone:
are you referring to this commit ?
https://github.com/pimcore/pimcore/commit/2b99cea2f0f39e42c79681aec81ccd1e42c5d699

weisswurstkanone:
temporary fix: use the old strategy if no visible fields are selected.
long-term solution: lazy load the metadata (just like the path formatter does)

blankse:
I don't test it, but after looking in the fix i think it is not enough. The getter method will also be executed. Am I wrong?

weisswurstkanone:
when, where ?

blankse:
https://github.com/pimcore/pimcore/blob/dd9d7a856bba6621c51a514d169d28ac76fcb7f9/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L566-L568

weisswurstkanone:
good catch

fashxp:
still relevant?

weisswurstkanone:
Don't think so. Just enable async loading. 

see https://github.com/pimcore/pimcore/commit/62204339d7c582612596e877b52f29fe834790ac

![image](https://user-images.githubusercontent.com/6142086/71171623-844b0280-225e-11ea-8b7e-f17e13e7c017.png)

PR #3739, opened 2018-12-11
VisibleFields Setting for ReverseManyToManyObjectRelation

blankse:
Visible Fields Feature from #3614 for ReverseManyToManyObjectRelation

### DIFF
--- a/models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php
+++ b/models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php
@@ -158,7 +158,9 @@ class AdvancedManyToManyRelation extends ManyToManyObjectRelation
     }
 
     /**
-     * @param $data
+     * @see QueryResourcePersistenceAwareInterface::getDataForQueryResource
+     *
+     * @param array $data
      * @param null|Model\DataObject\AbstractObject $object
      * @param mixed $params
      *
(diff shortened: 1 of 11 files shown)


models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php
```diff
@@ -24,7 +24,7 @@
 use Pimcore\Model\Document;
 use Pimcore\Model\Element;
 
-class AdvancedManyToManyRelation extends ManyToManyObjectRelation implements CustomResourcePersistingInterface, QueryResourcePersistenceAwareInterface
```
brusch: That was already fixed in 2f7e18f5e76d31e657a3cb83f39e6b0cff04c1bb 
Can you please take that away? (also for `AdvancedManyToManyObjectRelation` 
Thanks!

blankse: I rebased to master

brusch:
This PR causes the same issues as described in #3737 so we have to solve that first, so that we can use the same approach for reverse many-to-many as well.

brusch:
Closed for now. 
Needs to follow similar approach as used here  #4132 
After this PR was merged we can think about how to implement this for reverse relations.

PR #3743, opened 2018-12-12
[Email] Refactored email log to use the same default address format

brusch:


bundles/AdminBundle/Controller/Admin/EmailController.php
```diff
@@ -271,22 +271,40 @@ public function resendEmailAction(Request $request)
                 $mail->setBodyText($text);
             }
 
-            $mail->setFrom($emailLog->getFrom());
+            $fromArray = \Pimcore\Helper\Mail::parseEmailAddressField($emailLog->getFrom());
```
dpfaffenbauer: you are repeating yourself 5 times here, better to refactor that into an array?

brusch:
@dvesh3 especially the migration script would need additional reviewing/testing. Thanks!

PR #3747, opened 2018-12-13
Block data-type needs no query column

### LINKED ISSUE #3746: Cannot create object with Block-Datatype
See screenshot. Just tested on the latest demo

image

blankse:
Resolves #3746

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Block.php
+++ b/models/DataObject/ClassDefinition/Data/Block.php
@@ -66,13 +66,6 @@ class Block extends Data implements CustomResourcePersistingInterface, ResourceP
     public $maxItems;
 
     /**
-     * Type for the column to query
-     *
-     * @var string
-     */
-    public $queryColumnType = 'longtext';
-
-    /**
      * Type for the column
      *
      * @var string


weisswurstkanone:
thx!

brusch:
This fix is unfortunately not entirely right, tried to fix it the proper way with 51f47bda2107262a27f158716e00e3f419caa4d9

blankse:
Isn't it a bc break?

brusch:
Well, kind of ... but to be realistic, who does override/call useless functions on data-types?

dpfaffenbauer:
@brusch true, lets wait how my tests react 😜

blankse:
I mean: Every non core data type must implement these traits now. Because the data abstract class don't have these function anymore. Is this not a bc break?

brusch:
@blankse that's true ... maybe not such a good idea.
At least we shouldn't use them in the BC layer to determine whether to call `getDataForQueryResource()` or not :) 

see: https://github.com/pimcore/pimcore/commit/51f47bda2107262a27f158716e00e3f419caa4d9#diff-16f9a3cbf890a5bd4d3ba500c90e2478R255

brusch:
Done in 71729e5a6475c2f229316f8349e16579a3230368

blankse:
@brusch We should set the function direct in the abstract class and trigger a deprecation warning. So the custom data type owner can migrate to the new traits.

brusch:
@blankse done in 0c0fd117d82cddc7585fc2ef755fa94ad0a190d9

PR #3748, opened 2018-12-13
Fix: Unable to open objects with ManyToManyObjectRelation fields and visible fields but no classes

blankse:
# Steps to reproduce
1. Create class with Many-To-Many Object Relation
2. Select Many-To-Many Object Relation classes and visible fields
3. Save class
4. Create object
4. Remove Many-To-Many Object Relation classes
5. Save class
6. Open object


brusch:
Thanks!

PR #3749, opened 2018-12-13
add version preview for metadata fields of advanced multi relation types

markus-moser:
This pull request adds metadata columns of advanced multi relation data types to the data object version preview.

Styling is not that easy but should be acceptable:
image


brusch:
Thanks!

PR #3751, opened 2018-12-13
ElasticSerach  - add setter for searchAggregations

ckogler-elements:
Re add setter for custom searchAggregations


ckogler-elements:
Unwanted commits in it  - creating a new pr

PR #3758, opened 2018-12-17
Fixed usage of wrong variable in webservice import of videos

BlackbitNeueMedien:
The `getFromWebserviceImport` method for videos referenced the wrong variable in some places. I have replaced the variables in this PR.


weisswurstkanone:
thx!

PR #3759, opened 2018-12-17
Exception when copying an object recursive with updating references without children

daves-weblab:
When copying an object without children via "recursive, updating references" an exception is raised. This is due to the variable $childIds not being countable (not existent without children). Thus the for loop iterating over the children should only be accessible if there are children present.


weisswurstkanone:
thx!

Issue #3762, opened 2018-12-17
Imagegallery items missing context on upload in backend


gadget60:
Images in an imagegallery have no context (that is, having object id and fieldname set) when uploading an image in Pimcore backend.
This breaks dynamic setting the upload folder as described in  https://pimcore.com/docs/5.x/Development_Documentation/Extending_Pimcore/Event_API_and_Event_Manager.html (chapter "Asset Upload Path") when using imagegalleries.

Reason is: imagegallery is a container for on-the-fly generated hotspotimages, but their context won't be set. PR follows...

PR #3763, opened 2018-12-17
Fix setting the context for imagegallery items

gadget60:
Fixes https://github.com/pimcore/pimcore/issues/3762


weisswurstkanone:
thx!

Issue #3242, opened 2018-08-21
Consider using Hoa\Compiler for Placeholder instead of regex


dkarlovi:
# Feature Request 
- [x] There is no existing issue regarding the same topic

## Feature description 
  
 Currently, the [`Placeholder` uses regex](https://github.com/pimcore/pimcore/blob/132933484df724e6d31f837efab7bdc07c6ffca2/lib/Placeholder.php#L185-L186) to match individual placeholders and their configuration. While this is simple, it's also quite fragile to get wrong and run into problems you can't easily debug if you're an end user (or even a developer).

Using some dedicated component like [`Hoa\Compiler`](https://github.com/hoaproject/Compiler) there exists an opportunity to define the grammar in a much more robust way, together with strict validation and interpretation. It might provide additional opportunities in the future.

brusch:
Closed in favor of #7215

PR #3766, opened 2018-12-17
Add AppBundle placeholder path

ckogler-elements:


dpfaffenbauer:
you should refactor this and make this available through the container. adding a new path makes this now more complicated to refactor -> since it will be a BC break.

dpfaffenbauer:
And, since you are touching this, also consider open tickets like:

https://github.com/pimcore/pimcore/issues/3242

ckogler-elements:
@fashxp updated the docs.
@dpfaffenbauer : I think it makes  sence to add the AppBundle path as it would be the default location when you migrate from Pimocre 4 to 5.

fashxp:
PRs are welcome ;-)
for now this is better than it was before, therefore we merge it. Thx.

PR #3769, opened 2018-12-18
insertOrUpdateVoucherSeries() - return generated codes

ckogler-elements:
insertOrUpdateVoucherSeries() should return the generated codes if successfully (like it was in older versions)

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Controller/VoucherController.php
+++ b/bundles/EcommerceFrameworkBundle/Controller/VoucherController.php
@@ -151,10 +151,10 @@ class VoucherController extends FrontendController
                 $translator = $this->get('translator');
                 $params = ['id' => $request->get('id')]; //$request->query->all();
 
-                if ($result == true) {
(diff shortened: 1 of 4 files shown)


fashxp:
did you double check the problems of https://github.com/pimcore/pimcore/pull/3500?

ultramegatom:
@fashxp @ckogler-elements 
i think we would need to reverse this condition: 
https://github.com/pimcore/pimcore/blob/master/bundles/EcommerceFrameworkBundle/Controller/VoucherController.php#L154
from ```$result == true``` to ```$result === false```

and also make sure the abstractTokenManager / interface reflect the return type in the phpdoc.

ultramegatom:
@fashxp @ckogler-elements 
it really feels unconsistend to me however... Almost seems to me like this method should not be public and only be used internally. 
Maybe the SingleVoucher should return the code as well then? Then we would have a ```string|array|bool``` return type...


Maybe we could introduce a more cleaner way to generate tokens programatically? Like make a new method that generates and returns codes based on parameters instead of relying on the voucher configuration. That way we would have full control over it.

fashxp:
thx

PR #3770, opened 2018-12-18
refactor optimizers to use symfony container

dpfaffenbauer:
The reason for this should be quite clear, you can now register custom optimizers or change existing ones.


dpfaffenbauer:
BTW: having something like a service registry in the Pimcore Core would make this things much easier:

https://github.com/coreshop/Registry/blob/master/ServiceRegistry.php

adding a new compiler pass for a standard way of registering services, makes that far easier, more painless and consistent through the system:

https://github.com/coreshop/CoreShop/blob/b3f85464e9f9e77728ae5e4e704844b361bb3c8b/src/CoreShop/Bundle/CoreBundle/DependencyInjection/Compiler/RegisterReportsPass.php#L17

bundles/CoreBundle/Resources/config/commands.yml
```diff
@@ -0,0 +1,6 @@
+services:
+    Pimcore\Bundle\CoreBundle\Command\ThumbnailsOptimizeImagesCommand:
```
brusch: That shouldn't be necessary because it's autoregistered and autowired, right? Or any other reason for that?

bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
```diff
@@ -105,6 +105,8 @@ public function loadInternal(array $config, ContainerBuilder $container)
         $loader->load('analytics.yml');
         $loader->load('sitemaps.yml');
         $loader->load('aliases.yml');
+        $loader->load('optimizer.yml');
```
brusch: We should call this file `image_optimizers.yml`, to be more specific.

bundles/CoreBundle/PimcoreCoreBundle.php
```diff
@@ -82,5 +83,6 @@ public function build(ContainerBuilder $container)
         $container->addCompilerPass(new DebugStopwatchPass());
         $container->addCompilerPass(new LongRunningHelperPass());
         $container->addCompilerPass(new WorkflowPass());
+        $container->addCompilerPass(new RegisterOptimizersPass());
```
brusch: Here also `RegisterImageOptimizersPass`

bundles/CoreBundle/Resources/config/commands.yml
dpfaffenbauer: well, it would be autoconfigured/autowired if the path is configured, but Pimcore doesn't do that. Symfony does automatically recognize commands, but thats deprecated and they should be registered using the container starting with 4.0

bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
dpfaffenbauer: will do

bundles/CoreBundle/PimcoreCoreBundle.php
dpfaffenbauer: will do

bundles/CoreBundle/Resources/config/commands.yml
brusch: Ok thanks, I wasn't aware of the deprecation in 4.0 :) 
But still auto-registering works, since it's located in the standard namespace `BUNDLE\Command`, see https://symfony.com/doc/3.4/console.html#registering-the-command

I'm just thinking if we should leave this out of this PR for now and do the manual service definition for all Pimcore commands in a separate PR, so that we don't get into a mess (some commands defined, some not, ...)

bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
dpfaffenbauer: done

bundles/CoreBundle/PimcoreCoreBundle.php
dpfaffenbauer: done

bundles/CoreBundle/Resources/config/commands.yml
dpfaffenbauer: I don't 100% agree with you. If somebody touches a command, why not just go ahead and refactor it? You are right that all should be done, but somehow we need to start with it. and that is a start ;)

PR #3773, opened 2018-12-19
Added batch edit option for Advanced Many-To-Many Object Relation meta fields

### LINKED ISSUE #3674: Relation - Object Meta Data - Bulk /Batch Update , Filter Option
# Feature Request 
Relation - Object Meta Data - Bulk /Batch Update , Filter Option
## Please make sure all the following boxes are checked before submitting your feature request - thank you!
- [ ] There is no existing issue regarding the same topic

## Feature description 
Pimcore should  provide bulk update and filter as an option in admin for all the metadata fields as one can have more than 1k relations with the object.
screenshot from 2018-11-29 11-05-57

dvesh3:
Resolves #3674 

Batch edit option for meta fields in Advanced Many-To-Many Object Relation
batchupdate1
batchupdate2

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
@@ -133,7 +133,7 @@ pimcore.object.tags.advancedManyToManyObjectRelation = Class.create(pimcore.obje
 
             var cellEditor = null;
             var renderer = null;
-            var listeners = null;
+            var listeners = [];
 
             if (this.fieldConfig.columns[i].type == "number" && !readOnly) {
                 cellEditor = function() {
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
```diff
@@ -190,9 +190,9 @@ pimcore.object.tags.advancedManyToManyObjectRelation = Class.create(pimcore.obje
                     }
                 };
 
-                listeners = {
-                    "mousedown": this.cellMousedown.bind(this, this.fieldConfig.columns[i].key, this.fieldConfig.columns[i].type)
-                };
+                listeners = [
```
fashxp: does that change to an array of objects really work? 
docs say, listeners should be an object (not an array of objects). should we stick to documented ways of defining listeners?

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
```diff
@@ -230,6 +230,36 @@ pimcore.object.tags.advancedManyToManyObjectRelation = Class.create(pimcore.obje
 
 
         if (!readOnly) {
+            listeners.push({'afterrender':
```
fashxp: this is the wrong place for that, isn't it? 
because it modifies the listeners array, that is created, populated and assigned to the column config in the for-loop above. 
so, this event listener will only be added to the last column, not to all!

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
dvesh3: strange but it works! reverted as per your suggestion :-)

PR #3774, opened 2018-12-19
[UI Improvement] Batch edit & append UI changes

### LINKED ISSUE #3750: [UI Improvement] Better differenciation between batch edit all and selected
Hello,

Some pimcore users, using backend for production websites, found the batch edit command "risky" because sometimes they could use the wrong action (as they are side by side actions) and the context is not well defined in the confirmation popin.

I think two small UI improvements can help this.

1- Having a different icons for both action to prevent click on the wrong one

icones

2- Adapt the popin title to give the context. This one is the more important i think.

title

I link this issue to another one i opened some days ago which is on the same subject/scope (better information for batch process): https://github.com/pimcore/pimcore/issues/3666

Thanks.

dvesh3:
Resolves #3750


bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js
```diff
@@ -239,11 +239,13 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
                         if(formPanel.isValid()) {
                             this.batchProcess(jobs, append, editor, fieldInfo, true);
                         }
-                    }.bind(this)
+                    }.bind(this),
```
fashxp: was this added on purpose?

Issue #3778, opened 2018-12-20
[DataObjects] Multiple Deprecation Warnings


fashxp:
```
PHP Deprecated: Pimcore\Model\DataObject\ClassDefinition\Data\Textarea uses method Pimcore\Model\DataObject\ClassDefinition\Data::setColumnType from the abstract class. This won't work in v6.0, please use the proper interfaces and provided traits. in /home/ecommerce/demo-ecommerce/dev/pimcore/pimcore/models/DataObject/ClassDefinition/Data.php on line 1389
PHP Deprecated: Pimcore\Model\DataObject\ClassDefinition\Data\ManyToOneRelation uses method Pimcore\Model\DataObject\ClassDefinition\Data::setQueryColumnType from the abstract class. This won't work in v6.0, please use the proper interfaces and provided traits. in /home/ecommerce/demo-ecommerce/dev/pimcore/pimcore/models/DataObject/ClassDefinition/Data.php on line 1389
PHP Deprecated: Pimcore\Model\DataObject\ClassDefinition\Data\ManyToOneRelation uses method Pimcore\Model\DataObject\ClassDefinition\Data::setColumnType from the abstract class. This won't work in v6.0, please use the proper interfaces and provided traits. in /home/ecommerce/demo-ecommerce/dev/pimcore/pimcore/models/DataObject/ClassDefinition/Data.php on line 1389
PHP Deprecated: Pimcore\Model\DataObject\ClassDefinition\Data\ManyToManyObjectRelation uses method Pimcore\Model\DataObject\ClassDefinition\Data::setQueryColumnType from the abstract class. This won't work in v6.0, please use the proper interfaces and provided traits. in /home/ecommerce/demo-ecommerce/dev/pimcore/pimcore/models/DataObject/ClassDefinition/Data.php on line 1389
PHP Deprecated: Pimcore\Model\DataObject\ClassDefinition\Data\ManyToManyObjectRelation uses method Pimcore\Model\DataObject\ClassDefinition\Data::setColumnType from the abstract class. This won't work in v6.0, please use the proper interfaces and provided traits. in /home/ecommerce/demo-ecommerce/dev/pimcore/pimcore/models/DataObject/ClassDefinition/Data.php on line 1389
[...]

PR #3783, opened 2018-12-21
[DataObjects] Fix: No deprecation warning if property exists

blankse:
The deprecation warning should also be generated with existing properties


brusch:
The getters and setters should be removed entirely from the abstract class in future, so just moving the deprication warning is not the final solution.

brusch:
@blankse sorry, I thought this is also related to #3778 :) Thanks!

PR #3802, opened 2019-01-04
Fix Voucher FIeldcollection charactertype and length settings

NiklasBr:
As is currently the case the settings for this field allows the user to leave it as-is, causing `Pattern::generateCode()` to throw a fatal error and fail. The same goes for Length's `minValue` where a negative (why anyone would ever have a negative length I don't know) will cause a fatal error to be thrown. The length value needs to be `prefix+1` and prefix can be empty, so a value of  `1` should be a sane and usable default.

  
  

Resolves #

## Additional info  

Reproduce this issue by:

1. Create a new VoucherSeries,
2. Add a pattern, but leave the Length and/or Character Type fields empty (this is allowed in the interface)
3. Switch to the Voucher Details tab
4. Attempt to generate


fashxp:
thx very much ... but if, then I would go even further ... datatype can be unsigned integer and the same goes for the count attribute too. 

I changed that.

PR #3819, opened 2019-01-09
Fixes incorrect return methods in PHPDocs

NiklasBr:
Resolves Documentation vs code mismatches

### DIFF
--- a/models/DataObject/Listing.php
+++ b/models/DataObject/Listing.php
@@ -82,7 +82,7 @@ class Listing extends Model\Listing\AbstractListing implements \Zend_Paginator_A
     /**
      * @param $unpublished
      *
-     * @return bool
+     * @return $this
      */
     public function setUnpublished($unpublished)
     {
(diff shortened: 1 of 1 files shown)


models/DataObject/Listing.php
```diff
@@ -178,7 +178,7 @@ public function count()
      * @param int $offset
      * @param int $itemCountPerPage
      *
-     * @return array
+     * @return int
```
blankse: I think `@return array` was correct

brusch: Yep, count should return `int`

NiklasBr: In that `case @method int load()` should be `@method array load()`.

I assumed the higher order one was the correct one.

NiklasBr: (line 26 by the way)

blankse: In Line 26 is:
https://github.com/pimcore/pimcore/blob/50c71ce7ff1cc75caff5b427b8a09cdb373b434b/models/DataObject/Listing.php#L26

You can change it here also to `@return Model\DataObject[]`

models/DataObject/Listing.php
```diff
@@ -228,7 +228,7 @@ public function key()
     }
 
     /**
-     * @return mixed|void
+     * @return mixed|null
```
blankse: I think `@return mixed` should be enough. `null` is impossible. The `next` function will return `false` if the array is at the end.

PR #3825, opened 2019-01-11
Improves bookmarklet markup

NiklasBr:
See original at https://pimcore.com/docs/5.x/Development_Documentation/Development_Tools_and_Details/Cache/Full_Page_Cache.html#page_Bookmarklet

  
  

Resolves #


fashxp:
actually we want to have that as links, not as JS code sample. 
I changed to a <a> tag - this should do the trick that in docs it is visualized correctly.

PR #3828, opened 2019-01-11
Enrich object layout before preSendData event

jremmurd:
### Description
The field definitions of the keys for classification stores get enriched after the `pimcore.admin.dataobject.get.preSendData` is sent. So manipulating the field definitions of the classification store's keys is not possible yet.

### Use Case
I want to set a specific field of a single key in a classification store group to `notEditable`.


weisswurstkanone:
thx!

PR #3830, opened 2019-01-14
admin/website translations backend - sort by language fix

markus-moser:
Sorting by language value did not work in admin/website translations - with this fix it works.


brusch:
Thanks!

PR #3844, opened 2019-01-17
fix workflow place issue

iventura:
if workflow place would be changed from admin area, all changes from the last version including publish status will be merge into dataobject. In this case the live product will be away from the shop. this fix avoid it.

### DIFF
--- a/lib/Workflow/Manager.php
+++ b/lib/Workflow/Manager.php
@@ -243,7 +243,10 @@ class Manager
         $this->notesSubscriber->setAdditionalData([]);
 
         if ($saveSubject && $subject instanceof AbstractElement && method_exists($subject, 'save')) {
-            $subject->save();
+            if($subject->getPublished())
+                $subject->save();
+            else
+                $subject->saveVersion();
         }
 
         return $marking;


brusch:
@iventura please sign the CLA :) Thanks!

iventura:
@brusch i did it twice already, but on this page it's still on pending. Maybe i have to to something more?

brusch:
Please ensure that your email address used here on GitHub (in your profile) is the same as the one you use in your local Git config.
You can check that with `git config user.email`

brusch:
That looks good now!

ivan-vetoskin-racechip:
i tried my best :) thank you in advance!

lib/Workflow/Manager.php
```diff
@@ -243,7 +243,10 @@ public function applyWithAdditionalData(Workflow $workflow, $subject, string $tr
         $this->notesSubscriber->setAdditionalData([]);
 
         if ($saveSubject && $subject instanceof AbstractElement && method_exists($subject, 'save')) {
-            $subject->save();
+            if($subject->getPublished())
```
fashxp: this will cause problems with workflows on assets, doesn't it?

Issue #3842, opened 2019-01-16
Programatically adding items to fieldcollections is not working with Many-To-Many relationships


BlackbitNeueMedien:
I am adding entries to a fieldcollection via the PHP API. The fieldcollection contains a Many-to-many object relationship. When I add items to that field, they are not properly saved. New entries show up in the fieldcollection, but they are all empty. I am using this code:

```
$object = \Pimcore\Model\DataObject\Concrete::getById(10);
$relatedObject = \Pimcore\Model\DataObject\Concrete::getById(123);

/** @var \Pimcore\Model\DataObject\Fieldcollection $fc */
$fc = $object->getFieldcollection();

$data = new \Pimcore\Model\DataObject\Fieldcollection\Data\SomeFieldcollectionData();
$data->setRelation([$relatedObject]);

$fc->add($data);

$object->setFieldcollection($fc);

$object->save();
```

After some debugging, I found that the fieldcollection was not considered to be dirty. Since I'm using the `add` method, I figured I had to add a call to `$this->markFieldDirty('_self', true);` to that method in https://github.com/pimcore/pimcore/blob/37fe951a34de7d4d174a3f34458b05c63dc2ecea/models/DataObject/Fieldcollection.php#L156-L159
After doing so, things got weirder. I have added some logging to the code above to see the state of the MTM relation. The code used is this:

```
$printRelations = function (array $elements) {
    foreach ($elements as $element) {
        if (empty($element->getCategory())) {
            $this->io->text('Empty relation');
        } else {
            $this->io->text(count($element->getCategory()) . ' entries, first one is: ' . $element->getCategory()[0]->getFullPath());
        }
    }
};
```

It's called after initially loading the object, right after saving the object, and lastly after reloading the object. This is the result.

```
Initial state
----------

 /some/other/object/that/was/there/before

After save
----------

 /some/other/object/that/was/there/before
 /my/new/object/i/want/to/add

Object loaded again from DB
---------------------------

 Empty relation
 /my/new/object/i/want/to/add
```

So, initially the fieldcollection had already one entry. The MTM relation of that entry already has some data. After saving, it looks good at first. The collection now has two entries, the second entry is the one I just added. But after reloading the object, the MTM relation of the first entry is empty. Repeating this import will eventually leads to only empty MTM relations except for the last one:

```
Object loaded again from DB
---------------------------

 Empty relation
 Empty relation
 Empty relation
 /my/new/object/i/want/to/add

```

[...]

PR #3845, opened 2019-01-17
Bugfix: Relational items of field collections disappeared

BlackbitNeueMedien:
Solves #3842 

Field collections have not been marked dirty if an item got added or removed.
Furthermore we had the problem that the relations of existing field collection items got empty after adding a new field collection item. This is caused because when saving a field collection all relational items get deleted, see https://github.com/pimcore/pimcore/blob/master/models/DataObject/Fieldcollection/Dao.php#L40 and https://github.com/pimcore/pimcore/blob/master/models/DataObject/Fieldcollection/Dao.php#L237. For this reason they have to be recreated again but as the original field collection items and also their relations have not been touched to mark them dirty, they won't be saved again. For this reason I mark all fields of a field collection item as dirty after loading. This works but nevertheless feels a bit unclean.

### DIFF
--- a/models/DataObject/Fieldcollection/Dao.php
+++ b/models/DataObject/Fieldcollection/Dao.php
@@ -96,7 +96,7 @@ class Dao extends Model\Dao\AbstractDao
                             $collection->setValue($key, $value);
 
                             if ($collection instanceof DataObject\DirtyIndicatorInterface) {
-                                $collection->markFieldDirty($key, false);
+                                $collection->markFieldDirty($key, true);
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


models/DataObject/Fieldcollection/Dao.php
```diff
@@ -96,7 +96,7 @@ public function load(DataObject\Concrete $object)
                             $collection->setValue($key, $value);
 
                             if ($collection instanceof DataObject\DirtyIndicatorInterface) {
-                                $collection->markFieldDirty($key, false);
+                                $collection->markFieldDirty($key, true);
```
weisswurstkanone: what is the exact reason for this? after the load operation the data shouldn't be dirty, right ?

PR #3853, opened 2019-01-21
[Fix] Save default values for Checkbox and Quantity Value fields

### LINKED ISSUE #3585: Default Quantity Value does not get saved on object
# Bug Report
 
  
### Expected behavior 
If a new object is created which has a default value, the value should be saved after saving the object. Later this value should also be found on the corresponding object (DB, PHP API)
  
### Actual behavior  
All values - except for the default value - are saved. See DB. After changing the default value in the form, it will also be saved.

### Steps to reproduce
1. Create new class with Data Compontent "Quantity Value", set default value (in class definition) and new Quantity Value ("Euro")
2. Create new Object of this class
3. Edit the Object and do not change the default value!
4. Save

After that the value in the database is null.
  
### Code example 
```php
// your code
```

# Feature Request

### LINKED ISSUE #3784: Checkbox with default value checked is NULL in database
# Bug Report  
### Expected behavior 
  A object datatype checkbox with default option checked should insert 1 in the database if no user interaction takes place.

### Actual behavior  
  A object datatype checkbox with default option checked does insert NULL in the database if no user interaction takes place - same as with object datatype checkboxes which do not have a default option checked selected.
### Steps to reproduce  
Create a checkbox datatype in some object. Check the default value. 
  checkbox-default-checked
Create some object of this and save it. Value in database will be null. To get a 1, you'll need to uncheck the default checked checkbox and check it again.

Tested on Pimcore v5.5.4

prugala:
[Fix] Save default values for Checkbox and Quantity Value fields in database

resolves #3784 
resolves #3585

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/quantityValue.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/quantityValue.js
@@ -17,17 +17,6 @@ pimcore.object.tags.quantityValue = Class.create(pimcore.object.tags.abstract, {
     type: "quantityValue",
 
     initialize: function (data, fieldConfig) {
-        this.defaultValue = null;
-        this.defaultUnit = null;
-        if ((typeof data === "undefined" || data === null) && (fieldConfig.defaultValue || fieldConfig.defaultUnit)) {
-            data = {
-                value: fieldConfig.defaultValue,
-                unit: fieldConfig.defaultUnit,
-            };
-            this.defaultValue = data.value;
-            this.defaultUnit = data.unit;
-        }
-
         this.data = data;
         this.fieldConfig = fieldConfig;
 
--- a/models/DataObject/ClassDefinition/Data/Checkbox.php
+++ b/models/DataObject/ClassDefinition/Data/Checkbox.php
@@ -90,11 +90,9 @@ class Checkbox extends Data implements ResourcePersistenceAwareInterface, QueryR
      */
     public function getDataForResource($data, $object = null, $params = [])
     {
-        if (is_bool($data)) {
-            $data = (int)$data;
-        }
+        $data = is_null($data) ? $this->getDefaultValue() : $data;
 
-        return $data;
+        return (int)$data;
     }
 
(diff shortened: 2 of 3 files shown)


cwiest:
This still seems to be a problem, when quantity values with default units are used within an object brick.

cwiest:
This bugfix does not set default values or units in the initialize function, so its no wonder none are set.

prugala:
> This bugfix does not set default values or units in the initialize function, so its no wonder none are set.

Thank you. I will fix it within the next few days. Need some free time :)

dpfahlbusch:
Is there a new status of this bug?

prugala:
Sorry I had no time, I will try to fix it tomorrow.

weisswurstkanone:
IMHO, this is not the right approach.
As it says for other datatypes: "... Default values are not stored in the database table definition, nor are they set in empty Object instances. They are only available in the pimcore GUI ..."

Also see https://github.com/pimcore/pimcore/issues/4082

NiklasBr:
@weisswurstkanone FWIW I consider not saving defaults to DB an antipattern and is not a good design decision, but let's put that discussion away for the time. Even without documentation on why I will accept the the way it is as the PimCore way of doing it, there might be a good reason for that...

That said, even if default values are not actually saved to the database, they should at minimum always be returned by the getter function.

Let's assume that we create a `Product` class, and in it we add a `checkbox` called "Is Taxable", we set the default value to `true`  because >95% of all products are expected to be taxable.

When a content editor manually creates a `Product` object in PimCore they will almost always leave the "Is Taxable" checkbox checked (`true`) because they will think "*of course* it should be left checked when our products are almost always taxable".

There is a bug here, and...
* either the user interface needs to tell the user that "This value will not be saved", which is a *really bad solution*.
 * Or the default value need to be saved to the database.
 * Or the getter need to always return the default value in lieu of saved value.

However, even fixing it at the getter level is not optimal because I suspect that might actually be problematic with Listings and conditions.

I think this solution (when completed by @prugala) is good and vote for it 👍

PR #3855, opened 2019-01-21
Fixes faulty column name escaping on Date field

NiklasBr:
A missing ` would case this to not work when searching for objects with a specific date.

  
  

Resolves an error message when searching for data objects with a specific ("On") date.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Date.php
+++ b/models/DataObject/ClassDefinition/Data/Date.php
@@ -430,7 +430,7 @@ class Date extends Data implements ResourcePersistenceAwareInterface, QueryResou
             $db = Db::get();
 
             if ($this->getColumnType() == 'date') {
-                $condition = '`' . $params['name'] . ' = '. $db->quote($value);
+                $condition = '`' . $params['name'] . '` = '. $db->quote($value);
 
                 return $condition;
             } else {


brusch:
fixed it in a357e9cc705cae36782e9b417fb6761e7a210ac5 with proper `$db->quoteIdentifier()`

NiklasBr:
👍 @brusch 

I ranted a few days ago in the Gitter chat about some problems I have experienced with the documentation and I think this is a good example of what is missing. Nowhere in the documentation can I find anything about quoteIdentifier(). How was I expected to know it even exists to make a proper pull request when it was not even mentioned once?

NiklasBr:
I looked at the [lines below](https://github.com/pimcore/pimcore/commit/a357e9cc705cae36782e9b417fb6761e7a210ac5#diff-6108de5895694cf81d66e4a360a0e0cdL439) where `$filterField` was escaped with backticks...

There is no mention of Doctrine [in the docs](https://pimcore.com/docs/5.x/Development_Documentation/) how was I supposed to know that this was a third-party thing with docs on a third-party page without you mentioning it? When I said that there is a *ton* of unmentioned things in the documentation that maybe makes sense if you are a seasoned Pimcore developer but are entirely confusing for me I was not joking.

brusch:
@NiklasBr thanks for you valuable feedback - we'll take this into consideration 👍
Do you have any example where you think the docs are great?

NiklasBr:
I think some of the *Best Practices* sections are good.

Those few chapters are like "islands of self-contained knowledge" with mostly complete examples, some rules of thumbs. And they also show something I consider really valuable to understand frameworks: They contain some thinking/motivations of *why* from the writers on why some things should be like they are.

I (personally) think it is *so* much easier to learn from examples like those which are more complete as they can be than it is to learn from larger examples covering a broader topic with lots of omitted parts sacrificed for brevity. People of experience often omit parts which they think are not important, but they often are important to understand what is actually happening.

I also think the file `Doctrine/DBAL/Platforms/AbstractPlatform.php` is nicely documented. More like that 🌮

PR #3856, opened 2019-01-21
 Support locking classifcations via fielddefinition

### LINKED ISSUE #3834: Feature Request: Lock adding/removing groups from classification store
# Feature Request 
Currently it is not possible to lock a classification store on object level (i.e. field `attributes` in `Product` class) so adding/removing of groups would be disabled, but the fields inside the groups (i.e. Inhalt, Farbe) still stay editable. The current property `notEditable` inherits to all the fields in the store and also can't be overridden in the `pimcore.admin.object.get.preSendData` event. 

image

jremmurd:
# Feature Request 

Resolves #3834

## Additional info  
The PR only enables frontend layout changes i.e. via the `pimcore.admin.asset.get.preSendData` event and does NOT provide new configuration options via the class definition (maybe nice to have too) and also it does NOT include any backend permission checks.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/classificationstore.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/classificationstore.js
@@ -118,7 +118,7 @@ pimcore.object.tags.classificationstore = Class.create(pimcore.object.tags.abstr
 
         var tbarItems = [];
 
-        if (!this.fieldConfig.noteditable) {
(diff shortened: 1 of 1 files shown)


fashxp:
if we do it, then we should do it completely

fashxp:
cannot 'reuse' the `locked` setting, since this has another meaning. 

closed and see https://github.com/pimcore/pimcore/issues/4074

Issue #3848, opened 2019-01-17
Missing permissions check for Classificationstore


jremmurd:
# Bug Report
There is no permissions check in the ClassificationstoreController, thus as long as you are logged in to Pimcore you can perform all CRUD operations on the classification store definition (i.e. `createGroupAction()`, `deleteGroupAction()`, ... )

## Steps to reproduce 
>Tested in Pimcore v5.5.4
1. Log in to Pimcore without `classes` permission
2. Call any action of the Controller `Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\ClassificationstoreController`.
3. The actions will be performed even though you do not have the respective permission

PR #3857, opened 2019-01-21
Editing classifications in object context without classes permission

jremmurd:
# Bugfix 
> This PR is an addition to https://github.com/pimcore/pimcore/pull/3849
## Expected behaviour
Adding classification store groups and collections inside an object is possible without `classes` permission.

## Actual behaviour
Since resolving #3848 this is no longer possible due to missing `classes` permissions.


jremmurd:
@fashxp For explanation: the actions `"collectionsActionGet", "groupsActionGet", "addGroupsAction", "addCollectionsAction"` are used inside the objects to be able to list the groups in the popup window on adding new groups to an object. So I think it actually does not change datastructure. Further the `addGroupsAction` and `addCollectionsAction` do not perform any save until the product is saved. 

Since #3849 the add button inside any object was visible for users with `objects` permission but the popup for adding groups/collecitons resulted in an error.

PR #3858, opened 2019-01-21
Option to skip database structure on install is not working

### LINKED ISSUE #3852: Option to skip database structure on install is not working
# Bug Report
 
  
### Expected behavior 

Trying to install a new Pimcore version on an existing installation without loosing already existing DB data. The installer option `skip-database-structure` can be used, to prevent the initial deletion of the DB data. See issue #3035
  
### Actual behavior  

To reinstall Pimcore and update it to a new version, the following commands will be used:

```
composer install
./vendor/bin/pimcore-install --ignore-existing-config --skip-database-structure -n -e dev --mysql-username=xxx --mysql-password=xxx
```
Currently the `--skip-database-structure` has no impact, because in `vendor/pimcore/pimcore/bundles/InstallBundle/Command/InstallCommand.php` the method `installerNeedsOption()` will return false:

```
if ('install_options' === ($config['group'] ?? null) && InputOption::VALUE_OPTIONAL === ($config['mode'] ?? null)) {
```
In that case, this option will never added to the `$params`-Array. Because of that, the variable `$this->createDatabaseStructure` will be always `true` in `vendor/pimcore/pimcore/bundles/InstallBundle/Installer.php` in method `install()`:

```
//check skipping database creation or database data
if (array_key_exists('skip_database_structure', $params)) {
    $this->createDatabaseStructure = false;
}
``` 
The same for the other install options:
* skip-database-data
* skip-database-data-dump

 
### Steps to reproduce  

1. Add some data to an existing Pimcore installation (for example a new object)
2. Run  `./vendor/bin/pimcore-install --ignore-existing-config --skip-database-structure -n -e dev --mysql-username=xxx --mysql-password=xxx` on an existing Pimcore installation
3. Object is now deleted from the DB

mugge6:
Fix installer options:

* skip-database-structure
* skip-database-data
* skip-database-dump

  
  

Resolves #3852

## Additional info  

Comment and request from @dpfaffenbauer (https://github.com/pimcore/pimcore/issues/3852#issuecomment-456087497) is not included.

### DIFF
--- a/bundles/InstallBundle/Command/InstallCommand.php
+++ b/bundles/InstallBundle/Command/InstallCommand.php
@@ -117,20 +117,17 @@ class InstallCommand extends Command
             ],
             'skip-database-structure' => [
                 'description' => 'Skipping creation of database structure during install',
-                'mode' => InputOption::VALUE_OPTIONAL,
-                'default' => false,
+                'mode' => InputOption::VALUE_NONE,
                 'group' => 'install_options',
             ],
             'skip-database-data' => [
                 'description' => 'Skipping importing of any data into database',
-                'mode' => InputOption::VALUE_OPTIONAL,
-                'default' => false,
+                'mode' => InputOption::VALUE_NONE,
                 'group' => 'install_options',
             ],
             'skip-database-data-dump' => [
                 'description' => 'Skipping importing of provided data dumps into database (if available). Only imports needed base data.',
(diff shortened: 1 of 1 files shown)


brusch:
Thanks!

brusch:
Reverted with #3859 since it breaks the installer, needs more attention.

PR #3863, opened 2019-01-23
Add MySQL config option to set index table engine

### LINKED ISSUE #3850: [Ecommerce][IndexService] Fulltext search option to change engine to InnoDB
# Feature Request 
Option (or by default) to enable innoDB engine for product index tables.

## Feature description 
By default the createOrUpdateIndexStructures will create the index tables with engine=InnoDB. But as soon as there are fulltext search columns the table will be altered to MyISAM. 

In the installation requirements you require 'innodb fulltext search' but the fulltext search table is changed to MyISAM.

Why not remove the MyISAM alter query by default? If for a reason this is not possible, it would be nice to set the engine in the indexservice settings (yaml) or through the MySQLConfig.

Another reason to do this is because InnoDB tables are needed for database clusters to keep the tables in sync on multiple database servers.

RuudvdvNB:
Resolves #3850 

## Additional info  
Added a getter in the MySQL config for the ecommerce index service. This returns MyISAM by default but can be changed to what ever engine is needed.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/IndexService/Worker/Helper/MySql.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Worker/Helper/MySql.php
@@ -15,6 +15,7 @@
 namespace Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\Helper;
 
 use Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Config\IMysqlConfig;
+use Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Config\IMysqlEngineConfig;
 use Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Interpreter\IRelationInterpreter;
 use Pimcore\Cache;
 use Pimcore\Db\Connection;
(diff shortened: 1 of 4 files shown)


brusch:
There's the theoretical possibility of a BC break, if there are other index services implementing `IMysqlConfig`. 
Not sure if this is relevant or not.

dpfaffenbauer:
Every BC break is relevant :P

Just add a new Interface and let the implementation implement the IMysql and the new one -> no BC break then

RuudvdvNB:
Default pimcore source didn't implement the config in any other index service but ofcourse other bundles/packages could. 

Added a new interface that extends the default IMySQLConfig and implement the new interface on the DefaultMySQL index service.

bundles/EcommerceFrameworkBundle/IndexService/Worker/Helper/MySql.php
```diff
@@ -141,7 +141,7 @@ public function createOrUpdateIndexStructures()
                 Logger::info($e);
             }
 
-            $this->dbexec('ALTER TABLE `' . $this->tenantConfig->getTablename() . '` ENGINE = MyISAM;');
+            $this->dbexec('ALTER TABLE `' . $this->tenantConfig->getTablename() . '` ENGINE = ' . $this->tenantConfig->getTableEngine() . ';');
```
dpfaffenbauer: you also have to check here if the tenantConfig implements the IMysqlEngineConfig

bundles/EcommerceFrameworkBundle/IndexService/Worker/Helper/MySql.php
```diff
@@ -141,7 +141,17 @@ public function createOrUpdateIndexStructures()
                 Logger::info($e);
             }
 
-            $this->dbexec('ALTER TABLE `' . $this->tenantConfig->getTablename() . '` ENGINE = ' . $this->tenantConfig->getTableEngine() . ';');
+            $indexTableEngine = 'MyISAM';
+            try {
+                $reflectionConfig = new \ReflectionClass(get_class($this->tenantConfig));
```
dpfaffenbauer:
```php
if (this->tenantConfig instanceof IMysqlEngineConfig) {
}
```

should be enough, I don't see the need for reflection here

brusch:
I'm wondering if this whole change is even necessary or if we should just change the engine to `InnoDB`. I guess `MyISAM` is just a relic from Pimcore 4 when we supported MySQL versions < 5.6

brusch:
Changed engine to `InnoDB` - there's no reason for `MyISAM`.

PR #3870, opened 2019-01-24
Refactor how robots.txt files get persisted

dpfaffenbauer:
the current solution is unsuitable for a lot of sites with a proper deployment process

It now gets stored in a `reports.php` file in `var/config`

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/SettingsController.php
+++ b/bundles/AdminBundle/Controller/Admin/SettingsController.php
@@ -1506,25 +1506,17 @@ class SettingsController extends AdminController
     /**
      * @Route("/robots-txt", methods={"GET"})
      *
-     * @param Request $request
-     *
      * @return JsonResponse
      */
-    public function robotsTxtGetAction(Request $request)
+    public function robotsTxtGetAction()
     {
         $this->checkPermission('robots.txt');
 
-        $robotsPath = $this->getRobotsTxtPath($request);
-
-        $data = '';
-        if (is_file($robotsPath)) {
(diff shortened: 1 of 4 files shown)


dpfaffenbauer:
Currently it unsuitable to maintain a dynamic list of files in a deployment process that shares files between the release.

bundles/AdminBundle/Controller/Admin/SettingsController.php
```diff
@@ -1506,25 +1506,17 @@ public function videoThumbnailUpdateAction(Request $request)
...
-        $data = '';
-        if (is_file($robotsPath)) {
-            $data = file_get_contents($robotsPath);
-        }
+        $config = Config::getRobotsConfig();
 
         return $this->adminJson([
             'success' => true,
-            'data' => $data,
-            'onFileSystem' => file_exists(PIMCORE_WEB_ROOT . '/robots.txt')
```
brusch: Why you've removed this line? I think it's still a good thing to know if there's a robots.txt on the filesystem, that has priority over the one delivered by Pimcore. 
Could cause a lot of misunderstandings ...

dpfaffenbauer: Thought nobody uses it... Will add it again

dpfaffenbauer:
re-added on-filesystem feature, looks like this now:

**EXISTS**

bildschirmfoto 2019-01-24 um 15 53 40

**DON'T EXISTS**
bildschirmfoto 2019-01-24 um 15 53 53

PR #3873, opened 2019-01-24
Checkbox to exclude wysiwygs and textareas from the backend search index

andreas-gruenwald:
Feature: exclude wysiwyg and textarea contents from the Pimcore backend search.
Use cases:
- exclude sensitive information in search result
- exclude irrelevant data and reduce the fulltext search index size.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Textarea.php
+++ b/models/DataObject/ClassDefinition/Data/Textarea.php
@@ -54,6 +54,11 @@ class Textarea extends Data implements ResourcePersistenceAwareInterface, QueryR
     public $showCharCount;
 
     /**
+     * @var bool
+     */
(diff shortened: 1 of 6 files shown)


models/DataObject/ClassDefinition/Data/Textarea.php
```diff
@@ -243,6 +265,23 @@ public function getDiffVersionPreview($data, $object = null, $params = [])
...
+     *
+     * @param null|Model\DataObject\AbstractObject $object
+     * @param mixed $params
+     *
+     * @return string
+     */
+    public function getDataForSearchIndex($object, $params = [])
+    {
+        if ($this->isExcludeFromSearchIndex()) {
+            return "";
```
dpfaffenbauer: return null?

PR #3879, opened 2019-01-28
WebTestCase: ensure on Kernel::shutdown a container was instantiated

smoench:
If a WebTestCase with two or more test methods is used there will be a php error `Call to a member function get() on null`. After every test method the kernel will be shutdown and on every `bootKernel` will be ensured the kernel has been shutdown. Therefore it must be checked if there is a container instantiated.


brusch:
Thanks!

PR #3883, opened 2019-01-29
Fixed a bug with isValidCartItem()

ThomasKeil:
`\Pimcore\Bundle\EcommerceFrameworkBundle\CartManager\AbstractCart::isValidCartItem()` fails with an error "Call to a member function getId() on null" when trying to log a warning.
productId "0" was being handled as "unknown" in `AbstractCartItem::getProductId()`, trying to obtain the Item's object to get it's Id, which fails since the Id is not valid.
This patch fixes this.


brusch:
Thanks!

PR #3884, opened 2019-01-29
Add discernible name for link & button in cookie notice

### LINKED ISSUE #3805: discernible name for button in cookie notice
The button in cookie-policy-default-template.html should have an discernible name for a11y reasons (e.g. inner text or non-empty aria-label attribute).

It's a small code change, but an important enhancement for screen reader users. The missing discernible name gives also a penalty on https://web.dev/measure for Pimcore based websites which ranks this issue as "high impact".

Short explanation by Google's tool:

"When a button doesn't have an accessible name, screen readers announce it as "button", making it unusable for users who rely on screen readers."

Detailed explanation:
https://dequeuniversity.com/rules/axe/2.2/button-name

Thank you!

Edit: The same applies for the learn more link, of course.

dvesh3:
Resolves #3805


brusch:
Thanks!

PR #3886, opened 2019-01-29
Use proper binary KiB, MiB, ... for base 2 numbering in formatBytes()

NiklasBr:
See https://en.wikipedia.org/wiki/Kibibyte

### DIFF
--- a/lib/helper-functions.php
+++ b/lib/helper-functions.php
@@ -285,7 +285,7 @@ function return_bytes($val)
  */
 function formatBytes($bytes, $precision = 2)
 {
-    $units = ['B', 'KB', 'MB', 'GB', 'TB'];
+    $units = ['B', 'KiB', 'MiB', 'GiB', 'TiB'];
 
     $bytes = max($bytes, 0);
     $pow = floor(($bytes ? log($bytes) : 0) / log(1024));


NiklasBr:
Of course. That's because they use base 10 = MB/GB and not base 2 = MB/GiB. :)

brusch:
@NiklasBr that's actually not right, see
image (Win 10)

NiklasBr:
Apple has done it for 9 years now: https://www.apple.com/shop/question/answers/readonly/what-does-tb-mean-when-compared-to-gb/QXD9XA7YJ74X9XJJ2

NiklasBr:
Dropbox uses base 10: https://www.quora.com/Why-does-Dropbox-use-base-10-instead-of-base-2-for-storage-e-g-I-got-an-email-asking-to-upgrade-to-1-TB-and-they-provided-the-conversion-as-being-equal-to-1000-GB-instead-of-1024-GB

NiklasBr:
HP uses base 10: https://support.hp.com/gb-en/document/c01179504

brusch:
Oh dear ... then let's use base 10 ...

NiklasBr:
Bash uses base ten: From the man page: "If base# is omitted, then base 10 is used"

brusch:
follow up, see  #3889

PR #3892, opened 2019-01-30
Fix class name in PageSnippet::getUrl to work properly when used in a Site

LosHawlos:
The link generated for the new "Open in new Window"-Button does not work for Documents in a Site because PageSnippet::getUrl never uses the code to handle the url for a Site. It generates invalid urls that result in a 404 error.
This happens because the class name "Site" is not properly resolved.
Either add a 
use Pimcore\Model\Site;
or change the if contition to use the class name "Model\Site" instead of "Site".


brusch:
Thanks!

PR #3898, opened 2019-01-31
[Versions] Handle incompatible/ missing class more gracefully

ckogler-elements:
If a version file contains a class which doesn't exist anymore we get a " __PHP_Incomplete_Class" response. 
Therefore the we can't open the object anymore in the pimcore admin.

In this case -> delete the version as it is done if we can't get the data.


ckogler-elements:
An other option would be to return "false" instead of the deletion of the version. 
This way the "damaged" version file would be kept but we could at least open the Object in the admin...

PR #3903, opened 2019-02-01
Fix: Translation of predefined properties doesn't work

blankse:
### Steps to reproduce

1. Create a predefined property
2. Open predefined property select in a document
3. Go to Admin translations (The property name doesn't exist)
4. Create Admin translation
5. Open predefined property select in a document (Predefined Properties are not translated)


brusch:
Thanks!

PR #3907, opened 2019-02-01
Documentation: Get Values of Localized Block Entries

putzflorian:
Add the Info how to get Values of Localized Block Entries


brusch:
Thanks!

weisswurstkanone:
thx!

PR #3908, opened 2019-02-01
Object / Asset / Document - ::getById - fixes

ultramegatom:
Object / Asset / Document
If you pass any object to the getById method it will return the root element.
I fixed this by throwing an exception if a non numeric key is passed.

Asset / Document
The getById method does not typematch. e.g. \Pimcore\Model\Asset\Image::getById(...); could return any other Asset for example a \Pimcore\Model\Asset\Document. Same for Documents. 
The behavior is now the same as it is for Objects.

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -23,7 +23,6 @@ use Pimcore\Event\FrontendEvents;
 use Pimcore\Event\Model\AssetEvent;
 use Pimcore\File;
 use Pimcore\Logger;
-use Pimcore\Model\Element\ElementInterface;
 use Pimcore\Tool\Mime;
 use Symfony\Component\EventDispatcher\GenericEvent;
 
@@ -33,6 +32,8 @@ use Symfony\Component\EventDispatcher\GenericEvent;
  */
 class Asset extends Element\AbstractElement
 {
+    use Element\ChildsCompatibilityTrait;
+
     /**
      * possible types of an asset
      *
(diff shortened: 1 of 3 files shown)


ultramegatom:
this might actually be a possible bc break if the method was 'misused'. Maybe we should not it in the upgrade notes?

dpfaffenbauer:
you should also double check this pr, you made properties public again.

ultramegatom:
@dpfaffenbauer oh yes sry was from the wrong branch - should be fine now

dpfaffenbauer:
ok, fine now. Since this is a BC anyway. And I would be fine with that BC, why not just change the type of `getById($id)` to `getById(int $id)`?

dpfaffenbauer:
I think its better to do it right properly once, even though it means a lot more work.

brusch:
I don't think it is actually a real BC (while it is of course a theoretical), ... as long as you're passing a numeric value everything stays the same, if your application code passes some other values, it is currently misbehaving anyway...

brusch:
But tests are failing ... that needs to be investigated and fixed. Thanks!

brusch:
Maybe a broken string, because `is_numeric('2') === true`

models/Asset.php
```diff
@@ -228,16 +228,38 @@ public static function getByPath($path, $force = false)
...
 
+    /**
+     * @param Asset $asset
+     *
+     * @return bool
+     */
+    protected static function typeMatch(Asset $asset)
+    {
+        $staticType = get_called_class();
+        if ($staticType != 'Pimcore\Model\Asset') {
```
dpfaffenbauer: `Asset::class`

models/Document.php
```diff
@@ -237,16 +237,38 @@ public static function getByPath($path, $force = false)
...
 
+    /**
+     * @param Document $document
+     *
+     * @return bool
+     */
+    protected static function typeMatch(Document $document)
+    {
+        $staticType = get_called_class();
+        if ($staticType != 'Pimcore\Model\Document') {
```
dpfaffenbauer: `Document::class`

brusch:
Hmm, I'm just thinking, maybe it's better to return `null` anyway if `$id` is not numeric ... that would be more consistent (also in terms of BC, because `getById()` has not thrown exceptions in the past).

ultramegatom:
@brusch seems fine to me - see latest commit

PR #3910, opened 2019-02-03
Bugfix: AbstractBatchProcessingWorker getDefaultDataForIndex() should respect tenantConfig

Torty:
In the AbstractBatchProcessingWorker when you call getDefaultDataForIndex the getCategories method is hardcoded to the object. 
It would be better to use the tenantconfig here. (like on all the other workers) 
So you can use a different function/field to retrieve "categories" for your object even if you already have categories defined in your object.


fashxp:
makes sense. thx very much.

PR #3916, opened 2019-02-05
[Bug][Ecommerce] joinCustomer cast classId to integer

### LINKED ISSUE #3915: [Bug][Ecommerce] joinCustomer cast classId to integer
# Bug Report

### Expected behavior 
When you create a new order that adds a customer object the customer table will be joined in the "Back office - ordering". Our customer object identifier is 'customer' and not an integer so the join should be 'object_customer'.
  
### Actual behavior  
The customer join casts the classId to integer which results in a zero. This now tries to join a not existing view 'object_0'.
  
### Steps to reproduce  
- Create order and set a customer object
- In the admin go to "Back office - ordering"
- Open the corresponding order, this will throw an error

RuudvdvNB:
Resolves #3915 

## Additional info  
Removed the cast to integer since the identifier can also be a string.


fashxp:
thx

PR #3917, opened 2019-02-05
Expose all classes and interfaces of Assets for js

ultramegatom:
same as https://github.com/pimcore/pimcore/pull/3906 but for assets


ultramegatom:
there you go :)

PR #3918, opened 2019-02-05
Empty role prevents login fix

ultramegatom:
Not sure how this occurs but somethimes the roleId of a User can be empty - that results in a boolean value trying to call ->getName() and prevents the User from logging in.


brusch:
Thanks!

PR #3921, opened 2019-02-07
Detecting UTF-16 encoding fails in CSV import

BlackbitNeueMedien:
Importing a UTF-16 encoded file was not possible for me, since the encoding detection failed and the following conversion to UTF-8 messed up the file. According to the docs at http://php.net/manual/de/function.mb-detect-order.php, detection of UTF-8 and UTF-16 is not supported and will always fail. That why I've taken the BOM detection approach from http://php.net/manual/en/function.mb-detect-encoding.php#91051 and added it to the `detectEncoding` method.

### DIFF
--- a/lib/Tool/Text.php
+++ b/lib/Tool/Text.php
@@ -368,14 +368,31 @@ class Text
      */
     public static function detectEncoding($text)
     {
+        // Detect UTF-8, UTF-16 and UTF-32 by BOM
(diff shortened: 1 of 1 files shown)


brusch:
Thanks! 
But just for the sake of completeness ... if you're using characters/symbols which are only in utf16 or utf32 ... they won't work in utf8 anyway  😉

PR #3924, opened 2019-02-07
cache-buster functionality for Image Thumbnails

MercuryKojo:
added "cache-buster" option for the getHtml() function in Image Thumbnails so you can disable the Browsercache on things like Profile Images


MercuryKojo:
changed option name to camelcase to comply with other option names

brusch:
Grazie Mille!

PR #3926, opened 2019-02-11
Remove redundant isValidPassword helper function

blankse:
This function is also defined in line 2765
https://github.com/pimcore/pimcore/blob/f778e30b42aae7e4a771788d6f3b1ced85b874c2/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js#L2642-L2648

https://github.com/pimcore/pimcore/blob/f778e30b42aae7e4a771788d6f3b1ced85b874c2/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js#L2765-L2771


weisswurstkanone:
thx!

PR #3927, opened 2019-02-11
[Bug] ObjectBrick import does not import title & group

Shobba:
ObjectBrick import currently only sets parent class in general settings, but ignores title and group

## Additional Info
To reproduce:
- Create an object brick
- Set parent class, title and group in general settings
- Export object brick
- Delete object brick
- Create new and empty object brick
- Import from json
- Title & group fields are empty


brusch:
Thanks!

PR #3932, opened 2019-02-12
Add option to reload csv import column configuration

### LINKED ISSUE #3864: [CSV Import] No refresh possible of the column configuration after changing the delimter.
# Bug Report
 
  
### Expected behavior 
 After delimiter is changed it is possible to refresh the column configuration and map the right columns to the right attributes.

### Actual behavior 
The columns are not refreshed which makes it impossible to assign the correct column tot the correct attribute.
  
### Steps to reproduce
1. Make a PIPE("|") separated csv.
2. Use csv import for a random class.
3. Change delimiter to PIPE ("|")
4. Try to match right column to the right attribute.

dvesh3:
Resolves #3864 

## Additional info  
Added button to reload column configuration(with csv settings changes) on Import configuration
refresh_csvimport


brusch:
Thanks!

Issue #3736, opened 2018-12-11
[Data Objects] Lazy Loaded Relations within Localized Fields, Field Collections & Object Bricks


sidd-chaudhery:
# Bug Report

There is performance issue of Pimcore which comes when we add a large number of languages (assume 40-50) into a Pimcore instance.

The problem is many of the functionalities gets affected because of these high no. of languages like folder grid listing, object search, object tree load etc. How they get affected is, because all of these functionalities provide a list of objects and for each object Pimcore fetch the full data (data for localized + non localized attributes) and since the no. of languages is too high it cannot load the localized data in time and the "Communication Faliure" error popup comes up.

The same kind of problem is coming in my current project also. We have around 40 languages and the total no. of localized attributes is 180 (160 input type  + 20 relation type). I tried to look into the part where Pimcore is spending most of the time when getting the object data. I found there is a function load in file Localizedfield Dao file

which does all the work to load the localized data and this function is taking the maximum time. There is one loop running for each language which loads the respective language data in the model. Normally it is taking 3.5 sec to get the data of a single object and if I comment that loop it takes 0.04 sec. Any idea to increase the performance of the loop ?

Second thing which looks unusual is, for relation type fields which are localized Pimcore fires a separate query for each attribute and for each language. For eg. if I have 20 relation type fields and 40 no. of languages then for a single object it fires 800 (20*40) queries. I am not sure if this is the only way to get the data for relation type fields.
  
### Expected behavior   
Error popup should not come and object should load a bit faster.
 
### Actual behavior  
"Communication Failure" error popup comes most of the times.

fashxp:
see also https://github.com/pimcore/pimcore/issues/2998

ascheider:
@sidd-chaudhery yes it is really big issue with pimcore. Try to not using localized fields in relations, because they are not lazy loadable yet. We also optimized many database queries with sql instead of object loading via get-Methods. Sometimes you only need few informations, but pimcore loads all objects and also relations. This leads to huge amount of queries, which are like DDOS on database. Try to change the model, there is no quick fix for this issue i guess.

sidd-chaudhery:
@ascheider - Thanks. Will give it a try.

brusch:
fixed by #3938

ascheider:
@brusch it seems the relations are still not lazy loadable in localized fields, right?

byfareska:
why is that issue closed when the bug still occurrences?

brusch:
[...]

PR #3938, opened 2019-02-13
[Data Objects] Lazy Loaded Relations within Localized Fields, Field Collections & Object Bricks #3736

weisswurstkanone:
see https://github.com/pimcore/pimcore/issues/3736


models/DataObject/ClassDefinition/Data.php
```diff
@@ -738,6 +738,13 @@ public function getGetterCodeObjectbrick($brickClass)
         $code .= '*/' . "\n";
         $code .= 'public function get' . ucfirst($key) . " () {\n";
 
+        if ($this instanceof DataObject\ClassDefinition\Data\Relations\AbstractRelations && $this->getLazyLoading()) {
```
brusch: @weisswurstkanone we should move this part into `preGetData()`, this makes the migration script obsolete

PR #3941, opened 2019-02-14
[Docs] How to remove a cache item created be the templating helper

ultramegatom:
add doc on how to remove a cache item created by the templating helper

### DIFF
--- a/doc/Development_Documentation/02_MVC/02_Template/02_Templating_Helpers/README.md
+++ b/doc/Development_Documentation/02_MVC/02_Template/02_Templating_Helpers/README.md
@@ -106,6 +106,9 @@ of calculation or require a huge amount of objects (like navigations, ...).
 
 `$this->cache(string $name, [int $lifetime = null], [bool $force = false])`
 
+
+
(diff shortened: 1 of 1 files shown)


brusch:
I think it would be better to provide a method on the helper to clear an item, because the prefix may change in the future or additional suffixes such as `webp` will be added.

ultramegatom:
@brusch 
was thinking about that but not sure where to but that. Seems pretty useless inside the helper since i would never clear the cache inside a view. And getting a templating helper and calling a method on it in application code seems like super ugly to me.

Doesn't really seem to fit inside the \Pimcore\Cache class either.

So not sure whats our best option here.

dpfaffenbauer:
@ultramegatom Create a new Service that will also be used by the helper that takes care about Template Caches.

brusch:
@ultramegatom @dpfaffenbauer yes, a dedicated service that handles the caching in templates would be the best solution (should be also used by the helper). 
I'll close this PR for now, please create a new PR for the new service.

PR #3953, opened 2019-02-18
[DocType] Fix: Predefined Document Type Select in Settings Tab

blankse:
In the moment the select shows only undefined > undefined and doesn't work.


blankse:
@brusch Why did you remove my translation function?

brusch:
Ah, sorry, that was by mistake :)

brusch:
I've also added it for the group name.

blankse:
Thanks :)

blankse:
Shouln't be the > sign between group and name?

brusch:
Oh dear, yes, of course - thanks again man :)

PR #3954, opened 2019-02-18
Better performance for loading field collections / blocks

BlackbitNeueMedien:
Loading field collections in Pimcore's backend with e.g. > 20 items can be quite slow. According to https://www.sencha.com/blog/performance-optimization-for-layout-runs/ it is a best practice to not redraw ExtJS component for every child item but to suspend layout calls until all child items have been added. 
Especially field collections and blocks would profit by suspending layout updates. If you can imagine this to be useful also for other data types please comment.

Performance comparison (for a field collection with 47 items) for reloading object:
Without suspending layout: 116 seconds
With suspending layout: 24 seconds


weisswurstkanone:
makes sense, thx!

PR #3956, opened 2019-02-18
[WebsiteSettings] getByName shouldn't log a warning

blankse:
I think the WebsiteSetting::getByName function shouln't log an warning, if the setting doesn't exists.

It is impossible to check if a website settings does exists. So I can't avoid this warning.

Other get functions also doesn't log warnings. Return null should be enough.

### DIFF
--- a/models/WebsiteSetting.php
+++ b/models/WebsiteSetting.php
@@ -117,8 +117,6 @@ class WebsiteSetting extends AbstractModel
         try {
             $setting->getDao()->getByName($name, $siteId, $language);
         } catch (\Exception $e) {
-            Logger::warning($e->getMessage());
-
             if ($language != $fallbackLanguage) {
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Not even if there is one for the fallback language ?

dpfaffenbauer:
it should throw an exception actually.... but that would be a bc break.

blankse:
Exceptions would be nice. We could create a Service class for new get functions and set these functions deprecated. I know it from other projects, that load/create/update functions exists in Service classes and not in the model. I don't know your roadmap. Is this desired here?

dpfaffenbauer:
Unfortunately, I guess not.... Pimcore uses a Custom Active Record pattern, whereas it should use doctrine ORM, but Pimcore doesn’t see the advantage of it.

brusch:
Thanks!

brusch:
We're not logging anything on other entities, so I'd say it's ok and safe to remove the logging here.

PR #3960, opened 2019-02-19
Currently the AbstractTokenManager doesn't check if a voucher series is published…

### LINKED ISSUE #3959: Unpublished VoucherSeries considered valid
# Bug Report

### Expected behavior
The implementation of `AbstractTokenManager->checkToken()` (which is called by default implementation of `IVoucherService` when adding tokens to a cart), doesn't consider tokens of unpublished voucher series' as "valid".

### Actual behavior
The implementation of `AbstractTokenManager->checkToken()` (which is called by default implementation of `IVoucherService` when adding tokens to a cart), considers tokens of unpublished voucher series' as "valid".

### Steps to reproduce
1. Create a Voucher-Series 'Test-Voucher'
2. Add a simple token config (token-code 'simple')
3. Save the voucher 'Test-Voucher' without publishing it
4. "Generate" the token via the related voucher series tab
5. Configure a pricing rule
5.1. Name: Test-Discount
5.2. Active: yes
5.3. Condition: Voucher-Series 'Test-Voucher'
5.4. Action: Cart-Discount 10%
6. Add an item to the cart
7. Go to cart
8. Add voucher-code 'simple'
->> No error message or hint is displayed (e.g. https://demo-advanced.pimcore.org/en/cart/list)

### Notes
Although one might limit Pricing-Rules by date (periods) it's not nice from a user's point of view being able to add a voucher code without being notified about the fact that's not valid anymore. Users might think there is an error as they only see their expected discount not being applied.

dergab:
...in checkToken(). This results in vouchers/tokens that can't be restricted in use e.g. for a period of time by using the publish-schedueling.

* Added checkVoucherSeriesIsPublished(string $code) method to AbstractTokenManager. It checks if the voucher-series (for the related token) is published, and throws a VoucherServiceException if not.
* Added a call to checkVoucherSeriesIsPublished in checkToken().

Resolves #3959

### DIFF
--- a/bundles/EcommerceFrameworkBundle/VoucherService/TokenManager/AbstractTokenManager.php
+++ b/bundles/EcommerceFrameworkBundle/VoucherService/TokenManager/AbstractTokenManager.php
@@ -64,11 +64,38 @@ abstract class AbstractTokenManager implements ITokenManager
      * @param ICart $cart
      *
      * @return mixed
+     *
+     * @throws VoucherServiceException When validation fails for any reason
      */
     public function checkToken($code, ICart $cart)
     {
+        $this->checkVoucherSeriesIsPublished($code);
         $this->checkAllowOncePerCart($code, $cart);
         $this->checkOnlyToken($cart);
+    }
+
+    /**
+     * Only tokens of published voucher series' may be used.
+     *
+     * @param string $code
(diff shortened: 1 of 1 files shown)


fashxp:
thx

PR #3963, opened 2019-02-19
Thumbnail - preserve color space

ckogler-elements:
Do not change the colorspace if "PreserveColor" is set. Otherwise cmyk images would be converted to rgb...


brusch:
Thanks!

PR #3965, opened 2019-02-19
Update README.md

jacobsyentl:
Added an example to create and update objects since the previous description was wrong.
The body of an UPDATE/DELETE does not correspond with the response of a GET.

The user must not make use of the "data" property to construct it's body when creating/updating.


jacobsyentl:
Thread with other user who had problems with the documentation:
https://talk.pimcore.org/t/create-object-via-api-pimcore-model-dataobject-was-not-found/1569/8

brusch:
Thanks!

PR #3966, opened 2019-02-19
Add jobConfig settings to pimcore.document.print.prePdfGeneration event

ckogler-elements:
Add jobConfig settings to pimcore.document.print.prePdfGeneration event so we can modify the html rendering depending on custom pdf generation settings...
E.g use different Thumbnails...


fashxp:
thx

PR #3967, opened 2019-02-20
Centralize classes operations and allow to initialize using pimcore h…

msphn:
…elper

Improving database setup in Pimcore helper.

Especially for codeception stuff.

### DIFF
--- a/bundles/CoreBundle/Resources/config/services.yml
+++ b/bundles/CoreBundle/Resources/config/services.yml
@@ -192,6 +192,9 @@ services:
 
     Pimcore\Translation\Escaper\Xliff12Escaper: ~
 
+    Pimcore\Model\DataObject\ClassDefinition\ClassDefinitionManager:
+        public: true
+
     # Override Swiftmailer redirecting plugin
     swiftmailer.plugin.redirecting.abstract:
         abstract: true
--- a/bundles/CoreBundle/Resources/config/services.yml
+++ b/bundles/CoreBundle/Resources/config/services.yml
@@ -192,6 +192,9 @@ services:
 
     Pimcore\Translation\Escaper\Xliff12Escaper: ~
 
+    Pimcore\DataObject\ClassesManager:
+        public: true
+
     # Override Swiftmailer redirecting plugin
     swiftmailer.plugin.redirecting.abstract:
         abstract: true
--- a/bundles/CoreBundle/Resources/config/services.yml
+++ b/bundles/CoreBundle/Resources/config/services.yml
@@ -192,9 +192,6 @@ services:
 
     Pimcore\Translation\Escaper\Xliff12Escaper: ~
 
-    Pimcore\DataObject\ClassesManager:
-        public: true
-
     # Override Swiftmailer redirecting plugin
     swiftmailer.plugin.redirecting.abstract:
         abstract: true
--- a/lib/DataObject/ClassesManager.php
+++ b/lib/DataObject/ClassesManager.php
@@ -14,6 +14,8 @@
 
 namespace Pimcore\DataObject;
 
+use Pimcore\Db;
+use Pimcore\Db\Connection;
 use Pimcore\Model\DataObject\ClassDefinition;
 
 /**
@@ -27,11 +29,25 @@ class ClassesManager
     public const DELETED = 'deleted';
 
     /**
+     * @var Db
+     */
+    private $db;
+
+    /**
+     * ClassesManager constructor.
+     * @param Connection $db
+     */
+    public function __construct(Connection $db)
+    {
+        $this->db = $db;
+    }
+
+    /**
      * Delete all classes from db
      */
     public function deleteClasses(): array
     {
-        $classes = \Pimcore\Db::get()->fetchAll('SELECT * FROM classes');
+        $classes = $this->db->fetchAll('SELECT * FROM classes');
         $deleted = [];
 
         foreach ($classes as $class) {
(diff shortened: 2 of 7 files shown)


bundles/CoreBundle/Manager/ClassesManager.php
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Pimcore\Bundle\CoreBundle\Manager;
+
+use Pimcore\Model\DataObject\ClassDefinition;
+
+/**
+ * Class ClassesManager
+ * @package Pimcore\Bundle\CoreBundle
+ */
+class ClassesManager
```
dpfaffenbauer: I guess the class should be somewhere here: https://github.com/pimcore/pimcore/tree/master/lib/DataObject

msphn: Thank you! Good hint, will do so.

turko:
That is exactly what I was looking for! Please merge :)

lib/DataObject/ClassesManager.php
```diff
@@ -0,0 +1,84 @@
...
+    public const SAVED = 'saved';
+    public const CREATED = 'created';
+    public const DELETED = 'deleted';
+
+    /**
+     * Delete all classes from db
+     */
+    public function deleteClasses(): array
+    {
+        $classes = \Pimcore\Db::get()->fetchAll('SELECT * FROM classes');
```
dpfaffenbauer: you should inject $db

msphn:
Any chance we can get this merged? My CI / CD is actually based on this 💃

lib/DataObject/ClassesManager.php
```diff
@@ -0,0 +1,100 @@
...
+
+use Pimcore\Db;
+use Pimcore\Db\Connection;
+use Pimcore\Model\DataObject\ClassDefinition;
+
+/**
+ * Class ClassesManager
+ * @package Pimcore\DataObject
+ */
+class ClassesManager
```
brusch: I'd prefer `ClassDefinitionManager` to be consistent in naming. 
Maybe we should also put it under the `Pimcore\Model\DataObject\ClassDefinition` namespace.

lib/DataObject/ClassesManager.php
```diff
@@ -0,0 +1,100 @@
...
+     */
+    public function __construct(Connection $db)
+    {
+        $this->db = $db;
+    }
+
+    /**
+     * Delete all classes from db
+     */
+    public function deleteClasses(): array
```
brusch: We should name this method accordingly to what it actually does. 
What about `cleanupDeletedClassDefinitions()` or anything more meaningful?

lib/DataObject/ClassesManager.php
```diff
@@ -0,0 +1,100 @@
...
+            }
+        }
+
+        return $deleted;
+    }
+
+    /**
+     * Updates all classes from PIMCORE_CLASS_DIRECTORY
+     */
+    public function updateClasses(): array
```
brusch: Same here ... maybe `createOrUpdateClassdefinitions()` ?

msphn:
@brusch thanks for review! Hope that looks better!

bundles/CoreBundle/Resources/config/services.yml
```diff
@@ -192,6 +192,9 @@ services:
 
     Pimcore\Translation\Escaper\Xliff12Escaper: ~
 
+    Pimcore\DataObject\ClassesManager:
```
brusch: This needs to be changed as well.

lib/Model/DataObject/ClassDefinition/ClassDefinitionManager.php
```diff
@@ -0,0 +1,100 @@
...
+    public const CREATED = 'created';
+    public const DELETED = 'deleted';
+
+    /**
+     * @var Db
+     */
+    private $db;
+
+    /**
+     * ClassesManager constructor.
```
brusch: This can be removed.

bundles/CoreBundle/Resources/config/services.yml
msphn: ah damn it, sorry was in hurry, I removed the service declaration. There should be no need for it.

brusch: Sure we need a service definition, but the name should be according to the fully qualified class name: `Pimcore\Model\DataObject\ClassDefinition\ClassDefinitionManager`

lib/Model/DataObject/ClassDefinition/ClassDefinitionManager.php
brusch: Line `33` is still there, ... that should go.

bundles/CoreBundle/Resources/config/services.yml
msphn: Do we? The autowiring seems to work well for it.

tests/_support/Helper/Pimcore.php
```diff
@@ -161,10 +164,13 @@ protected function setupDbConnection()
 
         $connection = $this->getDbConnection();
 
-        $connected = false;
         if ($this->config['initialize_db']) {
             // (re-)initialize DB
             $connected = $this->initializeDb($connection);
+            if ($this->config['setup_objects']) {
+                $this->debug('[DB] Initializing objects');
+                $this->kernel->getContainer()->get(ClassDefinitionManager::class)->createOrUpdateClassDefinitions();
```
brusch: What about this line then, if it's not defined as a public service?

PR #3969, opened 2019-02-20
Improvement: support workflow permissions for all element types

mcuenez:
# Why merge?
The `WorkflowManagementListener` only allows to use permissions for elements of type `ConcreteObject`. However, in some situations it may be needed to restrict permissions for other element types (such as documents, and assets).

# What changed?
The code of the `WorkflowManagementListener` was updated to also allow other element types.


fashxp:
thx

PR #3973, opened 2019-02-21
[WebsiteSettings] getData should lazy load like Property class

blankse:
The Property class loads the elements but the WebsiteSetting class doesn't:
https://github.com/pimcore/pimcore/blob/652a614a0c50d1b33212471e548e3cc96a517aad/models/Property.php#L140-L152

I know its a kind of bc break. Perhaps it is possible in 5.7.0 with a bc notice?

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/01_Within_V5/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/01_Within_V5/README.md
@@ -1,11 +1,11 @@
 # Upgrade Notes for Upgrades within Pimcore 5
 
 ## Version 5.7.0
-- `WebsiteSetting` properties are now `protected` instead of `public`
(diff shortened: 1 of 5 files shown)


brusch:
It only breaks BC if you're accessing the `data` property in your code, or I'm missing something? 
I think we should then take the chance to change all the properties of `WebsiteSetting` to `protected`, can you add that to your PR? 
Please also put a note here https://github.com/pimcore/pimcore/blob/master/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/01_Within_V5/README.md#L2-L1
We'll merge it into 5.7 then.

blankse:
@brusch Done :)

dpfaffenbauer:
tests break cause they are accessing the `Property` properties directly

blankse:
I will correct the tests on Monday

blankse:
Tests are correct now

brusch:
@blankse needed to revert this PR, due to #4095

PR #3988, opened 2019-02-25
PhpDoc - loadIdList does not return an integer array.

ultramegatom:
loadIdList does not return an integer array.

### DIFF
--- a/models/Asset/Listing.php
+++ b/models/Asset/Listing.php
@@ -25,7 +25,7 @@ use Zend\Paginator\AdapterAggregateInterface;
  * @method Model\Asset[] load()
  * @method int getTotalCount()
  * @method int getCount()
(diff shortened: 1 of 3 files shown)


brusch:
Thanks!

dpfaffenbauer:
what else does it return? string array with ids? Cause it should be int[]

blankse:
The Ids are strings. Is it possible to cast these to integer?
```
array (size=4)
  0 => string '447' (length=3)
  1 => string '441' (length=3)
  2 => string '445' (length=3)
  3 => string '443' (length=3)
```

dpfaffenbauer:
thats the point I tried to make with my comment ;)

blankse:
I created a PR for this https://github.com/pimcore/pimcore/pull/4003

PR #3994, opened 2019-02-27
Fix PhpDoc 

jdreesen:
This fixes the order of the type & parameter name in the PhpDoc and adds a missing use statement.


weisswurstkanone:
thx

PR #3995, opened 2019-02-27
Link / Hardlink - relation improvements

ultramegatom:
Make the relations inside link/hardlink document types look and feel more like normal relations.

-> add open, delete, search button
-> add context menu


ultramegatom:
image

weisswurstkanone:
Thanks, as a follow-up task I would suggest to extract all that "copy&paste" common stuff  and replace it with helper functions.

PR #4000, opened 2019-02-27
video properties bugfix

markus-moser:
- The thumbnail processor might save the asset and then when the properties are already converted to an array an exception might occur.
- Therefore this fix does the array conversion later in the controller.


brusch:
Thanks!

PR #4004, opened 2019-02-28
[Targeting] allow to set custom VisitorInfo via event

dpfaffenbauer:
This allows to define a custom VisitorInfo. I have currently the use-case where I track page-views and sessions manually, but still wanna use the Targeting Engine. With this little change, it is possible for me to write my own VisitorInfo "Resolver" by defining the Visitor based on an event.


brusch:
Thanks!

PR #4005, opened 2019-02-28
[Video Editable] Fix: No poster if no thumbnail option is set

blankse:
If I create a video editable without set the thumbnail option I get no poster.
```
{{ pimcore_video('video', {
    'attributes': {
        'class': 'video-js vjs-fluid vjs-default-skin vjs-big-play-centered',
        'data-setup': '{}'
    }
}) }}
```


brusch:
Thanks!

PR #4008, opened 2019-03-01
notifyRoles and notifyUsers are only in notificationSettings configurable

smoench:
See at latest core bundle configuration class https://github.com/pimcore/pimcore/blob/master/bundles/CoreBundle/DependencyInjection/Configuration.php


fashxp:
thx

PR #4014, opened 2019-03-04
Add option to reset target groups for a visitor

Cruiser13:
There is no possibility to remove target groups from a visitor. This does add a function to reset (remove) the target groups.


brusch:
Thanks!

PR #4016, opened 2019-03-04
Es Bugfix

ckogler-elements:
- Create version dir if it doesn't exist
- Disable emergency logging (there are valid use cases where it is reasonable)


brusch:
Thanks!

PR #4019, opened 2019-03-05
Ecommerce order/order item tables prefix are not number but strings h…

krishjun:
…ence replacing %d with %s

  
  

Resolves #


fashxp:
thx

PR #4023, opened 2019-03-05
[Web2print] Fix background PDF generation with wkhtmltopdf on OS X

### LINKED ISSUE #4022: [Web2print] Fix background PDF generation on OS X
I tried to generate PDF print previews on OS x which did not work. I checked the Web2print log and found this error:

> nohup: can't detach from console: Inappropriate ioctl for device

This is very easy to fix, a PR will be submitted soon

Ben305:
Fix "nohup: can't detach from console: Inappropriate ioctl for device" error when generating PDFs on OS x using wkthmltopdf

Resolves #4022

### DIFF
--- a/lib/Tool/Console.php
+++ b/lib/Tool/Console.php
@@ -448,7 +448,7 @@ class Console
         Logger::debug('Executing command `' . $commandWrapped . '´ on the current shell in background');
         $pid = shell_exec($commandWrapped);
 
-        Logger::debug('Process started DARWIN with PID ' . $pid);
+        Logger::debug('Process started with PID ' . $pid);
 
         return $pid;
     }
--- a/lib/Tool/Console.php
+++ b/lib/Tool/Console.php
@@ -364,7 +364,7 @@ class Console
         if (self::getSystemEnvironment() == 'windows') {
             return self::execInBackgroundWindows($cmd, $outputFile);
         } else if (self::getSystemEnvironment() == 'darwin') {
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
ok, in my opinion: Pimcore doesn't support Mac OS as operating system, therefore Pimcore should not care about that. OSX also isn't a suitable server infrastructure to run Pimcore on.

BTW: I am also working on Mac and I never had issues with that before.

Ben305:
I wouldn't use OS X in production, but seems suitable for me as dev environment. Chances to break anything with this bugfix are fairly small.

dpfaffenbauer:
True, but thats not my concern. My concern is that this is another piece of code that has to be maintained and refactored as well some day. "just to support osx".

lib/Tool/Console.php
```diff
@@ -410,6 +414,45 @@ protected static function execInBackgroundUnix($cmd, $outputFile)
...
 
+    /**
+     * @static
+     *
+     * @param string $cmd
+     * @param string $outputFile
+     *
+     * @return int
+     */
+    protected static function execInBackgroundDarwin($cmd, $outputFile)
```
brusch: This function is actually a 1:1 copy of `execInBackgroundUnix` except the `nohup` command. I guess it would be better to just add an additional parameter `$useNohup` to the existing method.

Ben305: Has been fixed

Issue #3135, opened 2018-07-25
Support of 'Objects with Matadata' field with Non-owner object relation field


psuhane:
Support of 'Objects with Matadata' field with Non-owner object relation field

brusch:
This is a feature request, currently it's intended to work with "Objects" data-type only.

SEYentl:
Would be nice if it also works with "single" href.

dergab:
@psuhane, we found a kind of a hidden path to achieve that. At least with Pimcore v5.5. First add a standard object-relation field (e.g. F1) to your model (e.g. in class A), then add the non-owner relation field (e.g. F2) to the other side (e.g. class B) referencing F1. After that you can change ("Convert to") F1 in class A to objects with metadata type. After saving it the F2 still relates (shows) objects added to F1. ;)

@brusch, this is really a drawback to not have this feature "enabled" for objects with metadata. We frequently have to workaround this missing option. Maybe the above "workaround" hints on a quick way to implement it (properly), so it finds its way into the core soon? ;D

brusch:
@dergab maybe you should think about making a PR for that, instead of frequently working around this missing feature 😉

dergab:
> @dergab maybe you should think about making a PR for that, instead of frequently working around this missing feature 😉
I did think about it, but I'm not very experienced in PHP-dev. So I fear a PR (in such an important part of the core) could cause more time spend on reviewing and commenting than the actual implementation. ;)
...but maybe I'll give it a try anyways...

dergab:
As it turned out it wasn't too much of coding...basically nothing at all... 😬

fashxp:
Dear community, it's been a while since that issue was updated the last time, so we decided to close it for now. If you're still interested in contributing, feel free to add comments here or contact us before creating a pull request to discuss next steps.

Thanks for your understanding.

PR #4031, opened 2019-03-06
Added objects-with-metadata-relation to supported non-owner-field-types.

dergab:
Resolves/adds feature-request #3135 by adding objects-with-metadata type to list of allowed field-types when loading available fields for non-owner-object-relation in admin.


brusch:
Unfortunately this was not all ... displaying relations works, but it's not possible to add a new relation onto the reverse-many-to-many field. 
Please complete this functionality, then we can merge it :) 
Thanks!

dergab:
@brusch, I had a look at it. Without detailed knowledge (or documentation) about how Pimcore handles views and data configurations, this is not a task for me. There are no comments at all in the classes in /AdminBundle/Controller/Admin/DataObject/. I'm sorry, but I think that the time it would take me to sort this out would greatly exceed the value. We'll have to stick to the workaround.

PR #4036, opened 2019-03-07
do not throw exceptions when properties in data object versions do not exist any more

markus-moser:
Currently when for example someone deletes an object brick Pimcore will throw an Exception when someone tries to show a previous version. This fix silently ignores these fields in the version preview.


brusch:
Thanks!

PR #4040, opened 2019-03-07
Fix @var annotation and test for CONST with identical equals operator

NiklasBr:

Resolves no specific issue, but just things I noticed while reading the code.


fashxp:
thx

PR #4042, opened 2019-03-07
PHPDoc and empty conditions checks in Bracket.php 

NiklasBr:
Resolves no issue, just tiny things I noticed reading the code.

## Additional info  

`sizeof` is just an alias of `count` and computation time scales linearly with the size of the array, `empty` is a single operation. This might save many *μ-seconds* I tell you! ;)


brusch:
Thanks :)

PR #4043, opened 2019-03-07
Video conversion not working when storing assets in S3

BlackbitNeueMedien:
### Expected behavior
Converting videos (and creating video thumbnails) works regardless of the asset storage location.

### Actual behavior
When I use S3 as asset store (set up as described [in the docs](https://pimcore.com/docs/5.x/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/Amazon_AWS_Setup/Amazon_AWS_S3_Setup.html)), it is not possible to use the ffmpeg adapter to convert videos or generate video thumbnails. The problem lies with the creation of the ffmpeg cli command. In [Video/Adapter/Ffmpeg.php](https://github.com/pimcore/pimcore/blob/a4b5160f2e26d00bf2614b40d7bcf11b583898a5/lib/Video/Adapter/Ffmpeg.php#L120), the path to the file to convert is created as `realpath($this->file)`, where `$this->file` is in the format of `s3://my-bucket/assets/video.mp4`. The `realpath` call will then return false, and the resulting command has no input file argument.

### DIFF
--- a/lib/Video/Adapter/Ffmpeg.php
+++ b/lib/Video/Adapter/Ffmpeg.php
@@ -38,6 +38,11 @@ class Ffmpeg extends Adapter
     protected $arguments = [];
 
     /**
+     * @var array
+     */
+    private $tmpFiles = [];
+
+    /**
      * @return bool
      */
     public function isAvailable()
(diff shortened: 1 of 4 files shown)


BlackbitNeueMedien:
That check is for the destination file, not the video source. But still, you pointed me in the right direction here. Checking with `stream_is_local` at some additional places does the trick 😄

PR #4046, opened 2019-03-08
[Custom Reports] Add openUrl as row action

### LINKED ISSUE #4045: [Custom Reports] Add openUrl as row action
Use the column value als url and open it in a new window

image

datom:
resolves #4045


brusch:
Thanks!

PR #4048, opened 2019-03-08
Quoting User Roles to not cause a SQL Query error

MercuryKojo:
User roles can be string (role_hierarchy settings in yml) and would cause a SQL Query Error


MercuryKojo:
Sorry just found the real reason for the problem, please ignore!

PR #4052, opened 2019-03-11
[workflow] email notifications: set correct translator locale

markus-moser:
Currently the translator locale does not work correctly


fashxp:
thx

PR #4055, opened 2019-03-11
do not output useless language hint for field collections

markus-moser:
The language output does not make any sense here.


brusch:
Thanks!

PR #4059, opened 2019-03-12
[Routing] Defaults by Symfony Config

blankse:
We want use the constants to set a default bundle, controller and view. In the moment it is not possible.


dpfaffenbauer:
Please don't introduce anymore constants, use Container Parameters instead, you can fill them with constant values or env vars.

blankse:
These are no new constants. Are documented in the example:
https://github.com/pimcore/skeleton/blob/master/app/constants.example.php

But yes should be parameters...

blankse:
We could integrate these settings in the symfony configuration tree and set the default value to these constants. 
https://github.com/pimcore/pimcore/blob/master/bundles/CoreBundle/DependencyInjection/Configuration.php

WDYT? Any suggestions for the config tree?

dpfaffenbauer:
oh wow, true, these contants really exists. But as they don't work anyway, I would much rather use the Symfony Config Tree for that

brusch:
Config tree would be fine for me ... what about adding them to the existing `routing` node? 
```yml
pimcore: 
  routing: 
    defaults: 
      bundle: AppBundle
      controller: Default
      action: default
```

dpfaffenbauer:
👍

brusch:
@blankse can you update your PR, please?  :)

blankse:
@brusch Yes I have already started :)

blankse:
@brusch I updated the PR :)

bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -350,6 +350,20 @@ private function addRoutingNode(ArrayNodeDefinition $rootNode)
                 ->arrayNode('routing')
                     ->addDefaultsIfNotSet()
                     ->children()
+                        ->arrayNode('defaults')
+                            ->addDefaultsIfNotSet()
+                            ->children()
+                                ->scalarNode('bundle')
+                                    ->defaultValue(PIMCORE_SYMFONY_DEFAULT_BUNDLE)
```
brusch: Imho we can completely get rid of those constants and define the default values in the config-tree definition. 
If someone wants to use constants anyway, there's a special syntax for the use in the config.

blankse: I removed the constants and add an BC notice

blankse:
I added docs. There I noticed that it concerns only the routing of the documents. Should not it be in the documents tree?

brusch:
No, it's not only used for documents, but also for generic purposes, like here: 
https://github.com/pimcore/pimcore/blob/4c986d60d890898e466533a21dfd29bde2acb9f6/bundles/CoreBundle/EventListener/TemplateControllerListener.php#L110 
or for static routes, here: 
https://github.com/pimcore/pimcore/blob/433957d8aaf67d4054b0c2bcb918ee48e8c78a10/lib/Tool.php#L299

We could think about a dedicated config for documents, but that sounds a bit too much for me :)

dpfaffenbauer:
There are already pre-defined document types, so that should be enough

blankse:
Rebase because of merge conflict in README.md

brusch:
Hmm, but with your last changes you made the system setting for the default controller & action obsolete - which is fine for me, but then we should also remove it from the UI and announce that in the upgrade notes.

brusch:
Sorry, pressed the wrong button 🤓

blankse:
@brusch I removed the system settings for default controller and action and add the bc break notice

blankse:
@brusch Did you mean that the settings were no longer used? The Tool:getRoutingDefaults() function had used it.

brusch:
@blankse oh yes, you're absolutely right ... but the system settings had priority, so the 2 config options for action and controller were pretty much useless ... especially for existing setups where the default controller/action are set in system settings. I'd say the current implementation is more consistent and less confusing 👌

bundles/AdminBundle/Controller/Admin/MiscController.php
```diff
@@ -93,12 +94,17 @@ public function getAvailableControllersAction(Request $request, ControllerDataPr
...
+        $routingDefaults = $this->getParameter('pimcore.routing.defaults');
         $bundle = $request->get('moduleName');
         if (empty($bundle)) {
-            $bundle = 'AppBundle';
+            $bundle = $routingDefaults['bundle'];
         }
 
         $controller = $request->get('controllerName');
+        if (empty($controller)) {
+            $bundle = $routingDefaults['controller'];
```
brusch: This should be `$controller = ...` i guess.

blankse: Of course, I fixed it

brusch:
I guess that's it ... let's wait for the tests to be finished, then I'm going to merge the PR.

lib/Tool.php
```diff
@@ -289,29 +289,11 @@ public static function getLanguageFlagFile($language)
...
-
-            // get configured settings for defaults
-            $systemRoutingDefaults = $config->documents->toArray();
-
-            foreach ($routingDefaults as $key => $value) {
-                if (isset($systemRoutingDefaults['default_' . $key]) && $systemRoutingDefaults['default_' . $key]) {
-                    $routingDefaults[$key] = $systemRoutingDefaults['default_' . $key];
-                }
-            }
+        $container = \Pimcore::getContainer();
```
dpfaffenbauer: you could move that to a compiler pass, so there would be one less call to \Pimcore::getContainer() ;)

blankse: You mean the module-bundle mapping? I think we can do this in the extension class too. Than we can use everywhere the paramerter directly and can remove this function. Is this what you think?

dpfaffenbauer: Thinking about that: I guess you don't really need the getRoutingDefaults function at all. You can load it from the container directly when you need it. The only place you would need to call ```Pimcore::getContainer``` again is the Renderlet, cause they don't support dependency injection.

brusch: For me it's fine as it is now, I'd keep `getRoutingDefaults()` anyway to avoid possible BC issues.

PR #4062, opened 2019-03-13
[Document] Fix multihref editable elements visibility

### LINKED ISSUE #2117: Multihref Editable returns unpublished objects
```php
$this->multihref('contactObjects')->getElements()
```

returns unpublished objects since switch from Object to DataObject

dvesh3:
Resolves #2117


brusch:
Modifying `setElements()` can result into unexpected behavior, it would be better if the iterator considers the published state.

PR #4063, opened 2019-03-13
[GDPR Extractor] Fix sent emails full-text search condition

### LINKED ISSUE #3909: [Bug] GDPR Bug with email quote and fulltext search
# Bug Report
 

This behaviour increases the risk for legal trouble because the GDPR extractor might now export return results which are from the wrong user and therefore cause the customer service representative to export incorrect data and unknowingly leak personal information.

I think I have narrowed the problem down to `EmailController::emailsLogsAction()`, [here](https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Controller/Admin/EmailController.php#L40), and how the *@* is replaced and the entire string is quoted.

### Expected behaviour

When searching for *someuser2134@example.com* only results matching *someuser2134@example.com* should be returned
  
### Actual behaviour  

Any user matching only the domain part is returned.

### Steps to reproduce  

1. Open the GDPR data extractor
2. Search for an email address.
3. Look at the results

skarmbild 46

dvesh3:
Resolves #3909

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/EmailController.php
+++ b/bundles/AdminBundle/Controller/Admin/EmailController.php
@@ -61,8 +61,18 @@ class EmailController extends AdminController
                 $filterTerm = str_replace('%', '*', $filterTerm);
                 $filterTerm = htmlspecialchars($filterTerm, ENT_QUOTES);
 
+                if(strpos($filterTerm, '@')) {
+                    $parts = explode(' ', $filterTerm);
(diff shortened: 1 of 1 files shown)


brusch:
Unfortunately this causes new issues ... 
- boolean operators don't work anymore
- it is now always an "exact match" search, so it is unpossible to search for 2 terms (e.g. for a keyword in the subject and an email address)

PR #4067, opened 2019-03-13
Fix pimcore:migrations:migrate command

blankse:
I get following error if I want execute the migrate command:
`php bin/console pimcore:migrations:migrate -s pimcore_core`

```
In PimcoreMigrationsConfiguration.php line 123:
Type error: Return value of Pimcore\Migrations\Command\MigrateCommand::getConnection() must be an instance of Pimcore\Migrations\Command\Traits\Connection, instance of Pimcore\Db\Connection returned
```

Regression from https://github.com/pimcore/pimcore/pull/4044


brusch:
Thanks!

PR #4069, opened 2019-03-14
Protect also img and video thumbnails

putzflorian:
Add routing.yml info to protect also img and video thumbnails


brusch:
Thanks!

PR #4075, opened 2019-03-14
[Custom Layouts] Fix to disallow custom layout invalid name on update

### LINKED ISSUE #4061: Can't edit class custom layout after adding - to its name
# Bug Report

### Expected behavior
Can't save custom layout if name is invalid
### Actual behavior
Can add - to custom layout name when editing it and save. After that can't save this custom layout anymore. Always get error "Invalid name"
### Steps to reproduce

1.  Create custom layout for class
2. Edit name of custom layout class so it has - in its name
3. Save
4. Remove - from name and try saving again.

dvesh3:
Resolves #4061


weisswurstkanone:
thx!

PR #4091, opened 2019-03-19
"Show in tree" did not work if last element in tree page is folder

BlackbitNeueMedien:
When you click on "Show in tree" on a data object there is a bug in the current version when the last element of the data object tree page is a folder. In this case in https://github.com/pimcore/pimcore/blob/edd0f3e8457f0a230bf37819668716b01f9edcab/bundles/AdminBundle/Resources/public/js/pimcore/treenodelocator.js#L502-L508 it is only checked if last element is a folder - if it is, the tree gets paged forwards without checking if the folder name is alphabetically after the object being searched for.

This PR changes this behaviour and also simplifies the code. I have not understood why folders and objects are treated differently because - unlike as e.g. in Windows - Pimcore folders are not sorted after objects but alphabetically by name regardless of their type.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/treenodelocator.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/treenodelocator.js
@@ -337,7 +337,7 @@ pimcore.treenodelocator = function()
             var firstelementChild = childNodes[0];
             var lastelementChild = childNodes[childCount-1];
             
-            if (pagingState.elementType == "document") {
+            if (globalState.elementType == "document") {
                 direction = self.getDirectionForElementsSortedByIndex(
                     pagingState.elementKey,
                     firstelementChild,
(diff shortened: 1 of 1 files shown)


brusch:
Unfortunately it seems that you have over-optimized the code a bit 😉
You're right, documents (manual order) and objects (manual or a-z ordered) are sorted independently of their type, but not assets, they are sorted as known from the windows explorer 😊

Can you please update your PR? Thanks again!

BlackbitNeueMedien:
Now with folder-first order for assets.

dvesh3:
@BlackbitNeueMedien thanks for your PR! could you please share more details how to reproduce the issue with steps? thanks!

BlackbitNeueMedien:
How to reproduce:
# Create folder in data objects panel with name / key "a" in a certain root folder (can also be "Home")
# Copy and paste this object 30 times so data object tree gets paginated
# Create folder in data objects panel with name / key "b" in same root folder
# Copy and paste this object 30 times so data object tree gets paginated to 3 pages
# Create folder in data objects panel with name / key "c" in same root folder
# Copy and paste this object 30 times so data object tree gets paginated to 4 pages
# Create object with key "b1" (will be on page 2 just between folder "b" and "b_copy").
# go to page 3 in tree
# Open "b1" and click "show in tree"
# Result: Tree loads page 4 and does not find object "b1"

BlackbitNeueMedien:
Are there any remaining problems with this PR? @diveshp3 approved the changes but has not merged them yet.

PR #4093, opened 2019-03-19
[Translations] Added csv settings interface for translations import

### LINKED ISSUE #1644: Translation import - determineCsvDialect fails for quotechar if too many single quotes are in the first few lines
# Bug Report
If there are too many single quotes in the first few lines, the automatic csv dialect detection fails and detects single quote as quotechar. If you import a file in to an existing project all keys are duplicate with double quotes afterwards. Maybe it would be better to detect the first character of the line or the combination of quote character and delimiter.
  
### Code example 
```
"'%hostname%' is not a valid hostname for email address '%value%'";"'%hostname%' ist keine gültige Domain für E-Mail-Adresse '%value%'";""
"'%value%' appears to be a dns hostname but cannot match tld against known list";"'%value%' ist kein gültige TLD";""
"'%value%' appears to be a local network name but local network names are not allowed";"'%value%' ist keine gültige Domain";""
"'%value%' does not appear to be a valid ip address";"";""
"'%value%' does not appear to be a valid local network name";"";""
"'%value%' does not match the expected structure for a dns hostname";"'%value%' weist keine gültige Domain-Struktur auf.";""
"'%value%' is not a valid email address in the basic format local-part@hostname";"'%value%' ist keine gültige E-Mail-Adresse";
```

dvesh3:
Resolves #1644


bundles/AdminBundle/Controller/Admin/TranslationController.php
```diff
@@ -96,6 +109,39 @@ public function importAction(Request $request)
...
+
+        //ignore if line terminator is already hex otherwise generate hex for string
+        if (!empty($dialect->lineterminator) && empty(preg_match('/[a-f0-9]{2}/i', $dialect->lineterminator))) {
+            $dialect->lineterminator = bin2hex($dialect->lineterminator);
+        }
+
+        return $this->adminJson([
+            'success' => true,
+            'config' => [
+                'tmpFile' => $importFile,
```
brusch: Here we should only expose the filename not the entire server path.

bundles/AdminBundle/Controller/Admin/TranslationController.php
```diff
@@ -51,25 +51,38 @@ class TranslationController extends AdminController
     public function importAction(Request $request)
     {
         $admin = $request->get('admin');
+        $dialect = $request->get('csvSettings', null);
+        $tmpFile = $request->get('importFile');
```
brusch: And here we need to validate the filename and append it to the known path (`PIMCORE_SYSTEM_TEMP_DIRECTORY`).

PR #4104, opened 2019-03-20
Resolves #4103

### LINKED ISSUE #4103: Element::getByPath() doesn't work properly with class overrides
Hello,

I noticed the addition of this method in file `vendor\pimcore\pimcore\models\Asset.php`.

```php
<?php
    /**
     * @param Asset $asset
     *
     * @return bool
     */
    protected static function typeMatch(Asset $asset)
    {
        $staticType = get_called_class();
        if ($staticType != Asset::class) {
            if (!$asset instanceof $staticType) {
                return false;
            }
        }
        return true;
    }
```

This new method is used in getById of the same class to check the integrity of the fetched asset.

This breaks the currently allowed [class overrides of Pimcore](https://pimcore.com/docs/5.x/Development_Documentation/Extending_Pimcore/Overriding_Models.html) because it evaluates the called class to the default Pimcore asset class. This means if you call the `getByPath` method through an override class it will return `false` as the called class is not explicitly the default one.

SmashingQuasar:
Resolves #4103:

Rewrote the code to make it shorter and easier to read.
Switched double equal signs to triple equal signs to improve code quality.
Added an extra condition to allow this method to work with [Pimcore model override](https://pimcore.com/docs/5.x/Development_Documentation/Extending_Pimcore/Overriding_Models.html).

## Additional info  

I submitted an issue earlier about a new method that breaks the model override. I saw it had been assigned to a developer however I had already fixed it locally so I just submit this pull request to save you time.

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -236,13 +236,8 @@ class Asset extends Element\AbstractElement
     protected static function typeMatch(Asset $asset)
     {
         $staticType = get_called_class();
-        if ($staticType != Asset::class) {
-            if (!$asset instanceof $staticType) {
-                return false;
-            }
-        }
 
-        return true;
+        return ($staticType === Asset::class) || ($asset instanceof $staticType) || is_subclass_of($staticType, Asset::class);
     }
 
     /**


models/Asset.php
```diff
@@ -236,13 +236,8 @@ public static function getByPath($path, $force = false)
...
     {
         $staticType = get_called_class();
-        if ($staticType != Asset::class) {
-            if (!$asset instanceof $staticType) {
-                return false;
-            }
-        }
 
-        return true;
+        return ($staticType === Asset::class) || ($asset instanceof $staticType) || is_subclass_of($staticType, Asset::class);
```
dvesh3: what about situation when user requests wrong element from asset class e.g. `$asset = Pimcore\Model\Asset\Image::getById(1); // ID 1 is of folder type` introducing `is_subclass_of($staticType, Asset::class)` will return mismatched element

brusch:
closed - see explanation from @dvesh3

PR #4113, opened 2019-03-21
fixed #4105 - [DataObjects] (Advanced)ManyToMany(Objects) do not work…

### LINKED ISSUE #4105: [DataObjects] (Advanced)ManyToMany(Objects) do not work within localized fields
adding new elements in pimcore backend does not work - after saving and reloading object data is lost again

fashxp:
… within localized fields

fixed #4105


weisswurstkanone:
root cause:
https://github.com/pimcore/pimcore/blob/5f01109d02f13595d895de341e34842372663a1c/models/DataObject/Localizedfield.php#L172

PR #4122, opened 2019-03-23
Optimize relation loading in admin

### LINKED ISSUE #4083: [DataObjects] Optimize Loading Performance for AdvancedManyToManyObject Relation
optimize loading performance for many-to-many object relation when 1000ends of relations are assigned in one object. 

Potentially necessary steps: 

- make object in [`ObjectMetadata`](https://github.com/pimcore/pimcore/blob/3f2ddc27c60f791a48d3791006a90d8a12e577af/models/DataObject/Data/ObjectMetadata.php#L26) lazyloaded - store object id and only load object when `getObject` is called and object not yet loaded. 
- also modify corresponding dao accordingly, especially [`load` method](https://github.com/pimcore/pimcore/blob/cc229fa61180a4d15109161eb7780702579e423e/models/DataObject/Data/ObjectMetadata/Dao.php#L81)
- modify [`loadData`](https://github.com/pimcore/pimcore/blob/16a3370d1812addc8491a2a3ba8d9dc12074fd5a/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php#L97) method in data type implementation that is doesn't load destination object 

These changes should allow it that when calling a relation getter on the object, not all destination objects are loaded right away. 

Optimizations for Pimcore Admin: 
- optimize [`DataObjectController`](https://github.com/pimcore/pimcore/blob/7bcafb6b1dd4773109c37cc867688e17873e2d93/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L559) so that advanced relations are treated like normal relations (and loaded directly from relations table) and meta data is not loaded right away
- add mechanism to load object attributes (if necessary) and meta data asynchronously similar to nice path helper

Possibly these optimizations are also possible for AdvancedManyToMany relations.

fashxp:
adds async load option for meta data and object fields to (Advanced-)Many-To-Many(-Object) data types
image

partly fixes #4083


bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/advancedManyToManyObjectRelation.js
```diff
@@ -188,13 +188,22 @@ pimcore.object.classes.data.advancedManyToManyObjectRelation = Class.create(pimc
             name: "enableBatchEdit",
             value: this.datax.enableBatchEdit
         });
+
+        console.log(this);
```
dpfaffenbauer: debug ;)

fashxp: damn, thx

models/DataObject/ClassDefinition/Data/ManyToManyRelation.php
```diff
@@ -1046,4 +1046,12 @@ public function getDiffDataFromEditmode($data, $object = null, $params = [])
 
         return;
     }
+
+    /**
+     * @return bool
+     */
+    public function getEnableAdminAsyncLoad()
```
brusch: @fashxp Do we really want that enabled by default without having a choice? 
I think it would make sense to put the changes on `ManyToManyObjectRelation` into a trait and re-use it here and maybe also for `ReverseManyToManyObjectRelation`.

brusch: Since the frontend/JS part is there, I assume that's not intentional ... I'm going to fix that.

PR #4128, opened 2019-03-25
[Assets] exclude asset folder from tree preview

ssiergl:
Listing returned folders as well which results in empty tree preview on
instances with nested folder structures.

Fix it with limiting the
listing to asset types which actually have a thumbnail.

## Additional info  

To reproduce....let's assume a nested structure like this:

image

As you see there are a lot of folders which finally have images in it. Hovering over the Artikel folder is showing just 5 tree preview images, although there are a lot more present.

Due that we fetch as well folders in our query and filter out assets afterwards, which are not a folder, we get this poor result.

Hence query just for assets which really have a thumbnail and show the first 35 we get


ssiergl:
@weisswurstkanone this patch is not the reason for build breakage. The 7.3 php archive download timed out

weisswurstkanone:
just restarted the build

ssiergl:
thanks :+1:

brusch:
Thanks!

PR #4131, opened 2019-03-26
[Deployment] add customlayout cli import

ssiergl:
Resolves 

see: https://talk.pimcore.org/t/custom-layouts-via-bundle-installation/1851/2

## Additional info  

Ability to import custom layouts through cli. You simply need to specify target class and filename to import.
- If not present it will be created.
- If present it will be updated

### DIFF
--- a/models/DataObject/ClassDefinition/CustomLayout.php
+++ b/models/DataObject/ClassDefinition/CustomLayout.php
@@ -119,6 +119,28 @@ class CustomLayout extends Model\AbstractModel
     }
 
     /**
+     * @param string $name
+     *
+     * @return null|CustomLayout|mixed
+     */
+    public static function getByName(string $name)
+    {
+        try {
+            $customLayout = new self();
+            $id = $customLayout->getDao()->getIdByName($name);
+            if ($id) {
+                return self::getById($id);
+            } else {
+                throw new \Exception('There is no customlayout with the name: ' . $name);
+            }
+        } catch (\Exception $e) {
+            Logger::error($e);
+        }
+
+        return null;
+    }
+
+    /**
      * @param string $field
      *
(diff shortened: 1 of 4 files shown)


ssiergl:
@fashxp that's the way we are integrating this feature in our plugins. What do you think?

models/DataObject/ClassDefinition/CustomLayout.php
```diff
@@ -118,6 +118,35 @@ public static function getById($id)
...
+            throw new \Exception('Custom layout name is null');
+        }
+
+        try {
+            $customLayout = new self();
+            $customLayout->getDao()->getByName($name);
+
+            if ($customLayout->getId()) {
+                DataObject\Service::synchronizeCustomLayout($customLayout);
+                \Pimcore\Cache\Runtime::set($customLayout->getId(), $customLayout);
```
fashxp: this is dangerous because `id` is not unique systemwide, isn't it? 

in general, I would suggest that we implement the getByName method in a similar way as e.g. in ClassDefinition here https://github.com/pimcore/pimcore/blob/5d373e1085dc0f5e17306d5301a3ed90f63e9d00/models/DataObject/ClassDefinition.php#L224 ...
just get the id for a name and then call getById(). so we would avoid code duplication and also inconsistencies like the cache id.

ssiergl: I will check it tomorrow. Sorry for the late reply.....pretty busy right now due of high workload in my job.

fashxp: no worries.

ssiergl: Done

fashxp:
thank you very much

PR #4132, opened 2019-03-26
[DataObjects] Optimize Loading Performance for AdvancedManyToManyObject Relation

### LINKED ISSUE #4083: [DataObjects] Optimize Loading Performance for AdvancedManyToManyObject Relation

weisswurstkanone:
Resolves  #4083

### DIFF
--- a/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
+++ b/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
@@ -97,35 +97,60 @@ class AdvancedManyToManyObjectRelation extends ManyToManyObjectRelation
     public function loadData($data, $object = null, $params = [])
     {
         $ownerObject = $object;
-        $objects = [];
+        $list = [];
 
         if (is_array($data) && count($data) > 0) {
-            foreach ($data as $object) {
-                $source = DataObject::getById($object['src_id']);
-                $destination = DataObject::getById($object['dest_id']);
+            $db = Db::get();
+            $targets = [];
 
-                if ($source instanceof DataObject\Concrete && $destination instanceof DataObject\Concrete && $destination->getClassName() == $this->getAllowedClassId()) {
+            foreach ($data as $relation) {
+                $targetId = $relation['dest_id'];
+                $targets[] = $targetId;
+            }
+
+            $existingTargets = $db->fetchCol(
+                'SELECT o_id FROM objects WHERE o_id IN ('.implode(',', $targets).')'
+            );
+
+            foreach ($data as $relation) {
+                if ($relation['dest_id']) {
+                    $destinationId = $relation['dest_id'];
+
+                    if (!in_array($destinationId, $existingTargets)) {
+                        // destination object does not exist anymore
+                        continue;
+                    }
+
                     /** @var $metaData DataObject\Data\ObjectMetadata */
                     $metaData = \Pimcore::getContainer()->get('pimcore.model.factory')
                         ->build('Pimcore\Model\DataObject\Data\ObjectMetadata', [
                             'fieldname' => $this->getName(),
                             'columns' => $this->getColumnKeys(),
-                            'object' => $destination
+                            'object' => null
                         ]);
 
                     $metaData->setOwner($ownerObject, $this->getName());
+                    $metaData->setObjectId($destinationId);
 
-                    $ownertype = $object['ownertype'] ? $object['ownertype'] : '';
-                    $ownername = $object['ownername'] ? $object['ownername'] : '';
-                    $position = $object['position'] ? $object['position'] : '0';
+                    $ownertype = $relation['ownertype'] ? $relation['ownertype'] : '';
+                    $ownername = $relation['ownername'] ? $relation['ownername'] : '';
+                    $position = $relation['position'] ? $relation['position'] : '0';
 
-                    $metaData->load($source, $destination, $this->getName(), $ownertype, $ownername, $position);
-                    $objects[] = $metaData;
+                    $metaData->load(
+                        $ownerObject,
+                        $relation['dest_id'],
+                        $this->getName(),
+                        $ownertype,
+                        $ownername,
+                        $position
+                    );
+
+                    $list[] = $metaData;
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


models/DataObject/Data/ObjectMetadata.php
```diff
@@ -54,8 +55,26 @@ class ObjectMetadata extends Model\AbstractModel implements DataObject\OwnerAwar
     public function __construct($fieldname, $columns = [], $object = null)
     {
         $this->fieldname = $fieldname;
-        $this->object = $object;
         $this->columns = $columns;
+        $this->setObject($object);
```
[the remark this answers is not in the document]
fashxp: saving an object with lot of relations seems still quite expensive, event if the relations didn't change.

models/DataObject/Data/ObjectMetadata/Dao.php
```diff
@@ -78,17 +78,14 @@ protected function getTablename($object)
      *
      * @return null|Model\Dao\\Pimcore\Model\DataObject\AbstractObject
      */
-    public function load(DataObject\Concrete $source, $destination, $fieldname, $ownertype, $ownername, $position, $type = 'object')
+    public function load(DataObject\Concrete $source, $destinationId, $fieldname, $ownertype, $ownername, $position, $type = 'object')
```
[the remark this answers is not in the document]
weisswurstkanone: will clean that up, it is there for ages already

models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
```diff
@@ -114,38 +113,42 @@ public function loadData($data, $object = null, $params = [])
...
-                        $ownertype,
-                        $ownername,
-                        $position
-                    );
-
-                    $list[] = $metaData;
+                    if ($source instanceof DataObject\Concrete) {
+                        /** @var $metaData DataObject\Data\ObjectMetadata */
+                        $metaData = \Pimcore::getContainer()->get('pimcore.model.factory')
+                            ->build('Pimcore\Model\DataObject\Data\ObjectMetadata', [
```
dpfaffenbauer: `ObjectMetadata::class`?

models/DataObject/Data/ObjectMetadata.php
```diff
@@ -110,9 +129,9 @@ public function save($object, $ownertype = 'object', $ownername, $position)
      *
      * @return mixed
      */
-    public function load(DataObject\Concrete $source, $destination, $fieldname, $ownertype, $ownername, $position)
+    public function load(DataObject\Concrete $source, $destinationId, $fieldname, $ownertype, $ownername, $position)
```
blankse: Should be also changed in doc block

models/DataObject/Data/ObjectMetadata/Dao.php
```diff
@@ -78,17 +78,14 @@ protected function getTablename($object)
      *
      * @return null|Model\Dao\\Pimcore\Model\DataObject\AbstractObject
      */
-    public function load(DataObject\Concrete $source, $destination, $fieldname, $ownertype, $ownername, $position, $type = 'object')
+    public function load(DataObject\Concrete $source, $destinationId, $fieldname, $ownertype, $ownername, $position)
```
blankse: Should be also changed in doc block

models/DataObject/Data/ObjectMetadata.php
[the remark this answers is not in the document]
weisswurstkanone: the dirty map gets reset when loading is finished, so there isn't an issue

models/DataObject/Data/ObjectMetadata/Dao.php
weisswurstkanone: Thx, incorporated that!

blankse: type can be removed in doc block ;)

models/DataObject/Data/ObjectMetadata.php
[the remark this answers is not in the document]
fashxp: ok, saving is expensive because getter is called in Object\Concrete `update()` method, see also https://github.com/pimcore/pimcore/issues/4154 
has nothing to do with this PR, there everything is good.

models/DataObject/Data/ObjectMetadata/Dao.php
fashxp: done

models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
fashxp: done

PR #4137, opened 2019-03-27
TranslateValue-Operator does not translate "0"

woellundwoell:
When value to translate is String "0" the condition in line 44 will be false. So no prefix is set and no translation is executed.


weisswurstkanone:
thanks a lot!

Issue #1023, opened 2016-11-08
[Objects] Quantity Value: Do not store Unit IDs in class definitions


solverat:
when using units in class configuration, the definition stores unit Ids as reference:

![bildschirmfoto 2016-11-08 um 10 15 50](https://cloud.githubusercontent.com/assets/700119/20093295/717451ee-a59c-11e6-8368-02ec396f2c97.png)

Because of that, we can't share definitions in different environments. The dev DB is not always in sync with the staging or production DB. So we might have different Unit Ids.

It would be great to have a string reference, like the linked objectbricks for example:

![bildschirmfoto 2016-11-08 um 10 19 36](https://cloud.githubusercontent.com/assets/700119/20093394/db26c752-a59c-11e6-8b5d-14b9d94c071a.png)

cleggypdc:
+1 for this experience too.

weisswurstkanone:
Problem here is that there is no other mandatory unique field (other than the internal ID) that could identify the unit. Not even the abbreviation.

cleggypdc:
How about storing the classId in the php config file too, call it originalClassId. Then at least we can build a sanity check into either class deployment or custom deployments to check that the ID's map. It would make debugging easier :-)

solverat:
Because we can't share the unit id between environments, it's quite impossible to use them. The *quantity value* requires "valid units" and can't be empty. With empty fields, the unit will be stored with the given id: 

![bildschirmfoto 2016-12-07 um 10 50 38](https://cloud.githubusercontent.com/assets/700119/20962905/8f2bf17a-bc6b-11e6-8a05-2fef4ef38b26.png)

So, either allow empty "valid units" or even better: store units in class configuration as string or something like that.

brusch:
Proposed solution: 
- Make `id` column initially editable (not editable after it was initially set)
- change `id` columns to a `VARCHAR` in db schema (in definition table and data columns)
- make it visible per default (in grid). 
- save all classes, object bricks, field collections in update script

BlackbitNeueMedien:
@brusch, @weisswurstkanone: Is there a reason why you do not want to use the abbreviation as unique field? I actually do not know of any unit abbreviation which as different meanings in different contexts, do you?

brusch:
@BlackbitNeueMedien not really ... at least from my side.

rubanooo:
@brusch any updates on this issue?

blankse:
[...]

PR #4138, opened 2019-03-27
Quantity Value Unit conversion

### LINKED ISSUE #3529: Quantity value fields: Add conversion rates between different units
# Feature Request 

Currently units of a quantity value field are more or less only for viewing purposes. I saw preparations for unit conversion in \Pimcore\Model\DataObject\QuantityValue\Unit: there are the fields baseunit, reference, factor, conversionOffset. Currently these fields are not editable via Pimcore GUI. Are there plans to implement unit conversion in the near future or did I miss something and this is possible yet today?

BlackbitNeueMedien:
When using quantity values it was only possible to insert data in different units but now to convert between them. This PR adds the possibility to set base unit and conversion factor for quantity value units and also adds the method QuantityValue::convert(Unit $unit) to convert quantity value objects between units.

Resolves #3529


bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
```diff
@@ -87,14 +87,47 @@ pimcore.object.quantityValue.unitsettings = Class.create({
...
+
+        var baseUnitEditor = {
+            xtype: 'combobox',
+            queryMode: 'local',
+            valueField: 'id',
+            displayField: 'abbreviation',
+            forceSelection: true,
+            store: baseUnitStore
+        };
+
```
weisswurstkanone: is there a way to null it out ?

models/DataObject/Data/QuantityValue.php
```diff
@@ -104,6 +104,42 @@ public function getValue()
...
+
+        $toBaseUnit = $toUnit->getBaseunit();
+        if($toBaseUnit === null) {
+            $toUnit = clone $toUnit;
+            $toBaseUnit = $toUnit;
+            $toUnit->setFactor(1);
+            $toUnit->setConversionOffset(0);
+        }
+
+        if($fromBaseUnit === null || $toBaseUnit === null || $fromBaseUnit->getId() !== $toBaseUnit->getId()) {
```
weisswurstkanone: is this even possible ?

$toBaseUnit === null

ultramegatom:
Would an approach similar to what https://github.com/pimlie/php-unit-conversion does also be an option?

Right now this seems like a somewhat specific use case to me (only convert from a specific unit to its base unit).

Don't get me wrong here. Right now die QuantityValue is only storing values and has no logic whatsoever and your approach certainly improves the datatype. 

But i think if we add conversion functionality to the datatype we should have an API for parsing/converting/outputting units. 

This would of course be a complete refactor of the current datatype... so maybe a followup task?

what do you think @weisswurstkanone ?

bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
BlackbitNeueMedien: Added an empty option.

models/DataObject/Data/QuantityValue.php
BlackbitNeueMedien: Of course this is possible. When the target unit does not have a base unit.
For example when you have an object with a quantity value in unit "cm". 
And you have configured units m, cm and mm. base unit of cm and mm is m. m does not have a base unit.
When you want to convert cm to m, toUnit is m and it does not have a base unit. Then it is assumed that the target unit has itself as base unit.

BlackbitNeueMedien:
> Right now this seems like a somewhat specific use case to me (only convert from a specific unit to its base unit).

This PR does not only support converting to the base unit. 
Example:
When you have 3 units: m, cm and mm
You have a quantity value of 3 cm in your object.

You can then call $quantityValue->convert(Unit::getByAbbreviation('mm')); and it will give you a QuantityValue object of 30 mm. 
When converting the source value is converted to the base unit first and then to the target unit. In practice it is rcommended to use SI units as base units. All units which share the same base unit can be converted to each other.

BlackbitNeueMedien:
Are there any remaining problems with this PR @weisswurstkanone ?

brusch:
I'd prefer a more generic way ... what about configuring a conversion service that takes care of the calculation? Within the service you can then of course also use your preferred library, or just do simple factor based calculations.

BlackbitNeueMedien:
Do you mean a service which you inject the actual converter? By default a converter which is based on quantity value unit conversion factors and offsets (as configured in Pimcore backend) gets injected to this service - but you could also provide a converter based on any conversion library like https://github.com/pimlie/php-unit-conversion as @ultramegatom suggested.
This would also make it easy to use a currency conversion library which is based on official currency exchange rates API.
And then you would have Service::convert($quantityValue, $unit):QuantityValue? 

This would also separate available quanitity value units when editing data object's data from available target units (you could allow only to insert "meter" or "EUR" in data objects but be able to output "feet" and "USD". No idea if this is necessary but there will certainly be cases where you want to limit the available units in data editing but be able to convert to much more units in outputs.

brusch:
@BlackbitNeueMedien What I mean is an approach similar to what we do already with the _option provider service_ for select data-types. So a service that needs to implement a certain interface and that can be used as for transforming from one unit to another. 

I'm thinking of something link that: 
asd


```php
<?php

namespace AppBundle\DataObject\QuantityValue;

class MeterConverter implements QuantityValueConverterInterface
{
    const PREFIXES = [
        'k' => .001,
        'm' => 1000,
        'c' => 100,
        'd' => 10,
    ];

    public function convert(string $from, string $to, $value) {

        if(strlen($from) === 2) {
            $prefixFrom = substr($from, 0, 1);
            $value = $value / self::PREFIXES[$prefixFrom];
        };

        if(strlen($to) === 2) {
            $prefixTo = substr($to, 0, 1);
            $value = $value * self::PREFIXES[$prefixTo];
        };

        return $value;
    }
}
```


```php
<?php

namespace Pimcore\Model\DataObject\QuantityValue;

interface QuantityValueConverterInterface
{
    public function convert(string $from, string $to, $value); 
}
```

brusch:
Maybe we should also consider #1023 in this regard, a manually defined unique identifier would be quite helpful when working with a conversion service...

BlackbitNeueMedien:
Imho, only configuring unit conversion via programmed services is too geeky in my opinion. Most customers only want to convert metrical units to imperial for example (and they will look angry if they have to pay a programmer only for this simple, recurring taks). This is why I added a "convereter" column for each unit. The converter is a service class which has to implement the QuantityValueConverterInterface. If the converter is omitted the DefaultConverter (which is part or Pimcore and this PR) gets used. This DefaultConverter uses the configured conveersion factors / offsets.
This way you can use your own converter service but for simple cases can entry conversion factors in the Pimcore backend.

bundles/CoreBundle/Resources/migrations/Version20190408084129.php
```diff
@@ -14,6 +14,7 @@ public function up(Schema $schema)
     {
         $this->addSql("ALTER TABLE `quantityvalue_units`
             CHANGE `baseunit` `baseunit` INT(11) UNSIGNED,
+            ADD `converter` VARCHAR(255),
```
weisswurstkanone: please don't forget to add this to install.sql

BlackbitNeueMedien: Thanks for your hint. Added database changes to install.sql

bundles/InstallBundle/Resources/install.sql
```diff
@@ -865,11 +865,17 @@ CREATE TABLE `quantityvalue_units` (
   `group` varchar(50) DEFAULT NULL,
   `abbreviation` varchar(20) NOT NULL,
   `longname` varchar(250) DEFAULT NULL,
-  `baseunit` varchar(10) DEFAULT NULL,
+  `baseunit` `baseunit` INT(11) UNSIGNED DEFAULT NULL,
```
weisswurstkanone: I guess this is not right

BlackbitNeueMedien: Why not? The column baseunit contains the id of the base unit. If you do not set a base unit then it is null.

weisswurstkanone:
please note that the unit tests are still failing. please have a look at my comment before.

bundles/InstallBundle/Resources/install.sql
BlackbitNeueMedien: Oh sorry, my fault. Is fixed now

BlackbitNeueMedien:
Sorry, now tests are green.

brusch:
@BlackbitNeueMedien are you done already?

BlackbitNeueMedien:
@brusch yes, from my point of view this is ready to be checked / merged.

weisswurstkanone:
can we add a short section to the docs ?

BlackbitNeueMedien:
@weisswurstkanone of course but who is "we"? ;-)

BlackbitNeueMedien:
@weisswurstkanone I will add some documentation to https://github.com/pimcore/pimcore/blob/master/doc/Development_Documentation/05_Objects/01_Object_Classes/01_Data_Types/03_Number_Types.md#quantity-value

BlackbitNeueMedien:
Documentation done.

weisswurstkanone:
*) I am experiencing the following problem when I change the unit dropdown (tag)

Message: Type error: Return value of Pimcore\Model\DataObject\QuantityValue\Unit::getConverter() must be of the type string, null returned

*) Not sure if conversion should be done automatically or better be an option in the class definition

*) Oberservation: network panel indicates that there are more requests than needed to add a unit definition

BlackbitNeueMedien:
The error should be fixed now.

I will take a look at network panel when adding a unit...

BlackbitNeueMedien:
I checked the network tab when editing units. The calls originate from 3 stores which are used: one for the unit configuration itself, one for the base units combobox (I cannot use the same store as for the unit configuration because I need an "empty" option) and one for the data object dropdown (tag) - if it is still loaded.

BlackbitNeueMedien:
Added field configuration for auto-conversion

BlackbitNeueMedien:
Are there any problems remaining, @weisswurstkanone / @brusch ?

weisswurstkanone:
thanks a lot!

PR #4140, opened 2019-03-27
Fix classificationstore key tooltips

### LINKED ISSUE #4139: Tooltips in classificationstore keys not applied
# Bug Report

If you configure a tooltip for a classificationstore key it is not showing in the object instance.

image

jremmurd:
Configured tooltips of classificationstore keys get overriden and so are not displayed in the objects. This PR fixes the issue and uses the previous value as fallback.

Resolves #4139


weisswurstkanone:
thx!

PR #4141, opened 2019-03-27
Bugfix for wrong limit / missing start param in grid CSV export.

### LINKED ISSUE #4181: CSV Export get only 20 first objects
When use CSV export in Grid View, document contain only 20 first objects. The issue is reproducable at https://demo-advanced.pimcore.org/admin

Think, that solution can be connected or same as https://github.com/pimcore/pimcore/pull/4168/commits/c16c4f559626ecd87c9cb1b76d42892502c6641b

woellundwoell:
Bugfix for wrong limit / missing start param in grid CSV export.

### Expected behaviour
Export the number of records selected in "Elements per Page" or selected objects per checkbox in object grid view (folder or children)

### Actual behavior
Export is limited to 20 datasets in CSV export, even if all checkboxes (>20) are selected.

### Reproduce bug
1. Select folder (or children) view in object tree
2. Select object type
3. Click "CSV Export" button whith more than 20 datasets
4. CSV export file will contain max. 20 records

Resolves #4181

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js
@@ -502,6 +502,13 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
 
         var fields = this.getGridConfig().columns;
         var fieldKeys = Object.keys(fields);
+        
+        var limit = pimcore.helpers.grid.getDefaultPageSize(-1);
+        var start = 0;
+        if(this.store.lastOptions) {
+            limit = this.store.lastOptions.limit;
+            start = this.store.lastOptions.start;
+        }
 
         //create the ids array which contains chosen rows to export
         ids = [];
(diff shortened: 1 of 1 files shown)


dvesh3:
@woellundwoell sorry, this is incomplete solution as export only works if you are on first page of grid.

woellundwoell:
OK, thank you very much! But I think we should make the change in PR #4144 and set $start only when $ignoreLimit is not set:
https://github.com/pimcore/pimcore/blob/e9b524a075eabb9c9c2ece7f613f40ab6a9b6d79/bundles/AdminBundle/Helper/GridHelperService.php#L423-L428
to
<pre>
        if($ignoreLimit) {
        	$limit = null;
        }
        if ($requestParams['start'] && !$ignoreLimit) {
            $start = $requestParams['start'];
        }
</pre>
What do you think?

dvesh3:
let's change #4144 first

dvesh3:
Resolved by #4144

brusch:
Resolved by #4144

PR #4142, opened 2019-03-27
Change mail log / redirect handling to retain the original message data

das-peter:
Currently the mail redirecting in the mail debug mode leads to "information loss" in the email log.
The receivers are not retained and hence the log can't be searched for mails to a certain receiver. 

1. `Pimcore\Mail\Plugins\RedirectingPlugin` has been adjusted to put the receiver & sender information in the `originalData` property and restore the information once the mail is sent. It also takes care to inject properties to the message which provide information about the redirection that happened.
2. `Pimcore\Mail::sendWithoutRendering()` has been adjusted to not overwrite receiver information when the mail debug mode is enabled (this should be task of whoever handles the receiver manipulation anyways, right?).


brusch:
But the email was actually sent to the debug addresses and should be therefore logged as such.

das-peter:
@brusch  fair point. How about switching the approach then and logging the original data as message parameters? That still keeps the log searchable for the original data which is my ultimate goal here.

Background: I've extended the REST service to be able to fetch a sent mail by receiver - allowing me to create automated testing of our password-reset process.

brusch:
@das-peter message parameters sound ok for me, can you create a new PR for that? 
Thanks!

das-peter:
Erm, the branch was already updated to reflect the changes :/ But sure, I can re-branch and create another pull-request

brusch:
Oh, sorry haven't seen that and was not mentioned in the comments 😐 Anyway, thanks for the new PR 😉

PR #4144, opened 2019-03-28
Feature "Export all" in grid CSV export

woellundwoell:
# Feature request "Export all checkbox" in grid CSV export

In the current version, the CSV export in the data grid is set to "sets per page". If you want to export larger amounts of data (> 200) via CSV, all data must first be loaded in the browser view, which can lead to enormous performance problems.
The new checkbox in the export dialog "export all" allows an export of all data (without limit) of the current table view.

## Dependencies
Need merge of PR #4141


bundles/CoreBundle/Resources/translations/de.json
```diff
@@ -647,6 +647,8 @@
     "resolver_strategy": "Resolver Strategy",
     "resolver_settings": "Resolver Settings",
     "csv_settings": "CSV Einstellungen",
+    "export_settings": "Export Einstellungen",
```
brusch: Only `EN` translations are needed, the German translations are made here after this PR was merged: https://poeditor.com/join/project/VWmZyvFVMH please revert this changes :)

woellundwoell:
Reverted translation file ;)

brusch:
@woellundwoell the new-line seems to be still there and is causing a conflict: 

image


Thanks again!

brusch:
Thanks looks way better, thanks!

dvesh3:
@woellundwoell we don't need extra config. If no record is selected in grid then export click should export all records.

woellundwoell:
I think it will not behave that way. The limit is hard coded / intitialized in 
https://github.com/pimcore/pimcore/blob/e9b524a075eabb9c9c2ece7f613f40ab6a9b6d79/bundles/AdminBundle/Helper/GridHelperService.php#L402
and will not be set / overwritten without any data from request in
https://github.com/pimcore/pimcore/blob/e9b524a075eabb9c9c2ece7f613f40ab6a9b6d79/bundles/AdminBundle/Helper/GridHelperService.php#L420-L422

dvesh3:
how about adding extra param `export:true` here 
https://github.com/pimcore/pimcore/blob/015c27b6e9c3a4a004723b20487fb2e44408a41d/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js#L515

and then only set offset & limit is export param is NOT set or false:
https://github.com/pimcore/pimcore/blob/e9b524a075eabb9c9c2ece7f613f40ab6a9b6d79/bundles/AdminBundle/Helper/GridHelperService.php#L518

can you try this way?

woellundwoell:
Based on the desire of our customer to have a definite control, we implemented the feature in this way. But you are absolutely right. So we will change the PR as soon as possible.

woellundwoell:
When exporting but also possibly set ids (checkboxes in the grid view) should be considered. We have now taken this into account. OK like this?

dvesh3:
looks good 👍

dvesh3:
can you also add search query in export params as right now it is ignored in export:

Add `var searchQuery = this.searchField.getValue();` here:
https://github.com/pimcore/pimcore/blob/bd765d0703a89fa4e3096c1aca177c68f49d720c/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js#L493

and `query: searchQuery,` here:
https://github.com/pimcore/pimcore/blob/bd765d0703a89fa4e3096c1aca177c68f49d720c/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js#L525

Then we are ready to merge 👍

woellundwoell:
Here it is. ;)

brusch:
Resolved by #4144

PR #4149, opened 2019-03-28
Remove lazyKeys from Object Serialization

solverat:
Remove `lazyKeys` from object serialization. Otherwise, lazy fields will remain in a NULL state "randomly". We experienced several side effects and this took us hours of debugging! 

BTW: Pimcore needs way **more** **tests**, so we don't have to get to this point in the first place....

### DIFF
--- a/models/DataObject/AbstractObject.php
+++ b/models/DataObject/AbstractObject.php
@@ -1226,12 +1226,12 @@ class AbstractObject extends Model\Element\AbstractElement
 
         if (isset($this->_fulldump)) {
             // this is if we want to make a full dump of the object (eg. for a new version), including childs for recyclebin
-            $blockedVars = ['o_userPermissions', 'o_dependencies', 'o_hasChilds', 'o_versions', 'o_class', 'scheduledTasks', 'o_parent', 'omitMandatoryCheck', 'o_dirtyFields'];
+            $blockedVars = ['o_userPermissions', 'o_dependencies', 'o_hasChilds', 'o_versions', 'o_class', 'scheduledTasks', 'o_parent', 'omitMandatoryCheck', 'o_dirtyFields', 'lazyKeys'];
             $finalVars[] = '_fulldump';
             $this->removeInheritedProperties();
         } else {
             // this is if we want to cache the object
-            $blockedVars = ['o_userPermissions', 'o_dependencies', 'o_childs', 'o_hasChilds', 'o_versions', 'o_class', 'scheduledTasks', 'o_properties', 'o_parent', 'o___loadedLazyFields', 'omitMandatoryCheck'];
+            $blockedVars = ['o_userPermissions', 'o_dependencies', 'o_childs', 'o_hasChilds', 'o_versions', 'o_class', 'scheduledTasks', 'o_properties', 'o_parent', 'o___loadedLazyFields', 'omitMandatoryCheck', 'lazyKeys'];
         }
 
         foreach ($parentVars as $key) {


solverat:
Also: Please provide a new patch release as soon as possible!

dpfaffenbauer:
I am backing up the tests comment ;)

fashxp:
everybody is welcome to contribute more tests ;-) 
e.g. could you provide some tests that test for the side effects you experienced.

brusch:
This is unfortunately not the solution, it actually would make makes things even worse. 
`lazyKeys` stores the keys of the fields which were not loaded yet from the backend, so let's assume the following case: 
- Class `MyObject` has a lazy loaded many-to-many relation field called `categories` 
- `$object1 = MyObject::getById(123)`; -> not loaded from cache, `categories`  contains 3 relations
- `$serializedData = serialize($object);`
- `$object2 = unserialize($serializedData );`
- `$object2->getCategories();` -> empty because it doesn't get loaded from the backend (missing key in `lazyKeys`

dpfaffenbauer:
@brusch But exactly that is solved by that. Once you serialize $object, Pimcore`null's` all lazy loaded fields, but it doesn't reset the state of the `lazyKeys`. Thus you will get null on the unserialize.

@fashxp But it's not everybody responsability to contribute tests. No one will actually invest many hours just to create tests. If you change something in the core, you should also automatically test that, or provide tests that proof the things you change are working. Pimcore already is so big, you cannot test everything manually anymore. @maff did a good job at his time providing tests for things he changed or added. But since @maff is gone, no more tests have been really added or done.

brusch:
@dpfaffenbauer but `lazyKeys` does contain the fields which were **NOT** loaded!

dpfaffenbauer:
Exactly, thats the issue. 

Following two scenarios:

 - I load Product
 - I call lazy loaded field 'categories'
 - Then lazyKeys should contain 'categories'
 - I serialize product
 - I unserialize Product
 - Then categories should be null
 - I call lazy loaded field 'categories'
 - Then categories should not be null: BUT IT IS!

Second scenario:

 - I load Product
 - I serialize product
 - I unserialize Product
 - Then categories should be null
 - I call lazy loaded field 'categories'
 - Then categories should not be null: AND IT IS!

Pimcore doesn't serialize ALL lazy loaded fields, no matter if it has been loaded or not: https://github.com/pimcore/pimcore/blob/master/models/DataObject/Concrete.php#L718

So, all Lazy Loaded fields are null after `unserialize`: Reason: Cause the ObjectTags still see the lazyKeys cause they never get reset.

brusch:
@dpfaffenbauer 
> Then lazyKeys should contain 'categories'

That's the point ... after you've called the getter it doesn't contain `categories` anymore. 
See: https://github.com/pimcore/pimcore/blob/16a3370d1812addc8491a2a3ba8d9dc12074fd5a/models/DataObject/ClassDefinition/Data.php#L1371

dpfaffenbauer:
How is that supposed to work? I thought that the lazyKeys contain all loaded fields? Why remove them on getter when it is loaded?

brusch:
No, they contain the fields which were not yet loaded, that's what I wrote several times ;) 
But yes, there is still an issue, but this PR doesn't solve it.

dpfaffenbauer:
Wasn't it the other way around before? So Pimcore needs to add all those keys to the store before it does anything else? Why not the other way around, I guess that would make more sense. Add the key once it's loaded.

brusch:
Yes, it was that way before. 
Either way is fine, but we're just checking what else is wrong here ;-)

dpfaffenbauer:
:confused: Well... thats a very good note for Upgrade Notes... every custom lazy loaded object tag is now broken cause of that.

dpfaffenbauer:
But, since it contains keys that are not loaded yet, why do you add them in the `load` function:

https://github.com/pimcore/pimcore/blob/16a3370d1812addc8491a2a3ba8d9dc12074fd5a/models/DataObject/Objectbrick/Dao.php#L82
https://github.com/pimcore/pimcore/blob/16a3370d1812addc8491a2a3ba8d9dc12074fd5a/models/DataObject/Fieldcollection/Dao.php#L86
https://github.com/pimcore/pimcore/blob/16a3370d1812addc8491a2a3ba8d9dc12074fd5a/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php#L344
https://github.com/pimcore/pimcore/blob/16a3370d1812addc8491a2a3ba8d9dc12074fd5a/models/DataObject/ClassDefinition/Data/Fieldcollections.php#L337

That doesn't make any sense, that tells me that its now loaded...

brusch:
An object brick/field collction contains a lazy loaded relation, on load, the brick marks this relational field as "to be loaded" by calling ` $this->model->addLazyKey($lazyKey); `

See: 
image

dpfaffenbauer:
so you actually never know if the field has ever been loaded or is loaded? Cause, non of the lazy loaded fields are pre-filled into lazyKeys, cause it just says "to be loaded".

solverat:
That's truly a brainf*k but @brusch is right, this PR is obsolete since you've switched the query `hasLazyKey` vs `!hasLazyKey`... but still, there _is_ something wrong, we have several nulled manyToOne relations out of nowhere.

brusch:
Yes, there's actually something wrong (like I said already ...) and we're investigating it.

brusch:
Just for reference,  #4151 is the follow up PR for this topic.

PR #4150, opened 2019-03-28
Use flex instead of fixed width for columns in an advanced many-to-many relation

BlackbitNeueMedien:
Visible columns in advanced Many-to-Many object relations are 100px wide by default, often hiding parts of the content. The regular many-to-many-relation already makes use of the flex param, improving visibility.

Current view:
Bildschirmfoto 2019-03-28 um 15 07 26

When using flex:
Bildschirmfoto 2019-03-28 um 15 07 45


brusch:
Makes sense, thanks!

PR #4153, opened 2019-03-29
AssetController cannot generate thumbnails

mcuenez:
### How to reproduce
Simply, call the following url: **\<your-domain\>/admin/asset/get-image-thumbnail?id=\<your-asset-id\>&thumbnail=\<your-thumbnail-name\>&cropPercent=false**.

_Obviously, you'll have to use proper values for the URL._

### Changes made
The parameter is additionally checked for its value, ultimately, resolving the issue with this PR.


brusch:
Thanks!

PR #4157, opened 2019-03-29
[DataObjects] Fixed localized advanced many-to-many (object) relations update in object bricks &…

dvesh3:
Fixes relations data loss for non dirty languages of advanced relation fields in object bricks & field collections.

### DIFF
--- a/models/DataObject/Localizedfield/Dao.php
+++ b/models/DataObject/Localizedfield/Dao.php
@@ -150,7 +150,8 @@ class Dao extends Model\Dao\AbstractDao
                     ];
 
                     if ($fd instanceof DataObject\ClassDefinition\Data\Relations\AbstractRelations) {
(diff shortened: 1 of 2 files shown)


models/DataObject/Localizedfield/Dao.php
```diff
@@ -150,7 +150,11 @@ public function save($params = [])
                     ];
 
                     if ($fd instanceof DataObject\ClassDefinition\Data\Relations\AbstractRelations) {
-                        if ($this->model->isLanguageDirty($language) || $params['saveLocalizedRelations']) {
+                        if ( ($params['saveRelationalData']['saveLocalizedRelations']
```
brusch: I guess we need an `isset($params['saveRelationalData'])` here as well, in order to avoid notice log messages.

PR #4160, opened 2019-04-01
Fix regex if object path contains a "@"

FredoVelcro:
Resolves #

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -2016,7 +2016,7 @@ class DataObjectController extends ElementControllerBase implements EventedContr
                 $targetParent = DataObject::getById($request->get('targetParentId'));
             }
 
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -2016,7 +2016,7 @@ public function copyAction(Request $request)
                 $targetParent = DataObject::getById($request->get('targetParentId'));
             }
 
-            $targetPath = preg_replace('@^' . $sourceParent->getRealFullPath() . '@', $targetParent . '/', $source->getRealPath());
+             $targetPath = preg_replace('/^' . str_replace('/', '\/', $sourceParent->getRealFullPath()) . '/', $targetParent . '/', $source->getRealPath());
```
dpfaffenbauer: code style ;)

brusch:
We should go fir `preg_quote()` here, instead of some custom, search & replace.

PR #4164, opened 2019-04-02
[Gallery] set inline upload path of assets in datatype gallery

### LINKED ISSUE #4158: [DataObjects] inline upload of assets in datatype gallery does not consider upload-path setting
always uploaded in _default_upload-bucket

ssiergl:
Resolves #4158


brusch:
Thanks!

PR #4165, opened 2019-04-02
Skip Incomplete Classes in renewing References Process

solverat:
If a version stores incomplete (sub-) class, the [renewReferences](https://github.com/pimcore/pimcore/blob/a03da421d921a39249b76f72a1275f98e03994ab/models/Element/Service.php#L625) method fails in line:

https://github.com/pimcore/pimcore/blob/a03da421d921a39249b76f72a1275f98e03994ab/models/Element/Service.php#L663-L668

because the `method_exists` check may fails with `Error: method_exists(): The script tried to execute a method or access a property of an incomplete object.`.

This PR checks the incoming `$data` against valid classes and returns `null` otherwise.

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -624,6 +624,11 @@ class Service extends Model\AbstractModel
      */
     public static function renewReferences($data, $initial = true, $key = null)
     {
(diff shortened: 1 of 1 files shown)


brusch:
Makes sense, thanks a lot!

PR #4167, opened 2019-04-02
Use instanceof instead of get_class in Version Data Check

solverat:
This is a follow-up for #4165. `$data` could be `false` which triggers an exception in `get_class`method.

PS: Yes, I'm currently struggling with a some messed up versions. 😄


brusch:
Thanks again! ☺

Issue #4163, opened 2019-04-02
Batch edit all only updating the first 20 records


amarjeet6694:
I tried to bulk publish all my records from folder grid, but it published only the first 20 records.
The issue is reproducable at https://demo-advanced.pimcore.org/admin

brusch:
Related to #4141 and probably to c16c4f559626ecd87c9cb1b76d42892502c6641b

brusch:
Resolved by #4168

PR #4168, opened 2019-04-02
Grid: "Batch edit all" doesn't edit all or to many items

das-peter:
Using the "Batch edit all in a grid doesn't process all or to many items in the grid if the page size was adjusted.
This is because it doesn't sent the currently set page size - which causes `pimcore/vendor/pimcore/pimcore/bundles/AdminBundle/Helper/GridHelperService.php::prepareListingForGrid()` to fall back to it's default limit of 20 items.

`pimcore.object.helpers.gridTabAbstract.batchPrepare()` now sets the current pageSize as limit parameter.


dvesh3:
fixes #4163

PR #4170, opened 2019-04-03
fix admin translation javascript issues

markus-moser:
This PR fixes several issues with the current admin translations Javascript implementation:

- For objects with massive amounts of translations (for example select boxes with thousands of entries) the current call of `/admin/translation/add-admin-translation-keys` does not work. If there are massive amounts of keys the request will fail and Pimcore will show a 503 maintenance error. Additionally this call can contain the same keys multiple times. Therefore this PR limits the amount of a single request to 500 (and stores the rest for the next request). As this method is called each 30 seconds for each admin users the translation keys will be added anyway. 

- Currently not translated keys are added to the `/admin/translation/add-admin-translation-keys` call again and again. In complex systems this could be a serious performance issue. Therefore this PR checks if a key exits and not only if the key is translated.

- If case_insensitive mode is enabled the admin translations did not work correctly. This is fixed in this PR too.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/MiscController.php
+++ b/bundles/AdminBundle/Controller/Admin/MiscController.php
@@ -20,6 +20,7 @@ use Pimcore\Controller\Configuration\TemplatePhp;
 use Pimcore\Db;
 use Pimcore\File;
 use Pimcore\Tool;
+use Pimcore\Translation\Translator;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
@@ -169,7 +170,8 @@ class MiscController extends AdminController
             }
         }
 
-        $response = new Response('pimcore.system_i18n = ' . $this->encodeJson($translations) . ';');
+        $caseInsensitive = $translator instanceof Translator && $translator->getCaseInsensitive() ? "true" : "false";
+        $response = new Response('pimcore.system_i18n = ' . $this->encodeJson($translations) . ';pimcore.system_i18n_case_insensitive='. $caseInsensitive);
         $response->headers->set('Content-Type', 'text/javascript');
 
         return $response;
--- a/bundles/AdminBundle/Controller/Admin/MiscController.php
+++ b/bundles/AdminBundle/Controller/Admin/MiscController.php
@@ -170,8 +170,8 @@ class MiscController extends AdminController
             }
         }
 
-        $caseInsensitive = $translator instanceof Translator && $translator->getCaseInsensitive() ? "true" : "false";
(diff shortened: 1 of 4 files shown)


bundles/AdminBundle/Controller/Admin/MiscController.php
```diff
@@ -169,7 +170,8 @@ public function jsonTranslationsSystemAction(Request $request)
             }
         }
 
-        $response = new Response('pimcore.system_i18n = ' . $this->encodeJson($translations) . ';');
+        $caseInsensitive = $translator instanceof Translator && $translator->getCaseInsensitive() ? "true" : "false";
```
weisswurstkanone: boolean or string ?

bundles/AdminBundle/Resources/public/js/pimcore/startup.js
```diff
@@ -950,18 +950,22 @@ pimcore["intervals"]["translations_admin_missing"] = window.setInterval(function
     var missingTranslations = pimcore.globalmanager.get("translations_admin_missing");
```
weisswurstkanone: off topic:

do we currently really need 

translations_admin_missing
and
translations_admin_added

or can we can get rid of one of them ?

bundles/AdminBundle/Controller/Admin/MiscController.php
markus-moser: as it is used in the response js code it needs to be a string here

bundles/AdminBundle/Resources/public/js/pimcore/startup.js
markus-moser: I'm sure that it's possible to implement an algorithm with just a single array but this would be a separate task.

bundles/AdminBundle/Controller/Admin/MiscController.php
dpfaffenbauer: why not return true or false as json?

dpfaffenbauer: that would be a boolean then

weisswurstkanone: see my comment around an hour ago :-)

markus-moser: In the end it's a string but ok, you're right now json_encode is used.

markus-moser: It's a generated javascript code response and not a json so therefore it's a string in the PHP world and a boolean in the json world. ;)

dpfaffenbauer: now its a boolean in both worlds :=

brusch:
Thanks a lot!

PR #4173, opened 2019-04-03
Error handling and logging

till:
This is a proposal/food for thought or a drive-by PR if you will... For context, we've been trying to run Pimcore in a set of Docker containers.

# is_integer vs defined

Especially when you use environment variables to configure this setting, it's never ever going to be an integer as the code to "resolve constants" has no context/domain knowledge of the setting, doesn't cast the constant and therefor it's of type `string` and the `is_integer` is always false. You could probably use `is_numeric()` or so. But I haven't tested it.

# error_log

I changed this as well as it doesn't work when you e.g. set the `error_log` to `/proc/self/fd/2` to catch logs on `stdout` or `stderr`.

I would assume that when people set this environment variable (or the constant) they know what's going on and checking for writable is not necessary. The `is_writable` is also a hefty performance penalty each time this code is executed too. Removing this definitely fixes "my" problem, but I am also sure there are other use cases and maybe someone else has a better idea.

### DIFF
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -123,7 +123,7 @@ class Bootstrap
         $loader = include __DIR__ . '/../../../../vendor/autoload.php';
         self::defineConstants();
 
-        if (defined(PIMCORE_PHP_ERROR_REPORTING)) {
+        if (defined('PIMCORE_PHP_ERROR_REPORTING')) {
             error_reporting(PIMCORE_PHP_ERROR_REPORTING);
         }
 
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -123,14 +123,14 @@ class Bootstrap
         $loader = include __DIR__ . '/../../../../vendor/autoload.php';
         self::defineConstants();
 
-        if (is_integer(PIMCORE_PHP_ERROR_REPORTING)) {
+        if (defined(PIMCORE_PHP_ERROR_REPORTING)) {
             error_reporting(PIMCORE_PHP_ERROR_REPORTING);
         }
 
         \Pimcore::setAutoloader($loader);
         self::autoload();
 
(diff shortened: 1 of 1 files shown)


till:
Is a CLA necessary for 2 lines of code? Please let me know.

brusch:
@till yes, please. 
Just ensure you've configured your email used for the commit on GitHub.
You can check your email by running `git config user.email` in your repo and you can add additional email addresses to GitHub here: https://github.com/settings/emails

lib/Bootstrap.php
```diff
@@ -123,14 +123,14 @@ public static function boostrap()
         $loader = include __DIR__ . '/../../../../vendor/autoload.php';
         self::defineConstants();
 
-        if (is_integer(PIMCORE_PHP_ERROR_REPORTING)) {
+        if (defined('PIMCORE_PHP_ERROR_REPORTING')) {
```
brusch: This check is useless, the constant is always defined ;-) 
So I guess we can just remove the `if` condition entirely.

lib/Bootstrap.php
```diff
@@ -123,14 +123,14 @@ public static function boostrap()
...
             error_reporting(PIMCORE_PHP_ERROR_REPORTING);
         }
 
         \Pimcore::setAutoloader($loader);
         self::autoload();
 
-        if ('syslog' === PIMCORE_PHP_ERROR_LOG || is_writable(dirname(PIMCORE_PHP_ERROR_LOG))) {
+        if (defined('PIMCORE_PHP_ERROR_LOG')) {
```
brusch: Same here ...

brusch:
I've just changed the code accordingly.

lib/Bootstrap.php
till: Less code is even better! 🎉

PR #4185, opened 2019-04-04
[Thumbnails] introduce asTexture mode and fix broken setBackgroundImage method for system thumbnails

ssiergl:
This patch fixes two bugs:
 - set real image path for background image
 - use textureimage to set background image independent of background image size

## Additional info  

This method was introduced to render a background image on thumbnail generation if original image has transparencies.

First of all this method never did anything due of wrong path to `tree-preview-transparent-background.png`. We need to use the real path! So actual behavior on an image with transparency is this (white text with transparent background):

Thumbnail completely white:
image

Same:
image

With fixed path the original behavior is restored. But due that the background image has a size from 500x2000, images with larger width and/or height will be cropped:

This thumbnail sizes well:
image

This thumbnail is broken:
image

With this patch the background image is used as a repeated texture. This way the bug is fixed, we can use a way smaller image and the whole process is less cpu expensive for imagick:

image

image


ssiergl:
@brusch @fashxp we just saw this behavior on our productive instance. Can be replicated on your demo as well. This should fix it.

ssiergl:
added GD support as well

brusch:
Very nice, thank you!

brusch:
Also made `asTexture` available for custom thumbnail configs: 
image

See db903ba443049574829f5dea00dc58c37a4b9af2

PR #4189, opened 2019-04-05
[DataObjects] Allow multiple assignments on advanced (object) relations

### LINKED ISSUE #2022: Object relations - allow assignment of an item multiple times
# Feature Request 

Currently the relation object types (e.g. Multi-Href, Multi-Href with Metadata ) only allow the assignment of an item once. It would be great if we can assign one item multiple times.

dvesh3:
Resolves #2022


bundles/CoreBundle/Resources/migrations/Version20190403120728.php
```diff
@@ -0,0 +1,83 @@
...
+    public function up(Schema $schema)
+    {
+        try {
+            $db = Db::get();
+
+            // add index to primary key in relations tables
+            $relationTables = $db->fetchAll("SHOW TABLES LIKE 'object\_relations\_%'");
+            foreach ($relationTables as $table) {
+                $relationTable = current($table);
+                $this->addSql('ALTER TABLE `' . $relationTable . '` ADD PRIMARY KEY `PRIMARYKEY` (`src_id`, `dest_id`, `ownertype`, `ownername`, `fieldname`, `type`, `position`, `index`), DROP INDEX `PRIMARY`');
```
fashxp: why not drop primary key first and then create new primary key with same name? something like that: 

``` 
alter table object_relations_33 drop primary key;

alter table object_relations_33
	add primary key (src_id, dest_id, ownertype, ownername, fieldname, type, position, `index`);
```

bundles/CoreBundle/Resources/migrations/Version20190403120728.php
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            $metaTables = $db->fetchAll("SHOW TABLES LIKE 'object\_metadata\_%'");
+            foreach ($metaTables as $table) {
+                $metaTable = current($table);
+
+                // add index column to metadata tables
+                $this->addSql('ALTER TABLE `' . $metaTable . '` ADD COLUMN `index` int(11) unsigned NOT NULL DEFAULT \'0\'');
+                $this->addSql('ALTER TABLE `' . $metaTable . '` ADD PRIMARY KEY `PRIMARYKEY` (`o_id`, `dest_id`, `type`, `fieldname`, `column`, `ownertype`, `ownername`, `position`, `index`), DROP INDEX `PRIMARY`');
```
fashxp: see comment above

bundles/CoreBundle/Resources/migrations/Version20190403120728.php
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public function down(Schema $schema)
+    {
+        try {
+            $db = Db::get();
+            // remove index from primary key in relations tables
+            $relationTables = $db->fetchAll("SHOW TABLES LIKE 'object\_relations\_%'");
+            foreach ($relationTables as $table) {
+                $relationTable = current($table);
+                $this->addSql('ALTER TABLE `' . $relationTable . '` ADD PRIMARY KEY `PRIMARYKEY` (`src_id`, `dest_id`, `ownertype`, `ownername`, `fieldname`, `type`, `position`), DROP INDEX `PRIMARY`');
```
fashxp: see comment above. 
in addition to that, this might fail since there might be data where already multiple assignments took place

bundles/CoreBundle/Resources/migrations/Version20190403120728.php
```diff
@@ -0,0 +1,83 @@
...
+                        ON  mt.o_id = rl.src_id
+                            AND mt.dest_id = rl.dest_id
+                            AND mt.type = rl.type
+                            AND mt.fieldname = rl.fieldname
+                            AND mt.ownertype = rl.ownertype
+                            AND mt.ownername = rl.ownername
+                            AND mt.position = rl.position
+                        SET mt.index = rl.index');
+            }
+        } catch (\Exception $e) {
```
fashxp: never ever catch exceptions and to nothing!

bundles/CoreBundle/Resources/migrations/Version20190403120728.php
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            // drop index column from metadata tables
+            $metaTables = $db->fetchAll("SHOW TABLES LIKE 'object\_metadata\_%'");
+            foreach ($metaTables as $table) {
+                $metaTable = current($table);
+                $this->addSql('ALTER TABLE `' . $metaTable . '` DROP COLUMN `index`');
+                $this->addSql('ALTER TABLE `' . $metaTable . '` ADD PRIMARY KEY `PRIMARYKEY` (`o_id`, `dest_id`, `type`, `fieldname`, `column`, `ownertype`, `ownername`, `position`), DROP INDEX `PRIMARY`');
+            }
+        } catch (\Exception $e) {
```
fashxp: never ever catch exceptions and to nothing!

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyObjectRelation.js
```diff
@@ -856,7 +857,7 @@ pimcore.object.tags.manyToManyObjectRelation = Class.create(pimcore.object.tags.
                 var key;
 
                 if (rdata.success) {
-                    var rec = this.store.getById(item.id);
+                    var rec = newItem[0];
```
fashxp: why did you change that line?

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyRelation.js
```diff
@@ -669,21 +670,35 @@ pimcore.object.tags.manyToManyRelation = Class.create(pimcore.object.tags.abstra
...
 
-    requestNicePathData: function (targets) {
+    requestNicePathData: function(targets, isInitialLoad) {
         if (!this.object) {
             return;
         }
+
+        var context = this.getContext();
+        var loadEditModeData = false;
+        if(isInitialLoad && this.fieldConfig.optimizedAdminLoading && context['containerType'] == 'object') {
```
fashxp: why did you add that here? optimizedAdminLoading is never available for manyToManyRelations

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyObjectRelation.js
dvesh3: because id is not unique anymore!

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyRelation.js
dvesh3: good catch...I'll remove it ;)

bundles/CoreBundle/Resources/migrations/Version20190403120728.php
dvesh3: I agree...but still there is chance of downgrade without changing the data! Do you think it is better to leave the index as is?

fashxp: I think we should try and throw an error, if it is not possible - with hints, what needs to be done.

PR #4197, opened 2019-04-08
fix admin translation javascript issues - bug fix when translations are not loaded

markus-moser:
This is a important follow up to #4170 
There are situations when pimcore.system_i18n is not loaded but the t() function get's called. We should merge this PR urgently as otherwise very strange bugs can occur.


brusch:
Many thanks!

PR #4208, opened 2019-04-09
[BackendSearch] restart save transaction if it fails

ssiergl:

Adding data to the backend search is called on every object/asset/document save. The element save transactions are protected against database failures (deadlocks etc.)
Due that backend search transaction is called afterwards (EventListener) it is not protected and will not restart on failure.

To fix it go inline with element save methods

## Additional info

We are importing a big amount of objects and assets multi threaded and we ran into this issue


brusch:
Thanks!

PR #4209, opened 2019-04-10
Bugfix: Controllers in bundles have not been listed in static routes / document types settings

BlackbitNeueMedien:
When you have a bundle other than AppBundle which does not have its controllers being registered as services then its action methods have not been listed in static routes and document types settings.

How to reproduce:
1. Create Bundle named TestBundle
1. Create controller DefaultController with action method testAction
1. Go to static routes settings
1. Select Module=TestBundle, Controller=DefaultController
1. Open action dropdown -> it shows methods of AppBundle\DefaultController not of TestBundle\DefaultController

The same applies for choosing the action in document type settings.


brusch:
Thanks a lot!

PR #4211, opened 2019-04-10
Show unreferenced assets

BlackbitNeueMedien:
Sometimes you want to clean up your assets. To be able to show assets which are not referenced in any object or document this PR introduces a new checkbox in the asset list view.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -2268,6 +2268,10 @@ class AssetController extends ElementControllerBase implements EventedController
                 $conditionFilters[] = 'path LIKE ' . ($folder->getRealFullPath() == '/' ? "'/%'" : $list->quote($folder->getRealFullPath() . '/%'));
             }
 
+            if ($allParams['only_unreferenced'] == 'true') {
+                $conditionFilters[] = 'id NOT IN (SELECT targetid FROM dependencies WHERE targettype=\'asset\')';
+            }
+
             $conditionFilters[] = "type != 'folder'";
             $filterJson = $allParams['filter'];
             if ($filterJson) {
--- a/bundles/CoreBundle/Resources/translations/de.json
+++ b/bundles/CoreBundle/Resources/translations/de.json
@@ -435,6 +435,7 @@
     "columns": "Spalten",
     "children_grid": "Kinder Tabelle",
     "only_children": "Nur direkte Kindelemente",
+    "only_unreferenced": "Nur nicht referenzierte",
     "cut": "Ausschneiden",
     "paste_cut_element": "Ausgeschnittenes Element einfügen",
     "memorize_tabs": "Geöffnete Tabs merken",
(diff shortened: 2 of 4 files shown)


bundles/CoreBundle/Resources/translations/de.json
```diff
@@ -435,6 +435,7 @@
     "columns": "Spalten",
     "children_grid": "Kinder Tabelle",
     "only_children": "Nur direkte Kindelemente",
+    "only_unreferenced": "Nur nicht referenzierte",
```
brusch: Please revert this change, only EN is maintained on GitHub, all other languages are on POEditor :)

bundles/AdminBundle/Controller/Admin/AssetController.php
```diff
@@ -2268,6 +2268,10 @@ public function gridProxyAction(Request $request, EventDispatcherInterface $even
                 $conditionFilters[] = 'path LIKE ' . ($folder->getRealFullPath() == '/' ? "'/%'" : $list->quote($folder->getRealFullPath() . '/%'));
             }
 
+            if ($allParams['only_unreferenced'] == 'true') {
+                $conditionFilters[] = 'id NOT IN (SELECT targetid FROM dependencies WHERE targettype=\'asset\')';
```
brusch: I'm afraid that this could cause major performance issues on bigger installations, I guess we should try to reduce the result set in the sub-select as much as possible.

bundles/CoreBundle/Resources/translations/de.json
BlackbitNeueMedien: Ok, reverted this. I do not know anything about PO editor so I cannot add the non-english translations with it. Can this be done by somebody else?

bundles/AdminBundle/Controller/Admin/AssetController.php
BlackbitNeueMedien: I do not think that this is a problem as MySQL uses indexes for the subquery:
```SQL
explain select * FROM `assets` WHERE (path LIKE '/%' AND id NOT IN (SELECT targetid FROM dependencies WHERE targettype='asset') AND type != 'folder') ORDER BY `id` ASC LIMIT 25
```
gives me:
```
id	select_type	table	partitions	type	possible_keys	key	key_len	ref	rows	filtered	Extra
1	PRIMARY	assets		index	"fullpath,path"	PRIMARY	4		50	45.00	Using where
2	DEPENDENT SUBQUERY	dependencies		index_subquery	"targetid,targettype"	targetid	4	func	13	31.35	"Using where; Using index"
```
So there is no temporary table or anything worse than this.
For 30.000 assets this query is executed on our database (with SQL_NO_CACHE to get the performance without query cache) in under 0.5 seconds.

brusch:
Thanks for your feedback!

BlackbitNeueMedien:
What do you think @brusch - would this feature also make sense for data objects? So you can find data objects which are not referenced by any other object or document?

brusch:
I've just reverted this PR ... there seems to be something wrong with it. 
When clicking the checkbox, 2 requests are fired, one with the parameter one without. In my case the list was never filtered.

brusch:
Same happens when clicking _just direct children_ , see: 


Needs further investigation.

BlackbitNeueMedien:
Well, I saw the 2 requests but they were there also when there was only the "only children" checkbox. That is why I didn't care.
But now I found the reason: When calling `this.grid.filters.clearFilters();` the store got reloaded and then when calling setExtraParam it was reloaded again. I removed the filter clearing, now it works and furthermore does not loose column filters -> win-win ;-)

PR #4220, opened 2019-04-11
Make workflow name available in place configs and global actions.

andreas-gruenwald:
Sometimes it is useful to access the workflow name in a place config or a global action.
For that case a getter is available now which returns the name of the current workflow.

### DIFF
--- a/lib/Workflow/GlobalAction.php
+++ b/lib/Workflow/GlobalAction.php
@@ -37,11 +37,15 @@ class GlobalAction implements NotesAwareInterface
      */
     private $expressionService;
 
-    public function __construct(string $name, array $options, ExpressionService $expressionService)
+    /** @var string */
+    private $workflowName;
+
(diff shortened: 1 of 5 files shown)


fashxp:
thx

lib/Workflow/GlobalAction.php
```diff
@@ -37,11 +37,15 @@ class GlobalAction implements NotesAwareInterface
      */
     private $expressionService;
 
-    public function __construct(string $name, array $options, ExpressionService $expressionService)
+    /** @var string */
+    private $workflowName;
+
+    public function __construct(string $name, array $options, ExpressionService $expressionService, string $workflowName)
```
fashxp: creation of new global actions [here](https://github.com/pimcore/pimcore/blob/75ee283823dcb5dd008d7a41a43153d204d64174/lib/Workflow/Manager.php#L98) also needs to set the workflow name

PR #4222, opened 2019-04-11
Show unreferenced assets (Follow-up of #4211)

BlackbitNeueMedien:
Follow-up of #4211

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -2268,6 +2268,10 @@ class AssetController extends ElementControllerBase implements EventedController
                 $conditionFilters[] = 'path LIKE ' . ($folder->getRealFullPath() == '/' ? "'/%'" : $list->quote($folder->getRealFullPath() . '/%'));
             }
 
+            if ($allParams['only_unreferenced'] == 'true') {
+                $conditionFilters[] = 'id NOT IN (SELECT targetid FROM dependencies WHERE targettype=\'asset\')';
+            }
+
             $conditionFilters[] = "type != 'folder'";
             $filterJson = $allParams['filter'];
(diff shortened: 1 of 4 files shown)


brusch:
Thanks a lot!

BlackbitNeueMedien:
@brusch, I did not dare to call it "unused" because this could lead to the wrong assumption that these are assets which have never been called via URL / thumbnail.

brusch:
@BlackbitNeueMedien that's true ... let's see if this causes some misunderstandings, we can rename it again any time :)

NiklasBr:
I think this is a nice feature and thank you both for working on it!

Though, *please* consider not using end users as a platform for "wait and see" experiments. Decisions like these are cheap and easy to get right now. But expensive to fix when it is in production and a customer has misunderstood that "unused" actually meant "unreferenced". It might cause unnecessary work, costs, and pain.

brusch:
Whatever ... cf5f48722726ff1693977ebaff38b13cae63a1f8

BlackbitNeueMedien:
@brusch, non-english translations are missing. How is the process to add those?

brusch:
@BlackbitNeueMedien https://poeditor.com/projects/view?id=38068

BlackbitNeueMedien:
Hmm, I created an account there but then your link gets me to a "404 Whoa! Nothing this way my friend!"

brusch:
Login and then join the project, just tested, works as expected.

BlackbitNeueMedien:
Hmm, do I need an invitation to the project? All I see is: Bildschirmfoto 2019-04-16 um 09 35 21
There I could create a new project but not join any.

brusch:
1. Login
2. Open https://poeditor.com/join/project/VWmZyvFVMH
3. Click button **Join**

PR #4227, opened 2019-04-12
add postAdd(Object|Document)Tree event for the backend ui

ultramegatom:
add postAddObject event for the backend ui


brusch:
Note: assets are not really possible, because there are too many ways of you how can add assets.

ultramegatom:
thx

PR #4243, opened 2019-04-16
Prevent select options without value

### LINKED ISSUE #4215: [Objects] Select data-type: automatically set display value to value
A lot of our customers forget to set values for select field options. They only fill the "display name", then they edit these select fields in the data objects and wonder why their selection does not get saved.

What do you think of using the "display value" as "value" if no value got entered for an option?
Or: Do not allow to save the class / brick / field collection if there is an option without a value.

BlackbitNeueMedien:
Prevent select and multiselect options without a value.
For the CSV editor I do not present an error message but use the key if no value has been specified for a certain row because the user can check the options after applying the CSV.

Resolves #4215


bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/select.js
```diff
@@ -192,8 +193,10 @@ pimcore.object.classes.data.select = Class.create(pimcore.object.classes.data.da
             ],
             autoHeight: true,
             plugins: [
-                Ext.create('Ext.grid.plugin.CellEditing', {
-                    clicksToEdit: 1
+                Ext.create('Ext.grid.plugin.RowEditing', {
+                    clicksToEdit: 1,
+                    errorSummary: false,
+                    autoCancel: true
```
weisswurstkanone: default anyway ?

weisswurstkanone:
image

- Messes up the layout a bit. Can you confirm?
- Also this row editor feels a bit awkward. When pressing enter I would expect to get from left to right.

BlackbitNeueMedien:
Yes, layout is not optimal with RowEditing but CellEditing somehow does not forbid empty values although I set `allowBlank: false` in column editor.

weisswurstkanone:
Mhm, but even the roweditor does not prevent you from saving things like this.

image

weisswurstkanone:
Mhm, but even the roweditor does not prevent you from saving things like this.

image

BlackbitNeueMedien:
Ok, I returned to CellEditing. But as CellEditing reverts the change to the previous value if you edit the value to "", we need an initial value. Therefore I use the number os options as default value (which can then be changed to the user's needs). What do you think?

NiklasBr:
@BlackbitNeueMedien I like this approach.

It mimics the Drupal/CCK approach where you will automatically get keys if you omit them, but can easily specify them from the start (using strings or numeric keys).

brusch:
@BlackbitNeueMedien In my opinion we can even automatically set the **_Display Value_** as the default for **_Value_** (which can be changed afterwards) , I'd be happy with that as well (probably more practical than a numeric value).

BlackbitNeueMedien:
@brusch Now the "display value" gets set as "value" if no value has been set.

bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/select.js
```diff
@@ -192,7 +192,20 @@ pimcore.object.classes.data.select = Class.create(pimcore.object.classes.data.da
...
-                    clicksToEdit: 1
+                    clicksToEdit: 1,
+                    listeners: {
+                        edit: function(editor, e) {
+                            if(!e.record.get('value')) {
+                                e.record.set('value', e.record.get('key'));
+                            }
+                        },
+                        beforeedit: function(editor, e) {
+                            if(e.field === 'value') {
```
weisswurstkanone: just one last note: would it make sense to also avoid duplicates?

BlackbitNeueMedien: just added this. when you set a value to an already existing value, it gets reverted (silently). Do you think we need a notification message?

weisswurstkanone: Guess we can still add it later if it feels strange.

weisswurstkanone:
thx!

PR #4244, opened 2019-04-16
Catch error if a single quantity value fails

jremmurd:
## Catch error if a single quantity value fails
Currently if one quantity value fails to load all fail.

### Use Case
When manipulating an object's layout within the `postOpenObject` hook (i.e. removing a tab which contains quantity values) since quantity value fields are fetched asynchronosly, if a single quantity value field can't be resolved any more (target fields for which the data would be fetched are gone) all quantity units fail.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers/quantityValue.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers/quantityValue.js
@@ -22,11 +22,7 @@ pimcore.helpers.quantityValue.initUnitStore = function(callback, filters) {
         var newListener = function () {
             pimcore.helpers.quantityValue.storeLoaded = true;
             pimcore.helpers.quantityValue.storeLoading = false;
-            try {
-                pimcore.helpers.quantityValue.getData(callback, filters);
-            } catch (e) {
-                console.log("An error occured while fetching quantity value data.");
-            }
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/helpers/quantityValue.js
```diff
@@ -22,7 +22,11 @@ pimcore.helpers.quantityValue.initUnitStore = function(callback, filters) {
         var newListener = function () {
             pimcore.helpers.quantityValue.storeLoaded = true;
             pimcore.helpers.quantityValue.storeLoading = false;
-            pimcore.helpers.quantityValue.getData(callback, filters);
+            try {
+                pimcore.helpers.quantityValue.getData(callback, filters);
```
weisswurstkanone: wouldn't it be better to filter it out in getData instead of letting the entire call fail ?

jremmurd: Please have look at my latest commits. I didn't exactly know what you ment with "filter it".

weisswurstkanone: My impression was that getData allows to pass in a list of IDs

PR #4258, opened 2019-04-18
[Object] DateTime Data-Type: Support for Seconds in User-Interface

BlackbitNeueMedien:
Till now datetime field's time part only support hours and minutes. With this PR also seconds are shown.


bundles/CoreBundle/Resources/translations/de.json
```diff
@@ -894,4 +894,4 @@
     "disable_split_view": "Disable Split View",
     "split_view_object_dirty_title": "Please confirm",
     "split_view_object_dirty_msg": "There are unsaved modifications. You will lose all changes. Are you sure?"
-}
\ No newline at end of file
+}
```
BlackbitNeueMedien: Sorry, obviously there ware some changes in my IDE which do not belong to this PR. Somehow I cannot revert this extra line break...

dpfaffenbauer: mine (PHPStorm) does that too, as soon as I open a file, it creates the line break.

BlackbitNeueMedien: I even edited the file in github's browser UI but no success.
Perhaps this file can be regenerated with PO editor...

NiklasBr: I believe this is intentional: https://github.com/pimcore/pimcore/blob/master/.editorconfig#L10

brusch: No, that's just for `*.php` files, not for `*.json`

BlackbitNeueMedien: @brusch would be nice, if you could revert changes of de.json concerning this PR, thanks.

brusch: @BlackbitNeueMedien Should be easier/quicker for you, I'd have to clone your repo first ...

brusch:
I'm just thinking if seconds should be there by default as in most use-cases seconds are probably not necessary at all (was fine now without seconds for years ...). 
Any other opinions?

BlackbitNeueMedien:
@brusch would there be any harm if seconds are shown but not used?
I think the problems with the current implementation are bigger for use-cases where seconds are necessary (e.g. transaction data, financial data).

brusch:
@BlackbitNeueMedien well, as always it's a matter of the point of view ... for editors it's probably not very convenient if seconds are there but not used.
I personally don't mind ... but you know ... 😉
Let's wait a bit and see if someone else has an opinion on that. In the worst case we'd have to make it configurable ...

bundles/CoreBundle/Resources/translations/de.json
BlackbitNeueMedien: Ok, now it worked - who knows why I was not able to revert the change last time...

brusch:
@BlackbitNeueMedien Thanks for your patience! 
We've discussed this topic internally and came to the result that it would be best to give the developer the possibility to completely customize the date and the time format. 
So we'd add 2 additional fields to the data-type configuration, one for the date format and one for the time format. Both fields should be prefilled with the current values.

How does that sound to you?

BlackbitNeueMedien:
I also thought of this approach some days ago. Sounds good to me.

brusch:
@BlackbitNeueMedien if you'd like to contribute the feature discussed above, please open up a new PR for that. I'm closing this PR now, since the agreed solution defers a bit from this original PR.

PR #4261, opened 2019-04-18
Unique key constraint not considered when restoring from recycling bi…

weisswurstkanone:


models/Element/Recyclebin/Item.php
```diff
@@ -125,6 +127,20 @@ public function restore($user = null)
             if ($indentElement) {
                 $element->setKey($element->getKey().'_restore');
             }
+
+            // create an empty object first and clone it
+            // see https://github.com/pimcore/pimcore/issues/4219
+            Model\Version::disable();
+            $class = get_class($element);
+            $dummy = new $class();
```
dpfaffenbauer: shouldn't you use the pimcore model factory here?

PR #4262, opened 2019-04-18
[Objects] Getter / Setter for reverse many-to-many relations

### LINKED ISSUE #3920: [Object][Tag] One-to-many and Many-To-Many Reverse with getter
Imagine the following model:

Objects:
  - Customer
  - Addresses

**Customer** 
Fields:
 - Addresses (One-to-many)

**Address** 
Fields:
 - Customer (Many-to-one)
 - Street
 - Name
 - etc.

The Reverse side of Address (Customer), would be the One-to-Many. In theory Pimcore should be able to allow that type, it knows the relation and allowed-types and could create a getter called `getAddresses` and get me all Addresses defined from the reverse side.

As well as with the many-to-many which partly already is implemented, but only for `Many-To-Many Objects`. But: It requires to have a unique name, but doesn't generate a getter.

BlackbitNeueMedien:
This PR adds getters and setters for reverse many-to-many relations.
For example when you have:
Product
- categories (many-to-many-object-relation)

Category
- products (reverse many-to-many relation)

```php
$products = $category->getProducts();
$products[] = Product::getById(123);
$category->setProducts($products);
$category->save();
```

Partly resolves #3920


dpfaffenbauer:
YES, wanted to add that for months and never came to it! Amazing!

BlackbitNeueMedien:
Todos: 
- new version for main object
- currently changes are not visible when comparing versions of object which has the reverse relation

BlackbitNeueMedien:
Todos:
- Set modification timestamp of main object's new version so it appears as "published"
- Create migration which saves all classes, so new getters / setters are added to data object classes

brusch:
@BlackbitNeueMedien please let us know when you're done. Thanks! 😊

BlackbitNeueMedien:
@brusch Now this PR is ready to be checked.

BlackbitNeueMedien:
Now tests are green again.

BlackbitNeueMedien:
@brusch Did you already have the time to check this PR? Or is there anything open to do?

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
```diff
@@ -296,6 +273,190 @@ public function isOptimizedAdminLoading(): bool
...
+            }
+            $deletedRelation->set($this->getOwnerFieldName(), array_values($reverseObjects));
+
+            $version = $deletedRelation->saveVersion(true, true, $params['versionNote'] ?? null);
+            $db->update('objects', ['o_versionCount' => $version->getVersionCount(), 'o_modificationDate' => $version->getDate()], ['o_id' => $deletedRelation->getId()]);
+
+            Cache::remove('object_' . $deletedRelation->getId());
+        }
+
+        $db->deleteWhere('object_relations_' . $this->getOwnerClassId(), 'dest_id='.$db->quote($object->getId()).' AND fieldname='.$db->quote($this->getOwnerFieldName()).' AND ownertype = \'object\'');
```
weisswurstkanone: TODO: add missing note to docs that only allowed in objects

        object: true,
        objectbrick: false,
        fieldcollection: false,
        localizedfield: false,
        classificationstore : false

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
```diff
@@ -296,6 +273,190 @@ public function isOptimizedAdminLoading(): bool
...
+        }
+
+        $relations = array_map(function ($relation) {
+            $relation['dest_id'] = $relation['src_id'];
+            unset($relation['src_id']);
+            return $relation;
+        }, $relations);
+
+        // using PHP sorting to order the relations, because "ORDER BY index ASC" in the queries above will cause a
+        // filesort in MySQL which is extremely slow especially when there are millions of relations in the database
```
NiklasBr: I think you can do some micro-optimisation here like this:

```php
        usort($relations, static function ($a, $b) {
            if ($a['index'] === $b['index']) {
                return 0;
            }
            return ($a['index'] < $b['index']) ? -1 : 1;
        });
```

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
```diff
@@ -296,6 +273,190 @@ public function isOptimizedAdminLoading(): bool
...
+            $version = $deletedRelation->saveVersion(true, true, $params['versionNote'] ?? null);
+            $db->update('objects', ['o_versionCount' => $version->getVersionCount(), 'o_modificationDate' => $version->getDate()], ['o_id' => $deletedRelation->getId()]);
+
+            Cache::remove('object_' . $deletedRelation->getId());
+        }
+
+        $db->deleteWhere('object_relations_' . $this->getOwnerClassId(), 'dest_id='.$db->quote($object->getId()).' AND fieldname='.$db->quote($this->getOwnerFieldName()).' AND ownertype = \'object\'');
+
+        $return = [];
+        if (is_array($data) && count($data) > 0) {
```
NiklasBr: `count($data)` scales linearly with the number of items, `empty()` will be slightly faster, also on line 434, might be valuable if there are lots of relations.

BlackbitNeueMedien: No, count() is O(1), see https://stackoverflow.com/questions/5835241/is-phps-count-function-o1-or-on-for-arrays
Also empty() is considered a bad practice because it does not care about the data type.

NiklasBr: [Really?
](https://stackoverflow.com/a/29635982/4919561)You are already verifying that it is an array though, also `count(FALSE))` may return `1` which is problematic.

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
```diff
@@ -296,6 +273,190 @@ public function isOptimizedAdminLoading(): bool
...
+                        'type' => 'object',
+                        'fieldname' => $this->getOwnerFieldName(),
+                        'index' => $counter
+                    ];
+
+                    if(!in_array($reverseObject->getId(), $oldRelations)) {
+                        $version = $reverseObject->saveVersion(true, true, $params['versionNote'] ?? null);
+                        $db->update('objects', ['o_versionCount' => $version->getVersionCount(), 'o_modificationDate' => $version->getDate()], ['o_id' => $reverseObject->getId()]);
+
+                        Cache::remove('object_' . $reverseObject->getId());
```
weisswurstkanone: Off topic, but a convenience function for that is overdue.

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
```diff
@@ -296,6 +273,190 @@ public function isOptimizedAdminLoading(): bool
...
+            $reverseObjects = $deletedRelation->get($this->getOwnerFieldName());
+
+            foreach($reverseObjects as $index => $reverseObject) {
+                if($reverseObject->getId() === $object->getId()) {
+                    unset($reverseObjects[$index]);
+                }
+            }
+            $deletedRelation->set($this->getOwnerFieldName(), array_values($reverseObjects));
+
+            $version = $deletedRelation->saveVersion(true, true, $params['versionNote'] ?? null);
```
weisswurstkanone: Saving a version even versioning is disabled ?

BlackbitNeueMedien: Oh, certainly there is a check missing. Which disabling do you mean? I only know of the number of version items in system config.

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
```diff
@@ -296,6 +273,190 @@ public function isOptimizedAdminLoading(): bool
...
+            $deletedRelation->set($ownerFieldName, array_values($reverseObjects));
+        }
+
+        return parent::preSetData($object, $data, $params);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function prepareDataForPersistence($data, $object = null, $params = [])
```
weisswurstkanone: I fear this is not right. If I don't miss anything then stuff that was previously done implicitely by the save() method is now completely missing. Example: Updating the dependencies table.

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
```diff
@@ -296,6 +273,190 @@ public function isOptimizedAdminLoading(): bool
...
+                }
+            }
+            $deletedRelation->set($this->getOwnerFieldName(), array_values($reverseObjects));
+
+            $version = $deletedRelation->saveVersion(true, true, $params['versionNote'] ?? null);
+            $db->update('objects', ['o_versionCount' => $version->getVersionCount(), 'o_modificationDate' => $version->getDate()], ['o_id' => $deletedRelation->getId()]);
+
+            Cache::remove('object_' . $deletedRelation->getId());
+        }
+
```
weisswurstkanone: Also keep in mind that the database contains the data for the PUBLISHED version. So if the latest owner version is not the published one then this is wrong.

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
weisswurstkanone: exactly.

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
weisswurstkanone: any opinion on that ?

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
BlackbitNeueMedien: done

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
BlackbitNeueMedien: @weisswurstkanone what do you mean with "versioning is disabled"? In https://github.com/pimcore/pimcore/blob/b87de319676a7978f0f1e73d0163133d86eda114/models/DataObject/Concrete.php#L285-L289 there is the actual check for the number of versions to be saved. Or is there any other place where you can disable versioning except setting the number of versions to be saved?

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
BlackbitNeueMedien: Doc for myself:
As documentation: Concrete::save() calls AbstractObject::save() which calls AbstractObject::update() and there is the dependency handling:
https://github.com/pimcore/pimcore/blob/b87de319676a7978f0f1e73d0163133d86eda114/models/DataObject/AbstractObject.php#L810-L823
`$this->resolveDependencies()` refers to 
https://github.com/pimcore/pimcore/blob/b87de319676a7978f0f1e73d0163133d86eda114/models/DataObject/Concrete.php#L377-L392
There for every Data field the method resolveDependencies gets called. This is currently implemented to always `return []`. If I implement it, the dependency is always written in the wrong direction as the source ID gets fixed in https://github.com/pimcore/pimcore/blob/b87de319676a7978f0f1e73d0163133d86eda114/models/DataObject/AbstractObject.php#L812
I think this will have to be added to ReverseManyToManyObjectRelation::prepareDataForPersistence()

BlackbitNeueMedien: Dependencies are now working.

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
BlackbitNeueMedien: @weisswurstkanone Fixed.

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
weisswurstkanone: AFAIK, there is no central helper for that kind of decision. Please still keep in mind that only the published version of the data should go into the database.

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -2176,98 +2166,6 @@ public function previewAction(Request $request)
...
-            //TODO: lock ?!
-            if (method_exists($owner, $getter)) {
-                $currentData = $owner->$getter();
-                if (is_array($currentData)) {
-                    for ($i = 0; $i < count($currentData); $i++) {
-                        if ($currentData[$i]->getId() == $object->getId()) {
-                            unset($currentData[$i]);
-                            $owner->$setter($currentData);
-                            $owner->setUserModification($this->getAdminUser()->getId());
-                            $owner->save();
```
weisswurstkanone: Another thing which came to my mind: will it still trigger DataObject events such as preUpdate and postUpdate on the owner object? Maybe we should just come up with a solution that generates getters&setters but leaves the original saving and loading logic untouched ?

weisswurstkanone:
Just tried the following:
- enable inheritance for owner class A
- create owner object of type A
- create target (nonowner) object (B)
- assign owner object A via reverse manytomany grid UI (B)
- save nonowner
=> changes are not reflected in the object_query_ table
Not OK

- reload owner object (A)
- save it
=> object_query_ table is OK again

BlackbitNeueMedien:
@weisswurstkanone, now object_query table of owner object gets updated as well.
I also included validity check for owner field restrictions.

weisswurstkanone:
Unfortunately it doesn't handle inheritance.

- Create (owner) class A + enable inheritance
- Create target (nonowner) class B
- Create object X of type A
- Create child object Y of type A so that X is the parent of Y
- Create target object Z of type B

- Let's do it from a owner perspective
- Open object X (which is the parent of Y) and assign target Z
- Save X
- object_query_A table should contain two rows - important here is that the entry for Y states that there is a (inherited) relation to Z
=> fine

- start again and do the same thing from a nonowner perspective, that is 
- open target Z
- add source object X
- save it (object Z)
- object_query_A is fine for object X but not for child object Y. Expectation is that there it contains the same value as for its parent.

I also found strange values like a colon only.

weisswurstkanone:
Regarding the owner scenario:
Handled by InheritanceHelper + Concrete\Dao

BlackbitNeueMedien:
@weisswurstkanone I followed your steps which resulted in an error as you said - adding the object to the nonowner relation:
* Create (owner) class A + enable inheritance
* Create target (nonowner) class B
* Create object X of type A
* Create child object Y of type A so that X is the parent of Y
* Create target object Z of type B
* open object Z
* add source object X
* save it (object Z)
After this my object_query_A looks like this:
Bildschirmfoto 2019-11-07 um 22 26 01
130804 is X in your and the above steps, 130805 is Y - so Y automatically inherited the relation from X. 

`I also found strange values like a colon only.` With the above steps or how to reproduce that?

weisswurstkanone:
Could you please paste a sample showing how you create & assign the objects here?
Thanks a lot!

BlackbitNeueMedien:
Query table is now ok for:
* Create (owner) class A + enable inheritance
* Create target (nonowner) class B
* Create object X of type A
* Create child object Y of type A so that X is the parent of Y
* Create target object Z of type B
* open object Z
* add source object X
* save it (object Z)

BlackbitNeueMedien:
I fixed the mentioned bugs. @weisswurstkanone, could you please give it another try? Hopefully we will finish this before this PR has its first birthday ;-)

weisswurstkanone:
Not really started with the review but one thing I noticed in the database.  Delete and re-add the owner using the backend UI.

image

models/DataObject/ClassDefinition/Data/ReverseManyToManyObjectRelation.php
```diff
@@ -296,6 +278,249 @@ public function isOptimizedAdminLoading(): bool
...
+     * @param DataObject\Concrete $object
+     * @param                     $data
+     * @param array               $params
+     *
+     * @return array|null
+     * @throws \InvalidArgumentException
+     */
+    public function preSetData($object, $data, $params = [])
+    {
+        $oldRelations = (array)$object->get($this->getName());
```
weisswurstkanone: general comment about preSetData: what is your opinion regarding the performance topic ?

BlackbitNeueMedien: In my tests it is now much faster than with the old implementation in DataObjectController because many database updates for the owner objects are made with bulk UPDATE queries - in contrast to calling the save() method for every owner object.
Nevertheless I did not dare to remove the hint "The current object is not the owner of these relations, making changes here might make saving the object slow" ;-)

BlackbitNeueMedien:
@weisswurstkanone Could you please explain in more detail how to reproduce the bug with "Delete and re-add the owner using the backend UI."

The following scenarios work here:
* Add an owner object to the reverse relation object and save. Then remove it and saved again - then there is only 1 ID in the reverse query table
* Add an owner object to the reverse relation object and save. Then remove it and add again (without saving after removal) and save - then there is only 1 ID in the reverse query table.

Also tested with adding 2 objects to the reverse relation.

weisswurstkanone:
maybe the following recording can give an idea, let me know :-)

[2019-12-10_10-12-19.zip](https://github.com/pimcore/pimcore/files/3943969/2019-12-10_10-12-19.zip)

BlackbitNeueMedien:
@weisswurstkanone Thanks, that helped to understand the problem. I think it should be fixed now.

BlackbitNeueMedien:
@weisswurstkanone, @brusch if there is anything left to do here, please let me know.

BlackbitNeueMedien:
Fixed conflicts.
@brusch could you please review this in the near future as it is much, error-prone work to keep this branch up-to-date with master.

brusch:
@BlackbitNeueMedien To be honest, I'm actually not really happy about how the saving is done, because a different object is doing database operations on another object. That's of course a matter of taste and I know that there might be different opinions on that, but I'm not really feeling good about that fact. 
So what I suggest is that we're starting with just the getter implementation, which would be fine for us as it is absolutely non-critical.

BlackbitNeueMedien:
```I'm actually not really happy about how the saving is done, because a different object is doing database operations on another object```
That is exactly what makes it much faster than the current saving implementation because there are a lot of bulk updates. We could call save() method for all referenced objects but at the cost of performance.

brusch:
Yep, I absolutely do understand the intention, but I see 2 main risks: 
- maintaining the code
- data consistency

dpfaffenbauer:
`To be honest, I'm actually not really happy about how the saving is done, because a different object is doing database operations on another object`

But does it really? It only does for the relations table, where the two objects have a connection anyway.

weisswurstkanone:
It is a bit more complex. There is also the query table (inheritance), dependencies, and maybe more

brusch:
Yes, just have a look here: https://github.com/pimcore/pimcore/pull/4262/files#diff-0501f86fb04f8426e33d658f4e2b2b34R411 
There quite many things that are done here which are done in a central place as well, that's why I'm so worried about the maintainability.

dpfaffenbauer:
> Yes, just have a look here: https://github.com/pimcore/pimcore/pull/4262/files#diff-0501f86fb04f8426e33d658f4e2b2b34R411
> There quite many things that are done here which are done in a central place as well, that's why I'm so worried about the maintainability.

Alright, seeing that also scares me :D

brusch:
@BlackbitNeueMedien So I'd go for the getter only for the moment, that's better than nothing. We can then think about how to solve the setter challenge. Ok for you? Can you update your PR accordingly? 
Thanks a lot! (also for your patience 😉)

brusch:
Ping 🏓 😊

BlackbitNeueMedien:
@brusch Sorry, have not much time at the moment in the evenings.
And on worktime I am working on completely different areas at the moment. Will try to continue soon...

brusch:
Hey @BlackbitNeueMedien! 
What about this PR? Should we close it for now and you just open a new one as soon as you had the time? I guess this would make sense since the original topic changed anyway, since it's just about the getter anymore 😊

PR #4264, opened 2019-04-18
Account port on URL generation in toolbar

roland4432:
changed the way the url is generated when clicking "open in new window" in the toolbar of a document

  
  

Resolves #

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/page_snippet.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/page_snippet.js
@@ -223,7 +223,7 @@ pimcore.document.page_snippet = Class.create(pimcore.document.document, {
                     iconCls: "pimcore_icon_open_window",
                     scale: "medium",
                     handler: function () {
-                        window.open(this.data.url);
+                        window.open(this.data.path + this.data.key);
                     }.bind(this)
                 });
 


roland4432:
Hello Bernhard,

well... ok, if you regard it as workaround ;) Your cms, your rules.. BUT
The solution is pretty easy, it only used parameters which are already available.
Since you pointed me to the PageSnippet-Class, I could add something like:

```
    public function getUrl($hostname = null, $scheme = null)
    {
        if (!$scheme) {
            $scheme = 'http://';
            $requestHelper = \Pimcore::getContainer()->get('pimcore.http.request_helper');
            if ($requestHelper->hasMasterRequest()) {
                $scheme = $requestHelper->getMasterRequest()->getScheme() . '://';
            }
        }

        if (!$hostname) {
            if (!$hostname = \Pimcore\Config::getSystemConfig()->general->domain) {
                if (!$hostname = \Pimcore\Tool::getHostname()) {
                    throw new \Exception('No hostname available');
                }
            }
        }

        $port = "";
        if(!in_array($_SERVER['SERVER_PORT'], [80, 443])){
            $port = ':' .$_SERVER['SERVER_PORT'];
        }

        $url = $scheme . $hostname . $port . $this->getFullPath();

        $site = \Pimcore\Tool\Frontend::getSiteForDocument($this);
        if ($site instanceof Model\Site && $site->getMainDomain()) {
            $url = $scheme . $site->getMainDomain() . $port . preg_replace('@^' . $site->getRootPath() . '/?@', '/', $this->getRealFullPath());
        }

        return $url;
    }
```

but I refuse to add a PR for this ugly lines :)

I couldn't find the port in any of the Pimcore\Config::getXY() Methods, but maybe you can help me with that. I'd go with a PR then ;)

Regards

brusch:
Yes it is a workaround, because it doesn't cover sites (just the main site) ... so I'd be a regression. 
Anyway, the proposed fix above looks good so far, but only if sub-sites are sharing the same port as the main site, but that should be normally the case - not sure if we should handle that as well (would require an additional configuration on site level).

roland4432:
Yes. You're right. Doesn't cover the site-features :( My Bad (Still learning the internals of pimcore)
I'm eager to help, so I am very much into any suggestions and coop work on that item. It's a very small bug though, so maybe we don't put too much work in this.
I still wonder, why the other options to open a site in a new window account the port if they access the same Model file for building the url?

PR #4265, opened 2019-04-19
[Elements] Allow array parameter binding for listing conditions

### LINKED ISSUE #4112: [DataObjects] hydratation problem when using placeholder in Listing::setCondition() with "SELECT ... IN (...)" statement
When using object Listing->setCondition() method, if the condition contains something like "state IN (:states)", the underlying raw generated SQL query is correct and returns results. 
BUT the Listing->getObjects() return an empty array !

Exemple : 
```php
    public static function getByRegion(): array
    {
        $listing = new User\Listing();

        $listing->setCondition('role LIKE :role AND state IN (:states)', [
            'role' => 'Admin',
            'states' => ['East', 'West'] // Array that could come from a variable
        ]);

        $listing->load();

        return $listing->getObjects();
    }
```

This will returns an **empty** array BUT using the Profiler and retrieveing the underlying raw SQL query, it returns results (lauching on phpMyAdmin for instance)
Sample raw SQL query generated yet : 
```SQL
SELECT object_BM_User.o_id as o_id, `object_BM_User`.`o_type` FROM `object_BM_User` WHERE ((role LIKE 'Admin' AND state IN ('East', 'West') AND object_BM_User.o_type IN ('object','folder')) AND object_BM_User.o_published = 1)
```

The objects are yet not hydrated...

The only way to make it working is to inject the string of joined values like this : 
```php
$states = ['East', 'West'];
$listing->setCondition(sprintf('(rose LIKE :role AND state IN ("%s")', join('","', $states)), [
            'role' => 'Admin'
        ]);
```

dvesh3:
Resolves #4112

### DIFF
--- a/doc/Development_Documentation/05_Objects/03_Working_with_PHP_API.md
+++ b/doc/Development_Documentation/05_Objects/03_Working_with_PHP_API.md
@@ -92,7 +92,7 @@ $entries->addConditionParam("city = ?", "New York", "AND"); // concatenator can 
 //use array bindings for prepared statements (since version 5.8.0)
 $entries->setCondition("city IN (?)", [["New York", "Chicago"]]);
 // or
-$entries->setCondition("city IN (:cities)", ["cities" => ["New York", "Chicago"]]); // With PDO_Mysql you can use named parameters
+$entries->setCondition("city IN (:cities)", ["cities" => ["New York", "Chicago"]]); // named parameters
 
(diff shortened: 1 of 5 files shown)


brusch:
@dvesh3 I think we should also add a code example here https://pimcore.com/docs/5.x/Development_Documentation/Objects/Working_with_PHP_API.html#page_Object-Listings

Thanks!

PR #4267, opened 2019-04-23
Add missing trims for email-array getters

GALCF:
## Steps to Reproduce

Set your default email settings (receiver and copy) to include spaces like `john.doe@example.com; jane.doe@example.com`.  
Send an email with these default settings. It should throw an error that the second email is not compliant to the RFC 2822, 3.6.2

Adds trim command for all email split strings (found with `preg_split('/,|;/'`)


brusch:
Thanks!

PR #4268, opened 2019-04-24
Correct return type for inherited data

BlackbitNeueMedien:
Without this PR it is possible to get data from getters whose type does not match with the doc tags. This often confuses programmers and IDEs. This is especially annoying for relation getters which shall return arrays but do return `null` if neither the object itself has the relation field being filled nor its parents (because https://github.com/pimcore/pimcore/blob/master/models/DataObject/Concrete.php#L532-L543 is independent of the field type). 
But this also applies for non-inherited data. With this PR the returned data gets casted to the correct return type.

### DIFF
--- a/models/DataObject/ClassDefinition/Data.php
+++ b/models/DataObject/ClassDefinition/Data.php
@@ -648,26 +648,6 @@ abstract class Data
         $code .= "\t\t" . 'return $preValue;' . "\n";
         $code .= "\t" . '}' . "\n";
 
-        $docTypes = (string)$this->getPhpdocType();
-        $allowedReturnTypes = [];
-        foreach(explode('|', $docTypes) as $docType) {
-            $docType = trim($docType);
-            if(strpos($docType, '?') === 0) {
-                $allowedReturnTypes[] = 'null';
-                $docType = substr($docType, 1);
-            }
-            if(substr($docType, -2) === '[]') {
-                $allowedReturnTypes[] = 'array';
-            } else {
-                $allowedReturnTypes[] = $docType;
-            }
-        }
-
-        $typeCast = '';
-        if(count($allowedReturnTypes) === 1 && in_array($allowedReturnTypes[0], ['string', 'int', 'integer', 'bool', 'boolean', 'float', 'double', 'real', 'array'])) {
-            $typeCast = '('.$allowedReturnTypes[0].')';
-        }
-
         if (method_exists($this, 'preGetData')) {
             $code .= "\t" . '$data = $this->getClass()->getFieldDefinition("' . $key . '")->preGetData($this);' . "\n";
         } else {
@@ -677,15 +657,18 @@ abstract class Data
         // insert this line if inheritance from parent objects is allowed
         if ($class instanceof DataObject\ClassDefinition && $class->getAllowInherit() && $this->supportsInheritance()) {
             $code .= "\t" . 'if(\Pimcore\Model\DataObject::doGetInheritedValues() && $this->getClass()->getFieldDefinition("' . $key . '")->isEmpty($data)) {' . "\n";
-            $code .= "\t\t" . 'return '.$typeCast.'$this->getValueFromParent("' . $key . '");' . "\n";
+            $code .= "\t\t" . '$parentValue = $this->getValueFromParent("' . $key . '");' . "\n";
+            $code .= "\t\t" . 'if($parentValue !== null) {' . "\n";
+            $code .= "\t\t\t" . 'return $parentValue;' . "\n";
+            $code .= "\t\t" . '}' . "\n";
             $code .= "\t" . '}' . "\n";
         }
 
         $code .= "\t" . 'if ($data instanceof \\Pimcore\\Model\\DataObject\\Data\\EncryptedField) {' . "\n";
-        $code .= "\t\t" . '    return '.$typeCast.'$data->getPlain();' . "\n";
+        $code .= "\t\t" . '    return $data->getPlain();' . "\n";
         $code .= "\t" . '}' . "\n";
 
-        $code .= "\treturn " . $typeCast.'$data' . ";\n";
+        $code .= "\treturn " . '$data' . ";\n";
         $code .= "}\n\n";
 
         return $code;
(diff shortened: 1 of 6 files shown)


weisswurstkanone:
IMHO, typecasting (Data.php) is wrong as there is no way to return null anymore (null is not the same as the empty string ""). Have a look at this example for an input field. 

/**
* Get myinputfield - ifferl
* @return string
*/
public function getMyinputfield () {
	$preValue = $this->preGetValue("myinputfield"); 
	if($preValue !== null && !\Pimcore::inAdmin()) { 
		return $preValue;
	}
	$data = $this->myinputfield;
	if(\Pimcore\Model\DataObject::doGetInheritedValues() && $this->getClass()->getFieldDefinition("myinputfield")->isEmpty($data)) {
		return (string)$this->getValueFromParent("myinputfield");
	}
		    return (string)$data->getPlain();
	}
	return (string)$data;
}

BlackbitNeueMedien:
If it should really be possible to return null for input fields then the PHP return doc type should be set to "@return null|string" or "?null", shouldn't it?

weisswurstkanone:
*) I am experiencing the following problem when I change the unit dropdown (tag)

Message: Type error: Return value of Pimcore\Model\DataObject\QuantityValue\Unit::getConverter() must be of the type string, null returned

*) Not sure if conversion should be done automatically or better be an option in the class definition

*) Oberservation: network panel indicates that there are more requests than needed to add a unit definition

weisswurstkanone:
Discussed this topic internally. Please remove the typecasting entirely.
Regarding the other question: Please give string|null a try ($phpDocType). However, keep in mind that we have to check all datatypes.

BlackbitNeueMedien:
?? When I remove the typecasting there won't be much left in this PR.
The original goal of this PR was to always get arrays for relation field getters so we can always do 
```php
foreach($object->getManyToManyRelationABC() as $obj) {
...
}
```
At the moment this fails when neither the object nor its parent have any object in the field ManyToManyRelationABC.

Whether there is string or ?string or string|null in the @return type does not really matter to me.

weisswurstkanone:
OK, then why not just override getGetterCode for relational data types and provide a specific implementation there instead of doing this phpdoc guesswork? I am probably missing something :-)

BlackbitNeueMedien:
Hmm, that would mean to duplicate the whole getGetterCode, getGetterCodeObjectbrick, getGetterCodeFieldcollection etc. - only to typecast the return value of Concrete::getValueFromParent(). I also could fetch the field definition in Concrete::getValueFromParent() and use the same typecasting logic as I had in https://github.com/pimcore/pimcore/pull/4268/files#diff-4255251c227151b9e7966fe8a417fc47R651-R669 - but you would not accept that either, would you?

The main question to me is: Do the getters may return null for all fields or are sane defaults for empty values ok? 
If we need null, then we have to accept that the following constructs will not be possible without the danger of a PHP error:
* `foreach($object->getManyToManyRelationABC() ...)`
* `foreach($object->getFieldCollectionABC() ...)`

BlackbitNeueMedien:
Not I found a way: We return the inherited parent value only if it is not null, otherwise we return the value of the object's field. 
If you wonder how this may happen: Concrete::getValueFromParent() fetches the parent element and calls its getter. If the value is empty Concrete::getValueFromParent() is called again and so on. Eventually there is no parent object and then Concrete::getValueFromParent() returns null.

BlackbitNeueMedien:
@weisswurstkanone, @brusch Any problems remaining with this PR?

brusch:
@BlackbitNeueMedien 
- regarding field collections: that's a proper BC break, so I wouldn't do that now. 
- regarding the `getValueFromParent()` topic: I'm not 100% sure which side-effects that might have, so we need to further investigate this.

dpfaffenbauer:
regarding Field Collection: That is a proper BC, but I would like to have that in 6.0 :). 

In fact, it doesn't even matter code-wise if it is null or just empty, if it is just empty, I don't have to take care about creating it and just loop over it.

brusch:
Since we have 3 different topics within this single PR, I've fixed now the PHPDoc here 76a2e424bcf9221e52c312d5f5d4de5d8366bfea

Regarding the 2 remaining topics ... 
Field-Collections:   
Would be possible with v6, but the feature is not complete in this PR. The new behavior needs to be  considered at least in `\Pimcore\Model\DataObject\ClassDefinition\Data::isEmpty()`, so it is necessary to override this method in `\Pimcore\Model\DataObject\ClassDefinition\Data\Fieldcollections` and implement the functionality accordingly. This has probably also effects in the admin UI. 

`\Pimcore\Model\DataObject\Concrete::getValueFromParent`:  
Maybe we should take the change and get rid of the workaround using `null` anyway ... 
In my opinion we should check whether there's a parent or not, if not, then we return the original value. This would give us also the possibility that a parent can safely return `null` even if the original value is not null but in "empty" state. 

I'm thinking of something similar to the following code snippet: 
```php
public function getValueFromParent($key, $params = null)
    {
        $parent = $this->getNextParentForInheritance();
        if ($parent) {
            $method = 'get' . $key;
            if (method_exists($parent, $method)) {
                return call_user_func([$parent, $method], $params);
            } else {
                throw new \Exception(sprintf('Parent object does not have a method %s(), unable to retrieve value for key %s', $method, $key));
            }
        }

        throw new \Exception('No parent object available');
    }
```

and the getter code example (the `try/catch` is the new part): 

```php
public function getEnabled () {
	$preValue = $this->preGetValue("enabled"); 
	if($preValue !== null && !\Pimcore::inAdmin()) { 
		return $preValue;
	}
	$data = $this->enabled;
	if(\Pimcore\Model\DataObject::doGetInheritedValues() && $this->getClass()->getFieldDefinition("enabled")->isEmpty($data)) {
	    try {
            return $this->getValueFromParent("enabled");
        } catch (\Exception $e) {
	        // no data from parent available, continue ... 
        }
	}
		    return $data->getPlain();
	}
	return $data;
}
```

@weisswurstkanone @dpfaffenbauer @BlackbitNeueMedien  what do you think about that?

BlackbitNeueMedien:
Sounds good to me.

brusch:
Good, if nobody has concerns so far, I'll prepare a PR for that :)

PR #4284, opened 2019-04-25
[Elements] Fixed: locked elements should not be deleted

### LINKED ISSUE #4281: [Elements] Locked elements should not be deleted from grid
# Bug Report
### Expected behavior
Locked elements should not be deleted from grid
### Actual behavior
It is possible to delete **locked** elements from grid

### example
image

dvesh3:
Resolves #4281

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -570,7 +570,7 @@ class AssetController extends ElementControllerBase implements EventedController
                  * @var $asset Asset
                  */
                 $deletedItems[] = $asset->getRealFullPath();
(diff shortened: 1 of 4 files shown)


brusch:
@diveshp3 imho we should also check the lock status in the element's `delete()` methods, because it shouldn't be possible to delete locked elements via the API either.

dpfaffenbauer:
@brusch I don't think so, if the dev wants to delete it, even though it is locked, he should do so. If he cares about the locked status, he should check if it is locked before.

brusch:
Well, depends on the point of view and the definition of the lock status ... but we can leave this open for discussion. @diveshp3 ignore my previous message.

PR #4286, opened 2019-04-26
recipient calculation newsletter panel

zerlpaMMT:
Add the number of recipient to the dialog of "do you really want to send the newsletter to all".


dvesh3:
@zerlpaMMT your PR looks good.. can we also stop sending newsletter if no recipient is found (just to avoid below situation)
image

brusch:
@dvesh3 @zerlpaMMT yes, that would be indeed a nice future addition!

Issue #3970, opened 2019-02-20
[Web2Print] add global option to enable PDFReactor debug mode


weisswurstkanone:

dpfaffenbauer:
env var maybe?

weisswurstkanone:
see #4292

PR #4292, opened 2019-04-27
Add possibility to enable debug mode when generating PDFs with PDFreactor

Ben305:
Add possibility to enable debug mode when generating PDFs with PDFreactor.

This is especially useful when using Javascript, because console.log output is appended to the generated PDF.

PDFreactor Debug Mode Switch

This is a sample of the output (only the beginning, logs are really detailed) appended by PDFreactor at the end of the document
PDFreactor Debug Logs

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/web2print.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/web2print.js
@@ -151,6 +151,11 @@ pimcore.settings.web2print = Class.create({
                         fieldLabel: t("web2print_licence"),
                         name: 'pdfreactorLicence',
                         value: this.getValue("pdfreactorLicence")
+                    }, {
+                        xtype: 'checkbox',
+                        fieldLabel: t("web2print_enableDebugMode"),
+                        name: 'pdfreactorEnableDebugMode',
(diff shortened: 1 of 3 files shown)


weisswurstkanone:
Makes sense. Please also have a look at https://github.com/pimcore/pimcore/issues/3970
This would be useful when it comes to investigate general configuration issues.

Ben305:
@weisswurstkanone No problem, I've added an option to the Web2print settings.

Ben305:
@brusch Any chance to see this in 5.8.2?

bundles/AdminBundle/Resources/public/js/pimcore/settings/web2print.js
```diff
@@ -151,6 +151,11 @@ pimcore.settings.web2print = Class.create({
                         fieldLabel: t("web2print_licence"),
                         name: 'pdfreactorLicence',
                         value: this.getValue("pdfreactorLicence")
+                    }, {
+                        xtype: 'checkbox',
+                        fieldLabel: t("web2print_enableDebugMode"),
```
weisswurstkanone: Translation is probably missing. Please add it to en.json

weisswurstkanone:
thx!

PR #4318, opened 2019-05-03
Fix a set of PHP notices

NiklasBr:

Resolves no specific issues


models/DataObject/AbstractObject.php
```diff
@@ -712,7 +712,7 @@ public function save()
 
             if ($isUpdate) {
                 $updateEvent = new DataObjectEvent($this);
-                if ($differentOldPath) {
+                if (isset($differentOldPath) && strlen($differentOldPath) > 0) {
```
brusch: I think it would be better to declare `$differentOldPath = null`  in advance.

brusch: I quickly fixed that 😉

PR #4319, opened 2019-05-03
Fixes __clone method non-object by moving it to after instanceof

NiklasBr:
Error was: `CRITICAL: __clone method called on non-object ... _clone method called on non-object at vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/AssetController.php:69`

Resolves no specific bug report, but this has sometimes happened after assets were not fully uploaded and some traces of seemed to be left in caches (maybe).

Just hardening the code against those odd situations.


brusch:
Thanks a lot!

PR #4320, opened 2019-05-03
Fix down logic in Migration 20190403120728

pmuskovac:
Fix the used table when restoring the old primary keys.

Currently, this leaves the last modifed relation table without a primary key, and causes errors for the meta tables as you cannot add multiple primary keys to a table.


weisswurstkanone:
thx!

PR #4323, opened 2019-05-06
Use numeric field options also for cleaning up data in setter

BlackbitNeueMedien:
Currently when calling the setter of a numeric field the data gets written to the object without any validation or cleanup. With this PR the field's settings "integer", "decimal precision" and "unsigned" are used to clean-up set data.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Numeric.php
+++ b/models/DataObject/ClassDefinition/Data/Numeric.php
@@ -577,16 +577,8 @@ class Numeric extends Data implements ResourcePersistenceAwareInterface, QueryRe
      */
     public function preSetData($object, $data, $params = [])
     {
-        if($this->getInteger()) {
-            $data = (int)$data;
-        }
-
         if($this->getDecimalPrecision()) {
             $data = round($data, $this->getDecimalPrecision());
-        }
-
-        if($this->getUnsigned()) {
-            $data = max(0, $data);
         }
 
         return $data;
--- a/models/DataObject/ClassDefinition/Data/Numeric.php
+++ b/models/DataObject/ClassDefinition/Data/Numeric.php
@@ -567,4 +567,28 @@ class Numeric extends Data implements ResourcePersistenceAwareInterface, QueryRe
 
         return (float) $value;
     }
+
+    /**
+     * @param $object
+     * @param $data
+     * @param array $params
+     *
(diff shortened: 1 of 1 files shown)


brusch:
I don't think we should just silently cast the passed values, I think it's better to throw proper exceptions and this is what is done already here: https://github.com/pimcore/pimcore/blob/b48e588b69871456ee74ed1b3a8faeef62d4f5cf/models/DataObject/ClassDefinition/Data/Numeric.php#L464

BlackbitNeueMedien:
@brusch But checkValidity() does not check for number of decimals, does it?
That was the main point I stumbled upon because I wondered why two versions had different values in a numeric field with 2 decimals. In one version there were 4 decimals. Could we do this rounding in preSetData? Or also handle number of decimals in checkValidity()?

brusch:
Rounding should be automatically done by the database, so I don't think we need to cover this in `checkValidity()`.

BlackbitNeueMedien:
Of course is rounding done by the database. But I speak of the object before it has been saved to database.
By calling $object->setMyNumericField(1.2345); the value in the object has 4 decimals although in the field definition only 2 have been allowed. This leads also to wrong display in the versions (as the version gets the data from the PHP object in the moment of saving).

brusch:
Oh, yes that's right of course :) 
Let's do that in `preSetData()` i guess it's the right place for that. 
I'll reopen this PR so you can update it. Thanks! 👍😊

BlackbitNeueMedien:
Ok, I added the numerical rounding to preSetData.
One fundamental question remains to me: Should setters protect its invariants also in memory (when calling the setters) or is it enough to do that on saving via checkValidity()? In my opinion it would be good to always have integrity in the object's data (as value objects do that, see https://en.wikipedia.org/wiki/Value_object), not only after saving.

brusch:
Thank you very much for updating the PR! 

Regarding value objects: that's a fundamental question and changing this would be quite a huge topic and a proper BC break.  I think we should discuss this topic in an extra issue for v6.0.

petterjv:
@brusch Please revert this change. This PR breaks my code completely since I use bcmath a lot for calculations. It's not a good idea to use round in a setter here. It makes the correct string representation of a mysql decimal to be converted into a floating point number (getXYZ returns a float).

PR #4327, opened 2019-05-06
Add rest example for asset creation.

andreas-gruenwald:
Add example for creating an asset via Pimcore REST API with minimum data.


weisswurstkanone:
thx!

PR #4328, opened 2019-05-07
Documentation: Fix method call for admin session

kovinet:
There was a mistake in code example for getting admin session


brusch:
Thanks a lot!

PR #4330, opened 2019-05-07
Grouping role permissions for better overview

PGBrandi:
Added groups for permissions in roles
collapsed elements at initialization


andreas-gruenwald:
This is how the feature looks like:
image

- same feature for user permissions still pending.

brusch:
In general I like the idea of grouping the available permissions, but in my opinion we should choose a more generic approach. I'm thinking of an additional column `group` in the `users_permission_definitions` table, which is then used for grouping the permissions. 
What do you think?

PGBrandi:
I already thought about this option, but didn't want to alter anything in the database. So yes, this would be the nicer option. I'll look at it.

andreas-gruenwald:
Has this PR been closed on purpose?
Should we create another PR to continue with this topic, or just adapt the current solution?

fashxp:
lets create a new one as it is a complete different approach. thank you very much!

PR #4331, opened 2019-05-08
Use index for version maintenance query

BlackbitNeueMedien:
In the maintenance job at the step of removing excess versions the following SQL query gets executed (when using max. steps of object versions and not versioning by days):
https://github.com/pimcore/pimcore/blob/606c231f9dddd5fbc6458077014aafe4af315493/models/Version/Dao.php#L200
The actual query for example is then:
`SELECT cid,count(*) as amount FROM versions WHERE ctype = 'object' AND NOT public AND id NOT IN (0) GROUP BY cid HAVING amount > 10`

Currently this query is using a temporary table and filesort:
```
1 | SIMPLE | versions | NULL | ref | PRIMARY,cid,ctype | ctype | 2 | const | 83065 | 5.00 | Using index condition; Using where; Using temporary; Using filesort
```

In this PR I added an index with 2 columns `ctype`, `cid`. With this index the query plan is quite better:
```
1 | SIMPLE | versions | NULL | ref | PRIMARY,cid,ctype_cid | ctype_cid | 2 | const | 83064 | 5.00 | Using index condition; Using where
```

As MySQL always uses matching prefixes of indexes all queries which used the single-column index `ctype` until now, can now use the new 2-column index.

### DIFF
--- a/bundles/CoreBundle/Resources/migrations/Version20190508074339.php
+++ b/bundles/CoreBundle/Resources/migrations/Version20190508074339.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+
+class Version20190508074339 extends AbstractPimcoreMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE `versions`
+            DROP INDEX `ctype`,
+            ADD INDEX `ctype_cid` (`ctype`, `cid`);');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE `versions`
+            DROP INDEX `ctype_cid`,
+            ADD INDEX `ctype` (`ctype`);');
+    }
+}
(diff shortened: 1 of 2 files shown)


brusch:
@BlackbitNeueMedien Thanks a lot for your PR!
Should we also consider the column `public` for the key?

BlackbitNeueMedien:
@brusch I do not think that adding `public`to the index has any advantage because most of version datasets have `public`=0 - to be exact: (1 - 1/version steps) * 100 %
From https://dev.mysql.com/doc/refman/8.0/en/mysql-indexes.html:
`
When a query needs to access most of the rows, reading sequentially is faster than working through an index
`

BlackbitNeueMedien:
@brusch One question keeps coming to my head from time to time: The `NOT public` part of https://github.com/pimcore/pimcore/blob/606c231f9dddd5fbc6458077014aafe4af315493/models/Version/Dao.php#L200 was originally introduced for not deleting the currently published version, wasn't it?
As the currently published version is handled via objects.o_versionCount = versions.versionCount nowadays, is the maintenance query in https://github.com/pimcore/pimcore/blob/606c231f9dddd5fbc6458077014aafe4af315493/models/Version/Dao.php#L200 still correct? Because I have cases where the published version actually does get deleted:
Bildschirmfoto 2019-05-22 um 13 34 10

BlackbitNeueMedien:
@brusch have you seen my comment?

brusch:
@BlackbitNeueMedien `public` is used for documents, where you can have multiple public versions with individual public accessible URLs, which can be used e.g. for multi-variant testing. 
image

So basically no public version should be deleted at all.

Issue #3799, opened 2019-01-03
[Symfony 4] Define commands as services


brusch:
https://github.com/pimcore/pimcore/blob/master/bundles/CoreBundle/Resources/config/commands.yml

PR #4342, opened 2019-05-08
[Symfony 4] Define commands as services #3799

weisswurstkanone:


fashxp:
https://github.com/pimcore/pimcore/commit/cc117f07839630a73063109f424d0b9ecf6aa5bb

PR #4349, opened 2019-05-12
[Newsletter] Improved error handling while sending newsletter

Lexipowder:
If the domain of a recipient is misspelled or not existing, the existing process causes:

Expected response code 354 but got code "554", with message "554 5.5.0 No valid recipients have been specified

Improvement:
First Step: Avoiding any email domains with no MX entries.
Second Step: If any error occurs cancel the jamming smtp-process via $mailer->getTransport()->stop();

* Changed Logger:info to Logge r:warn otherwise no output in the “newsletter-sending-output.log”
* Added counter to see the exact progress of the sending process in the .log file, helps later in finding any further errors
* Added ReturnPath for hard bounces

### DIFF
--- a/bundles/CoreBundle/Command/InternalNewsletterDocumentSendCommand.php
+++ b/bundles/CoreBundle/Command/InternalNewsletterDocumentSendCommand.php
@@ -145,13 +145,12 @@ class InternalNewsletterDocumentSendCommand extends AbstractCommand
         $limit = $fifth > 10 ? 10 : ($fifth < 3 ? 3 : intval($fifth));
         $offset = 0;
         $hasElements = true;
+        $index = 1;
 
         while ($hasElements) {
             $tmpStore = Model\Tool\TmpStore::get($sendingId);
 
             $data = $tmpStore->getData();
-
-            Logger::info('Sending newsletter ' . $hasElements . ' / ' . $totalCount. ' [' . $document->getId(). ']');
 
             $data['progress'] = round($offset / $totalCount * 100, 2);
             $tmpStore->setData($data);
@@ -159,6 +158,8 @@ class InternalNewsletterDocumentSendCommand extends AbstractCommand
 
             $sendingParamContainers = $addressAdapter->getParamsForSingleSending($limit, $offset);
             foreach ($sendingParamContainers as $sendingParamContainer) {
+                Logger::warn('Sending newsletter ' . $index . ' / ' . $totalCount. ' [' . $document->getId(). ']');
+                
                 try {
                     $mail = \Pimcore\Tool\Newsletter::prepareMail($document, $sendingParamContainer, $hostUrl);
                     \Pimcore\Tool\Newsletter::sendNewsletterDocumentBasedMail($mail, $sendingParamContainer);
@@ -170,6 +171,7 @@ class InternalNewsletterDocumentSendCommand extends AbstractCommand
                     Logger::warn("Sending configuration for sending ID $sendingId was deleted. Cancelling sending process.");
                     exit;
                 }
+                ++$index;
             }
 
             $offset += $limit;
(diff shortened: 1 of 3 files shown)


bundles/CoreBundle/Command/InternalNewsletterDocumentSendCommand.php
```diff
@@ -145,20 +145,21 @@ protected function doSendMailInSingleMode(Model\Document\Newsletter $document, A
...
 
-            Logger::info('Sending newsletter ' . $hasElements . ' / ' . $totalCount. ' [' . $document->getId(). ']');
-
             $data['progress'] = round($offset / $totalCount * 100, 2);
             $tmpStore->setData($data);
             $tmpStore->update();
 
             $sendingParamContainers = $addressAdapter->getParamsForSingleSending($limit, $offset);
             foreach ($sendingParamContainers as $sendingParamContainer) {
+                Logger::warn('Sending newsletter ' . $index . ' / ' . $totalCount. ' [' . $document->getId(). ']');
```
brusch: Why severity `warning`?

lib/Mail.php
```diff
@@ -623,7 +623,13 @@ public function sendWithoutRendering(\Swift_Mailer $mailer = null)
 
         if ($event->hasArgument('mailer')) {
             $mailer = $event->getArgument('mailer');
-            $mailer->send($this);
+            try {
+                $mailer->send($this, $emailaddress);            
+            } catch (\Exception $e){
+                $mailer->getTransport()->stop();
+                sleep(10);
```
brusch: Why `sleep()`?

brusch:
@Lexipowder ping 😊

bundles/CoreBundle/Command/InternalNewsletterDocumentSendCommand.php
Lexipowder: Sorry, I'm super bussy this week.
The issue with the warning is, that an Logger::info isn't logging to the newsletter-sending-output.log.
But it is essential for error tracking, especially related to customers, that you know exactly which customers got the newsletter and which ones not. But maybe I'm missing the point how to partial change the logging to info-Level..

lib/Mail.php
Lexipowder: To make sure the transportation connection is really closed, otherwise we will run in occupied errors, Sleep() happen only in an error case. But I read a bit more about swiftmailer in the meantime, maybe the is a better solution, but this have to wait for the weekend.

bundles/CoreBundle/Command/InternalNewsletterDocumentSendCommand.php
brusch: Ok, thanks for clarifying! 
Please add a comment there explaining why we're using `warning` so that it doesn't get changed in the future.

lib/Mail.php
brusch: Ok, thanks, then I'll wait until you come back on this topic.

lib/Mail.php
```diff
@@ -623,7 +623,12 @@ public function sendWithoutRendering(\Swift_Mailer $mailer = null)
 
         if ($event->hasArgument('mailer')) {
             $mailer = $event->getArgument('mailer');
-            $mailer->send($this);
+            try {
+                $mailer->send($this, $emailaddress);            
+            } catch (\Exception $e){
+                $mailer->getTransport()->stop();
+                throw new \Exception($emailaddress[0].' - '.$e->getMessage());
+            }
```
Lexipowder: This should work well, without sleep(), according to swiftmailer docs:

/**
     * Check if this Transport mechanism is alive.
     *
     * If a Transport mechanism session is no longer functional, the method
     * returns FALSE. It is the responsibility of the developer to handle this
     * case and restart the Transport mechanism manually.
     *
     * @example
     *
     *   if (!$transport->ping()) {
     *      $transport->stop();
     *      $transport->start();
     *   }
     *
     * The Transport mechanism will be started, if it is not already.
     *
     * It is undefined if the Transport mechanism attempts to restart as long as
     * the return value reflects whether the mechanism is now functional.
     *
     * @return bool TRUE if the transport is alive
     */

brusch:
Thanks!

PR #4350, opened 2019-05-12
Make EcommerceFrameworkBundle::hasCurrentUserId() more reliable

das-peter:
## Problem

Currently `Pimcore\Bundle\EcommerceFrameworkBundle::hasCurrentUserId()` isn't very reliable because it checks `userId` for the value defined in `Pimcore\Bundle\EcommerceFrameworkBundle\Environment::USER_ID_NOT_SET` (-1).
However, there are multiple locations in which `userId` is set to `null` as default/re-set value.

A loose `empty()` check in `hasCurrentUserId()` could be fair to handle that in a very generic way but I'd say it's preferable to fix the consistency first.

Change two locations to use `Pimcore\Bundle\EcommerceFrameworkBundle\Environment::USER_ID_NOT_SET` when "re-"setting the `userId` property.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Environment.php
+++ b/bundles/EcommerceFrameworkBundle/Environment.php
@@ -174,7 +174,7 @@ class Environment implements IEnvironment
         $this->load();
 
         $this->customItems = null;
-        $this->userId = null;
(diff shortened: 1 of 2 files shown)


fashxp:
thx

PR #4356, opened 2019-05-13
[Fieldcollection] Allow class definition import with missing fieldcollections

blankse:
## Additional info  

`Definition::getByKey` should always return Fieldcollection/Definition objects. Otherwise it throws an exception. With try/catch it is possible to import the class and remove the missing fieldcollections.

The same is in the setAllowedTypes function:
https://github.com/pimcore/pimcore/blob/cc117f07839630a73063109f424d0b9ecf6aa5bb/models/DataObject/ClassDefinition/Data/Fieldcollections.php#L372-L377


brusch:
Thanks!

PR #4362, opened 2019-05-14
new Image cacheBuster Feature was not working

MercuryKojo:
new Image cacheBuster Feature was not working because the prefix was not equal to asset path (cache-buster-...)
added preg_replace to resolve this issue


brusch:
Grazie Mille!

PR #4366, opened 2019-05-14
Respect the default language setting when creating new users

pmuskovac:
Pimcore doesn't respect the default language setting in the system settings when creating new users. With this PR the language is set to the system settings value for every user on user creation.

## Expected behavior:
When setting the 'Default-Language in Admin-Interface' in System Settings > General to e.g. 'German', a new users 'Language' setting should by default be set to 'German'

## Actual Behavior:
The default language is always 'English'

## Steps to reproduce:
- Go to https://demo-basic.pimcore.org/admin/
- Open Settings > System Settings > General
- Change the default admin language to something other than English
- Save and agree to reload Pimcore
- Go to Settings > User and Roles > Users
- Create a new user
=> The language of the new user is 'English'

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/UserController.php
+++ b/bundles/AdminBundle/Controller/Admin/UserController.php
@@ -118,7 +118,8 @@ class UserController extends AdminController implements EventedControllerInterfa
                 'parentId' => intval($request->get('parentId')),
                 'name' => trim($request->get('name')),
                 'password' => '',
-                'active' => $request->get('active')
+                'active' => $request->get('active'),
+                'language' => \Pimcore\Config::getSystemConfig()->get('general')->get('language')
             ]);
 
             if ($request->get('rid')) {


brusch:
The default language in the system settings has nothing to do with the default language for editors in the UI. 

There are 2 different types of language settings: 
- The application languages, which are project-specific and configured in System Settings
- The UI languages, which are distributed by the Pimcore core

brusch:
Thanks anyway :)

pmuskovac:
Okay, so what setting do I need to change so that new Editors have their language set to German by default?
Yes, I can change the language of an Editor as an Admin, but that's rather error prone and the Editors don't have Permission to access the User configuration so can't change their own language setting.

We would also like to automate this, but for that I also would like to understand which setting I should be reading the default value from.

brusch:
There's currently no setting for the default UI language.
But you can use the event `\Pimcore\Event\UserRoleEvents::PRE_ADD` to set the language for new users.

pmuskovac:
Ok, thanks!

PR #4367, opened 2019-05-14
Fix: Missing field name translation in grid cell editor

blankse:
This PR translates the field name like in edit helper:
https://github.com/pimcore/pimcore/blob/8b21bdd0e5cbb42b4261e2e596849f57292d8529/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/edit.js#L134-L137

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridCellEditor.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridCellEditor.js
@@ -51,12 +51,15 @@ Ext.define('pimcore.object.helpers.gridCellEditor', {
             return;
         }
 
-        var title = fieldInfo.label ? fieldInfo.label : fieldInfo.key;
-
         this.context = this.editingPlugin.context;
         // this.callParent(arguments);
 
         var tagType = fieldInfo.layout.fieldtype;
+
+        // translate title
+        if(typeof fieldInfo.layout.title != "undefined") {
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridCellEditor.js
```diff
@@ -51,13 +51,16 @@ Ext.define('pimcore.object.helpers.gridCellEditor', {
...
-        var title = fieldInfo.label ? fieldInfo.label : fieldInfo.key;
-
         this.context = this.editingPlugin.context;
         // this.callParent(arguments);
 
         var tagType = fieldInfo.layout.fieldtype;
 
+        // translate title
+        if(typeof fieldInfo.layout.title != "undefined") {
+            fieldInfo.layout.title = t(fieldInfo.layout.title);
```
weisswurstkanone: shouldn't it be ts instead ?

blankse: `ts()` is only a alias for `t()`
https://github.com/pimcore/pimcore/blob/8b21bdd0e5cbb42b4261e2e596849f57292d8529/bundles/AdminBundle/Resources/public/js/pimcore/functions.js#L64-L66

So i thought `t()` should be preferred

blankse:
@brusch This change is not enough. So only the title of the Window is translated. But not the field itself.

brusch:
You are absolutely right, thanks!

PR #4368, opened 2019-05-14
[Symfony 4] Replace SensioGeneratorBundle #3275

### LINKED ISSUE #3275: [Symfony 4] Replace SensioGeneratorBundle
we are not using **symfony/maker-bundle** at the moment as it does not provide bundle skeleton generate command and also there are no changes in directory structure for backward compatibility.

dvesh3:
Resolves #3275


dpfaffenbauer:
what about https://github.com/symfony/maker-bundle?

brusch:
@dpfaffenbauer doesn't support generating bundles 😉

dpfaffenbauer:
@brusch but you could extend it to support it ;)

brusch:
@dpfaffenbauer this was also our first idea, but the maker-bundle uses the new folder structure, so this approach is less effort. As soon as we switch to the new folder structure it definitely makes sense to use the maker-bundle.

dpfaffenbauer:
ok

PR #4373, opened 2019-05-15
Fix: Deprecations of Twig, Google Client and Pimcore

blankse:
Twig functions and tests write deprecation warnings

### DIFF
--- a/lib/Config.php
+++ b/lib/Config.php
@@ -173,7 +173,7 @@ class Config
             $siteId = null;
             if (Model\Site::isSiteRequest()) {
                 $siteId = Model\Site::getCurrentSite()->getId();
-            } elseif (Tool::isFrontentRequestByAdmin()) {
+            } elseif (Tool::isFrontendRequestByAdmin()) {
                 // this is necessary to set the correct settings in editmode/preview (using the main domain)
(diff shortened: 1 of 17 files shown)


lib/Config.php
```diff
@@ -173,7 +173,7 @@ public static function getWebsiteConfig($language = null)
             $siteId = null;
             if (Model\Site::isSiteRequest()) {
                 $siteId = Model\Site::getCurrentSite()->getId();
-            } elseif (Tool::isFrontentRequestByAdmin()) {
+            } elseif (Tool::isFrontendRequestByAdmin()) {
```
brusch: What has changed here? Can you maybe revert those changes?

blankse: isFrontentRequestByAdmin is deprecated:
https://github.com/pimcore/pimcore/blob/7bff13be32bf5962a323ac56549c1a6b2ab12a73/lib/Tool.php#L362-L372

Should we write a deprecated warning if someone uses it and remove it in 6.0?

brusch: Ah, sorry, didn't see the difference at the first sight :) 
It's fine as it is, thanks!

PR #4381, opened 2019-05-16
Performance enhancement when moving object to folder with lots of children

### LINKED ISSUE #4332: Bad performance when moving objects to folders with lots of siblings
When you move a data object to a folder which already contains a lot of objects the performance is really slow - or in other words the system breaks down.

Steps to reproduce:
1. Take a folder which contains about 10.000 objects
1. Create an object outside of this folder
1. Move the object via drag-and-drop to the folder
1. In my example the AJAX request to /admin/object/update needed 292 seconds

The cause is this function:
https://github.com/pimcore/pimcore/blob/606c231f9dddd5fbc6458077014aafe4af315493/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L1107-L1140

In this function all sibling objects are fetched and their o_index gets updated. This results in 10.000 update statements to be executed - of course this takes some time.

I do not really now what the o_index column is for - I thought of sorting by index vs. sorting by key but my folders are all sort by key, so updating the index would not be necessary.

My naive approach would be:
```php
protected function updateIndexesOfObjectSiblings(DataObject\AbstractObject $updatedObject, $newIndex)
    {
        $updatedObject->saveIndex($newIndex);

        $list = new DataObject\Listing();
        Db::get()->executeUpdate('UPDATE '.$list->getDao()->getTableName().' SET o_index = o_index+1 WHERE o_parentId = ? AND o_id != ? AND o_index > ?', [$updatedObject->getParentId(), $updatedObject->getId(), $newIndex]);
    }
```
But I did not dare to put this in a PR because I am certainly missing some point why it is currently implemented by updating all siblings separately. 
But even if this naive approach is not useful we need to find a way to accelerate moving an object to a folder with lots of objects.

BlackbitNeueMedien:
As described in #4332 there is a major performance issue when moving a data object to a folder with lots of objects because all siblings get updated separately:
https://github.com/pimcore/pimcore/blob/606c231f9dddd5fbc6458077014aafe4af315493/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L1107-L1140

This PR updates all sibling's o_index values at once.

Performance comparison for a folder with 14.000 objects:
without this PR: 267 seconds
with this PR: 130 seconds

Resolves #4332

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -1128,15 +1128,16 @@ class DataObjectController extends ElementControllerBase implements EventedContr
         $updatedObject->saveIndex($newIndex);
 
         Db::get()->executeUpdate('UPDATE '.$list->getDao()->getTableName().' o, 
-        (
-            SELECT newIndex, o_id FROM (SELECT @n := IF(@n = ? - 1,@n + 2,@n + 1) AS newIndex, o_id 
-            FROM '.$list->getDao()->getTableName().', 
-            (SELECT @n := -1) variable 
-            WHERE o_id != ? AND o_parentId = ? AND o_type IN (\'' . implode("','", [DataObject\AbstractObject::OBJECT_TYPE_OBJECT, DataObject\AbstractObject::OBJECT_TYPE_VARIANT, DataObject\AbstractObject::OBJECT_TYPE_FOLDER]) . '\') 
-            ORDER BY o_index, o_id=?
-        ) tmp) order_table
-        SET o.o_index = order_table.newIndex
-        WHERE o.o_id=order_table.o_id',
+            (
+                SELECT newIndex, o_id FROM (SELECT @n := IF(@n = ? - 1,@n + 2,@n + 1) AS newIndex, o_id 
+                FROM '.$list->getDao()->getTableName().', 
+                (SELECT @n := -1) variable 
+                WHERE o_id != ? AND o_parentId = ? AND o_type IN (\'' . implode("','", [DataObject\AbstractObject::OBJECT_TYPE_OBJECT, DataObject\AbstractObject::OBJECT_TYPE_VARIANT, DataObject\AbstractObject::OBJECT_TYPE_FOLDER]) . '\') 
+                    ORDER BY o_index, o_id=?
+                ) tmp
+            ) order_table
+            SET o.o_index = order_table.newIndex
+            WHERE o.o_id=order_table.o_id',
             [
                 $newIndex, $updatedObject->getId(), $updatedObject->getParentId(), $updatedObject->getId()
             ]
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Could you please double check. Managed to do the following:

image

BlackbitNeueMedien:
@weisswurstkanone I even triple-checked ;-)  but the query is quite complex. In my tests it worked.
Can you explain what you did to get this result?

weisswurstkanone:
Example:
- Drag any news node and drop it somewhere inside the same folder (e.g. at 3rd position).
- Look at the database
- There will be two entries with o_index = 2

BlackbitNeueMedien:
Hmm, I can reproduce the problem. I tried several changes yesterday but in some case it is always wrong - but I am sure that there is a way to accomplish updating all siblings at one as in the end it is only a loop as it was before in the PHP code.

weisswurstkanone:
Do you want to update the PR ? :-)

BlackbitNeueMedien:
Of course - when I found a working solution. ;-)
If you have an idea how to change the query, please comment here.

weisswurstkanone:
Db::get()->executeUpdate('UPDATE '.$list->getDao()->getTableName().' o, 
            (
                SELECT newIndex, o_id FROM (SELECT @n := IF(@n = ? - 1,@n + 2,@n + 1) AS newIndex, o_id 
                FROM '.$list->getDao()->getTableName().', 
                (SELECT @n := -1) variable 
                WHERE o_id != ? AND o_parentId = ? AND o_type IN (\'' . implode("','", [DataObject\AbstractObject::OBJECT_TYPE_OBJECT, DataObject\AbstractObject::OBJECT_TYPE_VARIANT, DataObject\AbstractObject::OBJECT_TYPE_FOLDER]) . '\') 
                ORDER BY o_index, o_id=?
            ) tmp) order_table
            SET o.o_index = order_table.newIndex
            WHERE o.o_id=order_table.o_id',
            [
                $newIndex, $updatedObject->getId(), $updatedObject->getParentId(), $updatedObject->getId()
            ]
        );

BlackbitNeueMedien:
Great job, @weisswurstkanone - this really works!
I updated the PR.

As I wrote, it still needs about 130 seconds for 14.000 siblings. I have not investigated but I think this is caused by getting the last version of all those objects. What do you think, should we try to only select those objects whose latest version is newer than the published one (by joining the versions table)? (But when we do so, we have to keep in mind to call clearDependentCache() for all siblings). But we could also care about this in a separate PR...

weisswurstkanone:
Getting or saving the latest version? Could you please further investigate? I would prefer to create a separate PR for further improvements :-)
What we can definitely do is to only save the objects with modified indices.

weisswurstkanone:
@BlackbitNeueMedien , could you please give it a try and compare?
https://github.com/pimcore/pimcore/tree/update_siblings

BlackbitNeueMedien:
Works really well, now moving an object to a folder with 9500 siblings needs 2.8 seconds.

PR #4391, opened 2019-05-17
check template select on document

AlternateIf:
1) check template select on document instead of throwing an exception. 
the exception was thrown due to not taking the template select (as seen below) into account if there is no template matching the bundle/controller/action specifications
2) move building the legacy template to a new method to make it easier to override

@dvesh3 can you please review this. i'm not too sure about passing the request parameter holding the template-name entered in the template select straight through

image

### DIFF
--- a/bundles/CoreBundle/Templating/LegacyTemplateGuesser.php
+++ b/bundles/CoreBundle/Templating/LegacyTemplateGuesser.php
@@ -3,6 +3,7 @@
 namespace Pimcore\Bundle\CoreBundle\Templating;
 
 use Doctrine\Common\Persistence\Proxy;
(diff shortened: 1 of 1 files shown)


brusch:
We have the same issue for the core route `/admin/asset/get-preview-document`.
I don't think throwing an exception is even necessary, as it works like a charm if we don't do so. 
But in my opinion it makes sense to return the original template reference in the case the legacy template reference doesn't exist, so that the resulting exception doesn't show the legacy reference name.

brusch:
Btw. changed that now in 5a41ce4ef25c8675474ab1298ae51966fe812a44

AlternateIf:
@brusch 
I can confirm that the template select is now working as expected for me

Issue #4378, opened 2019-05-16
Remove all deprecated functions & BC layers for version 6


brusch:

brusch:
0cd37aad03281b56d67461e6c6a3cd66e9493316

PR #4405, opened 2019-05-20
Remove Legacy Mode and Support for v4 Compatibility Bridge

brusch:
- Remove support https://github.com/pimcore/pimcore4-compatibility-bridge/tree/master/src/Pimcore 
- Legacy Mode (Documents, Static Routes, ...) 

See also #4378


solverat:
@brusch I always wondered why the upgrade button in the extension manager has been gone. Never had the time to investigate. Now, 1.5 years later, I checked the file history and stumbled on this PR: "remove legacy mode".

Within this PR, the possibility to upgrade bundles via extension manager is gone. No note, no warning, nothing. Not nice.

https://github.com/pimcore/pimcore/pull/4405/files#diff-0637bbf7054ab2202ceddd30d833659470961c9cbc7716b6a2c5fdb491641f07L317-L360

Is there any reason for that?

brusch:
Updates were only possible for old extensions, right, this was just there because of the compatibility bridge.

solverat:
It worked with all bundles (Bundle with and also without Migration-Support). Just like the Install/Uninstall-Button still works with all kind of extensions.

dpfaffenbauer:
@brusch yep, what @solverat says is true...

brusch:
Ok, sorry, then I don't know ... is there any reason why this should be still there, not really, right as it should be done using composer anyway.

PR #4406, opened 2019-05-21
Bugfix: Send language param in grid tab 

woellundwoell:
When using the SQL editor in the grid view the selected language from grid config is not transfered.
After clicking the button to show SQL editor, the correct language is displayed but grid data is translated to default language.

This behavior is reproducable in demo system.


woellundwoell:
Sorry we screwed up some PR.

PR #4409, opened 2019-05-21
Fix: Language for TranslateValue operator

woellundwoell:
The TranslateValue operator does not take into account the language setting of the grid configuration (for view and CSV export).
## Steps to reproduce

1. Configure gridview with "Transformers > Operator Translate Value" + Input attribute
2. Select other language than default language
3. Value in grid column and CSV export column will not be translated

### DIFF
--- a/lib/DataObject/GridColumnConfig/Operator/Factory/TranslateValueFactory.php
+++ b/lib/DataObject/GridColumnConfig/Operator/Factory/TranslateValueFactory.php
@@ -19,7 +19,6 @@ namespace Pimcore\DataObject\GridColumnConfig\Operator\Factory;
 
 use Pimcore\DataObject\GridColumnConfig\Operator\OperatorInterface;
 use Pimcore\DataObject\GridColumnConfig\Operator\TranslateValue;
-use Pimcore\Tool;
 use Symfony\Component\Translation\TranslatorInterface;
 
 class TranslateValueFactory implements OperatorFactoryInterface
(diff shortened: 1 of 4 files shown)


lib/DataObject/GridColumnConfig/Operator/Factory/TranslateValueFactory.php
```diff
@@ -35,6 +36,9 @@ public function __construct(TranslatorInterface $translator)
 
     public function build(\stdClass $configElement, $context = null): OperatorInterface
     {
+	    if(null !== $context && isset($context['language']) && Tool::isValidLanguage($context['language'])) {
+		    $this->translator->setLocale($context['language']);
```
weisswurstkanone: I would pass both translator service and locale to the TranslateValue operator and set + restore the locale in getLabeledValue similar to https://github.com/pimcore/pimcore/blob/c62924ca5c3d0c482a71097cb0231aff143208d6/lib/DataObject/GridColumnConfig/Operator/LocaleSwitcher.php#L54

woellundwoell: We are not sure exactly what you mean. The LocalSwitcherOperator gets the language from the operator setting while the TranslatedValueOperator gets the language from the grid configuration / request. The Pimcore\Model\DataObject\Service receives the language from the request. The only possibility we saw was to pass this parameter in context.
Would the solution shown on the screenshot be OK?
Bildschirmfoto 2019-05-21 um 15 39 50

weisswurstkanone: yes :-)

weisswurstkanone:
thx!

PR #4413, opened 2019-05-21
[Symfony4] Fix deprecations

dvesh3:
Resolves #


solverat:
@dvesh3 one of our bundle tests just introduced an error in P6. Because of changing the MessageSelector class to IdentityTranslator, the `choose` method here:

https://github.com/pimcore/pimcore/blob/16c1d538462ca4b348aa447abd4ba0cc41d0242c/lib/Translation/Translator.php#L154

does not exist anymore:

```
  [Error] Call to undefined method Symfony\Component\Translation\IdentityTranslator::choose()  
```

PR #4416, opened 2019-05-22
Bugfix: Create asset from remote stream

BlackbitNeueMedien:
Currently there is a bug when you want to create an asset from a remote stream, reproducable with the following script:
```php
<?php
$fileStream = fopen('https://picsum.photos/200', 'rb');
\Pimcore\Model\Asset::create(1, [
    'filename' => 'test'.uniqid(),
    'stream' => $fileStream,
    'userOwner' => 0,
    'userModification' => 0,
]);
```

Problem is that PHP does not support rewinding remote streams. For this reason I keep the temporary file which originally was used to determine the MIME type and use this also for the data import.


brusch:
Thanks!

PR #4417, opened 2019-05-22
[Document|Object] Sort contextmenu by translated names

blankse:
In the moment the contextmenu of document and object are not sorted by name if you have set translations.


brusch:
Thanks!

PR #4418, opened 2019-05-22
[Data Object] fixes behavior of DataObject\Concrete::getValueFromParent() ...

brusch:
..., to work as expected without the workaround using 'null' - related to discussion here #4268

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -523,11 +523,9 @@ class Concrete extends AbstractObject implements LazyLoadedFieldsInterface
 
     /**
      * @param $key
-     * @param $params
-     *
+     * @param null $params
      * @return mixed
-     *
-     * @todo: return explicit null or false
+     * @throws \Exception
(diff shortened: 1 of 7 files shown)


models/DataObject/Concrete.php
```diff
@@ -536,10 +534,12 @@ public function getValueFromParent($key, $params = null)
             $method = 'get' . $key;
             if (method_exists($parent, $method)) {
                 return call_user_func([$parent, $method], $params);
+            } else {
+                throw new \Exception(sprintf('Parent object does not have a method called `%s()`, unable to retrieve value for key `%s`', $method, $key));
             }
         }
 
-        return;
+        throw new \Exception('No parent object available to get a value from');
```
dpfaffenbauer: It should throw a specific Exception to Pimcore to be able to specifically catch it

brusch: Yep, makes sense, thanks!

PR #4420, opened 2019-05-22
[Data Object] make call of preGetValue() hook optional

brusch:
Currently `preGetValue()` is called in every getter.
This PR introduces `PreGetValueHookInterface` which needs to be implemented by the extending class in order to use the `preGetValue()` hook.


dpfaffenbauer:
:+1

PR #4432, opened 2019-05-23
delete dynamic-contants.php

dpfaffenbauer:
First: it has a spelling error in the name
Second: It's not used anywhere?!


brusch:
It's just there for auto-completion in the IDE, because the constants are set dynamically here  https://github.com/pimcore/pimcore/blob/master/lib/Bootstrap.php... but yes, the typo im the file name needs to be fixed 😊

dpfaffenbauer:
🤦‍♂

brusch:
What's wrong with that?

dpfaffenbauer:
It's just a dirty hack for something that should be refactored anyway...

Yivan:
I think it should be kept.
For my case, stubs is of great help for AST analyser like phpstan.
I included them in autoload_files section so I can analyse project bundle code without phpstan warning.

Yivan:
@brush just the line `require_once '../config/constants.php';` is bad resolved... phpstan reports it does not exists.

brusch:
@Yivan You are referencing the wrong user name ;) 
I have no clue what you mean ...

Yivan:
@brush No it was well for this issue, i was telling:
- yes in Pimcore stub `dynamic-contants.php` should be renammed `dynamic-constants.php` : https://github.com/pimcore/pimcore/tree/master/stubs
- and Pimcore stubs are useful so it should be kept (for IDE like you said or in my case for php linter PHPStan).
That's all !

brusch:
@Yivan again ... wrong username referenced, it's `brusch` not `brush` 😅

Yivan:
Ha I get it !!! I correct it !! Thanks : ) I need some sleep ; )

PR #4433, opened 2019-05-23
don't use require_once in code, use composer autoloading instead

dpfaffenbauer:


composer.json
```diff
@@ -103,6 +103,10 @@
     "elasticsearch/elasticsearch": "2.0.0"
   },
   "autoload": {
+    "files": [
+      "lib/simple_html_dom.php",
```
brusch: I'm not sure if we should really load this huge library on every request, even if it's not used ... this doesn't feel right to me ...
I think we should instead migrate to the OOP port of simple_html_dom: https://packagist.org/packages/sunra/php-simple-html-dom-parser

dpfaffenbauer: To be honest: it doesn’t hurt PHP at all loading it. It’s getting opcached anyway. 

But: if the OOP port works, I am happy with that as well

brusch: So are you updating your PR?

dpfaffenbauer: yep, later today

brusch: Seems that `sunra/php-simple-html-dom-parser` has still some issues with PHP 7.3 (there are some open PRs for that), so probably not an alternative for now.

dpfaffenbauer:
ok, happy with that

brusch:
Changed Simple DOM HTML as well. 
It's used in 90% of the requests anyway, so no real benefit ...
Thanks again!

PR #4435, opened 2019-05-23
refactor tests to test for symfony 3.4 and 4.*

dpfaffenbauer:
refactor autoloader to don't require composer autoloader anymore.


dpfaffenbauer:
also: this now doesn't depend on the skeleton anymore. and you are also not forced to install it two levels deep into a directory structure to work.

Example: I like to have a `lib/` Folder where I develop certain bundles, I include them using the `repositories` section in the `composer.json` file. With Pimcore, it has to be two levels deep like it would be when it's installed with composer. So in my case it currently sits in `lib/Pimcore/core` where `lib/Pimcore` should be fine enough.

lib/Bootstrap.php
```diff
@@ -226,23 +228,24 @@ public static function defineConstants()
...
 
         // legacy mapping loader creates aliases for renamed classes
-        $legacyMappingLoader = new \Pimcore\Loader\Autoloader\AliasMapper($loader);
+        $legacyMappingLoader = new AliasMapper();
         $legacyMappingLoader->register(true);
 
         // the following code is out of `app/autoload.php`
         // see also: https://github.com/symfony/symfony-demo/blob/master/app/autoload.php
-        AnnotationRegistry::registerLoader([$loader, 'loadClass']);
+        // Don't think this is necessary anymore
```
brusch: Let's just kick this code if it's not necessary anymore 😉

lib/Loader/Autoloader/AliasMapper.php
```diff
@@ -61,26 +61,8 @@ class AliasMapper extends AbstractAutoloader
...
-        $aliases = array_keys($this->mapping, $class);
-        if (count($aliases)) {
-            $this->composerAutoloader->loadClass($class);
-            // the return of composer autoloader obviously doesn't work, be better check manually if the class really exists
-            if ($this->classExists($class, false)) {
-                foreach ($aliases as $alias) {
-                    class_alias($class, $alias);
-                }
-            }
+        if (isset($this->mapping[$class]) && !$this->classExists($class) && $this->classExists($this->mapping[$class])) {
```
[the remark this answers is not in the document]
dpfaffenbauer: But it isn't necessary to vice versa load them. It actually works with this solutions as well. The ecommerce framework still uses `href` and `multihref` and it works.

dpfaffenbauer:
- `app`, `bin` and `web`: let's move them into .travis to keep the root directory clean from it

I don't agree with that. It should be possible to clone Pimcore and install it as you would do with the skeleton. This makes it easier to create PRs for it. Otherwise you are forced to either copy files around or use a skeleton which is tricky with composer if you are less experienced.

brusch:
@dpfaffenbauer Pimcore is now a library ... the folders are not necessary in the root.

lib/Bootstrap.php
dpfaffenbauer: done

dpfaffenbauer:
@brusch alright, moved them....

dpfaffenbauer:
@brusch  builds are passing, what do you think about the autoloading nowß

lib/Loader/Autoloader/AliasMapper.php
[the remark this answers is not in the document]
brusch: Not working: 
```
$foo = new \Pimcore\Model\DataObject\ClassDefinition\Data\ManyToManyRelation();
dump($foo instanceof \Pimcore\Model\DataObject\ClassDefinition\Data\Multihref);
```

Same for the object autoloader: 
```
$foo = new \Pimcore\Model\DataObject\News();
dump($foo instanceof \Pimcore\Model\Object\News);
```

brusch:
@dpfaffenbauer we have to leave the autoloaders as they are, see examples above, thanks!

lib/Loader/Autoloader/AliasMapper.php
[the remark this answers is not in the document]
dpfaffenbauer: Since that’s legacy stuff anyway, it would be time to remove it finally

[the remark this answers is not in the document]
brusch: 😁I expected that answer. 
Please revert the changes regarding the autoloader, as it is a different topic anyway. Thanks! 😊

dpfaffenbauer:
@brusch reverted the autoloader changes

aarongerig:
Isn‘t that legacy stuff like `Object` or `Multihref` something to be removed with v6.0?

brusch:
Let's see ...

bin/console
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env php
```
brusch: This file can be reverted as well, see `.travis/bin/console`. Thanks!

lib/Bootstrap.php
```diff
@@ -14,8 +14,10 @@
 
 namespace Pimcore;
 
+use Composer\Autoload\ClassLoader;
 use Doctrine\Common\Annotations\AnnotationReader;
-use Doctrine\Common\Annotations\AnnotationRegistry;
+use Pimcore\Loader\Autoloader\AliasMapper;
```
brusch: This `use` statements are currently useless,  we can leave them, but then e.g. `$legacyMappingLoader = new \Pimcore\Loader\Autoloader\AliasMapper($loader);` should use only the class name.

.travis/setup-tests.sh
```diff
@@ -2,41 +2,16 @@
...
+cp .travis/extensions.template.php var/config/extensions.php
+cp app/config/parameters.example.yml app/config/parameters.yml
 
 # install composer dependencies
-cd /tmp/www
-COMPOSER_DISCARD_CHANGES=true COMPOSER_MEMORY_LIMIT=-1 composer install --no-interaction --optimize-autoloader
-cd ~/build/pimcore/pimcore
\ No newline at end of file
+COMPOSER_MEMORY_LIMIT=-1 composer install --no-interaction --optimize-autoloader
+COMPOSER_MEMORY_LIMIT=-1 composer require symfony/symfony:$SYMFONY_VERSION --no-interaction --optimize-autoloader
```
brusch: I guess if we change the order and use `--no-update` for requiring `symfony/symfony`, we could speed things up a bit. 
```
COMPOSER_MEMORY_LIMIT=-1 composer require symfony/symfony:$SYMFONY_VERSION --no-interaction --no-update
COMPOSER_MEMORY_LIMIT=-1 composer install --no-interaction --optimize-autoloader
```

bin/console
dpfaffenbauer: done

lib/Bootstrap.php
dpfaffenbauer: done

.travis/setup-tests.sh
dpfaffenbauer: done

brusch:
Thanks a lot!

dpfaffenbauer:
🥳🍻

PR #4436, opened 2019-05-24
Issue when System Language contains underscore

eliasfellner:
Google CSE search doesn't accept language parameters such as "pl_PL" "fr_FR" etc, it wants the language to be "pl", "fr" etc.
With the fix only the part before the underscore is used.


brusch:
Thanks!

PR #4447, opened 2019-05-27
Es Bugfix - Bool filter

ckogler-elements:
If a property is defined as "bool" the value in the query has to be a bool or a string "true" or "false". 
Otherwise a parsing exception is thrown.

"0" or "1" is passed from a filterdefinition which  won't work. 
Changing the value in the backend from 0/1 to true/false didn't worked either...

### DIFF
--- a/bundles/EcommerceFrameworkBundle/FilterService/FilterType/ElasticSearch/MultiSelect.php
+++ b/bundles/EcommerceFrameworkBundle/FilterService/FilterType/ElasticSearch/MultiSelect.php
@@ -64,12 +64,14 @@ class MultiSelect extends \Pimcore\Bundle\EcommerceFrameworkBundle\FilterService
                 }
             }
 
-            if ($filterDefinition->getUseAndCondition()) {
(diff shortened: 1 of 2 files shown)


bundles/EcommerceFrameworkBundle/FilterService/FilterType/ElasticSearch/MultiSelect.php
```diff
@@ -56,14 +56,20 @@ public function addCondition(AbstractFilterDefinitionType $filterDefinition, IPr
                     $quotedValues[] = $v;
                 }
             }
-            if (!empty($quotedValues)) {
```
fashxp: the not empty check got lost now? is that not necessary?

ckogler-elements: ah... sorry - its required for the "else" block - readded the empty check

fashxp:
thx

PR #4450, opened 2019-05-27
[Element] removed ChildsCompatibilityTrait from all Elements

brusch:
see #4378

### DIFF
--- a/models/Document/Dao.php
+++ b/models/Document/Dao.php
@@ -206,7 +206,7 @@ class Dao extends Model\Element\Dao
      *
      * @return array
      */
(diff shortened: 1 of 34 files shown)


models/Document/Dao.php
```diff
@@ -206,7 +206,7 @@ public function updateWorkspaces()
      *
      * @return array
      */
-    public function updateChildsPaths($oldPath)
+    public function updateChildrenPaths($oldPath)
```
NiklasBr: Why not `updateChildPaths` seems like that would be more "correct" English (see also line 219 below)?

PR #4461, opened 2019-05-29
[DataObject] Remove data types BC layers - see #4378

blankse:
see #4378


brusch:
Very nice, thanks a lot!

PR #4463, opened 2019-05-29
Enable localized fields in object bricks

Lexipowder:
Now the getter functions in the localized fields are set in the objectbrick Definition.


weisswurstkanone:
thx!

PR #4472, opened 2019-05-31
[Webservices] Remove useless & dangerous urldecode from query filter

ramundomario:
Removed useless & dangerous urldecode, which could cause an unwanted transformation, when take the query from the request to filter the results.

## Additional info  
As indicated in the [PHP documentation](https://www.php.net/manual/en/function.urldecode.php#refsect1-function.urldecode-notes), the `$ _GET` and `$ _REQUEST` variables are already decoded, so the urldecode could accidentally transform the query.

For example the query `{"field":{"$like":"%6232%"}}` is transformed to
`{"field":{"$like":"b32%"}}` because `%62` represents the char `b`.
Below a screenshot with the first value corresponds to the query taken from the request and the second the value `urldecoded` two times.

Screenshot from 2019-05-31 14-32-00

The behavior can be reproduced on the latest demo, in all calls where filters are made:

- Search Assets
- Search Documents
- Search Objects
- Get Asset Count
- Get Document Count
- Get Object Count
- Get Classification Store Definition
- Get QuantityValue unit Definition

### DIFF
--- a/bundles/AdminBundle/Controller/Rest/AbstractRestController.php
+++ b/bundles/AdminBundle/Controller/Rest/AbstractRestController.php
@@ -302,7 +302,7 @@ abstract class AbstractRestController extends AdminController
      */
     protected function buildCondition(Request $request)
     {
-        $q = trim(urldecode($request->get('q')));
+        $q = trim($request->get('q'));
         if (!$q) {
             return;
         }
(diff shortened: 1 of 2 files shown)


weisswurstkanone:
thx!

PR #4475, opened 2019-05-31
[Core] Migrate System Config to Symfony Core

### LINKED ISSUE #3268: [Core] Migrate System Config from PHP to  Symfony Config

dvesh3:
Resolves #3268


dpfaffenbauer:
Awesome! Better if it would not need a system.yml at all. Just directly use Symfony config (app/config.yml) and inject parameters into the DI container and load it from there.

dvesh3:
changes in progress....this is already a topic for internal discussion with team 😉

dpfaffenbauer:
awesome :)

bundles/InstallBundle/Command/InstallCommand.php
```diff
@@ -189,9 +190,15 @@ protected function configure()
     protected function initialize(InputInterface $input, OutputInterface $output)
     {
         // no installer if Pimcore is already installed
-        $configFile = Config::locateConfigFile('system.yml');
-        if ($configFile && is_file($configFile) && !$input->getOption('ignore-existing-config')) {
-            throw new \RuntimeException(sprintf('The system.yml config file already exists in "%s". You can run this command with the --ignore-existing-config flag to ignore this error.', $configFile));
+        $configFile = Config::locateConfigFile('parameters.yml');
```
dpfaffenbauer: not necessary, all that information is in the container and can be loaded from there

dpfaffenbauer:
- must it use parameters.yml for system configs? It should be easier. Plain simple symfony configuration stored in the container with some default values if necessary.
 - Must it write into the parameters.yml on saving system settings? Can't it just create a symfony config somewhere in var/config that is included in app/config.yml?

bundles/InstallBundle/Command/InstallCommand.php
dvesh3: this returns parameters.yml file path to check if Pimcore is already installed and prompt user to provide --ignore-existing-config to proceed.

dpfaffenbauer: but the parameters.yml is copied from Pimcore\Composer before an installation actually happened.

dvesh3: this "/app/config/local/parameters.yml" is different from /app/config/parameters.yml

dpfaffenbauer: puh.... not sure if thats better.

Is this test even necessary? If it is: Why not check the DB if a certain table exists?

brusch:
@dpfaffenbauer 

> Better if it would not need a system.yml at all. Just directly use Symfony config (app/config.yml) and inject parameters into the DI container and load it from there.

This is what we did now ... 

If you'd like, you can defined everything in your `app/config.yml` if you prefer to use the UI, those values are stored as parameters.

bundles/InstallBundle/Command/InstallCommand.php
brusch: How should we check if a certain table exists without being able to establish a DB connection (because Pimcore is not installed) ... sounds like an interesting approach 🤔

brusch:
@dpfaffenbauer if you think you have a better approach, please let us know in detail (maybe with an example) - thanks!

lib/Bootstrap.php
```diff
@@ -285,6 +286,22 @@ public static function kernel()
         \Pimcore::setKernel($kernel);
         $kernel->boot();
 
+        $conf = \Pimcore::getContainer()->getParameter('pimcore.config');
+
+        if ($conf instanceof \Pimcore\Config\Config) {
```
brusch: @dvesh3 `$conf` is an array.

dpfaffenbauer:
@brusch that is not true, you are forced to use the parameters.yml, you can override it with a config in app/config. But still: you are forced to have the parameters set, otherwise you end up having symfony configuration exceptions since parameters are missing.

Better approach: define default values in the Configuration that are necessary. Only thing for me that comes to mind is actually just the language. (Maybe mailer configurations as well, but most of it can be empty or defaulted)

Within the Admin System Settings, write the configuration into something like /app/config/pimcore/system.yml (that is included from app/config.yml) in the same format you define the Configuration. That way, all the configs are loaded from Symfony Configurations without the need of the parameters.yml.

And you also have the possibility to completely ignore what is saved from Admin System Settings. Cause in many cases, you don't want the user to change anything in there in production anyway.

BTW: you could also define the `app/config/pimcore` folder as "dynamically" generated configs directory. So in the future, other bundles can also persist configurations from the backend in there. (example: Workflow GUI). That can be then easily shared between servers or releases (depending on the deployment strategy). Could be also something like `var/config/pimcore/`

The point is: It should be as close to a symfony standard as possible and don't require any forced parameters at all.

brusch:
Are you kidding me? This is what we had already ... the you wrote ... 

> Better if it would not need a system.yml at all. Just directly use Symfony config (app/config.yml) and inject parameters into the DI container and load it from there.

dpfaffenbauer:
....

yes, but you forced loaded it in code and not included it from another configuration.

brusch:
This is not what you wrote earlier ... anyway ... we'll see. 
Parameters have also their benefits ... and parameters are Symfony standard.

dpfaffenbauer:
What I mean is this: Since you want to have the ability to change those settings in the backend, you have to persist it somewhere on the file system. My suggestion was to introduce a directory for dynamically created configurations.

What I don't meant and don't think is good: To force load any configuration besides the only one symfony needs (`/app/config/config_[ENV].yml`)

For example, if you use the directory `app/config/pimcore` for dynamically generated configs, the developer can decide if he wants to load them or not. It's up to me or you to import `system.yml` in `app/config.yml`, or even just `app/config_prod.yml` if it makes sense.

So, the Backend would create the file `app/config/pimcore/system.yml`, which looks like this:

```yml
pimcore:
    general:
        timezone: 'Vienna'
        domain: 'pimcore.com'
        language: 'en'
        validLanguages: ['de', 'en']
....
```

And that config is imported from `app/config/config.yml`:

```yml
imports:
 - { resource: 'pimcore/system.yml' }
```

Works the same, but you are more flexible.

The current way this PR is implemented, forces you to have that many parameters defined, if they aren't defined, you'll end up with an exception for a missing parameter. I don't think this is good and could cause certain issues. And besides that, many of them have either default values or non values at all...

dvesh3:
@dpfaffenbauer Agree! this approach is flexible and loosely coupled from configuration tree and that's what introduced in earlier commits https://github.com/pimcore/pimcore/pull/4475/commits/5102177c2cfb167684da7bf0b9420ff54aebf7ba https://github.com/pimcore/pimcore/pull/4475/commits/8969430bc1d0fa5d7a1eee0530cf6e4533f036c4 . so we can step back to earlier approach of introducing system.yml and can be overridden in app/config/config.yml

.travis/web/app_test.php
```diff
@@ -31,7 +31,7 @@
 Tool::setCurrentRequest($request);
 
 // redirect to installer if pimcore is not installed
-if (!is_file(\Pimcore\Config::locateConfigFile('system.php'))) {
+if (!is_file(\Pimcore\Config::locateConfigFile('system.yml'))) {
```
dpfaffenbauer: What I described, is to remove this, and add it as import to the file `.travis/app/config/config.yml`. As well for every other occurrence of it

bundles/AdminBundle/Controller/Admin/SettingsController.php
```diff
@@ -326,7 +327,7 @@ public function getSystemAction(Request $request)
     {
         $this->checkPermission('system_settings');
 
-        $values = Config::getSystemConfig();
+        $values = Config::getSystemConfiguration();
```
dpfaffenbauer: I would also remove internal usages of Config and get the parameters directly from the Kernel. I'd still leave the Config class for BC reasons though, but it should trigger a deprecation notice for `getSystemConfig`

.travis/app/config/system.yml
```diff
@@ -0,0 +1,114 @@
...
+            browserapikey: ''
+    cache:
+        enabled: false
+        lifetime: null
+        excludePatterns: ''
+        excludeCookie: ''
+    webservice:
+        enabled: true
+    httpclient:
+        adapter: Zend_Http_Client_Adapter_Socket
```
brusch: `Zend_Http_Client_Adapter_Socket` should be just `Socket`

PR #4477, opened 2019-06-03
Improved isValidCartItem()

ThomasKeil:
In some rare cases the check "is there _something_" is not sufficient.
This fix improves the check.


bundles/EcommerceFrameworkBundle/CartManager/AbstractCart.php
```diff
@@ -929,7 +930,7 @@ protected function validateVoucherTokenReservations()
      */
     protected static function isValidCartItem(CartItemInterface $item)
     {
-        if ($item->getProduct() != null) {
+        if ($item->getProduct() instanceof ProductInterface) {
```
fashxp: we even could go for CheckoutableInterface there, couldn't we?

fashxp: @ThomasKeil  ping

Issue #3934, opened 2019-02-12
Custom layout lost after workflow transition (and F5) reload


solverat:
## Workflow
- Open object and apply custom layout "example"
- Dispatch workflow transition

**Current** **behavior**: Object gets reloaded with layout "master"
**Expected** **behavior**: Object should be available with custom layout "example" after reload
 
## UI
- Open object and apply custom layout
- Hit F5

**Current** **behavior**: Object gets reloaded with layout "master"
**Expected** **behavior**: Object should be available with custom layout "example" after reload

fashxp:
is this fixed now?

PR #4479, opened 2019-06-04
Make the custom layout pop up properly after page refresh.

AlekMosingiewicz:
## Issue this PR references  
#3934 
 - treat the empty layoutId parameter in the getAction in DataObjectController as a null parameter

## Additional info  
Basically, when you apply a custom layout to an object and ascribe this custom layout to a user or group, when you first log in as this user or group and open an object the layout is applied correctly. However, when you refresh the page the custom layout is gone. This is because upon refreshing the Pimcore frontend puts an empty-valued `layoutId` parameter in the request. Why it does that is a different matter, but this empty parameter is treated as an actual parameter - thus the layout set for the user or group is lost. This way or another, an empty parameter shouldn't be processed at all.

If the issue is wrong then tell me - I'll create a different one for this problem specifically.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -425,7 +425,7 @@ class DataObjectController extends ElementControllerBase implements EventedContr
             $validLayouts = DataObject\Service::getValidLayouts($object);
 
             //master layout has id 0 so we check for is_null()
-            if (is_null($currentLayoutId) && !empty($validLayouts)) {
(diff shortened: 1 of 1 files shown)


AlekMosingiewicz:
I've already signed the CLA - why is it listed as unsigned?

AlekMosingiewicz:
I'm trying to recheck but it doesn't work either.

brusch:
Your email address used in Git has to be the same as the one used for GitHub. 
You can check that with the following command: `git config --global user.email`
On GitHub you can configure multiple email address :)

AlekMosingiewicz:
@brusch Thanks - signed!

weisswurstkanone:
Makes sense, but the actual problem also needs to be fixed :-)
See #4490

PR #4483, opened 2019-06-05
Fix wrong translation in video editable

blankse:


brusch:
Please fix the translation here: https://poeditor.com/projects/po_edit?id_language=55&id=38068
Only EN is maintained in Git, all the other languages are on POEditor and exported from there. 
Thanks 😊

blankse:
@brusch I changed it there now :)

Issue #4441, opened 2019-05-24
Document editable Many-To-Many add reload option


ultramegatom:
Reload the page after the data change similar to how other editables handle this.

brusch:
Sounds good, can you provide us a PR for that? Should be an quite easy one :)

PR #4485, opened 2019-06-05
add reload option to relations editable

ultramegatom:
add reload option to relations editable

see https://github.com/pimcore/pimcore/issues/4441


brusch:
Thanks! 
Just one more thing, can you please also add it to the docs?

ultramegatom:
already did :)

brusch:
Just saw it, thanks!

Issue #4310, opened 2019-05-02
Master (Admin Mode) not editable fields in fieldcollections


ultramegatom:
You should be able to edit everything while in Master (Admin Mode).
If you make a Fieldcollection with not editable fields these fields stay not editable.

![image](https://user-images.githubusercontent.com/7206334/57069943-481a6000-6cd6-11e9-8577-79ac0fcec139.png)

PR #4491, opened 2019-06-06
Fieldcollections: Master (Admin Mode) should unlock not-editable fields

ultramegatom:
see https://github.com/pimcore/pimcore/issues/4310

I don't know if this causes any side effects so please double check this :)


weisswurstkanone:
thx!

PR #4492, opened 2019-06-06
[EcommerceFramework][Datatrans-Lightbox] - add upp-start-target attribute

AlternateIf:
in case of the datatrans lightbox the additional attribute upp-start-target should be set to '_top'. this is necessary since chrome apparently blocks redirects from within the lightbox iframe.  Such a redirect may happen to the Access Control Server of the bank which can lead to the following error:

'Unsafe JavaScript attempt to initiate navigation .... The frame attempting navigation is targeting its top-level window, but is neither same-origin with its target nor has it received a user gesture'

@fashxp


fashxp:
thx

Issue #3128, opened 2018-07-24
(Multi-)Href allow (DataObject)folder restriction


ultramegatom:
Currently there is no way to restrict a href field for (DataObject)folders. 
If you want to allow folders at all you have to allow all classes.

Assets and Documents allow this restriction just fine.

Would be nice if we could virtually add 'folder' to the classes list so the restrictions would be cleaner.

![image](https://user-images.githubusercontent.com/7206334/43134453-67665ca6-8f42-11e8-89ad-2a4a2a55b5b3.png)

PR #4496, opened 2019-06-06
Relations allow (DataObject)-folder restriction

ultramegatom:
see https://github.com/pimcore/pimcore/issues/3128


ultramegatom:
@dvesh3 has nothing to do with this pr. Tests were failing on all commits last week...

dpfaffenbauer:
Rebase it onto master, then the Tests should pass.

Issue #1634, opened 2017-06-22
Two roles with same workspace but different rights


janwalther:
### Expected behavior 
Rights should be additive
  
### Actual behavior  
Rights are not additive but one role defines access rights for workspace (which one seems unclear at the moment but the steps below reproduce the bug for sure), others are ignored
  
### Steps to reproduce  
1. Create role "write" with right for assets and add asset folder X as asset workspace and assign it all rights (list, view, publish etc.)
1. Create role "read" without right for assets and add asset folder X as asset workspace and assign only list and view rights
1. Create user and assign both roles
1. Login as new user and right click on folder X -> cannot upload

markus-moser:
fixed in pimcore 4 and pimcore 5: permissions are now additive in this case (and not "random" anymore)

brusch:
This topic needs to be revised again, see #4499 - seems we need a more sophisticated query to achieve the desired behavior for the scenario described above.

brusch:
Additionally we should think about the behavior when having multiple roles assigned to a user and the roles have colliding or overlaping workspaces.

weisswurstkanone:
Already touched by https://github.com/pimcore/pimcore/pull/6670

Current behavior is that:

- Longer cpath always has higher priority regardless of whether defined in any role or user workspace
- For same cpath user setting always takes preference
- If there is no user workspace setting for the cpath then ...
- Overlapping roles settings are then additive (e.g. read write for the cpath get merged)

PR #4499, opened 2019-06-07
[User Permissions] Fix wrong inheritance of wider permissions

ramundomario:
The problem appears when setting restrictive permissions on a child of an element to which wider permissions have been defined.
Restrictive permissions on the child are ignored.
The behavior can be reproduced on Documents, Assets and Objects.

## Steps to reproduce  
The following steps were performed on the current basic demo.
1. Create a user with wider permissions to parent document and restrictive permissions on child
users_workspaces_permissions

2. The user can rightly perform all operations on the parent document
homepage

3. The user can wrongly perform all operations on the child document even if more restrictive permissions have been defined
homepage_child

### DIFF
--- a/models/Asset/Dao.php
+++ b/models/Asset/Dao.php
@@ -465,7 +465,7 @@ class Dao extends Model\Element\Dao
         $userIds[] = $user->getId();
 
         try {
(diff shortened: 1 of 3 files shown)


brusch:
Regression of #1634
Needs further attention.

brusch:
Thanks a lot!

PR #4505, opened 2019-06-11
Relation data types: Default values and Type hints

blankse:
If you create class definitions via PHP it is possible to get an PHP error in the count functions in Relation::getPhpDocClassString method.

I think we should set default values an type hints so this couldn't happen anymore.


brusch:
Unfortunately this changes causes issues with existing class definitions. 
E.g. when `assetsAllowed` or `assetTypes` is `null` in the definition file, you're unable to load the class. 
So this would need a migration script in order to allow fluent updates.

PR #4508, opened 2019-06-11
Add hint about magic id/variable names

jdreesen:
I stumbled upon this while trying to add a new navigation item, so I figured I should add a hint in the docs.

Improvements in the wording are very welcome, though ;)


brusch:
Thank you!

PR #4514, opened 2019-06-12
[Docs] Fix extension of parent class for CustomRestController

ramundomario:
Fix extension of wrong parent class for a CustomRestController


brusch:
The documentation is about completely custom REST API's (e.g. consumed by frontends, ...) and not about extending the REST API's provided by the core, so the code is fine as it is at the moment. 
But of course it would make sense to mention the possibility to do so.

PR #4515, opened 2019-06-12
update endpoint

AlternateIf:
soon payment.datatrans.biz will be replaced by pay.datatrans.com. although currently both are still available it makes sense to switch to the new one


AlternateIf:
@fashxp

brusch:
Thanks!

PR #4524, opened 2019-06-13
Extend Validation Exeption

mkauferstein:
currently you cannot see which field throws an exeption this should fix this

  
  

Resolves #


fashxp:
thx

PR #4530, opened 2019-06-17
[Objects] Quantity value tooltip showing value in compatible units

BlackbitNeueMedien:
Sometimes it can be useful to know the value of a quantity value in a different unit. There is already automatic conversion but due to rounding errors, readonly fields and lacking overview (if you want to know the value in different units - e.g. multiple currencies) I added a quantity value tooltip showing all compatible units of a certain quantity value field:
Bildschirmfoto 2019-06-17 um 13 40 04
(This is a readonly quantity value field)

The units in the tooltip are only based on the base unit of the currently set unit of the quantity value field - not the allowed units. This is implemented on purpose because this way you can set allowed units for editing (e.g. SI units) and see their converted values (e.g. imperial units). Thus the quanitity value field does not get cluttered in all the different possible units but you can keep the data clean.


weisswurstkanone:
Does this mean that 
- it will override the static tooltip if there is one ?
- there is not way to disable this feature (maybe make it optional and turned off by default)

Any other opinions?

BlackbitNeueMedien:
Ok, I will add a configuration for this feature.

Currently static tooltip and this unit tooltip get shown. As the unit tooltip is positioned fixed on the right of the field and the static tooltip moves with the mouse position, imho they could coexist. If you do not want this, we could "merge" the tooltips by using the static tooltip and adding the converted units.

BlackbitNeueMedien:
Configuration added.

Now I merge the contents of the static tooltip (if present) with the alternative units tooltip:
Bildschirmfoto 2019-06-17 um 15 28 21

brusch:
Since calculating the values for the other compatible units isn't that expensive, I wouldn't do this right here: 
https://github.com/pimcore/pimcore/blob/5686f817c6806ebff2cc8ad0651554e094a05629/models/DataObject/ClassDefinition/Data/QuantityValue.php#L292 and deliver the alternative values directly with the initial data request. 
Less code, better UX, easier to maintain 🤓

@BlackbitNeueMedien can you update your PR accordingly? 

Thanks a lot!

BlackbitNeueMedien:
"deliver the alternative values directly with the initial data request."
@brusch Which values do you mean? The converted values? This would not make much sense as the numerical value of the quantity value field can be changed. Or do you mean sending converted values  initially to display as long as the numerical value has not been changed? And after change use AJAX request?

Or do you mean sending conversion factors + offsets with the initial request? This would not work if a conversion service class gets used.

BlackbitNeueMedien:
@brusch, have you seen my questions?
Either I misunderstood you or delivering converted values in the initial request would only add additional code (because when another value gets entered, the tooltip has to show updated values).

BlackbitNeueMedien:
Is there anything left to do here? Or isn't this feature useful at all, @brusch , @weisswurstkanone ?

brusch:
@BlackbitNeueMedien There was a misunderstanding on my end, sorry for that 🤪
Regarding the tooltip ... what do you think about an extra button (that get's only showed if there are additional compatible units available), that triggers the info window/tooltip? 
We can then get rid of the extra config option and the user is not getting annoyed by the tooltip but can trigger the info when really needed. 

image

BlackbitNeueMedien:
@brusch Good idea! Now a button gets added if there are convertible units.

brusch:
@BlackbitNeueMedien Thanks for updating the PR! 
Maybe we should only update the content of the tooltip when the button is actually pressed, this would avoid `async: false,` on the XHR and probably saves some requests. Or was there a particular reason for updating the content on change?

BlackbitNeueMedien:
@brusch The reason for updating the tooltip contents on change is that it is immediately present when the button gets clicked. The `async: false` was from the previous implementation, we actually do not need it now.
Of course I could change it to fetch the tooltip contents only after clicking the button but then the user would have to wait for the response. What do you think?

brusch:
Well ... in practice I don't think that the value of a field changes that often, so I guess we're fine with the current implementation.

brusch:
Thanks a lot for your work and for your patience 🙏

PR #4532, opened 2019-06-18
use urldecode for special characters

putzflorian:
use urldecode for special characters to find the file on the Filesystem.


fashxp:
thx

PR #4540, opened 2019-06-18
getBundle returns object so array_pop does not work

BlackIkeEagle:
And getBundle does only accept one parameter.

When getParent returns a Bundle and we want to get the bundle from the
kernel there is a warning

```
Warning: array_pop() expects parameter 1 to be array, object given
```

Because \Symfony\Component\HttpKernel\Kernel::getBundle will always
return an object or throw an exception.

``` php
/**
 * {@inheritdoc}
 */
public function getBundle($name)
{
    if (!isset($this->bundles[$name])) {
        $class = \get_class($this);
        $class = 'c' === $class[0] && 0 === strpos($class, "class@anonymous\0") ? get_parent_class($class).'@anonymous' : $class;

        throw new \InvalidArgumentException(sprintf('Bundle "%s" does not exist or it is not enabled. Maybe you forgot to add it in the registerBundles() method of your %s.php file?', $name, $class));
    }

    return $this->bundles[$name];
}
```

So the getBundle will return us the desired object immediatly.

Resolves #

### DIFF
--- a/bundles/CoreBundle/Templating/LegacyTemplateGuesser.php
+++ b/bundles/CoreBundle/Templating/LegacyTemplateGuesser.php
@@ -96,8 +96,7 @@ class LegacyTemplateGuesser extends BaseTemplateGuesser
                     $bundleName = $bundle->getName();
                     break;
                 }
-                $bundles = $this->kernel->getBundle($parentBundleName, false);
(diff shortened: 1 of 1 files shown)


fashxp:
thx

PR #4542, opened 2019-06-18
Fix #4495: user.classes column too short for custom identifieres

### LINKED ISSUE #4495: Roles/Users: allowed types to create is limited by varchar(255) column
# Bug Report

**Pimcore 5.6.6
MariaDB 10.1.14**

- [x] This is not a security issue / vulnerability -> use [this form](https://pimcorehq.wufoo.com/forms/pimcore-security-report/) instead
- [x] Behavior is reproducible on http://demo-basic.pimcore.org/admin/ (admin/demo)
- [x] You're not able to fix the problem yourself and send us a pull request instead of an issue.
- [x] There's no existing ticket for the same issue

### Expected behavior

It should be possible to store a reasonable amount of classes a User/Role is allowed to create and use custom unique class identifiers.
 
### Actual behavior

Our classes use speaking unique identifiers, i.e. `AcmeProductCompositionItem`.
We cannot configure roles with more than x allowed classes because the `users.classes` column currently is limited to `varchar(255)`. With numeric class ids this might not have been critical, but since Pimcore allows for unique class identifiers contain up to *64 characters* this breaks permission management.

### Steps to reproduce

* Have about 10 classes with rather long unique identifiers
* Create a new Role
* Select 10 classes in "Allowed types to create"
* Save & refresh role tab
* Not all previously selected classes are selected

Check also `users` table for the created role: The `classes` column contains truncated list of class identifiers.

Bildschirmfoto 2019-06-06 um 15 31 55

macghriogair:
Resolves #4495

### DIFF
--- a/bundles/CoreBundle/Resources/migrations/Version20190618154000.php
+++ b/bundles/CoreBundle/Resources/migrations/Version20190618154000.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+
+class Version20190618154000 extends AbstractPimcoreMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->addSql(
+            'ALTER TABLE `users` 
+            CHANGE COLUMN `classes` `classes` VARCHAR(1000) NULL DEFAULT NULL;'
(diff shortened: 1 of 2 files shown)


fashxp:
please also update the install.sql
https://github.com/pimcore/pimcore/blob/81a3a6df30fe7293105a5629dd53774ad5273f3c/bundles/InstallBundle/Resources/install.sql#L661

macghriogair:
@fashxp okay, I updated my PR.

brusch:
Changed the column to `text` makes more sense I guess :)

Issue #4543, opened 2019-06-18
[Bug] Check requirements : improve mysql check.


Yivan:
When installing Pimcore 6 on a mariadb 10.3 (stock debian buster package), the mysql check requirement for `innodb_large_prefix` was not respected (cli message during check requirement).
It said `innodb_large_prefix` was not activated, but it was set to "1".
I tried set it to "On" but same message. Setting it to "ON" was ok.
I think "ON" is the less used value, "1" is more common or "On".

Check is here:
https://github.com/pimcore/pimcore/blob/78b0966810fac687afa5dfacc98bf97299f8cd54/lib/Tool/Requirements.php#L110

Strangly, for `innodb_file_per_table` (which was setted to "1"), it passed !

I don't know if the bug is one db wrapper used or Pimcore, but I think the check on Pimcore side for `innodb_large_prefix` and `innodb_file_per_table` should be done against ['1', 'On', 'ON'] so all case are covered, and install will be smother ; )

Thanks.

PR #4545, opened 2019-06-18
Improve mysql requirements check

Yivan:
Thias pull request resolve https://github.com/pimcore/pimcore/issues/4543

### DIFF
--- a/lib/Tool/Requirements.php
+++ b/lib/Tool/Requirements.php
@@ -107,7 +107,7 @@ class Requirements
         $largePrefix = $db->fetchRow("SHOW GLOBAL VARIABLES LIKE 'innodb\_large\_prefix';");
         $checks[] = new Check([
             'name' => 'innodb_large_prefix = ON ',
-            'state' => ($largePrefix && !in_arrayi(['on', '1', 1], $largePrefix['Value'])) ? Check::STATE_ERROR : Check::STATE_OK
+            'state' => ($largePrefix && !in_arrayi($largePrefix['Value'], ['on', '1', 1])) ? Check::STATE_ERROR : Check::STATE_OK
         ]);
 
         $fileFormat = $db->fetchRow("SHOW GLOBAL VARIABLES LIKE 'innodb\_file\_format';");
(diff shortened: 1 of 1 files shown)


lib/Tool/Requirements.php
```diff
@@ -107,19 +107,19 @@ public static function checkMysql(ConnectionInterface $db)
...
             'name' => 'innodb_large_prefix = ON ',
-            'state' => ($largePrefix && $largePrefix['Value'] != 'ON') ? Check::STATE_ERROR : Check::STATE_OK
+            'state' => ($largePrefix && !in_arrayi($largePrefix['Value'], ['on', '1', 1])) ? Check::STATE_ERROR : Check::STATE_OK
         ]);
 
         $fileFormat = $db->fetchRow("SHOW GLOBAL VARIABLES LIKE 'innodb\_file\_format';");
         $checks[] = new Check([
             'name' => 'innodb_file_format = Barracuda',
-            'state' => ($fileFormat && $fileFormat['Value'] != 'Barracuda') ? Check::STATE_ERROR : Check::STATE_OK
+            'state' => ($fileFormat && !in_arrayi('barracuda', $fileFormat['Value'])) ? Check::STATE_ERROR : Check::STATE_OK
```
BlackbitNeueMedien: `$fileFormat['Value']` is not an array here, is it?
Do you mean `strtolower($fileFormat['Value']) !== 'barracuda'`
Or you could do `!in_arrayi('barracuda', [$fileFormat['Value']])`

Yivan: @BlackbitNeueMedien Thanks, yes you are true, i keeped arrayi because used everywhere in this method, but if it is not an array it will raise an error or notice i think as arrayi use array_map whose parameter must be an array.
So it is better to replace it like you suggest by: `strtolower($fileFormat['Value']) != 'barracuda'` (i keeped only `!=` because it was a non strict non egality in the original code, but strict non equality is ok too!)

Yivan: @BlackbitNeueMedien @brusch I just made a new commit to fix this array problem and add same principe of strtolower to utf8mb4 check in case of.
Should be ok now and ready to merge : )

PR #4549, opened 2019-06-19
String-Replace-Operator: Commit configured label

BlackbitNeueMedien:
Currently set label does not get saved but reverted to default "Operator String Replace".


weisswurstkanone:
thx!

PR #4550, opened 2019-06-19
[Cache] Avoid redirects cache stampede

### LINKED ISSUE #4531: Caching: high database usage after cache cleanup
When the redirect cache is invalidated (either from the admin panel, or via CLI), or it expires (default to 28 days), it gets automatically re-populated with the next http request.

If the redirects list is big enough (~14k entries here), and the website serves enough traffic, this creates a cache stampede effect, where each http request tries to warm the redirect cache, creating a peak of queries on the database:

image

The relevant code is https://github.com/pimcore/pimcore/blob/08b6245bcc828321f7555faf37ea546ed4ab2633/lib/Routing/RedirectHandler.php#L220-L229

Adding a mutex would help the stability of the system.

dvesh3:
Resolves #4531

### DIFF
--- a/lib/Routing/RedirectHandler.php
+++ b/lib/Routing/RedirectHandler.php
@@ -218,13 +218,12 @@ class RedirectHandler implements LoggerAwareInterface
         }
 
         $cacheKey = 'system_route_redirect';
-        if (!($this->redirects = Cache::load($cacheKey))) {
+        if (($this->redirects = Cache::load($cacheKey)) === false) {
             // acquire lock to avoid concurrent redirect cache warm-up
             Lock::acquire($cacheKey);
 
             //check again if redirects are cached to avoid re-warming cache
-            if (!($this->redirects = Cache::load($cacheKey))) {
-
+            if (($this->redirects = Cache::load($cacheKey)) === false) {
                 try {
                     $list = new Redirect\Listing();
                     $list->setCondition('active = 1');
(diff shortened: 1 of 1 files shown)


giannello:
This will prevent concurrency when caching the redirects, but won't fully solve the issue, as the requests will "queue" up and re-fill the cache as soon as the previous lock is released. Given the right conditions (time to warm the cache, and amount of requests per second) won't this smoothen out the database load by spreading it over a longer time span (potentially of making some requests time out)?

Shouldn't we acquire the lock only if it's not already acquired by another request, in which case we should just wait until the lock is release and proceed _without_ re-warming the cache?

dvesh3:
@giannello Thanks for your review! I’ve made changes now to force write cache for redirects as previously it was deferred write at shutdown. And also added an extra check after lock is acquired so subsequent requests will not re-fill the cache. 

I’m not sure about proceed without re-warming the cache as it could have some side effects. Let’s wait for @brusch suggestions.

brusch:
The patch works as long as there's no cache write lock set during the first request that should write the item to the cache. If there's a write lock, the item doesn't get into the cache, so the 2nd (or any of the other requests) will not get the contents out of the cache, so we're in the same situation. 

Related to: #4573 

Let's wait with this PR until we have an answer to #4573 then we can decide if this is the right approach to solve the issue.

Issue #3723, opened 2018-12-07
[Asset] Image Editor: increase size of image that get loaded into the editor


Yivan:
Hello,

As Pimcore admin logged in the Pimcore backend, using miniPaint image editor on an image asset, the opened image is not open with the good resolution.
Concequence is that if the image is save just after being opened in the image editor, the new image lost quality as resolution is splitted by a factor 2 it seems.
So if an enduser use the editor, he will degrade the resolution of the assets... : ( (pretty dangerous for system in production, actually the only way to prevent this is to disable temporarly the image editor).

Here is an opened image on the Pimcore demo:

![image-editor](https://user-images.githubusercontent.com/9025839/49655788-e5c97600-fa3b-11e8-92e0-dbc32be65700.png)

Here is the same image opened in the online demo of miniPaint:

![image-editor-demo](https://user-images.githubusercontent.com/9025839/49655798-ecf08400-fa3b-11e8-8ac4-134f9ba549da.png)

So the problem seems to be on the Pimcore side i think.

Thanks.

brusch:
That's acutally no bug, it is necessary that Pimcore converts images to a smaller size and to a format miniPaint can read. It used to be 1000x1000 pixels at maximum, I've increased it now to 3000x3000 which seems to work on an average PC.

Yivan:
@brusch Thanks for your explaination and fix.

I understand this is a miniPaint limitation. Can i suggest the folowing:
- if image is too big to display a message saying "_Be carreful: picture was too big and has been reduced to XXX x YYY pixels_". Because if user just open the image, make a small modification and save it, he can miss the fact that resolution of the image asset has been reduced.
I really think it would be safer to inform the user via such an alert message, so he doesn't degrate asset resolution unintentionnaly.

Thanks.

Yivan:
@brusch 
Sorry to come back on this, we are using more and more pimcore as a couple PIM/website, so image asset have sometime really big resolution. I know it has been extended to 3000x3000, but with PIM sometimes there a more bigger images.
And actually customer can destroy the image quality (downsizing the resolution) without any message informing it. It is problematic for a PIM centralized image. Please read my previous comment. Do you think it is a good idea to prevent this ?

Thanks for reading me.

fashxp:
@Yivan  could you provide a PR for it?

Yivan:
@fashxp Yes, ok i will try to make it.
Thanks!

Yivan:
[...]

PR #4551, opened 2019-06-20
Fix max image size in image editor - No more limits

Yivan:
Hello,

This pull request resolve https://github.com/pimcore/pimcore/issues/3723

There is no more limits of 3000x3000.
Test on image resolution of less 3000x3000 and more than 3000, saving it on the image editor preserve original image size.
Resizing up the image work well too.

Thanks

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/Asset/imageEditor.html.php
+++ b/bundles/AdminBundle/Resources/views/Admin/Asset/imageEditor.html.php
@@ -70,7 +70,7 @@
 <div id="popup"></div>
 
 
-<img style="visibility: hidden" id='image' src='/admin/asset/get-image-thumbnail?id=<?= $this->asset->getId() ?>&width=3000&height=3000&contain=true'/>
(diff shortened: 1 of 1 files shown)


brusch:
Thanks @Yivan for your PR. 
Before we merge this PR I guess we should fist check which limitations miniPaint has.
If there are limitations on the amount of pixels in an image we should check them and display an info message in case an image is too big for miniPaint, we should then give the user the choice to cancel the editing or continue with a downsized image. 

I've tried to find our something about the limitations of miniPaint, but haven't found anything, maybe you can check it again or make your own experiences. 

Thanks a lot!

Yivan:
@brush i didn't find any limitation mentionned.
I trid with a 10000x10000 png image. Open it on https://viliusle.github.io/miniPaint/ , add some pencil stroke ; ) and save it.
It was ok, no limitation. The resulting image make 10000x10000.
I wanted to try 20000x20000 but my computer was in pain for memory !!!
So i think the limitation would be computer memory before miniPaint limitation ; )

brusch:
Ok, thanks for checking. 
I've just pushed ff2ee54f06c92ae6cf5781b33aeb88a78ed7c793 with some further optimizations.

Yivan:
@brush Thanks, and nice addition the file format conversion !

PR #4564, opened 2019-06-24
Output document thumbnail with correct content-type header

BlackbitNeueMedien:
When there is an error during document thumbnail generation Pimcore tries to load the SVG `PIMCORE_WEB_ROOT . '/bundles/pimcoreadmin/img/filetype-not-supported.svg'` (see https://github.com/pimcore/pimcore/blob/43e14cca6b5f9171ffc4abc21b1921b23feec33b/models/Asset/Document/ImageThumbnail.php#L123)

But in https://github.com/pimcore/pimcore/blob/43e14cca6b5f9171ffc4abc21b1921b23feec33b/bundles/AdminBundle/Controller/Admin/AssetController.php#L1329-L1332 there is always Content-Type image/png being set - also for the above SVG.


brusch:
Thanks!

PR #4568, opened 2019-06-25
[ElementServices] Extend functionality

aarongerig:
This pull request makes it possible to use the copy methods within `Document\Service`, `Asset\Service` and `DataObject\Service` without an existing Pimcore user (e.g. request from Front-End).

Also the `copyRecursive` methods now also copy unpublished Documents and DataObjects as well as DataObject variants.


aarongerig:
Thank you!

Issue #3327, opened 2018-09-09
Unable to install Pimcore on custom PIMCORE_ENVIRONMENT variable.


Yivan:
Hello,

Running `COMPOSER_MEMORY_LIMIT=3G composer create-project pimcore/demo-basic .` on an environment having PIMCORE_ENVIRONMENT set to `int` (this is our integration server) leads to an error:

`PHP Fatal error:  Uncaught Symfony\Component\Config\Definition\Exception\InvalidConfigurationException: File /var/www/html/app/config/config_int.yml  cannot be found. in /var/www/html/vendor/pimcore/pimcore/lib/Kernel.php:115`

We are unable to deploy fresh install of Pimcore on those environments... :-(

Here is all the cli text:

```
Installing pimcore/demo-basic (v1.0.4)
  - Installing pimcore/demo-basic (v1.0.4): Downloading (100%)
Created project in .
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 138 installs, 0 updates, 0 removals
  - Installing wikimedia/composer-merge-plugin (v1.4.1): Downloading (100%)
  - Installing ocramius/package-versions (1.3.0): Downloading (100%)
  - Installing endroid/installer (1.0.5): Downloading (100%)
  - Installing psr/cache (1.0.1): Downloading (100%)
  - Installing cache/tag-interop (1.0.0): Downloading (100%)
  - Installing cache/integration-tests (0.16.0): Downloading (100%)
  - Installing symfony/polyfill-ctype (v1.9.0): Downloading (100%)
  - Installing symfony/polyfill-mbstring (v1.9.0): Downloading (100%)
  - Installing twig/twig (v2.5.0): Downloading (100%)
  - Installing paragonie/random_compat (v9.99.99): Downloading (100%)
  - Installing symfony/polyfill-php70 (v1.9.0): Downloading (100%)
  - Installing symfony/polyfill-util (v1.9.0): Downloading (100%)
  - Installing symfony/polyfill-php56 (v1.9.0): Downloading (100%)
  - Installing symfony/polyfill-intl-icu (v1.9.0): Downloading (100%)
  - Installing symfony/symfony (v3.4.15): Downloading (100%)
  - Installing symfony/polyfill-apcu (v1.9.0): Downloading (100%)
  - Installing psr/simple-cache (1.0.1): Downloading (100%)
  - Installing psr/log (1.0.2): Downloading (100%)
[...]

PR #4574, opened 2019-06-26
Add notes to install documentation for custom PIMCORE_ENVIRONMENT

Yivan:
Hello,

Thias pull request add notes to documentation, so less problem could occurs in case of custom PIMCORE_ENVIRONMENT.
(see https://github.com/pimcore/pimcore/issues/3327)

Thanks.

### DIFF
--- a/doc/Development_Documentation/01_Getting_Started/00_Installation.md
+++ b/doc/Development_Documentation/01_Getting_Started/00_Installation.md
@@ -39,6 +39,8 @@ COMPOSER_MEMORY_LIMIT=-1 composer create-project pimcore/demo-basic-twig my-proj
 COMPOSER_MEMORY_LIMIT=-1 composer create-project pimcore/demo-ecommerce my-project
 ```
 
+> Note: if installation is made on a custom PIMCORE_ENVIRONMENT value (different than those provided, see [app/config](https://github.com/pimcore/skeleton/tree/master/app/config), you need to prepare before the `app/config/config_[env_name].yml` file or to run the command like this `PIMCORE_ENVIRONMENT= COMPOSER_MEMORY_LIMIT=-1 composer create-project ...`
+
 Point the document root of your vhost to the newly created `/web` folder (eg. `/your/project/web`).
 Keep in mind, that Pimcore needs to be installed **outside** of the **document root**.
 Specific configurations and optimizations for your webserver are available here:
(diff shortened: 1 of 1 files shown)


brusch:
I'm wondering if this is the right place for this information, since this page is mainly for Pimcore newbies that don't know what the additional note actually means and we don't want to confuse them :| 

Maybe it's better to put the note here https://pimcore.com/docs/5.x/Development_Documentation/Deployment/Multi_Environment.html ?

Yivan:
@brusch I don't know. I would prefer to have it on the installation page. When installing a new Pimcore, i go to this page, and i think other people do like this (experienced or not) to refer if some things changed for installing Pimcore. I hope to find all information on this same page to be able to install Pimcore even if multienv is activated. I think to have related information in same place is better.

Another way, would be to put it in `https://pimcore.com/docs/5.x/Development_Documentation/Deployment/Multi_Environment.html` and refer to it in the installation page. But i thin as it is command for installing, it should really directly be in the installation page documentation.

Tell me what you prefer, and i can change my PR if needed.

Yivan:
@brusch 
Thanks for your modification. But the real anoying problem is you need a custom skeleton...
**Composer cannot create project if directory is not empty**, so you cannot prepare the file before the create composer command.
Making `PIMCORE_ENVIRONMENT= COMPOSER_MEMORY_LIMIT=-1 composer create-project ...` resolves nicely this problem, as just after this you can add this `app/config/config_[env_name].yml` and go on the install process.

Maybe you could just change the sentence you added by:
```
If you're running the installation using a custom environment name, ensure you have already the right config files in place, e.g. app/config/config_[env_name].yml. You can create the composer project by using command `PIMCORE_ENVIRONMENT= COMPOSER_MEMORY_LIMIT=-1 composer create-project ...` which will unset PIMCORE_ENVIRONMENT just for the creation of the project.
```

At least people won't have to make a custom skeleton to create the composer project. And it is a nice tip for Continous Integration where installing fresh Pimcore.

dpfaffenbauer:
I don't see any problem here, this is perfectly normal Symfony behaviour, if you need custom envs in your setup, just create a custom skeleton, there is nothing wrong with it.

Yivan:
@dpfaffenbauer Thanks for your insights!
I would say it is more composer than synfony related (being able to bypass create-project to be able to add the config file).
I never done this, maybe it should be documented or at least point to the synfony doc explaining it ?
I just looked into what does exactly "create-project, and i found it here: `https://getcomposer.org/doc/03-cli.md#create-project`

So I imagine to install Pimcore on custom PIMCORE_ENVIRONMENT, it is something like this:
- `git clone https://github.com/pimcore/skeleton.git .`
- add app/config/config_[env_name].yml
- `COMPOSER_MEMORY_LIMIT=-1 composer install`

In all case having it documented could only help people to install Pimcore ; ) I doesn't hurt.
(this way by git clone + composer install, or this way `PIMCORE_ENVIRONMENT= COMPOSER_MEMORY_LIMIT=-1 composer create-project ...`)

@brush if you think it is valuable information i can make a new PR (just tell me what you wish for it), or we can leave this issue closed (people could always refer to this github issue to have some hints).

PR #4580, opened 2019-06-27
Support for exclude pattern feature in case of Multiple Websites

avinashpimcore:
Please check pimcore/pimcore/issues/4579


brusch:
Unfortunately this change is a major BC break therefore I cannot merge it as of now. 
Probably it's better to use a custom event listener anyway instead of the very limited exclude pattern functionality. 
In an custom event handler you can inject the cache listener service and disable the cache as needed ... much more flexible ;)

PR #4581, opened 2019-06-27
[Custom layouts] Prevent SQL error for non-numarical ID

BlackbitNeueMedien:
Nowadays custom layouts can have non-numerical IDs. In https://github.com/pimcore/pimcore/blob/eebc75491fffdc848d9435668f874d387b3cefa2/models/DataObject/Service.php#L913 they are used in an SQL query's IN-criteria but without quoting them. This results in an SQL error.

To reproduce:
1. Create custom layout with non-numerical ID
2. Assign custom layout to a user's workspace settings
3. Login with this user
4. Try to open an object where workspace settings from 2. apply


brusch:
Thanks!

PR #4586, opened 2019-06-29
[Notifications] Role permission are not inherited

prugala:
## Bugfix
Users without `notifications` permissions  with a role that has permission are not listed on recipient list.
Same problem is when we want to send notification to group, users without permission dont get the message.

## Steps to reproduce
1. Add Role with `notifications` permissions
2. Add user without `notifications` permissions and add him to created role
3. Try to send notification:
3.1. Try to send notification to him - he isnt listed on recipient list
3.2. Send notification to created group - used dont get the message

### DIFF
--- a/models/Notification.php
+++ b/models/Notification.php
@@ -32,12 +32,12 @@ class Notification extends AbstractModel
     protected $id;
 
     /**
-     * @var int
+     * @var string
      */
     protected $creationDate;
 
     /**
-     * @var int
+     * @var string
      */
     protected $modificationDate;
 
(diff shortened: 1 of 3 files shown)


models/Notification.php
```diff
@@ -32,12 +32,12 @@ class Notification extends AbstractModel
     protected $id;
 
     /**
-     * @var int
+     * @var string
```
fashxp: why can a date be a string?

prugala: It was string, I fixed docblock only. If you want I can change on DateTime object,

fashxp: hmm, in DB it is a timestamp therefore an INT, isn't it? 
so we should change all typehints to INT or - since it is not main focus of this PR - just leave it as it was

PR #4591, opened 2019-07-01
Prevent deadlock when updating sibling indexes

BlackbitNeueMedien:
When you move multiple items in the data object tree to another folder their `o_index` values get updated. But as requests for all moved objects are sent in parallel a deadlock can occur (resulting in an SQL error). This PR catches those deadlocks and tries to re-execute the database updates.
(The simultanous update of all `o_index` values got introduced in #4381)


brusch:
Thanks!

PR #4592, opened 2019-07-01
Minor config improvements in getWebsiteConfig()

NiklasBr:
Removes duplicate if().
Does stricter comparison for site ids (which should always be integers).
Fetches $key later, after some continue's to avoid unnecessary execution.

  
  

Resolves #

### DIFF
--- a/lib/Config.php
+++ b/lib/Config.php
@@ -190,11 +190,10 @@ class Config
 
                     $key = $item->getName();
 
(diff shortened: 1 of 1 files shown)


NiklasBr:
It is unclear to me why some tests which appear to be unrelated failed here,

fashxp:
please rebase to master ... its related to new version of guzzle, see https://github.com/pimcore/pimcore/issues/4595

PR #4593, opened 2019-07-01
[Translation] Allow trans to support plurals & Fix transChoice

### LINKED ISSUE #4587: [Pimcore6] Undefined method "choose" in IdentityTranslator
Because of changing the MessageSelector class to IdentityTranslator, the `choose` method here:

https://github.com/pimcore/pimcore/blob/16c1d538462ca4b348aa447abd4ba0cc41d0242c/lib/Translation/Translator.php#L154

does not exist anymore:

```
  [Error] Call to undefined method Symfony\Component\Translation\IdentityTranslator::choose()  
```

dvesh3:
Resolves #4587

### DIFF
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -30,7 +30,10 @@ use Symfony\Contracts\Translation\TranslatorTrait;
 
 class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleAwareInterface
 {
-    use TranslatorTrait;
+    use TranslatorTrait {
+        trans as protected doTrans;
+    }
+
     /**
      * @var TranslatorInterface|TranslatorBagInterface
      */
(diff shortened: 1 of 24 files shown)


lib/Translation/Translator.php
```diff
@@ -24,10 +24,16 @@
 use Symfony\Component\Translation\MessageCatalogue;
 use Symfony\Component\Translation\MessageCatalogueInterface;
 use Symfony\Component\Translation\TranslatorBagInterface;
-use Symfony\Component\Translation\TranslatorInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
+use Symfony\Contracts\Translation\LocaleAwareInterface;
+use Symfony\Contracts\Translation\TranslatorTrait;
 
-class Translator implements TranslatorInterface, TranslatorBagInterface
+class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleAwareInterface
```
brusch: We should also additionally implement the legacy interface `Symfony\Component\Translation\TranslatorInterface` here, same as Symfony does, see: 
https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Translation/Translator.php#L32

Otherwise it would be a proper BC break :)

dpfaffenbauer:
does this also work properly with Symfony 3.4? Does the Contract exist there as well? Not sure about that...

dpfaffenbauer:
is it also possible to add a test for that? So it can fail in the future if symfony changes something here again

PR #4610, opened 2019-07-03
[Quantity Value] Error when auto-converting unset unit

BlackbitNeueMedien:
Steps to reproduce:
1. Create 2 quantity value units "g" and "kg"
1. Set "kg" as base unit of "g", set conversion factor to 0.001
1. Create data object class with quantity value field, set allowed units to "kg" and "g", enable auto-conversion
1. Open / Create new object
1. Change quantity value field's unit
This results in an error because previously there was no unit being set and the controller returns `null`.


weisswurstkanone:
thx!

PR #4612, opened 2019-07-03
[Localization] Add command to clean unused locale data

### LINKED ISSUE #4525: [Bug] When removing a langage, corresponding sql views are not removed.
Hello,

When a language is removed in the GUI Pimcore (system settings), the corresponding object SQL views are kept.
Problem occurs if a field is removed from a class. The object SQL view for the no more configured langage is not updated and try to get the old column which doesn't exist anymore.
As one of the consequence for instance, the database could no more be exported (myssql send an error because the SQL view call an unexisting column).

How to reproduce:
- set 2 languages in system config
- create a class with at least two attributes
- first, remove one of the language
- then, remove one of the attribut
Try an SQL dump (or just click on the view object_query_<classname> on adminer it will send an sql error)

Thanks.

dvesh3:
Resolves #4525

## Additional info  
Localized tables and views are retained even if a language is removed from Pimcore GUI -> system settings -> Localization & Internationalization (i18n/l10n) or overriding `valid_languages` in system.yml.

This PR introduces a new command to cleanup unsued locale/language data: `bin/console pimcore:locale:delete-unused-tables`


bundles/CoreBundle/Command/DeleteUnusedLocaleDataCommand.php
```diff
@@ -0,0 +1,83 @@
...
+        $validLanguages = Tool::getValidLanguages();
+
+        $db = Db::get();
+
+        $tables = $db->fetchAll("SHOW FULL TABLES LIKE '%object_localized_%'");
+
+        foreach ($tables as $table) {
+            $table = array_values($table);
+
+            if (preg_match('/^object_localized_[0-9]+_/', $table[0]) ||
```
dpfaffenbauer: Class IDs can be strings as well, so this will not work then right?

dvesh3: this leads to more testing scenarios...thanks for the catch! 😉

bundles/CoreBundle/Command/DeleteUnusedLocaleDataCommand.php
```diff
@@ -47,37 +47,90 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        $db = Db::get();
```
dpfaffenbauer: inject `Connection` instead?

PR #4630, opened 2019-07-05
[Templating] simplify condition and remove unused variable

jdreesen:
This simplifies the code of #4628 a bit.

### DIFF
--- a/bundles/CoreBundle/Templating/LegacyTemplateGuesser.php
+++ b/bundles/CoreBundle/Templating/LegacyTemplateGuesser.php
@@ -47,10 +47,8 @@ class LegacyTemplateGuesser extends BaseTemplateGuesser
      */
     public function guessTemplateName($controller, Request $request, $engine = 'twig')
     {
-        if ($request->attributes->has('_template')) {
-            if ($template = $request->attributes->get('_template') instanceof TemplatePhp) {
-                $engine = 'php';
-            }
+        if ($request->attributes->get('_template') instanceof TemplatePhp) {
+            $engine = 'php';
         }
 
         //first lookup for new template name(snake_case)


bundles/CoreBundle/Templating/LegacyTemplateGuesser.php
```diff
@@ -47,10 +47,8 @@ public function __construct(KernelInterface $kernel, DelegatingEngine $templateE
      */
     public function guessTemplateName($controller, Request $request, $engine = 'twig')
     {
-        if ($request->attributes->has('_template')) {
-            if ($template = $request->attributes->get('_template') instanceof TemplatePhp) {
-                $engine = 'php';
-            }
+        if ($request->attributes->has('_template') && $request->attributes->get('_template') instanceof TemplatePhp) {
```
jdreesen: Strictly speaking, the `$request->attributes->has('_template')` isn't needed here.

jdreesen: Another possibility would be to use the real `engine` value of the annotation instead:
```suggestion
        if ($request->attributes->has('_template')) {
            $template = $request->attributes->get('_template');

            if ($template instanceof \Sensio\Bundle\FrameworkExtraBundle\Configuration\Template) {
                $engine = $template->getEngine();
            }
        }
```

dvesh3: $engine property is removed from \Sensio\Bundle\FrameworkExtraBundle\Configuration\Template in 5.x version as now only Twig is supported so this will not work.

dvesh3: go ahead! :)

jdreesen: Oh okay, I didn't think of that.

PR #4635, opened 2019-07-08
Start session for ecommerce indexer when run in cli

das-peter:
## Issue

If indexing products triggers the `\Pimcore\Bundle\EcommerceFrameworkBundle\PricingManager\PricingManager` and you run in cli you encounter an error like:
> Error Failed to start the session because headers have already been sent by "/var/www/pimcore/vendor/pimcore/pimcore/bundles/EcommerceFrameworkBundle/IndexService/Tool/IndexUpdater.php" at line 176.

This is because the logging in `IndexUpdater` begins before the `PricingManager` can start its required session.

### Solution

Start a session, if required, in `IndexUpdater::updateIndex()`.
I think it's fair that the `IndexUpdater` takes care of that as it seems to be mainly used in the cli by `Pimcore\Bundle\EcommerceFrameworkBundle\Command\IndexService` and in our case we trigger that using migrations too.

Added code to check for cli and existing session in `IndexUpdater::updateIndex()`. If cli & session isn't started we start a session.


das-peter:
Fixed merge conflict by "re-basing" it on latest master and squashing changes into one commit.

fashxp:
why did you add the session start at two places? 
bootstrap command is calling index updater anyway, isn't it?

das-peter:
@fashxp Oh, that's not intended, I'll try to fix that merge issue. Just a sec please.

das-peter:
Here we go. I've no clue how all that was that messed up so badly :|

dpfaffenbauer:
ecommerce framework is non of my business: but: I don't think it is a good idea to start a session in a CLI command, better to fix the session issue where it appears. And: You should use DI to inject the session if you are going with this solution

fashxp:
we're going to refactor that indexupdater class anyway - make it a service, get rid of static methods, use proper logging, etc. 
for now I would say we accept that as a workaround.

das-peter:
I agree with both here. Should be ultimately fixed appropriately in the pricing system / index handler but we need a stop gap for the time being :)
Thanks for the feedback & merge 👍

PR #4639, opened 2019-07-08
Show progress in Index Updater log.

andreas-gruenwald:
As the index updater can run very long, it is helpful to show the progress in %.


fashxp:
thx

andreas-gruenwald:
@fashxp  I just noticed two issues regarding the update.

1. $iPos is not incremented.
2. if a total of zero records will be processed, division by zero occurs.

I can send an update by tomorrow, or you may you update the code accordingly.

PR #4649, opened 2019-07-09
[Newsletter] Default URL prefix

### LINKED ISSUE #4343: [Newsletter] Resolved host url is not correct
# Bug Report
When sending a newsletter, the host url gets resolved incorrectly. This is because `[\Pimcore\Tool::getHostUrl()](https://github.com/pimcore/pimcore/blob/master/lib/Tool.php#L452)` can't resolve the current request.

The host url is used to for example prefix image resources. Now, if you are sending the newsletter from a secure domain, the `https` protocol is ignored and `http` is used to process the newsletter template. This causes a lot of issues, some email clients don't even display "insecure" image resources.

Also if you're using the `www` prefix for your website's domain, that also gets ignored. That shouldn't be a big issue though.

### Expected behavior
Images getting prefixed with the correct host url, e.g. `https://www.domain.com`

### Actual behavior
Images getting prefixed with the wrong host url, e.g. `http://domain.com`

### Steps to reproduce

1. Go to https://demo-basic.pimcore.org/admin and sign in
2. Open the newsletter template `newsletters / example-mailing`
3. Send a newsletter to a random email address (doesn't matter which one)
4. Preview the HTML email in the "Sent Emails" tab and inspect the large image with the palm trees in it

aarongerig:
Resolves #4343 

This PR introduces a new configuration parameter, which allows to define a default URL prefix for static assets within a newsletter template.


brusch:
Thanks a lot!

PR #4650, opened 2019-07-09
Update GitHub pull request template to new Pimcore version

jdreesen:
Updates the GitHub PR template to the new version 6 of Pimcore.

Is Pimcore 4 still maintained and should be kept in the template or is it EOL and can be removed?


brusch:
Thanks!
v4 is history :)

PR #4651, opened 2019-07-09
Iframe Layout Component: Correction of height and distances

blankse:
### Fixes
- Height of iframe
- Add missing margin-bottom
- Remove debug message in console

### Before
before

### After
after


brusch:
Nice, thanks!

PR #4652, opened 2019-07-09
[DataObjects] Fixed language restriction on collections

### LINKED ISSUE #3753: Users with restricted access to languages can access all localized fields in fieldcollections
# Expected behaviour
When a user is not allowed to edit or view certain languages (as defined under "special settings" in his workspace) , this should apply to all localized fields

# Actual behaviour
Localized fields within fieldcollections ignore the settings

# Steps to reproduce
* Create a class with localized fields both at class level and in a fieldcollection
* Create a user and assign him a workspace. Under "special settings", select a subset of the languages to view and edit
* Log in as the user, create or open an object. You will see the settings applied at class level, while in the fieldcollection, all languages are visible and editable

Here, I have set up the user to be able to view english and german only. He also should only be able to edit the german language fields.
Restricted view

dvesh3:
Resolves #3753

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -728,6 +728,18 @@ class DataObjectController extends ElementControllerBase implements EventedContr
                             unset($objectData['data'][$key]['data'][$language]);
                         }
                     }
+                } else if ($fd instanceof DataObject\ClassDefinition\Data\Fieldcollections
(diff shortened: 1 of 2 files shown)


weisswurstkanone:
IMHO, this is not needed anymore as it is now covered by your changes as well

https://github.com/pimcore/pimcore/blob/23948c29c2d18f8f2e11233f8a9299167b36e10b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L517

Can you remove pls ?

dvesh3:
Resolved...Thanks!

PR #4655, opened 2019-07-10
don't allow folders in ManyToManyObject Relations Search Dialog

dpfaffenbauer:
Without this fix, it always shows every folder in the Search Dialog within a ManyToMany Object Relation


fashxp:
thx

Issue #4411, opened 2019-05-21
[Data Objects] WYSIWYG data-type: use <br> instead of <p>


BlackbitNeueMedien:
Currently the inserted text of an wysiwyg field of a data object gets wrapped in a p-tag by default. We can disable this by setting `editor configuration` of the field definition to
```json
{
  enterMode:2
}
```

In our experience only few customers are aware of this "feature", most just recognize it when output layout gets scrambled.

Imho it would be better to set enterMode: 2 by default as it is much easier to wrap the output in a p-tag whereever needed than to remove the p-tag afterwards.

What do you think?

ckemptner:
Please undo this change asap. It is now not possible anymore to add a new paragraph by hitting the "Enter" key. (Which is the default text editing behavior in all major text editing software such as Word or Wordpress - Why make it different in Pimcore than how most users know it?)
I would expect that such a change gets voted on by asking at least some community content editors -before implementing it - as this is a major change in the of content editor experience.

NiklasBr:
I'm in agreement with @ckemptner

brusch:
Feel free to continue the discussion, until v7 we have some time left.

BlackbitNeueMedien:
The question is what should be default behaviour. 

If you want the old behaviour with p-tags back you can enter 
`{
  enterMode:1
}`
in the wysiwyg field's editor configuration.

katharina-kopf:
Please undo this change asap! I work a lot in pimcore and I prefer adding a paragraph by hitting the "Enter" key - the same way as it works in word!

NiklasBr:
> p-tags can be counter intuitive in my (and a lot of our customers') experience.

This is where my experience is the exact inverse of yours, customers expect consistency and paragraphs. And the default behaviour of other systems is now different from Pimcore.

BlackbitNeueMedien:
There are 2 problems which initially got me to issue this: 
1. When you only have one line in a wysiwyg field, e.g. `This is <b>bold</b>` and press save it gets converted to `<p>This is <b>bold</b></p>`. It is not possible to enter something without the p-tags - they always got wrapped around the content.
2. When you set a wysiwyg field's content by code, e.g. `$object->setXyz('This is <b>bold</b>');` the content really is `This is <b>bold</b>` but when you later open the object in the Pimcore admin GUI and save the object, it gets converted as in 1. Especially this behaviour is problematic because nobody actually changed the content of the field via GUI and nevertheless its content in the database changed.

[...]

PR #4656, opened 2019-07-10
[Documentation} Wysiwyg field configuration -> enterMode

BlackbitNeueMedien:
As discussed in #4411 there was a BC break in 6.0 because the default `enterMode` got changed. This PR explains the details how to reset it and what the consequences are.


brusch:
Thanks!

PR #4657, opened 2019-07-10
Fix w3c validation warning in HeadStyle template helper

knoepfchen23:
The HTML Tags "style" now get rendered without "text/css".


brusch:
Thanks!

PR #4664, opened 2019-07-10
[Core] Fix typo in Bootstrap::bootstrap() method name

podarcis:
BC (tests, existing web.php, ...) is assured by providing a deprecated
version with the old method name.

https://github.com/pimcore/skeleton, https://github.com/pimcore/demo-basic and https://github.com/pimcore/demo-basic-twig should be adjusted as well.


brusch:
Thanks!

PR #4665, opened 2019-07-10
Environment Variable Improvements

### LINKED ISSUE #4642: [Improvement] Update to new .env mechanism
Hello,

According to https://symfony.com/doc/current/configuration.html#managing-multiple-env-files, it could be possible to use `.env.local` or by environment .env files (like .env.stage.local or .env.stage).

Pimcore has not implemented the changes of october 2018. : (
It would be really nice to include them by default. It is one of the main new feature of Pimcore 6 (.env file) and being able to use .local or .env.{env} or .env.{env}.local will bring all the power of this new feature!!
Here is the roadmap to follow, it is just minor changes to Pimcore 6 i think, and it will be really usefull (i think to continuous integration, specific environment, etc.).
https://symfony.com/doc/current/configuration/dot-env-changes.html#updating-my-application

By the way, the default .gitignore should be updated by adding all .local files so they are ignored by GIT:
```
/.env.local
/.env.local.php
/.env.*.local
```
And removing `.env` from .gitignore (should be in the commits).

Thanks!

brusch:
- support for new DotEnv menachnism (resolves #4642) 
- avoid usage of `putenv()` and `getenv()` 
- support for `APP_DEBUG` and `APP_ENV` environment variable
- support for `.env.local.php` file

### DIFF
--- a/bin/pimcore-install
+++ b/bin/pimcore-install
@@ -14,6 +14,7 @@
  */
 
 use Pimcore\Config;
+use Pimcore\Bootstrap;
 use Pimcore\Bundle\InstallBundle\InstallerKernel;
 use Pimcore\Version;
 use Symfony\Bundle\FrameworkBundle\Console\Application;
@@ -41,16 +42,8 @@ if (file_exists($a = getcwd() . '/vendor/autoload.php')) {
     exit(1);
 }
 
-if (!defined('PIMCORE_PROJECT_ROOT')) {
-    define(
-        'PIMCORE_PROJECT_ROOT',
-        getenv('PIMCORE_PROJECT_ROOT')
-            ?: getenv('REDIRECT_PIMCORE_PROJECT_ROOT')
-            ?: realpath($projectRoot)
-    );
-}
-
-\Pimcore\Bootstrap::defineConstants();
+Bootstrap::setProjectRoot();
+Bootstrap::defineConstants();
 
 Debug::enable();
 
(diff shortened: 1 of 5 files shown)


jdreesen:
> support for `.env.local.php` file

How can it be used? 
How to create the file? 
What should it contain?

The [logic used in this PR](https://github.com/pimcore/pimcore/pull/4665/files#diff-0f24a255d82c3b7037c4b0bbb26470a7R171-R174) is really different from [the one in symfony](https://github.com/symfony/dotenv/blob/4.4/Dotenv.php#L126-L167) (which `load()` and `loadEnv()` use internally, too).

brusch:
@jdreesen this is probably because they changed it from `4.0` to `4.4`, see: 
https://github.com/symfony/dotenv/blob/4.0/Dotenv.php

jdreesen:
I see. What's the reason you don't just use `Dotenv::populate()`?

How do I create `.env.local.php` files? Is Symfony Flex usable with Pimcore?

dpfaffenbauer:
the idea is to have `.env` committed. So, if you have like a image you ship, you can pre-populate the cache and put it into a read-only filesystem without running into issues of not having certain env-vars available. (symfony throws an error if you want to use a non-existing env var). 

So, in your .env you basically put in the used vars. 
in .env.local, you put it for your local env. .env.local is not committed.

something like:


```env
//.env

GOOGLE_MAPS_KEY=
MAILCHIMP_KEY=
```

```env
//.env.local

GOOGLE_MAPS_KEY=ASDF
MAILCHIMP_KEY=JKLO
```

dpfaffenbauer:
(that is how symfony suggests it)

jdreesen:
That's right. But Symfony suggests compiling the `.env` vars into a `.env.local.php` file for the `prod` environment: https://github.com/symfony/demo/blob/master/.env#L13

Pimcore seems to support that too (even if the implementation differs) and my question is **how** to compile the `.env` vars in Pimcore. In Symfony it's done with Flex via `composer dump-env prod`, but can Flex be used with Pimcore (as Pimcore still uses the Symfony 3 file structure, etc.)?

dpfaffenbauer:
gotcha, didn't know about that.

Issue #4662, opened 2019-07-10
[Elements] Shortcuts for copying meta info to clipboard


podarcis:
# Feature Proposal / RFC

Provide a shortcut for copying meta info such as ID, full path and deeplink of an element (document, asset, data object) in user system's clipboard.

![image](https://user-images.githubusercontent.com/19773287/60965492-485a5b80-a316-11e9-907b-67f73fec1ee9.png)
![image](https://user-images.githubusercontent.com/19773287/60965504-4f816980-a316-11e9-8ed3-8d8525b0efcf.png)

## Feature description

Especially during development this could come in handy for quickly copying essential information without the need to open the meta info dialog, selecting the desired information and CTRL-c-ing. And the "copy deeplink" shortcut might be useful for the customer as well when working collaborative with others or sharing a link to an element in question.

## RFC

If such a feature is desired and has a chance of being merged I'll finish my prototype (screenshot is Pimcore 5, but will be Pimcore 6) and open a pull request.

ultramegatom:
+1

brusch:
@podarcis Great idea! Feature accepted - looking forward to your PR :)

brusch:
Resolved by #4672

PR #4672, opened 2019-07-11
[UI][Elements] Shortcuts for copying meta info to clipboard

podarcis:
Provide a shortcut for copying meta info (ID, full path and deeplink) of
an element (document, asset, data object) in user system's clipboard.

The metainfo-Button becomes a split-button. The button part still opens
the metainfo-Dialog. The dropdown arrow opens a menu with actions that
copy certain information into the clipboard.

Implements #4662


brusch:
Thanks a lot!

PR #4687, opened 2019-07-14
When sorting data objects in the tree, the new index isn't reflected in event listener

IgorBenko:
When sorting data objects in the tree, the new index isn't reflected in the object instance that is passed into the event listener.

While technically the object sort index hasn't been updated yet, it is still
confusing that the postUpdate listener doesn't get the updated index value.
I'm not sure if there is a reason for updating indexes of siblings after the save,
that is why I didn't put everything in one if statement.

The index in the object that is passed into the postUpdate listener has the correct index.


brusch:
Thanks!

PR #4688, opened 2019-07-14
Set mime-type in BinaryResponse for thumbnails. When loading assets from

IgorBenko:
S3, the MimeTypeGuesser can't handle the file from remote stream.

Specifically the error message is (in FileInfoMimetypeGuesser):
`Warning: finfo::file(): Failed identify data 0:invalid mode 00`

It seems that finfo extension can't handle remote streams. The extension
was installed on the system!

Since we are passing the MimeType with the Response, the MimeGuesser doesn't need to be used and thus finfo will not be used for S3.


brusch:
Thanks @IgorBenko !

PR #4690, opened 2019-07-15
No exception if opened object got deleted in the meantime

BlackbitNeueMedien:
Currently there is an exception thrown if you open an object which got deleted in the meantime.
Steps to reproduce:
1. Open Pimcore backend in 2 browser tabs
1. Browse the tree till you see a certain object in both browser tabs
1. Delete the object in browser tab 1
1. Try to open the object in tab 2


brusch:
Thanks!

PR #4695, opened 2019-07-15
[Perspective] Translation for name in toolbar

blankse:
With this change it is possible to translate the menu items in the perspective toolbar tree.

/app/config/pimcore/perspectives.php
```
return [
    'default' => [
    ],
    'location' => [
    ],
];
```

image


brusch:
Thanks!

PR #4696, opened 2019-07-15
[Assets] Set remote stream

BlackbitNeueMedien:
Currently it is not possible to set a remote stream as the new data source for an asset.
Reproducable with this script:
```php
$asset = \Pimcore\Model\Asset::getById(<ID of an existing Asset>);
$asset->setStream(fopen('https://picsum.photos/200/300', 'rb'));
$asset->save();
```

Results in lots of exceptions because PHP cannot rewind a remote stream (as discussed in #4416 -  the code in this PR is just copied from there).

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1310,15 +1310,12 @@ class Asset extends Element\AbstractElement
             $this->stream = $stream;
 
             $isRewindable = @rewind($this->stream);
-            $tmpFile = PIMCORE_SYSTEM_TEMP_DIRECTORY . '/asset-create-tmp-file-' . uniqid() . '.' . File::getFileExtension($this->getFilename());
-            $dest = fopen($tmpFile, 'w+', false, File::getContext());
-            stream_copy_to_stream($this->stream, $dest);
 
             if (!$isRewindable) {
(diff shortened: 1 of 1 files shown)


models/Asset.php
```diff
@@ -1308,7 +1308,15 @@ public function setStream($stream)
         if (is_resource($stream)) {
             $this->setDataChanged(true);
             $this->stream = $stream;
-            rewind($this->stream);
+
+            $isRewindable = @rewind($this->stream);
```
dpfaffenbauer:
```php
  $meta = stream_get_meta_data($stream);
  $seekable = $meta['seekable'];
```

BlackbitNeueMedien: Hmm, could also be done this way, but would be an additional function call because for local (rewindable) streams, the stream has to to rewinded.

brusch: I'm happy with `@rewind()` if it doesn't have any other implications.

brusch: @BlackbitNeueMedien One more thing, ... the temp file should be cleaned up, once the asset object get's destroyed ... just add the following line after your code and it should be fine :) 

```php
$this->_temporaryFiles[] = $tmpFile;
```

Thanks!

brusch: @BlackbitNeueMedien Maybe also rebase/merge to the current master, so that the tests can complete without any errors :) Thanks again!

PR #4699, opened 2019-07-15
Start session for ecommerce indexer when run in cli - v2

das-peter:
This is a follow up to #4635 
Turns out there are other locations that subsequently trigger the pricing system and hence need a session.
So I've moved the session creation code into its own function and call that function in pretty much every other function in `IndexUpdater` to ensure the session is ready before sending output.
I wonder if as of now it maybe even would make sense to but the session bootstrap into `IndexUpdater::log()` as this is the point at which the output is sent that could interfere with the later start of the session.


brusch:
@das-peter can you please resolve the conflict? Thanks a lot!

bundles/EcommerceFrameworkBundle/IndexService/Tool/IndexUpdater.php
```diff
@@ -161,6 +173,8 @@ public static function processUpdateIndexQueue($tenants = null, $maxRounds = nul
             $tenants = [$tenants];
         }
 
+        self::startSession();
```
NiklasBr: Are duplicate `startSession()`s required?

das-peter: Duplicated starts aren't necessary, but it's necessary to ensure the session is ready to use in every possible entry point which could trigger an output before starting the session.
A duplication of session starts is explicitly covered in the helper method which only starts a session if the code runs in CLI and there no session.

NiklasBr: Oh, I was confused by Github's review interface, didn't see that they where in two different methods. It looked like they where just on different sides of an foreach. Sorry! :D

das-peter: No worries :)

PR #4702, opened 2019-07-16
[Static Routes] Delegate `encode` Param to the Generator

Gerhard-Kanzler:
If you add the param encode in pimcoreUrl() (4. param) now you get the url with all utf8 stuff

### DIFF
--- a/lib/Routing/Staticroute/Router.php
+++ b/lib/Routing/Staticroute/Router.php
@@ -124,7 +124,7 @@ class Router implements RouterInterface, RequestMatcherInterface, VersatileGener
     /**
      * @inheritDoc
      */
-    public function generate($name, $parameters = [], $referenceType = self::ABSOLUTE_PATH)
+    public function generate($name, $parameters = [], $referenceType = self::ABSOLUTE_PATH, $encode = true)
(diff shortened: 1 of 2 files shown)


lib/Routing/Staticroute/Router.php
```diff
@@ -124,7 +124,7 @@ public function getRouteDebugMessage($name, array $parameters = [])
     /**
      * @inheritDoc
      */
-    public function generate($name, $parameters = [], $referenceType = self::ABSOLUTE_PATH)
+    public function generate($name, $parameters = [], $referenceType = self::ABSOLUTE_PATH, $encode = true)
```
brusch: Unfortunately this isn't possible because of `\Symfony\Component\Routing\Generator\UrlGeneratorInterface::generate($name, $parameters = [], $referenceType = self::ABSOLUTE_PATH)` 


So we'd have to stick to the `$parameters` approach, and fix it this way. 

Can you please update your PR? 
Thanks!

Gerhard-Kanzler: I changed the "encode" to $parameters.

PR #4706, opened 2019-07-16
[Customview] Icon not in center and doesn't invert on active

blankse:
## Fixes
- Icon is black on active now
- Icon is in center

## Before
image

## After
image

## Additional
The size of the icons are 24x24 but in this case they should 30x30. I tried to resize with background-size but it doesn't work (Chrome). Maybe it works if we remove the fixed size in the SVGs? Does someone have an idea?


brusch:
Thanks! 

Yep, I guess the SVG has to be 30x30 ... 😐

PR #4712, opened 2019-07-17
[Bug-Fix] Areablock not processing data from rest import properly

mcuenez:
## Description
When importing data from another Pimcore instance using REST, the `getFromWebserviceImport` method in the `AreaBlock.php` file incorrectly parses the `$indices`. This causes an error when trying to open the corresponding `document`, as the database now contains `stdClass` in the serialized data.

## What was changed?
The `$indices` are now properly parsed as seen in the commit. This resolves the issue, ultimately, making the content available as exptected.


mcuenez:
Btw, I also resolved the `todo` regarding the `and``s.

weisswurstkanone:
thx!

PR #4718, opened 2019-07-17
[Classes] Changed protected to public for $fieldDefinitions

vmalyk:
On Bulk export get error http://prntscr.com/ogbifc with unset($fieldCollection->fieldDefinitions) from Pimcore\Model\DataObject\ClassDefinition\Service line 129 and line 166


brusch:
The service should be aware of that, making the property public isn't a proper fix, just a workaround.

vmalyk:
I`ve just updated PR. Changed to methods and set Key and fieldDefinitions to null

brusch:
That looks much better, thanks a lot! 👍

PR #4729, opened 2019-07-18
Install custom layouts

zieglerh:
With the command "pimcore:deployment:classes-rebuild" it's possible to update/install classes/bricks/collections in Pimcore from existing files. The custom layouts were not considered yet.
This PR adds a new command "pimcore:deployment:custom-layouts-rebuild" and works like classes rebuild. 
The new command has been integrated to the Pimcore installer.

## Additional info  
I decided to split and integrate the custom layout into a separate command to not break anything. It could be integrated into the classes-rebuild process as well.


brusch:
Thanks a lot!

PR #4749, opened 2019-07-23
[DataObject] GridConfigDialog: Layout items in tree are not translated

blankse:
It would be good to display the translated name of tab panels.
We use translation identifiers for this panel labels.


brusch:
Thanks!

Issue #2466, opened 2018-01-18
Extension requirements ECommerce framework


andreas-gruenwald:
# Feature Request 

Can we add the following features to the Ecommerce Core:

1.  Add setter for default currency in Environment:

IEnvironment.php
```php
public function setDefaultCurrency(Currency $currency);
```

Environment.php
```
    public function setDefaultCurrency(Currency $currency)
    {
        $this->defaultCurrency = $currency;
    }
```

2. In the AttributePriceSystem.php line 139 in calculateAmount:
 ```
return Decimal::create((double) $product->$getter());
```
in order to avoid exceptions with null values.

fashxp:
can you provide a pull request for that?

fashxp:
ad 1) ... what would be the usecase for that?

ckogler-elements:
> ad 1) ... what would be the usecase for that?

I would also need the "setDefaultCurrency". Usecase:
You have 2 tenants. Tenant A has EUR and tenant B has CZK .

Currently you can't switch the currency per tenant because the currency can only be defined globaly.

andreas-gruenwald:
@fashxp  See PR https://github.com/pimcore/pimcore/pull/4764 for that.
Hope that will help to close the issue.

PR #4764, opened 2019-07-24
Add setter for default currency in ecommerce environment.

andreas-gruenwald:
Add setter for default currency in ecommerce environment.

Resolves: #2466


fashxp:
BC Break only if you don't use Pimcore Environment as base class when implementing your own Environment.

PR #4765, opened 2019-07-24
Ecommerce index bootstrapping for specific tenants.

andreas-gruenwald:
Add an option to the ecommerce index bootstrapping command, which will build/update the index only for specific tenants.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Command/IndexService/BootstrapCommand.php
+++ b/bundles/EcommerceFrameworkBundle/Command/IndexService/BootstrapCommand.php
@@ -35,7 +35,8 @@ class BootstrapCommand extends AbstractIndexServiceCommand
             ->addOption('create-or-update-index-structure', null, InputOption::VALUE_NONE, 'Use to create or update the index structure')
             ->addOption('update-index', null, InputOption::VALUE_NONE, 'Use to rebuild the index data')
             ->addOption('object-list-class', null, InputOption::VALUE_REQUIRED, 'The object list class to use', '\\Pimcore\\Model\\DataObject\\Product\\Listing')
-            ->addOption('list-condition', null, InputOption::VALUE_OPTIONAL, 'An optional condition for object list', '');
+            ->addOption('tenant', null, InputOption::VALUE_REQUIRED | InputOption::VALUE_IS_ARRAY, 'Tenant to perform action on (defaults to all)')
+        ;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


bundles/EcommerceFrameworkBundle/Command/IndexService/BootstrapCommand.php
```diff
@@ -47,6 +48,16 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $listCondition = $input->getOption('list-condition');
+        $tenants = count($input->getOption('tenant')) ? $input->getOption('tenant') : null;
+
+        //set active tenant workers.
+        if (!empty($tenants)) {
+            $tenantWorkerList = [];
+            foreach ($tenants as $tenantName) {
+                $tenantWorkerList[] = Factory::getInstance()->getIndexService()->getTenantWorker($tenantName);
+            }
+            Factory::getInstance()->getIndexService()->setTenantWorkers($tenantWorkerList);
```
fashxp: there is no setter `setTenantWorkers` in indexservice.

I'm also not sure, if we should just add the setter and modify the total tenant list due to potential side effects.
Maybe it would be better to refactor the `IndexUpdater`, remove the static methods, make it a service and add the possibility there to inject a customer index service instead of using the global one.

andreas-gruenwald: Sorry, I forgot to push the part of the IndexService, where the workers are set.

andreas-gruenwald: I just added the setter to be complete.

bundles/EcommerceFrameworkBundle/Command/IndexService/BootstrapCommand.php
```diff
@@ -35,7 +35,8 @@ protected function configure()
             ->addOption('create-or-update-index-structure', null, InputOption::VALUE_NONE, 'Use to create or update the index structure')
             ->addOption('update-index', null, InputOption::VALUE_NONE, 'Use to rebuild the index data')
             ->addOption('object-list-class', null, InputOption::VALUE_REQUIRED, 'The object list class to use', '\\Pimcore\\Model\\DataObject\\Product\\Listing')
-            ->addOption('list-condition', null, InputOption::VALUE_OPTIONAL, 'An optional condition for object list', '');
```
fashxp: sorry, one more thing: why this option was removed? 
I think we should readd it.

andreas-gruenwald: Yes! Done.

PR #4785, opened 2019-07-29
Process queue improvements

jremmurd:
# Process queue improvements

If you have got a lot of products (several hundred thousand) with a complex data structure the preparation may sometimes take quite some time if i.e. a product category with a lot of dependencies gets updated. So if I configured the preparation (`ecommerce:indexservice:process-queue preparation`) to run every 5 minutes but it takes about an hour to finish the job the server load continues to increase as more and more processes are running.
 
This PR now adds several new options to the `ecommerce:indexservice:process-queue` command:
- `ignore-lock` (default true for bc)
- `unlock`
- `lock-timeout`
- `timeout`

The lock-name is a checksum of the `queue` argument and the `tenant ` option.


andreas-gruenwald:
@fashxp We would need this one. May you please give us a quick update whether this is going to be released or not? If not we will find a project-specific solution.

fashxp:
thx!

PR #4787, opened 2019-07-30
Ability to select variants in ManyToOne object grid

### LINKED ISSUE #4750: Not able to select variants in manyToOneRelation
### Expected behavior
In the search grid I need to have a choice between `Object` and `Variant` and `All types`.

### Actual behavior
Only `Object` option is available in the types dropdown.
The issue seems to be related to the `allow Objets` checkbox in the manyToOneRelation settings.
See also `manyToOneRelation.js:openSearchEditor()`.
In pimcore 5 I was able to select variants, but this broke in pimcore 6.

### Steps to reproduce
- Go to classes configuration and open a class.
- Create a new `manyToOneRelation` field
- Check the box `allow Objects` and select a class to be allowed
- Go to an object and click the search icon (loop) in the relation grid.

bramstroker:
Resolves #4750

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyRelation.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyRelation.js
@@ -823,7 +823,7 @@ pimcore.object.tags.advancedManyToManyRelation = Class.create(pimcore.object.tag
             allowedSubtypes.object = [];
             if (this.fieldConfig.classes != null && this.fieldConfig.classes.length > 0) {
                 allowedSpecific.classes = [];
-                allowedSubtypes.object.push("object");
+                allowedSubtypes.object.push("object", "variant");
                 for (i = 0; i < this.fieldConfig.classes.length; i++) {
                     allowedSpecific.classes.push(this.fieldConfig.classes[i].classes);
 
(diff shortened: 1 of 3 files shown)


brusch:
Please let us discuss this internally first, if it fit's into the general concept of variants and if this could have any side effects. 

If we decide to go in this direction and allow relations to variants, this PR should also cover all other relational data-types, such as (Advanced) Many-To-Many (Object) Relations ...

bramstroker:
Hi @brusch. Thanks for your reply. We use relations to variants a lot and our whole application / domain is dependant on this. In Pimcore 5 we did not experienced any issues and it was not a limitation.
Within Pimcore 6 our content editors are not able to select variants anymore because they don't have this option.

Btw. When you don't configure a specific object class on the relation it is possible to select "variant" as a type.

bramstroker:
I agree we must also look at the other relation types, but I did have a look at this already when I made this PR and it does not seem and issue with the other relation types. I am able to select "variant" there without any problem.

brusch:
Must have been a regression of  #4496, but for me it seems this never worked properly :|

bramstroker:
Yes I agree, does not look like a regression of this particular PR when I have a quick look at the code. Because for `allowedClasses` configuration option only `object` subtype was available.
I suspect this never worked properly since the addition of the `type` select in Pimcore 6

bramstroker:
@brusch Do you need anything from me for this PR to be merged, or are you going to discuss this internally further?

brusch:
@bramstroker we have the same issue for the (advanced) many-to-many relations, right? 
See: https://github.com/pimcore/pimcore/blob/9b2ab13224a6f89a70415daca10f51d0665a6d4a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyRelation.js#L487 
and 
https://github.com/pimcore/pimcore/blob/9b2ab13224a6f89a70415daca10f51d0665a6d4a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyRelation.js#L826

I'd say we should fix it there as well, right?

brusch:
@bramstroker can you update the PR accordingly? Would be great, thanks a lot! :)

bramstroker:
@brusch I have updated PR accordingly. Also tested on my local instance. Everything is working for me as expected.

brusch:
Merci beaucoup!

PR #4794, opened 2019-07-31
[Documents] Newsletter: add option to maintain a manual edited plain-text version

Lexipowder:
Spam filters like SpamAssassin rate multipart emails higher than just html-email. Having the text part is worth one score point. The existing version just sends html-emails, also with the detour via html2text. (See comment in code).

Following database change has to be done first: (I don't know how to send the db-update in a PR)
ALTER TABLE `intranet`.`documents_newsletter` 
ADD COLUMN `plaintext` LONGTEXT NULL DEFAULT NULL AFTER `sendingMode`;

Best regards
Lexi

### DIFF
--- a/bundles/CoreBundle/Resources/translations/de.json
+++ b/bundles/CoreBundle/Resources/translations/de.json
@@ -52,8 +52,6 @@
     "send_newsletter": "Newsletter JETZT versenden",
     "object_filter": "Objekt Filter",
     "add_newsletter": "Newsletter hinzufügen",
-    "plain_text": "Klartext",
-    "plain_text_email_part": "E-Mail-Klartextteil",
     "newsletter": "Newsletter",
     "crm": "CRM",
     "notes_events": "Notizen und Termine",
--- a/bundles/CoreBundle/Resources/translations/de.json
+++ b/bundles/CoreBundle/Resources/translations/de.json
@@ -52,6 +52,8 @@
     "send_newsletter": "Newsletter JETZT versenden",
     "object_filter": "Objekt Filter",
     "add_newsletter": "Newsletter hinzufügen",
+    "plain_text": "Klartext",
+    "plain_text_email_part": "E-Mail-Klartextteil",
     "newsletter": "Newsletter",
     "crm": "CRM",
     "notes_events": "Notizen und Termine",
--- a/lib/Tool/Newsletter.php
+++ b/lib/Tool/Newsletter.php
@@ -69,6 +69,10 @@ class Newsletter
         if ($sendingContainer && $sendingContainer->getParams()) {
             $mail->setParams($sendingContainer->getParams());
         }
+        
+        if(strlen(trim($newsletterDocument->getPlaintext())) > 0) {
+            $mail->setBodyText(trim($newsletterDocument->getPlaintext()));
+        }
 
         $contentHTML = $mail->getBodyHtmlRendered();
         $contentText = $mail->getBodyTextRendered();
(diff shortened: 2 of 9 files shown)


lib/Tool/Newsletter.php
```diff
@@ -115,6 +119,8 @@ public static function prepareMail(
 
         $mail->setBodyHtml($contentHTML);
         $mail->setBodyText($contentText);
+        // Adds the plain text part to the message, that it becomes a multipart email
+        $mail->addPart($contentText, 'text/plain');
```
brusch: Why? See: 
https://github.com/pimcore/pimcore/blob/d688a67780b6ae733e145110bbbc04b3278e7190/lib/Mail.php#L567

bundles/CoreBundle/Resources/translations/de.json
```diff
@@ -52,6 +52,8 @@
     "send_newsletter": "Newsletter JETZT versenden",
     "object_filter": "Objekt Filter",
     "add_newsletter": "Newsletter hinzufügen",
+    "plain_text": "Klartext",
```
brusch: Languages other than EN are managed here: 
Please revert this change and add the translation in POEditor once this PR is merged.

lib/Tool/Newsletter.php
```diff
@@ -137,8 +143,6 @@ public static function sendNewsletterDocumentBasedMail(Mail $mail, SendingParamC
 
         if (!empty($mailAddress)) {
             $mail->setTo($mailAddress);
-            // Getting bounces
```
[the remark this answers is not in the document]
Lexipowder: Not necessary, as the SMTP Server will set it automatically set the return path when finally sending the email.

https://www.postmastery.com/about-the-return-path-header/

bundles/CoreBundle/Resources/translations/de.json
Lexipowder: Where exactly? Link missing...

lib/Tool/Newsletter.php
Lexipowder: Line 567 is part of the send function, the newsletter is send via sendWithoutRendering function:

Pimcore::getEventDispatcher()->dispatch(DocumentEvents::NEWSLETTER_PRE_SEND, $event);
            $mail->sendWithoutRendering($mailer);
            Pimcore::getEventDispatcher()->dispatch(DocumentEvents::NEWSLETTER_POST_SEND, $event);

bundles/CoreBundle/Resources/translations/de.json
brusch: https://poeditor.com/join/project/VWmZyvFVMH

image
image

lib/Tool/Newsletter.php
brusch: Good point 👍😊

lib/Tool/Newsletter.php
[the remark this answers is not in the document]
brusch: You added this in #4349 ... can't hurt to leave it as it is, right?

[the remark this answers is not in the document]
Lexipowder: I know it was me. :-) 
But on what I read in the meantime, I would recommend to leave it out, because some mailserver seems to reject or identify the email as spam, when return path is wrongly set. At the end if we leave it in, it won't hurt the sending process, !maybe! some emails won't receive the customer.

bundles/CoreBundle/Resources/translations/de.json
Lexipowder: Thank you, will do

PR #4799, opened 2019-08-02
Remove unused variables in document editmode dnd

GALCF:

Removed unused variables `iFrameElement` and `xy`

* `iframeElement` (spelled differently) is passed through initialize parameters
* `xy` is never used in the function scope


brusch:
Thanks!

Issue #4790, opened 2019-07-30
[Bug] Custom admin entrypoint don't work when path begining with "admin..." and doc not up to date


Yivan:
Hello,

We were trying implementing custom admin path.
We get two problems.

1- The easy one, is the documentation is not correct (invalid Yaml).
Here https://pimcore.com/docs/6.x/Development_Documentation/Development_Tools_and_Details/Custom_Admin_Login_Entry_Point.html, good code is:
```
my_custom_admin_entry_point:
    path: /my-custom-login-page
    controller: Pimcore\Bundle\CoreBundle\Controller\PublicServicesController::customAdminEntryPointAction
```

2- And we where strugling trying to make work path with `/admin-mytest` and it didn't work (don't go in the customAdminEntryPointAction and eventlistener (https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/EventListener/CustomAdminEntryPointCheckListener.php#L57) is runned before going into the routed action.
Same with `/adminmytest`.
But path with `/roberto` work well.
So it seems all path begining with `/admin...` don't work as custom admin entry point.
I think it should corrected because many people will do somethings custom but meaningfull so possibility of path starting with admin word is not negligible.
Globally it means we cannot do any routes in yaml beginning with admin even in the frontend... a more harder bug.

Thanks.

brusch:
1. changed that - probably a Symfony 3.4 vs 4.3 issue 
2. not possible to fix, added a not in the docs

Yivan:
@brush Thanks for your return / fix.

I think it need some code fixing at this level:
https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/EventListener/CustomAdminEntryPointCheckListener.php#L57

Because any custom route beginning with "/admin*" will fall in this exception. So imagine a front-end custom rule for acccessing a custom admin having url beginning by /admin*. It won't work.
Maybe at this level, it should only send the exception if we are trying to access to the admin BO Pimcore related bundle ?
And it will allow in the same way to use /admin* even for BO Pimcore custom route.

The real problem is this exeption is throw for any bundles... If it is not throw the route is well resolved after.

Thanks!

brusch:
As I said, ... I don't think we can fix that, because the context resolver uses /admin/* . However, if you think you can fix that without breaking anything else, we're happy to mere a PR.

dpfaffenbauer:
Container Parameters could work...

Yivan:
[...]

PR #4802, opened 2019-08-02
Fix paths in yml files for more robust context and firewall

Yivan:
Hello,

This PR make context matching more robust so it does not include URLs like:
```
/admin-project
/webservice-project
/plugin-project
...
```

Actually they was resolved as admin/webservice/plugin context.
Firewall paths have been changed too, so firewall is not activated on these custom URLs.

Now, something like:
```
project_myfront_entry_point:
    path: /admin-front
    controller: AppBundle\Controller\DefaultController::testFrontAdminAction
```
will work ! : )

And custom admin pimcore entrypoint can now begin with /admin* too.
Doc for admin entrypoint is updated in this PR too.
https://github.com/pimcore/pimcore/issues/4790 is resolved with this PR.

Tested locally, all seems to works pretty fine.
Thanks.

### DIFF
--- a/bundles/CoreBundle/Resources/config/pimcore/default.yml
+++ b/bundles/CoreBundle/Resources/config/pimcore/default.yml
@@ -202,18 +202,18 @@ pimcore:
     context:
         profiler:
             routes:
-                - { path: ^/_(profiler|wdt) }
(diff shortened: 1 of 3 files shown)


brusch:
Looks good, thanks a lot!

PR #4803, opened 2019-08-05
add possibility to change mode of wysiwyg editor 

knoepfchen23:
Added the possibility to configure the text editor to open in classic mode. The inline Mode is kept as default.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/tags/wysiwyg.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/tags/wysiwyg.js
@@ -94,6 +94,12 @@ pimcore.document.tags.wysiwyg = Class.create(pimcore.document.tag, {
 
             delete specificConfig.width;
 
+            var editorMode = "inline";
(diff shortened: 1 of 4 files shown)


brusch:
What is the use case for that? 
Does drag & drop of links work into the classic mode? 
To be honest, if possible I'd like to avoid maintaining 2 different modes of the wysiwyg :)

knoepfchen23:
Unfortunately drag & drop does not work in the classic mode. The problem of some of our clients is, that the inline-mode overlay of the wysiwyg editor hides other bricks or their edit toolbar, so you have to click somewhere to deactivate the overlay. Yes, we know that it’s only one click beside the editor, therefore we thought about a configuration possibility to switch to the classic mode. If there is another possibility to improve the usability of this matter, it would be ok.

brusch:
Dnd is a critical functionality for the WYSIWYG, so I'm going to close this PR for now. 
Feel free to create an issue regarding the toolbar issue, or even better provide a working fix in a PR :) 
Thanks!

PR #4806, opened 2019-08-06
[docs] add best practice to open an object by an external id

ultramegatom:


brusch:
Maybe we should also reference [this](https://pimcore.com/docs/6.x/Development_Documentation/Extending_Pimcore/Event_API_and_Event_Manager.html#page_Hook-into-the-Open-Document-Asset-Data-Object-dialog) alternative approach, just for the sake of completeness?

ultramegatom:
Oh i didnt actually know this exists. I mentioned it :)

PR #4808, opened 2019-08-06
[Ecommerce][CartManager] Calculate GiftItems of Cart only if not already calculated

AlternateIf:
This PR fixes calculating a already calculated cart.

Whenever getGiftItems or getGiftItem is called the calculate method is executed. This also triggers the applyCartRules function. However, this should not be the case if the cart was calculated before.

@fashxp


fashxp:
thx

PR #4809, opened 2019-08-06
Transmit values of system columns to grid operators

BlackbitNeueMedien:
Currently there is an error if you add a date formatter operator to a grid config and assign this date formatter operator the modificationDate or creationDate system columns. 
Bildschirmfoto 2019-08-06 um 13 00 35
(This screenshot is done including this PR's changes)

This results in the error
´´´
Call to a member function isEmpty() on null
Trace: 
#0 /path/to/pimcore/vendor/pimcore/pimcore/lib/DataObject/GridColumnConfig/Value/DefaultValue.php
...
´´´

This PR fixes this problem.


brusch:
Thanks!

PR #3943, opened 2019-02-15
Improve handling of Meta Data from Assets

PGBrandi:
- Analyse videos with EXIFTool
- Combined EXIF/IPTC/XMP Data
- Extended Meta Data for videos and images

PR #4813, opened 2019-08-06
Improve handling of Meta Data from Assets

mysliwietzflorian:

Add is_file-check for getXMPData.

For IPTC and Exif metadata a filecheck is performed. This was overlooked for XMP data.

```php
$xmp = $this->flattenArray($this->getXMPData($filePath));
$iptc = $this->flattenArray($this->getIPTCData($filePath));
$exif = $this->flattenArray($this->getEXIFData($filePath));
```
See: https://github.com/pimcore/pimcore/blob/master/models/Asset/MetaData/EmbeddedMetaDataTrait.php#L140

Follow up of #3943 and #3979.


brusch:
Thanks!

PR #4816, opened 2019-08-06
[Ecommerce] Checkout refactoring

fashxp:
### Basic goals of new architecture:
* Make payment integration easier and more transparent.
* Tackle the 'cart is readonly deadlock' problem.
* Introduce events for easier customization of the checkout process.

### Changes in a nutshell: 
* Cart can be configured that it is not readonly any more as soon as an active payment exists. 
* Checkout manager can be configured with different strategies for handling active payments and restart of payments like
  * **RecreateOrder**: Create new order every time a payment is started and leave old orders untouched. 
  * **CancelPaymentOrRecreateOrder** (default value): Cancel payments if possible and cart has not changed, create new order when cart has changed.
  * **ThrowException**: Throw exceptions to make handling of these cases in controller possible.
* Payment provider have now typed responses (e.g. form, url, json, etc.) to make integration into controllers/templates more transparent


yariksheptykin:
@fashxp this is a great project! Since we have a chance now to rewrite parts of e-commerce do you think adding some test coverage on unit level would be reasonable? New [CheckoutManager](bundles/EcommerceFrameworkBundle/CheckoutManager/V7/CheckoutManager.php) seems to get all its dependencies injected so writing a unit test should not be that difficult. What do you think?

PR #4819, opened 2019-08-07
[Documents] Navigation - Added option to provide class for sub menus

### LINKED ISSUE #4741: [Documents] Navigation - option to provide class for sub menus
# Feature Request
While using Navigation helper (https://pimcore.com/docs/5.x/Development_Documentation/Documents/Navigation.html) it is possible to provide class(ulClass) for main menu but there is no option to provide class for sub-menus or ulClass property should be applied on levels.

dvesh3:
Changes in this pull request  
Resolves #4741

## Additional info  
```php
    <?php
    echo $this->navigation()->menu()->renderMenu($mainNavigation, [
        'maxDepth' => 2,
        // apply different ulClass on depth levels 
        'ulClass'  => [
            0 => 'nav navbar-nav', //ulClass for first level
            1 => 'nav navbar-nav-second',
            2 => 'nav navbar-nav-third'
        ],
        //or apply same ulClass on depth levels
       'ulClass'  => [
          'default' => 'nav navbar-nav'
       ]
    ]);
    ?>
```

### DIFF
--- a/lib/Navigation/Renderer/Menu.php
+++ b/lib/Navigation/Renderer/Menu.php
@@ -761,7 +761,7 @@ class Menu extends AbstractRenderer
 
             //set ulCLass depth wise
             if(is_array($ulClasses)) {
-                $ulClass = ($ulClasses[$depth] ? $ulClasses[$depth] : $ulClasses['default']);
+                $ulClass = $ulClasses[$depth] ?? $ulClasses['default'];
             }
 
             if ($depth < $minDepth || !$this->accept($page)) {
--- a/lib/Navigation/Renderer/Menu.php
+++ b/lib/Navigation/Renderer/Menu.php
@@ -516,7 +516,7 @@ class Menu extends AbstractRenderer
 
         // UL class
         if (isset($options['ulClass']) && $options['ulClass'] !== null) {
-            $options['ulClass'] = (string) $options['ulClass'];
+            $options['ulClass'] = $options['ulClass'];
         } else {
             $options['ulClass'] = $this->getUlClass();
         }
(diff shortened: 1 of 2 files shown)


lib/Navigation/Renderer/Menu.php
```diff
@@ -756,6 +757,13 @@ protected function _renderMenu(
         foreach ($iterator as $page) {
             $depth = $iterator->getDepth();
             $isActive = $page->isActive(true);
+
+
+            //set ulCLass depth wise
+            if(is_array($ulClasses)) {
+                $ulClass = ($ulClasses[$depth] ? $ulClasses[$depth] : $ulClasses['default']);
```
brusch: @dvesh3 I think we should better use the cull-coalescing-operator `??` here, because it doesn't issue a warning if the array index doesn't exist (and it's shorter ;)). 
`$ulClass = $ulClasses[$depth] ?? $ulClasses['default']`

brusch: @dvesh3 updated it already, nothing else to do for you ;)

dvesh3: you're a lifesaver 😃

lib/Navigation/Renderer/Menu.php
```diff
@@ -697,7 +697,7 @@ protected function _renderDeepestMenu(
      * Renders a normal menu (called from {@link renderMenu()})
      *
      * @param  Container $container     container to render
-     * @param  string                    $ulClass       CSS class for first UL
+     * @param  string                    $ulClasses       CSS class for UL levels
```
yariksheptykin: `$ulClasses` can now be both `string` and `array`. Annotation `@param  string|string[]` would be more appropriate.

PR #4827, opened 2019-08-08
[Ecommerce] Consider json_decode errors on updating index

jremmurd:
## Improvement

I had the case that there was invalid json data in my index store table. As the json_decode failed the data was not added to the index itself and there always was some inconsistency between store and index - without any errors shown. This PR now throws an error if the json_decode fails.


fashxp:
thx

PR #4834, opened 2019-08-12
Allow _ in custom object layout names

markus-moser:
In previous versions there was no regex name check therfore some legacy names do not work anymore. I suggest to allow _ here as it is not a dangerous character.


brusch:
Thanks!

PR #4835, opened 2019-08-12
Update Version marshal to ignore Doctrine

dpfaffenbauer:
Otherwise, it would also copy entity-manager and that is not possible, this fixes CoreShop Builds: https://travis-ci.com/coreshop/CoreShop/jobs/224293440#L991


dpfaffenbauer:
Also, I guess it be good to have an extension point here to manually marshal and umarshal custom Data Types

PR #4837, opened 2019-08-12
Fix/undefined value2 index

davi-mo:
Resolves #

## Additional info  
When you are saving an item at the classification store, you can receive an undefined index value2.


weisswurstkanone:
thx!

PR #4842, opened 2019-08-13
[Session] Use PHP's default session mechanism

brusch:
This is just what the user expects ...


doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,5 +1,17 @@
 # Upgrade Notes
 
+## 6.1.2
```
NiklasBr: BC breaking change?

brusch: Don't think so ... nothing changes if you're using a custom session config. 
If you're using the default, it will continue to work by using PHP's default config.

brusch: I just thought it is worth to mention in the upgrade notes.

PR #4852, opened 2019-08-16
[Workflows] Allow non-integer object layout IDs

kjkooistra-youwe:
Object/custom layouts can have a string as ID. However, when configuring those for a workflow the system will throw an exception because of the configuration validation.


brusch:
Grazie Mille! 👌

PR #4860, opened 2019-08-19
[Document] Support for Folders & Links as a Fallback Document

brusch:
Currently only the types `page` `snippet` and `hardlink` are supported, but it would make sense to support `link` and `folder` as well. 

### Example: 

Folder structure [type in square brackets]: 
```
/ [page]
/en [link]
```

Route: `/en/news/Title~n123`

In the action of the route, the fallback document in `$this->document` is currently `/`, this PR sets the document to  the link `/en`.


solverat:
@brusch, there was a discussion the other day but I guess I'm too late anyway. maff ignored those types because of BC reasons.

https://github.com/pimcore/pimcore/issues/1733#issuecomment-362267021

Issue #4758, opened 2019-07-24
[Data Object] Decimal Value Data Loss in German Language


AroraVaishali:
# Bug Report

This issue produce in https://demo-basic.pimcore.org/admin/
### Expected behavior
NO data loss
### Actual behavior
decimal values convert into integer in next tab
### Steps to reproduce
1) Created Field collection with 2 tabs
2) Create a class and add that field collection
3)Each tab has 2 numneric fields (PFA fieldCollectionClass)
4) Create object by Admin user. Enter decimal values in each fields.
5) Login to German user (PFA : TestGermanUser). 
Note : **1.23 is display 1,23 in german**
6) **Change value in Tab1 and click on save button.**
7) Refresh page and check values in Tab2 . decimal data is convert into integer. 
 PFA 

![fieldCollectionClass](https://user-images.githubusercontent.com/19402815/61769775-0991ce80-ae09-11e9-8449-788235113fce.png)
![TestGermanUser](https://user-images.githubusercontent.com/19402815/61769773-0991ce80-ae09-11e9-9681-a8de6099427d.png)
![Befordecimalvalue](https://user-images.githubusercontent.com/19402815/61769809-2e864180-ae09-11e9-9ca0-ab7ddad15abb.png)
![AfterDecimalValue](https://user-images.githubusercontent.com/19402815/61769771-0991ce80-ae09-11e9-9e06-8a3f4efb6aa2.png)

### Code example
```php
// your code
```

# Feature Request

satisgupta:
@brusch Could you please share any update on this?

dvesh3:
@brusch @AroraVaishali @satisgupta issue is only reproducible when user(with language having comma as decimal seperator) changes the value in first tab and don't open another tabs.

satisgupta:
@dvesh3 It is reproducible if user changes value in any tab, decimal value get removed for the fields in the next tab(s).

satisgupta:
@brusch Thanks for fixing this issue. Can you please let me know, how can I apply this fix in previous version. I need to apply in version 5.4.4?

brusch:
See changes of the commit ... 257297cc2c6f841926a5219698da6edadfdccb7a
patch is available here https://github.com/pimcore/pimcore/commit/257297cc2c6f841926a5219698da6edadfdccb7a.patch

PR #4862, opened 2019-08-19
fix: taking into account comma on decimal validation

### LINKED ISSUE #4861: Decimal validation in error when comma is ussed
When the comma is used instead of the dot for the decimals (in french for example), the validation is not functional

### Expected behavior
Validation OK

### Actual behavior
image

### Steps to reproduce

Put the admin interface in french and set a decimal to 0,03 on a field that has a minimum value of 0,02 then publish.

rbonfils:
Resolves #4861

## Additional info  
Also checks for a comma before returning an integer


rbonfils:
Sorry, I did not see that the bug had already been fixed here: https://github.com/pimcore/pimcore/issues/4758

PR #4866, opened 2019-08-19
[Workflows] Layout not reloaded after transition

### LINKED ISSUE #4735: [Workflows] Custom DataObject layouts not transitioning with the current state
# Bug Report

### Expected behavior
The correct custom layout should be shown when transitioning a step in the workflow, as defined in the objectLayout configuration.

### Actual behavior
The same custom layout is always used.

### Steps to reproduce
- Create a DataObject class with two text fields
- Create two custom layouts with noticable differences, for example make one of the fields uneditable
- Configure a workflow with two places and configure the custom layouts:
```yml
pimcore:
    workflows:
        test_workflow:
            enabled: true
            priority: 0
            type: 'workflow'
            supports:
                - [your data object class]
            initial_place: place1
            places:
                place1:
                    label: Step 1
                    permissions:
                        - objectLayout: 1
                place2:
                    label: Step 2
                    permissions:
                        - objectLayout: 2
            transitions:
                step1_proceed:
                    from: place1
                    to: place2
                step2_return:
                    from: place2
                    to: place1
```
- Try creating an object instance and switch between the workflow steps. It will always show the same custom layout.

### Code example

Open 'pimcore/pimcore/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js':
In the reload method (line 808), the following must be uncommented to produce the expected behaviour.
```js
if(!params['layoutId']) {
    params['layoutId'] = this.data.currentLayoutId;
}
```
These lines might serve another purpose, so I am hesitating to just remove these in a PR.

A workaround I have used is to copy the code from the onAdminElementGetPreSendData event in 'pimcore/pimcore/bundles/CoreBundle/EventListener/WorkflowManagementListener'
and remove the $hasSelectedCustomLayout check, thus executing the part in the if condition below. That if condition will set the correct layout according to the current workflow step.

kjkooistra-youwe:
Updated object reload JavaScript function to allow resetting the layout ID,
which ensures the layout is reloaded as well.

Resolves #4735


fashxp:
thx

PR #4877, opened 2019-08-21
override default dump_directory from public to web

MercuryKojo:
Presta Sitemap now sets dump_directory to public by default for Symfony Flex.
Since Pimcore does not use Symfony Flex yet, this should be set by default.


fashxp:
thx

PR #4883, opened 2019-08-25
[doc] Rename upstream name to php-pimcore6

ceefour:
Rename upstream name to php-pimcore6


fashxp:
thc

PR #4886, opened 2019-08-26
Remove "prod" env from clear cache list

solverat:
Do not add (by guessing) "prod" to `env` list to clear cache after saving system settings: 
- Most of the time, it's not required (long build time)
- Others may use `production`, `staging` instead of `prod`, `stag` as env names.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/SettingsController.php
+++ b/bundles/AdminBundle/Controller/Admin/SettingsController.php
@@ -594,7 +594,7 @@ class SettingsController extends AdminController
         $this->forward(self::class . '::clearCacheAction', [
             'only_symfony_cache' => false,
             'only_pimcore_cache' => false,
-            'env' => array_unique(['dev', 'prod', \Pimcore::getKernel()->getEnvironment()])
+            'env' => array_unique(['dev', \Pimcore::getKernel()->getEnvironment()])
         ]);
 
         return $this->adminJson(['success' => true]);


solverat:
and i forgot the main reason: if `prod` is not defined, a exception will be thrown: 

`File [..]/config_prod.yml  cannot be found`

brusch:
I don't think just removing `prod` is the solution for this ... in probably 90% there's a `prod` environment, so that should be definitely covered.

solverat:
@brusch, I don't get it. I event wanted to update this pr to also remove the "dev" env. There is already the `\Pimcore::getKernel()->getEnvironment()` method, which returns the right environment, so why adding those two? 

Second: "probably 90%" - and that's why we should remove it, right? Besides that it works perfectly fine with the `getEnvironment()`itself.

Third: If you keep it that way, pimcore requires two environments `dev` and `prod` which is not documented - and in my opinion, absolutely unnecessary...

dpfaffenbauer:
I don't get it as well... Why force to clear all envs on saving system settings? That can't work right for all cases. What if I have more envs than 'prod' and 'dev'? or different envs? Why not just clear the current env? and let the dev clear the other envs manually? Is there maybe a way to get all defined envs and clear them?

An easy fix would be to introduce a new configuration, called something like `envs_to_clear_on_system_settings_save` and let it have default values for `prod` and `dev`. So if you define or rename your envs, you can change the config.

solverat:
@brusch, we really need to talk this out. Why should we ever rebuild and create a `dev` cache block in any other environment than `dev`? Why adding those two absolut unecessary variables if there is already `\Pimcore::getKernel()->getEnvironment()` which returns the only valid and current environment?

PR #4887, opened 2019-08-26
[FieldCollection] Json import ignores title and group

blankse:


brusch:
Thanks a lot!

models/DataObject/ClassDefinition/Service.php
```diff
@@ -149,14 +149,19 @@ public static function importFieldCollectionFromJson($fieldCollection, $json, $t
             $fieldCollection->setLayoutDefinitions($layout);
         }
 
-        $fieldCollection->setParentClass($importData['parentClass']);
+        foreach (['parentClass', 'title', 'group'] as $importPropertyName) {
+            if (isset($importData[$importPropertyName])) {
+                $fieldCollection->{'set' . $importPropertyName}($importData[$importPropertyName]);
```
yariksheptykin: @blankse Isn't `ucfirst` missing here?
i. e:
```
$fieldCollection->{'set' . ucfirst($importPropertyName)}($importData[$importPropertyName]);
```
It used to be `$fieldCollection->setParentClass`. Now it's become `$fieldCollection->setparentClass`. I'd expect a "missing method" error.

blankse: @yariksheptykin Yes ucfirst is missing. Does it work in class import? There is the same:
https://github.com/pimcore/pimcore/blob/master/models/DataObject/ClassDefinition/Service.php#L112

yariksheptykin: This is a good question. To my understanding that should not work either. Is there any magic that could convert `setparentClass` to `setParentClass` in AbstractModel / AbstractDao, @brusch ?

blankse: @yariksheptykin I checked it. Function names are case-insensitive in PHP.
See:
- https://www.php.net/manual/en/functions.user-defined.php
- https://stackoverflow.com/questions/33273941/php-case-sensitivity

I think we should nevertheless use the correct spelling

brusch: @blankse @yariksheptykin added `ucfirst()` thanks for reviewing!

PR #4890, opened 2019-08-26
[Assets] hasMetaData should be false if setMetaData(null) gets called

BlackbitNeueMedien:
Currently if an asset had metadata and `$asset->setMetadata(null)` or `$asset->setMetadata([])`gets called, the property `hasMetaData` stays true.


brusch:
Thanks a lot!

PR #4892, opened 2019-08-27
Hotspotimage does not use crop data when medias are available

### LINKED ISSUE #4891: Hotspotimage does not use crop data when medias are availabel
# Bug Report
When cropping an image in the Pimcore Backend, using a `Hotspotimage` with `media` in the thumbnail config, the image is not being cropped.

### Expected behavior
The image should be cropped.

### Actual behavior
The image is not cropped.

### Steps to reproduce
1. Create a thumbnail with a media query.
2. Render a `Hotspotimage`, using the thumbnail.
3. Compare image on desktop with same image on mobile.

mcuenez:
Fixes #4891

### DIFF
--- a/models/DataObject/Data/Hotspotimage.php
+++ b/models/DataObject/Data/Hotspotimage.php
@@ -181,6 +181,19 @@ class Hotspotimage implements OwnerAwareFieldInterface
         }
 
         if ($crop) {
+            if ($thumbConfig->hasMedias()) {
+                $medias = $thumbConfig->getMedias() ?: [];
+
+                foreach ($medias as $mediaName => $mediaConfig) {
+                    $thumbConfig->addItemAt(0, 'cropPercent', [
+                        'width' => $crop['cropWidth'],
+                        'height' => $crop['cropHeight'],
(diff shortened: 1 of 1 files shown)


models/DataObject/Data/Hotspotimage.php
```diff
@@ -181,6 +181,19 @@ public function getThumbnail($thumbnailName = null, $deferred = true)
         }
 
         if ($crop) {
+            if ($thumbConfig->hasMedias()) {
+                $medias = $thumbConfig->getMedias() ?: [];
+
+                foreach ($medias as $mediaName => $mediaConfig) {
+                    $thumbConfig->addItemAt(0, 'cropPercent', [
```
brusch: Shouldn't be this `$mediaConfig` instead of ` $thumbConfig`?

ultramegatom: Should be fine ```addItemAt``` handles this correctly with the last param

brusch: Ah, thanks, have overssen the last param, thanks for clarifying!

Issue #4810, opened 2019-08-06
Prevent data loss when two users save same object


BlackbitNeueMedien:
Currently Pimcore is prone to data loss when two users open the same object.
Steps to reproduce:
1. User A opens object 1
  * at this time data of object 1 gets loaded
1. User B opens object 1 (a little message appears that this object is already being opened by another user - this is currently the only protection against data loss through multiple user savings)
  * User A and B see the same data
1. User A changes data of object 1 and saves object
1. User B changes data of object 1 and saves object
1. Object 1 now has only the data of user B. All changes of user A have been overwritten due to user B's saving.

Proposal:
* Remember loaded version ID which underlies the loaded data of the opened object
* When object is getting saved, check if loaded version ID is identical with current version:
  * if true: save object
  * if not:
    1. Load current version's data
    1. Check if there are changes done in the same fields in the meantime as of the ones being saved now (and field's contents between current version and saved data differ)
       * if yes: error message because Pimcore cannot decide which data in a certain field is wanted; 
       * if not: save only fields which got changed and keep all other data from the current version

brusch:
Actually I think this would add a lot of extra complexity to the entire process which is currently already quite complex. Just think of nested data-types such as field-collections, ... that would be just a freaking nightmare. In my opinion it's just not worth it, even though I understand the problem. 

I'll close this issue for now, since that's not gonna be implemented by the core-team, but feel free the continue the discussion 😊

BlackbitNeueMedien:
I could imagine remembering the loaded version id and then remember all setter calls.
When save() gets called, the underlying version ID gets compared to the current version ID of the object. If they differ, the current version gets loaded and all setter calls get called again.
Am I underestimating this task?

brusch:
> Am I underestimating this task?
I'd think so ... 😉

But feel free to create a PR for that, just remember that we won't accept an extra layer of complexity.

BlackbitNeueMedien:
@brusch what do you think of showing the "another user has already opened this object" only to users who have the right to edit the object? Because if the current user only has the right to read the object, the info message does not make much sense, does it?

PR #4896, opened 2019-08-28
[Data Objects] Show editlock dialog only if user has permissions to edit object

BlackbitNeueMedien:
When two users open the same data object in Pimcore backend there is the message "The desired element is currently opened by another person" with the option to open it anyway. This dialogue's goal is to prevent the object from being edited by multiple users in parallel because this would overwrite the changes of each other. As it is currently too complex to implement a mechanism to allow parallel editing (#4810) this PR shows the edit-lock dialogue only to those users who have the right to edit the opened object. If a user does not have the right to open the object the dialogue does not make much sense because he cannot alter the object anyway.

Use case: Multiple support employees wiht only read permissions use Pimcore for researching data. When they open the same object the edit-lock dialogue appears - this confuses the user (and is not needed imho).

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -376,7 +376,7 @@ class DataObjectController extends ElementControllerBase implements EventedContr
 
         // check for lock
         $allowedToModify = $object->isAllowed('save') || $object->isAllowed('publish') || $object->isAllowed('unpublish') || $object->isAllowed('delete');
-        if (Element\Editlock::isLocked($request->get('id'), 'object') && $allowedToModify) {
+        if ($allowedToModify && Element\Editlock::isLocked($request->get('id'), 'object')) {
             return $this->getEditLockResponse($request->get('id'), 'object');
         }
 
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -365,13 +365,7 @@ class DataObjectController extends ElementControllerBase implements EventedContr
      */
     public function getAction(Request $request, EventDispatcherInterface $eventDispatcher)
     {
-        // check for lock
-        if (Element\Editlock::isLocked($request->get('id'), 'object')) {
-            return $this->getEditLockResponse($request->get('id'), 'object');
-        }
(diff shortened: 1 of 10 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -365,23 +365,24 @@ public function getIdPathPagingInfoAction(Request $request)
...
+        // check for lock
+        if (Element\Editlock::isLocked($request->get('id'), 'object') && ($object->isAllowed('save') || $object->isAllowed('publish') || $object->isAllowed('unpublish') || $object->isAllowed('delete'))) {
+            return $this->getEditLockResponse($request->get('id'), 'object');
+        }
+        Element\Editlock::lock($request->get('id'), 'object');
 
         // we need to know if the latest version is published or not (a version), because of lazy loaded fields in $this->getDataForObject()
-        $objectFromVersion = $latestObject === $objectFromDatabase ? false : true;
+        $objectFromVersion = $latestObject !== $objectFromDatabase;
         $object = $latestObject;
```
brusch: This is obsolete, as this is done already in line `376`, right? 
I think we can even directly use `$object`, I don't see any reason for using `$latestObject` first, I guess that has historical reasons :)

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -365,23 +365,24 @@ public function getIdPathPagingInfoAction(Request $request)
...
 
         // set the latest available version for editmode
         $latestObject = $this->getLatestVersion($objectFromDatabase);
+        $object = $latestObject;
+
+        // check for lock
+        if (Element\Editlock::isLocked($request->get('id'), 'object') && ($object->isAllowed('save') || $object->isAllowed('publish') || $object->isAllowed('unpublish') || $object->isAllowed('delete'))) {
+            return $this->getEditLockResponse($request->get('id'), 'object');
+        }
+        Element\Editlock::lock($request->get('id'), 'object');
```
brusch: To be consistent, locking should only be done when user has the rights to modify the object.

brusch:
Another consistency thing: Assets and Documents should behave the same.

BlackbitNeueMedien:
Now also works for assets and documents.
Opened element gets only locked if user has permission to edit it.

brusch:
Thanks a lot!

PR #4898, opened 2019-08-28
[Backend search] Find related objects

BlackbitNeueMedien:
Normally all data fields of data objects are used to fill the backend search index. Somehow this does not apply for relations. In 15c4f4fc88fc6070dfb6e37765aaad7d98034b09 there is a change which prevents relation's data from being written to search index with commit message "index optimization". 
Is this really on purpose or why should an object not be found if a user searches for a related object of it?
This PR only adds the full path of related objects to search index to not jumble search results up too much (otherwise if we wrote all data of related objects to search index it could happen that the related objects are listed before the object which actually contains the data).

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
+++ b/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
@@ -343,17 +343,6 @@ abstract class AbstractRelations extends Data implements CustomResourcePersistin
     }
 
     /**
-     * @param $object
-     * @param mixed $params
-     *
-     * @return string
-     */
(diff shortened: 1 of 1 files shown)


brusch:
The reason for this "optimization" was that the index got quite bloated when having hundreds or even thousands of relations on objects. 
Due to the way the index of MySQL/MariaDB works, the size of the index get's quickly up to hundreds of gigabytes. 
So I'd keep it as is is for now, maybe we need a different way of storing and searching for related objects ... 🤔

brusch:
Maybe you can create an issue for that to continue the discussion of how we can improve that. 
Thanks!

PR #4899, opened 2019-08-28
Don't load by tag parent by empty id

das-peter:
I'd say we can skip to attempt loading a 0 / empty parent tag id as it will lead to null anyways.
Checking that in the function itself is less overhead (no db query, no exception handling)


brusch:
Thanks!

PR #4900, opened 2019-08-28
add primary key to http_error_log

dpfaffenbauer:
This fixes the Percona Error:

`WSREP: Percona-XtraDB-Cluster doesn't recommend use of DML command on a table (fotobuchDB.http_error_log) without an explicit primary key with pxc_strict_mode = PERMISSIVE`

### DIFF
--- a/bundles/CoreBundle/Resources/migrations/Version20190828142756.php
+++ b/bundles/CoreBundle/Resources/migrations/Version20190828142756.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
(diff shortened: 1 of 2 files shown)


brusch:
Thanks!

blankse:
@dpfaffenbauer I get following error in this migration with MySQL 5.7:
```
Migration 20190828142756 failed during Execution. Error An exception occurred while executing 'ALTER TABLE `http_error_log` ADD `id` int(11) NULL AUTO_INCREMENT PRIMARY KEY FIRST;':

SQLSTATE[42000]: Syntax error or access violation: 1171 All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead
```

PR #4902, opened 2019-08-28
[DataObject] Avoid fatal errors if class definition file is missing

blankse:
If another developer forgets to check in a class definition, it would be good if the page were still usable.

The `DataObject\ClassDefinition::getById` method log the missing class now.

### DIFF
--- a/models/DataObject/ClassDefinition.php
+++ b/models/DataObject/ClassDefinition.php
@@ -22,6 +22,7 @@ use Pimcore\Db;
 use Pimcore\Event\DataObjectClassDefinitionEvents;
 use Pimcore\Event\Model\DataObject\ClassDefinitionEvent;
 use Pimcore\File;
+use Pimcore\Logger;
 use Pimcore\Model;
 use Pimcore\Model\DataObject;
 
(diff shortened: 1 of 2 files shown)


brusch:
Thanks!

dvesh3:
@blankse can you please explain the need for this change? https://github.com/pimcore/pimcore/pull/4902/files#diff-5ae87407f6fac8a1a2a6f448d3b89866R210

problem is that it is now throwing error notices on console command `bin/console pimcore:deployment:classes-rebuild -c -d`

```
06:56:38 ERROR     [pimcore] Exception: Class definition with name xyz or ID xyz does not exist in in vendor\pimcore\pimcore\models\DataObject\ClassDefinition.php:224
```

blankse:
@dvesh3 Before this PR there was an exception if a definition was missing. I thought it would be good if we could get at least one log entry if a definition is missing. If there are any problems with it, we can remove it.

Or should we move it to the Dao::load() function in an else branch?

dvesh3:
@blankse thanks for the explanation 👍

Issue #4911, opened 2019-08-30
Add Language option in Wirecard seamless  (payment provider) config in documentation


neha04:
Adding language option in Wirecard seamless (payment provider) config , if this language is not passed , de is default language in the code.

PR #4912, opened 2019-08-30
Adding Language Option in the config in Wirecard seamless payment 

neha04:
If language is not passed 'de' is hard coded.
Resolves https://github.com/pimcore/pimcore/issues/4911


fashxp:
thx!

PR #4914, opened 2019-08-30
Defer system thumbnail creation.

christophluehr:
If a lot of images are uploaded via the UI, the parallel processing can put a significant load on the server (depending on server configuration). By deferring the thumbnail creation, an (optional) hook via the asset.postAdd event could then e.g. create jobs in a queue with a worker process generating system and application specific thumbnails without straining CPU/RAM.

  
  

Resolves #

### DIFF
--- a/models/Asset/Image.php
+++ b/models/Asset/Image.php
@@ -78,12 +78,19 @@ class Image extends Model\Asset
         // now directly create "system" thumbnails (eg. for the tree, ...)
         if ($this->getDataChanged()) {
             try {
-                $path = $this->getThumbnail(Image\Thumbnail\Config::getPreviewConfig())->getFileSystemPath();
+                
+                // defer thumbnail creation in order to prevent overloading the server on
+                // mass file uploads via the gui
+                $path = $this->getThumbnail(Image\Thumbnail\Config::getPreviewConfig())->getFileSystemPath(true);
 
                 // set the modification time of the thumbnail to the same time from the asset
                 // so that the thumbnail check doesn't fail in Asset\Image\Thumbnail\Processor::process();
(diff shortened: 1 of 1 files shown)


brusch:
Thanks for your PR! 
I guess each approach has it's pros and cons, some pros of the current approach are: 
- preview thumbnail is instantly available in the admin UI (especially useful when dealing with complex or huge images) 
- if you would generate the thumbnails on demand you could get into real troubles when using the folder previews in the admin UI - huge load, long running processes, ... +
- low quality previews are generated out of the preview thumbnails, which is faster than from the original file and also allows faster delivery in the frontend

Imho there are more pros for the current approach that for your proposed alternative.
Maybe we should think about a config option for that? 

Btw. there's also a small issue with your PR that makes it useless when using low quality previews (which is on by default): 
https://github.com/pimcore/pimcore/blob/fb0b07c086fbc8b5152093acf3bcaab905dcd504/models/Asset/Image.php#L88
`generateLowQualityPreview()` calls `$this->getThumbnail()->getFileSystemPath()` as well ;-)

christophluehr:
Well, as the overall idea of this PR was being able to prevent any automatic system thumbnail generation, an additional config option would suffice - without that, i am unable to delay/defer this. (We had customers uploading 20+ x-megapixel / 80MB files that would trigger parallel thumbnail generation and hogging the server). 

I prevented mass system preview image generation (in the grid/tree) by hooking into asset.postUpdate/Add and setting/adding a generic thumbnail image file, btw (and of course disabling low quality previews via the option).

Should i do another PR with a config option to disable the on-create/update preview generation? Any suggestion for a name?

brusch:
Yes, that would be very nice. 
What about: 
```yml
pimcore: 
    assets:
        image:
            generate_preview_thumbnail_on_save: false # (default: true)
```

PR #4920, opened 2019-09-02
Use ucfirst consistently in dynamic setters

yariksheptykin:
Followup on the discussion in https://github.com/pimcore/pimcore/pull/4887#discussion_r319880018


brusch:
Thanks a lot!

PR #4930, opened 2019-09-03
Fix syntax error in migration

blankse:
I get following error in this migration with MySQL 5.7:
```
Migration 20190828142756 failed during Execution. Error An exception occurred while executing 'ALTER TABLE `http_error_log` ADD `id` int(11) NULL AUTO_INCREMENT PRIMARY KEY FIRST;':

SQLSTATE[42000]: Syntax error or access violation: 1171 All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead
```

See #4900


brusch:
Thanks! 
Interestingly it was correct in the schema definition :) 
https://github.com/pimcore/pimcore/blob/cb131ea13297154e98c6e81058e9b2005f53ddaf/bundles/InstallBundle/Resources/install.sql#L315

dpfaffenbauer:
Ups... sorry. One beer on me ;)

PR #4938, opened 2019-09-05
[Data objects] Do not hide tooltips automatically after 5s

BlackbitNeueMedien:
ExtJs closes Ext.Tooltip instances automatically after 5s (see https://docs.sencha.com/extjs/6.2.0/modern/Ext.tip.ToolTip.html#cfg-dismissDelay).
When you have long or complex helper texts in the tooltip of a data object field this can be annoying. This PR disables automatic closing of tooltips while the mouse is still on the field - of course it gets still hidden when the mouse leaves the field label.


brusch:
Thanks!

PR #4942, opened 2019-09-05
[DataObject] New field type: Geopolyline

blankse:
Corrected version of #4913

PR #4944, opened 2019-09-05
[DataObject] Geopolygon field doesn't respect cancel in edit mode

blankse:
# Steps to reproduce

1. Create a object with geopolygon field
2. Create a polygon and save the object
3. Edit the polygon and cancel the edit mode
4. Save the object
5. Reload the object
6. Edited polygon is saved instead of old version

Noticed at #4942


brusch:
Thanks!

PR #4946, opened 2019-09-06
[doc] Fix latitude and longitude in example

ceefour:

Previous example was invalid, valid range for latitude is -90 to +90 degrees


brusch:
Thanks!

PR #4953, opened 2019-09-08
[Data Objects] Added preserve ratio option on hotspotimage

### LINKED ISSUE #4952: [Data Objects] Image-Cropper ratio bug
When I try to crop the image by height, the crop jumps back.

Additionally, I don't have a preview of the height when I crop the image with the width.

cropped-image-big

My solution is really simple. I have added an additional field for the "Preserve Ratio". PR comes immediately.

In combination with #4949, the cropping function is now much better in usability

skoch98:
Resolves #4952

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/element/tag/imagecropper.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/element/tag/imagecropper.js
@@ -21,10 +21,15 @@ pimcore.element.tag.imagecropper = Class.create({
 
         this.ratioX = null;
         this.ratioY = null;
+        this.preserveRatio = false;
         if(typeof config == "object") {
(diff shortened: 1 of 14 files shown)


brusch:
Wouldn't it be better to just set `preserveRatio: true` if both `ratioX` and `ratioY` are configured? I actually don't see any benefit of an additional config option for preserve the ratio, or I#m getting something wrong? :) 
Thanks!

skoch98:
> Wouldn't it be better to just set `preserveRatio: true` if both `ratioX` and `ratioY` are configured? I actually don't see any benefit of an additional config option for preserve the ratio, or I#m getting something wrong? :)
> Thanks!

Hmm, looks like I've been doing some extra work :-D

I'll remove the option as soon as I can.

PR #4962, opened 2019-09-10
[Admin] Add absolute path option to getLanguageFlagFile

### LINKED ISSUE #4880: \Pimcore\Tool::getLanguageFlagFile - add option to retrieve either web or file system path
This function is quite handy, however we currently have to strip to path if we want to display the flags within a web interface.

Even Pimcore itself uses this ugly workaround to display the flags... https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Resources/views/Admin/Misc/iconList.html.php#L159

dvesh3:
Resolves #4880

## Additional info  
`\Pimcore\Tool::getLanguageFlagFile("en", false);` for relative paths


ultramegatom:
thank you :)

PR #4963, opened 2019-09-10
Low quality image preview generation: support for single asset ID(s) and progress bar.

andreas-gruenwald:
1. Support to generate previews for single asset ID(s).
2. Symfony progress bar.
3. Output message is displayed before generation starts, not afterwards.


brusch:
Nice, thanks a lot!

PR #4976, opened 2019-09-12
Ecommerce | Support for product mockup objects in order manager

andreas-gruenwald:
If the cart item class is overriden to operate with product mockup objects rather than the original Pimcore object, an exception occurs in the order manager process, as soon as an order item object is saved.

By setting the original object rather than the mockup, the error is prevented and there is no need o override the createOrderItem method anymore.


fashxp:
needs some more discussion

brusch:

PR #4983, opened 2019-09-14
Optimize version cleanup process

IgorBenko:
This optimizes the version cleanup process speed without affecting the functionality in any way. In a system with a huge number of versions (multiple millions) this reduces the query time from multiple minutes to few seconds.

### DIFF
--- a/models/Version/Dao.php
+++ b/models/Version/Dao.php
@@ -147,7 +147,7 @@ class Dao extends Model\Dao\AbstractDao
                     $versionIds = array_merge($versionIds, $tmpVersionIds);
                 } else {
                     // by steps
-                    $elementIds = $this->db->fetchCol('SELECT cid,count(*) as amount FROM versions WHERE ctype = ? AND NOT public AND id NOT IN (' . $ignoreIdsList . ') GROUP BY cid HAVING amount > ?', [$elementType['elementType'], $elementType['steps']]);
+                    $elementIds = $this->db->fetchCol('SELECT cid,count(*) as amount FROM versions WHERE ctype = ? AND NOT public AND id NOT IN (' . $ignoreIdsList . ') GROUP BY cid HAVING amount > ? LIMIT 1000', [$elementType['elementType'], $elementType['steps']]);
                     foreach ($elementIds as $elementId) {
                         $count++;
                         Logger::info($elementId . '(object ' . $count . ') Vcount ' . count($versionIds));


brusch:
But it does limit the amount cleaned up version per batch to 1000, right? 
So let's say if there are 3000 versions to be deleted it needs 3 runs of the maintenance command, right?

IgorBenko:
@brusch really sorry, I somehow missed your reply. A bit further down in the code there is any way a limit of 1000 elements per run. I did multiple tests back then when we had this issue and the same amount of versions was being cleaned up in one execution, but much much faster since the new query was really fast. So this just limits the initial query so it doesn't deliver more data than is needed later in the process. If you want to be safe, we could make this number higher, but I didn't see any gain in doing that. Before this modification, we had table locks that lasted multiple minutes.

BlackbitNeueMedien:
What do you think about this version?
```php
public function maintenanceGetOutdatedVersions($elementTypes, $ignoreIds = [])
    {
        $ignoreIdsList = implode(',', $ignoreIds);
        if (!$ignoreIdsList) {
            $ignoreIdsList = '0'; // set a default to avoid SQL errors (there's no version with ID 0)
        }
        $versionIds = [];

        Logger::debug("ignore ID's: " . $ignoreIdsList);

        if (!empty($elementTypes)) {
            $count = 0;
            $stop = false;
            foreach ($elementTypes as $elementType) {
                if (isset($elementType['days']) && $elementType['days'] > 0) {
                    // by days
                    $deadline = time() - ($elementType['days'] * 86400);
                    $tmpVersionIds = $this->db->fetchCol('SELECT id FROM versions as a WHERE (ctype = ? AND date < ?) AND NOT public AND id NOT IN (' . $ignoreIdsList . ')', [$elementType['elementType'], $deadline]);
                    $versionIds = array_merge($versionIds, $tmpVersionIds);
                } else {
                    // by steps
                    $versionData = $this->db->executeQuery('SELECT cid, GROUP_CONCAT(id ORDER BY id DESC) AS versions FROM versions WHERE ctype = ? AND NOT public AND id NOT IN (' . $ignoreIdsList . ') GROUP BY cid HAVING COUNT(*) > ? LIMIT 1000', [$elementType['elementType'], $elementType['steps']]);
                    while( $versionInfo = $versionData->fetch()) {
                        $count++;
                        Logger::info($versionInfo['cid'] . '(object ' . $count . ') Vcount ' . count($versionIds));
                        $elementVersions = \array_slice(explode(',', $versionInfo['versions']), $elementType['steps']);

                        $versionIds = array_merge($versionIds, $elementVersions);

                        // call the garbage collector if memory consumption is > 100MB
                        if (memory_get_usage() > 100000000 && ($count % 100 == 0)) {
                            \Pimcore::collectGarbage();
                            sleep(1);
                        }

                        if (count($versionIds) > 1000) {
                            $stop = true;
                            break;
                        }
                    }

                    if ($stop) {
                        break;
                    }
                }
            }
        }
        Logger::info('return ' .  count($versionIds) . " ids\n");

        return $versionIds;
    }
```

Things changed:
* Only one SQL query to get all outdated versions (via ordered GROUP_CONCAT) instead of n+1 with n being the number of elements which have outdated versions
* Remove `array_unique` calls as in my opinion it is not possible for one version to belong to multiple elements

I kept the `LIMIT 1000` as in https://github.com/pimcore/pimcore/blob/22a337f716b99409db2746410abbe2ba53b87513/models/Version/Dao.php#L166-L169 the loop gets aborted if there are more than 1000 deleted versions. And as the maximum number of elements with 1000 outdated versions is 1000 (every element has exactly 1 outdated version), the limit is correct (if some elements have more than 1 outdated versions, some elements are fetched to no avail - but this is also the case with the current implementation without the `LIMIT 1000` - but worse as even more useless data gets fetched)

BlackbitNeueMedien:
@IgorBenko have you seen my proposal? Perhaps you could include that into this PR. I actually wanted that you get the credits for the PR... so please react, otherwise I will create an own PR.

IgorBenko:
@BlackbitNeueMedien sorry for my late reply, I am on vacation and don't have much access to my computer for another few days. I couldn't test your proposal, but feel free to create a separate pull request, it looks good to me and seems to be a cleaner solution! Thanks!

brusch:
Closed in favor of #5113. 
Thanks anyway! ☺

PR #4984, opened 2019-09-14
Element\Service::getValidKey - fix for names starting with both dots and spaces

IgorBenko:
When using keys that start with dot and space (e.g. `. test`) we get a key that still begins with aspace (eg. ` test`).


brusch:
Thanks!

PR #4987, opened 2019-09-16
[Code] Specify array item class

BlackbitNeueMedien:
Replace unspecific `@return array` by more specific `@return Version[]`

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -323,7 +323,7 @@ class Concrete extends AbstractObject implements LazyLoadedFieldsInterface
     }
 
     /**
-     * @return array
+     * @return Model\Version[]
      */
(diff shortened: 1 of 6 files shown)


models/DataObject/Concrete.php
```diff
@@ -323,7 +323,7 @@ public function saveVersion($setModificationDate = true, $saveOnlyVersion = true
     }
 
     /**
-     * @return array
+     * @return Model\Version[]
```
weisswurstkanone: could you change that for assets and documents as well ? :-)

BlackbitNeueMedien: Give @weisswurstkanone a finger and he will grab the whole arm ;-)

weisswurstkanone:
thx a lot!

PR #4989, opened 2019-09-17
Update 23_Localized_Fields.md

fkaefer:
\Pimcore\Translate\Website was removed with Pimcore 5 so this change might also be relevant for 5.x documentation.
I'm not sure if this is the most elegant way, but Pimcore\Model\Translation\Website::getByKeyLocalized worked for me.

  
  

Resolves #


brusch:
Obsolete due to c8edcefd951ab31f7039ea85999ef186051f7bf7
Thanks anyway 😊

PR #4990, opened 2019-09-17
[XLIFF] ignore not exportable tags in localized blocks

markus-moser:
Only text fields can be exported. Images, relations and other field types need to be ignored.


brusch:
Thanks!

PR #4992, opened 2019-09-17
[Ecommerce] Consider json_encode errors on updating index

jremmurd:
# Improvement

I had the case that there were encoding errors (Error code 5) for some product related data. As the json_encode failed the data was added to the mysql store table but not sent to the elastic search index, so there always was some inconsistency between store and the elastic index - without any errors shown. 

This PR now throws an error if the json_encode fails. This PR is related to https://github.com/pimcore/pimcore/pull/4827.


fashxp:
thx

PR #4996, opened 2019-09-19
add render function to document tag to allow throwing exceptions

### LINKED ISSUE #4353: Exception Handling for Bricks
Since Bricks are actually nothing else than a Symfony Sub Request, shouldn't symfony be able to handle exceptions for them? For example: I am currently building a Security Area where all of the backend pages are build within Pimcore. I have to check for authorization and would just throw a AccessDeniedException, but that doesn't work cause Pimcore handles any throwable directly in the Document\Tag.

https://github.com/pimcore/pimcore/blob/master/models/Document/Tag.php#L475

What do you guys think about changing that behaviour for Bricks? For BC reasons, we could add a property/method to keep the behaviour as it is right now but have the possibility to throw exceptions and let them handle from symfony?

### LINKED ISSUE #4870: [Migrate Tag Naming Strategy Command] Fetch Inline Exceptions
I just spend two days migrating a huge project from P4 to P5. The tag naming migration command (render) works quite nice but i also struggeled with the area inline exceptions. If an exception rises within the area rendering, the migration service won't notice it, because it's just a rendered string. This is quite dangerous, since a lot of elements get skipped silently. 

I fixed that locally:

```php
$content = Document\Service::render(
    $document,
    [],
    false,
    ['pimcore_editmode' => true]
);

if (strpos($content, 'color:#f00') !== false) {
    preg_match_all('/<b style=\"color:#f00\">(.*?)<\/b>/s', $content, $matches);
    $excerptStart = strpos($content, $matches[1][0]);
    $message = substr($content, $excerptStart, 10000);
    throw new \Exception(sprintf('Inline error! Error: %s. Trace (excerpt): %s...',
        (is_array($matches[1]) ? join(', ', $matches[1]) : '--'), $message)
    );
}
```

pretty nasty, huh? 😄 

My question: Should i provide a PR for that? If yes: Which branch? I think it could be useful to do this within P5 but i need you're advice for that first.

Thanks!

dpfaffenbauer:
as discussed: https://github.com/pimcore/pimcore/issues/4353

resolves #4870 
resolves #4353

### Old
Bildschirmfoto 2019-09-19 um 09 13 47
Bildschirmfoto 2019-09-19 um 09 13 42

### New
Bildschirmfoto 2019-09-19 um 09 13 50
Bildschirmfoto 2019-09-19 um 09 13 55

### DIFF
--- a/doc/Development_Documentation/03_Documents/01_Editables/README.md
+++ b/doc/Development_Documentation/03_Documents/01_Editables/README.md
@@ -16,6 +16,19 @@ The following code makes the `<h1>` headline editable in a document:
 <h1>{{ pimcore_input("headline") }}</h1>
 ```
 
+In some cases, especially with areablocks, editables could throw exceptions. Since Pimcore internally uses the `__toString` method to render
+the editables, Pimcore can't throw exceptions there. Therefore Pimcore catches the exception, and puts it out as string if DEBUG Mode is enabled.
+
+To prevent that and to have Pimcore throw the Exception, you can call editables like:
+
+```php 
+<h1><?= $this->input("headline")->render() ?></h1>
+```
+
+```twig 
+<h1>{{ pimcore_input("headline").render() }}</h1>
+```
+
 </div>
 
 Pimcore automatically displays an input widget in the edit mode and renders the content when accessing the document via the frontend. 
--- a/models/Document/Tag.php
+++ b/models/Document/Tag.php
@@ -465,16 +465,26 @@ abstract class Tag extends Model\AbstractModel implements Model\Document\Tag\Tag
      *
      * @return string
      */
+    public function render()
+    {
+        if ($this->editmode) {
+            return $this->admin();
+        }
+
+        return $this->frontend();
+    }
+
+    /**
+     * direct output to the frontend
+     *
+     * @return string
+     */
     public function __toString()
     {
         $result = '';
 
         try {
-            if ($this->editmode) {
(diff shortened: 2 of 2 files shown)


ultramegatom:
IMO we should add it to the docs and state that this is the preferred way. Maybe even go further and deprecate the __toString method?

brusch:
@ultramegatom @dpfaffenbauer 
Docs 👍
But I don't think it's necessary to make one as the preferred way, it's up to the developer to decide which behavior is the right one for each particular use-case. I don't see any problem in supporting both ways in long term.

ultramegatom:
Kinda disagree here, but maybe i just dont see it :). 

What would be a legit use-case where i would want to catch an exception and output its content as a string? Other then a dirty hack or a workaround i cant think of anything.

Calling render will simply be superior in all ways and will prevent a lot of headaches when dealing with areas. So imo we should promote it as it is best practice. The only argument for not using it would be that its less code to write which in think is not a valid argument. (and for BC reasons of course)

brusch:
Maybe for areas, but for all the other editables it's much easier to use the shorthand :) 
But anyway, I guess we should mention both approaches in the docs, for now we only have to make clear the differences, not making any recommendation.

dpfaffenbauer:
ok, will update the docs

dpfaffenbauer:
@brusch docs are done

brusch:
Very nice, thanks a lot!

PR #4998, opened 2019-09-19
Ecommerce: getCart() in CartItem class must use user-defined class definition 

andreas-gruenwald:
The getCart() method must appy the user-defined cart class, and not the default one.
If the default one is used, the application may throw random errors as soon as custom methods are applied to a cart, which do not exist within the default class implementation of Pimcore\Bundle\EcommerceFrameworkBundle\CartManager\Cart.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/CartManager/CartItem.php
+++ b/bundles/EcommerceFrameworkBundle/CartManager/CartItem.php
@@ -16,6 +16,7 @@ namespace Pimcore\Bundle\EcommerceFrameworkBundle\CartManager;
 
 use Pimcore\Cache\Runtime;
 use Pimcore\Logger;
+use Pimcore\Bundle\EcommerceFrameworkBundle\Factory;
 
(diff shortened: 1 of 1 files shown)


bundles/EcommerceFrameworkBundle/CartManager/CartItem.php
```diff
@@ -43,7 +44,8 @@ public function getSortIndex()
     public function getCart()
     {
         if (empty($this->cart)) {
-            $this->cart = Cart::getById($this->cartId);
+            $cartClass = '\\'.Factory::getInstance()->getCartManager()->getCartClassName();
+            $this->cart = $cartClass::getById($this->class);
```
fashxp: really $this->class? should that not be $this->cartId .... also phpstan tests are failing because of that. 
please double check and fix, thx.

andreas-gruenwald: You are totally right, obviously a c&p failure.
I committed the change, let's see if all tests succeed.

Issue #4999, opened 2019-09-20
Mark the last key not dirty?! 


mike4git:
In the ObjectBricks loading mechanism there is an error we were stumbling over.

In line 
https://github.com/pimcore/pimcore/blob/f57c214c2b53e56d420a09f1a4b8efd7bbe6cdd2/models/DataObject/Objectbrick/Dao.php#L111
the last key will be marked as NOT dirty but we are sure it should have been propably the object itself:
`$brick->markFieldDirty('_self', false);`

weisswurstkanone:
please provide the build number. thx!

mike4git:
Build number: v5.8.4 (which I have used for fixing)
But we guess that the following commit
https://github.com/pimcore/pimcore/commit/a0518623c639f27ec0a6ae195a3c4c8a8e6d19af#diff-30cd79ff8a376b104bc4bd82d778d0f7L99
has introduced this bug.

weisswurstkanone:
I would say already fixed via
https://github.com/pimcore/pimcore/issues/4677

mike4git:
In Issue #4677 you have corrected the saving of objects but not the loading.
And during the loading there should be no field marked as dirty, as well as the object itself.
But the markFieldDirty('_self') would not be deleted, if you would not use my change.
And the current code in the loading method is marking always the LAST $key as not dirty.
I would say this is development by accident, isn't it?!

1. The change should not happen in the save BUT in the load method.
2. The wrong line of code uses $key outside of the foreach block where it has been defined. (s. line 111 above)

weisswurstkanone:
still ... could you please try to reproduce this with at least the latest pimcore 5 build. please note that 5.8.4 is kind of outdated :-)

mike4git:
Nice try...but this didn't work ;-)

In the current version (master and 6.2.1 - https://github.com/pimcore/pimcore/blob/master/models/DataObject/Objectbrick/Dao.php) you can see the following code block (for convenience reason I've extracted it here):

```
public function load(DataObject\Concrete $object, $params = [])
{
        $fieldDef = $object->getClass()->getFieldDefinition($this->model->getFieldname());
        $values = [];
        ...
                foreach ($fieldDefinitions as $key => $fd) {
                    ...
                } // end of foreach, where $key has been defined


                $setter = 'set' . ucfirst($type);
                if ($brick instanceof DataObject\DirtyIndicatorInterface) {

                    // here $key is used again OUTSIDE of foreach
[...]

PR #5001, opened 2019-09-23
changed marking flag for complete object instead of 'last' field

mike4git:
This fixes issue #4999 .
The problem can not be easily reproduced in the demo shop because we had it during our product update scenario from outside of Pimcore.
Changing one product in Primcore leaded to a deletion of objectbricks of an another product which was in relation to the one being updated. Why?
Simply because the dirty flag map of the object bricks were handled incorrectly.

### DIFF
--- a/models/DataObject/Objectbrick/Dao.php
+++ b/models/DataObject/Objectbrick/Dao.php
@@ -108,7 +108,7 @@ class Dao extends Model\DataObject\Fieldcollection\Dao
                 $setter = 'set' . ucfirst($type);
 
                 if ($brick instanceof DataObject\DirtyIndicatorInterface) {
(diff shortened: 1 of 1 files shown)


brusch:
@mike4git please sign the CLA - thanks ☺

weisswurstkanone:
thx a lot!

mike4git:
👍

Von meinem Huawei-Mobiltelefon gesendet

-------- Originalnachricht --------
Betreff: Re: [pimcore/pimcore] changed marking flag for complete object instead of 'last' field (#5001)
Von: Josef Aichhorn
An: pimcore/pimcore
Cc: Michael Albrecht ,Mention


Merged #5001<https://github.com/pimcore/pimcore/pull/5001> into master.

Issue #4901, opened 2019-08-28
Do not use MEMORY Storage-Engine


dpfaffenbauer:
Percona produces this lock

```
[ERROR] WSREP: Percona-XtraDB-Cluster prohibits use of DML command on a table (fotobuchDB.locks) that resides in non-transactional storage engine with pxc_strict_mode = ENFORCING or MASTER
```

Guess that happens cause Pimcore creates a transaction and executes commands on the MEMORY locks table, thus Percona complaining about it. Not sure though

brusch:
Let's change the locks table to `InnoDb` storage engine, not real benefit of the memory engine anyway.

PR #5009, opened 2019-09-24
Workaround for database services which doesn't support MyISAM

mgomse:
This PR adds a workaround for MySQL services which don't support MyISAM engine. It was created mainly for Azure Database for [MySQL service]((https://azure.microsoft.com/en-us/services/mysql/) - https://docs.microsoft.com/en-us/azure/mysql/concepts-limits#unsupported

## Additional info  

This PR should make no changes to the actual installation process - it is just an addition:

* `is-myisam-supported` -cert-path won't be asked interactively
* if `is-myisam-supported` parameter will be omitted in CLI (or env variables) then the installation is like it was until now

Theoretically, it could be done better but then `installerNeedsOption` usage should be rewritten (e.g. currently parameters like `skip-database-data-dump` doesn't work because of this method), so I decided that in one PR should be done only one change.

### DIFF
--- a/bundles/InstallBundle/Command/InstallCommand.php
+++ b/bundles/InstallBundle/Command/InstallCommand.php
@@ -117,7 +117,7 @@ class InstallCommand extends Command
             ],
             'is-myisam-supported' => [
                 'description' => 'If set to "no" then MyISAM engine is changed to InnoDB in all tables (useful e.g. in Azure database for MySQL)',
-                'mode' => InputOption::VALUE_REQUIRED,
+                'mode' => InputOption::VALUE_OPTIONAL,
                 'default' => 'yes',
                 'group' => 'db_credentials'
             ],
--- a/bundles/InstallBundle/Command/InstallCommand.php
+++ b/bundles/InstallBundle/Command/InstallCommand.php
@@ -115,6 +115,12 @@ class InstallCommand extends Command
                 'default' => 3306,
                 'group' => 'db_credentials',
             ],
+            'is-myisam-supported' => [
+                'description' => 'If set to "no" then MyISAM engine is changed to InnoDB in all tables (useful e.g. in Azure database for MySQL)',
+                'mode' => InputOption::VALUE_REQUIRED,
+                'default' => 'yes',
+                'group' => 'db_credentials'
+            ],
             'skip-database-structure' => [
                 'description' => 'Skipping creation of database structure during install',
                 'mode' => InputOption::VALUE_OPTIONAL,
--- a/bundles/InstallBundle/Installer.php
+++ b/bundles/InstallBundle/Installer.php
@@ -87,6 +87,11 @@ class Installer
     private $importDatabaseDataDump = true;
 
     /**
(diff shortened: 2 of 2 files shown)


bundles/InstallBundle/Command/InstallCommand.php
```diff
@@ -115,6 +115,12 @@ private function getOptions()
                 'default' => 3306,
                 'group' => 'db_credentials',
             ],
+            'is-myisam-supported' => [
+                'description' => 'If set to "no" then MyISAM engine is changed to InnoDB in all tables (useful e.g. in Azure database for MySQL)',
+                'mode' => InputOption::VALUE_REQUIRED,
```
brusch: Value should be optional, right?

bundles/InstallBundle/Command/InstallCommand.php
```diff
@@ -115,6 +115,12 @@ private function getOptions()
...
+                'default' => 'yes',
```
brusch: I think we should use a bool here, and call the option `force-innodb-tables`.
Then if the option is set we set the flag accordingly. I think we should then do the same with MEMORY tables. 
Thanks!

brusch:
@mgomse any update?

bundles/InstallBundle/Command/InstallCommand.php
mgomse: Right, fixex

bundles/InstallBundle/Command/InstallCommand.php
mgomse: To be true I'm not sure about it - MyISAM problem is that it has "lack of transaction support which can lead to cases of data loss." - that's why e.g Azure doesn't support it (https://feedback.azure.com/forums/597982-azure-database-for-mysql/suggestions/19271050-add-myisam-engine-support)

However each engine has got its own advantages (probably that's why not all tables are by default InnoDb). So in my opinion we shouldn't have one potion to force all tables to be the same but according to use cases we should have possibility to "turn off" this special functionalities. And in the Azure database use-case we would like to "turn off" only the MyISAM, becasue why we should also abandon MEMORY engine advantages?

@brusch what do you think?

mgomse:
@brusch I'm very sorry that I took so long ... but finally I fixed what I could and I answered where I do not fully agree. Please let me know what do you think about it?

bundles/InstallBundle/Command/InstallCommand.php
brusch: @mgomse The MEMORY storage engine also doesn't support transactions 😉
See: https://mariadb.com/kb/en/library/show-engines/

brusch: But we're going away from the MEMORY engine anyway, see: #4901

brusch: But please, let's rename the flag and make it a boolean, then we can merge it. Thanks!

brusch:
Obsolete, see #5218

PR #5013, opened 2019-09-25
Feature/cookie notice styling

### LINKED ISSUE #5012: [Feature Request] Cookie Notice hide CSS styles block
# Feature Request
Hide styles from cookie notice template to be able to use your own fonts, look, etc.

## Feature description
* Add a new system config to remove the styles block of the cookie notice
* Change CSS to use classes and lower specificity

hadl:
Resolves #5012


brusch:
Thanks for your PR!
Unfortunately this is a proper BC break and we can't merge that at the moment. 

Imho it would be anyway better to override the twig template for customizations, rather than introducing more and more config options.

dpfaffenbauer:
Overwriting Twig templates is anyway the preferred symfony way

PR #5020, opened 2019-09-27
Fix naming of nocache parameter in documentation

andreas-gruenwald:


weisswurstkanone:
related to https://github.com/pimcore/pimcore/commit/dd14b97fee36b79de0b8d127d3344b683ab70c6f

PR #5025, opened 2019-09-30
[DataObject] Classificationstore cannot add collections

kjkooistra-youwe:
Ran into this issue after updating to version 6.2.1.

How to reproduce
- Create a classification store with collections
- Add a classification store field to a data object
- Add classifications using collection(s)

No classifications are added since none are found. When debugging the controller I noticed the groups data query didn't return any results and when checking the file history I noticed this change, which causes the problem: https://github.com/pimcore/pimcore/commit/ccfc5c30869a18b05ba4cc415fdc32fedd8854e2


fashxp:
thx very much

PR #5027, opened 2019-09-30
[Data objects] Symfony service as calculated field calculator

BlackbitNeueMedien:
As (multi-)select fields, fields with path resolver and link formatter support providing a symfony service, this would be also quite nice for calculated value calculators. This PR implements this.
This enables developers to also use dependency injection for calculator classes.
At the moment this PR is implemented in a way that the `compute` and `getCalculatedValueForEditMode`methods do not need to be static, but can be. On the one hand this is quite handy, on the other hand this prevents us from providing a clean interface which would be really nice for calculator classes.
Imho calculator methods should not be static at all but changing that would result in a backward incompatible change - which is no option at the moment, I think.

### DIFF
--- a/doc/Development_Documentation/05_Objects/01_Object_Classes/01_Data_Types/33_Calculated_Value_Type.md
+++ b/doc/Development_Documentation/05_Objects/01_Object_Classes/01_Data_Types/33_Calculated_Value_Type.md
@@ -14,28 +14,24 @@ called `sum` placed inside a localizedfields container.
 ![Calculated Value Configuration](../../../img/classes-datatypes-calculated.png)
 
 
-The first step is to provide a PHP calculator class implementing at least one method which computes the result 
-for the `sum` field. An example is shown below.
+The first step is to provide a PHP calculator class implementing the `CalculatorClassInterface` interface. At least the `compute` method needs to be implemented which computes the result for the `sum` field. An example is shown below.
 
 The arguments passed into this method is the Pimcore object and the contextual information telling you which 
-calculated-value field is affected and where it is located it.
+calculated-value field is affected and where it is located at.
 
-The extent of information differs on which datatype is the owner of the calculated-value field 
+The extent of information depends on the datatype of the owner of the calculated-value field 
 (localizedfield, object brick etc.). The details are documented below.
 
 ```php
 namespace Website;
  
 use Pimcore\Model\DataObject\Concrete;
+use Pimcore\Model\DataObject\ClassDefinition\CalculatorClassInterface;
+use Pimcore\Model\DataObject\Data\CalculatedValue;
  
-class CalculatorDemo
+class CalculatorDemo implements CalculatorClassInterface
 {
-    /**
-     * @param $object Concrete
-     * @param $context \Pimcore\Model\DataObject\Data\CalculatedValue
-     * @return string
-     */
-    public static function compute($object, $context) {
+    public function compute(Concrete $object, CalculatedValue $context):string {
         if ($context->getFieldname() == "sum") {
             $language = $context->getPosition();
             return $object->getXValue($language) +  $object->getYValue($language);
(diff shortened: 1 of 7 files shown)


weisswurstkanone:
maybe mark the static stuff as deprecated?

dpfaffenbauer:
There should also be a Interface to implement. If not implemented (BC) it should throw a deprecation note.

BlackbitNeueMedien:
Implemented an interface with non-static methods.
Also implemented the runtime caching of the resolved classes / services - till now only the reading part of this cache was implemented but it was never filled with data, see for example https://github.com/pimcore/pimcore/blob/0eac74bb3effee2cff3b9c1944ce3dd1d0e5e7e5/models/DataObject/ClassDefinition/Helper/LinkGeneratorResolver.php#L31-L54

dpfaffenbauer:
What about the interface for `getCalculatedValueForEditMode`?

caching: well, guess that is my fault, next beer on me for you :P

BlackbitNeueMedien:
Also added `getCalculatedValueForEditMode` - at first I wanted to leave it optional but I think you are right to have it in the interface as well because otherwise you would have to look at the documentation to know about `getCalculatedValueForEditMode` and its signature.
Returning `$this->compute()` in `getCalculatedValueForEditMode` when you do not want to differentiate between normal and edit mode is much easier.

dpfaffenbauer:
I would have added a separate interface for it, since it is not required to implement `getCalculatedValueForEditMode `

BlackbitNeueMedien:
Hmm, two interfaces for calculated value field calculators (one for `compute` and one for `getCalculatedValueForEditMode` seems to be too much in my opinion. If you do not want another behaviour for edit mode you can just call `return $this->compute(...)` in `getCalculatedValueForEditMode`

One I thing I could imagine is an abstract class:
```php
namespace Pimcore\Model\DataObject\ClassDefinition;

use Pimcore\Model\DataObject\Data\CalculatedValue;

abstract class CalculatorClass implements CalculatorClassInterface
{
    public function getCalculatedValueForEditMode(Concrete $object, CalculatedValue $context): string
    {
        return $this->compute($object, $context);
    }
}
```

What do others think?

dpfaffenbauer:
That is bad DX, having a separate interface doesn't cost too much.

Alternatively, it could also be an AbstractClass, like Symfony Form does it. But in this case, I prefer a separate interface, since `getCalculatedValueForEditMode` is not required and Pimcore fallsback to `compute` anyway.

BlackbitNeueMedien:
Let's wait for other opinions on this. I will not resist on a second interface if the majority thinks this is a good idea.

BlackbitNeueMedien:
@brusch, @weisswurstkanone, @fashxp What is your opinion about this? Would you prefer one `CalculatorClassInterface` with compute() and getCalculatedValueForEditMode() methods or two interfaces, one with compute() and one with getCalculatedValueForEditMode() ?
I think either way we decide, it will be better than now.

weisswurstkanone:
I don't have any preferences :-)

dpfaffenbauer:
I know stumbled upon this. it is ridiculous to implement two methods when one is enough.

I really think we should change that two to interfaces, having this really is ridiculous, especially if you don't know the difference:

```php

class QualityCalculator implements CalculatorClassInterface
{
    public function compute(Concrete $object, CalculatedValue $context): string
    {
        return 'Blub';
    }

    public function getCalculatedValueForEditMode(Concrete $object, CalculatedValue $context): string
    {
        return $this->compute($object, $context);
    }
}

```

solverat:
@BlackbitNeueMedien too late, but: is there any reason why `compute` only allows `string` as return type? For example, I want to re-calcuate and return a `Carbon` instance  in my compute method.

BTW within `getCalculatedValueForEditMode` I totally agree with a string as return type.

BlackbitNeueMedien:
Hmm, you are absolutely right, @solverat. I only thought of Pimcore backend usage and there values of calculated fields are always text-based. But of course if you use them anywhere else for application logic, only `getCalculatedValueForEditMode()` would have to return `string` while `compute()` could return anything. Could you create a PR for this please? There also should not be a BC problem of classes which implement compute() method with return type as this is not a problem for PHP when interface does not define return type but implementing class does:
```php
<?php
interface Test {
    public function compute();
}

class A implements Test {
    public function compute(): string {
        return 'abc';
    }
}

$a = new A();
echo $a->compute(); // outputs "abc"
```

PR #5035, opened 2019-10-01
Fix auto generated redirects for multi-site

### LINKED ISSUE #4967: Redirects: moving document from one site to another
In System Settings I have checked an option "Create automatically a redirect when a page is moved or renamed", and I moved document from one site to another one, but the created redirect is not working. It has wrong Source and Target site. 
If I change source (by removing site key) and correct Source and Target Site it works.

Tested on the latest pimcore version.

dvesh3:
Resolves #4967


brusch:
Moved the duplicate code into a trait + tightened the source regex by adding `^`  at the beginning and `$` of the pattern -> exact matches only.

PR #5037, opened 2019-10-01
[Data Objects/Grid] Object batch edit permissions fix

### LINKED ISSUE #4868: [Data objects] Grid batch actions dosen't work without publish permission
# Bug Report

### Expected behavior
User without `publish` permission on objects can use batch actions
### Actual behavior
User without permission can't edit objects. Exception `Permission denied. You don't have the rights to save this object.` is thrown.

### Steps to reproduce
1. Create a user without `publish` permission for objects (workspaces)
2. Add object
3. Try to use batch action on grid

Problem is here: https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php#L2152

I think we should throw an exception only if a user wants to edit `publish` value without permission.

We can move if statement to line 2167 and change it on:
`if ($name === 'publish' && !$object->isAllowed('publish')) {`

prugala:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Tickets       | Fix #4868

Fix for object batch edit permissions

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -2149,7 +2149,9 @@ class DataObjectHelperController extends AdminController
             $object = DataObject::getById($request->get('job'));
 
             if ($object) {
-                if (!$object->isAllowed('publish')) {
+                $name = $request->get('name');
+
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
thx!

PR #5038, opened 2019-10-01
Feature/xlsx export

### LINKED ISSUE #4455: Export to Excel
# Feature Request
Possibility to export to xls or ods format from a grid perspective.
## Feature description
Expand grid CSV export feature to include xls format. Converting between csv and excel (or ods) format is rather easy. It can solve issues with interoperability caused by different encoding used by Microsoft Office and Pimcore. Also, it would allow skipping configuration step when opening an exported file in Libre Office.
If feature proposal is accepted, I could implement it.

kubaplas:
Resolves #4455

## Additional info  
I tried to rewrite the js code to allow adding new export types easily and without overwriting core code.
Xlsx part is right now the most trivial approach - I convert already generated CSV file to Xlsx.

Zrzut ekranu z 2019-10-01 22-19-35
Zrzut ekranu z 2019-10-01 22-19-24


weisswurstkanone:
thank you very much!

PR #5053, opened 2019-10-04
Target group events 6.x

### LINKED ISSUE #5044: Target group events
I've noticed, that there are no events firing upon creating/modifying/deleting Target groups. 
In my own deveIopment, I'd like to build upon such events, so I'm going to implement the followings: 
- an Event model called TargetGroupEvent, which extends Symfony\Component\EventDispatcher\Event
- an Event class called TargetGroupEvents, which contains POST_ADD, POST_UPDATE and POST_DELETE event constants
- event dispatches in bundles/AdminBundle/Contreoller/Admin/TargetingController with the newly created events above upon Target group creation, modification and deletion
## Additional info

simon-dcode:
Resolves #5044 for PimCore 6.x

### DIFF
--- a/lib/Event/Model/TargetGroupEvent.php
+++ b/lib/Event/Model/TargetGroupEvent.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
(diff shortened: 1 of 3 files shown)


lib/Event/Model/TargetGroupEvent.php
```diff
@@ -0,0 +1,63 @@
...
+class TargetGroupEvent extends Event
+{
+
+    /**
+     * @var TargetGroup
+     */
+    protected $targetGroup;
+
+    /**
+     * DocumentEvent constructor.
```
weisswurstkanone: TargetGroupEvent constructor

simon-dcode: Thanks, fixed it!

weisswurstkanone:
thx!

PR #5055, opened 2019-10-04
[Ecommerce][IndexService][Elastic] Add database index to ecommerce index

andreas-gruenwald:
## Summary 

Mass updates on the ecommerce product index store can cause the system to slow down, because no index is used when updating the worker. By adding a new index, the performance increases massively.

## Problem
On index updates, the ecommerce queue is processed. SQL update queries, such as follows, occur frequently.
```sql
UPDATE ecommerceframework_productindex_store_elastic SET worker_id = '5d97121450e87', 
worker_timestamp = '1570181652' WHERE (crc_current != crc_index OR ISNULL(crc_index)) AND 
tenant = 'AT_hr_elastic' AND (ISNULL(worker_timestamp) 
OR worker_timestamp < '1570181352') 
LIMIT 200
```

This caused our system to slow down, so we analyzed the update statement by rewritting it into a sql statement, and then using SQL ``explain``. We found out that no index is used on the update statement!

## Solution
We added an index and executed the statement again.

In our case the query time was reduced from **1.906** seconds to **0.016** seconds per query!
image


fashxp:
thx

PR #5059, opened 2019-10-05
Query filter conditional lte fixed

petterjv:
Fixed less or equal to condition.


weisswurstkanone:
thx a lot!

PR #5060, opened 2019-10-07
Asset doctype specification

BlackbitNeueMedien:
Some doctypes in Asset class only declared parameters of type `array`. As this makes auto-complete implossible this PR specifies the type of the array.

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1400,7 +1400,7 @@ class Asset extends Element\AbstractElement
     }
 
     /**
-     * @param array $properties
+     * @param Property[] $properties
      *
      * @return $this
      */
@@ -1491,7 +1491,7 @@ class Asset extends Element\AbstractElement
     }
 
     /**
-     * @param array $versions
+     * @param Version[] $versions
      *
      * @return $this
      */
--- a/models/Element/ValidationException.php
+++ b/models/Element/ValidationException.php
@@ -38,8 +38,11 @@ class ValidationException extends \Exception
     /**
      * @param array $subItems
      */
-    public function setSubItems(array $subItems)
+    public function setSubItems($subItems)
     {
+        if(!\is_array($subItems)) {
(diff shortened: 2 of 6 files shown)


models/Asset.php
```diff
@@ -1400,7 +1400,7 @@ public function getProperties()
     }
 
     /**
-     * @param array $properties
+     * @param Property[] $properties
```
weisswurstkanone: do we need a fix for Documents and DataObjects as well ?

models/Element/ValidationException.php
```diff
@@ -38,7 +38,7 @@ public function getSubItems()
     /**
      * @param array $subItems
      */
-    public function setSubItems($subItems)
+    public function setSubItems(array $subItems)
```
dpfaffenbauer: that actually is a BC break, even though I could live with it. Still...

models/Asset.php
BlackbitNeueMedien: Done

models/Element/ValidationException.php
BlackbitNeueMedien: You are right. In my opinion this method gets mostly used internally (in fact I just wondered why I got `null` when calling `getSubItems` although its method signature says `@return array`.
But if the core team sees any difficulties in adding the typehint we can first add a deprecation hint...

weisswurstkanone: Not sure about "mostly used internally" as the ValidationException class itself is mentioned in the docs.

https://pimcore.com/docs/5.x/Development_Documentation/Extending_Pimcore/Bundle_Developers_Guide/Plugin_Backend_UI.html

But yes, it is quite unlikely that the method is used quite often.

BlackbitNeueMedien: I changed the code to only trigger a deprecation warning. We can introduce the typehint later...

PR #5064, opened 2019-10-07
Add option to display calculated value field as textarea

BlackbitNeueMedien:
Currently calculated value fields are always displayed as input fields. With longer texts this has the disadvantage that not all content is visible (or only via horizontal scrolling). For example we have cases where whole product descriptions get generated by lots of object fields.
This PR adds an option to display calculated value fields as textareas. You can choose in class definition whether you want an input field or a textarea - input being the default.


weisswurstkanone:
thx a lot !

Issue #4600, opened 2019-07-02
Default values are not saved


BlackbitNeueMedien:
I know that configured default values in field definitions are not saved to database, they are only used to preset a field's value in the GUI
But why is this the case? Also in https://github.com/pimcore/pimcore/pull/3853#issuecomment-473885914 this fact is cited - but what is the reason for this?

By not saving default values to the database a lot of problems arise, e.g. the value does not appear in exports. When you then open the object to check if the value really has not been set, it gets even more confusing because there the default value gets applied.

weisswurstkanone:
- only for create, not on update. otherwise there is no way to null it out anymore.
- if inheritance is enabled and there is no parent value then take the default value.
- if inheritance is disabled, take the default value.

BlackbitNeueMedien:
Resolved by #5594

lherbst:
Hi all,

We are currently having the exact same problem as mentioned in this issue, so we checked whether updating to the Pimcore version where this issue was resolved would help here.

As far as I can see, this would currently not take an optional OptionProvider for Selects into account... or am I missing something?

weisswurstkanone:
True, see https://github.com/pimcore/pimcore/issues/5673

PR #5069, opened 2019-10-08
[Data objects] Add default value generator service

BlackbitNeueMedien:
Currently some fields support default values, others don't - but in any case these default values are static. This PR adds an input field in field definition to set a default value generator class or service. Use-cases are e.g. URL generation based on other fields of the object, Article number generation etc. - things which most often get only generated once per object and then stay the same forever.

The specified default value generator generates a value on saving if the field is empty.

This PR only implements this for input fields but it is easily extendable to all field types (except those where it does not make sense - calculated value e.g.).

PS: As soon as #5027 gets merged I will refactor the `DefaultValueGeneratorInterface` to also use `ClassResolver`.


models/DataObject/ClassDefinition/Helper/DefaultValueGeneratorResolver.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Pimcore\Model\DataObject\ClassDefinition\Helper;
+
+use Pimcore\Logger;
+use Pimcore\Model\DataObject\ClassDefinition\DefaultValueGeneratorInterface;
+
+class DefaultValueGeneratorResolver
```
dpfaffenbauer: why not use `ClassResolver`?

BlackbitNeueMedien: This is part of #5027 which hat not been merged yet.

dpfaffenbauer: true, so we should wait for #5027 to be merged before this

BlackbitNeueMedien:
I added context information to the Field definition class similar to `CalculatedValue`'s `ownerType`, `ownerName`, `index` etc. as these information can be quite useful for generating values.
Sadly I fail to get the name of the brick container field (in `CalculatedValue` this would be `ownerName`, see https://pimcore.com/docs/6.x/Development_Documentation/Objects/Object_Classes/Data_Types/Calculated_Value_Type.html#page_Context-Information-for-Calculation-Class). Any ideas how to get that?

weisswurstkanone:
I guess somehow related to
https://github.com/pimcore/pimcore/issues/4600

Can you give a judgement on what the consequences would be? If I interpret your PR correctly then it wouldn't be possible to null it out anymore?

BlackbitNeueMedien:
Now the default value is generated on loading and saving. That means if the object has an empty input field, it is tried to generate a value and display it in the editing panel  (`getDataForEditmode`). This is similar to what current static default values do.

If the user deletes the content, the default value gets generated (`getDataForResource`). This is the thing which current static default values do not so (and is addressed in #4600)

In my opinion it is ok to not be able to set the field empty, isn't it? 
If we remove the generation when the object is getting saved (`getDataForResource`), then it is possible to nullify the field but the behaviour is somehow strange - imagine the following case:
1. A product has a URL in an input field
1. Editor thinks that this URL is bad, it shall be newly generated
1. Editor removes content if URL field
1. Editor reloads and is happy
But actually the new value is only generated in `getDataForEditmode` - in the database the value has not changed. This is quite confusing as described in #4600.

brusch:
@BlackbitNeueMedien we have recently made some changes regarding default values, see #5594 .
I think this change is also relevant for this PR in terms of the nullify question. 
This PR should also implement the generator for all other data-types that support default values. 
Thanks a lot!

models/DataObject/ClassDefinition/Helper/DefaultValueGeneratorResolver.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    public static $cache = [];
+
+    /**
+     * @param $generatorClass
+     *
+     * @return DefaultValueGeneratorInterface
+     */
+    public static function resolveGenerator($generatorClass)
```
weisswurstkanone: There is already Pimcore\Model\DataObject\ClassDefinition\Helper\ClassResolver::resolve
which does the same magic

weisswurstkanone:
Could you please update the dev documentation as well?

weisswurstkanone:
@BlackbitNeueMedien ping. would be nice if we could add this to the next release

brusch:
Ping 🏓
Quite old already, so I guess it would be good to get it out of the way 😊

models/DataObject/ClassDefinition/Helper/DefaultValueGeneratorResolver.php
BlackbitNeueMedien: Actually this method does use `Pimcore\Model\DataObject\ClassDefinition\Helper\ClassResolver::resolve()` - what change are you thinking of?

weisswurstkanone: Comment is outdated then. Noticed that there are still changes going on. Please let me know when you are done. Thanks a lot!

models/DataObject/ClassDefinition/Data/Input.php
```diff
@@ -38,6 +39,11 @@ class Input extends Data implements ResourcePersistenceAwareInterface, QueryReso
      */
     public $width;
 
+    /**
+     * @var string
```
blankse:
```suggestion
     * @var string|null
```

models/DataObject/ClassDefinition/Data/Input.php
```diff
@@ -300,4 +308,37 @@ public function isFilterable(): bool
...
+     *
+     * @return null|int
+     */
+    protected function doGetDefaultValue($object, $context = [])
+    {
+        return $this->getDefaultValue();
+    }
+
+    /**
+     * @return int|null
```
blankse:
```suggestion
     * @return string|null
```

models/DataObject/ClassDefinition/Data/Input.php
```diff
@@ -300,4 +308,37 @@ public function isFilterable(): bool
...
+     */
+    public function getDefaultValue()
+    {
+        return $this->defaultValue;
+    }
+
+    /**
+     * @param int $defaultValue
```
blankse:
```suggestion
     * @param string $defaultValue
```

BlackbitNeueMedien:
@weisswurstkanone #5850 is about unique fields. This PR is about default values. I listed the data types which currently use `\Pimcore\Model\DataObject\Traits\DefaultValueTrait`.

weisswurstkanone:
Yes, but I noticed that the line talking about unique fields was touched as well. Maybe just a re-formatting thing.

BlackbitNeueMedien:
Default value generator would also make sense for URL slug data type, wouldn't it? I actually implemented this feature for URL slugs based on an object's current data - but as url slug datatype did not exist at that time, I used input.

weisswurstkanone:
Makes definitely sense and would be a nice improvement. Please add it to this PR if possible 👌

brusch:
@BlackbitNeueMedien any updates? 😊 Thanks a lot!

BlackbitNeueMedien:
Not much time at the moment. Could you review it without the support of URL slug data type?

weisswurstkanone:
Sure, created a follow-up ticket https://github.com/pimcore/pimcore/issues/6132
@BlackbitNeueMedien thanks so far

weisswurstkanone:
Could you please check or explain why the $context information is completely different when compared to a dynamic options provider for example?

image

For the default value generator it looks like this (fields are on the same level directly under the object).

image

bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/checkbox.js
```diff
@@ -82,7 +82,16 @@ pimcore.object.classes.data.checkbox = Class.create(pimcore.object.classes.data.
         });
 
         return [
-            defaultField, {
+            defaultField,
+            {
```
weisswurstkanone, line 86: new field should also be considered when cloning the checkbox (same for other data types). see method "applySpecialData"

weisswurstkanone:
@BlackbitNeueMedien ping

bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/checkbox.js
weisswurstkanone: added support for clone

PR #5073, opened 2019-10-09
[ClassificationStore] show error message in case a group with the same name already exists

### LINKED ISSUE #5072: [ClassificationStore] - adding groups with the same name
Currently it is not possible to add multiple classification store groups with the same name.
The create-group request returns true although the group is never added

It should either be possible to add a group with the same name (the group is identified by id anyway - similar to classification store keys where adding multiple keys with the same name is possible)
Or there should be displayed an error message upon trying to add a new group with a name that already exists
tested on the new pimcore demo

AlternateIf:
show error message in case a group with the same name already exists
resolves #5072


weisswurstkanone:
thx!

PR #5076, opened 2019-10-10
[Assets] Metadata name must be unique (or unique for certain language)

BlackbitNeueMedien:
When you set an asset's metadata via `setMetadata` method it is possible to have two metadata items with the same name (and language).
To reproduce:
```php
<?php
$asset = \Pimcore\Model\Asset::getById(1234);
$asset->setMetadata([
    [
        'name' => 'test',
        'type' => 'input',
        'data' => 'test',
        'language' => null
    ], [
        'name' => 'test',
        'type' => 'input',
        'data' => 'another_test',
        'language' => null
    ]
]);
$asset->save();
```

The result is:
Bildschirmfoto 2019-10-10 um 08 50 03

This PR changes this behaviour to overwrite existing items with same key (or same key+language for localized metadata items).

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1620,9 +1620,15 @@ class Asset extends Element\AbstractElement
      */
     public function setMetadata($metadata)
     {
-        $this->metadata = $metadata;
+        if(!empty($metadata)) {
+            foreach((array)$metadata as $metaItem) {
+                if(!isset($metaItem['language'])) {
+                    $metaItem['language'] = null;
+                }
 
-        $this->setHasMetaData(!empty($metadata));
+                $this->addMetadata($metaItem['name'], $metaItem['type'], $metaItem['data'], $metaItem['language']);
+            }
+        }
 
         return $this;
     }
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1620,6 +1620,7 @@ class Asset extends Element\AbstractElement
      */
     public function setMetadata($metadata)
     {
+        $this->metadata = [];
         if(!empty($metadata)) {
             foreach((array)$metadata as $metaItem) {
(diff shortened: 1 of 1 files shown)


dvesh3:
already handled by Primary index...please see https://github.com/pimcore/pimcore/pull/4375
image

dvesh3:
It is not possible to delete the existing metadata with your approach....as it will always append!!

BlackbitNeueMedien:
Oh you are right, I edited the PR to first clear the metadata array. This way it is possible to clear the metadata.

dvesh3:
Sorry, it maybe a BC break for situations when you want to save empty metadata. let's wait for @brusch comment!

BlackbitNeueMedien:
@dvesh3 do you mean when you want to set $this->metadata to `null`? This does not make sense to me as everywhere in the code there is `[]` (field initialization).
In the end we could even distinguish in setMetadata: `$this->metadata = ($metadata === null)?null:[]` but to be honest I think we need more typization not less.
But in the end we are open - as long as we will find a solution where the setter does not allow invalid entries during runtime (so using the setter complies with UI's and addMetadata's implementation).

dvesh3:
@BlackbitNeueMedien No, I mean when you want to set empty metadata like `$asset->setMetadata("title", "input", "", "en");` not sure if we should consider this case but maybe an edge case.

BlackbitNeueMedien:
You mean `$asset->setMetadata(["title", "input", "", "en"]);`, don't you? How is this a problem? `$metaItem['data'] ?? null` would return the empty string in this case. The null coalesce operator here is just to allow not setting the `data` key - as this is allowed in `addMetadata` - whatever sense this makes as metadata without value is not saved anyway at the moment. But this PR's goal is to make `setMetadata` compatible with `addMetadata`.

dvesh3:
you are right....thnx!

Issue #5078, opened 2019-10-10
Grid view doesn't filter by an item from the classification store


davi-mo:
# Bug Report

### Expected behavior
It should be possible to filter by items from the classification store.

### Actual behavior
An exception is thrown when I am trying to filter by the item.
<img width="698" alt="Screenshot 2019-10-09 at 12 25 18" src="https://user-images.githubusercontent.com/20536831/66552402-8c454280-eb49-11e9-92ab-db56daa39d56.png">

### Steps to reproduce

1. Go to Data Objects
2. Navigate to a folder that contains items in the grid view
3. Click on grid options.
4. Select an item from the classification store
5. Click on apply
6. Click on the column that contains the item from the classification store
7. Select filters and choose an item
8. See the popup with the exception
 
I could provide a PR for this issue.

davi-mo:
@weisswurstkanone PR is opened

davi-mo:
<img width="1073" alt="Screenshot 2019-10-10 at 11 15 13" src="https://user-images.githubusercontent.com/20536831/66555719-69b62800-eb4f-11e9-902b-5ba3dee38a4f.png">

weisswurstkanone:
IMHO the right place for that would be Pimcore\Model\DataObject\ClassDefinition\Data\Multiselect::getFilterConditionExt

davi-mo:
Hi @weisswurstkanone, why do you think that would be the right place? The problem is for this specific case, the method Pimcore\Model\DataObject\ClassDefinition\Data\Multiselect::getFilterConditionExt is not been called.

weisswurstkanone:
I my case it was called.
I created a CS key of type multiselect.
Then added the key to the grid and enabled filtering.

davi-mo:
Hi @weisswurstkanone, I discovered why is not been called. That's because I am using a type select and not multi-select. And in the select class, we don't have a method called getFilterConditionExt.
<img width="1019" alt="Screenshot 2019-10-10 at 14 57 28" src="https://user-images.githubusercontent.com/20536831/66570767-89a91400-eb6e-11e9-97b4-233f50bb4085.png">

weisswurstkanone:
Great :-) My assumption was that we are talking about a multiselect field.
So my suggestion would be to implement it for the datatype instead of relying on the fallback.

davi-mo:
@weisswurstkanone Great. Done.

PR #5079, opened 2019-10-10
Grid view doesn't filter by an item from the classification store

davi-mo:
Resolves https://github.com/pimcore/pimcore/issues/5078

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Select.php
+++ b/models/DataObject/ClassDefinition/Data/Select.php
@@ -494,4 +494,23 @@ class Select extends Data implements ResourcePersistenceAwareInterface, QueryRes
             return $data;
         }
     }
+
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/Select.php
```diff
@@ -494,4 +494,43 @@ public function getDataForGrid($data, $object = null, $params = [])
...
+    /**
+     * returns sql query statement to filter according to this data types value(s)
+     *
+     * @param  $value
+     * @param  $operator
+     * @param  $params
+     *
+     * @return string
+     */
+    public function getFilterCondition($value, $operator, $params = [])
```
weisswurstkanone: guess this method is not necessary as provided by the base class

PR #5085, opened 2019-10-11
[Assets] Fix preview of video located into root folder

### LINKED ISSUE #5084: [Asset] Video thumbnail preview doesn't work for assets in the 1st tree level
# Bug Report

### Expected behavior
Upload video in Assets root folder, open and see its preview
### Actual behavior
Upload video in Assets root folder, open and don't see its preview
### Steps to reproduce
1. Upload asset in root folder
Screenshot from 2019-10-11 10-45-50
2. Open video, after preview is made, it doesn't work
Screenshot from 2019-10-11 10-48-30

### Code example
https://github.com/pimcore/pimcore/blob/9223bd712b172bc4d524ee4c750aebaa25299a80/bundles/AdminBundle/Controller/Admin/AssetController.php#L1446-L1450
Since ```$asset->getVideoThumbnailSavePath()``` already contains the path of the asset, with ```preg_replace('@' . preg_quote($asset->getPath(), '@') . '@', '', urldecode($thumbnail['formats']['mp4']))``` the path of the asset is removed from the thumbnail path.
If the asset is in the root, with ```preg_replace``` all occurrences of ```/``` in the thumbnail path are replaced, producing an invalid path.
```/video-thumb__511__pimcore-system-treepreview/file_example_MOV_1920_2_2MB.mp4```
becomes
```video-thumb__511__pimcore-system-treepreviewfile_example_MOV_1920_2_2MB.mp4```
instead of
```video-thumb__511__pimcore-system-treepreview/file_example_MOV_1920_2_2MB.mp4```

ramundomario:
All occurrences of Asset path are wrongly removed causing bug for
preview of video stored into root folder.

  
  

Resolves #5084

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -1445,7 +1445,7 @@ class AssetController extends ElementControllerBase implements EventedController
 
         $config = Asset\Video\Thumbnail\Config::getPreviewConfig();
         $thumbnail = $asset->getThumbnail($config, ['mp4']);
-        $fsFile = $asset->getVideoThumbnailSavePath() . '/' . preg_replace('@' . preg_quote($asset->getPath(), '@') . '@', '', urldecode($thumbnail['formats']['mp4']));
+        $fsFile = $asset->getVideoThumbnailSavePath() . '/' . ltrim(urldecode($thumbnail['formats']['mp4']), $asset->getPath());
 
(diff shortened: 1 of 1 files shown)


brusch:
Thanks for your PR, but unfortunately your patch does introduce a new issue due to the way how `ltrim()` works. `ltrim()`'s 2nd parameter is a mask, that means that the order of the characters in the mask doesn't matter.  

### Example 
Asset path: `/vvv/vvv.mp4`

PR #5090, opened 2019-10-15
[quicksearch]: do not allow to consume mouse events behind quicksearch overlay

ssiergl:
## How to reproduce

- Be sure to have asset tree preview enabled
- open quicksearch
- hover over the asset tree
- preview will be opened

another case

- implement an iFrame with whatever application or target
- open quicksearch
- try to close quicksearch with click in the area of the iframe
- quicksearch does not close

## Additional information

current implementation allows the application to consume mouse events behind the quicksearch overlay

This is a bad UX on several places....e.g. if asset tree preview is enabled or if an iframe is implemented
which will consume the mouse events and prevent the quicksearch overlay to be closed.

Simply add an additional div to control this behaviour


brusch:
Nice, thanks!

Issue #1662, opened 2017-06-29
Asset folder -> List - add filter for Metadata


ckogler-elements:
## Feature description 
If you click on a asset folder you can view the assets of the folder in the "List" tab. 
It would be great if a user could filter for asset metadata in this view.

A use case for this could be: 
An asset is valid from day x to y. The user want's to filter all assets which are invalid.

Things to consider:
- Column selection (which metadatafields should be displayed)
- Saving column configuration (similar to object grids)
- Localized metadata (language selection)
- Editing Metadata in the grid view
- Show thumnails of the asset in the grid view
- CSV Export ?

PR #5093, opened 2019-10-15
[Assets] Grid Improvements - column configuration & export for Asset metadata

### LINKED ISSUE #4363: Asset csv export
# Feature Request

It would be great if there was the possibility to export the `list` tab for assets as a simple csv file.

image

dvesh3:
Resolves #4363 & partially #1662

## Additional info  
1. Added option to define and save metadata column configuration
image

2. Localized metadata
image
3. Thumbnail preview
image

4. Filter by metadata columns
image

5. CSV & XLSX Export
image


bundles/CoreBundle/Resources/migrations/Version20191015131700.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
```
weisswurstkanone: could you please remove or adapt this ?

weisswurstkanone:
I think we should try to share obvious stuff with the object grid configuration/csv export implementation which basically does the same/similar thing.

weisswurstkanone:
Seems that the filter does not respect the language selection.

- add default copyright text "XYZ"
- add german copyright text "XXX"
- switch to German
- XXX is displayed
- but now filter for XXX => no result (wrong)

- still in German filter for the default "XYZ" text
- the German result is shown (feels at least wrong)

image

dvesh3:
@weisswurstkanone filters are independent of language selection as we display fallback values for metadata with no language. so filtering on both text 'XXX' and 'XYZ' should work as of now but yes, it may lead to confusion.

**Solution**: Select language in grid config and have strict language match for metadata display then filter should also consider language however downside of this approach will be that you cannot see metadata with/without language in one view. e.g. title(en) & authorName.

Refactoring in separate PR.

brusch:
Additional topics to consider: 
- batch operations
- batch operations for selected elements

bundles/CoreBundle/Resources/migrations/Version20191015131700.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
weisswurstkanone: maybe changes in install.sql are missing?

dvesh3: done :)

models/Asset/Service.php
```diff
@@ -166,18 +171,98 @@ public function copyContents($target, $source)
...
+                'filename' => $asset->getKey(),
+                'creationDate' => $asset->getCreationDate(),
+                'modificationDate' => $asset->getModificationDate(),
+                'idPath' => Element\Service::getIdPath($asset),
+            ];
+
+            $requestedLanguage = str_replace("default","", $requestedLanguage);
+
+            foreach ($fields as $field) {
+                if ($field == "preview") {
```
weisswurstkanone: Please note that a metadata key can be called "preview". Such a key cannot be added to the grid!

weisswurstkanone: same for all system columns

bundles/AdminBundle/Resources/public/js/pimcore/asset/listfolder.js
```diff
@@ -154,35 +269,153 @@ pimcore.asset.listfolder = Class.create({
...
+                }));
+            } else if (field.type == "select") {
+                gridColumns.push({text: field.key,  width: this.getColumnWidth(field, 200), sortable: false,
+                    dataIndex: field.key, filter: 'string'});
+            } else if (field.type == "document" || field.type == "asset" || field.type == "object") {
+                gridColumns.push({text: field.key,  width: this.getColumnWidth(field, 300), sortable: false,
+                    dataIndex: field.key});
+            } else {
+                gridColumns.push({text: field.key,  width: this.getColumnWidth(field, 250), sortable: false,
+                    dataIndex: field.key, filter: 'string'});
```
weisswurstkanone: just a thought: for localized metadata: can we show the language flag + key instead of concatenated key~~language ?

PR #5095, opened 2019-10-15
Update system requirements docs

jremmurd:
Curl should be be a required extension, since following the installation instructions without curl installed results in the following error:

pimcore


brusch:
Thanks! 
I've also changed it in the requirements check: 88264eaffaea7aca695f9762dfae6ab1a860ba2b

PR #5096, opened 2019-10-15
Minor Code & Performance Optimizations

BlackbitNeueMedien:
Just some minor code and performance optimizations.

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -355,7 +355,9 @@ class Concrete extends AbstractObject implements LazyLoadedFieldsInterface
     {
         if (isset($this->$key)) {
             return $this->$key;
-        } elseif ($this->getClass()->getFieldDefinition($key) instanceof Model\DataObject\ClassDefinition\Data\CalculatedValue) {
+        }
+
+        if ($this->getClass()->getFieldDefinition($key) instanceof Model\DataObject\ClassDefinition\Data\CalculatedValue) {
             $value = new Model\DataObject\Data\CalculatedValue($key);
             $value = Service::getCalculatedFieldValue($this, $value);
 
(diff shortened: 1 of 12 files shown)


dpfaffenbauer:
I recommend not doing such messy PR's which include a lot of different things, they are hard to review cause so much happened. Better split them up into multiple ones.

BlackbitNeueMedien:
These are just some obvious things I recognized while scrolling through the code. Most people would keep that as it is because these are rather small changes - resulting in nobody ever changing those.
For changes with a certain topic I agree with you.

If maintainers do not want to review this PR, then we leave all these things as they are.

dpfaffenbauer:
I agree with most changes, but there are some that I would make a separate PR like:

 - https://github.com/pimcore/pimcore/pull/5096/files#diff-a15becd6f7488c1338b86fd7272640dfR241
 - https://github.com/pimcore/pimcore/pull/5096/files#diff-a15becd6f7488c1338b86fd7272640dfR321
 
Cause they potentially could change behavior. The rest is fine, thats basically just code-styles.

BlackbitNeueMedien:
For https://github.com/pimcore/pimcore/pull/5096/files#diff-a15becd6f7488c1338b86fd7272640dfR241 I do not agree because this query's result is only used in this function, nowhere else
https://github.com/pimcore/pimcore/blob/3b710827514e28a57fad78d8bd67f13430962e77/models/DataObject/AbstractObject/Dao.php#L241-L272
Or where do you see any potential behavioral changes?

For https://github.com/pimcore/pimcore/pull/5096/files#diff-a15becd6f7488c1338b86fd7272640dfR321 I also do not agree as o_id is always an int > 0. This means when there is a result, the old implementation would return `true`, otherwise false. This is the same with returning 1. With this query we only want to check if there is a dataset - we do not care which one it is.

dpfaffenbauer:
I understand what it does, I am just saying that IMO it doesn't belong to this PR. But anyhow: it is not my decision anyway, I just gave my 2 cents.

models/DataObject/Concrete.php
```diff
@@ -569,12 +573,12 @@ public function getNextParentForInheritance()
     {
         if ($this->getParent() instanceof AbstractObject) {
             $parent = $this->getParent();
-            while ($parent && $parent->getType() == 'folder') {
+            while ($parent && $parent->getType() === 'folder') {
```
NiklasBr: Use `Object_Abstract::OBJECT_TYPE_*` constants here and below?

BlackbitNeueMedien: Very good idea. Just changed that. Thanks!

weisswurstkanone:
thx!

PR #5101, opened 2019-10-16
fix #5100 Fix for input and textarea fields when field is not editable with cha…

### LINKED ISSUE #5100: Not working field with properties "not editable" and "show counter"
# Bug Report
### Expected behavior
Show not editable field (input or textarea) with characters counter
### Actual behavior
Pimcore admin panel stops working
throw js error:
`TypeError: this.component.setReadOnly is not a function`

### Steps to reproduce
Create a class with field input and checked properties "not editable" and "show character counter"
Create an object of the created class
Open created object

jacdobro:
resolves #5100 
Fix for input and textarea fields with selected properties "not editable" and "show counter"

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/input.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/input.js
@@ -99,7 +99,7 @@ pimcore.object.tags.input = Class.create(pimcore.object.tags.abstract, {
             //init word count
             this.updateCharCount(this.component, charCount);
 
-            return Ext.create("Ext.Panel", {
+            var panelWithCharCount = Ext.create("Ext.Panel", {
                 cls: "object_field",
                 style: "margin-bottom: 10px",
                 layout: {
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/tags/input.js
```diff
@@ -124,9 +129,12 @@ pimcore.object.tags.input = Class.create(pimcore.object.tags.abstract, {
 
 
     getLayoutShow: function () {
-
         this.component = this.getLayoutEdit();
-        this.component.setReadOnly(true);
+        if (this.fieldConfig.showCharCount) {
+            this.component.items.items[0].setReadOnly(true);
```
weisswurstkanone: I would prefer an approach similar to

https://github.com/pimcore/pimcore/blob/857b22beef2503abbac7f51b0479d4a1b9235982/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/localizedfields.js#L99

instead of exploting implementation details ;-)

jacdobro:
Better fix in #5102 pull request

PR #5102, opened 2019-10-16
#5100 Fix for input fields when field is not editable with characters…

jacdobro:
Fix for not editable input field with counter


weisswurstkanone:
thx!

PR #5103, opened 2019-10-16
[Data objects] Cache children and siblings

BlackbitNeueMedien:
Currently caching of an element's children and siblings seems a bit broken:
* fields `lastGetChildrenObjectTypes` and `lastGetSiblingObjectTypes` are never written
* if above fields were written, only exact matches of previously fetched object types were reused
* fields `o_hasChildren` and `o_hasSiblings` are never written
* `$unpublished` has not been regarded in cache as well as `o_childrenSortBy`

PS: The addition of the `array` typehints is not a BC break because in https://github.com/pimcore/pimcore/blob/b34dc774c6412219bd149640155c27588955c92a/models/DataObject/AbstractObject/Dao.php#L298-L310 `implode` gets executed with `$objectTypes` - this would fail if no array is given.


BlackbitNeueMedien:
Sorry @weisswurstkanone, I still work in this PR. I added "[WIP]" in the title. I will notify you if I am done...

weisswurstkanone:
Ah, missed that

brusch:
@BlackbitNeueMedien 
Regarding the `[WIP]` prefix used in the title: GitHub supports _draft pull requests_ for a while now, see: https://github.blog/2019-02-14-introducing-draft-pull-requests/ 

Maybe next time 😉

BlackbitNeueMedien:
While caching works now (and also the tests are green ;-) I wonder if the implementation is too complex to be maintained. The advantages of caching naturally only come to play if during a script's runtime the children / siblings are fetched more than once. What do you think?

weisswurstkanone:
It turned out that a unit test I am working on right now is failing.


---------
1) ObjectTest: Cache children and siblings
 Test  tests/cache/Runtime/ObjectTest.php:testCacheChildrenAndSiblings
Expected one child
Failed asserting that 3 matches expected 2.
/home/pimcore-5/demo-basic/vendor/phpunit/phpunit/src/Framework/Constraint/IsEqual.php:108

The test itself should be quite clear. It creates a parent foldert, then one child, calls getChildren (result count is one of course), creates a second child, result count is three then (instead of 2) !?

Code:

<?php

namespace Pimcore\Tests\Cache\Core;

use Pimcore\Model\DataObject\Folder;
use Pimcore\Model\DataObject\Service;
use Pimcore\Tests\Test\ModelTestCase;
use Pimcore\Tests\Util\TestHelper;

/**
 * Class ObjectTest
 * @package Pimcore\Tests\Cache\Core
 *
 * extends ModelTestCase so that DB gets cleaned up
 */
class ObjectTest extends ModelTestCase
{

    /**
     * see https://github.com/pimcore/pimcore/pull/5103/files
     * @throws \Exception
     */
    public function testCacheChildrenAndSiblings()
    {
        // create parent folder
        $root = Service::createFolderByPath("/" . uniqid());
        $this->assertTrue($root->getId() > 0, "Root must have a valid DB id");

        $firstChild = new Folder();
        $firstChild->setKey(uniqid());
        $firstChild->setParentId($root->getId());
        $firstChild->save();

        $firstChild = Folder::getById($firstChild->getId(), true);       // force from DB
        $this->assertTrue($firstChild->getId() > 0, "Child must have a valid DB id");
        $this->assertEquals($root->getId() + 1, $firstChild->getId(), "Expected different child ID, " . $firstChild->getId() . " " . $root->getId());


        $childs1 = $root->getChildren();
        $this->assertEquals(1, count($childs1), "Expected one child");


        $secondChild = new Folder();
        $secondChild->setKey(uniqid());
        $secondChild->setParentId($root->getId());
        $secondChild->save();

        $secondChild = Folder::getById($secondChild->getId(), true);       // force from DB
        $this->assertTrue($secondChild->getId() > 0, "Child must have a valid DB id");
        $this->assertEquals($firstChild->getId() + 1, $secondChild->getId(), "Expected different child ID");

        $childs2 = $root->getChildren();
        $this->assertEquals(2, count($childs2), "Expected one child");
    }
}

image

Could you please check?

Issue #3951, opened 2019-02-18
Changing metadata of AdvancedManyToManyObjectRelation does not mark field dirty


BlackbitNeueMedien:
# Bug Report

### Expected behavior 
Changing a meta field's value of an AdvancedManyToManyObjectRelation should mark the field dirty.
  
### Actual behavior  
Field does not get marked dirty and so does not get saved.  

### Code example 
```php
<?php
use Pimcore\Model\DataObject\Concrete;

// if you uncomment the follwoing line, everything works as expected
// Concrete::setDisableDirtyDetection(true);

$object = Concrete::getById(52807);

$refObject = Concrete::getById(52053);
$metaObject = new Pimcore\Model\DataObject\Data\ObjectMetadata('Distri', ['scaled_prices'], $refObject);
$metaObject->setScaled_prices('abc');
$object->setDistri([$metaObject]);
var_dump($object->getDistri()[0]);
$object->save();

$metaObject = $object->getDistri()[0];
$metaObject->setScaled_prices('def');
var_dump($object->getDistri()[0]);
$object->save();

// meta field scaled_prices still contains "abc"
```

In https://github.com/pimcore/pimcore/blob/848ec673a728e8d1e1bd85eefea1134992dc9a6c/models/DataObject/Data/ObjectMetadata.php#L85 $this->markMeDirty() gets called. This call is handled by https://github.com/pimcore/pimcore/blob/848ec673a728e8d1e1bd85eefea1134992dc9a6c/models/DataObject/Traits/OwnerAwareFieldTrait.php#L52 but as the field `_owner` is null (at least in my case), the relation field is not set dirty.

weisswurstkanone:
Unable to reproduce.
Are you sure that this isn't a duplicate of https://github.com/pimcore/pimcore/issues/3911

BlackbitNeueMedien:
Could be a duplicate of #3911 but the example there seems to be a bit too short.
With above script the meta field contains "def" at your test? 
Can you follow my point which lead to the relation field not being marked dirty from code perspective?

weisswurstkanone:
yes, it contains "def".
Place would be 
https://github.com/pimcore/pimcore/commit/6c153018297fbee914401c7060803d36cb75e284

BlackbitNeueMedien:
Ok, really works with #3911's change. I have not tested with the latest version.

Lexipowder:
Still the problem exists or most likly through code changes it came up again:

[...]

PR #5115, opened 2019-10-21
New option in classs definition: Disallow deletion of objects if dependencies exist 

Lexipowder:
image

Helps to keep data consistent. System critial data, like masterdata, can't be deleted as long dependencies exist.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -418,6 +418,7 @@ class DataObjectController extends ElementControllerBase implements EventedContr
             $objectData['general']['allowVariants'] = $objectFromDatabase->getClass()->getAllowVariants();
             $objectData['general']['showVariants'] = $objectFromDatabase->getClass()->getShowVariants();
             $objectData['general']['showAppLoggerTab'] = $objectFromDatabase->getClass()->getShowAppLoggerTab();
+            $objectData['general']['disallowDeleteIfDependencies'] = $objectFromDatabase->getClass()->getDisallowDeleteIfDependencies();
             if ($objectFromDatabase instanceof DataObject\Concrete) {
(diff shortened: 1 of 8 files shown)


BlackbitNeueMedien:
This currently only protects objects in the Pimcore GUI, doesn't it?
To be consistent we would also need to protect objects which are tried to be deleted via code, e.g.
```php
$object = new \Pimcore\Model\DataObject\Product();
$category = new \Pimcore\Model\DataObject\Category(); // must not be deleted when it has dependencies
$object->setCategories([$category]);
$object->save();
$category->delete();
```

Lexipowder:
Our idea was to avoid any inconsistencies caused by any kind of GUI-User / System-Admin. Deletion by code would be system designer hemisphere and we hope that they know what to code. 

But I can see the point, on the other hand due to the given database structure it is always possible to delete/destroy any object by a simple SQL-command:
DELETE FROM `database`.`objects` WHERE (`o_id` = '390');
which btw leads to dead fragments in the database. 

As pimcore is an open system and the database engine, as well the structure of the database allows any system designer to delete via code on several levels, I see no chance to avoid inconsistencies caused by system designer level with the given database structure

If $object->delete() has to be extended with the check if dependencies exist in order that the PR will be accepted we will look into it. 

Please let me know your thoughts.

BTW have you seen my comment here: https://github.com/pimcore/pimcore/issues/3951
Can you double check, if you have the same issue with the current version.

brusch:
First of all thanks a lot for your contribution! 
I'm not sure if this is topic is really of general interest or if this is just a special project specific requirement. I'd rather say it's not. 

This functionality can also be easily implemented using events - even a generic configurable bundle would be possible - so there's no need to have that directly in the core. 

Any other opinions on that? 
Thanks!

brusch:
Also this PR only does checks in the UI but not on the server-side, so `$object->delete()` wouldn't be covered. But let's wait for other opinions first 😉

brusch:
Since there were no other opinions on that, I'll close that for now.
Feel free to continue the discussion, in the meantime you can use event approach described above.

PR #5117, opened 2019-10-21
[DataObjects][Blocks][LocalizedFields] Inherited data of other langua…

### LINKED ISSUE #5105: [DataObjects][Blocks][LocalizedFields] Inherited data of other languages is cleaned once inheritance is broken
### Steps to reproduce
- create class definition with localized field (with at least one attribute) in block
- activate inheritance
- add data object and add block and add data in for all languages
- create sub object -> data in block is inherited for all languages
- overwrite data in just one language, leave others as they are
- save data object and reload
-> overwritten data in block is there, data of all other languages is empty.

weisswurstkanone:
fixed #5105


fashxp:
@weisswurstkanone  please doublecheck my modifications. worked now on my side.

PR #5119, opened 2019-10-21
Fix doctype for PageSnippet::setElement()

BlackbitNeueMedien:
Fixed doctype of PageSnippet::setElement()
Shortened some other doctypes.

Are there actually any plans for a future major version to add typehints (parameter and return types) to all functions?


brusch:
Thanks! 
Future plans: not yet, but sounds like a good idea! :)

BlackbitNeueMedien:
Perhaps you could globally introduce typehints when you plan to completely cover Pimcore with unit tests (this is planned in the medium feature when I understood the Inspire presentation correctly) as a typehint often can save multiple tests.

PR #5120, opened 2019-10-21
[ClassificationStore] Error validating mandatory fields

kjkooistra-youwe:
Error message: Call to a member function getMessage() on array
Caused by classification store validation setting array data as sub items
instead of nested validation exceptions, resulting in the error when iterating
through the sub items in the admin exception listener (recursiveAddValidationExceptionSubItems).
Fixed by creating validation exceptions on group and key level and adding these
as sub items of the field validation exception.
Additionally fixed an issue in the admin exception listener where the message
based on the sub items was never passed to the response data.
Updated Concrete to only return the sub items to prevent double messages.

Pimcore 6.2.1
To reproduce:
- Add a classification store to a data object
- Set a key to mandatory
- Save the data object while the field is still empty

While I tried to test as much as possible, I may have overlooked something. Hopefully someone with more knowledge of the Pimcore internals can pick this up and check if validation works correctly within other components.

### DIFF
--- a/bundles/AdminBundle/EventListener/AdminExceptionListener.php
+++ b/bundles/AdminBundle/EventListener/AdminExceptionListener.php
@@ -61,7 +61,6 @@ class AdminExceptionListener implements EventSubscriberInterface
 
             $data = [
                 'success' => false,
-                'message' => $message,
             ];
 
             if (\Pimcore::inDebugMode()) {
@@ -73,9 +72,16 @@ class AdminExceptionListener implements EventSubscriberInterface
                 $data['type'] = 'ValidationException';
                 $code = 403;
 
+                // Reset message of validation exception to prevent duplicate message
+                $message = 'Validation failed: ';
+                if (count($ex->getSubItems()) > 1) {
+                    $message .= '<br>';
+                }
+
                 $this->recursiveAddValidationExceptionSubItems($ex->getSubItems(), $message, $data['traceString']);
             }
 
+            $data['message'] = $message;
             $response = new JsonResponse($data, $code, $headers);
             $event->setResponse($response);
 
@@ -124,19 +130,20 @@ class AdminExceptionListener implements EventSubscriberInterface
     }
 
     /**
-     * @param $items
-     * @param $message
-     * @param $detailedInfo
+     * @param ValidationException[] $items
+     * @param string &$message
+     * @param string &$detailedInfo
+     * @param int $level
      */
-    protected function recursiveAddValidationExceptionSubItems($items, &$message, &$detailedInfo)
+    protected function recursiveAddValidationExceptionSubItems(array $items, string &$message, string &$detailedInfo, int $level = 0)
     {
         if (!$items) {
             return;
         }
-        /** @var $items ValidationException[] */
+
         foreach ($items as $e) {
             if ($e->getMessage()) {
-                $message .= '<b>' . $e->getMessage() . '</b>';
+                $message .= '<b>' . str_repeat('&nbsp;', $level * 2) . $e->getMessage() . '</b>';
                 $this->addContext($e, $message);
                 $message .= '<br>';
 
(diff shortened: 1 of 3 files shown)


weisswurstkanone:
Sorry, unable to reproduce.

image

kjkooistra-youwe:
@weisswurstkanone That is odd. I was trying to check one of the demo environments as well to see if I can reproduce it there, but they seem to be down. What I don't understand is the following:

https://github.com/pimcore/pimcore/blob/master/models/DataObject/ClassDefinition/Data/Classificationstore.php#L1009
Clearly it adds the wrong data here, since subItems should be set with ValidationException[] and not [seemingly legacy array data](https://github.com/pimcore/pimcore/blob/master/models/DataObject/ClassDefinition/Data/Classificationstore.php#L993):
`$errors[] = [
    'exception' => $e,
    'language' => $validLanguage
];`

https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/EventListener/AdminExceptionListener.php#L84
Here it retrieves the subItems and within the [recursive method](https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/EventListener/AdminExceptionListener.php#L146) it calls getMessage() on each subItem recursively. So how on earth can it do that on the above array?

brusch:
@kjkooistra-youwe There is a new demo at demo.pimcore.fun 😉

weisswurstkanone:
see screenshot :-)

kjkooistra-youwe:
Ah yes, I tried that before, but I used the wrong credentials it seems. Silly me. I also can't reproduce it there, so I will try to find why it breaks on my installation and come back to you.

weisswurstkanone:
With your change it doesn't show the errors on language / item level anymore, just that something is wrong with the classification store container in general. Can you revise your PR?

weisswurstkanone:
with your change (compare to screenshot above)

image

kjkooistra-youwe:
@weisswurstkanone and @brusch 
I finally figured it out. It only happens if 'Allow inheritance' is enabled for the object. I reproduced it on the demo environment now.
getMessage_on_array

weisswurstkanone:
can you adapt the PR accordingly?

kjkooistra-youwe:
Not without affecting the other other validation I'm afraid. That is also why I initially asked for someone with more knowledge of the Pimcore internals to check it. Removing the rethrow section in Concrete::update() fixes the issue, but I'm not sure what effect this will have on other logic:

`if ($e instanceof Model\Element\ValidationException) {
    throw $e;
}`

Maybe it should be fixed in another change. This change basically rewrites the AdminExceptionListener so the recursive validation exceptions in subItems are displayed in a more readable fashion, which I assume was the original intention of this method anyway.
Formatted_validation_message
In the current situation nothing seems to be done with the message generated by the recursiveAddValidationExceptionSubItems method, since the variable is never updated in the response data.

weisswurstkanone:
hi @kjkooistra-youwe , could you please resolve the conflicts?

kjkooistra-youwe:
@weisswurstkanone This issue was fixed in https://github.com/pimcore/pimcore/pull/5233

The changes might be relevant to format the validation error message though, since currently all fields are printed in one line, as you can see in the screenshot you posted before (https://github.com/pimcore/pimcore/pull/5120#issuecomment-546212192).

As mentioned in one of my previous posts it looks like the formatting was already implemented in the commit below, but doesn't actually work.
https://github.com/pimcore/pimcore/commit/6d3b819f53d9261a010887d1b5ade2c44477956c#diff-995c084a0084b0cbc2fc0eb9db2df4da

weisswurstkanone:
So if I understand it correctly it would be maybe better to close this PR and investigate the formatting issue based on the current master status ?

kjkooistra-youwe:
I think so. Maybe create a new issue for it with a link to this PR for reference purposes. Perhaps Brusch can shed some light on it, since he seems to have implemented the formatting in the aforementioned commit.

weisswurstkanone:
done, see #5489 

thx a lot!

PR #5124, opened 2019-10-22
Ecommerce Subtenant ambigious column o_id

daves-weblab:
When using a subtenant with the basic implementation, the o_id column is ambigious since it is used in the product index table and the subtenant assortment table as well.

This will lead to an error when using the variant mode "VARIANT_MODE_INCLUDE_PARENT_OBJECT". Using the default alias "a" in the condition solves the issue.


fashxp:
thx

PR #5125, opened 2019-10-22
[Object grid] Enable text selection

BlackbitNeueMedien:
Sometimes it is good to be able to copy field contents from the object grid. By default text selection is disabled for ExtJS grid panels.
While currently text selection is possible for editable fields via its editor (but at the cost of one extra click) it is impossible for system and non-editable fields. This PR enables text selection und so also copying values from the grid. It does not affect the behaviour of inline-editing.


weisswurstkanone:
nice improvement :-)

PR #5135, opened 2019-10-23
[Calculated value fields] readOnly instead of disabled

BlackbitNeueMedien:
Render calculated fields with `readOnly` attribute instead of `disabled` to be able to select and copy the content. Somehow we forgot this in #3693.


weisswurstkanone:
thx!

PR #5145, opened 2019-10-24
Admin Session Performance Optimizations

brusch:
Avoids writing the same contents again and again to the session storage.


bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
```diff
@@ -48,9 +50,14 @@ public function __construct(SessionInterface $session)
      */
     public function getSessionId()
     {
-        return $this->useSession(function (SessionInterface $session) {
-            return $session->getId();
-        });
+        if (!$this->session->isStarted()) {
+            // this is just to initialize the session :)
+            $this->useSession(function (SessionInterface $session) {
```
NiklasBr: Use `static function (...` for another micro-optimisation.

Source: https://github.com/Ocramius/GeneratedHydrator/pull/83

brusch: Thanks for reviewing!

PR #5149, opened 2019-10-25
Fix: Empty Object Tree after update to v6.2.2

### LINKED ISSUE #5148: Empty Object Tree after update to v6.2.2
# Bug Report
If a non admin has roles with object workspaces the tree is after the update empty.

After debugging it seems its a regression from #5096

The DataObject/AbstractObject/Dao::collectParentIds() has not the own ID now.
So the The DataObject/AbstractObject/Dao::isAllowed() function returns false.

blankse:
Resolves #5148


fashxp:
thx!

PR #5150, opened 2019-10-25
Don't use complete url

Cruiser13:
If the complete url is used, css and images will not be shown with wkhtmltopdf and web2print-output.log will have errors like this:

> Warning: Failed to load http://http//example.com/static/css/custom.css (ignore)
Warning: Failed to load http://http//example.com/images/example.png (ignore)


brusch:
Thanks a lot :)

PR #5157, opened 2019-10-28
[Asset] Grid Improvements - Batch edit Asset metadata

### LINKED ISSUE #1662: Asset folder -> List - add filter for Metadata

dvesh3:
Resolves #1662

## Additional info  
This is dependent of #5093


kubaplas:
@dvesh3 Nice work! As the code is similar to one in DataObject controllers, do you have plans to extract the common base to separate services or helpers?

@brusch as I've been looking into moving out some code from Controller is there a preffered place for code (sql, logic) out of them?

dvesh3:
Hi @kubaplas thanks for your review! we are trying our best to refactor the common functionality as it also leads to testing DataObject grid,..however it will be great if you can review it again after we merge this one and provide new PR for your changes :)

For SQL queries, I would say preferred way is to go with Dao classes.

weisswurstkanone:
what happens if one creates a predefined metadata definition out of the set of default metadata fields (copyright, alt, title) ?

image

image

bundles/AdminBundle/Resources/public/js/pimcore/asset/helpers/gridCellEditor.js
```diff
@@ -0,0 +1,163 @@
+/**
```
weisswurstkanone: the whole file seems to be 95% identical to the object's grid cell editor

bundles/AdminBundle/Resources/public/js/pimcore/asset/listfolder.js
```diff
@@ -14,7 +14,7 @@
 pimcore.registerNS("pimcore.asset.listfolder");
 pimcore.asset.listfolder = Class.create(pimcore.asset.helpers.gridTabAbstract, {
 
-    systemColumns: ["id", "type", "fullpath", "filename", "creationDate", "modificationDate", "preview", "size"],
+    systemColumns: ["id~system", "type~system", "fullpath~system", "filename~system", "creationDate~system", "modificationDate~system", "preview~system", "size~system"],
```
weisswurstkanone: please create a seperate issue for disallowing ~ in the metadata key

models/Asset/Service.php
```diff
@@ -185,6 +185,7 @@ public static function gridAssetData($asset, $fields = null, $requestedLanguage
 
         if ($asset instanceof Asset && !empty($fields)) {
             $data = [
+                'id' => $asset->getid(),
```
weisswurstkanone: ->getId()

dvesh3:
Predefined Metadata will not show up in grid config if it has same key as Default Metadata.
> what happens if one creates a predefined metadata definition out of the set of default metadata fields (copyright, alt, title) ?
> 
> image
> 
> image

PR #5159, opened 2019-10-28
Set correct parent value

simon-dcode:
I've noticed, that if PimCore Admin is in an iFrame, an error occurs: the parent.pimcore value will be set incorrectly and this will cause further issues. 
With a simple line this is fixable.

## Additional info  
This setting must be done before loading the library and internal scripts, because some of those rely on this variable too.

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/Index/index.html.php
+++ b/bundles/AdminBundle/Resources/views/Admin/Index/index.html.php
@@ -707,6 +707,7 @@ $scripts = array(
 <?php // pimcore constants ?>
 <script>
     pimcore.settings = <?= json_encode($this->settings, JSON_PRETTY_PRINT) ?>;
+    try {parent.pimcore} catch (err) {parent=window}
 </script>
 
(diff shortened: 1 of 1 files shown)


simon-dcode:
Travis checks are failing on:
```
0.01s$ mysql -e "SET GLOBAL innodb_file_format=Barracuda;"
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
The command "mysql -e "SET GLOBAL innodb_file_format=Barracuda;"" failed and exited with 1 during .
```

This has no connection to my code at all.

brusch:
@simon-dcode Seems that Travis has some issue with selecting the right MariaDB version ... actually it should use `MariaDB 10.3` but obviously uses `MySQL 5.7`. 

We'll try to restart the job later today, hopefully they fixed it until then :)

simon-dcode:
Thank you!

brusch:
I'm just wondering if it should be really possible to load the Pimcore admin UI into an Iframe by default. From a security perspective I'd say better not. I'd rather make it configurable, but the default should be off.  Any other opinions?

brusch:
Closed, details see: #5163

PR #5161, opened 2019-10-29
Fix undefined method errorerror

### LINKED ISSUE #5160: Call to undefined method Symfony\Bridge\Monolog\Logger::errorerror()
# Bug Report
Error: Call to undefined method Symfony\Bridge\Monolog\Logger::errorerror()

https://github.com/pimcore/pimcore/blob/master/lib/Maintenance/Tasks/DbCleanupBrokenViewsTask.php#L64

I guess it is pretty much self-explanatory:

The method name is incorrect, it should be `error()`

lherbst:
Resolves #5160


weisswurstkanone:
thx!

PR #5175, opened 2019-10-29
Objectbricks mandatory fields are not validated correctly

### LINKED ISSUE #5172: Objectbricks mandatory fields are not validated correctly
# Bug Report
Objectbreaks mandatory fields are not validated correctly

### Expected behavior
All mandatory fields in objectbricks are validated correctly 

### Actual behavior
The second added objectbrick is not validated correctly. Object can be published although mandatory fields are empty

### Steps to reproduce
Create a objectbrick with mandatory fields and use it in objects
Add a objectbrick then publish the object. 
image
After publishing add a second objectbrick to the same object. Do not enter any data in the required fields and publish the object. The object can be published although mandatory empty fields
image

weisswurstkanone:
resolves #5172

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Objectbricks.php
+++ b/models/DataObject/ClassDefinition/Data/Objectbricks.php
@@ -747,12 +747,8 @@ class Objectbricks extends Data implements CustomResourcePersistingInterface
                     /** @var DataObject\Objectbrick\Data\AbstractData $item */
                     $item = $data->$getter();
 
-                    if ($item) {
+                    if ($item instanceof DataObject\Objectbrick\Data\AbstractData) {
                         if ($item->getDoDelete()) {
-                            continue;
-                        }
-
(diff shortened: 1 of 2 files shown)


models/DataObject/ClassDefinition/Data/Objectbricks.php
```diff
@@ -740,39 +740,46 @@ public function checkValidity($data, $omitMandatoryCheck = false)
             if ($data instanceof DataObject\Objectbrick) {
                 $validationExceptions = [];
 
-                $items = $data->getItems();
```
brusch: @weisswurstkanone I'm wondering why the empty brick isn't included in `getItems()` shouldn't it be there as well?

weisswurstkanone: @brusch at that point of time items is not set for the newly created brick

PR #5181, opened 2019-10-30
Ecommerce index service improvements and advanced error handling events in ecommerce index preparation

andreas-gruenwald:
## Index Processing Performance Improvements
On large scale systems, several index SQL queries result in long waiting times ( > 10 sec.), which can cause the system to run inefficiently. Updates include:

1) Additional indices on ecommerce index store table.
2) The update query  in ``processUpdateIndexQueue`` has been split up into a select, and an update statement. By doing so, the system runs much more stable, as the old query took > 10 seconds on a system with > 1.000.000 data records in the store table. A transaction has been introduced so that parallel workers still can be executed.

## Index Preprocessing Error Handling Event

Currently entries in the index preparation queue are marked as processed, even if errors occured, such as on processing attributes.. In this PR a ``PreprocessAttributeErrorEvent`` event is introduced that allows to override the default behavior on processing attributes:

Example listener:

```php
class IndexPreprocessListener implements EventSubscriberInterface
{
    {
        return [
             IndexServiceEvents::ATTRIBUTE_PROCESSING_ERROR => 'onPreprocessError'
        ];
    }

    public function onPreprocessError(PreprocessErrorEvent $event) {
        $event->setSkipAttribute(false);
    }
}
```
In case that `` $event->setSkipAttribute(false);`` is set, then the queue entry will remain as unprocessed, and the entry is marked (new db columns preparation_status and preparation_error).

Another event ``PreprocessErrorEvent`` has been introduced that is fired when more general errors occurs, such as when the JSON cannot be encoded. General errors always result in a preparation queue error, nothing has to be set explicitly. ``stopPropagation`` of the event can be used to abort the index processing and throw an exception immediatly.

When the preparation queue is processed entries are sorted by their preparation_status ASC, so that the processing of other products can continue regardless.

TODOs:
- discussion on the implementation details.
- Migration?
- Documentation?
- Default behavior?

## Additional fields in store table
- ``preparation_status``: is used to mark processed store entries aus faulty, and sort them to the back, so that they can be reprocessed later, again. Is currently only used if `` $event->setSkipAttribute(false);`` is set explicitly.

- ``preparation_error``: is used to save the error message on a preparation error.

- ``trigger_info``: can be used when index updates are triggered by the application. A common use case is that a category is updated, and all related products have to be updated in the index as well. The developer may utilize this field to save the origin of the update. By doing so, it is easier to detect performance issues and trace errors.


bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractBatchProcessingWorker.php
```diff
@@ -361,24 +437,44 @@ public function processUpdateIndexQueue($limit = 200)
...
-        if ($entries) {
-            foreach ($entries as $entry) {
-                Logger::info("Worker $workerId updating index for element " . $entry['id']);
-                $data = json_decode($entry['data'], true);
-                $this->doUpdateIndex($entry['o_id'], $data);
+            //fetch open IDs and assign a worker-ID. SELECT and Update statements are separated, as a combined
+            //statement can take several seconds on large-scale systems.
+
+            $this->db->beginTransaction();
+            $query = "SELECT o_id, data FROM {$this->getStoreTableName()}
```
fashxp: we need to add a FOR UPDATE here ... otherwise select does not lock the rows.

bundles/EcommerceFrameworkBundle/Model/IndexStati.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+
+namespace Pimcore\Bundle\EcommerceFrameworkBundle\Model;
+
+
+class IndexStati
```
fashxp: lets rename it to IndexStatus ... reads a bit more logical when saying IndexStatus::PREPARATION_STATUS_DONE

bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractBatchProcessingWorker.php
```diff
@@ -216,7 +231,24 @@ public function prepareDataForIndex(IndexableInterface $object)
...
+
+                        if ($event->isSkipAttribute()) {
+                            Logger::err(
+                                sprintf(
+                                    'Exception in IndexService when processing the attribute "%s": %s',
+                                    $event->getAttribute()->getName(),
+                                    $event->getException()->getMessage()
+                                )
+                            );
+                        } elseif ($event->isPropagationStopped()) {
```
fashxp: we should not use stopPropagation for this but add an additional flag on the event to decide if exception should be thrown or not

bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractBatchProcessingWorker.php
```diff
@@ -235,22 +267,63 @@ public function prepareDataForIndex(IndexableInterface $object)
                 ]);
 
                 $jsonLastError = \json_last_error();
+                $generalErrors = [];
                 if ($jsonLastError !== JSON_ERROR_NONE) {
-                    throw new \Exception("Could not encode product data for updating index. Json encode error code was {$jsonLastError}, ObjectId was {$subObjectId}.");
+                    $e = new \Exception("Could not encode product data for updating index. Json encode error code was {$jsonLastError}, ObjectId was {$subObjectId}.");
+                    $event = new PreprocessErrorEvent($e);
+                    $this->eventDispatcher->dispatch(IndexServiceEvents::GENERAL_PREPROCESSING_ERROR, $event);
+                    if ($event->isPropagationStopped()) {
```
fashxp: we should not use stopPropagation for this but add an additional flag on the event to decide if exception should be thrown or not

fashxp:
default behavior as it is now, see https://github.com/pimcore/pimcore/issues/5313 to change that for Pimcore 7

PR #5186, opened 2019-11-01
Convert special chars in username to URL encoded values.

### LINKED ISSUE #5185: Login as user fails if there is special characters in the username.
https://github.com/pimcore/pimcore/blob/27633d378bbbeb10ec84e238575f3a814b72991b/pimcore/modules/admin/controllers/UserController.php#L673

tanaydin:
Resolves #5185

## Additional info  
Create a user with & in name of it. Then try login as this user. It will fail.
(I didn't try with pimcore v5 or v6 yet)

### DIFF
--- a/pimcore/modules/admin/controllers/UserController.php
+++ b/pimcore/modules/admin/controllers/UserController.php
@@ -670,7 +670,7 @@ class Admin_UserController extends \Pimcore\Controller\Action\Admin
         if ($user) {
             $token = Tool\Authentication::generateToken($user->getName(), $user->getPassword());
             $r = $this->getRequest();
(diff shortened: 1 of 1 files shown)


kubaplas:
@tanaydin Maybe it would be better to use user id instead of username?

tanaydin:
> @tanaydin Maybe it would be better to use user id instead of username?

so instead of fixing username, I should get user id and change internal "login as" process of old pimcore?

brusch:
Obsolete due to #5182

brusch:
Oh I haven't seen that this PR was for Pimcore 4, but anyway, Pimcore 4 is end of life. 
Thanks anyway :)

tanaydin:
you welcome, I'll try contribute more :)

PR #5187, opened 2019-11-04
Omit mandatory check when an object is not to publish

mischief24:
Resolves #

Pimcore doesn't allow to edit an object with empty required fields, directly from the grid even if the user doesn't want to publish changes, but only save.

image

After below changes editing and saving from the grid is possible.


weisswurstkanone:
thx!

PR #5191, opened 2019-11-04
Transfer elementType on cloning / copying calculated value field

BlackbitNeueMedien:
Currently when you copy or clone a calculated value field its element type (input or textarea) does not get copied. This PR fixes that.


weisswurstkanone:
thx!

BlackbitNeueMedien:
I noticed that this also affects other settings - e.g. the column length and the width. I never altered these fields so I think it is not a big deal but nevertheless we should fix this also (in a separate PR).

weisswurstkanone:
can you provide a PR ?

PR #5192, opened 2019-11-05
[Document] Fix: Tags getElement method returns only saved data

blankse:
I have an areablock with items. These items have editables like an input.
In the moment it is impossible to get the current data from this input in the controller.
Because the getElement method loads the saved version of the document.

I add a setDocument and getDocument method. So we can use the modified document.

### DIFF
--- a/models/Document/PageSnippet.php
+++ b/models/Document/PageSnippet.php
@@ -344,7 +344,7 @@ abstract class PageSnippet extends Model\Document
                 $this->elements[$name] = $element;
                 $this->elements[$name]->setDataFromEditmode($data);
                 $this->elements[$name]->setName($name);
-                $this->elements[$name]->setDocumentId($this->getId());
+                $this->elements[$name]->setDocument($this);
             }
         } catch (\Exception $e) {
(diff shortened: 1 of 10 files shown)


brusch:
@blankse Thanks a lot for your PR! 
I'm afraid that this was for a reason, namely because of caching and versioning where the documents get serialized. If we go in this direction we should correctly implement `__sleep()`.

blankse:
@brusch I add the new document property to the __sleep method

brusch:
@blankse Thanks a lot! 
Maybe one more thing ... what do you think about adding the following to `setDocumentId()`: 

```php
    public function setDocumentId($id)
    {
        $this->documentId = (int) $id;
        
        if($this->document instanceof PageSnippet && $this->document->getId() != $this->documentId) {
            $this->document = null; 
        }
        
        return $this;
    }
```

Just to ensure `$this->document` can't be a different document as stored in `$this->documentId`.

blankse:
@brusch I extended the `setDocumentId()` method 👍

brusch:
Thanks! :)

PR #5195, opened 2019-11-05
Transfer "width" and "columnLength" on cloning / copying calculated value field

BlackbitNeueMedien:
Follow-up of #5191


weisswurstkanone:
thx!

PR #5203, opened 2019-11-06
Ecommerce | Reindex Command Improvement

andreas-gruenwald:
Allow reset of index for all tenants at the same time (option  ``--tenant=*``).


fashxp:
thx

PR #5206, opened 2019-11-06
[Data objects] HTML-decode object key when editing / HTML-encode key after editing

BlackbitNeueMedien:
This PR solves the following bug:
1. Create new object with key "abc & a" or change key of existing object to "abc & a" via right-click in tree -> Rename
1. Reload parent element of this object in the tree (if you new object gets created in 1. you can skip this as parent gets reloaded automatically)
1. Right-click created object -> Rename
Now in the input there is "abc \&amp; a".

It uses `Ext.util.Format.htmlDecode` for the object key input field (as HTML entities do not work in input fields)
It also applies `Ext.util.Format.htmlEncode` after renaming a key - so the behaviour is the same as when fetching keys from PHP (there `htmlspecialchars()` gets applied).

Thus this PR also solves the bug that when you rename an object to "abc \&amp; a" by purpose (for whatever reason) you got "abc & a" displayed in the tree. After reloading the parent node it turned to the correct "abc \&amp; a".

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tree.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tree.js
@@ -996,7 +996,7 @@ pimcore.object.tree = Class.create({
             elementType: "object",
             elementSubType: record.data.type,
             id: record.data.id,
-            default: record.data.text
+            default: Ext.util.Format.htmlDecode(record.data.text)
         };
         pimcore.elementservice.editElementKey(options);
     },
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -611,7 +611,7 @@ class AssetController extends ElementControllerBase implements EventedController
 
         $tmpAsset = [
             'id' => $asset->getId(),
-            'text' => $asset->getFilename(),
+            'text' => htmlspecialchars($asset->getFilename()),
             'type' => $asset->getType(),
             'path' => $asset->getRealFullPath(),
             'basePath' => $asset->getRealPath(),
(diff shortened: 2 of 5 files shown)


fashxp:
this is necessary to prevent possible cross-site-scripting.

BlackbitNeueMedien:
@fashxp Of course. That is why I changed this PR to always use HTML-encoded keys.
Or is there still a problem with the current implementation of this PR?

bundles/AdminBundle/Resources/public/js/pimcore/object/tree.js
```diff
@@ -996,7 +996,7 @@ pimcore.object.tree = Class.create({
             elementType: "object",
             elementSubType: record.data.type,
             id: record.data.id,
-            default: record.data.text
+            default: Ext.util.Format.htmlDecode(record.data.text)
```
brusch: @BlackbitNeueMedien we should do the same for assets ( https://github.com/pimcore/pimcore/blob/23e82ab797a26e3b5837b622ecbd6b64c5e74604/bundles/AdminBundle/Resources/public/js/pimcore/asset/tree.js#L1259 ) and documents ( https://github.com/pimcore/pimcore/blob/d75f5a1c2faf16ac09c8697f07d78a6e3ca9122d/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js#L1499 ) right?

BlackbitNeueMedien: @brusch Done for assets (I was not aware that "&" is valid in asset keys but in fact it is). For documents & gets replaced by "-", so I think we do not need it there, do we?

brusch:
Thanks a lot!

PR #5208, opened 2019-11-06
[Asset] Metadata - disallow tilde(~) in names

dvesh3:
Related to #5093 & #5157


models/Asset.php
```diff
@@ -1659,6 +1659,7 @@ public function addMetadata($name, $type, $data = null, $language = null)
     {
         if ($name && $type) {
             $tmp = [];
+            $name = str_replace("~", '-', $name);
```
weisswurstkanone: just wondering if you try to replace a~b with a-b but a-b already exists

dvesh3: changed replacement to "---" so now it will be an edge case!

PR #5211, opened 2019-11-07
[Object grid] Quantity value filter

BlackbitNeueMedien:
This PR adds the grid column filter for quantity value fields. It supports selecting a unit and also converts the value to be compared to compatible units, e.g. when you filter for "< 10 cm" it also finds objects which have "50 mm" in the quantity value field (given those units have been set to be [convertable](https://pimcore.com/docs/5.x/Development_Documentation/Objects/Object_Classes/Data_Types/Number_Types.html#page_Quantity-Value-Unit-Conversion)).


weisswurstkanone:
$value in not an array if you open the filter but leave the unit combo empty. This will result in an error.

image

Apart from that, awesome stuff 👍

BlackbitNeueMedien:
Thanks @weisswurstkanone - nice to hear that all those evenings were worth it.

Fixed the mentioned bug.
And now the default unit gets preselected (if set).

Looking forward to your review. When this PR is merged I will start the even more important relation filter...

weisswurstkanone:
Excellent! Thanks a lot!

PR #5213, opened 2019-11-07
Move ss option of ffmpeg to the beginning

spotbot2k:
Helps to avoid longer encoding times, see [https://www.nrg-media.de/2010/11/creating-screenshots-with-ffmpeg-is-slow/](https://www.nrg-media.de/2010/11/creating-screenshots-with-ffmpeg-is-slow/)

This one appears on large local video files, in my case it was an full-HD *.wmv. When picking a frame to generate a preview it would take more time, the higher the time offset is.

**how to reproduce:**
Click "Use current position as preview" on a long bing video file with a big offset (e.g. > 300)


brusch:
Grazie Mille 👌😘

PR #5217, opened 2019-11-07
[Documents] Pretty URL warnings

### LINKED ISSUE #5199: Unique pretty urls per site/domain
### Expected behavior
Only one unique pretty url should be allowed per site/domain.

### Actual behavior
You can create the same pretty urls (e.g. /testurl) many times, but only the first one is resolved.
At least a warning would be nice.

Pimcore version 6.2.3

dvesh3:
Resolves #5199

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/PageController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PageController.php
@@ -248,18 +248,22 @@ class PageController extends DocumentControllerBase
         $path = rtrim($path, '/');
 
         $success = true;
+        $message = null;
 
         // must start with /
         if (strpos($path, '/') !== 0) {
             $success = false;
+            $message .= "\n URL must start with /.";
         }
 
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Controller/Admin/Document/PageController.php
```diff
@@ -248,18 +248,22 @@ public function checkPrettyUrlAction(Request $request)
...
 
         // must start with /
         if (strpos($path, '/') !== 0) {
             $success = false;
+            $message .= "\n URL must start with /.";
         }
 
         if (strlen($path) < 2) {
             $success = false;
+            $message .= "\n URL must be 2 characters long.";
```
NiklasBr: Shoudn't it say?

> ...must be *at least* 2 characters long

brusch: Thanks, done!

PR #5220, opened 2019-11-08
[Asset] Image Thumbnails: add cache-buster to media-query sources

putzflorian:
CacheBuster Config was not implemented for Thumbnails with Mediaqueries


brusch:
Made a bit more generic solution using less code in df136788d6086941472a69e134808ce963c76482

PR #5222, opened 2019-11-08
[Admin] Remove jQuery in Admin UI

### LINKED ISSUE #4760: Remove jQuery in Admin UI
Replace with vanilla JS or ExtJS equivalents.

dvesh3:
##  Changes in this pull request  
Resolves #4760


NiklasBr:
BC Breaks can't wait until 7.0.0?

brusch:
@NiklasBr jQuery is currently just a leftover of earlier times when Pimcore was using ExtJS 3.4 (which required a backend library such as jQuery), and unfortunately it was used here and there for whatever reason (probably because it was just there, but really very rarely as you can see in this PR). We also never recommended to use jQuery or even mentioned it somewhere in the documentation, as it really is/was just a 3rd party dependency. 

So we think this is really just a minor "break", but of course we'll include a small code example for how to get jQuery back into the UI, if necessary.

NiklasBr:
What you are describing is *exactly* the situation I do not want and want to prevent with semantic versioning. We cannot trust Pimcore when bugfixes and backward compatibility breaks are mixed like this. :(

brusch:
I agree for things which are documented and therefore approved for general use, but that's not the case for jQuery.

dpfaffenbauer:
Still a break though. You cannot assume that 3rd party providers didn't use it cause it was not documented. There are a lot of non-documented features in Pimcore, so they are, as you are saying, all exempt BC breaks then?

brusch:
Yes, but like I said a _minor_ break.

NiklasBr:
Where are *minor* version *major* BC break thresholds documented?

brusch:
@NiklasBr Currently not really (just here and there in discussions like this one), but probably it's time to do so 😉

brusch:
Just for reference, created a ticket for that: #5232

PR #5230, opened 2019-11-11
[Assets, Documents] Do not load listing data on Asset::getList() / Document::getList()

BlackbitNeueMedien:
Currently when you call Asset::getList() or Document::getList() the listing object is not only generated but also its items get loaded. This makes it impossible to have something like
```php
$listing = Asset::getList();
$listing->addConditionParam('type != "folder"');

// without this PR listing's items contain also assets with type="folder" 
// because Listing::load() is already executed in Asset::getList()
// thus any change to the listing parameters afterwards will be ignored
foreach($listing as $asset) {
 ...
}
```
For documents applies the same.

This is especially a bug as it is possible to modify a data object listing object after having it created via `AbstractObject::getList()`.

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -26,6 +26,7 @@ use Pimcore\Logger;
 use Pimcore\Model\Element\ElementInterface;
 use Pimcore\Tool\Mime;
 use Symfony\Component\EventDispatcher\GenericEvent;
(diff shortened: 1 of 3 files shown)


brusch:
Thanks a lot!

PR #5231, opened 2019-11-11
Fix: deleting element containing pimcore_areablock

roland4432:
Reproduce:
 * Create a brick containing a pimcore_areablock('ident')
 * Insert any brick in the newly generated brick
 * delete the parent brick containing the pimcore_areablock

=> Cannot read isAncestor of 'null'

Fix: inserted check if 
```
Ext.get(id) 
```
gets an element at all

  
  

Resolves #


brusch:
Thanks a lot!

roland4432:
Another great fix >.<

PR #5233, opened 2019-11-12
[ClassificationStorre] Error validating mandatory fields

kjkooistra-youwe:
Error message: Call to a member function getMessage() on array

Caused by a series of events:

- In ClassificationStore::checkValidity() the ValitionException sub items are set with an array of error data instead of nested ValidationExceptions.
- In Concrete::update() the ValidationException is rethrown when inheritance isallowed.
- In AdminExceptionListener::onKernelException() the method recursiveAddValidationExceptionSubItems() is called.
- This in turn causes the error because it expects the sub items to be ValidationException objects and tries to call getMessage() in the foreach loop.

Fixed by changing sub items to ValidationException objects in ClassificationStore::checkValidity()

Pimcore master
To reproduce:
- Data object allows inheritance
- Add a classification store to a data object
- Set a key to mandatory
- Save the data object while the field is still empty

Also see https://github.com/pimcore/pimcore/pull/5120
@weisswurstkanone

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Classificationstore.php
+++ b/models/DataObject/ClassDefinition/Data/Classificationstore.php
@@ -974,7 +974,7 @@ class Classificationstore extends Data implements CustomResourcePersistingInterf
         }
         $items = $data->getItems();
         $validLanguages = $this->getValidLanguages();
-        $errors = [];
+        $subItems = [];
 
         if (!$omitMandatoryCheck) {
             foreach ($activeGroups as $activeGroupId => $enabled) {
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Mhm, looks a bit weird for localized fields.

image

kjkooistra-youwe:
Ok, so then that's probably what the condition is for. So how can we fix this error?

weisswurstkanone:
well, it seems that the subitems get lost

kjkooistra-youwe:
It does. But if they don't get lost it causes the fatal error with classification store. I guess the classification store validation logic should still be fixed then to use ValidationException objects as sub items.

kjkooistra-youwe:
Changed the code so the fix is applied within the ClassificationStore.

weisswurstkanone:
thx!

PR #5237, opened 2019-11-12
Fix: Unable to restore deleted objects-folder recursively from recycle bin

### LINKED ISSUE #5235: Unable to restore deleted objects-folder recursively from recycle bin
If you delete a folder in data objects and try to restore it, only the empty folder is restored. the objects are gone.

This has been fixed for asset-folders (see #5022), but unfortunately, the issue still exists for object-folders.

### Expected behavior
restoring a deleted objects-folder from recycle-bin should restore all objects inside the folder, too.

### Actual behavior
restoring a deleted objects-folder only restores the folder, but not the objects.

### Steps to reproduce

1. visit https://demo.pimcore.fun/admin/
2. delete the folder Data Objects/News
3. restore the folder News

Please fix this as soon as possible. One of our clients already lost data because of this.

blankse:
Resolves #5235

### DIFF
--- a/models/Element/Recyclebin/Item.php
+++ b/models/Element/Recyclebin/Item.php
@@ -299,8 +299,9 @@ class Item extends Model\AbstractModel
 
         if (method_exists($element, 'getChildren')) {
             if ($element instanceof DataObject\AbstractObject) {
-                // don't use the getter because this will return an empty array (variants are excluded by default)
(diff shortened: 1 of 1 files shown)


brusch:
Thanks!

brusch:
the `$unpublished` argument shouldn't be necessary, but doesn't hurt i guess.

PR #5241, opened 2019-11-13
[Class definition] Show unique checkbox for newly created fields

BlackbitNeueMedien:
Currently there is the following bug:
1. Create class
1. Create field of type `input` (but the bug happens for all fields which support a unique index)
1. In the field's properties on the right panel there is no `unique` checkbox

This PR solves this.
I wonder what the tooltip is good for:
https://github.com/pimcore/pimcore/blob/b42962724ab4f1d8343e2afea5bab854c462350e/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/data.js#L201-L202
While it currently does not even work (at least I have not seen any tooltip on hovering the checkbox), I wonder why `t("unique")` should be shown as tooltip while this is already the checkbox's label. Is there a translation missing which shall explain what `unique` means?

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/data.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/data.js
@@ -96,7 +96,7 @@ pimcore.object.classes.data.data = Class.create({
             disabled: !in_array("unique",this.availableSettingsFields),
             autoEl: {
                 tag: 'div',
-                'data-qtip': t('unique')
+                'data-qtip': t('unique_qtip')
             },
             hidden: true
         });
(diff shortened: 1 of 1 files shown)


dvesh3:
yes, this should be t("unique_qtip") 

https://github.com/pimcore/pimcore/blob/b42962724ab4f1d8343e2afea5bab854c462350e/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/data.js#L202
> While it currently does not even work (at least I have not seen any tooltip on hovering the checkbox), I wonder why t("unique") should be shown as tooltip while this is already the checkbox's label. Is there a translation missing which shall explain what unique means?

please see https://github.com/pimcore/pimcore/blob/fceafd057f98048541d6bb7841a8d2b3fffb6b2e/bundles/CoreBundle/Resources/translations/en.extended.json#L539

BlackbitNeueMedien:
Thanks @dvesh3 - I updated this PR.

brusch:
Thanks a lot!

PR #5242, opened 2019-11-13
[UI][User] "Copy to clipboard" feature for "Login as this user" dialog

podarcis:
This PR brings a new button "Copy & Close" to the "Login as this user" dialog, so the user isn't required to select & copy the URL.

Also included are minor changes such as
* Meaningful alert title
* Increased textarea height so URL will usually fit
* Make textarea readonly, as changing the value doesn't make sense


brusch:
Very nice, thanks!

PR #5243, opened 2019-11-13
[Data objects] Truncate unique fields on copying

### LINKED ISSUE #4333: Copying an object with mandatory and unique field is not possible.
I see that there are already some issues regarding this, but I am not sure whether this is wanted behavior or not:

I have got a class that contains just one field named "remoteId" (unique, mandatory).
When I create an object, publish it and then copy & paste it, this is not possible 
with "unique constraint violation".

Version: v5.7.3
Git Hash: d2acc6e0bbf363175975fce77af0557b017739cb

BlackbitNeueMedien:
Resolves #4333 

This PR truncates unique fields on pasting an object because such objects could otherwise not be saved.
It also unpublishes the object if it has unique fields because otherwise it could not be saved if the unique field was also mandatory - and also to show the user that it was not a 1:1 copy but something was modified.

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -115,7 +115,7 @@ class Service extends Model\Element\Service
      * @param AbstractObject $target
      * @param AbstractObject $source
      *
-     * @return mixed
+     * @return AbstractObject
      */
     public function copyRecursive($target, $source)
     {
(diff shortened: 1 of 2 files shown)


models/DataObject/Service.php
```diff
@@ -115,7 +115,7 @@ public static function getObjectsReferencingUser($userId)
      * @param AbstractObject $target
      * @param AbstractObject $source
      *
-     * @return mixed
+     * @return AbstractObject
```
NiklasBr: On line 127 it can return `void` as well.

brusch:
Very nice, thanks!

blankse:
Does that no longer work at the moment?

- I create a Class in the pimcore demo with only 1 input field (mandatory, unique)
- Create a object
- Copy this object
- Error when paste:
image

PR #5245, opened 2019-11-13
Fix: Better check for project root and fall back to env

irgangla:

Extend project root detection and add option to provide project root as env variable.

Resolves #

Exception while booting/running tests.


brusch:
Thanks!

PR #5249, opened 2019-11-13
Fix: Don't call jQuery.noConflict

blankse:
We use jQuery in the frontend.

Now we get in the editmode following exceptions and all JS is broken:
`Uncaught TypeError: $ is not a function`

I think its a regression from https://github.com/pimcore/pimcore/pull/5222


brusch:
Thanks!

PR #5254, opened 2019-11-14
Fix: exception  'undefined index google' on running tests

irgangla:

Save access of $config values array in Config::mapLegacyConfiguration to avoid exceptions if values are missing.

Resolves #

Broken Codeception tests packages ecommerce, model, rest an service with inital Skeleton config and DB.

## Additional info  

On loading page and on running Codeception tests with initial skeleton DB and config exceptions occur because of missing config values, e.g. service.google.*. This breaks the tests and cause errors when loading the web app.

### DIFF
--- a/lib/Config.php
+++ b/lib/Config.php
@@ -263,7 +263,7 @@ class Config
         return $config;
     }
 
-    private static function getArrayValueSave($keys, $array)
+    private static function getArrayValue($keys, $array)
     {
         $len = count($keys);
         $pointer = $array;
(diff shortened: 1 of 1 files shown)


lib/Config.php
```diff
@@ -263,6 +263,22 @@ public static function getWebsiteConfigValue($key = null, $default = null, $lang
         return $config;
     }
 
+    private static function getArrayValueSave($keys, $array)
```
brusch: We should call this method just `getArrayValue()`, thanks!

irgangla: You are right, I changed the name.

PR #5257, opened 2019-11-14
Fix: Symfony deprecation event dispatcher

irgangla:

Change order of event and eventname calls for all dispatcher calls.

Resolves #

Symfony 4.3 deprecation log

## Additional info  

With Symfony 4.3 the order of the parameters of the dispatch function was changed. Using the old order cause a deprecation error log.

This change should not have any functional impact since the parameter order is switched anyway in dispatch after the error log:

vendor\symfony\symfony\src\Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher.php
	@trigger_error(sprintf('Calling the "%s::dispatch()" method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead.', EventDispatcherInterface::class), E_USER_DEPRECATED);
	$swap = $event;
	$event = $eventName ?? new Event();
	$eventName = $swap;

vendor\symfony\symfony\src\Symfony\Component\EventDispatcher\EventDispatcher.php
	@trigger_error(sprintf('Calling the "%s::dispatch()" method with the event name as the first argument is deprecated since Symfony 4.3, pass it as the second argument and provide the event object as the first argument instead.', EventDispatcherInterface::class), E_USER_DEPRECATED);
	$swap = $event;
	$event = $eventName ?? new Event();
	$eventName = $swap;

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -393,7 +393,7 @@ class AssetController extends ElementControllerBase implements EventedController
             $context = json_decode($context, true);
             $context = $context ? $context : [];
             $event = new \Pimcore\Event\Model\Asset\ResolveUploadTargetEvent($parentId, $filename, $context);
-            \Pimcore::getEventDispatcher()->dispatch(AssetEvents::RESOLVE_UPLOAD_TARGET, $event);
+            \Pimcore::getEventDispatcher()->dispatch($event, AssetEvents::RESOLVE_UPLOAD_TARGET);
             $filename = Element\Service::getValidKey($event->getFilename(), 'asset');
             $parentId = $event->getParentId();
         }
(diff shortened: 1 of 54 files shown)


dpfaffenbauer:
that is a proper BC break cause it doesn't work with Symfony 3 that way

brusch:
@irgangla Pimcore supports both Symfony 3.4 and ^4.1 so unfortunately we can't change that as of now (Pimcore 7 topic 😉) Thanks anyway!

romainbessugesmeusy:
Hey there, wouldn't it be possible to check the symfony version and call the method adequately? For devs in my team it's getting really painful to debug the logs.

romainbessugesmeusy:
It's tedious to work with logging. It's been 6 months this problem has been reported, and as a none core developer of Pimcore I wouldn't know how to begin with. Could you please at least take a look at possible workarounds?

Capture d’écran 2020-07-20 à 18 34 21

Issue #3085, opened 2018-07-16
Mandatory fields validation issue in Objects


psuhane:
# Bug Report
  
- Client side and server side validation message doesn't have mention of all the mandatory fields including localized. 
- In tab layout, fields in a tab only get mention in validation message if user click(focus) that tab.
- Client side validation doesn't work with localized fields.
- For localized fields, validation message doesn't have mention of language(s) in which value is missing.

In class have 5 mandatory fields 
- In first tab - name(localized)
- In second tab - EAN and Article Number
- In Third tab - Size and Gender

Open a object, name field can be seen marked *, but there are 4 more fields are mandatory in two other tabs
![tab1](https://user-images.githubusercontent.com/12061880/42746549-eaf4a288-88f5-11e8-8f9d-8f2675274b45.png)

Now click on 'Save and Publish' button, server side mandatory validation message will be displayed and no mention of specific language
![tab11](https://user-images.githubusercontent.com/12061880/42746558-f8156808-88f5-11e8-8209-4762a885444f.png)

Focus on second tab, now client side validation work and validation message have mention of EAN and Article Number.. no mention of localized field and the field available in third tab

![tab2](https://user-images.githubusercontent.com/12061880/42748640-82943f6a-88fe-11e8-8487-c62bec0b060a.png)

![tab3](https://user-images.githubusercontent.com/12061880/42748697-b89a4064-88fe-11e8-8b0f-8b2bcdc82177.png)

brusch:
@weisswurstkanone let's discuss this first internally

PR #5269, opened 2019-11-15
[DataObject] Mandatory setting for Geopolygon and Geopolyline

blankse:
This PR activates the mandatory settings in Geopolygon and Geopolyline fields.
The mandatory check works at the server.

At the client the validation is removed since #3085

I think we can remove all the isInvalidMandatory JS functions, right?


weisswurstkanone:
Thanks a lot! isInvalidMandatory can be removed. Can you add this to the PR or create a seperate one please?

blankse:
@weisswurstkanone I added a commit that removes the functions

weisswurstkanone:
thx!

PR #5271, opened 2019-11-15
Enable Editable Icons for Region Panels in DataObjects

daves-weblab:
Icons were already editable for basic panel in data object's edit mode. This PR just enables the same feature for region panels as well.


weisswurstkanone:
can we add the link to the icon library as well ?

weisswurstkanone:
also see https://github.com/pimcore/pimcore/pull/5273

jremmurd:
I gues this PR is independent of #5273 as it relates to the panel/region icons not on the class definition itself. If we want to add a select to the elements inside the class layout, shouldn't we add it for all layout types which support the icon config and per default add options for the whole icon library?

brusch:
@jremmurd agree! 
@daves-weblab can you please update your PR? Thanks a lot!

brusch:
@daves-weblab ping 😊

brusch:
Closed in favor of #5735

PR #5273, opened 2019-11-17
Add icon select to classdefinition

jremmurd:
# Classdefinition icon select

This PR adds
* a select field in the classdefinition tab 
* a set of icons, all different color variants of the default dataobject icon
* an event which is dispatched in the `getIconsAction()` to customize the icons available in the select 

image


bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
```diff
@@ -1745,6 +1745,45 @@ public function getBrickUsagesAction(Request $request)
...
+     */
+    public function getIconsAction(Request $request)
+    {
+        $classId = $request->get('classId');
+
+        $path = "/bundles/pimcoreadmin/img/object-icons";
+        $icons = scandir(PIMCORE_WEB_ROOT . $path);
+
+        array_shift($icons);
+        array_shift($icons);
```
NiklasBr: Is this to remove the `.` and `..` directories? Wouldn't it be nicer to use [Symfony Finder](https://symfony.com/doc/current/components/finder.html) and filter by file name?

dvesh3: or just use rscandir helper https://github.com/pimcore/pimcore/blob/4958e801727dc5541688cd6d83f9aa7c4be69d04/bundles/AdminBundle/Resources/views/Admin/Misc/iconList.html.php#L7

jremmurd: thx for your feedback please see the latest commit

PR #5275, opened 2019-11-18
[DataObject] Start Map Search in Geo Data fields with Enter key

blankse:
At the moment we need to click at the search icon. With this PR it is possible to click the enter key.


weisswurstkanone:
merci!

PR #5278, opened 2019-11-18
[Many-to-many object relation] add option to restrict "Allow to create a new object in relation field"

domiha:
Currently, it's not possible to restrict adding a new object in the `many-to-many object relation` field. (only in general).

So I've implemented an additional checkbox to enable/disable the plus button.

### DIFF
--- a/bundles/CoreBundle/Resources/translations/pl.extended.json
+++ b/bundles/CoreBundle/Resources/translations/pl.extended.json
@@ -715,6 +715,5 @@
     "invitation_link_sent": "A temporary login link has been sent to email address: \"%s\" ",
     "geopolyline": "Geographic Polyline",
     "classificationstore_error_group_exists_msg": "Group with this name already exists",
-    "classificationstore_group_by_key": "Group by key",
-    "allow_to_create_new_object":  "Allow to create a new object in relation field"
+    "classificationstore_group_by_key": "Group by key"
 }
--- a/bundles/CoreBundle/Resources/translations/pl.extended.json
+++ b/bundles/CoreBundle/Resources/translations/pl.extended.json
@@ -715,5 +715,6 @@
     "invitation_link_sent": "A temporary login link has been sent to email address: \"%s\" ",
     "geopolyline": "Geographic Polyline",
     "classificationstore_error_group_exists_msg": "Group with this name already exists",
-    "classificationstore_group_by_key": "Group by key"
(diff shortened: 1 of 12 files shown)


models/DataObject/ClassDefinition/Data/ManyToManyObjectRelation.php
```diff
@@ -75,6 +75,11 @@ class ManyToManyObjectRelation extends AbstractRelations implements QueryResourc
      */
     public $visibleFields;
 
+    /**
+     * @var bool
+     */
+    public $allowToCreateNewObject = false;
```
brusch: Default should be true as it is the current behavior, otherwise this would be a breaking change.

bundles/CoreBundle/Resources/translations/pl.extended.json
```diff
@@ -715,5 +715,6 @@
     "invitation_link_sent": "A temporary login link has been sent to email address: \"%s\" ",
     "geopolyline": "Geographic Polyline",
     "classificationstore_error_group_exists_msg": "Group with this name already exists",
-    "classificationstore_group_by_key": "Group by key"
+    "classificationstore_group_by_key": "Group by key",
+    "allow_to_create_new_object":  "Allow to create a new object in relation field"
```
brusch: Please add the translation key only to the `en.extended.json`, all other languages are covered by POEditor.

models/DataObject/ClassDefinition/Data/ManyToManyObjectRelation.php
```diff
@@ -941,6 +946,22 @@ public function getVisibleFields()
...
+    {
+        return (bool) $this->allowToCreateNewObject;
+    }
+
+    /**
+     * @param bool $allowToCreateNewObject
+     */
+    public function setAllowToCreateNewObject($allowToCreateNewObject)
+    {
+        $this->allowToCreateNewObject = $allowToCreateNewObject;
```
brusch: We should cast to bool here in the setter, not in `isAllowToCreateNewObject()`

bundles/CoreBundle/Resources/translations/pl.extended.json
brusch: Unfortunately the translations for `de` are still there, can you please remove them as well. Thanks a lot!

brusch:
Thanks!

PR #5280, opened 2019-11-18
[ImageOptimizerQueue] write file path to 'data' column instead of 'id' column

ljthiel:
Currently, file paths of thumbnails to be optimised are written to the tmp_store.id column which has a stipulated length of 190.
In some cases, file paths will be longer than that and any characters over the limit will be cut off when inserted into the database.
The ImageOptimizeTask maintenance script then dies, as it cannot find any files with the faulty paths.

This PR creates an md5 hash of the file path as the id and writes the actual file path to the tmp_store.data column.


brusch:
Thanks a lot!

PR #5282, opened 2019-11-18
Readd checkExcecutingUser method to check if the user is "valid"

ckogler-elements:
Was removed in #4475 - Don't think that this was intended

### DIFF
--- a/lib/Tool/Console.php
+++ b/lib/Tool/Console.php
@@ -502,4 +502,28 @@ class Console
             throw new \Exception('Script execution is restricted to CLI');
         }
     }
+
+    /**
+     * @param array $allowedUsers
+     *
+     * @throws \Exception
+     */
+    public static function checkExecutingUser($allowedUsers = [])
+    {
+        $configFile = \Pimcore\Config::locateConfigFile('system.php');
+        $owner = fileowner($configFile);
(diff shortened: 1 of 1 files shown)


dvesh3:
@ckogler-elements we removed it intentionally as it was not being used anywhere as per our understanding. Please see https://github.com/pimcore/pimcore/blob/a87b12af07b43e2bb2640d098378e339ffdf8589/doc/Development_Documentation/23_Installation_and_Upgrade/07_Updating_Pimcore/01_V5_to_V6.md#removed-classes-interfaces-methods-constants-functions-aliases. 

Can you please share your usecase for this method and also we need to adapt changes in PR for new config file "system.yml"

ckogler-elements:
Hi,
the use case is that you can verify that the user which is executing the script has proper permission. For example if you have a script which imports assets or write files the created files could have wrong permissions. Dev1 executes the script as root -> file permissions are set to root so a "correct user" can't edit the files or the files won't be editable in pimcore...

dpfaffenbauer:
@ckogler-elements since it's not used internally, there is no point for Pimcore to maintain the function. If you need it in your project/bundle, feel free to copy it in there. But for Pimcore that is basically dead code.

PR #5286, opened 2019-11-19
Fixed layout bug which granted access to admin layout without permissions

PGBrandi:
Getting Admin Layout without permissions:
To reproduce this bug, simply open a object but set the "layoutId" param to an empty string beforehand. Now pimcore will show the admin layout to the user even without the necessary permission.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -472,7 +472,7 @@ class DataObjectController extends ElementControllerBase implements EventedContr
             }
             if(!$ok){
                 if (count($validLayouts) > 0) {
-                    $currentLayoutId = $validLayouts[array_key_first($validLayouts)]->getId();
+                    $currentLayoutId = reset($validLayouts)->getId();
                 }
             }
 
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -462,6 +462,20 @@ public function getAction(Request $request, EventDispatcherInterface $eventDispa
...
+            //Uses first valid layout instead of admin layout when empty
+            $ok = false;
+            foreach ($validLayouts as $key => $layout){
+                if($currentLayoutId == $layout->getId()){
+                    $ok = true;
+                }
+            }
+            if(!$ok){
+                if (count($validLayouts) > 0) {
+                    $currentLayoutId = $validLayouts[array_key_first($validLayouts)]->getId();
```
weisswurstkanone: maybe use https://www.php.net/manual/en/function.reset.php instead ?

PGBrandi: Looks better now, thanks. Didn't know this function at all.

brusch:
Thanks a lot!

PR #5288, opened 2019-11-19
[Documents] Add required option for editables

### LINKED ISSUE #3672: [New Feature] Define Required editables like input or wysiwyg
If the `<h1>` is empty on a Document i can't publish a Document!
```<h1><?= $this->input('headline', [ 'required' => true ]); ?></h1>```

### PR DESCRIPTION
**Pull request by dvesh3**
Resolves #3672

```{{ pimcore_input("headerLine", {'width': 540, ```**```'required': true```**}```) }}```
## Additional info 
It prompts for missing values on **"Save & Publish"** of Documents.

Supported Types:
- **Input** 
- **WYSIWYG**
- **Textarea**
- **Numeric**

### CODE REVIEW

File: models/Document/PageSnippet.php [hunk 1]
```diff
@@ -607,4 +619,46 @@ public function getUrl($hostname = null, $scheme = null)
 
         return $url;
     }
+
+    /**
+     * @return int|null
+     */
+    public function isContentRequired()
+    {
+        return $this->contentRequired;
```
brusch: The property `contentRequired` isn't persisted to the database, as the database column is called `requireContent` so the auto-mapping doesn't work. So we need to unify the naming everywhere, I'd anyway prefer a name like `hasRequiredEditables` or something similar, which is more self-explaining. 

Please change that for all methods, properties, column names, variables, ... Thanks!

bundles/AdminBundle/Resources/public/js/pimcore/document/edit.js
```diff
@@ -391,11 +391,17 @@ pimcore.document.edit = Class.create({
 
             for (var i = 0; i < editables.length; i++) {
                 try {
-                    if (editables[i].getName() && !editables[i].getInherited()) {
-                        editableName = editables[i].getName();
-                        values[editableName] = {};
-                        values[editableName].data = editables[i].getValue();
-                        values[editableName].type = editables[i].getType();
+                    if(only == "validate") {
```
brusch: We can collect all editables wich are configured as required already on initialization here: 
https://github.com/pimcore/pimcore/blob/98f8f80933856fad19bba41678caa976e9562b86/bundles/AdminBundle/Resources/public/js/pimcore/document/edit/startup.js#L146

I'd say we can just reference them in another array, something like that should be sufficient: 
```js
try {
	let editable = getEditable(editableConfigurations[i]);
	editables.push(editable);
	if(editable.isRequired()) {
		requiredEditables.push(editable)
	}
} catch (e) {
```

bundles/AdminBundle/Resources/public/js/pimcore/document/page_snippet.js
```diff
@@ -445,5 +445,64 @@ pimcore.document.page_snippet = Class.create(pimcore.document.document, {
...
+    },
+
+    save : function ($super, task, only) {
+        if(task !== "publish") {
+            this.validateRequiredEditables(true);
+        }
+        $super(task, only);
+    },
+
+    validateRequiredEditables: function (dismissalert) {
```
brusch: `dismissalert` camel case for better readability plz 😉

bundles/AdminBundle/Resources/public/js/pimcore/document/tag.js
```diff
@@ -126,6 +128,27 @@ pimcore.document.tag = Class.create({
             fieldname: this.name
         }
         return context;
+    },
+
+    validateValue: function(value, el, parent, mark) {
```
brusch: Should we call that method `validateRequiredValue` ?

bundles/AdminBundle/Resources/public/js/pimcore/document/tags/input.js
```diff
@@ -90,13 +94,14 @@ pimcore.document.tags.input = Class.create(pimcore.document.tag, {
...
 
+        // @TODO validator is based on \Zend\Json\Expr and does not work with Twig templates, to be removed in v7.0
         if(options["validator"]) {
             this.element.isValid = options["validator"];
             this.validateElement();
         }
     },
 
-    checkValue: function () {
+    checkValue: function (mark) {
```
brusch: The `mark` argument seems to be useless, as `validateValue()` is called only when `this.required == true` and in this case `mark` is always `true`, so I'd say we can just call `this.validateValue(value, this.element, this, true);` instead. 

Please also change that for the other editables.

models/Document/PageSnippet.php
dvesh3: changed to requireEditableValues and is consistent now with db.

bundles/AdminBundle/Resources/public/js/pimcore/document/tags/input.js
dvesh3: We still need mark argument as on initial load checkValue() is called which also checks the required editables but it should not mark element as required(red highlight class).

brusch: ok, thanks for clarification

models/Document/PageSnippet.php
```diff
@@ -70,6 +71,11 @@ abstract class PageSnippet extends Model\Document
      */
     protected $contentMasterDocumentId;
 
+    /**
+     * @var null|int
+     */
+    protected $requireEditableValues = null;
```
brusch: I'd prefer `hasRequiredEditables` as the name for this property, as it indicates whether this document has at least one editable configured as required or not. 
@weisswurstkanone what do you think?

brusch: In other places we're also using e.g. `validateRequiredEditables` so I guess this would be more consistent.

dvesh3: @brusch that's not true, actually it states that if a document is missing values for any required editables.

brusch: @dvesh3 oh dear, my fault, sorry 😊
But I'm still not entirely happy with the name, what do you think about `missingRequiredEditable` or `hasMissingRequiredEditable`?

bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
```diff
@@ -89,6 +89,10 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
         var saveData = this.getSaveData(only);
 
         if (saveData) {
+            if(this.data.requireEditableValues !== null) {
+                saveData.requireEditableValues = intval(this.data.requireEditableValues);
```
brusch: I think we should work with `bool` here and let the DB adapter/setter do the required casing.

models/Document/PageSnippet.php
```diff
@@ -607,4 +619,51 @@ public function getUrl($hostname = null, $scheme = null)
...
+    {
+        return $this->requireEditableValues;
+    }
+
+    /**
+     * @param int|null $requireEditableValues
+     */
+    public function setRequireEditableValues($requireEditableValues)
+    {
+        $this->requireEditableValues = $requireEditableValues;
```
brusch: Cast to `bool`

models/Document/PageSnippet.php
```diff
@@ -607,4 +619,51 @@ public function getUrl($hostname = null, $scheme = null)
...
+    }
+
+    /**
+     * Validates if there is a missing value for required editable
+     */
+    public function checkRequireEditableValues() {
```
brusch: Should we make that `protected` as it is for internal use only?

bundles/AdminBundle/Resources/public/js/pimcore/document/edit.js
```diff
@@ -404,6 +404,34 @@ pimcore.document.edit = Class.create({
         catch (e2) {
         }
 
+        return values;
+    },
+
+    getRequiredEditableValues: function () {
```
brusch: In my opinion we should call this method `getEmptyRequiredEditables` as this is what this function is actually doing (it is not returning the values of the required editables).

bundles/AdminBundle/Resources/public/js/pimcore/document/edit.js
```diff
@@ -404,6 +404,34 @@ pimcore.document.edit = Class.create({
...
+        var values = {};
```
brusch: I think it's sufficient to use a simple array here for `values` and just return the names of the empty required editables: 

```js
var values = [];
// ...
values.push(editableName); 
// ...
```

models/Document/PageSnippet.php
dvesh3: @brusch I think `missingRequiredEditable` fits better ;)

models/Document/PageSnippet.php
```diff
@@ -607,4 +619,61 @@ public function getUrl($hostname = null, $scheme = null)
...
+                $document = Document::getById($this->getId());
+                if ($document instanceof self) {
+                    // rendering could fail if the controller/action doesn't exist, in this case we can skip the required check
+                    $tagNames = $tagUsageResolver->getUsedTagnames($document);
+                    foreach ($tagNames as $tagName) {
+                        $tag = $document->getElement($tagName);
+                        if ($tag instanceof Tag && in_array($tag->getType(), $allowedTypes)) {
+                            $documentOptions = $tag->getOptions();
+                            if ($tag->isEmpty() && isset($documentOptions['required']) && $documentOptions['required'] == true) {
+                                $this->setMissingRequiredEditable(true);
```
brusch: I think we can skip the check here, right?

dvesh3: yes! done :)

PR #5291, opened 2019-11-20
[Asset] Add option to delete Image Features

### LINKED ISSUE #4728: [Asset] Possibility to delete image features
# Report
After my [intial issue](https://github.com/pimcore/pimcore/issues/4625) (https://github.com/pimcore/pimcore/issues/4625) was fixed, I quickly came to realize that I forgot to mention the feature images in the description.

So, while focal points now work fine, the feature images are still displayed. Could you please remove/hide them when not needed?

### Expected behavior
If all focal points of an image are removed, the image features are removed too. Or at least stay hidden after one toggles the image features, so that they are invisble.

### Actual behavior
Removing all focal points will still show the image features after saving & reloading the asset.

### Steps to reproduce
Pretty much the same steps as in the referenced issue.

dvesh3:
Resolves #4728

## Additional info  
1. Toggle Image Features
2. Save Image


brusch:
I don't see the possibility to delete the existing image features :( 
Good would be an additional button in the UI to delete the existing image features (at the moment just faces) and disables the auto-detection of image features (maybe give it already a generic name e.g. `disableImageFeatureAutoDetection`. 


image

brusch:
Maybe we should start a new PR for that :)

dvesh3:
@brusch steps to disable Image Features from UI(not delete): 1. Toggle Image Features 2.Save Image

Just add to your suggestion to introduce "Delete Image Features", we should switch this button to "Set Image Features" to re-add it.

I think we should just change this PR.

PR #5292, opened 2019-11-20
Load .env variables before custom constants

Blackbam:
... because then we can use the values of .env variables in the custom constants. Or is there any good reason against changing the order?

Please check if there is any good reasons AGAINST this. If there is please reject.


brusch:
Yep, there's a reason ... `\Pimcore\Bootstrap::prepareEnvVariables()` has a dependency on the constants, at least on `PIMCORE_PROJECT_ROOT`.

PR #5293, opened 2019-11-20
[Data objects] Prevent "undefined index" notice

BlackbitNeueMedien:
I stumbled upon multiple notices like "Undefined index" for the context variables. There are a lot of conditions like `if ($context && $context['containerType'] === 'fieldcollection') ...` As $context can be not empty and still not contain the key `containerType` this notice gets emitted. With this PR it is checked whether `$context` really contains the index `containerType`.


weisswurstkanone:
thx!

PR #5297, opened 2019-11-20
A possibility to define environment specific constants

Blackbam:
The number of files to be changed in a deployment should be as small as possible (best case only .env files and the specific config files). As the constants file is loaded before the environment i detected, we should have a possibility to define environment specific constants after PIMCORE_ENVIRONMENT has been loaded.

# Use case
Disabling Amazon S3 on local (defining S3 variables only for staging and prod environment).

### DIFF
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -224,7 +224,7 @@ class Bootstrap
         // load custom constants
         $customConstantsEnvFile = PIMCORE_PROJECT_ROOT . '/app/constants-env.php';
         if (file_exists($customConstantsEnvFile)) {
-            include_once $customConstantsFile;
+            include_once $customConstantsEnvFile;
         }
 
         $resolveConstant = function (string $name, $default, bool $define = true) {
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -220,6 +220,12 @@ class Bootstrap
         }
 
         self::prepareEnvVariables();
+        
+        // load custom constants
+        $customConstantsEnvFile = PIMCORE_PROJECT_ROOT . '/app/constants-env.php';
+        if (file_exists($customConstantsEnvFile)) {
+            include_once $customConstantsFile;
+        }
 
         $resolveConstant = function (string $name, $default, bool $define = true) {
             // return constant if defined


brusch:
You can define the constants via env variables, so i actually don't see any need for this 😐
For your use-case, just add the following to you .env.local: 

```
PIMCORE_ASSET_DIRECTORY=/your/local/path/to/assets
...
...
```

Blackbam:
> You can define the constants via env variables, so i actually don't see any need for this

This is where the problem is coming from:

```
if(PIMCORE_ENVIRONMENT !== "local") {
    ########################### FOR S3 ##############################
    // defined according to: https://pimcore.com/docs/6.x/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/Amazon_AWS_Setup/Amazon_AWS_S3_Setup.html
    $s3BaseUrl = getenv("S3_BASE_URL");
    $s3BucketName = getenv("S3_BUCKET_NAME"); // this needs to be changed to the name of your S3 bucket
    $s3FileWrapperPrefix = "s3://" . $s3BucketName; // do NOT change

    // with this you can individualize the storage path of each entity in pimcore
    // you can of course keep some data locally and some data in a S3 bucket - it's completely up to you
    // please remember that you have to migrate existing contents manually if you have existing contents

    // the following 2 paths need configured public access in your bucket
    define("PIMCORE_ASSET_DIRECTORY", $s3FileWrapperPrefix . "/assets");
    define("PIMCORE_TEMPORARY_DIRECTORY", $s3FileWrapperPrefix . "/tmp");

    // constants for reference in the views
    define("PIMCORE_TRANSFORMED_ASSET_URL", $s3BaseUrl . "/" . $s3BucketName . "/assets");

    // the following paths should be private!
    define("PIMCORE_VERSION_DIRECTORY", $s3FileWrapperPrefix . "/versions");
    define("PIMCORE_RECYCLEBIN_DIRECTORY", $s3FileWrapperPrefix . "/recyclebin");
    define("PIMCORE_LOG_MAIL_PERMANENT", $s3FileWrapperPrefix . "/email");
    define("PIMCORE_LOG_FILEOBJECT_DIRECTORY", $s3FileWrapperPrefix . "/fileobjects");
    ########################### END: FOR S3 ##############################
}
```

Do not want to set those constants on local and I also do not want to search for their correct default value as it may change over time.  What would you suggest?

dpfaffenbauer:
use env variables instead.

Blackbam:
> use env variables instead.

But .env files do not support variables.

If you think a second constants would be unecessary overhead I accept your decision though ;-)

dpfaffenbauer:
But Symfony parameters do with env variables. so define your basic variables with env vars, and the rest with Symfony Parameters

PR #5299, opened 2019-11-20
Fixes the nullable container issue in Pimcore::shutdown() mentod

### LINKED ISSUE #5300: Error: Call to a member function getParameter() on null in .../Pimcore.php on line 275
# Bug Report

### Expected behavior
There is no errors after `Pimcore::shutdown()`

### Actual behavior
```
Error: Call to a member function getParameter() on null in /var/www/vendor/pimcore/pimcore/lib/Pimcore.php on line 275
```

### Steps to reproduce
My case.
There is a conflict between `\Symfony\Bundle\FrameworkBundle\Test\KernelTestCase` and `\Pimcore\Kernel`. When you create a test that extends the `KernelTestCase` it will cause an error after execution:
```
Error: Call to a member function getParameter() on null in /var/www/vendor/pimcore/pimcore/lib/Pimcore.php on line 275
```
This error caused by trying to call the `getParameter()` method on the nullable container because it was deleted by `KernelTestCase` already. [See this link](https://github.com/symfony/symfony/blob/996d8a708fd5fe5043c5098c3da4921f75b95e52/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L134) (doesn't matter whether it's 4.4 or 4.3. I'm using the 4.3 version)

MiragePresent:
Resolves #5300 

## Problem 
```
Error: Call to a member function getParameter() on null in /var/www/vendor/pimcore/pimcore/lib/Pimcore.php on line 275
```

## Reproduce guide
My case.
There is a conflict between `\Symfony\Bundle\FrameworkBundle\Test\KernelTestCase` and `\Pimcore\Kernel`. When you create a test that extends the `KernelTestCase` it will cause an error after execution:
```
Error: Call to a member function getParameter() on null in /var/www/vendor/pimcore/pimcore/lib/Pimcore.php on line 275
```
This error caused by trying to call the `getParameter()` method on the nullable container because it was deleted by `KernelTestCase` already. [See this link](https://github.com/symfony/symfony/blob/996d8a708fd5fe5043c5098c3da4921f75b95e52/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L134) (doesn't matter whether it's 4.4 or 4.3. I'm using the 4.3 version)

### DIFF
--- a/lib/Pimcore.php
+++ b/lib/Pimcore.php
@@ -270,6 +270,10 @@ class Pimcore
         // set inShutdown to true so that the output-buffer knows that he is allowed to send the headers
         self::$inShutdown = true;
 
+        if (self::getContainer() === null) {
+            return;
+        }
+
         // Check if this is a cache warming run and if this runs on an installed instance. If this is a cache warmup
         // we can't use self::isInstalled() as it will refer to the wrong caching dir.
         if (self::getKernel()->getCacheDir() === self::getContainer()->getParameter('kernel.cache_dir') && self::isInstalled()) {


dpfaffenbauer:
you can change the KernelTestCase to load a different kernel, which I advice you todo if you make Kernel Tests with Pimcore. Pimcore internally relies heavily on having the container publicly available. So this wouldn't actually solve your issues...

dpfaffenbauer:
take a look here how to setup PHP Unit tests with Pimcore: https://github.com/coreshop/CoreShop

MiragePresent:
> take a look here how to setup PHP Unit tests with Pimcore: https://github.com/coreshop/CoreShop

@dpfaffenbauer Thank you! I'll take a look at that link.  But as I understand there are other cases when the container is empty [see here](https://github.com/pimcore/pimcore/blob/14ecbb8bfc3fb4ffd67a01a71149246df43b2dd2/lib/Kernel.php#L198):
```
register_shutdown_function(function () {
    // check if container still exists at this point as it could already
    // be cleared (e.g. when running tests which boot multiple containers)
    if (null !== $container = $this->getContainer()) {
        $container->get('event_dispatcher')->dispatch(SystemEvents::SHUTDOWN);
    }
    \Pimcore::shutdown();
});
```

So the additional condition in this PR is not useless

brusch:
This check doesn't hurt, so we can include it. Thanks!

MiragePresent:
@brusch Thank you! I'll help a lot

Issue #4762, opened 2019-07-24
[Documents] Default option for Select Editable


dvesh3:
# Feature Request
Add option to set default value for select editable store https://pimcore.com/docs/5.x/Development_Documentation/Documents/Editables/Select.html

PR #5305, opened 2019-11-21
Update 30_Select.md

roland4432:
Clarified the usage of 'setDataFromResource' and added it to available method calls

  
  

Resolves #


brusch:
I think it would be better to introduce a config option for defining the default value as `setDataFromResource()` is not intended to be used for that.

brusch:
There's already an existing issue for that, see #4762

PR #5310, opened 2019-11-21
Always use PIMCORE_PHP_ERROR_REPORTING and fix PHP notices

### LINKED ISSUE #4708: [Improvement] No notice errors in back-office and Pimcore code + activate E_ALL in debug mode
Hello,

I have seen phpstan is on his way for Pimcore test suite, kool !
On our side, we are trying to enforce code quality, by setting `error_reporting(E_ALL);` for bundles in the `src` directory.

We cannot enforce this setting globally including Pimcore because Pimcore send some notice errors in back-end.
It would be nice a global pass could be done to eliminate all those notice errors.
For instance `$limit` here could not be defined (empty skeleton, just open the BO and check for error 500 in browser dev console): https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L196

image

There is more (i didn't check everywhere... the recycle bin for instance, etc.)

Idea behind this, is to be able to develop on the Pimcore GUI with notice errors being reported.
It will make Pimcore code more robust too.

The final objective is to be able to change this line: https://github.com/pimcore/pimcore/blob/31fd11bcf370ab92b6539b68fc88da53cfec0e12/lib/Pimcore.php#L73 by `error_reporting(E_ALL);`
This way people developping Pimcore extension would produce better code too.

blankse:
Partly resolves #4708

With this PR it is possible to set an PIMCORE_PHP_ERROR_REPORTING env to show all error inclusive notices in dev mode.

**.env.local.php**
```
<?php

return [
    'PIMCORE_PHP_ERROR_REPORTING' => E_ALL,
];

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -138,7 +138,7 @@ env:
     - PIMCORE_TEST_CACHE_REDIS_DATABASE=1
     - SYMFONY_VERSION=^4.0
     - PIMCORE_PROJECT_ROOT=$TRAVIS_BUILD_DIR
-
+    - PIMCORE_PHP_ERROR_REPORTING=32767 # E_ALL
 
 # see https://docs.travis-ci.com/user/job-lifecycle/
 before_install:
--- a/bin/pimcore-install
+++ b/bin/pimcore-install
@@ -21,7 +21,6 @@ use Symfony\Bundle\FrameworkBundle\Console\Application;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Debug\Debug;
 
-error_reporting(E_ALL & ~E_NOTICE & ~E_STRICT);
 @ini_set('display_errors', 'On');
 
 $maxExecutionTime = 0;
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetHelperController.php
@@ -283,10 +283,10 @@ class AssetHelperController extends AdminController
         $sharedConfigs = $classId ? $this->getSharedGridColumnConfigs($this->getAdminUser(), $classId, $searchType) : [];
         $settings = $this->getShareSettings((int)$gridConfigId);
         $settings['gridConfigId'] = (int)$gridConfigId;
-        $settings['gridConfigName'] = $gridConfigName;
-        $settings['gridConfigDescription'] = $gridConfigDescription;
-        $settings['shareGlobally'] = $sharedGlobally;
-        $settings['isShared'] = (!$gridConfigId || $shared) ? true : false;
+        $settings['gridConfigName'] = $gridConfigName ?? null;
+        $settings['gridConfigDescription'] = $gridConfigDescription ?? null;
+        $settings['shareGlobally'] = $sharedGlobally ?? null;
+        $settings['isShared'] = !$gridConfigId || ($shared ?? null);
 
         return [
             'sortinfo' => isset($gridConfig['sortinfo']) ? $gridConfig['sortinfo'] : false,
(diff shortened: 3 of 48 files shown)


brusch:
Nice!
But imho the error reporting level shouldn't be a system setting, I'd prefer configuring it via an environment variable.

dpfaffenbauer:
I'd prefer configuring it via php settings, where they usually are...

brusch:
For v7 yes -> proper BC break! 
But in the meantime I'd say it's better to do that via constant/env variable (like the other stuff).

dpfaffenbauer:
BC break cause Pimcore statically defines the levels?

brusch:
Yes, if your code relies on that ... booom.

dpfaffenbauer:
Just a minor BC then :P

Alright then, we should keep it on the list for Pimcore 7 then.

BTW: Also, when we're gonna change it to an setting, we have to use the current situation as default fallback then as well.

blankse:
There is already a PIMCORE_PHP_ERROR_REPORTING const. This is used only with disabled debug mode. But has a lower level (E_ALL & ~E_NOTICE & ~E_STRICT instead of E_ALL & ~E_NOTICE)

Should we use this for both modi? Or should I create a new one?

brusch:
@blankse in my opinion `PIMCORE_PHP_ERROR_REPORTING` should be used either way.

brusch:
Default should stay on `E_ALL & ~E_NOTICE & ~E_STRICT`. 
Let me know when this PR is ready for review.

brusch:
As far as I can see only one `error_reporting()` in `Boostrap.php` should be sufficient, we should get rid of all the other calls 😊
Thanks!

brusch:
@blankse ready for review? 😊

brusch:
Uh, tests are failing, can you fix them? Thanks!

blankse:
@brusch I am working on it. I'll contact you if the tests are successful :)

blankse:
@brusch I was able to correct all errors in the tests. Also the error in the current master branch (https://github.com/pimcore/pimcore/pull/5310/commits/41e7a38fa6d8b14a288638a5ca2f71cdb5df7d94).

blankse:
Rebase because of https://github.com/pimcore/pimcore/commit/6958beb148070696e3f30f5c615b934e782434b9

weisswurstkanone:
I am getting a undefined index notice in pimcore/lib/Config.php:784 when I try to load the backend UI.
Everything is find if I stay on the master branch.

Seems that this forces me to E_ALL although PIMCORE_PHP_ERROR_REPORTING is not defined.

image

blankse:
@weisswurstkanone Thank you for reviewing. You are right, Symfony Debug set the error reporting level to E_ALL in dev mode. I set the error reporting level argument now.

I get no error in the backend. In pimcore/lib/Config.php:784 should be a closing bracket?

brusch:
Very, very nice, thank you so much!

Yivan:
Howwww so great ! @blankse thanks a lot for having taken this problem.
(it was a big task and i didn't had time to handle it: https://github.com/pimcore/pimcore/issues/4708)
It is one of the best improvements to be able to improve code quality.
Mille merci : )

@brush for subsequent development to come, will the core code be developped in E_ALL on your side (+phpstan level 2 for instance) ? Would be great so we keep this abilities to develop in front and back end with maximum level error reporting.

brusch:
@Yivan wrong username again 😉 (u know `brush` vs `brusch`), it's not about me but about the poor real `brush` who get's the messages instead 😅

But to answer your question: Yes, that would be the plan.

Yivan:
@brusch You are right, sorry !!
Very great if this is the plan so ! Thanks. Have a nice evening.

PR #5314, opened 2019-11-21
[Data object] Copy visible fields when cloning many-to-many object relation field

BlackbitNeueMedien:
To reproduce bug:
1. Create data object class
2. create many-to-many object relation field, allow a class an select some visible fields
3. clone the field
Visible fields have not been copied. This PR fixes that.


weisswurstkanone:
thx!

PR #5324, opened 2019-11-23
Minor readability changes on instalaltion docs

NiklasBr:
The change from `bash` to `text` type for Cron config is because the `text` renders as very dark text on very dark background and is currently unreadable:

Screenshot 2019-11-23 at 18 19 53


brusch:
Thanks a lot! 😊

PR #5330, opened 2019-11-25
[Email] Add option to forward sent emails #5193

### LINKED ISSUE #5193: [Email] Possibility to forward email to different recipient in email log
Additional button in list to forward email to a different recipient, currently it's only possible to resend the email to all original recipients.

dvesh3:
Resolves #5193


bundles/AdminBundle/Resources/public/js/pimcore/settings/email/log.js
```diff
@@ -349,6 +349,37 @@ pimcore.settings.email.log = Class.create({
...
+
+                            Ext.Ajax.request({
+                                url: '/admin/email/show-email-log?id=' + rec.get('id') + '&type=details',
+                                success: function(response){
+                                    var data = Ext.decode( response.responseText );
+                                    var win = this.getForwardEmailWindow(data);
+                                    win.show();
+                                }.bind(this),
+                                failure: function () {
+                                    alert('Could not foward email');
```
weisswurstkanone: alert ?

brusch: @dvesh3 can you change that to a proper ExtJS alert box? Thanks!

dvesh3: @brusch done 😉

bundles/AdminBundle/Resources/public/js/pimcore/settings/email/log.js
```diff
@@ -431,5 +466,113 @@ pimcore.settings.email.log = Class.create({
...
+                    xtype: "form",
+                    bodyStyle: "padding:10px;",
+                    itemId: "form",
+                    items: [
+                        {
+                            xtype: 'hiddenfield',
+                            name: 'id',
+                            value: data.id
+                        },
+                        {
```
brusch: I think we only need one field here, for the email address to forward the email to, all other fields such as subject, from address and reply to should be taken from the original email. "To", CC and Bcc should be clared before sending.

dvesh3: I have kept "subject", "from", "replyTo" as display fields for reference. Only "To" is required to forward the email.

PR #5334, opened 2019-11-25
Fix: Error image path

blankse:
If the image thumbnail could not be generated, the error image have the absolute filesystem path.


brusch:
Thanks a lot! 
I fixed it as well for the other thumbnail types, and used a regex instead which is less error prone.

PR #5335, opened 2019-11-25
Fix: Multi page document thumbnails are not generated

blankse:
The thumbnails are only generated during the update. Should be also generated if they are deleted.


brusch:
Thanks a lot!

PR #5339, opened 2019-11-26
Refactor PageController::checkPrettyUrlAction()

NiklasBr:
You can verify the current behaviour on the error messages by switching to the *SEO & Settings* tab, then clicking in the *Pretty URL* input and then by clicking anywhere else, you will get an error message.

I considered first checking for string length in 1f64da0, but settled for an early exit due to edge cases.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/PageController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PageController.php
@@ -260,17 +260,17 @@ class PageController extends DocumentControllerBase
         // must start with /
         if (strpos($path, '/') !== 0) {
             $success = false;
-            $message .= "\n URL must start with /.";
+            $message = "URL must start with /.";
         }
 
         if (strlen($path) < 2) {
             $success = false;
-            $message .= "\n URL must be at least 2 characters long.";
+            $message .= "<br>URL must be at least 2 characters long.";
         }
 
         if (!Element\Service::isValidPath($path, 'document')) {
             $success = false;
-            $message .= "\n URL is invalid.";
+            $message .= "<br>URL is invalid.";
         }
 
         $list = new Document\Listing();
--- a/bundles/AdminBundle/Controller/Admin/Document/PageController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PageController.php
@@ -256,7 +256,8 @@ class PageController extends DocumentControllerBase
             $message .= "\n URL must start with /.";
         }
 
-        if (strlen($path) < 2) {
+        // Cannot check for length <2, because that would falsely flag empty pretty URL/URL slugs
+        if (strlen($path) === 1) {
             $success = false;
             $message .= "\n URL must be at least 2 characters long.";
         }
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/Document/PageController.php
```diff
@@ -244,12 +244,19 @@ public function displayPreviewImageAction(Request $request)
...
+        $path = (string) trim($request->get('path'));
         $path = rtrim($path, '/');
 
         $success = true;
         $message = null;
 
+        if ($path === '') {
+            return $this->adminJson([
+                'success' => $success,
+                'message' => $message
```
dvesh3: null message?

NiklasBr: I see no side effects in the console, but I can remove it.

dvesh3: got it...this is to bypass errors!

bundles/AdminBundle/Controller/Admin/Document/PageController.php
```diff
@@ -244,26 +244,32 @@ public function displayPreviewImageAction(Request $request)
     public function checkPrettyUrlAction(Request $request)
     {
         $docId = $request->get('id');
-        $path = trim($request->get('path'));
+        $path = (string) trim($request->get('path'));
         $path = rtrim($path, '/');
```
dvesh3: can you move rtrim after empty path check? otherwise path "/" will be valid now.

NiklasBr: Current validation after changes:

Input: `/`,  `//` or more repeating... This is not perfect, might want to add additional error message when the string only contains unwanted characters. But it flags an error at the very least.
Error: `URL must be at least 2 characters long.`

Input: `a`
Error: `URL must start with /.`
`URL must be at least 2 characters long.`

Input: `ab`
Error: `URL must start with /.`

Input: `/ab`
*Success*

dvesh3: It´s just a warning for the user....looks good to me!

bundles/AdminBundle/Controller/Admin/Document/PageController.php
```diff
@@ -254,23 +254,23 @@ public function checkPrettyUrlAction(Request $request)
             ]);
         }
 
-        $message = '';
```
dpfaffenbauer: why not refactor to use symfony validation...

NiklasBr: I did not know that it was possible to validate non-Symfony forms until you asked.

But I think I'm going to let this stand as is because the time/effort it is going to take to do such a refactor.

Good idea for 7.0 though :)

dpfaffenbauer: you can validate basically everything, it is independent from symfony form.

As it is not a BC break, it would work perfectly for 6.* and the time effort wouldn't be too bad.... But, if you don't want to spent the time, I totally understand that.

brusch:
Thanks a lot!

PR #5341, opened 2019-11-27
Open dependency elements via context menu

BlackbitNeueMedien:
Currently UI is not consistent: Sometimes you can open elements by left-click, sometimes by double click, sometimes via context menu -> Open.
This PR adds a contextmenu when you right-click a dependency element, so you can choose `Open`.


brusch:
Nice, thanks a lot! 🙏

PR #5344, opened 2019-11-27
Simplification of the ecommerce tracking for custom product data

andreas-gruenwald:
# Feature Request
Simplify custom product data attribute tracking. 
 
The method ``addProductAttributes`` has been introduced. It
- initializes the most common product data tracking attributes on one site,
- provides a hook to override the default behavior: just call ``item->addAdditionalAttribute('ean', 'test-ean');`` to add your custom attributes.

In addition, each tracking item has an additional ``addAdditionalAttribute`` method. This method can be used to add custom attributes not only to product data, but also to transactions, checkout steps etc.


fashxp:
thx!

PR #5345, opened 2019-11-27
Fix path of error image und thumbnail image with @ signs

blankse:
Fix regression from https://github.com/pimcore/pimcore/commit/50662ce9eebf5b1f09aefff00b3a231d061e22eb


brusch:
Thank you so much!

PR #5346, opened 2019-11-27
Fix GridCellEditor Namespace

solverat:
Use `new pimcore.element.helpers.gridCellEditor` instead of wrong `new pimcore.object.helpers.gridCellEditor` namespace.


brusch:
Thanks a lot!

PR #5349, opened 2019-11-28
While cloning a Select or Multiselect field, all attributes are not saved in cloned field.

neha04:
- While cloning a select field default value and column length are not saved in the cloned field.
- While cloning a multi select field optionsProviderClass and optionsProviderData are not saved in the cloned field


brusch:
Thanks!

PR #5350, opened 2019-11-28
Trim Document Key

### LINKED ISSUE #5317: Route not Found exception for documents ending with space.
- Create a new document ending with space - "test ".
- Open document.

error

route error

neha04:
Resolves #5317

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -940,6 +940,8 @@ class Service extends Model\AbstractModel
         if ($type === 'object') {
             $key = preg_replace('/[<>]/', '-', $key);
         } elseif ($type === 'document') {
+            $key = trim($key);
+            
             // replace URL reserved characters with a hyphen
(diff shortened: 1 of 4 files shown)


BlackbitNeueMedien:
I also stumbled upon this for object keys. These are also not trimmed. Is this on purpose, @weisswurstkanone / @brusch ?

brusch:
@BlackbitNeueMedien No, i guess it would be good to trim them as well, I can't think of a use-case where it wouldn't make sense. 

@neha04 can you update your PR please? Thanks a lot! :)

models/Element/Service.php
```diff
@@ -940,6 +940,8 @@ public static function getValidKey($key, $type)
         if ($type === 'object') {
             $key = preg_replace('/[<>]/', '-', $key);
         } elseif ($type === 'document') {
+            $key = trim($key);
```
dvesh3: just move this before ifelse should be enough and remove trim for else part.

brusch:
Nice, thanks a lot!

PR #5353, opened 2019-11-28
[Web2Print] Preserve type of processing options

podarcis:
For processing options of type bool the value was incorrectly stored as
string. When loading previously stored options, a value "false" (string)
evaluated true and therefore the option was incorrectly checked.

This fix lets ExtJS encode the options as JSON (preserving types) and
the controller action JSON decodes the request body. When storing the
options, all types are preserved.


brusch:
👍 Thanks a lot!

PR #5354, opened 2019-11-28
Support Iterator for all Listings

BlackbitNeueMedien:
This PR harmonizes all descendants of `Pimcore\Model\Listing\AbstractListing`, so all implement the `\Iterator` interface.
Originally I that that was meant with "For all Listing classes it's not necessary anymore to call the load() method before calling the actual getter for the items" in https://pimcore.com/docs/6.x/Development_Documentation/Installation_and_Upgrade/Updating_Pimcore/V5_to_V6.html#page_General-Changes but in the code a lot of Listing classes did not implement `\Iterator` yet.

This PR actually changes some setter return values. If you do not want to harmonize those for backwards compatibility reasons, I will remove the `return`s there to have the `void` return type they had before.


bundles/EcommerceFrameworkBundle/CartManager/Cart/Listing.php
```diff
@@ -23,10 +23,6 @@
  */
 class Listing extends \Pimcore\Model\Listing\AbstractListing
 {
-    /**
-     * @var array
-     */
-    public $carts;
```
dpfaffenbauer: BC break

bundles/EcommerceFrameworkBundle/CartManager/CartCheckoutData/Listing.php
```diff
@@ -23,10 +23,6 @@
  */
 class Listing extends \Pimcore\Model\Listing\AbstractListing
 {
-    /**
-     * @var array
-     */
-    public $cartCheckoutDataItems;
```
dpfaffenbauer: BC break

bundles/EcommerceFrameworkBundle/CartManager/CartItem/Listing.php
```diff
@@ -25,10 +25,6 @@
  */
 class Listing extends \Pimcore\Model\Listing\AbstractListing
 {
-    /**
-     * @var array
-     */
-    public $cartItems;
```
dpfaffenbauer: BC break

lib/Model/Listing/AbstractListing.php
```diff
@@ -445,4 +447,65 @@ public function getConditionVariablesFromSetCondition()
     {
         return $this->conditionVariablesFromSetCondition;
     }
+
+    public function getData(): ?array
```
dpfaffenbauer: BC break, don't set the return value, what if someone else extends from AbstractListing?

models/DataObject/Listing/Concrete.php
```diff
@@ -54,7 +54,7 @@ abstract class Concrete extends Model\DataObject\Listing
      */
     public function __construct()
     {
-        $this->initDao('\\Pimcore\\Model\\DataObject\\Listing\\Concrete');
+        $this->initDao(__CLASS__);
```
dpfaffenbauer: not sure if right, but doesn't ```__CLASS__``` return the actual class also on classes that extend Listing\Concrete?

lib/Model/Listing/AbstractListing.php
BlackbitNeueMedien: Ok, changed this as it is not really necessary.

bundles/EcommerceFrameworkBundle/CartManager/Cart/Listing.php
BlackbitNeueMedien: Yes it is. Perhaps we should leave classes with public item properties as they are. But as most other class variables are protected meanwhile we have to tackle this any time...

models/DataObject/Listing/Concrete.php
BlackbitNeueMedien: No, __CLASS__ always points to the class where this statement is in.

lib/Model/Listing/AbstractListing.php
BlackbitNeueMedien: But as this function did not exist before this PR I actually do not see a BC break.

dpfaffenbauer: but in many other Listings it did

BlackbitNeueMedien: I only found it in https://github.com/pimcore/pimcore/blob/bb4fd386fd09d3e5f4e26041f7aef2eccae0aa7e/models/Notification/Listing.php#L52

But anyway: The return type hint does not add any value imho, so I am fine as it is now (without return type hint)

bundles/EcommerceFrameworkBundle/CartManager/Cart/Listing.php
BlackbitNeueMedien: Resolved

bundles/EcommerceFrameworkBundle/CartManager/CartCheckoutData/Listing.php
BlackbitNeueMedien: Resolved

bundles/EcommerceFrameworkBundle/CartManager/CartItem/Listing.php
BlackbitNeueMedien: Resolved

BlackbitNeueMedien:
Now it works without BC breaks.
And also IDEs recognize the return type automatically via `@method >ModelClass> current()` in all Listing classes.

bundles/EcommerceFrameworkBundle/CartManager/Cart/Listing.php
```diff
@@ -62,6 +61,6 @@ public function getCarts()
      */
     public function setCarts($carts)
     {
-        $this->carts = $carts;
+        return $this->setData($carts);
```
NiklasBr: No longer `@return void`

models/DataObject/QuantityValue/Unit/Listing.php
```diff
@@ -49,18 +55,14 @@ public function isValidOrderKey($key)
...
+        return $this->getData();
     }
 
     /**
      * @param array $units
      */
     public function setUnits($units)
     {
-        $this->units = $units;
+        return $this->setData($units);
```
NiklasBr: Add `@return` in Doc?

models/ImportConfig/Listing.php
```diff
@@ -30,23 +31,24 @@ class Listing extends Model\Listing\AbstractListing
...
+        return $this->getData();
     }
 
     /**
      * @param array $importconfigs
      */
     public function setImportconfigs(array $importconfigs)
     {
-        $this->importconfigs = $importconfigs;
+        return $this->setData($importconfigs);
```
NiklasBr: Add `@return ...` to Doc

BlackbitNeueMedien:
@NiklasBr That i what I meant with "This PR actually changes some setter return values. If you do not want to harmonize those for backwards compatibility reasons, I will remove the returns there to have the void return type they had before."
When the Pimcore maintainers want to syncronize the Listings, I will update the return doctypes. Otherwise I will remove the `return` in the setters which had `void` return type till now. The latter would provide full BC compatibility.

NiklasBr:
@BlackbitNeueMedien What I requested was not either that or the other. I wanted to `@return` declaration *in* the Doc to actually reflect the reality. Whichever you choose to make the reality.

This is not correct:
```php
/**
 * @return void
 */
function exampleFunction() {
      return [];
}
```

This is:
```php
/**
 * @return array
 */
function exampleFunction() {
      return [];
}
```

And it has nothing to do with BC compatibility.

bundles/EcommerceFrameworkBundle/CartManager/Cart/Listing.php
BlackbitNeueMedien: Fixed

models/ImportConfig/Listing.php
BlackbitNeueMedien: Fixed

BlackbitNeueMedien:
Sorry @NiklasBr, you are right. Now return doctypes are corrected. 
But my original statement concerning the decision whether to harmonize all Listing setter methods with returning the Listing object at the cost of a slight BC creak remains. Waiting for a reply of the maintainers. If we want to keep BC, I will remove the `return` from those setter methods which had `@return void` before.

bundles/EcommerceFrameworkBundle/VoucherService/Token/Listing.php
```diff
@@ -20,13 +20,19 @@
...
 /**
  * @method Token[] load()
+ * @method Token current()
  * @method int getTotalCount()
  * @method \Pimcore\Bundle\EcommerceFrameworkBundle\VoucherService\Token\Listing\Dao getDao()
  */
-class Listing extends \Pimcore\Model\Listing\AbstractListing implements \Iterator, AdapterInterface, AdapterAggregateInterface
+class Listing extends \Pimcore\Model\Listing\AbstractListing implements AdapterInterface, AdapterAggregateInterface
 {
     public $tokens;
```
brusch: I guess we should mark those "replaced" properties in the listing classes as `@deprecated` to that we can remove them in v7.

brusch:
@BlackbitNeueMedien All the rest looks good - great work! Just waiting for the feedback to my previous comment - then it's ready to be merged 🤓
@NiklasBr @dpfaffenbauer Thanks for your valuable feedback/review!

bundles/EcommerceFrameworkBundle/VoucherService/Token/Listing.php
BlackbitNeueMedien: Done

BlackbitNeueMedien:
@brusch One last thing to mention: `\Pimcore\Model\Dao\AbstractDao` actually does not define a `load()` method. We also cannot add it because this Dao does also get used for other things as Listings. So there is a slight risk of a BC break if someone extended `AbstractListing` and its Dao class does not have a `load()` method.
The only way to clean this up would be to introduce a new `ListingDao` class which extends `\Pimcore\Model\Dao\AbstractDao` and contains the `load()` method. But this would of course be also a BC break as `AbstractListing::getDao()` then would need to return a `ListingDao` object.

brusch:
@BlackbitNeueMedien I'd value the risk as very low, however we can think about adding a `method_exists()` check before calling `$this->getDao()->load();` - the question is ... is it worth it? 
What do you think?

dpfaffenbauer:
I am pretty sure people extend the `AbstractDao` and `AbstractListing` for custom Pimcore DAO Models. So, I'd go for a `method_exist` in Pimcore 6 and change that to an actual function in Pimcore 7.

BlackbitNeueMedien:
I added the method_exists() call.
But nevertheless I would like to have a structure for new Listing classes. When somebody writes a new Listing class this dependency for a `load()` method of the Dao is nowhere explicitly defined. The mentioned `ListingDao` could define an abstract load method.
But perhaps this is something we should keep in mind for version 7...

dpfaffenbauer:
definitely something for Pimcore 7, everything else would be a BC break.

dpfaffenbauer:
you could trigger a notice when the load function doesn't exist for now, cause it is expected. Something like:

```
@trigger_error(
            'The DAO class expects a load method that currently wasn't declared in the AbstractClass. The method will be added in Pimcore 7,
            E_USER_DEPRECATED
        );
```

BlackbitNeueMedien:
Done. With this deprecation error we will easily find this place when V7 gets prepared ;-)
Now I am fine with this PR.

brusch:
@BlackbitNeueMedien just one more thing, can you please add the following code to `\Pimcore\Model\Listing\Dao\AbstractDao` : 

```php
    /**
     * @TODO abstract method to be activated in Pimcore 7
     * @return array
     */
    /*abstract public function load();*/
```

Then I'm also fine 😉
Thanks again!

PR #5356, opened 2019-11-29
Increase phpstan level to 1 and fix errors

blankse:


dpfaffenbauer:
very nice :)

models/Asset.php
```diff
@@ -948,6 +947,14 @@ public function hasChildren()
...
     }
 
+    /**
+     * @return Asset[]
+     */
+    public function getChildren()
+    {
+        return [];
+    }
+
```
NiklasBr: I suspect you added this method because the Interface requested it. But is it correct to always return an empty array? the Docs says it will return an array of asset objects (but that will never happen now).

What about throwing an error like "An asset cannot have children"?

models/Asset/Document/ImageThumbnail.php
```diff
@@ -82,10 +82,10 @@ public function generate($deferredAllowed = true)
         } elseif (!$this->filesystemPath) {
             $config = $this->getConfig();
             $config->setFilenameSuffix('page-' . $this->page);
+            $path = null;
+            $deferred = ($deferredAllowed && $this->deferred) ? true : false;
```
NiklasBr: It is enough to use `($deferredAllowed && $this->deferred)` only

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -1594,11 +1602,11 @@ public function importGetFileInfoAction(Request $request, ImportService $importS
                 'dataFields' => array_keys($data[0]),
                 'targetFields' => $availableFields,
                 'selectedGridColumns' => $selectedGridColumns,
-                'resolverSettings' => $resolverSettings,
-                'shareSettings' => $shareSettings,
+                'resolverSettings' => $resolverSettings ?? null,
+                'shareSettings' => $shareSettings?? null,
```
NiklasBr: Add one space, very important! ;)

models/Asset.php
blankse: I add this, because the delete method use this method:
https://github.com/pimcore/pimcore/blob/8ae243d75f710701f62e7f4d1e97bb57fd767d38/models/Asset.php#L1016-L1021

The method is only in the Asset/Folder Class. I think we could add the method with an empty array. Why should this not compatible with the Docs? A folder can also have a empty array.

NiklasBr: `@return Asset[]` means an array of `Asset` objects, but this only returns an array, The Doc should say `@return []`

blankse: The return type should be equal to all children classes. The Asset/Folder class returns a array with Assets. In the future we should add type hints, these could not changed with extend.

blankse: If you load a list of assets (folder, images mixed) and call this function. The IDE should know it can only Assets in the array, If we add here array the IDE doesn't no it. That makes sense or not?

NiklasBr: But that code should be on the Listing, no?

I don't even know why this exists, it will never run:

```php
            // remove children
            if ($this->hasChildren()) {
                foreach ($this->getChildren() as $child) {
                    $child->delete(true);
                }
            }
```
because `hasChildren()` [always returns ](https://github.com/pimcore/pimcore/blob/8ae243d75f710701f62e7f4d1e97bb57fd767d38/models/Asset.php#L942-L948)`false` 

My IDE wants to declare it as `array` return type when I try it :)

Skärmbild (254)

blankse: The delete method is also valid for `\Pimcore\Asset\Folder`. There the hasChildren method can return true. Yes it would better to create a abstract class or a interface. Than the IDE wouldn't be show this notice.

dvesh3:
@blankse great work 👏 

Can you please add more fixes to your PR:

isset here:
https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/lib/Config.php#L780
and also here:
https://github.com/pimcore/pimcore/blob/11e8d5f8ec44f22daef3435b9dea8dca6cd728c0/bundles/AdminBundle/Controller/Admin/IndexController.php#L434

default null:
https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/models/Document/Tag/Image.php#L176

define `extras`:
https://github.com/pimcore/pimcore/blob/4c986d60d890898e466533a21dfd29bde2acb9f6/lib/Templating/Helper/HeadLink.php#L451

models/Asset.php
NiklasBr: Oh, I thought `Folder` was `\Pimcore\Model\DataObject\Folder` - now I know there is *also* a `\Pimcore\Asset\Folder` :)

Now I need to understand why there are two different. In other words, keep it as is :)

blankse:
@dvesh3 Thanks for your review. I changed your suggestions.

models/Element/Service.php
```diff
@@ -48,6 +48,7 @@ class Service extends Model\AbstractModel
     public static function getIdPath($element)
     {
         $path = '';
+        $ne = null;
```
dvesh3: meaningful name :)

models/Search/Backend/Data/Listing/Dao.php
```diff
@@ -43,6 +43,7 @@ public function load()
             } elseif ($entryData['maintype'] == 'object') {
                 $element = DataObject::getById($entryData['id']);
             } else {
+                $element = null;
```
dvesh3: just call `\Pimcore\Model\Element\Service::getElementById` and check only for `$entryData['maintype']`

lib/Config.php
```diff
@@ -587,8 +587,8 @@ public static function getPerspectivesConfig()
         if (\Pimcore\Cache\Runtime::isRegistered('pimcore_config_perspectives')) {
             $config = \Pimcore\Cache\Runtime::get('pimcore_config_perspectives');
         } else {
+            $file = self::locateConfigFile('perspectives.php');
```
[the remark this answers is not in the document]
blankse: `$file` is used in catch

lib/Tool/Transliteration.php
```diff
@@ -213,12 +214,7 @@ private static function _transliterationReplace($ord, $unknown = '?', $langcode
                 $base = [];
                 // contains the $base variable
                 include($file);
-                if ($langcode != 'en' && isset($variant[$langcode])) {
```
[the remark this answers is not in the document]
blankse: There can't be a variable `$variant`. So its always unset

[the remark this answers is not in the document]
blankse: Ah it came from the include?

models/Element/Service.php
blankse: Suggestion?

lib/Tool/Transliteration.php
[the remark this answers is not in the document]
blankse: I set the $variant array before include now

models/DataObject/Concrete/Dao/InheritanceHelper.php
```diff
@@ -369,6 +369,7 @@ protected function buildTree($currentParentId, $fields = '', $parentIdGroups = n
         if (!$parentIdGroups) {
             $object = DataObject::getById($currentParentId);
             $query = "SELECT b.o_id AS id $fields, b.o_type AS type, b.o_classId AS classId, b.o_parentId AS parentId, o_path, o_key FROM objects b LEFT JOIN " . $this->storetable . ' a ON b.o_id = a.' . $this->idField . ' WHERE o_path LIKE '.\Pimcore\Db::get()->quote($object->getRealFullPath().'/%') . ' GROUP BY b.o_id ORDER BY LENGTH(o_path) ASC';
+            $queryCacheKey = null;
 
             if (self::$useRuntimeCache) {
                 $queryCacheKey = 'tree_'.md5($query);
```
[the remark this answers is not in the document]
blankse: I set it as default

models/Element/Service.php
blankse: I renamed it to `$parentElement`

models/Search/Backend/Data/Listing/Dao.php
blankse: I used Service::getElementById now

dvesh3:
@blankse I have requested some review changes. Can you please have a look? thanks!

And also there are more findings so if you could also fix these (sorry, I could have pushed, but i think it's better to keep these changes in one place):

1 default null:
https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/lib/Model/Listing/Dao/AbstractDao.php#L38

https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/bundles/AdminBundle/Controller/Admin/Asset/AssetHelperController.php#L347-L348

https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php#L1787

2 isset:
https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/lib/Model/Listing/Dao/AbstractDao.php#L43

https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/bundles/AdminBundle/Controller/Admin/Asset/AssetHelperController.php#L315

https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/bundles/AdminBundle/Controller/Admin/Asset/AssetHelperController.php#L325

https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/bundles/AdminBundle/Helper/GridHelperService.php#L413

https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/bundles/AdminBundle/Helper/GridHelperService.php#L416

https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/bundles/AdminBundle/Helper/GridHelperService.php#L456

https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/bundles/AdminBundle/Helper/GridHelperService.php#L566

https://github.com/pimcore/pimcore/blob/e22c0dcfddab0070a474ce6b04944284a9eda5ca/bundles/AdminBundle/Helper/GridHelperService.php#L569

https://github.com/pimcore/pimcore/blob/c31dc976057f48dd5bb357e1f3f180b5146f5b6d/models/Asset/Service.php#L202

3 ` instanceof Element\AbstractElement` :
https://github.com/pimcore/pimcore/blob/7f9525347ff8969a683c0b37d6c40f2592208967/bundles/AdminBundle/Controller/Admin/ElementController.php#L318

4 default false:
https://github.com/pimcore/pimcore/blob/c62924ca5c3d0c482a71097cb0231aff143208d6/lib/Localization/IntlFormatter.php#L181

5 fix assignment:
https://github.com/pimcore/pimcore/blob/015c27b6e9c3a4a004723b20487fb2e44408a41d/bundles/AdminBundle/Controller/Admin/DataObject/VariantsController.php#L131

models/Search/Backend/Data/Listing/Dao.php
```diff
@@ -36,16 +37,8 @@ public function load()
...
-                $element = Document::getById($entryData['id']);
-            } elseif ($entryData['maintype'] == 'asset') {
-                $element = Asset::getById($entryData['id']);
-            } elseif ($entryData['maintype'] == 'object') {
-                $element = DataObject::getById($entryData['id']);
-            } else {
-                $element = null;
-                Logger::err('unknown maintype ');
-            }
+            $element = Service::getElementById($entryData['maintype'], $entryData['id']);
```
dvesh3: please add `$entryData['maintype']` check and log error if unknown

blankse: I add a check

blankse:
@dvesh3 I changed your suggestions

brusch:
@blankse very, very nice, thank you so much!

Issue #5087, opened 2019-10-14
Cookie notice: Opt-in


BlackbitNeueMedien:
Currently the cookie notice is just a tooltip which shows a message like "This website uses cookies" with an accept button. 
In our opinion this is not enough after current jurisdiction of the European Court of Justice (see https://openjur.de/u/2180608.html - this is only in german) - especially for cookies which are not necessary for the usage of the website (e.g. tracking cookies).
In https://stackoverflow.com/questions/10306254/is-it-possible-to-block-cookies-from-being-set-using-javascript-or-php is shown how to prevent cookies being set. 
What do you think - should Pimcore include an opt-in solution for cookies and prevent cookies from being set if the visitor has not accepted cookie usage?

Perhaps we could extend the cookie notice like in https://github.com/ConnectHolland/cookie-consent-bundle and track which cookies have been tried to be set. In the backend we could then have a list of cookie names which have tried to be set with above's implementation. In this list the admin (or anyone having the to-be-created "cookie" right) could specify to which category certain cookie names belong. This way the visitor really could decide which cookie categories are acceptable to him.

BlackbitNeueMedien:
https://www.webtoffee.com/how-to-automatically-block-cookies-using-the-gdpr-cookie-consent-plugin/ does something similar for Wordpress

Cruiser13:
As far as I know necessary cookies (login, cart, checkout) are still allowed without opt-in.
If other cookie opt-ins are needed, we're currently using https://github.com/AmauriC/tarteaucitron.js which plays nicely with pimcore.

NiklasBr:
> Pimcore should support website owners with a legal cookie consent solution. What do you think?

100% agreement.

msphn:
@brusch I'd think a statement here would be great :)

NiklasBr:
@brusch ping :)

brusch:
First of all, thanks for your inputs on this topic.
We also see the need of a generic solution built-in in Pimcore, but currently it's probably the most worse time to implement a new solution. 

You may ask why?
Well, I'm not a lawyer but, from what I've read, there's a problem with the recent judgement of the European Court of Justice, because it's just a decision for one particular thing, but it doesn't answer all the other open questions where a jurisdiction is missing. Just think of the question which cookies are indispensable and which ones not and there are a lot more things that need clarification. 
[...]

PR #5358, opened 2019-11-29
Cookie notice: Opt-in

fkaefer:
This PR refactors the old CookiePolicyNoticeListener and makes it possible to not only cusomize the template but also the javascript that is used for it. Tracking cookies can now be activated when a user clicks on OK.

- allows custom js in cookie notice
- new default template with an additional button
- two additional keys in shared translations (no, header)

This should be sufficient for #5087 until we get more clarification.


bundles/CoreBundle/EventListener/Frontend/CookiePolicyNoticeListener.php
```diff
@@ -298,13 +309,69 @@ protected function getTranslations($locale)
...
+                'lt' => 'Cookies',
+                'ro' => 'Cookies',
             ]
         ];
 
         $translations = [];
 
         if ($this->getTranslator()) {
-            foreach (['text', 'linkText', 'ok', 'linkTarget'] as $key) {
+            foreach (['text', 'linkText', 'ok', 'linkTarget', 'header', 'no'] as $key) {
```
NiklasBr: Perhaps use `foreach (array_keys($defaultTranslations) as ...` to avoid hard-coding the strings twice?

bundles/CoreBundle/Resources/misc/cookie-policy-default-template.html
```diff
@@ -1,90 +1,119 @@
...
     }
-
 </style>
 
 <div id="pc-cookie-notice">
     <div id="pc-icon"></div>
     <div id="pc-message">
-        <span data-content="%text%"></span>
-        %link%
+        <h1>%header%</h1>
```
NiklasBr: Should this really be a H1 level heading? I'm thinking that search engines might pick it up instead of the primary heading on the page?

bundles/CoreBundle/EventListener/Frontend/CookiePolicyNoticeListener.php
fkaefer: "linkTarget" is not in the previous array, but that could be fixed. But your suggestion gave me a better idea. Instead of hard-coding the strings here we could just have a class variable with an array that contains all the keys needed from shared translations.
So if you need more text bricks for your template you could just use a setter and add all the keys needed.

bundles/CoreBundle/Resources/misc/cookie-policy-default-template.html
fkaefer: Yeah you're right about that. Is h3 in order here?

NiklasBr: H3 or H2 would be fine I think.

While we are improving it, why not add [a little accessibility](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role)?

```html
<div id="pc-cookie-notice" role="dialog" aria-labelledby="pc-heading" aria-describedby="pc-description">
    <div id="pc-icon" aria-hidden="true"></div>
    <div id="pc-message">
        <h2 id="pc-heading">%header%</h2>
        <span data-content="%text%" id="pc-description"></span> %link%
    </div>
    <div id="pc-accept">
        <button data-content="%ok%" aria-label="%ok%"></button>
    </div>
    <div id="pc-decline">
        <button data-content="%no%" aria-label="%no%"></button>
    </div>
</div>
```

bundles/CoreBundle/EventListener/Frontend/CookiePolicyNoticeListener.php
NiklasBr: Missed that one, sounds good! 👍

brusch:
Thanks for your PR! 
I see one issue regarding styles, if someone is using custom styling for the cookie bar, this change will have an impact on that. 
So in my opinion that's not really a good idea to make that change in a minor version.

NiklasBr:
@brusch Now would be a *really* good time to refer to that coming guide for which breaking changes are allowed and which are not. :) 

Is it ready yet?

fkaefer:
If we don't want to have breaking changes we could leave in the part where you can set your own javascript block and translation keys and just keep the old default template.
I think this could really help people to build a law compliant solution.

brusch:
@fkaefer sound like a plan 👍 Can you please update your PR? 
Thanks a lot!

PR #5360, opened 2019-12-02
WYSIWYG: Simplify Regex to check for dependencies with `pimcore_(id|type)` attributes

### LINKED ISSUE #5359: WYSIWYG: Simplify Regex to check for dependencies with `pimcore_(id|type)` attributes
I recently was searching for a way to reference assets in the text of a WYSIWYG field. During my search I found this line of code in: https://github.com/pimcore/pimcore/blob/master/lib/Tool/Text.php#L258

```
preg_match_all("@\<(a|img)[^>]*((?:pimcore_id|pimcore_type)+=\"[0-9]+\")[^>]*((?:pimcore_id|pimcore_type)+=\"[asset|document|object]+\")[^>]*\>@msUi", $text, $matches);
```

Everything is working fine, but I was wondering what this part of the expression is good for: `(?:pimcore_id|pimcore_type)+`

In the end this piece of RegEx allows the following HTML tags to be matched:

- `<a href="/some-file.html" pimcore_type="123" pimcore_id="asset">`
- ``

My question is: Is there a reason for this behavior? Otherwise I propose to simplify this RegEx to just:

```
preg_match_all("@\<(a|img)[^>]*(pimcore_id=\"[0-9]+\")[^>]*(pimcore_type=\"[asset|document|object]+\")[^>]*\>@msUi", $text, $matches);
```

manueldublanc:
My proposal as PR.

Resolves #5359

### DIFF
--- a/lib/Tool/Text.php
+++ b/lib/Tool/Text.php
@@ -255,7 +255,7 @@ class Text
         }
 
         //$text = Pimcore_Tool_Text::removeLineBreaks($text);
-        preg_match_all("@\<(a|img)[^>]*((?:pimcore_id|pimcore_type)+=\"[0-9]+\")[^>]*((?:pimcore_id|pimcore_type)+=\"[asset|document|object]+\")[^>]*\>@msUi", $text, $matches);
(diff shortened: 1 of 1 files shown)


brusch:
Thanks! 
I also don't see any reason for keeping  the expression `(?:pimcore_id|pimcore_type)+`

PR #5362, opened 2019-12-02
AbstractElement::isAllowed() doesn't work in commands

blankse:
In cli mode the isAllowed function throws an error
```
Error thrown while running command
In AbstractElement.php line 203:
Call to a member function isAdmin() on null
```

I think its ok, when it returns true in cli


brusch:
Thanks for your PR! 
I've changed it a bit, so that it works in tests etc. as well.

PR #5363, opened 2019-12-02
Add indexedValues to the variables passed to view

mc-supply:
indexedValues was not added to the ElasticSearch version of the SelectCategory.php, which results in an error if using ElasticSearch instead of default MysqlSearch.

  
  

Resolves #


fashxp:
thx

PR #5367, opened 2019-12-03
[Data Objects] Relational Fields with Image Previews from Advanced Images are not automatically adjusting the grids height

rquarta-extera:
This PR resolves the same problem discussed [here](https://github.com/pimcore/pimcore/pull/4337) but for hotspot images.

To reproduce:

1. Create class **A** and add an "Image Advanced" field
2. Create another class **B** and add an "Advanced Many to Many Object Relation" field, allow the first class and specify the image field as visible.
3. Create an object "A" and assign an image
4. Create another object "B" and assign "A" to the advanced many-to-many object relation. You will see something like this:
before-pr
5. After switching panel and get back to Edit panel:
after-pr

The solution uses the same timeout (and the same idea, too) used by @BlackbitNeueMedien .


fashxp:
thx for your PR. unfortunately it didn't work for me, and the current solution had some issues anyway. 
tried to fix them in #5378 ... could you please verify? thx

PR #5370, opened 2019-12-03
[Translations] Bugfix: "undefined" translation for new language

BlackbitNeueMedien:
Steps to reproduce bug:
1. Add language via system settings > localization
2. Go to admin translations
3. All translations for the newly added language show "undefined"

Fortunately this is only a display error: When you access a translation for the new language it is handled as empty translation.


BlackbitNeueMedien:
Sorry, my local master was not clean

PR #5372, opened 2019-12-03
Fix: Missing PHPDoc return tags

blankse:


fashxp:
thank you so much mate!

PR #5375, opened 2019-12-04
[Ecommerce][PricingManager] Rules should not not be loaded multiple times if no active rules exist

markus-moser:
Empty is not a valid check as an empty array is empty too.


fashxp:
thx!

Issue #4909, opened 2019-08-30
[Data Objects] Image Previews in Grid-View bouncing


skoch872:
Hello,

I found an error while displaying the images in the grid view.

I've tried to fix the bug but I'm unsure what to change without breaking anything. 

### Expected behavior
Like before, a grid view that doesn't flicker :-)

### Actual behavior
The grid view flickers as soon as you want it to output images.

I've tested the behavior in the demo and it appears there, too. Important: This only appears with the data type Image. So not with Image Advanced.

And currently every 500ms the thumbnail is requested from the server.

On Chrome (Build:  76.0.3809):
![flicker_grid_view](https://user-images.githubusercontent.com/54348050/63999185-2c4b8d00-cb03-11e9-96d5-ec8a92f08297.gif)

And with the developer tools:
![flicker_grid_view_tools](https://user-images.githubusercontent.com/54348050/63999189-2f467d80-cb03-11e9-99a8-513e01a11851.gif)

### Steps to reproduce
In the demo I used the class "blogArticle" and converted the "posterImage" to "Image".

Then I added new images to the objects below "/blog/articles" and opened the grid view of the folder.

I found this change using the browser's developer tools, which could trigger it:
d430016fe574d31e807c0c8b53cff1758387a912

PR #5378, opened 2019-12-04
[Data Objects] fixed grid row heights for images in (advanced)many-to-many-object relation fields

fashxp:
fixed #5367
considered #4909 and #4337


rquarta-extera:
Hi @fashxp, your solution is working for me, thanks!

PR #5381, opened 2019-12-04
Fix: Cannot instantiate abstract class Pimcore\Model\User\AbstractUser

blankse:
Regression from #5356 

The error occurs when trying to save a user


fashxp:
thx

PR #5382, opened 2019-12-04
Fix: Wrong PHPDocs in Pimcore\Model\Notification

blankse:


models/Notification.php
```diff
@@ -32,12 +32,12 @@ class Notification extends AbstractModel
     protected $id;
 
     /**
-     * @var int
+     * @var string
```
fashxp: then we would have to change getter/setter too, and that would be a BC

blankse: Getter and Setter have type hints with return string. That's why my phpstorm marks me that something is wrong here

fashxp: ahh sorry ... thought change was to int . all good.

PR #5390, opened 2019-12-05
Support localized calculated fields in object bricks

### LINKED ISSUE #5337: [Calculated value] Context for localized calculated field in object brick
When you have a *localized* calculated field in an object brick there is a conflict with the context. As https://pimcore.com/docs/6.x/Development_Documentation/Objects/Object_Classes/Data_Types/Calculated_Value_Type.html states about context:
* Localized fields:
  * position | the language ("en", "de", ...)
  * ownerType | "localizedfield"
  * ownerName | the name of the localized field ("localizedfields")
* Objectbricks
  * ownerType | "objectbrick"
  * ownerName | the name of the objectbrick field inside the object
  * fieldName | the name of the attribute inside the brick
  * index | the name of the brick
  * keyDefinition | the calculated-value field definition

For object bricks also field `position` gets set but to the same value as `fieldName`:
https://github.com/pimcore/pimcore/blob/22882bee401ac3c411a18deb36a3198e942ccc41/models/DataObject/ClassDefinition/Data/Objectbricks.php#L179-L180

When you want to determine the language for the value generation of the calculated field, we need to set `position` to the language here. This currently does not work because the object brick fetches all descendant calculated value fields, not only children - bypassing the localized value field which would add the language as field `position`:
https://github.com/pimcore/pimcore/blob/22882bee401ac3c411a18deb36a3198e942ccc41/models/DataObject/ClassDefinition/Data/Objectbricks.php#L175

BlackbitNeueMedien:
With this PR it is possible to have localized calculated value fields in object bricks.

Resolves #5337


weisswurstkanone:
could you resolve the conflicts please ? thx

models/DataObject/ClassDefinition/Data/CalculatedValue.php
```diff
@@ -338,12 +344,32 @@ public function getGetterCodeLocalizedfields($class)
         $code .= "\t\t" . '}' . "\n";
         $code .= "\t" . '}'  . "\n";
 
+        if($class instanceof DataObject\Objectbrick\Definition) {
+            $ownerType = 'objectbrick';
+            $ownerName = $class->getClassDefinitions()[0]['fieldname']; // todo this is not really correct. Actually we need to find out the name of the brick container (an instance of \Pimcore\Model\DataObject\Objectbrick). This must happen during runtime as one brick can be used in multiple brick containers.
```
weisswurstkanone: what is the right interpretation of the comment ?

BlackbitNeueMedien: This means that we cannot determine the name of the class with the brick container field when the brick gets saved. Actually we need to add some code in the generated getter function to determine the brick container field during runtime but I did not find a way to do that because the generated brick class does not know anything in which brick container / data object class it is being used.
And because of that I chose a way to at least enhance the status quo (although this solution is not perfect) by setting the `$ownerName` to the first configured brick container field of the object brick.

weisswurstkanone: Hi, just noticed something. 

https://github.com/pimcore/pimcore/blob/2598b5dc343e06f5100cc8ce8d90c33c6301baf8/models/DataObject/ClassDefinition/Data/Localizedfields.php#L142

The calculator only gets called if there are other fields with data in the localized field container?

What do you think?

weisswurstkanone: Why not just take it from $this ?

image

BlackbitNeueMedien: Is `fieldname` the calculated value field in the brick or the brick container field in the data object class? I need the latter.

weisswurstkanone: the one in the data object class

BlackbitNeueMedien: Oh, then have overseen this. Thanks, I will change it.

BlackbitNeueMedien: Done. My test was positive.

weisswurstkanone: Can you also check this "I noticed something" thing ?

BlackbitNeueMedien:
```
The calculator only gets called if there are other fields with data in the localized field container?

What do you think?
```
Could you give this a try? The customer which I needed this for has Pimcore 5.8 - there it worked: I have only calculated value fields in the localizedfields container in the brick.

I currently have not the time to set up a Pimcore 6 with a localozed calculated value in a brick. Could you please give it a try, @weisswurstkanone - if it does not work I will revert the last commit.

weisswurstkanone:
It bombs out here because there is no container, $context is empty

image

BlackbitNeueMedien:
Now it is the same call as in Localizedfield::getLazyLoadedFieldNames(). Better?

weisswurstkanone:
looks OK from my side

weisswurstkanone:
thanks a lot!

PR #5395, opened 2019-12-06
Add tag event listeners

simon-dcode:
I've added A TagEvents class and a TagEvent Event Model class. 
In the TagEvents class I declared `PRE_ADD_TO_ELEMENT`, `POST_ADD_TO_ELEMENT`, `PRE_REMOVE_FROM_ELEMENT`, `POST_REMOVE_FROM_ELEMENT` events.

Also I added event dispatchers that fire the above events before and after both tag addition and removal in the Admin's TagsController class.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/TagsController.php
+++ b/bundles/AdminBundle/Controller/Admin/TagsController.php
@@ -17,6 +17,8 @@ namespace Pimcore\Bundle\AdminBundle\Controller\Admin;
 use Pimcore\Bundle\AdminBundle\Controller\AdminController;
 use Pimcore\Event\AdminEvents;
 use Pimcore\Model\Element\Tag;
+use Pimcore\Event\TagEvents;
+use Pimcore\Event\Model\TagEvent;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\GenericEvent;
 use Symfony\Component\HttpFoundation\JsonResponse;
@@ -208,8 +210,10 @@ class TagsController extends AdminController
 
         $tag = Tag::getById($tagId);
         if ($tag) {
+            $event = new TagEvent($tag, array('assignmentElementId' => $assginmentCId, 'assignmentElementType' => $assginmentCType));
+            \Pimcore::getEventDispatcher()->dispatch(TagEvents::PRE_ADD_TO_ELEMENT, $event);
             Tag::addTagToElement($assginmentCType, $assginmentCId, $tag);
-
+            \Pimcore::getEventDispatcher()->dispatch(TagEvents::POST_ADD_TO_ELEMENT, $event);
             return $this->adminJson(['success' => true, 'id' => $tag->getId()]);
         } else {
             return $this->adminJson(['success' => false]);
@@ -231,8 +235,10 @@ class TagsController extends AdminController
 
         $tag = Tag::getById($tagId);
         if ($tag) {
+            $event = new TagEvent($tag, array('assignmentElementId' => $assginmentCId, 'assignmentElementType' => $assginmentCType));
+            \Pimcore::getEventDispatcher()->dispatch(TagEvents::PRE_REMOVE_FROM_ELEMENT, $event);
             Tag::removeTagFromElement($assginmentCType, $assginmentCId, $tag);
-
+            \Pimcore::getEventDispatcher()->dispatch(TagEvents::POST_REMOVE_FROM_ELEMENT, $event);
             return $this->adminJson(['success' => true, 'id' => $tag->getId()]);
         } else {
             return $this->adminJson(['success' => false]);
--- a/lib/Event/Model/TagEvent.php
+++ b/lib/Event/Model/TagEvent.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * Pimcore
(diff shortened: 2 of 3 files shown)


lib/Event/Model/TagEvent.php
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @return Tag
+     */
+    public function getElement()
+    {
+        return $this->getTag();
+    }
+
+    /**
+     * @return Array
```
NiklasBr: `@return Array` should be `@return array` because PHP interprets `Array` as a class name and IDE:s will report *Undefined class 'Array'*

simon-dcode: Thank you for your comment. I fixed it!

brusch:
@simon-dcode very nice! 
Can you please recreate the PR for the master branch (currently based on 5.8)? 
Thanks a lot!

bundles/AdminBundle/Controller/Admin/TagsController.php
```diff
@@ -208,8 +210,10 @@ public function addTagToElementAction(Request $request)
 
         $tag = Tag::getById($tagId);
         if ($tag) {
+            $event = new TagEvent($tag, array('assignmentElementId' => $assginmentCId, 'assignmentElementType' => $assginmentCType));
+            \Pimcore::getEventDispatcher()->dispatch(TagEvents::PRE_ADD_TO_ELEMENT, $event);
```
brusch: @simon-dcode I guess we should move the event dispatching into `Tag::addTagToElement()` so that it is also dispatched when used in custom scripts.

brusch:
@simon-dcode Additionally we should also add all the other events we know from the other entities (pre/post save/delete)... e.g. see https://github.com/pimcore/pimcore/blob/bf4799782bb6d10ebe2adab85d9eb555ce0b84f9/lib/Event/RedirectEvents.php#L17

simon-dcode:
> @simon-dcode very nice!
> Can you please recreate the PR for the master branch (currently based on 5.8)?
> Thanks a lot!

I'll create a PR for the master branch after this is done and accepted. I'd like to have these functionalities in both 5.8.x and 6.x if possible.

brusch:
@simon-dcode please create the PR for the master branch first, incl. the changes requested below. Afterwards please backport the changes to the 5.8 branch.

simon-dcode:
> @simon-dcode please create the PR for the master branch first, incl. the changes requested above. Afterwards please backport the changes to the 5.8 branch.

All right, no problem. However I'm quite busy right now, maybe I can do these on the next week.

brusch:
@simon-dcode sure, no problem😉

bundles/AdminBundle/Controller/Admin/TagsController.php
simon-dcode: If I dispatch this in `Tag::addTagToElement()`, how will I be able to differentiate between Tag assignments made on the Admin and made in the code somewhere else. 
Do you think I should create separated events for these situations then?

brusch: Why should you differentiate between admin or API? 
This is also not done for other events as far as I know. 
One possibility would be to pass additional arguments to the event object, like `source`, which can be optionally set.

simon-dcode: It would be pointless to explain how my system works. The point is, I need to differentiate these two. But since the TagEvent object has the arguments, I'll handle this difference with that. Thanks for the suggestion.
The question is, can I send in the PR with this differentiating or is this 'too custom'?

brusch: Good point, I've just checked if we have done this already in the past and unfortunately the answer is no, so I guess it would be inconsistent to do it now with tags 😟

So I guess it needs a workaround for your case. 
You can try the following in the event handler: 
```php
        $masterRequest = $requestStack->getMasterRequest();
        if($masterRequest) {
            $route = $masterRequest->get('_route');
            if($route === 'pimcore_admin_tags_addtagtoelement') {
                // do whatever you want
            }
        }
```

Not nice, but works :D

dpfaffenbauer: you can also check for the context... if context is pimcore admin, you know the request comes from pimcore backend

simon-dcode: Could you give me a code example of what you're thinking of?

dpfaffenbauer: https://github.com/coreshop/CoreShop/blob/b3f85464e9f9e77728ae5e4e704844b361bb3c8b/src/CoreShop/Component/Core/Locale/Context/PimcoreAdminLocaleContext.php#L75

dpfaffenbauer: You don't need to pass a request object, the ContextResolver fetches the current request if null

simon-dcode: Thank you!

brusch:
Closed due to inactivity (and because this PR is anyway for the wrong branch). 
@simon-dcode Please open another PR for the `master` branch as soon as you have time :) Thanks!

PR #5400, opened 2019-12-09
Remove duplicate indexes

BlackbitNeueMedien:
There is no advantage in single-column database indexes when there also exists a composite index which has the same column(s) as prefix. For example index on column A does not make any sense if you additionally have an index on columns A,B. - see for example https://stackoverflow.com/questions/51046457/find-duplicate-index-in-mysql-using-query#answer-51048636

As example the query execution plan for the following query stays the same - with or without the removed index:
```mysql
explain select * from objects where o_path='abc'
```
Including the index path(o_path):

1 | SIMPLE | objects | NULL | ref | fullpath,path | fullpath | 2298 | const | 1 | 100.00 | NULL

Removed index path(o_path)

1 | SIMPLE | objects | NULL | ref | fullpath | fullpath | 2298 | const | 1 | 100.00 | NULL

This PR removes all those redundant / duplicate indexes to improve write performance.

### DIFF
--- a/bundles/CoreBundle/Resources/migrations/Version20191208175348.php
+++ b/bundles/CoreBundle/Resources/migrations/Version20191208175348.php
@@ -3,8 +3,12 @@
 namespace Pimcore\Bundle\CoreBundle\Migrations;
 
 use Doctrine\DBAL\Schema\Schema;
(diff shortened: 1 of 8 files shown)


brusch:
Thanks a lot!

PR #5409, opened 2019-12-10
Attempt to close stream in getTemporaryFile()

### LINKED ISSUE #5402: Stream not closed when creating temporary file
# Bug Report

### Description
When public method Pimcore\Model\Asset::getTemporaryFile() is called without having a source stream, a new one is created. But it is not closed until descruct is called. If there is a long running job there is a lot of open files. 

The issue was found is when using imagick adapter with external source storage (stream is not local).
```php
// In Pimcore\Image\Adapter\Imagick
if (!stream_is_local($imagePath) && isset($options['asset'])) {
	// imagick is only able to deal with local files
	// if your're using custom stream wrappers this wouldn't work, so we create a temp. local copy
	$imagePath = $options['asset']->getTemporaryFile();
	$this->tmpFiles[] = $imagePath;
}
```

### Expected behavior
Stream is closed properly when creating temporary file.

### Actual behavior
Stream is not closed properly.

### Steps to reproduce
Load a list of assets, create temporary file for each asset. Between each loop, check open files with for the process for example by using lsof -p.

### Code example
My suggestion is to always close the source stream after creating the temporary file.

```php
/**
 * returns the path to a temp file
 *
 * @return string
 */
public function getTemporaryFile()
{
	$destinationPath = PIMCORE_SYSTEM_TEMP_DIRECTORY . '/asset-temporary/asset_' . $this->getId() . '_' . md5(microtime()) . '__' . $this->getFilename();
	if (!is_dir(dirname($destinationPath))) {
		File::mkdir(dirname($destinationPath));
	}

	$src = $this->getStream();
	$dest = fopen($destinationPath, 'w+', false, File::getContext());
	stream_copy_to_stream($src, $dest);
	fclose($dest);
	$this->closeStream(); // ADDED

	@chmod($destinationPath, File::getDefaultMode());

	$this->_temporaryFiles[] = $destinationPath;

	return $destinationPath;
}
```

NiklasBr:
Resolves #5402

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1525,6 +1525,7 @@ class Asset extends Element\AbstractElement
         $dest = fopen($destinationPath, 'w+', false, File::getContext());
         stream_copy_to_stream($src, $dest);
         fclose($dest);
+        $this->closeStream();
 
         @chmod($destinationPath, File::getDefaultMode());
 


brusch:
@NiklasBr Thanks for you PR! Unfortunately this seems to cause some issues -> see Travis. 
Can you please check that? 
Thanks a lot!

NiklasBr:
Do you have any comments or documentation on the unit tests? Because there are none I can't be sure where to change them.

brusch:
@NiklasBr We're currently working on the docs for the core tests. 
I think the root cause why the tests are failing can be found somewhere around here: 
https://travis-ci.org/pimcore/pimcore/jobs/623129275#L4459 
It seems that the type detection for assets is behaving a bit differently due to this change. 

Regarding the error suppression: As far as i can remember this is there because of remote filesystems, because there the stream can be closed from the remote end, but `is_resource()` is still returning true, while `fclose()` will fail.

weisswurstkanone:
@NiklasBr Should be quite easy to reproduce, try the following (I believe it doesn't happen for asset type unknown). Without your change dump output will be "123", with your change ""

        $asset = new Asset\Image();
        $asset->setParentId(1);
        $asset->setFilename(uniqid());
        $asset->setData("123");
        $asset->save();
        $reloaded = Asset::getById($asset->getId());
        var_dump($reloaded->getFullPath());
        var_dump($reloaded->getData());

brusch:
@NiklasBr any update on this? Thanks! 😊

NiklasBr:
Nothing. I still need to work through all the BC breaks between 6.1 and 6.3 (now 6.4) before I can start this again.

PR #5414, opened 2019-12-10
BREACH Attack fix for login and lost password forms

brusch:


lib/Templating/Helper/BreachAttackRandomContent.php
```diff
@@ -0,0 +1,45 @@
...
+        $length = 50;
+        $randomData = random_bytes($length);
+
+        return '<!--'
+            . substr(
+                base64_encode($randomData),
+                0,
+                ord($randomData[$length-1]) % 32
+            )
+            . '-->';
```
NiklasBr: This is a mitigation by increasing the number of attempts required, not the final fix, right?

Is there a long term plan to implement more complete protection?

brusch: This is just an additional protection, the proper fix is to renew the CSRF token on every request.

brusch: Done here: https://github.com/pimcore/pimcore/pull/5414/files#diff-85fefd7ecb60f6f6c28b38687d90c2aeR77

PR #5423, opened 2019-12-11
[Recycle Bin] Fix restoring relations in Localized Fields, Object Bricks and Field Collections

### LINKED ISSUE #5255: [Recycle Bin] Localized Fields, Object Bricks and Field Collections are not getting restored
If single item restore, works for nested restores

dvesh3:
Resolves #5255

### DIFF
--- a/models/Element/Recyclebin/Item.php
+++ b/models/Element/Recyclebin/Item.php
@@ -17,6 +17,7 @@
 
 namespace Pimcore\Model\Element\Recyclebin;
 
+use DeepCopy\DeepCopy;
 use Pimcore\Cache;
 use Pimcore\File;
 use Pimcore\Logger;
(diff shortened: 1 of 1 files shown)


brusch:
Well, so your suggestion is to just keep the bug? Alright, I understand 😉

brusch:
Let me clarify one thing here: we are not responsible to check any of our changes against compatibility with Doctrine entities, so this is actually not a BC break, it's just a bug fix nothing else.

dpfaffenbauer:
Changing behavior in a patch release is the point here. That change was not a bug, so it does not have to be in a patch release. Doing that with Pimcore 6.5 wouldn’t be an issue. 

It is just too hard to maintain Plugins for Pimcore with the way you keep changing and breaking stuff.

dpfaffenbauer:
BTW: you also recommend doctrine for custom entities in your docs: https://pimcore.com/docs/6.x/Development_Documentation/Extending_Pimcore/Custom_Persistent_Models.html

So it should also work in combination with Pimcore Elements.

PR #5432, opened 2019-12-12
Update WYSIWYG docu

Gerhard-Kanzler:
getData -> is the RAW string
frontend -> here you get the parsed version


brusch:
Thanks a lot!

PR #5433, opened 2019-12-12
[Grid operators] PHPCode: Catch Exception if no class has been set

BlackbitNeueMedien:
There is an uncaught exception if you 
1. go to a grid view
1. click "grid options"
1. add operator "Others > Operator PHP Code"


brusch:
Thanks a lot!

PR #5434, opened 2019-12-12
[Migration] Fix compatibility with MySQL

BlackbitNeueMedien:
`CREATE INDEX IF NOT EXISTS <index_name> on <table> (<column>);');` is not valid in MySQL.
Results in `SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IF NOT EXISTS ecommerceframework_cart_user  
  id_index on ecommerceframework_cart (u' at line 1`


brusch:
Grazie Mille! 🙏

PR #5436, opened 2019-12-12
Fix PHP Notice: Undefined index: REMOTE_ADDR (command line issue)

### LINKED ISSUE #5435: PHP Notice:  Undefined index: REMOTE_ADDR (command line issue)
# Bug Report

### Steps to reproduce

execute a command via the command line, like:

php bin/console cache:clear

### Expected behavior

No PHP warning.

### Actual behavior

PHP Notice:  Undefined index: REMOTE_ADDR in /web/web11715/vendor/pimcore/pimcore/lib/Tool.php on line 495

It's no surprise because $_SERVER['REMOTE_ADDR'] is undefined when running PHP from the command line. It's only a warning and everything seems to work. But maybe it's still something you want to fix.

blankse:
Resolve #5435

### DIFF
--- a/lib/Tool.php
+++ b/lib/Tool.php
@@ -477,7 +477,7 @@ class Tool
      *
      * @param Request|null $request
      *
-     * @return string
+     * @return string|null
      */
     public static function getClientIp(Request $request = null)
     {
@@ -491,8 +491,10 @@ class Tool
             $ip = $_SERVER['HTTP_CLIENT_IP'];
         } elseif (!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) {
             $ip = $_SERVER['HTTP_X_FORWARDED_FOR'];
-        } else {
+        } elseif (!empty($_SERVER['REMOTE_ADDR'])) {
             $ip = $_SERVER['REMOTE_ADDR'];
+        } else {
+            return null;
         }
 
         $ips = explode(',', $ip);


dpfaffenbauer:
what re the consequences of returning `null` here?

blankse:
`$request->getClientIp()` in line 486 can also return null. So it shouldn't have consquences.

brusch:
Thanks!

blankse:
@dpfaffenbauer I took another look. The error only occurs when debug with IP is set. Here is already checked for null
https://github.com/pimcore/pimcore/blob/69d492abc8835604382a6deaad9ac376d6acd73e/lib/Config.php#L1132-L1138

ssiergl:
@brusch you should notice that this "bug" broke any migrations on `composer update`. I just upgraded one instance from `6.2.1` to `6.3.6` and got the output:

`Skipping migrations ... (either Pimcore is not installed yet or current status of migrations is not available)`

Reason (executeMigrationsUp method in Composer.php):

```
        // execute migrations
        $currentVersion = null;
        try {
            $process = static::executeCommand($event, $consoleDir,
                'pimcore:migrations:status -s pimcore_core -o current_version', 30, false);
            $currentVersion = trim($process->getOutput());
        } catch (\Throwable $e) {
            // noting to do
        }

        if (!empty($currentVersion) && is_numeric($currentVersion)) {
            self::clearDataCache($event, $consoleDir);
            static::executeCommand($event, $consoleDir, 'pimcore:migrations:migrate -s pimcore_core -n');
            self::clearDataCache($event, $consoleDir);
        } else {
            $event->getIO()->write('<comment>Skipping migrations ... (either Pimcore is not installed yet or current status of migrations is not available)</comment>', true);
        }
```

Due of the php notice the cli output isn't the current Version any more and migrations will not be executed.

PR #5439, opened 2019-12-12
Save applied pricing Rule ID in the order / OrderPriceModification Fieldcollection

ckogler-elements:
Save the applied pricing rule id, because it is often required for fürther order processing (ERP APIs....)


fashxp:
Very nice! 

Should we try to make a migration similar to this one? 
https://github.com/pimcore/pimcore/blob/a296bfebae97628113e0f71eab2240b0cc50303f/bundles/CoreBundle/Resources/migrations/Version20190806160450.php

ckogler-elements:
Good idea - added the migration and adopted the code

fashxp:
thx!

PR #5440, opened 2019-12-12
Fix: Undefined index notices

blankse:
ClassController:
If you open an object with field collections in a group

DataObjectController:
If you rename the key of an object


brusch:
Thanks a lot!  🤘

PR #5444, opened 2019-12-13
Fix undefined index in Dashboard.php

blankse:
Try to open the welcome dashboard with a new user


brusch:
Thank you & have a nice weekend!

PR #5449, opened 2019-12-13
Fix: Argument 1 passed to AbstractListing::setData() must be of the type array, null given

blankse:
I have this error after upgrade the master branch. Regression from #5354

I have a wrapper class around the listing, which set the data to null if the the limit/offset is changed. Otherwise it is not possible to geht a new offset.

Now I can't set it to null...


brusch:
I changed it to nullable  `setData(?array $data)` instead of using a default value.

PR #5453, opened 2019-12-16
[Ecommerce][Productindex] Errorhandling

jremmurd:
This PR 

- **fixes** the namespace of the error event - otherwise the `->dispatch()` of the error event will again throw an error

- **changes** the index preparation exception handling to also catch Errors not only Exceptions (i.e.:
`call to a member function XXX on null in ...`)
- **changes** adds a setter to the event whether to throw an exception or not


fashxp:
is that finished now?

jremmurd:
> is that finished now?

Yess :)

fashxp:
thx!

Issue #4916, opened 2019-08-30
[ecommerceframework][elasticsearch] price sorting does not work with elasticsearch 6


markus-moser:
code in `loadWithoutPriceFilterWithPriceSorting()` needs to look something like this (at least when elasticsearch 6 is used):

```php $params = $this->getQuery();
        unset($params['body']['sort']);     // don't send the sort parameter, because it doesn't exist with offline sorting
        $params['body']['size'] = 10000;    // won't work with more than 10000 items in the result (elasticsearch limit)
        $params['body']['from'] = 0;
        $params['body']['_source'] = ['system.priceSystemName'];

        $result = $this->sendRequest($params);
        $objectRaws = [];
        if ($result['hits']) {
            $this->totalCount = $result['hits']['total'];
            foreach ($result['hits']['hits'] as $hit) {
                $objectRaws[] = ['id' => $hit['_id'], 'priceSystemName' => $hit['_source']['system']['priceSystemName']];
            }
        }`
```

PR #5457, opened 2019-12-16
[ecommerceframework][elasticsearch] price sorting does not work with elasticsearch 6

markus-moser:
additional fix for #4916
(source filtering existed in ES version 2.x too => therefore this should work for all supported versions: https://www.elastic.co/guide/en/elasticsearch/reference/2.4/search-request-source-filtering.html)


fashxp:
thx

PR #5458, opened 2019-12-16
[ecommerceframework] store table migration fixes

markus-moser:
The current migration implementation does not work if multiple tenants with the same store table implementations are configured.


brusch:
Thanks a lot! 👌

PR #5464, opened 2019-12-17
Fix: Undefined Indexes

blankse:
This errors occur if you want to set a workflow state operator in the object grid


brusch:
Thanks!

PR #5472, opened 2019-12-17
[PHPSTAN] Ignore Doctrine\DBAL\Driver\Connection::query() errors

blankse:
Ignores all phpstan level 2 errors:
`Method Doctrine\DBAL\Driver\Connection::query() invoked with 1 parameter, 0 required`

Line is from https://github.com/doctrine/dbal/blob/master/phpstan.neon.dist


brusch:
Thanks!

PR #5473, opened 2019-12-17
[Data objects] Add field index recommendations

BlackbitNeueMedien:
Pimcore offers the option to add indexes to data object class fields. To help developers adding indexes based on the application's queries this PR analyzes executed data object listing queries (when Pimcore is in debug mode) and gives hints if a column index could help to enhance the query performance.
Example:
```Query execution plan for "SELECT object_1.o_id as o_id, `object_1`.`o_type` FROM `object_1` WHERE ((article_number = ? AND object_1.o_type IN ('object','folder')) AND object_1.o_published = 1)" seems not to be optimal, please consider adding an index to field(s) "article_number" of data object class "Product"```

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/data.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/data.js
@@ -297,7 +297,7 @@ pimcore.object.classes.data.data = Class.create({
 
         var data = this.getData();
         data.name = trim(data.name);
-        var regresult = data.name.match(/[a-zA-Z][a-zA-Z0-9_]*/);
+        var regresult = data.name.match(/[a-zA-Z]\w*/);
 
         if (data.name.length > 1 && regresult == data.name
                             && in_array(data.name.toLowerCase(), this.forbiddenNames) == false) {
--- a/models/DataObject/Listing/Dao.php
+++ b/models/DataObject/Listing/Dao.php
@@ -201,11 +201,7 @@ class Dao extends Model\Listing\Dao\AbstractDao
 
                         if (count($indexCandidates) > 0) {
                             Logger::info(
-                                'Query execution plan for "'.$query
-                                .'" seems not to be optimal, please consider adding an index to field(s) '.implode(
-                                    ',', $indexCandidates
-                                ).' of data object class '.$classDefinition->getName()
-                            );
+                                'Query execution plan for "'.$query.'" seems not to be optimal, please consider adding an index to field(s) "'.implode('", "', $indexCandidates).'" of data object class "'.$classDefinition->getName().'"');
                         }
                     }
                 }
--- a/models/DataObject/Listing/Dao.php
+++ b/models/DataObject/Listing/Dao.php
@@ -193,8 +193,7 @@ class Dao extends Model\Listing\Dao\AbstractDao
                         foreach ($columnNames as $columnName) {
                             $fieldDefinition = $classDefinition->getFieldDefinition($columnName['columnName']);
 
-                            if ($fieldDefinition instanceof DataObject\ClassDefinition\Data
-                                && !$fieldDefinition->getIndex()) {
+                            if ($fieldDefinition instanceof DataObject\ClassDefinition\Data && !$fieldDefinition->getIndex() && !$fieldDefinition->getUnique()) {
                                 $indexCandidates[] = $fieldDefinition->getName();
                             }
                         }
--- a/models/DataObject/Listing/Dao.php
+++ b/models/DataObject/Listing/Dao.php
@@ -19,11 +19,12 @@ namespace Pimcore\Model\DataObject\Listing;
 
 use Pimcore\Db\ZendCompatibility\Expression;
 use Pimcore\Db\ZendCompatibility\QueryBuilder;
+use Pimcore\Logger;
 use Pimcore\Model;
 use Pimcore\Model\DataObject;
 
 /**
- * @property \Pimcore\Model\DataObject\Listing $model
+ * @property \Pimcore\Model\DataObject\Listing\Concrete $model
  */
 class Dao extends Model\Listing\Dao\AbstractDao
 {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by brusch** [comment]
Thanks for your PR, however I think there are already better existing tools for that kind of analysis, like the simple `log-queries-not-using-indexes` config option for MySQL/MariaDB.

**Comment by BlackbitNeueMedien** [comment]
@brusch Of course there are such tools but those tools are mostly used by database admins while the data models of Pimcore are created by developers. In my experience a lot of developers do not know much about indexing - but complain about bad performance. Then caching and such things get added and nobody thinks of database indexes (at least that was the case with all companies I worked for). In the end Pimcore gets blamed for being slow - and that is the worst thing, isn't it?

For me it isn't a valid justification that there are other tools which need to be separately installed / configured while we could give the developers hints to enhance their application's performance out-of-the-box.

**Comment by dpfaffenbauer** [comment]
Well, if a developer doesn't know anything about indexes, he/she should learn about it before complaining about anything. In the end, Pimcore is a System supposed to hold thousands of database entries, having a properly designed database is key here.... And understanding what you do as well.

BTW: Lazy developers don't even care about these things, not even if Pimcore tells them anything, cause they don't understand the problem anyway.

**Comment by BlackbitNeueMedien** [comment]
Ok, thanks for your opinion. Not all developers are as experienced as @brusch or you, Dominik. I thought we could assist unexperienced developers - and also those which simply forgot to add indexes to certain fields.
Experts can always look down from their ivory tower and shake their heads about certain implementations of juniors. Or they could assist them to become better...

**Comment by dpfaffenbauer** [comment]
You said it right, assist them, but not in the form of creating more code that needs maintained. Assist them by showing them how to get better at certain things.

**Comment by BlackbitNeueMedien** [comment]
Ok, I give up on that case.
@brusch would be nice to hear another statement from you. In the end Pimcore will be blamed for bad performance. In my opinion this is much worse than maintaining 30 lines of code.

**Comment by brusch** [comment]
Yes, please continue the discussion 😊Other opinions are highly appreciated.

**Comment by BlackbitNeueMedien** [comment]
I can only refer to for example symfony which tries to help developers when they misspell a `bin/console` command to be executed
```Command "pimcore:non-existing-command" is not defined.     
                                                 
  Did you mean one of these? 
...
```
Following @dpfaffenbauer's argumentation symfony could have said: "Why do we care for typos of developers who are not able to type the command name correctly? A command name suggestion would mean that we have to maintain xx lines of additional code - only because the developers are not able to correctly type the command name."
And following @brusch's argumentation they could have said "A message that the command does not exist is enough. There are tools like executing `bin/console list` to show valid command names".

dpfaffenbauer:
Oh common, that is something completely different. You are comparing apples and pears.

If you are so keen about this, create a Bundle for it...

dpfaffenbauer:
And if you are talking about DX, that feature is poorly implemented in that terms. You are querying the database for possible optimizations on every Listing load call for Entities. That 

 1. takes time to generate
 2. is it not expected to be in there
 3. it creates unasked log entries that eventually no-one every reads in debug anyway.

so, if you really want this feature, create it properly with some nice backend UI's that do exactly that. Then we can talk about DX here.

dpfaffenbauer:
(or at least a CLI command that gives clues about missing optimizations)

BlackbitNeueMedien:
I am open to implementation changes. 
1. Of course it takes time - that's why debug mode should not be enabled on production systems - there are also other places where things are done in debug mode which cost time
2. where else can be gather data about which fields are used for querying for this specific application?
3. I admit, that perhaps nobody will read the generated log entries. I am open for alternative solutions here. But I thought when someone investigates a performance issue perhaps he does look at the logs. Your mentioned CLI command would be another option, of course.

`If you are so keen about this, create a Bundle for it...`
This could only be done by providing own Listing classes which the developer would have to configure his classes to extend. And then this bundle would have no more advantage compared to tools like `log-queries-not-using-indexes`. This feature does only make sense out-of-the-box. If a user wants to further improve performance (e.g. with composite indexes) he will need to use other tools anyway as this is currently not possible in the class definition.

I have never said that this is the optimal way of recommending additional field indexes. Of course everything is enhancable. 
But I think you two are fighting this feature with so much effort that I do not think that it could get merged regardless of what I do.

dpfaffenbauer:
Bundle: not true,

You could create some sort of wrapper Class, that needs a Listing Class, in there you can get the query that Pimcore generates and use that for the EXPLAIN Query.

Similar to what I did with BatchListings: https://github.com/coreshop/Pimcore/blob/master/BatchProcessing/BatchListing.php

Instead you just use `getQuery` function and you got what you need. Done, nice DX, debug ability, output format in what ever you like, done.

I'd be happy to merge something like that into coreshop/pimcore, since there are a lot of tools that enhance DX.

BlackbitNeueMedien:
One last try: Please forget this PR's implementation.

The problem: I know from many production Pimcore systems that often there are nearly no indexes being set for the data object tables. Do we want to implement anything which works out-of-the-box (no additional bundle class which works as a listing wrapper / listing parent class) which could recommend additional indexes on data object fields based on the application's database queries? 
How could a solution look like?

dpfaffenbauer:
if you always find yourself in the situation of slow queries due to missing indexes, you should adjust your toolset and have these tools installed by default on your servers. they do much more than just a simple EXPLAIN on a query. maybe even create CI tools to check queries before deploying anything to production.

It is just not a Pimcore Problem, that is a Database Design Problem. And cause of that, I still stand behind my opinion: that it is not a Pimcore Problem to solve. 

BTW: If you actually experience slow queries, there already is something integrated into the Symfony Profiler. So that should end the discussion:

Bildschirmfoto 2019-12-18 um 17 35 17

BlackbitNeueMedien:
`It is just not a Pimcore Problem, that is a Database Design Problem`
Pimcore creates the database design (based on the class definitions). That is one main advantage compared to an application where you have to care for the database modelling yourself.

The Symfony toolbar also does an EXPLAIN SQL statement - but you need to be able to interpret this information. And which queries would you look at? Perhaps queries which now run in 0.6 seconds can profit by an additional index while another one which needs 1 second cannot.

But I see that our perspectives differ too much: you and @brusch assume that every Pimcore developer uses a performance tracking toolset and is also a database expert. I only wanted to support those who are not.

PS: We have monitoring for our customers. And we add indexes where appropriate. And we earn  money by doing that. But in the end my perspective is always how to improve Pimcore to developers of all levels of experience.

dpfaffenbauer:
Pimcore creates a database tables for you, yes. But, you still have to have a certain know-how to understand what is going on. And if you don't have that know-how, a EXPLAIN query will not solve the problem.... so, that would not help in that cases.

PR #5474, opened 2019-12-18
fix using fallback language values for AbstractRelations

AlternateIf:
In case of AbstractRelations the fallback values have not been considered when lazy loading is used beacause the loadLazyField function only loads the current language. This PR fixes this issue by resolving the fallback language lazy load fields in case they are needed

@weisswurstkanone


weisswurstkanone:
thx

PR #5481, opened 2019-12-18
fix checking fallback languages with inheritance for AbstractRelations

AlternateIf:
fix checking fallback languages with inheritance for AbstractRelations and add diagram describing inheritance and fallback languages for localized fields


AlternateIf:
@weisswurstkanone

PR #5482, opened 2019-12-18
[Assets] Set type to unknown for assets without data

BlackbitNeueMedien:
Currently in https://github.com/pimcore/pimcore/blob/5cb4a06b041e8b3f45337743836ef3430602e7b0/models/Asset/Listing/Dao.php#L47-L51
only assets with non-empty type get loaded.

I have the following script:
```php
<?php
$asset = new \Pimcore\Model\Asset();
$asset->setFilename('abc.txt');
$asset->setParent(\Pimcore\Model\Asset::getByPath('/'));
$asset->save();

$listing = $asset->getList();
$listing->addConditionParam('Filename = ?', 'abc.txt');
var_dump($listing->loadIdList()); // returns array including id of above asset
var_dump($listing->load()); // returns empty array
```

The array does not appear in the Pimcore backend (because of above's restriction in the Dao) - so I cannot delete the asset in the GUI.
Executing above script again results in a primary key collision because there already is a dataset with the same path and filename.

This PR solves that by setting the `type` of an asset to "unknown" if it is empty. We could also set `unknown` as initial value for the `type` class attribute but this is not as secure as checking the value on save.

PS: Alternatively we could prohibit creating assets without data / file. Please report if this is the way to go, then I will change this PR.

### DIFF
--- a/lib/Model/Listing/AbstractListing.php
+++ b/lib/Model/Listing/AbstractListing.php
@@ -32,7 +32,7 @@ abstract class AbstractListing extends AbstractModel implements \Iterator
     protected $order;
 
     /**
-     * @var string|array
+     * @var array
      */
     protected $orderKey;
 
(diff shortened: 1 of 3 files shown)


brusch:
Thanks!

PR #5483, opened 2019-12-18
Caching layer

### LINKED ISSUE #5427: [Data Object] Relation Types DB Caching Layer

weisswurstkanone:
resolves #5427


dpfaffenbauer:
where does the cache get invalidated?

brusch:
When the object gets updated (on save). The raw data is directly on the object as a proeprty, so no additional invalidation is needed.

models/DataObject/ClassDefinition.php
```diff
@@ -1283,4 +1299,24 @@ public function getLinkGenerator()
...
+    /**
+     * @return bool
+     */
+    public function getCacheRawRelationData(): bool
+    {
+        return $this->cacheRawRelationData;
+    }
+
+    /**
+     * @param $cacheRawRelationData
```
blankse: `@param bool $cacheRawRelationData`

models/DataObject/ClassDefinition/Service.php
```diff
@@ -242,9 +242,8 @@ public static function importObjectBrickFromJson($objectBrick, $json, $throwExce
     /**
      * @param array $array
      * @param bool $throwException
-     * @param bool $insideLocalizedField
```
blankse: This param exists

models/DataObject/ClassDefinition/Service.php
```diff
@@ -242,9 +242,8 @@ public static function importObjectBrickFromJson($objectBrick, $json, $throwExce
...
      *
-     * @return mixed
+     * @return bool
```
blankse: Can also be `DataObject\ClassDefinition\Data`

models/DataObject/Concrete.php
```diff
@@ -881,4 +881,59 @@ public static function getLatestVersionByObjectIdAndLatestModificationDate($obje
 
         return;
     }
+
+    /**
+     * @param $descriptor
```
blankse: `@param array $descriptor`

dpfaffenbauer:
what is the actual idea for this? I am obviously missing the point....

markus-moser:
@weisswurstkanone what happens when an object get saved? does it load the relations data always "fresh" from the database or is the cached used for it?

weisswurstkanone:
@blankse , thx!

models/DataObject/Concrete.php
```diff
@@ -883,7 +883,7 @@ public static function getLatestVersionByObjectIdAndLatestModificationDate($obje
     }
 
     /**
-     * @param $descriptor
+     * @param array $descriptor
      * @return array|mixed|mixed[]
```
NiklasBr: This method will now return these two types:
`array`: line 921
`mixed[]`: 936

models/DataObject/Concrete.php
```diff
@@ -881,4 +881,62 @@ public static function getLatestVersionByObjectIdAndLatestModificationDate($obje
...
+            foreach ($descriptor as $column => $expectedValue) {
+                if (is_string($expectedValue)) {
+                    $trimmed = rtrim($expectedValue, '%');
+                    if (strlen($trimmed) < strlen($expectedValue)) {
+                        $likes[$column] = $trimmed;
+                    }
+                }
+            }
+
+            $filterFn = function($row) use ($descriptor, $likes) {
```
brusch: @weisswurstkanone 
for a micro-optimzation we can use `static function` here as `$this` context is not needed

brusch:
@dpfaffenbauer  see docs page: 

> Enable this if you want to retrieve all rows of the affected `object_relation_` table in one go and cache the results.
This can be beneficial if your model contains many lazy loaded fields that are frequently used.

😊 

Imho this should be an advantage in any case, however we'd like to get some more real life experiences on that fist before we make ir the default behavior.

PR #5490, opened 2019-12-19
[Webservice] Set data object's parent id when no parent with this id exists

BlackbitNeueMedien:
In 59cc53bb872836ac29dbc66a113b773c50a88e57 the behaviour of AbstractObject::setParent() was changed: Before `o_parentId` got only updated if the parent existed. Now it gets set to 0 if the parent does not exist.

When creating a data object via `Pimcore\Model\Webservice\Data\DataObject\reverseMap` the data object gets recreated for each attribute via its setter methods. When inheritance is enabled the following chain gets executed:
1. Setter calls getter to get current value to mark field dirty on change. 
1. If value is empty, getter calls:
1. Concrete::getValueFromParent() which calls:
2. Concrete::getParent() which calls:
3. Concrete::setParent(AbstractObject::getById($this->getParentId())  - as it could be the case that no object with the parent id exists (as it was potentially fetched from another Pimcore) `AbstractObject::getById($this->getParentId()` returns `null` in this case
4. Concrete::setParent() calls Concrete::setParentId(0) in in this case - and so the parent id information from the webservice gets lost.

This already happens within the reverseMap loop in https://github.com/pimcore/pimcore/blob/5bebd464ba08352106964edb549ca334108c9914/models/Webservice/Data/DataObject/Concrete.php#L77-L82
First parent id gets set via Concrete::setParentId() but when the first next attribute gets handled whose getter calls Concrete::getValueFromParent() the parent id gets set to 0 (if the parent does not exist).

### DIFF
--- a/models/Webservice/Data/DataObject/Concrete.php
+++ b/models/Webservice/Data/DataObject/Concrete.php
@@ -112,5 +112,8 @@ class Concrete extends Model\Webservice\Data\DataObject
                 }
             }
         }
+
+        // potentially there is no parent with this parentId -> as the setter methods above call Concrete::getValueFromParent() which calls Concrete::getParent() which calls Concrete::setParent(AbstractObject::getById($this->parentId)). As AbstractObject::getById($this->parentId) is null in this case, Concrete::setParent() sets $this->parentId to 0
+        $object->setParentId($this->parentId);
     }
 }


weisswurstkanone:
IMHO, it is better to restore the old behavior of setParent. I think there is no valid reason for that change. Probably just a refactoring accident ...

BlackbitNeueMedien:
Actually I don't know if we simply can revert the changes of 59cc53b to setParent() method. I thought this change may have had a reason ;-)
Could you please check? Imho it is safer to only change the webservice behaviour.
If you think we should revert 59cc53b's changes to setParent(), please do that and close this PR.

weisswurstkanone:
Will merge this as it does not hurt anyway and rethink https://github.com/pimcore/pimcore/commit/59cc53bb872836ac29dbc66a113b773c50a88e57 again :-)

PR #5496, opened 2019-12-20
Added support for constant "PIMCORE_PROJECT_ROOT" for the autoloader

### LINKED ISSUE #5495: [Bug] "PIMCORE_PROJECT_ROOT" constant is not considered
# Bug Report

### Expected behavior
That the constant is considered when loading the autoload file in `dev\lib\Bootstrap.php` method `bootstrap`.

### Actual behavior
She is not considered.

### More informations
See documentation:
https://pimcore.com/docs/5.x/Development_Documentation/Getting_Started/Configuration.html#page_The-PIMCORE-_-PROJECT-_-ROOT-constant

PR will coming soon.

skoch98:
Resolves #5495

### DIFF
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -125,6 +125,8 @@ class Bootstrap
             $loader = include __DIR__ . '/../../../../vendor/autoload.php';
         } elseif (getenv('PIMCORE_PROJECT_ROOT') != '' && file_exists(getenv('PIMCORE_PROJECT_ROOT') . '/vendor/autoload.php')) {
             $loader = include getenv('PIMCORE_PROJECT_ROOT') . '/vendor/autoload.php';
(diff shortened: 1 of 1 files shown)


skoch98:
Oops there is a mistake with my companies and private email. I hope this is OK.

brusch:
@skoch98 I think we can replace the `getenv('PIMCORE_PROJECT_ROOT')`  with just `PIMCORE_PROJECT_ROOT` because `\Pimcore\Bootstrap::setProjectRoot()` is always called before `\Pimcore\Bootstrap::bootstrap()`, so the constant `PIMCORE_PROJECT_ROOT` is always set when `bootstrap()` is called.

brusch:
Fixed in 48d5b04ed533f2c151240745b275f574736a3026, see also #5495

PR #5498, opened 2019-12-20
Improved behavior of non-existent assets

### LINKED ISSUE #5497: [Improvement] Improve behavior of non-existent assets on AssetController
### Expected behavior
When I try to call up an image thumbnail that no longer exists I expect a 404 response.

### Actual behavior
The following is output:
Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function isAllowed() on null in /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/AssetController.php:1163

### Steps to reproduce
Has been noticed by us through an individual ERP import/export development.

The 404 response, as it already happens with "PublicServicesController", would be better filtered during tracking.

skoch98:
Resolves #5497


brusch:
Thanks!

PR #5508, opened 2019-12-30
[Ecommerce] OnlineShopOrder class: change column length of field cartId to 190 -> max. index length

brusch:
Problem: 
Index on column `cartId` cannot be used by e.g. the following query: 
```sql
SELECT object_EF_OSO.o_id as o_id, `object_EF_OSO`.`o_type` FROM `object_EF_OSO` WHERE ((cartId = 'AppBundle\\Ecommerce\\CartManager\\Cart_2596' AND  object_EF_OSO.o_type IN ('object','folder')) AND object_EF_OSO.o_published = 1);
```

Suppressed error message when saving the `OnlineShopOrder` class: 
```
An exception occurred while executing 'ALTER TABLE `object_query_EF_OSO` ADD INDEX `p_index_cartId` (`cartId`);': SQLSTATE[HY000]: General error: 1709 Index column size too large. The maximum column size is 767 bytes.
```

### DIFF
--- a/bundles/CoreBundle/Resources/migrations/Version20191230104529.php
+++ b/bundles/CoreBundle/Resources/migrations/Version20191230104529.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+use Pimcore\Model\DataObject\ClassDefinition;
+
+class Version20191230104529 extends AbstractPimcoreMigration
+{
+    public function doesSqlMigrations(): bool
+    {
+        return false;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $class = ClassDefinition::getById('EF_OSO');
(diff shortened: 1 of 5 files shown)


bundles/CoreBundle/Resources/migrations/Version20191230104529.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        return false;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $class = ClassDefinition::getById('EF_OSO');
```
fashxp: I would go by class name ... since old installations might have numeric class ids for onlineshoporder class.

bundles/CoreBundle/Resources/migrations/Version20191230104529.php
```diff
@@ -0,0 +1,47 @@
...
+            $class->save();
+        }
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        $class = ClassDefinition::getById('EF_OSO');
```
fashxp: I would go by class name ... since old installations might have numeric class ids for onlineshoporder class.

PR #5513, opened 2020-01-02
FIX: DatatransRequest now extending AbstractRequest

HuppertzL:
Added AbstractRequest to DatatransRequest because without it can't be used for its purpose

## Additional info

While working with DatatransRequest (https://pimcore.com/docs/6.x/Development_Documentation/E-Commerce_Framework/Checkout_Manager/Integrating_Payment.html#page_Initialize-Payment-in-Controller) i realized that this Example is not working because `$checkoutManager->startOrderPaymentWithPaymentProvider(AbstractRequest $config);` only accepts AbstractRequests.

In DatatransRequest this abstract was missing, therefore i had to add it.


fashxp:
thx!

PR #5514, opened 2020-01-02
FIX: PayPalSmartPaymentButton

HuppertzL:
+ transform stdClass to json
+ check if return is json

  
  

PayPalSmartPaymentButton fix wrong Parameter-Type in `startPayment`

## Additional info  
In PayPalSmartPaymentButton `startPayment(...)` failed because 
`$this->initPayment(...)` returnes `stdClass` and `Pimcore\Bundle\EcommerceFrameworkBundle\PaymentManager\V7\Payment\StartPaymentResponse\JsonResponse` is only acception `string` as value.

Added check for json and encoding if necessary

### DIFF
--- a/bundles/EcommerceFrameworkBundle/PaymentManager/Payment/PayPalSmartPaymentButton.php
+++ b/bundles/EcommerceFrameworkBundle/PaymentManager/Payment/PayPalSmartPaymentButton.php
@@ -164,7 +164,18 @@ class PayPalSmartPaymentButton extends AbstractPayment implements \Pimcore\Bundl
      */
     public function startPayment(OrderAgentInterface $orderAgent, PriceInterface $price, AbstractRequest $config): StartPaymentResponseInterface
     {
-        $json = $this->initPayment($price, $config->asArray());
+        $result = $this->initPayment($price, $config->asArray());
+
+        if ($result instanceof \stdClass) {
+            $json = json_encode($result);
+        } else {
+            json_decode($result);
+            if (json_last_error() == JSON_ERROR_NONE) {
+                $json = $result;
+            } else {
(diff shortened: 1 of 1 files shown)


bundles/EcommerceFrameworkBundle/PaymentManager/Payment/PayPalSmartPaymentButton.php
```diff
@@ -164,7 +164,18 @@ protected function buildRequestBody(PriceInterface $price, array $config)
...
     {
-        $json = $this->initPayment($price, $config->asArray());
+        $result = $this->initPayment($price, $config->asArray());
+
+        if ($result instanceof \stdClass) {
+            $json = json_encode($result);
+        } else {
+            json_decode($result);
+            if (json_last_error() == JSON_ERROR_NONE) {
+                $json = $result;
```
fashxp: hmm, am I missing something? 
right now, this if is useless ... $json always gets value of $result assigend?

HuppertzL: Well not always, just in case `$result` is not a `stdClass`.
The check if it is a json is in fact useless.

I was not sure, if it would make sense to throw an exception if is is not json

HuppertzL: changed the check and added exception

fashxp: looks better now!

fashxp:
thx very much!

PR #5517, opened 2020-01-03
Deprecated REST Webservice API

brusch:
The API will be removed from the core in Pimcore 7, use the [Pimcore Data-Hub](https://github.com/pimcore/data-hub) instead.

Edit: 

Our motivation behind deprecating the REST API: 
- very limited functionality due to the antiqued architecture
- very inflexible due to the fixed and full-featured data-structures
- not compatible with nowadays requirements of a configurable API
- no permissions, no limitations, no transformations, just one big API endpoint

Even if this functionality is removed from the core, it does not mean that it cannot continue to exist as a bundle. If you're interested, please let us know, maybe we can form a task force for this topic 😊

### DIFF
--- a/bundles/AdminBundle/Controller/Rest/AbstractRestController.php
+++ b/bundles/AdminBundle/Controller/Rest/AbstractRestController.php
@@ -26,6 +26,9 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 use Symfony\Component\Stopwatch\Stopwatch;
 
+/**
+ * @deprecated
+ */
 abstract class AbstractRestController extends AdminController
 {
     /**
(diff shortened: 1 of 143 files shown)


brusch:
Even if this functionality is removed from the core, it does not mean that it cannot continue to exist as a  bundle. If you're interested, please let us know, maybe we can form a task force for this topic 😊

weisswurstkanone:
Just a warning: I believe the webservice is used by the classdefinition json export.

brusch:
See #5548

FredoVelcro:
what about the API key management ?

brusch:
@FredoVelcro thx 👍 added a note there as well, see: ab2fcea65b99e350307644f87cbe4d238e7f082c

image

betterapp:
> Even if this functionality is removed from the core, it does not mean that it cannot continue to exist as a bundle. If you're interested, please let us know, maybe we can form a task force for this topic blush.

I am interested with it.

fashxp:
It is part of our LTS offering ... Please see https://pimcore.com/en/services/lts for details

PR #5519, opened 2020-01-03
adding DIRECTORY_SEPARATOR resolves #5518

### LINKED ISSUE #5518: [ApplicationLogger] Error on showing FileObject on Windows
Hi,

I have succesfully created a couple of entries on the ApplicationLogger.
When I try to open the attached FileObject I get the following error:

preg_match(): Compilation failed: unknown property name after \P or \p at offset 18

The error is in the LogController class; here the reference https://github.com/pimcore/pimcore/blob/91a113ea5a14479f49bad9e463349018aae71b93/bundles/AdminBundle/Controller/Admin/LogController.php#L247

The problem is that, on Windows, file paths contains the backslash "\\" character instead of the slash "/" as linux.
It's common to have a \pimcore subfolder in Pimcore's projects, but for the preg_match function, \p is a pattern escape, and so the error.

MohammedAlkutrani:
fixing directory bug 

## Additional info  
The problem is that, on Windows, file paths contains the backslash "\" character instead of the slash "/" as linux. now it can work on both

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/LogController.php
+++ b/bundles/AdminBundle/Controller/Admin/LogController.php
@@ -240,7 +240,7 @@ class LogController extends AdminController implements EventedControllerInterfac
     public function showFileObjectAction(Request $request)
     {
         $filePath = $request->get('filePath');
-        $filePath = PIMCORE_PROJECT_ROOT . '/' . $filePath;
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
thx!

MohammedAlkutrani:
Welcome

PR #5527, opened 2020-01-05
[Assets] Fix: Wrong document thumbnail if thumbnail generation is locked

blankse:
ImageThumbnail::generate() method must set the filesystemPath argument. A return value is not taken into account.

### DIFF
--- a/models/Asset/Document/ImageThumbnail.php
+++ b/models/Asset/Document/ImageThumbnail.php
@@ -69,8 +69,6 @@ class ImageThumbnail
 
     /**
      * @param bool $deferredAllowed
-     *
-     * @return string
      */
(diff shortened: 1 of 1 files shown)


brusch:
Path to error image is already set to a variable earlier in the code, so we can reuse that, did this in dd330486d285fada288c8ed3cb469c2f96c109f2 😉
Thanks anyway 👍

blankse:
@brusch It was a other image. Can we remove please-wait.png than? Its not used anymore. I also changed the phpdoc...

brusch:
Oh, sorry, you're right, so your PR was correct. 
Thanks for the info, fixed that 🙏

PR #5530, opened 2020-01-06
[PHPSTAN] Fix: PHPDoc tag param references unknown parameter

blankse:
Fix all phpstan level 2 errors with schema:
```
PHPDoc tag @param references unknown parameter: $xyz
```


brusch:
Thanks a lot! 🙏

Issue #3681, opened 2018-11-30
Timeout during installation


khusseini:
# Bug Report
 
  
### Expected behavior 
Installer should not run into timeout, or the timeout should be configurable
  
### Actual behavior  
`
 * The process "'/usr/bin/php' '/var/www/bin/console' 'pimcore:deployment:classes-rebuild' '-c'" exceeded the timeout of 60 seconds.
`

A lot of times when I install pimcore, I get a timeout from `pimcore:deployment:classes-rebuild`. Is there any way to increase the timeout?

brusch:
This tasks shouldn't take more than 1-2 secs on an average system, no idea why it takes so long for you. 
Anyway, I've increased the timeout to 120 secs, what  shouldn't hurt.

PR #5533, opened 2020-01-07
Increase process timeout for asset installation

### LINKED ISSUE #5532: Timeout during installation:  assets:install web --symlink --relative --ansi" exceeded the timeout of 60 seconds.
# Bug Report
Install script crashes with the following error:
```
 * The process "C:\xampp\php\php.exe "C:\src\cispa\pimcore2/bin/console" assets:install web --symlink --relative --ansi" exceeded the timeout of 60 seconds.
```

### Expected behavior
Installer should run through.

### Actual behavior
See above.

### Steps to reproduce
Install pimcore skeleton via composer; run the install script.

This issue occured for some of our project team, not for others. I am suspecting it has to do with our environment (Win 10, using git bash), but I am unsure why it occurs. Interestingly, installing the assets separately works, BUT it installs them as copies even thogh the --symlink option is given. Maybe this is the reason why it takes so long.
```
$ bin/console assets:install web --symlink --relative --ansi

 Trying to install assets as relative symbolic links.

 --------- -------------------- ----------------
            Bundle               Method / Error
 --------- -------------------- ----------------
  WARNING   PimcoreCoreBundle    copy
  WARNING   PimcoreAdminBundle   copy
 --------- -------------------- ----------------

 ! [NOTE] Some assets were installed via copy. If you make changes to these assets you have to run this command again.


 [OK] All assets were successfully installed.
```

beheist:
Related to #5532, increasing the process timeout fixes the issue. This was fixed in a similar fashion for #3681, so I suggest doing the same here.

  
  

Resolves #5532

### DIFF
--- a/lib/Tool/AssetsInstaller.php
+++ b/lib/Tool/AssetsInstaller.php
@@ -49,6 +49,7 @@ class AssetsInstaller
     public function install(array $options = []): Process
     {
         $process = $this->buildProcess($options);
+        $process->setTimeout(240);
         $process->run($this->runCallback);
 
         if (!$process->isSuccessful()) {


brusch:
Thanks!

PR #5538, opened 2020-01-07
Adding extended return type declaration

ChrisB9:
The method returns not just the Object but also null. 
I just added it to this one currently, but it should be added to every single model to satisfy phpstan

Resolves errors with phpstan and null comparison on model methods

## Additional info  

Reproduce with:
- phpstan active: we have max level activated
- write something like Object::getByPath('/') === null

this results in:
Strict comparison using === between null and                            
         Pimcore\Model\DataObject\AbstractObject will always evaluate to false.


brusch:
Thanks!

PR #5541, opened 2020-01-07
Add configuration for the initial_markings to replace the deprecated …

davidmaes:
…initial_place

<!--
- [x] We're not accepting any feature PR's only for **version 5** anymore, you have to provide a feature PR for both versions. 
- [x] Submit bugfixes for version 5 to the target branch `5.8`, version 6 is `master` branch.

Resolves #
The deprecation of initial_places in Symfony's workflow component. Using initial_markings instead while providing backward compatibility.

## Additional info  
None

### DIFF
--- a/bundles/CoreBundle/DependencyInjection/Configuration.php
+++ b/bundles/CoreBundle/DependencyInjection/Configuration.php
@@ -1434,6 +1434,7 @@ class Configuration implements ConfigurationInterface
                                     ->info('Will be applied when the current place is empty.')
                                 ->end()
                                 ->arrayNode('initial_markings')
+                                    ->info('Can be used to set the initial places (markings) for a workflow. Note that this option is Symfony 4.3+ only')
                                     ->beforeNormalization()
                                         ->ifString()
                                             ->then(function ($v) {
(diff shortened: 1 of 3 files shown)


bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -1433,6 +1433,18 @@ private function addWorkflowNode(ArrayNodeDefinition $rootNode)
                                     ->defaultNull()
```
fashxp: please set this option to deprecated.

bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -1433,6 +1433,18 @@ private function addWorkflowNode(ArrayNodeDefinition $rootNode)
...
                                     ->info('Will be applied when the current place is empty.')
                                 ->end()
+                                ->arrayNode('initial_markings')
+                                    ->beforeNormalization()
```
fashxp: add info node with short description & information that this option will work with symfony 4 only ... since symfony 3 does not allow an array as inital place ... see also https://github.com/symfony/workflow/blob/3.4/Definition.php

bundles/CoreBundle/DependencyInjection/Configuration.php
davidmaes: Ah, didn't even know that was an option. If Travis is happy, I think I've resolved both your comments. Thanks for the review!

fashxp:
thank you very much!

PR #5549, opened 2020-01-08
[Workflows] Transition notification user / role should be optional

BlackbitNeueMedien:
It should be possible to define either users or roles to get notified when a workflow transition happens (or both) - so these fields should be optional.

In addition I changed the SQL query for looking up the users to be notified to use `IN()` instead of `FIND_IN_SET()`. `FIND_IN_SET` cannot use the index on the `name` column in users. If there are commas in user or role names the FIND_IN_SET would not work.


brusch:
Makes sense, thanks a lot!

PR #5550, opened 2020-01-08
[Documents|Objects] Fix children and siblings result cache

### LINKED ISSUE #5521: [Documents|Objects] Non-deterministic results of getChildren and hasChildren
- if on a document `getChildren(true)` is called, also unpublished children are loaded into local `children` variable and returned. 
- if later in same process `getChildren()` is called, unpublished children are still returned even not expected. 

- when calling just `getChildren()` first and later `getChildren(true)`, on second call no unpublished children are returned even they are expected though. 

same goes for `hasChildren`. 

`$unpublished` parameter needs to be considered when caching the results of these methods.

dvesh3:
Resolves #5521
Resolves #5103


models/Document.php
```diff
@@ -653,7 +653,7 @@ public function getDependencies()
      */
     public function setChildren($children)
     {
-        $this->children = $children;
+        $this->children[0] = $children;
```
weisswurstkanone: What if $children contains unpublished ones or is this just for internal use ?

dvesh3: default index for internal use

weisswurstkanone:
Can you add a unit test ?

models/Document/Dao.php
```diff
@@ -379,9 +379,17 @@ public function getChildAmount($user = null)
      *
      * @return bool
      */
-    public function hasSiblings()
+    public function hasSiblings($unpublished = false)
```
blankse: Please add also to phpdoc

dvesh3: done! thnx :)

models/DataObject/AbstractObject.php
```diff
@@ -438,15 +441,14 @@ public function getChildren($objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJE
      */
     public function hasChildren($objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER], $unpublished = false)
     {
-        if (is_bool($this->o_hasChildren)) {
-            if (($this->o_hasChildren and empty($this->o_children)) or (!$this->o_hasChildren and !empty($this->o_children))) {
-                return $this->getDao()->hasChildren($objectTypes, $unpublished);
-            } else {
-                return $this->o_hasChildren;
+        if (isset($this->hasChildren[$unpublished])) {
```
blankse: `if (isset($this->o_hasChildren[$unpublished])) {`

models/DataObject/AbstractObject.php
```diff
@@ -411,9 +411,11 @@ protected static function typeMatch(AbstractObject $object)
      *
      * @return self[]
      */
-    public function getChildren($objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER], $unpublished = false)
+    public function getChildren(array $objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER], $unpublished = false)
     {
-        if ($this->o_children === null || $this->lastGetChildrenObjectTypes != $objectTypes) {
+        $cacheKey = implode('_', $objectTypes) . (!empty($objectTypes) ?'_' : '') . (int) $unpublished;
```
brusch: We should move that to a `protected` function so that we can reuse it in `hasChildren()` `getSiblings()` ... 
Maybe something generic like: 
```php
$this->getListingCacheKey(func_get_args()); 
```

models/DataObject/AbstractObject.php
brusch: We need a cache key considering the `$objectTypes` here as well. 
Use the to be created function `getListingCacheKey()` here.

models/DataObject/AbstractObject.php
brusch: We should also cast to `string` instead of `int` as it get's converted to `string` anyway.

models/DataObject/AbstractObject.php
```diff
@@ -468,30 +472,30 @@ public function getSiblings($objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJE
...
      */
-    public function hasSiblings($objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER])
+    public function hasSiblings($objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER], $unpublished = false)
     {
-        if (is_bool($this->o_hasSiblings)) {
-            if (($this->o_hasSiblings and empty($this->o_siblings)) or (!$this->o_hasSiblings and !empty($this->o_siblings))) {
-                return $this->getDao()->hasSiblings($objectTypes);
-            } else {
-                return $this->o_hasSiblings;
+        if (isset($this->o_hasSiblings[$unpublished])) {
```
brusch: Here we also need to consider the `$objectTypes` for caching, use `getListingCacheKey()` here as well.

models/DataObject/Listing.php
```diff
@@ -140,6 +140,8 @@ public function resetConditionParams()
      */
     public function setCondition($condition, $conditionVariables = null)
     {
+        $this->setData(null);
```
brusch: This needs to be done for other methods that affect the result as well, such as `addObjectbrick()` `setLocale()` `setIgnoreLocalizedFields()` ....

models/Document.php
```diff
@@ -674,16 +676,16 @@ public function setChildren($children)
      */
     public function getChildren($unpublished = false)
     {
-        if ($this->children === null) {
+        if (!isset($this->children[$unpublished])) {
```
brusch: I'd go for a `getListingCacheKey()` method here as well, even though a bool get's converted to an int when used as array key, we should cast it in advance, just to avoid future issues :)

models/DataObject/AbstractObject.php
dvesh3: thanks for the catch!

dvesh3:
@brusch I have pushed the changes requested..also I have moved data flushing for listing to AbstractListing.php. Can you please review again? Thanks! :)

models/DataObject/AbstractObject.php
```diff
@@ -1120,8 +1124,7 @@ public function setChildren($children)
             $this->o_hasChildren = [];
         } else {
             //default cache key
-            $cacheKey = implode('_', [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER]) . '_0';
-
+            $cacheKey = $this->getListingCacheKey();
```
blankse: Required parameter '$args' missing

models/Document.php
```diff
@@ -656,11 +656,12 @@ public function setChildren($children)
             $this->hasChildren = [];
             $this->children = [];
         } else {
-            $this->children[0] = $children;
+            $cacheKey = $this->getListingCacheKey();
```
blankse: Required parameter '$args' missing

models/Document.php
```diff
@@ -695,15 +700,16 @@ public function getChildren($unpublished = false)
...
     {
-        if (is_bool($this->hasChildren)) {
-            if (($this->hasChildren and empty($this->children)) or (!$this->hasChildren and !empty($this->children))) {
-                return $this->getDao()->hasChildren($unpublished);
-            } else {
-                return $this->hasChildren;
+        $cacheKey = $this->getListingCacheKey(func_get_args());
+
+        if (isset($this->hasChildren[$cacheKey])) {
+            if (($this->hasChildren[$cacheKey] and empty($this->children[$cacheKey])) or (!$this->hasChildren[$cacheKey] and !empty($this->children[$cacheKey]))) {
```
blankse: Is this type of ifs really necessary? If I call this function twice, the database will be queried twice because the children array are still empty. The 2 arguments are protected and  should always be synchronous.

brusch: I also don't think that this is necessary, changed that, see 0d44a8a3074057a561f42cd527684e6f90445fad

models/DataObject/AbstractObject.php
```diff
@@ -995,6 +987,7 @@ public function setParentId($o_parentId)
         }
         $this->o_parentId = $o_parentId;
         $this->o_parent = null;
+        $this->o_siblings = null;
```
blankse:
```
$this->o_siblings = [];
$this->o_hasSiblings = [];
```

brusch: @dvesh3 resetting on `setParentId()` should be also considered for documents.

dvesh3: done! thnx :)

models/Document/Dao.php
```diff
@@ -377,11 +377,21 @@ public function getChildAmount($user = null)
     /**
      * Checks if the document has siblings
      *
+     * @param bool $unpublished
+     *
      * @return bool
      */
-    public function hasSiblings()
+    public function hasSiblings($unpublished = false)
```
weisswurstkanone: please clarify the meaning of $unpublished. Maybe find a better name for the flag and/or document it in the phpdoc.

models/Document/Dao.php
```diff
@@ -377,11 +377,21 @@ public function getChildAmount($user = null)
...
     {
-        $c = $this->db->fetchOne('SELECT id FROM documents WHERE parentId = ? and id != ? LIMIT 1', [$this->model->getParentId(), $this->model->getId()]);
+        $sql = 'SELECT id FROM documents WHERE parentId = ? and id != ?';
+
+        if (Model\Document::doHideUnpublished() && !$unpublished) {
+            $sql .= ' AND published = 1';
```
weisswurstkanone: meaning that I can set the flag to whatever I want if the first check isn't satisfied ?

models/Document/Dao.php
brusch: maybe `$includingUnpublished` ?

brusch: The same would be also for `getChildren()` and possibly others.

models/Document/Dao.php
brusch: We should probably use `null` as the default value for `$unpublished` then we can check whether it was set or if the default from `doHideUnpublished()` should be used.

models/Document.php
```diff
@@ -1491,4 +1495,11 @@ public function setVersionCount(?int $versionCount): ElementInterface
 
         return $this;
     }
+
+    protected function getListingCacheKey(array $args = []) {
+        $unpublished = $args[0] ?? false;
```
weisswurstkanone: maybe cast to bool here

        $children = $doc->getChildren(true);
        $children = $doc->getChildren('true');

will return 2 different cache keys

models/DataObject/AbstractObject.php
```diff
@@ -1104,14 +1115,22 @@ public function setUserModification($o_userModification)
...
      */
     public function setChildren($children)
     {
-        $this->o_children = $children;
-        $this->o_hasChildren = (is_array($children) && count($children) > 0);
+        if($children === null) {
+            // unset all cached children
+            $this->o_children = [];
+            $this->o_hasChildren = [];
+        } elseif(is_array($children)) {
```
weisswurstkanone: Is my interpretation correct that setChildren([]) will keep the cached unpublished ones ?

models/Document.php
```diff
@@ -1110,6 +1112,8 @@ public function setParentId($parentId)
     {
         $this->parentId = (int) $parentId;
         $this->parent = null;
+        $this->siblings = [];
```
weisswurstkanone: Can you can explain this? The parent doesn't change until save is called. Otherwise you would have to do the same thing in setPublished

weisswurstkanone: Can you explain this. The real parent doesn't change until save is called.

models/Document/Dao.php
dvesh3: done!

models/Document/Dao.php
dvesh3: done!

models/Document.php
dvesh3: done!

models/DataObject/AbstractObject.php
dvesh3: you are right.. change to empty check!

models/Document.php
dvesh3: when you change `$parentId` then you should be able to fetch siblings based on new `$parentId` not cached one. please see https://github.com/pimcore/pimcore/blob/7111e7423bf43926a3620e43680b1a00204f42c1/models/Document.php#L723

models/DataObject/AbstractObject.php
```diff
@@ -407,91 +397,93 @@ protected static function typeMatch(AbstractObject $object)
...
      * Quick test if there are children
      *
      * @param array $objectTypes
-     * @param bool $unpublished
+     * @param bool|null $includingUnpublished
      *
      * @return bool
      */
-    public function hasChildren($objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER], $unpublished = false)
+    public function hasChildren($objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER], $includingUnpublished = null)
```
fashxp: array typehint missing

brusch: I guess because it wasn't there before ... however the annotation is correct.

fashxp: but it got added on all other methods in this class :)

PR #5553, opened 2020-01-09
[PHPSTAN] Create IdMapperInterface

blankse:
Fix errors with undefined methods on the $idMapper variables. I create a interface only for the phpdoc.

Before:
```
 [ERROR] Found 3551 errors
```

After:
```
 [ERROR] Found 3536 errors
```


brusch:
Thanks a lot!

PR #5558, opened 2020-01-09
[Workflows] add option to force a object layout after a transition or global action was performed

markus-moser:
This PR adds a option for transitions and global layouts to force a object layout after the transition/action was performed via the actions button.

Additionally some PHP notices are fixed.


fashxp:
what counts if target place also has a custom layout defined? 
we should clarify that in info tags

markus-moser:
@fashxp docs are clarified now

PR #5561, opened 2020-01-09
Fix deprecated (PHP 7.4) parameter order of implode()

jdreesen:
Fixes the parameter order of `implode()` that has been [deprecated in PHP 7.4](https://www.php.net/manual/en/migration74.deprecated.php#migration74.deprecated.core.implode-reverse-parameters).


brusch:
Thanks a lot!

PR #5564, opened 2020-01-09
[PHPSTAN] Fill empty php doc param types (bundles)

blankse:
Fix all phpstan level 2 errors in bundles/ (also doc/ und tests/ ...) with following schema:
```
```
Before:
```
 [ERROR] Found 3539 errors
```
After:
```
 [ERROR] Found 3251 errors
```


weisswurstkanone:
Thanks a lot!

PR #5565, opened 2020-01-10
[PHPSTAN] Fill empty php doc param types (lib) 

blankse:
Fix all phpstan level 2 errors in lib/ with following schema:
```
```
Before:
```
 [ERROR] Found 3539 errors
```
After:
```
 [ERROR] Found 3051 errors 
```


weisswurstkanone:
Mille Grazie!

PR #5568, opened 2020-01-10
[DOCS] Areablock configuration controlsTrigger is wrong

ultramegatom:
hover seems to be the default value, see: https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Resources/public/js/pimcore/document/tags/areablock.js#L50

not sure if this is on purpose i would rather prefer fixed as the default value.


brusch:
It used to be fixed in earlier versions (as far as I can remember in versions lower than 3), but now hover is the default. Thanks for your PR!

PR #5569, opened 2020-01-10
[Areabrick] Add some top padding

ultramegatom:
Added some top padding to the areabrick toolbar in fixed mode.


brusch:
Thanks a lot!

PR #5570, opened 2020-01-10
[PHPSTAN] Fill empty php doc param types (Asset)

blankse:
Fix all phpstan level 2 errors in Pimcore\Model\Asset namespace with following schema:
```
```
Before:
```
 [ERROR] Found 2731 errors
```
After:
```
 [ERROR] Found 2646 errors 
```


brusch:
Well done - thanks a lot! 🙏

PR #5571, opened 2020-01-10
[PHPSTAN] Fill empty php doc param types (Document)

blankse:
Fix all phpstan level 2 errors in Pimcore\Model\Document namespace with following schema:
```
```
Before:
```
 [ERROR] Found 2731 errors
```
After:
```
 [ERROR] Found 2632 errors
```

### DIFF
--- a/models/Document/PageSnippet.php
+++ b/models/Document/PageSnippet.php
@@ -113,7 +113,7 @@ abstract class PageSnippet extends Model\Document
     /**
      * @param bool $setModificationDate
      * @param bool $saveOnlyVersion
-     * @param $versionNote string version note
+     * @param string $versionNote
      *
      * @return null|Model\Version
      *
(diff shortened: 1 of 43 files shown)


models/Document/PageSnippet.php
```diff
@@ -113,7 +113,7 @@ protected function update($params = [])
     /**
      * @param bool $setModificationDate
      * @param bool $saveOnlyVersion
-     * @param $versionNote string version note
+     * @param string $versionNote
```
weisswurstkanone: "version note" description missing

brusch: I would say the variable name speaks for itself, so it's ok.

brusch:
Again, thank you so much!

models/Document/PageSnippet.php
blankse: Yes I also thought that the description has no added value

PR #5572, opened 2020-01-10
[PHPSTAN] Fill empty php doc param types (Model)

blankse:
Fix all phpstan level 2 errors in Pimcore\Model namespace (without Asset|DataObject|Document) with following schema:
```
```
Before:
```
 [ERROR] Found 2731 errors
```
After:
```
 [ERROR] Found 2318 errors
```

### DIFF
--- a/models/Dependency.php
+++ b/models/Dependency.php
@@ -99,8 +99,8 @@ class Dependency extends AbstractModel
     }
 
     /**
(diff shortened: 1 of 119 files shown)


brusch:
Awesome! 🙋‍♂️ Thanks!

PR #5573, opened 2020-01-11
[PHPSTAN] Fill empty php doc param types (DataObject)

blankse:
Fix many (not all) phpstan level 2 errors in Pimcore\Model\DataObject namespace with following schema:
```
```
Before:
```
 [ERROR] Found 2731 errors
```
After:
```
 [ERROR] Found 2267 errors
```

### DIFF
--- a/models/DataObject/ClassDefinition/CustomLayout/Dao.php
+++ b/models/DataObject/ClassDefinition/CustomLayout/Dao.php
@@ -30,7 +30,7 @@ class Dao extends Model\Dao\AbstractDao
     protected $model;
 
     /**
-     * @param null $id
+     * @param string|null $id
      *
      * @throws \Exception
      */
@@ -70,7 +70,7 @@ class Dao extends Model\Dao\AbstractDao
     }
 
     /**
-     * @param mixed $id
+     * @param string $id
      *
      * @return string|null
      */
@@ -89,7 +89,7 @@ class Dao extends Model\Dao\AbstractDao
 
     /**
      * @param string $name
-     * @param int    $classId
+     * @param string $classId
      *
      * @return int|null
      */
@@ -107,7 +107,7 @@ class Dao extends Model\Dao\AbstractDao
     }
 
     /**
-     * @return int|mixed
+     * @return int
      */
     public function getNewId()
     {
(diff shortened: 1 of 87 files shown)


models/DataObject/ClassDefinition/CustomLayout/Dao.php
```diff
@@ -107,7 +107,7 @@ public function getIdByNameAndClassId($name, $classId)
     }
 
     /**
-     * @return int|mixed
+     * @return int
```
brusch: Should be `int|null` right?

models/DataObject/Data/ObjectMetadata.php
```diff
@@ -117,23 +117,23 @@ public function __call($name, $arguments)
     /**
      * @param $object
      * @param string $ownertype
-     * @param $ownername
-     * @param $position
-     * @param $index
+     * @param string $ownername
+     * @param int $position
```
brusch: `$position`  can be a string as well (e.g. localized fields).

models/DataObject/Data/ObjectMetadata.php
```diff
@@ -117,23 +117,23 @@ public function __call($name, $arguments)
...
-     * @param $fieldname
-     * @param $ownertype
-     * @param $ownername
-     * @param $position
-     * @param $index
+     * @param int $destinationId
+     * @param string $fieldname
+     * @param string $ownertype
+     * @param string $ownername
+     * @param int $position
```
brusch: Same here: `string`

models/DataObject/ClassDefinition/CustomLayout/Dao.php
blankse: Why? If the DB return null, the value will be set to 1

brusch: You're right, sorry!

models/DataObject/Data/ObjectMetadata.php
blankse: You are right. I changed it

models/DataObject/Data/ObjectMetadata.php
blankse: You are right. I changed it

PR #5576, opened 2020-01-14
Fix wrong objectbricks phpdoc type

blankse:
After upgrade to 6.4.0 the return type in phpdocs of objectbricks are without \ at beginning.

image


brusch:
Thanks!

PR #5577, opened 2020-01-14
Adding null to getById

ChrisB9:
Adding the same thing to getById i added to getByPath
see: https://github.com/pimcore/pimcore/commit/3f0e55f6046a977db3004a98d590f1bc7bebf042


brusch:
Thanks! ☺

PR #5578, opened 2020-01-14
Object Preview doesn't work with relations

### LINKED ISSUE #5106: Object Preview doesn't work with lazy loaded relations
I debugged the issue, the problem are lazy loaded fields, since they don't get serialized without _fulldump. Also, when you add _fulldump for the session serialization, it get's removed later, cause \Pimcore\Tool\Session opens and closes the session like a million times per request... which unserializes it, removes the _fulldump and serializes it again.

weisswurstkanone:
Fixes #5106


models/Element/Service.php
```diff
@@ -1228,4 +1233,78 @@ public static function getNoteData(Note $note)
...
+            $element->setInDumpState(true);
+        }
+        $serializedData = Serialize::serialize($element);
+
+        TmpStore::set($tmpStoreKey, $serializedData, $tag);
+    }
+
+
+    /**
+     * @param $type
```
blankse: Please add types

doc/Development_Documentation/05_Objects/01_Object_Classes/05_Class_Settings/15_Link_Generator.md
```diff
@@ -8,75 +8,127 @@ Additionally they are also enabling the preview tab for data objects.
...
+     * @param array $params
+     * @return string
+     */
+    public function generateWithMockup(ProductInterface $object, array $params = []): string {
+        return $this->doGenerate($object, $params);
+    }
 
-        return $path;
+    /**
+     * @param $object
```
blankse:
```
     * @param ProductInterface $object
     * @param array $params
```

Issue #5584, opened 2020-01-15
Operator in filter method of data object listing breaks implementation


deltatag:
# Expected behavior
Usage of class listing with field name "operator" should be possible

### Actual behavior
Using a listing of a class with field name "operator" fails due to two method parameters with same name generated for listing.

This problem was created with implementation of #4351 / #4355 

### Steps to reproduce
* Create class with field name "operator" (e.g. input)
* Use listing to load objects
* Fails due to compile error "Compile Error: Redefinition of parameter $operator"

![image](https://user-images.githubusercontent.com/7627410/72415379-a1e38d00-3774-11ea-8577-b3df9484a3b7.png)

### Solution
All Pimcore used parameters should be blocked when creating the class definition (e.g. same as with reserved field names)

weisswurstkanone:
@BlackbitNeueMedien , can you have a look ?

weisswurstkanone:
fixed via #5585

PR #5585, opened 2020-01-15
[Data object listings] Prevent name collision for field named "operator"

BlackbitNeueMedien:
Solves #5584 

Now the `filterByXYZ` method always uses parameters `$data`and `operator`. It was not necessary to use the real field name here.


weisswurstkanone:
thanks a lot!

PR #5128, opened 2019-10-23
allow custom field-definitions to marshal and unmarshal version data differently

dpfaffenbauer:
This makes a lot of Problems with complex custom data-types where we in CoreShop link DataObjects with Doctrine Entities. This PR basically adds support for Custom Data Types to marshal and unmarshal Data for Versions as they need id.

PR #5588, opened 2020-01-15
[Recycle Bin] Optimize relations serialization

### LINKED ISSUE #5426: [Recycle Bin] Use ElementDescriptor to store relations
Currently the entire objects get serialized. 
Similar solution to what we do with versions: 
https://github.com/pimcore/pimcore/blob/1d54158b50d52c3f220d68fb1f667acc3cc99a4b/models/Version.php#L273

Using: `\Pimcore\Model\Version\ElementDescriptor`

dvesh3:
Resolves #5426


dpfaffenbauer:
Why don't you use the exact same behaviour as with versions?

brusch:
@dpfaffenbauer it is actually very similar, but the recycle bin contains also the data of children which needs a bit different processing 😊

brusch:
@dvesh3 please check travis: https://travis-ci.org/pimcore/pimcore/jobs/637340097#L4425

models/Element/Recyclebin/Item.php
```diff
@@ -329,6 +326,67 @@ protected function doRecursiveRestore(Element\ElementInterface $element)
...
+                }
+            }
+        );
+        $copier->addFilter(new Model\Version\SetDumpStateFilter(true), new \DeepCopy\Matcher\PropertyMatcher(Element\ElementDumpStateInterface::class, Element\ElementDumpStateInterface::DUMP_STATE_PROPERTY_NAME));
+
+        return $copier->copy($data);
+    }
+
+    /**
+     * @param $data
```
blankse: Please add type

dpfaffenbauer:
Please also add some kind of custom serialization for custom data-types. So I can add that to CoreShop as well

dvesh3:
@dpfaffenbauer Hey, i didn't understand your point. We are filtering based on value of fields(that's why i added ` $element instanceof Element\ElementInterface` check in matches method) not based on type of fields. Can you please explain your problem in detail or can also propose a PR if you have better solution. thanks!

models/Element/Recyclebin/Item.php
```diff
@@ -314,7 +311,7 @@ protected function doRecursiveRestore(Element\ElementInterface $element)
 
         if (method_exists($element, 'getChildren')) {
             if ($element instanceof DataObject\AbstractObject) {
-                $children = $element->getChildren([], true);
+                $children = $element->getChildren([DataObject::OBJECT_TYPE_FOLDER, DataObject::OBJECT_TYPE_VARIANT, DataObject::OBJECT_TYPE_OBJECT]);
```
weisswurstkanone: includingUnpublished droped by accident ?

models/Element/Recyclebin/Item.php
```diff
@@ -329,6 +326,67 @@ protected function doRecursiveRestore(Element\ElementInterface $element)
...
+                 * @return boolean
+                 */
+                public function matches($element)
+                {
+                    //compress only elements with full_dump_state = false
+                    return $element instanceof Element\ElementInterface && $element instanceof Element\ElementDumpStateInterface && !($element->isInDumpState());
+                }
+            }
+        );
+        $copier->addFilter(new Model\Version\SetDumpStateFilter(true), new \DeepCopy\Matcher\PropertyMatcher(Element\ElementDumpStateInterface::class, Element\ElementDumpStateInterface::DUMP_STATE_PROPERTY_NAME));
```
weisswurstkanone: Please check in which order the filters get applied. SetDumpStateFilter must be executed first I guess.

models/Element/Recyclebin/Item.php
```diff
@@ -329,6 +326,67 @@ protected function doRecursiveRestore(Element\ElementInterface $element)
         }
     }
 
+    /**
+     * @param Element\ElementInterface $data
+     *
+     * @return mixed
+     */
+    public function marshalData($data)
```
weisswurstkanone: please have a look at @dpfaffenbauer comment about the custom filter and incorporate his idea if possible

dpfaffenbauer:
@dvesh3 basically what I did here: https://github.com/pimcore/pimcore/pull/5128

I have lots of custom pimcore data types in CoreShop, they use doctrine entities to persist data. They don't work with default Pimcore things, so they need a custom way of serializing and unserializing.

PR #5590, opened 2020-01-15
respect error reporting in installer

solverat:
Regression of #5310 (@blankse): otherwise the installer uses `E_ALL` as default which will rise up all the notice errors while trying to boot the system.


brusch:
Thanks a lot! 🙋‍♂️

PR #5594, opened 2020-01-15
[Data Object] Default values should be also set to database on create

weisswurstkanone:
resolves #4846

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Checkbox.php
+++ b/models/DataObject/ClassDefinition/Data/Checkbox.php
@@ -229,6 +229,7 @@ class Checkbox extends Data implements ResourcePersistenceAwareInterface, QueryR
     }
 
     /**
+     * @deprecated
      * @param DataObject\AbstractObject $object
      * @param array $params
      *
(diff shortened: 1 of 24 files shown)


models/DataObject/ClassDefinition/Data/Checkbox.php
```diff
@@ -242,11 +244,10 @@ public function getForWebserviceExport($object, $params = [])
     /**
      * converts data to be imported via webservices
      *
-     * @deprecated
      * @param mixed $value
      * @param null|Model\DataObject\AbstractObject $object
      * @param mixed $params
-     * @param Model\Webservice\IdMapperInterface|null $idMapper
+     * @param $idMapper
```
blankse: This should be reverted. I think also the deprecated phpdocs and the isFilterable method.

weisswurstkanone: thx!

models/DataObject/ClassDefinition/Data/Checkbox.php
```diff
@@ -226,7 +229,6 @@ public function getFromCsvImport($importValue, $object = null, $params = [])
     }
 
     /**
-     * @deprecated
```
blankse: This should be also reverted

models/DataObject/ClassDefinition/Data/Checkbox.php
```diff
@@ -335,8 +337,10 @@ public function isEmpty($data)
         return $data === null;
     }
 
-    public function isFilterable(): bool
```
blankse: This should be also reverted

PR #5595, opened 2020-01-15
Only use active env to rebuild cache in system settings save action

solverat:
As discussed (a lot 😄), this PR only uses the current environment to clear the symfony cache after system settings has been updated. This will improve three things:

- **Performance**: only one cache container gets regenerated instead of two or three
- **Security**: there will be no "dev" container on production or any other environment but on dev only!
- **Stability**: if "prod" is not available in a given installation, this action will fail completely (`Uncaught InvalidConfigurationException: File config_prod.yml  cannot be found`)


dpfaffenbauer:
👍

brusch:
Let's see how it goes with that, if we see that there are any issues, we can think about providing a split button with multiple options ...

PR #5597, opened 2020-01-16
[Translation] Fix pluralization

### LINKED ISSUE #4979: [Feature request] Translation pluralisation support
# Feature Request

## Feature description

Consider strings such as *"%number reviews"* or "*Available in %count colours"* on a product page.

If *%number* is `1` then the translated string will not make sense in English. English is pretty straight-forward like this. But other Germanic language such as German (you know this better than me) can also change definite article going from "das *Noun*" to "die *Noune*" (with an -e suffix), thus making the problem even more glaring.

Pimcore already uses [Symfony translation](https://symfony.com/doc/current/translation.html), but the interface for shared translations does not expose the ability to translate sigular/plural strings.

This is complicated by the fact that languages also differ on the zero items count, where English, German, Ductch and so pluralises the noun, but other languages such as [Polish ](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#pl) and Ukrainian can have four different plural/singular forms depending in item count.

So, in summary, my feature request requires both a more expressive file format (likely the same as Symfony) for storing translation strings and a more complex UI. But I think it would be worth it long-term.

blankse:
Partly resolves #4979

It isn't possible now to use translation pluralization from symfony.

message.de.xlf
```
<?xml version="1.0"?>
<xliff version="1.2" xmlns="urn:oasis:names:tc:xliff:document:1.2">
    <file source-language="en" datatype="plaintext" original="file.ext">
        <body>
            <trans-unit id="1">
                <source>{0} Maximum number of participants reached.|{1} One registration possible.|]1,Inf[ %count% more registrations possible.</source>
                <target>{0} Maximale Teilnehmerzahl erreicht.|{1} Noch eine Anmeldung möglich.|]1,Inf[ Noch %count% Anmeldungen möglich.</target>
            </trans-unit>
        </body>
    </file>
</xliff>
```

template.html.twig
```
{{ '{0} Maximum number of participants reached.|{1} One registration possible.|]1,Inf[ %count% more registrations possible.'|trans({'%count%': count}) }}
```

## Additional info  
It would be nice to use always the trans method from symfony. So we can use the new pluralization format from symfony:
https://symfony.com/doc/4.4/translation/message_format.html

### DIFF
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -16,7 +16,6 @@ namespace Pimcore\Translation;
 
 use Pimcore\Cache;
 use Pimcore\Model\Translation\AbstractTranslation;
-use Pimcore\Model\Translation\TranslationInterface;
 use Pimcore\Tool;
 use Symfony\Component\HttpKernel\Kernel;
 use Symfony\Component\Translation\Exception\InvalidArgumentException;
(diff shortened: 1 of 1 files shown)


dvesh3:
@blankse Thanks for your PR but it is still not possible to use new pluralization format  from Symfony as our implementation is based on count parameter https://github.com/pimcore/pimcore/blob/c5066303f70abc6eb0db098f7b7832a944a53c63/lib/Translation/Translator.php#L99

blankse:
@dvesh3 Yes I know. It would be nice in 7.0 to change this ;) With my PR it is possible with the old way (count param)

dvesh3:
thank you :)

PR #5598, opened 2020-01-16
AdminStyle for Documents and Assets #5147

### LINKED ISSUE #5147: AdminStyle for Documents and Assets
# Feature Request
I was wondering whether we could add the admin style feature (currently available for data objects) to assets and documents as well.

## Feature description
As you can see in the screenshot below, we're using Pimcore's [Workflow Management](https://pimcore.com/docs/5.x/Development_Documentation/Workflow_Management/index.html). Now, we'd like to display an indicator, describing the workflow state so the user knows what elements are relevant without opening them.

image

The current workaround for this relies on the following events:
- `pimcore.admin.document.treeGetChildsById.preSendData`
- `pimcore.admin.asset.treeGetChildsById.preSendData`
- `pimcore.admin.dataobject.treeGetChildsById.preSendData`

However, when opening an element, then clicking on the `+` next to the node itself, the node is re-rendered. This causes the classes and, thus, the status indicator to be removed. At this point we didn't find a way to re-apply the classes in order to keep the status indicator.

Adding the possibility of an admin style, similar to the one for data objects, would resolve this issue.

weisswurstkanone:
resolves #5147


bundles/AdminBundle/Controller/Admin/Document/NewsletterController.php
```diff
@@ -89,6 +89,8 @@ public function getDataByIdAction(Request $request): JsonResponse
             'interfaces' => array_values(class_implements($email))
         ];
 
+        $this->addAdminStyle($email, Pimcore\Event\Admin\ElementAdminStyleEvent::CONTEXT_EDITOR, $data);
```
brusch: Let's use `use` and not the fully qualified class name, to make it a bit shorter ;)

bundles/AdminBundle/Resources/public/js/pimcore/document/hardlink.js
```diff
@@ -96,6 +96,14 @@ pimcore.document.hardlink = Class.create(pimcore.document.document, {
         var tabTitle = this.data.key;
         this.tabPanel = Ext.getCmp("pimcore_panel_tabs");
         var tabId = "document_" + this.id;
+
+        var iconClass;
```
brusch: Maybe we should move this code snippet to the abstract, as it is the same for every type.

doc/Development_Documentation/18_Tools_and_Features/02_Custom_Icons.md
```diff
@@ -0,0 +1,148 @@
+# Custom Icons & Tooltips for Documents/Assets & Data Objects
+
+Pimcore allows to dynamically define custom element icons & tooltips in the element tree. In addition, the icon of the editor tab can
+be changed.
+
+> Support for the event-based approach for Documents, Assets and Data Objects is available since release 4.6.1
```
brusch: `4.6.1` -> `6.4.1` 😉

models/Element/AdminStyle.php
```diff
@@ -32,12 +33,12 @@ class AdminStyle
     protected $elementIcon;
 
     /**
-     * @var string
+     * @var string string
```
brusch: just `string` is enough

models/Element/AdminStyle.php
```diff
@@ -32,12 +33,12 @@ class AdminStyle
...
      */
     protected $elementIconClass;
 
     /**
-     * @var array
+     * @var array sring
```
brusch: `sring` can go here

bundles/AdminBundle/Controller/Admin/Document/EmailController.php
```diff
@@ -119,12 +122,15 @@ public function saveAction(Request $request)
...
+                $this->addAdminStyle($page, ElementAdminStyleEvent::CONTEXT_EDITOR, $treeData);
+
                 return $this->adminJson([
                     'success' => true,
                     'data' => [
                         'versionDate' => $page->getModificationDate(),
                         'versionCount' => $page->getVersionCount()
-                    ]
+                    ],
+                    'treeDate' => $treeData
```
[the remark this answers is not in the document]
weisswurstkanone: thanks for the catch

bundles/AdminBundle/Controller/Admin/Document/EmailController.php
```diff
@@ -119,12 +122,15 @@ public function saveAction(Request $request)
                 $page->save();
                 $this->saveToSession($page);
 
+                $this->addAdminStyle($page, ElementAdminStyleEvent::CONTEXT_EDITOR, $treeData);
```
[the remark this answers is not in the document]
weisswurstkanone: no, because the style is used for updating the tree node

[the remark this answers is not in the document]
brusch: But why `CONTEXT_EDITOR` then, when it's used for the tree node? 🤔

bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
```diff
@@ -1561,6 +1569,25 @@ public function onKernelController(FilterControllerEvent $event)
         $this->_documentService = new Document\Service($this->getAdminUser());
     }
 
+    /**
+     * @param Model\Document $document
+     * @param null|int $context
+     * @param array $data
+     * @throws \Exception
+     */
+    protected function addAdminStyle(Document $document, $context = null, &$data) {
```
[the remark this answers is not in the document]
weisswurstkanone: done

bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
```diff
@@ -635,5 +634,14 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
...
         });
+    },
+
+    getIconClass: function () {
+        var iconClass;
+        if (this.data.iconCls) {
+            iconClass = this.data.iconCls;
+        } else if (this.data.icon) {
+            iconClass = pimcore.helpers.getClassForIcon(this.data.icon);
+        }
```
brusch: No return value?

brusch: We should move `getIconClass()` to `pimcore.element.abstract` then we can reuse it for assets.

bundles/AdminBundle/Resources/public/js/pimcore/asset/unknown.js
```diff
@@ -72,14 +72,22 @@ pimcore.asset.unknown = Class.create(pimcore.asset.asset, {
             items.push(this.workflows.getLayout());
         }
 
+        var iconClass;
```
brusch: That shouldn't be necessary, as this is one already here: 
https://github.com/pimcore/pimcore/blob/3b20298ddae3b9c905fdc9db3582c9098a50c363/bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js#L82
So in my opinion this file doesn't need any modification.

PR #5600, opened 2020-01-16
[Webservice] Add API key header option for REST

### LINKED ISSUE #4240: [Router] Remove logging of API Key of rest requests
router logs all requrest uris with something like that: 
```
[2019-04-15 16:55:51] request.INFO: Matched route "document_17". {"route":"document_17","route_parameters":{"_locale":"en_GB","_controller":"AppBundle:Content:default","_route":"document_17"},"request_uri":"http://academy.pim.zone/en/special-functions/productmodel","method":"GET"} []
```

therefore also API keys if they are in request uri ... 

we need to avoid that.

dvesh3:
Resolves #4240

### DIFF
--- a/bundles/AdminBundle/Security/Guard/WebserviceAuthenticator.php
+++ b/bundles/AdminBundle/Security/Guard/WebserviceAuthenticator.php
@@ -56,7 +56,12 @@ class WebserviceAuthenticator extends AbstractGuardAuthenticator implements Logg
      */
     public function getCredentials(Request $request)
     {
-        if ($apiKey = $request->get('apikey')) {
+        if ($apiKey = $request->headers->get('apikey')) {
+            // check for API key header
+            return [
+                'apiKey' => $apiKey
+            ];
+        } else if ($apiKey = $request->get('apikey')) {
             // check for API key parameter
             return [
                 'apiKey' => $apiKey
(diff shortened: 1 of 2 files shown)


NiklasBr:
The use of `X-` prefixes were [deprecated in 2012](https://tools.ietf.org/html/rfc6648):

>   Historically, designers and implementers of application protocols
>   have often distinguished between standardized and unstandardized
>   parameters by prefixing the names of unstandardized parameters with
>   the string "X-" or similar constructs.  In practice, that convention
>   causes more problems than it solves.  Therefore, this document
>   deprecates the convention for newly defined parameters with textual
>   (as opposed to numerical) names in application protocols.

It is not necessary in API specifications unless it is truly experimental, but since this is merged into the master branch I assume it is not experimental.

brusch:
Thanks for the info! The reason for preferring the `X-` prefix was the compatibility with all HTTP clients out there ... I personally don't care in the end, we can even support both as there are no negative effects to be expected.

NiklasBr:
Which clients cannot do headers not part of the [permanent set](https://www.iana.org/assignments/message-headers/message-headers.xhtml)?

It would perhaps be valuable to consider using the `Authorization: Basic [apiKey here]` or `Authorization: APIKEY [apiKey here]` header, because the API Key is effectively a usernameless password.

Do you consider the use of `X-API-Key` experimental or a regular/documented part of Pimcore Webservices? Because you now recommend the use of this header I would assume this is not experimental and should therefore avoid the prefix.

It doesn't really hurt either way, I'm mostly bikeshedding here, but I saw an an opportunity to maybe improve a small aspect. :)

PR #5609, opened 2020-01-17
Fix array key in relation editable docs

jdreesen:
The array key is correct in the PHP code example (few lines above) but wrong in the Twig example.

It's wrong in the docs of v6.x, too. Should I create a PR with a fix targeting the master branch or do you merge the branches somehow?


weisswurstkanone:
thx, I fixed it for 6.x as well

PR #5610, opened 2020-01-17
prevent invalid foreach argument if table has no rows

circular-code:
Speichern eines Objektes mit einer Tabelle ohne Zeilen führt zu fehlerhaftem Speichern.
Fehler wird nicht ans Frontend durchgereicht.

Resolves #

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Table.php
+++ b/models/DataObject/ClassDefinition/Data/Table.php
@@ -324,25 +324,30 @@ class Table extends Data implements ResourcePersistenceAwareInterface, QueryReso
     {
         $unserializedData = Serialize::unserialize((string) $data);
 
+        if ($unserializedData === null)
+            return $unserializedData;
+        
         //set array keys based on column configuration if set
         $columnConfig = $this->getColumnConfig();
-        if ($this->isColumnConfigActivated() && $columnConfig) {
-            $dataWithKeys = [];
-            foreach ($unserializedData as $row) {
-                $indexedRow = [];
-                $index = 0;
-                foreach ($row as $col) {
-                    $indexedRow[$columnConfig[$index]['key']] = $col;
-                    $index++;
-                }
-
-                $dataWithKeys[] = $indexedRow;
+        
+        if (!$this->isColumnConfigActivated() || !$columnConfig)
+            return $unserializedData;
+            
+        $dataWithKeys = [];
+        
+        foreach ($unserializedData as $row) {
+            $indexedRow = [];
+            $index = 0;
+            
+            foreach ($row as $col) {
+                $indexedRow[$columnConfig[$index]['key']] = $col;
+                $index++;
             }
 
-            return $dataWithKeys;
-        } else {
-            return $unserializedData;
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/Table.php
```diff
@@ -324,25 +324,30 @@ public function getDataFromResource($data, $object = null, $params = [])
     {
         $unserializedData = Serialize::unserialize((string) $data);
 
+        if ($unserializedData === null)
```
weisswurstkanone: how can this be null ?

brusch: @circular-code ping 🔔😊

brusch: @circular-code any feedback?

NiklasBr: I think it [should check for](https://www.php.net/manual/en/function.unserialize.php#refsect1-function.unserialize-notes) `false` rather than `null`?

weisswurstkanone: I am not even able to reproduce this.

NiklasBr: @weisswurstkanone `Serialize::unserialize()` will return `false` if it gets a malformed string or a nil value.

weisswurstkanone: @NiklasBr the first question is how it can be malformed. as I said, I am not experiencing this problem following the steps described above

circular-code: Hey guys, sry for the late reply.

The Problem occurs to me when I add fixed columns to my table type element and active the column configuration (see picture). When saving this state, it corrupts the class definition data, so when I later try to read the class definition everything below the table is missing.

image

weisswurstkanone: even then I am not able to reproduce on demo.pimcore.fun

circular-code: 1.	Create class A
2.	Add table T
3.	Set Columns to two
4.	Check Cols fixed
5.	Check Activate Column Configuration
6.	Save class A
7.	Create object AA of class A
8.	Open Database Admin and select data of object_store_<id of A>
9.	The T column contains the value „N;“ in the row of object AA

10.	DataObject\Concrete\Dao.php getById calls getData
11.	getData queries all field values from object_store_<id of A> and loops over fields
12.	each iteration calls $value->getDataFromResource, which for T calls the Table.php getDataFromResource function providing the stored „N;“
13.	Serialize::unserialize turns „N;“ to null
14.	foreach (null as $row) results in an exception
15.	The exception exits the foreach loop in Dao.php getData and is caught in Dao.php getById
16.	This way all fields which would have been processed after T are skipped

weisswurstkanone: Now I got you! Thanks for the detailed explanation 👍 And for the fix.

PR #5612, opened 2020-01-20
[Elements] Sanitize path via getValidKey()

BlackbitNeueMedien:
This PR enables Pimcore\Model\Element\Service::correctPath() to sanitize the path via getValidKey method of each path part. This allows querying objects like 
```php
Concrete::getByPath('/trailing-whitespace /abc')
```
This will find an object with path `/trailing-whitespace/abc`

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -224,7 +224,7 @@ class Asset extends Element\AbstractElement
      */
     public static function getByPath($path, $force = false)
     {
-        $path = Element\Service::correctPath($path);
+        $path = Element\Service::correctPath($path, 'asset');
 
         try {
             $asset = new Asset();
--- a/models/Asset/Service.php
+++ b/models/Asset/Service.php
@@ -277,7 +277,7 @@ class Service extends Model\Element\Service
      */
     public static function pathExists($path, $type = null)
     {
-        $path = Element\Service::correctPath($path);
+        $path = Element\Service::correctPath($path, 'asset');
 
         try {
             $asset = new Asset();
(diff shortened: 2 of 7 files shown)


brusch:
Hmm, should this be really possible? 
Shouldn't it return `null` instead?

BlackbitNeueMedien:
Also thought about that but what is the pursose of `correctPath()` then?

brusch:
Well just trimming and decoding if necessary, but `DataObject::getByPath('/trailing-whitespace /abc')` is imho a different story. But we can discuss to take away `correctPath()` as well, but at least the url decoding I'd keep.

BlackbitNeueMedien:
Ok, I know that it would be correct to return `null` from an engineer`s perspective. But from a user perspective it would be a nice feature if the key would get corrected.

I'd like to refer to Symfony console. This also tries to correct the command name - when there is only one possibility what you could have meant it simply executes the command, e.g.
`bin/console cache:clea` executes `bin/console cache:clear`

Concrete example where I need the correct path is in imports. Customers usually do not matter about trailing whitespaces or other things which get filteres out when saving Pimcore objects. Of course the developer is responsible to ensure data conversion to a correct format. 
I have no problem with this PR being closed. Was just a proposal to be more tolerant against user mistakes.

brusch:
Thanks! Let's see if there are other opinions on that, in the meanwhile I'll keep this PR open 😊

brusch:
Since there's no feedback so far, I'm going to close this PR for now. 
Feedback is of course still welcome.

PR #5613, opened 2020-01-20
[Data objects] Allow %_locale in data object preview URL

BlackbitNeueMedien:
Allow `%_locale` as variable in data object's preview URL. Of course you could use a link generator but often it is easier (and configurable via GUI) to enter a URL with placeholders.


weisswurstkanone:
thx!

PR #5614, opened 2020-01-20
Thumbnail config name should be copyable

BlackbitNeueMedien:
(Image / Video) Thumbnail config name should be copyable.


weisswurstkanone:
thx!

PR #5620, opened 2020-01-21
[Data objects] Open element when double-clicking relation item

BlackbitNeueMedien:
Originally I was annoyed of an advanced many-to-many relation with a lot of meta fields. When you want to open the related object you have to scroll to the complete right. This PR adds support for opening the related element via double-click.
To be consistent I also implemented this behaviour for all other relational types.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/abstractRelations.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/abstractRelations.js
@@ -194,6 +194,13 @@ pimcore.object.tags.abstractRelations = Class.create(pimcore.object.tags.abstrac
         }
 
         this.batchWin.close();
-    }
+    },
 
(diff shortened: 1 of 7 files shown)


weisswurstkanone:
what about reverseManyToManyObjectRelation ?

BlackbitNeueMedien:
Have not tested but I thought it would work because `pimcore.object.tags.reverseManyToManyObjectRelation` extends `pimcore.object.tags.manyToManyObjectRelation`
Have you tested, @weisswurstkanone - does it really not work?

brusch:
I'm just thinking ... since the handler function is always identical, should be move it to `pimcore.object.tags.abstractRelations` and just reference it on the implementations? 

e.g.
```js
listeners: {
    rowdblclick: this.gridRowDblClickHandler
}
```

Would save a lot of duplicate code.

brusch:
Thanks!

PR #5629, opened 2020-01-22
[Data objects] Minimizable modals for copying and deleting elements

BlackbitNeueMedien:
When copying or deleting elements in Pimcore backend GUI there is a progress bar modal with a backdrop which prevents any other interaction with the GUI. This PR makes this modal minimizable so the user is able to continue working in the GUI.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/elementservice.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/elementservice.js
@@ -141,6 +141,8 @@ pimcore.elementservice.deleteElementFromServer = function (r, options, button) {
                 text: t('initializing')
             });
 
+            var me = this;
+            var minimizedWindowListener;
             this.deleteWindow = new Ext.Window({
                 title: t("delete"),
                 layout:'fit',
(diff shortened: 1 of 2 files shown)


brusch:
Originally we made the progress windows modal intentionally. The idea was to prevent the user from doing anything that could cause any unexpected behavior in the UI while the system is processing jobs. In my opinion this still makes sense for some operations like deleting elements, but for the majority of tasks (like pasting, zipping, exporting, ...) this could be changed in general to a normal window. I'd prefer a smaller window in general for non-modal progress windows, so that there's no need to make it minimizable ... may implemented by just a draggable progressbar? 🤔

What do you think?

dpfaffenbauer:
Bildschirmfoto 2020-02-05 um 10 33 18

Maybe stick it to the bottom like the save confirm dialog?

BlackbitNeueMedien:
```The idea was to prevent the user from doing anything that could cause any unexpected behavior in the UI```
My current workaround when I do not want to wait for deletions to complete is to just open a new browser tab. Have you never deleted loads of objects and were disappointed about not being able to continue work in other areas?

brusch:
@dpfaffenbauer sounds good to me! 
@BlackbitNeueMedien actually not really, but I got your point 😊
To be honest, since deleting is a very critical operation I'm a bit worried about the possible side-effects. But starting with all the other operations would be a good start, wouldn't it? We can then decide about the deleting operation later on, after we have checked the possible side-effects in detail.

brusch:
@BlackbitNeueMedien 
Looks good in my opinion: 
image

brusch:
Closed, see #5914 😊

PR #5630, opened 2020-01-22
ObjectMetadata as return / param doctype for advanced many-to-many object relation getters / setters

BlackbitNeueMedien:
Analogous to advanced many-to-many relation, see https://github.com/pimcore/pimcore/blob/296fc17f0b00dc8f07926b5c34d127aa4ddb4f1b/models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php#L1103-L1106

Does anyone know a way to document getters / setters for meta columns of advanced relation fields?


weisswurstkanone:
thx!

PR #5638, opened 2020-01-22
[Data objects] Add listing filter methods for system fields

BlackbitNeueMedien:
This PR adds filter methods for system fields to generated data object listing classes, e.g. `filterById()` or `filterByKey()`.
I struggle a bit with `filterByPublished()` because it could lead to the false positive opinion that the following code will return unpublished objects:
```php
$listing = new Listing();
$listing->filterByPublished(false);
var_dump($listing->load());
```
In fact this will never return any object because it filters for unpublished objects but as long as `AbstractObject::getHideUnpublished()` returns `true` (this is the default value), this listing can only return published objects. Should we not add the `filterByPublished()` for that reason?


BlackbitNeueMedien:
Concerning `filterByPublished()`: Imho it is the same as 
```php
$listing = new Listing();
$listing->addConditionParam('o_published = 0');
var_dump($listing->load());
```
This will also never lead to any results as long as `AbstractObject::getHideUnpublished()` is `false`. So I see no problem being added by this PR.

models/DataObject/ClassDefinition.php
```diff
@@ -504,6 +504,40 @@ public function save($saveDefinitionFile = true)
...
+            $cd .= '* @param '.$systemFieldDataParamDoc."\n";
+            $cd .= '* @param '.$systemFieldOperatorParamDoc."\n";
+            $cd .= '* @return static'."\n";
+            $cd .= '*/' . "\n";
+
+            $cd .= 'public function filterBy' . ucfirst($fieldName) .' ($data, $operator = \'=\') {'."\n";
+            $cd .= "\t" . 'if(strpos($operator, \'?\') === false) {'."\n";
+            $cd .= "\t\t".'$operator .= \' ?\';'."\n";
+            $cd .= "\t}\n";
+            $cd .="\t" . '$listing->addConditionParam(\'`'.$systemField.'` \'.$operator, $data);' . "\n";
```
brusch: `$listing` is undefined in this context. 
I think it would make also sense to move this part
```php
	if(strpos($operator, '?') === false) {
		$operator .= ' ?';
	}
	$listing->addConditionParam('`o_key` '.$operator, $data);
```

to a helper function, so that we can apply changes without regenerating the class files.

BlackbitNeueMedien: Is ok now?

models/DataObject/Listing.php
```diff
@@ -211,4 +211,19 @@ public function addDistinct()
...
     }
+
+    /**
+     * @param string $field database column to use for WHERE condition
+     * @param string $operator SQL comparison operator, e.g. =, <, >= etc. You can use "?" as placeholder, e.g. "IN (?)"
+     * @param string|int|float|float|array $data comparison data, can be scalar or array (if operator is e.g. "IN (?)")
+     *
+     * @return static
+     */
+    public function filterByField($field, $operator, $data) {
```
brusch: Maybe we should call this method `addFilterByField()` and mark it as `@internal` to be consistent in naming?

brusch:
@BlackbitNeueMedien I'm just thinking ... we actually don't need to generate code, we just have to add the following to `\Pimcore\Model\DataObject\Listing\Concrete` that's it 😀


```php
/**
* Filter by path (system field)
* @param string|int|float|float|array $data  comparison data, can be scalar or array (if operator is e.g. "IN (?)")
* @param string $operator  SQL comparison operator, e.g. =, <, >= etc. You can use "?" as placeholder, e.g. "IN (?)"
* @return static
*/
public function filterByPath ($data, $operator = '=') {
	$this->addFilterByField('o_path', $operator, $data);
	return $this;
}

/**
* Filter by key (system field)
* @param string|int|float|float|array $data  comparison data, can be scalar or array (if operator is e.g. "IN (?)")
* @param string $operator  SQL comparison operator, e.g. =, <, >= etc. You can use "?" as placeholder, e.g. "IN (?)"
* @return static
*/
public function filterByKey ($data, $operator = '=') {
	$this->addFilterByField('o_key', $operator, $data);
	return $this;
}

/**
* Filter by id (system field)
* @param string|int|float|float|array $data  comparison data, can be scalar or array (if operator is e.g. "IN (?)")
* @param string $operator  SQL comparison operator, e.g. =, <, >= etc. You can use "?" as placeholder, e.g. "IN (?)"
* @return static
*/
public function filterById ($data, $operator = '=') {
	$this->addFilterByField('o_id', $operator, $data);
	return $this;
}

/**
* Filter by published (system field)
* @param string|int|float|float|array $data  comparison data, can be scalar or array (if operator is e.g. "IN (?)")
* @param string $operator  SQL comparison operator, e.g. =, <, >= etc. You can use "?" as placeholder, e.g. "IN (?)"
* @return static
*/
public function filterByPublished ($data, $operator = '=') {
	$this->addFilterByField('o_published', $operator, $data);
	return $this;
}

/**
* Filter by creationDate (system field)
* @param string|int|float|float|array $data  comparison data, can be scalar or array (if operator is e.g. "IN (?)")
* @param string $operator  SQL comparison operator, e.g. =, <, >= etc. You can use "?" as placeholder, e.g. "IN (?)"
* @return static
*/
public function filterByCreationDate ($data, $operator = '=') {
	$this->addFilterByField('o_creationDate', $operator, $data);
	return $this;
}

/**
* Filter by modificationDate (system field)
* @param string|int|float|float|array $data  comparison data, can be scalar or array (if operator is e.g. "IN (?)")
* @param string $operator  SQL comparison operator, e.g. =, <, >= etc. You can use "?" as placeholder, e.g. "IN (?)"
* @return static
*/
public function filterByModificationDate ($data, $operator = '=') {
	$this->addFilterByField('o_modificationDate', $operator, $data);
	return $this;
}
``` 

Can you update your PR again, please? 
Thanks a lot!

BlackbitNeueMedien:
Good catch! Just thought too complicated...

brusch:
Thanks a lot! 👍

PR #5642, opened 2020-01-22
Remove unused JavaScript code

jdreesen:
The variable `insertPoint` is not used anywhere else in the function in which it is declared.


brusch:
Good catch, thanks a lot!

PR #5646, opened 2020-01-22
Remove duplicate JavaScript object keys

jdreesen:
This PR removes some duplicate keys in JavaScript objects.

When there are duplicate keys in an object, the last one wins. 
Thus, I deleted the first one (if values are different).


brusch:
Thanks!

Issue #5648, opened 2020-01-23
[QuantityValue] BaseUnit dropdown is limited to 25 items


macghriogair:
# Bug Report

Our customer excessively uses Pimcore quantity units with conversion / base unit configuration. When there are more then 25 units, the base unit selection is broken. 

### Expected behavior

In QuantityValue panel:

Select a BaseUnit for a QuantityValue from *all* available units, not just first 25 items.

### Actual behavior

BaseUnit dropdown is loaded only once on page load and has an ExtJs default limit of 25 items.
Changing pagination / grid refresh has no effect. 

### Steps to reproduce

1. Having more than 25 Units. 

E.g. on a **TEST** system one could use following mysql procedure:

```sql

drop procedure if exists load_unit_test_data;

delimiter #
create procedure load_unit_test_data()
begin
set @i=0;
  truncate table `pimcore_test`.`quantityvalue_units`;

  start transaction;
  while @i < 50 do
    INSERT INTO `pimcore_test`.`quantityvalue_units` (abbreviation, longname)
    VALUES (
		concat('unit-',  lpad(@i, 3, '0')), concat('unit-',  lpad(@i, 3, '0') , ' long')
	);

    set @i=@i +1;
  end while;
  commit;
end #
```

2. In Pimcore, go to DataObjects > Quantity Value
3. Edit BaseUnit cell for any unit.
4. Dropdown shows only first 25 units:

![Bildschirmfoto 2020-01-23 um 10 04 52](https://user-images.githubusercontent.com/4060352/72974085-bfe16b00-3dce-11ea-906f-afdb98c2d940.png)

PR #5649, opened 2020-01-23
[QuantityValue] fix default limit on ExtJs BaseUnit store

macghriogair:
this PR fixes https://github.com/pimcore/pimcore/issues/5648

Background: Ext.data.JsonStore has a default pagination limit of 25, setting it to 0 disables it.


weisswurstkanone:
https://docs.sencha.com/extjs/6.2.0/classic/Ext.data.AbstractStore.html#cfg-pageSize

weisswurstkanone:
thanks a lot!

PR #5651, opened 2020-01-23
JavaScript: Remove unused renderer

jdreesen:

The renderer is overridden in the config object through the duplicate key in line 38. 

I'm not completely sure this is correct and if the custom one should be used instead of removing it, though.


weisswurstkanone:
thx!

PR #5652, opened 2020-01-23
JavaScript: Fix condition

jdreesen:
- `data` should be set if it **is not** `null` or `undefined`
- this check doesn't need to be repeated in the `else if` case


weisswurstkanone:
thanks a lot!

PR #5654, opened 2020-01-23
[Data objects] Allow DataObjects in the tree to be sorted in ascending and descending order #5540

### LINKED ISSUE #5540: Sort objects in tree from Z-A
Especially useful when objects are named by year/date to show latest date on top.

How it looks now
image

Suggested improvement
1) Rename "Key (Alphabetically)" to "A to Z"
2) Add option "Z to A"
3) Rename "Index (Ordered by Manually)" to "Manual"

=> Similar like Excel
image

mehlichmeyer:
Resolves #5540 

In the Data Object Tree you can now sort your DataObjects alphabetically in ascending or descending order.  

/e:
The DataObjects are now being sorted naturally, see Issue #5540


models/DataObject/AbstractObject.php
```diff
@@ -171,6 +172,11 @@ class AbstractObject extends Model\Element\AbstractElement
      */
     protected $o_childrenSortBy;
 
+    /**
+     * @var bool
+     */
+    protected $o_reverseSort;
```
dvesh3: please change property name to make it more clear and somehow consistent with o_childrenSortBy....maybe "o_childrenSortDesc" or "o_childrenSortOrder" with values 'ASC/DESC'

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -212,6 +217,22 @@ public function treeGetChildsByIdAction(Request $request, EventDispatcherInterfa
         return $this->adminJson($objects);
     }
 
+    private function sortChildrenNaturally(array $objects, string $sortBy, bool $reverse): array
```
dvesh3: phpdoc please

models/DataObject/AbstractObject.php
```diff
@@ -1443,4 +1449,24 @@ protected function getListingCacheKey(array $args = [])
...
+    {
+        $this->o_reverseSort = $o_reverseSort;
+
+        return $this;
+    }
+
+    /**
+     * @return bool
+     */
+    public function getReverseSort(): bool
```
dvesh3: we should also consider sort order in getChildren method here https://github.com/pimcore/pimcore/blob/c63c1b5e448503f6969bcbacd8f5ea83e9c01222/models/DataObject/AbstractObject.php#L412

dvesh3: also add here for nodes info https://github.com/pimcore/pimcore/blob/ae677a1de09d1c488cce9cea3ba95a3e44954ae0/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L230

mehlichmeyer: do you mean like: $list->setOrder($this->getChildrenSortOrder()); instead of $list->setOrder('asc');

dvesh3: yup, looks good 👍 also can you please add `'sortOrder' => $child->getChildrenSortOrder(),` here after SortBy https://github.com/pimcore/pimcore/blob/ae677a1de09d1c488cce9cea3ba95a3e44954ae0/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L230

bundles/InstallBundle/Resources/install.sql
```diff
@@ -369,7 +369,7 @@ CREATE TABLE `objects` (
   `o_classId` VARCHAR(50) NULL DEFAULT NULL,
   `o_className` varchar(255) DEFAULT NULL,
   `o_childrenSortBy` ENUM('key','index') NULL DEFAULT NULL,
-  `o_reverseSort` tinyint(1) UNSIGNED NULL DEFAULT '0',
+  `o_childrenSortOrder` VARCHAR(50) NULL DEFAULT NULL,
```
dvesh3: how about `o_childrenSortOrder ENUM('ASC','DESC') NULL DEFAULT NULL,`?

models/DataObject/AbstractObject.php
mehlichmeyer: I added the sortorder in the previous commit :)

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -216,19 +221,43 @@ public function treeGetChildsByIdAction(Request $request, EventDispatcherInterfa
         return $this->adminJson($objects);
     }
 
+    /**
+     * @param array $objects
+     * @param string $sortBy
+     * @param string $childrenSortOrder
+     * @return array
+     */
+    private function sortChildrenNaturally(array $objects, string $sortBy, string $childrenSortOrder): array
```
dvesh3: Hey @mehlichmeyer I tested your changes and found that getChildren() API call and tree hierarchy still have different children order, reason being that `sortChildrenNaturally()` is never called in `getChildren()`. I would suggest to move this method to `DataObject\Service` and then we call it in getChildren() also.
```php
public static function sortChildrenNaturally(array $objects, string $sortBy, string $childrenSortOrder): array
    {

        usort(
            $objects,
            function ($a, $b) use ($sortBy) {
                return strnatcasecmp($a->get($sortBy), $b->get('o_'. $sortBy));
            }
        );

        return $childrenSortOrder === 'ASC' ? $objects : array_reverse($objects);
    }
```

bundles/CoreBundle/Resources/translations/sv.json
```diff
@@ -692,7 +692,8 @@
     "login_token_no_password_error": "Användaren har inte ställt in lösenord.",
     "targeting_toolbar": "Targeting Toolbar",
     "sort_children_by": "Sortera barn efter",
-    "by_key": "Nyckel (Alfabetiskt)",
+    "by_key": "Nyckel (A till Z)",
+    "by_key_reverse": "Nyckel (Z till A)",
```
NiklasBr: *A till Ö* and *Ö till A* are the colloquial Swedish phrases that indicate sort order, even if ÅÄÖ may not actually be available in the possible valid keys.

brusch:
@mehlichmeyer please revert all the changes to the translations files, except the ones for English. Translations (except EN which is the master language) are managed here: https://pimcore.com/en/resources/translations

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
mehlichmeyer: Hi @dvesh3 could you check again? :)

models/DataObject/Service.php
```diff
@@ -1618,4 +1618,26 @@ public static function removeObjectFromSession($objectId)
...
+    public static function sortChildrenNaturally(array $objects, string $sortBy, string $childrenSortOrder): array
+    {
+        if('index' === $sortBy) {
+            $sortBy = 'idx';
+        }
+
+        usort(
+            $objects,
+            function ($a, $b) use ($sortBy) {
+                return strnatcasecmp($a[$sortBy], $b[$sortBy]);
```
dvesh3: Unfortunately, this will not work with getChildren() as it compares based on array keys(array generated for `DataObjectController::treeGetChildsByIdAction`) so instead 

1. please change the `sortChildrenNaturally` definition to something like mentioned here: 
https://github.com/pimcore/pimcore/pull/5654#discussion_r373507502
which compares based on the actual object properties e.g. `$a->get('o_key')` 

2. and then you also have to move `DataObject\Service::sortChildrenNaturally` call to this line in treeGetChildsByIdAction
https://github.com/pimcore/pimcore/blob/ac773a11cb5d15c08f14160c2ca96991c37c2f02/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L182

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
dvesh3: @mehlichmeyer please check my last comment. Thanks!

models/DataObject/Service.php
mehlichmeyer: Sorry, I previously misread your comment. Plese check again. :)

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -197,7 +201,8 @@ public function treeGetChildsByIdAction(Request $request, EventDispatcherInterfa
             'objects' => $objects,
         ]);
         $eventDispatcher->dispatch(AdminEvents::OBJECT_TREE_GET_CHILDREN_BY_ID_PRE_SEND_DATA, $event);
-        $objects = $event->getArgument('objects');
+
+        $objects = DataObject\Service::sortChildrenNaturally($event->getArgument('objects'), $object->getChildrenSortBy(), $object->getChildrenSortOrder());
```
dvesh3: I guess this one is not required anymore

dvesh3:
Hey @mehlichmeyer, we have an issue here :( natural sorting will not work properly if there is paginated list. In Objects tree, if children are paginated it will sort children naturally only on current page not on full list.

mehlichmeyer:
@dvesh3 I'll have a look!

brusch:
@mehlichmeyer any updates on this? Thanks a lot!

mehlichmeyer:
@brusch @dvesh3 sorry for the inactivity, but I am taking a look right now and hopefully update this PR sometime this week!

mehlichmeyer:
@dvesh3 The issue is that the pagination already takes place in the DataObject\Listing(), so another way to sort is not possible within the Listing itself. I don't have the time to debug the DataObject\Listing class and implement it there, but I can offer to implement a workaround which includes removing the limit and the offset from the listing, and then doing the pagination with array_chunk and calculating the page from the offset and the limit. What do you think?

dvesh3:
hello @mehlichmeyer we would not like to go with array_chunk workaround as it will have performance impacts and also breaks installations with hundreds of thousands items in one level. Also, I did some research to find a solution on SQL query level but couldn't find a decent one. So I would suggest to revert natural sorting changes and then we can merge your other changes. 

And we can create a new issue for natural sorting.

dvesh3:
Hey @mehlichmeyer any update on this? thanks!

dvesh3:
🏓 @mehlichmeyer

danielerne:
@dvesh3 @mehlichmeyer Any progress about this PR?

dvesh3:
@danielerne It's almost ready to be merged. Just waiting for @mehlichmeyer to revert natural sorting changes as per https://github.com/pimcore/pimcore/pull/5654#issuecomment-599961125.

brusch:
Seems that @mehlichmeyer isn't active anymore. 
Maybe someone is interesting in taking over this topic? 

Closed for now.

danielerne:
I removed natural sorting and cleaned it up. pr #6952

Issue #5301, opened 2019-11-20
RFC: Multi-class inheritance


BlackbitNeueMedien:
Often there is the challenge of wanting to be able to set default values for a subtree of other class objects. For example when you have a category and all products below this category have certain technical data in common. Currently you would have to add some "virtual" group product which gets the common product and below this the real products. Without this "group" product you would not be able to set the data once and inherit it to all products below.

#816 suggested multi-class object inheritance - but this approach was based on identical naming of fields for different classes. This seems not optimal to me as it mixes contexts and confuses users with fields which actually do not belong to the class itself.

What do you think of an extra tab in the object edit tab where you can set values for all other classes than the object's class itself? For example when you have a category object you could set attributes of the product class - in fact this would not change the category's object state but these values could be used for inheritance to Product objects below this category object.

I could imagine a similar UI like https://github.com/pimcore/output-data-config-toolkit - a separate tab on object editing listing all classes for each class an edit button which opens a window with the class definition. There you could set the value to be inherited to all objects of the specified class.

This is a "request for comments", so I appreciate any comment about this concept - be it positive or negative.

fashxp:
Dear community, it's been a while since that issue was updated the last time, so we decided to close it for now. If you're still interested in contributing, feel free to add comments here or contact us before creating a pull request to discuss next steps.

Thanks for your understanding.

PR #5658, opened 2020-01-23
[Data objects] Inheritance from "grandparents"

### LINKED ISSUE #5492: [Idea] AbstractObject ->getClosestParentOfType()
Hey there, happy holidays to everybody 🎄 

I think anyone who has been working with jQuery know the value of the method `closest(selector)` which allow the developer to go the nearest parent element that matches the selector. I hereby propose to add a somewhat similar method to the `AbstractObject` class, in the form of a new method that would allow developers to gain access to the closest parent of a certain type or class. 

Here's a concrete example:
Capture d’écran 2019-12-19 à 22 42 44

I have objects that are instances of `PrintingTechniques` (i.e. Screen Printing) and I use them as folders to store `PrintingOptions` (i.e. Press Type). I could create a `manyToOne` property called `technique` in the `PrintingOptions` class, restrict it to Objects of type `PrintingTechniques` and use the magic method `$printingOption->getTechnique()`. But this means that I have to ask the users to make sure that this field is correctly filled with respect to its parent Technique in the tree. 

I could also listen to events in the tree and programmatically update this field, but it feels like using a bazooka to kill a mosquito. 

Instead, I would rather use a method like :

```php
$printingOption->getClosestParentOfType(PrintingTechnique::class);
``` 

I noticed the `Concrete` class already implements a similar logic in the method `getNextParentForInheritance`. The proposed method `getClosestParentOfType` would simply call the getParent() until it the instanceof condition matches. 

By the way, I feel it would also resolve the issue raised here @dpfaffenbauer : https://github.com/pimcore/pimcore/issues/1944

What do you think? Do you think this could make its way through the core? 

Thank you

BlackbitNeueMedien:
When you have an object hierarchy it sometimes is useful to be able to get the parent of a certain class, see #5492 
Example: 
You want to get the category of a product. 
You have the tree hierarchy: category > master product > slave product. 
Not all products have slaves so you have to loop through the parents of a product until you get a category object. It would be easier via `$product->getClosestParentOfClass(Category::classId)`.

With the current implementation of this PR (as proposed by @romainbessugesmeusy in #5492) there is a side effect:
Currently objects can only inherit data from their parent element if this parent element is of the same class (or if you have some folders between, e.g. Product > Folder > Product). This PR adds the option to have objects of other classes between, e.g. News > Category > News and nevertheless inherit data from the first ancestor of the same class.

I wonder if this side effect is good or not and look forward to a discussion.
When we decide to not want this "grandparent" inheritance we can simply implement `Concrete::getNextParentForInheritance()` as it was before and nevertheless have the new `Concrete::getClosestParentOfClass()`.

Resolves #5492


models/DataObject/Concrete.php
```diff
@@ -594,9 +594,16 @@ public function getValueFromParent($key, $params = null)
      */
     public function getNextParentForInheritance()
     {
+        return $this->getClosestParentOfClass($this->getClassId());
+    }
+
+    /**
```
weisswurstkanone: @param string $classId

BlackbitNeueMedien: Added it. 
In my own projects I do not add param / return doctypes when this is already present via typehints. Is there a general rule for Pimcore to always add param / return doctypes - also if they do not add any value?

brusch:
In my opinion your proposed approach is the one someone would expect, so I'd go for it. 
But since the behavior can change, I'd definitely put a upgrade note here: https://github.com/pimcore/pimcore/blob/d26da42019bd0b1200e1f9492b96d27b4b8e73f2/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md#L4 (scheduled for 6.5.0). 
Thanks!

weisswurstkanone:
can you please add a simple test as well if it is not too much effort ?

BlackbitNeueMedien:
@weisswurstkanone Added a test but actually I do not know how to run it. Let's see if Travis can execute it... 
How can I get the files in `_output/var/classes/DataObject` locally? Or more general: Is there an instrcution how to run the tests?

weisswurstkanone:
Have a look at https://pimcore.com/docs/6.x/Development_Documentation/Development_Tools_and_Details/Testing/Core_Tests.html

BlackbitNeueMedien:
Test runs successfully.

weisswurstkanone:
@BlackbitNeueMedien 
"...When we decide to not want this "grandparent" inheritance we can simply implement Concrete::getNextParentForInheritance() as it was before and nevertheless have the new Concrete::getClosestParentOfClass()..."

I think it is not that easy. You have to make InheritanceHelper (it is responsible for propagating the query table entries down) make aware of the fact that you have your own implementation of getNextParentForInheritance.
So simply overwriting it is not sufficient.

Actually your change is even a bugfix because InheritanceHelper always behaved the way you are expecting it now :-)

mc-supply:
Just for my clarification, is the implementation as proposed by @BlackbitNeueMedien supposed to be in v6.5.2?  It still seems to be the original implementation with the limitations mentioned.

BlackbitNeueMedien:
Yes, it is merged in master, see https://github.com/pimcore/pimcore/blob/master/models/DataObject/Concrete.php#L565-L592

mc-supply:
I could be wrong, but that seems to be the implementation from @romainbessugesmeusy, where it only allows the same parent type or folder?  I wasn't on the original message, so I'm kinda reverse-engineering a few different conversations.

BlackbitNeueMedien:
Nope, it is this line which allows inheritance of same-class objects also if there are other class' objects between:
https://github.com/pimcore/pimcore/blob/12ca8e1fe84527111e54812ee3b2b6d90137fb5d/models/DataObject/Concrete.php#L579

mc-supply:
Gotcha.  My particular application works better with this:  https://github.com/pimcore/pimcore/pull/5658/commits/666411769757d50ddee99a95ae08771d70746298

What I have is a top-most parent of a different type (manufacturer), with products of type product underneath it.  I have a listener that looks to that top-most manufacturer parent to get the brand if the product manufacturer is blank in the product and not inherited from a parent product (mostly useful when adding a new top-level product directly under a manufacturer).  It might be so specific that others would not encounter the issue?

BlackbitNeueMedien:
I do not see your problem. What does 666411769757d50ddee99a95ae08771d70746298 do better in your opinion?

mc-supply:
It looks as though the main issue is not line 579, it is that:

https://github.com/pimcore/pimcore/blob/master/models/DataObject/Concrete.php#L585-L587 then removes the ability for a different parent class to be returned.  So since my parent type is a different type, null is returned.  Correct?

BlackbitNeueMedien:
This is correct because inheritance does only work with objects f the same class. You want to inherit a field with the same name but from another class, do you? I have written down some thoughts about that in #5301

mc-supply:
Right, so the way I'm getting around that is using event listeners to manually reach up the chain and grab that value.  To me, that defeats the whole purpose of the getClosestParentOfClass function and the addition of $parent->getClassId() !== $classId, cause that $parent object will never be returned.  I agree with the comments in https://github.com/pimcore/pimcore/issues/5301, however.  That would be a nice feature.

BlackbitNeueMedien:
getClosestParentOfClass's purpose is to find an object of the given class in the current object's ancestors.
For example when you have Manufacturer > Master product > Slave product you can call `$slaveProduct->getClosestParentOfClass(Manufacturer::classId);` and get the manufacturer object. Multi-class inheritance is a complete different story. And yes, that would be a nice feature (and we would like to have that as of yesterday ;-) ) but it is also a lot of work. Simply trying to inherit fields with the same name from different classes does not seem to be the right way imho.

And I can also not understand `the addition of $parent->getClassId() !== $classId, cause that $parent object will never be returned` because that would mean that inheritance does not work at all - but actually it does (for objects of the same class).

Perhaps I do not understand what you mean. Could you please write a little step-by-step instruction or a little script?

mc-supply:
Your example is actually precisely my situation, but the problem is that Manufacturer object will never be returned (cause 585 requires that the parent is of the same class as the calling class).

BlackbitNeueMedien:
Of course not - and this is by purpose as Pimcore currently only supports single-class inheritance. Please create a separate issue or even a PR if you have an idea how to support multi-class inheritance.

Issue #4473, opened 2019-05-31
If no tag exists, it is not possible to add a first tag.


Yivan:
Hello,

It seems tags have been slighlty modified (no more tag id 0 in database).
Result is if there is no tag created, enduser in the back-end GUI can no more create a first tag.

Please see the ecommerce demo and open the tag manager:

![tags](https://user-images.githubusercontent.com/9025839/58710969-cba29c00-83bd-11e9-9087-f437484d6eb3.png)

It is the same problem if you delete the existing tags in the basic demo.

Thanks.

Issue #4195, opened 2019-04-07
Tags should not be loaded while opening any element(object, asset, document)


psuhane:
My system have more that 200k tags and all tags are loading on each element click on tree. Tags should only be loaded when clicking on Tags tab.

Please refer following screen
![screenshot-ec2-3-120-129-78 eu-central-1 compute amazonaws com-2019 04 07-16-12-39](https://user-images.githubusercontent.com/12061880/55682376-5207a800-5950-11e9-817e-0544aaf955d5.png)

PR #5663, opened 2020-01-23
[Tags] server side filtering

### LINKED ISSUE #4725: Tag filter in assets search
# Bug Report

### Expected behavior
Being able to search for tags lower than the 5th level.

### Actual behavior
ext-all.js throwing an error in the console and no results are shown.
```
too much recursion
```
Furthermore the admin site has to be refreshed when closing the "Search Asset"-panel after the error occured, because the admin site isn't usable afterwards.
The same error occurs when using the "Search Documents" or "Search Objects"-panel.

### Steps to reproduce
1. create tags down to the 5th level:
grafik
2. open the "Filter for Tags" in "Search Assets"-panel
3. search for the tag on the 5th level:
grafik

**Edit:** I just noticed that the error only occurs when the tag-tree wasn't opened before. And it fails when opening the 3rd level.

dvesh3:
Resolves #4725

also considered https://github.com/pimcore/pimcore/issues/4195 https://github.com/pimcore/pimcore/issues/4473

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/TagsController.php
+++ b/bundles/AdminBundle/Controller/Admin/TagsController.php
@@ -108,6 +108,7 @@ class TagsController extends AdminController
         $assginmentCId = intval($request->get('assignmentCId'));
         $assginmentCType = strip_tags($request->get('assignmentCType'));
 
+        $recursiveChildren = false;
         $assignedTagIds = [];
         if ($assginmentCId && $assginmentCType) {
             $assignedTags = Tag::getTagsForElement($assginmentCType, $assginmentCId);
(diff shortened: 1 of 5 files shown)


bundles/AdminBundle/Controller/Admin/TagsController.php
```diff
@@ -125,9 +126,30 @@ public function treeGetChildrenByIdAction(Request $request)
         }
         $tagList->setOrderKey('name');
 
+        if(!empty($request->get('filter'))) {
+            $filterIds = [0];
+            $filterTagList = new Tag\Listing();
+            $filterTagList->setCondition('`name` LIKE "%'.$request->get('filter').'%"');
```
weisswurstkanone: SQL injection possible. Please properly quote the filter.
Example: http://pimcore-demo-basic.pim.zone/admin/tags/tree-get-children-by-id?_dc=1580295370027&showSelection=false&assignmentCId=&assignmentCType=&node=0&filter=%22;%20DROP%20TABLE%20objects;%20SELECT%20*%20FROM%20users%20WHERE%20name%20LIKE%20%22

will drop the objects table

PR #5664, opened 2020-01-24
[Asset] get not modified fileextension for Assets.

putzflorian:
In case when a filename with uppercase filextensionn like JPG you run in an error.


brusch:
Thanks for your PR! 
I've noticed that the replace function is no specific enough, so I've replaced it by `preg_replace()` which solves this issue as well 😊

PR #5679, opened 2020-01-27
[Ecommerce] Log and display index update errors

chilladelia:
Log and display the detailed error message, e.g., when running the bootstrap command.


fashxp:
thx

PR #5680, opened 2020-01-27
Do not use master for downloads

Cruiser13:
It will download Pimcore 6 instead of Pimcore 5. I used 1.3.0 as pimcore.phar does use the same now.


brusch:
Thanks a lot!

PR #5684, opened 2020-01-28
[PHPStan] level 2 fixes

weisswurstkanone:


bundles/AdminBundle/Controller/Rest/Element/DataObjectController.php
```diff
@@ -20,6 +20,7 @@
 use Pimcore\Model\DataObject;
 use Pimcore\Model\Webservice\Data\DataObject\Concrete\In as WebserviceObjectIn;
 use Pimcore\Model\Webservice\Data\DataObject\Concrete\Out as WebserviceObjectOut;
+use Pimcore\Model\Webservice\Data\DataObject\Concrete\Out;
```
jdreesen: This is `use`d twice. See also line 133.

weisswurstkanone: thx! 👍

bundles/AdminBundle/Controller/Rest/Element/DataObjectController.php
```diff
@@ -125,10 +125,11 @@ public function getAction(Request $request, $id = null)
             $stopwatch->start('ws', $profileName);
         }
 
-        /** @var WebserviceObjectOut|WebserviceFolderOut $out */
         if ($object instanceof DataObject\Folder) {
+            /** @var WebserviceObjectOut|WebserviceFolderOut $out */
             $out = $this->service->getObjectFolderById($id);
         } else {
+            /** @var WebserviceObjectOut|Out $out */
```
jdreesen: Not that the `use` statement is removed, this should be removed, too.

```suggestion
            /** @var WebserviceObjectOut $out */
```

models/Document/Tag/Renderlet.php
```diff
@@ -177,6 +177,7 @@ public function frontend()
                 Logger::error($e);
             }
         }
+        return null;
```
brusch: There's actually no return value expected, so `void` would be best I'd say.

models/Document/Tag/Snippet.php
```diff
@@ -177,6 +177,7 @@ public function frontend()
                 return 'ERROR: ' . $e->getMessage() . ' (for details see log files in /var/logs)';
             }
         }
+        return null;
```
brusch: Here `void` as well.

models/Document/Tag/Wysiwyg.php
```diff
@@ -203,5 +203,6 @@ public function rewriteIds($idMapping)
 
         $html->clear();
         unset($html);
+        return null;
```
brusch: Also no return value expected here, so `void`;

PR #5696, opened 2020-01-29
[DataObjects] Layout Component - Dynamic Text refactoring

### LINKED ISSUE #5599: Containerize Text Renderers
https://github.com/pimcore/pimcore/blob/master/models/DataObject/ClassDefinition/Layout/Text.php#L135

dvesh3:
Resolves #5599


models/DataObject/ClassDefinition/Layout/Text.php
```diff
@@ -126,15 +126,15 @@ public function setBorder(bool $border): void
...
-                $context['fieldname'] = $this->getName();
-                $context['layout'] = $this;
-                $result = call_user_func($renderingClass . '::renderLayoutText', $this->renderingData, $object, $context);
-                $this->html = $result;
-            }
+        $renderingClass = Model\DataObject\ClassDefinition\Helper\DynamicTextResolver::resolveRenderingClass(
+            $this->getRenderingClass()
+        );
+
+        if (method_exists($renderingClass, 'renderLayoutText')) {
```
weisswurstkanone: maybe instanceof DynamicTextLabelInterface in addition (keep method_exists for BC reasons).

models/DataObject/Service.php
```diff
@@ -946,7 +946,10 @@ public static function getValidLayouts(Concrete $object)
         }
 
         foreach ($list as $customLayout) {
-            $resultList[$customLayout->getId()] = $customLayout;
+            if($customLayout instanceof ClassDefinition\CustomLayout) {
```
dpfaffenbauer: doesn't belong to this PR, does it?

models/DataObject/ClassDefinition/Layout/Text.php
brusch: Good point, I'll add it.

dpfaffenbauer:
did you test if that is also backward-compatible? Cause the Class-Resolver makes a "new $class" if it doesn't come from DI. AFAIK currently the layout used a static function

models/DataObject/Service.php
dvesh3: you right this does not belong to this PR....my bad! it was miss :/

brusch: But doesn't hurt, so i kept it.

dvesh3:
@dpfaffenbauer yes it backward compatible... please see https://github.com/pimcore/pimcore/pull/5696/files#diff-f60d77aacb93b63df6601aa3bb681637R136

dpfaffenbauer:
@dvesh3 right, just checked `call_user_func` works with static functions on instances without notices. (https://www.php.net/manual/en/function.call-user-func.php#example-6382)

brusch:
Added a deprecation warning and fixed the example in the docs.

PR #5698, opened 2020-01-30
re-add clipboard to ckeditor

### LINKED ISSUE #5151: Removal of ckeditor-plugin "clipboard" (see #3831) breaks advanced content filter
In #3831, ckeditor's "clipboard"-plugin was removed. There is no need to remove the plugin just because the paste-buttons don't work anymore in modern browsers.
The main problem with this is that ckeditor needs the "clipboard"-plugin to trigger it's advanced content filter (acf). This is why even the most basic preset for any ckeditor configuration has this plugin enabled.

**The "clipboard"-plugin needs to be re-added to ckeditor.**

**Edit:** to clarify - the acf still works, because on reopening a document with a wysiwyg that contains unallowed styles, the content gets properly filtered.
However, if you just paste content to the wysiwyg and then save, the content is NOT being filtered (but it should).

### Expected behavior

Pasting any formatted text into an wysiwyg-editor triggers ckeditors acf. Tihs filter ist enabled by default and removes any unwanted formatting, ie. styles.

### Actual behavior
Pasting formatted text keeps any unwanted styles.

### Steps to reproduce
1. Copy the colorful line of text from https://ckeditor.com/docs/ckeditor4/latest/examples/acf.html in any wysiwyg-editor in pimcore.

solverat:
This PR:

- fixes #5151
- Updates ckeditor to CKEditor 4.13.1


brusch:
Grazie Mille 👌🙏

M-Ali-Aziz:
Updated to Pimcore 6.5.2 and still same problem :(
https://talk.pimcore.org/t/ckeditor-clipboard-plugin-missing-on-pimcore-6-2-x-6-3-x/2690

PR #5699, opened 2020-01-30
add doctrine filter to RecycleBin creation.

dpfaffenbauer:


dpfaffenbauer:
BTW: The behaviour is exactly the same as with Versions, why not re-use that? 

Currently it is copy paste stuff and leads to errors, like this already does...

albertmueller:
i can confirm, that with this PR the doctrine issue is fixed

brusch:
@dpfaffenbauer Thanks for your PR! 
Actually it is not exactly the same behavior (e.g. the recycling bin includes children, ...), that's why we're not sharing the code with versions and providing dedicated interfaces 😊

brusch:
Related to #5423

dpfaffenbauer:
can we expect a new patch release? or is this bug-fix coming with a minor release?

brusch:
Is coming with 6.5.0 next week.

PR #5700, opened 2020-01-30
[Advanced object relation] Removed hard-coded visible field "ID"

BlackbitNeueMedien:
Currently the column `ID` always appears in an advanced many-to-many object relation - ignoring if this has been set as visible field. 
If you have set `ID` as visible field it appears twice in the object edit panel.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
@@ -108,7 +108,6 @@ pimcore.object.tags.advancedManyToManyObjectRelation = Class.create(pimcore.obje
         var visibleFields = this.visibleFields;
 
         var columns = [];
-        columns.push({text: 'ID', dataIndex: 'id', width: 50});
 
         for (i = 0; i < visibleFields.length; i++) {
             if (!empty(this.fieldConfig.visibleFieldDefinitions) && !empty(visibleFields[i])) {
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
@@ -42,10 +42,6 @@ pimcore.object.tags.advancedManyToManyObjectRelation = Class.create(pimcore.obje
         }
 
         var fields = [];
-        if (typeof this.fieldConfig.visibleFields != "string") {
-            this.fieldConfig.visibleFields = "";
-        }
-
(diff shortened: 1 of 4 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
```diff
@@ -108,7 +108,6 @@ pimcore.object.tags.advancedManyToManyObjectRelation = Class.create(pimcore.obje
         var visibleFields = this.visibleFields;
 
         var columns = [];
-        columns.push({text: 'ID', dataIndex: 'id', width: 50});
```
weisswurstkanone: are there any noticable side effects if there are no columns at all (i.e. user error) :-)

BlackbitNeueMedien: I tested it. I do not see any problems - does not look beautiful but is exactly what is to be expected when there are neither visible fields nor meta fields:
Bildschirmfoto 2020-01-30 um 11 36 30
I added 1 object in this case.

Or should we implement a fallback if the edit grid has no columns then at least show ID (or path)?

weisswurstkanone: Fallback sounds reasonable because otherwise you cannot distinguish the rows.

BlackbitNeueMedien: Copied fallbacks from many-to-many object relation (except class because advanced m2m object relations can only have one class).

weisswurstkanone: so now it is always there ?

BlackbitNeueMedien: ??? No, the fallback fields get hidden of visibleFields are set:
`{text: 'ID', dataIndex: 'id', width: 50, hidden: !!visibleFields.length},`

The interesting part is `hidden: !!visibleFields.length`

weisswurstkanone: But hidden means that you can make it visible again using the column dropdown.
Is it what we really want ?

image

BlackbitNeueMedien: Ok, then it is also wrong for many-to-many object relation ;-)

I will change the PR to only add those fallback columns if no visible fields have been set...

BlackbitNeueMedien: Now the fallback columns are only added if no visible fields have been set.

weisswurstkanone:
thanks a lot

PR #5709, opened 2020-02-03
[Glossary] Deprecate Acronyms and move them to Abbr (Sucessor)

### LINKED ISSUE #5705: [Glossary] Change "acronym" (obsolete) to "abbr"
# Feature Request

## Feature description

the glossary produces HTML elements with "acronym" (obsolete) instead of "abbr"

Link: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/acronym

### Expected behavior
`<abbr class=pimcore_glossary title="Some Text">Text</abbr>`
### Actual behavior
`<acronym class=pimcore_glossary title="Some Text">Text</acronym>`
### Code example
```php
// My simple Code Sample
 $this->glossary()->start();

?>
<div class="content">
        Text - and all the Nice Features in my World
</div>
<?php
$this->glossary()->stop();
```
Thx ... Niki

brusch:
Resolves #5705

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/glossary.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/glossary.js
@@ -60,7 +60,7 @@ pimcore.settings.glossary = Class.create({
             '/admin/settings/glossary?',
             [
                 'id', {name: 'text', allowBlank: false}, 'language', 'casesensitive', 'exactmatch',
-                'site', 'link', 'acronym', 'creationDate', 'modificationDate'
+                'site', 'link', 'abbr', 'creationDate', 'modificationDate'
             ],
             itemsPerPage
         );
@@ -100,8 +100,6 @@ pimcore.settings.glossary = Class.create({
             {text: t("link"), flex: 200, sortable: true, dataIndex: 'link', editor: new Ext.form.TextField({}),
                                 tdCls: "pimcore_droptarget_input"},
             {text: t("abbr"), flex: 200, sortable: true, dataIndex: 'abbr', editor: new Ext.form.TextField({})},
-            {text: t("acronym"), flex: 200, sortable: true, dataIndex: 'acronym',
-                                editor: new Ext.form.TextField({})},
             {text: t("language"), flex: 50, sortable: true, dataIndex: 'language', editor: new Ext.form.ComboBox({
                 store: this.languages,
                 mode: "local",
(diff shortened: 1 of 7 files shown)


NiklasBr:
This is a BC break, right? I mean, this should be done, but there was no documentation and still no clear guidelines which BC breaks are acceptable and which are not.

brusch:
Shouldn't break under normal conditions, only if: 
- you're accessing directly the database
- your're accessing the public property `$acronym`

brusch:
The change is documented in the upgrade notes.

NiklasBr:
This directly affects frontend HTML `{% pimcoreglossary %} ...`.

There is still no documentation on acceptable BC breaks, which is *very* frustrating. The fact that all these BC breaks are allowed to creep into the code makes it a huge hassle to upgrade Pimcore.

brusch:
Well yes, it uses a different HTML tag, is this really that big problem that we can't change it with a minor release?

NiklasBr:
The problem is the very fact that you allow BC breaks to begin with, then the problem escalates when the rules you implement for *some* PR:s seem very arbitrary.

Example: About a month ago you denied a PR specifically because of minor HTML frontend changes. Now you allow it. Please publish those promised rules soon. Which are breaks acceptable?

brusch:
No further development wanted by @NiklasBr - thanks - closed. 🤷‍♂️

dpfaffenbauer:
what is always the hurry with BC breaks in minor released? why not wait for 7.0?

NiklasBr:
@dpfaffenbauer be careful, you might be labelled png. ;)

dpfaffenbauer:
@NiklasBr png?

NiklasBr:
That's an abbrev... or maybe an acronym for \<abbr title="persona non grata">unwelcome person\</abbr>. 😅

dpfaffenbauer:
Well, pretty sure some already see me that way. 

The thing just is, this are the things I stumble upon in a couple of weeks, then I have to research that just to see that it is an unnecessary/avoidable BC break. And it hasn't to be like that. If you just follow semver properly, everything is fine. 

But however...  I might just need to stop caring too much 🤷‍♂️

jdreesen:
Well, it's very annoying to even be afraid of patch updates, because you never know what will break due to unexpected/undocumented BC breaks (we had this several times in v5.x and it's sometimes hard to spot). So it would be great if Pimcore would care a bit more about these things, because not updating is also not really an option.

PR #5716, opened 2020-02-04
[Ecommerce][IndexService] add pre-send-data event in FilterDefinition Object

jremmurd:
This PR makes it possible to adjust the displayed value in the FilterDefiniton object.

In the screenshot the actual value is in brackets.

image

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Controller/IndexController.php
+++ b/bundles/EcommerceFrameworkBundle/Controller/IndexController.php
@@ -14,8 +14,9 @@
 
 namespace Pimcore\Bundle\EcommerceFrameworkBundle\Controller;
 
-use Pimcore\Event\Ecommerce\IndexServiceEvents;
+use Pimcore\Event\Ecommerce\AdminEvents;
 use Symfony\Component\EventDispatcher\GenericEvent;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Pimcore\Bundle\AdminBundle\Controller\AdminController;
 use Pimcore\Bundle\EcommerceFrameworkBundle\Factory;
 use Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\ProductList\ProductListInterface;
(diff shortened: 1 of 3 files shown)


lib/Event/Ecommerce/IndexServiceEvents.php
```diff
@@ -33,4 +33,13 @@ final class IndexServiceEvents
...
     const GENERAL_PREPROCESSING_ERROR = 'pimcore.ecommerce.indexservice.generalPreProcessingError';
+    
+       /**
+     * Fired when values in filter definition get fetched
+     *
+     * @Event("Symfony\Component\EventDispatcher\GenericEvent")
+     *
+     * @var string
+     */
+    const GET_VALUES_FOR_FILTER_FIELD_PRE_SEND_DATA = 'pimcore.ecommerce.indexservice.getValuesForFilterFieldPreSendData';
```
brusch: Since this event isn't related to the index service, I don't think this is the right place for it. 
I think a dedicated class would make sense here, e.g. `\Pimcore\Event\Ecommerce\AdminEvents`.

bundles/EcommerceFrameworkBundle/Controller/IndexController.php
```diff
@@ -92,6 +94,10 @@ public function getValuesForFilterFieldAction(Request $request)
                 $data = $helper->getGroupByValuesForFilterGroup($columnGroup, $productList, $request->get('field'));
             }
 
+            $event = new GenericEvent(null, ["data" => $data, "field" => $request->get('field')]);
+            \Pimcore::getEventDispatcher()->dispatch($event, IndexServiceEvents::GET_VALUES_FOR_FILTER_FIELD_PRE_SEND_DATA);
```
brusch: Please use DI to get the event-dispatcher (type hint).

jremmurd: done

lib/Event/Ecommerce/IndexServiceEvents.php
jremmurd: done

brusch:
Just made some smaller changes in 15732707626f653c46b9a95818389809bf561411 - now it should be ready to be merged, but let's wait what Travis has to say 😉

PR #5717, opened 2020-02-04
[Travis] Add PHPStan level 2 job but allow it to fail

weisswurstkanone:


blankse:
It would be possible to switch to level 2 and generate a baseline file with the current errors. So we get failed tests when someone generate new errors. See https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff

weisswurstkanone:
Sounds interesting. Is this something you could look at?
Nevertheless, we should try get rid of the current errors and the list is still quite long.

PR #5720, opened 2020-02-05
[PHPSTAN] Fix: "Method XYZ should return XYZ" errors

blankse:
This PR fix all "Method XYZ should return XYZ" PHPSTAN level 2 errors.

Before
```
 [ERROR] Found 1026 errors 
```
After
```
 [ERROR] Found 901 errors
```

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Model/AbstractOrderItem.php
+++ b/bundles/EcommerceFrameworkBundle/Model/AbstractOrderItem.php
@@ -129,8 +129,7 @@ class AbstractOrderItem extends \Pimcore\Model\DataObject\Concrete
      */
     public function getTotalNetPrice()
     {
-        //prevent throwing an exception for backward compatibility
-        Logger::err('getTotalNetPrice not implemented for ' . get_class($this));
+        throw new UnsupportedException('getTotalNetPrice is not implemented for ' . get_class($this));
     }
 
     /**
@@ -140,8 +139,7 @@ class AbstractOrderItem extends \Pimcore\Model\DataObject\Concrete
      */
     public function setTotalNetPrice($totalNetPrice)
     {
-        //prevent throwing an exception for backward compatibility
-        Logger::err('setTotalNetPrice not implemented for ' . get_class($this));
+        throw new UnsupportedException('setTotalNetPrice is not implemented for ' . get_class($this));
     }
 
     /**
@@ -151,8 +149,7 @@ class AbstractOrderItem extends \Pimcore\Model\DataObject\Concrete
      */
     public function getTaxInfo()
     {
-        //prevent throwing an exception for backward compatibility
-        Logger::err('getTaxInfo not implemented for ' . get_class($this));
+        throw new UnsupportedException('getTaxInfo is not implemented for ' . get_class($this));
     }
 
     /**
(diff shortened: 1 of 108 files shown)


bundles/EcommerceFrameworkBundle/Model/AbstractOrderItem.php
```diff
@@ -129,8 +129,7 @@ public function setTotalPrice($totalPrice)
      */
     public function getTotalNetPrice()
     {
-        //prevent throwing an exception for backward compatibility
-        Logger::err('getTotalNetPrice not implemented for ' . get_class($this));
+        throw new UnsupportedException('getTotalNetPrice is not implemented for ' . get_class($this));
```
brusch: This changes are BC breaks, we should put a `@TODO` here for v7.

blankse: I add it to the Upgrade Notes. It is not possible for v6.5.0?

brusch: Better not, I'm tired of those discussions 😉

dpfaffenbauer: ;)

blankse: Okay 😄 I will revert this.
How should we fix the error here? Should I return 0? Or should we stay this error till v7?

brusch: What about 
```php
    /**
     * @return void
     */
    public function getTaxInfo()
    {
        //prevent throwing an exception for backward compatibility
        Logger::err('getTaxInfo not implemented for ' . get_class($this));
    }
```

blankse: Yes is also possible. I just thought it would be good for developers to know what the function should return. Should then be commented about `@return void`

For v7 it would be good to convert it in a abstract class with abstract methods 😉

blankse: @brusch I changed it

bundles/EcommerceFrameworkBundle/PricingManager/Action/CartDiscount.php
```diff
@@ -93,7 +93,7 @@ public function toJSON()
     /**
      * @param string $string
      *
-     * @return ConditionInterface
```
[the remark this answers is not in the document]
blankse: I removed the use

bundles/EcommerceFrameworkBundle/PricingManager/Action/FreeShipping.php
```diff
@@ -69,7 +69,7 @@ public function toJSON()
     /**
      * @param string $string
      *
-     * @return ConditionInterface
```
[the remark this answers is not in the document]
blankse: I removed the use

brusch:
👌 excellent, thanks!

bundles/EcommerceFrameworkBundle/Model/AbstractOrderItem.php
blankse: @brusch Now we have new errors like
`Result of method Pimcore\Bundle\EcommerceFrameworkBundle\Model\AbstractOrder::getTaxInfo() (void) is used.` 😄

brusch: Oh dear 😁 Maybe we should just exclude them as they get changed anyway with v7.

blankse: Yes we should fix so much as possible and the rest ignore via the baseline feature (https://github.com/pimcore/pimcore/issues/5725)

PR #5723, opened 2020-02-05
Add client_max_body_size directive to nginx config

PimLab:
The default nginx client_max_body_size is set to 1m (http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size). On default nginx config the directive is not set in the highest level in "http". Based on the Pimcore Documentation client_max_body_size is set to 100m to prevent the "413 Request Entity Too Large" error.

  
  

Resolves #

"413 Request Entity Too Large" error using nginx while upload in frontend and backend files larger then 1m.


brusch:
Thanks a lot!

PR #5727, opened 2020-02-05
Use PDO MySQL Driver Option PDO::ATTR_EMULATE_PREPARES - Doctrine sho…

brusch:


weisswurstkanone:
also related to #5724

dpfaffenbauer:
this breaks Doctrine ORM. Not sure why but it doesn't work for me anymore with this enabled...

dpfaffenbauer:
https://github.com/doctrine/dbal/issues/1270

dpfaffenbauer:
there is a dirty fix available, but the question rather is: how compatible does Pimcore wanna be with orm? so we have these possible ways:

 - Pimcore reverts this change and casts integers in code. which it used to do
 - Pimcore reverts this change and sets the param for queries where it needs it
 - Pimcore implements the dirrty fix
 - Pimcore just don't cares about ORM anymore and people have to fix stuff themself

Here is the fix: https://stackoverflow.com/questions/31089893/doctrine-doesnt-persist-entity-with-boolean-values-and-pdoattr-emulate-prepar/35800979#35800979

weisswurstkanone:
would say that 1) is not an option

dpfaffenbauer:
whats the reason why? What is even the reason for this change?

brusch:
There's another option:
Someone fixes the issue in Doctrine 🤓 
They say they will include a fix, so best would be definitely to fix it there. 

In the meantime we can remove the config again from Pimcore, but that won't be permanently. We'll add this again soon (at the latest with v7), no matter if this was fixed in Doctrine or not. 

The reason for this change was originally discovered while implementing #5724. We have seen that this change significantly reduces the memory usage (`int` vs `string`).

dpfaffenbauer:
That being said, if doctrine doesn't fix it, Pimcore doesn't Support Doctrine ORM anymore with v7?

brusch:
No reason to panic now...
I've just said that this is an issue of Doctrine and therefore it should be fixed there and that we're going to re-add this config option in the future. If no one is willing to create a patch + tests for Doctrine this would be sad, but we still have some alternative options then.

dpfaffenbauer:
good.

PR #5729, opened 2020-02-05
Error when sorting product list with sub tenants

daves-weblab:
When joining the mysql productindex with a sub tenant an error is thrown on ordering since the o_id column is ambiguous. The default order key should be a.o_id instead of just o_id.


brusch:
Thanks a lot! 🙏

PR #5730, opened 2020-02-05
Use constant instead of its value

jdreesen:
The values of those PDO constants tend to change between PHP versions (see: https://stackoverflow.com/a/20990359) so it's better to not rely on them and use the real constant instead of its "magic number" value.

Follow-up to #5727


brusch:
Very nice, thanks a lot! 👍

PR #5733, opened 2020-02-06
[Document|Data Object] consider latest version for copying elements

### LINKED ISSUE #5721: [Document|Data Object] Content is not copied unless the content is saved & published.
Version: v3.6.3

Description:
Content is not copied unless the content is saved and published.

Steps to reproduce:

1. Create a new content page or modify one,
2. Only save, DON't Publish,
3. Copy and Paste with inheritance as a new language variant,
4. Notice that the modified content is not visible on the copied page.

Video: https://we.tl/t-tJzR0VFnEO

Current behavior:
Content is not copied when the user whats to paste it on a new page.

Expected behavior:

When the user wants to copy and paste a Page he should do it without publishing it.

dvesh3:
Resolves #5721

## Additional info  
Pasted element published status set to false, in case of "There is a newer not published version"
image

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -1974,6 +1974,11 @@ class DataObjectController extends ElementControllerBase implements EventedContr
         if ($target->isAllowed('create')) {
             $source = DataObject::getById($sourceId);
             if ($source != null) {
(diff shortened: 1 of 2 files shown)


brusch:
Just for the sake of completeness: I've pushed a small fix in 788b6bf8367025a8ae04f8820c63009e49b13314 because this PR didn't check for non-versionable types such as folders, links ...

PR #5736, opened 2020-02-06
Translate perspectives and docTypes in user/role settings

blankse:
The settings window should display the translated name of perspective and doc type


weisswurstkanone:
thx!

PR #5740, opened 2020-02-06
Translations for "Uncategorized"

blankse:
I want add a german translation for Uncategorized via POEditor


brusch:
Just for reference, it's used here: https://github.com/pimcore/pimcore/blob/ca75246d82fe1190ce2268767ace84f157b168d9/models/Document/Tag/Areablock.php#L446

brusch:
Thanks a lot! 👍

PR #5742, opened 2020-02-06
Update Important Notice Text of XLIFF Export

eliasfellner:
The current translation was confusing and did not indicate, whether the copied tree, or the original tree should be used for the export.
I believe the new translation would give clearer instructions to the user.


brusch:
Makes totally sense, thanks a lot for updating!

PR #5744, opened 2020-02-07
Ecommerce | Elastic Search |  Native Reindexing and Synonym Updates

andreas-gruenwald:
This PR contains some very handy Elastic Search (ES) utilities for reindexing and synonym updates. 

## Elastic Worker Extensions

First, it contains two extensions for the ``AbstractElasticSearch`` worker, which is the boilerplate for ES updates:

1)  Method ``performNativeReindexing`` for reindexing that utilizes "native" ES reindexing API, which is particulary useful during development, or if index mappings require a complete rebuild of an index.

2) Method ``refreshSynonymsInIndex`` for activating updates in the search index when working with synonym files.

## Elastic Search Sync Command

Second, a new command ``ecommerce:indexservice:elasticsearch-sync`` is introduced.
1) Mode "reindex": performs a native reindexing based on the ES Reindexing API. This is much faster than updates via the ecommerce ES update queue, and the action is processed atomically.

2) Mode ``refresh-synonyms``: copy the synonym files from a specified Pimcore asset directory to the ES configuration directory. Close and reopen the index to activate the changes.


brusch:
@andreas-gruenwald Thanks for your PR. Unfortunately there's an issue here: https://travis-ci.com/pimcore/pimcore/jobs/284765095#L1228 can you please fix that? 😊

brusch:
@andreas-gruenwald thanks a lot for fixing the issue. 
Just one more thing:
We should add all this changes to the docs: 
https://github.com/pimcore/pimcore/blob/master/doc/Development_Documentation/10_E-Commerce_Framework/05_Index_Service/01_Product_Index_Configuration/05_Data_Architecture_and_Indexing_Process.md

Again, thanks a lot!

andreas-gruenwald:
@brusch  Done. Feel free to change, modify, relocate. I think in general there is potential to improve the Ecommerce Elastic Search documentation (description of index mapping features, examples, etc.).

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -836,4 +840,126 @@ protected function createEsIndex(string $indexName)
         $this->settingsHash = md5(json_encode($this->tenantConfig->getIndexSettings()));
         $this->updateVersionFile();
     }
+
+    /**
+     * Get the next index version, e.g. if currently 13, then 14 will be returned.
+     * @return int
+     */
+    private function getNextIndexVersion() : int {
```
fashxp, line 848: should we make all private functions protected in order to give possibility to overwrite if necessary?

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -836,4 +840,126 @@ protected function createEsIndex(string $indexName)
...
+     * Create an index with the given name (delete existing ones), and force a rebuild in ES.
+     * Does not perform an index switch.
+     * @param string $indexName
+     * @throws BadRequest400Exception
+     * @throws \Elasticsearch\Common\Exceptions\NoNodesAvailableException
+     */
+    private function buildIndexAndPerformEsRebuild(string $indexName) {
+        $indexName = strtolower($indexName);
+        $esClient = $this->getElasticSearchClient();
+        Logger::info('Index '.$indexName.' already existed. Delete and rebuild it...');
```
fashxp, line 875:
```suggestion
        Logger::info('Index '.$indexName.' already existing. Delete and rebuild it...');
```

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -836,4 +840,126 @@ protected function createEsIndex(string $indexName)
...
+    private function getNextIndexVersion() : int {
+        return $this->getIndexVersion()+1;
+    }
+
+    /**
+     * Build the name of the index for a specific index number.
+     * @param int $id e.g. 13
+     * @return string the name of the index, such as at_de_elastic_13
+     */
+    private function buildIndexName(int $id) : string {
```
fashxp, line 857: why not provide an optional parameter to `getIndexNameVersion` and use this method instead? 
```
    public function getIndexNameVersion($indexVersionOverride = null)
    {
        $indexVersion = $indexVersionOverride ?? $this->getIndexVersion()
        return $this->indexName . '-' . $indexVersion;
    }
```
snippet not tested though.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
fashxp: and should we use this method in line [538](https://github.com/pimcore/pimcore/pull/5744/files#diff-d02a59391d96d7d11a3a41cfdf9270deL536) too?

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -836,4 +840,126 @@ protected function createEsIndex(string $indexName)
...
+            $esClient->indices()->delete(['index' => $indexName]);
+        }
+
+        $this->createEsIndex($indexName);
+
+        $body =
+            [
+                'source' => [
+                    'index' => $this->getIndexNameVersion(), //currently active ES index with data.
+                    /*
```
fashxp, line 888: what's that comment for?

fashxp: please remove unnecessary comments

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -836,4 +840,126 @@ protected function createEsIndex(string $indexName)
...
+     * @throws BadRequest400Exception
+     * @throws \Elasticsearch\Common\Exceptions\NoNodesAvailableException
+     */
+    private function buildIndexAndPerformEsRebuild(string $indexName) {
+        $indexName = strtolower($indexName);
+        $esClient = $this->getElasticSearchClient();
+        Logger::info('Index '.$indexName.' already existed. Delete and rebuild it...');
+        $result = $esClient->indices()->exists(['index' => $indexName]);
+        if ($result) {
+            Logger::info('Deleted '.$indexName.'.');
```
fashxp, line 878: technically at this point is should state 'deleting' ... it is not deleted yet, might be a misleading log message.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -836,4 +840,126 @@ protected function createEsIndex(string $indexName)
...
+        $this->settingsHash  = md5(json_encode($this->tenantConfig->getIndexSettings()));
+        $result = $this->updateVersionFile();
+        if (!$result) {
+            throw new \Exception("Can't write version file: " . $this->getVersionFile());
+        }
+
+        Lock::release(self::REINDEXING_LOCK_KEY);
+
+        if ($hasReindexIndexingModeCompleted) {
+            //continue with rebuild, or reset queue...
```
fashxp, line 950: missing code?

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -836,4 +840,126 @@ protected function createEsIndex(string $indexName)
...
+    /**
+     * Perform a native reindexing via the ES reindexing API.
+     * This is typically much faster than reindexing single records.
+     * @throws \Exception
+     */
+    public function performNativeReindexing() {
+        $hasReindexIndexingModeCompleted = false;
+        Lock::lock(self::REINDEXING_LOCK_KEY);
+        if ($this->isInReindexMode()) {
+            //stop reindexing process. If not yet finished, no alias switch will take place
```
fashxp, line 928: this does not stop the reindex-mode .. it just checks, if it is finished, and if so, it switches alias.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -836,4 +840,126 @@ protected function createEsIndex(string $indexName)
...
+        $indexName = strtolower($indexName);
+        $esClient = $this->getElasticSearchClient();
+        Logger::info('Index '.$indexName.' already existed. Delete and rebuild it...');
+        $result = $esClient->indices()->exists(['index' => $indexName]);
+        if ($result) {
+            Logger::info('Deleted '.$indexName.'.');
+            $esClient->indices()->delete(['index' => $indexName]);
+        }
+
+        $this->createEsIndex($indexName);
```
fashxp, line 882: this will result in inconsistent version file!! 
new index is created, but index number in version file is updated (since `$this->indexVersion` is not changed), but settings has is updated in version file.

fashxp: in `performNativeReindexing` the version file is then updated correctly, but `buildIndexAndPerformEsRebuild` must not be called without that context ... so maybe we should move that code to `performNativeReindexing`?

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
fashxp: so whats the point of these lines?

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -836,4 +840,126 @@ protected function createEsIndex(string $indexName)
...
+        Lock::lock(self::REINDEXING_LOCK_KEY);
+        if ($this->isInReindexMode()) {
+            //stop reindexing process. If not yet finished, no alias switch will take place
+            $this->completeReindexMode();
+            $hasReindexIndexingModeCompleted = true;
+        }
+
+        $nextIndex = $this->getNextIndexVersion();
+        $nextIndexName = $this->buildIndexName($nextIndex);
+        $this->buildIndexAndPerformEsRebuild($nextIndexName);
```
fashxp, line 935: what happens, if updates take place during ES reindex process? are they lost? 
or what happens if mapping updates take place?

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
andreas-gruenwald: To stay consistent I declared all private methods as protected.
I also replaced ``$this->indexVersion++`` with ``getNextIndexVersion()``.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
andreas-gruenwald: Logging message improved.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
andreas-gruenwald: God idea! So we got rid of the ``buildIndexName()`` method...

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
andreas-gruenwald: It was useful for testing, as smaller parts of the index are copied only.
However, unnecessary.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
andreas-gruenwald: Agree. Replaced.

andreas-gruenwald:
**Summary:**
 - we need to add lock in ``doUpdateIndex()``.
 - we need to add lock in ``doCreateOrUpdateIndexStructures()``.
- ``buildIndexAndPerformEsRebuild()`` should handle index version handling (``$this->switchIndexAlias();``).
 - Requires detailed testing.

fashxp:
after extensive testing I thought about an alternative approach for the reindex thing.
see https://github.com/pimcore/pimcore/pull/6437 ... what do you think. Would be a bit less complex.

andreas-gruenwald:
> after extensive testing I thought about an alternative approach for the reindex thing.
> see #6437 ... what do you think. Would be a bit less complex.

#6437 is the same approach as this one, but it goes further, by also removing the reindexing mode mechanism. I am going to vote for the more radical approach. Let's continue the discussion in #6437 then.

fashxp:
closed in #6437 ... for synonym update please provide a separate PR

Issue #4518, opened 2019-06-12
[Core] Remove usage of legacy system configuration


brusch:
- do not use `Pimcore\Config::getSystemConfig()`
- no usage of the object based config `$config->general->....`
- use DI (preferred) or array config

dvesh3:
TODO: add upgrade notes

brusch:
added to upgrade notes for 6.5.0 as it was deprecated with this release

PR #5745, opened 2020-02-07
[Core] Refactor System Configuration - Admin Bundle & CoreBundle

dvesh3:
Resolves partially #4518


lib/Config.php
```diff
@@ -39,6 +39,64 @@ class Config
...
+     * @see    ArrayAccess::offsetSet()
+     *
+     * @param  mixed $offset
+     * @param  mixed $value
+     *
+     * @return void
+     */
+    public function offsetSet($offset, $value)
+    {
+        if (is_null($offset)) {
```
brusch: I'd throw an Exception here, as modifying the config isn't allowed.

lib/Config.php
```diff
@@ -39,6 +39,64 @@ class Config
...
+    /**
+     * @see    ArrayAccess::offsetUnset()
+     *
+     * @param  mixed $offset
+     *
+     * @return void
+     */
+    public function offsetUnset($offset)
+    {
+        unset(static::$systemConfig[$offset]);
```
brusch: Here as well an Exception.

brusch:
Looks good to me, I'd go ahead with this approach 👍

PR #5747, opened 2020-02-07
[Document] Navigation: Possibility to specify depth + introduced new param config based build() method

### LINKED ISSUE #5662: [Document] Navigation Builder: Possibility to specify depth
Currently the entire tree is resolved and cached, especially for huge site with thousands of child pages under the root node, this could have a huge performance impact.

brusch:
Resolves #5662

### DIFF
--- a/lib/Templating/Helper/Navigation.php
+++ b/lib/Templating/Helper/Navigation.php
@@ -98,14 +98,13 @@ class Navigation extends Helper
 
     /**
      * Builds a navigation container by passing params
-     * Possible config params are: 'root', 'htmlMenuPrefix', 'pageCallback', 'cache', 'maxDepth'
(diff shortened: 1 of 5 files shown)


lib/Templating/Helper/Navigation.php
```diff
@@ -64,29 +65,66 @@ public function getName()
...
+    /**
+     * Builds a navigation container by passing params
+     * Possible config params are: 'root', 'htmlMenuPrefix', 'pageCallback', 'cache', 'maxDepth'
+     *
+     * @param Document $activeDocument
+     * @param array $params
+     * @return Container
+     * @throws \Exception
+     */
+    public function build(Document $activeDocument, array $params): Container
```
dpfaffenbauer: 👍 Like that a lot, that allows BC break safe changes ;)

PR #5750, opened 2020-02-07
[Ecommerce][Indexservice] Fix doPreIndexDataModification call

jremmurd:
# Bugfix
If the `$data = $this->doPreIndexDataModification($data);` gets called after the categorization of 

`$indexSystemData = [];
            $indexAttributeData = [];
            $indexRelationData = [];` 

the data in the mockup cache and the data which is written to the index may get inconsistent!

- Data `$data` which considers the adjustments done in `doPreIndexDataModification` is written to the cache in line https://github.com/pimcore/pimcore/blob/master/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L436
- Data `$this->bulkIndexData` which does not consider the adjustments done in `doPreIndexDataModification` is collected in https://github.com/pimcore/pimcore/blob/master/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L425


brusch:
Thanks a lot! 🙏

PR #5754, opened 2020-02-07
Fixed undefined index

weisswurstkanone:


blankse:
@weisswurstkanone The same line is 3 times in `\Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectHelperController`

I think it should be changed there too, right?

weisswurstkanone:
thx! see https://github.com/pimcore/pimcore/pull/5756

PR #5757, opened 2020-02-10
[Data Object] make relations caching layer the standard, remove option to disable it

weisswurstkanone:
also see #5483


models/DataObject/ClassDefinition.php
```diff
@@ -1329,20 +1314,7 @@ public function getLinkGenerator()
...
-     * @param bool $cacheRawRelationData
-     *
-     * @return $this
-     */
-    public function setCacheRawRelationData($cacheRawRelationData)
-    {
-        $this->cacheRawRelationData = (bool) $cacheRawRelationData;
-
-        return $this;
+    public function getCacheRawRelationData() {
```
brusch: You want to keep that for compatibility reasons? 
As this was just used internally I'd just kick it.

bundles/CoreBundle/Resources/migrations/Version20200210101048.php
```diff
@@ -0,0 +1,42 @@
...
+     * @param Schema $schema
+     * @throws \Exception
+     */
+    public function up(Schema $schema)
+    {
+        $list = new ClassDefinition\Listing();
+        $list = $list->load();
+
+        foreach ($list as $class) {
+            if ($class->getCacheRawRelationData()) {
```
brusch: This check is obsolete, the method `getCacheRawRelationData()` should go anyway, even if we keep it it always returns `true`

models/DataObject/Concrete.php
```diff
@@ -950,40 +953,52 @@ public function retrieveSlugData($descriptor)
     public function retrieveRelationData($descriptor)
```
brusch: Let's mark `retrieveRelationData()` `@internal`

models/DataObject/Concrete.php
```diff
@@ -950,40 +953,52 @@ public function retrieveSlugData($descriptor)
...
 
-            return $filteredData;
-        } else {
-            return $this->doRetrieveData($descriptor, 'object_relations_' . $this->getClassId());
+        return $filteredData;
+
+    }
+
+    /**
+     * @return array
```
brusch: Also `@internal` plz.

PR #5758, opened 2020-02-10
[Data Objects] Possibility to configure Interfaces on Class Definitio…

### LINKED ISSUE #5715: [Data Objects] Possibility to configure Interfaces on Class Definitions
# Feature Request

Create a possibility to define interfaces in Pimcore class definitions.

## Feature description

We already can define parent classes and traits:

class-definition

https://pimcore.com/docs/6.x/Development_Documentation/Objects/Object_Classes/Class_Settings/Inheritance.html

So why not interfaces? It would be useful for certain scenarios. Recently I was writing a form builder built on Data objects. It contains four class definitions:

FormInputText
FormInputOptions
FormInputSelect
FormInputFile

They all have some common methods, but a trait would be unnecessary as there is no code reuse. I just want the class definition to use the interface FormInput such that I can check with instanceof in the service.

Or did I miss something?

weisswurstkanone:
resolves #5715


bundles/CoreBundle/Resources/migrations/Version20200210101048.php
```diff
@@ -0,0 +1,42 @@
...
+     * @throws \Exception
+     */
+    public function up(Schema $schema)
+    {
+        $list = new ClassDefinition\Listing();
+        $list = $list->load();
+
+        foreach ($list as $class) {
+            if ($class->getCacheRawRelationData()) {
+                // get rid of the CacheRawRelationDataInterface
```
jdreesen: How is this related?

models/DataObject/AbstractObject.php
```diff
@@ -300,7 +300,7 @@ public static function getById($id, $force = false)
 
                     $object->__setDataVersionTimestamp($object->getModificationDate());
 
-                    if ($object instanceof CacheRawRelationDataInterface) {
+                    if ($object instanceof Concrete) {
```
jdreesen: How is this related?

models/DataObject/ClassDefinition.php
```diff
@@ -397,12 +405,30 @@ public function save($saveDefinitionFile = true)
...
+            foreach ($customParts as $interface) {
+                $interface = trim($interface);
+                if (Tool::interfaceExists($interface)) {
+                    $customParts[]= $interface;
+                } else {
+                    throw new \Exception("interface '" . $interface . "' does not exist");
+                }
+            }
+
+            $implementsParts[] = $this->getImplementsInterfaces();
```
jdreesen: This is somehow strange.

Is the `foreach` only for checking the existence? In that case line 420 `$customParts[]= $interface;` is not needed.

Otherwise, the line 426 `$implementsParts[] = $this->getImplementsInterfaces();` should be removed and line 420 should be `$implementsParts[]= $interface;` instead.

models/DataObject/Objectbrick/Definition.php
```diff
@@ -232,7 +232,26 @@ public function save($saveDefinitionFile = true)
...
+            foreach ($customParts as $interface) {
+                $interface = trim($interface);
+                if (Tool::interfaceExists($interface)) {
+                    $customParts[]= $interface;
+                } else {
+                    throw new \Exception("interface '" . $interface . "' does not exist");
+                }
+            }
+
+            $implementsParts[] = $this->getImplementsInterfaces();
```
jdreesen: Same as above.

models/Webservice/Data/ClassDefinition.php
```diff
@@ -103,9 +103,9 @@ class ClassDefinition extends Model\Webservice\Data
     public $showVariants = false;
 
     /**
-     * @var bool
+     * @var string
      */
-    public $cacheRawRelationData = false;
+    public $implementsInterfaces = false;
```
jdreesen: Is this intended?

bundles/CoreBundle/Resources/migrations/Version20200210101048.php
weisswurstkanone: It isn't :-)

models/DataObject/Fieldcollection/Definition.php
```diff
@@ -340,7 +347,27 @@ public function save($saveDefinitionFile = true)
...
+            foreach ($customParts as $interface) {
+                $interface = trim($interface);
+                if (Tool::interfaceExists($interface)) {
+                    $customParts[]= $interface;
+                } else {
+                    throw new \Exception("interface '" . $interface . "' does not exist");
+                }
+            }
+
+            $implementsParts[] = $this->getImplementsInterfaces();
```
jdreesen: Same as above.

models/DataObject/ClassDefinition/Service.php
```diff
@@ -358,4 +360,63 @@ public static function skipColumn($tableDefinitions, $table, $colName, $type, $d
...
         return false;
     }
+
+    /**
+     * @param array $implementsParts
+     * @param string $newInterfaces
+     * @return string
+     * @throws \Exception
+     */
+    public static function buildImplementsInterfaces($implementsParts = [], string $newInterfaces) {
```
jdreesen: As it is not possible to just skip this parameter because the next one is required, the signature should be as follows:
```suggestion
    public static function buildImplementsInterfaces(array $implementsParts, string $newInterfaces) {
```

models/DataObject/ClassDefinition/Service.php
```diff
@@ -358,4 +360,63 @@ public static function skipColumn($tableDefinitions, $table, $colName, $type, $d
 
         return false;
     }
+
+    /**
+     * @param array $implementsParts
+     * @param string $newInterfaces
```
jdreesen:
```suggestion
     * @param string $newInterfaces A comma separated list of interfaces
```

models/DataObject/ClassDefinition/Service.php
```diff
@@ -358,4 +360,63 @@ public static function skipColumn($tableDefinitions, $table, $colName, $type, $d
...
+        if ($implementsParts) {
+            return ' implements ' . implode(', ', $implementsParts);
+        }
+        return '';
+    }
+
+
+    /**
+     * @param array $useParts
+     * @param string $newTraits
```
jdreesen:
```suggestion
     * @param string $newTraits A comma separated list of traits
```

models/DataObject/ClassDefinition/Service.php
```diff
@@ -358,4 +360,63 @@ public static function skipColumn($tableDefinitions, $table, $colName, $type, $d
...
+     * @return string
+     * @throws \Exception
+     */
+    public static function buildUseTraits($useParts = [], string $newTraits) {
```
jdreesen: As it is not possible to just skip this parameter because the next one is required, the signature should be as follows:
```suggestion
    public static function buildUseTraits(array $useParts, string $newTraits) {
```

lib/Tool.php
```diff
@@ -659,11 +659,11 @@ public static function interfaceExists($class)
 
     /**
      * @param string $class
-     * @param string $type
+     * @param string $type (e.g. 'class', 'interface', 'trait')
      *
      * @return bool
      */
-    protected static function classInterfaceExists($class, $type)
+    public static function classInterfaceExists($class, $type)
```
jdreesen: Wouldn't it be better to leave this protected and add a facade method like for the other two options?

```php
    /**
     * @static
     *
     * @param string $trait
     *
     * @return bool
     */
    public static function traitExists($trait)
    {
        return self::classInterfaceExists($trait, 'trait');
    }
```

brusch:
@weisswurstkanone can you please resolve the conflicts? Thanks a lot ;)

models/DataObject/ClassDefinition/Service.php
```diff
@@ -358,4 +360,63 @@ public static function skipColumn($tableDefinitions, $table, $colName, $type, $d
...
+                } else {
+                    throw new \Exception("trait '" . $trait . "' does not exist");
+                }
+            }
+        }
+
+        if ($useParts) {
+            $result = '';
+            foreach ($useParts as $part) {
+                $result .= 'use ' . $part . ";\r\n";
```
jdreesen: Everywhere else, only `\n` is used.
```suggestion
                $result .= 'use ' . $part . ";\n";
```

lib/Tool.php
weisswurstkanone: don't see a reason why it can't be public

blankse:
What do you think of generating the php classes with an external generator?
For example with https://github.com/nette/php-generator

The result would be cleaner and could take PSR's code style into account.

brusch:
@blankse yes, sounds good, but I'd rather see this as an extra topic 😊
Would you like to do this? 🥺

blankse:
@brusch Yes, of course, is an extra topic. Just want to know if this is welcome. If I have time, I test it and do a PR.

models/DataObject/ClassDefinition.php
```diff
@@ -1326,4 +1328,43 @@ public function getLinkGenerator()
...
+
+
+    /**
+     * @param string|null $implementsInterfaces
+     * @return $this
+     */
+    public function setImplementsInterfaces(?string $implementsInterfaces)
+    {
+        $this->implementsInterfaces = $implementsInterfaces;
+        return $this;
```
jdreesen: Code style:
```suggestion

        return $this;
```

models/DataObject/ClassDefinition.php
```diff
@@ -1326,4 +1328,43 @@ public function getLinkGenerator()
...
+     */
+    public function getImplementsInterfaces(): ?string
+    {
+        return $this->implementsInterfaces;
+    }
+
+
+    /**
+     * @param string|null $implementsInterfaces
+     * @return $this
```
jdreesen: Code style:
```suggestion
     *
     * @return $this
```

models/DataObject/ClassDefinition.php
```diff
@@ -1227,14 +1245,10 @@ public function getGroup()
...
     /**
      * @param string $group
-     *
-     * @return $this
      */
     public function setGroup($group)
     {
         $this->group = $group;
-
-        return $this;
```
jdreesen: Why are those removed again? They were just added in #5759

lib/Tool.php
jdreesen: Maybe that's right. 

It is just strange, because `$type` can only take 3 values and there are explicit public methods for the other 2.

So why not add a third one, that makes the intention clear, instead of forcing the user to call a method named `classInterfaceExists` to check if a **trait** exists (and that's the only real reason to call this method because for the other cases there are simpler, better named methods).

models/DataObject/ClassDefinition/Service.php
```diff
@@ -358,4 +360,63 @@ public static function skipColumn($tableDefinitions, $table, $colName, $type, $d
...
         return false;
     }
+
+    /**
+     * @param array $implementsParts
+     * @param string $newInterfaces
+     * @return string|null
+     * @throws \Exception
+     */
+    public static function buildImplementsInterfaces($implementsParts = [], ?string $newInterfaces) {
```
jdreesen: Again: adding a default value to a parameter makes no sense when there are subsequent parameters without default values (`$newInterfaces` is just nullable, but has no default value).

Also, it does not force the parameter type to be an array, while the method will crash if it's not an array (and even if method is called with `null` for this parameter, it will be `null` and not an empty array - the default value does not help in this case).

This also counts for `buildUseTraits()`.

models/DataObject/ClassDefinition.php
weisswurstkanone: real showstopper :-)

jdreesen: Consistent code style makes code easier to read and reading it what we mostly do with code ;)

brusch:
@blankse great, yes, this is definitely welcome 😊

models/DataObject/ClassDefinition.php
```diff
@@ -1308,7 +1326,7 @@ public function getLinkGeneratorReference()
     /**
      * @param string $linkGeneratorReference
      *
-     * @return $this;
+     * @param $this
```
jdreesen:
```suggestion
     * @return $this
```

blankse:
```suggestion
     * @return $this
```

models/DataObject/ClassDefinition.php
```diff
@@ -1326,4 +1344,45 @@ public function getLinkGenerator()
 
         return $generator;
     }
+
+    /**
+     * @return bool
+     */
+    public function getCacheRawRelationData(): bool
```
brusch: #5757 ;)

models/DataObject/ClassDefinition/Service.php
```diff
@@ -105,8 +106,8 @@ public static function importClassDefinitionFromJson($class, $json, $throwExcept
         $class->setModificationDate(time());
         $class->setUserModification($userId);
 
-        foreach (['description', 'icon', 'group', 'allowInherit', 'allowVariants', 'showVariants', 'parentClass',
-                    'listingParentClass', 'useTraits', 'listingUseTraits', 'previewUrl', 'propertyVisibility',
+        foreach (['description', 'icon', 'group', 'allowInherit', 'allowVariants', 'cacheRawRelationData', 'showVariants', 'parentClass',
```
brusch: #5757

models/DataObject/Objectbrick/Definition.php
```diff
@@ -232,7 +232,26 @@ public function save($saveDefinitionFile = true)
         $cd .= 'use Pimcore\Model\DataObject\PreGetValueHookInterface;';
         $cd .= "\n\n";
 
-        $cd .= 'class ' . ucfirst($this->getKey()) . ' extends ' . $extendClass . ' implements \\Pimcore\\Model\\DataObject\\DirtyIndicatorInterface {';
+        $implementsParts = ['\\Pimcore\\Model\\DataObject\\DirtyIndicatorInterface'];
+
+        if ($this->getImplementsInterfaces()) {
+            $customParts = $this->getImplementsInterfaces();
```
brusch: We should use `DataObject\ClassDefinition\Service::buildImplementsInterfaces` here as well, right?

dpfaffenbauer:
> What do you think of generating the php classes with an external generator?
> For example with https://github.com/nette/php-generator
> 
> The result would be cleaner and could take PSR's code style into account.
@blankse 

I already once made a POC for that:

https://github.com/dpfaffenbauer/pimcore/commit/16f31ab79866360578a57326872dae59ada80353

It doesn't use nette, but it refactors the class-building into services.

That branch also does way more than just only separate the class-building, it also uses Doctrine DBAL Schema Manager to create DataObject Schemas for better migration between data-object changes. But I never finished it... (would be also a big BC as well)

PR #5759, opened 2020-02-10
class/fc/brick definition: chaining

weisswurstkanone:


models/DataObject/ClassDefinition.php
```diff
@@ -1242,10 +1242,13 @@ public function getGroup()
 
     /**
      * @param string $group
+     *
+     * @return $this
      */
     public function setGroup($group)
     {
         $this->group = $group;
+        return $this;
```
jdreesen: Missing blank line before `return`:

```suggestion

        return $this;
```

PR #5768, opened 2020-02-11
[XLIFF] Import - bypass element not found exceptions

### LINKED ISSUE #5764: XLIFF Import throws Error (objects)
### Expected behavior
XLIFF Import (for objects) works properly, although one or more objects have been deleted in the meantime after the export.

### Actual behavior
XLIFF Import (for objects) crashes at the object, which doesn't exist anymore and doesn't go on with the following objects

### Steps to reproduce
- Make a XLIFF Export of a folder with objects in it.
- Delete an object in this folder.
- Try to import translated XLIFF file --> throws Error

Return Line:
https://github.com/pimcore/pimcore/blob/v6.4.2/bundles/AdminBundle/Controller/Admin/TranslationController.php#L901

dvesh3:
Resolves #5764


bundles/AdminBundle/Controller/Admin/TranslationController.php
```diff
@@ -899,9 +899,15 @@ public function xliffImportElementAction(Request $request, ImportDataExtractorIn
             $importerService->import($attributeSet);
         } catch (\Exception $e) {
             Logger::err($e->getMessage());
+            $success = false;
+
```
weisswurstkanone: I guess this can be simplified

dvesh3: @weisswurstkanone how?

bundles/AdminBundle/Controller/Admin/TranslationController.php
```diff
@@ -899,9 +899,11 @@ public function xliffImportElementAction(Request $request, ImportDataExtractorIn
             $importerService->import($attributeSet);
         } catch (\Exception $e) {
             Logger::err($e->getMessage());
+            $success = $e instanceof Element\Exception\ElementNotFoundException ?? false;
```
jdreesen: I doubt that the result of `instanceof` can ever be `null`.

dvesh3: true!!....changed it 😉 thanks!

brusch:
@dvesh3 just pushed an approach that required less code, what do you think?

dvesh3:
@brusch I thought about the same approach initially but was just concerned that it will silently ignore the elements not found during the process. I believe user should be notified in some way(logs) that some jobs failed due to an exception.

brusch:
Just added logging, but with a lower severity (`warn`), because actually it doesn't break the import process.

PR #5769, opened 2020-02-11
[DataObjects] Recheck indices on relations table

### LINKED ISSUE #5687: [DataObjects] Recheck indices on relations table
- check if all index are necessary
- eventually use compound index 

--> reduce update/insert costs, reduce memory consumption for index

weisswurstkanone:
Resolves #5687


brusch:
@weisswurstkanone what about the primary key? I guess that can go as well, right?

PR #5773, opened 2020-02-11
[PHPStan] level 2 fixes

weisswurstkanone:


models/Document/Tag/Pdf.php
```diff
@@ -70,8 +70,8 @@ public function getDataForResource()
     public function getDataEditmode()
     {
         $pages = 0;
-        if ($asset = Asset::getById($this->id)) {
-            /** @var Asset\Document $asset */
+
+        if (/** @var Asset\Document $asset */ $asset = Asset::getById($this->id)) {
```
aarongerig: Why not put that comment above the conditional statement and not inside it? It looks very weird to have a comment in between code ...

models/Document/Tag/Video.php
```diff
@@ -382,11 +382,11 @@ public function getAssetCode($inAdmin = false)
                     $imageThumbnailConf['format'] = 'JPEG';
                 }
 
-                if ($this->poster && ($poster = Asset::getById($this->poster))) {
-                    /** @var Asset\Image $image */
+                if ($this->poster && (/** @var Asset\Image $poster */ $poster = Asset::getById($this->poster))) {
```
aarongerig: same here ...

models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
```diff
@@ -314,7 +314,7 @@ public function getDataFromGridEditor($data, $object = null, $params = [])
     }
 
     /**
-     * @param $data
+     * @param array|null$data
```
aarongerig: missing space after `array|null`

models/Document/Tag/Video.php
```diff
@@ -382,11 +382,11 @@ public function getAssetCode($inAdmin = false)
...
                     $image = $poster->getThumbnail($imageThumbnailConf);
                 } else {
-                    if ($asset->getCustomSetting('image_thumbnail_asset') && ($customPreviewAsset = Asset::getById($asset->getCustomSetting('image_thumbnail_asset')))) {
+                    if ($asset->getCustomSetting('image_thumbnail_asset')
+                        && (/** @var Asset\Image $customPreviewAsset */ $customPreviewAsset = Asset::getById($asset->getCustomSetting('image_thumbnail_asset')))) {
```
aarongerig: here ...

models/Document/Tag/Video.php
```diff
@@ -980,7 +980,7 @@ public function getPosterAsset()
      */
     public function getImageThumbnail($config)
     {
-        if ($this->poster && ($poster = Asset::getById($this->poster))) {
+        if ($this->poster && (/** @var Asset\Image $poster */ $poster = Asset::getById($this->poster))) {
```
aarongerig: and here.

models/Document/Tag/Pdf.php
weisswurstkanone: because Mr. Stan is not happy about that. maybe there is a better solution?

models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
weisswurstkanone: thx

models/Document/Tag/Pdf.php
aarongerig: So something like
```php
/** @var Asset\Document $asset */
if ($asset = Asset::getById($this->id)) {
    // ...
}
```
or
```php
$asset = Asset::getById($this->id);

if ($asset instanceof Asset) {
    // ...
}
```
doesn't work?

weisswurstkanone: Negative :-)

blankse: @aarongerig The second approach should work
@weisswurstkanone We can use Asset\Document::getById() and change the return to static|null. Like in AbstractObject:getById(). So we don't need the comments, right?

weisswurstkanone: @aarongerig yes, second one works but this is something you sometimes don't want

aarongerig: I'm sorry, of course I mean to write:

```php
$asset = Asset::getById($this->id);

if ($asset instanceof Asset\Document) {
    $pages = $asset->getPageCount();
}
```

Just tested that on my machine and it works. :)

weisswurstkanone: that's what I said. but imagine a if/else/... scenario which is a bit more complicated

weisswurstkanone: @blankse yep, this static stuff works. I would go for it.

models/Element/AbstractElement.php
```diff
@@ -172,7 +172,9 @@ public function isLocked()
         }
 
         // check for inherited
-        return $this->getDao()->isLocked();
+        /** @var Model\Document\Dao|Model\Asset|Dao|Model\DataObject\AbstractObject\Dao $dao */
+        $dao = $this->getDao();
```
blankse: I think we should use here a method comment on class. Like https://github.com/pimcore/pimcore/pull/5765/commits/dd826788a0bcfbdeab6c759ae0f0b67138ea2970

models/Document/Tag/Video.php
blankse:
```suggestion
        if ($this->poster && ($poster = Asset\Image::getById($this->poster))) {
```

models/Element/AbstractElement.php
```diff
@@ -222,20 +224,27 @@ public function isAllowed($type, ?Model\User\AbstractUser $user = null)
...
         $type = Service::getType($this);
-        $ids = $this->getDao()->unlockPropagate();
+
+        /** @var Model\Document\Dao|Model\Asset|Dao|Model\DataObject\AbstractObject\Dao $dao */
+        $dao = $this->getDao();
+        $ids = $dao->unlockPropagate();
 
         // invalidate cache items
         foreach ($ids as $id) {
+            /** @var Model\Document\Dao|Model\Asset|Dao|Model\DataObject\AbstractObject\Dao $element */
```
blankse: The Dao classes haven't the clearDependentCache method, right? We should add the clearDependentCache method to the ElementInterface.

weisswurstkanone: Mhm, ElementInterface or AbstractElement ?
Then I would say we should add things like setId() as well

models/Element/AbstractElement.php
```diff
@@ -222,20 +224,27 @@ public function isAllowed($type, ?Model\User\AbstractUser $user = null)
         }
 
         //everything is allowed for admin
+        /** @var Model\User $user */
         if ($user->isAdmin()) {
```
blankse: I think we should change the method param. If someone put a role it would be occurred an error. The Dao method also accept only User objects

models/Element/AbstractElement.php
blankse: I would prefer the interface. It should have alle necessary methods. The `Service::getElementById()` function have as return also the interface.

models/Document.php
```diff
@@ -297,9 +298,12 @@ public static function getById($id, $force = false)
                     }
                 }
 
+                /** @var Document $document */
                 $document = self::getModelFactory()->build($className);
                 \Pimcore\Cache\Runtime::set($cacheKey, $document);
-                $document->getDao()->getById($id);
+                /** @var Document\Dao $dao */
+                $dao = $document->getDao();
```
blankse: Here we can also use the method comment on the class

PR #5777, opened 2020-02-12
[PHPStan] Fix level 2 errors

blankse:
Before
`[ERROR] Found 610 errors`

After
`[ERROR] Found 477 errors`

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/FolderController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/FolderController.php
@@ -42,6 +42,10 @@ class FolderController extends DocumentControllerBase
     {
         $folder = Document\Folder::getById($request->get('id'));
 
+        if (!$folder) {
+            throw $this->createNotFoundException();
+        }
+
         // check for lock
         if ($folder->isAllowed('save') || $folder->isAllowed('publish') || $folder->isAllowed('unpublish') || $folder->isAllowed('delete')) {
             if (Element\Editlock::isLocked($request->get('id'), 'document')) {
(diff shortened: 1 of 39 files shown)


brusch:
@blankse Ready for review or still in progress?

blankse:
@brusch I think it's enough :) You can review it.

composer.json
```diff
@@ -101,12 +101,23 @@
     "cache/integration-tests": "^0.16.0",
     "codeception/codeception": "~2.4.5",
     "phpstan/phpstan": "^0.12",
-    "phpstan/phpstan-symfony": "^0.12"
+    "phpstan/phpstan-symfony": "^0.12",
+    "heidelpay/heidelpay-php": "^1.2.5.1",
```
brusch: Imho those dependencies should stay optional.

lib/Model/ModelInterface.php
```diff
@@ -0,0 +1,60 @@
...
+     * @throws \Exception
+     */
+    public function initDao($key = null, $forceDetection = false);
+
+    /**
+     * @param string $modelClass
+     *
+     * @return string|null
+     */
+    public static function locateDaoClass($modelClass);
```
brusch: Not really needed on the interface, right?

composer.json
blankse: Its only "require-dev". Is only installed if the composer is run directly on the package (root-only). See https://getcomposer.org/doc/04-schema.md#require-dev

brusch: Oh, sorry, missed that detail in the diff view 😊 Thanks for clarifying!

blankse: I added this packages also to `suggest`. So the user see, which packages can be added to increase functionality.

lib/Model/ModelInterface.php
blankse: You are right. I removed it.

bundles/AdminBundle/Controller/Admin/Document/FolderController.php
```diff
@@ -42,6 +42,10 @@ public function getDataByIdAction(Request $request)
     {
         $folder = Document\Folder::getById($request->get('id'));
 
+        if (!$folder) {
+            throw $this->createNotFoundException();
```
weisswurstkanone: maybe add a message

blankse: Default message is "Not found" and its a 404. Not enough?

weisswurstkanone: for now I would leave it as it is

blankse: I set it, so the user know at least the type

weisswurstkanone:
thanks a lot!

PR #5779, opened 2020-02-12
XLIFF Import - DataExtractor returns null

melissakittl:
Resolves #5768

## Additional info  
The fix in the pull request above resolves the issue #5764 
The extractElement function of the DataExtractor now returns null instead of an exception, but its not defined as return value, so the import throws exception 
```Return value of Pimcore\Translation\ImportDataExtractor\Xliff12DataExtractor::extractElement() must be an instance of Pimcore\Translation\AttributeSet\AttributeSet, null returned```


weisswurstkanone:
thx!

PR #5781, opened 2020-02-12
[Data Object] Objectbricks shouldn't extend from Field Collections #5774

### LINKED ISSUE #5774: [Data Object] Objectbricks shouldn't extend from Field Collections
Better separate out shared stuff into a trait, makes it a lot easier to check for field collections in code using `instanceof`.

weisswurstkanone:
resolves #5774


brusch:
@weisswurstkanone there's a conflict that needs to be resolved 😊 Thanks!

PR #5784, opened 2020-02-12
[Redirect] Performance & Usability Optimizations

### LINKED ISSUE #5766: [Redirect] Performance Optimizations
- default should be exact matching redirects (no-regexes)
- exact matching redirect lookup can be then done directly via database
- regex redirects are done the existing way incl. caching

brusch:
resolves #5766


brusch:
@dvesh3 thanks a lot for reviewing & fixing the issues! 🙏

PR #5786, opened 2020-02-12
[DataObjects] When renaming fieldname of object brick, brick tables a…

### LINKED ISSUE #5656: [DataObjects] When renaming fieldname of object brick, brick tables are not cleaned up
rename field name of object brick 
image

old values remain in store and query table
image

new values are added in addition. 

This results in potentially unexpected filter results. 

also cause for https://github.com/pimcore/demo/issues/169

weisswurstkanone:
resolves #5656

### DIFF
--- a/lib/Maintenance/Tasks/CleanupBrickTablesTask.php
+++ b/lib/Maintenance/Tasks/CleanupBrickTablesTask.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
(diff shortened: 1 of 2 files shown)


lib/Maintenance/Tasks/CleanupBrickTablesTask.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->logger = $logger;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function execute()
+    {
+        $db = Db::get();
```
dpfaffenbauer: inject via DI

brusch: @weisswurstkanone no, please not, it will break the installer ;) Thanks!

dvesh3:
same issue with field collections :)

PR #5793, opened 2020-02-13
[Object bricks] More specific return type for object brick container getter

BlackbitNeueMedien:
Currently for object brick fields in a data object class its getter only says `@return \Pimcore\Model\DataObject\Objectbrick`. While this is correct it could be more specific by documenting the brick container class as return type, for example `@return \Pimcore\Model\DataObject\DataObjectClassName\ObjectBrickContainerFieldName`.


weisswurstkanone:
thx!

PR #5794, opened 2020-02-13
[PHPStan] level 2 fixes

weisswurstkanone:


models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -69,9 +74,20 @@ public function setClasses($classes)
      */
     public function getLazyLoading()
     {
-        return true;
+        return $this->lazyLoading;
```
dvesh3: why?

brusch: Doesn't make sense, needs to be reverted 😊

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -72,6 +77,17 @@ public function getLazyLoading()
         return true;
     }
 
+    /**
+     * @param bool $lazyLoading
+     *
+     * @return $this
+     */
+    public function setLazyLoading($lazyLoading)
```
dvesh3: what about other changes in this file?

models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php
```diff
@@ -205,6 +205,7 @@ public function getDataForQueryResource($data, $object = null, $params = [])
         $d = [];
 
         if (is_array($data) && count($data) > 0) {
+            /** @var DataObject\Data\ElementMetadata $metaObject */
```
blankse: Than we can set the $data param to `DataObject\Data\ElementMetadata[]|null`, right?

weisswurstkanone: isn't this already the case?

blankse: No it's `Element\ElementInterface[]|null`. This var comment should no longer be necessary after that.

weisswurstkanone: Ah, now I got you! thx

models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php
```diff
@@ -186,7 +186,7 @@ public function loadData($data, $object = null, $params = [])
     }
 
     /**
-     * @param Element\ElementInterface[]|null $data
+     * @param Element\ElementMetadata[]|null $data
```
blankse:
```suggestion
     * @param DataObject\Data\ElementMetadata[]|null $data
```

PR #5800, opened 2020-02-13
[Data objects] Enable inheritance for field collections

### LINKED ISSUE #1410: Inherited field collection value not shown
# Bug Report

### Expected behavior 
Inherited field collections of parent objects should be shown.
  
### Actual behavior  
Inheritance marker (grren arrow in upper left panel corner) is present but content of inherited field collections is not visible.
  
### Steps to reproduce  
1. create field collection with 1 input field
1. create class with "inheritance allowed" 
1.1. add a field collection field to this class and assign the field collection from 1. as allowed type
1. create object of this class, add the field collection and enter any value to the input field of the field collection
1. create another object of the same class as child element of the first object
1. open second object and see that inherited field collection value is not shown

BlackbitNeueMedien:
I wondered why inheritance is not supported for field collections. I am quite certain that there must be a reason but I just tried to enable it and at first glance it works quite well. Have I missed something?

Resolves #1410


fashxp:
What about query tables?

BlackbitNeueMedien:
Which table exactly do you mean? In `object_query_<class ID>` is no information about field collections.

In `object_collection_<Field collection class ID>_<class id>` there is no entry for inherited field collection items - and that is correct imho.

fashxp:
But you never can query on inherited field collection data as there are no query tables for field collection data ... and that is inconsistent to all other data types - which is one of the main reasons why there is no inheritance for field collections yet.

BlackbitNeueMedien:
Hmm ok. But what if somebody does not want to filter for field collection data but only read their values? Or in other words: Use inheritance as tool to ease data maintenance.

brusch:
@BlackbitNeueMedien My recommendation would be then to override the class and implement a custom getter that allows inheritance. We can add this approach also to the docs, that would be fine for me.

BlackbitNeueMedien:
At first I wanted to agree but now as I am updating the docs I am again wondering about the reason because the data of course is in the database in the `object_collection_*` tables - and if you want to query it you can use this table. Or do you see any disadvantages?

I also cannot understand the fact that we do not provide out-of-the-box field collection inheritance only because this information is not inside a query table. That is like not introducing a working feature because another feature is not finished.
Imho this would be as if we would not have introduced many-to-many object relation type because there is no grid filter at the moment.

dpfaffenbauer:
@BlackbitNeueMedien The query is table is used to persist inherited values into the database, whereas the store table is used to persist the actual value (null if should be inherited) into the database. That is why this feature would be inconsistent to DataObjects and Bricks. What if I want to query the real values, not inherited one? Not possible then.

BlackbitNeueMedien:
Ah, you are right, Dominik.
Ok, then I will update the docs and leave the data class as it is.

weisswurstkanone:
IMHO, inheritance only makes sense if you do it on field level which is hard to achieve because you can mix field collection item types of any flavour.
If you just want to inherit the whole container then yes, maybe ... but you have to make it clear what you can expect.

dpfaffenbauer:
@weisswurstkanone good point as well... since with field-collections, if you change the index (eg position), it might end up messing with inheritance as well....

BlackbitNeueMedien:
Ok, updated the docs regarding @weisswurstkanone's hint.

weisswurstkanone:
Please also be aware that this affects things like grid (inherited data, filtering, probably sorting, etc).
So for object bricks for example it joins the brick query (maybe even the localized one) table together.  Question is how the UX would be for the user.

brusch:
Yep, @weisswurstkanone is right, the current implementation can have weird side-effects in the admin UI. So maybe it's better to enable the inheritance only if the scope is outside the admin UI? `!\Pimcore::inAdmin()`

brusch:
I'll close this for now as this isn't a topic for the core.
However, it would be of course nice to improve the documentation in that regard and also show some possible ways (see above) there how to achieve a "kind of" inheritance with all it's possible pitfalls.

PR #5803, opened 2020-02-13
[PHPStan] Fix level 2 errors

blankse:
Before
`[ERROR] Found 397 errors`

After
`[ERROR] Found 364 errors`


models/DataObject/ClassDefinition/CustomLayout.php
```diff
@@ -522,15 +522,15 @@ public function getDescription()
     }
 
     /**
-     * @param array $layoutDefinitions
+     * @param Layout $layoutDefinitions
```
weisswurstkanone: IMHO, this can be null as well (just tried to create a new class). Same for fielddefinitions. Getters&Setters.

models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -635,26 +636,6 @@ public function getLayout()
         return $this->layout;
     }
 
-    /**
-     * @param mixed $data
-     * @param array $blockedKeys
-     *
-     * @return $this
-     */
-    public function setValues($data = [], $blockedKeys = [])
```
weisswurstkanone: $blockedKeys not necessary anymore ?

blankse: I add it direct to the parent class `\Pimcore\DataObject\ClassDefinition\Data` because of:
https://github.com/pimcore/pimcore/blob/643f7384db4a433d9f439930ecc0b6cc7f2c10a9/models/DataObject/ClassDefinition/Service.php#L266

weisswurstkanone: Saw that, but the question is if the blockedKeys parameter isn't needed. It is not supported by the parrent, right ?

blankse: The parent class has now the blockedKeys parameter. So it is no longer necessary in this child class, as it is inherited:
https://github.com/pimcore/pimcore/blob/643f7384db4a433d9f439930ecc0b6cc7f2c10a9/models/DataObject/ClassDefinition/Data.php#L332-L350

models/DataObject/ClassDefinition/CustomLayout.php
blankse: I added null to layoutDefinitions/fieldDefinitions

weisswurstkanone:
can you fix the baseline issues ?

blankse:
@weisswurstkanone PhpStan wants an `is_array` everywhere at getFieldDefinitions otherwise it ignores the type... I think we should stay at `Data[]`. The ClassDefinition class already had `public $fieldDefinitions = [];`. But it exists also a `setFieldDefinitions(null)`. This could be changed to `setFieldDefinitions([])`, right?

weisswurstkanone:
I would say yes.  Same for layoutDefinitions.

In the setter something like
$this->layoutDefinitions = is_array($layoutDefinitions) ? $layoutDefinitions : []

blankse:
@weisswurstkanone $layoutDefinitions is no array, is a Layout Object. There is null okay.

weisswurstkanone:
ok

weisswurstkanone:
setFieldDefinitions will still accept null (e.g. from definition file) and then it will stay null, right ?

blankse:
@weisswurstkanone Good catch, I changed the setters

weisswurstkanone:
thanks a lot!

PR #5804, opened 2020-02-14
DeleteUnusedLocaleDataCommand: Fix wrong SHOW TABLES

blankse:
We have classes with same beginning. For example: magazine and magazineblock

This command causes an error. Because of the missing underscore in the sql:
```
An exception occurred while executing 'DROP TABLE IF EXISTS object_localized_query_magazine_object_localized_query_magazineblock_de'
```


weisswurstkanone:
thanks!

PR #5805, opened 2020-02-14
VideoTag add more valid youtube params

putzflorian:
add "playsinline" and "hl" according to https://developers.google.com/youtube/player_parameters?hl=en


brusch:
Thanks a lot! 🙏

PR #5807, opened 2020-02-14
Javascript: Replace ts function to t in js files

vmalyk:

Replace ts  to t function for translations, because in code it`s the same:
https://github.com/pimcore/pimcore/blob/f4062f6918c53628b495827bec1d322165e69a65/bundles/AdminBundle/Resources/public/js/pimcore/functions.js#L64-L66.

## Additional info
I suggest to mark function ts as deprecated in next release and remove in 7.0.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/tags/select.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/tags/select.js
@@ -119,7 +119,7 @@ pimcore.asset.tags.select = Class.create(pimcore.asset.tags.abstract, {
             options = options.split(',');
             for (var i = 0; i < options.length; i++) {
 
-                var key = ts(options[i]);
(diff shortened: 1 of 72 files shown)


brusch:
Marked as deprecated in #5840

dvesh3:
@vmalyk thanks a lot! also, can you please change here:

https://github.com/pimcore/pimcore/blob/baf00883878433f0d56ee5c7ae3f177e5b5f496d/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/table.js#L82

https://github.com/pimcore/pimcore/blob/8de2aed3ca979d82d4f1f5a85196dbe7281c87e4/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/edit.js#L409

vmalyk:
@dvesh3 , thanks for review! I`ve added fixes in PR.

PR #5808, opened 2020-02-14
Filter stays active while changing classes #5802

### LINKED ISSUE #5802: Filter stays active while changing classes
I have records of 3 different classes in the same folder. When I click on the folder, I have a selectbox to select the class.
First I select the class, in this example Product, which has a field with the name 'sku'. Then I filter on a value in the field 'sku' so it only shows me the results and it shows the 'Filter active'  and 'Clear filter' text at the top.
When I then switch class, the 'Filter active' and 'Clear filter' dissappear and it shows me the results of the other class. Strange thing is that in case the other class has the same field name, in this case 'sku', it applies the filter unwanted to this field.
If you switch back to the original class 'Product', it doesn't show the 'Filter active' and 'Clear filter' button but as I can see the filter is still applied. 
So this seems to be a bug.
I have recorded a video. You can find it at the url https://demo.pimcore.h1.nl/videos/screencast-localhost-2020.02.13-16_43_27.webm

weisswurstkanone:
resolves #5802 

changed so that filter stays active if you open&apply grid config settings but removed if one changes the class selection.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/folder/search.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/folder/search.js
@@ -218,9 +218,6 @@ pimcore.object.search = Class.create(pimcore.object.helpers.gridTabAbstract, {
         gridHelper.enableEditor = true;
         gridHelper.limit = itemsPerPage;
 
-
-        var propertyVisibility = klass.get("propertyVisibility");
-
         var existingFilters;
         if (this.store) {
             existingFilters = this.store.getFilters();
@@ -233,11 +230,9 @@ pimcore.object.search = Class.create(pimcore.object.helpers.gridTabAbstract, {
         this.store.getProxy().setExtraParam("only_direct_children", this.onlyDirectChildren);
         this.store.setPageSize(itemsPerPage);
 
-        if (existingFilters && this.classId == this.previousClassId) {
+        if (existingFilters && fromConfig) {
             this.store.setFilters(existingFilters.items);
         }
-
-        this.previousClassId = this.classId;
 
         var gridColumns = gridHelper.getGridColumns();
 
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/folder/search.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/folder/search.js
@@ -232,9 +232,12 @@ pimcore.object.search = Class.create(pimcore.object.helpers.gridTabAbstract, {
         }
         this.store.getProxy().setExtraParam("only_direct_children", this.onlyDirectChildren);
         this.store.setPageSize(itemsPerPage);
-        if (existingFilters) {
+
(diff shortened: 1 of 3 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/folder/search.js
```diff
@@ -232,10 +232,13 @@ pimcore.object.search = Class.create(pimcore.object.helpers.gridTabAbstract, {
         }
         this.store.getProxy().setExtraParam("only_direct_children", this.onlyDirectChildren);
         this.store.setPageSize(itemsPerPage);
-        if (existingFilters) {
+
+        if (existingFilters && this.classId == this.previousClassId) {
```
dvesh3: @weisswurstkanone how about if we keep the exisiting filter in case of `fromConfig=true`? and then everything else should reset the filter. 

Also, we should keep 'Filter active' & 'Clear Filter' buttons on config change.

weisswurstkanone: keep it even if the class changes ?

dvesh3: I mean:
```
if (!formConfig) {
    this.store.setFilters(existingFilters.items);
}	   
```

bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js
```diff
@@ -237,28 +237,31 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
...
+                hidden: true,
+                text: '<b>' + t("filter_active") + '</b>',
+                tooltip: t("filter_condition"),
+                handler: function (button) {
+                    Ext.MessageBox.alert(t("filter_condition"), button.pimcore_filter_condition);
+                }.bind(this)
+            });
+        }
 
+        if (!fromConfig) {
```
weisswurstkanone: same condition a second time

bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js
```diff
@@ -237,28 +237,31 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
...
+            this.clearFilterButton = new Ext.Button({
```
weisswurstkanone: button gets added in line 326. does it always exist?

dvesh3: yes, always exists and hidden. so now in case of change from grid configuration, we don't recreate the buttons.

dvesh3: I would also like to keep other filters intact in case of changing grid columns :)

PR #5810, opened 2020-02-14
[travis] build artifactes

weisswurstkanone:


blankse:
That only makes sense if a new baseline file is generated during the test, right?

weisswurstkanone:
it is just for archiving & inspecting the build artifactes, nothing more. similar to the test.log file

weisswurstkanone:
Example: so you could download the file and replace the repo version without running the tests.

image

blankse:
@weisswurstkanone But the file is the same as before the test, right? ;)
We must add the lines from the docs, which generate a new file:
https://pimcore.com/docs/6.x/Development_Documentation/Development_Tools_and_Details/Testing/Core_Tests.html#page_PHPStan-Baseline-Feature

weisswurstkanone:
should be the one that gets generated in $PROJECT_DIR

blankse:
@weisswurstkanone No new file is generated in the tests, it is only used.
Or where is the location where it is generated?

weisswurstkanone:
makes sense, do you want to add it ?

blankse:
Hmm, where should we add this? On every phpstan test? Then they have to execute the command twice and take twice as long... Or should we create our own job for it?

The idea with an bot like "Coding Standard Fixer" doesn't work?

weisswurstkanone:
@brusch  ?

weisswurstkanone:
maybe create a new job for now

brusch:
I assume the answer is now #5816 right? 😊

PR #5815, opened 2020-02-14
Fix Text Layout BC Break

### LINKED ISSUE #5814: Error with Text Layout after update to 6.5.0
I can't open data objects after update to 6.5.0 because we using the Text Layout with a rendering Class:

```
class DocumentationLink
{
    /** @var EngineInterface */
    protected $templating;

    public function __construct(EngineInterface $templating)
    {
        $this->templating = $templating;
    }

    public static function renderLayoutText($data, Concrete $object, $params): string
    {
        /** @var DocumentationLink $self */
        $self = \Pimcore::getKernel()->getContainer()->get(__CLASS__);

        return $self->templating->render(
            'template.html.twig',
            ['classIdentifier' => $object->getClassId()]
        );
    }
}
```

`Too few arguments to function DocumentationLink::__construct(), 0 passed in vendor/pimcore/pimcore/models/DataObject/ClassDefinition/Helper/ClassResolver.php on line 28 and exactly 1 expected`

Before it is only used static with call_user_func. Now it is used the ClassResolver which would create a object from it. We can migrate to this new service feature. But maybe its against your BC thought ;)

blankse:
Resolves #5814

### DIFF
--- a/models/DataObject/ClassDefinition/Helper/ClassResolver.php
+++ b/models/DataObject/ClassDefinition/Helper/ClassResolver.php
@@ -14,21 +14,19 @@ class ClassResolver
             if (isset(self::$cache[$class])) {
                 return self::$cache[$class];
             }
-            if (strpos($class, '@') === 0) {
-                $serviceName = substr($class, 1);
-                try {
+            try {
+                if (strpos($class, '@') === 0) {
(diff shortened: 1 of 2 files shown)


dvesh3:
thank you!

PR #5817, opened 2020-02-14
[Data Object] Improved lazy loading

### LINKED ISSUE #5791: [Data Object] Check if lazy loading could be improved
Currently the `load()` method of a data-type is called no matter if it is lazy loaded or not, maybe we can just skip this call in case lazy loading is active? 
https://github.com/pimcore/pimcore/blob/739957f849b8804683f71eef9be5c21493d932f5/models/DataObject/Concrete/Dao.php#L157

Something like that: 
```
if ($value instanceof CustomResourcePersistingInterface) {
    if(!$value instanceof LazyLoadingSupportInterface || !$value->getLazyLoading()) {
        // datafield has it's own loader
        $params = [
            'context' => [
                'object' => $this->model
            ]
        ];
        $value = $value->load($this->model, $params);
        if ($value === 0 || !empty($value)) {
            $this->model->setValue($key, $value);
        }
    }
}
```

This would also make the `force = true` param obsolete: 
https://github.com/pimcore/pimcore/blob/bf5070deb00e7d428f3c061878cbdc191751bc30/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php#L159
https://github.com/pimcore/pimcore/blob/bf5070deb00e7d428f3c061878cbdc191751bc30/models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php#L812

weisswurstkanone:
Resolves #5791

### DIFF
--- a/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
+++ b/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
@@ -698,7 +698,7 @@ class AdvancedManyToManyObjectRelation extends ManyToManyObjectRelation
         if ($object instanceof DataObject\Concrete) {
             $data = $object->getObjectVar($this->getName());
             if (!$object->isLazyKeyLoaded($this->getName())) {
-                $data = $this->load($object, ['force' => true]);
+                $data = $this->load($object);
 
                 $object->setObjectVar($this->getName(), $data);
                 $this->markLazyloadedFieldAsLoaded($object);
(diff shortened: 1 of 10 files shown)


brusch:
@NiklasBr @dpfaffenbauer do you see any possible issues with that? Thanks for you feedback! 😊

dpfaffenbauer:
I don't think so. Only thing that got me thinking was the 'force' part, but that is something each type handles for itself, right?

brusch:
Yep, that's why I think this would be fine to be merged.

NiklasBr:
I have no comments :)

dpfaffenbauer:
Not 100% sure if it is this PR, but it breaks loads of my tests:

https://travis-ci.com/coreshop/CoreShop/jobs/290537830

I think that cause some of the things have todo with loading relations or field-collections.

dpfaffenbauer:
OK, after locally fixing https://github.com/pimcore/pimcore/pull/5852, this PR seems to be fine afterall

PR #5821, opened 2020-02-14
[PHPStan] level 2 fixes

weisswurstkanone:


models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -185,7 +185,7 @@ public function getDimensions()
     }
 
     /**
-     * @return Asset
+     * @return Image
```
blankse: The Trait is also be used for Document and Video

weisswurstkanone: will look at this at a later time

models/User/UserRole.php
```diff
@@ -330,7 +330,7 @@ public function setWebsiteTranslationLanguagesEdit($websiteTranslationLanguagesE
      * checks if given parameter is string and if so splits it creates array
      * returns empty array if empty parameter is given
      *
-     * @param array $array
+     * @param array|string $array
      *
      * @return array|string
```
blankse:
```suggestion
     * @return array
```

PR #5830, opened 2020-02-17
Asset count

weisswurstkanone:


lib/Maintenance/Tasks/CleanupBrickTablesTask.php
```diff
@@ -56,6 +56,7 @@ public function execute()
                 $brickDef = Definition::getByKey($brickType);
                 if (!$brickDef) {
                     $this->logger->error("Brick '" . $brickType . "' not found. Please check table " . $tableName);
+                    continue;
```
brusch: This is not related to this PR, right?

PR #5832, opened 2020-02-17
[Bundles] Add install migration to available migrations

### LINKED ISSUE #5098: [Bundles] Install migration throws error on migrating updates
Extending `Pimcore\Extension\Bundle\Installer\MigrationInstaller` in a bundle's installer executes the install migration as version "00000001". 
https://github.com/pimcore/pimcore/blob/65720840f7219cb0d061c818eff5cfc0499b3d5c/lib/Extension/Bundle/Installer/MigrationInstaller.php#L351
(InstallVersion::INSTALL_VERSION is set to be 00000001).

The contents of this migration are fetched from the bundles' installer:
https://github.com/pimcore/pimcore/blob/65720840f7219cb0d061c818eff5cfc0499b3d5c/lib/Migrations/Migration/InstallMigration.php#L42-L45

This migration gets executed and added to the migrations database table.

Update migrations are fetched from the `migrations` directory in the bundle:
https://github.com/pimcore/pimcore/blob/65720840f7219cb0d061c818eff5cfc0499b3d5c/lib/Migrations/Configuration/ConfigurationFactory.php#L226-L234

When calculating the migrations of a bundle only the migrations in the `migrations` directory are used:
https://github.com/doctrine/migrations/blob/215438c0eef3e5f9b7da7d09c6b90756071b43e6/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php#L569-L582
(This is from doctrine/migrations 1.8.1 because Pimcore's composer contains `"doctrine/doctrine-migrations-bundle": "^1.3.1"` which currently resolves to 1.3.2. `doctrine/doctrine-migrations-bundle`1.3.2 requires `"doctrine/migrations": "^1.1"` which currently resolves to 1.8.1.

Therefore executing `bin/console pimcore:migration:migrate -b MyBundle -n` always complains:
```
WARNING! You have 1 previously executed migrations in the database that are not registered migrations.
    >>  (00000001)
```

Either we get the install migration to also be added to the list of available migrations or we exclude it when getting the already executed migrations in
https://github.com/pimcore/pimcore/blob/65720840f7219cb0d061c818eff5cfc0499b3d5c/lib/Migrations/Configuration/Configuration.php#L373-L383

BlackbitNeueMedien:
Currently the install migration gets added to `pimcore_migrations` table but in https://github.com/pimcore/pimcore/blob/65720840f7219cb0d061c818eff5cfc0499b3d5c/lib/Migrations/Configuration/Configuration.php#L373-L383 only migrations in the migration folder get used to determine available migrations. This leads to the following error when trying to execute migrations after the inital installation:
```
WARNING! You have 1 previously executed migrations in the database that are not registered migrations.
    >>  (00000001)
```

This PR adds the installer migration to the list of available migrations.

To reproduce bug:
1. Install bundle with installer which extends `Pimcore\Extension\Bundle\Installer\MigrationInstaller`
1. Execute `bin/console pimcore:migration:status -b <Bundle name>`
`Executed Unavailable Migrations` should be `0`.
1. Alternatively execute `bin/console pimcore:migration:migrate -b <Bundle name>`
Above error should not appear.

Resolves #5098

### DIFF
--- a/lib/Migrations/Configuration/Configuration.php
+++ b/lib/Migrations/Configuration/Configuration.php
@@ -383,6 +383,16 @@ class Configuration extends \Doctrine\DBAL\Migrations\Configuration\Configuratio
         return $newMigrations;
     }
 
+    public function getAvailableVersions()
+    {
+        $migrations = parent::getAvailableVersions();
+        if ($this->installer !== null) {
+            $migrations[] = $this->installer->getMigrationVersion();
+        }
+        return $migrations;
+    }
+
+
     /**
      * Handles simple placeholder handling in query. Makes queries more readable as we need to replace the configurable
      * columns in every query.


brusch:
Nice, thanks! 👍

PR #5835, opened 2020-02-17
[Users] Type-cast parent id to int

BlackbitNeueMedien:
I just stumbled about this when trying to get all users below a certain role folder via 
```php
$roleListing = new Listing();
$roleFolder = Role\Folder::getByName('abc');
$roles = array_filter($roleListing->load(), static function(AbstractUser $role) use ($roleFolder) {
    return $role->getParentId() === $roleFolder->getId();
});
```


brusch:
Thanks! 🙏

PR #5837, opened 2020-02-17
Asset|Document Listing getCount with Limit and Offset always returns 0

### LINKED ISSUE #5639: Asset/Listing getCount with Limit and Offset always returns 0
# Bug Report

### Expected behavior
Asset\Listing() ... $assetListing->getCount() with set Limit/Offset values should return correct assets count.

### Actual behavior
Asset\Listing() getCount() with Limit/Offset returns 0

### Steps to reproduce
- Create Asset\Listing instance
- Set Limit and Offset values
- Load list
- call getCount() on asset listing instance

### Code example
```php
$assetList = new Asset\Listing();
$assetList->setOffset(5);
$assetList->setLimit(10);

$assetList->load();
$x = $assetList->getCount();
// $x = 0
```

### Solution / additional informations
Not sure whats the expected solution, so no PR

If its a condition, that always before $myAssetList->getCount(), $myAssetList->load() must be called, then it just should return ``` count($this->model->getAssets())``` and not execute another query. 

If the current system with it's own query for getCount() has to stay this way, then the query in getCount() (models/Asset/Listing/Dao.php) should be reseted with:

```php
$select = $this->getQuery([new Expression('COUNT(*)')]);
$select->reset(QueryBuilder::LIMIT_COUNT);
$select->reset(QueryBuilder::LIMIT_OFFSET);
```

weisswurstkanone:
Resolves #5639

### DIFF
--- a/models/Asset/Listing/Dao.php
+++ b/models/Asset/Listing/Dao.php
@@ -99,10 +99,12 @@ class Dao extends Model\Listing\Dao\AbstractDao
      */
     public function getCount()
     {
-        $select = $this->getQuery([new Expression('COUNT(*)')]);
-        $amount = (int) $this->db->fetchOne($select, $this->model->getConditionVariables(), $this->model->getConditionVariableTypes());
-
-        return $amount;
+        if ($this->model->isLoaded()) {
+            return count($this->model->getAssets());
+        } else {
+            $idList = $this->loadIdList();
+            return count($idList);
+        }
     }
 
     /**
--- a/models/Asset/Listing/Dao.php
+++ b/models/Asset/Listing/Dao.php
@@ -99,11 +99,10 @@ class Dao extends Model\Listing\Dao\AbstractDao
      */
     public function getCount()
     {
-        if (count($this->model->getAssets()) == 0) {
-            $this->load();
-        }
+        $select = $this->getQuery([new Expression('COUNT(*)')]);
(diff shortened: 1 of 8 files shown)


models/Asset/Listing/Dao.php
```diff
@@ -99,10 +99,11 @@ public function loadIdList()
      */
     public function getCount()
     {
-        $select = $this->getQuery([new Expression('COUNT(*)')]);
-        $amount = (int) $this->db->fetchOne($select, $this->model->getConditionVariables(), $this->model->getConditionVariableTypes());
+        if (count($this->model->getAssets()) == 0) {
```
dpfaffenbauer: isn't that way slower? Cause it needs to hydrate the models.

weisswurstkanone: same approach as for objects. at least it produces the correct result

dpfaffenbauer: Why don't you use `getQuery` and change it up to make a COUNT(*) query?

https://github.com/pimcore/pimcore/blob/9167b96ed44aef0d5f029be73d120faec972f3a8/models/Asset/Listing/Dao.php#L64

dpfaffenbauer: just realised, that is exactly what has been proposed in #5639

PR #5839, opened 2020-02-17
[Ecommerce][Productlist] Consider ElasticSearch reverse nested aggregations

jremmurd:
# New Feature

The product list converts the aggregation bucket values returned from elastic search to "value" and "count" pairs. If I want to use nested aggregations, but only count the source documents I have to use reverse nested aggregations. The returned values in this type of aggregation gets lost in the `convertBucketValues()` method. 

This PR considers the values and adds an additional field `reverse_count`.

image

Also see https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-aggregations-bucket-reverse-nested-aggregation.html


brusch:
Thanks a lot! 🙋‍♂️

PR #5842, opened 2020-02-17
Fix HardlinkController setValuesToDocument function

davidhoeck:
Query all documents by `path` and not only hardlinks


brusch:
Regression of #5777

brusch:
Thanks a lot! 🙏

PR #5844, opened 2020-02-18
[Redirects] Filter by Test-URL

### LINKED ISSUE #5616: [Redirects] Filter by Test-URL
Inside the redirects list it would be cool if we could enter a test-URL and filter (or mark) only redirect entries that match that URL. 

This way it would be easy to find problematic redirects. Endusers or SEO's could test if their regex redirect works correctly quickly.

brusch:
Resolves #5616

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/RedirectsController.php
+++ b/bundles/AdminBundle/Controller/Admin/RedirectsController.php
@@ -22,7 +22,9 @@ use Pimcore\Bundle\AdminBundle\HttpFoundation\JsonResponse;
 use Pimcore\Logger;
 use Pimcore\Model\Document;
 use Pimcore\Model\Redirect;
+use Pimcore\Model\Site;
 use Pimcore\Routing\Redirect\Csv;
+use Pimcore\Routing\RedirectHandler;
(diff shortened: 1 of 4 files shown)


bundles/AdminBundle/Controller/Admin/RedirectsController.php
```diff
@@ -131,6 +134,16 @@ public function redirectsAction(Request $request)
...
                     $list->setCondition('id = ?', [$filterValue]);
+                } elseif(preg_match('@^https?://@', $filterValue)) {
+                    $dummyRequest = Request::create($filterValue);
+                    $site = Site::getByDomain($dummyRequest->getHost());
+                    $dummyResponse = $redirectHandler->checkForRedirect($dummyRequest, false, $site);
+                    if($dummyResponse && $redirectId = $dummyResponse->headers->get(RedirectHandler::RESPONSE_HEADER_NAME_ID)) {
+                        $list->setCondition('id = ?', [$redirectId]);
+                    } else {
+                        // do not return any results
+                        $list->setCondition('1 = 2');
```
dpfaffenbauer: srsly?

PR #5847, opened 2020-02-18
[Mail] Update docs

### LINKED ISSUE #5785: Pimcore Mail does not provide mail params in view model
# Bug Report

The documentation https://pimcore.com/docs/6.x/Development_Documentation/Development_Tools_and_Details/Email_Framework/Pimcore_Mail.html states that `setParams` sets the parameters for both placeholders and the email view. However, the view can not access these parameters as they aren't set

### Expected behavior

Mail param `foo` can be accessed in template by using `$this->foo`.

AFAICR that was the behavior in Pimcore 4.

### Actual behavior

View model does not contain key `foo` and thus cannot be accessed from within template (in this case PHP template).

However, the params are accessible through request's attributes (`$request->attributes`) from e.g. within controller action.

### Steps to reproduce

1) Prepare and send a mail with \Pimcore\Mail using params and a document: 
```php
$mail = new \Pimcore\Mail();
$mail->addTo('example@pimcore.org');
$mail->setDocument('/email/myemaildocument');
$mail->setParams(['foo' => 'bar');
$mail->send();
```

2) Create email document `/email/myemaildocument` and set `FooController` and `fooAction`.
3) Have `FooController` and `fooAction` render a template `foo.html.php`.
4) Template `foo.html.php` tries to access `$this->foo`, but its unset and thus null.

### Crude workaround

Set some/all request's attribute keys as view model variables.

```php
    public function onKernelController(FilterControllerEvent $event)
    {
        parent::onKernelController($event);

        // set params for view
        $request = $this->requestHelper->getCurrentRequest();
        $viewParams = $request->attributes->all();
        foreach ($viewParams as $key => $value) {
            $this->view->$key = $value;
        }
    }
```

---
Using Pimcore 6.4.2

dvesh3:
Resolves #5785

### DIFF
--- a/lib/Mail.php
+++ b/lib/Mail.php
@@ -409,7 +409,7 @@ class Mail extends \Swift_Message
         if (is_string($key) || is_integer($key)) {
             $this->params[$key] = $value;
         } else {
-            Logger::warn('$key has to be a string - Param ignored!');
+            Logger::warn('$key has to be a string or integer - Param ignored!');
         }
 
         return $this;
@@ -475,7 +475,7 @@ class Mail extends \Swift_Message
         if (is_string($key) || is_integer($key)) {
             unset($this->params[$key]);
         } else {
-            Logger::warn('$key has to be a string - unsetParam ignored!');
+            Logger::warn('$key has to be a string or integer - unsetParam ignored!');
         }
 
         return $this;
--- a/lib/Mail.php
+++ b/lib/Mail.php
@@ -381,7 +381,7 @@ class Mail extends \Swift_Message
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lib/Mail.php
```diff
@@ -397,7 +397,7 @@ public function setParams(array $params)
     }
 
     /**
-     * Sets a single parameter for the email view and the Placeholders
+     * Sets a single parameter to the request object and the Placeholders
```
weisswurstkanone: Please adapt the warning message as the code says $key is allowed to be a string

Logger::warn('$key has to be a string - Param ignored!');

PR #5849, opened 2020-02-18
[Users] Doctype: Return type for getBy*

BlackbitNeueMedien:
When you call `User::getById()` you get a `User` object. When you call `Role::getById()` you get a `Role` object etc.


brusch:
Nice, thanks a lot!

PR #5850, opened 2020-02-18
[Data objects] unique index for field type "User"

BlackbitNeueMedien:
Currently it seems to be only possible for Input and Numeric fields to add a unique index as they are the only classes which override 
https://github.com/pimcore/pimcore/blob/453a87daed484e85ad00120e264f5a04def5492b/models/DataObject/ClassDefinition/Data.php#L1197-L1200

This PR adds the possibility to create unique indexes also for `User` datatype.

I wonder why unique index is currently only possible for Numeric and Input datatype because the unique checkbox actually is present for all datatypes.


brusch:
Thanks a lot!

brusch:
Follow up issue for cleanup other data-types: #5861

PR #5852, opened 2020-02-18
[DataObjects] Refactor multiple assignments check for Advanced relation types

### LINKED ISSUE #5841: [DataObjects] Update. Move multiple assignements check to responsible datatype

dvesh3:
Resolves #5841

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -337,6 +337,41 @@ class Service extends Model\AbstractModel
     }
 
     /**
+     * @internal trigger deprecation error when a relation is passed multiple times, remove in Pimcore 7
+     * @param array $data
+     *
+     * @return array
+     *
+     * @throws \Exception
+     */
+    public static function filterMultipleElements($data)
+    {
+        $relationItems = [];
+        if (is_array($data)) {
+            foreach ($data as $item) {
+                $elementHash = null;
+                if ($item instanceof Model\DataObject\Data\ObjectMetadata || $item instanceof Model\DataObject\Data\ElementMetadata) {
+                    if ($item->getElement() instanceof Model\Element\ElementInterface) {
+                        $elementHash = Model\Element\Service::getElementHash($item->getElement());
+                    }
+                } elseif ($item instanceof Model\Element\ElementInterface) {
+                    $elementHash = Model\Element\Service::getElementHash($item);
+                }
+
+                if ($elementHash && isset($relationItems[$elementHash])) {
+                    @trigger_error(
+                        'Passing relations multiple times is deprecated since version 6.5.2 and will throw exception in 7.0.0, tried to assign ' . $elementHash .  ' multiple times.',
+                        E_USER_DEPRECATED
+                    );
+                }
+                $relationItems[$elementHash] = $item;
+            }
+        }
+
+        return array_values($relationItems);
+    }
+
+    /**
      * @static
      *
      * @param  string $type
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -339,14 +339,31 @@ class Service extends Model\AbstractModel
     /**
      * @internal trigger deprecation error when a relation is passed multiple times, remove in Pimcore 7
      * @param array $data
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|\Pimcore\Model\DataObject\Fieldcollection\Data\AbstractData $object
+     * @param string $fieldname
      *
      * @return array
      *
      * @throws \Exception
      */
-    public static function filterMultipleElements($data)
+    public static function filterMultipleElements($data, $object, $fieldname)
     {
         $relationItems = [];
+        $objectId = null;
+
+        if ($object instanceof DataObject\Concrete) {
+            $objectId = $object->getId();
+        } elseif (
+            $object instanceof DataObject\Fieldcollection\Data\AbstractData ||
+            $object instanceof DataObject\Localizedfield ||
+            $object instanceof DataObject\Objectbrick\Data\AbstractData
+        ) {
+            $object = $object->getObject();
+            if ($object) {
+                $objectId = $object->getId();
+            }
+        }
+
         if (is_array($data)) {
             foreach ($data as $item) {
                 $elementHash = null;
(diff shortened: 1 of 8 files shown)


models/Element/Service.php
```diff
@@ -336,6 +336,60 @@ public static function filterUnpublishedAdvancedElements($data)
...
+     */
+    public static function filterMultipleElements($data, $object, $fieldname)
+    {
+        $relationItems = [];
+        $objectId = null;
+
+        if ($object instanceof DataObject\Concrete) {
+            $objectId = $object->getId();
+        } elseif (
+            $object instanceof DataObject\Fieldcollection\Data\AbstractData ||
```
weisswurstkanone: does it work for blocks?

dvesh3: fixed for Blocks

dpfaffenbauer:
Ok, after hours of debugging, this PR is the reason 6.5.2 is bloody useless again....

Guys, if you do duplication checks, do them right, check if the item actually has an ID or not. `getElementHash` doesn't check if the Element has an ID. So, if you put two items without ID into an many-to-many relation, guess what happens.... YES: One of them gets removed, cause the hash is the same.

What actually is with your tests? This should be picked up. This is so crucial and still it happens ALL THE TIME you break major important things.

This needs to stop, why do keep the CoreShop tests or also @solverat tests fail and yours not on such an important piece?

dpfaffenbauer:
BTW: since when does Pimcore trigger anything on setters? 

Why not check that on save, trigger the notice there, where there has to be an ID. This PR is completely wrong and should never been had merged as it is. And also, this is a very crucial parts, where are tests if the hash function actually works?

dvesh3:
@dpfaffenbauer what is your usecase for "if you put two items without ID into an many-to-many relation"? 😕 why you want to do that?

and if you feel something is missing then you are most welcome to contribute 😉 

Also, we didn't change anything in `getElementHash` and it's just refactoring so that multiple assignment filter should work in container types as before it was not the case.

dpfaffenbauer:
@dvesh3 The use case is to save database roundtrips, and also prohibit to store unnecessary data in Pimcore. CoreShop **ALWAYS** creates an empty Cart Object for every visitor, BUT it doesn't persist it right away, that would create millions of empty carts. When the customer adds something to the cart, we create a cart-item that get's added to the cart, then it gets saved. 

That still works with your change BUT imagine something like a Multiple add-to-cart behaviour where the customer adds like 10 items to the cart, the cart itself is not persisted yet, thus has no ID, the cart-items get's all added to the cart using `setItems`, all of them have **NO** `ID` yet. So, then, after everything is calculated, the `CartManager` persists the cart. It temporarily removes and backups the items, saves the cart, saves the items, set's the items and saves the cart again. A bit ridiculous, but that is cause of how the relation works and how Pimcore needs to have a path for childs to be saved.

That has many advantages:

 1. We save loads of DB roundtrips, cause we don't need to save the cart first, save the item, calculate stuff, save the item again, call `setItems` and save the cart again
2. Doesn't create millions of unnecessary empty carts

True, you didn't change `getElementHash`, you changed where it gets called. Previously that happened in `update` on the data-object, where the `cart-item already` has an `ID`. There it **MUST** have an `ID`. You moved it to the setter, and the setter can **NOT** be responsible for that. 

You also have a **TODO** in there, that says to move that into the **validation**, but the validation also just gets called on `save`, where it belongs....

weisswurstkanone:
@dpfaffenbauer It feels like that there is an unresolved problem. Could you please open an issue for that? Referencing your comment here should be sufficient.

We agree that the setter shouldn't trigger the filtering, so please describe how you fixed it "locally" (I think you mentioned this in the other PR) and provide a PR fixing the issue.

Besides that, the policy should be that if you find bugs then contribute a test covering the issue.
This would also help others.

Thanks a lot.

brusch:
Just to clarify things, this is really kind of an edge-case and maybe also in a way subject to interpretation. But as always we're here to help, improve, discuss and fixing things quickly. 

I also want to state here that this is not the right place for publicly blaming people or the project owners. We maintain a friendly and collaborative approach in our community, all in all we are in the same boat and want to enjoy our daily work.
Don't get me wrong, we highly appreciate your professional work and your advice is more than welcome and so I think it would be for the benefit of all if we keep our way to communicate in a cooperative and respectful manner.

dpfaffenbauer:
@brusch Edge case: Not sure, subject of interpretation: definitely .

Just to clarify as well:
I didn't blame anyone at all at anytime, or at least didn't intend to blame anyone directly. I may seem harsh from time to time, but it is quite frustrating working with Pimcore for a couple of weeks/months now. More and more stuff gets merged without being tested properly or without having tests at all. Too many things get broken, with comments like: "edge-cases" or "Pimcore is not responsible".  I mean, I can recall at least one major bug for every version since 6.3.6. 

So currently, no version except 6.3.6 is usable in Production properly. And not even 6.3.6 cause you broke the ability to delete DataObjects which have a relation to Doctrine Entities. (And no, doctrine ORM in relation to Pimcore DataObjects should not be seen as an EdgeCase or anything, it is de-facto the standard ORM for PHP). And this is not the first discussion we have regarding this matter. 

The quality of Pimcore is getting worse... that is my main concern. And you're not doing anything against that trend. How can it be that community provided bundles test Pimcore better than Pimcore tests itself? 

You just keep adding stuff, without testing. Sometimes so obvious that it feels like you don't even test it manually... You keep on adding features poorly designed and buggy, maybe even knowing that there might be a better way. Writing tests isn't that hard. It just needs time... And that might be the issue here.... I don't know. And it improves the quality by so much. 

Regarding this specific issue: I don't see that as an edge case at all. I don't think that everytime someone calls a setter on a many-to-many relation in Pimcore, that the items are already saved. Another popular use-case for that scenario are Tests. I don't have to save anything to make assumption on them. 

Also, statements like this:

> and if you feel something is missing then you are most welcome to contribute 😉

and

> Besides that, the policy should be that if you find bugs then contribute a test covering the issue.
This would also help others.

NO. I don't think so. Appreciate that I just spent couple of hours, unpaid, to find what is going. That saves you at least the time I spent to fix it. Appreciate the help of the community, but don't force it on them. Be happy if they do help and commit PR's, and thankfully, they do quite a lot now, but still, nobody is obligated to do so. The policy rather should be: For every change, especially in the PIM part, needs to have a test coming along with it, testing the change and the new behaviour. If you create a PR, where you change crucial parts, and no tests fail, something gotta be wrong. (Not talking about implementing new features here, completely different approach to that)

Listen, Pimcore is great, and I don't wanna blame it all in any way. I just want to make it a better Product, just as you do. And I think I do a pretty good job with I do and supply to the Community (also for free). Maybe even more than anyone else (except for @solverat, which does an amazing job there as well). 

Pimcore needs to evolve more, it needs more tests (like a thousand more tests). It needs refactoring to get rid of loads of legacy stuff. It needs better/deeper Symfony integration. (eg. Symfony Flex) It maybe also needs Doctrine ORM (not for DataObjects, at first ;)). 

But little in that matter happened since @maff left, unfortunately. And the main point here, that evolvement needs to start from YOUR side, not the community. 

So yeah, just my two cents on current development of Pimcore.

brusch:
Thanks for your feedback. 
I agree that we have a couple of challenges and the whole team knows about that. We accepted the challenge and so we're continuously working on improving the situation not just with tests but also with improved workflows and other things. But also more and more tests are getting added, just have a look here: https://github.com/pimcore/pimcore/commits/master/tests 

In my opinion the situation is not as worse as you illustrate, but I respect your view on that and we got your point. Last but not least, please let me ask you to not repeat this here in the issue tracker over and over again, it's just not the right place for this kind of discussions.

Thanks for your understanding, keep on the good work and let's make Pimcore better together! 🤝💪

dpfaffenbauer:
If you stop repeating breaking Pimcore, I will stop as well ;)

To get back to the issue: Is it gonna get fixed or not?

brusch:
Yes. Well keep you updated here #5902

PR #5855, opened 2020-02-18
Read access for protected object properties

BlackbitNeueMedien:
More and more object properties got and will get converted from `public` to `protected` access. Of course this makes sense so nobody can inject any garbage into the objects. On the other hand it breaks a lot of legacy code. In the last time we see more and more customers upgrading to Pimcore 6 and a lot of applications use object property access instead of getters. I do not know why the original application developers decided to not use the getter methods but actually it was possible to use property access - and so they did use it.

While accessing the value of object properties result in a fatal PHP error and so are quite easy to find (at least when you have lots of tests - automatic or manual) , it cost us 5 hours to find this line:
```php
if(empty($fieldCollectionItem->name)) {
...
}
```
The problem here is that while the `name` property was `public` `empty($fieldCollectionItem->name)` returned `false` if a name had been set to the name field of the field collection. Now as it is `protected` this simply returns `true` and consequently does go into this if-block - this completely inverts the logic without any error, log or anything.

It is very difficult to find such code occurences, so I thought if there was a way to support developers during Pimcore version upgrade.
At first I thought to only implement magic `__isset()` and `__get()` methods with a deprecation notice (because `empty()` calls `__get()` when `__isset()` returns `true`). But in the end I wondered whether those deprecation notices actually get read in the logs (because there would actually be no need for developers to look at the logs while the system works as expected).
Then I went one step back and thought that the goal of setting object property access to `protected` was to prevent **setting** unexpected values to those properties. So imho there is no reason to not allow **read access** via object properties. To ensure correct data handling the magic `__get()` calls the actual getter method instead of returning the property value directly.

### DIFF
--- a/lib/Model/AbstractModel.php
+++ b/lib/Model/AbstractModel.php
@@ -284,6 +284,7 @@ abstract class AbstractModel implements ModelInterface
         if(\method_exists($this, $getter)) {
             return $this->$getter();
         }
+        return null;
     }
 
     /**
--- a/lib/Model/AbstractModel.php
+++ b/lib/Model/AbstractModel.php
@@ -273,6 +273,19 @@ abstract class AbstractModel implements ModelInterface
         return $result;
     }
 
+    public function __isset($name)
+    {
+        return isset($this->$name);
+    }
+
+    public function __get($name)
+    {
+        $getter = 'get'.ucfirst($name);
+        if(\method_exists($this, $getter)) {
+            return $this->$getter();
+        }
+    }
+
     /**
      * @return Factory
      */


dpfaffenbauer:
No, that is a step backwards! Refactor your legacy application instead!

BlackbitNeueMedien:
I actually do want to refactor it but how to find such cases as described above without having to check every code line - especially when there are no automated tests?
I would also be fine with adding a deprecation notice and remove those `__get()` and `__isset()` in Pimcore 7...

dpfaffenbauer:
what was the point of making them protected when you want to access them directly now again anyway?

BlackbitNeueMedien:
As said above: The goal of setting object property access to protected was to prevent **setting** unexpected values to those properties and also guarantee correct data handling (e.g. for lazy-loaded fields)

dpfaffenbauer:
yes, and this solution would introduce the same problem again...

BlackbitNeueMedien:
Why? You cannot set any value to the object properties with __get(). That is the whole point of this PR. We allow read access (including correct data handling through getter methods) while preventing write access to those object properties.

dpfaffenbauer:
right, you are calling the getter. 

Still, I don't like that solution. Pimcore went the way knowing it is a BC break... if you need that in your application, you can add that magic function to your data-objects or even other models, but this should not get back to the core. it introduces a magic method of accessing properties that is not necessary to have (since we have the getters anyway). I don't see any valid reason of adding that.

If you're about to refactor an older application, that doesn't help you at all, since you will run into the same problem again in the future. (when that is getting removed). you are basically just delaying that problem for you. and, if you add in a deprecation note, you still need to go through the whole website and check for usages and read a bloody long log file with thousands of deprecation notices from symfony and some from pimcore. So that doesn't help a lot as well...

BlackbitNeueMedien:
`If you're about to refactor an older application, that doesn't help you at all, since you will run into the same problem again in the future. `
That is the question whether this needs to be removed in the future. I do not see a disadvantage of allowing property read access via __get(). 
It would help lots of developers in upgrading their legacy applications...

brusch:
If you ask me personally, I have to honestly say that I wouldn't introduce that possibility again. 
I see too many risks there, just for example: 
- it's not consistent, ... container data-types such as field collections, LF, ... do not support it
- also other entities are not supporting it anymore
- it may causes some issues in the future, since we have to support 2 different ways of how data can be accessed 
...
...

But that's just my 2 cents for the moment, we can keep the discussion open and wait also for other opinions 😊

BlackbitNeueMedien:
Ok, that is really frustrating... 
Could you give me a hint how you would try to find such things like `empty($object->fieldName)`? It is hard to accept that the way to go is to look through the legacy code line by line.

Throwing an exception in `__isset()` or `__get()` would also not be a way to go?

@brusch, what do you mean with `container data-types such as field collections, LF, ... do not support it` and which `other entities are not supporting it anymore`?
Actually I also do not see that there are 2 different ways to access the data. It remains one way: through the getter method - this PR only introduces a wrapper to this getter...

aarongerig:
@BlackbitNeueMedien Why not search your code for something like `->property` with a regex? I'm sure that way it would be pretty easy to find all occurrences of direct property access usages.

brusch:
One very easy and pragmatic way to quickly fix those issues, would be to put this code into a trait and apply it to your class definitions, job done 😊

To fix the issues I would log the usage into a dedicated file including the file and line number where it was called. If you have loads of usages you can also think about logging them to a temporary database table, to make the list a bit easier to group/filter. 

Regarding your questions: 
1. With this it's not possible to use something like `$object->bricks->myBrick->myField` that's what I meant with consistency
2. Other entities, e.g. `$asset->filename` is also not possible.

BlackbitNeueMedien:
@aarongerig I tried `->\w+[^\(]` but that also finds method calls in PhpStorm (it does not highlight the bracket in this case but nevertheless there are too many false positives). And even if we found a regex which does match only object property access this would also find lots of internal class property access like `$this->field` which may be perfectly fine in custom application classes.

Thanks @brusch - we will do that: Create a trait and add that to data object classes...

brusch:
@BlackbitNeueMedien if you implement the logging as well, it would be nice if you can share the code here, maybe it's useful for others as well 😊

BlackbitNeueMedien:
@brusch regarding your answers:
`With this it's not possible to use something like $object->bricks->myBrick->myField that's what I meant with consistency`
I do not understand why this is not possible. With this PR this would call `$object->getBricks()->getMyBrick()->getMyField()` because for every field which is not accessible from the current scope the getter method would be called.

`Other entities, e.g. $asset->filename is also not possible.` I do not understand that either. Why is this not possible with this PR? When you try to access Asset::filename from a scope which does not have access to it, the `__get()` method gets called which calls `$asset->getFilename()`.

brusch:
@BlackbitNeueMedien You're absolutely right, shame on me 🤦‍♂️. I thought you put the methods on `AbstractObject` 😅sorry, for the confusion.

dpfaffenbauer:
@BlackbitNeueMedien it still is unnecessary magic that doesn't belong in there anymore.

BlackbitNeueMedien:
One last question remained unanswered: Throwing an exception in __isset() and __get() would also not be a way to go? 

If only we had this idea when Pimcore decided to change the access level to `protected`, we could have done that without BC break...

brusch:
What would be the benefit of an exception as there's anyway an error thrown (which is the correct behavior)?

BlackbitNeueMedien:
@brusch That was exactly the problem: With empty() or isset() there is no error thrown. Here again our problem from yesterday:
```php
if(empty($fieldCollectionItem->name)) {
...
}
```
The problem here is that while the `name` property was `public` `empty($fieldCollectionItem->name)` returned `false` if a name had been set to the name field of the field collection. Now as it is `protected` this simply returns `true` and consequently does go into this if-block - this completely inverts the logic without any error, log or anything.

dpfaffenbauer:
did you check the error log for notices? AFAIK PHP ignores missing properties and just throws a notice, doesn't it?

BlackbitNeueMedien:
Yes, I checked the logs, there is really nothing, see https://3v4l.org/KeiYp

dpfaffenbauer:
Alright, yes, true, cause `empty` calls `isset` and `isset` checks if the property is accessable. if you remove the empty, you'll see the error

BlackbitNeueMedien:
That's right, when you just try to access the value, you will get an error. 
But when upgrading a legacy Pimcore you actually do not know if somewhere there are those isset() or empty() calls. You only recognize it in inverted logic - if you recognize it at all ;-)

What do you think of throwing an exception in these cases?

dpfaffenbauer:
still don't like that. that is exactly a good reason why you shouldn't use empty, use `null !==` instead.

Pimcore introduced these BC breaks in a lot of versions ago, lot's of people already adapted to it and got used to it.

for your upgrade path: you have to go through the project anyway and test everything.... I don't get why that is a problem.

PR #5856, opened 2020-02-19
[Data Object] Automatically create redirects for changed URL slugs

### LINKED ISSUE #5767: [Data Object] Automatically create redirects for changed URL slugs
Related to #5766 which has to be done first.

brusch:
Resolves #5767

### DIFF
--- a/models/DataObject/ClassDefinition/Data/UrlSlug.php
+++ b/models/DataObject/ClassDefinition/Data/UrlSlug.php
@@ -21,6 +21,7 @@ use Pimcore\Db;
 use Pimcore\Logger;
 use Pimcore\Model;
 use Pimcore\Model\DataObject\ClassDefinition\Data;
+use Pimcore\Model\Redirect;
 
 class UrlSlug extends Data implements CustomResourcePersistingInterface, LazyLoadingSupportInterface
 {
@@ -91,8 +92,6 @@ class UrlSlug extends Data implements CustomResourcePersistingInterface, LazyLoa
     public function getDataForEditmode($data, $object = null, $params = [])
     {
         $result = [];
-
-        // for now we don't support sites (=> there is just a plain input field in the UI)
         if (is_array($data)) {
             foreach ($data as $slug) {
                 if ($slug instanceof Model\DataObject\Data\UrlSlug) {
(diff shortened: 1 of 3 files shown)


models/DataObject/ClassDefinition/Data/UrlSlug.php
```diff
@@ -286,6 +290,42 @@ public function save($object, $params = [])
                 }
             }
         }
+
+        // check for previous slugs and create redirects
+        if(is_array($data)) {
```
dpfaffenbauer:
```
if(!is_array($data)) {
     return;
}
```

models/DataObject/ClassDefinition/Data/UrlSlug.php
```diff
@@ -286,6 +290,42 @@ public function save($object, $params = [])
...
+            foreach($data as $slug) {
+                if($previousSlug = $slug->getPreviousSlug()) {
```
dpfaffenbauer:
```
if($previousSlug !== $slug->getPreviousSlug()) {
    continue;
}
```

brusch: I guess you mean `if($previousSlug === $slug->getSlug()) {` right?

models/DataObject/Data/UrlSlug.php
```diff
@@ -132,11 +137,33 @@ public function getSlug(): ?string
      */
     public function setSlug(?string $slug)
     {
+        if($this->slug && !$this->previousSlug && $slug !== $this->slug) {
+            $this->previousSlug = $this->slug;
```
weisswurstkanone: wouldn't it be better to do this in createFromDataRow?

brusch: But there it's not covering API updates ... so I guess this is the best place, right?

weisswurstkanone: Would still say that previousSlug should only be set when loaded from resource. It should be kind of a dirty detection, which causes a redirect, right ?

brusch: Yep, now I got your point 😉

brusch:
@dvesh3 regarding the need of calling `saveToSession()` on publish: 

In my opinion in would be sufficient to call `Service::removeElementFromSession()` here, since the latest content is then anyway stored in the published document: 
https://github.com/pimcore/pimcore/blob/9b0a4ebdc831dbd5cc21d314388ee34b8d4f8703/bundles/AdminBundle/Controller/Admin/Document/PageController.php#L171

But we can leave it as it is for now, as it doesn't hurt and the session is anyway cleaned up as soon as the document gets closed.

PR #5857, opened 2020-02-19
Tri-State-Checkbox update docs

ultramegatom:
Add steps to get the same behavior before the checkbox update


brusch:
Thanks a lot! 👍

PR #5858, opened 2020-02-19
Clean up data that is not really necessary for a clone

weisswurstkanone:


brusch:
@weisswurstkanone what about: 
- (`o_`) `dependencies`
- scheduledTasks
- (`o_`) userPermissions
? 

For assets maybe additionally `stream`
For documents: `fullPathCache` 
Just for the sake of completeness?

weisswurstkanone:
Not sure if wen can drop userPermissions

brusch:
Dropped user permission in #5864

PR #5860, opened 2020-02-19
[Data Object] Do not swallow exception in getData

weisswurstkanone:


brusch:
@weisswurstkanone does that help, because the exception is caught here anyway? 
https://github.com/pimcore/pimcore/blob/663b0887cf4623b5499987e2b15b9522f3342e11/models/DataObject/AbstractObject.php#L317

weisswurstkanone:
before: you were able to open the object even if something went wrong during loading of the data
after: object cannot be opened anymore because (as you point out correctly) object is null.
but:  maybe this needs further improvement because now no logging is done at all.

dpfaffenbauer:
When an exception is thrown here, something went terribly wrong, why not pass the exception on? How would somebody ever find out that something happened during hydration?

Of course, that should be done on 7.0 only, since otherwise it is a BC break ;)

weisswurstkanone:
So currently if something wents terribly wrong then this is results in data loss if the object is saved again. Compared to a BC break, well, ...

brusch:
What about Documents and Assets, does the same apply there? Thanks!

weisswurstkanone:
I would postpone this change.

brusch:
Agree, scheduled for v7. But behavior should be unified across all entities.

PR #5864, opened 2020-02-19
[Elements] get rid of of `userPermission` property and setter + code cleanup

brusch:


weisswurstkanone:
Because you are touching this piece of code. Can you fix this stuff as well (i.e. not set the property but array val instead)

 ------ ------------------------------------------------------------------ 
  Line   bundles/AdminBundle/Controller/Admin/AssetController.php          
 ------ ------------------------------------------------------------------ 
  91     Access to an undefined property Pimcore\Model\Asset\Text::$data.  
  93     Access to an undefined property Pimcore\Model\Asset\Text::$data.  
  96     Access to an undefined property                                   
         Pimcore\Model\Asset\Document::$pdfPreviewAvailable.               
  117    Access to an undefined property                                   
         Pimcore\Model\Asset\Video::$videoInfo.                            
  138    Access to an undefined property                                   
         Pimcore\Model\Asset\Image::$imageInfo.

PR #5865, opened 2020-02-19
[Data Object] Numeric Data-Type: fix schema if 'decimalPrecision' => ''

brusch:
If `'decimalPrecision' => ''` and `'decimalSize' = NULL` the resulting column type is `decimal(64,0)` which is wrong, correct would be the default type which is `double`.


brusch:
Should be backported to v5 as well.

PR #5867, opened 2020-02-19
[Document] "Create Page" dialog: title should be the first field, the rest should be pre-filled

roland4432:

Altered the way how pages (here: key, navigation) are created:
Key and navigation follow the entered title

Why? Since key is representing the navigation-url of the node in the backend tree those 2 values should match by default (but can still be altered) and the navigation entry will by default be a valid url with replacing invalid chars

Even if you think the replacement won't be sufficient, this might serve as a starting point to enhance it further

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
@@ -23,10 +23,6 @@ Ext.define('documentreemodel', {
     }]
 });
 
-const a = 'àáâäæãåāăąçćčđďèéêëēėęěğǵḧîïíīįìłḿñńǹňôöòóœøōõőṕŕřßśšşșťțûüùúūǘůűųẃẍÿýžźż·/_,:;';
-const b = 'aaaaaaaaaacccddeeeeeeeegghiiiiiilmnnnnoooooooooprrsssssttuuuuuuuuuwxyyzzz------';
-const p = new RegExp(a.split('').join('|'), 'g');
-
 pimcore.registerNS("pimcore.document.tree");
 pimcore.document.tree = Class.create({
 
@@ -1341,7 +1337,8 @@ pimcore.document.tree = Class.create({
                             }.bind(this), 100);
                         },
                         keyup: function (el) {
-                            this.setKeyAndNavigation(el.getValue(), pageForm);
+                            pageForm.getComponent("name").setValue(el.getValue());
+                            pageForm.getComponent("key").setValue(el.getValue());
                         }.bind(this),
                     }
                 },{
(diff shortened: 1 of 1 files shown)


brusch:
@roland4432 yep, agree! It makes sense to move the title to the top of the form and auto-fill the key and the navigation name based on the title. I disagree with the replacement functionality, because of the following reasons: 
- we're supporting UTF-8 slugs and all major search engines are totally fine with it (yes we're in 2020, even if some SEOs don't believe it 😉)
- we're therefore supporting also non-Latin based alphabets

Would be nice if you can update your PR accordingly. 

Btw. if you like to force your own slug/key policies you can use the following event: `\Pimcore\Event\SystemEvents::SERVICE_PRE_GET_VALID_KEY`

roland4432:
okay.. I can cut that from the code.. also keep capital letters and spaces?

brusch:
Yes, please!

roland4432:
Reverted the slug replacement and kept only the event to adopt the title to key and navigation, as well as the reordered textfields (title -> navigation -> key)

bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
```diff
@@ -1339,20 +1338,21 @@ pimcore.document.tree = Class.create({
...
+                            pageForm.getComponent("key").setValue(el.getValue());
+                        }.bind(this),
                     }
                 },{
                     xtype: "textfield",
                     itemId: "name",
                     fieldLabel: t('navigation'),
                     name: 'name',
-                    width: "100%"
+                    width: "100%",
```
brusch: Please remove the trailing `,` here, that's unfortunately not valid JS. 
It works in Chrome and FF, but anyway, we should have everything clean. ;)

bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
```diff
@@ -1339,20 +1338,21 @@ pimcore.document.tree = Class.create({
...
                 },{
                     xtype: "textfield",
-                    itemId: "title",
-                    fieldLabel: t('title'),
-                    name: 'title',
-                    width: "100%"
+                    width: "100%",
+                    fieldLabel: t('key'),
+                    itemId: "key",
+                    name: 'key',
```
brusch: Here as well please 😉

bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
```diff
@@ -1339,20 +1338,21 @@ pimcore.document.tree = Class.create({
                         },
                         keyup: function (el) {
                             pageForm.getComponent("name").setValue(el.getValue());
-                        }
+                            pageForm.getComponent("key").setValue(el.getValue());
+                        }.bind(this),
```
brusch: `,` ;-)

roland4432: meh.. my linter added them :D

bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
aarongerig: @brusch Just FYI, that is valid JavaScript (ES5) and also reduces git diff noise. Airbnb for example has it in their guidelines by default: https://github.com/airbnb/javascript#commas ;)

roland4432: that's my linter config (airbnb)... but it's up to @brusch to define his / their code styles so I apply to his rules :)

brusch:
Well, it's not about the style, the guys who still use IE11 will complain 😂

aarongerig:
Should work in IE11 as well, since that is valid ES5. :)

brusch:
Oh, yes you are right, for object literals it works, but not for function calls 😊

aarongerig:
Would a linter with a custom ruleset be of interest for Pimcore‘s JavaScript (ExtJS) base?

roland4432:
that might boost the next commit count for the dozens of thousands of EXTJS code lines :D

PR #5873, opened 2020-02-20
[Data Object] Unique and Index option should be only available on data-types th…

### LINKED ISSUE #5861: [Data Object] Unique option should be only available on data-types that support it
Currently this option is available across nearly all data-types even though they don't support it. 
See also #5850 

This option should be limited to data-types that support it, currently: input, numeric, user

weisswurstkanone:
resolves #5861

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/data.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/data.js
@@ -53,8 +53,8 @@ pimcore.object.classes.data.data = Class.create({
         }
 
         // per default all settings are available
-        this.availableSettingsFields = ["name","title","tooltip","mandatory","noteditable","index", "unique", "invisible",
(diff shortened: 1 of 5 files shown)


weisswurstkanone:
resolves #5861

brusch:
Looks good, should we implement the same for the index? Would make sense in my opinion. I think we can add this to this PR.

weisswurstkanone:
Good question. This would affect everything that implements ResourcePersistenceAwareInterface and is not relational.

brusch:
@weisswurstkanone well, many to one would work in theory 😊
But I think it would make sense, just to avoid misunderstandings, ... you don't think this is worth the effort?

weisswurstkanone:
Currently this is "enabled" by default. Changing this in the same manner would maybe break custom data types.

brusch:
@weisswurstkanone ok, then let's leave it as it is for now.

PR #5876, opened 2020-02-20
[Ecommerce]IndexService] ElasticSearch Worker only filter null values

### LINKED ISSUE #5874: [Ecommerce]IndexService]ElasticSearch Worker filters by boolean conversion
# Bug Report

The `AbstractElasticSearch` Index-Worker performs an `array_filter()` which results in filtering by default php boolean conversion `false`. This means that not only `null` values are filtered but also `0`, `false`, ... 

This may have a huge impact in the business logic if I filter by a field of type `integer` and if it does not exist in the document its sort value is infinit, but may actually be `0`!

A typical use case could be product-list order by sales-count.

Line: https://github.com/pimcore/pimcore/blob/be2a5932605df2832957cf0a1b740ea1dfc21627/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L425 

If we fix this issue, the interpreters and getters in the indexservice which are provided by Pimcore should be checked for the correct return values --> currently `false` maybe returned which actually should be `null`.

Also see https://www.php.net/manual/de/language.types.boolean.php#language.types.boolean.casting

jremmurd:
# Bugfix

Resolves #5874 

May be a bc-breaking, if there are interpreters getters in the application which do not return the correct type: a string or integer field may receive "false" - if `false` was returned which should actually be `null`. 

This may result in a filter like so:
image


brusch:
Thanks for your PR! Yes, this is kind of changing the behavior, so we schedule this patch for the next minor release.

PR #5882, opened 2020-02-21
[Element] unify getQuery() method on listing DAOs

brusch:


weisswurstkanone:
please have a look at the tests

models/DataObject/Listing/Dao.php
```diff
@@ -173,7 +173,7 @@ public function getCount()
      */
     public function loadIdList()
     {
-        $query = $this->getQuery();
+        $query = $this->getQuery([new Expression(sprintf('%s as o_id', $this->getTableName() . '.o_id')), 'o_type']);
```
weisswurstkanone: This looks odd, array or string ?

image

models/DataObject/Listing/Concrete/Dao.php
```diff
@@ -51,25 +51,20 @@ class Dao extends Model\DataObject\Listing\Dao
...
         $select = $this->db->select();
 
-        // create base
-        $field = $this->getTableName() . '.o_id';
         $select->from(
             [$this->getTableName()],
-            [
-                new Expression(sprintf('%s as o_id', $this->getSelectPart($field, $field))), 'o_type'
-            ]
+            [$columns]
```
weisswurstkanone: what if columns is already an array ?

models/DataObject/Listing/Dao.php
brusch: @weisswurstkanone well, that comes from here: https://github.com/pimcore/pimcore/blob/934c42b6cb2085cbd96889f542e4b888c4e02689/lib/Db/ZendCompatibility/QueryBuilder.php#L356

PR #5884, opened 2020-02-21
Check if redirect_to_maindomain is set to true

solverat:
This PR checks if `redirect_to_maindomain` is a true boolean. Otherwise, every instance will be redirected to `domain` => always!


brusch:
Thanks a lot!

PR #5886, opened 2020-02-21
[Asset] WebP thumbnails with media-queries are not working for cropped images

brusch:


andreas-gruenwald:
There is still a problem with the current solution: in desktop mode the wrong image is delivered, while in the mobile version this seems to work now.

I used this as a temporary workaround, may you please take a look?
```php
                    if ($isAutoFormat && Tool\DeviceDetector::getInstance()->isPhone()) {
                        $thumbConfigWebP = clone $thumbConfigRes;
                        $thumbConfigWebP->setFormat('webp');
                        $thumb = $image->getThumbnail($thumbConfigWebP, true);
                    }
```

PR #5887, opened 2020-02-21
Add ability to lock columns in the grid and save this in grid config

alexbaat:
ExtJs had a nice functionality for the grid to lock columns so it makes it easier to handle when you have a lot of columns in your grid. I have added this functionalty and made it a setting 'Enable grid locking' within the class setting.
In the grid you can lock/unlock columns. It is also saved within the grid configuration to make it persistent.


bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -501,6 +501,7 @@ public function doGetGridColumnConfig(Request $request, Config $config, $isDelet
                             'key' => $key,
                             'type' => 'system',
                             'label' => $key,
+                            'locked' => $sc['locked'],
```
dvesh3: please set lock property for saved calculatedcolumns(operators etc.) here:
https://github.com/pimcore/pimcore/blob/035465f49002e1278d134f0a6ab51b7414a6c4bc/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php#L797

and also set for container types(bricks, classificationstore etc.) somewhere here:
https://github.com/pimcore/pimcore/blob/035465f49002e1278d134f0a6ab51b7414a6c4bc/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php#L510

models/DataObject/ClassDefinition.php
```diff
@@ -179,6 +179,11 @@ class ClassDefinition extends Model\AbstractModel
         ]
     ];
 
+    /**
+     * @var bool
+     */
+    public $enableGridLocking = true;
```
dvesh3: I would keep this default `false` just like other options :)

dvesh3 (changes requested):

dvesh3:
Hey @alexbaat, nice work 👍 I like this feature and also would be nice to see this enabled for assets listing grid by default. what do you think?
image

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
alexbaat: I have applied this now also to the other fields like classificationstore, objectbricks and any operator.

models/DataObject/ClassDefinition.php
alexbaat: I've changed it to false.

alexbaat:
> Hey @alexbaat, nice work  I like this feature and also would be nice to see this enabled for assets listing grid by default. what do you think?
> image

I can change this if you like. Because this grid has no class settings, should it always be enabled?
I've notices some files, for instance bundles/AdminBundle/Resources/public/js/pimcore/asset/helpers/gridConfigDialog.js have different type of line endings (CRLF)  then the file bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridConfigDialog.js (LF).
What is the correct one to use?

dvesh3:
@alexbaat yes, for assets grid it should be enabled always. 

could you please fix line endings to Unix (LF)? thanks a lot!

alexbaat:
@dvesh3 I've added the gridlocking to the assets as well now. I've noticed there was a bug because in gridpanel in assets the autoscroll function was set to true and this cannot work with grid column locking. All works fine now.
I have not changed the line endings because I've noticed there are multiple files with CRLF setting and when I change it now the complete file is modified and you cannot see the real changes that I've made.

dvesh3:
@alexbaat when we lock preview column in assets list grid, it resets the height of other columns. can you please have a look? thanks!
image

alexbaat:
@dvesh3 Unfortunately changing forceFit to false doesn't help. I'll found out that the grid has a method called syncRowHeights(). This method will sync the heights between both parts of the grid.
I've tried to call this in the 'afterlayout' event but it doesn't work as expected. This is because the height of rows is changing when images are rendered within each row. So when I call the method to early, only the first rows have a synced height. But I cannot wait until all images are loaded because I don't know when this is.
Do you have any suggestions?

dvesh3:
@alexbaat Sorry for my late reply! I think calling syncRowHeights is quite challenging in this case like you mentioned. An easy and viable solution could be to wrap the image in fix height container(just like preview tab). wdyt?

dvesh3:
Hey, @alexbaat did you look at my last comment?

alexbaat:
Hi @dvesh3 , I've just made a new commit. I have used your suggestion by using the way it is used on the preview tab. Now the images are threaded like background images. I have to give it a fixed height to get it properly working. I think it works fine this way. Please have a look.

dvesh3 (approved):

weisswurstkanone:
@alexbaat could you resolve the confict please?

weisswurstkanone:
no worries, already done :-)

weisswurstkanone:
noticed one thing regarding the object grid.

1) in class definition, enable grid locking
2) in object grid,lock some columns and save the grid configuration
3) reload, columns are locked
4) in class definition, disable grid locking
5) reload object grid, columns are still locked

But I guess that's a minor issue :-)

alexbaat:
Hi @weisswurstkanone , I will fix it. Do you need it soon or can it be in a next release?
Do you suggest skipping the lock column functionality in Extjs when this is disabled in class or would it be better to update grid configs in database? I think first solution is easiest.
When I delete fields afterwards in the class, I see it also messes up the grid, because the fields doesn't fit the class anymore.

weisswurstkanone:
I would simply skip them. Updating the configs is not necessary I guess.
I think we are fine to put it in this release and do some minor improvements in the next one.
@alexbaat great improvement 👍

weisswurstkanone (approved):

BlackbitDevs:
Enabling grid locking causes the horizontal scrollbar only to appear at the bottom of the grid. If you do not have a touchpad or a mouse which supports scrolling even without scrollbar, you have to scroll to the bottom before being able to scroll horizontically.

PR #5894, opened 2020-02-24
Update Concrete.php

mc-supply:
Allow returning parent of calling class type or specified class type.  This seems to be the intended behavior and purpose of the function and the option to pass in $classId.  Current code only returns $parent if it is of the same class as calling class.

  
  

Resolves #


brusch:
I don't think this is not the right behavior now, because now it's either the same or the passed class `$classId`. 
Imho it should be just the one that get's passed as `$classId`.

brusch:
Put the change in #5904

Issue #5895, opened 2020-02-24
[Classification Store] Accessing undefined index to check dirty if new value is added


michel-paiva:
# Bug Report

When I add a new key to classification store and associate to an existing group that is associated to an object and I try to set a value to this key via setLocalizedKeyValue, if the object isn't dirty it will try to access an undefined index to make the check to mark _self as dirty.

```php
if (!$this->isFieldDirty('_self')) {
            if ($this->object) {
                $oldData = $this->items[$groupId][$keyId][$language];
                $oldData = $dataDefinition->getDataForResource($oldData, $this->object);
                $oldData = serialize($oldData);

                $newData = $dataDefinition->getDataForResource($value, $this->object);
                $newData = serialize($newData);

                if ($newData != $oldData) {
                    $this->markFieldDirty('_self');
                }
            } else {
                $this->markFieldDirty('_self');
            }
}
```

PR #5896, opened 2020-02-24
Dealing with undefined index if the field was empty before and the object is not dirty

michel-paiva:
PR solving the issue: #5895

### DIFF
--- a/models/DataObject/Classificationstore.php
+++ b/models/DataObject/Classificationstore.php
@@ -198,7 +198,7 @@ class Classificationstore extends Model\AbstractModel implements DirtyIndicatorI
 
         if (!$this->isFieldDirty('_self')) {
             if ($this->object) {
-                $oldData = $this->items[$groupId][$keyId][$language];
+                $oldData = $this->items[$groupId][$keyId][$language] ?? null;
                 $oldData = $dataDefinition->getDataForResource($oldData, $this->object);
                 $oldData = serialize($oldData);
 


models/DataObject/Classificationstore.php
```diff
@@ -198,14 +198,18 @@ public function setLocalizedKeyValue($groupId, $keyId, $value, $language = null)
 
         if (!$this->isFieldDirty('_self')) {
             if ($this->object) {
-                $oldData = $this->items[$groupId][$keyId][$language];
```
weisswurstkanone: Would say simply adding a trailing ?? null
would be sufficient.

models/DataObject/Classificationstore.php
```diff
@@ -198,14 +198,18 @@ public function setLocalizedKeyValue($groupId, $keyId, $value, $language = null)
...
-                $oldData = $dataDefinition->getDataForResource($oldData, $this->object);
-                $oldData = serialize($oldData);
+                $oldData = null;
+                
+                if (!isset($this->items[$groupId][$keyId][$language])) {
```
weisswurstkanone: remove the if

now we compare data which is transformed by getDataForResource with data that isn't

models/DataObject/Classificationstore.php
```diff
@@ -198,14 +198,18 @@ public function setLocalizedKeyValue($groupId, $keyId, $value, $language = null)
...
+                    $oldData = $this->items[$groupId][$keyId][$language];
+                    $oldData = $dataDefinition->getDataForResource($oldData, $this->object);
+                    $oldData = serialize($oldData);
+                }
 
                 $newData = $dataDefinition->getDataForResource($value, $this->object);
                 $newData = serialize($newData);
 
-                if ($newData != $oldData) {
+                if (!$oldData || $newData != $oldData) {
```
weisswurstkanone: shouldn't be necessary

models/DataObject/Classificationstore.php
michel-paiva: Done!

weisswurstkanone:
thx!

PR #5903, opened 2020-02-25
[Thumbnails] Add option to parallelize thumbnail pregeneration command

### LINKED ISSUE #5691: [Thumbnails] Add option to parallelize thumbnail pregeneration command
to speedup generation process

dvesh3:
Resolves #5691

### DIFF
--- a/bundles/CoreBundle/Command/ThumbnailsImageCommand.php
+++ b/bundles/CoreBundle/Command/ThumbnailsImageCommand.php
@@ -16,13 +16,23 @@ namespace Pimcore\Bundle\CoreBundle\Command;
 
 use Pimcore\Console\AbstractCommand;
 use Pimcore\Model\Asset;
-use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Webmozarts\Console\Parallelization\Parallelization;
 
 class ThumbnailsImageCommand extends AbstractCommand
 {
+    use Parallelization;
+
+    protected static $defaultName = 'pimcore:thumbnails:image';
+
+    /**
+     * @var Asset\Image\Thumbnail\Config[] $thumbnailsToGenerate
+     */
+    private $thumbnailConfigNames = [];
+
     protected function configure()
     {
         $this
@@ -65,13 +75,29 @@ class ThumbnailsImageCommand extends AbstractCommand
                 null,
                 InputOption::VALUE_NONE,
                 'do not generate high-res (@2x) versions of thumbnails'
+            )->addOption(
+                'processes',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'The number of parallel processes to run',
+                1
+            )->addOption(
+                'child',
+                null,
+                InputOption::VALUE_NONE,
+                'Set on child processes'
+            )->addArgument(
+                'item',
+                InputArgument::OPTIONAL,
+                'The item to process'
             );
     }
 
     /**
-     * @inheritDoc
+     * @inheritdoc
+     *
      */
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function fetchItems(InputInterface $input): array
     {
         $list = new Asset\Image\Thumbnail\Config\Listing();
         $thumbnailConfigList = $list->getThumbnails();
@@ -128,12 +154,6 @@ class ThumbnailsImageCommand extends AbstractCommand
             $thumbnailsToGenerate[] = Asset\Image\Thumbnail\Config::getPreviewConfig();
         }
 
-        $thumbnailConfigNames = [];
-        foreach ($thumbnailsToGenerate as $thumbnailConfig) {
-            $thumbnailConfigNames[] = $thumbnailConfig->getName();
-        }
-        $thumbnailConfigNames = array_unique($thumbnailConfigNames);
-
         // get only images
         $conditions = ["type = 'image'"];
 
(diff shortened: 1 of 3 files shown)


bundles/CoreBundle/Command/ThumbnailsImageCommand.php
```diff
@@ -149,54 +169,82 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+    protected function getItemName(int $count): string
+    {
+        return 1 === $count ? 'thumbnail' : 'thumbnails';
+    }
 
-        return 0;
+    /**
+     * @return \Symfony\Component\DependencyInjection\ContainerInterface
+     */
+    private function getContainer()
```
brusch: @dvesh3 for what purpose? Maybe a leftover? If we need it we should use DI and only inject the needed services.

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
```diff
@@ -65,13 +75,29 @@ protected function configure()
...
                 InputOption::VALUE_NONE,
                 'do not generate high-res (@2x) versions of thumbnails'
+            )->addOption(
+                'processes',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'The number of parallel processes to run',
+                1
+            )->addOption(
+                'child',
```
brusch: This options seems to be useless?

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
```diff
@@ -149,54 +169,82 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $list = new Asset\Listing();
         $list->setCondition(implode(' AND ', $conditions));
-        $total = $list->getTotalCount();
-        $perLoop = 10;
 
-        $totalToGenerate = $total * count($thumbnailsToGenerate);
+        $items = [];
```
brusch: I think we need to use `loadIdList()` here to avoid memory issues when having hundreds of thousands assets to process.

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
```diff
@@ -65,13 +75,29 @@ protected function configure()
...
+                1
+            )->addOption(
+                'child',
+                null,
+                InputOption::VALUE_NONE,
+                'Set on child processes'
+            )->addArgument(
+                'item',
+                InputArgument::OPTIONAL,
+                'The item to process'
```
brusch: We should make clear here that this is just used internally, nothing to care about 😉

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
dvesh3: used by Parallelization trait...changed visibility to protected

> Why is the progressbar gone :-(
> Should be still possible by using `runAfterBatch($input, $output, $items)`

Progressbar is still there (from Parallelization) w/o generated thumbnail info and i believe, it is quite hard to show generated thumbnail in `runAfterBatch` as it does not get the updated items

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
dvesh3: used by Parallelization trait. marked it "for internal use"

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
dvesh3: done!

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
dvesh3: done!

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
```diff
@@ -149,54 +169,88 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+    protected function getItemName(int $count): string
+    {
+        return 1 === $count ? 'thumbnail' : 'thumbnails';
+    }
 
-        return 0;
+    /**
+     * @return \Symfony\Component\DependencyInjection\ContainerInterface
+     */
+    protected function getContainer()
```
dpfaffenbauer: please use `ContainerAwareCommand` or `ContainerAwareInterface` with `ContainerAwareTrait`, that would be a bit more Symfony like.

dvesh3: done. thanks for review :)

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
```diff
@@ -149,54 +172,88 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
-        $output->writeln('');
+    /**
+     * @inheritdoc
+     *
+     */
+    protected function getItemName(int $count): string
+    {
+        return 1 === $count ? 'thumbnail' : 'thumbnails';
+    }
```
NiklasBr: Where is this used? Can't find it [in Symfony](https://github.com/symfony/symfony/search?q=ItemName&unscoped_q=ItemName) or [in Pimcore](https://github.com/pimcore/pimcore/search?q=ItemName&unscoped_q=ItemName), but it says that it is using `@inheritdoc`, inherited from where?

dvesh3: please see https://github.com/webmozarts/console-parallelization/blob/8f7440d1bdb5e350a081593395cc217705401ff0/src/Parallelization.php#L203

NiklasBr: Oh, thanks. Why not use Symfony?
`$translate->trans('1 thumbnail|%count% thumbnails', ['%count%' => $count])`?

brusch: Current approach would be fine for me as well, as we're not using translations anywhere in the commands anyway ... so kind of a nice to have 😉

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
brusch: 👍  thanks for clarifying & sorry! 😊

bundles/CoreBundle/Command/ThumbnailsImageCommand.php
NiklasBr: @brusch I think you misunderstand what the Symfony method is doing, it is not modifying language, it is modifying pluralisation. Why re-invent the wheel when we have access to a simple proper well functioning wheel?

PR #5904, opened 2020-02-25
[Data Object] Method getClosestParentOfClass($classId) should be independent on the calling class

brusch:
Follow up of #5894 
Related to #5658 

See also discussion here: https://github.com/pimcore/pimcore/pull/5658#issuecomment-590478950

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -570,7 +570,7 @@ class Concrete extends AbstractObject implements LazyLoadedFieldsInterface
     /**
      * @param string $classId
      *
-     * @return AbstractObject|null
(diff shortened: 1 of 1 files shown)


models/DataObject/Concrete.php
```diff
@@ -574,15 +574,15 @@ public function getNextParentForInheritance()
      */
```
weisswurstkanone: AbstractObject or Concrete ?

brusch: It was `AbstractObject` before as well. 
But yes, in my opinion  `AbstractObject` is right, because we if the direct parent of the called object can be a folder, and that should definitely work, right?

PR #5905, opened 2020-02-25
[Core] Fixed route collisions caused by same path prefixes of assets and documents in combination with unpublished documents

### LINKED ISSUE #2704: Asset gets processed as Document
Hi!

**Documents**
Home >
/series/batman (Unpublished)

**Assets**
/series/batman/batman.jpg

When opening the page /series/batman as preview, the image

is not visible (the asset throws a 404 error).

The debug mode shows that the ElementListener https://github.com/pimcore/pimcore/blob/73b00d44cb072e04da29a4ab7bd1a4324765f917/pimcore/lib/Pimcore/Bundle/CoreBundle/EventListener/Frontend/ElementListener.php#L122 processes the asset as document. The document is unpublished so the thumb gets not created.

### LINKED ISSUE #5893: Route collision when unpublished document has same name as asset folder
# Bug Report

When you have an unpublished document named "/test" and an asset under "/test/123.jpg" you cannot access a thumbnail of this image.

### Expected behavior

Thumbnail image should be shown

### Actual behavior

Exception thrown

### Steps to reproduce
1. Create thumbnail definition "abc"
2. Create document named "test" under "/", ensure it is unpublished
3. Create asset folder "test" under "/"
4. Upload image 123.jpg to asset folder "/test"
5. Create this little script test.php
```php
$asset = \Pimcore\Model\Asset::getByPath('/test/123.jpg');
echo $asset->getThumbnail('abc')->getPath());
```
6. Run script via `bin/console pimcore:run-script test.php`, should output something like `/test/image-thumb__410__abc/123.jpeg`
7. Open this URL via browser (and prepend domain name), e.g. https://demo.pimcore.fun/test/image-thumb__410__abc/123.jpeg

Now you get an exception thrown. 
If you publish the document "/test", it works and you get the thumbnail shown. 

PS: Beware that if you again unpublish the document and try to load the URL it still works. This is because then the generated thumbnail file exists and thus gets directly served without PHP being involved.

brusch:
Fixes #5893
Fixes #2704

Related #2826

### DIFF
--- a/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
@@ -21,7 +21,7 @@ use Pimcore\Http\Request\Resolver\SiteResolver;
 use Pimcore\Model\Document;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Symfony\Component\HttpKernel\Event\ControllerEvent;
+use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\OptionsResolver\OptionsResolver;
@@ -146,7 +146,7 @@ class DocumentFallbackListener implements EventSubscriberInterface
         }
     }
 
-    public function onKernelController(ControllerEvent $event) {
+    public function onKernelController(FilterControllerEvent $event) {
         // no document found yet - try to find the nearest document by request path
         // this is only done on the master request as a sub-request's pathInfo is _fragment when
         // rendered via actions helper
--- a/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
@@ -99,7 +99,7 @@ class DocumentFallbackListener implements EventSubscriberInterface
             // -> Symfony\Component\HttpKernel\EventListener\LocaleListener::onKernelRequest()
             // -> Pimcore\Bundle\CoreBundle\EventListener\Frontend\EditmodeListener::onKernelRequest()
             KernelEvents::REQUEST => ['onKernelRequest', 20],
-            KernelEvents::CONTROLLER => 'onKernelController',
+            KernelEvents::CONTROLLER => ['onKernelController', 20],
         ];
     }
 
--- a/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
@@ -20,6 +20,7 @@ use Pimcore\Http\Request\Resolver\PimcoreContextResolver;
 use Pimcore\Http\Request\Resolver\SiteResolver;
 use Pimcore\Model\Document;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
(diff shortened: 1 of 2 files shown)


BlackbitNeueMedien:
Hmm, does not work for me at all. When I use these changes, thumbnails are not generated at all.

brusch:
Hmm, very interesting. Nothing changed there afaik since 5 so, that should work also for 5. 
Any error message? Or what does _thumbnails are not generated at all_ actually mean? 
Thanks for your feedback! 😊

BlackbitNeueMedien:
I get
`Type error: Argument 1 passed to Pimcore\Targeting\EventListener\DocumentTargetGroupListener::assignDocumentTargetGroups() must be an instance of Pimcore\Model\Document, null given, called in /var/www/clients/client1/web28/web/vendor/pimcore/pimcore/lib/Targeting/EventListener/DocumentTargetGroupListener.php on line 76`

BlackbitNeueMedien:
Also used changes from https://github.com/pimcore/pimcore/commit/739957f849b8804683f71eef9be5c21493d932f5#diff-10617a31aeefe2eeabeb49a469750386R76 - now it works.

Thank you very much, @brusch !

brusch:
@BlackbitNeueMedien just wanted to as you to apply this change but you were a bit quicker 😊
Thanks a lot for testing and the feedback!

BlackbitNeueMedien:
There seems to be a BC break: For a URL `/de/calendar-data` in the controller which extends `Pimcore\Controller\FrontendController`  `$this->document` is `null` although a document `/de` exists (under a site example.org - so the document structure is / -> example.org -> de)

BlackbitNeueMedien:
There seems to be a BC break: For a URL `/de/calendar-data` in the controller which extends `Pimcore\Controller\FrontendController`  `$this->document` is `null` although a document `/de` exists (under a site example.org - so the document structure is / -> example.org -> de)

brusch:
@BlackbitNeueMedien and https://example.org/de/calendar-data is matching a route? (Not a document) is the document available on the request? Should be available with the key 'contentDocument'

brusch:
@BlackbitNeueMedien I'm unfortunately not able to reproduce the problem you've reported. 

I've tested the following cases: 

Site root is in all cases `/site`.   

### Case 1: Static Route
URL: `https://site.example.com/Magazine/Trends/bla/blub/asdasdasdfg~test~route`
Document path: `/site/Magazine/Trends`
Static route: `@~test~route@`
Controller extends `\Pimcore\Controller\FrontendController`, in the action `$this->document` is set to `/site/Magazine/Trends`   

**Correct behavior**

### Case 2: Document Route
URL: `https://site.example.com/Magazine/Trends`
Document path: `/site/Magazine/Trends`
Controller extends `\Pimcore\Controller\FrontendController`, in the action `$this->document` is set to `/site/Magazine/Trends`   

**Correct behavior**

### Case 3: Symfony Route
URL: `https://site.example.com/Magazine/Trends/symfony-route`
Document path: `/site/Magazine/Trends`
Symfony route: `@Route("{path}/symfony-route", requirements={"path"=".*?"})`
Controller extends `\Pimcore\Controller\FrontendController`, in the action `$this->document` is set to `/site/Magazine/Trends`   

**Correct behavior**

brusch:
But found another BC issue which is fixed with 146f169eeeae288aa9cc3834112f795706eaef63, so would be good to apply those changes as well 😊

BlackbitNeueMedien:
Case 3 was the problem in our case. I will try to test again...

solverat:
@brusch. This is a bc break anyway. If someone needs the document on a staticroute via kernelRequest for example (Like our i18n bundle), it's gone.

solverat:
BTW: Wouldn't it have been better to simply add an boolean attribute flag (eg `fallback_document`) to the request? In your element listener you could then skip the document.

brusch:
@solverat don't know what you actually mean, but the approach changed anyway already since this PR, see https://github.com/pimcore/pimcore/blob/master/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php

dpfaffenbauer:
This is a BC break, it moves the resolving of the fallback document after `kernel.request`, if somebody relied on the document during `kernel.request`, you get null now.

Basically the same as @solverat already confirmed. I just came across this using CoreShop/ThemeBundle

PR #5906, opened 2020-02-25
Bugfix translation

ckogler-elements:
Resolves #


dvesh3:
@ckogler-elements I guess we need this fix here as well:
https://github.com/pimcore/pimcore/blob/9e550ead94e5b35a8d8a0d3291d531e5c7fa76e7/models/Translation/Admin.php#L45

brusch:
@ckogler-elements any feedback? Thanks a lot!

dvesh3:
@ckogler-elements we should be good by just replacing this line https://github.com/pimcore/pimcore/blob/850475739c91be3d7ad2e33f225fbbb352616ba4/models/Translation/Admin.php#L64
with `return parent::getByKeyLocalized($id, $create, $returnIdIfEmpty, $language);`

ckogler-elements:
Hi,
i can't add something to this pull request, because my fork got crashed i had to delete it. 
Please just add it directly without a pull request.
@dvesh3

dvesh3:
alright. thanks, @ckogler-elements

PR #5910, opened 2020-02-26
[System-Config]: fix email settings detection

ssiergl:
still one regression left from https://github.com/pimcore/pimcore/pull/5745

email settings detection was false positiv

How to reproduce:

Frontend still has notification that email settings are not completed although completed (and other way round). False positive detection


ssiergl:
@dvesh3 still one bug was left :p

dvesh3:
thanks a lot! 🙌

PR #5912, opened 2020-02-26
Exceptions did not trigger an error message box

MercuryKojo:
If you write a Asset Listener to throw an error on certain conditions the error message was not showing in the backend UI.

XMLHttpRequest error is only called if there was an error transmitting.
if the transmission was successfull "load" is called, even if the status code is 400 and up


brusch:
Thanks!

Karthi96:
Still, I'm not getting error box though I have added those lines. Can you help me. 

In helpers.js, I have added https://github.com/pimcore/pimcore/pull/5912/commits/b5f59d53162d881a77a484b294e189cd7f459a67 these lines

MercuryKojo:
@Karthi96 
instead of throwing an exception, you can return a Response with status 500
this solution probably needs more work

Karthi96:
I tried returning the response with 500, it also failed to throw an exception. I can see that all asset events are not throwing a popup in UI, why ?

PR #5914, opened 2020-02-26
[UI] Progress bars shouldn't be modal if possible

brusch:
Related to #5629

Progress windows are getting stacked: 
image


dvesh3:
@brusch can we add some info to the progress windows? as now you can perform multiple tasks so it will be useful for the user to differentiate between different tasks.
image

brusch:
@dvesh3 yeah, this was also my original idea, but I had 2 concerns: 
- We need to enlarge the bar, which then covers other parts of the UI
- At the moment we don't have all the information available

brusch:
Reverted the change for batch updates in the object grid, as it seems that this causes some issues (race condition, ...).

PR #5917, opened 2020-02-27
Cache definition in Classification Store Service

kubaplas:
During the save process, getFieldDefinitionFromJson is called multiple times. In vanilla Pimcore at least twice for each definition - during save and when pushing data to backend search. When AOS plugin is installed - that's another one.

This change caches the result of getFieldDefinitionFromJson in static array. As the gain during regular save is minimal, the real benefit can be seen when saving a lot of objects in one go.

Here's the comparision for importing data from file (~100 objects with classification store). Before:
https://blackfire.io/profiles/43a7d2bf-0adc-45e9-9bf0-530a8c3e564a/graph
After:
https://blackfire.io/profiles/0be9a7e9-162a-40fa-90f7-247605a5895f/graph

### DIFF
--- a/models/DataObject/Classificationstore/Service.php
+++ b/models/DataObject/Classificationstore/Service.php
@@ -22,17 +22,34 @@ use Pimcore\Model\DataObject;
 class Service
 {
     /**
+     * @var array Used for storing definitions
+     */
+    protected static $definitionsCache = [];
+
+    /**
      * @param KeyConfig $keyConfig
      *
      * @return DataObject\ClassDefinition\Data
      */
     public static function getFieldDefinitionFromKeyConfig($keyConfig)
     {
+        if ($keyConfig instanceof KeyConfig) {
+            $cacheId = $keyConfig->getId();
+        }
+
+        if ($keyConfig instanceof KeyGroupRelation) {
+            $cacheId = $keyConfig->getGroupId() . "_" . $keyConfig->getKeyId();
+        }
+
+        if (array_key_exists($keyConfigId, self::$definitionsCache)) {
+            return self::$definitionsCache[$cacheId];
+        }
+
         $definition = $keyConfig->getDefinition();
         $definition = json_decode($definition, true);
         $type = $keyConfig->getType();
         $fd = self::getFieldDefinitionFromJson($definition, $type);
-
+        self::$definitionsCache[$cacheId] = $fd;
         return $fd;
     }
 
--- a/models/DataObject/Classificationstore/Service.php
+++ b/models/DataObject/Classificationstore/Service.php
@@ -35,10 +35,10 @@ class Service
     {
         if ($keyConfig instanceof KeyConfig) {
             $cacheId = $keyConfig->getId();
-        }
-
-        if ($keyConfig instanceof KeyGroupRelation) {
+        } else if ($keyConfig instanceof KeyGroupRelation) {
             $cacheId = $keyConfig->getKeyId();
+        } else {
+            throw new \Exception('$keyConfig should be KeyConfig or KeyGroupRelation');
         }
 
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
@kubaplas , makes sense.

Please have a look at the tests.

kubaplas:
@weisswurstkanone I had one missing rename, now it should be fine.

weisswurstkanone:
image

models/DataObject/Classificationstore/Service.php
```diff
@@ -21,18 +21,35 @@
 
 class Service
 {
+    /**
+     * @var array Used for storing definitions
+     */
+    protected static $definitionsCache = [];
+
     /**
      * @param KeyConfig $keyConfig
```
weisswurstkanone: KeyConfig|KeyGroupRelation

models/DataObject/Classificationstore/Service.php
```diff
@@ -21,18 +21,35 @@
...
      * @return DataObject\ClassDefinition\Data
      */
     public static function getFieldDefinitionFromKeyConfig($keyConfig)
     {
+        if ($keyConfig instanceof KeyConfig) {
+            $cacheId = $keyConfig->getId();
+        }
+
+        if ($keyConfig instanceof KeyGroupRelation) {
+            $cacheId = $keyConfig->getGroupId() . "_" . $keyConfig->getKeyId();
```
weisswurstkanone: even if it is a relation, the key and therefore the definition is still the same, right ?

models/DataObject/Classificationstore/Service.php
```diff
@@ -22,17 +22,34 @@
...
+        }
```
blankse: If $keyConfig is no KeyConfig or KeyGroupRelation, $cacheId is not defined.

```suggestion
        if ($keyConfig instanceof KeyConfig) {
            $cacheId = $keyConfig->getId();
        } elseif ($keyConfig instanceof KeyGroupRelation) {
            $cacheId = $keyConfig->getGroupId() . "_" . $keyConfig->getKeyId();
        } else {
            throw new \Exception('$keyConfig should be KeyConfig or KeyGroupRelation');
        }
```

models/DataObject/Classificationstore/Service.php
kubaplas: I think you're right - the definition is taken from KeyConfig table:
pimcore/models/DataObject/Classificationstore/KeyGroupRelation/Dao.php:43

models/DataObject/Classificationstore/Service.php
kubaplas: As Pimcore now requires PHP 7.2, maybe we can just typehint this?

models/DataObject/Classificationstore/Service.php
kubaplas: I've made the fix.

models/DataObject/Classificationstore/Service.php
blankse: PHP can only one class as typehint, right?

kubaplas: @weisswurstkanone is typehinting okay, or should we go with if/else?

weisswurstkanone:
I think the suggestion of @blankse looks fine.

models/DataObject/Classificationstore/Service.php
kubaplas: @blankse you're right, thanks for the tip!

weisswurstkanone:
Just wondering if there are scenarios where we want to wipe the cache. Long running script changing definitions or so.
Can we provide a method for that?

kubaplas:
@weisswurstkanone sure, it's there now

weisswurstkanone:
Thank you very much!

PR #5918, opened 2020-02-27
[DataObjects] Multiple assignments check for Relations Types bugfix

### LINKED ISSUE #5902: [DataObjects] Refactor multiple assignments check for Advanced relation types breaks Pimcore
https://github.com/pimcore/pimcore/pull/5852#issuecomment-590774419

Local fix: 
https://github.com/pimcore/pimcore/blob/b22960db68b4f30b468c769071f0128163fba9ba/models/Element/Service.php#L543

Returning `uniqeid` to sort of prove what is going wrong

> Besides that, the policy should be that if you find bugs then contribute a test covering the issue.
This would also help others.
> Thanks a lot.

I have no intention of creating a PR, that is not my responsibility. I already spent loads of hours in creating loads of plugins for FREE for the community and also supporting the community in many other ways. Introducing another bug in another release should be and is your responsibility to fix.

Just appreciate the fact that I already spent couple of hours investigating what is going on and not demanding PR's from the reporter!

Also, changing behaviour and not introducing new tests, is irresponsible for a product this size. And as you see, it breaks things not having them. Like with the release of 6.4 and now 6.5 as well.

dvesh3:
Resolves #5902

## Additional info  
Multiple assignments check for Relation Types moved from setter `preGetData` to `save` method so that it is still possible to assign relations without Id before persisting data.

### DIFF
--- a/models/DataObject/ClassDefinition/Data.php
+++ b/models/DataObject/ClassDefinition/Data.php
@@ -1208,7 +1208,7 @@ abstract class Data
      */
     protected function setDataToObject($data, $object, $params = [])
     {
-        $this->getSetDataObjectParam($object,  $params, $data, 'set');
+        $this->getSetData($object,  $params, $data, 'set');
     }
 
     /**
@@ -1221,7 +1221,7 @@ abstract class Data
      */
     protected function getDataFromObjectParam($object, $params = [])
     {
-        return $this->getSetDataObjectParam($object, $params);
+        return $this->getSetData($object, $params);
     }
 
     /**
@@ -1234,7 +1234,7 @@ abstract class Data
      *
      * @throws \Exception
      */
-    protected function getSetDataObjectParam($object,  $params = [], $setData = null, $method = 'get')
+    protected function getSetData($object,  $params = [], $setData = null, $method = 'get')
     {
         $data = null;
 
--- a/models/DataObject/ClassDefinition/Data.php
+++ b/models/DataObject/ClassDefinition/Data.php
@@ -1201,6 +1201,18 @@ abstract class Data
 
     /**
      * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|DataObject\Fieldcollection\Data\AbstractData $object
+     * @param mixed $data
+     * @param array $params
+     *
+     * @throws \Exception
+     */
+    protected function setDataToObject($data, $object, $params = [])
+    {
+        $this->getSetDataObjectParam($object,  $params, $data, 'set');
+    }
+
+    /**
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|DataObject\Fieldcollection\Data\AbstractData $object
      * @param array $params
      *
      * @return mixed
@@ -1208,6 +1220,21 @@ abstract class Data
      * @throws \Exception
      */
     protected function getDataFromObjectParam($object, $params = [])
+    {
+        return $this->getSetDataObjectParam($object, $params);
+    }
+
+    /**
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|DataObject\Fieldcollection\Data\AbstractData $object
+     * @param array $params
+     * @param mixed|null $setData
+     * @param string|null method
+     *
+     * @return mixed
+     *
+     * @throws \Exception
+     */
+    protected function getSetDataObjectParam($object,  $params = [], $setData = null, $method = 'get')
     {
         $data = null;
 
@@ -1226,7 +1253,7 @@ abstract class Data
                     if ($object instanceof DataObject\Concrete) {
                         $containerGetter = 'get' . ucfirst($fieldname);
                         $container = $object->$containerGetter();
-                        if (!$container && $context['containerType'] === 'block') {
+                        if ($method === 'get' && !$container && $context['containerType'] === 'block') {
                             // no data, so check if inheritance is enabled + there is parent value
                             if ($object->getClass()->getAllowInherit()) {
                                 try {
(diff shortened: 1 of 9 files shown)


models/DataObject/ClassDefinition/Data.php
```diff
@@ -1208,6 +1220,21 @@ public function getUnique()
...
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|DataObject\Fieldcollection\Data\AbstractData $object
+     * @param array $params
+     * @param mixed|null $setData
+     * @param string|null method
+     *
+     * @return mixed
+     *
+     * @throws \Exception
+     */
+    protected function getSetData($object,  $params = [], $setData = null, $method = 'get')
```
brusch: Should we maybe split this into 2 dedicated methods for better readability? 
For me it's seems to be quite hard to read the code.

dpfaffenbauer: I second that, that is super hard to read and also to understand....

dvesh3: I was just trying to save some duplication but yes, it leads to poor readability... so changed it :)

models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
```diff
@@ -602,6 +602,10 @@ public function save($object, $params = [])
         }
 
         $objectsMetadata = $this->getDataFromObjectParam($object, $params);
+        //TODO: move validation to checkValidity & throw exception in Pimcore 7
+        if ($object->isFieldDirty($this->getName()) || $object->isFieldDirty('_self')) {
```
weisswurstkanone: what if the trait isn't used?

models/DataObject/ClassDefinition/Data/ManyToManyRelation.php
```diff
@@ -699,6 +699,38 @@ public function getFromWebserviceImport($value, $relatedObject = null, $params =
...
     }
 
+    /**
+     * TODO: move validation to checkValidity & throw exception in Pimcore 7
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|\Pimcore\Model\DataObject\Fieldcollection\Data\AbstractData $object
+     * @param array $params
+     */
+    public function save($object, $params = [])
+    {
+        if (!DataObject\AbstractObject::isDirtyDetectionDisabled() && $object instanceof DataObject\DirtyIndicatorInterface) {
```
brusch: As this code is 100% identical with the one in `ManyToManyObjectRelation` they should share it. 
What about a trait `Pimcore\Model\DataObject\ClassDefinition\Data\Relations\ManyToManyRelationTrait` that has this `save()` method?

dvesh3: done!

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -415,4 +415,71 @@ public function loadLazyBrickField(DataObject\Objectbrick\Data\AbstractData $ite
...
+     *
+     * @param array|null $data
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|\Pimcore\Model\DataObject\Fieldcollection\Data\AbstractData $object
+     * @param array $params
+     *
+     * @return array
+     *
+     * @throws \Exception
+     */
+    public function filterMultipleAssignments($data, $object, $params)
```
weisswurstkanone: Can we call it $container instead of $object from now on?

models/DataObject/ClassDefinition/Data.php
```diff
@@ -1199,6 +1199,104 @@ public function getUnique()
...
     }
 
+    /**
+     * @param mixed $data
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|DataObject\Fieldcollection\Data\AbstractData $object
+     * @param array $params
+     *
+     * @throws \Exception
+     */
+    protected function setDataToObject($data, $object, $params = [])
```
weisswurstkanone: jost one thought: maybe mark it as internal and move it to AbstractRelations as it is only used there.

models/DataObject/ClassDefinition/Data.php
```diff
@@ -1199,6 +1199,104 @@ public function getUnique()
         return false;
     }
 
+    /**
+     * @param mixed $data
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|DataObject\Fieldcollection\Data\AbstractData $object
+     * @param array $params
+     *
+     * @throws \Exception
```
weisswurstkanone: @throws Exception ... wondering in which case

dvesh3: If relations are filtered then it calls `setDataToObject` which tries to set the data back to object so could be any scenario like `setLocalizedValue` fails etc.

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -415,4 +415,71 @@ public function loadLazyBrickField(DataObject\Objectbrick\Data\AbstractData $ite
...
+                        @trigger_error(
+                            'Passing relations multiple times is deprecated since version 6.5.2 and will throw exception in 7.0.0, tried to assign ' . $elementHash
+                            . ' multiple times in field' . $fieldName . ' of object id: ' . $objectId,
+                            E_USER_DEPRECATED
+                        );
+                    }
+                }
+
+                if(count($relationItems) !== count($data)) {
+                    $this->setDataToObject($relationItems, $object, $params);
```
weisswurstkanone: maybe change to setDataToObject(array_values($relationItems), ...)
but ... I am not sure if we should change the array keys anyway.
calling the setter, save it and then calling the getter will probably not return the same thing

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -415,4 +415,71 @@ public function loadLazyBrickField(DataObject\Objectbrick\Data\AbstractData $ite
...
+                    $elementHash = null;
+                    if ($item instanceof DataObject\Data\ObjectMetadata || $item instanceof DataObject\Data\ElementMetadata) {
+                        if ($item->getElement() instanceof Element\ElementInterface) {
+                            $elementHash = Element\Service::getElementHash($item->getElement());
+                        }
+                    } elseif ($item instanceof Element\ElementInterface) {
+                        $elementHash = Element\Service::getElementHash($item);
+                    }
+
+                    if($elementHash === null) {
```
weisswurstkanone: please check if this is still possible. on save every target must have an ID anyway

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
dvesh3: changed it setDataToObject(array_values($relationItems). Anyway passing associative array for relation types is not a valid structure.

models/DataObject/ClassDefinition/Data.php
dvesh3: marked it internal

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
dvesh3: done!

models/DataObject/ClassDefinition/Data.php
```diff
@@ -1199,6 +1199,102 @@ public function getUnique()
...
+    /**
+     *  @internal
+     *
+     * @param mixed $data
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|DataObject\Fieldcollection\Data\AbstractData $container
+     * @param array $params
+     *
+     * @throws \Exception
+     */
+    protected function setDataToObject($data, $container, $params = [])
```
dpfaffenbauer: can you explain me why this complicated function has to be in the "sort-of" abstract parent class for all object data-types? Can't it be extracted to corresponding types?

Also, I don't quite understand where this comes from, shouldn't this only change the behaviour of validation and nothing else?

dvesh3: A scenario for introducing this method is that you set duplicate relations to object and then save it, which then filters the duplicate data during save process so then filtered relations data should be set back to object to avoid having different information between object and database. 

The reason it is in Data class as data could be anything not just relation types and also it is the set counterpart of function `getDataFromObjectParam` which is called before applying filters.

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
dvesh3: yes, it is possible as you can save targets without ID however not sure what could be the use case.

models/DataObject/ClassDefinition/Data.php
dpfaffenbauer: so basically it is responsible for setting the filtered items back to the object after/on save? (brick, localized-fields,...)

If that is the case, I think you should completely re-think this solution, it is way to un-generic and should be done within the Object Data Tag itself. And it doesn't belong where it currently sits.

It is only used within the `filterMultipleAssignments` currently, where it is the wrong place as well. Every Object Data Tag **should** filter/handle that themself. The `filter` function already is called within the `AdvancedRelation` and obviously `AbstractRelation`, why not just call a setter with the filtered data there?! Then you would get completely rid of that function. So, the `filter` function returns only the allowed data-set, the Object Data Tag then persists said data and set's it back to the object.

brusch: The thing is that the data-type can be used in many different contexts/containers, you'd have to duplicate this code again and again, that's why we thought it makes sense to make this generic. It's marked internally anyway, if we don't see any further use, we're going to kick it with v7 together with this filtering stuff.

dpfaffenbauer: Alright then. Keeping it as a temp solution until Pimcore 7 is fine with me.

PR #5920, opened 2020-02-27
Fix $systemConfig not initialized in Config::offsetExists()

### LINKED ISSUE #5919: (data loss) (regression) Pimcore deletes document versions on its own
# Bug Report

### Steps to reproduce

1. make sure Pimcore is configured to save document versions.
2. save a document.
3. open the document and the versions tab after five minutes.

### Expected behavior

Previous versions are still there.

### Actual behavior

After saving a document a version entry is visible in the versions tab. It's still there after closing and opening the same document again. But after five minutes at the latest all versions are lost. So maybe it's triggered by something the maintenance cronjob does (runs every five minutes)?

I updated two Pimcore instances from 6.4.0 to 6.5.2 and both have the same issue. It was never an issue with version 6.4.0.

I cannot test on http://demo-basic.pimcore.org/admin/ (as requested in the issue template) because it does not exist.

Nevay:
Fixes Pimcore\Config::$systemConfig being not initialized in ::offsetExists().

Resolves #5919

### DIFF
--- a/lib/Config.php
+++ b/lib/Config.php
@@ -54,7 +54,7 @@ class Config implements \ArrayAccess
      */
     public function offsetExists($offset)
     {
-        return isset(static::$systemConfig[$offset]);
+        return self::getSystemConfiguration($offset) !== null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


brusch:
Thanks!

Issue #5761, opened 2020-02-10
Avoid deadlocks caused by many concurrent insert/update operations


brusch:
Deadlocks are mainly caused by delete/inserts into the `dependencies` and  `object_*_relations` table, but maybe there are also other tables affected, such as `versions`, ...

brusch:
One possible solution would be using something like that on  `Pimcore\Db\PimcoreExtensionsTrait`
```
    public function deleteWhereById($table, $where = '', $idColumnName = 'id')
    {
        $sql = sprintf('SELECT `%s` FROM `%s`', $idColumnName, $table);
        if ($where) {
            $sql .= ' WHERE ' . $where;
        }

        $ids = $this->fetchCol($sql);
        if(!empty($ids)) {
            return $this->deleteWhere($table, sprintf("`%s` IN (%s)", $idColumnName, implode(",", $ids)));
        }

        return null;
    }
```

brusch:
I'm wondering why we're experiencing deadlocks there, because that shouldn't happen at all :( 
https://dev.mysql.com/doc/refman/5.7/en/innodb-locks-set.html 
https://dev.mysql.com/doc/refman/5.7/en/innodb-locking.html

weisswurstkanone:
Maybe off-topic, but should be notices. Managed to create the following situation after
- open object with relation data
- clear cache
- save it without touching anyhting

The hash is different which triggers:
- remove all existing dependencies
- re-insert the same ones

![image](https://user-images.githubusercontent.com/6142086/75039096-3948a800-54b8-11ea-9b9b-f6c094a2b6fa.png)

weisswurstkanone:
group_concat returns a different result because the default max length is 1024 which doesn't scale very wellfor that purpose.
see https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_group_concat_max_len

weisswurstkanone:
maybe replace this:
https://github.com/pimcore/pimcore/blob/ae677a1de09d1c488cce9cea3ba95a3e44954ae0/models/DataObject/AbstractObject/Dao.php#L123

- delete only needed if it is not locked anymore.
- otherwise insert or update

weisswurstkanone:
not directly related. do we need both or is a composite enough?
https://github.com/pimcore/pimcore/blob/17d4ce07869ffad2aa0c920ed75aa55a3e6ff44e/bundles/InstallBundle/Resources/install.sql#L106

[...]

PR #5921, opened 2020-02-27
Avoid deadlocks caused by many concurrent insert/update operations #5761

weisswurstkanone:


lib/Db/PimcoreExtensionsTrait.php
```diff
@@ -611,4 +611,25 @@ public function setAutoQuoteIdentifiers($autoQuoteIdentifiers)
...
+        $sql = 'SELECT ' . $idColumn . '  FROM ' . $table;
+
+        if ($where) {
+            $sql .= ' WHERE ' . $where;
+        }
+
+        $idsForDeletion = $this->fetchCol($sql);
+
+        if (!empty($idsForDeletion)) {
+            $idString = implode(",", $idsForDeletion);
```
brusch: We should quote the IDs in case they are alphanumeric. 
`$idString = implode(',', array_map([$this, 'quote'], $idsForDeletion))`

lib/Db/PimcoreExtensionsTrait.php
```diff
@@ -611,4 +611,25 @@ public function setAutoQuoteIdentifiers($autoQuoteIdentifiers)
...
+    {
+        $sql = 'SELECT ' . $idColumn . '  FROM ' . $table;
+
+        if ($where) {
+            $sql .= ' WHERE ' . $where;
+        }
+
+        $idsForDeletion = $this->fetchCol($sql);
+
+        if (!empty($idsForDeletion)) {
```
brusch: Maybe we should think about processing them in chunks of maybe 50 IDs, to not exceed the max. query length `max_allowed_packet`.

models/Dependency/Dao.php
```diff
@@ -72,66 +74,96 @@ public function cleanAllForElement($element)
                 }
             }
 
-            $this->db->delete('dependencies', ['sourceid' => $id, 'sourcetype' => $type]);
+            $this->db->selectAndDeleteWhere('dependencies', 'id', "`sourceid` = '" . $id . "' AND  `sourcetype` = '" . $type . "'");
```
brusch: We should use proper `quoteInto()` here, just in case, u know 😉

models/Dependency/Dao.php
```diff
@@ -72,66 +74,96 @@ public function cleanAllForElement($element)
...
     /**
      * Clear all relations in the database for current source id
+     *
+     * @return void
      */
     public function clear()
     {
         try {
-            $this->db->delete('dependencies', ['sourceid' => $this->model->getSourceId(), 'sourcetype' => $this->model->getSourceType()]);
+            $this->db->selectAndDeleteWhere('dependencies', 'id', "`sourceid` = '" . $this->model->getSourceId() . "' AND  `sourcetype` = '" . $this->model->getSourceType() . "'");
```
brusch: Here as well `quoteInto()`

lib/Db/PimcoreExtensionsTrait.php
weisswurstkanone: default value is 67108864 so I think 50 is a bit pesimistic

brusch: 😁 alright, maybe you're right. But splitting it into chunks makes sense, right?

lib/Db/PimcoreExtensionsTrait.php
```diff
@@ -611,4 +611,28 @@ public function setAutoQuoteIdentifiers($autoQuoteIdentifiers)
...
     }
+
+    /**
+     * @param string $table
+     * @param string $idColumn
+     * @param string $where
+     */
+    public function selectAndDeleteWhere($table, $idColumn = 'id', $where = '')
+    {
+        $sql = 'SELECT ' . $idColumn . '  FROM ' . $table;
```
ultramegatom: i think $idColumn should be quoted here?

brusch: 👍 good catch!  `quoteIdentifier()`
@weisswurstkanone can you please update the PR! Thanks!

weisswurstkanone: thx!

PR #5923, opened 2020-02-27
Update links to 5.x docs to point to 6.x docs

limenet:

I noticed some links in Pimcore `master` (6.x) were still pointing to the docs for version 5.x and as such you'd be looking at the wrong version of the docs. This PR fixes these links.

Please let me know if you'd like me to make any changes!


brusch:
Thanks a lot for your PR! 
Please undo the changes on the translation files as the master for them is POEditor and they will be overridden anyway. 
Thanks a lot!

aarongerig:
@brusch @limenet Wouldn't it make sense to use `https://pimcore.com/docs/latest/...`, which always redirects to the latest version of the documentation?

brusch:
@aarongerig well, depends, if you're for instance still running v5, then linking to `latest` is probably not  desired.

aarongerig:
@brusch True, but then this should be done on every major release. Now all versions 6.0.0 to 6.5.2 have links to the old documentation, which isn't optimal as well. ;)

brusch:
@aarongerig fully agree!

limenet:
@brusch you're welcome! I've reverted the changes to the files in `bundles/CoreBundle/Resources/translations/` and as you and @aarongerig discussed, the links still point to the 6.x docs (and not `latest`).

Is there anything else you'd like me to change?

brusch:
Thanks a lot!

limenet:
You're very welcome, @brusch !

PR #5933, opened 2020-02-28
refactor data object setOrderKey and setGroupBy

### LINKED ISSUE #5927: [DataObject] check of sql backticks are missing on setOrderKey
# Bug Report

```php
$myStockList->setOrderKey( '`object_25`.o_modificationDate');
```
runs into a error, because the check of sql backticks is missing.

albertmueller:
Resolves #5927

## Additional info
second parameter is actually no longer necessary and could be removed later (Attention if yes then it is a BC break)

### DIFF
--- a/lib/Model/Listing/AbstractListing.php
+++ b/lib/Model/Listing/AbstractListing.php
@@ -233,7 +233,11 @@ abstract class AbstractListing extends AbstractModel implements \Iterator
                 if ($quote === false) {
                     $this->orderKey[] = $o;
                 } elseif ($this->isValidOrderKey($o)) {
-                    $this->orderKey[] = '`' . $o . '`';
+                    if (strpos($o, '`') !== false) {
+                        $this->orderKey[] = $o;
+                    } else {
+                        $this->orderKey[] = '`' . $o . '`';
+                    }
                 }
             }
         }
--- a/lib/Model/Listing/AbstractListing.php
+++ b/lib/Model/Listing/AbstractListing.php
@@ -15,6 +15,7 @@
 namespace Pimcore\Model\Listing;
 
 use Pimcore\Db;
+use Pimcore\Db\ZendCompatibility\Expression;
 use Pimcore\Model\AbstractModel;
 
 /**
@@ -214,11 +215,11 @@ abstract class AbstractListing extends AbstractModel implements \Iterator
 
     /**
      * @param string|array $orderKey
-     * @param bool $quote
+     * @param bool $backticks
      *
      * @return $this
      */
-    public function setOrderKey($orderKey, $quote = true)
+    public function setOrderKey($orderKey, $backticks = true)
     {
         $this->setData(null);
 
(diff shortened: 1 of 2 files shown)


albertmueller:
@brusch my idea related to #5927

lib/Model/Listing/AbstractListing.php
```diff
@@ -214,11 +215,11 @@ public function getOrderKey()
 
     /**
      * @param string|array $orderKey
-     * @param bool $quote
+     * @param bool $backticks
      *
      * @return $this
      */
-    public function setOrderKey($orderKey, $quote = true)
+    public function setOrderKey($orderKey, $backticks = true)
```
dpfaffenbauer: why the rename?

lib/Model/Listing/AbstractListing.php
```diff
@@ -230,7 +231,7 @@ public function setOrderKey($orderKey, $quote = true)
 
         if (is_array($orderKey)) {
             foreach ($orderKey as $o) {
-                if ($quote === false) {
+                if (!$backticks || $this->checkIfValueContainBackticks($o)) {
                     $this->orderKey[] = $o;
                 } elseif ($this->isValidOrderKey($o)) {
                     $this->orderKey[] = '`' . $o . '`';
```
dpfaffenbauer: better use `$db->quote` here?

lib/Model/Listing/AbstractListing.php
```diff
@@ -241,6 +242,35 @@ public function setOrderKey($orderKey, $quote = true)
...
+     * @return $this
+     */
+    public function setGroupBy($groupBy, $backticks = true)
+    {
+        $this->setData(null);
+
+        if ($groupBy) {
+            $this->groupBy = $groupBy;
+
+            if (!$backticks || $this->checkIfValueContainBackticks($groupBy)) {
```
dpfaffenbauer: Actually, this is super inconsistent and should be unified. Guess a TODO for Pimcore 7 would be alright for now?!

dpfaffenbauer: forget that, you consistent that now :)

lib/Model/Listing/AbstractListing.php
albertmueller: because these are not quotes, you want to say that you're using backticks in your value

lib/Model/Listing/AbstractListing.php
albertmueller: make sense, i will change it

albertmueller: @dpfaffenbauer do you recommend to use $db = Db::get(); or implement the ConnectionInterface? i'm not sure if the ConnectionInterface via autowire is working and $db = Db::get(); is the ugliest solution :)

lib/Model/Listing/AbstractListing.php
albertmueller: recommend to remove the $backticks parameter in pimcore 7 with a BC break note

lib/Model/Listing/AbstractListing.php
dpfaffenbauer: only option here is `DB:get`.

lib/Model/Listing/AbstractListing.php
dpfaffenbauer: no, you are quoting the key, that is the point here.

albertmueller: as we discussed lets using "escape" now

lib/Model/Listing/AbstractListing.php
albertmueller: done

brusch:
Thanks for your PR! 
However I think this is is all in all not necessary and also just wrong in certain cases. 

What about cases like: 
```php
$list->setOrderKey('(SELECT foo FROM bar WHERE objects.o_id = foo.id)', false);
```

brusch:
I'll close this for now, as I don't see any benefit, rather the opposite is the case. 
But feel free to continue the discussion, we can open the PR again anytime 😉

Issue #5940, opened 2020-03-02
[Ecommerce][Indexservice] Productlist cache inconsistency


jremmurd:
# Bug Report

The product mockup cache and the elastic search index gets inconsistent if using the we implement the ` doPreIndexDataModification()`method. https://github.com/pimcore/pimcore/blob/master/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L448

1. Worker writes data to Store Table, does data modification and writes the result in the mockup cache and ElasticSearch.
2. Somehow the pimcore cache gets cleared.
3. Now the AbstractMockupCacheWorker does not consider the method `doPreIndexDataModification` and so the data added or modified in this method does NOT get considered

See: https://github.com/pimcore/pimcore/blob/master/bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractMockupCacheWorker.php#L72

PR #5941, opened 2020-03-02
[Ecommerce][Indexservice] Productlist cache inconsistency

jremmurd:
# Bugfix
Resolves https://github.com/pimcore/pimcore/issues/5940

### DIFF
--- a/bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractMockupCacheWorker.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractMockupCacheWorker.php
@@ -70,6 +70,10 @@ abstract class AbstractMockupCacheWorker extends AbstractBatchProcessingWorker
         if (empty($data)) {
             $data = $this->db->fetchOne('SELECT data FROM ' . $this->getStoreTableName() . ' WHERE o_id = ? AND tenant = ?', [$objectId, $this->name]);
             $data = json_decode($data, true);
+            $worker = $this->getTenantConfig()->getTenantWorker();
+            if(is_callable([$worker, "doPreIndexDataModifications"], true)) {
+                $data = $worker->doPreIndexDataModification($data);
+            }
         }
 
         if ($this->tenantConfig instanceof MockupConfigInterface) {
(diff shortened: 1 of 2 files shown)


jremmurd:
Additional thoughs: why not load data from ElasticSearch rather than from the store table?

brusch:
@jremmurd Thanks for your PR. 
Can you please fix the issue reported here: 
https://travis-ci.com/pimcore/pimcore/jobs/292920687#L1287

Thanks a lot!

andreas-gruenwald:
> Additional thoughs: why not load data from ElasticSearch rather than from the store table?

I already had the same thought. But probably it doesnt hurt that much...

bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractMockupCacheWorker.php
```diff
@@ -70,6 +70,10 @@ public function saveToMockupCache($objectId, $data = null)
         if (empty($data)) {
             $data = $this->db->fetchOne('SELECT data FROM ' . $this->getStoreTableName() . ' WHERE o_id = ? AND tenant = ?', [$objectId, $this->name]);
             $data = json_decode($data, true);
+            $worker = $this->getTenantConfig()->getTenantWorker();
+            if(method_exists($worker, "doPreIndexDataModification")) {
```
fashxp: Should we introduce an interface for that?

fashxp:
> Additional thoughs: why not load data from ElasticSearch rather than from the store table?

Isn't that code not used by other adapters like optimized mysql, findologic, etc too?
There would be no elastic to load the data from. Or am I missing something?

bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractMockupCacheWorker.php
jremmurd: I though about something like that as well, another option would be to extract line 71 and 72 in a new method `getDataForMockupCache` or somthing like that and override it in the AbstractElasticSearch Worker? 
In that way we could (now or later) tackle the other idea as well, and fetch the data directly from ES.

jremmurd:
> > Additional thoughs: why not load data from ElasticSearch rather than from the store table?
> 
> I already had the same thought. But probably it doesnt hurt that much...

It can hurt dependent on what one is doing in `doPreIndexDataModification` now. Currently ElasticSearch is not in use for the product list mockups as far as I can see - to know that kinda hurts in itself ;)

fashxp:
closed in favor of #6320

PR #5960, opened 2020-03-04
Document::hasElement() should return true for inherited elements

### LINKED ISSUE #5959: [Documents] Document::hasElement() returns false for inherited elements
# Bug Report

### Expected behavior
`Document::hasElement()` should return `true` for inherited elements

### Actual behavior
`Document::hasElement()` returns `false` for inherited elements

### Steps to reproduce
1. Create document with editable elements
2. Copy document from 1. and set content master document to 1.
3. In the controller of both documents `Document::hasElement()` returns `true` for the original document but `false` for the inherited document.

I do not know if this is really a bug as the copied document does not really **have** the element in the sense of being the owner of the element but it is an unexpected behaviour that `Document::hasElement()` returns `false` but you actually can get the element via `Document::getElement()`.

BlackbitNeueMedien:
Resolves #5959

### DIFF
--- a/models/Document/PageSnippet.php
+++ b/models/Document/PageSnippet.php
@@ -404,7 +404,7 @@ abstract class PageSnippet extends Model\Document
     public function getElement($name)
     {
         $elements = $this->getElements();
-        if ($this->hasElement($name)) {
+        if (isset($this->elements[$name]) ){
             return $elements[$name];
         }
 
@@ -504,9 +504,7 @@ abstract class PageSnippet extends Model\Document
      */
     public function hasElement($name)
     {
-        $elements = $this->getElements();
-
-        return array_key_exists($name, $elements);
+        return $this->getElement($name) !== null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


brusch:
Thanks for your PR. 
I actually don't see a problem with that change, but we should change `removeELement()` to `isset()` as well, see: https://github.com/pimcore/pimcore/blob/81ffb899f0cec27173f0226dec33b7c221de59d8/models/Document/PageSnippet.php#L390

Can you please update your PR accordingly? 
Thanks a lot!

BlackbitNeueMedien:
Fixed `removeElement()`.

brusch:
Grazie Mille! 👌🙏

jdreesen:
This change has broken the targeting which now results in an endless loop.

The problem is, that `PageSnippet::hasElement()` now uses `getElement()` which is overridden in `TargetingDocument` and itself uses `hasElement()`.

See: https://github.com/pimcore/pimcore/blob/2ec7fc523ad6ec0f5322863f5d2b524782cdbbce/models/Document/TargetingDocument.php#L124-L130

brusch:
@jdreesen thanks for reporting! 
To keep track of bugs it's required to create a dedicated issue or even better a PR with a fix for it so that we can assign it to a milestone. 
Thanks a lot! 😊

PR #5963, opened 2020-03-04
[Events] Symfony version agnostic EventDispatcher

BlackbitNeueMedien:
As I was annoyed by all the deprecation notices like `Calling the "EventDispatcher::dispatch()" method with the event name as the first argument is deprecated since Symfony 4.3, pass it as the second argument and provide the event object as the first argument instead.` which come from the Pimcore core, I created this decorator. The reason for the dispatch calls to have wrong order (in the sense of Symfony 4) is that in Pimcore 6.* we want to support Symfony 3 where the order of arguments of `EventDispatcher::dispatch()` was switched compared to now.
With this decorator we can switch to the new `dispatch()` syntax step by step without BC breaks.

A disadvantage is that developers would not get deprecation notices about application-specific events. But perhaps we find a way to also solve that until we can throw away this decorator when Pimcore will not support Symfony 3 anymore.

### DIFF
--- a/bundles/CoreBundle/Resources/config/services.yml
+++ b/bundles/CoreBundle/Resources/config/services.yml
@@ -202,3 +202,7 @@ services:
     Pimcore\Bundle\CoreBundle\Request\ParamConverter\DataObjectParamConverter:
         tags:
             - { name: request.param_converter, priority: -2, converter: data_object_converter }
+
+    Pimcore\Event\EventDispatcher:
+        decorates: event_dispatcher
+        arguments: ['@Pimcore\Event\EventDispatcher.inner']
--- a/lib/Event/EventDispatcher.php
+++ b/lib/Event/EventDispatcher.php
@@ -1,0 +1,101 @@
+<?php
+
+namespace Pimcore\Event;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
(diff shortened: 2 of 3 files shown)


dpfaffenbauer:
IMO: the disadvantage is disadvantage enough for me to not implement that. How am I supposed to know what to change in my application when you disable this notices?

BlackbitNeueMedien:
@dpfaffenbauer That is what I said: Perhaps we find a way to distinguish between events dispatched by the core and events dispatched by the application.

dpfaffenbauer:
Not sure if that even is possible properly. Only way to know is by checking for the path the event comes from, and that is still not a nice solution then, cause if requires dirty things like debug_backtrace.

I'd rather opt to make that an Application specific implementation than to merge that into the core...

BlackbitNeueMedien:
Only works with Symfony 4 ;-(
In Symfony 3 there is
`Declaration of Pimcore\Event\EventDispatcher::dispatch($event) must be compatible with Symfony\Component\EventDispatcher\EventDispatcherInterface::dispatch($eventName, ?Symfony\Component\EventDispatcher\Event $event = NULL)`
But I cannot change the method signature because then it would be incpmpatible with the Symfony 4 variant...

PR #5964, opened 2020-03-04
Output actual name of Product class on indexing.

andreas-gruenwald:
## Improvement 

 Output the actual (short) name of the class instead of "product". This is useful as other entities might get indexed as well, and it that case the message "Updating product ..." might get confusing.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/IndexService/Tool/IndexUpdater.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Tool/IndexUpdater.php
@@ -88,7 +88,7 @@ class IndexUpdater
             self::log($loggername, '=========================');
 
             foreach ($products as $p) {
-                self::log($loggername, 'Updating product ' . $p->getId());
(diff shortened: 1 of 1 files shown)


bundles/EcommerceFrameworkBundle/IndexService/Tool/IndexUpdater.php
```diff
@@ -88,7 +88,7 @@ public static function updateIndex($objectListClass, $condition = '', $updateInd
             self::log($loggername, '=========================');
 
             foreach ($products as $p) {
-                self::log($loggername, 'Updating product ' . $p->getId());
+                self::log($loggername, 'Updating '.array_pop(explode('\\', get_class($p))).' '.$p->getId());
```
brusch: `$p->getClass()->getName()` would be easier ;-) 
Can you please update the PR?

brusch: @andreas-gruenwald  Just made the update, so nothing else to do for you  😊

PR #5967, opened 2020-03-04
Advanced Many To Many Objects metadata column names get "strtolowered…

### LINKED ISSUE #5908: Advanced Many To Many Objects metadata column names  get "strtolowered"
# Bug Report

I think this isn't supposed to be a feature, but if it's the case please forgive me. 

### Expected behavior & actual behaviours

As an admin, when I design DataObjects and create an Advanced Many To Many Object relation property, I want my metadata column names to take into account the correct capitalization of the names I input. 

For example, when I type `productionCost`, I expect that the name displayed and saved in the database to be `productionCost`, and not `productioncost`. I also expect the Metadata object to have a magic getter `$metadata->getProductionCost()` instead of `$metadata->getProductioncost()`.

### Steps to reproduce

DataObject > Classes > Add an Advanced M2M relation and create a column with a custom capitalization. Save the Class and reload definition.

weisswurstkanone:
Resolves #5908

### DIFF
--- a/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
+++ b/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
@@ -845,7 +845,7 @@ class AdvancedManyToManyObjectRelation extends ManyToManyObjectRelation
         $this->columns = [];
         $this->columnKeys = [];
         foreach ($columns as $c) {
-            $c['key'] = strtolower($c['key']);
+            $c['key'] = $c['key'];
             $this->columns[] = $c;
             $this->columnKeys[] = $c['key'];
         }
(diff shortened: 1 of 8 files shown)


models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
```diff
@@ -845,7 +845,7 @@ public function setColumns($columns)
         $this->columns = [];
         $this->columnKeys = [];
         foreach ($columns as $c) {
-            $c['key'] = strtolower($c['key']);
+            $c['key'] = $c['key'];
```
dvesh3: this can go away

weisswurstkanone: done

PR #5974, opened 2020-03-05
[DataObjects/Documents] Clear children cache on child delete

### LINKED ISSUE #5909: Pimcore | AbstractObject | Behaviour of hasChildren()
# Bug Report
In Pimcore 5 the following command returned ``false``:
 
```php
    $child = AbstractObject::getById(123); //fetch a data object
    $parent = $child->getParent();  //navigate to its parent
    $child->delete(); //delete the last child
    $parent->hasChildren(); //no more children -> false
````

However, in Pimcore 6.4.2. ``hasChildren`` will still return ``true`` if the last child is deleted, as the data object is obviously cached.

```php
$parent->hasChildren(); //returns true
AbstractObject::getById($parent->getId(), true)->hasChildren(); //returns false
```

What is the expected behavior? Would it make sense to adapt the code so that a child deletion will also affect the parent data object?

dvesh3:
Resolves #5909

### DIFF
--- a/doc/Development_Documentation/05_Objects/03_Working_with_PHP_API.md
+++ b/doc/Development_Documentation/05_Objects/03_Working_with_PHP_API.md
@@ -456,25 +456,3 @@ $list->onCreateQuery(function (\Pimcore\Db\ZendCompatibility\QueryBuilder query)
     echo $query;
 });
 ```
-
-
-### Working with Runtime Cache
-Pimcore heavily uses runtime cache to cache API results for performance reasons. However, it is very crucial to understand that how to deal with cached results so that correct data should utilized from the API. Let's take few examples to understand similar situations:
-
-```php
-//Delete item from the list
-$list = new DataObject\Test\Listing;
-$list->load();
-$list->current()->delete();
-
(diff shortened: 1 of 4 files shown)


brusch:
I'm just thinking, ... maybe it's time to introduce a `getCacheKey()` on the AbstractElement? 
@weisswurstkanone @dvesh3 what do you think?

dvesh3:
@brusch we already have one or you mean something else?
https://github.com/pimcore/pimcore/blob/83ece3824eb4666d212e5d2dfdbb0f80d979a07f/models/Element/AbstractElement.php#L120-L125

brusch:
@dvesh3 oh, perfect :) Since the cache tag and the cache key is always the same we could use that one as well, yes. Then we don't have to introduce another method which is actually doing the same. 
So I'd go for this one! We shoudl also use it then in the `getById()` methods.

dvesh3:
@brusch It is a different topic...we shouldn't update in this PR cause then it needs an instance for `getCacheTag()` which is wrong before runtime cache check.

brusch:
@dvesh3 implemented an alternative approach here #5994

PR #5976, opened 2020-03-05
[Advanced relations] Lowercase metadata columns

BlackbitNeueMedien:
Metadata field names are converted to lower case in https://github.com/pimcore/pimcore/blob/a393da91d63e5819bcd5eb776780414aa8fdae99/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php#L838-L854

And also the getter / setter access converts field names to lowercase in https://github.com/pimcore/pimcore/blob/a393da91d63e5819bcd5eb776780414aa8fdae99/models/DataObject/Data/ObjectMetadata.php#L94-L115

But with the following script you get an error:
```php
$meta = new \Pimcore\Model\DataObject\Data\ObjectMetadata('test', ['camelCase']);
$meta->setCamelCase('abc'); // error "Requested data camelcase not available"
```

This PR fixes that.

### DIFF
--- a/models/DataObject/Data/ElementMetadata.php
+++ b/models/DataObject/Data/ElementMetadata.php
@@ -62,7 +62,7 @@ class ElementMetadata extends Model\AbstractModel implements DataObject\OwnerAwa
     public function __construct($fieldname, $columns = [], $element = null)
     {
         $this->fieldname = $fieldname;
-        $this->columns = $columns;
+        $this->columns = array_map('strtolower', $columns);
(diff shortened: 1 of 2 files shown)


weisswurstkanone:
There is already a PR for this.
See https://github.com/pimcore/pimcore/pull/5967

BlackbitNeueMedien:
Oh, my fault.

weisswurstkanone:
Thanks anyway :+1:

Issue #5813, opened 2020-02-14
Versioning performance with a lot of relationships


belendel:
Hi,

It seems the recursive DeepCopy used when saving a new version has a serious performance impact when you have a lot of relationships. The following is a blackfire profile for a batch edit of 25 objects:

https://blackfire.io/profiles/9902f36c-0857-4511-ac3a-54bf25e10a29/graph

When I make a shallow copy it's significantly faster. Is there any reason for traversing all of the related objects? Of course direct relationships have to be included because they are shown in the version tab but any deeper than this is not included in the version data anyway.

brusch:
Related to #4154

weisswurstkanone:
Can you confirm that you are on the latest release? With 6.1.0 we added a ReplaceFilter which optimize the whole stuff.

weisswurstkanone:
Not directly related but for completeness: https://github.com/pimcore/pimcore/pull/5858

brusch:
With #4667 the issue with many relations should be resolved. 
#5858 brings also some additional improvements, especially for Data Objects. 

One thing that can be further improved is that we're setting `children` to null, before doing the deep copy.

weisswurstkanone:
Not setting them to null will still not cause that they get traversed, right?
Reason is that they are treated just like a relation (replaced with an ElementDescriptor)

kubaplas:
Hi,
I can contribute my profile also: https://blackfire.io/profiles/8252677b-5d00-406f-a1f1-53f8bbe2066a
Somewhat complex data model with classification store, but most of the fields are empty.
Using dev-master (80b6afa115a2d174b64cc9f69b2a9426d42173ef)

weisswurstkanone:
see https://github.com/pimcore/pimcore/pull/5917

belendel:
> Can you confirm that you are on the latest release? With 6.1.0 we added a ReplaceFilter which optimize the whole stuff.
> 
> See #4667

This was tested with 6.4.0. I believe it still traverses the whole tree even with the ReplaceFilter. I can  do a new profile with the latest version if you want.

weisswurstkanone:
thx!

kubaplas:
@weisswurstkanone I think that #5917 is not related to this task - I've made profiles for #5917 with versioning disabled.

weisswurstkanone:
It is related to the general performance of versioning :-) Just wanted to link them together.

weisswurstkanone:
[...]

PR #5981, opened 2020-03-05
Versioning performance with a lot of relationships #5813

weisswurstkanone:


brusch:
@weisswurstkanone is this also relevant for the recycle bin?

weisswurstkanone:
performance is probably less important there, but let's add it. I'll create a PR

PR #5983, opened 2020-03-05
[DOCS] Update 13_Building_Custom_Rest_APIs.md

pdchaudhary:
doc link: https://pimcore.com/docs/6.x/Development_Documentation/Best_Practice/Building_Custom_Rest_APIs.html
There is two error coming on 1) undefined method name checkPermission 2) undefined method name adminJson
so I changed this way to element these errors by extends 'Pimcore\Bundle\AdminBundle\Controller\Rest\AbstractRestController' and changing routes.

Thank you

  
  

Resolves #

### DIFF
--- a/doc/Development_Documentation/26_Best_Practice/13_Building_Custom_Rest_APIs.md
+++ b/doc/Development_Documentation/26_Best_Practice/13_Building_Custom_Rest_APIs.md
@@ -16,17 +16,17 @@ in the desired format.
 
 namespace AppBundle\Controller;
 
-use Pimcore\Controller\FrontendController;
 use Pimcore\Model\DataObject;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
+use Pimcore\Bundle\AdminBundle\Controller\Rest\AbstractRestController;
 
 
-class CustomRestController extends FrontendController
+class CustomRestController extends AbstractRestController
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
thx!

thgross:
But why is `Pimcore\Bundle\AdminBundle\Controller\Rest\AbstractRestController` deprecated?

brusch:
@thgross because the entire REST API is deprecated, see #5517

thgross:
I understand. But then the documentation shouldn't encourage using a deprecated API.

brusch:
The note is just missing on this documentation page, on the main page about REST services it's already mentioned. 
See also: https://github.com/pimcore/pimcore/blob/8ec544cafa2956bf8227c732babc657be9ab4fae/doc/Development_Documentation/24_Web_Services/README.md#L3

dvesh3:
@brusch I would say, it should not be suggested to use deprecated API in best practices. How about we change the doc to extend `AdminController` instead of `AbstractRestController`? this will also solve the checkPermission  & adminJson issue.

brusch:
@dvesh3 see 230217855e14b7865f8e7a8a0de269bbb977df16 😉

dvesh3:
@brusch nice! we should also do something about `adminJson` in example.

pdchaudhary:
routes endpoint also needs to change if REST API is deprecated. because using these routes "/webservice/rest/*" , it will use the deprecated api key method for authentication.

PR #5987, opened 2020-03-05
[Listings] Prevent infinite loop

### LINKED ISSUE #5985: [Data object listing] Endless loop
The following script never ends:
```php
<?php
$object = new \Pimcore\Model\DataObject\SupplierRequest();
$list = $object->getList();

foreach($list as $item) {
  var_dump($list->getTotalCount());
  var_dump($item->getId());
}
```
When I fetch `$list->getTotalCount()` before the loop and remove this call inside the loop everything works fine.
Of course this example this simplified - in my code it is not so easy to move the `getTotalCount()` call outside...

PS: Also this works:
```php
foreach($list->load() as $item) {
  var_dump($list->getTotalCount());
  var_dump($item->getId());
}
```
So it seems to have something to do with the `Iterator` interface.

BlackbitNeueMedien:
Resolves #5985

### DIFF
--- a/lib/Model/Listing/AbstractListing.php
+++ b/lib/Model/Listing/AbstractListing.php
@@ -433,8 +433,6 @@ abstract class AbstractListing extends AbstractModel implements \Iterator
      */
     public function setConditionVariables($conditionVariables)
     {
-        $this->setData(null);
-
         $this->conditionVariables = $conditionVariables;
 
         return $this;


brusch:
@BlackbitNeueMedien Stan is complaining, can you have a look plz? 
@dvesh3 I guess `$this->setData(null);` is done for a reason, right? Can you plz check? 😊

dvesh3:
@brusch yes, we changed it for fixing children and sibling cache result https://github.com/pimcore/pimcore/pull/5550/commits/f63394eb46b88c3853abb0419863d8af7a64d871 along with other changes. I'm ok with this PR change, since `setConditionVariables` is called internally by getCondition() only.

brusch:
@dvesh3 thanks for clarifying! 😊

brusch:
I'm wondering why Stan is complaining? Just a false positive?

dvesh3:
maybe we need to update baseline file. @weisswurstkanone ?

blankse:
The baseline builds were successful. The unsuccessful builds should be allowed to fail. Was there a travis mistake?

PR #5989, opened 2020-03-06
Add event to allow programmatic manipulation of element permissions

BlackbitNeueMedien:
This event gives developers a way to add their own custom logic when determining whether a user should be granted a certain permission.

I was facing a situation where, in a multi-tenant-setup, all tenants have similar basic roles.
Say, for data objects, I have folder structure like:

```
/My Tenants
    |-- tenant 1
           |-- customers
           |-- products
           |-- orders
    |-- tenant 2
           |-- customers
           |-- products
           |-- orders
    |-- tenant 3
           |-- customers
           |-- products
           |-- orders
```

A role "Product Manager" should have access to `/My Tenants/<whatever tenant I belong to>/products`. With hundreds of tenants and half a dozen or so roles it can be
cumbersome to handle this with the default setup of workspaces. With this event, I have to possibility to fine-tune permissions in any way I like.

### DIFF
--- a/lib/Event/AdminEvents.php
+++ b/lib/Event/AdminEvents.php
@@ -404,4 +404,19 @@ final class AdminEvents
      * @var string
      */
     const RESOLVE_ELEMENT_ADMIN_STYLE = 'pimcore.admin.resolve.elementAdminStyle';
+
+    /**
+     * Allows you to modify whether a permission on an element is granted or not
+     *
+     * Subject: \Pimcore\Model\Element\AbstractElement
+     * Arguments:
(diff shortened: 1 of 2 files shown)


brusch:
Makes sense, thanks a lot!

weisswurstkanone:
Please mention this nice feature in
https://pimcore.com/docs/6.x/Development_Documentation/Administration_of_Pimcore/Users_and_Roles.html
(just add a link to AdminEvents.php)

brusch:
@weisswurstkanone done by a499e3dee69841fae7493e0afa5b386472047bcb

PR #5991, opened 2020-03-06
[DataObject] Fix inconsistent version diffs

t-zilla:

Fixes a type juggling issue that leads to unintuitive comparison of data object versions.

For example:

* `null` changed to `0` _doesn't_ show up as modification, even though you can see the difference in the table
* `null` changed to empty string `""` _does_ show up as modification, even though you cannot see the difference in the table

## Additional info  

To reproduce:

1. Create an object and populate two fields with `null` value and save.
2. Change value of one field to `0`, change value of the other field to empty string `""` and save.
3. Compare object versions in admin panel.


weisswurstkanone:
Just thinking about the overview.

Maybe handle this better here (instead of in the template)

public function getVersionPreview

For a string it doesn't matter if it is null or ''
For a number it definitely makes a difference if it is null or 0

brusch:
@t-zilla can you please update your PR based on @weisswurstkanone feedback? 
Thanks a lot!

t-zilla:
@weisswurstkanone 

Can you elaborate on your idea? Correct me if I'm wrong, fixing this bug outside template would require a massive refactor.

`public function getVersionPreview` has many implementations and they aren't well type-hinted. I can't tell which implementation returns `string`, `int` or perhaps even something else.

weisswurstkanone:
As said above, it depends on the data type and should be handled there.

My suggestion is to generate a decent version preview using getVersionPreview 
and for comparison there isEqual.

getVersionPreview is just a short human-readable presentation of old and new data.
isEqual is responsible for comparing the data. If it yields wrong results it should be fixed there but not in the template.

weisswurstkanone:
@t-zilla ping, any news on this?

weisswurstkanone:
please rever diffVersions.html.php  as this is already handled by #6160 in an alternative way

t-zilla:
> r

Hey, thanks for linking. I'll update this as soon as isEqual is implemented for all datatypes (https://github.com/pimcore/pimcore/issues/6165) otherwise it would break current behaviour which gives us more consistency for the time being.

weisswurstkanone:
Please be aware that #6165 is scheduled for milestone 7 (https://github.com/pimcore/pimcore/milestone/70)

so this will take a while

brusch:
Closed for now, we should start a new PR once #6165 is fully implemented

PR #5993, opened 2020-03-06
Allow the user to be null in create function, save already allows this

questionmark78:
If you want to use the Recyclebin functionality in custom code, the Item::create function needs a user. However the save() function which is called inside the create function allows the user to be null.

This changes makes it possible to have the user parameter omitted in the create call.


questionmark78:
Commit was with wrong user account, need to change this

PR #5994, opened 2020-03-06
[Elements] Cache keys should be generated in a central place for consistency

brusch:
Follow up to #5974

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -480,6 +480,24 @@ class Service extends Model\AbstractModel
     }
 
     /**
+     * @param string $className
+     * @return string|null
+     */
+    public static function getElementTypeByClassName(string $className):?string
+    {
+        $className = trim($className, "\\");
+        if(is_subclass_of($className, AbstractObject::class) || $className === AbstractObject::class) {
+            return 'object';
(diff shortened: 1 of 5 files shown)


models/Element/Service.php
```diff
@@ -479,6 +479,24 @@ public static function getElementType($element)
...
+    public static function getElementTypeByClassName(string $className):?string
+    {
+        $className = trim($className, "\\");
+        if(is_subclass_of($className, AbstractObject::class) || $className === AbstractObject::class) {
+            return 'object';
+        } elseif (is_subclass_of($className, Asset::class) || $className === Asset::class) {
+            return 'asset';
+        } elseif (is_subclass_of($className, Document::class) || $className === Document::class) {
+            return 'document';
+        }
```
jdreesen:
```suggestion
        if (is_a($className, AbstractObject::class, true)) {
            return 'object';
        } elseif (is_a($className, Asset::class, true)) {
            return 'asset';
        } elseif (is_a($className, Document::class, true)) {
            return 'document';
        }
```

brusch: Thanks a lot! 👍

Issue #3974, opened 2019-02-21
[ASSETS] upload problems with large office files


putzflorian:
# Bug Report

Try to upload the Attached Excel File (or a certain bigger Excel File) you run in an error like "pimcore.ERROR: Couldn't load document /home/project/www/web/var/assets/_temp/downloads/2019_02_21_09_16_49_export.xlsx only PDF documents are currently supported [] []". After a few Minutes or you reload the Pimcore UI the File is successfully saved and is available!

It seems the error only occures on bigger Excel Files.

[2019_02_20_16_17_44_export.xlsx](https://github.com/pimcore/pimcore/files/2888413/2019_02_20_16_17_44_export.xlsx)

brusch:
Possible solution: do the processing in the background -> similar to the image optimization queue (using the `TmpStore` and a maintenance job).

PR #6001, opened 2020-03-06
Pdf preview is not working on grid

geraudi:
When you upload a new PDF, the thumbnail image is not generated immediatly and replaced by "filetype-not-supported.svg" image.

Capture du 2020-03-06 15-42-58


brusch:
Thanks for your PR!
However this behavior is intentional, the previews should be generated asynchronously in the background, because otherwise this will cause issues with huge documents. 
See also: #3974

PR #6004, opened 2020-03-08
[Asset] fix not authorized view on animated images if .htaccess publi…

ssiergl:
…c resriction is enabled

If pimcore setup restricts public asset access (like stated in the docs) animated images
cannot be viewed anymore due of direct path access.

add new controller action and return the animated image

## How to reproduce

- Change `.htaccess`like stated here and restrict some asset area: https://pimcore.com/docs/6.x/Development_Documentation/Assets/Restricting_Public_Asset_Access.html
- Upload an animated image into the restricted asset area
- Open image in backend....it will be broken
Screenshot 2020-03-08 at 20 02 25

## How it should be
At least in admin backend all images should be shown even if all assets are restricted from outside access.
Screenshot 2020-03-08 at 20 01 28
 

add new controller action and return the animated image

## Additional info  
I think for the backend ui this is still acceptable resource wise @brusch

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -124,7 +124,7 @@ class AssetController extends ElementControllerBase implements EventedController
 
             $imageInfo['previewUrl'] = sprintf('/admin/asset/get-image-thumbnail?id=%d&treepreview=true&hdpi=true&_dc=%d', $asset->getId(), time());
             if ($asset->isAnimated()) {
-                $imageInfo['previewUrl'] = $asset->getFullPath() . '?_dc=' . time();
+                $imageInfo['previewUrl'] = sprintf('/admin/asset/get-image?id=%d&_dc=%d', $asset->getId(), time());
             }
 
             if ($asset->getWidth() && $asset->getHeight()) {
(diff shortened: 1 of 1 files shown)


brusch:
@ssiergl Thanks a lot! Yes, for the UI that's totally fine. 
I've just renamed `get-image` to `get-asset` as the function itself is generic and maybe we need the same functionality somewhere else as well (can't hurt 😊).

bundles/AdminBundle/Controller/Admin/AssetController.php
```diff
@@ -124,7 +124,7 @@ public function getDataByIdAction(Request $request, EventDispatcherInterface $ev
 
             $imageInfo['previewUrl'] = sprintf('/admin/asset/get-image-thumbnail?id=%d&treepreview=true&hdpi=true&_dc=%d', $asset->getId(), time());
             if ($asset->isAnimated()) {
-                $imageInfo['previewUrl'] = $asset->getFullPath() . '?_dc=' . time();
+                $imageInfo['previewUrl'] = sprintf('/admin/asset/get-asset-?id=%d&_dc=%d', $asset->getId(), time());
```
blankse:
```suggestion
                $imageInfo['previewUrl'] = sprintf('/admin/asset/get-asset?id=%d&_dc=%d', $asset->getId(), time());
```

brusch: Thanks a lot! 🙏

ssiergl:
LGTM

PR #6005, opened 2020-03-08
[User|Role] be consistent on checkbox UI and use boxLabel

ssiergl:
most checkboxes have boxLabel in User and Role tab. Permission checkboxes should go inline.
This will allow long descriptions for custom permissions as well.

## Before
Screenshot 2020-03-08 at 21 01 54

Screenshot 2020-03-08 at 21 01 01

Screenshot 2020-03-08 at 21 07 52


brusch:
Nice, thanks a lot! 🙏

PR #6009, opened 2020-03-09
[Assets] Exception handling - trigger error message in admin UI - #5912

dvesh3:
related to #5912


bundles/AdminBundle/Controller/Admin/AssetController.php
```diff
@@ -316,23 +323,30 @@ public function addAssetAction(Request $request, Config $config)
      * @param Request $request
      * @param Config $config
      *
-     * @return JsonResponse
+     * @return Response/JsonResponse
```
blankse:
```suggestion
     * @return JsonResponse
```

dvesh3: thanks! not sure why i added that 😆

PR #6015, opened 2020-03-10
[Asset] Document content are no longer indexed in the search

blankse:
## Additional info  

The `document_page_count` custom setting is only set via cronjob. So the backend search can't get the text from the document and the logger writes a wrong log entry (document adapter is available)

### DIFF
--- a/models/Asset/Document.php
+++ b/models/Asset/Document.php
@@ -129,7 +129,7 @@ class Document extends Model\Asset
      */
     public function getText($page = null)
     {
-        if (\Pimcore\Document::isAvailable() && \Pimcore\Document::isFileTypeSupported($this->getFilename()) && $this->getCustomSetting('document_page_count')) {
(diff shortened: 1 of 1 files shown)


brusch:
Thanks for your PR, however your change would make the asynchronous processing obsolete, because the document would be converted on the fly with this change, so we're back to what it was before #3974

So I think updating the index needs to be done here: https://github.com/pimcore/pimcore/blob/11e8d5f8ec44f22daef3435b9dea8dca6cd728c0/lib/Maintenance/Tasks/AssetDocumentConversionTask.php#L55 

See also #6001

brusch:
@blankse #6016 should fix the issue. Thanks anyway 😊👍

blankse:
Ok, than I would change the Log message and i would allow it on cli calls. My log in scripts with documents a full of this message...

brusch:
@blankse CLI good idea, will update my PR!

PR #6016, opened 2020-03-10
[Asset] getText() on documents should process the asset synchronously called from CLI

brusch:
Follow-up to #6015

### DIFF
--- a/lib/Maintenance/Tasks/AssetDocumentConversionTask.php
+++ b/lib/Maintenance/Tasks/AssetDocumentConversionTask.php
@@ -51,9 +51,9 @@ final class AssetDocumentConversionTask implements TaskInterface
 
             try {
                 if ($asset instanceof Asset\Document) {
-                    $this->logger->debug(sprintf('Processing document with ID %s | Path: %s', $asset->getId(),
-                        $asset->getFullPath()));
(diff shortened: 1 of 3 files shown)


brusch:
@blankse can you confirm this works for you?

lib/Maintenance/Tasks/AssetDocumentConversionTask.php
```diff
@@ -50,9 +51,15 @@ public function execute()
...
-                        $asset->getFullPath()));
+                    $this->logger->debug(sprintf('Processing document with ID %s | Path: %s', $asset->getId(), $asset->getFullPath()));
                     $asset->processPageCount();
+                    $asset->save();
+
+                    $searchEntry = Data::getForElement($asset);
+                    if ($searchEntry instanceof Data and $searchEntry->getId() instanceof Data\Id) {
+                        $searchEntry->setDataFromElement($asset);
+                        $searchEntry->save();
+                    }
```
blankse: `$asset->save();` also generate the search entry, right?

blankse:
```suggestion
```

brusch: You're totally right, `save()` does already the job :)

PR #6018, opened 2020-03-10
Trigger useSession before executing HtmlToImage Command

solverat:
On most server, the document version compare does not work (It works on your demo instance though). Most of the time, this process will run for 60sec before a timeout kicks in. After that, no image has been generated.

After some research, I found a similar issue: https://github.com/KnpLabs/KnpSnappyBundle/issues/42

This PR will fix this issue.


brusch:
🍪 Nice, thanks a lot! 🙏

PR #6019, opened 2020-03-10
Update toolbar.js

roland4432:
Fix typo to check for correct user permissions when checking for quick search if corresponding user was only granted asset permission

  
  

Resolves wrong permission check


dvesh3:
@roland4432 thanks a lot!

roland4432:
This has to be shortest PR ever :D

PR #6022, opened 2020-03-10
Localized field translate labels in tabs

### LINKED ISSUE #6021: Localized field tab labels not translatable
### Expected behavior

Admin translations are applied to localized field tabs.

### Actual behavior

They are not.

### Steps to reproduce

- Log in on the demo environment.
- Open admin translations and find the 'German' key.
- Set the English translation to 'DE'.
- Open a data object containing localized fields (AccessoryPart for example).
- Notice the label in the tab is not translated

- Optionally edit the localized field in the classes and enable split view.
- Open the split view settings in the data object. The label is translated here.

Localized_field_translation

### Possible fix

Adding a translate function to the title.
https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/localizedfields.js#L479

I created a pull request for this, but I wanted to check two things first:

- Perhaps there is a specific reason this title wasn't translated?
- I noticed the file was [changed recently and all translations have been changes from ts() to t()](https://github.com/pimcore/pimcore/commit/0f26b1e0aec53cb109ea43ffac0fba152f769ad5#diff-067bfff709d898820f46a4250601a73c). Since ts appears to be associated with admin translations and t with shared translations I wanted to check whether this is correct.

As a sidenote: I'd like this to work since I have an environment with a lot of languages, but still want to show short labels and not icons only. Labels like 'English (United Kingdom)' take a lot of space compared to something like 'EN (UK)'.

kjkooistra-youwe:
Resolves #6021

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/localizedfields.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/localizedfields.js
@@ -476,7 +476,7 @@ pimcore.object.tags.localizedfields = Class.create(pimcore.object.tags.abstract,
             });
         } else {
             item.iconCls = "pimcore_icon_language_" + language.toLowerCase();
-            item.title = pimcore.available_languages[language];
+            item.title = t(pimcore.available_languages[language]);
         }
     },
 


brusch:
I'm wondering why `pimcore.available_languages` doesn't contain the translated values in your case, because this should be the case, see: https://github.com/pimcore/pimcore/blob/cd1d124b0a061f12608180d03c8d725edaed0bf4/lib/Tool.php#L193

Maybe the ICU package is missing on your system? 
But anyway, I guess passing the value through `t()` can't hurt.

kjkooistra-youwe:
As mentioned in issue #6021 I reproduced it on the demo environment. Anyway, thanks.

PR #6024, opened 2020-03-11
Make multiselect values in advanced many-to-many-relations translatable

BlackbitNeueMedien:
When editing multiselect values in an advanced many-to-many relation, the edit window will already show translated values. In the grid, however, the values are not translated as of yet. This PR adds a renderer to the multiselect column that applies the translation method to every selected option.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
@@ -156,44 +156,52 @@ pimcore.object.tags.advancedManyToManyObjectRelation = Class.create(pimcore.obje
                 cellEditor = function() {
                     return new Ext.form.TextField({});
                 };
-            } else if (this.fieldConfig.columns[i].type == "select" && !readOnly) {
-               var selectData = [];
+            } else if (this.fieldConfig.columns[i].type == "select") {
+                if(!readOnly) {
+                    var selectData = [];
 
-                if (this.fieldConfig.columns[i].value) {
-                    var selectDataRaw = this.fieldConfig.columns[i].value.split(";");
+                    if (this.fieldConfig.columns[i].value) {
+                        var selectDataRaw = this.fieldConfig.columns[i].value.split(";");
 
-                    for (var j = 0; j < selectDataRaw.length; j++) {
-                        selectData.push([selectDataRaw[j], t(selectDataRaw[j])]);
+                        for (var j = 0; j < selectDataRaw.length; j++) {
+                            selectData.push([selectDataRaw[j], t(selectDataRaw[j])]);
+                        }
                     }
+
+                    cellEditor = function(selectData) {
+                        return new Ext.form.ComboBox({
+                            typeAhead: true,
+                            queryDelay: 0,
+                            queryMode: "local",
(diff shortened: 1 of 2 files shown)


dvesh3:
@BlackbitNeueMedien Isn't it should be already handled by `pimcore.object.helpers.metadataMultiselectEditor`? https://github.com/pimcore/pimcore/blob/53e8d494faea79fc17c7e43071f0bbb59a4b7bfe/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/metadataMultiselectEditor.js#L47

BlackbitNeueMedien:
@dvesh3 In the editor window, yes. But after making a selection, the selected values are shown in the corresponding grid column. The translator is not applied there.

dvesh3:
then it should be done somewhere here: https://github.com/pimcore/pimcore/blob/53e8d494faea79fc17c7e43071f0bbb59a4b7bfe/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js#L394

dvesh3:
Hey @BlackbitNeueMedien, any concerns here?

BlackbitNeueMedien:
Hi @dvesh3,

thanks for your reply but I think `pimcore.object.helpers.grid.prototype.advancedRelationGridRenderer` does only get applied in the grid, not when editing an object.
When editing an object with an advanced many-to.many object relation the following function gets executed:
https://github.com/pimcore/pimcore/blob/b349c8621aabb450093217f60d4f366ba88ed2b7/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js#L529-L531

This calls `createLayout` which creates the columns for the meta fields in https://github.com/pimcore/pimcore/blob/b349c8621aabb450093217f60d4f366ba88ed2b7/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js#L141-L241
There is no call to `pimcore.object.helpers.grid.prototype.advancedRelationGridRenderer` here.
Moreover `select` meta fields get a special renderer which handles translations in https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js#L222-L226 - but multiselect fields do not.

dvesh3:
Gotcha! sorry, i was looking at wrong place (folder grid). Thanks for your 
explanation!

but your renderer will only work if field is **not readonly** https://github.com/pimcore/pimcore/blob/b349c8621aabb450093217f60d4f366ba88ed2b7/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js#L191 so we need to fix it by moving it next to select renderer https://github.com/pimcore/pimcore/blob/b349c8621aabb450093217f60d4f366ba88ed2b7/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js#L222-L226 or better if we can refactor the readonly check to move down for celleditor only.

dvesh3:
@BlackbitNeueMedien please see https://github.com/pimcore/pimcore/pull/6024#issuecomment-601087723

BlackbitNeueMedien:
Refactored the code to support renderer also for readonly fields.

dvesh3:
nice! thanks a lot 🙂

PR #6030, opened 2020-03-11
[Data Objects] Fixed inherited classification-store data is not displ…

### LINKED ISSUE #6029: [Data Objects] ClassificationStore grid view inherited data
For a future customer project, the classification store in the grid view must work with inheritance. Unfortunately this is not yet happening.

I already fixed that in the forked-repo and I will push that right now.

skochTC:
fixes #6029

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -332,35 +332,7 @@ class Service extends Model\Element\Service
                 } elseif (strpos($key, '~') === 0) {
                     $type = $keyParts[1];
                     if ($type === 'classificationstore') {
-                        $field = $keyParts[2];
-                        $groupKeyId = explode('-', $keyParts[3]);
-
-                        $groupId = $groupKeyId[0];
-                        $keyid = $groupKeyId[1];
-                        $getter = 'get' . ucfirst($field);
-                        if (method_exists($object, $getter)) {
-                            /** @var Classificationstore $classificationStoreData */
-                            $classificationStoreData = $object->$getter();
-
-                            /** @var Model\DataObject\ClassDefinition\Data\Classificationstore $csFieldDefinition */
-                            $csFieldDefinition = $object->getClass()->getFieldDefinition($field);
-                            $csLanguage = $requestedLanguage;
-                            if (!$csFieldDefinition->isLocalized()) {
-                                $csLanguage = 'default';
-                            }
-
-                            $fielddata = $classificationStoreData->getLocalizedKeyValue($groupId, $keyid, $csLanguage, true, true);
-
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Hi @skochTC , can you provide steps to reproduce + point out the significant change as some restructuring happened as well.
Thanks!

skochTC:
> 
> 
> Hi @skochTC , can you provide steps to reproduce + point out the significant change as some restructuring happened as well.
> Thanks!

Hi,

to report this you need to create a store with some keys. Make sure that inheritance is enabled in the class definition. Now create a parent and a child element. The inherited data of the parent element is displayed in the object view in the child. But they are not displayed in the GRID view.

For a better explanation, see the following screenshot:
image

After my changes, the inherited data from the store is displayed correctly in the GRID view:
image

> point out the significant change as some restructuring happened as well.

I'm not really sure I got that right. I' ve added a comment in the code

models/DataObject/Service.php
```diff
@@ -446,6 +417,16 @@ public static function gridObjectData($object, $fields = null, $requestedLanguag
                         }
                     }
 
+                    // because the key for the classification store has not a direct getter, you have to check separately if the data is inheritable
+                    if (strpos($key, '~') === 0 && empty($data[$key])) {
```
weisswurstkanone: guess the data-types isEmpty method should be used instead.

does (strpos($key, '~') === 0 also work for bricks and so on ?

models/DataObject/Service.php
```diff
@@ -724,6 +705,62 @@ private static function getValueForObject($object, $key, $brickType = null, $bri
...
+                    if (method_exists($definition, 'getDataForGrid')) {
+                        $fielddata = $definition->getDataForGrid($fielddata, $object);
+                    }
+
+                    return $fielddata;
+                }
+            }
+        }
+
+        return '';
```
weisswurstkanone: returning an empty string here is changing the original behavior

models/DataObject/Service.php
skochTC: > guess the data-types isEmpty method should be used instead.

For Classification-Store keys `$def` is `null`. So `isEmpty` does not work.

> does (strpos($key, '~') === 0 also work for bricks and so on ?

Right, I'll fix that right now.

models/DataObject/Service.php
skochTC: Okay in order to not change the original behavior, I return "null" like it was before.

models/DataObject/Service.php
skochTC: Fixed in commit: 3e03840b05b74a5aa98a9440d891d492bb4307ad

weisswurstkanone:
thanks!

PR #5935, opened 2020-02-29
[PHPStan] Fix level 3 errors

blankse:
### Level 3
Before (without baseline file)
`[ERROR] Found 1086 errors`

After (without baseline file)
`[ERROR] Found 492 errors`

### Level 2
Before (without baseline file)
`[ERROR] Found 315 errors`

After (without baseline file)
`[ERROR] Found 301 errors`

PR #6032, opened 2020-03-12
Fix return types for Relation::getClasses

BlackbitNeueMedien:
As `AbstractRelations::getClasses()` really exists we do not need the `@method string[] getClasses()` in the `Relation` trait - this only confuses the IDE.

I actually wonder why all the other methods do not exist or what the `@method` documentations are for - because there is neither a `__call()` function nor do the methods `getAssetTypes()`, `getDocumentTypes()` etc. exist.
They actually only confuse the IDE and if you use those functions you will get errors depending on the actual relation `Data` class.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
+++ b/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
@@ -37,7 +37,7 @@ abstract class AbstractRelations extends Data implements
      *
      * @var array
      */
-    public $classes;
+    public $classes = [];
 
     /** Optional path formatter class
      * @var null|string
--- a/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
+++ b/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
@@ -45,11 +45,13 @@ abstract class AbstractRelations extends Data implements
     public $pathFormatterClass;
 
     /**
-     * @return array
+     * @return array[
+     *  'classes' => string,
+     * ]
      */
     public function getClasses()
     {
-        return $this->classes;
+        return $this->classes ?: [];
     }
 
     /**
--- a/bundles/EcommerceFrameworkBundle/Controller/AdminOrderController.php
+++ b/bundles/EcommerceFrameworkBundle/Controller/AdminOrderController.php
@@ -278,7 +278,7 @@ class AdminOrderController extends AdminController implements EventedControllerI
                 $field = $order->getClass()->getFieldDefinition('customer');
                 if ($field instanceof ManyToOneRelation) {
                     $classes = $field->getClasses();
-                    if (count($classes) == 1) {
+                    if (count($classes) === 1) {
                         $class = 'Pimcore\Model\DataObject\\' . reset($classes)['classes'];
                         /* @var \Pimcore\Model\DataObject\Concrete $class */
 
--- a/models/DataObject/ClassDefinition/Data/Extension/Relation.php
+++ b/models/DataObject/ClassDefinition/Data/Extension/Relation.php
@@ -78,4 +78,11 @@ trait Relation
 
         return $class;
     }
+
+    /**
+     * @return array[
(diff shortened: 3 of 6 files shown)


blankse:
But the `Relation  `trait hasn't a `getClasses()` method. So this PHPDoc makes sense (also for PHPStan > level 2 I think). I correct the type also in #5935

BlackbitNeueMedien:
Exactly - it also does not have all the other methods described in the `@method`s
I wonder what these `@method`s are good for. So I also do not understand why you say `So this PHPDoc [for getClasses()] makes sense` because as you said there is no method `getClasses()` in the trait.

blankse:
But it is used in the `getPhpDocClassString()` method. So how should the IDE or PhpStan now the retun types?

blankse:
It would also be possible to create abstract methods for this ;)
They must be present in the class that uses this trait.

BlackbitNeueMedien:
Puuh, that seems completely wrong to me. Imho traits should be self-contained. These `@method` statements define that there will be a method `getAssetsAllowed()` in any class which uses the `Relation` trait. And that is not correct.

blankse:
Oh yes, getDocumentsAllowed and getAssetsAllowed are optional...

BlackbitNeueMedien:
Right, so I think the correct way would be to introduce those methods in `AbstractRelations` class, e.g. with 
`public function getDocumentsAllowed(){ return false;}`

And add those methods as abstract methods to the `Relation` trait.

blankse:
You can add all methods in the trait with false and empty arrays. So the classes can override it and we don't must call method_exists. Or?

BlackbitNeueMedien:
yes, that is even better

blankse:
`method_exists($this, 'getDocumentsAllowed')` and `method_exists($this, 'getAssetsAllowed')` can be removed than.

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -45,11 +45,13 @@ abstract class AbstractRelations extends Data implements
...
     /**
-     * @return array
+     * @return array[
+     *  'classes' => string,
+     * ]
      */
     public function getClasses()
     {
-        return $this->classes;
+        return $this->classes ?: [];
```
blankse: I think this could be done also for `getDocumentTypes` and `getAssetTypes`

We can also add a default value `[]` to the properties

BlackbitNeueMedien: But `getDocumentTypes()` and `getAssetTypes()` does not exist in `AbstractRelations`. So where do you think we could add this default value?

If we can set the default value to the property I cannot judge. I am always cautious doing that because Pimcore decides on a lot of places whether to load or not to load certain data based on a property being an (empty) array or null. Are you really sure it is ok to set the daefult value of `$classes` property in AbstractRelations to `[]`?

blankse: This methods exists in ManyToManyRelation and ManyToOneRelation.

If you create the fields in the admin interface this properties will filled with arrays. But if you create it via script it can be that this properties are null. So I would prefer to set it with a default value. I think that can't hurt.

In 7.0 we can protect this properties and we can make the correct type in the setter.
Right @weisswurstkanone :) ?

blankse: Yes also the default value for the `$classes` property in `AbstractRelations`

blankse: Also a default value for the `objectsAllowed` property in `ManyToManyRelation`. Than I think its perfect :)

BlackbitNeueMedien: Thanks for your assistance and feedback @blankse!

weisswurstkanone:
thx!

PR #6036, opened 2020-03-12
[THUMBNAILS] add option for single asset-image thumbnail generator

putzflorian:


dvesh3:
@putzflorian how about we accept multiple image IDs instead? Then you can generate thumbnails for specific images, so something like https://github.com/pimcore/pimcore/blob/dd9fc1a906b7b6eed72af8ebffa9733baf1f9708/bundles/CoreBundle/Command/LowQualityImagePreviewCommand.php#L32-L37

putzflorian:
@dvesh3 sounds good

dvesh3:
LGTM. Thanks a lot!

PR #6041, opened 2020-03-13
[PHPStan] Fixes

blankse:
Before
```
[ERROR] Found 488 errors
```

After
```
[ERROR] Found 480 errors
```

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/DocumentControllerBase.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/DocumentControllerBase.php
@@ -147,17 +147,17 @@ abstract class DocumentControllerBase extends AdminController implements Evented
 
     /**
      * @param Model\Document $document
+     * @param array $data
      */
-    protected function addTranslationsData(Model\Document $document)
+    protected function addTranslationsData(Model\Document $document, array &$data)
     {
         $service = new Model\Document\Service;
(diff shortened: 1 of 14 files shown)


bundles/AdminBundle/Controller/Admin/Document/DocumentControllerBase.php
```diff
@@ -214,8 +214,8 @@ protected function getFromSession($doc)
             // see also PageController::clearEditableDataAction() | this is necessary to reset all fields and to get rid of
             // outdated and unused data elements in this document (eg. entries of area-blocks)
 
-            if ($sessionDocument = Model\Document\Service::getElementFromSession('document', $doc->getId())
-                            && $documentForSave = Model\Document\Service::getElementFromSession('document', $doc->getId(), '_useForSave')) {
+            if (($sessionDocument = Model\Document\Service::getElementFromSession('document', $doc->getId())) &&
+                ($documentForSave = Model\Document\Service::getElementFromSession('document', $doc->getId(), '_useForSave'))) {
```
blankse: There might be a bug here at the moment. PHPStan means the variable is a boolean. First the && is executed and then the =. But in right side there is also a =. I don't know how this is handled. With braces it should be safe.

brusch:
Thanks a lot!

PR #6043, opened 2020-03-16
Fixes settings_abstract.js to align inputs & elements

NiklasBr:
There was a duplicate declaration of `defaults` which overrode the width/labelWidth settings inside the Controller, Action & Template fieldset.

Now it looks like this:

Skärmbild (315)


brusch:
Nice, thanks a lot! 👍

PR #6044, opened 2020-03-16
Feature: Change getSaveCopyName() to behave more like Operating Systems

NiklasBr:
Just a nice thing to have: If an element key already ends with _copy or copy_N, append/increment a digit to avoid _copy_copy_copy chains if pasting the same object multiple times.

For example, in Windows, pasting a file multiple times will result in: *filename - copy* and then *filename - copy (2)*, in Ubuntu it becomes *filename (copy 1)* and *filename (copy 2)* and similarly will MacOS.

Edit: By the way, I'm using `array_key_last` here, it's a PHP 7.3 requirement, but I'm assuming symfony/polyfill-php73 is loaded if anyone is using PHP 7.2.


brusch:
Makes sense, this came from a time where special characters were not supported in keys/filenames 😉
Thanks a lot!

PR #6047, opened 2020-03-16
[Elements] Allow user to set userModification & modificationDate

### LINKED ISSUE #4365: [Element] Not possible to set userModification/modificationDate
I have a custom object (product) importer that create and update objects. The import is started from a front controller or a console command. I have created a pimcore admin user to represent this import. I am trying to set this user as the user modifying the object using the setUserModification method of the object. This does not work, the object is modified with user 0 instead of the user I set on the object.

I have located the problem in the method updateModificationInfos of class Pimcore\Model\Element\AbstractElement. Here the userModification value I have set is overwritten with the current admin user instead. In my case this will be the system user since there are no active admin user. 

I am unsure how to best solve this. My first idea was to only get the current user if the userModification property of the object is null. The problem then is that it will rarely be null. When an existing object is loaded the userModification property has the value from the last save. There is, as I see it, no reliable way of knowing if the value of userModification is a new set value or the value from last save. 

Or maybe I am looking at this elephant from the completely wrong angle?

### Expected behavior
The user id set on object with setUserModification method should be saved to the objects table in database when calling save on the object.

### Actual behavior
System (id 0) is used instead.

### Code example
```php
$product = new \Pimcore\Model\DataObject\Product();
$product->setSku('test-product');
$product->setParentId(1);
$product->setUserModification(3);
$product->save();
```

dvesh3:
Resolves #4365


models/Element/Traits/KeyValueTrait.php
```diff
@@ -0,0 +1,49 @@
...
+ * @category   Pimcore
+ * @package    Element
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Model\Element\Traits;
+
+trait KeyValueTrait
```
brusch: Maybe this name is a bit too generic, what about `TemporaryKeyValueTrait` or `InitialValueHelperTrait`? 
We should also tag this trait as `@internal`.

models/Element/Traits/KeyValueTrait.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @var array
+     */
+    protected $_temporaryValues;
```
brusch: This property needs to be excluded here: https://github.com/pimcore/pimcore/blob/163f95e61ac97b8d5997932827ba93cc85755b0b/lib/Model/AbstractModel.php#L217

models/Element/Traits/KeyValueTrait.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * @param string $fieldName
+     * @param mixed $matchValue
+     * @return bool
+     */
+    public function matchValueFromTemp($fieldName, $matchValue) {
+        $result = false;
+
+        if (isset($this->_temporaryValues[$fieldName]) && $matchValue) {
```
brusch: Is it really necessary that `$matchValue` is set? I don't think so, it can be also `false` right? 
Maybe we should just do the following: 
```php
if (isset($this->_temporaryValues[$fieldName])) {
    $result = $this->_temporaryValues[$fieldName] === $matchValue;
}
```

dpfaffenbauer:
Wouldn't it be possible to use the Dirty Identification stuff for that as well? When modDate or modUser is dirty, it must have been user-changed, if not, update it as it was before?

dvesh3:
@dpfaffenbauer  yes, it will work for data objects but not for assets and documents as they don't use dirty detection.

dpfaffenbauer:
not yet ;)

brusch:
@dpfaffenbauer good point! Thanks for your feedback, we're trying this approach 👍

models/Element/DirtyIndicatorInterface.php
```diff
@@ -15,7 +15,7 @@
  * @license    http://www.pimcore.org/license     GPLv3 and PEL
  */
 
-namespace Pimcore\Model\DataObject;
```
dpfaffenbauer: you moved the file but forget to add the BC file in its original location, didn't you?

dvesh3: Added. Thanks for review!

bundles/CoreBundle/Migrations/Version20200317163018.php
```diff
@@ -0,0 +1,41 @@
...
+            $class->save();
+        }
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
```
weisswurstkanone: please remove

models/Element/AbstractElement.php
```diff
@@ -44,23 +46,29 @@ protected function updateModificationInfos()
...
+            $this->setCreationDate($this->getModificationDate());
         }
 
-        // auto assign user if possible, if no user present, use ID=0 which represents the "system" user
-        $userId = 0;
-        $user = \Pimcore\Tool\Admin::getCurrentUser();
-        if ($user instanceof Model\User) {
-            $userId = $user->getId();
+        // auto assign user if possible, if not changed explicitly, if no user present, use ID=0 which represents the "system" user
+        $userModificationKey = Service::getElementType($this) === 'object' ? 'o_userModification' : 'userModification';
```
weisswurstkanone: why not just instanceof

models/DataObject/DirtyIndicatorInterface.php
```diff
@@ -1,45 +1,36 @@
...
+ * @package    Element
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Model\DataObject;
+
+@trigger_error(
+    'Interface Pimcore\Model\DataObject\DirtyIndicatorInterface is deprecated since version 6.0.0 and will be removed in 7.0.0. ' .
```
brusch: Version `6.6.0` 😉

models/DataObject/Traits/DirtyIndicatorTrait.php
```diff
@@ -17,56 +17,20 @@
...
-     *
-     * @return bool
-     */
-    public function isFieldDirty($key)
-    {
-        if (is_array($this->o_dirtyFields) && array_key_exists($key, $this->o_dirtyFields)) {
-            return true;
-        }
+@trigger_error(
+    'DirtyIndicatorTrait is deprecated since version 6.0.0 and will be removed in 7.0.0. ' .
```
brusch: Same here `6.6.0`

models/DataObject/Fieldcollection/Definition.php
```diff
@@ -165,7 +165,7 @@ public function save($saveDefinitionFile = true)
         $cd .= 'use Pimcore\Model\DataObject\PreGetValueHookInterface;';
         $cd .= "\n\n";
 
-        $implementsParts = ['\\Pimcore\\Model\\DataObject\\DirtyIndicatorInterface'];
+        $implementsParts = ['\\Pimcore\\Model\\Element\\DirtyIndicatorInterface'];
```
brusch: We should move that to the abstract class as well, just apply the interface on `\Pimcore\Model\DataObject\Fieldcollection\Data\AbstractData` then this can go away. 

Same for the trait and the object bricks (missing at the moment), see 
https://github.com/pimcore/pimcore/blob/663b0887cf4623b5499987e2b15b9522f3342e11/models/DataObject/Fieldcollection/Definition.php#L175
and
https://github.com/pimcore/pimcore/blob/cc1059ad1d4458aa410f9d46ffe279336993a941/models/DataObject/Objectbrick/Definition.php#L231

models/Element/Traits/DirtyIndicatorTrait.php
```diff
@@ -0,0 +1,82 @@
...
+            unset($this->o_dirtyFields[$field]);
+        }
+    }
+
+    public function resetDirtyMap()
+    {
+        $this->o_dirtyFields = null;
+    }
+
+    public function checkFieldDirty($field, $value, $compareExisting = true)
```
brusch: I think we better should make this `protected`.

models/Element/AbstractElement.php
dvesh3: done

bundles/CoreBundle/Migrations/Version20200317163018.php
dvesh3: done

models/DataObject/DirtyIndicatorInterface.php
dvesh3: done

models/DataObject/Traits/DirtyIndicatorTrait.php
dvesh3: done

models/DataObject/Fieldcollection/Definition.php
dvesh3: changed for Fieldcollection\Definition, Objectbrick\Definition & DataObject\Classdefinition

models/Element/Traits/DirtyIndicatorTrait.php
dvesh3: done

models/DataObject/AbstractObject.php
```diff
@@ -999,9 +1000,7 @@ public function setId($o_id)
     public function setParentId($o_parentId)
     {
         $o_parentId = (int) $o_parentId;
-        if ($o_parentId != $this->o_parentId && $this instanceof DirtyIndicatorInterface) {
-            $this->markFieldDirty('o_parentId');
-        }
+        $this->checkFieldDirty('o_parentId', $o_parentId);
```
weisswurstkanone: still needed ?

dvesh3: just future-proof :)

brusch: Hmm, I'm wondering ... it's just used here, can't we avoid it? Then we can kick the additional method.

dvesh3: kicked it ;)

brusch: Nice, thanks!

bundles/CoreBundle/Migrations/Version20200317163018.php
```diff
@@ -0,0 +1,41 @@
...
+     * @throws \Exception
+     */
+    public function up(Schema $schema)
+    {
+        $list = new DataObject\ClassDefinition\Listing();
+        $list = $list->load();
+
+        foreach ($list as $class) {
+            $this->writeMessage(sprintf('Saving class: %s', $class->getName()));
+            $class->save();
```
dpfaffenbauer: I am wondering, since this actually changes the class-definition file as well, wouldn't it be more sufficient to just re-generate the source-code files and done?

that doesn't just affect this PR, but also all future ones.

weisswurstkanone: @dvesh3 there is an option for this

    public function ClassDefinition::save($saveDefinitionFile = true)

brusch: Cool, yes! Let's do that!  @dvesh3 can you please update the PR once again? Thanks a lot 🙏

dpfaffenbauer: Just for clarification, maybe we should introduce a new function: `generatePhpClassFiles` or something. Maybe also introduce a service responsible for class generation. Guess that has already been propsoed?

brusch: Not a bad idea, but out of scope for this PR. 😊
Hmm, can't remember a proposal for this, but anyway, like I said, not a bad idea at all 😉

brusch: Just saw that the parameter doesn't help in this case, so we need to move the code into another public method ...

dvesh3: Introduced new public method `generateClassFiles` and now migration calls this new method instead of save.

dpfaffenbauer: 👍 Would be fine to have that in a separate PR, but fine for me :)

Thanks @dvesh3

dvesh3: Thanks for review guys 🙌

weisswurstkanone:
@dvesh3 can you resolve the conflict?

brusch:
Conflict resolved. 😉

brusch:
@dvesh3 One last thing ... regarding field collections and object bricks. I guess we also have to re-save the class files for them in the migration (like we do for class definitions). For that it probably makes also sense to introduce the `generateClassFiles()` method there as well.

dvesh3:
@brusch totally makes sense! will refactor and add to migrations 😉

PR #6049, opened 2020-03-17
CKE and editmode adjustments/fixes

NiklasBr:
This PR has two changes:

1. In editmode, the CKE modal dialogues did not properly position themselves above the WYSIWYG button panel, using `.cke_dialog_container` resolved that.
2. Any time a CKE modal dialogue were opened the background document jumped sideways due to the disappearing scroll bar, this keeps the scrollbar. The modal will still be centred in the viewport if the user scrolls and scrolling scrollable fields in the modal works as expected.


dvesh3:
nice work 👍  thanks a lot!

PR #6059, opened 2020-03-18
[Import] Wrong setParent on GetBy resolver

rquarta-extera:
This PR resolves a bug on GetBy resolver.

To reproduce:
1. Start a CSV import on any class.
2. In **Resolver Settings** choose _Resolver Strategy -> Get By Attribute_ and configure the resolver in this way.
bugfix-getby
In _Column_ field you have to choose the column inside your CSV file that matches with _Attribute_.

When you start the CSV import, you will get this error:
bugfix-getby-error

This is caused by a wrong method call on _setParent_, passing the ID as parameter. The solution fixes the method call.


weisswurstkanone:
thx!

PR #6060, opened 2020-03-18
Improve saving for inheritance - especially when you have a lot of lan…

ckogler-elements:
Problem: Saving of objects with inherited values takes quite long when you have a lot of data in the objects table .

- only execute the query when childs are available
-  use runtime cache and use ids to do the "slow" "like" ..%" query only once.

### DIFF
--- a/models/DataObject/Concrete/Dao/InheritanceHelper.php
+++ b/models/DataObject/Concrete/Dao/InheritanceHelper.php
@@ -422,7 +422,7 @@ class InheritanceHelper
         if (!$parentIdGroups) {
             $object = DataObject::getById($currentParentId);
 
-            if($object->hasChildren([$object::OBJECT_TYPE_OBJECT, $object::OBJECT_TYPE_FOLDER, $object::OBJECT_TYPE_VARIANT],true)){
+            if($object->getDao()->hasChildren([$object::OBJECT_TYPE_OBJECT, $object::OBJECT_TYPE_FOLDER, $object::OBJECT_TYPE_VARIANT],true)){
                 /**
                  * get the object keys once and reuse it because we don't have a key on o_path and it can take quite long when you have a lot of data
                  *
--- a/models/DataObject/Concrete/Dao/InheritanceHelper.php
+++ b/models/DataObject/Concrete/Dao/InheritanceHelper.php
@@ -421,43 +421,66 @@ class InheritanceHelper
 
         if (!$parentIdGroups) {
             $object = DataObject::getById($currentParentId);
-            if (isset($params['language'])) {
-                $query = "SELECT a.language as language, b.o_id AS id $fields, b.o_classId AS classId, b.o_parentId AS parentId FROM objects b LEFT JOIN " . $this->storetable . ' a ON b.o_id = a.' . $this->idField . ' WHERE o_path LIKE ' . $this->db->quote($object->getRealFullPath() . '/%')
-                    . ' HAVING `language` = "' . $params['language'] . '" OR ISNULL(`language`)'
-                    . ' ORDER BY LENGTH(o_path) ASC';
-            } else {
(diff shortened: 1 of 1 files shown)


dvesh3:
@ckogler-elements can you please check why tests are failing? thanks!

ckogler-elements:
@dvesh3 :Ähm, no i can't, because i cannot see why the test is failing

"We couldn't display the repository
pimcore/pimcore
This repository may not exist or you may need to use another account to have access to it"

weisswurstkanone:
https://travis-ci.com/github/pimcore/pimcore/jobs/299476777
doesn't work ?

weisswurstkanone:
seems that is broken now for cases where you have a folder or different class inbetween

models/DataObject/Concrete/Dao/InheritanceHelper.php
```diff
@@ -421,43 +421,70 @@ protected function buildTree($currentParentId, $fields = '', $parentIdGroups = n
...
+                 * get the object keys once and reuse it because we don't have a key on o_path and it can take quite long when you have a lot of data
+                 *
+                 */
+                $query = 'SELECT o_id FROM objects WHERE o_path LIKE ' .  $this->db->quote($object->getRealFullPath() . '/%');
 
-            if (!$parentIdGroups) {
-                $result = $this->db->fetchAll($query);
+                if (self::$useRuntimeCache) {
+                    $queryCacheKey = 'tree_ids_'.md5($query);
+                    if(is_null(self::$runtimeCache[$queryCacheKey])){
```
dvesh3: check for index
image

weisswurstkanone:
I think hasChildren returns false here as it is taken from the runtime cache

weisswurstkanone:
also see https://github.com/pimcore/pimcore/pull/5550

models/DataObject/Concrete/Dao/InheritanceHelper.php
```diff
@@ -421,43 +421,64 @@ protected function buildTree($currentParentId, $fields = '', $parentIdGroups = n
...
+                }else{
+                    $ids = $this->db->fetchCol($query);
+                }
 
-            if (!$parentIdGroups) {
-                $result = $this->db->fetchAll($query);
 
                 if (isset($params['language'])) {
-                    $result = $this->filterResultByLanguage($result, $params['language'], 'language');
+                    $query = "SELECT a.language as language, b.o_id AS id $fields, b.o_classId AS classId, b.o_parentId AS parentId FROM objects b LEFT JOIN " . $this->storetable . ' a ON b.o_id = a.' . $this->idField . ' WHERE  o_id IN( ' . implode(',',$ids).')'
```
brusch: I have a concern regarding the `o_id IN()` condition. If you have hundreds of thousands children, you're probably hitting `max_allowed_packet` .

brusch:
Thanks a lot for your PR! 
Im wondering why there should be a performance difference if both queries are using an index. Can you maybe share some benchmarks (incl. the `EXPLAIN` for each query) of your installation that shows the difference? 
Thanks a lot in advance! 😊

ckogler-elements:
Hi,
sorry - it turned out that it was an Issue with the application. 
The problem was that the fullpath index of the objects table was
UNIQUE INDEX `fullpath` ( `o_key`,`o_path`),
instead of 
UNIQUE INDEX `fullpath` (`o_path`, `o_key`),

so the index wasn't used which caused this slow behaviour. In older versions a index was set on o_path so  this wasn't a problem...

The only thing which could be added is the 
if($object->getDao()->hasChildren([$object::OBJECT_TYPE_OBJECT, $object::OBJECT_TYPE_FOLDER, $object::OBJECT_TYPE_VARIANT],true)){
so that the queries are not executed when no result is possible - but it dosn't realy improve the performace significantly...

weisswurstkanone:
thanks for your investigations especially regarding the other optimization!

brusch:
Alright, thanks a lot for your feedback!

PR #6062, opened 2020-03-19
Dynamic select in grouped objectbricks

### LINKED ISSUE #6046: Dynamic select in objectbricks
# Bug Report

### Expected behavior
In Data Objects it is possible to add an option provider for a select or multiselect to add dynamic values. When loading a data object the options provider is used to load the dynamic options. This is possible for both classes and object bricks.

We created a second class (see screenshot) with a table in which we added the values for the options provider. (2 columns, key and value). As soon as we add a new row in the table we expect the select options for the object brick to change when we (re)load a data object.

image

### Actual behavior
When (re)loading a data object the options for the dynamic select in a object brick are not reloaded. The options for the select that is added directly in the class are loaded as expected.

### Steps to reproduce
1. Create a class (eg. product)
2. Add a select as field with an an option provider (https://pimcore.com/docs/6.x/Development_Documentation/Objects/Object_Classes/Data_Types/Dynamic_Select_Types.html), see the screenshot below for an example as we implemented it and see the code for our provider
3. Add object brick field
4. Create object brick with same select field you added in step 2
5. Create a second (dropdown) class with table (see screenshot)
4. Create a new data object from step 5 and add some rows to the table
5. Create a new data object for the class you created in step 1
6. Check the options in both selects
7. Add new rows to the dropdown object
8. Close and reopen the product object and check options

image

### Code example
**DropdownOptionsProvider.php**
```php
<?php

namespace AppBundle\Product\OptionsProvider;

use Exception;
use Pimcore\Model\DataObject\ClassDefinition\Data;
use Pimcore\Model\DataObject\ClassDefinition\DynamicOptionsProvider\SelectOptionsProviderInterface;
use Pimcore\Model\DataObject\Dropdown;

class DropdownOptionsProvider implements SelectOptionsProviderInterface
{
    /**
     * @param array $context
     * @param Data $fieldDefinition
     * @return array
     * @throws Exception
     */
    public function getOptions($context, $fieldDefinition)
    {
        $dropdownPath = $this->getOptionsProviderData($fieldDefinition);
        if (empty($dropdownPath->dropdown)) {
            throw new Exception('No dropdown path is provided!');
        }

        $dropdown = Dropdown::getByPath($dropdownPath->dropdown);
        if (empty($dropdown) || !$dropdown instanceof Dropdown) {
            throw new Exception('Dropdown path does not exsist!');
        }

        return $dropdown->getItems();
    }

    /**
     * Returns the value which is defined in the 'Default value' field
     * @param array $context
     * @param Data $fieldDefinition
     * @return mixed
     */
    public function getDefaultValue($context, $fieldDefinition)
    {
        return $fieldDefinition->getDefaultValue();
    }

    /**
     * @param array $context
     * @param Data $fieldDefinition
     * @return bool
     */
    public function hasStaticOptions($context, $fieldDefinition)
    {
        return true;
    }

    private function getOptionsProviderData(Data $fieldDefinition)
    {
        return json_decode($fieldDefinition->getOptionsProviderData());
    }

}
```

weisswurstkanone:
Resolves #6046

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
@@ -1237,7 +1237,6 @@ class ClassController extends AdminController implements EventedControllerInterf
             } else {
                 if ($forObjectEditor) {
                     $layout = $item->getLayoutDefinitions();
-                    DataObject\Service::enrichLayoutDefinition($layout, $object);
 
                     $currentLayoutId = $request->get('layoutId', null);
 
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
@@ -1179,6 +1179,8 @@ class ClassController extends AdminController implements EventedControllerInterf
         $fieldname = null;
         $className = null;
 
+        $object = DataObject\AbstractObject::getById($request->get('object_id'));
+
         if ($request->query->has('class_id') && $request->query->has('field_name')) {
             $classId = $request->get('class_id');
             $fieldname = $request->get('field_name');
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
```diff
@@ -1232,6 +1237,7 @@ public function objectbrickTreeAction(Request $request)
             } else {
                 if ($forObjectEditor) {
                     $layout = $item->getLayoutDefinitions();
+                    DataObject\Service::enrichLayoutDefinition($layout, $object);
```
dvesh3: Is it really required? as enrichment is done later in the block https://github.com/pimcore/pimcore/blob/6d1e4d376f70e263636d25a47edb0fe6853b1051/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php#L1256

weisswurstkanone: good point

PR #6066, opened 2020-03-19
[DOCS] Update 20_Building_Custom_Rest_APIs.md

pdchaudhary:
if REST API is deprecated then routes endpoint also needs to change . because using these routes "/webservice/rest/*" , it is using the deprecated api key method for authentication.

  
  

Resolves #


brusch:
Thank you!

PR #6076, opened 2020-03-22
[PHPStan] Use new generate baseline option

blankse:
With the new version 0.12.18 its easier to generate a new baseline file:
https://github.com/phpstan/phpstan/releases/tag/0.12.18

PR #6077, opened 2020-03-23
[PHPStan] Fix tests without baseline

blankse:
With #6076 I ignored the tests without baseline. With this it should work again


weisswurstkanone:
thanks! just wondering if we should still offer level < 3 jobs in addition just to make it clear that existing prio 1 and 2 issues should be fixed first.

PR #6078, opened 2020-03-23
[docs] [paypal] casting object to array looses data

ultramegatom:
casting the object to an array will loose some nested objects that are needed.


brusch:
Thanks a lot!

PR #6082, opened 2020-03-23
[Asset] JPEG Thumbnails should only use .jpg as file extension, not pjpeg or .jpeg

brusch:
This solves issues e.g. with mime detections by object storages like S3 and probably also für CDNs and other services. 
There is also no reason to not use the standard file extension at all.

### DIFF
--- a/models/Asset/Image/Thumbnail.php
+++ b/models/Asset/Image/Thumbnail.php
@@ -381,7 +381,7 @@ class Thumbnail
 
                     $sourceHtml = '<source ' . array_to_html_attribute_string($sourceTagAttributes) . ' />';
                     if ($isAutoFormat) {
-                        $sourceHtmlWebP = preg_replace(['@(\.)(jpg|png)@', '@(/)(jpeg|png)@'], '$1webp', $sourceHtml);
+                        $sourceHtmlWebP = preg_replace(['@(\.)(jpg|png)( \dx)@', '@(/)(jpeg|png)(")@'], '$1webp$3', $sourceHtml);
                         $html .= "\t" . $sourceHtmlWebP . "\n";
                     }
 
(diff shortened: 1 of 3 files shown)


dvesh3:
this also
https://github.com/pimcore/pimcore/blob/3816b4480486c6e5021f00e033fa391f08c4cd1d/bundles/CoreBundle/Controller/PublicServicesController.php#L84

brusch:
They all need to stay, because the format stays `PJPEG` just the file-extension changes to `.jpg` ;-)

models/Asset/Image/Thumbnail.php
```diff
@@ -381,7 +381,7 @@ public function getHtml($options = [], $removeAttributes = [])
 
                     $sourceHtml = '<source ' . array_to_html_attribute_string($sourceTagAttributes) . ' />';
                     if ($isAutoFormat) {
-                        $sourceHtmlWebP = preg_replace(['@(\.)(pjpeg|png)@', '@(/)(jpeg|png)@'], '$1webp', $sourceHtml);
+                        $sourceHtmlWebP = preg_replace(['@(\.)(jpg|png)@', '@(/)(jpeg|png)@'], '$1webp', $sourceHtml);
```
NiklasBr: This looks like it is a little bit too greedy, if the substring ".png" or ".jpg" happens to appear inside the file name it might replace the incorrect part of the string:

```php
$sourceHtml = "<source src='test.pngenerated.jpg' />";

$result = preg_replace(['@(\.)(jpg|png)@', '@(/)(jpeg|png)@'], '$1webp', $sourceHtml);

print_r($result); // <source src='test.webpenerated.webp' />
```

brusch: Yep makes sense to harden that, thanks for your feedback!

Issue #4647, opened 2019-07-09
[Feature request] Semantic versioning


NiklasBr:
# Feature Request

Fully commit to use [Semantic Versioning](https://semver.org/).

## Feature description

With the introduction of Composer support in 5.x updating is much easier from a technical perspective, but it is still not clear if we can trust minor and patch versions fully, for example #4645 which has some nice changes but is a BC break and appears soon to be ready to merge.

From my perspective it would add a lot of confidence in releases which are tagged as minor and patches if Pimcore promises not to add BC breaking changes without either:

 * Incrementing the major version number, or
 * Adding a compatibility layer until the major version number can be increased, or
 * Waiting with these changes until you feel a new major version is warranted, for example committing these to a 7.x branch while 6.x is the current stable release.

dpfaffenbauer:
+:1

julmot:
Can't believe no one from the Pimcore team took the responsibility for this yet :cry:
https://semver.org/#why-use-semantic-versioning

brusch:
Starting with 7.0 we're going to fully commit to semver and there will be guidelines similar to https://symfony.com/doc/current/contributing/code/bc.html

For the 6.x branch we promise to be as close to semver as possible.

julmot:
👍 👍 👍 Thanks @brusch!

![ezgif-3-d9cd9cc51de0](https://user-images.githubusercontent.com/8310677/80184184-64f82380-860a-11ea-8b6f-c8c2f5536d8d.gif)

PR #6085, opened 2020-03-24
[Asset] Support for all/custom media-queries in thumbnails

### LINKED ISSUE #5201: Image Thumbnails - Media Queries for min-width breakpoints
Image Thumbnails don't follow the mobile first concepts by providing only max-width media queries.
It should be possible to set up min-width media queries for thumbnails too (or in my opinion, technically it should be only the min-width - way to define media queries for thumbnails, but this will obviously break existing ones)

Currently you always have to set the largest variant of the image as default and then create media query variants for smaller devices. The mobile-first approach would be more to first define the smallest thumbnail as default and then create media querys for larger sizes.

Currently it is a bit weird concerning the mobile-first approach: You just start to code your mobile template but have to create a thumbnail definition with at least a large default and a media query size for mobile. You cannot start by setting the default to the small size and later add a larger media query when you develop the desktop template.
Even worse would be if you initially create the mobile size as default and later on add larger sizes. You develop the mobile template and are finished some time. Then you continue to the desktop version: You add a media query size and without knowing the previously defined default size will not be used anymore for small devices because of the max-width approach.

brusch:
Resolves #5201

In the past only `max-width` media queries were supported, now any media query can be defined. 

image

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
@@ -58,12 +58,11 @@ pimcore.settings.thumbnail.item = Class.create({
                 text: t("add_media_query"),
                 iconCls: "pimcore_icon_add",
                 handler: function () {
-                    Ext.MessageBox.prompt("", t("please_enter_the_maximum_viewport_width_in_pixels_allowed_for_this_thumbnail"), function (button, value) {
-                        if (button == "ok" && is_numeric(value)) {
-                            value = value + "w"; // add the width indicator here, to be future-proof
+                    Ext.MessageBox.prompt("", t("enter_media_query"), function (button, value) {
+                        if (button == "ok") {
                             this.addMediaPanel(value, null, true, true);
                         }
-                    }.bind(this));
+                    }.bind(this), null, false, '(min-width: 576px)');
                 }.bind(this)
             }]
         };
@@ -194,6 +193,11 @@ pimcore.settings.thumbnail.item = Class.create({
 
     addMediaPanel: function (name, items, closable, activate) {
 
+        if(name.match(/^\d+w$/)) {
+            // convert legacy syntax to new syntax/name
+            name = '(max-width: ' + name.replace("w", "") + 'px)';
+        }
+
         if (this.medias[name]) {
             return;
         }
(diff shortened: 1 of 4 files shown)


julmot:
@brusch @weisswurstkanone Please note: In this PR in the DataHub repository we've integrated Pimcore's srcset functionality into DataHub: https://github.com/pimcore/data-hub/pull/177. Changes in the Pimcore API may have an direct impact to this API, as it was in this case.

I was debugging our application and found out that the `descriptor` in this PR was no longer `768px` but `(max-width: 768px)`.

Unfortunately the change in this PR was merged with a minor version, instead of a major version, even though it's a breaking change. Most probably you haven't noticed yet, but please see this as an example why we urgently need to follow [semantic versioning](https://github.com/pimcore/pimcore/issues/4647). Also a reason is that Pimcore doesn't have a [changelog](https://github.com/pimcore/pimcore/releases), there are no descriptions in the releases. If this would have been the case then we'd have noticed that.

In the meantime, until v7 is out and semantic versioning is being used, would it be possible to announce this somehow in the DataHub repository, when things change due to Pimcore updates?

brusch:
I haven't seen this as a breaking change, because it was designed to be fully compatible, so I merged it. Imho this is really an edge case that could even happen when strictly using semver.

julmot:
Not sure why this was classified as non breaking change. I know this PR includes a migration, but I don't think the question is "is a migration included?". It should rather be "Would it be possible that code of users depend on this certain thing directly and could it be that it breaks their code?". If the answer is yes, which it is in this case, it should be classified as breaking change. In this case this user is DataHub/Pimcore itself, but custom implementations on top of Pimcore are also valid use cases that should not be ignored in my opinion.

A quote from semantic versioning:

> If even the tiniest backwards incompatible changes to the public API require a major version bump, won’t I end up at version 42.0.0 very rapidly?

> This is a question of responsible development and foresight. Incompatible changes should not be introduced lightly to software that has a lot of dependent code. The cost that must be incurred to upgrade can be significant. Having to bump major versions to release incompatible changes means you’ll think through the impact of your changes, and evaluate the cost/benefit ratio involved.

I think if it can not be guaranteed that you follow semver it should at least be documented in the release description, otherwise there's no chance for us users to know what changed, without going to all of the git commits manually. If it's documented and we know you will not follow semver, we can stick to a hardcoded version and no longer use e.g. Caret Version Range for this repository.

brankko:
The perfect opportunity to start with release notes :)

PR #6086, opened 2020-03-24
[Asset] Thumbnails: add configuration for 1x1 image placeholder, where no image is wanted

### LINKED ISSUE #217: Image Thumbnails Media Queries - no Image
It should be possible to specify no image / thumbnail for a certain media query, 
eg. a teaser element has no image in mobile view.

brusch:
Resolves #217

image

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
@@ -1351,7 +1351,7 @@ pimcore.settings.thumbnail.items = {
         return item;
     },
 
-    item1x1Placeholder: function (panel, data, getName) {
+    item1x1_pixel: function (panel, data, getName) {
 
         var niceName = t("1x1_pixel_placeholder");
         if (typeof getName != "undefined" && getName) {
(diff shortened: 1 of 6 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
```diff
@@ -1348,6 +1348,35 @@ pimcore.settings.thumbnail.items = {
...
+            id: myId,
+            style: "margin-top: 10px",
+            border: true,
+            bodyStyle: "padding: 10px;",
+            tbar: this.getTopBar(niceName, myId, panel),
+            html: t("1x1_pixel_placeholder_description"),
+            items: [{
+                xtype: "hidden",
+                name: "type",
+                value: "1x1_pixel"
```
dvesh3: value should be "1x1Placeholder" to match with function name to load save config. https://github.com/pimcore/pimcore/blob/bfcce7ca94c08f30fbca69831a13be574c7613d9/bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js#L263

brusch: Good catch, thanks a lot!

brusch: With 8b00edbccb31c90027a42bc7e7354afc6bfcc150 it should be fine now.

dvesh3:
when you set 1x1 pixel placeholder without any media query then it takes height and width, which is wrong:
MicrosoftTeams-image

Issue #6091, opened 2020-03-25
Import and Export CSV Configurations


MarcoGuiducci:
# Feature Request
Possibility to Import and Export the configurations created for CSV Imports

## Feature description
In Pimcore we can create a lot of CSV Column configurations, save and share those configurations between users, load them in a second moment and also import the export configurations.

The problem is that if I have more environments for the same Pimcore project (such as develop, test and production) I am not able to export and import those configuration through these environments. The only way is to extract them from the database.

I think that let people export and import the json configuration (as we can do for classes definitions) can save a lot of time, because in real scenarios CSV import could be very complex.

### Pull Request
#6092

weisswurstkanone:
@MarcoGuiducci , why did you close this one?

MarcoGuiducci:
> @MarcoGuiducci , why did you close this one?

I have closed because I've seen that an issue was opened when the Pull Request was created, so this issue is just a duplicate. Should I reopen it?

weisswurstkanone:
![image](https://user-images.githubusercontent.com/6142086/77924806-7288ea80-72a4-11ea-9cb7-e2c95c177540.png)

how do I export the current configuration?

MarcoGuiducci:
The export functionality becomes enabled as soon as you load a previously saved configuration in the database.

weisswurstkanone:
Can you post a video showing the whole procedure?

We would love to see this in the next release which is due tomorrow.

MarcoGuiducci:
That's a link on WeTransfer with videos of the Export and Import procedures

https://we.tl/t-P7ntYiz9aD

weisswurstkanone:
Thanks for sharing. Export.avi doesn't work for me. The export button stays gray.

Apart from that: not sure if this can confuse the user. So, right now you have to save the config. Then load it again so that you can export it as JSON ?

MarcoGuiducci:
> Thanks for sharing. Export.avi doesn't work for me. The export button stays gray.

I'll check fort this problem

> Apart from that: not sure if this can confuse the user. So, right now you have to save the config. Then load it again so that you can export it as JSON ?

Yes, but I think that just after saving the configuration the export has to be enabled, without reloading. I'll make sure that this happens

weisswurstkanone:
imho, export should be enabled all the time and even contain changes you do in the meantime. if latter one is not possible then at least show a warning that you have to save it first and prevent the export.
[...]

PR #6092, opened 2020-03-25
Import export csv import configurations

MarcoGuiducci:
Possibility to Import and Export the configurations created for CSV Imports

## Additional info  
Related Issue:
#6091


weisswurstkanone:
please revert it.json
en.json is the only file needed . The rest is done in POEditor.
see https://poeditor.com/join/project/VWmZyvFVMH

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2488,4 +2488,168 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
...
+            throw new \Exception("Please provide a valid import configuration Id");
+        }
+
+        try {
+            $importConfig = ImportConfig::getById($importConfigId);
+        
+            $configData = $importConfig->getConfig();
+            $configName = $importConfig->getName();
+            
+            $response = new Response();
```
weisswurstkanone: better use a JsonResponse. Example: 
        $jsonResponse = new JsonResponse($json, 200, [
            'Content-Disposition' => 'attachment; filename="export-data-object-' . $object->getId() . '.json"'
        ], true);

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2488,4 +2488,168 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
...
+        $selectedGridColumns = $configData['selectedGridColumns'];
+        $resolverSettings = $configData['resolverSettings'];
+        $shareSettings = $configData['shareSettings'];
+        $dialect = json_decode(json_encode($configData['csvSettings']), false);
+        
+        $success = true;
+        $supportedFieldTypes = ['checkbox', 'country', 'date', 'datetime', 'href', 'image', 'input', 'language', 'table', 'multiselect', 'numeric', 'password', 'select', 'slider', 'textarea', 'wysiwyg', 'objects', 'multihref', 'geopoint', 'geopolygon', 'geopolyline', 'geobounds', 'link', 'user', 'email', 'gender', 'firstname', 'lastname', 'newsletterActive', 'newsletterConfirmed', 'countrymultiselect', 'objectsMetadata'];
+
+        $classId = $request->get('classId');
+        $file = PIMCORE_SYSTEM_TEMP_DIRECTORY . '/import_' . $request->get('importId');
```
weisswurstkanone: why are we dealing with csv files suddenly? my understanding is that we want to import the exported JSON configuration file?

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2488,4 +2488,168 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
...
+
+        $configData = json_decode($json, true);
+
+        $selectedGridColumns = $configData['selectedGridColumns'];
+        $resolverSettings = $configData['resolverSettings'];
+        $shareSettings = $configData['shareSettings'];
+        $dialect = json_decode(json_encode($configData['csvSettings']), false);
+        
+        $success = true;
+        $supportedFieldTypes = ['checkbox', 'country', 'date', 'datetime', 'href', 'image', 'input', 'language', 'table', 'multiselect', 'numeric', 'password', 'select', 'slider', 'textarea', 'wysiwyg', 'objects', 'multihref', 'geopoint', 'geopolygon', 'geopolyline', 'geobounds', 'link', 'user', 'email', 'gender', 'firstname', 'lastname', 'newsletterActive', 'newsletterConfirmed', 'countrymultiselect', 'objectsMetadata'];
```
weisswurstkanone: see question below but this shouldn't be hardcoded anyway. hard to maintain and even impossible to add custom datatypes.

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
MarcoGuiducci: Done

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
MarcoGuiducci: My idea was to produce the same JSON response of the importGetFileInfoAction function.
The difference is that the json configuration is taken from the uploaded file instead of being loaded from database

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
MarcoGuiducci: As said in the previous comment, this is part of the code of the importGetFileInfoAction. How can I modify to avoid this code duplication?

weisswurstkanone: well, importGetFileAction is about importing CSV data. Here we are talking about importing JSON configurations. Maybe take a look at 
https://github.com/pimcore/pimcore/blob/663b0887cf4623b5499987e2b15b9522f3342e11/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php#L460
for another sample as well.

I guess you don't need the whole magic.

MarcoGuiducci: My idea was to use the imported json to refresh the CSV import interface the same way it's done when a saved configuration is loaded. I will follow your suggestion and try to make this more simple

weisswurstkanone: Thank you! 👍

MarcoGuiducci: Made some changes.
As i wrote in the code's comments, I have maintained the reload of data form original CSV to properly refresh the data preview on the import interface

bundles/CoreBundle/Resources/translations/en.json
```diff
@@ -942,5 +942,7 @@
   "domain_label_width": "Domain label width",
   "redirect_performance_warning": "Using this feature has very bad effects on the performance of the entire application, do you really want to continue?",
   "Uncategorized": "Uncategorized",
-  "methods": "Methods"
+  "methods": "Methods",
+  "export_configuration": "Export Configuration",
```
weisswurstkanone: maybe mention JSON here

something "export as json" and "import from json" just to avoid confusion with the other import

bundles/CoreBundle/Resources/translations/it.json
```diff
@@ -940,4 +940,4 @@
     "test_url": "Test URL",
     "create_redirects": "Create redirects to keep URLs working (incl. children)?",
     "auto_create": "Auto create"
-}
```
weisswurstkanone: please remove entirely if possible

weisswurstkanone:
can you resolve the conflict please?

in addition, please mention this feature in the docs
https://pimcore.com/docs/6.x/User_Documentation/DataObjects/CSV_Import/index.html

bundles/CoreBundle/Resources/translations/en.json
MarcoGuiducci: Done, thanks for the suggestion

bundles/CoreBundle/Resources/translations/it.json
MarcoGuiducci: I'm trying to revert this, but the blank line still remain

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2489,4 +2489,119 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
...
+     *
+     * @param Request $request
+     * 
+     * @return Response
+     */
+    public function exportJsonConfigAction(Request $request)
+    {
+        $importConfigId = $request->get('importConfigId');
+        
+        if($importConfigId == null || empty($importConfigId)){
```
NiklasBr: `if (!isset($importConfigId))` should be enough here because `isset(null)` is `false`

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2489,4 +2489,119 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
...
+                foreach ($rowData as $key => $value) {
+                    $tmpData['field_' . $key] = $value;
+                }
+
+                $tmpData['rowId'] = $count + 1;
+                $data[] = $tmpData;
+
+                $count++;
+
+                if ($count > 18) {
```
NiklasBr: `18` is a magic number of some sort, would be nice if it was described.

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
MarcoGuiducci: Done. Thanks for the suggestion

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
MarcoGuiducci: I have added a comment, but this was taken from originale Pimcore's code

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2489,4 +2489,122 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
...
+        
+        $success = true;
+        $classId = $request->get('classId');
+        
+        $file = PIMCORE_SYSTEM_TEMP_DIRECTORY . '/import_' . $request->get('importId');
+
+        $originalFile = $file . '_original';
+        // determine type
+        if (empty($dialect)) {
+            $dialect = Tool\Admin::determineCsvDialect($file . '_original');
```
NiklasBr: You have `$originalFile` above, it can be re-used rather than concatenating again.

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2500,4 +2500,121 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
...
+                 * Reached the number or rows for the preview
+                 */
+                if ($count > 18) {
+                    break;
+                }
+            }
+            fclose($handle);
+        }
+        
+        $availableConfigs = $this->getImportConfigs($importService, $this->getAdminUser(), $classId);
```
weisswurstkanone: why do we use the admin user here ? shouldn't it be the user that triggers the import action ?

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2500,4 +2500,121 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
...
+
+        return $this->adminJson([
+            'success' => $success,
+            'config' => [
+                'importConfigId' => $importConfigId,
```
weisswurstkanone: where does this importConfigId come from? I guess when yu import it from JSON it shouldn't be set. right now, would it be possible to overwrite a config I don't even own?

bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/import/configDialog.js
```diff
@@ -125,6 +125,14 @@ pimcore.object.helpers.import.configDialog = Class.create({
 
         buttons.push(this.deleteButton);
 
+        this.exportConfigButton = new Ext.menu.Item({
```
weisswurstkanone: From a user experience perspective I believe that the "Load" split button is not the right place for exporting something. Maybe rephrase the entire menu? Or even combine "Load" and "Save" split button ?

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2500,4 +2500,121 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
...
+    
+    /**
+     * @Route("/import-json-config", methods={"POST"})
+     *
+     * @param Request $request
+     * @param ImportService $importService
+     *
+     * @return JsonResponse
+     */
+    public function importJsonConfigAction(Request $request, ImportService $importService)
```
weisswurstkanone: maybe change this to
importCSVImportConfigFromJsonAction or something simpler (same for route).

Currently it is not clear what this is all about :-)

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
weisswurstkanone: we could also try to share identical code. see importGetFileInfoAction for example (same controller)

MarcoGuiducci: the importConfigId comes from the requests, it corresponds to the current loaded configuration, it's not overridden by the imported JSON

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
MarcoGuiducci: I'm trying to use the $this->getUser() function, but i got this exception.

Exception

The exception is thrown on the $importService->getSharedImportConfigs($user, $classId) function.

Which is the correct function to call to get the current user?

weisswurstkanone: \Pimcore\Tool\Admin::getCurrentUser() I believe

MarcoGuiducci: Thank you, fixed

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
MarcoGuiducci: > we could also try to share identical code. see importGetFileInfoAction for example (same controller)

Done sharing code

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
MarcoGuiducci: route and action renamed ;)

bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/import/configDialog.js
MarcoGuiducci: Sincerely, i'm not very skilled in terms of UX :)
I think that the "Import" button could remain in the Load split button. I can just move the "Export" behind the Save split button.
What do you think about it?

weisswurstkanone: better than now, because you will never search for an export button in a "Load" menu :-)

MarcoGuiducci: Moved the Export functionality on the Save split button

weisswurstkanone:
Resolved all conversations? Ready to be reviewed again ? 🤔

MarcoGuiducci:
> Resolved all conversations? Ready to be reviewed again ? 🤔

Yes, I think that I've done all the changes you suggested :)

bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/import/configDialog.js
```diff
@@ -773,5 +795,17 @@ pimcore.object.helpers.import.configDialog = Class.create({
...
             }.bind(this)
         });
+    },
+    
+    getExportConfigUrl: function(){
+        this.commitEverything();
+        var config = this.prepareSaveData();
+        config = Ext.encode(config);
+        
+        return "/admin/object-helper/export-json-config?classId="+this.classId+"&config="+config;
```
weisswurstkanone, line 805: same similar to import url

bundles/CoreBundle/Resources/translations/en.json
```diff
@@ -944,6 +944,8 @@
   "redirect_performance_warning": "Using this feature has very bad effects on the performance of the entire application, do you really want to continue?",
   "Uncategorized": "Uncategorized",
   "methods": "Methods",
+  "export_configuration": "Export As JSON",
```
weisswurstkanone, line 947: better Export Configuration as JSON
same for import

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2500,4 +2511,98 @@ protected function processAvailableFieldDefinitions($fds, &$firstOne, &$commonFi
             }
         }
     }
+    
+    /**
+     * @Route("/export-json-config", methods={"GET"})
```
weisswurstkanone, line 2516: same as for import, maybe 
export-csv-import-config-from-json

bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/import/configDialog.js
MarcoGuiducci: Done, renamed as "export-csv-import-config-as-json"

bundles/CoreBundle/Resources/translations/en.json
MarcoGuiducci: Done

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
MarcoGuiducci: Done, renamed as "export-csv-import-config-as-json"

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
weisswurstkanone: this is just a random number, simply for getting the first x number of lines as preview

weisswurstkanone:
@MarcoGuiducci looks good!

One last wish: Could you please check the user documentation (https://pimcore.com/docs/6.x/User_Documentation/DataObjects/CSV_Import/index.html)
and modify/extend it if necessary?

MarcoGuiducci:
> @MarcoGuiducci looks good!
> 
> One last wish: Could you please check the user documentation (https://pimcore.com/docs/6.x/User_Documentation/DataObjects/CSV_Import/index.html)
> and modify/extend it if necessary?

I've done it, but accidentally another PR was opened for that

weisswurstkanone:
Thanks a lot, definitely a useful improvement!

MarcoGuiducci:
> Thanks a lot, definitely a useful improvement!

It was my pleasure!!! Thanks for your suggestions during these days

PR #6096, opened 2020-03-26
[Staticroute] respect base url

dpfaffenbauer:
Respect base-url in static-routes. Use case:

I started developing Frontend Tests for CoreShop, since I need a fixed environment, I created `web/app_test.php` where I force an ENV. Since that is now my base-path for all generated Routes, Pimcore Static Routes should respect that as well.


dpfaffenbauer:
This now also fixes the Absolute Route Generation with non 80/443 ports.

dvesh3:
thanks a lot! 👍

PR #6109, opened 2020-03-27
Pass context when adding an asset via drag and drop to a document

BlackbitNeueMedien:
Passing the document context will trigger the "resolve upload target" event and gives the developer an opportunity to change the target folder.
(In my specific use case, I want to choose a folder based on the sub-site the document belongs to)


brusch:
Nice, thanks a lot! 🙏

PR #6114, opened 2020-03-30
[Video editable] Use imagethumbnail without thumbnail

BlackbitNeueMedien:
1. Upload video asset to Pimcore assets
1. Upload image asset
1. Create **image** thumbnail `abc`
1. Create controller with view
```php
$video = $this->video('video', [
        "imagethumbnail" => "abc"
]);

echo $video;
```
5. Create document, set above controller, assign uploaded asset video to editable and set uploaded image asset as poster image

Generated HTML will be:
```html
<video width="100%" height="300" poster controls="controls" class="pimcore_video">
<source type="video/mp4" src="/path/to/video.mp4">
</video>
```
No poster image is being set. It ignores the `imagethumbnail` config because this currently only works if `thumbnail` config is given, see https://github.com/pimcore/pimcore/blob/916a01bae967ea02d790fc9e7a8c710622c95c33/models/Document/Tag/Video.php#L358-L368

With this PR the HTML output is
```html
<video width="100%" height="300" poster="/var/tmp/image-thumbnails/0/4217/thumb__abc/poster-image.jpeg" controls="controls" class="pimcore_video">
<source type="video/mp4" src="/path/to/video.mp4">
</video>
```


brusch:
@BlackbitNeueMedien Thanks a lot! 👍

PR #6116, opened 2020-03-30
Allows PreviewRequestTrait to discover primcore_preview

NiklasBr:
Not 100% sure if this is the most optimal location for this change, but I'm submitting it like this. This change resolves no issue, but it helps with previewing Snippets and such objects.

### DIFF
--- a/bundles/CoreBundle/EventListener/Traits/PreviewRequestTrait.php
+++ b/bundles/CoreBundle/EventListener/Traits/PreviewRequestTrait.php
@@ -33,6 +33,10 @@ trait PreviewRequestTrait
             return true;
         }
 
+        if ($request->get('pimcore_preview') === 'true') {
+            return true;
+        }
+
         return false;
     }
 }


brusch:
@NiklasBr Thanks for your PR, can you please outline what this change actually improves? 
Thanks a lot! 😊

NiklasBr:
Documents have a regular preview tab in the admin interface, but Snippets does not. But they have the "eye icon - open in new window" kind of preview. 

While *editmode* is nice, it is often not very good as a true preview, by making this change, then using this Trait in the Snippet Controller like this:

```php
// ...
    use PreviewRequestTrait;

    public function defaultAction(Request $request)
    {
        // ...
        $this->view->previewmode = $this->isPreviewRequest($request);
    }
```

And that lets me include some extra CSS when previewing the Snippet:

```twig
{% if previewmode %}
    <style> ... </style>
{% endif %}
```

Just like *editmode*.

brusch:
Makes sense, thanks a lot man! 👍

PR #6117, opened 2020-03-31
[Video editable] Prevent error when neither using imagethumbnail nor thumbnail

BlackbitNeueMedien:
Follow-up of #6114

This PR prevents a null pointer exception when using a video editable without `imagethumbnail` and without `thumbnail`.
```php
$video = $this->video('video');

echo $video;
```
THe error happened in https://github.com/pimcore/pimcore/blob/9d2d3fe29899634c58c57f7801118916eed620af/models/Document/Tag/Video.php#L414


brusch:
Thanks a lot!

PR #6119, opened 2020-03-31
Import/export icon semiotic alignment

### LINKED ISSUE #6098: Import and Export Icons seem to be wrong
We are actually using the pimcore icons for „export“ and „import“, the colored and the non-colored (black & white) ones. The colored icons seem to be inverted the wrong way. The non-colored seem to be right.

The direction of the arrow of the export icon must be outside the element / circle / square.
The direction of the arrow of the import icon must be inside the element / circle / square.

In these lines, you can see that it’s interchanged:

> .pimcore_icon_export -> /bundles/pimcoreadmin/img/flat-color-icons/internal.svg
> .pimcore_icon_import -> /bundles/pimcoreadmin/img/flat-color-icons/external.svg
> 1197 - 1203 /bundles/AdminBundle/Resources/public/css/icons.css

The same is for ".pimcore_nav_icon_export" and ".pimcore_nav_icon_import". (Lines: 406 - 412)

NiklasBr:
Resolves #6098

See my comment in the issue for more info

### DIFF
--- a/bundles/AdminBundle/Resources/public/css/icons.css
+++ b/bundles/AdminBundle/Resources/public/css/icons.css
@@ -404,11 +404,11 @@
 }
 
 .pimcore_nav_icon_export {
-    background: url(/bundles/pimcoreadmin/img/flat-white-icons/internal.svg) center center no-repeat !important;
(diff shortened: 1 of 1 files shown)


brusch:
Thanks a lot! 🙏😊

PR #6137, opened 2020-04-02
Allow LIKE operator on Select data type filter condition

roynilsson:
In admin object grid, filtering on select fields that use an option provider does not work. Select fields that use option providers use a free text search instead of letting you pick from a list of options. Filter type is 'string' and not 'list' and operator is 'like' instead of 'in'. 

In getFilterConditionExt method of the Select class only the '=' operator is allowed. It needs to allow 'LIKE' operator as well for filtering to work when using option provider.

Additionally for filters using the = operator, the produced query was using LIKE operator. This is not needed. Using = as operator in query should be slightly better performance wise.


dvesh3:
thanks a lot!

PR #6145, opened 2020-04-03
Fixed wrong project root on pimcore-install command

skochTC:
Same behaviour as in issue #5495 but this time during installation.

I think the line I added was forgotten?
Otherwise the variable `$projectRoot` makes no sense.


brusch:
Thanks a lot! 
Just made a small change, since `Bootstrap::setProjectRoot();` is then obsolete 😉

PR #6151, opened 2020-04-04
Delete job will be executed right after its add to recycle bin job.

### LINKED ISSUE #6126: Wrong percentage of deleted objects shown while deleting objects from the grid.
While deleting selected/all the objects from the grid, all the objects are firstly added to recycle bin first and then all are Deleted.So no object is actually deleted till the status is 50 %.

If we are deleting a few objects it doesn't matter much as they are deleted very fast.
But if we are deleting 1000s of objects in that case till 50% no object is actually deleted.
try to delete 500 objects and before 50% hit reload.No object is deleted and some are added in the recycle bin.

Screenshot from 2020-03-31 16-09-14

kubaplas:
Resolves #6126
Instead of two arrays for job, we reuse one - now we have a chain of `add, delete, add, delete` jobs.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
+++ b/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
@@ -84,7 +84,6 @@ class ElementControllerBase extends AdminController
         $hasDependency = false;
         $errors = false;
         $deleteJobs = [];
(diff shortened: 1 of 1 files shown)


brusch:
Nice, thanks a lot!

PR #6160, opened 2020-04-06
[DataObject] Fix inconsistent version diffs

weisswurstkanone:
Alternative approach to #5991 

In the long term diffVersions should use the data object's isEqual method instead of comparing the version preview.

This however requires that every data type properly implements this method. Currently only done for 
relational data types and url slug (needed for dirty detection).
Default implementation in Data returns false not really useable currently.


brusch:
Would be fine for me as a temp. solution, but we should definitely create a follow up ticket for v7.

weisswurstkanone:
see #6165 regarding isEqual

PR #6166, opened 2020-04-06
[DataObject] [Slider] Improve display of the slider component

### LINKED ISSUE #5748: [DataObject] [Slider] Improve display of the slider component
When creating a new DataObject with a slider component the value of it is NULL. The Slider is however set to the lowest position that is some numeric value. To set the slider to the lowest value you have to first change the value and drag it back again.

This seems very confusing for customers. They essentially also have no idea what NULL means.
Maybe we should display something like 'No Value Set' instead of NULL and mark the slider differently somehow?

I think we should also introduce a way to reset the slider value back to NULL to enable inheritance again. Similar to the trashcan for the checkbox type.

image

What do you guys think?

weisswurstkanone:
resolves #5748


weisswurstkanone:
image

weisswurstkanone:
done

PR #6168, opened 2020-04-06
- Added new generic event to the object brick update method.

ascheider:
Description:
Currently there is no event in pimcore if you change object brick definition. With this new event it is possible to listen to the object brick changes.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
@@ -1085,6 +1085,12 @@ class ClassController extends AdminController implements EventedControllerInterf
                 $brickDef->setLayoutDefinitions($layout);
             }
 
+            $event = new GenericEvent($this, [
+                'brickDefinition' => $brickDef
+            ]);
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
```diff
@@ -1085,6 +1085,11 @@ public function objectbrickUpdateAction(Request $request)
                 $brickDef->setLayoutDefinitions($layout);
             }
 
+            $event = new GenericEvent($this, [
+                'brickDefinition' => $brickDef
+            ]);
+            \Pimcore::getEventDispatcher()->dispatch(AdminEvents::CLASS_OBJECTBRICK_UPDATE_DEFINITION, $event);
+
```
dvesh3, line 1092: fetch `brickDefinition` argument from event before save for the changes to be utilized. i.e. https://github.com/pimcore/pimcore/blob/8f09d1164447af032dc3fa972b8cdb58c9d2146b/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php#L189

ascheider: @dvesh3 Hey Divesh, you are right. PR-Updated. Should be fixed now.

dvesh3:
thanks a lot @ascheider 👍

PR #6170, opened 2020-04-06
[Ecommerce][FilterService] Category filter is resetting after page re…

### LINKED ISSUE #5996: [Ecommerce][FilterService] Category filter is resetting after page reload
# Bug Report

Hello.

We have also the same problem on Pimcore Demo. I will try to explain. If we do search like "red" (for example), we will see all results for this term:

Bildschirmfoto 2020-03-06 um 14 15 27

If we refresh the page, the category filter is gone, but results still same:

Bildschirmfoto 2020-03-06 um 14 15 41

We can try to search results for "mercedes" and filter is still gone:

Bildschirmfoto 2020-03-06 um 14 16 00

I have tried to clear pimcore cache and it works but only for one time.

weisswurstkanone:
may fix #5996

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
+++ b/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
@@ -386,13 +386,13 @@ abstract class AbstractRelations extends Data implements
      *
      * @throws \Exception
      */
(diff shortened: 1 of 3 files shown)


dvesh3:
closed in favour of #6176

PR #6171, opened 2020-04-06
Add kernel bootstrap to codeception boostrap file example in documentation

Yivan:
Hello,

This pull request add a missing kernel bootstrap in documentation for codeception.
I was working with codeception 4 and Pimcore 6.6.0, and without this, it was not possible to use Pimcore API on unit tests.

Same for any custom php file needing Pimcore API (php my-script.php), kernel must be run (I think on older Pimcore kernel was bootstraped in `\Pimcore\Bootstrap::bootstrap();` before). I don't know if there is some other places in documentation which need to be updated about this.

Thanks.


brusch:
Thanks!

PR #6174, opened 2020-04-07
[Ecommerce][FilterService] Category filter is resetting after page re…

weisswurstkanone:
…load (alternative approach)

Alternative approach for #5996 
(replaces PR #6170)

Cleaner solution but probably contradicts what we want to accomplish with the website config cache


dvesh3:
closed in favour of #6176

Issue #6177, opened 2020-04-07
[Thumbnail generation] not working since the update to allow various media queries


roland4432:
# Bug Report

### Expected behavior
When setting the thumbnail configuration in backend, I assume the frontend to use the correct image size depending on viewport width

### Actual behavior
Only the default filter will apply to the given image

### Steps to reproduce
Create a thumbnail configuration with:
 * max-width: 1200px
 * max-width: 600px

Create a document with an image and set the newly created thumbnail config.  
Resize your browser to match either max 1200px or 600px.
None of the none default filters will apply.

### Code example
Proof of concept to get it back to working in parts

pimcore/bundles/CoreBundle/Controller/PublicServicesController.php
```php
//line 119-123

// check if a media query thumbnail was requested
if (preg_match("#~\-~([\d\w-]+)#", $matches[1], $mediaQueryResult)) {
    $mediaQueryResult[1] = '(' . preg_replace('/(max|min)(-width)-(\d+px)/', '$1$2: $3', $mediaQueryResult[1]) . ')';
    $thumbnailConfig->selectMedia($mediaQueryResult[1]);
}
```

### Problem with ordering
The next issue: the order of the given mediaqueries in the picture won't apply if not sorted correctly  
a picture srcset with the following order:
```
<source srcset="..." media="(max-width: 1200px")>
<source srcset="..." media="(max-width: 600px")>
```
will *never* create a 600px image, since max-width: 1200px already matches for 1px viewport width.

However if you reorder the code to this:
```
<source srcset="..." media="(max-width: 600px")>
<source srcset="..." media="(max-width: 1200px")>
```
the correct thumbnail will be used.

So for convinience, since you cannot reorder thumbnail media query configs in the backend, you have to follow this:
when you go with _max-width_ queries, you have to start with the smallest filter, ending with the largest  
when you go with _min-width_ queries, you have to start with the largest, ending with the smallest

There is loads of todos with the new system to allow for more complex media queries for image thumbnails since the last update.  

I am eager to help, once this issue is reviewed from @brusch or @fashxp and you guys come to the same conclusion.

brusch:
@roland4432 can you please share an example, I don't get your point, sorry ....

roland4432:
Hello @brusch. Ofc :)

The current media queries don't work. You will get the filtered image from the default config for that specific thumbnail setup for all viewports.

[...]

PR #6188, opened 2020-04-08
[Asset] Thumbnails: Possibility to change order of media queries

brusch:
Reported in #6177

image


Iris57:
@brusch thank you for continuing on this, tested and it works as I expected. I think its a good solution for the problem we encountered in #6177 .

PR #6192, opened 2020-04-08
Fixes Symfony 4.2 deprecation warnings in TreeBuilder

NiklasBr:
See https://symfony.com/blog/new-in-symfony-4-2-important-deprecations#deprecated-tree-builders-without-root-nodes

Might have to be tested with Symfony 3.x just to be sure nothing breaks.


NiklasBr:
So, it seems it broke in the test, I'll leave the `root()` calls in but move the argument

PR #6194, opened 2020-04-08
Avoid deadlocks caused by object_relations table

weisswurstkanone:
see #5761 

primary keys has already been dropped in the table's create statement but was never removed via migration script


brusch:
See also #5769 and #5687

brusch:
@weisswurstkanone the index is getting dropped here https://github.com/pimcore/pimcore/blob/f5b02e99a581eaeef90b4122de94357a5d873d18/bundles/CoreBundle/Migrations/Version20200211115044.php#L25 so I don't think this migration is necessary, is it?

weisswurstkanone:
Ah right, then this one is obsolete.

PR #6195, opened 2020-04-08
[QuantityValue] Value should be a float

blankse:
Data from the database is a string. Should be casted to float.

quantity-value


brusch:
Makes sense, thanks a lot! 👍

PR #6198, opened 2020-04-09
Refresh of custom-view trees, too

baloraki:
The update of the trees currently only affects the standard trees.

## Resolves
This change should also cover custom views.


weisswurstkanone:
Hi @baloraki , I created a PR with a different approach as there is already a helper covering some aspects of tree handling.
see https://github.com/pimcore/pimcore/pull/6271

Issue #6199, opened 2020-04-09
Full-Page-Cache: Wrong links in multi site system


blankse:
# Bug Report

In a system with multiple sites I can't activate Full-Page-Cache. Because the links a often wrong. I think the backend previews have the same cache as the frontend.

In the backend the links are with the complete system uri. 
In the frontend, they should be shortened without the site uri.

brusch:
@blankse can you describe the problem a bit more in detail? 
Or even provide a fix, the full page cache is pretty simple, so maybe even quicker 😉
https://github.com/pimcore/pimcore/blob/23e5ff63213f688563101087ad6a78981f11d5ed/bundles/CoreBundle/EventListener/Frontend/FullPageCacheListener.php#L36

blankse:
@brusch I am in debugging in the moment and try to make a fix.

I think its not the problem of links in the current document but the links in snippets. The Snippet Tag has its own cache, when Full-Page-Cache is activated.

I can reproduce it:
- Clear Full-Page-Cache
- Open Site in Frontend (Links ok)
- Clear Full-Page-Cache
- Open Site in Backend
- Open Site in Frontend (Links wrong)

blankse:
@brusch I created a PR

PR #6200, opened 2020-04-09
[Full-Page-Cache] Fix: Wrong links in multi site system

blankse:
Resolves https://github.com/pimcore/pimcore/issues/6199

### DIFF
--- a/models/Document/Tag/Snippet.php
+++ b/models/Document/Tag/Snippet.php
@@ -18,10 +18,11 @@
 namespace Pimcore\Model\Document\Tag;
 
 use Pimcore\Cache;
-use Pimcore\Logger;
 use Pimcore\Model;
 use Pimcore\Model\Document;
+use Pimcore\Model\Site;
 use Pimcore\Targeting\Document\DocumentTargetingConfigurator;
+use Pimcore\Tool;
 use Pimcore\Tool\DeviceDetector;
 use Pimcore\Tool\Frontend;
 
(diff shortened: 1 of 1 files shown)


blankse:
I add isFrontend/isSiteRequest to the Snippet cache keys. This params are the indicators in the getPath method to switch between backend/frontend links.

I also remove the try/catch. This should be only part of the __toString() method. With the render() method it should be handled by symfony. See https://github.com/pimcore/pimcore/pull/6042 and https://github.com/pimcore/pimcore/pull/4996

models/Document/Tag/Snippet.php
```diff
@@ -114,71 +115,61 @@ public function frontend()
...
+            if ($this->snippet->getUseTargetGroup()) {
+                $cacheParams['target_group'] = $this->snippet->getUseTargetGroup();
             }
 
-            return $content;
-        } catch (\Exception $e) {
-            Logger::error($e);
+            $cacheParams['webp'] = Frontend::hasWebpSupport();
+            $cacheParams['frontend'] = Tool::isFrontend();
+            $cacheParams['siteRequest'] = Site::isSiteRequest();
```
brusch, line 151: I think we should better use the ID of the site here, then `isFrontend()` shouldn't be necessary.

blankse: I changed it

blankse:
@brusch Are there still any doubts?

brusch:
@blankse Sorry, haven't got the time yet for a detailed review and testing.

brusch:
Looks good, thanks a lot!

PR #6203, opened 2020-04-10
[Routing] Locale not set consistently (Translator, ...) when fallback document is used

brusch:
Regression of #5905


MercuryKojo:
This only fixes the issue partially.
The Translator now has the correct language, but Carbon and the path Helper still don't use the correct locale, because the request does not have the "_locale" attribute when the symfony LocaleListener is called, so it is always set to the default locale
@brusch

PR #6204, opened 2020-04-10
Field Lookup for Objects

### LINKED ISSUE #5277: Field Lookup for Objects
It would be great if we can have the "Field Lookup" Plugin directly in the core. In various projects it is really handy when you have large product classes.... to easy find the field location or its value.

image

weisswurstkanone:
resolves #5277


bundles/AdminBundle/Resources/public/js/pimcore/object/fieldlookup/helper.js
```diff
@@ -0,0 +1,248 @@
...
+                if (currentFilterValue) {
+                    var textLower = text.toLowerCase();
+                    var idx = textLower.indexOf(currentFilterValue);
+                    if (idx == -1) {
+                        return null;
+                    }
+                    var pre = text.substring(0, idx);
+                    var post = text.substring(idx + currentFilterValue.length);
+                    var match = text.substring(idx, idx + currentFilterValue.length);
+                    text = pre + "<font color='#00AA00'><b color='green'>" + match + "</b></font>" + post;
```
NiklasBr, line 144: `<font color='#00AA00'><b color='green'>` What? :D

weisswurstkanone: This is 1:1 from the existing bundle (it. highlights the filter match). Suggestions for better styling are welcome, either as part of this PR or afterwards :-)

NiklasBr: What about `<strong style="color: #00AA00">` considering the state of [\<font\>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font).

weisswurstkanone: changed it but I am wondering if we should replace this greenish highlight color with something else that causes less pain :-)

NiklasBr: What about  \#008040 (https://jsfiddle.net/10yogmak/1/)? Should hit most requirements for [color contrast](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrasts).

I first tried \#28a050 (the background colour on the Save & Publish button), but that one had a little too low contrast versus light colours at (3.37, 2.48, 5.61).

weisswurstkanone: thx! looks definitely better

image

weisswurstkanone:
Changed this in the same moment :-)

PR #6207, opened 2020-04-10
Search and replace pagination

t-zilla:
Results grid in File -> Search and Replace Assignments now supports pagination.

### DIFF
--- a/models/Dependency/Dao.php
+++ b/models/Dependency/Dao.php
@@ -21,6 +21,7 @@ use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use Pimcore\Logger;
 use Pimcore\Model;
 use Pimcore\Model\Element;
+use Symfony\Component\HttpFoundation\Exception\SuspiciousOperationException;
 
 /**
  * @property \Pimcore\Model\Dependency $model
@@ -176,7 +177,7 @@ class Dao extends Model\Dao\AbstractDao
     {
         $query = 'SELECT * FROM dependencies WHERE targetid = ? AND targettype = ?';
 
-        if ($offset !== null & $limit !== null) {
+        if ($offset !== null && $limit !== null) {
             $query = sprintf($query . ' LIMIT %d,%d', $offset, $limit);
         }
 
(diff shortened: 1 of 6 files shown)


kubaplas:
@t-zilla Getting a list of objects containing relation should be rewritten in a way that would limit the amount of objects using db query. Now it still collects the data about all of the relations, still resulting in communication failure.

dvesh3:
@t-zilla could you please address this concern? https://github.com/pimcore/pimcore/pull/6207#issuecomment-614097816

models/Dependency/Dao.php
```diff
@@ -196,6 +197,64 @@ public function getRequiredBy($offset = null, $limit = null)
...
+        if (!in_array($orderBy, ['id', 'type', 'path'])) {
+            $orderBy = 'id';
+        }
+
+        if (!in_array($orderDirection, ['ASC', 'DESC'])) {
+            $orderDirection = 'ASC';
+        }
+
+        $query = "
+            SELECT id, type, path
```
dvesh3, line 226: @t-zilla can you please explain what is the benefit of using such a complex query over a simple one in `getRequiredBy()`? AFAIK we still need to fetch the relation elements to check permissions and so.. or i missed a point here?

t-zilla: To enable pagination without a need to fetch _all_ dependencies which causes a communication error down the line. The simple query on `dependencies` table doesn't enable us to sort by `path` which appears on the S&R grid.

dvesh3: Thanks for your explanation! I believe we can keep it then as it is just used in S&R grid.

models/Dependency/Dao.php
```diff
@@ -196,6 +197,64 @@ public function getRequiredBy($offset = null, $limit = null)
...
     }
 
+    /**
+     * @param null $orderBy
+     * @param null $orderDirection
+     * @param null $offset
+     * @param null $limit
+     * @return array
+     */
+    public function getRequiredByWithPath($orderBy = null, $orderDirection = null, $offset = null, $limit = null)
```
dvesh3, line 207: I would still keep offset and limit params in first place to be consistent with `getRequiredBy()`

t-zilla: Done

dvesh3:
@t-zilla please revert all translation files except en.json. Translations (except English) are managed in POEditor: https://pimcore.com/en/resources/translations

models/Dependency/Dao.php
```diff
@@ -196,6 +197,64 @@ public function getRequiredBy($offset = null, $limit = null)
         return $requiredBy;
     }
 
+    /**
+     * @param null $orderBy
```
dvesh3, line 201: always null?

dvesh3:
@t-zilla nice work! thanks a lot 👍

PR #6209, opened 2020-04-10
Fix: New Request() in DocumentRenderer clears RequestContext Data

### LINKED ISSUE #3579: New Request() in DocumentRenderer clears RequestContext Data
https://github.com/pimcore/pimcore/blob/4c986d60d890898e466533a21dfd29bde2acb9f6/lib/Document/Renderer/DocumentRenderer.php#L129

As you can see here, pimcore generates a new request object if the request stack has no active requests (in cli mode for example).

That's ok if you're submitting just one email per request. But if you're going to send multiple mails (or something else with an involved request context) via CLI in a loop, this becomes a major problem. For example:

## Preparation
- Populate system settings `host` with `my-domain.com`
- Create a email
- Create a command with a containing loop (more than one cycles) and send the mail on each cycle:

```php
foreach ($infos as $info) {
   $email = Email::getByPath('/your-email');
   $url = $this->urlGenerator->generate('route_name', []);
   $mail = new Mail();
   $mail->setDocument($email);
   $mail->setTo('your-email-address');
   $mail->setParam('url', $url);
   $mail->send();
}
```

## Result
- Mail 1: Param `url` is `http://my-domain.com/my-url-structure`
- Mail 2: Param `url` is `http://localhost/my-url-structure`

## Explanation
1. The first mail generation is valid since the default RequestContext contains the default host (search for`pimcore.routing.router.request_context` service in your compiled container).
2. DocumentRenderer does not find any request and creates a new one
3. Empty request will be injected in `RouteListener` => `$this->context->fromRequest($request);`
4. In the next cycle, the context does not have any default data anymore.

dvesh3:
Resolves #3579

### DIFF
--- a/lib/Http/RequestHelper.php
+++ b/lib/Http/RequestHelper.php
@@ -16,6 +16,7 @@ namespace Pimcore\Http;
 
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\Routing\RequestContext;
 
 class RequestHelper
 {
@@ -26,12 +27,16 @@ class RequestHelper
      */
     protected $requestStack;
 
+    protected $requestContent;
+
     /**
      * @param RequestStack $requestStack
+     * @param RequestContext $requestContext
      */
-    public function __construct(RequestStack $requestStack)
+    public function __construct(RequestStack $requestStack, RequestContext $requestContext)
     {
         $this->requestStack = $requestStack;
+        $this->requestContext = $requestContext;
     }
 
     public function hasCurrentRequest(): bool
@@ -186,5 +191,23 @@ class RequestHelper
         $aip .= '255';
 
         return $aip;
+    }
+
+    /**
+     * @param string $uri
+     * @return Request
+     */
+    public function createRequestWithContext($uri = '/') {
+        $request =  Request::create(
+            $this->requestContext->getScheme().'://'.$this->requestContext->getHost().$this->requestContext->getBaseUrl().$uri,
+            $this->requestContext->getMethod(),
+            $this->requestContext->getParameters(), [], [],
+            [
+            'SCRIPT_FILENAME' => $this->requestContext->getBaseUrl(),
+            'SCRIPT_NAME' => $this->requestContext->getBaseUrl(),
+            ]
+        );
+
+        return $request;
     }
 }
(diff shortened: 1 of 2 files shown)


lib/Http/RequestHelper.php
```diff
@@ -187,4 +195,22 @@ public function anonymizeIp(string $ip)
...
         return $aip;
     }
+
+    /**
+     * @param string $uri
+     * @return Request
+     */
+    public function createRequestWithContext($uri = '/') {
+        $request =  Request::create(
+            $this->requestContext->getScheme().'://'.$this->requestContext->getHost().$this->requestContext->getBaseUrl().$uri,
```
brusch, line 205: We should consider non-standard ports here: 

```php
if ('http' === $scheme && 80 !== $this->requestContext->getHttpPort()) {
    $port = ':'.$this->requestContext->getHttpPort();
} elseif ('https' === $scheme && 443 !== $this->requestContext->getHttpsPort()) {
    $port = ':'.$this->requestContext->getHttpsPort();
}
```

lib/Http/RequestHelper.php
```diff
@@ -187,4 +195,22 @@ public function anonymizeIp(string $ip)
 
         return $aip;
     }
+
+    /**
+     * @param string $uri
```
brusch, line 200: Please mark as `@internal`

lib/Http/RequestHelper.php
```diff
@@ -187,4 +195,22 @@ public function anonymizeIp(string $ip)
...
+     * @return Request
+     */
+    public function createRequestWithContext($uri = '/') {
+        $request =  Request::create(
+            $this->requestContext->getScheme().'://'.$this->requestContext->getHost().$this->requestContext->getBaseUrl().$uri,
+            $this->requestContext->getMethod(),
+            $this->requestContext->getParameters(), [], [],
+            [
+            'SCRIPT_FILENAME' => $this->requestContext->getBaseUrl(),
```
brusch, line 209: Is it really necessary to set `SCRIPT_FILENAME` and `SCRIPT_NAME` ? 
If so, we should probably set them to the right values, right?

lib/Http/RequestHelper.php
dvesh3: done!

lib/Http/RequestHelper.php
dvesh3: done!

lib/Http/RequestHelper.php
dvesh3: you are right, it is not required. so removed it.

PR #6212, opened 2020-04-14
[Redirects] Fetch target site by domain if no target site set

BlackbitNeueMedien:
When you create a redirect without target site currently the general domain from the system settings gets used. This PR extends this by trying to find the target site by the request's hostname if no target site has been specified for the redirect. If none is found, the general domain gets used (as before)

### DIFF
--- a/lib/Routing/RedirectHandler.php
+++ b/lib/Routing/RedirectHandler.php
@@ -190,16 +190,11 @@ class RedirectHandler implements LoggerAwareInterface
                     return null;
                 }
             } else {
-                $redirectDomain = $this->config['general']['domain'];
-                $sites = new Site\Listing();
-                foreach ($sites as $site) {
-                    $siteDomains = $site->getDomains();
-                    $siteDomains[] = $site->getMainDomain();
-                    $siteIndex = \array_searchi($request->getHost(), $siteDomains);
-                    if($siteIndex !== false) {
-                        $redirectDomain = $siteDomains[$siteIndex];
-                        break;
-                    }
+                $site = Site::getByDomain($request->getHost());
+                if($site instanceof Site) {
+                    $redirectDomain = $request->getHost();
+                } else {
+                    $redirectDomain = $this->config['general']['domain'];
                 }
 
                 if ($redirectDomain) {
--- a/lib/Routing/RedirectHandler.php
+++ b/lib/Routing/RedirectHandler.php
@@ -174,24 +174,40 @@ class RedirectHandler implements LoggerAwareInterface
             $url = replace_pcre_backreferences($url, $matches);
         }
 
-        if ($redirect->getTargetSite() && !preg_match('@http(s)?://@i', $url)) {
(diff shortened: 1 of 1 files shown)


lib/Routing/RedirectHandler.php
```diff
@@ -174,25 +174,41 @@ protected function buildRedirectResponse(Redirect $redirect, Request $request, $
...
             } else {
-                $this->logger->error('Site with ID {targetSite} not found', [
-                    'redirect' => $redirect->getId(),
-                    'targetSite' => $redirect->getTargetSite()
-                ]);
+                $redirectDomain = $this->config['general']['domain'];
+                $sites = new Site\Listing();
+                foreach ($sites as $site) {
+                    foreach ($site->getDomains() as $siteDomain) {
+                        if ($request->getHost() === $siteDomain) {
```
dvesh3, line 197: why not consider site main domain also for comparison? and better use something like array_search() instead of second foreach

BlackbitNeueMedien: Done. I use `array_searchi()` as domain names are not case-sensitive.

lib/Routing/RedirectHandler.php
```diff
@@ -174,25 +174,42 @@ protected function buildRedirectResponse(Redirect $redirect, Request $request, $
...
+
+                    return null;
+                }
             } else {
-                $this->logger->error('Site with ID {targetSite} not found', [
-                    'redirect' => $redirect->getId(),
-                    'targetSite' => $redirect->getTargetSite()
-                ]);
+                $redirectDomain = $this->config['general']['domain'];
+                $sites = new Site\Listing();
```
dvesh3, line 194: I stumbled upon a method to get the site by domain host and i think we utilize it here, so something like this: https://github.com/pimcore/pimcore/blob/2c01ab9c60fd222063cbaf59b290f812fa7fcaa8/bundles/CoreBundle/EventListener/Frontend/FrontendRoutingListener.php#L167

It leads better performance when you have a lot of sites. @BlackbitNeueMedien  WDYT?

BlackbitNeueMedien: Hmm, does `Site::getByDomain()` really work? https://github.com/pimcore/pimcore/blob/004e782803f69a85c5d12bbf68bb932b70b4a747/models/Site/Dao.php#L60-L96

Currently the domains for a site are saved as serialized PHP array, not with the %-encoding which is for example used for relational data object fields.
Bildschirmfoto 2020-04-20 um 14 21 54
So the SQL condition does not work and the code in the foreach loops does only regard wildcard domains.

Do we have to fix `Site\Dao::getByDomain()` or have I overseen something and it actually does work?

dvesh3: I don't see any problem. It is using a LIKE operator so should work for additional domains also
image

image

BlackbitNeueMedien: Yes, you are right, my fault - changed tis PR.

PR #6213, opened 2020-04-14
[Core] Changed default permissions from 0775 to 0664

brusch:


brusch:
This needs some more attention since `.php` files need the execute flag. 
@dvesh3 can you take over please? Thanks 😊

Yivan:
@brusch @dvesh3 thanks a lot for this fix!

Just to know, why php files needs execute bit (for instance you added 775 here https://github.com/pimcore/pimcore/pull/6213/commits/f456d0f4990c8d9d97b47651f28e6d8f2e1ad625) ? I was thinking execute bit is only needed if called directly in cli so the php-cli can handle it directly, like .sh file instance. But calling them `php myfile.php` works without execute bit, same than bash myfile.sh.
So why generated classes, etc. need this +x bit ? I am very curious : )

brusch:
@Yivan @dvesh3 This was actually the issue: 
https://travis-ci.com/github/pimcore/pimcore/jobs/319097645#L2473
My first thought was it's caused by the missing `+x` but could be something else as well, but with 775 it's working fine again 😊

Yivan:
> 
> 
> @Yivan @dvesh3 This was actually the issue:
> https://travis-ci.com/github/pimcore/pimcore/jobs/319097645#L2473
> My first thought was it's caused by the missing `+x` but could be something else as well, but with 775 it's working fine again 😊

Could it come because the parent directory is not created with the +x permissions, so writing file inside is forbidden, just an idea where to investigate.

dvesh3:
@Yivan yes, you are right that it happened because parent directory was missing +x permissions to general files under it so we are working to see if we can revert the file permission changes and just keep changed permission for mkdir() calls. thanks a lot for your direction 😉 

@brusch do you have anything to add here?

brusch:
@dvesh3 no, sounds like a plan 😊 - please update the PR accordingly, thanks a lot!

PR #6222, opened 2020-04-15
Update icons.css for Quick Search

putzflorian:
set priority of pimcore_icon_asset_default otherwise it overrides the icons in quick search

no_icon


brusch:
Thanks a lot! 🙏

PR #6224, opened 2020-04-15
make document tags protected

### LINKED ISSUE #6050: Document\Tag Public properties
Remove Public Properties there as well to be consistent with DataObjects and Assets (https://github.com/pimcore/pimcore/blob/master/models/Document/Tag/Relation.php#L58). 

It is quite painful sometimes to work with that in templates, you always have to remember where to use a getter or the property name itself (which for private/protected properties, symfony checks for a getter and uses that one)

dpfaffenbauer:
resolves #6050 

Make Document Tags Protected with additional getters

### DIFF
--- a/models/Document/Tag/Areablock.php
+++ b/models/Document/Tag/Areablock.php
@@ -35,19 +35,19 @@ class Areablock extends Model\Document\Tag implements BlockInterface
      *
      * @var array
      */
-    public $indices = [];
+    protected $indices = [];
 
     /**
      * Current step of the block while iteration
      *
      * @var int
      */
-    public $current = 0;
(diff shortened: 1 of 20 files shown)


brusch:
Because of #6921 we have quite a lot of conflicts now. @dpfaffenbauer can you please resolve them or just recreate the PR (whatever is easier for you)? Thanks a lot! 🙏

dpfaffenbauer:
I am quite busy at the moment with other things... I'll just close it for now, maybe someone else finds time for this.

brusch:
👍 the issue is still there anyway, so we don't forget 😊 Thanks!

PR #6226, opened 2020-04-15
Import reporting is currently not scrollable.

ascheider:
So if you import many items, you are not able to see the whole result set.
This PR adds possibility to scroll in reporting tab


brusch:
Thanks!

PR #6228, opened 2020-04-15
Remove docblock data which changes everytime

### LINKED ISSUE #6211: Docblocks in generated classes/objectbricks
All generated definition files for classes and objectBricks contain docblocks with "Changed by" and "Generated at" fields.
As mentioned in the Pimcore manual these files should be committed to version control to keep DB structures consistent among DTAP environments. However these two fields obviously cause changes everytime a field is added and cause merge conflicts during GIT merge process.
In my opinion this information has no added value so it should be removed to make the GIT workflow less frustrating. 
We tried by extending/event listeners etc. to modify this behaviour, but this part of Pimcore code is designed in such a bad way this is no option. So the only solution left is to modify the Pimcore core code.
What do you think? Should we remove the two docblock fields or make this configurable somehow? We could provide a PR for this.

bramstroker:
Resolves #6211

As discussed there's not much added value in having this data in the docblocks. The disadvantages (merge conflicts when working with several feature branches) outweigh the advantages, so we remove this.

### DIFF
--- a/models/DataObject/ClassDefinition.php
+++ b/models/DataObject/ClassDefinition.php
@@ -600,10 +600,6 @@ class ClassDefinition extends Model\AbstractModel
         $cd .= '* Inheritance: '.($this->getAllowInherit() ? 'yes' : 'no')."\n";
         $cd .= '* Variants: '.($this->getAllowVariants() ? 'yes' : 'no')."\n";
 
-        if (isset($_SERVER['REMOTE_ADDR'])) {
-            $cd .= '* IP: '.$_SERVER['REMOTE_ADDR']."\n";
-        }
-
         if ($this->getDescription()) {
             $description = str_replace(['/**', '*/', '//'], '', $this->getDescription());
             $description = str_replace("\n", "\n* ", $description);
--- a/models/DataObject/ClassDefinition.php
+++ b/models/DataObject/ClassDefinition.php
@@ -597,14 +597,8 @@ class ClassDefinition extends Model\AbstractModel
 
         $cd .= '/** ';
         $cd .= "\n";
-        $cd .= '* Generated at: '.date('c')."\n";
         $cd .= '* Inheritance: '.($this->getAllowInherit() ? 'yes' : 'no')."\n";
         $cd .= '* Variants: '.($this->getAllowVariants() ? 'yes' : 'no')."\n";
-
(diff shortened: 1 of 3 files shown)


bramstroker:
Also changed it for fieldcollection definition, which is also used for objectbrick classes.

models/DataObject/ClassDefinition.php
```diff
@@ -597,15 +597,9 @@ protected function getInfoDocBlock()
...
         $cd .= '* Inheritance: '.($this->getAllowInherit() ? 'yes' : 'no')."\n";
         $cd .= '* Variants: '.($this->getAllowVariants() ? 'yes' : 'no')."\n";
 
-        $user = Model\User::getById($this->getUserModification());
-        if ($user) {
-            $cd .= '* Changed by: '.$user->getName().' ('.$user->getId().')'."\n";
-        }
-
         if (isset($_SERVER['REMOTE_ADDR'])) {
             $cd .= '* IP: '.$_SERVER['REMOTE_ADDR']."\n";
```
bramstroker, line 604: Another look at the PR, I think we also need to remove the IP. Agreed?

brusch: Yep

bramstroker: Made the changes btw

bramstroker:
> Please remove from CustomLayout also:
> 
> https://github.com/pimcore/pimcore/blob/663b0887cf4623b5499987e2b15b9522f3342e11/models/DataObject/ClassDefinition/CustomLayout.php#L288



> Please remove from CustomLayout also:
> 
> https://github.com/pimcore/pimcore/blob/663b0887cf4623b5499987e2b15b9522f3342e11/models/DataObject/ClassDefinition/CustomLayout.php#L288

Done

models/DataObject/ClassDefinition/CustomLayout.php
```diff
@@ -293,15 +293,6 @@ protected function getInfoDocBlock()
         $cd .= "\n";
         $cd .= '* Generated at: '.date('c')."\n";
```
jdreesen, line 294: Should this be removed, too?

bramstroker: Sharp! And done..

brusch:
Very nice 👌 Thanks a lot!

PR #6231, opened 2020-04-15
Update Tag.php : added Tag::getByName() method

FredoVelcro:
Added useful method : getByName()


weisswurstkanone:
I think you have to combine both pull requests

brusch:
Should be integrated into #6232 - thanks! 😊

PR #6232, opened 2020-04-15
update Tag DAO : added getByName($name) method

FredoVelcro:
Resolves #

### DIFF
--- a/models/Element/Tag/Dao.php
+++ b/models/Element/Tag/Dao.php
@@ -39,6 +39,20 @@ class Dao extends Model\Dao\AbstractDao
         $this->assignVariablesToModel($data);
     }
 
+        /**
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
shouldn't this behave in the same way as Tag::getById(), i.e. return null in case the tag does not exist ?

see https://github.com/pimcore/pimcore/blob/163f95e61ac97b8d5997932827ba93cc85755b0b/models/Element/Tag.php#L65

brusch:
And we need the static getter ... see https://github.com/pimcore/pimcore/blob/163f95e61ac97b8d5997932827ba93cc85755b0b/models/Element/Tag.php#L65

jdreesen:
I think this PR is superseded by #6234

PR #6236, opened 2020-04-16
Cache-Tree is not properly cleared when updating Block-Elements

### LINKED ISSUE #6113: Cache-Tree is not properly cleared when updating Block-Elements
# Bug Report

### Setup:

Following Data-Objects:
- ObjectA
   - ObjectB
       - Blocks
           - BlockElement
- ObjectC
  - Relation to ObjectB

I make changes to the BlockElement
BlockElement->setTitle('title')->save();

### Expected behavior

Saving (no difference, whether i save through web or php-api) the BlockElement, then accessing the values of BlockElement from either ObjectA or ObjectC shows the updated values

ObjectC->getObjectB()->"getBlocks()"->getBlockElement()->getTitle() === 'title';

### Actual behavior

Saving (no difference, whether i save through web or php-api) the BlockElement, then accessing the values of BlockElement from either ObjectA or ObjectC shows the old value, pre-update

ObjectC->getObjectB()->"getBlocks()"->getBlockElement()->getTitle() !== 'title';
ObjectC->getObjectB()->"getBlocks()"->getBlockElement()->getTitle() === 'previousTitle';

### Reason:

I narrowed the saving process down:
- while saving, the BlockElement clears the cache of the Parent
- it does not however "inform" the related objects about this change.

My Workaround so far is, that i get the CacheTags and clear manually:
- ObjectA
- ObjectB
- ObjectC

Doing that manually feels wrong to me, thus here i am opening this issue.
What could be a better way of solving this issue?

weisswurstkanone:
resolves #6113


models/DataObject/Data/BlockElement.php
```diff
@@ -110,11 +129,90 @@ public function setData($data)
...
 
+    protected function renewReferences()
+    {
+        $copier = new DeepCopy();
+        $copier->skipUncloneable(true);
+        $copier->addTypeFilter(
+            new \DeepCopy\TypeFilter\ReplaceFilter(
+                function ($currentValue) {
+                    if ($currentValue instanceof ElementDescriptor) {
+                        $cacheTag = $currentValue->getType() . "_" . $currentValue->getId();
```
brusch, line 140: Should we add a getCacheTag() method to the ElementDescriptor ?

brusch:
@weisswurstkanone can you please tidy up this PR?  Thanks!

weisswurstkanone:
@brusch done

models/Version/ElementDescriptor.php
```diff
@@ -74,4 +76,16 @@ public function setId(int $id)
 
         return $this;
     }
+
+    /**
+     * get the cache tag for the element
+     *
+     * @return string
+     */
+    public function getCacheTag()
```
brusch, line 85: Maybe we should call this `getCacheKey()` , see `\Pimcore\Model\Element\AbstractElement::getCacheKey`
What do you think? I guess this would be a bit more specific to it's use case.

weisswurstkanone:
Mhm, why should be there a difference to AbstractElement::getCacheTag ?

brusch:
@weisswurstkanone actually `getCacheKey()` is used for the cache keys, so `getCacheTag()` could be misleading... see https://github.com/pimcore/pimcore/blob/1f9fca52eacf4fcc8b8c7899b65c0b1900db9124/models/DataObject/AbstractObject.php#L275

weisswurstkanone:
changed it

brusch:
@weisswurstkanone can u plz check the tests?

PR #6237, opened 2020-04-16
Refactors login/lost-pw/2-factor form autocomplete functionality

NiklasBr:
Autocomplete=off on login forms is [not really a thing any more](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#The_autocomplete_attribute_and_login_fields).

Also, added autocomplete hints to improve compatibility with password managers.

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/Login/login.html.php
+++ b/bundles/AdminBundle/Resources/views/Admin/Login/login.html.php
@@ -27,7 +27,7 @@ if ($browser->getBrowser() == \Pimcore\Browser::BROWSER_OPERA && $browserVersion
 
 
 <div id="loginform">
-    <form method="post" action="<?= $view->router()->path('pimcore_admin_login_check', ['perspective' => strip_tags($view->request()->getParameter('perspective'))]) ?>" autocomplete="off">
+    <form method="post" action="<?= $view->router()->path('pimcore_admin_login_check', ['perspective' => strip_tags($view->request()->getParameter('perspective'))]) ?>">
 
         <?php if ($this->error) { ?>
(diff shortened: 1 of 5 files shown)


bundles/AdminBundle/Resources/views/Admin/Login/login.html.php
```diff
@@ -27,16 +27,16 @@
...
         <?php if ($this->error) { ?>
             <div class="text error">
                 <?= $this->translate($this->error) ?>
             </div>
         <?php } ?>
 
-        <input type="text" name="username" placeholder="<?= $this->translate("Username"); ?>" required autofocus/>
-        <input type="password" name="password" placeholder="<?= $this->translate("Password"); ?>" required/>
+        <input type="text" name="username" autocomplete="username" placeholder="<?= $this->translate("Username"); ?>" required autofocus>
+        <input type="password" name="password" placeholder="<?= $this->translate("Password"); ?>" required>
```
jdreesen, line 39: Maybe add `autocomplete="current-password"` here?

```suggestion
        <input type="password" name="password" autocomplete="current-password" placeholder="<?= $this->translate("Password"); ?>" required>
```

NiklasBr: Could have sworn I did

NiklasBr:
Regarding the `afterrender: ...` methods added, it is a limitation in ExtJs that the regular `autoComplete` value is only a boolean, this is required to use any other value than on/off.

brusch:
Nice, thanks a lot! 🙏

PR #6239, opened 2020-04-16
Prevents synchrounous XHttpRequest for update check

NiklasBr:
See https://xhr.spec.whatwg.org/#the-open()-method for more info on why


brusch:
Thanks a lot!

Issue #5965, opened 2020-03-04
[Data Objects] introduce onContainerSaved method


weisswurstkanone:
Call when a container definition gets saved, i.e.
- object brick definition
- fieldcollection  definition
- class definition

Two methods:
- preSaved
- postSaved

Input params:
* container definition
* additional params array containing additional contextual information

classSaved method is currently only when the field resides inside a classdefinition but not for other container types.

PR #6242, opened 2020-04-17
[Feature] Application Logger for class definition save actions

NiklasBr:
Resolves comment in #6211, at least potentially, by restoring the intention of logging these changes according to [comment](https://github.com/pimcore/pimcore/issues/6211#issuecomment-613462338)

## Additional info  

So, this PR is not complete, it only handles the save() action, nothing else. But I'm throwing it out here to ask if this seem to be a sensible way to move forward?

This stores the new version of the Classs as a file object in the Application Logger, allowing administrators to search through the logger by filtering on the class name and finding the class history there.

An alternative would be to save the JSON version in the file object, which would allow for a quicker path to restore the class definition by re-importing it.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
@@ -19,6 +19,8 @@ use Pimcore\Bundle\AdminBundle\HttpFoundation\JsonResponse;
 use Pimcore\Controller\EventedControllerInterface;
 use Pimcore\Db;
 use Pimcore\Event\AdminEvents;
+use Pimcore\Log\ApplicationLogger;
+use Pimcore\Log\FileObject;
 use Pimcore\Logger;
 use Pimcore\Model;
 use Pimcore\Model\Asset;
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Hi @NiklasBr , looks fine for me. Question is if we want to add this to brick and fieldcollection save as well ?

NiklasBr:
@weisswurstkanone I just thought of a second alternative I think might be better. Why not add a new [EventListener](https://github.com/pimcore/pimcore/tree/master/bundles/AdminBundle/EventListener) called something along the lines of `ObjectDefinitionListener` and have it log events on [Classes](https://github.com/pimcore/pimcore/blob/master/lib/Event/DataObjectClassDefinitionEvents.php)?

weisswurstkanone:
yes, it is on the list for milestone 7 :-)
https://github.com/pimcore/pimcore/issues/5965

weisswurstkanone:
@NiklasBr , so what's the outcome regarding bricks and fieldcollections ? Add it in the same fashion?

NiklasBr:
@weisswurstkanone I'll get back to this soon, might not be ready in time for project 83 however.

weisswurstkanone:
Take as much time as needed, this one is not super urgent. Thx @NiklasBr

NiklasBr:
Closing this because I accidentally deleted the branch and the easiest way was to create a new one from scratch

PR #6243, opened 2020-04-17
Add missing getter for PaymentNotSuccessFullException

jremmurd:
PaymentNotSuccessFullException misses Getters for protected properties `order` and `status`.


weisswurstkanone:
thx!

PR #6244, opened 2020-04-17
[Admin UI] Website Settings - Add checks for update/delete requests

### LINKED ISSUE #5998: TypeError in Website Settings if no language provided
# Bug Report

Key collisions in *Website Settings* cause silent errors and the admin interface must be reloaded.

### Expected behavior

Similar to *Properties* on objects, if the same setting is added a second time with the first setting missing a Language, the admin interface should prevent this. Or just allow it without the "hidden" JS error.

### Actual behavior

The entire tab dies and the admin interface must be reloaded.

### Steps to reproduce

1. Open Website Settings
2. Add a setting, for example key: "testSetting" of type *Text*
3. Add a value to the setting.
4. Add a second new setting with the same key
5. Add a third new setting with the same key
6. Set a Language to the first setting
7. Try to set a Language to the second setting.

(You can also try this with the JS console open to see the ExtJS error: `TypeError: k.el is null`)

dvesh3:
Resolves partially #5998

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/SettingsController.php
+++ b/bundles/AdminBundle/Controller/Admin/SettingsController.php
@@ -1777,11 +1777,7 @@ class SettingsController extends AdminController
                         case 'asset':
                         case 'object':
                             if (isset($data['data'])) {
-                                $path = $data['data'];
-                                $element = null;
-                                if ($path != null) {
-                                    $element = Element\Service::getElementByPath($setting->getType(), $path);
-                                }
+                                $element = Element\Service::getElementByPath($setting->getType(), $data['data']);
                                 $data['data'] = $element;
                             }
                             break;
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/SettingsController.php
```diff
@@ -1764,34 +1764,36 @@ public function websiteSettingsAction(Request $request)
...
-                                $element = Element\Service::getElementByPath($setting->getType(), $path);
+                if ($setting instanceof WebsiteSetting) {
+                    switch ($setting->getType()) {
+                        case 'document':
+                        case 'asset':
+                        case 'object':
+                            if (isset($data['data'])) {
+                                $path = $data['data'];
+                                $element = null;
+                                if ($path != null) {
```
NiklasBr, line 1782: This `if` should not be required, example:
```php
$data = [];
$data['data'] = null;
var_dump(isset($data['data'])); // bool(false)
```

dvesh3: Good point. Done!

PR #6248, opened 2020-04-17
[Feature] Allow PHP CS Fixer to run directly on Definition save

NiklasBr:
Once `friendsofphp/php-cs-fixer` is installed, this PR will automatically apply the `.php_cs.dist` rules on any file in `var/classes` whenever they are updated.

Because this is a potentially very computationally heavy process (I counted up to 48 seconds with debugger running in a VM) on a rather big class the request timeout was increased.

I did this primarily to better be able to diff class definition files when checking them into our repository, but it was also nice to apply these to `var/classes/DataObject/*` files as well, they look much nicer now.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/class.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/class.js
@@ -1435,6 +1435,7 @@ pimcore.object.classes.klass = Class.create({
                         values: n,
                         id: this.data.id
                     },
+                    timeout: 120000, // Longer than usual timout due to complex classes taking a long time if PHP CS Fixer runs
                     success: this.saveOnComplete.bind(this),
                     failure: this.saveOnError.bind(this)
                 });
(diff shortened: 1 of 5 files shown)


brusch:
First of all, thanks a lot for your PR! 
But to be honest, I don't like the idea😐. Saving classes is already at the moment a pretty time-intensive task, adding another layer of complexity and execution time makes this tasks even more prone to errors (webserver/proxy/PHP/... timeouts, ...) 

We should better work on generating compliant code in my opinion.

NiklasBr:
@brusch That's ok. Question: Why not store the JSON files as definition files? As I understand it they are never dynamically loaded in regular use.

brusch:
@NiklasBr well, that depends on the use-case, you are right vor very simple cases, but there are quite a lot of cases where the class definition is needed.

NiklasBr:
Do you have any example? I mean, if the class definition can be re-imported from a JSON file, why not read that JSON file?

PR #6251, opened 2020-04-19
[Bundles] Automatically execute bundle migrations on composer update

### LINKED ISSUE #5445: [Pimcore Bundles] Execute migrations of Pimcore Bundles automatically on composer update?
right now, composer update updates composer dependencies, but potential migrations are not executed - and therefor project might be broken unless migrations are executed.

BlackbitNeueMedien:
This PR provides a composer script to execute bundle migrations. Together with https://github.com/pimcore/skeleton/pull/22 this enables automatic execution of bundle migrations when doing `composer update`.

In my opinion the core and bundles should act in the same way. As core migrations get executed automatically in `composer update` by default (in pimcore skeleton's composer.json), bundle migrations should also be.
If someone does not want to execute bundle (and core) migrations automatically (as mentioned in #5445) he can remove the `post-update-cmd` script calls from `composer.json`.

Resolves #5445


lib/Composer.php
```diff
@@ -100,6 +101,52 @@ public static function executeMigrationsUp(Event $event)
...
+        $bundles = include('var/config/extensions.php'); // path is based on composer.json's path
+
+        $installedBundles = array_map(static function($bundleClassName, $active) {
+            if($active) {
+                return substr($bundleClassName, strrpos($bundleClassName, "\\")+1);
+            }
+            return null;
+        }, array_keys($bundles['bundle']), $bundles['bundle']);
+
+        $installedBundles = array_filter($installedBundles);
```
jdreesen, line 125: Would be easier if you filter first:

```suggestion
        $installedBundles = array_filter($bundles['bundle']);
        $installedBundles = array_map(static function($bundleClassName) {
            return substr($bundleClassName, strrpos($bundleClassName, "\\") + 1);
        }, array_keys($installedBundles));
```

---
I'd also suggest that you use reflection to get the short name of the class, as it's almost as fast, but expresses the intention much better:

```suggestion
        $installedBundles = array_map(static function($bundleClassName) {
            return (new \ReflectionClass($bundleClassName))->getShortName();
        }, array_keys(array_filter($bundles['bundle'])));
```

BlackbitNeueMedien: Filtering first does not work because this array_map sets array items to `null` if the respective bundles are disabled. The `array_filter` thereafter does remove those items.

Now `ReflectionClass` gets used to get the bundle class' short name, thanks.

brusch:
@BlackbitNeueMedien can you please check the code style (Travis)? Thanks a lot!

BlackbitNeueMedien:
@brusch I have no idea how this PR can cause failing tests in `bundles/AdminBundle/Controller/Admin/ElementController.php` and `bundles/AdminBundle/Controller/Admin/IndexController.php` (see https://travis-ci.com/github/pimcore/pimcore/jobs/321028534 and https://travis-ci.com/github/pimcore/pimcore/jobs/321028535). Do you have an idea?

dvesh3:
@BlackbitNeueMedien just merge master. please see https://github.com/pimcore/pimcore/pull/6252

lib/Composer.php
jdreesen: That would be the exact advantage of filtering first: then there's no need to set it to `null` because only active bundles (those with a value of `true`) arrive into `array_map` ;)

BlackbitNeueMedien: Oh you are right, sorry. Changed it.

BlackbitNeueMedien:
Now it is green, thanks @dvesh3

brusch:
Thanks a lot! 

Just one more thing, I think it would make sense to add this new functionality to the docs. 
We should at least describe how to enable/disable automatic bundle migrations.

BlackbitNeueMedien:
Added documentation.

lib/Composer.php
```diff
@@ -100,6 +100,52 @@ public static function executeMigrationsUp(Event $event)
...
+     */
+    public static function executeBundleMigrationsUp(Event $event)
+    {
+        $consoleDir = static::getConsoleDir($event, 'bundle migrations');
+
+        if (null === $consoleDir) {
+            return;
+        }
+
+        $bundles = include('var/config/extensions.php'); // path is based on composer.json's path
```
brusch, line 115: This is not ideal, because this path can be customized. 
Maybe we should use `./bin/console pimcore:bundle:list` to get the list of enables bundles? We could introduce another flag e.g. `--json` to get the list in an easy to read format. 
What do you think?

dpfaffenbauer:
This might also lead to some incompatibilities when bundles depend on each other and migrations might also depend on each other. So this also needs to take bundle priorities into account.

lib/Composer.php
BlackbitNeueMedien: You are right, I added the mentioned `--json` option for `pimcore:bundle:list`.

BlackbitNeueMedien:
@dpfaffenbauer I have not found anything about bundle priorities or migration dependencies for Symfony. Could you please direct me to the documentation for that?

dpfaffenbauer:
It's a Pimcore thing, not a Symfony thing. If you make things right, priority shouldn't matter, there are some edge cases though where it does matter.

Here is the documentation:

https://pimcore.com/docs/6.x/Development_Documentation/Extending_Pimcore/Bundle_Developers_Guide/Bundle_Collection.html

brusch:
So i guess it would be good to consider the priorities in the json dump as well.

BlackbitNeueMedien:
Added priorities to JSON dump. If for BC reasons you do not want to include this `Priority` column when `--json` is not set, please notify me.

Bundles migrations now get executed in ascending bundle priority order.

PR #6257, opened 2020-04-20
[Data object grid] Prevent error "Call to a member function getPermissions() on null"

BlackbitNeueMedien:
In https://github.com/pimcore/pimcore/blob/735282d92b7097ca2687f8881ed06075fbe116a5/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php#L1277 `DataObject\Service::enrichLayoutDefinition()` gets called with second parameter `null`.

In https://github.com/pimcore/pimcore/blob/735282d92b7097ca2687f8881ed06075fbe116a5/models/DataObject/Service.php#L1408 the method `self::getLanguagePermissions($object)` gets then called with `$object` being `null`.

And then in In https://github.com/pimcore/pimcore/blob/735282d92b7097ca2687f8881ed06075fbe116a5/models/DataObject/Service.php#L573 there is a null pointer exception `Call to a member function getPermissions() on null`.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -1274,7 +1274,7 @@ class DataObjectHelperController extends AdminController
                 $object = DataObject\AbstractObject::getById($objectId);
                 $context['object'] = $object;
             }
-            DataObject\Service::enrichLayoutDefinition($field, null, $context);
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Isn't this already fixed via 
https://github.com/pimcore/pimcore/commit/28f91cbf339e006fa7e63d1fee6b5732077f02e6

BlackbitNeueMedien:
Yes, 28f91cb does also work.

PR #6267, opened 2020-04-21
Move ElementDescriptor to Pimcore\Model\Element namespace

### LINKED ISSUE #6264: Move ElementDescriptor to Pimcore\Model\Element namespace
- it's also used for recycle bin and the block data-type, so it makes sense to move it to a more generic namespace 
- do proper deprecation (see https://github.com/pimcore/pimcore/blob/877413a3d6a454f4827139460fc9bf29e4384794/models/DataObject/ClassDefinition/Data/Href.php#L15) 
- provide an optional migration script (can be as Gist) that does search & replace in the versioning files (serialized objects) to be compatible with v7

weisswurstkanone:
resolves #6264

### DIFF
--- a/models/Element/Recyclebin/Item.php
+++ b/models/Element/Recyclebin/Item.php
@@ -350,7 +350,7 @@ class Item extends Model\AbstractModel
             new \DeepCopy\TypeFilter\ReplaceFilter(
                 function ($currentValue) {
                     $elementType = Element\Service::getType($currentValue);
-                    $descriptor = new Model\Version\ElementDescriptor($elementType, $currentValue->getId());
+                    $descriptor = new Element\ElementDescriptor($elementType, $currentValue->getId());
 
                     return $descriptor;
                 }
@@ -400,7 +400,7 @@ class Item extends Model\AbstractModel
         $copier->addTypeFilter(
             new \DeepCopy\TypeFilter\ReplaceFilter(
                 function ($currentValue) {
-                    if ($currentValue instanceof Model\Version\ElementDescriptor) {
+                    if ($currentValue instanceof Element\ElementDescriptor) {
                         $value = Element\Service::getElementById($currentValue->getType(), $currentValue->getId());
 
                         return $value;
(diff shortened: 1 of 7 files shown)


brusch:
This needs to wait until #6236 is merged, then this PR needs to be updated.

brusch:
@weisswurstkanone can u plz update this PR considering #6236

brusch:
The BC layer for the old class name is missing 😊

weisswurstkanone:
added missing commit

models/Element/Recyclebin/Item.php
```diff
@@ -400,7 +400,7 @@ public function unmarshalData($data)
         $copier->addTypeFilter(
             new \DeepCopy\TypeFilter\ReplaceFilter(
                 function ($currentValue) {
-                    if ($currentValue instanceof Model\Version\ElementDescriptor) {
+                    if ($currentValue instanceof Element\ElementDescriptor) {
```
dvesh3, line 403: when you have already deleted data with `Model\Version\ElementDescriptor` relations? it will be ignored when restoring from recyclebin, right?

models/Version/UnmarshalMatcher.php
```diff
@@ -35,7 +35,7 @@ public function __construct()
      */
     public function matches($element)
     {
-        if ($element instanceof ElementDescriptor) {
```
dvesh3, line 38: similar problem..unmarshalling will not work with existing version data.

models/Element/Recyclebin/Item.php
weisswurstkanone: did you see the alias ?

dvesh3: yes, you are right! it should work with alias. 😜

models/Version/UnmarshalMatcher.php
dvesh3: never mind. it works with alias!

PR #6269, opened 2020-04-21
[Data objects] Override inherited value with same value (break inheritance)

BlackbitNeueMedien:
Steps to reproduce:
1. Create class `Category`
1. Create class `Product` with inheritance enabled and add a many-to-many object relation field `categories` with allowed class `Category` and `lazy-loading` enabled
1. Create `Category` object with path `/category`
1. Run `bin/console pimcore:run-script` with the following script:
```php
<?php
$parent = new \Pimcore\Model\DataObject\Product;
$parent->setParentId(1);
$parent->setKey('parent');
$category = \Pimcore\Model\DataObject\Category::getByPath('/category');
$parent->setCategories([$category]);
$parent->save();

$child = new \Pimcore\Model\DataObject\Product;
$child->setParent($parent);
$child->setKey('child');
$child->save();
```
After this the child object with path `/parent/child` inherits the field `categories` from its parent.

Now call `bin/console pimcore:run-script` again but with the following script:
```php
<?php
$child = \Pimcore\Model\DataObject\Product::getByPath('/parent/child');
$category = \Pimcore\Model\DataObject\Category::getByPath('/category');
$child->setCategories([$category]);
$child->save();
```

Expected behaviour (at least for me) is that the categories relation gets set directly for the child object, so that the inheritance gets broken up.
Actual behaviour is that the `categories` field stays inherited because the object value did not change and so the setter does not set the value directly for the child object.

Currently I have to set the relation to `null` first and thereafter set it to the desired value to break up the inheritance.

With this PR inheritance gets disabled in the setter to get the current value. This way you can directly set the desired value, even if it is the same as the previously inherited one.

### DIFF
--- a/models/DataObject/ClassDefinition/Data.php
+++ b/models/DataObject/ClassDefinition/Data.php
@@ -840,14 +840,15 @@ abstract class Data
         }
 
         if ($this->supportsDirtyDetection()) {
-            $code .= "\t" . 'if(\Pimcore\Model\DataObject::doGetInheritedValues($this->getObject())) {' . "\n";
+            $code .= "\t" . '$class = $this->getObject() ? $this->getObject()->getClass() : null;' . "\n";
+            $code .= "\t" . 'if($class && $class->getAllowInherit()) {' . "\n";
             $code .= "\t\t" . '$inheritValues = $this->getObject()::getGetInheritedValues();'."\n";
             $code .= "\t\t" . '$this->getObject()::setGetInheritedValues(false);'."\n";
             $code .= "\t" . '}'."\n";
             $code .= "\t" . '$currentData = $this->get' . ucfirst($this->getName()) . '();' . "\n";
-            $code .= "\t" . 'if(\Pimcore\Model\DataObject::doGetInheritedValues($this->getObject())) {' . "\n";
+            $code .= "\t" . 'if($class && $class->getAllowInherit()) {' . "\n";
             $code .= "\t\t" . '$this->getObject()::setGetInheritedValues($inheritValues);'."\n";
-            $code .= "\t" . '}'."\n";
+            $code .= "\t" . '}' . "\n";
             $code .= "\t" . '$isEqual = $fd->isEqual($currentData, $' . $key . ');' . "\n";
             $code .= "\t" . 'if (!$isEqual) {' . "\n";
             $code .= "\t\t" . '$this->markFieldDirty("' . $key . '", true);' . "\n";
--- a/models/DataObject/ClassDefinition/Data.php
+++ b/models/DataObject/ClassDefinition/Data.php
@@ -740,7 +740,13 @@ abstract class Data
         }
 
         if ($this->supportsDirtyDetection()) {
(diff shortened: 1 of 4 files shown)


models/DataObject/ClassDefinition/Data.php
```diff
@@ -740,7 +740,13 @@ public function getSetterCode($class)
         }
 
         if ($this->supportsDirtyDetection()) {
+            if($class instanceof DataObject\ClassDefinition && $class->getAllowInherit()) {
+                $code .= "\t" . "self::setGetInheritedValues(false);\n";
+            }
             $code .= "\t" . '$currentData = $this->get' . ucfirst($this->getName()) . '();' . "\n";
+            if($class instanceof DataObject\ClassDefinition && $class->getAllowInherit()) {
+                $code .= "\t" . "self::setGetInheritedValues(true);\n";
```
weisswurstkanone, line 748: *) get original state before setting it to false and restore the state here instead of setting it to true
*) same change probably also needed for localized field and brick

BlackbitNeueMedien: Fixed.

models/DataObject/ClassDefinition/Data.php
```diff
@@ -833,7 +840,14 @@ public function getSetterCodeObjectbrick($brickClass)
         }
 
         if ($this->supportsDirtyDetection()) {
+            $code .= "\t" . 'if(\Pimcore\Model\DataObject::doGetInheritedValues($this->getObject())) {' . "\n";
```
weisswurstkanone, line 843: This looks wrong to me and probably causes the tests to fail.
I think it should be something like 

```
        if ($this->supportsDirtyDetection()) {
            $code .= "\t" . '$class = $this->getObject() ? $this->getObject()->getClass() : null;' . "\n";
            $code .= "\t" . 'if($class && $class->getAllowInherit()) {' . "\n";
            $code .= "\t\t" . '$inheritValues = $this->getObject()::getGetInheritedValues();'."\n";
            $code .= "\t\t" . '$this->getObject()::setGetInheritedValues(false);'."\n";
            $code .= "\t" . '}'."\n";
            $code .= "\t" . '$currentData = $this->get' . ucfirst($this->getName()) . '();' . "\n";
            $code .= "\t" . 'if($class && $class->getAllowInherit()) {' . "\n";
            $code .= "\t\t" . '$this->getObject()::setGetInheritedValues($inheritValues);'."\n";
            $code .= "\t" . '}' . "\n";
            $code .= "\t" . '$isEqual = $fd->isEqual($currentData, $' . $key . ');' . "\n";
            $code .= "\t" . 'if (!$isEqual) {' . "\n";
            $code .= "\t\t" . '$this->markFieldDirty("' . $key . '", true);' . "\n";
            $code .= "\t" . '}' . "\n";
        }
```

BlackbitNeueMedien: You are right. Committed those changes - let's look what Travis thinks about it ;-)

weisswurstkanone:
thx a lot @BlackbitNeueMedien

PR #6274, opened 2020-04-22
show validationExceptions for documents correctly as well

domiha:
As it was already implemented for objects, we would like to have a proper JS validation for documents as well. Without these changes, we can't close the validation box correctly (the background is still unclickable)

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
@@ -108,7 +108,7 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
             } catch (e) {
                 if (e instanceof pimcore.error.ValidationException) {
                         this.tab.unmask();
-                        pimcore.helpers.showPrettyError('document', t("error"), e.message);
+                        pimcore.helpers.showPrettyError('document', t("error"), t("saving_failed"), e.message);
                         return false;
                     }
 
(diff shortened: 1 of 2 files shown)


dvesh3:
@domiha nice work 👍 i think we should also add validation for `preSaveAsset`

bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
```diff
@@ -103,7 +103,21 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
 
             }
 
-            pimcore.plugin.broker.fireEvent("preSaveDocument", this, this.getType(), task, only);
+            try {
+                pimcore.plugin.broker.fireEvent("preSaveDocument", this, this.getType(), task, only);
+            } catch (e) {
+                if (e instanceof pimcore.error.ValidationException) {
+                        this.tab.unmask();
+                        pimcore.helpers.showPrettyError('document', t("error"), e.message);
```
dvesh3, line 111: just like objects...please use t("saving_failed") for text(3rd param) and e.message for errorText(4th param)

domiha: fixed

domiha:
> @domiha nice work 👍 i think we should also add validation for `preSaveAsset`

also for assets added 👍

dvesh3:
@domiha thanks a lot!

PR #6276, opened 2020-04-22
Fixed bug with tabs in read-only fieldcollection

### LINKED ISSUE #6138: [FieldCollection] Layout blocked when making field collection read-only.
#Bug report

## Expected behavior
It should be possible to switch tabs in a read-only field collection because relevant information should still be available for consultation.

## Actual behavior
Switching tabs is not possible because layout components also become read-only.
Screenshot 2020-04-02 at 15 23 49

## Steps to reproduce
1. Surf to the demo environment: https://demo.pimcore.fun/
2. Go to field collection definition.
3. Make a field collection with a tabpanel layout component.
4. Go to the object class definition.
5. Create a class with a read-only field collection attribute that is set on the Field collection created in step 3.
6. Make an object of the created class in step 5.
7. Add a field collection and try to switch tabs -> this won't work.

skochTC:
Resolve #6138

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/fieldcollections.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/fieldcollections.js
@@ -394,15 +394,21 @@ pimcore.object.tags.fieldcollections = Class.create(pimcore.object.tags.abstract
             this.currentData = blockData;
         }
 
-        var items =  this.getRecursiveLayout(this.layoutDefinitions[type], null,
+        var items =  this.getRecursiveLayout(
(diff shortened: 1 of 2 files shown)


skochTC:
> 
> 
> @skochTC Field collection items are not disabled anymore although field collection has been marked as read only.
> 
> image

Okay, I'll check it out. I have tested this with several data types with different layout types but actually not in a localized field. I have forgotten the use case. Sorry!

skochTC:
Someone has forgotten to set the showMode in the tab variant. For the dropdown the value was set correctly.

weisswurstkanone:
thanks @skochTC 👍

PR #6278, opened 2020-04-22
Added translation support for longer texts

### LINKED ISSUE #6277: [Bug] Texts that are longer than 190 characters cannot be translated
# Bug Report

### Expected behavior
That every text will be translated ;-)

### Actual behavior
It is not the case.

### Steps to reproduce
Simply write a longer text in the class definition into a layout element.

Now call the object and wait until the translation is added to the database with the request `/admin/translation/add-admin-translation-keys

The database now limits the key to 190 characters according to the column definition.

The key is cut off and can be translated in the admin-translations, but this has no effect, because the translation is always requested with the full text.
My solution is simple, the key is limited to 158 characters, so the context to the real text still exists.
So that the texts do not appear twice, if there are two different texts that start equal, a MD5 hash is created from the complete text and appended to the key.

### Edited on 2020/04/23
It is a bug, because texts that are longer than 190 characters are imported into the database, although they can never be accessed.

If it were the case that 190 characters long texts are not allowed at all I would most likely send a feature request ;-)

More information why this PR is useful, see below.

I hope the PR will be accepted.

skochTC:
Resolve #6277

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/TranslationController.php
+++ b/bundles/AdminBundle/Controller/Admin/TranslationController.php
@@ -352,7 +352,6 @@ class TranslationController extends AdminController
             $data = $this->decodeJson($request->get('data'));
 
             if ($request->get('xaction') == 'destroy') {
-                $data = $this->decodeJson($request->get('data'));
                 $t = $class::getByKey($data['key']);
                 $t->delete();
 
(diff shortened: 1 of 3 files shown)


brusch:
Thanks! 🙏

PR #6281, opened 2020-04-23
[Feature] Element Router

blankse:

It would be nice to can use the symfony/twig functions path/url to generate urls for elements.
I know it from other systems to use this functions also for objects.

It is also supported from symfony with the `VersatileGeneratorInterface` interface.

### DIFF
--- a/doc/Development_Documentation/05_Objects/01_Object_Classes/05_Class_Settings/30_Link_Generator.md
+++ b/doc/Development_Documentation/05_Objects/01_Object_Classes/05_Class_Settings/30_Link_Generator.md
@@ -125,6 +125,30 @@ would produce the following output
  
 ### Use in Views
 
+#### path() / url()
+
+<div class="code-section">
+
+```php
+<ul class="foo">
+    <?php foreach($this->carList as $car) { ?>
+        <li><a href="<?= $this->path($car); ?>"><?= $car->getName() ?></a></li>
+    <?php } ?>
+</ul>
+```
+
+```twig
+<ul class="foo">
+    {% for car in carList %}
+        <li><a href="{{ path(car) }}">{{ car.getName() }}</a></li>
+    {% endfor %}
+</ul>
+```
+
+</div>
+
+#### pimcoreUrl
+
 <div class="code-section">
 
 ```php
@@ -138,7 +162,7 @@ would produce the following output
 ```twig
 <ul class="foo">
     {% for car in carList %}
-        <li><a href="{{ path(car) }}">{{ car.getName() }}</a></li>
+        <li><a href="{{ pimcore_url({object: car}) }}">{{ car.getName() }}</a></li>
     {% endfor %}
 </ul>
 ```
--- a/doc/Development_Documentation/05_Objects/01_Object_Classes/05_Class_Settings/30_Link_Generator.md
+++ b/doc/Development_Documentation/05_Objects/01_Object_Classes/05_Class_Settings/30_Link_Generator.md
@@ -125,12 +125,22 @@ would produce the following output
  
 ### Use in Views
 
+<div class="code-section">
+
 ```php
 <ul class="foo">
     <?php foreach($this->carList as $car) { ?>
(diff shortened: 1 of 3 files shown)


brusch:
Wouldn't it be better to just implement `VersatileGeneratorInterface` on `\Pimcore\Routing\DataObjectRoute` and `\Pimcore\Routing\DocumentRoute` ?

brusch:
That wouldn't cover Assets but for them I don't think it matters at all. 😊

blankse:
@brusch The `VersatileGeneratorInterface` must set to a router and not to a route. I don't see any possibility there.

brusch:
Oh yes, sorry, you're right.

doc/Development_Documentation/05_Objects/01_Object_Classes/05_Class_Settings/30_Link_Generator.md
```diff
@@ -125,12 +125,22 @@ would produce the following output
  
 ### Use in Views
 
+<div class="code-section">
+
 ```php
 <ul class="foo">
     <?php foreach($this->carList as $car) { ?>
-        <a href="<?= $this->pimcoreUrl(['object' => $car]); ?>"><?= $car->getName() ?></a>
+        <li><a href="<?= $this->pimcoreUrl(['object' => $car]); ?>"><?= $car->getName() ?></a></li>
```
brusch, line 133: Here it would be `$this->path()` as well. 
Maybe it's even better to document both ways ... the path helper and the pimcoreUrl helper.

brusch: Additionally it would be good to add a short description, that the path helper can be used with assets, documents and data objects as well.

blankse: Yes we can document both ways.

But I think we should recommend path and set pimcoreUrl deprecated. What do you think?

PhpEngine is also deprecated from symfony site. Starting from Symfony 5.0 the php engine is removed. So we should also recommend twig.

brusch: > But I think we should recommend path and set pimcoreUrl deprecated. What do you think?
Yes, we can recommend using `path`. 
Regarding the deprecation of `pimcoreUrl`: let's think about that again and fix it in a dedicated PR 😊

> So we should also recommend twig. 
Yes, but also a different topic 😊

dpfaffenbauer: pimcoreUrl has some advantages that path doesn't have, so I think it is still legitimate. 

PimcoreUrl allows replacing certain parameters without doing some twig|merge stuff or anything like that. So I think that there is a use-case for it. At least for me, I use it in some cases.

blankse: @brusch Yes we can wait with the deprecation. 
@dpfaffenbauer Maybe we can transfer this functionality also to the router? 

There are also some point to do I think. The new slug datatype is so not supported I think. And the referenceType is also not considered. But should be a good start.

dpfaffenbauer: wouldn't work, since pimcoreUrl also works with static routes, maybe also with symfony routes.

blankse: @dpfaffenbauer We can extend the symfony chain router, so it can be done for all routes.

dpfaffenbauer: We can, but I don't think we should. I think it is fine to have these two. As a symfony developer I know what to expect with the default router (eg. path) and if I have use-cases I can use pimcoreUrl

PR #6287, opened 2020-04-23
[Feature Request] Classification-Store group limitation

### LINKED ISSUE #6285: [Feature Request] Classification-Store group limitation
# Feature Request

## Feature description
We use the Classification store for the classification system ETIM and will also integrate eCl@ss into a customer project in a few weeks.

Many classification systems are very extensive and allow only one class, in Pimcore group. Unfortunately there is no option to set this in Pimcore via the class definition.

Some months ago I implemented this for a customer project based on a bundle and now I want to share the code in the core.

skochTC:
Resolve #6285

I have a general question:
When I do adjustments, is it legitimate to make cs-fixes?

line 56 to 61, for example, triggered me a lot,
only I don't want to disrupt the Blame function.
bundles/pimcoreadmin/js/pimcore/object/classes/data/classificationstore.js


fashxp:
Since cs-fixer is executed automatically and would disrupt history too, I would say minor cs fixes are legitimate - as long as they don't obfuscate the actual adjustment and therefore make reviewing harder or even impossible!
If there would be bigger cs fixes, please create a separate PR.

skochTC:
> Since cs-fixer is executed automatically and would disrupt history too, I would say minor cs fixes are legitimate - as long as they don't obfuscate the actual adjustment and therefore make reviewing harder or even impossible!
> If there would be bigger cs fixes, please create a separate PR.

Thanks for this information 👍

bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/classificationstore.js
```diff
@@ -77,6 +76,14 @@ pimcore.object.classes.data.classificationstore = Class.create(pimcore.object.cl
             value: this.datax.allowedGroupIds
         });
 
+        this.specificPanel.add({
+            xtype: "numberfield",
+            fieldLabel: t("maximum_items"),
```
weisswurstkanone, line 81: can we use a dedicated label here so that it is clear to the user that it is about groups

skochTC: Yes sure, but I can only translate the English and German text.

Is there already an idea for translating them?

weisswurstkanone: Note that only en.json has to be touched. The other languages are managed in POEditor.
See https://pimcore.com/docs/6.x/Development_Documentation/Multi_Language_i18n/Admin_Translations.html

weisswurstkanone:
Looks from my side.

Open point:
- introduce translation key
- please don't forget to update the screenshot

Thx 👍

weisswurstkanone:
Maybe one additional idea: hide or disable the "plus" button when the limit is already reached. What do you think ?

skochTC:
> 
> 
> Maybe one additional idea: hide or disable the "plus" button when the limit is already reached. What do you think ?

Principally yes but I think a dialog message is better for the usability. When controls are changed, it can often be confusing.

weisswurstkanone:
I think the most important thing is to keep it consistent. Fieldcollections also allow to limit the items and there you also get a dialog basically saying the same thing. So keep it as it is 👍

bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/classificationstore.js
skochTC: Done ;-)

The linked "POEditor" link in the documentation is dead -> https://pimcore.com/en/resources/translations

No idea what the new / correct link is otherwise I would adjust it. If necessary you take it up as a to-do?

weisswurstkanone: thx for the notice, I fixed the link in https://github.com/pimcore/pimcore/commit/df66f7a13d050ba198108920b3ddf081029bbb7b

weisswurstkanone:
thank you very much @skochTC

PR #6288, opened 2020-04-23
Added classification-store translation support

### LINKED ISSUE #3888: Unify Classification Store Translation
There are several translation mismatches if I'm in a classification store context:

## Keys
If I'm defining a nice translation title for a simple key:

bildschirmfoto 2019-01-29 um 11 06 17

=> The name itself still shows up in admin translation:

bildschirmfoto 2019-01-29 um 11 10 52

***

If I'm adding a description:

bildschirmfoto 2019-01-29 um 11 22 02

=> The single key does not show up in the translator. Instead, the name (not title) + description will be added concatenated in backend translation:

bildschirmfoto 2019-01-29 um 11 21 55

## Groups
If I'm opening the CL search field, the name does not show up translated:
bildschirmfoto 2019-01-29 um 11 05 33

***

If I'm adding a description:

bildschirmfoto 2019-01-29 um 11 20 13

=> The single key does not show up in the translator. Instead, the name + description will be added concatenated in backend translation:

bildschirmfoto 2019-01-29 um 11 19 50

## Collections
The collection name and description skips the translation fetch at all (Probably because it's not shown up as a real field in the object layout). Logically, if I'm opening the CL search field, the name does not show up translated:
bildschirmfoto 2019-01-29 um 11 25 48

skochTC:
Resolves #3888

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classificationstore/keySelectionWindow.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classificationstore/keySelectionWindow.js
@@ -435,7 +435,37 @@ pimcore.object.classificationstore.keySelectionWindow = Class.create({
         this.store = new Ext.data.Store({
             remoteSort: true,
             proxy: proxy,
-            fields: readerFields
+            fields: readerFields,
+            filters: [
+                function(item) {
+                    // translate collection and group results
+                    if (item.get('name')) {
+                        item.set('name', t(item.get('name')));
+                    }
+
+                    if (item.get('description')) {
+                        item.set('description', t(item.get('description')));
+                    }
+
+                    // translate group by key results
+                    if (item.get('keyName')) {
+                        item.set('keyName', t(item.get('keyName')));
+                    }
+
+                    if (item.get('keyDescription')) {
+                        item.set('keyDescription', t(item.get('keyDescription')));
+                    }
+
+                    if (item.get('groupName')) {
+                        item.set('groupName', t(item.get('groupName')));
+                    }
+
+                    // remove modified state so there is no ui effect
+                    item.modified = {};
+
+                    return item;
+                }
+            ]
         });
 
         var pageSize = pimcore.helpers.grid.getDefaultPageSize(-1);
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classificationstore/keySelectionWindow.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classificationstore/keySelectionWindow.js
@@ -436,36 +436,39 @@ pimcore.object.classificationstore.keySelectionWindow = Class.create({
             remoteSort: true,
             proxy: proxy,
             fields: readerFields,
-            filters: [
-                function(item) {
-                    // translate collection and group results
-                    if (item.get('name')) {
-                        item.set('name', t(item.get('name')));
-                    }
+            listeners: {
+                // translate classification store texts
+                load: function (store, records, options) {
+                    Ext.Array.forEach(records, function(item) {
+                        // translate collection and group results
+                        if (item.get('name')) {
+                            item.set('name', t(item.get('name')));
+                        }
 
-                    if (item.get('description')) {
-                        item.set('description', t(item.get('description')));
-                    }
+                        if (item.get('description')) {
+                            item.set('description', t(item.get('description')));
+                        }
 
-                    // translate group by key results
-                    if (item.get('keyName')) {
-                        item.set('keyName', t(item.get('keyName')));
-                    }
+                        // translate group by key results
+                        if (item.get('keyName')) {
+                            item.set('keyName', t(item.get('keyName')));
+                        }
 
-                    if (item.get('keyDescription')) {
-                        item.set('keyDescription', t(item.get('keyDescription')));
-                    }
+                        if (item.get('keyDescription')) {
+                            item.set('keyDescription', t(item.get('keyDescription')));
+                        }
 
-                    if (item.get('groupName')) {
-                        item.set('groupName', t(item.get('groupName')));
-                    }
+                        if (item.get('groupName')) {
+                            item.set('groupName', t(item.get('groupName')));
+                        }
 
-                    // remove modified state so there is no ui effect
-                    item.modified = {};
+                        // remove modified state so there is no ui effect
+                        item.dirty = false;
+                    });
 
-                    return item;
-                }
-            ]
+                    this.store.setRecords(records);
+                }.bind(this)
+            }
         });
 
         var pageSize = pimcore.helpers.grid.getDefaultPageSize(-1);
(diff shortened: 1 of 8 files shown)


skochTC:
I used `Admin::getByKeyLocalized($var, true, true)` because the translation key would be created anyway in the admin ui via the route `pimcore_admin_translation_addadmintranslationkeys`.

I hope that' s okay.

skoch98:
I was reviewing my changes and discovered that it doesn't make sense to change the API so much. At the moment the texts are translated with very few exceptions based on javascript.

So I have chosen to modify my adjustments.
The fallback tooltip is included in my adjustments.

bundles/AdminBundle/Resources/public/js/pimcore/object/classificationstore/keySelectionWindow.js
```diff
@@ -435,7 +435,37 @@ pimcore.object.classificationstore.keySelectionWindow = Class.create({
         this.store = new Ext.data.Store({
             remoteSort: true,
             proxy: proxy,
-            fields: readerFields
+            fields: readerFields,
+            filters: [
+                function(item) {
```
weisswurstkanone, line 440: can you explain why this is needed and how it works ? don't get the point here especially because it is supposed to return a boolean ?

bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/edit.js
```diff
@@ -404,9 +404,25 @@ pimcore.object.helpers.edit = {
                             // apply tooltips
                             if(field.tooltip) {
                                 try {
+                                    var tooltipHtml = field.tooltip;
+
+                                    // special rule for classification store keys
+                                    // in this context there is no way to refer to the description of a classification store keys
+                                    // therefore the fallback tootip provided by the api will be separated and translated correctly
+                                    if (dataProvider.component !== undefined &&
+                                        dataProvider.component.bodyCls == 'pimcore_object_tag_classification_store' &&
```
weisswurstkanone, line 413: this looks a bit fragile - what if the bodyCls changes ... nobody will probably know to change the rule here as well.

Is there a reason why this cannot be done here?
https://github.com/pimcore/pimcore/blob/d07615a97ed4b95e4f4a1e82ed0354f82ed5b839/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/classificationstore.js#L419

bundles/AdminBundle/Resources/public/js/pimcore/object/classificationstore/keySelectionWindow.js
skochTC: It is used to translate the texts, because I have dropped the changes in the backend.

Filters are used to modify values, you can write a function to process the complete data set or define one filter per data key. This is necessary because the data is passed directly through the API into the GRID table over the proxy.

Well, you know, I missed the return type, and I'll fix that right up.

Reference to the ExtJs documentation:
https://docs.sencha.com/extjs/6.2.0/modern/Ext.data.AbstractStore.html#cfg-filters

weisswurstkanone: And this cannot be done in an event handler (e.g. on "load") ?

skochTC: This would also work but I find that filters are better for this.

With "load" I can change the data, but I have to overwrite it in the store with ".bind(this)". Personally, I find this a very bad solution.

If you prefer, I can make that like this way. If it really works in detail. I was just touching on that until I saw the filters.

weisswurstkanone: If it doesn't work with event then please add inline documentation explaing why it is done (similar to what you described above). I think otherwise this is not obvious :-)

skochTC: Done, with the load handler 👍

bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/edit.js
skochTC: That makes sense. I fixed it.

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/classificationstore.js
```diff
@@ -416,14 +416,15 @@ pimcore.object.tags.classificationstore = Class.create(pimcore.object.tags.abstr
                 context["applyDefaults"] = true;
             }
 
+            context["type"] = this.type;
```
weisswurstkanone, line 419: what I was suggestion was to manipulate the tooltip here instead of the central edit.js (which shouldn't be even aware of the data type it is dealing with). what do you think ?

skochTC: Well, I honestly don't know how I would best solve that.

At this point I am in the stroe context, not in the field.

I would have to run over the elements and change the text there somehow and in the recursive layout method I would still have to check if the text is already translated because otherwise the translated text would be translated.

weisswurstkanone: well, childItem is already the thing you need.

But I guess special handling would be still needed in edit.js. If you remove the if (...classificationstore ...) then the tooltip wil be still retranslated. So let's leave it as it is for now.

weisswurstkanone:
Notices one thing.

Open object with existing cs group.

image

Tooltip is there. Remove the group and re-add through the search dialog.

Tooltip is gone:

image

bundles/AdminBundle/Resources/public/js/pimcore/object/classificationstore/keySelectionWindow.js
```diff
@@ -435,7 +435,40 @@ pimcore.object.classificationstore.keySelectionWindow = Class.create({
         this.store = new Ext.data.Store({
             remoteSort: true,
             proxy: proxy,
-            fields: readerFields
```
weisswurstkanone, line 438: The longer I think about it .... why not simply use a column renderer.

Example can be found here:
image

skoch98:
> Notices one thing.
> 
> Open object with existing cs group.
> 
> (removed image)
> 
> Tooltip is there. Remove the group and re-add through the search dialog.
> 
> Tooltip is gone:
> 
> (removed image)

This only happens if no tooltip has been maintained on the store key. The fallback tooltip is set up in the API as is known.

In this context you don't have access to the description and that's a bug that is not caused by my changes 😉 

After a couple of :coffee: I found a nice solution to translate it. At least I kept the fallback in the API to ensure backwards compatibility.

bundles/AdminBundle/Resources/public/js/pimcore/object/classificationstore/keySelectionWindow.js
skoch98: Now looks a lot better :smiley:

weisswurstkanone:
thanks a lot @skoch98

Issue #5828, opened 2020-02-17
Calculated field not working in ObjectBricks LocalizedFields


dpfaffenbauer:
```
Message: Call to undefined method getName in class Pimcore\Model\DataObject\Objectbrick\Definition
Trace: 
#0 /var/www/html/vendor/pimcore/pimcore/models/DataObject/ClassDefinition/Data/CalculatedValue.php(502): Pimcore\Model\AbstractModel->__call('getName', Array)
#1 /var/www/html/vendor/pimcore/pimcore/models/DataObject/ClassDefinition/Data/Localizedfields.php(831): Pimcore\Model\DataObject\ClassDefinition\Data\CalculatedValue->getSetterCodeLocalizedfields(Object(Pimcore\Model\DataObject\Objectbrick\Definition))
#2 /var/www/html/vendor/pimcore/pimcore/models/DataObject/Objectbrick/Definition.php(277): Pimcore\Model\DataObject\ClassDefinition\Data\Localizedfields->getSetterCode(Object(Pimcore\Model\DataObject\Objectbrick\Definition))
#3 /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php(1082): Pimcore\Model\DataObject\Objectbrick\Definition->save()
#4 /var/www/html/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(151): Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\ClassController->objectbrickUpdateAction(Object(Symfony\Component\HttpFoundation\Request))
#5 /var/www/html/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(68): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#6 /var/www/html/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php(198): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#7 /var/www/html/web/app.php(50): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#8 {main
```

It assumes it is within a regular classes LocalizedFields.

brusch:
I guess this is currently just not supported. 
[...]

PR #6290, opened 2020-04-24
Reinvent calculated value context

### LINKED ISSUE #6128: Using calculated fields within blocks doesn't work
# Bug Report
The calculator isn't called if the field is placed within a block.

Appears in Pimcore 6.6 using services and simple classes.

### Expected behavior
The calculator being called for each field, with right context, even in blocks. If if blocks not supported, restrict fields which are not supported.

### Actual behavior
The calculator class ist not called and field stays blank.

### Steps to reproduce
Add a calculator class within a block and add simple calculator either as class or service. Value is not set and class never called.

weisswurstkanone:
Proposes a new and more flexible approach how context information is passed.

- Arbitrary nesting levels are possible now.
- Old way can still be used in parallel  but marked as deprecated.
- Blocks - which is a new feature - only support the new way.

resolves #6128 

also see #5828


bundles/CoreBundle/Migrations/Version20200423113512.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $list = new ClassDefinition\Listing();
+        $list = $list->load();
+
+        foreach ($list as $class) {
+            $class->save(false);
```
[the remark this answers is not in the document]
weisswurstkanone: thx!

brusch:
@weisswurstkanone maybe "stack" would be a better name than "chain", just as the `RequestStack`, `DocumentStack` or `BlockStateStack`.

models/DataObject/ContextChain/AbstractNode.php
```diff
@@ -0,0 +1,30 @@
...
+ * @category   Pimcore
+ * @package    Object
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Model\DataObject\ContextChain;
+
+abstract class AbstractNode
```
brusch, line 20: @weisswurstkanone 
Imho a context object should always have a subject, similar to Symfony events. 
The subject is then an instance of a Brick, Localized Fields, ... 
This would make the implementation classes a bit leaner, since we don't need getters/setters for inromation like names, ... since they can be obtained by the subject object. Only contextual information which is not included in the subject object needs to be added to the implementation class, such as index information, or the locale and others. 

What do you think?

dpfaffenbauer: Maybe also a factory should be responsible in creating these elements

brusch:
Closed for now, see also #6411

PR #6297, opened 2020-04-27
[Routing] Element Router must implement RouterInterface

dvesh3:
Follow up to #6281
Resolves issues with the data-hub

Screenshot

### DIFF
--- a/lib/Routing/Element/Router.php
+++ b/lib/Routing/Element/Router.php
@@ -25,11 +25,13 @@ use Symfony\Component\Routing\Exception\ResourceNotFoundException;
 use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Routing\Matcher\RequestMatcherInterface;
 use Symfony\Component\Routing\RequestContext;
+use Symfony\Component\Routing\RouteCollection;
+use Symfony\Component\Routing\RouterInterface;
 
 /**
  * A custom router implementation handling pimcore elements.
  */
-class Router implements RequestMatcherInterface, VersatileGeneratorInterface
+class Router implements RouterInterface, RequestMatcherInterface, VersatileGeneratorInterface
 {
     /**
      * @var RequestContext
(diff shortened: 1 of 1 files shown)


blankse:
Maybe the interface should be checked in the GraphQLController? Symfony also allows routers without this interface. Please refer:
https://github.com/symfony-cmf/Routing/blob/master/src/ChainRouter.php#L82-L98

brusch:
@blankse yep, would make sense.
@dvesh3 can you add this check? thanks a lot!

brusch:
Probably this change can then be discarded ...

brusch:
Doesn't make sense to implement interfaces which are actually not needed 😐

blankse:
I looked deeper. The ChainRouter always want the `getRouteCollection` function from its router without checking the Interface. Seems like a bug in Symfony than... Can't be changed in the GraphQLController.

dvesh3:
@blankse @brusch  the check was added to support ChainRouter https://github.com/symfony-cmf/Routing/blob/bee5389c27aa57ba0c0ca9bb09590ba77578ddd5/CHANGELOG.md#changelog

also something to note here: https://github.com/symfony/routing/blob/aa0ed350e84a464f53d429992ecaebd6e0c87c08/RouterInterface.php#L18

romainbessugesmeusy:
The clientname bug in Datahub is still there. Tried 6.6.2 and 6.6.4

brusch:
@romainbessugesmeusy doesn't seem to be reproducible on our public demo ... any more details? 😐

romainbessugesmeusy:
The problem resides in the data-hub repository. It has been fixed on dev-master but hasn't made it to a tag yet. Sorry for bothering here

brusch:
@weisswurstkanone maybe time for a release tag then? 😊

weisswurstkanone:
https://github.com/pimcore/data-hub/releases/tag/v0.6.2

PR #6306, opened 2020-04-28
[Redirects] HTML-encode redirect source and target

BlackbitNeueMedien:
Steps to reproduce bug:
1. Create redirect with source `/test?test=1&amp=1`
1. Save redirect

Expected behaviour: Redirect source is `/test?test=1&amp=1`
Actual behaviour: Redirect source is `/test?test=1&=1`

This is only a bug in the user interface - in the database and program logic everything works correctly.


brusch:
Thanks a lot! 🙏

PR #6309, opened 2020-04-28
[CSV Import] No error if header and data rows have different number of columns

blankse:

If you have a csv file with empty columns only on the end of the data rows, all seems okay in the admin interface. It shows the data in preview. Only the import is immediately successful without importing anything.

I think pimcore should immediately display an error message and not mislead the user.


brusch:
Thanks!

PR #6311, opened 2020-04-29
add site for prepareFrontendPath

zerlpaMMT:
To get an correrkt fullpath if document is in site.


brusch:
Unfortunately this would have a huge performance impact, because each (initial) call to `getFullPath()` on a document would result in an additional SQL query. So e.g. for navigations this would mean quite a lot of extra queries.

PR #6312, opened 2020-04-29
[Reports] Flex column width if width not explicitly set

BlackbitNeueMedien:
Currently if column width is not explicitly set ExtJS's default column width gets used. This is not optimal as it is not responsive. This PR sets `flex: 1` attribute for all columns which have no width being set. This way the columns use the whole available panel width.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/report/custom/report.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/report/custom/report.js
@@ -66,6 +66,8 @@ pimcore.report.custom.report = Class.create(pimcore.report.abstract, {
 
             if(colConfig["width"]) {
                 gridColConfig["width"] = intval(colConfig["width"]);
+            } else {
+                gridColConfig["flex"] = 1;
             }
 
             if(colConfig["filter"]) {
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/report/custom/report.js
```diff
@@ -232,9 +234,7 @@ pimcore.report.custom.report = Class.create(pimcore.report.abstract, {
             plugins: ['pimcore.gridfilters'],
             stripeRows: true,
             trackMouseOver: true,
-            viewConfig: {
-                forceFit: false
-            },
+            forceFit: false,
```
weisswurstkanone, line 237: Are you sure that this shouldn't be a viewConfig setting?
Maybe "false" is the default anyway so this can go away entirely?

BlackbitNeueMedien: I am quite sure, see https://docs.sencha.com/extjs/6.0.2/classic/Ext.grid.Panel.html#cfg-forceFit while for `viewConfig` of a grid panel "Any of the config options available for [Ext.view.Table](https://docs.sencha.com/extjs/6.0.2/classic/Ext.view.Table.html) can be specified here" - and `forceFit` is not listed there.

Sadly I cannot see the default value in the documentation, for this reason I kept it.

weisswurstkanone: ok, there is probably some Extjs magic going on :-)

PR #6313, opened 2020-04-29
[Admin] DataObjects - URL Slugs readonly layout fix

dvesh3:
Read-only URL Slug bug fix:
image


bundles/AdminBundle/Resources/public/js/pimcore/object/tags/urlSlug.js
```diff
@@ -257,7 +257,11 @@ pimcore.object.tags.urlSlug = Class.create(pimcore.object.tags.abstract, {
 
     getLayoutShow: function () {
         var layout = this.getLayoutEdit();
-        this.component.setReadOnly(true);
+        for (key in this.elements) {
```
weisswurstkanone, line 260: let key

weisswurstkanone:
this should be disabled as well

image

dvesh3:
> this should be disabled as well
> 
> image

Hide "Add Site" when readonly

PR #6320, opened 2020-04-29
Ecommerce | Load Elastic Search (ES) Data directly from Index

### LINKED ISSUE #5940: [Ecommerce][Indexservice] Productlist cache inconsistency

andreas-gruenwald:
would also fix #5940

# Feature Description
This PR provides an approch to use product data directly from the Elastic Search (ES) index instead of accessing the cache and the ecommerce store table. By doing so, the response is much faster for uncached product items. 

## Current State
1. A search request is sent to ES.
2. The product IDs are gathered from the ES response. The full source is ignored.
3. The product data is loaded from the **database store table per record**, if not existing in cache yet.
4. The store table data is added do the cache, so that 3 + 4 can be skipped the next time.

## Solution Suggestion
1. A search request is sent to ES.
2. The product IDs are gathered from the ES response. The full source is part of it.
3. The response is directly mapped onto a product mockup object and returned.

At the moment the solution includes a switch so that you can turn-on/off the direct mode by calling ``setLoadFromSource(true)`` explicitly. We found out that this is helpful to compare the results of the two different approaches, which simplifies testing. Later on, we could think to remove that option and just offer the new approach.

## Benchmarks + Testing
When the store table is not cached, the direct approach takes ms instead of seconds!

Results from a specific project setup:

 - No product data cached: between 1 and 4 seconds.
 - otherwise: between 300 and 600ms

Direct approach:
 - No product data cached: between 200 and 600ms
 - otherwise: between 200 and 600ms

In our specific project we tested the approach based on ``loadWithoutPriceFilterWithoutPriceSorting()``. The approach worked very well, and we didn't find any issues. However, ``loadWithoutPriceFilterWithPriceSorting`` should still be tested thoroughly.


andreas-gruenwald:
@fashxp Travis fails, but I don't see a related error in the logs.
Is there any problem with the PR which I don't see, or is it system-related?

weisswurstkanone:
@andreas-gruenwald the log states that there are new phpstan errors (violating the baseline rules).
This happens because there was an update. See
https://github.com/pimcore/pimcore/pull/6380

bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -1419,4 +1474,43 @@ public function valid()
...
+    }
+
+    /**
+     * Get the score from a loaded product list based on a (Pimcore) product Id.
+     * Precondition: isLoadFromSource - mode must be set to true.
+     * @param int $productId the Pimcore product Id.
+     * @return float the score returned by Elastic Search.
+     * @throws \Exception if loadFromSource mode is not true.
+     */
+    public function getScoreFromLoadedList(int $productId): float
```
fashxp, line 1502: what has that to do with the PR?

andreas-gruenwald: Based on the new approach it is possible to access the score which can be very helpful for debugging search implementations. Until now the score was not accessible based on productLists.

If you prefer, we can either remove the method completely, or create another PR for it.
However, I am not 100% if this is the best place for the solution anyway, so probably we should discuss that matter in another issue or PR after this PR has been completed?

bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -627,12 +645,20 @@ protected function loadWithoutPriceFilterWithPriceSorting()
...
+            $this->hitData = [];
+            $params['body']['_source'] = true;
+        }
         $result = $this->sendRequest($params);
         $objectRaws = [];
         if ($result['hits']) {
             $this->totalCount = $result['hits']['total'];
             foreach ($result['hits']['hits'] as $hit) {
-                $objectRaws[] = ['id' => $hit['_id'], 'priceSystemName' => $hit['_source']['system']['priceSystemName']];
+                $objectRaw = ['id' => $hit['_id'], 'priceSystemName' => $hit['_source']['system']['priceSystemName']];
```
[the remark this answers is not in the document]
andreas-gruenwald: In the Pimcore version I used there was a ``reset()``  contained in (removed) line 635, which causd side effects on debugging. That's why I introduced a variable - to simplify debugging. We can undo the change, if you prefer.

andreas-gruenwald:
@fashxp 
> Hmm, I'm not sure. Right now this feels like a half-way solution.
> 

> If the benefit of loading the mockup data directly from ES is so big, then we should get rid of mockup cache for ES implementations completely ...

The main reasons, why I didn't remove the entire mockup cache solution immediately were a) testability during the PR release, and b) downward compatibility: until now it didn't hurt that much if ES index updates for particular documents failed, were incomplete or outdated, as long as the full data was contained in the store table, which was is very likely. The old version is included to leave a backdoor and make updates less painfully. Without the old version this is going to be a BC...

If you prefer to make a hard switch, then let's do it ;-)

 that would also mean to get rid of `AbstractMockupCacheWorker` and update crc sums directly in [`doUpdateIndex` of the worker](https://github.com/pimcore/pimcore/blob/7e5478fbfaf066fb835ee5ccaa94cbf9e9326eb5/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L372).
> So we reduce complexity too.

Ok, that's another topic. If we get rid of the old version, we can also remove the mockupcache worker for the Elastic Search part, I agree.

> What is the difference between 'Results from a specific project setup' and 'Direct approach' in your testing?

'Results from a specific project setup' is the solution with the cache, where the data is not accessed directly from the ES index. 

'Direct approach' is the solution that was provided in the PR without accessing the cached mockups, instead reusing the data from the index directly.

> Did you do any comparison between your implementation and the original implementation with cache filled?

We cleared the cache for a particular product list and refreshed the page in the frontend several times, forcing the Pimcore cache to be filled.

The result is that both approaches took between 300 and 600ms (cf. initial PR description).

> Then of course we need to update docs and also give a hint that `_source` field in ES is mandatory (see also [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/mapping-source-field.html) )

Agree.

bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -648,9 +674,20 @@ protected function loadWithoutPriceFilterWithPriceSorting()
         } else {
             throw new \Exception('Not implemented yet - multiple pricing systems are not supported yet');
         }
+
         $raws = [];
+        if ($this->isLoadFromSource()) {
```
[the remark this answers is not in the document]
andreas-gruenwald: ok.

bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -850,9 +887,27 @@ protected function buildQueryConditions($queryFilters, $excludedFieldnames)
      */
     protected function loadElementById($elementId)
     {
-        return $this->tenantConfig->getObjectMockupById($elementId);
+        /** @var ElasticSearch $tenantConfig */
+        $tenantConfig = $this->getTenantConfig();
+        if ($this->isLoadFromSource()) {
+            $mockup = null;
+            if (isset($this->hitData[$elementId])) {
```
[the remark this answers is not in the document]
andreas-gruenwald: Do you have a specific idea in mind?
If we remove the caching feature, then it always needs to be a direct mapping, or are you thinking about some kind of static cache so that the mapping is only done once per (URL) request?

I am not 100% sure if I get your point?

fashxp:
lets finish #5744 and/or #6437 first, then decide on this one.

bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php
fashxp: let leave it there

bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -676,18 +665,10 @@ protected function loadWithoutPriceFilterWithPriceSorting()
...
 
         $raws = [];
-        if ($this->isLoadFromSource()) {
-            $hitDataFull = $this->hitData;
-            $this->hitData = [];
-        }
+
         foreach ($objectRaws as $raw) {
             $id = $raw['o_id'];
             $raws[] = $id;
```
NiklasBr:
```suggestion
            $raws[] = $raw['o_id'];
```

NiklasBr: Too late :)

fashxp: thx anyway!

fashxp:
@andreas-gruenwald  can you please double check my additional changes?

fashxp:
TODO before merging: adding note to Upgrade Notes

fashxp:
> @fashxp Review done. Looks good. Did you test the changes in the ecommerce demo?

sure

PR #6323, opened 2020-04-29
Added "restrict to roles" option in user-select data-type

### LINKED ISSUE #6322: [Feature-Request] User-Select restrict to roles
# Feature Request
## Feature description
In a customer project, I set up an option provider to select users of a specific role.

I recently discovered the data field "User" and thought it would be nice to be able to select a role there.

skochTC:
Resolves #6322

### DIFF
--- a/models/DataObject/ClassDefinition/Data/User.php
+++ b/models/DataObject/ClassDefinition/Data/User.php
@@ -32,6 +32,16 @@ class User extends Model\DataObject\ClassDefinition\Data\Select
      */
     public $unique;
 
+    /** Restrict selection to comma-separated list of roles.
+     * @var string|null
+     */
+    public $restrictTo = null;
+
+    /**
+     * @var int
+     */
+    public $width;
+
     /**
      * @return User
      */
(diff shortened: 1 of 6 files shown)


models/DataObject/ClassDefinition/Data/User.php
```diff
@@ -94,6 +104,23 @@ public function getDataForResource($data, $object = null, $params = [])
...
     {
         $list = new Model\User\Listing();
+
+        if (!empty($this->restrictTo)) {
+            $roles = explode(',', $this->restrictTo);
+            $expr = \Pimcore\Db::getConnection()->createQueryBuilder()->expr();
+
+            foreach ($roles as $roleId) {
+                $list->setCondition(
+                    $expr->orX(
```
skochTC, line 114: I think it would make sense to store this in a repository, but where is the best place to do this in the Pimcore context?

So `User\Listing::getByRoleId()` or `Model\User\Role::getById(1)->getUsers()` or something like that, what do you think?

skoch98: But I think this is a bigger todo for a future issue / PR to improve the roles query between user and role

models/DataObject/ClassDefinition/Data/User.php
```diff
@@ -179,4 +206,51 @@ public function setUnique($unique)
...
     }
+
+    /**
+     * @param string|null $restrictTo
+     *
+     * @return $this
+     */
+    public function setRestrictTo($restrictTo)
+    {
+        /**
```
weisswurstkanone, line 217: where is this comment copied from? this can be removed :-)

skochTC: Okay 👍 I have oriented me on the selection of countries

brusch:
I'm wondering if this should really go into the core, maybe it's better to solve this using a normal select type in combination with an option provider and keep the user select type as it is.

skochTC:
> I'm wondering if this should really go into the core, maybe it's better to solve this using a normal select type in combination with an option provider and keep the user select type as it is.

That's fine for me. For a customer project I have already implemented this as OptionProvoder.

I just thought it would be a nice feature to extend it.

For example, you can provide a select for the responsible person of the sales department in an order.
In my case, we needed that to set the SEO responsible for the product texts.

If you don't want this in the core, you can close the MR :)

brusch:
Thanks a lot for your feedback. 
I'll close this PR for now, but please feel free to continue the discussion, we can reopen the PR anytime if necessary 😊

PR #6324, opened 2020-04-30
[Users] Allow users to have children (not biologically ;-) )

BlackbitNeueMedien:
Currently Pimcore users can only be created within folders. This PR makes the user structure more flexible by allowing users to have children - other users as well as folders.

This can for example be useful in an application where one business department admin is responsible for its users. With this PR you could create the department employees directly under the department admin user.

Example:
We have department admins, e.g. user marketing-admin. This user is responsible for all employees in the marketing department, e.g. marketing-employee-1 and marketing-employee-2.
To structure this responsibility, I would like to add marketing-employee-1 and marketing-employee-2 directly under marketing-admin:
- marketing-admin
  - marketing-employee-1
  - marketing-employee-2

With folders we would need an additional folder marketing-users beside marketing-admin and put marketing-employee-1 and marketing-employee-2 to this folder, so we would have
- marketing-admin
- marketing-users
  - marketing-employee-1
  - marketing-employee-2

Additionally this PR makes the GUI compatible to what is possible via PHP since forever: With PHP you can set the `parentId` of a user object to the ID of another user. Currently everything works fine for this sub-user (login, permissions etc.) except that this sub-user does not get shown in the user settings tree. With this PR it does.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/UserController.php
+++ b/bundles/AdminBundle/Controller/Admin/UserController.php
@@ -71,7 +71,6 @@ class UserController extends AdminController implements EventedControllerInterfa
         $tmpUser = [
             'id' => $user->getId(),
             'text' => $user->getName(),
-            'elementType' => 'user',
             'type' => $user->getType(),
             'qtipCfg' => [
                 'title' => 'ID: ' . $user->getId()
@@ -80,6 +79,7 @@ class UserController extends AdminController implements EventedControllerInterfa
 
         // set type specific settings
         if ($user instanceof User\Folder) {
+            $tmpUser['elementType'] = 'userfolder';
             $tmpUser['leaf'] = false;
             $tmpUser['iconCls'] = 'pimcore_icon_folder';
             $tmpUser['expanded'] = true;
@@ -91,12 +91,21 @@ class UserController extends AdminController implements EventedControllerInterfa
                 $tmpUser['loaded'] = true;
             }
         } else {
+            $tmpUser['elementType'] = 'user';
             $tmpUser['leaf'] = true;
             $tmpUser['iconCls'] = 'pimcore_icon_user';
             if (!$user->getActive()) {
                 $tmpUser['cls'] = ' pimcore_unpublished';
             }
-            $tmpUser['allowChildren'] = false;
+            $tmpUser['allowChildren'] = true;
+            $tmpUser['expanded'] = true;
+
+            if ($user->hasChildren()) {
+                $tmpUser['expanded'] = false;
+                $tmpUser['leaf'] = false;
+            } else {
+                $tmpUser['loaded'] = true;
+            }
             $tmpUser['admin'] = $user->isAdmin();
         }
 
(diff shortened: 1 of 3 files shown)


weisswurstkanone:
"And thinking of the future this could also allow inheritance for users..."

I believe this will definitely happen and be the expectation from the beginning on :-)

weisswurstkanone:
Dangerous thing: Right click on the parent and delete it, guess what happens without any confirmation ...

BlackbitNeueMedien:
Hmm, there actually is a confirmation message when you delete anything with children:
Bildschirmfoto 2020-04-30 um 09 00 18

The JS does not differ between types in https://github.com/pimcore/pimcore/blob/cd10fe8ad587494362d670bd85f1901f0bf938ce/bundles/AdminBundle/Resources/public/js/pimcore/settings/user/panels/abstract.js#L48-L75

weisswurstkanone:
ok, didn't see this one

brusch:
@BlackbitNeueMedien Thanks for your PR! 
To be honest, I think folders are enough to structure users. I also don't think that inheritance for users is a good idea at all, that would be just too much to think of and would cause confusion in combination with the role model. The role concept is well known and users know how to deal with it, so I'd just stick to that to keep thinks simple.

BlackbitNeueMedien:
Ok, then no inheritance - but why not allow to add children to users?

brusch:
Why not using just folders, this is what they are designed for? 😉
I don't see any benefit to be honest. 
Any other opinions?

BlackbitNeueMedien:
The use-case is the follwoing: We have department admins, e.g. user `marketing-admin`. This user is responsible for all employees in the marketing department, e.g. `marketing-employee-1` and `marketing-employee-2`.
To structure this responsibility, I would like to add `marketing-employee-1` and `marketing-employee-2` directly under `marketing-admin`:
- marketing-admin
  - marketing-employee-1
  - marketing-employee-2

With folders we would need an additional folder `marketing-users` beside `marketing-admin` and put `marketing-employee-1` and `marketing-employee-2` to this folder, so we would have
- marketing-admin
- marketing-users
  - marketing-employee-1
  - marketing-employee-2

I do not see any added complexity by allowing users to have sub-users. It is only a GUI effect. Everything else stays the same.

NiklasBr:
I'm not convinced of the merit of this change, with the exception of value inheritance it seems this already works fine with folders and roles. Are we not also adding complexity and increased security risks due to inheritance as well?

BlackbitNeueMedien:
This PR does NOT implement inheritance for users!

Where is there any complexity added? This is only a GUI change. Everything else works as before!

NiklasBr:
That was a response to:
> "And thinking of the future this could also allow inheritance for users..."
>
> I believe this will definitely happen and be the expectation from the beginning on :-)

brusch:
I'll close this for now, feel free to continue the discussion 😊

BlackbitNeueMedien:
@brusch May I ask for the reason? This PR only makes the GUI compatible to what is possible with PHP anyway. The underlying logic stays the same.
Or in other words: There currently is the feature / the bug which allows to set a User as parent of another User via `setParentId()`.

brusch:
I'd classify this as a bug of the API.
Maybe we should also take the change to tag the user classes as `@internal`.

dpfaffenbauer:
I also feel this is unnecessary. Folders are the way to go. 

I wouldn't even classify that using the API to create Users under Users a bug, it doesn't have any side-effects doing that. The UI doesn't allow it cause it makes no real sense really. (BTW: there are more things possible via PHP than the Pimcore UI allows anyway, like converting variants to objects and vice-versa, not a bug to me)

PR #6327, opened 2020-04-30
[Sitemap] Ignore hardlinks

### LINKED ISSUE #6053: [Sitemap]  DocumentTypeFilter ignores Hardlinks
# Bug Report

### Expected behavior
DocumentTypeFilter should check if element is of type page/link/hardlink

### Actual behavior
DocumentTypeFilter ignores hardlinks because it is of the type 'Pimcore\Model\Document\Hardlink\Wrapper\Page' and returns 'page' instead of 'hardlink' on the function $element->getType()

[https://github.com/pimcore/pimcore/blob/master/lib/Sitemap/Document/Filter/DocumentTypeFilter.php](url)

dvesh3:
Resolves #6053

## Additional info  
please see https://github.com/pimcore/pimcore/issues/6053#issuecomment-603162070

### DIFF
--- a/lib/Sitemap/Document/Filter/DocumentTypeFilter.php
+++ b/lib/Sitemap/Document/Filter/DocumentTypeFilter.php
@@ -56,7 +56,7 @@ class DocumentTypeFilter implements FilterInterface
 
     public function canBeAdded(AbstractElement $element, GeneratorContextInterface $context): bool
     {
-        if (!$element instanceof Document) {
+        if (!$element instanceof Document || $element instanceof Document\Hardlink\Wrapper\Page) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


lib/Sitemap/Document/Filter/DocumentTypeFilter.php
```diff
@@ -56,7 +56,7 @@ public function __construct(array $documentTypes = null, array $containerTypes =
 
     public function canBeAdded(AbstractElement $element, GeneratorContextInterface $context): bool
     {
-        if (!$element instanceof Document) {
+        if (!$element instanceof Document || $element instanceof Document\Hardlink\Wrapper\Page) {
```
brusch, line 59: Should we better use `\Pimcore\Model\Document\Hardlink\Wrapper\WrapperInterface` here instead? That would cover all hardlinks, no matter which type, or is there a special reason for `Page` ?

dvesh3: you are right, it should cover all hardlinks...changed it.

PR #6336, opened 2020-05-04
[ElementRouter] Support query parameter and reference type

blankse:
Follow up to #6281


dpfaffenbauer:
does it also respect the base-url? like: https://github.com/pimcore/pimcore/pull/6096/files

blankse:
> does it also respect the base-url? like: https://github.com/pimcore/pimcore/pull/6096/files

@dpfaffenbauer I add the base url

brusch:
Thanks a lot!

PR #6337, opened 2020-05-04
[Admin] Asset Custom Metadata grid - fix data for rows with same name

dvesh3:
Resolves #
Bug fix: If there are metadata rows with same name and different language then onDrop always take the first occurrence in grid.
divesh pim zone __ Pimcore (1)


weisswurstkanone:
please make sure that there will be no conflicts with #6333

dvesh3:
I verified, there is none 😉

PR #6342, opened 2020-05-05
[Data Objects] [Inheritance] Values at child not getting inherited from parent.

### LINKED ISSUE #6338: [Data Objects] [Inheritance] Values at child not getting inherited from parent.
**Expected behavior**

A child object should be able to inherit the values from its parent.

**Actual behavior**

Fields are not shown marked as inherited in the UI but showing as copied instead.
Noticed inheritance issues on Number, QuantityValue and Select types.

Please see screenshot attached below.

**Steps to reproduce**

- Create a new class and allow inheritance.
- Add a text, number, quantityValue and select field in definition. 
- Create an object of the same class. Add some data to all the fields. Save or Save&Publish.
- Create a child object.
- The data for respective fields will not inherit but gets copied from parent to the child.

inheritance issue

weisswurstkanone:
resolves #6338

### DIFF
--- a/models/DataObject/Traits/DefaultValueTrait.php
+++ b/models/DataObject/Traits/DefaultValueTrait.php
@@ -77,19 +77,6 @@ trait DefaultValueTrait
                 $class = $owner->getObject()->getClass();
             }
 
-            if ($class && $class->getAllowInherit()) {
-                $params = [];
-
-                try {
-                    $data = $owner->getValueFromParent($this->getName(), $params);
-                    if (!$this->isEmpty($data)) {
-                        return $data;
-                    }
-                } catch (InheritanceParentNotFoundException $e) {
-                    // no data from parent available, use the default value
-                }
-            }
-
             if ($object !== null && !empty($this->defaultValueGenerator)) {
                 $defaultValueGenerator = DefaultValueGeneratorResolver::resolveGenerator($this->defaultValueGenerator);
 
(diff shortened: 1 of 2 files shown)


models/DataObject/Traits/DefaultValueTrait.php
```diff
@@ -130,7 +117,27 @@ protected function handleDefaultValue($data, $object = null, $params = [])
...
+            $configuredDefaultValue = $this->doGetDefaultValue($object, $context);
+            if (!$this->isEmpty($configuredDefaultValue)) {
+
+                if ($class && $class->getAllowInherit()) {
+                    $params = [];
+
+                    try {
+                        $data = $owner->getValueFromParent($this->getName(), $params);
+                        if (!$this->isEmpty($data)) {
+                            return $data;
```
dvesh3, line 133: if parent has value then it should return null so as to inherit the value form parent. right?

weisswurstkanone: No. It should return the value from the parent as this will go directly into the resource

dvesh3:
when using default value then null values does not show inherited value in UI even though db has correct values.
image

image

models/DataObject/Traits/DefaultValueTrait.php
dvesh3: please add a note in docs, if not there :)

PR #6344, opened 2020-05-06
[Data objects] Set default value for database column

BlackbitNeueMedien:
Currently data object tables get `NULL` as default value (or more exactly, no default value is specified while the columns allow `NULL` values).
This leads to the problem that when you add a new field with a default value this field is set to `NULL` for all existing objects. #5594 changed this behaviour for newly created objects but for already existing ones the problem still exists. 
Quoted from #4600: 
> By not saving default values to the database a lot of problems arise, e.g. the value does not appear in exports. When you then open the object to check if the value really has not been set, it gets even more confusing because there the default value gets displayed.

This PR is far from being finished, I only want to check other opinions to determine if it is worth to continue.

### DIFF
--- a/models/DataObject/ClassDefinition/Dao.php
+++ b/models/DataObject/ClassDefinition/Dao.php
@@ -168,7 +168,7 @@ class Dao extends Model\Dao\AbstractDao
                         } elseif ($value->getColumnType()) {
                             $defaultValue = '';
                             if(method_exists($value, 'getDefaultValue') && $value->getDefaultValue() !== '') {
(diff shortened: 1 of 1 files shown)


BlackbitNeueMedien:
Any opinions on that @brusch, @weisswurstkanone, @dvesh3 ?

fashxp:
see also https://github.com/pimcore/pimcore/issues/6376

brusch:
Don't think this is a good idea at all, as @fashxp referenced already above this will cause some major issues in combination with inheritance and probably some other places such as default value generators ... 
I'll close this PR for now, but please feel free to continue the discussion. Thanks!

PR #6346, opened 2020-05-06
[Quantity values] implement "isEqual()" for quantity value fields

BlackbitNeueMedien:


models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -708,4 +708,17 @@ public function getFilterCondition($value, $operator, $params = [])
 
         return implode(' OR ', $conditions);
     }
+
+    public function isEqual($oldValue, $newValue)
+    {
+        if(!$oldValue instanceof Model\DataObject\Data\QuantityValue) {
```
weisswurstkanone, line 714: can the new value be null?

dpfaffenbauer: yes, but I think what is missing is that the old and new value can be null, therefore it is still equal

BlackbitNeueMedien: ? I do not understand the question. Of course it can as this is a public method - old and new value can be of any type. Or what do you mean?

BlackbitNeueMedien: Added a check for old and new value being `null`

models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -708,4 +708,21 @@ public function getFilterCondition($value, $operator, $params = [])
 
         return implode(' OR ', $conditions);
     }
+
```
dvesh3, line 711: please add phpdocs

BlackbitNeueMedien: Done. Copied from the parent method.

dvesh3:
thanks a lot!

PR #6353, opened 2020-05-07
[DataObject] ObjectBrick/FieldCollection - Add Maximum limit validation in API

### LINKED ISSUE #6350: [Bug] Validation by updating objectbricks via the grid view
# Bug Report
I'm not sure if this is a real bug or if it's just deliberate. Please give me some feedback about this.

### Expected behavior
A validation message should be displayed.

### Actual behavior
The object is stored invalidly.

### Steps to reproduce
If an objectbrick is assigned in the object, you can still assign another one in the GRID view, but only one object brick is allowed according to the class definition. 
When the edit view is opened, you will only see one object brick and the message "limit reached" will appear immediately.

I have found this line that disables the validation:
https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L1719

I think instead of `||` you mean `&&`?

dvesh3:
Resolves #6350

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
@@ -1,5 +1,8 @@
 # Upgrade Notes
 
+## 6.7.0
+- If you are using the specific settings 'max. items' option for ObjectBricks & Fieldcollections on your class definition, then API will validate the max limit on save() calls from now on.
+
 ## 6.6.2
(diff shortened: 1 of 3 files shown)


weisswurstkanone:
Do not see how this works, omitMandatory check is true so no check is performed.
Is there maybe a dependency on unmerged PR https://github.com/pimcore/pimcore/pull/6354 ?

Besides that, I am not sure if this shouldn't be checked regardless of the omitMandatory check as this doesn't have anything to do with being mandatory or not.

dvesh3:
Yes,`max limit should be checked irrespective of omitMandatoryCheck check. Changed it now! :)

PR #6377, opened 2020-05-13
[Ecommerce][Indexservice] Utilize transactions for avoiding deadlocks

### LINKED ISSUE #5499: [Ecommerce][IndexService] fillupPreparationQueue can produce deadlocks
Under certain cercumstances the following exception might happen:

```
An exception occurred while executing 'UPDATE ecommerceframework_productindex_store_elastic SET in_preparation_queue = 1 WHERE tenant = ? AND o_id IN (35795,36015,36214,36343,36344,35796,36215,36016)' 
```

This happens when I save an object and the fillupPreparationQueue statements get's executed by the event listener. If the process queue cronjob runs for the same items, deadlocks might occur.

jremmurd:
# Bugfix

Tries to fix #5499 

As discribed in the issue, we experience deadlocks in the store table as well.

This PR adds a method `executeTransactionalQuery` which tries to commit a query for n times (default 3) with a given sleep timeout (default .5 seconds). 
There may be some more places in the worker class(es) where it could help to utilize this method, I added the calls to our pain points only.

The solution is inspired by `AbstractObject::save()` in https://github.com/pimcore/pimcore/blob/master/models/DataObject/AbstractObject.php#L664

### DIFF
--- a/bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractBatchProcessingWorker.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractBatchProcessingWorker.php
@@ -531,10 +531,9 @@ abstract class AbstractBatchProcessingWorker extends AbstractWorker implements B
     }
     
     /**
-     * @param string $query
-     * @param null $params
+     * @param \Closure $fn
      * @param int $maxTries
-     * @param int $waitTimeout
+     * @param float $sleep
      * @return bool
      * @throws \Exception
      */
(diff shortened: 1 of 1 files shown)


fashxp:
did that solve your deadlock issues? 
what bothers me is, why the deadlocks can occur in the first place.

jremmurd:
> did that solve your deadlock issues?
> what bothers me is, why the deadlocks can occur in the first place.

@fashxp I am totally on your side, that this is only a workaround. I had a short call with @weisswurstkanone some weeks ago and we concluded that for now I should provide a PR with this "solution".

fashxp:
thx

PR #6380, opened 2020-05-13
[PHPStan] Fix current errors after phpstan update

blankse:
After Update > 0.12.24 PHPStan merge inherited PHPDocs
See https://github.com/phpstan/phpstan/releases/tag/0.12.24

So some new errors appears.

### DIFF
--- a/models/Document/Hardlink/Wrapper.php
+++ b/models/Document/Hardlink/Wrapper.php
@@ -180,11 +180,11 @@ trait Wrapper
     }
 
     /**
-     * @throws \Exception
+     * @return \Exception
      */
-    protected function getHardlinkError()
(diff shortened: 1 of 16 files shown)


models/Document/Hardlink/Wrapper.php
```diff
@@ -172,9 +182,9 @@ public function hasChildren($unpublished = false)
     /**
      * @throws \Exception
```
dvesh3, line 183: @return \Exception

also, could you please fix the typo in exception message? thanks!

blankse: I fixed it

dvesh3:
@blankse thanks a lot!

PR #6384, opened 2020-05-13
Removes limitation on password length

NiklasBr:
It makes no sense to have an upper limit* on password length. Anyway, it was never enforced in the backend if you did a `$account->setPasswordField('somestring');`, only in the UI, and even then it never actually prevented submitting the data.

This PR also removes an unneccesary duplicate declaration of inputType = password.

Hashes which are harder to detect are generally always in the HEX 00 to FF `a-f0-9` character range and of a set length (4 bits per char, e.g. SHA512 = 512/4=128), only checking for >= 32 is too weak and makes very little sense in the world of password managers and diceware passwords.

\* Well, technically maybe a password in the kilobyte range or larger would be uneccesarily large and a [problem in itself](https://arstechnica.com/information-technology/2013/09/long-passwords-are-good-but-too-much-length-can-be-bad-for-security/), but only 32 non-multibyte chars is way too low.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/password.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/password.js
@@ -32,7 +32,6 @@ pimcore.object.tags.password = Class.create(pimcore.object.tags.abstract, {
             name: this.fieldConfig.name,
             componentCls: "object_field",
             inputType: "password",
-            maxLength: 30,
             listeners: {
                 afterrender: function (cmp) {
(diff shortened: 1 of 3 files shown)


weisswurstkanone:
The column type is still

    /**
     * Type for the column
     *
     * @var string
     */
    public $columnType = 'varchar(190)';


?

NiklasBr:
True, "[255 characters would be a good choice](https://www.php.net/manual/en/function.password-hash.php)", have changed.

brusch:
Thanks a lot!

PR #6394, opened 2020-05-15
Avoid Pimcore::getContainer() in Controller/Commands

blankse:
`Pimcore::getContainer()` should be avoided if possible

### DIFF
--- a/bundles/CoreBundle/Command/Targeting/MigrateElementNamesCommand.php
+++ b/bundles/CoreBundle/Command/Targeting/MigrateElementNamesCommand.php
@@ -19,7 +19,7 @@ namespace Pimcore\Bundle\CoreBundle\Command\Targeting;
 
 use Doctrine\DBAL\Query\QueryBuilder;
 use Pimcore\Console\AbstractCommand;
-use Pimcore\Db;
+use Pimcore\Db\ConnectionInterface;
 use Pimcore\Document\Tag\NamingStrategy\NamingStrategyInterface;
 use Pimcore\Document\Tag\NamingStrategy\NestedNamingStrategy;
 use Pimcore\Model\Document\Targeting\TargetingDocumentInterface;
@@ -31,7 +31,7 @@ use Symfony\Component\Console\Question\ConfirmationQuestion;
 class MigrateElementNamesCommand extends AbstractCommand
 {
     /**
-     * @var Db\Connection
+     * @var ConnectionInterface
      */
     private $db;
 
@@ -50,8 +50,9 @@ class MigrateElementNamesCommand extends AbstractCommand
      */
     private $runCommand = true;
 
-    public function __construct(NamingStrategyInterface $namingStrategy)
+    public function __construct(ConnectionInterface $connection, NamingStrategyInterface $namingStrategy)
     {
+        $this->db = $connection;
         $this->namingStrategy = $namingStrategy;
 
         parent::__construct();
@@ -106,8 +107,6 @@ class MigrateElementNamesCommand extends AbstractCommand
 
             return 1;
         }
-
-        $this->db = \Pimcore::getContainer()->get(Db\ConnectionInterface::class);
 
         $qb = $this->buildQuery();
         $stmt = $qb->execute();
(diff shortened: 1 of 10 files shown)


bundles/EcommerceFrameworkBundle/Controller/AdminOrderController.php
```diff
@@ -178,19 +185,27 @@ public function listAction(Request $request)
             'paginator' => $paginator,
             'pimcoreUser' => \Pimcore\Tool\Admin::getCurrentUser(),
             'listPricingRule' => new \Pimcore\Bundle\EcommerceFrameworkBundle\PricingManager\Rule\Listing(),
-            'defaultCurrency' => \Pimcore\Bundle\EcommerceFrameworkBundle\Factory::getInstance()->getEnvironment()->getDefaultCurrency(),
-            'formatter' => \Pimcore::getContainer()->get('pimcore.locale.intl_formatter')
+            'defaultCurrency' => Factory::getInstance()->getEnvironment()->getDefaultCurrency(),
+            'formatter' => $formatter,
```
dvesh3, line 189: trailing comma

bundles/EcommerceFrameworkBundle/Controller/AdminOrderController.php
```diff
@@ -354,15 +368,15 @@ public function detailAction(Request $request)
             'arrCustomerAccount' => $arrCustomerAccount,
             'geoAddressDelivery' => $geoAddressDelivery,
             'pimcoreSymfonyConfig' => $pimcoreSymfonyConfig,
-            'formatter' => \Pimcore::getContainer()->get('pimcore.locale.intl_formatter'),
-            'locale' => \Pimcore::getContainer()->get('pimcore.locale')
+            'formatter' => $formatter,
+            'locale' => $localeService,
```
dvesh3, line 372: trailing comma

bundles/EcommerceFrameworkBundle/Controller/AdminOrderController.php
jdreesen: What's the problem? Is it against the code style?

blankse: It is a good. Also symfony do this:
https://symfony.com/doc/current/contributing/code/standards.html#structure
`Add a comma after each array item in a multi-line array, even after the last one;`

So there is no change in the line above in diffs if someone add a line.

dvesh3: I was just concerned about neat code and was not aware that it has a benefit. Good to know that Symfony also recommends it 😉 thanks!

bundles/CoreBundle/Command/Targeting/MigrateElementNamesCommand.php
```diff
@@ -107,8 +108,6 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
-        $this->db = \Pimcore::getContainer()->get(Db\ConnectionInterface::class);
```
brusch, line 110: This is intentional, because if the DB connection isn't set up (e.g. before running the installer) this change will lead to errors (e.g. when running clear:cache). 
But maybe we just change it to `Pimcore\Db::get()` which is shorter and does the same :)

dpfaffenbauer: not sure, but does it also happen when you lazy load the command? https://symfony.com/doc/current/console/commands_as_services.html#lazy-loading

brusch: Unfortunately yes. At least last time i tried it.

dpfaffenbauer: alright. Oro Platform has similar issues, they fixed it by wrapping and catching things inside a special service: https://github.com/oroinc/platform/blob/master/src/Oro/Bundle/EntityBundle/Tools/SafeDatabaseChecker.php

That might work for these use-cases, I am just afraid people might use it in wrong scenarios as well. And: for now Pimcore::getDb might really be the best solution here.

brusch: @dpfaffenbauer thanks for your feedback! 
@blankse can you please update the PR accordingly? 

Thanks a lot guys!

blankse: I updated the PR

brusch:
Thanks a lot!

PR #6397, opened 2020-05-15
Typo fixed and new function getAllCartsForCurrentUser

MHundeloh:
Typo fixed and new function getAllCartsForCurrentUser to make it extendable

  
  

Makes it more easy to extend initSavedCarts method by overwriting the new getAllCartsForCurrentUser method. Otherwise you have to copy the whole method to change the the current user functionality.
## Additional info  
We need this in our current project where not the logged in user is related to the cart but another entity called accountUser which is also held in environment.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/CartManager/MultiCartManager.php
+++ b/bundles/EcommerceFrameworkBundle/CartManager/MultiCartManager.php
@@ -97,10 +97,7 @@ class MultiCartManager implements CartManagerInterface
 
     protected function initSavedCarts()
     {
-        $classname = $this->getCartClassName();
-
-        /* @var CartInterface[] $carts */
-        $carts = $classname::getAllCartsForUser($this->environment->getCurrentUserId());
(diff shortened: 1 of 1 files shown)


bundles/EcommerceFrameworkBundle/CartManager/MultiCartManager.php
```diff
@@ -130,6 +127,17 @@ protected function initSavedCarts()
...
     }
 
+    /**
+     * @return CartInterface[]|null
+     */
+    protected function getAllCartsForCurrentUser()
+    {
+        $classname = $this->getCartClassName();
+
+        /* @var CartInterface[] $carts */
```
blankse, line 137: There is no `$carts` variable. I think we can drop this comment

MHundeloh: Yes indeed!

fashxp:
thx

Issue #1469, opened 2017-04-28
Clone support for Documents, Assets & Objects


brusch:
Consistent support of  `clone` for all pimcore elements.

brusch:
+ dedicated docs page 'cloning elements'

PR #6398, opened 2020-05-15
implement custom copy function for class-definition types

dpfaffenbauer:
similar to VersionMarshal and RecycleBinMarshal just for copy

### DIFF
--- a/models/DataObject/ClassDefinition/Data/CustomDataCopyInterface.php
+++ b/models/DataObject/ClassDefinition/Data/CustomDataCopyInterface.php
@@ -27,5 +27,5 @@ interface CustomDataCopyInterface
      *
      * @return mixed
      */
-    public function createDataCopy($object, $data);
+    public function createDataCopy(Concrete $object, $data);
 }
--- a/models/DataObject/ClassDefinition/Data/CustomDataCopyInterface.php
+++ b/models/DataObject/ClassDefinition/Data/CustomDataCopyInterface.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @category   Pimcore
+ * @package    Element
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Model\DataObject\ClassDefinition\Data;
+
+use Pimcore\Model\DataObject\Concrete;
+
+interface CustomDataCopyInterface
+{
+    /**
+     * @param Concrete $object
+     * @param mixed $data
+     */
+    public function createDataCopy($object, $data);
+}
--- a/models/DataObject/ClassDefinition/Data/CustomDataCopyInterface.php
+++ b/models/DataObject/ClassDefinition/Data/CustomDataCopyInterface.php
@@ -24,6 +24,8 @@ interface CustomDataCopyInterface
     /**
      * @param Concrete $object
      * @param mixed $data
+     *
+     * @return mixed
      */
     public function createDataCopy($object, $data);
 }
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -32,9 +32,13 @@ use Pimcore\Model;
 use Pimcore\Model\Asset;
 use Pimcore\Model\DataObject;
 use Pimcore\Model\DataObject\AbstractObject;
+use Pimcore\Model\DataObject\ClassDefinition\Data;
+use Pimcore\Model\DataObject\Concrete;
 use Pimcore\Model\Dependency;
 use Pimcore\Model\Document;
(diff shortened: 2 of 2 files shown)


models/DataObject/ClassDefinition/Data/CustomDataCopyInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+use Pimcore\Model\DataObject\Concrete;
+
+interface CustomDataCopyInterface
+{
+    /**
+     * @param Concrete $object
+     * @param mixed $data
+     */
+    public function createDataCopy($object, $data);
```
weisswurstkanone, line 28: return type ?

dpfaffenbauer: added

models/DataObject/ClassDefinition/Data/CustomDataCopyInterface.php
```diff
@@ -0,0 +1,31 @@
...
+
+interface CustomDataCopyInterface
+{
+    /**
+     * @param Concrete $object
+     * @param mixed $data
+     *
+     * @return mixed
+     */
+    public function createDataCopy($object, $data);
```
brusch, line 30:
```suggestion
    public function createDataCopy(Concrete $object, $data);
```

dpfaffenbauer:
done

dpfaffenbauer:
Any ETA for when this will be merged? I wanna release CoreShop 2.2.2 and it depends on this...

fashxp:
what happens, if you clone the object, set ID to null and save it again?

dpfaffenbauer:
not sure what the question has to do with this PR?

fashxp:
If I see an interface `CustomDataCopyInterface` I probably would expect, that also cloning is covered and it might cause confusion if it is not. 
Maybe there is a solution that also covers the clone case? Or how do you solve this on your end?

dpfaffenbauer:
I don't clone objects on my end, I don't see a reason to do that. Anyhow: if you need to, use the Service::clone function that uses the CustomDataCopyInterface

dpfaffenbauer:
are there any issues with this now?

brusch:
Well, yes, we're thinking about the right solution ... I also think that someone would assume that this also applies for the  `clone` command. Wouldn't it be better to implement `__clone()` properly?

dpfaffenbauer:
so you mean we should also use DeepClone for `__clone`? The main problem is, that I cannot copy any Objects that are linked to Doctrine Entities right now since DeepClone is used for copy. Before it worked fine.

albertmueller:
i have also issues with it, can someone please merge this PR?

dpfaffenbauer:
@brusch Can we please proceed with this, we have lots of issues with copying stuff.

brusch:
@dpfaffenbauer I still think we should make `$foo = clone $bar` possible rather than relying on a service for copying objects.

dpfaffenbauer:
and how do you think that should work? you changed every internal clone call to DeepCopy, which now causes other problems. this is set to fix some of them and you want to go back to clone? which also had problems?

brusch:
Well, you can still use DeepCopy in `__clone`, right? 
Using DeepCopy here and there was more a workaround than a proper solution, I guess. Long term I'd really like to see `__clone()` implemented propertly, so that using `clone` returns the expected result.

dpfaffenbauer:
@brusch And how do you wanna use DependencyInjection within models in `__clone`?

brusch:
@dpfaffenbauer for waht, do you have an example?

dpfaffenbauer:
For exactly what this PR tries to solve. You might up implementing super custom stuff and have to 'clone' it differently or in a custom way. Like how I explained above to create a service that creates the DeepCopy instance and you can hook into and adapt it if needed.

BTW: This PR is not about `__clone` per-se, but about the current Implementation, if you really want to move stuff to `__clone` it is nothing this PR should care about.

brusch:
This doesn't answer my question, but anyway.

dpfaffenbauer:
Ok, let me try again. We have lots of custom stuff in CoreShop that keeps making us troubles with DeepCopy. That's why I think it would be good to have a service that creates the DeepCopy instance and throws it into the EventDispatcher where I can listen to and add certain Filters to solve my problems outside the Core. When you do the same with `__clone` that would not work anymore since the Model can not have any Dependencies injected. (Besides using `Pimcore::getContainer` and that is not a good practice) 

Basically I quite like the idea of moving the responsibility of copying certain properties of a DataObject into the corresponding FieldDefinition Tag Class. That way, you have always control of how you copy stuff.

But also: This PR is not about moving DeepCopy to `__clone`, this PR is about having the possibility to add custom `copy` functions for custom DataObject FieldDefinition Tags in order to deal with that as well. Same as we have for Versions or RecycleBin. 

I do understand the convenience of `__clone`, but it is certainly not the Topic of this PR.

dpfaffenbauer:
Why did you close this now? There is still an issue with this, what was the point of using DeepCopy if you wanna go back to clone again? You introduced this feature couple of years ago: https://github.com/pimcore/pimcore/issues/1469

Also, the new behaviour is targeted for 7.0, this PR was meant for 6.6....

brusch:
Thanks for your detailed explanation! 
I guess this doesn't hurt for now, but we should already think about deprecating it for v6, even if it was just introduced. But we can decide about that later on.

PR #6399, opened 2020-05-15
[Asset] fix download-filename when file extension of Asset is uppercased

bebuecherl:

* Modifies Admin/AssetController to replace the Assets file extension with case insensitive regexp
* Modifies Image/Thumbnail/Processer to create tmp files with correct filename by replacing the original file extension with case insensitive regexp

## Steps to reproduce

* Upload an Asset with an uppercased file-extension (e.g. `picture.TIF`)
* Download the "Web-Format" of this Asset

## Expected

* Downloaded file is named `picture.jpeg`

## Actual Behaviour

* Downloaded file is named `picture.tif`


brusch:
Thanks a lot!

PR #6400, opened 2020-05-15
[DataObjects] Behavior of default values in combination with inherita…

### LINKED ISSUE #6376: [Discussion][DataObjects] Behavior of default values in combination with inheritance
related to #4600 #6338 #6342

Right now default value, behavior is as follows: 
1) no inheritance: **default value** is persisted to store/query table on create
2) with inheritance and NO parent value: **default value** is persisted  to store/query table on create
3) with inheritance and existing parent value: **parent value** is persisted to store/query table on create, inheritance is broken.

For my understanding, 3) is confusing. More logical behaviors would be

3.a) with inheritance and existing parent value: **no value** is persisted to store table, **inherited value** is persisted to query table, inheritance is active

3.b) with inheritance and existing parent value: **default value** is persisted to store/query table on create, inheritance is broken.

What are your thoughts? How should it be?

weisswurstkanone:
resolves #6376

### DIFF
--- a/doc/Development_Documentation/05_Objects/01_Object_Classes/01_Data_Types/README.md
+++ b/doc/Development_Documentation/05_Objects/01_Object_Classes/01_Data_Types/README.md
@@ -155,7 +155,23 @@ See sub-pages of this page for detail documentation of different data types.
 
 ### Default values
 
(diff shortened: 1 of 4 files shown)


brusch:
@weisswurstkanone since this is a BC-break we should put a note into to the upgrade notes. Thanks!

PR #6402, opened 2020-05-15
[Workflow] Fix saving document workflow state

### LINKED ISSUE #6401: [Workflow] Workflow state incorrectly saved for documents if an object with the same id also has a workflow state
# Bug Report

### Expected behavior

When I use the workflow actions to change my snippet workflow state I expect this to actually change and be reflected in the GUI

### Actual behavior

The workflow state for my snippet stays the same while an object with the same id does have its workflow state changed

### Steps to reproduce

* Have an object and a document with the same id both with a workflow state (in the `element_workflow_state` table)
* Apply your workflow action to the document. You'll see it's still at the old workflow state even if you refresh it.
* Refresh the object and see that it has changed to the new workflow state instead

I will send in a merge request to fix this

MVermue:
Resolves #6401


fashxp:
thx!

PR #6403, opened 2020-05-15
[Admin] XSS Protection for grids

dvesh3:

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/metadata/predefined.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/metadata/predefined.js
@@ -137,11 +137,11 @@ pimcore.settings.metadata.predefined = Class.create({
             {text: t("description"), sortable: true, dataIndex: 'description',
                 getEditor: function() { return new Ext.form.TextArea({}); },
                 renderer: function (value, metaData, record, rowIndex, colIndex, store) {
-                                if(empty(value)) {
-                                    return "";
-                                }
-                                return nl2br(value);
-                           }
+                    if (empty(value)) {
+                        return "";
+                    }
+                    return Ext.util.Format.htmlEncode(nl2br(value));
+                }
             },
             {text: t("type"), width: 90, sortable: true,
                 dataIndex: 'type',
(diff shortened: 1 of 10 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/settings/metadata/predefined.js
```diff
@@ -137,11 +137,11 @@ pimcore.settings.metadata.predefined = Class.create({
...
                 renderer: function (value, metaData, record, rowIndex, colIndex, store) {
-                                if(empty(value)) {
-                                    return "";
-                                }
-                                return nl2br(value);
-                           }
+                    if (empty(value)) {
+                        return "";
+                    }
+                    return Ext.util.Format.htmlEncode(nl2br(value));
```
brusch, line 143: The order of `htmlEncode()` and `nl2br()` is wrong here. 
The current code would produce: `&lt;br /&gt;` instead of `<br />`
So `nl2br(Ext.util.Format.htmlEncode(value));` would be the correct order 😊

bundles/AdminBundle/Resources/public/js/pimcore/settings/properties/predefined.js
```diff
@@ -107,12 +107,12 @@ pimcore.settings.properties.predefined = Class.create({
...
-                                    if(empty(value)) {
-                                        return "";
-                                    }
-                                    return nl2br(value);
-                               }
+                renderer: function (value, metaData, record, rowIndex, colIndex, store) {
+                    if(empty(value)) {
+                        return "";
+                    }
+                    return Ext.util.Format.htmlEncode(nl2br(value));
```
brusch, line 114: Same order issue here.

dvesh3: Done :)

bundles/AdminBundle/Resources/public/js/pimcore/settings/metadata/predefined.js
dvesh3: Done :)

PR #6404, opened 2020-05-16
[Feature] Application Logger for class definition save actions

NiklasBr:
Updated PR for #6242


dvesh3:
@blankse could you please fix the tests. thanks!

blankse:
@dvesh3 I think you mean @NiklasBr :)

dvesh3:
@blankse yes, thanks for correction! missed you both on weekend 😄

NiklasBr:
I'll look at it @dvesh3

NiklasBr:
@dvesh3 Feel free to have a look again.

brusch:
We discussed this internally and for now we came to the conclusion that we're not going to put any core log entries into the application logger, that should stay (for now) exclusive for application related stuff. 
We agree that this would be a handy feature, but we don't think the application logger (as it is designed at the moment) is the right place for that. If we go in this direction, this feature should be also implemented for other topics.

brusch:
Thanks anyway! Maybe we should also create a follow-up issue for that feature. 🤔

Issue #3409, opened 2018-10-01
Select/multiselect data types with localized and centralized select options


markus-moser:
Select/multiselect fields only support a single value and label for one selected option and the provided select options need to be defined directly in the class editor (if no options provider is used). There a several problems with this approach in many projects:

- In many constellations it's important to localize the selected value. It's possible to do this via website translations with a special naming scheme (e.g. optionsvalue.my-option). But this could be quite challenging and a lot of work in a PIM system with hundreds of attributes and thousands of attribute values. 

- Sometimes the same select options need to be used in different fields. For example when object bricks are used for different product categories the same field could be needed in two different bricks with the same options.

- Often some users should have the right to maintain the provided select options (add new ones etc.). At the moment this is only possible when the user is allowed to use the class editor which is quite dangerous.

It's possible to use a options provider to create solutions for these kind of problems but a more standardized and out of the box way would be a really good improvement as option providers have several limitations:

- It's more work to create option providers than to use a good standard feature. :-)

- Export/import/grid view works not that perfect. A good out of the box solution could provide good solutions for that. For example with a special operator which can export the translated value. The grid view itself could show the translated value in the right language etc.

- If combined with Pimcore's output channel bundle (https://github.com/pimcore/output-data-config-toolkit) things become really nasty when using option providers.

Therefore I suggest a solution similar to this:

1.  Add a global "Select options" section similar to for example the classification store group/key definitions. There it might be possible to define for example select value groups and within the groups localized select values. A new user right can be added to define who is allowed to edit the select options.
2. Add 2 new data types for select/multiselect based on these select options and groups. In the class editor of the data type the used select options group needs to be configurable.
3. Add grid view operators for the new data types.

dpfaffenbauer:
Isn't that a bit overkill for just selects? Why not use something like https://github.com/coreshop/CoreShop/blob/master/src/CoreShop/Bundle/PimcoreBundle/CoreExtension/DynamicDropdown.php where the select value is from another Object?

markus-moser:
If done it similarly to that in several projects. But remember a data model with 400 select/multiselect attributes in 20 object bricks. It would get really big and additionally it's not that easy to create for example a CSV exporter via the object grid. I think the dynamic dropdown is the overkill for this kind of problem. I have the same situation in nearly each PIM project with a lot of attributes.

dpfaffenbauer:
👍 ok, the export/import argument makes sense to me ;)

[...]

PR #6406, opened 2020-05-17
[Tags] Translations for Tags #2839

### LINKED ISSUE #2839: [Tags] Translations for Tags
Tags should be optionally translatable to better support multi-locale setups. 

Translatable in the _Tag Configuration_ similar to how it is made with shared translations. 

image

roland4432:
Resolves #2839

## Additional info  

Added translation capability of tags as suggested in #2839 and stored the translation in a serialized array in 1 field to be able to add translations later with adding new languages. Will default to tagname if translation could not be found.


bundles/AdminBundle/Controller/Admin/TagsController.php
```diff
@@ -95,6 +95,43 @@ public function updateAction(Request $request)
...
+     */
+    public function updateTranslationForTagAction(Request $request){
+        $success = false;
+
+        if($tag = Tag::getById($request->get('id'))){
+            $translations = $tag->getTranslations();
+            $translations[$request->get('language')] = $request->get('translation');
+
+            $tag->setTranslations(serialize($translations));
+            $tag->setModificationDate(date('U'));
```
jdreesen, line 110: As the `modificationDate` property is of type `int`, probably `time()` should be used instead, as it returns the timestamp as an `int`, while `date('U')` returns it as a `string`.

```suggestion
            $tag->setModificationDate(time());
```

models/Element/Tag.php
```diff
@@ -55,6 +57,26 @@ class Tag extends Model\AbstractModel
...
+    public $creationDate;
+
+    /**
+     * @var int
+     */
+    public $modificationDate;
+
+    public function __construct() {
+        $this->creationDate = (int) date('U');
+        $this->modificationDate = (int) date('U');
```
jdreesen, line 77:
```suggestion
        $this->creationDate = time();
        $this->modificationDate = time();
```

models/Element/Tag.php
```diff
@@ -163,6 +185,7 @@ public static function getElementsForTag(
     public function save()
     {
         $this->correctPath();
+        $this->modificationDate = (int) date('U');
```
jdreesen, line 188:
```suggestion
        $this->modificationDate = time();
```

models/Element/Tag.php
```diff
@@ -319,4 +342,51 @@ public function correctPath()
...
+    /**
+     * @param int $modificationDate
+     */
+    public function setModificationDate($modificationDate) {
+        $this->modificationDate = $modificationDate;
+    }
+
+    public function __toString(){
+        $currentLanguage = \Pimcore::getContainer()->get('pimcore.locale')->getLocale();
+        return $this->getTranslations()['_' . $currentLanguage] ?? $this->name;
```
jdreesen, line 390: This won't work if there are no translations at all, because in this case `$this->getTranslations()` returns an empty array while `?? $this->name` only gets called for `null`.

```suggestion
        return $this->getTranslations()['_' . $currentLanguage] ?: $this->name;
```

models/Element/Tag.php
```diff
@@ -319,4 +342,51 @@ public function correctPath()
             $this->idPath = '/';
         }
     }
+
+    /**
+     * @return mixed
```
jdreesen, line 347:
```suggestion
     * @return array
```

or 

```suggestion
     * @return string[]
```

models/Element/Tag.php
```diff
@@ -319,4 +342,51 @@ public function correctPath()
...
+     */
+    public function getTranslations() {
+        return $this->translations ? unserialize($this->translations) : [];
+    }
+
+    /**
+     * @param mixed $translations
```
jdreesen, line 354:
```suggestion
     * @param array $translations
```

or 

```suggestion
     * @param string[] $translations
```

bundles/AdminBundle/Controller/Admin/TagsController.php
```diff
@@ -95,6 +95,43 @@ public function updateAction(Request $request)
...
+     * @Route("/update-translation-for-tag")
+     */
+    public function updateTranslationForTagAction(Request $request){
+        $success = false;
+
+        if($tag = Tag::getById($request->get('id'))){
+            $translations = $tag->getTranslations();
+            $translations[$request->get('language')] = $request->get('translation');
+
+            $tag->setTranslations(serialize($translations));
```
jdreesen, line 109: I think the serializing should be done in the setter, as the unserializing is done in the getter, too.

```suggestion
            $tag->setTranslations($translations);
```

bundles/AdminBundle/Controller/Admin/TagsController.php
jdreesen: Actually, this call in not even needed here, because the `modificationDate` gets updated in ths `save()` method anyways.

```suggestion
```

models/Element/Tag.php
```diff
@@ -319,4 +342,51 @@ public function correctPath()
...
+     */
+    public function setTranslations($translations) {
+        $this->translations = $translations;
```
jdreesen, line 357: The fact that the values are serialized internally should be transparent for the user of this class.

```suggestion
        $this->translations = serialize($translations);
```

models/Element/Tag.php
```diff
@@ -55,6 +57,26 @@ class Tag extends Model\AbstractModel
      */
     public $parent;
 
+    /**
+     * @var string
+     */
+    public $translations;
```
jdreesen, line 63: Maybe those properties should be `private` or `protected` because at least this one contains a serialized string and thus should be accessed through getters/setters.

models/Element/Tag.php
```diff
@@ -18,6 +18,8 @@
 namespace Pimcore\Model\Element;
 
 use Pimcore\Model;
+use Pimcore\Tool;
+use Symfony\Component\HttpFoundation\Request;
```
jdreesen, line 22: I don't think this is used here:
```suggestion
```

bundles/AdminBundle/Controller/Admin/TagsController.php
jdreesen, line 109: Maybe we could even add an `addTranslation(string $language, string $translation)` method in the model?

Then this would only be:
```suggestion
            $tag->addTranslation($request->get('language'), $request->get('translation'));
```

models/Element/Tag.php
roland4432: This will lead to an "undefined index" error, when requesting non-existent translations whereas ?? does not

models/Element/Tag.php
roland4432: Totally aggree on this one, hence all models in the pimcore lifecyle have public properties, therefore I went with public as well

models/Element/Tag.php
jdreesen: Oh, you're totally right. `??` is correct here!

models/Element/Tag.php
fashxp: yes, but we are in the process of changing them to protected ... so please go for protected for new ones

roland4432: You have no clue, how much this decission makes me happy.. there is a ton of problems that will just be gone with wind with this approach

bundles/AdminBundle/Resources/public/js/pimcore/element/tag/configuration.js
```diff
@@ -37,18 +40,121 @@ pimcore.element.tag.configuration = Class.create({
 
             var tree = new pimcore.element.tag.tree();
 
+            treeLayout = tree.getLayout();
+            treeLayout.addListener({
+                itemclick: function(tree, record){
+                    this.fillPanelRow(record);
+                }.bind(this),
+            })
```
fashxp, line 48:
```suggestion
            });
```

bundles/InstallBundle/Resources/install.sql
```diff
@@ -501,6 +501,9 @@ CREATE TABLE `tags` (
   `parentId` int(10) unsigned DEFAULT NULL,
   `idPath` varchar(190) DEFAULT NULL,
   `name` varchar(255) DEFAULT NULL,
+  `translations` text DEFAULT NULL,
```
fashxp, line 504: migration is missing

roland4432: Omg, you're right.. I added them immediatly...

roland4432:
One of the sql checks fails, but I don't think it's caused by my PR

bundles/CoreBundle/Migrations/Version20200518205733
```diff
@@ -0,0 +1,33 @@
+<?php
```
jdreesen, line 1: This file should have a `.php` extension.

roland4432: It was autogenerated from console via pimcore:migrations:generate but you're right

brusch:
I'm wondering if it wouldn't be better to just use the admin translations ... this would be then consistent with other UI parts where we're doing the same. 

@fashxp what do you think?

fashxp:
In my point of view, admin translations as they are right now, are not suitable for it. 

If we want to make a more generic solution, then we would have to think much bigger and improve usage of admin translations for different contexts. 

Similar challenges we have with translation of select options, see also one aspect of https://github.com/pimcore/pimcore/issues/3409

roland4432:
Did you guys come to any conclusion yet?

roland4432:
can be closed as will be obsolete in favor of #6568

PR #6413, opened 2020-05-19
[Data objects] Fulltext search for select fields

BlackbitNeueMedien:
Currently you can filter a select field in a data object's edit panel only be prefix. For example when you have the options `Test` and `My test` and you type `test` in the input field, you will only get `Test` suggested but not `My test`. With this PR the searched string can be anywhere in the option to get suggested.


BlackbitNeueMedien:
This PR is not causing the failing test - Travis reports `The command "mysql --version" failed and exited with 127 during .`

PR #6416, opened 2020-05-19
[Documentation] Fix typos

BlackbitNeueMedien:
By the way: What does `Passing multiple relations (w/o multiple assignments check) in data objects is deprecated and will throw exception in Pimcore 7` actually mean or to which pull request / commit does it refer?


dvesh3:
@BlackbitNeueMedien means if you pass multiple relations with same id to `ManyToManyRelation` field on save through API and `Allow Multiple Assignments` option is disabled for that field then API strips out extra relations with same id to save only unique data. This leads to data loss unknowingly so it will throw exception in Pimcore 7. please see changes here: https://github.com/pimcore/pimcore/pull/5852

BlackbitNeueMedien:
Thanks @dvesh3 for the explanation.

PR #6420, opened 2020-05-19
Get Versions by UserId

pdchaudhary:
there is no way to get version by userid without going to use direct query. we can add this function for filter versions by user-specific.

  
  

Resolves #

### DIFF
--- a/models/DataObject/Concrete/Dao.php
+++ b/models/DataObject/Concrete/Dao.php
@@ -431,6 +431,26 @@ class Dao extends Model\DataObject\AbstractObject\Dao
 
         return $versions;
     }
+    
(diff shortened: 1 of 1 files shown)


NiklasBr:
Can't this be accomplished with something like below?
```php
$versions = new Pimcore\Model\Version\Listing();
$versions->setCondition('userId = ?', [$userId]);
$versions->load();
```

pdchaudhary:
Yes but for specific object we can use directly that way

brusch:
I think the proposed way by @NiklasBr is the way to go. Let's keep it simple, this functionality seems to be an edge case to me.

PR #6421, opened 2020-05-19
Fix reference to where the constants are generated now

jdreesen:
`pimcore/config/constants.php` doesn't exist anymore, they're generated in `lib/Bootstrap.php` nowadays.


brusch:
Thanks man! 😊

PR #6422, opened 2020-05-19
Update Listing.php

zman-dwa:
Removed typecasting to INT for the `$classId` in the joinProduct() method in the <a href="https://pimcore.com/docs/api/master/Pimcore/Bundle/EcommerceFrameworkBundle/OrderManager/IOrderList.html">Order List Interface</a>.

## Additional info  

The method classId() in AbstractProduct or any Concrete object should return an INT according to the Pimcore API, ie., for <a href="https://pimcore.com/docs/api/master/Pimcore/Bundle/EcommerceFrameworkBundle/Model/AbstractProduct.html">AbstractProduct</a>.

However, a STRING can be entered as an identifier when creating a class. In this case, a STRING is returned with classId(). And joinProduct() fails.

## To Reproduce
1. Create a class with an identifier that is not an INT, for example "CAR"
2. Call joinProduct():
`$orderManager = Factory::getInstance()->getOrderManager();`
`$orderList = $orderManager->createOrderList();`
`$orderList->joinProduct( ClassFromStep1::classId() );`

This will fail when displaying if the classId() is type casted to an INT.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/OrderManager/Order/Listing.php
+++ b/bundles/EcommerceFrameworkBundle/OrderManager/Order/Listing.php
@@ -226,7 +226,7 @@ class Listing extends AbstractOrderList implements OrderListInterface
     }
 
     /**
-     * @param int $classId
+     * @param string $classId
      *
      * @return $this
      */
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
> 👍
> 
> Also change `@param int $classId` to `@param int|string $classId`

it is never an int, it always is a string

NiklasBr:
True

brusch:
Changes that 😊

brusch:
Fixed remaining issue in 633a758b0643baa2571a426a50dc8a41c8092e9a

PR #6423, opened 2020-05-19
Fix column label loss for the 'Operator Merge' type in object grids

mloock:
This fix solves the issue when you try to add or edit an 'Operator Merge' column with a custom label (other than 'Operator Merge').

The column label was uncorrectly reported in the grid display and in the related CSV export. It displayed 'Operator Merge' whatever you edited for this label in the 'Grid Options' window.


dvesh3:
thnx!

PR #6426, opened 2020-05-20
Fix: Data::isEmpty() methods have wrong PHP docs

blankse:
Fix wrong PHP docs of isEmpty methods and optimize some code inside.


brusch:
Thanks a lot!

PR #6430, opened 2020-05-21
use ElementDescriptor for Session saving

dpfaffenbauer:
That solves issues with DataObjects Linked to Doctrine Entities, also it optimizes the save-size of relations, not the full element is serialized, only its type and id.

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -1299,7 +1299,23 @@ class Service extends Model\AbstractModel
             if ($data) {
                 $element = Serialize::unserialize($data);
 
-                return $element;
+                $copier = new DeepCopy();
+                $copier->addTypeFilter(
+                    new \DeepCopy\TypeFilter\ReplaceFilter(
+                        function ($currentValue) {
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
BTW: I think we should find some sort of Service for all DeepCopy things that happen in Pimcore, they get more and more and the always cause us issues with Doctrine. Maybe it helps bundling them into services and make them extendable that way too, so we can fix stuff on our side and not on Pimcore side, wdyt?

fashxp:
still, what can be fixed in a generic way should be fixed on Pimcore side - so that also other can benefit from it.

dpfaffenbauer:
Yep, but hotfixing is not possible with Pimcore when certain part is not extendable. With a service, hotfixing is possible ;)

fashxp:
I'm completely with you - as long as the fixes find their way back ;-)

dpfaffenbauer:
sure

brusch:
Thanks!

PR #6431, opened 2020-05-21
Fix uncorrect CSV/XLS export for 'Operator Merge' values

mloock:
This fix is intented to correct CSV/XLS data export for merged values via the 'Operator Merge' formatter for object grids.

Before:

| Operator Merge   |
|----------|
| Array |
| Array |
| Array |

Now:

| Operator Merge   |
|----------|
| value1,value2 |
| value1,value3 |
| value4 |

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -2075,7 +2075,7 @@ class DataObjectHelperController extends AdminController
                     if ($helperDefinitions[$field]) {
                         $cellValue = DataObject\Service::calculateCellValue($object, $helperDefinitions, $field, ['language' => $requestedLanguage]);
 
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
How does your grid configuration look like? It feels like you want to use a Concatenator instead?
The merge operator will always return an array (just as the php array_merge function). 
The reason why this works in the grid is because the ExtJS template engine will (by default) do something like this when it renders the cell.

Suppose:
let v = ["x",y"];
let output=v+'';

The result will be "x,y"

I am not sure if we really want to mimic this behavior for the CSV or better use the right operator.

What do you think?

mloock:
You're right, I expected to concatenate the values from two fields in one column, I may have use another operator.
Anyway, displaying "Array" in the CSV exported file is not a proper render in my opinion.
Let me know if you want another rendering.

weisswurstkanone:
The user expects the same behavior so take this approach for now. Maybe we can make this configurable later if necessary.

PR #6437, opened 2020-05-22
[Ecommerce][Elastic Search] Replace reindexing-mode with ES Native Reindexing

fashxp:
this is an alternative implementation to #5744 for the reindexing part. 

Removes ecommerce framework reindexing mode completely and forces user to start ES native reindex when mapping update is not possible. 

During ES native reindex
- any other mapping updates are not possible
- data updates are only updated in store table and updated via process update queue to ES index after reindex is finished 

Thus it would reduce complexity. 

Open question: 
- does ES serve requests from old index while building the new one? 
- what to do with delete operations when reindex is running?

### DIFF
--- a/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
@@ -881,7 +881,7 @@ abstract class AbstractElasticSearch extends Worker\AbstractMockupCacheWorker im
                 throw new \Exception($errorMessage);
             } else {
                 //only write log message once a minute to not spam up log file when running update index
-                if($this->lastLockLogTimestamp > time() + 60) {
+                if($this->lastLockLogTimestamp < time() + 60) {
                     $this->lastLockLogTimestamp = time();
                     Logger::warning($errorMessage . ' (will suppress subsequent log messages of same type for next 60 seconds)');
                 }
--- a/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
@@ -615,7 +615,7 @@ abstract class AbstractElasticSearch extends Worker\AbstractMockupCacheWorker im
         }
     }
 
-    protected function doCreateOrUpdateIndexStructures($exceptionOnFailure = false)
+    protected function doCreateOrUpdateIndexStructures()
     {
         $this->checkIndexLock(true);
 
--- a/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
@@ -550,7 +550,7 @@ abstract class AbstractElasticSearch extends Worker\AbstractMockupCacheWorker im
             if (is_array($matches) && count($matches) > 1) {
                 $version = (int)$matches[1];
                 if ($version != $this->indexVersion) {
-                    $indexNameVersion = $this->getIndexNameVersion();
+                    $indexNameVersion = $this->getIndexNameVersion($version);
                     Logger::info('Index-Actions - Delete old Index ' . $indexNameVersion);
                     $this->deleteEsIndexIfExisting($indexNameVersion);
                 }
--- a/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
@@ -862,6 +862,11 @@ abstract class AbstractElasticSearch extends Worker\AbstractMockupCacheWorker im
     }
 
     /**
+     * @var int
+     */
+    protected $lastLockLogTimestamp = null;
+
+    /**
      * Verify if the index is currently locked.
      * @param bool $throwException if set to false then no exception will be thrown if index is locked
      * @return bool returns true if no exception is thrown and the index is locked
@@ -874,7 +879,11 @@ abstract class AbstractElasticSearch extends Worker\AbstractMockupCacheWorker im
             if ($throwException) {
                 throw new \Exception($errorMessage);
             } else {
-                Logger::warning($errorMessage);
+                //only write log message once a minute to not spam up log file when running update index
+                if($this->lastLockLogTimestamp > time() + 60) {
+                    $this->lastLockLogTimestamp = time();
+                    Logger::warning($errorMessage . ' (will suppress subsequent log messages of same type for next 60 seconds)');
+                }
             }
 
             return true;
@@ -889,5 +898,6 @@ abstract class AbstractElasticSearch extends Worker\AbstractMockupCacheWorker im
 
     protected function releaseIndexLock() {
         Lock::release(self::REINDEXING_LOCK_KEY);
+        $this->lastLockLogTimestamp = null;
     }
 }
--- a/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
@@ -15,6 +15,7 @@
 namespace Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\ElasticSearch;
 
 use Elasticsearch\Common\Exceptions\BadRequest400Exception;
+use Elasticsearch\Common\Exceptions\Missing404Exception;
 use Elasticsearch\Common\Exceptions\NoNodesAvailableException;
 use Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Config\ElasticSearch;
 use Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Config\ElasticSearchConfigInterface;
@@ -561,6 +562,11 @@ abstract class AbstractElasticSearch extends Worker\AbstractMockupCacheWorker im
         return $data;
     }
 
+    /**
+     * @param int $objectId
+     * @param IndexableInterface|null $object
+     * @throws \Exception
+     */
     protected function doDeleteFromIndex($objectId, IndexableInterface $object = null)
     {
         $esClient = $this->getElasticSearchClient();
@@ -568,6 +574,10 @@ abstract class AbstractElasticSearch extends Worker\AbstractMockupCacheWorker im
         $storeEntry = \Pimcore\Db::get()->fetchRow('SELECT * FROM ' . $this->getStoreTableName() . ' WHERE  o_id=? AND tenant=? ', [$objectId, $this->getTenantConfig()->getTenantName()]);
         if ($storeEntry) {
 
+            $isLocked = $this->checkIndexLock(false);
+            if($isLocked) {
+                throw new \Exception("Delete not possible due to product index lock. Please re-try later.");
+            }
 
             try {
                 $esClient->delete([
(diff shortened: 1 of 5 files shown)


andreas-gruenwald:
Comparison of advantages and disadvantages compared to #5744:

**Advantages**
 - removes a central part of the reindexing complexity
 - simplifies the index update process and makes it more transparent to the developer.
 - only have one locking mechanism (Pimcore Lock) in place, not two.

**Disadvantages**
 - Additional tool ``ecommerce:indexservice:elasticsearch-sync reindex`` becomes necessary for the ES stack. This is a deviation from the other indexing processes (``MySQL``).

**Conclusio**
In my opinion the advantages outhweight the disadvantage. However, now we are that far, I would even **go one step further** and try to remove the error-prone local index version file as well.

**Idea of how to get rid of the local index file**: 
1)  ``determineAndSetCurrentIndexVersion()``: index version is not determined based on the local file anymore, but based on the index name that is currently assigned to the ES alias. 
Example: ``AT_de_elastic -> AT_de_elastic_11`` -> index name is ``AT_de_elastic_11`` and index version is ``11``.
 2) Then we could also go one step further and also name the index that is build in parallel ``AT_de_elastic_tmp`` instead of ``AT_de_elastic_11``, which is another simplification.
3) On index switch, then ``AT_de_elastic_tmp`` is first renamed into ``AT_de_elastic_12`` and then the alias index is switched. 

What do you think?

---
Open question:
>- does ES serve requests from old index while building the new one? 

Yes of course! I also just tried it by myself. 

>what to do with delete operations when reindex is running?

Probably needs to be discussed in more detail. One solution could to be introduce a "delete" flag in the store table and delete the store table entry only after it has been ensured that the index entry has been deleted as well. While reindexing, deletion wouldn't be possible, resulting in an error log in the store table. Another approach could be to perform the delete operation while reindexing on both indices, hoping that the document will be eliminated by doing so.

fashxp:
getting rid of versions file might be a bit harder, as it also stores a settings version of the index settings ... see [here](https://github.com/pimcore/pimcore/blob/4e9a36b8e79c2ea7ae2d53866a6f95b1a31c3b85/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L644) 

not sure how we can tackle this.

andreas-gruenwald:
@fashxp The settings version is only required in ``doCreateOrUpdateIndexStructures()``? In the new version, the developer will probably prefer to execute the native reindexing instead of that method, as the native reindexing will always update the index settings and the mappings (while locking), and will also reindex the documents accordingly and in a timely manner. 

Maybe the native reindexing could always take place after / or instead of the initial index structure has been created with ````doCreateOrUpdateIndexStructures()``? The overhead would be probably acceptable.

Another solution could be to retrieve the settings directly from the ES index via the ES API, so that it is not required anymore to hash the settings inside the local versions file?

fashxp:
I like the idea of running reindex always on `doCreateOrUpdateIndexStructures()` - I would add a 'if necessary' though ... means reindex should be executed automatically instead of throwing an exception if `putMapping` fails. 
Then also your disadvantage is not valid anymore as developer doesn't need any additional command. 

Also executing putSettings on every `doCreateOrUpdateIndexStructures()` might work and so we could really get rid of the versioning file. I'll have a look at it and update the PR accordingly.

fashxp:
- removed versioning file
- added automatic execution of reindex if `doCreateOrUpdateIndexStructures` fails

Open: 
- delete of elements during reindex

fashxp:
> Isn't accessing the version file with the highest number a bit dangerous?
Wouldn't it be better to get the index version based on the alias assignment?
See GET /at_de_elastic/_alias.
good point, changed it

fashxp:
delete now throws an exception.

andreas-gruenwald:
@fashxp Based on your last updates I think that you are pretty done?
The only open issue is how to deal with the synonym file upload, which is currently only possible for the local server.

fashxp:
@andreas-gruenwald  yes, synonym file upload should be a separate PR.

andreas-gruenwald:
> @andreas-gruenwald yes, synonym file upload should be a separate PR.

Ok, but the plan is still to add it to ``EsSyncCommand``?
Probably we can take care of the PR as soon as this one is completed / merged?
Should I provide a new PR then, or are you going to take care of it?

fashxp:
> Ok, but the plan is still to add it to EsSyncCommand?
yes
> Probably we can take care of the PR as soon as this one is completed / merged?
yes
> Should I provide a new PR then, or are you going to take care of it?
as you like

fashxp:
but please test the implementation of this PR first if you got some time.

andreas-gruenwald:
@fashxp 

Final feedback based on another project-specific testround:

### Redundant Logging

**1. Avoid redundant logging in concurrent commands while index lock is active**
The log messages that appear in other commands (bootstrap, preparation) should appear only once in a while instead of popping up multiple times in the console output:

https://github.com/pimcore/pimcore/blob/e2269f3ba20376c8e7b7c2ddb64ca525ca14e593/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L869

Suggestion: only output log message every x seconds to avoid that the console is polluted. if the lock is released, the next time the message appears it has to be displayed, again.

Also, when the ``~/www/bin/console ecommerce:indexservice:process-queue update-index --tenant=AT_de_elastic`` is started while the lock is active it must be ensured that not thousand redundant log messages pop up in the console and inside the data logs.

### Index Cleanup
**2. Use existing method to determine old index name on index cleanup in switchIndexAlias():**

Replace 

https://github.com/pimcore/pimcore/blob/e2269f3ba20376c8e7b7c2ddb64ca525ca14e593/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L548-L549

with 
```php
$oldVersionName = $this->getIndexNameVersion($version);
Logger::info('Index-Actions - Delete old Index ' . $oldVersionName);
$this->deleteEsIndexIfExisting($oldVersionName);
``` 

---

**2. Move cleanup of old indices into separate method instead of keeping it in switchIndexAlias()**
We could move that block into a separate method ``cleanupOldEsIndices(int $currentIndexVersion)`` to make the code structure even more clean.
https://github.com/pimcore/pimcore/blob/e2269f3ba20376c8e7b7c2ddb64ca525ca14e593/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L541-L557

**3. Show error message if a index cannot be deleted on cleanup**
Sometimes it happens that indices cannot be cleaned up / deleted for any reason (e.g., ES index locked, closed, readonly, whatever). Obviously the API confirms the deletion but the index is still in place. For that case it could be helpful to add another

```php
$esClient = $this->getElasticSearchClient();
$result = $esClient->indices()->exists(['index' => $oldVersionName]);
if ($result) {
    Logger::error("Could not delete index {$oldVersionName} while cleanup. Please remove the index manually.");
}
```
after the cleanup attempt in ``switchIndexAlias()``.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -866,7 +879,11 @@ protected function checkIndexLock(bool $throwException = true) : bool {
             if ($throwException) {
                 throw new \Exception($errorMessage);
             } else {
-                Logger::warning($errorMessage);
+                //only write log message once a minute to not spam up log file when running update index
+                if($this->lastLockLogTimestamp > time() + 60) {
```
andreas-gruenwald, line 883: The first time ``$this->lastLockLogTimestamp`` is null, so probably Travis will complain.
Either cast it to an integer or initialize ``lastLockLogTimestamp`` with 0 instead of null.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -539,14 +539,19 @@ public function switchIndexAlias()
...
+
+    protected function cleanupUnusedEsIndices(): void {
+        $esClient = $this->getElasticSearchClient();
         $stats = $esClient->indices()->stats();
         foreach ($stats['indices'] as $key => $data) {
             preg_match('/'.$this->indexName.'-(\d+)/', $key, $matches);
             if (is_array($matches) && count($matches) > 1) {
                 $version = (int)$matches[1];
                 if ($version != $this->indexVersion) {
                     Logger::info('Index-Actions - Delete old Index ' . $this->indexName.'-'.$version);
```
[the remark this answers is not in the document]
fashxp: done

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
fashxp: done

andreas-gruenwald:
https://github.com/pimcore/pimcore/blob/012091e2271a2523d63e0eed2d3e45422e361a9a/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L884 has to be replaced with ``if($this->lastLockLogTimestamp < time() - 60) {``

podarcis:
I'm not sure if this is related, but when calling console command `ecommerce:indexservice:elasticsearch-sync reindex` this error occurs:

```
bin/console ecommerce:indexservice:elasticsearch-sync reindex -v
Process tenant "products" (mode "reindex")...

In DefaultElasticSearch5.php line 80:
                                   
  [Exception]                      
  Unknown Type for mapping params  
                                   

Exception trace:
  at /var/www/pimcore6/vendor/pimcore/pimcore/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/DefaultElasticSearch5.php:80
 Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\ElasticSearch\DefaultElasticSearch5->getMappingParams() at /var/www/pimcore6/vendor/pimcore/pimcore/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php:780
 Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\ElasticSearch\AbstractElasticSearch->createEsIndex() at /var/www/pimcore6/vendor/pimcore/pimcore/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php:883
 Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\ElasticSearch\AbstractElasticSearch->startReindexMode() at /var/www/pimcore6/vendor/pimcore/pimcore/bundles/EcommerceFrameworkBundle/Command/IndexService/EsSyncCommand.php:72
 Pimcore\Bundle\EcommerceFrameworkBundle\Command\IndexService\EsSyncCommand->execute() at /var/www/pimcore6/vendor/symfony/symfony/src/Symfony/Component/Console/Command/Command.php:255
 Symfony\Component\Console\Command\Command->run() at /var/www/pimcore6/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:1018
 Symfony\Component\Console\Application->doRunCommand() at /var/www/pimcore6/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php:97
 Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at /var/www/pimcore6/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:271
 Symfony\Component\Console\Application->doRun() at /var/www/pimcore6/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php:83
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /var/www/pimcore6/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:147
 Symfony\Component\Console\Application->run() at /var/www/pimcore6/bin/console:36

ecommerce:indexservice:elasticsearch-sync [--tenant [TENANT]] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [--ignore-maintenance-mode] [--maintenance-mode] [-e|--env ENV] [--no-debug] [--] <command> <mode>
```

Seems like `getMappingParams()` is called without the type from within this context.

Environment:
* Pimcore 6.6.5 (dirty, as I'm on PR 6487)
* Elasticsearch5 adapter using Elasticsearch 2.4

PR #6438, opened 2020-05-22
Fix on object variant grid: batch edit for 'select' fields + batch edit all with selected items

mloock:
This PR fixes two issues in the object variant grid:

1/ broken batch edit for 'select' fields: the code worked for the object type 'object', not 'variant', so the select options were not populated in the batch edit popup,
2/ broken 'batch edit all' when some lines were selected in the grid.


dvesh3:
thanks a lot!

PR #6441, opened 2020-05-23
Batch processes performance

kubaplas:
Hi, I would like to discuss and improve the performance of batch processes and exports. I think it would be best if we discuss this using examples, that's why it is a PR not an issue.

In this PR I found two places that influence performance:
- size of the chunks that are being exported in a single request - currently set to 20, I've raised it to 100, gaining great general performance boost with the cost of longer requests
- the delay between requests - I reduced it from 400 to 50ms. I guess its purpose is to not overload backend or frontend of the application but in most cases, it is the main factor of the processing time.

Bigger chunks in export.
Lower the delay between batch processes.

## Possible further improvements
- implement working in chunks in batch processing, similar to the export
- work on the list of ids in the backend, so the list of ids wouldn't have to be sent between frontend and the backend; this would have additional pros: possibility to resume job, the backend could decide dynamically how many jobs it should handle (e.g. 5 seconds instead of 20 jobs), jobs could be processed in the background

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -1795,7 +1795,7 @@ class DataObjectHelperController extends AdminController
 
         $ids = $list->loadIdList();
 
-        $jobs = array_chunk($ids, 100);
+        $jobs = array_chunk($ids, 20);
 
         $fileHandle = uniqid('export-');
         file_put_contents($this->getCsvFile($fileHandle), '');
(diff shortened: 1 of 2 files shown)


msphn:
This will work for simple objects but probably not for more complex ones, I fear the processes could queue up and produce a lot of load. For now those imports even work fine on live environments.

What could be handy, allowing exports of grid configurations via cli.

kubaplas:
@msphn to export grids in cli you can try https://github.com/DivanteLtd/pimcore-scheduled-export
About queuing - you mean the delay in js? I think timeout is run after job is finished, so there would be no queue build up.

msphn:
I don't know if the JS is using simultaneous requests to process data or is waiting for the previous request to finish or not. Just if you would allow every 50ms to run a new one it could be that the previous one haven't finished yet. That's all what came into my mind :).

brusch:
Chunk size `100` is definitely too big for more complex data objects, I've just reverted that. Reducing the delay is totally fine I guess.

cancan101:
Why is there a delay at all when sending?

PR #6442, opened 2020-05-23
Load ids in batches.

### LINKED ISSUE #6408: CustomReport | CSV Download | Add pagination
# Bug Report

## Problem Description
Currently on CSV-Download, the entire data is fetched at once from the database. 
Offset-/Limit is not taken into consideration internally:

This causes internal server errors / SQL timeouts, if many datasets are part of the result (e.g. 600.000 records). 

## Solution Suggestion
Fetch the dataset using blocks shouldn't be that hard.
Line:
https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Controller/Reports/CustomReportController.php#L409

One just has to verify if the SQL statement defined in the custom report already defines offset / limit on its own. If yes, then the download can remain the same. If no, then pagination can be applied.

We can provide a PR if you like the suggestion.

kubaplas:
For large sets of data loadIdList (e.g. used in count or export) can cause MySQL error.
Specifically, list of ids can be larger than default max_allowed_packet of MySQL server.
Changed how loadIdList work - for large sets of id loads them in batches of 10000.
Partially resolves #6408

### DIFF
--- a/models/DataObject/Listing/Dao.php
+++ b/models/DataObject/Listing/Dao.php
@@ -27,6 +27,8 @@ use Pimcore\Model\DataObject;
  */
 class Dao extends Model\Listing\Dao\AbstractDao
 {
+    CONST MAX_ROWS_IN_QUERY = 10000;
+
     /**
      * @var \Closure
      */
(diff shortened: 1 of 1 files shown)


kubaplas:
Closing as it break current counting method.

PR #6453, opened 2020-05-25
Fix application logs archive table useless creation when no entries to insert

mloock:
This PR is intented to fix the useless creation of the monthly application logs archive table when there are no related entries in the original application_logs table


mturalenka:
thanks a lot. it wasn't just creating table, it was permanently moving all records from application log table and moving them to archive table on each run. as result application table was empty all time.

mturalenka:
@brusch Hi,
Why this fix wasn't released? i don't see it in commit history for the 6.6.9

brusch:
@mturalenka that went to `master` therefore it will be included in `6.7`.

PR #6461, opened 2020-05-27
[Tracking] fix quantity for trackCartProductActionAdd

jremmurd:
# Bugfix

Currently the `GoogleTagManager` will always track quantity 1, this PR considers the passed quantity in the method signature.


fashxp:
ui, thx!

PR #6471, opened 2020-05-28
fix flatpickr compatibility

cannonerd:
Change date format to ISO to match flatpickr date format.

The Orderlist filtering tool shows a different date than what the filter uses.

Go to orderlist, check the date in UI, go to browser  inspection tools and see that the value given is correct, but in wrong form

```html
<input 
type="text" 
id="date-from" 
class="form-control date flatpickr-input active" 
name="from"
placeholder="Date from" 
value="01.05.2020"
> 
```

it should be: 

```html
<input 
type="text" 
id="date-from" 
class="form-control date flatpickr-input active" 
name="from"
placeholder="Date from" 
value="2020-05-01"
>
```


dvesh3:
thnx!

PR #6477, opened 2020-05-28
[Data Objects] fields in bricks are not set on object creation

### LINKED ISSUE #6454: [Objects] Relations in bricks are not set on object creation
# Bug Report

Verified in Pimcore v6.6.2:

### Use case: 

I want to create an object (or a variant) add it a brick and set a relation field inside that brick. The brick gets added but the relation is not set in the brick. If I do a `$variant = Product::getById(xxx)` all works fine, but if the object was just created the relations is not set.

### Dummy code (to reproduce)
```php
// all listeners disabled
// inheritence enabled

$product = Product::getById(123);

$variant = new Product();
$variant->setType(AbstractObject::OBJECT_TYPE_VARIANT); // also does not work if its a regular object
$variant->setParent($product);
$variant->setPublished(true);
$variant->setKey(time());

if ($brickContainer = $variant ->getMyBrickField()) {
    if (!($brick = $brickContainer->getMyBrick())) {
        $brick = new MyBrick($variant );
        $brickContainer->setMyBrick($brick);
    }
}

$someObject = SomeClass::getById(123456789);
if ($someObject) {
    $brick->setSomeClass($someObject);                // the relation field
} else {
    throw new \Exception("Object not found");
}

$variant ->save();

// variant is created, but no relation 'someClass' is set
```

### LINKED ISSUE #6455: [Objects] Inheritence in bricks bug
# Bug Report
Verified in Pimcore v6.6.2

### Use Case:

I want set a field on an object's child brick field. The parent has a brick with a field A which has a value, the children sets a value to another field B of that brick. Now the children brick should inherit the value of field A and should have a value set for field B, but field A is not empty an thus the value is not inherited, but the value of the parent is set explicitly to the child object.

The behaviour is the same if  `$variant` was just created as well as if it already existed.

### Dummy code (to reproduce)
```php
// all listeners disabled
// inheritence enabled

$product = Product::getById(123);

$variant = new Product();
$variant->setType(AbstractObject::OBJECT_TYPE_VARIANT); // also does not work if its a regular object
$variant->setParent($product);
$variant->setPublished(true);
$variant->setKey(time());

if ($brickContainer = $variant ->getMyBrickField()) {
    if (!($brick = $brickContainer->getMyBrick())) {
        $brick = new MyBrick($variant );
        $brickContainer->setMyBrick($brick);
    }
}

$brick->setPrice(100);

$variant ->save();
```

image

image

dvesh3:
Resolves #6454

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20200604110336.php
+++ b/bundles/CoreBundle/Migrations/Version20200604110336.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+use Pimcore\Model\DataObject;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20200604110336 extends AbstractPimcoreMigration
+{
+    public function doesSqlMigrations(): bool
+    {
+        return false;
+    }
+
+    /**
+     * @param Schema $schema
+     *
+     * @throws \Exception
+     */
+    public function up(Schema $schema)
+    {
+        $list = new DataObject\Objectbrick\Definition\Listing();
+        $list = $list->load();
+        foreach ($list as $brickDefinition) {
+            $this->writeMessage(sprintf('Saving object brick: %s', $brickDefinition->getKey()));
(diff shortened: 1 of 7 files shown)


brusch:
todo: 
- tests
- migration script

dvesh3:
> todo:
> 
> * tests
> * migration script

Done

PR #6480, opened 2020-05-28
Ecommerce fix notice filter

Yivan:
Hello,

This PR fix a notice error because undefined index when using select to SelectFromMultiselect filter (seems it has been forgotten for this one as the MultiselectFromMultiselect filter check well if set/empty already).

Thanks.


fashxp:
thx

PR #6483, opened 2020-05-29
[xliff import] fix XML special characters for input fields

elements-team:
## Steps to reproduce
- Export a localized data object input field with content "A & B"
- Export it via XLIFF
- Translate and reimport it with the same content

=> The target input field will contain `"A &amp; B"`

This PR fixes this problem by implementing an improved version for parsing the inner XML of the target nodes.


markus-moser:
I was logged in with the wrong Github user but the PR should work as expected. ;) 
We tested it with various different cases.

PR #6484, opened 2020-05-29
[Objects] Optional PHP type declarations for Data Object Classes

weisswurstkanone:


bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2073,7 +2073,8 @@ protected function getCsvFieldData(Request $request, $field, $object, $requested
                 // check for objects bricks and localized fields
                 if (DataObject\Service::isHelperGridColumnConfig($field)) {
                     if ($helperDefinitions[$field]) {
-                        return DataObject\Service::calculateCellValue($object, $helperDefinitions, $field, ['language' => $requestedLanguage]);
+                        $val = DataObject\Service::calculateCellValue($object, $helperDefinitions, $field, ['language' => $requestedLanguage]);;
```
dpfaffenbauer, line 2076: not necessary to store it in a var and then return it.

models/DataObject/ClassDefinition/Data/ManyToOneRelation.php
```diff
@@ -734,6 +734,14 @@ public function isFilterable(): bool
         return true;
     }
 
+    public function getInputType() {
```
dpfaffenbauer, line 737: actually could be more detailed defined if only one type has been configured, right?

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -493,4 +493,18 @@ public function filterMultipleAssignments($data, $container, $params)
 
         return $data;
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function getInputType() {
```
dpfaffenbauer, line 500: does it return null when nothing is set? or does it return a empty array?

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
weisswurstkanone: not really related :-)

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
weisswurstkanone: Hard to decide in the base class. So be on the safe side.

models/DataObject/ClassDefinition/Data/ManyToOneRelation.php
weisswurstkanone: Sure. We could do that.

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
dpfaffenbauer: sure, you changed it with this PR

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -2073,7 +2073,7 @@ protected function getCsvFieldData(Request $request, $field, $object, $requested
                 // check for objects bricks and localized fields
                 if (DataObject\Service::isHelperGridColumnConfig($field)) {
                     if ($helperDefinitions[$field]) {
-                        return DataObject\Service::calculateCellValue($object, $helperDefinitions, $field, ['language' => $requestedLanguage]);
+                        return DataObject\Service::calculateCellValue($object, $helperDefinitions, $field, ['language' => $requestedLanguage]);;
```
jdreesen, line 2076:
```suggestion
                        return DataObject\Service::calculateCellValue($object, $helperDefinitions, $field, ['language' => $requestedLanguage]);
```

models/DataObject/ClassDefinition/Data.php
```diff
@@ -664,13 +678,22 @@ protected function getPreGetValueHookCode(string $key): string
     public function getGetterCode($class)
     {
         $key = $this->getName();
+
+
+        if ($class->getUseTypeHints() && $this->getReturnType()) {
+            $typehint = ': ' . $this->getReturnType();
```
NiklasBr, line 684: This will intruduce a crash:y scenario on Table data types, which have declared an `array` docblock return type, but in reality will return either a string (i.e. no data) or an array. Either we make a BC break on the Table datatype by making it only return an array, or we can change the return type to `array|string` and validate here if the datatype has a pipe char and not append it as a typehint. See #6532, thoughts?

weisswurstkanone: this is only the default return type. it is up to the datatype to generate a better/different return type hint.

jdreesen: But as long as are we're not on PHP 8 with union types, it's not possible to add the return type of `array|string` for the mentioned method, so we would have to skip it. Is this supported?

NiklasBr: @jdreesen Relation types already return pipe-concatenated types, for example: [1](https://github.com/pimcore/pimcore/blob/c485e8d4ea1de0d7bb97fd06331205cce54cede9/models/DataObject/ClassDefinition/Data/ManyToOneRelation.php#L675), [2](https://github.com/pimcore/pimcore/blob/163f95e61ac97b8d5997932827ba93cc85755b0b/models/DataObject/ClassDefinition/Data/ManyToManyObjectRelation.php#L754), [3](https://github.com/pimcore/pimcore/blob/1f9fca52eacf4fcc8b8c7899b65c0b1900db9124/models/DataObject/ClassDefinition/Data/ManyToManyRelation.php#L885). These must also be solved

models/DataObject/ClassDefinition/Data.php
```diff
@@ -654,6 +654,20 @@ protected function getPreGetValueHookCode(string $key): string
         return $code;
     }
 
+    /**
+     * @return string
+     */
+    public function getInputType() {
```
brusch, line 660: Should we really go for a default here on the abstract? 
Imho it would be better to introduce an `TypeHintSupportInterface` that needs to be implemented by the data-type to support PHP type-hints.

weisswurstkanone:
In addition:
getPhpInputDocType
getPhpReturnDocType

models/DataObject/ClassDefinition/Data.php
```diff
@@ -654,6 +654,34 @@ protected function getPreGetValueHookCode(string $key): string
...
+     * @return string|null
+     */
+    public function getTypeHintReturnType(): ?string {
+        return '?' . $this->getPhpdocReturnType();
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getPhpDocInputType(): ?string {
```
jdreesen, line 674: It's used as this (with lowercase 'd') everywhere else (including the interface).
```suggestion
    public function getPhpdocInputType(): ?string {
```

models/DataObject/ClassDefinition/Data.php
```diff
@@ -654,6 +654,34 @@ protected function getPreGetValueHookCode(string $key): string
...
+        return $this->getPhpdocType();
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getPhpDocReturnType(): ?string {
```
jdreesen, line 681: It's used as this (with lowercase 'd') everywhere else (including the interface).
```suggestion
    public function getPhpdocReturnType(): ?string {
```

models/DataObject/ClassDefinition/Data.php
weisswurstkanone: thx!

jdreesen:
"Type hint" is a legacy naming from PHP 5 - since PHP 7 it's called "type declaration" (because they're not "hints" but "real" types) or respectively "parameter/property/return type", so maybe we should use this naming too?

models/DataObject/ClassDefinition/Data.php
```diff
@@ -654,6 +654,34 @@ protected function getPreGetValueHookCode(string $key): string
         return $code;
     }
 
+    /**
+     * @return string|null
+     */
+    public function getTypeHintInputType(): ?string {
```
jdreesen, line 660: I think "parameter type" would be a more appropriate name than "input type".

weisswurstkanone: changed it

brusch:
@jdreesen @weisswurstkanone fully agree, type declaration is the right wording! Thanks for the hint!

models/DataObject/ClassDefinition/Data.php
```diff
@@ -654,6 +654,34 @@ protected function getPreGetValueHookCode(string $key): string
         return $code;
     }
 
+    /**
+     * @return string|null
```
dvesh3, line 658: Is it possible to return null?

weisswurstkanone: yes

models/DataObject/ClassDefinition/Data/Objectbricks.php
```diff
@@ -706,7 +713,7 @@ public function getGetterCode($class)
         $code .= '/**' . "\n";
         $code .= '* @return ' . $classname . "\n";
         $code .= '*/' . "\n";
-        $code .= 'public function get' . ucfirst($key) . " () {\n";
+        $code .= 'public function get' . ucfirst($key) . " ()" . $typehint .  " {\n";
```
jdreesen, line 716:
```suggestion
        $code .= 'public function get' . ucfirst($key) . " ()" . $typeDeclaration .  " {\n";
```

brusch:
@jdreesen thanks a lot!

PR #6487, opened 2020-05-29
Ecommerce | Refactoring queue commands

andreas-gruenwald:
This PR strives to optimize the Ecommerce queueing commands and the entire queuing process by making the following adjustments:

1. Make the ``BootstrapCommand`` and the ``ProcessQueueCommand``s parallelizable by using the Webmozart's Parallelization trait. Instead of ``ProcessQueueCommand`` now ``ProcessPreparationQueueCommand`` and ``ProcessUpdateIndexQueueCommand`` exist.
2. The commands take care of parallelization and locking by themselves, so as a fact we can get rid of the worker information in the store tables.
3. Introduce ``IndexUpdateService`` which encapsulates logic for the retrieval of open queue entries based on the ecommerce store tables. The service is meant to provide useful methods  for store-table related updates. that are commonly needed, that's why also common methods to reset preparation queue and update-index queue entries have been added.
4. The preparation - queue is processsed product per product, which means that for 200.000 products and 10 tenants 200.000 (very expensive) product data object reads instead of 200.000*10=2.000.000 will take place (=reduction by 90%!).
5. Also, a TimeoutTrait has been introduced which can be used to unify the common timeout behavior. It is built on top of https://github.com/pimcore/pimcore/pull/6452, but may be integrated into the ``ParallelizationTrait`` extension.

**TODOs:**
1. Continue refactoring.
2. ``getStoreTableName()`` and ``doUpdateIndex`` change their signature. 
3. Especially, for ``doUpdateIndex()`` and the findologic solution we have to verify if the process update index queue solution will work as well.
4. Get rid of the old ProcessQueueCommand.
5. Get rid of all worker information and all the methods in the workers that become obsolete because there are no index updates necessary anymore (because the worker-id does not need to be saved anymore). This will reduce the complexity and the queuing a lot!
6. Get rid of the static ``IndexUpdater`` class.
7. Edit migration scripts and edit the upgrade notes.
8. Take care of the documentation.
9. Test

**Questions**
- Is making doUpdate public adequate?
- Should the common methods in the commands be moved to AbstractIndexServiceCommand? -> if yes, shift them.
- Are the commands already services or to they need to be registered (because ``@required`` is used to inject indexupdateservice)? -> please test in demo and add them if required.


bundles/EcommerceFrameworkBundle/Command/IndexService/ProcessPreparationQueueCommand.php
```diff
@@ -0,0 +1,167 @@
...
+    use Parallelization
+    {
+        Parallelization::runBeforeFirstCommand as parentRunBeforeFirstCommand;
+        Parallelization::runAfterBatch as parentRunAfterBatch;
+    }
+
+    /**
+     * @var IndexUpdateService
+     */
+    protected $indexUpdateService;
```
andreas-gruenwald, line 42: just out of curiosity: why is this one protected. The command shouldn't be overriden.

bundles/EcommerceFrameworkBundle/Command/IndexService/ProcessUpdateIndexQueueCommand.php
```diff
@@ -0,0 +1,182 @@
...
+    use Parallelization
+    {
+        Parallelization::runBeforeFirstCommand as parentRunBeforeFirstCommand;
+        Parallelization::runAfterBatch as parentRunAfterBatch;
+    }
+
+    /**
+     * @var IndexUpdateService
+     */
+    protected $indexUpdateService;
```
andreas-gruenwald, line 38: Same here. protected or private?

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -434,8 +433,6 @@ protected function doUpdateIndex($objectId, $data = null, $metadata = null)
             $this->bulkIndexData[] = $bulkIndexData;
             $this->indexStoreMetaData[$objectId] = $routingId;
 
-            //update crc sums in store table to mark element as indexed
-            $this->db->query('UPDATE ' . $this->getStoreTableName() . ' SET crc_index = crc_current WHERE o_id = ? and tenant = ?', [$objectId, $this->name]);
```
andreas-gruenwald, line 438: That line was important to be removed.

bundles/EcommerceFrameworkBundle/Command/IndexService/ProcessPreparationQueueCommand.php
fashxp: maybe not overwritten, but extended?

bundles/EcommerceFrameworkBundle/Command/IndexService/ProcessUpdateIndexQueueCommand.php
fashxp: maybe not overwritten, but extended?

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
fashxp: but it is removed, right?

andreas-gruenwald: yes

bundles/EcommerceFrameworkBundle/Resources/config/index_service_configs_workers.yml
```diff
@@ -67,15 +67,25 @@ services:
 
     Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\OptimizedMysql:
         parent: Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\AbstractWorker
+        arguments:
+            $workerMode: '%pimcore_ecommerce.index_service.worker_mode%'
 
     Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\ElasticSearch\DefaultElasticSearch5:
         parent: Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\AbstractWorker
+        arguments:
+            $workerMode: '%pimcore_ecommerce.index_service.worker_mode%'
```
podarcis, line 76: I'm not sure if this is relevant for BC. However, if you've got a service extending one of the existing workers (e.g. DefaultElasticSearch5) you'll now have to provide the argument $workerMode in the service definition, as `ProductCentricBatchProcessingWorker`'s constructor has new non-nullable `string $workerMode` as 4th argument and auto-wiring doesn't work anymore:

```
Fatal error: Uncaught Symfony\Component\DependencyInjection\Exception\RuntimeException: Cannot autowire service "pimcore_ecommerce.index_service.products.worker": argument "$workerMode" of method "Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\ElasticSearch\AbstractElasticSearch::__construct()" is type-hinted "string", you should configure its value explicitly. in /var/www/pimcore6/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Compiler/DefinitionErrorExceptionPass.php on line 54
```

Example definition to get it working again:

```yaml
services:
    # Index service worker
    AppBundle\EcommerceFramework\IndexService\Worker\Products:
        parent: Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Worker\AbstractWorker
        arguments:
            $workerMode: '%pimcore_ecommerce.index_service.worker_mode%'
```

podarcis:
@andreas-gruenwald @fashxp 
Thank you so much for digging into this, I really appreciate it!

What do you think: Is the current state ready for an early (pre-)alpha test on our side or should we wait a few more days?

andreas-gruenwald:
@fashxp Webmozarts created a new release, so you should be able to remove those parts of the parallelization trait that deal with the quotation of input options:
https://github.com/webmozarts/console-parallelization/releases/tag/1.0.1

lib/Console/Traits/Parallelization.php
```diff
@@ -103,4 +108,191 @@ protected function getContainer()
...
+                    $this->runTolerantSingleCommand((string) $item, $input, $output);
+
+                    $progressBar->advance();
+                }
+
+                $this->runAfterBatch($input, $output, $items);
+            }
+        } else {
+            // Distribute if we have multiple segments
+            $consolePath = realpath(getcwd().'/bin/console');
```
podarcis, line 249: @andreas-gruenwald 

Why is current working directory assumed? In my test it failed when the invocation of the master process (calling `bin/console`) happened outside of the project root, with `realpath()` returning `false`.

```
In Assert.php line 2220:
 
  [InvalidArgumentException]
  Expected a string. Got: boolean
 

Exception trace:
  at /var/www/share/REDUCTED/pimcore/htdocs/vendor/webmozart/assert/src/Assert.php:2220
 Webmozart\Assert\Assert::reportInvalidArgument() at /var/www/share/REDUCTED/pimcore/htdocs/vendor/webmozart/assert/src/Assert.php:233
 Webmozart\Assert\Assert::string() at /var/www/share/REDUCTED/pimcore/htdocs/vendor/webmozart/assert/src/Assert.php:1592
 Webmozart\Assert\Assert::fileExists() at /var/www/share/REDUCTED/pimcore/htdocs/vendor/pimcore/pimcore/lib/Console/Traits/Parallelization.php:252
 Pimcore\Bundle\EcommerceFrameworkBundle\Command\IndexService\ProcessUpdateIndexQueueCommand->executeMasterProcess() at /var/www/share/REDUCTED/pimcore/htdocs/vendor/webmozarts/console-parallelization/src/Parallelization.php:287
```

andreas-gruenwald: This is a known issue.
See https://github.com/webmozarts/console-parallelization/issues/32.

You are right, we should find a solution for that. I am going to take a look.

andreas-gruenwald: I just created a PR for that.

bundles/EcommerceFrameworkBundle/Resources/config/index_service_configs_workers.yml
fashxp: should be fixed, workerMode is optional now

lib/Console/Traits/Parallelization.php
andreas-gruenwald: @fashxp It's now possible to override the console path, for pimcore, if the problem is still existing: https://github.com/webmozarts/console-parallelization/pull/43

PR #6491, opened 2020-06-02
Possibility to override the admin locale in view templates

albertmueller:
In some projects I had issues with the admin locale, for example: When you open a document with the locale "hr" or "zh" in Pimcore backend, it shows just the basic key without translation because "hr" is not available as an admin translation and "zh" is only available as "zh_Hans".

This PR allows you to override the admin locale with following mapping like:
```
pimcore:
    translations:
        admin_translation_mapping:
            'hr': 'en'
            'zh': 'zh_Hans'
```


brusch:
Thanks a lot! 🙏

PR #6497, opened 2020-06-02
[Object bricks] Prevent fatal error if unserialize for data object is not possible

BlackbitNeueMedien:
Steps to reproduce:
1. Create class named `Abc`, add object brick field `bricks`
1. Create object brick `Brick`, allow it to be assigned to `Abc::bricks`
1. Override class by adding 
```yaml
pimcore:
    models:
        class_overrides:
            'Pimcore\Model\DataObject\Abc': 'AppBundle\Model\DataObject\Abc'
```
to `app/config/config.yml`
1. Create PHP class `AppBundle\Model\DataObject\Abc` under `src/AppBundle/Model/DataObject`
1. Clear cache
2. Create object of class `Abc`, assign brick `Brick` and save object
3. Rename class `AppBundle\Model\DataObject\Abc` to `AppBundle\Model\DataObject\AbcNew`
4. Update `app/config/config.yml` to:
 ```yaml
pimcore:
    models:
        class_overrides:
            'Pimcore\Model\DataObject\Abc': 'AppBundle\Model\DataObject\AbcNew'
```
5. Open object `Abc` and open versions tab.
6. Click on latest version (this should be the one which got created while the class was still named `Abc`)

Expected behaviour:
Version cannot be opened

Actual behaviour:
Fatal error `Fatal error: Pimcore\Model\DataObject\Objectbrick\Data\AbstractData::__wakeup(): The script tried to execute a method or access a property of an incomplete object. Please ensure that the class definition "AppBundle\Model\DataObject\Abc" of the object you are trying to operate on was loaded _before_ unserialize() gets called or provide an autoloader to load the class definition in /var/www/clients/client1/web1/web/vendor/pimcore/pimcore/models/DataObject/Objectbrick/Data/AbstractData.php on line 304`

### DIFF
--- a/models/DataObject/Objectbrick.php
+++ b/models/DataObject/Objectbrick.php
@@ -313,8 +313,12 @@ class Objectbrick extends Model\AbstractModel implements DirtyIndicatorInterface
         $brickGetter = null;
 
         // for backwards compatibility
-        if (isset($this->object) && $this->object && !$this->object instanceof  \__PHP_Incomplete_Class) {
-            $this->objectId = $this->object->getId();
+        if (isset($this->object) && $this->object) {
+            if($this->object instanceof  \__PHP_Incomplete_Class) {
+                Logger::error('Parent object of '.$brickGetter.' could not be unserialized (probably its class does not exist anymore)');
+            } else {
+                $this->objectId = $this->object->getId();
+            }
         }
 
         // sanity check, remove data requiring non-existing (deleted) brick definitions
--- a/models/DataObject/Objectbrick.php
+++ b/models/DataObject/Objectbrick.php
@@ -313,7 +313,7 @@ class Objectbrick extends Model\AbstractModel implements DirtyIndicatorInterface
         $brickGetter = null;
 
         // for backwards compatibility
-        if (isset($this->object) && $this->object) {
+        if (isset($this->object) && $this->object && !$this->object instanceof  \__PHP_Incomplete_Class) {
             $this->objectId = $this->object->getId();
         }
 
--- a/models/DataObject/Objectbrick/Data/AbstractData.php
+++ b/models/DataObject/Objectbrick/Data/AbstractData.php
@@ -306,7 +306,7 @@ abstract class AbstractData extends Model\AbstractModel implements Model\DataObj
 
     public function __wakeup()
     {
-        if ($this->object) {
+        if ($this->object && !$this->object instanceof \__PHP_Incomplete_Class) {
             $this->objectId = $this->object->getId();
         }
     }


weisswurstkanone:
I think we shouldn't silently drop such as errors. Can we show an errror message instead which is more meaningful to the user instead? Besides that I think there are more scenarios like third-party datatype got removed etc. This should be handled in the similar way then.

BlackbitNeueMedien:
It may seem like a small thing when I look at one version and this is not possible to be restored - but the problem is the fatal error. This is uncatchable so when you have an import script which fetches the last version of objects in a loop and one of them is not unserializable the whole script crashes.

BlackbitNeueMedien:
Added error log. Now it is implemented as for non-existing bricks, see
https://github.com/pimcore/pimcore/blob/dc2608fb318d33d9446f66f5c57adbdebe1c6342/models/DataObject/Objectbrick.php#L322-L339

weisswurstkanone:
I am not able to reproduce the fatal error. 
Instead, the problem is handled here 
https://github.com/pimcore/pimcore/blob/9c9c11fe6774e076eb0fd35e7bda2ee2b94c4a24/models/Version.php#L418
and behaves like the version does not exist.

What could be the difference?

I also don't see why handling this in __wakeup should solve what you describe because this piece of code is just for backwards compatibility.
The $object should not even get serialized.

Could you please double-check ?

BlackbitNeueMedien:
@weisswurstkanone This line is for serialization. At that time the class `AppBundle\Model\DataObject\Abc` existed. When you afterwards rename the class to `AppBundle\Model\DataObject\AbcNew` (only the PHP class and the YAML config, not in the serialized version) then the version points to a non-existing class. Is this handled anywhere, too?

weisswurstkanone:
That is what I did. I am not aware of that this is handled somewhere else. As said what I see the version tab shows the "Version does not exist" message because of the line I mentioned in Version.php

BlackbitNeueMedien:
The problem is in line https://github.com/pimcore/pimcore/blob/9c9c11fe6774e076eb0fd35e7bda2ee2b94c4a24/models/Version.php#L416

This `unserialize()` call triggers the fatal error as in this moment it is tried to also unserialize the object brick. To do that the __wakeUp() in the ObjectBrick class tries to assign the object id.

weisswurstkanone:
I attached my version file. Do you see any obvious difference?
[version.zip](https://github.com/pimcore/pimcore/files/4730113/version.zip)

weisswurstkanone:
Can you attach your version file as well ?

BlackbitNeueMedien:
Also thought about that, actually not relevant. On the other hand: The work is done, so I do not see any disadvantage of ensuring this BC. But I would neither be angry if you close this PR.

fashxp:
lets close this for now. we can reopen if a realistic use case occurs.

PR #6498, opened 2020-06-02
[Ecommerce][Indexservice] Fix timeout for transactional queries

jremmurd:
Calling `sleep` will cast the given value to `int` so a default value auf `0.5` would be treated as `sleep(0)`. This PR utilizes `usleep` to be able to use float values.


fashxp:
thx

PR #6499, opened 2020-06-02
CustomReport | CSV Download with progress bar #6408

### LINKED ISSUE #6408: CustomReport | CSV Download | Add pagination

PGBrandi:
Resolves #6408
The creation of the file gets split in several chunks to prevent a timeout. The progress is returned after each chunk and the client keeps resuming until the server says that it's finished with the creation of the file.
A progress bar to show the current progress was also added.

### DIFF
--- a/bundles/AdminBundle/Controller/Reports/CustomReportController.php
+++ b/bundles/AdminBundle/Controller/Reports/CustomReportController.php
@@ -15,6 +15,7 @@
 namespace Pimcore\Bundle\AdminBundle\Controller\Reports;
 
 use Pimcore\Model\Tool\CustomReport;
+use Symfony\Component\Filesystem\Exception\FileNotFoundException;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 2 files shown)


brusch:
@PGBrandi Please fix the tests / PhpStan, see: 
https://travis-ci.com/github/pimcore/pimcore/jobs/342543011#L1189

Thanks a lot!

bundles/AdminBundle/Controller/Reports/CustomReportController.php
```diff
@@ -369,14 +370,13 @@ public function chartAction(Request $request)
     }
 
     /**
-     * @Route("/download-csv", name="pimcore_admin_reports_customreport_downloadcsv", methods={"GET"})
+     * @Route("/create-csv", name="pimcore_admin_reports_customreport_createcsv", methods={"GET"})
      *
      * @param Request $request
      *
      * @return BinaryFileResponse
```
brusch, line 377:
```suggestion
     * @return JsonResponse
```

weisswurstkanone:
What will be the outcome if the data changes in the meantime?

weisswurstkanone:
@PGBrandi ping

PGBrandi:
> What will be the outcome if the data changes in the meantime?

It depends on the changed data set. If one entry was removed/changed after already being processed, the data simply stays in the file the way it was. However, the new data from unprocessed entries is updated in the file.

PR #6500, opened 2020-06-02
further usage of routing

dpfaffenbauer:
Some things we missed with the JS Routing PR

### DIFF
--- a/lib/Tool/Admin.php
+++ b/lib/Tool/Admin.php
@@ -87,10 +87,11 @@ class Admin
      * @static
      *
      * @param string $scriptContent
+     * @param bool $asUrl
      *
      * @return mixed
      */
-    public static function getMinimizedScriptPath($scriptContent)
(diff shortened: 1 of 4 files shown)


lib/Tool/Admin.php
```diff
@@ -103,7 +104,15 @@ public static function getMinimizedScriptPath($scriptContent)
             '_dc' => \Pimcore\Version::getRevision()
         ];
 
-        return '/admin/misc/script-proxy?' . array_toquerystring($params);
+        if ($asUrl) {
+            @trigger_error(
+                'Calling Pimcore\Tool::getMinimizedScriptPath with $asUrl true is deprecated and will be removed with Pimcore 7.1.0',
```
jdreesen, line 109: Why remove it with Pimcore `v7.1` and not with `v7.0`?

dpfaffenbauer: good catch, can be removed with 7.0 as well

jdreesen: #6513

PR #6506, opened 2020-06-03
Dispatch pimcore.system.cache.clear event when calling pimcore:cache:clear command

BlackbitNeueMedien:
Currently the `pimcore.system.cache.clear` event does only get dispatched when the cache gets cleared via GUI, see https://github.com/pimcore/pimcore/blob/fb4af6928847b0635b6f782b53055e7fbdff9d8b/bundles/AdminBundle/Controller/Admin/SettingsController.php#L750

With this PR the event gets also dispatched when the cache gets cleared via CLI command `bin/console pimcore:cache:clear`


brusch:
Thank you very much!

PR #6515, opened 2020-06-03
[Document] Areablocks: Add option to limit certain bricks

### LINKED ISSUE #6366: [Document] Areablock: Possibility to limit the usage of certain bricks
Currently it is only possible to limit the amount of total bricks globally independent of the type, which is not always sufficient. 

### Feature Request
It would be nice to limit the usage on brick level individually, so that for instance a certain brick can be only used twice within a block. 

### Config Example
```twig
{{ pimcore_areablock("myAreablock", {
		"allowed": ["iframe","teasers","wysiwyg"],
		"limits": {
			"iframe": 1,
			"teasers": 2
		},
		"limit": 5
	})
}}
```
Default should stay on unlimited.
So in this example you can use only 1 `iframe` and two `teaser` bricks in this block but as many `wysiwyg` bricks, as long as it doesn't exceed the global limit of 5 bricks. 
Of course the `limits` config should also work independently of the `limit` option.

dvesh3:
Resolves #6366

## Additional info  
```twig
{{ pimcore_areablock("myAreablock", {
		"allowed": ["iframe","teasers","wysiwyg"],
		"limits": {
			"iframe": 1,
			"teasers": 2
		},
		"limit": 5
	})
}}
```


bundles/AdminBundle/Resources/public/js/pimcore/document/tags/areablock.js
```diff
@@ -22,6 +22,7 @@ pimcore.document.tags.areablock = Class.create(pimcore.document.tag, {
         this.id = id;
         this.name = name;
         this.elements = [];
+        this.elementCounter = [];
```
brusch, line 25: Should we name this `brickTypeUsageCounter` instead? Or something else which is a bit more specific and self describing?

dvesh3: changed it to `brickTypeUsageCounter`. thanks!

dvesh3:
> I think we should also enforce this configuration on the backend side, what do you think?

Done!

models/Document/Tag/Areablock.php
```diff
@@ -54,6 +54,11 @@ class Areablock extends Model\Document\Tag implements BlockInterface
      */
     protected $blockStarted;
 
+    /**
+     * @var array
+     */
+    private $elementTypeUsageCounter = [];
```
brusch, line 60: Maybe here `$brickTypeUsageCounter` to be consistent with the JS code?

dvesh3: done!

blankse:
@dvesh3 I get in every Areablock an "Notice: Undefined index" exception now. I created a PR for this.

PR #6374, opened 2020-05-12
[DataObject] Fully implement isEqual for all datatypes

dvesh3:
## Changes in this pull request  
Resolves #6165

PR #6517, opened 2020-06-03
[Ecommerce] Out-of-the-Box support for Elastic Search Synonyms

### LINKED ISSUE #6490: [Ecommerce] Out-of-the-Box support for Elastic Search Synonyms
# Feature Suggestion 
This issue provides a out-of-the-box solution suggestion for the synchronization of elastic search (ES) synonyms, by the ecommerce framework.

## History
- PR #5744 suggested a native reindexing approach for the elastic search (ES) ecommerce product index. 
- It also contained a suggestion on how to support ES synonym updates out-of-the-box.
- PR #5744 has been closed by #6437, which is a more radical approach. 
- The synonym feature, which was part of #5744 is not included in #6437 anymore, as the approach was not convenient for multiserver-environments, where ES is not installed on the same server as the web application.

## Requirements
The suggested solution depends on the native reindexing approach, so basically the following workflow has to be executed: 
- Merge PR #5744 
- create another PR on top of #5744 and based on the discussed solution in this approach
- approve PR.

### Update Synonyms 
Introduce a new method  in the ``AbstractElasticSearch`` worker to just update the synonym filters in the settings, nothing else.
This approach works quite well and does not require reindexing. The only disadvantage is that the index needs to be closed and reopened, which will case the index to be inavaible for some milliseconds. 
```php
   /**
     * Perform a synonym update "live". Attention: the index will be closed and opened,
     * so it might not be available for a moment (typically some milliseconds).
     */
    public function updateSynonyms() {
        try {
            $this->activateIndexLock(); //lock all other processes

            $currentIndexName = $this->getIndexNameVersion();
            $settings = $this->tenantConfig->getIndexSettings();

            $filters = @$settings['analysis']['filter'];
            if ($filters) {
                $indexPart = [
                    'analysis' =>
                        [
                            'filter' => []
                        ]
                ];

                foreach ($filters as $name => $filter) {
                    if (in_array($filter['type'],['synonym'])) {
                        $indexPart['analysis']['filter'][$name] = $filter;
                    }
                }
            }

            $esClient = $this->getElasticSearchClient();

            $esClient->indices()->close(['index' => $currentIndexName]);
            $esClient->indices()->putSettings([
                'index' => $this->getIndexNameVersion(),
                'body' => [
                    'index' => $indexPart
                ]
            ]);
            $esClient->indices()->open(['index' => $currentIndexName]);

        } finally {
            $this->releaseIndexLock();
        }
    }
```

#### EsSyncCommand
- see https://github.com/andreas-gruenwald/pimcore/blob/44904e7cd6f2662584ecd6184d9f51eda6332ccd/bundles/EcommerceFrameworkBundle/Command/IndexService/EsSyncCommand.php. 
 - Essentially the ``refresh-synonyms`` mode from PR #5744 needs to be restored.
 - instead of copying the synonym files, the newly introduced method ``$elasticWorker->updateSynonyms()`` should be called.

#### TODOs
- refine filtering of relevant synonym filters.
- only update the index if the settings actually changed (compare settings from server with current settings).
- in future versions, even ``_reload_search_analyzers`` (https://www.elastic.co/guide/en/elasticsearch/reference/7.x/indices-reload-analyzers.html) could be applied, but this feature is only available since ES 7.x and is basically a technical optimization.

### Support Synonym Synchronisation on Multi-Server Environments
The approach in PR #5744 was based on synonyms that are stored in a local file.

Example:
```yaml
pimcore_ecommerce_framework:
index_service:
        tenants:
           mytenantName:
               config_options:
                    index_settings:
                        analysis:
                          filter:
                                my_synonyms:
                                    type: synonym
                                    synonyms_path: "ele_elasticsearch/synonyms_%%tenant%%.txt" # must reside in local /etc/elasticsearch/
```

Now the idea is to add the synonyms directly to the index instead of putting them to a local file, because then the local synonym file does not need to be transmitted to a remote file system.

This is the proof of concept:

### ElasticSearch-Config
``Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\Config\ElasticSearch``

```php

    /**
     * @return array
     */
    public function getIndexSettings()
    {
        $indexSettings = $this->preparseIndexSettings($this->indexSettings);
        return $indexSettings;
    }

    /**
     * @param array $indexSettings
     * @throws \Exception
     */
    protected function preparseIndexSettings(array $indexSettings) {
        if (isset($indexSettings['analysis']['filter'])) {
            $filters = $indexSettings['analysis']['filter'];
            foreach ($filters as $filterName => $filter) {
                if ('synonym' === $filter['type']) {
                    if (array_key_exists('synonyms_path', $filter)) {
                        $path = $filter['synonyms_path'];
                        $filename = array_pop(explode_and_trim('/', $path));
                        $localHome = '/home/my-project/www/web/var/assets/System/Webshop/Search/';
                        $fullPath = $localHome.$filename;
                        if (!file_exists($fullPath)) {
                            throw new \Exception(sprintf('Synonym path "%s" does not exist.', $fullPath));
                        }
                        $content = file_get_contents($fullPath);
                        $synonymLines = explode_and_trim(PHP_EOL, $content);

                        $rewrittenFilter = $filter;
                        unset($rewrittenFilter['synonyms_path']);
                        $rewrittenFilter['synonyms'] = $synonymLines;

                        $indexSettings['analysis']['filter'][$filterName] = $rewrittenFilter;
                    }
                }
            }
        }

        //p_r($indexSettings['analysis']['filter']); exit;

        return $indexSettings;
    }
```
Basically the synonyms are parsed from the local synonym file and are then added to the index directly. The index configuration may look like this:

```yaml
pimcore_ecommerce_framework:
index_service:
        tenants:
           mytenantName:
               config_options:
                    index_settings:
                        analysis:
                          filter:
                                my_synonyms:
                                    type: synonym
                                    synonyms_path: "%kernel.root_dir%/var/web/var/assets/System/Webshop/Search/synonyms_%%tenant%%.txt"
```

#### TODOs
- while the solution is very simple, it still requires a local file, which is a small disadvantage. Instead of parsing the synonyms from a local file, a synonym-provider service could be introduced which can provide synonyms from a specific location. This needs some discussion.
- the right way to parse the synonyms and modify the index settings needs to be defined. Do we need some kind of ``getIndexSettingsModified()``, or ``getIndexSettingsOriginal()``?

andreas-gruenwald:
resolves #6490. 

**Suggestion for Synonym Providers:**
index_services.yml
```yaml
AT_de_elastic:
    config_id: '%ecommerce.index_service.elastic.config_id%'
    worker_id: '%ecommerce.index_service.elastic.worker_id%'
    config_options: 
        es_client_params:
            ...
        client_config:
            ...
        synonym_providers:
            app_synonym_filter:
               provider_id: Pimcore\Bundle\EcommerceFrameworkBundle\IndexService\SynonymProvider\FileSynonymProvider
               options:
                  synonymFile: '%kernel.project_dir%/web/var/assets/System/Webshop/Search/synonyms_%%tenant%%.txt'

        index_settings:
            ...
            analysis:
                analyzer:
                    ...
                normalizer:
                ...
                tokenizer:
                    ...
                filter:
                    app_synonym_filter: # mapping between synonym_provider and filter based on name
                        type: synonym # or synonym_graph or any other ES type
                        synonyms: []
                    ...
        ...
```

**TODO's**
- Final configuration "wording".
- Documentation.
- Testing on another system.
- Test without the new configuration option.


andreas-gruenwald:
Notes:
 - instead of using ``synonyms_provider`` in filters, link the synonym provider based on the filter name.
- Only update synonyms if there is a difference between the synonyms on the ES index and the local synonyms.
- Refactoring.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -871,6 +873,135 @@ public function startReindexMode()
...
+                return;
+            }
+
+            $esClient = $this->getElasticSearchClient();
+
+            if (!$skipComparison) {
+                $indexSettingsCurrentEs = $esClient->indices()->getSettings(['index' => $indexName])[$indexName]['settings']['index'];
+                $indexSettingsSynonymPartEs = $this->extractMinimalSynonymFiltersTreeFromIndexSettings($indexSettingsCurrentEs);
+
+                if (json_encode($indexSettingsSynonymPartEs) == json_encode($indexSettingsSynonymPartLocalConfig)) {
```
fashxp, line 913: compare arrays directly and not their json_strings ... should be faster.

fashxp:
also please fix travis errors, thx.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
fashxp: see also https://github.com/pimcore/pimcore/pull/6374#discussion_r427750038

andreas-gruenwald: done.

andreas-gruenwald:
> also please fix travis errors, thx.

Fixed parameter. Travis still running, but hope it helped.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -871,6 +873,135 @@ public function startReindexMode()
...
+            }
+
+            $indexName = $indexNameOverride ? : $this->getIndexNameVersion();
+
+            $indexSettingsSynonymPartLocalConfig = $this->extractMinimalSynonymFiltersTreeFromTenantConfig();
+            if (empty($indexSettingsSynonymPartLocalConfig)) {
+                Logger::info('No index update required, as no synonym providers are configured. '.
+                    'If filters have been removed, then reindexing will help to get rid of old configurations.'
+                );
+                return;
```
fashxp, line 904: lock is never released

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
```diff
@@ -871,6 +873,135 @@ public function startReindexMode()
...
+            $esClient = $this->getElasticSearchClient();
+
+            if (!$skipComparison) {
+                $indexSettingsCurrentEs = $esClient->indices()->getSettings(['index' => $indexName])[$indexName]['settings']['index'];
+                $indexSettingsSynonymPartEs = $this->extractMinimalSynonymFiltersTreeFromIndexSettings($indexSettingsCurrentEs);
+
+                if ($indexSettingsSynonymPartEs == $indexSettingsSynonymPartLocalConfig) {
+                    Logger::info(sprintf('The synonyms in ES index "%s" are identical with those of the local configuration. '.
+                        'No update required.', $indexName));
+                    return;
```
fashxp, line 916: lock is never released

fashxp:
please merge master branch once more and make sure, that it works with changes of https://github.com/pimcore/pimcore/pull/6594 

especially check if `updateSynonyms` is still called in `createEsIndex`,  and not in `putIndexMapping`

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
andreas-gruenwald: Are you sure? Lock is released here:
https://github.com/pimcore/pimcore/blob/3596cfa74f27a1d8ff53cdd6a92e3d47fd83db6d/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L940

See https://www.php.net/manual/en/language.exceptions.php
> If a return statement is encountered inside either the try or the catch blocks, the finally block will still be executed. Moreover, the return statement is evaluated when encountered, but the result will be returned after the finally block is executed.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
andreas-gruenwald: Lock is released in ``finally``.

bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
fashxp: hmm, somehow I managed to produce a leftover lock ... will test again after master is merged again.

PR #6524, opened 2020-06-04
[Quantity values] Only save default value with value AND unit to database

BlackbitNeueMedien:
Currently when you have set a default unit for a quantity value field but no default value and create a new object this quantity value without value gets saved to the database, even if you call `$object->setQuantityField(null)`. I wonder if this is intended.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/QuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/QuantityValue.php
@@ -648,7 +648,7 @@ class QuantityValue extends Data implements ResourcePersistenceAwareInterface, Q
      */
     protected function doGetDefaultValue($object, $context = [])
     {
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Yes and no. Problem here is that if you forget to define one part of the pair in the class definition then absolutely nothing will be written to the database. I think this is quite hard to notice. Can we catch this in the UI already maybe?

BlackbitNeueMedien:
~~Hmm, you are right as with this PR setting default unit without value does not have any effect at all, that is nonsense. We have to find a way to apply default value / unit in the GUI indepenently of the `handleDefaultValue()` function, so that those values get prefilled in the UI but not get applied when the object gets created programmatically.~~
But this would mean we wuld have the same problem as described in #4600. So I think the current implementation is correct.

PR #6529, opened 2020-06-04
Fix example code in the docs

jdreesen:
If I'm not mistaken, it works like this.


dvesh3:
It is about **extending** existing tags so i don't see a problem in example.

dvesh3:
well..it's just an example but yes might be helpful for beignners. please change it to `pimcore.document.tags.textarea`

dvesh3:
thanks!

Issue #6387, opened 2020-05-14
[Data Object] Optional PHP type declarations on generated getters and setters


brusch:
There should be a config option on the class definition to generate the PHP class including PHP (return) type hints. 

The result should be then: 

```php
/**
* @return string
*/
public function getMyName ():string {
	// ...
	return $data;
}

/**
* @param string $myName
* @return \Pimcore\Model\DataObject\MyClass
*/
public function setMyName (string $myName):\Pimcore\Model\DataObject\MyClass {
    // ...
	return $this;
}
```

msphn:
Maybe combine it with setting strict_types?

brusch:
@msphn yep!

dpfaffenbauer:
🎉 Sounds amazing :)

PR #6532, opened 2020-06-04
Corrects doctype for Table ClassDefinition inputs

NiklasBr:
I'm not 100% sure if this is the best solution, but the fact that an empty table will return a string should be accounted for. Personally I would prefer that it always returned an array, but since that might be a (minor) BC break I want to avoid doing that change for now.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Table.php
+++ b/models/DataObject/ClassDefinition/Data/Table.php
@@ -99,7 +99,7 @@ class Table extends Data implements ResourcePersistenceAwareInterface, QueryReso
      *
      * @var string
      */
-    public $phpdocType = 'array';
+    public $phpdocType = 'array|string';
 
     /**
      * @return int


jdreesen:
I'm afraid this won't work with #6387 / #6484.

NiklasBr:
I see, maybe change the default/empty value from `""` to `[]` then? Currently I need to do even thought the doctype only says `array`:

```php
$value = $product->getTable();
if (is_array($value)) {
    $value[] = $newData;
} else {
  $value = [$newData];
}
$product->setTable($value);
```

Would be nice to just have to ability to do this no matter if the value is empty or not:

```php
$value = $product->getTable();
$value[] = $newData;
$product->setTable($value);
```
Is that an acceptable BC break?

brusch:
In my opinion this would be an acceptable BC break for 6.7. 
I'd also classify this as a bug, because returning a string or null for the table data-type is just not right.  
@NiklasBr could you please update this PR accordingly (or create a new PR)? 
Btw. it seems that null is currently also a possible value, see https://github.com/pimcore/pimcore/blob/1e49e5cb106b1e701e979782fdf7cefaae96174d/models/DataObject/ClassDefinition/Data/Table.php#L436

brusch:
#6387 / #6484 will introduce an interface for PHP type hints, so we can also leave this out for the table data-type and fix it in v7. 
Then there's no break, the only drawback is that PHP type hints are not going to be supported for the table data type until v7, but that would be ok in my opinion.

brusch:
Scheduled a follow up issue for v7: #6661

NiklasBr:
Thanks @brusch

PR #6536, opened 2020-06-05
[Data Objects] Class Definition JSON Export should be independent fro…

### LINKED ISSUE #5548: [Data Objects] Class Definition JSON Export should be independent from REST Webservices
See: 
https://github.com/pimcore/pimcore/blob/263f13ceed1e0b53f04dc02c7fba94444dc86f3d/models/DataObject/ClassDefinition/Service.php#L40

weisswurstkanone:
resolves #5548


dvesh3:
should be changed for bulk export as well https://github.com/pimcore/pimcore/blob/0abcbeca8b11732d9165a8f49214376112e901b1/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php#L1657

weisswurstkanone:
done

PR #6538, opened 2020-06-05
fix brick route

dpfaffenbauer:
regression of #6146


brusch:
Grazie Mille 🙏👌 (even already to the right branch 😎)

PR #6539, opened 2020-06-05
Check for customConfig passed to ckeditor

roland4432:
Checking only for custom set toolbarGroups and (in case none is found) passing a default for toolbarGroups won't allow to overwrite the settings given in the customConfig file passed via options.

One could argue, that the customFile is not passing a toolbarGroup config, but the user could at least then overwrite the default settings from line 86 - 97 or accept the default given in ckeditor settings.


roland4432:
uuhh.. something stranged happend when committing the change to the docs.. there are 2 commits with the 9de6139 being empty? The 6f607bd was commited correctly though... what ever happened there, can you see the changes to the docs correctly @brusch ?

brusch:
Thank you very much!

PR #6545, opened 2020-06-08
Remove erroneous line-break from the docs

jdreesen:
There should be no line break here, because otherwise it results in being rendered as follows:
image


fashxp:
thx!

PR #4529, opened 2019-06-14
[XLIFF] Support for blocks in localized fields and field collections containing localized fields

mcuenez:
Added support for exporting and importing both **blocks in localized fields** and **field collections with localized fields**.

PR #6549, opened 2020-06-08
Fix xliff import overriding source values

msphn:
This is a bug fix for latest release and master. The bug was introduced with https://github.com/pimcore/pimcore/pull/4529/files.

Having a translated block doesn't work properly when importing an XLIFF file.

As seen in this screenshot, the language isn't changed, but the value is, that caused Pimcore to override it's original value instead of the source value. (Which by some magic later on is set anyway. - Didn't investigate)
Screenshot 2020-06-08 at 17 26 55

I enforce the '_language' variable in the block element which is supposed to be present when in context of a translation.

I hope I didn't miss something and I'd be happy to hear some feedback!


msphn:
Failed tests are not related to my change, right? Do they just die sometimes?

brusch:
@msphn Thanks for your PR! 
Since this is a bug fix for the latest release branch, please base this PR on the `6.6` branch. I tried to change the base, but that unfortunately didn't work as expected. 

Thanks a lot!

msphn:
@brusch done! :)

brusch:
Thanks a lot 😊

PR #6554, opened 2020-06-09
Fix to keep icon sizes regular and centered

NiklasBr:
Previously larger icons from the library would not look good, such as Twemoji and flags, because they did not properly fit into the available box.

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/Misc/admin-css.html.php
+++ b/bundles/AdminBundle/Resources/views/Admin/Misc/admin-css.html.php
@@ -23,7 +23,7 @@
             $treetype = $cv["treetype"] ? $cv["treetype"] : "object";
             ?>
     .pimcore_<?= $treetype ?>_customview_icon_<?= $cv["id"]; ?> {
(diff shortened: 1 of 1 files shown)


brusch:
This one needs to be based on `6.6` as well - thanks a lot!

NiklasBr:
As I forked from master it would be best to merge back to master because a lot of other things are following along then (like my other PR:s)

brusch:
Ok for this time, next time please base the PR on the right branch, otherwise we're getting in a mess 😉 Thanks a lot!

NiklasBr:
Yup, I never noticed that you had started to work on branches like this

brusch:
No problem at all, that's why we're telling you 😉
Btw. here are the updated guidelines: 
https://raw.githubusercontent.com/pimcore/pimcore/f1e4343e530c564e8e0c5cb485973bd960a39551/.github/PULL_REQUEST_TEMPLATE.md

brusch:
Just merged them to `master` and cherry-picked them to `6.6` - so will be available in `6.6.6` (incl. #6555 and #6564)

PR #6566, opened 2020-06-10
[Documents] Fix moving multiple documents for auto redirects

### LINKED ISSUE #6541: When moving multiple documents - Path value is only updated for one item, not all
# Bug Report

### Expected behavior
When selecting multiple documents in the document tree on the left side of the window and dragging them to another parent document and dropping them, the value of the `SEO & Settings -> Path, Key & ID -> Path` field is changed for every moved document.

### Actual behavior
After dropping the items, only the last active item has an updated and correct `Path` field value. The other items still have the old value (including their children - recursively).

### Steps to reproduce
- Log into admin area.
- Create 5 documents
  - Doc A
  - Doc B
  - Doc C
  - Doc D
  - Doc E
- Select all documents from Doc B (first selected) to Doc E (last selected) while holding your Shift key.
- Move selected documents (drag & drop) to Doc A so they become children of Doc A.
- Do not create links to keep SEO credit (click no on the popup).
- Have a look at the `SEO & Settings -> Path, Key & ID` section of each document from Doc B to Doc E.
- Only one document will have a correct `Path` set. All other documents will still have the old value.

### Structure
My folder structure looks like this right now (after moving):
image

The structure before was without the `home` document. It was created after all the other documents and every document was on the same layer, like this:
- en
  - faq
  - shop
  - contact-us
  - upcoming
  - home

I then moved all documents except `home` inside the `home` document resulting in:
- en
  - home
    - faq
    - shop
    - contact-us
    - upcoming

After that, every document (faq, shop, contact-us) except `upcoming` had the old `Path` value set.

dvesh3:
Resolves #6541

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
@@ -288,13 +288,12 @@ pimcore.document.tree = Class.create({
             (node.data.type === 'page' || node.data.type === 'hardlink') &&
             pimcore.globalmanager.get("user").isAllowed('redirects')
         ) {
-            var data = {
+            this.nodesToMove.push({
                 "id": node.data.id,
                 "params": params,
                 "moveCallback": moveCallback,
-            };
+            });
 
-            this.nodesToMove.push(data);
             // ask the user if redirects should be created, if node was moved to a new parent
             Ext.MessageBox.confirm("", t("create_redirects"), function (buttonValue) {
                 for (let nodeIdx in this.nodesToMove) {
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
```diff
@@ -284,12 +288,22 @@ pimcore.document.tree = Class.create({
...
+                "moveCallback": moveCallback,
+            };
+
+            this.nodesToMove.push(data);
             // ask the user if redirects should be created, if node was moved to a new parent
             Ext.MessageBox.confirm("", t("create_redirects"), function (buttonValue) {
+                let redirectParam = [];
                 if (buttonValue == "yes") {
-                    params['create_redirects'] = 'true';
+                    redirectParam['create_redirects'] = 'true';
```
brusch, line 302: Can't we just use `this.nodesToMove[nodeIdx]['params']['create_redirects'] = 'true';` here?

dvesh3: changed it.

PR #6575, opened 2020-06-10
[CSV Import] import broke due to added exception

### LINKED ISSUE #6548: [CSV Import] import broke due to added exception
# Bug Report

### Expected behavior
I make export of Car objects from demo.pimcore.fun

### Actual behavior
I successfully import the file

### Steps to reproduce
I get the error "Different number of columns"

The problem arose 5 days ago after the merger of this pull request
https://github.com/pimcore/pimcore/pull/6309

For some reason, after receiving the last line of $fields from the csv file, we get an "empty line" where $fields = [0 = null] (screen 1)
Because of this, every csv file exported from pimcore does not pass verification
Some files do not have this problem (for example, when there is 1 column in the csv file)
Снимок экрана от 2020-06-08 19-21-28
[export.zip](https://github.com/pimcore/pimcore/files/4746917/export.zip)

weisswurstkanone:
resolves #6548

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -1551,7 +1551,7 @@ class DataObjectHelperController extends AdminController
         }
 
         $csv = new \SplFileObject($originalFile);
-        $csv->setFlags(\SplFileObject::READ_CSV);
+        $csv->setFlags(\SplFileObject::READ_CSV | \SplFileObject::SKIP_EMPTY| \SplFileObject::READ_AHEAD);
         $csv->setCsvControl($dialect->delimiter, $dialect->quotechar, $dialect->escapechar);
         $rows = 0;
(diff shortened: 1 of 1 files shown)


albertmueller:
@weisswurstkanone i have an issue there, if you have a list of more then 20 objects, the line break is missing.
that generates a unusable csv in the end, because the latest element of the first do-export request will merged with the first element of the next do-export request.

 i found out where the issue happens:
```
 for ($i = 0; $i < $lineCount; $i++) {
            $line = $csv[$i];
            if ($addTitles && $firstLine) {
                $firstLine = false;
                $line = implode($delimiter, $line);
                fwrite($fp, $line);
            } else {
                fputs($fp, implode($delimiter, array_map([$this, 'encodeFunc'], $line)));
            }
            if ($i < $lineCount - 1) {
                fwrite($fp, "\r\n");
            }
        }
```

if i change the last if to:
```
if ($i < $lineCount) {
```

then it works correctly.
can you take a look at this issue please?

PR #6582, opened 2020-06-11
[Object search] Sort by class name

BlackbitNeueMedien:
With this PR it gets possible to sort data object search results by class name.


brusch:
Thank you very much! 🙏

PR #6583, opened 2020-06-11
Fix return type annotation

jdreesen:
`Pimcore\Model\Document\TargetingDocument::getElement()` often returns `parent::getElement()` which may return `null`:
https://github.com/pimcore/pimcore/blob/1e49e5cb106b1e701e979782fdf7cefaae96174d/models/Document/PageSnippet.php#L402-L404


brusch:
👍thanks a lot!

PR #6584, opened 2020-06-11
[Data objects] Option to disable inheritance for single fields

BlackbitNeueMedien:
Currently you can either enable inheritance for all fields in a class or completely disable inheritance. This PR adds the option to define on field level which fields shall be inheritable and which not.
To now this is only a proof of concept and only works for input fields.

Currently the `inheritable` setting does only apply if class inheritance is enabled but I would like to discuss whether the class-level inheritance checkbox would actually make sense if we could define inheritance on field level. 

Would love to hear some opinions and perhaps which pitfalls to watch for.


brusch:
To be honest, I don't think that is is currently a good idea. 
I'd rather prefer overriding the class and just custom implement the functionality (by using `\Pimcore\Model\DataObject\Concrete::getValueFromParent()` if needed for single fields.

PR #6585, opened 2020-06-11
Prevent `get_class(null)` calls

jdreesen:
They're not allowed anymore since PHP 7.2: https://wiki.php.net/rfc/get_class_disallow_null_parameter

This also removes some duplicated code.


brusch:
Nice, thanks a lot! 👌

PR #6595, opened 2020-06-12
Revert to latest phpstan release

weisswurstkanone:


NiklasBr:
Why?

weisswurstkanone:
Because I downgraded to an older version a couple of days ago

NiklasBr:
That was my question. Why revert PHPStan to an outdated version?

weisswurstkanone:
It isn't.
As you see in the build log it uses 0.12.28 now (instead of 0.12.25)

PR #6600, opened 2020-06-14
Notification events for add/delete

mloock:
Hi,

This PR is intented to bring new events regarding the notification feature.
This adds the pre/post add/update/delete events.
For example, this could be useful to send an email for a new notification.

### DIFF
--- a/lib/Event/Model/NotificationEvent.php
+++ b/lib/Event/Model/NotificationEvent.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Event\Model;
+
+use Pimcore\Event\Traits\ArgumentsAwareTrait;
+use Pimcore\Model\Notification;
+use Symfony\Component\EventDispatcher\Event;
+
+class NotificationEvent extends Event implements ElementEventInterface
+{
+    use ArgumentsAwareTrait;
+
+    /**
+     * @var Notification
+     */
+    protected $notification;
+
+    /**
+     * DataObjectEvent constructor.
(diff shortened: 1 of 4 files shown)


lib/Event/Model/NotificationEvent.php
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public function getNotification()
+    {
+        return $this->notification;
+    }
+
+    /**
+     * @param Notification $notification
+     */
+    public function setObject($notification)
```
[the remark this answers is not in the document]
mloock: You're right! :)

brusch:
I'd prefer a bit simpler event structure, with just `pre/post save` and `pre/post delete` just as other simple entities do, see https://github.com/pimcore/pimcore/blob/bf4799782bb6d10ebe2adab85d9eb555ce0b84f9/lib/Event/RedirectEvents.php#L17 for example. 

Can you please update the PR accordingly? 
Thanks a lot!

mloock:
> I'd prefer a bit simpler event structure, with just `pre/post save` and `pre/post delete` just as other simple entities do, see
> 
> https://github.com/pimcore/pimcore/blob/bf4799782bb6d10ebe2adab85d9eb555ce0b84f9/lib/Event/RedirectEvents.php#L17
> 
> for example.
> Can you please update the PR accordingly?
> Thanks a lot!

Hi @brusch !
Of course, I'll PR this in a few minutes ;)
The update events would have been useful for the first read action, because we update the actual notification object in this case.
But no problem, I amend that right now!

Thanks for the review

lib/Event/NotificationEvents.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Pimcore\Event;
+
+final class NotificationEvents
+{
+    /**
+     * @Event("Pimcore\Event\Model\NotificationEvent")
+     *
+     * @var string
+     */
+    const PRE_ADD = 'pimcore.notification.preAdd';
```
brusch, line 24: This should be `PRE_SAVE`

lib/Event/NotificationEvents.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * @Event("Pimcore\Event\Model\NotificationEvent")
+     *
+     * @var string
+     */
+    const POST_ADD = 'pimcore.notification.postAdd';
```
brusch, line 31: This should be `POST_SAVE`

brusch:
That looks good now! Thanks a lot!

PR #6603, opened 2020-06-15
[Asset] Add media query support for Video Thumbnails

### LINKED ISSUE #6364: [Asset] Media Query Support for Video Thumbnails
... just identical to the image thumbnails.

dvesh3:
Resolves #6364 

## Additional info  
Added media query support for video thumbnails
image
```php
$asset = Asset::getById(123);
$thumbnail = $asset->getThumbnail("myVideoThumbnail"); // returns an array
 /* OUTPUTS:
     Array(
       "mp4" => "/website/var/tmp/video.....mp4",
       "medias" => array:1 [
            "mp4" => array:2 [  
                  "(min-width: 576px)" => "/website/var/tmp/video..~-~%28min-width%3A%20576px%29.mp4"
                  "(min-width: 800px)" => "/website/var/tmp/video..~-~%28min-width%3A%20800px%29.mp4"
       )
*/

```


dvesh3:
Discussion: do we need frontend code generation? as it might be not supported in many browsers https://bugzilla.mozilla.org/show_bug.cgi?id=1513109 https://walterebert.com/playground/video/media-queries/

brusch:
Ok, it seems that [MPEG-DASH](https://en.wikipedia.org/wiki/Dynamic_Adaptive_Streaming_over_HTTP) is currently the only standardized way how to implement adaptive progressive video-streaming. 
Unfortunately it requires a polyfill for all major browsers: https://github.com/Dash-Industry-Forum/dash.js 

So i guess the best solution for now is that we generate a MPEG-DASH (`.mpd`) file using FFMPEG in case there are media queries defined, if no media queries are defined everything stays as it is now. The generated HTML markup also stays the same, just that we reference the single `.mpd` file. 
The polyfill is not included in the generated markup, it's up to the developer to care about that.

dvesh3:
> So i guess the best solution for now is that we generate a MPEG-DASH (.mpd) file using FFMPEG in case there are media queries defined, if no media queries are defined everything stays as it is now. The generated HTML markup also stays the same, just that we reference the single .mpd file.

@brusch For this approach, I believe we just need a config option to generate .mpd file and there's no need for media queries config since it is not used anyway for .mpd file generation. right?

brusch:
It seems that only the first media segment is included in the `mpd` file. 

I have the following thumbnail definition: 
```
<?php 

return [
    "bernie" => [
        "items" => [
            [
                "method" => "scaleByWidth",
                "arguments" => [
                    "width" => 100
                ]
            ]
        ],
        "medias" => [
            "200K" => [
                [
                    "method" => "scaleByWidth",
                    "arguments" => [
                        "width" => 200
                    ]
                ]
            ],
            "300K" => [
                [
                    "method" => "scaleByWidth",
                    "arguments" => [
                        "width" => 300
                    ]
                ]
            ],
            "400K" => [
                [
                    "method" => "scaleByWidth",
                    "arguments" => [
                        "width" => 400
                    ]
                ]
            ]
        ],
        "name" => "bernie",
        "description" => "",
        "group" => "",
        "videoBitrate" => 800,
        "audioBitrate" => 196,
        "modificationDate" => 1609765839,
        "creationDate" => 1609752951,
        "id" => "bernie"
    ]
];
```

The resulting `MPD`  is the following: 
```xml
<?xml version="1.0" encoding="utf-8"?>
<MPD xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xmlns="urn:mpeg:dash:schema:mpd:2011"
        xmlns:xlink="http://www.w3.org/1999/xlink"
        xsi:schemaLocation="urn:mpeg:DASH:schema:MPD:2011 http://standards.iso.org/ittf/PubliclyAvailableStandards/MPEG-DASH_schema_files/DASH-MPD.xsd"
        profiles="urn:mpeg:dash:profile:isoff-live:2011"
        type="static"
        mediaPresentationDuration="PT9.5S"
        minBufferTime="PT19.1S">
        <ProgramInformation>
        </ProgramInformation>
        <Period id="0" start="PT0.0S">
                <AdaptationSet id="0" contentType="video" segmentAlignment="true" bitstreamSwitching="true" lang="und">
                        <Representation id="0" mimeType="video/mp4" codecs="avc1.64000d" bandwidth="200000" width="400" height="224" frameRate="24000/1001">
                                <BaseURL>video-converter-Volkswagen-Van-stream0.mp4</BaseURL>
                                <SegmentList timescale="1000000" duration="9592917" startNumber="1">
                                        <Initialization range="0-807" />
                                        <SegmentURL mediaRange="808-222912" indexRange="808-859" />
                                </SegmentList>
                        </Representation>
                </AdaptationSet>
                <AdaptationSet id="1" contentType="audio" segmentAlignment="true" bitstreamSwitching="true">
                </AdaptationSet>
        </Period>
</MPD>
```

So as you can see there's only the 200k representation included. 
Thanks for checking 😊

brusch (changes requested):
Please have a look at the issue described above. Thanks!

dvesh3:
> Please have a look at the issue described above. Thanks!

@brusch fixed it. please review again. thanks!

brusch:
@dvesh3 Thanks for fixing! However, there's one small remaining issue, which is related to #8562 . 
The problem is, that the `mpd` file is indeed moved to the storage here: 
https://github.com/pimcore/pimcore/blob/0a330fb7a7205895123392db5f1f63c0fb15e146/models/Asset/Video/Thumbnail/Processor.php#L212 

But not the files which are referenced in the `mpd` file. 
This works as long as you're on a local file-system but not if e.g. using S3. 
So we need to parse the `mpd` file and move the files enclosed by `<BaseURL>` as well to the same location (if not already there), e.g.: `<BaseURL>temp-file-60644cf4de35d-602a7cdd663ca1e3a3b9692d4b2294-stream0.mp4</BaseURL>`

PR #6604, opened 2020-06-15
Fix xliff import overriding source values

msphn:
Followup to: https://github.com/pimcore/pimcore/pull/6549

@brusch I guess I changed the setValues method and forgot about it, using the original pimcore sources my patch didn't solve the problem. I extended the OwnerAwareInterface and trait to allow updating the language, we are not aware of the owning container at the import so it's the best I came up with. I hope the tests will still work, and sorry - was my mistake.

### DIFF
--- a/lib/Translation/ImporterService/Importer/DataObjectImporter.php
+++ b/lib/Translation/ImporterService/Importer/DataObjectImporter.php
@@ -19,6 +19,8 @@ use Pimcore\Model\Element;
 use Pimcore\Translation\AttributeSet\Attribute;
 use Pimcore\Translation\ExportDataExtractorService\DataExtractor\DataObjectDataExtractor;
 
+use function explode;
+
 class DataObjectImporter extends AbstractElementImporter
 {
     /**
@@ -70,12 +72,10 @@ class DataObjectImporter extends AbstractElementImporter
             $blockData = $element->{'get' . $blockName}($targetLanguage);
             $blockItem = !empty($blockData) && $blockData[$blockIndex] ? $blockData[$blockIndex] : $originalBlockItem;
 
-            // ensure correct target language
-            /* @var $blockItemField DataObject\Data\BlockElement */
-            $blockItemField = $blockItem[$fieldname];
-            $blockItemField->setValue('_language', $targetLanguage);
-
             $blockItemData = !empty($blockData) ? $blockItem[$fieldname] : clone $originalBlockItemData;
+
+            /* @var $blockItemData DataObject\Data\BlockElement */
+            $blockItemData->setLanguage($targetLanguage);
 
             $blockItemData->setData($attribute->getContent());
             $blockItem[$fieldname] = $blockItemData;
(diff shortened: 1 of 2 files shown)


brusch:
@msphn We can't change `OwnerAwareFieldInterface` for BC reasons. It's also used by not localized data-types, so that's not right anyway. So we need a different approach. Can you please check? 
Thanks a lot!

msphn:
Is it breaking? I'm kinda confused, setOwner is already accepting a language parameter and I'd just like to change it without calling setOwner again. It's a little messy already, the DataObjectImporter isn't aware of the elements container and therefore I struggle setting the language properly.

Should we try making the OwnerAwareFieldInterface fully language agnostic?

brusch:
Changing an interface is always breaking 😉
Why not call `setOwner()` with the right parameters?

msphn:
That's what I intedend, but I don't get the owning container in the DataObjectImporter, I thought I try to avoid bigger refactoring ..

brusch:
@weisswurstkanone do you have an idea maybe? Thanks!

weisswurstkanone:
Hello @msphn , could you please share more details on how to reproduce your problem? 
I.e. data model + steps.
Ideally on the demo.

I will have a look then.

msphn:
The model:
```
- translation container
   - block
       - input
```

Given the locales DE (source) and EN:

- fill values for DE (make sure nothing set in EN)
- export xliff
- use poedit or similar to translate the values, use unique ones like "EN Test 1" for "DE Test 1" or similar
- import the xliff

You will see, that the values will be set for the source language as well.

weisswurstkanone:
Can you attach both xliffs for reference ?

models/DataObject/OwnerAwareFieldInterface.php
```diff
@@ -20,9 +20,16 @@
...
      * @param string|null $language
      */
     public function setOwner($owner, string $fieldname, $language = null);
+
+    /**
+     * @param string $language
+     *
+     * @return void
+     */
+    public function setLanguage(string $language);
```
weisswurstkanone, line 34: could be dangerous anyway if there is already a setter called "setLanguage". guess we should add it in pimcore 7 but do it in a different way.

but you could  add a setter directly to BlockElement and use it instead of setValue

msphn: I initially thought about this but abandoned it because it would make the actual language unaware class aware of a property introduced by a trait, that felt unnatural.

But if you like that more I can do it.

weisswurstkanone: I would say this is the only option without a BC break

msphn: Yes, I agree, will do so.

weisswurstkanone: thx, could you update the PR accordingly?

msphn: Ill do tomorrow!

msphn:
I'll updated, let's wait what the CI says.

lib/Translation/ImporterService/Importer/DataObjectImporter.php
```diff
@@ -19,6 +19,8 @@
 use Pimcore\Translation\AttributeSet\Attribute;
 use Pimcore\Translation\ExportDataExtractorService\DataExtractor\DataObjectDataExtractor;
 
+use function explode;
```
weisswurstkanone, line 22: can you revert this line please?

msphn: I can do but it's considered best practice in symfony.

weisswurstkanone:
thx!

PR #6605, opened 2020-06-15
Update ClassificationstoreController.php

mc-supply:
Fix route name for /groups (POST/PUT)

  
  

Resolves #


weisswurstkanone:
Isn't this fixed already via https://github.com/pimcore/pimcore/pull/6580 ?

mc-supply:
Looks like it is.  Didn't see that.

blankse:
The 6.6 branch isn't merged to master yet.

brusch:
Yes, we do that just from time to time or after a bug fix release (which is common practice). 
In general, bug fixes only go into the latest release branch, so e.g. `6.6`. 
See also: https://github.com/pimcore/pimcore/blob/master/CONTRIBUTING.md#contributing-pull-requests 😉

Thanks anyway!

PR #6606, opened 2020-06-16
Localized classification store tab label translations

kjkooistra-youwe:
Admin translations are not applied to localized classification store tab labels. Also see https://github.com/pimcore/pimcore/pull/6022


kjkooistra-youwe:
Admin_translations

weisswurstkanone:
thanks a lot!

PR #6607, opened 2020-06-16
Update presta/sitemap-bundle

### LINKED ISSUE #6340: Upgrade to PrestaSitemapBundle v2
# Feature Request
## Feature description
PrestaSitemapBundle v2 has been released on January 8th and requires at least PHP 7.1 as well as Symfony 3.4. Therefore, I think it could be upgraded without any problems. As far as I have seen there have no changes to be made to the Pimcore code: https://github.com/prestaconcept/PrestaSitemapBundle/releases/tag/v2.0.0

What do you think?

aarongerig:
This PR changes the version constraint of presta/sitemap-bundle to version 2.1 and above.
Resolves #6340 

## Additional info
Also removes the ugly deprecation messages when running `composer update`. ;)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -74,7 +74,7 @@
     "pear/net_url2": "^2.2",
     "phive/twig-extensions-deferred": "^2.0",
     "piwik/device-detector": "^3.9",
-    "presta/sitemap-bundle": "^2.1",
+    "presta/sitemap-bundle": "^1.5 || ^2.1",
     "ramsey/uuid": "^3.8",
     "sabre/dav": "^3.2",
     "sensio/framework-extra-bundle": "^5.4",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -74,7 +74,7 @@
     "pear/net_url2": "^2.2",
     "phive/twig-extensions-deferred": "^2.0",
     "piwik/device-detector": "^3.9",
-    "presta/sitemap-bundle": "^1.5",
+    "presta/sitemap-bundle": "^2.1",
```
blankse, line 77: Should we allow both versions to be on the safe side? So somebody who need version 1 can set it to the own composer.json and can downgrade it.

```suggestion
    "presta/sitemap-bundle": "^1.5 || ^2.1",
```

brusch:
Thanks a lot!

PR #6612, opened 2020-06-16
Dataloss in localized fields and maps

### LINKED ISSUE #6509: Dataloss in localized fields and maps
Class loses all input of localized fields and points on the maps. Using 26 languages.

[bulk_export (2).json.txt](https://github.com/pimcore/pimcore/files/4722362/bulk_export.2.json.txt)

weisswurstkanone:
resolves #6509


models/DataObject/Localizedfield/Dao.php
```diff
@@ -167,7 +167,7 @@ public function save($params = [])
                             )
                             || (((!$container instanceof DataObject\Fieldcollection\Definition || $container instanceof DataObject\Objectbrick\Definition)
                                     && $this->model->isLanguageDirty($language))
-                                || $params['saveRelationalData']['saveLocalizedRelations'])) {
+                                || ($params['saveRelationalData']['saveLocalizedRelations'] ?? false))) {
```
blankse, line 170: We could simplify the if when we wrote that in a variable beforehand:
`$saveLocalizedRelations = $params['saveRelationalData']['saveLocalizedRelations'] ?? false;`

What do you think?

blankse: @weisswurstkanone It's not the only line. You could also save the isset calls 6/7 lines above...

models/DataObject/Localizedfield/Dao.php
```diff
@@ -157,17 +157,14 @@ public function save($params = [])
                     }
 
                     $isUpdate = isset($params['isUpdate']) && $params['isUpdate'];
-                    $childParams = $this->getFieldDefinitionParams($fd->getName(), $language, ['isUpdate' => $isUpdate, 'context' => $context]);
+                    $childParams = $this->getFieldDefinitionParams($fd->getName(), $language, ['isUpdate' => $isUpdate]);
```
blankse, line 160: This should stay ;)

```suggestion
                    $childParams = $this->getFieldDefinitionParams($fd->getName(), $language, ['isUpdate' => $isUpdate, 'context' => $context]);
```

weisswurstkanone: travis will complain ;-)

PR #6615, opened 2020-06-16
Update Pimcore DB master/slave to primary/replica

NiklasBr:
Following [Doctrine/DBAL's recent deprecation](/doctrine/dbal/commit/7047798fbb5484703e2f628ab6110b08f887d550) of master/slave in favour of primary/replica Pimcore should do the same


brusch:
Do we need to increase the DBAL dependency as well? I guess so right?

NiklasBr:
It applies to the [future version](https://github.com/doctrine/dbal/milestone/77)  and no changes to Pimcore's composer.json should be needed. The current logic (`"doctrine/dbal": "^2.8.0",`) allows `2.10.2` to be installed and should allow `2.11` when released.

brusch:
Well I'm just thinking if you're on an existing installation and you want to follow the docs .. might be confusing, but yes, nevermind, I can live with that 😉

Thanks a lot! (as always 😎)

NiklasBr:
No hurry to merge :)

PR #6616, opened 2020-06-16
[CS-Fixer] Add trailing comma in multiline array rule

blankse:
Also Symfony do this:
https://symfony.com/doc/current/contributing/code/standards.html#structure
`Add a comma after each array item in a multi-line array, even after the last one`


brusch:
Thanks a lot!

PR #6627, opened 2020-06-18
Fix issues with shutdown when running cache warming / building the container 

### LINKED ISSUE #4709: Changes related to #4555 broke cache warmup.
The changes related to #4555 seem to have broken the cache clearing / warming (`bin/console cache:clear`).
   1. https://github.com/pimcore/pimcore/commit/8c41e8155805b16dc2d6c6aecb77e2ad871f452c
   2. https://github.com/pimcore/pimcore/commit/4f7725c70586767fc5ddd7e0fb8399adffd0be0f#diff-ff0c921492cef999788fd76dcb807a22R306

The second change seems to trigger following error:
```
Warning: Uncaught ErrorException: Warning: require(/var/www/pimcore/var/cache/in_/ContainerLEnsbem/removed-ids.php): failed to open stream: No such file or directory in /var/www/pimcore/var/cache/int/ContainerLEnsbem/appAppKernelIntDebugContainer.php:634
Stack trace:
#0 /var/www/pimcore/var/cache/int/ContainerLEnsbem/appAppKernelIntDebugContainer.php(634): require()
#1 /var/www/pimcore/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Container.php(161): ContainerLEnsbem\appAppKernelIntDebugContainer->getRemovedIds()
#2 /var/www/pimcore/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Container.php(392): Symfony\Component\DependencyInjection\Container->set('container.env_v...', Object(Symfony\Component\DependencyInjection\ServiceLocator))
#3 /var/www/pimcore/var/cache/int/ContainerLEnsbem/appAppKernelIntDebugContainer.php(1259): Symfony\Component\DependencyInjection\Container->getEnv('trim:file:DB_US...')
#4 /var/www/pimcore/vendor/symfony/symfony/src/Symfony/Component/DependencyInjectio in /var/www/pimcore/var/cache/int/ContainerLEnsbem/appAppKernelIntDebugContainer.php on line 634

Fatal error: ContainerLEnsbem\appAppKernelIntDebugContainer::getRemovedIds(): Failed opening required '/var/www/pimcore/var/cache/in_/ContainerLEnsbem/removed-ids.php' (include_path='/var/www/pimcore/vendor/pear/archive_tar:/var/www/pimcore/vendor/pear/console_getopt:/var/www/pimcore/vendor/pear/net_url2:/var/www/pimcore/vendor/pear/pear-core-minimal/src:/var/www/pimcore/vendor/pear/pear_exception:.:/usr/local/lib/php') in /var/www/pimcore/var/cache/int/ContainerLEnsbem/appAppKernelIntDebugContainer.php on line 634
```

It seems like the used `\Pimcore\Db::get()` in `\Pimcore::isInstalled()` doesn't work properly during a cache warmup.
During a cache warmup the return values of `Kernel->getCacheDir()` and `Kernel->getContainer()->getParameter('kernel.cache_dir')` are incosistent.
Symfonys UPGRADE-4.0.md says:
>  The `getCacheDir()` method of your kernel should not be called while building the container.
>  Use the `%kernel.cache_dir%` parameter instead. Not doing so may break the `cache:clear` command.

The problem seems to be that the Environment used in `\Pimcore\Db::get()` will try to access the temporary warmup directory which in the meantime already has been moved into place of the regular cache.
This is because the warmup is completed before our shutdown handler runs and the environment isn't cleaned up.

The only thing I could think of for the moment to work around this is a condition like this:
```php
        // Check if this is a cache warming run and if this runs on an installed instance. If this is a cache warmup
        // we can't use self::isInstalled() as it will refer to the wrong caching dir.
        if (self::getKernel()->getCacheDir() === self::getContainer()->getParameter('kernel.cache_dir') && self::isInstalled()) {
```
So `bin/console cache:clear --no-warmup` will go into the condtion while `bin/console cache:clear` will skip it.

brusch:
fixes #4709

### DIFF
--- a/bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
+++ b/bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
@@ -56,6 +56,10 @@ class PimcoreCoreExtension extends ConfigurableExtension implements PrependExten
      */
     public function loadInternal(array $config, ContainerBuilder $container)
     {
+        // on container build the shutdown handler shouldn't be called
+        // for details please see https://github.com/pimcore/pimcore/issues/4709
+        \Pimcore::disableShutdown();
+
         // performance improvement, see https://github.com/symfony/symfony/pull/26276/files
         if (!$container->hasParameter('container.dumper.inline_class_loader')) {
             $container->setParameter('container.dumper.inline_class_loader', true);
--- a/lib/Pimcore.php
+++ b/lib/Pimcore.php
@@ -40,6 +40,11 @@ class Pimcore
     private static $inShutdown = false;
 
     /**
+     * @var bool
+     */
+    private static $shutdownEnabled = true;
+
+    /**
      * @var KernelInterface
      */
     private static $kernel;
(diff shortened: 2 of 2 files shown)


bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
```diff
@@ -56,6 +56,10 @@ public function getAlias()
      */
     public function loadInternal(array $config, ContainerBuilder $container)
     {
+        // on container build the shutdown handler shouldn't be called
+        // for details please see https://github.com/pimcore/pimcore/issues/4709
+        \Pimcore::disableShutdown();
```
jdreesen, line 61: Shouldn't it be enabled again anywhere?

brusch:
No, when building the contaier it should stay just off.

PR #6634, opened 2020-06-19
[Workflow] fix #6013 initial place empty on add with `single_state` store

### LINKED ISSUE #6013: [Workflows] Initial place of single_state workflows does not get saved in database
Steps to reproduce:
1. Create Class ABC with input field `workflowState`
2. Create workflow which supports `Pimcore\Model\DataObject\ABC` and has initial place `started`, and set
```yml
marking_store:
      type: single_state
      arguments:
         - workflowState
```
3. Create object of class ABC and publish it

Expected behaviour: Field `workflowState` should contain `started`
Current behaviour: Field `workflowState` is empty

Tested in Pimcore 6.3.3

macghriogair:
- fixes invalid place "" exception as of Symfony >= 4.4.8
- set place on load for compat with objects created prior to fix

Resolves #6013 

## Additional info  

Tested mainly with marking store of type `single_state` where on adding objects the initial place was not saved. The custom [StateTableMarkingStore](https://github.com/pimcore/pimcore/blob/master/lib/Workflow/MarkingStore/StateTableMarkingStore.php) seems not to be affected by the bug b/c loose checking is still applied (also, on object add the initial place get's stored to the `element_workflow_state` table).

Setting the initial place programmatically in [onAdminElementGetPreSendData](https://github.com/macghriogair/pimcore/blob/bugfix/issue-6013-initial-workflow-place-on-add/bundles/CoreBundle/EventListener/WorkflowManagementListener.php#L167) seems necessary to me as there might be (as in our case) objects persisted with empty string in the db. Feel free to suggest any alternatives. (The method could be run by a batch job as well or via maintenance, but this is maybe too case specific).

### DIFF
--- a/lib/Workflow/Manager.php
+++ b/lib/Workflow/Manager.php
@@ -336,12 +336,11 @@ class Manager
     }
 
     /**
-     * Compatibility Fix: forces an initial place being set (and stored) if the current place is empty.
+     * Forces an initial place being set (and stored) if the current place is empty.
      * We cannot apply a regular transition b/c it would be considered invalid by the state machine.
      *
      * As of Symfony 4.4.8 built-in implementations of @see \Symfony\Component\Workflow\MarkingStore\MarkingStoreInterface
-     * use strict `null` comparison when retrieving the current marking. This will throw exceptions for elements
-     * stored with empty marking despite an initial state was configured in prior versions of Pimcore.
+     * use strict `null` comparison when retrieving the current marking and throw an exception otherwise.
      *
      * @param Asset|Concrete|PageSnippet $subject
      *
@@ -349,7 +348,7 @@ class Manager
      *
      * @throws \Exception
      */
-    public function ensureBackwardCompatibleInitialPlace(string $workflowName, $subject): bool
+    public function ensureInitialPlace(string $workflowName, $subject): bool
     {
         if (!$workflow = $this->getWorkflowIfExists($subject, $workflowName)) {
             return false;
--- a/lib/Workflow/Manager.php
+++ b/lib/Workflow/Manager.php
@@ -23,6 +23,7 @@ use Pimcore\Model\Element\AbstractElement;
 use Pimcore\Model\Element\ValidationException;
 use Pimcore\Workflow\EventSubscriber\ChangePublishedStateSubscriber;
 use Pimcore\Workflow\EventSubscriber\NotesSubscriber;
+use Pimcore\Workflow\MarkingStore\StateTableMarkingStore;
 use Pimcore\Workflow\Place\PlaceConfig;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Workflow\Exception\InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


weisswurstkanone:
could you please have a look at the unit tests ?
https://travis-ci.com/github/pimcore/pimcore/jobs/351372916

macghriogair:
updated, let's see if phpstan evaluates  `version_compare` correctly 🤞

brusch:
@macghriogair since this is a bugfix and should go into the next bugfix release, it would be good if you can rebase your branch on `6.6` and then change the merge target of this PR to `6.6` as well. 
Thanks a lot!

macghriogair:
@brusch done

lib/Workflow/Manager.php
```diff
@@ -333,4 +334,74 @@ public function getTransitionByName(string $workflowName, string $transitionName
...
+     * use strict `null` comparison when retrieving the current marking. This will throw exceptions for elements
+     * stored with empty marking despite an initial state was configured in prior versions of Pimcore.
+     *
+     * @param Asset|Concrete|PageSnippet $subject
+     *
+     * @return bool true if initial state was applied
+     *
+     * @throws \Exception
+     */
+    public function ensureBackwardCompatibleInitialPlace(string $workflowName, $subject): bool
```
fashxp, line 352: Actually this is not only a b/c fix, but also is needed, when a new workflow is configured for existing data objects. Or do I miss something there? 
So maybe we should name this method something like `ensureInitialPlace`

lib/Workflow/Manager.php
```diff
@@ -333,4 +334,74 @@ public function getTransitionByName(string $workflowName, string $transitionName
...
+
+        return false;
+    }
+
+    public function getInitialPlacesForWorkflow(Workflow $workflow): array
+    {
+        $initialPlaces = [];
+        $definition = $workflow->getDefinition();
+
+        if(method_exists($definition, 'getInitialPlaces')) {
```
fashxp, line 397: I would prefer to use `$definition->getInitialPlace()` for Pimcore 6 and switch to `$definition->getInitialPlaces()` with Pimcore 7 - when we need to get rid of all deprecation messages for symfony 5 compatibility anyway. 

This construct here blows up the code and we will never find it again when we are starting with Pimcore 7.

lib/Workflow/Manager.php
macghriogair: True, if a new workflow with initial place is added _or_ an initial place is configured for a workflow that had none previously, all existing objects need to be handled. Will change the method name 👍

lib/Workflow/Manager.php
macghriogair: I get your point, but does it fit into the concept of [multiple initial places](https://symfony.com/blog/new-in-symfony-4-3-workflow-improvements#allow-to-configure-many-initial-places)?

`$definition->getInitialPlace()` will return only the 1st element. 🤔

```
    /**
     * @deprecated since Symfony 4.3. Use getInitialPlaces() instead.
     */
    public function getInitialPlace(): ?string
    {
        /// ...
        return reset($this->initialPlaces);
    }
```
(on the other hand I'm not quite sure if Pimcore 6 supports multiple initial places)

fashxp: good point ... multiple initial places should be supported since #5541

then I guess we need to do it the way it is 👍

fashxp:
thank you very much.

PR #6647, opened 2020-06-22
[QuantityValue] Fix notice: Undefined offset: 1

blankse:
image


blankse:
I wonder why the unit id is used. This is not the same in different systems. Wouldn't it be better to use the abbreviation and the space as a separator? Then you could also import quantities from outside. I can create a new PR for the master branch if you want. Of course with BC layer.

brusch:
@blankse I agree - @weisswurstkanone what about you?

PR #6649, opened 2020-06-23
Improvement - Set allowed units to empty array when saving a quantity value attribute.

### LINKED ISSUE #6641: Improvement - Set allowed units to empty array when saving a quantity value attribute.
## Description
When adding a new unit for quantity values in Pimcore, we must redefine the class definitions to allow the unit to be able to use them, instead of just leaving no unit selected in the class definition so that by default all units are allowed.

### Expected behavior
We would expect that this attribute behaves the same as an object relation. On object relations: if no classes are selected as allowed, then all classes are allowed. The same would be expected for quantity value units.

### Actual behavior
When we don't select allowed units, then only the units at the time of saving will be allowed for the attribute, any attributes made on-demand will not be added to the list of allowed units.

### Steps to reproduce
1. Add a quantity value attribute to a class.
2. Create a unit on-demand programmatically 
3. Set unit on attribute programmatically
4. The unit will not show on the object since it's not in the collection of allowed units.

weisswurstkanone:
resolves #6641 

this will reset the unit cache if we find out that the unit id is not in the store


brusch:
@weisswurstkanone please rebase onto the master. Thanks!

PR #6651, opened 2020-06-23
Grid export - move logic to Service

ckogler-elements:
Move the logic for the grid export to the Service and call the Service functions in the Controller


msphn:
The current service is a beast, what about separating the export logic?

weisswurstkanone:
@msphn , would make sense. First step is to pull it out of the controller :-)

msphn:
That's absolutely a good idea ;D

PR #6652, opened 2020-06-23
Correction of reading ACL for tree element

roland4432:
Sort rows correctly to ensure that access rights for specific document are fully taken into account

  
  

Resolves issue #6654 with overlapping roles for the same tree element

## Additional info  

With the old select statement and the following setup, this will get false values for checking if a right has to be given or not

2 Roles:

GiveRight  
| cid  | cpath  | userId  | list  | view  | ... |
|---|---|---|---|---|---|
| 2 | /tree-element  | 100  | 1  | 1  | ...  |

DeclineRight  
| cid  | cpath  | userId  | list  | view  | ... |
|---|---|---|---|---|---|
| 2 | /tree-element  | 100  | 0  | 0  | ...  |

=> expected result: the user should be able to list and view the element _/tree-element_, but he won't since:
```
$permissionsParent = $this->db->fetchOne('SELECT `' . $type . '` FROM users_workspaces_document WHERE cid IN (' . implode(',', $parentIds) . ') AND userId IN (' . implode(',', $userIds) . ') ORDER BY LENGTH(cpath) DESC, ABS(userId-' . $user->getId() . ') ASC LIMIT 1');
```
**can** (the sorting might be random when none of the given parameters for sort makes a different, eg. no user specific workspace is set for this element) result (without the limit 1) in something like this:

| type (eg list) |
|---|
| 0 |
| 1 |

and the *Limit 1* will then ommit the second result row which has the correct right for this case.  
I therefore changed the ordering to first check for user specific rights on the element
```
FIELD(userId, ' . $user->getId() . ') DESC
```
and then check (if no user specific right is set) for the rights delivered from the results by sorting DESC (to see if there is a 1 in the resultset)
```
`' . $type . '` DESC 
```

It's tricky to get a setup which reproduces the error, but it does exist.


brusch:
This needs to be checked in detail as this could break compatibility in certain cases.

roland4432:
Regarding the SQL syntax or semantic?

brusch:
Both 😉

roland4432:
Can I collaborate in any way?

brusch:
@roland4432 not really, I think it's important that someone else has a critical view on the changes 😉

brusch:
@roland4432 since this is a bug fix, can you please rebase this PR on the `6.6` branch and also change the merge target to `6.6` please? Thanks a lot!

roland4432:
moved to #6670

PR #6667, opened 2020-06-25
[QuantityValue] CSV Export/Import with abbreviation instead of unit ID

blankse:
See comment in https://github.com/pimcore/pimcore/pull/6647

### DIFF
--- a/models/DataObject/ClassDefinition/Data/QuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/QuantityValue.php
@@ -412,8 +412,12 @@ class QuantityValue extends Data implements ResourcePersistenceAwareInterface, Q
     public function getForCsvExport($object, $params = [])
     {
         $data = $this->getDataFromObjectParam($object, $params);
-        if ($data instanceof \Pimcore\Model\DataObject\Data\QuantityValue) {
-            return $data->getValue() . '_' . $data->getUnitId();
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -430,15 +434,26 @@ public function getForCsvExport($object, $params = [])
...
-            $number = (float) str_replace(',', '.', $values[0]);
-            $value = new Model\DataObject\Data\QuantityValue($number, $values[1]);
+            return new Model\DataObject\Data\QuantityValue($number, $unitId);
         }
 
-        return $value;
+        if (strpos($importValue, ' ') !== false) {
+            [$number, $abbreviation] = explode(' ', $importValue);
+            $number = (float)str_replace(',', '.', $number);
+            $unit = Model\DataObject\QuantityValue\Unit::getByAbbreviation($abbreviation);
```
brusch, line 447: I think for compatibility reasons we should still support the import with unit IDs, shouldn't be an issue, right?

blankse: Is there in lines 437-442

brusch: 🤦‍♂️ Sorry - thanks a lot! 😊

PR #6668, opened 2020-06-25
[Staticroute] allow to define non-required parameters

dpfaffenbauer:
As a Bundle provider, I wan't to supply the Developer with default routes and I want to use these routes within my Bundle. In various use-cases, these static routes get changed. (eg. from `/en/shop` to just `/`). Since my routes require certain parameters (eg. `_locale`), which get set on route-generation, Pimcore attaches these Params as GET Parameters `?_locale=en`, which is not necessary in some cases. 

So basically we don't want to attach certain parameters as GET Parameters when not directly set within the reverse route.

This PR allows to define non-required parameters for a static route.

### DIFF
--- a/models/Staticroute.php
+++ b/models/Staticroute.php
@@ -83,6 +83,11 @@ class Staticroute extends AbstractModel
     public $methods;
 
     /**
+     * @var array
+     */
+    public $nonRequiredParams = [];
+
+    /**
      * @var int
      */
     public $priority = 1;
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
PS: UI is still missing, I wanna know if you're happy with it

brusch:
Basically the idea would be fine for me, but I'm not entirely happy with the naming, because actually the parameters are not "non-required" but more "ignored" on assembling the link, right? 
I'm also not really happy with "ignored" but maybe there are other ideas 😊

aarongerig:
Those parameters would only be ignored, if they are not set within the reverse route. Because of that they are **not always** ignored, but rather not always mandatory (= non-required). Therefore I'd suggest the following naming alternatives:

- `optional` parameters
- `arbitrary` parameters

brusch:
Hmm, they are just ignored for the `query`, so maybe `ignoredQueryParams`?

dpfaffenbauer:
So, what are we using now?

brusch:
I'd go for `ignoredQueryParams` as it is more specific and meaningful in my opinion 😐

brusch:
🏓😊

dpfaffenbauer:
will do next week ;)

brusch:

Issue #6501, opened 2020-06-02
[UI] Disabled toolbar buttons are hard to see


ultramegatom:
By default the toolbar buttons are hard to see when disabled.

Proposal
```css
.pimcore_main_toolbar .x-btn.x-btn-disabled {
    background-color: #484B4E;
    opacity: 1;
}

.pimcore_main_toolbar .x-btn.x-btn-disabled .x-btn-inner {
    color: #8e8e8e;
}
```
![image](https://user-images.githubusercontent.com/7206334/83517877-f4320a00-a4d9-11ea-8a32-fa0d76d48ad5.png)

andreas-gruenwald:
@ultramegatom Maybe you can prepare a PR, wouldn't be that much effort?

ultramegatom:
if i get feedback that the proposal mentioned above is fine then yes i can certainly copy this 3 css rules into a pull request

ultramegatom:
https://github.com/pimcore/pimcore/pull/6675

PR #6675, opened 2020-06-25
[UI] Disabled toolbar buttons are hard to see

ultramegatom:
see https://github.com/pimcore/pimcore/issues/6501


brusch:
Thanks a lot! 🙏😊

PR #6678, opened 2020-06-25
[REST Webservice] Support overridden document classes

BlackbitNeueMedien:
When you override a document class, its data is not accessible via REST API.
Steps to reproduce:
1. Put the followiing to app/config/config.yml:
```yml
pimcore:
    models:
        class_overrides:
            'Pimcore\Model\Document\Page': 'AppBundle\Model\Document\Page'
```
2. Clear cache.
3. Create page document
4. Access document via https://\<Pimcore-Hostname\>/webservice/rest/document/id/\<ID of new document\>

PS: I know, REST webservice is deprecated but as long as it is not removed we should make it work correctly.

### DIFF
--- a/models/Webservice/Service.php
+++ b/models/Webservice/Service.php
@@ -57,8 +57,7 @@ class Service
         try {
             $folder = Document::getById($id);
             if ($folder instanceof Document\Folder) {
-                $className = Webservice\Data\Mapper::findWebserviceClass($folder, 'out');
-                $apiFolder = Webservice\Data\Mapper::map($folder, $className, 'out');
+                $apiFolder = Webservice\Data\Mapper::map($folder, '\\Pimcore\\Model\\Webservice\\Data\\Document\\Folder\\Out', 'out');
 
                 return $apiFolder;
             }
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Can you explain the relevant change? This works for me without the change in Service.php

BlackbitNeueMedien:
Currently for documents `Webservice\Data\Mapper::findWebserviceClass()` gets called to get the Webservice model class (note that this is only done for documents whereas for assets and data objects this method is not used):
https://github.com/pimcore/pimcore/blob/030b0e711367d6a1fd3cf39674f812568245323f/models/Webservice/Data/Mapper.php#L36-L76

In this method in my case `$object` is an instance of `AppBundle\Model\Document\Page`. In the method the condition https://github.com/pimcore/pimcore/blob/030b0e711367d6a1fd3cf39674f812568245323f/models/Webservice/Data/Mapper.php#L56 is true. There `$orgclass = str_replace('Pimcore\\Model\\', '', get_class($object));` gets executed - which results in `$orgclass` being `AppBundle\Model\Document\Page` (there is no replacement). As of course this class is not in the array https://github.com/pimcore/pimcore/blob/030b0e711367d6a1fd3cf39674f812568245323f/models/Webservice/Data/Mapper.php#L38-L49 the returned API class is `AppBundle\Model\Document\Page`: https://github.com/pimcore/pimcore/blob/030b0e711367d6a1fd3cf39674f812568245323f/models/Webservice/Data/Mapper.php#L68 - and as this does not have any public properties the REST API call export a quite empty object.

I have to admin that the project where I need that does not have the current Pimcore version but when I look at the current code I do not see any difference concerning this REST webservice part.

weisswurstkanone:
Got it, thanks a lot!

PR #6685, opened 2020-06-26
Fix: FOSJsRouting for symfony 3.4

### LINKED ISSUE #6648: Pimcore Update from 6.0.5 to >6.6.4
# Bug Report
As per https://github.com/pimcore/pimcore/pull/6146 one cannot login to Pimcore anymore.
According to Chrome's Dev-Tools, there is an error when `/js/routing` is called.

This also means that any sub-sequent route fails too, which leads to being stuck with the loading spinner.

The root cause is
`The controller for URI "/js/routing" is not callable: Class "fos_js_routing.controller" does not exist.`

## Workaround
Downgrade to Pimcore 6.6.4.

blankse:
Resolves #6648

### DIFF
--- a/bundles/CoreBundle/Resources/config/routing.yml
+++ b/bundles/CoreBundle/Resources/config/routing.yml
@@ -36,7 +36,8 @@ _pimcore_service_thumbnail:
         filename: '.*'
 
 _fos_js_routing:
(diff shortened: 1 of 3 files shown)


dpfaffenbauer:
awesome :)

dvesh3:
@NiklasBr thanks a lot! great work 👏

PR #6689, opened 2020-06-29
Performance enhancements when deleting elements

BlackbitNeueMedien:
This PR consists of 2 changes:
1. Slight modification of `RecycleBinController::addAction()`: 
Refactored to use `getList()` method which all deletable elements have. And it now also checks for unpublished child elements (in the check if there are more than 100 children).
1. Change of `AbstractObject::delete()`: 
    - It first checked for `hasChildren()` and if this was `true` called `getChildren()` - this is redundant because `hasChildren()` does a database query and afterwards we fetch the children via `getChildren()` which needs an additional database query. Currently only `hasChildren()` can reuse the results of `getChildren()` but not vice-versa.
    - Currently the dependencies of the deleted element get fetched. This is not necessary as in the next line they get cleared by `cleanAllForElement()`.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
+++ b/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
@@ -157,7 +157,7 @@ class ElementControllerBase extends AdminController
 
                 if ($hasChilds) {
                     // get amount of childs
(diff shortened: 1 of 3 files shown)


brusch:
Thanks a lot! 👍

PR #6693, opened 2020-06-29
[DataObject] Possible inconsistent query table when missing entries are recreated

### LINKED ISSUE #6691: [DataObject] Possible inconsistent query table when missing entries are recreated
when rows in query tables are missing, they are recreated. this can result in inconsistent query table entries when certain data attributes are not dirty. 

steps to reproduce in demo application: 
- execute `DELETE FROM object_query_AP WHERE oo_id = 408;`
- save data object with id `408` in Pimcore backend without doing any changes 
- execute `SELECT * FROM object_query_AP WHERE oo_id = 408` 

--> columns like `manufacturer__id`, `manufacturer__type`, `series__id`, `series__type`, `mainCategory__id` and `mainCategory__type` are empty.

weisswurstkanone:
resolves #6691

### DIFF
--- a/models/DataObject/Concrete/Dao.php
+++ b/models/DataObject/Concrete/Dao.php
@@ -185,6 +185,14 @@ class Dao extends Model\DataObject\AbstractObject\Dao
     {
         parent::update($isUpdate);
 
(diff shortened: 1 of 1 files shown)


fashxp:
since this is quite an edge case scenario, we decided to avoid additional database queries and close this in favor of https://github.com/pimcore/pimcore/commit/2f7afb5533ef8e8de475dc6c26d7294c2ce56ef8

PR #6694, opened 2020-06-29
[Areablock] Fix Notice: Undefined index

blankse:
After #6515 I get following notice in every Areablock tag:
```
Notice: Undefined index: xyz in vendor/pimcore/pimcore/models/Document/Tag/Areablock.php (line 218)
```


dvesh3:
thanks a lot!

PR #6711, opened 2020-06-30
Fix JS error in areablock when no limits are set

### LINKED ISSUE #6710: JS error in editables in master branch
# Bug Report

I have multiple bricks with wysiwyg editables and edit button in one document.

## 6.6 Branch
All is okay.

## master Branch
Only the first brick has an edit button and there is an error in the Chrome console:
```
TypeError: Cannot read property 'wysiwyg' of undefined
```
This error is from here:
https://github.com/pimcore/pimcore/blob/7a5acaa75d7f8fbe614824905314c6a66135c34a/bundles/AdminBundle/Resources/public/js/pimcore/document/edit/startup.js#L153

I want to test it in the pimcore demo. But there is branch 6.6. Exists a demo with master branch?

blankse:
Resolves #6710 

## Additional info  
Regression from #6515


dvesh3:
thanks a lot! 🤝

PR #6712, opened 2020-06-30
fix seo panel tree route

solverat:
SEO Document editor does not work because of wrong route definition. Regression of https://github.com/pimcore/pimcore/pull/6146.


blankse:
@solverat I think this should be done for 6.6 branch

solverat:
Damn you‘re right. Going to rebase it tomorrow!

solverat:
done

weisswurstkanone:
thanks a lot!

PR #6717, opened 2020-07-01
[Data Objects] CSV Import - New Objects Based On Key

### LINKED ISSUE #5303: [Data Objects] CSV Import - New Objects Based On Key
# Bug Report

### Expected behavior
Trying to create new data objects by CSV import. I have a CSV file in which the "keys/filenames" are included. 
https://demo.pimcore.fun/admin/login/deeplink?object_763_folder
[csv-import-test-key.zip](https://github.com/pimcore/pimcore/files/3873955/csv-import-test-key.zip)

image

### Actual behavior
Data objects do not get named as configured in import. New objects are created even though there are existing ones with the same key in the CSV.
image

weisswurstkanone:
resolves #5303

### DIFF
--- a/lib/DataObject/Import/Resolver/Filename.php
+++ b/lib/DataObject/Import/Resolver/Filename.php
@@ -17,7 +17,6 @@
 
 namespace Pimcore\DataObject\Import\Resolver;
 
-use Pimcore\File;
 use Pimcore\Model\Element\Service;
 use const FILTER_VALIDATE_BOOLEAN;
 use Pimcore\Model\DataObject;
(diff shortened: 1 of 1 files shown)


lib/DataObject/Import/Resolver/Filename.php
```diff
@@ -56,21 +58,33 @@ public function resolve(\stdClass $config, int $parentId, array $rowData)
...
+        if ($overwrite) {
+            $objectKey = $rowData[$this->getIdColumn($config)];
+            $intendedPath = $parent->getRealFullPath() . '/' . $objectKey;
+        } else {
+            $objectKey = $rowData[$this->getIdColumn($config)];
+            $objectKey = Service::getValidKey($objectKey, "object");
+            $intendedPath = $parent->getRealFullPath() . '/' . $objectKey;
+
+            if (AbstractObject::getByPath($intendedPath)) {
+                $objectKey = $prefix;
```
dvesh3, line 76: prefix is ignored for new objects
image
and when object exists(without overwrite check) then it creates new numbered objects with prefix.
image

weisswurstkanone: this numbering should happen if overwrite is not checked and the key already exists

dvesh3: I expect it to take key not number when filename resolver strategy is used.

PR #6720, opened 2020-07-01
[GridHelperService] Fix Notice: Undefined index: value

blankse:
1. Open Asset Folder
2. Filter file name with no input
3. Get error:
image


weisswurstkanone:
Just curious how exactly this can be reproduced. Was not able to reproduce this by enabling the filter checkbox. But doesn't hurt anyway :-)

thx!

blankse:
@weisswurstkanone You must also set PIMCORE_PHP_ERROR_REPORTING to E_ALL

weisswurstkanone:
I know, it is turned on :-)

blankse:
Strange... :thinking:

PR #6721, opened 2020-07-01
[Search] Allow duplicate words in search_backend_data

BlackbitNeueMedien:
Currently duplicate words are removed in `search_backend_data` datasets. But as search results are ordered by relevance (by default) an object with multiple occurences of the search term should be ranked higher than an object where the search term only occurs once.

### DIFF
--- a/models/Search/Backend/Data.php
+++ b/models/Search/Backend/Data.php
@@ -507,11 +507,6 @@ class Data extends \Pimcore\Model\AbstractModel
         $data = str_replace("\t", '', $data);
         $data = preg_replace('#[ ]+#', ' ', $data);
 
-        // deduplication
-        $arr = explode(' ', $data);
-        $arr = array_unique($arr);
-        $data = implode(' ', $arr);
-
         return $data;
     }
 
--- a/models/Search/Backend/Data.php
+++ b/models/Search/Backend/Data.php
@@ -29,6 +29,9 @@ use Pimcore\Model\Search\Backend\Data\Dao;
  */
 class Data extends \Pimcore\Model\AbstractModel
 {
+    // if a word occures more often than this number it will get stripped to keep the search_backend_data table from getting too big
+    const MAX_WORD_OCCURENCES = 3;
+
     /**
      * @var Data\Id
      */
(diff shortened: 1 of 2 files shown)


brusch:
The main idea behind this deduplication was to reduce the size of the full-text index, which can get huge very quickly. But this is definitely a valid point, but we should still be aware of the index size. What do you think about the following approach, we remove all: 
- occurrences more than 3
- words with less than 3 characters

WDYT?

brusch:
@BlackbitNeueMedien ping 🏓😉

BlackbitNeueMedien:
I would alter your proposal a bit:
* Remove words with less than `innodb_ft_min_token_size` (see https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_ft_min_token_size) and more than `innodb_ft_max_token_size`
* Remove occurences above 3

I also thought about using something like https://github.com/voku/stop-words to filter out some stopwords which are not necessary for searching anyway. 
We could remove stopwords in https://github.com/pimcore/pimcore/blob/master/models/DataObject/ClassDefinition/Data/Localizedfields.php#L399-L421 for localized fields and for all other data we could use the set default language to filter out stopwords.
WDYT?

brusch:
Sounds good, can you update your PR accordingly? 
Regarding the stopwords, this also sounds good to me, but I think this should be definitely a topic for another PR 😊

models/Search/Backend/Data.php
```diff
@@ -507,6 +510,28 @@ protected function cleanupData($data)
         $data = str_replace("\t", '', $data);
         $data = preg_replace('#[ ]+#', ' ', $data);
 
+        $minWordLength = $this->getDao()->getMinWordLengthForFulltextIndex();
+        $maxWordLength = $this->getDao()->getMaxWordLengthForFulltextIndex();
+
+        $words = explode(' ', $data);
+
+        $wordOccurrences = [];
+        foreach($words as $key => &$word) {
```
jdreesen, line 519: Since you don't write `$word` in the loop, there's no need in assigning it as a reference.

BlackbitNeueMedien: Is only to not create a copy of the array - saves memory.

jdreesen: `foreach` loops in PHP are copy-on-write ;)

BlackbitNeueMedien: Thanks, you are right ;-)
```
It re-uses the existing value in memory. With zero performance cost. It's called "copy-on-write" and means that PHP doesn't make any copies unless you try to MODIFY the value.
```
https://www.php.net/manual/de/control-structures.foreach.php#120710

brusch:
Good work - thanks a lot! 🙏

PR #6724, opened 2020-07-01
Fix ecommerce filter to keep original ordering

Yivan:
Hello,

We are using ecommerce ecommerce filters and current implementation break desired ordering forced in productListing, even if no order is configured in the filterDefinition.

This PR provide a fix.

Please check the comments for the dump command which explains the problem:

```
$productListing = $factory->getIndexService()->getProductListForCurrentTenant();

// Set order
$productOrderIds = MyProduct::getOrderIds();
$productListing->setOrderKey('FIELD(a.o_id, ' . $productOrderIds . ')');
$productListing->setOrder('ASC');

// dump($productListing->getProducts()); // Order is ok, as desired

// We create and init filter service
\Pimcore\Bundle\EcommerceFrameworkBundle\FilterService\Helper::setupProductList($filterDefinition, $productListing, $params, $viewModel, $filterService, true);

// dump($productListing->getProducts()); // Order is reset ! Even if the filter definition no order is given
```

The filter definition has no order setted, so it should keep the setOrderKey  of the productListing:

image

Thanks.

PR #6731, opened 2020-07-02
Fix ecommerce filter to keep order

Yivan:
Hello,

It replace https://github.com/pimcore/pimcore/pull/6724
This new one PR is based on 6.6 branch and for 6.6 branch so.

Thanks


fashxp:
thx!

PR #6738, opened 2020-07-06
Event to modify exported data

ckogler-elements:
Add pimcore.dataobject.postCsvItemExport event to modify exported CSV/XSLx data


lib/Event/DataObjectEvents.php
```diff
@@ -112,4 +112,11 @@ final class DataObjectEvents
      * @var string
      */
     const POST_COPY = 'pimcore.dataobject.postCopy';
+
+    /**
+     * @Event("Pimcore\Event\Model\DataObjectEvent")
+     *
```
weisswurstkanone, line 118: arguments should be listed here

ckogler-elements: arguments are added

PR #6745, opened 2020-07-07
[Data objects] Generate default values on update

BlackbitNeueMedien:
Currently default values are only applied on object creation. This makes default value generators completely useless for manually created objects because in the moment of creation all fields are empty. So when you want to create a value based on other fields - e.g. a URL slug based on a product's name - this becomes impossible because on object creation the product has no name and on first saving it is an update operation and thus default values are not applied.
This PR changes this behaviour so default values are also applied for updates.

Of course then there is the problem that you will not be able to nullify this field (except if your default value generator supports a special value for this). But this is also the case for inheritance (it is not possible to override an inherited field with an empty string / array / null or whatever. So why would one want to set a field with a default value to `null`? And why do we need to allow it for default values but not for inheritance? The only clean solution I can think of would be a "nullify-button" - be it as a button beside the field or via context menu on right-clicking the field. Any other suggestions?

### DIFF
--- a/models/DataObject/Traits/DefaultValueTrait.php
+++ b/models/DataObject/Traits/DefaultValueTrait.php
@@ -47,7 +47,6 @@ trait DefaultValueTrait
     protected function handleDefaultValue($data, $object = null, $params = [])
     {
         $context = isset($params['context']) ? $params['context'] : [];
-        $isUpdate = isset($params['isUpdate']) ? $params['isUpdate'] : true;
 
         /**
          * if inheritance is enabled and there is no parent value then take the default value.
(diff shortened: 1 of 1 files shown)


fashxp:
The decision to apply default values only while create was made on purpose. We see them as default values, not as fallback or calculated values and therefore they should not be applied on every update. 
If you want to implement the use case of setting/updating values on every update (or calculate them based on other field values), you always can (and should) use event listeners. 
That's also the reason why we removed default value generator from url slug (because it didn't make any sense). 

And since we don't see default values as fallback values, not being able to null them out is a major aspect to us - also in contrast to inheritance.
Using a nullify button is not an option to us either - what happens next time the data object is saved (and nullify-button is not pressed), or how can that be archived via API? 

Feel free to continue discussion. Maybe we find some arguments we didn't considered yet and an even better solution. For now we won't merge.

BlackbitNeueMedien:
Ok, thanks for the hint @fashxp . Then we will change our bundle to use an event listener.

PR #6749, opened 2020-07-08
[DataObject] Performance enhancements when rebuilding classes

aclDeveloper:
When you change class definitions and execute the `pimcore:deployment:classes-rebuild` command, the underlying data model is altered if `skipColumn` evaluates to **false**

We experienced heavy workloads and poor performance on systems with lots of columns (100+) and/or records (200k+). Investigation lead to the fact that columns are altered even though they don't have changed. This is based on the fact that MySQL syntax supports multiple ways of defining data types and treats `decimal(10)` the same as it does `Decimal(10)`, etc. 

We therefore propose to **normalize** the type comparison in `skipColumn` in order to skip columns that have not changed in type during class rebuilding.


brusch:
Yep, makes sense, thanks a lot!

PR #6752, opened 2020-07-08
[Data objects] Objects of wrong folder displayed

### LINKED ISSUE #6751: [Data objects] Objects of wrong folder displayed
Steps to reproduce:
In Pimcore backend:
1. Create folder `_0` under `/`
2. Create folder `00` under `/`
3. Create object `A` in `/_0`
4. Create object `B` in `/00`
5. Open folder `_0` --> you will see both objects in the grid although actually only `A` is in folder `/_0` while `B` is not.

weisswurstkanone:
resolves #6751

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -1666,7 +1666,9 @@ class AssetController extends ElementControllerBase implements EventedController
 
         $conditionFilters = [];
         $list = new Asset\Listing();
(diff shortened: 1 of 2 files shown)


brusch:
@weisswurstkanone  I'm afraid that we have this / or similar issues anywhere where we're doing a `path LIKE` right? 
So maybe we should put this into a helper function [here](https://github.com/pimcore/pimcore/blob/master/lib/Db/PimcoreExtensionsTrait.php) and also fix all the other places.

weisswurstkanone:
I agree, I am afraid this will be a bigger change

weisswurstkanone:
closed in favor of https://github.com/pimcore/pimcore/pull/6809

PR #6753, opened 2020-07-08
Bug fix: "Use current player position as preview" Button not working

joFra210:
"Use current player position as preview" button now is working again

  
  

Rearranged small block in video.js (line 202)

## Additional info  
Bug can be reproduced on https://demo.pimcore.fun/
### Expected behavior
Video-thumbnail should be set when clicking on "Use current player position as preview" button.
### Actual behavior
Nothing happens, but the browser console logs error: 
`TypeError: Cannot read property 'getTime' of undefined`
### Steps to reproduce
Upload Video, try to set thumbnail


dvesh3:
thanks a lot!

PR #6755, opened 2020-07-08
Fix: Pimcore can't load object bricks with localized block fields

blankse:
The data could be saved, but if you want load it you get an SQL error.


weisswurstkanone:
thanks a lot!

PR #6756, opened 2020-07-08
Allow usage of `Pimcore\Navigation\Builder` without request

jdreesen:
I want to use the navigation builder inside a console command but it fails because obviously there's no request available which results in a `There is no master request available.` exception.

The builder actually doesn't need a request if the first parameter (`$activeDocument`) is not `null` because it uses that document to determine which pages are active if it could not be determined from the request.


dvesh3:
thanks a lot!

PR #6757, opened 2020-07-08
Additional Icons

passioneight:
We just added the white icons in black and updated the icon library to display them as well.


brusch:
Thanks a lot for your PR. However I don't think we need the black icons in the core, so maybe this is better maintained in a bundle?

PR #6758, opened 2020-07-08
Garbage collection - only close db connection if there is no active transaction

ckogler-elements:
When calling \Pimcore::collectGarbage(); the Db connection should only be closed if no active transaction is running.

Currently you can't call \Pimcore::collectGarbage(); in a "preDelete" or "postDelete" Event because it closes the connection and a error occurs... because the transaction is not available anymore...


dpfaffenbauer:
What is the use case for "Pimcore Garbarge Collection" in preDelete?

ckogler-elements:
> What is the use case for "Pimcore Garbarge Collection" in preDelete?

I have a exporter wich writes data to a csv... and the client can configure if the data should be written directly when the object is saved or in the background. On the preDelete event i have to delete the data from the tareget file. Therefore it can happen that lots of data have to be loaded/processed where i have to call the garabage collector. 
Normaly the client shouldn't configure the export like this - but he can...

Nevertheless i wouldn't expect that the garbage collector closes the DB connection if the db connection is still required - which is the case when i'm in a transaction...

dpfaffenbauer:
Alright, makes sense. That is what I would call: house-made problems ;)

PR #6759, opened 2020-07-08
[DOCS] no add function on $this->view

putzflorian:
there is no add function available on $this->view

### DIFF
--- a/doc/Development_Documentation/02_MVC/02_Template/02_Templating_Helpers/README.md
+++ b/doc/Development_Documentation/02_MVC/02_Template/02_Templating_Helpers/README.md
@@ -301,11 +301,17 @@ index.html.twig
 IndexController.php (whatever controller / method is designated for /some/other/document in the document tree)
 ```php
 public function otherDocumentAction(Request $request) {
-    $this->view->add([
-        'parameterToPass' => $request->get('parameterToPass')
-    ]);
+    $this->view->parameterToPass = $request->get('parameterToPass');
 }
 ```
+
+more Convenient way
+```php
+public function otherDocumentAction(Request $request) {
+    return $this->render(":Default:someOtherDocument.html.twig", ['parameterToPass' => $request->get('parameterToPass')]);
+}
+```
+
 
 someOtherDocument.html.twig (whatever Twig template is actually for /some/other/document in the document tree)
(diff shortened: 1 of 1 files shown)


doc/Development_Documentation/02_MVC/02_Template/02_Templating_Helpers/README.md
```diff
@@ -301,12 +301,18 @@ index.html.twig
 IndexController.php (whatever controller / method is designated for /some/other/document in the document tree)
 ```php
 public function otherDocumentAction(Request $request) {
-    $this->view->add([
-        'parameterToPass' => $request->get('parameterToPass')
-    ]);
+    $this->view->parameterToPass = $request->get('parameterToPass');
```
fashxp, line 304: isn't also following possible (if yes, we should use it instead of `$this->view`
```
return ['parameterToPass' => $request->get('parameterToPass')]; 
```

dpfaffenbauer: honestly: should also not be used. Auto View Rendering is a thing Pimcore added to be more backward compatible to zend. Those times are long gone now.... A bit down is the right way of doing it:

```
return $this->render(":Default:someOtherDocument.html.twig", ['parameterToPass' => $request->get('parameterToPass')]);
```

blankse: We use this function to add parameters to the view: `$this->view->getParameters()->add()`

putzflorian: i would agree with @dpfaffenbauer comment.
better to remove this line `$this->view->parameterToPass = $request->get('parameterToPass');`

putzflorian: as @fashxp explained `return ['parameterToPass' => $request->get('parameterToPass')]; ` will also work

brusch: @putzflorian can you please update the PR accordingly? Thanks a lot!

PR #6762, opened 2020-07-10
Pimcore | Bootstrap | Initialize .env variables before constants init

andreas-gruenwald:
When implementing app/constants.php access to environment variables may be required, for instance when setting up a connection to S3 based on a docker environment.

This PR recommends to initialize the .env variables before the initialization of the constants take place.


brusch:
Thanks a lot!

PR #6763, opened 2020-07-10
[Documents] Copy master request context to sub-request

### LINKED ISSUE #6650: Unpublished objects showing on 'Error' page
When there is a page showing e.g. news DataObjects and this page is also set as the default Error page in the System Settings it shows the unpublished News items (DataObjects) when there is displayed through an error. 
In case of a 'normal' visit it behaves as expected and unpublished items are not displayed.

I found out that in Pimcore\Bundle\CoreBundle\EventListener\PimcoreContextListener in the onKernelRequest in case of an error the check on `$event->isMasterRequest()` is returning false and so the `initializeContext()` is not called.
In the `initializeContext()` function the value for hideUnpublised DataObjects is set to true as the default value is to false.

My first reaction would be to set the default value to true, but it is hard to oversee all the implications and perhaps changed behaviour. Or what was the design decision to have the default value false there? Should it be set to some value in any other case?

dvesh3:
Resolves #6650

### DIFF
--- a/bundles/CoreBundle/EventListener/PimcoreContextListener.php
+++ b/bundles/CoreBundle/EventListener/PimcoreContextListener.php
@@ -80,9 +80,16 @@ class PimcoreContextListener implements EventSubscriberInterface, LoggerAwareInt
                     'path' => $request->getPathInfo()
                 ]);
             }
-
-            $this->initializeContext($context, $request);
+        } else {
+            // copy master pimcore context to sub-request if available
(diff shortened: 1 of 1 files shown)


brusch:
@dvesh3 this needs some more attention. Actually we don't want to initialize the context for every single sub-request, so we need to check why the context is not initialized correctly in case the error page get's displayed.

brusch:
Closed in favor of #6811

PR #6769, opened 2020-07-13
[Data Objects] Fix API-saving unpublished with empty mandatory field throws exception

### LINKED ISSUE #6587: [Data Objects] API-saving unpublished with empty mandatory field throws exception
### Expected behavior
Saving unpublished data object via API with unset mandatory field should not throw ValidationException (just like when saving via the UI)

### Actual behavior
Pimcore\Model\Element\ValidationException is thrown if there is an empty mandatory field, even if data object is unpublished.

dvesh3:
Resolves #6587

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -76,7 +76,7 @@ class Concrete extends AbstractObject implements LazyLoadedFieldsInterface
     protected $scheduledTasks = null;
 
     /**
-     * @var bool
+     * @var bool|null
      */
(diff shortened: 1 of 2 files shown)


models/DataObject/Concrete.php
```diff
@@ -78,7 +78,7 @@ class Concrete extends AbstractObject implements LazyLoadedFieldsInterface
     /**
      * @var bool
```
blankse, line 79:
```suggestion
     * @var bool|null
```

Issue #6696, opened 2020-06-29
Accessing Pimcore Assets via WebDAV not working


woellundwoell:
# Bug Report

### Steps to reproduce
* Open https://demo.pimcore.fun/admin/asset/webdav in webbrowser (using demo credentials)
* check var/logs/prod.log 
`[2020-06-29 16:54:14] pimcore.ERROR: Symfony\Component\Routing\Exception\MissingMandatoryParametersException: Some mandatory parameters are missing ("path") to generate a URL for route "pimcore_admin_webdav"`

### Possible reason
Missing default or emtpy value for parameter "path" in 
https://github.com/pimcore/pimcore/blob/c4e091969e509f33b3b92b5ed573c1ecc2ba616d/bundles/AdminBundle/Controller/Admin/AssetController.php#L901

### Possible solution
Add a default for $path
`public function webdavAction(Request $request, string $path = '')`

Is this solution feasible? If so we could provide a PR

brusch:
Fixed by #6777

PR #6777, opened 2020-07-14
Bugfix Webdav Link Generating

Gerhard-Kanzler:
It is not possible to generate the link without a path.
I add the '/' as BaseUri to generate the rigth URI


brusch:
Fixes #6696

brusch:
Thanks a lot! 🙏

PR #6784, opened 2020-07-14
Add additional Twig functions for Site model

kjkooistra-youwe:
Needed to retrieve the current site in a Twig template not related to a document and figured it would be useful to add this to the core.


brusch:
Thanks a lot! 🙏

PR #6787, opened 2020-07-14
Cope with missing target tags in xliff import file

### LINKED ISSUE #6785: XLIFF Import fails when target tag is missing
# When exporting not all values get target tags in XML
No target tag:
missing_target
Crashes here:
crash_without_target
Can be fixed with a simple check for an existing target tag, I added isset check to see if it exists or not.

So should it be fixed during import or export?

msphn:
Resolves #6785


msphn:
@weisswurstkanone

weisswurstkanone:
thx!

Issue #6530, opened 2020-06-04
Pagination disappears when saving an object


BlackbitNeueMedien:
I have an object that has more than 30 direct children. I get a pagination in the tree. 

<img width="907" alt="Bildschirmfoto 2020-06-04 um 16 54 51" src="https://user-images.githubusercontent.com/8749138/83773199-87a73e80-a684-11ea-80d5-8419c4b97716.png">

Now I open the parent element and at the same time I open the list of child elements via the plus icon in the tree. When I save the parent element, the pagination of the open parent element disappears in the tree.

<img width="915" alt="Bildschirmfoto 2020-06-04 um 16 54 59" src="https://user-images.githubusercontent.com/8749138/83773197-85dd7b00-a684-11ea-84eb-5bc49827002c.png">

----
General Information:
Detected in Pimcore Version: v6.5.3 (also in latest release, see demo.pimcore.fun)

weisswurstkanone:
Unable to reproduce. Can you share a short video demonstrating the problem please?

BlackbitNeueMedien:
I have attached a zip file with a short one minute log video which demonstrate the issue. 

[Bildschirmvideo aufnehmen 2020-06-09 um 13.28.17-720p.mov.zip](https://github.com/pimcore/pimcore/files/4751867/Bildschirmvideo.aufnehmen.2020-06-09.um.13.28.17-720p.mov.zip)

weisswurstkanone:
Thx, seems to be a nasty side effect of this line:

https://github.com/pimcore/pimcore/blob/a9f787995202fa4729ab26405c940464ae7f35f4/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js#L229

brusch:
@BlackbitNeueMedien can you confirm that?

brusch:
Fixed by #6577

patrickbitzer:
> @BlackbitNeueMedien please check if #6577 solves your problem & if you notice any side effects

Sorry for this late response. But at the moment we have many work :D.

This fix does not has any effects. I build in this changes in our clients code, but the problem i described has already been there. Any other solutions?

patrickbitzer:
Sorry. The adjustment does work. I got the line wrong. It's all fixed and working fine now.
Thanks again. I delete the Pull Request again ;-)

PR #6791, opened 2020-07-15
Addition to PullRequest #6577 - Treepaging disappears after save object

BlackbitNeueMedien:
Hello, I took another look inside and saw that the "record.ptb" is destroyed and deleted in the event "beforeitemupdate". I added the exception that this only happens when no paging is needed.

Can you check again if I am on the right way? But this adjustment works fine ;-)


weisswurstkanone:
Doesn't hurt but I don't see why there should be a record.ptb if record.needsPaging **was** false

patrickbitzer:
> Doesn't hurt but I don't see why there should be a record.ptb if record.needsPaging **was** false

Sorry. The adjustment form #6530  does work. I got the line wrong. It's all fixed and working fine now.
Thanks again. I delete the Pull Request again ;-)

PR #6793, opened 2020-07-15
[Docs] Wrong document resolving (BC break in 6.6.9)

BlackbitNeueMedien:
I wondered why in one project Pimcore documents did not resolve correctly after updating to 6.6.9. The reason was that the controller overrode `Pimcore\Controller\setContainer()` and set some default variables via `$this->view->variableName = 123;`. The problem with this is that this sets the view model in the request via https://github.com/pimcore/pimcore/blob/24a5ff9eca55b0a14b310ccaee3213df76fb4fad/lib/Controller/FrontendController.php#L47 -> https://github.com/pimcore/pimcore/blob/24a5ff9eca55b0a14b310ccaee3213df76fb4fad/lib/Http/Request/Resolver/ViewModelResolver.php#L56-L65 -> https://github.com/pimcore/pimcore/blob/24a5ff9eca55b0a14b310ccaee3213df76fb4fad/lib/Http/Request/Resolver/ViewModelResolver.php#L97

In https://github.com/pimcore/pimcore/commit/7a99c74104749e761c2bd2ffbdf84c03bc902964#diff-cafe6135bd57517b8d11e7e4edf3cb0cR93 document resolving was moved to `kernel.controller` event. The found document gets written to `Request::attributes[DynamicRouter::CONTENT_KEY]` in https://github.com/pimcore/pimcore/blob/24a5ff9eca55b0a14b310ccaee3213df76fb4fad/bundles/CoreBundle/EventListener/Frontend/ElementListener.php#L136 and https://github.com/pimcore/pimcore/blob/24a5ff9eca55b0a14b310ccaee3213df76fb4fad/lib/Http/Request/Resolver/DocumentResolver.php#L47-L53 
But the Request's view model does not get updated here. Beside documenting this BC break (what this PR does), I wonder if we should update the view model's document here.

When you then later on call `$this->view->document` the previously generated, cached view model gets returned:
https://github.com/pimcore/pimcore/blob/24a5ff9eca55b0a14b310ccaee3213df76fb4fad/lib/Http/Request/Resolver/ViewModelResolver.php#L56-L58
This might contain the wrong document - it will for example when in edit mode and the document gets loaded from session. 
https://github.com/pimcore/pimcore/blob/24a5ff9eca55b0a14b310ccaee3213df76fb4fad/models/Document/Tag/Area/Info.php#L220

This prevents the user from editing any area brick in the documents because as soon as you add or remove a brick or just change its data, the document gets saved to session and then reloaded. But as the view model gets fixed too early due to our access to `$this->view` in the overridden `FrontendController::setContainer()`, the wrong document gets used to fetch the elements to be rendered (the document from the database gets used, not the one from the session - so you can never change anything which automatically saves the document to session).

### DIFF
--- a/models/Webservice/Data.php
+++ b/models/Webservice/Data.php
@@ -129,12 +129,6 @@ abstract class Data
      */
     public function reverseMap($object, $disableMappingExceptions = false, $idMapper = null)
     {
-        $fallbackProperties = [];
-        $existingObject = Element\Service::getElementByPath('document', $this->path.$this->key);
-        if($existingObject) {
-            $fallbackProperties = $existingObject->getProperties();
-        }
-
         $keys = get_object_vars($this);
         foreach ($keys as $key => $value) {
             $method = 'set' . ucfirst($key);
@@ -165,10 +159,7 @@ abstract class Data
                     }
 
                     if ($id) {
-                        try {
                             $dat = Element\Service::getElementById($type, $id);
-                        } catch(\Exception $e) {
-                        }
                     }
 
                     if (is_numeric($propertyWs['data']) and !$dat) {
@@ -176,12 +167,6 @@ abstract class Data
                             throw new \Exception('cannot import property [ ' . $type . ' ] because it references unknown ' . $propertyWs['data']);
                         } else {
                             $idMapper->recordMappingFailure('object', $object->getId(), $type, $propertyWs['data']);
-                        }
-
-                        foreach($fallbackProperties as $fallbackProperty) {
-                            if($fallbackProperty->getName() == $propertyWs['name'] && $fallbackProperty->getType() == $propertyWs['type']) {
-                                $dat = $fallbackProperty->getData();
-                            }
                         }
                     }
                 } elseif ($type == 'date') {
--- a/models/Webservice/Data.php
+++ b/models/Webservice/Data.php
@@ -129,6 +129,12 @@ abstract class Data
      */
     public function reverseMap($object, $disableMappingExceptions = false, $idMapper = null)
     {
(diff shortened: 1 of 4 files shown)


models/Webservice/Data.php
```diff
@@ -129,6 +129,12 @@ private function mapProperties($value)
      */
     public function reverseMap($object, $disableMappingExceptions = false, $idMapper = null)
     {
+        $fallbackProperties = [];
```
brusch, line 132: How is this related to this PR?

brusch:
Thanks for sharing your findings! I guess it would be best to update the view model as well, if it exists already on the request at this time. Then this shouldn't cause any further isses and I think then this PR is also obsolete, right?

models/Webservice/Data.php
BlackbitNeueMedien: Sorry, it isn't. Reverted those

BlackbitNeueMedien:
Yes, if the view model got updated in https://github.com/pimcore/pimcore/blob/24a5ff9eca55b0a14b310ccaee3213df76fb4fad/lib/Http/Request/Resolver/DocumentResolver.php#L47-L53 then this PR is obsolete. Would you do that or shall I?

brusch:
@BlackbitNeueMedien I'm on it ;)

BlackbitNeueMedien:
Could you please also apply the changes in https://github.com/pimcore/pimcore/pull/6793/files#diff-d07902447e77f7587968dd55d8b26e27R353 and https://github.com/pimcore/pimcore/pull/6793/files#diff-437a631008c6eaefb2c98052941de7adR114 ?

PR #6794, opened 2020-07-16
ViewModel: "document" parameter doesn't get set/updated when view model was initialized before document was resolved 

brusch:
Follow up of  #6793

Fixed regression of #6203 and #5905, regarding the document parameter on the view model, which isn't set/updated when view model was created before the document was resolved by the event listener (onKernelController)


brusch:
@BlackbitNeueMedien can you confirm that this fixes your issue described in #6793 ? Thanks a lot!

BlackbitNeueMedien:
Works like a charm! Thanks!

brusch:
Thanks for the feedback!

blankse:
@brusch Is only merged to master?

brusch:
@blankse oh, you're right, thanks a lot for the hint.

brusch:
Cherry picked into `6.6` - thanks again!

PR #6799, opened 2020-07-16
[AdminBundle] Change core templates to Twig

dvesh3:
Resolves Partially #4680


lib/Twig/Extension/HelpersExtension.php
```diff
@@ -18,13 +18,38 @@
...
+    {
+        return [
+            new TwigFilter('basename', [$this, 'basenameFilter']),
+        ];
+    }
+
+    public function getFunctions()
+    {
+        return [
+            new TwigFunction('callStatic', function ($class, $method, $args = array()) {
```
dpfaffenbauer, line 40: not a good idea I think. This seems to a bad practice. Better create specific twig extensions where this is needed.

dpfaffenbauer: also, naming scheme is not persisted, use camel case and prefix with pimcore_

bundles/AdminBundle/Controller/Admin/IndexController.php
```diff
@@ -448,6 +450,8 @@ protected function addCustomViewSettings(ViewModel $settings)
 
     public function onKernelController(FilterControllerEvent $event)
     {
+        // enable view auto-rendering
+        $this->setViewAutoRender($event->getRequest(), true, 'twig');
```
dpfaffenbauer, line 454: if you already on that, why don't you return $this->renderTemplate or $this->render on every action instead? Auto View Rendering is not good imo.

brusch:
@dvesh3 this needs to be finished in order to start working on #7224

bundles/AdminBundle/Controller/Admin/LoginController.php
```diff
@@ -286,4 +288,29 @@ public function twoFactorAuthenticationAction(Request $request, BruteforceProtec
     public function twoFactorAuthenticationVerifyAction(Request $request)
     {
     }
+
+    /**
+     * @return bool
+     */
+    public function detectBrowser()
+    {
+        $supported      = false;
```
dpfaffenbauer, line 297: create a twig extension that does that?

bundles/AdminBundle/Resources/views/SearchAdmin/Search/Quicksearch/asset.html.twig
```diff
@@ -0,0 +1,22 @@
+{# @var $element \Pimcore\Model\Asset\Image|\Pimcore\Model\Asset\Document|\Pimcore\Model\Asset\Video #}
+{% set previewImage = null %}
+{% set params = { 'id': element.id, 'treepreview': true, 'hdpi': true } %}
+
+{% if element is instanceof('\\Pimcore\\Model\\Asset\\Image') %}
+    {% set previewImage = path('pimcore_admin_asset_getimagethumbnail', params) %}
+{% elseif element is instanceof('\\Pimcore\\Model\\Asset\\Video') and callStatic('\\Pimcore\\Video', 'isAvailable') %}
```
dpfaffenbauer, line 7: `callStatic` is not good DX, better create twig extensions:

`pimcore_video_is_available`
`pimcore_document_is_available`

bundles/AdminBundle/Resources/views/SearchAdmin/Search/Quicksearch/info-table.html.twig
```diff
@@ -0,0 +1,77 @@
...
+                <th>{{ 'mimetype'|trans([],'admin') }}</th>
+                <td>{{ element.getMimetype() }}</td>
+            </tr>
+        {% endif %}
+
+        {% if language is not empty %}
+            <tr>
+                <th>{{ 'language'|trans([],'admin') }}</th>
+                <td style="padding-left: 40px; background: url({{ pimcore_language_flag(language, false) }}) left top no-repeat; background-size: 31px 21px;">
+                    {% set locales = callStatic('\\Pimcore\\Tool', 'getSupportedLocales') %}
```
dpfaffenbauer, line 23: `callStatic` is not good DX, better create twig extensions:

`pimcore_supported_locales`

bundles/AdminBundle/Controller/Admin/LoginController.php
dvesh3: quite specific for login request...what is the benefit?

dpfaffenbauer: I think it's more elegant...

brusch: I think it's fine as it is ...

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
```diff
@@ -484,10 +484,13 @@ public function quicksearchAction(Request $request, EventDispatcherInterface $ev
                     $data['iconCls'] .= ' pimcore_icon_' . $element->getType();
                 }
 
+                $validLanguages = \Pimcore\Tool::getValidLanguages();
```
dpfaffenbauer, line 487: twig extension here might be a good idea as well:`pimcore_valid_languages`

bundles/AdminBundle/Controller/Admin/IndexController.php
```diff
@@ -247,6 +247,7 @@ protected function buildPimcoreSettings(Request $request, array &$templateParams
             'document_tree_paging_limit' => $config['documents']['tree_paging_limit'],
             'object_tree_paging_limit' => $config['objects']['tree_paging_limit'],
             'maxmind_geoip_installed' => (bool) $this->getParameter('pimcore.geoip.db_file'),
+            'hostname' => htmlentities(\Pimcore\Tool::getHostname(), ENT_QUOTES, 'UTF-8'),
```
dpfaffenbauer, line 250: maybe here too? Not sure though if it really helps

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
brusch: Don't think this could be of general use, ... I'd just leave it for now as it is.

bundles/AdminBundle/Controller/Admin/IndexController.php
brusch: I'd assume Twig's `escape` would be also sufficient here, right?

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
dpfaffenbauer: I think it could, I've had use case for it.

bundles/AdminBundle/Controller/Admin/IndexController.php
dpfaffenbauer: also the call to getHostname might be useful in some cases.

bundles/AdminBundle/Controller/Admin/LoginController.php
```diff
@@ -286,4 +288,29 @@ public function twoFactorAuthenticationAction(Request $request, BruteforceProtec
...
+        $browser        = new \Pimcore\Browser();
```
brusch, line 298: @dvesh3 Please consider #7084 here, so basically just use `\Browser` here 😉

PR #6803, opened 2020-07-17
AbstractListing order attribute is always an array

blankse:
Only the setter can have a string as parameter. But the attribute and the getter is always an array.


brusch:
Thanks a lot!

PR #6804, opened 2020-07-20
Implementation of the Hobex Payment Provider 

andreas-gruenwald:
This PR implements a payment provider for "Hobex".


dpfaffenbauer:
Is it really necessary to implement all Payment Providers into the Core? Shouldn't that be just a separate bundle for Pimcore you can install? Just think about bug-fixes and security related issues, you can handle that way faster outside the core....

andreas-gruenwald:
> Is it really necessary to implement all Payment Providers into the Core? Shouldn't that be just a separate bundle for Pimcore you can install? Just think about bug-fixes and security related issues, you can handle that way faster outside the core....

I agree. But then all payment providers need to be shifted at once. See #6805.

brusch:
Yep, we'll do that for v7 😉

bundles/EcommerceFrameworkBundle/PaymentManager/V7/Payment/StartPaymentResponse/HobexResponse.php
```diff
@@ -0,0 +1,85 @@
...
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Bundle\EcommerceFrameworkBundle\PaymentManager\V7\Payment\StartPaymentResponse;
+
+class HobexResponse implements StartPaymentResponseInterface
```
fashxp, line 17: why not use [SnippetResponse](https://github.com/pimcore/pimcore/blob/68046984953e4a69dc6ff5cdf3cfb7a2e7e8af77/bundles/EcommerceFrameworkBundle/PaymentManager/V7/Payment/StartPaymentResponse/SnippetResponse.php#L41) here?

fashxp: at least as parent class, then rendering the form would be more unified.

andreas-gruenwald: You are right. I replaced the HobexResponse with SnippetResponse: https://github.com/pimcore/pimcore/pull/6804/commits/5653cb38da485bd8f7f66485600c2fa1f9f242d3

PR #6812, opened 2020-07-21
Hide all unpublished elements in document relation tags

jdreesen:
While `Pimcore\Model\Document\Tag\Relations` checked for both, `DataObject`s and `Document`s whether to hide unpublished ones, `Pimcore\Model\Document\Tag\Relation` only checked this for `Document`s.

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -126,7 +126,7 @@ class Service extends Model\AbstractModel
      *
      * @internal
      *
-     * @param DataObject|Document $element
+     * @param AbstractObject|Document $element
      *
      * @return string
      *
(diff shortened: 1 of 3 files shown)


models/Element/Service.php
```diff
@@ -267,6 +267,17 @@ public static function getDependedElement($config)
...
     }
 
+    /**
+     * @static
+     *
+     * @return bool
+     */
+    public static function doHideUnpublished($element)
+    {
+        return ($element instanceof DataObject && DataObject::doHideUnpublished())
```
weisswurstkanone, line 277: Maybe use AbstractObject instead ?

weisswurstkanone:
thanks a lot!

PR #6813, opened 2020-07-21
[Database] Convert search_backend_data.published to tinyint

BlackbitNeueMedien:
Same column definition as e.g. objects.o_published


weisswurstkanone:
you have to change it in install.sql as well

brusch:
Thanks a lot!

PR #6817, opened 2020-07-22
[Asset] Apply 8BIM embedded clipping path to thumbnails

### LINKED ISSUE #5505: [Asset] Thumbnails: Support for embedded clipping mask in JPEG files
Example files: 
[images.zip](https://github.com/pimcore/pimcore/files/4009504/images.zip)

brusch:
resolves #5505

Known issue: it seems that `-clip` doesnt work with ImageMagick 6.9.7-4 Q16 x86_64 20170114 (which is used in Debian 9)

### DIFF
--- a/lib/Image/Adapter/Imagick.php
+++ b/lib/Image/Adapter/Imagick.php
@@ -143,6 +143,18 @@ class Imagick extends Adapter
             if (!$this->isPreserveColor()) {
                 $this->setColorspaceToRGB();
             }
+
+            // check for the existence of an embedded clipping path (8BIM / Adobe profile meta data)
+            $identifyRaw = $i->identifyImage(true)['rawOutput'];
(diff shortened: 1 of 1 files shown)


adriancespedes:
@brusch I may be wrong, but I believe that for the next release, it’s important to maintain backward compatibility. Setting clip_auto_support to default to false (as it was in Pimcore 10) would help avoid unexpected issues during upgrades. The **main problem we encountered is that images are automatically cropped in Pimcore when a clipping path was created in Photoshop but not enabled**. This led to unexpected behavior during our upgrade from Pimcore 10 to 11, and we spent time troubleshooting it. Defaulting clip_auto_support to false would help ensure smoother upgrades for other users. 

It’s just my opinion, and I didn’t mean to bother you—the great job you’re doing is much appreciated. Thanks!

PR #6819, opened 2020-07-22
Wrap page/hardlink deletion into transaction

jdreesen:
[`Pimcore\Model\Document\Page::delete()`](https://github.com/pimcore/pimcore/blob/34f953311a296113636803b48880d87e2675bb1d/models/Document.php) and [`Pimcore\Model\Document\Hardlink::delete()`](https://github.com/pimcore/pimcore/blob/34f953311a296113636803b48880d87e2675bb1d/models/Document/Hardlink.php#L240-L252) delete items from the database but don't wrap this into a transaction. 

Their parent classes/methods [`Pimcore\Model\Document\PageSnippet::delete()`](https://github.com/pimcore/pimcore/blob/34f953311a296113636803b48880d87e2675bb1d/models/Document/PageSnippet.php#L184-L206) and [`Pimcore\Model\Document::delete()`](https://github.com/pimcore/pimcore/blob/34f953311a296113636803b48880d87e2675bb1d/models/Document.php#L789-L852) are wrapped into a transaction which is why I think these two should be, too.

Note: the diff should be viewed [with disabled whitespace changes](https://github.com/pimcore/pimcore/pull/6819/files?diff=unified&w=1).

### DIFF
--- a/models/Document/Hardlink.php
+++ b/models/Document/Hardlink.php
@@ -17,6 +17,9 @@
 
 namespace Pimcore\Model\Document;
 
+use Pimcore\Event\DocumentEvents;
+use Pimcore\Event\Model\DocumentEvent;
+use Pimcore\Logger;
 use Pimcore\Model;
 use Pimcore\Model\Document;
 use Pimcore\Model\Redirect;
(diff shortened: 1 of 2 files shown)


brusch:
Makes sense, thanks a lot!

PR #6826, opened 2020-07-27
Sitemap - Empty Hardlinks causing error

MosiiAUT:
If an empty hardlink is present, the DocumentTreeGenerator will throw an Exception "canBeAdded on null" / "hashChildren on null"


lib/Sitemap/Document/DocumentTreeGenerator.php
```diff
@@ -145,7 +145,7 @@ private function visit(Document $document, DocumentGeneratorContext $context): \
             $document = Document\Hardlink\Service::wrap($document);
```
dvesh3, line 145: since it is specific to hard links, it is better to check for document is empty here and return.

dvesh3:
thank you!

PR #6836, opened 2020-07-28
adjusts portrait images in the search preview to prevent overflow

PGBrandi:
Prevents portrait preview images in the search to overflow and keeps image within the container.
## Resolves
Overflow of image in the search preview

##Additional Info
Example Screenshots: 
Before:

After:


brusch:
Nice - thanks a lot! 🙌

PR #6839, opened 2020-07-28
 Depend on `StreamOutput` instead of `ConsoleOutput` in `Pimcore\Console\Dumper`

jdreesen:
I wanted to write a Test according to https://symfony.com/doc/4.4/console.html#testing-commands but it fails because Symfony injects a `Symfony\Component\Console\Output\StreamOutput` instance, while Pimcore expects a `Symfony\Component\Console\Output\ConsoleOutput`.

As `ConsoleOutput` is a subclass of `StreamOutput` and all methods that are used in the `Dumper` are actually defined in `StreamOutput` it should be safe to switch the dependencies.

### DIFF
--- a/lib/Console/Dumper.php
+++ b/lib/Console/Dumper.php
@@ -14,7 +14,7 @@
 
 namespace Pimcore\Console;
 
-use Symfony\Component\Console\Output\ConsoleOutput;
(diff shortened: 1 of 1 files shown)


brusch:
Thanks! 😊

PR #6843, opened 2020-07-28
[Version preview] Add whitespace between class name and element fullpath for many-to-many relation fields

BlackbitNeueMedien:
Currently when you look at the version preview of a many-to-many relation it looks something like 
```
Pimcore\Model\DataObject\Product/path/to/my/object
Pimcore\Model\DataObject\Product/path/to/another/object
```
This is a bit difficult to read as you have to exactly pay attention where there are backslashes and where forward slashes.
This PR adds a colon and a whitespace between the class name and the element's fullpath:
```
Pimcore\Model\DataObject\Product: /path/to/my/object
Pimcore\Model\DataObject\Product: /path/to/another/object
```


brusch:
Makes absolutely sense - thanks a lot! 👌

BlackbitNeueMedien:
@brusch What do you think of harmonizing the output like for advanced many-to-many-relations: see https://github.com/pimcore/pimcore/blob/a82c930b105a92cfbd5ab6f11eac6c48ca2de82b/models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php#L439
Imho nobody needs the FQN of the class here but the element type and fullpath would be enough. WDYT?

Issue #6825, opened 2020-07-27
[Cache] Tests are failing


dvesh3:
Cache tests are failing on master like: 
```cli
1) TaggableArrayAdapterProxyTest: Multiple tags
 Test  tests/Cache/Adapter/SymfonyProxy/TaggableArrayAdapterProxyTest.php:testMultipleTags
Failed asserting that true is false.

2) TaggableArrayAdapterProxyTest: Invalidate tag
 Test  tests/Cache/Adapter/SymfonyProxy/TaggableArrayAdapterProxyTest.php:testInvalidateTag
Item should be cleared when tag is invalidated
Failed asserting that true is false.
```
regression of Symfony new version: 4.4.11, 3.4.43

reported here: https://github.com/symfony/symfony/issues/37667

dvesh3:
closed by #6847

PR #6847, opened 2020-07-29
[Composer] Add conflicts for Symfony 3.4.43 & 4.4.11 due to regressions

dvesh3:
Resolves temporarily #6825

## Additional info  
Temporary fix needs to be reverted


brusch:
Don't think we need to revert it, we can just keep it, as only specific versions are blocked.

brusch:
Merged to `master`.

PR #6853, opened 2020-07-29
[Data object version preview] Equal version preview of many-to-many relations and advanced many-to-many relations

BlackbitNeueMedien:
Follow-up of #6843


brusch:
@BlackbitNeueMedien I think we should do the same also for `ManyToOneRelation` , can you update the PR? 
Thanks a lot! 😊

brusch:
Nice, thanks a lot! 👌

PR #6860, opened 2020-07-30
Remove unused `$isNested` parameter in `delete()` methods of `DataObject`'s and `Document`'s

jdreesen:
Not sure if this is desired or an oversight (actually, I don't understand the necessity of this parameter at all) but I've noticed that the method's argument is ignored when calling `parent` and instead the value is hardcoded to `true`.

Maybe someone of the core-developers can validate if this behavior is correct (in that cast I think it should be documented!) or if it should be patched by this PR.

### DIFF
--- a/models/DataObject/AbstractObject.php
+++ b/models/DataObject/AbstractObject.php
@@ -517,11 +517,9 @@ class AbstractObject extends Model\Element\AbstractElement
     }
 
     /**
-     * @param bool $isNested
-     *
      * @throws \Exception
      */
-    public function delete(bool $isNested = false)
+    public function delete()
     {
         \Pimcore::getEventDispatcher()->dispatch(DataObjectEvents::PRE_DELETE, new DataObjectEvent($this));
 
--- a/models/DataObject/AbstractObject.php
+++ b/models/DataObject/AbstractObject.php
@@ -530,7 +530,7 @@ class AbstractObject extends Model\Element\AbstractElement
             $children = $this->getChildren([self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER, self::OBJECT_TYPE_VARIANT], true);
             if (count($children) > 0) {
                 foreach ($children as $child) {
-                    $child->delete(true);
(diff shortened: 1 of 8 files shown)


dvesh3:
Sorry, even i'm not sure about this behavior of passing hard-coded value to `parent::delete`. But if we decide to go with your changes, then please make sure to also update all `$object->delete()` calls to `$object->delete(true)`. like this one:
https://github.com/pimcore/pimcore/blob/2d146e18936a1bbee0dd6f517397fd9f83cccb17/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L974

dpfaffenbauer:
Is $isNested even needed?

jdreesen:
> Is $isNested even needed?

I don't think so, at least for `DataObject`'s and `Documens`'s. The only place it's actually used is in `Asset`'s, afaik:
https://github.com/pimcore/pimcore/blob/15e1132729f5c6dd08198ec434828066c5f69fbb/models/Asset.php#L1066-L1072

fashxp:
maybe a leftover of https://github.com/pimcore/pimcore/commit/9279f30874d5e0c6cf3114c2c42bd72c9f2245d7

dpfaffenbauer:
Very like a leftover, can be removed then right? and maybe for Asset it could be changed to something like `$this->getTransactionNestingLevel()  === 1` and add `getTransactionNestingLevel` to `AbstractDao` like `$this->db->getTransactionNestingLevel()`

dpfaffenbauer:
thinking about that nestinglevel thing, no, that would not work. if you'd wrap a transaction yourself, this would fail then.

But: Is it even necessary to skip deletion of assets if nestedß

brusch:
@jdreesen since we're changing the signature of the `delete()` method (even if it's quite safe) I'd rather not merge this to `6.6`. Can you please rebase it onto `master`? Thanks a lot!

brusch:
@jdreesen please also add an appropriate upgrade note to the docs - just to be sure 😊
Thanks again!

PR #6863, opened 2020-07-30
Fix : Order not updating when quantity is updating

pdchaudhary:
Order not updating when quantity is updating in order details page from pimcore admin
Issue : https://github.com/pimcore/ecommerce-framework-bundle/issues/81
Editing quantity from pimcore order details page it is not updating order item total and order total so using this we can update order on chaning qunitiy,

  
  

Resolves #


fashxp:
thank you very much ... as this is a bug fix, could you please rebase (or create a new one) to 6.6 branch?

PR #6865, opened 2020-07-30
Fix : Order not updating when quantity is updating

pdchaudhary:
Order not updating when quantity is updating in order details page from pimcore admin
Issue : pimcore/ecommerce-framework-bundle#81
Editing quantity from pimcore order details page it is not updating order item total and order total so using this we can update order on chaning qunitiy,

### DIFF
--- a/bundles/EcommerceFrameworkBundle/OrderManager/Order/Agent.php
+++ b/bundles/EcommerceFrameworkBundle/OrderManager/Order/Agent.php
@@ -143,8 +143,7 @@ class Agent implements OrderAgentInterface
 
         return $note;
     }
-
-    /**
+   /**
      * change order item
      *
      * @param OrderItem $item
@@ -155,6 +154,7 @@ class Agent implements OrderAgentInterface
     public function itemChangeAmount(OrderItem $item, $amount)
     {
         // init
+       
         $amount = floatval($amount);
 
         // add log note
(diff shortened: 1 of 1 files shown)


bundles/EcommerceFrameworkBundle/OrderManager/Order/Agent.php
```diff
@@ -163,15 +163,35 @@ public function itemChangeAmount(OrderItem $item, $amount)
...
         $note->addData('amount.old', 'text', $oldAmount);
         $note->addData('amount.new', 'text', $amount);
-
+        $oldTotalPrice = $item->getTotalPrice();
+        $unitPrice = $oldTotalPrice/$oldAmount;
+        $newTotalPrice =$unitPrice*$amount;
         // change
         $item->setAmount($amount);
-
+        $item->setTotalPrice( $newTotalPrice);
```
fashxp, line 171: in that case we also would need to update netprice and tax information (if available)
image

bundles/EcommerceFrameworkBundle/OrderManager/Order/Agent.php
```diff
@@ -163,15 +163,35 @@ public function itemChangeAmount(OrderItem $item, $amount)
...
+       
+        $order = $item->getParent();
+        $subTotalNetPrice = $order->getSubTotalNetPrice();
+        $subTotalPrice = $order->getSubTotalPrice();
+        $totalNetPrice = $order->getTotalNetPrice();
+        $totalPrice = $order->getTotalPrice();
+        $replacedvalue  = $newTotalPrice -$oldTotalPrice;
+        $order->setSubTotalNetPrice($subTotalNetPrice +$replacedvalue);
+        $order->setSubTotalPrice($subTotalPrice+$replacedvalue);
+        $order->setTotalNetPrice($totalNetPrice+$replacedvalue);
```
fashxp, line 192: that calculation is wrong ... cannot reduce gross and net price with same amount. also need to update the tax information.

fashxp (changes requested):

PR #6866, opened 2020-07-30
[Document] Add properties for Area-Bricks

### LINKED ISSUE #6370: [Document] Properties for Area-Bricks
Similar to properties on elements and the successor of the `edit.php`. 
The `edit.php` is sometimes quite hard to configure/setup (layout issues, CSS issues, ...), so it makes more sense to move this functionality to a more structured and configurable concept. 

This is how it should work: 

There should be a button to configure / edit the properties the area brick: 
image

To enable this button the area brick class has to implement a certain interface: 
```php
<?php

namespace AppBundle\Document\Areabrick;

use Pimcore\Extension\Document\Areabrick\PropertySupportInterface;

class Blockquote extends AbstractAreabrick implements PropertySupportInterface
{
    
}
```

Implementing this interface allows you to freely define properties (similar to the element properties). 
If you'd like to preconfigure properties you have to implement `PropertyConfigurationInterface`: 

```php
<?php

namespace AppBundle\Document\Areabrick;

use Pimcore\Extension\Document\Areabrick\PropertySupportInterface;
use Pimcore\Extension\Document\Areabrick\PropertyConfigurationInterface;
use Pimcore\Model\Document\Tag\Area\Info;

class Blockquote extends AbstractAreabrick implements PropertySupportInterface
{
    /**
     * @param Info $info
     * @param array $params contains the custom brick params defined in areablock option `params` or `globalParams`
     * @return array
     */
    public function configureProperties(Info $info, array $params = []):array
    {
        return [
            'myName' => [
                'type' => 'input',
                'description' => 'foo bar awsome ...', // optional
                'default' => 'my custom text'
            ],
            'mySelectProp' => [
                'type' => 'select',
                'options' => [
                    'key1' => 'value1',
                    'key2' => 'value2',
                ]
            ]
        ];
    }
}
```

This is how the grid should look like. 
The description should be displayed on hover. 

image

Persisting: 
Maybe serializing them into the areablock is the best way to persist the values.

dvesh3:
Resolves #6370


dpfaffenbauer:
You know what I really don't like about this approach? You quite limit the types by a lot, Pimcore supports loads of Document Tags, but you only support these types: 'text', 'select', 'object', 'document', 'asset', 'bool(checkbox)'. But what about things like: 'Image', 'Multihref', 'Link', 'WYSIWYG' or even Custom Tags?

Like this, I don't see this as a good alternative for edit.php... And honestly, having Properties and edit.php doesn't make a lot sense. Why would you wanna add "custom" properties like in documents for bricks? You have to code anyway to do something with a property. 

If you take a look at the Toolbox, they solved that properly by also supporting all kind of Tags. Also custom ones.

dpfaffenbauer:
see, this is something the toolbox generates without writing any twig/html, just by using symfony configurations. I quite like that a lot, and it is totally user-friendly and everything is configurable and supports the all pimcore document tags.

Bildschirmfoto 2020-07-31 um 09 24 08

brusch:
@dpfaffenbauer This is a valid point, a shame that this hasn't been addressed already here  #6370
The original idea was to make the UI as lean as possible, therefore the "properties" approach with an editing grid. But yes, of course it would be also nice to use all types of editables. 
We'll discuss this topic internally.

dpfaffenbauer:
Well, it was addressed in #6370: https://github.com/pimcore/pimcore/issues/6370#issuecomment-626959556

brusch:
Well, not very specific, isn't it? But never mind, it's too late now anyway 😉

brusch:
closed, see #6923

PR #6868, opened 2020-07-31
Depend on `OutputInterface` instead of `StreamOutput` in `Pimcore\Console\Dumper`

jdreesen:
In #6839 I already changed the output dependecy from `ConsoleOutput` to `StreamOutput`, but this is not sufficient in all cases, so it should depend on `OutputInterface` instead.

A use case is when running commands programmatically, like this:
```php
function runCommand(Application $application, string $command, array $parameters = []): void
{
    $input = new ArrayInput(array_merge(['command' => $command], $parameters));
    $output = new BufferedOutput();

    if (0 !== $application->run($input, $output)) {
        throw new \RuntimeException(sprintf('Error running "%s": %s', $command, $output->fetch()));
    }
}

$kernel = ...
$application = new Application($kernel);
$application->setAutoExit(false);

runCommand($application, 'awesome:command', ['-v' => true]);
```

The `BufferedOutput` implements `OutputInterface` but does not extend `StreamOutput`.

### DIFF
--- a/lib/Console/Dumper.php
+++ b/lib/Console/Dumper.php
@@ -14,6 +14,7 @@
 
 namespace Pimcore\Console;
 
+use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Output\StreamOutput;
 use Symfony\Component\VarDumper\Cloner\VarCloner;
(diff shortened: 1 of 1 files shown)


brusch:
Thanks!

Issue #6462, opened 2020-05-27
[Data Objects] Bug with import when headers contains comma


j-ts:
# Bug Report

I found out that if you try to import CSV-file from Google Spreadsheets that contains coma in first row, Pimcore won't recognise columns correctly.
![image](https://user-images.githubusercontent.com/58213627/83003709-5ccf4180-a017-11ea-8427-088062e27f4b.png)
Well, I knew about it for a long time and fixed it by re-saving file with semicolon delimiter. But it would be more comfortable to fix such things inside import itself. 
### Expected behavior
Correct column recognising
### Actual behavior
Pimcore perceives file as _semicolon-separated_ with line terminator `0d0a` or `0a` (or sometimes can't recognise it at all), although Spreadsheets created comma-separated file with line terminator `0d0a`.
No changes in **CSV settings** or **Reload column configurations** features does't help.
### Steps to reproduce
1. Create a spreadsheet (https://spreadsheet.new/), add some data with comma in first row and download file.
![image](https://user-images.githubusercontent.com/58213627/83007919-0cf37900-a01d-11ea-832e-1c5139aa7cc9.png)

2. Import it to any class.

fashxp:
Importer got removed

PR #6870, opened 2020-07-31
fix: csv dataobject import reload column configuration

youwe-petervanderwal:
Error: CSV DataObject import, tab panel "CSV Settings", button "Reload column configuration" triggered error "Uncaught TypeError: this.callback is undefined" in updateColumnConfig
Caused by: Argument "showReload" was missing in constructor call
Introduced by: 62202949a129c806754f57c15498a60bf753943b
Relates to: Issue #6462

Steps to reproduce:
- Create a CSV file
- Start a CSV DataObject import with any random class
- Go to the CSV Settings tab
- Change settings and hit "Reload column configuration"

- Nothing happens, browser console displays: Uncaught TypeError: this.callback is undefined in updateColumnConfig


dvesh3:
thanks a lot!

PR #6871, opened 2020-07-31
[Element] Remove multiple triggers of POST_DELETE_FAILURE event

### LINKED ISSUE #6821: [Element] POST_DELETE_FAILURE could be triggered multiple times
In case you're deleting a page and that fails, let's say here: 
https://github.com/pimcore/pimcore/blob/2662d44a8828ad8fc7c8d878020b115f8ba61d69/models/Document.php#L823

Then the `DocumentEvents::POST_DELETE_FAILURE` event get's triggert 3 times here: 
https://github.com/pimcore/pimcore/blob/2662d44a8828ad8fc7c8d878020b115f8ba61d69/models/Document.php#L840
https://github.com/pimcore/pimcore/blob/a9f23dd2c7913a5d573e3dd101e3ae2da7751f63/models/Document/PageSnippet.php#L202
https://github.com/pimcore/pimcore/blob/e62ed40c25b802ae56c571ef70398088a8046e17/models/Document/Page.php#L100

Maybe we can use the `$isNested` argument to avoid this behavior.

dvesh3:
Resolves #6821

## Additional info  
As `PRE_DELETE` & `POST_DELETE` events are triggered only on abstract classes so `POST_DELETE_FAILURE` should only be triggered from there.
https://github.com/pimcore/pimcore/blob/2662d44a8828ad8fc7c8d878020b115f8ba61d69/models/Document.php#L791

Reason for `master` base https://github.com/pimcore/pimcore/issues/6821#issuecomment-662882535

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -249,7 +249,6 @@ class Concrete extends AbstractObject implements LazyLoadedFieldsInterface
             $this->commit();
         } catch (\Exception $e) {
             $this->rollBack();
-            \Pimcore::getEventDispatcher()->dispatch(DataObjectEvents::POST_DELETE_FAILURE, new DataObjectEvent($this));
             Logger::crit($e);
             throw $e;
         }
--- a/models/Document/Hardlink.php
+++ b/models/Document/Hardlink.php
@@ -259,7 +259,6 @@ class Hardlink extends Document
             $this->commit();
         } catch (\Exception $e) {
             $this->rollBack();
-            \Pimcore::getEventDispatcher()->dispatch(DocumentEvents::POST_DELETE_FAILURE, new DocumentEvent($this));
             Logger::error($e);
             throw $e;
         }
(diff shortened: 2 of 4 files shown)


brusch:
@dvesh3 but what if an exception is thrown e.g. here: 
https://github.com/pimcore/pimcore/blob/e62ed40c25b802ae56c571ef70398088a8046e17/models/Document/Page.php#L97

then the `DocumentEvents::POST_DELETE_FAILURE` is not triggered. 

I see the same problem with `PRE_DELETE` and `POST_DELETE`, I'd assume that `PRE_DELETE` is trigger before any action takes place, but that's actually not the case, since e.g. redirects are getting deleted here before `PRE_DELETE` is being triggered. 

I guess this means that we need to properly implement `$isNested` (see also #6860)  to decide whether to trigger the event or not.

brusch:
Closed. There will be a follow up PR that resolves the entire issues described above :)

PR #6872, opened 2020-07-31
[DataObject] Fix: Checkbox/Link/Slider/Datetime label width is not respected

blankse:


dvesh3:
for me it works when i set label width on panel
image

blankse:
@dvesh3 Yes direct in panel it works. The edit helper fill this option than.

It doesn't work in localizedfields.

dvesh3:
@blankse ok, I understand the problem. So then we also need to fix other types using `FieldContainer` like slider, datetime.

blankse:
@dvesh3 I added the fix to slider, datetime. The other fields with FieldContainer should respect the labelWidth option already

dvesh3:
thank you very much 🙏

PR #6873, opened 2020-07-31
[DataObject] Add css class for object field types

blankse:
We want add css properties to special object field types.
Can we add a css class with the type of the field?

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/input.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/input.js
@@ -52,7 +52,7 @@ pimcore.object.tags.input = Class.create(pimcore.object.tags.abstract, {
         var input = {
             fieldLabel: this.fieldConfig.title,
             name: this.fieldConfig.name,
-            componentCls: "object_field",
+            componentCls: "object_field object_field_type_" + this.type,
             labelWidth: 100
         };
 
@@ -100,7 +100,7 @@ pimcore.object.tags.input = Class.create(pimcore.object.tags.abstract, {
             this.updateCharCount(this.component, charCount);
 
             return Ext.create("Ext.Panel", {
-                cls: "object_field",
+                cls: "object_field object_field_type_" + this.type,
                 style: "margin-bottom: 10px",
                 layout: {
                     type: 'vbox',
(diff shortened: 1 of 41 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/tags/input.js
```diff
@@ -52,10 +52,13 @@ pimcore.object.tags.input = Class.create(pimcore.object.tags.abstract, {
         var input = {
             fieldLabel: this.fieldConfig.title,
             name: this.fieldConfig.name,
-            componentCls: "object_field",
             labelWidth: 100
         };
 
+        if (!this.fieldConfig.showCharCount) {
```
weisswurstkanone, line 58: what is the intention of this switch?

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/textarea.js
```diff
@@ -63,10 +63,13 @@ pimcore.object.tags.textarea = Class.create(pimcore.object.tags.abstract, {
             width: this.fieldConfig.width,
             height: this.fieldConfig.height,
             fieldLabel: this.fieldConfig.title,
-            componentCls: "object_field",
             labelWidth: labelWidth
         };
 
+        if (!this.fieldConfig.showCharCount) {
```
weisswurstkanone, line 69: same here. please explain.

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/input.js
blankse: I think the css class should only be once per field and at the outer component. If showCharCount is activated, the text field get a container with this class.

weisswurstkanone: That's actual a good question. I would agree with you.
However, if there is a charCount then it is not added at all?

blankse: If there is a charCount it will added to the Panel:
https://github.com/pimcore/pimcore/pull/6873/files/7186d47bf7f5d5f6eab95d03cd738db3ed6eb42a#diff-05180e356c10a77013221780b2e13ebaR106

weisswurstkanone: aah, it is getting late :-) then everything should be fine ...

weisswurstkanone:
thanks a lot @blankse

PR #6881, opened 2020-08-03
[Reverse object relation] More defensive check if item is allowed

BlackbitNeueMedien:
On the current implementation calling `allowObjectRelation(null)` would throw `Fatal error: Uncaught Error: Call to a member function getClassId() on null`.


weisswurstkanone:
thx!

PR #6894, opened 2020-08-04
Object UI breaks if checkbox or Calculated field is added under field container

### LINKED ISSUE #2919: Object UI breaks if checkbox or Calculated field is added under field container
### Expected behavior 
  Object UI must not break , if checkbox or Calculated field is added under field container
### Actual behavior  
  Object UI breaks if checkbox or Calculated field is added under field container
### Steps to reproduce  
 Just add a checkbox under field container.The ui of  field container breaks.

add checkbox in class

ui of object breaks
After Adding checkbox in field container all the data fields inside that field container becomes invisible.

Same is the case with Calculated field.
Add Calculated field under field container. 
calculated field in class

calculated field breaked ui

weisswurstkanone:
resolves #2919


blankse:
@weisswurstkanone Destination branch should be 6.6, right?

weisswurstkanone:
thx, will create a new one

PR #6897, opened 2020-08-04
[Installer] Added option to skip database connection setup (database.yml)

### LINKED ISSUE #6638: [Installer] Possibility to skip database connection setup (database.yml)
# Bug Report

I am trying to update the installer for Pimcore on Platform.sh, and running into an issue.  Specifically, even though an `/appconfig/installer.yml` file is included, the installer is tryign to write an `/app/config/local/database.yml` file.  I want it to skip that entirely, because the DB information will always be provided via environment variables and injected into the container through a custom parameters file.

I don't believe this problem existed on earlier versions, but with the latest Pimcore release I'm running into this wall and haven't figured out a way around it.

### Expected/Desired behavior

Installer detects that DB information is already available in the container and just keeps on going without trying to write any YAML files to disk.

### Actual behavior

The installer fails with:

```
                                                                                                                        

 * Warning: file_put_contents(/app/app/config/local/database.yml): failed to open stream: Read-only file system
```

Unfortunately I think this hard-blocks the Pimcore template on Platform.sh for the time being.

dvesh3:
Resolves #6638

## Additional info  
`./vendor/bin/pimcore-install  --skip-database-config` use this option to skip writing database config (database.yml)

### DIFF
--- a/bundles/InstallBundle/Command/InstallCommand.php
+++ b/bundles/InstallBundle/Command/InstallCommand.php
@@ -177,6 +177,11 @@ class InstallCommand extends Command
                 null,
                 InputOption::VALUE_NONE,
                 'Do not abort if a <comment>system.yml</comment> file already exists'
+            )->addOption(
+                'skip-database-config',
+                null,
+                InputOption::VALUE_NONE,
+                'Do not write a database config file: <comment>database.yml</comment>'
             );
 
         foreach ($this->getOptions() as $name => $config) {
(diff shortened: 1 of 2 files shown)


yariksheptykin:
@dvesh3 I was wondering if we could start adding some test coverage along with PRs. This changeset is quite trivial so it should not be that difficult to cover just this bit of Installer with a unit test. What do you think?

dvesh3:
@yariksheptykin yes, seems good opportunity. Do you want to contribute?

yariksheptykin:
> @yariksheptykin yes, seems good opportunity. Do you want to contribute?

I could give it a try. Is there any way for adding changes into your PR? Fork your pimcore version and add a PR against installer_skip_db_config?

dvesh3:
@yariksheptykin this branch is on pimcore/pimcore and i think it is not possible for you to add changes to this PR coz you need write access. Better just provide a seperate PR with you changes.

jdreesen:
It should be possible to create a PR with this branch (`installer_skip_db_config`) as target branch.

PR #6902, opened 2020-08-05
InstallCommand: Correct info to log file path

blankse:
Log dir was changed here: https://github.com/pimcore/pimcore/commit/daba49f34e064cf72fca3cf0047981670a245759


dvesh3:
thanks! 🙏

PR #6920, opened 2020-08-07
Fix: Copying an object with mandatory and unique field is not possible.

blankse:
See #5243

Steps to reproduce:
- Create a Class in the pimcore demo with only 1 input field (mandatory, unique)
- Create a object
- Copy this object
- Error when paste:
image

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -148,6 +148,7 @@ class Service extends Model\Element\Service
                 if ($fieldDefinition->getUnique()) {
                     $new->set($fieldDefinition->getName(), null);
                     $new->setPublished(false);
+                    $new->setOmitMandatoryCheck(true);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


dvesh3:
I can reproduce the issue but the question is, why it is required to set **omitMandatory** check when publish status is **false**. It should not be checked by default when just saving the object. @blankse do you know the reason?

dvesh3:
@blankse or this already fixes the issue https://github.com/pimcore/pimcore/pull/6769? can you please try on master

blankse:
@dvesh3 Yes, in master branch it works. Can we backport #6769 to 6.6 branch?

brusch:
@blankse unfortunately we can't backport #6769 because it changes the behavior, so we agreed to NOT do this change in a bug-fix release.

dvesh3:
already fixed in #6769

PR #6923, opened 2020-08-07
[Document] Areablock: Editable Dialog Box

brusch:
alternative approach for #6866 / #6370


dpfaffenbauer:
so basically quite similar to the toolbox approach?

brusch:
Yes

brusch:
## Example

```php
<?php

/**
 * Pimcore
 *
 * This source file is available under two different licenses:
 * - GNU General Public License version 3 (GPLv3)
 * - Pimcore Enterprise License (PEL)
 * Full copyright and license information is available in
 * LICENSE.md which is distributed with this source code.
 *
 *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
 *  @license    http://www.pimcore.org/license     GPLv3 and PEL
 */

namespace AppBundle\Document\Areabrick;

use Pimcore\Extension\Document\Areabrick\EditableDialogBoxConfiguration;
use Pimcore\Extension\Document\Areabrick\EditableDialogBoxInterface;
use Pimcore\Model\Document;
use Pimcore\Model\Document\Tag\Area\Info;

class WysiwygWithImages extends AbstractAreabrick implements EditableDialogBoxInterface
{
    /**
     * {@inheritdoc}
     */
    public function getName()
    {
        return 'WYSIWYG w. Images';
    }

    public function getEditableDialogBoxConfiguration(Document\Tag $area, ?Info $info): EditableDialogBoxConfiguration
    {
        $config->setWidth(600);
        //$config->setReloadOnClose(true);
        $config->setItems([
            'type' => 'tabpanel',
            'items' => [
                [
                    'type' => 'panel',
                    'title' => 'Tab 1',
                    'items' => [[
                        'type' => 'checkbox',
                        'name' => 'myDialogCheckbox'
                    ], [
                        'type' => 'input',
                        'name' => 'myDialogInput'
                    ]]
                ],
                [
                    'type' => 'panel',
                    'title' => 'Tab 2',
                    'items' => [[
                        'type' => 'input',
                        'name' => 'myNumber3'
                    ]]
                ]
            ]
        ]);


        return $config;
    }
}

```

image

## Current status
- `input` and `checkbox` editables can be used
- 2 layouts: `panel` and `tabpanel` 
- editables need to implement `render()` method to be supported

## Concerns
- No support for `block` editable

brusch:
@dvesh3 @dpfaffenbauer @markus-moser @aarongerig @jdreesen WDYT? Thanks for your feedback!

dpfaffenbauer:
Why don't just use straight html as result? So you don't need to change the editables to support the render method. Basically sort of a auto-generate edit.html.twig?

But: I do like the layouts, tabs and panel. gives it more structure, but can also be solved with HTML output and no direct use of extjs.

brusch:
@dpfaffenbauer the idea was to initialize/render the editables in the dialog box only when needed, because this saves a lot of initial loading time. Additionally I think that the current approach (rendering them in the constructor) is anyway not really a good practice. And changing existing custom editables to support the `render()`method shouldn't be too hard and just a matter of minutes each. 

Regarding ExtJS vs. plain HTML: I don't have a preference there, I just used ExtJS for the PoC because it was quicker 😉

dpfaffenbauer:
@brusch Yes, but this makes things more complicated, you then have to implement the JS render and also the "traditional" output of the field, right?

Re initialize/render: I think the HTML overhead is not too bad having them "rendered" directly on view-rendering. We do that with the toolbox for years now and have never had any performance issues.

brusch:
@dpfaffenbauer no, the `render()` method is replacing the traditional way 😊
Another disadvantage of the view-rendering is that you have to move the DOM nodes to get the content into the dialog box, which is not really nice -> see also the code here: https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Resources/public/js/pimcore/document/tags/areablock.js#L930

markus-moser:
@brusch I like the approach. I looked at the first PR too in the past and I'm glad that you rethinked it again in this way. Without deep knowledge on how it works internally I'd agree with what @dpfaffenbauer said but am sure that you will find the right implementation details. But I would appreciate it very much if either all or most of the regular document tags would work. The panel/layout approach definitely is much nicer then the simple grid.

brusch:
Added the possibility to let the editable decide how it should get rendered in the dialog box by implementing the `renderInDialogBox()` method. If this method isn't implemented the editable gets rendered just like in the normal editmode and wrapped in a fieldset container.

dvesh3 (approved):

julmot:
@brusch @fashxp @dvesh3 We're using the edit template only, for all bricks. The reason is, that:

* We don't have to support Pimcore brick HTML manipulations, so it's easier and more efficient for us to build HTML with an isolated FE template that doesn't necessarily have to be built specifically for Pimcore
* We can still edit the data at the place where it's visible
* We're avoiding ugly frontends with UI controls that would mess it up

The downside is that we had to build our own "hack" to manipulate Pimcore WYSIWYG to look like actual form fields. I've seen that in Pimcore v7 we should use the new [controller based editable dialog instead](https://pimcore.com/docs/pimcore/current/Development_Documentation/Documents/Editables/Areablock/Bricks.html#page_Editable-Dialog-since-6-8). The refactoring will take us a couple of time so we hope it's worth it. While the documentation covers simple field types like select and input, it doesn't cover a documentation about all valid types and doesn't mention blocks.

How do you translate the following Twig template using `pimcore_iterate_block` and `pimcore_block` into the new edit dialog controller based approach?

```twig
{%- set utilityNavLinks = [] -%}
{%- for i in pimcore_iterate_block(pimcore_block('utilityNavLinks')) -%}
  {%- set link = pimcore_link('link-' ~ i) -%}
  {%- set utilityNavLinks = utilityNavLinks | merge([{
    title: link.getText(),
    href: link.getHref(),
    selected: link.getClass() == 'selected'
  }]) -%}
{%- endfor -%}
```

brusch:
@julmot unfortunately blocks cannot be supported 😐

julmot:
@brusch Oh! Then, what's the plan once the edit template approach is really removed? Are we forced to drop our entire ecosystem of bricks into ugly WYSIWYG fields in the frontend, making our frontend Pimcore specific and make it difficult for users to differentiate between brick data fields like input and form fields like select or block that just have to appear somewhere where they cause the less damage?

brusch:
We're still thinking how we can make blocks possible, but so far all approaches are not really satisfying in terms of usability. 

But if your whole ecosystem depends on that, you can just extend the Areablock and integrate the edit.php support again to your custom data-type.

julmot:
@brusch As far as I understand the intention of this change, it should improve the editing experience in the CMS area. I very much appreciate the intention, because our customers already mentioned that the edit overlay is – I'm quoting "the worst CMS they ever worked with". I didn't make this experience, however, I can understand the points. We have editables that were built for WYSIWYG only and we're embedding these fields in an edit overlay where they are e.g. invisible since the border only gets active on focus – so we had to build our own "hack" to make them look like native form fields which is still not a nice looking solution.

Wouldn't it be possible to have an option per editable to specify what pimcore generates? Let's say we'd use the usual WYSIWYG approach with only a view template, then we could keep it like it's now.  However, for those using the edit overlay with edit template they could specify an option like e.g. `nativeFormField: true` per editable which would then render a native form control rather than an editable div or soemthing like that.

While this would not mean a paradigm change, it would improve the experience massively. We could then still build our own logic for tabs etc.

dpfaffenbauer:
@julmot you can still build your own logic and your own editables, I don't see a problem here?

julmot:
@dpfaffenbauer We would like to continue using the edit overlay, but this is only possible if we also have blocks. Blocks, however, arne't possible with the current alternative for the deprecated edit overlay. So, this is "the problem".
Therefore, I've mentioned the possibility to improve the experience of the edit overlay (what seems to be the intention of this issue), without loosing the edit template approach.

christopher-siegel:
**First of all:** really great to see pimcore implement this! Much better than the first iteration with properties 👍

> If this method isn't implemented the editable gets rendered [...] **and wrapped in a fieldset container.**

What's the intention behind wrapping every editable inside its own fieldset container @brusch ? 

To me it just looks like a lot of space is "wasted" around the elements:
Bildschirmfoto 2020-11-04 um 08 45 46

It's great we can change it with `renderInDialogBox()`, but I think the default doesn't need fieldset containers? 
Or... Is it possible to add multiple editables into the same fieldset?

Maybe I'm missing something though...

brusch:
@christopher-siegel well, the idea was to have a generic as possible way how to display the label. 
If you have any other and better idea, please share 😊

mittererr:
A possibility for a simple textbox would be great. Just for some explaining comments in the DialogBox. A divider like "<hr>" would be great too.

BenHeer:
> @brusch Oh! Then, what's the plan once the edit template approach is really removed? Are we forced to drop our entire ecosystem of bricks into ugly WYSIWYG fields in the frontend, making our frontend Pimcore specific and make it difficult for users to differentiate between brick data fields like input and form fields like select or block that just have to appear somewhere where they cause the less damage? Having a fixed amount of links (slots) in the above example or e.g. in a navigation brick with a big menu will it make difficult to manage content for users as we'd need to have 100-200 fixed slots. Also it would mean that we're loosing all the existing brick data.

I have the same problem as you. Especially with pimcore_iterate_block and pimcore_block. My first solution is to use snippets for blocks and place them over the editable dialog. Im not happy with that, but its possible.

PR #6924, opened 2020-08-10
Update getters/setters in Breadcrumbs.php

manud99:
I did some minor changes on Breadcrumbs.php

* Return `$this` from the `setTemplate` method.
* Make `getPartial` and `setPartial` real aliases of `getTemplate` and `setTemplate` by actually calling the alias methods.
* Reorder: getters before setters.


dvesh3:
thanks! 🙏

PR #6926, opened 2020-08-10
Micro-optimisation: Change from spl_object_hash() to _id()

NiklasBr:
Resolves no specific issue. However, I am currently having trouble with performance with the Navigation system. While hunting down the reason for my problems I found some quick micro-optimisations which could be done.

Moving from `spl_object_hash()` to `spl_object_id()` should be faster by [skipping XOR:ing and using integers rather than strings](https://github.com/php/php-src/pull/2611) ([polyfill exists](https://github.com/symfony/polyfill/commit/c3ce3729c1697d865e0ae7ba7af8a289bf672f33)). I could see some memory gains (250kB/request with a large menu).

Using the root namespace (e.g. https://github.com/symfony/symfony/issues/21020) for the `spl` function saved two steps of stack trace for each call when Opcache was active, performance improvements was a few hundreds of a second for the request, so it did not do much. But it all adds up. :)

### DIFF
--- a/lib/Navigation/Page.php
+++ b/lib/Navigation/Page.php
@@ -1192,13 +1192,13 @@ abstract class Page extends Container
     }
 
     /**
-     * Returns a hash code value for the page
+     * Returns a unique code value for the page
      *
-     * @return string  a hash code value for this page
+     * @return int a unique code value for this page
      */
     final public function hashCode()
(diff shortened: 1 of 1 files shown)


brusch:
Thank you! 👌

PR #6929, opened 2020-08-11
[Versions] Support version-compressing for stream wrappers which do not support gzopen

BlackbitNeueMedien:
When stream wrappers do not support binary operations, version compression will fail. This is for example the case for Google Cloud Storage and AWS S3, see https://github.com/googleapis/google-cloud-php-storage/blob/9412bf831a3fdd895378203b7e42b782deee8795/src/StreamWrapper.php#L147-L193

To nevertheless support version compressing for such cloud storage providers this PR adopts the suggested way from https://github.com/aws/aws-sdk-php/issues/1576#issuecomment-404446446

### DIFF
--- a/lib/Maintenance/Tasks/VersionsCompressTask.php
+++ b/lib/Maintenance/Tasks/VersionsCompressTask.php
@@ -40,7 +40,6 @@ final class VersionsCompressTask implements TaskInterface
     {
         $perIteration = 100;
         $alreadyCompressedCounter = 0;
-        $overallCounter = 0;
 
         $list = new Version\Listing();
         $list->setCondition('date < ' . (time() - 86400 * 30));
@@ -59,8 +58,6 @@ final class VersionsCompressTask implements TaskInterface
             $versions = $list->load();
 
             foreach ($versions as $version) {
-                $overallCounter++;
-
                 if (file_exists($version->getFilePath())) {
                     gzcompressfile($version->getFilePath(), 9);
                     @unlink($version->getFilePath());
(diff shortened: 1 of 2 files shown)


brusch:
@BlackbitNeueMedien can you please check PhpStan: https://travis-ci.com/github/pimcore/pimcore/jobs/370616824#L1088 
Thanks a lot!

BlackbitNeueMedien:
@brusch Fixed. 
Travis now fails all tests for not being able to install `imagick`

brusch:
@BlackbitNeueMedien I'm just thinking, ... why not just use `deflate_*()` if it works in all cases?

BlackbitNeueMedien:
Hmm, do not know if there are certain optimizations in `gzopen()` and the other gz functions. But if you prefer cleaner code, give me a sign and I will edit this PR to only use `deflate_*` functions.

brusch:
@BlackbitNeueMedien I think we should keep it as simple as possible - so if using `deflate_*` is backward-compatible (reading should work as well) we should go for it. 
Thanks a lot!

BlackbitNeueMedien:
Fixed the code.
As `deflate_*` methods are present since PHP 7, I do not see a problem as Pimcore requires PHP >= 7.2 anyway.

brusch:
@BlackbitNeueMedien Thanks a lot! 
What I meant was if the `deflate_*` produced files are compatible with `gzdecode()` used here: 
https://github.com/pimcore/pimcore/blob/9c9c11fe6774e076eb0fd35e7bda2ee2b94c4a24/models/Version.php#L403

Can you confirm that restoring a version compressed by `deflate_*` works? 
Thanks again!

BlackbitNeueMedien:
Yes, it works (tested with Google Cloud Storage and https://github.com/googleapis/google-cloud-php-storage/). I can see the version details on the versions tab of the data object.

brusch:
Grazie Mille 🙏👌

PR #6931, opened 2020-08-11
Fix collecting condition parameters

t-zilla:
* Fix collecting parameters which were added to DataObject listing via `$listing->addConditionParam($key, $value, $concatenator)`

## Additional info

Example listing that fails:

```
$listing = new \Pimcore\Model\DataObject\Product\Listing();
$listing->setObjectTypes(['object', 'folder', 'variant']);
$listing->setOrderKey('o_id');
$listing->setOrder('ASC');
$listing->setOffset(0);
$listing->setLimit(25);
$listing->setUnpublished(true);
$listing->addConditionParam('3256 = 3256 AND suppliers LIKE ?', '%,15793,%', 'AND');
$listing->addConditionParam('8665 = 8665 AND manufacturer__id = ? AND manufacturer__type = ?', ['7', 'object'], 'AND');
$listing->load();  // Throws exception
```

```
An exception occurred while executing 'SELECT object_localized_6_en.o_id as o_id, `object_localized_6_en`.`o_type` FROM `object_localized_6_en` WHERE ((3256 = 3256 AND suppliers LIKE ?) AND (8665 = 8665 AND manufacturer__id = ? AND manufacturer__type = ?) AND object_localized_6_en.o_type IN ('object','folder','variant')) ORDER BY `o_id` ASC LIMIT 25' with params ["7", "object"]: SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens
```

There should be 3 parameters but only 2 are passed.

### DIFF
--- a/lib/Model/Listing/AbstractListing.php
+++ b/lib/Model/Listing/AbstractListing.php
@@ -310,8 +310,12 @@ abstract class AbstractListing extends AbstractModel implements \Iterator
                 // If there is not a placeholder, ignore value!
                 if (!$value['ignore-value']) {
                     if (is_array($value['value'])) {
-                        foreach ($value['value'] as $v) {
-                            $params[] = $v;
+                        foreach ($value['value'] as $k => $v) {
+                            if (is_int($k)) {
+                                $params[] = $v;
+                            } else {
+                                $params[$k] = $v;
+                            }
                         }
                     } else {
                         $params[] = $value['value'];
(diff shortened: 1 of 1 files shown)


lib/Model/Listing/AbstractListing.php
```diff
@@ -310,8 +310,8 @@ public function getCondition()
                 // If there is not a placeholder, ignore value!
                 if (!$value['ignore-value']) {
                     if (is_array($value['value'])) {
-                        foreach ($value['value'] as $k => $v) {
-                            $params[$k] = $v;
+                        foreach ($value['value'] as $v) {
+                            $params[] = $v;
```
dvesh3, line 314: @t-zilla I'm afraid, this change will break named params.

t-zilla: @dvesh3 Good point, I accounted for it in the next commit

dvesh3:
thanks a lot!

PR #6936, opened 2020-08-12
fixing matchRequest resolving for dataobjects with urlslug

sgc-fireball:
# Issue
Call: \Pimcore\Model\DataObject\Data\UrlSlug::resolveSlug("/my/path", 1)
SELECT * FROM object_url_slugs WHERE slug = "/my/path" AND siteId = 1 OR siteId = 0 ORDER BY siteId DESC LIMIT 1

# Solution
SELECT * FROM object_url_slugs WHERE slug = "/my/path" AND (siteId = 1 OR siteId = 0) ORDER BY siteId DESC LIMIT 1

### DIFF
--- a/models/DataObject/Data/UrlSlug.php
+++ b/models/DataObject/Data/UrlSlug.php
@@ -330,7 +330,7 @@ class UrlSlug implements OwnerAwareFieldInterface
      *
      * @return UrlSlug|null
      */
-    public static function resolveSlug($path, $siteId = 0)
+    public static function resolveSlug($path, int $siteId = 0)
     {
         $cacheKey = $path . '~~' . $siteId;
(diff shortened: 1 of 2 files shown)


models/DataObject/Data/UrlSlug.php
```diff
@@ -330,7 +330,7 @@ public static function createFromDataRow($rawItem): UrlSlug
      *
      * @return UrlSlug|null
      */
-    public static function resolveSlug($path, $siteId = 0)
+    public static function resolveSlug($path, int $siteId = 0)
```
brusch, line 333:
```suggestion
    public static function resolveSlug($path, $siteId = 0)
```

brusch:
Thanks a lot! 🙏😊

PR #6939, opened 2020-08-12
Pass and consume correct error messages when declining a workflow transition

roland4432:

The controller passes non-translated default error messages to the user trying to commit a transition while using workflows for objects.
These changes get the correct error messages being passed from the event, collects them, passes them to the transitionPanel and translates each item via backend translations


roland4432:
PHPStan failed, though the given exception is not right. How can I rerun the checks?

brusch:
@roland4432 restarted the job 😊

roland4432:
TY sir..
Do you think that failing check is correct or a false positive?

brusch:
@roland4432 there's definitely an issue with Symfony 3.4
The method `buildTransitionBlockerList()` doesn't exists in 3.4 it seems. 
See: https://travis-ci.com/github/pimcore/pimcore/jobs/371294641#L1086

roland4432:
I've checked the logs and was irritated for a short time, that 5 other jobs running symfony 3.4 passed, but those jobs are doing other things.
I took a look at https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Workflow/Workflow.php and the assumption was correct, that the method doesn't exist in symfony 3.4.

I still think the feature to pass the correct translated messages is important. Will pimcore 7.0 drop symfony 3.x support so maybe this might come in handy at a later time?

brusch:
@roland4432 yep, let's schedule it for 7 which drops Sf 3.4 support

roland4432:
Shall I take a note or do you have any internal workflow to postpone this PR?

brusch:
I put this PR on the 7  milestone already (see right sidebar 😉)

roland4432:
Wheeeeee... perfect.. thank you

brusch:
@roland4432 can you please rebase/merge your branch onto the current master? 
Thanks a lot!

roland4432:
Merged the current master into my branch @brusch

brusch:
Thanks a lot!

Issue #6888, opened 2020-08-04
[Document] Change PHP/JS namespace from Document\Tag to Document\Editable


brusch:
- Makes the naming consistent and easier to understand. 
- incl. BC layer
- deprecate old namespace

Also some class names and methods need to be renamed: 
```
\Pimcore\Event\Model\Document\TagNameEvent
\Pimcore\Templating\Renderer\TagRenderer
\Pimcore\Templating\Renderer\TagRenderer::tagExists() // ... and some more methods in this class
\Pimcore\Document\Tag\TagUsageResolver // ... including some methods in this class

Classes and methods in this folder: 
lib/Document/Tag
```

PR #6940, opened 2020-08-12
[Documents] Add new css class selector for document editables & deprecate old selector

dvesh3:
Resolves partially #6888


brusch:
@dvesh3 we have some more conflicts here as well 😊🙏

dvesh3:
@brusch 
> we have some more conflicts here as well 😊🙏

done

PR #6949, opened 2020-08-13
fix [Imagick]: add compatibility fix for Imagick 3.4.4 and ImageMagick 7

ssiergl:
regression from https://github.com/pimcore/pimcore/issues/5505

  
  

## Changes in this pull request  

fixes:

image

## Additional info  

see here: 

https://github.com/Imagick/imagick/blob/060c34ac7e7600ca8443059cc6104f80e8c7ef13/imagick_helpers.c#L1741

and here:

https://github.com/ImageMagick/ImageMagick/blob/fab704b6fc4ca4736278601e3790ce54ae684db4/MagickCore/image.h#L41

PR #6950, opened 2020-08-13
change[Thumbnails]: make auto clipping configurable

ssiergl:
https://github.com/pimcore/pimcore/issues/5505 introduced auto clipping for images
with embedded clipping path.

This is a pretty wonky feature and often broken through several ImageMagick releases. As well
this is a high cpu intensive task. People should be carful to use it.

hence we make it globaly configurable

  
  

follow up on discussion:

https://github.com/pimcore/pimcore/issues/5505

and regression fix

https://github.com/pimcore/pimcore/pull/6949

## Additional info  

As mentioned a pretty wonky ImageMagick feature especially if used with Imagick. Several stable ImageMagick releases are broken (again current 7.0.10 release since 8 minor releases) as well some popular 6.9 releases.

The most reliably implementation would be to call convert / imagick over cli. We are doing this over enqueue queues in background if new images are uploaded (see Screenshot). This works on most imagemagick releases and workaround the shity Imagick ImageMagick MagickWall alpha implementation.

image

But........TBH I do not see a direct cli call in pimcore imagick core adapter. This way we would need to add server platform dependent calls....this should never be in pimcore core.

hence just let the admin decide to enable/disable clipping and check beforehand all requirements

### DIFF
--- a/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
+++ b/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
@@ -327,7 +327,7 @@ If you prefer not using WebP, you can disable the support by adding the followin
 ```
 
 ## Clipping Support 
-Automatically clip images with embedded clipping path (8BIM / Adobe profile meta data) to generate thumbnails. 
+Images with an embedded clipping path (8BIM / Adobe profile meta data) are automatically clipped when generating thumbnails of them. 
     
 If you do not want to use thumbnail auto clipping, you can disable the support by adding the following config option: 
 ```yml
--- a/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
+++ b/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
@@ -326,6 +326,17 @@ If you prefer not using WebP, you can disable the support by adding the followin
                 webp_auto_support: false
 ```
 
+## Clipping Support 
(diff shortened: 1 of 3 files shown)


NiklasBr:
Why not respect the image metadata? If we respect metadata such as colour space and rotation, then it makes sense to also respect clipping masks--especially when this behaviour is only applied to JPEG images and not all image types.

brusch:
I agree with @NiklasBr 
Either way, we cannot change the behavior now as this would be a breaking change. The only solution I can think of would be to set the default value of this config option to `true` as this wouldn't change the behavior.

ssiergl:
@NiklasBr

yeah basically I have the same opinion......I just want to prevent that a lot pimcore user will flame pimcore for not working clipping or very high load.....which basically is always an Imagick/Imagemagick problem....

That beeing said I am ok with `true` as default....I will change it and changing the doc accordingly

ssiergl:
@brusch @NiklasBr done...enabled by default now

doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
```diff
@@ -326,6 +326,17 @@ If you prefer not using WebP, you can disable the support by adding the followin
                 webp_auto_support: false
 ```
 
+## Clipping Support 
+Automatically clip images with embedded clipping path (8BIM / Adobe profile meta data) to generate thumbnails.
```
brusch, line 330:
```suggestion
Images with an embedded clipping path (8BIM / Adobe profile meta data) are automatically clipped when generating thumbnails of them. 
```

brusch:
Thanks a lot! Since it's not actually a bug fix, I've changed the base to `master`  and changed the target milestone to 6.8.0.

PR #6953, opened 2020-08-15
Add --permit-file-read flag to ghostscript command.

### LINKED ISSUE #6911: Ghostscript versions >9.50 have problem to get page count
# Bug Report

If you have a ghostscript version >9.50 the maintenance script don't create thumbnails
```
 Exception: Unable to get page-count of /var/www/pimcore/web/var/assets/test.pdf in /var/www/pimcore/vendor/pimcore/pimcore/lib/Document/Adapter/Ghostscript.php:148
```

If I test the command I get following error:
```
gs -dNODISPLAY -q -c '(/var/www/pimcore/web/var/assets/test.pdf) (r) file runpdfbegin pdfpagecount = quit'
Error: /invalidfileaccess in --file--
Operand stack:
   (/var/www/pimcore/web/var/assets/test.pdf)   (r)
Execution stack:
   %interp_exit   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--   --nostringval--   --nostringval--   false   1   %stopped_push   .runexec2   --nostringval--   --nostringval--   --nostringval--   2   %stopped_push   --nostringval--
Dictionary stack:
   --dict:731/1123(ro)(G)--   --dict:0/20(G)--   --dict:75/200(L)--
Current allocation mode is local
Last OS error: Permission denied
GPL Ghostscript 9.52: Unrecoverable error, exit code 1
```

It seems ghostscript has a SAFER mode on and we need to add a"permit-file-read" param:
https://stackoverflow.com/questions/40156190/ghostscript-error-invalidfileaccess-in-file

manud99:
This PR resolves #6911 

It adds the `--permit-file-read` flag with the file's path so that ghostscript is allowed to read the file.

**Note:** I am not sure, if this still works with older Ghostscript versions. I used Ghostscript version 9.52 to test it.

### DIFF
--- a/lib/Document/Adapter/Ghostscript.php
+++ b/lib/Document/Adapter/Ghostscript.php
@@ -164,6 +164,7 @@ class Ghostscript extends Adapter
     {
         $command = self::getGhostscriptCli() . " -dNODISPLAY -q";
 
+        // Adding permit-file-read flag to prevent issue with Ghostscript's SAFER mode which is enabled by default as of version 9.50.
         if ($this->getVersion() >= 9.50) {
             $command .= " --permit-file-read='" . $this->path . "'";
         }
@@ -174,9 +175,9 @@ class Ghostscript extends Adapter
     }
 
     /**
-     * Get the version of the installed CLI.
+     * Get the version of the installed Ghostscript CLI.
      *
-     * @return float|null
+     * @return float
      * @throws \Exception
      */
     protected function getVersion()
--- a/lib/Document/Adapter/Ghostscript.php
+++ b/lib/Document/Adapter/Ghostscript.php
@@ -141,7 +141,7 @@ class Ghostscript extends Adapter
      */
     public function getPageCount()
     {
-        $pages = Console::exec(self::getGhostscriptCli() . " -dNODISPLAY -q -c '(" . $this->path . ") (r) file runpdfbegin pdfpagecount = quit'", null, 120);
+        $pages = Console::exec(self::getGhostscriptCli() . " -dNODISPLAY -q --permit-file-read='" . $this->path . "' -c '(" . $this->path . ") (r) file runpdfbegin pdfpagecount = quit'", null, 120);
         $pages = trim($pages);
 
         if (!is_numeric($pages)) {
--- a/lib/Document/Adapter/Ghostscript.php
+++ b/lib/Document/Adapter/Ghostscript.php
@@ -27,6 +27,11 @@ class Ghostscript extends Adapter
     protected $path;
 
     /**
+     * @var float|null
+     */
+    private $version = null;
+
+    /**
      * @return bool
      */
     public function isAvailable()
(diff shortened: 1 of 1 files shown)


dvesh3:
@manud99 how about we add a method to check version in `/lib/Document/Adapter/Ghostscript.php` and only add `--permit-file-read` parameter, if current version > 9.50. so something like:
```php 
public static function getVersion()
    {
        try {
            if (self::isAvailable()) {
                return (float) console::exec('gs --version');
            }
        } catch (\Exception $e) {
            Logger::warning($e);
        }

        return null;
    }
```

manud99:
@dvesh3 I think it depends on whether an older ghostscript versions throws an "Unknown argument" exception. If it does, a `getVersion` method would be the solution I'd use. Does it break with a ghostscript version older than `9.52`?

dvesh3:
@manud99 yes, it breaks on v9.19 with error "/undefinedfilename in (-c)" when adding `--permit-file-read` parameter

manud99:
Okay, then I will add a `getVersion` method for building the right command as you have suggested.

manud99:
@dvesh3 I added a `getVersion` method to build the right command. If there is a ghostscript version newer then 9.50, it will add the `--permit-file-read` flag. I tested it with version 9.27 and 9.52.

lib/Document/Adapter/Ghostscript.php
```diff
@@ -141,14 +146,46 @@ public function getPdf($path = null)
...
+
+        $command .= " -c '(" . $this->path . ") (r) file runpdfbegin pdfpagecount = quit'";
+
+        return $command;
+    }
+
+    /**
+     * Get the version of the installed CLI.
+     *
+     * @return float|null
```
dvesh3, line 179: please correct the return type..not possible to return null

lib/Document/Adapter/Ghostscript.php
```diff
@@ -141,14 +146,46 @@ public function getPdf($path = null)
...
+
+    /**
+     * @return string
+     * @throws \Exception
+     */
+    protected function buildPageCountCommand()
+    {
+        $command = self::getGhostscriptCli() . " -dNODISPLAY -q";
+
+        if ($this->getVersion() >= 9.50) {
```
dvesh3, line 167: please add a a short about version check and adding permit-file flag for better readability.

manud99:
@dvesh3 Return type updated and inline comment added. Please review again.

dvesh3:
@manud99 one last change and very imp 😅 ... we need to change the target base to maintenance branch `6.7` since it is a bug fix. https://github.com/pimcore/pimcore/blob/master/CONTRIBUTING.md#contributing-pull-requests thanks!

manud99:
@dvesh3 Okay, no problem. Thanks for notifying. It took me some time to figure out how to rebase the branch, but I got it working. Hopefully everything is fine now 😅

dvesh3:
@manud99 thanks! it looks good 👍

dvesh3:
thank you!

PR #6957, opened 2020-08-17
Fix PhpDoc type

jdreesen:
The type of the `$data` parameter in [`Pimcore\Model\Document\Tag\TagInterface::setDataFromEditmode($data)`](https://github.com/pimcore/pimcore/blob/master/models/Document/Tag/TagInterface.php#L53-L60) is `mixed`.


brusch:
That was the wrong target branch... current one is  6.7. @weisswurstkanone can you cherry-pick?

PR #6964, opened 2020-08-18
[Tags] Add Tag::getByPath() and Tag::__toString()

BlackbitNeueMedien:
Similar to objects, assets and documents this PR adds the methods `Tag::getByPath()` and `Tag::__toString()`

### DIFF
--- a/models/Element/Tag.php
+++ b/models/Element/Tag.php
@@ -160,8 +160,17 @@ class Tag extends Model\AbstractModel
         return $tag->getDao()->getElementsForTag($tag, $type, $subtypes, $classNames, $considerChildTags);
     }
 
+    /**
+     * @param string $path
+     *
+     * @return Tag|null
+     */
     public static function getByPath($path) {
-        return (new self)->getDao()->getByPath($path);
+        try {
+            return (new self)->getDao()->getByPath($path);
+        } catch(\Exception $e) {
+            return null;
+        }
     }
 
     public function save()
--- a/models/Element/Tag.php
+++ b/models/Element/Tag.php
@@ -280,7 +280,7 @@ class Tag extends Model\AbstractModel
 
         $parentNames = array_reverse($parentNames);
 
-        return '/' . implode('/', $parentNames) . '/';
+        return '/' . implode('/', $parentNames);
     }
 
     public function __toString()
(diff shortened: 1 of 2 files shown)


models/Element/Tag.php
```diff
@@ -160,6 +160,10 @@ public static function getElementsForTag(
         return $tag->getDao()->getElementsForTag($tag, $type, $subtypes, $classNames, $considerChildTags);
     }
 
+    public static function getByPath($path) {
```
dvesh3, line 163: please rename it to `getByNamePath` since it works only with name path.

models/Element/Tag/Dao.php
```diff
@@ -292,4 +292,38 @@ public function getElementsForTag(
...
+    /**
+     * @param string $tagPath separated by "/"
+     *
+     * @return null|Tag
+     */
+    public function getByPath($tagPath) {
+        $parentTagId = 0;
+
+        $tag = null;
+        foreach (explode('/', $tagPath) as $tagItem) {
```
dvesh3, line 305: It should work with same value return by `getNamePath()` which is with leading slash e.g. `"/key1/key2/key2"`

models/Element/Tag.php
BlackbitNeueMedien: Wanted to use the same name as for data objects, assets etc. There it is also named `getByPath()` and not `getByFilenamePath()` for assets and `getByKeyPath()` for data objects.
Please confirm that I really should use a different name here.

models/Element/Tag.php
```diff
@@ -160,6 +160,10 @@ public static function getElementsForTag(
...
+        return (new self)->getDao()->getByPath($path);
```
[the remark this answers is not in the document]
BlackbitNeueMedien: Fixed.

models/Element/Tag/Dao.php
BlackbitNeueMedien: Fixed.

models/Element/Tag.php
dvesh3: Ok, let's keep the name `getByPath()` for consistency. At least, add doc stating that it works only with name path. thanks!

BlackbitNeueMedien: Fixed.

PR #6965, opened 2020-08-18
[Objects] Quantity Value: Do not store Unit IDs in class definitions

### LINKED ISSUE #1023: [Objects] Quantity Value: Do not store Unit IDs in class definitions

weisswurstkanone:
resolves #1023 

see comment https://github.com/pimcore/pimcore/issues/1023#issuecomment-662382202

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
@@ -234,7 +234,7 @@ pimcore.object.quantityValue.unitsettings = Class.create({
     },
 
     onAdd: function (btn, ev) {
-        Ext.MessageBox.prompt(' ', t('enter_the_name_of_the_new_item'),
+        Ext.MessageBox.prompt(' ', t('unique_identifier'),
             function (button, value, object) {
                 var regresult = value.match(/[a-zA-Z0-9_\-]+/);
                 if (button == "ok") {
@@ -282,4 +282,3 @@ pimcore.object.quantityValue.unitsettings = Class.create({
         this.grid.store.remove(rec);
     }
 });
-
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
@@ -63,9 +63,14 @@ pimcore.object.quantityValue.unitsettings = Class.create({
     getRowEditor: function () {
 
         var baseUnitStore = Ext.create('Ext.data.JsonStore', {
+            fields: [{
+                name: 'id',
+                type: 'string'
+            }, 'abbreviation', 'longname', 'group', 'baseunit', 'factor', 'conversionOffset', 'reference', 'converter'],
             proxy: {
                 type: 'ajax',
-                async: false,
+                async: true,
+                batchActions: false,
                 url: Routing.generate('pimcore_admin_dataobject_quantityvalue_unitproxyget'),
                 reader: {
                     type: 'json',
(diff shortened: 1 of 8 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
```diff
@@ -229,14 +234,40 @@ pimcore.object.quantityValue.unitsettings = Class.create({
...
+                            this.cellEditing.startEditByPosition({
+                                row: 0,
+                                column: 0
+                            });
+
+                        }.bind(this)
+                    });
+
+                } else {
+                    Ext.Msg.alert(' ', t('failed_to_create_new_item'));
```
dvesh3, line 266: As a user, I do not expect this message when clicking cancel or closing the dialog box.

bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
```diff
@@ -229,14 +234,40 @@ pimcore.object.quantityValue.unitsettings = Class.create({
...
     onAdd: function (btn, ev) {
-        var u = {};
-        //id = -1;
-        this.cellEditing.completeEdit();
-        this.grid.store.insert(0, u);
-        this.cellEditing.startEditByPosition({
-            row: 0,
-            column: 0
-        });
+        Ext.MessageBox.prompt(' ', t('enter_the_name_of_the_new_item'),
```
dvesh3, line 237: title missing

models/DataObject/QuantityValue/Unit/Dao.php
```diff
@@ -89,24 +90,35 @@ public function getById($id)
...
+            if ($id) {
+                $id++;
+            } else {
+                $id = "1";
+            }
+        }
+
+
+        $this->model->setId($id);
+
```
dvesh3, line 121:
```suggestion
        if (!$id = $this->model->getId()) {
            // mimic autoincrement
            $id = $this->db->fetchOne('select CONVERT(SUBSTRING_INDEX(id,\'-\',-1),UNSIGNED INTEGER) AS num FROM quantityvalue_units ORDER BY num DESC LIMIT 1 ');
            $id = $id > 0 ? ($id + 1) : 1;
            $this->model->setId($id);
        }

```

bundles/CoreBundle/Migrations/Version20200807105448.php
```diff
@@ -0,0 +1,65 @@
...
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+
+class Version20200807105448 extends AbstractPimcoreMigration
+{
+    public function doesSqlMigrations(): bool
+    {
+        return false;
```
dvesh3, line 12:
```suggestion
        return true;
```

models/DataObject/QuantityValue/Unit/Dao.php
```diff
@@ -17,6 +17,7 @@
 
 namespace Pimcore\Model\DataObject\QuantityValue\Unit;
 
+use Pimcore\Db;
```
dvesh3, line 20: not required as `AbstractDao` already instantiates `$this->db`

bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
weisswurstkanone: moved the messagebox into the validation block

bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
weisswurstkanone: pls search for this line of code. we do it exactly that way at dozens of other places. maybe a refactoring / general improvement task ?

dvesh3: well anyway this is misleading, as it asks for "name" which is actually an Id.

dvesh3: Also, it would be nice to set default auto incremented Id as value so user doesn't need to take care of it everytime.

bundles/AdminBundle/Resources/public/js/pimcore/object/quantityvalue/unitsettings.js
```diff
@@ -229,14 +234,43 @@ pimcore.object.quantityValue.unitsettings = Class.create({
...
```
dvesh3, line 237:
```suggestion
        Ext.MessageBox.prompt(' ', t('unique_identifier'),
```

PR #6971, opened 2020-08-20
[Tags] Tag names should be unique in the same hierarchy level

### LINKED ISSUE #6970: [Tags] Tag names should be unique in the same hierarchy level
At the moment it's possible to create the same tag on the same level, what doesn't make sense at all. 
Example: 

- foo
  - baz
  - baz
- bar
- foo

Solution: create a unique compound db index over `idPath` and `name`

Follow up to #6964

dvesh3:
Resolves #6970

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20200820151104.php
+++ b/bundles/CoreBundle/Migrations/Version20200820151104.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+
+class Version20200820151104 extends AbstractPimcoreMigration
+{
+    public function doesSqlMigrations(): bool
+    {
+        return true;
+    }
+
+    /**
+     * @param Schema $schema
+     */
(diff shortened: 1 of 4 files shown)


bundles/CoreBundle/Migrations/Version20200820151104.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        try {
+            $this->addSql('ALTER TABLE `tags` CHANGE COLUMN `name` `name` varchar(255) DEFAULT NULL COLLATE utf8_bin AFTER `idPath`');
+            $this->addSql('ALTER TABLE `tags` ADD UNIQUE INDEX `idPath_name` (`idPath`,`name`)');
```
NiklasBr, line 22: Some sort of migration of duplicate tags must happen here, right?

brusch: @dvesh3 I agree with @NiklasBr - we should rename duplicate tags in the migration to ensure seamless upgrades.

dvesh3: Added migration for renaming duplicate tags with console warnings.

PR #6989, opened 2020-08-25
Undefined index disable_stack_trace notice in saveVersion

weisswurstkanone:


dvesh3:
@weisswurstkanone default set from config but yeah it doesn't harm..
https://github.com/pimcore/pimcore/blob/90f8fb355d590b4b2958181c82f3f7f5a0c88c17/bundles/CoreBundle/DependencyInjection/Configuration.php#L581-L588

dvesh3:
@weisswurstkanone thanks! 🙏

PR #6990, opened 2020-08-25
Fix doc block return types for relations

belendel:
By changing Data::getPhpdocType() to Data::getPhpdocReturnType() the call to Relation::getPhpDocClassString() for relationship data types was lost. For example in ManyToOneRelation::getPhpdocType()

This in turn causes the IDE's to break and is causing issues with symfony serializer which uses doc block for typehinting.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/ManyToOneRelation.php
+++ b/models/DataObject/ClassDefinition/Data/ManyToOneRelation.php
@@ -31,7 +31,6 @@ class ManyToOneRelation extends AbstractRelations implements QueryResourcePersis
     use DataObject\ClassDefinition\Data\Relations\AllowObjectRelationTrait;
     use DataObject\ClassDefinition\Data\Relations\AllowAssetRelationTrait;
     use DataObject\ClassDefinition\Data\Relations\AllowDocumentRelationTrait;
-    use DataObject\ClassDefinition\NullablePhpdocReturnTypeTrait;
 
     /**
      * Static type of this element
@@ -773,6 +772,18 @@ class ManyToOneRelation extends AbstractRelations implements QueryResourcePersis
             return $listing;
         }
         throw new \InvalidArgumentException('Filtering '.__CLASS__.' does only support "=" operator');
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getPhpdocReturnType(): ?string
+    {
+        if ($this->getPhpdocType()) {
+            return $this->getPhpdocType() . '|null';
+        }
+
+        return null;
     }
 }
 


weisswurstkanone:
I think it is better to fix this in AbstractRelations and return the trait there

belendel:
Hi Josef,

Thanks for your reply. I'm not quite sure I get you. Do you mean remove the trait from the relationship data types and move it to AbstractRelations and override the getPhpdocReturnType() method there?

Also it seems getPhpdocReturnType() is called without checking if the TypeDeclarationSupportInterface is implemented.

weisswurstkanone:
Yes exactly.

Regarding the check: Do you mean in the trait? Well, this shouldn't be an issue. It is marked as internal and all internal data types implement the TypeDeclarationInterface.

belendel:
https://github.com/pimcore/pimcore/blob/dee82b6069915b0b1a42913cfd70c74ec37254cd/models/DataObject/ClassDefinition/Data.php#L730

I meant in Data::getGetterCode() for example. It doesn't check if the current instance implements Pimcore\Model\DataObject\ClassDefinition\Data\TypeDeclarationSupportInterface. 

Not sure what the fallback should be though. Otherwise it would be better to implement the interface on Data class which would enforce it.

belendel:
Ooh sorry I see you have the methods on Data class already. Just not the interface.

weisswurstkanone:
But there is a default implementation in line 700, so it should always be there, right?

belendel:
Yea I missed it. Sorry about that! I will update the pull request.

belendel:
The NullablePhpdocReturnTypeTrait was only added to ManyToOneRelation. So moving it to AbstractRelations could break things for the other relation data types.

weisswurstkanone:
Why not move it to AbstractRelations and overwrite whatever code is needed in the subclasses ?

belendel:
Hi Josef,

ManyToOneRelation is the only subclass of AbstractRelations that uses NullablePhpdocReturnTypeTrait. All the others uses the default methods from Data.

By moving the trait to AbstractRelations I would have to override all the subclasses again with the same behaviour that's in Data. I think this is cleaner?

weisswurstkanone:
As said, we are not forced to use the trait everywhere. The intention is to avoid code duplication, nothing else. If it doesn't fit there then just remove it and implement the method as needed.

weisswurstkanone:
Sorry, I think I cannot follow you.

A many-to-one on the master branch generates the following code: What is the issue with it?

```
/**
* Get sub - sub
* @return \Pimcore\Model\Document\Page | \Pimcore\Model\Document\Snippet | \Pimcore\Model\Document | \Pimcore\Model\Asset | \Pimcore\Model\DataObject\AbstractObject|null
*/
public function getSub (): ?\Pimcore\Model\Element\AbstractElement {
	if($this instanceof PreGetValueHookInterface && !\Pimcore::inAdmin()) { 
		$preValue = $this->preGetValue("sub"); 
		if($preValue !== null) { 
			return $preValue;
		}
	} 

	$data = $this->getClass()->getFieldDefinition("sub")->preGetData($this);

		    return $data->getPlain();
	}

	return $data;
}
```

belendel:
Hi Josef,

The issue is that before it generated the following code (copied from one of our projects):

```php

/**
* Get Serie - Serie
* @return \Pimcore\Model\DataObject\Serie
*/
public function getSerie () {
        if($this instanceof PreGetValueHookInterface && !\Pimcore::inAdmin()) { 
                $preValue = $this->preGetValue("Serie"); 
                if($preValue !== null) { 
                        return $preValue;
                }
        } 
​
        $data = $this->getClass()->getFieldDefinition("Serie")->preGetData($this);
​
                    return $data->getPlain();
        }
​
        return $data;
}
``

This was done in Pimcore\Model\DataObject\ClassDefinition\Data\Extension\Relation::getPhpDocClassString(). This was called from Pimcore\Model\DataObject\ClassDefinition\Data\ManyToOneRelation::getPhpdocType() which is now skipped because of Pimcore\Model\DataObject\ClassDefinition\NullablePhpdocReturnTypeTrait::getPhpdocReturnType();

The missing class name breaks typehinting and autocomplete in IDE and break symfony serializer which also uses docblock for typehinting.

dvesh3:
Hi @belendel @weisswurstkanone I tried to understand the issue.

It seems like the current approach does not consider restricted return types in doc block e.g. when I select "Car" in Allowed Objects then return type should be :
Correct
image

actually it is:
Wrong
image

So this PR fixes this issue by removing usage of `NullablePhpdocReturnTypeTrait` which is using default [$phpdocType property]( https://github.com/pimcore/pimcore/blob/01e71d5be10f95582a16e24d3d2d4aedd7fe7670/models/DataObject/ClassDefinition/Data/ManyToOneRelation.php#L73) for doc block
https://github.com/pimcore/pimcore/blob/dee82b6069915b0b1a42913cfd70c74ec37254cd/models/DataObject/ClassDefinition/NullablePhpdocReturnTypeTrait.php#L30

@belendel please correct me, if i'm wrong :)

belendel:
@dvesh3 that's exactly it

dvesh3:
@belendel since this is a fix, please rebase it to 6.7 branch.

brusch:
@belendel ping 🏓

brusch:
Fixed by d3dc72613aee7334325e6ea7db9331232e821bc5

PR #6947, opened 2020-08-13
[DataObject] Fix deleting folders

dvesh3:
## Changes in this pull request  
Resolves #6945

Issue #6945, opened 2020-08-13
Can not delete folders in version 6.7.0


rubanooo:
# Bug Report

I updated PimCore to the newest version 6.7.0 today and tried to delete a folder. This issue is reproducable on the demo site.

### Actual behavior

Deleting a folder gives you the following error:
`Call to undefined method getClassName in class Pimcore\Model\DataObject\Folder`

### Steps to reproduce

Create a folder or select an existing folder and try to delete it by opening the right mouse button menu.

### Trace

```json
 [{
 	"file": "\/var\/www\/pimcore\/vendor\/pimcore\/pimcore\/models\/DataObject\/AbstractObject.php",
 	"line": 1350,
 	"function": "__call",
 	"class": "Pimcore\\Model\\AbstractModel",
 	"type": "-\u003E",
 	"args": ["getClassName", []]
 }, {
 	"file": "\/var\/www\/pimcore\/vendor\/pimcore\/pimcore\/models\/DataObject\/AbstractObject.php",
 	"line": 361,
 	"function": "__call",
 	"class": "Pimcore\\Model\\DataObject\\AbstractObject",
 	"type": "-\u003E",
 	"args": ["getClassName", []]
 }, {
 	"file": "\/var\/www\/pimcore\/vendor\/pimcore\/pimcore\/bundles\/AdminBundle\/Controller\/Admin\/ElementControllerBase.php",
 	"line": 160,
 	"function": "getList",
 	"class": "Pimcore\\Model\\DataObject\\AbstractObject",
 	"type": "::",
 	"args": [{
 		"unpublished": true
 	}]
 }, {
 	"file": "\/var\/www\/pimcore\/vendor\/pimcore\/pimcore\/bundles\/AdminBundle\/Controller\/Admin\/DataObject\/DataObjectController.php",
 	"line": 92,
 	"function": "deleteInfoAction",
 	"class": "Pimcore\\Bundle\\AdminBundle\\Controller\\Admin\\ElementControllerBase",
 	"type": "-\u003E",
 	"args": [{
 		"attributes": {},
 		"request": {},
 		"query": {},
 		"server": {},
 		"files": {},
 		"cookies": {},
 		"headers": {}
 	}]
 }, {
 	"file": "\/var\/www\/pimcore\/vendor\/symfony\/symfony\/src\/Symfony\/Component\/HttpKernel\/HttpKernel.php",
 	"line": 158,
 	"function": "deleteInfoAction",
[...]

PR #6993, opened 2020-08-25
[DataObject] Fix deleting folders with class override

ssiergl:
Regression from #6947 and #6945

as discussed external bundles with class override on DataObject\Folder class will still break.

Fix it with explicity checking if class is instanceof Concrete before constructing DataObject class and listing


ssiergl:
@dvesh3 as discussed

should be scheduled for 6.7.2 as this is a bug fix?

dvesh3:
@ssiergl yes, scheduled for 6.7.2. thank you!

PR #6997, opened 2020-08-25
[Admin] DataObjects - Disable bool meta column for non-editable relations

### LINKED ISSUE #6994: Advanced many-to-many object relation - not editable check not working on added columns ( only Bool type)
# Bug Report

### Expected behavior
All types of columns must be non-editable.
### Actual behavior
Being non-editable does not work for columns of type Bool.
### Steps to reproduce
- Create a class with an "Advanced many-to-many object relation" type field, containing a "Bool" column type.
- Add some object to the relation.
- Set the "not editable" check true inside the field configuartion.
- Modify the "Bool" column for objects added earlier in the report.

# Feature Request
- Make each type of column non-editable when the "not editable" check is set to true.

dvesh3:
Resolves #6994


weisswurstkanone:
thx!

PR #6998, opened 2020-08-26
TrackingCodeFlashMessageListener: Avoid autostart session by accessing the FlashBag

das-peter:
With session autostart enabled accessing the FlashBag on the session container will create a session.
I think it would be beneficial to start sessions only if necessary.
Sessions make upstream caching e.g. with Varnish hard.
Explicitly checking if a session has been started before attempting to read if there are any data in the session shouldn't hurt the functionality in any way.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/EventListener/Frontend/TrackingCodeFlashMessageListener.php
+++ b/bundles/EcommerceFrameworkBundle/EventListener/Frontend/TrackingCodeFlashMessageListener.php
@@ -65,13 +65,16 @@ class TrackingCodeFlashMessageListener implements EventSubscriberInterface
             return;
         }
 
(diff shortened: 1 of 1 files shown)


brusch:
@das-peter can you please resolve the conflict? 
Thanks a lot! 😊

dvesh3:
@das-peter This PR changes will break the tracking codes functionality, as session is not started yet and with the new check tracking codes will not be processed. 

I have created a PR #7135 with different approach: introducing a cookie that we use to check if there's a flash bag, we only start the session if the cookie is there and also disable the full page cache for those requests.

Please have a look and let me know what do you think?

brusch:
Closing this PR in favor of #7135
Thanks anyway for pointing out the issue!

PR #7000, opened 2020-08-26
[DataObject] Fix block inside localized field

### LINKED ISSUE #6976: block inside localized field
# Bug Report
I am currently having troubles when I try to use a block inside a localized field. You can see the class structure in the attached 
image 
Class

### Actual behavior
If I am creating a new blockitem inside the localized field and save it everything seems to work but as soon as I reload pimcore and try to save the Object again the backend instantly goes into maintenance mode.

### Steps to reproduce
create class -> create object -> add block item -> save -> reload pimcore -> try to save again

dvesh3:
Resolves #6976

### DIFF
--- a/models/DataObject/Data/BlockElement.php
+++ b/models/DataObject/Data/BlockElement.php
@@ -177,26 +177,7 @@ class BlockElement extends AbstractModel implements OwnerAwareFieldInterface
      */
     public function __sleep()
     {
-        $copier = new DeepCopy();
-        $copier->skipUncloneable(true);
-        $copier->addTypeFilter(
-            new \DeepCopy\TypeFilter\ReplaceFilter(
-                function ($currentValue) {
-                    if ($currentValue instanceof ElementInterface) {
-                        $elementType = Service::getType($currentValue);
-                        $descriptor = new ElementDescriptor($elementType, $currentValue->getId());
-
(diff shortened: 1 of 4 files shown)


weisswurstkanone (approved):

weisswurstkanone (changes requested):
unit tests

dvesh3:
> unit tests

@weisswurstkanone reverted

weisswurstkanone:
@dvesh3  still failing

models/DataObject/Data/BlockElement.php
```diff
@@ -150,7 +148,17 @@ function ($currentValue) {
...
+                /**
+                 * @param mixed $element
+                 *
+                 * @return bool
+                 */
+                public function matches($element)
+                {
+                    return $element instanceof ElementInterface;
+                }
+            }
```
jdreesen, line 161: Why do you need an anonymous class here? Shouldn't this be sufficient?

```suggestion
            new TypeMatcher(ElementInterface::class)
```

weisswurstkanone:
We will come up with a generic approach.

Discussion points:
* https://github.com/pimcore/pimcore/pull/5799
* https://github.com/pimcore/pimcore/pull/6236

PR #7001, opened 2020-08-26
PageSippet: set/remove editable loses existing elements on save

### LINKED ISSUE #6992: PageSippet: set/remove editable loses existing elements on save
Steps to reproduce:

```
$doc = Document::getById(99);
$doc->setEditable("somekey", "somedata");    // it doesn't check if existing editables are already loaded, so overwrites them ...
$doc->save();
```

see also https://github.com/pimcore/pimcore/pull/6982

weisswurstkanone:
resolves #6992

### DIFF
--- a/models/Document/PageSnippet.php
+++ b/models/Document/PageSnippet.php
@@ -420,6 +420,7 @@ abstract class PageSnippet extends Model\Document
      */
     public function setEditable(string $name, Tag $data)
     {
+        $this->getEditables();
         $this->editables[$name] = $data;
 
         return $this;
@@ -444,6 +445,7 @@ abstract class PageSnippet extends Model\Document
      */
     public function removeEditable(string $name)
     {
+        $this->getEditables();
         if (isset($this->editables[$name])) {
             unset($this->editables[$name]);
         }
@@ -604,6 +606,7 @@ abstract class PageSnippet extends Model\Document
     public function getEditables(): array
     {
         if ($this->editables === null) {
+            $this->editables = [];
             $this->setEditables($this->getDao()->getEditables());
         }
 
--- a/models/Document/PageSnippet.php
+++ b/models/Document/PageSnippet.php
@@ -606,7 +606,6 @@ abstract class PageSnippet extends Model\Document
     public function getEditables(): array
     {
         if ($this->editables === null) {
-            $this->editables = [];
             $this->setEditables($this->getDao()->getEditables());
         }
 
--- a/models/Document/PageSnippet/Dao.php
+++ b/models/Document/PageSnippet/Dao.php
@@ -54,8 +54,9 @@ abstract class Dao extends Model\Document\Dao
             $editable->setDataFromResource($editableRaw['data']);
 
             $editables[$editableRaw['name']] = $editable;
-            $this->model->setEditable($editableRaw['name'], $editable);
         }
+
+        $this->model->setEditables($editables);
 
         return $editables;
     }


models/Document/PageSnippet.php
```diff
@@ -604,6 +606,7 @@ public function getElements()
     public function getEditables(): array
     {
         if ($this->editables === null) {
+            $this->editables = [];
```
brusch, line 609: This shouldn't be necessary since `$this->getDao()->getEditables()` always returns an array, right?

brusch (changes requested):

models/Document/PageSnippet.php
weisswurstkanone: Yes, but since $this->editables is not initialized (i.e. null) and getEditables calls setEditable this will result in an endless recursion otherwise. See https://github.com/pimcore/pimcore/pull/6991#issue-473165647

jdreesen: No, it doesn't call `setEditable` but `setEditables` which just sets the field. I don't see why this should result in an endless loop.

brusch: @weisswurstkanone is right, with this PR `setEditable()` is again calling `getEditables()` which would result in an endless loop.

brusch: But I'm wondering why this is happening, ... it should be sufficient if `\Pimcore\Model\Document\PageSnippet\Dao::getEditables()` just returns the editables ... calling `setEditable()` first and then calling `PageSnippet::setEditables()` with the return value of `Dao::getEditables()` sounds a bit weird to me.

weisswurstkanone:
image

models/Document/PageSnippet.php
msphn: why not initialise editables as an empty array? otherwise the strict type change will fail here anyway

weisswurstkanone:
$this->editable === null currently indicates that the editabes haven't been loaded and this still needs to be done.
I can think of a dedicated flag instead ....

models/Document/PageSnippet/Dao.php
```diff
@@ -54,9 +54,10 @@ public function getEditables()
             $editable->setDataFromResource($editableRaw['data']);
 
             $editables[$editableRaw['name']] = $editable;
-            $this->model->setEditable($editableRaw['name'], $editable);
         }
 
+        $this->model->setEditables($editables);
```
brusch, line 59: I guess this can go as well ... since the caller uses the return value anyway: 
https://github.com/pimcore/pimcore/blob/338cd24e38090c8a13e0580228aa32b62f5bc860/models/Document/PageSnippet.php#L608

jdreesen: In line 77/78 `$this->model->setVersions($versions);` should be removed, too, as it's the same.

jdreesen: But actually... this pattern of setting the value on the from from inside the DAO is used in most of the DAO implementations... so maybe it should still be done - consistency wise - even if it's redundant...

brusch: But then we should remove it from the other place, ... calling the same thing twice for nothing doesn't make sense 😉

jdreesen: That's right. But maybe in a follow-up PR as it's not really related to this fix and might be a bigger changeset.

brusch: you're right, this doesn't belong to this PR

PR #7004, opened 2020-08-26
Avoid function aliases

jdreesen:
Some functions in PHP have aliases. For consistency in the codebase, these should be avoided.


brusch:
@jdreesen well done, thanks! But one thing, since it's actually not a bug fix it should be merged to the master, can you please change the base? Thanks again!

brusch:
Grazie Mille! 👌

PR #7007, opened 2020-08-26
boolean select csv import and export not working

### LINKED ISSUE #7003: boolean select csv import and export not working
# Bug Report

### Expected behavior
When CSV exporting a boolean select which is set to "No" or "empty" corresponding values should be exported.
It should be possible to import "No" and "empty" values via CSV.

### Actual behavior
"No" and "empty" attributes leave an empty cell in a CSV export.
"No" and "empty" cannot be imported via CSV import.

### Steps to reproduce
1. Add a boolean select field to a class.
2. In an object set its value to "no" or "empty"
3. Export a CSV and look up the corresponding cell which will be empty

weisswurstkanone:
resolves #7003

### DIFF
--- a/models/DataObject/ClassDefinition/Data/BooleanSelect.php
+++ b/models/DataObject/ClassDefinition/Data/BooleanSelect.php
@@ -476,6 +476,6 @@ class BooleanSelect extends Data implements ResourcePersistenceAwareInterface, Q
         } else {
             $value = null;
         }
-        return $importValue;
+        return $value;
     }
 }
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/BooleanSelect.php
```diff
@@ -435,4 +435,47 @@ public function isEqual($oldValue, $newValue): bool
...
+     */
+    public function getFromCsvImport($importValue, $object = null, $params = [])
+    {
+        if ($importValue === "1") {
+            $value = self::YES_VALUE;
+        } else if ($importValue === "0") {
+            $value = self::NO_VALUE;
+        } else {
+            $value = null;
+        }
```
dvesh3, line 478:
```suggestion
        if ($importValue === "1") {
            $value = true;
        } else if ($importValue === "0") {
            $value = false;
        } else {
            $value = null;
        }
```
since we need bool here https://github.com/pimcore/pimcore/blob/01e71d5be10f95582a16e24d3d2d4aedd7fe7670/models/DataObject/ClassDefinition/Data/BooleanSelect.php#L95

PR #7011, opened 2020-08-27
[Docs] Display all code examples with two languages (php & twig) in tabs

jdreesen:
In some places, the code examples, when given in Twig & PHP, are already displayed in tabs, like here: 
https://pimcore.com/docs/6.x/Development_Documentation/Documents/Editables/Block.html#page_Using-Manual-Mode

But in other places, they're just displayed below each other, like here: 
https://pimcore.com/docs/6.x/Development_Documentation/Documents/Editables/Video.html#page_Basic-Usage-a-Local-Asset

This PR tries to align it, so that all examples existing in both languages are displayed in tabs.

### DIFF
--- a/doc/Development_Documentation/02_MVC/02_Template/01_Layouts.md
+++ b/doc/Development_Documentation/02_MVC/02_Template/01_Layouts.md
@@ -21,6 +21,8 @@ For more details about template inheritance and layouts, please have a look at t
 
 ###### A Simple Sample Layout Looks Like the Following:  
 
+<div class="code-section">
+
 ```php
 <?php
 /**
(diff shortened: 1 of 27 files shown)


aarongerig:
Just as a quick question or follow up to this topic. Would it also be possible to save the user's preference (twig or php) to the local storage of the browser, so that you don't always have to change tabs on other documentation pages? I think that's something that could further improve the docs. :)

jdreesen:
I thought about that, too, it would be really nice!

But as there may be tabbed code examples in other languages, too (I'm thinking of e.g. the Apache & nginx snippets [here](https://pimcore.com/docs/6.x/Development_Documentation/Assets/Restricting_Public_Asset_Access.html)) I'm not sure how to implement this at the moment.

jdreesen:
It would be a follow-up ticket anyway, as this must be changed in the https://github.com/pimcore/pimcore-docs repo.

Maybe we can start by just saving the preference for tabs with only Twig & PHP code examples...

brusch:
Just thinking if it's worth the effort ... PHP will go anyway in Pimcore 7 😉
But what we can already think about is changing the default to Twig.

dvesh3:
@jdreesen this is an improvement, so should go into master. could you please rebase it? thanks!

jdreesen:
Done

dvesh3:
thnx! 🙏

PR #7012, opened 2020-08-27
[Docs] Update 03_Navigation.md

luisaugusto:
Updated the [breadcrumbs section](https://github.com/pimcore/pimcore/blob/master/doc/Development_Documentation/03_Documents/03_Navigation.md#breadcrumbs) to include a Twig example that accompanies the PHP example.


brusch:
Thanks! 👌

PR #7014, opened 2020-08-28
Migrates tables from tinyint(4) to tinyint(1) for boolean-type content

NiklasBr:
## Additional info  

I prepared this PR, and then I remembered that it had [no effect on storage](https://dev.mysql.com/doc/refman/5.7/en/integer-types.html)/memory or anything like that, it's just a display value as long as it is a `tinyint`, oh well, do what you like with it, at least now it's consistent for all boolean-type columns :)

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20200828091305.php
+++ b/bundles/CoreBundle/Migrations/Version20200828091305.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace App\Migrations;
+namespace Pimcore\Bundle\CoreBundle\Migrations;
 
 use Doctrine\DBAL\ConnectionException;
 use Doctrine\DBAL\DBALException;
--- a/bundles/CoreBundle/Migrations/Version20200828091305.php
+++ b/bundles/CoreBundle/Migrations/Version20200828091305.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace App\Migrations;
+
+use Doctrine\DBAL\ConnectionException;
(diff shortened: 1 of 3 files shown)


brusch:
@NiklasBr  Yes, I think it would make sense to have it consistent across the system. Thanks a lot!

bundles/CoreBundle/Migrations/Version20200828091305.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+namespace App\Migrations;
```
dvesh3, line 3:
```suggestion
namespace Pimcore\Bundle\CoreBundle\Migrations;
```

NiklasBr: Thanks!

bundles/CoreBundle/Migrations/Version20200828091305.php
```diff
@@ -0,0 +1,69 @@
...
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Log\Handler\ApplicationLoggerDb;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+
+/**
+ * Class Version20200828091305
+ *
+ * Migrates tables from tinyint(4) to tinyint(1) for boolean-type content
+ *
+ * @package App\Migrations
```
jdreesen, line 16:
```suggestion
```

dvesh3:
thank you!

PR #7018, opened 2020-08-28
Symfony templates location

NiklasBr:
## Additional info  

[Using Symfony's location](https://symfony.com/doc/current/templates.html#template-location) `templates` for templates already work fine, but they never show up when selecting a template for a document, this resolves that problem.

### DIFF
--- a/lib/Controller/Config/ControllerDataProvider.php
+++ b/lib/Controller/Config/ControllerDataProvider.php
@@ -233,7 +233,7 @@ class ControllerDataProvider
         }
 
         $symfonyPath = realpath(implode(DIRECTORY_SEPARATOR, [PIMCORE_PROJECT_ROOT, 'templates']));
-        if ($symfonyPath && file_exists($symfonyPath) && is_dir($symfonyPath)) {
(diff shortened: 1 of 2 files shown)


lib/Controller/Config/ControllerDataProvider.php
```diff
@@ -232,6 +232,11 @@ public function getTemplates(): array
             $templates = array_merge($templates, $this->findTemplates($appPath));
         }
 
+        $symfonyPath = realpath(implode(DIRECTORY_SEPARATOR, [PIMCORE_PROJECT_ROOT, 'templates']));
+        if ($symfonyPath && file_exists($symfonyPath) && is_dir($symfonyPath)) {
```
jdreesen, line 236: I doubt that checking for both, `file_exists` & `is_dir` is really necessary.

```suggestion
        if ($symfonyPath && is_dir($symfonyPath)) {
```

NiklasBr: That's correct, `is_dir(null)` will return a `false` boolean. I just copied the above and below checks, so I'll change those as well.

dvesh3:
thanks a lot!

PR #7019, opened 2020-08-29
[Feature] Allows to pretty print ClassDefinitions

benammann:
Sometimes it happens that you have to merge the class definitions. But the native format of `var_export` causes some problems here.

The library https://github.com/brick/varexporter makes the created PHP code much more readable and mergeable.

This feature can be activated with the environment variable `PIMCORE_CLASS_DEFINITION_PRETTY_PRINT=true`.

Once enabled the class definitions are created via the `brick/varexporter` library.


brusch:
Nice, thanks a lot for your PR! 
Just 2 more things: 
- I don't think we need an option to turn that on (if it is necessary for whatever reason, it should be a Symfony Config instead - not a constant) 
- If we use `VarExporter` instead of `var_export` we should take the chance to check the export/loading in a unit test, just to be sure it works for all data-types

Thanks!

composer.json
```diff
@@ -95,7 +95,8 @@
     "scheb/two-factor-bundle": "^3.26",
     "phpoffice/phpspreadsheet": "^1.9",
     "html2text/html2text": "^4.2.1",
-    "webmozarts/console-parallelization": "^1.1.0"
+    "webmozarts/console-parallelization": "^1.1.0",
+    "brick/varexporter": "^0.3.2"
```
jdreesen, line 99: This should be an optional dependency, I think, since it's only used in one place and on demand.

brusch:
Just a note for later reference: If the 2 points mentioned above are getting fixed and this PR get's merged, we should replace all usages of `var_export_pretty()` with `VarExporter` in a follow up PR.

benammann:
Hi @brusch @jdreesen,

thanks for your feedback!

As soon as I have time again I will take care of the points and update the pull request.

brusch:
@benammann any updates on this?

brusch:
Closed due to inactivity. 
Please feel free to come back on this PR as soon as you have time again 😊
Thanks and happy new year! 🎉

PR #7020, opened 2020-08-31
Fix wrong PHPDocs in \Pimcore\Model\Document\Tag\Image

blankse:


models/Document/Tag/Image.php
```diff
@@ -421,7 +423,7 @@ public function getImage()
     }
 
     /**
-     * @param Asset\Image $image
+     * @param Asset\Image|null $image
```
NiklasBr, line 426: `$this` cannot be `null` right?

blankse: The image param can be null. To reset the tag. See:
https://github.com/pimcore/pimcore/pull/7020/files#diff-3bbec05fe0a61dca677cb7e8c81411dfR803

NiklasBr: But the method returns `$this`

blankse: Yes `$this` cannot be null. And the return PHPDoc is correct, right?

NiklasBr: says `@return Model\Document\Tag\Image|null` now

jdreesen: No, it doesn't @blankse only changed the `@param` annotation, not the `@return` one.

NiklasBr: image

Incognito says I was wrong, might have seen at an old commit or something. Sorry!

brusch:
Thanks!

PR #7021, opened 2020-08-31
Correct the call to get the correct img thumbnail in image tags in backend/ edit mode

roland4432:
Fixed parameters passed to this.getThumbnailConfig

  
  

The function merging parameters to create the img.src is being called with wrong parameters, hence I corrected it

## Additional info  
none - easy fix


weisswurstkanone:
Could you please rebase it to 6.7 ? Thx!

jdreesen (approved):
This broke in #6778

roland4432:
Rebased to 6.7

jdreesen:
You need to change the target branch of the PR as well.

jdreesen:
And only keep the original commit (e0f75e9) when rebasing, skipping everything else.

roland4432:
Close this one and opened #7022 since rebasing again would duplicate commits

jdreesen:
No, it wouldn't (if done right). You have to do an interactive rebase, skipping everything not related to the actual fix ;)

roland4432:
> No, it wouldn't (if done right). You have to do an interactive rebase, skipping everything not related to the actual fix ;)

Agreed, but not if the branched is already rebased (which it was). I would need to create a new branch then anways... which I could already base on pimcore/6.7 when checking out...

jdreesen:
> I would need to create a new branch then anways...

No, it's still possible with rebasing the branch (almost anything is possible in git :D). 

Anyways, creating a new branch based on the desired target branch and cherry-picking the commit may be faster/easier in this case. Except that you have to create a new PR on GitHub then...

PR #7024, opened 2020-09-01
video image-thumbnails should be delivered web-optimized

AlternateIf:
image thumbnails for videos should be created the same way ordinary image thumbnails are. 

### **Expected behavior**

Thumbnails will be generated according to the thumbnail config format 
image

### **Actual behavior**

Thumbnails are created with the png format in case of "source"-format (=Auto). See [Video-ImageThumbnail](https://github.com/pimcore/pimcore/blame/master/models/Asset/Video/ImageThumbnail.php#L199)

This PR changes the behaviour to match the image - thumbnail logic (see [Image-Thumbnail](https://github.com/pimcore/pimcore/blob/master/models/Asset/Image/Thumbnail.php#L467))

### DIFF
--- a/models/Asset/Video/ImageThumbnail.php
+++ b/models/Asset/Video/ImageThumbnail.php
@@ -192,14 +192,6 @@ class ImageThumbnail
      */
     protected function createConfig($selector)
     {
-        $config = Image\Thumbnail\Config::getByAutoDetect($selector);
-        if ($config) {
-            $format = strtolower($config->getFormat());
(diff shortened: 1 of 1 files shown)


dvesh3 (approved):
format 'source' to 'png' conversion already happens here: https://github.com/pimcore/pimcore/blob/2662d44a8828ad8fc7c8d878020b115f8ba61d69/models/Asset/Image/Thumbnail/Processor.php#L116-L117

PR #7031, opened 2020-09-01
Report niceName/group not translateable in Panel Title and Menu

blankse:
The report name and group only translated in the left menu of the report page.


weisswurstkanone:
thx!

PR #7034, opened 2020-09-02
[DataObject][QuantityValue] Fixes for zero value, default unit (and inheritance)

podarcis:
This PR fixes two things:

1. Make zero (0) a legitimate value.
2. Doesn't enforce value zero (0) when a default unit is configured for the QuantityValue field.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/QuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/QuantityValue.php
@@ -257,7 +257,7 @@ class QuantityValue extends Data implements ResourcePersistenceAwareInterface, Q
      */
     public function getDataFromResource($data, $object = null, $params = [])
     {
-        if ($data[$this->getName() . '__value'] || $data[$this->getName() . '__unit']) {
+        if (strlen($data[$this->getName() . '__value']) > 0 || $data[$this->getName() . '__unit']) {
             $quantityValue = new Model\DataObject\Data\QuantityValue((float)$data[$this->getName() . '__value'], $data[$this->getName() . '__unit']);
 
             if (isset($params['owner'])) {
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -257,8 +257,9 @@ public function getDataForResource($data, $object = null, $params = [])
      */
     public function getDataFromResource($data, $object = null, $params = [])
     {
-        if ($data[$this->getName() . '__value'] || $data[$this->getName() . '__unit']) {
-            $quantityValue = new Model\DataObject\Data\QuantityValue((float)$data[$this->getName() . '__value'], $data[$this->getName() . '__unit']);
+        if (strlen($data[$this->getName() . '__value']) > 0 || $data[$this->getName() . '__unit']) {
```
weisswurstkanone, line 260: Why not simply compare it with  === null instead of the strlen magic ?

podarcis: I tried to be consistent with other existing code/ checks (e.g. ```ClassDefinition\Data\Numeric```). Feel free to modify my PR. If the value is guaranteed to be null in that case (and not an empty string) I'd prefer a type safe comparison as well.

weisswurstkanone: I don't see that "Numeric" does anything in getDataFromResource ...

weisswurstkanone: Besides that, qv_value is stored as a float in the db ...

podarcis: You're right, it's ```getDataFromResource()``` scope. In that case let's change it to null comparison. Would you mind altering my PR?

PR #7035, opened 2020-09-02
AbstractUser: Fix return PHPdoc

blankse:
PHPStan classify `$role = User\Role::create($options);` as AbstractUser now. Should be Role.


brusch:
Thanks!

PR #7036, opened 2020-09-02
fix wrong brackets for dao sql query generation

roland4432:
Corrected the way the brackets are set in the sql query generation method

## Additional info  
none - easy fix


brusch:
Thanks!

roland4432:
You're welcome :)

PR #7037, opened 2020-09-02
Increase the number of characters for workflow places

davidmaes:
Resolves #
This fixes the problem when you have complex workflows with too many parallel places.

255 characters is not enough to create complex workflows, hence the field is updated to text

When you have more then a few places in your workflow you run out of characters in the database table.
for instance 
translation_english_done,translation_spanish_done, etc. etc.
This results in the final string being cut off and breaking the workflows. You then have to enter the database and reset the field manually. Changing this to a text instead of varchar makes sure you can create much more complex workflows.

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20200902111642.php
+++ b/bundles/CoreBundle/Migrations/Version20200902111642.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+
+class Version20200902111642 extends AbstractPimcoreMigration
+{
+    /**
+     * @param Schema $schema
+     *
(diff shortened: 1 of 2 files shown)


brusch:
Thanks for your PR! 
One more thing, this needs to be fixed in the initial schema as well, see: 
https://github.com/pimcore/pimcore/blob/6cded713a05ac762fcde6b35e92341c3402c9a42/bundles/InstallBundle/Resources/install.sql#L876

@davidmaes can you please update your PR accordingly? Thanks!

davidmaes:
> Thanks for your PR!
> One more thing, this needs to be fixed in the initial schema as well, see:
> 
> https://github.com/pimcore/pimcore/blob/6cded713a05ac762fcde6b35e92341c3402c9a42/bundles/InstallBundle/Resources/install.sql#L876
> 
> @davidmaes can you please update your PR accordingly? Thanks!

Yes, didn't know that was a thing, good catch!

brusch:
Nice, thanks  a lot!

PR #7042, opened 2020-09-03
Image Optimization "dead lock"

das-peter:
The image optimize pipeline only works if format is set to "source".
I don't understand why that is.
However, even with format set as "source" the image optimization is _not_ triggered if an image url is directly requested for the first time.
E.g. requesting `/content/hero-slider/2020/09/image-thumb__52057__hero-slider-image/hero-slider-image-1.png` will pass through `Pimcore\Bundle\CoreBundle\Controller\PublicServicesController::thumbnailAction()`.
Now if the format is set to "source" this action will adjust the thumbnail config to set the format to the file extension.
Which then leads `\Pimcore\Model\Asset\Image\Thumbnail\Processor::process()` to never enable `$contentOptimizedFormat`  because of this condition `if ($format == 'source' || empty($format)) {`.
Format is never empty or "source" in that case - as such once the thumbnail is generated it wont be optimized ever again.

I do not know why calling the image optimization has anything to do with the 'source' format - and besides that why "original" wouldn't be considered for optimization either.
As far as I see it the optimization should happen for all _supported_ formats (keep excluding webp) and output channels where it is beneficial (exclude print / tiff).

I've made an attempt to change the approach.

### DIFF
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -168,6 +168,10 @@ class Processor
         $fileExtension = $format;
         if ($format == 'original') {
             $fileExtension = \Pimcore\File::getFileExtension($fileSystemPath);
+            // If we can optimize we need to set the concrete format to do so.
+            if ($contentOptimized = in_array($fileExtension, ['jpg', 'jpeg', 'png', 'pjpeg'])) {
+                $format = $fileExtension;
+            }
         } elseif ($format === 'pjpeg' || $format === 'jpeg') {
             $fileExtension = 'jpg';
         }
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -359,6 +359,11 @@ class Processor
             }
         }
 
+        // @TODO: Should this somehow check if a format switch is acutally allowed?
+        if ($contentOptimized && !self::hasWebpSupport()) {
+            $format = $image->getContentOptimizedFormat();
+        }
+
         $tmpFsPath = preg_replace('@\.([\w]+)$@', uniqid('.tmp-', true) . '.$1', $fsPath);
         $image->save($tmpFsPath, $format, $config->getQuality());
         @rename($tmpFsPath, $fsPath); // atomic rename to avoid race conditions
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -152,7 +152,6 @@ class Processor
         if (self::hasWebpSupport() && $image->supportsFormat('webp')) {
             // We can use wepb - no need to optimize.
             // @TODO Why can we just adjust the format here without checking if
-            // the format  is source / original and / or if any preserving
             // option is active?
             $optimizedFormat = $optimizedContent = false;
             $format = 'webp';
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -151,7 +151,6 @@ class Processor
 
         if (self::hasWebpSupport() && $image->supportsFormat('webp')) {
             // @TODO Why can we just adjust the format here without checking if
-            // option is active?
             $optimizedFormat = $optimizedContent = false;
             $format = 'webp';
         }
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -100,8 +100,8 @@ class Processor
         $generated = false;
         $errorImage = PIMCORE_WEB_ROOT . '/bundles/pimcoreadmin/img/filetype-not-supported.svg';
         $format = strtolower($config->getFormat());
-        // Optimize by default.
-        $contentOptimized = true;
+        // Optimize if allowed to strip info.
+        $contentOptimized = (!$config->isPreserveColor() && !$config->isPreserveMetaData());
 
         if (self::containsTransformationType($config, '1x1_pixel')) {
             return 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7';
@@ -168,8 +168,9 @@ class Processor
         $fileExtension = $format;
         if ($format == 'original') {
             $fileExtension = \Pimcore\File::getFileExtension($fileSystemPath);
-            // If we can optimize we need to set the concrete format to do so.
-            if ($contentOptimized = in_array($fileExtension, ['jpg', 'jpeg', 'png', 'pjpeg'])) {
+            // If we can optimize we check if the original is compatible and
+            // set the concrete target format to do so.
+            if ($contentOptimized && $contentOptimized = in_array($fileExtension, ['jpg', 'jpeg', 'png', 'pjpeg'])) {
                 $format = $fileExtension;
             }
         } elseif ($format === 'pjpeg' || $format === 'jpeg') {
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -149,7 +149,7 @@ class Processor
 
         $image = Asset\Image::getImageTransformInstance();
 
-        if (self::hasWebpSupport() && $image->supportsFormat('webp')) {
+        if ($optimizedFormat && self::hasWebpSupport() && $image->supportsFormat('webp')) {
             // @TODO Why can we just adjust the format here without checking if
             $optimizedFormat = $optimizedContent = false;
             $format = 'webp';
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -150,7 +150,6 @@ class Processor
         $image = Asset\Image::getImageTransformInstance();
 
         if (self::hasWebpSupport() && $image->supportsFormat('webp')) {
-            // We can use wepb - no need to optimize.
             // @TODO Why can we just adjust the format here without checking if
             // option is active?
             $optimizedFormat = $optimizedContent = false;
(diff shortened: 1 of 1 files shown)


models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -360,7 +371,7 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
 
         $generated = true;
 
-        if ($contentOptimizedFormat) {
+        if ($contentOptimized) {
```
brusch, line 374: Here would be the right place then for checking if an image should be optimized or not. 
But be careful here again! Pimcore supports custom optimizers based on the image type, so we can't hard-code this here. 
See also: `\Pimcore\Image\ImageOptimizerInterface`
Probably it's better to just add all images regardless of which type it is to the optimizer queue, if the 2 options mentioned before are not set.

brusch (changes requested):

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -350,7 +360,8 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
             }
         }
 
-        if ($contentOptimizedFormat && !self::hasWebpSupport()) {
+        // Ensure the image optimization is only called with a compatible format.
+        if ($contentOptimized && !self::hasWebpSupport() && !in_array($format, ['jpeg', 'png', 'pjpeg'])) {
             $format = $image->getContentOptimizedFormat();
```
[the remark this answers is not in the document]
das-peter: >  there's also the thumbnail config option to preserve the meta-data and the color profile of the source image

I've addressed that in the second to last commit.

> $image->getContentOptimizedFormat() returns a format depending on the source file or transformation

I removed all optimization related format checks / adjustments.
That said I know that there's another png issue with forcing Imagemagick to use _always_ use png32 but that's probably something for another pull request.

models/Asset/Image/Thumbnail/Processor.php
das-peter: > Here would be the right place then for checking if an image should be optimized or not.

By now the optimization is always queued, unless some kind of preserving option (color, metadata) is set or it's clear the thumbnail is not intended for web (print, tiff).

However, I currently wonder about the forced webp format. I've added a comment there because I don't understand how it is okay for that to just swoop in and hijack the format as soon as it is supported by the adapter.
I see that there's some sort of handling in `Asset\Image\Thumbnail::getHtml()` . But again, `Controller\PublicServicesController::thumbnailAction()` doesn't handle that. So if understand correctly a direct http request to an unavailable thumbnail, and with `webp_auto_support` enabled - which it is by default, will overwrite the output format regardless of the requested image format.

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -350,17 +359,13 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
             }
         }
 
-        if ($contentOptimizedFormat && !self::hasWebpSupport()) {
-            $format = $image->getContentOptimizedFormat();
```
brusch, line 354: Why have you removed that section? 
As I said: `getContentOptimizedFormat()` is a completely different functionality, it returns the image type (PNG or JPEG) depending on the image content, it has nothin to do with any optimization.

brusch (changes requested):

models/Asset/Image/Thumbnail/Processor.php
das-peter: It hijacks the format because it has no regard for the actually requested format.
So if you've got a png without alpha and format is `png`, the function will set the format to `pjpeg` - regardless if some sort of "auto-detection" is requested.
As you said, it's a completely different function and jet it still is/was somehow related to the optimization pipeline (?)
However, restored the code.

das-peter: That whole thing bothered me to much. I've split the flag indicating content and / or format optimization into two separate variables. Format optimization is only allowed if no specific format is requested.

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -100,7 +100,9 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
         $generated = false;
         $errorImage = PIMCORE_WEB_ROOT . '/bundles/pimcoreadmin/img/filetype-not-supported.svg';
         $format = strtolower($config->getFormat());
-        $contentOptimizedFormat = false;
+        // Optimize if allowed to strip info.
+        $optimizedContent = (!$config->isPreserveColor() && !$config->isPreserveMetaData());
```
brusch, line 104:
```suggestion
        $optimizeContent = (!$config->isPreserveColor() && !$config->isPreserveMetaData());
```

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -114,14 +116,17 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
...
             if ($format === 'jpeg') {
                 $format = 'pjpeg';
             }
-            $contentOptimizedFormat = true; // format can change depending of the content (alpha-channel, ...)
         }
 
         if ($format == 'print') {
+            // Don't optimize images for print as we assume we want images as
+            // untouched as possible.
+            $optimizedFormat = $optimizedContent = false;
```
brusch, line 129:
```suggestion
            $optimizedFormat = $optimizeContent = false;
```

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -134,6 +139,7 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
                 return $asset->getFullPath();
             }
         } elseif ($format == 'tiff') {
+            $optimizedFormat = $optimizedContent = false;
```
brusch, line 142:
```suggestion
            $optimizedFormat = $optimizeContent = false;
```

models/Asset/Image/Thumbnail/Processor.php
brusch, line 104:
```suggestion
        $optimizeContent = (!$config->isPreserveColor() && !$config->isPreserveMetaData());
```

models/Asset/Image/Thumbnail/Processor.php
brusch, line 129:
```suggestion
            $optimizedFormat = $optimizeContent = false;
```

models/Asset/Image/Thumbnail/Processor.php
brusch, line 142:
```suggestion
            $optimizedFormat = $optimizeContent = false;
```

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -143,7 +149,12 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
 
         $image = Asset\Image::getImageTransformInstance();
 
-        if ($contentOptimizedFormat && self::hasWebpSupport() && $image->supportsFormat('webp')) {
+        if (self::hasWebpSupport() && $image->supportsFormat('webp')) {
+            // We can use wepb - no need to optimize.
+            // @TODO Why can we just adjust the format here without checking if
+            // the format  is source / original and / or if any preserving
+            // option is active?
+            $optimizedFormat = $optimizedContent = false;
```
brusch, line 157:
```suggestion
            $optimizedFormat = $optimizeContent = false;
```

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -360,7 +371,7 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
 
         $generated = true;
 
-        if ($contentOptimizedFormat) {
+        if ($optimizedContent) {
```
brusch, line 374:
```suggestion
        if ($optimizeContent) {
```

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -143,7 +149,12 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
 
         $image = Asset\Image::getImageTransformInstance();
 
-        if ($contentOptimizedFormat && self::hasWebpSupport() && $image->supportsFormat('webp')) {
+        if (self::hasWebpSupport() && $image->supportsFormat('webp')) {
+            // We can use wepb - no need to optimize.
+            // @TODO Why can we just adjust the format here without checking if
```
brusch, line 154:
```suggestion
```

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -143,7 +149,12 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
...
+            // the format  is source / original and / or if any preserving
```
brusch, line 155:
```suggestion
```

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -143,7 +149,12 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
...
+            // option is active?
```
brusch, line 156:
```suggestion
```

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -143,7 +149,12 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
 
         $image = Asset\Image::getImageTransformInstance();
 
-        if ($contentOptimizedFormat && self::hasWebpSupport() && $image->supportsFormat('webp')) {
+        if (self::hasWebpSupport() && $image->supportsFormat('webp')) {
+            // We can use wepb - no need to optimize.
```
brusch, line 153:
```suggestion
```

brusch (changes requested):

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -143,7 +149,12 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
 
         $image = Asset\Image::getImageTransformInstance();
 
-        if ($contentOptimizedFormat && self::hasWebpSupport() && $image->supportsFormat('webp')) {
+        if (self::hasWebpSupport() && $image->supportsFormat('webp')) {
```
brusch, line 152:
```suggestion
        if ($optimizedFormat && self::hasWebpSupport() && $image->supportsFormat('webp')) {
```
WebP is only used if format is set to `auto` (`$format == 'source'`), otherwise the defined target format is used.

brusch (changes requested):

das-peter:
Sweet thanks!

PR #7051, opened 2020-09-03
Corehandler refactoring

### LINKED ISSUE #7039: [Cache] Serialization Performance Improvements
Currently all items that are added to the cache queue are getting serialized right away, even in case they don't get written to the cache. 

This happens here: 
https://github.com/pimcore/pimcore/blob/2662d44a8828ad8fc7c8d878020b115f8ba61d69/lib/Cache/Core/CoreHandler.php#L464

My proposal for optimization is that in case an object of type `ElementInterface` is getting added to the queue we do not serialize the item, but we just add an `ElementDescriptor` to the queue. 

Then later in process shutdown, when `\Pimcore\Cache\Core\CoreHandler::writeSaveQueue` is called, we just look for `ElementDescriptors` in the cache items and replace them by freshly loaded objects (using the `$force` argument)  from the database and serialize them ( e.g. `DataObject::getById(123, true)`  ).

weisswurstkanone:
Resolves #7039


lib/Cache/Core/CoreHandler.php
```diff
@@ -270,8 +276,29 @@ public function load($key)
 
         if ($item->isHit()) {
             $data = $item->get();
+
             $data = unserialize($data);
 
+            // TODO if this change is really for release 6.8 we could simplify this and
+            // use the service instead.
+            // see  "Add service for DeepCopy/ElementDescriptor tasks", https://github.com/pimcore/pimcore/pull/6983
+            $copier = new DeepCopy();
```
brusch, line 285: I guess we should use the new service `\Pimcore\Model\Element\Service::getDeepCopyInstance()` here

brusch: This is not going to be part of a bug fix release anyway 😉

lib/Cache/Core/CoreHandler.php
```diff
@@ -270,8 +276,17 @@ public function load($key)
...
         if ($item->isHit()) {
             $data = $item->get();
+
             $data = unserialize($data);
 
+            $context = [
+                'source' => __METHOD__,
+                'conversion' => 'unmarshal'
+            ];
+            $copier = Service::getDeepCopyInstance($data, $context);
```
brusch, line 286: Can we avoid this somehow? 
Doing a deep copy on load is slowing down the retrieval of data from the cache, so if possible we shouldn't do this. If necessary, we'd have to do some extensive performance testing so see how big the impact of this is.

models/Asset.php
```diff
@@ -290,7 +290,9 @@ public static function getById($id, $force = false)
 
                 $asset->resetDirtyMap();
 
-                \Pimcore\Cache::save($asset, $cacheKey);
+                if (!$force) {
```
brusch, line 293: I wouldn't do this. 
I think it would be better to just set a private flag in CoreHandler as soon as `writeSaveQueue()` is called (reset to `false` when done), which is then checked in `addToSaveQueue()` and ignores queuing items if set to `true`. With that we have covered this topic entirely in the cache where it belongs to.

brusch (changes requested):

brusch:
I'd prefer #7060 over this one, because this keeps the `load()` method of the cache as it is without adding any additional complexity  and therefore processing time.

lib/Cache/Core/CoreHandler.php
brusch: Do we really need that? This seems to be unnecessary as the dump state is anyway set to `false` before serialization, see: https://github.com/pimcore/pimcore/pull/7051/files#diff-903c6798cb9358159948a1eca1bb31fdR594 so on unserializing it it should be still `false` ... 
@weisswurstkanone can you please check - thanks!

brusch:
I'd prefer #7060 over this approach - closed.

PR #7052, opened 2020-09-03
block inside localized field 

weisswurstkanone:
see #6976


weisswurstkanone:
note that this PR needs #7051 changes are need for the unit tests to succeed

PR #7053, opened 2020-09-03
[Admin] Documents Tree - Fix AdminStyle on save

### LINKED ISSUE #7047: Text is not struck through after unpublishing document
Open published document, click "unpublish" => the document in the tree should be struck through, but it's not

image

dvesh3:
Resolves #7047

## Additional info  
AdminStyle should be updated for tree node on all document type save

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
@@ -47,7 +47,6 @@ use Symfony\Component\Routing\Annotation\Route;
  */
 class DocumentController extends ElementControllerBase implements EventedControllerInterface
 {
-    use AdminStyleTrait;
     use DocumentTreeConfigTrait;
 
     /**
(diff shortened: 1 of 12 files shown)


bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
```diff
@@ -48,6 +48,7 @@
 class DocumentController extends ElementControllerBase implements EventedControllerInterface
 {
     use AdminStyleTrait;
```
weisswurstkanone, line 50: Is the AdminStyleTrait still needed here ?

dvesh3: yes, still calling here: https://github.com/pimcore/pimcore/blob/969f4efbc7c9a480690e366dc0d0e1a9b5940440/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php#L106

https://github.com/pimcore/pimcore/blob/969f4efbc7c9a480690e366dc0d0e1a9b5940440/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php#L785

dvesh3: oh, you are right. we don't need it, since it is already used in DocumentTreeConfigTrait

PR #7056, opened 2020-09-04
Fix Pimcore 6.7 BC Break in Layout\Text rendering class

### LINKED ISSUE #7055: Pimcore 6.7: BC Break in Layout\Text rendering class
# Bug Report
After upgrade to Pimcore 6.7.x I have an error in the table option dialog:
```
Call to a member function getClassId() on null
```

After debug I see, the error came from my renderLayoutText method of an Layout\Text rendering class.

The interface say we get always an **Concrete**, but now we get **null** here.

The code in the documentation should throw also an error now:
https://github.com/pimcore/pimcore/blob/6324fa8ee9e85b83d894e091e561e50aa30291be/doc/Development_Documentation/05_Objects/01_Object_Classes/03_Layout_Elements/01_Dynamic_Text_Labels.md

We can change the PhpDoc of the interface, but its an BC break, right?

Regression from #6204

blankse:
Resolves #7055

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -1400,7 +1400,7 @@ class Service extends Model\Element\Service
     {
         $context['object'] = $object;
 
-        if (method_exists($layout, 'enrichLayoutDefinition')) {
+        if (method_exists($layout, 'enrichLayoutDefinition') && $object) {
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
thanks a lot!

PR #7058, opened 2020-09-04
Fix for BlockElement (Remove child's owner, cause it make recursion o…

kt-team-asimonov:
…n serialization)

  
  

Resolves #


weisswurstkanone:
thx but will be fixed via #7052 anyway

PR #7060, opened 2020-09-04
Corehandler: Collaboration approach

### LINKED ISSUE #6976: block inside localized field

### LINKED ISSUE #7039: [Cache] Serialization Performance Improvements

weisswurstkanone:
Resolves #7039
Fixes #6976


weisswurstkanone:
would make #7051 and #7052 obsolete

PR #7062, opened 2020-09-07
[Users] Fix auto-focus on user search

BlackbitNeueMedien:
Currently when you click `Search` on the user panel the input field (actually it is a combo box) does not get auto-focussed because the `afterrender` event fires too early.


BlackbitNeueMedien:
@dvesh3 Never did rebasing before. To not merge unwanted things from master which were already in the branch I just redid the changes based on 6.7 branch - certainly there will be a more elegant way...

dvesh3:
@BlackbitNeueMedien It looks good. thanks!

PR #7063, opened 2020-09-07
[Ghostscript] Use version_compare to compare ghostscript versions

BlackbitNeueMedien:
Currently the code compares versions with floats:
```php
if ($this->getVersion() >= 9.50) {
```

This works as long as there won't be a version 9.100 - for sure this is more of a hypothetical case but in my experience comparing versions is not as trivial as a numerical comparison.

### DIFF
--- a/lib/Document/Adapter/Ghostscript.php
+++ b/lib/Document/Adapter/Ghostscript.php
@@ -166,7 +166,7 @@ class Ghostscript extends Adapter
         $command = self::getGhostscriptCli() . ' -dNODISPLAY -q';
 
         // Adding permit-file-read flag to prevent issue with Ghostscript's SAFER mode which is enabled by default as of version 9.50.
-        if ($this->getVersion() >= 9.50) {
+        if (version_compare($this->getVersion(), '9.50') >= 0) {
             $command .= " --permit-file-read='" . $this->path . "'";
         }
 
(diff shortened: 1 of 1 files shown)


lib/Document/Adapter/Ghostscript.php
```diff
@@ -185,7 +185,7 @@ protected function buildPageCountCommand()
     protected function getVersion()
     {
         if (is_null($this->version)) {
-            $this->version = (float) Console::exec(self::getGhostscriptCli() . ' --version');
+            $this->version = Console::exec(self::getGhostscriptCli() . ' --version');
```
blankse, line 188: PHP doc type of the atttribute should be also changed, right?

BlackbitNeueMedien: Fixed, thanks.

dvesh3:
@BlackbitNeueMedien please see https://travis-ci.com/github/pimcore/pimcore/jobs/381870793#L1088

lib/Document/Adapter/Ghostscript.php
blankse: I mean not the method return (but is also needed, yes) but the attribute var doc

lib/Document/Adapter/Ghostscript.php
```diff
@@ -178,14 +178,14 @@ protected function buildPageCountCommand()
...
-     * @return float
+     * @return string
      *
      * @throws \Exception
      */
     protected function getVersion()
     {
         if (is_null($this->version)) {
-            $this->version = (float) Console::exec(self::getGhostscriptCli() . ' --version');
+            $this->version = Console::exec(self::getGhostscriptCli() . ' --version');
```
jdreesen, line 188: The PhpDoc type should be changed in line 30, too.

BlackbitNeueMedien: Fixed

dvesh3 (approved):

PR #7066, opened 2020-09-07
[Application Logger] Write log also to file object (if set)

BlackbitNeueMedien:
Currently you can provide a file object when logging a message with the `ApplicationLogger`, see https://pimcore.com/docs/6.x/Development_Documentation/Tools_and_Features/Application_Logger.html#page_Special-context-variables.

The goal of this should be to provide some more insights on the error in a separate file (as far as I understood). But currently the data for the `FileObject` can only be set when creating the `FileObject` object. When later on you call `$logger->alert('Something bad happened', ['fileObject' => $fileObject]);` the file object does only get used for linking the application log entry to the `FileObject` file. This PR changes this behaviour as it also adds the log message to the `FileObject` file. This way it is possible to create a `FileObject` for a bulk operation and then log all messages which happen during this operation to the `FileObject` file. When then there is an error, the user sees that in the Application Logger and can directly jump to the logs for this bulk operation and also sees some other logs which happened before the error but were not critical (e.g. in an import an error happens during saving of a data object - in this moment it may help to know which data got written to the object).
```php
$logFileObject = new FileObject('My bulk operation');

$dataObject = new Product;
$dataObject->setKey($key);
$logger->info('Key='.$key,  ['fileObject' => $logFileObject]);
$dataObject->setSKU($sku);
$logger->info('SKU='.$sku,  ['fileObject' => $logFileObject]);
try {
  $dataObject->save();
  $logger->info('Object '.$dataObject->getId().' successfully saved',  ['fileObject' => $logFileObject]);
catch(\Throwable $e) {
  $logger->alert('Object could not be saved: '.$e,  ['fileObject' => $logFileObject]);
}
```

Whenever you see `Object could not be saved` in your application logs you can open the linked file object and then you see the other fields' values:
```
My bulk operation
[info] Key=
[info] SKU=1234
[alert] Object could not be saved: Key is empty
```
This way it becomes easy to analyse the cause of the error. Without the info log with the SKU it would be much harder to analyse which object had en empty key.

Without this PR you have to collect all the logs in a variable and then create a `FileObject` with the collected messages and thereafter log a dummy log so that the linked file object appears in the Application Logger.

### DIFF
--- a/lib/Log/ApplicationLogger.php
+++ b/lib/Log/ApplicationLogger.php
@@ -158,7 +158,7 @@ class ApplicationLogger implements LoggerInterface
                 $context['fileObject'] = str_replace(PIMCORE_PROJECT_ROOT, '', $context['fileObject']->getFilename());
             }
 
-            file_put_contents(PHP_EOL.$context['fileObject'], $message, FILE_APPEND);
+            file_put_contents(PHP_EOL.$context['fileObject'], PHP_EOL.'['.$level.'] '.$message, FILE_APPEND);
         }
 
         $relatedObject = null;
--- a/lib/Log/ApplicationLogger.php
+++ b/lib/Log/ApplicationLogger.php
@@ -157,6 +157,8 @@ class ApplicationLogger implements LoggerInterface
             } else {
                 $context['fileObject'] = str_replace(PIMCORE_PROJECT_ROOT, '', $context['fileObject']->getFilename());
             }
+
+            file_put_contents($context['fileObject'].PHP_EOL, $message, FILE_APPEND);
         }
 
         $relatedObject = null;
--- a/lib/Log/ApplicationLogger.php
+++ b/lib/Log/ApplicationLogger.php
@@ -158,7 +158,7 @@ class ApplicationLogger implements LoggerInterface
                 $context['fileObject'] = str_replace(PIMCORE_PROJECT_ROOT, '', $context['fileObject']->getFilename());
             }
 
-            file_put_contents(PHP_EOL.$context['fileObject'], PHP_EOL.'['.$level.'] '.$message, FILE_APPEND);
+            file_put_contents($context['fileObject'], PHP_EOL.'['.$level.'] '.$message, FILE_APPEND);
         }
 
         $relatedObject = null;
--- a/lib/Log/ApplicationLogger.php
+++ b/lib/Log/ApplicationLogger.php
@@ -158,7 +158,7 @@ class ApplicationLogger implements LoggerInterface
                 $context['fileObject'] = str_replace(PIMCORE_PROJECT_ROOT, '', $context['fileObject']->getFilename());
             }
 
-            file_put_contents($context['fileObject'].PHP_EOL, $message, FILE_APPEND);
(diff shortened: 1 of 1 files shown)


lib/Log/ApplicationLogger.php
```diff
@@ -158,7 +158,7 @@ public function log($level, $message, array $context = [])
                 $context['fileObject'] = str_replace(PIMCORE_PROJECT_ROOT, '', $context['fileObject']->getFilename());
             }
 
-            file_put_contents(PHP_EOL.$context['fileObject'], $message, FILE_APPEND);
+            file_put_contents(PHP_EOL.$context['fileObject'], PHP_EOL.'['.$level.'] '.$message, FILE_APPEND);
```
jdreesen, line 161:
```suggestion
            file_put_contents($context['fileObject'], PHP_EOL.'['.$level.'] '.$message, FILE_APPEND);
```

BlackbitNeueMedien: Thanks, fixed this. I should not edit files manually after testing ;-)

brusch:
Hmm,  the file-object functionality is there to append some context sensitive data, e.g. the raw JSON request data to a custom REST API, not to add more logging data. The logging data is anyway stored in the database, I don't think it makes sense to add this also ti the file object. 
Also the current PR would break the entire functionality, as it would override the appended file (e.g. a JSON or XML file) with the logging message (by using `filr_put_contents()`. 

Thanks anyway for your PR 😊

BlackbitNeueMedien:
```Also the current PR would break the entire functionality, as it would override the appended file (e.g. a JSON or XML file) with the logging message (by using filr_put_contents().```
Nope, as this PR uses `FILE_APPEND` in `file_get_contents()` nothing gets overwritten.

Anyway, how would you encapsulate the logs of one process to later analyse what caused severe errors?

brusch:
It would break, ... e.g. if you attach an `xlsx` as an `fileObject` (from an importer or so) then you wouldn't be able to open the file anymore ... 😐

Good question, I know what you mean, but the `fileObject` isn't the right tool for that. 
Maybe the PID is the right thing? 
Filtering in the UI by the PID would give you all the log entries for the specific process 🤔

BlackbitNeueMedien:
@brusch Ok, you are right.
When working with the `FileObject` I wondered whether these actually get cleaned up. I did not find anything. I could imagine to delete all referenced file objects when the corresponding application log entries get archived. Did I miss this cleanup or shall I prepare a PR for that?

brusch:
@BlackbitNeueMedien You're right, there is not such cleanup yet.
I guess it would make sense to integrate this into the existing task here: https://github.com/pimcore/pimcore/blob/0ca66e1ec4b16f8adf1358fa714e7213180b8517/lib/Maintenance/Tasks/LogArchiveTask.php#L47 

Also I think it would make sense to run this task only once a day, same as here: 
https://github.com/pimcore/pimcore/blob/ca75246d82fe1190ce2268767ace84f157b168d9/lib/Maintenance/Tasks/CacheCleanupTask.php#L52

A PR would be really nice 😊

PR #7069, opened 2020-09-07
[Doc] Remove some unnecessary whitespace

jdreesen:
Some docs had `CRLF` line endings. Those got changed to `LF` which makes them appear completely changed in the diff ;)

It's best to view the diff while [ignoring whitespace changes](https://github.com/pimcore/pimcore/pull/7069/files?w=1).


brusch:
Thanks @jdreesen excellent!👌

PR #7071, opened 2020-09-08
[Translations] %count% parameter doesn't work with translation values longer than 190 characters

brusch:


NiklasBr:
@brusch this change broke use cases like:

```twig
{{ '{0} list.product-count.zero|{1} list.product-count.one|]1,Inf[ list.product-count.plural'|trans({'%count%': totalCount}) }}
```

Now it just returns the key:
image

Issue #1764, opened 2017-07-26
Object Editor: Localized Fields data inside fieldcollections not properly saved when there is more than one localized field container inside the field collection definition


weisswurstkanone:
Example:
- Fieldcollection XYZ
               + Localizedfields
                             + Input field 1
                             + Input field 2
               + Localizedfields
                             + Input field 3

Result: only data of input field 3 is saved

weisswurstkanone:
build 81

PR #7072, opened 2020-09-08
[Admin] DataObject - Fix only save last filled localizedfields in Objectbricks/Blocks

### LINKED ISSUE #7059: [Admin] DataObject - only save last filled localizedfields in Objectbricks/Blocks
# Bug Report

### Steps to reproduce
- Create Class with Objectbrick:

- Create Objectbrick with 2 localizedfields
image

- Fill both localizedfields and save

- Reload

### Expected behavior
- Both localizedfields are filled

### Actual behavior
- Only the second localizedfield is filled

dvesh3:
Resolves #7059

## Additional info  
Similar fix like fieldcollections https://github.com/pimcore/pimcore/issues/1764

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Block.php
+++ b/models/DataObject/ClassDefinition/Data/Block.php
@@ -684,7 +684,7 @@ class Block extends Data implements CustomResourcePersistingInterface, ResourceP
         }
 
         if ($def instanceof DataObject\ClassDefinition\Data) {
-            $existing = $fields[$def->getName()];
+            $existing = $fields[$def->getName()] ?? false;
             if ($existing && method_exists($existing, 'addReferencedField')) {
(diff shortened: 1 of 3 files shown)


models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -684,7 +684,15 @@ public function doGetFieldDefinitions($def = null, $fields = [])
         }
 
         if ($def instanceof DataObject\ClassDefinition\Data) {
-            $fields[$def->getName()] = $def;
+            $existing = $fields[$def->getName()];
```
blankse, line 687:
```suggestion
            $existing = $fields[$def->getName()] ?? null;
```

dvesh3: already pushed..thanks!

blankse (approved):

PR #7074, opened 2020-09-08
Change admin UI state only when (un)publishing documents succeeds

jdreesen:
We're validating documents (and their editables) on the server-side when publishing them and throw a `ValidationException` when they're not in a valid state.

The moment I click on the "Save & Publish" button, the buttons in the admin UI changes: the "Save" button disappears and the "Unpublish" button appears, etc.

But if now the exception is thrown, the document is not saved and the buttons display a wrong state. 

This PR tries to fix this.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
@@ -77,7 +77,7 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
         tabPanel.setActiveItem(tabId);
     },
 
-    save: function (task, only, callback, failureCallback) {
+    save: function (task, only, callback, successCallback) {
 
         if (this.tab.disabled || this.tab.isMasked()) {
             return;
@@ -125,6 +125,10 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
                 success: function (response) {
                     try {
                         var rdata = Ext.decode(response.responseText);
+                        if (typeof successCallback == 'function') {
+                            //the successCallback function retrieves response data information
+                            successCallback(rdata);
+                        }
                         if (rdata && rdata.success) {
                             pimcore.helpers.showNotification(t("success"), t("saved_successfully"), "success");
                             this.resetChanges();
@@ -160,10 +164,6 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
                     if (this.newerVersionNotification) {
                         this.newerVersionNotification.setVisible(newerVersionNotificationBackup);
                     }
-
-                    if (typeof failureCallback == "function") {
-                        failureCallback();
-                    }
                 }.bind(this),
             });
         } else {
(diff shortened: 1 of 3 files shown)


dvesh3:
@jdreesen how about we use `successCallback` just like objects, then it is consistent and also we don't need to retain previous state. WDYT?
https://github.com/pimcore/pimcore/blob/0abcbeca8b11732d9165a8f49214376112e901b1/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js#L690

jdreesen:
Thx for the hint, I updated the code accordingly.

It's best to [look at the changes when ignoring whitespace](https://github.com/pimcore/pimcore/pull/7074/files?w=1).

jdreesen:
I fixed some inconsistencies in the `object.js` as well.

bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
```diff
@@ -175,63 +176,69 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
...
         this.publish(null, function () {
-            var tabPanel = Ext.getCmp("pimcore_panel_tabs");
-            tabPanel.remove(this.tab);
+            this.close();
         }.bind(this));
     },
 
     publish: function (only, callback) {
-        this.data.published = true;
+        this.save("publish", only, callback, function (rdata) {
```
dvesh3, line 197: update page_snippet::publish to pass successCallback to $super https://github.com/pimcore/pimcore/blob/b82377b72d2b1d9deb4f1db8c9e9a74b72ff911e/bundles/AdminBundle/Resources/public/js/pimcore/document/page_snippet.js#L450

bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
```diff
@@ -77,7 +77,7 @@ pimcore.document.document = Class.create(pimcore.element.abstract, {
         tabPanel.setActiveItem(tabId);
     },
 
-    save: function (task, only, callback) {
+    save: function (task, only, callback, successCallback) {
```
dvesh3, line 80: update page_snippet::save to pass successCallback to $super:
https://github.com/pimcore/pimcore/blob/b82377b72d2b1d9deb4f1db8c9e9a74b72ff911e/bundles/AdminBundle/Resources/public/js/pimcore/document/page_snippet.js#L460

jdreesen: Good catch.

bundles/AdminBundle/Resources/public/js/pimcore/document/document.js
jdreesen: Is this really needed? The signature of `publish` hasn't changed.

dvesh3: you are right..only changing save should be sufficient.

dvesh3:
👍 thanks a lot!

Issue #2677, opened 2018-03-09
parentId class property breaks o_parentId relation on first object save


ascheider:
# Bug Report
## Please make sure all the following boxes are checked before submitting your issue - thank you!
- [x] This is not a security issue / vulnerability -> send them to security@pimcore.org instead
- [x] Behavior is reproducible on http://demo.pimcore.org/admin/ (admin/demo) 
or on a clean install of the latest build (out of your project, to avoid problems with custom code, plugins, ...) 
- [ ] You're not able to fix the problem yourself and send us a pull request instead of an issue. 
- [x] There's no existing ticket for the same issue
- [x] Remove unused sections from below and this bullet points

### Expected behavior 
  1. Create object and save it.
  2. Reload tree, object still present
### Actual behavior  
  1. Create object and save it.
  2. Reload tree, object missing.
### Steps to reproduce  
  1. Create class with parentId property
![image](https://user-images.githubusercontent.com/6016485/37210450-0592c9a6-23a9-11e8-9017-0ca07d02d901.png)
  2. Create object, input some data and save. After tree reloading, object disappears from list and o_parentId of object is NULL.
![image](https://user-images.githubusercontent.com/6016485/37210628-b873bbac-23a9-11e8-97c5-c91a1d6ba409.png)
  3. Save object again and reload sidebar. Object is now present and o_parentId is correctly set in the database.
![image](https://user-images.githubusercontent.com/6016485/37210651-cb6b2236-23a9-11e8-8c55-39f3c894d47d.png)

weisswurstkanone:
added parentId to the list of forbidden names.

build 219

ascheider:
@weisswurstkanone wow that was fast;) thx.

PR #7082, opened 2020-09-11
fix: forbidden names can be added to class

kjkooistra-youwe:
Caused by case sensitive check while field name is changed to lower case in the isValid() check.
Fixed by converting all forbidden names to lower case.

To reproduce: add the field 'parentid', 'parentId', 'parentID' or 'pArEnTiD' to a class (you get the idea). It's all allowed.
Relates to https://github.com/pimcore/pimcore/issues/2677


dvesh3:
thnx!

PR #7084, opened 2020-09-11
Pimcore\Browser deprecation in favour of Browser

NiklasBr:

Allows us to better keep [Browser.php](https://github.com/cbschuld/Browser.php) updated through Composer. I'm a little unsure if extending the class is the best way to do this, or if I should try to use `class_alias()` instead. Also, is actual deprecation itself acceptable? I'm just guessing here :)


brusch:
Thanks a lot for your PR! 
I'd say `extend` is fine in this case, because `class_alias()` makes only sense if we can do the loading vice versa, like here: 
https://github.com/pimcore/pimcore/blob/04cbbfd0916ffc926226b302c58f7440450e1738/models/Document/Tag/Href.php#L25
but that's actually not possible.

PR #7099, opened 2020-09-14
Fixed: programmatically setting children of type "variant" will not be loaded when calling getChildren([AbstractObject::OBJECT_TYPE_VARIANT])

mcuenez:
## Description
When programmatically creating an object and its variants then setting the variants via `setChildren($variants)` will result in unexpected behaviour when trying to call `getChildren([AbstractObject::OBJECT_TYPE_VARIANT])`, if the object/variants is/are never saved.

## Expected Behaviour
Whenever the method `setChildren` is called, I'd expect it to return the same children when calling `getChildren`.

## Actual Behaviour
When calling `getChildren([AbstractObject::OBJECT_TYPE_VARIANT])`, the returned array is empty.

## Example Code
```php
$product = new Product();
$product->setChildren($variants);
p_r($product->getChildren([AbstractObject::OBJECT_TYPE_VARIANT]));
```

## The Problem
Setting the children works differently than getting the children when it comes to building the cache key.
In the method `getChildren` one automatically defines the cache key with the passed arguments:

```php
$cacheKey = $this->getListingCacheKey(func_get_args());
```
However, in the method `setChildren`, the cache key is always the same:

```php
$cacheKey = $this->getListingCacheKey();
```

## The Solution
Building the cache key the same way as `getChildren` does, will resolve this issue:
```php
public function setChildren($children, array $objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER], $includingUnpublished = false)
    {
        if ($children === null) {
            // unset all cached children
            $this->o_children = [];
            $this->o_hasChildren = [];
        } elseif (is_array($children)) {
            //default cache key
            $cacheKey = $this->getListingCacheKey([$objectTypes, $includingUnpublished]);
            $this->o_children[$cacheKey] = $children;
            $this->o_hasChildren[$cacheKey] = (bool) count($children);
        }

        return $this;
    }
```

Now, when setting variants using `setChildren($variants, [AbstractObject::OBJECT_TYPE_VARIANT])` it is possible to retrieve the same variants using `getChildren([AbstractObject::OBJECT_TYPE_VARIANT])`.

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
@@ -7,6 +7,7 @@ If you have custom editables or wysiwyg global config then please change namespa
 - Javascript - Class `pimcore.document.tag` is deprecated as well and will be removed in 7. Use new class `pimcore.document.editable` instead.
 - All classes in namespace `Pimcore\Document\Tag` moved to new namespace `Pimcore\Document\Editable` (including their services) for better readability and marked as deprecated. Please update custom document editable classes and mappings. [#6921](https://github.com/pimcore/pimcore/pull/6921)
 - All `document_tag_` css classes are deprecated and will be removed in Pimcore 7. Use new classes `document_editable_` for custom styling of document editables.
+- If `AbstractObject` was extended and `setChildren()` overridden, you'll run into a **BC**.
 
 ## 6.7.0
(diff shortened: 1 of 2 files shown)


brusch:
We should add a note to the upgrade notes that the method signature changed.

brusch:
Should go into the 6.8 release as there's a very little chance that it breaks existing code (if `AbstractObject` was extended and `setChildren()` overridden, what is very unlikely but still possible).

dvesh3:
@mcuenez please see requested changes ;) thanks!

mcuenez:
see #7125

brusch:
`master` branch was fine as this patch should go into `6.8`, just the docs were missing, but now this PR is fine to be merged, just waiting for Travis to finish the jobs 😊

dvesh3:
thanks a lot!

PR #7100, opened 2020-09-14
Added translation support in the version tab of the data objects

### LINKED ISSUE #7098: [DataObject] Missing admin translations on versions
The column "name" will not be translated in the versions tab.

### Expected behavior

object-edit

### Actual behavior
object-versions

### Steps to reproduce
Class definiton:

class-definition

PR will coming soon.

skochTC:
Closes #7098 

Not sure if this is ok with " \Pimcore\Model\Translation\Admin::getByKeyLocalized" but " $this->translate" is only for the shared translations.

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/DataObject/DataObject/diffVersions.html.php
+++ b/bundles/AdminBundle/Resources/views/Admin/DataObject/DataObject/diffVersions.html.php
@@ -11,6 +11,7 @@
 
 use Pimcore\Model\DataObject;
 
+$this->get('translate')->setDomain('admin');
 $fields = $this->object1->getClass()->getFieldDefinitions();
 ?>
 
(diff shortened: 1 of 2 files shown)


dvesh3:
@skochTC 

> Not sure if this is ok with " \Pimcore\Model\Translation\Admin::getByKeyLocalized" but " $this->translate" is only for the shared translations.

You can set admin domain for translator `$this->get("translate")->setDomain("admin");`
https://github.com/pimcore/pimcore/blob/001b7ee3b4677b2071813ecf56906373377b1a92/bundles/AdminBundle/Resources/views/Admin/Login/login.html.php#L5

tvedderTC:
> @skochTC
> 
> > Not sure if this is ok with " \Pimcore\Model\Translation\Admin::getByKeyLocalized" but " $this->translate" is only for the shared translations.
> 
> You can set admin domain for translator `$this->get("translate")->setDomain("admin");`
> https://github.com/pimcore/pimcore/blob/001b7ee3b4677b2071813ecf56906373377b1a92/bundles/AdminBundle/Resources/views/Admin/Login/login.html.php#L5

I updated the translation support in the version tab of the data objects for @skochTC .

dvesh3:
thanks a lot!

PR #7118, opened 2020-09-16
Support stream wrappers in LowQualityImagePreviewTask

BlackbitNeueMedien:
As https://github.com/pimcore/pimcore/blob/5dba19710d71e7843d19d2fed8e82feec022c348/lib/Image/Adapter/Imagick.php#L71-L76 correctly says Imagick does not support stream wrappers. The `LowQualityImagePreviewTask` does currently not handle this. 
This is especially necessary when `PIMCORE_TEMPORARY_DIRECTORY` uses a stream wrapper.


brusch:
Thanks!

PR #7125, opened 2020-09-17
Fixed: programmatically setting children of type "variant" will not be loaded when calling getChildren([AbstractObject::OBJECT_TYPE_VARIANT])

mcuenez:
see #7099


dvesh3:
should stay on master so we use https://github.com/pimcore/pimcore/pull/7099 only

PR #7126, opened 2020-09-17
Pass params from link editable to link generator

jremmurd:
With this PR the defined params in the link editable now get passed to the link generator service. 

image

### DIFF
--- a/models/Document/Editable/Link.php
+++ b/models/Document/Editable/Link.php
@@ -296,12 +296,17 @@ class Link extends Model\Document\Editable
                                 if ($realPath) {
                                     $this->data['path'] = $object->getFullPath();
                                 } else {
+                                    $parameters = [];
+                                    parse_str($this->getParameters(), $parameters);
                                     $this->data['path'] = $linkGenerator->generate(
                                         $object,
-                                        [
-                                            'document' => $this->getDocument(),
-                                            'context' => $this,
-                                        ]
+                                        array_merge(
+                                            $parameters,
+                                            [
+                                                'document' => $this->getDocument(),
+                                                'context' => $this,
+                                            ]
+                                        )
                                     );
                                 }
                             }


brusch:
@jremmurd thanks a lot for your PR. As I'd consider this as a bug, can you please rebase on `6.7`? 
Thanks a lot!

jremmurd:
@brusch I just noticed that this section will add the parameters anyways. I am not sure what's the expected behaviour. I would suggest to not add the configured parameters by hard to the url path, if they get now passed to the link generator?

see https://github.com/jremmurd/pimcore/blob/patch-43/models/Document/Editable/Link.php#L253

dvesh3:
@jremmurd https://github.com/jremmurd/pimcore/blob/patch-43/models/Document/Editable/Link.php#L253 this adds params for all types no matter if you use link generator or not.

jremmurd:
> @jremmurd https://github.com/jremmurd/pimcore/blob/patch-43/models/Document/Editable/Link.php#L253 this adds params for all types no matter if you use link generator or not.
> 
> So that means it already works 😉 can we close this PR?

No.

The PR allows to consider the configured params in the link generator service. I.e. if one wants to manipulate them or to extend them ... 
the problem is that, if the params are already set in the generator they will be added twice in line 253 i.e. `www.test.at?food=bar?foo=bar`. To me, if a link generator is configured the params should not be added anymore.

dvesh3:
@jremmurd ok, I understand your concern now. Still you can use context to manipulate or extend params...right? https://github.com/pimcore/pimcore/blob/f82b720a56978cfeac0be2eb28ce50af388cd30a/models/Document/Editable/Link.php#L303

jremmurd:
> @jremmurd ok, I understand your concern now. Still you can use context to manipulate or extend params...right?
> 
> https://github.com/pimcore/pimcore/blob/f82b720a56978cfeac0be2eb28ce50af388cd30a/models/Document/Editable/Link.php#L303

You are right, gues I missed that.

PR #7131, opened 2020-09-17
[Dependencies] - Sort dependencies by type and path

BlackbitNeueMedien:
One of our clients is working with the dependencies on the objects and asked if it is possible to sort the list of data alphabetically by path - without clicking the path column on every page to sort (which of course only sorts for current page).

I customized the SQL query a bit so that JOIN can be used to find the path of the other tables (document, assets and objects).


BlackbitNeueMedien:
Perhaps the same would make sense for https://github.com/pimcore/pimcore/blob/3d6d51756e678e1dd257573bec0404472cc9e2ec/models/Dependency/Dao.php#L47

tests/Model/Element/DependenciesTest.php
```diff
@@ -78,7 +78,9 @@ public function testObjectDependencies()
         $source = TestHelper::createEmptyObject();
 
         /** @var Unittest[] $targets */
-        $targets = TestHelper::createEmptyObjects('', true, 3);
+        for ($i = 0; $i <= 2; $i++) {
+            $targets[] = TestHelper::createEmptyObjects($i, true, 1)[0];
```
dvesh3, line 82:
```suggestion
            $targets[] = TestHelper::createEmptyObject($i, true);
```

tests/Model/Element/DependenciesTest.php
```diff
@@ -103,7 +105,9 @@ public function testDocumentDependencies()
         /** @var Document $source */
         $source = TestHelper::createEmptyDocumentPage();
         /** @var Unittest[] $targets */
-        $targets = TestHelper::createEmptyObjects('', true, 3);
+        for ($i = 0; $i <= 2; $i++) {
+            $targets[] = TestHelper::createEmptyObjects($i, true, 1)[0];
```
dvesh3, line 109:
```suggestion
            $targets[] = TestHelper::createEmptyObject($i, true);
```

tests/Model/Element/DependenciesTest.php
```diff
@@ -128,7 +132,11 @@ public function testAssetDependencies()
         /** @var Asset $source */
         $source = TestHelper::createImageAsset();
         /** @var Unittest[] $targets */
-        $targets = TestHelper::createEmptyObjects('', true, 3);
+        $targets = [];
+        for($i=0;$i<=2;$i++) {
+            $targets[] = TestHelper::createEmptyObjects($i, true, 1)[0];
```
dvesh3, line 137:
```suggestion
            $targets[] = TestHelper::createEmptyObject($i, true);
```

tests/Model/Element/DependenciesTest.php
BlackbitNeueMedien: Fixed.

tests/Model/Element/DependenciesTest.php
BlackbitNeueMedien: Fixed.

tests/Model/Element/DependenciesTest.php
BlackbitNeueMedien: Fixed.

dvesh3:
thank you!

PR #7139, opened 2020-09-18
[Document] Link editable - generate absolute path for sub-site links

dvesh3:

When using sub-site path in link editable, generate absolute url based on sub-site domain.

## Additional info  
subsite_link

image


dvesh3:
should stay on master as link editable class is refactored

brusch:
@dvesh3 At the moment we're only using the full qualified path in the admin interface when referencing documents, I guess this change would introduce an inconsistency here, right?

brusch:
Closed in favor of #7187 

The challenge is that the right link should be already generated in `\Pimcore\Model\Document::getFullPath()`, but since it's not clear  for which context the link should be generated it could be wrong in some cases.

PR #205, opened 2015-07-31
REST Service: use class mapping in object creation

lorextera:
REST Service use class mapping only in update mode with Object::getById() method.
Now, with this fix, also in add mode.

PR #7143, opened 2020-09-18
Extends the configurability of custom views

Lexipowder:
With this extension, you can define to which depth in a custom view object tree a specific object is available.
It's also backward compatible, both markups can be used:

- 'classes' => 'PRODUCT,3,CONFIGITEM'

- 'classes' => [
      "PRODUCT" => 1,  
      "3" => 3,
      "CONFIGITEM" => 4
   ],

This extension, in combination with a very limited custom view, is very helpful when working with automated hierarchy trees that depend heavily on inheritance and order levels. 
   
Plus an additional PR:
If an object, for example, has been locked by code, the unlock button in the context menu can be hidden as well.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -149,9 +149,9 @@ class DataObjectController extends ElementControllerBase implements EventedContr
 
                 if ($cv['classes']) {
                     $cvConditions = [];
-                    $cvClasses = explode(',', $cv['classes']);
-                    foreach ($cvClasses as $cvClass) {
-                        $cvConditions[] = "objects.o_classId = '" . $cvClass . "'";
+                    $cvClasses = $cv['classes'];
+                    foreach ($cvClasses as $key => $cvClass) {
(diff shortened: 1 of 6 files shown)


brusch:
Can you please update the docs accordingly as well? 
https://github.com/pimcore/pimcore/blob/cd6e0ffae5c9013cf461e06a251fb47842b5cf16/doc/Development_Documentation/05_Objects/01_Object_Classes/05_Class_Settings/20_Custom_Views.md#L1
Thanks!

weisswurstkanone:
Nice! Would it be possible to add this here as well?
https://github.com/pimcore/demo/blob/master/app/config/pimcore/customviews.example.php
https://github.com/pimcore/demo/blob/master/app/config/pimcore/customviews.php

brusch:
@Lexipowder 🏓😊

Lexipowder:
Please see [#205](https://github.com/pimcore/demo/pull/205)

dvesh3:
@Lexipowder thanks for your changes. It would be nice to update https://github.com/pimcore/skeleton/blob/master/config/pimcore/customviews.example.php as well.

Lexipowder:
[#41](https://github.com/pimcore/skeleton/pull/41)

PR #7150, opened 2020-09-22
Improve the verbosity of validation messages

IgorBenko:
This pull request fixes and improves very incomplete display of validation messages when saving Data Objects. The fix will improve validation messages of Localized Fields and Field collections. The context is also added so it now can be recognised in which language and which Field-Collection-Entry the field is incomplete. 

Before:

After:

Above messages are based on this empty form:

The example message above might seem too complex, but this is because a huge amount of fields was on purpose left empty for demonstration purposes.


weisswurstkanone:
thx!

PR #7156, opened 2020-09-22
[Object Grid] Missing field name and values when edit related fields from/in grid view table popup

### LINKED ISSUE #7140: [Object Grid] Missing field name and values when edit related fields from/in grid view table popup
# Bug Report

### Expected behavior
- I open the grid view https://demo.pimcore.fun/admin/login/deeplink?object_94_folder
- I add the "Categories" relation field to the table
- I double-click the category field in row (e.g id: 98)  to edit the categories
- I will see the category path and other fields selected in object definition

### Actual behavior
image

image

image

### Steps to reproduce
- see "Expected behavior"

weisswurstkanone:
fixes #7140

### DIFF
--- a/models/DataObject/ClassDefinition/Layout/DynamicTextLabelInterface.php
+++ b/models/DataObject/ClassDefinition/Layout/DynamicTextLabelInterface.php
@@ -23,7 +23,7 @@ interface DynamicTextLabelInterface
 {
     /**
      * @param string $data as provided in the class definition
-     * @param Concrete $object
+     * @param Concrete|null $object
      * @param mixed $params
      *
      * @return string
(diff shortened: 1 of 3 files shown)


models/DataObject/ClassDefinition/Layout/DynamicTextLabelInterface.php
```diff
@@ -23,7 +23,7 @@ interface DynamicTextLabelInterface
 {
     /**
      * @param string $data as provided in the class definition
-     * @param Concrete $object
+     * @param Concrete|null $object
```
blankse, line 26: This can't be null.
The object param in Pimcore\Model\DataObject\ClassDefinition\Layout\Text::enrichLayoutDefinition can be null now.

weisswurstkanone: In the current implementation it can't be null, true. 

Better to move the change in the documentation into a separate task for 7.0? IMHO, it should be able to handle null values as well solely based on the $params. Purpose could be a "preview" or whatever.

weisswurstkanone:
Humble question: is there a specific why the "height" is not covered?

dvesh3 (approved):

PR #7158, opened 2020-09-22
[Document] Navigation: Add missing argument to recursive method call

jdreesen:
#5747 added the possibility to limit the navigation depth, but misses passing the `$maxDepth` value in the recursive call.

Thus, it only works if `$maxDepth` is `1` (when the method is called directly). After that, in the recursive calls, `$maxDepth` is always `null` which means it will never trigger.


dvesh3:
Thanks!

Issue #699, opened 2016-06-24
[Improvement] Grid filter improvement: link (href/mutlihref), images, and empty values.


Yivan:
On the object grid, it is possible to filter "path" column. But for instance "href"/"multihref"(asset/objectdocument link) type column can not be filtered.

By the way adding the possibility to filter "empty" values would help administrator to find for instance object with empty href/multihref link. Or object where image (image type field) is not filled. I know it could be done by making a custom report to find empty value, but is is more convenient for enduser to use directly the grid when he work on his dataset.

kubaplas:
@brusch do you have any plans regarding this issue?

fashxp:
not on our prio list yet, we are happy to merge PRs for that though (maybe with a short discussion about the planned solution before implementing it completely)

greefhorst:
@fashxp I've been working on the issue for some time, but ran into the same issues as @BlackbitNeueMedien in #3031: how to fill the filters and trigger an ajax call? In the meantime, I was able to do build a well working list filter, but this might result in a very long options list (although 200 options is already accepted for Country select fields). How do you feel about this? Or do you have any suggestions on how linking the values from a popup to the filter?
<img width="1361" alt="image" src="https://user-images.githubusercontent.com/70212046/94911631-b9031000-04a6-11eb-99c4-ede2ed0fbf21.png">
(I used the object key in the example, but this might also be another column (path?), or maybe using the PathFormatter class configured for the field.

fashxp:
@greefhorst  thank you for your input. 
> How do you feel about this?
looks nice, but I'm not sure if this really scales and is robust enough. what, if you have thousands of objects of the liked class (or no class limit in the class definition at all, so all data objects in the system are valid options). 

> Or do you have any suggestions on how linking the values from a popup to the filter?
sorry, not really. 
maybe not implement it as grid filter from ext but as some custom filter similar to the sql filter?

kubaplas:
@greefhorst Did you continue working on this feature? If you dropped it, maybe you can share the code? I can try to finish the job ;)

brusch:
Fixed by #10022 🥳

PR #7160, opened 2020-09-22
[Grid View] Make Direct SQL Button visible through permissions

greefhorst:
# Feature Request

Making the Direct SQL Query functionality available for non-admin users

## Feature description
As Improvement #699 has not yet received any priority, using the Direct SQL Query functionality allows users with basic SQL and data structure knowledge to filter on relations. In ancient versions (PimCore 4.5) normal users had this option as well, but nowadays the feature is only available for admin users. The proposed changes make the Direct SQL Query available for other users as well - though only if the new permission `direct_sql_query` is set.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js
@@ -175,7 +175,7 @@ pimcore.object.helpers.gridTabAbstract = Class.create({
             iconCls: "pimcore_icon_sql",
             enableToggle: true,
             tooltip: t("direct_sql_query"),
-            hidden: !pimcore.currentuser.admin,
+            hidden: pimcore.currentuser.permissions.indexOf("direct_sql_query") < 0,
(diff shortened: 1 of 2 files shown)


brusch:
I'm not sure if we really should introduce a dedicated permission for that, if you ask me, I wouldn't do that. 
We're trying to keep the permissions as lean as possible otherwise it will get harder and harder to initially do the permissioning for users. 

Even if we'd introduce a new permission, it should be checked on the server-side as well, not just on the client-side:
https://github.com/pimcore/pimcore/blob/1e49aaaf53686624e6eada9260b624dc8fadb328/bundles/AdminBundle/Helper/GridHelperService.php#L589

It would be also necessary to include the permission in the intial setup: 
https://github.com/pimcore/pimcore/blob/12e8ce54e686b37a0210e261d291b8842c0a7c0f/bundles/InstallBundle/Installer.php#L817

dvesh3:
@greefhorst Direct SQL Query button is only available to administrators for security reasons e.g. SQL injection. It will be a comprise if we add a config for this option. Anyway, thanks for your efforts :)

PR #7161, opened 2020-09-23
Fix "Undefined index: id" error after save

blankse:
The error occurred after i removed an object brick

image


weisswurstkanone:
Thanks a lot!

PR #7164, opened 2020-09-23
[Application logs] Cleanup referenced FileObjects

BlackbitNeueMedien:
Follow-up of https://github.com/pimcore/pimcore/pull/7066#issuecomment-689345674

### DIFF
--- a/lib/Maintenance/Tasks/LogArchiveTask.php
+++ b/lib/Maintenance/Tasks/LogArchiveTask.php
@@ -42,6 +42,7 @@ final class LogArchiveTask implements TaskInterface
     /**
      * @param Db\ConnectionInterface $db
      * @param Config $config
+     * @param LoggerInterface $logger
      */
     public function __construct(Db\ConnectionInterface $db, Config $config, LoggerInterface $logger)
     {
--- a/lib/Maintenance/Tasks/LogArchiveTask.php
+++ b/lib/Maintenance/Tasks/LogArchiveTask.php
@@ -65,10 +65,10 @@ final class LogArchiveTask implements TaskInterface
             $tablename = $db->quoteIdentifier($this->config['applicationlog']['archive_alternative_database']).'.'.$tablename;
         }
 
-        $archive_treshold = (int) ($this->config['applicationlog']['archive_treshold'] ?? 30);
+        $archive_threshold = (int) ($this->config['applicationlog']['archive_treshold'] ?? 30);
 
         $timestamp = time();
-        $sql = 'SELECT %s FROM '.ApplicationLoggerDb::TABLE_NAME.' WHERE `timestamp` < DATE_SUB(FROM_UNIXTIME('.$timestamp.'), INTERVAL '.$archive_treshold.' DAY)';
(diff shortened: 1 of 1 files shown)


lib/Maintenance/Tasks/LogArchiveTask.php
```diff
@@ -79,5 +89,28 @@ public function execute()
...
+        $lockId = self::class;
+        if (!Lock::isLocked($lockId, 86400) && date('H') <= 4) {
+            // execution should be only sometime between 0:00 and 4:59 -> less load expected
+            Lock::lock($lockId);
+            $this->logger->debug('Deleting referenced FileObjects of application_logs which are older than '. $archive_treshold.' days');
+            $fileIterator = new \DirectoryIterator(PIMCORE_LOG_FILEOBJECT_DIRECTORY);
+            $fileIterator = new \CallbackFilterIterator(
+                $fileIterator,
+                static function (\SplFileInfo $fileInfo) use ($archive_treshold) {
+                    return time() - $archive_treshold * 86400 > $fileInfo->getMTime();
```
jdreesen, line 102: `time() - $archive_treshold * 86400` should be calculated outside the closure. There's no need to repeat it every time.

BlackbitNeueMedien: You are right, fixed that. Thanks.

lib/Maintenance/Tasks/LogArchiveTask.php
```diff
@@ -77,7 +87,32 @@ public function execute()
                     ) ENGINE = ARCHIVE ROW_FORMAT = DEFAULT;");
 
             $db->query('INSERT INTO '.$tablename.' '.sprintf($sql, '*'));
-            $db->query('DELETE FROM '.ApplicationLoggerDb::TABLE_NAME.' WHERE `timestamp` < DATE_SUB(FROM_UNIXTIME('.$timestamp.'), INTERVAL '.$archive_treshold.' DAY);');
+            $db->query('DELETE FROM '.ApplicationLoggerDb::TABLE_NAME.' WHERE `timestamp` < DATE_SUB(FROM_UNIXTIME('.$timestamp.'), INTERVAL '.$archive_threshold.' DAY);');
+        }
+
+        $lockId = self::class;
+        if (!Lock::isLocked($lockId, 86400) && date('H') <= 4) {
```
NiklasBr, line 94: This hard-coded time window is not something I like. For example our QA testing instances on AWS are all on a schedule where they are shut down every night to save money. This would effectively never run on those instances. Can you make it configurable?

BlackbitNeueMedien: This is what @brusch suggested in https://github.com/pimcore/pimcore/pull/7066#issuecomment-689345674 - for sure we can make that configurable. Shutting down your servers means that current implementation of `CacheCleanupTask` does not work for you either. @brusch what do you think?

brusch: I'd keep it as it is for now, as we use the same approach also in different places. We can then create a follow up ticket where we can think about how to make that configurable or forceable.

brusch:
Thank you!

PR #7167, opened 2020-09-24
[Webservice] Quantity Value: Only use mapped unit id if a mapping exists

BlackbitNeueMedien:
Currently when the id mapper does not return an id the unit id gets set to `false` - at least in case of https://github.com/pimcore/RestImporter/blob/deb8daaa4136aa2ba7d10e154b6f918d6a4d5a3b/src/Elements/Bundle/RestImporterBundle/IdMapper.php#L49-L71). This PR keeps the transfered unit id if there isn't any id mapping.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/QuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/QuantityValue.php
@@ -537,7 +537,10 @@ class QuantityValue extends Data implements ResourcePersistenceAwareInterface, Q
             if (array_key_exists('value', $value) && array_key_exists('unit', $value) && array_key_exists('unitAbbreviation', $value)) {
                 $unitId = $value['unit'];
                 if ($idMapper) {
-                    $unitId = $idMapper->getMappedId('unit', $unitId);
+                    $mappedId = $idMapper->getMappedId('unit', $unitId);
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
May this should be fixed in the RestImporter by returning null instead of false.

Just thinking about your change ... let us assume that the remote unit ID is 8 for let's say "centimeters" but there is no mapping.
On the local instance there is a unit definition with the same ID 8 for kilograms.
So as far as I see it centimeters will be mapped to kilograms.

BlackbitNeueMedien:
Hmm, you are right. I did not want to put too much effort in this dying web service feature. The actual problem is that the RestImporter does not write anything to the table for mapped ids for quantity value units. So we synchronized those units and their ids on the source and target system to at least be able to transfer quantity values at all. 

Then we will create those mappings manually, then everything should work also with the current implementation.

weisswurstkanone:
As far as I remember you have to add a UnitDefinition importer to your job list.
Maybe this could be a starting point for you:
https://github.com/pimcore/RestImporter/blob/c1188f0a8aa09e2d39275e01feb596374a979c29/src/Elements/Bundle/RestImporterBundle/Importer/UnitDefinition.php

PR #7169, opened 2020-09-24
Extend correct document class in email hardlink wrapper

loriskoch:
In my opinion this class should extend the email document class, as all the other wrapper classes extend their own base document class too.

The current state causes errors if you want to call an email document via hardlinks. The problem is that "isDirectRouteDocument()" will always return false because it doesn't extend the PageSnippet class (pimcore/lib/Routing/Dynamic/DocumentRouteHandler.php:361).

Because of that the route never gets assigned the _controller attribute and therefore the page can't be rendered in the frontend and throws the following error:

Bildschirmfoto 2020-09-24 um 10 17 21


dvesh3:
Thanks!...kudos for your first PR 👍

PR #7170, opened 2020-09-24
[Twig] Add template helpers to get elements by path

BlackbitNeueMedien:
Currently there are only twig helpers to get elements by id: pimcore_document, pimcore_asset, pimcore_object. This can be a problem when you transmit data between Pimcore systems because then Ids may change. For this use case this PR introduces helpers to get elements by path: pimcore_document_by_path, pimcore_asset_by_path, pimcore_object_by_path

Also they are better understandable as `pimcore_object_by_path("/path/to/special/category")` is more expressive than `pimcore_object(123)`.

### DIFF
--- a/doc/Development_Documentation/02_MVC/02_Template/00_Twig.md
+++ b/doc/Development_Documentation/02_MVC/02_Template/00_Twig.md
@@ -128,16 +128,24 @@ generates HTML as otherwise the HTML will be escaped by twig.
 
 #### Loading Objects
 
-The following functions can be used to load Pimcore objects in a template (use `<type>`::getById()`):
+The following functions can be used to load Pimcore elements from within a template:
 
 * `pimcore_document`
+* `pimcore_document_by_path`
 * `pimcore_site`
 * `pimcore_asset`
+* `pimcore_asset_by_path`
 * `pimcore_object`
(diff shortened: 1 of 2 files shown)


brusch:
I'm just thinking, ... maybe we should just add path support to the existing helpers, ... should be pretty easy by checking the parameter with `is_numeric()`?

BlackbitNeueMedien:
Also thought about this. But I could imagine that some time we might change element IDs to not be numeric anymore (e.g. UUIDs). Of course then you might check for slashes in the parameter... but actually I like SRP for methods. Also a lot of effort has been put into type safety lately - e.g. with the phpstan checks. And moreover I thought of the static methods `getById()` and `getByPath()`, why should we differ in the twig helpers?
But if you give me a sign and I will change this PR so `pimcore_object` and the other helpers support also paths.

brusch:
Ok, let's see if there are other opinions😊
But I think since the naming of the current helpers is very generic (`pimcore_object()`), it would be perfect to be used as the universal tool for obtaining elements.

dpfaffenbauer:
I like having a separate method better. It gives you a clear indication of what to except from the function.

brusch:
Thank you very much! 🚀

PR #7172, opened 2020-09-24
Omit mandatory check when copying objects

bramstroker:
When using a field which is both unique and mandatory copy action will fail.
This is caused by the copy routine setting all the values of "unique" fields to null, but the object is saved with `omitMandatoryCheck` on `false`, which causes a validation exception.

Reproduce:
- Create a field and check both `Mandatory field` and `Unique`
- Now create an object and give the field a value. 
- Save the object
- Copy the object and past as child.
- Now you'll get an exception "Validation failed: Empty mandatory field [ fieldname ]"

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -149,6 +149,7 @@ class Service extends Model\Element\Service
             foreach ($new->getClass()->getFieldDefinitions() as $fieldDefinition) {
                 if ($fieldDefinition->getUnique()) {
                     $new->set($fieldDefinition->getName(), null);
+                    $new->setOmitMandatoryCheck(true);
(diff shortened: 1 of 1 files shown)


dvesh3:
This should work with changes done on v6.7.0 https://github.com/pimcore/pimcore/blob/master/models/DataObject/Concrete.php#L503 since new object publish state is set to `false` https://github.com/pimcore/pimcore/pull/7172/files#diff-ed339135ec840520ae1e3876b28c842bR153. right?

bramstroker:
@dvesh3 Yes you are right! I was still running on 6.6.9 which had the bug. It seems all fine in 6.7.2. 
Thanks and case closed!

PR #7176, opened 2020-09-25
[Class Export] Export modificationDate

blankse:
Can we export the modificationDate? It would be helpful to check whether a class needs to be imported.

The import set the modificationDate to `time()`, so it would not a problem there.


brusch:
Should be ok I guess

PR #7177, opened 2020-09-25
Add Pimcore JS events to enable virtual data folder integration.

andreas-gruenwald:
## Feature Request

We are working on a bundle that allows developers to integrate folders and objects as part of a virtual data object tree.

The idea is very similar to the [AlternateObjectTrees](https://github.com/elements-at/AlternateObjectTrees) bundle, but the implementation 
- focuses more on the developer perspective,
- and the UI integration into Pimcore's tree structure is more tightly.

Preview:
image

In order to support the integration for data object trees and the children grid view, this PR adds additional JS events that make the core more flexible.

We tried to keep the changes as minimalistic as possible.


brusch:

PR #7179, opened 2020-09-25
fixed [Ecommerce][Productindex] elasticsearch preparation queue children do not update if parent is not in index

### LINKED ISSUE #7168: [Ecommerce][Productindex] elasticsearch preparation queue children do not update if parent is not in index
In previous Pimcore versions the product index mechanism did always reset the preparation queue for all children if they are in the index.

In one of the last releases this if was added (seems like for performance reasons):
https://github.com/pimcore/pimcore/blob/master/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L361

Now the children are only updated if the parent is in the index too. I dont't think that this is a good idea as there are many constellations where this might happen.

fashxp:
fixes #7168  

`fillupPreparationQueue` got moved, so that children are updated nevertheless if parent is in index or not


fashxp:
@markus-moser @andreas-gruenwald

PR #7186, opened 2020-09-28
Delete application_logs archive tables after configurable amount of time

### LINKED ISSUE #7163: Deleting application_logs_archive tables after some configured time
Currently application_logs get archived to application_logs_archive_MM_YYYY. They never get deleted and thus bloat the size of database dumps. Of course you could delete those if you customize the call to mysqldump or have any other backup strategy. But I wonder if it is really necessary in all cases to be able to see application logs from 2 years ago for example.
What do you think of deleting those tables after a configurable time?

BlackbitNeueMedien:
Resolves #7163

By default remove application_logs_archive_MM_YYYY tables after 6 months. You can configure this time threshold in PImcore system settings.


brusch:
Thanks!

PR #7210, opened 2020-10-01
Deprecate auto view rendering

### LINKED ISSUE #6760: Deprecate Auto View Rendering with 6.7 and remove with 7.0
Since that feature was introduced for BC reasons with Zend, it is time to finally remove it and stick more to Symfony standard behaviour. WDYT?

dvesh3:
Resolves #6760

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/IndexController.php
+++ b/bundles/AdminBundle/Controller/Admin/IndexController.php
@@ -23,6 +23,7 @@ use Pimcore\Config;
 use Pimcore\Controller\Configuration\TemplatePhp;
 use Pimcore\Controller\EventedControllerInterface;
 use Pimcore\Db\ConnectionInterface;
+use Pimcore\Event\Admin\IndexActionSettingsEvent;
 use Pimcore\Event\Admin\IndexSettingsEvent;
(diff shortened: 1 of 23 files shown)


brusch (changes requested):
I think it would be good if we trigger a deprecation notice here: 
https://github.com/pimcore/pimcore/blob/2662d44a8828ad8fc7c8d878020b115f8ba61d69/bundles/CoreBundle/EventListener/TemplateControllerListener.php#L111

I think `ViewModelValueResolver` can be deprecated as well, right? 

Here https://github.com/pimcore/pimcore/blob/da18c52e7e9b902b01c13cde23568f2f63ab91ab/lib/Controller/FrontendController.php#L48 we should trigger a deprecation message as well I guess. 

Or maybe better directly in: 
`\Pimcore\Http\Request\Resolver\ViewModelResolver::createViewModel()`

brusch:
Btw. can you please also resolve the conflicts? 😊

brusch:
@dvesh3 tests are failing and unfortunately there's again a conflict 😉

dvesh3:
> @dvesh3 tests are failing and unfortunately there's again a conflict 😉

@brusch fixed :)

PR #7213, opened 2020-10-01
[Docs] Fix dead links

danmichaelo:
These should point to the demo repo, right?


brusch:
Exactly, thanks a lot!

PR #7215, opened 2020-10-01
Support for Twig in Emails and Web2Print Documents

### LINKED ISSUE #3692: Twig instead of Placeholders
Wouldn't it be way nicer to use Twig instead of the Placeholder Engine for email templates? That would allow it to have way more logic in there, such as twig helpers, ifs and loops :)

Just an idea ;)

brusch:
Resolves #3692

Replaces Placeholder support in v7.

### DIFF
--- a/doc/Development_Documentation/03_Documents/02_Document_Types/15_Print_Documents.md
+++ b/doc/Development_Documentation/03_Documents/02_Document_Types/15_Print_Documents.md
@@ -51,7 +51,7 @@ If you updated an existing Pimcore installation, please get the default implemen
 ## PDF Rendering
 
 Both web-to-print documents have an additional tab that is the place for rendering documents to print-ready PDFs. 
-When rendering, the print is first rendererd to HTML and then rendered to an PDF. PDF rendering itself is done by an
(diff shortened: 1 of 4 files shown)


doc/Development_Documentation/03_Documents/02_Document_Types/15_Print_Documents.md
```diff
@@ -51,7 +51,7 @@ If you updated an existing Pimcore installation, please get the default implemen
 ## PDF Rendering
 
 Both web-to-print documents have an additional tab that is the place for rendering documents to print-ready PDFs. 
-When rendering, the print is first rendererd to HTML and then rendered to an PDF. PDF rendering itself is done by an
+When rendering, the print is first rendererd to HTML, then rendered as a Twig template (yes, you can use Twig expressions in the document) and then rendered to an PDF. PDF rendering itself is done by an
```
dpfaffenbauer, line 54: doesn't make sense? first html, than twig than html? Better rendered from twig to html and then to pdf

brusch: Does make sense, the rendered content is again rendered as twig as it can contain twig expressions, e.g. in editables ...

PR #7231, opened 2020-10-05
Trigger LoginSuccessEvent when user is authenticated into admin

### LINKED ISSUE #4094: Post Login Admin Events
# Feature Request

Currently there is  \Pimcore\Event\AdminEvents::LOGIN_CREDENTIALS that triggers before authentication.We need an event post login like Login_Success and Login_Failure.

Actually we need this for Login Audit logs...Currently only last login time is stored in Pimcore.

NiklasBr:
Resolves #4094

### DIFF
--- a/bundles/AdminBundle/Security/Guard/AdminAuthenticator.php
+++ b/bundles/AdminBundle/Security/Guard/AdminAuthenticator.php
@@ -20,6 +20,7 @@ use Pimcore\Bundle\AdminBundle\Security\User\User;
 use Pimcore\Cache\Runtime;
 use Pimcore\Event\Admin\Login\LoginCredentialsEvent;
 use Pimcore\Event\Admin\Login\LoginFailedEvent;
+use Pimcore\Event\Admin\Login\LoginSuccessEvent;
 use Pimcore\Event\AdminEvents;
 use Pimcore\Model\User as UserModel;
 use Pimcore\Tool\Admin;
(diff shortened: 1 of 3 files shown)


bundles/AdminBundle/Security/Guard/AdminAuthenticator.php
```diff
@@ -252,6 +253,9 @@ public function getUser($credentials, UserProviderInterface $userProvider)
             }
         }
 
+        // trigger LOGIN_SUCCESS event if user has been fully authenicated
+        $this->dispatcher->dispatch(AdminEvents::LOGIN_SUCCESS, new LoginSuccessEvent($user));
```
dvesh3, line 257: this is not the right place for success event as it will be triggered on every getUser() call. please see https://symfony.com/doc/current/security/guard_authentication.html#step-2-create-the-authenticator-class

Please move it to the right place considering scenarios like 2FA, deeplink etc. Thanks!

dpfaffenbauer:
since Pimcore uses the Symfony Firewall, doesn't Symfony already trigger events?

dvesh3:
> since Pimcore uses the Symfony Firewall, doesn't Symfony already trigger events?

yes, but then you might not get sufficient information from passed event argument.

dpfaffenbauer:
> > since Pimcore uses the Symfony Firewall, doesn't Symfony already trigger events?
> 
> yes, but then you might not get sufficient information from passed event argument.

What do you except to get? The User right? Which is what you get....

dvesh3:
@NiklasBr How about using `security.interactive_login`? https://symfony.com/doc/current/components/security/authentication.html#authentication-events

NiklasBr:
I'll look into it @dvesh3

dvesh3:
@NiklasBr any luck here?

NiklasBr:
Not had the time to properly do it yet, still on the table though.

dvesh3:
@NiklasBr closed due to inactivity. feel free to reach out to continue working on this in future 😉 thanks!

PR #7223, opened 2020-10-02
Deprecated Tool\Lock, created Symfony Lock factory service

brusch:
Fixes #7200

PR #7233, opened 2020-10-05
[Maintenance] getLastExecution() should check if TmpStore item exists

dvesh3:
Fixes exception "getData() on null" inside getLastExecution()

## Additional info  
Regression of #7223


brusch:
Merci! 👌😊

PR #7247, opened 2020-10-07
Removed ViewModel and Auto-Rendering

### LINKED ISSUE #7199: Remove ViewModel and Auto-Rendering
Follow up to #6760

brusch:
resolves #7199 
related to #6760


dpfaffenbauer:
nice, what do you think if we also add this for $document and $editmode: https://symfony.com/doc/current/controller/argument_value_resolver.html

I would create a PR for that if you agree

brusch:
@dpfaffenbauer 
https://github.com/pimcore/pimcore/blob/c62924ca5c3d0c482a71097cb0231aff143208d6/lib/Controller/ArgumentValueResolver/EditmodeValueResolver.php#L22
https://github.com/pimcore/pimcore/blob/c62924ca5c3d0c482a71097cb0231aff143208d6/lib/Controller/ArgumentValueResolver/DocumentValueResolver.php#L26

There since Pimcore 5.0 😉

dpfaffenbauer:
alright then :D, never used them. good to know :)

PR #7249, opened 2020-10-07
Grid filters null check

t-zilla:
Add a null check to gridTabAbstract. It crashes if filters isn't defined in this.grid yet.


brusch:
Thank you! 👍

PR #7255, opened 2020-10-07
Fixed Monolog Bundle incompatibility cased by 3.6.0 release of symfony/monolog-bundle

brusch:
Set fixed constraint to `"symfony/monolog-bundle": "3.5.0"` 
Seems to be only an issue with Symfony 3.4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -82,7 +82,7 @@
     "sensiolabs/ansi-to-html": "^1.1",
     "symfony-cmf/routing-bundle": "^2.0",
     "symfony/contracts": "^1.1",
-    "symfony/monolog-bundle": "^3.1.0",
+    "symfony/monolog-bundle": "3.5.0",
     "symfony/swiftmailer-bundle": "^3.2.2",
     "symfony/symfony": "^3.4.26 || ^4.1.12",
     "tijsverkoyen/css-to-inline-styles": "^2.2.1",


dpfaffenbauer:
Is that related to this error?

```
Fatal error: Uncaught Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: Unused binding "Symfony\Contracts\HttpClient\HttpClientInterface" in service "instanceof.Monolog\Handler\HandlerInterface.0.Pimcore\Log\Handler\ApplicationLoggerDb". in /home/runner/work/CoreShop/CoreShop/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Compiler/ResolveBindingsPass.php on line 50

Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: Unused binding "Symfony\Contracts\HttpClient\HttpClientInterface" in service "instanceof.Monolog\Handler\HandlerInterface.0.Pimcore\Log\Handler\ApplicationLoggerDb". in /home/runner/work/CoreShop/CoreShop/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Compiler/ResolveBindingsPass.php on line 50
```

brusch:
Yep, see also: https://github.com/symfony/monolog-bundle/issues/365

dpfaffenbauer:
the same happens for 6.7 too now: https://github.com/coreshop/CoreShop/pull/1497/checks?check_run_id=1225039656#step:13:9

brusch:
Yep, needs to be fixed on project basis then if you're not upgrading to `6.7`. 
It's anyway not an Pimcore issue, the issue is on the monolog bundle.

dpfaffenbauer:
yep 👍

dpfaffenbauer:
this would also do the trick:

```yml
    Pimcore\Log\Handler\ApplicationLoggerDb:
        public: true
        autowire: false
        autoconfigure: false
        calls:
            - [pushProcessor, ['@monolog.processor.psr_log_message']]
            - [pushProcessor, ['@Pimcore\Log\Processor\ApplicationLoggerProcessor']]
            - [pushProcessor, ['@pimcore.app_logger.introspection_processor']]
```

not sure if it has any side-effects, but I don't think so.

also, a conflict with monolog-bundle:3.6 might also work

jdreesen:
A conflict with `v3.6.0` would have been the better fix than hard-requiring `v3.5.0`, imho.

brusch:
They will drop support for 3.4 anyway, so fix is ok.

dpfaffenbauer:
yep, but they will do with 3.7, and that means you have to change the composer file again, with a conflict this would work automatically.

brusch:
Nothing needs to be changed again (3.6 doesnt work and 3.7 wouldn work anyway with 3.4, so what?), but whatever ....

dpfaffenbauer:
yeah, true.

jdreesen:
By pinning it to `v3.5.0` you force Pimcore users to stay on that exact version while they could profit from improvements in newer versions when they're on Symfony 4, if it still was `^3.1.0` with a conflict for `v3.6.0` ;)

brusch:
PRs are welcome

jdreesen:
#7268

PR #7268, opened 2020-10-08
[Followup] Fixed Monolog Bundle incompatibility caused by 3.6.0 release of symfony/monolog-bundle

jdreesen:
Only the `v3.6.0` is incompatible, so only this one should be prevented from being installed.

Followup of: #7255

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -82,7 +82,7 @@
     "sensiolabs/ansi-to-html": "^1.1",
     "symfony-cmf/routing-bundle": "^2.0",
     "symfony/contracts": "^1.1",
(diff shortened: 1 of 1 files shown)


brusch:
That was quick 😉 - Thanks a lot!

BlackbitNeueMedien:
@brusch Can we backport this or #7255 change also for 5.x versions? Because currently it is not possible to upgrade a 5.x pimcore to the latest 5.x version.

brusch:
@BlackbitNeueMedien no, unfortunately not, v5 is EOL. You have to make this setting in your root composer.json.

PR #7277, opened 2020-10-09
Get data for 1.1.1970 00:00, too

ThomasKeil:
Check on timestamp "0" would return null although it's valid.


dpfaffenbauer:
you somewhat messed up the PR, but I would only check for `null === $data`.

ThomasKeil:
It's indeed messed up, although I don't understand why. I'll close it.

jdreesen:
You created the patch on `master` but selected `6.8` as target ;)

PR #7285, opened 2020-10-12
[Core] Do not statically set error reporting level

### LINKED ISSUE #5311: [Core] Do not statically set error reporting level
see also #5310

PHP setting (default behavior) should be used instead.

blankse:
Resolves #5311


brusch:
Nice, thanks a lot! 👍

PR #7286, opened 2020-10-12
Only go to current working directory if it is accessible

BlackbitNeueMedien:
In https://github.com/pimcore/pimcore/blob/1430b30519dd28af41c938944eb431ae2593e01c/lib/Bootstrap.php#L76 it is tried to change directory to current working directory. This can be a problem when user `abc` calls a command with `sudo -u xyz bin/console command:name` when user `xyz` does not have access to the directory from which the script was called (e.g. user `abc`'s home folder).


brusch:
Thanks!

PR #7289, opened 2020-10-12
[Data Objects] Support selecting multiple items for relational fields

BlackbitNeueMedien:
Currently in object edit mode you can only select 1 referenced item for a relational field at the same time. This PR adds multi-select. This is for example useful when you want to move multiple objects via drag & drop. Multiple items can be selected via Shift or Ctrl key + mouseclick.


brusch:
Makes absolutely sense, thank you so much! 👌

PR #7291, opened 2020-10-12
[Roles] Sort permissions by label

### LINKED ISSUE #7245: User Permissons - Sorting
The permissions in the user settings tab should be sorted alphabetically, so they "stick together" when you use a custom prefix for your own bundle.

BlackbitNeueMedien:
Currently permissions are not sorted at all, making it difficult to find a certain one (especially if there are some plugin- / project-specific ones added or if the current user's language is not english):
Bildschirmfoto 2020-10-12 um 15 00 18

This PR sorts permissions based on their (translated) labels.

Resolves #7245


brusch:
Thank you very much!

PR #7296, opened 2020-10-12
[Core] Use ElementInterface instead of AbstractElement

blankse:
In some places a check or type hint is made for the AbstractElement class instead of the ElementInterface. 

With a major release coming up now, I think we could change that.


brusch:
Thanks a lot!

PR #7299, opened 2020-10-13
[Element] Added setProperties() and setProperty() to ElementInterface

brusch:
Follow up to #7296

### DIFF
--- a/models/Element/ElementInterface.php
+++ b/models/Element/ElementInterface.php
@@ -17,6 +17,7 @@
 
 namespace Pimcore\Model\Element;
 
+use Pimcore\Model\Asset;
 use Pimcore\Model\Dependency;
 use Pimcore\Model\ModelInterface;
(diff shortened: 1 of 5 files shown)


models/Element/ElementInterface.php
```diff
@@ -154,6 +162,16 @@ public function getProperties();
      */
     public function getProperty($name, $asContainer = false);
 
+    /**
+     * @param $name
+     * @param $type
+     * @param $data
+     * @param false $inherited
+     * @param false $inheritable
+     * @return mixed
```
jdreesen, line 171:
```suggestion
     * @param string $name
     * @param string $type
     * @param mixed $data
     * @param bool $inherited
     * @param bool $inheritable
     *
     * @return $this
```

brusch: Thanks!

PR #7308, opened 2020-10-13
[Data Objects] Removed BC Layers in Data-Types

### LINKED ISSUE #6535: remove method_exists($value, 'getDataForResource') calls
As already mentioned in the upgrade guide 5 to 6 custom datatype has to implement the 
ResourcePersistenceAwareInterface

https://github.com/pimcore/pimcore/blob/1bac46d5626babf2fdabc4fb2ec20300b1bba152/doc/Development_Documentation/23_Installation_and_Upgrade/07_Updating_Pimcore/01_V5_to_V6.md#L115

see #4378

brusch:
Resolves #6535


models/DataObject/ClassDefinition/Dao.php
```diff
@@ -159,9 +159,9 @@ public function update()
         // add non existing columns in the table
         if (is_array($this->model->getFieldDefinitions()) && count($this->model->getFieldDefinitions())) {
             foreach ($this->model->getFieldDefinitions() as $key => $value) {
-                if ($value instanceof DataObject\ClassDefinition\Data\ResourcePersistenceAwareInterface || method_exists($value, 'getDataForResource')) {
+                if ($value instanceof DataObject\ClassDefinition\Data\ResourcePersistenceAwareInterface) {
                     // if a datafield requires more than one column in the datastore table => only for non-relation types
-                    /** @var Data&DataObject\ClassDefinition\Data\ResourcePersistenceAwareInterface $value */
+                    /** @var DataObject\ClassDefinition\Data\ResourcePersistenceAwareInterface $value */
```
blankse, line 164: @brusch These comments can now also be removed. PHPStan and IDEs should be know it with the if statement above.

brusch: 👍thanks a lot!

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -477,11 +477,7 @@ public function filterMultipleAssignments($data, $container, $params)
                 } elseif (!isset($relationItems[$elementHash])) {
                     $relationItems[$elementHash] = $item;
                 } else {
-                    @trigger_error(
-                            'Passing relations multiple times is deprecated since version 6.5.2 and will throw exception in 7.0.0, tried to assign ' . $elementHash
-                            . ' multiple times in field' . $fieldName . ' of object id: ' . $objectId,
-                            E_USER_DEPRECATED
-                        );
+                    throw new \Exception(sprintf('Passing relations multiple times is not allowed: %s in field %s of object id: %s', $elementHash, $fieldName, $objectId));
```
blankse, line 480: @brusch Is there a plan for custom exceptions in Pimcore 7? It would be good to have different exception classes for catching

brusch: @blankse yes, we can think about that where it makes sense 👍

PR #7309, opened 2020-10-13
[Data Objects] Removed getO_*() and setO_*() BC layer

brusch:


brusch:
That's really a bit tricky now 😊 Let's see maybe I find the issue tomorrow 😅

brusch:
Should be done together with #7309

Issue #7192, opened 2020-09-29
[Data Object] Modifing an object having a fieldcollection with inside a localized fields delete the localized values


Yivan:
Hello,

This is a critical issue as it impact several Pimcore versions (at least from 6.6.4 to 6.7.x-dev) and **lead to lost datas silently**. It could impact many project in production.

It is easily reproducible on online demo, and this is a pretty classic setup on multilingual project.

Create an object with a field collection:

![image](https://user-images.githubusercontent.com/9025839/94551404-9928dd80-0255-11eb-818c-b934800c9945.png)

Create a field collection:

![image](https://user-images.githubusercontent.com/9025839/94551432-a5ad3600-0255-11eb-8393-e0f291616c59.png)

Create an instance of this object, and fill some values and save&publish:

![image](https://user-images.githubusercontent.com/9025839/94551524-c4abc800-0255-11eb-80c3-60888fcfe6e4.png)

Close and reopen the object, and change the value in the text field, save&publish, and reload the object, **the multihref in localized field has been emptied**:

![image](https://user-images.githubusercontent.com/9025839/94551641-f4f36680-0255-11eb-9641-ddc4b249667a.png)

This affect production project for now (we are on 6.6.4 on production projects) and last Pimcore version.

Thanks.

Note : we encountered the same problem with localized fields inside an object brick for information. After investigation, it seems only one table in the database get updated, the other not (localized fields are in two table in the database in case of them inside an object brick). Maybe can be a hint for this issue.

brusch:
Fixed by #7197

weisswurstkanone:
@Yivan , I believe there was a major change via #6612  which was done after 6.6.4

Yivan:
Thanks very much @weisswurstkanone for your answer on this subject.

I just try, the only line changing is the 160 : https://github.com/pimcore/pimcore/pull/6612/files#diff-16f9a3cbf890a5bd4d3ba500c90e2478R160
(the just below part was already modified from the code of this new commit from this issue, I have take all the if block of this fix).

Sadly problem is always here with this second commit backported : (
I tried disably lazyloading, change nothing.

[...]

PR #7318, opened 2020-10-14
Untouched relations get lost when contained in localizes fields in object bricks on save

### LINKED ISSUE #7302: Untouched relations get lost when contained in localizes fields in object bricks on save
# Bug Report
If you have a relation field (e.g. Many-To-One) in an localized field in an object brick and you are editing the content of this relation in one locale while leaving the other locales untouched, the content of the other locales is lost on saving the object.

### Expected behavior
If the content for one locale of a relation field in a localized object brick is changed, the content of the other locales should not get lost, when untouched.

### Actual behavior
The content of the other locales is lost.

### Steps to reproduce
Create an object brick with localized fields and a relation attribute (e.g. Many-To-One). Add this object brick to a class. Create an object of that class. Add the object brick to that object and fill the relation attribute for one locale. Save the object. Now reopen the object and fill the relation attribute for a second locale. Save again. After reloading the object, the content of the first locale is gone.

## Possibly related:
https://github.com/pimcore/pimcore/issues/7059 But the commit that solves the issue there, does not resolve the bug as described here.

weisswurstkanone:
resolves #7302

### DIFF
--- a/models/DataObject/Localizedfield/Dao.php
+++ b/models/DataObject/Localizedfield/Dao.php
@@ -135,12 +135,11 @@ class Dao extends Model\Dao\AbstractDao
          */
         DataObject\Concrete\Dao\InheritanceHelper::setUseRuntimeCache(true);
         foreach ($validLanguages as $language) {
-            if ((!isset($params['newParent']) || !$params['newParent'])
-                && isset($params['isUpdate'])
-                && $params['isUpdate']
+            if (empty($params['newParent'])
+                && !empty($params['isUpdate'])
                 && !$this->model->isLanguageDirty($language)
-                 && !$forceUpdate
-                ) {
+                && !$forceUpdate
+            ) {
                 continue;
             }
             $inheritedValues = DataObject\AbstractObject::doGetInheritedValues();
--- a/models/DataObject/Localizedfield/Dao.php
+++ b/models/DataObject/Localizedfield/Dao.php
@@ -135,9 +135,12 @@ class Dao extends Model\Dao\AbstractDao
          */
         DataObject\Concrete\Dao\InheritanceHelper::setUseRuntimeCache(true);
         foreach ($validLanguages as $language) {
-            if ((!isset($params['newParent']) || !$params['newParent']) && isset($params['isUpdate']) && $params['isUpdate'] && !$this->model->isLanguageDirty(
-                    $language
-                )) {
+            if ((!isset($params['newParent']) || !$params['newParent'])
+                && isset($params['isUpdate'])
+                && $params['isUpdate']
+                && !$this->model->isLanguageDirty($language)
+                 && !$forceUpdate
+                ) {
                 continue;
             }
             $inheritedValues = DataObject\AbstractObject::doGetInheritedValues();
--- a/tests/Model/LazyLoading/ManyToManyRelationTest.php
+++ b/tests/Model/LazyLoading/ManyToManyRelationTest.php
@@ -342,8 +342,16 @@ class ManyToManyRelationTest extends AbstractLazyLoadingTest
         //prepare data object
         $object = $this->createDataObject();
         $brick = new LazyLoadingLocalizedTest($object);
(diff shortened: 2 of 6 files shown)


models/DataObject/Localizedfield/Dao.php
```diff
@@ -135,9 +135,12 @@ public function save($params = [])
...
         foreach ($validLanguages as $language) {
-            if ((!isset($params['newParent']) || !$params['newParent']) && isset($params['isUpdate']) && $params['isUpdate'] && !$this->model->isLanguageDirty(
-                    $language
-                )) {
+            if ((!isset($params['newParent']) || !$params['newParent'])
+                && isset($params['isUpdate'])
+                && $params['isUpdate']
+                && !$this->model->isLanguageDirty($language)
+                 && !$forceUpdate
+                ) {
```
jdreesen, line 143: `!isset($params[...]) || !$params[...]` can be simplified to `empty($params[...])` 
and
`isset($params[...]) && $params[...]` can be simplified to `!empty($params[...])` 

```suggestion
            if (empty($params['newParent'])
                && !empty($params['isUpdate'])
                && !$this->model->isLanguageDirty($language)
                && !$forceUpdate
            ) {
```

dold:
@weisswurstkanone  I just tried to patch our pimcore version (6.6.10) with the fix in models/DataObject/Localizedfield/Dao.php, but it still does not work. Can you give me any hint, if other parts of the code need to be updated? I can see, that there is no "isUpdate" attribute in the params array. And the $forceUpdate variable is not defined at all in the save method.
 
We have this as a quite pressing issue with a customer, but we cannot update the pimcore version right now to the latest release.

weisswurstkanone:
No? https://github.com/pimcore/pimcore/blob/41c6a9106110449816aae6fcbfbdfaea1cbf0b7f/models/DataObject/Localizedfield/Dao.php#L101

weisswurstkanone:
isUpdate set flag is set here
https://github.com/pimcore/pimcore/blob/4304973c91f2587be57ad683312e4cf82827a0c2/models/DataObject/Objectbrick/Data/Dao.php#L125

but this was there already before 6.6.10

dold:
The $forceUpdate variable is not in 6.6.10. Let me check the isUpdate param.

weisswurstkanone:
the forceUpdate flag was added 2 weeks ago. see https://github.com/pimcore/pimcore/commit/ebf150676da3e4da96866ad3bee1b5229dee4e43

the isUpdate param was there before

dold:
Thanks. I guess there are then other files that also need to change, so that this is working. We are planning now an update to 6.8.2 and see if we can patch it there.

tests/Model/LazyLoading/ManyToManyRelationTest.php
```diff
@@ -342,15 +342,33 @@ public function testLocalizedBrickAttributes()
...
+        $brick->getLocalizedfields()->setLocalizedValue('lrelations', $this->loadRelations()->load(), 'de');
 
         $object->getBricks()->setLazyLoadingLocalizedTest($brick);
         $object->save();
 
         $brick->setLInput(uniqid());
         $object->save();
         $object = Concrete::getById($object->getId(), true);
-        $this->assertTrue(count($object->getBricks()->getLazyLoadingLocalizedTest()->getLRelations()) > 0);
+        $this->assertTrue(count($object->getBricks()->getLazyLoadingLocalizedTest()->getLRelations('en')) > 0);
```
dvesh3, line 354: test for other relation types?..at-least for many-to-one?

weisswurstkanone: @dvesh3 added a test for many to one

dvesh3 (approved):

Yivan:
@dold We get the problem to backport manually a fix on the same subject (see : https://github.com/pimcore/pimcore/issues/7192#issuecomment-702227966 ) it is linked I Think (fix has been availbale only starting from v6.7.3).
But we didn't manage to backport it... : ( So we have no choice to update Pimcore.
If you have any success I would be glad you give some hints on it !
Best regards.

PR #7335, opened 2020-10-15
[Ecommerce Framework] IntlFormatter tried to resolve locale before LocaleListener

BlackbitNeueMedien:
In https://github.com/pimcore/pimcore/blob/33c2039c2e0d716c04ca05479f2c399b7a20a19f/lib/Localization/IntlFormatter.php#L62-L65
the locale to be used for number formatting is tried to be found. But in this moment the `Symfony\Component\HttpKernel\EventListener\LocaleListener` has not run yet and so the `LocaleService` cannot get the current request's language in https://github.com/pimcore/pimcore/blob/33c2039c2e0d716c04ca05479f2c399b7a20a19f/lib/Localization/LocaleService.php#L63-L75

This PR only injects the `LocaleService` into `IntlFormatter` and as soon as a locale is necessary it the `LocaleService`gets asked. 

This bug is for example reproducable with a static route which supports `_locale` parameter.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetHelperController.php
@@ -278,7 +278,7 @@ class AssetHelperController extends AdminController
         $settings['shareGlobally'] = $sharedGlobally ?? null;
         $settings['isShared'] = !$gridConfigId || ($shared ?? null);
 
-        $context = $gridConfig['context'];
+        $context = $gridConfig['context'] ?? null;
         if ($context) {
             $context = json_decode($context, true);
         }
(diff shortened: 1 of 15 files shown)


brusch:
I'd classify this as a bug, can you please rebase onto `6.8`? Thanks a lot! 😊

BlackbitNeueMedien:
Always have problems with rebasing. Created new branch based on 6.8 and redid changes and then force-pushed to bugfix/intlformatter-locale. Is this ok?

brusch:
Sure, just changed the target branch of the PR, seems to be fine now 😊 Thanks!

blankse:
@BlackbitNeueMedien I think the commit https://github.com/pimcore/pimcore/pull/7335/commits/131d5578aa45fd9b727a6a1d3816bdeba74ad330 should be removed.

Otherwise it will https://github.com/pimcore/pimcore/pull/7345 reverted.

BlackbitNeueMedien:
Thanks @blankse, fixed that.

brusch:
Thanks a lot! 🙏

PR #7340, opened 2020-10-16
[Translations] Add alias to "bundles" key for plugins, extensions

vmalyk:
Added and updated aliases to "bundles" key in translations. 

## Additional info
In Permissions for User I still see "Extensions" menu item and for easy match current menu item with right permission definition I suggest to use the same title.    

Please review,


brusch:
Thank you 😊

PR #7345, opened 2020-10-16
[Asset] Fix Stream error when saving an asset to an S3 bucket

### LINKED ISSUE #6339: Stream error when saving an asset to an S3 bucket
# Bug Report

### Expected behavior
Assets are saved to the S3 bucket and the corresponding element is registered in the database without errors.

### Actual behavior
This message is showing on the admin interface everytime I'm trying to save an asset.
MicrosoftTeams-image

After saving or creating an asset I found that the asset was uploaded/updated on the S3 bucket but the UI kept showing the previous error. 
When creating a new asset I found that a row was not created in the asset table. 
While debugging the logs showed that the database was being rolled back after the error was thrown. 

This is the stack trace of the error:

MicrosoftTeams-image(2)
MicrosoftTeams-image(1)

### Steps to reproduce
Setup:
- Amazon S2 for Pimcore with Apache
- Amazon S3 for Assets as per the documentation
https://pimcore.com/docs/master/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/Amazon_AWS_Setup/Amazon_AWS_S3_Setup.html

### Code example
I found that the result of the stream_get_meta_data method was returning a closed resource.
I added the following validation to the vendor/pimcore/pimcore/models/Asset.php class on the getStream method.

get_resource_type($this->stream) != "stream"

**Original Code** ([source](https://github.com/pimcore/pimcore/blob/master/models/Asset.php))
```php
    /**
     * @return resource
     */
    public function getStream()
    {
        if ($this->stream) {
            $streamMeta = stream_get_meta_data($this->stream);
            if (!@rewind($this->stream) && $streamMeta['stream_type'] === 'STDIO') {
                $this->stream = null;
            }
        }

        if (!$this->stream && $this->getType() != 'folder') {
            if (file_exists($this->getFileSystemPath())) {
                $this->stream = fopen($this->getFileSystemPath(), 'r', false, File::getContext());
            } else {
                $this->stream = tmpfile();
            }
        }

        return $this->stream;
    }
```

**Fix**
```php
    /**
     * @return resource
     */
    public function getStream()
    {
        if ($this->stream) {
           if(get_resource_type($this->stream) != "stream"){
	    	    $this->stream = null;
            }else{
                $streamMeta = stream_get_meta_data($this->stream);
                if (!@rewind($this->stream) && $streamMeta['stream_type'] === 'STDIO') {
                    $this->stream = null;
                }
            }
        }

        if (!$this->stream && $this->getType() != 'folder') {
            if (file_exists($this->getFileSystemPath())) {
                $this->stream = fopen($this->getFileSystemPath(), 'r', false, File::getContext());
            } else {
                $this->stream = tmpfile();
            }
        }
	
        return $this->stream;
    }
```

dvesh3:
Resolves #6339

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1295,9 +1295,13 @@ class Asset extends Element\AbstractElement
     public function getStream()
     {
         if ($this->stream) {
-            $streamMeta = stream_get_meta_data($this->stream);
-            if (!@rewind($this->stream) && $streamMeta['stream_type'] === 'STDIO') {
+            if (get_resource_type($this->stream) != "stream") {
                 $this->stream = null;
+            } else {
+                $streamMeta = stream_get_meta_data($this->stream);
+                if (!@rewind($this->stream) && $streamMeta['stream_type'] === 'STDIO') {
+                    $this->stream = null;
+                }
             }
         }
 
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1295,7 +1295,7 @@ class Asset extends Element\AbstractElement
     public function getStream()
     {
         if ($this->stream) {
-            if (get_resource_type($this->stream) != "stream") {
+            if (get_resource_type($this->stream) !== "stream") {
                 $this->stream = null;
             } else {
                 $streamMeta = stream_get_meta_data($this->stream);
--- a/models/Asset/Service.php
+++ b/models/Asset/Service.php
@@ -374,9 +374,9 @@ class Service extends Model\Element\Service
                 $instance = $loader->build($item['type']);
 
                 if ($mode == 'grid') {
-                    $transformedData = $instance->getDataFromListfolderGrid($item['data'], $item);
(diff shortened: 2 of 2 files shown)


models/Asset.php
```diff
@@ -1295,9 +1295,13 @@ public function setData($data)
     public function getStream()
     {
         if ($this->stream) {
-            $streamMeta = stream_get_meta_data($this->stream);
-            if (!@rewind($this->stream) && $streamMeta['stream_type'] === 'STDIO') {
+            if (get_resource_type($this->stream) != "stream") {
```
aarongerig, line 1298: Maybe use strict comparison (`!==`) here?

dvesh3: Done.

dpfaffenbauer:
@dvesh3 do you mind read? Just had this Problem today and wanted to create a PR on monday :)

PR #7349, opened 2020-10-17
Allow Doctrine Inflector v2 & Common v3

jdreesen:
Doctrine Common v3 has [no relevant breaking changes](https://github.com/doctrine/common/releases/tag/3.0.0) (apart from dropping some dependencies that must now be required directly, which we already do).

Doctrine Inflector v2 [drops the legacy static API](https://github.com/doctrine/inflector/releases/tag/2.0.0).

---

I would have targeted the v6.9 because this is not a bugfix, yet people may befinit from it prior to the v7 release, but it seems no branch exists yet.


brusch:
@jdreesen thanks a lot! can you please check the tests, something is going wrong there it seems 😊

jdreesen:
I pushed some fixes.

Not sure how the PHPStan failures are related...

brusch:
Thanks a lot!

brusch:
Related to #7391

PR #7356, opened 2020-10-19
[EcommerceFramework][Hobex-PaymentProvider] bugfixes

rezoti:
This PR fixes some bugs in Hobex Payment Provider: 

- unlike stated in Hobex documentation, merchantTransactionId only allows numeric values (N20) => use customParameters for internalTransactionId and create numeric merchantTransactionId from internalTransactionId 

- result code 000.000.000 should be considered as successfull response code too (see https://hobex.docs.oppwa.com/reference/resultCodes )

- returned status (authorized or cleared) depends on intitial payment Type - for (at time of this PR only possible) paymentTpye 'DB' Status 'cleared' will be returned, status 'authorized' is not correct as payment is already cleared 

- instead of checkout id, paymentId should be used as payment Reference in Payment Info (checkout id is stored in additional provider data, so it won't get lost ) . Checkout it is never needed again afterwards, but payment Id is needed in order to do reversals, refunds and so on 

## Additional info  
method createMerchantId  can be hopefully solved better if payment is refactored in pimcore 7 - but with the current structure, I couldn't find a better solutation for creating a unique, but reproducable integer-based transaction id


bundles/EcommerceFrameworkBundle/PaymentManager/Payment/Hobex.php
```diff
@@ -257,7 +269,7 @@ public function handleResponse($response)
 
             $this->logger->debug('Received JSON response in ' . self::class . '::handleResponse', $jsonResponse);
 
-            $internalPaymentId = $jsonResponse['merchantTransactionId'];
+            $internalPaymentId = $jsonResponse['customParameters']['internalTransactionId'];//$jsonResponse['merchantTransactionId'];
```
brusch, line 272:
```suggestion
            $internalPaymentId = $jsonResponse['customParameters']['internalTransactionId'];
```

brusch (changes requested):

brusch:
Thanks a lot! 🙏

PR #7360, opened 2020-10-20
"owner missing" error after upgrading from 6.0.x/6.6.9 to current 6.7.2

### LINKED ISSUE #7226: "owner missing" error after upgrading from 6.0.x/6.6.9 to current 6.7.2
# Bug-Report

We updated 2 pimcore instances (one from 6.0.x and one from 6.6.9) to the current version of pimcore.
After this update the "owner missing" error appears, when opening an object with blocks.
After a "cache clear" the error disappears for a random time, but then comes back again.

Maybe a regression of #6280 ?

weisswurstkanone:
resolves #7226

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Block.php
+++ b/models/DataObject/ClassDefinition/Data/Block.php
@@ -1063,7 +1063,7 @@ class Block extends Data implements CustomResourcePersistingInterface, ResourceP
         if ($object instanceof DataObject\Concrete) {
             $data = $object->getObjectVar($this->getName());
             if ($this->getLazyLoading() && !$object->isLazyKeyLoaded($this->getName())) {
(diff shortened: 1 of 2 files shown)


models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -1058,6 +1058,8 @@ public function delete($object, $params = [])
     public function preGetData($object, $params = [])
     {
         $data = null;
+        $params['owner'] = $object;
+        $params["fieldname"] = $this->getName();
```
dvesh3, line 1062: Issue still exists as params are not passed in load() call below.

dvesh3 (changes requested):
Also fix for Blocks inside localizedfields container

PR #7368, opened 2020-10-20
[Permissions] Allow setting permission category after creating it

BlackbitNeueMedien:
Currently the code 
```php
$permission = \Pimcore\Model\User\Permission\Definition::create('my_permission');
$permission->setCategory('my category');
$permission->save();
```
does result in a permission `my_permission` but without category `my category`. This PR changes that by allowing to set a permission category after creating it.


brusch:
Thanks a lot! 😊

PR #7370, opened 2020-10-21
Correct file path to fosjsrouting router.js

nicklassandell:
## Issue
Pull #6146 introduced an issue that causes editmode to break in the admin if you are using asset versioning. A file path is generated using `$this->package->getUrl()` which also includes asset versioning (if used) in the path, and since the file is fetched with `file_get_contents()` at one point, this fails because `file_get_contents()` does not understand the query param appended to the path.

## Steps to reproduce
Add this to your `app/config/config.yml` file and load a document in editmode.

```
  framework:
      assets:
          version: 'xxx'
```

You should get an error from file_get_contents() failing to read from a path like this:

`/var/www/html/web/bundles/fosjsrouting/js/router.js?23471` <-- see query param

## Solution
Using a relative path (like the other files included) seems to work just fine. I also commented on the [original PR](https://github.com/pimcore/pimcore/pull/6146#issuecomment-712938852) to check if the contributor (@dpfaffenbauer) knew the reason for the `getUrl()` call, but he's not sure anymore.

### DIFF
--- a/bundles/CoreBundle/EventListener/Frontend/EditmodeListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/EditmodeListener.php
@@ -318,7 +318,7 @@ class EditmodeListener implements EventSubscriberInterface
     {
         return array_merge(
             [
-                $this->package->getUrl('bundles/fosjsrouting/js/router.js'),
+                '/bundles/fosjsrouting/js/router.js',
                 '/bundles/pimcoreadmin/js/pimcore/functions.js',
                 '/bundles/pimcoreadmin/js/pimcore/overrides.js',
                 '/bundles/pimcoreadmin/js/pimcore/tool/milestoneslider.js',


dpfaffenbauer:
There was a reason why I used package->getUrl here, but I can't remember anymore. Might be so it prefixes the right path (eg. app_test.php) if that was your entrypoint.

brusch:
Maybe we should just trim off query params?

nicklassandell:
Is there a reason why it's treated differently to the rest of the files included? What difference does it make if there's a different entry point to the app?

brusch:
@nicklassandell not that I know of, but @dpfaffenbauer thinks that there was a reason 🤔

dpfaffenbauer:
@brusch @nicklassandell I will check next week for the reason. I can't really remember anymore, but I'll let you know

brusch:
@dpfaffenbauer thanks a lot!

dpfaffenbauer:
I can confirm, this is fine. I have no idea anymore why I added it like that....

nicklassandell:
That's great, thanks for looking into it @dpfaffenbauer!

I'm still waiting for my CLA to be checked, but once that's approved I would guess it's good to be merged?

brusch:
@nicklassandell you have to sign the CLA by clicking the button. Please ensure the email address you use in your local Git config (you can check this here: https://patch-diff.githubusercontent.com/raw/pimcore/pimcore/pull/7370.patch)  is also configured on your GitHub profile. Thanks!

brusch:
Closed in favor of #7459

nicklassandell:
Thanks @brusch the email was the culprit. I've gone ahead and created a new PR with the email corrected #7459

nicklassandell:
Wow that was fast, thanks. 👍

PR #7374, opened 2020-10-21

Pi-Solution:
"Pimcore\Bundle\EcommerceFrameworkBundle\FilterService\FilterType\ElasticSearch\Input.php" currently has error where in line 25 return object insted of string, that later cause error  "Illegal offset type" at line 28 and 38. Fix that I implemented is nothing more that copy of code from other FilterType.


dvesh3:
@Pi-Solution thanks for your PR. please sign the CLA 😉

PR #7377, opened 2020-10-21

Pi-Solution:
Fixed the issue regarding FilterType input for elasticSearch in old one value that is returner is Object instead of the string. This cause offset type error later in line 28 ans 38


fashxp:
hi, since this is a bugfix please rebase to 6.8 branch

PR #7379, opened 2020-10-22

Pi-Solution:
Fix from my previous pull request that you told me marge to 6.8 instead of master
Pull request fix line 25 returning object instead of string, this later cause offset bug at line 28 and 38


fashxp:
thank you so much!!!

PR #7380, opened 2020-10-22
[Maintenance] Fix versions cleanup stacktrace db task

ssiergl:
Newly introduced VersionsCleanupStackTraceDbTask is collecting all table entries which have stackTrace != NULL

On big environments (e.g. we have multiple customers with about 4 million versions) the current implementation will eat up all RAM resources already on collecting all versions at once from database.

As well if you have customers which generate a high number of versions per week this can be a problem.

Just collect the data in smaller packages to avoid this.

Regression from https://github.com/pimcore/pimcore/commit/9da5a11bfb62592a1e4976114bb45e47513f1576


ssiergl:
@dvesh3

dvesh3:
thanks a lot!

PR #7386, opened 2020-10-23
[Maintenance] fix parallel execution of maintenance tasks

ssiergl:
`$lock->isAquired` respects only aquired locks from the SAME process.
This will never happen. Like the maintenance task is ment it should check over
all running proccesses (tasks) if a lock with the process name exists and skip

See as well Symfony docs:

image

We run on several instances the maintenance task every 5 Minutes in parallel overlapping mode......means if the maintenance task is still not finished (for whatever reason...image optimisation etc) 5 minutes later cron is starting the next maintenance task process. To avoid conflicts it should skip the task already running in another process

### DIFF
--- a/lib/Maintenance/Executor.php
+++ b/lib/Maintenance/Executor.php
@@ -81,15 +81,13 @@ final class Executor implements ExecutorInterface
 
             $lock = $this->lockFactory->createLock('maintenance-' . $name, 86400);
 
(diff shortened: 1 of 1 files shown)


ssiergl:
@dvesh3 another one ;)

dvesh3:
thank you!

PR #7389, opened 2020-10-25
Order Details Tab - fix inconsistent permission check

das-peter:
The Order Details Tab for the ecommerce backoffice detail view has stricter access check than the actual detail page.
It doesn't seem to be related to `bundle_ecommerce_pricing_rules` but there was an access check for that in place - re-organized.


fashxp:
thx for you PR ... since it is a bugfix, please rebase to 6.8.

PR #7393, opened 2020-10-26
generate public url through asset listeners

davidhoeck:
## Description of the bug 

The public url displayed in the admin backend of pimcore always displays an url, with the hostname of the Pimcore application as a base. 

If you use AWS S3 or AWS Cloudfront for example, the public url should be the specific url of the used system.

This PR generates the public url of an asset via the asset events. 
If the url does not include http (no public url) it will prepend the hostname of the Pimcore application.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -25,6 +25,7 @@ use Pimcore\Db;
 use Pimcore\Event\Admin\ElementAdminStyleEvent;
 use Pimcore\Event\AdminEvents;
 use Pimcore\Event\AssetEvents;
+use Pimcore\Event\FrontendEvents;
 use Pimcore\File;
 use Pimcore\Loader\ImplementationLoader\Exception\UnsupportedException;
 use Pimcore\Logger;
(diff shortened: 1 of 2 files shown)


brusch:
@davidhoeck Thanks a lot for your PR! 
I'm just thinking if it would make sense to just add an optional argument to `Asset::getFullPath()` to force returning the frontend path, this would avoid duplicating code. WDYT?

davidhoeck:
@brusch good idea!  I also thought of that, should I update the PR ?

brusch:
@davidhoeck yes, but for BC reasons it's probably better to introduce a new method `getFrontendFullPath()` (tagged as `@internal`)  which is then used in `getFullPath()` and in the controller.

davidhoeck:
@brusch done ✅

bundles/AdminBundle/Controller/Admin/AssetController.php
```diff
@@ -25,6 +25,7 @@
 use Pimcore\Event\Admin\ElementAdminStyleEvent;
 use Pimcore\Event\AdminEvents;
 use Pimcore\Event\AssetEvents;
+use Pimcore\Event\FrontendEvents;
```
brusch, line 28:
```suggestion
```

brusch:
Nice, thanks a lot! 👍

PR #7397, opened 2020-10-27
[Docs] Renderlet can accept 'auto' as height as well

michaelclaesson:
Clarify that you can use 'auto' as height to have the renderlet dynamically change height depending on content, just like a snippet.


brusch:
Thank you!

PR #7403, opened 2020-10-27
Order Details Tab - fix inconsistent permission check

das-peter:
The Order Details Tab for the ecommerce backoffice detail view has stricter access check than the actual detail page.
It doesn't seem to be related to `bundle_ecommerce_pricing_rules` but there was an access check for that in place - re-organized.
Maybe the proper base for https://github.com/pimcore/pimcore/pull/7389


dvesh3:
thank you!

PR #7404, opened 2020-10-27
Make "GoogleTagManager::formatPrice()" compatible with its default value

das-peter:
`GoogleTagManager::formatPrice()` accepts and defaults to NULL for `$price`. But `Decimal::fromNumeric($price)` wont work with that and throw a `\InvalidArgumentException`.
Add a check to be compatible with the method signature - similar to the code handling in `EnhancedEcommerce::transformProductImpression()`


dvesh3:
@das-peter Can you please rebase the PR on 6.8? since it is a fix! thanks!

brusch:
Thanks a lot!

PR #7405, opened 2020-10-28
[Maintenance] fix not updating TmpStore on executing maintenance task

### LINKED ISSUE #7347: Maintenance warning after upgrading from 6.7.x to 6.8.x
Hi,

Since the upgrade to 6.8.x I see the maintenance warning in PimCore the maintenance is not setup properly:

image

The maintenance is setup properly and is running every 5 minutes. Also when running maintenance manually with -vvv there are no errors.

Before 6.7.x no issues.

ssiergl:
TmpStore does not get updated till default expiration time of 7 days (86400 * 7).
Just use TmpStore::set instead of TmpStore:add to always update the maintenance entry as soon
it is executed.

fixes #7347

### DIFF
--- a/lib/Maintenance/Executor.php
+++ b/lib/Maintenance/Executor.php
@@ -119,7 +119,7 @@ final class Executor implements ExecutorInterface
      */
     public function setLastExecution()
     {
-        TmpStore::add($this->pidFileName, time());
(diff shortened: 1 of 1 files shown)


brusch:
Thanks a lot 👌🙏

PR #7407, opened 2020-10-28
[Grid config] Error when opening saved configuration which contains fields of deleted object brick

BlackbitNeueMedien:
Steps to reproduce:
1. Create class `Test` with `object brick` field `bricks`
2. Create object brick `testbrick` with an input field `testfield`
3. Create object `A` in `Home` folder (tree root)
4. Open `Home` in grid view
5. Edit table configuration and add field `testfield` of brick `testbrick`
6. Save configuration under `My configuration`
7. Delete object brick `testbrick`
8. Reload grid for `Home` folder
9. Try to open previously saved table configuration `My configuration` -> results in error `Call to a member function getFieldDefinition() on null`

This PR fixes that. If the brick of a field in the saved table configuration does not exist anymore, this field will be skipped when generating the table.


brusch:
Thank you very much! 👍

PR #7410, opened 2020-10-28
Update `doctrine/dbal` to `v2.10` and fix deprecations

jdreesen:
I started this some time ago and create this draft PR now, so nobody does duplicate work here.

The update to `v2.10` with its deprecation fixes should be complete. 

But in the meantime `v2.11` and `v2.12` were released with new deprecations which I'm going to address in the next days.


brusch:
Closed in favor of #7453

PR #7414, opened 2020-10-29
[Maintenance] fix remove duplicate lock for CacheCleanupTask and LowQualityImagePreviewTask

ssiergl:
Executor is handling all task locks. Remove own lock handling from CacheCleanupTask and LowQualityImagePreviewTask
to prevent duplicate lock entries in lock_keys table.

@dvesh3 @brusch


ssiergl:
forget it...to late to do such stuff...everything is valid...closing this

PR #7418, opened 2020-10-29
Fixed locking - follow up to #7386

brusch:
follow up to #7386

### DIFF
--- a/lib/Maintenance/Tasks/CacheCleanupTask.php
+++ b/lib/Maintenance/Tasks/CacheCleanupTask.php
@@ -54,9 +54,8 @@ final class CacheCleanupTask implements TaskInterface
      */
     public function execute()
     {
-        if (!$this->lock->isAcquired() && date('H') <= 4) {
+        if ($this->lock->acquire() && date('H') <= 4) {
             // execution should be only sometime between 0:00 and 4:59 -> less load expected
-            $this->lock->acquire();
             $this->logger->debug('Execute purge() on cache handler');
             $this->cacheHandler->purge();
         } else {
--- a/lib/Maintenance/Tasks/CacheCleanupTask.php
+++ b/lib/Maintenance/Tasks/CacheCleanupTask.php
@@ -54,7 +54,7 @@ final class CacheCleanupTask implements TaskInterface
      */
     public function execute()
     {
-        if ($this->lock->acquire() && date('H') <= 4) {
+        if (date('H') <= 4 && $this->lock->acquire()) {
             // execution should be only sometime between 0:00 and 4:59 -> less load expected
             $this->logger->debug('Execute purge() on cache handler');
             $this->cacheHandler->purge();
--- a/lib/Maintenance/Tasks/LowQualityImagePreviewTask.php
+++ b/lib/Maintenance/Tasks/LowQualityImagePreviewTask.php
@@ -46,7 +46,7 @@ final class LowQualityImagePreviewTask implements TaskInterface
      */
     public function execute()
     {
-        if ($this->lock->acquire() && date('H') <= 4) {
+        if (date('H') <= 4 && $this->lock->acquire()) {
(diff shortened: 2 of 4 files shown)


lib/Maintenance/Tasks/CacheCleanupTask.php
```diff
@@ -54,9 +54,8 @@ public function __construct(CoreHandlerInterface $cacheHandler, LoggerInterface
      */
     public function execute()
     {
-        if (!$this->lock->isAcquired() && date('H') <= 4) {
+        if ($this->lock->acquire() && date('H') <= 4) {
```
blankse, line 57:
```suggestion
        if (date('H') <= 4 && $this->lock->acquire()) {
```
Should not lock out of time 0:00 and 4:59, or?

lib/Maintenance/Tasks/LowQualityImagePreviewTask.php
```diff
@@ -46,9 +46,8 @@ public function __construct(LoggerInterface $logger, LockFactory $lockFactory)
      */
     public function execute()
     {
-        if (!$this->lock->isAcquired() && date('H') <= 4) {
+        if ($this->lock->acquire() && date('H') <= 4) {
```
blankse, line 49:
```suggestion
        if (date('H') <= 4 && $this->lock->acquire()) {
```

lib/Maintenance/Tasks/CacheCleanupTask.php
brusch: Yep, makes sense! Actually it shouldn't make any difference in the functionality, since the lock lasts 24 hours, but it's definitely nicer to not call `acquire()` if not necessary 😊

brusch: Fixed it on the master as well: a46dccf1045abe1126767333f93e9171242b4d1a

models/Asset/Document/ImageThumbnail.php
```diff
@@ -95,12 +95,12 @@ public function generate($deferredAllowed = true)
...
 
-                    if (!is_file($path) && !$lock->isAcquired()) {
-                        $lock->acquire();
+                    $lockAcquired = $lock->acquire();
+                    if (!is_file($path) && $lockAcquired) {
                         $converter->saveImage($path, $this->page);
                         $generated = true;
                         $lock->release();
-                    } elseif ($lock->isAcquired()) {
+                    } elseif (!$lockAcquired) {
```
blankse, line 103: I would here use `$lock->isAcquired()` and let the `$lock->acquire()` behind `!is_file($path)`

Should not be locked if the path exists. Only if the file is missing

brusch: @blankse But `isAcquired()` only works for the same process, we actually want to process only one document at the time here, so we have to use `acquire()`

blankse: Is the same process in the if above

brusch: Sorry, I don't get your point😐
It's an `if` /  `elseif` so if we got the lock in this process ( by `$lockAcquired = $lock->acquire();`) we never reach the `elseif` anyway, so this is just for the case that we're in another process where we didn't got the lock because the conversion is already in progress by another process.

blankse: Okay, I get your point. The lock was also selected before if the file exists. I think this is a unnecessary SQL. Why not make 2 ifs?
```
if (!is_file($path)) {
    if ($lock->acquire()) {
        ...
    } else {
       ...
    }
}
```

blankse: Or can exist the file but is not ready... Then we should leave it as it is.

brusch: Yep - done - thanks! 😊

PR #7420, opened 2020-10-30
[Pimcore7] implement minimize scripts

dpfaffenbauer:


lib/Twig/Extension/AdminExtension.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        return [
+            new TwigFunction('pimcore_minimize_scripts', [$this, 'minimze']),
+        ];
+    }
+
+    public function minimze($value)
+    {
+        return Admin::getMinimizedScriptPath($value, false);
+    }
```
aarongerig, line 40: Missing return types and doc blocks. :)

brusch:
The entire concatenation and minification will be moved out of the templates anyway, since it's definitely not the right place to do that. We just wanted to keep things simple so we migrated this part as well, but this will be refactored again. 

Thanks anyway 😉

dpfaffenbauer:
yes, I thought so too, but as of now, you cannot run the Pimcore 7-dev without DEVMODE, which makes things quite slow. What if we merge this for now and remove later?

PR #7428, opened 2020-11-01
Fix null data on table with columns config enabled

### LINKED ISSUE #7365: Adding Table with columns config activated on existing object class creates an exception on DAO load
# Bug Report
### Expected behavior
Adding a table with colum config on an existing object should work
### Actual behavior
It works in pimcore ui but make an exception in PHP API
### Steps to reproduce
On an existaing object, create a table with config colums activated, then try a foreach loop on a listing of your objects. you can't access your objects anymore

### Code example
```php
$skus = new Sku\Listing();
foreach ($skus as $sku) {
  echo 'ok'
} // each loop will generate an exception without executing the core loop code
```

The error is located in this function :
[Table.php:324](https://github.com/pimcore/pimcore/blob/master/models/DataObject/ClassDefinition/Data/Table.php#L324)

I can do a PR but i don't know what is the best solutuion to solve the problem : 
This bug appears because $data (witch is null) is casted to string ('') before unserializing. so the next condition on null is never verified. Then the foreach on line 341 is trying to iterate on an empty string.

- Solution 1: remove the string cast
- Solution 2: replace the '=== null' condition by an 'empty()' condition
- Solution 3: protect the foreach loop with an 'is_array' condition

ChristopheBraud:
Resolves #7365 

## Additional info  
Fix an issue with table access when columns config is enabled

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Table.php
+++ b/models/DataObject/ClassDefinition/Data/Table.php
@@ -325,8 +325,8 @@ class Table extends Data implements ResourcePersistenceAwareInterface, QueryReso
     {
         $unserializedData = Serialize::unserialize((string) $data);
 
-        if ($unserializedData === null) {
-            return $unserializedData;
+        if ($data === null || $unserializedData === null) {
+            return null;
         }
 
         //set array keys based on column configuration if set


fashxp:
thx for your PR ... can you please sign the CLA so that we can merge.

ChristopheBraud:
It seems that there is a bug with the CLA, I've signed it before the PR (because I've read CONTRIBUTING.md)
image

ChristopheBraud:
oops sorry, I've used my other github account :/

PR #7429, opened 2020-11-02
[Dashboard] Set flex=1 for columns without specified width for custom report portlet

BlackbitNeueMedien:
Like in https://github.com/pimcore/pimcore/blob/12900c5c22d5cde7284295ba155f9a23dce8ba99/bundles/AdminBundle/Resources/public/js/pimcore/report/custom/report.js#L71-L75 this PR sets flex width for columns, which do not have a width being explicitly set, to use the full availyble width of the portlet.


brusch:
Thanks a lot!

PR #7431, opened 2020-11-02
sylius/theme-bundle allows translations per theme, by that it adds the theme-name to the locale with @

dpfaffenbauer:
Resulting in Pimcore in an error that the cache is using invalid characters. I think we should allow any character here, since symfony does too


brusch:
👍 Thanks!

PR #7435, opened 2020-11-02
[QuantityValueUnit] make setters chainable

dpfahlbusch:
Make it possible to chain setters

Not sure if its a "bugfix" or "feature"

Resolves #

### DIFF
--- a/models/DataObject/QuantityValue/Unit.php
+++ b/models/DataObject/QuantityValue/Unit.php
@@ -67,7 +67,9 @@ class Unit extends Model\AbstractModel
      */
     public $conversionOffset;
 
-    /** @var string */
+    /**
+     * @var string
+     */
     public $converter;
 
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Nice, can we add according phpdoc in the same round?

models/DataObject/QuantityValue/Unit.php
```diff
@@ -182,26 +184,44 @@ public function __toString()
         return ucfirst($this->getAbbreviation() . ' (' . $this->getId() . ')');
     }
 
+    /**
+     * @param string $abbreviation
+     *
+     * @return Unit
```
jdreesen, line 190: Should we use this instead for chainable methods?
```suggestion
     * @return $this
```

weisswurstkanone:
thx!

PR #7442, opened 2020-11-03
[Document] Improve usability of block editable for Twig templates

brusch:
## Old style
```twig
    {% for i in pimcore_iterate_block(pimcore_block('contentblock')) %}
        <h2>{{ pimcore_input('subline') }}</h2>
    {% endfor %}
```

## New style
```twig
    {% for i in pimcore_block('contentblock').iterator %}
        <h2>{{ pimcore_input('subline') }}</h2>
    {% endfor %}
```

### DIFF
--- a/doc/Development_Documentation/02_MVC/02_Template/00_Twig.md
+++ b/doc/Development_Documentation/02_MVC/02_Template/00_Twig.md
@@ -143,19 +143,6 @@ detailed description of every helper:
 * `pimcore_cache`
 * `pimcore_url`
 
-
-#### Block elements
-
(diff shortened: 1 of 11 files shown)


NiklasBr:
Just a thought here... Maybe use `getIterator()` instead of `getIterable()` to follow the [PHP style](https://www.php.net/manual/en/arrayobject.getiterator.php)?

brusch:
Yeah, I was also thinking about that - thanks for your input!

brusch:
BC break caused by: 
https://github.com/pimcore/pimcore/blob/2119ce36697201475395728d4acb00bb12c091a5/models/Document/Editable/BlockInterface.php#L22

PR #7453, opened 2020-11-03
Pimcore requires doctrine/dbal ^2.11.0 and php ^7.3

blankse:
The PR https://github.com/pimcore/pimcore/pull/7327 use \Doctrine\DBAL\Exception which is first created in doctrine/dbal 2.11.0.
doctrine/dbal 2.11.0 requires php ^7.3

### DIFF
--- a/lib/Db/PimcoreExtensionsTrait.php
+++ b/lib/Db/PimcoreExtensionsTrait.php
@@ -14,7 +14,12 @@
 
 namespace Pimcore\Db;
 
+use Doctrine\DBAL\Cache\CacheException;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
+use Doctrine\DBAL\Driver\Exception as DriverException;
+use Doctrine\DBAL\Driver\PDO\Statement as PDOStatement;
+use Doctrine\DBAL\Driver\ResultStatement;
+use Doctrine\DBAL\Exception as DBALException;
 use Pimcore\Db;
 use Pimcore\Db\ZendCompatibility\Expression;
 use Pimcore\Db\ZendCompatibility\QueryBuilder;
@@ -51,11 +56,11 @@ trait PimcoreExtensionsTrait
     }
 
     /**
-     * @see \Doctrine\DBAL\Connection::query
+     * @see \Doctrine\DBAL\Connection::executeQuery
      *
-     * @return \Doctrine\DBAL\Driver\Statement
+     * @return ResultStatement
      *
-     * @throws \Doctrine\DBAL\DBALException
+     * @throws DBALException
      */
     public function query(...$params)
     {
@@ -63,7 +68,7 @@ trait PimcoreExtensionsTrait
         // eg. $db->query("UPDATE myTest SET date = ? WHERE uri = ?", [time(), $uri]);
         if (func_num_args() === 2) {
             if (is_array($params[1])) {
-                return parent::executeUpdate($params[0], $params[1]);
+                return parent::executeQuery($params[0], $params[1]);
             }
         }
 
@@ -71,20 +76,20 @@ trait PimcoreExtensionsTrait
             $params[0] = $this->normalizeQuery($params[0], [], true);
         }
 
-        return parent::query(...$params);
+        return parent::executeQuery(...$params);
     }
 
     /**
      * @see \Doctrine\DBAL\Connection::executeQuery
      *
-     * @param string                                      $query  The SQL query to execute.
-     * @param array                                       $params The parameters to bind to the query, if any.
-     * @param array                                       $types  The types the previous parameters are in.
-     * @param \Doctrine\DBAL\Cache\QueryCacheProfile|null $qcp    The query cache profile, optional.
+     * @param string $query The SQL query to execute.
+     * @param array $params The parameters to bind to the query, if any.
+     * @param array $types The types the previous parameters are in.
+     * @param QueryCacheProfile|null $qcp The query cache profile, optional.
      *
-     * @return \Doctrine\DBAL\Driver\Statement The executed statement.
+     * @return ResultStatement The executed statement.
      *
-     * @throws \Doctrine\DBAL\DBALException
+     * @throws DBALException
      */
     public function executeQuery($query, array $params = [], $types = [], QueryCacheProfile $qcp = null)
     {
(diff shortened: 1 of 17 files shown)


brusch:
Just wanted to comment that Travis needs to be changed 😉 Thanks for updating the PR.

brusch:
Related to #7410
@blankse can you maybe also fix the deprecations in this PR?

jdreesen:
`doctrine/dbal v2.11` has many more deprecations that should also be fixed: https://github.com/doctrine/dbal/blob/2.12.x/UPGRADE.md#upgrade-to-211

blankse:
@brusch @jdreesen I fixed many 2.11 deprecations.

There are deprecations in QueryBuilder and a deprecation with the getParams() (is internal now).
QueryBuilder is also deprecated, so I ignore it for now.
For getParams exists no alternative. We should load it direct from the settings?

But we can leave it at that for now, or?

brusch:
@blankse nice, thanks a lot! 
`QueryBuilder`  👍
`getParams()` yep, we have to get the info then from the settings - are you updating the PR? 😊

brusch:
@blankse regarding `>=` PHP 7.3 we should update this in the docs as well: 
https://github.com/pimcore/pimcore/blob/ae7b62cc4a0ade1bac5ad70d720187215344f827/doc/Development_Documentation/23_Installation_and_Upgrade/01_System_Requirements.md#L12

blankse:
@brusch I looked in the container. There are no symfony parameter for the db settings. They are only set to the Connection cunstructur.

brusch:
@blankse hm, too bad, then we need to put it on the container as a parameter I guess, probably here is a good place for that: 
https://github.com/pimcore/pimcore/blob/9af9958fbcfb9e86406f8076212196a52b831b81/bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php#L70
wdyt?

blankse:
@brusch From where do you want get the settings in the PimcoreCoreExtension? The config variable have only the settings from the pimcore configuration tree.

The db settings are in the doctrine configuration tree.

brusch:
@blankse 🤦‍♂️ you're right, sorry 😊

brusch:
I think using the `@internal` tagged `getParams()` is fine for now, if Doctrine changes something there in the future we can always add our own getter to `\Pimcore\Db\PimcoreExtensionsTrait`.

blankse:
@brusch I agree with you

lib/Db/PimcoreExtensionsTrait.php
```diff
@@ -14,7 +14,12 @@
 
 namespace Pimcore\Db;
 
+use Doctrine\DBAL\Cache\CacheException;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
+use Doctrine\DBAL\Driver\Exception as DriverException;
+use Doctrine\DBAL\Driver\PDO\Statement as PDOStatement;
```
brusch, line 20:
```suggestion
use Doctrine\DBAL\Driver\Result;
```

lib/Db/PimcoreExtensionsTrait.php
```diff
@@ -278,12 +288,13 @@ public function fetchCol($sql, $params = [], $types = [])
         // unfortunately Mysqli driver doesn't support \PDO::FETCH_COLUMN, so we have to do it manually
         $stmt = $this->executeQuery($sql, $params, $types);
         $data = [];
-        while (($row = $stmt->fetchColumn()) || $row !== false) {
-            $data[] = $row;
+        if ($stmt instanceof PDOStatement) {
```
brusch, line 291:
```suggestion
        if ($stmt instanceof Result) {
```

lib/Db/PimcoreExtensionsTrait.php
```diff
@@ -315,15 +328,18 @@ public function fetchOne($sql, $params = [], $types = [])
...
      */
     public function fetchPairs($sql, array $params = [], $types = [])
     {
         $params = $this->prepareParams($params);
-        $statement = $this->executeQuery($sql, $params, $types);
+        $stmt = $this->executeQuery($sql, $params, $types);
         $data = [];
-        while ($row = $statement->fetch(\PDO::FETCH_NUM)) {
-            $data[$row[0]] = $row[1];
+        if ($stmt instanceof PDOStatement) {
```
brusch, line 339:
```suggestion
        if ($stmt instanceof Result) {
```

lib/Targeting/Storage/DbStorage.php
```diff
@@ -18,8 +18,9 @@
 namespace Pimcore\Targeting\Storage;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\PDO\Statement as PDOStatement;
```
brusch, line 21:
```suggestion
use Doctrine\DBAL\Driver\Result;
```

lib/Targeting/Storage/DbStorage.php
```diff
@@ -86,11 +87,12 @@ public function all(VisitorInfo $visitorInfo, string $scope): array
         $this->addExpiryParam($qb, $scope);
 
         $stmt = $qb->execute();
-        $result = $stmt->fetchAll();
-
         $data = [];
-        foreach ($result as $row) {
-            $data[$row['name']] = json_decode($row['value'], true);
+
+        if ($stmt instanceof PDOStatement) {
```
brusch, line 92:
```suggestion
        if ($stmt instanceof Result) {
```

lib/Targeting/Storage/DbStorage.php
```diff
@@ -119,7 +121,11 @@ public function has(VisitorInfo $visitorInfo, string $scope, string $name): bool
         $this->addExpiryParam($qb, $scope);
 
         $stmt = $qb->execute();
-        $result = (int)$stmt->fetchColumn();
+        $result = 0;
+
+        if ($stmt instanceof PDOStatement) {
```
brusch, line 126:
```suggestion
        if ($stmt instanceof Result) {
```

lib/Targeting/Storage/DbStorage.php
```diff
@@ -177,7 +183,11 @@ public function get(VisitorInfo $visitorInfo, string $scope, string $name, $defa
         $this->addExpiryParam($qb, $scope);
 
         $stmt = $qb->execute();
-        $result = $stmt->fetchColumn();
+        $result = false;
+
+        if ($stmt instanceof PDOStatement) {
```
brusch, line 188:
```suggestion
        if ($stmt instanceof Result) {
```

lib/Targeting/Storage/DbStorage.php
```diff
@@ -292,9 +302,12 @@ private function loadDate(VisitorInfo $visitorInfo, string $scope, string $selec
         $this->addExpiryParam($qb, $scope);
 
         $stmt = $qb->execute();
-        $date = $this->convertToDateTime($stmt->fetchColumn());
 
-        return $date;
+        if ($stmt instanceof PDOStatement) {
```
brusch, line 306:
```suggestion
        if ($stmt instanceof Result) {
```

brusch (changes requested):
I think it's more generic if we use `\Doctrine\DBAL\Driver\Result` instead of `\Doctrine\DBAL\Driver\PDO\Statement` then `Mysqli` is still an option 😊

PR #7427, opened 2020-10-31
[Pimcore7] Remove php engine completely

dpfaffenbauer:
- Remove any reference to templating
 - Switch to new Template Filename structure (@BundleName/Path/to.html.twig)
 - disable templating (twig enabled by default)
 - enable assets (was enabled by templating before)
 - remove all CompilerPasses related to PHP Engine
 - Switch references from Engine to Twig

PR #7454, opened 2020-11-03
[Core] Template Engine Delegation

dvesh3:
https://github.com/pimcore/pimcore/pull/7427#issuecomment-720499944


dpfaffenbauer:
does that bring back all the deprecation messages again?

dpfaffenbauer:
also, Symfony\Component\Templating\DelegatingEngine has been removed with 5.0.

dvesh3:
> does that bring back all the deprecation messages again?

@dpfaffenbauer Nope! It is using Component classes.

dpfaffenbauer:
nevermind, that should actually work.

lib/Templating/DelegatingEngine.php
```diff
@@ -0,0 +1,84 @@
...
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Templating;
```
brusch, line 15: Shouldn't we just use Symfony's `DelegatingEngine` and configure it properly, e.g.: 

```yml
    pimcore.templating.engine.delegating:
        public: true
        class: Symfony\Component\Templating\DelegatingEngine
        arguments: ['@twig']
```

?

brusch (changes requested):

brusch:
> also, Symfony\Component\Templating\DelegatingEngine has been removed with 5.0.

Still there: 
https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/Templating/DelegatingEngine.php

Surprisingly also `PhpEngine` 🤔
https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/Templating/PhpEngine.php

dpfaffenbauer:
@brusch Yep, I corrected myself later. It's fine like that. I've looked at the FrameworkBundle and not the Component. The Component is not deprecated, just within the FrameworkBundle. 

Re PhpEngine.php: I guess they'll keep the Templating Component for BC reasons and just removed the PHP Templating Engine from the FrameworkBundle and stopped maintenance for the Templating Component.

lib/Templating/DelegatingEngine.php
dvesh3: Yes, that was my first thought. Unfortunately, it does not support `Twig\Environment` directly so we need to introduce a wrapper. Also, overriding DelegatingEngine here means we can switch the case for Twig or custom engines(TwigEngine: Twig Wrapper, PhpEngine), so it is still faster for instances using just Twig.

PR #7456, opened 2020-11-03
[Image thumbnails] Do not filter requested attributes for image thumbnail HTML

### LINKED ISSUE #7438: [Image Thumbnails] Support loading="lazy"
Pimcore already has support for lazy-loading image thumbnails with `lowQualityPlaceholder` parameter. While this does work, it requires a JS implementation for actually loading the images. Meanwhile most browsers support this feature natively with [`loading="lazy"`](https://web.dev/browser-level-image-lazy-loading/) in the `img` tag. Would be cool if we could add a parameter for that too, e.g. `$image->getThumbnail()->getHtml(['lazy' => true])`

BlackbitNeueMedien:
Resolves #7438

Currently it is possible to set arbitrary attributes to generated HTML code for image thumbnails. For the W3C-specified ones you can provide them directly, e.g. `$image->getThumbnail()->getHtml(['class' => 'my-image'])`. But for those not specified in https://github.com/pimcore/pimcore/blob/530098ea48745926d8f155cce5e6228c222968ac/models/Asset/Image/Thumbnail.php#L205-L208 you have to use a special `attributes` array key, e.g. `$image->getThumbnail()->getHtml(['attributes' => ['loading' => 'lazy']])`
Although this `loading` attribute might not be in an official W3C standard, it is implemented by most browser vendors, so imho it does not make much sense to distinguish between W3C attributes and custom ones.

Additionally in the [documentation](https://pimcore.com/docs/pimcore/current/Development_Documentation/Assets/Working_with_Thumbnails/Image_Thumbnails.html#page_Advanced-Examples) this special `attributes` options key is not mentioned.

### DIFF
--- a/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
+++ b/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
@@ -129,6 +129,9 @@ $height = $thumbnail->getHeight();
  
 // get the html "img" tag for the thumbnail incl. custom class:
 echo $thumbnail->getHtml(["class" => "custom-class"]);
+
+// get the html "img" tag including attribute to enable lazy.loading
+echo $thumbnail->getHtml(["loading" => "lazy"]);
  
 // get the path to the thumbnail
 $path = $thumbnail->getPath();
@@ -160,10 +163,8 @@ echo $thumbnail; // prints something like /var/tmp/....png
 <?= $asset->getThumbnail("content")->getHtml(['alt' => 'top priority alt']) ?>
 // or
 <?= $asset->getThumbnail("content")->getHtml(['defaultalt' => 'default alt, if not set in image']) ?>
-
-
 ```
-
+Additionally there are some special parameters to [customize generated image HTML code](https://pimcore.com/docs/pimcore/current/Development_Documentation/Documents/Editables/Image.html#page_Configuration).
 
 ## Using ICC Color Profiles for CMYK -> RGB 
 Pimcore supports ICC color profiles to get better results when converting CMYK images (without embedded color profile) 
--- a/models/Asset/Image/Thumbnail.php
+++ b/models/Asset/Image/Thumbnail.php
@@ -202,11 +202,6 @@ class Thumbnail
             }
         }
 
-        $w3cImgAttributes = ['alt', 'align', 'border', 'height', 'hspace', 'ismap', 'longdesc', 'usemap',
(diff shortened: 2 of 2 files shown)


christopher-siegel:
👍  Lots of frontend libraries use attributes that aren't "valid". I think it's a good change.

brusch:
Actually there's a reason why there's the `attributes` config option, because not all config options should be used as HTML attributes. 
Just consider the following example: 
```php
$asset->getThumbnail('content')->getHtml([
    'loading' => 'lazy',
    'defaultalt' => 'foo',
    'disableAutoAlt' => true,
    'lowQualityPlaceholder' => true,
]);
```

With your fix this would result in: 

```
<img width="1140" height="760" loading="lazy" defaultalt="foo" disableAutoAlt="1" lowQualityPlaceholder="1" alt="" src="data:image/svg+xml;base64,P...." data-src="/Sample%20Content/Example%20Images/image-thumb__362__content/Fiat%20500%20-%20city%20drive.webp" data-srcset="/Sample%20Content/Example%20Images/image-thumb__362__content/Fiat%20500%20-%20city%20drive.webp 1x, /Sample%20Content/Example%20Images/image-thumb__362__content/Fiat%20500%20-%20city%20drive@2x.webp 2x" />
```
So all scalar config options are getting added as HTML attributes. 

I see 2 possible solutions: 
1. to be consistent we only allow HTML attributes which are passed to the `attributes` config option
2. we just add `loading` to the allowed w3c attributes

christopher-siegel:
I see, that makes sense. 

Filtering out the pimcore specific ones would be another option.
I didn't think about the `attributes` config option. That should be enough for most use cases.

But I'd add `loading` to the w3c attributes as well.

brusch:
I'd also go for option 2 for now. @BlackbitNeueMedien can you please update your PR accordingly? 
Thanks 😊

BlackbitNeueMedien:
Option 2 breaks all current implementations like `$asset->getThumbnail('content')->getHtml(['class' => 'my-image']);`
Don't you think that it is better to filter out the "special" options like `lowQualityPlaceholder`, `defaultalt` etc.?

brusch:
@BlackbitNeueMedien Can you update the PR, or should I do it? Thanks 😊

BlackbitNeueMedien:
@brusch If you have the time, this would be great. I do not have a customer paying for this. I just stumbled upon an article about browser-native lazy loading and thought that was something cool for Pimcore. If you don't I will try to change it this evening...

brusch:
Closed in favor of #7461

PR #7459, opened 2020-11-04
Correct file path to fosjsrouting router.js

nicklassandell:
Identical to #7370 but with the author email corrected so that the CLA can be approved.
  
  
## Issue
Pull #6146 introduced an issue that causes editmode to break in the admin if you are using asset versioning. A file path is generated using `$this->package->getUrl()` which also includes asset versioning (if used) in the path, and since the file is fetched with `file_get_contents()` at one point, this fails because `file_get_contents()` does not understand the query param appended to the path.

## Steps to reproduce
Add this to your `app/config/config.yml` file and load a document in editmode.

```
  framework:
      assets:
          version: 'xxx'
```

You should get an error from file_get_contents() failing to read from a path like this:

`/var/www/html/web/bundles/fosjsrouting/js/router.js?23471` <-- see query param

## Solution
Using a relative path (like the other files included) seems to work just fine. I also commented on the [original PR](https://github.com/pimcore/pimcore/pull/6146#issuecomment-712938852) to check if the contributor (@dpfaffenbauer) knew the reason for the `getUrl()` call, but he's not sure anymore.

### DIFF
--- a/bundles/CoreBundle/EventListener/Frontend/EditmodeListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/EditmodeListener.php
@@ -318,7 +318,7 @@ class EditmodeListener implements EventSubscriberInterface
     {
         return array_merge(
             [
-                $this->package->getUrl('bundles/fosjsrouting/js/router.js'),
+                '/bundles/fosjsrouting/js/router.js',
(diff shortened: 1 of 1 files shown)


brusch:
Thank you very much!

nicklassandell:
Happy to help!

PR #7460, opened 2020-11-04
Fixes/improvements to sort in MiscController

NiklasBr:

Sorts Actions, Bundles, Controllers, Templates better. The previous `sort()` method was triggered on a nested array and did not actually work. Also adds flag to sort naturally case insensitive.

Previously:
```
Areas/templage.twig
Documents/template.twig
bundles/template.twig
```

Now:
```
Areas/templage.twig
bundles/template.twig
Documents/template.twig
```


brusch:
Thanks a lot!

PR #7461, opened 2020-11-04
[Asset] Image thumbnails: allow loading=lazy attribute

brusch:


BlackbitNeueMedien:
First thanks, @brusch - could you please add a hint to setting non-standard attributes to the documentation - e.g. `$image->getThumbnail()->getHtml(['attributes' => ['non-standard' => 'value']])`?

christopher-siegel:
@BlackbitNeueMedien There is a section in the Image Editable Docs explaining this already: 
https://pimcore.com/docs/pimcore/current/Development_Documentation/Documents/Editables/Image.html#page_An-Example-Using-Custom-Attributes

brusch:
I've added an example here #7436 - as the examples are written in PHP at the moment, it doesn't make sense to still add this in this PR 😉

PR #7464, opened 2020-11-05
Quoting null value when for preparation_worker_id leads to '' being the value

das-peter:
It seems that quoting the null value to insert into preparation_worker_id `$data['preparation_worker_id'] = $this->db->quote(null);` leads to the value stored being `''`. This means preparation_worker_id is not empty which will prevent the processor from picking up the entry!


das-peter:
Tottaly missed that the lease is also checked and hence the ugly value isn't a blocker and just is ugly.

brusch:
Thanks a lot!

Issue #6187, opened 2020-04-08
Zend Paginator is abondoned, replace with Laminas?


NiklasBr:
[zendframework/zend-paginator](/zendframework/zend-paginator) is abandoned and suggests  [laminas/laminas-paginator](/laminas/laminas-paginator) instead. Listing classes will need to implement `Laminas\Paginator\Adapter\AdapterInterface` which is likely a BC break.

Something for Pimcore 7.0 perhaps?

vmalyk:
@brusch , I suggest to change Zend packages to Laminas in next major release and use laminas/laminas-zendframework-bridge for using new packages without code update now and prepare to full code migration in 7.0 (or maybe a bit later - 7.1, 7.2).
What do you think?

Thanks!

tecbird:
+1, please update and use laminas instead of deprecated zend modules

PR #7465, opened 2020-11-05
[Travis] Add tests with lowest dependencies & changed zend-* packages to laminas

blankse:
I think it would be good to test with lowest dependencies. Than we can see when a class is used, which is added in a newer version as defined in composer.json

Also removed old symfony 3 tests, load only symfony as source and use composer 2 to reduce the test duration.


brusch:
Fixes #6187

brusch:
Very nice, thanks a lot! (Composer 2 is just great, tests are twice as fast now 🎉)

blankse:
@brusch Yes and only symfony as source should also be faster ;)

brusch:
But prefer source was there before as well, or was it ignored for a reason?

blankse:
@brusch I removed it and add only source for symfony (see config section in composer.json [root only]). source is slower than dist.

PR #7468, opened 2020-11-05
Fix quantity value comparator

t-zilla:

* Fix comparison of Quantity Values with no units

## Additional info  

Empty `unitId` can have multiple representations: `null`, `0` or `"0"` which are semantically the same

### DIFF
--- a/models/DataObject/ClassDefinition/Data/QuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/QuantityValue.php
@@ -748,7 +748,7 @@ class QuantityValue extends Data implements ResourcePersistenceAwareInterface, Q
             return false;
         }
 
-        return $oldValue->getValue() === $newValue->getValue() &&
-            intval($oldValue->getUnitId()) === intval($newValue->getUnitId());
+        return $oldValue->getValue() === $newValue->getValue()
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -748,6 +748,7 @@ public function isEqual($oldValue, $newValue): bool
             return false;
         }
 
-        return $oldValue->getValue() === $newValue->getValue() && $oldValue->getUnitId() === $newValue->getUnitId();
+        return $oldValue->getValue() === $newValue->getValue() &&
+            intval($oldValue->getUnitId()) === intval($newValue->getUnitId());
```
jdreesen, line 752:
```suggestion
        return $oldValue->getValue() === $newValue->getValue()
            && (int)$oldValue->getUnitId() === (int)$newValue->getUnitId();
```

brusch:
We can't cast to `int` as the ID can be a string as well.

t-zilla:
@brusch What about `empty()`?

brusch:
@t-zilla yep, should work, created a follow up here: #7469

PR #7470, opened 2020-11-05
Travis: use Composer v2

brusch:


.travis/setup-tests.sh
```diff
@@ -12,6 +12,7 @@ cp .travis/extensions.template.php var/config/extensions.php
 cp app/config/parameters.example.yml app/config/parameters.yml
 
 # install composer dependencies
+composer self-update --2
 composer config preferred-install source
```
blankse, line 16: @brusch Should be removed ;)

brusch: 👍😺 absolutely, thanks!

PR #7490, opened 2020-11-09
Fixes wrong file path in docs

rubanooo:
File path in last step does not match the paths specified above


brusch:
Thank you very much!

PR #7491, opened 2020-11-09
[Cache/Core] No error message if storeCacheItem() returns false

ingogascher:
Additional error message is not necessary, all messages are logged in storeCacheItem(). Not every return=false is an error, in this case we get a false negative.


brusch:
👍 makes sense! Thanks!

PR #7492, opened 2020-11-09
Fix formatDebugReceivers for rfc2822 3.6.2 validation

NiklasBr:
Resolves a bug where rfc2822 3.6.2 did not validate on re-sending email from the admin log due to:

https://github.com/pimcore/pimcore/blob/cf90256bcbd3dc55b7ad49159c26575d524c966e/lib/Helper/Mail.php#L408-L436

and Swiftmailer:

https://github.com/swiftmailer/swiftmailer/blob/f7d2961c858520dc83a594c66cde9d550f6d2226/lib/classes/Swift/Mime/Headers/MailboxHeader.php#L345-L357

## Additional info  

Pimcore would prior to this patch save the email in the Mail Log in the format `address@domain.ext (Name)` but it would then fail.

### DIFF
--- a/lib/Helper/Mail.php
+++ b/lib/Helper/Mail.php
@@ -129,17 +129,17 @@ CSS;
      */
     protected static function formatDebugReceivers(array $receivers)
     {
-        $tmpString = '';
-        foreach ($receivers as $mail => $name) {
-            $tmpString .= $mail;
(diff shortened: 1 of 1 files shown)


brusch:
Thanks  a lot! 👍

PR #7493, opened 2020-11-10
Update docs for Objects -> Select Types

MyZik:
The current example for getting the values from multiselect field doesn't work. So this PR fixes it


brusch:
Thank you!

PR #7495, opened 2020-11-10
Fix Notice: Undefined index: locked

blankse:
I get following error if I load a object grid with error reporting E_ALL:
image


brusch:
👍 Thank you so much!

PR #7500, opened 2020-11-10
[Data objects] Do not try to load deleted object bricks

BlackbitNeueMedien:
At our customer's system there was the problem that a data object could not be opened anymore in Pimcore backend because its object brick container contained an object brick which had been removed since last saving. I do not know exactly how the brick got deleted, I cannot reproduce the problem in the Pimcore demo. But nevertheless I think it is a good idea to check if the called dynamic method exists.


brusch:
Yep, makes sense, thanks a lot! 👍

PR #7502, opened 2020-11-10
Require `matomo/device-detector` instead of `piwik/device-detector`

jdreesen:
Piwik changed its name to Matomo some time ago and the package got a new name, too. The old one is abandoned.


brusch:
Nice, thanks!

PR #7503, opened 2020-11-10
Performance optimizations for Data Inheritance

aclDeveloper:

We experienced heavy performance degradation when using localized fields together with large data tables (300k+ objects). This is due to the architecture generating language-specific entries for every field available to the class. 

The current approach uses LIKE operators on path segments to determine the child objects to a relative parent. These are heavy optimizations in conjunction with the sorting of the path length and take ~500ms per query. When using [w-vision/data-definitions](https://pimcore.com/de/w-vision/data-definitions_e21825) to upsert 300k records we'd experience 

300k (records) x 500 (ms) x 3 (languages) 

of overhead for the localized fieldset alone. 

This is why we propose to change the resolution of the parent/child relation by using recursive SQL queries. Rather than using the object's path, we utilize the `o_parentId` column to resolve descendance across the element tree. In our case this increased overall performance by well over 1000%.


brusch:
@aclDeveloper Thanks a lot for your PR!
Currently the tests are failing, but this should be fixed since #7510 has been merged. Can you please rebase/merge your PR onto the latest upstream/master, that should fix the issue. Thanks!

aclDeveloper:
It seems we're still failing with the same problem. Local tests seem to work, I also wonder what is the difference between the Percona and Mysql/MariaDB images, as we don't have the former for investigative purposes.

brusch:
The problem I currently see with this PR is that it isn't compatible with MySQL 5.7. 
Since e.g. AWS Aurora isn't on 8.0 yet, I see a potential issue here for now.

brusch:
This PR also should include a proper test to ensure the query is not changing any behavior.

weisswurstkanone:
@aclDeveloper could you please have a look at the unit tests? thx!

aclDeveloper:
> @aclDeveloper could you please have a look at the unit tests? thx!

Hey, 

The tests for mysql 5.7 fail because CTEs are only supported from 8.0 onwards. This makes the query fail on older mysql releases. I though about the following options to tackle this and would like to discuss them here before altering the code:

* Check `server_version` in the database.yml and execute the old statement in case its not set or 5.7. This would maintain compatibility with existing solutions as nothing changes unless one configures the correct server version either thru .env or in the doctrine config.
* Implement the CTE as an anonymous procedure and execute it along with the query (see http://guilhembichot.blogspot.com/2013/11/with-recursive-and-mysql.html for reference)

Happy to hear about your opinions!

brusch:
We've decided to make MySQL 8 and MariaDB 10.3 as a minimum requirement for Pimcore X, so this shouldn't be an issue anymore. 
Can you please rebase/merge your PR onto the latest master? The tests should then run through. 

We have scheduled this PR for Pimcore X but before we need to check whether this functionality is fully covered by our tests.

aclDeveloper:
Even better, will do right away!

models/DataObject/Concrete/Dao/InheritanceHelper.php
```diff
@@ -418,15 +429,53 @@ protected function filterResultByLanguage($result, $language, $column)
...
+                    FROM objects p
+                    INNER JOIN cte on (p.o_parentId = cte.id)
+                ) SELECT l.language AS `language`,
+                         x.id AS id,
+                         x.classId AS classId,
+                         x.parentId AS parentId
+                         $fields
+                    FROM cte x
+                    LEFT JOIN $storeTable l ON x.id = l.$idfield
+                   WHERE COALESCE(`language`, " . $this->db->quote($language) . ") = " . $this->db->quote($language) .
```
weisswurstkanone, line 457: Hi @aclDeveloper can you explain with an example how the coalesce is equivalent to the having clause as it was before?

aclDeveloper: So the original query utilizes a `HAVING` clause to implicitly group languages together. However there is no `GROUP BY` clause present, so it just behaves as the `WHERE` would, as you can see in the following 2 examples:

image

image

With the recursive query we more or less have the same behavior. I propose to put the language filter in the `WHERE` clause because there's no `GROUP BY` and so its easier to read for people who are not extra-fond in SQL. 

image

The use of COALESCE is used as a short-hand here so that we don't have to incorporate the `OR ISNULL(language)` part. If we replace NULL-Values with the expected value, we can get the same result and save on one more condition on an already complicated query.

brusch:
@aclDeveloper well done, thanks a lot!

Issue #6771, opened 2020-07-13
[Permissions] Inconsistent permissions implementation


fashxp:
Permissions implementation for search and tree listing is not consistent. 

![image](https://user-images.githubusercontent.com/8792145/87313558-54c55580-c522-11ea-825e-4358373de325.png)

- navigation in tree shows alfa romeo cars
- search doesn't find any alfa romeo cars

additional aspect 
- bad performance impact of permission implementation in tree navigation (see #6383) 
- maybe also #1634
- implementation of CMF would also not find alfa romeo cars, see https://github.com/pimcore/customer-data-framework/blob/master/src/Listing/Filter/Permission.php#L11 - but is much faster

ckogler-elements:
I had just a similar issue. 
In my use case the user had workspace permissions and a lot of items matched where the user did not have the list permission. Because the first matches were empty (no permissions to the objects) the "real matches" didn't show up. 
Therefore the workspaces paths should be added to the DB query to get the right data. 

I just solved it by creating an event listener but it would be great if the core also handles workspaces...

`
class SearchBackend
{

    public function searchAdmin(\Symfony\Component\EventDispatcher\GenericEvent $event){

        $searcherList = $event->getArgument('list');
        $context =$event->getArgument('context');
        if($context['type'] == 'object'){
            $db = \Pimcore\Db::get();
            $user = \Pimcore\Tool\Authentication::authenticateSession();
            if(!$user->isAdmin()){
                $validPaths = [];
                foreach($user->getWorkspacesObject() as $ws){
                    if($ws->getList()){
                        $validPaths[] = $ws->getCpath();
                    }
                }

                foreach($user->getRoles() as $rId){
                    $role = Role::getById($rId);
                    /**
                     * @var $ws \Pimcore\Model\User\Workspace\DataObject
                     */
                    foreach($role->getWorkspacesObject() as $ws){
                        if($ws->getList()){
                            $validPaths[] = $ws->getCpath();
                        }
                    }
                }
                $validPaths = array_filter(array_unique($validPaths));

                if($validPaths){
                    $queryParts = [];
                    foreach($validPaths as $path){
                        $queryParts[] = ' ' . $db->quoteIdentifier('fullpath') . ' LIKE '. $db->quote($path."%");
                    }
                    $q = implode(' OR ',$queryParts);
                    $searcherList->setCondition($searcherList->getCondition().' AND (`type` = "object" AND (' .$q.'))'  );
                }
            }
        }

    }
}
[...]

PR #7506, opened 2020-11-10
[Dashboard] Show all elements in widget changed objects / assets / documents

BlackbitDevs:
Currently the changed objects / assets / documents widget shows only elements where the last change has been made by the current user. This PR changes this behaviour to display all elements which the user has the permission to view.

Filtering for current user does especially not make sense because when user A changes object X, it appears in his dashboard. When then user B changes object X, it disappears from A's dashboard. Then user A wonders where his changes are or if he actually saved the element.


weisswurstkanone:
relevant change: https://github.com/pimcore/pimcore/pull/3270

BlackbitNeueMedien:
@weisswurstkanone Actually I do not understand #3270. What has the portlet / widget to do with the edit permission? For example in https://github.com/pimcore/pimcore/blob/e3e542a4f3dc4dafdbcfc05bdec7142682c4d934/bundles/AdminBundle/Controller/Admin/PortalController.php#L339 it is checked anyway if the current user has access to the object. 

And what has it to do with the "rememnbered tabs" function?

weisswurstkanone:
Not sure either, just wanted to add a reference to the original change.
@brusch , @ckogler-elements can you provide your view on this ?

BlackbitNeueMedien:
See a problem with this PR currently: If the user hasn't view permission for one of the 10 loaded items, there won't be 10 results - in the worst case there will be 0.
But I think there is a table where the objects which a certain user has access to are in - have seen a join somewhere I think...

ckogler-elements:
Well, if i remember it correctly the main problem was that all items were displayed although the user didn't had permissions to view them. When the user tried to open the item he got an error. Everytime the user logged in to pimcore he got an error because i guess its the pimcore.helpers.openMemorizedTabs tried to open this item where the user didn't had permissions to. To handle it "correctly" i guess the pimcore.helpers.openObject... would have to check if the item is still viewable by the user.

The reason for filtering for the user was, that the user almost never got relevant items for him E.g. A importer ran and saved 100 objects -> they would a appear in the list...

Maybe showing all modified objects with an additional colum for filter the user would make sense? As Blackbit mentioned the filtering should be done in DB as the user otherwise might get no results.

BlackbitNeueMedien:
The mentioned permission tables are `users_workspaces_object`, `users_workspaces_asset` and `users_workspaces_document`. Will try to join to those like it is done in https://github.com/pimcore/pimcore/blob/e3e542a4f3dc4dafdbcfc05bdec7142682c4d934/models/DataObject/AbstractObject/Dao.php#L459

brusch:
@BlackbitNeueMedien yep, makes sense! Thanks

BlackbitDevs:
I implemented the view permission check with a SQL condition.

models/Asset.php
```diff
@@ -403,7 +403,7 @@ public static function create($parentId, $data = [], $save = true)
     /**
      * @param array $config
      *
-     * @return mixed
+     * @return Listing
```
fashxp, line 406: this has nothing to do with this PR

models/Asset.php
```diff
@@ -415,6 +415,7 @@ public static function getList($config = [])
 
         $listClass = Listing::class;
 
+        /** @var Listing $list */
```
fashxp, line 418: this has nothing to do with this PR

models/Asset/Listing.php
```diff
@@ -79,19 +79,20 @@ public function getItems($offset, $itemCountPerPage)
...
-    public function filterAccessibleByUser(Model\User $user)
+    public function filterAccessibleByUser(Model\User $user, $permission = 'list')
     {
         if (!$user->isAdmin()) {
             $userIds = $user->getRoles();
             $userIds[] = $user->getId();
 
             $condition = '(
-                (SELECT list FROM users_workspaces_asset WHERE userId IN ('.implode(',', $userIds).') AND LOCATE(CONCAT(path,filename),cpath)=1 ORDER BY LENGTH(cpath) DESC, FIELD(userId, '.$user->getId().') DESC, list DESC LIMIT 1)=1
+                (SELECT `'.$permission.'` FROM users_workspaces_asset WHERE userId IN ('.implode(',', $userIds).') AND LOCATE(CONCAT(path,filename),cpath)=1 ORDER BY LENGTH(cpath) DESC, FIELD(userId, '.$user->getId().') DESC, `'.$permission.'` DESC LIMIT 1)=1
```
fashxp: why not use `$this->quote($permission)` here?

models/DataObject/Listing.php
```diff
@@ -184,4 +184,29 @@ public function addFilterByField($field, $operator, $data)
...
+     * @internal
+     */
+    public function filterAccessibleByUser(Model\User $user, $permission = 'list')
+    {
+        if (!$user->isAdmin()) {
+            $userIds = $user->getRoles();
+            $userIds[] = $user->getId();
+
+            $condition = '(
+                (SELECT `'.$permission.'` FROM users_workspaces_object WHERE userId IN ('.implode(',', $userIds).') AND LOCATE(CONCAT(o_path,o_key),cpath)=1 ORDER BY LENGTH(cpath) DESC, FIELD(userId, '.$user->getId().') DESC, `'.$permission.'` DESC LIMIT 1)=1
```
fashxp: same here, why not use `$this->quote()`

models/Document/Listing.php
```diff
@@ -131,4 +131,30 @@ public function getItems($offset, $itemCountPerPage)
...
+     * @internal
+     */
+    public function filterAccessibleByUser(Model\User $user, $permission = 'list')
+    {
+        if (!$user->isAdmin()) {
+            $userIds = $user->getRoles();
+            $userIds[] = $user->getId();
+
+            $condition = '(
+                (SELECT `'.$permission.'` FROM users_workspaces_document WHERE userId IN ('.implode(',', $userIds).') AND LOCATE(CONCAT(path,`key`),cpath)=1 ORDER BY LENGTH(cpath) DESC, FIELD(userId, '.$user->getId().') DESC, `'.$permission.'` DESC LIMIT 1)=1
```
fashxp: same here, why not use `$this->quote()`

bundles/AdminBundle/Controller/Admin/PortalController.php
```diff
@@ -291,24 +294,24 @@ public function portletModifiedAssetsAction(Request $request)
...
              */
-            if ($doc->isAllowed('view')) {
-                $response['assets'][] = [
-                    'id' => $doc->getId(),
-                    'type' => $doc->getType(),
-                    'path' => $doc->getRealFullPath(),
-                    'date' => $doc->getModificationDate(),
-                ];
-            }
+            $response['assets'][] = [
```
fashxp: shouldn't we keep that check for extra security? we also have them in the tree controllers...

bundles/AdminBundle/Controller/Admin/PortalController.php
```diff
@@ -327,21 +330,21 @@ public function portletModifiedObjectsAction(Request $request)
...
         foreach ($list as $object) {
-            if ($object->isAllowed('view')) {
-                $response['objects'][] = [
-                    'id' => $object->getId(),
-                    'type' => $object->getType(),
-                    'path' => $object->getRealFullPath(),
-                    'date' => $object->getModificationDate(),
-                ];
-            }
+            $response['objects'][] = [
```
fashxp: shouldn't we keep that check for extra security? we also have them in the tree controllers...

fashxp (changes requested):

models/Asset.php
BlackbitDevs: Well it has, otherwise the class of $list in https://github.com/pimcore/pimcore/pull/7506/files#diff-28f27197a7d7ff638b16394c0ae3c22e477459333caedc93e91fc02f719d6220R300 cannot be determined.

models/Asset.php
BlackbitDevs: Without this PhpStan test fails (see https://github.com/pimcore/pimcore/runs/2782266277) because `self::getModelFactory()->build($listClass)` returns `AbstractModel` (when we have `@return Listing` in the PHPDoc)

models/Asset/Listing.php
BlackbitDevs: `$this->quote()` adds quotes(`'`) but these are database column names, here we need backticks (`)

models/DataObject/Listing.php
BlackbitDevs: Let's discuss in above conversation. When we find another method there, I will apply it also here.

models/Document/Listing.php
BlackbitDevs: Let's discuss in above conversation. When we find another method there, I will apply it also here.

bundles/AdminBundle/Controller/Admin/PortalController.php
BlackbitDevs: Let's discuss in above conversation. When we find another method there, I will apply it also here.

bundles/AdminBundle/Controller/Admin/PortalController.php
BlackbitDevs: No, `isAllowed()` executes the same database queries again but then of course for every item in the loop which is always a bad idea:
Compare
https://github.com/pimcore/pimcore/blob/86ab63e336ce5d36201370381b3aaf177e56b435/models/Asset/Dao.php#L429-L449
with https://github.com/pimcore/pimcore/blob/86ab63e336ce5d36201370381b3aaf177e56b435/models/Asset/Listing.php#L85-L101
But of course then for every item in the loop! So beside being slow this is also redundant.

These `filterAccessibleByUser` methods in the Listings have potential to be used for all permission checks in the backend within the actual data-fetching query. No additional `isAllowed()` calls in loops necessary anymore. This will result in big performance improvements...

models/Asset/Listing.php
ctippler: I guess $db->quoteIdentifier() would do the trick.

BlackbitDevs: Ok, I changed it to `Db::get()->quoteIdentifier($permission)`.
Actually I cannot understand such pickiness while in other places it is implemented exactly how I had it with the backticks, see for example https://github.com/pimcore/pimcore/blob/86ab63e336ce5d36201370381b3aaf177e56b435/models/Asset/Dao.php#L430

fashxp: Thank you very much and sorry for that pickiness ... but better safe than sorry. 
The simple backticks approach opens up the door for SQL injection. Just use ```$type = 'list` FROM assets; DELETE assets; -- ';``` as `$type` and bang. And when we already touch things, we should do them right. 

Besides that, we also need to fix the other places too of course ... see https://github.com/pimcore/pimcore/issues/9412

models/Asset.php
fashxp: good point, sorry.

models/Asset.php
fashxp: I see

bundles/AdminBundle/Controller/Admin/PortalController.php
fashxp: Again sorry for the pickiness, but that is not the case. 

The query of `filterAccessibleByUser` compares based on path, the `isAllowed` compares based on ids. There are edge cases, where that matters ... e.g. constellations like 
image

These constellations are filtered with the `isAllowed` call - also in the trees and grids, see for example [here](https://github.com/pimcore/pimcore/blob/092422059721cb76426194c60cddab433ce1d3c2/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L216) or [here](https://github.com/pimcore/pimcore/blob/092422059721cb76426194c60cddab433ce1d3c2/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L1814).

Besides that, `isAllowed` fires an `AdminEvents::ELEMENT_PERMISSION_IS_ALLOWED` event - which of course needs to be fired in the dashboard case too. 

So it is about consistency and security. Permissions is a very delectat topic and thus we have to be very careful there. 

In general, there is room for improvement for the whole permission implementation, especially when it comes to performance, see also https://github.com/pimcore/pimcore/issues/6771 ... but still, what we change/improve must not introduce any regressions.

BlackbitDevs: Ok, I added the `isAllowed()`. Perhaps we will find a correct way in #6771 which also enhances performance.

fashxp (approved):

fashxp:
from a technical perspective, this should be fine now. 
from a UX perspective we once more discussed internally and came to the conclusion, that we should not change default behavior of existing widgets. we should either add a config option or add additional widgets that contain all modified elements.

brusch:
@BlackbitNeueMedien can you update your PR according to @fashxp 's feedback? Thanks a lot!

brusch:
I'm closing this PR now due to inactivity. 
Please just open up a new PR if you like to contribute the feature as described above. 
Thanks!

PR #7508, opened 2020-11-10
Pimcore::Tool wrong merging of parameters in useFrontendOutputFilters

weisswurstkanone:


lib/Tool.php
```diff
@@ -382,10 +382,10 @@ public static function useFrontendOutputFilters(Request $request = null)
             return false;
         }
 
-        $requestKeys = array_merge([
+        $requestKeys = array_merge(
             array_keys($request->query->all()),
             array_keys($request->request->all()),
```
blankse, line 387:
```suggestion
            array_keys($request->request->all())
```

PR #7513, opened 2020-11-11
Fatal with caching on localized fields caused by ElementDescriptor

### LINKED ISSUE #7367: Fatal with caching on localized fields caused by ElementDescriptor
This is caused by the CoreHandler in the latest version. If i revert it to the state before the error is gone.
I think it relates to this:
https://github.com/pimcore/pimcore/issues/7039

grafik

weisswurstkanone:
fixes #7367


brusch:
@weisswurstkanone can you please check Travis

weisswurstkanone:
phpstan is not happy.

weisswurstkanone:
@brusch should be OK now

PR #7519, opened 2020-11-12
Fix batch edit null error

t-zilla:

* Fix null pointer error

## Additional info 

Object ID is stored in `this.data`, as indicated on line 471 `if (this.data.id)`
Causes a console error during batch edit


brusch:
Thanks a lot! 👍

PR #7522, opened 2020-11-12
Auto save / Drafts for objects and documents

ckogler-elements:
Implement Autosave functionality for documents/objects.

To enable it, set the saving interval for the items to > 0 in your paramter.yml

e.g.
parameters:
    pimcore.admin.draft_saving_interval_document : 3
    pimcore.admin.draft_saving_interval_object: 5


models/Version/Listing.php
```diff
@@ -21,7 +21,7 @@
 
 /**
  * @method \Pimcore\Model\Version\Listing\Dao getDao()
- * @method \Pimcore\Model\Version\Listing\Dao loadIdList()
+ * @method array loadIdList()
```
NiklasBr, line 24: Is this correct? I think it should actually reference the class

dvesh3:
@ckogler-elements nice feature 🚀 please rebase it on master, as milestone is set to Pimcore X(10.0.0). thanks!

brusch:
@ckogler-elements ping 🏓

ckogler-elements:
@brusch : is changed

brusch:
Thank a lot! One thing that comes directly to my mind is that we should stop the saving interval as soon as a tab goes to background (`hide` event on tab panel). Otherwise this could have bad impact on browser-performance. Wdyt?

ckogler-elements:
Yes i absolutely agree with that. https://github.com/pimcore/pimcore/pull/7522/commits/53b46ead3a62a7b0f1e81545faae8f2cffcd194d should do the trick.

brusch:
@ckogler-elements yep, but it should be re-enabled again when activated.

ckogler-elements:
Should be the case when this.tab.on("activate", this.startChangeDetector.bind(this)); is called. This calls checkForChanges() and this starts the draft saving if its dirty. 
The problem is that i currently can't test it because i don't have a pimcore instance with the changes anymore ;-)

brusch:
Squashed this PR into #8838 to make some further changes.

PR #7531, opened 2020-11-13
Move Twig Helpers from coreshop/pimcore to pimcore/pimcore

dpfaffenbauer:
These are quite some handy twig extensions that should be part of the Core.

@brusch there are many more things in coreshop/pimcore that might make sense, if you mgiht take a look at it what you think might help here as well: https://github.com/coreshop/CoreShop/tree/master/src/CoreShop/Component/Pimcore

### DIFF
--- a/lib/Twig/Extension/AssetHelperExtensions.php
+++ b/lib/Twig/Extension/AssetHelperExtensions.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Twig\Extension;
+
+use Pimcore\Model\Asset;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigTest;
+
+final class AssetHelperExtensions extends AbstractExtension
+{
+    public function getTests(): array
+    {
+        return [
+            new TwigTest('pimcore_asset', static function ($object) {
+                return is_object($object) && $object instanceof Asset;
+            }),
+            new TwigTest('pimcore_asset_archive', static function ($object) {
+                return is_object($object) && $object instanceof Asset\Archive;
+            }),
+            new TwigTest('pimcore_asset_audio',static  function ($object) {
+                return is_object($object) && $object instanceof Asset\Audio;
+            }),
+            new TwigTest('pimcore_asset_document', static function ($object) {
+                return is_object($object) && $object instanceof Asset\Document;
+            }),
+            new TwigTest('pimcore_asset_folder', static function ($object) {
+                return is_object($object) && $object instanceof Asset\Folder;
+            }),
+            new TwigTest('pimcore_asset_image', static function ($object) {
+                return is_object($object) && $object instanceof Asset\Image;
+            }),
+            new TwigTest('pimcore_asset_text', static function ($object) {
+                return is_object($object) && $object instanceof Asset\Text;
+            }),
+            new TwigTest('pimcore_asset_unknown', static function ($object) {
+                return is_object($object) && $object instanceof Asset\Unknown;
+            }),
+            new TwigTest('pimcore_asset_video', static function ($object) {
+                return is_object($object) && $object instanceof Asset\Video;
+            }),
+        ];
+    }
+}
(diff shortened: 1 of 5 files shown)


lib/Twig/Extension/AssetHelperExtensions.php
```diff
@@ -0,0 +1,58 @@
...
+use Twig\Extension\AbstractExtension;
+use Twig\TwigTest;
+
+final class AssetHelperExtensions extends AbstractExtension
+{
+    public function getTests(): array
+    {
+        return [
+            new TwigTest('pimcore_asset', static function ($object) {
+                return is_object($object) && $object instanceof Asset;
```
jdreesen, line 30: The `is_object` check should not be needed. 
The PHP docs say: 
`instanceof does not throw any error if the variable being tested is not an object, it simply returns FALSE. Constants, however, were not allowed prior to PHP 7.3.0.`

```suggestion
                return $object instanceof Asset;
```

dpfaffenbauer: true, not sure why I used that here, it for sure had a reason :) but might be to older PHP Versions, will remove it.

dpfaffenbauer:
@jdreesen done

brusch:
Thanks a lot for your PR! 
Regarding other Twig extensions ... we're quite open here, proposals are definitely welcome! 🙏😊

brusch:
I'll leave this PR open until we have a complete set of helpers we gonna want to have in the core.

dpfaffenbauer:
@brusch the question is what helpers ;). Some seem a bit outdated (like coreshop_path, since I think you already implemented something to alle path(data_object) right?)

dpfaffenbauer:
@dvesh3 commited your changes

dvesh3:
> @dvesh3 commited your changes

thanks! I guess we also have to change the type hints.

dpfaffenbauer:
> > @dvesh3 commited your changes
> 
> thanks! I guess we also have to change the type hints.

Thinking about it, since it is twig helpers, I think we should not allow Thumnbail or array as a config. I am using this helpers for years now and never had the need to use anything else than string.

dvesh3:
> > > @dvesh3 commited your changes
> > 
> > 
> > thanks! I guess we also have to change the type hints.
> 
> Thinking about it, since it is twig helpers, I think we should not allow Thumnbail or array as a config. I am using this helpers for years now and never had the need to use anything else than string.

 @dpfaffenbauer TBH not sure, if we really need these functions since you can call getThumbnail directly on image like,. 
```twig
{{ image.thumbnail('exampleScaleWidth').html({
    'class': 'thumbnail-class',
    'data-my-name': 'my value',
}) }}
```

dpfaffenbauer:
@dvesh3 this is more of a separation of concerns. It bugs me that the model classes in Pimcore have too much responsability. The image itself should not be responsible for creating the thumbnail, a service should do that. this PR will make Pimcore future proof in case this ever gets refactored.

dvesh3:
> @dvesh3 this is more of a separation of concerns. It bugs me that the model classes in Pimcore have too much responsability. The image itself should not be responsible for creating the thumbnail, a service should do that. this PR will make Pimcore future proof in case this ever gets refactored.

@dpfaffenbauer got your point 👍 let's revert my suggestions then 😉

lib/Twig/Extension/ImageThumbnailExtension.php
```diff
@@ -0,0 +1,79 @@
...
+     * @param string $thumbnail
+     * @param array  $options
+     * @param array  $removeAttributes
+     * @param bool   $deferred
+     *
+     * @return string
+     */
+    public function getImageThumbnailHtml(
+        Image $image,
+        string|array|Image\Thumbnail\Config $thumbnail,
```
dvesh3, line 72:
```suggestion
        string $thumbnail,
```

lib/Twig/Extension/ImageThumbnailExtension.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return [
+            new TwigFunction('pimcore_image_thumbnail', [$this, 'getImageThumbnail'], ['is_safe' => ['html']]),
+            new TwigFunction('pimcore_image_thumbnail_html', [$this, 'getImageThumbnailHtml'], ['is_safe' => ['html']]),
+        ];
+    }
+
+    /**
+     * @param Image  $image
+     * @param string|array|Image\Thumbnail\Config $thumbnail
```
dvesh3, line 51:
```suggestion
     * @param string $thumbnail
```

dpfaffenbauer:
🎉

dvesh3:
@dpfaffenbauer thanks a lot 🙏

brusch:
Sorry, one more, thing  .... docs 😉
Can someone please add them here: 
https://github.com/pimcore/pimcore/blob/34896ec5737a49cd5ea51898391fefb441c947d2/doc/Development_Documentation/02_MVC/02_Template/02_Template_Extensions/README.md#L115

Please also check if there are some others which need to be added. 
Thanks.

PR #7534, opened 2020-11-13
Blocks with many entries causes fatal error / probably infinite loop …

### LINKED ISSUE #7297: Blocks with many entries causes fatal error / probably infinite loop
# Bug Report

### Expected behavior

- Objects with several block items load correctly

### Actual behavior

- Objects with several block items won't load

### Steps to reproduce

(Tested on live demo system, pimcore 6.8.x-dev)

- Create Class (block with Advanced Many-to-Many Object Relations with 1 meta data field for car objects)

Bildschirmfoto 2020-10-13 um 08 57 02

- Create object and populate block attribute with about 15 entries (maybe some more or less) and assign car-objects with some META data

- Save the object

- Reload or click in tree

- Check php.log: 

Bildschirmfoto 2020-10-13 um 08 56 36

weisswurstkanone:
fixes #7297


models/DataObject/Data/BlockElement.php
```diff
@@ -154,6 +155,18 @@ function ($currentValue) {
...
         );
+
+        $copier->addFilter(new \DeepCopy\Filter\KeepFilter(), new class() implements \DeepCopy\Matcher\Matcher {
+            public function matches($object, $property)
+            {
+                if ($object instanceof AbstractElement) {
+                    return true;
+                }
+
+                return false;
```
jdreesen, line 166: Could be simplified to:

```suggestion
                return $object instanceof AbstractElement;
```

PR #7537, opened 2020-11-13
Fixes array_to_html_attribute_string to allow value-less attributes

NiklasBr:
Allows attributes such as `data-*`, `contenteditable`, `required`, or `autofocus` now which may be used without any value by using the attribute name as key and `null` as the value.

### DIFF
--- a/lib/helper-functions.php
+++ b/lib/helper-functions.php
@@ -222,7 +222,7 @@ function array_toquerystring($args)
 }
 
 /**
- * @param array $array
+ * @param array $array with attribute names as keys, and values as values
  *
  * @return string
  */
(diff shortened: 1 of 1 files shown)


lib/helper-functions.php
```diff
@@ -235,6 +235,8 @@ function array_to_html_attribute_string($array)
                 $data .= ' ';
             }
             $data .= $key . '="' . htmlspecialchars($value) . '"';
+        } elseif (is_string($key) && is_null($value)) {
+            $data .= $key;
```
dvesh3, line 239: space missing between multiple items.

NiklasBr: Thanks, done

NiklasBr:
Something odd with Travis today:

> Your requirements could not be resolved to an installable set of packages.

dvesh3:
> Something odd with Travis today:

please see https://github.com/pimcore/pimcore/pull/7541

dvesh3:
thanks!

Issue #5827, opened 2020-02-17
Update Codeception


brusch:
To the latest version. 
Also update here: 
https://github.com/pimcore/demo/blob/master/composer.json#L24
https://github.com/pimcore/skeleton/blob/master/composer.json#L16

brusch:
See also https://github.com/pimcore/skeleton/issues/18

brusch:
There seem to be problems with Symfony 3.4 support in newer version of codeception, so I'm afraid we have to wait until Pimcore v7. 
Or did anyone else got it to work? 😊

jdreesen:
Can you give more details about the problem? 
We used Codeception v3 for the acceptance tests (with Selenium) in one of our projects and it worked fine.

weisswurstkanone:
As far as I remember we struggled a lot to get it working for both Symfony 3.4 and 4 at the same time. 
We would be happy about any input about this topc. Would be nice if you could share your experiences ;-)

brusch:
Resolved by #7542

PR #7542, opened 2020-11-16
Update Codeception #5827

weisswurstkanone:


tests/_support/Helper/DataType/TestDataHelper.php
```diff
@@ -1187,8 +1187,9 @@ public function assertObjectsWithMetadata(Concrete $object, $field, Concrete $co
      */
     protected function assertObjectMetadataEqual($expected, $value)
     {
-        $this->assertInternalType('array', $expected);
-        $this->assertInternalType('array', $value);
+        // see https://github.com/sebastianbergmann/phpunit/commit/50ad7e1c4e74dce3beff17bf9c9f5a458cbe9958
+        $this->assertTrue(is_array($expected), "expected an array");
+        $this->assertTrue(is_array($value), "expected an array");
```
jdreesen, line 1192: `assertInternalType` got removed in PHPUnit 9, while PHPUnit 8 added `assertIsArray` for this type of check:

```suggestion
        $this->assertIsArray($expected);
        $this->assertIsArray($value);
```

<br>

But as it seems that you're still using PHPUnit 7, I'm not sure why this change is necessary at the moment.

weisswurstkanone: In this case I would say let's try to keep it under our own control instead of relying on specialized assertion methods which are now or then depecrated or need an replacement and basically do the same. It is the only occurence.

weisswurstkanone: TestDataHelper basically extends from Module which used this Asset codeception trait. But it does not implement it anymore.

```
namespace Codeception;

use Codeception\Exception\ModuleException;
use Codeception\Lib\Interfaces\RequiresPackage;
use Codeception\Lib\ModuleContainer;
use Codeception\Util\Shared\Asserts;

/**
 * Basic class for Modules and Helpers.
 * You must extend from it while implementing own helpers.
 *
 * Public methods of this class start with `_` prefix in order to ignore them in actor classes.
 * Module contains **HOOKS** which allow to handle test execution routine.
 *
 */
abstract class Module
{
  => use Asserts;
```

It seems that it in the past this was just delegated to

```
    protected function assertInternalType($type, $actual, $description = '')
    {
        \PHPUnit\Framework\Assert::assertInternalType($type, $actual, $description);
    }
```

which is not the case now. As said, I would prefer to be on the safe and flexible side :-)

PR #7543, opened 2020-11-16
Upgrade Doctrine Bundles, introduce new Migrations & Installer

### LINKED ISSUE #7408: Update Doctrine dependencies to latest version

brusch:
Resolves #7408


brusch:
Extending the functionality of `DoctrineMigrationsBundle` is not any longer possible the way we did it in the past. Therefore we're switching to standard Doctrine migrations, this means also that migration sets are not supported anymore and that the available migrations have to be configured manually or by using flex. 

```yml
doctrine_migrations:
    migrations_paths:
        'Pimcore\Bundle\DataHubBundle\Migrations': '@PimcoreDataHubBundle/Migrations'
        'CustomerManagementFrameworkBundle\Migrations': '@PimcoreCustomerManagementFrameworkBundle/Migrations'
```

However, we've extended the doctrine commands to accept an optional `--prefix` option, which let's you filter configured migration classes. This is in a way an adequate replacement for the `-s` (sets) option.  

`./bin/console doctrine:migrations:list --prefix=Pimcore\\Bundle\\CoreBundle`

PR #7545, opened 2020-11-16
[Workflows] Fix deeplink for workflow notifications

justbane:
Resubmitting to proper branch (6.8) for bugfix

  
  

Resolves #

## Additional info  

I am getting object links when I am editing pages.

For instance here is the deeplink to the page: https://www.site.com/admin/login/deeplink?document_1824_page

Here is the link that came in the email when i edited the page and submitted the workflow: https://www.site.com/admin/login/deeplink?object_1824_object

The ID is correct but the link is wrong.

### DIFF
--- a/lib/Workflow/Notification/NotificationEmailService.php
+++ b/lib/Workflow/Notification/NotificationEmailService.php
@@ -90,7 +90,7 @@ class NotificationEmailService extends AbstractNotificationService
                     $type = $subject->getType();
                 }
 
-                $deeplink = $hostUrl . $this->router->generate('pimcore_admin_login') . '/deeplink?'.$objectType.'_' . $subject->getId() . '_'. $type;
+                $deeplink = $hostUrl . $this->router->generate('pimcore_admin_login_deeplink') . '?'.$objectType.'_' . $subject->getId() . '_'. $type;
             }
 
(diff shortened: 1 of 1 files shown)


lib/Workflow/Notification/NotificationEmailService.php
```diff
@@ -78,7 +78,19 @@ public function sendWorkflowEmailNotification(array $users, array $roles, Workfl
...
+                if ($subject instanceof \Pimcore\Model\Document) {
+                    $objectType = "document";
+                    $type = $subject->getType();
+                }
+                if ($subject instanceof \Pimcore\Model\Asset) {
+                    $objectType = "asset";
+                    $type = $subject->getType();
+                }
+
+                $deeplink = $hostUrl . $this->router->generate('pimcore_admin_login') . '/deeplink?'.$objectType.'_' . $subject->getId() . '_'. $type;
```
blankse, line 93: We should also use the correct route:
```suggestion
                $deeplink = $hostUrl . $this->router->generate('pimcore_admin_login_deeplink') . '?'.$objectType.'_' . $subject->getId() . '_'. $type;
```

dvesh3:
thank you!

PR #7546, opened 2020-11-17
Add setItems method to CartInterface

blankse:
Is used here:
https://github.com/pimcore/pimcore/blob/b926dc320b5ccda17e20665303ca239404a3d821/bundles/EcommerceFrameworkBundle/CartManager/SessionCartItem.php#L42


brusch:
Thanks!

PR #7547, opened 2020-11-17
Add getArgument method to ElementEventInterface

blankse:
Is used very often. 
For example here:
https://github.com/pimcore/pimcore/blob/b926dc320b5ccda17e20665303ca239404a3d821/bundles/CoreBundle/EventListener/ElementTagsListener.php#L50

### DIFF
--- a/lib/Event/Model/ElementEventInterface.php
+++ b/lib/Event/Model/ElementEventInterface.php
@@ -22,4 +22,15 @@ interface ElementEventInterface
      * @return ElementInterface
      */
     public function getElement();
+
+    /**
+     * Get argument by key.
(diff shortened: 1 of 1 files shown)


brusch:
Should we add the other methods defined here as well, or are they not used anywhere? 
https://github.com/pimcore/pimcore/blob/c62924ca5c3d0c482a71097cb0231aff143208d6/lib/Event/Traits/ArgumentsAwareTrait.php#L17

blankse:
@brusch Yes make sense to add all the arguments methods. I changed it.

brusch:
Thanks 😎👌

PR #7548, opened 2020-11-17
Add getId method to ProductInterface

blankse:
Is used very often. For example:
https://github.com/pimcore/pimcore/blob/b926dc320b5ccda17e20665303ca239404a3d821/bundles/EcommerceFrameworkBundle/Model/AbstractSetProductEntry.php#L82


brusch:
Thanks!

PR #7549, opened 2020-11-17
Add getRule method to ModificatedPriceInterface

blankse:
Is used here:
https://github.com/pimcore/pimcore/blob/b926dc320b5ccda17e20665303ca239404a3d821/bundles/EcommerceFrameworkBundle/OrderManager/OrderManager.php#L349
https://github.com/pimcore/pimcore/blob/b926dc320b5ccda17e20665303ca239404a3d821/bundles/EcommerceFrameworkBundle/OrderManager/V7/OrderManager.php#L136


brusch:
Thanks!

PR #7550, opened 2020-11-17
Numeric filter fix

### LINKED ISSUE #7423: Floating point filtering issue
## Expected behavior
I can filter number fields using float values.

### Actual behavior
When filtering using floating point values, e.g. `0.00002`, no results are shown. Also, removing filters doesn't reverts to original grid.

### Steps to reproduce
1. Create a numeric field with decimal precision of at least 5, or with no precision.
2. Set the value in object editor to 0.00002
Zrzut ekranu z 2020-10-30 12-17-39

3. Open the grid, filter for objects with value equal to 0.00002 - no results are found
Zrzut ekranu z 2020-10-30 12-17-54

4. Remove the filters - all the results are missing now, folder reload is needed
Zrzut ekranu z 2020-10-30 12-18-09

5. Using direct sql query - results are shown
Zrzut ekranu z 2020-10-30 12-19-08

### Possible related issue
It may be related to #5088 ?

t-zilla:
Resolves #7423

Numeric grid filters now have the same `decimalPrecision` as in the class definition. Previously it defaulted to 2.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
@@ -289,6 +289,10 @@ pimcore.object.helpers.grid = Class.create({
                     if (tag) {
                         var fc = tag.prototype.getGridColumnConfig(field);
                         fc.width = this.getColumnWidth(field, 100);
+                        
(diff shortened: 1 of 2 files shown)


weisswurstkanone:
thanks a lot!

PR #7552, opened 2020-11-17
[Data Object] Table Data-Type should always return an array

### LINKED ISSUE #6661: [Data Object] Table Data-Type should always return an array
At the moment the data-type returns and empty string or even `null` if no value is set. 
With v7 it should always return an array, in case of no data, just an empty array (`[]`). 

See original discussion here: #6532

weisswurstkanone:
fixes #6661

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Table.php
+++ b/models/DataObject/ClassDefinition/Data/Table.php
@@ -300,7 +300,7 @@ class Table extends Data implements ResourcePersistenceAwareInterface, QueryReso
      * @param null|DataObject\Concrete $object
      * @param mixed $params
      *
-     * @return string
+     * @return string|null
      */
     public function getDataForResource($data, $object = null, $params = [])
     {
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/Table.php
```diff
@@ -326,7 +331,7 @@ public function getDataFromResource($data, $object = null, $params = [])
         $unserializedData = Serialize::unserialize((string) $data);
 
         if ($data === null || $unserializedData === null) {
-            return null;
+            return [];
```
blankse, line 334: Change also PHPDoc:
`@return array`

weisswurstkanone: done

models/DataObject/ClassDefinition/Data/Table.php
```diff
@@ -100,7 +100,7 @@ class Table extends Data implements ResourcePersistenceAwareInterface, QueryReso
      *
      * @var string
      */
-    public $phpdocType = 'array|string';
+    public $phpdocType = 'array';
```
dvesh3, line 103: we should rebuild classes to update phpdoc

dvesh3: nevermind! It should be covered in a generic task for other changes on v7.

PR #7553, opened 2020-11-17
Make AbstractOrder/AbstractOrderItem methods abstract and add missing methods

blankse:
Add following methods:

- AbstractOrder::getCartHash()
- AbstractOrderItem::getComment()
- AbstractOrderItem::setComment()

Removed unused methods:

- AbstractOrder::getDeliveryEMail()
- AbstractOrder::setDeliveryEMail()
- AbstractOrder::getCartModificationTimestamp()

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Model/AbstractOrder.php
+++ b/bundles/EcommerceFrameworkBundle/Model/AbstractOrder.php
@@ -426,4 +426,11 @@ abstract class AbstractOrder extends Concrete
      * @return \Pimcore\Model\DataObject\OnlineShopOrder
      */
     abstract public function setVoucherTokens($voucherTokens);
+
+    /**
+     * Get cartHash - Cart Hash
+     *
+     * @return float|null
+     */
+    abstract public function getCartHash();
 }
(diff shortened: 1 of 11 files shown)


bundles/EcommerceFrameworkBundle/Model/AbstractOrder.php
```diff
@@ -724,62 +400,30 @@ public function hasDeliveryAddress()
...
-    }
-
-    /**
-     * return cart modification time stamp
-     *
-     * @throws UnsupportedException
-     *
-     * @return int
-     */
-    public function getCartModificationTimestamp()
```
fashxp, line 781: please add instead

```
/**
* Get cartHash - Cart Hash
* @return float|null
*/
public function getCartHash ();
```

fashxp (changes requested):
nice!

bundles/EcommerceFrameworkBundle/Model/AbstractOrder.php
blankse: I added it: https://github.com/pimcore/pimcore/pull/7553/commits/528daa63348fe3c960ba06f95b16579b2f7b97b5

fashxp (approved):

PR #7554, opened 2020-11-17
[Data Objects] Relation Data-Types: Validation error if object without an ID was assigned

### LINKED ISSUE #5953: [Data Objects] Relation Data-Types: Validation error if object without an ID was assigned
The following example should throw an Exception: 
```php
$newObject = new MyDataObject(); 
$existingObject->setMyRelations([$newObject]);
$existingObject->save(); 
```
At the moment a `0` value is stored for `dest_id` in the relation table.

weisswurstkanone:
resolves  #5953

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Relations/AllowAssetRelationTrait.php
+++ b/models/DataObject/ClassDefinition/Data/Relations/AllowAssetRelationTrait.php
@@ -16,6 +16,10 @@ trait AllowAssetRelationTrait
      */
     protected function allowAssetRelation($asset)
     {
+        if ($asset instanceof Asset && $asset->getId() <= 0) {
+            return false;
+        }
+
         $allowedAssetTypes = $this->getAssetTypes();
         $allowedTypes = [];
         $allowed = true;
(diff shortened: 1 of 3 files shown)


models/DataObject/ClassDefinition/Data/Relations/AllowAssetRelationTrait.php
```diff
@@ -16,6 +16,10 @@ trait AllowAssetRelationTrait
      */
     protected function allowAssetRelation($asset)
     {
+        if ($asset instanceof Asset && $asset->getId() <= 0) {
```
dvesh3, line 19: instanceof check really needed? as already done in checkValidity

weisswurstkanone: I would keep it for consistency as the same is not true for objects. Besides that, this is not an internal method.

dvesh3: then better would be `!$asset instanceof Asset || $asset->getId() <= 0`

weisswurstkanone: done

dvesh3 (approved):

PR #7564, opened 2020-11-18
[Application Logger] Support opening FileObject when using stream wrappers in PIMCORE_LOG_FILEOBJECT_DIRECTORY

BlackbitNeueMedien:
Currently stream wrappers are not supported when using a `Pimcore\Log\FileObject` together with stream wrappers in constant `PIMCORE_LOG_FILEOBJECT_DIRECTORY`. This PR solves that.


brusch:
Thanks a lot! 👌

PR #7566, opened 2020-11-19
Make Note->addData() chainable

danmichaelo:
Since the other methods on `Note` support chaining, it would be nice if `addData()` did too, so you could do stuff like this:

```php
(new Note())
    ->setElement($product)
    ->setType('notice')
    ->setTitle('Text')
    ->addData('originalPath', 'text', $oldPath)
    ->save();
```

Not completely sure if this should be considered a bugfix or an improvement though :) An improvement I think, but let me know if I should rebase it against 6.9.

### DIFF
--- a/models/Element/Note.php
+++ b/models/Element/Note.php
@@ -94,6 +94,8 @@ class Note extends Model\AbstractModel
      * @param string $name
      * @param string $type
      * @param mixed $data
+     *
+     * @return $this
      */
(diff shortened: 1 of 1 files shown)


dvesh3:
@danmichaelo please rebase it on 6.9. Thanks!

brusch:
`master` is fine as well, will be then part of v7 😊 Changed the milestone.

PR #7568, opened 2020-11-19
[Path formatter] Refresh view when using path formatter

BlackbitNeueMedien:
The solution in #5378 was only for (advanced) many to many object relations but not for advanced many to many relations. 
Additionally I think that the current solution in https://github.com/pimcore/pimcore/blob/f0019e7b7f776c9a0898b6c48d48f0e5267f9325/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyObjectRelation.js#L859-L863 cannot work because the variable `var nicePathRequested = pimcore.helpers.requestNicePathData(...)` can never become truthy, as there is no return path other than returning nothing in https://github.com/pimcore/pimcore/blob/9ffbad3967cf6a2492d2e4adf75b3383a27afa4f/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js#L2533-L2589
For this reason I added a `return true` there. Perhaps @weisswurstkanone can say if c115de8a6a0e1c8f2cb617222ebed41ec4f5cdfc was missing this when checking the return value of `pimcore.helpers.requestNicePathData()` or if it was on purpose. And if the changes in this PR make sense.


weisswurstkanone:
@BlackbitNeueMedien, nice, maybe put this into 6.8? What do you think?

BlackbitNeueMedien:
Done

weisswurstkanone:
Could you please have a look at traivs please? Thx!

BlackbitNeueMedien:
Are the failing tests really my fault? The tests are actually not being executed at all. There is only the message `The command "if [ $TASK_TESTS == 1 ]; then .travis/setup-tests.sh; fi" failed and exited with 1 during .`

BlackbitNeueMedien:
Ah, now tests are green ;)

weisswurstkanone:
Merci!

PR #7572, opened 2020-11-20
Remove leftover QR-Code parts

jdreesen:
QR-Codes were removed in #7262 but there was still some code left.

The Document preview panel has a "QR-Codes" button that's now broken because the route doesn't exist anymore, as its controller has been removed.

Note: there's a translation with the key `qr_codes` left, that doesn't seem to be used anywhere. Not sure if it should be deleted, too (and if so, in which translation files - only English or all?)

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
@@ -1551,39 +1551,6 @@ class DocumentController extends ElementControllerBase implements KernelControll
     }
 
     /**
-     * @Route("/qr-code", name="pimcore_admin_document_document_qrcode", methods={"GET"})
-     *
-     * @param Request $request
-     *
-     * @return BinaryFileResponse
-     */
-    public function qrCodeAction(Request $request)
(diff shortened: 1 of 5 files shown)


brusch:
Thanks a lot for your PR! 
I'd keep this functionalitity, so this was removed by mistake. Maybe we move the action just to the document controller, I guess this would make sense, wdyt? 😊

jdreesen:
I'll update this PR accordingly.

jdreesen:
I restored the deleted controller action into the `Pimcore/Bundle/AdminBundle/Controller/Admin/Document/DocumentController` and removed the unnecessary parts.

As the preview button is currently only in `bundles/AdminBundle/Resources/public/js/pimcore/document/pages/preview.js`, should I move it into the `Bundle/AdminBundle/Controller/Admin/PageController` instead?

brusch:
Yes, then PageController would be nice 😊 Thanks!

jdreesen:
I moved it to the `PageController`.

brusch:
Thanks a lot!

PR #7578, opened 2020-11-20
[Admin] Fix open action for folder relations

### LINKED ISSUE #7571: Can't open object-folder in manyToOne-relation if folder is assigned with script
### Expected behavior
Assign folder via php to a objects manyToOne-relation field and then open it by clicking on the relation or the "open folder"-icon.
### Actual behavior
Assigning works without problems, relations are ok. If you try to open it, it is trying to open the folder via /get instead of /get-folder and will throw this error:

**Argument 1 passed to Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectController::getLatestVersion() must be an instance of Pimcore\Model\DataObject\Concrete, instance of Pimcore\Model\DataObject\Folder given, called in /home/protina/www/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php on line 366**

If you then assign the folder again per hand in the Pimcore-Backend, opening it is no problem anymore.
Asset-Folders are also no problem. I don't about Document-Folders :)

### Steps to reproduce
Create an object with a manyToOne-relation (doesn't matter if restricted only to folders) and then assign a object-folder via a php-script.

### Code example
```php
$testFolder = Object\Folder::getById(21);
$testObj->setTestFolder($testFolder);
$testObj->save();
```

dvesh3:
Resolves #7571

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
@@ -305,7 +305,7 @@ pimcore.helpers.recordElement = function (id, type, name) {
 };
 
 pimcore.helpers.openElement = function (idOrPath, type, subtype) {
-    if (typeof subtype != "undefined") {
+    if (typeof subtype != "undefined" && subtype !== null) {
         if (type == "document") {
             pimcore.helpers.openDocument(idOrPath, subtype);
         }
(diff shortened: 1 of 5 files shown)


weisswurstkanone:
thx!

PR #7579, opened 2020-11-23
Model::getById() ::getByPath() ... shouldn't catch any exceptions

### LINKED ISSUE #7322: Model::getById() ::getByPath() ::getByName() ...  shouldn't catch any Exceptions

weisswurstkanone:
resolves #7322

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -29,7 +29,7 @@ use Pimcore\Model\Asset\Listing;
 use Pimcore\Model\Asset\MetaData\ClassDefinition\Data\Data;
 use Pimcore\Model\Asset\MetaData\ClassDefinition\Data\DataDefinitionInterface;
 use Pimcore\Model\Element\ElementInterface;
-use Pimcore\Model\Element\Exception\ElementNotFoundException;
+use Pimcore\Model\Exception\NotFoundException;
 use Pimcore\Tool;
 use Pimcore\Tool\Mime;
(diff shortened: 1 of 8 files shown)


models/Element/Exception/ElementNotFoundException.php
```diff
@@ -0,0 +1,22 @@
...
+ * @category   Pimcore
+ * @package    Element
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Model\Element\Exception;
+
+class ElementNotFoundException extends \Exception
```
jdreesen, line 20: This should extend the [`RuntimeException`](https://www.php.net/manual/de/class.runtimeexception.php) class, as this only occurs on runtime.

```suggestion
class ElementNotFoundException extends \RuntimeException
```

models/Asset.php
```diff
@@ -232,7 +232,7 @@ public static function getByPath($path, $force = false)
             $asset->getDao()->getByPath($path);
 
             return static::getById($asset->getId(), $force);
-        } catch (ElementNotFoundException $e) {
+        } catch (NotFoundExceptio $e) {
```
jdreesen:
```suggestion
        } catch (NotFoundException $e) {
```

jdreesen (changes requested):

PR #7580, opened 2020-11-23
OwnerAwareFieldInterface: add setters and getters

### LINKED ISSUE #6635: OwnerAwareFieldInterface: add setters and getters
follow up #6604

weisswurstkanone:
resolves #6635

### DIFF
--- a/models/DataObject/OwnerAwareFieldInterface.php
+++ b/models/DataObject/OwnerAwareFieldInterface.php
@@ -24,19 +24,19 @@ interface OwnerAwareFieldInterface
      *
      * @return $this;
      */
-    public function __setOwner($owner);
+    public function _setOwner($owner);
 
     /**
      * @param string|null $fieldname
      * @return $this
      */
-    public function __setOwnerFieldname(?string $fieldname);
+    public function _setOwnerFieldname(?string $fieldname);
 
     /**
      * @param string|null $language
      * @return $this
      */
-    public function __setOwnerLanguage(?string $language);
+    public function _setOwnerLanguage(?string $language);
 
     /**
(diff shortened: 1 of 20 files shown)


models/DataObject/OwnerAwareFieldInterface.php
```diff
@@ -21,8 +21,35 @@ interface OwnerAwareFieldInterface
 {
     /**
      * @param mixed $owner
-     * @param string $fieldname
+     *
+     * @return $this;
+     */
+    public function __setOwner($owner);
```
jdreesen, line 27: Probably the "double underscore" should mark the methods as "private", but it would be better to use another prefix for this (e.g. only one underscore) because functions/methods that start with `__` are considered as "magic" by PHP and should not be defined in userland as stated here: https://www.php.net/manual/en/language.oop5.magic.php

weisswurstkanone: The double underscore is mean to avoid name clashes with existing methods such as "setFieldname". Of course we can change this to something else. Any preferences?

weisswurstkanone: The original inspiration was this method I guess which is not magic as well:  "__setDataVersionTimestamp"
I think it would be sufficient to use a single underscore which maybe would be even more consistent as the internally it is stored in _owner, _fieldname and _language.

jdreesen: I'd vote for a single underscore, too.

weisswurstkanone: done :-)

Issue #7184, opened 2020-09-28
Shared translations broken in 6.7.2


Cruiser13:
# Bug Report
### Expected behavior
Translations work as they used to. This will be properly translated if placed in a twig view template:
`{{ "Role"|trans }}`
Works in 6.7.0.

### Actual behavior
Starting from 6.7.2, this will no longer use the proper translated value but the english value only.

### Reason
Must be in this commit:
https://github.com/pimcore/pimcore/pull/7071/commits/3f4e0e3b81f330ccb287716d3a5c46318e08d3b2#diff-35148f8556f54cb705805f95b0f19fed
Reverting it fixes the bug.

brusch:
Unable to reproduce ... just tested this on the public demo:

![image](https://user-images.githubusercontent.com/142037/94441071-991bd580-01a2-11eb-9120-2b20e6e8ccc7.png)
![image](https://user-images.githubusercontent.com/142037/94441107-a33dd400-01a2-11eb-865c-d746ba99b09b.png)
![image](https://user-images.githubusercontent.com/142037/94441127-ab960f00-01a2-11eb-8a30-45c8ab16d27b.png)

Cruiser13:
Weird, we can reproduce it. Might be settings specific? Here are ours:
![settings locale](https://user-images.githubusercontent.com/2771909/94441846-89e95780-01a3-11eb-92b3-1d848afb2fbf.png)

brusch:
Just tried the same (added allback language and set default to `de`) ... works as expected ... 😐

dpfaffenbauer:
try clearing the cache, maybe that works.

brusch:
Did it twice ... still works

Cruiser13:
Clearing the cache was the first thing we did too... no success.

We nailed it down to this line of code. Commenting out 
`$lookForFallback = $comparisonId == $translated;`
makes it work for us. No idea why yet.

Cruiser13:
We have a second pimcore instance now which shows the same behaviour at 6.7.2.
I think the issue appears, if the key and one of the translations are the same. In the following setup:

![key replacement shared translations](https://user-images.githubusercontent.com/2771909/94533384-5825ce80-023f-11eb-86c7-e567c3227097.png)

This will be translated properly:
`{{ "ask_for_callback"|trans }}`

This will fail:
`{{ "Rückruf vereinbaren"|trans }}`

brusch:
[...]

PR #7584, opened 2020-11-23
Consider parameters in fallback translations

jremmurd:
# Bugfix

Steps to reproduce the bug:

1. Setup two languages, one the fallback of the other (i.e. de and de_at, de is fallback for de_AT)
2. Now create a translation for de_AT with a placeholder i.e. cart.incl-vat --> "Taxrate is %percent%%", where %percent% is the placeholder which should be replaced.
3. The output will be "Taxrate is %percent%%" and thus the parameters won't be replaced.
4. On second call the translation will be right, the output is i.e. "Taxrate is 20%"

With this PR this behaviour gets fixed and the parameters are already replaced at the very first call.

### DIFF
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -357,7 +357,7 @@ class Translator implements LegacyTranslatorInterface, TranslatorInterface, Tran
                     // update fallback value in original catalogue otherwise multiple calls to the same id will not work
                     $this->getCatalogue($locale)->set($normalizedId, $fallbackValue, $domain);
 
-                    return $this->translator->trans($fallbackValue, $parameters);
+                    return strtr($fallbackValue, $parameters);
                 }
             }
         }
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -357,7 +357,7 @@ class Translator implements LegacyTranslatorInterface, TranslatorInterface, Tran
                     // update fallback value in original catalogue otherwise multiple calls to the same id will not work
                     $this->getCatalogue($locale)->set($normalizedId, $fallbackValue, $domain);
 
-                    return $fallbackValue;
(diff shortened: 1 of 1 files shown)


dvesh3:
I guess this is also broken for default translations(https://github.com/pimcore/pimcore/issues/7184#issuecomment-726340626), so should be fixed in general not just for fallback.

jremmurd:
> I guess this is also broken for default translations([#7184 (comment)](https://github.com/pimcore/pimcore/issues/7184#issuecomment-726340626)), so should be fixed in general not just for fallback.

In 6.6.2 all works fine, but translations with placeholders using fallback language as far as I can see.

brusch:
Closed in favor of #7588 that fixes multiple issues. Thanks anyway @jremmurd !

jremmurd:
> Closed in favor of #7588 that fixes multiple issues. Thanks anyway @jremmurd !

This does  not fix this issue though, have you tried the above steps to reproduce the bug?

brusch:
Ah sorry, I mixed #7184 with #7584 - just looked at the last 2 digits 😊Sorry again.

lib/Translation/Translator.php
```diff
@@ -357,7 +357,7 @@ protected function checkForEmptyTranslation($id, $translated, $parameters, $doma
                     // update fallback value in original catalogue otherwise multiple calls to the same id will not work
                     $this->getCatalogue($locale)->set($normalizedId, $fallbackValue, $domain);
 
-                    return $fallbackValue;
+                    return $this->translator->trans($fallbackValue, $parameters);
```
dvesh3, line 360:
```suggestion
                    return strtr($fallbackValue, $parameters);
```

brusch:
Thanks a lot guys!

PR #7587, opened 2020-11-25
Installation of Pimcore

aidatulaqilah:
Resolves #


dvesh3:
@aidatulaqilah why renamed doc file? number prefix is used to sort docs.

brusch:
Don't think this makes sense 🙂

aidatulaqilah:
> @aidatulaqilah why renamed doc file? number prefix is used to sort docs.

I am really sorry, Idk how to use this actually

PR #7583, opened 2020-11-23
[Data Objects] Data-Types: remove getPhpdocType() BC layer

weisswurstkanone:
resolves #7310

PR #7589, opened 2020-11-25
Remove abstract methods from `Pimcore\Model\DataObject\ClassDefinition\Data` that are also in `Pimcore\Model\DataObject\ClassDefinition\Data\TypeDeclarationSupportInterface`

jdreesen:
This is a PR for the PR #7583

Since the methods of `Pimcore\Model\DataObject\ClassDefinition\Data\TypeDeclarationSupportInterface` are only called in `Pimcore\Model\DataObject\ClassDefinition\Data`, they should not be defined as `abstract` methods in this class as well. This would make the interface useless as the existence of these methods would be forced by the class and therefore they are no longer optional.

### DIFF
--- a/models/DataObject/ClassDefinition/Data.php
+++ b/models/DataObject/ClassDefinition/Data.php
@@ -610,31 +610,6 @@ abstract class Data
     }
 
     /**
-     * @abstract
-     *
-     * @return string|null
-     */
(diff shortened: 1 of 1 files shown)


brusch:
Thanks a lot! 👌

PR #7593, opened 2020-11-25
Hobex Payment Provider: if transaction was already processed, query transaction

rezoti:
if checkout-resource (v1/checkouts/xxxx/payment) is called twice after
transaction is completed (e.g. if user reloads page before transaction is
completed), hobex will throw an error.
To avoid this, check if transaction was already processed (=> hobex-Brick
exists and status != pending); if this is the case, transaction can
be queried from /v1/query/{transactionid}

  
  

Resolves #


brusch:
Obsolute due to #7596

PR #7594, opened 2020-11-26
[Grid] Auto-adjust column width depending on currently shown content

BlackbitNeueMedien:
Currently in the grid view the columns have fixed width which can get overridden by setting the width in the field definitions. But this fixed width is not optimal in most situations and leads to a lot of the columns being cropped - in this simple case ID and path are truncated:
Bildschirmfoto 2020-11-26 um 08 01 29

This PR changes this behaviour by automatically adjusting the column width depending on the contents which are currently shown. Of course the widths can still be overridden in the field definitions. Without any other changes the grid from above screenshot now looks like this:
Bildschirmfoto 2020-11-26 um 08 05 41

Of course this might lead to very big columns for text fields but imho it is still a better default setting than to truncate nearly all columns and nevertheless waste much space to the right.


brusch:
Makes absolutely sense, thanks a lot!

PR #7598, opened 2020-11-26
[Data objects Grid] Fix search

ramundomario:
The problem appears when apply search in data objects grid, where objects that don't match the search could be found
if there are documents or assets with the same id which match the search

## Steps to reproduce  
The following steps were performed on the current demo.
1. Open Product Data/Cars folder & search "fiat"
Screenshot from 2020-11-26 22-27-12

2. The selected items don't match the search but there are documents/assets with the same id which match the search
Screenshot from 2020-11-26 22-33-46
Screenshot from 2020-11-26 22-33-24


brusch:
Thanks a lot! 👌

PR #7487, opened 2020-11-06
Fix phpstan level 3 errors

blankse:
Partly #6780

Before:
`[ERROR] Found 407 errors`

After:
`[ERROR] Found 324 errors`

PR #7615, opened 2020-11-27
[Navigation] Improve active trail fallback code

jdreesen:
This code was improved by #7487 a bit, but we can do a little better.

Actually, if `$activeDocument` isn't an instance of `Document`, we don't even need to try building the trail, because `$activeTrail` would always be false anyways.


brusch:
Thanks a lot!

PR #7610, opened 2020-11-27
Tags filter not works if tags are uppercase

Se7en-IT:
When tags are uppercase, when i search on tags filter section, i don't find nothing.

PR #7617, opened 2020-11-27
Fixup: tags filter not works if tags are uppercase

jdreesen:
This is a fixup for #7610 

Like I said in that branch, I would not rely on the input and call `strtolower` nonetheless.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/TagsController.php
+++ b/bundles/AdminBundle/Controller/Admin/TagsController.php
@@ -109,13 +109,13 @@ class TagsController extends AdminController
     public function treeGetChildrenByIdAction(Request $request)
     {
         $showSelection = $request->get('showSelection') == 'true';
-        $assginmentCId = intval($request->get('assignmentCId'));
-        $assginmentCType = strip_tags($request->get('assignmentCType'));
+        $assignmentCId = intval($request->get('assignmentCId'));
+        $assignmentCType = strip_tags($request->get('assignmentCType'));
 
         $recursiveChildren = false;
         $assignedTagIds = [];
-        if ($assginmentCId && $assginmentCType) {
-            $assignedTags = Tag::getTagsForElement($assginmentCType, $assginmentCId);
+        if ($assignmentCId && $assignmentCType) {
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/TagsController.php
```diff
@@ -133,7 +133,7 @@ public function treeGetChildrenByIdAction(Request $request)
         if (!empty($request->get('filter'))) {
             $filterIds = [0];
             $filterTagList = new Tag\Listing();
-            $filterTagList->setCondition('`name` LIKE '. $filterTagList->quote('%'. $request->get('filter') .'%'));
+            $filterTagList->setCondition('`name` LIKE '. $filterTagList->quote('%'. strtolower($request->get('filter')) .'%'));
```
NiklasBr, line 136:
```suggestion
            $filterTagList->setCondition('`name` LIKE '. $filterTagList->quote('%'. mb_strtolower($request->get('filter')) .'%'));
```

But shouldn't `LIKE` be case-insensitive anyway when Pimcore tables are `utf8mb4_*_ci`?
https://mariadb.com/kb/en/like/
https://dev.mysql.com/doc/refman/5.7/en/case-sensitivity.html

brusch: @NiklasBr https://github.com/pimcore/pimcore/blob/9a1bde84311eb9aa02fbd00b1d7c13a67ce6e993/bundles/InstallBundle/Resources/install.sql#L494

brusch: The column `name` is `utf8_bin` because of the uniqueness of a name, but of course for the search it makes sense to be case-insensitive.

brusch:
@jdreesen can you please rebase this PR onto the `6.8` branch, since #7610 was also based on `6.8` - Thanks a lot!

jdreesen:
Rebased it onto `6.8` and added two small (unrelated) fixes.

brusch:
Thanks a lot! 🙏

PR #7619, opened 2020-11-28
Fix: Change the published state, regardless of the previous settings.

patrick-schmid:
If you open a document, switch to the "settings" tab and then try to publish or unpublish the document, the new status is not applied.

### Reason:
If you switch to the tab "settings", all settings - including the current published/unpublished status - are transferred to the controller, which overwrites the new desired status.

### Solution:
Set the new status after all other settings are set.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/PageController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PageController.php
@@ -148,13 +148,6 @@ class PageController extends DocumentControllerBase
 
         $page->setUserModification($this->getAdminUser()->getId());
 
-        if ($request->get('task') == 'unpublish') {
-            $page->setPublished(false);
-        }
-        if ($request->get('task') == 'publish') {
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/Document/PageController.php
```diff
@@ -182,6 +175,13 @@ public function saveAction(Request $request)
         if (($request->get('task') == 'publish' && $page->isAllowed('publish')) || ($request->get('task') == 'unpublish' && $page->isAllowed('unpublish'))) {
             $this->setValuesToDocument($request, $page);
 
+            if ($request->get('task') == 'unpublish') {
+                $page->setPublished(false);
+            }
+            if ($request->get('task') == 'publish') {
```
dvesh3, line 181:
```suggestion
            } elseif ($request->get('task') == 'publish') {
```

PR #7623, opened 2020-11-30
Optimize Dependency::isRequired()

blankse:
getRequiredBy returns always an array and generate sql queries.


brusch:
Thanks a lot!

PR #7624, opened 2020-11-30
PHPDoc: Missing returns

blankse:


models/Asset/MetaData/ClassDefinition/Data/Date.php
```diff
@@ -37,6 +37,8 @@ public function marshal($value, $params = [])
     /**
      * @param mixed $value
      * @param array $params
+     *
+     * @return mixed
```
weisswurstkanone, line 41: IMHO, this has always to return a string or null, otherwise the implementation is wrong.
version preview is primarily only there to create a human readable overview (plain or html) of the data. wdyt?

blankse: @weisswurstkanone Yes, but than we have to declare the value in Data::getVersionPreview also to string or we make an `if (is_string())`. Parent/Child methods must have the same types

blankse: @weisswurstkanone I cast it to string. So okay?

weisswurstkanone: should be perfectly fine

brusch:
Thanks a lot! 👍

PR #7632, opened 2020-12-01
[Doc] Remove $info->getView()

blankse:


brusch:
Nice, thanks a lot! 👍

PR #7640, opened 2020-12-01
Areabrick names are not translated

blankse:
After upgrade from 6.7.* to 6.8.* all areabrick names are not translated anymore.

We use translation keys as names. so this is a blocker :(


brusch:
Thanks a lot!

PR #7646, opened 2020-12-02
[Thumbnails] Throw exception if requested thumbnail definition does not exist

BlackbitNeueMedien:
Currently you do not get an error when doing
```php
$asset = Image::getById(1234);
$thumbnail = $asset->getThumbnail('non_existing_thumbnail_definition');

echo $thumbnail->getPath();
```
But you get the original image in this case. Moreover you do not even get something logged which could hint you that the thumbnail definition does not exist.
This is especially bad if the thumbnail definition once existed and everything worked and somebody wanted to "clean up".

This PR changes this behaviour to throw an excpetion if a non-existing thumbnail definition gets requested.


brusch:
I'd keep the static getter as is (`return nulll`), to be consistent with the other models.
This change would also require some more changes where `getByName()` is used, e.g. https://github.com/pimcore/pimcore/blob/4fd1f8d1889b0c5e5f315b6aa1a6dad55fb2e362/bundles/AdminBundle/Controller/Admin/SettingsController.php#L1275

Maybe we should throw the exception here instead: 
https://github.com/pimcore/pimcore/blob/16fde8bbab89f07801f9a66e1ef39b103d487b02/models/Asset/Image/Thumbnail.php#L123
https://github.com/pimcore/pimcore/blob/961a9d033bc5004ff06bbe56b523da34653f2212/models/Asset/Document/ImageThumbnail.php#L111
https://github.com/pimcore/pimcore/blob/df7f57d291293a64f81a1f7b5a95f7e07332f9fa/models/Asset/Video/ImageThumbnail.php#L142

brusch:
If we change the behavior, we should also trigger an appropriate deprecation message in `6.9` if no valid config is given.

brusch:
@BlackbitNeueMedien 🏓😊

BlackbitNeueMedien:
Oh, missed that. Try to continue working on this during the holidays (I hope)...

brusch:
@BlackbitNeueMedien Thanks a lot! 😊

brusch:
@BlackbitNeueMedien friendly reminder 😉

BlackbitNeueMedien:
@brusch Sorry for the delay.

In https://github.com/pimcore/pimcore/blob/16fde8bbab89f07801f9a66e1ef39b103d487b02/models/Asset/Image/Thumbnail.php#L121-L124
I cannot check if a non-existing thumbnail definition has been requested because if `$this->config` is `null` there this could be due to having called `$imageAsset->getThumbnail()` or `$imageAsset->getThumbnail('non-existing-thumbnail-def)`. 

But I found another solution. What do you think?

brusch:
@BlackbitNeueMedien thanks for your update. 
The question is whether we should allow `$image->getThumbnail()` in the future or not. 
In my opinion we should not allow that anymore, because one should use `$image->getFullPath()` instead and calling `$image->getThumbnail()` should always require a valid config. 

WDYT?

models/Asset/Image/Thumbnail.php
```diff
@@ -396,6 +396,12 @@ public function getMedia($name, $highRes = 1)
      */
     protected function createConfig($selector)
     {
-        return Thumbnail\Config::getByAutoDetect($selector);
+        $thumbnailConfig = Thumbnail\Config::getByAutoDetect($selector);
+
+        if(!empty($selector) && $thumbnailConfig === null) {
+            throw new \Exception('Thumbnail definition "' . (is_string($selector)? $selector:'') . '" does not exist');
```
blankse, line 402: I would prefer `Pimcore\Model\Exception\NotFoundException` and add it to the PHPDoc (also the other exceptions)

brusch: Agree.

BlackbitNeueMedien: Fixed. Thanks @blankse

BlackbitNeueMedien:
@brusch I am not a big fan of BC breaks only to enforce to use another method without any behavioral change. But if you want, I can remove those checks for `!empty($selector)` like in 
https://github.com/pimcore/pimcore/blob/49bb23a5fe1ba5236b5dbf0eb99bdbf020933afd/models/Asset/Image/Thumbnail.php#L404

models/Asset/Video.php
```diff
@@ -129,6 +129,10 @@ public function getThumbnailConfig($config)
 
         if (is_string($config)) {
             $thumbnail = Video\Thumbnail\Config::getByName($config);
+
+            if($thumbnail === null) {
+                throw new Model\Exception\NotFoundException('Video Thumbnail definition "' . $config . '" does not exist');
```
blankse, line 134: Missing PHPdoc (also for models/Asset/Video/ImageThumbnail.php )

BlackbitNeueMedien: Fixed. Thank you.

brusch:
@BlackbitNeueMedien hmm, I can follow your argumentation, in the end it doesn't hurt at all if `getThumbnail()` returns the path to the original image if no config is given. So let's keep it as it is. Thanks!

brusch:
@BlackbitNeueMedien can you please resolve the merge conflict?

brusch:
@BlackbitNeueMedien one more thing: the check in `\Pimcore\Model\Asset\Document\ImageThumbnail::createConfig()` is still missing. Thanks 😊

BlackbitNeueMedien:
@brusch I resolved conflict and added check in `\Pimcore\Model\Asset\Document\ImageThumbnail::createConfig()`.

brusch:
Thanks a lot!

PR #7653, opened 2020-12-02
Fixed wrong annotation

brusch:


models/DataObject/ClassDefinition/Data.php
```diff
@@ -1700,7 +1699,7 @@ public function isEqual($oldValue, $newValue)
     }
 
     /**
-     * @param LazyLoadingInterface $object
+     * @param DataObject\ClassDefinition\Data\LazyLoadingSupportInterface $object
```
blankse, line 1702: I think it should be DataObject\Concrete.
This interface is for field types. But here concrete object are passed.

PR #7666, opened 2020-12-03
Simplify regex for static files in nginx config

jdreesen:
This is a followup to #4372

In the regex for static file serving in the nginx config the `(?!/admin/asset/webdav/)(?!/admin)` part is redundant, because when the path may not start with `/admin`, it may not start with `/admin/asset/webdav/` either, no need to specify this explicitly again.


brusch:
Thanks!

PR #7670, opened 2020-12-04
[Models] Implement JsonSerializable interface

BlackbitDevs:
For exports it would be nice to be able to serialize Pimcore elements to JSON:
```php
$product = Product::getById(123);
echo json_encode($product);
```

For this to work this PR implements PHP's JsonSerialize interface.

Current problem and reason why this is only a draft:
* currently not always works because of `Recursion detected` -> this could be solved with things like https://stackoverflow.com/questions/44323296/recursion-detected-in-json-encode - but this seems ugly and also slow (because of serializing + unserializing the object to remove recursion). Perhaps someone has another idea?

Is this actually interesting for anyone so more research would be worth it?

### DIFF
--- a/lib/Model/AbstractModel.php
+++ b/lib/Model/AbstractModel.php
@@ -14,6 +14,7 @@
 
 namespace Pimcore\Model;
 
+use JsonSerializable;
 use Pimcore\Logger;
 use Pimcore\Model\DataObject\Traits\ObjectVarTrait;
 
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
The perfect solution would be to actually serialize it, like something with symfony/serializer. That means also that data-types either have to support it, or we have to write serializers for complex types (fieldcollections, bricks, quantity-value, etc.)

brusch:

PR #7674, opened 2020-12-04
Fix wrong QuantityValue PHPDocs

blankse:
- getUnitId(), getUnit(), getValue() can return null
- Cunstructor should use string instead of int as unit


brusch:
@blankse We did some major changes to `OwnerAwareFieldInterface` for v10 recently  (see #7580), which also affects this PR. So I'd  base this PR directly on `master` and consider the changes of #7580. Ok for you? Thanks a lot!

blankse:
@brusch Okay, I rebased to master

brusch:
@blankse thanks a lot!

weisswurstkanone:
thx!

PR #7692, opened 2020-12-08
Fix docs for "Custom Icons & Tooltips for Documents/Assets & Data Objects"

jdreesen:
I removed the link to the demo site of the "Custom Icon for the Car DataObject" example for now, because it's broken and redirects to the homepage of [elements.at](https://www.elements.at/de) and I'm not sure where to link to instead.

Also, I moved setting the `elementQtipConfig` out of the condition in the "Custom Style for Assets" example, so that it matches the description.


brusch:
Thanks a lot!. 
Regarding the dead link: I guess it's fine without a link, maybe we can include a screenshot instead in the future.

PR #7541, opened 2020-11-16
composer.json conflict with "phpstan/phpstan": "^0.12.55"

brusch:
See https://github.com/phpstan/phpstan/issues/4067

PR #7693, opened 2020-12-08
Conflict only with PHPStan 0.12.55

jdreesen:
This is a followup of #7541

Apparently the bug in PHPStan has been fixed in `0.12.56`, so only the broken version `0.12.55` should be listed in composers `conflict` section.


weisswurstkanone:
thx!

PR #7694, opened 2020-12-08
Fix Ghostscript::getVersion()

### LINKED ISSUE #7683: Ghostscript getVersion() returns wrong format
# Bug Report

### Expected behavior
$this->getVersion() should return string without newline
e.g. "9.50"
### Actual behavior
$this->getVersion() returns:
e.g. "9.50
"
--> if clause is always **false**

### Steps to reproduce
add **var_dump($this->version);** in Ghostscript Adapter before line 192
(We use Ubuntu 20.04.1 LTS and Ghostscript 9.50)
### Code example
```php
protected function getVersion()
{
    if (is_null($this->version)) {
        $this->version = Console::exec(self::getGhostscriptCli() . ' --version');
    }
    var_dump($this->version);
    return $this->version;
}
```

### Possible Solution
add **trim()** function to **return $this->version**
```php
protected function getVersion()
{
    if (is_null($this->version)) {
        $this->version = Console::exec(self::getGhostscriptCli() . ' --version');
    }
    return trim($this->version);
}
```
This could be a solution for #7665

jdreesen:
Removes possible line breaks around Ghostscript's console output of `--version`.

Fixes #7683

### DIFF
--- a/lib/Document/Adapter/Ghostscript.php
+++ b/lib/Document/Adapter/Ghostscript.php
@@ -185,7 +185,7 @@ class Ghostscript extends Adapter
     protected function getVersion()
     {
         if (is_null($this->version)) {
-            $this->version = Console::exec(self::getGhostscriptCli() . ' --version');
+            $this->version = trim(Console::exec(self::getGhostscriptCli() . ' --version'));
         }
 
         return $this->version;


dvesh3:
thanks!

Issue #5628, opened 2020-01-22
AWS S3 integration issue in download assets zip file


kudeepak:
# Bug Report

### Expected behavior
Image should be douwnload by clicking button "Download selected items as ZIP" under assets listing. 
### Actual behavior
It is giving 500 internal server error.
### Steps to reproduce
-> Go in assets
-> Click on list
-> Select checkbox 
-> Click "Download selected items as ZIP"

### Code example
```php
// your code
```

# Feature Request

brusch:
Please provide the full error message incl. the stack trace, otherwise we're not able to help. You should find this information in your log files.

kudeepak:
Oops! An Error Occurred
The server returned a "500 Internal Server Error".
Something is broken. Please let us know what you were doing when this error occurred. We will fix it as soon as possible. Sorry for any inconvenience caused.


The file "/var/www/fssqa.joyousminds.com/pmifss01/var/tmp/download-zip-5e27e6b4688e2.zip" does not exist


in vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/File/File.php (line 36)

kudeepak:
It is a path issue. But we have this path and that one is writable as well. Then why this error is coming. 
We are using Pimcore 6. Also we followed all the steps of Pimcore 6 to integrate the S3.

brusch:
You don't have a stack trace? 
Would be nice if you could fix that and provide us a PR as we don't have a dev-system with S3 ready. 
Should be quite easy, just have a look at this action: 
https://github.com/pimcore/pimcore/blob/ca75246d82fe1190ce2268767ace84f157b168d9/bundles/AdminBundle/Controller/Admin/AssetController.php#L1766

Btw. `PIMCORE_SYSTEM_TEMP_DIRECTORY` shouldn't be on S3 ... if that's the case, change it to a local path.

kudeepak:
Thanks Brusch, Will look as per your suggestion. 
As you hint will check in that direction.

I will update in this ticket only.

kudeepak:
Hi Brusch,
Just after updating the Line no 1880 of file   
`pimcore/bundles/AdminBundle/Controller/Admin/AssetController.php

`$zip->addFile($a->getFileSystemPath(), preg_replace('@^' . preg_quote($asset->getRealPath(), '@') . '@i', '', $a->getRealFullPath()));`
with 
[...]

PR #7695, opened 2020-12-08
[Assets] Fix "allowed memory size exhausted" error

aweichler:
Fix an error, which occurs when using "Download selected items as ZIP" function at a folder view in assets.
It occurs when selecting large files. In our case there where 3 Videos ~200M. PHP Memory Limit was at 256M or 512M.
The function "file_get_contents" reads the whole file into memory before inserting it into the ZIP archive. With "addFile" of the class "ZipArchive" this limitation is bypassed.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -2041,7 +2041,11 @@ class AssetController extends ElementControllerBase implements KernelControllerE
                     if ($a->isAllowed('view')) {
                         if (!$a instanceof Asset\Folder) {
                             // add the file with the relative path to the parent directory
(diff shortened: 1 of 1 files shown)


brusch:
We used that already in the past, but this caused problems when using remote streams, see #5628
I'd suggest to check whether `$a->getFileSystemPath()` is a local or remote stream (`stream_is_local()`) and then use `addFromString()` or `addFile()` accordingly. 

Can you update your PR? 😊 Thanks a lot!

aweichler:
Would it also be possible to apply the changes in Pimcore 6.x? Our customer uses this feature and would have to wait very long for Pimcore 10.

brusch:
@aweichler yes, `6.9` would be possible, can you please rebase your PR onto this branch and change base accordingly? Thanks!

aweichler:
I am not sure how to do this. Would it be ok if i provide a new pull request for branch 6.9?

brusch:
@aweichler yes, of course, just create a new PR 😉

PR #7696, opened 2020-12-08
Fix race condition in workflow transition saves

davidmaes:
This fixes a problem where a save is overwritten with an older save when you save an object in a event triggered by a transition. After saving the object, another save is triggered in the WorkflowManager that puts the object in an older state again undoing everything we did in the transistion event itself. The save itself seems not needed since both notes and transitions are being updated correctly even without this save

  
  

Resolves #

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/WorkflowController.php
+++ b/bundles/AdminBundle/Controller/Admin/WorkflowController.php
@@ -110,7 +110,7 @@ class WorkflowController extends AdminController implements KernelControllerEven
 
         if ($workflow->can($this->element, $request->get('transition'))) {
             try {
-                $workflowManager->applyWithAdditionalData($workflow, $this->element, $request->get('transition'), $workflowOptions, true);
+                $workflowManager->applyWithAdditionalData($workflow, $this->element, $request->get('transition'), $workflowOptions);
 
                 $data = [
                     'success' => true,
--- a/lib/Workflow/Manager.php
+++ b/lib/Workflow/Manager.php
@@ -241,25 +241,13 @@ class Manager
      * @throws ValidationException
      * @throws \Exception
      */
-    public function applyWithAdditionalData(Workflow $workflow, $subject, string $transition, array $additionalData, $saveSubject = false)
+    public function applyWithAdditionalData(Workflow $workflow, $subject, string $transition, array $additionalData)
     {
         $this->notesSubscriber->setAdditionalData($additionalData);
 
         $marking = $workflow->apply($subject, $transition);
 
         $this->notesSubscriber->setAdditionalData([]);
-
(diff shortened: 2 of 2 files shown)


brusch:
@davidmaes Thanks a lot for your PR. Can you please rebase your PR onto the `6.8` branch, since it's a bug fix? 
Thanks a lot! 😊

dpfaffenbauer:
@brusch that is a BC break! there might be cases where this is intended and works!

brusch:
Thanks for your feedback ( btw. no exclamation marks needed when writing to me 😉 ). Let's keep this PR open as it is and discuss how to solve this  in the future.

davidmaes:
Ok, so what to do? Because this might be BC break, but it is definitely not OK (or needed) to save here. We could make sure the $subject is "refreshed" before the save, so it doesn't break. I mean, at this moment, you can't save anything in a transition event because it will be overwritten. That seems like a bigger problem than BC break for niche cases.

dpfaffenbauer:
@davidmaes The niche case depends on how you look at it, for me it is a niche case to change it during a transition ;)

davidmaes:
@dpfaffenbauer I kindly disagree. What do you even have the events for then? What else am I supposed to do in events except alter the object at hand. That is the main case for having events at all, and definitely not niche. I say your case is niche because 1) this logic is only called through the frontend. If you want to transition programatically, you can use the registry. You wouldn't even go through the code that I changed. 

Now, please suggest a better way, or I would suggest just merging this branch, because as I see it, there isn't even any BC break.

davidmaes:
I still disagree. This won't break anything. The save is unneeded, and it only accessed through the frontend. That is definitely not bc breaking. Only for people that actively hacked into the frontend and did stuff there.

Now, I still need a solution because with your logic, you wouldn't even be able to change anything to an object in an event. Which is a major bug. Can we at least rehydrate the object before saving it, so the changes are added to it?

blankse:
Its a BC Break, we use the Event to change Field Values and use this save methods to save it. If these a gone we need to execute the save methods by yourself.

davidmaes:
> Break

That is impossible, because your changed to the field values in the event itself are what is being fixed here.

davidmaes:
> Its a BC Break, we use the Event to change Field Values and use this save methods to save it. If these a gone we need to execute the save methods by yourself.

@blankse Also, your code is then broken when you push an object through the transition programmatically via the WorkflowRegistry. Because then you won't hit this part of the code, and you are missing a save anyway.

blankse:
Yes we need the workflows only for the backend now. 
It would be good for v10 to solve this.
But I think also its a BC if someone like me change field values in the event.

dpfaffenbauer:
I do change field values in events for the workflow. The workflow is triggered from the frontend and it works just fine....

PR #7702, opened 2020-12-09
Removed API link, updated docs, & https usage

NiklasBr:
By the way https://talk.pimcore.org doesn't exist any more either, will that be restored or migrated?


brusch:
@NiklasBr Thanks a lot! 
Regarding https://talk.pimcore.org : we are moving over to GitHub Discussions, which makes much more sense for us to have everything integrated here on GitHub. Regarding migration: currently there doesn't seem to be a way of doing that 😐

brusch:
Just for reference: https://github.com/pimcore/pimcore/discussions/7704

PR #7706, opened 2020-12-09
Deep copy in Version::marshalData() corrupts in-memory Document object

### LINKED ISSUE #7628: Deep copy in Version::marshalData() corrupts in-memory Document object
# Bug Report
This appeared when upgrading to 6.8 (v6.8.5) using both PHP 7.3 and 7.4.
We have some custom EventHandlers that do validation and data postprocessing on Relations in Pages.
In `pimcore.document.preUpdate` the Page object passed to the event looks fine.
In `pimcore.document.postUpdate` we run into a strange crash:
`[Error] Call to undefined method Pimcore\Model\Element\ElementDescriptor::getProperty()`

Condensing the code we run, it's
`DocumentEvent::getDocument()->getEditables()['mainArea:1.relatedObject']->element->getProperty('x')`

Using xdebug to look at the data when the Page is saved:
'mainArea:1.relatedObject' is an instance of `Pimcore\Model\Document\Editable\Relation`
- before Version::save():
 Relation::$element is an instance of `Pimcore\Model\DataObject\Article`
- after Version::save():
 Relation::$element is an instance of `Pimcore\Model\Element\ElementDescriptor`

so it seems like DeepCopy->copy() is also modifying the source object, and not just the clone, resulting in a corrupted Page object being passed to `pimcore.document.postUpdate`.

### Expected behavior
identical objects are passed to `pimcore.document.preUpdate` and `pimcore.document.postUpdate`

### Actual behavior
the Document object passed to `pimcore.document.postUpdate` contains marshalled relations.

### Steps to reproduce
- create a Page with Relation Editables on the page.
- create an EventHandler for `pimcore.document.postUpdate` that includes the code example below

### Code example
```php
        foreach ($event->getDocument()->getEditables() as $editable) {
            if ($editable instanceof Relation) {
                $editable->getElement()->getProperties();
            }
        }
```

weisswurstkanone:
resolves #7628


pmuskovac:
Looks good. Added the patch to our environment and the Page objects look correct again, and all related tests are passing.

models/Document/PageSnippet.php
```diff
@@ -678,12 +678,12 @@ public function getHref()
         return $this->getFullPath();
     }
 
-    /**
-     * @TODO: remove with $this->elements
-     */
     public function __wakeup()
     {
-        $this->editables = & $this->elements;
+        if ($this->editables === null && $this->elements !== null) {
```
dvesh3, line 683: @weisswurstkanone why not incorporate changes for $inheritedElements as well 😉

models/Document/PageSnippet.php
```diff
@@ -110,7 +111,6 @@ abstract class PageSnippet extends Model\Document
 
     public function __construct()
     {
-        $this->elements = & $this->editables;
         $this->inheritedElements = & $this->inheritedEditables;
```
jdreesen, line 114: This should be removed then, too, I think (and be made private)

weisswurstkanone: @jdreesen 👍

models/Document/PageSnippet.php
```diff
@@ -99,20 +100,15 @@ abstract class PageSnippet extends Model\Document
     /**
      * @var array
      *
-     * @deprecated since v6.7 and will be removed in Pimcore 10. Use getter/setter methods or $this->inheritedEditables
+     * @deprecated since v6.7 and will be removed in Pimcore 10. Use getter/setter methods h
```
dvesh3, line 103:
```suggestion
     * @deprecated since v6.7 and will be removed in Pimcore 10.
```

models/Document/PageSnippet.php
```diff
@@ -678,12 +674,17 @@ public function getHref()
...
-     */
     public function __wakeup()
     {
-        $this->editables = & $this->elements;
+        if ($this->editables === null && $this->elements !== null) {
+            $this->editables = $this->elements;
+            unset($this->elements);
+        }
+
+        if ($this->inheritedEditables === null && $this->inheritedElements !== null) {
```
dvesh3, line 684:
```suggestion
        if (empty($this->inheritedEditables) && !empty($this->inheritedElements)) {
```

models/Document/PageSnippet.php
```diff
@@ -99,20 +100,15 @@ abstract class PageSnippet extends Model\Document
     /**
      * @var array
      *
-     * @deprecated since v6.7 and will be removed in Pimcore 10. Use getter/setter methods or $this->inheritedEditables
+     * @deprecated since v6.7 and will be removed in Pimcore 10.
```
jdreesen, line 103:
```suggestion
     * @deprecated since v6.7 and will be removed in Pimcore 10.
     *                From 6.9 on this property will be private instead of protected.
```

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,5 +1,7 @@
 # Upgrade Notes
+
 ## 6.9.0
+- `PageSnippet::$elements` property visibility will change from `protected` to `private`
```
jdreesen, line 4:
```suggestion
- `PageSnippet::$elements` property visibility changed from `protected` to `private` 
- `PageSnippet::$inheritedElements` property visibility changed from `protected` to `private` 
```

weisswurstkanone: @jdreesen merci!

models/Document/PageSnippet.php
```diff
@@ -60,9 +60,10 @@ abstract class PageSnippet extends Model\Document
      *
      * @var array
      *
-     * @deprecated since v6.7 and will be removed in Pimcore 10. Use getter/setter methods or $this->editables
+     * @deprecated since v6.7 and will be removed in 10. Use getter/setter methods getEditables/setEditables instead.
+     *                From 6.9 on this property will be private instead of protected.
```
jdreesen, line 64: Does this make sense? Actually it is already `private` here.

It should probably rather be added in the `6.8` branch.

(And it's missing for `$inheritedElements`)

brusch: We thought `6.9` would be better, as it potentially could have an impact on existing project related code.

jdreesen: What I meant was, that only the message `From 6.9 on this property will be private instead of protected.` goes into the branch `6.8` while the actual change goes into `6.9.` (like it is actually targeted).

weisswurstkanone: @jdreesen valid and good point 👍

weisswurstkanone: @jdreesen see https://github.com/pimcore/pimcore/pull/7706/commits/bffe7251107cba22d8a2730587dfbc7e3d860ddd

PR #7720, opened 2020-12-10
Boolean select preview

t-zilla:

Fixed version preview of Boolean Select fields. Currently, false and empty values are indistinguishable and appear as empty strings.

## Additional info  

The reason for the change in `DataObject\Concrete` is that its method is called in `bundles/AdminBundle/Resources/views/Admin/DataObject/DataObject/previewVersion.html.php:207` which makes it impossible to later distinguish a field that's unset/empty from a field with value false.


brusch:
Nice, thanks a lot!

PR #7730, opened 2020-12-11
Log metadata Error for getXMPData

putzflorian:
upload of images disrupted if metadata has an error

This Log an Error Message and continue the Upload of the Image


weisswurstkanone:
could you check travis please ?

weisswurstkanone:
thx!

PR #7734, opened 2020-12-11
[Settings] Fulltext search in locale dropdown

BlackbitNeueMedien:
In System settings > Localization & Internationalization (i18n/l10n) there is a combo box for adding new locales. Currently this only supports searching for prefixes. I just had the pleasure to find `en_gb` in this list - took me nearly a minute ;-)
With this PR the typeahead feature does full text search on all available locales instead of only prefix search.


brusch:
Makes absolutely sense, thanks a lot @BlackbitNeueMedien !

PR #7735, opened 2020-12-11
Localized fields in bricks translation import bugfix

markus-moser:
The translation import of localized fields in bricks currently does not work anymore. The reason is that a PHPStan level 2 fix added a check for a wrong parent class. 

Generally it would be a good idea to test each single line if we do such fixes/changes. I appreciate it really much if we improve the PHPStan level, but in this case I've got really serious problems and needed to fix missing data via scripts...


weisswurstkanone:
@markus-moser rather a topic for 6.8? could you please rebase?

markus-moser:
@weisswurstkanone I switched the branch to 6.8

weisswurstkanone:
Hi @markus-moser could you please check the issue phpstan reported ? 
https://travis-ci.com/github/pimcore/pimcore/jobs/457983993

PR #7737, opened 2020-12-11
[WIP] improve display of scheduled versions

jdreesen:
Today I studied the scheduled publication of documents a bit and I find the current scheduling behavior very confusing.

My observations so far are as follows:

1. The "Version" field in the "Schedule" tab is only considered when the "Publish Version" action is selected - but if you select another action (Publish, Unpublish, Delete) you can still select a version if you want to.
1.1. If you *do* select a version, the according version gets displayed as "Scheduled" in the "Versions" tab by displaying the time of the task in this column. But if the action is not "Publish Version", this version is **not** the version that the task operates on!

2. If you select the "Publish" action, some "opaque" version will be published when the task runs. 
2.1. Which version this will be can nowhere be seen (as far as I know).
2.2. I expected this to be the latest available version, but apparently it's the version that was published last - which means it will be the **first** version (the raw document that was created when the user clicked "add page" - without any changes) if the page has never been published yet.

#### So my question is: if you want to publish a document for the first time using the scheduler, you **have** to use the "Publish Version" action if you don't want a blank page to be published. Is that correct?

---

To tackle the problems described in `1.1` and `2.1` I tried to improve this a bit with the PoC in this PR:
1.1: With this code only the selected versions of "Publish Version" tasks are considered in the "Versions" task. 
2.1: Additionally, the next scheduled "Publish" task gets displayed at the real version it operates on. 

The "Unpublish" or "Delete" tasks don't get displayed in the "Versions" tab because they operate independent of the versions displayed there:
- "Unpublish" unpublishes the currently published version (there can only be one)
- "Delete" deletes the whole document, independent of any versions

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/ElementController.php
+++ b/bundles/AdminBundle/Controller/Admin/ElementController.php
@@ -602,7 +602,7 @@ class ElementController extends AdminController
                         if ($task->getActive()) {
                             if ($task->getAction() === 'publish-version' && $task->getVersion()) {
                                 $schedules[$task->getVersion()] = $task->getDate();
-                            } else {
+                            } elseif ($task->getAction() === 'publish') {
                                 $hasSchedule = $task->getDate();
                             }
                         }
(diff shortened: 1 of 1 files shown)


dvesh3:
@jdreesen PR is closed coz we deleted the branch 6.8 and now new maintenance branch is 6.9. could you please recreate it based on 6.9? thanks! sorry for the inconvenience.

PR #7740, opened 2020-12-11
Fix #7739: Localized field mustn't break context for the non-localized fields

### LINKED ISSUE #7739: Localized field in Objectbrick breaks the context for all the following non-localized fields.
# Bug Report

### Expected behavior
Pimcore displays variants in the variants grid.

### Actual behavior
Pimcore displays an error popup message "Call to a member function supportsInheritance() on null".

### Steps to reproduce
Create an ObjectBrick with two fields: one is localized, placed above another, non-localized field. Create an object of this class. Create a variant, the non-localized field must be allowed to be displayed in the variants grid. Try to open the parent object and show the variants.
Unfortunately, I'm unable to reproduce it on a demo server as it displays another error: No alive nodes found in your cluster.

### Code example
https://github.com/pimcore/pimcore/blob/b3fbaeb862479e7b3213c2a58ec4ea1e82ddb672/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php#L773-L778

CountZero1981:
Resolves #7739

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -770,12 +770,12 @@ class DataObjectHelperController extends AdminController
                 if ($bf instanceof DataObject\ClassDefinition\Data\Localizedfields) {
                     $localizedFieldDefinitions = $bf->getFieldDefinitions();
 
-                    $context = [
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
@CountZero1981 could you please rebase this to 6.8 ?

brusch:
Fixed by #7749 (which is on the right branch) - thanks a lot!

PR #7747, opened 2020-12-14
[Data Objects] Retry transaction when tables for newly added locale do not exist yet

BlackbitNeueMedien:
Steps to reproduce bug:
* Create data object class `Test` with localized text field `test`
* Add new locale in system settings > localization / i18n
* Create new object of class `Test`
* An exception occurs `missing table created, start next run ... ;-)`

As https://github.com/pimcore/pimcore/blob/eb493066f5c3bafc68f1a9540886bd42161758d9/models/DataObject/Localizedfield/Dao.php#L260-L261 says this exception should cause the saving transaction to be retried - but currently this is not done because https://github.com/pimcore/pimcore/blob/eb493066f5c3bafc68f1a9540886bd42161758d9/models/DataObject/AbstractObject.php#L665-L695 does not handle this type of exception separately.
This PR fixes this by introducing a new exception class `LanguageTableDoesNotExistException` (which inherits DBAL's RetryableException) and by restarting the save transaction for all RetryableExceptions (not only DeadlockException).

And as little side note it fixes the little UI quirk in editor languages settings (table is broader than columns):
Before:
editor-settings-bug

After:
editor-settings-bug-solved

### DIFF
--- a/models/DataObject/Localizedfield/Dao.php
+++ b/models/DataObject/Localizedfield/Dao.php
@@ -17,6 +17,7 @@
 
 namespace Pimcore\Model\DataObject\Localizedfield;
 
+use Doctrine\DBAL\Exception\DeadlockException;
 use Pimcore\Db;
 use Pimcore\Logger;
 use Pimcore\Model;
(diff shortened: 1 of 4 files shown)


models/DataObject/Localizedfield/Dao.php
```diff
@@ -17,6 +17,7 @@
 
 namespace Pimcore\Model\DataObject\Localizedfield;
 
+use Doctrine\DBAL\Exception\DeadlockException;
```
brusch, line 20:
```suggestion
```

brusch (changes requested):

brusch:
Thanks a lot! 👍😊

PR #7752, opened 2020-12-14
[Data objects] Fallback to first matching locale if requested language does not exist

BlackbitNeueMedien:
Currently there is the following problem:
1. Set current user's account language to `English`
1. Add locales `gb` as default language (if not already present), `de_DE`, `en_GB` etc. in system settings > localization / i18n
3. Create class `Category` with localized input field `name`
2. Create class `Product` with many-to-many object relation `categories`, allowed class `Category`and visible field `name`
4. Create `Category` object with name (de_DE) = `Kategorie de`, name (en) = `Category en`, name (en_GB) = `Category en_gb`
5. Create `Product` object and assign category object to `categories` relation -> see column `name` being filled with `Category en` (correct) -> save object
6. Change current user's account language to `German`
7. Reload Pimcore and open the `Product` object again -> column `name` of `categories` relation still filled with `Category en` and not `Kategorie de`

The problem is that we added `de_DE` as valid locale but the current user's language is `de`. When Pimcore cannot find the requested locale, it falls back to the default language - and this is still `English` (see 2.)

This PR changes this behaviour so that when an invalid language gets requested it is tried to find a valid locale which has this language as prefix. 
This also works for example when you have a system with `German` being the default language but you also have locales `en_GB` and `en_US`. Currently users with language `English` would see values in German. With this PR they see values in the first valid locale for `en_*` (which is much more understandable to them than the default language German in this case).


dvesh3:
@BlackbitNeueMedien tests are failing, so please rebase master 😉

BlackbitNeueMedien:
@dvesh3 Actually thought I did today morning, but perhaps haven't `git fetch`ed. Now it should work...

brusch:
Thanks a lot for your PR! However, in my opinion the current behavior is correct as it shouldn't be based on the users locale.
With this PR the behavior would be different for users with different locales and it would also make it more difficult to debug, since the behavior isn't following the configuration anymore. 

So imho we should close this PR and let the behavior entirely based on the config.

BlackbitNeueMedien:
What do you mena with user's locale? Do you mena the browser locale? This is not getting used at all.

The problem simply is that currently backend languages are languages-only (like de, en etc.) and localized field languages can be locales (de_DE, de_AT, en_GB, en_US etc.).

When a user has set his backend language to German and the default language in system settings is English, he will not be able to see localized values for de_DE for visible fields in relations for example.

BlackbitNeueMedien:
Any news on this @brusch ? In your last post I had the impression that there was a misunderstanding in what this PR actually does.

brusch:
@BlackbitNeueMedien  sorry for the late response 😊 
I think there was a misunderstanding, what I mean is, that your changes should only apply within the admin interface, but shouldn't be the general behavior. 
So I'd suggest to just wrap your changes in `if(\Pimcore::inAdmin())` and keep the `throw ...`. 😊

models/DataObject/Localizedfield.php
```diff
@@ -310,6 +310,15 @@ public function getLanguage(string $language = null): string
             if (Tool::isValidLanguage($locale)) {
                 return $locale;
             }
+
+            if (\Pimcore::inAdmin()) {
+                foreach (Tool::getValidLanguages() as $validLocale) {
+                    if (strpos($validLocale, $locale.'_') === 0) {
```
brusch, line 316:
```suggestion
                    if (str_starts_with($validLocale, $locale.'_')) {
```

PR #7756, opened 2020-12-15
Revert "Fix batch edit null error"

brusch:
Reverts pimcore/pimcore#7519

Causes other problems see @skochTC  comment here: https://github.com/pimcore/pimcore/commit/9c4b38a463519e850936f22dc5475fc1ad2d9c2b#commitcomment-45129223


skochTC:
Thanks 👍

PR #7757, opened 2020-12-15
[Ecommerce][Productindex][Elasticsearch] fix execution of process update index queue command with multiple tenants

markus-moser:
The `ecommerce:indexservice:process-update-queue` does not work corretly if used with multiple tenants (which do not contain the same products). The `commitBatchToIndex()` will only be executed for the tenants of the first combined product row. Therefore it's needed to call the command multiple times to get all items processed.

This PR collects all relevant workers of all rows and therefore the `commitBatchToIndex()` will be executed for all of them. Now a single call of the command is enough to process all items in the queue!


brusch:
Thanks a lot! 👍

PR #7759, opened 2020-12-15
Symfony Cache Tags

dpfaffenbauer:
Symfony Cache does not consider the key a tag, add key as tag in CoreCacheHandler


brusch:
Thanks a lot! 👌 - related to the changes made by #7635

PR #7767, opened 2020-12-16
install cache on installation

dpfaffenbauer:
Install cache tables, otherwise it creates unneeded and confusing logs.


brusch:
Makes perfectly sense, thanks a lot! 👌

brusch:
This seems to cause a new issue -> reverting. 

```
The service "Pimcore\Bundle\InstallBundle\Command\InstallCommand" has a dependency on a non-existent service "pimcore.cache.adapter.pdo".
```

PR #7769, opened 2020-12-16
[Assets] Fix "allowed memory size exhausted" error

aweichler:
Fix an error, which occurs when using "Download selected items as ZIP" function at a folder view in assets.
It occurs when selecting large files. In our case there where 3 Videos ~200M. PHP Memory Limit was at 256M or 512M.
The function "file_get_contents" reads the whole file into memory before inserting it into the ZIP archive. With "addFile" of the class "ZipArchive" this limitation is bypassed.


brusch:
Nice, thanks a lot! 👌

PR #7774, opened 2020-12-16
Create cache table on install

brusch:
Fixes #7767


dpfaffenbauer:
ok, cool, makes sense too. as long as the table is created I am happy with it :)

PR #7776, opened 2020-12-16
Allow to create editables in new `pimcore.document.editables` namespace

jdreesen:
The docs say that the JS for custom editables should be created in the `pimcore.document.editables` namespace (since `pimcore.document.tags` was deprecated). But this is currently not possible. This PR fixes that. Also, an improved error message is thrown when an editable class does not exist.
<br>
Note: this PR is best looked at [without whitespace changes](https://github.com/pimcore/pimcore/pull/7776/files?w=1) ;)


brusch:
Thank you!

PR #7777, opened 2020-12-16
Fix date editable

jdreesen:


bundles/AdminBundle/Resources/public/js/pimcore/document/editables/date.js
```diff
@@ -38,8 +38,6 @@ pimcore.document.editables.date = Class.create(pimcore.document.editable, {
             this.config.value = this.data;
         }
 
-        this.config.name = id + "_editable";
```
jdreesen, line 41: This was called in the wrong place, because `id` isn't defined here. So we better do this in `initialize`.

models/Document/Editable/Date.php
```diff
@@ -94,7 +92,6 @@ public function frontend()
      */
     public function getDataForResource()
     {
-        $this->checkValidity();
```
jdreesen, line 97: I don't think this makes sense here because this class doesn't implement the `checkValidity` method, which means the one from `parent` is called which only returns `true`.

jdreesen (commented):
Instead of the PR description the changes are described here ;)

models/Document/Editable/Date.php
dvesh3: yes, good catch!

dvesh3:
thank you!

PR #7779, opened 2020-12-16
Resolve #7764 - Doc copy and paste for table data type

### LINKED ISSUE #7764: [bug] Table datatype: paste clipboard on object instance don't work, it paste only one column
Hello,

On a table datatype on an object instance, using "paste" feature, it only fill the first colum.
I tried with | , ; as data separator with or without " or ', but it always fill juste the first column.
Is there some specific format to use or is there a bug to fix ?

On the online demo create a class like:

image

Create an object, open the paste popin:

image

And copy/paste date like:
```
keyA|valA
keyB|valB
```
(this is format used here: https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Data_Types/Table.html)

It gives:

image

Use case: if data comes from a CSV files having many lines and that data are updated regularly, using paste feature is very conveniant for this as it prevent having to fill all manually!

Best regards.

Yivan:
Resolve #7764 by updating documentation for table data type for copy&paste feature.


weisswurstkanone:
Excellent, thanks a lot!

PR #7786, opened 2020-12-17
DBAL Schema - Generator

dpfaffenbauer:
Move Schema Generation to DBAL Schema and let DBAL do the diff. This allows to deploy migrations without touching class-definition.

This is a super early draft and be considered a POC.

Also extracts the File Generation into separate services, this would allow to just create PHP Files, without the need of DB changes. (Separation of concerns)


dpfaffenbauer:
Most important is the new SchemaBuilder: He is responsible for creating the Tables for a ClassDefinition. This should be more structured and be distributed better. Maybe also have Data Classes be able to create tables. Not sure, since this is a PoC we can see where it goes.

brusch:

PR #7800, opened 2020-12-22
[Translations] Deprecated Translation\Website|Admin and Introduced new Translation model

### LINKED ISSUE #7504: Deprecate Translation\Website|Admin introduce Translation
Introduce new `Translation` Model, which just extends `AbstractTranslation`. 

Deprecate: 
`AbstractTranslation`
`Website` (incl. sub-classes)
`Admin` (incl. sub-classes)

New translation model should have "domain" support and persist each domain in a dedicated table ( e.g. `translations_DOMAIN`), if the table doesn't exist, it should be created automatically. 

We should also make `messages` as the default and don't use `website` anymore.

dvesh3:
Resolves #7504


bundles/CoreBundle/Migrations/Version20201222101114.php
```diff
@@ -0,0 +1,28 @@
...
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20201222101114 extends AbstractPimcoreMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->addSql('RENAME TABLE translations_website TO translations_messages;');
```
brusch, line 18: I guess this is a bc-break, we should do the renaming in v10 and keep the current table name for now. 
We did that also here #7263

models/Translation/Dao/DaoInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ * @category   Pimcore
+ * @package    Translation
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Model\Translation\Dao;
+
+interface DaoInterface
```
brusch, line 20: I think there's no reason for keeping this interface, as `getTableName()` is now anyway generic and based on the domain.

models/Translation/Listing/Dao/DaoInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ * @category   Pimcore
+ * @package    Translation
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Model\Translation\Listing\Dao;
+
+interface DaoInterface
```
brusch, line 20: Same here, I guess this can go as well.

models/Translation.php
```diff
@@ -0,0 +1,478 @@
...
+use Pimcore\Event\TranslationEvents;
+use Pimcore\File;
+use Pimcore\Model\Translation\TranslationInterface;
+use Pimcore\Model\Translation\Website;
+use Pimcore\Tool;
+
+/**
+ * @method \Pimcore\Model\Translation\Dao getDao()
+ */
+class Translation extends Website implements TranslationInterface
```
brusch, line 32: Maybe we should just rename `models/Translation/AbstractTranslation.php` to `models/Translation/Translation.php` (and the class name) to keep the Git history and then do the following for the `AbstractTranslation`

```php
abstract class AbstractTranslation extends Translation
```

This would keep things comaptible and we're not duplicating code.

brusch (changes requested):

bundles/CoreBundle/Migrations/Version20201222101114.php
brusch: We can think of using a view as kind of an alias, I guess that would work as well. 
Then in v10 we just drop the BC view.

models/Translation.php
brusch: This would also make #7808 easier, as we're just deleting the deprecated classesand while keeping the history of the `AbstractTranslation` in `Translation`.

bundles/CoreBundle/Migrations/Version20201222101114.php
dvesh3: Added BC view.

models/Translation/Dao/DaoInterface.php
dvesh3: Kicked it.

models/Translation/Listing/Dao/DaoInterface.php
dvesh3: Kicked it.

brusch (changes requested):
Did some code cleanup and changes in f216977ec11366aaf96433f5c3284f661974b900, please review again 😊 Thanks a lot!

PR #7804, opened 2020-12-22
Use \Browser to detect device and move default setting to later in method

NiklasBr:


brusch:
We're taking away the device detector, see #7807
Thanks anyway 😉

NiklasBr:
@brusch see my reply in the linked issue. I think you have misunderstood what Google is doing.

PR #7805, opened 2020-12-22
[CustomReport] Fix mixed grid/drilldown filters in SQL Adapter

rquarta-extera:
Resolves a bug when mixed grid and drilldown filters are used in SQL adapter inside a custom report.

## To reproduce

https://user-images.githubusercontent.com/43776120/102893558-15e8e380-4462-11eb-8a38-9044a47be5f5.mp4


dvesh3:
thanks a lot!

PR #7793, opened 2020-12-18
Provide Doctrine Query Builder on Listing classes

dvesh3:
## Changes in this pull request  
Resolves #7744

PR #7817, opened 2020-12-28
Introduced support for Doctrine Query builder on listing classes

### LINKED ISSUE #7744: Provide Doctrine Query Builder on Listing classes
Alternative (and replacement in v10) for `getQuery()` on listing classes (e.g. `\Pimcore\Model\DataObject\Listing\Concrete\Dao::getQuery` )

brusch:
Resolves #7744
Alternative approach for #7793

This approach will use the legacy Zend query builder if `onCreateQuery()` is used and the Doctrine QB otherwise. Imho this shouldn't cause any bc-breaks while you can use already the Doctrine QB.


dvesh3:
@brusch ready for review 😉

HageMaster3108:
Guess it'd be good to have this in the docs?
https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Working_with_PHP_API.html#page_Access-and-modify-internal-object-list-query

brusch:
@HageMaster3108 Thanks for the hint, done: https://github.com/pimcore/pimcore/commit/0ffd3aafaf47e9f9eb3a1ca99cfe6abdbceb6722

PR #7846, opened 2021-01-04
Improve error message when missing permission to create assets

danmichaelo:
### Current behaviour

Given that you don't have the "create" permission for an asset folder, then the frontend will automatically hide many of the options for uploading files to that folder, but there's still some options left, where it will be up to the backend to reject the upload. I've found two:

1. when you drag-n-drop a file onto any folder
2. when you upload a file to a media component on a data object.

In those cases, AssetController returns a 200 response with `{status: false}`, which results in a rather sparse error message:

Screenshot 2021-01-04 at 18 44 27

### Proposed change

This PR uses `createAccessDeniedHttpException()` method to create a proper exception, resulting in a 403 response and an error message like this:

Screenshot 2021-01-04 at 19 45 06

I chose to use the word "create" over "upload", since it's the name of the permission you need.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -396,7 +396,6 @@ class AssetController extends ElementControllerBase implements EventedController
      */
     protected function addAsset(Request $request, Config $config)
     {
-        $success = false;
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
thx!

PR #7847, opened 2021-01-05
Incompatibilities due to fixed Doctrine server_version

brusch:
reverts 12c936dc6695dd9ac40ec6acc84b194194bd33b3

The problem with hard-coding the `server_version` is that schema/table comparisons are platform specific. This causes problems when for example running `./bin/console doctrine:migrations:version --all --add -n -q`. 

```text

In MetadataStorageError.php line 13:

  The metadata storage is not up to date, please run the sync-metadata-storage command to fix this issue.


doctrine:migrations:version [--add] [--delete] [--all] [--range-from [RANGE-FROM]] [--range-to [RANGE-TO]] [--configuration CONFIGURATION] [--db-configuration DB-CONFIGURATION] [--prefix [PREFIX]] [--] [<version>]
```

### DIFF
--- a/bundles/CoreBundle/Resources/config/pimcore/default.yml
+++ b/bundles/CoreBundle/Resources/config/pimcore/default.yml
@@ -47,7 +47,6 @@ doctrine:
                 charset: UTF8MB4
                 logging: false
                 profiling: false
-                server_version: '5.6'
                 default_table_options:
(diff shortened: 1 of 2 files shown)


brusch:
haven't seen any issues with `6.x` , so we can keep it there for now.

PR #7857, opened 2021-01-07
Add missing field pricingRuleId to field collection OrderPriceModification

IgorBenko:
Resolves a problem with a missing field `pricingRuleId` in the field collection `OrderPriceModification`.

This field is normally added with the migration Version20200113120101.

In the following scenario this migration will not do any changes:
- Pimcore is installed
- during install all pimcore_core migrations are executed
- the migration Version20200113120101 checks if Ecommerce Bundle is installed. Nothing is changed at this point as the bundle isn't there yet, see: `https://github.com/pimcore/pimcore/blob/6.9/bundles/CoreBundle/Migrations/Version20200113120101.php`
- the migration is marked as successful
- we enable and install the ecommerce framework
- the field definition from `bundles/EcommerceFrameworkBundle/Resources/install/fieldcollection_sources/fieldcollection_OrderPriceModifications_export.json` is installed
- the pricingRuleId is missing and it's never added

This pull request adds the file directly to the class definition in the bundle.


fashxp:
thx very much.
since this is a bugfix, can you please rebase to 6.8?

IgorBenko:
Hi Christian, 
I rebased to 6.8

fashxp:
thx very much!

PR #7860, opened 2021-01-07
Remove data type BC layer ('multihrefMetadata','objectsMetadata','objects','multihref','href','nonownerobjects')

blankse:
See #3653 and #3728


brusch:
Nice, thanks a lot! 👍
Follow-up to #7308

PR #7866, opened 2021-01-07
Require Composer ^2.0

### LINKED ISSUE #7419: Make Composer 2.0 requirement

brusch:
Resolves #7419
See also: 
https://github.com/pimcore/demo/commit/dafcb86a512fad19d8dfd60a510397f31601b5b1
https://github.com/pimcore/skeleton/commit/bf80dd8e96f814ef6b629c265dc518d1fb378339

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,7 @@
     "ext-pdo_mysql": "*",
     "ext-zip": "*",
     "ext-zlib": "*",
+    "composer-runtime-api": "^2.0",
     "psr/simple-cache": "^1.0.0",
     "amnuts/opcache-gui": "^3.0.1",
     "cbschuld/browser.php": "^1.9.6",


dpfaffenbauer:
I don't get the point of this?

brusch:
@dpfaffenbauer same as with PHP 8, it's just to be prepared for the future 😉

fashxp:
and make sure that shipped composer.lock files work properly (because depending on composer versions different dependencies might be loaded) ... but that would be maybe more a topic of the skeletons...

NiklasBr:
That shoudn't be a problem from my understanding:
> - `composer.lock` files are interoperable between versions, so you can upgrade to 2.0 and roll back easily if needed.

fashxp:
not if some dependencies, that are locked in the lock file, require composer 2 ... then you can't install pimcore with composer 1 anymore. 
at least we had that issue already in some constellations.

dpfaffenbauer:
Requirements are the problem here, "ocramius/package-version" in Pimcore's case. You should rather require the root-package causing it, instead of composer v2

brusch:
Well, not much difference, isn't it?

PR #7868, opened 2021-01-07
Remove deprecated getElement() method on NotificationEvent and TranslationEvent

blankse:
See #7077


brusch:
Thank you very much!

PR #7844, opened 2021-01-04
PHP 8 Support

brusch:
Dependencies:

PR #7870, opened 2021-01-07
ocramius/package-versions: ^2.3

brusch:
Follow up to #7869 should be merged after #7844 was merged (PHP 8 dependency)


brusch:
Integrated into #7844 -> closed

PR #7876, opened 2021-01-07
Make all UserRole setters chainable

danmichaelo:
All the setters in `AbstractUser` and many in `UserRole` were already chainable, so this is just to make the remaining setters in `UserRole` chainable too.

Not sure if this should go to master or 6.9. I'm fine with both, let me know if I should rebase.


brusch:
Thank you!

PR #7878, opened 2021-01-08
 Removed colon-in-path-workaround for imagemagick

fscholl:
Resolves an issue where imagemagick couldn't find images with a colon ( : ) in the path.
The removed workaround prefixes a path that contains a colon with another colon and therefore imagemagick can't find the image and throws an error (the error is from a Windows system, but was reproducible with Ubuntu): 
`
pimcore.ERROR: ImagickException: UnableToOpenBlob ':C:\folder\structure/public/var/assets/logo/logo-3x.png[0]': Invalid argument @ error/blob.c/OpenBlob/3315
`

## Additional info  

The workaround was introduced in #2829 but the underlying issue was fixed later that year by imagemagick itself https://github.com/ImageMagick/ImageMagick/issues/1402.


brusch:
Thanks a lot! However I'd rather merge this into the `master` branch, since it potentially changes the behavior for existing installations. Can you please rebase? Thanks again.

blankse:
The colon should only be set, if in the path is no colon. See https://github.com/pimcore/pimcore/pull/2888
In your path is a colon, so this shouldn't be happened.

Not everone have ImageMagick v7. For example we still have v6.8. 
But I think its okay for Pimcore X

brusch:
Ah, didn't notice that this is ImageMagick v7, there was no milestone assigned on GitHub 😊
If this is v7 only I wouldn't change it for now, since a lot of linux distributions (such as Debian, ...) are still delivering v6 as the default.

blankse:
In the issue are commits linked. There are also commits that tagged in v6.9

brusch:
ok, then it's probably fine. 
@fscholl please rebase on `master`  😊 Thanks in advance!

fscholl:
> 
> 
> The colon should only be set, if in the path is no colon. See #2888
> In your path is a colon, so this shouldn't be happened.
> 
> Not everone have ImageMagick v7. For example we still have v6.8.
> But I think its okay for Pimcore X

I wasn't aware that this was ImageMagick 7 only, since the fix was from 2018. Also your code behaves correctly because it should set a colon if there is a colon in the path (You probably misread your old commit there https://github.com/pimcore/pimcore/pull/2888/commits)

@brusch feel free to close the PR if you've come to a final decision about the v6/v7 compatibility.

brusch:
@fscholl I guess it's fine for v10, can you please rebase your PR onto the master? Thanks 😊

fscholl:
I'm sorry somehow I can't get the rebase working. I'll create a new clean PR for the master and close this one.

PR #7888, opened 2021-01-11
[Calculated value fields] Support display type HTML

BlackbitNeueMedien:
Currently calculated value fields support display types "input field" and "textbox". This PR adds support for HTML output allowing to fully flexibilize calculated field output. In the end this seems to be the same as dynamic text layout element but this for example lacks multi-language support and a getter method. So to support dynamically generated HTML in the "Edit" panel of the object which is also possible to be reused via its getter method I added the "HTML" display type to calculated value fields.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/calculatedValue.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/calculatedValue.js
@@ -52,7 +52,8 @@ pimcore.object.classes.data.calculatedValue = Class.create(pimcore.object.classe
                 labelWidth: 140,
                 store: [
                     ['input', t('input')],
-                    ['textarea', t('textarea')]
+                    ['textarea', t('textarea')],
+                    ['html', t('HTML')]
                 ]
             },
             {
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/calculatedValue.js
```diff
@@ -52,7 +52,8 @@ pimcore.object.classes.data.calculatedValue = Class.create(pimcore.object.classe
                 labelWidth: 140,
                 store: [
                     ['input', t('input')],
-                    ['textarea', t('textarea')]
+                    ['textarea', t('textarea')],
+                    ['html', t('HTML')]
```
weisswurstkanone, line 56: shouldn't this be lower cased? at least I couldn't find it in the master translation file.

BlackbitNeueMedien: Oh sorry, you are right. Fixed it.

weisswurstkanone:
great, thank you!

PR #7891, opened 2021-01-12
[Object bricks] Brick container class might contain removed brick types after unserializing

BlackbitNeueMedien:
In the same context like in #7407 we now had the problem that we could not save an object which previously had an object brick assigned which got removed in the meantime. The problem is that the brick container class gets serialized together with the data object - and so `$this->brickGetters` also contains the brick type which got removed in the meantime. When we now want to save the object the following code gets executed: 
https://github.com/pimcore/pimcore/blob/69b4285b91e926eb6c70d9ccc83f301d7a32bdf8/models/DataObject/Objectbrick.php#L177-L180
Here the getter for the removed brick type gets called causing a fatal error because the getter does not exist anymore.

So the main problem is that the non-existing brick type remains present in `$this->brickGetters`. I also tried to remove the non-existing brick type in the `wakeUp` method https://github.com/pimcore/pimcore/blob/69b4285b91e926eb6c70d9ccc83f301d7a32bdf8/models/DataObject/Objectbrick.php#L325-L328 but this did not work but I have not researched any further why.

### DIFF
--- a/models/DataObject/Objectbrick.php
+++ b/models/DataObject/Objectbrick.php
@@ -135,12 +135,9 @@ class Objectbrick extends Model\AbstractModel implements DirtyIndicatorInterface
      */
     public function getBrickGetters()
     {
-        $getters = [];
-        foreach ($this->brickGetters as $bg) {
-            $getters[] = 'get' . ucfirst($bg);
-        }
-
-        return $getters;
+        return array_map(function($brickType) {
+            return 'get'.ucfirst($brickType);
+        }, $this->getAllowedBrickTypes());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Hi @BlackbitNeueMedien what do you think about this alternative approach ? https://github.com/pimcore/pimcore/pull/7912

BlackbitNeueMedien:
That is exactly what I meant with
```I also tried to remove the non-existing brick type in the `wakeUp` method https://github.com/pimcore/pimcore/blob/69b4285b91e926eb6c70d9ccc83f301d7a32bdf8/models/DataObject/Objectbrick.php#L325-L328 but this did not work but I have not researched any further why.```

Perhaps I did something wrong. Sadly I cannot reproduce the problem for the customer where this happened as now everything is cleaned up but actually #7912 should work... let's try and check again if it happens again.

brusch:
Closed in favor of #7912

PR #7808, opened 2020-12-22
[Translations] Consolidate Translation models & remove deprecated classes

dvesh3:
## Changes in this pull request  
Resolves #7505 

## Additional info  
Branched off from  #7800 - Once it is merged, then rebase this PR on master and change target-base to master.

PR #7893, opened 2021-01-12
Removed Migration Version20201222101114

brusch:
leftover of #7800


brusch:
will be covered by #7808

PR #7898, opened 2021-01-12
Fix Notice: Undefined index: data

blankse:
## Steps to reproduce

1. Set env `PIMCORE_PHP_ERROR_REPORTING=E_ALL`
2. Try to set focial point on new image asset
3. Get following error:
image


dvesh3:
thanks!

Issue #4056, opened 2019-03-12
[Ecommerce][Bug] Price rule AND/OR conditions


NiklasBr:
# Bug Report

### Expected behavior

The price rule condition Product AND in conjunction with the Product Discount action should apply to all matches and only those matching.

The price rule Product OR in conjunction with the Product Discount action should apply to any matches.

([Gitter conversation](https://gitter.im/pimcore/pimcore?at=5c82754f3162ec7bc73c0e60))

### Actual behavior

The rule condition Product AND causes all products in cart to get a Product Discount, even those not matching the condition.

The rule condition Product OR causes only the first product in the cart to get the Product Discount, even if other products match the condition.

### Steps to reproduce

1. Create a rule (active, additive) 
2. With the conditions [product: 123] AND [product: 567]
3. And the action [product discount: 10%]
4. Add product 123, 567, and 985 to the cart

Result: All products gets the product discount.
Expected behaviour: The two products 123 and 567 should be discounted.

5. Create a rule (active, additive)
6. With the conditions [product: 123] OR [product: 567]
7. And the action [product discount: 10%]
8. Add product 123, 567, and 985 to the cart

Result: Only the first product of 123 or 567 gets discounted.
Expected behaviour: The two products 123 and 567 should be discounted.

This might be a bit of semantic wrangling, but the last expected behaviour is not an OR behaviour it's an XOR behaviour. I would expect the OR to mean "may match this one or may match that one, may match both to return true for those matching".

AlternateIf:
I also experienced the second test case (OR Condition). It seems like the first item is always AND coupled. Until the bug is fixed, you can put a parenthese around your condition or you can move the first condition down then change the condition to OR and move it back up

![image](https://user-images.githubusercontent.com/14984260/54189915-12fe4800-44b3-11e9-9695-fa8fb47017c5.png)

dvesh3:
fixed by #7904

Daniel-Ateles:
This bug seems to have resurfaced in later versions, eg 10.4.1.

dvesh3:
@Daniel-Ateles can you please try on latest version 10.5.2?

Daniel-Ateles:
@dvesh3 now I'm on 10.5.2 and this problem still exists. Doesn't matter if I use paranthesis with AND/OR etc. All products in the cart get the discount, even products not in the rules.

Daniel-Ateles:
[...]

PR #7902, opened 2021-01-13
Fix first rule condition is always AND

tarolinus:
Resolves https://github.com/pimcore/pimcore/issues/4056

## Additional info  

As you can see in [this issue](https://github.com/pimcore/pimcore/issues/4056), the first operator is always `AND`

The first condition in a bracket shouldn't have an operator.


dvesh3:
@tarolinus thanks for your PR! please rebase your branch on 6.8 branch and change target base to 6.8, since it is a fix. thanks!

https://github.com/pimcore/pimcore/blob/master/CONTRIBUTING.md#contributing-pull-requests

PR #7904, opened 2021-01-14
Remove operator from first condition

### LINKED ISSUE #4056: [Ecommerce][Bug] Price rule AND/OR conditions

tarolinus:
The first condition should not have an operator (currently it is AND)

  
  

Resolves #4056

## Additional info  

Due to uncomfortable merge conflicts in https://github.com/pimcore/pimcore/pull/7902, I created a new branch/PR

### DIFF
--- a/bundles/EcommerceFrameworkBundle/PricingManager/Condition/Bracket.php
+++ b/bundles/EcommerceFrameworkBundle/PricingManager/Condition/Bracket.php
@@ -65,6 +65,11 @@ class Bracket implements BracketInterface
 
             // test condition
             $check = $condition->check($environment);
+            
+            if ($num === 0) {
+                $state = $check;
+                continue;
+            }
 
             // check
             switch ($this->operator[$num]) {


dvesh3:
thanks!

PR #7908, opened 2021-01-15
Update link to app.php in the upgrade docs

IgorBenko:
The app.php doesn't exist anymore, thus the link was broken. I put in a link that points directly to a commit before that change.


dvesh3:
thanks!

PR #7911, opened 2021-01-15
Remove innodb_file_format

yariksheptykin:
As of mariadb 10.2.2 this option is deprecated and as of 10.3.1 removed. See: https://mariadb.com/kb/en/innodb-system-variables/#innodb_file_format

The value defaults to `Barracuda`.


brusch:
Thank you!

PR #7916, opened 2021-01-17
6.9 Block Editable loop sorting random

tobbert:
Hi, 

i think a sorting feature for loop a Block Editable would be nice, especially for a random sorting.
Something like that:

$this->block("contentblock",["sort" => "rand"])->loop()

Options for sorting maybe: asc, desc, random

It would be attractive for slideshows, image-gallerys and other contents.

br 

  
  

Resolves #


dvesh3:
@tobbert Please use [Discussions idea section](https://github.com/pimcore/pimcore/discussions/categories/ideas) or create a feature request issue for proposals. thanks!

tobbert:
> @tobbert Please use [Discussions idea section](https://github.com/pimcore/pimcore/discussions/categories/ideas) or create a feature request issue for proposals. thanks!

allrighty

PR #7922, opened 2021-01-18
Pimcore Backend | Grid View | Value "0" is not shown

andreas-gruenwald:
Bugfix: when a select field has the value "0", then the actual select name should be shown in the grid view, not an empty value.

See screenshot:

image


weisswurstkanone:
could you please rebase this on 6.8 ? thx

andreas-gruenwald:
> could you please rebase this on 6.8 ? thx

I have bad internet today, so I created another PR which is based on V6.8. Hope that helps.

PR #7925, opened 2021-01-18
Explicitly generate non-HDPI and HDPI system thumbnails

limenet:

Generating system thumbnails using `bin/console pimcore:thumbnails:image --system` only outputs non-HDPI versions. When these thumbnails are then requested by calling `/admin/asset/get-image-thumbnail?hdpi=1`, the HDPI thumbnails are not available and have to generated on the fly which can cause high load (when e.g. the files are Photoshop files).

This PR generates the HDPI versions as well, provided `--skip-high-res` is not set. To make the code more explicit, the existing calls to `Asset\Image\Thumbnail\Config::getPreviewConfig()` now set the first argument to `false` when generating the non-HDPI version.

## Additional info  

The presence of the bug in version 6.8.6 and the absence of the bug with the patch have been verified.

### DIFF
--- a/bundles/CoreBundle/Command/ThumbnailsImageCommand.php
+++ b/bundles/CoreBundle/Command/ThumbnailsImageCommand.php
@@ -205,15 +205,19 @@ class ThumbnailsImageCommand extends AbstractCommand
             if (!$input->getOption('thumbnails')) {
                 $thumbnailsToGenerate = [];
             }
+
             $thumbnailsToGenerate[] = Asset\Image\Thumbnail\Config::getPreviewConfig(false);
+
             if (!$input->getOption('skip-high-res')) {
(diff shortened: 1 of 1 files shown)


brusch:
@limenet thanks a lot for your PR! 
I'd rather classify this as an improvement, can you please rebase your PR onto the `6.9` branch? Thanks again!

limenet:
@brusch you're welcome!

While I do see your point, the customer's server which we managed to overload due to this bug would appreciate this being a bugfix in the upcoming 6.8.7 so we can get this fixed asap. What do you think?

Otherwise I'll be happy to rebase the PR.

brusch:
@limenet Yes, please rebase onto `6.9` since `6.8` is really strictly just for bug fixes. Thanks!

limenet:
@brusch unfortunately I cannot change the source branch of this MR and had to create #7927 instead.

Issue #7926, opened 2021-01-18
Version Steps and Days "null" deletes all Versions


dpfaffenbauer:
Configuring Versioning by nulling or 0ing steps and days results in all Versions being deleted. Whereas I would expect that to disable the Version Cleanup completely.

brusch:
Partly resolved by #7928 
Closed in favor of #7944

PR #7928, opened 2021-01-18
Maintainance Task VersionCleanup: set default steps to 10

JHeimbach:

We updated a pimcore application from v5 to v6 and between updating the application and running the migration scripts the cronjob deleted some version because it couldn't find the system.yml configs.
The Versioncleanup job deleted all versions because the default step value is set to zero. 

I propose to change the default value to 10, because thats the value that is set when i'm installing pimcore.

### DIFF
--- a/lib/Maintenance/Tasks/VersionsCleanupTask.php
+++ b/lib/Maintenance/Tasks/VersionsCleanupTask.php
@@ -58,7 +58,7 @@ final class VersionsCleanupTask implements TaskInterface
 
         foreach ($conf as $elementType => $tConf) {
             $versioningType = 'steps';
-            $value = $tConf['steps'] ?? 0;
+            $value = $tConf['steps'] ?? 10;
 
             if (isset($tConf['days']) && (int)$tConf['days'] > 0) {
                 $versioningType = 'days';


dpfaffenbauer:
Related to https://github.com/pimcore/pimcore/issues/7926

I propose to disable version cleanup when step and days is null

JHeimbach:
Disabling if steps and days is null is also a good solution. 

if you want i can change my PR.

dpfaffenbauer:
It also seems that Pimcore doesn't save versions when you set days and steps to null....

dvesh3:
this change is ok for 6.8, since this fixes the issue of all versions being deleted with migration updates and let's disable cleanup with steps and days = null for Pimcore X, as it changes the behavior.

brusch:
I think this might be also relevant here: 
https://github.com/pimcore/pimcore/blob/master/models/Asset.php#L850
https://github.com/pimcore/pimcore/blob/master/models/Document/PageSnippet.php#L158
https://github.com/pimcore/pimcore/blob/master/models/DataObject/Concrete.php#L288

brusch:
@JHeimbach can you please update your PR accordingly? That would be nice, thanks in advance 😊

JHeimbach:
Hey @brusch 
i'm not sure what I should update.

As far as I understood @dvesh3 comment, we change the default 0 to a 10 in 6.8.7 and in Pimcore X we change so that no versions don't get cleaned up if the config is empty (null).

Shouldn't I rather create a new PR with the changes for Pimcore X?

brusch:
You're right, let's schedule this for 10. 
Thanks!

PR #7931, opened 2021-01-19
Use const `DOMAIN_ADMIN` instead of magic string `admin` in `Pimcore\Model\Translation`

jdreesen:


dvesh3:
follow up to https://github.com/pimcore/pimcore/pull/7800

dvesh3:
@jdreesen good catch! thank you 😉

PR #7943, opened 2021-01-19
Upgrade to scheb/2fa

### LINKED ISSUE #7845: Migrate to scheb/2fa
The latest version of this bundle (which is `v5`) has been moved to https://github.com/scheb/2fa 
Upgrade guide is available here: https://github.com/scheb/2fa/blob/5.x/UPGRADE.md

dvesh3:
Resolves #7845


composer.json
```diff
@@ -92,7 +92,9 @@
     "laminas/laminas-code": "^3.3.1",
     "laminas/laminas-paginator": "^2.7",
     "laminas/laminas-servicemanager": "^3.2",
-    "scheb/two-factor-bundle": "^3.26",
+    "scheb/2fa-bundle": "^5.3",
+    "scheb/2fa-google-authenticator": "^5.3",
+    "scheb/2fa-qr-code": "^5.3",
```
brusch, line 97: @dvesh3 do we really need `scheb/2fa-qr-code`? As we're generating the QR code using `endroid/qr-code` here: https://github.com/pimcore/pimcore/blob/4fd1f8d1889b0c5e5f315b6aa1a6dad55fb2e362/bundles/AdminBundle/Controller/Admin/UserController.php#L821

bundles/AdminBundle/Security/User/User.php
```diff
@@ -148,7 +148,7 @@ public function getGoogleAuthenticatorUsername(): string
      *
      * @return string
      */
-    public function getGoogleAuthenticatorSecret(): string
+    public function getGoogleAuthenticatorSecret(): ?string
```
jdreesen, line 151: As far as I can see, this method never returns `null`, so why this change?

composer.json
dvesh3: Good catch! reverted https://github.com/pimcore/pimcore/commit/69219ce19887a8f77746c59b7eaca8038ac7056f

bundles/AdminBundle/Security/User/User.php
dvesh3: Signature changed in the interface https://github.com/scheb/2fa/blob/5.x/src/google-authenticator/Model/Google/TwoFactorInterface.php#L23. Returning null now instead of empty string https://github.com/pimcore/pimcore/commit/69219ce19887a8f77746c59b7eaca8038ac7056f

PR #7953, opened 2021-01-20
[Assets] Show asset filesizes in UI when using stream wrappers + [WebDAV] Bugfix uploading new files + show assets without files

BlackbitNeueMedien:
In https://github.com/pimcore/pimcore/blob/de3119d12fd5a003db20dc896e757c7d7342664d/models/Asset/Service.php#L217
the file name contains an additional slash between `PIMCORE_ASSET_DIRECTORY` and the asset's full path. While this is no problem with local storage it becomes a problem with stream wrappers. Additionally there is no need to build the path here as `Asset::getFileSystemPath()` does it correctly anyway:
https://github.com/pimcore/pimcore/blob/de3119d12fd5a003db20dc896e757c7d7342664d/models/Asset.php#L466-L469
This is for Pimcore backend (sorry for mixing this in this WebDAV PR):
Before:
Bildschirmfoto 2021-01-20 um 17 12 16

After:
Bildschirmfoto 2021-01-21 um 08 36 54

The other important change in this PR is that in Pimcore backend you also see assets whose files do not exist anymore. To have the same behavior in WebDAV we need to suppress errors in https://github.com/pimcore/pimcore/blob/de3119d12fd5a003db20dc896e757c7d7342664d/models/Asset/WebDAV/File.php#L181
Compare this to https://github.com/pimcore/pimcore/blob/de3119d12fd5a003db20dc896e757c7d7342664d/models/Asset/Service.php#L218 which gets used when listing assets in Pimcore backend.

And a 3rd thing added: When I use Cyberduck as WebDAV client uploading new files did not work at first. The reason was that the stream was not at the beginning in https://github.com/pimcore/pimcore/blob/de3119d12fd5a003db20dc896e757c7d7342664d/models/Asset/WebDAV/Folder.php#L121-L122
Because other than the PHPdoc previously said `$data` is a stream here - at least when this method gets called from https://github.com/sabre-io/dav/blob/4258420f15425a5f128fe5cad454e00ab4a68ae5/lib/DAV/CorePlugin.php#L504
Beware that this does not happen when using the WebDAV browser interface because this sends POST requests and not PUT.

The other things are just minor fixes.

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -351,7 +351,7 @@ class Asset extends Element\AbstractElement
             } else {
                 $mimeType = Mime::detect($data['sourcePath'], $data['filename']);
                 if (is_file($data['sourcePath'])) {
-                    $data['stream'] = fopen($data['sourcePath'], 'r', false, File::getContext());
+                    $data['stream'] = fopen($data['sourcePath'], 'rb', false, File::getContext());
                 }
 
                 unset($data['sourcePath']);
@@ -708,7 +708,7 @@ class Asset extends Element\AbstractElement
                         unlink($destinationPath);
                     }
 
-                    $dest = fopen($destinationPath, 'w', false, File::getContext());
+                    $dest = fopen($destinationPath, 'wb', false, File::getContext());
                     if ($dest) {
                         stream_copy_to_stream($src, $dest);
                         if (!fclose($dest)) {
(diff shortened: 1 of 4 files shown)


brusch:
Nice! Thanks a lot!

PR #7969, opened 2021-01-22
Update AbstractData.php

mzulis:
Objectbricks: add localized field support in "get", "set" functions

## Additional info  
Solutions taken from Fieldcollection https://github.com/pimcore/pimcore/blob/master/models/DataObject/Fieldcollection/Data/AbstractData.php#L152


weisswurstkanone:
@mzulis thanks a lot!

PR #7973, opened 2021-01-22
check hardlink realpath with site

zerlpaMMT:
get correct path of documents from multiple sites in pimcore


brusch:
Closed in favor of #7996

PR #7975, opened 2021-01-23
Added Locking to prevent generating same thumbnail in concurent requests

alesak:
Added Locking to prevent wasting of CPU resources on generating same Image Thumbnail in parallel by my multiple independent requests


brusch:
Nice thanks a lot!

PR #7981, opened 2021-01-25
[Versions] Show current database version for unpublished elements

BlackbitNeueMedien:
When a data object is unpublished there currently is no way to recognize which version in the versions tab currently is saved in the database. This PR changes this with a new icon in the `Published` column:

Before:
Bildschirmfoto 2021-01-25 um 13 44 11

After:
Bildschirmfoto 2021-01-25 um 13 47 05

If the object is published, everything stays the same.


weisswurstkanone:
same for documents ?

BlackbitNeueMedien:
Yep, now also for documents:
Bildschirmfoto 2021-01-25 um 14 54 30

weisswurstkanone:
Nice! Thanks a lot!

PR #7988, opened 2021-01-26
[Ecommerce][Productindex][Elasticsearch] do not close index on index creation

markus-moser:
AWS hosted Elasticsearch version 6.x does not support the index close operation therefore index creation currently fails.

Therefore this PR directly adds the synonyms to the create index call. This way everything works correctly and the only drawback is that you can't use the `bin/console ecommerce:indexservice:elasticsearch-sync update-synonyms` command which isn't a real problem as you can still use the `bin/console ecommerce:indexservice:elasticsearch-sync reindex` command which is the better choice on live environments anyways.


fashxp:
thx

PR #8023, opened 2021-01-29
Fix CSRF Attack warning when using wrong custom login screen

### LINKED ISSUE #8016: Incorrect value in system settings prevents user from login
# Bug Report
Hi, 
I found a bug or definitely unexpected behavior in system settings. 

When an admin user sets an invalid URL string into the **login_screen_custom_image** field in **system settings**. after logout, another login to administration is unsuccessful and ends with an error. The error prevents users from login.

```
"code":400,
"message":"Detected CSRF Attack! Do not do evil things with pimcore ... ;-)"
```

This problem is hard to debug because, after this small issue, you are no longer able to log into the application.

### Expected behavior
* I would expect that field would be validated.
* Such issues should not prevent users from login.

### Actual behavior
It prevents users from logging in and fixing system settings.

### Steps to reproduce
* Update system settings > login screen custom URL with value that is not corect url. 
* eg.: "#FF0000asdasdasd" 
* save settings
* logout
* refresh browser
* try to login
* error is thrown

### Code example
```php
pimcore_admin:
    branding:
        color_login_screen: '#FF0000'
        color_admin_interface: ''
        login_screen_custom_image: '#FF0000asdasdasd'
        login_screen_invert_colors: false
```

dvesh3:
Resolves #8016

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/Login/layout.html.php
+++ b/bundles/AdminBundle/Resources/views/Admin/Login/layout.html.php
@@ -23,7 +23,8 @@ $config = $this->config;
 <body class="pimcore_version_6 <?= $config['branding']['login_screen_invert_colors'] ? 'inverted' : '' ?>">
 
 <?php
-    if (empty($backgroundImageUrl = $config['branding']['login_screen_custom_image'])) {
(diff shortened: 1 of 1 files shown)


brusch:
@dvesh3 is this also relevant for `master` (Twig templates)?

dvesh3:
> @dvesh3 is this also relevant for `master` (Twig templates)?

@brusch yes, please see https://github.com/pimcore/pimcore/commit/417e3a42b6035371663d397ce29a63f4fdc63e7b

PR #8026, opened 2021-01-31
Use searchSring instead of filter in website translation activation

solverat:
After hitting the `pimcore.globalmanager.get('translationwebsitemanager').activate` method,  the request will raise a json exception since the wrong parameter `filter` gets passed. I just adapted the logic from the admin translation class: 

https://github.com/solverat/pimcore/blob/2f700bd6beb4ce4c3c7844369e20a8a0dd272ddc/bundles/AdminBundle/Resources/public/js/pimcore/settings/translation/admin.js#L32


weisswurstkanone (approved):

brusch:
Thanks a lot!

PR #8042, opened 2021-02-01
[CSV Export] restrict AnyGetter to Pimcore Elements (Asset/Document/Object) and check user permission 

### LINKED ISSUE #2965: [CSV Export] restrict AnyGetter to Pimcore Elements (Asset/Document/Object) and check user permission

weisswurstkanone:
resolves #2965

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/startup.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/startup.js
@@ -164,11 +164,13 @@ Ext.onReady(function () {
     Ext.Ajax.on('requestexception', function (conn, response, options) {
         console.log("xhr request failed");
 
-        var jsonData = null;
-        try {
-            jsonData = Ext.decode(response.responseJson);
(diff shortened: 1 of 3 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/startup.js
```diff
@@ -166,7 +166,7 @@ Ext.onReady(function () {
 
         var jsonData = null;
         try {
-            jsonData = Ext.decode(response.responseText);
+            jsonData = Ext.decode(response.responseJson);
```
brusch, line 169: @weisswurstkanone What if the response isn't JSON? Maybe check for both properties here? 
Probably related to this here, right: https://forum.sencha.com/forum/showthread.php?471298-Undocumented-breaking-change-responseText-undefined ?

brusch (changes requested):

brusch (approved):

PR #8050, opened 2021-02-01
Fixes #8049 Add subitems to ValidationException rethrow

### LINKED ISSUE #8049: Pimcore\Model\DataObject\Concrete::update when a ValidationException is re-thrown it does not have the subitems from the original exception
# Bug Report

### Expected behavior
When a validation exception is thrown the exception should contain the sub items for each of the fields that cause the exception. (For example mandatory fields set as empty)
### Actual behavior
When the exception happens the subitems from the validation exception are ignored.
### Steps to reproduce
For example the method Pimcore\Model\DataObject\ClassDefinition\Data\Localizedfields::checkValidity throws a ValidationException with the subitems for each field, when this exception is catch in Pimcore\Model\DataObject\Concrete::update, a new exception is created instead and ignoring the sub items from the original one.

strentz:
Resolves #8049

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -141,7 +141,9 @@ class Concrete extends AbstractObject implements LazyLoadedFieldsInterface
                                     throw $e;
                                 }
                                 $exceptionClass = get_class($e);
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
thx

PR #8058, opened 2021-02-02
[Docs] Clearer instrucutions how to set up custom report adapters

BlackbitNeueMedien:
Changes:
* There is no need that report adapter class needs to be in namespace `Pimcore\Model\Tool\CustomReport\Adapter`
* Description of PHP Adapter class was duplicated resulting in the impression that one needs to implement 2 classes
* the list items about the adapter factory were a bit misleading because the step `Add your Adapter Factory to the configuration` seemed to be only necessary for "complex adapter classes" but in fact it is for all adapters, regardless if they use an own factory or `DefaultCustomReportAdapterFactory`


brusch:
Thanks a lot! 🙏

PR #8061, opened 2021-02-03
Fix JSON parse error, when clicking on edit button of a brick

aweichler:
There was an JSON parse error, when I clicked on the edit button of a brick with the following code:
```php
    public function getEditableDialogBoxConfiguration(Document\Editable $area, ?Info $info): EditableDialogBoxConfiguration
    {

        $config->setItems([
            [
                'type' => 'select',
                'name' => 'test',
                'config' => [
                    'store' => [
                        ["image 1", '<img src="/bundles/pimcoreadmin/img/flat-color-icons/idea.svg">'],
                    ],
                ],
            ]
        ]);

        return $config;
    }
```
This pull request fixes the error.


dvesh3:
thanks!

PR #8066, opened 2021-02-03
[Notifications] Prevent exception when trying to open deleted notification

BlackbitNeueMedien:
When a user receives a notification he can open the notification with one of the icons in the appearing modal in the bottom-right corner. If the notification got deleted in the meantime an exception gots thrown. This PR fixes this so that he gets the usual message "Element not found" (as when trying top open an object which got deleted in the meantime).


brusch:
Thank you! 👍
Have a nice weekend!

PR #8076, opened 2021-02-03
[Static Routes] Do not merge request params on static-route generate

### LINKED ISSUE #3394: Do not pass parameters by default in Staticrouter-Router
# Bug Report
The Symfony router interface (generate) and the StaticRouter interface (with reset parameter) are not compatible. I think thus, when Pimcore was upgraded to V5, the followin line in the static router has been added:

 Pimcore\Routing\Staticroute\Router.php
```php
$reset  = isset($parameters['reset']) ? (bool)$parameters['reset'] : false;
```
In my opinion this is very dangerous, as arbitrary parameters that are added by the user to the URL are passed through, without the awareness of the developer.
  
### Expected behavior 
Never automatically append parameters that are not explicitly defined to a route!

### Actual behavior  
Arbitrary parameters are added.
  
### Steps to reproduce  
https://www.my-project.com/?_prefix=insecure-parameter....

```php
router = \Pimcore::getContainer()->get('router');
//route should be a static route.
$url = $this->router->generate('app_shop_product_overview', [
          'slug' => $object->getSlug($locale) ? : File::getValidFilename($object->getName($locale)),
]);
 ```

# Feature Request 
Can we change the line so that reset by default is always true in Pimcore 6?
Optionally we could define a global parameter (parameters.yml) where it is possible to disable that behavior for a specific project.

### PR DESCRIPTION
**Pull request by dvesh3**
Resolves #3394

### DIFF
--- a/models/Staticroute.php
+++ b/models/Staticroute.php
@@ -438,33 +438,6 @@ class Staticroute extends AbstractModel
      */
     public function assemble(array $urlOptions = [], $encode = true)
     {
-        // get request parameters
-        $blockedRequestParams = ['controller', 'document'];
-
-        // allow blocked params if we use it as variables
-        $variables = explode(',', $this->getVariables());
-        foreach ($variables as $name) {
-            $pos = array_search($name, $blockedRequestParams);
-            if ($pos !== false) {
-                unset($blockedRequestParams[$pos]);
-            }
-        }
-
-        $requestParameters = [];
-
-        // merge route params from static routes here
-        $request = \Pimcore::getContainer()->get('request_stack')->getCurrentRequest();
-        if (null !== $request && $request->attributes->get('_route_params')) {
-            $requestParameters = array_merge($requestParameters, $request->attributes->get('_route_params'));
-        }
-
-        // remove blocked parameters from request
-        foreach ($blockedRequestParams as $key) {
-            if (array_key_exists($key, $requestParameters)) {
-                unset($requestParameters[$key]);
-            }
-        }
-
         $defaultValues = $this->getDefaultsArray();
 
         // apply values (controller, ... ) from previous match if applicable (only when )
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: models/Staticroute.php [hunk 1]
```diff
@@ -451,32 +450,26 @@ public function assemble(array $urlOptions = [], $reset = false, $encode = true)
...
-        } else {
-            $requestParameters = \Pimcore::getContainer()->get('pimcore.routing.router.request_context')->getParameters();
+        $requestParameters = [];
 
-            // merge route params from static routes here
-            $request = \Pimcore::getContainer()->get('request_stack')->getCurrentRequest();
-            if (null !== $request && $request->attributes->get('_route_params')) {
-                $requestParameters = array_merge($requestParameters, $request->attributes->get('_route_params'));
-            }
+        // merge route params from static routes here
```
brusch, line 455: @dvesh3 This entire code block can go I guess, it was used in the `else` clause and therefore only used if `$reset = false`.

brusch (changes requested):

models/Staticroute.php
dvesh3: you are right...kicked it

brusch: Thanks 😊

dpfaffenbauer:
I don't know why and how, but this breaks route like:

```
2 => [
        "id" => 2,
        "name" => "coreshop_category_list",
        "pattern" => "/(\\w+)\\/shop\\/(.*?)\\~c([0-9]+)\$/",
        "reverse" => "/%_locale/shop/%name~c%category",
        "controller" => "CoreShop\\Bundle\\FrontendBundle\\Controller\\CategoryController:indexAction",
        "variables" => "_locale,name,category",
        "defaults" => NULL,
        "siteId" => NULL,
        "methods" => [

        ],
        "priority" => 2,
        "creationDate" => 1611743537,
        "modificationDate" => 1611743537
    ],
```

variable _locale is now set to "_locale" instead of the actual locale

dvesh3:
@dpfaffenbauer thanks for reporting 👍 please see #8103

PR #8078, opened 2021-02-03
[CoreBundle] Invalid definition for service GeoIp : lint:container

ogmdev:
On my environment Pimcore 6.8.7 (PHP 7.4) I have this error when  I run Symfony command lint:container 

````bash
/www/current $ php bin/console lint:container

In CheckTypeDeclarationsPass.php line 302:
                                                                                                                                                                 
  Invalid definition for service "Pimcore\Targeting\DataProvider\GeoIp": argument 1 of "GeoIp2\Database\Reader::__construct()" accepts "string", "NULL" passed.  
 ````

This  PR fixes this error, setting an empty string instead of null value.


brusch:
Thanks a lot!

PR #8116, opened 2021-02-08
[DataObject] Classificationstore + QuantityValue not properly restored

### LINKED ISSUE #8114: [DataObject] Classificationstore + QuantityValue not properly restored
Regression of https://github.com/pimcore/pimcore/pull/6965

weisswurstkanone:
resolves #8114

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20210211152319.php
+++ b/bundles/CoreBundle/Migrations/Version20210211152319.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+use Pimcore\Model\DataObject\ClassDefinition;
+use Pimcore\Model\DataObject\Objectbrick\Definition;
+
+class Version20210211152319 extends AbstractPimcoreMigration
+{
+    public function doesSqlMigrations(): bool
+    {
+        return false;
+    }
+
+    /**
+     * @param Schema $schema
+     *
+     * @throws \Exception
+     */
+    public function up(Schema $schema)
+    {
+        $list = new ClassDefinition\Listing();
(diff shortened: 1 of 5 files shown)


dvesh3 (changes requested):
Data not saved for direct inputQuantity fields on object class as generated tables unit column is bigint and it should be varchar
image

line: https://github.com/pimcore/pimcore/blob/48102717faaa6e0283644636457c9cdb77961c25/models/DataObject/ClassDefinition/Data/QuantityValue.php#L95

https://github.com/pimcore/pimcore/blob/48102717faaa6e0283644636457c9cdb77961c25/models/DataObject/ClassDefinition/Data/QuantityValue.php#L95

dvesh3 (changes requested):
if InputQuantityValue field is inside Classificationstore, then on save it throws validation error..
image

weisswurstkanone:
@dvesh3 unable to reproduce the validation error

bundles/CoreBundle/Migrations/Version20210211152319.php
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @throws \Exception
+     */
+    public function up(Schema $schema)
+    {
+        $list = new ClassDefinition\Listing();
+        $list = $list->load();
+        foreach ($list as $class) {
+            $this->writeMessage(sprintf('Saving php files for class: %s', $class->getName()));
+            $class->generateClassFiles(false);
```
dvesh3, line 28: generating classes is not enough.....we need to up db tables as well for column type update

dvesh3 (approved):

dvesh3 (changes requested):
Update migration

bundles/CoreBundle/Migrations/Version20210211152319.php
dvesh3: Even calling `$class->save()` in migration has no impact on existing classes.

dvesh3 (approved):

PR #8120, opened 2021-02-08
Consider existing parameters in DataObject Linkgenerators

jremmurd:
# Bugfix

Consider additional parameters in object linkgenerators.

## Use Case:
The link generator produces a link with querystring zB www.example.com?test1=123

In the Pimcore backend additional parameters get configured zB test2=123

The currently resulting link is www.example.com?test1=123?test2=123

This PR changes the behaviour to the following result: www.example.com?test1=123&test2=123


dvesh3:
thank you!

PR #8134, opened 2021-02-10
Custom Login screen - Fix and support external image URL

### LINKED ISSUE #8129: Login-Background: Custom Image not displaying anymore
# Bug Report
When defining a custom login background image path in the system settings, which isn't hosted on the same domain, the image isn't displayed anymore. [This commit](https://github.com/pimcore/pimcore/commit/417e3a42b6035371663d397ce29a63f4fdc63e7b) is causing this.

### Expected behavior
Before it was possible to host the background image on another host. This had the benefit, that you wouldn't need to always store the same images on all Pimcore instances.

### Actual behavior
Now the code checks if the image exists on the filesystem, which of course is `false` when the image lies on another system.

### Steps to reproduce

- Define an image of a different host in the system settings
- Logout
- Default Pimcore background images are display instead of your own custom image

dvesh3:
Resolves #8129

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/Login/layout.html.php
+++ b/bundles/AdminBundle/Resources/views/Admin/Login/layout.html.php
@@ -26,7 +26,7 @@ $config = $this->config;
     $customImage = $config['branding']['login_screen_custom_image'];
     //https://github.com/pimcore/pimcore/issues/8016
     //https://github.com/pimcore/pimcore/issues/8129
-    if (preg_match('@^https?://@', $customImage) && file_get_contents($customImage)) {
+    if (preg_match('@^https?://@', $customImage)) {
         $backgroundImageUrl = $customImage;
     } elseif (file_exists(PIMCORE_WEB_ROOT . $customImage)) {
         $backgroundImageUrl = $customImage;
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/views/Admin/Login/layout.html.php
```diff
@@ -23,8 +23,14 @@
 <body class="pimcore_version_6 <?= $config['branding']['login_screen_invert_colors'] ? 'inverted' : '' ?>">
 
 <?php
-    $backgroundImageUrl = file_exists(PIMCORE_WEB_ROOT . $config['branding']['login_screen_custom_image']) ? $config['branding']['login_screen_custom_image'] : '';
-    if (empty($backgroundImageUrl)) {
+    $customImage = $config['branding']['login_screen_custom_image'];
+    //https://github.com/pimcore/pimcore/issues/8016
+    //https://github.com/pimcore/pimcore/issues/8129
+    if (preg_match('@^https?://@', $customImage) && file_get_contents($customImage)) {
```
NiklasBr, line 29:
```suggestion
    if (preg_match('@^https?://@', $customImage) && is_readable($customImage)) {
```

brusch, line 29: I woudn't do a `file_get_contents()` here. 
I'd say it's up to the administrator to ensure the image exists.

brusch (changes requested):

bundles/AdminBundle/Resources/views/Admin/Login/layout.html.php
dvesh3: reverted!

PR #8135, opened 2021-02-10
Custom Login screen - Fix and support external image URL for Pimcore 10

dvesh3:
Follow up to https://github.com/pimcore/pimcore/commit/417e3a42b6035371663d397ce29a63f4fdc63e7b
Related fix for #8129

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/Login/layout.html.twig
+++ b/bundles/AdminBundle/Resources/views/Admin/Login/layout.html.twig
@@ -19,7 +19,7 @@
         {% set customImage = config['branding']['login_screen_custom_image'] %}
         {# https://github.com/pimcore/pimcore/issues/8016 #}
         {# https://github.com/pimcore/pimcore/issues/8129 #}
-        {% if customImage matches '@^https?://@' and pimcore_file_get_contents(customImage) %}
+        {% if customImage matches '@^https?://@' %}
             {% set backgroundImageUrl = customImage %}
         {% elseif pimcore_file_exists(constant('PIMCORE_WEB_ROOT') ~ customImage) %}
             {% set backgroundImageUrl = customImage %}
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Resources/views/Admin/Login/layout.html.twig
```diff
@@ -16,12 +16,16 @@
...
-        {% if pimcore_file_exists(constant('PIMCORE_WEB_ROOT') ~ config['branding']['login_screen_custom_image']) %}
-            {% set backgroundImageUrl = config['branding']['login_screen_custom_image'] %}
-        {% endif %}
-        {% if backgroundImageUrl is empty %}
-            {% set defaultImages = ['pimconaut-ecommerce.svg', 'pimconaut-world.svg', 'pimconaut-engineer.svg', 'pimconaut-moon.svg', 'pimconaut-rocket.svg'] %}
-            {% set backgroundImageUrl = '/bundles/pimcoreadmin/img/login/' ~ random(defaultImages) %}
+        {% set customImage = config['branding']['login_screen_custom_image'] %}
+        {# https://github.com/pimcore/pimcore/issues/8016 #}
+        {# https://github.com/pimcore/pimcore/issues/8129 #}
+        {% if customImage matches '@^https?://@' and pimcore_file_get_contents(customImage) %}
```
brusch, line 22: Same here, wouldn't fire a HTTP request here. 
If the target server is not reachable this would completely block the Pimcore login.

brusch (changes requested):

bundles/AdminBundle/Resources/views/Admin/Login/layout.html.twig
dvesh3: reverted!

PR #8139, opened 2021-02-10
added async option to headscript

alen125:

Adds additional option to use "async" in headscript


brusch:
Nice, thanks a lot! 👍

PR #8141, opened 2021-02-11
 Fix the problem that not all fields are shown in "Visible Fields" at…

aweichler:
This pull request fixes an issue in the many to many object relation, where not all fields are shown in "Visible Fields".
If there are two input fields in class Product where the second Input field is after a block field, it is not shown in the "Visible Fields" of the many to many object relation for this class.
In the attached zip File i created an example.
In the ProductGroup class you can see the issue.
The field artnr is not shown when you click at products field.
[classes.zip](https://github.com/pimcore/pimcore/files/5963676/classes.zip)


weisswurstkanone:
thanks a lot!

jacobsyentl:
@weisswurstkanone This issue is also present in Pimcore X. Should I make a new ticket for this?

brusch:
@jacobsyentl yes, please create a new ticket for that. Thanks!

PR #8146, opened 2021-02-11
[DataObjects] Quantity Value: Fix adding new units

podarcis:
(Fix is based on master as it doesn't occur in current 6.9).

This fixes adding and editing new units by checking for an undefined index 'baseunit'. There seems to be a changed behavior (no property 'baseunit' in payload) how newly persisted entries are updated.

How to reproduce:
1) Click "new unit" and enter ID "g" into modal.
2) Edit abbreviation to "g".
3) Lose focus, so entry is saved.
4) Exception is shown instead of update.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/QuantityValueController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/QuantityValueController.php
@@ -110,7 +110,7 @@ class QuantityValueController extends AdminController
                 $data = json_decode($request->get('data'), true);
                 $unit = Unit::getById($data['id']);
                 if (!empty($unit)) {
-                    if ($data['baseunit'] === -1) {
+                    if (isset($data['baseunit']) && $data['baseunit'] === -1) {
                         $data['baseunit'] = null;
                     }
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Helloo @podarcis , did you find a reason why this doesn't happen on 6.9?

podarcis:
@weisswurstkanone I just verified that the sent payload for `unit-proxy?xaction=update` is `data: {"id":"g","abbreviation":"g"}` for both versions. And the array key `baseunit` already wasn't set in 6.9 as well, so there's also an "undefined index" notice.

However I suspect that the PHP notice are handled differently in master and are caught as exception.

(Both installations dev-master and current-6.9 are freshly installed and Debug-Mode=On and are using same PHP 7.4).

podarcis:
What do you mean? This PR fixes the issue that the array is accessed with an undefined index. A similar check is already in place some lines below in `xaction=create`.

If you mean investigating the differing behavior regarding error handling: Can't promise, but I'll take a look.

weisswurstkanone:
Thanks a lot!

PR #8181, opened 2021-02-15
Proper method name and fewer if-checks in Imagick adapter

NiklasBr:
- `setcolorspace` -> `setColorspace` method name
- Fewer checks for `!$this->isPreserveColor()`

### DIFF
--- a/lib/Image/Adapter/Imagick.php
+++ b/lib/Image/Adapter/Imagick.php
@@ -98,20 +98,6 @@ class Imagick extends Adapter
             $i = new \Imagick();
             $this->imagePath = $imagePath;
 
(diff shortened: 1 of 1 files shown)


NiklasBr:
I didn't understand the build error and need to check up on it

dvesh3:
@NiklasBr `$this->setColorspaceToRGB();` requires $this->resource to be set, so it should be called after the line https://github.com/pimcore/pimcore/blob/527cf3700c85fd53ad7e58d41f192469c57c4229/lib/Image/Adapter/Imagick.php#L130 which set's the resource.

brusch:
Thank you!

PR #8187, opened 2021-02-16
DocumentDataExtractor editable check

zerlpaMMT:
$targetDocument->getEditable() can return null and function call on null is not good.


dvesh3:
@zerlpaMMT thanks!

PR #8189, opened 2021-02-16
Fix Wrong language used if key equals translation in shared translations

### LINKED ISSUE #8117: Wrong language used if key equals translation in shared translations
# Bug Report
### Expected behavior
If the translation key "Artikel suchen" has an english translation "search article" and german translation "Artikel suchen" it should read "Artikel suchen" on a german website.

### Actual behavior
It does show "search article" instead.

### Steps to reproduce
Setup a pimcore instance with languages german and english. Use english as fallback language for german.
Setup a website using the german language.
Use a translation key in twig like this:
```twig
{{ 'Artikel suchen'|trans }}
```
Fill in "search article" as english translation in the shared translations and "Artikel suchen" as german translation.

This issue has been discussed in #7184 before. Reproducable in Pimcore 6.8.8.

dvesh3:
Resolves #8117

### DIFF
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -284,7 +284,7 @@ class Translator implements LegacyTranslatorInterface, TranslatorInterface, Tran
             $normalizedId = mb_strtolower($id);
         }
 
-        if (isset($parameters['%count%'])) {
+        if (isset($parameters['%count%']) && $translated) {
             $normalizedId = $id = $translated;
         }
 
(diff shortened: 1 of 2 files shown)


lib/Translation/Translator.php
```diff
@@ -366,7 +361,7 @@ protected function checkForEmptyTranslation($id, $translated, $parameters, $doma
             }
         }
 
-        return $translated;
+        return !empty($translated) ? $translated : $id;
```
matthewfedak, line 364: Was this change done by accident? It is now impossible to know if something was translated or not. For example:

Let's say I call the method checkForEmptyTranslation() with string $id "taxi" and target language "tr" (Turkish). 

If I get back "taksi" I'll know the word was translated because it is significantly different. However if no translation exists I will still get back "taxi" and I'll have no idea if the translation existed or not.

PR #8200, opened 2021-02-17
[DataObject] Remove columnType & queryColumnType property from definition export

### LINKED ISSUE #8158: [DataObject] ColumnType change for existing classes
# Improvement
At the moment, if there's a change in Object Datatype (query) column type then it is quite hard to update existing class tables as they are always rebuild based on generated definition file.

Solution: Consider column type from Datatype class on every class save.

Also remove temp solution: https://github.com/pimcore/pimcore/pull/8116/files#diff-d2564b1d2701cc3ab300240f945b53cb21a2466232ad24ad2e5d75f10f5789faR679-R688

dvesh3:
Resolves #8158

### DIFF
--- a/models/DataObject/ClassDefinition/CustomLayout.php
+++ b/models/DataObject/ClassDefinition/CustomLayout.php
@@ -268,7 +268,7 @@ class CustomLayout extends Model\AbstractModel
      */
     public static function cleanupForExport(&$data)
     {
-        if (method_exists($data, 'resolveBlockedVars')) {
+        if ($data instanceof DataObject\ClassDefinition\Data\VarExporterInterface) {
             $blockedVars = $data->resolveBlockedVars();
             foreach ($blockedVars as $blockedVar) {
                 if (isset($data->{$blockedVar})) {
--- a/models/DataObject/ClassDefinition/CustomLayout.php
+++ b/models/DataObject/ClassDefinition/CustomLayout.php
@@ -268,8 +268,17 @@ class CustomLayout extends Model\AbstractModel
      */
     public static function cleanupForExport(&$data)
     {
-        if (isset($data->fieldDefinitionsCache)) {
-            unset($data->fieldDefinitionsCache);
+        if (method_exists($data, 'resolveBlockedVars')) {
+            $blockedVars = $data->resolveBlockedVars();
+            foreach ($blockedVars as $blockedVar) {
+                if (isset($data->{$blockedVar})) {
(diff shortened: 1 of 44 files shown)


models/DataObject/ClassDefinition/CustomLayout.php
```diff
@@ -268,8 +268,17 @@ public function getDefinitionFile()
      */
     public static function cleanupForExport(&$data)
     {
-        if (isset($data->fieldDefinitionsCache)) {
-            unset($data->fieldDefinitionsCache);
+        if (method_exists($data, 'resolveBlockedVars')) {
```
brusch, line 271: Maybe we should go for an interface here? Or we just add it the the abstract data-type?

brusch (changes requested):

models/DataObject/ClassDefinition/CustomLayout.php
dpfaffenbauer: Yes, definitely use an interface here

dvesh3: Added VarExportInterface for Data-types. please see https://github.com/pimcore/pimcore/pull/8200/commits/65a78b0aa1242b32cc94d14a2669b572c79dc585

dpfaffenbauer:
to be honest: I really don't like this solution. I would much rather have a more robust and future proof solution for creating class-definitions. something that might also work in the future with YML definitions or XML definitions. I think it would be much better to get rid of VarExporter and use maybe Symfony Serializer or a custom normalization and serialization process for that. 

Main reason: The current solution depends on PHP for the class files, which works, but doesn't work with DependencyInjection for custom tags. If we get rid of VarExporter, we could also work on DI for tags and future proof class-definitions even better.

dvesh3:
> to be honest: I really don't like this solution. I would much rather have a more robust and future proof solution for creating class-definitions. something that might also work in the future with YML definitions or XML definitions. I think it would be much better to get rid of VarExporter and use maybe Symfony Serializer or a custom normalization and serialization process for that.
> 
> Main reason: The current solution depends on PHP for the class files, which works, but doesn't work with DependencyInjection for custom tags. If we get rid of VarExporter, we could also work on DI for tags and future proof class-definitions even better.

@dpfaffenbauer well, this is a different topic. It requires complete refactoring and for sure a BC break, as it changes the definition file persistence so please create a ticket for Pimcore 10.. thanks!

weisswurstkanone (approved):

PR #8205, opened 2021-02-18
Translation fallback kicks in on valid translation

das-peter:
Scenario:
Translation Key (EN): "Stock"
Translations:
* DE: Lager
* FR: Stock

A call with FR as locale to translator resolves this to "Lager"!
This is because `\Pimcore\Translation\Translator::checkForEmptyTranslation()` triggers the language fallback handler if the translation key and the translated value match.
I think the translator tries to do to much here - if locale catalogue has a value for the key, the value should be used and not necessarily further fiddled with.
Even checking if the value is non empty might be already to much post processing in terms of expected behavior - it means that you can't "hide" / "remove" an irrelevant translation in a language.

Obviously you could workaround by using a dedicated translation key - but insisting on that would likely collide with a loooot of real world usage (https://symfony.com/doc/current/translation.html#using-real-or-keyword-messages).

### DIFF
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -294,7 +294,7 @@ class Translator implements LegacyTranslatorInterface, TranslatorInterface, Tran
         } elseif ($normalizedId == $translated) {
             if ($this->getCatalogue($locale)->has($normalizedId, $domain)) {
                 $translated = $this->getCatalogue($locale)->get($normalizedId, $domain);
-                if ($normalizedId != $translated && $translated) {
+                if ($translated) {
                     return $translated;
                 }
             } elseif ($backend = $this->getBackendForDomain($domain)) {


brusch:
@das-peter Sounds like a bug, can you please rebase your PR onto the `6.8` branch? 
@dvesh3 has added some tests for translations recently, I think we should a test-case for this scenario as well. 
Thanks a lot!

dvesh3:
@das-peter this should be fixed by https://github.com/pimcore/pimcore/pull/8189 could you please confirm? thanks!

das-peter:
As far as I understand this still goes to the fallback handling if the key matches the translated value:
```
if ($normalizedId != $translated && $translated) {
    return $translated;
}
```

dvesh3:
In #8189 catalogue is not lazy initialized with key as value(if value doesn't exist for locale) and we look for fallback when translated value is empty(as oppose to before looking for fallback when `$key = $translated`).

@das-peter are you already on 6.8.10?

dvesh3:
@das-peter could you please confirm https://github.com/pimcore/pimcore/pull/8205#issuecomment-781208806? thanks!

dvesh3:
Closed due to inactivity. Please feel free to discuss further 😉

PR #8217, opened 2021-02-18
[BUGFIX] fix format() of Grid DateFormatter

storbahn:

Steps to reproduce
Using DateFormatter in DataObject Grid failed.

If `$theValue` was an `int` it would have been returned as such when no `$this->format` is not defined and resulted in an error otherwise because `date()` was passed a `Carbon` instance instead of an `int` timestamp. This fixes it.


brusch:
Thank you very much!

PR #8222, opened 2021-02-18
ramsey/uuid: ^4.0

brusch:


dpfaffenbauer:
maybe use this instead? https://symfony.com/doc/current/components/uid.html

brusch:
@dpfaffenbauer valid point, thanks for pointing out!

PR #8223, opened 2021-02-18
Use Symfony/Uuid instead of Ramsey/Uuid

brusch:

PR #8224, opened 2021-02-18
Use UUID namespace constant instead of magic string

jdreesen:
Followup of #8223 
See: https://github.com/pimcore/pimcore/pull/8223#discussion_r578539747


jdreesen:
Ooh no! It's a Symfony 5.3 feature and not available in 5.2 yet :(

PR #8230, opened 2021-02-19
Tool\Console Deprecations

### LINKED ISSUE #8136: Tool\Console Deprecations
Most of the stuff in `Tool\Console` is covered by Symfony's `Process` component, which should be used instead. 
In `6.9`we should change already everything to Symfony/Process.

dvesh3:
Resolves #8136

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -43,6 +43,7 @@ use Symfony\Component\HttpFoundation\ResponseHeaderBag;
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
 use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
 use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\Process\Process;
 use Symfony\Component\Routing\Annotation\Route;
 
 /**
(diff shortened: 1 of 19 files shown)


weisswurstkanone:
@dvesh3 please remove the exclusion from here when your PR gets merged: https://github.com/pimcore/pimcore/blob/8b18a7e7126948df2e77fb0b76a188c29dec1c78/phpstan-baseline.neon#L13

BlackbitNeueMedien:
With `Console\Tool::executeInBackground()` and `runPhpScriptInBackground()` it was tried to use `nice` to lower the process priority. Is it wanted that we lose this because I could imagine that cleaning some things up does not need to be executed with the same priority as serving real requests. Perhaps we could add this to Console\Tool::

BlackbitNeueMedien:
Any reaction about dropping automatic use of `nice` and `nohup`?

brusch:
@BlackbitNeueMedien I think it would make sense to add `nice` to certain commands, but not to all of them as it was. 
Maybe just for FFMPEG?

BlackbitNeueMedien:
In which case can a background task be as important as serving real requests? To me it was reasonable that every background process got low priority automatically.

brusch:
Would be "nice" 😊 to see this in Symfony, see symfony/symfony#35488
That wouldn't require an additional layer inbetween.

BlackbitNeueMedien:
Of course would be good if Symfony had this but they will never automatically set low priority because a Symfony Command does not know if a process is a background process or not.
But in Pimcore we had the layer to distinguish those two types of processes. I already can imagine customers complaining why their site is so slow while agencies say that there are some "background" task running which are treated with the same priority as real user requests.

brusch:
@BlackbitNeueMedien see #8329 - not perfect, but with this approach we can control whether a process is low priority or not.

PR #8242, opened 2021-02-22
[Application Logger] Stream log file object

BlackbitNeueMedien:
When the log file object is very big memory problems can arise with `file_get_contents()`. With this PR the log file gets streamed to prevent this.


brusch:
Nice, thanks a lot! 👍

PR #8246, opened 2021-02-23
[Thumbnails] Performance improvements for cleaning thumbnails

BlackbitNeueMedien:
This PR merges the 2 regular expressions and uses `SplFileInfo::getFilename()` for the check. I also moved the `isDir()` call after the regular expression check because file system calls can be quite slow - especially when using stream wrappers (we currently have > 100000 files in Google Cloud Storage).

And the cleanup of temporary files is delegated to `pimcore:thumbnails:clear` command in the background.

PS: I actually tried to enhance performance even more but as https://github.com/googleapis/google-cloud-php-storage does currently not support `glob` operations we really have to loop through every asset. It is especially bad that the iterator also returns files but I did not find a way to change this (tried RecursiveCallbackFilterIterator but this has to return `true` for all directories, otherwise recursion does not work).

### DIFF
--- a/models/Asset/Image/Thumbnail/Config/Dao.php
+++ b/models/Asset/Image/Thumbnail/Config/Dao.php
@@ -128,14 +128,12 @@ class Dao extends Model\Dao\PhpArrayTable
         $enabled = \Pimcore::getContainer()->getParameter('pimcore.config')['assets']['image']['thumbnails']['auto_clear_temp_files'];
         if ($enabled) {
             $arguments = [
-                Console::getPhpCli(),
-                PIMCORE_PROJECT_ROOT.'/bin/console',
                 'pimcore:thumbnails:clear',
                 '--type=image',
                 '--name='.$this->model->getName()
             ];
 
-            Console::execInBackground(implode(' ', $arguments));
+            Console::runPhpScriptInBackground(realpath(PIMCORE_PROJECT_ROOT.DIRECTORY_SEPARATOR.'bin'.DIRECTORY_SEPARATOR.'console'), implode(' ', $arguments));
         }
     }
 }
--- a/models/Asset/Image/Thumbnail/Config/Dao.php
+++ b/models/Asset/Image/Thumbnail/Config/Dao.php
@@ -132,7 +132,7 @@ class Dao extends Model\Dao\PhpArrayTable
                 '--name='.$this->model->getName()
             ];
 
-            Console::runPhpScriptInBackground(realpath(PIMCORE_PROJECT_ROOT.DIRECTORY_SEPARATOR.'bin'.DIRECTORY_SEPARATOR.'console'), implode(' ', $arguments));
+            Console::runPhpScriptInBackground(realpath(PIMCORE_PROJECT_ROOT.DIRECTORY_SEPARATOR.'bin'.DIRECTORY_SEPARATOR.'console'), $arguments);
         }
     }
 }
--- a/models/Asset/Image/Thumbnail/Config/Dao.php
+++ b/models/Asset/Image/Thumbnail/Config/Dao.php
@@ -98,7 +98,7 @@ class Dao extends Model\Dao\PhpArrayTable
             $thumbnailDefinitionAlreadyExisted = $this->db->getById($this->model->getName()) !== null;
             $this->db->insertOrUpdate($data, $this->model->getName());
 
-            if (!$thumbnailDefinitionAlreadyExisted) {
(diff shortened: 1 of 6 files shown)


brusch:
@BlackbitNeueMedien can you please check why the tests are failing? thanks!

models/Asset/Image/Thumbnail/Config/Dao.php
```diff
@@ -106,7 +112,16 @@ protected function autoClearTempFiles()
...
+            $arguments = [
+                Console::getPhpCli(),
+                PIMCORE_PROJECT_ROOT.'/bin/console',
+                'pimcore:thumbnails:clear',
+                '--type=image',
+                '--name='.$this->model->getName()
+            ];
+
+            $process = new Process($arguments);
+            $process->start();
```
brusch, line 124: This unfortunately doesn't work,  as the process doesn't get detached,  it is killed as soon as the main process ends, see also  #8371

brusch (changes requested):

brusch:
I guess tests are failing because cleanup is now done asynchronously.  Maybe we should also add a switch/argument to force synchronous cleanup.

BlackbitNeueMedien:
Ah ok, thx @brusch .
I reverted back to `Console::execInBackground()`.

Concerning forcing synchronous cleanup: Currently `Pimcore\Model\Asset\Image\Thumbnail\Config\Dao\autoClearTempFiles()` gets only executed when saving a thumbnail definition - and as it is `protected` you cannot execute it from outside. When someone needs synchronous cleanup he would not do it by saving a thumbnail definition but by calling `Asset::clearTempFiles()` directly.

models/Asset/Image/Thumbnail/Config/Dao.php
```diff
@@ -69,7 +71,7 @@ public function exists(string $name): bool
     /**
      * @throws \Exception
      */
-    public function save()
+    public function save($autoClearThumbnails = true)
```
brusch, line 74: I'm just thinking if `$autoClearThumbnails` is really a good idea, because it conflicts in a way with the Symfony config `auto_clear_temp_files`. So even if you call `save(true)` but `auto_clear_temp_files` is set to `false` it won't run the cleanup. 
So I'd really go for a `$forceClearTempFiles = false`, which does call `$this->model->clearTempFiles();` synchronously if set to `true`.

brusch (changes requested):

models/Asset/Image/Thumbnail/Config/Dao.php
BlackbitNeueMedien: You are right. Changed the code as you requested.

models/Asset/Image/Thumbnail/Config/Dao.php
```diff
@@ -89,24 +92,50 @@ public function save()
...
-            $this->model->clearTempFiles();
+            $arguments = [
+                Console::getPhpCli(),
+                PIMCORE_PROJECT_ROOT.'/bin/console',
+                'pimcore:thumbnails:clear',
+                '--type=image',
+                '--name='.$this->model->getName()
+            ];
+
+            Console::execInBackground(implode(' ', $arguments));
```
brusch, line 138: We should use `\Pimcore\Tool\Console::runPhpScriptInBackground()` instead, saves a bit of code and does proper escaping 😉

models/Asset/Image/Thumbnail/Config/Dao.php
```diff
@@ -89,24 +91,48 @@ public function save()
...
         $enabled = \Pimcore::getContainer()->getParameter('pimcore.config')['assets']['image']['thumbnails']['auto_clear_temp_files'];
         if ($enabled) {
-            $this->model->clearTempFiles();
+            $arguments = [
+                'pimcore:thumbnails:clear',
+                '--type=image',
+                '--name='.$this->model->getName()
+            ];
+
+            Console::runPhpScriptInBackground(realpath(PIMCORE_PROJECT_ROOT.DIRECTORY_SEPARATOR.'bin'.DIRECTORY_SEPARATOR.'console'), implode(' ', $arguments));
```
brusch, line 135:
```suggestion
            Console::runPhpScriptInBackground(realpath(PIMCORE_PROJECT_ROOT.DIRECTORY_SEPARATOR.'bin'.DIRECTORY_SEPARATOR.'console'), $arguments);
```

brusch (changes requested):

PR #8252, opened 2021-02-23
Fix getDefaultConfig of imagegallery.js overwriting previously set configuration

Brainshaker95:

The `getDefaultFieldConfig` function of `imagegallery.js` returns a fieldConfig object that doesn't take the previously set configuration in consideration. This is problematic since this config gets passed to the hotspotimage constructor in line 146.
E.g. When a "predefined data template" is present on the ImageGallery this template will not be passed to the hotspot image and so will not be taken into account inside of `hotspotimage.js`.

```js
// imagegallery.js
  var fieldConfig = this.getDefaultFieldConfig();
  var hotspotImage = new pimcore.object.tags.hotspotimage(itemData, fieldConfig, this.hotspotConfig);
 ```
 ```js
 // hotspotimage.js
  if (fieldConfig.predefinedDataTemplates) {
    try {
      this.predefinedDataTemplates = Ext.decode(fieldConfig.predefinedDataTemplates);
    } catch (e) {
      console.log(e);
    }
  }
 ```
 
 In this example the `predefinedDataTemplates` option of `hotspotimage` will be null, since the `getDefaultConfig` function doesn't return the full fieldConfig object. This is probably not intended since the opened edit window of the Hotspotimage will not contain the definied config.
 
 This PR aims to fix this issue by assigning all previously set configuration options to the fieldConfig object inside of `getDefaultFieldConfig` .

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/imagegallery.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/imagegallery.js
@@ -107,7 +107,6 @@ pimcore.object.tags.imageGallery = Class.create(pimcore.object.tags.abstract, {
         var fieldConfig = Object.assign({}, this.fieldConfig, {
             width: itemWidth,
             height: itemHeight,
-            uploadPath: this.fieldConfig.uploadPath
         });
     
         return fieldConfig;
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/imagegallery.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/imagegallery.js
@@ -103,13 +103,13 @@ pimcore.object.tags.imageGallery = Class.create(pimcore.object.tags.abstract, {
     getDefaultFieldConfig: function () {
         var itemWidth = this.fieldConfig.width ? this.fieldConfig.width : 150;
         var itemHeight = this.fieldConfig.height ? this.fieldConfig.height : 150;
-
-        var fieldConfig = {
+    
+        var fieldConfig = Object.assign({}, this.fieldConfig, {
             width: itemWidth,
             height: itemHeight,
-            uploadPath: this.fieldConfig.uploadPath,
-        };
-
+            uploadPath: this.fieldConfig.uploadPath
+        });
+    
         return fieldConfig;
     },
 

### CODE REVIEW

File: bundles/AdminBundle/Resources/public/js/pimcore/object/tags/imagegallery.js [hunk 1]
```diff
@@ -103,13 +103,13 @@ pimcore.object.tags.imageGallery = Class.create(pimcore.object.tags.abstract, {
...
-
-        var fieldConfig = {
+    
+        var fieldConfig = Object.assign({}, this.fieldConfig, {
             width: itemWidth,
             height: itemHeight,
-            uploadPath: this.fieldConfig.uploadPath,
-        };
-
+            uploadPath: this.fieldConfig.uploadPath
```
jdreesen, line 110: This should not be necessary anymore.
```suggestion
```

Brainshaker95: Oh my bad, you are correct

Brainshaker95:
> [CLA assistant check](https://cla-assistant.io/pimcore/pimcore?pullRequest=8252)
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/pimcore/pimcore?pullRequest=8252) before we can accept your contribution.
> **1** out of **2** committers have signed the CLA.
> 
> ✅ Brainshaker95
> ❌ patrickrupp
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/pimcore/pimcore?pullRequest=8252) it.

I am Brainshaker95 and patrickrupp... patrickrupp doesn't have a corresponding github account. Is this problematic?

brusch:
@Brainshaker95 please check that your email address which was used in the commit is also configured on Github. 
You can check the email address used here: https://patch-diff.githubusercontent.com/raw/pimcore/pimcore/pull/8252.patch 
and add the address here: https://github.com/settings/emails

It might be necessary to close and reopen the PR to get the information updated.
Thanks a lot!

brusch:
@Brainshaker95 can you please check again (see above) - we'd like to include this patch in `6.8.11` which will be released soon. Thanks!

Brainshaker95:
@brusch Now it should work, can you reopen the PR again? Sorry it took so long, I just didn't have the time until now

brusch:
@Brainshaker95 thanks a lot, however, I've applied the main change already with e4d680dd5349d72948cf1238b215f35af2e9a197 so it's done already 😊

Brainshaker95:
@brusch Ah I didn't notice that. Thanks!

PR #8263, opened 2021-02-23
fix twig 3.0 bc breaks

fashxp:


bundles/EcommerceFrameworkBundle/Resources/views/admin_order/detail.html.twig
```diff
@@ -213,7 +214,9 @@
                                     <small>
                                         {% set amount = null %}
                                         {% set break = false %}
-                                        {% for providerKey, providerClass in paymentProviders if not break %}
+                                        {% for providerKey, providerClass in paymentProviders %}
+                                            {% if not break %}
```
blankse, line 218: @fashxp I think the if block should be over the whole for block

bundles/EcommerceFrameworkBundle/Resources/views/admin_order/detail.html.twig
```diff
@@ -213,7 +214,9 @@
...
+                                            {% endif %}
```
dvesh3, line 219: wrap amount getter code inside if block..

bundles/EcommerceFrameworkBundle/Resources/views/admin_order/detail.html.twig
fashxp: thx, good catch!

PR #8269, opened 2021-02-24
Update 03_Amazon_AWS_S3_Setup.md

andreas-gruenwald:
Fix S3 setup description to make the synchronisation of versions and fileobjects work.


brusch:
Thank you!

PR #8271, opened 2021-02-24
Query builder missing condition variables

dturopoli:
Doing $list->getQuery()->execute() where $list is DataObject\Listing would fail because conditionVariables weren't set in QueryBuilder

### DIFF
--- a/lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
+++ b/lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
@@ -76,7 +76,8 @@ trait QueryBuilderHelperTrait
         }
 
         if ($condition) {
-            $queryBuilder->where($condition);
+            $queryBuilder->where($condition)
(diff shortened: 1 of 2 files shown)


dvesh3:
@dturopoli Thanks for your PR. However, I'm afraid this change could break existing code using `getQuery()` and don't forget `onCreateQuery` callback which is used to modify the query before execute.

brusch:
Would be a topic for `10.x` then. Can you please rebase the PR?

brusch:
Ok, just rebasing it not enought, since there'e were quite a lot of changes there for `10.x`. 
Additionally I guess this would also apply for other listings such as Assets and Documents.

dvesh3:
basically listings which are using `QueryBuilderHelperTrait`

dvesh3:
@dturopoli are you on it? https://github.com/pimcore/pimcore/pull/8271#issuecomment-786725790 thanks!

dvesh3:
@dturopoli thanks!

PR #8272, opened 2021-02-24
Fixes Unique constraint violation on UrlSlugs

NiklasBr:
Upon saving a *product variant* this would throw an error even if it was actually unique: `Unique constraint violated. Slug already used by object` which resolved the issue


NiklasBr:
> Test redis DB is not configured 

Not related to these changes, right?

dvesh3:
@NiklasBr Redis tests are skipped. There are 3 failed tests related to slug change https://github.com/pimcore/pimcore/runs/1968991517#step:11:3777

NiklasBr:
We should allow an object to update its own slug, so this need to change then.

NiklasBr:
I don't have time to understand and fix the failing Tests right now. Closing.

PR #8278, opened 2021-02-24
Fix Classification Store doesn't allow null

### LINKED ISSUE #8277: [bug] Classification Store doesn't allow null
# Bug Report

After [this change](https://github.com/pimcore/pimcore/commit/508d4e9d069d3259f348c5ff020181872a1a351a#diff-5959ee96b2dc092ba3e0a5d9834745486efc70ed786934bd466e72e310a575f7R104), we noticed a regression where all our quantity values were suddenly giving `0` where they should be empty, corrupting the values (because zero is a valid value in some cases).

The linked change prevents classification store to ever be `null`, which doesn't seem expected.

### Expected behavior

Allow `null` in classification store.

### Actual behavior

`null` cast to an empty string which is then interpreted as `0` by quanity unit.

### Steps to reproduce

1. add Quanity unit to a CS
2. set it to something
3. delete the value, make it empty
4. reload the object, now it's set to `0`

dvesh3:
Resolves #8277

### DIFF
--- a/models/DataObject/Classificationstore/Dao.php
+++ b/models/DataObject/Classificationstore/Dao.php
@@ -101,8 +101,8 @@ class Dao extends Model\Dao\AbstractDao
                     }
                     $value = $fd->marshal($value, $object);
 
-                    $data['value'] = isset($value['value']) ? $value['value'] : '';
-                    $data['value2'] = isset($value['value2']) ? $value['value2'] : '';
+                    $data['value'] = $value['value'] ?? null;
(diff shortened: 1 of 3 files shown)


dkarlovi:
Maybe a regression test here too?

dvesh3:
> Maybe a regression test here too?

@dkarlovi here you go https://github.com/pimcore/pimcore/pull/8278/commits/8040b0ca4f0ab6e147767928476e47d7e9d52837

dkarlovi:
Thanks for working on this, @dvesh3!

PR #8286, opened 2021-02-25
added file uri scheme in mail helper for absolute paths

alen125:
Added file scheme uri in Mail helper for resolving absolute paths. This is useful when i.e. generating PDF's with wkhtml and we pass sources to assets with file scheme. Currently it appends host to the file scheme URI.


brusch:
Thank you!

Issue #6475, opened 2020-05-28
[UI] Timeout when saving system settings in certain cases


alexgit2k:
# Bug Report

Saving settings at System settings always fails with "Error: Saving failed!" on a clean demo-installation. However, when reloading the page the changed settings have been stored.

### Expected behavior
Show a success message

### Actual behavior
Message says: "Error: Saving failed!"
No entry in logs or PHP-error-Log

### Steps to reproduce
1. COMPOSER_MEMORY_LIMIT=-1 composer create-project pimcore/demo application
2. cd application && vendor/bin/pimcore-install
3. Go to Settings -> System Settings
4. Click Save button

### Environment
PHP 7.4

NiklasBr:
I also encounter this quite a lot, it seems the ExtJS timeout is too short to account for some system load.

fashxp:
now, system settings can be stored in settings store, details see https://pimcore.com/docs/platform/Pimcore/Tools_and_Features/System_Settings/

NiklasBr:
That's not the same thing @fashxp

fashxp:
Well it is in the consequence ... as the timeout most likely was issues by a container rebuild. With saving settings in settings store, container rebuild is not necessary anymore. 

If there is another cause for that problem, we are happy to get more information and contributions to fix it. Thank you.

NiklasBr:
Would you like me to reopen this issue?

fashxp:
We would appreciate it - if the problem still exists - when you provide additional information as comment in that issue or, if possible, even create pull request to improve that behavior.

PR #8290, opened 2021-02-26
[UI] Increase timeout of buildDefaultStore

jremmurd:
# Enhancement

I suggest to increase the default timeout of proxy stores created via `pimcore.helpers.grid.buildDefaultStore` from 30s to 120s.

This is espcially useful for `/chart` or `/data` requests of custom report, which may take longer depending on the implementation.

 
image


NiklasBr:
Is it possible to increase the same for other places like System Settings as well? See #6475

brusch:
@NiklasBr well, shouldn't this covered by `Ext.Ajax.setTimeout(900000)` here: 
https://github.com/pimcore/pimcore/blob/0d9275da9ed5348a268bc09042d311100b85ba0e/bundles/AdminBundle/Resources/public/js/pimcore/startup.js#L157 
as this isn't using the proxy store?

brusch:
Thanks!

PR #8305, opened 2021-03-01
Make GDPR Exporters independent from REST Webservices

### LINKED ISSUE #7280: Make GDPR Exporters independent from REST Webservices
The data providers for objects an assets were relying on the REST webservices, since #7279 removed them, this needs to be fixed. 

Affected classes: 
```
\Pimcore\Bundle\AdminBundle\GDPR\DataProvider\DataObjects
\Pimcore\Bundle\AdminBundle\GDPR\DataProvider\Assets
```

See `@TODO` in the classes. 

Maybe using the `marshal()` method on the data-types would be an appropriate replacement. 
Also consider pimcore/advanced-object-search#90

Please also introduce a new interface for `marshal()` and `unmarshal()` ... `MarshallableInterface`.

weisswurstkanone:
resolves #7280

### DIFF
--- a/bundles/AdminBundle/GDPR/DataProvider/Exporter.php
+++ b/bundles/AdminBundle/GDPR/DataProvider/Exporter.php
@@ -1,0 +1,179 @@
+<?php
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+declare(strict_types=1);
+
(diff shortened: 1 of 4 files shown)


bundles/AdminBundle/GDPR/DataProvider/Exporter.php
```diff
@@ -0,0 +1,179 @@
...
+use Pimcore\Model\DataObject\Objectbrick;
+
+/**
+ * Class Exporter
+ * @package Pimcore\Bundle\AdminBundle\GDPR\DataProvider
+ *
+ * @internal
+ * @final
+ */
+class Exporter
```
dvesh3, line 33: we should make this independent service. please see https://github.com/pimcore/advanced-object-search/issues/90

weisswurstkanone: IMHO, it should be decoupled from the core, why again provide a centralized service ? wdyt ?

brusch: In this particular case I would also think it make sense to keep it decoupled.

Issue #7840, opened 2021-01-04
Geopoint latitude/longitude constructor


KatrinAckermannEk:
# Improvement Request

Change the signature of Geopoint by flipping the arguments _longitude_ and _latitude_.

## Improvement description

A Geopoint or Geocode is commonly defined as a pair of value consisting of latitude and longitude.
The signature of the constructor in Pimcore defines the reverse order and expects the longitude at first.

As this is not intuitive, I suggest to change the order to adjust it to the common sense:

In pimcore/pimcore/models/DataObject/Data/Geopoint.php:
Is: 
`public function __construct($longitude = null, $latitude = null) {}`

Should: 
`public function __construct($latitude = null, $longitude = null) {}`

NiklasBr:
Seems like a good idea: https://en.wikipedia.org/wiki/ISO_6709#Order,_sign,_and_units

weisswurstkanone:
Hi @NiklasB, ,this definitely makes sense (!) but I don't see a migration scenario from a code perspective. I don't believe that existing code (e.g. import scripts) can be updated without effort and even the customers (and developers) will be aware that the signature changed. Scripts will still work but the data will be wrong. There will be no hard error ;-)

Cruiser13:
We stumbled upon this quite a few times in the past. But you usually note it in development. I'd not change the behaviour as this would require a lot of code changes in customer projects and as weisswurstkanone said, this will throw no deprecation warnings or anything to let developers know about the change.

NiklasBr:
It would be possible to solve by deprecating the `Geopoint` class, while introducing a new effectively identical class in 6.9.0 but with the correct order of the constructor arguments.

brusch:
I'd prefer @NiklasBr  proposal.

weisswurstkanone:
@NiklasBr do you have a meaningful name for the new class ?

brusch:
What do you think about `GeoCoordinates` which is also used by schema.org

NiklasBr:
Seems like a good choice, the only thing I don't like is that it feels odd with a pluralised classname when all others are singular such as as `RgbaColor`.

KatrinAckermannEk:
What about `GeoPoint`, just as in the context of Pimcore's Targeting and Data Type?

brusch:
Done by #8306

PR #8306, opened 2021-03-01
Geopoint latitude/longitude constructor

weisswurstkanone:


weisswurstkanone:
partly resolves #7840

kaurov:
What for DataObjects class Geopoint was renamed to GeoCoordinates ?

dvesh3:
@kaurov It was done to use the correct ordering of the lat/long in the construct parameters. please see https://github.com/pimcore/pimcore/issues/7840

PR #8311, opened 2021-03-02
[Data Objects] Fieldcollection with Localized Fields deletes and overwrites content after unpublished save

### LINKED ISSUE #7974: Fieldcollection with Localized Fields deletes and overwrites content after unpublished save
# Bug Report

### Expected behavior
The Fieldcollections with Localized Fields should keep the data at the right positions after saving. 

### Actual behavior
1. After adding new FC and saving without any changes, adding afterwards data and hit save again, data of next element is missing.
2. Another Bug appears, when you add a new FC above the current elements, add Data and hit save multiple times afterwards, contents of data of elements below get overwritten.

### Steps to reproduce
Add following Test FC and Class to demo:
[TEST_FC_BUG.zip](https://github.com/pimcore/pimcore/files/5856519/TEST_FC_BUG.zip)

First behavior:
1. Create Object from Class Test
2. Add 1 FC element "test", add data and hit save + reload object
image

3. Add another FC element before and hit save without reload
image

4. Add data to the new FC element and hit save
image

5. reload and data from first element should be missing
image

Second behavior::
1. Create Object from Class Test
2. Add 3 FC element "test" add unique data to the elements and hit save + reload
image

3. add new field before the other 3 elements, add data and save multiple times 
image

4. Fields below last element should have same content as previous
image

foxgreek:
Resolves #7974

## Additional info  
Fieldcollections and Blocks always send their data without checking if content isDirty, but a LocalizedFields element itself checks again the isDirty() function of its children.

Therefore, a small check of the context container allows to ignore the isDirty() function of LocalizedFields in Fieldcollections and Blocks.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/localizedfields.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/localizedfields.js
@@ -492,6 +492,7 @@ pimcore.object.tags.localizedfields = Class.create(pimcore.object.tags.abstract,
     getValue: function () {
         var localizedData = {};
         var currentLanguage;
+        var ignoreIsDirty = ['fieldcollection'].includes(this.getContext().containerType)
 
         for (var i = 0; i < this.frontendLanguages.length; i++) {
             currentLanguage = this.frontendLanguages[i];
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Thanks a lot!

PR #8320, opened 2021-03-02
JS helper to format durations

NiklasBr:
After and before:

image

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/video.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/video.js
@@ -125,7 +125,7 @@ pimcore.asset.video = Class.create(pimcore.asset.asset, {
                 detailsData[t("height")] = this.data.customSettings.videoHeight;
             }
             if(this.data.customSettings['duration']) {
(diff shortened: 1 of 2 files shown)


weisswurstkanone:
Hi @NiklasBr, I like the idea! 

But I fear that there will be localization issues. What do you think about this ISO standard?
https://en.wikipedia.org/wiki/ISO_8601

NiklasBr:
@weisswurstkanone this is not a date format, but a [duration](https://en.wikipedia.org/wiki/ISO_8601#Durations), I mean yes we can show the format like `12H30M5S` but that seems too technical to me,

brusch:
Why not just `01:45:51`? (`hh:mm:ss`) It's international and everyone knows what it means?

NiklasBr:
Apparently both iTunes and WMP show durations on minutes without the leading "0" when it is a single digit, maybe I should copy that?
image

brusch:
@NiklasBr yep, fine for me as well 😊 Thanks!

NiklasBr:
image

brusch:
Thanks a lot! 👍

PR #8324, opened 2021-03-03
bugfix fullpath of hardlinks

zerlpaMMT:
documents inside hardlinks from different sites used the wrong domain


models/Document.php
```diff
@@ -879,6 +879,7 @@ public function getFullPath(bool $force = false)
         if (!$link && \Pimcore\Tool::isFrontend() && Site::isSiteRequest() && !FrontendTool::isDocumentInCurrentSite($this)) {
             if ($masterRequest && ($masterDocument = $masterRequest->get(DynamicRouter::CONTENT_KEY))) {
                 if ($masterDocument instanceof WrapperInterface) {
+                    $hardlinkPath = null;
```
brusch, line 882:
```suggestion
                    $hardlinkPath = '';
```
Better use a `string` here, as `strpos()` would convert it otherwise.

brusch:
Thanks a lot! 👌

Issue #7790, opened 2020-12-18
Update Core Bundles to be consistent with new folder structure


dvesh3:
# Feature Request
follow up to #7679

e.g. update config files `*.yml` to `*.yaml`

brusch:
Done by referenced PRs above.

PR #8325, opened 2021-03-03
It should be possible to override a bundles config - follow up to #8186

brusch:
follow up to #8186

### DIFF
--- a/lib/Kernel.php
+++ b/lib/Kernel.php
@@ -130,21 +130,21 @@ abstract class Kernel extends SymfonyKernel
      */
     public function registerContainerConfiguration(LoaderInterface $loader)
     {
-        $this->microKernelRegisterContainerConfiguration($loader);
-
         $loader->load(function (ContainerBuilder $container) {
             $this->registerExtensionConfigFileResources($container);
         });
+
+        $bundleConfigLocator = new BundleConfigLocator($this);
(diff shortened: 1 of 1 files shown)


lib/Kernel.php
```diff
@@ -130,22 +130,22 @@ protected function configureRoutes(RoutingConfigurator $routes): void
      */
     public function registerContainerConfiguration(LoaderInterface $loader)
     {
-        $this->microKernelRegisterContainerConfiguration($loader);
-
         $loader->load(function (ContainerBuilder $container) {
             $this->registerExtensionConfigFileResources($container);
         });
 
+        $bundleConfigLocator = new BundleConfigLocator($this);
```
dvesh3, line 137: I think this should go when https://github.com/pimcore/pimcore/issues/7790 is implemented. 
```suggestion
       //@TODO Pimcore 10: Remove when https://github.com/pimcore/pimcore/issues/7790 is done
        $bundleConfigLocator = new BundleConfigLocator($this);
```

brusch: @dvesh3 don't think so, as this loads the config from `Resources/config/pimcore` which isn't obsolete with #7790. 
But probably we need to include `yaml` here: https://github.com/pimcore/pimcore/blob/5fcf30280c6d95dd8e3a08e9663300b0c5b9b579/lib/Config/BundleConfigLocator.php#L103

PR #8331, opened 2021-03-04
[Ecommerce] Fix pricing rule save misleading result message

jremmurd:
# Bugfix

Clicking the save button inside a pricing rule's tab always showed a success notification, not considering the actual response. With this PR the returned result is taken into account and the response message is shown to the user.


brusch:
Thanks a lot! 👍

PR #8332, opened 2021-03-04
SettingsStore

fashxp:
Adds a settings store to Pimcore, that persists simple key-value data to Pimcore database.

### DIFF
--- a/models/Tool/SettingsStore/Dao.php
+++ b/models/Tool/SettingsStore/Dao.php
@@ -30,7 +30,7 @@ class Dao extends Model\Dao\AbstractDao
 
     /**
      * @param string $id
-     * @param $data
+     * @param mixed $data
      * @param string|null $scope
      * @param string $type
      * @return bool
--- a/models/Tool/SettingsStore/Dao.php
+++ b/models/Tool/SettingsStore/Dao.php
@@ -1,0 +1,92 @@
+<?php
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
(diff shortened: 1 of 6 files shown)


jdreesen:
Nice, but I see some problems with the current implementation. 
As far as I can see, the `id`'s are unique, which means they *must* be namespaced somehow to prevent clashes between extensions etc. 
But when `id`'s are namespaced... do we still need the `scope`? 
So, maybe the primary key should consist of `id` *and* `scope`?
Otherwise, it should be made *very clear* in the docs that you *have to* namespace your `id`'s, no matter if you use `scopes` or not.

models/Tool/SettingsStore/Dao.php
```diff
@@ -0,0 +1,92 @@
...
+     */
+    public function getById(string $id): bool
+    {
+        $item = $this->db->fetchRow('SELECT * FROM ' . self::TABLE_NAME . ' WHERE id = ?', $id);
+
+        if (is_array($item) && array_key_exists('id', $item)) {
+            $this->assignVariablesToModel($item);
+
+            $data = $item['data'] ?? null;
+            settype($data, $this->model->getType());
```
jdreesen, line 76: This is already done in the model: https://github.com/pimcore/pimcore/pull/8332/files#diff-4902f5fa9a4944a318a52a600c1c09365edbf626bfa473202120d1b7e773e2caR196

brusch:
@jdreesen thanks a lot for your review & valuable feedback! 
I agree, id makes sense to span the PK over `id` and `scope`.

PR #8333, opened 2021-03-04
Override isEmpty in Classificationstore.php

Daniel-Ateles:
When using variants that inherits classificationstores the overridden/inherited values is not fetched because isEmpty will check if the classificationstore object is empty/null - which it never is. Hence we need to override isEmpty (like all other Data-classes) and check if the classificationstores *items* are empty, not the object itself.


weisswurstkanone:
thx!

PR #8341, opened 2021-03-05
[AdminBundle] Fix icon library

podarcis:
The icon library could not be opened and threw an exception in the Twig template.

```
Too few arguments to function Pimcore\Model\Site::getById(), 0 passed in /var/www/pimcore-master/var/cache/dev/twig/.....php on line 48 and exactly 1 expected
```


brusch:
Thanks a lot!

PR #8342, opened 2021-03-05
[AdminBundle] Fix opening application logger

podarcis:
Ext JS KeyNav constructor doesn't support multiple arguments anymore and
thus threw an exception. The functionality "search on key ENTER" was
already fixed/implemented in #2539, so this code seems obsolete.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/log/admin.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/log/admin.js
@@ -294,14 +294,6 @@ pimcore.log.admin = Class.create({
                     handler: this.find.bind(this),
                     iconCls: "pimcore_icon_search"
                 }],
-                listeners: {
-                    afterRender: function(formCmp) {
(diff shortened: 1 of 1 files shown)


brusch:
Thanks a lot!

podarcis:
@brusch You might want to pick a fixup podarcis/pimcore@695a4ce0a as well.

brusch:
@podarcis this is still in use, isn't it? 
see https://github.com/pimcore/pimcore/blob/a9f787995202fa4729ab26405c940464ae7f35f4/bundles/AdminBundle/Resources/public/js/pimcore/log/admin.js#L300

podarcis:
@brusch Not since pimcore/pimcore@b5fb2bc860, which was the original commit of my PR. The second commit removed the now  obsolete variable.

brusch:
@podarcis oh dear, your're right, pulling the latest code helps 😊 Done d85612673295f7035cfd242c34d775d9310f8a20

PR #8357, opened 2021-03-08
[Web2Print] Fix Pdf generation in background process

### LINKED ISSUE #8355: [Web2Print] Starting Processing not blocking anymore
regression of https://github.com/pimcore/pimcore/pull/8230/files#diff-b09ab260fb424fed749d1fdf730a6e1f12249fdf63d14728b0c6a48d25d46c0aR93

running non-blocking is more important than writing to an output file.

dvesh3:
Resolves #8355
- also fixes running php script in background

### DIFF
--- a/lib/Web2Print/Processor.php
+++ b/lib/Web2Print/Processor.php
@@ -90,11 +90,13 @@ abstract class Processor
 
         if (!$disableBackgroundExecution) {
             $process->start();
-            $logHandle = fopen(PIMCORE_LOG_DIRECTORY . DIRECTORY_SEPARATOR . 'web2print-output.log', 'a');
-            $process->wait(function ($type, $buffer) use ($logHandle) {
-                fwrite($logHandle, $buffer);
-            });
-            fclose($logHandle);
+            //we can't write output to log file, since wait() is a blocking method
+            //see https://github.com/pimcore/pimcore/issues/8355
+//            $logHandle = fopen(PIMCORE_LOG_DIRECTORY . DIRECTORY_SEPARATOR . 'web2print-output.log', 'a');
+//            $process->wait(function ($type, $buffer) use ($logHandle) {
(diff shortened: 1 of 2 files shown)


lib/Web2Print/Processor.php
```diff
@@ -90,11 +90,13 @@ public function preparePdfGeneration($documentId, $config)
 
         if (!$disableBackgroundExecution) {
             $process->start();
-            $logHandle = fopen(PIMCORE_LOG_DIRECTORY . DIRECTORY_SEPARATOR . 'web2print-output.log', 'a');
-            $process->wait(function ($type, $buffer) use ($logHandle) {
-                fwrite($logHandle, $buffer);
-            });
-            fclose($logHandle);
+            //we can't write output to log file, since wait() is a blocking method
```
brusch, line 93: @dvesh3 @fashxp Don't think this is a problem so we can remove all the comments. But one thing: 
https://github.com/pimcore/pimcore/blob/dd9fc1a906b7b6eed72af8ebffa9733baf1f9708/bundles/CoreBundle/Command/Web2PrintPdfCreationCommand.php#L53 
This should be logged to the standard log, so that this message isn't getting lost.

brusch (changes requested):

fashxp:
background task dies as soon as main tasks exits, so we need another solution for that.

lib/Web2Print/Processor.php
fashxp: @dvesh3  can you please revert all these changes, we will solve that with https://github.com/pimcore/pimcore/pull/8371

dvesh3: we can close this PR and copy over upgrade notes to #8371

brusch: Done

brusch:
Obsolete #8371

PR #8363, opened 2021-03-09
Fix: "Call to undefined method getInherited" when saving a document a second time

### LINKED ISSUE #7760: "Call to undefined method getInherited" when saving a document a second time
I have the following error in multiple projects with Pimcore 6.8.5.

### Steps to reproduce

1. open a document
2. save the document
3. repeat step 2

### Expected behavior

The document is saved after step 2 and after step 3.

### Actual behavior

The document is saved after step 2 but after step 3 an error message appears. I have to close the document and open it again, then saving the document works again but don't work after a second attempt to save.

```
URL: /admin/page/save?task=publish
Method: PUT
Message: Call to undefined method getInherited in class Pimcore\Model\DataObject\ConfigWebsite
Trace: 
in /web/web12545/vendor/pimcore/pimcore/lib/Model/AbstractModel.php:251
#0 /web/web12545/vendor/pimcore/pimcore/models/DataObject/AbstractObject.php(1338): Pimcore\Model\AbstractModel->__call('getInherited', Array)
#1 /web/web12545/vendor/pimcore/pimcore/models/Document/PageSnippet.php(128): Pimcore\Model\DataObject\AbstractObject->__call('getInherited', Array)
#2 /web/web12545/vendor/pimcore/pimcore/models/Document.php(424): Pimcore\Model\Document\PageSnippet->update(Array)
#3 /web/web12545/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/Document/PageController.php(184): Pimcore\Model\Document->save()
#4 /web/web12545/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(151): Pimcore\Bundle\AdminBundle\Controller\Admin\Document\PageController->saveAction(Object(Symfony\Component\HttpFoundation\Request))
#5 /web/web12545/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php(68): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#6 /web/web12545/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php(200): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#7 /web/web12545/web/app.php(36): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#8 {main}
```
I want to note that `ConfigWebsite` represents a data object in our projects which is used fairly a lot within these projects. This makes it probably a bit difficult for you to reproduce the issue. What I found out:

In /vendor/pimcore/pimcore/models/Document/PageSnippet.php there is a method `update($params = [])`. In this method you can find the following block:

```
      if (is_array($this->getEditables()) and count($this->getEditables()) > 0) {
            foreach ($this->getEditables() as $name => $editable) {
                if (!$editable->getInherited()) {
                    $editable->setDao(null);
                    $editable->setDocumentId($this->getId());
                    $editable->save();
                }
            }
        }
```

If I change the following line:

`if (!$editable->getInherited()) {`

to:

`if (method_exists($editable, 'getInherited') && !$editable->getInherited()) {`

… the problem does no longer exist. But I don't know if this change has some other unwanted consequences.

So the main question is: Should it be possible at all that there is a $editable within the foreach loop without a getInherited method? If no, then there is probably a bug somewhere because it happens that the getInherited method does not exist for every $editable. But I really have no idea what exactly in our project is triggering this. Please let me know if I can somehow help to find the culprit.

blankse:
Resolve #7760

If the document is cloned in the save process (checkMissingRequiredEditable and saveVersion), the editable is referenced to the original document.
DeepCopy copy the element array and so the original objects are lost in the source. When DeepCopy copies the editable array the hashes can based on other objects.

### DIFF
--- a/models/Document/PageSnippet.php
+++ b/models/Document/PageSnippet.php
@@ -682,9 +682,24 @@ abstract class PageSnippet extends Model\Document
     /**
      * @TODO: remove with $this->elements
      */
+    public function __clone()
+    {
+        unset($this->elements, $this->inheritedElements, $this->editables, $this->inheritedEditables);
+
+        $this->inheritedEditables = [];
+        $this->elements = & $this->editables;
+        $this->inheritedElements = & $this->inheritedEditables;
+
+        parent::__clone();
+    }
+
+    /**
+     * @TODO: remove with $this->elements
+     */
     public function __wakeup()
     {
-        $this->editables = & $this->elements;
+        $this->elements = & $this->editables;
+        $this->inheritedElements = & $this->inheritedEditables;
 
         parent::__wakeup();
     }


fashxp:
this PR breaks the documents ... it is not possible to add any elements (areabricks/blocks) anymore

Issue #8369, opened 2021-03-09
[Translations] Using parameter %count% in simple translation values will create additional translation keys


fashxp:
having a translation `count_key` that translates to `%count% Count` 

calling ` $this->translator->trans('count_key', ['%count%' => 2])` will create a new translation entry with key  `2 Count` and empty value.

dvesh3:
Fixed by #8375

PR #8370, opened 2021-03-09
Fix missing parameters for translation listings

fashxp:
doing something like fails: 

```
$translations = new Translation\Listing();
$translations->setDomain('messages');
$translations = $translations->getTranslations(); 
```

tests fail because of https://github.com/pimcore/pimcore/issues/8369

### DIFF
--- a/tests/Unit/Translation/TranslatorTest.php
+++ b/tests/Unit/Translation/TranslatorTest.php
@@ -18,6 +18,7 @@ declare(strict_types=1);
 namespace Pimcore\Tests\Unit\Translation;
 
 use Codeception\Util\Stub;
+use Pimcore\Model\Translation;
 use Pimcore\Model\Translation\Website;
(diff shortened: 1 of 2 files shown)


tests/Unit/Translation/TranslatorTest.php
```diff
@@ -205,4 +206,22 @@ public function testTranslateCaseSensitive()
...
         $this->assertEquals($this->translations['en']['case_key'], $translator->trans('CASE_KEY'));
     }
+
+    public function testLoadingTranslationList() {
+
+        $translations = new Translation\Listing();
+        $translations->setDomain('messages');
+
+        $translations = $translations->getTranslations();
+        $this->assertCount(9, $translations);
```
dvesh3, line 216:
```suggestion
        $this->assertCount(count($this->translations['en']), $translations);
```

PR #8371, opened 2021-03-09
Revert execInBackground

### LINKED ISSUE #8355: [Web2Print] Starting Processing not blocking anymore

fashxp:
Symfony Process background processes are terminated as soon as parent process exists. So this is a change in behavior. 
For now, reverted execInBackground, will switch to queue implementations in Pimcore X.x versions. 

fixes #8355

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/NewsletterController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/NewsletterController.php
@@ -364,10 +364,10 @@ class NewsletterController extends DocumentControllerBase
             'progress' => 0,
         ], 'newsletter');
 
-        Console::runPhpScript(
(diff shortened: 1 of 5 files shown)


dvesh3 (changes requested):
we should revert all `runPhpScriptInBackground()` usages replaced in https://github.com/pimcore/pimcore/pull/8230/files 

and Is this required anymore? https://github.com/pimcore/pimcore/pull/8329

brusch (approved):

PR #8400, opened 2021-03-11
[Data object versions] Show version note

BlackbitNeueMedien:
In some use-cases it is good to add a description or some meta data to data object versions. There is a field `note` in the `versions` table. Currently this can only be filled by manually calling the `Version::setNote()` method as far as I know. But currently this field is not displayed anywhere in the Pimcore backend. With this PR the version note gets shown in the data object preview / comparison - only if it is filled. So for projects which do not use version notes everything will look the same as before but for those which do the version note is displayed.

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/DataObject/DataObject/previewVersion.html.twig
+++ b/bundles/AdminBundle/Resources/views/Admin/DataObject/DataObject/previewVersion.html.twig
@@ -16,6 +16,14 @@
         <th>Key</th>
         <th>Value</th>
     </tr>
+    {% if versionNote %}
+        <tr class="system">
+            <td>Version Note</td>
+            <td>&nbsp;</td>
+            <td>{{ versionNote }}</td>
+        </tr>
+        <p><b>Version Note: </b> {versionNote}</p>
+    {% endif %}
     <tr class="system">
         <td>Date</td>
         <td>o_modificationDate</td>
--- a/bundles/AdminBundle/Resources/views/Admin/DataObject/DataObject/previewVersion.html.twig
+++ b/bundles/AdminBundle/Resources/views/Admin/DataObject/DataObject/previewVersion.html.twig
@@ -16,14 +16,6 @@
         <th>Key</th>
         <th>Value</th>
     </tr>
-    {% if versionNote %}
-        <tr class="system">
-            <td>Version Note</td>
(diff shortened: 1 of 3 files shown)


dvesh3:
@BlackbitNeueMedien you also pass optional parameter on save like `$object->save(["versionNote" => "my new version"]);` and it displayed on versions list next to scheduled column:
image

and it would be nice to see it in version preview as well. thanks for your PR! 😊

bundles/AdminBundle/Resources/views/Admin/DataObject/DataObject/previewVersion.html.twig
```diff
@@ -16,6 +16,14 @@
         <th>Key</th>
         <th>Value</th>
     </tr>
+    {% if versionNote %}
+        <tr class="system">
+            <td>Version Note</td>
+            <td>&nbsp;</td>
+            <td>{{ versionNote }}</td>
+        </tr>
+        <p><b>Version Note: </b> {versionNote}</p>
```
dvesh3, line 25: having a row is enough...please kick it 
image and please move it under "published". thanks!

BlackbitNeueMedien:
Oh I see - my screen was only too small ;-)
Bildschirmfoto 2021-03-11 um 09 28 12

But nevertheless makes sense to also display it in version preview.

bundles/AdminBundle/Resources/views/Admin/DataObject/DataObject/previewVersion.html.twig
BlackbitNeueMedien: Fixed. The `<p>` was actually not supposed to be committed.

BlackbitNeueMedien:
Do you really think that it makes more sense below `Published` field? I thought that it would mix up contexts because id, key etc. (system fields) belong to the object while not note belongs to the version. For this reason I put it on top. But now I have put it under `Published` as you requested. What do you think?

dvesh3:
> Do you really think that it makes more sense below `Published` field? I thought that it would mix up contexts because id, key etc. (system fields) belong to the object while not note belongs to the version. For this reason I put it on top. But now I have put it under `Published` as you requested. What do you think?

I would like to see this at the last position to be consistent with list view and the context is already mixed since date belongs to version, which is on top.

dvesh3:

PR #8404, opened 2021-03-11
[Twig] Removed argument $removeAttributes from pimcore_image_thumbnail_html 

brusch:
... doesn't exist anymore in v10


brusch:
Follow up to #7531

PR #8412, opened 2021-03-11
[Data objects] Do not create non-unique index in object_store table

BlackbitNeueMedien:
Follow-up of #8409

Currently when you tick `indexed` and `unique` checkboxes for a field 2 indexes get created in the `object_store_*` table slowing down write performance. With this PR `object_store_*` tables only get unique indexes (if `unique` is ticked) and `object_query_*` tables only get non-unique indexes (when `indexed` is ticked).


brusch:
@BlackbitNeueMedien thanks for your follow-up PR - since this is not really a bug, can you please rebase on `master`? 
Thanks! 😊

brusch:
Thanks a lot! 👍

PR #8415, opened 2021-03-11
Shorter Twig syntax for odd rows

NiklasBr:
Follow-up on #8403
see https://twig.symfony.com/doc/3.x/tests/odd.html

I was too slow to give feedback on the PR :)


brusch:
Nice, thanks a lot! 😊

PR #8416, opened 2021-03-11
[BUGFIX] fixes marshalling of DataObject Link

### LINKED ISSUE #8381: Increased Memory Usage when Saving DataObject with Block in LocalizedFields
# Bug Report

Increased Memory Usage when saving DataObjects which contains Blocks in LocalizedFields

### Expected behavior
The used memory should not differ.

### Actual behavior
After reloading a DataObject (from Cache), changing it and saving it again, then the used memory is increased.

### Steps to reproduce
**(1) Use class export json to create a new Class StorageBug**
[class_StorageBug_export.zip](https://github.com/pimcore/pimcore/files/6109688/class_StorageBug_export.zip)

**(2) Create a new DataObject with one link in Main Block and one link in Additional Block**

image

**(3) Check the memory usage in the table _object_localized_data_storagebug_**

use the following query to check the length of the fields

SELECT LENGTH(mainBlock), LENGTH(additionalBlock) FROM object_localized_data_storagebug;

I got a length of **537** characters for both fields.

image

**(4) Reload the DataObject, change the Link of the Main Block and save it again**

image

**(5) Check the memory usage in the table _object_localized_data_storagebug_ again**

Now I got **538** characters for the first changed field and **17287** characters for the second field (cache).

image

### Assumption

My assumption is that it relates to the PR https://github.com/pimcore/pimcore/pull/6328.

The problem is that the owner will be serialized too, when it comes from cache.

A workaround for me was to comment the following lines of the class 

_pimcore/pimcore/models/DataObject/Data/BlockElement.php_

image

storbahn:
Resolves #8381

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Link.php
+++ b/models/DataObject/ClassDefinition/Data/Link.php
@@ -30,6 +30,7 @@ class Link extends Data implements ResourcePersistenceAwareInterface, QueryResou
     use Extension\ColumnType;
     use Extension\QueryColumnType;
     use DataObject\ClassDefinition\NullablePhpdocReturnTypeTrait;
+    use DataObject\Traits\ObjectVarTrait;
 
     /**
      * Static type of this element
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
how is this related to #8381 ?

weisswurstkanone:
could you please add a reference to the right issue including steps to reproduce. thx a lot!

storbahn:
In the example I mentioned in the issue I used a Link in a Block in a LocalizedField and when saving it, I could recognize an increased memory usage in the db table. After implementing this Bugfix I could not reproduce it anymore.

It seems that the original problem was the Marshalling of the Link DataObject.

weisswurstkanone:
please be careful here. is it backwards compatible ? I guess in unmarshal we have to take precautions

storbahn:
In my opinion it should be backwards compatible. (therefore I check if $data is an array)

weisswurstkanone:
thanks a lot!

PR #8427, opened 2021-03-12
[Translations] getAllTranslations must not cache with single cache-key

### LINKED ISSUE #8426: [Translations] getAllTranslations must not cache with single cache-key
needs to consider listing-settings ... like `getOrder`, `getGroupBy`, `getOffsetLimit`  and `getCondition`

fashxp:
fixed #8426

### DIFF
--- a/models/Translation/Listing/Dao.php
+++ b/models/Translation/Listing/Dao.php
@@ -98,7 +98,9 @@ class Dao extends Model\Listing\Dao\AbstractDao
                 $translations[$t['key']]->setModificationDate($t['modificationDate']);
             }
 
-            Cache::save($translations, $cacheKey, ['translator', 'translate'], 999);
+            if(empty($this->model->getConditionParams())) {
+                Cache::save($translations, $cacheKey, ['translator', 'translate'], 999);    
+            }
         }
 
         return $translations;
--- a/models/Translation/Listing/Dao.php
+++ b/models/Translation/Listing/Dao.php
@@ -73,11 +73,10 @@ class Dao extends Model\Listing\Dao\AbstractDao
      */
     public function getAllTranslations()
     {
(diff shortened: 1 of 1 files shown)


models/Translation/Listing/Dao.php
```diff
@@ -73,11 +73,10 @@ public function getCount()
      */
     public function getAllTranslations()
     {
-        $cacheKey = $this->getDatabaseTableName().'_data';
-        if (!$translations = Cache::load($cacheKey)) {
+        $queryBuilder = $this->getQueryBuilder(['*']);
+        $cacheKey = $this->getDatabaseTableName().'_data_' . md5((string)$queryBuilder);
+        if (!empty($this->model->getConditionParams()) || !$translations = Cache::load($cacheKey)) {
```
dvesh3, line 78: if `addConditionParam()` called on listing, then it will overwrite the cache everytime... Is this expected behavior?

fashxp: either we need to include all the condition params to the cache key, or we deactivate caching as soon as condition params are set. 
I was thinking that adding the condition params to the cache will create tons of cache keys, that are not used very often anyway (because the queries are very specific...)

dvesh3: but it will not deactivate caching completely as on line https://github.com/pimcore/pimcore/pull/8427/files#diff-f28aba119c2334883e102db25bad594b03bb7900a84da7e69f470a4a3a40913aR101 still creates a new cache item with generated cache key..

fashxp: good catch!

fashxp: please recheck

dvesh3 (approved):

PR #8443, opened 2021-03-15
Filesize check makes only sense when file is read

das-peter:
Mime detection based on the file extension works perfectly fine on an empty file.
So check filesize only if we "fallback" to finfo.

### DIFF
--- a/lib/Tool/Mime.php
+++ b/lib/Tool/Mime.php
@@ -30,21 +30,22 @@ class Mime
             throw new \Exception('File ' . $file . " doesn't exist");
         }
 
+
+        if (!$filename) {
+            $filename = basename($file);
+        }
+
+        $extensionMapping = \Pimcore::getContainer()->getParameter('pimcore.mime.extensions');
+
+        // check for an extension mapping first
+        if ($filename) {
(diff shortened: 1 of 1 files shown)


dvesh3:
candidate for 6.8?

brusch:
I'm wondering if this is really right, a file called `foo.jpg` which is empty, does it really have the mime-type `image/jpeg`, i guess not, right?

das-peter:
> I'm wondering if this is really right, a file called foo.jpg which is empty, does it really have the mime-type image/jpeg, i guess not, right?

Well, the system as is uses file extensions if a mapping is available. As such the contents don't matter at all. I can upload a binary and name it not-an-executable.jpg and if ".jpg" is mapped to "image/jpeg" it will be happily categorized as image.
Considering that, why bother with the filesize on a 0 byte file only?
I think if we want to bother with filesize we should also always rely on "magic bytes" detection to determine file types.

brusch:
Yes, you're right, probably we should just switch to Symfony Mime component and just rely on `fileinfo`.

das-peter:
Totally agree, but that felt out of scope for a quick win ;)

brusch:
Closed in favor of #8447

PR #8449, opened 2021-03-15
Use `JSON_THROW_ON_ERROR` instead of custom JSON error handling

jdreesen:
Since this was added to `master` in #8420 and #8435 we should use PHP >= 7.3 features.


brusch:
Makes absolutely sense, thanks a lot! 👍

PR #8451, opened 2021-03-15
Fixes deleting a Website Setting not clearing it from cache

NiklasBr:
If a setting is first deleted:

```php
$settingObject->delete();
```

It was then still available to the process and returned via 

```php
WebsiteSetting::getByName()
```

Which should not have happened. This overrides the Dao to to resolve the issue.

### DIFF
--- a/models/WebsiteSetting.php
+++ b/models/WebsiteSetting.php
@@ -20,7 +20,6 @@ use Pimcore\Model\Element\Service;
 /**
  * @method \Pimcore\Model\WebsiteSetting\Dao getDao()
  * @method void save()
- * @method void delete()
  */
 class WebsiteSetting extends AbstractModel
 {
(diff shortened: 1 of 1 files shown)


brusch:
@NiklasBr can you please check PhpStan? Thanks a lot!

NiklasBr:
Yes, found the issue

models/WebsiteSetting.php
```diff
@@ -311,4 +310,19 @@ public function clearDependentCache()
     {
         \Pimcore\Cache::clearTag('website_config');
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function delete()
+    {
+        $nameCacheKey = $this->getName() . '~~~' . $this->getSiteId() . '~~~' . $this->getLanguage();
```
brusch, line 319: I think it would be good to have this in a `protected static function`, then we can share it with `getByName`.

brusch (changes requested):

models/WebsiteSetting.php
brusch: I'd leave it as it is for now, as it also has different parameters (just the `id`). Thanks!

brusch:
LGTM - thanks a lot 👍

Issue #7189, opened 2020-09-29
Mark internal/final classes, methods, properties


brusch:
PhpDoc: `@internal` 
PhpDoc: `@final` or directly `final` 

- all admin controllers should be `final` and tagged as `@internal`

PR #8453, opened 2021-03-15
[AdminBundle] Mark classes internal/final

dvesh3:
Resolves partially #7189

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/PageController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PageController.php
@@ -29,7 +29,7 @@ use Symfony\Component\Routing\Annotation\Route;
  *
  * @internal
  */
-class PageController extends DocumentControllerBase
+final class PageController extends DocumentControllerBase
 {
     use ElementEditLockHelperTrait;
 
--- a/bundles/AdminBundle/Controller/Admin/Document/PageController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PageController.php
@@ -26,6 +26,8 @@ use Symfony\Component\Routing\Annotation\Route;
 
 /**
  * @Route("/page")
+ *
+ * @internal
  */
 class PageController extends DocumentControllerBase
 {
--- a/bundles/AdminBundle/Controller/Admin/Document/PrintpageControllerBase.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PrintpageControllerBase.php
@@ -25,7 +25,7 @@ use Symfony\Component\HttpFoundation\Request;
 /**
  * @internal
  */
-class PrintpageControllerBase extends DocumentControllerBase
+abstract class PrintpageControllerBase extends DocumentControllerBase
 {
     use ElementEditLockHelperTrait;
 
--- a/bundles/AdminBundle/Controller/Admin/Document/PrintpageControllerBase.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PrintpageControllerBase.php
@@ -22,6 +22,9 @@ use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 
+/**
+ * @internal
+ */
 class PrintpageControllerBase extends DocumentControllerBase
 {
     use ElementEditLockHelperTrait;
--- a/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
+++ b/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
@@ -35,7 +35,7 @@ use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 /**
  * @internal
  */
-class ElementControllerBase extends AdminController
+abstract class ElementControllerBase extends AdminController
 {
     /**
      * @param ElementInterface $element
--- a/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
+++ b/bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
@@ -32,6 +32,9 @@ use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
+/**
+ * @internal
+ */
 class ElementControllerBase extends AdminController
 {
     /**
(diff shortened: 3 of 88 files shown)


bundles/AdminBundle/Controller/Admin/Document/PageController.php
```diff
@@ -26,6 +26,8 @@
 
 /**
  * @Route("/page")
+ *
+ * @internal
  */
 class PageController extends DocumentControllerBase
```
brusch, line 32:
```suggestion
final class PageController extends DocumentControllerBase
```

bundles/AdminBundle/Controller/Admin/Document/PrintpageControllerBase.php
```diff
@@ -22,6 +22,9 @@
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 
+/**
+ * @internal
+ */
 class PrintpageControllerBase extends DocumentControllerBase
```
brusch, line 28:
```suggestion
abstract class PrintpageControllerBase extends DocumentControllerBase
```

bundles/AdminBundle/Controller/Admin/ElementControllerBase.php
```diff
@@ -32,6 +32,9 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
+/**
+ * @internal
+ */
 class ElementControllerBase extends AdminController
```
brusch, line 38:
```suggestion
abstract class ElementControllerBase extends AdminController
```

bundles/AdminBundle/Controller/GDPR/PimcoreUsersController.php
```diff
@@ -27,8 +27,10 @@
  * @Route("/pimcore-users")
  *
  * @package GDPRDataExtractorBundle\Controller
```
brusch, line 29:
```suggestion
```

bundles/AdminBundle/Controller/GDPR/SentMailController.php
```diff
@@ -27,8 +27,10 @@
  * @Route("/sent-mail")
  *
  * @package GDPRDataExtractorBundle\Controller
```
brusch, line 29:
```suggestion
```

bundles/AdminBundle/Controller/Reports/AnalyticsController.php
```diff
@@ -27,6 +27,8 @@
 
 /**
  * @Route("/analytics")
+ *
+ * @internal
  */
 class AnalyticsController extends ReportsControllerBase implements KernelControllerEventInterface
```
brusch, line 33:
```suggestion
final class AnalyticsController extends ReportsControllerBase implements KernelControllerEventInterface
```

bundles/AdminBundle/Controller/Reports/CustomReportController.php
```diff
@@ -24,6 +24,8 @@
 
 /**
  * @Route("/custom-report")
+ *
+ * @internal
  */
 class CustomReportController extends ReportsControllerBase
```
brusch, line 30:
```suggestion
final class CustomReportController extends ReportsControllerBase
```

bundles/AdminBundle/Controller/Reports/ReportsControllerBase.php
```diff
@@ -17,6 +17,9 @@
 use Pimcore\Bundle\AdminBundle\Controller\AdminController;
 use Pimcore\Config;
 
+/**
+ * @internal
+ */
 class ReportsControllerBase extends AdminController
```
brusch, line 23:
```suggestion
abstract class ReportsControllerBase extends AdminController
```

bundles/AdminBundle/Controller/Reports/SettingsController.php
```diff
@@ -22,6 +22,8 @@
 
 /**
  * @Route("/settings")
+ *
+ * @internal
  */
 class SettingsController extends ReportsControllerBase
```
brusch, line 28:
```suggestion
final class SettingsController extends ReportsControllerBase
```

bundles/AdminBundle/EventListener/CustomAdminEntryPointCheckListener.php
```diff
@@ -24,6 +24,9 @@
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\HttpKernel\KernelEvents;
 
+/**
+ * @internal
+ */
 class CustomAdminEntryPointCheckListener implements EventSubscriberInterface
```
brusch, line 30:
```suggestion
final class CustomAdminEntryPointCheckListener implements EventSubscriberInterface
```

brusch (changes requested):

brusch (approved):

roland4432:
What's the reason for all classes to be marked as final?

brusch:
@roland4432 Because they are not intended to be overridden or customized. The BC promise / SemVer doesn't apply to those classes.

roland4432:
Could you maybe leave that decission to the programmer?
This breaks dozens of projects and doesn't make any sense in the symfony flow at all.

And.. it doesn't even have the desired effect, since you can still override the controller just not extend it.

roland4432:
I am interested what top contributors like @BlackbitNeueMedien @dpfaffenbauer @jdreesen think about that...

BlackbitNeueMedien:
I never understood the argument to make classes final. Arguments like 
> Using the class for something else by overriding part of its behavior may jeopardize its integrity. The state of the extended object may become unpredictable or inconsistent with the state of the original object.
from https://matthiasnoback.nl/2018/09/final-classes-by-default-why/

are simply not true. If you want to protect internal state of properties, then it is fine for me to make them private. But to prohibit extending a class will only lead to a lot of copy-paste. There are a lot of use-cases where reusing / extending the AdminBundle controllers will on the one hand make custom features easier to implement and on the other hand make them easier to migrate. Because what will happen is that all the custom features which currently extend the AdminBundle controllers which @roland4432 mentions will have to copy the methods from the original classes to still work. Then this at first will work but when those AdminBundle classes get changed, the copied methods might again not work. Of course now you will say that developers have to check their code anyway for Pimcore updates - but what is the advantage of making those classes final?

brusch:
For me it would be ok as well to just tag them as `@internal`, but that would still mean that we're not maintaining backward-compatibility for those classes - "use at your own risk".

roland4432:
Would be totally fine with that.

PR #8477, opened 2021-03-16
lazy load fix when using localized field

hstjep:
Please review the fix for the lazy loading issue inside localized fields. 

After lazy loaded fields are removed on serialization for cache, the lazy loaded fields are now also unmarked as loaded.

## Additional info  
There is an issue with lazy loading when using localized fields.

Steps to reproduce:
1. On a class, add "one-to-many" relation type under localized fields.
2. When the relation field is requested for the first time, the value gets lazy loaded.
3. When it is called the second time in a row, it returns null.

What happens is that after the object is loaded from the database and its lazy loaded values are retrieved, the object gets serialized for cache.

In Pimcore\Model\DataObject\Localizedfield.php, inside __sleep() method, all lazy loaded fields are unset before writing to cache.
However, there is also $this->loadedLazyKeys array (inside LazyLoadedRelationTrait) that still keeps the key marked as loaded.

That causes the issue on the next object load, because the object is pulled from cache (without lazy loaded fields), and $this->isLazyKeyLoaded (used in loadLazyField method inside Localizedfield.php) returns true because the key stayed marked as loaded. Therefore, the localized field value is not fetched from the database and it returns null.

The proposed solution would be to also unmark lazy loaded key as loaded, after the lazy loaded values are removed on serialization for cache. When that's done, $this->isLazyKeyLoaded will return false and the lazy loaded field will be loaded when requested.

### DIFF
--- a/models/DataObject/Localizedfield.php
+++ b/models/DataObject/Localizedfield.php
@@ -385,7 +385,7 @@ class Localizedfield extends Model\AbstractModel implements
      */
     private function loadLazyField(Model\DataObject\ClassDefinition\Data $fieldDefinition, $name, $language)
     {
-        $lazyKey = $name . LazyLoadedFieldsInterface::LAZY_KEY_SEPARATOR . $language;
+        $lazyKey = $this->buildLazyKey($name, $language);
         if (!$this->isLazyKeyLoaded($lazyKey) && $fieldDefinition instanceof Model\DataObject\ClassDefinition\Data\CustomResourcePersistingInterface) {
             $params['language'] = $language;
             $params['object'] = $this->getObject();
@@ -577,7 +577,7 @@ class Localizedfield extends Model\AbstractModel implements
         $forceLanguageDirty = false;
         $isLazyLoadedField = ($fieldDefinition instanceof LazyLoadingSupportInterface || method_exists($fieldDefinition, 'getLazyLoading'))
                                     && $fieldDefinition->getLazyLoading();
-        $lazyKey = $name . LazyLoadedFieldsInterface::LAZY_KEY_SEPARATOR . $language;
+        $lazyKey = $this->buildLazyKey($name, $language);
 
         if ($isLazyLoadedField) {
             if (!$this->isLazyKeyLoaded($lazyKey)) {
@@ -636,7 +636,7 @@ class Localizedfield extends Model\AbstractModel implements
                 foreach (Tool::getValidLanguages() as $language) {
                     unset($this->items[$language][$fieldName]);
 
-                    $lazyKey = $fieldName . LazyLoadedFieldsInterface::LAZY_KEY_SEPARATOR . $language;
+                    $lazyKey = $this->buildLazyKey($fieldName, $language);
                     $this->unmarkLazyKeyAsLoaded($lazyKey);
                 }
             }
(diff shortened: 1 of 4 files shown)


brusch:
@hstjep Thanks a lot for your PR, however, since this is a bug-fix it should go into the bug-fix branch `6.8`. Can you please rebase? Thanks!

dvesh3:
@hstjep Cannot reproduce on 6.8.10. could you please share exact steps to reproduce the issue(is it Admin issue or API) ? in case of API, please share code snippet. thanks!

Also, try to reproduce it on https://demo.pimcore.fun/admin

hstjep:
The case seems to be a little bit more complicated. Here's more details on how to reproduce it using the PHP API on the demo version:

1. In Pimcore administration, create a class and add relation type under the localized fields. For example, Configuration class with a relation for an email document:

image

2. In Pimcore administration, create an object instance of the created class and set its email relation field.

image

3. Add that object under the Website Settings (important).

image

4. Create some Symfony route and, using the PHP API, try calling methods in the following order:

```
/**
* @Route("/api/test", name="test")
*/
public function testAction(Request $request)
{
      // Just saving ANY object (seems to clear WebsiteSettings from cache)
      $news = News::getById(773);
      $news->save();
    
      /* Fetching entire WebsiteSettings. If not cached, it loads the data from db and serializes all the values for cache
        (including our previously added Configuration object).
        On serialization, __sleep() method gets called in \Pimcore\Model\DataObject\Localizedfield.php 
        and removes lazy loaded fields before caching. */
      $websiteSettings = \Pimcore\Config::getWebsiteConfig();
    
      // After that, lazy loaded value not available
      $lazyLoadedEmailDocument = $websiteSettings->get('configuration')->getSomeEmailDocument();
    
      // Same thing can be reproduced if fetching only that setting by name
      $lazyLoadedEmailDocument = WebsiteSetting::getByName('configuration')->getData()->getSomeEmailDocument();

      // Or if the object is fetched by id, after the website settings are loaded before.
      // $configuration = Configuration::getById(1139);
      // $lazyLoadedEmailDocument = $configuration->getSomeEmailDocument();

      return new Response($lazyLoadedEmailDocument->getId());
}
```

Try calling api route few times, and you should be able to reproduce it.

Therefore, if_sleep() method inside Localizedfield.php needs to unset lazy loaded fields, it seems like the simplest solution would be to also unmark them as loaded. So, the $this->items and $this->loadedLazyKeys array (inside LazyLoadedRelationTrait) are in sync. Then, on the next object load, $this->isLazyKeyLoaded will return false and the lazy loaded field will be loaded (if requested).

Let me know what you think.

models/DataObject/Localizedfield.php
```diff
@@ -635,6 +635,9 @@ public function __sleep()
             foreach ($lazyLoadedFields as $fieldName) {
                 foreach (Tool::getValidLanguages() as $language) {
                     unset($this->items[$language][$fieldName]);
+
+                    $lazyKey = $fieldName . LazyLoadedFieldsInterface::LAZY_KEY_SEPARATOR . $language;
```
dvesh3, line 639: please refactor `$lazykey` generation to a separate method since used at many places. thanks!

hstjep:
Added separate method to the LazyLoadedRelationTrait where the other lazy load methods are placed, then used where needed. Also, added signatures to the LazyLoadedFieldsInterface.

dvesh3:
@hstjep Adding signature to LazyLoadedFieldsInterface will lead to BC  break. please revert it and let's just keep the method in trait. thanks!

hstjep:
ok, signatures removed.

dvesh3 (approved):

PR #8500, opened 2021-03-18
[Document] Fixed Symfony 5 compatibility / template reference for areabricks

brusch:


lib/Document/Editable/EditableHandler.php
```diff
@@ -342,13 +342,17 @@ protected function buildBrickTemplateReference(TemplateAreabrickInterface $brick
     {
         if ($brick->getTemplateLocation() === TemplateAreabrickInterface::TEMPLATE_LOCATION_BUNDLE) {
             $bundle = $this->bundleLocator->getBundle($brick);
+            $bundleName = $bundle->getName();
+            if ('Bundle' === substr($bundleName, -6)) {
```
aarongerig, line 346:
```suggestion
            if (str_ends_with($bundleName, 'Bundle')) {
```

see https://www.php.net/manual/de/function.str-ends-with.php

brusch: Thanks! 👍
Just copied it from there: https://github.com/symfony/symfony/blob/5.x/src/Symfony/Bundle/TwigBundle/DependencyInjection/TwigExtension.php#L176 😁

aarongerig: You're welcome. I thought since Pimcore X has the PHP8 requirement, it should be fine to use this method. :)

PR #8502, opened 2021-03-18
[DataObjects] Fix Encrypted fields with Ownerawareinterface leads to recursion on save

### LINKED ISSUE #8501: [DataObjects] Encrypted fields with ownerawareinterface leads to recursion on save
# Bug Report
when saving an object with encrypted field having "_owner" property, it leads to recursion. see https://github.com/pimcore/pimcore/blob/fb48e8b95b53fa0cd8efa33de36bea01705210c5/models/DataObject/Data/EncryptedField.php#L104

dvesh3:
Resolves #8501

### DIFF
--- a/models/DataObject/Data/EncryptedField.php
+++ b/models/DataObject/Data/EncryptedField.php
@@ -136,7 +136,7 @@ class EncryptedField implements OwnerAwareFieldInterface
                 $data = Serialize::unserialize($data);
 
                 if ($data instanceof OwnerAwareFieldInterface) {
-                    $data->setOwner($this, '_owner');
+                    $data->setOwner($this->_owner, $this->_fieldname);
                 }
 
                 $this->plain = $data;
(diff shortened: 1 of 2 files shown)


models/DataObject/Data/EncryptedField.php
```diff
@@ -129,6 +134,11 @@ public function __wakeup()
                 $data = Crypto::decrypt($this->encrypted, $key, true);
 
                 $data = Serialize::unserialize($data);
+
+                if ($data instanceof OwnerAwareFieldInterface) {
+                    $data->setOwner($this, '_owner');
```
dvesh3, line 139: this seems wrong as owner should be object not encryptedField..

dvesh3: used `$this->_owner` as encrypted field is just for delegation.

weisswurstkanone (approved):

PR #7101, opened 2020-09-14
[Data Object] unpublish-publish-permission-fix

fabfischer:
Currently it is possible that a user with the permission "save" can unpublish or publish an object even though the user explicitly has no permission to do so. 

When right-clicking in the tree, the options "publish" and "unpublish" may not appear, but the user could currently unpublish a DataObject with the following (exemplary) cURL request:

```
curl 'http://localhost:1337/admin/object/save?task=unpublish' \
  -X 'PUT' \
  -H 'Connection: keep-alive' \
  -H 'Pragma: no-cache' \
  -H 'Cache-Control: no-cache' \
  -H 'X-Requested-With: XMLHttpRequest' \
  -H 'X-pimcore-csrf-token: TOKEN' \
  -H 'User-Agent: an agent' \
  -H 'DNT: 1' \
  -H 'Content-Type: application/x-www-form-urlencoded; charset=UTF-8' \
  -H 'Accept: */*' \
  -H 'Origin: http://localhost:1337' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Referer: http://localhost:1337/admin/?_dc=1600084528&perspective=' \
  -H 'Accept-Language: en-GB,en;q=0.9' \
  -H 'Cookie: PHPSESSID=session_id; pimcore_admin_sid=1' \
  --data-raw 'id=123' \
  --compressed
```

result:
```
{
   "success":true,
   "general":{
      "o_modificationDate":1600086362,
      "versionDate":1600086362,
      "versionCount":1
   },
   "treeData":{
      "id":123,
      "type":"object",
      "path":"\/example\/object",
      "basePath":"\/example\/",
      "elementType":"object",
      "locked":false,
      "lockOwner":false,
      "isTarget":true,
      "allowDrop":true,
      "allowChildren":true,
      "leaf":false,
      "cls":"pimcore_class_icon pimcore_unpublished  ",
      "published":false,
      "className":"EXAMPLECLASS",
      "allowVariants":null,
      "icon":null,
      "iconCls":"pimcore_icon_object",
      "qtipCfg":{
         "title":"ID: 123",
         "text":"Type: EXAMPLECLASS"
      },
      "expanded":true,
      "permissions":{
         "save":true,
         "unpublish":false,
         "lEdit":false,
         "lView":false,
         "layouts":false,
         "list":false,
         "view":false,
         "publish":false,
[...]

PR #8503, opened 2021-03-18
Bugfix/dataobject unpublish publish permission fix

AlexanderHeidrich:
The DataObjectController::saveAction() will now throw errors when unpublishing/publishing is not allowed.
Otherwise the DataObject will get crossed out while no publishing/unpublishing is actually done.

## Additional info  
If a customer wants to publish/unpublish an object via context menu, where he/she doesn't has any permissions, the saveAction() will end up in the final elseif-Condition `} elseif ($object->isAllowed('save')) {`.
Therefor even if the user has no permissions, a new (per default) unpublished version is stored and the response `treeData` contains `published: false`.

This fix is an adaption of https://github.com/pimcore/pimcore/pull/7101/files


dvesh3:
Hi @AlexanderHeidrich since this is a bug fix, please rebase the PR on 6.8 branch and then change base target here. thanks!

PR #8545, opened 2021-03-23
Model::getById() ::getByPath() ... shouldn't catch any exceptions

weisswurstkanone:
Follow up to https://github.com/pimcore/pimcore/issues/7322

### DIFF
--- a/models/DataObject/Concrete/Dao.php
+++ b/models/DataObject/Concrete/Dao.php
@@ -43,10 +43,13 @@ class Dao extends Model\DataObject\AbstractObject\Dao
         $this->inheritanceHelper = new DataObject\Concrete\Dao\InheritanceHelper($this->model->getClassId());
     }
 
+
     /**
      * Get the data for the object from database for the given id
      *
      * @param int $id
+     *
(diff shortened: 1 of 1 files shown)


models/DataObject/Concrete/Dao.php
```diff
@@ -50,19 +50,15 @@ public function init()
...
-            Logger::warning($e);
+        $data = $this->db->fetchRow("SELECT objects.*, tree_locks.locked as o_locked FROM objects
+            LEFT JOIN tree_locks ON objects.o_id = tree_locks.id AND tree_locks.type = 'object'
+                WHERE o_id = ?", $id);
+
+        if (!empty($data['o_id'])) {
+            $this->assignVariablesToModel($data);
+            $this->getData();
+        } else {
+            throw new \Exception('Object with the ID ' . $id . " doesn't exists");
```
blankse, line 61: Please use `Model\Exception\NotFoundException` and add it to the PhpDocs

weisswurstkanone: done

PR #8546, opened 2021-03-23
Improved \Pimcore\File::getFileExtension() - …

### LINKED ISSUE #8539: Unable to upload assets (pimcore 6.8.11)
# Bug Report

After updating to 6.8.11, I can no longer upload Images to any assets-folder. It looks like pimcore builds a wrong path to the uploaded file, see the screenshot below.

### Expected behavior

It is possible to upload images to an assets folder.

### Actual behavior

pimcore shows an error dialog:

image

### Steps to reproduce

1. install pimcore 6.8.11.
2. drag any image file to the Assets / Home folder.

brusch:
fixes issue with tmp files in certain folders

fixes #8539

### DIFF
--- a/lib/File.php
+++ b/lib/File.php
@@ -39,10 +39,13 @@ class File
     public static function getFileExtension($name)
     {
         $name = strtolower($name);
-        $parts = explode('.', $name);
 
-        if (count($parts) > 1) {
-            return $parts[count($parts) - 1];
+        $pos = strrpos($name, '.');
+        if($pos) {
+            $extension = substr($name, $pos+1);
+            if ($extension && strpos($extension, '/') === false) {
+                return $extension;
+            }
         }
 
         return '';


lib/File.php
```diff
@@ -39,10 +39,13 @@ class File
...
-        $parts = explode('.', $name);
 
-        if (count($parts) > 1) {
-            return $parts[count($parts) - 1];
+        $pos = strrpos($name, '.');
+        if($pos) {
+            $extension = substr($name, $pos+1);
+            if ($extension && strpos($extension, '/') === false) {
+                return $extension;
+            }
```
NiklasBr, line 48: Why does not `pathinfo($path , PATHINFO_EXTENSION)` work here?

brusch: I was not sure if `pathinfo()` works with all the possible (and currently used)  strings which are getting passed to this and I wanted to avoid any breaks here. We should change that in 10 I guess.

Issue #8520, opened 2021-03-22
Classificationsstore - Keys that are deleted are still active in groups.


rndfm:
When keys are deleted they are still visible in the groups tab in the classificationsstore editor.

### Expected behavior
Deleted keys should not be visible. They should behave like they are deleted.

### Actual behavior
Keys are still visible under each group in the groups tab.

### Steps to reproduce
Create a key
Create a group
Associate key to group.
Delete the key.
Observe that key is still visible in the group.

weisswurstkanone:
already fixed

brusch:
fixed in #8579

PR #8579, opened 2021-03-25
[Classificationstore] Deleting key removes it from groups that contai…

BabovicT:
Fix for #8520. Still needs to select other group then first one when deleting key to refresh the group data.

### DIFF
--- a/models/DataObject/Classificationstore/KeyConfig/Dao.php
+++ b/models/DataObject/Classificationstore/KeyConfig/Dao.php
@@ -108,6 +108,12 @@ class Dao extends Model\Dao\AbstractDao
             if (in_array($key, $this->getValidTableColumns(self::TABLE_NAME_KEYS))) {
                 if (is_bool($value)) {
                     $value = (int) $value;
+                    
+                    if(!$value) {
+                        $this->db->delete(
+                            Model\DataObject\Classificationstore\KeyGroupRelation\Dao::TABLE_NAME_RELATIONS,
+                            ['keyId' => $this->model->getId()]);
+                    }
                 }
                 if (is_array($value) || is_object($value)) {
                     if ($this->model->getType() == 'select') {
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Why not add this here ?
https://github.com/pimcore/pimcore/blob/ace9eac8fe8b8ee06d6d5fbed1ec83c13ea3ac2b/models/DataObject/Classificationstore/KeyConfig/Dao.php#L93

weisswurstkanone:
Besides that, could you please rebase this to 6.9 ?

BabovicT:
> Why not add this here ?
> 
> https://github.com/pimcore/pimcore/blob/ace9eac8fe8b8ee06d6d5fbed1ec83c13ea3ac2b/models/DataObject/Classificationstore/KeyConfig/Dao.php#L93

Not sure what You mean. That line would delete key from classificationstore_key table but currently You are soft deleting key with disabling it not removing it from database.

Yea i can rebase it.

weisswurstkanone:
You are right ;-)

Background why this doesn't work anymore: Initially there was no "enabled" column. Since it got introduced key definitions are not deleted anymore (exept via API). Instead they are marked as hidden.

So this constraint doesn't do anything anymore if removed via the user interface.

https://github.com/pimcore/pimcore/blob/6ca8dd8389681f23de81754c5849f548f07b0f1e/bundles/InstallBundle/Resources/install.sql#L850

models/DataObject/Classificationstore/KeyConfig/Dao.php
```diff
@@ -110,6 +110,12 @@ public function update()
             if (in_array($key, $this->getValidTableColumns(self::TABLE_NAME_KEYS))) {
                 if (is_bool($value)) {
                     $value = (int) $value;
+                    
+                    if(!$value) {
```
weisswurstkanone, line 114: However, I think we should check against
$key === "enabled" here

BabovicT: > However, I think we should check against
> $key === "enabled"
> IN ADDITION here

I rebased it to 6.9. Also I added that check You asked.

weisswurstkanone:
thanks a lot!

PR #8585, opened 2021-03-26
Fix undefined getFieldDefinition in ClassRebuildCommand

blankse:
If you have a new class definition with an object brick and execute follow command (without this class definition in the db), you get an exception:
```
php bin/console pimcore:deployment:classes-rebuild -n
```

### DIFF
--- a/bundles/CoreBundle/Command/ClassesRebuildCommand.php
+++ b/bundles/CoreBundle/Command/ClassesRebuildCommand.php
@@ -126,7 +126,11 @@ class ClassesRebuildCommand extends AbstractCommand
                 $output->writeln(sprintf('%s saved', $brickDefinition->getKey()));
             }
 
-            $brickDefinition->save(false);
+            try {
+                $brickDefinition->save(false);
+            } catch (\Exception $e) {
+                $output->write((string)$e);
+            }
         }
 
         if ($output->isVerbose()) {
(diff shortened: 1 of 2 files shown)


bundles/CoreBundle/Command/ClassesRebuildCommand.php
```diff
@@ -126,7 +126,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 $output->writeln(sprintf('%s saved', $brickDefinition->getKey()));
             }
 
-            $brickDefinition->save(false);
+            try {
+                $brickDefinition->save(false);
```
dvesh3, line 130: fix fieldcollection as well?

blankse: I don't think fieldcollection have this problem. There it loads all existing classes from the db.

dvesh3: please see https://github.com/pimcore/pimcore/blob/c62924ca5c3d0c482a71097cb0231aff143208d6/models/DataObject/Fieldcollection/Definition/Listing.php#L27

blankse: This loads fieldcollections and no classes

dvesh3: got it....you mean the problem is with `createContainerClasses()` and since we don't have that for fieldcollection as it is other way to define it in class.

blankse: Yes :)

dvesh3:
@blankse thnx!

PR #8587, opened 2021-03-28
Possibility to overwrite executables with env variables

blankse:
I want set the executables via .env and .env.local files.

Now it is possible to set it via symfony parameters:

```
parameters:
    pimcore_executable_php: '%env(PIMCORE_EXECUTABLE_PHP)%'
```

But if the env doesn't exists, Pimcore now deactivate the executeable and do not check with `whereis`.

I think it would be good if Pimcore itself looks for these env variables.

### DIFF
--- a/lib/Tool/Console.php
+++ b/lib/Tool/Console.php
@@ -76,6 +76,11 @@ class Console
         if (method_exists(__CLASS__, $customSetupMethod)) {
             self::$customSetupMethod();
         }
+        
+        // CHECK for ENV variables
+        if ($value = getenv('PIMCORE_EXECUTABLE_' . strtoupper($name))) {
+            return $value;
+        }
 
         // use DI to provide the ability to customize / overwrite paths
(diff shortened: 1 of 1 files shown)


brusch:
I'd keep it with the parameter configuration as it is but check with `whereis` if the config is empty. 
Otherwise there are again multiple ways how to configure it, what can be confusing.

blankse:
@brusch That would be ok for me. Now it is only disabled if the parameter is false. So it could be set to null and Pimcore looks with whereis.

For example:
```
parameters:
    env(PIMCORE_EXECUTABLE_PHP): ~
    pimcore_executable_php: '%env(PIMCORE_EXECUTABLE_PHP)%'
```

brusch:
Thanks! But maybe a topic for `10.0` as it changes the behavior?

blankse:
@brusch I rebased to master

brusch:
@blankse thanks a lot!

PR #8598, opened 2021-03-29
Cleanup of conditions in Classificationstore's classdefinition

NiklasBr:
- `array_keys($array, true, true)` returns all keys which matches the second argument (`true`)
- `$groupList` got a proper Order Key, a listing without it had no effect as we could see.
- general QoL things like removing unnecessary assignment on `$relation = $relation->load();` and phpdoc cleanup.
- Tooltip fallback with description only if description contains any text, would often end with ugly " - " otherwise.
- Don't call `recursiveGetActiveGroupCollectionMapping()` until required.


weisswurstkanone (approved):

brusch:
Thanks a lot! 👍

PR #8603, opened 2021-03-30
[Grid] Enable inheritance for grid operators

BlackbitNeueMedien:
Currently when you use grid operators you have to manually enable inheritance, otherwise values do not get inherited by parent elements. But for normal fields (without operators) inherited values do get shown by default. 

To reproduce this:
1. Create class Product, enable inheritance, add many-to-many object relation `test`
2. Create an object `/related`
2. Create an object `/parent` and assign `/related` in the relation field `test`
3. Create an object `/parent/child`.
4. Create a PHP class in `src/AppBundle/Operator/Attributes.php` with the follwoing code:
```php
<?php
namespace AppBundle\Operator;

use Pimcore\DataObject\GridColumnConfig\Operator\AbstractOperator;
use Pimcore\DataObject\GridColumnConfig\ResultContainer;
use Pimcore\Model\DataObject;

class Attributes extends AbstractOperator
{
    public function __construct(\stdClass $config, $context = null)
    {
        parent::__construct($config, $context);
    }

    public function getLabeledValue($element)
    {
        $output = implode('<br>', array_map(function($element) {
            return $element->getFullPath();
        }, $element->getTest()));

        $result = new ResultContainer();
        $result->setValue($output);

        return $result;
    }
}
```
4. Open grid, in grid config add field `test` and add operator `PHP code` with PHP class `AppBundle\Operator\Attributes`
The result looks like this:
Bildschirmfoto 2021-03-30 um 12 37 14
You see that although in column `Test` there are related items, in the operator column there is only for the parent element, not for the child.

With this PR the result looks like this:
Bildschirmfoto 2021-03-30 um 12 39 33

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -545,12 +545,14 @@ class Service extends Model\Element\Service
             return null;
         }
 
+        $inheritanceEnabled = AbstractObject::getGetInheritedValues();
+        AbstractObject::setGetInheritedValues(true);
         $result = $config->getLabeledValue($object);
         if (isset($result->value)) {
             $result = $result->value;
 
             if (!empty($config->renderer)) {
-                $classname = 'Pimcore\\Model\\DataObject\\ClassDefinition\\Data\\' . ucfirst($config->renderer);
+                $classname = 'Pimcore\\Model\\DataObject\\ClassDefinition\\Data\\'.ucfirst($config->renderer);
                 /** @var Model\DataObject\ClassDefinition\Data $rendererImpl */
                 $rendererImpl = new $classname();
                 if (method_exists($rendererImpl, 'getDataForGrid')) {
@@ -560,6 +562,7 @@ class Service extends Model\Element\Service
 
             return $result;
         }
+        AbstractObject::setGetInheritedValues($inheritanceEnabled);
 
         return null;
     }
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
mhm, is there an easy way to display the green inheritance indicator as well? 🤔

BlackbitNeueMedien:
@weisswurstkanone It is possible if we add the inheritance status to `ResultContainer` class (or if we add an `inherited` field to those operators which return `stdclass` for `getLabeledValue()` like `Concatenator`). But what to display for example for `Concatenator` operator when one child field is inherited and one not, should inheritance icon be displayed then? Might get even more confusing when the actual origin of the value is not regocnizable at first sight (e.g. when the operator uses a field value to fetch data from elsewhere).

weisswurstkanone:
Sure, for concatenator (and other operators) there is no way. I would leave it this way for now. Only thing is that it is kind of confusing because in your example both columns show exactly the same data 😄 What do you think?

BlackbitNeueMedien:
Actually it was my goal that both columns show the same data because my custom operator does exactly the same as when the relational field itself is used as grid field. So I only wanted to show the difference between native fields using inheritance while operators do not.

weisswurstkanone:
thanks a lot!

PR #8609, opened 2021-03-30
Fix: removeUnusedColumns should invalidate table-columns-cache

youwe-petervanderwal:
Bug: Deleting a object brick field causes an 

> An exception occurred while executing 'ALTER TABLE `object_brick_store_BrickId_ClassId` DROP COLUMN `fieldName`;':  
> SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP 'fieldName'; check that column/key exists

on deployment classes-rebuild

Caused by:
- \Pimcore\Model\DataObject\Objectbrick\Definition\Dao::createUpdateTable() is called twice: first via \Pimcore\Model\DataObject\ClassDefinition\Data\Objectbricks::classSaved(), second via \Pimcore\Model\DataObject\Objectbrick\Definition::save()
- $existingColumnsStore = $this->getValidTableColumns($tableStore, false) returned OLD columns on the second call
- This is due to the fact that the runtime cache wasn't cleared after the drop column


dvesh3:
@youwe-petervanderwal please merge 6.8 branch to fix the test. thanks!

youwe-petervanderwal:
@dvesh3 Rebased, and changed the clear-cache method. `offsetUnset` was causing an exception in PHPUnit if the cache entry wasn't registered - so I've wrapped that call in an `if isRegistered` block

brusch:
Thanks a lot! 👍

PR #8618, opened 2021-03-31
Support dbal 3.0

blankse:


jdreesen:
I think an upgrade to `v2.13` and fixing all deprecations is needed first, see: [Migrate to DBAL 3 (at the bottom of the page)](https://www.doctrine-project.org/2021/03/29/dbal-2.13.html).

This will probably a bit of work, though...

blankse:
@jdreesen It doesn't update to 3.0. Because doctrine/migrations and doctrine/orm is not ready for dbal 3. It't a long way...

brusch:
@blankse yep, that is the reason why we're still on `2` and wasn't changed with #7543 , see also

dpfaffenbauer:
doctrine/orm 2.10 is now compatible with dbal 3: https://github.com/doctrine/orm/releases/tag/2.10.0

brusch:
That's nice, but doesn't help ... `doctrine/migrations` is still not compatible with `3.0`.

PR #8623, opened 2021-03-31
Simplify local phpstan usage

blankse:
After that it is possible to execute the phpstan without setting the PIMCORE_PROJECT_ROOT env. So it can be setup in the PhpStorm inspector.

Before:
```
git clone https://github.com/pimcore/pimcore.git
cd pimcore
.github/ci/scripts/setup-pimcore-environment.sh
cp -rv .github/ci/files/models .
export PIMCORE_PROJECT_ROOT=$(pwd)
composer install
vendor/bin/phpstan analyse -c .github/ci/files/phpstan.actions.neon bundles/ lib/ models/ -l 3
```

After:
```
git clone https://github.com/pimcore/pimcore.git
cd pimcore
.github/ci/scripts/setup-pimcore-environment.sh
cp -rv .github/ci/files/models .
composer install
vendor/bin/phpstan analyse
```

Support also in project mode:
```
git clone https://github.com/pimcore/skeleton.git
cd skeleton
composer install
composer require "phpstan/phpstan:^0.12" "phpstan/phpstan-symfony:^0.12" "elasticsearch/elasticsearch:^7.11" "composer/composer:*"
cp -rv vendor/pimcore/pimcore/.github/ci/files/models/* var/classes/
cp vendor/pimcore/pimcore/doc/Samples/phpstan.neon .
vendor/bin/phpstan analyse
```


.github/ci/files/bin/console
```diff
@@ -16,11 +16,13 @@
 ob_get_clean();
 
 include __DIR__ . "/../vendor/autoload.php";
-\Pimcore\Bootstrap::setProjectRoot();
 
+define('PIMCORE_PROJECT_ROOT', __DIR__ . '/..');
```
brusch, line 20: That might be a problem if you're using custom directories (env variable). 
Why is this necessary, should be covered `\Pimcore\Bootstrap::setProjectRoot()` anyway, right?

brusch (changes requested):

.github/ci/files/bin/console
blankse: setProjectRoot assumes that it is in the vendor folder. When you use this console, you are in the pimcore only scope. I can path setProjectRoot, so that looks after pimcoreOnly/project scope if you prefer there

brusch: Oh, my fault, I'm so sorry, it's only for CI Actions, so it's of course fine 😊

PR #8626, opened 2021-03-31
[Reports] Exporting reports with > 5000 items did not work

BlackbitNeueMedien:
Currently reports with more than 5000 items cannot be exported. In the resulting file are maximum of 5000 rows. Reason is that in https://github.com/pimcore/pimcore/blob/ef8ad357b3de0c56b30b4a77e67e3d1ffc69ee25/bundles/AdminBundle/Controller/Reports/CustomReportController.php#L437 only the basename gets transfered to the client but in the next batch `$exportFile` was not prepended with `PIMCORE_SYSTEM_TEMP_DIRECTORY`:
https://github.com/pimcore/pimcore/blob/ef8ad357b3de0c56b30b4a77e67e3d1ffc69ee25/bundles/AdminBundle/Controller/Reports/CustomReportController.php#L416-L421

This bug was introduced in 1786bdd4962ee51544fad537352c2b4223309442


dvesh3:
@BlackbitNeueMedien thanks!

PR #8633, opened 2021-04-01
Allows pimcore:thumbnails:image to process thumbnails in more than one parent

NiklasBr:


bundles/CoreBundle/Command/ThumbnailsImageCommand.php
```diff
@@ -42,7 +42,7 @@ protected function configure()
                 'parent',
                 'p',
                 InputOption::VALUE_OPTIONAL,
-                'only create thumbnails of images in this folder (ID)'
+                'only create thumbnails of images in this folder (comma separated ID:s eg 543,1077)'
```
dvesh3, line 45:
```suggestion
                'only create thumbnails of images in this folder (comma separated IDs eg. 543,1077)'
```

brusch:
Thanks a lot! 🙏

Issue #8636, opened 2021-04-01
[Ecommerce] Invalid VoucherTokenInformationDetails for cart vs. product context


jremmurd:
# Bug Report

**Issue**
The `getPricingManagerTokenInformationDetails` method in the `VoucherServiceInfercae` returns pricing rules, which contain a vouchercondition for tokens, but where not applied due to other failing conditions in the same rule. This logic is based on the `getAppliedPricingRules` method of the `CartPriceCalculator`, which does not distinguish between applied rules on cart or products. This is problematic in some cases.

**Example**
I.e. if I have products with regular and sales-prices in my shop. Then I create a rule with a birthday vouchertoken which should apply a 10% discount on the cart, but only if there are no products with sales prices in the cart.

In this case the rule would be considered as `applied` (`getNotAppliedRules()` in `VoucherTokenInformation` is empty) as soon as any none-sale product is in the cart. The token does not result in any action, because only a cart-action is configured, even though it is considered applied.

**Possible solutions**
1. I gues the simplest solution would be to add a context to the `getAppliedPricingRules()` method, so you can decide if you want to check for cart or product actions.
2. An even better approach could be to check the applied actions of a rule automatically, it would be necesry to add a method to identify in which context(s) an action gets applied to the `ActionInterface`.

What do you think?

fashxp:
fixed via https://github.com/pimcore/pimcore/issues/8647

PR #8641, opened 2021-04-01
[Ecommerce] Only consider rules as applied on items if product actions exist

jremmurd:
Partially fixes https://github.com/pimcore/pimcore/issues/8636

### DIFF
--- a/bundles/EcommerceFrameworkBundle/CartManager/CartPriceCalculator.php
+++ b/bundles/EcommerceFrameworkBundle/CartManager/CartPriceCalculator.php
@@ -407,6 +407,10 @@ class CartPriceCalculator implements CartPriceCalculatorInterface
             }
         }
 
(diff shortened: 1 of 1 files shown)


fashxp:
fine for now, for Pimcore X we should find a more generic solution. see #8647

fashxp:
on second thoughts ... this might break custom rules, that do changes on product prices

jremmurd:
I see your point, it can be problematic, if there are custom actions on products which are not considered in the `hasProductActions()` method of the `Rule` class. 

To me it is still a bug, which even can be easily overlooked, if you do not test your pricing rules containing voucher conditions separatly for product and cart actions well enough.

fashxp:
agree, this is a bug ... but right now I don't see a possibility to fix this bug without introducing other possible problems. 
solution would be #8647 what we will introduce in Pimcore X

PR #8645, opened 2021-04-01
Allow tags to be dropped to other tags, that don't have children yet

roland4432:
With the current settings one can only drop tags to an element if it already has children. Setting to _false_ will allow to set it as children as well.


brusch:
Thanks a lot!

PR #8656, opened 2021-04-02
BC layer for version unserialization of geopoint

weisswurstkanone:


models/Version.php
```diff
@@ -360,6 +361,8 @@ public function loadData($renewReferences = true)
         }
 
         if ($this->getSerialized()) {
+            class_alias(GeoCoordinates::class, 'Pimcore\Model\DataObject\Data\Geopoint');
```
brusch, line 364:
```suggestion
            // this makes it possible to restore data object versions from older Pimcore versions
            class_alias(GeoCoordinates::class, 'Pimcore\Model\DataObject\Data\Geopoint');
```

PR #8661, opened 2021-04-02
Follow-up on PR #8598 (Classificationstore changes)

NiklasBr:
Turns out my cleanup prior to committing my [previous changes](https://github.com/pimcore/pimcore/pull/8598) were too hasty and introduced a bug.

`$activeGroups` should keep the format `[id => bool]` in  `recursiveGetActiveGroupsIds()` previous change incorrectly changed it to `[int => id]`, see:
image

## Additional info  

Might actually need a new 6.8.x release as previous incorrect change makes editing Classification Stores hard in the admin UI as not all groups were properly returned.


NiklasBr:
@weisswurstkanone Should I make this to 6.8 as well?

weisswurstkanone:
6.8 is closed already

NiklasBr:
Unfortunate but ok.

PR #8662, opened 2021-04-02
Fixes for object merger

weisswurstkanone:


weisswurstkanone:
steps to reproduce:
compare /Product Data/Cars/jaguar/E-Type with a color variant.

PR #8664, opened 2021-04-02
fixed undefined index notice in Asset::getFrontendFullPath

weisswurstkanone:


brusch:
This value should be always set, as there is a default value configured... 🤔

weisswurstkanone:
ok, then maybe a misconfiguration on my instance

brusch:
See https://github.com/pimcore/pimcore/blob/2fabce94fa9a9a13a68a9013afe183ec302bdb24/bundles/CoreBundle/DependencyInjection/Configuration.php#L406 
So should be fine. 
Please re-open if problem persists. 😊

PR #8672, opened 2021-04-03
Remove `final` keyword in ConnectionFactory.php

NiklasBr:
Fixes PHP Warning: `Warning: Private methods cannot be final as they are never overridden by other classes in /var/www/pimcore/vendor/pimcore/pimcore/lib/Storage/Redis/ConnectionFactory.php on line 112`


brusch:
Makes sense 😉 Thanks a lot!

PR #8683, opened 2021-04-06
Pimcore X Upgrade notes

dvesh3:
Resolves #


doc/Development_Documentation/23_Installation_and_Upgrade/01_System_Requirements.md
```diff
@@ -37,9 +37,8 @@ Both **mod_php** and **FCGI (FPM)** are supported.
 - [graphviz](https://www.graphviz.org/) (for rendering workflow overview)
 
 ### Database Server
-- MariaDB >= 10.2
-- MySQL >= 5.7
-- AWS Aurora (supported versions see MySQL)
```
yariksheptykin, line 42: @brusch , what is the reason for dropping AWS Aurora Support? Is Pimcore X not compatible with Aurora any longer? Do you happen to know which feature / commit breaks the compatibility?

brusch: @yariksheptykin Aurora doesn't support MySQL 8 yet, that's the reason, so you have to use RDS in the meantime. The announced already MySQL 8 support, but it seems it's still not rolled out yet.

yariksheptykin: Thanks for the quick reply!

PR #8684, opened 2021-04-06
Fix PHPStan level 4 errors

blankse:
Before:
```
 [ERROR] Found 717 errors 
```

After:
```
[ERROR] Found 105 errors
```


weisswurstkanone:
Brilliant. Thanks a lot!

brusch:
@blankse wow, well done - only 105 left, maybe we can even bring it down to 0 for the launch of X - that would be great 😍

PR #8686, opened 2021-04-06
[Document version preview] Prevent error when xvfb-run is not available

BlackbitNeueMedien:
WIthout this PR you get the error `Warning: array_merge(): Expected parameter 1 to be an array, string given` when trying to display a document version.


brusch:
Thanks a lot! 👍

PR #8691, opened 2021-04-06
style: remove unnecessary space in export filename

aarongerig:
**Filename before:**
`export_ website_translations.csv`

**Filename after:**
`export_website_translations.csv`


weisswurstkanone:
thx!

PR #8671, opened 2021-04-02
Fix classificationstore bugs

skoch98:
There are two bug-fixes in this Pull-Request. Next time I'll split that into two PRs - sorry!

### Description
The selected fields of a classificaton store are not all included in the GRID view.

> Bugfixes need a short guide how to reproduce them -> target branch latest maintenance branch, e.g. `6.6`

The bug exists in Pimcore X (10) and also in Pimcore 6. How you will handel it?

### Steps to reproduce
1. Create a Group and assign more than 15 keys:
Group Relations

2. Select more than 15 keys in the GRID-View relation panel:
Key-Selection

3. Only 15 keys are included in the GRID view
only 15

### Description
The select field in the class definition shows options but without a name.

The bug only exists in Pimcore X (10). The serializer called by "adminJson" can't handle the "StoreConfig"-DataObject.
The serializer could be causing problems at other places?

### Steps to reproduce
1. Create one ore more classificationstores
2. Place the data field "Classfication Store" in a class definition
3. Scroll down in the field definition to "Store" option
image

After save and reopen the class definition the extjs id is displayed:
image

PR #8693, opened 2021-04-06
Fix grid-view classificationstore limit

skoch98:
Fixes part 1 of #8671

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
@@ -788,6 +788,11 @@ class ClassificationstoreController extends AdminController implements EventedCo
         $limit = 15;
         $orderKey = 'name';
         $order = 'ASC';
+        $relationIds = $request->get('relationIds');
+
+        if ($relationIds) {
+            $relationIds = json_decode($relationIds, true);
+        }
 
         if ($request->get('dir')) {
             $order = $request->get('dir');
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
```diff
@@ -817,6 +826,7 @@ public function relationsActionGet(Request $request)
         if ($limit > 0) {
```
weisswurstkanone, line 826: why not move this up to the $request->getLimit path (line 815) ?

bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
```diff
@@ -788,13 +788,19 @@ public function relationsActionGet(Request $request)
         $limit = 15;
```
weisswurstkanone, line 788: not sure if we even this, IMHO this should work without the default. wdyt

bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
skoch98: I dont understand. The limit will only be set if it is greater than 0.

`$list` does not exist in line 815 because it is initialized in line 824.

weisswurstkanone: as far as I see limit is set to 15 by default in line 788 and only overwritten if passes as a request parameter. 
but let's leave it as it is for, should be good enough ;-)
any objections ?

bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
skoch98: I have no idea what the background idea of this limit is. I would leave it in for safety, unless you know exactly that it can not cause problems somewhere else ;-)

bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
skoch98: Nope. Refactoring potential exists in some places 🤔 
But you can't change too much at once with so many contributors. Because there are usually merge conflicts.

Otherwise I should already send a few PSR12 PRs ^^

weisswurstkanone:
thanks a lot!

PR #8695, opened 2021-04-06
[Data Objects] Allow HTML-returning path formatter for many to one relation

BlackbitNeueMedien:
Currently the assigned object of a many-to-one relation is displayed in a text field. This PR changes it to a [`DisplayField`](https://docs.sencha.com/extjs/6.2.0/classic/Ext.form.field.Display.html). This has the advantage that you can use a path formatter which returns HTML:
Bildschirmfoto 2021-04-06 um 15 45 54

As far as I know the capabilities of a text field are not used anyway (e.g. direct typing is disabled).


brusch:
Makes absolutely sense, thanks!

PR #8699, opened 2021-04-07
Increase imagick minimum suggested version

NiklasBr:
PHP 7 support was [added in version 3.4.0](https://github.com/Imagick/imagick/blob/master/ChangeLog#L187..L189), any earlier version requirement should be moot


brusch:
Thanks a lot!

PR #8701, opened 2021-04-07
[Data Objects] - DAO - Prevent (array) access on empty $data

roland4432:
Return directly after SELECT statement (if no result is found) to prevent unnecessary calls on empty $data (especially the $data[$key] in former line 179 which will lead to exceptions)


brusch:
Makes absolutely sense, thanks a lot!

PR #8705, opened 2021-04-07
[Admin] Fix Validation "message" is empty in error response

### LINKED ISSUE #8692: Validation "message" is empty in error response
### Expected behavior
Exception "message" is returned in data array

### Actual behavior
message is empty, because "exception getSubItems" is empty, we use it without any subitems

### Steps to reproduce
throw an validation exception in controller action or somewhere else

### Code example
```php
throw new ValidationException('hallo 123');
```

dvesh3:
Resolves #8692

### DIFF
--- a/bundles/AdminBundle/EventListener/AdminExceptionListener.php
+++ b/bundles/AdminBundle/EventListener/AdminExceptionListener.php
@@ -67,11 +67,6 @@ class AdminExceptionListener implements EventSubscriberInterface
                 }
             }
 
-            $data = [
-                'success' => false,
-                'message' => $message,
-            ];
-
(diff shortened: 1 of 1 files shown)


weisswurstkanone (approved):

blankse:
@dvesh3 Why not only remove line 84? Now trace, traceString and type is removed. Is that intended? And the $data array is not initiated.

dvesh3:
@blankse good catch! fixed in a follow-up commit https://github.com/pimcore/pimcore/commit/3ac8daa15323ab24720c96983ee2d9dce0a579ed thanks! 👍

PR #8721, opened 2021-04-08
Version.php -> class_alias GeoPoint?

### LINKED ISSUE #8718: Version.php -> class_alias GeoPoint?
Not sure why, but here:

https://github.com/pimcore/pimcore/blob/master/models/Version.php#L365

You make a class_alias for GeoPoint to the same Class Name which triggers the following error:

```
Warning: Cannot declare class Pimcore\Model\DataObject\Data\Geopoint, because the name is already in use in vendor/pimcore/pimcore/models/Version.php line 365
```

weisswurstkanone:
resolves #8718

### DIFF
--- a/models/Version.php
+++ b/models/Version.php
@@ -362,7 +362,7 @@ class Version extends AbstractModel
 
         if ($this->getSerialized()) {
             // this makes it possible to restore data object versions from older Pimcore versions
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
thanks :)

PR #8723, opened 2021-04-08
Allow resetting offset to 0 in Listings

fracsi:
There is currently no way to reset a listing offset to 0.

  
  

Allows resetting offset to 0 in Listing classes.

## Additional info  
If using KnpPaginator, you can set it to fix the pagination if the paginator runs out of bounds. The paginator resets the offset to the last available page. If there is only 1 page, the offset is reset to 0, but `Listing` class only allows numbers greater than 0
.
https://github.com/pimcore/pimcore/blob/055d64a90e8cd145795ed1e28f16c24fc7629c10/lib/Model/Listing/AbstractListing.php#L169-L178


brusch:
Thanks a lot!

PR #8727, opened 2021-04-08
[WebDAV] Enhance file listing performance

BlackbitNeueMedien:
Almost every customer who we tell about Pimcore's WebDAV interface is first impressed but then says it is unusable slow. This PR is a try to improve that.

To check permissions I copied the logic from https://github.com/pimcore/pimcore/blob/4c85c70e172bffef38a14c906013463cd32d6590/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php#L256-L271


models/Asset/WebDAV/Folder.php
```diff
@@ -50,39 +51,56 @@ public function getChildren()
...
      */
     public function getChild($name)
     {
-        $nameParts = explode('/', $name);
-        $name = Element\Service::getValidKey($nameParts[count($nameParts) - 1], 'asset');
         $asset = null;
 
         if (is_string($name)) {
+            $nameParts = explode('/', $name);
+            $name = Element\Service::getValidKey($nameParts[count($nameParts) - 1], 'asset');
```
aarongerig, line 100:
```suggestion
            $name = Element\Service::getValidKey(substr($name, strrpos($name, '/') + 1), 'asset');
```

models/Asset/WebDAV/Folder.php
```diff
@@ -50,39 +51,56 @@ public function getChildren()
...
-            foreach ($this->asset->getChildren() as $child) {
-                if ($child->isAllowed('view')) {
-                    try {
-                        if ($child = $this->getChild($child)) {
-                            $children[] = $child;
-                        }
-                    } catch (\Exception $e) {
-                        Logger::warning($e);
-                    }
+        $db = Db::get();
```
aarongerig, line 54: Use DI here:

```php
Pimcore\Db\ConnectionInterface $connection
```

BlackbitNeueMedien: This is not a symfony service but it gets instantiated in https://github.com/pimcore/pimcore/blob/0c8b8e9ddc97dd5107a0a4f082857da2e651151c/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php#L896

Of course we could change that but as there are a lot of other classes where `Db::get()` is used (like the above mentioned `AssetController`), I do not see a problem that this PR also does.

brusch: I think it's fine for now.

brusch:
@BlackbitNeueMedien but maybe we can avoid copy&paste and put this piece of code into a generic private function?

models/Asset/WebDAV/Folder.php
BlackbitNeueMedien: Never mind, now I provide the parameter separately for `setCondition()`, then we do not have to manually quote it.

BlackbitNeueMedien:
@brusch you mean the code from `AssetController`? A private method would not help here because we could not access this from both places. Possible would be a trait or to add a method in `Asset\Folder` like `getChildrenForUser()`. Or even integrate it into `getChildren()`?

brusch:
@BlackbitNeueMedien oh, sorry, there was a misunderstanding on my side 😊 
My favorite is also `filterAccessibleByUser()` tagged as `@internal` on `Asset\Listing`, can you update your PR accordingly? 
Thanks a lot!

BlackbitNeueMedien:
@brusch I extracted `Asset\Listing::filterAccessibleByUser()`.

brusch:
@BlackbitNeueMedien very well done, thanks a lot! 🙏

Issue #7533, opened 2020-11-13
[Document] Select Editable: Behavior for default values


NiklasBr:
# Bug Report

I just tested what should be the future 6.8.5 release and similar to #6459, any defaultValue is not available unless the panel has been opened and saved, rendering `getData()` useless.

### Expected behavior

`defaultValue` should be made available via `getData()`

### Actual behavior

Default value is not returned.

### Steps to reproduce

This code:
```php
            [
                'type' => 'select',
                'label' => $this->translate('brick.options.label.background-colour'),
                'name' => 'backgroundColorConfig',
                'config' => [
                    'store' => $this->getColors(),
                    'width' => 300,
                    'defaultValue' => '#232736',
                ],
            ]

```

### Code example
This returns `null`
```twig
{% set backgroundColor = pimcore_select("backgroundColorConfig").getData() %}
```

# Feature Request

cadeyrn:
It's more like #6458 than #6459.

brusch:
There we should have the same behavior as with the default values at data objects. 
Since it's a BC break anyway, we can discuss this topic for v7.

NiklasBr:
I do not agree that this should be pushed to Pimcore 7 @brusch, since it was the new dialogue box which introduced the BC break in the first place and fixing this now would solve the BC break. Something which should be done in 6.x.

The primary problem here lies in the fact that previously you had the settings in the editable itself, embedded in the document. So when adding a brick to a document and then you saved the document, the default value would saved and then exposed through the `getData()` method, however, since the settings are not loaded the same way in the dialogue box, even when saving the document the data is no longer exposed through the `getData()` method.

So, now the getData() method was actually changed in 6.x in a breaking way, and I would like it to remain as it was previously: Compatible with the old way. This is why I consider this a bug and not a change to be punted to 7.

brusch:
I do not agree that the editable dialog box introduced a bc break! This is just wrong,  it is a new feature and the behavior is as it is. So changing it would be a bc break.

NiklasBr:
Are we talking about the same thing @brusch? It reads like and feels like you are dismissing my concern without considering or explaining the impact.  _You literally wrote in the documentation_ that they should behave the same way, but in this issue you claim it should be different. I don't understand.

> [...] you can use all supported editables in a structured layout in the context of the current brick. The editing interface is configured by implementing the EditableDialogBoxInterface on your brick class and by providing a simple config array.
>[...]
[...]

PR #8729, opened 2021-04-08
[Editables] - Select - Empty value (even with defaultValue) on initial load

roland4432:
# Bug Report

### Expected behavior
Since 6.8 there is an config for dialog options. A select-type can have a defaultValue via Code which should be accessible right in the twig template right after adding the brick.

### Actual behavior
After adding the brick and accessing the value from the select in the optionDialog, the value will always be null
Only if you reload the current page, the defaultValue will be used if no other value has been set in the meantime.

### Steps to reproduce

add a select type in the dialog option
```php
public function getEditableDialogBoxConfiguration(Document\Editable $area, ?\Pimcore\Model\Document\Editable\Area\Info $info): EditableDialogBoxConfiguration {
        $config
            ->setWidth(700)
            ->setHeight(570)
            ->setReloadOnClose(true)
            ->setItems([
                [
                    'type' => 'select',
                    'label' => 'my cool label',
                    'name' => 'my-identifier',
                    'config' => [
                        'store' => [
                            ['default', 'my-default-label'],
                            ['anothervalue', 'my-other-label'],
                        ],
                        'defaultValue' => 'default',
                    ],
                ],
            ])
        ;

        return $config;
    }
```

```twig
{{ dump(pimcore_select('my-identifier').data) }} {# this will be null right after adding the brick and have the default value after reload / save&reload #}
```

### Steps to solve
Right after the config is set to the select, check if there is a defaultValue coming with the passed array and if so and if the current $text is null, then set the defaultValue to the $text variable

### DIFF
--- a/models/Document/Editable/Select.php
+++ b/models/Document/Editable/Select.php
@@ -90,6 +90,18 @@ class Select extends Model\Document\Editable
     }
 
     /**
+     * @param array $config
+     * @return $this
+     */
+    public function setConfig($config)
+    {
+        $this->config = $config;
+        $this->text = $this->text ?? $this->config['defaultValue'] ?? null;
+
+        return $this;
+    }
+
+    /**
      * @return bool
      */
     public function isEmpty()


brusch:
See #8669
Unfortunately this is a BC break, so we fixed that in `10`. 
Thanks anyway 😊

roland4432:
Sorry to bother, but I don't see where this is BC breaking. That's why I used the _setConfig_ method. This will only be called in Pimcore >= 6.8 and then only be applied if there is a defaultValue. In versions < 6.8 this method will of course not be called, therefore I don't get the BC break issue?

dvesh3:
@roland4432 with your changes, it will not possible to nullify value anymore:

```
'store' => [
      ['default', 'my-default-label'],
      ['anothervalue', 'my-other-label'],
      [null, 'Empty'],
  ],
  'defaultValue' => 'default',
```

roland4432:
Yes it is, since null is the ultimate fallback for the defaultValue

=> $this->text ?? $config['defaultValue'] ?? null;  # <-- this will return null in the almost never to be seen case in reality

so if defaultValue = null, then null will be passed as defaultValue

roland4432:
Ah... and to add again.. 
without my code and null as defaultValue... it's not working either.
So my code doesn't break anything but fixes 99.99% of the issues.

NiklasBr:
I agree. The current behaviour is not logical to me, see issue Select Editable: Behavior for default values (#7533) (from an implementation standpoint). It does not make sense to be required to have two places to define default values which behave subtly different, especially when the documentation says that they ought to behave the same.

roland4432:
I see @dvesh3 point with a non-nullable value as a store key. But I still see more advantages in a functional defaultValue.

Therefore, if you wanna override the default behavior and are aware of the downsides, you can always use the _composer.json_ to override parts of the pimcore bundle:

```
autoload:
    ...
    "exclude-from-classmap": [
      "vendor/pimcore/pimcore/models/Document/Editable/Select.php"
    ],
    "files": [
      "src/AppBundle/Model/Document/Editable/Select.php"
    ],
    ...
```

Then just copy the **exact** file content (yes even with the wrong namespace) from this PR to _/src/AppBundle/Model/Document/Editable/Select.php_ and give it a 
> composer dump-autoload -o

Congratz, you now have a working defaultValue in the optionDialog (except for null being a store key and a !null value being defaultValue)

dvesh3:
Well.. this is more about changing behavior than fixing what you think is not logical. There are existing projects with business logics based on this behavior and changing it means BC BREAK.

The difference is documented here: https://pimcore.com/docs/pimcore/master/Development_Documentation/Documents/Editables/Select.html#page_Configuration

And yes, this is changed on master https://github.com/pimcore/pimcore/pull/8669

roland4432:
Let's agree, that we don't aggree. I can see your point from a platform developer to keep things BC.
I don't insist on merging. I am just giving solutions to anybody that stumbles across the same issue and want to help cope this issue.
As my point is to make things working on the current projects I am on and my change does the trick.

So I am totally fine in not merging ;)

PR #8732, opened 2021-04-08
[Image thumbnails] Do not delete downloaded thumbnails when permanent thumbnail definition is used

BlackbitNeueMedien:
Assume you have created a thumbnail definition and enabled it to be listed as a download option for assets. When you download an asset from the asset detail page in the Pimcore backend, then the thumbnail file gets deleted afterwards. Actually it could be reused if the same thumbnail definition from this asset gets accessed again. But even worse: It gets even deleted if it existed before (e.g. when it got created by a frontend application or whatever called `$image->getThumbnail('name-of-thumbnail-config')`). So this breaks the whole create-once-always-reuse-lifecycle of asset thumbnails.


brusch:
Makes sense, thanks a lot! 😊

PR #8740, opened 2021-04-09
[Quantity Values] Sort by base unit

### LINKED ISSUE #8734: [Quantity Values] Sort by base unit instead of name of specific unit
# Feature Request

Current Quantity Values seem to be sorted alphabetically, this makes sense in most context and is a sensible default. But with quantity values this is less intuitive because related measurements are spread across a spectrum rather than grouped together in a way that makes semantic sense.

Very few people are going to mix millimeters and milliliters and Mega-Ohm in the same input fields. But they are grouped together thanks to the sorting.

image

I suggest the default sorting is based on base unit rather than specific unit.

I can try to make a PR if it is interesting.

NiklasBr:
Resolves #8734 

## Additional info  

Sorry about the duplicate PR:s, totally failed when I tried to rebase from source.


NiklasBr:
Replaced by #8780

PR #8744, opened 2021-04-10
Fix: Empty Bundle List

blankse:
In Skeleton is no Bundle in the ExtensionManager. It should be at least PimcoreEcommerceFrameworkBundle.

PHP 8 has a new tokenizer constant T_NAME_QUALIFIED.


brusch:
Nice, thanks a lot! 🙏

PR #8745, opened 2021-04-10
Fix pimcore_outputfilters_disabled param

blankse:
Noticed by #8684 

?pimcore_outputfilters_disabled=true not work. 

$requestKeys has the param names as values


dvesh3:
thanks!

PR #8753, opened 2021-04-12
Fix Array to string conversion

blankse:
If a bundle throws an exception in a migration, it shows only: "Notice: Array to string conversion"


weisswurstkanone:
thx!

PR #8770, opened 2021-04-12
Placeholder/Twig functionality is missing with PdfReactor

blankse:
It is only in WkHtmlToPdf and in the buildPdfFromString (which is only used in the test page)


brusch:
Thank you very much!

PR #8774, opened 2021-04-12
Save Pimcore Classes in /config/pimcore/classes

### LINKED ISSUE #7980: Save Pimcore Classes in /config/pimcore/classes
# Feature Request
Currently, several configuration files are managed in app/config/pimcore as soon as they are copied there.

Can we apply the same behavior for pimcore classes (data object definitions, custom layouts, etc.)?
Meaning, as soon as app/config/pimcore/classes exists, all definition files and sub directories are written there.

This will simplify deployment scenarios in which the class definitions are part of the repository.

weisswurstkanone:
resolves #7980

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
@@ -235,8 +235,7 @@ final class ClassController extends AdminController implements KernelControllerE
     {
         $customLayout = DataObject\ClassDefinition\CustomLayout::getById($request->get('id'));
         $isWriteable = $customLayout->isWritable();
-        $customLayout = json_encode($customLayout->getObjectVars());
-        $customLayout = json_decode($customLayout, true);
+        $customLayout = $customLayout->getObjectVars();
         $customLayout['isWriteable'] = $isWriteable;
 
         return $this->adminJson(['success' => true, 'data' => $customLayout]);
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
@@ -218,8 +218,7 @@ final class ClassController extends AdminController implements KernelControllerE
         $class = DataObject\ClassDefinition::getById($request->get('id'));
         $class->setFieldDefinitions([]);
         $isWriteable = $class->isWritable();
-        $class = json_encode($class);
-        $class = json_decode($class, true);
+        $class = $class->getObjectVars();
         $class['isWriteable'] = $isWriteable;
 
         return $this->adminJson($class);
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
@@ -679,8 +679,7 @@ final class ClassController extends AdminController implements KernelControllerE
         $fc = DataObject\Fieldcollection\Definition::getByKey($request->get('id'));
 
         $isWriteable = $fc->isWritable();
-        $fc = json_encode($fc);
-        $fc = json_decode($fc, true);
+        $fc = $fc->getObjectVars();
         $fc['isWriteable'] = $isWriteable;
 
         return $this->adminJson($fc);
@@ -1058,8 +1057,7 @@ final class ClassController extends AdminController implements KernelControllerE
         $fc = DataObject\Objectbrick\Definition::getByKey($request->get('id'));
 
         $isWriteable = $fc->isWritable();
-        $fc = json_encode($fc);
-        $fc = json_decode($fc, true);
+        $fc = $fc->getObjectVars();
         $fc['isWriteable'] = $isWriteable;
 
         return $this->adminJson($fc);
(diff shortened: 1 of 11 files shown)


models/DataObject/ClassDefinition.php
```diff
@@ -742,7 +756,13 @@ public function getDefinitionFile($name = null)
             $name = $this->getName();
         }
 
-        $file = PIMCORE_CLASS_DIRECTORY.'/definition_'.$name.'.php';
+        $customFile = PIMCORE_CUSTOM_CONFIGURATION_DIRECTORY . '/classes/definition_'.$name.'.php';
+        if (is_file($customFile)) {
+            return $customFile;
+        } else {
```
blankse, line 762: - No else needed (if has an return)
- return in line 767 is so not reachable

jdreesen: The same applies to the other classes.

models/DataObject/ClassDefinition.php
```diff
@@ -518,7 +521,7 @@ public function generateClassFiles($generateDefinitionFile = true)
 
         $cd .= "}\n";
         $cd .= "\n";
-
+        //TODO here
```
brusch, line 524: @weisswurstkanone ?

models/DataObject/ClassDefinition.php
```diff
@@ -731,6 +734,17 @@ protected function deletePhpClasses()
         @unlink($this->getDefinitionFile());
     }
 
+
+    public function isWritable()
+    {
+        $name = $this->getName();
+        $customFile = PIMCORE_CUSTOM_CONFIGURATION_DIRECTORY . '/classes/definition_' . $name . '.php';
```
brusch, line 741: I'd go for the following here: 
```php
return (strpos($this->getDefinitionFile(), PIMCORE_CUSTOM_CONFIGURATION_DIRECTORY) === 0);
```
This has the benefit that we have the way how the path is being build is just in one place. 
Of course this should be done for the other classes as well.

brusch (changes requested):
@weisswurstkanone The save button in the UI should be grayed out if `isWritable() === false`. 
Please also consider @blankse feedback. 
Thanks!

models/DataObject/ClassDefinition.php
jdreesen: Since we're on PHP 8, `str_starts_with` should be used:
```php
return str_starts_with($this->getDefinitionFile(), PIMCORE_CUSTOM_CONFIGURATION_DIRECTORY);
```

brusch: Absolutely 👍 Thanks! 😊

bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
```diff
@@ -217,6 +217,10 @@ public function getAction(Request $request)
     {
         $class = DataObject\ClassDefinition::getById($request->get('id'));
         $class->setFieldDefinitions([]);
+        $isWriteable = $class->isWritable();
+        $class = json_encode($class);
```
brusch, line 221: Why not `$class->getObjectVars()` or `object2array()` or all together `\object2array($class->getObjectVars())`?

models/DataObject/Objectbrick/Definition.php
```diff
@@ -671,17 +679,32 @@ protected function doEnrichFieldDefinition($fieldDefinition, $context = [])
         return $fieldDefinition;
     }
 
+    public function isWritable()
```
brusch, line 682:
```suggestion
    public function isWritable(): bool
```

models/DataObject/Fieldcollection/Definition.php
```diff
@@ -255,15 +264,28 @@ public function delete()
         }
     }
 
+    public function isWritable()
```
brusch, line 267:
```suggestion
    public function isWritable(): bool
```

models/DataObject/ClassDefinition/CustomLayout.php
```diff
@@ -251,14 +254,28 @@ private function saveCustomLayoutFile($saveDefinitionFile = true)
         }
     }
 
+    public function isWritable()
```
brusch, line 257:
```suggestion
    public function isWritable(): bool
```

brusch (changes requested):
We should also add a proper documentation for this feature: 
https://pimcore.com/docs/pimcore/current/Development_Documentation/Deployment/Deployment_Tools.html#page_Pimcore-Class-Definitions

doc/Development_Documentation/21_Deployment/05_Deployment_Tools.md
```diff
@@ -19,6 +19,9 @@ can be defined.
 As with Pimcore configurations also Pimcore class definitions are saved as PHP configuration files and therefore can 
 be added to version control systems and be deployed to different deployment stages. 
 
+The PHP configuration files will be written to the `var/classes` directory by default. However, you can create a copy
+at `/config/pimcore/classes` which is then read-only.
```
brusch, line 23:
```suggestion
at `/config/pimcore/classes` which is then read-only. The generated classes, e.g. `Pimcore\Model\DataObject\Example` can be placed in the `src/` folder. 
```

lib/File.php
```diff
@@ -137,15 +137,18 @@ public static function put($path, $data)
 
     /**
      * @param string $path
-     * @param mixed $data
+     * @param string $data
+     * @return int
```
blankse, line 141:
```suggestion
     * @return int|false
```

PR #8776, opened 2021-04-12
Fix overwriting where condition in customviews

fracsi:
Allow adding extra where conditions in customviews. (It broke in 6.9, works in 6.8)

## Additional info  
DoctrineQueryBuilder `where` function overrides the complete where condition, thus in customviews you cannot add a simple where condition to the existing one.

Previously in 6.8, in customviews you could add a simple extra filter to the query like `o_published = 1`, so only the published elements would show up in the tree. In 6.9 (using Doctrine Query Builder) currently the whole `where` condition would be only `o_published = 1` and the original where conditions built in the `DataObjectController::treeGetChildsByIdAction` method are lost.


fracsi:
Note: #7954 introduced the bug.

dvesh3:
thanks!

PR #8777, opened 2021-04-12
Ecommerce index service problem with fraction numbers in elastic search

kkarkus:
Let's say we have double type in ES, and some getter on a product returns e.g.: 5.0, but without this flag `JSON_PRESERVE_ZERO_FRACTION`, we'll end up with integer instead after json_encode, and this leads to an error during indexing new products.

This is only the case for numbers like 1.0, 2.0, 3.0, and so on.
Every other fraction number works ok.


fashxp:
thx.

PR #8746, opened 2021-04-11
Use DataObject and ProductListInterface constants

NiklasBr:
## Changes in this pull request  
Resolves no specific issue, only trying to avoid hard-coded strings where possible and the class constants are available.

PR #8781, opened 2021-04-13
Use DataObject and ProductListInterface constants

NiklasBr:
Replaces pull #8746


dvesh3:
@NiklasBr thanks a lot!

PR #8783, opened 2021-04-13
[Image thumbnails] Bugfix: Automatic image clipping by embedded path

BlackbitDevs:
We have a problem with images like the following (with embedded clipping path AND transparency) with enabled `clip_auto_support`
[217001-00-WTAL__ANW-010.tif.zip](https://github.com/pimcore/pimcore/files/6302523/217001-00-WTAL__ANW-010.tif.zip)

WIth current code it looks like this:
Bildschirmfoto 2021-04-13 um 11 09 24

With this PR applied it looks like this:
Bildschirmfoto 2021-04-13 um 11 11 52

This project uses Imagick 3.4.4, and Image Magick 
```
Version: ImageMagick 6.9.7-4 Q16 x86_64 20170114 http://www.imagemagick.org
Copyright: © 1999-2017 ImageMagick Studio LLC
License: http://www.imagemagick.org/script/license.php
Features: Cipher DPC Modules OpenMP 
Delegates (built-in): bzlib djvu fftw fontconfig freetype jbig jng jpeg lcms lqr ltdl lzma openexr pangocairo png tiff wmf x xml zlib
```

When I upload this file to https://demo.pimcore.fun/ the preview looks even worse:
Bildschirmfoto 2021-04-13 um 11 30 41
But with this PR it also works there:
Bildschirmfoto 2021-04-13 um 11 33 01

I am quite sure that this PR is not the solution because I removed code which is actually there to ensure compatiblity with different image magick versions. But perhaps we can find a way to correctly display above image and also keep the current behaviour for other image magick versions?

### DIFF
--- a/bundles/CoreBundle/Resources/config/services.yml
+++ b/bundles/CoreBundle/Resources/config/services.yml
@@ -225,4 +225,6 @@ services:
     Pimcore\Image\Adapter\Imagick:
         shared: false
         public: true
-    Pimcore\Image\Adapter\GD: ~
+    Pimcore\Image\Adapter\GD:
+        shared: false
+        public: true
--- a/bundles/CoreBundle/Resources/config/services.yml
+++ b/bundles/CoreBundle/Resources/config/services.yml
@@ -222,5 +222,7 @@ services:
         arguments:
             $locator: !tagged_locator { tag: flysystem.storage }
 
-    Pimcore\Image\Adapter\Imagick: ~
+    Pimcore\Image\Adapter\Imagick:
+        shared: false
+        public: true
     Pimcore\Image\Adapter\GD: ~
--- a/bundles/CoreBundle/Resources/config/services.yml
+++ b/bundles/CoreBundle/Resources/config/services.yml
@@ -225,6 +225,7 @@ services:
     Pimcore\Image\Adapter\Imagick:
         shared: false
         public: true
+
     Pimcore\Image\Adapter\GD:
         shared: false
         public: true
--- a/bundles/CoreBundle/Resources/config/services.yml
+++ b/bundles/CoreBundle/Resources/config/services.yml
@@ -221,3 +221,6 @@ services:
         public: true
         arguments:
             $locator: !tagged_locator { tag: flysystem.storage }
+
+    Pimcore\Image\Adapter\Imagick: ~
+    Pimcore\Image\Adapter\GD: ~
--- a/lib/Image.php
+++ b/lib/Image.php
@@ -14,6 +14,7 @@
 
 namespace Pimcore;
 
+use Pimcore;
 use Pimcore\Image\Adapter;
 
 class Image
@@ -42,9 +43,9 @@ class Image
     {
         try {
             if (extension_loaded('imagick')) {
-                return new Adapter\Imagick();
+                return Pimcore::getContainer()->get(Adapter\Imagick::class);
             } else {
-                return new Adapter\GD();
+                return Pimcore::getContainer()->get(Adapter\GD::class);
             }
         } catch (\Exception $e) {
             Logger::crit('Unable to load image extensions: ' . $e->getMessage());
--- a/lib/Image/Adapter/Imagick.php
+++ b/lib/Image/Adapter/Imagick.php
@@ -135,27 +135,11 @@ class Imagick extends Adapter
                 $identifyRaw = $i->identifyImage(true)['rawOutput'];
                 if (strpos($identifyRaw, 'Clipping path') && strpos($identifyRaw, '<svg')) {
                     // if there's a clipping path embedded, apply the first one
-
                     try {
-                        // known issues:
-                        // - it seems that -clip doesnt work with the ImageMagick version
-                        //   ImageMagick 6.9.7-4 Q16 x86_64 20170114 (which is used in Debian 9)
-                        // - Imagick 3.4.4 with ImageMagick 7 on OSX has horrible broken clipping support
-                        $i->setImageAlphaChannel(\Imagick::ALPHACHANNEL_TRANSPARENT);
                         $i->clipImage();
                     } catch (\Exception $e) {
                         Logger::info(sprintf('Although automatic clipping support is enabled, your current ImageMagick / Imagick version does not support this operation on the image %s', $imagePath));
                     }
-
-                    // Imagick version compatibility
-                    // Since Imagick 3.4.4 compiled against ImageMagick 7 ALPHACHANNEL_OPAQUE was removed for whatever reason
-                    // ImageMagick is still defining and using OpaqueAlphaChannel in ImageMagick 7 releases
-                    // Let's hardcode the current ImageMagick 7 enum number to workaround this issue
-                    $alphaChannel = 11;
-                    if (defined('Imagick::ALPHACHANNEL_OPAQUE')) {
-                        $alphaChannel = \Imagick::ALPHACHANNEL_OPAQUE;
-                    }
-                    $i->setImageAlphaChannel($alphaChannel);
                 }
             }
         } catch (\Exception $e) {


brusch:
@BlackbitNeueMedien Yep, this fix breaks embedded clipping path in JPEGs. The examples linked here https://github.com/pimcore/pimcore/issues/5505#issue-543824108 for instance are broken after applying this patch.

brusch:
@BlackbitNeueMedien would be nice if you can find a solution that works for both given examples. Thanks in advance!

BlackbitNeueMedien:
What do you think of applying the current solution for JPGs and this PR's solution for all other images? Because actually the current code includes all the workarounds which should not be necessary if ImageMagick wordked correctly. I will try to implement that and test with the images from https://github.com/pimcore/pimcore/issues/5505#issue-543824108

brusch:
@BlackbitNeueMedien This was also my first thought, but I don't know if this is specific to the image format or if this depends on the type and content of the meta-info that contains the clipping path. 🤔

BlackbitNeueMedien:
;-) This is also my concerns. Let's just do it this way and if it does not work, hopefully there will be some example images soon so that we can adjust the implementation.

BlackbitNeueMedien:
Hmm, when I upload the files from https://github.com/pimcore/pimcore/issues/5505#issue-543824108 to https://demo.pimcore.fun/admin (without this PR's changes!) I get this:
Bildschirmfoto 2021-04-14 um 09 42 54

NiklasBr:
@BlackbitNeueMedien can you try to upload that image but with the settings on the thumbnail config set to *preserve metadata* without the changes in this PR?

brusch:
hmmm, interesting, for me they worked, but I was on latest `master`, env: 
```
PHP Version 8.0.3

imagick module version | 3.4.4
-- | --
imagick classes | Imagick, ImagickDraw, ImagickPixel, ImagickPixelIterator, ImagickKernel
Imagick compiled with ImageMagick version | ImageMagick 6.9.10-23 Q16 x86_64 20190101 https://imagemagick.org
Imagick using ImageMagick library version | ImageMagick 6.9.10-23 Q16 x86_64 20190101 https://imagemagick.org
ImageMagick copyright | © 1999-2019 ImageMagick Studio LLC
ImageMagick release date | 20190101


```

BlackbitNeueMedien:
@NiklasBr I am currently only looking at the preview in the Pimcore backend. But I tried it by creating a separate thumbnail definition with output format JPG, background color red and preserve metadata. And that gives me:

brusch:
Same here, just checked the Pimcore preview.

BlackbitNeueMedien:
@brusch: And with this PR it gets wrong on your "Pimcore preview"? Do you mean in https://demo.pimcore.com/admin/ ? Perhaps we have to check for certain image magick versions?

brusch:
@BlackbitNeueMedien yes, with this PR I'm getting a white background for these images https://github.com/pimcore/pimcore/issues/5505#issue-543824108

BlackbitNeueMedien:
Hmm, while `$i->clipImage()` does not work `$i->clipImagePath('#1', false);` works for the images of https://github.com/pimcore/pimcore/issues/5505#issue-543824108 - this is simply a bug in ImageMagick that it removes the wrong side of the clipping the path (the screwdriver instead of the background). For this reason I now use `$i->clipImagePath('#1');` for JPGs. This way it works here for the #5505 images and for the above image. Can you please try on your setup @brusch?

lib/Image/Adapter/Imagick.php
```diff
@@ -143,21 +143,25 @@ public function load($imagePath, $options = [])
                         // - it seems that -clip doesnt work with the ImageMagick version
                         //   ImageMagick 6.9.7-4 Q16 x86_64 20170114 (which is used in Debian 9)
                         // - Imagick 3.4.4 with ImageMagick 7 on OSX has horrible broken clipping support
-                        $i->setImageAlphaChannel(\Imagick::ALPHACHANNEL_TRANSPARENT);
-                        $i->clipImage();
+
+                        if(in_arrayi($this->sourceImageFormat, ['jpeg', 'pjpeg', 'jpg'])) {
```
brusch, line 147:
```suggestion
                        if(in_arrayi($this->sourceImageFormat, ['jpeg', 'pjpeg', 'jpg', 'pjpeg', 'jpe'])) {
```

brusch:
@BlackbitNeueMedien Unfortunately I'm now seeing the same behavior as you had without the patch: 

image

brusch (changes requested):

lib/Image/Adapter/Imagick.php
BlackbitNeueMedien: But this does not solve the wrong display on your system, does it?

brusch: @BlackbitNeueMedien exactly, that was just a missing format I noticed before I did the test 😊
Must be related to the underlying ImageMagick version, because our Imagick (PHP extension) version is actually the same (`3.4.4`). 
So maybe we need to somehow utilize `Imagick::getVersion()` to fix that?

BlackbitNeueMedien: I already tried. The problem is that I have 6.9.7-4, you have 6.9.10-23 and I have also tried it with 7.0.10-25. The problem now is that it works with this PR in 6.9.7-4 and 7.0.10-25 but not in your 6.9.10-23. So it will be difficult to check with `version_compare()`.
So the problem seems to be what @ssiergl said in https://github.com/pimcore/pimcore/issues/5505#issuecomment-673386221 
> Imagemagick quite often breaks clipping. Aside that there are some broken imagemagick 6.9 Versions (like used in debian 9) it is pretty often broken in imagemagick 7 as well (E.g. right now the last eight stable releases are broken again)

brusch: @BlackbitNeueMedien ok, agree. 
I'd then like to propose the following solution: 
For 6.9 we leave everything as it is, so we don't change any behavior. 

For X: 
We implement it as the expected behavior of those IM clipping functions is and we just ignore the bugs of ImageMagick. 
Additionally we refactor `\Pimcore\Image` to return non shared services for either `Imagick` or `GD`. 
Then it's easy to override `Imagick` including some fixes for the individual broken ImageMagick version. 
Wdyt?

BlackbitNeueMedien:
Ok, changed base branch to master and removed the handling of buggy ImageMagick versions.
Also introduced services for `Adapter\ImageMagick` and `Adapter\GD`. Is this what you meant, @brusch ?

bundles/CoreBundle/Resources/config/services.yml
```diff
@@ -221,3 +221,6 @@ services:
         public: true
         arguments:
             $locator: !tagged_locator { tag: flysystem.storage }
+
+    Pimcore\Image\Adapter\Imagick: ~
```
brusch, line 225:
```suggestion
    Pimcore\Image\Adapter\Imagick:
        shared: false
        public: true
```

bundles/CoreBundle/Resources/config/services.yml
```diff
@@ -221,3 +221,6 @@ services:
...
+    Pimcore\Image\Adapter\GD: ~
```
brusch, line 226:
```suggestion
    Pimcore\Image\Adapter\GD:
        shared: false
        public: true
```

brusch (changes requested):
Thanks, that's exactly what i meant. 
I've seen that there's an open conflict and I think it would be better to have the services non-shared.

BlackbitNeueMedien:
There is a conflict? I cannot see it. Also merged master again but only says "Already up to date"

brusch:
Seems that GitHub has some caching issues, gone now, sorry for confusion 😊

bundles/CoreBundle/Resources/config/services.yml
```diff
@@ -221,3 +221,10 @@ services:
         public: true
         arguments:
             $locator: !tagged_locator { tag: flysystem.storage }
+
+    Pimcore\Image\Adapter\Imagick:
+        shared: false
+        public: true
+    Pimcore\Image\Adapter\GD:
```
brusch, line 228:
```suggestion

    Pimcore\Image\Adapter\GD:
```

lib/Image.php
```diff
@@ -42,9 +43,9 @@ public static function create()
     {
         try {
             if (extension_loaded('imagick')) {
-                return new Adapter\Imagick();
+                return Pimcore::getContainer()->get(Adapter\Imagick::class);
```
dvesh3, line 46: this throws exception when running `./vendor/bin/pimcore-install`
image

PR #8797, opened 2021-04-14
System yml cleanup

brusch:


dvesh3:
@brusch we missed https://github.com/pimcore/demo/blob/783fdb07051360d5c1a229d2bc2354849be9b9e8/var/config/system.template.yml

brusch:
@dvesh3 I did changed that 5 days ago on the master: https://github.com/pimcore/demo/blob/master/var/config/system.template.yml so it should be alright 😊

PR #8804, opened 2021-04-14
Fix Ghostscript parameter

joFra210:
Fixes GS command, which could not be executed due to wrong parameter formatting

## Additional info  
To reproduce upload any pdf or other document and try to create a thumbnail


dvesh3:
thanks!

PR #8817, opened 2021-04-16
Revert service introduced for imagick & GD library - fix broken installer

dvesh3:
Regression https://github.com/pimcore/pimcore/pull/8783#discussion_r614763600


brusch:
Created an alternative approach: #8821 
I think it's just not needed to utilize the container for the requirements check.

PR #8818, opened 2021-04-16
Update Thumbnail.php

mediafant:
when I use "lowQualityPlaceholder" in pimcore_image, "srcset" is replaced by "data-srcset". That is correct, but w3c validator is complaining that srcset is not available. For this I think it is better not to delete this attribute but fill it with small content.

  
  

Resolves #

### DIFF
--- a/models/Asset/Image/Thumbnail.php
+++ b/models/Asset/Image/Thumbnail.php
@@ -244,7 +244,7 @@ final class Thumbnail
 
             if (isset($options['previewDataUri'])) {
                 $sourceTagAttributes['data-srcset'] = $sourceTagAttributes['srcset'];
-                unset($sourceTagAttributes['srcset']);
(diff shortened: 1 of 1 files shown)


brusch:
@mediafant to be honest, haven't seen `'data:,1w'` before, where did you get it from and how compatible it is with browsers? Thanks for clarifying 😊

brusch:
@mediafant 🏓😊

mediafant:
Sorry late answer - I've found it in forum, but I haven't seen this solution also never before, but on w3c-validator it looks nice.

mediafant:
@brusch - found source: https://github.com/aFarkas/lazysizes/issues/500

brusch:
@mediafant nice, thanks a lot for clarifying! 👍

brusch:
@mediafant I had to revert this PR, see #9576
So we need to look for an alternative approach which doesn't break the functionality.

PR #8821, opened 2021-04-16
Fixed installer, regression of #8817

brusch:
Alternative approach for / fixes  #8817


brusch:
In my opinion there's no need to get the service from the container. For the requirements check I'd rather initiate the service manually.

PR #8829, opened 2021-04-19
[Data object grid] Prevent ambiguous columns when filtering / sorting by object brick field

BlackbitNeueMedien:
Steps to reproduce problem:
1. Create class `Product`. Add a `select` field `test` and option `A`and `B`. Add an object brick field named `bricks`.
2. Create an object brick `Brick`. Add a `select` field `test` and option `A`and `B`. Assign `Product.bricks` as allowed brick field.
3. Create an object `/A` of class `Product`, assign object brick `Brick`, select `A` in `Brick.test`.
4. Click `Home` in data object tree, click `Grid options` add field `test` from object brick `Brick`
5. Sort by field `test` -> first ambiguous column error appears
6. Filter by field `test=A` -> second ambiguous column error appears

This PR prefixes the sort / filter field names with the brick table alias.

### DIFF
--- a/bundles/AdminBundle/Helper/GridHelperService.php
+++ b/bundles/AdminBundle/Helper/GridHelperService.php
@@ -557,6 +557,9 @@ class GridHelperService
                         $orderKey = $orderKeyParts[0].'.'.$orderKeyParts[1];
                         $doNotQuote = true;
                     }
+                } else {
+                    $orderKey = $list->getDao()->getTableName().'.'.$orderKey;
+                    $doNotQuote = true;
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


dvesh3 (changes requested):
ambiguous column error appears when sorting direct field "test" along with brick field "test"

dvesh3:
@BlackbitNeueMedien please see https://github.com/pimcore/pimcore/pull/8829#pullrequestreview-642034828

BlackbitNeueMedien:
Now filtering and sorting does work for the class field and for the brick field.

dvesh3:
thanks a lot!

PR #8833, opened 2021-04-19
[Grid exports] Correct order of grid export types

BlackbitNeueMedien:
As long as there are only the grid exports for CSV and Excel which are included in Pimcore, the result with and without this PR are the same. But if a bundle adds another grid export type, the result is wrong because 
https://github.com/pimcore/pimcore/blob/f1b0ccb8b3d5c7028416a73df9ea4bab508b165b/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js#L330
always uses the last added export type for the main button. This means currently the last added export type will get on the main export button, all others go into the button dropdown. With this PR the `CSV Export` will always stay on the main button while XLSX and other custom export types go into the button dropdown.


dvesh3:
thanks!

PR #8862, opened 2021-04-22
fix classification store language permisions via workspace

frithjof:
Fixes localized classification store not honoring workspace language settings for users

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -1460,13 +1460,6 @@ class Service extends Model\Element\Service
                     }
                 }
                 if (!($haveAllowedViewDefault && count($allowedView) == 0)) {
-                    /*
-                    $layout->{'permissionView'} = AdminTool::reorderWebsiteLanguages(
-                        AdminTool::getCurrentUser(),
-                        array_keys($allowedView),
-                        true
-                    );
-                    */
                     $layout->setPermissionView(AdminTool::reorderWebsiteLanguages(
                         AdminTool::getCurrentUser(),
                         array_keys($allowedView),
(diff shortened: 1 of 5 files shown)


frithjof:
i am not sure how to fix the phpstan complains about
$layout->{'permissionView'} = AdminTool::reorderWebsiteLanguages(
                        AdminTool::getCurrentUser(),
                        array_keys($allowedView),
                        true
                    );

weisswurstkanone:
please add a protected property + respective getter and setter

frithjof:
if i move them to protected or even define things start breking in js ... so i kept the old logic and adjusted the phpstan ignore for the error that was already there for this

models/DataObject/Service.php
```diff
@@ -1450,7 +1450,7 @@ public static function enrichLayoutDefinition(&$layout, $object = null, $context
      */
     public static function enrichLayoutPermissions(&$layout, $allowedView, $allowedEdit)
     {
-        if ($layout instanceof Model\DataObject\ClassDefinition\Data\Localizedfields) {
+        if ($layout instanceof Model\DataObject\ClassDefinition\Data\Localizedfields || $layout instanceof Model\DataObject\ClassDefinition\Data\Classificationstore && $layout->localized === true) {
```
weisswurstkanone, line 1453: Please try to avoid extending the baseline. Would it be possible to do something like this instead ?
```
$layout = $layout->getObjectVars();
$layout['whateveryKey'] = $wahteverValue;
```

frithjof: i think this is worse because layout is by reference an object there. Also this wont work with protected properties.

I think the whole js handling of localized field permissions has to be adjusted either way or you add it to baseline

Localized Fields was probably in there because of this reason.

I agree that it is not pretty but if i start applying changes there it will be more than just a small bugfix.

weisswurstkanone:
The main reason for getObjectVars is to read protected properties.

See https://github.com/pimcore/pimcore/blob/254d1d7830bac4d6fddc244d931c2f769f4a3c74/models/DataObject/Traits/ObjectVarTrait.php#L25

frithjof:
i had to leave them as public properties like the other properities (title, width, height etc. ) that go into js layout because json encoder wont pick them up and the layout still needs to be an object during the process ... getObjectVars would convert it to an array.

brusch:
@frithjof at least we should tag them as `@internal` - thanks!

PR #8863, opened 2021-04-22
Bugfix for incorrect maxDepth comparison

NiklasBr:

Resolves a bug which would happen in only specific instances.

Consider the following scenario:

```twig
{% set menuOptions = { active: document, maxDepth: 5, expandSiblingNodesOfActiveBranch: true, onlyActiveBranch: true } %}
{{ pimcore_render_nav(pageNavigation, 'menu', 'renderMenu', menuOptions) }}
```

With a document structure that is at least 2 levels deeper than `maxDepth`, example where _Produkter_ is the root document for this menu.

image

This issue seems to be sensitive to cache:ing, so to reproduce it, first click around in the rendered menu to make sure it has been cached. Also good to do this in a clean browser session for some reason.

Then, *Unpublish* the page at the same level as the configured `maxDepth` by right-clicking and using the context menu option. In the screenshot above this would be _BLL produkter_.

Now click around in the menu again. Most pages will work fine until you once again click on the page which in the screen shot corresponds to _Energi_. At which point the rendered HTML will break.

This bug does not manifest if the unpublished page is on any other level than the exact one set in `maxDepth`.

### DIFF
--- a/lib/Navigation/Builder.php
+++ b/lib/Navigation/Builder.php
@@ -332,7 +332,7 @@ class Builder
 
                 $page->setClass($page->getClass() . $classes);
 
-                if ($child->hasChildren() && (!$maxDepth || $maxDepth > $this->currentLevel)) {
+                if ($child->hasChildren() && (!$maxDepth || $maxDepth >= $this->currentLevel)) {
                     $childPages = $this->buildNextLevel($child, false, $pageCallback, $parents, $maxDepth);
                     $page->setPages($childPages);
                 }


brusch:
@NiklasBr Thanks for your PR! We should probably move this to `10` milestone as it changes the behavior, right?

NiklasBr:
I'd consider it a bug in 6.x because without this fix we get HTML which looks like this:

image

brusch:
I agree, thanks a lot @NiklasBr ! 🙏

NiklasBr:
Thanks @brusch!

However a colleague alerted me to another potential side-effect of this change with regards to how many levels are printed.

Might have to change how currentLevel works to something like this instead:
```php
                if ($child->hasChildren() && (!$maxDepth || $maxDepth > $this->currentLevel+1)) {
                    $this->currentLevel++;
                    $childPages = $this->buildNextLevel($child, false, $pageCallback, $parents, $maxDepth);
                    $page->setPages($childPages);
                }
```

brusch:
Hmm, maybe I didn't got the point, but is `$this->currentLevel++;` really necessary? because at the beginning of `buildNextLevel()` it's incremented again.

brusch:
@NiklasBr can you maybe make a follow up PR, that would be very nice 😊

NiklasBr:
@brusch I might indeed have to do that. The odd thing here is that this patch works 100% of the time in my local environment but when we applied it in a Prod environment the issue would still appear. Current working theory is that this fix is maybe not the proper fix but rather just mostly fixed it.

brusch:
@NiklasBr Thanks for your feedback, that sounds very weird 🤪 Maybe a caching issue? 
Maybe you can check that by disabling the cache for the navigation (`cacheLifetime: false`).

brusch:
@NiklasBr Thanks for your feedback, that sounds very weird 🤪 Maybe a caching issue? 
Maybe you can check that by disabling the cache for the navigation (`cacheLifetime: false`).

NiklasBr:
We will do that as well 🤔 meanwhile feel free to revert this patch if you have any doubts about it.

NiklasBr:
Considering making a PR with these changes instead @brusch, they make sure the navigation is rebuild every time I do the following step in the initial description of this issue:

> Then, Unpublish the page at the same level as the configured maxDepth by right-clicking and using the context menu option. In the screenshot above this would be BLL produkter.

It appears to work as well. Thoughts?

```diff
@@ -44,6 +44,8 @@
      */
     private $currentLevel = 0;
 
+    private $navCacheTags = [];
+
     /**
      * @param RequestHelper $requestHelper
      * @param string|null $pageClass
@@ -106,6 +108,8 @@
         if (!$navigation || !$cacheEnabled) {
             $navigation = new \Pimcore\Navigation\Container();
 
+            $this->navCacheTags = ['output', 'navigation'];
+
             if ($navigationRootDocument->hasChildren()) {
                 $this->currentLevel = 0;
                 $rootPage = $this->buildNextLevel($navigationRootDocument, true, $pageCallback, [], $maxDepth);
@@ -115,7 +119,7 @@
             // we need to force caching here, otherwise the active classes and other settings will be set and later
             // also written into cache (pass-by-reference) ... when serializing the data directly here, we don't have this problem
             if ($cacheEnabled) {
-                CacheManager::save($navigation, $cacheKey, ['output', 'navigation'], $cacheLifetime, 999, true);
+                CacheManager::save($navigation, $cacheKey, $this->navCacheTags, $cacheLifetime, 999, true);
             }
         }
 
@@ -338,6 +342,8 @@
                     $pageCallback($page, $child);
                 }
 
+                $this->navCacheTags[] = $page->getDocument()->getCacheTag();
+
                 $pages[] = $page;
             }
         }

```

brusch:
@NiklasBr LGTM 👍 Can you put that into a PR? Release would be planned for today 😊

PR #8866, opened 2021-04-22
CountryOptionsProvider for Country/Countrymultiselect

blankse:
Country and Countrymultiselect must at the moment always save all options, because they only generate the options on creation.

We should use here the OptionsProvider. Than we don't need to save in the definition file.


weisswurstkanone:
Hi @blankse , I still still all countries (for a single select) in the definition file. Could you please double-check? Thx!

blankse:
@weisswurstkanone I mean the JSON export file. With jsonSerialize function it removes the options. But you are right. The definition php file don't need it either. I will fix this also for alle selections

weisswurstkanone:
Nice! Thanks a lot!

PR #8887, opened 2021-04-26
Fix Text Layout without renderer class

blankse:
Try to open a Event object in the Pimcore Demo:
image

Regression from https://github.com/pimcore/pimcore/pull/8859


blankse:
Here it was decided that the `renderLayoutText` should handle nullable object params: https://github.com/pimcore/pimcore/pull/7156

brusch:
Thanks a lot @blankse !

PR #8892, opened 2021-04-26
[Data objects] Prevent fatal error when an unpublished version object gets opened after new meta field got added

BlackbitNeueMedien:
Steps to reproduce:
1. Create class `Product`, add advanced many-to-many relation field `test`, allow assets
2. Create object of class `Product`, assign an asset, click `Only save new version`
3. Add meta column `abc` to field `test` in class `Product`
4. Reload object -> Uncaught exception

The reason is that the serialized `ElementMetadata` object does not have the newly added column but in https://github.com/pimcore/pimcore/blob/881f8a4448f23bddcafb8c01ccc5954cc00dc40f/models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php#L342-L345
the current columns of the relational field get used. Thus there is an exception thrown in https://github.com/pimcore/pimcore/blob/881f8a4448f23bddcafb8c01ccc5954cc00dc40f/models/DataObject/Data/ElementMetadata.php#L101

### DIFF
--- a/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
+++ b/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
@@ -17,6 +17,7 @@
 namespace Pimcore\Model\DataObject\ClassDefinition\Data;
 
 use Pimcore\Db;
+use Pimcore\Logger;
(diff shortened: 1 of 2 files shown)


weisswurstkanone:
thx!

PR #8906, opened 2021-04-27
Fix PHPStan level 4 errors

blankse:
Follow up #8684 

Before:
```
 [ERROR] Found 130 errors 
```

After:
```
[ERROR] Found 89 errors
```


weisswurstkanone:
Merci!

PR #8916, opened 2021-04-27
2FA login form should use relative url

picman:
Here should be a relative not absolute URL. It makes problems if Pimcore is behind a proxy.

Original output of the view:

`<form method="post" action="http://xyx.cz/admin/login/2fa-verify">`

After the change

`<form method="post" action="/admin/login/2fa-verify">`


brusch:
Thanks a lot!

brusch:
Should be fine for X: 
https://github.com/pimcore/pimcore/blob/9af9958fbcfb9e86406f8076212196a52b831b81/bundles/AdminBundle/Resources/views/Admin/Login/twoFactorAuthentication.html.twig#L14

PR #8918, opened 2021-04-27
[Areablock editable] Support twig condition for "areablock is (not) empty"

BlackbitNeueMedien:
When you have a document template with an area block like this:
```twig
{% set areablock = pimcore_areablock('container') %}

{% if areablock is not empty %}
    {{ areablock | raw }}
 {% endif %}
```
then you will see the area block's contents duplicated. This is because of how Twig checks for `empty`, see https://twig.symfony.com/doc/2.x/tests/empty.html - the problematic part is `For objects that implement the __toString() magic method (and not Countable), it will check if an empty string is returned.` Without this PR this means that https://github.com/pimcore/pimcore/blob/7e0b0a68cc32fa09b16e75d8de3ee8e49a4b8dc7/models/Document/Editable.php#L463-L469 gets called and this renders the area block items.
With this PR this does not happen because Twig then uses the `count()` method to check if the area block is empty.

### DIFF
--- a/models/Document/Editable/Areablock.php
+++ b/models/Document/Editable/Areablock.php
@@ -31,7 +31,7 @@ use Pimcore\Tool\HtmlUtils;
 /**
  * @method \Pimcore\Model\Document\Editable\Dao getDao()
  */
-class Areablock extends Model\Document\Editable implements BlockInterface
+class Areablock extends Model\Document\Editable implements BlockInterface, \Countable
 {
     /**
      * Contains an array of indices, which represent the order of the elements in the block
@@ -617,6 +617,11 @@ class Areablock extends Model\Document\Editable implements BlockInterface
     public function getCount()
     {
         return count($this->indices);
+    }
+
+    public function count()
+    {
+        return $this->getCount();
     }
 
     /**


brusch:
Why not just use: 
```twig
{% set areablock = pimcore_areablock('container') %}

{% if not areablock.empty %}
    {{ areablock | raw }}
 {% endif %}
```

If we'd like to support Twig's `empty` test, we'd have to do so for all editables, but I guess just using `.empty` should be sufficient 🤔

BlackbitNeueMedien:
Of course this is always possible. You could also use `{% if areablock.count() > 0 %}` but when we say we support Twig, such simple language features as the `empty` check have to be supported. For example we have developers who have worked with Twig on non-Pimcore projects and we assume that they can use their knowledge also to create Pimcore templates. But they get confused when such simple things do not work - especially the frontend guys who won't be able to understand why `{% if not areablock.empty %}` works and `{% if areablock. is not empty %} does not.

brusch:
I agree that it would be nice that Twig's `empty` tests should work for editables, but we need to ensure that this is the case for all editables and it would be a topic for `X`.

brusch:
I've converted this PR to a draft for now. 
Criteria for being merged is as mentioned above: 
- Base: `master` (as it isn't actually a bug) 
- All editables need to be supported

BlackbitNeueMedien:
Ok, perhaps also we did something wrong but to me it is a bug when an if condition outputs something. But anyway, this was only supposed to be a quick fix as I did not know that other editables to not support `empty` either. I am normally not doing much Twig, thus I currently cannot invest time into this.

PR #8922, opened 2021-04-27
Marking internal classes & methods as @internal (lib)

brusch:


NiklasBr:
@brusch Why was [Tool/Transliteration](/pimcore/pimcore/blob/master/lib/Tool/Transliteration.php) marked as `@internal` and reduced in functionality? It was really useful, for example in Pretty URL and URL Slug generation, as well as creating usernames from regular names.

brusch:
@NiklasBr because we're actually not using it anymore in the core and it was quite a huge library, which was copy & pasted from Drupal so also very hard to maintain. 
I'd recommend to use https://github.com/cocur/slugify instead.

PR #8924, opened 2021-04-27
Update Nginx Setup Guide

putzflorian:
replace app.php with index.php and fix path to thumbnails


brusch:
Nice, thanks a lot! 👍

PR #8931, opened 2021-04-28
Revert "[DataObjects] UrlSlug value does not update when we save it blank"

dvesh3:
Reverts pimcore/pimcore#8914. This should go on 6.9 branch


brusch:
🍒 - picked #8914 to `6.9` branch (see https://github.com/pimcore/pimcore/commit/884282e006c698a302c98e10da437546c752a8ae), so can be closed.

PR #8932, opened 2021-04-28
Add timeout option to elastic search client

### LINKED ISSUE #7803: Ecommerce | Index Service | Elastic Search Timeout Settings
# Improvement
Support timeout in Elastic Search requests.

### Expected behavior
It should be possible to set the timeout of an Elastic Search (ES) request, ideally in milliseconds.
The parameter can help to avoid performance issues on ES server bottlenecks.

### Current behavior
There is already a timeout parameter in the product list service:
https://github.com/pimcore/pimcore/blob/8e02b8deca7fb75ad25dea9e6750228ba8b20d91/bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php#L40
However, the implementation is missing.

### Solution Suggestion
In https://github.com/pimcore/pimcore/blob/84313b448af88570e7e8deeb721e5ea3be64949f/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php#L189

add the following lines: 

```php
$builder = \Elasticsearch\ClientBuilder::create();
$builder->setConnectionParams([
    'client' => [
        'curl' => [
            CURLOPT_TIMEOUT_MS => $timeout
        ]
    ]
]);

```

We could change the signature of
 ``public function getElasticSearchClient()`` to ``public function getElasticSearchClient(array $customConnectionParams = [])``

and then pass the ``timeout`` variable of https://github.com/pimcore/pimcore/blob/8e02b8deca7fb75ad25dea9e6750228ba8b20d91/bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php#L40

as a parameter.  It is very handy to use ms insead of seconds for testing purposes. With this approach, ms (CURLOPT_TIMEOUT_MS) can still be passed by overriding ``AbstractElasticSearch`` and providing the parameter in the array directly.

----

I can create a PR as soon as we agree on the solution.

andreas-gruenwald:
Resolves #7803 by adding default timeout parameters and the possibility to overwrite them via ecommerce index settings.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
+++ b/bundles/EcommerceFrameworkBundle/IndexService/Worker/ElasticSearch/AbstractElasticSearch.php
@@ -40,6 +40,9 @@ abstract class AbstractElasticSearch extends Worker\ProductCentricBatchProcessin
 
     const REINDEXING_LOCK_KEY = 'elasticsearch_reindexing_lock';
 
+    const DEFAULT_TIMEOUT_MS_FRONTEND = 20000; // 20 seconds
+    const DEFAULT_TIMEOUT_MS_BACKEND =  120000; // 2 minutes
+
     /**
      * Default value for the mapping of custom attributes
      *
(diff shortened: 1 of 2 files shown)


fashxp:
thx very much!

PR #8935, opened 2021-04-28
ControllerReferenceMigrationCommand: Don't load all documents at once

blankse:
I have problems to execute this command. It progresses but is still being terminated...

### DIFF
--- a/bundles/CoreBundle/Command/Migration/ControllerReferenceMigrationCommand.php
+++ b/bundles/CoreBundle/Command/Migration/ControllerReferenceMigrationCommand.php
@@ -85,8 +85,8 @@ class ControllerReferenceMigrationCommand extends AbstractCommand
         }
 
         // documents
-        $documents = new Document\Listing();
-        $documents->setUnpublished(1);
+        $documentListing = new Document\Listing();
+        $documentListing->setUnpublished(1);
         $condition = 'type NOT IN (:types)';
 
         if ($input->getOption('documentId')) {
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Hello @blankse , what are the steps to reproduce (except executing the command) ?

blankse:
@weisswurstkanone I think many documents with many area bricks and little memory. I still haven't managed to migrate. i think i need more optimization. I must look deeper. But the command always will killed without message...

Maybe we can only load documents that still have the wrong controller schema?

weisswurstkanone:
@blankse, did you try to execute Pimcore::collectGarbage() inside the loop

blankse:
@weisswurstkanone No, thanks for the hint. I will try :)

dpfaffenbauer:
does it actually has to load the documents? Wouldn't it be efficient enough to just check the documents_snippet and document_page tables directly? That would massively increase the performance. Especially when you deal with lots and big documents

weisswurstkanone:
@dpfaffenbauer , yes maybe there is room for improvement. Perfomance always matters but you execute a migration script only once. Top priority is to get the migration done in a **correct way**. So let us focus on this point first ;-)

dpfaffenbauer:
> @dpfaffenbauer , yes maybe there is room for improvement. Perfomance always matters but you execute a migration script only once. Top priority is to get the migration done in a **correct way**. So let us focus on this point first ;-)

Performance for a migration is also super important. you cannot let the migration run for an hour during a deployment process... that makes the website unusable during that time ;)

dpfaffenbauer:
there you go https://github.com/pimcore/pimcore/pull/8949

blankse:
Close in favor of #8949

PR #8938, opened 2021-04-28
correct path for User images is /user-image

putzflorian:


brusch:
Thanks a lot for the fix!

PR #8939, opened 2021-04-28
[QuickSearch] fix mb-issue in quick-search

fashxp:
when path contains utf8 characters like umlauts quicksearch sometimes looked like 

image

### DIFF
--- a/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
+++ b/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
@@ -523,7 +523,7 @@ class SearchController extends AdminController
             if (strlen($shortPath) <= 50 || $i === 0) {
                 break;
             }
-            array_splice($parts, $i - 1, 1, '...');
+            array_splice($parts, $i - 1, 1, '…');
         }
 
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Searchadmin/SearchController.php
```diff
@@ -526,8 +526,8 @@ protected function shortenPath($path)
             array_splice($parts, $i - 1, 1, '...');
         }
 
-        if (strlen($shortPath) > 50) {
-            $shortPath = substr($shortPath, 0, 47) . '...';
+        if (mb_strlen($shortPath) > 50) {
+            $shortPath = mb_strstr($shortPath, 0, 47) . '...';
```
NiklasBr, line 530:
```suggestion
            $shortPath = mb_strstr($shortPath, 0, 49) . '…';
```

NiklasBr: An actual ellipsis 😄

brusch: Changed it also for the path a few lines above 😊

PR #8941, opened 2021-04-28
Migration to increase prettyUrl field size to 255

NiklasBr:

If the UI field already allows 255 characters, so should the database. Keeping it short is an editorial issue.

https://github.com/pimcore/pimcore/blob/c4f055340935b9028b358f459267fe9ae42137f3/bundles/AdminBundle/Resources/public/js/pimcore/document/pages/settings.js#L213-L217

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20210428145320.php
+++ b/bundles/CoreBundle/Migrations/Version20210428145320.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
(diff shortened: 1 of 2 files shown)


brusch:
@NiklasBr thanks a lot for your PR! Can you please change the length here as well (for new installations): 
https://github.com/pimcore/pimcore/blob/e61c6a15e900096d7bf529aeab32f914adc143a8/bundles/InstallBundle/Resources/install.sql#L186 

Thanks again!

brusch:
Thanks a lot! 🙏

Issue #7488, opened 2020-11-06
Migration causes Error


ckogler-elements:
Migration Causes an error beause a primary key allready exists on the tables...

---------------------
Migrating up to 20201001133558 from 20200812160035

  ++ migrating 20200817133132

     -> ALTER TABLE `object_relations_1` ADD COLUMN `id` BIGINT(20) NOT NULL  AUTO_INCREMENT FIRST
Migration 20200817133132 failed during Execution. Error An exception occurred while executing 'ALTER TABLE `object_relations_1` ADD COLUMN `id` BIGINT(20) NOT NULL  AUTO_INCREMENT FIRST':

SQLSTATE[42000]: Syntax error or access violation: 1075 Incorrect table definition; there can be only one auto column and it must be defined as a key

In AbstractMySQLDriver.php line 128
------------
Solution:
Drop the primary key first 

`                if (!$schema->getTable($relationTable)->hasColumn('id')) {
                    $this->addSql('ALTER TABLE `'.$relationTable.'` DROP PRIMARY KEY');
                    $this->addSql('ALTER TABLE `'.$relationTable.'` ADD COLUMN `id` BIGINT(20) NOT NULL PRIMARY KEY AUTO_INCREMENT FIRST');
                }`

brusch:
Was fixed already by #7258

blankse:
@brusch Doesn't work for me:
```
ALTER TABLE `object_relations_test` DROP PRIMARY KEY;                                                                                                                                                               
  Migration 20200817133132 failed during Execution. Error An exception occurred while executing 'ALTER TABLE `object_relations_dfimportdomain` DROP PRIMARY KEY;':                                                                      
                                                                                                                                                                                                                                        
  SQLSTATE[42000]: Syntax error or access violation: 1075 Incorrect table definition; there can be only one auto column and it must be defined as a key             
```

PR #8944, opened 2021-04-29
Don't remove primary key from new relation id columns

blankse:
See #7258 / #7488


brusch:
@blankse Nice, thanks a lot! (I was just about to look into it because of your comment https://github.com/pimcore/pimcore/issues/7488#issuecomment-828987348 😉) Luckily I've seen your PR first 😊

PR #8947, opened 2021-04-29
[PimcoreX] Translations Table migration is allowed to fail

dpfaffenbauer:
Migration failed for a different reason, but didn't told me. so it should fail and not further proceed. Otherwise you'll end up with following error when trying to access pimcore:

```
InvalidFieldNameException
An exception occurred while executing 'INSERT INTO `translations_admin` (`key`, `type`, `language`, `text`, `modificationDate`, `creationDate`) VALUES (?, ?, ?, ?, ?, ?) ON DUPLICATE KEY UPDATE `key` = ?, `type` = ?, `language` = ?, `text` = ?, `modificationDate` = ?, `creationDate` = ?;' with params ["login", null, "en", "", 1619684040, 1619684040, "login", null, "en", "", 1619684040, 1619684040]: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'type' in 'field list'
```

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20210324152822.php
+++ b/bundles/CoreBundle/Migrations/Version20210324152822.php
@@ -31,19 +31,15 @@ final class Version20210324152822 extends AbstractMigration
      */
     public function up(Schema $schema): void
     {
-        try {
-            $db = Db::get();
+        $db = Db::get();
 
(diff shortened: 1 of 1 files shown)


dvesh3:
makes sense....thanks!

PR #8955, opened 2021-04-29
random_bytes for default passwords

NiklasBr:
By scanning the net for fresh installations or enumerating new accounts on existing sites an attacker could potentially get an idea of the default password. E.g. catching the when the new account appeared, or the creationDate of the root document or the basic Pimcore assets given by visiting the site could clue the attacker in. To prevent this (small) security hole we remove the dependency on `microtime()`.

For the controller, `uniqid()` is also based on microtime and `md5()`:ing it is actually ever so slightly decreasing the strength of the string given by `uniqid()`.

If nothing else this will prevent naive code analysis tools from warning us that *OMG the words "md5" and "password" appears on the same line in the code: must be insecure* ;)

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/UserController.php
+++ b/bundles/AdminBundle/Controller/Admin/UserController.php
@@ -1161,7 +1161,7 @@ final class UserController extends AdminController implements KernelControllerEv
             if (empty($message)) {
                 //generate random password if user has no password
                 if (!$user->getPassword()) {
-                    $user->setPassword(random_bytes(16));
+                    $user->setPassword(bin2hex(random_bytes(16)));
                     $user->save();
                 }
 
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Controller/Admin/UserController.php
```diff
@@ -1161,7 +1161,7 @@ public function invitationLinkAction(Request $request)
             if (empty($message)) {
                 //generate random password if user has no password
                 if (!$user->getPassword()) {
-                    $user->setPassword(md5(uniqid()));
+                    $user->setPassword(random_bytes(16));
```
brusch, line 1164:
```suggestion
                    $user->setPassword(bin2hex(random_bytes(16)));
```

bundles/InstallBundle/Installer.php
```diff
@@ -633,7 +633,7 @@ protected function createOrUpdateUser($config = [])
     {
         $defaultConfig = [
             'username' => 'admin',
-            'password' => md5(microtime()),
+            'password' => random_bytes(16),
```
brusch, line 636:
```suggestion
            'password' => bin2hex(random_bytes(16)),
```

brusch (changes requested):

NiklasBr:
random_bytes() should already return a hash:able string, but your suggestions doesn't hurt 👍

brusch:
@NiklasBr Thanks for your feedback - that's what I thought as well - doesn't hurt 😉

PR #8956, opened 2021-04-29
Add `@method` annotation for `getDao()` to `AbstractListing`

jdreesen:
This way it must not be annotated via `@var` each time (also for external users).


brusch:
Nice, thanks a lot 👍 Have a nice weekend @jdreesen !

PR #8968, opened 2021-04-30
Ecommerce | Cart | order is lost when more than one items are added within one second

### LINKED ISSUE #6464: Ecommerce | Cart | order is lost when more than one items are added within one second
# Bug Report
A common use case is to fill the cart with items based on a fast entry mask or a file-based import.

### Actual behavior
When multiple items are added to the cart within one second, then currently the order is not maintained.

The reason is that the default card item order is based on the ``creationDateTimestamp`` attribute,
which currently only saves the time in seconds. 

### Expected behavior
The items have to be listed in exactly the same order as they have been added to the cart.

### Suggested Solution
Our project-specific solution was to add the following line to the 
``CartManager/Cart::updateItem()`` method:

```php
//set sortindex to nanotime to improve order on mass-insertion (fast-entry, imports)
$item->setSortIndex(hrtime(true)); 
``` 
because then the cart items will be sorted by the sort index.

(a) A better solution would probably be to either add an additional field to the database where the ``hrtime()`` is stored and use the field for the default sorting. 

(b) Another option would be to replace the ``creationDateTimestamp`` with the nanotime value. Option (b) is more likely to cause errors in existing projects.

I can provide a PR after discussion.

fashxp:
fixed #6464

also merge https://github.com/pimcore/demo/pull/226  when this is merged

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20210430124911.php
+++ b/bundles/CoreBundle/Migrations/Version20210430124911.php
@@ -30,7 +30,8 @@ final class Version20210430124911 extends AbstractMigration
     public function down(Schema $schema): void
     {
         if ($schema->hasTable(CartItem\Dao::TABLE_NAME)) {
-            $this->addSql('UPDATE ecommerceframework_cartitem SET addedDateTimestamp = addedDateTimestamp / 1000000;');
+            $this->addSql('UPDATE ecommerceframework_cartitem SET addedDateTimestamp = FLOOR(addedDateTimestamp / 1000000);');
+            $this->addSql('ALTER TABLE ecommerceframework_cartitem modify addedDateTimestamp int(10) not null;');
         }
     }
 }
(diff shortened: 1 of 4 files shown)


bundles/CoreBundle/Migrations/Version20210430124911.php
```diff
@@ -0,0 +1,36 @@
...
+            $this->addSql('ALTER TABLE ecommerceframework_cartitem modify addedDateTimestamp bigint not null;');
+            $this->addSql('UPDATE ecommerceframework_cartitem SET addedDateTimestamp = addedDateTimestamp * 1000000;');
+        }
+
+    }
+
+    public function down(Schema $schema): void
+    {
+        if ($schema->hasTable(CartItem\Dao::TABLE_NAME)) {
+            $this->addSql('UPDATE ecommerceframework_cartitem SET addedDateTimestamp = addedDateTimestamp / 1000000;');
```
brusch, line 33:
```suggestion
            $this->addSql('UPDATE ecommerceframework_cartitem SET addedDateTimestamp = FLOOR(addedDateTimestamp / 1000000);');
            $this->addSql('ALTER TABLE ecommerceframework_cartitem modify addedDateTimestamp int(10) not null;');
```

brusch (changes requested):

PR #8979, opened 2021-05-03
[Pimcore X][ClassDefinition] quantityValue unit selection broken

### LINKED ISSUE #8974: [Pimcore X][ClassDefinition] quantityValue unit selection broken
### Expected behavior
'Default unit' dropdown field and 'Valid units'-Field are populated with existing units.
### Actual behavior
Both fields are empty.

### Steps to reproduce
Create Class with quantityValue and try to select default unit and valid units.

weisswurstkanone:
resolves #8974

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/QuantityValueController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/QuantityValueController.php
@@ -184,8 +184,9 @@ final class QuantityValueController extends AdminController
             $list->setCondition('id IN (' . $string . ')');
         }
 
+        $result = [];
         $units = $list->getUnits();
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/QuantityValueController.php
```diff
@@ -184,8 +184,9 @@ public function unitListAction(Request $request)
             $list->setCondition('id IN (' . $string . ')');
         }
 
+        $result = [];
         $units = $list->getUnits();
-        foreach ($units as $unit) {
+        foreach ($units as &$unit) {
```
jdreesen, line 189: Why the reference?

PR #8981, opened 2021-05-03
[Maintenance] Remove duplicate maintenance job CheckErrorLogsDbTask

### LINKED ISSUE #8969: Duplicated maintenance job?
# Bug Report
There are two very similar (same) maintenance jobs running.

These two jobs seem to be doing the exact same thing:
https://github.com/pimcore/pimcore/blob/master/lib/Maintenance/Tasks/CheckErrorLogsDbTask.php
https://github.com/pimcore/pimcore/blob/master/lib/Maintenance/Tasks/LogMailMaintenanceTask.php

They are both registered in:
https://github.com/pimcore/pimcore/blame/master/bundles/CoreBundle/Resources/config/maintenance.yml

Am I missing something here or is this just a refactoring mishap? 🙂

dvesh3:
Resolves #8969


IgorBenko:
Thanks a lot Divesh!

PR #8983, opened 2021-05-03
[Data object grid] Validate grid export form on submit

BlackbitNeueMedien:
There is no validation currently for grid exports. For example when you request a CSV export without a column separator you will get something like 
`"id""fullpath""published""creationDate""modificationDate""name""shortDescription"`

This PR adds validation and also makes the column separator a required field.
The validation also makes sense for custom exports which perhaps also might need some validation.


weisswurstkanone:
thx!

PR #8985, opened 2021-05-04
Fix PHPStan level 4 errors

blankse:
Follow-up #8906 

phpstan-symfony understands container parameter types now. So phpstan finds new errors.
See https://github.com/phpstan/phpstan-symfony/releases


weisswurstkanone:
@blankse , just one thing https://github.com/pimcore/pimcore/runs/2499224606 fails
Could you have a look please ?

blankse:
@weisswurstkanone  See #8987

PR #8990, opened 2021-05-04
Fix: NotificationService::sendToGroup() has unsafe role condition

blankse:
If I send a notification to a role with ID 2, I will also send notifications to role 20 and others.

Maybe it would be better if there was a leading and a trailing comma in this type of column.


dvesh3:
thanks!

PR #9000, opened 2021-05-04
HeadlessChrome Processor: Fix EventDispatcher call

blankse:
The $event should be the first argument in Symfony 5


dvesh3:
thanks!

PR #9005, opened 2021-05-04
[WebDAV] Performance improvement

BlackbitNeueMedien:
This PR uses SQLite for the lock file if available and falls back to file storage if not. This is recommended on https://sabre.io/dav/locks/

Additionally this PR removes the Sync plugin - I actually did not find a reason what this is good for in the Pimcore context, see https://sabre.io/dav/sync/


bundles/CoreBundle/Migrations/Version20210505093841.php
```diff
@@ -0,0 +1,47 @@
...
+final class Version20210505093841 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'WebDAV locks with SQLite instead of file-based';
+    }
+
+    public function up(Schema $schema): void
+    {
+        if(!file_exists(PIMCORE_SYSTEM_TEMP_DIRECTORY.'/webdav-locks.sqlite')) {
```
jdreesen, line 24: Is it right to create this once via migration inside a temp directory? What if the temp directory is somehow cleared and the migration already ran? Does that mean that from there on the lock-file (`webdav-locks.dat`) will be used?

BlackbitNeueMedien: Ah you are right. But I actually do not want to check on every request if the database tables have been set up... I will have look how to solve this differently.

brusch:
@BlackbitNeueMedien I think we need to go for MySQL/MariaDB anyway, because it shouldn't be necessary anymore to share `PIMCORE_SYSTEM_TEMP_DIRECTORY` across multiple app servers/containers.

BlackbitNeueMedien:
@brusch Yes, I also thought about that. I will implement a MySQL Sabre-DAV-Backend and will create a PR for sabre/dav to allow to set the table names for `\Sabre\DAV\Locks\Backend\PDO`

brusch:
@BlackbitNeueMedien Nice, thanks a lot! Maybe you can even contribute the Doctrine lock backend to `sabre/dav`, I can imaging that this could be from general interest.

brusch:
@BlackbitNeueMedien just found https://github.com/sabre-io/dav/issues/778#issuecomment-186645884
So I guess we're also fine in just using the PDO lock backend and get the PDO connection from Doctrine: 
```php
        $lockBackend = new \Sabre\DAV\Locks\Backend\PDO(\Pimcore\Db::get()->getWrappedConnection());
```

brusch:
@BlackbitNeueMedien sorry for spamming you 😉 `\Sabre\DAV\Locks\Backend\PDO::$tableName` is public, so that should be fine as well (at least for now): 
```php
$lockBackend = new \Sabre\DAV\Locks\Backend\PDO(\Pimcore\Db::get()->getWrappedConnection());
$lockBackend->tableName = 'webdav_locks'; 
```

BlackbitNeueMedien:
Oh great. I give that a try. Give me a few seconds

brusch (changes requested):
@BlackbitNeueMedien 2 more things 😊: 
1. Idempotent migrations
2. `CREATE TABLE` should be in `install.sql` as well: https://github.com/pimcore/pimcore/blob/7fbbbf9413885e0b97fe64f2d52871021a7da0da/bundles/InstallBundle/Resources/install.sql#L944

BlackbitNeueMedien:
Sorry, I should not commit 1 minute before a customer call... should be ok now.

brusch:
Very nice, thanks a lot! 👍

PR #9010, opened 2021-05-05
Update phpstan-symfony (Non-empty parameter array might be empty in different environment)

blankse:
The new phpstan-symfony version ignores empty array parameters now.
See https://github.com/phpstan/phpstan-symfony/releases/tag/0.12.32


brusch:
Nice, thanks a lot! 👍

PR #9014, opened 2021-05-05
Element\Service::getSafeCopyName() created deprecated alias and removed unnecessary argument

### LINKED ISSUE #9004: Element\Service::getSaveCopyName renamed to Element\Service::getSafeCopyName without notice
- add alias-method 
- rediscuss if we really should make it internal ... is also quite pratical for importers and other use cases.

brusch:
Resolves #9004


blankse:
@brusch See https://github.com/pimcore/pimcore/pull/9010 for the Stan Error

PR #9019, opened 2021-05-05
[Predefined Asset metadata] Groups

BlackbitDevs:
This PR adds the option to define groups for predefined asset metadata. This way you can manage metadata which belongs to the same asset type (e.g. document) but is for different document types (e.g. PDF manuals, PDF certificates etc.).

On predefined metadata panel there is a new column `group` (in this screenshot it is named "Gruppe" (german)):
Bildschirmfoto 2021-05-05 um 14 17 59

Then on asset's `Custom Metadata` panel the button `Add predefined definitions` is now a split button. In the dropdown are the available groups for this asset type:
Bildschirmfoto 2021-05-05 um 14 17 35

When you then select one of the groups, only the metadata from this group gets added to the asset (in this case the group `Manual` was clicked):
Bildschirmfoto 2021-05-05 um 14 18 24

When you click the main button `Add predefined definitions` only metadata without group gets added to the asset. This way it is possible to have a basic set of metadata fields which can be applied to all assets of a certain type. When we would add all metadata such non-specific assets would get cluttered with all the metadata from the different groups.


bundles/CoreBundle/Resources/translations/de.json
```diff
@@ -435,6 +435,7 @@
     "columns": "Spalten",
     "children_grid": "Kinder Tabelle",
     "only_children": "Nur direkte Kindelemente",
+    "only_unreferenced": "Nur nicht referenzierte",
```
weisswurstkanone, line 438: No need for de.json. en is the master file

BlackbitNeueMedien: Is only a draft, please do not review yet. We try to teach also other colleagues to create PRs. We will fix it and convert it to a real PR soon.

weisswurstkanone:
@BlackbitNeueMedien Just one thought ... would it make sense to allow multiple groups (comma separated maybe) in a way to enable overlapping stuff?

BlackbitDevs:
@weisswurstkanone I tried to implement it similar to the target subtype (document, image etc). As there is currently only 1 subtype selectable, I did the same for groups.

weisswurstkanone:
Perfectly fine ;-)

bundles/AdminBundle/Controller/Admin/SettingsController.php
```diff
@@ -236,11 +236,14 @@ public function getPredefinedMetadataAction(Request $request)
...
         $type = $request->get('type');
         $subType = $request->get('subType');
+        $group = $request->get('group');
         $list = Metadata\Predefined\Listing::getByTargetType($type, [$subType]);
         $result = [];
         foreach ($list as $item) {
-            $item->expand();
-            $result[] = $item->getObjectVars();
+            if ((string)$group === $item->getGroup()) {
+                $item->expand();
```
weisswurstkanone, line 243: this doesn't work anymore since the properties are protected now (since 10).
please add $item = $item->getObjectVars() after $item->expand()

models/Metadata/Predefined.php
```diff
@@ -78,6 +78,11 @@ final class Predefined extends Model\AbstractModel
      */
     protected $language;
 
+    /**
+     * @var string
+     */
+    public $group;
```
weisswurstkanone: please change this to protected

weisswurstkanone:
@BlackbitNeueMedien 🏓

bundles/AdminBundle/Controller/Admin/SettingsController.php
BlackbitDevs: Fixed. (I actually developed this for Pimcore 6 and just copied the changes)

models/Metadata/Predefined.php
BlackbitDevs: Fixed. (I actually developed this for Pimcore 6 and just copied the changes)

weisswurstkanone:
Hello @BlackbitNeueMedien , thanks a lot! Would it be possible to stick to the old behavior regarding main button. I.e. add all definitions (even with a group assignment), but add an entry to the split menu in a way like this:

Main (adds all)
    - add undefined
    - seprator line
    - add group 1
    - add group 2
    - ...

wdyt? 🤔

BlackbitDevs:
Of course this is possible but in which cases would it make sense to apply all metadata? As mentioned above the groups are for me an extension of the asset subtype. As it would be nonsense to have a possibility to add all metadata fields for all asset subtypes, it is the same for adding metadata from all groups.
But it would not be a problem for me. If you think adding all metadata with the main button is a good idea, just confirm again and I will change it.

weisswurstkanone:
If it is not too much effort, it would be nice! 😄

BlackbitDevs:
@weisswurstkanone Ok, I fixed it. Main button now adds all predefined metadata and in dropdown menu there is an item to only add predefined metadata without group.

weisswurstkanone:
@BlackbitNeueMedien nice addition! thx a lot!

PR #9023, opened 2021-05-05
[Docs] Fix upgrade instructions

BlackbitNeueMedien:
`web/var/tmp` does not exist anymore after the step `Rename web/ to public/` is done.

After moving `app/AppKernel.php` you have to put it to namespace `App` and rename the class to `Kernel`, otherwise you get `Defined Kernel Class \App\Kernel not found`


dvesh3:

PR #9024, opened 2021-05-05
[X installation] Make migrations idempotent

BlackbitNeueMedien:
Steps to reproduce problem:
1. Install Pimcore 10 with `composer create-project --prefer-dist pimcore/skeleton:dev-master` and `./vendor/bin/pimcore-install`
2. Directly after run `composer update`

Without this PR a migration will fail with 
```
An exception occurred while executing 'ALTER TABLE `versions` ADD `autoSave` TINYINT(4) NOT NULL DEFAULT 0':                                                                                                                                                                                                       
SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name 'autoSave' 
```

The actual reason is that this operation is already done in https://github.com/pimcore/pimcore/blob/7fbbbf9413885e0b97fe64f2d52871021a7da0da/bundles/InstallBundle/Resources/install.sql#L723
This PR makes the migration idempotent so that it does not matter if the column exists or not, result will always be the same.

When you think one step further the question is what we should put in `/bundles/InstallBundle/Resources/install.sql` - I usually put the initial setup there and freeze this forever. All the changes are only done with migrations. This way you never endanger to execute a change in both: install SQL and migration. WDYT?

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20210408153226.php
+++ b/bundles/CoreBundle/Migrations/Version20210408153226.php
@@ -30,8 +30,15 @@ class Version20210408153226 extends AbstractMigration
      */
     public function up(Schema $schema): void
     {
-        $this->addSql('ALTER TABLE `versions` ADD `autoSave` TINYINT(4) NOT NULL DEFAULT 0');
-        $this->addSql('ALTER TABLE `versions` ADD INDEX `autoSave` (`autoSave`)');
+        $versionsTable = $schema->getTable('versions');
+
+        if(!$versionsTable->hasColumn('autoSave')) {
+            $this->addSql('ALTER TABLE `versions` ADD `autoSave` TINYINT(4) NOT NULL DEFAULT 0');
+        }
+
+        if (!$versionsTable->hasIndex('autoSave')) {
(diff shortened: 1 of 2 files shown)


brusch:
@BlackbitNeueMedien I'm wondering why this happend to you, because actually the installer marks all migrations as done here: 
https://github.com/pimcore/pimcore/blob/c5da5cf7bc1ed94e71ddcaf653cff35534d5d940/bundles/InstallBundle/Installer.php#L481 

Just tried that and worked fine for me. 

To be honest I don't like the idea of running all migrations on install, because if you look e.g. at Pimcore `6.9`: 
https://github.com/pimcore/pimcore/tree/6.9/bundles/CoreBundle/Migrations
We have exactly 100 migrations, that would need to run on install ... that would take some time and makes it probably also even more confusing for first time users 😐

BlackbitNeueMedien:
Hmm, I don't know, why this has not been done. But ok, if it works on your side, perhaps I made a mistake - you probably have installed Pimcore 10 more often than me (I did exactly once ;-) )
But nevertheless don't you think idempotent migrations are good? If not, please jsut close this PR.

brusch:
Idempotent migrations can't hurt 😉
@dvesh3 @weisswurstkanone @fashxp please consider in future. Thanks!

PR #9026, opened 2021-05-05
[Docs] Symfony 5 does only support 1 config file with symfony firewalls

BlackbitNeueMedien:


dvesh3:
maybe just change last point here: https://github.com/pimcore/pimcore/blob/8eec5479e48d32901335c8875bb557e24df535cd/doc/Development_Documentation/23_Installation_and_Upgrade/07_Updating_Pimcore/10_V6_to_V10.md#adapt-security-config

BlackbitNeueMedien:
@dvesh3 The problem is that when you check that after the update you won't be able to execute `bin/console debug:config security firewalls` anymore - you are actually not able to execute anything.

BlackbitNeueMedien:
I merged the 2 paragraphs - but into the `before the update` section.

dvesh3:
@BlackbitNeueMedien thanks! 🙏

PR #9027, opened 2021-05-05
8300: Reduce varchar length for quantity value unit ids

### LINKED ISSUE #8300: Units using varchar(50) IDs causes tables to reach maximum row length quickly
Hey all,

this is not really a bug per se, but a bit of an nuisance nevertheless. 
I noticed that the ID field for units uses varchar(50) on the database level, instead of int(11) as e.g. the assets or objects tables do. This causes this field type to be used for the foreign key fields in object tables where units are used as well, which in turn leads to unnecessarily big row lengths. In classes that use lots of quantity values, tables reach the maximum row length very quickly. Can this field be changed to a shorter ID in future releases, so that we do not run into this issue as fast?
Thanks!

johannesdold:
Resolves #8300 

## Additional info
The column definition for quantity value units was recently changed from bigint(20) to varchar(50). If you have many quantity value data fields in one class, the maximum row length of the innodb table is reached very quickly. This pr reduces the varchar length to 15 - which should be enough for defining the unit ids.


johannesdold:
@weisswurstkanone Maybe another solution would be possible. If the dynamic row format is used, than varchar fields above a certain size are stored on overflow pages ([https://mariadb.com/kb/en/troubleshooting-row-size-too-large-errors-with-innodb/](https://mariadb.com/kb/en/troubleshooting-row-size-too-large-errors-with-innodb/)). In our test, if I set the size to 255, I do not get the row size too large error. So maybe we should instead of reducing the size, increase it. What do you think?

weisswurstkanone:
Please give it a try! 👍

weisswurstkanone:
Just one thought. I think there should be a validation on save. So if it doesn't satisfy the limits then throw an exception.

johannesdold:
@weisswurstkanone I am not sure that I understood your last comment correctly. I changed the size now to 64 and added an exception, if the user tries to create a quantity value unit with a unit ID exceeding the 50 characters count imposed by the restriction in the quantityvalue_units table.

weisswurstkanone:
@johannesdold could you please provide steps how to reproduce this on the official demo. Besides that, in my opinion 50 is too long anyway. I guess that the goal is to use the unit abbreviation as unit id (like cm, mm, etc) instead of an autoincrement integer.

johannesdold:
@weisswurstkanone It would be essential for us, that we get this adaption in a 6.9 release. I will try to reproduce the issue on the demo and come back to you.

johannesdold:
@weisswurstkanone To reproduce the issue on the demo, add a class with 40 quantity value fields and try to save:

large-class

The limit is reached somewhere between 30 and 40 fields.

weisswurstkanone:
Could you please check what's configured for your
"innodb_file_format" variable?

johannesdold:
@weisswurstkanone It is Barracuda.

weisswurstkanone:
please check if "dynamic" works for you

johannesdold:
@weisswurstkanone  As I understand it, the innodb_file_format is Barracuda (or Antelope), but the innodb_row_format can take the value dynamic. Which it does in my setup. This leads to the information being stored in overflow pages, if the varchar is equal to or greater than 64 (in a utf8mb4 context). Here is a quite nice description of the problem:

https://mariadb.com/kb/en/troubleshooting-row-size-too-large-errors-with-innodb/

The issue can be reproduced on the demo installation, as shown above. I guess you are also using the file format Barracuda and row format dynamic.

johannesdold:
@weisswurstkanone Could you please give me feedback ob how you will proceed with this topic? Can we get a solution into a 6.9.x release?

brusch:
@johannesdold as this isn't really a bug there wont be a patch for `6.9` anymore. Sorry & thanks for your understanding.

brusch:
@johannesdold can you please rebase this PR onto `10.x`? Thanks a lot!

weisswurstkanone:
@johannesdold 🏓

johannesdold:
@weisswurstkanone  Ups. I guess I just made a mistake when rebasing. Sorry, I am not sure how to clean that up.

weisswurstkanone:
maybe just create a new PR

johannesdold:
Yes, will do that.

PR #9028, opened 2021-05-05
[Workflows] Fix object not unpublished with force_unpublished option

### LINKED ISSUE #7586: Workflow with changePublishedState force_unpublished will create new version instead of unpublishing the object
# Bug Report

### Expected behavior

Workflow with two simple transitions (publish and unpublish) on DataObject:
```yaml
type: 'state_machine'
    marking_store:
        type: 'single_state'
        arguments: [ workflowState ]
```
...

```yaml
into_inactive:
into_unpublished:
    guard: is_fully_authenticated()
    from: [ published ]
    to: unpublished
    options:
        label: 'Unpublish'
        notes:
            commentEnabled: false
            commentRequired: false
        changePublishedState: force_unpublished
into_published:
    guard: is_fully_authenticated()
    from: [ unpublished ]
    to: publish
    options:
        label: 'Publish'
        notes:
            commentEnabled: false
            commentRequired: false
        changePublishedState: force_published
```
These transitions should toggle between publish and unpublish the DataObject.

### Actual behavior

Publishing is working fine. But unpublishing the DataObject will create a new version and leaves the object published. I expect, that the DataObject is unpublished afterwards.

ergosarapu:
Resolves #7586

## Additional info  

[Fix] Call element `saveVersion()` when workflow option `save_version` is set, otherwise call element `save()`. This allows element to unpublish when `force_unpublished` option is set.

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
@@ -1,6 +1,7 @@
 # Upgrade Notes
 
 ## 6.9.0
+- [Workflows] Fixed wrong behavior of force_unpublished option, which created a new version (in case of published) instead of unpublishing the object. If you still want to save version, just use save_version option.
 - [Documents] Announcement: Run the `pimcore:documents:migrate-elements` command before (!) you upgrade to Pimcore 10. Not needed if you don't intend to keep your document versions.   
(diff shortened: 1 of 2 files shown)


dvesh3:
@ergosarapu thanks a lot for the fix. We need to add upgrade notes for the behavior change, even though this is fixes the wrong behavior of `force_unpublished` option.

Could you please add this line to upgrade notes? _"[Workflows] Fixed wrong behavior of `force_unpublished` option, which created new version(in case of published) instead of unpublishing the object. If you still want to save version, just use `save_version` option_."

dvesh3:
@ergosarapu thanks!

ergosarapu:
Thanks for quick replies and merge!

PR #9031, opened 2021-05-06
Get exclusive lock when writing PHP config files

BlackbitNeueMedien:
When you have a (temporarily) slow write I/O on the server you might get the problem that generated PHP config files get messed up. We experienced this in the context of #9019. When you create a predefined metadata field, the ExtJS store automatically saves this. When you then create another field before the first save operation is finished `var/config/predefined-asset-metadata.php` might end up with something like
```php
<?php 

return [
    1 => [
        "id" => 1,
        "name" => "author",
        "description" => "",
        "type" => "input",
        "targetSubtype" => NULL,
        "data" => "",
        "config" => NULL,
        "language" => NULL,
        "creationDate" => 1557394593,
        "modificationDate" => 1566383852
    ]
];
];
```

Of course this gives errors afterwards as this file is no valid PHP. You can only fix this manually.

The try-catch is for filesystems which do not support exclusive locking - for those the behaviour remains the same.

### DIFF
--- a/lib/File.php
+++ b/lib/File.php
@@ -15,6 +15,8 @@
 
 namespace Pimcore;
 
+use Throwable;
+
 class File
 {
     /**
@@ -99,7 +101,12 @@ class File
             self::mkdir(dirname($path));
         }
 
-        $return = file_put_contents($path, $data, null, self::getContext());
+        try {
+            $return = file_put_contents($path, $data, LOCK_EX, self::getContext());
+        } catch(Throwable $e) {
+            $return = file_put_contents($path, $data, null, self::getContext());
+        }
+
         @chmod($path, self::$defaultMode);
 
         return $return;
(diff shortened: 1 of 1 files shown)


brusch:
@BlackbitNeueMedien what do you think about #9032 ? 
It's the same concept Flysystem is using.

BlackbitNeueMedien:
I think the result is exactly the same as this PR, so I am fine with this ;-)
But I would add a fallback for file systems which do not support file locking, just google for "Exclusive locks are not supported for this stream"

brusch:
I'm wondering if we really need a fallback or if just the config should be changed in such cases. Flysystem e.g. doesn't have a fallback, you just to change the default config.

brusch:
I'd go for #9032 for now, if we see the need for a fallback, we can add it without breaking anything in the future. 
Thanks a lot for pointing out this issue 👍

PR #9032, opened 2021-05-06
Get exclusive lock when writing files 

brusch:
alternative approach to #9031

### DIFF
--- a/lib/File.php
+++ b/lib/File.php
@@ -29,6 +29,11 @@ class File
     protected static $context = null;
 
     /**
+     * @var int
+     */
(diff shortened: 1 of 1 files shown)


BlackbitNeueMedien:
@brusch I don't know exactly but can this also happen to generated data object classes or does this PR also apply there? It never happened to me there but I could imagine that it would be very frustrating to clean up a messed definition file which got written from 2 processes.

brusch:
@BlackbitNeueMedien it applies also there. Also for many other places where it can't hurt. (System Settings, ...).

PR #9035, opened 2021-05-06
Update pimcore image gallery hotspots from database type text to longtext

mattamon:
when adding a lot of hotspots with additional data (objects, e.g.) the database datatype text is not sufficient.
the hotspots text is then cut off and the hotspots won't load anymore in the pimcore backend. 
Therefore I propose to update to database type longtext instead to make sure all hotspots are stored correctly


brusch:
Thanks a lot! 👍

PR #8054, opened 2021-02-02
Fix: No thumbnail for PDF documents with ) sign

blankse:
Resolve #7938

PR #9041, opened 2021-05-06
Fix: No thumbnail for PDF documents with ) sign again

blankse:
Was already fixed in https://github.com/pimcore/pimcore/pull/8054 but get lost


brusch:
Thanks a lot! 👍

PR #9043, opened 2021-05-06
Fix: getText not work with ghostscript only

blankse:
When you have only installed the gs command without the pdftotext command, the getText method write an error in the maintenance cronjob and the method return no text.


brusch:
Thanks a lot!

PR #9047, opened 2021-05-06
Fix PhpStan errors

blankse:
This fix all PhpStan errors in the baseline file. False positives are ignored inline.


blankse:
@brusch The PHP-CS-Fixer seems not working with fork branches...

blankse:
@brusch Unfortunately still does not work :(

brusch:
@blankse ok, now it fails at a different step 🤔

blankse:
@brunsch I read something about the pull_request_target workflow event for forks. See https://github.com/stefanzweifel/git-auto-commit-action:
`The following workflow runs php-cs-fixer (a code linter and fixer for PHP) when a pull_request is opened. We've added the pull_request_target-trigger too, to make it work for forks.`

brusch:
@blankse thanks for the hint, I've seen this as well, but at the end they also write: 
> For Workflows which run linters and fixers (like the example above) we recommend running them when a push happens on the main-branch.

So I've excluded PRs for now, which should be also fine.

brusch:
@blankse can you please rebase & push again? Thanks 😊

blankse:
@brusch Okay 👍🏼 Should now work 😄 Have also a nice weekend and I looking forward to the release next week 🎉

blankse:
@weisswurstkanone I see you change the milestone to 10.1

I think this could be too late for some changes because of BC. Should I split this changes in a seperate PR, so you can review this and merge before 10.0 Release?

brusch:
@blankse that would be great, thanks a lot!

brusch:
Closed in favor of #9067

PR #9049, opened 2021-05-06
Pass extra arguments to post-add/update events

danmichaelo:
When saving a Data Object, any additional arguments passed to the `save()` method are currently passed to the `DataObjectEvents::PRE_ADD` and `DataObjectEvents::PRE_UPDATE`, but not to the POST events.

This PR adds any arguments passed to `save()` to the events 
- `DataObjectEvents::POST_ADD`
- `DataObjectEvents::POST_UPDATE`
- `DataObjectEvents::POST_ADD_FAILURE`
- `DataObjectEvents::POST_UPDATE_FAILURE`

## Additional info

I'm a bit puzzled by [this forum post](https://talk.pimcore.org/t/how-can-i-find-from-where-event-has-triggered/3234/4) which seems to claim that additional params are already passd to the post update event, but from what I can see, they are not.

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -583,19 +583,19 @@ class Asset extends Element\AbstractElement
             $this->clearDependentCache($additionalTags);
             $this->setDataChanged(false);
 
+            $postEvent = new AssetEvent($this, $params);
             if ($isUpdate) {
-                $updateEvent = new AssetEvent($this);
                 if ($differentOldPath) {
-                    $updateEvent->setArgument('oldPath', $differentOldPath);
+                    $postEvent->setArgument('oldPath', $differentOldPath);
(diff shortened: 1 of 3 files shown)


brusch:
@danmichaelo Makes absolutely sense, but we should do the same for assets and documents. Can you please update your PR accordingly? Thanks a lot 👍

danmichaelo:
> @danmichaelo Makes absolutely sense, but we should do the same for assets and documents. Can you please update your PR accordingly? Thanks a lot 👍

Good point, updated the PR now!

dvesh3 (approved):

dvesh3:
@danmichaelo thanks a lot!

PR #9053, opened 2021-05-07
[EDITABLES] Fix wrong array key check

roland4432:
Correct way to check for array key existence to prevent errors / warnings

  
  

Correct check for array key existence


dvesh3:
@roland4432 thanks!

PR #9067, opened 2021-05-10
Fix PhpStan errors

blankse:
See https://github.com/pimcore/pimcore/pull/9047

Before:
```
[OK] Baseline generated with 52 errors.
```

After:
```
[OK] Baseline generated with 24 errors.
```

PR #9068, opened 2021-05-10
[Workflows] Deprecated uasort usage

### LINKED ISSUE #9065: [Workflows] Deprecated uasort usage
Deprecated: uasort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero in /var/www/html/vendor/pimcore/pimcore/lib/Workflow/Manager.php on line 167

brusch:
resolves #9065

### DIFF
--- a/lib/Workflow/Manager.php
+++ b/lib/Workflow/Manager.php
@@ -163,7 +163,10 @@ class Manager
         $this->workflows[$workflowName] = new WorkflowConfig($workflowName, $options);
 
         uasort($this->workflows, function (WorkflowConfig $a, WorkflowConfig $b) {
-            return $a->getPriority() < $b->getPriority();
+            if ($a->getPriority() == $b->getPriority()) {
+                return 0;
+            }
+            return ($a->getPriority() > $b->getPriority()) ? -1 : 1;
         });
     }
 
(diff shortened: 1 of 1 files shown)


blankse:
@brusch I fixed the new phpstan error here: https://github.com/pimcore/pimcore/pull/9067
The getById static function already exists in the ElementInterface.

brusch:
@blankse pulled that into #9071 for now, so that we can continue independently from #9067

lib/Workflow/Manager.php
```diff
@@ -163,7 +163,10 @@ public function registerWorkflow(string $workflowName, array $options = [])
         $this->workflows[$workflowName] = new WorkflowConfig($workflowName, $options);
 
         uasort($this->workflows, function (WorkflowConfig $a, WorkflowConfig $b) {
-            return $a->getPriority() < $b->getPriority();
+            if ($a->getPriority() == $b->getPriority()) {
+                return 0;
+            }
+            return ($a->getPriority() > $b->getPriority()) ? -1 : 1;
```
jdreesen: That's a case for the spaceship operator:

```suggestion
            return $b->getPriority() <=> $a->getPriority();
```

PR #9076, opened 2021-05-10
Fix return phpdoc of ManyToOneRelation

blankse:
Example

Before:
```
/**
* Get preSelect - Pre Select
* @return \\Pimcore\Model\DataObject\Category|null
*/
```

After:
```
/**
* Get preSelect - Pre Select
* @return \Pimcore\Model\DataObject\Category|null
*/
```

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Extension/Relation.php
+++ b/models/DataObject/ClassDefinition/Data/Extension/Relation.php
@@ -20,7 +20,7 @@ trait Relation
     /**
      * @internal
      *
-     * @param bool|false $asArray
+     * @param bool $asArray
      *
      * @return string[]
      */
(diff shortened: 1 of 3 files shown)


weisswurstkanone:
merci!

jdreesen:
For the future: you can display diffs as follows

<pre>
```diff
  /**
  * Get preSelect - Pre Select
- * @return \\Pimcore\Model\DataObject\Category|null
+ * @return \Pimcore\Model\DataObject\Category|null
  */
```
</pre>

which makes it easier to see such subtle changes

```diff
  /**
  * Get preSelect - Pre Select
- * @return \\Pimcore\Model\DataObject\Category|null
+ * @return \Pimcore\Model\DataObject\Category|null
  */
```

PR #9083, opened 2021-05-11
[EDITABLE] - Video - various fixes to make the video editable work

roland4432:
* return null for poster image if none is set
* corrected check for array values with a certain key
* replaced upload time format (ISO8601 was removed in php 7.2)
* used https instead of http for schema.org snippet


brusch:
Thanks a lot! 👍

PR #9110, opened 2021-05-13
Fix migration error when upgrading from 6.9.3 to 10.0

ramundomario:
Migration error upgrading from 6.9.3 to 10.0:
Screenshot from 2021-05-13 21-34-13
  

Resolves #

## Additional info  
Migration 20210324152822 will fail when upgrading from 6.9:
https://github.com/pimcore/pimcore/blob/1a33a2b93782be8c43683935623e3e771d5caef9/bundles/CoreBundle/Migrations/Version20210324152822.php#L32-L44
The reason is translation tables are dynamically retrieved with `SHOW TABLES` which returns both tables and views causing error.

Table `translations_website` was renamed here and a BC view was created:
https://github.com/pimcore/pimcore/blob/b36e6568fd3adb39980b7b65c456d049abee97ab/bundles/CoreBundle/Migrations/Version20201222101114.php#L16-L24

This PR add a migration, before of 20210324152822, to remove the BC view which had to be removed in v10 in #7808 as described [here](https://github.com/pimcore/pimcore/pull/7800#discussion_r549352650)

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20210324152821.php
+++ b/bundles/CoreBundle/Migrations/Version20210324152821.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20210324152821 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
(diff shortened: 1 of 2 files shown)


dvesh3:
@ramundomario yes, it makes sense! thanks for your PR. maybe we can also improve `Version20210324152822` by checking type like:
```php
$translationsTables = $db->fetchAll("SHOW FULL TABLES LIKE 'translations\_%'");
        foreach ($translationsTables as $table) {
            if ($table['Table_type'] != "BASE TABLE") {
                continue;
            }

            $translationsTable = current($table);

            if (!$schema->getTable($translationsTable)->hasColumn('type')) {
                $this->addSql('ALTER TABLE `'.$translationsTable.'` ADD COLUMN `type` varchar(10) DEFAULT NULL AFTER `key`');
            }
        }
```
same for downgrade...wdyt?

ramundomario:
@dvesh3 My first solution was to retrieve only tables in `Version20210324152822`, but I removed it as I read that the BC view would be removed in v10
Anyway adding an extra control doesn't hurt

WDYT of db filter like this?
`$translationsTables = $db->fetchAll("SHOW FULL TABLES WHERE Tables_in_{$db->getDatabase()} LIKE 'translations\_%' AND Table_type = 'BASE TABLE'");`

dvesh3:
> WDYT of db filter like this?
$translationsTables = $db->fetchAll("SHOW FULL TABLES WHERE Tables_in_{$db->getDatabase()} LIKE 'translations\_%' AND Table_type = 'BASE TABLE'");

@ramundomario looks good 👍 please change it in this PR. thanks!

ramundomario:
@dvesh3 done :wink:

dvesh3:
@ramundomario thanks a lot! 🙏

PR #9140, opened 2021-05-17
fix mail sender on GUI + enrich doc for email address format

Yivan:
Hello,

This PR fix a bug when using address email in settings of document email in back-office. Symfony wants a string and trigger an error because it was null (if we use the classic format `user@mail.com` to set emails in the GUI).

I added a hint to documentation to use the extended format to pass full username as well (I discovered it by parsing the code !! 😄)

Thanks.


dvesh3:
@Yivan thanks for your PR! please rebase it on branch 10.0 since it is a fix 😉

Yivan:
@dvesh3 This is done!

PR #9159, opened 2021-05-19
[Data Objects] Fix loading lazy loaded localized fields on save versio…

### LINKED ISSUE #9136: [Data Objects] Relation fields inside localized fields loss data when saved as draft
# Bug Report

IMPORTANT: DO NOT SUBMIT ISSUES WHICH WERE NOT VERIFIED ON THE LATEST VERSION OF PIMCORE! 
Bug reports that do not meet the conditions listed below will be closed/deleted without comment.

- [X] Do not submit reports for unsupported versions, supported versions can be viewed here https://github.com/pimcore/pimcore/milestones
- [X] This is not a security issue / vulnerability -> use [this form](https://pimcorehq.wufoo.com/forms/pimcore-security-report/) instead
- [X] Behavior is reproducible on https://demo.pimcore.fun/admin/ (admin/demo)
or on a clean install of the latest build (out of your project, to avoid problems with custom code, plugins, ...)
- [X] You're not able to fix the problem yourself and send us a pull request instead of filing an issue.
- [X] There's no existing ticket for the same issue
- [X] Remove unused sections from below and this bullet points, submit only clean reports

### Expected behavior
Save (as draft) object with updated data in a relation field inside localized fields & see changes
### Actual behavior
Save (as draft) object with updated data in a relation field inside localized fields & loss changes 
### Steps to reproduce

1. Add a relation field inside localized fields (Tested with Many-To-One Relation & Many-To-Many Relation)
Screenshot from 2021-05-17 16-12-19
2. Add an initial data to relation field (pdf-test.pdf)
3. Replace relation field data (pdf-test_1.pdf)
y0P1qm8Ca6

After saving the object as a draft and reloading, the text field (title) shows the updated data, while the relational field shows the old data 
In addition: when the object is saved and published, the old data are saved.

Found on Pimcore 6.9 but also reproduced on Pimcore X

dvesh3:
Resolves #9136

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Localizedfields.php
+++ b/models/DataObject/ClassDefinition/Data/Localizedfields.php
@@ -189,10 +189,11 @@ class Localizedfields extends Data implements CustomResourcePersistingInterface,
         $inheritanceAllowed = $class->getAllowInherit();
         $inherited = false;
 
-        $dataItems = $data->getInternalData($params['objectFromVersion'] ? false : true);
+        $loadLazy = !($params['objectFromVersion'] ?? false);
+        $dataItems = $data->getInternalData($loadLazy);
         foreach ($dataItems as $language => $values) {
             foreach ($this->getFieldDefinitions() as $fd) {
-                if ($fd instanceof LazyLoadingSupportInterface && $fd->getLazyLoading() && !($params['objectFromVersion'] ?? false)) {
+                if ($fd instanceof LazyLoadingSupportInterface && $fd->getLazyLoading() && $loadLazy) {
                     $lazyKey = $data->buildLazyKey($fd->getName(), $language);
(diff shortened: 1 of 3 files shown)


models/DataObject/ClassDefinition/Data/Localizedfields.php
```diff
@@ -189,10 +189,10 @@ private function doGetDataForEditMode($data, $object, &$fieldData, &$metaData, $
         $inheritanceAllowed = $class->getAllowInherit();
         $inherited = false;
 
-        $dataItems = $data->getInternalData(true);
+        $dataItems = $data->getInternalData($params['objectFromVersion'] ? false : true);
```
ramundomario: Could be simplified with `!$params['objectFromVersion']`?
Why is there no default value as done below?

ramundomario (changes requested):

ramundomario:
@dvesh3 anyway tested & it's working 😉

PR #9171, opened 2021-05-20
Remove problematic URLSlug validation

### LINKED ISSUE #9169: URLSlug should allow all valid URL characters
# Bug Report

### Expected behavior

A URL slug should allow [all URI/URL allowed characters](https://www.ietf.org/rfc/rfc3986.html).

### Actual behavior

Pimcore only allows URLSlugs which are compatible with admin element keys.

### Steps to reproduce

Create a URLSlug which may look like `/product/powerline/>36kVolt`

### Code example
Remove to fix:

https://github.com/pimcore/pimcore/blob/f31a2e44e662e1ca65be446c4145f13a9a7f13b3/models/DataObject/ClassDefinition/Data/UrlSlug.php#L199-L202 

Or add a new section in https://github.com/pimcore/pimcore/blob/cda1a4c6211cb65c7d708af3076f2529cbc6dabb/models/Element/Service.php#L1020 to deal with slugs rather than relying on document key rules.

NiklasBr:
Resolves #9169

### DIFF
--- a/models/DataObject/ClassDefinition/Data/UrlSlug.php
+++ b/models/DataObject/ClassDefinition/Data/UrlSlug.php
@@ -196,10 +196,6 @@ class UrlSlug extends Data implements CustomResourcePersistingInterface, LazyLoa
                         if (strlen($part) === 0) {
                             throw new Model\Element\ValidationException('Slug ' . $slug .' not valid');
                         }
-                        $sanitizedKey = Model\Element\Service::getValidKey($part, 'document');
-                        if ($sanitizedKey != $part) {
-                            throw new Model\Element\ValidationException('Slug part ' . $part .' not valid');
-                        }
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


brusch:
@NiklasBr we should still check for reserved characters, right? see https://www.ietf.org/rfc/rfc3986.html#section-2.2

NiklasBr:
I think we should be careful about that as well, for example Wikipedia-style URL:s would be disallowed then: 

https://en.wikipedia.org/wiki/The_Last_Days_of_Pompeii_(1913_film)

NiklasBr:
Does Pimcore/ExtJs support some kind of "notice" that could say something along the line "this works, but is not recommended" while still allowing the data to be saved?

brusch:
Not really, what we could do it displaying a tooltip or some kind of custom HTML notice. 
But I'm wondering if we should really support characters which are officially not supported, even if Wikipedia does 🤔
Maybe we should just `%` encode them as stated in rfc3986?

NiklasBr:
Ok. I have made an update here. What do you think?

NiklasBr:
```
Function str_starts_with not found.
Function str_contains not found.
```
Aha, [Symfony 3.4 does not include polyfill for php8](https://github.com/symfony/symfony/blob/3.4/composer.json#L33-L34)

jdreesen:
We could require the polyfill in Pimcore 6.9 ourselves. WDYT @brusch ?

But maybe... it's a little bit late now and not worth it anymore...

brusch:
I don't think it's worth to add a new dependency just for that.

Issue #9141, opened 2021-05-17
[Pimcore X] New debug system based on symfony and email debug too tighted + dev bundle registration.


Yivan:
Hello,

On Pimcore X, now debug mode is tightly tied to Symfony debug mode, and the debug email rely on this state. It makes it difficult to handle those separately... in fact I didn't manage : )

Use cases are:
- I want to active real 404/500 page (no debug mode) but keep emails in debug mode. For instance, for staging environment where customer wants to see/show to their direction the beautiful 404 page designers work hard on it (and not seeing Symfony debug information during pre-showcase!), but for sure we don't want sent emails to real customers !
- I am a developer or a project manager and I want to check, the errors designed page is ok and works fine, but I don't want to trigger real emails during my recipe.
- I am in production, I need to debug an email problem and make some touchy test, so I want to redirect emails on debug email but I don't want to display errors on production
- I am in production, I want to display errors (i configure firewall so just me can access), but I don't want customer lost cronjobs emails (i don't want they go to debug email address)

All of this could arrive one day or another (we had those needs several times).

On Pimcore X the only way I found to get something more fluent for developers/debuggers, is to:
- remove `isDebug` from the condition here : https://github.com/pimcore/pimcore/blob/10.x/lib/Kernel.php#L328 so dev bundles are available even in non-debug mode
- playing with APP_DEBUG=0 or APP_DEBUG=1 in .env.local
It works fine, as well if dev & stage environment for instance are in `getEnvironmentsForDevBundles`, a developer can easily switch debug/no debug to test the real errors page, but with the risk of sending email to real people during those test ! : (( He has to be careful, but it works.

_We needed to remove the condition isDebug because it prevents disabling debug mode on dev environment... (if we set APP_DEBUG=0 on dev environment, Symfony is not happy because he needs the profiler, and developers must change config yaml to adapt them, not really friendly...). This way he can just with a variable enable/disable debug mode (a little like Pimcore 6 with the BO checkbox)._

So this help a little to facilitate those use cases, but always risky concerning email delivery.

**To summarise, this issue is about two things:**
- removing isDebug condition `lib/Kernel.php`, `getEnvironmentsForDevBundles` should be sufficient and it helps switch debug mode off/on with APP_DEBUG.
[...]

PR #9178, opened 2021-05-20
Simpler error handling to display Pimcore or Symfony 404

Yivan:
Hello,

This is an improvement suggestion to make:
- 404 page error handling easier to manage for developers (testing real display)
- switching APP_DEBUG without any errors on dev environments (with fresh Symfony you can set `dev` with APP_DEBUG 0 or 1 it works out of the box, not for Pimcore)
- use more easily Symfony 404 pages if wanted instead Pimcore error page

Purpose is to make it easy, even on the `dev` environment, after this fix just doing:
- APP_DEBUG=0 => display 404 page
- if error document is set in BO: use this one
- if no error document is set : use the Symfony error page and way of customise it (https://symfony.com/doc/current/controller/error_pages.html)
- APP_DEBUG=1 => display debug page

This is linked to investigation made here: https://github.com/pimcore/pimcore/issues/9141#issuecomment-845465542

This is a proposition, which can make a **breaking change**, because on production instances of Pimcore having not the 404 page set:
- before: redirect to home page
- after: will display the Symfony 404 page (Oops! An Error Occurred):

image

I made this suggestion because, it seems more natural. Now all is handled by APP_DEBUG, so make it simpler for error page like it has been done for email make sense and coherence.

If not, can be closed ; )

Thanks.

### DIFF
--- a/bundles/CoreBundle/DependencyInjection/Configuration.php
+++ b/bundles/CoreBundle/DependencyInjection/Configuration.php
@@ -73,7 +73,7 @@ final class Configuration implements ConfigurationInterface
                             ->end()
                         ->end()
                     ->end()
-                    ->setDeprecated('The "%node%" option is deprecated.')
+                    ->setDeprecated('The "%node%" option is deprecated since Pimcore 10.1, it will be removed in Pimcore 11.')
                 ->end()
                 ->arrayNode('bundles')
                     ->addDefaultsIfNotSet()
--- a/bundles/CoreBundle/DependencyInjection/Configuration.php
+++ b/bundles/CoreBundle/DependencyInjection/Configuration.php
@@ -59,21 +59,6 @@ final class Configuration implements ConfigurationInterface
 
         $rootNode
             ->children()
-                ->arrayNode('error_handling')
-                    ->addDefaultsIfNotSet()
-                    ->children()
-                        ->booleanNode('render_error_document')
-                            ->info('Render error document in case of an error instead of showing Symfony\'s error page')
-                            ->defaultTrue()
-                            ->beforeNormalization()
-                                ->ifString()
-                                ->then(function ($v) {
-                                    return (bool)$v;
-                                })
-                            ->end()
-                        ->end()
-                    ->end()
-                ->end()
                 ->arrayNode('bundles')
                     ->addDefaultsIfNotSet()
                     ->children()
--- a/bundles/CoreBundle/DependencyInjection/Configuration.php
+++ b/bundles/CoreBundle/DependencyInjection/Configuration.php
@@ -59,6 +59,22 @@ final class Configuration implements ConfigurationInterface
 
         $rootNode
             ->children()
+                ->arrayNode('error_handling')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->booleanNode('render_error_document')
+                            ->info('Render error document in case of an error instead of showing Symfony\'s error page')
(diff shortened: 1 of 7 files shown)


brusch:
I wouldn't classify this as a breaking change in traditional sense, because it actually doesn't break anything, it just changes the content of the error page in case no error page is configured.

brusch:
The only thing we'd need to keep is the entry in the config tree and deprecate it (without having any use), otherwise it would break if you're using this in your project config.

brusch:
@Yivan 🏓

Yivan:
@brusch Thanks for looking into this suggestion. You mean I need to update my PR by adding deprecated here ?
- https://github.com/pimcore/pimcore/pull/9178/files#diff-bdc109f44f3002230731bc28875b9a9c016397113fac5510e7c51913e60fc50aL82
- https://github.com/pimcore/pimcore/pull/9178/files#diff-bdc109f44f3002230731bc28875b9a9c016397113fac5510e7c51913e60fc50aL82

If yes, I am not sure how to do it, adding a comment `# Deprecated` is ok ? : )
If not could you explain me or if you prefer/faster you can edit my PR directly (I will check how you do it for next time!).

bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -59,21 +59,6 @@ public function getConfigTreeBuilder()
 
         $rootNode
             ->children()
-                ->arrayNode('error_handling')
```
brusch: We cannot remove this config node, because this would be a breaking change. 
We need to keep it, even if it has no use anymore and deprecate it, so that we can remove it with Pimcore 11.  Use  `->setDeprecated('...')` for that.

brusch (changes requested):

bundles/CoreBundle/DependencyInjection/Configuration.php
brusch: @Yivan Sorry, I should have commented this info onto the actual code line from the beginning on 😊 I hope it's now clear why I mean. Thanks!

Yivan: @brusch Thanks for the info! This is done, I updated the PR.

bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
```diff
@@ -85,8 +85,6 @@ public function loadInternal(array $config, ContainerBuilder $container)
 
         $container->setParameter('pimcore.web_profiler.toolbar.excluded_routes', $config['web_profiler']['toolbar']['excluded_routes']);
 
-        $container->setParameter('pimcore.response_exception_listener.render_error_document', $config['error_handling']['render_error_document']);
```
brusch, line 88: I'm afraid we need to keep this parameter as well -> BC break. 
But please add a deprecation note: `// @deprecated since Pimcore 10.1, parameter will be removed in Pimcore 11`.

brusch (changes requested):

bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
Yivan: @brusch This is done (i changed by the way the node option deprecated message with same information you gave) !

brusch:
Looks good to me now, thanks a lot 👍

PR #9184, opened 2021-05-21
Fixes broken sorting in HTTP Errors view

NiklasBr:
Resolves #

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/MiscController.php
+++ b/bundles/AdminBundle/Controller/Admin/MiscController.php
@@ -612,7 +612,7 @@ class MiscController extends AdminController
 
         $limit = intval($request->get('limit'));
         $offset = intval($request->get('start'));
-        $sortInfo = json_decode($request->get('sort'), JSON_OBJECT_AS_ARRAY)[0];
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/MiscController.php
```diff
@@ -612,8 +612,9 @@ public function httpErrorLogAction(Request $request)
 
         $limit = intval($request->get('limit'));
         $offset = intval($request->get('start'));
-        $sort = $request->get('sort');
-        $dir = $request->get('dir');
+        $sortInfo = json_decode($request->get('sort'), JSON_OBJECT_AS_ARRAY)[0];
```
jdreesen: The flag `JSON_OBJECT_AS_ARRAY` is meant to be used as the 4th argument, but setting the second one to `true` does the same.

```suggestion
        $sortInfo = json_decode($request->get('sort'), true)[0];
```

NiklasBr: Thanks!

dvesh3:
thanks!

PR #9186, opened 2021-05-21
Youtube Video Error (no query)

Gerhard-Kanzler:
if the link looks like: https://youtu.be/bKgd_08Rpg0 there is no query part!


dvesh3:
@Gerhard-Kanzler thanks a lot!

PR #9188, opened 2021-05-21
fix: locales-all checking

adamtang79:
fix: Returned 'Montag' instead of 'Dienstag' while timedatectl show Asia/Kuala_Lumpur or UTC

  
  

Resolves #


brusch:
Thanks a lot!

PR #9192, opened 2021-05-21
Less aggressive stripping of backend data during indexing

NiklasBr:
Follow-up of #7154 which was closed due to branching.

## Additional info  
Also: replace tabs with spaces to prevent unwanted concatenation.

### DIFF
--- a/models/Search/Backend/Data.php
+++ b/models/Search/Backend/Data.php
@@ -510,7 +510,7 @@ class Data extends \Pimcore\Model\AbstractModel
      */
     protected function cleanupData($data)
     {
-        $data = preg_replace('/(<\?.*?(\?>|$)|<.*?>)/s', '', $data);
+        $data = preg_replace('/(<\?.*?(\?>|$)|<[^<]+>)/s', '', $data);
 
         $data = html_entity_decode($data, ENT_QUOTES, 'UTF-8');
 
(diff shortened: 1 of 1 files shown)


models/Search/Backend/Data.php
```diff
@@ -510,16 +510,13 @@ public function setDataFromElement(Element\ElementInterface $element)
      */
     protected function cleanupData($data)
     {
-        $data = strip_tags($data);
+        $data = preg_replace('/(<\?.*?(\?>|$)|<.*?>)/s', '', $data);
```
dvesh3: this ignores the text if there's a tag after "<" ... e.g. 
"In compliance with UNE 0064-1:2020 . Bacterial filtration efficiency (BFE) ≥ 95%. Breathability (differential pressure) **<60 Pa/cm2. Non-reusable. \<test\>**"

results in "compliance with UNE 0064-1 2020 Bacterial filtration efficiency (BFE) 95%. Breathability (differential pressure)"

jdreesen: Good catch. Changing the regex to this fixes it:
```suggestion
        $data = preg_replace('/(<\?.*?(\?>|$)|<[^<]+>)/s', '', $data);
```

Note: maybe the `+` should be replaced by `*` depending on whether `<>` should be matched or not.

dvesh3:
thanks a lot!

PR #9196, opened 2021-05-22
fix: chrome halt halfway on admin

adamtang79:
fix: firefox is okay but chrome will halt leaving admin page stuck

  
  

Resolves #


weisswurstkanone:
Is this reproducible on the demo instance ?

weisswurstkanone:
maybe we should consider to add it to the prepackaged ext-all

PR #9199, opened 2021-05-24
Update CoreHandler.php - performance

kkarkus:
Like title says,
Comparison: https://blackfire.io/profiles/compare/50b7a021-0f5d-4dee-9c59-75f835114062/graph

### DIFF
--- a/lib/Cache/Core/CoreHandler.php
+++ b/lib/Cache/Core/CoreHandler.php
@@ -386,24 +386,6 @@ class CoreHandler implements LoggerAwareInterface, CoreHandlerInterface
     {
         $this->saveQueue[$item->getKey()] = $item;
 
-        // order by priority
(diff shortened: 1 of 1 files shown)


kkarkus:
@brusch please take a look on this PR, I closed it because I don't have time to fix unit tests, you should be able easily to increase performance.

brusch:
@kkarkus Thanks for your thoughts, but actually there's a problem with your suggested change. The whole point of this code is to reference only 50 items in the queue, the others should be released, so that the GC can free memory if necessary. Doing that in `writeSaveQueue()`  make the code obsolete, because at this point we don't need to reduce the size of the queue, instead we could just skip after writing 50 items. 
So I'm afraid we need to look for a different solution. Maybe we do the sorting & splicing only every X times, so maybe if the queue is filled up to 100 items or so.

PR #9200, opened 2021-05-24
Set property in trait to public so that it can be loaded

### LINKED ISSUE #9198: [Class Definition] Default Value Generator Class in Region or Panel cannot be loaded
# Bug Report

### Expected behavior

I save a default generator value in class definition. After a reload of the page the value still exists.

### Actual behavior

A default generator is defined. After clicking the save button, the value is saved but not loaded anymore.

### Steps to reproduce

- Create a new input element
- Add a custom default generator class
- Save the form
- Reload the page
- Default generator class is missing

cmuench:
The property is currently protected which prevents that the value can be accessed from outside. If a class definition is loaded, the value is empty in the UI.

  
  

Resolves #9198


weisswurstkanone:
thx!

PR #9209, opened 2021-05-25
New Classificationstore groups added to variant is not visible

Daniel-Ateles:
When adding a new classificationstore group to a variant and then do a refresh of the variant in the admin interface results in the added group is gone and only the parents groups are visible.
The fault is that recursiveGetActiveGroupsIds throws away the variant groupid's instead of adding them to the parent ids.
The fix consists of adding the parent id:s to the variant instead of overwriting the array with the parent ids.

  
  

Resolves #


Daniel-Ateles:
This fix is important for our customer.

weisswurstkanone:
thanks a lot!

PR #9220, opened 2021-05-25
[Data objects] Disable mouse wheel scrolling for numeric, quantity value fields

BlackbitDevs:
Currently when you edit a data object and click a numeric / quantity value field and afterwards want to scroll down (or up) to another field the value in the field gets changed (due to the mouse wheel scroll). As in a PIM system I think nearly nobody uses the mouse scroll wheel to change a field's value this PR proposes to disable this "feature". 

And now I am waiting for comments who do use the mouse wheel to change numeric fields' values so that we have to add a checkbox in the field definition ;-)

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/numeric.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/numeric.js
@@ -26,6 +26,7 @@ pimcore.document.editables.numeric = Class.create(pimcore.document.editable, {
         config.value = data;
         config.name = id + "_editable";
         config.decimalPrecision = 20;
+        config.mouseWheelEnabled = false;
 
         if(config["required"]) {
             this.required = config["required"];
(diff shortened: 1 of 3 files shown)


NiklasBr:
That's the problem with ExtJs implementing what should be a `<input type="number">` as `<input type="text">` with a listener and all kinds of extra HTML... I'm supporting this 💯 :)

(regular `type="number"` [supports scroll wheel editing without JS as long as it has focus](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) and the cursor is on top of it)

brusch:
I'm supporting this as well, but currently ExtJS behaves exactly like regular `type="number"`. Scolling only works if the pointer is over the input.

brusch:
Btw.: if we agree to change the behavior, we should change the Numeric editable as well (documents).

brusch:
Thanks a lot! 👍

PR #9227, opened 2021-05-26
[Cache] Performance improvements

brusch:
Follow up to #9199

- Do not double-serialize data -> let Symfony care about serialization 
- Do not clean up queue on every `save()` call


brusch:
@kkarkus this is the follow up to #9199, but for Pimcore X, which is a bit different, but the same concept should work also for `6.x`.

PR #9229, opened 2021-05-26
[CLASSES] - Remove final modifier from all classes

roland4432:
Remove all final modifier from each and every class. It's not protecting code from overwriting, it's just making it harder to stick as close as possible to framework changes.


brusch:
@roland4432 thanks for your PR, however I don't think it's a good idea to remove `final` from all classes, as it definitely has it's advantages. If there are specific classes which are problematic, please let us discuss them.

roland4432:
Which classes do you think *really* deserve to be final?
It takes 5 seconds to overwrite them via composer. So imho it's easier to stick with @internal than making it _final_.
WDYT?

brusch:
Well, yes, overriding them via Composer is of course possible, but having them just `@internal` can lead to misunderstandings. Imho there's for instance absolutely no reason for having compiler passes not `final`.

roland4432:
Okay. Maybe you can at least agree to not having models being final?

Speaking of classes in the /models folder for instance...

brusch:
Not that many classes in `/models` are final, do you have an actual use case for overriding one of them?

roland4432:
Yes. We have a usecase where we did an override for the Asset Thumbnail generation to add a custom filter to show that images do not have any copyright set in the metadata.

So this is affecting _/models/Asset/Image/Thumbnail.php_ for example.

brusch:
@roland4432 I see, but Pimcore itself still uses `Pimcore\Model\Asset\Image\Thumbnail` so how does that actually help? 🤔

roland4432:
We are also overriding the ThumbnailProcessor and the Image Model. But these classes aren't final. So this is how things come together.

roland4432:
I don't get it. There is **no** point as framework / platform deliverer to restrict people from using your base code in a way you cannot foresee?
Final classes? Private methods? Could you please by any means leave it to the guys programming and matching customer requirements which part of pimcore can be overridden?

You cannot prevent anyone from overriding pimcore code. You are just making it unnecessary complicated.

brusch:
Symfony as a framework has private, final and internal classes/methods/properties as well. It just makes sense if we want to make clear which classes are intended to be used and which are e.g. just used by our user interface, etc.

roland4432:
Yes. But Symfony has an architecture, that makes it possible to override stuff via config.
It's a dependency injection based framework with an open mind set for alternatives in regards of convention meets configuration (e.g. CompilerPasses, ResponseTypes, Controllers, etc etc)
I totally get the idea of final classes and private methods. But if you make something final, than better have a config at hands that makes it possible to pass a custom implemention to a core function.

That's one of the few things why I get mad about pimcore. Sometimes you stumble upon architectural decissions that are hard to understand and overcome. E.g a Thumbnail generator implementation in this case.

Making a class final of course clarifies the usage for a programmer => don't mess with this stuff or things might go bad
But this has under no obligation to be your business what the community does with it. And if you don't make things configuratable in a sense of: this is my implementation, this is how things work for a core functionality, but here is a config, feel free to pass an implemention of your own to this method and change the behavior to one that better suits your customer requirements, then I am very unhappy with the current outcome.

Don't get me wrong. I didn't mean to tell, that _final class_ or _private function_ is a nogo, but not letting anybody change the cascade of used classes is.

dpfaffenbauer:
@roland4432 I am pretty much a heavy Pimcore User and Developer. I am using Pimcore longer than most here (except the Pimcore Devs obviously :P). In 9 Years of using Pimcore, I rarely had the use-case to override Pimcore Core Models. In all of these cases it was either wrong or for one and only one reason: hotfixing. Hotfixing a Pimcore Bug that was fixed upstream but for several reasons I was not able to update.

If you have the need to override core classes, recheck your architecture. You are definitely doing something wrong then.

roland4432:
Hey @dpfaffenbauer thx for dropping in to the party.
Just for the fun of it I will go in detail through the whole thought process and show a complete stack of code, which was used in 6.x and cannot be used in 10.x anymore due to private and final modifiers.

The customer wanted a brief method to not use images, that don't have a copyright metadata set. Since Assets cannot be unpublished, there should be a generic way to find a solution to make a subtle note to the user across all thumbnail configs, that the image does not have a copyright set and somewhat render the image, that the customer does not end up with any copyright infringements.

Idea:
Blur all images to a state where a google picture search would not find the image anymore and add a png overlay to all images without a copyright metadata showing that the copyright is missing.

Config:
```yaml
pimcore:
    models:
        class_overrides:
            'Pimcore\Model\Asset\Image': 'AppBundle\Model\DataObject\Image\Image'
```

Custom Image.php integration to implement a custom Thumbnail class which will finally use another Processor to get the job done.

```php
use Pimcore\Model\Asset\Image as BaseImage;

class Image extends BaseImage
{

    /**
     * Returns a path to a given thumbnail or an thumbnail configuration.
     *
     * @param string|array|Image\Thumbnail\Config $config
     * @param bool $deferred
     *
     * @return Thumbnail
     */
    public function getThumbnail($config = null, $deferred = true)
    {
        return new Thumbnail($this, $config, $deferred); #Thumbnail is also overriding the original Pimcore class
    }

    public function save() {
        $this->clearThumbnails(true);
        return parent::save();
    }
}
```

So then the Thumbnail generator will use our custom processor:
```php
use Pimcore\Logger;
use Pimcore\Model\Asset\Image\Thumbnail as BaseThumbnail;
use Symfony\Component\EventDispatcher\GenericEvent;

class Thumbnail extends BaseThumbnail
{
    /**
     * @param bool $deferredAllowed
     */
    public function generate($deferredAllowed = true)
    {
        $errorImage = PIMCORE_WEB_ROOT . '/bundles/pimcoreadmin/img/filetype-not-supported.svg';
        $deferred = false;
        $generated = false;

        if (!$this->asset) {
            $this->filesystemPath = $errorImage;
        } elseif (!$this->filesystemPath) {
            // if no correct thumbnail config is given use the original image as thumbnail
            if (!$this->config) {
                $this->filesystemPath = $this->asset->getRealFullPath();
            } else {
                try {
                    $deferred = $deferredAllowed && $this->deferred;
                    $this->filesystemPath = Processor::process($this->asset, $this->config, null, $deferred, true, $generated); #Processor again is overriden to finally come to the point
                } catch (\Exception $e) {
                    $this->filesystemPath = $errorImage;
                    Logger::error("Couldn't create thumbnail of image " . $this->asset->getRealFullPath());
                    Logger::error($e);
                }
            }
        }

        if ($this->hasListeners(AssetEvents::IMAGE_THUMBNAIL)) {
            \Pimcore::getEventDispatcher()->dispatch(AssetEvents::IMAGE_THUMBNAIL, new GenericEvent($this, [
                'deferred' => $deferred,
                'generated' => $generated
            ]));
        }
    }
}

```
and since we are now using a Processor of our own, we can now finally add the code to get things done:
```php
use Pimcore\File;
use Pimcore\Logger;
use Pimcore\Model\Asset;
use Pimcore\Model\Asset\Image\Thumbnail\Config;
use Pimcore\Model\Asset\Image\Thumbnail\Processor as BaseProcessor;
use Pimcore\Model\Tool\TmpStore;

class Processor extends BaseProcessor
{
...
    public static function process(Asset $asset, Config $config, $fileSystemPath = null, $deferred = false, $returnAbsolutePath = false, &$generated = false)
    {
        ...
            if(!$asset->getMetadata('copyright')){
                $transformations[] = [
                    'method' => 'gaussianBlur',
                    'arguments' => [
                        'radius' => 20,
                        'sigma' => 20,
                    ]
                ];
                $transformations[] = [
                    'method' => 'addOverlayFit',
                    'arguments' => [
                        'path' => '/web/static/img/no-copyright.png',
                        'composite' => 'COMPOSITE_DEFAULT',
                    ],
                ];
            }

            foreach ($transformations as $transformation) {
               ...
            }
        ...
    }
}
```

So this is how things ended up. Maybe there are another million solutions but this is what has been working damn fine in 6.x. So now in 10.x, I find that _Pimcore\Model\Asset\Image\Thumbnail_ is final. So I cannot override it anymore and extend it, but I need to copy the whole class and cut the inheritance.  
Ok, so I copy the class and have like a 99% code doubling in it for no reason.  
Let's take a look at the Processor now. While the Thumbnail\Processor is not final (so why not? or why is the Thumbnail class final then?) I can extend the original class, but now I stumble upon a lot of private static methods, which I would need when inheriting the Processor like:
* containsTransformationType
* getAllowedFormat  

Private static methods totally make sense from a performance state of view. Why not make them protected when the class is not final?
Why are classes from the same namespace sometimes final, sometimes not?

@dpfaffenbauer I know you are like the very grandmaster of pimcore. Your point of sometimes rethinking the architecture is legit. I know there are another million solutions and maybe better ways to get this done, what we are intended to do.  
But I don't see any point, why some things in the pimcore bundle are the way they are, therefore I wanted to talk about this, so.. here we are.

Happy for any input on this.

dpfaffenbauer:
why not create a twig extension that does that? You can pass in either a thumbnail-config name to the getThumbnail function, or an array as a config. If you use a existing thumbnail config, load the config, add the blur effect and pass it to the thumbnail config. With that, you are also not dependent on Pimcore API changes

dpfaffenbauer:
BTW: You are right in some points, Pimcore needs to overhaul its api and be more "injectable" and maybe "decorateable". But that doesn't change the fact that overriding stuff is not good.

roland4432:
Yes, I will come up with a solution. That's for sure (it's not even the point I am trying to make). I just hoped to have a little impact on how things are considered in the pimcore universe and that there are developers out there that don't take things right out-of-the-box but work with and extend them.

dpfaffenbauer:
You didn't get my point: You are using the wrong way to "don't take things right out-of-the-box". 

If I was Pimcore, I would also remove the option of class_overrides. Too much magic and bad practices there. Never saw the use-case for it in 9 years of Pimcore.

PR #9239, opened 2021-05-26
[Documents] Table Editable - do not allow columns to be draggable

### LINKED ISSUE #9138: Table Editable not working properly
# Bug Report
When creating an editable of type "Table" via "pimcore_table" it is possible to rearrange columns inside the rendered table editor via drag and drop in the admin UI.

### Expected behavior
The expected behaviour is that these changes trigger rearrangement of the data array layer and the change is recognized by the UI resulting in an unsaved document.

### Actual behavior
The change is displayed visually but saving the document and reloading afterwards restores the old state. The rendered frontend always shows the old state. Document preview also shows the old state, even when no reload of the editmode is done.

### Steps to reproduce
Add "pimcore_table" to your view.html add some columns and data to differentiate the column order, rearrange the columns via drag and drop.

### Code example
```html.twig
{{ pimcore_table("demo_table", {
    "defaults": {
        "cols": 2,
        "rows": 2,
        "data": [
            ["1", "A"],
            ["2", "B"],
            ["3", "C"]
        ]
    }
}) }}
```

dvesh3:
Resolves #9138

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/table.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/table.js
@@ -74,7 +74,8 @@ pimcore.document.editables.table = Class.create(pimcore.document.editable, {
                         allowBlank: true
                     }),
                     hideable: false,
-                    sortable: false
+                    sortable: false,
+                    draggable: false
                 });
             }
         }


dvesh3:
fixed by #9138

PR #9246, opened 2021-05-27
Fix missing service monolog processor when using custom environments

ghost:
Resolves #
https://github.com/pimcore/pimcore/discussions/9243
## Additional info  
This fixes the missing service definition when using custom environments such as "phpunit" and "behat" by recreating a fix that was already written.


dvesh3:
@davidmaes could you please rebase it on 10.0 branch? since it is a fix. thanks!

ghost:
Hi @dveshp3 that should be available here then: https://github.com/pimcore/pimcore/pull/9254

PR #9247, opened 2021-05-27
[Core] Data Object Param Converter - Suppress unpublished object warning in Admin preview mode

dvesh3:
Related to #5554

### DIFF
--- a/bundles/CoreBundle/Request/ParamConverter/DataObjectParamConverter.php
+++ b/bundles/CoreBundle/Request/ParamConverter/DataObjectParamConverter.php
@@ -17,6 +17,7 @@ namespace Pimcore\Bundle\CoreBundle\Request\ParamConverter;
 
 use Pimcore\Model\DataObject\AbstractObject;
 use Pimcore\Model\DataObject\Concrete;
+use Pimcore\Tool;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
 use Sensio\Bundle\FrameworkExtraBundle\Request\ParamConverter\ParamConverterInterface;
 use Symfony\Component\HttpFoundation\Request;
@@ -54,7 +55,7 @@ class DataObjectParamConverter implements ParamConverterInterface
         $object = $class::getById($value);
         if (!$object) {
             throw new NotFoundHttpException(sprintf('Invalid data object ID given for parameter "%s".', $param));
-        } elseif (!$object->isPublished()) {
+        } elseif (!$object->isPublished() && !Tool::isElementRequestByAdmin($request, $object)) {
             throw new NotFoundHttpException(sprintf('Data object for parameter "%s" is not published.', $param));
         }
 


bundles/CoreBundle/Request/ParamConverter/DataObjectParamConverter.php
```diff
@@ -54,7 +55,7 @@ public function apply(Request $request, ParamConverter $configuration)
         $object = $class::getById($value);
         if (!$object) {
             throw new NotFoundHttpException(sprintf('Invalid data object ID given for parameter "%s".', $param));
-        } elseif (!$object->isPublished()) {
+        } elseif (!Tool::isFrontendRequestByAdmin() && !$object->isPublished()) {
```
brusch: @dvesh3 We need to verify the login status in case `isFrontendRequestByAdmin() === true` otherwise you could expose confidential (not published) data just by adding one of those params: 
https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/lib/Http/RequestHelper.php#L162

brusch (changes requested):

bundles/CoreBundle/Request/ParamConverter/DataObjectParamConverter.php
dvesh3: fixed it by adding `Tool\Authentication::authenticateSession($request)` check 😉

bundles/CoreBundle/Request/ParamConverter/DataObjectParamConverter.php
```diff
@@ -55,7 +56,9 @@ public function apply(Request $request, ParamConverter $configuration)
         if (!$object) {
             throw new NotFoundHttpException(sprintf('Invalid data object ID given for parameter "%s".', $param));
         } elseif (!$object->isPublished()) {
-            throw new NotFoundHttpException(sprintf('Data object for parameter "%s" is not published.', $param));
+            if (!Tool::isFrontendRequestByAdmin() || !Tool\Authentication::authenticateSession($request)) {
```
brusch:
```suggestion
            if (!(Tool::isFrontendRequestByAdmin() && Tool\Authentication::authenticateSession($request))) {
```

brusch (changes requested):

bundles/CoreBundle/Request/ParamConverter/DataObjectParamConverter.php
dvesh3: @brusch we need to throw exception if `isFrontendRequestByAdmin() == false` or authenticate session when `isFrontendRequestByAdmin() == true`

bundles/CoreBundle/Request/ParamConverter/DataObjectParamConverter.php
```diff
@@ -55,7 +56,9 @@ public function apply(Request $request, ParamConverter $configuration)
         if (!$object) {
             throw new NotFoundHttpException(sprintf('Invalid data object ID given for parameter "%s".', $param));
         } elseif (!$object->isPublished()) {
-            throw new NotFoundHttpException(sprintf('Data object for parameter "%s" is not published.', $param));
+            if (!(Tool::isFrontendRequestByAdmin() && Tool\Authentication::authenticateSession($request))) {
```
brusch: @dvesh3 Please also consider the workspace permissions for the logged in user (`$user->isAllowed('view')`).

brusch (changes requested):

dvesh3:
this should utilize method introduced in https://github.com/pimcore/pimcore/pull/9308

PR #9250, opened 2021-05-27
8300 Change varchar size for quantity value unit ids

### LINKED ISSUE #8300: Units using varchar(50) IDs causes tables to reach maximum row length quickly

johannesdold:
Resolves #8300 

## Additional info
The column definition for quantity value units was recently changed from bigint(20) to varchar(50). If you have many quantity value data fields in one class, the maximum row length of the innodb table is reached very quickly. This pr increases the varchar length to 64. Then the values are stored in overflow pages. It also adds an exception, when creating a unit with length greater than 50.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/QuantityValueController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/QuantityValueController.php
@@ -134,6 +134,10 @@ class QuantityValueController extends AdminController
                     throw new \Exception('unit with ID [' . $id . '] already exists');
                 }
 
+                if (mb_strlen($id) > 50) {
+                    throw new \Exception('The maximal character length for the unit ID is 50 characters, the provided ID has ' . mb_strlen($id) . ' characters.');
+                }
+
                 $unit = new Unit();
                 $unit->setValues($data);
                 $unit->save();
--- a/models/DataObject/ClassDefinition/Data/QuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/QuantityValue.php
@@ -98,7 +98,7 @@ class QuantityValue extends Data implements ResourcePersistenceAwareInterface, Q
      */
     public $queryColumnType = [
         'value' => 'double',
-        'unit' => 'varchar(50)',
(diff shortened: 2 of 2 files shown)


johannesdold:
@weisswurstkanone Here is the new pr. Sorry for the hassle. I am on family holiday in the austrian mountains right now, and do not have my working laptop with me.

weisswurstkanone:
Enjoy it!

weisswurstkanone:
Hi @johannesdold what is the reason for increasing it to 64 but at the same time limiting it to 50.
In general I believe that something like varchar(15) is enough anyway. What do you think ?

johannesdold:
Hi @weisswurstkanone, I had explained the reason in the former pr, that was messed up due to my rebasing. Here my comment from before:

_Maybe another solution would be possible. If the dynamic row format is used, than varchar fields above a certain size are stored on overflow pages (https://mariadb.com/kb/en/troubleshooting-row-size-too-large-errors-with-innodb/). In our test, if I set the size to 64, I do not get the row size too large error. So maybe we should instead of reducing the size, increase it. What do you think?_

I added the check to the limit of 50, as this is the format that is used in the dedicated quantity value table. If we increase the format in the object tables to 64, we can safely keep the 50 and at the same time do not run in the table row size limit problem. If we set the format to varchar(15), the limit will be reached, however with a factor 3 later than with 50.

Also from the previous pr:

_To reproduce the issue on the demo, add a class with 40 quantity value fields and try to save._ 

With that, the problem will arise probably  at around 120 quantity value fields in a class. However, probably one would use than the classification store for that type of data.

weisswurstkanone:
thx

PR #9254, opened 2021-05-27
Fix missing service monolog processor psr log message 10.0

ghost:
Rebase for https://github.com/pimcore/pimcore/pull/9246#issuecomment-849559552


dvesh3:
@davidmaes thanks a lot 🙏

PR #9270, opened 2021-05-28
Localized error pages

dturopoli:
I think it would be nice if we had an option to define localized error pages. We had that problem on multiple projects, only way we managed to add some kind of localized error pages is via overriding of ResponseExceptionListener. So I would suggest something this:

Screenshot from 2021-05-28 17-39-41


bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -661,13 +661,20 @@ private function addDocumentsNode(ArrayNodeDefinition $rootNode)
                 ->scalarNode('default_controller')
                     ->defaultValue('App\\Controller\\DefaultController::defaultAction')
                 ->end()
-                ->arrayNode('error_pages')
```
brusch: This is acutally a BC break, we need to keep this as it is. 
I'd suggest to move the localized error pages under the `error_pages` node, so that it looks like: 
```yml
error_pages: 
    default: /foo/bar
    localized: 
        en: /en/error
        af: /af/error
```

bundles/CoreBundle/EventListener/ResponseExceptionListener.php
```diff
@@ -177,4 +187,56 @@ protected function logToHttpErrorLog(Request $request, $statusCode)
...
+            ['page', 'snippet', 'hardlink', 'link', 'folder']
+        );
+
+        if ($document) {
+            if ($document->getProperty('language')) {
+                $locale = $document->getProperty('language');
+            }
+        }
+
+        // If locale can't be determined or localized error page for given locale is not defined
```
brusch: We should consider the user-agent preferences as a fallback here as well (`\Symfony\Component\HttpFoundation\Request::getLanguages`)

brusch (changes requested):
Thanks a lot for your PR. I agree that this would be a very nice addition, however there are some things we need to consider (see inline comments). 

Additionally: to be consistent we'd need that for site specific error pages as well. 
Thanks!

bundles/CoreBundle/DependencyInjection/Configuration.php
dturopoli: Adjusted as you requested.

dturopoli:
Additionally: to be consistent we'd need that for site specific error pages as well.

For this part I'm not sure where to put inputs for localized error pages for site. Should we add them to popup for editing site?

brusch:
> Additionally: to be consistent we'd need that for site specific error pages as well.
> 
> For this part I'm not sure where to put inputs for localized error pages for site. Should we add them to popup for editing site?

Yes, I think that would be the right place. But maybe we should change the popup to a tab, then we have more space 😊

dturopoli:
Added localized error pages to site edit popup

Screenshot from 2021-06-01 09-16-41

bundles/CoreBundle/EventListener/ResponseExceptionListener.php
dturopoli: Adjusted as you requested.

dvesh3:
@dturopoli please check that you have unrelated commits in this PR. thanks!

brusch:
@dturopoli can you please resolve the conflicts in `ResponseExceptionListener.php`? Thanks again 😊

dturopoli:
Conflicts resolved

brusch (approved):

brusch:
Thanks a lot, very nice addition! 👍

PR #9275, opened 2021-05-30
Docs: Fix link to DoctrineMigrationsBundle

danmichaelo:
DoctrineMigrationsBundle uses a different versioning, so 5.2 doesn't exist. We can either link to "3.1" or to "current". I suggest using "current", and rather change the link to a fixed version in the future if Pimcore at some point no longer supports the current version.


dvesh3:
@danmichaelo could you please rebase on branch 10.0? thanks!

PR #9288, opened 2021-05-31
Symfony 5.3 compatibility - use UserInterface::getUserIdentifier() if exists.

weisswurstkanone:
see https://github.com/symfony/symfony/commit/8afd7a3765ff1258ef46ca70a1102dbe51762644#diff-e483d6cdb224e0e551d992dd9cad2a7c060f20adcdb5487b2b2b42001adfb380L77

### DIFF
--- a/lib/Security/Encoder/Factory/UserAwareEncoderFactory.php
+++ b/lib/Security/Encoder/Factory/UserAwareEncoderFactory.php
@@ -54,8 +54,10 @@ class UserAwareEncoderFactory extends AbstractEncoderFactory
             ));
         }
 
-        if (isset($this->encoders[$user->getUsername()])) {
-            return $this->encoders[$user->getUsername()];
+        $username = method_exists($user, 'getUserIdentifier') ? $user->getUserIdentifier() : $user->getUsername();
+
+        if (isset($this->encoders[$username])) {
+            return $this->encoders[$username];
         }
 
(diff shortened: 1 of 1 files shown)


dvesh3:
Other failing tests fixed in https://github.com/pimcore/pimcore/pull/9297

jdreesen:
The method `UserInterface::getUsername()` has just been re-added in [Symfony `v5.3.1`](https://github.com/symfony/symfony/releases/tag/v5.3.1) for compatibility reasons: https://github.com/symfony/symfony/pull/41493

wouterj:
But this PR is still valid. Calling `getUsername()` will result in a deprecation notice (as the method will be removed in 6.0), so it's better to update all code to not call it when the new method is already available.

lib/Security/Encoder/Factory/UserAwareEncoderFactory.php
```diff
@@ -54,8 +54,17 @@ public function getEncoder($user)
             ));
         }
 
-        if (isset($this->encoders[$user->getUsername()])) {
-            return $this->encoders[$user->getUsername()];
+        $username = null;
```
aarongerig, line 57: Heck, I would even rename the variable to `$userIdentifier` or `$identifier`, since `$username` is not applicable anymore. 😅

PR #9294, opened 2021-06-01
Fixes `aspectration` to `aspectratio` in wysiwyg.js

NiklasBr:

The incorrect parameter can be seen when dragging an asset into a WYSIWYG field and inspecting image properties.


dvesh3:
@NiklasBr thanks!

PR #9296, opened 2021-06-01
Add symfony/doctrine-messenger dependency to fix Symfony 5.3 compatibility issues

brusch:


brusch:
Seems to be obsolete.

brusch:
Sometimes it still appears, no idea why, but adding this package fixes it, so I guess it doesn't hurt.

jdreesen:
What appears? Why is this necessary?

brusch:
This happens when building the container: 

```
  RuntimeException {#6618
    #message: "The autoloader expected class "Symfony\Component\Messenger\Transport\Doctrine\DoctrineTransportFactory" to be defined in file "/home/demos/build/opensource-master/source-master/vendor/compose
  r/../symfony/symfony/src/Symfony/Component/Messenger/Transport/Doctrine/DoctrineTransportFactory.php". The file was found but the class was not in it, the class name or namespace probably has a typo."
    #code: 0
    #file: "./vendor/symfony/symfony/src/Symfony/Component/ErrorHandler/DebugClassLoader.php"
    #line: 401
    trace: {
      ./vendor/symfony/symfony/src/Symfony/Component/ErrorHandler/DebugClassLoader.php:401 { …}
      ./vendor/symfony/symfony/src/Symfony/Component/ErrorHandler/DebugClassLoader.php:358 { …}
      Symfony\Component\ErrorHandler\DebugClassLoader->loadClass() {}
      ./vendor/doctrine/doctrine-bundle/DependencyInjection/DoctrineExtension.php:995 { …}
      ./vendor/doctrine/doctrine-bundle/DependencyInjection/DoctrineExtension.php:79 { …}
      ./vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Compiler/MergeExtensionConfigurationPass.php:76 { …}
      ./vendor/symfony/symfony/src/Symfony/Component/HttpKernel/DependencyInjection/MergeExtensionConfigurationPass.php:39 { …}
      ./vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Compiler/Compiler.php:91 { …}
      ./vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/ContainerBuilder.php:749 { …}
      ./vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:545 { …}
      ./vendor/pimcore/pimcore/lib/Kernel.php:223 { …}
      ./vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:786 { …}
      ./vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:125 { …}
      ./vendor/pimcore/pimcore/lib/Kernel.php:202 { …}
      ./vendor/pimcore/pimcore/lib/Bootstrap.php:256 { …}
      ./vendor/pimcore/pimcore/lib/Bootstrap.php:60 { …}
      ./bin/console:44 {
        › /** @var \Pimcore\Kernel $kernel */
        › $kernel = \Pimcore\Bootstrap::startupCli();
        › $application = new \Pimcore\Console\Application($kernel);
      }
    }
  }
```

brusch:
Ok, I've found the reason, `dist` is different from `source` for `symfony/symfony`. 
So actually https://github.com/symfony/symfony/archive/refs/tags/v5.3.0.zip  is different from what you see here https://github.com/symfony/symfony/tree/v5.3.0 
Maybe @fabpot can have a look at this.

fabpot:
Short answer: you should never use the `symfony/symfony` package. Since a long time now, individual components should be used instead, we "only" keep `symfony/symfony` as a mono-repo where development happens (and for BC reasons, but we might remove the package from Packagist at some point).

brusch:
Ok, I see. 
Just for redcord, it's because of: https://github.com/symfony/symfony/blob/5.4/.gitattributes#L4

brusch:
So we need that fix for now, until #9312 is fixed

Issue #6906, opened 2020-08-05
[DataObjects] Move common stuff from DataObjectController & VariantsController to trait


weisswurstkanone:

weisswurstkanone:
related #6898

PR #9307, opened 2021-06-02
"Show in Tree" only works if sorting is alphabetical

### LINKED ISSUE #8476: "Show in Tree" only works if sorting is alphabetical
Hi guys, 

when I am on an object and I run the "Show in Tree" function, it only works if the objects are sorted alphabetically. Is that the way it is supposed to be?

Best regards

Pimcore-Version: 6.8.10

_Originally posted by @polymed in https://github.com/pimcore/pimcore/discussions/8474_

weisswurstkanone:
resolves #8476

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -1125,7 +1125,7 @@ class DataObjectController extends ElementControllerBase implements EventedContr
 
 
     /**
-     * @param DataObject\AbstractObject $parent
+     * @param DataObject\AbstractObject $parentObject
      * @param string $currentSortOrder
      */
     protected function reindexBasedOnSortOrder(DataObject\AbstractObject $parentObject, string $currentSortOrder) {
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -956,7 +956,6 @@ class DataObjectController extends ElementControllerBase implements EventedContr
             $sortOrder = $request->get('childrenSortOrder');
 
             $currentSortBy = $object->getChildrenSortBy();
-            $currentSortOrder= $object->getChildrenSortOrder();
 
             $object->setChildrenSortBy($sortBy);
             $object->setChildrenSortOrder($sortOrder);
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -952,8 +952,27 @@ class DataObjectController extends ElementControllerBase implements EventedContr
     {
         $object = DataObject::getById($request->get('id'));
         if ($object) {
-            $object->setChildrenSortBy($request->get('sortBy'));
-            $object->setChildrenSortOrder($request->get('childrenSortOrder'));
+            $sortBy = $request->get('sortBy');
+            $sortOrder = $request->get('childrenSortOrder');
+
+            $currentSortBy = $object->getChildrenSortBy();
+            $currentSortOrder= $object->getChildrenSortOrder();
+
+            $object->setChildrenSortBy($sortBy);
+            $object->setChildrenSortOrder($sortOrder);
+
+            if ($currentSortBy != $sortBy) {
+                $user = Tool\Admin::getCurrentUser();
+
+                if (!$user->isAdmin()) {
+                    return $this->json(['success' => false, 'message' => 'Changing the sort method is only allowed for admin users']);
+                }
+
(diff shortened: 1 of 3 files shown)


bundles/CoreBundle/Resources/translations/en.json
```diff
@@ -938,5 +938,6 @@
   "preview_item": "Preview Item",
   "version_currently_saved_in_database": "Currently saved in database (but not published)",
   "preview_item": "Preview Item",
-  "deprecated_feature": "<p style='color: orange;'>Deprecated feature. Use %s<br>instead</p>"
+  "deprecated_feature": "<p style='color: orange;'>Deprecated feature. Use %s<br>instead</p>",
+  "reindex_warning": "Chaning it from alphabatical sort order to index will reindex all item. Do you want to continue?"
```
blankse:
```suggestion
  "reindex_warning": "Changing it from alphabetical sort order to index will reindex all item. Do you want to continue?"
```

jdreesen: Additionally:
```suggestion
  "reindex_warning": "Chaning it from alphabatical sort order to index will reindex all items. Do you want to continue?"
```

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -952,8 +952,27 @@ public function changeChildrenSortByAction(Request $request)
     {
         $object = DataObject::getById($request->get('id'));
         if ($object) {
-            $object->setChildrenSortBy($request->get('sortBy'));
-            $object->setChildrenSortOrder($request->get('childrenSortOrder'));
+            $sortBy = $request->get('sortBy');
+            $sortOrder = $request->get('childrenSortOrder');
+
+            $currentSortBy = $object->getChildrenSortBy();
+            $currentSortOrder= $object->getChildrenSortOrder();
```
dvesh3: unused var `$currentSortOrder`? should we pass this in `reindexBasedOnSortOrder()`?

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -1169,6 +1123,126 @@ protected function updateIndexesOfObjectSiblings(DataObject\AbstractObject $upda
         }
     }
 
+
+    /**
+     * @param DataObject\AbstractObject $parentObject
+     * @param string $currentSortOrder
+     */
+    protected function reindexBasedOnSortOrder(DataObject\AbstractObject $parentObject, string $currentSortOrder) {
```
dvesh3, line 1130: I would move this to model..maybe `AbstractObject` and mark this as internal

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -1169,6 +1123,126 @@ protected function updateIndexesOfObjectSiblings(DataObject\AbstractObject $upda
...
+            }
+            $latestVersion->save();
+        }
+    }
+
+    /**
+     * @param DataObject\AbstractObject $updatedObject
+     * @param int $newIndex
+     */
+    protected function updateIndexesOfObjectSiblings(DataObject\AbstractObject $updatedObject, $newIndex)
```
dvesh3, line 1189: I would move this to model

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -1073,78 +1092,13 @@ public function updateAction(Request $request)
         return $this->adminJson(['success' => $success]);
     }
 
-    /**
-     * @param DataObject\AbstractObject $updatedObject
-     * @param int $newIndex
-     */
-    protected function updateIndexesOfObjectSiblings(DataObject\AbstractObject $updatedObject, $newIndex)
-    {
+    private function executeInsideTransaction(callable $fn) {
```
dvesh3, line 1094: I would move this to model

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
weisswurstkanone: 1) thx
2) I wouldn't take the value from the databse but the current value from the admin ui instead

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
weisswurstkanone: Not directly related, but I would do this in one go. There is even a task for this 
https://github.com/pimcore/pimcore/issues/6906

dvesh3 (changes requested):
exception when changing order to index (manually)
image

weisswurstkanone:
Unable to reproduce.

dvesh3 (approved):

PR #9308, opened 2021-06-02
[Elements] Add method to verify element preview request by admin user

### LINKED ISSUE #5164: [DataObject][Preview] Add option to verify in controller if request comes from object preview
- check if request comes from object preview
- check for existing admin session etc. to verify if request 

then also update demo, see https://github.com/pimcore/demo/issues/154

dvesh3:
Resolves #5164

### DIFF
--- a/lib/Tool.php
+++ b/lib/Tool.php
@@ -18,6 +18,7 @@ namespace Pimcore;
 use GuzzleHttp\RequestOptions;
 use Pimcore\Http\RequestHelper;
 use Pimcore\Localization\LocaleServiceInterface;
+use Pimcore\Model\Element;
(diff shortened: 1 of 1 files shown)


lib/Tool.php
```diff
@@ -328,6 +329,29 @@ public static function isFrontendRequestByAdmin(Request $request = null)
...
+            return false;
+        }
+
+        $user = Tool\Authentication::authenticateSession($request);
+
+        if (!($user && $element->isAllowed('view', $user))) {
+            return false;
+        }
+
+        return true;
```
jdreesen: For me, it's more clear when it's written like this:

```suggestion
        return $user && $element->isAllowed('view', $user);
```

PR #9311, opened 2021-06-02
[Ecommerce] Return mock product in CartItem when actual product is not available

### LINKED ISSUE #9280: [Ecommerce] Error with deleted product in cart
# Bug Report

If you add an product to the cart and delete it in the backend it came to following error:
```
Call to a member function getOSName() on null
```

### Expected behavior
No error. Product should be removed.

dvesh3:
Resolves #9280

### DIFF
--- a/bundles/EcommerceFrameworkBundle/CartManager/SessionCart.php
+++ b/bundles/EcommerceFrameworkBundle/CartManager/SessionCart.php
@@ -15,7 +15,9 @@
 
 namespace Pimcore\Bundle\EcommerceFrameworkBundle\CartManager;
 
+use Pimcore\Bundle\EcommerceFrameworkBundle\Model\CheckoutableInterface;
 use Pimcore\Bundle\EcommerceFrameworkBundle\Tools\SessionConfigurator;
+use Pimcore\Model\DataObject;
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
 
 class SessionCart extends AbstractCart implements CartInterface
@@ -175,6 +177,14 @@ class SessionCart extends AbstractCart implements CartInterface
 
         // set current cart
         foreach ($this->getItems() as $item) {
+
+            // if a product is deleted in backend, then remove item from SessionCart
+            $product = DataObject::getById($item->getProductId());
+            if (!$product instanceof CheckoutableInterface) {
+                $this->removeItem($item->getItemKey());
+                continue;
+            }
+
             $item->setCart($this);
 
             if ($item->getSubItems()) {
--- a/bundles/EcommerceFrameworkBundle/CartManager/SessionCart.php
+++ b/bundles/EcommerceFrameworkBundle/CartManager/SessionCart.php
@@ -15,9 +15,7 @@
 
 namespace Pimcore\Bundle\EcommerceFrameworkBundle\CartManager;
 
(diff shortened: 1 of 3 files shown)


bundles/EcommerceFrameworkBundle/CartManager/SessionCart.php
```diff
@@ -175,6 +177,14 @@ public function __wakeup()
 
         // set current cart
         foreach ($this->getItems() as $item) {
+
+            // if a product is deleted in backend, then remove item from SessionCart
+            $product = DataObject::getById($item->getProductId());
```
fashxp: we should call `getProduct()` here ... then also PHPStan is happy ...

fashxp (changes requested):

bundles/EcommerceFrameworkBundle/CartManager/SessionCart.php
dvesh3: done.

dvesh3: Changed approach here as it impacts the performance coz when updating cart, it checks for product exists by calling `DataObject::getById()` which loads the product. 

so now we use `MockProduct` for CartItem in getProduct() calls when actual product doesn't exist.

bundles/EcommerceFrameworkBundle/Model/MockProduct.php
```diff
@@ -0,0 +1,115 @@
...
+use Pimcore\Bundle\EcommerceFrameworkBundle\PriceSystem\PriceInterface;
+use Pimcore\Bundle\EcommerceFrameworkBundle\PriceSystem\PriceSystemInterface;
+use Pimcore\Model\DataObject\Concrete;
+
+/**
+ * Mock Product class which should be used as a product when actual product is not available in the System.
+ */
+class MockProduct extends Concrete implements ProductInterface, IndexableInterface, CheckoutableInterface
+{
+    public function getAvailabilitySystemName(): ?string
```
fashxp, line 31: I would return null where ever possible. 
so also with `getAvailabilitySystemName`, `getPriceSystemImplementation`, `getAvailabilitySystemImplementation`, `getOSPriceInfo`, `getOSAvailabilityInfo`, `getPriceSystemName`

fashxp (changes requested):

bundles/EcommerceFrameworkBundle/Model/MockProduct.php
dvesh3: It is not possible to return null for these methods due to incompatibility between CartItemInterface & CheckoutableInterface. I have created an issue for Pimcore 11 https://github.com/pimcore/pimcore/issues/9405

fashxp (approved):

blankse:
@dvesh3 @fashxp AbstractCart::isValidCartItem() now is always true and can be removed?

dvesh3:
@blankse valid point....but then let's just mark it deprecated coz removing it, is a BC break!

blankse:
@dvesh3 BC break is also change the behavior. Should we check to MockProduct then?

dvesh3:
@blankse 👍 added MockProduct check https://github.com/pimcore/pimcore/commit/9787cc1cabd1fe79e71ab2dfc8240251c135e16a, so the behavior is same as before. thanks for review 😉

PR #9314, opened 2021-06-02
Classificationsstore KeysConfig - GetByName should return list of keys or Key names should be unique

### LINKED ISSUE #8518: Classificationsstore KeysConfig - GetByName should return list of keys or Key names should be unique.
### Expected behavior
As there could be keys with the same name. The function GetByName should return a list or Key names should be unique.

### Actual behavior
The first key with a matching name is returned. Even though it might be "deleted" (marked as not enabled in db)

### Steps to reproduce
Create two keys with identical names. 
Use keyconfig getByName function to get the keys.
Observe that only one of the keys is returned.

Maybe delete that key in pimcore backend.
Try getByName again.
Observe that the deleted key is still returned.

weisswurstkanone:
resolves #8518

### DIFF
--- a/models/DataObject/Classificationstore/KeyConfig.php
+++ b/models/DataObject/Classificationstore/KeyConfig.php
@@ -16,6 +16,7 @@
 namespace Pimcore\Model\DataObject\Classificationstore;
 
 use Pimcore\Cache;
+use Pimcore\Db;
 use Pimcore\Event\DataObjectClassificationStoreEvents;
 use Pimcore\Event\Model\DataObject\ClassificationStore\KeyConfigEvent;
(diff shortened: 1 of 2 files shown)


dvesh3:
@weisswurstkanone I believe this does not solve the problem reported on the issue which is "_The first key with a matching name is returned. Even though it might be "**deleted**" (marked as not enabled in db)_".

Adding a new method is fine so that it is possible to get all key configs with the same name, however i would still adapt condition for not returning deleted key from db. wdyt?

dvesh3:
please see https://github.com/pimcore/pimcore/issues/8518#issuecomment-855739167

PR #9316, opened 2021-06-02
Variants tab in object shows only default language for classificationstore values

### LINKED ISSUE #9304: Variants tab in object shows only default language for classification store values
### Expected behavior
When I click on the Variants Tab in Pimcore on an Object, then I want to see the data with the current language.

### Actual behavior
At the moment there is a bug with the classification store. You see only the default values in the view. (You can edit it and then the new value is in the right language) 
grafik

On the folders it is working. It's only a problem in the variants tab.

### Steps to reproduce
1. Add a classification store
2. Add a object with classification store and variants in it
3. Add in the classification store any input field
4. Go to the object on the variants tab
5. Add english and german in the pimcore settings

weisswurstkanone:
resolves #9304

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/VariantsController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/VariantsController.php
@@ -70,8 +70,13 @@ class VariantsController extends AdminController
     {
         // get list of variants
 
-        if ($request->get('language')) {
-            $request->setLocale($request->get('language'));
+        $requestedLanguage = $request->get('language') ?? null;
+        if ($requestedLanguage) {
+            if ($requestedLanguage != 'default') {
+                $request->setLocale($requestedLanguage);
+            }
+        } else {
+            $requestedLanguage = $request->getLocale();
         }
 
         if ($request->get('xaction') == 'update') {
--- a/bundles/AdminBundle/Controller/Admin/DataObject/VariantsController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/VariantsController.php
@@ -86,15 +86,6 @@ class VariantsController extends AdminController
             $object = DataObject\Concrete::getById($data['id']);
             $class = $object->getClass();
 
-            $requestedLanguage = $allParams['language'] ?? null;
-            if ($requestedLanguage) {
-                if ($requestedLanguage != 'default') {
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/VariantsController.php
```diff
@@ -70,8 +70,13 @@ public function getVariantsAction(Request $request, GridHelperService $gridHelpe
     {
         // get list of variants
 
-        if ($request->get('language')) {
-            $request->setLocale($request->get('language'));
+        $requestedLanguage = $request->get('language') ?? null;
```
blankse:
```suggestion
        $requestedLanguage = $request->get('language');
```

blankse:
Line 89-96 can than be removed:
https://github.com/pimcore/pimcore/blob/e6e3e165bc83f6c71f309341fe13fd5cd91c08c3/bundles/AdminBundle/Controller/Admin/DataObject/VariantsController.php#L89-L96

brusch:
@niederberger-qw can you please confirm this fixes the issue? Thanks a lot!

niederberger-qw:
@brusch 
Unfortunately no success. I cannot detect any change.
grafik

Normal field are displayed. Only the fields from the classification store are in the default language (Nothing displayed).

weisswurstkanone:
Hello @blankse , did you try the patch. Notice that the PR didn't got merged yet.

brusch:
@blankse I guess @weisswurstkanone actually wanted to ping @niederberger-qw so don't mind 😉 
@niederberger-qw please apply [the patch of this PR](https://patch-diff.githubusercontent.com/raw/pimcore/pimcore/pull/9316.patch) before testing. Thanks for your feedback 😊

niederberger-qw:
grafik

Did apply the patch, but again... In the variants tab only the default language of the classification store value is displayed.

dvesh3:
@weisswurstkanone the fix is not working for me too..
image

dvesh3:
> @weisswurstkanone the fix is not working for me too..
> image

https://github.com/pimcore/pimcore/pull/9316/commits/5cbe196e7157675f55c07da11d198465c39ba043 this should fix the problem by passing `$requestedLangauge` to gridObjectData() call

PR #9319, opened 2021-06-03
Fixed all KernelEvent::isMasterRequest() deprecation messages

h4rdc0m:
Fixes the deprecation messages in the log about isMasterRequest deprecation, changed all of them into isMainRequest as per Symfony standard.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -2696,8 +2696,8 @@ class AssetController extends ElementControllerBase implements KernelControllerE
      */
     public function onKernelControllerEvent(ControllerEvent $event)
     {
-        $isMasterRequest = $event->isMasterRequest();
-        if (!$isMasterRequest) {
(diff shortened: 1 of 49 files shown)


jdreesen:
This would require updating to Symfony 5.3 as minimum version, as the method `isMainRequest()` has [been introduced](https://github.com/symfony/symfony/blob/5.3/UPGRADE-5.3.md#httpkernel) in that version.

h4rdc0m:
Seems that the test suite installs symfony 5.2 instead of 5.3 when you install per instruction

h4rdc0m:
> This would require updating to Symfony 5.3 as minimum version, as the method `isMainRequest()` has [been introduced](https://github.com/symfony/symfony/blob/5.3/UPGRADE-5.3.md#httpkernel) in that version.

Problem is that when you install per instruction Symfony 5.3 is installed, makes reading logs very hard.
"symfony/symfony": "^5.2.0", means that anything up to 6.0.0 will be installed on new installs.

brusch:
I think you should just exclude the deprecation messages in Monolog.

PR #9322, opened 2021-06-03
Apply fallback languages after changing setting

BlackbitDevs:
Currently there is the following problem:
1. Configure `de` and `en` as languages in system settings
1. Create class with localized field
2. Create object of this class
3. Enter data in the localized field for `de`
4. Add `de` as fallback language for `en`
5. First problem: Data in query table is not updated
6. Save object again -> second problem: data for `en` in query table is still not updated because Localizedfields support dirty detection and when the localized field's value does not change, nothing gets changed in the database.

So currently the only way to apply the fallback language to all existing objects is to create a script which sets the field value to mark it `dirty` and then save it.

Same problem applies when a language once had a fallback language and you remove it. In the query table the fallback entries will stay.

This PR changes this behaviour by immediately applying fallback languages to the query tables.


fashxp:
does this scale to millions of data records? 🤔

brusch:
@fashxp @BlackbitNeueMedien I don't think that it scales well that way. Maybe it's better to provide this as a command and put a note to the UI to run this manually? 
Actually this is probably anyway necessary, e.g. for automated (multi-environment) deployment scenarios ... wdyt?

brusch:
@BlackbitNeueMedien 🏓😊

brusch:
I'll close this PR for now. @BlackbitNeueMedien If you agree with my last comment, it would be nice if you could provide a PR for that. Thanks a lot!

PR #9349, opened 2021-06-07
Classificationstore Multiselct -> empty values

ctippler:
Problem: Empty values are not saved in the classificationstore when a Multiselect is used.

Reproduce:
1. select some values in the multiselect -> save -> data is saved correctly
2. unselect all values -> save object -> and reload the object in the admin interface
-> The old values appear because the empty selection was not saved.


weisswurstkanone:
please rebase on 10.0

ctippler:
creating a new pr

PR #9350, opened 2021-06-07
Update Redirect Listing Dao.php

kkarkus:
There's the possibility that when dao tries to load redirects with a certain condition it will end up with a list of mixed-up nulls and redirects objects because somebody removed some before dao loaded it.


brusch:
LGTM, thanks a lot! I guess that's as well relevant for some other entities 🤔

kkarkus:
@brusch Yea, probably the same case for other entities :) But we struggled with this in our project, cause we have a huge amount of redirects and sometimes we got an error saying RedirectHandler (one of the anonymous functions) got null value instead of Redirect object, causing 500 on every request.

PR #9355, opened 2021-06-07
[Core] composer.json: Migrate to individual symfony/* packages

### LINKED ISSUE #9312: composer.json: Migrate to individual symfony/* packages
See: https://github.com/pimcore/pimcore/pull/9296#issuecomment-852919765

dvesh3:
Resolves #9312


doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -2,6 +2,7 @@
 
 ## 10.1.0
 - [InstallBundle] Installer preconfiguration path changed from `app\config/installer.yml` to `config/installer.yaml`
+- [Core] composer.json: `symfony/symfony` package requirement has been replaced by `symfony/*` individual bundles. **Note for Bundles**: if you are using `symfony/symfony` dependency, it will now conflict with package `pimcore/pimcore`. Please move your bundle requirements to Symfony individual bundles.
```
```suggestion
- [Core] composer.json: `symfony/symfony` package requirement has been replaced by `symfony/*` individual bundles. **Note for Bundles**: if you are using `symfony/symfony` dependency, it will now conflict with package `pimcore/pimcore`. Please move your bundle requirements to Symfony individual component packages.
```

.github/ci/scripts/symfony-require-dev.sh
```diff
@@ -0,0 +1,4 @@
+SYMFONY_VERSION=${SYMFONY_VERSION:-local}
```
jdreesen, line 1: What's `local` for a version?

brusch: @jdreesen you are right, but actually it doesn't hurt as there is still the empty check for `SYMFONY_VERSION` in the workflow.yml's. 
@dvesh3 I think we should fix that anyway, not that it causes confusion in future 😊

dvesh3: yes, a leftover from debugging. fixed it on 10.x branch 😉 thanks!

kaurov:
The reason of it:
you should never use the symfony/symfony package. Since a long time now, individual components should be used instead, we "only" keep symfony/symfony as a mono-repo where development happens (and for BC reasons, but we might remove the package from Packagist at some point).


e.g. conflicts:

- https://github.com/symfony/symfony/blob/5.4/.gitattributes#L4
- dist is different from source for symfony/symfony. So actually https://github.com/symfony/symfony/archive/refs/tags/v5.3.0.zip is different from what you see here https://github.com/symfony/symfony/tree/v5.3.0

PR #9359, opened 2021-06-07
fix asset frontend URL with and without frontend prefixes

bramalho:
This change allows to use the `source` of `frontend_prefixes` and keeps the context scheme and host

## Additional info  
if we use a different asset file system and we want to change the URL prefix:
```
pimcore:
    assets:
        frontend_prefixes:
            source: 'https://storage.googleapis.com/%env(GCP_BUCKET_NAME)%'
```

Or if you change the `frontendPath` in the event `pimcore.frontend.path.asset`

The frontend URL will always be something like: `url: "http://localhosthttps://storage.googleapis.com/....`

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -873,6 +873,14 @@ class Asset extends Element\AbstractElement
         $path = urlencode_ignore_slash($path);
 
         $prefix = \Pimcore::getContainer()->getParameter('pimcore.config')['assets']['frontend_prefixes']['source'];
+        if (!$prefix) {
+            $prefix = sprintf(
+                '%s://%s',
+                \Pimcore::getContainer()->getParameter('router.request_context.scheme'),
+                \Pimcore::getContainer()->getParameter('router.request_context.host')
+            );
+        }
+
         $path = $prefix . $path;
 
         $event = new GenericEvent($this, [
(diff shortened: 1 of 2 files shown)


brusch:
@bramalho thanks a lot for your PR. Can you please rebase onto the `10.0` branch and change the target of the PR to `10.0` as well (it's a bug fix). Thanks a lot!

models/Asset.php
```diff
@@ -873,6 +873,14 @@ public function getFrontendFullPath()
         $path = urlencode_ignore_slash($path);
 
         $prefix = \Pimcore::getContainer()->getParameter('pimcore.config')['assets']['frontend_prefixes']['source'];
+        if (!$prefix) {
+            $prefix = sprintf(
```
brusch, line 877: This would add the scheme and hostname even if not necessary, e.g. in frontend templates. On big pages this could result in a significant bigger response body. Maybe it's better to just hande this in the `AssetController`?

brusch (changes requested):

models/Asset.php
bramalho: yes, I also think it should be handled there, however, in the `AssetController` I can only validate it there is a schema, or check it this `['frontend_prefixes']['source']` is not empty

brusch: I'd check for a schema with a regex, because the frontend path could be also changed by an event and not only by config 😊

bramalho: ok, going to close this one and open a new MR, thanks

PR #9360, opened 2021-06-07
fix asset frontend public URL with and without frontend prefixes

bramalho:
This change allows to use the `source` of `frontend_prefixes` and keeps the context scheme and host

## Additional info  
if we use a different asset file system and we want to change the URL prefix:
```
pimcore:
    assets:
        frontend_prefixes:
            source: 'https://storage.googleapis.com/%env(GCP_BUCKET_NAME)%'
```

Or if you change the `frontendPath` in the event `pimcore.frontend.path.asset`

The frontend URL will always be something like: `url: "http://localhosthttps://storage.googleapis.com/....`

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -191,7 +191,9 @@ class AssetController extends ElementControllerBase implements KernelControllerE
         $data['fileExtension'] = File::getFileExtension($asset->getFilename());
         $data['idPath'] = Element\Service::getIdPath($asset);
         $data['userPermissions'] = $asset->getUserPermissions();
-        $data['url'] = $request->getSchemeAndHttpHost() . $asset->getFrontendFullPath();
+        $data['url'] = preg_match('/^http(s)?:\\/\\/.+/', $asset->getFrontendFullPath()) ?
+            $asset->getFrontendFullPath() :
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -191,7 +191,9 @@ public function getDataByIdAction(Request $request, EventDispatcherInterface $ev
         $data['fileExtension'] = File::getFileExtension($asset->getFilename());
         $data['idPath'] = Element\Service::getIdPath($asset);
         $data['userPermissions'] = $asset->getUserPermissions();
-        $data['url'] = $request->getSchemeAndHttpHost() . $asset->getFrontendFullPath();
+        $data['url'] = preg_match('/^http(s)?:\\/\\/.+/', $asset->getFrontendFullPath()) ?
+            $asset->getFrontendFullPath() :
+            $request->getSchemeAndHttpHost() . $asset->getFrontendFullPath();
```
```suggestion
        $frontendPath = $asset->getFrontendFullPath();
        $data['url'] = preg_match('/^http(s)?:\\/\\/.+/', $frontendPath) ?
            $frontendPath :
            $request->getSchemeAndHttpHost() . $frontendPath;
```

brusch:
Just added a small micro performance optimization, since every call of `\Pimcore\Model\Asset::getFrontendFullPath()`  builds the URL from scratch, including event dispatching, ... 😊

PR #9365, opened 2021-06-07
[DataObjects] Fix Service::getOptionsForSelectField not working with optionsProviderClass

### LINKED ISSUE #9340: DataObject\Service::getOptionsForSelectField not working with optionsProviderClass
# Bug Report

I already created an discussion (https://github.com/pimcore/pimcore/discussions/9172) for this but with no response yet. As far as I see this seems to be a bug.

### Expected behavior

Calling `DataObject\Service::getOptionsForMultiSelectField` or `DataObject\Service::getOptionsForSelectField` on an object class with optionsProviderClass returns array with elements.

### Actual behavior

Null is returned.

**Warning: foreach() argument must be of type array|object, null given**

### Steps to reproduce

1. Add select field to object class
2. Define an optionsProviderClass which returns values
3. Use `DataObject\Service::getOptionsForMultiSelectField` or `DataObject\Service::getOptionsForSelectField` to get the values.

### Code example

Responsible is the follwing code in `DataObject\Service::getOptionsForSelectField`:

```php
            if ($definition instanceof ClassDefinition\Data\Select || $definition instanceof ClassDefinition\Data\Multiselect) {
                $_options = $definition->getOptions();

                foreach ($_options as $option) {
                    $options[$option['value']] = $option['key'];
                }
            }
```

**`$definition->getOptions()`** returns NULL in my case and causes the Warning.

I use this in the `getEditableDialogBoxConfiguration`.

The optionsProviderClass works correct in the edit mode of an Pimcore\Model\DataObject\Article item.

# Feature Request

dvesh3:
Resolves #9340

## Additional info  
- not sure how this worked on ver. 6.8

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -849,7 +849,16 @@ class Service extends Model\Element\Service
             }
 
             if ($definition instanceof ClassDefinition\Data\Select || $definition instanceof ClassDefinition\Data\Multiselect) {
-                $_options = $definition->getOptions();
+                $optionsProvider = DataObject\ClassDefinition\Helper\OptionsProviderResolver::resolveProvider(
+                    $definition->getOptionsProviderClass(),
+                    DataObject\ClassDefinition\Helper\OptionsProviderResolver::MODE_MULTISELECT
+                );
+
+                if ($optionsProvider instanceof DataObject\ClassDefinition\DynamicOptionsProvider\MultiSelectOptionsProviderInterface) {
+                    $_options = $optionsProvider->getOptions(['fieldname' => $definition->getName()], $definition);
+                } else {
+                    $_options = $definition->getOptions();
+                }
 
                 foreach ($_options as $option) {
                     $options[$option['value']] = $option['key'];


weisswurstkanone (approved):

Yivan:
@dvesh3 Thanks for this fix!
But looking at the code, it is like fix has been done for dynamic multiselect only but not dynamic select:
- here we check for both: https://github.com/pimcore/pimcore/blob/fff6ff5c1aa194b4268ccb7cf6bc8fcce7665147/models/DataObject/Service.php#L851
- but here we only check for dynamic multiselect, not dynamic select: https://github.com/pimcore/pimcore/blob/fff6ff5c1aa194b4268ccb7cf6bc8fcce7665147/models/DataObject/Service.php#L857
Is all ok ?
This is just in case of because we don't check for `DataObject\ClassDefinition\DynamicOptionsProvider\SelectOptionsProviderInterface` but only `DataObject\ClassDefinition\DynamicOptionsProvider\MultiSelectOptionsProviderInterface`, if all is ok you can ignore my comment.

dvesh3:
Hi, @Yivan!  `SelectOptionsProviderInterface` extends `MultiSelectOptionsProviderInterface` so it should work for select fields as well.  thanks anyway for review 😉

Yivan:
Thanks, you are right @dvesh3 , I should have checked this, sorry for the noise !

PR #9378, opened 2021-06-08
[Data objects] Throw ValidationException in case same item got assigned multiple times to a relational field

BlackbitDevs:
The `checkValidity` methods for Many to many (object) relations and blocks usually throw `ValidationException`s - with one exception: When the same item gets assigned multiple times although the relational field does not allow this just `Exception` gets thrown. This PR changes this to always throw `ValidationException`.

An an example: https://github.com/pimcore/pimcore/blob/18a268ae3993daea952e8b1ac94155cdc0b38814/models/DataObject/ClassDefinition/Data/ManyToManyObjectRelation.php#L358-L388


weisswurstkanone:
thx!

PR #9414, opened 2021-06-11
Correct AbstractCart::isValidCartItem Warning Message

blankse:
MockProduct should have the ID from the missing Object

### DIFF
--- a/bundles/EcommerceFrameworkBundle/CartManager/AbstractCart.php
+++ b/bundles/EcommerceFrameworkBundle/CartManager/AbstractCart.php
@@ -952,7 +952,7 @@ abstract class AbstractCart extends AbstractModel implements CartInterface
     protected static function isValidCartItem(CartItemInterface $item)
     {
         $product = $item->getProduct();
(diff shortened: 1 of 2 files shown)


bundles/EcommerceFrameworkBundle/CartManager/AbstractCart.php
```diff
@@ -952,11 +952,11 @@ protected function validateVoucherTokenReservations()
     protected static function isValidCartItem(CartItemInterface $item)
     {
         $product = $item->getProduct();
-        if ($product instanceof CheckoutableInterface && !$product instanceof MockProduct) {
```
dvesh3: I would still keep `CheckoutableInterface` check, in case somebody has overridden the getProduct() method.

blankse: @dvesh3 Okay, I add it

dvesh3:
@blankse thanks a lot!

PR #9418, opened 2021-06-11
Redirect CSV export needs same handling as RedirectHandler

Daniel-Ateles:
Because the CSV-export is not a site request the function Document->getFullPath will return the target with the root document path.
I essentially copied what the RedirectHandler does.
(The same problem exists for RedirectController that shows the wrong target for auto-created redirects)

  
  

Resolves #

### DIFF
--- a/lib/Routing/Redirect/Csv.php
+++ b/lib/Routing/Redirect/Csv.php
@@ -23,6 +23,7 @@ use League\Csv\Statement;
 use League\Csv\Writer;
 use Pimcore\Model\Document;
 use Pimcore\Model\Redirect;
+use Pimcore\Model\Site;
 use Pimcore\Tool\Admin;
 use Pimcore\Tool\ArrayNormalizer;
 use Pimcore\Tool\Text;
--- a/lib/Routing/Redirect/Csv.php
+++ b/lib/Routing/Redirect/Csv.php
@@ -77,7 +77,12 @@ class Csv
                 $document = Document::getById((int)$redirect->getTarget());
 
                 if ($document) {
-                    $target = $document->getRealFullPath();
+                    $targetSiteId = $redirect->getTargetSite();
+                    $targetSite = Site::getById($targetSiteId);
+                    $rootDocument = $targetSite->getRootDocument();
+                    $rootDocumentPath = $rootDocument->getFullPath();
+                    $target = $document->getFullPath();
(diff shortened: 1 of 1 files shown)


lib/Routing/Redirect/Csv.php
```diff
@@ -77,7 +78,12 @@ public function createExportWriter(Redirect\Listing $list): Writer
                 $document = Document::getById((int)$redirect->getTarget());
 
                 if ($document) {
-                    $target = $document->getRealFullPath();
+                    $targetSiteId = $redirect->getTargetSite();
+                    $targetSite = Site::getById($targetSiteId);
+                    $rootDocument = $targetSite->getRootDocument();
```
dvesh3: this will break csv export when exporting main site entries where targetSite is null

dvesh3 (changes requested):
we should also fix site specific targets for auto-created redirects in `RedirectController`

dvesh3:
@Daniel-Ateles could you please fix this https://github.com/pimcore/pimcore/pull/9418#pullrequestreview-684888130? thanks!

Daniel-Ateles:
> @Daniel-Ateles could you please fix this [#9418 (review)](https://github.com/pimcore/pimcore/pull/9418#pullrequestreview-684888130)? thanks!

Will hopefuly start looking on it today.

dvesh3:
> > @Daniel-Ateles could you please fix this [#9418 (review)](https://github.com/pimcore/pimcore/pull/9418#pullrequestreview-684888130)? thanks!
> 
> Will hopefuly start looking on it today.

sounds good! just an heads up that the release 6.9.6 (last release for 6.x) is planned for 23rd June.

Daniel-Ateles:
> we should also fix site specific targets for auto-created redirects in `RedirectController`

New PR for the controller: https://github.com/pimcore/pimcore/pull/9523

brusch:
As this PR is now it would break the import. 
See also my comment here: https://github.com/pimcore/pimcore/pull/9523#issuecomment-865954020 

So I'd leave it as is is or include the main domain of the site, so that it is clear to which document the redirect refers to. The we'd also need to consider this in the import and also discuss which protocol should be used.

brusch:
I'll close this for now, as it isn't anyway a bug and therefore not relevant for `6.9`. 
However we can continue the discussion how we can improve that for `10.x` (considering my thoughts from above). 
Thanks!

PR #9434, opened 2021-06-14
Fix `Pimcore 6 CE` EOL in readme

jdreesen:
According to the milestone for [Pimcore 6.9.6](https://github.com/pimcore/pimcore/milestone/130) the EOL for Pimcore 6.x is 23.06.21.


weisswurstkanone:
Hi @jdreesen , I think it should be better to put this into 10.0

jdreesen:
Done. But as `v10.x` is the default branch which is shown when browsing the repo on GitHub it should be merged back into that branch soon, I think.

dvesh3:
> Done. But as `v10.x` is the default branch which is shown when browsing the repo on GitHub it should be merged back into that branch soon, I think.

It will be reflected on 10.1 branch, once it is tagged.

PR #9437, opened 2021-06-14
Bugfix: getCurrentIndex returns invalid index

mittererr:
Calling `{% set brickKey = brick.getEditable().getCurrentIndex() %}` in view returns incorrect key. This PR fixed this issue for me.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "abytemore/pimcore",
+  "name": "pimcore/pimcore",
   "type": "library",
   "description": "Content & Product Management Framework (CMS/PIM/E-Commerce)",
   "keywords": [
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "pimcore/pimcore",
+  "name": "abytemore/pimcore",
   "type": "library",
   "description": "Content & Product Management Framework (CMS/PIM/E-Commerce)",
   "keywords": [
--- a/models/Document/Editable/Areablock.php
+++ b/models/Document/Editable/Areablock.php
@@ -669,7 +669,7 @@ class Areablock extends Model\Document\Editable implements BlockInterface
      */
     public function getCurrentIndex()
     {
-        return $this->indices[$this->getCurrent()]['key'] ?? null;
+        return $this->indices[$this->current]['key'] ?? null;
     }
 
     /**


brusch:
@mittererr Thanks a lot for your PR! 
Can you please give us a bit more insights how to reproduce the issue? (example) With the current code snippet it's quite hard 😉 
Also, please rebase your PR onto the `10.0` as it is a bug-fix. 
Thanks a lot!

mittererr:
I will try to clearify the situation. 

Steps to reproduce the issue:

1. Create a really basic Areabrick:
```PHP
<?php
namespace App\Document\Areabrick;

use Pimcore\Extension\Document\Areabrick\AbstractTemplateAreabrick;
use Pimcore\Extension\Document\Areabrick\EditableDialogBoxConfiguration;
use Pimcore\Extension\Document\Areabrick\EditableDialogBoxInterface;
use Pimcore\Model\Document\Editable;

class AppDebug extends AbstractTemplateAreabrick implements EditableDialogBoxInterface {

    CONST DEFAULT_ICON_PATH = '/bundles/pimcoreadmin/img/flat-white-icons/';

    /**
     * Get template location
     *
     * @see \Pimcore\Extension\Document\Areabrick\AbstractTemplateAreabrick::getTemplateLocation()
     */
    public function getTemplateLocation()
    {
        return static::TEMPLATE_LOCATION_GLOBAL;
    }

    /**
     * @return string
     */
    public function getName () {
        return 'Debug';
    }

    /**
     * @return string
     */
    public function getDescription () {
        return 'Element';
    }

    /**
     * @return string|null
     */
    public function getIcon() {
        return self::DEFAULT_ICON_PATH .'grid.svg';
    }

    /**
     * @param Editable $area
     * @param \Pimcore\Model\Document\Editable\Area\Info|null $info
     * @return EditableDialogBoxConfiguration
     */
    public function getEditableDialogBoxConfiguration(Editable $area, ?Editable\Area\Info $info): EditableDialogBoxConfiguration
    {
        $dialogBoxConfig = new EditableDialogBoxConfiguration();
        $dialogBoxConfig->setWidth(800);
        $dialogBoxConfig->setHeight(800);
        $dialogBoxConfig->setItems([
            [
                'type' => 'input',
                'name' => 'text',
                'label' => 'Testlabel'
            ]
        ]);

        return $dialogBoxConfig;
    }
}
```

2. with a really basic view
```TWIG
{% set editables = document.getEditables() %}

{% block content %}
    <div>
        <p>
            <b>DEBUG:</b><br>
            getCurrentIndex(): {{ brick.getEditable().getCurrentIndex() is null ? 'null' : brick.getEditable().getCurrentIndex() }}<br>
            getCurrent(): {{ brick.getEditable().getCurrent() is null ? 'null' : brick.getEditable().getCurrent() }}<br>

            {{ dump(editables) }}
            <hr/>
        </p>
    </div>

{% endblock %}
```

3. Place this brick three times on a document and check the output
grafik

**Problem:**
If you want to access the `editables` array, maybe by 
```TWIG
{% set myProp = editables['content:' ~ brick.getEditable().getCurrentIndex() ~ '.text'] %}
```

it will fail because in the first case `getCurrentIndex()` returns null and in the other cases the return index is to low (1 instead of 2)?

brusch:
Ok, now it's clearer to me, thanks 😊
The problem is that `getCurrentIndex()` only work in the context of the areablock, but not from outside. 

So your change would break the entire areablock functionality, see here: 
https://github.com/pimcore/pimcore/blob/ecd55571f92ccbe7837211fca192c9c607c526cc/models/Document/Editable/Areablock.php#L145

I don't know what you're trying to achieve, but `getCurrentIndex()` is not the right tool for that.

PR #9438, opened 2021-06-14
Fixed missing language flag for chinese traditional

ascheider:
Currently if you create zh_Hant language, you have no flag in administration. _unknown.svg is shown instead. This pr fixes missing mapping for the flag.


weisswurstkanone:
thx!

PR #9439, opened 2021-06-14
[Data Object] Introduce additional interfaces for data-types

### LINKED ISSUE #8827: [Data Object] Introduce additional interfaces for data-types
For the current functionalities still `method_exists()` is used, that should be changed to proper interfaces: 

```
rewriteIds
classSaved
preGetData
preSetData
enrichLayoutDefinition
enrichFieldDefinition

marshalBeforeEncryption
unmarshalAfterDecryption

// ... possibly some more ... 
```

Incl. BC-layer.

weisswurstkanone:
resolves #8827


models/DataObject/ClassDefinition/Data/IdRewriterInterface.php
```diff
@@ -0,0 +1,38 @@
...
+     *  "asset" => array(...)
+     * )
+     *
+     * @param mixed $object
+     * @param array $idMapping
+     * @param array $params
+     *
+     * @return array
+     */
+    public function rewriteIds($object, $idMapping, $params = []);
```
jdreesen: I think the contract should be typed:

```suggestion
    public function rewriteIds(mixed $object, array $idMapping, array $params = []): array;
```

models/DataObject/ClassDefinition/Data/IdRewriterInterface.php
```diff
@@ -0,0 +1,38 @@
...
+     * array(
+     *  "document" => array(
+     *      SOURCE_ID => TARGET_ID,
+     *      SOURCE_ID => TARGET_ID
+     *  ),
+     *  "object" => array(...),
+     *  "asset" => array(...)
+     * )
+     *
+     * @param mixed $object
```
jdreesen: Is this really `mixed` or rather `object`?

models/DataObject/ClassDefinition/Data/PreGetDataInterface.php
```diff
@@ -0,0 +1,31 @@
...
+
+interface PreGetDataInterface
+{
+    /**
+     * @param Concrete|Localizedfield|AbstractData|\Pimcore\Model\DataObject\Fieldcollection\Data\AbstractData $object
+     * @param array $params
+     *
+     * @return array
+     */
+    public function preGetData($object, $params = []);
```
jdreesen:
```suggestion
    public function preGetData(Concrete|Localizedfield|AbstractData|\Pimcore\Model\DataObject\Fieldcollection\Data\AbstractData $object, array $params = []): array;
```

models/DataObject/ClassDefinition/Data/PreSetDataInterface.php
```diff
@@ -0,0 +1,30 @@
...
+interface PreSetDataInterface
+{
+    /**
+     * @param DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|DataObject\Fieldcollection\Data\AbstractData $object
+     * @param array|null $data
+     * @param array $params
+     *
+     * @return mixed
+     */
+    public function preSetData($object, $data, $params = []);
```
jdreesen:
```suggestion
    public function preSetData(DataObject\Concrete|DataObject\Localizedfield|DataObject\Objectbrick\Data\AbstractData|DataObject\Fieldcollection\Data\AbstractData $object, ?array $data, array $params = []): mixed;
```

models/DataObject/ClassDefinition/Data/IdRewriterInterface.php
weisswurstkanone: after glancing over the code it can be anything.

models/DataObject/ClassDefinition/Data/AfterDecryptionUnmarshallerInterface.php
```diff
@@ -0,0 +1,28 @@
...
+interface AfterDecryptionUnmarshallerInterface
+{
+    /**
+     * @param mixed $value
+     * @param mixed $object
+     * @param array $params
+     *
+     * @return mixed
+     */
+    public function unmarshalAfterDecryption(mixed $value, mixed $object = null, $params = []) : mixed;
```
jdreesen:
```suggestion
    public function unmarshalAfterDecryption(mixed $value, mixed $object = null, array $params = []) : mixed;
```

models/DataObject/ClassDefinition/Data/BeforeEncryptionMarshallerInterface.php
```diff
@@ -0,0 +1,28 @@
...
+interface BeforeEncryptionMarshallerInterface
+{
+    /**
+     * @param mixed $value
+     * @param mixed $object
+     * @param array $params
+     *
+     * @return mixed
+     */
+    public function marshalBeforeEncryption(mixed $value, mixed $object = null, $params = []) : mixed;
```
jdreesen:
```suggestion
    public function marshalBeforeEncryption(mixed $value, mixed $object = null, array $params = []) : mixed;
```

models/DataObject/ClassDefinition/Data/FieldDefinitionEnrichmentInterface.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Pimcore\Model\DataObject\ClassDefinition\Data;
+
+interface FieldDefinitionEnrichmentInterface
+{
+    /**
+     * @param array $context
+     * @return Data
+     */
+    public function enrichFieldDefinition($context = []) : Data;
```
jdreesen:
```suggestion
    public function enrichFieldDefinition(array $context = []) : Data;
```

dpfaffenbauer:
I think that method_exist should then also trigger a deprecation note?

lib/DataObject/BlockDataMarshaller/EncryptedField.php
```diff
@@ -109,7 +111,11 @@ public function encrypt($data, $params = [])
             $fd = $params['fieldDefinition'];
             $delegateFd = $fd->getDelegate();
 
-            if (method_exists($delegateFd, 'marshalBeforeEncryption')) {
+            if ($delegateFd instanceof BeforeEncryptionMarshallerInterface || method_exists($delegateFd, 'marshalBeforeEncryption')) {
+                if (!$delegateFd instanceof BeforeEncryptionMarshallerInterface) {
+                    @trigger_error(sprintf('Usage of method_exists is deprecated since version 10.1 and will be removed in Pimcore 11.' .
+                        'Implement the %s interface instead.', BeforeEncryptionMarshallerInterface::class), E_USER_DEPRECATED);
```
jdreesen: What about using [Symfony's deprecation contract](https://github.com/symfony/deprecation-contracts) for triggering deprecations in a somehow standardized way? It should already be installed with the `symfony/contracts` package.

jdreesen: So that's a no? Or was it resolved automatically?

weisswurstkanone: Let me first go over the rest of the open interfaces and then I will do everything in one go.

weisswurstkanone: done

bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
```diff
@@ -1023,7 +1024,12 @@ public function addCollectionsAction(Request $request)
                     $context['keyId'] = $keyData->getKeyId();
                     $context['groupId'] = $groupId;
                     $context['keyDefinition'] = $definition;
-                    if (method_exists($definition, 'enrichLayoutDefinition')) {
+                    //TODO Pimcore 11: remove method_exists BC layer
+                    if ($definition instanceof DataObject\ClassDefinition\Data\LayoutDefinitionEnrichmentInterface || method_exists($definition, 'enrichLayoutDefinition')) {
+                        if (!$definition instanceof DataObject\ClassDefinition\Data\LayoutDefinitionEnrichmentInterface) {
+                            trigger_deprecation('pimcore/pimcore', '10.1', sprintf('Usage of method_exists is deprecated since version 10.1 and will be removed in Pimcore 11.' .
+                                'Implement the %s interface instead.', DataObject\ClassDefinition\Data\LayoutDefinitionEnrichmentInterface::class));
```
jdreesen: `trigger_deprecation()` uses `printf` internally, just pass additional arguments after the message:

```suggestion
                            trigger_deprecation('pimcore/pimcore', '10.1', 'Usage of method_exists is deprecated since version 10.1 and will be removed in Pimcore 11.' .
                                'Implement the %s interface instead.', DataObject\ClassDefinition\Data\LayoutDefinitionEnrichmentInterface::class);
```

jdreesen: This is also true for all other calls to `trigger_deprecation` ;)

weisswurstkanone: true, let's create a follow up a follow up ticket. not even sure if it makes sense to invest time on deprecation notices style.

dvesh3 (commented):
@weisswurstkanone BC Break? if there are custom data types extending and overriding methods from the core, then introducing typehints will break the existing systems.

jdreesen:
If it's a BC break depends on where the types are added. 

Parameter types can be added without any problems, because the user can just leave them out (contravariance), which means Pimcore can add them and the user doesn't have to - it still works and thus is no BC break.

For return types it's the other way around (covariance). The user can add them even if Pimcore doesn't has them. But if Pimcore adds them, the user *must* add them, too (which would be a BC break).

See: https://www.php.net/manual/en/language.oop5.variance.php
and: https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters

models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
```diff
@@ -574,7 +575,7 @@ public function save($object, $params = [])
     /**
      * {@inheritdoc}
      */
-    public function preGetData($object, $params = [])
+    public function preGetData(mixed $object, $params = []): mixed
```
dvesh3:
```suggestion
    public function preGetData(/** mixed */ $object, /** array */ $params = []) /**: mixed */
```

models/DataObject/ClassDefinition/Data/Fieldcollections.php
```diff
@@ -544,13 +536,13 @@ public function getGetterCode($class)
...
 
-        if (method_exists($this, 'preGetData')) {
-            $code .= "\t" . '/** @var \\' . static::class . ' $fd */' . "\n";
-            $code .= "\t" . '$fd = $this->getClass()->getFieldDefinition("' . $key . '");' . "\n";
-            $code .= "\t" . '$data = $fd->preGetData($this);' . "\n\n";
-        } else {
-            $code .= "\t" . '$data = $this->' . $key . ";\n\n";
-        }
+        // TODO Pimcore 11: remove method_exists BC layer
+        // TODO this should be needed at all as preGetData is always there
```
dvesh3:
```suggestion
        // TODO else part should not be needed at all as preGetData is always there
```

models/DataObject/ClassDefinition/Data/IdRewriterInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+namespace Pimcore\Model\DataObject\ClassDefinition\Data;
+
+use Pimcore\Model\DataObject\Concrete;
+
+interface IdRewriterInterface
+{
+    /**
+     * @param mixed $container any container type like Concrete, Localizedfields, Fieldcollection, etc...
```
dvesh3, line 33:
```suggestion
     * Rewrites id from source to target, $idMapping contains
     * array(
     *  "document" => array(
     *      SOURCE_ID => TARGET_ID,
     *      SOURCE_ID => TARGET_ID
     *  ),
     *  "object" => array(...),
     *  "asset" => array(...)
     * )
     *
     * @param mixed $container any container type like Concrete, Localizedfields, Fieldcollection, etc...
```

models/DataObject/ClassDefinition/Data/ManyToManyObjectRelation.php
```diff
@@ -436,7 +437,7 @@ public function resolveDependencies($data)
      *
      * @return array
      */
-    public function preGetData($object, $params = [])
+    public function preGetData(mixed $object, $params = []): mixed
```
dvesh3:
```suggestion
    public function preGetData(/** mixed */ $object, /** array */ $params = []) /**: mixed */
```

models/DataObject/ClassDefinition/Data/ManyToOneRelation.php
```diff
@@ -454,7 +454,7 @@ public function resolveDependencies($data)
      *
      * @return null|Element\ElementInterface
      */
-    public function preGetData($object, $params = [])
+    public function preGetData(mixed $object, $params = []): mixed
```
dvesh3:
```suggestion
    public function preGetData(/** mixed */ $object, /** array */ $params = []) /** : mixed */
```

models/DataObject/ClassDefinition/Data/Wysiwyg.php
```diff
@@ -325,7 +325,7 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
      *
      * @return string
      */
-    public function preGetData($object, $params = [])
+    public function preGetData(mixed $object, $params = []): mixed
```
dvesh3:
```suggestion
    public function preGetData(/** mixed */ $object, /** array */ $params = []) /** : mixed */
```

models/DataObject/ClassDefinition/Data/LayoutDefinitionEnrichmentInterface.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Override point for Enriching the layout definition before the layout is returned to the admin interface.
+     *
+     * @param Concrete|null $object
+     * @param array $context additional contextual data
+     * @return $this
+     *
+     * @throws \Exception
+     */
+    public function enrichLayoutDefinition(/*?Concrete */ $object , /**  array */ $context = []) /* : self */;
```
jdreesen, line 32: Why don't you add parameter types to the new interfaces? It shouldn't be a breaking change. Only adding return types would be.

dvesh3: Please feel free to provide a follow-up PR. thanks! 😊

dvesh3 (approved):

PR #9444, opened 2021-06-14
Upate 07_Properties.md to clarify Select configuration

NiklasBr:
Took some trial-and-error to understand this. :)


weisswurstkanone:
thx!

PR #9447, opened 2021-06-15
[Documents] Static Page Generator

### LINKED ISSUE #8196: [Documents] Static Page Generator
It should be possible to export pages statically onto the public var directory, where the page can then be delivered directly by the web-server (without PHP!). For this the page needs to be saved to a structure that can be resolved by the web-server depending on the request uri. The concept is similar to the one we also use for delivering Assets and Thumbnails. 

#### Examples
Path of document: `/en/Magazine/Lifestyle`
Path of static rendered HTML document: `public/var/tmp/pages/en/Magazine/Lifestyle.html`  

Pretty URL of document: `/Culture`
Path of static rendered HTML document: `public/var/tmp/pages/Culture.html`

# Detailed description

- additional settings (stored in `documents_page` table for static generator
image
- those settings should be also possible to set on document-types 
- maintenance job that (re-)generates the static pages  (business logic should be in a shared service that can be used by the command as well)
- command that generates the static pages incl. the possibility to filter pages e.g. by parent path
- Apache/Nginx rule to serve static pages directly (similar to Assets & thumbnails), in case using a local storage
- public action (in `PublicServicesController` that can serve static pages even if using a remote storage
- #8081 could probably also be helpful

dvesh3:
Resolves #8196

## Additional info  
Tasks for the PR:
- [x] business logic should be in a shared service that can be used by the command as well
- [x] settings should be also possible to set on document-types
- [x] maintenance job that (re-)generates the static pages
- [x] command that generates the static pages incl. the possibility to filter pages e.g. by parent path (resolved in https://github.com/pimcore/pimcore/pull/9468)
- [x] Apache/Nginx rule to serve static pages directly (similar to Assets & thumbnails), in case using a local storage
- [x] Router that delivers static pages stored in a remote storage (enabled by config + optional pattern matching) 
- [x] Migrations to update db tables
- [x] UI Improvements - New page icon(Grey) & Last Generated info on page toolbar
- [x] Documentation

### .htaccess changes (local storage)
```
# static pages
RewriteCond %{REQUEST_METHOD} ^(GET|HEAD)
RewriteCond %{QUERY_STRING}   !(pimcore_editmode=true|pimcore_preview|pimcore_version)
RewriteCond %{DOCUMENT_ROOT}/var/tmp/pages%{REQUEST_URI}.html -f
RewriteRule ^(.*)$ /var/tmp/pages%{REQUEST_URI}.html [PT,L]
```
### Static page router (remote storage)
```yaml
pimcore:
    documents:
        static_page_router:
            enabled: true
            route_pattern: '@^/(en/Magazine|de/Magazin)@'
```

### command
`bin/console pimcore:documents:generate-static-pages`


dpfaffenbauer:
💪 A Command to generate them would also be nice. I could help if you need me to.

dvesh3:
> 💪 A Command to generate them would also be nice. I could help if you need me to.

@dpfaffenbauer yes, would be great 🤩 maybe create a dependent PR (using the service introduced in this PR). thanks in advance 😉

weisswurstkanone:
Question: What will/should happen if you
- disable the generator again
- and then delete a page ?

dvesh3:
> Question: What will/should happen if you
> 
> * disable the generator again
> * and then delete a page ?

@weisswurstkanone In both cases, the corresponding static file should be deleted.

Some more cases:
- Unpublish Document ~ Delete the static file
 - Publish a Version / Autosave ~ Do not touch static file

lib/Document/StaticPageGenerator.php
```diff
@@ -0,0 +1,113 @@
...
+        $storage = Storage::get('document_static');
+        $startTime = microtime(true);
+
+        $lockKey = 'document_static_' . $document->getId() . '_' . md5($storagePath);
+        $lock = $this->lockFactory->createLock($lockKey);
+
+        $lock->acquire(true);
+
+        try {
+            $renderedDocumentData = $this->documentRenderer->render($document);
```
dpfaffenbauer:
```suggestion
            $renderedDocumentData = $this->documentRenderer->render($document, ['static_page_generator' => true]);
```

dpfaffenbauer (approved):
Migration for new DB fields missing, other than that, works like a charm. I'll create a PR for Command Line Generation based on this PR

lib/Document/StaticPageGenerator.php
dpfaffenbauer: I think that would be nice, so can react to static page generation

dvesh3: 👍 thanks!

brusch:
@dvesh3 I think we have an issue here with code-injection event listeners like Google Analytics, it seems they are not getting applied.

dvesh3:
> @dvesh3 I think we have an issue here with code-injection event listeners like Google Analytics, it seems they are not getting applied.

@brusch fixed this issue by changing the approach a bit, so now we generate static page on `kernel.reponse` event after all the code-injection listeners are applied instead of directly generating on `document.save|update` event.

Also, new request context resolver `StaticPageResolver` is added to allow code-injection listeners to be applied for the requests from CLI for static page generation.

models/Document/DocType.php
```diff
@@ -99,6 +99,13 @@ class DocType extends Model\AbstractModel
      */
     protected $modificationDate;
 
+    /**
+     * @internal
+     *
+     * @var null|int
+     */
+    protected $staticGeneratorEnabled;
```
blankse, line 107: Why is this an integer and not an boolean?

brusch: I think that was by mistake, but hard to fix now, without a BC break, I'd say that's a topic for 11 then.

brusch: #10819 made an issue so we don't forget 😊

PR #9454, opened 2021-06-15
[ControllerDataProvider] Fix bundle templates in getTemplates() calls

### LINKED ISSUE #9453: Template path for external bundles is invalid
# Bug Report

### Expected behavior

Template selection drop down should display all templates from configured template paths as follows: @<namespace>/<template_path>/<twig_template>. Otherwise all templates from third party bundles has to be manually edited and the drop down is useless.

### Actual behavior

On documents the template selection drop down shows the wrong path in the settings path. For example the CMF templates are shown as follows:

image

Preview shows the following error because the namespace is not correctly used:

image

dvesh3:
Resolves #9453

### DIFF
--- a/lib/Controller/Config/ControllerDataProvider.php
+++ b/lib/Controller/Config/ControllerDataProvider.php
@@ -17,7 +17,6 @@ declare(strict_types=1);
 
 namespace Pimcore\Controller\Config;
 
-use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\HttpKernel\Bundle\BundleInterface;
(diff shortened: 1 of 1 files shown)


lib/Controller/Config/ControllerDataProvider.php
```diff
@@ -203,29 +200,14 @@ private function findTemplates(string $path, string $bundleName = null): array
             $finder->name($namePattern);
         }
 
+        if ($bundleName && 'Bundle' === substr($bundleName, -6)) {
```
jdreesen:
```suggestion
        if ($bundleName && str_ends_with($bundleName, 'Bundle')) {
```

Issue #6814, opened 2020-07-21
[DataObjects][DataTypes] Clarify and eventually unify enrichFieldDefinition vs. enrichLayoutDefinition


fashxp:
for example see 
https://github.com/pimcore/pimcore/blob/45ef287afc65e199353e976c91b14b059c5c1959/models/DataObject/ClassDefinition/Data/Select.php#L412 vs https://github.com/pimcore/pimcore/blob/45ef287afc65e199353e976c91b14b059c5c1959/models/DataObject/ClassDefinition/Data/Select.php#L436

- would it not be enough to apply options from options provider in just one of these methods? 
- should the calculation of options not be the same - one time inheritance is explicitly activated, one time not...

brusch:
See also: https://github.com/pimcore/frontend-permission-toolkit/blob/2d209ea43fc5ea02817cced8c048b136ddb63169/src/CoreExtensions/ClassDefinitions/DynamicPermissionResource.php#L351

brusch:
Please also consider: #8827

brusch:
#9439 introduced already interfaces for those 2 methods. 

Todos: 
- Describe methods on the interfaces: 
https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/ClassDefinition/Data/FieldDefinitionEnrichmentInterface.php#L27
https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/ClassDefinition/Data/LayoutDefinitionEnrichmentInterface.php#L33
- Check `Select` & `Multiselect` data-types, see issue description above.

PR #9486, opened 2021-06-18
Enrich Layout for CS Attributes in Grid

daves-weblab:
Without enriching the layout definition, batch editing does not work for selects / multiselect with static options. By enriching the layout before sending it to the grid proxy the options are correctly provided and batch editing also works for select / multiselect.


fashxp:
@weisswurstkanone  probably related to #6814 ?

weisswurstkanone:
@fashxp , topic wise related

brusch:
@daves-weblab could you please rebase onto `6.9`, this should fix the tests ( see #9511 ) Thanks! 
P.S.: would be urget if this PR should make it into `6.9.6` (before end of life) which is going to be released tomorrow morning.

PR #9495, opened 2021-06-18
[Geographical Point] Correct order of latitude and longitude in version preview

BlackbitDevs:
Currently it is displayed as longitude, latitude. The usual order is latitude, longitude - as for example used in CSV exports:
https://github.com/pimcore/pimcore/blob/d1b4d14b599950f892a6fbdcfdccc215081eb4d6/models/DataObject/ClassDefinition/Data/Geopoint.php#L206

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Geopoint.php
+++ b/models/DataObject/ClassDefinition/Data/Geopoint.php
@@ -190,7 +190,7 @@ class Geopoint extends AbstractGeo implements
     public function getVersionPreview($data, $object = null, $params = [])
     {
         if ($data instanceof DataObject\Data\GeoCoordinates) {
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Hello @BlackbitNeueMedien , just a suggestion. Maybe rebase this on 10.0 ?

BlackbitDevs:
Hmm, I do not have a problem with this. I thought that everything which is in branch 10.x will be released in the next release anyway. Did I understand something wrong?

BlackbitDevs:
Ah I see, bugfixes should go in 10.0, new features in 10.x?

weisswurstkanone:
Exactly. Thanks a lot!

Issue #5010, opened 2019-09-24
[DataObject] MultiSelection value with commas not working


ramundomario:
# Bug Report

### Expected behavior
Insert a value with commas in Selection Options of Multiselection type into DataObject class definition.
After save an object and reload, retrieve correct value.

### Actual behavior
Currently after an object is saved & refreshed, the value previously selected (with commas) is not shown

### Steps to reproduce

1. Insert a value with commas in Selection Options of Multiselection type into DataObject class definition.
 <img width="1402" alt="class_definition" src="https://user-images.githubusercontent.com/16133208/65531891-2e82ea80-defb-11e9-934b-c9e8681e73aa.png">
2. Select this value with commas into object creation
<img width="1408" alt="object_creation" src="https://user-images.githubusercontent.com/16133208/65532012-5c682f00-defb-11e9-898a-a0eef19903cd.png">
3. Reload object without correct value selected
<img width="1404" alt="object_reload" src="https://user-images.githubusercontent.com/16133208/65532058-730e8600-defb-11e9-8105-79d12106b041.png">

This is the value saved into the database:
<img width="1404" alt="database_value" src="https://user-images.githubusercontent.com/16133208/65532088-7efa4800-defb-11e9-88a6-970a02064b67.png">

Since the values are saved with commas in the multiselect data type into database, it seems that the values are saved as two distinct elements.

HeikoNF:
This issue is more than a year old and it ist still not possible to use commas neither as key nor as value in multiselects. Is there any known workaround or plan to changed this in the near future?

podarcis:
We just got hit by this bug/behavior, too.

Imagine having values for a field `scale` with possible values `1:2` and `1:2,5`. The field value is stored as `1:2,1:2,5` and thus values `1:2`, `1:2` and `5` are read back.

MadeP:
Strangely enough, this was not a problem in PIMCORE 6. In PIMCORE X it suddenly doesn't work anymore.

podarcis:
@MadeP Please elaborate on what's not working anymore. Previously commas were accepted but led to silent data loss. Now there's a sanity check as a workaround in place, which just prevents commas.

MadeP:
Ok, the silent loss could not be determined in the way so far. Have checked and it is true. Then the workaround that no commas are allowed is probably best for now. Would still be good if commas work in the future though.

fashxp:
[...]

PR #8972, opened 2021-04-30
[Data Objects] introduce onContainerSaved method 

weisswurstkanone:
resolves #5965

PR #9501, opened 2021-06-19
[DataObjects] Deny option values with comma for multiselect field type

podarcis:
This PR performs a sanity check on multiselect option values (both static and ones provided by a provider implementing `MultiSelectOptionsProviderInterface`) checking for a value containing commas, which are not supported for now and lead to silent data corruption due to the way the value is persisted (comma-separated), see #5010.

While it would be nice to see multiselect option values actually supporting commas, it doesn't seem to be a huge pain point. A correct fix could involve some kind of value escaping, but might consider BC / create proper migrations as well.

This PR instead prevents using commas in the first place, and probably saving a project from data loss/corruption. Please note that there is some kind of BC break as well, as already existing faulty definitions will throw an exception on class save - IMO in that rare case it's a wanted behavior telling the developer, that data loss may have already been happened in the past.

Addresses #5010

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Multiselect.php
+++ b/models/DataObject/ClassDefinition/Data/Multiselect.php
@@ -31,7 +31,8 @@ class Multiselect extends Data implements
     \JsonSerializable,
     NormalizerInterface,
     LayoutDefinitionEnrichmentInterface,
-    FieldDefinitionEnrichmentInterface
+    FieldDefinitionEnrichmentInterface,
+    DataContainerAwareInterface
 {
     use DataObject\Traits\SimpleComparisonTrait;
     use Extension\ColumnType;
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/Multiselect.php
```diff
@@ -645,6 +645,35 @@ public function resolveBlockedVars(): array
...
+    {
+        $optionsProvider = DataObject\ClassDefinition\Helper\OptionsProviderResolver::resolveProvider(
+            $this->getOptionsProviderClass(),
+            DataObject\ClassDefinition\Helper\OptionsProviderResolver::MODE_MULTISELECT
+        );
+        if ($optionsProvider) {
+            $context = [];
+            $context['fieldname'] = $this->getName();
+
+            $options = $optionsProvider->{'getOptions'}($context, $this);
```
jdreesen:
```suggestion
            $options = $optionsProvider->getOptions($context, $this);
```

models/DataObject/ClassDefinition/Data/Multiselect.php
```diff
@@ -645,6 +645,35 @@ public function resolveBlockedVars(): array
...
+        } else {
+            $options = $this->getOptions();
+        }
+        if (is_array($options) && array_reduce($options, static function ($containsComma, $option) {
+            return $containsComma || (strpos($option['value'], ',') !== false);
```
jdreesen: Since Pimcore 10 uses PHP 8, we should use [str_contains](https://www.php.net/str_contains):

```suggestion
            return $containsComma || str_contains($option['value'], ',');
```

models/DataObject/ClassDefinition/Data/Multiselect.php
podarcis: Agree, I copied that part 1:1 some from some lines above. Not sure why it's there in the first place, as there aren't any special characters in getter.

dvesh3:
Please merge 10.0 branch into current to fix the tests. thanks!

podarcis:
In further tests I realized that `classSaved()` is really only called from `DataObject\ClassDefinition::save()`, but not from another container like a FieldCollection. So this sanity check seems to only work if the field is directly added to a class definition. On container types it won't work. I don't see an easy fix or the best practice to perform such a check.

weisswurstkanone:
@podarcis , maybe just as a comment ... this one will be introduced in 10.1
https://github.com/pimcore/pimcore/issues/5965

brusch:
I'd see that as an improvement, so this should go into `10.x` (please rebase this PR). 
I'd also add a validation to the UI, to give instant feedback. 
Last but not least, since this PR is now going into `10.x` we can already use `onContainerSaved` as described in #8972

@podarcis could you please update your PR accordingly? Thanks in advance 😊

brusch (changes requested):
See above.

podarcis:
@brusch Unfortunately I cannot adjust my PR in a timely manner. Either it has to wait, or - and I won't be mad ;-) - someone else takes over this fix.

brusch:
@podarcis no worries, it's not urgent anyway. 
Maybe we'll ping you from time to time, if you don't mind 😊

brusch:
Works like a charm, thanks a lot! 👍

PR #9502, opened 2021-06-19
[Docs] Add Notice for getThumbnail-Function

matthiashamacher:
Add Information to the `getThumbnail()`-Function that the Thumbnail is not generated during the call.

### DIFF
--- a/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
+++ b/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
@@ -27,6 +27,9 @@ To retrieve a thumbnail from an asses simply call `$asset->getThumbnail("thumbna
 an `\Pimcore\Model\Asset\Image\Thumbnail` object. The thumbnail object's `__toString()` method returns the path to the thumbnail file, for example: 
 `/Car%20Images/ac%20cars/image-thumb__68__content/automotive-car-classic-149813.jpg`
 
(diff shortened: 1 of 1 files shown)


doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
```diff
@@ -27,6 +27,9 @@ To retrieve a thumbnail from an asses simply call `$asset->getThumbnail("thumbna
 an `\Pimcore\Model\Asset\Image\Thumbnail` object. The thumbnail object's `__toString()` method returns the path to the thumbnail file, for example: 
 `/Car%20Images/ac%20cars/image-thumb__68__content/automotive-car-classic-149813.jpg`
 
+**Important**: The function `getThumbnail()` does not generate the Thumbnail itself. It just returns the path were the thumbnail file will be stored.
```
dvesh3, line 30: this is already explained here: https://github.com/pimcore/pimcore/blob/10.x/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md#deferred-rendering-of-thumbnails

matthiashamacher: Yes, that's true. But at this point of the documentation, it is not clear that that is the case. So to make that clear and give some context I've added that information. I also linked to the other section so that you can get more information about that there.

brusch:
Can't hurt to have this mentioned here as well  😉

PR #9508, opened 2021-06-21
Add to conflicts "google/apiclient-services": ">=0.200.0" to solve API class not found error

dvesh3:
Resolves #

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -106,7 +106,8 @@
     "symfony/symfony": "3.4.43 || 4.4.11",
     "webmozarts/console-parallelization": "1.2.2",
     "symfony/monolog-bundle": "3.6.0",
-    "laminas/laminas-code": ">=4.4"
+    "laminas/laminas-code": ">=4.4",
+    "google/apiclient-services": ">=0.200.0"
   },
   "require-dev": {
     "cache/integration-tests": "^0.16.0",


lib/Google/Cse.php
```diff
@@ -138,9 +139,9 @@ public function load()
     public $query = '';
 
     /**
-     * @var array
+     * @var CustomSearchAPI\Search
```
blankse:
```suggestion
     * @var CustomSearchAPI\Search|null
```

lib/Google/Cse.php
```diff
@@ -262,7 +263,7 @@ public function setRaw($raw)
     }
 
     /**
-     * @return array
+     * @return CustomSearchAPI\Search
```
blankse:
```suggestion
     * @return CustomSearchAPI\Search|null
```

lib/Google/Cse/Item.php
```diff
@@ -15,12 +15,13 @@
...
 namespace Pimcore\Google\Cse;
 
+use Google\Service\CustomSearchAPI\Result;
 use Pimcore\Model;
 
 class Item
 {
     /**
-     * @var \Google_Service_Customsearch_Result
+     * @var Result
```
blankse, line 24:
```suggestion
     * @var Result|null
```

blankse:
@dvesh3 The static analysis Test in 6.9 doen't test with lowest dependencies. I think we must add the namespace version to composer.json

dvesh3:
@blankse old class names should be still supported with the installed version. please see https://github.com/googleapis/google-api-php-client-services/issues/383

blankse:
@dvesh3 But you use now the new class names. So Pimcore require "google/apiclient-services": "^0.201.0". When you execute composer with --prefer-lowest you will get error in phpstan I think.

dvesh3:
> @dvesh3 But you use now the new class names. So Pimcore require "google/apiclient-services": "^0.201.0". When you execute composer with --prefer-lowest you will get error in phpstan I think.

yes, i agree with you! but we decided not to go with this change at first place, as it will still break custom implementations. so our first preference is that it should be fixed on third-party repo otherwise we need to provide some BC layer along with this change.

blankse:
@dvesh3 So you can set "google/apiclient-services": ">=0.200.0" in conflict and use the namespaces only in 10.x

PR #9510, opened 2021-06-21
Fix default value for document editable select

aweichler:
With this fix it is possible to define a default / initial value for a select editable in documents.


dvesh3:
Please merge 10.0 branch into current to fix the tests. thanks!

brusch:
@aweichler what is the actual issue, can you provide us an example? 
For me the documented way how to use default values with the select editable, works just fine: 

```twig
{{ pimcore_select("valid_for", {
	"store": [
		["one-month", "One month"],
		["three-months", "Three months"],
		["unlimited", "Unlimited"]
	],
	"defaultValue" : "unlimited"
}) }}
```

aweichler:
You are right, "defaultValue" works fine. I am sure i have tried with that configuration before, but maybe i have done an mistake.
Sorry for that.

PR #9498, opened 2021-06-18
Quicksearch: use class icons when available

NiklasBr:
## Changes in this pull request  
Show custom class icon in quicksearch for root products (variants still keep their icon)

## Additional info  

image

PR #9512, opened 2021-06-21
[Application logger] Make text selectable / Link related object

BlackbitDevs:
With this PR you can select and copy the text in the log grid. 
Additionally you can directly jump to the related object by clicking the link in the  `Related object` column.
Bildschirmfoto 2021-06-21 um 13 29 22


BlackbitDevs:
Failing tests are imho not caused by this PR.

NiklasBr:
I think #9508 will resolve them

weisswurstkanone:
maybe related https://github.com/pimcore/pimcore/pull/9498

brusch:
@BlackbitNeueMedien can you please rebase your PR in order to fix the tests? Thanks!

weisswurstkanone:
Hello  @BlackbitNeueMedien , would it make sense to beatify it with the open icon? 
image
What do you think ?

BlackbitDevs:
Imho there is not much space anyway because the most important thing is the error message. And as the link already shows that something will happen when you click it, I do not know if it is necessary. But would also be not much work. If you want it, just tell, then I will include it in the renderer function.

weisswurstkanone:
@BlackbitNeueMedien thanks a lot for your contribution!

PR #9523, opened 2021-06-22
Added code to remove root document path

Daniel-Ateles:
The redirect list shows incorrect target path with the root path included if it is an auto redirect (with id instead of path).
This removes the root path, like the RedirectHandler.

  
  

Resolves #


brusch:
I'm not really sure if this is the right and expected behavior. 
In the Pimcore admin UI, you only work with the fully qualified path of a document, no matter if it's used within a site-context or not. This change would be not consistent anymore and maybe causes confusion.

brusch:
I'd also not classify this as a bug even if we come to the conclusion that it makes sense 😊

brusch:
Closed, see https://github.com/pimcore/pimcore/pull/9418#issuecomment-866003918

PR #9524, opened 2021-06-22
Fix inherited relation values not being visible after path formatting

### LINKED ISSUE #9509: Path formatter breaks on inheritance
# Bug Report

### Expected behavior

This is what the root object looks like:
image

Inherited values, like their parent values, should be displayed when using [Path Formatter](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Class_Settings/Path_Formatter.html), but are not.

### Actual behavior

The content becomes blank, like the inherited values here:

image

### Steps to reproduce

See code example on a `Product` class with variants.

### Code example
```php
<?php

namespace AppBundle\Service;

use AppBundle\Model\DataObject\Product;
use Pimcore\Model\DataObject\ClassDefinition\PathFormatterInterface;

class PathFormatterService implements PathFormatterInterface
{
    {
        foreach ($targets as $key => $item) {
            if ($item["type"] === "object") {
                $targetObject = Concrete::getById($item["id"]);
                if ($targetObject instanceof Product) {
                    $result[$key] = $this->getPath($targetObject, 'getOSName');
                }
            }
        }

        return $result;
    }

    protected function getPath(Concrete $object, string $getter): string
    {
        $result = [];

        while ($object instanceof Concrete) {
            if ($value = $object->{$getter}()) {
                $result[] = $value;
            }

            $object = $object->getNextParentForInheritance();
        }

        return implode(' &rarr; ', array_reverse($result));
    }
}
```

NiklasBr:
Resolves #9509

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/ElementController.php
+++ b/bundles/AdminBundle/Controller/Admin/ElementController.php
@@ -562,15 +562,17 @@ class ElementController extends AdminController
             if ($ownerType == 'object' && method_exists($source, $methodName)) {
                 $data = $source->$methodName();
                 $editModeData = $fd->getDataForEditmode($data, $source);
-                if (is_array($editModeData)) {
+                // Inherited values show as an empty array
+                if (is_array($editModeData) && !empty($editModeData)) {
                     foreach ($editModeData as $relationObjectAttribute) {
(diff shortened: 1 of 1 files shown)


brusch:
Thank you very much!

Issue #5107, opened 2019-10-17
[Feature request] Remove outdated password requirements


NiklasBr:
# Feature Request
You've all [read this XKCD comic](https://xkcd.com/936/) by now, haven't you? `correctbatteryhorsestaple` and all that....

It is no longer recommended to use complex rules for user passwords, for example these which recommend against the current implementation:

- [Microsoft password guidance, anti-pattern #2](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/06/Microsoft_Password_Guidance-1.pdf)
- [NIST on complexity composition rules](https://pages.nist.gov/800-63-3/sp800-63b.html)
- [GCHQ guidance](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/458857/Password_guidance_-_simplifying_your_approach.pdf)

In summary: Just a minimum character length please.

brusch:
+ do server-side validation as well

ondrejd:
I'm looking into sources of 6.4.2 and doesn't see these changes.

PR #9530, opened 2021-06-23
Fix wrong Password hint

blankse:
Hint is wrong since https://github.com/pimcore/pimcore/issues/5107


brusch:
Thank you very much!

blankse:
@brusch Translation changes from https://poeditor.com/projects/ only get in Feature Releases?

brusch:
@blankse at the moment yes. It's on our todo list to automate the procedure via Actions, but that's not as easy as you'd think it is 😉

PR #9545, opened 2021-06-23
Fix issue with missing indexName in ecommerce indexer configuration.

mugge6:
Fix issue with missing indexName in ecommerce indexer configuration. If indexName ist not set, then the tenant name will be used. But currently a missing arrey index exception is thrown.


fashxp:
LGTM .. thx

PR #9549, opened 2021-06-23
fix(kernel) allow tests to finish with cleared container

### LINKED ISSUE #9531: Kernel getContainer() throws a LogicException on shutdown
# Bug Report

### Expected behavior
Successfull shutdown.

### Actual behavior
```sh
Fatal error: Uncaught LogicException: Cannot retrieve the container from a non-booted kernel. in /app/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:307
Stack trace:
#0 /app/vendor/pimcore/pimcore/lib/Kernel.php(238): Symfony\Component\HttpKernel\Kernel->getContainer()
#1 [internal function]: Pimcore\Kernel->Pimcore\{closure}()
#2 {main}
  thrown in /app/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php on line 307
```

### Steps to reproduce
Running tests which boot multiple containers.
You use parent getContainer() and then check the value
https://github.com/pimcore/pimcore/blob/e9f1d5b03614a5d13e6a2994c25faa22d5f80076/lib/Kernel.php#L238
but the parent throws a LogicException earlier

https://github.com/symfony/symfony/blob/1a2e10137e03ad29c1f4f6fb1ab95ecd02be4518/src/Symfony/Component/HttpKernel/Kernel.php#L307

# Feature Request

Use try-finally to get the container value

ubaranouski:
Resolves [#9531](https://github.com/pimcore/pimcore/issues/9531)

## Additional info  
Allow tests to finish with cleared container.

### DIFF
--- a/lib/Kernel.php
+++ b/lib/Kernel.php
@@ -43,6 +43,7 @@ use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\Config\Resource\FileExistenceResource;
 use Symfony\Component\Config\Resource\FileResource;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 use Symfony\Component\EventDispatcher\GenericEvent;
 use Symfony\Component\HttpKernel\Bundle\BundleInterface;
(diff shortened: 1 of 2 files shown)


lib/Kernel.php
```diff
@@ -236,10 +237,14 @@ protected function initializeContainer()
...
             try {
-                $this->getContainer()->get('event_dispatcher')->dispatch(new GenericEvent(), SystemEvents::SHUTDOWN);
-            } finally {
-                \Pimcore::shutdown();
+                $container = $this->getContainer();
+            } catch (\LogicException) {
+                // Container is cleared. Allow tests to finish.
             }
+            if (isset($container) && $container instanceof ContainerInterface) {
+                $container->get('event_dispatcher')->dispatch(new GenericEvent(), SystemEvents::SHUTDOWN);
```
brusch, line 245: Shouldn't we move that inside the `try` block?

brusch (changes requested):

lib/Kernel.php
ubaranouski: I wouldn't do that because you'll lose exceptions from container and dispatcher. We shouldn't catch them.

brusch:
Since this is a bug-fix we should merge it into the `10.0` branch, can you please rebase your PR and change the target branch accordingly? Thanks again!

ubaranouski:
Rebased.

dvesh3:
@ubaranouski i tried to change the target base and still see lot of unrelated commits. It doesn't seem like rebased correctly. thanks!

brusch:
Thank you very much!

brusch:
It's fine in `10.x` as well, I'm going to cherry-pick into `10.0`.

PR #9551, opened 2021-06-23
update detailAction (timeline)

zman-dwa:
Changed three lines to use the null coalescing operator (??) - the ternary operator (?) throws an error if the key is not found (ie., when the title is different that the options provided). Could use isset() instead.

To reproduce error, add a note with a title other than those defined in $arrIcons and $arrContext

  
  

Resolves #

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Controller/AdminOrderController.php
+++ b/bundles/EcommerceFrameworkBundle/Controller/AdminOrderController.php
@@ -348,13 +348,13 @@ class AdminOrderController extends AdminController implements KernelControllerEv
 
             // add
             $arrTimeline[$group][] = [
-                'icon' => $arrIcons[$note->getTitle()],
-                'context' => $arrContext[$note->getTitle()] ?: 'default',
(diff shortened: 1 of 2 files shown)


dvesh3:
Hi @zman-dwa thanks for your PR! you need to remove ignore error rule from phpstan baseline https://github.com/pimcore/pimcore/blob/10.x/phpstan-baseline.neon#L39-L41 to fix the tests.

and also rebase this PR on branch 10.0, since this is a fix.

zman-dwa:
@dvesh3 I have rebased this PR and removed the ignore error rule. (I am new to this environment and hope I did these two steps correctly).

jdreesen:
You have to skip all commits that don't belong to your branch when rebasing.

dvesh3:
@zman-dwa please see https://github.com/pimcore/pimcore/pull/9551#issuecomment-869874266

brusch:
Fixed by 521d78e5cd8e641d32393bcccda073690404ae8c

PR #9553, opened 2021-06-24
[Mail] First log to "Sent mails" then try to send

BlackbitDevs:
This PR changes the order of processing when emails are sent via `\Pimcore\Mail`. The problem currently is that when you call the `send()` method that it is first tries to send the mail and only if this was successful the mail is logged / saved to "Sent mails" in Pimcore backend. If sending the mail was not successful an exception is thrown and thus the mail does not get logged to "Sent Emails":
https://github.com/pimcore/pimcore/blob/d37fd30fa0768adeebd5161e8c8df07e0df3e88b/lib/Mail.php#L548-L573

The problem is when you recognize in your application code that no mail has been sent you do not know if your application is wrong or if the email configuration is wrong. With this PR the mail to be sent is first logged to "Sent mails". This way even when mail configuration is wrong and the mail cannot be sent, you see that at least it was tried to be sent in the "Sent mails".


fashxp:
hmm, but then its not a sent mails log anymore... ;-)
I guess, then we also would need to add a flag that indicates if sending was possible or not...

BlackbitDevs:
You are right, @fashxp - I implemented a more sophisticated solution. Now there is a new icon in the "Sent Mails" panel. When you click it you see the error which happened during sending. If errors occured the whole line in the grid gets red background:
Bildschirmfoto 2021-06-24 um 09 38 42

Bildschirmfoto 2021-06-24 um 09 38 48

models/Tool/Email/Log.php
```diff
@@ -138,6 +138,13 @@ class Log extends Model\AbstractModel
      */
     protected $subject;
 
+    /**
+     * Error which happened when mail got sent - empty if no error occured
```
dvesh3:
```suggestion
     * Error log, when mail send resulted in failure - empty if successfully sent
```

dvesh3 (changes requested):
I assume "forward" and "resend" options are disabled for the failed logs, since this will lead to same the error.

BlackbitDevs:
Especially for mails with errors I would not disable "resend" because when the error is about a wrong SMTP configuration or something else with the mail server, this "resend" allows to send the mails again after the infrastructure problem has been fixed.

dvesh3:
> Especially for mails with errors I would not disable "resend" because when the error is about a wrnig SMTP configuration or something else with the mail server, this "resend" allows to send the mails again after the infrastrcuture problem has been fixed.

yes, that's a valid point. 👍

dvesh3 (approved):

dvesh3:
@BlackbitNeueMedien for  "wrong SMTP configuration" exception handling we need to refactor the mail class. As it will throw exception when loading mailer from the container, so here https://github.com/pimcore/pimcore/blob/10.x/lib/Mail.php#L534 (which is not in try catch) and moreover, we are catching only `TransportExceptionInterface` related to send() not `InvalidArgumentException`.

BlackbitDevs:
Ok, now all exceptions get caught.

dvesh3:
@BlackbitNeueMedien Nice improvement 👍  thanks a lot!

PR #9554, opened 2021-06-24
[Data Object] WYSIWYG: char counter and maxChar limit in datatype

### LINKED ISSUE #3720: Add Character count to WYSIWYG
# Feature Request 
## Feature description 
With the tickets #3717 and #3714, a character count was added to Textarea and Input elements. It would be great to have the same feature for WYSIWYG fields as well. 

This would include: 
- A character count info on the WYSIWYG Editor
- A setting in Pimcore DataObject how many characters are allowed for this input

There are plugins for CKEditor to realize this feature, e.g. this one: http://w8tcha.github.io/CKEditor-WordCount-Plugin/ 

My recommendation would be to include the plugin with the configuration
```
config.extraPlugins = 'wordcount';
config.wordcount = {
    // Whether or not you want to show the Word Count
    showWordCount: false,

    // Whether or not you want to show the Paragraphs Count
    showParagraphs: true,

    // Whether or not you want to show the Char Count
    showCharCount: true,

    // Whether or not you want to count Spaces as Chars
    countSpacesAsChars: true,

    // Maximum allowed Char Count
    maxCharCount: ##PIMCORE_VALUE##
};
```

Does anyone else need this feature?

blazevici:
Resolves #3720

## Additional info  
I've added the wordcount (and notification) plugin to CKEditor for this feature which allows us to have constant character counter in WYSIWYG and also going into the WYSIWYG DataType we can now limit the number of characters (picture below). 

image

It's important to state that if a object already has WYSIWYG content going over the char limit it won't shorten it, instead there is a red notification that you've went over the limit. 

Here's a picture of the counter in action:

image


brusch:
@blazevici Thanks a lot for your PR! 
So far everything looks good to me, but there's just one small thing that's missing: 
Please also add this plugins to the build config of CKEditor, otherwise they won't be included anymore as soon as we're updating CKEditor to the latest version. 
See: https://github.com/pimcore/pimcore/blob/3ad43b47041676e8c276bc32bd47926e7c01e735/bundles/AdminBundle/Resources/public/js/lib/ckeditor/build-config.js#L97

Thanks a lot!

bundles/AdminBundle/Resources/public/js/lib/ckeditor/config.js
```diff
@@ -7,4 +7,5 @@ CKEDITOR.editorConfig = function( config ) {
 	// Define changes to default configuration here. For example:
 	// config.language = 'fr';
 	// config.uiColor = '#AADC6E';
+    config.extraPlugins = 'wordcount,notification';
```
brusch: Maybe It's anyway a good idea to just use the build config and then build CKEDITOR again (see https://github.com/pimcore/pimcore/blob/697e81cc40f4c879dd638d4656d7926bb80b8fa5/bundles/AdminBundle/Resources/public/js/lib/PIMCORE_MODIFICATIONS.md#L1) to include the additional plugins. 
Then this line shouldn't be necessary, because they are loaded in `ckeditor.js`.

brusch (changes requested):

bundles/AdminBundle/Resources/public/js/lib/ckeditor/config.js
blazevici: Hey @brusch, I've rebuilt CKEditor as you asked me to do. My flow was uploading the existing `build-config.js` file, then adding the wordcount plugin to the builder and downloading the optimized version which I then added to my branch instead of the existing ckeditor files. I've tested it locally beforehand and it worked fine. Pushed the changes to this branch.

brusch:
@blazevici Thanks! 
Just one more thing, you have now included the `/samples` folder, which unnecessarily bloats the repo and also contains potential security issues (e.g. `/ckeditor/samples/old/assets/posteddata.php` ). 
To fix it, simply delete the entire `/samples` folder. 
Thanks again!

blazevici:
Hey @brusch, sorry for leaving the /samples folder, I was thinking of removing it firsthand but left it instead just in case. I've removed it from the CKEditor folder and pushed the changes here. Tell me if there's anything else left to do.

brusch:
Now we are there, thanks a lot! 👍😊

PR #9570, opened 2021-06-25
Fix video asset image thumbnail parameter

ghost:
Use `setimage` instead of `settime` for video thumbnail type `image_thumbnail_asset`
(Route `pimcore_admin_asset_getvideothumbnail`)

Without this change it is not possible to set an image as thumbnail for a video.


dvesh3:
@siwamhader Pimcore 6 reached EOL. please rebase on 10.0 branch. thanks!

brusch:
@siwamhader target branch change was possible without rebasing, so nothing else to do for you 😊

PR #9588, opened 2021-06-28
[XLIFF] The export removes zeros (#9586)

### LINKED ISSUE #9586: [XLIFF] The export removes zeros
# Bug Report
When I translate an object, zeros that are alone in a HTML Element are deleted.

## Expected behavior
The XLIFF file should always output the content of a field and never remove content from it.

## Actual behavior
Currently, zeros that are alone in an HTML element are removed.

image
image

## Steps to reproduce
1. open an object with a WYSWIG field
2. add a 0 in a paragraph, without any other text
3. add a 0 in another paragraph, with some text.
4. Create a table
5. add a 0 in a single cell
6. add a 0 and random text to another cell
7. save the object
8. open XLIFF Export in the menu
9. drag the saved object into it and export it
10. open the xliff file
=> All zeros without text have been removed

BabovicT:
Export correctly exporting zeros now.

Resolves #9586

### DIFF
--- a/lib/Translation/Escaper/Xliff12Escaper.php
+++ b/lib/Translation/Escaper/Xliff12Escaper.php
@@ -52,7 +52,7 @@ class Xliff12Escaper
             $parts = explode('>', $match);
             $parts[0] .= '>';
             foreach ($parts as $part) {
-                if (!empty(trim($part))) {
+                if (!empty(trim($part)) || $part === '0') {
                     if (preg_match("/<([a-z0-9\/]+)/", $part, $tag)) {
                         $tagName = str_replace('/', '', $tag[1]);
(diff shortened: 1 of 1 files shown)


lib/Translation/Escaper/Xliff12Escaper.php
```diff
@@ -52,7 +52,7 @@ public function escapeXliff(string $content): string
             $parts = explode('>', $match);
             $parts[0] .= '>';
             foreach ($parts as $part) {
-                if (!empty(trim($part))) {
+                if (!empty(trim($part)) || $part === '0') {
```
dvesh3:
```suggestion
                if (!empty(trim($part)) || trim($part) === '0') {
```

PR #9603, opened 2021-06-29
[Workflows] Fixes error response

fashxp:
fixes error responses for workflow transition requests


bundles/AdminBundle/Controller/Admin/WorkflowController.php
```diff
@@ -132,14 +132,14 @@ public function submitWorkflowTransitionAction(Request $request, Registry $workf
...
-                    'reason' => $reason,
+                    'reasons' => [$reason],
 
                 ];
             } catch (\Exception $e) {
                 $data = [
                     'success' => false,
                     'message' => 'error performing action on this element',
-                    'reason' => $e->getMessage(),
+                    'reasons' => [$e->getMessage()],
```
dvesh3, line 142: we need to fix it for `submitGlobalAction` as well, at line 192 & 199

PR #9612, opened 2021-06-29
[Security] Add Password Hasher Factory & deprecate Password Encoder Factory

dvesh3:

Starting with Symfony 5.3, Password encoder factory has been deprecated and new Password Hasher factory was introduced as a replacement. We need to provide the new decorator factory for  Password Hasher factory, so as to align with Symfony authentication system. 

So, bumped `security-core` & `security-bundle` requirement to `^5.3.0` and deprecated Pimcore Password encoder factory with the fact that, it is still default factory used on Pimcore.

If you wish to move to new Passsword hasher Factory, then change the default factory type to `password_hasher` as follows:

```yaml
pimcore:
    security:
        factory_type: password_hasher
```

## Additional info  
 - Revert changes from https://github.com/pimcore/pimcore/pull/9611 along with https://github.com/pimcore/pimcore/blob/1b434a2f4e1f5f4014afe554ec598e47b1f40012/bundles/CoreBundle/Resources/config/services.yml#L41

Resolves #9611


dpfaffenbauer:
💪

dvesh3:
>  - Revert changes from https://github.com/pimcore/pimcore/pull/9611 along with https://github.com/pimcore/pimcore/blob/1b434a2f4e1f5f4014afe554ec598e47b1f40012/bundles/CoreBundle/Resources/config/services.yml#L41

Done!

Issue #9594, opened 2021-06-28
ClassificationStore on Grid view - sorting/filtering is not working


mgomse:
There is a reproducible issue on the grid view with Classification Store - both sorting and filter are not working with CS.

You can find my reproduction from https://demo.pimcore.com/admin below video:
https://user-images.githubusercontent.com/16497506/123675540-22039280-d843-11eb-9ffc-4abff4c64a65.mp4

We've found it at the start on Pimcore 6.9.5, checked that on 6.9.6 it is not fixed and then I checked on Pimcore X (demo.pimcore.com) and the problem is still there.

mgomse:
Good morning - is there any chance for checking this issue? :) Seems quite important in my opinion.

brusch:
Fixed by #9624

PR #9620, opened 2021-06-30
Fixes issue (#9594)

dturopoli:
Fixes issue https://github.com/pimcore/pimcore/issues/9594


dvesh3:
@dturopoli thanks a lot for your PR! Unfortunately, this only fixes the sort issue. the issue is completely fixed in https://github.com/pimcore/pimcore/pull/9624

Issue #9600, opened 2021-06-29
Doctrine migrations ignores --prefix for migration generation


HaaseIT:
# Bug Report

I added this config to my config/config.yaml

```
doctrine_migrations:
    migrations_paths:
        'App\Migrations': '%kernel.project_dir%/src'
```

Side-Note, the docs are wrong here, they say config.yml and also the directory is incorrect, it says:

```
doctrine_migrations:
    migrations_paths:
        'App\Migrations': '%kernel.project_dir%/src/App'
```

I then ran:
```php bin/console doctrine:migrations:generate --prefix=App\\Migrations```

This was the output, file generation is as the output states:

```
 Generated new migration class to "/var/www/html/vendor/pimcore/pimcore/bundles/CoreBundle/Migrations/Version20210629080600.php"                                                                                   
                                                                                                                                                                                                                   
 To run just this migration for testing purposes, you can use migrations:execute --up 'Pimcore\Bundle\CoreBundle\Migrations\Version20210629080600'                                                                 
                                                                                                                                                                                                                   
 To revert the migration you can use migrations:execute --down 'Pimcore\Bundle\CoreBundle\Migrations\Version20210629080600'                                                                                        
```

### Expected behavior

Generates the migration file in src/Migrations

### Actual behavior

Generates the migration file in vendor/pimcore/pimcore/bundles/CoreBundle/Migrations

### Steps to reproduce

See above

**Comment by dpfaffenbauer** [comment]
Had the same issue, must be like this:

```
doctrine_migrations:
    migrations_paths:
        'App\Migrations': '%kernel.project_dir%/src/App/Migrations'
```

HaaseIT:
ok, I'll test this. If it works like this on my environment, I'll make a quick pullrequest for the documentation.

HaaseIT:
@dpfaffenbauer nope, doesnt work.

![grafik](https://user-images.githubusercontent.com/7584444/123779555-96394700-d8d2-11eb-84fc-88569b2bc49a.png)

cipribucur:
The `--prefix` is meant to filter the results of the read commands like `doctrine:migrations:list`, for the generation you need to use the `namespace` argument `doctrine:migrations:generate --namespace=App\\Migrations` that should do the trick.

dvesh3:
Fixed by #9623

PR #9623, opened 2021-06-30
added sample command for generating project specific migration

HaaseIT:
This is a PullRequest to clarify #9600


cipribucur:
@HaaseIT Thanks for the PR.

PR #9625, opened 2021-06-30
Increase colour contrast on locked elements in tree

NiklasBr:

I have alwayt thought the contrast was a bit low. Is this ok?

Lock colour changes (updated on top):
image

Inherited Lock colour changes (updated on top):
image


brusch:
Yes, totally makes sense - thanks a lot! 😊

PR #9626, opened 2021-06-30
ImageOptimizer: Remove unused mime type guesser

blankse:
The class property is unuesd and the classes are final.


brusch:
LGTM, thanks a lot! 👍

PR #9627, opened 2021-06-30
[Ecommerce][OrderManager] Set localized pricing rule label

jremmurd:
With this PR pricing rule labels in order items are set in the correct locale.


jdreesen:
Should probably go into the `10.0` branch, as it's a bugfix.

dvesh3:
Fixed directly on 10.0 branch https://github.com/pimcore/pimcore/commit/7dfec29ae2fb0c177ee38b5fc674940122a1ed3d

PR #9634, opened 2021-06-30
Allow Concrete::__callStatic() to retrieve variants in getters

### LINKED ISSUE #9457: Concrete::__callStatic does not allow variants
# Bug Report

Calling a static getter code like `Product::getBySomefield('valueInVaraint')` should return any `Product` regardless of whether or not is is an object or variant. Or the getter must accept a parameter that allows us to toggle this behaviour like for regular DataObject listings.

### Steps to reproduce

1. Add an object
2. Fill it with some data
3. Add a variant of that object
4. Change some field in the varaint
5. Attempt to fetch the variant (or object) via a static getter.

### Code example
```php
// A MyClass object which is a variant will never be returned here even thought it exists in the database.
Product::getBySomeField('data', 1);
```

NiklasBr:
Resolves #9457

## Additional info  

Only tested on 6.9.6 so far, but should probably work on 10.x as well :)

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20210702102100.php
+++ b/bundles/CoreBundle/Migrations/Version20210702102100.php
@@ -46,7 +46,7 @@ final class Version20210702102100 extends AbstractMigration
     /**
      * @throws \Exception
      */
-    private function rebuildClassesCommand()
+    private function regenerateClasses()
     {
         $listing = new Listing();
         foreach ($listing->getClasses() as $class) {
--- a/bundles/CoreBundle/Migrations/Version20210702102100.php
+++ b/bundles/CoreBundle/Migrations/Version20210702102100.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+use Pimcore\Tool\Console;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\Process;
+
+final class Version20210608094532 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Updates class definition files';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->rebuildClassesCommand();
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->rebuildClassesCommand();
+    }
+
+    /**
+     * @throws \Exception|ProcessFailedException
+     */
+    private function rebuildClassesCommand()
+    {
+        $cmd = [
+            Console::getPhpCli(),
+            'bin/console',
+            'pimcore:deployment:classes-rebuild',
+            '--create-classes',
+        ];
+
+        $process = new Process($cmd);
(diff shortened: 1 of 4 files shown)


dvesh3 (changes requested):
please add an example here https://github.com/pimcore/pimcore/blob/10.x/doc/Development_Documentation/05_Objects/03_Working_with_PHP_API.md#get-objects-matching-a-value-of-a-property thanks!

models/DataObject/ClassDefinition.php
```diff
@@ -504,7 +504,7 @@ private function generateClassFiles($generateDefinitionFile = true)
                             $this->getName()
                         ).' getBy'.ucfirst(
                             $def->getName()
-                        ).'($field, $value, $locale = null, $limit = 0, $offset = 0)'."\n";
+                        ).'($field, $value, $locale = null, $limit = 0, $offset = 0, $objectTypes = null)'."\n";
```
dvesh3, line 507: since this changes the class generation code. we need a migration, which calls`generateClassFiles` for exisiting generated classes something like https://github.com/pimcore/pimcore/blob/6.9/bundles/CoreBundle/Migrations/Version20200529121630.php#L34

bundles/CoreBundle/Migrations/Version20210702102100.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $this->rebuildClassesCommand();
+    }
+
+    /**
+     * @throws \Exception|ProcessFailedException
+     */
+    private function rebuildClassesCommand()
+    {
+        $cmd = [
```
dvesh3: rebuild command still saves the class and updates the db. As said before, we just need to regenerate the class files here.

bundles/CoreBundle/Migrations/Version20210702102100.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+use Pimcore\Model\DataObject\ClassDefinition\Listing;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\Process;
+
+final class Version20210608094532 extends AbstractMigration
```
dvesh3:
```suggestion
final class Version20210702102100 extends AbstractMigration
```

bundles/CoreBundle/Migrations/Version20210702102100.php
```diff
@@ -0,0 +1,57 @@
...
+
+    public function down(Schema $schema): void
+    {
+        $this->rebuildClassesCommand();
+    }
+
+    /**
+     * @throws \Exception
+     */
+    private function rebuildClassesCommand()
```
dvesh3:
```suggestion
    private function regenerateClasses()
```

bundles/CoreBundle/Migrations/Version20210702102100.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->rebuildClassesCommand();
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->rebuildClassesCommand();
```
dvesh3:
```suggestion
        $this->regenerateClasses();
```

bundles/CoreBundle/Migrations/Version20210702102100.php
```diff
@@ -0,0 +1,57 @@
...
+     * {@inheritDoc}
+     */
+    public function getDescription(): string
+    {
+        return 'Updates class definition files';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->rebuildClassesCommand();
```
dvesh3:
```suggestion
        $this->regenerateClasses();
```

dvesh3:
@NiklasBr please rebase to fix the test. see https://github.com/pimcore/pimcore/commit/241ccc21ddce3ac6e70bbbe56b0b914f6813c0c0 thanks!

dvesh3:
@NiklasBr 🏓

dvesh3 (approved):

dvesh3:
@NiklasBr thanks a lot!

PR #9635, opened 2021-06-30
Quicksearch: use class icons when available

NiklasBr:
Resolves #9498 which broke during rebase.


bundles/AdminBundle/Controller/Searchadmin/SearchController.php
```diff
@@ -478,14 +481,25 @@ public function quicksearchAction(Request $request, EventDispatcherInterface $ev
                     'fullpath' => htmlspecialchars($element->getRealFullPath()),
                     'fullpathList' => htmlspecialchars($this->shortenPath($element->getRealFullPath())),
                     'iconCls' => 'pimcore_icon_asset_default',
+                    'icon' => '',
                 ];
 
                 if ($element instanceof Asset) {
```
brusch: I think we don't need that whole code block at all (incl. for Assets), because I think that should be now covered by `$this->addAdminStyle($element, null, $data);`. See also: https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/models/Element/AdminStyle.php#L61

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
```diff
@@ -478,14 +481,25 @@ public function quicksearchAction(Request $request, EventDispatcherInterface $ev
...
+                        $data['icon'] = $element->getClass()->getIcon();
+                        $data['iconCls'] = '';
+                    } else {
+                        $data['iconCls'] .= ' pimcore_icon_' . $element->getType();
+                    }
                 } else {
                     $data['iconCls'] .= ' pimcore_icon_' . $element->getType();
                 }
 
+                $this->addAdminStyle($element, null, $data);
```
brusch: I think we should introduce a new context for the search `\Pimcore\Event\Admin\ElementAdminStyleEvent::CONTEXT_SEARCH` instead of passing `null`.

brusch (changes requested):

brusch:
@NiklasBr 🏓Thanks 😉

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
NiklasBr: In that case I will need to refactor how classes are treated for `variants` where the parent object has a custom icon since one is added by a `class` and the other as a an image `src`: 
image

brusch:
Excellent, thanks a lot! 👍

PR #9637, opened 2021-07-01
Consistent method name getMimeType/setMimeType

blankse:
Asset\Image has getMimetype and Asset\Image\Thumbnail has getMimeType.

PHP method names are case-insensitive. So it shouldn't be a BC break.


blankse:
I see there are much more "mimetype" nameings (setter, class property, db fieldname).

brusch:
@blankse are you updating your PR with the missing things (if possible w/o breaking)? There's also a conflict that needs to be solved 😊 Thanks a lot!

brusch:
@blankse 🏓😊 btw. there's no also a conflict that needs to be resolved. Thanks a lot!

blankse:
@brusch I rebased and changed also the setter

brusch:
Thanks a lot! 👍

Issue #9638, opened 2021-07-01
Error with Thumbnail::getStream() when Thumbnail::getMimeType() is called before


blankse:
# Bug Report

### Code example
```php
$image = Asset\Image::getById(123);
$imageThumbnail = $image->getThumbnail('jpeg');

// Generate file system path (but generate not the image)
$imageThumbnail->getMimeType();

// Try to load the file system path
$imageThumbnail->getStream();
```

### Output
```
Warning: fopen(/var/www/html/web/var/tmp/image-thumbnails/image-thumb__123__jpeg/test.jpg): failed to open stream: No such file or directory 
```

brusch:
Fixed by #9681

PR #9639, opened 2021-07-01
FIX: Thumbnail::getMimeType() fix (#9638)

dturopoli:
Fixes issue #9638


brusch:
Thanks! However, I think we can improve that a bit, I'll create a follow-up PR.

PR #9654, opened 2021-07-02
Fix: ObjectBrick with multiple Localized fields are broken after update to 6.9/10.0

### LINKED ISSUE #9646: ObjectBrick with multiple Localized fields are broken after update to 6.9/10.0
# Bug Report
If you have an object brick with multiple localized fields, its impossible to get the fields or save it in the backend.
For example:
image

image

image

This was all working in 6.7

### Expected behavior

Save and Read the fields

### Actual behavior

Getter doesn't exists and save in the admin interface doesn't work

blankse:
Resolve #9646

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20210706090823.php
+++ b/bundles/CoreBundle/Migrations/Version20210706090823.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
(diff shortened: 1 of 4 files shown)


blankse:
Classificationstore has also setReferencedFields/addReferencedField but no getFieldDefinitions method which call getReferencedFields. Is it there still needed?

dvesh3 (changes requested):
since this needs to regenerate brick tables...please add migration to save objectbricks

bundles/CoreBundle/Migrations/Version20210706090823.php
```diff
@@ -0,0 +1,62 @@
...
+use Pimcore\Model\DataObject;
+
+/**
+ * @internal
+ */
+final class Version20210706090823 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
```
dvesh3:
```suggestion
        return 'Rebuild class definitions to fix Objectbricks with multiple Localized fields';
```

dvesh3:
@blankse thanks a lot!

PR #9655, opened 2021-07-02
Properly save on save button

### LINKED ISSUE #9644: Template Reloading doesn’t work on unpublished page
**Action:**
1. Open Pimcore Backend
2. Create new page (type blank)
3. Change Template under „SEO & Settings“
4. Click on „Save“
5. Switch back to tab „Edit“
6. Click on „Refresh“ or clear cache and reload page

 <= Refresh-Button

**Expected Behaviour:**

Template Reloads

**Actual Behaviour:**

Nothing happens

**Tested on Versions:**

10.x-dev
10.0.2
6.9.3
6.9.4

(applies to all of them)

**Note:**

This only applies when the page is not published. After a click on the button „Save & Publish“, everything works fine.

BabovicT:
Resolves #9644


dvesh3:
@BabovicT thanks for your PR! I'm afraid this is not a proper fix, as it will still save a version in saveAction https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Controller/Admin/Document/PageController.php#L209 (which is the intended behavior).

please see PR https://github.com/pimcore/pimcore/pull/9657 to fix this issue, by checking if the request is from admin, then always use latest version for building route.

dvesh3:
We always appreciate contributions from our lovely community 🥰, so it is sad for us to close contributed PRs. I would like to suggest you guys to discuss your solutions on the issue itself before working on it. then it solves 2 problems for sure, 1) we know that someone is working it and saves rework 2) the solution is well discussed and proper 😉

PR #9658, opened 2021-07-02
[Docs] Remove false statement about calling bin/console

podarcis:
Explicitly executing `bin/console` w/ interpreter `php` makes no difference to implicit given interpreter in shebang `#!/usr/bin/env php` in regards to the effective user of the process.


brusch:
Thanks!

PR #9662, opened 2021-07-02
[DataObjects][Calculated Value] Add option to define calculation via symfony expression language

### LINKED ISSUE #9546: [DataObjects][Calculated Value] Add option to define calculation via symfony expression language
- allow definition of expression directly in class definition 
  - pass same parameters as to calculate method to execution context
- as an alternative to calculation service

- also add preview/test field with data object to drag for preview

----

unregister (or overwrite) `constant` function, just to be sure... see also [here](https://symfony.com/doc/current/components/expression_language/syntax.html#working-with-functions)

fashxp:
fixed #9546


fashxp:
missing

- [x] docs

PR #9678, opened 2021-07-05
[Admin Translations] Translations keys should be wrapped with + sign if translated value is missing

### LINKED ISSUE #9482: [Admin Translations] Debug Admin-Translations not working properly
`+` signs are only shown reliably when translation keys are not added to admin translations table. 
As soon as they are already added to the table (but still not translated), only some of them are shown.

### Steps to reproduce
- open admin translations and cleanup 
- reload pimcore admin ... all translation keys are wrapped in `+` signs
- reload pimcore admin once more ... not all translation keys are wrapped anymore

dvesh3:
Resolves #9482

## Additional info  
and debug is enabled

### DIFF
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -220,7 +220,7 @@ class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleA
 
                             //wrap non-translated keys with "+", if debug admin translations is enabled
                             if  ($debugAdminTranslations) {
-                                $translationKey = '+' . $translationKey . '+';
+                                $translationTerm = '+' . $translationTerm. '+';
                             }
                         }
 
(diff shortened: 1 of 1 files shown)


lib/Translation/Translator.php
```diff
@@ -216,6 +217,11 @@ public function lazyInitialize($domain, $locale)
 
                         if (empty($translationTerm)) {
                             $translationTerm = $translationKey;
+
+                            //wrap non-translated keys with "+", if debug admin translations is enabled
+                            if  ($debugAdminTranslations) {
+                                $translationKey = '+' . $translationKey . '+';
```
dvesh3:
```suggestion
                                $translationTerm = '+' . $translationTerm. '+';
```

fashxp (approved):

PR #9680, opened 2021-07-05
changed old method name to current

HaaseIT:
The Info object still had getTag(), which was replaced by getEditable().

Not sure if the line below needs a wording change.


brusch:
Thanks a lot 👍

PR #9688, opened 2021-07-06
[Data Object][BlockElement] Improve block data type memory usage

markus-moser:
Currently the object where a block element is used is loaded once from the cache for each block element within the object. So if we have for example 50 block elements within an object it will be loaded 51 times when you open it for example in the Pimcore backend. This could lead to serious memory problems.

This PR avoids this and the object will be loaded twice only instead of 51 times in the example above.


weisswurstkanone:
@markus-moser  thanks a lot!

PR #9689, opened 2021-07-06
Support endroid/qr-code 4

blankse:
I use endroid/qr-code-bundle and I get deprecation notices in V3. V4 has PHP 8 support and fixed this notices.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/PageController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PageController.php
@@ -15,6 +15,8 @@
 
 namespace Pimcore\Bundle\AdminBundle\Controller\Admin\Document;
 
+use Endroid\QrCode\Builder\Builder;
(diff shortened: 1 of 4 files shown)


brusch:
@blankse are there any BC breaks in v4?

blankse:
@brusch Yes the API in v4 is completely different.
But is it a BC break for Pimcore? See https://github.com/semver/semver/issues/148
If someone has also used the qr-code dependency with v3 the composer would not be update until he also update to v4.
The qr-code-bundle is unusable with PHP 8. Every page has outprinted notices...

I can try to make a wrapper to support v3 and v4 so somebody can stay on v3.

brusch:
@blankse I'm fine with that, thanks a lot! Can you just resolve the conflict?

blankse:
@brusch I rebased the branch.

brusch:
@blankse Thanks a lot! 👍

PR #9693, opened 2021-07-06
[Assets] Update children path on remote storage

### LINKED ISSUE #9367: Asset paths not updated after S3 setup
Expected Behaviour

If you move an asset folder with its content in the directory tree then the full path of the folder and the assets should be updated respectively 

Actual behaviour
If you setup S3 file storage and move an asset folder with its content in the directory tree then the full path of the folder is being updated but the full path of the the assets inside the folder does not change.

Example:
Original Pimcore X Asset Directory
 
image

Move /Hrs/Anw/sample.jpg => /Media/Images/Detail/Anw/sample.jpg

 
image

Expected full path: /Media/Images/Detail/Anw/sample.jpg

Actual path: /Hrs/Anw/sample.jpg

Steps to reproduce:

1. Right click on specific asset folder and select “Cut”
2. Right click on destination folder and select “Paste cut-out element”
3. Left click on specific asset in the moved folder and from the menu “copy full path to clipboard”

dvesh3:
Resolves #9367

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1988,6 +1988,8 @@ class Asset extends Element\AbstractElement
         } catch (UnableToMoveFile $e) {
             // noting to do
         }
+
+        return $this->getDao()->updateChildPaths($oldPath);
     }
 
     /**
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -542,11 +542,12 @@ class Asset extends Element\AbstractElement
                         try {
                             $storage->move($oldPath, $this->getRealFullPath());
                         } catch (UnableToMoveFile $e) {
-                            //nothing to do
+                            //update children, if unable to move parent
+                            $this->updateChildPaths($storage, $oldPath);
                         }
 
                         $this->getDao()->updateWorkspaces();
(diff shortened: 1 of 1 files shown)


models/Asset.php
```diff
@@ -1963,6 +1964,34 @@ public function clearThumbnails($force = false)
...
     }
 
+    /**
+     * @param string $oldPath
+     *
+     * @return array
+     *
+     * @throws \League\Flysystem\FilesystemException
+     */
+    private function updateChildPaths(string $oldPath)
```
brusch: Just moving the parent folder should be enough, right? Because all children of the parent will be moved anyway when calling `$storage->move()` for it, or what is the reason for that? 
If not we'd have to think about a more robust solution, because this only works for a certain amount of children (I'm just thinking about a folder with millions of assets in it 😁).

brusch (changes requested):

models/Asset.php
dvesh3: `$storage->move()` fails when you move a sub folder to another folder(S3 does not allow you to move directories), so this is required so that children of parent folder are also moved as well.

PR #9694, opened 2021-07-06
Element Router with string name

blankse:
It is not possible to set an route with an object as name. We must use a string and give the object as parameter.

Before:
`{{ url(document) }}`
`{{ url(object) }}`

After:
`{{ url('pimcore_element', {'element': document}) }}`
`{{ url('pimcore_element', {'element': object}) }}`

### DIFF
--- a/lib/Routing/Element/Router.php
+++ b/lib/Routing/Element/Router.php
@@ -75,7 +75,7 @@ class Router implements RouterInterface, RequestMatcherInterface, VersatileGener
      */
     public function supports($name)
     {
-        return $name instanceof ElementInterface;
+        return $name === 'pimcore_element';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


brusch:
But this would break existing `{{ url(document) }}` code, right? 🤔

blankse:
@brusch This code doesn't work anymore. Symfony 5 has a Typehint string on the name param. So it is only the fullpath string and the router dosn't match.

dpfaffenbauer:
1. It should either resolve the full path
2. or it is a bc break: you cannot break what is already broken right? :P

lib/Routing/Element/Router.php
```diff
@@ -75,16 +75,17 @@ public function getContext()
      */
     public function supports($name)
```
jdreesen, line 76:
```suggestion
    public function supports(string $name)
```

brusch:
@blankse I see, then it's fine of course 😊

lib/Routing/Element/Router.php
```diff
@@ -73,18 +73,19 @@ public function getContext()
     /**
      * {@inheritdoc}
      */
-    public function supports($name)
+    public function supports(string $name)
```
brusch:
```suggestion
    public function supports($name)
```

lib/Routing/Element/Router.php
brusch: Unfortunately this breaks the code: 
Declaration of Pimcore\Routing\Element\Router::supports(string $name) must be compatible with Symfony\Cmf\Component\Routing\VersatileGeneratorInterface::supports($name)

PR #9710, opened 2021-07-09
[Docs] Fix default rounding mode

jremmurd:
This PR changes the described default rounding mode from `PHP_ROUND_HALF_EVEN` to `PHP_ROUND_HALF_UP`, 
see https://github.com/pimcore/pimcore/blob/10.x/bundles/EcommerceFrameworkBundle/Type/Decimal.php#L114


dvesh3:
@jremmurd thanks!

PR #9716, opened 2021-07-09
[DataObjects] Fix localized calculated fields 

jremmurd:
# Bugfix

With this PR calculated fields in localized fields return the correct locale in cli context.

## Steps to reproduce
1. Create calculated field inside a localized field (this could be the name of a product for example)
image
 2. create command
 3. set locale to a none-default locale via `pimcore.locale` service
 4. calculated field will always return default locale


brusch:
Thanks a lot! 👍

PR #9717, opened 2021-07-09
[Login][Events] Added event before redirecting to login page

fashxp:
Adds an event before redirecting to login page when user is not logged in. This event allows you to influence path of the login page, e.g. for SSO integrations.

### DIFF
--- a/lib/Event/Admin/Login/LoginRedirectEvent.php
+++ b/lib/Event/Admin/Login/LoginRedirectEvent.php
@@ -9,7 +9,7 @@ class LoginRedirectEvent extends Event
 {
 
     /**
-     * @var
+     * @var string
      */
     protected $routeName;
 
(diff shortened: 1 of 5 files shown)


lib/Event/Admin/Login/LoginRedirectEvent.php
```diff
@@ -0,0 +1,64 @@
...
+    protected $routeName;
+
+    /**
+     * @var array
+     */
+    protected $routeParams;
+
+    /**
+     * LoginRedirectEvent constructor.
+     * @param $routeName
```
weisswurstkanone: string ?

lib/Event/Admin/Login/LoginRedirectEvent.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @var array
+     */
+    protected $routeParams;
+
+    /**
+     * @param string $routeName
+     * @param array $routeParams
+     */
+    public function __construct(string $routeName, array $routeParams)
```
brusch:
```suggestion
    public function __construct(string $routeName, array $routeParams = [])
```
I think this could be optional, right?

brusch (changes requested):

PR #9725, opened 2021-07-12
[Fixed] Error: "Getting Array to string conversion" on installation

pdchaudhary:
**Steps to reproduce:**
1. Run "COMPOSER_MEMORY_LIMIT=-1 composer create-project pimcore/skeleton my-project" cmd

**Error:**
Script Pimcore\Composer::clearCache handling the pimcore-scripts event terminated with an exception
[ErrorException]
Array to string conversion

### DIFF
--- a/lib/Composer.php
+++ b/lib/Composer.php
@@ -172,7 +172,7 @@ class Composer
 
         //$event->getIO()->write('Run command: ' . implode(' ', $command), false);
 
-        $process = new Process($command, null, null, null, $timeout);
+        $process = new Process(implode(' ', $command), null, null, null, $timeout);
         $process->run(function ($type, $buffer) use ($event, $writeBuffer) {
             if ($writeBuffer) {
                 $event->getIO()->write($buffer, false);


brusch:
@pdchaudhary Are you sure?  In Symfony 5 `new Process()` only accepts arrays as the first argument, see: 
https://github.com/symfony/symfony/blob/5.3/src/Symfony/Component/Process/Process.php#L141

pdchaudhary:
@brusch Yes, you are right about accepts arrays but I have tried again it is still giving me the attached error. 

image

but it is giving me error on below code:

```
  $process->run(function ($type, $buffer) use ($event, $writeBuffer) {
            if ($writeBuffer) {
                $event->getIO()->write($buffer, false);
            }
        });
```

Platform: **Windows**

brusch:
Maybe it's a Windows issue, but the code in question is definitely correct. 😊

pdchaudhary:
Okay thx, I will try on Linux subsystem.

PR #9616, opened 2021-06-30
Move Marketing Settings to SettingsStore 

cipribucur:
## Changes in this pull request  
Resolves #8555

PR #9729, opened 2021-07-12
Fix Error: "var/config/reports.php doesn't exist"

blankse:
Error in migration Version20210630062445:
```
[error] Migration Pimcore\Bundle\CoreBundle\Migrations\Version20210630062445 failed during Execution. Error: "/var/www/html/var/config/reports.php doesn't exist"

In Config.php line 922:
                                                      
  /var/www/html/var/config/reports.php doesn't exist  
```


brusch:
Related to #9726  and caused by #9616

brusch:
Thanks a lot! 👍

PR #9730, opened 2021-07-12
Non-parsable SQL query checking grid permissions

kjkooistra-youwe:
Doctrine\DBAL\Driver\PDO\Exception: SQLSTATE[42S22]:
Column not found: 1054 Unknown column '' in 'order clause'
Caused by permissions check without type for custom grid field definitions
https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/Service.php#L1182
Fixed by adding the type order by section conditionally when a type is provided

**To reproduce**
Refresh the grid view in a data object folder when logged in with a user restricted (via roles) using workspaces.
The query doesn't appear to be causing an error in the interface, but I ran into it via a monitoring system.

### DIFF
--- a/models/DataObject/AbstractObject/Dao.php
+++ b/models/DataObject/AbstractObject/Dao.php
@@ -548,7 +548,8 @@ class Dao extends Model\Element\Dao
                 return $firstPermission;
             }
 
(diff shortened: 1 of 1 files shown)


dvesh3:
thanks!

PR #9753, opened 2021-07-14
Consistent behaviour for document title/navigation/key

### LINKED ISSUE #8907: Consistent behaviour for document Title/Navigation/Key
The field order for "Add Page" is Title, Navigation, Key.
When changing the Title, the other fields get prefilled automatically.

image

The order in "Create Translation" (Document Language Overview) is Key, Navigation, Title.
Changing the Key prefills Navigation, Title stays empty. Changing the Title does nothing.

image

The same logic should be used for use cases.

In addition it would be nice to create a nice key/slug automatically.

As shown in the pimcore demo the following Keys are used:

image

Which results in URLs like: https://demo.pimcore.com/en/Magazine/Pick%20of%20the%20Day

In my opinion the Key should get prefilled with a nice slug (pick-of-the-day instead of Pick of the Day) if the field is empty.

Version: dev-master|add74462508e662a032f297149e1a7dfe5c042d2

valentin-:
Resolves #8907

## Additional info  

- Changed field order
- Added/moved key event to title

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/document_language_overview.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/document_language_overview.js
@@ -242,15 +242,16 @@ pimcore.document.document_language_overview = Class.create({
                         },
                         items: [{
                             xtype: "textfield",
+                            itemId: "title",
+                            fieldLabel: t('title'),
+                            name: 'title',
                             width: "100%",
-                            fieldLabel: t('key'),
-                            itemId: "key",
-                            name: 'key',
                             enableKeyEvents: true,
                             listeners: {
                                 keyup: function (el) {
                                     pageForm.getComponent("name").setValue(el.getValue());
-                                }
+                                    pageForm.getComponent("key").setValue(el.getValue());
+                                }.bind(this)
                             }
                         },{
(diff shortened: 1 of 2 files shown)


dvesh3:
@valentin- thanks for your PR. please sign CLA  https://github.com/pimcore/pimcore/pull/9753#issuecomment-879638066

bundles/AdminBundle/Resources/public/js/pimcore/document/document_language_overview.js
```diff
@@ -242,15 +242,16 @@ pimcore.document.document_language_overview = Class.create({
                         },
                         items: [{
                             xtype: "textfield",
+                            itemId: "title",
```
dvesh3, line 245: I would also make it consistent for translation new document option here https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/document/document.js#L481

valentin-: @dvesh3 thanks, good input. I've added another commit for this.

brusch:
Nice, thanks a lot! 👍

PR #9767, opened 2021-07-14
[Data objects] Grid: Flexible field width for class selection dropdown list

BlackbitDevs:
When you have long class names it is currently difficult to select the right one (you have to horizontically scroll if visible class prefix is completely identical):
Bildschirmfoto 2021-07-14 um 16 02 12

With this PR the dropdown list's width gets automatically adjusted to the longest class name in the dropdown:
Bildschirmfoto 2021-07-14 um 16 01 24

Furthermore with this PR you can search within the classname not only at the beginning of it.


brusch:
@BlackbitNeueMedien can you please check the static analysis action that failed? Thanks a lot!

jdreesen:
@brush I doubt that it's related, as this PR only changes some JavaScript.

blankse:
@brusch I created a PR which fix the errors https://github.com/pimcore/pimcore/pull/9770

BlackbitDevs:
Shall I do anything? Or simply wait for #9770 to get merged and afterwards merge branch `10.x` to my branch?

dvesh3:
> Shall I do anything? Or simply wait for #9770 to get merged and afterwards merge branch `10.x` to my branch?

@BlackbitNeueMedien please wait and merge branch `10.x` afterwards. thanks!

brusch:
@BlackbitNeueMedien can you please merge/rebase the up-to-date `10.x` branch? Thanks :)

weisswurstkanone:
Just as a discussion point: Ext says that "matchFieldWidth" is true by default. Maybe set it to false via an override (override.js) meaning that this would apply to all dropdowns.

BlackbitDevs:
@weisswurstkanone Actually I think this would make sense but on the other hand I am wondering if ExtJs had a reason for those default values or if those were only chosen for backwards compatiblity. 
The same also applies for `enableTextSelection` for grid panels. Over the time I have created 4 or 5 PRs to allow selecting / copying text from the various grids. And last week I noticed that it is still not possible to copy asset metadata.
Also `anyMatch` for combo boxes could be a candidate.

weisswurstkanone:
Exactly. Thanks for your valuable feedback. Maybe let us move this "general" thought to the discussion area as not directly related to your PR.

PR #9770, opened 2021-07-14
Fix new phpstan error with Symfony 5.4-dev

blankse:
The current 10.x branch have PhpStan errors with Symfony 5.4.x-dev
```
------ ----------------------------------------------------------------------------------- 
  Line   bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php  
 ------ ----------------------------------------------------------------------------------- 
  1182   If condition is always true.                                                       
 ------ ----------------------------------------------------------------------------------- 

  Line   bundles/AdminBundle/Controller/Admin/EmailController.php             
  171    Call to function is_subclass_of() with non-empty-string and          
         'Pimcore\\Model\\Element\\ElementInterface' will always evaluate to  
         false.                                                               
  171    Result of && is always false.                                        
  528    Call to function is_subclass_of() with non-empty-string and          
         'Pimcore\\Model\\Element\\ElementInterface' will always evaluate to  
         false.                                                               
  528    Result of && is always false.                                        

 ------ ---------------------------------------------------- 
  Line   models/DataObject/ClassDefinition/Data/Numeric.php  
 ------ ---------------------------------------------------- 
  524    Left side of && is always false.                    
  528    Left side of && is always false.                    
 ------ ---------------------------------------------------- 

 ------ -------------------------------------------------------------------- 
  Line   models/DataObject/ClassDefinition/Data/QuantityValue.php            
 ------ -------------------------------------------------------------------- 
  171    Comparison operation ">" between int<1, max> and 0 is always true.  
 ------ -------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------- 
  Line   models/DataObject/Classificationstore/KeyConfig/Listing/Dao.php  
 ------ ----------------------------------------------------------------- 
  84     If condition is always true.                                     
  88     Unreachable statement - code above always terminates.            
 ------ ----------------------------------------------------------------- 

 ------ --------------------------------------------- 
  Line   models/Staticroute.php                       
 ------ --------------------------------------------- 
  398    Ternary operator condition is always false.  
 ------ --------------------------------------------- 
```

### DIFF
--- a/models/DataObject/ClassDefinition/Data/QuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/QuantityValue.php
@@ -57,7 +57,7 @@ class QuantityValue extends Data implements ResourcePersistenceAwareInterface, Q
     /**
      * @internal
      *
-     * @var float
+     * @var float|null
      */
     public $defaultValue;
 
@@ -154,23 +154,17 @@ class QuantityValue extends Data implements ResourcePersistenceAwareInterface, Q
     /**
      * @return float|null
      */
-    public function getDefaultValue()
+    public function getDefaultValue(): ?float
     {
-        if ($this->defaultValue !== null) {
-            return (float) $this->defaultValue;
-        }
-
-        return null;
+        return $this->defaultValue;
     }
 
     /**
-     * @param int $defaultValue
+     * @param float|null $defaultValue
      */
-    public function setDefaultValue($defaultValue)
+    public function setDefaultValue(?float $defaultValue): void
     {
-        if (strlen((string)$defaultValue) > 0) {
-            $this->defaultValue = $defaultValue;
-        }
+        $this->defaultValue = $defaultValue;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -154,23 +154,17 @@ public function setUnitWidth($unitWidth)
     /**
      * @return float|null
      */
-    public function getDefaultValue()
+    public function getDefaultValue(): ?float
```
brusch: I'm afraid this is a BC break, right?

models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -154,23 +154,17 @@ public function setUnitWidth($unitWidth)
...
-        return null;
+        return $this->defaultValue;
     }
 
     /**
-     * @param int $defaultValue
+     * @param float|null $defaultValue
      */
-    public function setDefaultValue($defaultValue)
+    public function setDefaultValue(?float $defaultValue): void
```
brusch: This as well?

brusch (changes requested):

models/DataObject/ClassDefinition/Data/QuantityValue.php
blankse: @brusch Changed it

brusch:
Thanks a lot! 👍

PR #9780, opened 2021-07-16
Fix: labelWidth must be an integer

### LINKED ISSUE #9761: [Bug]: Data Object Class "Label Width" within localized fields
### Expected behavior

If you set a label width "150" for localized fields the label width is as expected.

### Actual behavior

The label width is not "150" but "1505px" in the dom element and breaks the whole field display.

### Steps to reproduce

- Add localized fields to an data object class
- Set label width "150" for the localized fields
- Save the class
- Open class item

blankse:
Resolves #9761

## Additional info  
ExtJS doesn't support labelWidth in percent. Must be an integer.
See: https://docs.sencha.com/extjs/7.4.0/classic/Ext.form.field.Text.html

> labelWidth : Number 
> The width of the fieldLabel in pixels. Only applicable if labelAlign is set to "left" or "right".

Open discussion: https://forum.sencha.com/forum/showthread.php?339242-How-to-specify-labelWidth-in-percentage


brusch:
LGTM, thanks a lot! 👍

PR #9789, opened 2021-07-16
fixed speed - better to split into 2 queries then use OR condition

dlhck:
When working with fieldcollections and having a lot of objects of the same type the `save()` gets kinda slow, because big tables has to go through all rows so its killing the index.

### DIFF
--- a/models/DataObject/Fieldcollection/Dao.php
+++ b/models/DataObject/Fieldcollection/Dao.php
@@ -248,11 +248,13 @@ class Dao extends Model\Dao\AbstractDao
         }
 
         $where = "(ownertype = 'fieldcollection' AND " . $this->db->quoteInto('ownername = ?', $this->model->getFieldname())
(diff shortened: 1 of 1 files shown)


dvesh3:
Hi @davidhoeck! Pimcore 6.x reached it EOL. Could you please rebase it on 10.0 branch? thanks!

dvesh3:
@DaffyM could you please sign CLA? thanks!

brusch:
@davidhoeck rebasing not necessary anymore - I've changed the base and history is still clean. 
@DaffyM but signing the CLA would be still necessary. Thanks!

dlhck:
@brusch @dvesh3 does it mean that 6.9 won't get any updates anymore, even if they are coming from community??

brusch:
@davidhoeck yes, EOL means no more updates 😉

brusch:
For those wo still want support for versions that have already reached end-of-life in the community version, we have our commercial long term support offering: 
https://pimcore.com/en/services/lts

brusch:
Thanks a lot!

PR #9797, opened 2021-07-19
Fix wrong PhpDocs

blankse:


brusch:
Looks very good to me, thanks a lot! 👍

PR #9798, opened 2021-07-19
Added table copy all functionality

karlozrilic:
I think it would be nice to have copy functionality since there is paste functionality already present. We encountered issue multiple times where we had to manually copy paste every cell data from one table to another. So I suggest copy button that opens up modal with text box that can be edited and you can easily copy all it contents with one click.
Screenshot_3
Screenshot_4


brusch:
Makes totally sense and works fine - well done! 👍

PR #9809, opened 2021-07-20
Greater lock TTL for parallel commands

BlackbitDevs:
By default all locks currently use the PDOStore, see https://github.com/pimcore/pimcore/blob/eed98a084d6e8852788cbbcaf8f2b67fe60a1ba6/bundles/CoreBundle/Resources/config/services.yml#L226-L229

This store supports TTL and the default TTL is 300 seconds: https://github.com/symfony/lock/blob/69902b2719376c83712c83a265ed5469b1e7c92d/LockFactory.php#L56

For parallel commands which use the `Parallelization` trait currently the default TTL is used (no second argument given for TTL):
https://github.com/pimcore/pimcore/blob/eed98a084d6e8852788cbbcaf8f2b67fe60a1ba6/lib/Console/Traits/Parallelization.php#L137

This leads to multiple parallel processes if the command execution needs longer than 300 seconds. Depending on how long the first process needs, this can stack up until the server goes down.
This PR changes the TTL to 1 day, in analogy to https://github.com/pimcore/pimcore/blob/eed98a084d6e8852788cbbcaf8f2b67fe60a1ba6/lib/Maintenance/Executor.php#L86

PS: I would prefer using a lock store which does not care about TTL like FlockStore but this may be discussed elsewhere.

### DIFF
--- a/lib/Console/Traits/Parallelization.php
+++ b/lib/Console/Traits/Parallelization.php
@@ -120,7 +120,7 @@ trait Parallelization
      */
     private function lock($name = null, $blocking = false)
     {
-        $this->lock = \Pimcore::getContainer()->get(LockFactory::class)->createLock($name ?: $this->getName());
+        $this->lock = \Pimcore::getContainer()->get(LockFactory::class)->createLock($name ?: $this->getName(), 86400);
 
         if (!$this->lock->acquire($blocking)) {
             $this->lock = null;


jdreesen:
Wouldn't it be better to leave the TTL low and instead refresh the lock while the command is still running? As explained in: https://symfony.com/doc/current/components/lock.html#expiring-locks

BlackbitDevs:
@jdreesen Yes, would also work, we could implement refreshing the lock in https://github.com/pimcore/pimcore/blob/eed98a084d6e8852788cbbcaf8f2b67fe60a1ba6/lib/Console/Traits/Parallelization.php#L93
But I guarantee you that there will be a use-case where one batch will need longer than 300 seconds and then it will fail again. But I am quite sure that there also will be use cases where the TTL of 1 day for the complete process will not be sufficient. Those problems can only be solved by TTL-independent lock stores...

brusch:
@BlackbitNeueMedien Thanks a lot for your PR. 
Unfortunately `FlockStore` has some other drawbacks, that's why we had to change the locking mechanism in the parallelization trait in #9117
I'd give it a try with the TTL of 1 day and see how it goes ... 😊

PR #9814, opened 2021-07-20
Fix deprecation warnings

blankse:
```
Since symfony/dependency-injection 5.1: Not setting the attribute "package" of the "deprecated" option in "/var/www/html/vendor/pimcore/pimcore/bundles/CoreBundle/DependencyInjection/../Resources/config/services.yml" is deprecated.

Since symfony/dependency-injection 5.1: Not setting the attribute "version" of the "deprecated" option in "/var/www/html/vendor/pimcore/pimcore/bundles/CoreBundle/DependencyInjection/../Resources/config/services.yml" is deprecated.

Since symfony/dependency-injection 5.1: Not setting the attribute "package" of the "deprecated" option in "/var/www/html/vendor/pimcore/pimcore/bundles/CoreBundle/DependencyInjection/../Resources/config/templating.yml" is deprecated.

Since symfony/dependency-injection 5.1: Not setting the attribute "version" of the "deprecated" option in "/var/www/html/vendor/pimcore/pimcore/bundles/CoreBundle/DependencyInjection/../Resources/config/templating.yml" is deprecated.

Since symfony/dependency-injection 5.2: The "Symfony\Component\DependencyInjection\Definition::setPrivate()" method is deprecated, use "setPublic()" instead.
```

### DIFF
--- a/bundles/CoreBundle/DependencyInjection/Compiler/WorkflowPass.php
+++ b/bundles/CoreBundle/DependencyInjection/Compiler/WorkflowPass.php
@@ -214,7 +214,7 @@ final class WorkflowPass implements CompilerPassInterface
             // Enable the AuditTrail
             if ($workflowConfig['audit_trail']['enabled']) {
                 $listener = new Definition(Workflow\EventListener\AuditTrailListener::class);
-                $listener->setPrivate(true);
+                $listener->setPublic(false);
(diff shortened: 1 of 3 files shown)


blankse:
PhpStan errors are fixed with https://github.com/pimcore/pimcore/pull/9815

dvesh3:
thanks a lot!

Issue #8197, opened 2021-02-17
[Documents] Headless CMS


brusch:
It should be possible to define a content structure without the need of specifying a template. 
The concept could be similar to the one we're using already in a minimal version for the [Editable Dialog](https://pimcore.com/docs/pimcore/current/Development_Documentation/Documents/Editables/Areablock/Bricks.html#page_Editable-Dialog-since-6-8). 

Possible approach: 
- Defining the content-structure in a YAML file
- The YAML file is then specified as the "template" on a page/snippet
- If a YAML file is provided as a template, then Pimcore knows it's a headless content strucure
- The editing interface is generated according to the content structure defined in the YAML

Additionally it would be nice to have a preview generator, that is used to generate the preview URL (see also [Link Generators for Objects](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Class_Settings/Link_Generator.html) )

### Example YAML

Could be stored in `/config/pimcore/headless-definitions/example.yaml`

```yaml
content:
    headline:
        type: input
        config:
            placeholder: 'Headline'
            required: true

    selectTest:
        type: select
        config:
            store:
                - ["one-month", "One month"]
                - ["three-months", "three-months"]
                - ["unlimited", "unlimited"]
    mainContent:
        type: wysiwyg
        config:
            height: 200

    contentBlocks:
        type: areablock
        bricks:
            teasers:
                teaser_1:
                    type: snippet
                teaser_2:
                    type: snippet
            gallery:
                imagesBlock:
                    type: block
                    items:
                        image:
                            type: image
                            width: 200
                            height: 200
                            thumbnail: gallery
            video:
                video:
                    type: video
            productListing:
                products:
                    type: relations
                    types: ["object"]
                    classes: ["Product"]
```

julkue:
@brusch Thanks for bringing this up. I've also brought this up in a similar way in the DataHub repository:
https://github.com/pimcore/data-hub/issues/113

brusch:
We've decided to implement this feature in a bundle, all necessary core changes have been made (#9816), so I'll close this ticket here.

PR #9816, opened 2021-07-20
[Core] Changes to support headless documents

dvesh3:
Related to https://github.com/pimcore/pimcore/issues/8197


models/Document/Dao.php
```diff
@@ -98,7 +98,7 @@ public function update()
     {
         $typeSpecificTable = null;
         $validColumnsTypeSpecific = [];
-        if (in_array($this->model->getType(), ['email', 'newsletter', 'hardlink', 'link', 'page', 'snippet'])) {
+        if (in_array($this->model->getType(), ['email', 'newsletter', 'hardlink', 'link', 'page', 'snippet', 'headlesspage'])) {
```
brusch: I guess it's time to move the available types to Symfony config, wdyt?

brusch (changes requested):

models/Document/Dao.php
dvesh3: yes, makes sense! Moved it 😉

PR #9819, opened 2021-07-20
NodeDefinition::setDeprecated() - Fix deprecation warnings

blankse:
```
Since symfony/config 5.1: The signature of method "Symfony\Component\Config\Definition\Builder\NodeDefinition::setDeprecated()" requires 3 arguments: "string $package, string $version, string $message", not defining them is deprecated.
```


dvesh3:
follow up to #9178

dvesh3:
thanks!

PR #9828, opened 2021-07-21
Remove document object from Mail object before sending

### LINKED ISSUE #9824: [Bug]:  Circular reference error when sending emails using Symfony Messenger
### Expected behavior

When using a messenger transport the given object should be serializable with the default symfony serializer to whatever transport you have defined. In my case this is the `symfony/doctrine-messenger`

### Actual behavior

Because Symfony tries to serialize the `\Pimcore\Mail` class and this class contains the property `$document` of type `\Pimcore\Model\Document\Email` it fails to serialize the class because the document has a `$dao` property which has `$model` property which is the same as the `$document` property as before.

### Steps to reproduce

Use the Symfony serializer to serialize a `\Pimcore\Mail` object OR Setup the Symfony messenger to send `Symfony\Component\Mailer\Messenger\SendEmailMessage` class over a transport that needs to serialize the class.

See documentation on how to do this here: https://symfony.com/doc/current/mailer.html#sending-messages-async

rubanooo:
Resolves #9824

## Additional info  

This is not an optimal solution, but probably the best solution until Pimcore supports serializing objects using the Symfony serializer.


dvesh3:
Hi @rubanooo Could you please rebase on branch 10.0? since this is a fix. thanks!

rubanooo:
@dvesh3 I have created a new pull request to to proper branch because I messed up my rebase :(

https://github.com/pimcore/pimcore/pull/9838

PR #9838, opened 2021-07-22
Remove document from mail object before sending

### LINKED ISSUE #9824: [Bug]:  Circular reference error when sending emails using Symfony Messenger

rubanooo:
Resolves  #9824

## Additional info  

This is not an optimal solution, but probably the best solution until Pimcore supports serializing objects using the Symfony serializer.

### DIFF
--- a/lib/Mail.php
+++ b/lib/Mail.php
@@ -488,6 +488,10 @@ class Mail extends Email
 
         $this->subject($this->getSubjectRendered());
 
+        // Remove the document property because it is no longer needed and makes it difficult
+        // to serialize the Mail object when using the Symfony Messenger component
+        $this->setDocument(null);
+
         return $this->sendWithoutRendering($mailer);
     }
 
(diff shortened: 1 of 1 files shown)


lib/Mail.php
```diff
@@ -488,6 +488,10 @@ public function send(MailerInterface $mailer = null)
 
         $this->subject($this->getSubjectRendered());
 
+        // Remove the document property because it is no longer needed and makes it difficult
+        // to serialize the Mail object when using the Symfony Messenger component
+        $this->setDocument(null);
```
dvesh3, line 493: please adapt `setDocument()` method to nullify `$this->document` property. At the moment, it will throw exception `document is not an instance of Pimcore\Model\Document\Email` https://github.com/pimcore/pimcore/blob/e814e9cb0ae8300a2cf27d1b78db391168a0c0c9/lib/Mail.php#L769-L774

rubanooo: Ah, I added this to my local test version but forgot to add it to the PR

dvesh3:
thanks!

PR #9862, opened 2021-07-26
[Actions] Added PHP 8.1 support

brusch:


brusch:
So far everything prepared from Pimcore's side, however there are some dependencies which currently do not support PHP 8.1 yet. Also it was necessary to use a dev version of `league/csv` which needs to be reverted as soon as they have tagged a release including the fix.

PR #9866, opened 2021-07-26
[Data objects] Cloning many-to-one relation keeps reference

BlackbitDevs:
Actually `\Pimcore\Model\Service::cloneMe()` is supposed to create deep clones of elements. But if an element is directly related to the element which shall be cloned (e.g. many-to-one relations, properties of type object / asset / document etc.) the clone currently still references the original object.
Steps to reproduce:
1. Create class with a many-to-one relation field named `href`, allow objects of all classes
2. Create object `A` under `/`, save & publish
3. Create object `B`, assign `A` in the many-to-one relation
4. Use this script:
```php
<?php
$object = \Pimcore\Model\DataObject\Concrete::getByPath('/B');
$clone = \Pimcore\Model\Element\Service::cloneMe($object);
$clone->getHref()->setKey(uniqid());
var_dump($object->getHref()->getKey()); // outputs a uniqid instead of A
```

When the elements are not directly connected but for example a property consists of an **array of Pimcore elements**, then the related objects get really cloned.

Perhaps @weisswurstkanone remembers why the deepcopy filter which I have removed in this PR got added. Or perhaps why the constructor argument of this anonymous class is not used anywhere, perhaps it was supposed to do something different than just to check if the property value implements `ElementInterface`.


BlackbitDevs:
Test error is not caused by this PR's changes.

weisswurstkanone:
@BlackbitNeueMedien  Yes, I agree with you. Don't forget that the code base changed a little bit in the meantime.

BlackbitDevs:
> Don't forget that the code base changed a little bit in the meantime.

Yes, was no offense. Actually I stumbled upon the problem because we had memory issues when elements have lots of dependencies which have even more dependencies etc.
For this reason I looked how Pimcore core does it - of course by using the `KeepFilter` for elements a lot of memory is saved - but actually it makes cloning useless.
Do you know if deepcopy loads all the lazy-loaded stuff by default? Because if we remove the `KeepFilter` with this there might be a memory-related issue soon...

weisswurstkanone:
@BlackbitNeueMedien Thx! 😄 Regarding loading lazy fields: https://github.com/pimcore/pimcore/blob/9485aafb2ca1f78c3227dfa4bb159893a48a8853/models/Element/Service.php#L1264

weisswurstkanone:
@BlackbitNeueMedien,

After thinking about it twice ... it is not the intention to clone the related items because:

- The goal is to get a copy of the source item, modify some properties and then persist it. That's why it is called "cloneMe" and not "cloneTheUniverse". Just imagine if "/A" has a reference to "/X" and so on ... 
DeepCopy would freak up. So we don't copy any outging stuff. The current constructor thing is useless I guess. 

- Where it came from: So the real world scenario would be: In the Backend UI. Do a "copy & paste" and save it.

- Related items will not get persisted anyway because they didn't change. They are just a reference.

- Memory and processing time issues where the main reason!

I would suggest to introduce an optional parameter to turn the KeepFilter off. 

What do you think ?

BlackbitDevs:
Hi @weisswurstkanone ,
> The goal is to get a copy of the source item, modify some properties and then persist it. That's why it is called "cloneMe" and not "cloneTheUniverse". Just imagine if "/A" has a reference to "/X" and so on ...
DeepCopy would freak up. So we don't copy any outging stuff.

Ok, then perhaps I understood the method wrong but then its behaviour with many-to-many relations is wrong because those currently do get cloned - in fact everything which is assigned to the cloned object via an array gets cloned and everything which is assigned to an object property directly stays referenced. And if it really shall be a shallow copy, we could use PHP's default `clone` instead of deepcopy, couldn't we? So I actually think the method originally was intended to create a deep copy.

But anyway: I would suggest to use the same behaviour for many-to-one relations (and other 1:1 relations to the cloned element like properties) as for many-to-many relations (and image galleries, advanced relations where multiple items can be related to the cloned element etc.) - so either adjust the `KeepFilter` to be applied also for arrays of `ElementInterface` elements or remove it.

weisswurstkanone:
ok, could you adapt this accordingly ?

BlackbitDevs:
You mean `adjust the KeepFilter to be applied also for arrays of ElementInterface elements`? Because the `or remove it` part is currently implemented by this PR ;-)

weisswurstkanone:
yes, the first one. But still not sure why DeepCopy would keep the elements even if it is an array. maybe time to debug this :-)

weisswurstkanone:
@BlackbitNeueMedien 🏓

brusch:
@BlackbitNeueMedien any updates? If not, I'll close this PR for now - thanks for a quick feedback 😊

BlackbitDevs:
Hi @brusch, I updated the PR to have the same behaviour for many-to-many object relations as for many-to-one as @weisswurstkanone requested. 

I also extended the checks for `ObjectMetadata`, `ElementMetadata`, `HotspotImage` because these datatypes are also only valid in combination with elements. So they are consistent with many-to-many relations.

jdreesen:
I assume this was not intentionally closed, but because its target branch `10.1` was deleted.

dvesh3:
> I assume this was not intentionally closed, but because its target branch `10.1` was deleted.

@BlackbitNeueMedien Jacob is right ☝️ could you please create a new one for 10.2 branch? sorry for the inconvenience!

PR #9874, opened 2021-07-27
[Adminer] Establish SSL connection if SSL driver options are configured

### LINKED ISSUE #9872: [Bug]: Adminer not using SSL connection
### Expected behavior

Adminer should connect not only w/ configured host/port and credentials, but also using SSL (exclusively!), if configured.

### Actual behavior

Depending whether MySQL/MariaDB user is configured to only accept SSL connections `REQUIRE SSL`:

## Case "SSL optional"

Pimcore/Doctrine connects via SSL, but Adminer silently uses an unencrypted connection. This can be verified by query
```sql
SHOW STATUS LIKE 'Ssl_cipher';
```
, which shows none. This might be a security risk, as all data sent via Adminer is unencrypted while the developer assumes it's encrypted as well.

## Case "SSL required"

Auto-Login w/ Adminer fails (as no SSL cert is being used for connection) and enters a Login-Fail-Retry-Loop, until there's a ban `Too many unsuccessful logins, try again in %d minute(s).` using a ban-file `/tmp/adminer.invalid` ("global" BTW) written.

### Steps to reproduce

1) Configure Doctrine to use SSL:
```yaml
doctrine:
  dbal:
    connections:
      default:
        host: my-example-instance.eu-central-1.rds.amazonaws.com
        port: 3306
        user: pim
        password: '******'
        dbname: pim
        mapping_types:
          enum: string
          bit: boolean
        options:
          !php/const PDO::MYSQL_ATTR_SSL_CA: '%kernel.project_dir%/config/misc/eu-central-1-bundle.pem'
```
2) Verify with internal Doctrine connection and `SHOW STATUS LIKE 'Ssl_cipher';` that SSL is being used.
3) Open Adminer from within Pimcore.
4) Result:
    1) "SSL optional": Query `SHOW STATUS LIKE 'Ssl_cipher';` shows that connection is unencrypted.
    2) "SSL required": Login-Fail-Retry-Loop resulting in a ban `/tmp/adminer.invalid`, so no connection at all.

# Possible solution

* Make use of Adminer-Login `\AdminerLoginSsl($ssl)` and set connection options for `MYSQL_ATTR_SSL_*`.
* If there won't be fix: Warn about unencrypted Adminer connection or deny Adminer connection at all.

podarcis:
This PR uses configured SSL-related driver options (SSL_KEY, SSL_CERT and SSL_CA) for the Adminer connection by using the Adminer plugin `\AdminerLoginSsl`.

Resolves #9872

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/External/AdminerController.php
+++ b/bundles/AdminBundle/Controller/Admin/External/AdminerController.php
@@ -200,6 +200,17 @@ namespace {
                 new \AdminerDumpAlter,
             ];
 
+            // support for SSL (at least for PDO)
+            $driverOptions = \Pimcore\Db::get()->getParams()['driverOptions'] ?? [];
+            $ssl = [
+                'key' => $driverOptions[\PDO::MYSQL_ATTR_SSL_KEY] ?? null,
+                'cert' => $driverOptions[\PDO::MYSQL_ATTR_SSL_CERT] ?? null,
+                'ca' => $driverOptions[\PDO::MYSQL_ATTR_SSL_CA] ?? null,
+            ];
+            if ($ssl['key'] !== null || $ssl['cert'] !== null || $ssl['ca'] !== null) {
+                $plugins[] = new \AdminerLoginSsl($ssl);
+            }
+
             class AdminerPimcore extends \AdminerPlugin
             {
                 /**


dpfaffenbauer:
💪

brusch:
LGTM, thanks a lot!

PR #9876, opened 2021-07-27
Force id to be string to show the currently active Roles for a User

StefanGaffga:
## Bugfix: Active Roles of existing Users not shown

When editing Users, the Roles they are in are not shown as selected. This is due to the fact, that the ExtJS
MultiSelect Box uses Strings to determine which entries to select, but the ArrayStore contains integers as the id. 

## How to Reproduce

Create some Roles and a User. Assign some Roles to this User. Save it. Close the User Tab!
Reopen the User-Tab: The current Roles are not shown selected in the List, but they should be.


dvesh3:
@StefanGaffga already fixed on 10.0.5 https://github.com/pimcore/pimcore/pull/9484?

StefanGaffga:
@dvesh3 Oh yes, indeed it is... I just realized I did not search ALL Pull Requests but only the open ones since the filter for open is predefined... I did not pay attention to that,,, my bad.

PR #9881, opened 2021-07-28
convertToWebPath can return null

HaaseIT:
This method can return null, which will result in a TypeError:

> Pimcore\Model\Asset\Video\ImageThumbnail::convertToWebPath(): Return value must be of type string, null returned

This commit fixes the type error.

### DIFF
--- a/models/Asset/Thumbnail/ImageThumbnailTrait.php
+++ b/models/Asset/Thumbnail/ImageThumbnailTrait.php
@@ -283,7 +283,7 @@ trait ImageThumbnailTrait
      *
      * @param array $pathReference
      *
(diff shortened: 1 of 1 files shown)


models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -285,7 +285,7 @@ public function getFileExtension()
      *
      * @return string
      */
-    protected function convertToWebPath(array $pathReference): string
+    protected function convertToWebPath(array $pathReference): ?string
```
dvesh3, line 288: please update phpdoc as well. thanks!

PR #9884, opened 2021-07-28
[Data Objects] Password field: added optional minimum password length

brusch:


bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/password.js
```diff
@@ -174,6 +174,12 @@ pimcore.object.classes.data.password = Class.create(pimcore.object.classes.data.
                 hideLabel: true,
                 value: t('width_explanation')
             },
+            {
+                xtype: "numberfield",
+                fieldLabel: t("min_length"),
+                name: "minimumLength",
+                value: this.datax.minimumLength
```
weisswurstkanone, line 182: Nice. Could you please add this to the config (just to avoid to get into the negative range)

minValue: 0

weisswurstkanone (approved):

Issue #9889, opened 2021-07-28
[Bug]: Deadlock when updating elastic search store


passioneight:
### Expected behavior

Executing `bin/console ecommerce:indexservice:process-queue update-index --timeout=5` does not cause deadlock.

### Actual behavior

Executing `bin/console ecommerce:indexservice:process-queue update-index --timeout=5` results in deadlock. 

```
[2021-07-28 07:39:03] console.ERROR: Error thrown while running command "ecommerce:indexservice:process-queue update-index --timeout=5".

Message: "An exception occurred while executing 'UPDATE ecommerceframework_productindex_store_elastic SET crc_index = crc_current WHERE o_id = ? and tenant = ?' <OMITTED> Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction"
```

### Steps to reproduce

Reproducing is going to be tricky. However, the curlpit is https://github.com/pimcore/pimcore/blob/43b0963dd9ff7d0faff39db339fd0c15ad324a50/bundles/EcommerceFrameworkBundle/IndexService/Worker/AbstractMockupCacheWorker.php#L94

as it is not transactional.

Use the following code instead:
```php
$this->executeTransactionalQuery(function () use ($data, $objectId) {
    $this->db->query('UPDATE ' . $this->getStoreTableName() . ' SET crc_index = crc_current WHERE o_id = ? and tenant = ?', [$objectId, $this->name]);
});
```

brusch:
Fixed by #9892

PR #9892, opened 2021-07-28
Deadlock when updating elastic search store

passioneight:
- Fixes deadlock issue
- see [Bug]: Deadlock when updating elastic search store #9889 for the description


brusch:
Thank you very much! 👍

PR #9893, opened 2021-07-28
Check if array is empty before imploding

passioneight:
If  `$filter['value']` is an empty array (`[]`)n the `$conditionPartsFilters` will contain an invalid condition, causing an SQL exception.

Same check as for the following line -  just a different use case: 

https://github.com/pimcore/pimcore/blob/4888053a44f8847866ee1a5822a07626d99e3c1b/bundles/AdminBundle/Helper/GridHelperService.php#L316


dvesh3:
@passioneight thanks!

PR #9902, opened 2021-07-29
Trigger fallback language

adriaans:
A fallback language for a parent language is never triggered because `$lookForFallback = empty($translated)` always returns false in function `Pimcore\Translation\Translator::checkForEmptyTranslation`. 

This is because the value for argument `$translated` in `checkForEmptyTranslation` always is equal to the `$id` if no translation is found, see these functions:
- function Pimcore\Translation\Translator::trans
- function Symfony\Component\Translation::get

Trigger the fallback language if `$normalizedId` and `$translated` are equal (this assumes no translation was found).

### DIFF
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -283,7 +283,6 @@ class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleA
             $translated = $this->translator->trans($normalizedId, $parameters, $domain, $locale);
         }
 
-        $lookForFallback = empty($translated);
         if ($normalizedId != $translated && $translated) {
             return $translated;
         } elseif ($normalizedId == $translated) {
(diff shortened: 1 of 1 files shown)


dvesh3:
@adriaans I'm afraid with your changes, it will fallback when translation key & value are same e.g. `key: "Artikel" en: "Article" de: "Artikel"`, in this case it will always fallback to English, which is wrong.  As per the fix https://github.com/pimcore/pimcore/pull/8189/files#diff-348ef94471940979eacfda402822c86f98fc9fdf4abc5c08afb74211d3898898L238-L241 message catalogue should be initialized with empty values and then `$lookForFallback` check should work. 

For some strange reason, the code is still there https://github.com/pimcore/pimcore/blob/701b4bac06f472321e71959f7f3c8b9386d4503b/lib/Translation/Translator.php#L219

Removing it should do the trick. could you please update your PR with this change? thanks!

dvesh3:
@adriaans 🏓

adriaans:
@dvesh3 Looks like your suggestion isn't doing the trick either. I will investigate further and report back here.

dvesh3:
@adriaans any update on this? thanks!

dvesh3:
Closed due to inactivity. Please feel free to discuss further!

PR #9903, opened 2021-07-29
Fix type error

Blackbam:
Returned the ID as string instead of integer although the annotation says integer.


weisswurstkanone:
@Blackbam Where does this exactly pop up? Is there a test scenario? Many thanks!

brusch:
Don't think something pops up here, the type of the return value is just not consistent with the PhpDoc 😉
@Blackbam Thanks a lot!

PR #9904, opened 2021-07-29
[Bug]: Objectbricks ManyToOneRelation throws warning if no relations are present. #9900

### LINKED ISSUE #9900: Objectbricks ManyToOneRelation throws warning if no relations are present.
### Expected behavior

Opening an object in the backend with an object brick that has a many-to-one relation without a relation should open.

### Actual behavior

When the object is opened and there is no relation present, the following error is presented.
`Warning: Undefined array key 0`

Happens cause relations can be empty but is not checked for empty on.
vendor/pimcore/pimcore/models/DataObject/ClassDefinition/Data/Objectbricks.php:251

### Steps to reproduce

Open an object in the backend that has an object brick with a many-to-one relation field that has no value.

weisswurstkanone:
Resolves #9900

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Objectbricks.php
+++ b/models/DataObject/ClassDefinition/Data/Objectbricks.php
@@ -247,7 +247,7 @@ class Objectbricks extends Data implements CustomResourcePersistingInterface, Ty
             }
             $data = [];
 
-            if ($fielddefinition instanceof ManyToOneRelation && count($relations) > 0) {
+            if ($fielddefinition instanceof ManyToOneRelation && isset($relations[0])) {
                 $data = $relations[0];
             } else {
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/Objectbricks.php
```diff
@@ -247,7 +247,7 @@ private function getDataForField($item, $key, $fielddefinition, $level, $baseObj
             }
             $data = [];
 
-            if ($fielddefinition instanceof ManyToOneRelation) {
+            if ($fielddefinition instanceof ManyToOneRelation && count($relations) > 0) {
```
dvesh3:
```suggestion
            if ($fielddefinition instanceof ManyToOneRelation && isset($relations[0])) {
```

PR #9907, opened 2021-07-30
Bug disableWidthHeightAttributes fixed

Gerhard-Kanzler:
https://pimcore.com/docs/pimcore/current/Development_Documentation/Assets/Working_with_Thumbnails/Image_Thumbnails.html#page_Generating-HTML-for-Thumbnails
The disableWidthHeightAttributes has no effect, but should remove width="" and height="" on image tag


dvesh3:
@Gerhard-Kanzler thanks!

PR #9909, opened 2021-07-30
Display faulty countrycodes in ECF backoffice display

ThomasKeil:
If there's a country code stored in OnlineShopOrder/customerCountry that is not handled by https://github.com/umpirsky/country-list an exception is thrown on the Ecommerce Framework Back Office display page.

This PR fixes this.


dvesh3:
Hi @ThomasKeil, thanks for your PR. Could you please rebase it on branch 10.0 branch? since this is a fix. thanks!

brusch:
Pushed straight to `10.0` so that it is included in `10.0.9` release. 
@ThomasKeil Thanks a lot for the fix! 👍

PR #9910, opened 2021-07-30
[Bug]: Sending test emails are not handled as frontend requests

### LINKED ISSUE #9899: [Bug]: Sending test emails are not handled as frontend requests
### Expected behavior

Sending test emails with assets should have the asset urls prefixed by the value given in the config as mentioned in the [documentation](https://pimcore.com/docs/pimcore/current/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/File_Storage_Setup.html#page_File-Storage-Setup)

### Actual behavior

When sending a test email with assets. The urls of assets are not prefixed because the request is handled as an admin request. 

Possible solution is to set the `_pimcore_frontend_request` attribute in the `sendTestEmailAction` of the `Pimcore\Bundle\AdminBundle\Controller\Admin\EmailController` 

I can provide a PR if you think this is the right solution

### Steps to reproduce

Create an email that renders an asset url, send the test email and see that the asset url is not prefixed by the url set in the config

rubanooo:
Resolves #9899

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/EmailController.php
+++ b/bundles/AdminBundle/Controller/Admin/EmailController.php
@@ -16,6 +16,7 @@
 namespace Pimcore\Bundle\AdminBundle\Controller\Admin;
 
 use Pimcore\Bundle\AdminBundle\Controller\AdminController;
+use Pimcore\Http\RequestHelper;
 use Pimcore\Logger;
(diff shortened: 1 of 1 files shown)


dvesh3:
@rubanooo thanks!

PR #9912, opened 2021-07-30
10.x core constructor property promotion

Blackbam:
Constructor property promotion for the core bundle (including tiny comment and code style improvements). Should not change any functionality, just saves development time and works better with IDEs.

### DIFF
--- a/bundles/CoreBundle/EventListener/WorkflowManagementListener.php
+++ b/bundles/CoreBundle/EventListener/WorkflowManagementListener.php
@@ -34,7 +34,6 @@ use Pimcore\Workflow\Place;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\EventDispatcher\GenericEvent;
 use Symfony\Component\HttpFoundation\RequestStack;
-use Symfony\Component\Workflow\Registry;
 
 /**
  * @internal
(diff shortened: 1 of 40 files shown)


bundles/CoreBundle/EventListener/WorkflowManagementListener.php
```diff
@@ -46,39 +46,12 @@ class WorkflowManagementListener implements EventSubscriberInterface
...
 
-    /**
-     * @var Manager
-     */
-    private $workflowManager;
-
-    /**
-     * @var Registry
-     */
-    private $workflowRegistry;
```
dvesh3, line 57: $workflowRegistry missing in new contructor

Blackbam: Yes. It was removed because it is a private property which is never called and therefore I assumed it is an artifact which is not needed anymore. Is there any benefit in adding it?

dvesh3: My bad, I didn't check for its usage. It is fine then we can remove it. Please remove the use on line 37

Blackbam: @dvesh3 Removed (+ little type fix). Should fit now?

dvesh3: yes, thanks!

dvesh3 (approved):

brusch:
Thank you very much!

PR #9915, opened 2021-08-01
Add info about lower_case_table_names setting

boenner:
The linked [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_lower_case_table_names) explains the limitations when setting `lower_case_table_names` but using `debconf-set-selections` didn't work for me ([and some other folks](https://bugs.mysql.com/bug.php?id=90695)), so I added a workaround to the docs.


brusch:
Thanks! 👍

PR #9919, opened 2021-08-02
Adds individual class name to DataObject components

NiklasBr:
Adds the `name` field from DataObject class definitions as a HTML class to help targeting elements in the admin UI with e.g. CSS styles.

## Additional info  
Example:
image

image


brusch:
@NiklasBr Nice addition, but there are some data-types missing, right? (e.g. table, geobounds, ...) Can you please make this consistent for all data-types (at least to all that currently have the class `object_field object_field_type_XXXX`. 
Thanks!

NiklasBr:
Right, I'll do that!

dpfaffenbauer:
Can that be a bit more generalized? Every new data-type has to think about that....

NiklasBr:
@dpfaffenbauer I had that thought as well, but could not spend the time to refactor at that level

brusch:
That's a good point. Maybe by adding `getWrapperClassName("additional classes added at the end")` to the `pimcore.object.tags.abstract` ?

brusch:
Shouldn't be that much extra effort, right?

NiklasBr:
I can give it a shot now that I know that it is in `pimcore.object.tags.abstract` :)

dpfaffenbauer:
I mean even more... Like in a abstract/parent were everything is constructed to add the classes. I am not really sure right now if that is possible. but the less effort for new tags the better :)

brusch:
That would be nice of course, but introducing a method for building the class names is i guess a good starting point 😉

brusch:
With the method it looks already much cleaner 👍 
Just let us know when it's ready for review again 😊 Thanks

brusch:
Looks good to me, thanks a lot! 👍

PR #9921, opened 2021-08-02
[Bug] Moved downloading image to local file system after deferred check

rubanooo:

When generating deferred thumbnails, the asset was always downloaded to the `PIMCORE_SYSTEM_TEMP_DIRECTORY` because `$asset->getLocalPath()` was called before generating the deferred thumbnail path

The only reason I could find is that the file extension was needed to get the proper format. I have changed this to getting the extension from the asset filename, which should always be the same, right?

### DIFF
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -163,7 +163,7 @@ class Processor
 
         $fileExtension = $format;
         if ($format == 'original') {
-            $fileExtension = \Pimcore\File::getFileExtension($fileSystemPath);
+            $fileExtension = $fileExt;
         } elseif ($format === 'pjpeg' || $format === 'jpeg') {
             $fileExtension = 'jpg';
         }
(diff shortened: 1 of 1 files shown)


models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -171,7 +163,7 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
 
         $fileExtension = $format;
         if ($format == 'original') {
-            $fileExtension = \Pimcore\File::getFileExtension($fileSystemPath);
+            $fileExtension = $fileExt;
```
brusch, line 166: `$fileSystemPath` was used here as well (pushed that directly to your branch), but we can use directly `$fileExt` instead, so that's fine 😊

rubanooo: Missed that one 😲
Thanks!!

brusch:
Thanks a lot! 👍

PR #9932, opened 2021-08-03
Fix DocumentFallbackListener::$fallbackDocument must not be accessed before initialization

weisswurstkanone:
Partly reverted https://github.com/pimcore/pimcore/pull/9912 as it causes the following error on login:

image

@Blackbam Could you please recheck this whenever you find some time? Thx!

### DIFF
--- a/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
@@ -49,7 +49,7 @@ class DocumentFallbackListener implements EventSubscriberInterface
     /**
      * @var Document|null
      */
-    private ?Document $fallbackDocument;
+    private ?Document $fallbackDocument = null;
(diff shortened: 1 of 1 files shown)


bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
```diff
@@ -41,23 +41,48 @@ class DocumentFallbackListener implements EventSubscriberInterface
...
     /**
      * @var array
      */
-    protected array $options;
+    protected $options;
 
     /**
      * @var Document|null
      */
-    private ?Document $fallbackDocument;
```
dpfaffenbauer, line 52: `private ?Document $fallbackDocument = null;`

would solve that

Blackbam: Thanks!

Issue #5183, opened 2019-11-01
Date & time field wrong time in object edit field 


ascheider:
# Bug Report

### Expected behavior
User select date and time in pimcore admin and save object. Each user no matter from which timezone should see the same time.

### Actual behavior
User 1 from germany set time to 10 o'clock and save object. User from another timezone open the object and see different time which is for example 12 o'clock. So the time which is stored in db !== time which is shown in the object view. Pimcore should respect timezone settings of the pimcore installation.

### Steps to reproduce
1. Create Date & time field in classes.
2. Open the object and select date and time.
3. Install Chrome extension: https://chrome.google.com/webstore/detail/change-timezone-time-shif/nbofeaabhknfdcpoddmfckpokmncimpj
4. Select another timezone (not your timezone)
5. Reload pimcore backend an open the object (you will see another time, not the time you initial selected)

![Bildschirmfoto 2019-11-01 um 10 52 07](https://user-images.githubusercontent.com/6016485/68017478-dbdde080-fca7-11e9-948b-2d6bee3b9eba.png)
![Bildschirmfoto 2019-11-01 um 10 53 23](https://user-images.githubusercontent.com/6016485/68017487-e304ee80-fca7-11e9-9cf4-037852986920.png)
![Bildschirmfoto 2019-11-01 um 10 53 36](https://user-images.githubusercontent.com/6016485/68017496-e6987580-fca7-11e9-993d-97deb6384322.png)

fashxp:
Dear community, it's been a while since that issue was updated the last time, so we decided to close it for now. If you're still interested in contributing, feel free to add comments here or contact us before creating a pull request to discuss next steps.

Thanks for your understanding.

PR #9933, opened 2021-08-03
fix: shift timezone on date field

youwe-petervanderwal:
This PR fixes a bug same as https://github.com/pimcore/pimcore/issues/5183 but for the Date field instead of the Datetime field. I gave prio to the Date field as that one was even more obscure. I do want to build an improvement on the Datetime field as well though, based on a similar approach.

Possible BC-BREAK 1: If you have a custom field type extending \Pimcore\Model\DataObject\ClassDefinition\Data\Date, then the change in getDataFromEditmode / getDataFromGridEditor might cause issues. I've changed it to consequently use the normal timestamp. The getDataFromEditmode method was the only PHP method using the Javascript millisecond format

Possible BC-BREAK 2: If you have a custom editable type extending \Pimcore\Model\Document\Editable\Date, then the change in setDataFromEditmode might cause issues. I've changed it to consequently use a timestamp (trough the entire system).

--- 

Error: Depending on the timezone of the browser and the timezone of the
Pimcore server the stored date may be either correct, one day earlier or
one day later than the date entered by the user.

Caused by: The date is selected in the browser using the user's timezone
and sent to the server as timestamp. For example if the user is in CEST
and selects "2021-08-03" then the timestamp for "2021-08-03 00:00:00
CEST" is sent to the server. But if the server is configured to use the
New York timezone, it converts that timestamp to "2021-08-02 18:00:00
EDT" and thus stores "2021-08-02" (if DATE MySQL column is used).
Similar issues occurs for the TIMESTAMP MySQL column.

Resolved by: Shifting timezone on the entered date when receiving data
from or sending data to Pimcore.

### DIFF
--- a/models/Document/Editable/Date.php
+++ b/models/Document/Editable/Date.php
@@ -74,6 +74,10 @@ class Date extends Model\Document\Editable
      */
     public function frontend()
     {
+        if (!$this->date instanceof \DateTimeInterface) {
+            return '';
+        }
+
         $format = null;
 
         if (isset($this->config['outputFormat']) && $this->config['outputFormat']) {
@@ -81,12 +85,10 @@ class Date extends Model\Document\Editable
         } elseif (isset($this->config['format']) && $this->config['format']) {
             $format = $this->config['format'];
         } else {
-            $format = 'Y-m-d\TH:i:sO'; // ISO8601
+            return $this->date->toIso8601String();
         }
 
-        if ($this->date instanceof \DateTimeInterface) {
-            return $this->date->formatLocalized($format);
-        }
+        return $this->date->formatLocalized($format);
     }
 
     /**
@@ -118,9 +120,8 @@ class Date extends Model\Document\Editable
      */
     public function setDataFromEditmode($data)
     {
-        if (strlen($data) > 5) {
-            $timestamp = strtotime($data);
-            $this->setDateFromTimestamp($timestamp);
+        if (is_numeric($data)) {
+            $this->setDateFromTimestamp((int)$data);
         }
 
         return $this;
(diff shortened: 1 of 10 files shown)


models/Document/Editable/Date.php
```diff
@@ -74,19 +74,21 @@ public function getDataEditmode()
...
+        }
+
         $format = null;
 
         if (isset($this->config['outputFormat']) && $this->config['outputFormat']) {
             $format = $this->config['outputFormat'];
         } elseif (isset($this->config['format']) && $this->config['format']) {
             $format = $this->config['format'];
         } else {
-            $format = 'Y-m-d\TH:i:sO'; // ISO8601
```
youwe-petervanderwal, line 84: The old implementation displayed a literal `Y-m-d\TH:i:sO` in the browser, as the `formatLocalized` method expected the `%Y-%m...` notation.

youwe-petervanderwal:
Rebased my commit to the 10.x branch

youwe-petervanderwal:
Hi @brusch / @weisswurstkanone  any thoughts on the approach used in this PR? We're getting more and more clients (and QA testers) in different timezones so this is quite an annoying bug

models/Document/Editable/Date.php
```diff
@@ -74,19 +74,21 @@ public function getDataEditmode()
      */
     public function frontend()
     {
+        if (!$this->date instanceof \DateTimeInterface) {
```
jdreesen, line 77: Should we check for `Carbon` here? Both, `toIso8601String()` and `formatLocalized()` don't exist in the `DateTimeInterface` but only in `Carbon`.

wcchew76:
Since the team is fixing timezone issue, can I request the team to look into Object VERSION view for date / datetime data type value being rendered when user browser and server timezone is different?

youwe-petervanderwal:
Did some further testing, my current approach is causing problems with daylight saving time which might actually propagate within the 'same' timezone (CET vs CEST)... Closing this PR as it needs a rewrite with a better approach

PR #9935, opened 2021-08-03
Add support for condition variable types when using the Note Listing

samynw:

Previously the Listing class for Notes did not take the condition variable types into account when building the query. This made it impossible to perform some advanced queries on the notes, such as using the `IN(?)` operator with an array of strings.

## Additional info  

Example
```
$notes->setCondition("type IN (?)", [['erp_import', 'workflow]]);
```

Without the `conditionVariableTypes` the array of strings is not processed correctly when preparing the query.


dvesh3:
@samynw thanks!

PR #9946, opened 2021-08-05
Data loss for localized fields in field collection

### LINKED ISSUE #9941: Data loss for localized fields in field collection
### Expected behavior

Entered data for all languages in localized fields within field collection remain intact.

### Actual behavior

Data loss for all languages but the first visible language (first tab) on save when field collection is modified/dirty, e.g. new entry, entry deletion or resorting.

**Additional info:**
  * The problem seems to be in the frontend (ExtJs) as inspecting the request (XHR) shows that the other languages all have an empty value.
  * Affected/tested versions are at least: 10.0.7, 10.0.8 and branch 10.x

### Steps to reproduce

1. Configure multiple languages in the system, e.g. `en` and `de`
2. Create a field collection TestCollection
    - Add a localized field
    - Add an input field to the localized field
3. Create a data object class
    - Add a field collection field and allow previous created field collection. Use defaults for settings.
4. Create a data object of this class
    - Add items to the field collection and fill in all languages into the localized input
    - Save the data object
5. Refresh or re-open the data object **without touching the localized fields tabs**
    - Add a new item or delete an item or reorder items
    - Save the data object again
6. Refresh or re-open the data object again and verify that for each field collection item the languages, that weren't visible in the first language tab, are gone.

weisswurstkanone:
resolves #9941

related: https://github.com/pimcore/pimcore/pull/9674

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Fieldcollections.php
+++ b/models/DataObject/ClassDefinition/Data/Fieldcollections.php
@@ -183,7 +183,7 @@ class Fieldcollections extends Data implements CustomResourcePersistingInterface
                 $collectionKey = $collectionRaw['type'];
 
                 $oIndex = null;
-                if (!isset($params['objectFromVersion'])) {
+                if (!isset($params['objectFromVersion']) || $params['objectFromVersion'] == false) {
                     $oIndex = $collectionRaw['oIndex'] ?? null;
                 }
 
--- a/models/DataObject/ClassDefinition/Data/Fieldcollections.php
+++ b/models/DataObject/ClassDefinition/Data/Fieldcollections.php
@@ -182,10 +182,7 @@ class Fieldcollections extends Data implements CustomResourcePersistingInterface
                 $collectionData = [];
                 $collectionKey = $collectionRaw['type'];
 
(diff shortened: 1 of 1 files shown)


podarcis:
@weisswurstkanone Is the commit e865eea already meant to fix the issue? I just tested on branch `lc_inside_fc` and the problem still exists. Nevermind me if you're still working on that issue.

weisswurstkanone:
hi @podarcis in general yes. So the main identifed issue was that oIndex didn't get repespected anymore.

podarcis:
@weisswurstkanone In that case, as mentioned, it does not fix the issue #9941. The language data is already absent in the AJAX request (so it seems to be an issue in frontend context).

Do you want me to provide more information? Are my "steps to reproduce" missing something?

weisswurstkanone:
HI @podarcis language data being absent in the AJAX request shouldn't be a problem because by principle only dirty fields get submitted. 

Regarding https://github.com/pimcore/pimcore/issues/9941 ... the issue is far more easier to reproduce than described. Just open the object with a fieldcollection and localized field inside. Change the value of one language without touching the other language tabs. Save it.

PR #9968, opened 2021-08-06
Avoid var_dump-ing exceptions

jremmurd:
For long stack traces and nested previous exceptions the `var_dump()` of an exception object may fail with error `mmap() failed: [12] Cannot allocate memory`, which is especially fatal because the `logException()` method is generally called inside a ´catch` block.

This PR suggests to just print a string message including the stacktrace of the exceptions and not doing a `var_dump()`. 

What do you think about the idea? 
What about the suggested formatting, should we better use JSONs for the FileObjects format?

### DIFF
--- a/lib/Log/ApplicationLogger.php
+++ b/lib/Log/ApplicationLogger.php
@@ -399,17 +399,30 @@ class ApplicationLogger implements LoggerInterface
         ], $context));
     }
 
-    private static function createExceptionFileObject(\Throwable $exceptionObject)
+    private static function exceptionToString(\Throwable $exceptionObject, bool $includeStackTrace, bool $includePrevious = false): string
     {
-        //workaround to prevent "nesting level to deep" errors when used var_export()
-        ob_start();
-        var_dump($exceptionObject);
-        $dataDump = ob_get_clean();
+        $data = [
+            $exceptionObject->getMessage(),
+            "File: " . $exceptionObject->getFile(),
+            "Line: " . $exceptionObject->getLine(),
+            "Code: " . $exceptionObject->getCode(),
+        ];
 
-        if (!$dataDump) {
-            $dataDump = $exceptionObject->getMessage();
+        if ($includeStackTrace) {
+            $data[] = "Trace: " . $exceptionObject->getTraceAsString();
         }
 
-        return new FileObject($dataDump);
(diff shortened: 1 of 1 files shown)


brusch:
I think JSON is not such a good idea, because it will escape `/` and `\` which makes it hard to read. 
But yes, I agree, formatting the exception properly instead of just dumping the data makes totally sense. 
I just changed the formatting a bit (added `\n`), so that's a bit easier to read: 

```
Bar
File: /var/www/html/src/Controller/BernieController.php
Line: 57
Code: 300
Trace:
#0 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(157): App\Controller\BernieController->testAction(Object(Symfony\Component\HttpFoundation\Request))
#1 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(79): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#2 /var/www/html/vendor/symfony/http-kernel/Kernel.php(199): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#3 /var/www/html/public/index.php(36): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#4 {main}

Previous:
Foo
File: /var/www/html/src/Controller/BernieController.php
Line: 57
Code: 0
Trace:
#0 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(157): App\Controller\BernieController->testAction(Object(Symfony\Component\HttpFoundation\Request))
#1 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(79): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#2 /var/www/html/vendor/symfony/http-kernel/Kernel.php(199): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#3 /var/www/html/public/index.php(36): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#4 {main}
```

PR #9983, opened 2021-08-09
`bin/console list` throws error when symfony/maker-bundle is installed

jheimbach:
The maker-bundle throws an exception on running `bin/console list`:

```
In MakerCommand.php line 120:
                               
  Application cannot be null.  
```

Steps to reproduce:
 1. Install Pimcore 10.x (tested with 10.0.9 and 10.1.0)
 2. Add MakerBundle `composer require symfony/maker-bundle`
 3. Activate MakerBundle (add `\Symfony\Bundle\MakerBundle\MakerBundle::class => ['dev' => true]` to `config/bundles.php`)
 4. run `bin/console list`

I could fix this with the changes I made in this PR.

`$definition` is only needed when the command is an Instance of DoctrineCommand, therefore i think it is okay to load it after we checked the objecttype.

### DIFF
--- a/lib/Console/Application.php
+++ b/lib/Console/Application.php
@@ -104,9 +104,9 @@ final class Application extends \Symfony\Bundle\FrameworkBundle\Console\Applicat
 
     public function add(Command $command)
     {
-        $definition = $command->getDefinition();
-
         if ($command instanceof DoctrineCommand) {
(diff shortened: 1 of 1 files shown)


dvesh3:
@JHeimbach thanks a lot!

PR #9992, opened 2021-08-10
[Data Object] CSV export: prevent formula injection

weisswurstkanone:


brusch:
@weisswurstkanone I think this should be fixed for the other CSV exports as well (Assets, Redirects, ...).

brusch:
See also: https://csv.thephpleague.com/9.0/interoperability/escape-formula-injection/#prevents-csv-formula-injection

models/Element/Service.php
```diff
@@ -49,6 +51,12 @@
  */
 class Service extends Model\AbstractModel
 {
+
+    /**
+     * @var EscapeFormula
+     */
+    protected static $formatter = null;
```
brusch:
```suggestion
    private static ?EscapeFormula $formatter = null;
```

brusch (changes requested):

models/Element/Service.php
```diff
@@ -49,6 +51,12 @@
  */
 class Service extends Model\AbstractModel
 {
+
+    /**
+     * @var EscapeFormula
```
brusch:
```suggestion
     * @var EscapeFormula|null
```

PR #9995, opened 2021-08-11
[Data Objects] Exacter typehints for data object class getter / setter methods

BlackbitDevs:
Replaces hard-coded class FQNs with `::class`.
Replaces `array` by more specific type plus `[]` appendix.

If we need a migration which regenerated data object classes, jsut give me a note. But as no behaviour is changed, imho this is not necessary.


brusch:
Makes totally sense - thanks a lot! 👍

PR #10009, opened 2021-08-12
Removes references to getBricks() as it does not exist

NiklasBr:
## Additional info  

Not 100% sure this is correct, but it aligns better with what works in my environment at least :) 

`Call to undefined method getBricks in class Model\DataObject\Category`


brusch:
I think the docs are correct, but maybe the name for the bricks field (`bricks` so that the getter is named `getBricks()`) was not well decided 😊
So basically the class definition has a brick field called `bricks` and this brick field has a brick `tyre` assigned.

NiklasBr:
Aha! I'll see if I can clarify it better :)

brusch:
@NiklasBr 🏓Would be nice to have this out of the way - Thanks a lot! 😉

NiklasBr:
Not a lot of time this week to finish it, sorry

brusch:

PR #10014, opened 2021-08-12
Preventing unnecessary calls to transformation methods (especially fo…

roland4432:
Prevents image transformations to be called multiple times and do funny stuff to images.
Example: you have 2 transformations: 1 for rotation (from e.g. exif data) and 1 for highres manipulation.
The chain will apply the rotation and then the resizing for highres, then the _goto_ command will start from the beginning and then reapplying the rotation which will result in a wrong orientation.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+PIMCORE_PROJECT_ROOT="~/pimcore-rebase"
--- a/.env
+++ b/.env
@@ -1,1 +1,0 @@
-PIMCORE_PROJECT_ROOT="~/pimcore-rebase"
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -299,7 +299,7 @@ class Processor
             // sorry for the goto/label - but in this case it makes life really easier and the code more readable
             prepareTransformations:
 
-            foreach ($transformations as $transformation) {
+            foreach ($transformations as &$transformation) {
                 if (!empty($transformation) && !$transformation['isApplied'] ?? true) {
                     $arguments = [];
 
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -368,11 +368,11 @@ class Processor
                     ksort($arguments);
                     if (!is_string($transformation['method']) && is_callable($transformation['method'])) {
                         $transformation['method']($image);
-                        $transformation['isApplied'] = true;
                     } elseif (method_exists($image, $transformation['method'])) {
                         call_user_func_array([$image, $transformation['method']], $arguments);
-                        $transformation['isApplied'] = true;
                     }
+
+                    $transformation['isApplied'] = true;
                 }
             }
         }
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -306,14 +306,6 @@ class Processor
                     if (is_string($transformation['method'])) {
                         $mapping = self::$argumentMapping[$transformation['method']];
 
-                        if (in_array($transformation['method'], ['cropPercent'])) {
-                            //avoid double cropping in case of $highResFactor re-calculation (goto prepareTransformations)
-                            if ($imageCropped) {
-                                continue;
-                            }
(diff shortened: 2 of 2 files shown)


roland4432:
@brusch or @dvesh3 could one of you maybe give me a hint why these tests fail? I don't see any connections to my changes. Thank you in advance

brusch:
@roland4432 it is related to your changes. It seems that the thumbnail (for whatever reason) cannot be generated, that's why calling this method fails: 
https://github.com/pimcore/pimcore/blob/4888053a44f8847866ee1a5822a07626d99e3c1b/models/Asset/Thumbnail/ImageThumbnailTrait.php#L319

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -299,8 +299,8 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
             // sorry for the goto/label - but in this case it makes life really easier and the code more readable
             prepareTransformations:
 
-            foreach ($transformations as $transformation) {
-                if (!empty($transformation)) {
+            foreach ($transformations as &$transformation) {
+                if (!empty($transformation) && !$transformation['isApplied'] ?? true) {
```
brusch:
```suggestion
                if (!empty($transformation) && !$transformation['isApplied'] ?? false) {
```
I guess that's the issue

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -368,8 +368,10 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
                     ksort($arguments);
                     if (!is_string($transformation['method']) && is_callable($transformation['method'])) {
                         $transformation['method']($image);
+                        $transformation['isApplied'] = true;
                     } elseif (method_exists($image, $transformation['method'])) {
                         call_user_func_array([$image, $transformation['method']], $arguments);
+                        $transformation['isApplied'] = true;
                     }
```
brusch:
```suggestion
                        $transformation['method']($image);
                    } elseif (method_exists($image, $transformation['method'])) {
                        call_user_func_array([$image, $transformation['method']], $arguments);
                    }
                    
                    $transformation['isApplied'] = true;
```

brusch (changes requested):

brusch:
@roland4432 I guess I've found the issue, please review/test my suggested changes. Thanks!

roland4432:
Thx a lot. I will review it asap

models/Asset/Image/Thumbnail/Processor.php
roland4432: Hello @brusch,
just took a quick check on your changes. You r totally right with the 
```php
       if (!empty($transformation) && !$transformation['isApplied'] ?? false) {
```
part.

Regarding the 2nd change:
just wanted to make sure, that you noticed, that the _$transformation['isApplied'] = true;_ was placed intentionally inside the _if_ and _elseif_ part to only set the 'isApplied' value to _true_ if one of the 2 switches applied a transformation. 
If both (if and elseif) statements are false, you would still set _isApplied_ to true with your suggested change.

roland4432: I did some further testings. Seems like PHP 8 still doesn't like
```php
if(!$unsetKey['key'] ?? false) {...}
#but you have to go with the ordinary isset check
```

I'll update my PR and check if the tests run.

roland4432:
Everything should be fixed. WDYT about this?

.env
```diff
@@ -0,0 +1 @@
+PIMCORE_PROJECT_ROOT="~/pimcore-rebase"
```
brusch: I guess this was commited by mistake 😉

brusch (changes requested):
Looks good to me. 
Just 2 small things: 
- `.env` (see comment above) 😉
- I think this code is then obsolute with your change: https://github.com/pimcore/pimcore/blob/eb1140df9ba7f6bbc9cc8a34e0e4e6529cfa20fa/models/Asset/Image/Thumbnail/Processor.php#L309 can you please remove it? Thanks!

.env
roland4432: whoops..

brusch:
Thanks a lot! 👍

PR #10018, opened 2021-08-12
Fix translation export

Navino16:
When exporting Admin translations with keys which doesn't have any translations, an exception is raised.
Checking if key exist inside translation array (and if not return null) fix the problem.


Navino16:
Wrong branch

PR #10019, opened 2021-08-12
Fix admin translation export

Navino16:
When exporting Admin translations with keys which doesn't have any translations, an exception is raised.
Checking if key exist inside translation array (and if not return null) fix the problem.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/TranslationController.php
+++ b/bundles/AdminBundle/Controller/Admin/TranslationController.php
@@ -261,7 +261,7 @@ class TranslationController extends AdminController
         foreach ($translations as $t) {
             $tempRow = [];
             foreach ($columns as $key) {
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/TranslationController.php
```diff
@@ -261,7 +261,7 @@ public function exportAction(Request $request)
         foreach ($translations as $t) {
             $tempRow = [];
             foreach ($columns as $key) {
-                $value = $t[$key];
+                $value = array_key_exists($key, $t) ? $t[$key] : null;
```
blankse:
```suggestion
                $value = $t[$key] ?? null;
```

dvesh3:
@Navino16 thanks!

PR #10020, opened 2021-08-12
fstat is only for files, if we using real streams (not from file, lik…

tecbird:
…e s3 etc) fstat returns false

  
  

Resolves #


brusch:
@tecbird thanks a lot for your PR. However, it's based on the wrong branch, can you please rebase to `10.1` ? 
Thanks a lot!

tecbird:
mhh,

PR #10021, opened 2021-08-12
[Data objects] Trigger preObjectSave and postObjectSave JS events for data object folders

BlackbitDevs:
While for asset and document folders the JS events `preAssetSave` and `preDocumentSave` get triggered, this is not the case for data object folders. 

With this PR `preObjectSave` and `postObjectSave` get triggered when saving a data object folder.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/folder.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/folder.js
@@ -343,6 +343,22 @@ pimcore.object.folder = Class.create(pimcore.object.abstract, {
 
         this.tab.mask();
 
+        try {
+            pimcore.plugin.broker.fireEvent('preSaveObject', this, 'object');
+        } catch (e) {
+            if (e instanceof pimcore.error.ValidationException) {
+                this.tab.unmask();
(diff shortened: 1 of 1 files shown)


dvesh3:

bundles/AdminBundle/Resources/public/js/pimcore/object/folder.js
```diff
@@ -343,6 +343,22 @@ pimcore.object.folder = Class.create(pimcore.object.abstract, {
 
         this.tab.mask();
 
+        try {
+            pimcore.plugin.broker.fireEvent('preSaveObject', this, 'object');
+        } catch (e) {
+            if (e instanceof pimcore.error.ValidationException) {
+                this.tab.unmask();
```
weisswurstkanone, line 350: Just a short question. This unmasking ... shouldn't it be done for any kind of exception?

BlackbitDevs: I think yes, in this case I just copied it from https://github.com/pimcore/pimcore/blob/8b948ea0f182b250b7744122772b7f5f9f9677e6/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js#L749-L763

PR #10028, opened 2021-08-13
[ADMIN] Change the way, the admin index template is used

roland4432:
Change from hardcoded template rendering to the Symfony way of doing things, so you can easily override the admin template if needed

  
  

Remove hardcoded template reference to annotation


brusch:
This should be already the way Symfony is doing things ... not? 😊
Overriding should be possible, even when using `$this->render()`. 

See also: 
https://symfony.com/doc/current/controller.html#rendering-templates
https://symfony.com/doc/current/templates.html#creating-templates

Actually the `@Template` annotation isn't the Symfony standard way, it's provided by the SensioFrameworkExtraBundle: 
https://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/view.html

brusch:
Btw. not directly related, but still important: when overriding the admin template, please don't forget to **NOT** remove any of our copyrights (also the visible ones!).

roland4432:
Both of you r totally right.. the PR was just plain stupid.. sry about that.
I forgot about the possibility to override just the template, even tho I already used that a lot in the current project :/
I blame it on friday :(

PR #10031, opened 2021-08-13
Fixes malformed @ sign in SECURITY.md

NiklasBr:
Fixes malformed @ sign in SECURITY.md and links address with mailto


brusch:
This was actually intentional for spam reasons 😉

NiklasBr:
I don't believe that

PR #10038, opened 2021-08-13
use fstat result only if returned an valid result

tecbird:
Changed base: https://github.com/pimcore/pimcore/pull/10020


brusch:
Tests are failing because of #10049

PR #10054, opened 2021-08-17
[Assets] undefined index warning in asset xls export

weisswurstkanone:
Delimiter is not set for xls.


bundles/AdminBundle/Controller/Admin/Asset/AssetHelperController.php
```diff
@@ -671,7 +671,7 @@ public function doExportAction(Request $request, LocaleServiceInterface $localeS
         $ids = $request->get('ids');
         $settings = $request->get('settings');
         $settings = json_decode($settings, true);
-        $delimiter = $settings['delimiter'] ? $settings['delimiter'] : ';';
+        $delimiter = isset($settings['delimiter']) ? $settings['delimiter'] : ';';
```
brusch:
```suggestion
        $delimiter = $settings['delimiter'] ?? ';';
```

brusch (changes requested):

PR #10055, opened 2021-08-17
[Document Tree] - Correct check for variable localizedErrorDocumentsData

roland4432:
Corrected call to check wether or not the variable _localizedErrorDocumentsData_ is set. Else will run into errors and the _edit site_ menu won't open at all.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
@@ -1618,7 +1618,7 @@ pimcore.document.tree = Class.create({
                     fieldLabel: t("error_page") + " (" + availableLanguages[language] + ")",
                     name: "errorDocument.localized." + language,
                     fieldCls: "input_drop_target",
-                    value: localizedErrorDocumentsData && localizedErrorDocumentsData[language] ? localizedErrorDocumentsData[language] : '',
+                    value: (localizedErrorDocumentsData && localizedErrorDocumentsData[language]) ? localizedErrorDocumentsData[language] : '',
                     labelWidth: 200,
                     width: 500,
                     xtype: "textfield",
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
```diff
@@ -1618,7 +1618,7 @@ pimcore.document.tree = Class.create({
                     fieldLabel: t("error_page") + " (" + availableLanguages[language] + ")",
                     name: "errorDocument.localized." + language,
                     fieldCls: "input_drop_target",
-                    value: localizedErrorDocumentsData[language] ? localizedErrorDocumentsData[language] : '',
+                    value: localizedErrorDocumentsData && localizedErrorDocumentsData[language] ? localizedErrorDocumentsData[language] : '',
```
brusch:
```suggestion
                    value: (localizedErrorDocumentsData && localizedErrorDocumentsData[language]) ? localizedErrorDocumentsData[language] : '',
```
Brackets are missing, otherwise `value` would be mostly boolean, right?

brusch (changes requested):

bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
roland4432: Parentheses won't make much of a difference in a tenary expression, but from the looks.. I am fine with it :)

roland4432:
Parantheses won't make much of a difference in a tenery operation, but for the looks and readability. Let's put them :)

bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
brusch: Can you just quickly commit the suggestion? (I'm not able to do so - it's blocked by your settings 😉). Thanks!

brusch: Ah sorry, caching issue on GitHub 😉 All fine - thanks a lot!

roland4432: Very welcome

PR #10056, opened 2021-08-17
[Data object] Advanced relation disable options

kjkooistra-youwe:
Disable delete, add and create options for advanced many to many (object) relation field via field configuration. This is useful when the records are added programmatically and therefore shouldn't be added or removed, but the metadata should still be editable.


bundles/CoreBundle/Resources/translations/nl.extended.json
```diff
@@ -613,6 +613,9 @@
     "right": "right",
     "provide_split_view": "Provide Split View",
     "allow_multiple_assignments": "Allow Multiple Assignments",
+    "disable_delete": "Verwijderen uitschakelen",
+    "disable_add": "Toevoegen uitschakelen",
+    "disable_create": "Aanmaken uitschakelen",
```
youwe-petervanderwal: Do we need to add it to the other languages as well?

youwe-petervanderwal (approved):

bundles/CoreBundle/Resources/translations/nl.extended.json
kjkooistra-youwe: Looks like they automatically fall back to English.

brusch: Translations please only for EN, other languages are managed at POEditor.

brusch:
Thanks a lot for your PR! 
I'd like to leave this PR open a bit and discuss the actual need out there - just to make sure we're not adding more and more features that are very niche and are actually just making it harder to maintain the product. 

Just one more thing that comes to my mind ... if we decide to merge this PR, the settings should be consistent for all relational data-types - it seems some are actually missing, right?

kjkooistra-youwe:
@brusch It's specifically added for the **advanced** many to many relation fields, since they contain metadata fields which can be edited. There is no way to disable adding or deleting records without also disabling the metadata fields. For regular relation fields the 'not editable' option usually is sufficient. It can be expanded to those fields as well of course.

brusch:
Since there were no notable reactions on this and that there's a merge conflict, I'll close this PR for now. Please feel free to continue the discussion here 😊

kjkooistra-youwe:
@brusch Will it be merged if I fix the merge conflict? Or is a discussion mandatory? And if so, who should be involved?

kjkooistra-youwe:
I updated the branch, but I guess the issue won't update when closed.

kjkooistra-youwe:
@brusch Can you please answer my previous question?

brusch:
Since there were not notable demand for that feature, I'd leave it closed for now - just to keep things easier to maintain.

kjkooistra-youwe:
@brusch That's fine, but who should be involved for further discussion on whether the feature is useful or not?

brusch:
Normally people are upvoting or commenting on features, this a a good indicator for us how big the interest is 😊

kjkooistra-youwe:
I guess this may be a very specific use case.

PR #10061, opened 2021-08-17
[Encrypted field] Explain how to set up encryption key for field type "encrypted field"

BlackbitDevs:
When you first try to save a data object with an encrypted field, you will get an error message "could not load key". This is not very helpful. This PR changes it by giving instructions how to set up encrytion secret.

Actually it would be much more convenient if we could just set the config `pimcore.encryption.secret` on Pimcore installation. What do you think? And in which file? We could put it in `config/config.yaml` by reading and writing it with Symfony Yaml. WDYT?

### DIFF
--- a/models/DataObject/ClassDefinition/Data/EncryptedField.php
+++ b/models/DataObject/ClassDefinition/Data/EncryptedField.php
@@ -126,7 +126,7 @@ class EncryptedField extends Data implements ResourcePersistenceAwareInterface, 
             try {
                 $key = Key::loadFromAsciiSafeString($key);
             } catch (\Exception $e) {
-                throw new \Exception('could not load key');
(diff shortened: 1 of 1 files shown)


brusch:
Thanks a lot! 
I'd leave it as it is, the key should anyway live in an environment variable.

BlackbitDevs:
> the key should anyway live in an environment variable.

Really? This would mean that you need to set the environment variable to the same key when you copy the database to another server, otherwise you won't be able to decrypt the data. I just wonder how long it would take me to find this to be the reason when I copy the database to a dev server for example - or even worse to a local dev environment.

brusch:
Well that's a matter of taste, at least it shouldn't live in a VCS repo 😉

PR #10073, opened 2021-08-18
Update 02_Bricks.md

AlexTravix:
For auto-discover the "areas" must be written in lower-case


brusch:
Thanks a lot! I've pushed your changes directly to `10.1` branch (668b27daad18b2a098c064d8c487a2508bdb4e1d), as I would classify this as a bug in the docs 😊

PR #10074, opened 2021-08-18
[Installation] Check for mysql version >= 8

BlackbitDevs:
Currently during installation MySQL version does not get checked. While everything works fine during installation, later it causes fatal errors with constructs like https://github.com/pimcore/pimcore/blob/015a594d171e1deea7cd0f6c9a0f252371c335b1/models/DataObject/Concrete/Dao/InheritanceHelper.php#L282-L296

### DIFF
--- a/lib/Tool/Requirements.php
+++ b/lib/Tool/Requirements.php
@@ -77,7 +77,7 @@ final class Requirements
         // MySQL version
         $checks[] = new Check(
             [
-                'name' => 'InnoDB Support',
+                'name' => 'MySQL version >= 8.0',
                 'state' => version_compare($db->fetchOne('SELECT VERSION()'), 8, '>=')
             ]
         );
--- a/lib/Tool/Requirements.php
+++ b/lib/Tool/Requirements.php
@@ -74,6 +74,14 @@ final class Requirements
     {
         $checks = [];
 
+        // MySQL version
+        $checks[] = new Check(
+            [
+                'name' => 'InnoDB Support',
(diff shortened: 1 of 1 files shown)


brusch:
Thanks a lot!

brusch:
@BlackbitNeueMedien I had to revert this PR, because this causes issues with MariaDB.

BlackbitDevs:
Ok, so what do your propose? Simply trying to execute such a `WITH RECURSIVE` query and check if an exception gets thrown?

NiklasBr:
An idea:

```mysql
mysql> SHOW VARIABLES LIKE "version%";
+-------------------------+------------------------------+
| Variable_name           | Value                        |
+-------------------------+------------------------------+
| version                 | 8.0.26                       |
| version_comment         | MySQL Community Server - GPL |
| version_compile_machine | x86_64                       |
| version_compile_os      | Linux                        |
| version_compile_zlib    | 1.2.11                       |
+-------------------------+------------------------------+
5 rows in set (0.00 sec)
```

A combination of checking against `version_comment` and `version` could potentially work if MariaDB also could return something like that. I don't have it installed at the moment, so I cannot check for myself.

BlackbitDevs:
I created a new PR. I think it is better to check if the database really supports this "WITH RECURSIVE" feature instead of only checking the version. 
Nevertheless thanks @NiklasBr for your proposal!

@brusch You only reverted this PR's changes in branch `10.x`, not in `10.1`. But #10113 cares for that so that hopefully in the end there won't be any merge conflict.

PR #10079, opened 2021-08-19
[Data objects] Do not use admin mode when reloading object as admin user

BlackbitDevs:
Steps to reproduce bug:
1. Log in as admin user
1. Create class with input field `abc`, enable `Not editable`
2. Create object of this class
3. You see that field `abc` is not editable -> this is correct
4. Click `Reload` button
5. Field `abc` is editable -> this is wrong

The reason is that the object gets automatically reloaded in admin mode, although the default reload button was clicked. 
This PR fixes this behaviour:
- with default reload button readonly fields are not editable
- with reload split button item `Reload (Admin mode)` readonly fields are editable

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -531,9 +531,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
                 $objectData['validLayouts'] = [];
 
                 foreach ($validLayouts as $validLayout) {
-                    if ($validLayout->getId() != $currentLayoutId) {
-                        $objectData['validLayouts'][] = ['id' => $validLayout->getId(), 'name' => $validLayout->getName()];
-                    }
+                    $objectData['validLayouts'][] = ['id' => $validLayout->getId(), 'name' => $validLayout->getName()];
                 }
 
                 $user = Tool\Admin::getCurrentUser();
@@ -542,12 +540,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
                     $layout = DataObject\Service::getSuperLayoutDefinition($object);
                     $objectData['layout'] = $layout;
                 } elseif (!empty($currentLayoutId)) {
-                    // check if user has sufficient rights
-                    if (is_array($validLayouts) && isset($validLayouts[$currentLayoutId])) {
-                        $objectData['layout'] = $validLayouts[$currentLayoutId]->getLayoutDefinitions();
-                    } else {
-                        $currentLayoutId = 0;
-                    }
+                    $objectData['layout'] = $validLayouts[$currentLayoutId]->getLayoutDefinitions();
                 }
 
                 $objectData['currentLayoutId'] = $currentLayoutId;
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -492,7 +492,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
 
             $this->addAdminStyle($object, ElementAdminStyleEvent::CONTEXT_EDITOR, $objectData['general']);
 
-            $currentLayoutId = $request->get('layoutId', 0);
+            $currentLayoutId = $request->get('layoutId', null);
 
             $validLayouts = DataObject\Service::getValidLayouts($object);
 
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -523,11 +523,14 @@ class DataObjectController extends ElementControllerBase implements KernelContro
                     }
                 }
             }
+
             if (!empty($validLayouts)) {
-                $objectData['validLayouts'] = [ ];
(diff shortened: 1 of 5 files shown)


weisswurstkanone:
@BlackbitNeueMedien I would say this is not really objectbricks related.

Root cause is I believe a regression from https://github.com/pimcore/pimcore/pull/5286

here

https://github.com/pimcore/pimcore/blob/f4ece489e71797c0080b50a301d8d55099e865ab/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L499

It falls back to the "super layout" (which is completely editable then) if layout id with empty string is provided. Wdyt? 
Could you have a closer look? Would warmly appreciate it.

BlackbitDevs:
@weisswurstkanone Yes, this little object bricks change was only because I initially thought readonly fields would not work for object bricks. It is just a useless line of code I removed. Do you really want me to remove it from this PR?

You are right for the super layout fallback: The first available layout got used, this meant "admin mode" for admins. I changed the order of the reload options, now it also works without this JS change. WDYT @weisswurstkanone ?

weisswurstkanone:
@BlackbitNeueMedien The change looks good so far. But just thinking (and keep in mind that this is more about UI) ... by providing a invalid layoutId param it will still fall back to the super layout?

BlackbitDevs:
> by providing a invalid layoutId param it will still fall back to the super layout?

Actually I don't think so because it always uses the first valid layout in this case. And as I changed the order of the layout options to be first "Master" and only as second option "Admin mode" it should use "Master" when an invalid `layoutId` gets sent.

In short https://github.com/pimcore/pimcore/blob/f4ece489e71797c0080b50a301d8d55099e865ab/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L495-L511
tiogether with this PR's changes is the same as the following PHP code 
```php
$arr[0] = 'master';
$arr[-1] = 'admin mode';
var_dump(reset($arr)); // outputs 'master'
```

If you provide an empty `layoutId` (e.g. null, '' or 0), `$ok` gets set to `true` because of the type-tolerant comparison in https://github.com/pimcore/pimcore/blob/f4ece489e71797c0080b50a301d8d55099e865ab/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L503

If you provide an invalid `layoutId`, `$ok` will be false and then above logic gets used to get the first layout - which is `master` (without admin mode) then.

weisswurstkanone:
@BlackbitNeueMedien Looks good! 

Only one thing I noticed: This also affects the sort order in the backend UI. Maybe the super layout should be still top most. Wdyt?

Before:
image

After:
image

BlackbitDevs:
Ok. I wanted to hear if this is a problem first - but obviously it is ;-)  Will change it...

Newbie question: What is actually the difference of "Master" layout and just clicking the Reload icon of the split button? Perhaps we could remove "Master" completely if they are the same?

BlackbitDevs:
Have removed the "Master" entry from the split button as this is the same as the main reload button afaik. Prevents some UI confusion if there is a difference between main button and "Master". 
Bildschirmfoto 2021-08-25 um 11 54 23

WDYT?

weisswurstkanone:
Just for clarification:
- Master is the layout as defined in the class definition. Respects readonly and so on
- Master (Admin Mode) allows everything. Even the readonly fields. Only available for admins of course.

So for admins both should be there.

BlackbitDevs:
Ok, now I think it's good. Initially the object gets loaded with `layoutId` not being set -> gets null in https://github.com/pimcore/pimcore/blob/358e02afa612d0e8d8571db13f928a2eaad0fbe2/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L495

Then it is checked if the current layout id is a valid layout: https://github.com/pimcore/pimcore/blob/358e02afa612d0e8d8571db13f928a2eaad0fbe2/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L503
The problem here is that `null` == `0`. and thus `$currentLayoutId` being `null` was recognized to be ok ($ok was `true`).
This then led to the problem that when clicking the main reload button the current layout id was still null. And for this reason in https://github.com/pimcore/pimcore/blob/358e02afa612d0e8d8571db13f928a2eaad0fbe2/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L507-L511 the `$currentLayoutId` was set to -1 as this is the first layout for admin users.

I refactored it a bit, now it should be ok. Also order of admin layouts stays the same as before.

The only thing which I added beside the bugfix for the main button: In the split menu now only layouts different from the current one get shown because this causes UX problems as it supposes that the main button and the respective layout split button option were different things.

weisswurstkanone:
Regarding the split button. Not sure if this is a really good idea because then the user does not have any idea on which layout he is on. In addition, would happens if there is only one layout (the active one) for non-admin users ?

BlackbitDevs:
Ok, I reverted it so all layouts are shown. But is the current layout somehow highlighted or how to see which layout is currently loaded?

weisswurstkanone:
As shown in the screenshot above. Bold.
Why not just put the layout with id -1 on top as it was before. Should be feasible to reorder on js side,

BlackbitDevs:
> Why not just put the layout with id -1 on top as it was before.

It is now. Everything is the same as before. Except that now the initially loaded layout gets remembered whereas in current 10.1 it is `null`.

weisswurstkanone:
thanks a lot!

PR #10087, opened 2021-08-19
[Data Objects] Save default values to versions

BlackbitDevs:
With #5594 saving field's default value to database was introduced. The problem is that these default values do not get saved to versions currently. This is especially bad because when you publish this version, you will again have `null` in the respective fields.

Steps to reproduce:
1. Create class with input field `abc`, default value `test`
2. Create object of above class and save it
3. In the database tables now there is `test` in column `abc`
4. Go to versions tab, select jsut saved version -> there `abc` is empty
5. Restore the just saved version -> The value is now also `null` in the database column -> this is really bad because you can not restore the exact state the object had at this point (what is the main purpose of versioning beside tracking changes)

With this PR the default values are also saved to the database (and also to the PHP object being saved - which is important, if you have further operations with it after saving).


BlackbitDevs:
Are the failing tests caused by this PR? I don't think so but am not sure

brusch:
@BlackbitNeueMedien failing tests seem to be related, base branch runs through without any issues.

weisswurstkanone:
@BlackbitNeueMedien I guess a unit test would definitely make sense here. wdyt 😉

weisswurstkanone:
@BlackbitNeueMedien This needs to be done for fieldcollections, bricks & blocks as well.

BlackbitDevs:
Added test, added support for localized fields, object bricks, field collections.

BlackbitDevs:
@weisswurstkanone For the test I wanted to add an input field with a default value. How does this work? Obviously it is not enough to add it to the definition file in https://github.com/pimcore/pimcore/pull/10087/files#diff-07e12400f01e9028add55e187301705004f356fbc4a4c23ce29b2113a31d88e9R572

But CI complains about `[Exception] Call to undefined method getInputWithDefault in class Pimcore\Model\DataObject\Unittest`

dvesh3:
@BlackbitNeueMedien you have to add a datachild with name `inputWithDefault` (and call `setDefaultValue`) in `Unittest` class setup here https://github.com/pimcore/pimcore/blob/10.x/tests/_support/Helper/Model.php#L549

weisswurstkanone:
Thanks Divesh. 

@BlackbitNeueMedien just for clarification: The classes are generated by code. The json files are just generated for your convenience.

weisswurstkanone:
@BlackbitNeueMedien what is the status regarding the tests ? 🤔

BlackbitDevs:
The status is that I currently do not know why the tests fail. If you have some time to support, this would be great.

weisswurstkanone:
I am on it! Thx!

weisswurstkanone:
Hello @BlackbitNeueMedien , regarding the test. I believe it should be

```
        $data = [
            'lblockadvancedRelations' => new BlockElement('lblockadvancedRelations', 'advancedManyToManyRelation', $reference),
        ];
        $source->setLtestblock([$data], 'de');
        $source->save();

        //link source on target
        $target = TestHelper::createEmptyObject();
        $target->setHref($source);
        $target->save(); //block data should retain here

        //update block element - manyToManyRelations
        $referenceNew = TestHelper::createEmptyObject();
        $referenceNewId = $referenceNew->getId();
        $referenceNew = new BlockElement('lblockadvancedRelations', 'advancedManyToManyRelation', $referenceNew);
        $source->getLtestblock('de')[0]['lblockadvancedRelations']->setData($referenceNew);
        $source->save();
```

so lblockadvancedRelations instead of blockmanyToManyRelations.

BlackbitDevs:
@weisswurstkanone Fixed the problem with field collections but the test with the block ist still failing...

BlackbitDevs:
Finally it's working !!! 
The BlockTest was completely screwed:
- there was a Concrete added to a **Advanced**ManyToManyRelation
- only the object got added to the relation -> it actually must be an array of objects
- localized block and non-localized block were mixed up

Concerning the changes in https://github.com/pimcore/pimcore/pull/10087/files#diff-8ff96247c8a26d5a3ecbf9b79e38c31b8fc8a638bd145a46b9cdb7ee27d4869bR158 and https://github.com/pimcore/pimcore/pull/10087/files#diff-fdf5ea53db66a5ec71cdc16ba799cd7d068b27e01ca83ea4f5544ca1087df07aR985: Without these changes the actual ValidationException gets swallowed away - you see the result in https://github.com/pimcore/pimcore/runs/3554296716:
`[Pimcore\Model\Element\ValidationException] Validation failed:  fieldname=localizedfields ( fieldname=ltestblock[ localizedfields-de ]) `
You now have to guess which field caused the error and what the actual error is, you only know that it happened in one of the fields in the block element. With these changes the sub items of the ValidationException get set so that the exception cause gets clearer.

tests/Model/DataObject/ObjectTest.php
```diff
@@ -160,4 +160,19 @@ public function testCustomModificationDate()
         $object->save();
         $this->assertGreaterThanOrEqual($currentTime, $object->getModificationDate(), 'Expected auto assigned modification date');
     }
+
+    /**
+     * Verifies that an object can be saved with custom modification date.
+     *
```
jdevries14: empty space

BlackbitDevs: test description fixed.

weisswurstkanone:
@BlackbitNeueMedien , the change with the ValidationException is not really related or needed for this issue, right ?

BlackbitDevs:
> the change with the ValidationException is not really related or needed for this issue, right ?
Right. This was only the find the cause why the BlockTest failed. The original test was completely useless as mentioned and the cause was hidden as the ValidationException got swallowed away. Would be nice if you could review it together with this PR.

weisswurstkanone:
Hello @BlackbitNeueMedien , just tried the following situation:

image

"myinput" is mandatory and I left it empty.

Before the change there was an error message. After the change the error gets swallowed completely. In other words, even though mandatory, the object got saved.

Can we please split this up into two separate PRs because not directly related and second thing is that this needs to be tested for  other constellations as well.

BlackbitDevs:
@weisswurstkanone I reverted the changes where the ValidationExceptions get thrown. 

PS: Ah I see, I forgot to throw the Exception for localized fields. I will set up another PR...

PPS: Sorry, now the test is failing again. Had to redo the changes for throwing ValidationException. Will notify you when it works.

weisswurstkanone:
@BlackbitNeueMedien any updates on this one. Merci! 😄

BlackbitDevs:
@weisswurstkanone I have no idea why the test https://github.com/pimcore/pimcore/pull/10087/checks?check_run_id=3743086036 fails. I have set up the same in my local Pimcore:
Bildschirmfoto 2021-09-29 um 14 46 04

And then executed this script:
```php
$source = \Pimcore\Model\DataObject\Concrete::getById(4888);

$data = [
    'lblockadvancedRelations' => new \Pimcore\Model\DataObject\Data\BlockElement('lblockadvancedRelations', 'advancedManyToManyRelation', [new \Pimcore\Model\DataObject\Data\ElementMetadata('lblockadvancedRelations', [], \Pimcore\Model\DataObject\Concrete::getById(4889))]),
];
$source->setLtestblock([$data], 'de');
$source->save();
```

And this works perfect:
Bildschirmfoto 2021-09-29 um 14 47 50

Do you have an idea why the test fails?

weisswurstkanone:
@BlackbitNeueMedien Hello Jan, tests are fixed now. Reason was that "allowed classes" for the relation field are case sensitive. Since this field obviously was never used (at least not for an object as target) this was never noticed.

BlackbitDevs:
Big thanks @weisswurstkanone !!!
Will split up this PR into 2 tomorrow: one for saving default values to versions and one for the changes with ValidationException.

brusch:
@BlackbitNeueMedien I'll close this PR since you're separating them anyway, right 😊 Looking forward to the 2 PRs.

BlackbitDevs:
@brusch No. Please reopen it. I just remove the ValidationException things from this PR here and put those changes into another one.

brusch:
@BlackbitNeueMedien sure, no problem 😊

BlackbitDevs:
Ok, I reverted all changes related to the ValidationExceptions. This PR is now ready for review

BlackbitDevs:
Sorry, @weisswurstkanone's fix got missing because I am too stupid to use git ;-)
Added it again.

brusch:
@BlackbitNeueMedien I wouldn't introduce this change in a bug-fix release, because someone could rely on the current behavior. I think it would be ok to do this in a minor release, but before we gonna merge this PR we should discuss this as well. 

In the meantime, could you please rebase this PR onto `10.x`. Thanks a lot!

BlackbitDevs:
Rebased to `10.x`.
Actually it really is a bug because you cannot restore a certain state by publishing a version like I described in https://github.com/pimcore/pimcore/pull/10087#issue-974488622 - but I do not have any problem that this gets released with the next minor release.

brusch:
@BlackbitNeueMedien one more minor thing, we should mention the behavior change in the upgrade notes. Thanks!

BlackbitDevs:
@brusch I added documentation in the upgrade notes.

weisswurstkanone:
thanks a lot!

PR #10091, opened 2021-08-19
Memorized asset issue

weisswurstkanone:
image

Scenario:
- get the asset tab get memorized in the local storage 
- delete the asset by using a different browser
- reload the backend ui

There will be an endless round of requests because
- the item does not exist
- success is false but the response flag never gets respected
- item type is different so it will start the next cycle

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js
@@ -32,7 +32,7 @@ pimcore.asset.asset = Class.create(pimcore.element.abstract, {
         try {
             this.data = Ext.decode(response.responseText);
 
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js
```diff
@@ -32,6 +32,10 @@ pimcore.asset.asset = Class.create(pimcore.element.abstract, {
         try {
             this.data = Ext.decode(response.responseText);
 
+            if (!this.data.success && this.options && this.options.ignoreNotFoundError) {
```
dvesh3: this breaks memorized asset tab in general as `success = true` is not passed when image is found
image
```suggestion
            if (this.data.success === false && this.options && this.options.ignoreNotFoundError) {
```

PR #10095, opened 2021-08-19
[Workflows] Add configuration validation for state_machine workflows

fashxp:
adds validation for configuration for state_machine workflows that ensures that only one parallel place is allowed. 

discuss if this is a bugfix or a new feature

### DIFF
--- a/bundles/CoreBundle/DependencyInjection/Configuration.php
+++ b/bundles/CoreBundle/DependencyInjection/Configuration.php
@@ -1790,6 +1790,32 @@ final class Configuration implements ConfigurationInterface
                                 })
                                 ->thenInvalid('"supports" or "support_strategy" should be configured.')
                             ->end()
+                            ->validate()
+                                ->ifTrue(function($v) {
+
+                                    if(($v['type'] ?? 'workflow') === 'state_machine') {
+
+                                        foreach ($v['transitions'] ?? [] as $transition) {
+
+                                            if(count($transition['to']) > 1) {
+                                                return true;
+                                            }
+
+                                        }
+
+                                        foreach ($v['globalActions'] ?? [] as $transition) {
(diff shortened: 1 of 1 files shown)


bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -1790,6 +1790,32 @@ private function addWorkflowNode(ArrayNodeDefinition $rootNode)
...
+                                            if(count($transition['to']) > 1) {
+                                                return true;
+                                            }
+
+                                        }
+                                    }
+
+                                    return false;
+                                })
+                                ->thenInvalid('Type `state_machine` does not support multiple `to` definitions for transitions and global actions. Change definition or type to `workflow`.')
```
brusch, line 1817:
```suggestion
                                ->then(function ($v) {
                                    trigger_deprecation('pimcore/pimcore', '10.2', 'Type `state_machine` does not support multiple `to` definitions for transitions and global actions. Change definition or type to `workflow`. This will trigger an error in Pimcore 11', __METHOD__);
                                })
```

I'd be careful with this change, since it could break existing deployments.
What do you think about my suggestion?

brusch (changes requested):

bundles/CoreBundle/DependencyInjection/Configuration.php
fashxp: I'm with you ... but on the other hand, state_machines with multiple `to` definitions also don't work properly yet. 
But I'm also fine with deprecation warning now and a invalid error with Pimcore 11

brusch: Triggering an error is fine as it was validated already before -> regression.

PR #10098, opened 2021-08-20
Update Readme - Stable is X already

dpfaffenbauer:
Same version on both links


brusch:
Thanks for the hint! 
Actually the demo.pimcore.fun was using the wrong branch - now it's fixed and I've changed the hostname of dev, to be more clear.

dpfaffenbauer:
alright too :)

PR #10103, opened 2021-08-20
Fix dashboard illegal chars

roland4432:
Resolves issue with illegal charset when creating dashboards. Illegal chars lead to non-functional dashboards (you cannot open them anymore).

=> **I consider this as draft, the solution is simple, but might confuse users since chars recently entered might disappear**

## Reproduce error
Create a dashboard with an identifier containing special chars like ?&%$, etc. or spaces.


dvesh3:
It looks good to me 👍

PR #10107, opened 2021-08-20
Fix: Width issue with multiselect field

### LINKED ISSUE #10089: Width issue with multiselect field
### Expected behavior

Width of class definition should be respected.

### Actual behavior

Width setting gets ignored.

### Steps to reproduce

- Create new class with select and multiselect field
- Set width to something arbitrary on both of the fields
- Create a new object
- Width gets respected for the select but not for the multiselect

Codewise it looks fine but maybe a 3rd party issue?

Maybe related: https://github.com/pimcore/pimcore/pull/6899

image

image

blankse:
Resolve #10089


weisswurstkanone:
Looks good. Thanks a lot!

PR #10112, opened 2021-08-20
Make Version20210323082921 more resilient to NULL values

NiklasBr:

Our `website-settings.php` had a lot of `null` values on languages, this helps alleviate fatal errors which would otherwise happen during migration.


brusch:
Thanks a lot!

PR #10113, opened 2021-08-20
[System requirements / Installation] Check for support for "WITH RECURSIVE"

BlackbitDevs:
Follow-up of #10074

Check if database supports SQL construct "WITH RECURSIVE" like in https://github.com/pimcore/pimcore/blob/015a594d171e1deea7cd0f6c9a0f252371c335b1/models/DataObject/Concrete/Dao/InheritanceHelper.php#L282-L296


BlackbitDevs:
But this time the failing test is not caused by this PR ;-)

brusch:
Thanks a lot! 👍

PR #10116, opened 2021-08-23
Url slug route fix, resolves #10005

### LINKED ISSUE #10005: [Bug]: URL Slugs that partially matches a Document path gives 404
### Expected behavior

Any published Data Object with a URL Slug should be reachable

### Actual behavior

Any time a part of the URL Slug matches a full path of an unpublished Document page, the URL Slug will return a 404.

### Steps to reproduce

1. Create a URL Slug on a data object, .e.g. `/a/b/c/123`
2. Create a Document page structure like with a Pretty URL like `/a/b/c`
3. Save, and unpublish the page `c`
4. Visit  `/a/b/c/123` on the front end.

Result: 404

dturopoli:
resolves [#10005](https://github.com/pimcore/pimcore/issues/10005)

### DIFF
--- a/lib/Routing/Dynamic/DataObjectRouteHandler.php
+++ b/lib/Routing/Dynamic/DataObjectRouteHandler.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 
 namespace Pimcore\Routing\Dynamic;
 
+use Pimcore\Bundle\CoreBundle\EventListener\Frontend\ElementListener;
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
I do agree with that change, but I see it as a BC break. Older installations might expected that unpublished elements are not "routed".

brusch:
That could be an issue, yes ... not sure if this is really relevant but anyway.

brusch:
Put #10005 on `11.0` milestone.

PR #10126, opened 2021-08-23
[Toolbar] Added toolbar menu entry ids

fashxp:
to ease adding additional menu items to sub menus

for example
```js
            const userMenu = settingsMenu.items.getByKey('pimcore_menu_settings_users');
            if(!userMenu) {
                return;
            }

            const someSubMenu = Ext.create('Ext.menu.Item', {
                // ...
            });

            userMenu.menu.add(someSubMenu);
```


brusch:
@fashxp not all items got an `itemId` , maybe it would make sense to consistently add it to all of them, wdyt?

PR #10129, opened 2021-08-23
[DataObject] Convert empty string values to nullable values

MyZik:
Hi!

This small PR fixes the bug with empty string values in Pimcore.

Actually, all input fields values will be saved in a database as empty string `""`. By developing of a REST API, that will be problematic because we cannot provide the good quality of data from data objects :)

## Additional info  
-

### DIFF
--- a/lib/Twig/Extension/CacheExtension.php
+++ b/lib/Twig/Extension/CacheExtension.php
@@ -109,7 +109,7 @@ class CacheExtension extends AbstractExtension
      */
     public function end()
     {
-        if ($this->captureEnabled[$this->key]) {
+        if ($this->captureEnabled[$this->key] ?? false) {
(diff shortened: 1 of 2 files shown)


dvesh3:
@MyZik I'm not really in favor of touching this generic method to fix input fields data, since it is being used for all datatypes. (This is the reason tests are failing).

MyZik:
@dvesh3 Hmm. I've tried to find the place, where the object data setters is called, there is only once place, AbstractModel...

dvesh3:
@MyZik I think of Input data type implementing `PreSetDataInterface` and handing this in `preSetData` method.

Please note this is a breaking change as from a semantic point of view an empty string is not the same as a null value.

PR #10143, opened 2021-08-25
[ExtJS] Translate UI labels

BlackbitDevs:
Currently for example grid filter labels like "Before", "After", "On" for date filters are not translated to the current user's language.
This PR only translates the grid column context menu item "Filters". It would be better if we could use ExtJS's language files (see https://docs.sencha.com/extjs/7.0.0/guides/core_concepts/localization.html) to translate all of ExtJS's UI labels - but neither did I found those files to download not did I get it working with the overrides (I tried to implement it in `bundles/AdminBundle/Resources/public/extjs/js/bootstrap-ext-all.js`). Perhaps someone can give me a hint, otherwise now at least the "filter" context menu label gets translated...

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/listfolder.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/listfolder.js
@@ -259,7 +259,13 @@ pimcore.asset.listfolder = Class.create(pimcore.asset.helpers.gridTabAbstract, {
             columns : gridColumns,
             enableLocking: true,
             bufferedRenderer: false,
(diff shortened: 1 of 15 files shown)


brusch:
I think this would need to be integrated into the build of ExtJS which is described here: 
https://github.com/pimcore/pimcore/blob/e8ea66e09032ea465e6438450dc1a9695101cc01/bundles/AdminBundle/Resources/public/js/lib/PIMCORE_MODIFICATIONS.md#ext-js

In the meantime, this PR can be also done with a single line of code 😉#10479

PR #10144, opened 2021-08-25
[Admin] Documents Link - allow assets dragndrop

### LINKED ISSUE #8446: Unable to drag`n`drop assets on a navigation link
# Bug Report
In Pimcore 6 it`s not possible to drag n drop an asset into a navigation link.
Our costumers are very confused about it. I double checked it in fewer versions of pimcore it was possible. In some cases this is very useful, to link the terms (a pdf) for example. 

### Expected behavior
It would be nice if I could drag n drop the pdf directly into the navigation link.

### Actual behavior
At the moment its not possible to drag an drop, I have to copy the path manually.

### Steps to reproduce
Open a Pimcore 6.X, set a navigation point (type link) and try to drag n drop a pdf from your assets folder.

dvesh3:
Obselete


brusch:
@dvesh3 Objects should be also supported, but aren't at the moment. I'd also classify this as a bug (rebase on `10.1`). Thanks!

Issue #7627, opened 2020-11-30
[DataObject][Inheritance] Quantity Value vs Simple Objects


zsala:
# Bug Report

We have the following setup. Parent child product relation for a Pimcore DataObject. For the description I will use the `Cars` data object from Pimcore demo website. In order to be able to test this bug we need to add a new value to the `Cars` data object. We will compare the behaviour of an existing field `Country` with the newly added field of type Quantity Value.

P.S In my project we only use this setup but the bug is also relevant to variant data object setup. 

### Expected behaviour
If I empty the value of the `Country` field and `save` OR `save & publish` my changes, the child product will get the value of the parent. I would expect the same for a class field of quantity value too.

### Actual behaviour
In case of our custom quantity value (price), if I empty the value, the parent value is taken and inherited only if I do `save & publish`. The `save` functionality does not overwrite the child's empty value!

### Possible cause
Every data object generates a php class that has predefined get / set functions. In this function inheritance checkbox is checked. Code example follows:

`
         if(\Pimcore\Model\DataObject::doGetInheritedValues() && $this->getClass()->getFieldDefinition("myCustomQuantityValuesName")- 
         >isEmpty($data)) {
		try {
			return $this->getValueFromParent("myCustomQuantityValuesName");
		} catch (InheritanceParentNotFoundException $e) {
			// no data from parent available, continue ...
		}
	}
`

As seen in the if statement the parent value is only taken if the childs value is empty. This works fine unless this custom field is an object. If it's an object the objects value is never empty. The values of the objects could be empty, but the check does not take care of that case, if I understand the code correctly.

MarcoLeko:
I think the actual problem is actually in the DataObjectController line ```671```:
There we have a check:
 ``` if ($fielddefinition->isEmpty($fieldData) && !empty($parent) ... ```
Where we try to call the function ```getDataForField``` recursively. But since a QuantityValue fieldcollection can never reach this condition ```$fielddefinition->isEmpty($fieldData)``` it will set the data from the parent. One fix would be to extend the check with: 
```
            if ($fielddefinition->isEmpty($fieldData) && !empty($parent) ||
                ($fieldData instanceof DataObject\Data\QuantityValue && $fielddefinition->isEmpty((string)$fieldData)))
            {
                    ...
            }
```

fashxp:
fixed via #10147?

PR #10147, opened 2021-08-25
[DataObject] Added isEmpty to QuanitiyValue data types

fashxp:
The thing is, that loading a data object with an empty quantity value data attributes results in different data depending if it is loaded from database or if it is loaded from a version. 

loaded from database --> null
loaded from version --> quantity value object with null values in it

this is because when saving a data object (from pimcore backend) with an empty quality value attribute, the data object has a quantity value object with null values in it ... and this is then serialized when saving the version. 
when loading from database, there is a check for null values and only if values are not null, a quantity value object is returned. 

this results in for example in https://github.com/pimcore/pimcore/issues/7627 --> inheritance does not work properly for unpublished data objects with quantity value attributes. 

this PR fixes is by adding `isEmpty` methods that consider empty quantiy value objects
not sure, if we can fix it already during saving time without major refactorings. 

there also might other data types to be checked?

### DIFF
--- a/models/DataObject/ClassDefinition/Data/InputQuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/InputQuantityValue.php
@@ -189,5 +189,4 @@ class InputQuantityValue extends QuantityValue
 
         return null;
     }
-
 }
--- a/models/DataObject/ClassDefinition/Data/InputQuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/InputQuantityValue.php
@@ -189,4 +189,16 @@ class InputQuantityValue extends QuantityValue
 
         return null;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isEmpty($data)
+    {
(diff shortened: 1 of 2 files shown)


models/DataObject/ClassDefinition/Data/InputQuantityValue.php
```diff
@@ -189,4 +189,16 @@ public function denormalize($value, $params = [])
 
         return null;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isEmpty($data)
```
blankse: InputQuantityValue extends QuantityValue. So this method should not be necessary.

weisswurstkanone:
Apart from @blankse 's comment it looks good to me.

weisswurstkanone:
@fashxp 🏓

models/DataObject/ClassDefinition/Data/InputQuantityValue.php
fashxp: done

weisswurstkanone (approved):

PR #10150, opened 2021-08-25
Fix PhpDoc: getByField methods can return null

blankse:
`$test = DataObject\Test::getByTest('Test', null, 1)` can return null.
PhpDoc is wrong and PhpStan not happy if I use it in an if :)

### DIFF
--- a/models/DataObject/ClassDefinition.php
+++ b/models/DataObject/ClassDefinition.php
@@ -509,7 +509,7 @@ final class ClassDefinition extends Model\AbstractModel
                             $this->getName()
                         ).'\Listing|\\Pimcore\\Model\\DataObject\\'.ucfirst(
                             $this->getName()
-                        ).' getBy'.ucfirst(
+                        ).'|null getBy'.ucfirst(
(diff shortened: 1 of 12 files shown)


models/DataObject/ClassDefinition.php
```diff
@@ -509,7 +509,7 @@ public function generateClassFiles($generateDefinitionFile = true)
                             $this->getName()
                         ).'\Listing|\\Pimcore\\Model\\DataObject\\'.ucfirst(
                             $this->getName()
-                        ).' getBy'.ucfirst(
+                        ).'|null getBy'.ucfirst(
```
dvesh3, line 512: Migration to regenerate class definitions?

dvesh3: 🏓 @NiklasBr

blankse: @dvesh3 I added a migration

dvesh3:
@blankse thanks a lot!

PR #10154, opened 2021-08-25
[Data Objects] Advanced relations must not be dirty directly after loading

BlackbitDevs:
Steps to reproduce bug:
1. Create class `ABC` with advanced many-to-many object relation named `relation`, allowed class is the class itself (for simplicity)
2. Create object `a` in `/`, save and publish
3. Create object `b` in `/`, assign `a` in field `relation`, save and publish
4. Create the following script:
```php
$object = \Pimcore\Model\DataObject\ABC::getByPath('/b');
$object->getRelation();
var_dump($object->isFieldDirty('relation'));
```
This will result in `true`

The reason is that when loading the advanced relation, the setter methods get used: https://github.com/pimcore/pimcore/blob/7efe8331d012641a6aca0e500bd280e0363d01bf/models/DataObject/Data/ObjectMetadata/Dao.php#L96

And those setter methods trigger `$this->markMeDirty()`, see for example https://github.com/pimcore/pimcore/blob/7efe8331d012641a6aca0e500bd280e0363d01bf/models/DataObject/Data/ObjectMetadata.php#L157
But of course the `ObjectMetadata` objects have to set the field dirty when they get called from the application after loading, so they are correct there.

This PR resets the dirty state after loading the data from the Dao.

### DIFF
--- a/tests/Model/LazyLoading/AdvancedManyToManyObjectRelationTest.php
+++ b/tests/Model/LazyLoading/AdvancedManyToManyObjectRelationTest.php
@@ -98,7 +98,7 @@ class AdvancedManyToManyObjectRelationTest extends AbstractLazyLoadingTest
         $object = LazyLoading::getByPath('/lazy1');
         $this->assertFalse($object->isFieldDirty('advancedObjects'), 'Advanced relation must not be dirty directly after loading');
 
-        $object->setAdvancedObjects()[0]->setMetadataUpper('some-other-metadata');
+        $object->getAdvancedObjects()[0]->setMetadataUpper('some-other-metadata');
         $this->assertTrue($object->isFieldDirty('advancedObjects'), 'Advanced relation must be dirty after changing a metadata field');
     }
 
--- a/tests/Model/LazyLoading/AdvancedManyToManyObjectRelationTest.php
+++ b/tests/Model/LazyLoading/AdvancedManyToManyObjectRelationTest.php
@@ -55,6 +55,7 @@ class AdvancedManyToManyObjectRelationTest extends AbstractLazyLoadingTest
         $object = $this->createDataObject();
         $object->setAdvancedObjects($this->loadMetadataRelations('advancedObjects', 'metadataUpper'));
         $object->save();
+
         $parentId = $object->getId();
         $childId = $this->createChildDataObject($object)->getId();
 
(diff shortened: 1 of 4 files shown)


dvesh3 (approved):

weisswurstkanone:
@BlackbitNeueMedien thanks a lot, looks good so far! Can we cover this scenario by extending an existing unit test. Should be a oneliner I guess. Wdyt?

weisswurstkanone (approved):

tests/Model/LazyLoading/AdvancedManyToManyObjectRelationTest.php
```diff
@@ -83,6 +84,24 @@ public function testClassAttributes()
...
 
+    public function testDirty()
+    {
+        $object = $this->createDataObject();
+
+        $relatedObjects = $this->loadMetadataRelations('advancedObjects', 'metadataUpper');
+
+        $object->setAdvancedObjects($relatedObjects);
+        $object->save();
+        $this->assertFalse($object->isDirty('advancedObjects'), 'Advanced relation must not be dirty after saving');
```
weisswurstkanone: maybe isFieldDirty instead ?

BlackbitDevs: Thanks, fixed this.
Do you know where the remaining problems come from, @weisswurstkanone ?
```
 ------ ---------------------------------------------------------------- 
  Line   bundles/AdminBundle/Controller/Admin/TranslationController.php  
 ------ ---------------------------------------------------------------- 
  1085   Left side of && is always true.                                 
 ------ ---------------------------------------------------------------- 

  Line   bundles/AdminBundle/Controller/Searchadmin/SearchController.php      
  525    Comparison operation ">=" between int<0, max> and 0 is always true.  

Error: ] Found 2 errors                                                         

Error: Left side of && is always true.
Error: Comparison operation ">=" between int<0, max> and 0 is always true.
Error: Process completed with exit code 1.
```

I tried to merge the base branch but with no success.

dvesh3: > Thanks, fixed this.
> Do you know where the remaining problems come from, @weisswurstkanone ?
> 
> ```
>  ------ ---------------------------------------------------------------- 
>   Line   bundles/AdminBundle/Controller/Admin/TranslationController.php  
>  ------ ---------------------------------------------------------------- 
>   1085   Left side of && is always true.                                 
>  ------ ---------------------------------------------------------------- 
> 
>  ------ --------------------------------------------------------------------- 
>   Line   bundles/AdminBundle/Controller/Searchadmin/SearchController.php      
>  ------ --------------------------------------------------------------------- 
>   525    Comparison operation ">=" between int<0, max> and 0 is always true.  
>  ------ --------------------------------------------------------------------- 
> 
> Error: ] Found 2 errors                                                         
> 
> Error: Left side of && is always true.
> Error: Comparison operation ">=" between int<0, max> and 0 is always true.
> Error: Process completed with exit code 1.
> ```
> 
> I tried to merge the base branch but with no success.

These issues are fixed on base branch 10.1. please rebase your PR on 10.1. thanks!

BlackbitDevs: Thanks @dvesh3 - I rebased 10.1 and now tests are successful.

weisswurstkanone:

PR #10156, opened 2021-08-25
[BUG] Admin Translations - Fix search domain for queries and handle $t == null

roland4432:
Pass $domain to the model to query in the correct table and add handle to care for $t == null since this is a valid return value

## Reproduce
Open admin translations. Search for any string
=> error 500 => fix with _if($t = ...) { ... }_
After this query again and get no results even if the search string is present in the translation keys
=> pass $domain to _getByKey_ Method to search in the right table


dvesh3:
Regression of #10013

dvesh3:
@roland4432 thanks!

PR #10163, opened 2021-08-26
Url slug route fix, Resolves #10005

### LINKED ISSUE #10005: [Bug]: URL Slugs that partially matches a Document path gives 404

dturopoli:
Resolves #10005, but with option to switch behavior off and on via config

### DIFF
--- a/bundles/CoreBundle/DependencyInjection/Configuration.php
+++ b/bundles/CoreBundle/DependencyInjection/Configuration.php
@@ -687,6 +687,15 @@ final class Configuration implements ConfigurationInterface
                                 ->end()
                             ->end()
                         ->end()
+                        ->booleanNode('allow_processing_unpublished_elements_for_url_slug')
+                            ->beforeNormalization()
+                            ->ifString()
+                                ->then(function ($v) {
+                                    return (bool)$v;
+                                })
+                                ->end()
+                            ->defaultFalse()
+                        ->end()
                     ->end();
         $classDefinitionsNode = $objectsNode
             ->children()
--- a/bundles/CoreBundle/DependencyInjection/Configuration.php
+++ b/bundles/CoreBundle/DependencyInjection/Configuration.php
@@ -687,20 +687,6 @@ final class Configuration implements ConfigurationInterface
                                 ->end()
                             ->end()
                         ->end()
-                        ->booleanNode('allow_processing_unpublished_elements_for_url_slug')
-                            ->beforeNormalization()
-                            ->ifString()
-                                ->then(function ($v) {
-                                    return (bool)$v;
-                                })
-                                ->end()
-                            ->defaultFalse()
-                            ->setDeprecated(
-                                'pimcore/pimcore',
-                                '10.1',
-                                'The "%node%" option is deprecated since Pimcore 10.1, it will be removed in Pimcore 11.'
-                            )
-                        ->end()
                     ->end();
         $classDefinitionsNode = $objectsNode
             ->children()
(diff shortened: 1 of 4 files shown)


brusch:
@EqualString @dturopoli Unfortunately adding an annotation is not enough. We also need to call `trigger_deprecation()` in case you're accessing an object slug but it's getting blocked by the underlying document. So that needs to be done somewhere in `\Pimcore\Bundle\CoreBundle\EventListener\Frontend\ElementListener` i guess. 
Thanks again!

lib/Routing/Dynamic/DataObjectRouteHandler.php
```diff
@@ -114,6 +123,17 @@ private function buildRouteForFromSlug(DataObject\Data\UrlSlug $slug, DataObject
         $route->setDefault('object', $object);
         $route->setDefault('urlSlug', $slug);
 
+        trigger_deprecation(
```
jdreesen: I'd say the deprecation should not be triggered if the value is the one that is not deprecated. I.e. it should only be triggered if the value is `false`. Otherwise, you can't fix the deprecation in your code prior to updating to the version where the behavior changes.

brusch: This is the wrong place anyway, that needs to be done in ElementListener as mentioned before. 
Here should be the right place: 
https://github.com/pimcore/pimcore/blob/e37d843d43299593391baa3d24e46cd442cf82ec/bundles/CoreBundle/EventListener/Frontend/ElementListener.php#L96-L95

Something like the following: 
```php
                if($request->get('object') && $request->get('urlSlug')) {
                    trigger_deprecation(...);
                }
```

brusch (changes requested):
Additionally I think we'd need to do that for Static Routes as well.

bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -687,6 +687,20 @@ private function addObjectsNode(ArrayNodeDefinition $rootNode)
                                 ->end()
                             ->end()
                         ->end()
+                        ->booleanNode('allow_processing_unpublished_elements_for_url_slug')
```
brusch: This is now the wrong location for the config, since it also covers static routes. 
I'd propose the following location/naming: `pimcore.routing.allow_processing_unpublished_fallback_document`

bundles/CoreBundle/EventListener/Frontend/ElementListener.php
```diff
@@ -94,6 +96,20 @@ public function onKernelController(ControllerEvent $event)
...
+                    (
+                        ($request->get('object') && $request->get('urlSlug')) ||
+                        $request->get('pimcore_request_source') == 'staticroute'
+                    ) &&
+                    !$this->config['objects']['allow_processing_unpublished_elements_for_url_slug']
+                ) {
+                    trigger_deprecation(
+                        'pimcore/pimcore',
+                        '10.2',
+                        '`allow_processing_unpublished_elements_for_url_slug` is deprecated, in Pimcore 11 it will always be `true`'
```
brusch:
```suggestion
                        'Blocking routes where the underlying fallback document is unpublished is deprecated and will be removed in Pimcore 11. If you rely on this behavior please change your controllers accordingly and set the config option `pimcore.routing.allow_processing_unpublished_fallback_document=true`'
```

brusch (changes requested):
We are almost there, just two more small changes 😉

brusch:
Thanks a lot! 👍

PR #10181, opened 2021-08-27
Update `.gitattributes` to save space and trees

jdreesen:
Resolves #10177

## Additional info  

Test this locally:

```bash
git checkout 10.1
git archive --format zip HEAD > archive-10.1.zip
ls -la archive-10.1.zip # 63MB
git checkout update-gitattributes
git archive --format zip HEAD > archive-update-gitattributes.zip
ls -la archive-update-gitattributes.zip # 29MB
```

The package downloaded by Composer every time Pimcore is required is 63MB. After applying this PR, the size decreases by 54% to 29MB.

By using `export-ignore` in a `.gitattributes` file, it is possible to dramatically reduce the file size of a Composer package download.

When Composer downloads a package, a zip/tar file is created using the `git archive` command. This behavior can be changed, but by default, Composer downloads an archive of the specified package. When using GitHub, GitLab, or BitBucket, Composer can directly download the zip file, and all those services will imitate a `git archive` command output to generate the downloadable archive.

The `git archive` command looks up for files matching an `export-ignore` rule, and if matched, it will exclude those files from the archive.

With `export-ignore` rules in the `.gitattributes` file, Composer packages can opt to exclude build files and test files (such as GitHub CI configuration file, tests and configuration files, documentation, etc.) when the package is downloaded by Composer.

All these excluded files will remain in the repository when it is cloned/forked, and are otherwise available to those who download the package source files; The `export-ignore` rule will exclude them in the zip file downloads, thus reducing the download size and time when the packages are consumed.

Link: https://php.watch/articles/composer-gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,7 +3,7 @@
 
 bundles/AdminBundle/Resources/public/js/lib/* linguist-vendored
 
-# Exclude build/test files from archive
+# Exclude build/test files from archive to reduce ZIP size for composer dist download
 /.github export-ignore
 /doc export-ignore
 /tests export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,3 +2,17 @@
 * -text
 
 bundles/AdminBundle/Resources/public/js/lib/* linguist-vendored
+
+# Exclude build/test files from archive
+/.github export-ignore
+/doc export-ignore
+/tests export-ignore
+/.editorconfig export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.markdownlint.json export-ignore
+/.php-cs-fixer.dist.php export-ignore
+/CLA.md export-ignore
+/CONTRIBUTING.md export-ignore
+/codeception.dist.yml export-ignore
+/phpstan* export-ignore


.gitattributes
```diff
@@ -2,3 +2,17 @@
 * -text
 
 bundles/AdminBundle/Resources/public/js/lib/* linguist-vendored
+
+# Exclude build/test files from archive
```
brusch:
```suggestion
# Exclude build/test files from archive to reduce ZIP size for composer dist download
```

Fahl-Design:
FYI: @brusch @jdreesen this PR breaks https://pimcore.com/docs/pimcore/current/Development_Documentation/Development_Tools_and_Details/Testing/Application_Testing.html#page_Codeception because you remove the hole tests/_support folder and not just that big chunk of example data

brusch:
@Fahl-Design just set `pimcore/pimcore` to `prefer-source` in your projects `composer.json`.

Fahl-Design:
@brusch sounds good to me, thanks
you may need to add this to the docs as well

jdreesen:
You may need to delete your `vendor/pimcore` folder before updating to make Composer actually download the source.

ITspirit:
Well, that should be added definitely to the docs with such a change, not only in an issue comment ;-)

But the solution "preferred-install" or "preferred-source" won´t work this way with satis. Maybe satis it´s exotic, but i don´t think so. We rely on many projects on satis repositories to speed up development and deployment.

Would be great it there could be a more suitable solution than just "downsizing" the archive, maybe an separate "test-repo"?

Could be working now with adding prefered-install to the project composer.json, but this definitely is skipping satis and needs connection to github for every pipeline run.

Correct me if i am wrong.

brusch:
@ITspirit @Fahl-Design regarding docs - agree, can you please provide a PR for that? Thanks in advance! 😊

Just wanted to add that other frameworks are doing that as well: 
https://github.com/symfony/symfony/blob/5.4/.gitattributes#L2

Please feel free to continue the discussion here. If it's really a common need, we'll sure find a solution 😊

ITspirit:
a PR shoudn´t be a big deal. But first a suitable solution for some test and CI/DD Specialists should be found. ;-)

Wouldn´t compare Pimcore with Symfony, but ok. Outsourcing of the test Section would be a nice option.

- Split the "/tests" folder out in a own project "pimcore/tests" for example
- Adding these project as a "Submodule" to the pimcore 
- If needed only an additional "git submodule init" and "git submodule update" should be fine (and/or git pull in these folder).

Would keep the pimcore source Dist small, and on the other hand with a separate repository it´s still usable with SATIS and other private Package Registries.

What do you think?

Fahl-Design:
[Off Topic] @ITspirit we had the same issue with SATIS some years ago, maybe migrate to https://github.com/repman-io/repman  (self hosted, + composer proxy, + composer v2 support, + source checkout works and  `prefer-source` works as expected, oauth gitlab/github integration, and based on symfony with async message support)

jdreesen:
> Outsourcing of the test Section would be a nice option.

Yes, but why include it as a git submodule instead of making it a composer package that can be installed if desired?

PR #10191, opened 2021-08-30
[Admin] Preview version - disable output escaping only for html preview

dvesh3:


dvesh3:
already fixed https://github.com/pimcore/pimcore/commit/6321e8296bed7c0eefd4fbcfa3f90bffb54966b4

PR #10192, opened 2021-08-30
Handle errors from preSave doing sanity checks for getOptions

### LINKED ISSUE #9991: [Bug]: Classes-rebuild fails for class dependent optionprovider
### Expected behavior

The console command pimcore:deployment:classes-rebuild should work under any circumstances.

### Actual behavior

The command fails if the order of the classes in the database is defined in an order that will break the rebuild command.

When running the classes-rebuild command, the build mechanism for the files under var/classes/DataObject will trigger the method _presave_, which will then call the _getOptions_ from the optionprovider and fail under certain cirumstances.

### Steps to reproduce

* Create a class 'A' with a select and set an optionprovider for its values (and create the optionprovider)
* Create a class 'B'
* The method _getOptions_ in the optionprovider will instance the listing from class 'B'
* Delete var/classes/DataObject folder for a clean rebuild (to simulate a deployment e.g. on Azure, where the artifact is always created without existing DataObjects or Listings)
* Call console command pimcore:deployment:classes-rebuild -c
* If the database table _classes_ is set to order A before B, then classes-rebuild will fail with the message: 'Cannot load class Listing in OptionProvider.php (the one created under the first step)

roland4432:
in case an error occurs, we don't have options, so therefore no sanity check (finding commas in values) is needed

  
  

Resolves #9991

## Additional info  
The method _preSave_ in the Multiselect only has 1 responsibility => check for commas in the value fields. Therefore, whenever an error occurs (e.g. when creating classes and their listings) there is no need to check for commas afterwards since we don't have values (for any reason). For a clean approach the _$options_ is set to null to not trigger the following _if_ statement

So a simple try catch does the trick here.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Multiselect.php
+++ b/models/DataObject/ClassDefinition/Data/Multiselect.php
@@ -707,7 +707,13 @@ class Multiselect extends Data implements
             $context = [];
             $context['fieldname'] = $this->getName();
 
-            $options = $optionsProvider->getOptions($context, $this);
+            try{
+                $options = $optionsProvider->getOptions($context, $this);
+            }
+            catch (\Exception $e){
+                // error from getOptions => no values => no comma => no problems
+                $options = null;
+            }
         } else {
             $options = $this->getOptions();
         }
--- a/bundles/AdminBundle/Controller/Admin/TranslationController.php
+++ b/bundles/AdminBundle/Controller/Admin/TranslationController.php
@@ -1080,10 +1080,10 @@ class TranslationController extends AdminController
                     libxml_clear_errors();
                     $html = $doc->saveHTML();
 
-                    $bodyStart = strpos($html, '<body>') + 6;
+                    $bodyStart = strpos($html, '<body>');
                     $bodyEnd = strpos($html, '</body>');
                     if ($bodyStart && $bodyEnd) {
-                        $html = substr($html, $bodyStart, $bodyEnd - $bodyStart);
+                        $html = substr($html, $bodyStart + 6, $bodyEnd - $bodyStart);
                     }
 
                     $output .= $html;
--- a/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
+++ b/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
@@ -524,7 +524,7 @@ class SearchController extends AdminController
 
         for ($i = $count; $i >= 0; $i--) {
             $shortPath = '/' . implode('/', array_unique($parts));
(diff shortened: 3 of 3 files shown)


roland4432:
Did the checks change? I did not alter any of the 2 given classes failing in the static analysis? oO

dvesh3:
@roland4432 please rebase on 10.1 to fix phpstan failed tests. thanks!

roland4432:
Hello @dvesh3,

the branch is forked from 10.1, a rebase onto this branch does not show any changes :/

dvesh3:
@roland4432 let me re-run the jobs and let's see if it fixes the issue.

dvesh3:
@roland4432 no luck 😞 please fetch and try rebasing again.

roland4432:
Made a fetch, rebased, but I don't know why this is implying file changes for the other 2 files the checks failed for.
But as long as it works??? I am fine with it :D
WDYT @dvesh3 ?

models/DataObject/ClassDefinition/Data/Multiselect.php
```diff
@@ -707,7 +707,13 @@ public function preSave($containerDefinition, $params = [])
             $context = [];
             $context['fieldname'] = $this->getName();
 
-            $options = $optionsProvider->getOptions($context, $this);
+            try{
+                $options = $optionsProvider->getOptions($context, $this);
```
dvesh3, line 711: or this should be handled in the Option provider implementation? since there's no reference of exception on interface https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/ClassDefinition/DynamicOptionsProvider/MultiSelectOptionsProviderInterface.php#L22-L28

dvesh3:
> Made a fetch, rebased, but I don't know why this is implying file changes for the other 2 files the checks failed for.
> But as long as it works??? I am fine with it :D
> WDYT @dvesh3 ?

we can rebase or merge 😉

models/DataObject/ClassDefinition/Data/Multiselect.php
roland4432: :thinking: 
not sure. The point is to prevent the method _getOptions_ to bubble errors to the calling code. Since it might fail with the scenario given in #9991. Therefore an implementation of error handling in many places would always be discouraged by a handling in a central place IMHO.
Also the error is not thrown to higher callings, it's just about not validating multiselect data if there is no data.

But I am always open to improvements. How would you try to tackle this situation?

dvesh3:
@roland4432 thanks!

PR #10201, opened 2021-08-31
[Fixed]Error:"Getting Array to string conversion" on installation in Windows platform

pdchaudhary:
Related to PR: #9725

Steps to reproduce:
1. Run "COMPOSER_MEMORY_LIMIT=-1 composer create-project pimcore/skeleton my-project" cmd

Error:
Script Pimcore\Composer::clearCache handling the pimcore-scripts event terminated with an exception
[ErrorException]
Array to string conversion

### DIFF
--- a/lib/Composer.php
+++ b/lib/Composer.php
@@ -171,6 +171,10 @@ class Composer
         $command = array_merge($command, $cmd);
 
         //$event->getIO()->write('Run command: ' . implode(' ', $command), false);
+        
+        if (substr(php_uname(), 0, 7) == "Windows") {
+            $command = implode(' ', $command);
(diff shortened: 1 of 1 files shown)


pdchaudhary:
@dvesh3 done

brusch:
Don't think that is right, as mentioned already in #9725 - the code of Symfony/Process is clearly not accepting strings: 
https://github.com/symfony/symfony/blob/5.3/src/Symfony/Component/Process/Process.php#L141

If it's a general issue, we definitely need a different solution.

Shradhanjalisahu (commented):
Where i changes this code.

PR #10224, opened 2021-09-02
Fix Multiselect getVersionPreview array_map argument order

GALCF:
Bug reproducable by adding a Multiselect field to any dataobject, filling this field and then searching globally (admin mode) for this object. The error looks something like:

```
array_map(): Argument #1 ($callback) must be a valid callback, array must have exactly two members
```


jdreesen (approved):

dvesh3:
@GALCF thank you!

PR #10242, opened 2021-09-06
[Data objects] Add getter method for reverse relation field type

BlackbitDevs:
Follow-up of #4262

This time it implements only the getter method.


brusch:
@BlackbitNeueMedien can you please check the failing tests? Thanks a lot!

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -176,11 +177,43 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
...
+        }, $relations);
+
+        // using PHP sorting to order the relations, because "ORDER BY index ASC" in the queries above will cause a
+        // filesort in MySQL which is extremely slow especially when there are millions of relations in the database
+        usort($relations, static function ($a, $b) {
+            if ($a['index'] == $b['index']) {
+                return 0;
+            }
+
+            return ($a['index'] < $b['index']) ? -1 : 1;
```
NiklasBr:
```suggestion
            return $a['index'] <=> $b['index'];
```

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -208,6 +241,20 @@ public function isOptimizedAdminLoading(): bool
     {
         return true;
     }
+
+    protected function enrichDataRow($object, array $params, ?string &$classId, &$row = [], string $srcCol = 'src_id')
```
NiklasBr: `string $srcCol = 'src_id'` is not used as far as I can see

BlackbitDevs: This method overrides https://github.com/pimcore/pimcore/blob/338fcc9554fff718429275ac42b19574c46f0e1a/models/DataObject/Traits/ContextPersistenceTrait.php#L36 and thus has to have the same parameters.

weisswurstkanone:
Could you please rebase on the latest version `?

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -102,13 +103,17 @@ public function getOwnerClassName()
...
-     * @return string
+     * @return string|null
      */
     public function getOwnerClassId()
     {
         if (empty($this->ownerClassId)) {
             try {
                 $class = DataObject\ClassDefinition::getByName($this->ownerClassName);
+                if(!$class instanceof DataObject\ClassDefinition) {
+                    Logger::error('Reverse relation '.$this->getName().' has not owner class assigned');
```
weisswurstkanone:
```suggestion
                    Logger::error('Reverse relation '.$this->getName().' has no owner class assigned');
```

BlackbitDevs:
@weisswurstkanone: Is now up-to-date with `10.x` branch.

weisswurstkanone:
Thanks. One more thought: A unit test would be nice 😃

weisswurstkanone:
I think this stuff shouldn't be grayed out

image

Please check https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/reverseObjectRelation.js#L35

weisswurstkanone:
The rest looks good so far. @BlackbitNeueMedien , could you please have a look at the 2 comments? Merci!

weisswurstkanone:
Just one more thing I noticed, maybe we can pack this into this PR.

This line will bomb out if both ownerClassName and ownerClassId is null because getById expects a string.

https://github.com/pimcore/pimcore/blob/9e6f0b56f4636757e802f96d6c475fce95352fa1/models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php#L95

tests/_support/Test/DataType/AbstractDataTypeTestCase.php
```diff
@@ -710,6 +710,18 @@ public function testObjects()
...
 
+    public function testReverseRelation()
+    {
+        // // this will create a couple of objects which can be used for references
+        $testObjects = TestHelper::createEmptyObjects();
+
+        $this->createTestObject('objects');
+
+        $this->refreshObject();
+        $this->assertCount(1, $testObjects[0]->getNonowner());
```
weisswurstkanone, line 724: can't see where the relation got populated

BlackbitDevs: I also struggle a bit with the tests ;-)
But as the reverse relation has the same class as itself as `ownerClass` and the field `objects` and owner field, it should be enough to assign the objects to the `objects` field (this is done in the test by `$this->createTestObject('objects')`. Or am I wrong?

blankse: The `refreshObject()` method load the object to the testObject property. So this could work:
```suggestion
        $this->assertCount(1, $this->testObject->getNonowner());
```

blankse: The `$testObjects = TestHelper::createEmptyObjects();` can be removed. The `$this->createTestObject('objects');` creates also an object to the testObject property

blankse: Oh I'm wrong. The `$this->testObject` should relates to the `$testObjects`. This is missing. You need to give an array with fields to the createTestObject method.

tests/_support/Test/DataType/AbstractDataTypeTestCase.php
```diff
@@ -710,6 +710,18 @@ public function testObjects()
...
     }
 
+    public function testReverseRelation()
+    {
+        // // this will create a couple of objects which can be used for references
+        $testObjects = TestHelper::createEmptyObjects();
+
+        $this->createTestObject('objects');
+
+        $this->refreshObject();
```
blankse: Maybe so?
```suggestion
        $this->createTestObject('objects');
        $testObjects = $this->testObject->getObjects();

```

weisswurstkanone:
Spent some time debugging it, I would say the test should be like

```
        $testObjects = TestHelper::createEmptyObjects();

        $this->createTestObject('objects');

        $this->refreshObject();
        $this->assertCount(1, $this->testObject->getNonowner());
        $this->testDataHelper->assertObjectsEqual($this->testObject->getNonowner()[0], $this->testObject);
```

and then it succeeds.

**BUT**

I think we still have a major problem.

Everything relies on

```
        $this->refreshObject();
```

which is not really practical in a real life.

The problem is that without that

```
https://github.com/pimcore/pimcore/blob/acb371a1bff695d56084c9f372d7eab41361f305/models/DataObject/ClassDefinition/Data/ManyToManyObjectRelation.php#L453
```

thinks that the field is already loaded.

It gets marked as loaded here

image

Could imagine to never mark reverse stuff as loaded but I am not sure what the consequences would be.

wdyt ?

BlackbitDevs:
@weisswurstkanone Thank you, I fixed the test.
The remaining failing test has nothing to do with this PR imho, does it? I already merged `10.x` but to no success.

Concerning changes to the owner relation after the reverse relation got loaded: You are right, we cannot get track of those. The only way I see is to watch the original relation for changes somewhere in the save method - or even better in the setters to also get non-saved changes. This is not easy as we have no direct relation between the saved object and the watching reverse relation - we could only use the runtime cache for this imho. 
For now I always reload the reverse relation to always keep track of changes to the owner relation. I think this is ok because it goes in the same direction as the warning message in the backend "Would you still like to implicitly  modify it with this action?" when an object is opened which the reverse relation contains.

tests/_support/Test/DataType/AbstractDataTypeTestCase.php
BlackbitDevs: Thanks for your input @blankse , I fixed the test.

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -176,11 +184,39 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
...
-    public function getForCsvExport($object, $params = [])
+    public function load($object, $params = [])
     {
-        return '';
+        $db = Db::get();
+
+        if($this->getOwnerClassId() === null) {
+            return [];
+        }
+
```
brusch:
```suggestion
        if($this->getOwnerClassId() === null) {
            return [];
        }

        $db = Db::get();
```

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -176,11 +184,39 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
...
+        $relations = array_map(static function ($relation) {
+            $relation['dest_id'] = $relation['src_id'];
+            unset($relation['src_id']);
+            return $relation;
+        }, $relations);
+
+        // using PHP sorting to order the relations, because "ORDER BY index ASC" in the queries above will cause a
+        // filesort in MySQL which is extremely slow especially when there are millions of relations in the database
+        usort($relations, static function ($a, $b) {
```
brusch: Does it even make sense to order them here, as the order of the relations in the originating data object has absolutely no value here, right?

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -208,6 +244,28 @@ public function isOptimizedAdminLoading(): bool
     {
         return true;
     }
+
+    protected function enrichDataRow($object, array $params, ?string &$classId, &$row = [], string $srcCol = 'src_id')
```
brusch: `enrichDataRow()` is only used by `save()`, so don't see any point in implementing it.

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -208,6 +244,28 @@ public function isOptimizedAdminLoading(): bool
...
+            $classId = $this->getOwnerClassId();
+        }
+    }
+
+    public function preGetData($container, $params = [])
+    {
+        if($container instanceof DataObject\LazyLoadedFieldsInterface) {
+            $container->unmarkLazyKeyAsLoaded($this->getName());
+        }
+        return parent::preGetData($container, $params);
```
brusch:
```suggestion
        return $this->load($container);
```
Since reverse relations are only allowed in the object context (not in localized fields, field collections, object bricks, it should be fine to just return the loaded data. No need to store it on the property or care about lazy loading - let's keep things simple 😉

models/DataObject/ClassDefinition.php
```diff
@@ -551,7 +551,7 @@ public function generateClassFiles($generateDefinitionFile = true)
 
         if (is_array($this->getFieldDefinitions()) && count($this->getFieldDefinitions())) {
             foreach ($this->getFieldDefinitions() as $key => $def) {
-                if (!$def instanceof DataObject\ClassDefinition\Data\ReverseObjectRelation && !$def instanceof DataObject\ClassDefinition\Data\CalculatedValue
+                if (!$def instanceof DataObject\ClassDefinition\Data\CalculatedValue
```
brusch: This can be reverted, no need to include a property that doesn't get used at all 😉

brusch (changes requested):
@BlackbitNeueMedien we're nearly there, just some smaller things are left, would be nice if you could update the PR accordingly. Thanks a lot!

models/DataObject/ClassDefinition.php
BlackbitDevs: Done

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
BlackbitDevs: Right, I removed it because Reverse relations are set "untouchable" in 
https://github.com/pimcore/pimcore/blob/af69638badd601feb3b4d695e6eec0eb9d7462bb/models/DataObject/Concrete/Dao.php#L205-L211
and thus in
https://github.com/pimcore/pimcore/blob/af69638badd601feb3b4d695e6eec0eb9d7462bb/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php#L89-L91
saving gets skipped and so https://github.com/pimcore/pimcore/blob/af69638badd601feb3b4d695e6eec0eb9d7462bb/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php#L117 does never get called.

Removed it (above explanation is only for our future selfs ;-)

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
BlackbitDevs: You are right, ordering by `index` does not make sense. Removed it for now.
Would it make sense to order by `src_id`?

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
BlackbitDevs: Right, I removed it because Reverse relations are set "untouchable" in 
https://github.com/pimcore/pimcore/blob/af69638badd601feb3b4d695e6eec0eb9d7462bb/models/DataObject/Concrete/Dao.php#L205-L211
and thus in
https://github.com/pimcore/pimcore/blob/af69638badd601feb3b4d695e6eec0eb9d7462bb/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php#L89-L91
saving gets skipped and so https://github.com/pimcore/pimcore/blob/af69638badd601feb3b4d695e6eec0eb9d7462bb/models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php#L117 does never get called.

Removed it (above explanation is only for our future selfs ;-)

BlackbitDevs:
@brusch Do you know what Stan wants to tell me?

> models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php, Line 95  
> Negated boolean expression is always false.

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
brusch: No case I can think of, so I'd just leave it without sorting.

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -208,6 +237,11 @@ public function isOptimizedAdminLoading(): bool
     {
         return true;
     }
+
+    public function preGetData($container, $params = [])
+    {
+        return $this->load($container);
+    }
```
brusch, line 244:
```suggestion
    }
    
    public function supportsInheritance()
    {
        return false;
    }
```

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -84,13 +85,16 @@ public function setOwnerClassName($ownerClassName)
...
     /**
-     * @return string
+     * @return string|null
      */
     public function getOwnerClassName()
     {
         //fallback for legacy data
         if (empty($this->ownerClassName)) {
             try {
+                if(!$this->ownerClassId) {
```
brusch:
```suggestion
                if(empty($this->ownerClassId)) {
```

brusch (changes requested):

brusch:
Thank you very much!

PR #10244, opened 2021-09-06
Remove CKEditor wordcount plugin

### LINKED ISSUE #10033: [Improvement]: Remove CKEditor wordcount plugin
### Improvement description

Was introduced by #9554
The problem is that the counters are not part of the toolbar, so you can't configure the place where it should be displayed, which is awful.

brusch:
Resolves #10033
Reverts #9554


cancan101:
with this being reverted, should the config field have been removed?:
image

PR #10250, opened 2021-09-06
[Select field] Throw exception when manually creating options with wrong parameter structure

BlackbitDevs:
Steps to reproduce problem:
1. Create class `Test`
2. Create select field `test`
3. Run the following script:
```php
<?php
$class = \Pimcore\Model\DataObject\ClassDefinition::getByName('Test');
$field = $class->getFieldDefinition('test');
$field->setOptions([['label 1', 'value 1']]);
$class->save();
```
You will get no error and in field configuration everything will look fine:
Bildschirmfoto 2021-09-06 um 15 41 28

Yet when you look at the available options in the dropdown you will notice that something is wrong:
Bildschirmfoto 2021-09-06 um 15 40 55

We first thought that something about translations is wrong but actually the cause is
`$field->setOptions([['label 1', 'value 1']]);`
which is wrong, it actually has to be
`$field->setOptions([['key' => 'label 1', 'value' => 'value 1']]);`
But it took us some time to notice that. With this PR an exception gets thrown when `setOptions()` gets called with wrong parameter structure.

PS: I am not sure if we have to allow for `null` here but as it was possible before, I kept this behaviour.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Select.php
+++ b/models/DataObject/ClassDefinition/Data/Select.php
@@ -201,6 +201,7 @@ class Select extends Data implements
         if(is_array($options)) {
             $this->options = [];
             foreach ($options as $option) {
+                $option = (array)$option; // support stdclass option
                 if (!array_key_exists('key', $option) || !array_key_exists('value', $option)) {
                     throw new InvalidArgumentException('Please provide select options as associative array with fields "key" and "value"');
                 }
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/Select.php
```diff
@@ -195,9 +196,20 @@ public function getOptions()
      *
      * @return $this
      */
-    public function setOptions($options)
+    public function setOptions(?array $options)
```
dvesh3, line 199: update phpdoc to have null

BlackbitDevs: Done

dvesh3:

PR #10253, opened 2021-09-06
Check if file is SVG when generating thumbnail

### LINKED ISSUE #10252: [Bug]: Cannot upload SVG Asset
### Expected behavior

SVG can be added as any other asset.

### Actual behavior

Upload modal freezes due to error in asset upload. In network tab, request /add-asset seems to return HTTP 500 with following error:

> Pimcore\Model\Asset\Image\Thumbnail::getLocalFileFromStream(): Return value must be of type string, null returned

### Steps to reproduce

Try to upload an SVG file into assets in Pimcore Admin.

iidaloiki:
Resolves #10252

## Additional info  

Adds check in thumbnail generation whether the image asset is .svg. If it is, it returns the asset file itself without modifications.

### DIFF
--- a/models/Asset/Image/Thumbnail.php
+++ b/models/Asset/Image/Thumbnail.php
@@ -158,8 +158,9 @@ final class Thumbnail
         $generated = false;
 
         if ($this->asset && empty($this->pathReference)) {
-            // if no correct thumbnail config is given use the original image as thumbnail
-            if (!$this->config) {
+            $isSvg = $this->asset->getMimetype() == 'image/svg+xml' ? true : false;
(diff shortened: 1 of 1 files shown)


brusch:
@iibs Thanks a lot for your PR, however, this is not entirely right, because we also need to consider this setting: 
image


Since this PR is a bug fix, it also needs to be rebased onto the `10.1` branch.

brusch:
Closed for now, as #10252 is not reproducible. 
Let's continue the discussion on #10252 - I'll reopen here if necessary 😊 Thanks!

PR #10254, opened 2021-09-06
[Assets] Display image at natural size, if smaller than max dimensions

iidaloiki:
Normal images work fine with current styling (defining max-width and max-height), but when .svg file is previewed, it is not shown due to lack of definition of width and height.

This PR defines those based on image dimensions while keeping maximum dimensions. Object fit is added so that image contains its aspect ratio.


brusch:
Actually this PR makes it even worse for me 🤔

image

The preview image get's really small.

PR #10261, opened 2021-09-07
[Shared Translations] Show all languages for admin users

BlackbitDevs:
Steps to reproduce:
1. Ensure in `Settings > Localization` there is only `English` 
2. Create admin user and enable checkbox for viewing and editing shared translation for `English`
3. In `Settings > Localization` add `German`
4. After reload go to shared translations
5. You will see that current user will only see english (as configured)
6. When you now go to user settings the panel for languages to be shown in shared translations is not displayed due to https://github.com/pimcore/pimcore/blob/7edef74976d037e913a6ea7d2c497a566d62db57/bundles/AdminBundle/Resources/public/js/pimcore/settings/user/user/settings.js#L557-L559

As the panel is not shown for admin users, there is no way to allow the newly added language `german` - only if the `admin` checkbox is removed and the user tab gets reloaded, the language panel will appear again.

This PR does 2 things:
- immediately show and hide language panel when the admin checkbox gets enabled or disabled (so that it works the same as when the user tab gets closed and reopened again)
- admin users can see and edit all languages in shared translations

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/user/user/settings.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/user/user/settings.js
@@ -399,11 +399,13 @@ pimcore.settings.user.user.settings = Class.create({
                         this.typesSet.hide();
                         this.permissionsSet.hide();
                         this.userPanel.workspaces.disable();
+                        this.websiteTranslationSettings.getPanel().hide();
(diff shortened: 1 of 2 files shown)


dvesh3 (approved):

dvesh3:
@BlackbitNeueMedien thanks!

PR #10265, opened 2021-09-07
[Classification Store] Add static getById methods

BlackbitDevs:
Add `getById` methods for `CollectionGroupRelation` and `KeyGroupRelation` - similar to all the other classification store classes like https://github.com/pimcore/pimcore/blob/7edef74976d037e913a6ea7d2c497a566d62db57/models/DataObject/Classificationstore/CollectionConfig.php#L66-L76

### DIFF
--- a/models/DataObject/Classificationstore/CollectionGroupRelation.php
+++ b/models/DataObject/Classificationstore/CollectionGroupRelation.php
@@ -61,8 +61,8 @@ final class CollectionGroupRelation extends Model\AbstractModel
     }
 
     /**
-     * @param int|null $colId
      * @param int|null $groupId
+     * @param int|null $colId
      *
      * @return self|null
      */
--- a/models/DataObject/Classificationstore/CollectionGroupRelation.php
+++ b/models/DataObject/Classificationstore/CollectionGroupRelation.php
@@ -66,7 +66,7 @@ final class CollectionGroupRelation extends Model\AbstractModel
      *
      * @return self|null
      */
-    public static function getByColAndGroupId($colId = null, $groupId = null)
+    public static function getByGroupAndColId($groupId = null, $colId = null)
     {
         try {
             $config = new self();
--- a/models/DataObject/Classificationstore/CollectionGroupRelation.php
+++ b/models/DataObject/Classificationstore/CollectionGroupRelation.php
@@ -63,6 +63,8 @@ final class CollectionGroupRelation extends Model\AbstractModel
     /**
      * @param int|null $colId
      * @param int|null $groupId
+     *
+     * @return self|null
      */
     public static function getById($colId = null, $groupId = null)
     {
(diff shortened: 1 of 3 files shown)


models/DataObject/Classificationstore/KeyGroupRelation.php
```diff
@@ -85,6 +85,22 @@ public static function create()
         return $config;
     }
 
+    /**
+     * @param int|null $keyId
+     * @param int|null $groupId
+     */
+    public static function getById($keyId = null, $groupId = null)
```
dvesh3: Do we really need this? we already have `getByGroupAndKeyId` https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/Classificationstore/KeyGroupRelation.php#L238

models/DataObject/Classificationstore/CollectionGroupRelation.php
```diff
@@ -60,6 +60,22 @@ public static function create()
         return $config;
     }
 
+    /**
+     * @param int|null $colId
+     * @param int|null $groupId
+     */
```
dvesh3:
```suggestion
    /**
     * @param int|null $colId
     * @param int|null $groupId
     *
     * @return self|null
     */
```

models/DataObject/Classificationstore/KeyGroupRelation.php
BlackbitDevs: You are right @dvesh3 - I removed the getById() but refactored `getByGroupAndKeyId()` to use the already existing https://github.com/pimcore/pimcore/blob/0b563345ac022a823e3bf16ac9fd1eafc72fd30f/models/DataObject/Classificationstore/KeyGroupRelation/Dao.php#L33-L47

models/DataObject/Classificationstore/CollectionGroupRelation.php
```diff
@@ -60,6 +60,24 @@ public static function create()
         return $config;
     }
 
+    /**
+     * @param int|null $colId
+     * @param int|null $groupId
+     *
+     * @return self|null
+     */
+    public static function getByColAndGroupId($colId = null, $groupId = null)
```
dvesh3: how about we change the method name to `getByGroupAndColId` along with correct param order? so that it is consistent with `KeyGroupRelation::getByGroupAndKeyId`

models/DataObject/Classificationstore/KeyGroupRelation.php
dvesh3: looks good to me. thanks!

models/DataObject/Classificationstore/CollectionGroupRelation.php
BlackbitDevs: Done

models/DataObject/Classificationstore/CollectionGroupRelation.php
```diff
@@ -60,6 +60,24 @@ public static function create()
         return $config;
     }
 
+    /**
+     * @param int|null $colId
+     * @param int|null $groupId
```
NiklasBr:
```suggestion
     * @param int|null $groupId
     * @param int|null $colId
```

dvesh3:

PR #10270, opened 2021-09-07
improvement: remove advanced relation double click event

### LINKED ISSUE #10051: [Improvement]: correct advanced many to many (object) relation event to open element on double click
### Improvement description

When using an advanced many to many (object) relation field, double clicking the editable metadata columns triggers the event to open the element related to the record. This is especially annoying with select fields, since it requires two clicks to open the options dropdown.

### To reproduce

1. Edit (or create) any class and configure an advanced many to many object relation with at least a select column.
2. Open a data object of this class and add a relation object.
3. Double click the select to open the options dropdown.

### Possible solutions

1. Remove the event entirely, since an icon is provided to open the element.
2. Only trigger the event from the columns containing element data, ignoring the editable cells. Possibly the [celldblclick event](https://docs.sencha.com/ext/6.5.1/classic/Ext.grid.Panel.html#event-celldblclick) could be used for this.
3. Attempt to ignore the editable cells in the current event listeners. This may be rather tricky though, since you'll need to figure out a way to determine what the cell contents are in a row based event.

In addition, should the select options dropdown be opened with one click? And if so, how can this be achieved?

### Technical details

Both the [advanced many to many object relation](https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js#L430) and [advanced many to many relation](https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyRelation.js#L396) use the [abstract relations gridRowDblClickHandler](https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/abstractRelations.js#L198).

kjkooistra-youwe:
When double clicking metadata fields of an advanced many to many relation row
the related data object is opened. This is very annoying, as it requires two
clicks to open metadata select dropdowns or focus the metadata input field.
Resolved by removing the double click event for the advanced relation fields.
An open icon is already available to open the record.
Resolves #10051

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyObjectRelation.js
@@ -425,10 +425,7 @@ pimcore.object.tags.advancedManyToManyObjectRelation = Class.create(pimcore.obje
             bodyCls: "pimcore_object_tag_objects pimcore_editable_grid",
             plugins: [
                 this.cellEditing
-            ],
-            listeners: {
-                rowdblclick: this.gridRowDblClickHandler
-            }
+            ]
         });
 
         if (!readOnly) {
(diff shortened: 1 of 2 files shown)


dvesh3:
@kjkooistra-youwe thanks!

PR #10276, opened 2021-09-08
[Select field] Support stdclass in select options

BlackbitDevs:
Follow-up of #10250

As some points in code use `json_decode()` without second parameter so that JSON objects get decoded to assosicative arrays, select field options for classification store select fields can end up as array of `stdclass` objects, e.g. in https://github.com/pimcore/pimcore/blob/606484e4708bb6d6d2080d8b90152451164c67c5/bundles/AdminBundle/Helper/GridHelperService.php#L110

This PR supports stdclass notation for `setOptions()`


dvesh3:
@BlackbitNeueMedien thanks!

PR #10283, opened 2021-09-09
Crisp SVG preview

### LINKED ISSUE #10086: [Bug]: SVG image blurred
### Expected behavior

SVG image thumbs are sharp

### Actual behavior

SVG image thumbnails are all blurred.

### Steps to reproduce

Go to pimcore Demo:
https://demo.pimcore.com/admin/login/deeplink?document_151_page

 

Go to the snippet: 
https://demo.pimcore.com/admin/login/deeplink?document_166_snippet
and change the image (need to override) to this svg: https://demo.pimcore.com/admin/login/deeplink?asset_368_image

iidaloiki:
Another take on PRs #10253 and #10254

Resolves #10086

## Additional info  

When working with small svgs, current thumbnail generation blurs them, because they are rasterized even when they shouldn't (tree preview generates checkered background).

This PR takes on account the rasterizeSVG setting. Minimum dimensions 200x200 shouldn't be an issue, since tree preview generates scale by 400px width when thumbnail is used.

Two example images: One with Pimconaut (large svg) and a simple user icon (small svg).

Näyttökuva 2021-9-9 kello 13 45 53
Näyttökuva 2021-9-9 kello 13 48 41

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/image.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/image.js
@@ -438,8 +438,11 @@ pimcore.asset.image = Class.create(pimcore.asset.asset, {
                     this.initPreviewImage();
                     var area = this.displayPanel.getEl().down('img');
                     if(area) {
+                        area.setStyle('min-width', "200px");
+                        area.setStyle('min-height', "200px");
                         area.setStyle('max-width', (width - 340) + "px");
                         area.setStyle('max-height', (height - 40) + "px");
+                        area.setStyle('object-fit', "contain");
                     }
                 }
             }.bind(this));
--- a/models/Asset/Image/Thumbnail.php
+++ b/models/Asset/Image/Thumbnail.php
@@ -159,7 +159,7 @@ final class Thumbnail
 
         if ($this->asset && empty($this->pathReference)) {
             // if no correct thumbnail config is given use the original image as thumbnail
-            if (!$this->config) {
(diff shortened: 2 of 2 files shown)


brusch:
@iibs I'm afraid that this is still not solving the issue in a generic way. The actual problem is that the SVG isn't rasterized properly by Imagick, just not rasterizing the image is just covering a small range of use-cases. 
In this case it's solving the issue with the preview, but not for custom thumbnails with `rasterizeSVG: true`.

brusch:
Actually the functionality is already there, see: 
https://github.com/pimcore/pimcore/blob/1e2adbc3f4da869767e1f0966c438a0a9e72ebff/models/Asset/Image/Thumbnail.php#L139
But it's not used by previews because of: 
https://github.com/pimcore/pimcore/blob/4d10439eb75b76acb8c1e07007f57d0d4e43078f/models/Asset/Image/Thumbnail/Config.php#L853

So I think we really need to check why Imagick is not generating crisp rastered images out of PNGs.

iidaloiki:
I can try to look into it next week, if I could come up with a more generic solution. I tried to debug the issue, but didn't quite catch yet why the thumbnail was generated the way it is. I'll look into it a bit further.

Is it ok if I'll make changes to this PR if I find a better alternative?

brusch:
@iibs cool, thanks! Yes, of course you can update this PR, I'll leave it open for now.

brusch:
@iibs any updates on this? I'll otherwise close this PR for now. Thanks for a quick feedback 😊

brusch:
Closed due to inactivity - please feel free to come back on this PR - we're happy to reopen it 😊

PR #10287, opened 2021-09-09
[Editables] - Allow scrolling inside renderlet

roland4432:

Queried the bodyWrap element to add styling to it, which allows for scrolling in editmode

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/renderlet.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/renderlet.js
@@ -11,10 +11,10 @@
  * @license    http://www.pimcore.org/license     GPLv3 and PCL
  */
 
-pimcore.registerNS("pimcore.document.editables.renderlet");
+pimcore.registerNS('pimcore.document.editables.renderlet');
 pimcore.document.editables.renderlet = Class.create(pimcore.document.editable, {
 
-    defaultHeight: 100,
+    defaultHeight: 150,
 
     initialize: function(id, name, config, data, inherited) {
         this.id = id;
@@ -35,28 +35,32 @@ pimcore.document.editables.renderlet = Class.create(pimcore.document.editable, {
         if (this.config.defaultHeight) {
             this.defaultHeight = this.config.defaultHeight;
         }
-        if (!this.config.height) {
+
+        if (this.config.height) {
+            this.initalHeightSet = true;
+        }
+        else {
+            this.initalHeightSet = false;
             this.config.height = this.data.id ? 'auto' : this.defaultHeight;
         }
 
-        this.config.name = id + "_editable";
+        this.config.name = id + '_editable';
         this.config.border = false;
-        this.config.bodyStyle = "min-height: 40px;";
+        this.config.bodyStyle = 'min-height: 40px;';
     },
 
     render: function() {
         this.setupWrapper();
 
         this.element = new Ext.Panel(this.config);
-        this.element.on("render", function (el) {
-
+        this.element.on('render', function (el) {
             // register at global DnD manager
             dndManager.addDropTarget(el.getEl(), this.onNodeOver.bind(this), this.onNodeDrop.bind(this));
 
-            this.getBody().insertHtml("beforeEnd",'<div class="pimcore_editable_droptarget"></div>');
-            this.getBody().addCls("pimcore_editable_snippet_empty");
+            this.getBody().insertHtml('beforeEnd','<div class="pimcore_editable_droptarget"></div>');
+            this.getBody().addCls('pimcore_editable_snippet_empty');
 
-            el.getEl().on("contextmenu", this.onContextMenu.bind(this));
+            el.getEl().on('contextmenu', this.onContextMenu.bind(this));
 
         }.bind(this));
 
@@ -68,7 +72,6 @@ pimcore.document.editables.renderlet = Class.create(pimcore.document.editable, {
     },
 
     onNodeDrop: function (target, dd, e, data) {
-
         if(!pimcore.helpers.dragAndDropValidateSingleItem(data)) {
             return false;
         }
@@ -135,21 +138,21 @@ pimcore.document.editables.renderlet = Class.create(pimcore.document.editable, {
     },
 
     getBodyWrap: function () {
-        var bodyId = this.element.getEl().query("." + Ext.baseCSSPrefix + "panel-bodyWrap")[0].getAttribute("id");
+        var bodyId = this.element.getEl().query('.' + Ext.baseCSSPrefix + 'panel-bodyWrap')[0].getAttribute('id');
         return Ext.get(bodyId);
     },
 
     getBody: function () {
         // get the id from the body element of the panel because there is no method to set body's html
         // (only in configure)
-        var bodyId = this.element.getEl().query("." + Ext.baseCSSPrefix + "panel-body")[0].getAttribute("id");
+        var bodyId = this.element.getEl().query('.' + Ext.baseCSSPrefix + 'panel-body')[0].getAttribute('id');
         return Ext.get(bodyId);
     },
 
     updateContent: function () {
         var self = this;
 
-        this.getBody().removeCls("pimcore_editable_snippet_empty");
+        this.getBody().removeCls('pimcore_editable_snippet_empty');
         this.getBody().dom.innerHTML = '<br />&nbsp;&nbsp;Loading ...';
 
         var params = this.data;
@@ -159,7 +162,7 @@ pimcore.document.editables.renderlet = Class.create(pimcore.document.editable, {
         try {
             // add the id of the current document, so that the renderlet knows in which document it is embedded
             // this information is then grabbed in Pimcore_Controller_Action_Frontend::init() to set the correct locale
-            params["pimcore_parentDocument"] = window.editWindow.document.id;
+            params['pimcore_parentDocument'] = window.editWindow.document.id;
         } catch (e) {
         }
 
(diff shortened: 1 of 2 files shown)


brusch:
@roland4432 can you please share a screenshot and an code-example that shows the effect of this PR? 
Thanks a lot!

roland4432:
Sure thing my friend. Here you go:

Screenshot *before* the changes (no scrollbars, content is hidden if larger than the visible area):

Bildschirmfoto vom 2021-09-10 10-28-21

and a short video *after* the changes:

https://user-images.githubusercontent.com/16651237/132824956-83d242cd-e032-476d-b3d7-c64c1929e709.mp4

There is no code involved. The container from the renderlet just autodecides if it needs vertical scrollbars or not.

brusch:
@roland4432 thanks a lot! 
However, I'm just wondering if it shouldn't automatically adjust the size 🤔

roland4432:
At first I thought so too, but the body element is styled like this:
```css
min-height: 40px;
overflow: auto;
width: 1396px;
height: auto;
left: 0px;
top: 49px;
```
and this won't allow for scrolling with _height: auto;_. This rather follows the parent container, which has a fixed height (either via config or fallback from code) and _overflow: hidden;_ therefore disallowes scrolling (this is set for the _bodyWrap_ element)

```css
.x-panel-bodyWrap: {
    overflow: hidden;
    position: static;
    height: 100%;
    width: 100%;
}
```


Just thinking, if this is only needed, if you pass a _height_ attribute to the pimcore_renderlet...

brusch:
Yeah, I think that's just necessary if using `height` in the renderlet. 
But your changes shouldn't take affect anyway if `height` was not set manually, right? Can you maybe just quickly check that? Thanks again!

roland4432:
If you don't set a height, then a default min-height will be set in the constructor of the renderlet editables js code

```js
//renderlet.js
 ...
    initialize: function(id, name, config, data, inherited) {
        ...
        // height management
        this.defaultHeight = 100;
        if (this.config.defaultHeight) {
            this.defaultHeight = this.config.defaultHeight;
        }
        if (!this.config.height) {
            this.config.height = this.defaultHeight;
        }
        ...
    },
... 
```

So there is no case with this code, where the renderlet is auto deciding its height.
I tested it both with tiny, large and null|undefined values passing as height.
The scrollbar works fine and won't be shown if you put a height > your content height from the renderlet.

Seems fine to me :)

brusch:
Hmm, but when updating the content of the renderlet, the height of the body element is set to `auto`. So I'd expect to not have a scrollbar if I don't set a height manually.

roland4432:
Now I get it :D sry.. long time ..
You wanna setup the renderlet like this:

```twig
pimcore_renderlet('my-renderlet', {height: 'auto'})
```

So yeah.. I could totally alter the PR to check, if height is set to _auto_ and dont change the style of the bodyWrap element, since with _auto_ scrollbars won't be necessary at all

Regarding:
> when updating the content of the renderlet, the height of the body element is set to auto

Totally true, but the body element cannot enlarge its parent with **auto** if the parent has a fixed height which it has, if you setup the renderlet via:
```
pimcore_renderlet('my-renderlet, {height: 400})
```

brusch:
Regarding auto-height: yes, but `auto` should be default, so: 
`pimcore_renderlet('my-renderlet')` should automatically adjust the height after assigning an element to the renderlet. 

If using a fixed height, so: 
`pimcore_renderlet('my-renderlet, {height: 400})` it should use a scrollbar 😉

roland4432:
Alrighty then. I'll get this behaviour in the PR and update asap.

bundles/AdminBundle/Resources/public/js/pimcore/document/editables/renderlet.js
```diff
@@ -37,7 +37,7 @@ pimcore.document.editables.renderlet = Class.create(pimcore.document.editable, {
             this.defaultHeight = this.config.defaultHeight;
         }
         if (!this.config.height) {
-            this.config.height = this.defaultHeight;
+            this.config.height = 'auto';
```
brusch:
```suggestion
            this.config.height = this.defaultHeight;
```
This is not necessary, as we're setting the body to auto anyway if a content is set. 
Happens here: https://github.com/pimcore/pimcore/blob/ad99f9a2061365848d15534266310a1b2ed83834/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/renderlet.js#L215

bundles/AdminBundle/Resources/public/js/pimcore/document/editables/renderlet.js
```diff
@@ -55,9 +55,15 @@ pimcore.document.editables.renderlet = Class.create(pimcore.document.editable, {
             dndManager.addDropTarget(el.getEl(), this.onNodeOver.bind(this), this.onNodeDrop.bind(this));
 
             this.getBody().setStyle({
-                overflow: "auto"
+                overflow: 'auto',
             });
 
+            if (this.config.height !== 'auto') {
+                this.getBodyWrap().setStyle({
```
brusch: I think we need to move this to `updateDimensions()` and 
```javascript
        this.getBody().setStyle({
            height: "auto"
        });
``` 

needs to be inside the `if` condition.

brusch (changes requested):

bundles/AdminBundle/Resources/public/js/pimcore/document/editables/renderlet.js
roland4432: Small update here and some notes:

the default body height has to take into account if there is data set or not, hence the:
```
        if (!this.config.height) {
            this.config.height = this.data.id ? 'auto' : this.defaultHeight;
        }
```
Else the drop zone icon would glitch out and looks broken.

2nd part as for the: 
> I think we need to move this to updateDimensions() and
>        this.getBody().setStyle({
>            height: "auto"
>        });
>needs to be inside the if condition.

As far as I checked thie _getBody().setStyle()_ has to be outside the if, since it needs to be called under any condition.
Else 2 visual bugs appear. If you pass a fixed height, the body needs to draw its height with auto to trigger the parent to make a vertical scrollbar.
If you did not pass a fixed height to the renderlet, the body needs to be a 100% which auto does as well so that the parents will spawn in a vertical dimension to show the complete content of the renderlet body.

WDYT?

brusch: The `defaultHeight` setting should be only applied if no content was set - that's the only reason for this config. 
I think we should just do it the way the `snippet` editable does, just to be consistent.

roland4432: That's exactly what is being done with: 
```javascript
//line 38++
        if (!this.config.height) {
            this.config.height = this.data.id ? 'auto' : this.defaultHeight;
        }
```

or do you want it to be matching the snippet syntax with
```javascript
        if (!this.config.height && !this.data.id) {
            this.config.height = this.defaultHeight;
        }
```
This would not work, since the container would not be set to 'auto' as soon as it has content.

The current PR approach is working and taking into account wether the renderlet has content or not.


FYI: The snippet is only taking the height: \<integer\> attribut to render its container if it has no content. As soon as you drop content into a snippet, it will go to height: 'auto'. Is this the desired behavior?

brusch: > FYI: The snippet is only taking the height: <integer> attribut to render its container if it has no content. As soon as you drop content into a snippet, it will go to height: 'auto'. Is this the desired behavior?

But what's the point then for `defaultHeight`? Actually `defaultHeight` should be the height if no content is set and `height` should be used no matter if content is set or not, right? 
And this is exactly what the `snippet` editable is doing. 😊

roland4432: Sorry, but no. The snippet is not doing what you think it should.
If you set
```twig
{{ pimcore_snippet('my-snippet', {height: 500}) }}
```

Then you will have a snippet d'n'd area being 500px height initially (or 100px as fallback if you don't pass the height attribute). But as soon as you drag a snippet onto that, then the 500px from this example will be replaced by height => auto and the snippet is as high as the content needs it to be.

IMHO: 
* defaultHeight => fallback if no explicit height has been set via call (e.g. pimcore_renderlet('my-renderlet)) => content is heigher than defaultHeight => height is set to auto to fully show the content of the element without scrollbars
* height => height of the container (snippet or renderlet) set to exactly this value (e.g. pimcore_renderlet('my-renderlet, {height: 500})), wether content is set or not (if set, and content is heigher => show scrollbars)

:D do you agree on that or are we still miles apart?

brusch: @roland4432 I agree. Can you change `snippet` to behave the same?

roland4432: Sure thing. :+1:

roland4432:
There are a lot more things to consider, than I thought.
The renderlet features a config `{reload: true}` and an optional config for `{title: 'some title'}`.

The snippet on the other hand had some bugs with duplication of the drop icon in the upper right corner. So there were only some minor things to consider.
Anyways tldr; remember the way both editables are configured initially with the `{height: <integer>}` as optional parameter, therefore save this state in a variable and consider this as well as wether content is present or not.

(And did some cleanup / linting of the code)

Looking forward to feedback.

bundles/AdminBundle/Resources/public/js/pimcore/document/editables/renderlet.js
```diff
@@ -11,10 +11,10 @@
  * @license    http://www.pimcore.org/license     GPLv3 and PCL
  */
 
-pimcore.registerNS("pimcore.document.editables.renderlet");
+pimcore.registerNS('pimcore.document.editables.renderlet');
 pimcore.document.editables.renderlet = Class.create(pimcore.document.editable, {
 
-    defaultHeight: 100,
+    defaultHeight: 150,
```
brusch:
```suggestion
    defaultHeight: 100,
```
We should leave the default height as it was (same as snippet).

brusch (changes requested):

brusch:
@roland4432 that looks much better now - thanks a lot! 👍

bundles/AdminBundle/Resources/public/js/pimcore/document/editables/renderlet.js
roland4432: Okay :)

PR #10289, opened 2021-09-09
Update Renderlet doc

CiLLiTxBanger:
I found a mistake in the pimcore doc. May apply to all 10.x versions


dvesh3:
@CiLLiTxBanger directly pushed on 10.1 branch, since this is a fix. thanks! https://github.com/pimcore/pimcore/commit/94447c9f31da84b53d32cb7d39264a9335811090

PR #10290, opened 2021-09-09
[DataObjects] Fix Table data type does not use the table config

### LINKED ISSUE #10285: [Bug]: Table in data object does not use the table config
### Expected behavior

When you create a table config and create a new data object it should use it such as:

**Config**
Screenshot 2021-09-09 at 13 59 24

**Behavior**
Screenshot 2021-09-09 at 13 59 35

### Actual behavior

Actual behavior is very strange, first the table is totally empty, not even the columns are showing, also you first have to create a new row, new columns, than click trash button to get your own data etc.

Screenshot 2021-09-09 at 14 00 03

### Steps to reproduce

Check in discussion https://github.com/pimcore/pimcore/discussions/10284

dturopoli:
Resolves #10285

data is always array and when its empty on object creation if wouldn't pass and this.getInitialData() won't be called

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/table.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/table.js
@@ -21,9 +21,8 @@ pimcore.object.tags.table = Class.create(pimcore.object.tags.abstract, {
 
         this.fieldConfig = fieldConfig;
 
-        if (!data) {
+        if (data.length < 1) {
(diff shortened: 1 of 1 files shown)


dvesh3 (approved):

jdevries14:
@dvesh3 can u also merge it into 10.2?

brusch:
@jdevries14 we do that anyway from time to time 😉 Latest after tagging a bug-fix release from `10.1`.

PR #10291, opened 2021-09-10
Use NotFoundException for all getById methods

blankse:
NotFoundException is used in some models but not in all.


models/Redirect.php
```diff
@@ -147,7 +148,7 @@ public static function getById($id)
             $redirect->getDao()->getById($id);
 
             return $redirect;
-        } catch (\Exception $e) {
+        } catch (NotFoundException $e) {
```
brusch, line 151: I'm just thinking, this could be seen as a BC break. Any opinions on that?

blankse: I see no change of API. So it should no BC break, right?
It should only catches own exceptions and not all. Make it hard to find errors.

brusch: The break is if a custom exception would be thrown that's then not caught anymore. 
Let's say a `ConnectionException`.

blankse: Why should a getById method returns null if there is a `ConnectionException`? This is what I think is wrong for example. This exception should be thrown and logged.

brusch: Well, yes, just wanted to mention, that it could be the case under certain circumstances (not in particular `ConnectionException` but any other exception as well). 
For me it sounds like a theoretical "problem" but you know, ... 😉

brusch: I'd take the risk, if we see that it causes some real issues, we can fix/revert that anyway later on.

jdreesen: You are right, there may be exceptions that were previously caught but are not now. But indeed I'd think this is a more theoretical problem and probably was never intended. So this might also be seen as a bugfix to only catch what is intended.

PR #10293, opened 2021-09-10
UrlSlug table name constant and cache fix

NiklasBr:
Resolves issue where static cache would persist on delete and adds a constant for the urlslug table as used through the system.


brusch:
Totally makes sense, thanks man! 🙏

PR #10296, opened 2021-09-10
Sorter in initial sorting of classificationstore groups in collections not considered

### LINKED ISSUE #10273: Sorter in initial sorting of classificationstore groups in collections not considered
### Expected behavior

Sorter is considered in initial sorting of groups in collections

### Actual behavior

Sorter is not considered in initial sorting, but after reload.
Initial sorting is via interal id

### Steps to reproduce

1. Go to current demo
2. Create Classification Store
3. Create collection and some groups
4. Add groups to collection and define sorter
5. Add classification store to object
6. Add collection of group
7. Sorter is ignored, save object and reload
8. Sorting is correct after reload

dturopoli:
Resolves #10273


weisswurstkanone:
thx!

PR #10299, opened 2021-09-10
[Data Object] CRM Consent-Type not visible because of CSS-Class

### LINKED ISSUE #10006: [Bug]: Pimcore X (10.0.1) Object: CRM Consent-Type not visible because of CSS-Class
### Expected behavior

Expected to see the Consent Checkbox (with the Label "Consent given for field" with the Fieldname, Date and Time.)

image

or 

image

### Actual behavior

The Checkbox and the Label isn't visible because of this CSS: 

.x-box-layout-ct {
    overflow: hidden;
}

[CSS File](https://github.com/pimcore/pimcore/blob/a4bc3905900cbb09e00c0e519489e6c7750c9ea5/bundles/AdminBundle/Resources/public/extjs/css/PimcoreApp-all_1.css#L1724)

image

### Steps to reproduce

1. Open Demo-Project 
2. Add a new Panel and a new Consent in the Customer Class 

image

3. Open an Customer --> Consent Tab --> no Checkbox visible 
4. if you edit the CSS and delete the CSS-Class, you'll see the Checkbox

mcop1:
Closes #10006

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Consent.php
+++ b/models/DataObject/ClassDefinition/Data/Consent.php
@@ -197,6 +197,9 @@ class Consent extends Data implements ResourcePersistenceAwareInterface, QueryRe
             }
             $noteId = $note->getId();
         }
+        elseif ($oldData instanceof DataObject\Data\Consent) {
+            $noteId = $oldData->getNoteId();
+        }
 
         return new DataObject\Data\Consent($data, $noteId);
     }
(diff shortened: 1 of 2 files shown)


models/DataObject/ClassDefinition/Data/Consent.php
```diff
@@ -197,6 +197,9 @@ public function getDataFromEditmode($data, $object = null, $params = [])
             }
             $noteId = $note->getId();
         }
+        elseif ($oldData instanceof DataObject\Data\Consent) {
```
weisswurstkanone, line 200: @mcop1 could you please explain this change because it is not about CSS at all

mcop1: That´s because the noteId wasn´t always set properly, when the corresponding checkbox got (un)checked. Therefore it could happen, that when you save your changes, the noteId would still be missing, resulting in the consent note not being displayed properly (even with the fixed css).

weisswurstkanone:
thx!

mcop1:
Fixed by PR #10006

PR #10300, opened 2021-09-10
Do not create system thumbnails by adding image asset

blankse:
This creation make the upload with many images very slow and leads to timeouts. The thumbnails are generated during use anyway. So why should we make the adding slow?


brusch:
This is intentional, so that preview thumbnails are instantly available. I agree that this is not an ideal solution, but removing that will cause some more issues (e.g. when creating folder previews). 

For me this is a perfect candidate for using a queue (Symfony Messenger).

PR #10312, opened 2021-09-13
Add Asset::removeMetadata() method

blankse:
See https://github.com/pimcore/pimcore/discussions/10311


models/Asset.php
```diff
@@ -1676,6 +1676,34 @@ public function addMetadata($name, $type, $data = null, $language = null)
...
+    {
+        if ($name) {
+            $tmp = [];
+            $name = str_replace('~', '---', $name);
+            if (!is_array($this->metadata)) {
+                $this->metadata = [];
+            }
+
+            foreach ($this->metadata as $item) {
+                if ($item['name'] != $name || $language != $item['language']) {
```
weisswurstkanone: Is there  a way to delete all items for the given name regardless of the language ?

blankse: I add language = '*' option

models/Asset.php
```diff
@@ -1676,6 +1676,35 @@ public function addMetadata($name, $type, $data = null, $language = null)
...
+    {
+        if ($name) {
+            $tmp = [];
+            $name = str_replace('~', '---', $name);
+            if (!is_array($this->metadata)) {
+                $this->metadata = [];
+            }
+
+            foreach ($this->metadata as $item) {
+                if ($item['name'] === $name && ($language == $item['language'] || $language === '*')) {
```
weisswurstkanone, line 1695: I like the idea. I think the docs should be enriched then accordingly 👍

blankse: I add it to the docs

weisswurstkanone:
Thanks a lot!

Issue #10239, opened 2021-09-03
Cannot assign Pimcore\Model\Element\ElementDescriptor to property Pimcore\Model\DataObject\Objectbrick::$object


piotrecky:
### Expected behavior

Pimcore cannot write and read the object with object bricks

### Actual behavior

PHP Fatal error:  Uncaught TypeError: Cannot assign Pimcore\Model\Element\ElementDescriptor to property Pimcore\Model\DataObject\Objectbrick::$object of type ?Pimcore\Model\DataObject\Concrete in /var/www/vendor/myclabs/deep-copy/src/DeepCopy/DeepCopy.php:253
Stack trace:
#0 /var/www/vendor/myclabs/deep-copy/src/DeepCopy/DeepCopy.php(253): ReflectionProperty->setValue(Object(Pimcore\Model\DataObject\Produkt\Parametry), Object(Pimcore\Model\Element\ElementDescriptor))
#1 /var/www/vendor/myclabs/deep-copy/src/DeepCopy/DeepCopy.php(208): DeepCopy\DeepCopy->copyObjectProperty(Object(Pimcore\Model\DataObject\Produkt\Parametry), Object(ReflectionProperty))
#2 /var/www/vendor/myclabs/deep-copy/src/DeepCopy/DeepCopy.php(144): DeepCopy\DeepCopy->copyObject(Object(Pimcore\Model\DataObject\Produkt\Parametry))
#3 /var/www/vendor/myclabs/deep-copy/src/DeepCopy/DeepCopy.php(253): DeepCopy\DeepCopy->recursiveCopy(Object(Pimcore\Model\DataObject\Produkt\Parametry))
#4 /var/www/vendor/myclabs/deep-copy/src/DeepCopy/DeepCopy.php(208): DeepCopy\DeepCopy->copyObjectProperty(Object(Pimcore\Model\DataObject\Data\BlockElement), Object(ReflectionProperty))
#5 /var/www/vendor/myclabs/deep-copy/src/DeepCopy/DeepCopy.php(144): DeepCopy\DeepCopy->copyObject(Object(Pimcore\Model\DataObject\Data\BlockElement))
#6 /var/www/vendor/myclabs/deep-copy/src/DeepCopy/DeepCopy.php(94): DeepCopy\DeepCopy->recursiveCopy(Object(Pimcore\Model\DataObject\Data\BlockElement))
[...]

PR #10319, opened 2021-09-14
Cannot assign Pimcore\Model\Element\ElementDescriptor to property Pimcore \Model\DataObject\Objectbrick::$object

weisswurstkanone:
see #10239

Hi @blankse , this seems to be somehow related to the phpstan 4 level changes (https://github.com/pimcore/pimcore/pull/8684) and I am really not sure how we should tackle this in the ideal world. 

Please understand this change as a discussion point.

So what happens is that while deepcopying the related object gets condensed (i.e. transformed into a ElementDescriptor) and DeepCopy afterwards does this reflection thing 

```
        // Copy the property
        $property->setValue($object, $this->recursiveCopy($propertyValue));
```

which of course does not work because the $object is not a Concrete anymore.

### DIFF
--- a/models/DataObject/Objectbrick.php
+++ b/models/DataObject/Objectbrick.php
@@ -47,7 +47,7 @@ class Objectbrick extends Model\AbstractModel implements DirtyIndicatorInterface
      *
      * @var Model\DataObject\Concrete|null
      */
-    protected ?Concrete $object = null;
+    protected $object = null;
 
     /**
      * @internal
--- a/models/DataObject/Objectbrick/Data/AbstractData.php
+++ b/models/DataObject/Objectbrick/Data/AbstractData.php
@@ -54,7 +54,7 @@ abstract class AbstractData extends Model\AbstractModel implements Model\DataObj
     /**
      * @var Concrete|null
      */
-    protected ?Concrete $object = null;
+    protected $object = null;
 
     /**
      * @var int|null


blankse:
@weisswurstkanone I don't see any other option either. I would add ElementDescriptor to the PhpDoc and it is possible to cover this by a test?

piotrecky:
I tested this changes and the type should also be removed in the Pimcore\Model\DataObject\Objectbrick::$object file.

After these two changes, the error no longer occurs and the caching works as expected.

brusch:
@piotrecky can you confirm those 2 changes in this PR are fixing the issue for you: 
https://github.com/pimcore/pimcore/pull/10319/files

Thanks a lot!

piotrecky:
Yes, I confirm. This changes correct the error. Thanks!

brusch:
@piotrecky thanks for you feedback!

GALCF:
@brusch I have the same issue with Fieldcollections, right here:

https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/Fieldcollection/Data/AbstractData.php#L44

dvesh3:
> @brusch I have the same issue with Fieldcollections, right here:
> 
> https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/Fieldcollection/Data/AbstractData.php#L44

Fixed by #10878

PR #10320, opened 2021-09-14
Fix Scheduled Tasks

### LINKED ISSUE #10318: [Bug]: Scheduled Tasks
### Expected behavior

Can save a schedule task

### Actual behavior

Can not save a schedule task

### Steps to reproduce

1. Open https://demo.pimcore.fun/admin/login/deeplink?object_765_object
2. Save a new draft
3. Go to the Schedule tab and create a new item
4. Save only schedule tasks
5. Reload

blankse:
Resolves #10318 

Also fix following issue:
The Pimcore\Maintenance\Tasks\ScheduledTasksTask can not set the active property to false.
So every maintenance execute will publish all active tasks again and again.

It is because the save method will generate a new id for every task and so the save of the task doesn't work.

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1783,18 +1783,16 @@ class Asset extends Element\AbstractElement
 
     private function saveScheduledTasks()
     {
-        $this->getScheduledTasks();
-        $this->getDao()->deleteAllTasks();
-
-        if (is_array($this->getScheduledTasks()) && count($this->getScheduledTasks()) > 0) {
-            foreach ($this->getScheduledTasks() as $task) {
-                $task->setId(null);
-                $task->setDao(null);
-                $task->setCid($this->getId());
-                $task->setCtype('asset');
-                $task->save();
-            }
+        $scheduledTasks = $this->getScheduledTasks();
+        $ignoreIds = [];
+        foreach ($scheduledTasks as $task) {
+            $task->setDao(null);
+            $task->setCid($this->getId());
+            $task->setCtype('asset');
+            $task->save();
+            $ignoreIds[] = $task->getId();
         }
+        $this->getDao()->deleteAllTasks($ignoreIds);
     }
 
     /**
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -31,6 +31,7 @@ use Pimcore\Model\Asset\Listing;
 use Pimcore\Model\Asset\MetaData\ClassDefinition\Data\Data;
 use Pimcore\Model\Asset\MetaData\ClassDefinition\Data\DataDefinitionInterface;
 use Pimcore\Model\Element\ElementInterface;
+use Pimcore\Model\Element\Traits\ScheduledTasksTrait;
 use Pimcore\Model\Exception\NotFoundException;
 use Pimcore\Tool;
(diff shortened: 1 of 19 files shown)


models/Schedule/Task.php
```diff
@@ -28,42 +28,42 @@ class Task extends Model\AbstractModel
     /**
      * @var int
      */
-    protected $id;
+    public $id;
```
brusch: I think we should leave them `protected` and use `getObjectVars()` instead, in the controller.

models/Document/Dao.php
```diff
@@ -322,10 +322,16 @@ public function deleteAllPermissions()
 
     /**
      * Deletes all scheduled tasks assigned to the document.
+     *
+     * @param int[] $ignoreIds
      */
-    public function deleteAllTasks()
+    public function deleteAllTasks(array $ignoreIds = [])
```
brusch: Should we maybe move this function to a trait, similar we do with `VersionDaoTrait`?

models/Asset.php
```diff
@@ -1783,18 +1783,16 @@ public function setScheduledTasks($scheduledTasks)
...
-        if (is_array($this->getScheduledTasks()) && count($this->getScheduledTasks()) > 0) {
-            foreach ($this->getScheduledTasks() as $task) {
-                $task->setId(null);
-                $task->setDao(null);
-                $task->setCid($this->getId());
-                $task->setCtype('asset');
-                $task->save();
-            }
-        }
+        $scheduledTasks = $this->getScheduledTasks();
```
brusch: I think we should move the existing `ScheduledTasksTrait` to `Pimcore\Model\Element\Traits` and use it for all elements.

brusch (changes requested):

models/Schedule/Task.php
blankse: See https://github.com/pimcore/pimcore/pull/10320/commits/bbd11a4e66fe83da4271eef142e83e60b33ec852

models/Document/Dao.php
blankse: See https://github.com/pimcore/pimcore/pull/10320/commits/305e744f98535945e427f478b96ed5e513f540a6

models/Asset.php
blankse: See https://github.com/pimcore/pimcore/pull/10320/commits/1f548009cd3c1b1357eaab98667609b3270b226e

models/Element/Traits/ScheduledTasksTrait.php
```diff
@@ -59,20 +61,18 @@ public function setScheduledTasks($scheduledTasks)
         return $this;
     }
 
-    public function saveScheduledTasks()
+    public function saveScheduledTasks(): void
```
brusch:
```suggestion
    public function saveScheduledTasks()
```
Adding the return type is a BC break.

brusch (changes requested):

brusch:
Well done - thanks a lot! 👍

dasraab:
i think this is also an issue in v6.9.6 - at least i'm experiencing the same problem. scheduled task do not get set to active=false after they ran.

PR #10324, opened 2021-09-15
Improve URL Slug section of 65_Others.md

NiklasBr:
Code example better aligns with img/classes-datatypes-urlslug.png and info on validation and tip on what to use to generate ones.


brusch:
Thanks a lot!

PR #10325, opened 2021-09-15
Hotfix get options from provider

### LINKED ISSUE #9991: [Bug]: Classes-rebuild fails for class dependent optionprovider

roland4432:
Resolves #9991 

## Additional info  
Not much description needed. Use Throwable instead of Exception to catch errors from non-existing classes.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Multiselect.php
+++ b/models/DataObject/ClassDefinition/Data/Multiselect.php
@@ -707,7 +707,13 @@ class Multiselect extends Data implements
             $context = [];
             $context['fieldname'] = $this->getName();
 
-            $options = $optionsProvider->getOptions($context, $this);
+            try{
+                $options = $optionsProvider->getOptions($context, $this);
+            }
+            catch (\Exception $e){
+                // error from getOptions => no values => no comma => no problems
+                $options = null;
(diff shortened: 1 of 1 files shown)


brusch:
Thanks a lot!

PR #10326, opened 2021-09-15
Use Service::getElementType() and trigger deprecated warning

blankse:


brusch:
Very nice, thanks a lot!

models/Element/Service.php
```diff
@@ -591,6 +591,12 @@ public static function getElementHash(ElementInterface $element): ?string
      */
     public static function getType($element)
     {
+        trigger_deprecation(
+            'pimcore/pimcore',
+            '10.0',
```
jdreesen, line 596: Shouldn't this be `10.2`? 
Docs say: `@param string $version The version of the package that introduced the deprecation`

blankse: Is deprecated since 10.0

brusch: yep, just `trigger_deprecation()` was missing 😉

PR #10329, opened 2021-09-15
Symfony Messenger for Internal Background Tasks and Maintenance Jobs.

dpfaffenbauer:
This PR does 2 things.

1. It removes internal commands to Symfony Messenger (VideoConvert and Newsletter Sending)
2. It lets the maintenance Tasks run as Messenger Tasks. (Not required, the Tasks could still be ran with CLI)

Let me know what you think so far @brusch


brusch:
Looks already very promising to me. 
Just a few things: 
- How should we deal with running `messenger:consume` since, upgrading would require to setup the workers? Should we maybe spin them up using the maintenance (when running without a special flag, e.g. `--no-messenger` - not ideal, but allows smooth upgrades). 

Example: 
```bash
bin/console pimcore:maintenance # automatically starts 2 workers, with a time limit.
bin/console pimcore:maintenance --no-messenger # is used when having setup the workers properly
```

- I think for most cases we can also get away from using `TmpStore` and just put the necessary information directly into the message object, which makes it easier to read/maintain/understand of what's going on.

dpfaffenbauer:
> How should we deal with running messenger:consume since, upgrading would require to setup the workers? Should we maybe spin them up using the maintenance (when running without a special flag, e.g. --no-messenger - not ideal, but allows smooth upgrades).

Asked myself the very same question. The Problem is: the messenger is not a scheduler. So something has to trigger the maintenance tasks to be added to the queue and something needs to process the queue. One idea I worked around with is: to change how tasks are executed. if every task itself is adding itself to the queue again with a delay of a set amount of time, we could continuously run the queue and it also functions as sort of a scheduler. 

The other idea would be similar to your suggestion: add a new parameter for the maintenance command that first of all, adds the maintenance tasks to the queue, but then also processes these directly. not sure how what the best approach is here.

> I think for most cases we can also get away from using TmpStore and just put the necessary information directly into the message object, which makes it easier to read/maintain/understand of what's going on.

I don't think so. The tmpStore is needed by the Controller triggered the task. for example: the NewsletterController needs it to know what the status about sending is. With the Queue itself, I don't know if we can query it for the current Status within the queue. As well as updating the entry in the queue with the new status.

brusch:
@dpfaffenbauer 

Workers: I'd go for 
> maintenance command that first of all, adds the maintenance tasks to the queue, but then also processes these directly.   

But this shouldn't be an additional flag, but the default behavior, so that no changes are required after the update. We have then already the `--async` option that disables this default behavior. 

TmpStore: 
Your're right, for certain things it makes sense to keep the TmpStore. But for certain things like generating the system preview thumbnail or similar simple things we could get around the TmpStore 😊

dpfaffenbauer:
@brusch updated as discussed

Also added Upgrade Notes
Also added Docs for messenger:consume command

bundles/CoreBundle/Command/MaintenanceCommand.php
```diff
@@ -80,8 +86,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            $validJobs,
+            $excludedJobs,
+            (bool)$input->getOption('force'),
+            $async
+        );
+
+        if (!$async) {
+            trigger_deprecation(
+                'pimcore/pimcore',
+                '10.1',
```
dvesh3: If this is for 10.1, then we need to change the base branch to 10.1 😉

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,5 +1,8 @@
 # Upgrade Notes
 
+## 10.1.4
+- [Maintenance] Maintenance Commands are now handled with Symfony Messenger. The `pimcore:maintenance` command will trigger the maintenance messages to the bus. You should now also add `bin/console messenger:consume pimcore_core pimcore_maintenance` as a Cron Job
```
dvesh3: Is it default behavior? or you should use `--async` option to use Symfony Messenger?

bundles/CoreBundle/Command/MaintenanceCommand.php
jdreesen: This is more of a new feature than a bugfix, no? So it should go into the next 10.x

brusch: Absolutely :)

brusch:
```suggestion
                '10.2',
```

bundles/CoreBundle/Command/MaintenanceCommand.php
```diff
@@ -80,8 +86,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            $excludedJobs,
+            (bool)$input->getOption('force'),
+            $async
+        );
+
+        if (!$async) {
+            trigger_deprecation(
+                'pimcore/pimcore',
+                '10.2',
+                'Running Maintenance Command without --async and not having the messenger consume message yourself is deprecated and will be removed from Pimcore in 10.2.',
```
dvesh3: @brusch
```suggestion
                '10.2',
                'Running Maintenance Command without --async and not having the messenger consume message yourself is deprecated and will be removed from Pimcore in 11.',
```

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,5 +1,8 @@
 # Upgrade Notes
 
+## 10.1.4
```
```suggestion
## 10.2.0
```

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
brusch:
```suggestion
- [Maintenance] Maintenance tasks are now handled with Symfony Messenger. The `pimcore:maintenance` command will add the maintenance messages to the bus and runs them afterwards immediately from the queue. However it's recommended to setup independent workers that process the queues, by running `bin/console messenger:consume pimcore_core pimcore_maintenance` (using e.g. Supervisor) and adding `--async` option to the `pimcore:maintenance` command that stops the maintenance command to process the queue directly. Details about setting it up for production environments, please check [Symfony Messenger Component docs](https://symfony.com/doc/current/messenger.html#deploying-to-production). 
```

bundles/CoreBundle/Command/MaintenanceCommand.php
```diff
@@ -80,8 +86,33 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                'pimcore/pimcore',
+                '10.2',
+                'Running Maintenance Command without --async and not having the messenger consume message yourself is deprecated and will be removed from Pimcore in 11.',
+                __CLASS__
+            );
+
+            $command = $this->getApplication()->find('messenger:consume');
+
+            $arguments = [
+                'receivers' => ['pimcore_core'],
```
brusch:
```suggestion
                'receivers' => ['pimcore_core', 'pimcore_maintenance'],
```
We should also run the `pimcore_maintenance` immediately.

doc/Development_Documentation/01_Getting_Started/00_Installation.md
```diff
@@ -71,6 +71,7 @@ have a look at the logs as a starting point when debugging installation issues.
 
 ```bash
 */5 * * * * /your/project/bin/console pimcore:maintenance
+*/5 * * * * /your/project/bin/console messenger:consume pimcore_core pimcore_maintenance --time-limit=300
```
brusch, line 78:
```suggestion

# it's recommended to run the following command using a process control system like Supervisor
# please follow the Symfony Messenger guide for a best practice production setup: 
# https://symfony.com/doc/current/messenger.html#deploying-to-production
*/5 * * * * /your/project/bin/console messenger:consume pimcore_core pimcore_maintenance --time-limit=300
```

lib/Maintenance/ExecutorInterface.php
```diff
@@ -20,14 +20,16 @@
...
+
     /**
      * Execute the Maintenance Task
      *
      * @param array $validJobs
      * @param array $excludedJobs
      * @param bool $force
      */
-    public function executeMaintenance(array $validJobs = [], array $excludedJobs = [], bool $force = false);
+    public function executeMaintenance(array $validJobs = [], array $excludedJobs = [], bool $force = false, bool $async = false);
```
brusch:
```suggestion
    public function executeMaintenance(array $validJobs = [], array $excludedJobs = [], bool $force = false);
```
We don't need the `$async` status here, as we're always adding the tasks to the queue.

lib/Maintenance/Executor.php
```diff
@@ -44,69 +46,83 @@ final class Executor implements ExecutorInterface
...
 
-            $lock = $this->lockFactory->createLock('maintenance-' . $name, 86400);
-
-            if (!$lock->acquire() && !$force) {
-                $this->logger->info('Skipped job with ID {id} because it already being executed', [
-                    'id' => $name,
-                ]);
-
-                continue;
+            if ($async) {
```
brusch: This `if` is not necessary, we're always dispatching to the bus.

lib/Maintenance/Executor.php
```diff
@@ -44,69 +46,83 @@ final class Executor implements ExecutorInterface
...
-                $this->logger->info('Finished job with ID {id}', [
-                    'id' => $name,
-                ]);
-            } catch (\Exception $e) {
-                $this->logger->error('Failed to execute job with ID {id}: {exception}', [
-                    'id' => $name,
-                    'exception' => $e,
-                ]);
+            else {
+                $this->executeTask($name, $force);
```
brusch: This can then also go 😉 
Unfortunately it doesn't let me create a suggestion.

lib/Maintenance/Executor.php
```diff
@@ -44,69 +46,83 @@ final class Executor implements ExecutorInterface
...
+    public function executeTask(string $name, bool $force = false)
+    {
+        if (!in_array($name, $this->getTaskNames(), true)) {
+            throw new \Exception(sprintf('Task with name "%s" not found', $name));
+        }
+
+        $task = $this->tasks[$name];
+        $lock = $this->lockFactory->createLock('maintenance-' . $name, 86400);
+
+        if (!$lock->acquire() && !$force) {
```
brusch, line 70: Shouldn't we leave the `$force` check in `executeMaintenance()`. There's no need to add them to the queue if not necessary, right?

brusch (changes requested):

bundles/CoreBundle/Command/MaintenanceCommand.php
dpfaffenbauer: this runs immediately. checkout the task executor

dpfaffenbauer: nevermind, just read the other comment ;)

lib/Maintenance/Executor.php
dpfaffenbauer: the task could be added to the queue from other places too... so I thought the TaskHandler/executeTask should be responsible for it

bundles/CoreBundle/Command/MaintenanceCommand.php
brusch: I thought we could even let `messenger:consume` process the maintenance tasks if running in legacy mode, since we're calling `messenger:consume` anyway. But yes, the current implementation works as well.

brusch:
@dpfaffenbauer thanks a lot - great work 👍

dpfaffenbauer:
🍻 🚀

PR #10331, opened 2021-09-16
Update `codeception/module-symfony`

TavoNiievez:
There are no **b**reaking **c**hanges between version 1.x and version 2.x if you were already using Symfony 4.4 or higher, so it should be safe to upgrade.

See https://github.com/Codeception/module-symfony/releases .

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -167,7 +167,7 @@
   "require-dev": {
     "cache/integration-tests": "dev-master",
     "codeception/codeception": "^4.1.12",
-    "codeception/module-symfony": "^2.0.0",
+    "codeception/module-symfony": "^2.0.2",
     "codeception/phpunit-wrapper": "^9",
     "phpstan/phpstan": "^0.12.85",
     "phpstan/phpstan-symfony": "^0.12.32",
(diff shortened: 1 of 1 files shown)


brusch:
@TavoNiievez seems to cause a small problem, see: https://github.com/pimcore/pimcore/pull/10331/checks?check_run_id=3619279351#step:11:1078 

Can you please check? 
Thanks a lot!

TavoNiievez:
Related to https://github.com/Codeception/module-symfony/pull/129 in `v2.0.2`.

TavoNiievez:
@brusch The change that is generating this error was introduced in version `2.0.2`, so version `2.0.0` and `2.0.1` are not affected.

Setting the requirement to `^2.0.2` will cause composer to download the latest version, in which the problem is still present.

really the solution is that someone send a PR to `module-symfony` partially reversing the changes of this PR  https://github.com/Codeception/module-symfony/pull/129/files , especially the line 157:

```php
unset($parameters['doctrine.connections']);
```

or that the Pimcore tests adapt to work with those changes.

I'm not very familiar with testing in Pimcore so I can't be of much help there, but I can quickly merge a patch for this if someone submits it.

brusch:
I thought it was fixed in 2.0.2, that's why I've set the constraint to `^2.0.2` 😉
Can't see a quick way how to fix that, to be honest, for us it doesn't have priority, so if you are not continue to work on that PR I'd close it for now.

PR #10332, opened 2021-09-16
Remove redundant scheduledTasks properties

blankse:
Followup to https://github.com/pimcore/pimcore/pull/10320

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -207,13 +207,6 @@ class Asset extends Element\AbstractElement
     /**
      * @internal
      *
-     * @var array|null
-     */
-    protected $scheduledTasks = null;
-
-    /**
-     * @internal
-     *
      * @var bool
      */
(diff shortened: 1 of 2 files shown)


brusch:
Totally missed them as well 😊 Thanks a lot!

dasraab:
i think you can also delete [here](https://github.com/pimcore/pimcore/blob/5f9c86682639403f22475bf5dc11758d962ba9fd/models/Asset.php#L1896) and [here](https://github.com/pimcore/pimcore/blob/5f9c86682639403f22475bf5dc11758d962ba9fd/models/DataObject/Concrete.php#L825)

blankse:
@dasraab The property exist in the ScheduledTasksTrait and I think it should be set to null by clone. WDYT @brusch ?

brusch:
Yep, that should stay as it is, as this isn't covered by the trait.

PR #10336, opened 2021-09-16
Fix deprecations

blankse:
- Server::exec() -> Server::start()
- Link::getObject() -> Link::getElement() + trigger_deprecation
- Link::setObject() -> Link::setElement() + trigger_deprecation

### DIFF
--- a/models/Document/Link.php
+++ b/models/Document/Link.php
@@ -354,7 +354,7 @@ class Link extends Model\Document
         trigger_deprecation(
             'pimcore/pimcore',
             '10.0',
-            'The Link::setObject() method is deprecated, use Link::setObject() instead.'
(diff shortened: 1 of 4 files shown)


models/Document/Link.php
```diff
@@ -340,6 +351,12 @@ public function getObject()
      */
     public function setObject($object)
     {
+        trigger_deprecation(
+            'pimcore/pimcore',
+            '10.0',
+            'The Link::setObject() method is deprecated, use Link::setObject() instead.'
```
NiklasBr:
```suggestion
            'The Link::setObject() method is deprecated, use Link::setElement() instead.'
```

brusch:
Very nice, thanks a lot!

PR #10337, opened 2021-09-16
Fixed getMasterRequest() deprecations

brusch:


blankse:
@brusch getMainRequest doesn't exists in symfony/http-foundation 5.2.0
Is part of 5.3.0. So you must bump composer, right?

brusch:
@blankse you're totally right - good catch! Thanks!

blankse:
@brusch I wonder why the phpstan test lowest also get 5.3. Seems to be impossible to get 5.2.0 now.

Edit: symfony/security-bundle is 5.3 and require  symfony/http-foundation 5.3. So it can't be 5.2

brusch:
@blankse true, but should be fixed anyway 😉

lib/Http/RequestHelper.php
```diff
@@ -82,17 +82,32 @@ public function getRequest(Request $request = null)
      */
     public function hasMasterRequest(): bool
```
jdreesen, line 83: I'd say this should be deprecated and replaced by `hasMainRequest()`, too.

brusch: Absolutely 😊 See b3fd6ee1370c764478daaf1a50fe66632ec6067d

PR #10348, opened 2021-09-17
Fix deprecation in league/csv

blankse:
## Additional info  

league/csv add PHP 8 support in 9.6.2: https://github.com/thephpleague/csv/releases/tag/9.6.2
So it should bumped anyway.


brusch:
Thank you very much!

PR #10349, opened 2021-09-19
function is not available in document

archanagoyal07:
function assignVariablesToModel is called in model file. but not define anywhere.

  
  

Resolves #


jdreesen:
The code example doesn't seem to be complete.

archanagoyal07:
The code is working in my local project.

Well it's up to you, you can test and merge branch.

Thanks
Archana

brusch:
Hmm, `assignVariablesToModel()` is available on `\Pimcore\Model\Dao\AbstractDao` so shouldn't be necessary. If you need custom functionality there, you can of course override it - but that just plain PHP 😉

archanagoyal07:
Ok, but this was giving error when it called in model function. So I
thought that it is not available. Well thanks for the information. I will
check this.

PR #10350, opened 2021-09-20
[Predefined Asset metadata] default action for "Add predefined definitions" button not working as expected

mcop1:
Right now, if you directly click on the button no metadata definitions are added. 
#9019 mentions that all metadata defintions should be added in this case.


weisswurstkanone:
thx!

PR #10351, opened 2021-09-20
Add checkbox to disable/enable to create new object in advanced many to many object relation

MarcW09:
Add checkbox for disabling or enabling to create new objects in advanced many to many relation admin form.


weisswurstkanone:
Currently enabled by default but no way to tick it off.

@MarcW09 Merci!

Issue #9330, opened 2021-06-04
[Core] Config Deployment Improvements


brusch:
Deployments vs. local configs
See also https://github.com/pimcore/data-hub/issues/357#issuecomment-842412425

- [x] custom reports 
- [x] document types 
- [x] predefined properties
- [x] predefined asset meta-data
- [x] thumbnails
- [x] web2print config
- [x] static routes (no settings-store)
- [x] perspectives
- [x] custom views

brusch:
This also fixes #6987

jdreesen:
Is this already completely done? It seems it was automatically closed by mistake.

jdreesen:
Aaand it's closed again without being completely done :wink:

brusch:
@jdreesen 😁 Thanks

jdreesen:
<h3>:partying_face:</h3>

PR #10352, opened 2021-09-20
[Core] Config Deployment Improvements predefined asset metadata

mcop1:
part of #9330

### DIFF
--- a/models/Metadata/Predefined/Dao.php
+++ b/models/Metadata/Predefined/Dao.php
@@ -31,7 +31,7 @@ class Dao extends Model\Dao\PimcoreLocationAwareConfigDao
         $config = \Pimcore::getContainer()->getParameter("pimcore.config");
 
         parent::configure([
-            'containerConfig' => $config['metadata']['predefined']['definitions'] ?? [],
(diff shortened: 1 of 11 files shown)


models/Metadata/Predefined/Dao.php
```diff
@@ -22,16 +22,25 @@
...
+    use Model\Dao\AutoIncrementTrait;
+
     public function configure()
     {
-        parent::configure();
-        $this->setFile('predefined-asset-metadata');
+        $config = \Pimcore::getContainer()->getParameter("pimcore.config");
+
+        parent::configure([
+            'containerConfig' => $config['metadata']['predefined']['definitions'] ?? [],
```
weisswurstkanone: As this is only asset specific I would assume that this should go into the assets namespace similar to
            'containerConfig' => $config['documents']['doc_types']['definitions'] ?? [],

wdyt ?

mcop1: Since metadatas are used in assets only, I would say you are right. I will change the namespace and data structure accordingly.

weisswurstkanone:
thanks a lot!

PR #10356, opened 2021-09-20
Fixes Key/name labels on Properties

NiklasBr:
Fixes a mismatch where currently what is called "Name" on the element is actually the "Key" of the property:

image

I could make a follow-up patch to also add the actual "Name" column on the element as well.


brusch:
Thanks a lot! 👍
Yes, I think it would make sense to display the `Name` as well if available (similar to `Description`). Would  be nice if you could provide a PR for that. Thanks in advance! 😊

PR #10367, opened 2021-09-21
Fix Bug: Object Scheduled Tasks

### LINKED ISSUE #10365: [Bug]: Object Scheduled Tasks
### Expected behavior

When you create schedule task on some object and reload this object, you expect to see same information you enter in form

### Actual behavior

After creation schedule task, entry is stored into database, but after reloading object, server return empty object for created task in array scheduledTasks and grid can not show real information

### Steps to reproduce

1. Go to public demo https://demo.pimcore.fun/admin
2. open some object (_e.g. News_)
3. Go to tab Schedule
4. Create one entry for schedule
5. Save object
6. Reload object

> You will see one entry with empty fields (_except date which show 1970-01-01_); Also in network yo can see empty object in `scheduledTasks` image (20)

blankse:
Resolves #10365
Followup to #10320

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -33,6 +33,7 @@ use Pimcore\Model\DataObject\ClassDefinition\Data\ManyToManyObjectRelation;
 use Pimcore\Model\DataObject\ClassDefinition\Data\Relations\AbstractRelations;
 use Pimcore\Model\DataObject\ClassDefinition\Data\ReverseObjectRelation;
 use Pimcore\Model\Element;
+use Pimcore\Model\Schedule\Task;
 use Pimcore\Model\Version;
 use Pimcore\Tool;
 use Symfony\Component\EventDispatcher\GenericEvent;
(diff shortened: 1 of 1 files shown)


gjorgic:
Hi @blankse  I check this and now i can see details in grid, but there is issue with time; each time when i save and reload object, time is increased

image

After saving and reloading i see this

image

blankse:
@gjorgic I havent't that problem. Whats your "Default timezone" and "date.timezone" values in the php.ini? Should be UTC I guess.

gjorgic:
My "Default timezone" is UTC, but "date.timezone" no value

However, this pull request solve the issue I reported  #10365
 :+1: Thanks @blankse

dvesh3:
@gjorgic please make sure to set correct timezone in `config.yaml` for Pimcore to show date times correctly:
```
pimcore:
   general:
       timezone: Europe/Berlin
```

dvesh3:
@blankse thanks a lot!

blankse:
@gjorgic @dvesh3 I see we also set the timezone in the config.yaml. So it makes a different, if the client use a other timezone? Maybe we must set the timezone in JS for the date?

gjorgic:
~@dvesh3 is there any chance to merge/cherry-pick this fix into 10.0?~
_Ignore this_

PR #10376, opened 2021-09-22
Fix invisible field inside block element

dturopoli:
When env is set to DEV an error appears when trying to save an object that has block element with invisible field (test1 is invisible field)

Screenshot from 2021-09-22 10-51-58

Also when you save an object that has invisible field inside block element using Master (admin mode) 
Screenshot from 2021-09-22 10-57-16
value inside invisible field wouldn't be updated.


weisswurstkanone:
Thx!

PR #10377, opened 2021-09-22
Versions -> Delete empty folders

### LINKED ISSUE #7717: Versions -> Delete empty folders
Currently when versions are deleted only the version file get removed but the directory of the version dosn't get deleted. 
Over the time this can cause a lot of disc usage. 

Tested it with 2 Projekt:
First:
Version directory had 8,3GB and after "find ./var/versions/ -type d -empty -delete" it has only 205MB.
Second: 
Version directory had 125GB and after "find ./var/versions/ -type d -empty -delete" it has only 111GB. (3.6Million empty folders deleted)

weisswurstkanone:
resolves #7717

### DIFF
--- a/lib/File.php
+++ b/lib/File.php
@@ -15,6 +15,8 @@
 
 namespace Pimcore;
 
+use League\Flysystem\FilesystemOperator;
+
 class File
 {
     /**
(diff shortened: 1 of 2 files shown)


mcop1:
thx 👍

PR #10381, opened 2021-09-22
[MessageBus] Fixed public service/alias compiler pass

brusch:
Follow-up to #10329

### DIFF
--- a/bundles/CoreBundle/DependencyInjection/Compiler/MessageBusPublicPass.php
+++ b/bundles/CoreBundle/DependencyInjection/Compiler/MessageBusPublicPass.php
@@ -28,6 +28,6 @@ final class MessageBusPublicPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
-        $container->findDefinition(MessageBusInterface::class)->setPublic(true);
+        $container->getAlias(MessageBusInterface::class)->setPublic(true);
     }
 }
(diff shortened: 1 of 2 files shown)


bundles/CoreBundle/DependencyInjection/Compiler/MessageBusPublicPass.php
```diff
@@ -28,6 +28,6 @@ final class MessageBusPublicPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
-        $container->findDefinition(MessageBusInterface::class)->setPublic(true);
+        $container->getAlias(MessageBusInterface::class)->setPublic(true);
```
dpfaffenbauer, line 31: use `find` instead, checks for definition or alias

brusch: True, will check that, thanks!

brusch: `findDefinition()` is not an option, it returns the service, but `setPublic()` on the alias doesn't work (only on the target service, which is `messenger.default_bus`).

dpfaffenbauer: alright

PR #10388, opened 2021-09-23
Fix batch edit in grid when using object brick filter,  resolves #10380

### LINKED ISSUE #10380: [Bug]: Batch Edit All raises Exception with active filters and Object Bricks
### Expected behavior

Pimcore 10.1.3
Also verified in 6.8.12, 10.x-dev

- Configure a grid with some custom columns from an object brick
- Filter the grid 
- Click any column and select "Batch Edit All"
- All filtered records get updated

### Actual behavior

- Configure a grid with some custom columns from an object brick
- Filter the grid 
- Click any column and select "Batch Edit All"
- /admin/object-helper/get-batch-jobs raises a 500 because the generated SQL cannot handle the Object Brick Fields in the relation
- 
image

### Steps to reproduce

- Log in to https://demo.pimcore.com/
- Open any folder (e.g. Alfa Romeo) 
- Open Grid Options and add some Object Brick columns (SaleInformation > Condition)
- Filter for Condition "reworked"
- Choose any column and click "Batch Edit All" in the Column menu

dturopoli:
Resolves #10380

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/element/helpers/gridColumnConfig.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/element/helpers/gridColumnConfig.js
@@ -779,6 +779,10 @@ pimcore.element.helpers.gridColumnConfig = {
             params["only_unreferenced"] = this.checkboxOnlyUnreferenced.getValue();
         }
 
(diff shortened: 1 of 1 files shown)


dvesh3:
@dturopoli thanks a lot!

PR #10389, opened 2021-09-23
[BUG] Listing with offset and limit

Gerhard-Kanzler:
Listing with offset and limit get always limit of 1 so you need brackets if there is multiple assignments in if statements


jdreesen (approved):

dvesh3:
pushed directly on 10.1 branch https://github.com/pimcore/pimcore/commit/94828e1816cb4c6e43499fd802b72ccfa9cf9537 thanks!

Issue #6790, opened 2020-07-15
[Asset thumbnails] Inverted colors for CMYK images


BlackbitNeueMedien:
We get inverted colors in the thumbnails for some images, e.g. https://www.dropbox.com/s/rhdzn76wtosu5tj/10555J__01.tif.zip?dl=0 (had to zip it, otherwise Dropbox had an error - for testing please unzip it)

Obviously in https://github.com/pimcore/pimcore/blob/b17b05a84a3946fbc1dc5eb88b0fc33682de2690/lib/Image/Adapter/Imagick.php#L319-L331 the comment https://www.php.net/manual/en/imagick.setimagecolorspace.php#107716 has been applied - but with the addition of setting the colorspace to sRGB:
https://github.com/pimcore/pimcore/blob/b17b05a84a3946fbc1dc5eb88b0fc33682de2690/lib/Image/Adapter/Imagick.php#L330

With above image I get this result in the default asset preview in Pimcore backend
![get-image-thumbnail](https://user-images.githubusercontent.com/8749138/87510852-9bc36000-c674-11ea-8ec9-2be0a3701c99.jpg)

When I remove https://github.com/pimcore/pimcore/blob/b17b05a84a3946fbc1dc5eb88b0fc33682de2690/lib/Image/Adapter/Imagick.php#L330
then it looks correctly:
![get-image-thumbnail](https://user-images.githubusercontent.com/8749138/87510911-beee0f80-c674-11ea-80b3-f9d146a10788.jpg)

On https://demo.pimcore.fun it nevertheless works also with this line. And also without. 
Do you know what this line is good for - or where thumbnail generation does not work properly without this line?

ImageMagick is version 6.9.11-23 Q16 x86_64 2020-07-14, Imagick 3.4.4 - so this is nearly the same as on https://demo.pimcore.fun (ImageMagick 6.9.10-23 Q16 x86_64 20190101, Imagick 3.4.3)
We have no additional color profiles set in system settings.

brusch:
This seems to be a known ImageMagick bug, see also: https://talk.pimcore.org/t/cmyk-to-rgb-displaying-as-negative/908/17?u=brusch

PR #10401, opened 2021-09-23
[System Requirements] Check for ImageMagick delegate LCMS

BlackbitDevs:
As already described in #6790 there can be a problem with negative colors when generating thumbnails from CMYK images. As proposed in https://talk.pimcore.org/t/cmyk-to-rgb-displaying-as-negative/908/17 and https://www.php.net/manual/en/imagick.profileimage.php#114781 this is caused by ImageMagick missing the `lcms` delegate. This PR adds a check in the system requirements for this delegate.

### DIFF
--- a/lib/Tool/Requirements.php
+++ b/lib/Tool/Requirements.php
@@ -19,6 +19,7 @@ use Pimcore\Db\ConnectionInterface;
 use Pimcore\File;
 use Pimcore\Image;
 use Pimcore\Tool\Requirements\Check;
+use Symfony\Component\Process\Process;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


lib/Tool/Requirements.php
```diff
@@ -615,6 +616,16 @@ public static function checkPhp()
             'state' => class_exists('Imagick') ? Check::STATE_OK : Check::STATE_WARNING,
         ]);
 
+        if(class_exists('Imagick')) {
+            $imageMagickLcmsDelegateInstalledProcess = Process::fromShellCommandline('convert -list configure | grep DELEGATES');
```
brusch: We should use `\Pimcore\Tool\Console::getExecutable()` to determine the path to `convert`.

lib/Tool/Requirements.php
```diff
@@ -615,6 +616,16 @@ public static function checkPhp()
...
+            $imageMagickLcmsDelegateInstalledProcess->run();
+            $checks[] = new Check([
+                'name' => 'ImageMagick LCMS delegate',
+                'link' => 'https://talk.pimcore.org/t/cmyk-to-rgb-displaying-as-negative/908/17',
```
brusch: I'd add a note about LCMS here: https://github.com/pimcore/pimcore/blob/7fff27a970182dc18717b3aa1727d29e388ba72d/doc/Development_Documentation/23_Installation_and_Upgrade/01_System_Requirements.md#L40 and change the link to the docs.

lib/Tool/Requirements.php
brusch: I'd prefer to not use `grep` but pure PHP.

brusch (changes requested):

lib/Tool/Requirements.php
BlackbitDevs: Done.

lib/Tool/Requirements.php
BlackbitDevs: Done.

brusch:
Very nice, thanks a lot! 👍

PR #10406, opened 2021-09-23
[DataObjects] Fix Table data type initial data 

### LINKED ISSUE #10405: When using table in object brick whole pimcore crashes
### Discussed in https://github.com/pimcore/pimcore/discussions/10396

<div type='discussions-op-text'>

<sup>Originally posted by **jdevries14** September 23, 2021</sup>
Screenshot 2021-09-23 at 12 05 50

This is my current object brick, and when selecting the object brick in my class it works fine, look:

Screenshot 2021-09-23 at 12 06 40

But when I add a table to my object brick and give it a name without any settings as following:
Screenshot 2021-09-23 at 12 07 24

Then when I choose this object brick, it crashes and I can't use Pimcore anymore. Also it doesn't give an exception or so.
Also all my fields etc disappear as following:

Screenshot 2021-09-23 at 12 07 59

I do get a console error:

Screenshot 2021-09-23 at 12 08 52

</div>

dvesh3:
Resolves #10405

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/table.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/table.js
@@ -21,7 +21,7 @@ pimcore.object.tags.table = Class.create(pimcore.object.tags.abstract, {
 
         this.fieldConfig = fieldConfig;
 
-        if (data.length < 1) {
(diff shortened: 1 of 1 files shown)


dvesh3:
closed in favor of #10400

PR #10419, opened 2021-09-27
Preview Generator Command: exclude & parent options

### LINKED ISSUE #8149: [Document] Preview Generator Command: exclude & parent options
It would be very helpful if you could specify a parent path for the generator (e.g. `/en`)  and additionally to exclude certain patterns (path regex).

mcop1:
Closes #8149

### DIFF
--- a/bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
+++ b/bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
@@ -120,7 +120,6 @@ class GeneratePagePreviews extends AbstractCommand
              * @var Document\Page $doc
              */
             try {
-                $this->writeInfo($doc->getPath() . " " . $doc->getRealFullPath());
                 $success = Document\Service::generatePagePreview($doc->getId(), null, $hostUrl);
             } catch (\Exception $e) {
                 $this->io->error($e->getMessage());
--- a/bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
+++ b/bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
@@ -84,7 +84,7 @@ class GeneratePagePreviews extends AbstractCommand
             }
         }
 
-        $regex = $input->getOption('regex');
+        $regex = $input->getOption('exclude-patterns');
         if($regex)
             $conditions[] = "path NOT REGEXP " . $db->quote($regex);
 
--- a/bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
+++ b/bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
@@ -53,7 +53,7 @@ class GeneratePagePreviews extends AbstractCommand
             )
             ->addOption(
                 'exclude-patterns',
-                'e',
+                'x',
                 InputOption::VALUE_OPTIONAL,
                 'Excludes all documents whose path property matches the regex pattern.'
             );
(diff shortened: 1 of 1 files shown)


bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
```diff
@@ -43,9 +44,60 @@ protected function configure()
                 InputOption::VALUE_OPTIONAL,
                 'Prefix for the document path, eg. https://example.com, if not specified, Pimcore will try use the main domain from system settings.',
                 null
+            )
+            ->addOption(
+                'parent',
+                'p',
+                InputOption::VALUE_OPTIONAL,
+                'Only create preview of objects in this folder (ID|path).'
```
dvesh3:
```suggestion
                'Only create preview of documents in parent hierarchy (ID|path).'
```

bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
```diff
@@ -43,9 +44,60 @@ protected function configure()
...
+            )
+            ->addOption(
+                'regex',
+                'r',
```
dvesh3: I think it is more clear
```suggestion
                'exclude-patterns',
                'e',
```

bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
```diff
@@ -43,9 +44,60 @@ protected function configure()
...
+                InputOption::VALUE_OPTIONAL,
+                'All child objects matching the regex pattern will be excluded.'
```
dvesh3:
```suggestion
                'All documents matching the regex pattern will be excluded.'
```

bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
```diff
@@ -43,9 +44,60 @@ protected function configure()
...
+            }
+            if ($parent instanceof Document) {
+                $conditions[] = "path LIKE " . $db->quote($parent->getRealFullPath() . "%");
+            } else {
+                $this->writeError($parentIdOrPath . ' is not a valid id or path!');
+                exit(1);
+            }
+        }
+
+        $regex = $input->getOption('regex');
```
dvesh3:
```suggestion
        $regex = $input->getOption('exclude-patterns');
```

bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
mcop1: The regex tests the path property, so maybe we should even clarify it further? Something like "Excludes all documents whose path property matches the regex pattern"?
What do you think?

dvesh3: yes, even better 👍

bundles/CoreBundle/Command/Document/GeneratePagePreviews.php
```diff
@@ -43,9 +44,60 @@ protected function configure()
...
+            )
+            ->addOption(
+                'parent',
+                'p',
+                InputOption::VALUE_OPTIONAL,
+                'Only create preview of documents in parent hierarchy (ID|path).'
+            )
+            ->addOption(
+                'exclude-patterns',
+                'e',
```
dvesh3: we cannot use option "e" as it is already being used for environment variable 
image
```suggestion
                'x',
```

PR #10426, opened 2021-09-27
Fix admin user not seeing all languages inside Shared translations

cuca24:
There is a bug when admin is not able to see all languages in shared translations
Steps to reproduce

1. Add at least 2 languages to System Settings
2. Create new user role and inside Shared translations settings for that role uncheck at least one of the languages
3. Create new user and assign newly created role from step 2.
4. Now that user can become an admin in future, so edit user and Check that he is admin
5. Login as user from step 2. Go to Shared Translations and this user won't be able to see the language that is unchecked from step 2. 

Expected behavior: Admin should always see all languages in shared translations no matter his previous role.

### DIFF
--- a/models/User.php
+++ b/models/User.php
@@ -745,6 +745,10 @@ final class User extends User\UserRole
      */
     public function getAllowedLanguagesForViewingWebsiteTranslations()
     {
+        if ($this->isAdmin()) {
+            return Tool::getValidLanguages();
+        }
+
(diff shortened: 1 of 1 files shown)


dvesh3:
@cuca24 Fixed already for 10.2 https://github.com/pimcore/pimcore/pull/10261. We can just cherry pick this change on 10.1.

dvesh3:
🍒 picked on 10.1 https://github.com/pimcore/pimcore/commit/551b74f7fc671be5d232a37547d80471706e2559

Issue #5824, opened 2020-02-17
CleanupBrickTables maintenance task does not recognize bricks with capitalization in name


efnack:
Maintenance task does not recognize bricks with capitalization in name causing ERROR in processing. 
Was this intended for some reason?

Ref: lib/Maintenance/Tasks/CleanupBrickTablesTask.php

Fixed locally

```php
private function getBrickDef($brickType) {
        $list = new Definition\Listing();
        $list = $list->load();
        
        foreach ($list as $brickDefinition) {
            if (strtolower($brickDefinition->getKey()) === $brickType) {
                return $brickDefinition;
            }
        }
}
```
And replace line 56
```php
$brickDef = $this->getBrickDef($brickType);
```

weisswurstkanone:
Unable to reproduce.

Example: 
- brick name "LazyLoadingTest"
- class: numbertest
- renamed field from "brickfield" to "brickfield2"
- executed command
- table "object_brick_store_LazyLoadingTest_30" got cleaned up properly

brusch:
@efnack see @weisswurstkanone answer ... can you please share some more details please? 
Thanks a lot!

weisswurstkanone:
Can you please explain:
- how is you brick called (was it renamed maybe)
- how it your class name
- which table causes the problem?

efnack:
If I reverse the changes and create a brick called TestBrick and save, it will create the tables
- object_brick_query_testbrick_5
- object_brick_store_testbrick_5

And when the maintenance task will give error:
10:52:37 ERROR     [app] Brick 'testbrick' not found. Please check table object_brick_store_testbrick_5
10:52:37 ERROR     [app] Brick 'testbrick' not found. Please check table object_brick_query_testbrick_5

As i've read the code it uses the table names to get brick definition. Is there a setting which makes table names keep capitalization of brick name as it does in your system?

Hope this clears things up

weisswurstkanone:
Mhm, if I create a brick called "HelloYou" it creates a table object_brick_store_HelloYou_[classid]

efnack:
Yes, I've found the issue. It's cause my development database is located on a Windows server, with `lower_case_table_names` set to 1. Which is default on Windows systems. So it will decapitalize, which will make the approch to checking if brick by key exists fails in the task.

Perhaps using the actual list of bricks as I suggest (or something more elegant) would be safer (or more universal) approch.

Thanks for the brainstorm :)

brusch:
[...]

PR #10428, opened 2021-09-27
Updated documentation on the setting lower_case_table_names.

### LINKED ISSUE #10427: [Bug]: CleanupBrickTables maintenance task does not recognize bricks with capitalization in name
### Expected behavior

CleanupBrickTables runs smoothly, without any errors.

### Actual behavior

I've managed to reproduce the problem from #5824.  
Our devops used [this manual](https://github.com/pimcore/pimcore/edit/10.x/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/05_DB_Setup.md) and added lower_case_table_names=1 in accordance with it.  
Unfortunately, he did it on the production DB, on Linux and on AWS, so I didn't even have access to mysql system settings and I wasn't aware of the changes.  
My next morning started with shouts of angry users, who were completely unable to work with the website. After some research, I changed the actual table names to lower case (luckily, I had access to mysqldump/mysql on production).  
So I appreciate the decision taken in Mysql 8 to disallow changing this setting if DB directory has already been initialized.  
But now we're getting the error with CleanupBrickTables mentioned in #5824, and I was able to reproduce it on a fresh installation, see the steps to reproduce.

### Steps to reproduce

Use Linux, set lower_case_table_names=1 in Mysql config. Create a new DataObject class, named with a capital letter. This creates tables with lower_case, but actual class files are still created case-sensitive. So CleanupBrickTables gets lower_cased Mysql table name and tries to look up a class with a lower_cased filename on a case-sensitive linux filesystem.
But even if we lower_case the filename of the class, datatype definitions (which are string literals in Pimcore) cannot be lower_cased - and Pimcore on Linux still tries to look them up in case-sensitive mode. It should not be a problem on Windows, though - datatype definitions and filenames can be CamelCased there, but CleanupBrickTables should still be able to get them, because the filesystem is case-insensitive there.

I've made a test using fresh & clean Pimcore installation, the result is shown [here](https://raw.githubusercontent.com/CountZero1981/pimcore/image/Untitled.png)

CountZero1981:
Resolves #10427, #5824

## Additional info  
I think that it makes sense to use `lower_case_table_names` only on Windows, where both PHP and Mysql filenames are case-insensitive and it's actually possible to make mess with two DataObject classes with different letter cases. And this comment is worth adding to this documentation.

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/05_DB_Setup.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/05_DB_Setup.md
@@ -1,9 +1,9 @@
-# Database Setup 
+# Database Setup
 
 Pimcore requires a standard MySQL database, the only thing you should assure is that the database uses `utf8mb4` as character set.  
 If you create a new database just set the character set to `utf8mb4`.
 
-> Note: You have to create the database manually before you can continue with the web-based installer, 
+> Note: You have to create the database manually before you can continue with the web-based installer,
 > which automatically creates the underlying database schema for Pimcore.
 
 ### Command to Create a new Database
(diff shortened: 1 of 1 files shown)


CountZero1981:
hmm... I see the problem... php functions are case insensitive... so if we create two classes, they might overlap...

CountZero1981:
I've tried to create two classes and objectbricks with different letter cases, Pimcore doesn't allow to do it. so I think that there're no reasons to use `lower_case_table_names` on Linux.

brusch:
OK, thanks for clarifying, I'll close this PR then.

PR #10458, opened 2021-09-29
IncludeRenderer and Snippet Editable cache should be kept with admin changes

blankse:
Steps to reproduce

1. Enable Full Page Cache with lifetime
```
    full_page_cache:
        enabled: true
        lifetime: 3600
        exclude_patterns: ''
        exclude_cookie: ''
```
2. Open a page with embeded snippet (Cache for Main Request and Snippets are created)
3. Change a element (Cache with Tag "output" will be deleted)
4. Open a other page with the same embedded snippet
-> Snippet will be new created but should still kept because of lifetime setting


brusch:
Makes totally sense, thanks a lot! 👌

PR #10460, opened 2021-09-29
GridConfig::getById() should return null instead of throw NotFoundException

blankse:
Should be consistent to the other models.
No BC break because its a internal model class.


brusch:
True, it's internal, so we can go 4 it - thanks man! 👍

PR #10473, opened 2021-09-30
Deadlock-preventing execution of all database write operations

BlackbitDevs:
Currently for example in https://github.com/pimcore/pimcore/blob/9a5b9adef0889af8fdade11ff3254ff80c47dd6a/models/DataObject/AbstractObject.php#L671-L748 we have an implementation which tries to prevent deadlocks but for example in https://github.com/pimcore/pimcore/blob/9a5b9adef0889af8fdade11ff3254ff80c47dd6a/models/Property/Dao.php#L56 we do not have such a logic which may cause errors like `SQLSTATE[40001]: Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction ...`. 
As MySQL already gives the hint to restart it, this PR implements deadlock-prevention for all database write operations.


BlackbitDevs:
Does not work this way because the transaction has to be restarted, not the query itself.

PR #10485, opened 2021-10-02
Add autoScroll to add scrollbar if necessary

### LINKED ISSUE #10371: [Class Definition] Missing horizontal scroller
### Expected behavior

There should be a horizontal scrollbar

### Actual behavior

Scrollbar missing. No way to open the dropdown for example.

image

### Steps to reproduce

Open class editor on small screen

matthiashamacher:
Resolves #10371 

## Additional Information
Also added autoScroll to rootPanel for field collections and object bricks as the issue is there also present.


weisswurstkanone:
Seems to fix it for the root node ("General Settings") but not for the nodes below it. @matthiashamacher , could you please check again? Thx!

matthiashamacher:
@weisswurstkanone You are right. I've added the autoScroll config to the data and layout components so that they are also scrollable.

weisswurstkanone:
Merci, double scrollbars feel a bit like an obstacle. However, I believe this would be a nice improvement as before one wasn't even able to edit the data on a small screen. Thanks a lot @matthiashamacher 👍 

image

PR #10486, opened 2021-10-04
Add foreign keys to data object tables

BlackbitDevs:
From time to time we have the problem that there is an item in the `object_store_*` table but not in `objects`. As there is a unique data field it is then neither possible to create a new item with the same value, nor to find the existing one (because of INNER JOINS on the `objects` table. I do not know the reason, perhaps there is a transaction missing somewhere but anyway, it would be cleaner if the data which belongs together would be explicitly connected with a foreign key. Then such situations could not occur anymore, no matter how one of the datasets belonging to the same object got removed / not created.

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20211018104331.php
+++ b/bundles/CoreBundle/Migrations/Version20211018104331.php
@@ -45,6 +45,9 @@ final class Version20211018104331 extends AbstractMigration
 
                 $fkName = 'fk_'.$table.'__'.$objectIdColumn;
                 if (!$tableSchema->hasForeignKey($fkName)) {
+                    if($tableSchema->getColumn($objectIdColumn)->getCustomSchemaOption('unsigned') === false) {
+                        $tableSchema->changeColumn($objectIdColumn, ['unsigned' => true]);
+                    }
                     $tableSchema->addForeignKeyConstraint('objects', [$objectIdColumn], ['o_id'], ['ON DELETE CASCADE'], $fkName);
                 }
             }
--- a/bundles/CoreBundle/Migrations/Version20211018104331.php
+++ b/bundles/CoreBundle/Migrations/Version20211018104331.php
@@ -46,7 +46,8 @@ final class Version20211018104331 extends AbstractMigration
 
                 $fkName = AbstractDao::getForeignKeyName($table, $objectIdColumn);
                 if (!$tableSchema->hasForeignKey($fkName)) {
-                    if($tableSchema->getColumn($objectIdColumn)->getCustomSchemaOption('unsigned') === false) {
+                    $column = $tableSchema->getColumn($objectIdColumn);
+                    if(!$column->hasCustomSchemaOption('unsigned') || $column->getCustomSchemaOption('unsigned') === false) {
                         $tableSchema->changeColumn($objectIdColumn, ['unsigned' => true]);
                     }
 
--- a/bundles/CoreBundle/Migrations/Version20211018104331.php
+++ b/bundles/CoreBundle/Migrations/Version20211018104331.php
@@ -20,6 +20,7 @@ namespace Pimcore\Bundle\CoreBundle\Migrations;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\SchemaException;
 use Doctrine\Migrations\AbstractMigration;
+use Pimcore\Model\Dao\AbstractDao;
 use Pimcore\Model\DataObject;
 
 final class Version20211018104331 extends AbstractMigration
@@ -43,11 +44,12 @@ final class Version20211018104331 extends AbstractMigration
                     continue;
                 }
 
-                $fkName = 'fk_'.$table.'__'.$objectIdColumn;
+                $fkName = AbstractDao::getForeignKeyName($table, $objectIdColumn);
                 if (!$tableSchema->hasForeignKey($fkName)) {
                     if($tableSchema->getColumn($objectIdColumn)->getCustomSchemaOption('unsigned') === false) {
                         $tableSchema->changeColumn($objectIdColumn, ['unsigned' => true]);
                     }
+
                     $tableSchema->addForeignKeyConstraint('objects', [$objectIdColumn], ['o_id'], ['ON DELETE CASCADE'], $fkName);
                 }
             }
--- a/bundles/CoreBundle/Migrations/Version20211018104331.php
+++ b/bundles/CoreBundle/Migrations/Version20211018104331.php
@@ -1,0 +1,105 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Schema\SchemaException;
+use Doctrine\Migrations\AbstractMigration;
+use Pimcore\Model\DataObject;
+
+final class Version20211018104331 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $list = new DataObject\ClassDefinition\Listing();
+
+        foreach ($list as $class) {
+            $foreignKeys = $this->getForeignKeys($class);
+
+            foreach($foreignKeys as $table => $objectIdColumn) {
+                try {
+                    $tableSchema = $schema->getTable($table);
+                } catch(SchemaException $e) {
+                    continue;
+                }
+
+                $fkName = 'fk_'.$table.'__'.$objectIdColumn;
+                if (!$tableSchema->hasForeignKey($fkName)) {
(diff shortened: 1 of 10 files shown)


brusch:
@BlackbitNeueMedien I agree, that totally makes sense! 
Just some more thoughts on it: 
- we should implement it consistently across all object related tables, so including localized fields, object bricks and field collections
- I think we can remove `ON UPDATE` as updating the ID is anyway not supported
- we should create migrations for existing tables
- we should remove resulting obsolete code 

For now I'd focus on data objects, then we could and should extend it at least to all other entities (Documents & Assets).

BlackbitDevs:
Is now implemented for all data object-related tables.
I removed the "ON UPDATE CASCADE" (as @brusch mentioned)
Wrote a migration for existing tables.
Removed now unnecessary deletion of object-related data as this now gets automatically deleted via foreign keys.

bundles/CoreBundle/Migrations/Version20211018104331.php
```diff
@@ -0,0 +1,110 @@
...
+            foreach($foreignKeys as $table => $objectIdColumn) {
+                try {
+                    $tableSchema = $schema->getTable($table);
+                } catch(SchemaException $e) {
+                    continue;
+                }
+
+                $fkName = AbstractDao::getForeignKeyName($table, $objectIdColumn);
+                if (!$tableSchema->hasForeignKey($fkName)) {
+                    if($tableSchema->getColumn($objectIdColumn)->getCustomSchemaOption('unsigned') === false) {
```
brusch:
```suggestion
                    $column = $tableSchema->getColumn($objectIdColumn);
                    if(!$column->hasCustomSchemaOption('unsigned') || $column->getCustomSchemaOption('unsigned') === false) {
```
Otherwise this would result in `Warning: Undefined array key "unsigned" `

bundles/CoreBundle/Migrations/Version20211018104331.php
```diff
@@ -0,0 +1,110 @@
...
+    private function getForeignKeys(DataObject\ClassDefinition $class): array
+    {
+        $foreignKeys = [
+            'object_query_'.$class->getId() => 'oo_id',
+            'object_store_'.$class->getId() => 'oo_id',
+            'object_relations_'.$class->getId() => 'src_id',
+            'object_classificationstore_groups_'.$class->getId() => 'o_id',
+            'object_classificationstore_data_'.$class->getId() => 'o_id',
+            'object_metadata_'.$class->getId() => 'o_id',
+            'object_localized_data_'.$class->getId() => 'ooo_id'
```
brusch, line 91: `object_localized_query_*` tables are missing here.

bundles/CoreBundle/Migrations/Version20211018104331.php
```diff
@@ -0,0 +1,110 @@
...
+                    continue;
+                }
+
+                $fkName = AbstractDao::getForeignKeyName($table, $objectIdColumn);
+                if (!$tableSchema->hasForeignKey($fkName)) {
+                    if($tableSchema->getColumn($objectIdColumn)->getCustomSchemaOption('unsigned') === false) {
+                        $tableSchema->changeColumn($objectIdColumn, ['unsigned' => true]);
+                    }
+
+                    $tableSchema->addForeignKeyConstraint('objects', [$objectIdColumn], ['o_id'], ['ON DELETE CASCADE'], $fkName);
```
brusch:
```suggestion
                    $tableSchema->addForeignKeyConstraint('objects', [$objectIdColumn], ['o_id'], ['onDelete' => 'CASCADE'], $fkName);
```
The current code however adds the FK, but with default action `RESTRICT`.

bundles/CoreBundle/Migrations/Version20211018104331.php
```diff
@@ -0,0 +1,110 @@
...
+            $foreignKeys = $this->getForeignKeys($class);
+
+            foreach ($foreignKeys as $table => $objectIdColumn) {
+                try {
+                    $tableSchema = $schema->getTable($table);
+                } catch (SchemaException $e) {
+                    continue;
+                }
+
+                $fkName = 'fk_'.$table.'__'.$objectIdColumn;
```
brusch:
```suggestion
                $fkName = AbstractDao::getForeignKeyName($table, $objectIdColumn);
```

bundles/CoreBundle/Migrations/Version20211018104331.php
```diff
@@ -0,0 +1,110 @@
...
+            'object_relations_'.$class->getId() => 'src_id',
+            'object_classificationstore_groups_'.$class->getId() => 'o_id',
+            'object_classificationstore_data_'.$class->getId() => 'o_id',
+            'object_metadata_'.$class->getId() => 'o_id',
+            'object_localized_data_'.$class->getId() => 'ooo_id'
+        ];
+
+        $brickList = new DataObject\Objectbrick\Definition\Listing();
+        foreach ($brickList->load() as $brickDefinition) {
+            $foreignKeys['object_brick_query_'.$brickDefinition->getKey().'_'.$class->getId()] = 'o_id';
```
brusch, line 100: `object_brick_store_*` tables are missing here.

bundles/CoreBundle/Migrations/Version20211018104331.php
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * @param DataObject\ClassDefinition $class
+     * @return string[]
+     */
+    private function getForeignKeys(DataObject\ClassDefinition $class): array
+    {
+        $foreignKeys = [
+            'object_query_'.$class->getId() => 'oo_id',
```
brusch, line 85: Additionally we should also add `object_url_slugs.objectId` , this also needs to be considered in `install.sql`.

models/DataObject/Concrete/Dao.php
```diff
@@ -419,19 +419,6 @@ public function saveChildData()
...
     {
-        $this->db->delete('object_query_' . $this->model->getClassId(), ['oo_id' => $this->model->getId()]);
-        $this->db->delete('object_store_' . $this->model->getClassId(), ['oo_id' => $this->model->getId()]);
-        $this->db->delete('object_relations_' . $this->model->getClassId(), ['src_id' => $this->model->getId()]);
-
-        // delete fields which have their own delete algorithm
-        if ($this->model->getClass()) {
-            foreach ($this->model->getClass()->getFieldDefinitions() as $fd) {
-                if ($fd instanceof CustomResourcePersistingInterface) {
-                    $fd->delete($this->model);
```
brusch: This needs to stay, because there could be data-types which are e.g. not using the database at all but e.g. Redis.

models/DataObject/ClassDefinition/Dao.php
```diff
@@ -127,32 +127,35 @@ public function update()
...
               `type` varchar(50) NOT NULL DEFAULT '',
               `fieldname` varchar(70) NOT NULL DEFAULT '0',
               `index` int(11) unsigned NOT NULL DEFAULT '0',
               `ownertype` enum('object','fieldcollection','localizedfield','objectbrick') NOT NULL DEFAULT 'object',
               `ownername` varchar(70) NOT NULL DEFAULT '',
               `position` varchar(70) NOT NULL DEFAULT '0',
               INDEX `forward_lookup` (`src_id`, `ownertype`, `ownername`, `position`),
-              INDEX `reverse_lookup` (`dest_id`, `type`)
+              INDEX `reverse_lookup` (`dest_id`, `type`),
+			  CONSTRAINT `".self::getForeignKeyName($objectDatastoreTableRelation, 'src_id')."` FOREIGN KEY (`src_id`) REFERENCES objects (`o_id`) ON DELETE CASCADE
```
brusch, line 158: We could do that for `dest_id` column as well, right?

brusch (changes requested):

bundles/CoreBundle/Migrations/Version20211018104331.php
BlackbitDevs: Fixed.

bundles/CoreBundle/Migrations/Version20211018104331.php
BlackbitDevs: Fixed.

bundles/CoreBundle/Migrations/Version20211018104331.php
BlackbitDevs: Fixed.

models/DataObject/Concrete/Dao.php
BlackbitDevs: Fixed.

models/DataObject/ClassDefinition/Dao.php
BlackbitDevs: I think this is not possible because `dest_id` could also refer to assets or documents. 

Edit: This seems to be named `polymorphic foreign key` and it is [possible](https://stackoverflow.com/questions/2002985/mysql-conditional-foreign-key-constraints#67708114) but this seems to get too complicated imho. Also I do not know if all supported DBMS vendors support such *generated columns*. What do you think, @brusch ?

brusch: @BlackbitNeueMedien you're right, I'd leave it as it is for now. Thanks a lot for your changes 😊

lib/Model/Dao/AbstractDao.php
```diff
@@ -95,4 +95,13 @@ public function resetValidTableColumnsCache($table)
         }
         Cache::clearTags(['system', 'resource']);
     }
+
+    public static function getForeignKeyName($table, $column) {
+        $fkName = 'fk_'.$table.'__'.$column;
+        if (strlen($fkName) > 64) {
+            $fkName = 'fk_'.md5($fkName);
```
brusch:
```suggestion
            $fkName = substr($fkName, 0, 55) . '_' . hash('crc32', $fkName);
```
I think keeping a prefix here would be nice and `crc32` should be enough randomness.

brusch:
Well done 👍 Thanks a lot!

PR #10487, opened 2021-10-04
[Data object grid] Fix JS error: Reload tree after deleting objects via grid

BlackbitDevs:
Steps to reproduce:
1. Create at least 2 data objects
2. Open grid which shows those 2 objects
3. Select both, right-click, click "Delete"

JS error occurs. This PR fixes this.


weisswurstkanone:
thx!

PR #10488, opened 2021-10-04
Fix Predefined getByKey and getByName methods

blankse:
This methods doesn't work anymore because assignVariablesToModel needs an array and not an object.


brusch:
👍 Thanks a lot!

PR #10493, opened 2021-10-04
NotFoundException for all getBy* methods

blankse:
Follow up to https://github.com/pimcore/pimcore/pull/10291


brusch:
Compatibility see also https://github.com/pimcore/pimcore/pull/10291#pullrequestreview-755102003

Thanks a lot @blankse ! 🙏

PR #10497, opened 2021-10-04
[Documents] Editable Dialog Box - destroy dialog box on removing editable from areablock

### LINKED ISSUE #10474: [Bug]: Wrong Content In Editable Dialog Box
### Expected behavior

A correct editable dialog should open, even after replacing the brick at the same index.

### Actual behavior

First placed brick dialog config shows up, even if I'm replacing it.

### Steps to reproduce

Screen Capture

dvesh3:
Resolves #10474

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/areablock.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/areablock.js
@@ -851,18 +851,19 @@ pimcore.document.editables.areablock = Class.create(pimcore.document.area_abstra
 
     removeBlock: function (element) {
         let container = Ext.get(element);
+
+        let dialogBoxDiv = container.query('.pimcore_block_dialog[data-name="' + this.name + '"]')[0];
+        if (dialogBoxDiv) {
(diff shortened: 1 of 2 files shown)


brusch (approved):

bundles/AdminBundle/Resources/public/js/pimcore/document/editables/areablock.js
```diff
@@ -851,19 +851,20 @@ pimcore.document.editables.areablock = Class.create(pimcore.document.area_abstra
 
     removeBlock: function (element) {
         let container = Ext.get(element);
+
+        let dialogBoxDiv = container.query('.pimcore_block_dialog[data-name="' + this.name + '"]')[0];
+        if (dialogBoxDiv) {
+            let dialogBoxId = dialogBoxDiv.dataset.dialogId;
+            this.removeEditableDialogbox(dialogBoxId);
```
brusch, line 858: @dvesh3 JFI: I've moved that code block to the beginning, because actually `container.remove();` removed the `.pimcore_block_dialog` container as well, because it's a child. Your code worked for whatever reason, however I thought it might be a bit confusing 😉

PR #10502, opened 2021-10-05
[Data Objects] Text Layout - Add option to define content via twig

### LINKED ISSUE #9547: [Data Objects][Text Layout] Add option to define content via twig
similar to #9546

- probably directly inside existing wysiwyg editor? 
- pass same parameters as to renderLayoutText method to execution context
- also add preview/test view with data object to drag for preview

dvesh3:
Resolves #9547

## Additional info  
Twig content:
dynamic_textlabel_3

Preview:
dynamic_textlabel_4


bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
```diff
@@ -1887,4 +1887,49 @@ public function suggestCustomLayoutIdentifierAction(Request $request)
...
+     * @Route("/text-layout-preview", name="pimcore_admin_dataobject_class_textlayoutpreview")
+     *
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function textLayoutPreviewAction(Request $request)
+    {
+        $objPath = $request->get('previewObject', '');
+        $obj = DataObject::getByPath($objPath) ?? new DataObject\Concrete();
```
fashxp: any change to instantiate the correct class and not concrete ... otherwise we will get error messages like the following, as the specific methods do not exist: 
image

bundles/AdminBundle/Resources/public/js/pimcore/object/classes/layout/text.js
```diff
@@ -48,49 +53,139 @@ pimcore.object.classes.layout.text = Class.create(pimcore.object.classes.layout.
...
+                                    xtype: "textfield",
+                                    fieldLabel: t("rendering_data"),
+                                    width: 600,
+                                    value: this.datax.renderingData,
+                                    name: "renderingData"
+                                },
+                                {
+                                    xtype: 'container',
+                                    style: 'padding-top:10px;',
+                                    html: 'You can use the following markup (in source edit mode) to make custom alerts: <br> <pre>&lt;div class=&quot;alert alert-success&quot;&gt;Your Message&lt;/div&gt;</pre>The following contextual classes are available: <pre>alert-primary, alert-secondary, alert-success, alert-danger, alert-warning, alert-info</pre>'
```
fashxp: should we add a info for twig support here too?

fashxp (changes requested):

bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
dvesh3: passing null when no object is provided to the `enrichLayoutDefinition`.

bundles/AdminBundle/Resources/public/js/pimcore/object/classes/layout/text.js
dvesh3: Added

bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
fashxp: does null solve the issue? then I would get a null error message. could we not just pass the class id to the request and then create an empty dataobject of correct type?

dvesh3: Changed it to use default class object when none is provided in the field.

PR #10505, opened 2021-10-05
Initialize global variables "editmode" and "document" on kernel.request with defaults

### LINKED ISSUE #10504: [Bug]: Global Twig Variable "editmode" does not exist.
### Expected behavior

Rendering partials should always be possible. 

### Actual behavior

If you're using the delegated templating engine before the `KernelEvents::CONTROLLER` state, additional renderings will fail (if variable `editmode` is given).

### Steps to reproduce

Stupid example but easy to reproduce: Use this class, create any object and try to save:

image

```php
<?php

namespace App\EventListener;

use Pimcore\Event\DataObjectEvents;
use Pimcore\Event\Model\DataObjectEvent;
use Pimcore\Mail;
use Pimcore\Model\Document\Email;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\Templating\EngineInterface;

class TestListener implements EventSubscriberInterface
{
    public EngineInterface $template;

    public function __construct(EngineInterface $template)
    {
        $this->template = $template;
    }
    
    {
        return [
            DataObjectEvents::PRE_UPDATE => 'onUpdate',
        ];
    }

    public function onUpdate(DataObjectEvent $event)
    {
         $data = $this->template->render(
            'example/data.html.twig',
            ['data' => []]
        );

         // use a template with a containing editmode variable
         $doc = Email::getByPath('/path/to/email');

         $mail = new Mail();

         $mail->setParam('data', $data);
         $mail->setDocument($doc);
         $mail->send();
    }
}
```

This happens because of this line:

https://github.com/pimcore/pimcore/blob/4d10439eb75b76acb8c1e07007f57d0d4e43078f/bundles/CoreBundle/EventListener/Frontend/GlobalTemplateVariablesListener.php#L75 

=> `$this->extensionSet->isInitialized()` will be `true` after the first render task (`$data`) and therefor this exception will be thrown after the second render task (email template).

brusch:
fixes #10504

### DIFF
--- a/bundles/CoreBundle/EventListener/Frontend/GlobalTemplateVariablesListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/GlobalTemplateVariablesListener.php
@@ -23,6 +23,7 @@ use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerAwareTrait;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\ControllerEvent;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
 use Symfony\Component\HttpKernel\Event\ResponseEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
 use Twig\Environment;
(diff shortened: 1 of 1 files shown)


brusch:
@solverat any feedback? Thanks!

PR #10515, opened 2021-10-06
Consider session service in \Pimcore\Http\RequestHelper::createRequestWithContext()

brusch:


dpfaffenbauer:
session service was deprecated by symfony and should not be used anymore. use the session from the request object instead. https://symfony.com/blog/new-in-symfony-5-3-session-service-deprecation

brusch:
@dpfaffenbauer yep, I'm aware of that, but actually the problem is that the session is not on the request object and throws an exception 😅 Let me check if there's an alternative 😊

brusch:
Closed in favor of scheb/2fa#109

PR #10525, opened 2021-10-07
Fix empty Hotspots and Markers in Image-Editable

MosiiAUT:
Fix empty Hotspots and Markers in Image-Editable:

How to reproduce:
- create Image-Editable and fill it with an image
- set marker or hotspot on image and fill it with data (doesn't matter which)
- save the document and reload
- data will now be gone in the markers/hotspots (but is still saved in DB)
- if the document is saved again, all de data will be overwritten and thus deleted

Change protected vars back do public

## Additional info  
object2array() - helper function casts the data to Json and back to assoc-array, but only public data will be converted.


brusch:
@MosiiAUT Thanks a lot for your PR! However, it's unfortunately on the wrong branch, could you please rebase it onto `10.1`? Thanks a lot!

PR #10528, opened 2021-10-07
[Data objects] Better aggregation of validation exception messages for nested fields with > 2 levels

BlackbitDevs:
Steps to reproduce:
1. Create data object class `Product`
2. Create data object class `Test` like this:
Bildschirmfoto 2021-10-07 um 18 10 31
For the many to one relation allow class `Test`

3. Create object of class `Product`
4. Create object of class `Test`
5. Create the following PHP script:
```php
<?php
$object = \Pimcore\Model\DataObject\Test::getById(2); // or whatever id the just created object of class Test has
$object->setPublished(true);
$data = [
    "relation" => new \Pimcore\Model\DataObject\Data\BlockElement('relation', 'manyToOneRelation', \Pimcore\Model\DataObject\Product::getById(1)), // or whatever id the just created object of class Product has
];

$object->setBlock([$data], 'en'); 
$object->save();
```

Without this PR you will get the following error:
```
[Pimcore\Model\Element\ValidationException]                                             
  Validation failed:  fieldname=localizedfields ( fieldname=block[ localizedfields-de ]) 
```
This does not say anything about which error actually occured - only that it is a validation exception and where it happened.

With this PR, you will get the error:
```
[Pimcore\Model\Element\ValidationException]                                                                                                                          
  Validation failed: Invalid data in field `relation` [type: manyToOneRelation][ block-0 ] fieldname=block[ localizedfields-de ] fieldname=localizedfields
```
This `Invalid data in field 'relation'` is the actual error which occured for adding an object of the wrong class in https://github.com/pimcore/pimcore/blob/d54aaae8d294f9e7d91fa0a8ed899335d389bfb9/models/DataObject/ClassDefinition/Data/ManyToOneRelation.php#L418

This lack of information about the actual error caused a lot of headaches in repairing the `BlockTest` in #10087

The reason why currently som errors are swallowed away, is that in https://github.com/pimcore/pimcore/blob/b6512cb45e06a5058ee7a2ffd0214ca2a905213e/models/DataObject/Concrete.php#L172-L204 only 2 levels of validation exceptions are taken into account while in above example there are 3 levels: relation -> block -> localized fields.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Block.php
+++ b/models/DataObject/ClassDefinition/Data/Block.php
@@ -1124,10 +1124,13 @@ class Block extends Data implements CustomResourcePersistingInterface, ResourceP
                 }
 
                 if ($validationExceptions) {
-                    $aggregatedExceptions = new Model\Element\ValidationException($validationExceptions[0]->getMessage());
-                    $aggregatedExceptions->setSubItems($validationExceptions);
-
-                    throw $aggregatedExceptions;
+                    $errors = [];
+                    /** @var \Exception $e */
+                    foreach ($validationExceptions as $e) {
+                        $errors[] = $e->getAggregatedMessage();
+                    }
+                    $message = implode(' / ', $errors);
+                    throw new Model\Element\ValidationException($message);
                 }
             }
         }
--- a/models/DataObject/ClassDefinition/Data/Block.php
+++ b/models/DataObject/ClassDefinition/Data/Block.php
@@ -1116,18 +1116,18 @@ class Block extends Data implements CustomResourcePersistingInterface, ResourceP
                             }
 
                             $fd->checkValidity($data, false, $params);
-                        } catch (Model\Element\ValidationException $ve) {
-                            $ve->addContext($this->getName() . '-' . $idx);
-                            $validationExceptions[] = $ve;
+                        } catch (Model\Element\ValidationException $e) {
+                            $exceptionClass = get_class($e);
+
+                            $newException = new $exceptionClass($e->getMessage().' fieldname='.$fd->getName(), $e->getCode(), $e->getPrevious());
+                            $subItems = $e->getSubItems();
+                            array_unshift($subItems, $e);
+                            $newException->setSubItems($subItems);
+                            $newException->addContext($this->getName().'-'.$idx);
+
+                            throw $newException;
                         }
                     }
-                }
-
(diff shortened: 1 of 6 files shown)


models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -1116,18 +1116,18 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
                             }
 
                             $fd->checkValidity($data, false, $params);
-                        } catch (Model\Element\ValidationException $ve) {
-                            $ve->addContext($this->getName() . '-' . $idx);
-                            $validationExceptions[] = $ve;
-                        }
-                    }
-                }
+                        } catch (Model\Element\ValidationException $e) {
```
dvesh3: so block fields validation errors are not aggregated anymore with this change? as we throw directly in the catch on first error.

dvesh3: @BlackbitNeueMedien ??

models/Element/ValidationException.php
```diff
@@ -22,19 +22,19 @@ class ValidationException extends \Exception
...
+     * @return \Exception[]
      */
     public function getSubItems()
     {
         return $this->subItems;
     }
 
     /**
-     * @param array $subItems
+     * @param \Exception $subItems
```
blankse:
```suggestion
     * @param \Exception[] $subItems
```

models/DataObject/ClassDefinition/Data/Block.php
BlackbitDevs: Sorry, this was not intended. I updated the PR description, the actual problem is this:
> The reason why currently som errors are swallowed away, is that in https://github.com/pimcore/pimcore/blob/b6512cb45e06a5058ee7a2ffd0214ca2a905213e/models/DataObject/Concrete.php#L172-L204 only 2 levels of validation exceptions are taken into account while in above example there are 3 levels: relation -> block -> localized fields.

mcop1 (approved):

models/Element/ValidationException.php
```diff
@@ -72,4 +72,35 @@ public function __toString()
 
         return $result;
     }
+
+    public function getAggregatedMessage() {
```
blankse:
```suggestion
    public function getAggregatedMessage(): string
    {
```

PR #10529, opened 2021-10-07
[Data objects: Check Validity] Check inherited value only if field supports inheritance and value is not empty

BlackbitDevs:
In `checkValidity()` currently for all object and localized fields whenever the class supports inheritance and the object's value is not valid, then it is tried to get the value with inheritance. When the field does not support inheritance or when the value is not empty currently, we cannot expect that any other value be returned in this case with inheritance as without. This PR removes this extra try with inheritance if the field type does not support inheritance or the current value is not empty.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Localizedfields.php
+++ b/models/DataObject/ClassDefinition/Data/Localizedfields.php
@@ -950,7 +950,7 @@ class Localizedfields extends Data implements CustomResourcePersistingInterface,
                 foreach ($this->getFieldDefinitions() as $fd) {
                     try {
                         try {
-                            if (!isset($dataForValidityCheck[$language]) || !isset($dataForValidityCheck[$language][$fd->getName()])) {
+                            if (!isset($dataForValidityCheck[$language][$fd->getName()])) {
                                 $dataForValidityCheck[$language][$fd->getName()] = null;
                             }
                             $fd->checkValidity($dataForValidityCheck[$language][$fd->getName()], false, $params);
--- a/models/DataObject/ClassDefinition/Data/Localizedfields.php
+++ b/models/DataObject/ClassDefinition/Data/Localizedfields.php
@@ -955,7 +955,7 @@ class Localizedfields extends Data implements CustomResourcePersistingInterface,
                             }
                             $fd->checkValidity($dataForValidityCheck[$language][$fd->getName()], false, $params);
                         } catch (\Exception $e) {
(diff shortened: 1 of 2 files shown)


models/DataObject/ClassDefinition/Data/Localizedfields.php
```diff
@@ -950,13 +950,12 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
...
-                                $fd->checkValidity($dataForValidityCheck[$language][$fd->getName()], false, $params);
-                            } else {
-                                $fd->checkValidity(null, false, $params);
+                            if (!isset($dataForValidityCheck[$language][$fd->getName()])) {
+                                $dataForValidityCheck[$language][$fd->getName()] = null;
                             }
+                            $fd->checkValidity($dataForValidityCheck[$language][$fd->getName()], false, $params);
                         } catch (\Exception $e) {
-                            if ($data->getObject()->getClass()->getAllowInherit()) {
+                            if ($fd->supportsInheritance() && $fd->isEmpty($dataForValidityCheck[$language][$fd->getName()]) && $data->getObject()->getClass()->getAllowInherit()) {
```
dvesh3: I would check `$data->getObject()->getClass()->getAllowInherit()` first..

models/DataObject/Concrete.php
```diff
@@ -134,7 +134,7 @@ protected function update($isUpdate = null, $params = [])
                     try {
                         $fd->checkValidity($value, $omitMandatoryCheck, $params);
                     } catch (\Exception $e) {
-                        if ($this->getClass()->getAllowInherit()) {
+                        if ($fd->supportsInheritance() && $fd->isEmpty($value) && $this->getClass()->getAllowInherit()) {
```
dvesh3: I would check `$this->getClass()->getAllowInherit()` first..

dvesh3:
@BlackbitNeueMedien please see my comments above. thanks!

models/DataObject/ClassDefinition/Data/Localizedfields.php
BlackbitDevs: I have changed the order.

models/DataObject/Concrete.php
BlackbitDevs: I have changed the order.

dvesh3 (approved):

dvesh3:
I'll merge this PR as it corrects the wrong behavior. 

Please note this changes the behavior of validations in API, as now if the inherited object has an invalid value, then it will a throw validation exception (as oppose to previous behavior of saving the object successfully).

PR #10536, opened 2021-10-08
[Docs] Thumbnails: removed section about disabling WebP support in <img> tags

### LINKED ISSUE #10533: [Bug]: Unrecognized option "webp_auto_support" under "pimcore.assets.image.thumbnails".
### Expected behavior

Following [the documentation](https://pimcore.com/docs/pimcore/current/Development_Documentation/Assets/Working_with_Thumbnails/Image_Thumbnails.html#page_WebP-Support) we should not have WebP images generated by the server.

### Actual behavior

We get the following Symfony error:
> Unrecognized option "webp_auto_support" under "pimcore.assets.image.thumbnails". Available options are "auto_clear_temp_files", "clip_auto_support".

### Steps to reproduce

In `config.yaml` set:
```yaml
pimcore:
    assets:
        image:
            thumbnails:
                webp_auto_support: false
```

brusch:
related to #7661
fixes #10533

### DIFF
--- a/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
+++ b/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
@@ -374,18 +374,6 @@ of the image is on the focal point.
   
 ![Image thumbnails cover transformation considering focal point](../../img/image_thumbnails_cover_focal_point.png)
 
-## WebP Support 
-Pimcore  delivers automatically thumbnails in WebP format when using the `Auto` configuration for the 
-target format and when the client does support WebP (checking by evaluating the `Accept` request header).  
-    
-If you prefer not using WebP, you can disable the support by adding the following config option: 
-```yml
-    assets:
-        image:
-            thumbnails:
-                webp_auto_support: false
-```
-
 ## Clipping Support 
 Images with an embedded clipping path (8BIM / Adobe profile meta data) are automatically clipped when generating thumbnails of them. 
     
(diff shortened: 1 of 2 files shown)


NiklasBr:
The bug is in the missing configuration option, not in the docs.

brusch:
See https://github.com/pimcore/pimcore/issues/10533#issuecomment-938490740

NiklasBr:
@brusch this PR is effectively hiding the issue rather than fixing it. Since nothing has happened with the WebP format since 6.9 to improve the bugs in Imagemagick (or GD for that matter as far as I can see) I don't see how this improves the situation at all. This is not a valid fix from my perspective.

brusch:
@NiklasBr please understand, as mentioned here https://github.com/pimcore/pimcore/issues/10533#issuecomment-938490740 the config option in question was for a completely different reason!

NiklasBr:
@brusch your explanation did not help resolve the issue unfortunately. The issue is that webp as a format should not be delivered even if the browser says it has support for it (because of graphical issues with the format preventing us from using it). `webp_auto_support` neatly resolved that earlier this year, but in Pimcore X there is appears to be currently no choice but to deliver broken images?

brusch:
I totally got your point and your need, but this is not a bug or related to the config option in question - it was just for a different purpose, even if it solved your requirement. Like I wrote here https://github.com/pimcore/pimcore/issues/10533#issuecomment-938490740 I'm open to introduce a config for that, maybe something like: 

```yaml
pimcore:
    assets:
        image:
            thumbnails:
                auto_formats: ["jpeg", "webp"]
```

PR #10537, opened 2021-10-08
[Asset] Thumbnails: <picture> tag should only include webp if supported

brusch:


brusch:
@NiklasBr does that work for you? You should just ensure that Imagick doesn't have `webp` in it's list of supported formats.

PR #10545, opened 2021-10-10
Fixes stretched assets in grid

NiklasBr:
Before and after:

image
image


blankse:
@NiklasBr Wouldn't it be nicer to center the thumbnails?
```
.asset_folder_preview .thumb-wrap {
  display: flex;
  align-items: center;
  justify-content: center;
}
```

NiklasBr:
Good idea!

brusch:
@NiklasBr Thanks a lot for your PR, however I'd classify this as a bug, so could you please rebase onto `10.1` branch? Thanks in advance! 

Regression of #10271

NiklasBr:
@brusch There are too many conflicts to fix when rebasing on 10.1, I thought this would be a quick PR. Will try again for 10.1.6 instead if you don't want it in 10.2?

brusch:
Pushed the changes right to `10.1` since it's just a small change anyway - hope that is ok for you 😊

PR #10547, opened 2021-10-11
Fixes Warning: Undefined array key "host" at Video.php:493

NiklasBr:


brusch:
Thank you very much! 🙏

PR #10548, opened 2021-10-11
Order assets by modificationDate to process new/updated assets earlier

NiklasBr:
The motivation for this change is where cron is being used to generate thumbnails and huge folders with many thumbnail configs will spend a lot of time just working through existing files on disk. If the command exits early for some reason this provides a greater possibility of having those thumbs generated.


brusch:
Makes sense, thanks a lot!

PR #10556, opened 2021-10-11
Fix mail issues

### LINKED ISSUE #10550: [Bug]: Mail Issues
### Expected behavior

Mail exception handling works properly. Correct newsletter mailer in test mode has been defined.

### Actual behavior

The opposite of the expected behavior...

## Wrong exception management
This array never gets populated (btw a count() > 0 would be the nicer approach to check against population)
https://github.com/pimcore/pimcore/blob/2e71f2ac4eb74cc501d92aea97469837ecdbad31/lib/Mail.php#L574

## Wrong mailer nulled out
Instead of `newsletter` it should be `pimcore_newsletter`:
https://github.com/pimcore/pimcore/blob/417b50202ceba7721311eb73f92c308f84236546/bundles/CoreBundle/Resources/config/pimcore/test.yaml#L9

dvesh3:
Resolves #10550

### DIFF
--- a/lib/Mail.php
+++ b/lib/Mail.php
@@ -569,7 +569,7 @@ class Mail extends Email
             $failedRecipients = [];
 
             try {
-                $mailer->send($this);
+                $mailer->send($this, $failedRecipients);
             } catch (\Exception $e) {
                 if (isset($failedRecipients[0])) {
                     $sendingFailedException = new \Exception($failedRecipients[0] . ' - ' . $e->getMessage(), 0, $e);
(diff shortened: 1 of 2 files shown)


lib/Mail.php
```diff
@@ -569,7 +569,7 @@ public function sendWithoutRendering(MailerInterface $mailer = null)
             $failedRecipients = [];
 
             try {
-                $mailer->send($this);
+                $mailer->send($this, $failedRecipients);
```
blankse, line 572: @dvesh3 @brusch Does this really work? I see a Envelope typehint in this method and no array.
https://github.com/pimcore/pimcore/blob/10.1/lib/Mail/Mailer.php#L48

dvesh3: @blankse you are right! It does not accept array anymore as it was left over from Swiftmailer code. This will just throw exception when sending fails so pushed change directly on 10.1 https://github.com/pimcore/pimcore/commit/f3bd29ad5258a0572d460be4684cfd5be174d7f9

PR #10557, opened 2021-10-11
Update 06_Block.md

danielerne:
Set false in blockStart() 
Otherwise it will produce blockControls() twice

  
  

Resolves #


brusch:
Thanks a lot! 👍

PR #10561, opened 2021-10-12
Remove wrong ValidationException parameters

blankse:
Fixes PhpStan level 5 errors.
2 argument must be a integer, not null.


brusch:
Thanks a lot!

PR #10562, opened 2021-10-12
[Assets] Fix versions data can not be properly restored

### LINKED ISSUE #10477: [Bug]: Versions data can not be properly restored
### Expected behavior

When I have two or more versions of an asset I'm expecting to be able to both preview the version (if an image) or publish the older one. I'm also expecting the versions to have different file data.

### Actual behavior

Right now it displays the same image when previewing and if I try to publish an older version of an image I get an error message and if it's a PDF file the preview tab disappears. Also if I check the versions folder on the volume all of the files have the same data.

error-message-image

no-preview-tab

### Steps to reproduce

1. Upload an image or a PDF to an asset folder
2. Upload a new version to the same asset
3. Check file data and confirm that all versions have the same file data
4. Try to preview or publish an older version of the asset and confirm that it's not possible

dvesh3:
Resolves #10477

### DIFF
--- a/models/Version.php
+++ b/models/Version.php
@@ -242,9 +242,8 @@ final class Version extends AbstractModel
             $useHardlinks = \Pimcore::getContainer()->getParameter('pimcore.config')['assets']['versions']['use_hardlinks'];
             $storage->write($this->getBinaryStoragePath(), '1'); // temp file to determine if stream is local or not
             if ($useHardlinks && stream_is_local($this->getBinaryFileStream()) && stream_is_local($data->getStream())) {
-                $linkPath = stream_get_meta_data($this->getBinaryFileStream())['uri'];
-                $storage->delete($this->getBinaryStoragePath());
-                $linked = @link(stream_get_meta_data($data->getStream())['uri'], $linkPath);
+                $linkPath = stream_get_meta_data($data->getStream())['uri'];
+                $linked = @link(stream_get_meta_data($this->getBinaryFileStream())['uri'], $linkPath);
             }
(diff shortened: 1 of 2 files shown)


models/Version.php
```diff
@@ -242,9 +242,8 @@ public function save()
             $useHardlinks = \Pimcore::getContainer()->getParameter('pimcore.config')['assets']['versions']['use_hardlinks'];
             $storage->write($this->getBinaryStoragePath(), '1'); // temp file to determine if stream is local or not
             if ($useHardlinks && stream_is_local($this->getBinaryFileStream()) && stream_is_local($data->getStream())) {
-                $linkPath = stream_get_meta_data($this->getBinaryFileStream())['uri'];
-                $storage->delete($this->getBinaryStoragePath());
-                $linked = @link(stream_get_meta_data($data->getStream())['uri'], $linkPath);
+                $linkPath = stream_get_meta_data($data->getStream())['uri'];
+                $linked = @link(stream_get_meta_data($this->getBinaryFileStream())['uri'], $linkPath);
```
brusch: This would try to create a link from the one byte binary storage file created in line `243` (`$this->getBinaryFileStream()`) to the existing asset file (`$data->getStream()`), which fails, since both paths exist already, so `$linked` will be `false` and the binary data get's therefore written in line `250`. 

Remember: `link($target, $link);` (while `$target` is the path to the existing data).

brusch (changes requested):

brusch (approved):

PR #10582, opened 2021-10-13
[Admin] Minimize bundle JS files in production

brusch:
This is mainly because of maximum connections per host in Chrome which is 256 since Chrome 81. 
See also: https://stackoverflow.com/questions/985431/max-parallel-http-connections-in-a-browser

### DIFF
--- a/lib/Twig/Extension/AdminExtension.php
+++ b/lib/Twig/Extension/AdminExtension.php
@@ -18,6 +18,7 @@ declare(strict_types=1);
 namespace Pimcore\Twig\Extension;
 
 use Pimcore\Tool\Admin;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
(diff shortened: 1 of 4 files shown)


lib/Twig/Extension/AdminExtension.php
```diff
@@ -26,23 +27,48 @@
...
 {
+    public function __construct(private UrlGeneratorInterface $generator)
+    {
+    }
+
     public function getFunctions()
     {
         return [
             new TwigFunction('pimcore_minimize_scripts', [$this, 'minimize']),
+            new TwigFunction('pimcore_script_exists', [$this, 'exists']),
```
jdreesen, line 38: This is unused and the `exists` method does not exist in this class.

jdreesen: #10583

PR #10583, opened 2021-10-13
Remove unused twig function without implementation

jdreesen:
Resolves https://github.com/pimcore/pimcore/pull/10582#discussion_r728008621

## Additional info  

This was added in #10582 (probably by mistake).


brusch:
Thanks a lot!

PR #10586, opened 2021-10-13
Cleanup `bundles/AdminBundle/Resources/views/Admin/Index/index.html.twig` a bit

jdreesen:
Move some code to where it's used and remove a duplicated line.


brusch:
Thanks again!

PR #10588, opened 2021-10-13
Fix xliff export

thgross:
### Adds missing namespace for xliff-Element
some tools need the correct namespace:
`xmlns="urn:oasis:names:tc:xliff:document:1.2"`
(see https://docs.oasis-open.org/xliff/v1.2/xliff-profile-html/xliff-profile-html-1.2-cd02.html#Structure_MappingAttributes)

### Encodes text via `htmlspecialchars()` instead of wrapping every text in CDATA sections
Some translation providers, like Across, don't support CDATA sections.
Also, the use of CDATA sections is not recommended (see https://docs.oasis-open.org/xliff/v1.2/xliff-profile-html/xliff-profile-html-1.2-cd02.html#General_CDATA).

### Properly encodes some special characters like `<` and `>`
Previously, when you used these characters as normal text in a wysiwyg-editable, they would be used as HTML instead of text after importing the translated xliff-file.

**For example:**
  > this is \<strong\>bold\</strong\> text

would be

  > this is <strong>bold</strong> text

after export, translation and import.

### DIFF
--- a/lib/Translation/Escaper/Xliff12Escaper.php
+++ b/lib/Translation/Escaper/Xliff12Escaper.php
@@ -35,18 +35,14 @@ class Xliff12Escaper
         // remove nasty device control characters
         $content = preg_replace('/[\x00-\x09\x0B\x0C\x0E-\x1F\x7F]/', '', $content);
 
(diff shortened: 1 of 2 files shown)


dvesh3 (approved):

dvesh3:
@thgross thanks!

PR #10589, opened 2021-10-13
Fix undefined array_key

### LINKED ISSUE #10587: [Bug]: Undefined array_key
### Expected behavior

Pimcore works

### Actual behavior

Pimcore doesnt work

### Steps to reproduce

1. Install plain pimcore 10.2
2. Open pimcore
3. Error, Undefined array_key 'default'

dpfahlbusch:
Resolves #10587

### DIFF
--- a/lib/Perspective/Config.php
+++ b/lib/Perspective/Config.php
@@ -254,11 +254,7 @@ final class Config
     {
         $masterConfig = self::get()->toArray();
         
-        if (!array_key_exists($name, $masterConfig)) {
-            return [];
-        }
-
-        $config = $masterConfig[$name];
+        $config = $masterConfig[$name] ?? null;
         if (!$config) {
             $config = [];
         }
--- a/lib/Perspective/Config.php
+++ b/lib/Perspective/Config.php
@@ -254,12 +254,9 @@ final class Config
     {
         $masterConfig = self::get()->toArray();
         
-        $config = $masterConfig[$name] ?? null;
-        if (!$config) {
-            $config = [];
-        }
+        $config = $masterConfig[$name] ?? static::getStandardPerspective()['default'];
 
-        $tmpResult = $config['elementTree'] ?? [];
+        $tmpResult = $config['elementTree'];
         
         $result = [];
 
--- a/lib/Perspective/Config.php
+++ b/lib/Perspective/Config.php
@@ -253,10 +253,18 @@ final class Config
     protected static function getRuntimeElementTreeConfig($name)
     {
         $masterConfig = self::get()->toArray();
+        
+        if (!array_key_exists($name, $masterConfig)) {
+            return [];
+        }
 
         $config = $masterConfig[$name];
         if (!$config) {
             $config = [];
+        }
+        
+        if (!array_key_exists('elementTree', $config)) {
+            return [];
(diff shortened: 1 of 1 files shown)


lib/Perspective/Config.php
```diff
@@ -253,11 +253,19 @@ public static function getRuntimePerspective(\Pimcore\Model\User $currentUser =
     protected static function getRuntimeElementTreeConfig($name)
     {
         $masterConfig = self::get()->toArray();
+        
+        if (!array_key_exists($name, $masterConfig)) {
+            return [];
+        }
 
         $config = $masterConfig[$name];
```
jdreesen:
```suggestion
        
        $config = $masterConfig[$name] ?? null;
```

lib/Perspective/Config.php
```diff
@@ -253,11 +253,19 @@ public static function getRuntimePerspective(\Pimcore\Model\User $currentUser =
...
         if (!$config) {
             $config = [];
         }
+        
+        if (!array_key_exists('elementTree', $config)) {
+            return [];
+        }
```
jdreesen: Not sure that this is right, as `$config['elementTree']` seems to be optional. 

Maybe get it like this `$tmpResult = $config['elementTree'] ?? [];` and skip the following `if (is_null($tmpResult)) {...}` (sorry, GitHub does not allow me to create a suggestion for this)

dpfahlbusch: Sounds good, thank you :)

lib/Perspective/Config.php
```diff
@@ -253,16 +253,18 @@ public static function getRuntimePerspective(\Pimcore\Model\User $currentUser =
...
             $config = [];
         }
-
-        $tmpResult = $config['elementTree'];
-        if (is_null($tmpResult)) {
-            $tmpResult = [];
+        
+        if (!array_key_exists('elementTree', $config)) {
+            return [];
         }
```
jdreesen: This is not necessary anymore (and probably wrong anyways)
```suggestion
        
        if (!array_key_exists('elementTree', $config)) {
            return [];
        }
```

dpfahlbusch: GitHub-Web messed it up ^^ should be fixed now :)

dvesh3:
@dpfahlbusch thanks for your PR! However i'm afraid this is not an optimal solution as it still does not load the default config if none available from legacy php file, settings-store and symfony container. I would just simply load standard perspective config if none found here https://github.com/pimcore/pimcore/blob/10.x/lib/Perspective/Config.php#L85
```
if (!count($config)) {
    $config = static::getStandardPerspective();
}
```
Note: Nonexistent perspective requested by the user is handled in the event listener https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/EventListener/UserPerspectiveListener.php#L89

dpfahlbusch:
@dvesh3 hope this is the right way :)

dvesh3:
@dpfahlbusch I would just use standard config in `get()` method, as this called at every step of loading perspectives. please see my PR https://github.com/pimcore/pimcore/pull/10591

dvesh3:
closed in favor of #10591

PR #10611, opened 2021-10-15
Update Web2Print/Config - Config could be null

dpfaffenbauer:
$repository->loadConfigByKey can return null for config, but Config\Config expects an array


brusch:
Thanks a lot! 🙏

blankse:
@brusch Should be cherry picked to 10.2 branch.

brusch:
@blankse on it 😉

PR #10621, opened 2021-10-15
disable the output-cache if the client sends an authorization header

thgross:
pimcore 10 enables the full page cache for `GET` requests by default in the `prod`-environment. It disables the cache automatically when a session is in use; however, if the application identifies it's users using an authorization-header, the full page cache is still enabled.

disables the output-cache if the client sends an authorization header.

Resolves #10610


brusch:
Thanks a lot!

PR #10622, opened 2021-10-16
Add "Set as favourite" option for gridconfigs

### LINKED ISSUE #9629: Add "Set as favourite" to Grid Configuration Window
# Feature Request
It would be nice to have an additional checkbox in the `Grid Configuration` window, so one can automatically set a grid view as favourite for users.

image

## Feature description
When checked, the `Save & Share` button should automatically set the favourite grid view of the users the grid view is shared with - unless, of course, they already set a favourite grid view themselves.

matthiashamacher:
Resolves [#9629](https://github.com/pimcore/pimcore/issues/9629)

## Additional info  
When the option is checked the grid will be set as the favourite for all shared users if they don't already have a favourite grid for that object. If they don't have a favourite grid at all the shared grid will also be set as the global favourite.

CleanShot 2021-10-16 at 12 13 24@2x


mcop1:
@matthiashamacher The "share globally" checkbox isn´t considered in the code. It only works if a user or role was selected from the corresponding dropdowns. Maybe we should consider this checkbox too? Or at least show a warning that this feature only works when a user/role was selected? What do you think?

matthiashamacher:
Oh yeah, you're right.
I think we should consider that checkbox too as otherwise, you need to add all users manually. This feature is for admins who want to set some defaults for the users, so in my opinion, this should work too.
But maybe we should implement that with kind of weight. What I mean is that currently if the user has already a favourite it is not overridden. But I think it would be useful to allow overriding if the current favourite of the user is a shared grid of another user (not owned by the current user) and is shared globally the favourite should be overridden.
I will definitely implement the logic for the "share globally" checkbox, what do you think of my proposal regarding the overriding?

mcop1:
Sounds good, maybe an additional checkbox to activate/confirm the overwriting would be a good idea?

matthiashamacher:
@mcop1 I've updated the PR. Please review again.

I've implemented the logic as followed:
If "share globally" and "set as favourite" are checked all users and roles are loaded and the grid config is set for all as favourite. If "share globally" is not checked the grid config is only set as favourite for the specified users and roles. If one of these users or roles already has a grid config it is checked 1. if the user is the owner. If that is the case we do not update the favourite. And 2. if the grid config is shared globally. If that is **not** the case we also do not update.

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -991,6 +1000,119 @@ protected function updateGridConfigShares($gridConfig, $metadata)
...
+                (int) $id,
+                $gridConfig->getClassId(),
+                (int) $objectId,
+                $gridConfig->getSearchType()
+            );
+
+            // If the user has already a favourite for that object we check the current favourite and decide if we update
+            if ($favourite instanceof GridConfigFavourite) {
+                // Check if the user is the owner. If that is the case we do not update the favourite
+                if ((int) $favourite->getOwnerId() === $currentUser->getId()) {
```
matthiashamacher: We need to check against `$id` instead of `$currentUser->getId()`

bundles/AdminBundle/Controller/Admin/Asset/AssetHelperController.php
```diff
@@ -614,6 +623,90 @@ protected function updateGridConfigShares($gridConfig, $metadata)
...
+            $favourite = GridConfigFavourite::getByOwnerAndClassAndObjectId(
+                (int) $id,
+                $gridConfig->getClassId(),
+                0,
+                $gridConfig->getSearchType()
+            );
+
+            if ($favourite instanceof GridConfigFavourite) {
+                // Check if the user is the owner. If that is the case we do not update the favourite
+                if ((int) $favourite->getOwnerId() === $currentUser->getId()) {
```
matthiashamacher: We need to check against `$id` instead of `$currentUser->getId()`

mcop1:
Looks very good now, just one more thing. Could you please confirm that "set as favourite" in combination with "shared roles" isn´t working? In my experience it gets shared, but never set as favourite.

matthiashamacher:
Yes, you are right. I removed the role logic as it is not needed. Also, I added the information that it will only work for users to the documentation.

mcop1:
Looks good 👍 thank you!

PR #10625, opened 2021-10-17
Fix scheduled block change listener not firing

### LINKED ISSUE #10415: [Bug]: Scheduled Block Change Listener not firing
### Expected behavior

The change listener on the date input of a scheduled block should fire.

### Actual behavior

When changing the date inputs value of a scheduled block the change listener isn't fired 

### Steps to reproduce

* Go to the Pimcore demo and add the "time-slider-featurette" to a page. 
* Set a marker for a specific date and time
* Change the date to another one

In this case, the previous marker will still be visible.
As far as I can tell the problem is that the change listener of the date input isn't fired. I am not sure why.

```js
// /bundles/AdminBundle/Resources/public/js/pimcore/document/editables/scheduledblock.js:68
this.dateField = new Ext.form.DateField({
    cls: "pimcore_block_field_date",
    value: initialDate,
    region: 'west',
    listeners: {
        'change': function() { // this function is never called
            this.loadTimestampsForDate();
        }.bind(this)
    }
});
```

Edit:
I think I found a fix. I'm not sure though if it is the right way. Maybe someone with more ExtJs knowledge can chime in

```js
// /bundles/AdminBundle/Resources/public/js/pimcore/document/editables/scheduledblock.js
// lines 68 - 76

this.dateField = new Ext.form.DateField({
  cls: "pimcore_block_field_date",
  value: initialDate,
  region: 'west',
});

this.dateField.on('select', function() {
  this.loadTimestampsForDate();
}.bind(this));
```

Brainshaker95:
Resolves #10415

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/scheduledblock.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/scheduledblock.js
@@ -69,12 +69,12 @@ pimcore.document.editables.scheduledblock = Class.create(pimcore.document.editab
             cls: "pimcore_block_field_date",
             value: initialDate,
             region: 'west',
-            listeners: {
-                'change': function() {
-                    this.loadTimestampsForDate();
-                }.bind(this)
-            }
         });
+          
+        this.dateField.on('select', function() {
+            this.loadTimestampsForDate();
+        }.bind(this));
+      
         controlItems.push(this.dateField);
 
         this.slider = Ext.create('Ext.pimcore.slider.Milestone', {
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/scheduledblock.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/scheduledblock.js
@@ -68,7 +68,7 @@ pimcore.document.editables.scheduledblock = Class.create(pimcore.document.editab
         this.dateField = new Ext.form.DateField({
             cls: "pimcore_block_field_date",
             value: initialDate,
-            region: 'west',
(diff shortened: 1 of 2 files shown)


dvesh3 (changes requested):
This now hides all markers, if a date is changed for a marker:
image

dvesh3 (approved):
> This now hides all markers, if a date is changed for a marker:
image

Nevermind, I didn't have entries for that day, so it works as expected. When you change a date it shows markers for that date.

dvesh3:
@Brainshaker95 thanks a lot!

Brainshaker95:
@dvesh3 No problem. I was confused at first and not sure what you meant, but it's alright.

I can't open another PR right now, but I think I stumbled upon another issue in the scheduledblock.js file yesterday

`cleanupTimestamps` is called 2 times, both times with `false`. I think the second call should pass `true` so that all timestamps get deleted when the corresponfing button is pressed.

blankse:
@dvesh3 Shouldn't it go into the 10.2 branch? Can you cherry pick it?

dvesh3:
@Brainshaker95 sorry for the confusion. I'll have a look at `cleanupTimestamps` issue.

@blankse yes, i'm on it. 

thanks!

dvesh3:
> cleanupTimestamps is called 2 times, both times with false. I think the second call should pass true so that all timestamps get deleted when the corresponfing button is pressed.

@Brainshaker95 thanks for reporting. I pushed the fix on 10.2 branch directly https://github.com/pimcore/pimcore/commit/4ddb6427d0a37881e5681a0536d084ab45eb49f9

PR #10627, opened 2021-10-18
Pass FC LayoutDefinitions to CLASS_FIELDCOLLECTION_LIST_PRE_SEND_DATA

dpfaffenbauer:
Pass the LayoutDefinition to the Event so EventListener can actually change the layout (eg. sort of a custom-layout thing for Field Collections)

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
@@ -907,9 +907,11 @@ class ClassController extends AdminController implements KernelControllerEventIn
         $event = new GenericEvent($this, [
             'list' => $definitions,
             'objectId' => $request->get('object_id'),
+            'layoutDefinitions' => $layoutDefinitions
         ]);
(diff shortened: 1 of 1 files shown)


mcop1:
@dpfaffenbauer 
PR looks good. There are 2 occasions where CLASS_FIELDCOLLECTION_LIST_PRE_SEND_DATA is called, but you added "layoutDefinitions" only once. Could you please add it to the other as well?

https://github.com/pimcore/pimcore/blob/0854b5a6c119b7bad423853d7065658a465fbfda/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php#L921
respectively
https://github.com/pimcore/pimcore/blob/0854b5a6c119b7bad423853d7065658a465fbfda/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php#L979

Thank you for your efforts!

dpfaffenbauer:
layoutDefinitions is not used in the second call at all, I am not even sure why it actually loads stuff within the loop in the second function at all. so there is no point in adding it to the second call.

mcop1:
You are right, we can omit it here.
Thanks :) 👍

PR #10633, opened 2021-10-18
[Layout][Toolbar] Fix UI throwing errors when a perspective has a space inside its name

### LINKED ISSUE #10632: [Bug]: UI throwing error when a perspective has a space inside its name
### Expected behavior

When you have a perspective with a space inside its name, you should be able to use and navigate the UI as before.

### Actual behavior

After updating to Pimcore 10.2, if a perspective has a space inside its name, the UI throws the following error:

`Uncaught Error: Invalid component "itemId": "pimcore_menu_file_perspective_My Perspective"`

Probably caused by #10126, as HTML elements cannot have space.

A solution would be to replace all spaces inside the perspective names with hyphens or underscores.
If this seems a good solution to you, I can open a PR.

### Steps to reproduce

1. Change your perspectives.php file and add a perspective with a key containing one or more spaces
2. Refresh the UI. You won't be able to click on any toolbar entry.

Not reproducible neither on https://demo.pimcore.fun/ as its version is 10.1.x-dev nor on https://demo.pimcore.com/ as the "superuser" user doesn't have permissions to explore server files.

emanmagni:
Resolves #10632 

## Additional info
Issue solved by replacing any space inside the name with hyphens.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/layout/toolbar.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/layout/toolbar.js
@@ -34,7 +34,7 @@ pimcore.layout.toolbar = Class.create({
                         var itemCfg = {
                             text: t(perspective.name),
                             disabled: perspective.active,
-                            itemId: 'pimcore_menu_file_perspective_' + perspective.name.replace(/\s/g, '-'),
+                            itemId: 'pimcore_menu_file_perspective_' + perspective.name..replace(/[^a-z0-9\-_]/i, '-'),
                             handler: this.openPerspective.bind(this, perspective.name)
                         };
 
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/layout/toolbar.js
```diff
@@ -34,7 +34,7 @@ pimcore.layout.toolbar = Class.create({
                         var itemCfg = {
                             text: t(perspective.name),
                             disabled: perspective.active,
-                            itemId: 'pimcore_menu_file_perspective_' + perspective.name,
+                            itemId: 'pimcore_menu_file_perspective_' + perspective.name.replace(/\s/g, '-'),
```
brusch:
```suggestion
                            itemId: 'pimcore_menu_file_perspective_' + perspective.name..replace(/[^a-z0-9\-_]/i, '-'),
```

brusch (changes requested):

brusch:
Just replacing `  `  isn't enough, we need to ensure the resulting ID validates against: 
image

emanmagni:
@brusch Thanks!

PR #10638, opened 2021-10-18
[Admin] DataObjects - Fix Url Slug domain inputs visibility

### LINKED ISSUE #10604: [Bug]: Domain input for URL Slug is hidden since Pimcore X
### Expected behavior

All fields should be visible in the Ext JS interface.

### Actual behavior

The domain-specific URL Slug field (and its button is hidden:
image
image

By manually adjusting the height of the elements in the DOM it is possible to show that both fields are available, just hidden behind some inline `height` attributes:
image

### Steps to reproduce

Putting a URL Slug in a fieldset which in turn is in a Localized fields 

image

dvesh3:
Resolves #10604


NiklasBr:
Looks like this patch works fine @dvesh3!

dvesh3:
> Looks like this patch works fine @dvesh3!

@NiklasBr thanks for testing 😉

PR #10640, opened 2021-10-18
[Data objects] Service::cloneMe() should not clone related elements

BlackbitDevs:
Follow-up of unintentionally closed #9866 

`\Pimcore\Model\Service::cloneMe()` is not supposed to create deep clones of elements. For this reason there is a `KeepFilter` in https://github.com/pimcore/pimcore/blob/b6512cb45e06a5058ee7a2ffd0214ca2a905213e/models/Element/Service.php#L1216-L1247. But for relations with more than one element like (advanced) many-to-many relations, image galleries etc. the related objects still get cloned.
Steps to reproduce:
1. Create class with a many-to-many object relation field named `m2m`
2. Create object `A` under `/`, save & publish
3. Create object `B`, assign `A` in the many-to-many relation, save and publish
4. Use this script:
```php
<?php
$object = \Pimcore\Model\DataObject\Concrete::getByPath('/B');
$clone = \Pimcore\Model\Element\Service::cloneMe($object);
$clone->getM2M()[0]->setKey(uniqid());
var_dump($object->getM2M()[0]->getKey()); // outputs a uniqid instead of A
```

When the elements are directly connected like for many-to-one relations then the related objects does currently not get cloned.

Originally I thought the `cloneMe` method *was* intended to create deep-clones but in https://github.com/pimcore/pimcore/pull/9866#issuecomment-892463899 there is the reasons why it is not.


BlackbitDevs:
Also here, this is a PR and I am looking forward to feedback...

dvesh3:
@BlackbitNeueMedien could you please confirm if the issue is still valid? I'm not able to reproduce it on latest 10.4.4 with given steps. thanks!

dvesh3:
@BlackbitNeueMedien 🏓

BlackbitDevs:
Hi @dvesh3,

if above test now works, then we can close this PR:
>  Create class with a many-to-many object relation field named m2m
> Create object A under /, save & publish
    Create object B, assign A in the many-to-many relation, save and publish
    Use this script:
> <?php
$object = \Pimcore\Model\DataObject\Concrete::getByPath('/B');
$clone = \Pimcore\Model\Element\Service::cloneMe($object);
$clone->getM2M()[0]->setKey(uniqid());
var_dump($object->getM2M()[0]->getKey()); // outputs a uniqid instead of A

dvesh3:
> if above test now works, then we can close this PR:

yes, it works on latest version 10.4.5. thanks!

PR #10641, opened 2021-10-18
[Search] Index is not updated on adding/updating elements

brusch:


manasaR507:
bin/console pimcore:search-backend-reindex while running this command. Throwing error like - In Image.php line 390:
Cannot use object of type Pimcore\Model\DataObject\Data\BlockElement as array

Issue #9731, opened 2021-07-12
[Bug]: Video previews are no longer working


nklotter:
### Expected behavior

Video previews should be playable from within the Asset section of the Admin console ("Preview" tab)

### Actual behavior

Video previews do not play

----------

See line 1632 of the Pimcore\Bundle\AdminBundle\Controller\Admin\Asset\AssetController class. The following code modification resolved the issue...

REPLACE =>
preg_replace('@^' . preg_quote($asset->getPath(), '@') . '@', '', urldecode($thumbnail['formats']['mp4']))
WITH =>
preg_replace('@.*?' . preg_quote($asset->getPath(), '@') . '@', '', urldecode($thumbnail['formats']['mp4']))

### Steps to reproduce

1. Upload a video
2. Wait for the preview to generate
3. Click on the "Play" button

brusch:
@nklotter for me this works, also in the demo. 
Can you please share more details on it, like which path the asset has and how to reproduce the issue on our [public demo](https://github.com/pimcore/pimcore#demo-community-edition). 

Thanks!

nklotter:
@brusch We're using AWS Cloudfront to house our external assets and thumbnails. The absolute URL paths are set under the assets -> frontend_prefixes section of the config.yaml.

Thanks for the quick response!

brusch:
@nklotter Ah, got it - but then we'd need to implement a different fix. The reason why it fails in this case is that `getPath()` is prefixed with `https://tavi-12345678990.cloudfront.net/....` 
Will try to provide a fix soon 😉

nklotter:
@brusch Awesome, thanks for looking into this!

PR #10644, opened 2021-10-19
[Image thumbnails] Support absolute URLs in getHtml()

BlackbitDevs:
This PR adds support for absoute URLs in `Asset\Image\Thumbnail::getHTML()`.

Usage:
```php
$asset = \Pimcore\Model\Asset\Image::getById(16);
var_dump($asset->getThumbnail('test')->getHtml(['absoluteUrl' => true]));
```
This outputs:
```html
<picture >
	<source srcset="http://example.org/image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51.webp 1x, http://example.org/image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51@2x.webp 2x" type="image/webp" />
	<source srcset="http://example.org/image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51.png 1x, http://example.org/image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51@2x.png 2x" type="image/png" />
	<img src="http://example.org/image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51.png" width="2618" height="198" alt="" loading="lazy" />
</picture>
```

The default protocol is defined in https://github.com/pimcore/pimcore/blob/777984e9d301bc1df1dbbe2cdbda04d27e74b750/lib/Tool.php#L433
Imho nowadays we should change this to `https`. Especially when doing exports from CLI there is no Symfony request object which the real protocol could be retrieved from. Alternatively a configuration like for the main domain in `System Settings > Website` would also work.
Because of the current approach, this PR implements to set the protcol to be used via `absoluteUrlProtocol`
```php
var_dump($asset->getThumbnail('test')->getHtml(['absoluteUrl' => true, 'absoluteUrlProtocol' => 'https']));
```

### DIFF
--- a/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
+++ b/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md
@@ -94,7 +94,8 @@ You can configure the generated markup with the following options:
 | `imgCallback`                  | callable | A callable to modify the attributes for the generated `<img>` tag. There 1 argument passed, the array of attributes.  |
 | `disableImgTag`                | bool     | Set to `true` to not include the `<img>` fallback tag in the generated `<picture>` tag.   |
 | `useDataSrc`                   | bool     | Set to `true` to use `data-src(set)` attributes instead of `src(set)`.   |
-
+| `absoluteUrl`                  | bool     | Set to `true` to use absolute URLs.   |
+| `absoluteUrlProtocol`          | string   | Protocol to be used for absolute URLs, e.g. "https"   |
 
 
 ## Usage Examples
--- a/models/Asset/Image/Thumbnail.php
+++ b/models/Asset/Image/Thumbnail.php
@@ -219,7 +219,6 @@ final class Thumbnail
     {
         $srcSetValues = [];
         $sourceTagAttributes = [];
-        $thumb = null;
 
         foreach ([1, 2] as $highRes) {
             $thumbConfigRes = clone $thumbConfig;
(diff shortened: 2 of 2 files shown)


brusch:
Maybe it's just easier to use the prefix config in this case? 
 ```yaml
pimcore:
    assets:
        frontend_prefixes:
            # Prefix used for the original asset files
            source: https://oreo-12345678990.cloudfront.net
            # Prefix used for all generated image & video thumbnails
            thumbnail: https://tavi-12345678990.cloudfront.net
```

**Comment by BlackbitNeueMedien** [comment]
How does this work?
I have 
```yaml
pimcore:
    assets:
        frontend_prefixes:
            thumbnail: https://example.org
```

Cleared cache, Symfony shows it:
```
bin/console debug:config pimcore assets.frontend_prefixes.thumbnail

Current configuration for "pimcore.assets.frontend_prefixes.thumbnail"
======================================================================

'https://example.org'
```

But with the script
```php
$asset = \Pimcore\Model\Asset\Image::getById(16);
echo $asset->getThumbnail('test')->getHtml();
```
I only get
```html
<picture >
	<source srcset="/image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51.webp 1x, /image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51@2x.webp 2x" type="image/webp" />
	<source srcset="/image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51.png 1x, /image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51@2x.png 2x" type="image/png" />
	<img src="/image-thumb__16__test/Bildschirmfoto 2021-10-07 um 12.16.51.png" width="2618" height="198" alt="" loading="lazy" />
</picture>
```

BlackbitDevs:
Ok this also would also work.
```yaml
pimcore:
    assets:
        frontend_prefixes:
            thumbnail_deferred: https://example.org
```

But this works only if I remove https://github.com/pimcore/pimcore/blob/777984e9d301bc1df1dbbe2cdbda04d27e74b750/models/Asset/Thumbnail/ImageThumbnailTrait.php#L295
What do you think, @brusch ? Can we remove this line generally to also use this configuration for CLI scripts?

brusch:
@BlackbitNeueMedien yes, that's a known problem that needs to be solved an additional parameter or a specific method. The background for this is: #9731 and #9768

brusch:
I think we should go for the prefix here and fix the challenge with the `Tool::isFrontend()`. 
That's by the way also the case when doing `$asset->getFullPath()`.

PR #10646, opened 2021-10-19
[Admin] Grids - Replace cellEditing with RowEditing

dvesh3:
In Admin interface, adding and updating settings on grids is cumbersome with cell-editing plugin as each cell edit fires an update request, which sometimes breaks when multiple requests are fired in close intervals, so I would like to propose using RowEditing instead of CellEditing plugin.

With Row Editing plugin, only one request is fired when clicking on update or pressing Enter. 
image

Grids covered:
 - Glossary
 - Unit settings
 - Document Types
 - Predefined Properties
 - Static Routes
 - Translations
 - Website Settings
 - Redirects

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/translations.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/translations.js
@@ -190,14 +190,13 @@ pimcore.settings.translations = Class.create({
             listeners: {
                 beforeedit: function(editor, e) {
                     let cm = this.grid.getColumnManager().getColumns();
-                    for (var i=0; i < cm.length; i++) {
-                        if (cm[i].dataIndex === 'data') {
-                            let editor = this.getCellEditor(e.record.get('type'));
+                    for (let i=0; i < cm.length; i++) {
+                        let columnId = cm[i].id;
+                        if (columnId.startsWith('translation_column_')) {
+                            let editor = this.getCellEditor(e.column.dataIndex.substring(1), e.record);
                             if (editor) {
-                                e.grid.columns[i].setEditor(editor);
+                                e.grid.getColumnManager().columns[i].setEditor(editor);
                             }
-
-                            break;
                         }
                     }
                 }.bind(this)
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/translations.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/translations.js
@@ -111,7 +111,7 @@ pimcore.settings.translations = Class.create({
                 sortable: true,
                 dataIndex: "_" + languages[i],
                 filter: 'string',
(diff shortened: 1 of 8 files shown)


dvesh3:
@fashxp @brusch what do you guys think about this change?

brusch:
Works like a charm - just one thing - could we maybe stop the row-editor from opening if the row isn't writable? Then it would be super smooth 😊
image

dvesh3:
> Works like a charm - just one thing - could we maybe stop the row-editor from opening if the row isn't writable? Then it would be super smooth 😊 image

@brusch yes, that's possible with `beforeedit` event listener like:
```js
beforeedit: function (editor, context, eOpts) {
    if (!context.record.data.writeable) {
        return false;
    }
}
```

brusch:
@dvesh3 nice, then I'd go for it 👍

bundles/AdminBundle/Resources/public/js/pimcore/settings/translations.js
```diff
@@ -184,25 +184,23 @@ pimcore.settings.translations = Class.create({
...
-                beforeedit: function(editor, context, eOpts) {
-                    editor.editors.each(function (e) {
-                        try {
-                            // complete edit, so the value is stored when hopping around with TAB
-                            e.completeEdit();
-                            if (in_array(context.field, this.languages)) {
-                                Ext.destroy(e);
+                beforeedit: function(editor, e) {
+                    let cm = this.grid.getColumnManager().getColumns();
+                    for (var i=0; i < cm.length; i++) {
```
matthiashamacher: As you use `let` already for the other variables, you could use it here too instead of `var` so it is consistent.

bundles/AdminBundle/Resources/public/js/pimcore/settings/website.js
```diff
@@ -256,25 +255,23 @@ pimcore.settings.website = Class.create({
...
-                            e.completeEdit();
-                            Ext.destroy(e);
-                        } catch (exception) {
-                            // garbage collector was faster
-                            // already destroyed
-                        }
-                    });
+                beforeedit: function(editor, e) {
+                    let cm = this.grid.getColumnManager().getColumns();
+                    for (var i=0; i < cm.length; i++) {
```
matthiashamacher: Same here, please use `let`

brusch:
@dvesh3 Looks good to me, just one thing, the quantity value grid doesn't save any values, can you please fix that 😊 Thanks a lot!

brusch (changes requested):
@dvesh3 Looks good to me, just one thing, the quantity value grid doesn't save any values, can you please fix that 😊 Thanks a lot!

dvesh3:
> @dvesh3 Looks good to me, just one thing, the quantity value grid doesn't save any values, can you please fix that 😊 Thanks a lot!

@brusch Fixed.

Issue #10613, opened 2021-10-15
[Improvement]: Doctrine  DBAL v3


brusch:
Should fix #5361 as well. 
Consider discussion here: #10648

See also: #12917

goetas:
FYI, 3.3.0 has been released

brusch:
@goetas thanks a lot 👍

jdreesen:
Meanwhile, `doctrine/migrations` is now compatible with DBAL v3 and DBAL v2 will get out of maintenance in April (plus an additional year of security fixes).

See: [Sunsetting Doctrine DBAL 2](https://www.doctrine-project.org/2022/01/22/sunsetting-dbal-2.html)

brusch:
Should be now ready to work on as #12917 got merged

Issue #5361, opened 2019-12-02
Memory leaks on bulk objects save


FredoVelcro-zz:
We are importing lots of objects (more than 100000) and the server memory keep increasing.
We tried many things to localize the origin of the problem, and it seems to be in the save() method/process itself.

What we tried to free memory :
- unset the objects vars upon completed save() or finish using it
- use \Pimcore::collectGarbage(); every X turns (ex: 100, 500, 1000, 5000 etc)
- use Version::disable(); eahc time before $myObject->save();
- use native calls of gc_* PHP garbage functions

and nothing changed...!
The only thing that prevent the memory growing was...commenting the save() call..

Can you tell us if you did some changes that could affect this since version 6.2.3 ? (DAO ?)

Best regards,

Fred

dpfaffenbauer:
https://github.com/pimcore/pimcore/issues/3104

dvesh3:
Long story, short https://github.com/pimcore/pimcore/issues/3104#issuecomment-442073399

dpfaffenbauer:
But: what I don't get is, why doesn't that happen with Doctrine ORM? That also uses doctrine/dbal?

FredoVelcro:
we did not have the problem with Pimcore 6.2.3 ...

brusch:
@FredoVelcro if possible it would be nice if you can do a profiler comparison between the old version and 6.2.3, imho that's the only way how we can find out what the problem actually is.

brusch:
@FredoVelcro ping

NiklasBr:
F.W.I.W. I'm experiencing this with Pimcore 6.6.4 and PHP 7.4, running a CLI script which does this:

```php
        $csvArray = $csv->decode(file_get_contents($filePath), CsvEncoder::FORMAT);
        foreach ($csvArray as $i => $item) {
            if ($i % 11 === 0) {
                \Pimcore::collectGarbage();
            }

            foreach ($item as $levelText => $name) {
                $output->writeln("Processing {$name} at level {$level}");

                $category = new Category();
                // ... set some values
                $category->save();

                $page = new Page();
                // ... set some values
                $page->save();
            }
        }
```

After about 100 rows memory usage has increased to about 4GB.

dpfaffenbauer:
1. streaming the csv helps
2. disable doctrine logging helps a lot
3. https://github.com/doctrine/dbal/issues/3047 still is the main issue here.

NiklasBr:
[...]

PR #10648, opened 2021-10-19
[DRAFT] Update to dbal 3.1

dpfaffenbauer:
https://github.com/pimcore/pimcore/issues/5361
https://github.com/pimcore/pimcore/issues/10613

@brusch @dvesh3 It is a BC break on multiple occasions here:
1. `Doctrine\DBAL\Connection` does not implement `Doctrine\DBAL\Driver\Connection` and has a different signature
2. `public function query(...$params)` changed to `public function query(string $sql): Result`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
     "doctrine/annotations": "^1.6.0",
     "doctrine/collections": "^1.5.0",
     "doctrine/persistence": "^2.1.0",
-    "doctrine/dbal": "^2.11.0",
+    "doctrine/dbal": "^3.1.3",
     "doctrine/doctrine-bundle": "^2.2.1",
     "doctrine/doctrine-migrations-bundle": "^3.0.1",
     "doctrine/inflector": "^1.4.0 || ^2.0.0",
--- a/lib/Db/Connection.php
+++ b/lib/Db/Connection.php
@@ -15,6 +15,8 @@
 
 namespace Pimcore\Db;
 
+use Doctrine\DBAL\Result;
+
 class Connection extends \Doctrine\DBAL\Connection implements ConnectionInterface
 {
     use PimcoreExtensionsTrait;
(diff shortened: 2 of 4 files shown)


brusch:
@dpfaffenbauer so a topic for Pimcore 11 then 🤔

composer.json
```diff
@@ -52,7 +52,7 @@
     "doctrine/annotations": "^1.6.0",
     "doctrine/collections": "^1.5.0",
     "doctrine/persistence": "^2.1.0",
-    "doctrine/dbal": "^2.11.0",
+    "doctrine/dbal": "^3.1.3",
```
brusch, line 55:
```suggestion
    "doctrine/dbal": "^2.11.0 || ^3.1.3",
```
Any concerns about supporting both?

dpfaffenbauer:
@brusch might be. maybe we could wrap the connection completely and still be compatible? Like ConnectionInterface doesn't have the query function. not sure if that might work?

composer.json
dpfaffenbauer: as is that will not work. signatures are to different

brusch:
@dpfaffenbauer don't think that would work, since currently you could do: 
```php
$db instanceof \Doctrine\DBAL\Connection && $db instanceof \Doctrine\DBAL\Driver\Connection
```

That's not gonna work, even if wrapping ... 😐

dpfaffenbauer:
@brusch then keep it for Pimcore X+1 (😉 ) and remove the custom Wrapper completely? I don't think there is still a lot need for it. We could prepare Pimcore X and move stuff from the "Custom" API's to the actual DBAL API's and properly deprecate old stuff.

brusch:
@dpfaffenbauer yep, agree 👍
So we close this PR and continue to work on preparing for XI (changing to DBAL API and deprecating the legacy stuff in the wrapper).

dpfaffenbauer:
perfect :)

brusch:
Btw. PRs are welcome, we can do the migration to Doctrine API also in smaller chunks, so don't be afraid 😉
Last thing should be then the deprecation of the legacy APIs. Maybe we can get this already into `10.3`.

PR #10652, opened 2021-10-19
[Documents] Blocks Editable - Ignore adding extra indices on required check and xliff export

### LINKED ISSUE #10443: [Bug]: Required fields in Blocks fail serverside check
### Expected behavior

In more complex sites, validating required fields within block elements fails.

This is checked in:
vendor/pimcore/pimcore/models/Document/PageSnippet.php:664

Here, we get the list of used editable names. The issue is that, for some block elements, a non-existing ediable name is added with the number of 1000000

I have come so far, as to see, this is added in:
vendor/pimcore/pimcore/lib/Event/Model/Document/EditableNameEvent.php:117
although when you look at the page object, there is no editable with that name.

There will be, for example an editable:
stage_imageset_block:1.stage_image
but this ediable name will also be added to the list:
stage_imageset_block:1000000.stage_image

I will add some screenshots and object dumps.

### Actual behavior

Validation works properly.

### Steps to reproduce

No easy way found yet, to reproduce, will look further.

dvesh3:
Resolves #10443

### DIFF
--- a/models/Document/Editable/Block.php
+++ b/models/Document/Editable/Block.php
@@ -24,7 +24,10 @@ use Pimcore\Tool\HtmlUtils;
  */
 class Block extends Model\Document\Editable implements BlockInterface
 {
-    const IGNORE_EDITMODE_INDICES = '_block_ignore_extra_editmode_indices';
+    /**
+     * @internal
+     */
+    const ATTRIBUTE_IGNORE_EDITMODE_INDICES = '_block_ignore_extra_editmode_indices';
 
     /**
      * Contains an array of indices, which represent the order of the elements in the block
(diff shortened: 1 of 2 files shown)


models/Document/Editable/Block.php
```diff
@@ -437,4 +439,14 @@ private function getBlockAttributes(): string
...
     }
+
+    /**
+     * @return bool
+     */
+    private function isIgnoreEditmodeIndices(): bool
+    {
+        $requestStack = \Pimcore::getContainer()->get('request_stack');
+
+        return $requestStack->getCurrentRequest()->get(self::IGNORE_EDITMODE_INDICES, false);
```
brusch: Just to be sure, I'd add a check here if `getCurrentRequest() !== null` .

brusch (changes requested):

models/Document/Editable/Block.php
dvesh3: Returning `false` on `getCurrentRequest() === null`

brusch (approved):

PR #10653, opened 2021-10-19
remove all internal usages of Pimcore ConnectionInterface/PimcoreExtensionsTrait

dpfaffenbauer:
`Pimcore\Db::get` currently includes a BC break, cause it changes the doc-type for the return class. This should still be `ConnectionInterface|Connection`. For now I removed it to get a idea of how many usages there actually are. Once Stan doesn't report any "unknown function" calls, we can re-add that and then merge it.


dpfaffenbauer:
@brusch this should go into the direction we discussed. could you take a brief look at it and give me some feedback?

dpfaffenbauer:
ping @brusch ?

brusch:
@dpfaffenbauer sorry, this is a quite huge PR and I'm quite busy at the moment. What is the status of this PR, is it already finished from your side, as it is still flagged as draft?

dpfaffenbauer:
@brusch yes, if someone could have a look at it...

brusch:
I'd like to start this topic again 😊 But I think it would make sense to split this monster PR 👾 up into smaller pieces. Maybe the first part could be to deprecate the interface, change the return types and introduce the helpers. And then we can continue with changing the DI stuff, the methods, .... WDYT?

brusch:
@dpfaffenbauer sorry, missed to tag you 🤗

brusch:
@dpfaffenbauer do you think you can do this anytime soon? Thanks for a quick info 😉

PR #10657, opened 2021-10-20
Document-types remove int-cast and is_numeric-function

MosiiAUT:
Document-types configurations are now yaml-files with a string ID

Removes int-cast and is_numeric query of the "docTypeId" param for document-types


dvesh3:
@MosiiAUT thanks!

PR #10665, opened 2021-10-20
Fix error if redirect is already deleted

blankse:
Error occurs when double-clicking. Or if another user has already removed the redirect.

Remove:
```
Call to a member function delete() on null
```

Update:
```
Call to a member function setValues() on null
```


brusch:
Thank you very much 😊

PR #10666, opened 2021-10-21
[BUG] PimcoreLocationAwareConfigDao: Runtime cache doesn´t cache datasource

mcop1:
PimcoreLocationAwareConfigDao: Runtime cache doesn´t cache datasource, therefore datasource is null for all cached items. As a result "isWriteable" Method doesn´t work like intended.

### DIFF
--- a/lib/Model/Dao/PimcoreLocationAwareConfigDao.php
+++ b/lib/Model/Dao/PimcoreLocationAwareConfigDao.php
@@ -70,10 +70,12 @@ abstract class PimcoreLocationAwareConfigDao implements DaoInterface
 
         list($data, $this->dataSource) = $this->locationAwareConfigRepository->loadConfigByKey($id);
 
-        self::$cache[$this->settingsStoreScope][$id] = [
-            'datasource' => $this->dataSource,
-            'data' => $data,
-        ];
+        if($data) {
+            self::$cache[$this->settingsStoreScope][$id] = [
+                'datasource' => $this->dataSource,
+                'data' => $data,
+            ];
+        }
 
         return $data;
     }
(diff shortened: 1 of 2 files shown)


lib/Model/Dao/PimcoreLocationAwareConfigDao.php
```diff
@@ -63,13 +65,17 @@ protected function getDataByName(string $id)
...
+            isset(self::$cache[$this->settingsStoreScope][$id][self::cacheKeyDataSource])) {
+
+            $this->dataSource = self::$cache[$this->settingsStoreScope][$id][self::cacheKeyDataSource];
             return self::$cache[$this->settingsStoreScope][$id];
         }
 
         list($data, $this->dataSource) = $this->locationAwareConfigRepository->loadConfigByKey($id);
 
         self::$cache[$this->settingsStoreScope][$id] = $data;
+        self::$cache[$this->settingsStoreScope][$id][self::cacheKeyDataSource] = $this->dataSource;
```
brusch:
```suggestion
        self::$cache[$this->settingsStoreScope][$id] = [
            'datasource' => $this->dataSource,
            'data' => $data,
        ];
```

This assumes `$data` is an array, it would also override a config node called `datasource`. so I'd wrap this in another array container. I think we can then as well live without having `datasource` as a constant 😊

brusch (changes requested):

lib/Model/Dao/PimcoreLocationAwareConfigDao.php
```diff
@@ -63,13 +63,20 @@ protected function getDataByName(string $id)
     {
         $this->id = $id;
 
-        if (isset(self::$cache[$this->settingsStoreScope][$id])) {
-            return self::$cache[$this->settingsStoreScope][$id];
+        if (isset(self::$cache[$this->settingsStoreScope][$id]) &&
+            isset(self::$cache[$this->settingsStoreScope][$id]['datasource'])) {
```
brusch:
```suggestion
```
This is obsolete as well, as `datasource` is always set if the cache key exists.

brusch (changes requested):

brusch:
Line `75`: 
```php
self::$cache[$this->settingsStoreScope][$id] = $data;
```
is also obsolete.

PR #10670, opened 2021-10-21
[Documents] Image editable - fix missing alt & title in frontend

ramundomario:
Regression of #7661

## Steps to reproduce 

- Set alt value in editmode for image editable 
- Alt & title attributes are empty when image is displayed in frontend

### DIFF
--- a/models/Asset/Image/Thumbnail.php
+++ b/models/Asset/Image/Thumbnail.php
@@ -371,8 +371,8 @@ final class Thumbnail
             }
         }
 
-        $altText = $options['alt'] ?? '';
-        $titleText = $options['title'] ?? '';
+        $altText = !empty($options['alt']) ? $options['alt'] : (!empty($attributes['alt']) ? $attributes['alt'] : '');
+        $titleText = !empty($options['title']) ? $options['title'] : (!empty($attributes['title']) ? $attributes['title'] : '');
 
         if (empty($titleText) && (!isset($options['disableAutoTitle']) || !$options['disableAutoTitle'])) {
             if ($image->getMetadata('title')) {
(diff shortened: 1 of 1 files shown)


dvesh3:
Sorry, not able to reproduce the issue. please pass image attributes as below:
```twig
{{ pimcore_image('pimcore-image',
    {
        thumbnail: 'content',
        imgAttributes: {
            title: 'title text',
            alt: 'alt text'
        }
    }) }}
```

ramundomario:
@dvesh3 
Issue refers to the value entered in editmode which should be printed by default if it is not specified when printing the editable:
`{{ pimcore_image('pimcore-image') }}`

Please see this example reproduced on demo:
Oc5pE5pgzX

dvesh3:
@ramundomario sorry, misunderstood the steps. It makes sense 👍

ramundomario:
@dvesh3 
Don't worry :+1: 
BTW: We should also handle alt and title passed via code as in your example.
Who should take precedence? The value entered in editmode or the one in the code?

dvesh3:
> BTW: We should also handle alt and title passed via code as in your example.
> Who` should take precedence? The value entered in editmode or the one in the code?

yes, true. I would say editmode takes the precedence since it is set by document author and the one defined in the code is used as a fallback.

ramundomario:
> > BTW: We should also handle alt and title passed via code as in your example.
> > Who` should take precedence? The value entered in editmode or the one in the code?
> 
> yes, true. I would say editmode takes the precedence since it is set by document author and the one defined in the code is used as a fallback.

Updated as discussed!
_Note_: used `!empty(...)` instead `??` since alt & text from editmode are empty string instead `null`.

dvesh3:
@ramundomario thanks a lot!

PR #10672, opened 2021-10-22
[Document] Areablock: hide/show button lacks status #10596

### LINKED ISSUE #10596: [Bug]: Hide/show button lacks status
### Expected behavior

Clicking on the strike-through'd eye should toggle the buttons status (and and preferably there should also be a tooltip describing what it does).

### Actual behavior

Clicking on it works in that it will hide/show the brick once the content is saved, but it has no status indicator if it is actually on or off.

### Steps to reproduce

Click the eye in the button bar:
image

karlozrilic:
Resolves #10596

## Additional info 
I also added tooltip with translation key to describe what hide/show button does


dvesh3:
@karlozrilic thanks!

PR #10681, opened 2021-10-25
[Asset] Pixel flood validation fix

brusch:


models/Asset.php
```diff
@@ -400,6 +405,18 @@ public static function create($parentId, $data = [], $save = true)
...
     }
 
+    private static function checkMaxPixels(string $localPath): void
+    {
+        $maxPixels = (int) \Pimcore::getContainer()->getParameter('pimcore.config')['assets']['image']['max_pixels'];
+        if($size = getimagesize($localPath)) {
+            $imagePixels = (int) ($size[0] * $size[1]);
+            if($imagePixels > $maxPixels) {
+                throw new ValidationException(
+                    'Image exceeds max pixel size of ' . $maxPixels . ', you can change the value in config pimcore.assets.image.max_pixels');
```
NiklasBr, line 415: This error message is fine for developers, but not for users. I'm not sure end users are going to understand, or able to resolve it given the hint.

NiklasBr:
To expand on the error message, we solved it like this in an Event Listener:

```php
    public function checkImageDimensions(Asset $asset)
    {
        $max = 15_000_000;
        $dimensions = getimagesizefromstring(stream_get_contents($asset->getStream()));
        $pixels = $dimensions[0] * $dimensions[1];
        if ($pixels > $max) {
            $diff = sqrt(1 + ($max / $pixels));
            $suggestion_0 = (int) round($dimensions[0] / $diff, -2, PHP_ROUND_HALF_DOWN);
            $suggestion_1 = (int) round($dimensions[1] / $diff, -2, PHP_ROUND_HALF_DOWN);
            $mp = $max/1_000_000;
            throw new UploadException("<p>Image dimensions of <em>{$asset->getFilename()}</em> are too large.</p>
<p>Max size: <code>{$mp}</code> <abbr title='Million pixels'>Megapixels</abbr></p>
<p>Suggestion: resize to <code>{$suggestion_0}&times;{$suggestion_1}</code> pixels or smaller</p>");
        }
    }
```

brusch:
Totally makes sense! Thanks for sharing @NiklasBr - incorporated here #10688

PR #10690, opened 2021-10-25
Check if $o exists before call method_exists

j-jonas:
otherwise u will get the error, if $o is not set
"method_exists(): Argument #1 ($object_or_class) must be of type object|string, null given"

  
  

Resolves #

### DIFF
--- a/lib/DataObject/GridColumnConfig/Operator/AnyGetter.php
+++ b/lib/DataObject/GridColumnConfig/Operator/AnyGetter.php
@@ -137,20 +137,24 @@ final class AnyGetter extends AbstractOperator
                     if (is_array($value)) {
                         $subValues = [];
                         foreach ($value as $o) {
-                            if ($this->attribute && method_exists($o, $getter)) {
-                                $subValues[] = $o->$getter($this->getParam1());
-                            } elseif ($this->attribute && method_exists($o, $fallbackGetter)) {
-                                $subValues[] = $o->$fallbackGetter($this->getParam1());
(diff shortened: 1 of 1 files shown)


lib/DataObject/GridColumnConfig/Operator/AnyGetter.php
```diff
@@ -137,20 +137,24 @@ public function getLabeledValue($element)
                     if (is_array($value)) {
                         $subValues = [];
                         foreach ($value as $o) {
-                            if ($this->attribute && method_exists($o, $getter)) {
-                                $subValues[] = $o->$getter($this->getParam1());
-                            } elseif ($this->attribute && method_exists($o, $fallbackGetter)) {
-                                $subValues[] = $o->$fallbackGetter($this->getParam1());
+                            if($o) {
```
dpfaffenbauer, line 140: if(!$o) {
  continue;
}

j-jonas: Maybe you have an array with 2 entries. What will happen, if the first is null and second is an object?

brusch:
@j-jonas Thanks a lot for your PR! Can you please rebase it onto the `10.2` branch, since it's a bug fix?

dvesh3:
@j-jonas pushed directly on 10.2 branch https://github.com/pimcore/pimcore/commit/4343980737a7031442847a8f94bcfbc80244dbd5. thanks!

PR #10694, opened 2021-10-26
Fix: skip CSRF check for all read methods

### LINKED ISSUE #10693: [Improvement]: Check CSRF token for write requests only
### Improvement description

The HEAD operation should not be checked for a CSRF token.

Some external software, we've encountered the issue with Excel, makes HEAD requests to embedded links. Our client uses Excel for some internal documents, and embeds deeplinks into their sheets. Our Sentry is lit up with exception reports over CSRF token issues: `Detected CSRF Attack! Do not do evil things with pimcore ... ;-)` on URIs like `/admin/login/deeplink?object_...`. The thing is, these are requests with http HEAD method, and any READ requests should not be checked against the CSRF token - the token is not provided anyway.

damijank:
Resolves #10693

## Additional info  
The change in this PR uses the Symfony Request's native method to resolve whether the incoming request is a READ or a WRITE operation. On any READ operation, the CSRF token check is skipped.

### DIFF
--- a/bundles/AdminBundle/EventListener/CsrfProtectionListener.php
+++ b/bundles/AdminBundle/EventListener/CsrfProtectionListener.php
@@ -71,7 +71,7 @@ class CsrfProtectionListener implements EventSubscriberInterface
 
         $this->csrfProtectionHandler->generateCsrfToken();
 
-        if ($request->getMethod() == Request::METHOD_GET) {
+        if ($request->isMethodCacheable()) {
             return;
         }
 


brusch:
@damijank Thanks a lot for your PR, totally makes sense! However, just one thing, could you please rebase your PR onto the `10.2` branch, since it's a bug fix?

damijank:
@brusch Hi, I've rebased onto 10.2. This also included other commits that were on the 10.x branch, can you confirm this is what you meant?

damijank:
Okay, fixed it and rebased again.

dvesh3:
@damijank thanks!

PR #10709, opened 2021-10-27
FIX(10699): fix no scroll in grid/list view of assets folder

### LINKED ISSUE #10699: [Bug]: Grid/list view fails to become scrollable due to img lazy-loading
### Expected behavior

When opening a grid/list view and it should be possible to scroll all the way down.

### Actual behavior

It appears that the lazy-loading of images interfere with how ExtJS calculates the content height.

### Steps to reproduce

Navigate assets in list view:
image

BabovicT:
Resolves #10699

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/helpers/grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/helpers/grid.js
@@ -199,7 +199,9 @@ pimcore.asset.helpers.grid = Class.create({
                         locked: this.getColumnLock(field),
                         renderer: function (value) {
                             if (value) {
-                                return '<img style="max-width: 200px; max-height: 100px;" src="' + value + '" loading="lazy">';
+                                return  '<div class="list-img-container">' +
+                                    '<img style="max-width: 150px; max-height: 100px; object-fit: contain;" src="' + value + '" loading="lazy">'+
+                                    '</div>'
                             }
                         }.bind(this)
                     });
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/helpers/grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/helpers/grid.js
@@ -200,7 +200,7 @@ pimcore.asset.helpers.grid = Class.create({
                         renderer: function (value) {
                             if (value) {
                                 return  '<div class="list-img-container">' +
-                                    '<img style="max-width: 150px; max-height: 100px; object-fit: contain;" src="' + value + '" loading="lazy">'+
+                                    '<img src="' + value + '" loading="lazy">'+
                                     '</div>'
                             }
                         }.bind(this)
--- a/bundles/AdminBundle/Resources/public/css/admin.css
+++ b/bundles/AdminBundle/Resources/public/css/admin.css
@@ -571,6 +571,12 @@
     width: 150px;
 }
 
+.list-img-container img {
+    max-width: 150px;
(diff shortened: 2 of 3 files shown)


dvesh3:
@BabovicT thanks for your PR! Unfortunately, the fix is not working.

brusch:
@BabovicT 🏓

BabovicT:
I hit the wall with this one and my colleague will continue today with this one. :ping_pong:

brusch:
@BabovicT Any updates? 😊 Thanks!

BabovicT:
@brusch we are unable to fix this. Collegue says that it is bug in ExtJS.
Only solution that does "work" is changing max-height of the cell the picture is to height and force the cell height to be "desired" size.
Problem is that there is no way to resize parent panel on image load or to know when all the images are done loading so that we refresh the parent panel. Parent panel is loaded and rendered with child grid that has no thumbnails and because there is no thumbnails every row height is set to  like 10-20px.
We also tried to add scroll but the scroll wasnt working. Its like something is overlapping with scroll and also when parent panel reactivates(go to any other tab inside folder and return to list view) there would be 2 scrolls.

NiklasBr:
I agree with @BabovicT that this is probably primarily a ExtJS issue. A regular HTML scroll/overflowing container would behave as expected and it appears Sencha has made a design decision not to allow ExtJS work like that. Unless this is fixed, would it be possible to disable the lazy-loading on assets in admin views?

brusch:
We introduced lazy loading for this reason: #10271 
So just removing it is probably not the best idea ... however what we could do is wrapping the image by a container with a fixed height, similar to what we do in the _Preview_ tab: 

image


WDYT?

BabovicT:
That solution will work. I just tried it with css from _Preview_ tab and it looks like this(I just changed height and width)
image

brusch:
@BabovicT looks good to me, can you update this PR accordingly? 
Maybe one thing, where is the left and right padding coming from, can we avoid that somehow? 
Thanks!

BabovicT:
@brusch padding is comming from .x-grid-cell-inner and every cell has it. 
I can add   'margin-right: -10px;' and  'margin-left: -10px;' to counter that padding.

brusch:
Yep, I think that would make sense, that we're not  wasting too much space.

BabovicT:
Just to be sure. Option 2 is preffered one?
1.
image
2. 
image

brusch:
@BabovicT Yep, looks good to me 👍

bundles/AdminBundle/Resources/public/js/pimcore/asset/helpers/grid.js
```diff
@@ -199,7 +199,9 @@ pimcore.asset.helpers.grid = Class.create({
                         locked: this.getColumnLock(field),
                         renderer: function (value) {
                             if (value) {
-                                return '<img style="max-width: 200px; max-height: 100px;" src="' + value + '" loading="lazy">';
+                                return  '<div class="list-img-container">' +
+                                    '<img style="max-width: 150px; max-height: 100px; object-fit: contain;" src="' + value + '" loading="lazy">'+
```
blankse: You can add this styles also to the admin.css `.list-img-container > img`

BabovicT: you are right.
Done

dvesh3 (approved):

dvesh3:
@BabovicT works like a charm now 👍...thanks a lot!

PR #10711, opened 2021-10-27
FIX(10708): fix for image assets that don't have set dimensions.

### LINKED ISSUE #10708: [Bug]: Can't open DWG file
### Expected behavior

- Open an asset with dwg extension. (DEBUG MODE)

### Actual behavior

- Can't open an asset with dwg extension without any error. ( Error appears only in debug mode)

### Steps to reproduce

- Set Debug mode on.
- Load an asset with dwg extension.
- Open an asset with dwg extension.
STEP_1
STEP_2
STEP_3
)
STEP_4

BabovicT:
Resolves #10708
Admin is no longer giving error, but as far as i can see .dwg is not supported by pimcore.

### DIFF
--- a/models/Asset/Image.php
+++ b/models/Asset/Image.php
@@ -407,7 +407,7 @@ EOT;
             return $dimensions['width'];
         }
 
-        return "";
+        return 0;
     }
 
     /**
@@ -421,7 +421,7 @@ EOT;
             return $dimensions['height'];
         }
 
-        return "";
+        return 0;
     }
 
     /**
--- a/models/Asset/Image.php
+++ b/models/Asset/Image.php
@@ -403,7 +403,11 @@ EOT;
     {
         $dimensions = $this->getDimensions();
 
-        return $dimensions['width'];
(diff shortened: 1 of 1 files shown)


models/Asset/Image.php
```diff
@@ -403,7 +403,11 @@ public function getWidth()
     {
         $dimensions = $this->getDimensions();
 
-        return $dimensions['width'];
+        if (isset($dimensions)) {
```
brusch:
```suggestion
        if ($dimensions) {
```
As `null` is returned in case the dimensions cannot be read, a simple `if` should be sufficient.

models/Asset/Image.php
```diff
@@ -413,7 +417,11 @@ public function getHeight()
     {
         $dimensions = $this->getDimensions();
 
-        return $dimensions['height'];
+        if (isset($dimensions)) {
```
brusch:
```suggestion
        if ($dimensions) {
```

brusch (changes requested):

PR #10712, opened 2021-10-27
FEATURE(10710): DataObject::getBy system properties

BabovicT:
Because key is unique only to folder not to the object, we can get more than 1 object from database.


brusch:
Maybe it's easier and more generic if we include it in `\Pimcore\Model\DataObject\Concrete::__callStatic()`, because then we could include some other system properties, such as `classId` or `className` ... ?

BabovicT:
@brusch Something like that?

models/DataObject/Concrete.php
```diff
@@ -709,10 +698,14 @@ public static function __callStatic($method, $arguments)
             return $list;
         }
 
-        // there is no property for the called method, so throw an exception
-        Logger::error('Class: DataObject\\Concrete => call to undefined static method ' . $method);
+        try {
+            return parent::$method($arguments);
```
danielczestki: We call static method from DataHub importer as:
$object = $className::$getter($data, 1);
so $arguments will be array with value: [$data, 1]

so to parent we should not send array but again
return parent::$method($data, 1);

danielczestki: should be: return call_user_func_array([parent::class, $method], $arguments);

BabovicT: Changed and pushed,,

jdreesen: Or maybe: `return parent::$method(...$arguments);`

BabovicT: Yea, i use that second one usually and usually forget to add ...
But they use call_user_func in their code so its more consistently with rest of code base

jdreesen: Both is used by pimcore... in the end it doesn't matter :)

models/DataObject/AbstractObject.php
```diff
@@ -79,6 +82,13 @@ abstract class AbstractObject extends Model\Element\AbstractElement
      */
     protected static $disableDirtyDetection = false;
 
+    /**
+     * @internal
+     *
+     * @var string[]
+     */
+    protected static $objectColumns = ['o_id', 'o_parentid', 'o_type', 'o_key', 'o_classid', 'o_classname'];
```
dvesh3: Any specific reason why `o_path` is missing?

BabovicT: You already did getByPath so I excluded it. 
If you think that it should be here then do i delete getByPath from abstract object?

dvesh3: I would just add it to the list so that you have both options to get object as we already have `getById` as well.

dvesh3: one more thing.. how getBy`o_parent`, `o_type`, `o_classid` & `o_classname` suppose to work? as this yield all objects

BabovicT: Same as other autogenerated methods from class. If there is one object u get that object back, if there is more then we return listing and null if there is no result.

BabovicT: I added o_path

dvesh3 (approved):

dvesh3:
@BabovicT thanks a lot!

PR #10717, opened 2021-10-28
FEATURE: Sites permission to allow user to edit sites

### LINKED ISSUE #10682: [Improvement][Site] Add Permission to create/edit/delete site
### Improvement description

right now, this is only allowed for admins, see 
https://github.com/pimcore/pimcore/blob/32e81d2b5b88262184236a6933e181b6cbd75b8f/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js#L790

BabovicT:
Resolves #10682

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/tree.js
@@ -787,7 +787,7 @@ pimcore.document.tree = Class.create({
                 });
             }
 
-            if (record.data.id != 1 && user.admin && record.data.type == "page") {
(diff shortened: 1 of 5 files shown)


dvesh3:
@BabovicT please add migration for adding new permission to existing installations. like https://github.com/pimcore/pimcore/blob/10.x/bundles/CoreBundle/Migrations/Version20201008091131.php#L39

Please note: This migration deletes the permission so please make sure to do otherwise.

BabovicT:
@dvesh3 like that?

dvesh3:
> @dvesh3 like that?

yes, looks good. thanks!

dvesh3 (approved):

PR #10719, opened 2021-10-28
[Data objects] Disable inheritance for single fields

### LINKED ISSUE #5416: Field Level Controlled Inheritance
Sometimes, you don't want certain fields to inherit but you want some to inherit. In that case, I always work with "dummy" values to break inheritance. (Thinks like a Metadata ManyToMany with Warehouses and Stock Quantity, a product can not inherit stock-values, since it is SKU based)

I think it should be quite easy possible for Pimcore to check on a field-definition level if inheritance is enabled or not, if not, don't inherit.

WDYT?

BlackbitDevs:
Sometimes there is the wish to enable inheritance for most fields but to exclude some others from being inherited. The suggestion in https://github.com/pimcore/pimcore/pull/6584#issuecomment-660097524 sadly is not enough to accomplish this because in the backend data is not loaded with inheritance through the getter methods but with 
https://github.com/pimcore/pimcore/blob/9bb42bc29e54493573aee21e197fe6efecdab991/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L596
and afterwards
https://github.com/pimcore/pimcore/blob/9bb42bc29e54493573aee21e197fe6efecdab991/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L681
This way it is not possible to exclude single fields from inheritance by overriding `\Pimcore\Model\DataObject\Concrete::getValueFromParent()` (as this is not used for loading backend data).

This PR adds a new parameter to `$fieldName` to `Concrete::getNextParentForInheritance()`. This way you can override this method in a data object class and implement something like
```php
public function getNextParentForInheritance($fieldName = null)
{
    if($fieldName === 'imageGallery'){
       return null;
    }
    return parent::getNextParentForInheritance($fieldName);
}
```

Resolves #5416


BlackbitDevs:
Failing test is not caused by this PR but something with Docker.

passioneight:
I feel like this should be a checkbox in the class editor. What do you think?

BlackbitDevs:
@passioneight Yes, me too but I first wanted the hear if this PR has a chance to get merged because in #6584 I already had it implemented with a checkbox (but with a different data model approach) and this was simply closed. To not waste that effort again, I would first like to hear some opinions of the Pimcore maintainers.

brusch:
For me this sounds like a good solution: 
- it gives the control over inheritance to the dev
- it doesn't break anything
- it doesn't require any additional configuration

@dvesh3 what's your opinion on that? Any concerns? 😊

brusch:
@BlackbitNeueMedien as this might be of general interest, could you please also add this possibility to the docs? Thanks in advance!

brusch:
This would resolve #5416

BlackbitDevs:
@brusch I will update the docs.
As this PR now seems to get accepted, shall I also add a checkbox to disable inheritance in the field definition as @passioneight suggested? Actually I can copy most of the JS code from #6584 for this.

brusch:
@BlackbitNeueMedien I'd leave it a coding only feature for now. We can think of a UI later on 😊

brusch:
Thank you very much! 🙏

brusch:
Reverted due to #10931

PR #10720, opened 2021-10-28
[Assets] Fix opening .zip file causes infinite requests

### LINKED ISSUE #10716: [Bug]: Assets - opening .zip file causes infinite requests
### Expected behavior

Zip Asset should be open

### Actual behavior

Infinite requests on opening zip asset and asset doesn't open 
image

### Steps to reproduce

Steps: 
Upload & Open a zip file in Admin interface

dvesh3:
Resolves #10716

## Additional info  
Additionally fixes:
 - Zero byte size file upload
 - Undefined array index "extensions" issue

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js
@@ -23,7 +23,7 @@ pimcore.asset.asset = Class.create(pimcore.element.abstract, {
             }.bind(this),
             params: {
                 id: this.id,
-                type: this.type
+                    type: this.type
             }
         });
     },
@@ -42,7 +42,7 @@ pimcore.asset.asset = Class.create(pimcore.element.abstract, {
             }
 
             if (this.type !== this.data.type) {
-                this.type = this.data.type;
+                Ext.MessageBox.alert(t("warning"), t("asset_type_changed"));
             }
 
             this.addTab();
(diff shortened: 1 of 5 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js
```diff
@@ -41,10 +41,8 @@ pimcore.asset.asset = Class.create(pimcore.element.abstract, {
                 throw "asset is locked";
             }
 
-            if(this.type !== this.data.type) {
-                pimcore.helpers.closeAsset(this.id);
-                pimcore.helpers.openAsset(this.id, this.data.type);
-                return;
+            if (this.type !== this.data.type) {
```
brusch, line 44: @dvesh3 Should we really change the type on the fly? As `type` is kind of a static variable that is defined by the actual implementation, see e.g.: 
https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/asset/audio.js#L21

Actually if the type in the UI doesn't match with the one sent from the server, there's something pretty wrong ... but maybe we should show an error message instead, that recommends to reload the admin ui? wdyt?

brusch (changes requested):

bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js
dvesh3: yes, I think a warning to reload the admin UI is a better idea than silently changing the type on the fly. This could lead to unwanted behavior in admin interface, if the response is not correct from the server for some reason.

Good suggestion! I'll update it 😉

dvesh3: However, we have to fix the type for zip archives response in the server as it returns unknown at the moment and then will end up with a warning every time you upload the zip file.

dvesh3: Added archive js type(to be consistent with models) and a warning message when the asset type is changed in the server.

PR #10721, opened 2021-10-28
Follow-Up PR#7177: Add Pimcore JS events to enable virtual data folder integration 

valentin-ki:
This is a follow-up Pull-Request for the PR https://github.com/pimcore/pimcore/pull/7177  from @andreas-gruenwald :

## Feature Request 

We are working on a bundle that allows developers to integrate folders and objects as part of a virtual data object tree.

The idea is very similar to the AlternateObjectTrees bundle, but the implementation

focuses more on the developer perspective,
and the UI integration into Pimcore's tree structure is more tightly.
Preview:

image

In order to support the integration for data object trees and the children grid view, this PR adds additional JS events that make the core more flexible.

We tried to keep the changes as minimalistic as possible.

## Additional info  

Original Contributer: @andreas-gruenwald

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/folder.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/folder.js
@@ -47,10 +47,14 @@ pimcore.object.folder = Class.create(pimcore.object.abstract, {
 
 
     getData: function () {
+
+        var eventData =  {requestParams: {id: this.id}};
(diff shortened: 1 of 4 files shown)


andreas-gruenwald:
As @valentin-ki wrote, this is a follow up PR.
Basically nothing has changed, but the PR is now based on PimcoreX while the old one was based on Pimcore 6.

dvesh3:
@valentin-ki thanks!

PR #10723, opened 2021-10-28
FIX(10600): update object_url_slug to correct character set and collate

### LINKED ISSUE #10600: [Bug]: Incorrect collation in column
### Expected behavior

`SELECT * FROM object_url_slugs WHERE slug = '🐌';` will not cause any errors.

### Actual behavior

`SELECT * FROM object_url_slugs WHERE slug = '🐌';` causes `pimcore.ERROR: PDOException: SQLSTATE[HY000]: General error: 1267 Illegal mix of collations (utf8_bin,IMPLICIT) and (utf8mb4_general_ci,COERCIBLE) for operation '='`.

> Occurred on Pimcore 6, reproducible on Pimcore X.

Relates to:
- https://github.com/pimcore/pimcore/issues/6698,
    - which also references an [apparent patch via a script](https://gist.github.com/podarcis/cdce3371b8149e86924e9bb5f91e0961), 
    - and resulted in a [documentation update]( https://github.com/pimcore/pimcore/pull/6773/files), which is [no longer available](https://pimcore.com/docs/4.6.x/Development_Documentation/Installation_and_Upgrade/Upgrade_Notes/Witin_V4.html)

Caused by:
- a migration for existing instances https://github.com/pimcore/pimcore/blob/08d0fd81e419909d4cac0bed59f4b0682a96c2b6/bundles/CoreBundle/Resources/migrations/Version20200121095650.php#L23
- the installer for new instances https://github.com/pimcore/pimcore/blob/eecf55da3317d82cc5f2b8b80c1f57f5c3257d38/bundles/InstallBundle/Resources/install.sql#L950

### Steps to reproduce

- Install Pimcore X.
- Execute `SELECT * FROM object_url_slugs WHERE slug = '🐌';`.

> You can check the collation of the `slug` column by executing `SHOW FULL COLUMNS FROM object_url_slugs;`.

### Solution for existing projects
Execute `ALTER TABLE object_url_slugs MODIFY slug VARCHAR(765) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;`.

### Solution for new projects
Update this line: https://github.com/pimcore/pimcore/blob/eecf55da3317d82cc5f2b8b80c1f57f5c3257d38/bundles/InstallBundle/Resources/install.sql#L950

BabovicT:
Resolves #10600

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20211028155535.php
+++ b/bundles/CoreBundle/Migrations/Version20211028155535.php
@@ -1,0 +1,41 @@
+<?php
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * @internal
+ */
+final class Version20211028134037 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema): void
+    {
+        $this->addSql("ALTER TABLE object_url_slugs MODIFY slug VARCHAR(765) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci;");
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema): void
+    {
+        $this->addSql("ALTER TABLE object_url_slugs MODIFY slug VARCHAR(765) CHARACTER SET utf8 COLLATE utf8_bin;");
+    }
+}
--- a/bundles/CoreBundle/Migrations/Version20211028155535.php
+++ b/bundles/CoreBundle/Migrations/Version20211028155535.php
@@ -21,7 +21,7 @@ use Doctrine\Migrations\AbstractMigration;
 /**
  * @internal
  */
-final class Version20211028134037 extends AbstractMigration
+final class Version20211028155535 extends AbstractMigration
 {
     /**
      * @param Schema $schema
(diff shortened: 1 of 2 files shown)


bundles/CoreBundle/Migrations/Version20211028155535.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * @internal
+ */
+final class Version20211028134037 extends AbstractMigration
```
dvesh3:
```suggestion
final class Version20211028155535 extends AbstractMigration
```

bundles/InstallBundle/Resources/install.sql
```diff
@@ -947,7 +947,7 @@ CREATE TABLE `object_url_slugs` (
       `ownertype` ENUM('object','fieldcollection','localizedfield','objectbrick') NOT NULL DEFAULT 'object',
       `ownername` VARCHAR(70) NOT NULL DEFAULT '',
       `position` VARCHAR(70) NOT NULL DEFAULT '0',
-      `slug` varchar(765) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL, /* slug in utf8 (3-byte) using the full key length of 3072 bytes */
+      `slug` varchar(765) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL, /* slug in utf8 (3-byte) using the full key length of 3072 bytes */
```
dvesh3: Comment should be changed as well.

BabovicT: like that?

dvesh3: Also the charset to `utf8mb4`. thanks!

BabovicT: There. Sry about that.

bundles/InstallBundle/Resources/install.sql
```diff
@@ -947,7 +947,7 @@ CREATE TABLE `object_url_slugs` (
       `ownertype` ENUM('object','fieldcollection','localizedfield','objectbrick') NOT NULL DEFAULT 'object',
       `ownername` VARCHAR(70) NOT NULL DEFAULT '',
       `position` VARCHAR(70) NOT NULL DEFAULT '0',
-      `slug` varchar(765) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL, /* slug in utf8 (3-byte) using the full key length of 3072 bytes */
+      `slug` varchar(765) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci NOT NULL, /* slug in utf8mb4 (4-byte) using the full key length of 4096 bytes */
```
brusch:
```suggestion
      `slug` varchar(765) NOT NULL, /* slug in utf8mb4 (4-byte) using the full key length of 3072 bytes */
```
We can omit setting the charset and collation here, as it is the default anyway. Also the max key length doesn't change, it's a MySQL limitation.

brusch (changes requested):

brusch:
Thanks a lot! 👍

PR #10725, opened 2021-10-28
[Maintenance] Add possibility to exclude maintenance handler jobs 

### LINKED ISSUE #10714: [Bug]: In 10.2 pimcore:maintenance no longer excludes excludedJobs
### Expected behavior

The command `bin/console maintenance --excludedJobs=logmaintenance,cleanuplogfiles,imageoptimize` should not do any of the following:
- logmaintenance
- cleanuplogfiles
- imageoptimize

### Actual behavior

They are all being executed anyway

### Steps to reproduce

Run the command in 10.1:
 - All the above jobs are excluded as expected

Run the command in 10.2:
- All jobs are executed

dvesh3:
Resolves #10714


dpfaffenbauer:
Feels a bit hacky... maybe there might be a better solution with messenger envelopes/stamps? not sure, just thinking out loud

dvesh3:
> Feels a bit hacky... maybe there might be a better solution with messenger envelopes/stamps? not sure, just thinking out loud

I was looking for a better solution in Symfony Docs, but couldn't find it. I'll give it a try with Messenger Envelopes. Thanks!

lib/Messenger/Middleware/MaintenanceMessageMiddleware.php
```diff
@@ -0,0 +1,51 @@
...
+    public function addSkipMessage($message): void
+    {
+        $this->skipMessages[] = $message;
+    }
+
+    public function handle(Envelope $envelope, StackInterface $stack): Envelope
+    {
+        /** @var MaintenanceTagStamp $stamp */
+        if (null !== $stamp = $envelope->last(MaintenanceTagStamp::class)) {
+            if (in_array($stamp->getTag(), array_values($this->skipMessages))) {
```
jdreesen:
```suggestion
            if (in_array($stamp->getTag(), $this->skipMessages, true)) {
```

dvesh3:
> Feels a bit hacky... maybe there might be a better solution with messenger envelopes/stamps? not sure, just thinking out loud

@brusch @dpfaffenbauer Changed approach to use stamps (for maintenance tags) with middleware implementation which checks if any message should be skipped. And if there's any then it assigns `DispatchAfterCurrentBusStamp` to the message envelop and continues.

lib/Messenger/Middleware/MaintenanceMessageMiddleware.php
```diff
@@ -0,0 +1,52 @@
...
+
+    public function handle(Envelope $envelope, StackInterface $stack): Envelope
+    {
+        /** @var MaintenanceTagStamp $stamp */
+        $stamp = $envelope->last(MaintenanceTagStamp::class);
+        if (null !== $stamp) {
+            if (in_array($stamp->getTag(), $this->skipMessages, true)) {
+                return $envelope->with(new DispatchAfterCurrentBusStamp());
+            }
+        }
```
jdreesen:
```suggestion
        if (null !== $stamp && in_array($stamp->getTag(), $this->skipMessages, true)) {
            return $envelope->with(new DispatchAfterCurrentBusStamp());
        }
```

brusch:
I'm just thinking ... maybe it's even better if we put those tasks into a dedicated transport. 
That would allow you to "exclude" image optimizations even if you're running `pimcore:maintenance --aync` and  `messenger:consume` using Supervisor ... so gives you more flexibility. Additionally less code and better maintainable. 

Wdyt? 

Example: 
```yml
framework: 
    messenger:
        transports:
            pimcore_core: "doctrine://default"
            pimcore_maintenance: "doctrine://default"
            pimcore_image_optimizer: "doctrine://default"
        routing:
            'Pimcore\Messenger\OptimizeImageMessage': pimcore_image_optimizer
```

dvesh3:
Closed in favor of https://github.com/pimcore/pimcore/pull/10947

@brusch this is now handled by introducing a config `pimcore.maintenance.exclude_messages`, where a user can defined messages which shouldn't be dispatched for the messenger bus. Otherwise these messages will stay in the database and never gets consumed.

PR #10728, opened 2021-10-28
add info that %Text( paramname) deprecated

frithjof:
Resolves #


doc/Development_Documentation/19_Development_Tools_and_Details/25_Email_Framework/README.md
```diff
@@ -75,6 +75,10 @@ $mail->setParams($params);
 $mail->send();
 ```
 
+you can access the parameters in your mail content now in standard twig syntax instead of the old %Text(paramname);
+```php
```
blankse:
```suggestion
```twig
```

doc/Development_Documentation/19_Development_Tools_and_Details/25_Email_Framework/README.md
```diff
@@ -75,6 +75,10 @@ $mail->setParams($params);
 $mail->send();
 ```
 
+you can access the parameters in your mail content now in standard twig syntax instead of the old %Text(paramname);
```
brusch:
```suggestion
you can access the parameters in your mail content in standard twig syntax.
```
Since this is the current documentation, we don't need to mention the old (and removed) syntax here.

brusch (changes requested):

doc/Development_Documentation/19_Development_Tools_and_Details/25_Email_Framework/README.md
frithjof: I would mention it because it is still in the image above used

brusch: I think we should then update the image as well. Doesn't make sense to have that still in the image if it doesn't work anymore 😊

frithjof:
image is now updated too and added example for using an Object

brusch:
Very nice, thanks a lot! 👍

PR #10733, opened 2021-10-29
[DataObject] don't increase the version count when versioning is disabled

fabfischer:
I noticed that the version count was increased, even with versioning disabled. the small fix/check honors such setting now:
```
if (Model\Version::$disabled === false) {
    $this->setVersionCount($this->getDao()->getVersionCountForUpdate() + 1);
}
```

## Additional info  
fixed one minor typo too

### DIFF
--- a/models/Element/AbstractElement.php
+++ b/models/Element/AbstractElement.php
@@ -49,7 +49,9 @@ abstract class AbstractElement extends Model\AbstractModel implements ElementInt
      */
     protected function updateModificationInfos()
     {
-        $this->setVersionCount($this->getDao()->getVersionCountForUpdate() + 1);
+        if (Model\Version::isEnabled() === true) {
+            $this->setVersionCount($this->getDao()->getVersionCountForUpdate() + 1);
+        }
 
         if ($this->getVersionCount() > 4200000000) {
             $this->setVersionCount(1);
--- a/models/Version.php
+++ b/models/Version.php
@@ -168,14 +168,6 @@ final class Version extends AbstractModel
     /**
      * @return bool
      */
-    public static function isDisabled(): bool
-    {
-        return (self::$disabled === true);
-    }
-
-    /**
-     * @return bool
-     */
     public static function isEnabled(): bool
     {
         return (self::$disabled === false);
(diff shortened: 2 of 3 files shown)


models/Element/AbstractElement.php
```diff
@@ -49,7 +49,9 @@ abstract class AbstractElement extends Model\AbstractModel implements ElementInt
      */
     protected function updateModificationInfos()
     {
-        $this->setVersionCount($this->getDao()->getVersionCountForUpdate() + 1);
+        if (Model\Version::$disabled === false) {
```
brusch: I'd prefer introducing a static getter for that. Can you change the PR accordingly? Thanks in advance!

brusch (changes requested):

models/Element/AbstractElement.php
fabfischer: Hi, 
thanks for your feedback. I hope I understood you correctly - I added two static methods that tell you if versioning is enabled or not:

```php
# models/Element/AbstractElement.php
if (Model\Version::isEnabled() === true) {
    $this->setVersionCount($this->getDao()->getVersionCountForUpdate() + 1);
}
```

```php
# models/Version.php
/**
* @return bool
*/
public static function isDisabled(): bool
{
    return (self::$disabled === true);
}

/**
* @return bool
*/
public static function isEnabled(): bool
{
    return (self::$disabled === false);
}
```

models/Version.php
```diff
@@ -165,6 +165,22 @@ public static function enable()
...
     }
 
+    /**
+     * @return bool
+     */
+    public static function isDisabled(): bool
+    {
+        return (self::$disabled === true);
+    }
+
```
brusch:
```suggestion
```

brusch: `isEnabled` should be sufficient, can be used with `!`.

models/Version.php
```diff
@@ -165,6 +165,14 @@ public static function enable()
         self::$disabled = false;
     }
 
+    /**
+     * @return bool
+     */
+    public static function isEnabled(): bool
+    {
+        return (self::$disabled === false);
```
brusch:
```suggestion
        return !self::$disabled;
```

brusch:
Thanks! 👍

PR #10734, opened 2021-10-29
[DataObject] [6.9] don't increase the version count when versioning is disabled

fabfischer:
see [PR-10733](https://github.com/pimcore/pimcore/pull/10733) for Pimcore 10.2.
We noticed the bug with Pimcore 6.9, which we are currently still using. But I fixed it for both versions, with two PRs


brusch:
I'm sorry, `6.9` is already EOL.

PR #10739, opened 2021-11-02
[eCommerceFramework] Fixed token filtering in VoucherSeries details tab

ThomasKeil:
If you try to filter for tokens in the details tab on a OnlineShopVoucherSeries, you won't get any results unless you provide from- and to-dates.
This bug was introduced with https://github.com/pimcore/pimcore/pull/7482/commits/72e61b1024fefe2dde370e0136fec078cc7d66ac and is fixed with this patch

### DIFF
--- a/bundles/EcommerceFrameworkBundle/VoucherService/Token/Listing.php
+++ b/bundles/EcommerceFrameworkBundle/VoucherService/Token/Listing.php
@@ -67,7 +67,7 @@ class Listing extends \Pimcore\Model\Listing\AbstractListing implements Paginate
                 $this->addConditionParam('length = ?', $filter['length']);
             }
 
-            if (isset($filter['creation_from']) && $filter['creation_from'] !== '') {
+            if (!empty($filter['creation_from'])) {
                 $this->addConditionParam("DATE(timestamp) >= STR_TO_DATE(?,'%Y-%m-%d')", $filter['creation_from']);
             }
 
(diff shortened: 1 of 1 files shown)


bundles/EcommerceFrameworkBundle/VoucherService/Token/Listing.php
```diff
@@ -67,11 +67,11 @@ public function setFilterConditions($seriesId, $filter = [])
                 $this->addConditionParam('length = ?', $filter['length']);
             }
 
-            if (isset($filter['creation_from'])) {
+            if (isset($filter['creation_from']) && $filter['creation_from'] !== '') {
```
brusch:
```suggestion
            if (!empty($filter['creation_from'])) {
```
That should be enough as well, right?

bundles/EcommerceFrameworkBundle/VoucherService/Token/Listing.php
```diff
@@ -67,11 +67,11 @@ public function setFilterConditions($seriesId, $filter = [])
...
                 $this->addConditionParam("DATE(timestamp) >= STR_TO_DATE(?,'%Y-%m-%d')", $filter['creation_from']);
             }
 
-            if (isset($filter['creation_to'])) {
+            if (isset($filter['creation_to']) && $filter['creation_to'] !== '') {
```
brusch:
```suggestion
            if (!empty($filter['creation_to'])) {
```

brusch (changes requested):

bundles/EcommerceFrameworkBundle/VoucherService/Token/Listing.php
ThomasKeil: Yes, it should - the additional check was to solve the problem like already existing line 62:

`if (isset($filter['usages']) && $filter['usages'] !== '') {`

dvesh3 (approved):

dvesh3:
thanks @ThomasKeil

PR #10750, opened 2021-11-03
Website settings in multi site environment are not fetched correctly

mittererr:
In my case a custom website setting "contactFormEmailRecipient" is set for main site and a sub site. For the sub site the according domain is selected in website settings.
The `$data` array holds two results and later, for the sub site `$data[0]` is taken. This is not correct. With this fix, `$data` holds only one result and everything is fine.

Bug fix: 6.9


brusch:
Thanks for your PR, however, `6.9` is already end of life, so we're also not accepting any further PRs for that branch.

PR #10756, opened 2021-11-03
[Bugfix]: Links having long URLs not editable in data objects

### LINKED ISSUE #10753: [Bug]: Links having long URLs not editable in data objects
### Expected behavior

Edit links as expected

### Actual behavior

As the buttons are positioned using js, they get cut off when used long URLs:

image

Replaced by a shorter URL:

image

### Steps to reproduce

Use a long URL in a data object link field

karlozrilic:
Resolves #10753


bundles/AdminBundle/Resources/public/js/pimcore/object/tags/link.js
```diff
@@ -90,6 +90,10 @@ pimcore.object.tags.link = Class.create(pimcore.object.tags.abstract, {
             text = this.data.path;
         }
 
+        if (text.length > 100) {
+            text = text.slice(0, 97)
+            text += '...'
```
brusch:
```suggestion
            text += '…'
```
Let's use ellipsis here 😉

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/link.js
```diff
@@ -165,6 +169,11 @@ pimcore.object.tags.link = Class.create(pimcore.object.tags.abstract, {
...
```
brusch:
```suggestion
            text += '…'
```

brusch (changes requested):

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/link.js
NiklasBr: 👍 … and in that case it is also possible to use `text.slice(0, 99)`

brusch:
Thanks! 🙏

PR #10758, opened 2021-11-03
Classification store groups and keys sorting by groupId and keyId not working properly

dturopoli:
Inside Classification Store tab on overview of all the groups inside a collection and all the keys inside the group sorting by groupId and KeyId  is not working properly after you add groups/keys and refresh page.

Screenshot from 2021-11-03 14-59-19

Screenshot from 2021-11-03 14-59-11


brusch:
Thanks! 👍

PR #10761, opened 2021-11-04
DocumentFallbackListener

ctippler:
2 Issues:

First: When a document is requested from a static route e.g /de/shop a document resolver in an Subscriber can't get the document, because it is never set. 
This is fixed by 
`
$this->documentResolver->setDocument($event->getRequest(), $this->fallbackDocument);
`
Second:
If we run into a Error page the fallback Document is not set because 
`
 if ($this->fallbackDocument && $event->isMainRequest()) {
`
is not fulfilled.
$event->isMainRequest() is false, so the fallback document is not set.

The following line solves it, but i don't know if its correct ;-)
`
if ($this->fallbackDocument && !$this->documentResolver->getDocument()) {
`


brusch:
The fallback document is resolves on `kernel.request`  but set on the request on `kernel.controller`. 
See: https://github.com/pimcore/pimcore/blob/61d92fa0d8e3d9ff1d35bbc7e58f47b8a0e677dd/bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php#L159
So your subscriber would need to listen on `kernel.controller` in order to get the right value from the document resolver. 

That has several reasons, see also: #6227 #6203 
So I'm afraid we can't change that as it would break many other things.

ctippler:
Hi,
thx for the hint. It was allready listening on the kernel.controller event but the priority of the subscriber was too high and some weird project specific implementations caused the problem. Got it to work now :-).

PR #10764, opened 2021-11-04
Fix return types / php docs

blankse:
PhpStan throws errors if you check for null value with the ?? operator but the phpdoc says it can't be null.
I now decided not to return a null value, because it is partially the case that the default value is already 0, '' or is casted in the getter

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -94,9 +94,9 @@ class Asset extends Element\AbstractElement
     /**
      * @internal
      *
-     * @var string
+     * @var string|null
      */
-    protected $filename = '';
+    protected $filename;
 
     /**
      * @internal
@@ -1146,7 +1146,7 @@ class Asset extends Element\AbstractElement
     }
 
     /**
-     * @return string
+     * @return string|null
      */
     public function getFilename()
     {
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -68,14 +68,14 @@ class Asset extends Element\AbstractElement
      *
      * @var int
      */
-    protected $id;
+    protected $id = 0;
 
     /**
      * @internal
      *
      * @var int
      */
-    protected $parentId;
+    protected $parentId = 0;
 
     /**
      * @internal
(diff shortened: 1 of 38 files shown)


models/Asset.php
```diff
@@ -213,14 +213,14 @@ class Asset extends Element\AbstractElement
      *
      * @var bool
      */
-    protected $_dataChanged = false;
+    protected $dataChanged = false;
```
jdreesen, line 216: Isn't this a BC break, as the property is `protected`?

blankse: Is a internal property

brusch:
Hmm using `0` instead of `null` feels not right to me. Where exactly is the issue? Can't we solve it a different way?

blankse:
@brusch Than we should change the phpdocs to int|null and remove the 0 default values and castings in the getter to be consistant?

```
$asset = new Asset()
$asset->getId() // 0 because of casting in getter
$asset->getParentId() // null but phpdoc says int
$asset->getKey() || $asset->getFilename() // '' because of casting in getter

$object = new DataObject();
$object->getId() // 0 because of default value 0 on attribute
$object->getParentId() // null but phpdoc says int
$object->getKey() // null but phpdoc says string

$document = new Document();
$document->getId() // 0 because of casting in getter
$document->getParentId() // null but phpdoc says int
$document->getKey() // null but phpdoc says string
```

When you use `$element->getKey() ?? ''` because you have a return typehint to string and have an error with null you get an PhpStan error because getKey should only return string (phpdoc)

blankse:
@brusch I make id, key, parentId, path nullable. I also think this is better :)

models/Tool/CustomReport/Config.php
```diff
@@ -70,62 +70,62 @@ class Config extends Model\AbstractModel implements \JsonSerializable
...
-     * @var int
+     * @var int|null
      */
     protected $creationDate;
 
     /**
      * @var bool
      */
-    protected $shareGlobally;
+    protected $shareGlobally = true;
```
brusch, line 118:
```suggestion
    protected $shareGlobally = false;
```
I'd set this `false` to not change the behavior.

brusch (changes requested):

models/Tool/CustomReport/Config.php
blankse: If you create a CustomReport in the backend the checkbox is checked by default. So I think it would be good to have the same behavior in PHP. But I can change it.

brusch: @blankse ah, sorry, I thought it's off by default - thanks for clarifying, nothing else to change then 😉 Thanks!

blankse: Oh, is not checked... I change it. Hmm I tought I checked this...

brusch: 👍 just wanted to double check as well 😊 Thanks!

brusch: For me it's enabled by default 🤔 So seems ok to me.

blankse: Okay, in 6.9 it was unchecked. In 10.2 it is checked by default. :)

PR #10765, opened 2021-11-04
Fix GenerateWeb2PrintPdfHandler: Add missing tag

### LINKED ISSUE #10655: [Bug]: Regression with web2print : no pdf gets generated for print documents
### Expected behavior

Generate & preview pdf documents for a print document or container should produce a pdf, which ever the tool used (PDFreactor, WkHtmlToPdf...)

### Actual behavior

The progression bar remains to 0% and nothing gets generated.
image

### Steps to reproduce

Problem is found on new Pimcore installation using 10.2.x and ALSA on the demo https://demo.pimcore.fun/.

blankse:
Fix #10655

### DIFF
--- a/bundles/CoreBundle/Resources/config/message_handler.yaml
+++ b/bundles/CoreBundle/Resources/config/message_handler.yaml
@@ -30,7 +30,6 @@ services:
         tags:
             - { name: messenger.message_handler }
 
-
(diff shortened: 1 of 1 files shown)


brusch:
Thank you very much! 👍

PR #10769, opened 2021-11-05
Fix missing translation show_hide_areablock

blankse:
Is added to es instead of en https://github.com/pimcore/pimcore/commit/1055e75d25728a2a6768ea8c52aaee117b41495b

image


dvesh3:
Good catch 😄 thanks!

PR #10772, opened 2021-11-05
Fixes incorrect default value casting in QuantityValues

### LINKED ISSUE #10754: [Bug]: Classification store inheritance is broken
### Expected behavior

According to [the docs](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Data_Types/Classification_Store.html#page_Inheritance) a variant should be able to inherit data from a parent object.

### Actual behavior

Certain fields appear to not inherit properly using the standard getters *OR* the documentation is missing.

### Steps to reproduce

1. Add data to a parent object:
image

2. Make sure it is visible on the variant:
image

3. Try to get the data in a controller so that it can be sent to the front end, but none of the expected methods work:

`$classificationStore->getActiveGroups()` does not return this data
`$classificationStore->getGroupIdsWithData()` does not return this data
`$classificationStore->getItems()` does not return this data either.

NiklasBr:
Resolves #10754

When `defaultValue` was for example an empty string it would be cast to `0.0` causing discrepancy between what was shown in the admin UI and other locations. This PR resolves the issue.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/QuantityValue.php
+++ b/models/DataObject/ClassDefinition/Data/QuantityValue.php
@@ -167,7 +167,7 @@ class QuantityValue extends Data implements ResourcePersistenceAwareInterface, Q
      */
     public function setDefaultValue($defaultValue)
     {
-        if (!empty($this->defaultValue)) {
+        if (!empty($defaultValue)) {
             $this->defaultValue = (float) $defaultValue;
         } else {
             $this->defaultValue = null;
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -167,7 +167,7 @@ public function getDefaultValue()
      */
     public function setDefaultValue($defaultValue)
     {
-        if ($defaultValue !== null) {
+        if (!empty($this->defaultValue)) {
```
dvesh3:
```suggestion
        if (!empty($defaultValue)) {
```

NiklasBr: @dvesh3 I did what you wanted, but I am unsure if that is *actually* correct. Can you describe why?

dvesh3:
thanks @NiklasBr!

PR #10781, opened 2021-11-08
[Docs] Add doc: Style backend depending on the application environment

podarcis:
This adds documentation for the feature implemented in #9794.


brusch:
Thank you very much!

PR #10782, opened 2021-11-08
using reference to edit cmd variable

j-jonas:
all of the calls of this function do not overwrite the variable, they all look like:

Console::addLowProcessPriority($command);
instead of:
$command = Console::addLowProcessPriority($command);

without & nothing will happen

  
  

Resolves #

### DIFF
--- a/lib/Tool/Console.php
+++ b/lib/Tool/Console.php
@@ -364,7 +364,7 @@ final class Console
      *
      * @return void
      */
-    public static function addLowProcessPriority(&$cmd): void
+    public static function addLowProcessPriority(&$cmd)
     {
         $nice = (string) self::getExecutable('nice');
         if ($nice) {
--- a/lib/Tool/Console.php
+++ b/lib/Tool/Console.php
@@ -364,7 +364,7 @@ final class Console
      *
      * @return void
      */
-    public static function addLowProcessPriority(&$cmd)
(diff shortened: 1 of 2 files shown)


lib/Tool/Console.php
```diff
@@ -364,7 +364,7 @@ protected static function execInBackgroundWindows($cmd, $outputFile)
      *
      * @return array|string
      */
-    public static function addLowProcessPriority($cmd)
+    public static function addLowProcessPriority(&$cmd)
```
dvesh3: maybe we get rid of return statement since it is not used anywhere and it is not a BC break as the method is marked as internal...could you please change it? thanks!

j-jonas: return value is removed

dvesh3:
@j-jonas correct approach would be to fix the usage here https://github.com/pimcore/pimcore/blob/10.x/models/Asset/Image.php#L112 so it doesn't depend on the return value.

j-jonas:
sry, i didnt see the error. i fixed it. hopefully it is alright now.

dvesh3:
> sry, i didnt see the error. i fixed it. hopefully it is alright now.

no worries. Looks good now 👍

dvesh3:
@j-jonas thanks!

PR #10784, opened 2021-11-08
Fix version display when an object stored in an older version does no…

mhgtlc:
Fix: When trying to display a version that contains reference to an object that is already deleted, it fails to display the version


brusch:
@mhgtlc thanks for your PR! However, I think we should make that consistent for all reference data-types: 
- AdvancedManyToManyRelation
- Hotspotimage
- Image
- ImageGallery
- Link
- ManyToManyObjectRelation
- ManyToManyRelation
- ManyToOneRelation
- Video 

For all data-types extending from `AbstractRelations` I could think of a generic solution, that also returns a translated text. The text should be also not that offensive (no or just one exclamation mark and not in capital letters). 

Could you change your PR accordingly? 
Thanks in advance!

brusch:
@mhgtlc any update? would be nice if you could change your PR accordingly. Thanks in advance!

mhgtlc:
> @mhgtlc any update? would be nice if you could change your PR accordingly. Thanks in advance!

Sorry for the late response, no, sorry I do not have time

brusch:

PR #10785, opened 2021-11-08
Label Width: Remove wrong width explanation

blankse:
Since https://github.com/pimcore/pimcore/pull/9780 this explanation is wrong. Can only be integer

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/localizedfields.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/localizedfields.js
@@ -179,11 +179,6 @@ pimcore.object.classes.data.localizedfields = Class.create(pimcore.object.classe
                     value: this.datax.labelWidth
                 },
                 {
-                    xtype: "displayfield",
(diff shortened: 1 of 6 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/localizedfields.js
```diff
@@ -178,11 +178,6 @@ pimcore.object.classes.data.localizedfields = Class.create(pimcore.object.classe
                     fieldLabel: t("label_width"),
                     value: this.datax.labelWidth
                 },
-                {
-                    xtype: "displayfield",
-                    hideLabel: true,
-                    value: t('width_explanation')
```
jdreesen, line 184: Can the associated translation be removed? Or is it still used somewhere else?

blankse: Is still used in the width/height fields, where % values are possible

brusch:
Thanks a lot!

PR #10792, opened 2021-11-09
Provide a Tool\Sorter class and use it for array sorting functions as callback

NiklasBr:
This PR started as a "housekeeping" thing to make all the sorters as static functions and use the spaceship operator. But then I got the idea that perhaps all these could be collected into a single utility class because there were a lot of overlap (code re-use yay!).

I also noticed think this could be the starting point for a wider normalization of array shapes and serve as a reminder that every time a new array (or object) structure is introduced which is intended to be sorted by PHP its sortable attribute should conform to existing ones.

Thoughts?

### DIFF
--- a/lib/Tool/Sorter.php
+++ b/lib/Tool/Sorter.php
@@ -1,0 +1,155 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Tool;
+
+use Pimcore\Bundle\AdminBundle\GDPR\DataProvider\DataProviderInterface;
+use Pimcore\Cache\Core\CacheQueueItem;
+use Pimcore\HttpKernel\BundleCollection\ItemInterface;
+use Pimcore\Model\Element\Tag;
+use Pimcore\Model\Staticroute;
+use Pimcore\Targeting\Model\TargetGroupAssignment;
+use Pimcore\Workflow\WorkflowConfig;
+
+/**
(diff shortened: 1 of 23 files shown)


lib/Tool/Sorter.php
```diff
@@ -0,0 +1,155 @@
...
+
+    public static function priority(array $a, array $b): int
+    {
+        return $a['priority'] <=> $b['priority'];
+    }
+
+    public static function sortIndex(array $a, array $b): int
+    {
+        return $a['sortIndex'] <=> $b['sortIndex'];
+    }
```
jdreesen, line 68: Those could be replaced by a ["higher-order" function](https://en.wikipedia.org/wiki/Higher-order_function): 

```suggestion
    public static function byArrayKey(string $name): \Closure
    {
        return static fn (array $a, array $b): int => $a[$name] <=> $b[$name];
    }
```

The usage would be like: `usort($data, Sorter::byArrayKey('date'))`

lib/Tool/Sorter.php
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Order by key length, longer key have priority
+     * (%abcd prior %ab, so that %ab doesn't replace %ab in [%ab]cd)
+     * @param string|null $a
+     * @param string|null $b
+     */
+    public static function strlen(?string $a, ?string $b): int
+    {
+        return strlen($b) <=> strlen($a);
```
jdreesen: Passing `null` to `strlen` is deprecated in PHP 8.1
```suggestion
        return strlen($b ?? '') <=> strlen($a ?? '');
```

lib/Tool/Sorter.php
NiklasBr: I wanted to avoid higher-order functions because it would result in additional overhead. The PHP array sorter functions are really smart because they will not even attempt to invoke the closure if the input array is fewer than two items.

But I do agree that this results in neater code.

brusch:
I agree for functions which are used multiple times and can be shared, but quite a lot are very specific for just one case, there it just makes it more difficult to release in my opinion.
Example: 
https://github.com/pimcore/pimcore/pull/10792/files#diff-ea06636172bbd5182617cdd5e57da91c132fe16c02e481a32191c1462639a083

So if we limit it to just functions which are at least used twice, I'm fine with it 😊

brusch:

PR #10801, opened 2021-11-10
web2print processor: do not load print document from runtime cache

jheimbach:
With the move of web2print pdf generation to the messenger, the runtime cache saves the first version of documents but if the document changes, the processor still uses the first version.

### Steps to reproduce
 - create new web2print document
 - generate pdf with messenger
 - change some content of the document
 - regenertate pdf

### What should happen: 
we see the new changes in the generated pdf

### What happens: 
we see the old version as generated pdf, and the new changes in the document will be overwritten (because the pdf processor saves the last generated time and overwrites the changes with the cached version, see [here](https://github.com/pimcore/pimcore/blob/10.x/lib/Web2Print/Processor.php#L130))

Maybe that is a bigger issue with the messenger long running process and the runtime cache and we have to fix this in some other place?

### DIFF
--- a/lib/Web2Print/Processor.php
+++ b/lib/Web2Print/Processor.php
@@ -183,7 +183,7 @@ abstract class Processor
      */
     protected function getPrintDocument($documentId)
     {
-        $document = Document\PrintAbstract::getById($documentId);
+        $document = Document\PrintAbstract::getById($documentId, true);
         if (empty($document)) {
             throw new \Exception('PrintDocument with ' . $documentId . ' not found.');
         }


brusch:
Good catch - thanks a lot! 👍

brusch:
This could be an issue in other places as well, I think it makes sense to flush the runtime cache when a worker receives a message. I'll provide a PR for that.

brusch:
Should be obsolete by #10802 which implements a more generic approach. 
@JHeimbach can you please confirm #10802 fixes your issue as well? 
Thanks again!

jheimbach:
@brusch yes, your PR fixes my issue! Thanks

brusch:
@JHeimbach thanks for your feedback! 😊

PR #10803, opened 2021-11-10
[Doc] Fix config/pimcore/classes path

blankse:
Path must relative and not absolute

### DIFF
--- a/doc/Development_Documentation/21_Deployment/05_Deployment_Tools.md
+++ b/doc/Development_Documentation/21_Deployment/05_Deployment_Tools.md
@@ -1,47 +1,47 @@
 # Deployment Tools
 
-Following tools are provided by Pimcore to support deployment processes. 
+Following tools are provided by Pimcore to support deployment processes.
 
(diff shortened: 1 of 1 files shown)


blankse:
@brusch I set `PIMCORE_CLASS_DEFINITION_WRITABLE=true` in the .env file but `getenv('PIMCORE_CLASS_DEFINITION_WRITABLE')` in the isWritable methods is false. Seems there is a bug.

blankse:
I create another PR for this: https://github.com/pimcore/pimcore/pull/10804

blankse:
@brusch Next problem: The `pimcore:deployment:classes-rebuild` command doesn't work if `PIMCORE_CLASS_DEFINITION_WRITABLE=true` is not set. So the live system can't generate the database changes and the DataObject folder.

The DataObject folder is also in config/pimcore/classes. So the autoload doesen't work anymore.
Has anyone used this feature before? :D

brusch:
Thanks!

PR #10804, opened 2021-11-10
Fix PIMCORE_CLASS_DEFINITION_WRITABLE env

blankse:
`PIMCORE_CLASS_DEFINITION_WRITABLE=true` can't set via .env file


blankse:
I want move only the definition files to config/pimcore/classes. The class and listing files should be in var/classes/DataObject, because the config folder should be readonly. This is not possible. Because the `PIMCORE_CLASS_DIRECTORY` env variable is used for both. Should we create a new env variable only for the definition files?

models/DataObject/Objectbrick/Definition.php
```diff
@@ -739,7 +736,7 @@ protected function doEnrichFieldDefinition($fieldDefinition, $context = [])
      */
     public function isWritable(): bool
     {
-        if (getenv('PIMCORE_CLASS_DEFINITION_WRITABLE')) {
+        if (($_SERVER['PIMCORE_CLASS_DEFINITION_WRITABLE'] ?? 'false') === 'true') {
```
brusch: Should we just do a `(bool) ($_SERVER['PIMCORE_DEV_MODE'] ?? false)` here, as we also do in other places? 
See e.g. https://github.com/pimcore/pimcore/blob/a35a1544b969e049489cb29054673542b4c70025/lib/Pimcore.php#L62, also Symfony is doing that for boolean switches: 
https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/5.2/public/index.php#L18

brusch (changes requested):

models/DataObject/Objectbrick/Definition.php
blankse: Can change this. But than i would change the doc to PIMCORE_CLASS_DEFINITION_WRITABLE=1/0 because true/false is both true.

brusch:
@blankse why do you want them in different places? Currently that's not possible, because we think it should be either or and always deployed together.

models/DataObject/Objectbrick/Definition.php
brusch: Yes, I think that would make sense, existing `true` will continue to work as well.

blankse:
@brusch

1. The class and listing files are not in the versionierung (.gitignore), so the server must in the deploy process write them (I think the server should write only in var and public/var). The classes-rebuild command always write these files. Only the definition files are untouched.
2. The composer autoload is set in var/classes/DataObject (so this is missing in the doc than)

brusch:
I prefer to have the classes checked in as well - but may depend on the use-case - not sure if we really should create another config for that 🤔

blankse:
@brusch But then the class rebuild command may not write these files, otherwise the db cannot be updated.

I would prefer this change. I don't want check in things, which I can generate in the build process.

brusch:
@blankse That's true ... hmm, ok, but we should put that in another PR as it is not really a bug fix, so target branch would be `10.x`.

blankse:
@brusch Ok, I split it in 2 PR

brusch:
Thanks a lot 👍

PR #10812, opened 2021-11-11
Introduce PIMCORE_CLASS_DEFINITION_DIRECTORY env variable

blankse:
Dependents on https://github.com/pimcore/pimcore/pull/10804
Will rebase after merge this in 10.x branch


brusch:
@blankse seems that there are some unrelated changes in this PR as well, could you please recheck & resolve the conflict? 
Thanks!

blankse:
@brusch I corrected it.

models/DataObject/Traits/LocateFileTrait.php
```diff
@@ -20,16 +20,19 @@
...
-        } else {
-            return sprintf('%s/' . $pathTemplate,
-                PIMCORE_CLASS_DIRECTORY, $key);
         }
+
+        return sprintf('%s/' . $pathTemplate, PIMCORE_CLASS_DEFINITION_DIRECTORY, $key);
+    }
+
+    protected function locateFile(string $key, string $pathTemplate): string
+    {
```
brusch, line 35: I'd keep the custom file check for `locateFile()` as well, just in case someone has checked in the generated classes for deployment.

brusch (changes requested):

models/DataObject/Traits/LocateFileTrait.php
blankse: Why should Pimcore look in PIMCORE_CUSTOM_CONFIGURATION_DIRECTORY ?
This will not work for class-build command (with -c param). The class definition listing only check PIMCORE_CLASS_DEFINITION_DIRECTORY/PIMCORE_CLASS_DIRECTORY. So I think we can remove it also from locateDefinitionFile()

I think we must also change the env. Also before this change, right? The is_file is a additional IO load.

brusch: Because you can put your definitions and generated classes there, see: 
https://github.com/pimcore/pimcore/blob/10.x/doc/Development_Documentation/21_Deployment/05_Deployment_Tools.md#pimcore-class-definitions  

> The PHP configuration files and PHP classes will be written to the `var/classes` directory by default. However, you can create a copy
at `config/pimcore/classes` which is then read-only.  

So this needs to be definitely checked.

blankse: Ah ok by manuell copy. Okay I will keep it for both :)

brusch:
Looks good to me - thanks a lot! 👍😊

PR #10815, opened 2021-11-11
Admin Event POST_GET_RUNTIME_PERSPECTIVE

blankse:
Since the perspectives and custom views are in symfony yaml settings it is not possible to set dynamic root nodes.

So I think it would be good to integrate an event so we can change the static root nodes afterwards via event listener.


brusch:
Sounds reasonable to me. But since this is an admin-only functionality, I'd add the event to the existing `Pimcore\Event\AdminEvents`.

blankse:
I will move it to AdminEvents

lib/Event/AdminEvents.php
```diff
@@ -524,4 +524,16 @@ final class AdminEvents
...
+     * This event is fired after Pimcore generates the runtime Perspective
+     *
+     * Arguments:
+     *  - result | The result array
+     *
+     * @Event("Symfony\Component\EventDispatcher\GenericEvent")
+     *
+     * @var string
+     */
+    public const POST_GET_RUNTIME_PERSPECTIVE = 'pimcore.perspective.postGetRuntimePerspective';
```
brusch:
```suggestion
    public const PERSPECTIVE_POST_GET_RUNTIME = 'pimcore.admin.perspective.postGetRuntimePerspective';
```

lib/Perspective/Config.php
```diff
@@ -254,7 +256,12 @@ public static function getRuntimePerspective(\Pimcore\Model\User $currentUser =
 
         $result['elementTree'] = self::getRuntimeElementTreeConfig($currentConfigName);
 
-        return $result;
+        $event = new GenericEvent(null, [
+            'result' => $result,
+        ]);
+        \Pimcore::getEventDispatcher()->dispatch($event, AdminEvents::POST_GET_RUNTIME_PERSPECTIVE);
```
brusch:
```suggestion
        \Pimcore::getEventDispatcher()->dispatch($event, AdminEvents::PERSPECTIVE_POST_GET_RUNTIME);
```

brusch (changes requested):
I'd change the name of the constant and the value, to be consistent with the others in `AdminEvents` (they all start with the main purpose and have a `.admin.` in the value.

brusch:
Thank you very much! 👍

PR #10822, opened 2021-11-12
GlossaryTokenParser: Trigger deprecation only when used

blankse:
Deprecation warning is triggered in every twig template. Should only be triggerd if it is really used.


brusch:
Makes sense, thanks a lot!

PR #10823, opened 2021-11-12
[Docs] Mitigate https://httpoxy.org/ vulnerabilities in `nginx.conf` for production

jdreesen:
See: https://www.nginx.com/blog/mitigating-the-httpoxy-vulnerability-with-nginx/#Defeating-the-Attack-using-NGINX-and-NGINX-Plus


brusch:
Thanks!

PR #10527, opened 2021-10-07
ChildsCompatibilityTrait: Trigger deprecation

blankse:
Deprecated since 4.4. Should be removed in Pimcore 11.

PR #10824, opened 2021-11-12
Fix: Deprecations with childs properties

blankse:
Every page throw many deprecation warnings because class definitions still have childs properties.
See https://github.com/pimcore/pimcore/pull/10527

- [x] Migration (Depends on https://github.com/pimcore/pimcore/pull/10852)


bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -632,7 +632,7 @@ public function addFolderAction(Request $request)
      */
     public function deleteAction(Request $request)
     {
-        if ($request->get('type') == 'childs') {
+        if ($request->get('type') === 'children') {
```
NiklasBr: I like this! 👍 
But it looks like it might be a minor BC break, right? What if a custom bundle/integration calls `/admin/asset/delete` with `type=childs`. It's mostly the usage of the getters/setters inside Pimcore I was commenting about.

blankse: Yes I was searching for all childs typing. I will revert this.

models/DataObject/ClassDefinition/Layout.php
```diff
@@ -80,7 +80,7 @@ class Layout
     /**
      * @var array
      */
-    public $childs = [];
+    public $children = [];
```
jdreesen, line 107: Is this a BC break? This class or property doesn't seem to be `@internal`.

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
jdreesen: Maybe both can be supported, while deprecating the `childs` variant, and removing it in `v11`?

blankse: Yes I add both options and trigger an deprecation warning when childs is used

models/DataObject/ClassDefinition/Layout.php
blankse: I think these attributes should be internal. It should be used the getter and setter. There are only public to send it to JavaScript. Can we make it internal in 10.3 or is this also a BC for you?
Otherwise we must live with these deprecations till v11 or we can try to make both attributes. Maybe one tree with childs properties and one tree with children properties.

brusch: What if we make it a reference in the constructor? 
Similar to: https://github.com/pimcore/pimcore/blob/6.9/models/Document/Editable.php#L118

blankse: But this reference then no longer works in the json response to javascript right? I can imagine that this json will be quite large with complex layouts. But I can test it out.

brusch: The reference itself not, but is that necessary? The data is after JSON serialization in both properties, so that should be enough to work in the UI.

blankse: I changed it. But as i feared the response will be much bigger:
For example:
Before: 6,7 kB
After: 85,2 kB

But if that doesn't matter, we can leave it that way. An alternative would be a BC layer in PHP with __set and __get methods

blankse: I create the BC layer in PHP with __set/__get methods and create for JS a childs tree. So the response is much smaller
For example:
Before: 6,7 kB
After: 14,4 kB

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -559,6 +559,7 @@ static function (Task $task) {
             DataObject\Service::enrichLayoutDefinition($objectData['layout'], $object);
             $eventDispatcher->dispatch($event, AdminEvents::OBJECT_GET_PRE_SEND_DATA);
```
weisswurstkanone, line 560: Just wondering what happens if the event listener tries to find a 'child' node. If I understand it correctly $objectData['layout'] uses 'children' nodes from now on.

blankse: @weisswurstkanone Should be working because of the new __get/__set methods in models/Element/ChildsCompatibilityTrait.php
https://github.com/pimcore/pimcore/blob/840f4bdeefc6aed2f443158cff654a7027ade3e5/models/Element/ChildsCompatibilityTrait.php

fashxp:
also test with versions and recyclebin

mcop1:
Thank you 👍

dpfaffenbauer:
This actually is a BC break. Applications might expect and access the definitions called `childs`: https://github.com/coreshop/CoreShop/blob/94c133f65a704a8486e9bb1cd9687f852028e937/src/CoreShop/Component/Pimcore/DataObject/AbstractDefinitionUpdate.php#L65

dpfaffenbauer:
my workaround is this: https://github.com/coreshop/CoreShop/pull/1858

ugly though

mcop1:
@dpfaffenbauer please elaborate a bit. Could you give an example of what your code is doing exactly, respectively where you are experiencing problems?

PR #10825, opened 2021-11-12
Fix version display for archived version files

mhgtlc:
This commit fixes the case for older archived version files that have extension '.gz' (as on pimcore6), at the moment pimcore does not look for archived files


brusch:
Thanks a lot for your PR. However, `gzip` support has been removed in Pimcore X, see also upgrade notes: 
https://pimcore.com/docs/pimcore/current/Development_Documentation/Installation_and_Upgrade/Updating_Pimcore/V6_to_V10.html#page_Decompress-your-zipped-version-data

PR #10826, opened 2021-11-12
Fix unserialize for versions generated in pimcore6

mhgtlc:
Fix unserialize for versions generated in pimcore6: Cannot assign null to property Pimcore\Model\DataObject\Localizedfield:: of type array
Screenshot 2021-11-12 at 15 39 50
y


brusch:
LGTM, thanks a lot 👍

PR #10829, opened 2021-11-12
Data import improve logs

shamoh19:
Adding tooltip about the Datarow values.

Instead of clicking the open to see the row values, users can just hover to the messages to get information

Related to this [improvement](https://github.com/pimcore/data-importer/issues/111)

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/log/admin.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/log/admin.js
@@ -193,6 +193,7 @@ pimcore.log.admin = Class.create({
                     flex: 220,
                     sortable: true,
                     renderer: function (s) {
+                        meta.tdAttr='data-qtip="'+s+'"';
                         return Ext.util.Format.htmlEncode(s);
                     }
                 },{


mcop1:
@shamoh19 I tried to apply the patch, but after that the log tab is broken. You referred to another pr, but it was merged already. Did I miss something?

shamoh19:
@mcop1 only these two PR's. But this is only adding the tooltip, I am not sure whether this is really breaking the log tab

bundles/AdminBundle/Resources/public/js/pimcore/log/admin.js
```diff
@@ -193,6 +193,7 @@ pimcore.log.admin = Class.create({
                     flex: 220,
                     sortable: true,
                     renderer: function (s) {
+                        meta.tdAttr='data-qtip="'+s+'"';
```
jdreesen, line 196: What's `meta`? Seems it isn't defined anywhere :confused:

shamoh19: It's part of renderer function. It's been almost two months, I made this PR. It was working fine during that time. I have to check again.

May be we need to alter the function as below. 

renderer: function (s, metadata, record) {
                        meta.tdAttr='data-qtip="'+s+'"';
                        return Ext.util.Format.htmlEncode(s);
                    }

jdreesen: yeah, seems like the parameter is missing

brusch: @shamoh19 🏓 can you please recheck? Thanks!

brusch:

PR #10834, opened 2021-11-15
[Elements] Show in tree does not work when using HTML special characters in element key

BlackbitDevs:
Steps to reproduce:
1. Create data object named `Sneaker "test"` in a folder
2. Copy and paste this object 30 times until yet get a pagination in the data object tree
3. Open the object on page 2 (should be named `Sneaker "test"_copy_9`
4. Click `Show in tree` button

Object does not get found.

The reason is that in https://github.com/pimcore/pimcore/blob/bdba843682564b55292ae2c2ae724186d82db0d8/bundles/AdminBundle/Resources/public/js/pimcore/treenodelocator.js#L475-L481 the `text` of the tree item gets used. This gets filled in https://github.com/pimcore/pimcore/blob/bdba843682564b55292ae2c2ae724186d82db0d8/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L267

This `htmlspecialchars()` causes that in above JS function `getDirectionForElementsSortedByKey`
`elementKey.toUpperCase() = 'SNEAKER "TEST"_COPY_9"'`
`firstElementChild.data.text.toUpperCase() = 'SNEAKER &quot;TEST&quot;_COPY_9"'`

Because of this the comparison goes wrong and the pagination returns `-1` and thus does not go to page 2.

With this PR the real element key gets used for comparison.

Also resolves #6526

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/AssetController.php
@@ -611,7 +611,7 @@ class AssetController extends ElementControllerBase implements EventedController
 
         $tmpAsset = [
             'id' => $asset->getId(),
-            'text' => $asset->getFilename(),
+            'text' => htmlspecialchars($asset->getFilename()),
             'type' => $asset->getType(),
             'path' => $asset->getRealFullPath(),
             'basePath' => $asset->getRealPath(),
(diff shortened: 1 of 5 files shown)


BlackbitDevs:
Sorry, wrong merge branch

PR #10836, opened 2021-11-15
FIX: Snippet publish fails with required input fields

### LINKED ISSUE #10584: [Bug]: Snippet publish fails with required input fields
### Expected behavior

If you have a snippet template with a required input field like

`{{ pimcore_input('test', {'required': true}) }}`

save & publish is successfull with filled out input.

### Actual behavior

An Exception is thrown:

_**Prevented publishing document - missing values for required editables**_

This error seems to occur only on the first "Save & Publish". If you have the input field not required and publish and then make the field required the problem doesn't appear.

### Steps to reproduce

1. Create a snippet template with a required input field
2. Create a new snippet document based on the template
3. Enter data in the required input field
4. Click "Save & Publish"

BabovicT:
Resolves #10584

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/SnippetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/SnippetController.php
@@ -158,6 +158,10 @@ class SnippetController extends DocumentControllerBase
             $snippet->setPublished(true);
         }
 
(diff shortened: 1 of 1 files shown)


brusch:
Thanks a lot! 👍

PR #10837, opened 2021-11-15
Show Public URL for documents in backend

### LINKED ISSUE #10832: [Improvement]: Show Public URL for documents in backend
### Improvement description

Similar as we already have it for assets.
image

Would be helpful in some situations to quickly retrieve the URL of a document (without having to open it in the frontend).

Should also consider pretty URLs & sites.

jelena-hrga:
Resolves #10832
Added public URL and pretty URL on document info popup

### DIFF
--- a/models/Document/PageSnippet.php
+++ b/models/Document/PageSnippet.php
@@ -598,7 +598,11 @@ abstract class PageSnippet extends Model\Document
         }
 
         $url = $scheme . $hostname;
-        $url .= $this->getPrettyUrl() ?: $this->getFullPath();
+        if($this instanceof Page && $this->getPrettyUrl()) {
+            $url .= $this->getPrettyUrl();
+        } else {
+            $url .= $this->getFullPath();
+        }
 
         $site = \Pimcore\Tool\Frontend::getSiteForDocument($this);
         if ($site instanceof Model\Site && $site->getMainDomain()) {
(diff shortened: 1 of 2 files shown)


brusch:
Thanks for your PR! 
I think we should have just one field für the public URL. If there's a pretty URL, this should be used, if not, the path of the document. I think we'd need to fix this here: https://github.com/pimcore/pimcore/blob/36a44b071f98d6e632312e6e797006ef4af20f94/models/Document/PageSnippet.php#L600

Can you please change your PR accordingly? 
Thanks a lot!

jelena-hrga:
@brusch Thank you for the feedback, I changed my PR accordingly.

models/Document/PageSnippet.php
```diff
@@ -597,7 +597,8 @@ public function getUrl($hostname = null, $scheme = null)
             }
         }
 
-        $url = $scheme . $hostname . $this->getFullPath();
+        $url = $scheme . $hostname;
+        $url .= $this->getPrettyUrl() ?: $this->getFullPath();
```
brusch:
```suggestion
        if($this instanceof Page && $this->getPrettyUrl()) {
            $url .= $this->getPrettyUrl();
        } else {
            $url .= $this->getFullPath();
        }
```

brusch (changes requested):

models/Document/PageSnippet.php
brusch: We need to check if `$this` is a `Page` because `Snippet` doesn't have the method `getPrettyUrl()`.

brusch:
Thank you very much!

PR #10842, opened 2021-11-16
Elements expand icon issue with user permissions

dturopoli:
If we define objects like this 
Screenshot from 2021-11-16 14-00-10

and then define workspaces for user like this
Screenshot from 2021-11-16 14-00-55

When that user opens object tree he will see expand icon (+ plus icon) even though he doesn't have permission to see neither child objects, after click on plus icon the icon will disappear.

Screenshot from 2021-11-16 14-03-42

Screenshot from 2021-11-16 14-03-50

Same thing happens to documents and assets too.

Issue is caused because hasChildren method doesn't check user permissions. So fix for this can be if we use different method but maybe checking for user permissions needs to be added to hasChildren method.


brusch:
Totally makes sense - thanks a lot!

PR #10849, opened 2021-11-17
Elasticsearch - Timeout bugfix

ctippler:
Bugfix timeout - took the wrong timeout because
php_sapi_name() == PHP_SAPI
is equal in frontend requests


bundles/CoreBundle/EventListener/Frontend/DocumentFallbackListener.php
```diff
@@ -120,6 +120,7 @@ public function onKernelRequest(RequestEvent $event)
                 if ($document->getProperty('language')) {
                     $request->setLocale($document->getProperty('language'));
                 }
+                $this->documentResolver->setDocument($event->getRequest(), $this->fallbackDocument);
```
brusch: @ckogler-elements this seems to be an unrelated change - can you fix that? Thanks!

brusch (changes requested):

brusch:
That went wrong it seems 😉 Nevermind, I've just pushed the changes straight to `10.2` as the target branch was also not right 😊

PR #10852, opened 2021-11-17
Add DefinitionWriteException and catch this in migrations

blankse:
If a class definition file is in the config/pimcore/classes folder it can't be saved. So the migrations failed. Should be catched and give the user a hint.

### DIFF
--- a/models/DataObject/ClassDefinition/Exception/WriteException.php
+++ b/models/DataObject/ClassDefinition/Exception/WriteException.php
@@ -1,0 +1,24 @@
+<?php
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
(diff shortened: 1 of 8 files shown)


models/DataObject/ClassDefinition/Exception/WriteException.php
```diff
@@ -0,0 +1,24 @@
...
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\DataObject\ClassDefinition\Exception;
+
+class WriteException extends \Exception
```
brusch: Since `WriteException` is also used by field-collections and object bricks, I'd move it to `Pimcore\Model\DataObject\Exception` and maybe rename it to `DefinitionWriteException`

brusch (changes requested):

models/DataObject/ClassDefinition/Exception/WriteException.php
blankse: I renamed and moved the exception

brusch:
Thanks a lot!

PR #10857, opened 2021-11-18
Reset prepration queue - 

ctippler:
Use the integer in the query because it is a int field - not a string...


dvesh3:
@ckogler-elements Please rebase on 10.2 branch as it is a bugfix. thanks!

brusch:
@ckogler-elements that went wrong it seems 😊

brusch:
🍒-picked to `10.2` - 53e0547916aad653d7c67d08d23a38a38ebe9e6e 
Thanks!

PR #10860, opened 2021-11-18
[Kernel] Remove ini_set() without effect of "PCRE recursion limit"

podarcis:
The current call to ini_set() has no effect, as:
* Typo in ini identifier "pcre.recursion-limit" (dash instead of underscore)
* Bundled library simple_dom_html has been removed


brusch:
Thank you very much! 👍

PR #10862, opened 2021-11-18
Update SelectionProviderTrait.php

AlexanderWeckmer:
The existing version worked with the Select field constant from the resolver. Therefore only the Option Provider worked for the Select field. With the change, the Multiselect also works.
The resolver resolves now 😊.
Lexi


brusch:
Makes sense, thanks a lot! 😊 Just one more thing, could you please rebase your PR onto `10.2` as it is a bug fix? Thanks!

brusch:
Rebasing unfortunately went wrong it seems (many unrelated commits). I've 🍒-picked b4040605bbbb0140bb97560ca4a6d1e1c7738e44 into `10.2` so fixed. Thanks a lot 😊

PR #10864, opened 2021-11-19
Delete local temp file

j-jonas:
I think $workingPath is always a local file, which is created at line 38 "$workingPath = File::getLocalTempFilePath($extension);"

  
  

Resolves #


dvesh3:
@j-jonas yes, it was missed during refactoring. thanks!

PR #10867, opened 2021-11-19
[DOCS] Prevent upgrading errors

Cruiser13:
Add conflict with monolog bundle to prevent upgrading errors because of incompability with monolog
Do not use bin/console from master as it's version 10.x and does not fit the 5.x environment

docs only


brusch:
Thank you very much!

brusch:
`5.8` is eol, but I think it makes sense to update at least the docs 😉

PR #10870, opened 2021-11-20
fix capital A of global template location of Areabricks

grizzlydotweb:
Arabricks default global location in Pimcore X is now `templates/areas` 
 with a lowercase 'a'


brusch:
Thank you very much! 👍

PR #10871, opened 2021-11-21
Update select.js

AlexanderWeckmer:
image
image

The selected value of a ComboBox breaks when the key is Html. It seems to be caused by the sencha Library, reproducible in the sencha sandbox as well.
The changes resolve the problem with using a template for the options ("tpl") and a template for the selected value/key ("displayTpl"), and the problem causing dispalyfield is gone.

image


brusch:
@Lexipowder is that the case for multiselect as well? Could you please verify that? Thanks a lot! 😊

dvesh3:
> is that the case for multiselect as well? Could you please verify that? Thanks a lot! 😊

@brusch multiselect works fine.

@Lexipowder thanks!

Issue #10869, opened 2021-11-20
[Bug]: Cant create new VoucherTokenTypeSingle


jansarmir:
### Expected behavior

After saving **OnlineShopVoucherSeries** and clicking on **Voucher Detail** i should see Manager and Statistics

### Actual behavior

Clicking on **Voucher Detail** for new **OnlineShopVoucherSeries** without tokens, we got error

### Steps to reproduce

Create **OnlineShopVoucherSeries** with **VoucherTokenTypeSingle** and click on **Voucher Detail**

brusch:
![image](https://user-images.githubusercontent.com/142037/142847955-2449af3a-2325-4471-8be2-4344127c7a76.png)

brusch:
Fixed by #10876

PR #10876, opened 2021-11-22
Fix: Cant create new VoucherTokenTypeSingle

blankse:
Fix https://github.com/pimcore/pimcore/issues/10869

Regression from #9067


brusch:
Thank you!

PR #10877, opened 2021-11-22
[Workflow] Fix issue when creating element, ni ID given, unable to sa…

brusch:
Not entirely sure where this issue comes from, but actually calling `getMarking()` here: 
https://github.com/pimcore/pimcore/blob/412486842a2485016f198f616b317740cc0f1226/bundles/CoreBundle/EventListener/WorkflowManagementListener.php#L94 

triggers a workflow save, which fails because there is no element ID yet. 
Moving the event listener to post add solves the issue. 

How to reproduce: 
- Create new asset folder in our demo instance
- Error 😉

### DIFF
--- a/bundles/CoreBundle/EventListener/WorkflowManagementListener.php
+++ b/bundles/CoreBundle/EventListener/WorkflowManagementListener.php
@@ -59,9 +59,9 @@ class WorkflowManagementListener implements EventSubscriberInterface
     public static function getSubscribedEvents()
     {
         return [
-            DataObjectEvents::PRE_ADD => 'onElementPreAdd',
-            DocumentEvents::PRE_ADD => 'onElementPreAdd',
-            AssetEvents::PRE_ADD => 'onElementPreAdd',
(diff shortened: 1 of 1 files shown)


bundles/CoreBundle/EventListener/WorkflowManagementListener.php
```diff
@@ -59,9 +59,9 @@ public function __construct(
     public static function getSubscribedEvents()
     {
         return [
-            DataObjectEvents::PRE_ADD => 'onElementPreAdd',
-            DocumentEvents::PRE_ADD => 'onElementPreAdd',
-            AssetEvents::PRE_ADD => 'onElementPreAdd',
+            DataObjectEvents::POST_ADD => 'onElementPreAdd',
```
blankse: Should we also rename the method to `onElementPostAdd`? Is a internal class.

brusch: Yep, makes sense, done - thanks!

PR #10878, opened 2021-11-22
Fix: Cannot assign ElementDescriptor to property $object

blankse:
Followup to https://github.com/pimcore/pimcore/pull/10319


dvesh3:
@blankse thanks!

Issue #10868, opened 2021-11-19
[Bug]: VoucherService fails at isReserved


jansarmir:
### Expected behavior

Cannot add token to cart

### Actual behavior

after calling $cart->addVoucherToken with existing token, with no token in cart, operation fails

https://github.com/pimcore/pimcore/blob/91531281537b7cc8307682517fa9cc03aa617e4f/bundles/EcommerceFrameworkBundle/VoucherService/Token/Dao.php#L62

Fails when $reservation === NULL

### Steps to reproduce

Create token, add to cart, process fails with error

brusch:
fixed by #10881

PR #10881, opened 2021-11-22
Fix: VoucherService fails at isReserved

blankse:
Fix https://github.com/pimcore/pimcore/issues/10868

Regression from https://github.com/pimcore/pimcore/pull/10493


brusch:
Thank you very much!

PR #10883, opened 2021-11-23
Elasticsearch

ctippler:
Prevent warning when no hits are returned.


dvesh3:
@ckogler-elements thanks!

PR #7824, opened 2020-12-29
ExtJS 7 for X

weisswurstkanone:
resolves #6362

PR #10888, opened 2021-11-24
Restore ExtJS locale files

### LINKED ISSUE #10809: [Bug]: No switch of decimal separator in backend GUI (works under 6.9)
### Expected behavior

Decimal separator switches according to the selected language of the logged in user for all number inputs (number, quantity value, etc.).

### Actual behavior

Decimal separator is always the same: "."
The entry of a comma character is allowed but all data following it will be truncated. Only the first "." sign is treated as decimal separator

### Steps to reproduce

- Log in to the pimcore demo under https://demo.pimcore.com/admin
- Check if user language is set to "English"
- Select a car DataObject to edit
- Select the tab "Sale Information"
- Enter a price with cents in the attribute field "Price in EUR" by using a "." as decimal separator
- Switch user language to "German", save and reload backend
- Repeat the same as above and use a "," as decimal separator

blankse:
Fix https://github.com/pimcore/pimcore/issues/10809

Restores locale files from 6.9

Regression from #7824


brusch:
@blankse I think we could live without the `-debug` files, right? 
I would also check if the locale file existis before we include it, as it could lead to a `404` in certain cases. 
Can you please update the PR accordingly? Thanks in advance!

blankse:
@brusch I changed it.

mcop1 (approved):

PR #10892, opened 2021-11-24
Option to allow free input in select editable

dturopoli:
It would be usefull if we allow free input inside select editable by passing new option 'editable' when creating select editable


brusch:
Looks good to me, thanks a lot! 👍

PR #10894, opened 2021-11-24
fixes #10893: Tabpanel allows to set tabPosition to null, but throws …

### LINKED ISSUE #10893: [Bug]: Tabpanel allows to set tabPosition to null, but throws an exception when code is trying to get it
### Expected behavior

Pimcore UI allows to set Tabpanel's tabPosition to null. Also, this is the default value in old product definitions. setTabPosition allows to use null values, getTabPosition should also allow it. Or it should return `top` by default (as Sencha actually doesn't allow null as a value). The same with Localizedfields.

### Actual behavior

Getter throws an exception `The method "Pimcore\Model\DataObject\ClassDefinition\Layout\Tabpanel::getTabPosition()" returned "null", but expected type "string". Did you forget to initialize a property or to make the return type nullable using "?string"?`

### Steps to reproduce

For example, tabPosition getter is called during serialization. I was unable to find it anywhere else in . So if you want to reproduce, set circular references handler to solve the problem with serialization, then try to serialize a product.

CountZero1981:
…an exception when code is trying to get it.

Resolves #10893

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/localizedfields.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/localizedfields.js
@@ -137,6 +137,7 @@ pimcore.object.classes.data.localizedfields = Class.create(pimcore.object.classe
                             fieldLabel: t('tab_position'),
                             name: 'tabPosition',
                             value: this.datax.tabPosition,
+                            editable: false,
                             store: [['top', t('top')], ['left', t('left')], ['right', t('right')], ['bottom', t('bottom')]]
                         },
                         {
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/layout/tabpanel.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/layout/tabpanel.js
@@ -50,6 +50,7 @@ pimcore.object.classes.layout.tabpanel = Class.create(pimcore.object.classes.lay
                     fieldLabel: t('tab_position'),
                     name: 'tabPosition',
                     value: this.datax.tabPosition,
+                    editable: false,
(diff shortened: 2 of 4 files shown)


blankse:
1. Can you add `|null` to the phpdocs of setter/attribute?
https://github.com/pimcore/pimcore/blob/b60f2a712c7658750c1fad82e42d2bc7c85fb73c/models/DataObject/ClassDefinition/Data/Localizedfields.php#L115
https://github.com/pimcore/pimcore/blob/b60f2a712c7658750c1fad82e42d2bc7c85fb73c/models/DataObject/ClassDefinition/Data/Localizedfields.php#L1321
https://github.com/pimcore/pimcore/blob/b60f2a712c7658750c1fad82e42d2bc7c85fb73c/models/DataObject/ClassDefinition/Layout/Tabpanel.php#L35
https://github.com/pimcore/pimcore/blob/b60f2a712c7658750c1fad82e42d2bc7c85fb73c/models/DataObject/ClassDefinition/Layout/Tabpanel.php#L64
2. Can you add `editable: false` to the ui combos?
https://github.com/pimcore/pimcore/blob/b60f2a712c7658750c1fad82e42d2bc7c85fb73c/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/localizedfields.js#L135-L141
https://github.com/pimcore/pimcore/blob/b60f2a712c7658750c1fad82e42d2bc7c85fb73c/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/layout/tabpanel.js#L48-L54

CountZero1981:
@blankse will do at the weekend

blankse (approved):
Thanks for the changes :)

mcop1:
Thank you 👍

PR #10903, opened 2021-11-25
Remove unused variable

jdreesen:
`$templateReference` is immediately overridden in line 371 and thus always set.


dvesh3:
10.2 branch?

jdreesen:
sure

PR #10908, opened 2021-11-25
Ensure `$output` actually is `BufferedOutput` before calling `fetch()`

### LINKED ISSUE #10759: [Bug]: ExtensionManagerController NullOutput error
### Expected behavior

-

### Actual behavior

I've installed CoreShop in a Pimcore 10.2 project. Opening the Bundles & bricks page throws an error. 

```
Message: Attempted to call an undefined method named "fetch" of class "Symfony\Component\Console\Output\NullOutput".
Trace: 
in /app/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/ExtensionManager/ExtensionManagerController.php:495
#0 /app/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/ExtensionManager/ExtensionManagerController.php(412): Pimcore\Bundle\AdminBundle\Controller\ExtensionManager\ExtensionManagerController->getInstallerOutput(Object(CoreShop\Bundle\CoreBundle\CoreShopCoreBundle))
#1 /app/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/ExtensionManager/ExtensionManagerController.php(315): Pimcore\Bundle\AdminBundle\Controller\ExtensionManager\ExtensionManagerController->buildBundleInfo(Object(CoreShop\Bundle\CoreBundle\CoreShopCoreBundle))
#2 /app/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/ExtensionManager/ExtensionManagerController.php(78): Pimcore\Bundle\AdminBundle\Controller\ExtensionManager\ExtensionManagerController->getBundleList()
#3 /app/vendor/symfony/http-kernel/HttpKernel.php(156): Pimcore\Bundle\AdminBundle\Controller\ExtensionManager\ExtensionManagerController->getExtensionsAction()
#4 /app/vendor/symfony/http-kernel/HttpKernel.php(78): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#5 /app/vendor/symfony/http-kernel/Kernel.php(199): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#6 /app/public/index.php(35): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#7 {main}
```

$output is type hinted as BufferedOutput, but it looks like it can also return a NullOutput, which doesn't provide the fetch() method. I'm not sure if this is an install error and "should not happen", but it would be better to check the $output instance.

https://github.com/pimcore/pimcore/blob/4550a053fac47a1d2c0406bf15c9799ef078bbb7/bundles/AdminBundle/Controller/ExtensionManager/ExtensionManagerController.php#L491

### Steps to reproduce

In my case the error occured after installing CoreShop, but I have no idea if it's related to this.

m0nken:
Resolves #10759 / #10763

## Additional info  
`InstallerInterface::getOutput()` allows to return any `OutputInterface`, but only `BufferedOutput` provides a `fetch()` method.

### DIFF
--- a/bundles/AdminBundle/Controller/ExtensionManager/ExtensionManagerController.php
+++ b/bundles/AdminBundle/Controller/ExtensionManager/ExtensionManagerController.php
@@ -487,9 +487,8 @@ class ExtensionManagerController extends AdminController implements KernelContro
 
         $installer = $this->bundleManager->getInstaller($bundle);
         if (null !== $installer) {
-            /** @var \Symfony\Component\Console\Output\BufferedOutput $output */
             $output = $installer->getOutput();
-            if (!empty($output)) {
+            if ($output instanceof BufferedOutput) {
                 $converter = new AnsiToHtmlConverter(null);
 
                 $converted = Encoding::fixUTF8($output->fetch());
--- a/bundles/AdminBundle/Controller/ExtensionManager/ExtensionManagerController.php
+++ b/bundles/AdminBundle/Controller/ExtensionManager/ExtensionManagerController.php
@@ -29,6 +29,7 @@ use Pimcore\Logger;
 use Pimcore\Routing\RouteReferenceInterface;
 use Pimcore\Tool\AssetsInstaller;
 use SensioLabs\AnsiConverter\AnsiToHtmlConverter;
+use Symfony\Component\Console\Output\BufferedOutput;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Event\ControllerEvent;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;


dvesh3:
@m0nken please check tests are failing. thanks!

m0nken:
Fixed. Tests should run now.

dvesh3:
@m0nken thanks!

jdreesen:
The `OutputInterface` return type of `InstallerInterface::getOutput()` is problematic here, since most of the `OutputInterface` implementations don't have a way to fetch the output (they stream their output directly into a resource defined in their constructor), which means we can only really handle `BufferedOutput` or `NullOutput` here. So maybe everything except these two should be deprecated and the return type of `getOutput()` should be changed to `BufferedOutput|NullOutput` in Pimcore 11.

dvesh3:
> The `OutputInterface` return type of `InstallerInterface::getOutput()` is problematic here, since most of the `OutputInterface` implementations don't have a way to fetch the output (they stream their output directly into a resource defined in their constructor), which means we can only really handle `BufferedOutput` or `NullOutput` here. So maybe everything except these two should be deprecated and the return type of `getOutput()` should be changed to `BufferedOutput|NullOutput` in Pimcore 11.

@jdreesen yes, i think it makes sense to change the return type here.  thanks!

please see https://github.com/pimcore/pimcore/issues/10915

PR #10917, opened 2021-11-29
Small commit for generating "title" from "name" field

tanaydin:
That commit converts "testFieldName1" to "Test Field Name1" in data object fields.

## Additional info  
Create an object and a field under it, and write fields name in camelCase, it will add spaces in.


dvesh3:
@tanaydin thanks! 😃

PR #10923, opened 2021-11-29
Deprecate unused `Kernel::getRootDir()`

jdreesen:
`KernelInterface::getRootDir()` was deprecated in Symfony 4.2 and has been removed in Symfony 5.0.

This method is not used in Symfony and Pimcore anymore and could be deleted.

*Note:* should we deprecate this method instead, in case someone calls it somewhere?

### DIFF
--- a/bundles/InstallBundle/InstallerKernel.php
+++ b/bundles/InstallBundle/InstallerKernel.php
@@ -55,14 +55,6 @@ class InstallerKernel extends Kernel
     /**
      * {@inheritdoc}
      */
-    public function getRootDir()
-    {
-        return $this->projectRoot . '/var/installer';
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function getLogDir()
     {
         return $this->projectRoot . '/var/log';
@@ -73,7 +65,7 @@ class InstallerKernel extends Kernel
      */
     public function getCacheDir()
     {
-        return $this->getRootDir() . '/cache';
+        return $this->projectRoot . '/var/installer/cache';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


dpfaffenbauer:
> Note: should we deprecate this method instead, in case someone calls it somewhere?

I would, could be that a Bundle uses that.

dvesh3:
@jdreesen Dominik is right, it is even used in Pimcore [bundle](https://github.com/pimcore/bundle-generator/blob/1.x/src/Command/BaseGeneratorCommand.php#L71). Please mark it deprecated for now and we'll remove it in Pimcore 11 😉 thanks!

jdreesen:
Done.

lib/Kernel.php
```diff
@@ -67,12 +67,15 @@ abstract class Kernel extends SymfonyKernel
      * @var BundleCollection
      */
     private $bundleCollection;
-
-    /**
-     * {@inheritdoc}
-     */
+
```
blankse:
```suggestion

    /**
     * @deprecated
     */
```

jdreesen: thx

bundles/InstallBundle/InstallerKernel.php
```diff
@@ -52,14 +52,6 @@ public function getProjectDir()
         return $this->projectRoot;
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function getRootDir()
```
dvesh3, line 58: please revert this also and mark it deprecated.

jdreesen: `InstallerKernel` is `@internal`

dvesh3: Nevermind. Just noticed it it marked as `internal`

PR #10927, opened 2021-12-01
[Tests] PHP 8.1: Remove league/csv:9.x-dev requirement

blankse:
Followup to https://github.com/pimcore/pimcore/pull/9862


dvesh3:
@blankse good catch...thanks!

PR #10929, opened 2021-12-01
[Messenger] Create custom messenger bus for pimcore core

fashxp:
Bundles could create their own messenger buses (and they might be defined it as default automatically) thus Pimcore core and maintenance messages then use the bundles buses. 

To be more robust: Pimcore now defines its own bus and does not rely on the default bus anymore.

### DIFF
--- a/bundles/CoreBundle/Resources/config/maintenance.yaml
+++ b/bundles/CoreBundle/Resources/config/maintenance.yaml
@@ -8,7 +8,7 @@ services:
             - 'maintenance.pid'
             - '@logger'
             - '@Symfony\Component\Lock\LockFactory'
-            - '@Symfony\Component\Messenger\MessageBusInterface'
+            - '@messenger.bus.pimcore-core'
 
     Pimcore\Maintenance\Tasks\ScheduledTasksTask:
         arguments:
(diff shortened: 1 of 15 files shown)


bundles/CoreBundle/Resources/config/pimcore/default.yaml
```diff
@@ -43,6 +43,10 @@ framework:
             'Pimcore\Messenger\VersionDeleteMessage': pimcore_core
             'Pimcore\Messenger\OptimizeImageMessage': pimcore_core
             'Pimcore\Messenger\MaintenanceTaskMessage': pimcore_maintenance
+        default_bus: messenger.bus.portal-engine
```
blankse, line 46:
```suggestion
        default_bus: messenger.bus.pimcore-core
```

bundles/CoreBundle/Resources/config/maintenance.yaml
```diff
@@ -8,7 +8,7 @@ services:
             - 'maintenance.pid'
             - '@logger'
             - '@Symfony\Component\Lock\LockFactory'
-            - '@Symfony\Component\Messenger\MessageBusInterface'
+            - '@messenger.bus.pimcore-core'
```
dvesh3, line 11: How about we use a diff bus for maintenance tasks? Just to make it future-proof, so that we can add specific middlewares for maintenance bus.

dvesh3: Like for this use case https://github.com/pimcore/pimcore/pull/10725

PR #10931, opened 2021-12-01
[Data objects] Disable inheritance for single fields -> query table

BlackbitDevs:
Follow-up of #10719 

We forgot to request for field-specific `getNextParentForInheritance($field)` when the data for the query table gets written. Currently inherited data still gets written to the query tables (and can for example be seen in the grid). With this PR this is fixed. 

Arg, still does not work when saving the parent object. AFAIK this then gets written to the children with the `InheritanceHelper`. But this fetches all the descendant objects with a `WITH RECURSIVE` query. Does someone know how we could add a call to `getNextParentForInheritance($field)` there?


brusch:
@BlackbitNeueMedien Regarding your question ... hmmm ... I'm afraid that there's just no practical way of doing so for parent updates - because this wouldn't be acceptable in cases with massive children (there are cases with million children inheriting from a single parent (product inheritance structures). 
So I think for now we have to limit the `getNextParentForInheritance($field)` "functionality" to dynamic calls 🤔

brusch:
Will close this here for now ... however we need to continue the discussion, either here or in the original ticket #5416

BlackbitDevs:
@brusch What did you mean with
> So I think for now we have to limit the getNextParentForInheritance($field) "functionality" to dynamic calls

? Disabling inheritance for singel fields would really be useful...

BlackbitDevs:
@brusch Have we given up too early on this? 
When I look again on https://github.com/pimcore/pimcore/blob/828fdff008a5426daa77741cdbba3f5e1df3ecac/models/DataObject/Concrete/Dao/InheritanceHelper.php#L272-L299
the query with the `WITH RECURSIVE` actually only fetches the object ids of the children, it does not write anything. The writing is only done in https://github.com/pimcore/pimcore/blob/828fdff008a5426daa77741cdbba3f5e1df3ecac/models/DataObject/Concrete/Dao/InheritanceHelper.php#L298

And there we can add a check for `getNextParentForInheritance($field)` for the fields of `$originalEntry`, can't we?

BlackbitDevs:
Or @fashxp - what do you think? We currently have been asked to implement disabling inheritance for single fields again. Do you see any other problems except for updating the query table as described in https://github.com/pimcore/pimcore/pull/10931#issuecomment-1282069386 ?

PR #10934, opened 2021-12-01
[Improvement] Use Chrome Headless to generate Page Previews 

### LINKED ISSUE #10679: [Improvement]: Use Chome Headless to generate Page Previews
### Improvement description

instead of wkhtmltoimage. 

wkhtmltoimage however should be still supported (to be removed in PXI) 

Example command: 

```
chromium --headless --disable-gpu --screenshot=/path/to/file.png --window-size=1200,2000 "http://pimcore.com/foo/bar"
```

we should check for `chromium` first and then for `chrome`. 

Could be just integrated here: 
https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/lib/Image/HtmlToImage.php#L59

An also in the requirements check as an optional tool. 
I think we can just replace the check for wkhtmltoimage here.

dvesh3:
Resolves #10679


lib/Image/Chromium.php
```diff
@@ -0,0 +1,102 @@
...
+        if ($xvfb = Console::getExecutable('xvfb-run')) {
+            $command = [$xvfb, '--auto-servernum', '--server-args=-screen 0, 1280x1024x24',
+                self::getChromiumBinary()];
+        } else {
+            $command = [self::getChromiumBinary()];
+        }
+        $command = array_merge($command, $options);
+        Console::addLowProcessPriority($command);
+        $process = new Process($command);
+        //p_r($process->getCommandLine()); die;
```
dvesh3:
```suggestion
```

dvesh3 (commented):
Document page versions comparison is broken as it requires admin session to generate correct page previews and seems it is not possible to pass cookies in chromium cli directly. As, it is passed in htmltoimage adapter https://github.com/pimcore/pimcore/blob/aac548046cdbd0522f23816ed1daa3d3bab53b05/lib/Image/HtmlToImage.php#L77-L83

https://github.com/chrome-php/chrome might be an option to use for generating correct page previews.

brusch:
@dvesh3 yes, I'd go for https://github.com/chrome-php/chrome  as an optional dependency then. If `chrome-php/chrome` and Chrome is installed it's used, otherwise not.

brusch:
For me the following code works fine using https://github.com/chrome-php/chrome and the latest Pimcore docker images: 

```php

if(!$chromeBin = Console::getExecutable('chromium')) {
    $chromeBin = Console::getExecutable('chrome');
}

$browserFactory = new BrowserFactory($chromeBin);
$browser = $browserFactory->createBrowser([
    'noSandbox' => file_exists('/.dockerenv'),
    'startupTimeout' => 120,
    'windowSize' => [1280, 1024],
]);

try {
    $page = $browser->createPage();
    $page->navigate('https://demo.pimcore.fun/en/Magazine/Lifestyle')->waitForNavigation();

    $screenshotBase64 = $page->screenshot([
        'captureBeyondViewport' => true,
        'clip' => $page->getFullPageClip(),
    ])->getBase64(120000);

    return new Response(base64_decode($screenshotBase64), 200, [
        'Content-Type' => 'image/png',
    ]);

} finally {
    $browser->close();
}
```

So I'd go for this approach.

lib/Image/Chromium.php
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            $page->screenshot([
+                'captureBeyondViewport' => true,
+                'clip' => $page->getFullPageClip(),
+            ])->saveToFile($outputFile);
+        } finally {
+            $browser->close();
+
+            return false;
```
jdreesen: Is it correct to return always `false`?

dvesh3: Changed it to `true`

bundles/CoreBundle/EventListener/Frontend/ElementListener.php
```diff
@@ -143,7 +143,7 @@ protected function handleVersion(Request $request, Document $document)
     {
         if ($request->get('v')) {
             if ($version = Version::getById($request->get('v'))) {
-                if ($version->getPublic()) {
+                if ($version->getPublic() || $request->get('pimcore_version')) {
```
brusch: This would allow unauthenticated users to request any version, no matter if it was marked as public or not and could be seen as a data breach. 🤔

composer.json
```diff
@@ -174,13 +175,15 @@
     "phpunit/phpunit": "^9.3",
     "spiritix/php-chrome-html2pdf": "^1.6",
     "elasticsearch/elasticsearch": "^7.11",
-    "composer/composer": "*"
+    "composer/composer": "*",
+    "chrome-php/chrome": "^1.4.0"
```
brusch, line 177: I'd make `ext-sockets` and `chrome-php/chrome` as `suggest` only - as this could possibly break an existing installation if `ext-sockets` is not enabled.

brusch (changes requested):

brusch:
@dvesh3 there seems to be a bug in `chrome-php/chrome` regarding custom headers ... just implemented a workaround: 
https://github.com/pimcore/pimcore/pull/10934/files#diff-48c7411580a8efd2225afdf0851c253e959b509e1437cb27611c1418f49c8828R86

PR #10938, opened 2021-12-02
Support link generator for mockup objects

ctippler:
Currently it is not possible to use link generators for mockup objects. 
This leads to the problem, that in some places you can use {{ pimcore_url({'object': product}) }} because you have the orgininal object and in other cases {{ pimcore_url({'object': product}) }} won't work because you have a mockup object....

This PR adds support for Link generators with Mockup objects...

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Model/DefaultMockup.php
+++ b/bundles/EcommerceFrameworkBundle/Model/DefaultMockup.php
@@ -50,10 +50,7 @@ class DefaultMockup implements ProductInterface, LinkGeneratorAwareInterface
 
     public function getLinkGenerator(): ?DataObject\ClassDefinition\LinkGeneratorInterface {
         if($classId = $this->params['o_classId'] ?? null){
-            if(!array_key_exists($classId,static::$linkGenerators)){
-                static::$linkGenerators[$classId] = DataObject\ClassDefinition::getById($classId)->getLinkGenerator();
-            }
-            return static::$linkGenerators[$classId] ?? null;
+            return static::$linkGenerators[$classId] ??= DataObject\ClassDefinition::getById($classId)->getLinkGenerator();
         }
         return null;
     }
--- a/bundles/EcommerceFrameworkBundle/Model/DefaultMockup.php
+++ b/bundles/EcommerceFrameworkBundle/Model/DefaultMockup.php
@@ -16,8 +16,9 @@
 namespace Pimcore\Bundle\EcommerceFrameworkBundle\Model;
 
 use Pimcore\Logger;
+use Pimcore\Model\DataObject;
 
-class DefaultMockup implements ProductInterface
+class DefaultMockup implements ProductInterface, LinkGeneratorAwareInterface
 {
     /** @var int */
     protected $id;
@@ -27,6 +28,12 @@ class DefaultMockup implements ProductInterface
 
     /** @var array */
     protected $relations;
+
+    /**
+     * contains link generators by class type (just for caching)
+     * @var array
+     */
+    protected static array $linkGenerators = [];
 
     public function __construct($id, $params, $relations)
     {
@@ -39,6 +46,16 @@ class DefaultMockup implements ProductInterface
                 $this->relations[$relation['fieldname']][] = ['id' => $relation['dest'], 'type' => $relation['type']];
             }
         }
+    }
+
+    public function getLinkGenerator(): ?DataObject\ClassDefinition\LinkGeneratorInterface {
+        if($classId = $this->params['o_classId'] ?? null){
+            if(!array_key_exists($classId,static::$linkGenerators)){
+                static::$linkGenerators[$classId] = DataObject\ClassDefinition::getById($classId)->getLinkGenerator();
+            }
+            return static::$linkGenerators[$classId] ?? null;
+        }
+        return null;
     }
 
     /**
--- a/bundles/EcommerceFrameworkBundle/Model/LinkGeneratorAwareInterface.php
+++ b/bundles/EcommerceFrameworkBundle/Model/LinkGeneratorAwareInterface.php
@@ -4,7 +4,7 @@ namespace Pimcore\Bundle\EcommerceFrameworkBundle\Model;
 use Pimcore\Model\DataObject;
 
 /**
- * Interface IndexableInterface
+ * Interface LinkGeneratorAwareInterface
  */
 interface LinkGeneratorAwareInterface
 {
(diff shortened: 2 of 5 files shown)


bundles/EcommerceFrameworkBundle/Model/LinkGeneratorAwareInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Pimcore\Bundle\EcommerceFrameworkBundle\Model;
+use Pimcore\Model\DataObject;
+
+/**
+ * Interface IndexableInterface
```
blankse:
```suggestion
 * Interface LinkGeneratorAwareInterface
```

models/DataObject/ClassDefinition/LinkGeneratorInterface.php
```diff
@@ -15,15 +15,16 @@
...
 {
     /**
-     * @param Concrete $object
+     * @param Concrete|DefaultMockup $object
      * @param array $params
      *
      * @return string
      */
-    public function generate(Concrete $object, array $params = []): string;
+    public function generate(Concrete|DefaultMockup $object, array $params = []): string;
```
jdreesen: Should the `EcommerceFrameworkBundle` code be mixed into the general non-bundle code?

bundles/EcommerceFrameworkBundle/Model/DefaultMockup.php
```diff
@@ -41,6 +48,16 @@ public function __construct($id, $params, $relations)
...
 
+    public function getLinkGenerator(): ?DataObject\ClassDefinition\LinkGeneratorInterface {
+        if($classId = $this->params['o_classId'] ?? null){
+            if(!array_key_exists($classId,static::$linkGenerators)){
+                static::$linkGenerators[$classId] = DataObject\ClassDefinition::getById($classId)->getLinkGenerator();
+            }
+            return static::$linkGenerators[$classId] ?? null;
+        }
+        return null;
+    }
```
jdreesen:
```suggestion
   public function getLinkGenerator(): ?DataObject\ClassDefinition\LinkGeneratorInterface 
   {
        if ($classId = $this->params['o_classId'] ?? null) {
            return static::$linkGenerators[$classId] ??= DataObject\ClassDefinition::getById($classId)->getLinkGenerator();
        }

        return null;
    }
```

ctippler: thx. added the suggestions

models/DataObject/ClassDefinition/LinkGeneratorInterface.php
ctippler: Good point - maybe just allowing any object is sufficient because in the end it depends on the generator how to handle it....
Changed it to 
`
public function generate(object $object, array $params = []): string;
`

models/DataObject/ClassDefinition/LinkGeneratorInterface.php
```diff
@@ -15,15 +15,13 @@
...
 {
     /**
-     * @param Concrete $object
+     * @param object $object
      * @param array $params
      *
      * @return string
      */
-    public function generate(Concrete $object, array $params = []): string;
+    public function generate(object $object, array $params = []): string;
```
dvesh3: @ckogler-elements BC break? As existing implementations will break with Incompatible Declaration error.

ctippler: Yes, but i don't see a way around it...

blankse: Is it really a BC break? Concrete is a specific object. So it should still work, right?

dvesh3: Like i said, it will throw declaration error as `generate()` implementations needs to change the param type hint as well. https://php.watch/versions/8.0/lsp-errors

jdreesen: It is, because parameter types are contravariant. So the implementation is allowed to be *less specific*. But if you change the interface to be *less specific*, all existing implementations are *more specific* (except they left out the parameter type), which is not allowed and thus a BC break.

blankse: Yes you are right. I test it. I think with class and subclasses it should work. But object and class doesn't.
See: https://www.php.net/manual/en/language.oop5.variance.php

Edit: Okay in params type hints the behavior is reverse to the return type hints. Make sense.

blankse: @ckogler-elements This means you can revert this change. You can create LinkGenerator classes with object type hints anyway. This works.
You should add only a comment to change this interface in Pimcore 11.

ctippler: @blankse : thx - that works :-)
@dvesh3 : Changed it back and added a hint to the documentation

models/DataObject/ClassDefinition/LinkGeneratorInterface.php
```diff
@@ -20,6 +20,11 @@
 interface LinkGeneratorInterface
 {
     /**
+     *
+     * If you want to support mockups or arbitrary objects you can change the typehint to
+     * public function generate(object $object, array $params = []): string
+     *
+     * Concrete $object
```
blankse:
```suggestion
     * TODO: Change stypehint in Pimcore 11
     *
```

models/DataObject/ClassDefinition/LinkGeneratorInterface.php
```diff
@@ -20,6 +20,12 @@
 interface LinkGeneratorInterface
 {
     /**
+     *
+     * If you want to support mockups or arbitrary objects you can change the typehint to
+     * public function generate(object $object, array $params = []): string
```
jdreesen, line 25: Do we really need this comment? That's basic PHP contravariance behavior, so you could basically add this to every method...

models/DataObject/ClassDefinition/LinkGeneratorInterface.php
```diff
@@ -20,6 +20,12 @@
...
+     *
+     * TODO: Change stypehint in Pimcore 11
```
jdreesen:
```suggestion
     * TODO: Change type of `$object` parameter to `object` in Pimcore 11
```

dvesh3:
@ckogler-elements thanks!

PR #10943, opened 2021-12-02
Do not initialize db schema in constructor.

andreas-gruenwald:
When the ecommerce bundle is activated, on the /admin page additional 1.8000 - 4.000 db schema queries are executed:

image

In Microsoft Aure (Azure DB for Maria DB 10.4) this results in a TTFB between betwen 15 and 30 seconds.

Do not execute ``$this->db->getSchemaManager()->createSchema();`` in the constructor of the Ecommerce framework bundle installer. Instead, initialize it lazy when needed.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Tools/Installer.php
+++ b/bundles/EcommerceFrameworkBundle/Tools/Installer.php
@@ -413,15 +413,9 @@ class Installer extends AbstractInstaller
 
     /**
      * @return Schema
-     * @throws \Exception
      */
     protected function getSchema(): Schema
     {
-        if ($this->db instanceof Connection) {
-            $this->schema ??= $this->db->getSchemaManager()->createSchema();
-            return $this->schema;
-        } else {
-            throw new \Exception('Cannot create schema.');
-        }
+        return $this->schema ??= $this->db->getSchemaManager()->createSchema();
     }
 }
--- a/bundles/EcommerceFrameworkBundle/Tools/Installer.php
+++ b/bundles/EcommerceFrameworkBundle/Tools/Installer.php
@@ -415,7 +415,7 @@ class Installer extends AbstractInstaller
     /**
      * @return Schema
      */
-    public function getSchema() : Schema {
+    public function getSchema() : ?Schema {
         if ($this->db instanceof Connection && is_null($this->schema)) {
             $this->schema = $this->db->getSchemaManager()->createSchema();
         }
--- a/bundles/EcommerceFrameworkBundle/Tools/Installer.php
+++ b/bundles/EcommerceFrameworkBundle/Tools/Installer.php
@@ -17,6 +17,7 @@ namespace Pimcore\Bundle\EcommerceFrameworkBundle\Tools;
 
 use Doctrine\DBAL\Schema\Schema;
 use Pimcore\Db\ConnectionInterface;
+use Doctrine\DBAL\Connection;
 use Pimcore\Extension\Bundle\Installer\AbstractInstaller;
 use Pimcore\Extension\Bundle\Installer\Exception\InstallationException;
 use Pimcore\Model\DataObject\ClassDefinition;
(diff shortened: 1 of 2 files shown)


bundles/EcommerceFrameworkBundle/Tools/Installer.php
```diff
@@ -414,4 +411,14 @@ public function needsReloadAfterInstall()
     {
         return true;
     }
+
+    /**
+     * @return Schema
+     */
+    public function getSchema() : ?Schema {
```
jdreesen: If this returns `null` the calls that use this getter will fail.

bundles/EcommerceFrameworkBundle/Tools/Installer.php
```diff
@@ -414,4 +411,14 @@ public function needsReloadAfterInstall()
...
     }
+
+    /**
+     * @return Schema
+     */
+    public function getSchema() : Schema {
+        if (!$this->schema) {
+            $this->schema = $this->db->getSchemaManager()->createSchema();
+        }
+        return $this->schema;
```
jdreesen:
```suggestion
        return $this->schema ?? = $this->db->getSchemaManager()->createSchema();
```

bundles/EcommerceFrameworkBundle/Tools/Installer.php
```diff
@@ -414,4 +409,16 @@ public function needsReloadAfterInstall()
     {
         return true;
     }
+
+    /**
+     * @return Schema
+     */
+    public function getSchema(): Schema
```
blankse:
```suggestion
    protected function getSchema(): Schema
```

bundles/EcommerceFrameworkBundle/Tools/Installer.php
andreas-gruenwald: Resolved. Schema is required for installation anyway.

bundles/EcommerceFrameworkBundle/Tools/Installer.php
andreas-gruenwald: Changed syntax.

bundles/EcommerceFrameworkBundle/Tools/Installer.php
andreas-gruenwald: Makes sense. Changed.

lib/Db/ConnectionInterface.php
```diff
@@ -297,4 +297,9 @@ public function escapeLike(string $like): string;
      * @return \PDO
      */
     public function getWrappedConnection();
+    
+    /**
+     * @return \Doctrine\DBAL\Schema\AbstractSchemaManager
+     */
+    public function getSchemaManager();
```
fashxp: cannot add that since it would be BC break ...

bundles/EcommerceFrameworkBundle/Tools/Installer.php
```diff
@@ -414,4 +409,12 @@ public function needsReloadAfterInstall()
     {
         return true;
     }
+
+    /**
+     * @return Schema
+     */
+    protected function getSchema(): Schema
+    {
+        return $this->schema ??= $this->db->getSchemaManager()->createSchema();
```
fashxp, line 419: I think we need to do it the same way as we did it previously ... since we cannot add the `getSchemaManager` method to the interface. 

So do something like that: 
```
        if ($this->db instanceof Connection) {
            $this->schema ??= $this->db->getSchemaManager()->createSchema();
        } else {
            throw new \Exception('Cannot create schema');
        }
```

fashxp (changes requested):

lib/Db/ConnectionInterface.php
blankse: @fashxp And when we add this via `@method` PhpDoc and make this as comment for Pimcore 11? Then everybody has the notice to add this method. An exception is thrown either way.

andreas-gruenwald: I reverted the change as @fashxp  suggested. If you are not happy with it yet, then please just add your suggestion in another PR. By all means...this is just the Ecommerce installer and not a core component.

bundles/EcommerceFrameworkBundle/Tools/Installer.php
andreas-gruenwald: Implemented as suggested.

lib/Db/ConnectionInterface.php
andreas-gruenwald: @blankse just decided to add your suggestion. hope that I got it right...

fashxp (approved):

PR #10895, opened 2021-11-24
Update CustomReportController.php

milesimarco:
## Changes in this pull request  
Resolves #

Status: 500 | Internal Server Error
URL: /admin/reports/custom-report/get
Message: Call to a member function getObjectVars() on null

Trace: 
in /pimcore/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Reports/CustomReportController.php:164
#0 /pimcore/vendor/symfony/http-kernel/HttpKernel.php(156): Pimcore\Bundle\AdminBundle\Controller\Reports\CustomReportController->getAction()
#1 /pimcore/vendor/symfony/http-kernel/HttpKernel.php(78): Symfony\Component\HttpKernel\HttpKernel->handleRaw()
#2 /pimcore/vendor/symfony/http-kernel/Kernel.php(199): Symfony\Component\HttpKernel\HttpKernel->handle()
#3 /pimcore/public/index.php(35): Symfony\Component\HttpKernel\Kernel->handle()
#4 {main}

PR #10945, opened 2021-12-02
CustomReportController: Send HTTP Status 404 if report configs not found

blankse:
Alternative to https://github.com/pimcore/pimcore/pull/10895


dvesh3:
Thanks @blankse

PR #10946, opened 2021-12-02
Fix set stream for asset model

mhgtlc:
when an asset is renamed, or a new version is uploaded, display for old versions is broken:
Screenshot 2021-12-02 at 16 54 14

That happens because If stream is not rewindable it goes to:
`public function getTemporaryFile(bool $keep = false)
    {
        return self::getTemporaryFileFromStream($this->getStream(), $keep);
    }`

then, in `$this->getStream()` because stream is not rewindable, it is set to null and created from: `$this->stream = Storage::get('asset')->readStream($this->getRealFullPath());` 
This way stream is created from asset real full path stored in version (it is possible that it does not exist any more, or the file  is different from the file it was in version) and not from `$version->getBinaryFileStream()` as it should be

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1350,7 +1350,7 @@ class Asset extends Element\AbstractElement
             $isRewindable = @rewind($this->stream);
 
             if (!$isRewindable) {
-                $tempFile = $this->getTemporaryFile();
+                $tempFile = $this->getLocalFileFromStream($this->stream);
                 $dest = fopen($tempFile, 'rb', false, File::getContext());
                 $this->stream = $dest;
             }


mcop1:
Thanks!

BlackbitDevs:
This also happened with this:
```php
$asset = new Asset();
$asset->setFilename('abc');
$asset->setParentId(1);
$asset->setStream(fopen('http://example.org/image.jpg', 'rb'));
$asset->save();
```
because in `setStream()` `$this->getTemporaryFile()` got called. For remote files this called `self::getTemporaryFileFromStream($this->getStream(), $keep);` and this `$this->getStream()` set `$this->stream` to null in https://github.com/pimcore/pimcore/blob/a3b8da954582f15d499e02b0e35cebe71ea08f68/models/Asset.php#L1318-L1329 ending up with an empty `tmpfile()`.
Thx @mhgtlc for fixing!

PR #10947, opened 2021-12-02
[Maintenance] Add possibility to exclude maintenance messages

### LINKED ISSUE #10714: [Bug]: In 10.2 pimcore:maintenance no longer excludes excludedJobs

dvesh3:
Resolves #10714


models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -402,7 +403,8 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
 
                 $generated = true;
 
-                if ($optimizeContent) {
+                $config = PimcoreConfig::getSystemConfiguration('maintenance');
+                if ($optimizeContent && !($config['exclude-messages']['optimizeimage'] ?? false)) {
```
blankse:
```suggestion
                if ($optimizeContent && !in_array('optimizeimage', $config['exclude_messages'], true)) {
```
optimizeimage is a value and not a key

fashxp:
probably also already add this option to deactivate via configuration to all the other maintenance tasks too

brusch:
I'm not sure if this is the right approach ... the problem I see is, that with this approach the messages are not getting dispatched to the queue at all, so there's no possibility to process the tasks in any way. 
Although this could be also an approach, but then the config should be in a different context, e.g. in `pimcore.assets.image.thumbnails.opimize_auto_format` for thumbnails, because it's not directly related to the queue, but just disables the optimization in general. 

Maybe a custom handler is the better way to go in this case? 🤔

NiklasBr:
I like this design because currently our messages queue gets "polluted" so that it grows with about 6k rows per day because they are [not competing properly](https://github.com/pimcore/pimcore/issues/10714#issuecomment-953709325) and therefore preventing the queue from progressing.

brusch:
Ok, so image optimization is a kind of a special case, because the approach changed here. 
All other maintenance tasks should be still fine to exclude with the existing `-J` (`--excludedJobs=EXCLUDEDJOBS`) option, since the jobs are not getting added to the queue if provided. 

So imho the solution could be this config option, but just a a different tree structure.

brusch:
Closed in favor of #11121

PR #10950, opened 2021-12-02
fixes #10949: Objectbrick::getLocalizedTableName() returns wrong value

### LINKED ISSUE #10949: [Bug]: Objectbrick::getLocalizedTableName() returns wrong value
### Expected behavior

Objectbrick::getLocalizedTableName() should return something like `object_brick_Basedata_localized_24`

### Actual behavior

it returns `object_collection_Basedata_localized_24`

### Steps to reproduce

create an Objectbrick and call the function.
Objectbrick\Dao (models/DataObject/Objectbrick/Definition/Dao.php) is inherited from Fieldcollection\Dao, but overrides all parent functions excepting getLocalizedTableName.

probably Objectbrick\Dao should be inherited directly from AbstractDao, because by now all methods in Fieldcollection\Dao are overriden. also Fieldcollection and Objectbrick represent two separate field types.

CountZero1981:
Resolves #10949 Objectbrick::getLocalizedTableName() returns wrong value

### DIFF
--- a/models/DataObject/Objectbrick/Definition/Dao.php
+++ b/models/DataObject/Objectbrick/Definition/Dao.php
@@ -23,8 +23,14 @@ use Pimcore\Model\DataObject;
  *
  * @property \Pimcore\Model\DataObject\Objectbrick\Definition $model
  */
-class Dao extends Model\DataObject\Fieldcollection\Definition\Dao
+class Dao extends Model\Dao\AbstractDao
 {
+    use DataObject\ClassDefinition\Helper\Dao;
+
+    /**
+     * @var array|null
+     */
+    protected $tableDefinitions = null;
     /**
      * @param DataObject\ClassDefinition $class
      * @param bool $query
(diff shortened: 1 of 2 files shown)


models/DataObject/Objectbrick/Definition/Dao.php
```diff
@@ -40,6 +46,21 @@ public function getTableName(DataObject\ClassDefinition $class, $query = false)
...
+     * @param string $language
+     *
+     * @return string
+     */
+    public function getLocalizedTableName(DataObject\ClassDefinition $class, $query = false, $language = 'en')
+    {
+        if ($query) {
+            return 'object_brick_localized_query_' . $this->model->getKey() . '_' . $class->getId() . '_' . $language;
+        } else {
+            return 'object_brick_localized_' . $this->model->getKey() . '_' . $class->getId();
```
dvesh3, line 61:
```suggestion
            return 'object_brick_localized_' . $this->model->getKey() . '_' . $class->getId() . '_' . $language;
```

dvesh3: Ok. This is correct in the PR. Could you please fix here also?https://github.com/pimcore/pimcore/blob/3a986d76cd918072bef3c056e6eb55fe8b856b80/models/DataObject/Objectbrick/Definition.php#L132 thanks!

dvesh3: Pushed directly on 10.2 branch https://github.com/pimcore/pimcore/commit/8818767b22dd64be082831f95f1d6cba5528b6dd

PR #10962, opened 2021-12-04
No edit button for empty EditableDialogBoxConfiguration

### LINKED ISSUE #7671: Improvement: EditableDialogBox - Do not render button for empty configuration
# Description
In case the `EditableDialogBoxInterface` was implemented, the corresponding edit-button is still displayed if the configuration is empty, ultimately opening an empty modal.

An **example use case** is the restriction of the available configuration based on the user's permissions:

```php
/**
 * @inheritDoc
 */
public function getEditableDialogBoxConfiguration(Editable $area, ?Info $info): EditableDialogBoxConfiguration
{
    $dialogBoxConfiguration = new EditableDialogBoxConfiguration();

    $user = Tool\Authentication::authenticateSession();

    if($user && $user->isAllowed('change-background-color')) {
        $dialogBoxConfiguration->addItem([
            'type' => 'select',
            'name' => 'backgroundColor',
            'label' => 'Background Color'
        ]);
    }
    return $dialogBoxConfiguration;
}
```

LosHawlos:
This change will have the areablock check that the EditableDialogBoxConfiguration contains items prior to creating the edit button. This will avoid showing an empty dialog box window.

Resolves #7671

## Additional info  
Another way to solve this would be a change in the signature of getEditableDialogBoxConfiguration to allow the return of null, but that would cause a BC break.

### DIFF
--- a/models/Document/Editable/Areablock.php
+++ b/models/Document/Editable/Areablock.php
@@ -452,7 +452,11 @@ class Areablock extends Model\Document\Editable implements BlockInterface
         $brick = $areabrickManager->getBrick($this->indices[$this->current]['type']);
         if ($this->getEditmode() && $brick instanceof EditableDialogBoxInterface) {
             $dialogConfig = $brick->getEditableDialogBoxConfiguration($this, $info);
-            $dialogConfig->setId('dialogBox-' . $this->getName() . '-' . $this->indices[$this->current]['key']);
+            if( $dialogConfig->getItems()) {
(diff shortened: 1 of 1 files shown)


dvesh3:
@LosHawlos thanks!

dvesh3:
Cherry picked on 10.2 https://github.com/pimcore/pimcore/commit/a43122658ff1b2e11b7878989f953a9f12d59203

PR #10737, opened 2021-11-01
Update to PhpStan v1

blankse:
Update PhpStan to the new 1.0.0 version:
https://github.com/phpstan/phpstan/releases/tag/1.0.0

PR #10963, opened 2021-12-04
ClientFactory: Fix proxy uri

blankse:
?? operator has a less priority than . operator

Splitting from https://github.com/pimcore/pimcore/pull/10737 for branch 10.2


dvesh3:
@blankse thanks!

PR #10983, opened 2021-12-08
fix: check if fieldcollection has localized fields

zoidbergx:
Resolves a bugfix if having a FieldCollection within a LocalizedField with the same name than a LocalizedField in the object itself wich having also the FieldCollection included

## Additional info  
The FieldCollections itself doesn't support inerhitance out of the box and must still be done with custom code.

With this change this is still possible to support custom inheritance but also then it never should inherit a single LocalizedField (e.g. one language) instead it should be inherited only all languages of a FieldCollection - LocalizedField


dvesh3:
@zoidbergx Sorry, Pimcore 6.9 CE reached it's EOL so we are not merging any changes on 6.9 branch. you can check supported versions here https://github.com/pimcore/pimcore#supported-versions

If it is reproducible on Pimcore X, then please create a PR for 10.2 branch. thanks!

PR #10984, opened 2021-12-08
Update LogMailMaintenanceTask.php

frithjof:
This is currently not working because the array of receivers needs to be set as Address and assigned to addto for each entry
This Pull requests makes the mailing again usable
  
  

Resolves #


dvesh3:
pushed directly on 10.2 branch https://github.com/pimcore/pimcore/commit/eaef8fd537d32990f5f8f4ca7acade43cacd397d

dvesh3:
@frithjof thanks!

PR #10991, opened 2021-12-09
[Admin UI] DataObject Class - escape general settings input values

dvesh3:
Resolves #


bundles/AdminBundle/Resources/public/js/pimcore/object/classes/class.js
```diff
@@ -607,7 +607,12 @@ pimcore.object.classes.klass = Class.create({
                 for (var i = 0; i < items.length; i++) {
                     var item = items[i];
                     if (typeof item.getValue == "function") {
-                        this.data[item.name] = item.getValue();
+                        let value = item.getValue();
+                        if (typeof item.config.xtype != 'undefined' && item.config.xtype === 'textfield') {
```
dvesh3:
```suggestion
                        if (typeof item.config.xtype !== 'undefined' && item.config.xtype === 'textfield') {
```

PR #11009, opened 2021-12-13
ElementInterface: add setParent method as annotation

blankse:
data-importer use this method.
See https://github.com/pimcore/data-importer/pull/141


models/Element/ElementInterface.php
```diff
@@ -22,6 +22,9 @@
 use Pimcore\Model\User;
 use Pimcore\Model\Version;
 
+/**
+ * @method self setParent(?ElementInterface $parent)
```
blankse:
```suggestion
 * @method static setParent(?ElementInterface $parent)
```

brusch:
Thanks!

PR #11013, opened 2021-12-14
Do not load metadata converter unless required

NiklasBr:


NiklasBr:
Am I missing something that PHPStan sees? To me all uses of `$convert` are preceded by a check against `$raw`

https://github.com/pimcore/pimcore/blob/09ede7dc9a72a222506f417b1b0d4c3627345402/models/Asset.php#L1798-L1802

https://github.com/pimcore/pimcore/blob/09ede7dc9a72a222506f417b1b0d4c3627345402/models/Asset.php#L1814-L1818

https://github.com/pimcore/pimcore/blob/09ede7dc9a72a222506f417b1b0d4c3627345402/models/Asset.php#L1829-L1831

models/Asset.php
```diff
@@ -1772,19 +1772,21 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
         \Pimcore::getEventDispatcher()->dispatch($preEvent, AssetEvents::PRE_GET_METADATA);
         $this->metadata = $preEvent->getArgument('metadata');
 
-        $convert = function ($metaData) {
-            $loader = \Pimcore::getContainer()->get('pimcore.implementation_loader.asset.metadata.data');
-            $transformedData = $metaData['data'];
+        if (!$raw) {
+            $convert = function ($metaData) {
```
brusch, line 1776: It's because the declaration is done within this `if` ... but `$convert` is used further down outside this `if`. 
I also don't understand what this change should improve ... it's just the function declaration that's not done if `$raw = true`. 🤔

NiklasBr:
It's a performance improvement. When benchmarking assets getting the loader has a significant penalty, this change will make sure the loader is only being loaded if actually needed.

blankse:
Is it not possible to convert this function to a private method? Does this also improve performance?

brusch:
@blankse would make sense, yes, it would also solve the static analysis issue. Can someone create a PR for that? Thanks in advance!

PR #11016, opened 2021-12-14
Remove Document from runtime path cache after it was deleted

jdreesen:
In [`Document::getByPath()`](https://github.com/pimcore/pimcore/blob/10.x/models/Document.php#L217-L239) the Document gets cached (unlike the same method on `Assets` and `DataObjects`) but the cache is never reset. So when the was deleted, `getByPath()` still returns the now non-existing object.

```php
$doc = Document::getByPath('/some/path');
$doc->delete();

// still returns the doc!
$doc = Document::getByPath('/some/path');
```

This PR resets the path cache when the document gets deleted.


models/Document.php
```diff
@@ -827,6 +827,7 @@ public function delete()
 
         //clear document from registry
         \Pimcore\Cache\Runtime::set(self::getCacheKey($this->getId()), null);
+        \Pimcore\Cache\Runtime::set('document_path_' . md5($this->getRealFullPath()), null);
```
jdreesen: Should we create a `getPathCacheKey($path)` method?

dvesh3: yes, please refactor it to `getPathCacheKey`

jdreesen: done

models/Document.php
```diff
@@ -24,6 +24,7 @@
 use Pimcore\Model\Document\Hardlink\Wrapper\WrapperInterface;
 use Pimcore\Model\Document\Listing;
 use Pimcore\Model\Element\ElementInterface;
+use Pimcore\Model\Element\Service;
```
dvesh3: unused?

jdreesen: oops!

jdreesen:
Do we have to consider the `update()` and `save()` methods here, too? They also update the "id cache".

dvesh3:
> Do we have to consider the `update()` and `save()` methods here, too? They also update the "id cache".

Good catch! I would do it otherwise would not see latest changes on subsequent calls to `getByPath()`

dvesh3 (approved):

dvesh3:
@jdreesen thanks!

brusch:
@jdreesen this is probably also relevant for assets and data objects, right?

jdreesen:
No, because they don't cache `getByPath()` calls.

PR #11017, opened 2021-12-14
[Thumbnails] Support Asset Image for overlay transformations

### LINKED ISSUE #10649: [Improvement][Thumbnails] Add option to upload files for overlay directly
### Improvement description

or assign an asset... 

without the need for placing a file on the server.

dvesh3:
Resolves  #10649

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
@@ -380,7 +380,7 @@ pimcore.settings.thumbnail.items = {
                 }.bind(el),
 
                 onNodeOver: function (target, dd, e, data) {
-                    if (data.records.length === 1 && data.records[0].data.elementType === "asset") {
+                    if (data.records.length === 1 && data.records[0].data.type === "image") {
                         return Ext.dd.DropZone.prototype.dropAllowed;
                     }
                 },
@@ -390,7 +390,7 @@ pimcore.settings.thumbnail.items = {
                         var record = data.records[0];
                         var data = record.data;
 
-                        if (data.elementType === "asset") {
+                        if (data.type === "image") {
                             this.setValue(data.path);
                             if (this.previousSibling()) {
                                 this.previousSibling().setValue(data.id);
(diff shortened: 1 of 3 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
```diff
@@ -369,6 +369,44 @@ pimcore.settings.thumbnail.items = {
...
+            // add dnd support
+            let dd = new Ext.dd.DropZone(el.getEl(), {
+                reference: this,
+                ddGroup: "element",
+                getTargetFromEvent: function (e) {
+                    return this.getEl();
+                }.bind(el),
+
+                onNodeOver: function (target, dd, e, data) {
+                    if (data.records.length === 1 && data.records[0].data.elementType === "asset") {
```
mcop1: Maybe we should use 
```suggestion
                    if (data.records.length === 1 && data.records[0].data.type === "image") {
```
here? Because otherwise it is possible to dnd e.g. xlsx files, which doesn´t make sense, I think.

bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
```diff
@@ -369,6 +369,44 @@ pimcore.settings.thumbnail.items = {
...
+                        return Ext.dd.DropZone.prototype.dropAllowed;
+                    }
+                },
+
+                onNodeDrop: function (target, dd, e, data) {
+                    if (pimcore.helpers.dragAndDropValidateSingleItem(data)) {
+                        var record = data.records[0];
+                        var data = record.data;
+
+                        if (data.elementType === "asset") {
```
mcop1: Same here, maybe change to 
```suggestion
                        if (data.type === "image") {
```
otherwise I get the icon that the drop is not allowed, but I am still able to drop the asset.

mcop1 (changes requested):

bundles/AdminBundle/Resources/public/js/pimcore/settings/thumbnail/item.js
dvesh3: Good catch! 👍

mcop1 (approved):

PR #11018, opened 2021-12-14
Don't return Document from runtime cache when `$force` parameter is true in `Document::getByPath()`

jdreesen:
When calling `Document::getByPath('/some/path/', true)`, I'd expect a fresh document loaded from the DB. But as the method has a cache, a cached instance is returned.

When the `$force` parameter is `true`, the cache is skipped and the document gets loaded from the DB.

### DIFF
--- a/models/Document.php
+++ b/models/Document.php
@@ -224,8 +224,6 @@ class Document extends Element\AbstractElement
             return \Pimcore\Cache\Runtime::get($cacheKey);
         }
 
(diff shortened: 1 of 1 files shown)


dvesh3:
10.2 branch?

jdreesen:
Sure, just forgot to select the right branch.

dvesh3:
@jdreesen thanks!

PR #11020, opened 2021-12-14
[Assets] Add dialog to overwrite assets on upload

### LINKED ISSUE #6893: Asset Upload | Replace existing asset instead of adding a duplicate
# Feature Request

## Problem Description
When uploading the same asset several times, Pimcore adds multiple versions of the same asset, by adding an incremental number at the end of the file name.

Example:
image

```
test.txt
test_1.txt
test_2.txt
```

are identical.

---

## Improvement Suggestion
Add a configuration option, so that when uploading assets with the same file name, the user is asked if the existing file should be replaced, instead of adding an additional file without prompting.

There are several challenges related to this issue, as mass uploads, etc. need to be considered as well. However, as this feature seems to be quite important for some of our customers, because they want to avoid redundantly saved data, I hope that other members of the community will vote for that issue as well, so that it might get realized some day.

BlackbitDevs:
In case of a file path collision when uploading an asset file currently the new file gets uploaded with `_1` suffix (or `_2`, `_3` etc.).
With this PR in such a case this dialog opens:
Bildschirmfoto 2021-12-14 um 16 05 46

The buttons explained:
- *Overwrite*: Use the uploaded file as data for the already existing asset
- *Keep both files*: Upload new file with suffix (current behaviour)
- *Overwrite all*: When uploading multiple files at once, all existing files will get overwritten. Files which did not exist before, are created as usual.
- *Close window*: Abort upload -> No files will be changed

Resolves #6893

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -394,6 +394,23 @@ class AssetController extends ElementControllerBase implements KernelControllerE
     }
 
     /**
+     * @Route("/exists", name="pimcore_admin_asset_exists", methods={"GET"})
+     *
+     * @param Request $request
+     *
+     * @return JsonResponse
+     *
+     * @throws \Exception
+     */
+    public function existsAction(Request $request)
+    {
+        $parentAsset = \Pimcore\Model\Asset::getById((int)$request->get('parentId'));
+        return new JsonResponse([
+            'exists' => Asset\Service::pathExists($parentAsset->getRealFullPath().'/'.$request->get('filename'))
+        ]);
+    }
+
+    /**
      * @param Request $request
      * @param Config $config
      *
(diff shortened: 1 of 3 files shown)


brusch:
Just a note: needs to be tested carefully with uploading multiple files at once ...

mcop1:
Hello @BlackbitNeueMedien ,
pr looks really good, thank you guys! Since we have an "Overwrite all" option, the only thing I am missing a bit is a similiar option for "keep both". 

@NiklasBr postet a screenshot here how that could look like:
https://github.com/pimcore/pimcore/issues/6893#issuecomment-668529774

I think it would be a good idea to remove the "overwrite all" button and add the "apply to all" checkbox to the asset dialog. What do you guys think?

mcop1:
@BlackbitNeueMedien thank you for your work. I will merge the pr now. 
Additionally I created an follow up improvement #11315 to add the mentioned checkbox in the future.

PR #11026, opened 2021-12-15
Fix has children detection

mhgtlc:
After this PR was merged: https://github.com/pimcore/pimcore/pull/10842/files some ajax calls for non admin users may take about 30 seconds (for example: `/admin/object/tree-get-childs-by-id`) and if it overflows the `max_execution_time` limit it is aborted, that happens because it counts all the objects and that is not efficient. 
Screenshot 2021-12-15 at 12 34 58

Also there is a discrepancy from `hasChildren` and `getChildAmount` methods, taking in account that `getChildAmount` method checks the user, but  `hasChildren` doesn't do this, there can be situations when `getChildAmount` returns 0, but `hasChildren` returns true, which is confusing.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -697,7 +697,7 @@ class AssetController extends ElementControllerBase implements KernelControllerE
             ],
         ];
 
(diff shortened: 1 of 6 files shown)


dvesh3:
@mhgtlc LGTM, thanks!

PR #11027, opened 2021-12-15
[Docs] Update 09_Adding_Document_Editables.md

DzemalBecirevic:
Documentation for event listener is using incorrect event for an editable. Should use "pimcore.bundle_manager.paths.editmode_js" to only be loaded in editmode, after all other editables.

Same change applies for CSS using "editmode_css"

  
  

Resolves #


brusch:
Thank you very much!

PR #11031, opened 2021-12-15
[Session] Get session from current request as session could not be set in RequestStack

dvesh3:
Follow up #10909

## Additional info  
Fixes warnings "There is currently no session available" in logs

### DIFF
--- a/bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
+++ b/bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
@@ -55,7 +55,7 @@ class AdminSessionHandler implements LoggerAwareInterface, AdminSessionHandlerIn
      */
     public function getSessionId()
     {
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
```diff
@@ -238,6 +242,26 @@ public function loadSession(): SessionInterface
...
+     *
+     * @return SessionInterface
+     */
+    private function getSession(): SessionInterface
+    {
+        return $this->requestStack->getCurrentRequest()->getSession();
+    }
+
+    /**
+     * Gets the Session.
```
dvesh3:
```suggestion
     * returns if the session is set
```

dvesh3:
closed in favor of https://github.com/pimcore/pimcore/pull/11041

PR #11033, opened 2021-12-15
Bugfix: add UTF-8 support to Input regex validation

r-vanstraaten:
Regex validation in dataobjects currently does not support UTF-8, making unicode regex validation impossible.
Added 'u' modifier to checkValidity method to resolve.

Steps to reproduce:

- Add regex validation to Dataobject field with Unicode (Example "a-zA-Z\x{00C0}\-\x{024F}0-9" to support diacritics)
- Attempting to save object should result in the following exception: Exception: ErrorException: Warning: preg_match(): Compilation failed: character code point value in \x{} or \o{} is too large

  
  

Resolves #

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Input.php
+++ b/models/DataObject/ClassDefinition/Data/Input.php
@@ -304,7 +304,7 @@ class Input extends Data implements
     public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
     {
         if (!$omitMandatoryCheck && $this->getRegex() && strlen($data) > 0) {
-            if (!preg_match('#' . $this->getRegex() . '#', $data)) {
+            if (!preg_match('#' . $this->getRegex() . '#u', $data)) {
                 throw new Model\Element\ValidationException('Value in field [ ' . $this->getName() . " ] doesn't match input validation '" . $this->getRegex() . "'");
             }
         }


dvesh3:
@r-vanstraaten thanks for your PR! since this is a bugfix, please rebase it on branch 10.2.

r-vanstraaten:
@dvesh3 I believe that should have fixed it, apologies for the incorrect branch initially!

dvesh3 (dismissed):
Please add unicode flag to input.js:
https://github.com/DivanteLtd/pimcore/blob/93d49b761457c8016047d6506dfa45e91974610a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/input.js#L99

https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/input.js#L116

 otherwise it will highlight the field with validation message even when the value is correct.

r-vanstraaten:
@dvesh3 
Are you sure? I've tested it locally myself and asked a colleague to check, neither of us are getting validation errors/highlights/flashes. Tested with save without publish and direct save & publish, I've been unable to reproduce any validation messages.
Tested in Firefox, Chrome and Brave browser.
Screenshots of a test field below:
image
image
image

dvesh3:
@r-vanstraaten thanks!

PR #11040, opened 2021-12-16
Don't execute "Git Mirror" workflow on forks

blankse:
I get error emails from my fork. See: https://github.com/blankse/pimcore/actions

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -9,6 +9,7 @@ on:
 jobs:
     to_github:
         runs-on: ubuntu-18.04
+        if: github.repository_owner == 'pimcore'
         steps:
             - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror.yml
```diff
@@ -9,9 +9,10 @@ on:
 jobs:
     to_github:
         runs-on: ubuntu-18.04
+        if: github.repository_owner == 'pimcore'
```
brusch:
```suggestion
        if: ${{ github.repository == 'pimcore/pimcore' }}
```
I think we can make this even more restrictive as we do here as well: https://github.com/pimcore/pimcore/blob/2320fe9185aa164296aa5e71815cb5feb1766fcd/.github/workflows/lts-repo-sync.yml#L11

brusch (changes requested):

brusch:
Thanks!

PR #11043, opened 2021-12-16
[Core] Check (and fix) Dao & Models for redundant data-setters

### LINKED ISSUE #7049: [Core] Check (and fix) Dao & Models for redundant data-setters
See here: https://github.com/pimcore/pimcore/pull/7001#pullrequestreview-481701211
There are probably some more such cases, as @jdreesen already mentioned, see https://github.com/pimcore/pimcore/blob/da18c52e7e9b902b01c13cde23568f2f63ab91ab/models/Document/PageSnippet.php#L645

mcop1:
fixes #7049

### DIFF
--- a/models/Element/Traits/VersionDaoTrait.php
+++ b/models/Element/Traits/VersionDaoTrait.php
@@ -60,10 +60,6 @@ trait VersionDaoTrait
             'ctype' => Element\Service::getElementType($this->model),
         ])->setOrderKey('id')->setOrder('ASC');
 
-        $versions = $list->load();
-
-        $this->model->setVersions($versions);
-
-        return $versions;
+        return $list->load();
     }
 }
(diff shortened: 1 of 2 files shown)


models/Element/Traits/VersionDaoTrait.php
```diff
@@ -60,10 +60,6 @@ public function getVersions()
             'ctype' => Element\Service::getElementType($this->model),
         ])->setOrderKey('id')->setOrder('ASC');
 
-        $versions = $list->load();
-
-        $this->model->setVersions($versions);
-
-        return $versions;
+        return $list->load();
```
NiklasBr, line 63:
```suggestion
        return $list->getVersions();
```

mcop1: load method has been called before, so I wouldn´t necessarily change it. The property is set here for example:
https://github.com/pimcore/pimcore/blob/10.x/models/Asset.php#L1492 
What do you think? :)

NiklasBr: Right you are :)

PR #11044, opened 2021-12-16
[DataObject] pre update validation exception event

kjkooistra-youwe:
Dispatch an event before the validation exception is thrown.
Allow adding custom validation exceptions and manipulation of the output.

Without output changes.
```php
$customValidationExceptions = [
    new ValidationException('Custom field validation | This value is too long. It should have 20 characters or less.'),
];
$validationExceptions = array_merge(
    $event->getArgument('validationExceptions'),
    $customValidationExceptions
);
$event->setArgument('validationExceptions', $validationExceptions);
```
Validation_before

With output changes.
```php
$event->setArgument('message', '');
$event->setArgument('separator', '<br>');
```
Validation_after

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -167,15 +167,23 @@ class Concrete extends DataObject implements LazyLoadedFieldsInterface
             }
         }
 
+        $preUpdateEvent = new DataObjectEvent($this, [
+            'validationExceptions' => $validationExceptions,
+            'message' => 'Validation failed: ',
+            'separator' => ' / ',
+        ]);
+        \Pimcore::getEventDispatcher()->dispatch($preUpdateEvent, DataObjectEvents::PRE_UPDATE_VALIDATION_EXCEPTION);
+        $validationExceptions = $preUpdateEvent->getArgument('validationExceptions');
+
         if ($validationExceptions) {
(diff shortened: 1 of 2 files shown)


kjkooistra-youwe:
@brusch @weisswurstkanone Hope you don't mind me tagging you. Any chance this PR can be merged?

weisswurstkanone:
Makes sense for me to add additional rules. Thx a lot!

models/DataObject/Concrete.php
```diff
@@ -167,15 +167,23 @@ protected function update($isUpdate = null, $params = [])
             }
         }
 
+        $preUpdateEvent = new DataObjectEvent($this, [
```
brusch, line 170: This could be inside the `if` condition, right?

brusch (changes requested):

models/DataObject/Concrete.php
kjkooistra-youwe: No, because then you cannot add validations if there are no exceptions from the default Pimcore validations.

brusch: Ah, sorry, missed the one crucial line, thanks for clarifying 😊

brusch:
Thank you!

PR #11046, opened 2021-12-16
[Image Thumbnails] Fix system image thumbnails generation when no thumbnails are configured

rquarta-extera:
Resolves this bug: when no thumbnails are configured in Pimcore, it's not possible to generate system thumbnails.

## To reproduce

- Install Pimcore Skeleton
- Leave image thumbnails configuration empty (don't configure any thumbnail)
- Open a terminal and prompt the command to generate system thumbnails
`bin/console thumbnails:image -s`
- No thumbnails are generated


mcop1:
Thank you

PR #11050, opened 2021-12-17
Pricing Rules - Saving pricing rule with date range condition leads to "setTime() on bool" error

### LINKED ISSUE #11028: [Bug]: Pricing Rules - Saving pricing rule with date range condition leads to "setTime() on bool" error
### Expected behavior

Saving pricing rules with date range condition should be possible.

### Actual behavior

On saving Pimcore shows an error dialogue with the following message:
image
No further stacktrace is given. 

Error is triggered in fromJSON() function in Pimcore\Bundle\EcommerceFrameworkBundle\PricingManager\Condition\DateRangeCondition.

`$starting = \DateTime::createFromFormat('Y-m-d\TH:i:s', $json->starting);`
`$starting->setTime(0, 0, 0);`
` $ending = \DateTime::createFromFormat('Y-m-d\TH:i:s', $json->ending);`
`$ending->setTime(23, 59, 59);`

Starting and ending keys in the decoded json array are not of expected format, therefore DateTime returns null.

### Steps to reproduce

1) Open Pimcore Demo
2) Open arbitrary pricing rule
3) Add a date range condition on tab "conditions" and fill in date values
4) Click "save"

mcop1:
Fixes #11028

### DIFF
--- a/bundles/EcommerceFrameworkBundle/PricingManager/Condition/DateRange.php
+++ b/bundles/EcommerceFrameworkBundle/PricingManager/Condition/DateRange.php
@@ -15,6 +15,7 @@
 
 namespace Pimcore\Bundle\EcommerceFrameworkBundle\PricingManager\Condition;
 
+use DateTimeZone;
 use Pimcore\Bundle\EcommerceFrameworkBundle\PricingManager\ConditionInterface;
 use Pimcore\Bundle\EcommerceFrameworkBundle\PricingManager\EnvironmentInterface;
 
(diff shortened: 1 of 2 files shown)


mcop1:
@asysrieke could you please try and check if it is working for you :)?

asysrieke:
@mcop1 
It does solve the saving problem.

But the function is not working properly.

If I save a date range from 2021/12/18 to 2021/12/18 as a condition in a pricing rule it is perfectly possible to add a voucher connected to that pricing rule today (2021/12/17), which should not be possible.

After some debugging I found that on saving the pricing rule the string is transported as UTC so when I type 2021/12/18 in the date condition field it transports the string as "2021-12-17T23:00:00.000Z"
image

In Condition\DateRange::fromJSON() it then creates the DateTime Object from that string(s) and sets the time to (0,0,0) resp. (23/59/59) so that $starting and $ending look like this:
image

As it later validates the current timestamp against those $starting and $ending objects in check() functin the outcome is positive on 2021/12/17 though the date range is 2021/12/18.

My Condition configuration looks like this:
image

mcop1:
@asysrieke you are right, I was able to reproduce your problem. I pushed a possible solution for this, would you mind to try it again, please?

PR #11054, opened 2021-12-17
[Cache] consider none-string values

jremmurd:
If the cache write fails and the value is not a string, the error log will result in an error itself during item size calculation.
This PR serializes values other than strings for size calculation and also considers multibyte characters in string length calculation.

### DIFF
--- a/lib/Cache/Core/CoreCacheHandler.php
+++ b/lib/Cache/Core/CoreCacheHandler.php
@@ -591,11 +591,25 @@ class CoreCacheHandler implements LoggerAwareInterface
         if ($result) {
             $this->logger->debug('Added entry {key} to cache', ['key' => $item->getKey()]);
         } else {
+            
+             $itemData = $item->get();
+            if (!is_string($itemData)) {
+                try {
+                    $itemData = serialize($itemData);
+                } catch (\Throwable) {
+                    $itemSizeText = "unknown";
+                }
+            }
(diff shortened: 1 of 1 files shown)


lib/Cache/Core/CoreCacheHandler.php
```diff
@@ -591,11 +591,24 @@ function ($currentValue) {
         if ($result) {
             $this->logger->debug('Added entry {key} to cache', ['key' => $item->getKey()]);
         } else {
+            
+            $itemData = $item->get();
+            if (is_string($itemData)) {
```
brusch, line 596:
```suggestion
            if (is_scalar($itemData)) {
```

lib/Cache/Core/CoreCacheHandler.php
```diff
@@ -591,11 +591,24 @@ function ($currentValue) {
...
+                $itemSizeText = formatBytes(mb_strlen($itemData));
```
brusch, line 597:
```suggestion
                $itemSizeText = formatBytes(mb_strlen((string) $itemData));
```

brusch (changes requested):
I think we don't need to serialize other scalar data types, just casting them to `string` should be suffcient, right?

brusch:
Closed in favor of #11108 - thanks a lot for reporting!

PR #11055, opened 2021-12-17
Class field rename save error

dturopoli:
When you create a class with a field named 'test' and then change field name to 'Test' and save this error appears

Screenshot from 2021-12-17 12-34-55

I couldn't reproduce this error on demo projects, but it happened on every project I'm working on (10.1.4 and 10.2.6 versions)  and on fresh docker installation of skeleton version


dvesh3:
@dturopoli thanks!

PR #11056, opened 2021-12-17
fix: add UTF-8 support to javascript regex validation

r-vanstraaten:
add 'u' modifier to javascript input.js regex checks, removing validation error highlights with usage of \p{L} and similar
relates to #11033 

Without this fix:
image

With fix:
image

(Apologies Dvesh, I didn't test this case enough in my previous PR, should have been included in the mentioned PR)

  
  

Resolves #


dvesh3:
@r-vanstraaten thanks! at least we got it right before the release 😉

dvesh3:
@r-vanstraaten please see https://github.com/pimcore/pimcore/pull/11586 we have to revert the unicode flag changes as it breaks the existing expressions with backslashes. Now you should be able to set specific flags in the classdefinition itself.

Issue #7290, opened 2020-10-12
Change JS core events for bundles to native JS events


brusch:
see https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#example

Event names should be consistent with the Symfony ones, so e.g. `pimcore.document.preSave` (see e.g. https://github.com/pimcore/pimcore/blob/be857eb3ccdaf8fea5ab6639b8a8ef9c5c5cc49e/lib/Event/DocumentEvents.php#L18 ) 

We should change: 
`pimcore.eventDispatcher` : https://github.com/pimcore/pimcore/blob/e8514a18bcaef81f9f57507ed415e3d700f1afa3/bundles/AdminBundle/Resources/public/js/pimcore/event-dispatcher.js

and

`pimcore.plugin.broker` : 
https://github.com/pimcore/pimcore/blob/e8514a18bcaef81f9f57507ed415e3d700f1afa3/bundles/AdminBundle/Resources/public/js/pimcore/plugin/broker.js

BC layer should be possible to implement in both classes, where the class register for the new event and dispatches then the old style events in the new listener: 

### Example dummy
```js
pimcore.plugin.broker = {
    initialize: function() {
        const eventMappings = {
            "preSaveDocument": "pimcore.document.preSave",
            // ...
        };

        for (let oldKey in eventMappings) {
            document.addEventListener(eventMappings[oldKey], function (e) {
                
                pimcore.plugin.broker.fireEvent(oldKey, '...');
            });
        }
    }
    
    // ...
};
``

brusch:
+ BC Layer

gjorgic:
Hi @brusch, please can you tell me what is the further intention behind this migration to native JS? Are you planning to change all current plugin logic including `Class.create`?

Can i simple make my own broker which will act as proxy between JS native events and my component?

brusch:
@gjorgic Sure: using native events makes it a lot easier to hook into the UI, because you actually don't need a bundle anymore or following a specific structure - you just do it the normal JS way - and of course the code is more readable and shorter 😉 

Broker: yes, this is what we actually did here for `10.5`: 
https://github.com/pimcore/pimcore/pull/12566/files#diff-068bb2dae95fa59414280126f267d3e7a2a212ff07f753f24273493f75338f8c

So for 10.5 you're fine, for 11 you need to change it or re-intruduce this BC layer in your custom code (not recommended 😉)

PR #11057, opened 2021-12-17
[Feature]: Add pimcoreEditableAdded javascript event

GALCF:

Adds a javascript event "pimcoreEditableAdded" after an editable was added.

## Additional info  

This feature adds a possibility to run js-brick-logic in editmode without needing to set needsReload to true in bricks.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/areablock.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/areablock.js
@@ -848,12 +848,15 @@ pimcore.document.editables.areablock = Class.create(pimcore.document.area_abstra
                 },
                 success: function (response) {
                     let res = Ext.decode(response.responseText);
+                    let element;
+
                     if(!this.elements.length) {
-                        Ext.get(this.id).setHtml(res['htmlCode']);
+                        element = Ext.get(this.id).setHtml(res['htmlCode']);
+                        element = element.down('.pimcore_block_entry');
                     } else if (this.elements[index-1]) {
-                        Ext.get(this.elements[index-1]).insertHtml('afterEnd', res['htmlCode'], true);
+                        element = Ext.get(this.elements[index-1]).insertHtml('afterEnd', res['htmlCode'], true);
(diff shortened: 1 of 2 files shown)


GALCF:
Example usage in frontend JS (jQuery example for brevity, can be used without):

```js
const initializeBrick = ($brick) => {
  // My logic, yay
  console.log($brick);
};

const brickClass = '.wysiwyg-brick';

// Runs on DOM ready
initializeBrick($(brickClass));

// Runs on editable added
$(document).on('pimcoreEditableAdded', (event) => {
  const $element = $(event.target).find(brickClass);
  initializeBrick($element);
});
```

brusch:
In general this makes sense, but since we have a custom event implementation until we've switched to #7290 we should use that one for now, see `pimcore.plugin.broker.fireEvent()`. We also have to make this consistent, not just for the use with areablocks, so I'd say the right place for dispatching the event would be in `pimcore.document.editables.manager.addByDefinition()`

Can you please update your PR accordingly? Thanks!

GALCF:
> In general this makes sense, but since we have a custom event implementation until we've switched to #7290 we should use that one for now, see `pimcore.plugin.broker.fireEvent()`. We also have to make this consistent, not just for the use with areablocks, so I'd say the right place for dispatching the event would be in `pimcore.document.editables.manager.addByDefinition()`
> 
> Can you please update your PR accordingly? Thanks!

The `pimcore.plugin.broker.fireEvent()` requires the programmer to register a pimcore plugin. The native event trigger could be used directly via `document.addEventListener('pimcoreEditableAdded', /* ... */)`

For example I would like to initialize my brick with the same code (e.g. Slider-Initialization) that my business logic uses, not via pimcore plugin, because the plugin would only run in editmode.

brusch:
@GALCF That's true, but until now it is they way how it works and I keep it consistent for now. 
But I'm also open to change all the existing events to `addEventListener()` incl. a BC layer, so if you're interested in contributing that, we'd be happy to review/merge a PR 😊

brusch:

PR #11058, opened 2021-12-17
Add missing type check to the Document path cache

jdreesen:
`Element::getById()` checks the type of the cached element before returning it.
`Document::getByPath()` should do this, too.


dvesh3:
@jdreesen thanks!

PR #11061, opened 2021-12-20
[DataObjects] PCX update consider empty tables 

### LINKED ISSUE #11059: [Bug]: PCX Update tables in DataObject migration
### Expected behavior

Tables in DataObjects get migrated to new format during Pimcore Update Migrations. 

Previously the array got serialized:
I.e. empty table was `s:0:"";` --> now it's a null value

### Actual behavior

After the update an error is shown:

`Pimcore\Model\DataObject\MyObject::setMyTable(): Argument #1 ($myValue) must be of type ?array, string given, called in /var/www/html/vendor/pimcore/pimcore/lib/Model/AbstractModel.php on line 206`

### Steps to reproduce

Create a class with a table field in Pimcore < v10, perform Update to Pimcore >10, load the object.

## Suggested solutions:

1. Writing migration 
2. 
Changing https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/ClassDefinition/Data/Table.php#L357 
to
`if ($data === null || $unserializedData === null || $unserializedData === "") {`

so empty strings get considered as well.

jremmurd:
Fixes #11059

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Table.php
+++ b/models/DataObject/ClassDefinition/Data/Table.php
@@ -354,7 +354,7 @@ class Table extends Data implements ResourcePersistenceAwareInterface, QueryReso
     {
         $unserializedData = Serialize::unserialize((string) $data);
 
-        if ($data === null || $unserializedData === null) {
+        if ($data === null || $unserializedData === null || $unserializedData === "") {
             return [];
         }
 
--- a/models/DataObject/ClassDefinition/Data/Table.php
+++ b/models/DataObject/ClassDefinition/Data/Table.php
@@ -354,7 +354,7 @@ class Table extends Data implements ResourcePersistenceAwareInterface, QueryReso
     {
         $unserializedData = Serialize::unserialize((string) $data);
 
-        if ($data === null || $unserializedData === null || $unserializedData === "") {
+        if ($data === null || empty($unserializedData)) {
             return [];
         }
 


models/DataObject/ClassDefinition/Data/Table.php
```diff
@@ -354,7 +354,7 @@ public function getDataFromResource($data, $object = null, $params = [])
     {
         $unserializedData = Serialize::unserialize((string) $data);
 
-        if ($data === null || $unserializedData === null) {
+        if ($data === null || $unserializedData === null || $unserializedData === "") {
```
dvesh3: since it is already casted to string on line 355, don't think so null check is required
```suggestion
        if ($data === null || empty($unserializedData)) {
```

jremmurd: @dvesh3 But what if the table has 1 column an 1 row with value `0`?

dvesh3: data will be null in that case. please see https://github.com/blankse/pimcore/blob/66c89bf84736c0128b982cbfad42cabff2bfe779/models/DataObject/ClassDefinition/Data/Table.php#L455-L467

dvesh3:
@jremmurd one last thing.. we have to rebase it on 10.2 branch, since it is a fix.

PR #11067, opened 2021-12-21
[Improvement] Creating system thumbnails for grid view (like treepreview thumbnails) to improve performance

LucaExtera:
In order to improve performance in projects with many images to show in data objects grid view, we added a new system thumbnail type (like trepreview thumbnails) that the command pimcore:thumbnails:image -s now can create, instead of let Pimcore doing the (auto) thumbnails on demand.


brusch:
I'd say it would be even better to also use the already pre-generated system-preview image (which is also used for the tree) for the grid. Even if it's by far too big, it probably saves a lot if time and resources ... wdyt?

brusch:
@LucaExtera 🏓- can you please update your PR according to the feedback above? Thanks a lot in advance! 😊

brusch:
@LucaExtera last call 😉 Thanks!

brusch:
#

PR #11070, opened 2021-12-21
[Improvement] Only show area EditableDialogBox dialog when it actually has items to edit

rberneder:
**Situation**
An area is used multiple times throughout different templates.

**Problem**
The area shall offer an EditableDialogBox in the one template, but must not have such in the other one.

**Solution**
The area checks if there are items to be edited. If so, the dialog is output.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -697,7 +697,7 @@ class AssetController extends ElementControllerBase implements KernelControllerE
             ],
         ];
 
-        $hasChildren = (bool)$asset->getChildAmount($this->getAdminUser());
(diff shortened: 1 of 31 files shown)


blankse:
@rberneder Should already be fixed. See https://github.com/pimcore/pimcore/pull/10962

Edit: I see it is fixed for areablock and not for area. Maybe we should use the same fix?

rberneder:
You are right @blankse. I have adopted the implementation to the solution in #10962.

dvesh3:
@rberneder please rebase on 10.2 branch. thanks!

dvesh3:
@rberneder Now  we have unrelated commits in the PR. Could you please create a new PR with relevant changes? thanks!

dvesh3:
Fixed directly on 10.2 branch https://github.com/pimcore/pimcore/commit/82dae6902d9a78c48e8de9e4fdbd2ecf0cec7816 thanks!

PR #11073, opened 2021-12-21
[Data Object] Values of options provider for (multi) select data-type are strings

AlexanderWeckmer:
If custom provider provides an integer as value, the List-View of the multiselect field won't show selected items after reload.
Tags-View works, so I'm assuming a handling "Ext.ux.form.MultiSelect" vs. "Ext.form.field.Tag".

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/multiselect.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/multiselect.js
@@ -98,7 +98,7 @@ pimcore.object.tags.multiselect = Class.create(pimcore.object.tags.abstract, {
                 if(label.indexOf('<') >= 0) {
                     label = replace_html_event_attributes(strip_tags(label, "div,span,b,strong,em,i,small,sup,sub2"));
                 }
-                storeData.push({id: value, text: label});
+                storeData.push({id: value.toString(), text: label});
             }
         }
 


brusch:
I'm wondering if the provider should just return a string, as this is the type as it will be stored as well .... 🤔 It would also make sense in terms of switching over to being more strictly typed. Then of course it should be documented. 
Any other opinions?

brusch:
But just in case it would make sense, then we should cast the values to `string` already in the PHP code. This would be somewhere in this area: https://github.com/pimcore/pimcore/blob/b4040605bbbb0140bb97560ca4a6d1e1c7738e44/models/DataObject/ClassDefinition/DynamicOptionsProvider/SelectionProviderTrait.php#L52

AlexanderWeckmer:
I'm not so sure, maybe for custom fields a value of type integer would be nice and helpful, or when you want do some calculation based on the selected key?

brusch:
Values are being stored as strings anyway ... so I'd just work with strings.

brusch:
@Lexipowder any update? (today we're releasing `10.3` 😉)

AlexanderWeckmer:
Sorry no update. The project is currently on hold. Hopefully we will continue the project around April.
No time before that.

brusch:
Ok, then I'll close that for now. Thanks for your feedback 😊

PR #11086, opened 2021-12-23
[Bug] Fixed Warning: Undefined array key in $actions variable

pdchaudhary:
## Resolves 
Warning: Undefined array key "" in $actions variable

## Additional info  
Steps to reproduce

-   Enable debug mode
-   Create global targeting rule using action Code-Snippet
-   Check on the frontend side will get ' Warning: Undefined array key "codesnippet" '

image

### DIFF
--- a/lib/Targeting/EventListener/TargetingListener.php
+++ b/lib/Targeting/EventListener/TargetingListener.php
@@ -248,7 +248,7 @@ class TargetingListener implements EventSubscriberInterface
                 continue;
             }
 
-            if (!is_array($actions[$type])) {
+            if (array_key_exists($type,$actions) && !is_array($actions[$type])) {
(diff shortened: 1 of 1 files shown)


lib/Targeting/EventListener/TargetingListener.php
```diff
@@ -248,7 +248,7 @@ private function getResponseActions(VisitorInfo $visitorInfo): array
                 continue;
             }
 
-            if (!is_array($actions[$type])) {
+            if (array_key_exists($type,$actions) && !is_array($actions[$type])) {
```
blankse:
```suggestion
            if (!isset($actions[$type])) {
```

brusch:
Thank you very much!

PR #11095, opened 2021-12-27
Fix issue with remaining tmp files created by getting metadata

croitoruion:
Resolves #
Unlink tmp files after getting metadata info

## Additional info  

In Search Backend Data.php model, after setting data from element, methods "getEXIFData()" and "getIPTCData" is creating local files in tmp folder, but doesn't unlink them, so you can wake up with full storage.

### DIFF
--- a/models/Search/Backend/Data.php
+++ b/models/Search/Backend/Data.php
@@ -461,7 +461,9 @@ class Data extends \Pimcore\Model\AbstractModel
                 }
             } elseif ($element instanceof Asset\Image) {
                 try {
-                    $metaData = array_merge($element->getEXIFData(), $element->getIPTCData());
+                    $elementLocalPath = $element->getLocalFile();
+                    $metaData = array_merge($element->getEXIFData($elementLocalPath), $element->getIPTCData($elementLocalPath));
+                    unlink($elementLocalPath);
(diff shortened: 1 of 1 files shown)


brusch:
This shouldn't be necessary, since the cleanup is done here https://github.com/pimcore/pimcore/blob/557612f50201debd5a23e5b8cb4b2d0a63577d6d/lib/Helper/TemporaryFileHelperTrait.php#L78 in case you're dealing with a remote file. 

and `getEXIFData()` as well as `getXMPData()` is using 
https://github.com/pimcore/pimcore/blob/2e3829cf651a0c5a6de06c4943927c1aae871c67/models/Asset/MetaData/EmbeddedMetaDataTrait.php#L126

where `getLocalFile()` is called.

flamadrid-jsmdg:
@brusch @croitoruion I think we have similar issue with the command  **pimcore:search-backend-reindex**
We have about 70K assets in S3, so the tmp files generated will be cleaned up only after the script finish to execute due the register_shutdown_function. In our case the cronjob (as a kubernetes job) gets killed due lack of storage reaching around 65GB. I'm not sure if deleting the files at the end is the best strategy with that many assets. Could we improve it somehow, what do you think?

PR #11101, opened 2021-12-28
Fix: document as pimcore_build_nav param not possible anymore

blankse:
Get following error after update to Pimcore X:

`ContainerPcgwqDE\Navigation_62d2b4b::build(): Argument #1 ($params) must be of type array, Pimcore\Model\Document\Page given, called in /var/www/html/vendor/pimcore/pimcore/lib/Twig/Extension/NavigationExtension.php on line 72`


brusch:
Thanks a lot! 👍

PR #11103, opened 2021-12-29
[Imagick] Performance optimization when dealing with RGB color spaces

brusch:


lib/Image/Adapter/Imagick.php
```diff
@@ -345,17 +345,24 @@ private function setColorspaceToRGB()
...
+            // no need to process (s)RGB images
+            return $this;
+        }
+
         $profiles = $this->resource->getImageProfiles('icc', true);
 
-        // Workaround for ImageMagick (e.g. 6.9.10-23) bug, that let's it crash immediately if the tagged colorspace is
-        // different from the colorspace of the embedded icc color profile
-        // If that is the case we just ignore the color profiles
         if (isset($profiles['icc']) && in_array($imageColorspace, [\Imagick::COLORSPACE_CMYK, \Imagick::COLORSPACE_SRGB])) {
```
jdreesen: I think `COLORSPACE_SRGB` can be removed here, as we already return in line 350 if it's this colorspace.

PR #11111, opened 2021-12-30
Fix getChildren/hasChildren/getSiblings/hasSiblings on unsaved new models

### LINKED ISSUE #11104: [Bug]: getChildren() method of AbstractObject results in doctrine SyntaxErrorException error when object has null id
### Expected behavior

when calling `->getChildren()` on newly created object (before the first save), empty array should be returned.

### Actual behavior

In version 10.2.7 (not sure in which version exactly this change happened) newly created objects have `id=null` before the first save. As a result, Doctrine\DBAL\Exception\SyntaxErrorException exception is thrown.

### Steps to reproduce

Create a new object of class that inherits from AbstractObject class. Do not save that object. Call `->getChildren()` method on it.

blankse:
Resolves https://github.com/pimcore/pimcore/issues/11104

### DIFF
--- a/models/User/UserRole/Folder.php
+++ b/models/User/UserRole/Folder.php
@@ -28,14 +28,14 @@ class Folder extends Model\User\AbstractUser
     /**
      * @internal
      *
-     * @var array
+     * @var array|null
      */
-    protected $children = [];
+    protected $children;
 
     /**
      * @internal
      *
-     * @var bool
+     * @var bool|null
      */
     protected $hasChildren;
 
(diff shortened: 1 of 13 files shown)


models/User/UserRole/Folder.php
```diff
@@ -75,11 +79,13 @@ public function getChildren()
...
-            $this->hasChildren = true;
-        } else {
-            $this->hasChildren = false;
+        if (is_array($children)) {
+            $this->children = $children;
+            if (count($children) > 0) {
+                $this->hasChildren = true;
+            } else {
+                $this->hasChildren = false;
+            }
```
jdreesen:
```suggestion
            $this->hasChildren = count($children) > 0;
```

mcop1:
Thanks!

PR #11107, opened 2021-12-30
Website settings are not selected correctly in multi site environment

mittererr:
Website settings are not selected correctly (using `WebsiteSetting::getByName()`) when $siteID parameter is given and $langauge parameter ist not given. 
Changing the "ORDER BY" order fixed it for me.

PR #11112, opened 2022-01-01
Get website settings in multi site environment

mittererr:
Website settings are not selected correctly (using WebsiteSetting::getByName()) when $siteID parameter is given and $language parameter ist not given.
Changing the "ORDER BY" order fixed it for me. (Redone PR #11107)

### DIFF
--- a/models/WebsiteSetting/Dao.php
+++ b/models/WebsiteSetting/Dao.php
@@ -58,7 +58,7 @@ class Dao extends Model\Dao\AbstractDao
         if ($name != null) {
             $this->model->setName($name);
         }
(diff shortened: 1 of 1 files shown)


brusch:
I'm not sure if this is entirely correct, this needs to be checked in detail before we do any changes here. 
I'd also add tests first, so that we can be sure, everything works as expected.

blankse:
@mittererr @brusch Add an alternative PR with test and correct fix https://github.com/pimcore/pimcore/pull/11137

mittererr:
See #11137

Issue #10624, opened 2021-10-17
[Bug]: Configuration Properties Cache Issue


solverat:
### Expected behavior

Creating & using config properties on the fly.

### Actual behavior

Instant created configuration flags are not available in same request.

### Steps to reproduce

Example static route (happend in some of our functional tests):

```php
$route = new Staticroute();
$route->setValues([
    'name'      => 'test_route',
    'pattern'   => '/([a-zA-Z0-9-_]*)\\/come-on$/',
    'reverse'   => '/{%%_locale}/come-on',
    'controller'    => '',
    'variables' => '_locale',
]);

$route->save();

$path = $this->container->get('router')->generate('test_route', [], false);

dump($path);
```
![image](https://user-images.githubusercontent.com/700119/137626705-e7035a55-3ad3-4a4c-b1ba-842cde0e6f15.png)

dpfaffenbauer:
I don't see that as a problem honestly. It is the same behaviour as it is with Symfony Routes or any Symfony Configuration. Only issue: it is a BC break to how it worked with 10.0.

solverat:
While this is absolutely true, and I would agree to ignore this, these properties (we're not only talking about routes, this affects every`LocationAwareConfigRepository` configuration [each configuration node in `var/config`]) are editable in backend and so this **may** could lead to funny side effects or at least to misinterpretation/wrong handling on dev side. I'm going to close this for now.

brusch:
@solverat you could switch to `settings-store` backend for your tests, see: 
https://github.com/pimcore/pimcore/blob/10.x/tests/_bootstrap.php#L30

PR #11126, opened 2022-01-04
[LocationAwareConfig] Cleanup cache when saving or deleting configuration

fashxp:
otherwise saving config and getting it in same process afterwards will return the old config

### DIFF
--- a/lib/Model/Dao/PimcoreLocationAwareConfigDao.php
+++ b/lib/Model/Dao/PimcoreLocationAwareConfigDao.php
@@ -90,6 +90,17 @@ abstract class PimcoreLocationAwareConfigDao implements DaoInterface
     }
 
     /**
+     * Removes config with corresponding id from the cache.
+     * A new cache entry will be generated upon requesting the config again.
+     *
+     * @param string $id
+     */
+    protected function invalidateCache(string $id)
+    {
+         unset(self::$cache[$this->settingsStoreScope][$id]);
+    }
+
+    /**
      * @param string $id
      * @param array $data
      *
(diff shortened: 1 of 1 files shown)


lib/Model/Dao/PimcoreLocationAwareConfigDao.php
```diff
@@ -89,6 +89,17 @@ protected function loadIdList(): array
         return $this->locationAwareConfigRepository->fetchAllKeys();
     }
 
+    /**
+     * Removes config with corresponding id from the cache.
+     * A new cache entry will be generated upon requesting the config again.
+     *
+     * @param string $id
+     */
+    protected function invalidateCache(string $id)
```
blankse:
```suggestion
    protected function invalidateCache(string $id): void
```

solverat:
@fashxp While that's may be true, we're now dealing with the same issue described here again: https://github.com/pimcore/pimcore/issues/10624#issuecomment-945434601. 

Back then, we switched to the settings-store and the tests passed.
But now, they are failing again, it's not possible to run multiple tests with dynamically generated static routes. 

I guess that's because of this

https://github.com/pimcore/pimcore/blob/53336ca3d41ec874efaaffb48eea0f0252506301/lib/Model/Dao/PimcoreLocationAwareConfigDao.php#L100

but `self::$cache` somehow never gets populated again.

The weird part is: It works if I'm using different static route names in each test. Using the same static route name in multiple test cases won't work.

fashxp:
@solverat could you please debug a but on your end and see, if it comes to [that line](https://github.com/pimcore/pimcore/pull/11126/files#diff-857e6dafe136726af90a8c9d936a762f7174ee8ea67ef76dda1846ae4896f10cL72) the second time and if the data is loaded properly.

solverat:
Yes. I've debugged it. Its not the recreation itself. However, it is not possible anymore to generate a static route with the same name multiple times in the same request (if the route also gets deleted). This is mostly required in test cases and maybe a rare situation in real life.

solverat:
Btw. adding some (very simple) functional tests for setting store would have shown you this long before releasing anything in core.

fashxp:
any PR providing such a test is highly appreciated (since you seem to have them in place already anyway)!

PR #11127, opened 2022-01-04
[Imagick] Performance optimization for detecting clipping paths

brusch:


lib/Image/Adapter/Imagick.php
```diff
@@ -159,6 +163,34 @@ public function load($imagePath, $options = [])
         return $this;
     }
 
+    private function has8BIMClippingPath(): bool
+    {
+        $count = 0;
+        $chunkSize = 1024;
+        $handle = fopen($this->imagePath, 'rb');
+        $overlapString = '';
+        while ($chunk = fread($handle, $chunkSize)) {
```
NiklasBr, line 172: Why not load the first 100kB in a single `fread()` call rather than running in a `while` loop? That would eliminate the need for using `$overlapString` and `$count` variables altogether, no?

brusch: I was afraid that the regex get's horribly slow on bigger chunks, but I just did a small benchmark test and it seems it's doing still fine with 100k. Thanks for the feedback! See #11132

PR #11133, opened 2022-01-05
[Documents] Possibility to configure an URL prefix for generating the preview

### LINKED ISSUE #11078: [Bug][Documents] Compare Versions results in error
### Feature description

image

brusch:
In certain use cases, such as when running Pimcore in a Docker container, it isn't possible for Pimcore to determine the right URL prefix for generating the preview, so it is necessary to specify a prefix manually. 

Example for the use with the default `docker-compose.yml`: 
```yml
pimcore:
    documents:
        preview_url_prefix: 'http://nginx'
```

Fixes #11078


blankse:
@brusch I think that should be added to the documentation.

PR #11134, opened 2022-01-05
Composer: Bump requirements to the lowest possible version

blankse:
Checked the lowest possible version with:
`composer update --prefer-lowest`

Mostly is the version the first with PHP 8 support.
Otherwise there are dependencies on other packages.


brusch:
Totally makes sense - thanks a lot! 👌

PR #11137, opened 2022-01-05
Wrong website settings in multi site environment

blankse:
Alternative, correct fix for: https://github.com/pimcore/pimcore/pull/11112

- Add also test
- Add possibility to execute tests without PIMCORE_PROJECT_ROOT env


mittererr:
@blankse THX for fixing!

weisswurstkanone:
thanks a lot!

PR #11145, opened 2022-01-07
Remove unnecessary `iterator_to_array()` call in `Pimcore\Navigation\Container::addPages()`

jdreesen:
Calling `iterator_to_array()` shouldn't be necessary, if the only thing we do is iterating over it.

### DIFF
--- a/lib/Navigation/Container.php
+++ b/lib/Navigation/Container.php
@@ -155,6 +155,7 @@ class Container implements \RecursiveIterator, \Countable
      */
     public function addPages($pages)
     {
+        // @phpstan-ignore-next-line this should be checked via parameter type in Pimcore 11
         if (!$pages instanceof self && !is_array($pages)) {
             throw new \Exception('Invalid argument: $pages must be an array or an instance of ' . self::class);
         }
--- a/lib/Navigation/Container.php
+++ b/lib/Navigation/Container.php
@@ -155,12 +155,8 @@ class Container implements \RecursiveIterator, \Countable
      */
     public function addPages($pages)
     {
-        if ($pages instanceof self) {
-            $pages = iterator_to_array($pages);
-        }
-
-        if (!is_array($pages)) {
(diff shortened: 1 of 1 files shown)


brusch:
@jdreesen can you please check PhpStan? Thanks!

jdreesen:
PHPStan failed with:
```
------ ------------------------------------------------------------------- 
  Line   lib/Navigation/Container.php                                       
 ------ ------------------------------------------------------------------- 
  158    Result of && is always false.                                      
         💡 Because the type is coming from a PHPDoc, you can turn off this  
         check by setting treatPhpDocTypesAsCertain: false in your          
         phpstan.neon.                                                      
 ------ ------------------------------------------------------------------- 
```
In the [PHPStan docs](https://phpstan.org/config-reference#treatphpdoctypesascertain) it says:
> PHPStan by default doesn’t differentiate between PHPDoc and native types. It considers them both as certain.
>
> This might not be what you want in case you’re writing a library whose users might pass a wrong argument type to a function. Setting `treatPhpDocTypesAsCertain` to `false` relaxes some of the rules around type-checking.

Since the second paragraph applies to Pimcore, I think we should disable this option.

But now it fails with 6 other errors (5 of them are in the baseline and now don't occur anymore).

@blankse Could you please have a look? I'm not that experienced with PHPStan. Would it be okay to just remove the 5 baseline entries?

blankse:
@jdreesen I was already thinking about setting this phpstan property. However, I left it that way and included the errors in the baseline.
I think these exceptions should be thrown by type hints in a long term. So this errors will be a reminder for this. Maybe this is possible in Pimcore 11. You can add an `// @phpstan-ignore-next-line` comment above the if statement and a comment that this should be handled by type hint in Pimcore 11. WDYT?

jdreesen:
@blankse sounds good to me, thx!

dvesh3:
@jdreesen thanks!

PR #11146, opened 2022-01-07
Fix `preg_match()` usage in condition

jdreesen:
`preg_match()` returns `0` when there's no match and `false` on error (see [the docs](https://www.php.net/manual/en/function.preg-match.php#refsect1-function.preg-match-returnvalues)), so checking `!== 0` is dangerous because the condition is true when there's a match *or* an error.


brusch:
Totally makes sense, thanks!

PR #11147, opened 2022-01-07
Don't add the same page multiple times in `Pimcore\Navigation\Container::findAllBy()`

jdreesen:
If the page property is an array, and we iterate over it, we should skip the rest when we find a match. Otherwise, the same page may be added twice to the result array.


dvesh3:
@jdreesen Agree with your point. thanks!

PR #11149, opened 2022-01-07
Optimize finding active pages when building the navigation

jdreesen:
I just dug up an old branch from a previous project with many CMS pages (~2600) for which I once optimized the menu building to me more performant.

A lot of time is spent by iterating the whole tree to find the active path, which is not really necessary, because branches whose URI isn't a prefix of the requested URI can be skipped, resulting in a much faster iteration (time decreased from ~5 sec to ~2 sec in that project for the whole menu building).


mcop1:
Thanks :)

PR #11165, opened 2022-01-12
Remove superfluous `array_merge()` call

jdreesen:
`$bind` is an associative array, which means calling `$bind = array_merge($bind, $bind)` results in the exact same array and thus can be removed. See: https://3v4l.org/d47Wt


dpfaffenbauer:
🙈

brusch:
Thank you very much!

PR #10979, opened 2021-12-07
[Translations] Merge Admin & Shared translations into single UI 

dvesh3:
## Changes in this pull request  
Resolves partially #6568

## Additional info  
- Merge Admin & Shared translations within one single UI
- Shortcuts for domain filtering

PR #11168, opened 2022-01-12
[Translations] follow up to #10979

### LINKED ISSUE #11139: [Bug] Admin/Shared Translations not working
### Expected behavior

can edit translations

### Actual behavior

cannot edit translations

### Steps to reproduce

open admin translations and try to edit a translation after filtered ... 

image

brusch:
Resolves #11139
Related to #10979

Changed to row editing to: 

If the value is neither multiline plain text or HTML markup you can choose between 2 additional options or inline edit the single line content: 
image

if the existing value is already different from plain single line text, the options are limited based on the value type: 

image


bundles/AdminBundle/Resources/public/js/pimcore/settings/translation.js
```diff
@@ -167,6 +165,7 @@ pimcore.settings.translation.domain = Class.create({
                     pimcore.layout.refresh();
 
                 } catch (e) {
+                    console.log(e);
```
jdreesen:
```suggestion
                    console.error(e);
```

bundles/CoreBundle/Resources/translations/en.extended.json
```diff
@@ -672,6 +672,5 @@
   "sites": "Sites",
   "show_hide_areablock": "Show or Hide on frontend",
   "asset_type_changed": "Asset type was changed in the meantime. Please reload the admin interface to see the latest changes.",
-  "translation_domain": "Domain",
-  "translations": "Translations"
+  "translation_domain": "Domain"
```
dvesh3: Not used anywhere
```suggestion
```

dvesh3 (approved):

PR #11170, opened 2022-01-12
fixed custom report bar and line chart labels, as well as tooltips

### LINKED ISSUE #3005: [Custom Reports] Better titles in charts
**Bar chart**
- chart uses titles like "ext-chart-series-bar-*", better would be Label from configuration or column name
- for displaying more then 1 column, combined type would be better, now they are placed over, so only one bar is vissible

**Line chart**
- chart uses column names, better would be Label from configuration, is exists

### LINKED ISSUE #6444: Custom report - Chart tooltip values
https://github.com/pimcore/pimcore/blob/986d2f25467da5b626d38a2c14fec572e09dc3b9/bundles/AdminBundle/Resources/public/js/pimcore/report/custom/report.js#L363

Tootltip for multi yAxis bar/line charts displays value of last yAxis
tooltip should be related to item.field like this
````
tooltip.setHtml(record.get(data.xAxis) + ': ' + record.get(item.field));
````

### LINKED ISSUE #7110: Custom Report Bar Chart : Impossible to edit legends
# Bug Report

### Expected behavior

Set a custom legend to a custom report which use a Bar Chart
### Actual behavior

cannot configure the legends of the chart

### Steps to reproduce

- Create a custom report, configure chart settings. 

- No place to configure the legends (ext-chart-series-bar-1 etc displayed)

custom_report_bo

custom_report_legend

Using Pimcore 6.7.1

Thanks

Qubus0:
Resolves #3005, Resolves #6444, Resolves #7110

thorough bug report in #7110
tooltip issue in #6444

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/layout/portlets/customreports.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/layout/portlets/customreports.js
@@ -216,7 +216,7 @@ pimcore.layout.portlets.customreports = Class.create(pimcore.layout.portlets.abs
             for(var i = 0; i < data.yAxis.length; i++) {
                 var yAxis = data.yAxis[i];
                 series.push({
-                    displayName: this.columnLabels[data.yAxis[i]],
+                    title: this.columnLabels[data.yAxis[i]],
                     type: (data.chartType == 'line' ? 'line' : 'bar'),
                     xField: data.xAxis,
                     yField: yAxis,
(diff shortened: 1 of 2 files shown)


brusch:
@Qubus0 Thanks a lot for your PR. Could you please rebase it onto `10.2` branch, as it is a bug fix? 
Thanks in advance!

Qubus0:
@brusch whoops, done. thanks

PR #11172, opened 2022-01-12
Parameter &?pimcore_nocache dosn't work

ctippler:
The Service has to be public because it is accessed in the Cache.php  file via the Container...


brusch:
Thanks a lot! 👍

PR #11177, opened 2022-01-13
[Image gallery] Only load image gallery items with an image / Remove gallery item when deleting image asset

BlackbitDevs:
Steps to reproduce:
1. Create data object class with an image gallery
2. Upload an image to Pimcore assets
3. Create an object and assign the uploaded image to the gallery
4. Delete the image asset
5. Reload the object -> There is an empty placeholder in the image gallery

It is not possible to assign metadata to an image gallery item which has no image:
https://github.com/pimcore/pimcore/blob/4661359d9af40387ed5e7f14d00a81a7fb1979a0/models/DataObject/ClassDefinition/Data/ImageGallery.php#L268-L282
There the `getDataFromResource()` from `Hotspotimage` gets called and this returns `null` when the image cannot be loaded:
https://github.com/pimcore/pimcore/blob/4661359d9af40387ed5e7f14d00a81a7fb1979a0/models/DataObject/ClassDefinition/Data/Hotspotimage.php#L179-L180

So it does not make sense to load an image gallery item when the HotspotImage could not be loaded. This PR prevents loading such values as `null` and so preventing the empty image gallery item from being loaded.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/ImageGallery.php
+++ b/models/DataObject/ClassDefinition/Data/ImageGallery.php
@@ -278,7 +278,9 @@ class ImageGallery extends Data implements ResourcePersistenceAwareInterface, Qu
             ];
 
             $itemResult = $fd->getDataFromResource($itemData, $object, $params);
-            $resultItems[] = $itemResult;
+            if($itemResult instanceof DataObject\Data\Hotspotimage) {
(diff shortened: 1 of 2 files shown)


weisswurstkanone (approved):

weisswurstkanone:
thanks!

PR #11159, opened 2022-01-11
ON DUPLICATE KEY UPDATE will set row id to 0 if 2 processes try to insert the same value at the same time

HaaseIT:
## Changes in this pull request  
Resolves 11109

PR #11180, opened 2022-01-13
[DB] Added PK cache to DB insertOrUpdate()

brusch:
Follow up to #11159

### DIFF
--- a/lib/Db/PimcoreExtensionsTrait.php
+++ b/lib/Db/PimcoreExtensionsTrait.php
@@ -37,6 +37,8 @@ trait PimcoreExtensionsTrait
      */
     protected $autoQuoteIdentifiers = true;
 
+    private static array $tablePrimaryKeyCache = [];
+
     /**
      * @see \Doctrine\DBAL\Connection::connect
      */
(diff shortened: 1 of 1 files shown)


lib/Db/PimcoreExtensionsTrait.php
```diff
@@ -341,8 +343,13 @@ public function fetchPairs($sql, array $params = [], $types = [])
...
         // get the field name of the primary key
-        $tableDetails = $this->getSchemaManager()->listTableDetails($table);
-        $fieldsPrimaryKey = (array)$tableDetails->getPrimaryKeyColumns();
+        if(isset(self::$tablePrimaryKeyCache[$table])) {
+            $fieldsPrimaryKey = self::$tablePrimaryKeyCache[$table];
+        } else {
+            $tableDetails = $this->getSchemaManager()->listTableDetails($table);
+            $fieldsPrimaryKey = (array) $tableDetails->getPrimaryKeyColumns();
+            self::$tablePrimaryKeyCache[$table] = $fieldsPrimaryKey;
+        }
```
jdreesen: This could be written as follows - not sure if it's more readable, though:

```suggestion
        $fieldsPrimaryKey = self::$tablePrimaryKeyCache[$table] ??= (array) $this->getSchemaManager()->listTableDetails($table)->getPrimaryKeyColumns();
```

brusch: Yep, makes sense, thanks 😊

PR #11187, opened 2022-01-14
decode quotation marks to make exact search working

croitoruion:
Resolves #
ensure we have double quote (") character on search query to make Boolean Full-Text Searches "exact search" working, as now quote is encoded (&quot;)

## Additional info  

https://dev.mysql.com/doc/refman/5.6/en/fulltext-boolean.html


brusch:
@croitoruion Thanks a lot for your PR. Could you please rebase it onto `10.2` as it is a bug fix? Thanks in advance!

brusch:
Closed in favor of #11195

PR #11189, opened 2022-01-14
Pass editmode as attribute to document render events

jremmurd:
# Enhancement

Pass editmode as attribute to document render events.

This is especially useful, because documents get now rendered in `Pimcore\Document\Editable\EditableUsageResolver` while saving a document in the pimcore backend and in this context data object inheritance is disabled and the EditmodeResolve will not recognize that the editmode is active. If you depend on data object inheritance during the rendering process the saving process will fail. 

Passing the editmode attribute to the render event enables developers to enable inheritance during redering in editmode on document save.

__
As an alternative Pimcore could set the data object inheritance to the same setting as during the frontend rendering process by default.

### DIFF
--- a/lib/Document/Renderer/DocumentRenderer.php
+++ b/lib/Document/Renderer/DocumentRenderer.php
@@ -28,6 +28,7 @@ use Pimcore\Templating\Renderer\ActionRenderer;
 use Pimcore\Twig\Extension\Templating\Placeholder\ContainerService;
 use Symfony\Component\HttpKernel\Fragment\FragmentRendererInterface;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+use Pimcore\Http\Request\Resolver\EditmodeResolver;
 
 class DocumentRenderer implements DocumentRendererInterface
 {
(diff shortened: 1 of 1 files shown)


lib/Document/Renderer/DocumentRenderer.php
```diff
@@ -111,8 +111,14 @@ public function setContainerService(ContainerService $containerService)
      */
     public function render(Document\PageSnippet $document, array $attributes = [], array $query = [], array $options = []): string
     {
+        $eventAttributes = array_filter(
+            $attributes,
+            fn($key) => in_array($key, [EditmodeResolver::ATTRIBUTE_EDITMODE]),
```
jdreesen: I think you forgot to add a `use` statement: `use Pimcore\Http\Request\Resolver\EditmodeResolver`

jremmurd: thx!

jremmurd:
Any updates if you plan to merge this PR any time soon?

brusch:
@jdreesen why not passing all attributes? 🤔

jdreesen:
I think you wanted to mention @jremmurd :wink:

jremmurd:
> @jdreesen why not passing all attributes? 🤔

@brusch No specific reason, I don't know for what those other attributes are all for, I just didn't want to bloat the event unnecessarily. If you think, we do not need to filter it, then I'll change it ;-)

brusch:
@jremmurd I'd include all attributes, as it doesn't hurt 😊Can you please update the PR? Thanks!

jdreesen:
It's currently targeted for `10.4`. Maybe it still fits into `10.3`, small as it is?

brusch:
Merci! 😻

Issue #11193, opened 2022-01-14
[Bug]: Expected variable in catch block in initializeContainer()


antonio-masotti:
### Expected behavior

in `pimcore/lib/Kernel.php` at line 260 the catch block is missing the variable for the exception and this throws an error:
The exception should be handled, logged or ignored.

I think it could just be enough to give the exception a variable name like this:

~~~php
 try {
                $container = $this->getContainer();
            } catch (\LogicException $e) {
                // Container is cleared. Allow tests to finish.
            }
~~~

I'll try to submit a PR if this can help.

### Actual behavior

Current code: 

~~~php
 try {
                $container = $this->getContainer();
            } catch (\LogicException) {
                // Container is cleared. Allow tests to finish.
            }
~~~

once in the catch block, I get this error:

`syntax error, unexpected ')', expecting '|' or variable (T_VARIABLE)`

![image](https://user-images.githubusercontent.com/91308510/149531515-40dd2be9-3708-4aa1-aec9-871a0ccf8468.png)

### Steps to reproduce

It happens when the a LogicException occurs while initializing the container.

jdreesen:
There's nothing wrong with it, it's [supported in PHP 8](https://wiki.php.net/rfc/non-capturing_catches). Are you sure you're using the right PHP version?

antonio-masotti:
Interesting... I didn't know (thanks for the docs) but yes  I'm sure:

![image](https://user-images.githubusercontent.com/91308510/149560564-62a32633-f2eb-4618-94a8-e1de7287e33c.png)

Still I get that error.

antonio-masotti:
I've closed the PR. 

I still had composer associated with PHP 7.4 instead of 8.0. 
jdreesen was of course right.

PR #11194, opened 2022-01-14
assign variable name to LogicException

antonio-masotti:
Resolves (hopefully) #11193

## Additional info  

see Issue


jdreesen:
As the variable is not used, it makes no sense adding it. [PHP 8 supports to do that](https://wiki.php.net/rfc/non-capturing_catches).

antonio-masotti:
@jdreesen  I've found the problem. You were of course right, it's ok in PHP 8.

I would then not purse this PR anymore. Thanks for the link.

PR #11195, opened 2022-01-14
Fix: Decode quotation marks to make exact search working

croitoruion:
Resolves #
ensure we have double quote (") character on search query to make Boolean Full-Text Searches "exact search" working, as now quote is encoded (")
## Additional info  
https://dev.mysql.com/doc/refman/5.6/en/fulltext-boolean.html


brusch:
Thank you very much!

Issue #11186, opened 2022-01-14
[Bug]: copy & pasting object with URL slug fails


passioneight:
### Expected behavior

When copy & pasting an object containing an `URL Slug` field, which has a value set, the slug itself is not copy & pasted to avoid URL conflicts.

### Actual behavior

The slug is copy & pasted, resulting in an error as shown in the image.

![image](https://user-images.githubusercontent.com/57256630/149491287-0969a332-d602-4167-82d4-b6a721f7954d.png)

### Steps to reproduce

1. Create a class definition containing an `URL Slug`
2. Create an object
3. Set the slug to any valid value (e.g. `/en/test/object-1`)
4. Copy & Paste the object

weisswurstkanone:
The Slug itself has to be unique. What we maybe could do is to leave the field empty.

passioneight:
Yes, that's what I suggested.

blankse:
@passioneight I created an PR: https://github.com/pimcore/pimcore/pull/11197

Can you test it?

passioneight:
@blankse thanks, I'll give it a try shortly.

##### Update
Unfortunately, it doesn't work - still the same behaviour. If I find some time, I'll further investigate. For now, I'll just not copy-paste the objects.

blankse:
@passioneight I tested it with an AccessoryPart object in the demo and it works for me.

brusch:
Yes, I agree, the value of the slug shouldn't be copied. @blankse thanks for your PR 😉

passioneight:
FYI, just updated to the latest version and the error still persists. Did you try with localized slugs too?

blankse:
@passioneight No the url slug in the demo AccessoryPart class is not localized. The copy process checks no localized fields for unique. This should be extended.

blankse:
@passioneight I create an PR for that. Can you test https://github.com/pimcore/pimcore/pull/11214 ?

passioneight:
@blankse thanks, I've added suggested changes based on my test - see https://github.com/pimcore/pimcore/pull/11214#pullrequestreview-856725869

PR #11197, opened 2022-01-16
Fix: copy & pasting object with URL slug fails

blankse:
Resolves https://github.com/pimcore/pimcore/issues/11186


weisswurstkanone:
thanks @blankse

PR #11202, opened 2022-01-17
[Custom Reports] Validate & escape CSS classes and names

brusch:


bundles/AdminBundle/Resources/public/js/pimcore/report/custom/item.js
```diff
@@ -894,7 +894,25 @@ pimcore.report.custom.item = Class.create({
 
     save: function () {
 
-        var m = this.getValues();
+        let m = this.getValues();
+        let error = false;
+
+        ['group', 'groupIconClass', 'iconClass', 'niceName', 'reportClass'].forEach(function (name) {
```
dvesh3, line 900:
```suggestion
        ['group', 'groupIconClass', 'iconClass', 'name', 'niceName', 'reportClass'].forEach(function (name) {
```

PR #11205, opened 2022-01-17
[Targeting] Escape/validate names of rules properly

brusch:


bundles/AdminBundle/Controller/Admin/TargetingController.php
```diff
@@ -37,6 +37,11 @@ class TargetingController extends AdminController implements KernelControllerEve
 {
     // RULES
 
+    private function correctName(string $name): string
+    {
+        return preg_replace('/[#\?\*\:\\\\<\>\|"%&@=;\+]/', '-', $name);
```
jdreesen: This much escaping is not necessary here, the following should be sufficient:

```suggestion
        return preg_replace('/[#?*:\\\\<>|"%&@=;+]/', '-', $name);
```

brusch: You're right, thanks!

PR #11206, opened 2022-01-17
[Ecommerce] Discounts shouldn't be negative

### LINKED ISSUE #11024: [Ecommerce]: Pricing Rules - Do not allow negative discounts
### Expected behavior

Not possible to set negative discounts

### Actual behavior

It is possible to set negative discounts for Cart & Product discount actions, then adds it to to total amount.

### Steps to reproduce

1. Login to the application https://demo.pimcore.fun/admin
2. Navigate to Online shop -> Pricing Rules -> Voucher Discount -> Actions``
3. Enter Negative amount in Cart Discount and click on save.

brusch:


fashxp:
this might be a BC break ... if someone (mis)uses pricing rules as additional addons to the cart. 
should we do this with 10.3?

fashxp:
fixes #11024

PR #11208, opened 2022-01-17
[Settings] Validate SVG uploads for branding

brusch:


bundles/AdminBundle/Controller/Admin/SettingsController.php
```diff
@@ -109,6 +109,12 @@ public function uploadCustomLogoAction(Request $request)
             throw new \Exception('Unsupported file format');
         }
 
+        if($fileExt === 'svg') {
+            if(strpos(file_get_contents($_FILES['Filedata']['tmp_name']), '<script')) {
+                throw new \Exception('Scripts in SVG files are not supported');
+            }
+        }
```
blankse:
```suggestion
        if ($fileExt === 'svg' && stripos(file_get_contents($_FILES['Filedata']['tmp_name']), '<script')) {
            throw new \Exception('Scripts in SVG files are not supported');
        }
```
Use stripos to be case-insensitive

brusch: 👍👍 Thanks!

PR #11213, opened 2022-01-18
Headless Chrome settings not read

mittererr:
Headless Chrome settings are not read correctly from Config object. I think the right way to do this stuff is to use the getter!?


blankse:
@mittererr Seems to be a bugfix. You should create this on branch 10.2

PR #11214, opened 2022-01-18
Copy DataObject: Check also localized fields for unique

blankse:
It's impossible to copy objects with localized unique fields. See https://github.com/pimcore/pimcore/issues/11186

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -32,6 +32,7 @@ use Pimcore\Model\DataObject\ClassDefinition\Data\IdRewriterInterface;
 use Pimcore\Model\DataObject\ClassDefinition\Data\LayoutDefinitionEnrichmentInterface;
 use Pimcore\Model\Element;
 use Pimcore\Model\Element\DirtyIndicatorInterface;
+use Pimcore\Tool;
 use Pimcore\Tool\Admin as AdminTool;
 use Pimcore\Tool\Session;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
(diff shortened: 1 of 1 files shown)


models/DataObject/Service.php
```diff
@@ -217,7 +213,16 @@ public function copyAsChild($target, $source)
 
         if ($new instanceof Concrete) {
             foreach ($new->getClass()->getFieldDefinitions() as $fieldDefinition) {
-                if ($fieldDefinition->getUnique()) {
+                if ($fieldDefinition instanceof DataObject\ClassDefinition\Data\Localizedfields) {
+                    foreach ($fieldDefinition->getFieldDefinitions() as $localizedFieldDefinition) {
+                        if ($fieldDefinition->getUnique()) {
```
passioneight: This needs to be `$localizedFieldDefinition` instead of `$fieldDefinition`.

models/DataObject/Service.php
```diff
@@ -217,7 +213,16 @@ public function copyAsChild($target, $source)
...
+                            foreach (Tool::getValidLanguages() as $language) {
+                                $new->set($fieldDefinition->getName(), null, $language);
```
passioneight: This needs to be `$localizedFieldDefinition` instead of `$fieldDefinition`.

passioneight (changes requested):
I've tested your PR. Great work! Only 2 changes are needed to resolve the issue.

models/DataObject/Service.php
blankse: Thanks for testing, fixed it https://github.com/pimcore/pimcore/pull/11214/commits/c1fb7af9d7e4c8cddc5db3f77de5a66d49e22988

models/DataObject/Service.php
blankse: Thanks for testing, fixed it https://github.com/pimcore/pimcore/pull/11214/commits/c1fb7af9d7e4c8cddc5db3f77de5a66d49e22988

weisswurstkanone:
Thank you!

PR #11218, opened 2022-01-18
[Bug] Headless Chrome settings not reader

mittererr:
Headless Chrome settings are not read correctly from Config object. I think the right way to do this stuff is to use the getter!?


dvesh3:
@mittererr thanks!

Issue #10654, opened 2021-10-19
[Improvement]: Use foreign keys for `ON DELETE` whenever possible


brusch:
### Improvement description

- [x] `assets_metadata.cid` ( #11072 ) 
- [x] `documents_*` tables (#11236)
- [ ] ~~`edit_lock.cid/ctype`~~ not possible as reference is based on multiple tables
- [ ] ~~`dependencies` (source & target - if possible)~~
- [ ] ~~`element_workflow_state.cid/ctype`~~
- [x] `gridconfig_favourites` (#11219)
...

I'd go for a PR for each entity to make reviewing easier.

kanevbg:
I was searching for info related to why pimcore does not have any foreign keys for the objects and found that improvement issue, which interestingly still does not mention objects.
Can you provide some info on why there is no FK for ON DELETE for objects.
I'd like being able to truncate object_X table, where X is the id of the class and all the related data for these object to be deleted from DB.

PR #11219, opened 2022-01-18
[GridConfig] Use foreign keys for ON DELETE #10654

dvesh3:
Resolves partially #10654

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20220119082511.php
+++ b/bundles/CoreBundle/Migrations/Version20220119082511.php
@@ -16,8 +16,9 @@ final class Version20220119082511 extends AbstractMigration
 
     public function up(Schema $schema): void
     {
+        $this->addSql('ALTER TABLE `gridconfig_favourites` CHANGE `gridConfigId` `gridConfigId` int(11) NOT NULL');
         $this->addSql('ALTER TABLE `gridconfig_favourites`
-            CHANGE `gridConfigId` `gridConfigId` int(11) NOT NULL,
+            ADD INDEX `grid_config_id` (`gridConfigId`),
             ADD CONSTRAINT `fk_gridconfig_favourites_gridconfigs`
             FOREIGN KEY (`gridConfigId`)
             REFERENCES `gridconfigs` (`id`)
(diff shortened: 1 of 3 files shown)


bundles/CoreBundle/Migrations/Version20220119082511.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Use foreign key for delete cascade on gridconfig_favourites & gridconfig_shares';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE `gridconfig_favourites`
+			CHANGE `gridConfigId` `gridConfigId` int(11) NOT NULL,
```
jdreesen: Tabs to spaces:

```suggestion
        CHANGE `gridConfigId` `gridConfigId` int(11) NOT NULL,
```

bundles/CoreBundle/Migrations/Version20220119082511.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Use foreign key for delete cascade on gridconfig_favourites & gridconfig_shares';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE `gridconfig_favourites`
+        CHANGE `gridConfigId` `gridConfigId` int(11) NOT NULL,
```
jdreesen: I'm sorry, I used too few spaces :/ 
This is right:
```suggestion
            CHANGE `gridConfigId` `gridConfigId` int(11) NOT NULL,
```

bundles/InstallBundle/Resources/install.sql
```diff
@@ -879,6 +879,7 @@ CREATE TABLE `gridconfig_favourites` (
     PRIMARY KEY (`ownerId`, `classId`, `searchType`, `objectId`),
 	INDEX `classId` (`classId`),
 	INDEX `searchType` (`searchType`),
+    INDEX `grid_config_id` (`gridConfigId`),
```
blankse, line 882: The other indexes are in camel case, why not this also?

dvesh3: So that constraint references column instead of index. Otherwise you won't be able to delete index in down migration.

blankse: Okay, thanks for clarifying

PR #11221, opened 2022-01-19
Sorting bug in classification store

mcop1:
1.) Create classification store with 2 groups (e.g. a, b)
2.) Create an object and add group a, make sure inheritance is activated for this object
3.) Try to import a child object with data objects importer and add group b
4.) Import succeeds, but there will be an exception when you try to open the imported object

Found while testing pr pimcore/data-importer#148


weisswurstkanone:
@mcop1 , can we add a unit test for this scenario ?

fashxp:
should not hurt, but there is an underlying problem: 
see https://github.com/pimcore/pimcore/pull/11270

fashxp (approved):

PR #11226, opened 2022-01-20
Document Types list breaks when showing creation or modification date

### LINKED ISSUE #11224: [Bug]: Document Types list breaks when showing creation or modification date
### Expected behavior

In the Document Types List all available columns can be shown properly

### Actual behavior

As soon as either the creation date or the modification date columns are selected to be shown, the new date columns are empty, and additionally the buttons in the delete and translate columns disappear.

There is also a JavaScript error in the console:
ext-all.js?_dc=a23ccb2bfaf92b1723e16ddd94e6cd296c50387f:1878 [W] XTemplate evaluation exception: Cannot read properties of undefined (reading 'format')

### Steps to reproduce

Reproduced on https://demo.pimcore.com/

Click on Settings > Document Types

document_types_before

Click the arrow down button on any of the visible columns
Check either the Creation Date or the Modification Date (or both)

=> See that the newly enabled columns are empty
=> Additionally the delete and translate buttons are gone.

document_types_after

dturopoli:
Fixes #11224

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/docTypes.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/docTypes.js
@@ -191,7 +191,7 @@ pimcore.settings.document.doctypes = Class.create({
                 renderer: function (d) {
                     if (d !== undefined) {
                         var date = new Date(d * 1000);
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
thanks a lot!

PR #11227, opened 2022-01-20
[Bugfix]: Users and Roles permissions on Custom Reports aren't displayed, resolves #11167

### LINKED ISSUE #11167: Users and Roles permissions on Custom Reports aren't displayed
### Expected behavior

When a user fills the "Visible To Users" and "Visible To Roles" boxes in a Custom Report setting, the information that was entered should be displayed even after reloading.

### Actual behavior

Users and Roles' names aren't loaded when you close and re-open the Custom Report setting, even though user permissions are working properly and, analyzing the GET request which loads Custom Report data, you can see they are returned inside the "sharedUserNames" and "sharedRoleNames" arrays.

### Steps to reproduce

1. Create a Custom Report
2. Under the "Permission" section, uncheck the "Shared Globally" option and select some users or roles from the tag fields underneath
3. Save the Custom Report
4. Close and re-open it. You'll notice "Visible To Users" and "Visible To Roles" boxes are empty, even though permissions are working, as long as you don't save again the report with the unfilled "Permissions" section.

https://user-images.githubusercontent.com/56591443/149167544-8f47ae2d-14e1-464b-90b2-bd473762f91b.mp4

dturopoli:
Fixes #11167

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/report/custom/item.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/report/custom/item.js
@@ -442,7 +442,7 @@ pimcore.report.custom.item = Class.create({
             });
 
             var userSharingField = Ext.create('Ext.form.field.Tag', {
-                name: "sharedUserIds",
+                name: "sharedUserNames",
                 width: '100%',
                 height: 100,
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Merci!

PR #11233, opened 2022-01-20
[DataObjects] Remove @internal from getRelationData()

jremmurd:
The docs mention `getRelationData()` as best practice to get data from Reverse Object Relation Datatype fields, so it should NOT be an internal method.

See https://pimcore.com/docs/pimcore/10.3/Development_Documentation/Objects/Object_Classes/Data_Types/Reverse_Object_Relation_Type.html#page_Working-with-PHP-API


brusch:
Thanks, you are totally right! 👍

PR #11240, opened 2022-01-21
Don't catch exception when resolving a class

passioneight:
For example, when using a `LinkGenerator`, the error message gets logged and that's about it.

This change will let the application throw an exception, so that any invalid reference used displays an error message right away.


dvesh3:
BC Break? As existing implementations with missing classes will start throwing exception.

passioneight:
Fair point, though, I'm not sure if there is any use case where the developer just left it as is when `null` was returned. But it could be. May be better off in v11?

passioneight:
<s>Can anyone list the steps for a proper rebase?

Here's what I did:
1. Cloned my repo.
2. Added pimcore/pimcore as remote
3. git fetch remote
4. Checkout patch-1
5. git rebase pimcore/11.x
6. Resolved merge conflict
7. Commit & Push
</s>

Turns out the issue was some setting in PhpStorm that merged the branch from origin after rebasing.

PR #11246, opened 2022-01-21
[Improvement]: default value for preview config

### LINKED ISSUE #11245: [Improvement]: default value for preview config
### Improvement description

Currently, the preview config in the [documentation](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Class_Settings/Preview_Generator.html#page_Sample-PreviewProvider-Implementation) looks like this:

```php
public function getPreviewConfig(\Pimcore\Model\DataObject\Concrete $object): array {
    return [
        [
            'name' => '_locale',
            'label' => 'Locale',
            'values' => [
                'English' => 'en',
                'German' => 'de'
            ]
        ],
        // Omitted ...
    ];
}
```

It would be great to also be able to define a `defaultValue`, like so (pseudo-code):

```php
public function getPreviewConfig(\Pimcore\Model\DataObject\Concrete $object): array {
    return [
        [
            'name' => '_locale',
            'label' => 'Locale',
            'values' => [
                'English' => 'en',
                'German' => 'de'
            ],
            'defaultValue' => 'en'
        ],
        // Omitted ...
    ];
}
```

This feature is especially helpful, when using locales to generate the preview as it allows us to pre-select the locale based on some logic (e.g., the user's locale).

dturopoli:
Resolves #11245

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/preview.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/preview.js
@@ -124,6 +124,7 @@ pimcore.object.preview = Class.create({
 
             let paramSelect = Ext.create('Ext.form.ComboBox', {
                 fieldLabel: params[i].label ? params[i].label : params[i].name,
+                value: params[i].defaultValue ? params[i].defaultValue : '',
                 name: params[i].name,
                 store: selectOptions,
                 queryMode: 'local',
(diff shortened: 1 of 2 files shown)


passioneight:
@dturopoli thanks, works.

brusch:
LGTM, thanks a lot! 👍

PR #11249, opened 2022-01-21
Add test env files/folders to .gitignore

blankse:
Splitting from #10737

If you execute the setup environment script `.github/ci/scripts/setup-pimcore-environment.sh` it should be no files for versioning.


brusch:
I'm not sure whether this could have some dangerous side-effects, can you maybe explain the background of this change a bit more in detail? Thanks 😊

blankse:
@brusch If I work in pimcore only environment to execute the codeception or phpstan tests (like the github action does):
```
git clone https://github.com/pimcore/pimcore.git
cd pimcore
.github/ci/scripts/setup-pimcore-environment.sh
composer install

vendor/bin/codecept run -c . -vvv --json
vendor/bin/phpstan analyse --memory-limit=-1
```

After that you have this files in the git versioning:  `git status`
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        .env
        bin/console
        composer.lock
        config/config.yaml
        config/packages/
        config/routes/
        config/services.yaml
        config/system.yml
        public/
        src/
        templates/
        var/
        vendor/
```

After this change:
`nothing to commit, working tree clean`

Than you make changes/add new files. You have to separate it from the folder/files from `setup-pimcore-environment.sh` / composer. This should be added to the .gitignore to simplify it to `git add . && git commit -m "Message"`

A .gitignore only can affect files in its path and below. Because of the / sign at the beginning of each search line it can only affect files in the same path as the .gitignore itself. So these folders are still possible in subfolders. So I see no side effects.

```
mkdir -p test/bin
touch test/bin/console
git status
```
```Untracked files:
  (use "git add <file>..." to include in what will be committed)
        test/
```

Only lines without / at the beginning can match with files/folders in subfolders.

What are your concerns?

brusch:
Thanks for clarifying - makes totally sense 👍😊 Thanks!

PR #11250, opened 2022-01-21
FIX: invisible fields saved incorectly when fieldcollections swap places

### LINKED ISSUE #11201: [Bug]: Invisible elements inside of fieldcollections getting mixed up when the fieldcollection is sorted
### Expected behavior

When sorting fieldcollections with the arrow buttons in the backend, every child element of the collection should keep it's parent-child relation.

### Actual behavior

When sorting fieldcollections with the arrow buttons in the backend, invisible elements inside of the collection are keeping their positions and therefore are getting mixed up and becoming children of a different fieldcollection.

Tested in Pimcore 10.2.4 + 10.2.8

### Steps to reproduce

Create a new fieldcollection containing a panel, an input field (e.g. 'collectionTitle') and a second input field ("invisibleTitle"). Keep the second input visible for now.

Create a new class to add the fieldcollection to.

Create an object of this class and add at least two fieldcollections to it. Enter distinctive titles for the collections (e.g. "collection-1", "collection-2") and the "invisibleText" fields (e.g. "invisble-1", "invisible-2") . Save the created object.

If you change the position of the fieldcollection and save and reload the object, everything should be ok.

Now change the visibility of the second input field to "invisble" . Afterwards reload the created test object, change the position of the fieldcollections and save the object. Revert the visibility of the second input field to "visible", reload the test object and the 
 elements should be mixed up.
 
 [edit]: The first version of this bug report only mentionend invisible block elements, but it seems that it's a general behaviour with any invisible element inside of fieldcollections.

BabovicT:
Problem was occurring when swapping 2 fieldcollections inside object swap places using arrows. Problem was only occurring if user saves(auto save is also causing this) and then publishes object.  Resolves #11201

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/fieldcollections.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/fieldcollections.js
@@ -102,7 +102,7 @@ pimcore.object.tags.fieldcollections = Class.create(pimcore.object.tags.abstract
 
     postSaveObject: function(object, task) {
 
-        if (object.id == this.object.id && task == "publish") {
+        if (object.id == this.object.id) {
             for (var itemIndex = 0; itemIndex < this.component.items.items.length; itemIndex++) {
                 var item = this.component.items.items[itemIndex];
                 item["pimcore_oIndex"] = itemIndex;


mcop1:
Thanks!

PR #11252, opened 2022-01-22
Set Layout properties to internal

blankse:
Splitting from https://github.com/pimcore/pimcore/pull/10824

I think these attributes should be internal. It should be used the getter and setter. There are only public to send it to JavaScript.


brusch:
@blankse totally makes sense - thank you very much!

Issue #11144, opened 2022-01-07
[Bug]: Caching - ClassificationStore KeyConfig - Pimcore 10.2.5


awroblewski:
### Expected behavior

Hello,

This is about creating/modifiying Classification store Keys programatically. There is no backend involvement here.

Step1 - create key
I create a classification store key using constructor \Pimcore\Model\DataObject\Classificationstore\KeyConfig, setting neccessary fields subsequently saving the KeyConfig using the ->save() method.

Step2 - modify key
Retrieving the field using KeyConfig::getByName($name, $storeId), modify the keys again saving using ->save() method.

Step - use key
Retrieve the field using KeyConfig::getByName($name, $storeId), expect to see updated data from Step 2.

### Actual behavior

Step1 - create key
I create a classification store key using constructor \Pimcore\Model\DataObject\Classificationstore\KeyConfig, setting neccessary fields subsequently saving the KeyConfig using the ->save() method -> all OK

Step2 - modify key
Retrieving the field using KeyConfig::getByName($name, $storeId), modify the keys again saving using ->save() method.

Step - use key
Retrieving the field using KeyConfig::getByName($name, $storeId), currently I see old outdated date from Step 1 as long as I do not clear the pimcore data cache using console p:c:c

### Steps to reproduce

Step 1 - create key
this is about creating classification store keys programatically. When creating a key using the \Pimcore\Model\DataObject\Classificationstore\KeyConfig constructor setting the fields subsequently saving the data using the save() method, the data is being written to the database and the correspondig cache item, if exists, is being invalidated using the calculated cache key

```
vendor/pimcore/pimcore/models/DataObject/Classificationstore/KeyConfig.php
(Line 296)            $cacheKey = 'cs_keyconfig_' . $this->getId();
(Line 297)            Cache::remove($cacheKey);
```

Step 2 - modify key
Next I want to modify keydata. First I retrieve the key by name using static KeyConfig::getByName($name, $storeId) from the KeyConfig class. This tries to retrieve the keyConfig from the cache (in my env it is redis) using the cache key

```
vendor/pimcore/pimcore/models/DataObject/Classificationstore/KeyConfig.php
(Line 155)  $cacheKey = 'cs_keyconfig_' . $storeId . '_' . md5($name);
(Line 164)  Cache::load($cacheKey);

```
[...]

PR #11254, opened 2022-01-22
#11144: adding cache invalidation for classification store KeyConfig

awroblewski:
Pull request as per issue #11144

@brusch , as suggested I've created a static method to separate the cache key creation logic as per your given example. The cache keys for the classification store keys are now calculated using this method.

@weisswurstkanone , I've implemented the same for the GroupConfig, however, the GroupConfig did not use caching on behalf of the group name but just the id. So the current cache invalidation stays the same. Does it probably make sense to enable the same caching on behalf of the group name as it is done with the KeyConfig or was it left out on purpose?

Thank you

### DIFF
--- a/lib/Cache/RuntimeCacheTrait.php
+++ b/lib/Cache/RuntimeCacheTrait.php
@@ -1,12 +1,15 @@
 <?php
 /**
+ * Pimcore
  *
- * Created by PhpStorm.
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
  *
- * @date: 02.02.2022
- *
- * @author: Andreas Wroblewski <andreas.wroblewski@codafish.net>
- * @copyright: codafish><> GmbH, https://codafish.net
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
  */
 
 namespace Pimcore\Cache;
--- a/lib/Cache/RuntimeCacheTrait.php
+++ b/lib/Cache/RuntimeCacheTrait.php
@@ -1,0 +1,83 @@
+<?php
+/**
+ *
+ * Created by PhpStorm.
+ *
+ * @date: 02.02.2022
+ *
+ * @author: Andreas Wroblewski <andreas.wroblewski@codafish.net>
+ * @copyright: codafish><> GmbH, https://codafish.net
+ */
+
+namespace Pimcore\Cache;
+
+use Exception;
+use Pimcore\Cache;
+
+trait RuntimeCacheTrait
+{
+    /**
+     * @var bool
+     */
+    private static bool $cacheEnabled = false;
+
+    /**
+     * @param bool $cacheEnabled
+     */
+    public static function setCacheEnabled(bool $cacheEnabled): void
+    {
+        self::$cacheEnabled = $cacheEnabled;
+    }
+
+    /**
+     * @return bool
+     */
+    public static function getCacheEnabled(): bool
+    {
+        return self::$cacheEnabled;
+    }
+
+    /**
+     * Set cache item for a given cache key
+     *
+     * @param mixed $config
+     * @param string $cacheKey
+     */
+    private static function setCache(mixed $config, string $cacheKey): void
+    {
+        if (self::$cacheEnabled) {
+            Cache\Runtime::set($cacheKey, $config);
+        }
+
+        Cache::save($config, $cacheKey, [], null, 0, true);
+    }
+
+    /**
+     * Remove a cache item for a given cache key
+     *
+     * @param string $cacheKey
+     */
+    private static function removeCache(string $cacheKey): void
+    {
+        Cache::remove($cacheKey);
+        Cache\Runtime::set($cacheKey, null);
+    }
+
+    /**
+     * Get a cache item for a given cache key
+     *
+     * @param string $cacheKey
+     *
+     * @return mixed
+     *
+     * @throws Exception
+     */
+    private static function getCache(string $cacheKey): mixed
+    {
+        if (self::$cacheEnabled && Cache\Runtime::isRegistered($cacheKey) && $config = Cache\Runtime::get($cacheKey)) {
+            return $config;
+        }
+
+        return Cache::load($cacheKey);
+    }
+}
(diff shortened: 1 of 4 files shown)


brusch:
@awroblewski you are right, I think it also would make sense to do the "named" based caching as well for `GroupConfig`.  Could you implement this as well in this PR? Thanks a lot!

models/DataObject/Classificationstore/KeyConfig.php
```diff
@@ -424,16 +396,60 @@ public function setStoreId($storeId)
...
+
+        return Cache::load($cacheKey);
+    }
+
+    /**
+     * Calculate cache key
+     *
      * @param int $id
-     * @param string|null $name
+     * @param string $name
```
blankse, line 383:
```suggestion
     * @param string|null $name
```

awroblewski:
Hello @brusch 

I have implemented caching for the group as well expanded the same onto the collections.

I got one more thing. I have again modified the KeyConfig, as there was an additonal bug I discovered during my tests.

The KeyConfig.php implements another runtime caching (next to the Pimcore Runtime Caching) in the KeyConfig class using a local array which was toggled using the variable $cacheenabled. Whenever you request a KeyConfig using an id and have $cacheenbaled on true the if clause tries to retrieve a value from the array by the given index $id. This fails with an exception when the requested $id was not stored yet. Example:

`if(self::$cacheEnabled && self::$cache[$id]) {...}`

If I request a KeyConfig with id 29 which is not yet stored in the array with cacheenabled = true it yíelds an exception

`"No entry is registered for key '29'"`

I'm not sure why the KeyConfig implements another local array caching in the first place while at the same time using the Runtime Cache. It is practically the same, right?

I have remove this local array caching for now, however, if i miss something here I can put that back in, let me know.

As well I have consolidated the cache getting, setting and removing into static functions for Key-,Group-, and CollectionConfig. This made the implementation more consisten across the three scopes.


Thank You :)

models/DataObject/Classificationstore/KeyConfig.php
awroblewski: @blankse 

thanks for the catch :)

models/DataObject/Classificationstore/CollectionConfig.php
```diff
@@ -262,4 +323,64 @@ public function setStoreId($storeId)
...
+     * @return CollectionConfig|bool
+     */
+    private static function getCache(string $cacheKey): CollectionConfig|bool
+    {
+        if (self::$cacheEnabled && Cache\Runtime::isRegistered($cacheKey) && $config = Cache\Runtime::get($cacheKey)) {
+            return $config;
+        }
+
+        return Cache::load($cacheKey);
+    }
```
brusch: To avoid code duplication and make it easier to maintain, we should put that code into a trait. Since this is very generic, i think we can put this trait here: `lib/Cache/RuntimeCacheTrait.php`. 
As a follow up we could then also migrate existing usages of the runtime cache to this trait.

models/DataObject/Classificationstore/CollectionConfig.php
```diff
@@ -108,6 +131,22 @@ public static function create()
...
+
+    /**
+     * @return bool
+     */
+    public static function getCacheEnabled()
+    {
+        return self::$cacheEnabled;
+    }
+
```
brusch: This should go to the trait as well.

models/DataObject/Classificationstore/CollectionConfig.php
```diff
@@ -25,13 +26,18 @@
...
     /**
-     * Group id.
-     *
-     * @var int
+     * @var bool
+     */
+    public static $cacheEnabled = false;
+
+    /**
+     * @var int|null
```
brusch: Also a candidate for the trait and additionally it could be `private`.

models/DataObject/Classificationstore/CollectionConfig.php
```diff
@@ -247,6 +294,20 @@ public function getCreationDate()
...
+     */
+    public function getRelations()
+    {
+        $list = new CollectionGroupRelation\Listing();
+        $list->setCondition('colId = ' . $this->id);
+        $list = $list->load();
+
+        return $list;
+    }
+
```
brusch: This code is not really related to this PR, so we should put it into another PR targeting `10.x`.

brusch (changes requested):
Thanks a lot, looks already pretty good, however there are just some more small optimizations 😉 Could you update the PR until noon - that would be nice, then we could include it into the `10.2.10` release. Thanks in advance!

models/DataObject/Classificationstore/CollectionConfig.php
awroblewski: Done...

models/DataObject/Classificationstore/CollectionConfig.php
awroblewski: Done...

models/DataObject/Classificationstore/CollectionConfig.php
awroblewski: Done...

models/DataObject/Classificationstore/CollectionConfig.php
awroblewski: You're right I will raise another PR for this. Removed...

lib/Cache/RuntimeCacheTrait.php
```diff
@@ -0,0 +1,83 @@
+<?php
+/**
+ *
+ * Created by PhpStorm.
+ *
+ * @date: 02.02.2022
+ *
+ * @author: Andreas Wroblewski <andreas.wroblewski@codafish.net>
+ * @copyright: codafish><> GmbH, https://codafish.net
+ */
```
brusch:
```suggestion
/**
 * Pimcore
 *
 * This source file is available under two different licenses:
 * - GNU General Public License version 3 (GPLv3)
 * - Pimcore Commercial License (PCL)
 * Full copyright and license information is available in
 * LICENSE.md which is distributed with this source code.
 *
 *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
 *  @license    http://www.pimcore.org/license     GPLv3 and PCL
 */
```

brusch (changes requested):

awroblewski:
@brusch I have refactored the classes. They are now using the Trait in lib/Cache/RuntimeCacheTrait.php.

I was not sure how to handle the Trait signature. Currently i left the standard signature which applies to myself but shouldn't be your Pimcore signature in here? Let me know if I have to change this.

many thx

brusch:
@awroblewski thanks a lot, looks good to me now. The license header should be the one from Pimcore, see code suggestion. Thanks!

awroblewski:
@brusch , just have seen this while writing the comment, all good now

PR #11255, opened 2022-01-22
[Perspective] Omit "not allowed perspective"-warning with perspective= as GET

ferdinand-it:
With the default installation every time reloading the admin UI with the default GET-parameter "&perspective=", a WARNING "User pimcore is not allowed requested perspective ." is logged. But as seen a few lines above, this is not stated as a request for a new perspective. "Perspective for user {user} was not requested." as debug log makes sense here which is logged after this change.

Additionally, fixed wording in a warning log-message.


dvesh3:
@ferdinand-it thanks for your PR. Could you please rebase it on 10.2 (bugfix) branch?

ferdinand-it:
> @ferdinand-it thanks for your PR. Could you please rebase it on 10.2 (bugfix) branch?

Did I do it right to open a new PR for this or would it has been an easier way? Anyways: https://github.com/pimcore/pimcore/pull/11262

dvesh3:
Closed in favor of #11262

PR #11256, opened 2022-01-24
Thumbnail Processor: Avoid output of complete error stacktraces

blankse:
If you execute the ThumbnailsImageCommand with some orphan image assets you get many exceptions with stacktraces. 
  
I think it should be enough to output only the message. So you can easier find the orphan images.


brusch:
Thanks! 👍

PR #11257, opened 2022-01-24
[Doc] Add custom logo migration to Update Notes

blankse:
The custom logo location was changed here: https://github.com/pimcore/pimcore/pull/8980


dvesh3:
@blankse thanks!

PR #11260, opened 2022-01-24
ClassDefinition Import ignores previewGeneratorReference, showFieldLookup and enableGridLocking

blankse:
If you export a class and import it to a other pimcore instance the previewGeneratorReference field is empty


weisswurstkanone:
Thxs !

PR #11261, opened 2022-01-24
[Documents] Refactor page preview generation to queue in Page::save()

### LINKED ISSUE #11091: [Improvement]: Add page preview generation to queue in Page::save()
### Improvement description

Follow-up to: #11023
It's not necessary to call another endpoint for adding the page to the queue. Moving this to Page::save() also has the benefit that previews are updated even when updating the page via SDK.

dvesh3:
Resolves #11091

### DIFF
--- a/models/Document/Page.php
+++ b/models/Document/Page.php
@@ -277,6 +277,8 @@ class Page extends TargetingDocument
 
     public function save()
     {
+        $response = parent::save();
+
         // Dispatch page preview message, if preview is enabled.
         $documentsConfig = \Pimcore\Config::getSystemConfiguration('documents');
         if ($documentsConfig['generate_preview'] ?? false) {
@@ -286,6 +288,6 @@ class Page extends TargetingDocument
             );
         }
 
-        return parent::save();
+        return $response;
     }
 }
--- a/models/Document/Page.php
+++ b/models/Document/Page.php
@@ -15,8 +15,8 @@
 
 namespace Pimcore\Model\Document;
 
+use Pimcore\Messenger\GeneratePagePreviewMessage;
 use Pimcore\Model\Redirect;
-use Pimcore\Model\Site;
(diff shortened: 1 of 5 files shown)


models/Document/Page.php
```diff
@@ -274,4 +274,18 @@ public function getPreviewImageFilesystemPath()
...
+        // Dispatch page preview message, if preview is enabled.
+        $documentsConfig = \Pimcore\Config::getSystemConfiguration('documents');
+        if ($documentsConfig['generate_preview'] ?? false) {
+
+            \Pimcore::getContainer()->get('messenger.bus.pimcore-core')->dispatch(
+                new GeneratePagePreviewMessage($this->getId(), \Pimcore\Tool::getHostUrl())
+            );
+        }
+
+        return parent::save();
```
blankse, line 289: Shouldn't it be sent to the messenger bus after saving? Not that the old page is still generated?

blankse: Shouldn't it be sent to the messenger bus after saving? Not that the old page is still generated?
```suggestion
        parent::save();

        // Dispatch page preview message, if preview is enabled.
        $documentsConfig = \Pimcore\Config::getSystemConfiguration('documents');
        if ($documentsConfig['generate_preview'] ?? false) {
            \Pimcore::getContainer()->get('messenger.bus.pimcore-core')->dispatch(
                new GeneratePagePreviewMessage($this->getId(), \Pimcore\Tool::getHostUrl())
            );
        }

        return $this;
```

jdreesen: What about doing this in an `DocumentEvents::POST_ADD`/`DocumentEvents::POST_UPDATE` event listener instead?

dvesh3: @blankse is right. We should generate preview after calling parent::save() 👍. 

@jdreesen I'd not create a new listener for this task.

PR #11263, opened 2022-01-24
PhpStan: Generate baseline on failure

blankse:
Similar to https://github.com/pimcore/data-hub/pull/463
Creates only the baseline file on failure and uploaded it to the github action artifacts


weisswurstkanone:
Thanks @blankse !

PR #11268, opened 2022-01-25
[DataObject] Link fields accept strings

blankse:
In Pimcore 6.9: If you set an string value to an Link field the value is saved and returned also if you reload the object.

In Pimcore X there is a typehint on the setter method. So it is not possible to set strings. But it can be still in the db.

I add a test and optimize the methods so it can only be Link object or null (from db/user)


blankse:
Ok the typehint is only if generateTypeDeclarations is true. So this problem appears also in Pimcore X. I rebase to 10.2 and add a validation test

PR #11269, opened 2022-01-25
[Docs] Configuration Storage Locations - revalidating existing configurations on prod

### LINKED ISSUE #11011: Improve docs for Configuration Storage locations
### Improvement description

Add more clarity regarding different storage location scenarios. 

For example, when you deploy `symfony-config` configurations from `dev` to `prod` and would like revalidate the existing configs on production.

dvesh3:
Resolves #11011

### DIFF
--- a/doc/Development_Documentation/21_Deployment/03_Configuration_Environments.md
+++ b/doc/Development_Documentation/21_Deployment/03_Configuration_Environments.md
@@ -66,3 +66,10 @@ You can do so by adding the following to your `.env` or just set the env variabl
 ```
 PIMCORE_WRITE_TARGET_CUSTOM_REPORTS=settings-store
 ```
+
+#### Revalidate existing configuration on production
+With `settings-store` target, one can update/change configurations in production environment, which in turn revalidates the generated files e.g. Image Thumbnails, Video thumbnails for subsequent requests.
+
+This is not the case with `symfony-config` write target, as configurations are read-only and deployed from different environment. So we need to explicitly revalidate the generated files either through command or custom script. 
+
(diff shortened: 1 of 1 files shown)


doc/Development_Documentation/21_Deployment/03_Configuration_Environments.md
```diff
@@ -66,3 +66,10 @@ You can do so by adding the following to your `.env` or just set the env variabl
 ```
 PIMCORE_WRITE_TARGET_CUSTOM_REPORTS=settings-store
 ```
+
+#### Revalidate existing configuration on production
+With `settings-store` target, one can update/change configurations in production environment, which in turn revalidates the generated files e.g. Image Thumbnails, Video thumbnails for subsequent requests.
+
+This is not the case with `symfony-config` write target, as configurations are read-only and deployed from different environment. So we need to explicitly revalidate the generated files either through command or custom script. 
+
+For example, to revalidate image or video thumbnails either run command `pimcore:thumbnails:clear` or call `Asset\Image\Thumbnail\Config::clearTempFiles()` after deploying changes on thumbnail configurations.
```
NiklasBr, line 75: This will clear *all* thumbnail images, right? Not just those which are actually updated with a new config? Seems like a more specific command could do here.

dvesh3: With `pimcore:thumbnails:clear` command, you can pass thumbnail name for which you want to revalidate image/video thumbnails.

NiklasBr: Exactly my point

PR #11270, opened 2022-01-25
classification store getter must not load store from parent object

fashxp:
### Prerequisites: 
- class definition with classification store attribute and inheritance activated, see [class_Test_export.zip](https://github.com/pimcore/pimcore/files/7933824/class_Test_export.zip)

- create parent data object with at least one attribute of a group added to classification store field

### Problem: 
try to create child object and add additional classification store group/key

```php
        $child = new Test();
        $child->setKey('c10 ' . time());
        $child->setParentId(24798);
        $child->setPublished(true);

        $store = $child->getStore();
        $store->setLocalizedKeyValue(75, 326, 'foo');
        $store->setActiveGroups($store->getActiveGroups() + [75 => true]);

        $child->save();
```
does not add any classification store data in created object, when loading object again it only contains inherited values. 

### Reason: 
Getter `getStore()` returns classification store of parent object, not the childs object one ... when modifying it, the classification store object of the parent object will be modified, but when saving the child object, the empty child classification store is stored. 

### TODOs: 

- [x] double check
- [x] remove not needed commented code
- [x] create a test for this scenario

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Classificationstore.php
+++ b/models/DataObject/ClassDefinition/Data/Classificationstore.php
@@ -1386,4 +1386,55 @@ class Classificationstore extends Data implements CustomResourcePersistingInterf
 
         return null;
     }
+
+
+    /**
+     * Creates getter code which is used for generation of php file for object classes using this data type
+     *
+     * @param DataObject\ClassDefinition|DataObject\Objectbrick\Definition|DataObject\Fieldcollection\Definition $class
+     *
+     * @return string
+     */
+    public function getGetterCode($class)
+    {
+        $key = $this->getName();
+
+        if ($class->getGenerateTypeDeclarations() && $this instanceof DataObject\ClassDefinition\Data\TypeDeclarationSupportInterface && $this->getReturnTypeDeclaration()) {
+            $typeDeclaration = ': ' . $this->getReturnTypeDeclaration();
+        } else {
+            $typeDeclaration = '';
+        }
+
+        $code = '/**' . "\n";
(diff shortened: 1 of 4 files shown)


models/DataObject/ClassDefinition/Data/Classificationstore.php
```diff
@@ -1386,4 +1386,55 @@ public function denormalize($value, $params = [])
...
+     * @return string
+     */
+    public function getGetterCode($class)
+    {
+        $key = $this->getName();
+
+        if ($class->getGenerateTypeDeclarations() && $this instanceof DataObject\ClassDefinition\Data\TypeDeclarationSupportInterface && $this->getReturnTypeDeclaration()) {
+            $typeDeclaration = ': ' . $this->getReturnTypeDeclaration();
+        } else {
+            $typeDeclaration = '';
```
cipribucur: you can declare typeDeclaration = '', on top and remove the "else".

dvesh3 (commented):
I guess we have to set store back on child to override the store from parent with:

`$child->setTeststore($store);`

maybe change docs for inheritance?

models/DataObject/ClassDefinition/Data/Classificationstore.php
dvesh3: Done 😉

fashxp:
> I guess we have to set store back on child to override the store from parent with:
> 
> `$child->setTeststore($store);`
> 
> maybe change docs for inheritance?

that's not a proper solution as it would break inheritance for existing attributes in the store and also has some other problems.

PR #11286, opened 2022-01-27
Delete unused `DocumentStack` and `DocumentStackListener`

jdreesen:
They were initially introduced to support WebP thumbnail generation (respectively to disable it for certain document types). 
But in #7661 the server-side WebP support detection was removed again - however, said classes were forgotten.


brusch:
@jdreesen Thanks a lot! 
Just 2 small things, we should also remove the service definitions: 
https://github.com/pimcore/pimcore/blob/1892d5f4e24a1013d34f8b6b80c07e3cf9ac4d19/bundles/CoreBundle/Resources/config/event_listeners.yaml#L58
https://github.com/pimcore/pimcore/blob/417b50202ceba7721311eb73f92c308f84236546/bundles/CoreBundle/Resources/config/documents.yaml#L67

jdreesen:
Oh, good catch!
I deleted the service definitions.

brusch:
Nice, thanks a lot!

PR #11296, opened 2022-01-28
[Tests] Git should ignore object resources

blankse:
After execution of codeception this JSONs are all changed. The JSONs are all generated before they are used. So it can be ignored by versioning


brusch:
Thank you very much! 😻

PR #11302, opened 2022-01-30
Fix: "Predefined Document Type"-Select not working

### LINKED ISSUE #11283: [Bug]: "Predefined Document Type"-Select not working correctly in Document Settings
### Expected behavior

When creating a blank Page it should be possible to change the Document Type according to the created predefined Document Types in the SEO & Settings Section of the Document.

image

image

### Actual behavior

When selecting a Document Type in the SEO & Settings Section of the Document it should prefill the predefined Controller and Template, but nothing happens.

image
image

### Steps to reproduce

The behavior can be reproduced on a fresh Pimcore install and on https://demo.pimcore.com/ too.

1. Create a new Document Type in `Settings > Document Types`
2. Create a new blank Page in the Documents-Tree
3. In the `SEO & Settings` Section of the newly created Document, select a predefined Document Type

blankse:
Fixes https://github.com/pimcore/pimcore/issues/11283

The id of predefined document types are no integer anymore

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/settings_abstract.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/settings_abstract.js
@@ -19,6 +19,9 @@ pimcore.document.settings_abstract = Class.create({
     },
 
     setDocumentType: function (field, newValue, oldValue) {
+        if (!newValue.data.id) {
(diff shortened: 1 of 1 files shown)


mcop1:
Thanks again 👍

PR #11303, opened 2022-01-30
Fix: Exception thrown in "Search Assets" window

### LINKED ISSUE #11281: [Bug]: Exception thrown in "Search Assets" window
### Expected behavior

If no search is provided initially and the table is sorted, all assets should be displayed in correct order.

grafik

### Actual behavior

If no search is provided initially and the table is sorted, an exception is thrown.

grafik

Exception:
```
Pimcore\Bundle\AdminBundle\Controller\Searchadmin\SearchController::filterQueryParam(): Argument #1 ($query) must be of type string, null given, called in /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Searchadmin/SearchController.php on line 74

```

### Steps to reproduce

1. Head over to [https://demo.pimcore.fun/admin/?perspective=CMS](https://demo.pimcore.fun/admin/?perspective=CMS)
2. Create a blank page
3. Add an image element
grafik
4. Right click the element > Search
grafik
5. Click directly on a table header, e.g. "Path", to sort the content. Do not type anything in the search field beforehand.
grafik

blankse:
Fix https://github.com/pimcore/pimcore/issues/11281

### DIFF
--- a/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
+++ b/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
@@ -71,7 +71,7 @@ class SearchController extends AdminController
 
         $allParams = $filterPrepareEvent->getArgument('requestParams');
 
-        $query = $this->filterQueryParam($allParams['query']);
+        $query = $this->filterQueryParam($allParams['query'] ?? '');
 
         $types = explode(',', $allParams['type'] ?? '');
         $subtypes = explode(',', $allParams['subtype'] ?? '');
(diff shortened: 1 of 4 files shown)


BabovicT:
@blankse it does fix the part where exception is thrown but now there is another bug. Go to dociment search for example and press the path. Now because type param is not send, after you press it, you see resoults for everything nit just documents.

blankse:
@BabovicT Fix this bug also.

brusch:
@blankse thanks a lot for your PR, however, since it's a bug fix, it should be based on `10.2` branch, right? Could you please rebase? Thanks in advance!

blankse:
@brusch Was created for 10.2 :) Only the branch in PR was wrong. Now it should be correct.

mcop1:
Looks great, thanks!

PR #11313, opened 2022-01-31
Add missing backslash in front of FQCN in the docs

jdreesen:
Then it can be copy/pasted into code that's in a namespace.

See: #11311


dvesh3:
@jdreesen thanks!

PR #11317, opened 2022-01-31
Fix namespaces in the Pimcore 6 upgrade docs

jdreesen:
I know that Pimcore 6 is EOL, but maybe this helps with updates.


brusch:
Docs are fine, even for EOL - thanks @jdreesen 😉

PR #11318, opened 2022-01-31
data object tab reload, maskEl error

### LINKED ISSUE #10975: [Bug] [Backend UI]: Object/Tab reload -> console error
### Expected behavior

The object must be reloaded correctly.

### Actual behavior

The object is reloaded, but with an error in the browser console.

### Steps to reproduce

1. Create the simple plugin, that reloads an Product object, after object was updated:

```javascript

    getClassName: function () {
    },

    initialize: function () {
    },

        if (object.data.general.o_className === 'Product') {
        
            object.reload();
            
            // Also console error
            // let tabpanel = Ext.getCmp("pimcore_panel_tabs");
            // let activeTab = tabpanel.getActiveTab();
            // if (activeTab) {
            //     if (activeTab.initialConfig.object) {
            //         activeTab.initialConfig.object.reload();
            //     }
            // }
        }
    },
});

```

2. Save the object.
3. Reload works, but we'll get an error in browser console:

Bildschirmfoto 2021-12-07 um 15 31 59

kingjia90:
fixes #10975

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js
@@ -872,6 +872,7 @@ pimcore.object.object = Class.create(pimcore.object.abstract, {
                 tabIndex: currentTabIndex,
                 uiState: uiState
             };
+            this.tab = null;
 
             window.setTimeout(function (id) {
                 pimcore.helpers.openObject(id, "object", options);


kingjia90:
The error is caused by this [unmask ](https://github.com/pimcore/pimcore/blob/cd7dd3f74a78f6376e3e9e7e14272a93136684eb/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js#L822). The reason seems to be that ["postSaveObject"](https://github.com/pimcore/pimcore/blob/cd7dd3f74a78f6376e3e9e7e14272a93136684eb/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js#L797) is just fired earlier than that (which normally should be fine) but since OP wants to reload as soon as you save via postSaveObject, the reloading function simply [closes ](https://github.com/pimcore/pimcore/blob/cd7dd3f74a78f6376e3e9e7e14272a93136684eb/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js#L882) the tab and reopens it after 500ms delay, making the unmask being called in a moment that no (masked) tab exists (closed and the new is scheduled). 

The workaround suggested by SamyMP to add 1 second delay helps the reload function  to be fired after the unmask, and this also explains why it seems working without throwing errors.

There were some possible solutions, like moving the postSaveObject after the unmask or change the ext-all with a if null check, nullify this.tab on "destroy", but i feel this could be an easy fix (as the commit) for this specific case without altering the usual behaviour.
Open to suggestions

brusch:
I think this is a good approach! 👌

MarcW09:
This related commit leads to "Object.reload" not working anymore for me. When removing the Commit it works again, but with the maskEl error.
Screenshot_09_02_22__14_58

brusch:
@MarcW09 can you please open a new ticket for that? thanks a lot!

brusch:
@MarcW09 ah sorry, there was already an issue, so @kingjia90 reopened it, all good 😉

PR #11319, opened 2022-02-01
Store thumbnails for each asset in folder -> prevent looping through all thumbnails for assets in same folder

BlackbitDevs:
When you update an asset file (also when you update a new one), all thumbnails of this asset get deleted, see
https://github.com/pimcore/pimcore/blob/6d6774a139e7cadd3ed446ef300c50523838a4c8/models/Asset/Image.php#L52

This calls https://github.com/pimcore/pimcore/blob/6d6774a139e7cadd3ed446ef300c50523838a4c8/models/Asset.php#L1987-L2006
The folder structure for thumbnail storage is currently:
`<asset storage>/<path to asset>/image-thumb__<asset id>__<thumbnail name>/<asset filename>`
To find all thumbnails which belong to the just saved asset, above function needs to loop through all the thumbnail files in `<asset storage>/<path to asset>/` and filter for those whose path contains `image-thumb__<asset id>__`.

Depending on the asset folder structure and storage type this may need a lot of time. For example when you have 50000 images in one asset folder and you have 20 thumbnail definitions (not regarding the automatically or dynamically created thumbnail definitions, e.g. tree preview or custom size download), this folder will contain up to 1M files. 
In addition when the thumbnails are not stored locally but on a cloud storage provider, the problem gets even worse.
In our case uploading one asset to such a folder needed several minutes to complete.

This PR changes the folder structure for storing thumbnails to `<asset storage>/<path to asset>/image-thumb__<asset id>/<thumbnail name>/<asset filename>`. This way in case of updating an asset file, we only need to delete the directory `<asset storage>/<path to asset>/image-thumb__<asset id>`. No loop necessary anymore.


dpfaffenbauer:
Love this, but it is a BC break and should go into 11.

fashxp:
maybe provide a config option to change folder structure and keep old structure as bc layer?

dpfaffenbauer:
@fashxp that works

BlackbitDevs:
@fashxp How do you imagine this config option? What is the initial value? Because as soon as one has executed the migration the files are moved. If we do not move but copy and keep both thumbnail locations in the migration (so that you can later decide whether to use the new or old directory structure), it will duplicate the necessary hard disk space for thumbnails.
Links are neither possible because not all sotrage providers support links.

dpfaffenbauer:
Don't add a migration, create a Command that can be manually executed

BlackbitDevs:
Would really like to get to know @brusch's opinion about this. Imho maintaining old and new thumbnail directory structure will complicate future development in an unnecessary way. Why should one not apply the new folder structure?

brusch:
I agree with @dpfaffenbauer that this would be a breaking change, the most critical thing is that it would require to change the config of the webserver (see `.htaccess` changes). 
I also agree with @BlackbitNeueMedien that it doesn't make sense to maintain both approaches, because as we can see in this PR, quite a lot of things need to be considered and it doesn't make sense to duplicate all the stuff. 

If we would change the directory to the following pattern, it should be still compatible with the webserver config imho: 
`<thumbnail storage>/<path to asset>/<asset id>/image-thumb__<asset id>__<thumbnail name>/<asset filename>` while we can still implement the same optimization. 

With that structure in mind, I could also imagine that it would be much easier to make it configurable whether the old or the new format should be used (incl. a migration command), without duplicating stuff or making it more complex. 

WDYT?

BlackbitDevs:
Brillant idea, @brusch! I changed the code.
Only with the migration I am not sure what to do. Shall I simply put it to a command? Then we should add a recommendation to the docs I think - otherwise nobody will migrate to the new folder structure. Or do you mean something else?

brusch:
@BlackbitNeueMedien can you please check our friend Stan? 😊

bundles/CoreBundle/Resources/config/routing.yml
```diff
@@ -15,7 +15,7 @@ _pimcore_service_common_files_apple_touch_icon:
         filename: 'apple\-touch\-icon.*'
 
 _pimcore_service_thumbnail:
-    path: '{prefix}image-thumb__{assetId}__{thumbnailName}/{filename}'
+    path: '{prefix}{assetId}/image-thumb__{assetId}__{thumbnailName}/{filename}'
```
brusch: This is not necessary, right? It's already covered by `{prefix}` and `{assetId}` later in the pattern.

BlackbitDevs:
@brusch Is fixed, regex had syntax error.

brusch (changes requested):
Looks already quite good to me. 
The only important open question to me now, is whether it's reasonable to do this change in a minor release. It's not a breaking change in a common sense anymore, but still it could have an impact on the system stability (even if we provide a migration command). 

Any thoughts on that? @BlackbitNeueMedien @dpfaffenbauer @NiklasBr @blankse @jdreesen

bundles/CoreBundle/Resources/config/routing.yml
BlackbitDevs: Correct, reverted this.

BlackbitDevs:
I see one BC problem currently: When someone has external applications which depend on the old URL structure. But for this we could add a check if the thumbnail file exists in the new folder structure in https://github.com/pimcore/pimcore/blob/4cd9301b1f40b9e67ac73c3e49fcaeaab9f33fb7/bundles/CoreBundle/Controller/PublicServicesController.php#L48

Edit: I think this is not necessary as the old URLs still work.

dpfaffenbauer:
@brusch 👍

brusch:
@BlackbitNeueMedien agree. So the only thing missing is to transform the migration to a command - I think then we're good to go 👍

blankse:
@brusch 👍🏼

NiklasBr:
Generally 👍 from me

> So the only thing missing is to transform the migration to a command

Since this is kind of a large change maybe add an "empty" migration which does nothing other prints a warning to run the command?

BlackbitDevs:
Ok, I added `pimcore:thumbnails:migrate-folder-structure` command.
@NiklasBr I added a note to the upgrade notes - because nothing breaks if you do not execute the command. The thumbnails will just be created again in the new folder structure.

brusch:
Thanks guys for your feedback, I really appreciate it! ♥
@BlackbitNeueMedien great, we'll review and further test as soon as possible!

NiklasBr:
> because nothing breaks if you do not execute the command

I don't fully agree here, there will be breakage, it takes many weeks for the many hundreds of thousand thumbnails to be re-created, so customers will complain that they disappeared and visitors will have a bad experience due to slower page load.

BlackbitDevs:
Ok, but who does a Pimcore update without reading upgrade notes :/
But anyway, what do you think, @brusch? Do we need an empty migration which just writes something like `Please run bin/console pimcore:thumbnails:migrate-folder-structure after the update`? Imho people will wonder why they have to do that themselves and why the migration does not already do it (but they will actually anyway when they read the upgrade notes ;)

brusch:
I'm afraid that an automatic migration could take quite some time, especially in case of multiple million or even billion thumbnails. 
A possible solution could be the recommendation to grab the migration command and deploy and run that in advance of the upgrade. 

We could even think of further optimizing the command, so that it provides additional functionality for filesystems supporting symlinks or hardlinks. The command could then create the links first and then after the upgrade you could run it again with a flag that removes the links so that you have the final new structure.

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,14 +1,13 @@
 # Upgrade Notes
 ## 10.3.0
-- **Important notice**: [Symfony Messenger] Pimcore Core & Maintenance messages are now routed to different queues instead of default. It is
+- **Important**: [Symfony Messenger] Pimcore Core & Maintenance messages are now routed to different queues instead of default. It is
   required to run command `bin/console messenger:consume pimcore_core pimcore_maintenance` before the upgrade, so that
   the messages on default queue gets consumed.
-- [Documents] Introduced additional interfaces for editable methods `getDataEditmode()`, `rewriteIds()` & `load()`. Existing `method_exists` calls are deprecated and will be removed in Pimcore 11.
+- **Important**: The folder structure for storing thumbnails changed, please run `bin/console pimcore:thumbnails:migrate-folder-structure` after the update to copy existing thumbnails to new folder structure
```
brusch: This would be a topic for `10.4.0` then 😉

brusch (changes requested):

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
BlackbitDevs: Sorry, fixed.

BlackbitDevs:
> A possible solution could be the recommendation to grab the migration command and deploy and run that in advance of the upgrade.

Seems too complicated imho. I always wonder when I do an update for old Pimcore systems and such external files are referenced in the upgrade notes, if they still work. Then I would prefer a migration which outputs a message to execute the command afterwards.

Concerning the links: With some apache configurations links are not followed. I always have the problem for example with the official Pimcore docker demo, that I have to execute `assets:install` with hard copies, otherwise the JS files are not accessible.

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,14 +1,15 @@
 # Upgrade Notes
+## 10.4.0
+- **Important**: The folder structure for storing thumbnails changed, please run `bin/console pimcore:thumbnails:migrate-folder-structure` after the update to copy existing thumbnails to new folder structure
```
```suggestion
- **Important**: The folder structure for storing thumbnails changed, please run `bin/console pimcore:thumbnails:migrate-folder-structure` after the update to copy existing thumbnails to new folder structure. If you're dealing with a huge amount of thumbnails you should consider that this change might increase the load on your system as well as page-loading times during the migration command is executed, as non-existing thumbnails are then generated on demand. 
```

brusch:
Ok for me - I've added some more info in the upgrade notes, I think that should be fine then.

brusch:
@BlackbitNeueMedien jfyi pushed c31eed6 with a fix for dynamic generation which was broken

brusch:
@BlackbitNeueMedien  One more thing: 
`\Pimcore\Model\Asset::relocateThumbnails()` also needs to be changed / optimized. 
Thanks in advance!

BlackbitDevs:
Ok, currently looking at it. Does this actually work currently? Because in 
https://github.com/pimcore/pimcore/blob/2322eebf98296fc00e21f0ca0054947c1328e184/models/Asset.php#L2051
the thumbnail directory for the asset gets processed in shallow mode - but there cannot be a file (`$attributes->isFile()`) whose path contains `image-thumb_` - or is this possible?
I thought the things with `image-thumb_` in their path are always folders?

brusch:
@BlackbitNeueMedien yes, that works because it's checking the full path of the file, not the filename: 
`strstr($attributes['path'], 'image-thumb_')` 😉

so `attributes['path']` is e.g. `Sample/362/image-thumb__362__galleryThumbnail/Tavi.jpg` , which is a file and matches with `image-thumb_` in the path 😊

Thanks for rechecking

brusch:
@BlackbitNeueMedien 🏓 just a quick reminder - not that much left anymore, so would be cool if we could finish that 😊Thanks!

BlackbitDevs:
> yes, that works because it's checking the full path of the file, not the filename:
strstr($attributes['path'], 'image-thumb_')

Still struggle with this because `$storage->listContents($oldParent)` is currently only executed in *shallow* mode. It will never find anything which is a file and which has a path containing `image-thumb_`

And if it would delete the thumbnails (when `listContents` would use *deep* mode) in the old path directory, then there were no thumbnail files left to be moved for https://github.com/pimcore/pimcore/blob/e57752751ecda8aa40190344ee4d85850ebc0ea6/models/Asset.php#L2065-L2074

BlackbitDevs:
Ok, have adjusted `Asset::relocateThumbnails()`. WDYT?

models/Asset.php
```diff
@@ -2036,20 +2025,13 @@ private function updateChildPaths(FilesystemOperator $storage, string $oldPath)
...
-            foreach ($contents as $item) {
-                $storage->delete($item['path']);
-            }
-
             //remove destination parent folder thumbnails
-            $contents = $storage->listContents($newParent)->filter(fn (StorageAttributes $attributes) => ($attributes->isFile() && strstr($attributes['path'], 'image-thumb_')));
+            $contents = $storage->listContents($newParent)->filter(fn (StorageAttributes $attributes) => str_contains($attributes['path'], 'image-thumb_'));
             /** @var StorageAttributes $item */
             foreach ($contents as $item) {
                 $storage->delete($item['path']);
```
brusch: I'm afraid that this is still not entirely right 🤔
What we need to do is, deleting all the folder previews (generated [here](https://github.com/pimcore/pimcore/blob/486a9f9f3611c7159b89554d0ab8bf900a79e673/models/Asset/Folder.php#L115)) of the parent folders of the source and the target folder. 

So I think we could implement a `private` helper function, where we pass an instance of `Asset/Folder` and then this function goes up the tree on by one and deletes the preview images. 

Example: 
```php
    private function clearFolderThumbnailsOfParents(Asset $asset): void
    {
        while ($asset = $asset->getParent()) {
            $asset->clearThumbnails(true);
        }
    }
```

brusch (changes requested):

models/Asset.php
BlackbitDevs: Ah you mean because the just moved thumbnails may be inside the source folder's preview? Now I understand what this is good for ;-)

~But then we only need to delete files containing `-folder-preview` right?~
As folders do only have this one folder preview thumbnail, we can use `$assetFolder->clearThumbnails(true);` - as you proposed

BlackbitDevs: Ok, changed it. What do you think now?

models/Asset.php
```diff
@@ -2036,39 +2026,34 @@ private function updateChildPaths(FilesystemOperator $storage, string $oldPath)
...
                         $storage->move($item->path(), $replacePath);
                     }
                 }
             }
 
-            //required in case if renaming or moving parent folder
+            //required in case of renaming or moving parent folder
             try {
-                $storage->move($oldPath, $this->getRealFullPath());
+                $storage->move($oldPath, $newParent);
```
```suggestion
                $storage->move($oldThumbnailsPath, $newThumbnailsPath);
```

models/Asset.php
```diff
@@ -2036,39 +2026,34 @@ private function updateChildPaths(FilesystemOperator $storage, string $oldPath)
      */
     private function relocateThumbnails(string $oldPath)
     {
-        $oldParent = dirname($oldPath);
-        $newParent = dirname($this->getRealFullPath());
+        $oldParent = dirname($oldPath).'/'.$this->getId();
+        $newParent = dirname($this->getRealFullPath()).'/'.$this->getId();
```
```suggestion
        $oldThumbnailsPath= $oldPath . '/' . $this->getId();
        $newThumbnailsPath= $this->getRealFullPath() . '/' . $this->getId();
```

models/Asset.php
```diff
@@ -2036,39 +2026,34 @@ private function updateChildPaths(FilesystemOperator $storage, string $oldPath)
...
         $storage = Storage::get('thumbnail');
 
         try {
-            //remove source parent folder thumbnails
-            $contents = $storage->listContents($oldParent)->filter(fn (StorageAttributes $attributes) => ($attributes->isFile() && strstr($attributes['path'], 'image-thumb_')));
-            /** @var StorageAttributes $item */
-            foreach ($contents as $item) {
-                $storage->delete($item['path']);
+            //remove source parent folder preview thumbnails
+            $sourceFolder = Asset::getByPath($oldPath);
```
```suggestion
            $sourceFolder = Asset::getByPath(dirname($oldPath));
```
We need the parent folder of the current here, because `$oldPath` is not available anymore (it's `$this`, so just got moved 😉).

bundles/CoreBundle/Command/ThumbnailsMigrateFolderStructureCommand.php
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * @internal
+ */
+class ThumbnailsMigrateFolderStructureCommand extends AbstractCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('pimcore:thumbnails:migrate-folder-structure')
```
brusch:
```suggestion
            ->setName('pimcore:migrate:thumbnails-folder-structure')
```
Maybe we should cluster all individual migration commands in `pimcore:migrate:` for better structure? 
If ok, please also change the FQCN to `Pimcore\Bundle\CoreBundle\Command\Migrate\ThumbnailsFolderStructureCommand` and also the docs. 😉

brusch (changes requested):
We're nearly there, plz review my comments 😉

BlackbitDevs:
Ok, have applied the changes. Do not know what the tests want to say to me...

bundles/CoreBundle/Command/ThumbnailsMigrateFolderStructureCommand.php
BlackbitDevs: Applied this and also moved `pimcore:migrate:storage` to `Pimcore\Bundle\CoreBundle\Command\Migrate` namespace.

bundles/CoreBundle/Command/Migrate/StorageCommand.php
```diff
@@ -26,7 +26,7 @@
 /**
  * @internal
  */
-class MigrateStorageCommand extends AbstractCommand
+class StorageCommand extends AbstractCommand
```
brusch, line 29: This command is defined here: 
https://github.com/pimcore/pimcore/blob/417b50202ceba7721311eb73f92c308f84236546/bundles/CoreBundle/Resources/config/commands.yaml#L76

So this needs to be changes as well when changing the FQCN😉

BlackbitDevs: Done.

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,14 +1,15 @@
 # Upgrade Notes
+## 10.4.0
+- **Important**: The folder structure for storing thumbnails changed, please run `bin/console pimcore:thumbnails:migrate-folder-structure` after the update to copy existing thumbnails to new folder structure. If you're dealing with a huge amount of thumbnails you should consider that this change might increase the load on your system as well as page-loading times during the migration command is executed, as non-existing thumbnails are then generated on demand.
```
```suggestion
- **Important**: The folder structure for storing thumbnails changed, please run `bin/console pimcore:migrate:thumbnails-folder-structure` after the update to copy existing thumbnails to new folder structure. If you're dealing with a huge amount of thumbnails you should consider that this change might increase the load on your system as well as page-loading times during the migration command is executed, as non-existing thumbnails are then generated on demand. 
```

brusch (changes requested):

fashxp:
did somebody ever test the migration command, doesn't seem so. have the following issues:
- [x] did not start migration at all with error `cannot traverse an already closed generator`
- [x] filter for list contents didn't work due to wrong directory handling and wrong regex
- [x] video thumbnails weren't considered
- [x] asset_cache storage was not considered
- [x] deletion of old files did not always work, since it also could be a file

tried to fix via https://github.com/BlackbitNeueMedien/pimcore/commit/92b83b455d1404868f1f52258946c5ce7347dc1c

fashxp:
further problems I found: 
- [x] folder previews weren't cleared - fix see https://github.com/BlackbitNeueMedien/pimcore/commit/e1fd7cad5e80c277dd0fbdd9a43f98a367820abc
- [x] did not clear folder thumbnails when deleting asset in folder (might be there before too) - fix see https://github.com/BlackbitNeueMedien/pimcore/commit/d95f58525290bea62b6f4693e7ebd573e1aba78e
- [x] relocate thumbnails didn't work for moving assets from one folder into another - fix see https://github.com/BlackbitNeueMedien/pimcore/commit/2f500188723390f5e5bfbad60b570ab80400c99c

fashxp:
@BlackbitNeueMedien  could you please double check my adaptions, thx.

fashxp (approved):

bundles/CoreBundle/Command/Migrate/StorageCommand.php
```diff
@@ -15,81 +15,104 @@
...
-
-        $output->writeln('Migrating asset cache (document previews, etc.) ...');
-        $this->doMigrateStorage($output, $assetCacheStorage);
-        $output->writeln("\nSuccessfully moved asset cache files to new folder structure");
-
-        return 0;
-    }
+            try {
+                $sourceStorage = $this->locator->get($storageSourceName);
+                $targetStorage = $this->locator->get($storageTargetName);
```
BlackbitDevs: @fashxp This means that you have to configure `pimcore.thumbnail.storage.source` and `target`, otherwise you cannot profit of the new directory structure. Is this really necessary or should we fall back to same storage location? Because usually those services will not have been set if the system never siwtched thumbnail storage.

brusch: @BlackbitNeueMedien No, look at the command name, it's about the existing storage migration command, see https://pimcore.com/docs/pimcore/current/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/File_Storage_Setup.html#page_Storage-Migration not about the thumbnail structure 😉
The Thumbnail structure can be migrated using the `ThumbnailsFolderStructureCommand`.

BlackbitDevs: Ah sorry, thanks for explanation.
Rest of @fashxp's changes look good to me.

PR #11324, opened 2022-02-01
Fix pagination issue #11322

### LINKED ISSUE #11322: [Bug]: Pagination with ES index display always 10 items on first page
### Expected behavior

Get the items limited to the given pageLimit from paginationAdapter in Pimcore\Model\Paginator\EventSubscriber\PaginateListingSubscriber:
https://github.com/pimcore/pimcore/blob/ece35aa05d1ac1d067b928d6452b2688fe687203/lib/Model/Paginator/EventSubscriber/PaginateListingSubscriber.php#L28-L33
reference to
https://github.com/pimcore/pimcore/blob/ece35aa05d1ac1d067b928d6452b2688fe687203/bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php#L1383-L1389

### Actual behavior

Due to the previous call (in line 29)
https://github.com/pimcore/pimcore/blob/ece35aa05d1ac1d067b928d6452b2688fe687203/lib/Model/Paginator/EventSubscriber/PaginateListingSubscriber.php#L28-L33
the productList is already loaded with the ES default limit of 10 when calling the getProducts method the second time (in line 30 above)
therefore the list is not reloaded with the right limit due to the fact the condition in line 218 is now false.
https://github.com/pimcore/pimcore/blob/ece35aa05d1ac1d067b928d6452b2688fe687203/bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php#L216-L223
From the second page the productList is reseted at line 451 in the setOffset method
https://github.com/pimcore/pimcore/blob/ece35aa05d1ac1d067b928d6452b2688fe687203/bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php#L449-L455

### Steps to reproduce

- Use elastic search adapter by according e-commerce framework setting
- Set a limit by a e-commerce filter object different from 10
- Run the controller code from pimcore demo
- First page will always have 10 items. From page 2 the item count will be like filterConfig

woellundwoell:
Resolves #11322

### DIFF
--- a/lib/Model/Paginator/EventSubscriber/PaginateListingSubscriber.php
+++ b/lib/Model/Paginator/EventSubscriber/PaginateListingSubscriber.php
@@ -26,8 +26,8 @@ class PaginateListingSubscriber implements EventSubscriberInterface
         $paginationAdapter = $event->target;
 
         if ($paginationAdapter instanceof PaginateListingInterface) {
-            $event->count = $paginationAdapter->count();
             $items = $paginationAdapter->getItems($event->getOffset(), $event->getLimit());
+            $event->count = $paginationAdapter->count();
             $event->items = $items;
             $event->stopPropagation();
         }


woellundwoell:
This PR is a "quick fix". 
The long term solution might be to make a hash over the filter parameters and then check for change on load the product list?

brusch:
Since this is a bugfix I've 🍒-picked it into `10.2` with 4013e6c6df912c4a060d53e1a854b8c2b743289d. 
Btw. ... nice commit message 😂

brusch:
Thanks!

woellundwoell:
Thanks a lot for the fast processing! ;)

Issue #11323, opened 2022-02-01
[Bug]: create method for classification store classes KeyGroupRelation and CollectionGroupRelation yield PDOException


awroblewski:
### Expected behavior

When issuing either

```
$colRel = CollectionGroupRelation::create()
$groupRel = KeyGroupRelation::create()

```
$colRel is a new object of type CollectionGroupRelation
$groupRel is a new object of type KeyGroupRelation

### Actual behavior

When issuing either

```
$colRel = CollectionGroupRelation::create()
$groupRel = KeyGroupRelation::create()

```
a PDOException is thrown.

`PDOException: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'keyId' cannot be null`

### Steps to reproduce

This happens as the create method instatiates the class and subsequently tries to save the object to the database.

```
    /**
     * @return Model\DataObject\Classificationstore\KeyGroupRelation
     */
    public static function create()
    {
        $config = new self();
        $config->save();

        return $config;
    }

```

However, the primary key for each entry consists of combined groupId and keyId for the KeyGroupRelation and combined colId and groupId for the CollectionGroupRelation.

When the object is instantiated using the static create method and subsequently saved, both entries are NULL which causes the exception.

The create method is used for the Key-,Group- and CollectionConfig as well, however, these entries have their own primary key "id" which is set as part of the create method using

```
$this->db->insert(self::TABLE_NAME_KEYS, []);
$this->model->setId($this->db->lastInsertId());

```
This, however, did not apply for the two classes KeyGroupRelation and CollectionGroupRelation.

To resolve this we coud either remove the create method for the two classes KeyGroupRelation and CollectionGroupRelation or, to keep it consistent with the implementation of the Key-, Group- and CollectionConfig classes, we just return the instatiated object. I would go for the latter and stay consistent.

awroblewski:
New PR Raised

https://github.com/pimcore/pimcore/pull/11326

PR #11325, opened 2022-02-01
PR for issue #11323

awroblewski:
Fix for issue #11323 create method for classification store classes KeyGroupRelation and CollectionGroupRelation yield PDOException


awroblewski:
wrong branch

PR #11326, opened 2022-02-01
fixing bug for create method on classification store group and collec…

awroblewski:
Fix as described in issue #11323 create method for classification store classes KeyGroupRelation and CollectionGroupRelation yield PDOException

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
@@ -380,3 +380,6 @@ framework:
 - Removed deprecated `marshal()` and `unmarshal()` methods from object data-types.
 - `DynamicTextLabelInterface::renderLayoutText()` must handle nullable object param.
 - [AdminBundle] Marked classes and controllers as @internal/final - please see all changes here: https://github.com/pimcore/pimcore/pull/8453/files & https://github.com/pimcore/pimcore/pull/8988/files
+
+## 10.3.0
+- Method `create()` from `Pimcore\Model\DataObject\Classificationstore\CollectionGroupRelation` and `Pimcore\Model\DataObject\Classificationstore\KeyGroupRelation` are no longer save the new object but just returns the instance. See all related changes here: https://github.com/pimcore/pimcore/pull/11326/files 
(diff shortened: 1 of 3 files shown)


brusch:
Since this changes the behavior a bit, we should mention this in the upgrade notes. 
I can't think of any alternative approach how to solve this, so imho it's fine to do that in a bug fix release but we should mention it. 
Could you quickly update the PR accordingly? Thanks!

awroblewski:
@brusch
I'm happy to do that, but don't know exactly what to do next. How do i modify the PR accordingly to mention this change in the upgrade notes, can you tell?

many thx

brusch:
@awroblewski sure, 2 things, as we have a behavior change I'd not include it in the bug-fix release, but in our next minor release (`10.3.0` which will be released today, so it's a bit urgent if it should be included 😉). To do so you need to rebase and change the target of your PR to `10.x`. 

Then add a note about the behavior change here: 
https://github.com/pimcore/pimcore/blob/bfe097fa3e9120b695394bb73698e815edff1375/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md#L2

Thanks!

awroblewski:
@brusch I have changed to 10.x and modified the upgrade notes. Is the statement in the upgrade notes ok?

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -380,3 +380,6 @@ framework:
 - Removed deprecated `marshal()` and `unmarshal()` methods from object data-types.
 - `DynamicTextLabelInterface::renderLayoutText()` must handle nullable object param.
 - [AdminBundle] Marked classes and controllers as @internal/final - please see all changes here: https://github.com/pimcore/pimcore/pull/8453/files & https://github.com/pimcore/pimcore/pull/8988/files
+
+## 10.3.0
+- Method `create()` from `Pimcore\Model\DataObject\Classificationstore\CollectionGroupRelation` and `Pimcore\Model\DataObject\Classificationstore\KeyGroupRelation` are no longer save the new object but just returns the instance. See all related changes here: https://github.com/pimcore/pimcore/pull/11326/files
```
```suggestion
- Method `create()` from `Pimcore\Model\DataObject\Classificationstore\CollectionGroupRelation` and `Pimcore\Model\DataObject\Classificationstore\KeyGroupRelation` no longer saves the new object, but just returns the instance. Related changes can be found here: https://github.com/pimcore/pimcore/pull/11326/files 
```

brusch:
Thank you very much!

PR #11331, opened 2022-02-01
check for classes property

solverat:
image

While using the https://github.com/pimcore/perspective-editor bundle, the `classes` property is not available in configuration which will lead to the exception above.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -152,7 +152,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
             if ($request->get('view')) {
                 $cv = Element\Service::getCustomViewById($request->get('view'));
 
-                if (isset($cv['classes']) && is_array($cv['classes'])) {
+                 if (!empty($cv['classes'])) {
                     $cvConditions = [];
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -152,7 +152,7 @@ public function treeGetChildsByIdAction(Request $request, EventDispatcherInterfa
             if ($request->get('view')) {
                 $cv = Element\Service::getCustomViewById($request->get('view'));
 
-                if ($cv['classes']) {
+                if (isset($cv['classes']) && is_array($cv['classes'])) {
```
blankse:
```suggestion
                if (!empty($cv['classes'])) {
```
otherwise an empty array would now only filter folders

solverat: puah. that's a good catch! thank you.

solverat: Three commits to fix this. Heavy duty :D

brusch:
Squashed and 🍒-picked into `10.2` as it is a bug fix 😉 see 3518f2a7a60d78d0f177daca424c07e9be2d7e34

PR #11332, opened 2022-02-01
[Feature]: Possibility to change the storage path of configs via env var

### LINKED ISSUE #11229: [Feature]: Possibility to change the storage path of configs via env var
### Feature description

E.g. https://github.com/pimcore/pimcore/blob/03877bf9a6608ff161b0ca65ed5d2f40fc90f438/models/Asset/Image/Thumbnail/Config/Dao.php#L38

It would be nice if you could change the storage path to a different location, so that you can for instance put the storage path to `/config/image-thumbnails` in the `dev` environment to write `YAML` files to the `config/` directory.

mcop1:
resolves #11229

### DIFF
--- a/lib/Kernel.php
+++ b/lib/Kernel.php
@@ -167,9 +167,19 @@ abstract class Kernel extends SymfonyKernel
             $loader->load($systemConfigFile);
         }
 
-        foreach (['image-thumbnails', 'video-thumbnails', 'custom-reports', 'document-types', 'web-to-print', 'predefined-properties', 'predefined-asset-metadata', 'staticroutes', 'perspectives', 'custom-views'] as $configDir) {
+        foreach (['video-thumbnails', 'custom-reports', 'document-types', 'web-to-print', 'predefined-properties', 'predefined-asset-metadata', 'staticroutes', 'perspectives', 'custom-views'] as $configDir) {
             $configDir = PIMCORE_CONFIGURATION_DIRECTORY . "/$configDir/";
             if (is_dir($configDir)) {
+                // @phpstan-ignore-next-line
+                $loader->import($configDir);
+            }
+        }
+
+        $configArray = ['\Pimcore\Model\Asset\Image\Thumbnail\Config'];
+        foreach($configArray as $class) {
+            $config = new $class;
+            $configDir = $config->getDao()->getStorageDirectory();
+            if(is_dir($configDir)) {
                 // @phpstan-ignore-next-line
                 $loader->import($configDir);
             }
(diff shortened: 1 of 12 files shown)


lib/Kernel.php
```diff
@@ -167,13 +167,34 @@ public function registerContainerConfiguration(LoaderInterface $loader)
...
         }
+
+        $configArray = [
+            ['storageDirectoryEnvVariableName' => 'PIMCORE_CONFIG_STORAGE_DIR_IMAGE_THUMBNAILS',
+                'defaultStorageDirectoryName' => 'image-thumbnails']
+        ];
+
+        foreach($configArray as $config) {
+            $configDir = $config['storageDirectoryEnvVariableName'];
+            $configDir = $_ENV[$configDir] ?? null;
```
jdreesen: Is `$_ENV` enough here? Or should we check `$_SERVER` too?

brusch: `$_ENV` should be fine, because of: 
https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Dotenv/Dotenv.php#L151

blankse: @brusch This line set the `$_ENV` values to `$_SERVER` not the other way around

blankse:
```suggestion
            $configDir = $_SERVER[$configDir] ?? null;
```

brusch: Yes, and therefore `$_SERVER` should work as well, right?

blankse: Yes, I think we should use the $_SERVER var

brusch: For what reason? We're using `$_ENV` already in other places as well ... 🤔

jdreesen: I wouldn't rely on Symfony DotEnv, since it's not necessarily used, especially in prod the variables are probably defined otherwise.

> We're using $_ENV already in other places as well ...

Not really. You're mostly using `$_SERVER`, sometimes with a fallback to `$_ENV`. 
There's only two places where only `$_ENV` is used: 
- https://github.com/pimcore/pimcore/blob/10.x/lib/Config/LocationAwareConfigRepository.php#L203
- https://github.com/pimcore/pimcore/blob/10.x/bundles/InstallBundle/Installer.php#L382-L383

brusch: Ok, I'm fine with `$_SERVER` as well, but we should then change it in the other places where `$_ENV` is used as well. Can you provide a PR for that? 
@mcop1 Please consider 😉

mcop1: Changed it to $_SERVER

PR #11333, opened 2022-02-01
[Element] Fix permission issue for listing children

roland4432:

Changes to order by method of the 3 DAOs for docs, assets and objects to ensure that the result will be correct if a user has multiple roles / acl settings.

## Additional info  
Recreate the bug:

* create an element (in any of the trees)
* create 2 children with 1 being named longer than the other one
* create 2 roles and set them up, that one mustn't list the children with the longer name and at least list rights for the shorter named child
* setup a user and give it both roles we just created
* login as this user and check, that the user is not having a `+` icon next to the level 1 entry in the tree, since the longer named path won't return a 1 for list and the other element is not being considered due to the limit 1 part of the sql query

=> solution: 
rather sort by field `list` desc instead of length(cpath) to ensure, that entries with `list = 1` will be on top

### DIFF
--- a/models/Asset/Dao.php
+++ b/models/Asset/Dao.php
@@ -332,7 +332,7 @@ class Dao extends Model\Element\Dao
             $userIds = $user->getRoles();
             $userIds[] = $user->getId();
 
-            $query .= ' AND (select `list` as locate from `users_workspaces_asset` where `userId` in (' . implode(',', $userIds) . ') and LOCATE(cpath,CONCAT(a.path,a.filename))=1  ORDER BY LENGTH(cpath) DESC LIMIT 1)=1';
+            $query .= ' AND (select `list` as locate from `users_workspaces_asset` where `userId` in (' . implode(',', $userIds) . ') and LOCATE(cpath,CONCAT(a.path,a.filename))=1  ORDER BY list DESC LIMIT 1)=1';
         }
         $query .= ' LIMIT 1;';
         $c = $this->db->fetchOne($query, $this->model->getId());
--- a/models/DataObject/AbstractObject/Dao.php
+++ b/models/DataObject/AbstractObject/Dao.php
@@ -316,7 +316,7 @@ class Dao extends Model\Element\Dao
             $userIds = $user->getRoles();
             $userIds[] = $user->getId();
 
-            $sql .= ' AND (select `list` as locate from `users_workspaces_object` where `userId` in (' . implode(',', $userIds) . ') and LOCATE(cpath,CONCAT(o.o_path,o.o_key))=1 ORDER BY LENGTH(cpath) DESC LIMIT 1)=1';
+            $sql .= ' AND (select `list` as locate from `users_workspaces_object` where `userId` in (' . implode(',', $userIds) . ') and LOCATE(cpath,CONCAT(o.o_path,o.o_key))=1 ORDER BY list DESC LIMIT 1)=1';
         }
         $sql .= ' LIMIT 1';
 
--- a/models/Document/Dao.php
+++ b/models/Document/Dao.php
@@ -341,7 +341,7 @@ class Dao extends Model\Element\Dao
             $userIds = $user->getRoles();
             $userIds[] = $user->getId();
 
-            $sql .= ' AND (select `list` as locate from `users_workspaces_document` where `userId` in (' . implode(',', $userIds) . ') and LOCATE(cpath,CONCAT(d.path,d.`key`))=1  ORDER BY LENGTH(cpath) DESC LIMIT 1)=1';
+            $sql .= ' AND (select `list` as locate from `users_workspaces_document` where `userId` in (' . implode(',', $userIds) . ') and LOCATE(cpath,CONCAT(d.path,d.`key`))=1  ORDER BY list DESC LIMIT 1)=1';
         }
 
         $sql .= ' LIMIT 1';


roland4432:
@brusch anything missing or do you think this might have unwanted sideeffects?

brusch:
@roland4432 unfortunately we haven't got the time yet to review your PR - as it is about permissions we need to review thoughtful 😉 We'll probably take it into our sprint next week (it's prioritized already in our backlog).

roland4432:
Thank you for your feedback. Justed wanted to check in, if I could support anyhow, not to put pressure. But I think you know that ;)

kingjia90:
@roland4432 i took a quick look but i didn't fully get this step, could you please rephrase it or provide a screenshot of the role permissions?
> * create 2 roles and set them up, that one mustn't list the children with the longer name and at least list rights for the shorter named child

roland4432:
No problem:

* setup 2 elements under a parent element, so that 1 of them is longer string-wise (e.g. test and testtest)
* add 2 roles, and set them up that 1 role is not allowed to list the testtest element, and the other role is allowed to list test
* create a user and apply those 2 roles to the user

=> you are not allowed to see the children since the order is currently set to use the cpath _desc_ to check for children visibility rights

kingjia90:
@roland4432 thank you for clarifying!

My findings are that while it works and fixes for the case of longer name, it also seems that the change would make the `+` icon appear on every folder/element, despite of the fact that all its children are not allowed (therefore the `+` is present but when you click on it, it will show nothing). 

To reproduce, just switch the shorter one `test` to be also not allowed and having no other elements (beside the already disallowed `testest`) in the same folder. 
Could you please double-check that?

On a SQL query level, the change you proposed seem always return all the children. Hope it helps.
image
I've set the LIMIT to 3 to debug, it takes only one to return true on `hasChildren`

roland4432:
So.. things go worse.. since I cannot replicate the error on 2 different (but version and data-wise identical) instances but I managed to track the issue to this:

image

and of course with the statement from the DAO `LIMIT 1` this will falsely reduce to:

image

and the whole statement for the `hasChildren` method:

image

As you see we have a lot of roles assigned to this user and some of them neglect `list` permissions, some apply them. I thought this might happen to sorting issues, hence my PR, since the following edit to `ORDER BY locate` will solve this false negative:

image


Right now I am more confused than yesterday how and why this happens. But I think this dataset might give you a rough idea how this happens.

kingjia90:
Thank you for tracking down the issue, i admit that i am confused as well (i am new to all this 😄) but these info shed a light.

I was surprised by the order of `userId` ordered by the same `cpath`. 
It's not totally random but it's quite unpredictable (referring to 768-771 then 775-759).

This lead me to find that [here](https://github.com/pimcore/pimcore/blob/e05650390cb6e515fe3f37afeb5ef995eda10f80/models/Asset/Dao.php#L453) we have a second and third sorting parameters that would refine it, and will be used [here](https://github.com/pimcore/pimcore/blob/e05650390cb6e515fe3f37afeb5ef995eda10f80/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php#L302) (which is after you click the `+` icon). So it would make sense that they somehow match, otherwise it would lead to the discrepancy i noticed (that `+` appears but shows nothing). 
Therefore, the proposed change in the PR is not applicable at the moment, either we change both places to sort by `list` (and any other places to following same/similar sorting rules) or add the "missing" second and/or third sorting parameter to `hasChildren`, **but** it seems it still won't fix the issue.

I guess the idea of sorting by cpath was to find the most nested item, which could makes sense but this is voided by elements with long names (in other words _folder1/disallowedfilewithanexceptionallongname_ > _folder1/folder2/nestedfile_ > _folder1/file_). 
If that's true, then i was  wondering if there would be a possibility to improve the SQL with some `SPLIT by the slashes and COUNT` and/or maybe `GROUP BY cpath HAVING list = 1` to avoid conflict when multiple roles are assigned 🤔 

That's what i am thinking so far, but unfortunately haven't found any good idea.

roland4432:
Thx for your detailed answer. Glad to hear, that you are just as lost as I am :D
Regarding the nested elements you were thinking about, which caught me as well, **but** they are not nested, since you have a `where parent_id = ?` in the outer statement and are not checking nested children, only direct (meaning 1st level) children. The decendent children (grandchildren and so on) are queried in separate calls.

So that's where I am stuck as well. But it is definitly not working as intended. I just don't see the part which solves it all that has to be altered :/

kingjia90:
>Regarding the nested elements you were thinking about, which caught me as well, **but** they are not nested, since you have a `where parent_id = ?` in the outer statement and are not checking nested children, only direct (meaning 1st level) children. The decendent children (grandchildren and so on) are queried in separate calls.
> 
The subquery has `LOCATE(cpath,CONCAT(a.path,a.filename))=1` condition which actually does a "textual check" on all grandchildren and further, by simply getting any element in the table that have a starting path that matches the path of the parent.

roland4432:
Yes you are right, I misread the statement. But the subquery does check grandchildren++. I'll have a thought about that and how it affects the SQL statement.

kingjia90:
any new updates on this topic? 
if no, i hope you won't mind if i close this PR for the moment (because it cannot be merged, but is a good finding though) and open a new issue with more "condensed" details of what we found here (i will link this PR as documentation and ping you there if you want to join the upcoming suggestions and discussion). 

It seems clear to me that this `ORDER BY LENGTH(cpath)` approach (which has 38 occurences in 12 files) need to be re-thinked at a bigger picture, because in all these places, it would also break the expected behaviour if there's a file with a long name.

roland4432:
I am fine with closing this PR and tracking issues in another one. The PR is obsolete, since that error will affect more places in the code.

I am playing around with some `group by` and `order by` statements and also kick subqueries as I think there are more elegant solutions for this problem.

Let's keep us updated on the issue. Maybe you can just ping me from the following issue. That would be awesome.

Cheers :)

Issue #11316, opened 2022-01-31
[Bug]: Thumbnail <img> fallback does not have srcset attribute


John5:
### Expected behavior

In Pimcore 6.x thumbnail images generate an img element with srcset attribute with multiple versions. Only if media queries are defined are picture/source used.

I would expect the img element to have a srcset attribute like it did before.
Also I expect to have control over which HTML elements are used.

Maybe adding an option like this for use cases when the picture/source  elements are undesirable:
image.thumbnail('myThumb').html({'tag': 'img'})

### Actual behavior

In Pimcore 10.x picture/source is now always used for all images. The img element is included as a fallback but it does not have the srcset attribute.

This causes a tool like wget to not download the high resolution image when downloading the page.

### Steps to reproduce

Add a thumbnail mage to a template and observe generated HTML source. Look for img element and see that srcset attribute is missing.

brusch:
Since the `<img>` tag is just a fallback, I'd not include the `srcset` to keep it simple...

John5:
Like I mentioned in the bug report there are clients like GNU Wget 1.21.2  that do support srcset but not the source/picture elements and it is causing real compatibility issues.

I understand if this has no priority for you but I think I can make a fix and submit a PR and I hope you will consider merging it.

brusch:
@John5 thanks for your feedback - I'd be ok with a PR 😻

Brainshaker95:
@brusch With this PR, the issue of #8818 pops up again. LQIP do not load the content of src attribute with srcset present.

brusch:
@Brainshaker95 can you please create a new issue for that, or even better a PR fixing the issue? Thanks a lot in advance!

Brainshaker95:
#12045

brusch:
@Brainshaker95 Thanks a lot! 👍

PR #11338, opened 2022-02-02
Thumbnail <img> fallback should have srcset attribute

John5:
Addresses issue #11316. img element will now have srcset attribute like it used to have in Pimcore 6 for improved compatibility with HTTP clients that can use this attribute but not source/picture elements.

### DIFF
--- a/models/Asset/Image/Thumbnail.php
+++ b/models/Asset/Image/Thumbnail.php
@@ -421,6 +421,7 @@ final class Thumbnail
             $attributes = $callback($attributes);
         }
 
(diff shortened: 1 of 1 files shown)


brusch:
This duplicates quite some code, could you please optimize that? Thanks in advance!

John5:
I will look into optimizing it, but it may take a few weeks before I get around to it.

brusch:
Ok, thanks for your feedback - I'll close the PR in the meantime, when you're ready, just push your updates and we'll reopen it again 😊 Looking forward to the improvements!

Issue #11341, opened 2022-02-02
[Improvement]: add method Relations to ClassificationStore class CollectionConfig


awroblewski:
### Improvement description

Hi,

The classificationStore class GroupConfig implements a method called "Relations" giving a list of all keys belonging to a group. However, the CollectionConfig class did not have such a method giving a  list of all groups belonging to a collection.
To stay consistent across the implementation of the classificationstore classes GroupConfig and CollectionConfig we should implement the Relations method for CollectionConfig as well.

brusch:
Fixed by #11342

PR #11342, opened 2022-02-02
add method Relations to class collectionConfig

awroblewski:
adding Relations method as described in Improvement request #11341


jdreesen:
As an improvement, it should probably target the `10.x` branch.

dvesh3:
@awroblewski thanks!

PR #11343, opened 2022-02-02
Fix for exception in Table.php

tanaydin:
Fix for Table data type column size change exception.
To reproduce
* create a table datatype with fixed number of columns
* save some data in there with a object
* close the object
* add or remote columns to table
* try to reopen object.

URL: /admin/object/save?task=publish
Method: PUT
Message: Warning: Undefined array key 6
Trace: 
in /var/www/html/vendor/pimcore/pimcore/models/DataObject/ClassDefinition/Data/Table.php:375

  
  

Resolves #


kingjia90:
Thank you, it works!

To reproduce the error, i needed to have "Activate Column Configuration" flagged and set; under other circumstances it works fine.

PR #11348, opened 2022-02-02
Added information about calling getImageTag

John5:
Update to documentation about using thumbnail images.


dvesh3:
@John5 thanks!

PR #11349, opened 2022-02-02
Passing null to parameter #1 ($bundle) of type string is deprecated

xlttj:
ResourceBundle::getLocales(): Passing null to parameter #1 ($bundle) of type string is deprecated in PHP 8.1


brusch:
Thank you very much!

PR #11353, opened 2022-02-03
change grid read method to post instead of get. If the data gets comp…

ascheider:
…lex with many long field names, the get request fails, du maximum length.

Currently the grid use GET method for the read operations. On environments with big classes and much bricks and field collections you get the error if the maximum url length is reached.
Alle Operations of the grid already use POST method except GET, which is better, because you never know the data.


dvesh3:
@ascheider Could please rebase it on 10.3 branch? since it is a fix.

dvesh3:
🏓 @ascheider

dvesh3:
Pushed the change to 10.3 branch https://github.com/pimcore/pimcore/commit/c76da3ced68266b888680d1a76b3ac13e24e41df. thanks!

PR #11356, opened 2022-02-03
Patch 2

xlttj:

Fixes a deprecation message 

> Deprecated: Return type of Pimcore\Document\Editable\Block\BlockState::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/vendor/pimcore/pimcore/lib/Document/Editable/Block/BlockState.php on line 108

### DIFF
--- a/lib/Document/Editable/Block/BlockState.php
+++ b/lib/Document/Editable/Block/BlockState.php
@@ -105,7 +105,7 @@ final class BlockState implements \JsonSerializable
     /**
      * {@inheritdoc}
      */
-    public function jsonSerialize()
+    public function jsonSerialize(): mixed
     {
         return [
             'blocks' => $this->blocks,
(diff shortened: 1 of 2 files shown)


lib/Document/Editable/Block/BlockState.php
```diff
@@ -105,7 +105,7 @@ public function clearIndexes()
     /**
      * {@inheritdoc}
      */
-    public function jsonSerialize()
+    public function jsonSerialize(): mixed
```
blankse, line 108:
```suggestion
    public function jsonSerialize(): array
```

blankse:
@xlttj I create an PR with all usages of jsonSerialize():
https://github.com/pimcore/pimcore/pull/11358

lib/Document/Editable/Block/BlockState.php
xlttj: See https://www.php.net/manual/en/class.jsonserializable.php

```php
/**
 * Objects implementing JsonSerializable
 * can customize their JSON representation when encoded with
 * <b>json_encode</b>.
 * @link https://php.net/manual/en/class.jsonserializable.php
 * @since 5.4
 */
interface JsonSerializable
{
    /**
     * Specify data which should be serialized to JSON
     * @link https://php.net/manual/en/jsonserializable.jsonserialize.php
     * @return mixed data which can be serialized by <b>json_encode</b>,
     * which is a value of any type other than a resource.
     * @since 5.4
     */
    #[TentativeType]
    public function jsonSerialize(): mixed;
}
```

blankse: See https://www.php.net/manual/en/language.oop5.variance.php

> Covariance allows a child's method to return a more specific type than the return type of its parent's method. Whereas, contravariance allows a parameter type to be less specific in a child method, than that of its parent.

PR #11358, opened 2022-02-03
Fix deprecations with jsonSerialize() methods

blankse:
See https://github.com/pimcore/pimcore/pull/11356
Fix all usages of jsonSerialize() method

Additional:
- "return $this" should be return typehint "static"
(Maybe in PHP 8.2 it is possible to use return type $this: https://wiki.php.net/rfc/this_return_type)
- EncryptedField: Add missing LayoutDefinitionEnrichmentInterface

### DIFF
--- a/bundles/AdminBundle/HttpFoundation/JsonResponse.php
+++ b/bundles/AdminBundle/HttpFoundation/JsonResponse.php
@@ -28,7 +28,7 @@ class JsonResponse extends BaseJsonResponse
     /**
      * {@inheritdoc}
      */
-    public function setData($data = [])// : self
+    public function setData($data = [])// : static
     {
         $serializer = Serialize::getAdminSerializer();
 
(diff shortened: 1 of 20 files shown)


xlttj (changes requested):
See See https://www.php.net/manual/en/class.jsonserializable.php

Return type shoult be `mixed` to match the interface, not `array` or `self`

blankse:
@xlttj See https://www.php.net/manual/en/language.oop5.variance.php

> In PHP 7.2.0, partial contravariance was introduced by removing type restrictions on parameters in a child method. As of PHP 7.4.0, full covariance and contravariance support was added.
> 
> Covariance allows a child's method to return a more specific type than the return type of its parent's method. Whereas, contravariance allows a parameter type to be less specific in a child method, than that of its parent.

xlttj:
@blankse ack.

dvesh3 (approved):

dvesh3:
@blankse thanks a lot!

PR #11361, opened 2022-02-03
Prevent error when version file does not exist

BlackbitDevs:
Currently when the version file does not exist or is not readable Flysystem will throw a `UnableToReadFile` exception - for example when opening a data object in Pimcore backend. Thus you cannot open the object at all. 
With this PR the exception gets caught.


blankse:
Why not only catch this type of exception?

BlackbitDevs:
You mean not to catch `Throwable` but the real exceptions?

blankse:
Yes, I think we should avoid to catch just Throwable. Only named and known exceptions should be caught. Otherwise we make it difficult for ourselves to find bugs/errors (for example there was throwing/catching exceptions in dao classes for not found models and there was an error in the SQL which also throws an exception).

BlackbitDevs:
Ok, changed it to the 2 mentioned exceptions from https://github.com/thephpleague/flysystem/blob/9f8b4260031bfef5497477da53dd57b2aaff0e06/src/FilesystemReader.php#L34-L38

models/Version.php
```diff
@@ -366,7 +368,11 @@ public function delete()
     public function loadData($renewReferences = true)
     {
         $storage = Storage::get('version');
-        $data = $storage->read($this->getStorageFilename());
+        try {
+            $data = $storage->read($this->getStorageFilename());
+        } catch(UnableToReadFile | FilesystemException $e) {
```
jdreesen: I don't think it's necessary to add both here, because the `UnableToReadFile` is already a `FilesystemException`, no?

jdreesen: Well okay, if Flysystem actually mentions both... maybe we should leave it as is...

blankse: It would be good to know when which exception is caused. Maybe we shouldn't always delete the version if the error is only temporary.

blankse: In the docs there is only one: https://flysystem.thephpleague.com/v2/docs/usage/exception-handling/
Edit: FilesystemException is for catching all. But we only use read() here. So UnableToReadFile should be enought.

brusch:
@BlackbitNeueMedien can you please check the failing check - thanks a lot 😊

blankse:
@brusch This error is in all new PRs. I fixed it here: https://github.com/pimcore/pimcore/pull/11367

brusch:
Oh, I'm sorry 😅 Wasn't yet at this PR 😉

models/Version.php
BlackbitDevs: Changed it to only catch `UnableToReadFile`

PR #11365, opened 2022-02-03
[Documents] fix Listing cachekey empty string

### LINKED ISSUE #10990: [Bug]: cachekey generation will result in empty string
### Expected behavior

Getting children of hardlinks or documents should get its appropriate children.

### Actual behavior

Throws exception:
An exception has been thrown during the rendering of a template ("Warning: Undefined array key """).

Caused by: 
/models/Document/Hardlink/Wrapper:161
```php
return $this->children[$cacheKey];
```

The corresponding $cacheKey is generated via:
/models/Document:1376
```php
protected function getListingCacheKey(array $args = [])
{
    $unpublished = (bool)($args[0] ?? false);
    $cacheKey = (string)$unpublished;

    return $cacheKey;
}
```
This will create an empty string for $cacheKey if $unpublished is _false_. I am not quite sure, which key should have been generated at this point, else I would drop a PR for that.

But maybe @dvesh3 or @brusch can enlighten me :)

Thx

### Steps to reproduce

Create hardlink. Get the hardline via e.g. id. Then call _getChildren()_ without any parameters.

dvesh3:
Resolves #10990


models/Document.php
```diff
@@ -1407,10 +1407,9 @@ public function setVersionCount(?int $versionCount): ElementInterface
      */
     protected function getListingCacheKey(array $args = [])
     {
-        $unpublished = (bool)($args[0] ?? false);
-        $cacheKey = (string)$unpublished;
+        $includingUnpublished = (bool)($args[0] ?? false);
 
-        return $cacheKey;
+        return 'document_list' . (!empty($includingUnpublished) ? '_' : '') . $includingUnpublished;
```
blankse:
```suggestion
        return 'document_list_' . ($includingUnpublished ? '1' : '0');
```

PR #11366, opened 2022-02-03
Tool\Email\Log in the Flysystem

### LINKED ISSUE #11357: [Improvement]: Storage for Tool\Email\Log should also use Flysystem
### Improvement description

This is still using local path: 
https://github.com/pimcore/pimcore/blob/66c89bf84736c0128b982cbfad42cabff2bfe779/models/Tool/Email/Log.php#L394

It should get it's own Flysystem storage here: 
https://github.com/pimcore/pimcore/blob/417b50202ceba7721311eb73f92c308f84236546/bundles/CoreBundle/Resources/config/pimcore/flysystem.yaml#L12

And implemented similar to all the other storages: 
https://github.com/pimcore/pimcore/blob/c906d2ba8d772f7c1e3800d05e916c58cd6a0517/bundles/AdminBundle/Controller/Admin/MiscController.php#L131

kingjia90:
resolves #11357


ramundomario (changes requested):
What do you think about creating function to get storage file name (.log/.html) instead composing it in multiple places? :blush:

PR #11367, opened 2022-02-03
Fix current phpstan error

blankse:
With the line `composer config minimum-stability "dev"` all dependencies are in dev version, also phpstan. I think only the symfony packages should be in dev. The phpstan repository seems to have a bug in dev-master.


brusch:
But with this change Symfony is not in `dev` anymore, right?

blankse:
The remain on `dev`. All symfony packages are on the 5.4-dev branch, but the other packages are on stable.

brusch:
Ah yes, true - thanks a lot 👍

PR #11369, opened 2022-02-04
Fix messenger configuration

### LINKED ISSUE #11355: [Bug]: Invalid option(s) "queue_name" passed to the Redis Messenger transport.
### Expected behavior

Switch the transports (Redis, In Memory, etc) for the messenger component

### Actual behavior

The problem is, that with 10.3 there was a messenger configuration "queue_name" introduced that is only available if you use Doctrine as transport. In case you use other transports, Redis in my example, everything fails because the option is not available in the Redis configuration. Therefore you need to switch to Doctrine or create your own transports without the option to fix the problem.

### Steps to reproduce

1. Install Redis
2. Configure Redis as transport
3. Try to save something in the backend or run the command `bin/console messenger:consume pimcore_core pimcore_maintenance` 

I don't know if it was intended that only doctrine is working out of the box from now on but at least a hint the Upgrade Notes would be helpful, that you need to configure completely new transports if you want to use other systems.

matthiashamacher:
Resolves #11355


matthiashamacher:
@dvesh3 Any Idea why the check fails?

dvesh3:
Please merge latest changes from 10.3 branch. Stan is fixed on base https://github.com/pimcore/pimcore/pull/11367

dvesh3:
@matthiashamacher thanks!

PR #11370, opened 2022-02-04
[Bug]: Zoom in Image Standard Preview after using 360° Viewer

### LINKED ISSUE #11052: [Bug]: Zoom in Image Standard Preview after using 360° Viewer
### Expected behavior

When viewing an Image Asset in the admin backend preview, selecting the "360° Viewer" option, then toggling back to "Standard Preview", the original preview should be restored.

### Actual behavior

When viewing an Image Asset in the admin backend, selecting the "360° Viewer" option, then toggling back to "Standard Preview" leads to an unintended zoom if the image width/height ratios diverge too heavily from each other. See these two images on the pimcore public demo system for reference (images were created using content from the public demo system, the list preview of the "Car Images" folder).
The first screenshot was taken before the 360° Viewer had been triggered on the image :

Bildschirmfoto 2021-12-17 um 10 45 56

The second screenshot was taken after the 360° Viewer had been triggered on the image :

Bildschirmfoto 2021-12-17 um 10 46 08

### Steps to reproduce

* upload image with low width and high height value (e.g. 800 width and 1800 height, as seen in the example) into pimcore public demo system
* select image to trigger preview
* toggle "360° Viewer" button on the right sidebar
* toggle "Standard Preview" button on the right sidebar
* find the image preview in zoomed-in state

mcop1:
closes #11052

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/image.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/image.js
@@ -20,6 +20,8 @@ pimcore.asset.image = Class.create(pimcore.asset.asset, {
         this.id = intval(id);
         this.setType("image");
         this.addLoadingPanel();
+        this.previewPanelWidth = 800;
+        this.previewPanelHeight = 600;
 
         pimcore.plugin.broker.fireEvent("preOpenAsset", this, "image");
 
@@ -107,7 +109,7 @@ pimcore.asset.image = Class.create(pimcore.asset.asset, {
             this.editPanel = new Ext.Panel({
                 title: t("edit"),
                 html: '<iframe src="' + url + '" frameborder="0" ' +
-                'style="width: 100%;" id="' + frameId + '"></iframe>',
+                    'style="width: 100%;" id="' + frameId + '"></iframe>',
                 iconCls: "pimcore_material_icon_edit pimcore_material_icon"
             });
             this.editPanel.on("resize", function (el, width, height, rWidth, rHeight) {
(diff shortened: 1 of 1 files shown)


kingjia90:
Looks fine and ready to merge, but why not just moving the area max size setter inside the `initPreviewImage` (so it will stay similar to `initPreviewVr`, avoiding the extra method `setPreviewPanelSize`), something like this:
```
  this.displayPanel.on('resize', function () {
      if(this.previewMode == 'vr') {
          this.initPreviewVr();
      } else {
          this.initPreviewImage();
      }
  }.bind(this));
}
```
while changing `width` to `this.displayPanel.getWidth()` (same for the height, which already works similarly [here](https://github.com/pimcore/pimcore/blob/35c5e83d99c8244979b09817dac198ce1726d260/bundles/AdminBundle/Resources/public/js/pimcore/asset/image.js#L456)) like this:
```
initPreviewImage: function () {
    var html = '<img src="' + this.data.imageInfo['previewUrl'] + '">';
    ....
    var area = this.displayPanel.getEl().down('img');
    if(area) {
        area.setStyle('max-width', (this.displayPanel.getWidth() - 340) + "px");
        area.setStyle('max-height', (this.displayPanel.getHeight() - 40) + "px");
    }
},
```
so we could clean these legacy(?) parameters on the resize event 
https://github.com/pimcore/pimcore/blob/35c5e83d99c8244979b09817dac198ce1726d260/bundles/AdminBundle/Resources/public/js/pimcore/asset/image.js#L434

wdyt?

mcop1:
@kingjia90 you´re right, I changed it. Regarding the legacy parameters, I don´t know where they come from, but since they are not  used anywhere I removed them.

PR #11373, opened 2022-02-04
Refactor Classes Build

dpfaffenbauer:
PHP Class Building does NOT need a Database to work. Currently there is only one way of doing that with `deployment:classes-rebuild`. That command also makes the DB changes though. So in order the create Classes in a PHP Command, we need to separate the Class Generation. This also helps to get rid of business logic in ClassDefinitions and move it to separate services. They can then also be extended if needed.

This PR refactors the Class Generation into separate Services. This also prepares Pimcore to use something like https://github.com/nette/php-generator to make that even prettier and more sexy.

**This does NOT make the `classes-rebuild` command obsolete, it just refactors the Code to make it able to generate PHP Classes in build processes**

Next step: Refactor DB Changes into a separate command to completely differentiate DB Changes (env specific) with Class Changes (source code or build specific). That is more complicated though. I already started to build a POC here: https://github.com/pimcore/pimcore/pull/7786. But it never got much attention....


dpfaffenbauer:
@BlackbitNeueMedien @jdreesen @solverat @dkarlovi As part of the "core" community, what is your Opinion about something like this?

Obviously: @brusch @dvesh3 As pimcore team: WDYT about that?

dkarlovi:
@dpfaffenbauer I'm absolutely for a change where we can know the changes which will be done to the database at deploy time before deploy time. Having a migrations-like system for Pimcore is from my experience one of the largest drawbacks Pimcore currently has and we've already had issues because of the ad-hoc nature of DDL queries running on our production systems.

The advantage of migrations are many and introducing support for it should be viewed as a high priority issue. Any step taken in that direction is a right one. :+1:

brusch:
@dpfaffenbauer I like the idea ♥ However, the current approach seems a bit over engineered to me. I don't think we need interfaces for all the builders and I also think that it would be sufficient enough to have just 2 builder services (class and listing) per type (class definition, field collection, brick). I understand the intention, but the actual benefit is low imho, but it makes it even more complex to maintain. WDYT?

dpfaffenbauer:
@brusch Sure, I can simplify where possible. It still will be more then 2 per type, cause fields and doc-blocks are reused.

brusch:
@dpfaffenbauer sounds good 👍

dpfaffenbauer:
@brusch simplified, we now have following services:
 - ClassBuilder for DataObjects
 - FieldCollectionClassBuilder
 - ObjectBrickClassBuilder
 - ObjectBrickContainerClassBuilder
 - FieldDefinitionPropertiesBuilder
 - FieldDefinitionDocBlockBuilder
 - FieldDefinitionBuilder
 - ListingClassBuilder
 - ListingClassPropertiesBuilder
 - Persisters for all generated classes (DataObject and Listing, FC, Brick and BrickContainer)

bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php
```diff
@@ -0,0 +1,81 @@
...
+        protected PHPObjectBrickClassWriterInterface $brickClassWriter,
+        protected PHPObjectBrickContainerClassWriterInterface $brickContainerClassWriter,
+    ) {
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this
+            ->setName('pimcore:build:classes')
```
brusch, line 44: I think we should choose a more self explaining name here, having also `pimcore:deployment:classes-rebuild` in mind. Maybe also in the same namespace and we can also think about renaming `pimcore:deployment:classes-rebuild` and providing a fallback? WDYT?

bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php
```diff
@@ -0,0 +1,81 @@
...
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $objectClassesFolder = PIMCORE_CLASS_DEFINITION_DIRECTORY;
+        $files = glob($objectClassesFolder.'/*.php');
+
+        $changes = [];
+
+        foreach ($files as $file) {
+            $class = include $file;
```
brusch, line 61: Why a different approach for classes? 
It's required that the class definition is also in the `classes` table, so why don't use the listing here as well, as `writePHPClasses()` doesn't create the database entry for the class definition.

bundles/CoreBundle/Resources/config/class_builder.yaml
```diff
@@ -0,0 +1,63 @@
+services:
+    _defaults:
+        public: true
+
+    Pimcore\DataObject\ClassBuilder\ClassBuilderInterface: '@Pimcore\DataObject\ClassBuilder\ClassBuilder'
+    Pimcore\DataObject\ClassBuilder\ClassBuilder:
+        arguments:
+            - '@Pimcore\DataObject\ClassBuilder\FieldDefinitionDocBlockBuilderInterface'
```
brusch: This should be possible with autowiring as well, right?

bundles/CoreBundle/Resources/config/class_builder.yaml
```diff
@@ -0,0 +1,63 @@
...
+    Pimcore\DataObject\ClassBuilder\ListingClassFieldDefinitionBuilderInterface: '@Pimcore\DataObject\ClassBuilder\ListingClassFieldDefinitionBuilder'
+    Pimcore\DataObject\ClassBuilder\ListingClassFieldDefinitionBuilder: ~
+
+    Pimcore\DataObject\ClassBuilder\ObjectBrickClassBuilderInterface: '@Pimcore\DataObject\ClassBuilder\ObjectBrickClassBuilder'
+    Pimcore\DataObject\ClassBuilder\ObjectBrickClassBuilder:
+        arguments:
+            - '@Pimcore\DataObject\ClassBuilder\FieldDefinitionDocBlockBuilderInterface'
+
+    Pimcore\DataObject\ClassBuilder\ObjectBrickContainerClassBuilderInterface: '@Pimcore\DataObject\ClassBuilder\ObjectBrickContainerClassBuilder'
+    Pimcore\DataObject\ClassBuilder\ObjectBrickContainerClassBuilder: ~
```
brusch: I'd prefer to not have service aliases, as they are also public then by default, anything that's against using just the "interface"-services? 
So: 
```yaml
Pimcore\DataObject\ClassBuilder\ObjectBrickContainerClassBuilderInterface:
    class: Pimcore\DataObject\ClassBuilder\ObjectBrickContainerClassBuilder
```

brusch (changes requested):

bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php
dpfaffenbauer: That is intentional. This command should be able to run WITHOUT Database.

bundles/CoreBundle/Resources/config/class_builder.yaml
dpfaffenbauer: this is basically the way I always do it in Symfony. I think it is also the Symfony way of doing it... Can change though if you want.

bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php
dpfaffenbauer: Idea is to create all PHP Classes during a Docker build for example where no db is available

bundles/CoreBundle/Resources/config/class_builder.yaml
dpfaffenbauer: sure

bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php
dpfaffenbauer: It is not a fallback to `pimcore:deployment:classes-rebuild`. It is a different approach, this here just builds the classes, it doesn't run database schema changes or even creation of the classes. So, the idea was to introduce the 'build' namespace for things that can happen during a eg. docker build.

lib/DataObject/ClassBuilder/PHPClassWriterInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\DataObject\ClassBuilder;
+
+use Pimcore\Model\DataObject\ClassDefinition;
+
+interface PHPClassWriterInterface
+{
+    public function writePHPClasses(ClassDefinition $classDefinition): void;
```
dkarlovi: Note, Symfony typically uses `dump` for this type of role, it's a "dumper", not a "writer".

dkarlovi (approved):

lib/DataObject/ClassBuilder/PHPClassWriterInterface.php
dpfaffenbauer: good to know, I'll rename it

dpfaffenbauer: done

bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php
brusch: Seems like a misunderstanding 😉 I know it's for a different purpose, I just thought it would make sense to have related stuff in the same groups. I'm not sure if it makes sense to introduce another namespace for that ... we can leave that open for discussion for now.

bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php
brusch: Got it ... I'm just wondering whether we should change `ClassDefinition\Listing` to your approach anyway 🤔

bundles/CoreBundle/Resources/config/class_builder.yaml
brusch: Well, this case is probably special, because here we have both `ObjectBrickContainerClassBuilderInterface` and `ObjectBrickContainerClassBuilder` `public` ... and the latter shouldn't be public imho.

bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php
dpfaffenbauer: Would that be a BC break? That is the question...

bundles/CoreBundle/Resources/config/class_builder.yaml
dpfaffenbauer: true, that was for convenience so I don't have to make single services public. I can change that though

brusch: Yeah, would make sense, as it is best practice and afaik `public` services are also more expensive.

bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php
brusch: Hmm, yes, probably, because of possible conditions ... forget about it, topic for XI then 😉

bundles/CoreBundle/Resources/config/class_builder.yaml
brusch: So only this is left then, the rest looks good to me. Thanks!

brusch:
#10242 needs to be considered vice versa, depending on which PR get's merged first 😉

bundles/CoreBundle/Resources/config/class_builder.yaml
dpfaffenbauer: changed

dpfaffenbauer: changed

dpfaffenbauer: changed

brusch:
@dpfaffenbauer can you please check the tests and resolve the conflicts? 
Conflict caused by https://github.com/pimcore/pimcore/pull/10242/files#diff-c26bc31ef67835011f2d069ad6edb77dd43d22b385d9936631a98bfbcf054497L578

Thanks!

dpfaffenbauer:
@brusch done

models/DataObject/Objectbrick/Definition.php
```diff
@@ -633,7 +464,7 @@ private function createContainerClasses()
      *
      * @return string
```
brusch, line 466:
```suggestion
     * @internal
     * @return string
```

models/DataObject/Objectbrick/Definition.php
```diff
@@ -644,7 +475,7 @@ private function getContainerClassName($classname, $fieldname)
...
```
brusch, line 478:
```suggestion
     * @internal
     * @return string
```

models/DataObject/Objectbrick/Definition.php
```diff
@@ -654,7 +485,7 @@ private function getContainerNamespace($classname, $fieldname)
...
```
brusch, line 489:
```suggestion
     * @internal
     * @return string
```

models/DataObject/Objectbrick/Definition.php
```diff
@@ -762,7 +593,7 @@ public function getDefinitionFile($key = null)
...
```
brusch, line 598:
```suggestion
     * @internal
     * @return string
```

models/DataObject/Objectbrick/Definition.php
```diff
@@ -472,7 +397,7 @@ private function updateDatabase()
      *
```
brusch:
```suggestion
     * @internal
```

models/DataObject/Fieldcollection/Definition.php
```diff
@@ -323,48 +265,8 @@ public function getDefinitionFile($key = null)
      *
      * @return string
```
brusch, line 267:
```suggestion
     * @internal
     * @return string
```

models/DataObject/ClassDefinition.php
```diff
@@ -821,12 +607,12 @@ public function getDefinitionFile($name = null)
         return $this->locateDefinitionFile($name ?? $this->getName(), 'definition_%s.php');
     }
```
brusch, line 609:
```suggestion

    /**
     * @internal
     */
```

models/DataObject/ClassDefinition.php
```diff
@@ -821,12 +607,12 @@ public function getDefinitionFile($name = null)
...
 
-    private function getPhpClassFile(): string
+    public function getPhpClassFile(): string
     {
         return $this->locateFile(ucfirst($this->getName()), 'DataObject/%s.php');
     }
```
brusch, line 617:
```suggestion

    /**
     * @internal
     */
```

brusch (changes requested):

brusch:
Changed the services so that only the dumpers are public - the others don't need to be public. 
Also added `@internal` to the methods where the visibility has changed.

brusch:
Thank you very much!

dpfaffenbauer:
🍻

jdreesen:
Maybe you wanna add this to the `10.4` milestone?

models/DataObject/ClassDefinition.php
```diff
@@ -668,64 +505,13 @@ public function generateClassFiles($generateDefinitionFile = true)
 
             $data = '<?php';
             $data .= "\n\n";
-            $data .= $infoDocBlock;
```
blankse, line 671: @dpfaffenbauer Why did you removed the info doc block from the definition files? We have only these files in the versioning and there we can fast see field changes. Can we restore it again?

jdreesen: I noticed this, too. It was quite useful to get a quick overview of what fields/types the class/brick/... has, which is otherwise very difficult to grasp.

blankse: I created a PR for this https://github.com/pimcore/pimcore/pull/12257

PR #11374, opened 2022-02-04
Allow Resetting Translator Initialization

solverat:
If you're processing data via messenger, all cache properties won't get updated until the worker gets restarted. We're pushing modified data in real-time based on customer's translation, so it's required to allow to reset translator's initialization. Otherwise, the translation catalog won't get updated while handling incoming envelopes.

### DIFF
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -256,6 +256,14 @@ class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleA
     }
 
     /**
+     * Reset Catalogues initialization
+     */
+    public function resetCache()
+    {
+        $this->initializedCatalogues = [];
+    }
+
+    /**
      * @param string $id
      * @param string $translated
      * @param array $parameters


lib/Translation/Translator.php
```diff
@@ -255,6 +255,14 @@ public function lazyInitialize($domain, $locale)
         }
     }
 
+    /**
+     * Reset Catalogues initialization
+     */
+    public function resetCache()
```
kingjia90, line 261: It looks fine, but feels like a rare practical use case and sort of too "destructive" of the purpose of lazy initialization.
What about something with a more specific scope like this?
```
    /**
     * Reset Catalogues initialization
     */
    public function resetInitialization($domain, $locale)
    {
        $cacheKey = 'translation_data_' . md5($domain . '_' . $locale);
        unset($this->initializedCatalogues[$cacheKey]);
    }
```
Would it fit in your case as well?

solverat: @kingjia90 I'm not sure about the "rare" since the messenger processes in pimcore is new, and I guess there will be a lot of upcoming new issues regarding those cached properties. However, I'm totally fine with your proposed changes. Should I update the PR?

kingjia90: @solverat For rare, i meant that the entire catalogue needing to be reinitialized while envelopes are handled one-by-one or in small batches, made me wondering if `resetCache` would be used at the beginning of each batch or so, **but** take it with a grain of salt because i don't know much about the messenger or the project needs. 

Yes, if everything's fine for you, please update the PR and i will merge it.

kingjia90: @solverat update: i had a quick chat with the team and concluded that it would be good to have that both options (clear all/clear one). 
So i am merging your original PR and i will open a new PR with my suggestion above. No further action is needed. 
Thank you again for your contribution!

solverat: awesome. thanks!

PR #11265, opened 2022-01-25
Search and move improvement (#11110)

karlozrilic:
## Changes in this pull request  
Based on #11110 suggestion I added checkboxes for easier asset selection.

## Additional info  
Clicking on already selected checkbox, asset is unselected and removed from store.

PR #11376, opened 2022-02-07
Asset selection fixes

karlozrilic:
After #11265 I noticed 2 small bugs.
First one is that if assets are filtered and user removes asset from selection, asset wouldn't uncheck.
Second one is when filtering assets, assets that are already selected wouldn't have their checkbox selected.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/element/selector/asset.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/element/selector/asset.js
@@ -256,11 +256,16 @@ pimcore.element.selector.asset = Class.create(pimcore.element.selector.abstract,
 
                             var checkbox = new Ext.grid.column.Check();
 
-                            if (value || rec) {
+                            if (typeof value ==='undefined' && rec !== null){
+                                this.resultPanel.getStore().getAt(rowIndex).set('asset-selected', true);
                                 return checkbox.renderer(true);
-                            } else {
-                                return checkbox.renderer(false);
                             }
+                              
+                            if (value && rec === null) {
+                                return checkbox.renderer(true);
+                            }
+                            
+                            return checkbox.renderer(false);
                         }.bind(this)
                     }
                 );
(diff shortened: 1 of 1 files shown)


fashxp:
since it is a bugfix, please rebase to 10.3. thx.

bundles/AdminBundle/Resources/public/js/pimcore/element/selector/asset.js
```diff
@@ -247,7 +249,19 @@ pimcore.element.selector.asset = Class.create(pimcore.element.selector.abstract,
...
                         dataIndex : 'asset-selected',
-                        sortable: false
+                        sortable: false,
+                        renderer: function (value, metaData, record, rowIndex) {
+                            var currentElementId = this.resultPanel.getStore().getAt(rowIndex).id;
+                            var rec = this.selectionStore.getData().find("id", currentElementId);
+
+                            var checkbox = new Ext.grid.column.Check();
+
+                            if (value || rec) {
```
kingjia90: nice work, but at the moment, it's never rendering unchecked checkbox, this OR condition is always true when clicked, on the first click of an empty checkbox, `value` is true and `rec` is null. if you click it again, `value` would be false but `rec` is set.
image

kingjia90: something like this works for me:

```
// mark checkbox when existing entries are found
// 'value' is always undefined here when loading/reloading, as no user interaction is made
if (typeof value ==='undefined' && rec !== null){
  this.resultPanel.getStore().getAt(rowIndex).set('asset-selected',true);
  return checkbox.renderer(true);
}

// when 'rec' is not already set and interaction is made, then mark the checkbox
if (value && rec === null) {
  return checkbox.renderer(true);
}
//stay unchecked in any other cases
return checkbox.renderer(false);
```

karlozrilic: Tnx for comment I didn't catch that before for some reason

PR #11378, opened 2022-02-07
[Feature]: add PRE_COPY event

### LINKED ISSUE #11264: [Feature]: add PRE_COPY event
### Feature description

I'd love to see a `PRE_COPY` event, similar to these lines:

https://github.com/pimcore/pimcore/blob/b03e38dcd0be38b3e2b048c98910b82fd500b998/models/DataObject/Service.php#L180-L183

> Should probably also exist for assets and documents.

dturopoli:
Resolves #11264 
PRE_COPY event added to documents, objects and assets

### DIFF
--- a/lib/Event/AssetEvents.php
+++ b/lib/Event/AssetEvents.php
@@ -118,7 +118,7 @@ final class AssetEvents
 
     /**
      * Arguments:
-     *  - base_element | Pimcore\Model\Asset | contains the base asset used in copying process
+     *  - target_element | Pimcore\Model\Asset | contains the target asset used in copying process
      *
      * @Event("Pimcore\Event\Model\AssetEvent")
      *
--- a/lib/Event/AssetEvents.php
+++ b/lib/Event/AssetEvents.php
@@ -15,6 +15,8 @@
 
 namespace Pimcore\Event;
 
+use Pimcore\Model\Asset;
+
 final class AssetEvents
 {
     /**
@@ -116,7 +118,17 @@ final class AssetEvents
 
     /**
      * Arguments:
-     *  - base_element | Pimcore\Model\Document | contains the base document used in copying process
+     *  - base_element | Pimcore\Model\Asset | contains the base asset used in copying process
+     *
+     * @Event("Pimcore\Event\Model\AssetEvent")
+     *
+     * @var string
+     */
+    const PRE_COPY = 'pimcore.asset.preCopy';
+
+    /**
+     * Arguments:
+     *  - base_element | Pimcore\Model\Asset | contains the base document used in copying process
      *
      * @Event("Pimcore\Event\Model\AssetEvent")
      *
--- a/lib/Event/DataObjectEvents.php
+++ b/lib/Event/DataObjectEvents.php
@@ -15,6 +15,8 @@
 
 namespace Pimcore\Event;
 
+use Pimcore\Model\DataObject\AbstractObject;
+
 final class DataObjectEvents
 {
     /**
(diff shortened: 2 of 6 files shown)


models/DataObject/Service.php
```diff
@@ -139,6 +139,10 @@ public function copyRecursive($target, $source)
         //load all in case of lazy loading fields
         self::loadAllObjectFields($source);
 
+        // triggers actions before object cloning
+        $event = new DataObjectEvent($source);
```
dvesh3: Should we also trigger the event when copying children?

dvesh3: @dturopoli 🏓

dturopoli: @dvesh3 Post copy event is already also triggering on copying children so I just used the same logic for pre copy

dvesh3: you are right, then it should be consistent with post copy

dvesh3: One more thing, like post copy contains `$new` as main element and passes `$source` as `base_element` in params. I think then it makes sense to pass `$target` as `target_element` in params of pre copy. Then you would know what will be the target and work with that info. wdyt?

dturopoli: I agree

dvesh3: could you please change it? 😄

dturopoli: Added target_element

models/Asset/Service.php
```diff
@@ -78,6 +78,12 @@ public function copyRecursive($target, $source)
 
         $source->getProperties();
 
+        // triggers actions before asset cloning
+        $event = new AssetEvent($source, [
+            'target_element' => $target,
+        ]);
+        \Pimcore::getEventDispatcher()->dispatch($event, AssetEvents::PRE_COPY);
```
dvesh3, line 85: we have to get the `$target` from event argument `$target = $event->getArgument('target_element');` after event dispatch, so that the latest one is used afterwards.

lib/Event/AssetEvents.php
```diff
@@ -116,7 +118,17 @@ final class AssetEvents
 
     /**
      * Arguments:
-     *  - base_element | Pimcore\Model\Document | contains the base document used in copying process
+     *  - base_element | Pimcore\Model\Asset | contains the base asset used in copying process
```
dvesh3:
```suggestion
     *  - target_element | Pimcore\Model\Asset | contains the target asset used in copying process
```

lib/Event/DataObjectEvents.php
```diff
@@ -118,7 +120,17 @@ final class DataObjectEvents
 
     /**
      * Arguments:
-     *  - base_element | Pimcore\Model\Document | contains the base document used in copying process
+     *  - base_element | Pimcore\Model\AbstractObject | contains the base object used in copying process
```
dvesh3:
```suggestion
     *  - target_element | Pimcore\Model\AbstractObject | contains the target object used in copying process
```

lib/Event/DocumentEvents.php
```diff
@@ -163,6 +163,16 @@ final class DocumentEvents
      */
     const PRINT_MODIFY_PROCESSING_CONFIG = 'pimcore.document.print.processor.modifyConfig';
 
+    /**
+     * Arguments:
+     *  - base_element | Pimcore\Model\Document | contains the base document used in copying process
```
dvesh3:
```suggestion
     *  - target_element | Pimcore\Model\Document | contains the target document used in copying process
```

dvesh3 (approved):

dvesh3:
@dturopoli 👍 for quick changes.....thanks a lot!

Issue #11379, opened 2022-02-07
[Bug]: Updating from Pimcore 10.2.7 to 10.3 through Composer results in Migration Exception


abcpremium:
### Expected behavior

Hello all,

I updated Pimcore from version 10.2.7 to 10.3 by following the guide at [https://pimcore.com/docs/pimcore/current/Development_Documentation/Installation_and_Upgrade/Updating_Pimcore/index.html](https://pimcore.com/docs/pimcore/current/Development_Documentation/Installation_and_Upgrade/Updating_Pimcore/index.html), executing `COMPOSER_MEMORY_LIMIT=-1 composer update`, however, the migration script that gets executed afterwards results in an exception. I never got this exception when performing updates before. Can anyone tell me what exactly is wrong? Or how I can fix this?

Thanks in advance!

### Actual behavior

```
[notice] Migrating up to Pimcore\Bundle\CoreBundle\Migrations\Version20220201132131
[error] Migration Pimcore\Bundle\CoreBundle\Migrations\Version20211103055110 failed during Execution. Error: "An exception occurred while executing 'ALTER TABLE object_relations_BS CHANGE COLUMN
                        `type` `type` VARCHAR(50)  NULL DEFAULT NULL ;':

SQLSTATE[42000]: Syntax error or access violation: 1171 All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead"

In AbstractMySQLDriver.php line 125:

  An exception occurred while executing 'ALTER TABLE object_relations_BS CHAN
  GE COLUMN
                          `type` `type` VARCHAR(50)  NULL DEFAULT NULL ;':

  SQLSTATE[42000]: Syntax error or access violation: 1171 All parts of a PRIM
  ARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead


In Exception.php line 18:

  SQLSTATE[42000]: Syntax error or access violation: 1171 All parts of a PRIM
  ARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead


In PDOConnection.php line 141:

  SQLSTATE[42000]: Syntax error or access violation: 1171 All parts of a PRIM
  ARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead


[...]

PR #10760, opened 2021-11-03
Change type column of object_relations_* tables to enum

BabovicT:
Resolves #10662

PR #11383, opened 2022-02-07
[DataObjects] Migrations - fix error: All parts of a PRIMARY KEY must be NOT NULL

dvesh3:
Resolves https://github.com/pimcore/pimcore/issues/11379

## Additional info  
follow up to #10760

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20211103055110.php
+++ b/bundles/CoreBundle/Migrations/Version20211103055110.php
@@ -36,7 +36,7 @@ class Version20211103055110 extends AbstractMigration
 
             if ($schema->hasTable($objectDatastoreTableRelation)) {
                 $this->addSql(
-                    "UPDATE $objectDatastoreTableRelation SET `type` = NULL WHERE `type` = 'object'"
+                    "UPDATE $objectDatastoreTableRelation SET `type` = 'object' WHERE `type` = NULL"
                 );
                 $this->addSql(
                     "ALTER TABLE $objectDatastoreTableRelation CHANGE COLUMN
--- a/bundles/CoreBundle/Migrations/Version20211103055110.php
+++ b/bundles/CoreBundle/Migrations/Version20211103055110.php
@@ -36,7 +36,7 @@ class Version20211103055110 extends AbstractMigration
 
             if ($schema->hasTable($objectDatastoreTableRelation)) {
                 $this->addSql(
-                    "UPDATE $objectDatastoreTableRelation SET `type` = 'object' WHERE `type` = NULL"
+                    "UPDATE $objectDatastoreTableRelation SET `type` = 'object' WHERE `type` = NULL OR `type` =''"
                 );
                 $this->addSql(
                     "ALTER TABLE $objectDatastoreTableRelation CHANGE COLUMN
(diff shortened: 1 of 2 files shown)


brusch:
@dvesh3 we need to fix that in the existing migration `Version20211103055110` as well, so that it can run through without issues 😊

dvesh3:
> @dvesh3 we need to fix that in the existing migration `Version20211103055110` as well, so that it can run through without issues 😊

done  😉

bundles/CoreBundle/Migrations/Version20211103055110.php
```diff
@@ -36,15 +36,11 @@ public function up(Schema $schema): void
 
             if ($schema->hasTable($objectDatastoreTableRelation)) {
                 $this->addSql(
-                    "ALTER TABLE $objectDatastoreTableRelation CHANGE COLUMN 
-                        `type` `type` VARCHAR(50)  NULL DEFAULT NULL ;"
+                    "UPDATE $objectDatastoreTableRelation SET `type` = NULL WHERE `type` = 'object'"
```
blankse:
```suggestion
                    "UPDATE $objectDatastoreTableRelation SET `type` = 'object' WHERE `type` = NULL"
```

blankse, line 39:
```suggestion
                    "UPDATE $objectDatastoreTableRelation SET `type` = 'object' WHERE `type` = NULL"
```

dvesh3: good catch! thanks 👍

dvesh3: I think the correct form is `WHERE type = ''` like using the previous default value.

bundles/CoreBundle/Migrations/Version20211103055110.php
```diff
@@ -36,15 +36,11 @@ public function up(Schema $schema): void
 
             if ($schema->hasTable($objectDatastoreTableRelation)) {
                 $this->addSql(
-                    "ALTER TABLE $objectDatastoreTableRelation CHANGE COLUMN 
-                        `type` `type` VARCHAR(50)  NULL DEFAULT NULL ;"
+                    "UPDATE $objectDatastoreTableRelation SET `type` = 'object' WHERE `type` = NULL"
```
dvesh3:
```suggestion
                    "UPDATE $objectDatastoreTableRelation SET `type` = 'object' WHERE `type` = ''"
```

blankse:
```suggestion
                    "UPDATE $objectDatastoreTableRelation SET `type` = 'object' WHERE `type` = NULL OR `type` =''"
```

brusch:
@abcpremium can you please confirm this patch does solve #11379  Thanks!

abcpremium:
> @abcpremium can you please confirm this patch does solve #11379 Thanks!

It didn't unfortunately, I had to do these changes in vendor/pimcore/pimcore/bundles/CoreBundle/Migrations/Version20211103055110.php

There I did those changes and then it worked.

brusch:
@abcpremium ok, so it did work, just just did the changes of this PR manually to `Version20211103055110` and then you ran it again? This would be ok, because this PR is still open and not part of a release yet 😉

abcpremium:
> @abcpremium ok, so it did work, just just did the changes of this PR manually to `Version20211103055110` and then you ran it again? This would be ok, because this PR is still open and not part of a release yet 😉

Yes exactly, so I think this would solve the issue =)

PR #11386, opened 2022-02-07
Svg sanitization

kingjia90:
Added SVG sanitization on file upload

Follow up to https://github.com/pimcore/pimcore/pull/11208

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -449,7 +449,7 @@ class AssetController extends ElementControllerBase implements KernelControllerE
             if ($sanitizedFileContent) {
                 File::put($sourcePath, $sanitizedFileContent);
             }else{
-                throw new \Exception('Sanitization failed, probably due badly formatted XML');
+                throw new \Exception('SVG Sanitization failed, probably due badly formatted XML. Filename:'.$sourcePath);
             }
         }
 
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -439,20 +439,13 @@ class AssetController extends ElementControllerBase implements KernelControllerE
 
         // SVG Sanitize check
         $fileMimeType = MimeTypes::getDefault()->guessMimeType($sourcePath);
-
         if ($fileMimeType === 'image/svg+xml') {
-            $fileContent = file_get_contents($sourcePath);
-
-            $sanitizer = new Sanitizer();
-            $sanitizedFileContent = $sanitizer->sanitize($fileContent);
-
-            if ($sanitizedFileContent) {
-                File::put($sourcePath, $sanitizedFileContent);
-            }else{
-                throw new \Exception('SVG Sanitization failed, probably due badly formatted XML. Filename:'.$sourcePath);
+            try{
+                File::put($sourcePath,$this->sanitizeSVG($sourcePath));
+            } catch (\Exception $e) {
+                throw $e;
             }
         }
-
 
         $parentId = $request->get('parentId');
         $parentPath = $request->get('parentPath');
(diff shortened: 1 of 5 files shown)


NiklasBr:
Not all assets are added via the `/add-asset` upload path, maybe adding an event listener on [AssetEvents](https://github.com/pimcore/pimcore/blob/10.x/lib/Event/AssetEvents.php) would be better, what do you think?

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -435,6 +436,24 @@ protected function addAsset(Request $request, Config $config)
             throw new \Exception('The filename of the asset is empty');
         }
 
+
+        // SVG Sanitize check
+        $fileExt = File::getFileExtension($filename);
+
+        if ($fileExt === 'svg') {
```
jdreesen: I think you should check the mime-type instead of the file extension here.

jdreesen, line 443: I think you should check the mime-type instead of the file extension here.

kingjia90:
@NiklasBr thank you for the suggestion, i gave it a try but got a problem. 
The PRE_ADD is triggered only for new files, if you upload a file to replace (by overwriting existing one). the [PRE_UPDATE ](https://github.com/pimcore/pimcore/blob/10.x/models/Asset.php#L534) event is triggered instead. 
The problem is that if i put the svg sanitization on PRE_UPDATE, then it would also trigger every time you save any existing svg asset.

NiklasBr:
> The problem is that if i put the svg sanitization on PRE_UPDATE, then it would also trigger every time you save any existing svg asset.

I think [getDataChanged()](https://github.com/pimcore/pimcore/blob/e714878c981d5373b59e3a12c32b3ee488dd04f5/models/Asset.php#L1380-L1386) could be used to check if an existing asset has updated svg data or not. E.g. the *Upload new version* button.

bundles/CoreBundle/EventListener/AssetSanitizationListener.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * @param ElementEventInterface $e
+     */
+    public function sanitizeAsset(ElementEventInterface $e)
+    {
+        $element = $e->getElement();
+        if ($element instanceof Asset) {
+            $assetStream = $element->getStream();
+            if (isset($assetStream)) {
+                $mime = mime_content_type($assetStream);
```
jdreesen: Why don't you use the Symfony Mime component anymore?

```suggestion
                $mime = MimeTypes::getDefault()->guessMimeType($element->getStream());
```

kingjia90: that's because it expects a file path and i don't have the temp file until [there](https://github.com/pimcore/pimcore/blob/e28eab0d25bec6fda4cf09a6ce41fd3dcd9707c7/models/Asset.php#L744), which is 
 [called](https://github.com/pimcore/pimcore/blob/e28eab0d25bec6fda4cf09a6ce41fd3dcd9707c7/models/Asset.php#L561) after [preAdd/preUpdate](https://github.com/pimcore/pimcore/blob/e28eab0d25bec6fda4cf09a6ce41fd3dcd9707c7/models/Asset.php#L536).
Couldn't think of a better solution at the moment, beside creating the temp file "twice" or moving the events to the postAdd/postUpdate (in this case i could just use [getMimeType()](https://github.com/pimcore/pimcore/blob/e28eab0d25bec6fda4cf09a6ce41fd3dcd9707c7/models/Asset.php#L764)).

bundles/CoreBundle/EventListener/AssetSanitizationListener.php
```diff
@@ -41,13 +42,15 @@ public static function getSubscribedEvents()
     public function sanitizeAsset(ElementEventInterface $e)
     {
         $element = $e->getElement();
-        if ($element instanceof Asset) {
+        if ($element instanceof Asset && $element->getType() === 'image') {
```
NiklasBr:
```suggestion
        if ($element instanceof Asset\Image) {
```

bundles/CoreBundle/EventListener/AssetSanitizationListener.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * @param ElementEventInterface $e
+     */
+    public function sanitizeAsset(ElementEventInterface $e)
+    {
+        $element = $e->getElement();
+
+        if ($element instanceof Asset &&
+            $element->getType() === 'image' &&
+            $element->getDataChanged()
```
jdreesen: Is `$dataChanged` also `true` for `PRE_ADD` events?

kingjia90: yes, i confirm this, it is changed to `true` everytime `setStream` is called and on creation of a new Asset it is initiated [there](https://github.com/pimcore/pimcore/blob/73142e5464f60f91313ad06aeff618de721a6df0/models/Asset.php#L380) and then [set](https://github.com/pimcore/pimcore/blob/73142e5464f60f91313ad06aeff618de721a6df0/models/Asset.php#L398)

dvesh3 (approved):

brusch:
Reverted by #11554

PR #11389, opened 2022-02-07
Remove legacy piwik settings check

jremmurd:
Remove legacy piwik settings tab permissions check in SettingsController, as there is no Piwik/Matomo Tab anymore.


dvesh3:
@jremmurd yes, it was a miss. thanks!

PR #11399, opened 2022-02-08
[Messenger] Process certain messages in batches to speed things up

brusch:
In certain cases it could take too long to process the messages, especially when dealing with a lot of updates. 
It seems to be caused by the overhead that's needed to process the messages one by one. Processing them in batches should improve the overall performance. 

See also https://github.com/symfony/symfony/pull/43354

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -106,7 +106,7 @@
     "symfony/intl": "^5.2.0",
     "symfony/lock": "^5.2.0",
     "symfony/mailer": "^5.2.0",
-    "symfony/messenger": "^5.2.0",
+    "symfony/messenger": "^5.4.0",
     "symfony/mime": "^5.2.0",
     "symfony/monolog-bridge": "^5.2.0",
     "symfony/notifier": "^5.2.0",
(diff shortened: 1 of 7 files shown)


lermontex:
Hey @brusch, 

Is it enough to implement the BatchHandlerInterface interface so that messages are batched? It works? It seems that you need to use a special stamp when submitting to enable buffering

> Handlers implementing this interface should expect a new $ack optional argument to be provided when __invoke() is called. **If you don't provide the $ack argument, the message is handled synchronously as usual.** If you provide $ack, __invoke() is expected to buffer the message and its $ack function, and to return the number of pending messages in the batch.

https://symfony.com/blog/new-in-symfony-5-4-messenger-improvements#handle-messages-in-batches

fashxp:
sometimes some messages remain in queue ... especially when there are other transports that also have batch handlers. 
not sure though, if this is relevant in real life. 

besides that, everything looks fine.

fashxp (approved):

PR #11404, opened 2022-02-08
[Feature]: enables fallback locale for admin translations

laurenz-glueck:
Implements fallback local handling for admin translations

## Additional info  

I unfortunately got no response on my question regarding this topic in the Gitter chat. So that's why I have implemented the changes as a proposal to enable the local handling for admin translations.

Depending on the translation domain, I updated the check to see if the translation is empty or for the admin domain equal to the id. 

I have also updated the docs and some translation keys for Ext JS - I unfortunately can not provide all translations ;)

Looking forward to your feedback.

### DIFF
--- a/doc/Development_Documentation/18_Tools_and_Features/25_System_Settings.md
+++ b/doc/Development_Documentation/18_Tools_and_Features/25_System_Settings.md
@@ -12,7 +12,7 @@ Contains settings about changing appearance of Pimcore admin like login screen c
 ## Localization & Internationalization (i18n/l10n) 
 These settings are used in documents to specify the content language (in properties tab), for objects in localized-fields, 
 for shared translations, ... simply everywhere the editor can choose or use a language for the content.
-Fallback languages are currently used in object's localized fields and shared translations.
+Fallback languages are currently used in object's localized fields, shared translations and admin translations.
 
 ## Debug
 
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -280,7 +280,6 @@ class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleA
             $translated = $this->translator->trans($normalizedId, $parameters, $domain, $locale);
         }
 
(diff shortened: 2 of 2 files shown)


brusch:
@laurenz-glueck thanks a lot for your PR - there's one small thing, could you please revert the changes in `bundles/CoreBundle/Resources/translations/de.extended.json` as they are managed with POEditor. 

Thanks!

laurenz-glueck:
@brusch No problem - done ✅

kingjia90:
@laurenz-glueck  Thank you for the PR, but unfortunately, this change for Admin translations doesn't work for me. 
As far as i tested, most of the time it would just fallback to English (if it is set, otherwise is raw text). 
Feel free to share your configs and cases/examples, to see if i missed or misconfigured something. 

However, it seems that the suggested approach may potentially generate other issues, since the fallback languages you set in the `system settings` GUI are exclusive for the `Frontend` and weren't "designed" to directly work along with the `Admin-interface` ones.
As you can see the two lists do not fully match:
image

For admin translation, probably the change should be done by extending and improving something here in the config.yaml `admin_translation_mapping`
https://pimcore.com/docs/pimcore/current/Development_Documentation/Multi_Language_i18n/Admin_Translations.html#page_Override-language-of-admin-translations-in-editmode-of-documents

laurenz-glueck:
@kingjia90 Thanks for your feedback!

My use case was, that I have added `de_AT` for our website and wanted to use our existing `de` translations for the admin translations that we have in our documents. They are used for labels of editable for example.

The difference between frontend and admin languages might be a problem...
If I find some time in the next few days, I will try the `admin_translation_mapping` approach. It might already solve my problem :)

Otherwise I will also do some tests and then get back to you...

kingjia90:
@laurenz-glueck That's great to hear and i am glad i could be of help 😄 

I am closing this PR for the moment (since it cannot be merged) and invite you to open a new git issue (possibly mentioning this PR) with any upcoming follow-up or suggestions.

PR #11407, opened 2022-02-08
Added a check if country key does exist

BentBr:
Resolves the issue of geoip targeting (country) not working for IP's not assigned directly to a country.
Since PHP8 we are failing horribly due to the non-existing offset of the array in line 74 of country.php.
(Before it was a warning and not that bad)

## Additional info  
You can reproduce the issue with certain IP addresses not being listed as part of a county but a contitent (use 66.249.93.204 for testing). You can simply set it in index.php: $_SERVER['REMOTE_ADDR'] = '66.249.93.204';

@pimcore guys :) Thx for your software so far, let's make it even better!


dvesh3:
Hi @BentBr thanks for your PR! Could you please rebase it on branch 10.3? since it is a bug fix. 

please see contribution guidelines here https://github.com/pimcore/pimcore/blob/10.x/CONTRIBUTING.md#contributing-pull-requests

BentBr:
@dvesh3  better? :)

dvesh3:
@BentBr Nope...we have unrelated commits now. Please try again with interactive rebasing pick your changes and drop all other commits. Or you can just close this PR and create a new one based on 10.3 branch.

PR #11408, opened 2022-02-08
Fixed the bug concerning missing data in city array for geoip targeting

BentBr:
Changes in this pull request
Resolves the issue of geoip targeting (country) not working for IP's not assigned directly to a country.
Since PHP8 we are failing horribly due to the non-existing offset of the array in line 74 of country.php.
(Before it was a warning and not that bad)

Additional info
You can reproduce the issue with certain IP addresses not being listed as part of a county but a contitent (use 66.249.93.204 for testing). You can simply set it in index.php: $_SERVER['REMOTE_ADDR'] = '66.249.93.204';

@pimcore guys :) Thx for your software so far, let's make it even better!


dvesh3:
@BentBr thanks!

Issue #11382, opened 2022-02-07
[Improvement] Application Logger File Objects should be stored via flysystem storage


fashxp:
### Improvement description

right now, it is saved directly on filesystem via File:put

brusch:
Assigned @BlackbitNeueMedien

PR #11409, opened 2022-02-08
Allow to open log file object when using custom PIMCORE_PRIVATE_VAR

BlackbitDevs:
Steps to reproduce bug:
1. Create `/config/pimcore/constants.php` with
    ```php
    <?php
    const PIMCORE_PRIVATE_VAR = '/tmp/private-var';
    ```
2. Create `ApplicationLogger` log with `LogFileObject`, e.g. with following script named `test.php`
    ```php
    <?php
    $logger = \Pimcore\Log\ApplicationLogger::getInstance('ProcessManager', true);
    $fileObject = new \Pimcore\Log\FileObject('some interesting data');
    $logger->error('my error message', ['fileObject' => $fileObject]);
    ```
3. Run script with `bin/console pimcore:run-script test.php`
3. Go to Pimcore backend, open Tools > Application Logger
4. Click "Open" in the column "File Object" for the just created log
5. You will get a 403 response -> with this PR you will see the created FileObject file with content `some interesting data`

The reason is that in https://github.com/pimcore/pimcore/blob/fa0162f80154a6c71e8cbc1b94ae3f81a29d9dad/bundles/AdminBundle/Controller/Admin/LogController.php#L240-L250 it is assumed that `PIMCORE_LOG_FILEOBJECT_DIRECTORY` is always inside `PIMCORE_PROJECT_ROOT`. But with above configuration it is not because of https://github.com/pimcore/pimcore/blob/fa0162f80154a6c71e8cbc1b94ae3f81a29d9dad/lib/Bootstrap.php#L199 
So I changed the check to only test if the given `$filePath` is somewhere under `PIMCORE_LOG_FILEOBJECT_DIRECTORY` - what `PIMCORE_PROJECT_ROOT` is, does not matter.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/LogController.php
+++ b/bundles/AdminBundle/Controller/Admin/LogController.php
@@ -238,7 +238,15 @@ class LogController extends AdminController implements KernelControllerEventInte
     {
         $filePath = $request->get('filePath');
 
-        if (!str_starts_with($filePath, PIMCORE_LOG_FILEOBJECT_DIRECTORY)) {
+        if (!filter_var($filePath, FILTER_VALIDATE_URL)) {
+            $filePath = PIMCORE_PROJECT_ROOT.DIRECTORY_SEPARATOR.$filePath;
+            $filePath = realpath($filePath);
+            $fileObjectPath = realpath(PIMCORE_LOG_FILEOBJECT_DIRECTORY);
+        } else {
+            $fileObjectPath = PIMCORE_LOG_FILEOBJECT_DIRECTORY;
+        }
+
+        if (!str_starts_with($filePath, $fileObjectPath)) {
             throw new AccessDeniedHttpException('Accessing file out of scope');
         }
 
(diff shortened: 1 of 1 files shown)


brusch:
@BlackbitNeueMedien This unfortunately opens a vulnerability, if you pass `/path/to/directory/../../../../../etc/passwd` it would go through the check and still show the contents of a file out of scope. 

Btw. should we directly switch to Flysystem? See #11382 Could you implement this? This would be great 😊
By this chance, we should maybe also go away from passing the path to the file-object, maybe it's better to use an ID instead, wdyt?

BlackbitDevs:
Ah you are right. I changed it, now it should not be possible to load a file out of `PIMCORE_LOG_FILEOBJECT_DIRECTORY`.

With Flysystem: Yes, I could implement #11382 but would prefer this PR as quick fix and then later we can fully change to Flysystem... WDYT?

brusch:
Yep, approach is fine for me - just wanted to tell you that this PR needs more attention, then you pushed e1042ea 😅

BlackbitDevs:
Yes, sorry for premature comment. Does work now, doesn't it?

brusch:
LGTM, thanks a lot! 👍

PR #11418, opened 2022-02-10
Tool::getHostname() -> return configured website domain when not in HTTP request context

BlackbitDevs:
Currently for CLI processes `\Pimcore\Tool::getHostname()` always returns null. With this PR in such cases the domain from system settings gets returned.

### DIFF
--- a/lib/Tool.php
+++ b/lib/Tool.php
@@ -395,8 +395,9 @@ final class Tool
     {
         $request = self::resolveRequest($request);
 
-        if (null === $request) {
-            return null;
+        if (null === $request || !$request->getHost() || $request->getHost() === 'localhost') {
+            $domain = \Pimcore\Config::getSystemConfiguration('general')['domain'];
+            return $domain ?: null;
         }
 
         return $request->getHost();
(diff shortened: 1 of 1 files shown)


dvesh3 (commented):
Bug fix? please rebase on 10.3 branch.

BlackbitDevs:
Hmm, actually it is not a bug. But please confirm again to base to 10.3, then I will do it.

dvesh3:
I think it's a bug as `Tool::getHostUrl()` considers the system settings and `Tool::getHostname()` not so there's an inconsistency between two methods of same class. wdyt?

BlackbitDevs:
Ah yes, this explains why I thought that this fallback is already being done - I mixed up `Tool::getHostUrl()` and `Tool::getHostname()`.
Rebased to 10.3.

dvesh3:
@BlackbitNeueMedien we have some unrelated commits here. Could you please check? thanks!

BlackbitDevs:
? Where do you see unrelated commits? In the files tab I only see the changes for this PR.

dvesh3:
> ? Where do you see unrelated commits? In the files tab I only see the changes for this PR.

image

BlackbitDevs:
Hmm ok, I will do the changes again based on a fresh 10.3 branch... one second.

BlackbitDevs:
Now looks better

dvesh3:
@BlackbitNeueMedien thanks!

PR #11420, opened 2022-02-10
[Translations] Only auto-create if currently no translation for key exists -> not on every exception

BlackbitDevs:
We faced a problem where some admin translations got removed. Currently we do not know why but I found a potential problem in https://github.com/pimcore/pimcore/blob/2322eebf98296fc00e21f0ca0054947c1328e184/models/Translation.php#L272-L289

There could also be other exceptions being thrown by `$translation->getDao()->getByKey($id);` - even if there is an existing translation for the given key. In those cases of course we do not want to auto-create an empty translation (and thus overwrite the existing translation).

BTW:
Is it actually correct to return `null` in https://github.com/pimcore/pimcore/blob/2322eebf98296fc00e21f0ca0054947c1328e184/models/Translation.php#L275-L277
regardless of `$returnIdIfEmpty`? Imho this is not correct because if `$returnIdIfEmpty = true` and `$create = false` we should return `$id`, shouldn't we?

### DIFF
--- a/models/Translation.php
+++ b/models/Translation.php
@@ -22,6 +22,7 @@ use Pimcore\Event\TranslationEvents;
 use Pimcore\File;
 use Pimcore\Localization\LocaleServiceInterface;
 use Pimcore\Tool;
+use Symfony\Component\Translation\Exception\NotFoundResourceException;
 
 /**
  * @method \Pimcore\Model\Translation\Dao getDao()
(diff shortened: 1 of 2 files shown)


dvesh3:
@BlackbitNeueMedien yes, looks good to me that we create new translation only on `NotFoundResourceException`.


About the `$returnIdIfEmpty` check, It sets the id for translations which doesn't exist and return it https://github.com/pimcore/pimcore/blob/39293699967f4935a89831f37797d923db703388/models/Translation.php#L294-L302

please note the method can only return `static|null` no strings

BlackbitDevs:
@dvesh3 Yes but if we get to
https://github.com/pimcore/pimcore/blob/39293699967f4935a89831f37797d923db703388/models/Translation.php#L277
we do not get to https://github.com/pimcore/pimcore/blob/39293699967f4935a89831f37797d923db703388/models/Translation.php#L294-L302

I adjusted this PR to return a `Translation` object if no item with given `$id` could be found and `$returnIdIfEmpty == true`.

dvesh3:
@BlackbitNeueMedien This change https://github.com/pimcore/pimcore/pull/11420/commits/3ef37c406784d1fc287cbfb1a0067462c3f37be1 is a BC break as:
Before:
```
$trans = Translation::getByKey('doesntexist'); //returns null
```
After:
```
$trans = Translation::getByKey('doesntexist'); //returns Pimcore\Model\Translation
```
I would still return null based on `!$create & !$returnIdIfEmpty`.

BlackbitDevs:
Ok, fixed the case `!$create && !$returnIdIfEmpty` -> now `null` gets returned in this case (as before)

dvesh3:
LGTM...thanks! 👍

PR #11425, opened 2022-02-10
[Documents] Refactor editables update after saving the document

### LINKED ISSUE #11421: [Bug]: Restore document from recycle bin fails
### Expected behavior

can restore pimcore documents

### Actual behavior

fails due to database error .... 

Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`pimcore`.`documents_editables`, CONSTRAINT `fk_documents_editables_documents` FOREIGN KEY (`documentId`) REFERENCES `documents` (`id`) ON DELETE CASCADE) in /var/www/pimcore/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:117 Stack trace: #0 /var/www/pimcore/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php(117): PDOStatement->execute(NULL)

### Steps to reproduce

create document, delete document, try to restore it from recyclebin

dvesh3:
Resolves #11421

## Additional info  
On saving the document page, the editables are cleaned up and re-interested into the database before the document is updated. 
Similarly, when restoring the document from recycle-bin, the editables are inserted first and then document is inserted into the db.

After introducing the foreign key on `documents_editables` table which references `documents` https://github.com/pimcore/pimcore/pull/11236/files#diff-16a66ecef029997901ca218419d1d956f02bad7ae4361808f4eee946aba41134R79, it is throwing exception as when restoring the document from recyclebin, it first tries to insert the editables with no corresponding entry in `documents` table.

This PR fixes the issue by refactoring the editables update after the document is updated.

### DIFF
--- a/models/Document/PageSnippet.php
+++ b/models/Document/PageSnippet.php
@@ -128,15 +128,14 @@ abstract class PageSnippet extends Model\Document
      */
     protected function update($params = [])
     {
-        // update this
-        parent::update($params);
-
         // update elements
-        $this->getEditables();
+        $editables = $this->getEditables();
         $this->getDao()->deleteAllEditables();
 
-        if (is_array($this->getEditables()) && count($this->getEditables()) > 0) {
-            foreach ($this->getEditables() as $name => $editable) {
+        parent::update($params);
+
+        if (is_array($editables) && count($editables) > 0) {
+            foreach ($editables as $name => $editable) {
                 if (!$editable->getInherited()) {
                     $editable->setDao(null);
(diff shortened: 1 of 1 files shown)


brusch:
@dvesh3 LGTM, I've only changed the order a bit back, so that 
```php
        $editables = $this->getEditables();
        $this->getDao()->deleteAllEditables();
```
is done before 
```php
        parent::update($params);
```

Actually it shouldn't matter but, just to be sure that no editable information is getting lost 😉

PR #11426, opened 2022-02-10
Pimcore Translator Closes Transaction - fixes #11004

### LINKED ISSUE #11004: [Bug]: Pimcore Translator Closes Transaction
### Expected behavior

I create a product or do anything else using a transaction. Inside the transaction I use Pimcore translator. This shouldn't have any effect on my transaction.

### Actual behavior

If a translation entity with a translation key doesn't exists, using translator commits transaction prematurely.

### Steps to reproduce

Open a transaction. Try to translate a key which doesn't exists in the DB yet. Commit the transaction. Receive a DB error "No active transaction".  
DBAL counts transactions correctly, the error comes from the DB in this case.  
According to the MySQL documentation (https://dev.mysql.com/doc/refman/8.0/en/implicit-commit.html), `CREATE TABLE` causes an implicit transaction commit.
[models/Translation/Dao.php:148](https://github.com/pimcore/pimcore/blob/9edd80dadbb1c0eff7f64af81f0382a03b0496a5/models/Translation/Dao.php#L140) runs `CREATE TABLE` every time when an unknown key met.
Maybe instead of running this (heavy and in the most cases useless) query it makes sense to store all translations in the same table and add a field `domain` to filter by domain?
Another possible solution, check for table existence separately from attempt to create and before it - should allow to avoid implicit commits in the most cases.

CountZero1981:
Resolves #11004


brusch:
@CountZero1981 Thanks a lot for your PR, however, I'm afraid that it doesn't solve the issue described in #11004 because `createOrUpdateTable();` is still being called directly. 

A possible solution could be to leave the `save()` method as it as and introduce a new `addToSaveQueue()` where the translations registers itself in the service and the service then calls just `save()` on the translation object.

WDYT?

CountZero1981:
your solution is better, will do it in 1-2 days.

brusch:
@CountZero1981 Thanks, looking forward to it! 👍

brusch:
@CountZero1981 just a quick reminder - we're going to release `10.3.1` tomorrow in the afternoon, so would be really cool to have this ready until then 😊 Thanks in advance!

CountZero1981:
done

CountZero1981:
done

lib/Translation/TranslationEntriesDumper/TranslationEntriesDumper.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Pimcore\Translation\TranslationEntriesDumper;
```
brusch:
```suggestion
namespace Pimcore\Translation;
```

There's no need to put that into a dedicated namespace, or what was the idea behind this? 😊 If there's no specific reason, I'd move it one level up. Thanks!

brusch (changes requested):

lib/Translation/TranslationEntriesDumper/TranslationEntriesDumper.php
CountZero1981: a level above is only Translator and other services have separate namespaces, even Xliff12Escaper in Escaper - though it doesn't implement or extend something. So I did the same. But moved it one level up, done.

brusch:
Works like a charm - just pushed another small micro optimization with b61a649067579131537beb22713fa35876de7a33 😉

brusch:
Thanks a lot! 👌

PR #11432, opened 2022-02-10
asset tmp file to be created on local instead of the flysystem

### LINKED ISSUE #11422: [Improvement]: Create tmp files during asset create on local tmp directory (not in flysystem)
### Improvement description

when using S3 (with versioning activated) as storage, the creation of tmp files results in double the space needed. 

see also https://github.com/pimcore/pimcore/blob/e714878c981d5373b59e3a12c32b3ee488dd04f5/models/Asset.php#L739

kingjia90:
resolves https://github.com/pimcore/pimcore/issues/11422

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -733,30 +733,31 @@ class Asset extends Element\AbstractElement
             if ($this->getDataChanged()) {
                 $src = $this->getStream();
 
-                // Write original data to temp path for writing stream
-                // as original file will be deleted before overwrite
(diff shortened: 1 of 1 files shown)


brusch:
Rebased to `10.3` as I'd classify this as a bug.

brusch:
Changed the approach, because using a local file as described in #11422 isn't the right approach.

brusch:
@kingjia90 plz review 😉

PR #11436, opened 2022-02-11
[Improvement] Show unpublished elements in dependency-list consistently with line-through

ferdinand-it:
The result can be seen in the screenshots below. Same style is also used e.g. in the many-to-many-relationshop-grid.

grafik
grafik


dvesh3:
@ferdinand-it thanks 👍

PR #11443, opened 2022-02-12
Remove duplicate function call

### LINKED ISSUE #11345: [Improvement]: Possible Duplication of lines / Called the same function twice after each other
### Improvement description

In the reverseObjectRelation.js the function this.specificPanel.removeAll() is called twice, once in line 56 as well as 58. 
Is this intentional or actually a mistake?

For clarification: we are talking about:
/vendor/pimcore/pimcore/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/reverseObjectRelation.js
or
https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/reverseObjectRelation.js

matthiashamacher:
Resolves #11345


dvesh3:
@matthiashamacher thanks!

PR #11445, opened 2022-02-14
[Docs] Cache config file needs to be put in local/ folder to get loaded with default config.yaml

BlackbitDevs:
`<Pimcore-Root>/config/cache.yaml` will not get loaded with default `config/config.yaml`, see https://github.com/pimcore/skeleton/blob/7f2ce31352c392dac9ef297b40727cf1ad84cede/config/config.yaml#L1-L2

### DIFF
--- a/doc/Development_Documentation/19_Development_Tools_and_Details/09_Cache/README.md
+++ b/doc/Development_Documentation/19_Development_Tools_and_Details/09_Cache/README.md
@@ -21,7 +21,7 @@ Pimcore uses the `pimcore.cache.pool` Symfony cache pool, you can configure it a
 that the pool supports tags.
 
 ```yaml
-# config/local/cache.yaml
+# config/packages/cache.yaml
 framework:
     cache:
         pools:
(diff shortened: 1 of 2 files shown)


jdreesen:
I don't think it should be put into the `config/local/` folder because that's in the [`.gitignore`](https://github.com/pimcore/skeleton/blob/7f2ce31352/.gitignore#L16). 
I think `config/packages/cache.yaml` would be the right location because everything in this folder gets loaded automatically [via the kernel](https://github.com/pimcore/pimcore/blob/10.x/lib/Kernel.php#L120-L121).

dvesh3:
@BlackbitNeueMedien could you please update the PR as per Jacob's comment https://github.com/pimcore/pimcore/pull/11445#issuecomment-1038842812
 
 Also change here https://github.com/pimcore/pimcore/blob/10.x/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/09_Performance_Guide.md?plain=1#L228
 
 Thanks!

BlackbitDevs:
@jdreesen Yes, I thought the cache config is supposed to be system-specific because it could be that on a dev or local machine you do not have redis installed - but of course we can also put it in `config/packages`. Please confirm again, @dvesh3 , @jdreesen

jdreesen:
I'd still say `config/packages/` is the right location. If you want it to be different per `env` you can also put it into `config/packages/{env}/`.

dvesh3:
Also that's what Symfony is doing in the demo https://github.com/symfony/demo/blob/main/config/packages/cache.yaml

BlackbitDevs:
Is fixed.

dvesh3:
@BlackbitNeueMedien thanks!

PR #11447, opened 2022-02-14
[Admin] Security - Add handler to enable Content Security Policy

dvesh3:
Resolves #

## Additional info  
Add option to enable [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) header for admin interface


bundles/AdminBundle/DependencyInjection/Configuration.php
```diff
@@ -54,6 +54,15 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->end()
                 ->end()
             ->end()
+            ->arrayNode('admin_csp_header')
+                ->addDefaultsIfNotSet()
+                ->children()
+                    ->booleanNode('enabled')
```
brusch: We could use `->canBeEnabled()` here instead.

bundles/CoreBundle/Resources/config/pimcore/default.yaml
```diff
@@ -57,6 +57,7 @@ twig:
         # this is only here for compatibility/dev reasons and may be removed later
         container: '@service_container'
         pimcore_csrf: '@Pimcore\Bundle\AdminBundle\Security\CsrfProtectionHandler'
+        pimcore_nonce: '@Pimcore\Bundle\AdminBundle\Security\ContentSecurityPolicyHandler'
```
brusch: Should we maybe call the variable `pimcore_csp`, as it is more than just the nonce?

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
```diff
@@ -0,0 +1,89 @@
...
+        $this->cspHeaderOptions = $resolver->resolve($cspHeaderOptions);
+    }
+
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults([
+            "default-src" => "'self'",
+            "img-src" => "* data: blob:",
+            "media-src" => "'self' data:",
+            "script-src" => "'self' 'nonce-" . $this->nonce() . "' 'unsafe-inline' 'unsafe-eval' https://code.jquery.com/ https://cdnjs.cloudflare.com/ajax/libs/popper.js/ https://stackpath.bootstrapcdn.com/bootstrap/4.1.1/ http://unpkg.com/react/umd/ http://unpkg.com/axios/dist/ http://unpkg.com/react-dom/umd/",
```
brusch: Where are all the external URLs used? Can only find a couple of them? 
I think we should document that somehow for later reference.

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
```diff
@@ -0,0 +1,89 @@
...
+            "font-src" => "'self' https://cdn.jsdelivr.net/npm/",
+        ]);
+    }
+
+    /**
+     * @return string
+     */
+    public function getCspHeader(): string
+    {
+        $cspHeaderOptions = array_map(function ($k, $v) {
```
brusch, line 79: I think it would be cool and very handy to provide an API for adding additional stuff to the headers. This would be especially useful for bundles that would like to support Pimcore with CSP enabled but using external resources.

brusch (changes requested):

bundles/AdminBundle/DependencyInjection/Configuration.php
dvesh3: done

bundles/CoreBundle/Resources/config/pimcore/default.yaml
dvesh3: done

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
dvesh3: External URLs are kicked out so individual bundle should set the allowed url through `setAllowedUrls` method

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
dvesh3: Added `setCspHeader(string $key, string $value): self` method

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
```diff
@@ -0,0 +1,156 @@
...
+    {
+        return $this->config['admin_csp_header']['enabled'] ? ' nonce="' . $this->nonce() . '"' : '';
+    }
+
+    /**
+     * Generates a random nonce parameter.
+     *
+     * @return string
+     */
+    public function nonce(): string
```
brusch:
```suggestion
    private function getNonce(): string
```
Here the same, wdyt?

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
```diff
@@ -0,0 +1,156 @@
...
+        return isset($this->allowedUrls[$key]) && is_array($this->allowedUrls[$key]) ? implode(' ', $this->allowedUrls[$key]) : '';
+    }
+
+    /**
+     * @param string $key
+     * @param array $value
+     *
+     * @return $this
+     */
+    public function setAllowedUrls(string $key, array $value): self
```
brusch:
```suggestion
    public function addAllowedUrls(string $key, array $value): self
```
I think `add` fits better here.

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
```diff
@@ -0,0 +1,156 @@
...
+        $this->cspHeaderOptions[$key] = $value;
+
+        return $this;
+    }
+
+    /**
+     *
+     * @return string
+     */
+    public function getNonce(): string
```
brusch:
```suggestion
    public function getNonceHtmlAttribute(): string
```
Wdyt, should we name this more appropriate?

bundles/AdminBundle/DependencyInjection/Configuration.php
```diff
@@ -54,6 +54,10 @@ public function getConfigTreeBuilder(): TreeBuilder
                     ->end()
                 ->end()
             ->end()
+            ->arrayNode('admin_csp_header')
+                ->canBeEnabled()
+                ->info('Can be used to enable or disable the Content Security Policy headers.')
+            ->end()
```
brusch, line 93: I think it would be cool here, if you could configure additional URLs.
That would make it very easy for bundles to add their custom stuff (without using a custom compiler pass). 
Example: 
```yaml
pimcore_admin:
    admin_csp_header:
        additional_urls:
            script_src: 
                - "https://oreo.cat/scripts/meow.js"
                - "https://bagheera.cat/*"
```

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
```diff
@@ -0,0 +1,156 @@
...
+        return implode(';' ,$cspHeaderOptions);
+    }
+
+    /**
+     * @param string $key
+     * @param bool $flatten
+     *
+     * @return array|string
+     */
+    public function getAllowedUrls(string $key, bool $flatten = true): array|string
```
brusch:
```suggestion
    private function getAllowedUrls(string $key, bool $flatten = true): array|string
```

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
```diff
@@ -0,0 +1,156 @@
...
+        return $this;
+    }
+
+    /**
+     * @param string $key
+     * @param string $value
+     *
+     * @return $this
+     */
+    public function setCspHeader(string $key, string $value): self
```
brusch, line 126: Should we even provide this method? Or is it better to use `addAllowedUrls()` anyway?

brusch (changes requested):

bundles/AdminBundle/DependencyInjection/Configuration.php
dvesh3: Added configuration node to define additional urls and am internal compiler pass which adds it.

bundles/AdminBundle/Security/ContentSecurityPolicyHandler.php
dvesh3: It will be useful for additional csp directives which are not defined in the default list

doc/Development_Documentation/26_Best_Practice/75_Security_Concept.md
```diff
@@ -25,6 +25,42 @@ it easy and comfortable to keep all dependencies of Pimcore and your project up-
...
+
+In case, you are using third party bundles or custom implementation that extends the admin backend interface with custom views then you would need to use generated nonce string in your scripts.
+If a script does not contain valid a nonce, then it is stopped from being executed wih a warning in console like:
+
+`Refused to execute inline script because it violates the following Content Security Policy directive: ...`
+
+This issue can be resolved either by using Pimcore [Headscript extension](../02_MVC/02_Template/02_Template_Extensions/03_HeadScript.md) or add nonce script to inline scripts as follows:
+
+```twig
+<script {{ pimcore_csp.getNonceHtmlAttribute()|raw }}>
```
jdreesen, line 62: Maybe create a Twig extension for this. As far as I see, this is the only method that should be called in templates. With an extension, it could look like: `<script {{ pimcore_csp_nonce_attr() }}>`

brusch: There will be eventually more in the future, so I'd keep it as it is for now ☺️

PR #11448, opened 2022-02-14
Redirects based on status code

### LINKED ISSUE #11282: [Feature]: Possibility to put a directory or page on error code 410
### Feature description

Hi peps,

I do need the possibility to send for 
- certain pages
- certain paths of my page via regex'es 
the status code 410 (for SEO issues). 

Is there any SEO out there who already has a workaround or a developer who would like to implement the feature. I guess it can be made via various ways, i.e. with an asset or within the siteconfig 

Thanks for your help!

spotbot2k:
Resolves #11282

## Additional info  

This is work in progress but needs to be discussed in advance - do not waste more time if the feature is not needed or the aproach might be wrong.

It adds additional options to the redirects UI/Table to handle redirects based on response status via EventListener.


spotbot2k:
So the current version of this PR is functional, but it adds new columns to the "Redirects" making its UI less clear. For the few people it would not be worth it.

Maybe a option in YAML config would be a better solution.

PR #11458, opened 2022-02-15
[Feature]: Possibility to store version data in database table

### LINKED ISSUE #11340: [Feature]: Possibility to store version data in database table
### Feature description

It should be possible to configure a database connection for storing the versioning information. 
This could be also a completely different to the main connection. I'd suggest to just configure a Doctrine connection. 
It should be also possible to configure a threshold for the max. size, so if a version exceeds this size it get's still dumped on the file system.

mcop1:
resolves #11340


bundles/CoreBundle/DependencyInjection/Compiler/VersionStorageAdapterPass.php
```diff
@@ -0,0 +1,34 @@
...
+use Symfony\Component\DependencyInjection\Reference;
+
+class VersionStorageAdapterPass implements CompilerPassInterface
+{
+    const TAG_NAME = 'pimcore.version.storage.adapter';
+
+    /**
+     * @inheritDoc
+     */
+    public function process(ContainerBuilder $container)
```
brusch: We don't need that compiler pass at all, since the actual adapters have anyway `getStorageType()` implemented, which then can be called by the `DelegateVersionStorageAdapter`.

bundles/CoreBundle/Resources/config/services.yaml
```diff
@@ -257,3 +257,11 @@ services:
...
         shared: false
         public: true
+
+    Pimcore\Model\Version\Adapter\VersionStorageAdapterInterface:
+        public: true
+        alias: Pimcore\Model\Version\Adapter\FileSystemVersionStorageAdapter
+
+    Pimcore\Model\Version\Adapter\FileSystemVersionStorageAdapter:
+        tags:
+            - { name: pimcore.version.storage.adapter, storageType: fs }
```
brusch:
```suggestion
```
Since we don't need the compiler pass, we also don't need the tags here.

models/Version/Adapter/DelegateVersionStorageAdapter.php
```diff
@@ -0,0 +1,134 @@
...
+
+    public function getStorageType(string $metaData,
+                                   mixed $binaryDataStream = null): string {
+        if(empty($this->fallbackAdapter) === false) {
+            $binarySize = 0;
+            $metaDataSize = strlen($metaData);
+
+            if (isset($binaryDataStream) === true) {
+                $stats = fstat($binaryDataStream);
+                $binarySize = $stats['size'];
```
brusch: I'm not sure, whether `size` is available for all types of streams (e.g. S3). 
So I'd go for the storage path here and use `$storage->fileSize($path)` instead.

models/Version/Adapter/DelegateVersionStorageAdapter.php
```diff
@@ -0,0 +1,134 @@
...
+
+class DelegateVersionStorageAdapter implements VersionStorageAdapterInterface
+{
+    public function __construct(protected array $adapters,
+                                protected int $byte_threshold,
+                                protected VersionStorageAdapterInterface $defaultAdapter,
+                                protected VersionStorageAdapterInterface $fallbackAdapter)
+    {
+
+    }
```
brusch:
```suggestion
    private array $adapters = []; 

    public function __construct(protected int $byte_threshold,
                                protected VersionStorageAdapterInterface $defaultAdapter,
                                protected VersionStorageAdapterInterface $fallbackAdapter)
    {
        $this->adapters[$defaultAdapter->getStorageType()] = $defaultAdapter;
        $this->adapters[$fallbackAdapter->getStorageType()] = $fallbackAdapter;
    }
```

brusch (changes requested):

bundles/CoreBundle/Migrations/Version20220214110000.php
```diff
@@ -0,0 +1,47 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20220214110000 extends AbstractMigration
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function getDescription(): string
+    {
+        return 'Add storageType column to database';
```
brusch:
```suggestion
        return 'Add `storageType` column to `version` database table';
```

models/Version.php
```diff
@@ -15,11 +15,10 @@
 
 namespace Pimcore\Model;
 
-use League\Flysystem\UnableToReadFile;
+use GuzzleHttp\Handler\Proxy;
```
brusch:
```suggestion
```
This seems to be obsolete

models/Version/Adapter/DatabaseVersionStorageAdapter.php
```diff
@@ -0,0 +1,161 @@
...
+class DatabaseVersionStorageAdapter implements VersionStorageAdapterInterface
+{
+    CONST versionsTableName = "versionsData";
+
+    protected function getDb() : Connection
+    {
+        return \Pimcore::getContainer()->get($this->database_connection);
+    }
+
+    public function __construct(protected string $database_connection)
```
brusch:
```suggestion
    public function __construct(protected ConnectionInterface $databaseConnection)
```
The Doctrine Bundle registers all configured connections automatically as a service (using the following scheme `doctrine.dbal.[name]_connection`), so we can inject the service directly rather than using the service name.

models/Version/Adapter/DatabaseVersionStorageAdapter.php
```diff
@@ -0,0 +1,161 @@
...
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\Version\Adapter;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\Result;
+use Doctrine\DBAL\ParameterType;
+
+class DatabaseVersionStorageAdapter implements VersionStorageAdapterInterface
```
brusch, line 24:
```suggestion
final class DatabaseVersionStorageAdapter implements VersionStorageAdapterInterface
```

models/Version/Adapter/DatabaseVersionStorageAdapter.php
```diff
@@ -0,0 +1,161 @@
...
+                         string $binaryFileHash = null,
+                         int $binaryFileId = null): void
+    {
+
+        if(isset($binaryDataStream) === true &&
+            isset($binaryFileId) === false) {
+            $contents = stream_get_contents($binaryDataStream);
+        }
+
+        $this->getDb()->executeStatement("insert into " . self::versionsTableName . "(id, cid, ctype, metaData, binaryData) values (:id, :cid, :ctype, :metaData, :binaryData)",
```
brusch: We should use `$this->getDb()->insert()` here instead.

models/Version/Adapter/DatabaseVersionStorageAdapter.php
```diff
@@ -0,0 +1,161 @@
...
+     * @throws \Doctrine\DBAL\Exception
+     */
+    protected function loadData(int    $id,
+                                int    $cId,
+                                string $cType,
+                                bool   $binaryData = false): mixed {
+
+        $data = null;
+        $dataColumn = $binaryData ? 'binaryData' : 'metaData';
+        $resultSet = $this->getDb()->executeQuery("SELECT " . $dataColumn . " FROM " . self::versionsTableName . " WHERE id = :id AND cid = :cid and ctype = :ctype",
```
brusch: We should use `$this->getDb()->fetchOne()` here directly, saves some lines of code😉

models/Version/Adapter/DatabaseVersionStorageAdapter.php
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    public function delete(int $id,
+                           int $cId,
+                           string $cType,
+                           string $storageType,
+                           bool $isBinaryHashInUse,
+                           int $binaryFileId = null): void
+    {
+        $this->getDb()->executeStatement("delete from " . self::versionsTableName . " where id=:id and cid=:cid and ctype=:ctype",
```
brusch: We should use `$this->getDb()->delete()` here directly.

models/Version/Adapter/DelegateVersionStorageAdapter.php
```diff
@@ -0,0 +1,130 @@
...
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\Version\Adapter;
+
+class DelegateVersionStorageAdapter implements VersionStorageAdapterInterface
+{
+    private array $adapters = [];
+    public function __construct(protected int $byte_threshold,
```
brusch:
```suggestion
    public function __construct(protected int $byteThreshold,
```
Variable names should always be camelCase, snake_case is only used by Symfony configs.

brusch (changes requested):

doc/Development_Documentation/18_Tools_and_Features/01_Versioning.md
```diff
@@ -22,6 +22,81 @@ You can configure the versioning behavior in the ![Settings](../img/Icon_setting
...
+Therefore, you are able to provide a connection to a completely separate database which may contains only the version data.    
+
+```yml
+Pimcore\Model\Version\Adapter\VersionStorageAdapterInterface:
+    public: true
+    alias: Pimcore\Model\Version\Adapter\DatabaseVersionStorageAdapter
+
+Pimcore\Model\Version\Adapter\DatabaseVersionStorageAdapter:
+    arguments:
+        $database_connection: doctrine.dbal.versioning
```
brusch:
```suggestion
        $database_connection: '@doctrine.dbal.versioning_connection'
```

brusch (changes requested):

bundles/CoreBundle/Migrations/Version20220214110000.php
```diff
@@ -0,0 +1,47 @@
...
+        if ($schema->getTable('versions')->hasColumn('storageType') === false) {
+            $this->addSql("ALTER TABLE `versions` ADD COLUMN `storageType` varchar(5) DEFAULT NULL;");
+            $this->addSql("update `versions` set storageType = 'fs'");
+        }
+    }
+
+    public function down(Schema $schema): void
+    {
+        if ($schema->getTable('versions')->hasColumn('storageType')) {
+            $this->addSql('ALTER TABLE `versions` DROP COLUMN `storageType`;');
```
fashxp, line 44: should we really drop that column? otherwise we at least would keep the information where versions are stored in case of the update is redone later on.

models/Version.php
```diff
@@ -226,41 +236,32 @@ public function save()
             $dataString = $data;
         }
 
-        $isAssetFile = false;
-        if ($data instanceof Asset && $data->getType() != 'folder') {
-            $isAssetFile = true;
-
+        if($data instanceof Asset) {
```
fashxp: why `$data->getType() != 'folder'` was removed as condition there?

models/Version/Adapter/DatabaseVersionStorageAdapter.php
```diff
@@ -0,0 +1,144 @@
...
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\Version\Adapter;
+
+use Doctrine\DBAL;
+
+final class DatabaseVersionStorageAdapter implements VersionStorageAdapterInterface
+{
+    CONST versionsTableName = "versionsData";
```
fashxp, line 26: should we automatically create the table it not existing? or do we really put that into the hands of the developer who sets it up?

fashxp (changes requested):
we should definitely add tests for that.

bundles/CoreBundle/Migrations/Version20220214110000.php
brusch: We do a proper rollback also in all other places, so I'd do it here as well (dropping columns, tables, ...). It's very unlikely that there will be that many versions in between of an upgrade and rollback where also the config gets changed to a different storage backend. If you do not change the adapter, nothing changes anyway when rolling back 😉

models/Version/Adapter/DatabaseVersionStorageAdapter.php
brusch: This we could do by catching `TableNotFoundException` and then running the DDL. 
But we need to be careful: we need to rollback the active db transaction, then run the CREATE TABLE and then throw an exception, so that the transaction get's restarted.

models/Version.php
mcop1: Added the type check again

brusch:
@mcop1 todo: adding tests

models/Version/Adapter/VersionStorageAdapterInterface.php
```diff
@@ -0,0 +1,109 @@
...
+     * @return void
+     */
+    public function save(int $id,
+                         int $cId,
+                         string $cType,
+                         string $storageType,
+                         string $metaData,
+                         mixed $binaryDataStream = null,
+                         string $binaryFileHash = null,
+                         int $binaryFileId = null) : void;
```
brusch:
```suggestion
    public function save(Version $version) : void;
```
@mcop1 I'm wonderning whether we should change the signatures of all the methods to just use `Version`, to make it less extensive, or was there a particular reason to specify all the individual parameters?

brusch (changes requested):

models/Version/Adapter/VersionStorageAdapterInterface.php
mcop1: @brush just wanted to keep most of the pimcore dependencies and logic separated from the storage adapter. If we pass the `Version` object, we have to set some properties inside the save method of the storageadapter or double check if the data is an asset:
https://github.com/pimcore/pimcore/pull/11458/files#diff-440c14d8343f7c3272fb27e0ec1577e5178eb73eef4f165f0ead5f0326794973R239

On the other hand just passing `Version` would be less extensive indeed.

brusch: Yep understand 😊 But since those 2 things are anyway deeply connected, I'd go for `Version` instead and maybe have some helper methods in a trait or so if it makes sense. @fashxp @dvesh3 wdyt?

dvesh3: I would really use the property params if this interface is or will be used by other types. If this is specific to `version` then i don't see harm in passing the object itself.

brusch: I'd say it's completely specific to `Version` right? I don't see any other use for it 🤔

bundles/CoreBundle/Migrations/Version20220214110000.php
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public function getDescription(): string
+    {
+        return 'Add `storageType` column to `version` database table';
+    }
+
+    public function up(Schema $schema): void
+    {
+        if ($schema->getTable('versions')->hasColumn('storageType') === false) {
+            $this->addSql("ALTER TABLE `versions` ADD COLUMN `storageType` varchar(5) DEFAULT NULL;");
```
blankse, line 36: Why is this different to the declaration in install.sql? I got errors now:
image

passioneight:
<s>Just updated to 10.4.3 - got the following error when trying to open a specific snippet:

image

Did anyone experience this already?
</s>

Nevermind, I imported the prod database a few seconds ago and forgot about this: https://pimcore.com/docs/pimcore/current/Development_Documentation/Administration_of_Pimcore/Cleanup_Data_Storage.html#page_Reduce-the-Amount-of-Versioning-Steps-of-Assets-Objects-and-Documents

> Pimcore stores the meta-data for version in the database, however, the data itself is stored on the filesystem (var/versions) as compressed files (gzip) to keep the database as lean as possible.

Issue #9029, opened 2021-05-05
[Assets] Aspect-Ratio of folder preview images - wanted? 


fashxp:
![image](https://user-images.githubusercontent.com/8792145/117198199-018dfd80-ade9-11eb-8de3-d4be8f100311.png)

brusch:
Actually this was wanted by me, so that you can see the entire contents of an image in folder preview an nothing is cut off. I've heard sometimes that very similar contents are together in one folder, and if we cut off contents it looks like 2 identical images are in there. 
I personally don't mind, it's just a preview.

thgross:
Or we could use a frame transformation for the previews to keep the aspect ratio and don't cut off anything.

brusch:
Would be possible as well, but then you'll loose the tile look & feel 🤔

PR #11459, opened 2022-02-15
Asset tree optimization

### LINKED ISSUE #11446: [Improvement]: [Asset] optimize tree requests regarding thumbnails
### Improvement description

When using remote storages this call to `exists()` can slow down the tree-requests significantly: 
https://github.com/pimcore/pimcore/blob/d34b60574994ea91db51758603c056b3033a3e52/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php#L755 

Possible improvements: 
- remove `exists()` check
- extend delay to show thumbnails 
- check existence of file in thumbnail delivery action
- if not exists, throw error
- handle error in JS and do not display preview (maybe using `onerror` ?)

brusch:
- get rid of `<iframe>` solution, got obsolete by #7900 / simplification
- return `404` if there's no preview available, show preview only on success
- do not generate previews on request for the tree - check for existence
- do not check the existence of a preview thumbnail when generating the tree config (remote storage performance!) 

Resolves #11446

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -741,7 +741,7 @@ class AssetController extends ElementControllerBase implements KernelControllerE
             $tmpAsset['expanded'] = !$hasChildren;
             $tmpAsset['loaded'] = !$hasChildren;
             $tmpAsset['permissions']['create'] = $asset->isAllowed('create');
-            $tmpAsset['thumbnail'] = $this->getThumbnailUrl($asset);
+            $tmpAsset['thumbnail'] = $this->getThumbnailUrl($asset, ['origin' => 'treeNode']);
         } else {
             $tmpAsset['leaf'] = true;
             $tmpAsset['expandable'] = false;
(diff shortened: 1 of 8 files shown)


brusch:
Missing: do add thumb generation to message queue if it doesn't exist

kingjia90:
looks good  to me and works great, it feels much quicker.




Just two thing that i have noticed during the tests, not strictly related to this PR changes.

first one is more related to #7900:
the folder preview doesn't update when you delete an asset, because [this](https://github.com/pimcore/pimcore/blob/b6d4a89e88d5ae761a4311e8594e004c6fcf38a5/models/Asset/Folder.php#L130) line is not able to detect the modificationDate change (since the entry is directly removed from the `assets` table). It also doesn't update any of the parents' folder preview (from bottom to top).

The second one: 
It happened that i had some unusual sized image (extralong, extrawide, 360°-ready for past PR review) and they look bad  in the folder preview, because [imagecopyresampled](https://github.com/pimcore/pimcore/blob/b6d4a89e88d5ae761a4311e8594e004c6fcf38a5/models/Asset/Folder.php#L174) tend to shrink and stretch and adds "random" black bars, not sure if it's intended, but probably this is for a new issue.

brusch:
@kingjia90 thanks a lot for reviewing! 

Regarding your feedback: 
1. valid ... I'll push a fix
2. see #9029 😉

PR #11464, opened 2022-02-16
[DataObjects] Localizedfields - throw meaningful exception when non existing attribute is requested in getLocalizedValue resolves

### LINKED ISSUE #11456: [Bug]: Localized fields in blocks
### Expected behavior

Localized fields in blocks do not throw exceptions, as described in the docs.

See 
* https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Data_Types/Blocks.html  
* Get Values of Localized Block Entries:

### Actual behavior

Getting data from localized fields in blocks throw exceptions 

image

### Steps to reproduce

Follow the docs 
```

        $object = DataObject\BlockClass::getById(48);
        /** @var  $blockData DataObject\ClassDefinition\Data\Block */
        $blockItems = $object->getBlockElement1();
        /** @var  $firstBlockItem DataObject\Data\BlockElement */
        $firstBlockItem = $blockItems[0];
        $localizedfields = $firstBlockItem["localizedfields"]->getData();
        // after that use it as you would to it with `Pimcore\Model\DataObject\LocalizedField::getLocalizedValue`
```

dvesh3:
Resolves #11456 

## Additional info  
please see https://github.com/pimcore/pimcore/issues/11456#issuecomment-1041346281

### DIFF
--- a/models/DataObject/Localizedfield.php
+++ b/models/DataObject/Localizedfield.php
@@ -462,7 +462,9 @@ final class Localizedfield extends Model\AbstractModel implements
      * @param string|null $language
      * @param bool $ignoreFallbackLanguage
      *
-     * @return mixed
+     * @return mixed|null
+     *
+     * @throws \Exception
      */
     public function getLocalizedValue(string $name, string $language = null, bool $ignoreFallbackLanguage = false)
     {
@@ -559,7 +561,7 @@ final class Localizedfield extends Model\AbstractModel implements
         }
 
         //TODO Pimcore 11: remove method_exists BC layer
-        if ($fieldDefinition instanceof PreGetDataInterface || ($fieldDefinition && method_exists($fieldDefinition, 'preGetData'))) {
+        if ($fieldDefinition instanceof PreGetDataInterface || method_exists($fieldDefinition, 'preGetData')) {
             if (!$fieldDefinition instanceof PreGetDataInterface) {
                 trigger_deprecation('pimcore/pimcore', '10.1', sprintf('Usage of method_exists is deprecated since version 10.1 and will be removed in Pimcore 11.' .
                     'Implement the %s interface instead.', PreGetDataInterface::class));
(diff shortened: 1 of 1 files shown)


models/DataObject/Localizedfield.php
```diff
@@ -472,6 +472,10 @@ public function getLocalizedValue(string $name, string $language = null, bool $i
         $context = $this->getContext();
         $fieldDefinition = $this->getFieldDefinition($name, $context);
 
+        if (!$fieldDefinition instanceof ClassDefinition\Data) {
+            throw new Model\Exception\NotFoundException(sprintf('Field "%s" does not exist in localizedfields', $name));
```
blankse, line 479: Please add this also to phpdoc

dvesh3: done 😉

models/DataObject/Localizedfield.php
```diff
@@ -462,7 +462,9 @@ private function loadLazyField(Model\DataObject\ClassDefinition\Data $fieldDefin
      * @param string|null $language
      * @param bool $ignoreFallbackLanguage
      *
-     * @return mixed
+     * @return mixed|null
+     *
+     * @throws \Exception
```
blankse, line 467:
```suggestion
     * @throws \Exception
     * @throws Model\Exception\NotFoundException
```

dvesh3: wdyt?
image

blankse: I want to know which exceptions can be thrown. From wich line came the \Exception? Should this be catched? You can additional add `@throws \Exception` also if not. But it would be good to have only named exceptions. So I can catch it individuel... We can also add a comment when which exception is thrown.

dvesh3: Let's add additional `@throws \Exception` for now.

PR #11469, opened 2022-02-16
[GDPR Data Extractor] DataObject - Allow extending dataObjects provider with custom search and download url

dvesh3:
Resolves #


bundles/AdminBundle/GDPR/DataProvider/Exporter.php
```diff
@@ -145,7 +145,7 @@ public static function doExportObject(Concrete $object, &$result = [])
 
             if ($fd instanceof Data\Fieldcollections) {
```
blankse: Fieldcollections can also be null
```suggestion
            if ($fd instanceof Data\Fieldcollections && $value instanceof Fieldcollection) {
```

PR #11470, opened 2022-02-17
[Improvement] Add attribute level placeholders support in index-service config

jremmurd:
This PR adds support for placeholder on attribute level to the index-service config.

**Example**:

```
    pimcore_ecommerce_framework:
        index_service:
            tenants:
                 INT:
                     name_en: &name
                        locale: '%%locale%%'
                        filtergroup: string
                 BE:
                     name_fr:
                        <<: *name
                        placeholders:
                            '%%locale%%': fr
                    name_de:
                        <<: *name
                        placeholders:
                            '%%locale%%': de
                    name_nl:
                        <<: *name
                        placeholders:
                            '%%locale%%': nl
```


dvesh3 (changes requested):
Please add an example here https://github.com/pimcore/pimcore/blob/10.x/doc/Development_Documentation/10_E-Commerce_Framework/05_Index_Service/01_Product_Index_Configuration/README.md?plain=1#L62

dvesh3:
@jremmurd thanks!

PR #11479, opened 2022-02-17
Fixed guessing mime types on temp files

rubanooo:
Im using the flysystem with the google cloud storage adapter. When you copy a folder or file to another place the following error is thrown: `The "php://temp" file does not exist or is not readable.`  This is because of the new SVG sanitation https://github.com/pimcore/pimcore/pull/11386 

I have added a small check to see if the current stream is a temp stream, since we can't read the mime type of a temp stream.

### DIFF
--- a/bundles/CoreBundle/EventListener/AssetSanitizationListener.php
+++ b/bundles/CoreBundle/EventListener/AssetSanitizationListener.php
@@ -47,11 +47,14 @@ class AssetSanitizationListener implements EventSubscriberInterface
 
             if (isset($assetStream)) {
                 $streamMetaData = stream_get_meta_data($assetStream);
-                $mime = MimeTypes::getDefault()->guessMimeType($streamMetaData['uri']);
+                
+                if ($streamMetaData['uri'] !== 'php://temp') {
+                    $mime = MimeTypes::getDefault()->guessMimeType($streamMetaData['uri']);
 
-                if ($mime === 'image/svg+xml') {
-                    $sanitizedData = $this->sanitizeSVG(stream_get_contents($assetStream));
-                    $element->setData($sanitizedData);
+                    if ($mime === 'image/svg+xml') {
+                        $sanitizedData = $this->sanitizeSVG(stream_get_contents($assetStream));
+                        $element->setData($sanitizedData);
+                    }
                 }
             }
         }


bundles/CoreBundle/EventListener/AssetSanitizationListener.php
```diff
@@ -47,11 +47,14 @@ public function sanitizeAsset(ElementEventInterface $e)
 
             if (isset($assetStream)) {
                 $streamMetaData = stream_get_meta_data($assetStream);
-                $mime = MimeTypes::getDefault()->guessMimeType($streamMetaData['uri']);
+
```
kingjia90, line 50: thank you for the PR, it would fix the specific case (which i was unable to reproduce with S3 adapter + Minio) but won't cover any other possible problem of `uri` field not returning a physical file, and it is actually skipping the sanitization (even though it should be fine for a copy/paste).

what about: 
```
if (file_exists($streamMetaData['uri'])){
    $mime = MimeTypes::getDefault()->guessMimeType($streamMetaData['uri']);}
else{
    $mime = mime_content_type($assetStream);
}

if ($mime === 'image/svg+xml') {
    $sanitizedData = $this->sanitizeSVG(stream_get_contents($assetStream));
    $element->setData($sanitizedData);
}
```
even though it was sort of "disrecommended" [here](https://github.com/pimcore/pimcore/pull/11386/commits/b5ea47670641ea14963dfc45c0ea937de56de742#r801659837) before i found out that could get it using stream_get_meta_data, would still be a quicker approach than write the stream in a temp file, read it with guessMimeType and delete the temp file. 
Let me know if this approach works as well

rubanooo: Looks good to me, I have tested it in our repository and it seems to work!

brusch: Maybe we should use `stream_is_local()` instead of `file_exists(...)`

brusch (changes requested):

bundles/CoreBundle/EventListener/AssetSanitizationListener.php
kingjia90: i just checked and to my surprise, `var_dump(stream_is_local("php://temp"));` returns `bool(true)`

kingjia90: but it should work for `stream_is_local($assetStream)` theoretically

brusch: Ok 😅 But I'm fine with the current approach as well ... so no problem at all 😉

kingjia90:
@rubanooo could you please do this one more test, to see if it works with `stream_is_local($assetStream)` instead of `file_exists`? if yes, then please push the changes with stream_is_local, otherwise let's go for file_exists.

Looking forward to have your final commit and i will be glad to merge it before the next release (Wednesday). 
Thank you in advance for the support 😃

brusch:
Closed in favor of #11554

PR #11481, opened 2022-02-17
[Bug]: Inconsistent getter behaviour when getting values from object brick in preSave listener

### LINKED ISSUE #11181: [Bug]: Inconsistent getter behaviour when getting values from object brick in preSave listener
### Expected behavior

No matter what fields user edits on the UI, the backend getters of object should behave in consistent manner. 

### Actual behavior

There are differences in getter behaviour, depending on whether an autosave was triggered during editing.

### Steps to reproduce

Create an object with a object brick, that has a ManyToManyObjectRelation field (object→object brick→relation field). Add one other field (e.g. input) on object brick level. Add a relation and fill in the input and save the instance. Add a listener that will "catch" this object type on preSave.

Behaviour 1:
Go to created object, remove the relation and save object. In the listener, if you call: `$relation = $object->getTestBrick()->getTestRelation()` the value of `$relation` will be `null`. This I would call expected behaviour.

Behaviour 2:
Go to created object, and edit the input field - the edit will trigger an autosave. Than, remove the relation and click save. In the listener, if you now try to call  `$relation = $object->getTestBrick()->getTestRelation()`, value of `$relation` will be a relation object. This in not expected.

This difference in behaviour did not occur in 10.2.1. I suspect it may be related to this changes https://github.com/pimcore/pimcore/pull/10820 and this issue https://github.com/pimcore/pimcore/issues/10786

mcop1:
resolves #11181

### DIFF
--- a/models/DataObject/Objectbrick.php
+++ b/models/DataObject/Objectbrick.php
@@ -185,7 +185,7 @@ class Objectbrick extends Model\AbstractModel implements DirtyIndicatorInterface
         $getters = $this->getBrickGetters();
 
         foreach ($getters as $getter) {
-            $brick = $this->$getter();
+            $brick = $this->$getter(true);
 
             if ($brick instanceof Objectbrick\Data\AbstractData) {
                 if ($brick->getDoDelete()) {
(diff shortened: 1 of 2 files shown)


fashxp:
- [ ] rebase to 10.x 
- [ ] add upgrade note
- [ ] add migration for regenerating class/brick definitions

mcop1:
See https://github.com/pimcore/pimcore/pull/11707

PR #11487, opened 2022-02-18
[DataObject] Classdefinition - Fix available sites selection in url slug field

### LINKED ISSUE #10604: [Bug]: Domain input for URL Slug is hidden since Pimcore X

dvesh3:
Related to https://github.com/pimcore/pimcore/issues/10604#issuecomment-1044176324

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/urlSlug.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/urlSlug.js
@@ -114,7 +114,7 @@ pimcore.object.classes.data.urlSlug = Class.create(pimcore.object.classes.data.d
             new Ext.ux.form.MultiSelect({
                 fieldLabel: t("available_sites"),
                 name: "availableSites",
-                value: availableSites,
+                value: datax.availableSites,
                 displayField: "domain",
                 valueField: "id",
                 store: sitesStore,


NiklasBr:
I don't think this seems to fix the issue linked, it is only ~hiding the~ fixing another problem.

dvesh3:
> I don't think this seems to fix the issue linked, it is only ~hiding the~ fixing another problem.

yes, the description was wrongly typed. Now fixed it by stating "Related to".

NiklasBr:
Please revert this. it made #10604 worse

dvesh3:
@NiklasBr how?

NiklasBr:
The entire field disappeared. Removing this PR manually in 10.3.3 at least restores some of the fields.

dvesh3:
This change is for classdefinition, so it shows selected values on reload. I don't understand how it affects the fields visibility 😕

NiklasBr:
In 10.2.10 the fields were visible and the URL Slugs could be edited in the admin interface, but after updating to 10.3.2 and 10.3.3 respectively this no longer works. I don't understand it either, which is why I need your documentation to fix it the way you want it fixed, but at least reverting this change locally improves situation so that at least some parts works.

dvesh3:
Sorry, this change does not affect the fields visibility on edit view. Infact fixes the multiselect selection view in classdefinition so we don't need to revert this.

PR #11500, opened 2022-02-21
Delete temporary files when calling Pimcore::collectGarbage()

BlackbitDevs:
For long-running processes from a lot of places `\Pimcore::collectGarbage()` gets called. This clears up some resources but it currently does not remove temporary files which got created by 
https://github.com/pimcore/pimcore/blob/e05650390cb6e515fe3f37afeb5ef995eda10f80/lib/Helper/TemporaryFileHelperTrait.php#L56-L84

This may become a problem for jobs with lots of such temporary files like https://github.com/pimcore/pimcore/blob/e05650390cb6e515fe3f37afeb5ef995eda10f80/lib/Maintenance/Tasks/LowQualityImagePreviewTask.php#L57-L82

In worst case this maintenance task will copy *all* asset files to `PIMCORE_SYSTEM_TEMP_DIRECTORY` and will only delete them when the whole script is finished. This means that you need at least the same amount of hard disk space as the assets already consume.

With this PR temporary files will get deleted when calling `\Pimcore::collectGarbage()`.

### DIFF
--- a/doc/Development_Documentation/05_Objects/05_External_System_Interaction.md
+++ b/doc/Development_Documentation/05_Objects/05_External_System_Interaction.md
@@ -112,7 +112,7 @@ You can pass in a string instead of an array if you only want to supply a single
 
 ## Temporary files
 
-Temporary files which get created during processing are usually deleted when the current process finishes. In some cases this may be a problem when thousands of temporary files get created during processing of the current request. If you want to clear temporary files before the script ends, you can call
+Temporary files which get created during processing are usually deleted when the current process finishes. In some cases this may be a problem when thousands of temporary files get created within a process. If you want to clear temporary files before the script ends, you can call
 ```php
 \Pimcore::deleteTemporaryFiles();
 ```
--- a/doc/Development_Documentation/05_Objects/05_External_System_Interaction.md
+++ b/doc/Development_Documentation/05_Objects/05_External_System_Interaction.md
@@ -86,8 +86,7 @@ For simple CSV exports, Pimcore backend interface provides a CSV export function
  
  
 ## Memory Issues
-If you're using / creating very much objects you should call the Pimcore garbage collector after several cycle to 
-prevent memory issues
+If you're using / creating very much objects you should call the Pimcore garbage collector after several cycles to prevent memory issues
 
 ```php
 // just call this static method
@@ -110,3 +109,10 @@ $longRunningHelper->removePimcoreRuntimeCacheProtectedItems(["myVeryImportantKey
 
 ```
 You can pass in a string instead of an array if you only want to supply a single key.
+
+## Temporary files
+
+Temporary files which get created during processing are usually deleted when the current process finishes. In some cases this may be a problem when thousands of temporary files get created during processing of the current request. If you want to clear temporary files before the script ends, you can call
+```php
+\Pimcore::deleteTemporaryFiles();
+```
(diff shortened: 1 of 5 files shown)


lib/Helper/LongRunningHelper.php
```diff
@@ -153,4 +157,18 @@ protected function resolveOptions(string $method, array $options)
 
         return [];
     }
+
+    /**
+     * Register a temp file which will be deleted on next call of cleanUp()
+     */
```
brusch:
```suggestion
    /**
     * @internal
     * Register a temp file which will be deleted on next call of cleanUp()
     */
```

brusch (changes requested):

lib/Helper/TemporaryFileHelperTrait.php
```diff
@@ -75,6 +75,7 @@ protected static function getTemporaryFileFromStream($stream, bool $keep = false
         fclose($dest);
 
         if (!$keep) {
+            LongRunningHelper::addTmpFilePath($tmpFilePath);
```
fashxp: Isn't that dangerous as the file still might be needed after calling `collectGarbage`? 
At least it is a potential change in behavior I would say.

brusch: Maybe we should make `deleteTemporaryFiles()` public and do not call it implicitly when calling `cleanup()`. 
Then we don't have to deal with unexpected behavior and it's up to the dev whether to clean tmp files or not (the possibility to do so, makes definitely sense, because currently it's not possible at all).

brusch: @BlackbitNeueMedien can you please update the PR accordingly? Thanks in advance 😊

BlackbitDevs: Done. Have also created `Pimcore::deleteTemporaryFiles()` (easier for developers than to deal with the service container).

doc/Development_Documentation/05_Objects/05_External_System_Interaction.md
```diff
@@ -110,3 +109,10 @@ $longRunningHelper->removePimcoreRuntimeCacheProtectedItems(["myVeryImportantKey
 
 ```
 You can pass in a string instead of an array if you only want to supply a single key.
+
+## Temporary files
+
+Temporary files which get created during processing are usually deleted when the current process finishes. In some cases this may be a problem when thousands of temporary files get created during processing of the current request. If you want to clear temporary files before the script ends, you can call
```
brusch:
```suggestion
Temporary files which get created during processing are usually deleted when the current process finishes. In some cases this may be a problem when thousands of temporary files get created within a process. If you want to clear temporary files before the script ends, you can call
```

lib/Helper/LongRunningHelper.php
```diff
@@ -153,4 +156,19 @@ protected function resolveOptions(string $method, array $options)
 
         return [];
     }
+
+    /**
+     * @internal
+     * Register a temp file which will be deleted on next call of cleanUp()
+     */
+    public static function addTmpFilePath(string $tmpFilePath) {
```
brusch: I'd prefer to not have this static as well as the `$tmpFilePaths` property, then it's easier to test and consistent with the other stuff.

brusch (changes requested):
We're almost there ☺️

lib/Helper/LongRunningHelper.php
BlackbitDevs: Done

PR #11501, opened 2022-02-22
[Bug][Assets] Uploading invalid zip just hangs without error message

### LINKED ISSUE #11393: [Bug][Assets] Uploading invalid zip just hangs without error message
### Expected behavior

show error message that zip file is not valid

### Actual behavior

import just hangs

image

### Steps to reproduce

select some other file than a zip file within zip upload

mcop1:
resolves #11393

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -2231,8 +2231,7 @@ class AssetController extends ElementControllerBase implements KernelControllerE
         else {
             return $this->adminJson([
                 'success' => false,
-                'message' => $this->trans('could_not_open_zip_file', [], 'admin'),
-                'ZipArchiveReturnCode' => $retCode
+                'message' => $this->trans('could_not_open_zip_file', [], 'admin')
             ]);
         }
     }
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -2199,7 +2199,8 @@ class AssetController extends ElementControllerBase implements KernelControllerE
         copy($_FILES['Filedata']['tmp_name'], $zipFile);
 
         $zip = new \ZipArchive;
-        if ($zip->open($zipFile) === true) {
+        $retCode = $zip->open($zipFile);
+        if ($retCode === true) {
             $jobAmount = ceil($zip->numFiles / $filesPerJob);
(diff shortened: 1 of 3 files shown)


bundles/CoreBundle/Resources/translations/en.extended.json
```diff
@@ -671,5 +671,6 @@
   "max_characters": "Maximum Characters",
   "sites": "Sites",
   "show_hide_areablock": "Show or Hide on frontend",
-  "asset_type_changed": "Asset type was changed in the meantime. Please reload the admin interface to see the latest changes."
+  "asset_type_changed": "Asset type was changed in the meantime. Please reload the admin interface to see the latest changes.",
+  "could_not_open_zip_file": "The uploaded zip file could not be opened."
```
brusch: This should be defined in `en.json` as `en.extended.json` is intended to be used by admin-only stuff.

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -2215,18 +2216,25 @@ public function importZipAction(Request $request)
...
+            ]);
 
-        return new Response($responseJson);
+            return new Response($responseJson);
+        }
+        else {
+            return $this->adminJson([
+                'success' => false,
+                'message' => $this->trans('could_not_open_zip_file', [], 'admin'),
+                'ZipArchiveReturnCode' => $retCode
```
```suggestion
```

brusch (changes requested):

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -2215,18 +2216,24 @@ public function importZipAction(Request $request)
...
+            ]);
 
-        return new Response($responseJson);
+            return new Response($responseJson);
+        }
+        else {
+            return $this->adminJson([
+                'success' => false,
+                'message' => $this->trans('could_not_open_zip_file', [], 'admin')
+            ]);
```
kingjia90, line 2235: 'admin' domain needs to be removed, as it changed from `en.extended.json` to `en.json`.
Other than that, everything works fine, ready to merge😃

mcop1: Does the message get displayed properly on your side? Because it shouldn´t make a difference, since 'admin' is the default value for the domain argument. But I agree, it would be more "clean".

kingjia90: ahh, it works, like you mentioned it is the default domain anyway, i overlooked that. 
Thank you for cleaning then 👍

PR #11506, opened 2022-02-22
[DataObject] Do not update search backend when auto save or only saving version

### LINKED ISSUE #11475: [Bug] Searchbackend messages are dispatched when autosaving elements
### Expected behavior

do not dispatch searchbackend messages when autosave since they are useless (because handler loads published version of element anyway).

also related to https://github.com/pimcore/pimcore/issues/11474

### Actual behavior

searchbackend messages are dispatched on every autosave

### Steps to reproduce

open data object and change data -> autosave is triggered -> searchbackend message is dispatched

dvesh3:
Resolves #11475

### DIFF
--- a/bundles/CoreBundle/EventListener/SearchBackendListener.php
+++ b/bundles/CoreBundle/EventListener/SearchBackendListener.php
@@ -49,7 +49,7 @@ class SearchBackendListener implements EventSubscriberInterface
             DocumentEvents::PRE_DELETE => 'onPreDeleteElement',
             AssetEvents::PRE_DELETE => 'onPreDeleteElement',
 
-            DataObjectEvents::POST_UPDATE => 'onPostUpdateElement',
+            DataObjectEvents::POST_UPDATE => 'onPostUpdateDataObject',
             DocumentEvents::POST_UPDATE => 'onPostUpdateElement',
             AssetEvents::POST_UPDATE => 'onPostUpdateElement',
         ];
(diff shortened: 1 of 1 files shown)


brusch:
@dvesh3 should be handled for documents as well, right?

dvesh3:
> @dvesh3 should be handled for documents as well, right?

@brusch you are right, pushed the changes for Documents as well 😉

bundles/CoreBundle/EventListener/SearchBackendListener.php
```diff
@@ -49,8 +49,8 @@ public static function getSubscribedEvents()
             DocumentEvents::PRE_DELETE => 'onPreDeleteElement',
             AssetEvents::PRE_DELETE => 'onPreDeleteElement',
 
-            DataObjectEvents::POST_UPDATE => 'onPostUpdateElement',
-            DocumentEvents::POST_UPDATE => 'onPostUpdateElement',
+            DataObjectEvents::POST_UPDATE => 'onPostUpdateDataObjectDocument',
+            DocumentEvents::POST_UPDATE => 'onPostUpdateDataObjectDocument',
```
jdreesen: A bit strange, that the listener method for `DocumentEvents::POST_UPDATE` is `onPostUpdateDataObjectDocument`.

bundles/CoreBundle/EventListener/SearchBackendListener.php
```diff
@@ -87,4 +87,21 @@ public function onPostUpdateElement(ElementEventInterface $e)
...
+    {
+        //do not update index when auto save or only saving version
+        if (
+            ($e->hasArgument('isAutoSave') && $e->getArgument('isAutoSave')) ||
+            ($e->hasArgument('saveVersionOnly') && $e->getArgument('saveVersionOnly'))
+        ) {
+            return;
+        }
+
+        $this->onPostAddElement($e);
```
jdreesen: Why does `onPostUpdateDataObjectDocument` call `onPostAddElement` and not `onPostUpdateElement`?

jdreesen: Hmm, maybe because `onPostAddElement` and `onPostUpdateElement` are the same...

What about having only one method for `onPostAddElement`, `onPostUpdateElement` and `onPostUpdateDataObjectDocument` as all three are basically the same?

bundles/CoreBundle/EventListener/SearchBackendListener.php
dvesh3: suggestion?

bundles/CoreBundle/EventListener/SearchBackendListener.php
dvesh3: Changed it `onPostAddUpdateElement` for all elements.

PR #11512, opened 2022-02-23
[Console::getExecutable()] Search for executables also outside open_basedir folders

BlackbitDevs:
Currently executables which are not in a directory which is allowed in PHP's `open_basedir` setting do not get found. I already described this in https://github.com/symfony/symfony/issues/41006#issuecomment-971419456 and fixed it for PHP binary in #10843

But for other executables this problem still exists. Imho `open_basedir` defines from which folders PHP is allowed to read files. But we do not want to read a file from there but we want to get the executable path and execute it independent of PHP.

For this reason I added the code from https://github.com/pimcore/pimcore/blob/01ae62571cb2d738cc9b166dad7b0622ab5cbc1d/lib/Tool/Console.php#L112-L118 to `Console::getExecutable()`.

### DIFF
--- a/lib/Tool/Console.php
+++ b/lib/Tool/Console.php
@@ -125,6 +125,16 @@ final class Console
 
                 $executableFinder = new ExecutableFinder();
                 $fullQualifiedPath = $executableFinder->find($executablePath);
+
+                if(!$fullQualifiedPath) {
+                    $checkCmd = 'which '.escapeshellarg($executablePath);
+                    if (self::getSystemEnvironment() === 'windows') {
+                        $checkCmd = 'where '.escapeshellarg($path).':'.$name;
+                    }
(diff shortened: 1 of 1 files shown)


brusch:
Just for reference, the added code was removed in #8230 
Should we maybe wait with that until Symfony has made a final decision here symfony/symfony#41006 

In the meantime you can use the parameters to configure the path to the execs: 
https://github.com/pimcore/skeleton/blob/10.x/config/services.yaml#L9

BlackbitDevs:
Exactly, #8230 removed it believing that Symfony will have cared for this - and most of the time this is true ;-)

I will create a PR for Symfony then because there are too many issues, nobody seems to look at the comment. In the meantime let's keep this PR open until Symfony decided...

brusch:
I'm closing this for now - it seems that the discussion at Symfony also got stuck. As there's a workaround using parameters (see above) I think we can live with that for now 😊

PR #11513, opened 2022-02-23
Reset translation catalogue initialization

kingjia90:
related to PR #11374, more details in the review conversation

### DIFF
--- a/lib/Translation/Translator.php
+++ b/lib/Translation/Translator.php
@@ -160,7 +160,7 @@ class Translator implements TranslatorInterface, TranslatorBagInterface, LocaleA
      */
     public function lazyInitialize($domain, $locale)
     {
-        $cacheKey = 'translation_data_' . md5($domain . '_' . $locale);
+        $cacheKey = $this->getCacheKey($domain,$locale);
 
         if (isset($this->initializedCatalogues[$cacheKey])) {
             return;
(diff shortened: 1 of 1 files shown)


lib/Translation/Translator.php
```diff
@@ -254,6 +254,15 @@ public function lazyInitialize($domain, $locale)
         }
     }
 
+    /**
+    * Resets the initialization of a specific catalogue 
+    */
+    public function resetInitialization($domain, $locale)
+    {
+        $cacheKey = 'translation_data_' . md5($domain . '_' . $locale);
```
dvesh3: I think it is better to refactor cachekey generation to a separate method which then also called in `lazyInitialize()`

dvesh3, line 262: I think it is better to refactor cachekey generation to a separate method which then also called in `lazyInitialize()`

dvesh3: I think it is better to refactor cachekey generation to a separate method which then also called in `lazyInitialize()`

lib/Translation/Translator.php
```diff
@@ -447,4 +459,15 @@ public function __call($method, $args)
...
         return call_user_func_array([$this->translator, $method], $args);
     }
+
+    /**
+     * @param string $domain
+     * @param string $locale
+     *
+     * @return string
+     */
+    private function getCacheKey($domain, $locale)
```
jdreesen: Why not real types instead?

```suggestion
    private function getCacheKey(string $domain, string $locale): string
```

PR #11511, opened 2022-02-23
Update Single.php

DzemalBecirevic:
When generating a single token I got the following errors:
Line 76: "Undefined array index 0".
Line 189: "undefined method "getUsages" on null.

  
  

## Changes in this pull request  
Resolves #

PR #11515, opened 2022-02-23
[Voucher] Add null check and default value for statistic usages

mysliwietzflorian:

After the creation of an OnlineShopVoucherSeries object in the Pimcore admin-backend, the voucher details page throws a "call on null" exception. This PR checks for this exception and adds the default value of 0 (zero) to the usage count.

## Additional info 

Recreate the problem in the Pimcore-demo instance:

- create a new OnlineShopVoucherSeries object
- fill in SingleType token settings, save, and reload. E.g.:
image

- on the tab "Voucher details" you will see a "500 Internal Server Error"

image


jdreesen:
This gets also fixed in #11511

dvesh3:
Closed in favor of #11511 since the other PR fixes more issues. @mysliwietzflorian thanks!

PR #11520, opened 2022-02-23
Fix: Quicksearch exception if brick found

### LINKED ISSUE #11460: Quicksearch exception if brick found
### Expected behavior

I should be able to use quick search without error.

### Actual behavior

Currently if you have bricks with localized fields and use quick search, you get exception: Variable "languages" does not exist in pimcore/bundles/AdminBundle/Resources/views/SearchAdmin/Search/Quicksearch/object.html.twig
quicksearch
It seems pimcore tries to access "languages" which are not defined in this elseif branch
image

### Steps to reproduce

Create brick with localized fields and input some data into brick. Run pimcore:search-backend-reindex and try to search in the quicksearch for your keyword.

blankse:
Fixes #11460


dvesh3 (approved):

dvesh3:
@blankse thanks!

PR #11521, opened 2022-02-23
Fix: The Calculated Field cloning skips Type and Expression

### LINKED ISSUE #11452: [Bug]: The Calculated Field cloning skips Type and Expression.
### Expected behavior

When the 'Calculated value' field is cloned, then:
-  'Calculator type' is copied and is the same as of source field, not empty
- 'Calculator expression' is the same as of source field, not empty

### Actual behavior

When Calculate field is cloned, then:
-  'Calculator type' property is empty but should be the same as of source field
- 'Calculator expression' property is empty but should be the same as of source field

### Steps to reproduce

1. Click on Settings -> Data Objects -> Classes,
2. choose existing or create the new Class
3. choose existing or create the new Calculated Field (Add Data Component -> Others -> Calculated Value)
4. Fill 'Calculator type' property 
5. Fill 'Calculator expression' property
6. Save
7. Click on the existing Calculated field with a right mouse button and select 'Clone'
8. Open the newly cloned Calculated field for editing and see the following: <See Actual behavior>

behavior

blankse:
Fixes #11452

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/calculatedValue.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/calculatedValue.js
@@ -135,7 +135,9 @@ pimcore.object.classes.data.calculatedValue = Class.create(pimcore.object.classe
             }
             Ext.apply(this.datax,
                 {
+                    calculatorType: source.datax.calculatorType,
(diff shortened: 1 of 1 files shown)


dvesh3 (approved):

dvesh3:
@blankse thanks!

PR #11523, opened 2022-02-24
[Asset folder] Go to list view if there are no previewable assets inside

BlackbitDevs:
Currently when you open an asset folder which does not contain any previewable items, it looks like this:
Bildschirmfoto 2022-02-24 um 08 27 11
The user might get the impression that the folder is empty.

With this PR if there are no previewable assets inside the folder, the list view will automatically get opened and the preview panel gets removed:
Bildschirmfoto 2022-02-24 um 08 26 19

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/folder.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/folder.js
@@ -62,6 +62,11 @@ pimcore.asset.folder = Class.create(pimcore.asset.asset, {
             fields: ['url', "filename", "filenameDisplay", "type", "id", "idPath"],
             listeners: {
                 "load": function () {
+                    if(this.store.getCount() === 0) {
+                        this.tabbar.setActiveItem(this.listfolder.getLayout());
+                        this.tabbar.remove(this.dataview);
+                    }
+
                     try {
                         this.dataview.reload();
                     }


bundles/AdminBundle/Resources/public/js/pimcore/asset/folder.js
```diff
@@ -62,6 +62,11 @@ pimcore.asset.folder = Class.create(pimcore.asset.asset, {
             fields: ['url', "filename", "filenameDisplay", "type", "id", "idPath"],
             listeners: {
                 "load": function () {
+                    if(this.store.getCount() === 0) {
+                        this.tabbar.setActiveItem(this.listfolder.getLayout());
+                        this.tabbar.remove(this.dataview);
```
kingjia90, line 67: Great, it works and is ready to merge.

it will also work with `this.tabbar.remove(this.dataview);` alone, since it will just "fallback" to `listfolder` (because it is the next one), if `dataview` is not present.
So `this.tabbar.setActiveItem(this.listfolder.getLayout());` is not strictly necessary, wondering if it would be worth to be removed (because it's obsolete at the moment) or just kept it (more clear and future-proof). wdyt?

BlackbitDevs: It might be that some plugin inserted another tab between `Preview` and `List View` via `postOpenAsset` event. For this reason I would keep the `setActiveItem()`.

kingjia90: yes, sounds right to me, i am merging it. Thank you again for the PR and for clarifyng

PR #11524, opened 2022-02-24
Added nullcheck before method_exists

ESalfinger:
Breaks saving of objects that extend AbstractProduct but have no value for the index field which is interpreted as IdList.

This happens if an index is used for multiple object types which do not share some of the indexed fields.


dvesh3:
@ESalfinger Thanks for your PR! please sign the CLA by clicking the button on CLAassistant comment, so that we can proceed with the review and merge.

brusch:
LGTM, thanks!

PR #11536, opened 2022-02-25
Do not reset condition variables

### LINKED ISSUE #11535: [Bug]: calling getConditionVariables resets condition variables
### Expected behavior

On any `AbstractListing`, calling `setConditionVariables` will have the DAO use the set condition variables. So the implementation should be the same as for `getConditionVariableTypes`:

https://github.com/pimcore/pimcore/blob/b26d41f4f7dfd7a6b606b30495bab56ffc305d7c/lib/Model/Listing/AbstractListing.php#L96-L103

### Actual behavior

When calling `setConditionVariables`, the set condition variables are not used in the SQL query, due to the DAO calling `getConditionVariables`. This call will overwrite all condition variables based on the logic implemented in `getCondition`:

https://github.com/pimcore/pimcore/blob/b26d41f4f7dfd7a6b606b30495bab56ffc305d7c/lib/Model/Listing/AbstractListing.php#L472-L477

### Steps to reproduce

1. Create a class
2. Use the classes listing to call `setConditionVariables(['test-value'])`
3. Call `$list->load()` (use xdebug to check the condition variables for this test case)

passioneight:
Fixes #11535

### DIFF
--- a/lib/Model/Listing/AbstractListing.php
+++ b/lib/Model/Listing/AbstractListing.php
@@ -471,7 +471,7 @@ abstract class AbstractListing extends AbstractModel implements \Iterator, \Coun
      */
     public function getConditionVariables()
     {
-        if(!$this-conditionVariables) {
+        if(!$this->conditionVariables) {
             $this->getCondition();
         }
         
--- a/lib/Model/Listing/AbstractListing.php
+++ b/lib/Model/Listing/AbstractListing.php
@@ -471,8 +471,10 @@ abstract class AbstractListing extends AbstractModel implements \Iterator, \Coun
      */
     public function getConditionVariables()
     {
-        $this->getCondition();          // this will merge conditionVariablesFromSetCondition and additional params into conditionVariables
-
+        if(!$this-conditionVariables) {
+            $this->getCondition();
+        }
+        
         return $this->conditionVariables;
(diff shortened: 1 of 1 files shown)


dvesh3 (commented):
@passioneight I'm afraid this will change will not work for array condition variables. As on calling `setConditionVariables(['test-value'])` sets the `$this->conditionVariables` and later calling `getConditionVariableTypes()` in loadList() returns null and lead to wrong parameter binding. 

please see https://github.com/pimcore/pimcore/pull/4265 for more details.

dvesh3:
A better solution would be to consider existing `conditionVariables` for merge in `getCondition()` https://github.com/pimcore/pimcore/blob/3680ddf728620a5dd219f9cfdcc798b459507f7c/lib/Model/Listing/AbstractListing.php#L330. wdyt?

passioneight:
Not too sure about your suggestion as it would mean that every other part of the `getCondition` method is executed, which seems like an overhead - unless there's some use case I'm not seeing?

dvesh3 (approved):

dvesh3:
@passioneight thanks! yes, it solves the problem of resetting the `conditionVariables`. Just the steps in #11535 were confusing.

PR #11537, opened 2022-02-25
Fix nice command

Ricardocanul7:
When addLowProcessPriority method is called, it add the nice command and '-n 19' but when this is executed by Symfony/Process the execution fails throwing the error `nice: invalid number ' 19'` which seems to be failing because of the blank space before the number

## Additional info  
I found this problem when generating thumbnails from PDF and calling `saveImage()` in `lib/Document/Adapter/Ghostscript.php`


brusch:
Thanks a lot! 👍

Cruiser13:
@Ricardocanul7 thanks, we ran into this bug on pdf generation and directly had a fix - awesome!

PR #11545, opened 2022-02-28
Add missing dao method annotations

blankse:
These methods are used in customer-data-framework
See https://github.com/pimcore/customer-data-framework/pull/248


dvesh3:
@blankse thanks!

PR #11552, opened 2022-03-01
[Reports] Support drag & drop when configuring columns of a custom report

BlackbitDevs:
Currently when configuring a custom report you can only change the order of the columns with the up and down button. Then some time you want to add a new column - this will be added at the very end. Then you have to click th up button to move it one position above. Then move mouse cursor to the move up arrow again and so on. This is not optimal.

With this PR drag and drop gets supported. You can simply take the row you want to move and then drag it to the desired position.
I have kept the up and down arrows for now to not irritate people. Please comment if they do not make sense anymore, then I will remove them.


brusch:
@BlackbitNeueMedien Should we maybe somehow indicate/show that dnd is possible?

BlackbitDevs:
Normally this is done with the drop target icon - but here it is not that you can drop an element there but only reorder. Do you have an idea how to visualize this?

brusch:
Hmm, good question ... I was thinking about a short note between the headline and the grid ... but then I remembered that we have this in other places as well. So I think it's fine for now, but we should create a follow-up ticket, where we should think about a drag and drop visualization for all the grid das support row-ordering via dnd. 
Maybe an icon next to the up- and down arrows could help?

BlackbitDevs:
I am wondering if we could replace https://github.com/pimcore/pimcore/blob/54021096da2b0f74df2d65b68b727791d6adf923/bundles/AdminBundle/Resources/public/css/ext-modifications.css#L309-L311
with `cursor: move` for grid panels which support drag & drop reordering.

brusch:
@BlackbitNeueMedien This would apply to all grid cells then, even if they don't support dnd. 
But, maybe we could use `cursor: grab` and  `cursor: grabbing`  (which seems to be more common for reordering stuff) on the action columns with the arrows?

BlackbitDevs:
@brusch Have changed this PR so that the `cursor: grab` gets used for grid panels which have CSS class `dragdrop`. 
WDYT?
But this approach would mean that we have to remember adding the `dragdrop` CSS class to all grid panels which support `gridviewdragdrop` plugin.

brusch:
@BlackbitNeueMedien I'd prefer to have `grab` only on the action columns for moving the rows. Because we have also grids where you can edit stuff, there the `grab` cursor could be confusing. This should be not the case when doing that only on the action columns.

BlackbitDevs:
@brusch Hmm, I just tried but `griddragdrop` does not work on the action columns (have tested in FF) because the browser behaviour of dragging and dropping the icon image gets used and so the dnd for the grid dragdrop does not even start. O I reverted this so we can enable dnd here for now without visual hint and in another PR care for visual hinting for all grids which support dnd. Is this ok for you?

PS: Imho the failing Stan test is not caused by this PR. Have merged `10.x` branch again but to no success.

brusch:
👍 Thanks - created a follow up ticket for the UI stuff #11667

PR #11555, opened 2022-03-01
[Application logger] Correct log file object path if PIMCORE_PROJECT_ROOT is /app

BlackbitDevs:
Imagine you have an `ApplicationLogger` log with `logFileObject=var/application-logger/3/620e5330c29b24.63887070`

When PIMCORE_PROJECT_ROOT is `/app` then the current replacement in https://github.com/pimcore/pimcore/blob/131ec1750f20b35142afeb4d1a0addeb1edfd38c/lib/Log/FileObject.php#L70 will return the path `varlication-logger/3/620e5330c29b24.63887070`

### DIFF
--- a/lib/Log/FileObject.php
+++ b/lib/Log/FileObject.php
@@ -67,7 +67,7 @@ final class FileObject
      */
     public function getFilename()
     {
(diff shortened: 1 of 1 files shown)


brusch:
Thanks a lot! 👍

PR #11557, opened 2022-03-01
[Bug]: Added Multi-site 'id' to DataObject Route Key

### LINKED ISSUE #11502: [Bug]: PimcoreUrl cannot resolve a data object's site-specific URL slug
### Steps to reproduce

We have a shop with categories. These categories are data objects and have a URL slug property. We also have two different sites set up. Now add a fallback URL slug to one category. Now add a slug for one specific site to the same category. (If you refresh the data object in the backend gui now the slugs will disappear but this a different matter and will receive a separate bug report or pull request today depending on what time allows us to do.)

Call the fallback slug URL: works. Creating links based on this (e.g. pagination) works as well.

Call the site specific slug URL: works as long as there are no links created in the rendered template.

### Actual behavior

Having links generated in the rendered template (e.g. pagination of articles inside a category in a shop) leads to a RouteNotFoundException:
`An exception has been thrown during the rendering of a template ("None of the chained routers were able to generate route: Route 'data_object_12345_/exampleSlug' not found").`

### Expected behavior

The route can be resolved and links/URLs can be generated.

Assumption: It works for the fallback slug but not for the site-specific slug because this line ultimately resolves the route "data_object_12345_/exampleSlug": https://github.com/pimcore/pimcore/blob/10.x/lib/Routing/Dynamic/DataObjectRouteHandler.php#L61
However this method has no knowledge of sites and therefore site specific slugs cannot be resolved.

We're happy to provide a pull request for this matter but we definitely need some pointers and directions on how to fix this to be able to do so in a timely manner.

kingjia90:
fixes #11502

### DIFF
--- a/lib/Routing/DataObjectRoute.php
+++ b/lib/Routing/DataObjectRoute.php
@@ -89,7 +89,7 @@ final class DataObjectRoute extends Route implements RouteObjectInterface
     public function getRouteKey(): ?string
     {
         if ($this->object) {
-            return sprintf('data_object_%d_%s', $this->object->getId(), $this->getPath());
+            return sprintf('data_object_%d_%s_%s', $this->object->getId(), $this->getDefault('site'), $this->getPath());
         }
 
         return null;
--- a/lib/Routing/DataObjectRoute.php
+++ b/lib/Routing/DataObjectRoute.php
@@ -89,7 +89,7 @@ final class DataObjectRoute extends Route implements RouteObjectInterface
     }
 
     /**
-     * @param Site $slug
+     * @param Site $site
      *
      * @return $this
      */
--- a/lib/Routing/DataObjectRoute.php
+++ b/lib/Routing/DataObjectRoute.php
@@ -17,6 +17,7 @@ namespace Pimcore\Routing;
 
 use Pimcore\Model\DataObject\Concrete;
 use Pimcore\Model\DataObject\Data\UrlSlug;
+use Pimcore\Model\Site;
 use Symfony\Cmf\Component\Routing\RouteObjectInterface;
 use Symfony\Component\Routing\Route;
 
--- a/lib/Routing/DataObjectRoute.php
+++ b/lib/Routing/DataObjectRoute.php
@@ -90,11 +90,11 @@ final class DataObjectRoute extends Route implements RouteObjectInterface
     }
 
     /**
-     * @param Site $site
+     * @param Site|null $site
      *
      * @return $this
      */
-    public function setSite(Site $site): self
+    public function setSite(?Site $site): self
     {
         $this->site = $site;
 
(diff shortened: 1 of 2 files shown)


kingjia90:
the suggested approach doesn't completely fix but improves the situation.

BEFORE:
no-multi-site: query with obsolete `AND (siteId = 0 OR siteId = 0)`
multi-site 2: doesn't find the slug when siteId is missing or not resolveable (by siteResolver either because it's a "decoding" function like `getRouteByName(string)`, or in preview generator because of FrontendRequestByAdmin isn't able to get the current site)
multi-site 2+: same problem

AFTER:
no-multi-site: query on entire table is fine, the column siteId is 0 anyways.
multi-site 2: when siteId is not set, opt for the slug of the a site (!=0 non-root), if not "fallbacks" to 0. thanks to  `ORDER BY siteId DESC LIMIT 1`. If is set, just do `(siteId = X OR siteId = 0)` as originally expected
multi-site 2+:  when siteId is not set and there's a site with same slug, there's a risk that the query is not able to pick the right record and returning the right object id, otherwise still works

still looking for a better approach

lib/Routing/Dynamic/DataObjectRouteHandler.php
```diff
@@ -98,13 +97,15 @@ public function matchRequest(RouteCollection $collection, DynamicRequestContext
...
     {
+        $site = $this->siteResolver->getSite();
         $route = new DataObjectRoute($slug->getSlug());
         $route->setOption('utf8', true);
         $route->setObject($object);
         $route->setSlug($slug);
         $route->setDefault('_controller', $slug->getAction());
         $route->setDefault('object', $object);
         $route->setDefault('urlSlug', $slug);
+        $route->setDefault('site', $site ? $site->getId() : 0);
```
brusch: I think we can just set the site object here, instead of the id (similar to the `object` above), or was there a particular reason for using the id?

lib/Routing/DataObjectRoute.php
```diff
@@ -89,7 +89,7 @@ public function getContent(): ?object
     public function getRouteKey(): ?string
     {
         if ($this->object) {
-            return sprintf('data_object_%d_%s', $this->object->getId(), $this->getPath());
+            return sprintf('data_object_%d_%s_%s', $this->object->getId(), $this->getDefault('site'), $this->getPath());
```
brusch: I think it would make sense to also add a `protected ?Site $site` property to this class, rather than using the default functionality from the parent. (see also `slug` and `object` property)

lib/Routing/Dynamic/DataObjectRouteHandler.php
```diff
@@ -57,8 +57,8 @@ public function __construct(
      */
     public function getRouteByName(string $name)
     {
-        if (preg_match('/^data_object_(\d+)_(.*)$/', $name, $match)) {
-            $slug = DataObject\Data\UrlSlug::resolveSlug($match[2]);
+        if (preg_match('/^data_object_(\d+)_(\s+)_(.*)$/', $name, $match)) {
```
brusch:
```suggestion
        if (preg_match('/^data_object_(\d+)_(\d+)_(.*)$/', $name, $match)) {
```
Since we use the id of the site `\d+` should be sufficient.

brusch (changes requested):

PR #11566, opened 2022-03-02
Clears created tmp files on image optimization exceptions

### LINKED ISSUE #11522: [Improvement]: Delete temp files from image optimisation on error
### Improvement description

We found a problem with the image optimisation algorithm.
I one project we had 330G images in the folder "var/tmp" witch are from the image optimisation.
The problem is that there was an huge tiff file. For that file there was a thumbnail with ~10M.
The optimizer "pngout" run very long for that image an run into a timeout (witch is 60s default from the symfony process component).
We also found many temp files in many other projects, with much smaller thumbnails.

I think there could be 2 things to avoid such a huge amount of temp files.

First in the `Pimcore\Image\Optimizer\AbstractCommandOptimizer` the [symfony process](https://github.com/pimcore/pimcore/blob/b6d4a89e88d5ae761a4311e8594e004c6fcf38a5/lib/Image/Optimizer/AbstractCommandOptimizer.php#L35) should be initialized with an configurable timeout value which defaults to 60s.

The second thing is that there is a random file generated for the optimisation process. 
When an exception occurs like symfony process timeout the whole process stops an the ramdon file is still there.
I think there should be a try catch arround of the [complete optimisation](https://github.com/pimcore/pimcore/blob/b6d4a89e88d5ae761a4311e8594e004c6fcf38a5/lib/Image/Optimizer.php#L41-L54) for example in `Pimcore\Image\Optimizer::optimizeImage`.
When an Error or an Exception occurs, the temp file should be deleted.

GALCF:
Fixes #11522

### DIFF
--- a/lib/Image/Optimizer.php
+++ b/lib/Image/Optimizer.php
@@ -35,13 +35,16 @@ class Optimizer implements ImageOptimizerInterface
         $extension = File::getFileExtension($path);
         $storage = Storage::get('thumbnail');
         $optimizedImages = [];
+        $failedImages = [];
         $workingPath = File::getLocalTempFilePath($extension);
         file_put_contents($workingPath, $storage->read($path));
 
         foreach ($this->optimizers as $optimizer) {
             if ($optimizer->supports($storage->mimeType($path))) {
+                $tmpFilePath = File::getLocalTempFilePath($extension);
+
                 try {
-                    $optimizedFile = $optimizer->optimizeImage($workingPath, File::getLocalTempFilePath($extension));
+                    $optimizedFile = $optimizer->optimizeImage($workingPath, $tmpFilePath);
 
                     $optimizedImages[] = [
                         'filesize' => filesize($optimizedFile),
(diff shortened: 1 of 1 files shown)


lib/Image/Optimizer.php
```diff
@@ -72,6 +76,12 @@ public function optimizeImage($path)
             unlink($tmpFile['path']);
         }
 
+        foreach ($failedImages as $failedImage) {
+            if (is_file($failedImage)) {
+                unlink($failedImage);
+            }
+        }
+
         if (is_file($workingPath)) {
```
blankse: Is this not always false? The workingPath seems to be a dir not a file

GALCF: I think `workingPath` is just a misleading variable name, it seems to be a temporary file to work on with the contents from Storage (could be cloud-based): `file_put_contents($workingPath, $storage->read($path));`

brusch:
I'd delete the tmp file straigt away if optimization fails ... just pushed the change.

lib/Image/Optimizer.php
blankse: You are right. Thanks for explaining.

PR #11572, opened 2022-03-02
[Task]: escaping fields in SERP preview

kingjia90:
prevents this
image


dvesh3:
please rebase to fix the tests. Fixed here https://github.com/pimcore/pimcore/pull/11576

PR #11573, opened 2022-03-02
[Task]: escaping 'key' custom property field in elements

kingjia90:


dvesh3:
please rebase to fix the tests. Fixed here https://github.com/pimcore/pimcore/pull/11576

PR #11576, opened 2022-03-03
Fix new PhpStan errors

blankse:
We get new errors with the newest version https://github.com/phpstan/phpstan/releases/tag/1.4.7


brusch:
Thank you very much!

PR #11587, opened 2022-03-03
[Data Object] Fallback for mandatory inherited data object brick

### LINKED ISSUE #11538: [Bug]: ObjectBrick does not properly inherit data from parent object when validating fields
### Expected behavior

Saving a child element containing an Objectbrick with inherited mandatory values takes parent's values into consideration when validating, as hinted correctly in UI.

### Actual behavior

This applies to all objects' children (Variants, Children) where inheritance is enabled and object bricks are present. After creating a new child element, the values of the parent object's brick are correctly shown. Publishing the child element triggers a ValidationException.

Callstack_invalid_brick

During validation in `models/DataObject/Concrete.php/update()` in line 128 `$this->$getter()` returns a new/blank instance of the Objectbrick instead of the parent's values. I suppose this is because of an existing entry in `object_brick_store_<name>_<clsId>` table which is channeled through the DAO and not enriched with the parent's data?

image

In `models/DataObject/ClassDefinition/Data/Objectbricks.php/checkValidity()` the validation fails due to empty field contents on the mandatory field by definition.

See the call stack here:

Timestamp: Fri Feb 25 2022 12:19:57 GMT+0100 (Mitteleuropäische Normalzeit)
Status: 422 | Unprocessable Content
URL: /admin/object/save?task=publish
Method: PUT
Message: Validation failed: invalid brick misc: Empty mandatory field [ specialNumber ][ misc ]
Trace: 
in /var/www/vendor/pimcore/pimcore/models/DataObject/Concrete.php:188
#0 /var/www/vendor/pimcore/pimcore/models/DataObject/AbstractObject.php(734): Pimcore\Model\DataObject\Concrete->update(true, Array)
#1 /var/www/vendor/pimcore/pimcore/models/DataObject/Concrete.php(714): Pimcore\Model\DataObject\AbstractObject->save(Array)
#2 /var/www/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php(1367): Pimcore\Model\DataObject\Concrete->save()
#3 /var/www/vendor/symfony/http-kernel/HttpKernel.php(152): Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectController->saveAction(Object(Symfony\Component\HttpFoundation\Request))
#4 /var/www/vendor/symfony/http-kernel/HttpKernel.php(74): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#5 /var/www/vendor/symfony/http-kernel/Kernel.php(202): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#6 /var/www/public/index.php(36): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#7 {main}

### Steps to reproduce

1. Create a new Object definition
2. In said definition enable Inheritance and Variants
image

3. Add an Object Brick field to the definition
image

4. Create a new Object Brick definition
5. In said Object Brick definition add a mandatory field
image

6. Link the definition with the Object's brick field
image

7. Create a new Object and fill all data, save
8. Create a new Variant of that Object 
9. You can see the inherited value
image

10. When trying to publish you will get the validation error.

kingjia90:
fixes #11538

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Objectbricks.php
+++ b/models/DataObject/ClassDefinition/Data/Objectbricks.php
@@ -620,7 +620,13 @@ class Objectbricks extends Data implements CustomResourcePersistingInterface, Ty
                             try {
                                 $key = $fd->getName();
                                 $getter = 'get' . ucfirst($key);
-                                $fd->checkValidity($item->$getter(), false, $params);
+                                
+                                $value = $item->$getter();
+                                if (empty($value)) {
+                                    $value = $item->getValueFromParent($key);
+                                }
+                                $fd->checkValidity($value, false, $params);
+                                
                             } catch (Model\Element\ValidationException $ve) {
                                 $ve->addContext($this->getName());
                                 $validationExceptions[] = $ve;
--- a/models/DataObject/ClassDefinition/Data/Objectbricks.php
+++ b/models/DataObject/ClassDefinition/Data/Objectbricks.php
@@ -618,21 +618,34 @@ class Objectbricks extends Data implements CustomResourcePersistingInterface, Ty
 
                     if (!$omitMandatoryCheck) {
                         foreach ($collectionDef->getFieldDefinitions() as $fd) {
+                            $key = $fd->getName();
+                            $getter = 'get' . ucfirst($key);
+
                             try {
-                                $key = $fd->getName();
-                                $getter = 'get' . ucfirst($key);
-
-                                $value = $item->$getter();
-                                if ($fd->getMandatory() && empty($value)) {
-                                    try {
-                                        $value = $item->getValueFromParent($key);
-                                    }catch (InheritanceParentNotFoundException $e){}
-                                }
-                                $fd->checkValidity($value, false, $params);
-
+                                $fd->checkValidity($item->$getter(), false, $params);
                             } catch (Model\Element\ValidationException $ve) {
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/Objectbricks.php
```diff
@@ -620,7 +621,15 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
                             try {
                                 $key = $fd->getName();
                                 $getter = 'get' . ucfirst($key);
-                                $fd->checkValidity($item->$getter(), false, $params);
+
```
fashxp, line 626: hmm, I thing we should do it the same way as we do this check for class attributes, see [here](https://github.com/pimcore/pimcore/blob/b42dbbf37d10f6460662f5dc518b24d4028054b0/models/DataObject/Concrete.php#L133)

please adapt accordingly, because currently it is not checked, if inheritance is activated for the class or not, and if the field definition supports inheritance at all...

fashxp (changes requested):

kingjia90:
closing this PR, i had troubles to switch to 10.3
moved to #11639

PR #11591, opened 2022-03-04
Add DuplicateFullPathException

### LINKED ISSUE #11589: [Improvement]: throw DuplicateFullPathException instead of Exception
### Improvement description

Ever so often, one needs to handle a duplicate full path situation gracefully. For example, when a user registers with the same email twice (with the email being used as the object's key).

Pimcore only triggers an `Exception`, which is just too generic:

https://github.com/pimcore/pimcore/blob/aa0ed48fd97757fc0b30bd2985eef2a237b373c6/models/DataObject/AbstractObject.php#L857

It would be nice to have an explicit exception, such as `DuplicateFullPathException` to catch. Otherwise, one has to check beforehand if the user already exists or check if the error message contains `Duplicate full path`.

Where the former is prone to error because some values might not be set at the time of checking the path - especially when using event subscribers. The latter, though, is also prone to error, because the exception might be caused by the application trying to create another already existing object (e.g., an `Address`).

Ultimately, the final application code should look something like this:

```php
try {
    $user->save(['versionNote' => 'User registered']);
} catch (DuplicateFullPathException $e) {
    $emailService->sendAlreadyRegisteredEmail($user);
    return $this->redirectToRoute('app_foo_bar');
}
```

dpfahlbusch:
Resolves #11589


passioneight:
I wonder if we should also pass the duplicate object to the exception, so one can retrieve it in the `catch` clause, e.g., `$e->getDuplicateElement();`. Wdyt?

dpfahlbusch:
I implemented it :)

@brusch could you have a look?

models/Exception/DuplicateFullPathException.php
```diff
@@ -0,0 +1,35 @@
...
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\Exception;
```
brusch:
```suggestion
namespace Pimcore\Model\Element;
```
Since we have paths only for `ElementInterface` types, it probably makes sense, to use a more specific namespace.

brusch (changes requested):

models/Element/DuplicateFullPathException.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        $this->duplicateElement = $duplicateElement;
+
+        parent::__construct($message, $code, $previous);
+    }
+
+    public function getDuplicateElement(): ?AbstractElement
+    {
+        return $this->duplicateElement;
+    }
```
blankse: I would prefer ElementInterface and a setter method.

```suggestion
    private ?ElementInterface $duplicateElement = null;

    public function setDuplicateElement(?ElementInterface $duplicateElement)
    {
        $this->duplicateElement = $duplicateElement;
    }

    public function getDuplicateElement(): ?ElementInterface
    {
        return $this->duplicateElement;
    }
```

dpfahlbusch: Good idea :) Changed it

models/Asset.php
```diff
@@ -709,7 +710,14 @@ public function correctPath()
...
             if ($duplicate instanceof Asset && $duplicate->getId() != $this->getId()) {
-                throw new \Exception('Duplicate full path [ ' . $this->getRealFullPath() . ' ] - cannot save asset');
+                $duplicateFullPathException = new DuplicateFullPathException(
+                    'Duplicate full path [ ' . $this->getRealFullPath() . ' ] - cannot save asset',
+                    0,
+                    null
+                );
+                $duplicateFullPathException->setDuplicateElement($duplicate);
+
+                throw $duplicateFullPathException;
```
blankse: Parameter 0 and null can be removed. This should be enough (Also for object/document):
```suggestion
                throw new DuplicateFullPathException('Duplicate full path [ ' . $this->getRealFullPath() . ' ] - cannot save asset')
                    ->setDuplicateElement($duplicate);
```

dpfahlbusch:
Dont know why codeception is failing :(

brusch:
@dpfahlbusch there's a problem with the latest `bitnami/minio:latest` container 😉

models/Asset.php
```diff
@@ -709,7 +710,10 @@ public function correctPath()
         if (Asset\Service::pathExists($this->getRealFullPath())) {
             $duplicate = Asset::getByPath($this->getRealFullPath());
             if ($duplicate instanceof Asset && $duplicate->getId() != $this->getId()) {
-                throw new \Exception('Duplicate full path [ ' . $this->getRealFullPath() . ' ] - cannot save asset');
+                $duplicateFullPathException = new DuplicateFullPathException('Duplicate full path [ ' . $this->getRealFullPath() . ' ] - cannot save asset',);
```
aryaantony92: We should remove the trailing comma. 
```suggestion
                $duplicateFullPathException = new DuplicateFullPathException('Duplicate full path [ ' . $this->getRealFullPath() . ' ] - cannot save asset');
```

models/DataObject/AbstractObject.php
```diff
@@ -854,7 +855,10 @@ protected function correctPath()
         if (Service::pathExists($this->getRealFullPath())) {
             $duplicate = DataObject::getByPath($this->getRealFullPath());
             if ($duplicate instanceof self && $duplicate->getId() != $this->getId()) {
-                throw new \Exception('Duplicate full path [ '.$this->getRealFullPath().' ] - cannot save object');
+                $duplicateFullPathException = new DuplicateFullPathException('Duplicate full path [ '.$this->getRealFullPath().' ] - cannot save object',);
```
aryaantony92: We should remove the trailing comma.

```suggestion
                $duplicateFullPathException = new DuplicateFullPathException('Duplicate full path [ '.$this->getRealFullPath().' ] - cannot save object');
```

models/Document.php
```diff
@@ -555,7 +556,10 @@ private function correctPath()
         if (Document\Service::pathExists($this->getRealFullPath())) {
             $duplicate = Document::getByPath($this->getRealFullPath());
             if ($duplicate instanceof Document && $duplicate->getId() != $this->getId()) {
-                throw new \Exception('Duplicate full path [ ' . $this->getRealFullPath() . ' ] - cannot save document');
+                $duplicateFullPathException = new DuplicateFullPathException('Duplicate full path [ ' . $this->getRealFullPath() . ' ] - cannot save document',);
```
aryaantony92: We should remove the trailing comma.

```suggestion
                $duplicateFullPathException = new DuplicateFullPathException('Duplicate full path [ ' . $this->getRealFullPath() . ' ] - cannot save document');
```

models/Asset.php
dpfahlbusch: Done 👍🏻

models/DataObject/AbstractObject.php
dpfahlbusch: Done 👍🏻

models/Document.php
dpfahlbusch: Done 👍🏻

aryaantony92:
@blankse @dpfahlbusch @passioneight Thank you

Issue #6808, opened 2020-07-20
[Core] Build all entities using DI


brusch:

brusch:
A possible way could be to register all the needed non-shared services dynamically by a compiler pass. 
We also need to consider class mappings here. 
Then the ModelFactory could just obtain the initialized objects from the container. 

Any other ideas?

dpfaffenbauer:
what entities are meant here?

brusch:
Mainly all `ElementInterface`, so e.g. `Pimcore\Model\Document\Page`

brusch:
Unfortunately this seems to be impossible ... see details https://github.com/pimcore/pimcore/pull/11674#issuecomment-1149559394

dpfaffenbauer:
Honestly, this is considered an anti-pattern anyway, why should models that get's instanced multiple times in one request get injected services?

It is better to have services getting data from the models and using it. The models (ElementInterface) should only hold data. Persisting the Data should be done using a Service (Similar to EntityManager and UnitOfWork in Doctrine ORM). That is a lot of work though... But it can be approached model by model. Starting with the simpler ones. E.g. Thumbnail Generation in Assets should be done using a Service not a function in the model.

brusch:
Agree as well. 
I think a possible compromise could be using the DAOs as services and trying to move more and more stuff out of the `save()` methods. This would ensure compatibility while still being able to better separate persistence.

dkarlovi:
@brusch it seems like you're looking to migrate away from active record to data mapper. :+1: 

An approach we took there (on a non-Pimcore project) is to move save logic from the model into a service, then for BC inject the service into each model and have the model `save()` call `service->save($this)`, roughly. This gave us enough legroom to be able to convert in chunks because the previous code mostly worked, new code started to get written exclusively in services.

Target is for the models to at first only be DTOs, later they can be made better by injecting services into them to directly use, not to hold.

PR #11595, opened 2022-03-04
Fix: Blocking Events recursion

### LINKED ISSUE #11526: [Bug]: Calling certain methods in EventListener can cause an infinite recursion
### Expected behavior

No infinite recursion

### Actual behavior

Calling certain methods, like `$asset->getThumbnail($name)`, in postUpdate Eventlistener can cause an infinite recursion. 
If the thumbnail is not already existing, the method will call `$asset->save()` and therefore the EventListener will be fired again. Thus getThumbnail() gets called again, asset will be saved again and so on.

Possible solutions:

1. Implement a method, that prevents the events from being fired. Similiar to Version::disable()
2. Provide a method, that only saves the custom settings/metadata, without really saving the asset (without events, without versioning, etc...)
3. Rethink the whole approach, since most of the functionality can be replaced by checking for the existence of the file in the filesystem (file_exists method)

### Steps to reproduce

1.) Register an Eventlistener for postUpdate on assets
2.) call ` $asset->getThumbnail($name);`  and make sure that the thumbnail doesn´t exist already

kingjia90:
fixes #11526


jdreesen (commented):
What about creating a new function for dispatching events, like:

```php
protected function dispatchEvent(Event $event, string $name): void {
	if ($this->isAllowedEvent($name)){
		\Pimcore::getEventDispatcher()->dispatch($event, $name);
	}
}
```

Then you wouldn't have to wrap every single dispatch into an `if` check.

---

Maybe everything (including `$disabledEvents`, `disableEvents()` and `isAllowedEvent()`) can even be put into a trait and be reused.

kingjia90:
@jdreesen Thank you for the feedback! I will surely follow your suggestion, the code would look much cleaner

brusch:
Probably it would also make sense to put that code into a trait that can be used anywhere necessary 🤔

lib/Event/Traits/DisableEventTrait.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    protected static $disabledEvents = [];
+
+    /**
+     * Sets the list of disabled events
+     *
+     * @param array $events
+     * @return void
+     */
+    public static function disableEvents(array $events = []): void
```
brusch: This is not used anywhere? I think we can kick it anyway.

lib/Event/Traits/DisableEventTrait.php
```diff
@@ -0,0 +1,63 @@
...
+
+use Symfony\Contracts\EventDispatcher\Event;
+
+trait DisableEventTrait
+{
+
+    /**
+     * @var array
+     */
+    protected static $disabledEvents = [];
```
brusch:
```suggestion
    private $activeDispatchingEvents = [];
```
I guess we need to consider this here as well then? 
https://github.com/pimcore/pimcore/blob/9859aacdc0653090182c1d403bcc4b60226bf4e8/models/Element/AbstractElement.php#L391

lib/Event/Traits/DisableEventTrait.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Event\Traits;
+
+use Symfony\Contracts\EventDispatcher\Event;
+
+trait DisableEventTrait
```
brusch:
```suggestion
trait RecursionBlockingEventDispatchHelperTrait
```
Maybe a more meaningful class name?

lib/Event/Traits/DisableEventTrait.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Checks if an event is not currently disabled
+     *
+     * @param string $eventName
+     * @return bool
+     */
+    private function isAllowedEvent(string $eventName): bool
+    {
+        return !in_array($eventName, self::$disabledEvents);
+    }
```
brusch:
```suggestion
```
We just need that once, so we don't need an extra method for that imho. 😉

lib/Event/Traits/DisableEventTrait.php
```diff
@@ -0,0 +1,63 @@
...
+     * @param Event $event
+     * @param string $name
+     * @return void
+     */
+    protected function dispatchEvent(Event $event, string $name): void
+    {
+        if ($this->isAllowedEvent($name)) {
+            \Pimcore::getEventDispatcher()->dispatch($event, $name);
+        }
+    }
```
brusch:
```suggestion
    protected function dispatchEvent(object $event, string $eventName = null): ?object
    {
        $eventName = $eventName ?? \get_class($event);
        if (!isset($this->activeDispatchingEvents[$eventName])) {
            $this->activeDispatchingEvents[$eventName] = true;
            \Pimcore::getEventDispatcher()->dispatch($event, $eventName);
            unset($this->activeDispatchingEvents[$eventName]);
        }
    }
```
I'd use the same method signature as `\Symfony\Contracts\EventDispatcher\EventDispatcherInterface::dispatch

brusch (changes requested):

lib/Event/Traits/DisableEventTrait.php
kingjia90: i've tested it, nice approach, works simply fine for the purpose. 

The main differences are that:
- `activeDispatchingEvents` approach is automatical (like by default in the system), while `disableEvents` is more manual, up to the developer and situational
- the `activeDispatchingEvents` stops the same event that was called by the parent event, but doesn't have control over stopping/preventing any other events from firing. In other words, it won't help to stop PRE_UPDATE in a recursive POST_UPDATE. With `disableEvents`, it would allow to manually decide to stop any other ones before that
-  `disableEvents` could be used in other circumstances beside from preventing recursion.

Not sure if these could be some valid points to consider, but the recommendation was to follow the `Version::disable()` approach, that's why it looks like that (like `disableEvents` not used anywhere in the PR).

brusch: As discussed, we only want to block recursion for the same object at the moment (as this is a bug). 
Can you please update the PR accordingly? Thanks a lot 😊

lib/Event/Traits/RecursionBlockingEventDispatchHelperTrait.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Dispatches an event, avoids recursion by checking if the active dispatch event is the same
+     *
+     * @param object $event
+     * @param string|null $eventName
+     * @return void
+     */
+    protected function dispatchEvent(object $event, string $eventName = null): void
+    {
+        $eventName = $eventName ?? \get_class($event);
```
jdreesen:
```suggestion
        $eventName ??= \get_class($event);
```

brusch:
@kingjia90 thanks for updating the PR, however, one thing u missed: 😉
> I guess we need to consider this here as well then? 
https://github.com/pimcore/pimcore/blob/9859aacdc0653090182c1d403bcc4b60226bf4e8/models/Element/AbstractElement.php#L391

lib/Event/Traits/RecursionBlockingEventDispatchHelperTrait.php
```diff
@@ -0,0 +1,42 @@
...
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Event\Traits;
+
+trait RecursionBlockingEventDispatchHelperTrait
```
brusch, line 21:
```suggestion
/**
 * @internal
 */
trait RecursionBlockingEventDispatchHelperTrait
```
One last thing, I guess we should keep this trait `@internal` 😉

brusch:
I'm just thinking ... we should probably do this for other entities as well to be consistent, right, so this would be: 
```
DataObjectClassDefinitionEvents
DataObjectClassificationStoreEvents
DataObjectCustomLayoutEvents
DataObjectQuantityValueEvents
NotificationEvents
RedirectEvents
SearchBackendEvents
TagEvents
TargetGroupEvents
TranslationEvents
UserRoleEvents
VersionEvents
```

@kingjia90 Can you please update the PR accordingly?

lib/Event/Traits/RecursionBlockingEventDispatchHelperTrait.php
```diff
@@ -0,0 +1,45 @@
...
+     * @param object $event
+     * @param string|null $eventName
+     * @return void
+     */
+    protected function dispatchEvent(object $event, string $eventName = null): void
+    {
+        $eventName ??= \get_class($event);
+        if (!isset($this->activeDispatchingEvents[$eventName])) {
+            $this->activeDispatchingEvents[$eventName] = true;
+            \Pimcore::getEventDispatcher()->dispatch($event, $eventName);
```
dpfaffenbauer, line 41: that is not good. You are going from a Container injected Event Dispatcher to a statically called injector with `\Pimcore::getContainer`. That is a step back..

dpfaffenbauer: (In the context of controllers)

brusch: We not gonna use that in the context of controllers 😉But thanks for the hint. 
As soon as #6808 is done, we're going to change `dispatchEvent()` to `@required` and remove the static getter - so it's already peparing for the good 😉

brusch: Ah, I've just seen that it was used in a controller ... that's not needed anyway as far as I can see, so this can reverted 😉

bundles/AdminBundle/Controller/Admin/TargetingController.php
```diff
@@ -240,7 +249,7 @@ public function targetGroupAddAction(Request $request, CoreCacheHandler $cache,
         $targetGroup->save();
 
         $event = new TargetGroupEvent($targetGroup);
-        $eventDispatcher->dispatch($event, TargetGroupEvents::POST_ADD);
+        $this->dispatchEvent($event, TargetGroupEvents::POST_ADD);
```
brusch: This is a bit out of context, but we should also consider this in this PR. 
The `TargetGroupEvents` should be dispatched in the `save()` method of the model and not in the controller -> similar to other models. 
@kingjia90 can you change that accordingly? Thanks

brusch (changes requested):

models/Tool/Targeting/TargetGroup.php
```diff
@@ -192,4 +192,24 @@ public function getActive()
...
+     * @return void
+     */
+    public function save(): void
+    {
+        if ($this->getId()) {
+            $this->dispatchEvent(new TargetGroupEvent($this), TargetGroupEvents::POST_UPDATE);
+        }else{
+            $this->dispatchEvent(new TargetGroupEvent($this), TargetGroupEvents::POST_ADD);
+        }
+    }
```
jdreesen: The actual action is missing here. You should call `delete()` and `save()` on the `dao`, I think.

lib/Event/Traits/RecursionBlockingEventDispatchHelperTrait.php
```diff
@@ -0,0 +1,45 @@
...
+     * Dispatches an event, avoids recursion by checking if the active dispatch event is the same
+     *
+     * @param object $event
+     * @param string|null $eventName
+     * @return void
+     */
+    protected function dispatchEvent(object $event, string $eventName = null): void
+    {
+        $eventName ??= \get_class($event);
+        if (!isset($this->activeDispatchingEvents[$eventName])) {
```
dpfaffenbauer, line 39: What if you have depending events? So if you have a listener that saves another DataObject that should then trigger the same event just for the other object? That might not work with that solution right?

brusch: That should work, since the dispatching is just blocked for the originating object as it isn't a static variable.

models/Tool/Targeting/TargetGroup/Dao.php
```diff
@@ -83,6 +85,7 @@ public function save()
     public function delete()
     {
         $this->db->delete('targeting_target_groups', ['id' => $this->model->getId()]);
+        $this->model->dispatchEvent(new TargetGroupEvent($this->model), TargetGroupEvents::POST_DELETE);
```
jdreesen: Hmm, not sure if the DAO should dispatch the events... I think it should be the responsibility of the model (like everywhere else).

brusch: @jdreesen I agree - @kingjia90 so you can revert your last change and just add `$this->getDao()->save/delete()` to the methods instead.

lib/Model/AbstractModel.php
```diff
@@ -29,6 +30,7 @@
 abstract class AbstractModel implements ModelInterface
 {
     use ObjectVarTrait;
+    use RecursionBlockingEventDispatchHelperTrait;
```
brusch: Since `RecursionBlockingEventDispatchHelperTrait` isn't used for all implementations that use `AbstractModel` I wouldn't include it here, but rather use it individually on the classes that make use of `dispatchEvent()`.

brusch (changes requested):

lib/Model/AbstractModel.php
kingjia90: What about the __sleep? can i keep it in `AbstractModel`?

```
public function __sleep()
    {
        $blockedVars = ['dao', 'o_dirtyFields', 'activeDispatchingEvents'];
```

or should i also add
```
   public function __sleep()
    {
        $blockedVars = ['activeDispatchingEvents'];

        return array_diff($parentVars, $blockedVars);
    }
```
in each class?

brusch: the `__sleep` we can leave as it is, as it doesn't hurt 😉

models/Asset.php
```diff
@@ -52,6 +53,7 @@ class Asset extends Element\AbstractElement
 {
     use ScheduledTasksTrait;
     use TemporaryFileHelperTrait;
+    use RecursionBlockingEventDispatchHelperTrait;
```
```suggestion
```
This is not needed, as we have the trait already on `AbstractElement`.

models/Asset.php
```diff
@@ -22,6 +22,7 @@
 use Pimcore\Event\AssetEvents;
 use Pimcore\Event\FrontendEvents;
 use Pimcore\Event\Model\AssetEvent;
+use Pimcore\Event\Traits\RecursionBlockingEventDispatchHelperTrait;
```
```suggestion
```

models/Element/AbstractElement.php
```diff
@@ -388,7 +390,7 @@ public function getVersions()
     public function __sleep()
     {
         $parentVars = parent::__sleep();
-        $blockedVars = ['dependencies'];
+        $blockedVars = ['dependencies','activeDispatchingEvents'];
```
brusch:
```suggestion
        $blockedVars = ['dependencies'];
```
This is redundant, as we have it already on `AbstractModel`.

brusch (changes requested):
Just some more cosmetics, the rest looks good to me ☺️

PR #11596, opened 2022-03-04
[Assets] Add normalize/denormalize method for metadata

### LINKED ISSUE #8489: [Assets] Add normalize/denormalize method for metadata
analog to #8487

dvesh3:
Resolves #8489

### DIFF
--- a/models/Asset/MetaData/ClassDefinition/Data/Asset.php
+++ b/models/Asset/MetaData/ClassDefinition/Data/Asset.php
@@ -43,7 +43,11 @@ class Asset extends Data
             $element = Service::getElementById('asset', $value);
         }
 
-        return $element ? $element->getRealFullPath() : '';
+        if (isset($params['fullpath'])) {
+            return $element ? $element->getRealFullPath() : '';
+        }
+
+        return $element;
     }
 
     /**
--- a/models/Asset/MetaData/ClassDefinition/Data/Asset.php
+++ b/models/Asset/MetaData/ClassDefinition/Data/Asset.php
@@ -47,6 +47,35 @@ class Asset extends Data
     }
 
     /**
+     * @param mixed $value
+     * @param array $params
+     *
+     * @deprecated use denormalize() instead, will be removed in Pimcore 11
+     *
+     * @return string
+     */
+    public function unmarshal($value, $params = [])
+    {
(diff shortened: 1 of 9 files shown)


models/Asset/MetaData/ClassDefinition/Data/Asset.php
```diff
@@ -20,41 +20,57 @@
...
+    }
+
+
     /**
      * @param mixed $value
      * @param array $params
      *
      * @return null|int
+     *
+     * @deprecated use normalize() instead, will be removed in Pimcore 11
```
blankse: We should add trigger_deprecation() in method body

dvesh3: Good catch! done 😉

models/Asset/Service.php
```diff
@@ -22,6 +22,7 @@
 use Pimcore\Model\Asset;
 use Pimcore\Model\Asset\MetaData\ClassDefinition\Data\Data;
 use Pimcore\Model\Element;
+use Pimcore\Normalizer\NormalizerInterface;
```
fashxp: why do we need that import here?

fashxp (changes requested):
the data retrieved via getter and after a normalize-denormalize cascade should be the same. 

is not the case for metadata of type `asset`, `document` and `object`

we should also add a test for that  ... snippet for a headstart of that test could be as follows: 

```php 
        $loader = \Pimcore::getContainer()->get('pimcore.implementation_loader.asset.metadata.data');
        
        $asset = Asset::getById(68);
        $metadataList = $asset->getMetadata();

        foreach($metadataList as $metaData) {
            try {
                /** @var Data $instance */
                $instance = $loader->build($metaData['type']);

                $marshalledData = $instance->normalize($metaData['data']);
                $unmarshalledData = $instance->denormalize($marshalledData);

                p_r($metaData['name'] . ': ');
                p_r($unmarshalledData === $metaData['data']);
                p_r("-------------------");

            } catch (UnsupportedException $e) {
            }


        }
```

models/Asset/Service.php
dvesh3: yes, we can get rid of it,

fashxp (approved):

PR #11597, opened 2022-03-04
small correction of siteId in the resolveSlug query

kingjia90:
It fixes the query to not filtering by `(siteId = 0 OR siteId = 0)` when siteId is 0

### DIFF
--- a/models/DataObject/Data/UrlSlug.php
+++ b/models/DataObject/Data/UrlSlug.php
@@ -341,11 +341,16 @@ class UrlSlug implements OwnerAwareFieldInterface
         $db = Db::get();
 
         try {
+            $filterSiteId = 'siteId = 0';
+            if ($siteId){
+                $filterSiteId = sprintf('(siteId = %d OR siteId = 0)', $siteId)
+            }
+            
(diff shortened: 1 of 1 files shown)


brusch:
Hmm, I can't reproduce the issue, for me `(siteId = 0 OR siteId = 0)` queries work 🤔.

### Example: 
```php
        $query = sprintf(
            'SELECT * FROM %s WHERE slug = %s AND (siteId = %d OR siteId = 0) ORDER BY siteId DESC LIMIT 1',
            'object_url_slugs',
            $db->quote('/this-is-just-a-test'),
            0
        );
```

kingjia90:
i am sorry, my PR description was unclear, i meant that i wanted to avoid from making the query with the WHERE conditions as `(siteId = 0 OR siteId = 0)` which is not meaningful to have when siteId is 0 (but it's fine when is different than 0).
It's not a bug, no perfomance killer, it still works, just it shouldn't have this "obsolete" OR in the default case (no multi-sites) or for siteId=0.

brusch:
Thanks for clarifying 😊

PR #11601, opened 2022-03-07
[Application logger] Correct log file object path if PIMCORE_PROJECT_ROOT is /app

BlackbitDevs:
Follow-up of #11555 where we forgot this place.


lib/Log/ApplicationLogger.php
```diff
@@ -150,9 +150,9 @@ public function log($level, $message, array $context = [])
 
         if (isset($context['fileObject'])) {
             if (is_string($context['fileObject'])) {
-                $context['fileObject'] = str_replace(PIMCORE_PROJECT_ROOT, '', $context['fileObject']);
+                $context['fileObject'] = preg_replace('/^'.preg_quote(\PIMCORE_PROJECT_ROOT, '/').'/', '', $context['fileObject']);
             } else {
```
brusch: Should we maybe add an `instanceof` here?

brusch (changes requested):

lib/Log/ApplicationLogger.php
BlackbitDevs: In this `is_string` case? Or do you mean https://github.com/pimcore/pimcore/pull/11601/files#diff-935282905cd71faeda62809e162086e37f92682e8f041d8e6e40eafe55f2c11cR155

brusch: Yeah, I mean changing the `else` to an `elseif($context['fileObject'] instanceof FileObject)`.

BlackbitDevs: Fixed

brusch:
Nice, thanks a lot! 👍

PR #11605, opened 2022-03-07
Fix Link Editable display in editmode

youwe-petervanderwal:
- Fix HTML encoding of entered text (a `htmlspecialchars` was applied in the `frontend` method, but wasn't applied in edit mode)
- Apply and display textPrefix and textSuffix configuration in editmode


brusch:
Thanks a lot! 😊

PR #11612, opened 2022-03-07
[Documents] Fix deleting a brick does not delete its content

### LINKED ISSUE #8122: Deleting a brick does not delete its content in tmp_store
# Bug Report

### Expected behavior
Deleting a brick should also delete its content. (or is this an intended feature as a "backup"?)

### Actual behavior
When you delete a brick and and add the same type at the same position the former content will be restored.

The update method of each brick writes contents to the tmp_store table. Deleting and adding a brick will trigger a reload from the tmp_store table. Conclusion is therefore, that the delete does not remove the content from the tmp_store and then gets the content from this table when you readd the same brick at the same index.

### Steps to reproduce

Add a page. Add a brick (e.g. a Text-Brick). Enter some text. Delete the brick and readd the same brick type at the same position.
=> Text from removed brick will reappear.

dvesh3:
Resolves #8122

## Additional info  
Steps to reproduce:
1. Add new Document with areablock editable.
2. Add a brick & data
3. Save and Reload Document 
4. Remove existing brick and re-add the same brick type on same index.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/PageController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PageController.php
@@ -443,7 +443,8 @@ class PageController extends DocumentControllerBase
         $blockStateStackData = json_decode($request->get('blockStateStack'), true);
         $blockStateStack->loadArray($blockStateStackData);
 
-        $document = Document\PageSnippet::getById($request->get('documentId'));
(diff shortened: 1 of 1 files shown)


brusch:
@dvesh3 just pushed a simpler fix that should do the trick as well, could you please double check the approach? Thanks ☺️

PR #11614, opened 2022-03-07
Fix: DocumentResolver::getDocument() don't return document in new areabricks

blankse:
If a areabrick service has a dependency to DocumentResolver, the getDocument() method returns null if you create a new areabrick. If you reload the edit iframe, the method returns the document correct.


blankse:
@brusch No candidate for 10.3.3? :(

brusch:
LGTM, thank you very much! 👍

PR #11616, opened 2022-03-08
[Admin] Documents - Fix not possible to open documents in localhost -…

### LINKED ISSUE #11590: [Admin] Documents - not possible to open documents in localhost
### Expected behavior

A document should be opened 

### Actual behavior

When hostname is not set in system settings then exception is thrown on opening a Document:
image

### Steps to reproduce

1. Setup Pimcore with Docker
2. Log into Admin interface
3. Open a Document

Regression of https://github.com/pimcore/pimcore/pull/11418

dvesh3:
Resolves #11590

## Additional info  
Reverted the change `$request->getHost() === 'localhost'` introduced in #11418 as it changes the behavior internally for calls to `Tool::getHostname()`. Before the change the `localhost` host was returned from the method and after it started returning `null` if there's no hostname set in the System Settings or configs, which in turn broke some functions.

### DIFF
--- a/lib/Tool.php
+++ b/lib/Tool.php
@@ -395,7 +395,7 @@ final class Tool
     {
         $request = self::resolveRequest($request);
 
-        if (null === $request || !$request->getHost() || $request->getHost() === 'localhost') {
+        if (null === $request || !$request->getHost()) {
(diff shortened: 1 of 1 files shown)


jdreesen:
Target branch should be `10.3`, no?

PR #11620, opened 2022-03-08
Added exception where pregmatch would return unexpected $matches

BentBr:
Hey Folks!

Resolves the issue of receiving a 500 in backend if a user in frontend requests a non-existent image url.

For example the image
https://www.baufritz.com/01_H%C3%A4user/Kundenh%C3%A4user/2004/Schauer/Bilder/image-thumb__3005__houseStage/22112005170126.jpg
does exist. Changing the name to a non-existing one returns a 404, which is great.

But requesting that URI without an image returns a 200 and internally creates a 500 due to the not existing key $matches[1]
https://www.baufritz.com/01_H%C3%A4user/Kundenh%C3%A4user/2004/Schauer/Bilder/image-thumb__3005__houseStage/

Now, with the runtimeException thrown the default behaviour of Pimcore is being set in motion (throws a 404 as it should).

p.s. keep going - I love pimcore :)

### DIFF
--- a/bundles/CoreBundle/Controller/PublicServicesController.php
+++ b/bundles/CoreBundle/Controller/PublicServicesController.php
@@ -116,6 +116,10 @@ class PublicServicesController extends Controller
 
                     preg_match("@([^\@]+)(\@[0-9.]+x)?\.([a-zA-Z]{2,5})@", $filename, $matches);
 
+                    if (empty($matches) || ! isset($matches[1])) {
+                        throw new \RuntimeException('Requested asset does not exist');
+                    }
+
                     if (array_key_exists(2, $matches) && $matches[2]) {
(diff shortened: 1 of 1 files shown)


BentBr:
Thx @jdreesen and @brusch for your hints.... that's the result of working with too many different accounts :facepalm:

brusch:
@BentBr Thanks a lot for your PR, however, it seems that this doesn't change anything in behavior.
When throwing a `RuntimeException` we still end up here: 
https://github.com/pimcore/pimcore/blob/eb1140df9ba7f6bbc9cc8a34e0e4e6529cfa20fa/bundles/CoreBundle/Controller/PublicServicesController.php#L172
So the error image get's displayed, which is wrong behavior. 
Anyway, I've created an alternative PR for that topic: #11669
Please let me know if that works for you as well. 
Thanks in advance!

PR #11621, opened 2022-03-08
[Tracking] Fix warning undefined array index in GoogleTagManager

jremmurd:
Fixes Warning undefined array key "impressions".

`$this->deferred["impressions"]` is not always set

image


brusch:
Thank you very much! 👍

PR #11624, opened 2022-03-08
[Admin] DataObject Tree - Fix tree child nodes listing permission

### LINKED ISSUE #11540: [Bug]: Nested children permission not properly considered, when a file/path with longer name is present
### Expected behavior

if a subfolder has a children that you are allowed to list, it should have `+` icon and be clickable to open it. 
If you are allowed to list a nested children, it should show the `+` icon on any folder/element in between root and the nested children folder.

### Actual behavior

if there's a file with a longer named file or path (cpath), the folder fails to display the `+` icon and it doesn't open the folder on click neither (so it's not just a visual bug). 
It takes a refresh (right-click on the folder) to get the '+' icon and turn it clickable and list the content.

### Steps to reproduce

Something like this: 
cpath problem

This issue is actually the "extracted juice" from some finding with @roland4432 and started with this PR #11333, which may contain further details.

https://github.com/pimcore/pimcore/blob/81ba8b753b8d8609eba67df1a18f57c0f4542add/models/DataObject/AbstractObject/Dao.php#L319
It seems that `LOCATE(cpath,CONCAT(o.o_path,o.o_key))=1 ORDER BY LENGTH(cpath)`, while it's a good approach to (i guess) avoid nested loops, sub-queries or recursion to get the most deeply nested children, is not working in the case of a file with a longer name
image
_abcdefghilmnoprst_ gets the priority here.

The problem is therefore happening in any other places that rely on ` ORDER BY LENGTH(cpath)`

### LINKED ISSUE #11603: [Bug]: Pimcore assets/objects folders tree view not displaying for private admin users from side bar. If we are mapping with sub child folders.
### Expected behavior

If we gave the access for sub child folder in workspace. While logging with that private user. By default, access folder tree view should be display on the sidebar. 

### Actual behavior

Screenshot (71)
Screenshot (70)

Hi,

Currently, In Pimcore10 admin panel. I have created one private user and give the asset access for the one only one sub child folder. While logging with that private user. By default, assets tree view is not displaying. This scenario is creating. When we give the access for sub child folders not for parent folder. Please observe in attached screenshots. Can any body please check on this issue. 

### Steps to reproduce

.

dvesh3:
Resolves #11540
Resolves #11603

Also consider https://github.com/pimcore/pimcore/issues/6771 for testing purpose.
## Additional info  
Fix `hasChildren` & children listing by considering `list=1` condition

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -170,9 +170,9 @@ class DataObjectController extends ElementControllerBase implements KernelContro
                 $userIds[] = $this->getAdminUser()->getId();
                 $condition .= ' AND
                 (
-                    (SELECT list FROM users_workspaces_object WHERE userId IN (' . implode(',', $userIds) . ') AND LOCATE(CONCAT(objects.o_path,objects.o_key),cpath)=1 ORDER BY LENGTH(cpath) DESC, FIELD(userId, '. $this->getAdminUser()->getId() .') DESC, list DESC LIMIT 1)=1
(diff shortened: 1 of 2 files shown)


manasaR507 (approved):

kingjia90:
closing in favor of the approach in #11714.
The changes here were always showing the `+` icon, even when you disallow all the children

PR #11625, opened 2022-03-08
Align `webdav_locks` table creation in `install.sql`

jdreesen:
Aligns the `webdav_locks` table creation with the rest of the statements in the `install.sql`.

It was originally added in #9005.

This also helps [scripts like this](https://github.com/pimcore/demo/blob/10.x/dump/create-dump.php#L12) to get the table names.


brusch:
Totally makes sense - thank you very much!

PR #11632, opened 2022-03-09
Added srcset attribute to img element

John5:
Addresses issue #11316 by adding a srset attribute to img elements for thumbnail images.


brusch:
Thank you very much!

PR #11636, opened 2022-03-09
Remove javascript alert().

timwiechers:
If an error occurs when changing a workflow place, a JavaScript alert is currently displayed.

My suggestion would be to delete the error handling here altogether, because I can't think of a sensible way to handle it either. A server error is displayed to the user anyway.


brusch:
Thank you!

PR #11639, opened 2022-03-09
[Data Object] Fallback for mandatory inherited data object brick

kingjia90:
fixes https://github.com/pimcore/pimcore/issues/11538


fashxp (approved):
lgtm

PR #11640, opened 2022-03-10
[Ecommerce] Fix warning: Undefined array key "key" in ElasticSearch ProductList

jremmurd:
Fix warning 
> Warning: Undefined array key "key" File: /var/www/html/vendor/pimcore/pimcore/bundles/EcommerceFrameworkBundle/IndexService/ProductList/ElasticSearch/AbstractElasticSearch.php Line: 1255

For range aggregation the `key` may not be set, but `from` and `to` for example.


brusch:
🍒-picked into `10.3` as it is a bug-fix, see: 98a68aff52295e37d2fb1d080682ead1a0366982

Issue #11507, opened 2022-02-23
[Bug]: Create document translation -> inherit controller


valentin-:
### Expected behavior

When creating a document translation with inheritance the controller is set to defaultAction.

I've created a listener to inherit the controller, but imho this should be the default behaviour as the document type for main page and translation should match.

### Actual behavior

Document EN:
![image](https://user-images.githubusercontent.com/90033/155289335-73614a2b-0d77-4d0d-8762-558303d1c27b.png)

Creating a translation for DE will use the default controller:
![image](https://user-images.githubusercontent.com/90033/155289425-d7867bfc-ff70-4021-8b9b-cdca30a75c10.png)

### Steps to reproduce

Create a document with a custom controller.
Create a translation for this document.

Pimcore 10.3

dvesh3:
Fixed by #11641

PR #11641, opened 2022-03-10
[Fix] create language inherited child didn't set the controller & template

youwe-petervanderwal:
Resolves bug #11507

## Additional info  
Related snippet from DocumentController::addAction():

```
$docType = Document\DocType::getById($request->get('docTypeId'));
if ($docType) {
   // ...
} elseif ($request->get('translationsBaseDocument')) {
    // This section was never reached because $docType was always a DocType object
    $translationsBaseDocument = Document::getById($request->get('translationsBaseDocument'));
    if ($translationsBaseDocument instanceof Document\PageSnippet) {
        $createValues['template'] = $translationsBaseDocument->getTemplate();
        $createValues['controller'] = $translationsBaseDocument->getController();
    }
}
```

## Note
Seems that this bug is introduced in f008e2b973f1ffbad20f982f70230287df1a18d9 - not 100% sure this is the right fix to this but I assume that change was also done for a reason

### DIFF
--- a/models/Document/DocType.php
+++ b/models/Document/DocType.php
@@ -117,6 +117,10 @@ class DocType extends Model\AbstractModel
      */
     public static function getById($id)
     {
+        if (empty($id)) {
+            return null;
+        }
+
         try {
             $docType = new self();
             $docType->getDao()->getById($id);
--- a/models/Document/DocType/Dao.php
+++ b/models/Document/DocType/Dao.php
@@ -45,27 +45,22 @@ class Dao extends Model\Dao\PimcoreLocationAwareConfigDao
      *
      * @throws \Exception
      */
-    public function getById($id = null)
+    public function getById(?string $id = null): void
     {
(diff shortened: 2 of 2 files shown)


dvesh3:
@youwe-petervanderwal This PR looks good to me. Could you please also revert this change https://github.com/pimcore/pimcore/pull/10478/files#diff-f7053f28596f2f8ceaf52906b900bceb1c8093e4cc701dafc9b17156a811cadcR51-R55? as Dao::getById() isn't suppose to return value. thanks!

youwe-petervanderwal:
@dvesh3 
- Reverted commit
- Rewritten the logic of that method so that your original goal is still achieved (otherwise it somehow complained on passing null to getDataByName() which wasn't allowed)
- Added argument and void return type to that method

dvesh3 (approved):

dvesh3:
@youwe-petervanderwal thank you!

PR #11647, opened 2022-03-11
Set _locale when handling slug route

lukadschaak:
## Changes in this pull request and Additional info
When processing a DataObject with a slugged route, AND the route contains a locale part, then the locale is never processed.

This PR add's some funtionality, like the \Pimcore\Model\Staticroute::match() does. If a static route has a `_locale` parameter and a proper pattern given, it will be resolved. This PR looks anly at the first part of the path of the url and uses it as `_locale`

Maybe the static route mechanism would make sense for slugs too, but this is out of my scope.

### DIFF
--- a/lib/Routing/Dynamic/DataObjectRouteHandler.php
+++ b/lib/Routing/Dynamic/DataObjectRouteHandler.php
@@ -125,9 +125,10 @@ final class DataObjectRouteHandler implements DynamicRouteHandlerInterface
         preg_match("/\/(\w+)\/.*/", $slugUrl, $matches);
 
         if (is_array($matches) && count($matches) === 2) {
-            // TODO check, whether locale is valid. Own SluggableInterface implementations may have
-            // other patterns
-            $route->setDefault('_locale', $matches[1]);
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
This doesn't feel right either... Pimcore knows from the slug field (inside the localized-field) what locale a certain slug should serve the content in. So Localization isn't properly thought about with slugs I guess? @brusch?

lukadschaak:
Ah, what it see know is: `Slug` has a db field named `position`, which can be read by `$slug->getPosition()` and it contains the language key like `de` or `en`. Is this the language from the localized field? Than this would be a one liner.

brusch:
Well yes, this is probably something we can think about but this proposed solution in this PR is not generic enought, because it assumes to have the locale in the first part of the path info. 
I think the best solution would be to set `_locale` accordingly when the matching slug is inside a localized field - that would be the intentional behavior.

brusch:
@lukadschaak oh, I was a bit too late - but same idea ☺️
Created already a follow-up PR: #11649

lukadschaak:
Nice! Thanks a lot :+1:

PR #11649, opened 2022-03-11
[Data Object] Set _locale when slug of localized field matches

brusch:
See also #11647


brusch:
@lukadschaak plz let me know if that works for you ☺️

lukadschaak:
@brusch That works for me. Thanks :+1:

brusch:
Thanks for the feedback 👍

PR #11670, opened 2022-03-15
Fix showing correct sort options for non admin users

### LINKED ISSUE #11618: [Bug]: User unable to change sortMode to index or key
### Expected behavior

When the current sort mode of an object is `index` the user should be able to change it to `key`, and vice versa.

### Actual behavior

When the current mode is `index` the user is only able to change it to `index` (which makes no sense as it is the current mode already). When you are an admin all the 3 possible sort modes are available.

I have looked in the code and the bug seems to be in https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/object/tree.js#L725.

`currentSortMethod == "index"` should be `currentSortMethod != "index"` imo, same applies for the condition `currentSortMethod == "key"`.

### Steps to reproduce

- Login is an non admin user
- Right click an object. And check sort mode, by_index is missing

bramstroker:
Resolves #11618

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tree.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tree.js
@@ -709,7 +709,7 @@ pimcore.object.tree = Class.create({
 
                     let currentSortMethod = record.data.sortBy;
 
-                    if (currentSortMethod == "key" || user.admin) {
+                    if (currentSortMethod !== "key" || user.admin) {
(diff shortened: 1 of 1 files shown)


brusch:
🍒-picked into `10.3` as it is a bug fix: d38e67ad1ee222f6021dd0f155631445a84c4595

bramstroker:
Thanks for the quick merge

PR #11672, opened 2022-03-15
[Assets] Use correct file extension when downloading svg image as jpg

BlackbitDevs:
Steps to reproduce bug:

1. Upload svg image to Pimcore assets
2. Open asset
3. On right side under "Custom download" select Format: JPG
4. Click Download

Currently you will get an svg file provided for download but it is actually a jpg with wrong file extension.


BlackbitDevs:
Seems to have only been a problem in Pimcore 6. But I wonder if it got even worse under Pimcore X because when I upload an SVG to https://demo.pimcore.fun/admin/ or https://demo.pimcore.com/admin/ I only see errors - no preview and of course no chance to test if downloading SVG as JPG now works...

PR #11674, opened 2022-03-15
Build ElementInterface entities using DI

### LINKED ISSUE #6808: [Core] Build all entities using DI

brusch:
Partly resolves #6808

- [x] Performance benchmark against current implementation. Report below
- [ ] replace all `new` usages in the controllers, etc. 

To be done in extra PRs: 
- [ ] Change all other models to use the entity builder (like, Redirect, ...) 
- [ ] Consider class-definition loading, it's basically about building services in `__set_state()` e.g. here https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/models/DataObject/ClassDefinition/Helper/VarExport.php#L55 but there are also specific implementations

#### Execution time on building 10000 elements through:

ClassMapLoader (Old):
| Assets | Documents | DataObjects |
|--------|-----------|-------------|
| 11ms   | 7ms 	     | 16ms 	   |

ContainerBuilder (New):
| Assets | Documents | DataObjects |
|--------|-----------|-------------|
| 40ms  | 35ms 	 |  32ms     	   |


lib/Event/Traits/RecursionBlockingEventDispatchHelperTrait.php
```diff
@@ -38,8 +42,26 @@ protected function dispatchEvent(object $event, string $eventName = null): void
...
             unset($this->activeDispatchingEvents[$eventName]);
         }
     }
+
+    /**
+     * @return EventDispatcherInterface|null
+     */
+    private function getEventDispatcher(): ?EventDispatcherInterface
+    {
+        return $this->eventDispatcher ?? \Pimcore::getEventDispatcher();
```
jdreesen:
```suggestion
        return $this->eventDispatcher ??= \Pimcore::getEventDispatcher();
```

bundles/CoreBundle/DependencyInjection/Compiler/EntityModelFactoryPass.php
```diff
@@ -0,0 +1,94 @@
...
+            Document\Hardlink::class,
+            Document\Link::class,
+            Document\Newsletter::class,
+            Document\Page::class,
+            Document\Printcontainer::class,
+            Document\Printpage::class,
+            Document\Snippet::class,
+
+            DataObject\Folder::class,
+        ];
```
jdreesen: I think this should be a `const`.

brusch: I'm planning to go for a generic approach anyway ... so basically scanning the classes in `model/` and initializing them.

brusch:
Unfortunately, we've hit a dead end with this approach ... 
The problem is the cache, or more specific retrieving items from the cache that are using injected services. 

### Example
Consider a model override, that uses DI 
```php
namespace App\Model;

use Symfony\Component\HttpFoundation\RequestStack;

class Car extends \Pimcore\Model\DataObject\Car {
    private RequestStack $requestStack;

    /**
     * @required 
     */
    public function setRequestStack(RequestStack $requestStack) {
        $this->requestStack = $requestStack;
    }
}
```

That works great as long as the object is built using the factory, so when the data is not coming from the cache. 
But the most common case is, that objects are coming from the cache and therefore the data is just unserialized and services cannot be injected.

Therefore, in the example above `$this->requestStack` won't contain the expected value (the current request stack) but it will contain the value from a former request that wrote the item to the cache. 

However, we could make the situation a bit better by just building the `Dao` through the container ... since they are not going to the cache anyway ... 

Any thoughts?

passioneight:
@brusch why would you want DI in models? IMO, models should only hold data.

If you need to apply some logic, one should use a service and pass the model's instance to the desired method. The service can use DI for the requeststack.

If there is no way around and the requeststack _has to be_ in the model, you can always use the controller to inject the requeststack and call setRequeststack on the model.

Would love to know your use case for this. I'm also interested in the community's take on this.

brusch:
@passioneight I agree. 
This feature was requested very often by the community, I've used `RequestStack` just as an example, because it illustrates the problem very well 😉
Imho it's mostly because of getting rid of public services in models (using `Pimcore::getContainer()`) - but I think that we could mostly solve when using `Dao`'s as services ... and that seems to be possible.

PR #11675, opened 2022-03-15
[Bug]: (Grid view) adds "empty" option in select

### LINKED ISSUE #11604: [Bug]: (Empty) value not available in grid edit
### Expected behavior

The expected behavior is that we have the possibility to select the (Empty) option/value in the grid edit view.

Example:

Screenshot 2022-03-07 at 13 14 56

### Actual behavior

All values of a select field are displayed except the (Empty) option/value.

 
image

### Steps to reproduce

Create a select field in an object with a few values and try to select the (Empty) value in the grid edit view.

kingjia90:
fixes #11604


kingjia90:
Need to double-check if other editables also miss the `empty` option, let's see if I can put all in this PR before the release

PR #11676, opened 2022-03-15
Update install.sql

AlexanderWeckmer:
The 'error' field is missing. 
Without the field, errors while sending emails aren't get stored. As a result, it's not possible to see any errors in the Sent Emails Panel of the Newsletter module.

This is how it should look like, and does after adding the field and clearing cache.

image

Please merge

### DIFF
--- a/bundles/InstallBundle/Resources/install.sql
+++ b/bundles/InstallBundle/Resources/install.sql
@@ -273,6 +273,7 @@ CREATE TABLE `email_log` (
   `bcc` longtext,
   `sentDate` int(11) UNSIGNED DEFAULT NULL,
   `subject` varchar(500) DEFAULT NULL,
+  `error` text, DEFAULT NULL, 
   PRIMARY KEY (`id`),
   KEY `sentDate` (`sentDate`, `id`),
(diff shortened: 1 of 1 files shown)


bundles/InstallBundle/Resources/install.sql
```diff
@@ -273,6 +273,7 @@ CREATE TABLE `email_log` (
   `bcc` longtext,
   `sentDate` int(11) UNSIGNED DEFAULT NULL,
   `subject` varchar(500) DEFAULT NULL,
+  `error` text, DEFAULT NULL,
```
dvesh3:
```suggestion
  `error` TEXT NULL,
```

dvesh3:
Regression of https://github.com/pimcore/pimcore/pull/9553

dvesh3:
@AlexanderWeckmer thanks!

blankse:
@dvesh3 Wrong branch? Shouldn't it be 10.3?

brusch:
@blankse thanks for the hint - just 🍒-picked into `10.3`.

PR #11682, opened 2022-03-16
Update 02_Nginx_Configuration.md

fheyer:
In development config for nginx we have to set `include fastcgi.conf;` or else php is not working. 
Therefor i uncommented this line just like in production config.


jdreesen (approved):

brusch:
🍒-picked into `10.3`

PR #11686, opened 2022-03-16
fix deadlock issues in classificationstore, fieldcollections and relations

hirakch:
fix deadlock issues in classificationstore, fieldcollections and relations

  
  

## Changes in this pull request  
Resolves #8228 

## Additional info  

Delete without having the data and insertion on the same table creates gap lock which causes deadlock for parallel processing on the same class

Also, created another PR to fix in 6.9 for legacy codes #11688

PR #11688, opened 2022-03-16
fix deadlock issues in classificationstore, fieldcollections and rela…

### LINKED ISSUE #8228: [DataObject][Classificationstore] regular deadlocks when saving data objects with classification store data in a parallelized way
Deadlocks occur on tables `object_classificationstore_data_*` on a regular base when saving classification store data in a parallelized ways.

Maybe we can find ways to make locking more sophisticated to reduce deadlocks -> would improve performance of importing/updating classification store data.

hirakch:
fix deadlock issues in classificationstore, fieldcollections and relations table

  
  

Resolves #8228 

## Additional info  

Will create a separate PR to resolve this in 10.x as well #11686

### DIFF
--- a/models/DataObject/Classificationstore/Dao.php
+++ b/models/DataObject/Classificationstore/Dao.php
@@ -62,7 +62,11 @@ class Dao extends Model\Dao\AbstractDao
         $dataTable = $this->getDataTableName();
         $fieldname = $this->model->getFieldname();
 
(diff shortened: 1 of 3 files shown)


dvesh3:
Sorry, Pimcore 6.9 CE reached it's EOL so we are not merging any changes on 6.9 branch. you can check supported versions here https://github.com/pimcore/pimcore#supported-versions

PR #11702, opened 2022-03-17
[Asset] Thumbnails: added status cache for image thumbnails

brusch:
This improves the performance significantly when using (remote) object storage for website projects.

### DIFF
--- a/models/Asset/Dao.php
+++ b/models/Asset/Dao.php
@@ -500,7 +500,7 @@ class Dao extends Model\Element\Dao
         return false;
     }
 
-    public function addThumbnail(string $name, string $filename): void
+    public function addToThumbnailCache(string $name, string $filename): void
     {
         $assetId = $this->model->getId();
         $time = time();
@@ -517,7 +517,7 @@ class Dao extends Model\Element\Dao
         }
     }
 
-    public function hasThumbnail(string $name, string $filename): ?int
+    public function getCachedThumbnailModificationDate(string $name, string $filename): ?int
     {
         $assetId = $this->model->getId();
 
@@ -540,7 +540,7 @@ class Dao extends Model\Element\Dao
         return self::$thumbnailStatusCache[$assetId][$hash] ?? null;
     }
 
-    public function deleteThumbnail(?string $name = null): void
+    public function deleteFromThumbnailCache(?string $name = null): void
     {
         $assetId = $this->model->getId();
         $where = [
--- a/models/Asset/Dao.php
+++ b/models/Asset/Dao.php
@@ -15,6 +15,7 @@
 
 namespace Pimcore\Model\Asset;
 
+use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use Pimcore\Db;
 use Pimcore\Loader\ImplementationLoader\Exception\UnsupportedException;
 use Pimcore\Logger;
@@ -31,6 +32,8 @@ class Dao extends Model\Element\Dao
 {
     use Model\Element\Traits\ScheduledTasksDaoTrait;
     use Model\Element\Traits\VersionDaoTrait;
+
+    private static array $thumbnailStatusCache = [];
 
     /**
      * Get the data for the object by id from database and assign it to the object (model)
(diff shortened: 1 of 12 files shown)


models/Asset/Dao.php
```diff
@@ -494,4 +499,59 @@ public function __isBasedOnLatestData()
 
         return false;
     }
+
+    public function addThumbnail(string $name, string $filename): void
```
fashxp:
```suggestion
    public function addToThumbnailCache(string $name, string $filename): void
```

models/Asset/Dao.php
```diff
@@ -494,4 +499,59 @@ public function __isBasedOnLatestData()
...
+                self::$thumbnailStatusCache[$assetId][$hash] = $thumb['modificationDate'];
+            }
+        }
+
+        $hash = $name . $filename;
+
+        return self::$thumbnailStatusCache[$assetId][$hash] ?? null;
+    }
+
+    public function deleteThumbnail(?string $name = null): void
```
fashxp:
```suggestion
    public function deleteFromThumbnailCache(?string $name = null): void
```

models/Asset/Dao.php
```diff
@@ -494,4 +499,59 @@ public function __isBasedOnLatestData()
...
+            'modificationDate' => $time,
+        ]);
+
+        if(isset(self::$thumbnailStatusCache[$assetId])) {
+            $hash = $name . $filename;
+            self::$thumbnailStatusCache[$assetId][$hash] = $time;
+        }
+    }
+
+    public function hasThumbnail(string $name, string $filename): ?int
```
fashxp:
```suggestion
    public function getCachedThumbnailModificationDate(string $name, string $filename): ?int
```

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -178,21 +179,37 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
...
-            } else {
-                // delete the file if it's not valid anymore, otherwise writing the actual data from
-                // the local tmp-file to the real storage a bit further down doesn't work, as it has a
-                // check for race-conditions & locking, so it needs to check for the existence of the thumbnail
-                $storage->delete($storagePath);
+
+        $modificationDate = null;
+        $statusCacheEnabled = \Pimcore::getContainer()->getParameter('pimcore.config')['assets']['image']['thumbnails']['status_cache'];
+        if($statusCacheEnabled) {
+            $modificationDate = $asset->getDao()->hasThumbnail($config->getName(), $filename);
```
fashxp: add additional fallback in that case the thumbnail is not in cache ... but might exist on file system.

fashxp (changes requested):

fashxp:
another scenario: 
if thumbnail is deleted from storage (but not cleared from thumbnail cache), it will never be generated when calling `getPath` or `getPathReference` ...

this for example happens, when calling `clear temporary files` from backend ui

--> see also failing test

models/Asset/Image/Thumbnail/Processor.php
brusch: Doing so would make the cache obsolete ... the whole point of this cache is to avoid these checks by assuming that the thumbnail doesn't exist if it's not in the cache. 

Consider the following example: 
You have a new image, and you request `$thumb->getHtml()` for it. For that you need to generate 7-8 thumbnails for the `<picture>` tag. Without having the fallback there wouldn't be a single request to the filesystem, because the assumtion is that the thumbnail is not generated when not being in cache. So if we do implement a fallback, then you'd have 7-8 requests to the filesystem, because there's nothing in the cache because no thumb was generated.

brusch:
> another scenario: if thumbnail is deleted from storage (but not cleared from thumbnail cache), it will never be generated when calling `getPath` or `getPathReference` ...
> 
> this for example happens, when calling `clear temporary files` from backend ui
> 
> --> see also failing test

If the cache is enabled, this is intended behavior, otherwise (in case we double check the existence of the file) the cache is getting obsolete ... 
But you're right, this needs to be considered in "clear temp files" thogh.

dpfaffenbauer:
YES! Can confirm, before this a Page Load could take up to 10 seconds, with this change, it is < 1 sec.

brusch:
@dpfaffenbauer Glad to hear! Thanks for sharing/confirming 👍

dpfaffenbauer:
@brusch this now still goes a lot to the database getting info about thumbnails. what if we use the Symfony Cache instead? That way it uses either the DB or Redis or Whatever depending on configuration and therefore be faster too.

brusch:
@dpfaffenbauer this was actually my first thought, but this approach has some major drawbacks. The biggest problem I see is, that most of the cache backends are LRU-caches, means that the thumbnail cache could get less reliable with time passing by ... but we could think of a 2-layer solution, with having the cache and as a fallback the database ... but I'm not sure if this would be a real benefit 🤔

PR #11713, opened 2022-03-20
Doc fix about object_(id) View

LionelGalilee:
assertion in table is wrong compare to just above

  
  

Resolves #


dvesh3:
@LionelGalilee thanks!

Issue #6383, opened 2020-05-13
[Core][DataObjects] Severe performance degradation when opening folders with many objects in the tree 


fabfischer:
# Bug Report

Hello,
we have a project with 19 `DataObjects` definitions and a large number of objects (nearly 1 million in total). The majority of the objects are of one type and located in one folder. 
If a user with limited permissions (no admin rights) opens this folder, loading the tree or grid will take a very, very long time (we're talking here about minutes...) The request blocks other requests and possibly the whole system, which makes it impossible to work with the system during this time.

### Expected behaviour
Context: user with limited permissions

A tree-list or grid-view of folders with a large number of objects are loaded in seconds, not minutes (as it is with an admin user).

### Actual behaviour
If you click on a folder with thousands of thousands of items, it takes minutes to load and display the data. (If the request is not already timed out anyway)

### Steps to reproduce
1. create a folder
2. create a large number of objects (600.000 or more) with the folder as the parent
3. create an user with limited permissions. give user permission to "list" the content of the folder 
4. log in with the limited user in another browser
5. open tree or grid for created folder.
6. wait till the data was loaded (which will take time)

### Code example
I was able to narrow down the problem to the following location in the [code](https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L141):

```php
# pimcore/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php:141
if (!$this->getAdminUser()->isAdmin()) {
    $userIds = $this->getAdminUser()->getRoles();
    $userIds[] = $this->getAdminUser()->getId();
    $condition .= ' AND (
                    (SELECT list FROM users_workspaces_object WHERE userId IN (' . implode(',', $userIds) . ') AND LOCATE(CONCAT(objects.o_path,objects.o_key),cpath)=1 ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
                    OR
                    (SELECT list FROM users_workspaces_object WHERE userId IN (' . implode(',', $userIds) . ') AND LOCATE(cpath,CONCAT(objects.o_path,objects.o_key))=1 ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
                )';
}
```

The subqueries within the condition will take most of the time because these are executed for each element in the objects table. 

stats of a real-world EXPLAIN-statement (with obscured ids for the parent and user (roles)))
```
EXPLAIN SELECT
  objects.o_id as o_id,
[...]

PR #11714, opened 2022-03-20
[Admin] Fix element tree child listing permission problems and adding related tests

### LINKED ISSUE #11540: [Bug]: Nested children permission not properly considered, when a file/path with longer name is present

kingjia90:
fixes #11540, #11603
improves #6383

The mismatch of the tree view `+` icon and the actual folder content is caused by a discrepancy between `isAllowed` and `hasChildren` in the DAOs

https://github.com/pimcore/pimcore/blob/18d8b50455f3e4c2c2e1aef38a499fd406e8a865/models/DataObject/AbstractObject/Dao.php#L482-L513
https://github.com/pimcore/pimcore/blob/18d8b50455f3e4c2c2e1aef38a499fd406e8a865/models/DataObject/AbstractObject/Dao.php#L299-L325

The `hasChildren` didn't perfom the equivalent of `$permissionsChildren` (isAllowed) which simply lists all the folder in between the current path and a nested children, to allow the user to reach any nested item which is allowed to.

Since we are scanning the children element of a given folder/parent, we can obtain and store the permissionsParent equivalent (that's the purpose of `isInheritingPermission`) instead of having it in a subquery fired on every item.

In the new `hasChildren` the logic looks different but it's the same, it's just from an upper level and tried to improve perfomance.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -274,7 +274,7 @@ class AssetController extends ElementControllerBase implements KernelControllerE
             // get assets
             $childsList = new Asset\Listing();
             $childsList->addConditionParam('parentId = ?', [$asset->getId()]);
-            $childsList->filterAccessibleByUser($this->getAdminUser());
+            $childsList->filterAccessibleByUser($this->getAdminUser(), $asset);
 
             if (!is_null($filter)) {
                 $childsList->addConditionParam('CAST(assets.filename AS CHAR CHARACTER SET utf8) COLLATE utf8_general_ci LIKE ?', [$filter]);
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -130,7 +130,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
         $filteredTotalCount = 0;
         $limit = 0;
 
-        if ($object->hasChildren($objectTypes, null, $this->getAdminUser())) {
+        if ($object->hasChildren($objectTypes)) {
             $limit = (int)$request->get('limit');
             if (!is_null($filter)) {
                 if (substr($filter, -1) != '*') {
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -217,7 +217,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
             foreach ($childs as $child) {
                 $tmpObject = $this->getTreeNodeConfig($child);
 
-                if ($child->isAllowed('list')) {
+                if ($child->isAllowed('list', $this->getAdminUser())) {
                     $objects[] = $tmpObject;
                 }
             }
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -166,16 +166,23 @@ class DataObjectController extends ElementControllerBase implements KernelContro
             // custom views end
 
             if (!$this->getAdminUser()->isAdmin()) {
-                $userIds = $this->getAdminUser()->getRoles();
-                $userIds[] = $this->getAdminUser()->getId();
 
-                $inheritedPermission = $object->getDao()->isInheritingPermission('list',$userIds);
+                $roleIds = $this->getAdminUser()->getRoles();
+                $currentUserId = $this->getAdminUser()->getId();
+                $permissionIds = array_merge($roleIds ,[$currentUserId]);
+
+                $inheritedPermission = $object->getDao()->isInheritingPermission('list', $permissionIds );
+
+                $existsChildren = 'EXISTS(SELECT list FROM users_workspaces_object uwo WHERE userId IN (' . implode(',', $permissionIds) . ') AND list=1 AND LOCATE(CONCAT(objects.o_path,objects.o_key),cpath)=1 AND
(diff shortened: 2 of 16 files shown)


kingjia90:
Just some outdated notes that were in the main comment

## Preface
The current behaviour is as described here
https://github.com/pimcore/pimcore/issues/1634#issuecomment-831885117

It's only related to users without the Admin permission

## LENGTH(cpath) approach is not ideal

It was supposed to be a quick way to detected the deepest level, but it's affected by long filename.
More details here (#11540).

## HasChildren consider only specified children permissions

The problem of #11603 was related to this `LOCATE(cpath,CONCAT(d.path,d.`key`))=1` which checks if there are any elements that have a longer name, but would miss the "inherited" permission.

It could be solved with #11624, the problem is that this change
https://github.com/pimcore/pimcore/pull/11624/files#diff-51d8f734290cf822c76411e21fe077c9d01d38fc6d49648dea6b6e10ce5269d7R319
`OR LOCATE(CONCAT(o.o_path,o.o_key), cpath)=1)` (to check if `/` is found in `/test/banner/1`)
but along with the `LIST=1` in the query, meant that anything is returned `true` (since every item has a permitted parent in th upper-tree), therefore the `+` icon is always be appearing and ignoring any `LIST=0` rules.

## Scenario
Given a folder tree structure as: 
`root/a/b/c/d/.../z`
if you are allowed to see `root` and blocked only on folder `h`, would it be expected to see any folder between `root` and `g`?

## TLDR
currently the entire folder tree would be not visible because the rule with a long cpath is found and it is list=0.

The "new" `hasChildren` would instead allow everything from root and `g`, blocking `h` folder and its children.

For each folder:
1) It checks if there anything nested with permission `list=1`, if yes, then any folder in between are permitted.
2) If nothing is found, checks if current direct children are blocked or not. 
3) If doesn't find any specific workspace rules, it takes parents' rules (inheritance) 

It works for the above, because the folder from a-g have the inheritance from root (cond. 3). When checking children of `g` it would return nothing because direct children `h` is list 0 (cond. 2).
If we add a new permission eg. `q` with list=1, it won't be checking (cond. 2) for `h` is blocked or not, because it's already returning true on rule (cond. 1) until `q`

For the example in #11603, the `/test/banner/1` will trigger (cond. 1) from root to it.

## More Tecnical Details and Notes

1) `EXISTS(subquery)`) : doesn't care of cpath length, for the purpose of knowing if there's allowed children elements and due the additive behaviour, anything with `LIST=1` is good to go.

   **perfomance-wise:** compared to old approach, it avoids multiple sortings which are quite resource intensive full table scans, uses EXISTS operator for better readability compared to `(select 1 from .. limit=1) = 1`

2) `LEFT JOIN,.. ON list=0` : I mean direct element, because it should NOT affect eg.`b` from displaying `c` because of `a/b/c/d/[..]/h` is blocked. 
In other words, for list=1 the best predictor that current folder should have children and be visibile would be the furthest allowed element and returning true on every folder in between, even granting permission on them, but only for the folder, not its content (otherwise would be unreachable), but for list=0 everything is the opposite, only the closest permission rule counts, we don't want to block the folder in between.

   **perfomance-wise:** In this case, to avoid the subquery (like the commented `NOT EXISTS` in [models/DataObject/AbstractObject/Dao.php](https://github.com/pimcore/pimcore/compare/10.3...tree_view_permission?expand=1#diff-51d8f734290cf822c76411e21fe077c9d01d38fc6d49648dea6b6e10ce5269d7)) and having a matching indexed column between the tables for the purpose, the JOIN should be the best.

3) `IFNULL, $inheritedPermission = 1`: The reason for `isInheritingPermission` is also because the second subquery is unnecessary and expensive to do on each row, the parent permission are almost same for each object with the same parent and is more likely that not everything has specific rules (list=1|0) or entries in the workspace table, so most of the time is doing the subquery

   **perfomance-wise:** It avoids an extra subquery on each found element
https://github.com/pimcore/pimcore/compare/10.3...tree_view_permission?expand=1#diff-23e25a13fb0b0b211542a0d0cb82b9b73d551c51cd72c1f2115246d84ed7e479L91-L95
on 30 elements, it was like 30+30+(1-30) depending on the result of the subquery before the `OR`

## TODO
Apply changes to other methods like `getChildAmount`
Add some rules for UserId in the LEFT JOIN, giving priority to the specific user rules than roles rules. Same to the `EXISTS`.

fashxp:
added some tests ... for my understanding, all the tests should pass --> found some additional problems.

fashxp:
TODOs:
- [x] fix failing tests
- [x] add test for #11540
- [x] do some performance tests

kingjia90:
The problems related to the user level rules having precedence seem fixed with the latest change, i have put some details in the commit description.

The search tests are still failing because the search uses a different [approach](https://github.com/pimcore/pimcore/issues/11822#issuecomment-1090966317), the current PR's changes do not cover or affect the search anyhow, so if it's okay, i would suggest to "split" it up.

PR #11725, opened 2022-03-21
[Versions] Marshalling performance

BlackbitDevs:
This PR accelerates version marshalling by 2 steps:
1. Properties which do not get serialized anyway by https://github.com/pimcore/pimcore/blob/e8362bf58a6a8b7a63e21b24fe4410b9bd540938/models/Version.php#L222, do not need to be marshalled but can be set to `null`. This reduces recursive marshalling effort for objects which do not get saved in the version anyway
2. Add the same filter which is used in https://github.com/pimcore/pimcore/blob/e8362bf58a6a8b7a63e21b24fe4410b9bd540938/models/Element/Service.php#L1215-L1231 to prevent cloning objects which is not necessary for versioning. When #10640 got a final decision we might adjust this the same here.


brusch:
@BlackbitNeueMedien can you please check the failing tests? Thanks 😊

brusch:
@BlackbitNeueMedien 🏓 😊

brusch:

PR #11726, opened 2022-03-22
[Calculated value] Prevent entering an undefined calculated value field type

BlackbitDevs:
Steps to reproduce:
1. Create data object class with calculaterd value field
2. In field `Type` type "invalid"
3. Save class
4. Reload class definition -> Field type "invalid" is really saved

For good luck https://github.com/pimcore/pimcore/blob/e8362bf58a6a8b7a63e21b24fe4410b9bd540938/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/calculatedValue.js#L60-L66 falls back to a text field but nevertheless it does not make sense to allow custom values here.


mcop1:
Thanks 👍

PR #11728, opened 2022-03-22
[Improvement] Add events to detect changes in the published state of data objects.

timwiechers:
This pull request will add events to detect changes in the publication status of data objects.

This would be quite useful if, for example, changes in the publication status are to be synchronized with external systems or notifications are to be sent when changes are made.

List of added events:
- pimcore.dataobject.prePublish
- pimcore.dataobject.postPublish
- pimcore.dataobject.postPublishFailure
- pimcore.dataobject.preUnpublish
- pimcore.dataobject.postUnpublish
- pimcore.dataobject.postUnpublishFailure

The current documentation would cover this feature in my opinion.


brusch:
To be honest, I think this can and should be done already with the existing update events and we shouldn't create dedicated events for single actions such as publishing, renaming, moving, ... 
So I'm closing this PR for now, but feel free to start a discussion about it - thanks anyway for the PR 😊

PR #11731, opened 2022-03-22
[Maintenance] Log job start

BlackbitDevs:
Currently when running `bin/console maintenance -vv` you only see which jobs are finished or skipped. If you see that the maintenance command takes a lot of resources / time, it would help to know which job causes this. Thus this PR adds a log when starting a certain maintenance job. This way you can call  `bin/console maintenance -vv` and when it slows down, you see which job is responsible.


mcop1:
Thanks again!

PR #11733, opened 2022-03-22
Do not ignore creating translations table on `TableNotFoundException`

NiklasBr:
Resolves issue where adding new translation domains would silently fail because of a DBAL error.


dvesh3:
@NiklasBr thanks!

PR #11735, opened 2022-03-22
Setup Content Security Policy header based on new handler.

aryaantony92:
Follow up to pimcore/pimcore#11447

## Additional info  

Added the whitelisting for the urls. But the page is not loaded since there is some issue with the inline scripts in external opcache page.


NiklasBr:
Shouldn't these be added to the [default configuration](https://symfony.com/doc/current/components/config/definition.html#default-and-required-values) rather than in the code?

aryaantony92:
No, we don't want to override the default configuration

NiklasBr:
Why?

dvesh3:
@NiklasBr because these URLs should be merged with the ones provided from bundle configs. These are whitelisted here for Opcache view in admin interface.

PR #11739, opened 2022-03-22
fix(images): Unhandled stream exception

marlonbasten:
## Changes in this pull request  Resolves
With this fix it informs the developer that they may have forgot to install the imagick plugin.

## Additional info  
When trying to load an SVG file the following function throws a null error without providing any information why.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -1355,6 +1355,11 @@ class AssetController extends ElementControllerBase implements KernelControllerE
         }
 
         $stream = $thumbnail->getStream();
+        
+        if ($stream === null) {
+            throw new \Exception("Could not generate the image thumbnail. Maybe you forgot to install the needed extension?");
+        }
+        
         $response = new StreamedResponse(function () use ($stream) {
             fpassthru($stream);
         }, 200, [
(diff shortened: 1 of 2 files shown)


brusch:
Hmm, don't see this connected with `Imagick` - also `getStream()` can return `null` - so probably this is the wrong place for your fix?

marlonbasten:
@brusch Yeah, that's exactly the problem that getStream() can return null. It throws a null exception on the get-image-thumbnail route in the AdminBundle when trying to load SVGs without having Imagick installed.

marlonbasten:
image

That's the error I'm trying to make clearer for users.

brusch:
Thanks for your feedback. 
Still, I'd add the `null` check in the action and log the message there if the thumbnail couldn't be generated (whatever the reason is, could be other image formats such as PSD as well). In this case the action should return the error image. 
Can you please update the PR accordingly? Thanks a lot!

brusch:
@marlonbasten 🏓

marlonbasten:
Sorry for the late response! Will do a PR tomorrow, thanks for the ping!

brusch:
@marlonbasten 🏓🙃 Thanks!

brusch:
@marlonbasten last call 😉 Thanks in advance!

marlonbasten:
@brusch Should be done now. I didn't read the "action should return the error image" part. Thanks for being so patient with me :)

brusch:
Just directly added the missing annotation reported by PhpStan.

brusch:
Just directly added the missing annotation reported by PhpStan.

brusch:
Thank you very much!

PR #11742, opened 2022-03-23
 Handle aborted requests to stop spamming console errors

croitoruion:
If we make several ajax requests that abort previous one (i.e. press next page several time) it will spam the console with errors from ext.js, because it try to decode "undefined"-json.
If the request is aborted, don't raise the error from ext by sending safe flag & also send console.log instead of console.error from startup.js.

  
  

Resolves #

## Additional info  
Before & after
:: Pimcore 2022-03-23 13-32-17


brusch:
Totally makes sense, thanks a lot! 👍😊

PR #11755, opened 2022-03-24
Update Additional Tools Installation for GitHub source URL:s

NiklasBr:
`wget` against `blob` will download the HTML page. We want the `raw` part with the actual binary content.

  
  

Resolves #


fashxp:
thx

PR #11758, opened 2022-03-24
Fix cache argument to be priority instead of lifetime

jdreesen:
I'm pretty sure the last argument of the `Cache::save()` call is meant to be the priority, like in many other places, instead of the lifetime, for which it would be a rather strange value.


brusch:
Very nice find - thanks a lot! 
... we could even think about using `$force = true` in this case, but then prio `999` is quite high, so should be ok as well 😉

PR #11762, opened 2022-03-25
[DataObject] Classdefinitions - Add BC layer for childs property

dvesh3:
Follow up to https://github.com/pimcore/pimcore/pull/10824

## Additional info  
For some bundles, the `childs` property is accessed in JS, which leads to exception when childs is missing. please see https://github.com/pimcore/asset-metadata-class-definitions/issues/50


blankse:
For JS I created the getLayoutDefinitionArray method in the DataObjectController class. Why does this not work? https://github.com/blankse/pimcore/blob/10.x/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L2356-L2390

With this reference the response will increase very much in complex layouts, see my comments in https://github.com/pimcore/pimcore/pull/10824

dvesh3:
> For JS I created the getLayoutDefinitionArray method in the DataObjectController class. Why does this not work? https://github.com/blankse/pimcore/blob/10.x/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L2356-L2390
> 
> With this reference the response will increase very much in complex layouts, see my comments in #10824

This BC layer only works in case of `/object/get` calls and for other bundles to work they need to introduce this change and then they are broken for 6.9 version (if supported), which is the case at the moment with pimcore bundles `pimcore/asset-metadata-class-definitions` & `pimcore/output-data-config-toolkit`. yes, the response is bigger but i couldn't think of a better solution.

mcop1:
@dvesh3 I think there is a problem with localized fields. I created an asset metadata set and applied it to an asset in 10.3:
image

After that I switched to 10.x and, as expected, it wasn´t working anymore:
image

And then I applied this pr, this is the result (nr. field is displayed correctly, localized fields are missing)
image

Edit: Applying https://github.com/pimcore/asset-metadata-class-definitions/pull/51 doesn´t help

dvesh3:
> @dvesh3 I think there is a problem with localized fields. I created an asset metadata set and applied it to an asset in 10.3: image
> 
> After that I switched to 10.x and, as expected, it wasn´t working anymore: image
> 
> And then I applied this pr, this is the result (nr. field is displayed correctly, localized fields are missing) image
> 
> Edit: Applying [pimcore/asset-metadata-class-definitions#51](https://github.com/pimcore/asset-metadata-class-definitions/pull/51) doesn´t help

@mcop1 after merging 1.2 branch on 1.x on [asset-metadata-class-definitions](https://github.com/pimcore/asset-metadata-class-definitions), it works again.

PR #11769, opened 2022-03-28
Fix non "thumbnail-able" images to be prefixed (eg. SVG's)

dpfaffenbauer:
SVG's don't get thumbnailed but should also be serveable via prefixes from CDN's or Blob Storages. In my case it is Google Cloud Storage. Configuring Pimcore like mentioned in the Docs doesn't work with SVGs. This fixes it.


brusch:
#11777 is related

brusch:
Thank you very much - while reviewing I stumbled upon #11777 as well 😉

dpfaffenbauer:
💪

PR #11770, opened 2022-03-28
[Grid column config] For admin users allow to show invisible fields

BlackbitDevs:
Currently data object fields which are set to be `invisible` can not be shown in the grid. In the edit view there is the special `Admin mode` which enables admin users to see invisible fields. In grid column config this does not exist.
With this PR admin users are allowed to add invisible fields to grid column config. 
If such a grid column config gets shared with non-admin users, those will not see invisible fields.

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -1336,11 +1336,18 @@ class Service extends Model\Element\Service
     {
         if ($layout instanceof ClassDefinition\Data) {
             $name = $layout->getName();
-            if (empty($fieldDefinitions[$name]) || $fieldDefinitions[$name]->getInvisible()) {
+            if (empty($fieldDefinitions[$name])) {
                 return false;
             }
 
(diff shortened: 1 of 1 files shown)


mcop1:
@BlackbitNeueMedien thank you for your pr, but right now we would prefer to not implement that. The suggested approach would be to set the field to visible and create 2 layouts. One which doesn´t include the field and gets distributed to the users and one which does include the field and gets distributed to the admins.

Nevertheless thank you for submitting this pr.

BlackbitDevs:
If there is a temporary problem and you only want to look up this invisible field in the grid, it seems to be a too long way to having to create a custom layout first...
Why shall the grid work differently than the edit view (where you have the "master mode")?

mcop1:
@BlackbitNeueMedien because the master mode is a separate mode that needs to be explicitly activated by the user and does more than just showing invisible fields.

PR #11771, opened 2022-03-28
Add possibility to save version note via ajax

croitoruion:
Resolves #

## Additional info  
By adding this we can call saving objects & asset, with new param via ajax "versionNote"

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -1023,7 +1023,9 @@ class AssetController extends ElementControllerBase implements KernelControllerE
                 // save to session only
                 Asset\Service::saveElementToSession($asset);
             } else {
-                $asset->save();
+                $asset->save([
+                    'versionNote' => $request->get('versionNote') ?? null
+                ]);
             }
 
             $treeData = $this->getTreeNodeConfig($asset);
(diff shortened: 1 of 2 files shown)


brusch:
Changed base to `10.x` as it isn't a bug.

brusch:
Can you please describe the use case a bit more in detail? Thanks!

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -1023,7 +1023,9 @@ public function saveAction(Request $request, EventDispatcherInterface $eventDisp
                 // save to session only
                 Asset\Service::saveElementToSession($asset);
             } else {
-                $asset->save();
+                $asset->save([
+                    'versionNote' => $request->get('versionNote') ?? null
```
```suggestion
                    'versionNote' => $request->get('versionNote')
```
` ?? null` is redundant, as `->get()` returns anyway `null` 🙃

croitoruion:
> Can you please describe the use case a bit more in detail? Thanks!

We call save() objects & assets via AJAX, but we can't save it with versionNote, so by adding handling for new request param we resolve this.

brusch:
Thanks for clarifying 😊 However, the endpoints of the Pimcore admin UI are not intended to be used by custom implementations or bundles as they do not ensure backward compatibility. Therefore I'd suggest you to create a custom endpoint supporting the required functionalities for your project. We're not adding functionalities to the admin UI which are actually not used by Pimcore itself, because this makes it impossible to maintain long term.

PR #11780, opened 2022-03-28
fix: PHP 8.1 doesn't allow passing null to explode()

dkarlovi:
## Additional info  

When on PHP 8.1, this produces a deprecation:
```
Deprecated: explode(): Passing null to parameter #2 ($string) of type string is deprecated
```

### DIFF
--- a/models/Staticroute.php
+++ b/models/Staticroute.php
@@ -215,9 +215,14 @@ final class Staticroute extends AbstractModel
      */
     private function getDefaultsArray()
     {
-        $defaults = [];
+        $array = $this->getDefaults();
+        if ($array === null) {
+            return [];
+        }
+        
+        $defaults = [];      
 
-        $t = explode('|', $this->getDefaults());
+        $t = explode('|', $array);
         foreach ($t as $v) {
             $d = explode('=', $v);
             if (strlen($d[0]) > 0 && strlen($d[1]) > 0) {
--- a/models/Staticroute.php
+++ b/models/Staticroute.php
@@ -216,7 +216,7 @@ final class Staticroute extends AbstractModel
     private function getDefaultsArray()
     {
         $array = $this->getDefaults();
-        if ($array === null) {
+        if (empty($array)) {
             return [];
         }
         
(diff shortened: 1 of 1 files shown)


models/Staticroute.php
```diff
@@ -215,9 +215,14 @@ public static function create()
      */
     private function getDefaultsArray()
     {
-        $defaults = [];
+        $array = $this->getDefaults();
+        if ($array === null) {
```
brusch:
```suggestion
        if (empty($array)) {
```
Using `empty()` would cover empty strings as well.

brusch (changes requested):

models/Staticroute.php
dkarlovi: I don't really use `empty` because it's too lenient (exactly how you show), one exception is checking for property initializations. Accepted your suggestion. :+1:

models/Staticroute.php
```diff
@@ -215,9 +215,14 @@ public static function create()
      */
     private function getDefaultsArray()
     {
-        $defaults = [];
+        $array = $this->getDefaults();
```
jdreesen: `$array` is a bit strange as a variable name, considering that the return type of `getDefaults()` is `string`.

dkarlovi: @jdreesen good point!

brusch:
```suggestion
        $defaultsString = $this->getDefaults();
```

models/Staticroute.php
```diff
@@ -215,9 +215,14 @@ public static function create()
...
+        if (empty($array)) {
```
brusch:
```suggestion
        if (empty($defaultsString)) {
```

models/Staticroute.php
```diff
@@ -215,9 +215,14 @@ public static function create()
...
+            return [];
+        }
+        
+        $defaults = [];      
 
-        $t = explode('|', $this->getDefaults());
+        $t = explode('|', $array);
```
brusch:
```suggestion
        $t = explode('|', $defaultsString);
```

brusch (changes requested):

models/Staticroute.php
brusch: @jdreesen valid point, renamed it, thanks 🙃

brusch:
Thanks!

Issue #11687, opened 2022-03-16
[Improvement]: PHP 8.1 deprecation messages


Ricardocanul7:
### Improvement description

When running tests in a project with PHP8.1 I receive a lot of deprecation messages from Pimcore code

a few of the messages are:

`Deprecated: strlen(): Passing null to parameter #1 ($string) of type string is deprecated`
`Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated`
`Deprecated: strip_tags(): Passing null to parameter #1 ($string) of type string is deprecated`
`Deprecated: array_slice(): Passing null to parameter #2 ($offset) of type int is deprecated`
`Deprecated: strip_tags(): Passing null to parameter #1 ($string) of type string is deprecated`
`Deprecated: PDO::quote(): Passing null to parameter #1 ($string) of type string is deprecated`
`Deprecated: explode(): Passing null to parameter #2 ($string) of type string is deprecated`

fashxp:
also check why phpstan is not catching that.

blankse:
@fashxp Parameter types are checked by PhpStan in level 5. We are on level 4.
https://phpstan.org/user-guide/rule-levels

fashxp:
also related: 
- https://github.com/pimcore/pimcore/pull/11792 
- https://github.com/pimcore/pimcore/pull/11785
- https://github.com/pimcore/pimcore/pull/11786
- https://github.com/pimcore/pimcore/pull/11787
- https://github.com/pimcore/pimcore/pull/11788
- https://github.com/pimcore/pimcore/pull/11789

aryaantony92:
@Ricardocanul7  Could you please provide the logs regarding the deprecated error messages in PHP 8.1 ?

aryaantony92:
Could you please provide the complete logs regarding the deprecation error messages in PHP 8.1 ?

fashxp:
closed due to inactivity. please come back, if they still occur with more insights. thx.

PR #11782, opened 2022-03-28
fix: PHP 8.1 deprecation: passing null to strlen()

dkarlovi:
## Additional info  

```
Deprecated: strlen(): Passing null to parameter #1 ($string) of type string is deprecated
```

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Extension/Text.php
+++ b/models/DataObject/ClassDefinition/Data/Extension/Text.php
@@ -36,7 +36,7 @@ trait Text
      */
     public function isEmpty($data)
     {
-        return strlen($data) < 1;
+        return empty($data) || strlen($data) < 1;
     }
 
     /**
--- a/models/DataObject/ClassDefinition/Data/Extension/Text.php
+++ b/models/DataObject/ClassDefinition/Data/Extension/Text.php
@@ -36,7 +36,7 @@ trait Text
      */
     public function isEmpty($data)
     {
-        return !is_string($data) || strlen($data) < 1;
+        return strlen((string) $data) < 1;
     }
 
     /**
--- a/models/DataObject/ClassDefinition/Data/Extension/Text.php
+++ b/models/DataObject/ClassDefinition/Data/Extension/Text.php
@@ -36,7 +36,7 @@ trait Text
      */
     public function isEmpty($data)
     {
-        return empty($data) || strlen($data) < 1;
+        return !is_string($data) || strlen($data) < 1;
     }
 
     /**


dkarlovi:
@brusch these seem like they should be caught by PHPStan? Is the level currently too low?

jdreesen:
Yes it is, see https://github.com/pimcore/pimcore/issues/11687#issuecomment-1073924029

models/DataObject/ClassDefinition/Data/Extension/Text.php
```diff
@@ -36,7 +36,7 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
      */
     public function isEmpty($data)
     {
-        return strlen($data) < 1;
+        return empty($data) || strlen($data) < 1;
```
jdreesen: I'm not entirely sure where this is used, but `empty` is dangerous, as it the string `"0"` is considered empty, too (see: https://3v4l.org/mgSdd).

dkarlovi: Right, which is why I don't use it, but was suggested by @brusch in the other PR. :) I guess this needs some more thinking.

jdreesen: Well, `$data` is expected to be `string` or `null`, so why not just cast it to `string` then? => https://3v4l.org/57cNL

dkarlovi: I try to avoid type casts because they're quite fragile. It's better to handle the intended types directly in the code IMO. If you know null is possible here, check for null. If others things start creeping in, that will allow it to fail, but casting will silently "help".

blankse: $data should be string or null. So this should be enough:
```suggestion
        return $data === null || strlen($data) < 1;
```
Edit: Alternative we could check for string
```suggestion
        return !is_string($data) || strlen($data) < 1;
```

brusch: Using `!is_string($data)` is dangerous here, as it is a possible BC break - see failing tests.
Reason: currently it also works with numeric data such as `2345` which is not a `string` and would treat it as empty.

models/DataObject/ClassDefinition/Data/Extension/Text.php
```diff
@@ -36,7 +36,7 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
      */
     public function isEmpty($data)
     {
-        return strlen($data) < 1;
+        return !is_string($data) || strlen($data) < 1;
```
brusch:
```suggestion
        return strlen((string) $data) < 1;
```
I'd go for a cast here instead, which covers numeric data as well.

brusch (changes requested):

brusch:
Thanks!

PR #11783, opened 2022-03-28
Update Concrete Exception Throwing

dpfaffenbauer:
`$e` can be anything here, assuming it has 3 constructor parameters is wrong.


brusch:
🍒-picked into `10.3` as I'd classify this as a bug (24e91964290ddaeb362d955cee732f6671830c48)

dpfaffenbauer:
thanks

PR #11784, opened 2022-03-28
Fix strlen deprecation in User

dpfaffenbauer:


models/User.php
```diff
@@ -147,7 +147,7 @@ public function getPassword()
      */
     public function setPassword($password)
     {
-        if (strlen($password) > 4) {
+        if (!empty($password) && strlen($password) > 4) {
```
blankse, line 150:
```suggestion
        if (is_string($password) && strlen($password) > 4) {
```
What value was $password? null? Than we should also update the phpdoc.

dpfaffenbauer: What do you mean? $password is supposed to be a string, as the PHPDoc also says.

blankse: When it is always a string, than we should not need this check. It must be a other type.

brusch: @blankse is right, must have been something other than `string` - I'd assume it was `null`.

brusch:
Closed in favor of #11792

PR #11787, opened 2022-03-28
Fix deprecation in Numeric

dpfaffenbauer:


models/DataObject/ClassDefinition/Data/Numeric.php
```diff
@@ -561,7 +561,7 @@ public function isDiffChangeAllowed($object, $params = [])
      */
     public function isEmpty($data)
     {
-        return strlen($data) < 1;
+        return empty($data) || strlen($data) < 1;
```
jdreesen: Not sure if this is right, as `"0"` is considered empty (see: https://3v4l.org/mgSdd).

dpfaffenbauer: true, lets wait on how we resolve that on the other PR

dpfaffenbauer: I think we should check if $data is actually a number, if not it is invalid.

brusch:
```suggestion
        return is_numeric($data);
```
This should do the trick in this case.

brusch (changes requested):

models/DataObject/ClassDefinition/Data/Numeric.php
```diff
@@ -561,7 +561,7 @@ public function isDiffChangeAllowed($object, $params = [])
      */
     public function isEmpty($data)
     {
-        return strlen($data) < 1;
+        return is_numeric($data);
```
brusch:
```suggestion
        return !is_numeric($data);
```

dpfaffenbauer:
thanks @brusch

PR #11789, opened 2022-03-28
Fix deprecation in Select

dpfaffenbauer:


models/DataObject/ClassDefinition/Data/Select.php
```diff
@@ -384,7 +384,7 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
      */
     public function isEmpty($data)
     {
-        return strlen($data) < 1;
+        return empty($data) || strlen($data) < 1;
```
brusch:
```suggestion
        return strlen((string) $data) < 1;
```
Casting to string is probably the best in this case. Because the currently implementation would still return `true` for value `0`, which is wrong.

dpfaffenbauer:
thanks @brusch

PR #11792, opened 2022-03-29
[User] Fix strlen deprecation in setPassword()

brusch:
Alternative approach for #11784 on the right target branch `10.3`

### DIFF
--- a/models/User.php
+++ b/models/User.php
@@ -33,7 +33,7 @@ final class User extends User\UserRole
     protected $type = 'user';
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $password;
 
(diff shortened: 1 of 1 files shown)


models/User.php
```diff
@@ -133,21 +133,21 @@ final class User extends User\UserRole
...
     /**
-     * @param string $password
+     * @param string|null $password
      *
      * @return $this
      */
     public function setPassword($password)
     {
-        if (strlen($password) > 4) {
+        if (is_string($password) && strlen($password) > 4) {
```
brusch:
```suggestion
        if (strlen((string) $password) > 4) {
```

PR #11793, opened 2022-03-29
[Data Object] Multiselect follow up fix

brusch:
see #11788


dkarlovi:
@brusch I think Pimcore codebase relies way too much on casting: https://3v4l.org/984YH

It will make using static analysis tools almost worthless in the future because you rely on type coercion, basically `==` instead of `===` everywhere. Type casting should be the exception, not the default solution. /cc @jdreesen

brusch:
Yes, I understand the concern, better approaches (not breaking the system) are always welcome as a PR ☺️
More strict typing we can only implement with major versions.

PR #11796, opened 2022-03-29
Deprecation fix for Data

dpfaffenbauer:


dpfaffenbauer:
@brusch @aryaantony92 any ETA on this?

blankse:
Maybe we should return always a string in the getTitle() method? Whats the advantage to have it nullable?

brusch:
@dpfaffenbauer I agree with @blankse : I also don't see a reason for having the title nullable.

brusch:
Btw. we should base this on `10.3` ☺️

dpfaffenbauer:
@brusch done

brusch:
Thanks a lot! 👍

PR #11805, opened 2022-03-30
[Admin UI] Unlock elements when browser window gets closed

BlackbitDevs:
When an element gets opened in Pimcore admin UI, it gets locked so a message gets shown to other users that this element is already open. 

When the element tabs get closed, those locks get released.

But when a user simply closes the browser, the locks do not get released. On the one hand this causes the funny effect that you get messages that the current user himself has the object currently open (even when using only one browser window). And on the other hand this irritates other users because they may not dare to change anything although the other user actually already closed his browser (and with it all the elements).


dvesh3:
@BlackbitNeueMedien This does not solve the problem when second user is trying to open the element without reloading the admin interface as elements are unlocked on startup...right? If yes, then we should think of an approach as this could be a common scenario that the other users tries to open the element without reloading the interface.

dvesh3:
@BlackbitNeueMedien 🏓

BlackbitDevs:
@dvesh3 Sorry, I do not understand your question. Do you mean that 2 users are logged in to Pimcore backend, user A opens an element, then closes the browser and then user B opens the same element? Exactly in this case this PR will work because when user A closes the browser the element gets unlocked.

dvesh3:
> @dvesh3 Sorry, I do not understand your question. Do you mean that 2 users are logged in to Pimcore backend, user A opens an element, then closes the browser and then user B opens the same element? Exactly in this case this PR will work because when user A closes the browser the element gets unlocked.

I mean there are 2 users (on different machines), User A opens an element, then closes the browser and user B opens the element sees the error message. This solves the problem when 2 users are using the same machine but still a problem when machines are different.

BlackbitDevs:
> I mean there are 2 users (on different machines), User A opens an element, then closes the browser and user B opens the element sees the error message. 

Why? When user A closes the browser the edit lock gets released with this PR.

dvesh3:
> > I mean there are 2 users (on different machines), User A opens an element, then closes the browser and user B opens the element sees the error message.
> 
> Why? When user A closes the browser the edit lock gets released with this PR.

@BlackbitNeueMedien please read [this ](https://github.com/pimcore/pimcore/pull/11805#issuecomment-1092651020)again.

I mean the user B is on different machine so his `pimcore_opened_elements` local storage will not have id of the closed element on user A machine. how this will be unlocked on user B machine?

BlackbitDevs:
Somehow there is a misunderstanding here ;-)

The workflow with 2 users on different machines is:
1. User A opens element X -> element X gets added to user A's localStorage `pimcore_opened_elements`, see https://github.com/pimcore/pimcore/pull/11805/files#diff-a63c12be882eef30c0dcfacd324ed3a2840b65340a7e156f6cc7eb180a3699c7R195 and https://github.com/pimcore/pimcore/pull/11805/files#diff-a63c12be882eef30c0dcfacd324ed3a2840b65340a7e156f6cc7eb180a3699c7R83
2. User A closes his browser -> localStorage `pimcore_opened_elements` gets used to unlock element X via https://github.com/pimcore/pimcore/pull/11805/files#diff-7e133e0be50ec93c2432982e19ac64d5c27e2cdc286a1e1a7e0bdd40ca61582dR152 and https://github.com/pimcore/pimcore/pull/11805/files#diff-d5ef24fccda46c5099d4a78e84d7bffbbec24c67bfecb385dbddebba168c4442R76 - so element X is really not locked anymore, also server-side
3. User B opens element X -> no lock present -> no edit lock warning

So actually the localStorage `pimcore_opened_elements` is similar to https://github.com/pimcore/pimcore/blob/ffe18e9a90566cd0c5c9e488cbc6314730372329/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js#L885-L891 and in fact I would use this - but it is limited to 10 entries, for this reason I introduced a new localStorage variable. An alternative would be to remove the limit of 10 items for `pimcore_opentabs` localStorage entry and only load the latest 10 in https://github.com/pimcore/pimcore/blob/ffe18e9a90566cd0c5c9e488cbc6314730372329/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js#L918

And now I re-read 
> This does not solve the problem when second user is trying to open the element without reloading the admin interface as elements are unlocked on startup...right?

No, this is not true, the elements get unlocked on `beforeunload` JS event, see https://github.com/pimcore/pimcore/pull/11805/files#diff-7e133e0be50ec93c2432982e19ac64d5c27e2cdc286a1e1a7e0bdd40ca61582dR138. This gets triggered when the browser gets closed.

dvesh3:
> No, this is not true, the elements get unlocked on beforeunload JS event, see https://github.com/pimcore/pimcore/pull/11805/files#diff-7e133e0be50ec93c2432982e19ac64d5c27e2cdc286a1e1a7e0bdd40ca61582dR138. This gets triggered when the browser gets closed.

Gotcha!! I was looking for this answer before 😄 thanks for clarifying 👍.

dvesh3:
> An alternative would be to remove the limit of 10 items for pimcore_opentabs localStorage entry and only load the latest 10 in

yes, I would go for this approach. Could you please change it? thanks!

dvesh3:
🏓 @BlackbitNeueMedien

dvesh3 (changes requested):
@BlackbitNeueMedien the fix is not working anymore after the refactor. could you please check? thanks!

BlackbitDevs:
@dvesh3 Sorry, now it works.

BlackbitDevs:
@dvesh3 Do you know how to fix the PhpStan problems? Imho those are not caused by this PR, are they?

blankse:
The error came with the new PhpStan Version 1.6.0. I fixed it here https://github.com/pimcore/pimcore/pull/11965

dvesh3:
@BlackbitNeueMedien please rebase on 10.x branch to fix PhpStan checks. thanks!

dvesh3:
@BlackbitNeueMedien sorry, still not working. could you please check again? thanks!

BlackbitDevs:
What exactly does not work, @dvesh3 ? Actually I tested it this morning and when I close the browser the locks get removed from `edit_lock` table.

dvesh3:
@BlackbitNeueMedien for me the lock is **not** getting removed from `edit_lock` table. It is hard to debug the server call as browser get's closed.

BlackbitDevs:
I debugged with browser window reloading - there at least the browser console can be used to analyse the request. Will try again with another browser...

dvesh3:
@BlackbitNeueMedien any chance to look at it? I would like to include this in the 10.4 release today 😉

BlackbitDevs:
Which browser have you used, @dvesh3 ? I tested in Firefox and this worked. Will test now again in Chrome as it may be a problem with sending requests during `onbeforeunload` event handler.

dvesh3:
@BlackbitNeueMedien I used **Chrome 89.0.4389.90**

BlackbitDevs:
Yes, can confirm, that it does not work in Chrome. Checking what the problem is...

BlackbitDevs:
Chrome blocks nearly everything in `onbeforeunload`: https://stackoverflow.com/questions/9626059/window-onbeforeunload-in-chrome-what-is-the-most-recent-fix
Hmmm...

dvesh3:
I'm wondering it worked before (on same chrome version) the refactoring 😄

BlackbitDevs:
@dvesh3 Can't really believe that ir worked before. I get error 
`Failed to execute 'send' on 'XMLHttpRequest': Failed to load 'http://localhost:2000/admin/element/unlock-elements': Synchronous XHR in page dismissal. See https://www.chromestatus.com/feature/4664843055398912 for more details.`
in browser console.

BlackbitDevs:
It may work with `navigator.sendBeacon()` Currently struggling with "Detected CSRF Attack! Do not do evil things with pimcore ... ;-)"

BlackbitDevs:
@dvesh3 Now it works in Firefox and Chrome (and hopefully all other modern browsers)

dvesh3 (approved):

dvesh3:
@BlackbitNeueMedien works like a charm 😉 thanks a lot!

PR #11806, opened 2022-03-30
Ignore not found properties in deep copies

### LINKED ISSUE #11801: [Bug]: cannot publish documents that where created with pimcore 6.9
### Expected behavior

I want to publish a document which was created before the update to pimcore 10.x (currently on 10.2.5)

Only happens if the document was never published.

### Actual behavior

I get the error `cannot access property starting with \0`.

The error is thrown here: https://github.com/pimcore/pimcore/blob/10.x/models/Element/Service.php#L1227
called from `Pimcore\Model\Document\PageSnippet::checkMissingRequiredEditable` 

### Steps to reproduce

(migration script was executed prior to pimcore 10 update)

I think it is because the version has a serialized Page Object which has still properties `module` `action` and `elements` but these where removed. 

I'm not sure how to fix it, because I have no idea how to remove properties from an unserialized object of a class that does not have these properties (or serialized without some very janky regex foo). If someone has an idea, i can create a PR for that.

useful grep command for finding the affected versions: `grep -rnw './var/versions/document' -e 'O:27:"Pimcore\\Model\\Document\\Page":36'`

jheimbach:
Resolves #11801

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -1224,7 +1224,12 @@ class Service extends Model\AbstractModel
                     return false;
                 }
                 $reflectionProperty->setAccessible(true);
-                $myValue = $reflectionProperty->getValue($object);
+
+                try {
+                    $myValue = $reflectionProperty->getValue($object);
+                } catch (\Throwable $e) {
+                    return false;
+                }
 
                 return $myValue instanceof ElementInterface;
             }
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -1220,12 +1220,7 @@ class Service extends Model\AbstractModel
             {
                 try {
                     $reflectionProperty = new \ReflectionProperty($object, $property);
-                } catch (\Exception $e) {
-                    return false;
-                }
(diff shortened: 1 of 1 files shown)


models/Element/Service.php
```diff
@@ -1224,7 +1224,12 @@ public function matches($object, $property)
                     return false;
                 }
                 $reflectionProperty->setAccessible(true);
-                $myValue = $reflectionProperty->getValue($object);
+
+                try {
```
dvesh3: I would keep these in one try catch for cleaner code:
```
try {
    $reflectionProperty = new \ReflectionProperty($object, $property);
    $reflectionProperty->setAccessible(true);
    $myValue = $reflectionProperty->getValue($object);
} catch (\Exception $e) {
    return false;
}
```

jheimbach: i merged the try catch blocks.
notice that we need to catch \Throwable instead of \Exception

dvesh3:
@JHeimbach thanks!

PR #11809, opened 2022-03-30
[Bug]: Video thumbnail generation fails on remote filesystems 

### LINKED ISSUE #11645: [Bug]: Video thumbnail generation failes in distributed environments
### Expected behavior

The message queue consumer should be able to create a video thumbnail regardless of the container/vm that created the message originally.

### Actual behavior

The video processor [creates a local temporary copy](https://github.com/pimcore/pimcore/blob/10.x/models/Asset/Video/Thumbnail/Processor.php#L130) of the video before passing the processing job to the message queue. In cases where the queue consumer does not run in the very same environment, the processing job fails. This can be the case when you have a separate container running the consumer, or multiple VMs behind a load balancer. The consumer might not have access to the same temporary directory that was used in the original request, thus failing with a "file not found" error.

I would suggest to move the creation of the temporary local file to the processing of the queue message and not do that beforehand. What do you think?

### Steps to reproduce

To simulate this behavior, have the queue worker use a different `PIMCORE_SYSTEM_TEMP_DIRECTORY` then the pimcore instance.
Alternatively, you can create two separate containers, one for pimcore, one for the worker. Once you trigger video thumbnail generation in pimcore, the processing job in the worker will fail if the two containers do not share the same temporary directory.

bbdammert:
Resolves #11645

## Additional info  
Fix the issue of missing source file in
Pimcore\Model\Asset\Video\Thumbnail\Processor::execute() by
 - creating a new temporary file from the original asset and
 - passing the new source file to the converter adapter(s).

Additionally, fix a minor issue with duplicate slashes in the path to
the thumbnail directory.

### DIFF
--- a/models/Asset/Video/Thumbnail/Processor.php
+++ b/models/Asset/Video/Thumbnail/Processor.php
@@ -131,7 +131,7 @@ class Processor
         $instance->setDeleteSourceAfterFinished($sourceFile);
 
         foreach ($formats as $format) {
-            $thumbDir = $asset->getRealPath() . '/video-thumb__' . $asset->getId() . '__' . $config->getName();
+            $thumbDir = rtrim($asset->getRealPath(), '/') . '/video-thumb__' . $asset->getId() . '__' . $config->getName();
             $filename = preg_replace("/\." . preg_quote(File::getFileExtension($asset->getFilename()), '/') . '/', '', $asset->getFilename()) . '.' . $format;
             $storagePath = $thumbDir . '/' . $filename;
             $tmpPath = File::getLocalTempFilePath($format);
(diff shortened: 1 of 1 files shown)


brusch:
@bbdammert can you please sign the CLA? Thanks a lot :)

brusch:
@bbdammert Please also check the failing tests: https://github.com/pimcore/pimcore/runs/5818105638?check_suite_focus=true#step:8:23

jdreesen:
> Please also check the failing tests: [pimcore/pimcore/runs/5818105638?check_suite_focus=true#step:8:23](https://github.com/pimcore/pimcore/runs/5818105638?check_suite_focus=true#step:8:23)

It just needs a rebase onto the latest `10.3`.

brusch:
@jdreesen ah, yes, thanks for clarifying - @bbdammert CLA would be the most important, rebase: just if it's easy for you 😉

brusch:
@bbdammert Thanks a lot for your PR, however there's a small issue with it: 
The change causes orphaned temp files, which are not cleaned up again, because actually the temp file created here https://github.com/pimcore/pimcore/blob/6b706aacc16ae0b51a62f4dcc16af24b94f7140b/models/Asset/Video/Thumbnail/Processor.php#L131 is getting overridden and therefore not clean up anymore. 

I've created an alternative PR fixing the issue, would be nice if you can double check and confirm that it fixes the issue for you as well. 

Thanks again! 👍

brusch:
@bbdammert forgot to link the alternative PR 🙃 here it is #11862

PR #11810, opened 2022-03-30
Static Page from Site needs Current Site

dpfaffenbauer:
Static Pages Resolver for Site Pages need to have Sites resolved and use them to serve them. This adds a bit of overhead and maybe looses a bit of performance, but without it, site Static Pages don't work.

### DIFF
--- a/bundles/CoreBundle/EventListener/Frontend/StaticPageGeneratorListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/StaticPageGeneratorListener.php
@@ -98,11 +98,12 @@ class StaticPageGeneratorListener implements EventSubscriberInterface
             $filename = urldecode($request->getPathInfo());
 
             if (Site::isSiteRequest()) {
-                $path = Site::getCurrentSite()->getRootPath();
-
                 if ($request->getPathInfo() === '/') {
                     $path = '';
                     $filename = '/' . Site::getCurrentSite()->getRootDocument()->getKey();
+                }
+                else {
+                    $path = Site::getCurrentSite()->getRootPath();
                 }
             }
             $filename = $path .  $filename  . '.html';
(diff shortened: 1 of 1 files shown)


bundles/CoreBundle/EventListener/Frontend/StaticPageGeneratorListener.php
```diff
@@ -93,7 +94,19 @@ public function onKernelRequest(RequestEvent $event)
         $storage = Storage::get('document_static');
 
         try {
-            $filename = urldecode($request->getPathInfo()) . '.html';
+            $path = '';
+            $filename = urldecode($request->getPathInfo());
+
+            if (Site::isSiteRequest()) {
+                $path = Site::getCurrentSite()->getRootPath();
```
dvesh3: I would move this in else part to save an extra call in case of `$request->getPathInfo() === '/'`

dpfaffenbauer: @dvesh3 good catch. done

bundles/CoreBundle/EventListener/Frontend/StaticPageGeneratorListener.php
```diff
@@ -93,7 +94,20 @@ public function onKernelRequest(RequestEvent $event)
         $storage = Storage::get('document_static');
 
         try {
-            $filename = urldecode($request->getPathInfo()) . '.html';
+            $path = '';
+            $filename = urldecode($request->getPathInfo());
+
+            if (Site::isSiteRequest()) {
+                if ($request->getPathInfo() === '/') {
+                    $path = '';
```
dvesh3: no need of re-initialization
```suggestion
```

dvesh3:
@dpfaffenbauer thanks!

PR #11812, opened 2022-03-31
[Improvement] Copy the icon path by clicking the icon

ogmdev:
Since adding inline svg icons to html to reduce concurrent requests ( #11089 ), the icon's url is no longer easily accessible.

This PR adds the possibility to retrieve the relative path to the icon in the clipboard by clicking on the image.

### DIFF
--- a/bundles/AdminBundle/Resources/views/Admin/Misc/iconList.html.twig
+++ b/bundles/AdminBundle/Resources/views/Admin/Misc/iconList.html.twig
@@ -51,6 +51,9 @@
             width: 50px;
         }
 
+        .language-icon img{
+            width: 16px;
+        }
     </style>
 </head>
 <body>
@@ -126,7 +129,15 @@
     {% endfor %}
 </table>
 
+{{ pimcore_minimize_scripts(['pimcore/common.js', 'pimcore/functions.js','pimcore/helpers.js']) | raw }}
+<script>
+    [...document.getElementsByTagName('img')]
+        .forEach(el => {
+            el.addEventListener("click", () => {
+                pimcore.helpers.copyStringToClipboard(el.dataset.imgpath)
+            });
+        });
+</script>
 
 </body>
 </html>
-
--- a/bundles/AdminBundle/Resources/views/Admin/Misc/iconList.html.twig
+++ b/bundles/AdminBundle/Resources/views/Admin/Misc/iconList.html.twig
@@ -25,6 +25,7 @@
             float: left;
             font-size: 10px;
             word-wrap: break-word;
+            cursor: pointer;
         }
 
         .icon.black {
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Resources/views/Admin/Misc/iconList.html.twig
```diff
@@ -126,7 +129,15 @@
     {% endfor %}
 </table>
 
+{{ pimcore_minimize_scripts(['pimcore/common.js', 'pimcore/functions.js','pimcore/helpers.js']) | raw }}
```
brusch:
```suggestion
<script src="/bundles/pimcoreadmin/js/pimcore/common.js"></script>
<script src="/bundles/pimcoreadmin/js/pimcore/functions.js"></script>
<script src="/bundles/pimcoreadmin/js/pimcore/helpers.js"></script>
```
We want to get rid of `pimcore_minimize_scripts`, so let's just use the script directly - no real benefit in this case anyway ☺️

brusch (changes requested):

bundles/AdminBundle/Resources/views/Admin/Misc/iconList.html.twig
```diff
@@ -126,7 +129,17 @@
     {% endfor %}
 </table>
 
+<script src="/bundles/pimcoreadmin/js/pimcore/common.js"></script>
+<script src="/bundles/pimcoreadmin/js/pimcore/functions.js"></script>
+<script src="/bundles/pimcoreadmin/js/pimcore/helpers.js"></script>
+<script>
```
brusch:
```suggestion
<script {{ pimcore_csp.getNonceHtmlAttribute()|raw }}>
```
We need the nonce here in order to get the inline script to work 😉

brusch (changes requested):
Added `cursor: pointer` to the icons and a short info text.

brusch:
Thank you very much! 👍

NiklasBr:
I'm a few hours late, but there is also a `cursor: copy;` CSS keyword which is made for this :)

brusch:
@NiklasBr thanks for the hint, just tried that on Win 11 and it looks pretty weird 🙃 Don't know how this is on other systems.

NiklasBr:
Quick testing in Windows 10 with Firefox, and Edge, Sublime Text, PHP Storm, and Spotify show the same "+" type icon (some subtle differences) as it does when you ctrl-drag an item in File Explorer to copy it. main difference is that File Explorer also has a short descriptive "copy _____" text.

brusch:
I'm fine with it, just pushed it straight to `10.x` 9ecae6b4a0e53b3d9fd4603dc4ba707efbe3583e

PR #11813, opened 2022-03-31
[Bug]: In the Classification store view the X button deletes without warning

### LINKED ISSUE #11754: [Bug]: In the Classification store view the X button deletes without warning
### Expected behavior

We need a message "are you sure to delete the group with all values? After the next save the values from that group will be empty".

### Actual behavior

Deletes the whole data without a warning. (Missclick is a problem here)

### Steps to reproduce

Click on the X after you added a group and then save.

mcop1:
Resolves #11754

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/classificationstore.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/classificationstore.js
@@ -534,39 +534,42 @@ pimcore.object.tags.classificationstore = Class.create(pimcore.object.tags.abstr
     },
 
     deleteGroup: function(groupId) {
-        var currentLanguage;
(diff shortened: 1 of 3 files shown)


kingjia90 (changes requested):
the alert is fine, but the change is in the wrong file/datatype, it was supposed to be in the classificationstore (deleteGroup), not in the fieldcollection. 

IMHO The message is a little too long and "obvious", i would remove the part "After the next save all values from this group will be empty." 🤔

mcop1:
> the alert is fine, but the change is in the wrong file/datatype, it was supposed to be in the classificationstore (deleteGroup), not in the fieldcollection.
> 
> IMHO The message is a little too long and "obvious", i would remove the part "After the next save all values from this group will be empty." 🤔

You are right, I added the message for the deleteGroup method too and changed the label to something more generic. Regarding the text I think the message should communicate that the whole group gets deleted, not only a single value.

PR #11814, opened 2022-03-31
Fix: application log table drop fails with exception

thgross:
on certain database names, dropping the application log tables fails because the database name is unquoted.

For example, when the database name is `customer1-pimcore`, dropping the application log table fails.


mcop1:
Thank you!

Since it is a bug I cherry picked the commit to 10.3 too 
https://github.com/pimcore/pimcore/commit/f03d39ca581d7a7c80f4c15b955fd9ea57e0dad1

PR #11815, opened 2022-03-31
[Application logger] Support manual setting of "source" via context

BlackbitDevs:
Consider the following code to log something to application logger:
```php
$logger = $this->get(ApplicationLogger::class);
$logger->error('my error message', [
    'relatedObject' => $myObject, 
    'component'     => 'different component'
]);
```

Now in https://github.com/pimcore/pimcore/blob/f415841de9a3b09308a2be99580261a31e4fec40/lib/Log/ApplicationLogger.php#L179 it gets calculated from which method the `log()` method got called. 
There are 2 problems with this:
1. You cannot disable this. Sometimes it might simply not be necessary to log the `source` because it will always be the same (e.g. when you have a central Logger object which logs to the ApplicationLogger - then `source` will be the same value for all the logs)
2. You cannot customize it. In some cases you may want to set a different `source` - again with the example of the central `Logger` object in your application, you might want to use the method which called your central `Logger::log()` method. Or you may even want to input some other context in here.

This PR allows to set `source` similar to the other `context` fields like `relatedObject`, `component` etc.:
```php
$logger = $this->get(ApplicationLogger::class);
$logger->error('my error message', [
    'relatedObject' => $myObject, 
    'component'     => 'different component'
    'source' => 'My source'
]);
```

### DIFF
--- a/doc/Development_Documentation/18_Tools_and_Features/17_Application_Logger.md
+++ b/doc/Development_Documentation/18_Tools_and_Features/17_Application_Logger.md
@@ -230,7 +230,7 @@ class TestController
             'fileObject'    => $fileObject,
             'relatedObject' => $myObject, 
             'component'     => 'different component',
-            'source'        => 'Stack trace or context-relevant information' // if empty, gets automatically filled with class:method:line from where the log got executed
+            'source'        => 'Stack trace or context-relevant information' // optional, if empty, gets automatically filled with class:method:line from where the log got executed
         ]);
     }
 }
--- a/doc/Development_Documentation/18_Tools_and_Features/17_Application_Logger.md
+++ b/doc/Development_Documentation/18_Tools_and_Features/17_Application_Logger.md
@@ -229,7 +229,8 @@ class TestController
         $logger->error('my error message', [
             'fileObject'    => $fileObject,
             'relatedObject' => $myObject, 
-            'component'     => 'different component'
+            'component'     => 'different component',
+            'source'        => 'Stack trace or context-relevant information' // if empty, gets automatically filled with class:method:line from where the log got executed
         ]);
     }
 }
--- a/lib/Log/ApplicationLogger.php
+++ b/lib/Log/ApplicationLogger.php
@@ -176,7 +176,9 @@ class ApplicationLogger implements LoggerInterface
             $context['relatedObjectType'] = Service::getElementType($relatedObject);
         }
 
-        $context['source'] = $this->resolveLoggingSource();
+        if(!isset($context['source'])) {
+            $context['source'] = $this->resolveLoggingSource();
+        }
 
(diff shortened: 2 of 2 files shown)


aryaantony92:
@BlackbitNeueMedien Could you please explain the second problem (mentioned above) once more with a valid use case?

BlackbitDevs:
The problem is that currently in https://github.com/pimcore/pimcore/blob/f415841de9a3b09308a2be99580261a31e4fec40/lib/Log/ApplicationLogger.php#L195-L214 it is tried to find out which method triggered logging. When you have a central logger class like this:
```php
class CompositeLogger implements LoggerInterface {
  /** @var LoggerInterface[] */
  private $loggers;

  public function addLogger(LoggerInterface $logger) {
    $this->loggers[] = $logger;
  }

  public function log($level, $message, array $context = array()) {
    foreach ($this->loggers as $logger) {
      $logger->log($level, $message, $context);
    }
  }
}

$logger = new CompositeLogger;
$logger->addLogger(\Pimcore::getContainer()->get('logger'));
$logger->addLogger(ApplicationLogger::getInstance());
$logger->log('info', 'Test');
```

The source in the application_logs table will now always be the log `CompositeLogger::log()` method. If you use this class to log errors through your whole application / bundle, this information will be useless. It would be a good idea if we could override this default behaviour by providing a more useful `source` in `CompositeLogger::log()` like 
```php
public function log($level, $message, array $context = array()) {
  if(!isset($context['source'])) {
    $context['source'] = 'Test'; // custom logic here to determine source / cause of the error
  }
  foreach ($this->loggers as $logger) {
    $logger->log($level, $message, $context);
  }
}
```

brusch:
@BlackbitNeueMedien Looks good to me, just one more thing, we should add this to the docs 😊 (btw. the sooner the better, tomorrow is the release of `10.4.0` 😉) Thanks!

BlackbitDevs:
@brusch Have added this feature to the docs.

doc/Development_Documentation/18_Tools_and_Features/17_Application_Logger.md
```diff
@@ -229,7 +229,8 @@ class TestController
         $logger->error('my error message', [
             'fileObject'    => $fileObject,
             'relatedObject' => $myObject, 
-            'component'     => 'different component'
+            'component'     => 'different component',
+            'source'        => 'Stack trace or context-relevant information' // if empty, gets automatically filled with class:method:line from where the log got executed
```
brusch:
```suggestion
            'source'        => 'Stack trace or context-relevant information' // optional, if empty, gets automatically filled with class:method:line from where the log got executed
```

brusch:
Thank you! 👌

PR #11817, opened 2022-03-31
[Maintenance] Optimize tables with full text indexes periodically

BlackbitDevs:
Full text indexes get slower over time because of gaps caused by deleting entries and because word counts get fragmented. In a concrete case we had the problem of ~600K entries in `search_backend_data` and the query
```mysql
SELECT id FROM search_backend_data WHERE MATCH (`data`,`properties`) AGAINST ('123456' IN BOOLEAN MODE);
```
took 28 seconds.
This gets used in https://github.com/pimcore/pimcore/blob/f415841de9a3b09308a2be99580261a31e4fec40/bundles/AdminBundle/Helper/GridHelperService.php#L617
and also in other places like object search.

After running 
`OPTIMIZE TABLE search_backend_data`
this same query takes 0.1 seconds.
It is even documented in the MySQL docs: https://dev.mysql.com/doc/refman/8.0/en/fulltext-fine-tuning.html#fulltext-optimize

This PR adds a maintenance task which optimizes the tables with fulltext indexes once every 7 days - currently those are `search_backend_data` and `email_log`. I also thought about fetching all tables which have fulltext indexes and optimize those but this would pull control of third-party tables like from plugins or application-specific implementation away from the owners.

### DIFF
--- a/bundles/CoreBundle/Resources/config/maintenance.yaml
+++ b/bundles/CoreBundle/Resources/config/maintenance.yaml
@@ -107,3 +107,7 @@ services:
             - '@logger'
         tags:
             - { name: pimcore.maintenance.task, type: documents_static_page_generate }
+
+    Pimcore\Maintenance\Tasks\FullTextIndexOptimizeTask:
+        tags:
+            - { name: pimcore.maintenance.task, type: optimize_fulltext_indexes }
--- a/lib/Maintenance/Tasks/FullTextIndexOptimizeTask.php
+++ b/lib/Maintenance/Tasks/FullTextIndexOptimizeTask.php
@@ -1,0 +1,46 @@
+<?php
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Maintenance\Tasks;
+
+use Pimcore\Db;
+use Pimcore\Maintenance\TaskInterface;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Lock\LockFactory;
+
+/**
+ * @internal
+ */
+class FullTextIndexOptimizeTask implements TaskInterface
+{
+    /** @var \Symfony\Component\Lock\LockInterface */
+    private $lock;
+
+    public function __construct(LockFactory $lockFactory)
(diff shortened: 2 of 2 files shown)


brusch:
LGTM, but tasks need to be registered as services, see: 
https://github.com/pimcore/pimcore/blob/f611fabc7b4b58900a030a54409cb57afcde8edd/bundles/CoreBundle/Resources/config/maintenance.yaml#L59

Can you please update your PR accordingly? 
Thanks in advance!

BlackbitDevs:
Sorry, forgot the service. Is fixed now.

brusch:
👍 Thanks!

BlackbitDevs:
Just as a note: I tried to migrate this maintenance task to a Pimcore 6.9 system (with `PdoStore` as lock backend) and stumbled upon the MySQL error
```
General error 2014 Cannot execute queries while other unbuffered queries are active. Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.
```

There I changed the calls in https://github.com/pimcore/pimcore/blob/ffe18e9a90566cd0c5c9e488cbc6314730372329/lib/Maintenance/Tasks/FullTextIndexOptimizeTask.php#L41-L42
to use `fetchAll()` instead of `exec()` and this works. Do not know what caused the error but if this ever occurs also for Pimcore 10, this may become useful...

DrLuke:
> to use `fetchAll()` instead of `exec()` and this works. Do not know what caused the error but if this ever occurs also for Pimcore 10, this may become useful...

We just upgraded to Pimcore 10 and had this issue.

Can be solved by overriding the class:
`overrides/Pimcore/Maintenance/Tasks/FullTextIndexOptimizeTask.php`
```php
<?php

namespace Pimcore\Maintenance\Tasks;

use Pimcore\Db;
use Pimcore\Maintenance\TaskInterface;
use Symfony\Component\Lock\LockFactory;

/**
 * @internal
 */
class FullTextIndexOptimizeTask implements TaskInterface
{
    /** @var \Symfony\Component\Lock\LockInterface */
    private $lock;

    public function __construct(LockFactory $lockFactory)
    {
        $this->lock = $lockFactory->createLock(self::class, 86400 * 7, false);
    }

    /**
     * {@inheritdoc}
     */
    public function execute()
    {
        // Override to make it work
        // See: https://github.com/pimcore/pimcore/pull/11817#issuecomment-1105026544
        if ($this->lock->acquire(false)) {
            Db::get()->fetchAll('OPTIMIZE TABLE search_backend_data');
            Db::get()->fetchAll('OPTIMIZE TABLE email_log');
        }
    }
}
```

And then in your `composer.json`:

```json
"psr-4": {
  "App\\": "src/",
  "Pimcore\\Maintenance\\Tasks\\": "overrides/Pimcore/Maintenance/Tasks",
}
```

Issue #11609, opened 2022-03-07
[Bug]: Users.docTypes column is being truncated


DzemalBecirevic:
### Expected behavior

A roles document types are all stored and listed when reloading the admin role editor.

### Actual behavior

Not all document types of the role is selected .

### Steps to reproduce

Have more than 7 document types.

1. Add all 7 document types to a single role. 
2. Save the role
3. Close and open the role for editing in Pimcore admin.

DzemalBecirevic:
I believe that this issue comes from that the Users.docTypes columns is a varchart(255) column and that worked fine since before Pimcore X the docTypes where stored as a commaseparated list of IDs. In Pimcore X the docTypes are stored as a commaseparated list of GUID/Hashes and therefore more than 6.5 documents can not be stored and the field is truncated.

6.5 because the last line is truncated.

Post data is: 

   "docTypes":[
      "f83755ff-3ca7-4315-abdf-3b9d48638715",
      "56112d81-5018-4fc8-890a-57fdfbebf26f",
      "99802d6b-d93d-4234-91fd-ae6356d93975",
      "29cfb4bb-8204-4810-a742-843cb17b722c",
      "39a60490-c448-4263-9e81-8a4e2d38bfac",
      "fde0190c-e72a-45d0-81b9-0e331adc17af",
      "28d153ed-2544-4b5c-ba4e-aa54f6f7891d",
      "2c0c8f98-e90a-42a9-b557-6b3103bb5888",
      "40374f33-260a-4139-bf8e-9b41b4e34d87",
      "440c6910-efaf-47bf-b657-b8ca149e6989",
      "4c3cc5d8-d78a-4b0c-a0b8-2344449b7455",
      "e1791de7-46d4-4ac2-8c23-70bdb26ebede",
      "e221b7cc-88f5-4759-a5f0-4563474254d5"
   ],
   
   
   
Data in database is:

f83755ff-3ca7-4315-abdf-3b9d48638715,
56112d81-5018-4fc8-890a-57fdfbebf26f,
99802d6b-d93d-4234-91fd-ae6356d93975,
29cfb4bb-8204-4810-a742-843cb17b722c,
39a60490-c448-4263-9e81-8a4e2d38bfac,
fde0190c-e72a-45d0-81b9-0e331adc17af,
28d153ed-2544-4b5c-ba4e-aa54f6f78


It obvious that the column should change datatype but I'm not sure to what.

blankse:
[...]

PR #11824, opened 2022-04-02
Fix: Set doc type permissions to users doesn't work

blankse:
Fix https://github.com/pimcore/pimcore/issues/11609


brusch:
Works like a charm - thanks @blankse !

Issue #11549, opened 2022-02-28
[Bug]: getById issue in Pimcore\Model\DataObject\Classificationstore\KeyGroupRelation\Dao 


FrederikHulleman:
### Expected behavior

The method getById in Pimcore\Model\DataObject\Classificationstore\KeyGroupRelation\Dao should return the correct record from both the classificationstore_relations & classificationstore_keys tables.

### Actual behavior

The query below returns the correct record from classificationstore_relations, but it will use the first record from classificationstore_keys

`$data = $this->db->fetchRow('SELECT * FROM ' . self::TABLE_NAME_RELATIONS
            . ',' . Model\DataObject\Classificationstore\KeyConfig\Dao::TABLE_NAME_KEYS . ' WHERE keyId = ? AND groupId = ?', [$this->model->getKeyId(), $this->model->getGroupId()]);`

In the were statement the following part is missing: `classificationstore_relations.keyId = classificationstore_keys.id`

### Steps to reproduce

I encountered this issue when I was working with the Pimcore Data Importer bundle. 

1. Import Settings (of a Data object importer) > Import settings > Mappings > use a Classification store key in the 'Data target'. 
2. Save the import config 
3. Close this import config
4. Reopen this import config
5. Lookup the data target from step 1. Here you'll see the wrong group-key relation

blankse:
@FrederikHulleman I created a PR https://github.com/pimcore/pimcore/pull/11825
Can you test it?

brusch:
Fixed by #11825

PR #11825, opened 2022-04-02
Fix KeyGroupRelation/CollectionGroupRelation: getById returns wrong result

blankse:
Fix https://github.com/pimcore/pimcore/issues/11549


brusch:
Thank you very much! 👍

PR #11827, opened 2022-04-03
Fix new PhpStan Error

blankse:
With the new PhpStan Version 1.5.4 one new error appears
https://github.com/phpstan/phpstan/releases/tag/1.5.4


fashxp:
thx

PR #11826, opened 2022-04-02
Fix PhpStan level 5 errors

blankse:
Before:
`[OK] Baseline generated with 414 errors.`
After:
`[OK] Baseline generated with 62 errors.`

- [x] Rebase/Merge if splitted PRs are merged and in 10.x branch
- [x] Create more splitted PRs

PR #11829, opened 2022-04-03
Fix Asset\Service::copyContents()

blankse:
Fix the Asset\Service::copyContents() method. Element\Service::cloneMe() accept only objects with ElementInterface as parameter. The other copyContents() (Document/Object) also set the properties direct.
Splitting up from #11826

### DIFF
--- a/models/Asset/Service.php
+++ b/models/Asset/Service.php
@@ -161,13 +161,12 @@ class Service extends Model\Element\Service
      * @param Asset $target
      * @param Asset $source
      *
-     * @return mixed
+     * @return Asset
      *
      * @throws \Exception
      */
     public function copyContents($target, $source)
     {
-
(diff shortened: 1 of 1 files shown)


dvesh3:
@blankse thanks!

brusch:
Could be dangerous, as it keeps the references, right? 
Should we do at least a deep copy? Or maybe `array_map(... clone ...)`.

blankse:
@brusch Yes this was also in my thoughts. But this must be fixed/changed in all 3 element types.

dvesh3:
@brusch @blankse please see https://github.com/pimcore/pimcore/pull/11860

dvesh3:
@brusch @blankse please see https://github.com/pimcore/pimcore/pull/11860

PR #11835, opened 2022-04-04
reapeating Headers of Custom-Report export

elements-team:
The header of a custom-report when exporting to CSV was repeating every 5000 rows. Changed this to only print the header once at the beginning.


dvesh3:
@elements-team thanks for your PR. please rebase it on 10.3 branch as it is a fix and also check stan is failing.

jdreesen:
For PHPStan you just need to rebase on top of `10.3`, because it was already fixed in #11827

brusch:
🍒-picked into `10.3` ( d79f617a6a1f204ae104aae6fd1a82dacf0c0fde ) 
Thanks @elements-team

PR #11842, opened 2022-04-04
Fix PHP-FPM status/ping routes in nginx config examples

jdreesen:
The PHP-FPM status/ping routes need the `SCRIPT_FILENAME` FastCGI param to be set for PHP-FPM to recognize them.

This can either be done by including the `fastcgi.conf` (which is not available on all systems) instead of `fastcgi_params` or, if the latter is used, manually like in this PR.


brusch:
Thank you very much!

PR #11843, opened 2022-04-05
Fix catching fatal errors

jantomicky:
Fixes uncaught "call to a member function on null" errors

  
  

Resolves #

### DIFF
--- a/lib/Maintenance/Tasks/VersionsCleanupStackTraceDbTask.php
+++ b/lib/Maintenance/Tasks/VersionsCleanupStackTraceDbTask.php
@@ -60,7 +60,7 @@ class VersionsCleanupStackTraceDbTask implements TaskInterface
                     $version->setGenerateStackTrace(false);
                     $version->setStackTrace(null);
                     $version->getDao()->save();
-                } catch (\Exception $e) {
+                } catch (\Throwable $e) {
                     $this->logger->debug('Unable to cleanup stack trace for version ' . $version->getId() . ', reason: ' . $e->getMessage());
                 }
             }


brusch:
@jantomicky Thanks for your PR. However, I think we should better fix the root cause (null check, ...) and not just catching all errors. Can you please update your PR accordingly? Thanks in advance!

jantomicky:
Of course, I'll try! Thanks for the feedback. Just out of curiosity:
1. How is it possible for `null` to end up i the list in the first place, should I look into that?
2. What errors do you, as PImcore developers, consider worth catching or not? Because in this case, the only difference is where (Application Logger or simple file log) the error gets logged.
3. By null check, do you mean something like this?
```
if (!$version) {
    continue;
}
```

To provide more context, I encountered this issue only when I linked Monolog to our Slack channel, it then got sent as an emergency error to our monitoring channel.

brusch:
@jantomicky Since I don't have any details or a stack trace it's hard for me to follow you 😊 But if the error is thrown in this line: 
https://github.com/pimcore/pimcore/blob/8c6ececed91c321045b786672670c5342e0dd2c7/lib/Maintenance/Tasks/VersionsCleanupStackTraceDbTask.php#L60 
I agree, that it is strange that a null value is in the list ... this needs probably fixes somewhere here: 
https://github.com/pimcore/pimcore/blob/10.x/models/Version/Listing/Dao.php#L52

But then it's still strange that an ID is part of the result but you can't get the detail of this version. Maybe you can further investigate this. Thanks!

jantomicky:
I'll look into it, thanks!

brusch:
@jantomicky 🏓just a quick reminder 😊

brusch:

Issue #6381, opened 2020-05-13
[Classification Store] Quantity Value: Default units not preselected after saving the object


aarongerig:
# Bug Report

This issue is a follow up to https://github.com/pimcore/pimcore/issues/6139
If you add a classification store which holds Quantity Values types with a default unit to an object, but leave the actual value of a Quantity Value empty and save the object, the unit is not "preselected" after reloading the object anymore.

### Expected behavior
Default unit is always preselected if the value of the field is empty or no other unit was previously selected.

### Actual behavior
The unit field is empty after saving and reloading the object.

### Steps to reproduce

- Log in to https://demo.pimcore.fun/admin
- Create a dummy classification store with a key type of Quantity Value
- Create a dummy class which includes the previously created Classification Store
- Create a dummy object, add the classification store but leave it empty and save it
- Reload the object

Newly created object with blank quantity value:
![first_open](https://user-images.githubusercontent.com/17384333/81799547-c62c5c00-9511-11ea-9a17-02e540791316.png)

Object after reload:
![after_save](https://user-images.githubusercontent.com/17384333/81799623-e2c89400-9511-11ea-9298-35e7bd302277.png)

BlackbitNeueMedien:
I currently experiencing the same problem also with classification store but for normal quantity value fields inside a class. When you set a default unit this does not get preselected. I kind of understand that because this would lead to setting a unit for a quantity value field which was `null` before (and afterwards would have value `null` and default unit set).

I could imagine that it would be better to apply the default unit when saving in case no unit has been provided. WDYT?

hellwy:
I think this should be solved somehow. Even by taking into account what @BlackbitNeueMedien is saying (make sense), i think PIM should be MAX optimized for saving operators clicks.

hellwy:
probably, as idea, the default unit could be set in moment of writing data into the fields.

ctippler:
Hi,

i currently encountered the same problem and it turned out that it works with the inputQuantityValue datatype but not with the quantityValue data type. 

This is because the inputQuantityValue  sets the default values in the initialize() method and the quantityValue doesn't do that.

I fixed this with 
![image](https://user-images.githubusercontent.com/8791970/156441237-ab2b419a-e54b-4578-9e4d-20d1e7030e4f.png)

I guess it is because of this:
https://github.com/pimcore/pimcore/pull/6191/commits/3e1b626e279df23cbed7c3caf09c2f1464bffdfc 

`
if (isNew) {
[...]

PR #11845, opened 2022-04-05
Quantity value bugfix

ctippler:
Set the default value for quantityValue type (as the inputQuantityValue already does it correctly)
This should fix the issue described at https://github.com/pimcore/pimcore/issues/6381


NiklasBr:
Looks good to me 👍

brusch:
🍒-picked into `10.3` as it is a bug fix 😊 (see 882ffbe30511640832f2be2bf7f0750daf645abd ) 
Thanks a lot for your PR!

PR #11847, opened 2022-04-05
Do not run Codeception and static analysis on Markdown files

NiklasBr:
Resolves #

## Additional info  

IMO it should not be necessarily to waste computing resources when e.g. docs and other .md files are updated.

### DIFF
--- a/.github/workflows/codeception.yml
+++ b/.github/workflows/codeception.yml
@@ -6,11 +6,15 @@ on:
         branches:
             - "[0-9]+.[0-9]+"
             - "[0-9]+.x"
+        paths:
+            - '!*.md'
     push:
         branches:
             - "[0-9]+.[0-9]+"
             - "[0-9]+.x"
             - "*_actions"
+        paths:
+            - '!*.md'
 
 env:
     PIMCORE_PROJECT_ROOT: ${{ github.workspace }}
--- a/.github/workflows/codeception.yml
+++ b/.github/workflows/codeception.yml
@@ -6,17 +6,17 @@ on:
         branches:
             - "[0-9]+.[0-9]+"
             - "[0-9]+.x"
-        paths:
-            - '!docs/**'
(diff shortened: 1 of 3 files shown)


brusch:
Good idea - but probably also relevant for other file types, right? What about CSS, JS, ...?

NiklasBr:
@brusch what if someone moves/renames those files, could have side effects? E.g. [PimcoreEcommerceFrameworkBundle.php](https://github.com/pimcore/pimcore/blob/10.x/bundles/EcommerceFrameworkBundle/PimcoreEcommerceFrameworkBundle.php) and that would be useful to know, right?

brusch:
@NiklasBr Don't think this is getting checked by Codeception at the moment ... so it should be safe to ignore. I'm also wondering whether it would be better to just exclude pathes, like `/docs`, `bundles/AdminBundle/Resources/public`, ... WDYT?

NiklasBr:
Let's try that @brusch

dvesh3:
@NiklasBr what happens with this change when a PR changes both code & docs? tests should run in that case.

NiklasBr:
The docs says it should work:

> When all the path names match patterns in paths-ignore, the workflow will not run. If any path names do not match patterns in paths-ignore, even if some path names match the patterns, the workflow will run.

dvesh3:
> The docs says it should work:
> 
> > When all the path names match patterns in paths-ignore, the workflow will not run. If any path names do not match patterns in paths-ignore, even if some path names match the patterns, the workflow will run.

@NiklasBr 👍 I would also change Documentation Action to run only if there's a change in `docs/` path.

NiklasBr:
@dvesh3 did you see my change?

dvesh3:
> @dvesh3 did you see my change?

LGTM. thanks!

dvesh3 (approved):

dvesh3:
@NiklasBr thanks!

Issue #11529, opened 2022-02-24
[DISCUSSION]: ReverseObjectRelation should not support inheritance


brusch:
### Expected behavior

Actually I don't think it makes sense that this data-type supports inheritance - for what reason, any use-cases? 

### Actual behavior

Actually the `ReverseObjectRelation` does support inheritance.

### Steps to reproduce

...

fashxp:
possible use case: 
- products have relation to category, category has corresponding reverse object relation to its products. 
- products have inheritance activated, variants are implemented via child objects and inherit category relation from their master product
- on category reverse object relation, you might wanna see also variants that are assigned to that category

brusch:
The described use-cases are about the inheritance on the originating objects, I was thinking about the actual reverse relation field, so e.g. if the reverse relation field on category object can't find reverse relations, should it try to get reverse relations from the parent category....

All of the 3 described use-cases are anyhow not possible at the moment, as the inherited data is not gettings persited in the corresponding `object_relation_*` tables, thus it cannot be queried.

brusch:
Added 
https://github.com/pimcore/pimcore/blob/b26d41f4f7dfd7a6b606b30495bab56ffc305d7c/models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php#L248
in #10242 as it's currently anyway not working 😉

Closing this for now.

PR #11848, opened 2022-04-05
[Bug]: ReverseObjectRelation data is inherited

### LINKED ISSUE #11818: [Bug]: ReverseObjectRelation data is inherited
### Expected behavior

ReverseObjectRelation data is inherited should not be inherited as it was discussed here:
https://github.com/pimcore/pimcore/issues/11529

### Actual behavior

Data is inherited. As a starting point see these lines of code
https://github.com/pimcore/pimcore/blob/f415841de9a3b09308a2be99580261a31e4fec40/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L619-L626
...

### Steps to reproduce

As described in https://github.com/pimcore/data-hub/issues/498

mcop1:
Resolves #11818

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -618,7 +618,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
 
             $relations = $object->getRelationData($refKey, !$fielddefinition instanceof ReverseObjectRelation, $refId);
 
-            if (empty($relations) && !empty($parent)) {
+            if (empty($relations) && !empty($parent) && $fielddefinition->supportsInheritance()) {
                 $this->getDataForField($parent, $key, $fielddefinition, $objectFromVersion, $level + 1);
             } else {
                 $data = [];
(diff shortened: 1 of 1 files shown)


fashxp:
please rebase to 10.3

mcop1:
> please rebase to 10.3

Needs to be implemented in 10.x (10.4), because the method is not available in 10.3, see
https://github.com/pimcore/pimcore/issues/11529#issuecomment-1056779990

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -618,7 +618,7 @@ private function getDataForField($object, $key, $fielddefinition, $objectFromVer
 
             $relations = $object->getRelationData($refKey, !$fielddefinition instanceof ReverseObjectRelation, $refId);
 
-            if (empty($relations) && !empty($parent)) {
+            if (empty($relations) && !empty($parent) && $fielddefinition->supportsInheritance()) {
```
dvesh3: I would check if field supports inheritance before checking relations & parent
```suggestion
            if ($fielddefinition->supportsInheritance() && empty($relations) && !empty($parent)) {
```

dvesh3 (approved):

PR #11849, opened 2022-04-05
Open variant on double click in 'variants'tab

mhgtlc:
Open variant on double click
Screenshot 2022-04-05 at 16 50 42

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/variantsTab.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/variantsTab.js
@@ -211,7 +211,7 @@ pimcore.object.variantsTab = Class.create(pimcore.object.helpers.gridTabAbstract
             listeners: {
                 rowdblclick: function (grid, record, tr, rowIndex, e, eOpts) {
                     var subtype = record.get('subtype');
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/variantsTab.js
```diff
@@ -210,7 +210,11 @@ pimcore.object.variantsTab = Class.create(pimcore.object.helpers.gridTabAbstract
             tbar: tbar,
             listeners: {
                 rowdblclick: function (grid, record, tr, rowIndex, e, eOpts) {
-
+                    var subtype = record.get('subtype');
+                    if (record.get('type') === "object" && record.get('subtype') !== "folder" && record.get('subtype') !== null) {
```
brusch:
```suggestion
                    if (record.get('type') === "object" && subtype !== "folder" && subtype !== null) {
```

brusch (changes requested):

brusch:
Changed base to `10.x` as I'd see this as an improvement 😊

brusch:
Makes sense - thanks a lot! 👍

PR #11851, opened 2022-04-05
Properties: Show Predefined Name in Grid

blankse:
You select the properties by the predefined name but than you can't see it in the grid. Only the key and the description is there. I think the user should also see the name that he has selected.


bundles/AdminBundle/Resources/public/js/pimcore/element/properties.js
```diff
@@ -270,6 +297,14 @@ pimcore.element.properties = Class.create({
                         renderer: Ext.util.Format.htmlEncode,
                         width: 230
                     },
+                    {
+                        text: t('name'),
+                        dataIndex: 'predefinedName',
```
dvesh3, line 302: Should we display "Custom" here for custom properties?

dvesh3: 🏓 @blankse

blankse: @dvesh3 Good idea, I added it.

dvesh3:
@blankse thanks!

PR #11855, opened 2022-04-06
[Maintenance] Fix VersionsCleanupStackTraceDbTask (performance + logic)

BlackbitDevs:
Current logic in https://github.com/pimcore/pimcore/blob/9ecae6b4a0e53b3d9fd4603dc4ba707efbe3583e/lib/Maintenance/Tasks/VersionsCleanupStackTraceDbTask.php#L55 is wrong because:
1. Imagine you have 600 matching versions
2. Initially you select with `LIMIT 0,500`
3. The `stackTrace` for those 500 found versions gets set to `NULL`
4. When you select with `LIMIT 500,500` in the next iteration, you will not find anything because after 3. there are only 100 versions left which match the condition `stackTrace IS NOT NULL`

But the even more severe problem is performance: We just had these processes:
Bildschirmfoto 2022-04-06 um 11 40 35
(Don't know if maintenance job locking did not work but was running with Pimcore 10.0, so this does not matter for now)

The problem is the `filesort` which has to be done because the database uses the index on `autoSave` (this condition gets added in https://github.com/pimcore/pimcore/blob/9ecae6b4a0e53b3d9fd4603dc4ba707efbe3583e/models/Version/Listing/Dao.php#L30-L36)
We could create an optimized index but ordering in this case is not necessary anyway as we want to change stack trace for all found versions.

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20220419120333.php
+++ b/bundles/CoreBundle/Migrations/Version20220419120333.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+use Pimcore\Model\DataObject\ClassDefinition\Listing;
+
+final class Version20220419120333 extends AbstractMigration
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function getDescription(): string
+    {
+        return 'Add index to versions.stackTrace to accelerate maintenance task "VersionsCleanupStackTraceDbTask"';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $versionsTable = $schema->getTable('versions');
+
+        if (!$versionsTable->hasIndex('stackTrace')) {
+            $versionsTable->addIndex(['stackTrace'], 'stackTrace', [], ['lengths' => [1]]);
+        }
+    }
+
+    public function down(Schema $schema): void
+    {
+        $versionsTable = $schema->getTable('versions');
+
+        if (!$versionsTable->hasIndex('stackTrace')) {
+            $versionsTable->dropIndex('stackTrace');
+        }
+    }
+}
--- a/bundles/InstallBundle/Resources/install.sql
+++ b/bundles/InstallBundle/Resources/install.sql
@@ -339,7 +339,7 @@ CREATE TABLE `notes` (
   PRIMARY KEY (`id`),
   KEY `cid_ctype` (`cid`, `ctype`),
   KEY `date` (`date`),
-  KEY `user` (`user`),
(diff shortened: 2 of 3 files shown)


brusch:
@BlackbitNeueMedien Thanks a lot for your PR, however I think we could even further optimize the process: 
1. we can fetch the versions in a while loop and only set the limit at the beginnen (then we also don't need the `getTotalCount()` and so on ... 
2. we can implement a dedicated method in the `Dao` for clearing the stack trace, to make it more efficient (e.g. `clearStackTrace()`. 

WDYT? 😊

BlackbitDevs:
@brusch I adjusted it to use a do-while-loop.

I did not add `Version\Dao::clearStackTrace()` as I do not understand how this should make clearing the stack trace more efficient. Currently the database data is always set with the setter methods of the `AbstractModel` objects, not directly with the `Dao` objects. Or have I understood you wrong?

BlackbitDevs:
This continues to kill the database in our client's case:
Bildschirmfoto 2022-04-19 um 10 07 20
(This was before the do-while-loop but this will not make those queries any better)

~~The problem is that `stackTrace IS NOT NULL` condition cannot be indexed~~ and `autoSave=0` and `date < NOW() - 7 days` match a lot of versions so that the indexes are not efficient.
~~What do you think of 
`Db::get()->executeUpdate('UPDATE versions SET stackTrace=NULL WHERE date < ?', [time() - 86400 * 7]);`
?~~
Direct update of course has same performance problem...

The only alternative I see is a "computed column" as described in https://stackoverflow.com/a/12048065

BlackbitDevs:
I added an index to `versions.stackTrace` with length 1 (more is not necessary to check if it is `NULL` or not). Now the `stacktrace IS NOT NULL` condition is very fast (0,0001s vs. 120s before).

bundles/CoreBundle/Migrations/Version20220419120333.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        return 'Add index to versions.stackTrace to accelerate maintenance task "VersionsCleanupStackTraceDbTask"';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $versionsTable = $schema->getTable('versions');
+
+        if (!$versionsTable->hasIndex('stackTrace')) {
+            $versionsTable->addIndex(['stackTrace'], 'stackTrace', [], ['lengths' => [1]]);
```
brusch, line 39: We need to add the index in the `install.sql` as well, see here: 
https://github.com/pimcore/pimcore/blob/265d95f285689b5df5c4f5b22c5e72c281bffb24/bundles/InstallBundle/Resources/install.sql#L757

brusch (changes requested):
Regarding setting the stack trace to null:
What I meant is that we're currently saving the whole version again by calling `$version->getDao()->save();`. I'd rather implement a dedicated method, like `$version->getDao()->clearStrackTrace();` that just uses: 
```php
        $this->db->update('versions', ['stackTrace' => null], ['id' => $this->model->getId()]);
```
to set the stack trace to `null`, which will be much more efficient, right?

bundles/CoreBundle/Migrations/Version20220419120333.php
BlackbitDevs: Done

BlackbitDevs:
@brusch Ok, I thought it had a reason that we save the version again. But if not, then we can completely simplify this task to 
```php
Db::get()->executeUpdate('UPDATE versions SET stackTrace = NULL WHERE date < ? AND stackTrace IS NOT NULL', [time() - 86400 * 7]);
```
WDYT?
(I already adjusted this PR)

brusch:
@BlackbitNeueMedien LGTM, can you just quickly fix the issue reported by PhpStan? Thanks again!

BlackbitDevs:
Done. This PR is really a simplification ;-)

brusch:
Love it 💕 Thanks a lot!

Issue #9727, opened 2021-07-12
[Bug]: Performance Data Object Grid


jremmurd:
### Expected behavior

 DB Queries should not take 10+ seconds inside Pimcore. Pimcore optimizes queries inside admin interface to the fullest.

### Actual behavior

Opening any data object grid in Pimcore executes 2 main queries:

```
SELECT COUNT(*) AS `totalCount`
FROM `object_customer`
WHERE ((o_path = '/XXX/XXX/XXX' OR o_path LIKE '/XXX/XXX/XXX/%') AND object_customer.o_type IN ('object', 'folder'))
```

and

```
SELECT object_customer.o_id as o_id, `object_customer`.`o_type`
FROM `object_customer`
WHERE ((o_path = '/XXX/XXX/XXX' OR o_path LIKE '/XXX/XXX/XXX/%') AND object_customer.o_type IN ('object', 'folder'))
ORDER BY `o_id` ASC
LIMIT 25
```

which both may take several seconds, or even result in a visible error to the user if a request exceeds the js-defined 30 seconds limit.

Explain statement for 700+ rows:
![image](https://user-images.githubusercontent.com/9052094/125296314-74a27b80-e326-11eb-98c3-7bbf38657dc8.png)

### Steps to reproduce

Open data object grid with several hundred thousand children.

andreas-gruenwald:
This query is triggered in the children grid of the Pimcore object tree.

markus-moser:
Now as Pimcore dropped the MySql < 8 Support it is possible to use CTE https://dev.mysql.com/doc/refman/8.0/en/with.html. In theory it should be possible to replace the like with hierarchical queries which should be much faster.

markus-moser:
This is a big problem in projects with massive amounts if objects. I tried to use CTE but in my first tests the performance was not better then the like. This is a problem which is very hard to be solved (without something like elasticsearch in the background...) but it's a major topic when Pimcore wants to be ready for PIM systems with hundrets of thousands or even millions of products. At the moment its not really practicle. #6383 is even much a bigger topic but also this issue really limits Pimcore here.

markus-moser:
Just as a reference the CTE I tried:

```sql
WITH recursive cte (o_id, o_classid, o_parentid) AS
(
       SELECT o_id,
              o_classid,
              o_parentid
       FROM   objects
       WHERE  o_parentid = 539285
       UNION ALL
       SELECT     p.o_id,
                  p.o_classid,
[...]

PR #11856, opened 2022-04-06
[Grid] Enhance performance to get data object classes for opened folder

BlackbitDevs:
Currently we execute https://github.com/pimcore/pimcore/blob/9ecae6b4a0e53b3d9fd4603dc4ba707efbe3583e/models/DataObject/AbstractObject/Dao.php#L452 to get the classes for the objects in the grid. This `GROUP BY` is very slow when you have millons of objects.

This PR uses a different approach. It iteratively searches for an object of a class which it does not already know. This way no grouping is necessary.

In our client project I could reduce query times from 26s to 1s with this approach (in this case the folder had only objects of one class).

### DIFF
--- a/models/DataObject/AbstractObject/Dao.php
+++ b/models/DataObject/AbstractObject/Dao.php
@@ -449,7 +449,16 @@ class Dao extends Model\Element\Dao
             if (!$this->model->getId() || $this->model->getId() == 1) {
                 $path = '';
             }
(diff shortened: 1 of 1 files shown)


brusch:
This is related to #9727 and partly resolves it.

brusch:
Well done 🧑‍🚀👍 Thanks!

BlackbitNeueMedien:
> This is related to https://github.com/pimcore/pimcore/pull/11856 and partly resolves it.

@brusch Think you copied the wrong number as this PR is #11856

Issue #10924, opened 2021-11-30
Custom Layouts for object bricks + field collections


BlackbitDevs:
### Feature description

With custom layouts you can limit the field which a certain user / role can see / edit. But this currently only exists for classes, not for object bricks or field collections. It would be nice if we could extend this functionality also to those.

matthiashamacher:
Seems to be the same as https://github.com/pimcore/pimcore/issues/1002

dvesh3:
> Seems to be the same as #1002

thanks! I would keep this one open as it is clearly described and voted 😜

plixus:
Hi, I came across this when I was looking for a solution to give a field collection a custom layout.

@BlackbitDevs as mentioned in #12084, are you planning to implement this feature for field collections as well?

BlackbitDevs:
@plixus I would love to but this needs some time and currently we have no client who would pay for this. If you are interested, please send an email to info@blackbit.de then we can discuss how to approach.

PR #11857, opened 2022-04-06
[Listing] Question mark and colon in condition in addConditionParam() are always interpreted as prepated statement parameters

BlackbitDevs:
Step to reproduce:
```php
$listing = new Listing();
$listing->addConditionParam('name="Brand: Test product"');
$listing->load();
```
This will result in an SQL error / PDO exception:
`SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens`

The reason is that in https://github.com/pimcore/pimcore/blob/da43b3f5fb18b59bdadfa49a77b43eb69d7ebf5e/lib/Model/Listing/AbstractListing.php#L258-L266 it is always assumed that when there is a `:` or a `?` in the SQL condition that this is a query parameter.
So the same problem appears when
```php
$listing->addConditionParam('name="This is an object name?"');
``` 

Above are simplified examples, of course normally you would use 
```php
$listing->addConditionParam('name=?', ['Brand: Test product']);
```
but sometimes this is not easily possible.

An alternative solution would be to add `Listing::addCondition()` to add conditions without parameters.
Imho, it is not a BC break, so if you want I could also rebase to `10.3` branch.


jdreesen:
This contains a commit from PR #11817

lib/Model/Listing/AbstractListing.php
```diff
@@ -256,7 +256,7 @@ public function addConditionParam($key, $value = null, $concatenator = 'AND')
 
         $key = '('.$key.')';
         $ignore = true;
-        if (strpos($key, '?') !== false || strpos($key, ':') !== false) {
+        if ((strpos($key, '?') !== false || strpos($key, ':') !== false) && func_num_args() > 1) {
```
jdreesen:
```suggestion
        if (func_num_args() > 1 && (str_contains($key, '?') || str_contains($key, ':'))) {
```

BlackbitDevs: Thought again about `func_num_args()` this is not a good idea because then you could not do an OR condition likt this:
```php
$listing->addConditionParam('name is null', null, 'OR');
```

Now I look for a way to check if the `?` or `:` is within quotes - if it is, then it is not a prepared statement parameter.

aryaantony92:
@BlackbitNeueMedien Since this is a bug, could you please rebase this to 10.3 branch ?

BlackbitDevs:
@aryaantony92 Is now for 10.3 branch.

aryaantony92:
@BlackbitNeueMedien Thanks for rebasing. 

But even after applying the patch/commit, when I try to do

$listing->addConditionParam('name="Brand: Test product"'); Or
$listing->addConditionParam('name="This is an object name?"');

I can still see the error message 

"SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens"

BlackbitDevs:
Sorry, regex was not correct, now it should work, even for
```php
$listing->addConditionParam('name="Brand: " and o_path=? and o_key="test?"', ['/']);
```

brusch:
I'd say this is quite a critical thing that should be covered by a test, could you please add one for all the possible cases? Thanks in advance!

brusch:
@BlackbitNeueMedien btw. writing tests is now super easy, see: #11872 and https://pimcore.com/docs/pimcore/current/Development_Documentation/Development_Tools_and_Details/Testing/Core_Tests.html 

Thanks in advance!

brusch:
@BlackbitNeueMedien 🏓 one more try 😉

BlackbitDevs:
Yep, did not forget it. Currently implementing #10924, try to come back to this PR here for the tests (or perhaps you would be so kind and create this?)

brusch:
@BlackbitNeueMedien ah cool 👍 then we wait until you're finished with the other topic 😉 Thanks in advance!

BlackbitDevs:
@brusch I added the tests.

brusch:
Thank you very much!

PR #11858, opened 2022-04-07
Reset fallback document after assigning it

xlttj:
A problem occurs in some special cases, i.e. when a cli script or some long running processes perform subsequent requests like so:

```
$kernel   = Kernel::get();
$request  = Request::create($somePath, 'GET');
$response = $kernel->handle($request);
```

The problem is that in the above use case the document once assigned in line 119 will be used in subsequent calls. That will overwrite an already correctly set document in the DocumentResolver in line 159.

When the fallback document is set to `null` afterwards, that won't happen.

  
  

Resolves #


brusch:
LGTM and shouldn't have any side-effects. Thanks!

xlttj:
@brusch You added this to the 10.3.5 milestone, but it was merged into 10.x, so this is not in 10.3.5. My fault, I guess…

brusch:
@xlttj Thanks for the hint - 🍒-picked it into `10.3` so it will be part of 10.3.6 then :)

PR #11859, opened 2022-04-07
Fix xliff import

valentin-:
## Additional info  

1. Fix special chars and remove empty new lines

Example navigation title:

image

2. Prevent PHP 8 errors

### DIFF
--- a/lib/Translation/ImportDataExtractor/Xliff12DataExtractor.php
+++ b/lib/Translation/ImportDataExtractor/Xliff12DataExtractor.php
@@ -83,6 +83,10 @@ class Xliff12DataExtractor implements ImportDataExtractorInterface
             $content = $transUnit->target->asXml();
             $content = $this->xliffEscaper->unescapeXliff($content);
 
+            if (is_string($content)) {
+                $content = html_entity_decode(trim($content));
+            }
+
             $attributeSet->addAttribute($type, $name, $content);
         }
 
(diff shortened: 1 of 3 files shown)


lib/Translation/ImportDataExtractor/Xliff12DataExtractor.php
```diff
@@ -83,6 +83,10 @@ public function extractElement(string $importId, int $stepId): ?AttributeSet
             $content = $transUnit->target->asXml();
             $content = $this->xliffEscaper->unescapeXliff($content);
 
+            if (is_string($content)) {
+                $content = html_entity_decode(trim($content));
```
brusch, line 87: I think here would be a better place for doing this: 
https://github.com/pimcore/pimcore/blob/514ffeb569f6305ce2e18009014ffeb8432c0bbc/lib/Translation/Escaper/Xliff12Escaper.php#L101
Actually it is already decoded, but only when wrapped in <bpt> tags. 
Maybe we can just add an else clause there, where we do the decode in case it's just a simple string and no markup. WDYT?

brusch (changes requested):

lib/Translation/ImportDataExtractor/Xliff12DataExtractor.php
brusch: @valentin- 🏓😉

valentin-: @brusch sorry I'm currently on vacation and can do it in 1,5-2 weeks at the earliest.

brusch: @valentin- 🏓 just a quick reminder 😊 Thanks!

brusch:
Closed in favor of #12448
Thanks @valentin- for your efforts 🙋

lib/Translation/ImportDataExtractor/Xliff12DataExtractor.php
valentin-: @brusch sorry I was quite busy and then oof again

PR #11860, opened 2022-04-07
[Elements] Clone properties on copyContents

dvesh3:
follow up to #11829

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -1265,10 +1265,10 @@ class Service extends Model\AbstractModel
 
     /**
      * @template T
-     * @param T $properties
-     * @return T
      *
-     * @return mixed
+     * @param T $properties
+     *
+     * @return T
      */
     public static function cloneProperties(mixed $properties): mixed
     {
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -1264,7 +1264,9 @@ class Service extends Model\AbstractModel
     }
 
     /**
-     * @param mixed $properties
+     * @template T
+     * @param T $properties
+     * @return T
      *
      * @return mixed
      */
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -1264,6 +1264,21 @@ class Service extends Model\AbstractModel
     }
 
     /**
+     * @param mixed $properties
+     *
+     * @return mixed
+     */
+    public static function cloneProperties(mixed $properties): mixed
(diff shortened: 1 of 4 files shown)


blankse (commented):
Maybe the PHPDocs, Typehint should be nullable.

models/Element/Service.php
```diff
@@ -1263,6 +1263,21 @@ function (Concrete $object, Data $fieldDefinition, $property, $currentValue) {
         return $theCopy;
     }
 
+    /**
+     * @param mixed $properties
```
blankse: Maybe nullable?
```suggestion
     * @param Model\Property[] $properties
```

models/Element/Service.php
```diff
@@ -1263,6 +1263,21 @@ function (Concrete $object, Data $fieldDefinition, $property, $currentValue) {
...
+     *
+     * @return mixed
```
blankse: Maybe nullable?
```suggestion
     * @return Model\Property[]
```

models/Element/Service.php
```diff
@@ -1263,6 +1263,21 @@ function (Concrete $object, Data $fieldDefinition, $property, $currentValue) {
...
+     */
+    public static function cloneProperties(mixed $properties): mixed
```
blankse, line 1273:
```suggestion
    public static function cloneProperties(array $properties): array
```

models/Element/Service.php
dvesh3: I thought about making it specific at first. Just wondering if this could be used for other types in future.

blankse: Then we should at least use generics so that PhpStan/PhpStorm knows that the return value is the same as the parameter value.
```suggestion
     * @template T
     * @param T $properties
     * @return T
```

models/Element/Service.php
```diff
@@ -1263,6 +1263,23 @@ function (Concrete $object, Data $fieldDefinition, $property, $currentValue) {
         return $theCopy;
     }
 
+    /**
+     * @template T
+     * @param T $properties
+     * @return T
+     *
+     * @return mixed
```
blankse: Remove the second return
```suggestion
     * @template T
     *
     * @param T $properties
     *
     * @return T
```

kingjia90: thank you! i have just noticed after the commit but you were faster 😄

PR #11862, opened 2022-04-07
[Asset] Video thumbnail processing not working with remote storages

### LINKED ISSUE #11645: [Bug]: Video thumbnail generation failes in distributed environments

brusch:
… fixes #11645
fixes #11809

### DIFF
--- a/models/Asset/Video/Thumbnail/Processor.php
+++ b/models/Asset/Video/Thumbnail/Processor.php
@@ -128,7 +128,6 @@ class Processor
             $tmpPath = File::getLocalTempFilePath($format);
 
             $converter = \Pimcore\Video::getInstance();
-            $converter->load($sourceFile, ['asset' => $asset]);
             $converter->setAudioBitrate($config->getAudioBitrate());
             $converter->setVideoBitrate($config->getVideoBitrate());
             $converter->setFormat($format);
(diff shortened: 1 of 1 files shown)


brusch:
@bbdammert @denniskorbginski can you please check if this PR fixes your issue? Thanks in advance!

denniskorbginski:
Looks good, works for me 👍

brusch:
@denniskorbginski thanks for your feedback!

PR #11866, opened 2022-04-08
Add JavaScript events for "preDelete" and "postDelete" for Assets, Documents and Data objects.

timwiechers:
This PR will add JavaScript events for "preDelete" and "postDelete" for Assets, Documents and Data objects.

## Additional info  
In this PR only the ID of the deleted data type is passed to the event listeners. 
I think it could also be useful to pass the complete information about the affected data type. If there are no concerns about this, I could extend this PR to do so.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/elementservice.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/elementservice.js
@@ -123,8 +123,14 @@ pimcore.elementservice.deleteElementFromServer = function (r, options, button) {
 
         let ids = Ext.isString(id) ? id.split(',') : [id];
         ids.forEach(function (elementId) {
+           try {
+                pimcore.plugin.broker.fireEvent(preDeleteEventName, elementId);
+            } catch (e) {
+                pimcore.helpers.showPrettyError('asset', t("error"), t("delete_failed"), e.message);
+                return false;
+            }
+
             pimcore.helpers.addTreeNodeLoadingIndicator(elementType, elementId);
-            pimcore.plugin.broker.fireEvent(preDeleteEventName, elementId);
         });
 
         var affectedNodes = pimcore.elementservice.getAffectedNodes(elementType, id);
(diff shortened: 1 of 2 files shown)


brusch:
This should be also documented properly, see: 
https://github.com/pimcore/pimcore/blob/2f203b0ff7d23af18b213d00138c7972ba481371/doc/Development_Documentation/20_Extending_Pimcore/13_Bundle_Developers_Guide/06_Plugin_Backend_UI.md#L106

fashxp:
I think, we should allow to throw errors in the preDelete events ... similar to the preSave events, see [here](https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js#L746-L760)

For that I think it would be better to fire all events first ... and then start setting loading indicators and doing the delete jobs.

Could you adapt the PR accordingly?

fashxp:
@timwiechers ping

bundles/AdminBundle/Resources/public/js/pimcore/elementservice.js
```diff
@@ -119,10 +119,12 @@ pimcore.elementservice.deleteElementFromServer = function (r, options, button) {
         var successHandler = options["success"];
         var elementType = options.elementType;
         var id = options.id;
+        const preDeleteEventName = 'preDelete' + elementType.charAt(0).toUpperCase() + elementType.slice(1);
 
         let ids = Ext.isString(id) ? id.split(',') : [id];
         ids.forEach(function (elementId) {
             pimcore.helpers.addTreeNodeLoadingIndicator(elementType, elementId);
+            pimcore.plugin.broker.fireEvent(preDeleteEventName, elementId);
```
dpfahlbusch:
```suggestion
           try {
                pimcore.plugin.broker.fireEvent(preDeleteEventName, elementId);
            } catch (e) {
                pimcore.helpers.showPrettyError('asset', t("error"), t("delete_failed"), e.message);
                return false;
            }

            pimcore.helpers.addTreeNodeLoadingIndicator(elementType, elementId);
```

something like this @fashxp ?

timwiechers:
@fashxp i applied the suggested changes of @dpfahlbusch which implement your suggestion:

> I think, we should allow to throw errors in the preDelete events ... similar to the preSave events, see [here](https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/object/object.js#L746-L760)
> For that I think it would be better to fire all events first ... and then start setting loading indicators and doing the delete jobs.

fashxp:
@timwiechers @dpfahlbusch thx for the contributions, throwing errors didn't work though. 
I fixed it now...

PR #11874, opened 2022-04-09
Chromium: Fix new PhpStan Level 5 errors

blankse:
Followup to https://github.com/pimcore/pimcore/pull/10934
Noticed after merging 10.x branch to https://github.com/pimcore/pimcore/pull/11826


dvesh3:
@blankse thanks!

PR #11882, opened 2022-04-11
[Thumbnails] Fix auto-cleaning in PublicServicesController and unify getStream()

### LINKED ISSUE #11880: PublicServicesController deletes folder of thumbnails
### Expected behavior

Once a thumbnail is created and not changed, Pimcore should not delete it or re-generate it.

### Actual behavior

Pimcore deletes the Thumbnail Folder and therefore needs to re-create all images again.

### Steps to reproduce

Use remote storage and create a png thumbnail, then a webp. The png is now deleted. This goes on and on for different types. (https://github.com/pimcore/pimcore/blob/10.x/bundles/CoreBundle/Controller/PublicServicesController.php#L143)

brusch:
fixes #11880

Related to #11702

Made `\Pimcore\Model\Asset\Thumbnail\ImageThumbnailTrait::getStream()` to work the same as `\Pimcore\Model\Asset\Image\Thumbnail::getStream()`

### DIFF
--- a/bundles/CoreBundle/Controller/PublicServicesController.php
+++ b/bundles/CoreBundle/Controller/PublicServicesController.php
@@ -100,10 +100,6 @@ class PublicServicesController extends Controller
                         }
 
                         $imageThumbnail = $asset->getImageThumbnail($thumbnailConfig, $time);
-                        if (!$imageThumbnail->existsOnStorage()) {
-                            $asset->clearThumbnail($thumbnailConfig->getName());
-                        }
-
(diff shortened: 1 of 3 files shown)


dpfaffenbauer:
looks good to me

brusch:
Actually the whole thing got obsolete by 5ac4a712f4aee61a3b8648f8370a8162e984b231
Reason: the cache is only used when `$deferred = true`, since this is not the case when using `getStream()` the cache doesn't need to be considered, yay 🥳

PR #11884, opened 2022-04-11
[DataObjects] Quote columns in getBy* static getters

podarcis:
For static getters like getByFoo() (handled by magic Concrete::__callStatic)
the field name (foo) is not quoted in the query. This might result in an SQL
error when using SQL reserved words like "usage" (getByUsage()).

This fix quotes the column for both regular and localized fields.


aryaantony92:
@podarcis Thank you

PR #11889, opened 2022-04-11
QuantityValue: Abbreviation is in grid not translateable

blankse:
In the edit mode it is translated, but not in the grid view.


aryaantony92:
@blankse Thank you for the fix.

PR #11890, opened 2022-04-12
Fix: Element Router generates "element%5B____pimcore_cache_item__%5D" query param

blankse:
The param should be unset. The element should be only in the url.


brusch:
LGTM, thanks! 👍

blankse:
This was fast 😄 👍🏼

PR #11773, opened 2022-03-28
Custom reports shared user role fix

elements-team:
## Changes in this pull request  
Save of a Custom-Report was saving the ID instead of the Name for the sharedUserName and sharedRoleNames. 
This lead to Users not seeing their Reports, since the logic was expecting a name in the DB.

PR #11892, opened 2022-04-12
fix for Custom-Reports shared user/role id

dmaerzendorfer:
Same problem/fix as:
https://github.com/pimcore/pimcore/pull/11773

However, made a new pull request on my actual github account for better traceability.


dmaerzendorfer:
@aryaantony92

aryaantony92:
The failed tests were not related to the this change.

PR #11895, opened 2022-04-12
[IndexFieldSelection] Implement NormalizerInterface to support in blocks

### LINKED ISSUE #11861: [Bug]: Index Field Selection with "Multi Select from Remote" not working in Blocks
### Expected behavior

Data in blocks should be persisted in the database.

### Actual behavior

 Index Field Selection with "Multi-Select from Remote" is not persisted into a serialized PHP object, indifferent to inheritance settings.

### Steps to reproduce

Reproducible in the Pimcore Demo. A minimal class definition is provided in the attachment.  
Fill the block data, save, and reload. I will also take a further look into the problem.

indexFieldSelectionTest

[class_IndexFieldSelectionTest_export.zip](https://github.com/pimcore/pimcore/files/8440755/class_IndexFieldSelectionTest_export.zip)

mysliwietzflorian:

Resolves #11861

 Fix to support Index Field Selection in blocks again in PCX.

## Additional info  

- The other classes in the EcommerceFrameworkBundle seem to work fine (those classes implicitly inherit that interface).
- No real type check is performed before the construction of the `ObjectData\IndexFieldSelection`. Should be correct from DB and could intervene when the constructor parameters of `ObjectData\IndexFieldSelection` will be typed in the future.


mcop1:
Thank you

mcop1:
Thank you

PR #11899, opened 2022-04-12
[Admin] Improve tree view perfomance for non-admin users

### LINKED ISSUE #6383: [Core][DataObjects] Severe performance degradation when opening folders with many objects in the tree

kingjia90:
improves #6383


kingjia90:
getUserPermissions required 14 consecutive queries (one for each permission/column), this PR changes reduce to 3.
Due the nature of the user level permission having more power than role level permission, i am trying to "shortcircuit" when found one by `if ($parentRow['userId'] == $currentUserId){`, 
if not found we check the roles (having the same `cid` from first query, it guarantees that there are role level but not any workspace at the user level) and do the additive calculation (breaking the sub-loop as soon as permission=1 is found).
Another potential benefit of this approach is that when `cid` for the element itself is not found, we check the parent one on the second query and the db engine should be internally "cache" it, compared to the [isAllowed](https://github.com/pimcore/pimcore/blob/89ecd1b81d4e955164027ca3a9129fa4b0d82679/models/DataObject/AbstractObject/Dao.php#L499) which is more one-time use only.

some reference values that i have noticed during manual tests:
when listing a folder with 30 data objects (the default max per pagination)

admin: 200-250ms
with this PR: 450-500ms
with #11714 changes: 600-750ms
prior #11714: 900ms-1s+

more details here: https://github.com/pimcore/pimcore/issues/6383#issuecomment-1073309284

kingjia90:
Both tests  (before/after the PR) are matching and passing.


Changes required to test the before:
In pimcore\models\Element\AbstractElement.php
add `$user` parameter on
`public function getUserPermissions($user = null)` and `$permissions[$name] = $this->isAllowed($name, $user);`
then `public function isAllowed` should return (int) instead of (bool), because `false` would just be empty and in the test i am using `0`, so the array_diff_assoc wrongly triggers.

fashxp (changes requested):
- [x] add note to docs that for list permission in listings, only further restricting, but not further allowing is possible. 
list=1 -> list=0 --> possible
list=0 -> list=1 --> NOT possible

tests/Model/Permissions/ModelDocumentPermissionsTest.php
```diff
@@ -301,6 +309,52 @@ public function testIsAllowed()
...
+        // User 1 doesn't have publish=1 based on the inherited `Testrole`
+        $this->doAreAllowedTest($this->groupfolder, ['save'=> 1,'publish' =>1], true, false, true);
+        // Check admin and User2 are not simply returning all true to pass the above test, confirms User2's user-level permissions getting priority over individual roles
+        $this->doAreAllowedTest($this->groupfolder, ['save'=> 1,'publish' =>1 ,'versions' => 0], false, false, true);
+        // `Dummyrole`(0,0) is the "weakest" when conflicting, so it should never true, User 2 doesn't have user-level permission, so this confirms the additive behaviour on conflicting roles
+        $this->doAreAllowedTest($this->groupfolder, ['save'=> 0,'publish' =>0], false, false, false);
+        // Checks User1 is correctly getting the workspace rule from roles
+        $this->doAreAllowedTest($this->groupfolder, ['save'=> 1,'publish' =>0], false, true, false);
+
+//        //check when no parent workspace is found, it should be allow list=1 when children are found, in this case for admin and user1 to get to `c`
```
fashxp: why this test is commented?

tests/Model/Permissions/ModelAssetPermissionsTest.php
```diff
@@ -301,6 +309,52 @@ public function testIsAllowed()
...
+        // User 1 doesn't have publish=1 based on the inherited `Testrole`
+        $this->doAreAllowedTest($this->groupfolder, ['delete'=> 1,'publish' =>1], true, false, true);
+        // Check admin and User2 are not simply returning all true to pass the above test, confirms User2's user-level permissions getting priority over individual roles
+        $this->doAreAllowedTest($this->groupfolder, ['delete'=> 1,'publish' =>1 ,'versions' => 0], false, false, true);
+        // `Dummyrole`(0,0) is the "weakest" when conflicting, so it should never true, User 2 doesn't have user-level permission, so this confirms the additive behaviour on conflicting roles
+        $this->doAreAllowedTest($this->groupfolder, ['delete'=> 0,'publish' =>0], false, false, false);
+        // Checks User1 is correctly getting the workspace rule from roles
+        $this->doAreAllowedTest($this->groupfolder, ['delete'=> 1,'publish' =>0], false, true, false);
+
+        //check when no parent workspace is found, it should be allow list=1 when children are found, in this case for admin and user1 to get to `c`
```
fashxp: why this test is commented?

fashxp (changes requested):

fashxp:
- [x] I think we should rebase that for 10.x ... as it is no bugfix, but an improvement.

fashxp:
we first merge https://github.com/pimcore/pimcore/pull/11938 ... and then we can merge that one

fashxp (approved):

fashxp:
my tests show following improvements with a folder of 20.000 direct child objects on a default local docker setup on a windows machine:  

| action | 10.3.4 | prior that PR | with that PR | 
| ------- | ------ | -------------- | ------------- |
| grid proxy (loading 25 items) | 1.3 sec | 1.3 sec | 1.0 sec |
| tree get childs (loading 30 items) | 2.7sec | 1.5 sec | 1.2 sec |

I did not notice any significant changes for admin users, they always were around 300ms per request. 

That's what we can archive right now.

kingjia90:
Interesting data, It would be nice to have the tests for grid proxy with pagination 200 (how that 300ms improvement is by scaling up) and one for one item in one folder (to see the minimum required time no matter how many items)

PR #11907, opened 2022-04-13
Create redirect if pretty url is changed

### LINKED ISSUE #11592: [Improvement]: Create redirect if Pretty URL is changed
### Improvement description

Currently, if there is an existing Pretty URL for a document and you change it to a new one, opening the previous Pretty URL will result in a 404.

It would be nice if a redirect would automatically be created in this case.

matthiashamacher:
Resolves #11592 

## Additional info  
A redirect is only created if:
- The pretty URL has changed
- The new pretty URL is not empty (if the pretty URL was removed a 404 is the correct error code)
- The old pretty URL is not empty(if there was no pretty URL before we do not need to redirect)

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/PageController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/PageController.php
@@ -232,6 +232,7 @@ class PageController extends DocumentControllerBase
                 $redirect->setSource($oldPage->getPrettyUrl());
                 $redirect->setTarget($page->getPrettyUrl());
                 $redirect->setStatusCode(301);
+                $redirect->setType(Redirect::TYPE_AUTO_CREATE);
                 $redirect->save();
             }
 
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/Document/PageController.php
```diff
@@ -222,6 +223,18 @@ public function saveAction(Request $request, StaticPageGenerator $staticPageGene
                 $data['staticLastGenerated'] = $staticPageGenerator->getLastModified($page);
             }
 
+            if ($page->getPrettyUrl() !== $oldPage->getPrettyUrl()
+                && empty($oldPage->getPrettyUrl()) === false
+                && empty($page->getPrettyUrl()) === false
+            ) {
+                $redirect = new Redirect();
+
+                $redirect->setSource($oldPage->getPrettyUrl());
```
aryaantony92, line 232: You should set the redirect type also while creating the Redirects. 
$redirect->setType(Redirect::TYPE_AUTO_CREATE);

aryaantony92:
@matthiashamacher Thank you

PR #11910, opened 2022-04-14
Change sort method permission

### LINKED ISSUE #11716: [Improvement]: Allow change of sort method of data object children based on a user role
### Improvement description

Since Pimcore 6.9 only admin users are allowed to change the sort method (key or index) of children of data objects.
Is there a specific reason why?
And would it be possible to change this behaviour so users with a specific role are able to change the sort method?

That's the old pull request which introduced the changes vor Pimcore 6.9: https://github.com/pimcore/pimcore/pull/9307

If you accept the change from admin user only to a specific user role, i would be able to work on a new pull request for this improvement.

andi-schlosser:
Resolves #11716

## Additional info  

A backend user with the new permission `objects_sort_method` is now able to change the sorting method (key or index) of children of data objects without needing to be a full admin.

### DIFF
--- a/bundles/CoreBundle/Resources/translations/de.json
+++ b/bundles/CoreBundle/Resources/translations/de.json
@@ -967,5 +967,6 @@
     "asset_upload_want_to_overwrite": "Do you want to overwrite file %s?",
     "asset_upload_overwrite": "Overwrite",
     "asset_upload_keep_both": "Keep both files",
-    "asset_upload_overwrite_all": "Overwrite all"
+    "asset_upload_overwrite_all": "Overwrite all",
+    "change_data_objects_sort_method": "Sortiermethode für Data Objects ändern"
 }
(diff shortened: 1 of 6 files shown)


bundles/CoreBundle/Resources/translations/de.json
```diff
@@ -967,5 +967,6 @@
     "asset_upload_want_to_overwrite": "Do you want to overwrite file %s?",
     "asset_upload_overwrite": "Overwrite",
     "asset_upload_keep_both": "Keep both files",
-    "asset_upload_overwrite_all": "Overwrite all"
+    "asset_upload_overwrite_all": "Overwrite all",
+    "objects_sort_method": "Sortiermethode für Data Objects ändern"
```
fashxp: please only provide english translations. all other translations are provided via POeditor.

fashxp (changes requested):
thx very much, looks good to me. 
could you please remove the german translation.

bundles/CoreBundle/Resources/translations/de.json
andi-schlosser: Thank you for your feedback. I removed the german translation.

fashxp (approved):

PR #11916, opened 2022-04-15
Overwrite the redirect status codes in yaml file

alyna:
Allow to overwrite the redirect status codes.
See: https://github.com/pimcore/pimcore/issues/11282


dvesh3:
@alyna Hi, thanks for your PR. please sign the CLA(by clicking on CLAassistant button) as this is your first contribution.

brusch:
@alyna also docs are missing 😉 Thanks

bundles/AdminBundle/Controller/Admin/RedirectsController.php
```diff
@@ -264,4 +264,28 @@ public function cleanupAction()
...
+     */
+    public function statusCodesAction()
+    {
+        $this->checkPermission('redirects');
+        $statusCodes = Redirect::getStatusCodes();
+        $codes = [];
+        foreach ($statusCodes as $statusCode => $label) {
+            $codes[] = [
+                'statusCode' => $statusCode,
+                'display' => $label . " ($statusCode)",
```
jdreesen: I think mixing variable expansion and string concatenation is somehow strange and unnecessary:
```suggestion
                'display' => "$label ($statusCode)",
```

models/Redirect.php
```diff
@@ -569,6 +570,6 @@ public function delete()
      */
     public static function getStatusCodes(): array
     {
-        return self::$statusCodes;
+        return Config::getSystemConfiguration('redirects')['status_codes'] ?: self::$statusCodes;
```
dvesh3: we should completely move the status codes list to the Symfony config(define the array list in the `default.yaml` of CoreBundle) instead of falling back to the hard coded property. So one can add to the list just by defining more codes in the config.yaml.

Also the node should be part of `pimcore` config not `pimcore_admin`.

These should be done:
- [ ] Add redirects node under pimcore config and define default status codes in default.yaml:
```yaml
pimcore:
    redirects:
        status_codes:
            300: 'Multiple Choices'
            301: 'Moved Permanently'
            302: 'Found'
            303: 'See Other'
            307: 'Temporary Redirect'

```
- [x] Remove hardcoded values in Redirect.php
- [x] Update docs to explain to add custom status codes.

Thanks!

doc/Development_Documentation/02_MVC/04_Routing_and_URLs/04_Redirects.md
```diff
@@ -39,3 +39,13 @@ Pimcore provides the ability to automatically create Redirects when renaming and
 ![Redirect Priority](../../img/redirects4.png)
 
 ![Redirect Priority](../../img/redirects5.png)
+
+
+#### Creating custom redirect status codes
+The redirect status codes can be overwritten by adding custom codes in a config file. (ex:redirects.yaml)
+
+![Redirect Priority](../../img/redirects6.png)
```
dvesh3, line 47: I would go for code example here:
```yaml
pimcore:
    redirects:
        status_codes:
            409: Conflict
            410: Gone

```

bundles/AdminBundle/Controller/Admin/RedirectsController.php
```diff
@@ -264,4 +264,28 @@ public function cleanupAction()
             return $this->adminJson(['success' => false]);
         }
     }
+
+    /**
+     * @Route("/get-statuscodes", name="pimcore_admin_redirects_statuscodes", methods={"GET"})
+     *
+     * @return \Symfony\Component\HttpFoundation\JsonResponse
+     */
+    public function statusCodesAction()
```
dvesh3:
```suggestion
    /**
     * @Route("/get-statuscodes", name="pimcore_admin_redirects_statuscodes", methods={"GET"})
     *
     * @return JsonResponse
     */
    public function statusCodesAction(): JsonResponse
```

dvesh3 (commented):
so you cannot assign non-redirection status codes to [RedirectReponse ](https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/lib/Routing/RedirectHandler.php#L231)as Symfony throws exception in that case:
image

Should we catch the exception and create a Response with the status code..but then it is not a redirect anymore? now i'm asking myself it really makes sense to introduce this change 😞

dvesh3 (approved):
Ok. we decided to go with this PR as it anyway improves the extensibility and in the end it is developer's responsibility to use the correct status code for Redirects.

dvesh3:
@alyna thanks a lot!

alyna:
> so you cannot assign non-redirection status codes to [RedirectReponse ](https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/lib/Routing/RedirectHandler.php#L231)as Symfony throws exception in that case: image
> 
> Should we catch the exception and create a Response with the status code..but then it is not a redirect anymore? now i'm asking myself it really makes sense to introduce this change disappointed

Hello,
Regarding this exception that Symfony throws for non-redirection status codes would it be ok if we dispatch an event after line 230 in order to be able to handle the behaviour for these status codes?

dvesh3:
> Hello,
Regarding this exception that Symfony throws for non-redirection status codes would it be ok if we dispatch an event after line 230 in order to be able to handle the behaviour for these status codes?

@alyna and how you will handle it? coz anyway these codes should not be used in case of redirection.

alyna:
> > Hello,
> > Regarding this exception that Symfony throws for non-redirection status codes would it be ok if we dispatch an event after line 230 in order to be able to handle the behaviour for these status codes?
> 
> @alyna and how you will handle it? coz anyway these codes should not be used in case of redirection.

Sometimes, especially during migration of older system, for some links first redirects are in place and after some time they should be result in a 410 Gone response. The redirects user interface could stay the same only the target would be empty or ignored. In the event listener we could just throw a GoneHttpException in case the status is 410 and show an according error page via an exception listener.

dvesh3:
@alyna please provide a follow up PR to add event `pimcore.redirect.preBuild`. thanks!

PR #11919, opened 2022-04-15
[Improvement][JS][Dataobject]: Adding willClose property (save,publish and close/publish and close )

### LINKED ISSUE #11823: [Improvement]: JavaScript: Conflict between "onPostUpdate" event and closing of an object.
### Expected behavior

When I use the "postSaveObject" event to reload an object after saving it (because for example data was changed server-side after saving, see "Steps to reproduce"), I have the information if the object should be closed after saving, so I can decide if the reload is necessary.

### Actual behavior

I don't see a way to determine if the object will be closed and therefore cannot decide wether to reload the object or not. 
So after clicking on the "Save, publish and close" button, the reload always opens it again.

### Steps to reproduce

1. Create a simple plugin, that reloads an object after it was updated:
```

        getClassName: function () {
        },

        initialize: function () {
        },

            object.reload();
        },
    });

```
2. Click "Save, publish and close"-Button.
3. Object gets reopened.

kingjia90:
improves #11823

~~I wouldn't consider it as a bug, but~~ this change will give more versatility (eg. decide wheter reload or not like described in #11823 or implementing a post save event listener that "cancels" the closing action, by turning `object.willClose` to `false`.
It seems it that makes 'callback' parameter in save() obsolete because it was mostly used to do the closing.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js
@@ -13,7 +13,7 @@
 
 pimcore.registerNS("pimcore.asset.asset");
 pimcore.asset.asset = Class.create(pimcore.element.abstract, {
-
+    willClose: false,
     getData: function () {
         Ext.Ajax.request({
             url: Routing.generate('pimcore_admin_asset_getdatabyid'),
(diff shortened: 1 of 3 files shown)


fashxp (dismissed):
please rebase to 10.x

PR #11923, opened 2022-04-19
[Admin] Fix displaying multiselect with options provider and render type "list"

### LINKED ISSUE #11623: [Bug]: Inconsistent behavior with Multiselect using OptionsProvider providing numeric values
### Expected behavior

When the options provider returns an integer (e.g. Pimcore ID of an object) for `value`, for instance: ` ['key' => 'foobar', 'value' => 1]`:

- Persisted values of multiselect should be the selected options when opening the object.
- Consistent behavior regardless of render type "List" or "Tags".

### Actual behavior

- Multiselect (render type "List") has no selected options reflecting the persisted values.
- Render type "List" and "Tags" should behave the same - here only render type "List" is affected.

### Steps to reproduce

1. Create a DataObject class.
2. Add a field of type Multiselect with render type "List" (default).
3. Make it a "Dynamic Multiselect" by specifying an options provider (see below). Return valid options according to the docs of "Dynamic Select Typed" in format `['key' => 'foobar', 'value' => 1]` - use an integer for value.
4. Open the data object, select multiple options and save.
5. Reload the object and discover that no option is selected (value is persisted correctly, though).

## Possible solutions:

1. Specify the option provider option format to always return a string for value and fix the documentation.
2. Server-side: Transform/cast the options' values provided by the options provider.
3. Client-side `pimcore.object.tags.multiselect`: Cast the `value` in `prepareStoreDataAndFilterLabels()` to string.

## Code snippet options provider

```php
<?php

namespace App;

use Pimcore\Model\DataObject;

class MultiOptionsProvider implements DataObject\ClassDefinition\DynamicOptionsProvider\MultiSelectOptionsProviderInterface
{

    /**
     * @inheritDoc
     */
    public function getOptions($context, $fieldDefinition)
    {
        return [
            ['key' => 'foo1', 'value' => 1],
            ['key' => 'foo2', 'value' => 2],
            ['key' => 'foo3', 'value' => 3],
        ];
    }

    /**
     * @inheritDoc
     */
    public function hasStaticOptions($context, $fieldDefinition)
    {
        return true;
    }
}
```

podarcis:
If a multiselect (render type "list") with an options provider returns numeric IDs
(as given as an example in the documentation) the multiselect element isn't correctly
rendered in the edit layout: The set items are not selected.

This fix forces the "id" field of the store to be handled as a string, exactly as a
static definition of items would be.

Resolves #11623

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/multiselect.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/multiselect.js
@@ -74,7 +74,7 @@ pimcore.object.tags.multiselect = Class.create(pimcore.object.tags.abstract, {
         var storeData = this.prepareStoreDataAndFilterLabels(field.layout);
 
         var store = Ext.create('Ext.data.JsonStore', {
-            fields: ['id', 'text'],
+            fields: [{name: 'id', type: 'string'}, 'text'],
             data: storeData
         });
 
@@ -120,7 +120,7 @@ pimcore.object.tags.multiselect = Class.create(pimcore.object.tags.abstract, {
         }
 
         var store = Ext.create('Ext.data.Store', {
-            fields: ['id', 'text'],
+            fields: [{name: 'id', type: 'string'}, 'text'],
             data: storeData
         });
 


mcop1:
Thank you!

PR #11924, opened 2022-04-19
[Grid] Remove unnecessary child count when loading classes

BlackbitDevs:
When loading the data object classes for the grid, the call
https://github.com/pimcore/pimcore/blob/fad54cc4ee3ddd4c177ecc4a7972a1f86b819000/models/DataObject/AbstractObject/Dao.php#L469
is expensive and not necessary (needs 5 seconds in our case with > 4M objects in a folder)

This PR is a follow-up of #11856
Improves problems mentioned in #9727


dvesh3:
@BlackbitNeueMedien looks good...thanks! 👍

PR #11938, opened 2022-04-20
[Bug][Search]: Fixes Search and Quick Search not returning children when parent folder is list=0

### LINKED ISSUE #11822: [Bug] Search doesn't show results under certain cirumstances
### Expected behavior

search finds all matching elements

### Actual behavior

search doesn't find all elements, see also https://github.com/pimcore/pimcore/issues/6771#issuecomment-674850816

### Steps to reproduce

- create dataset with many (> limit size of find action, default = 25) similar elements, e.g. all elements containing some string `manyelement` 
- add additional element with specific string e.g. `manyelement X`
- create user and make sure, that this user only sees `manyelement X`, but not the other elements
- search for `manyelement` 
 --> user doesn't find any element, also not `manyelement X`

### LINKED ISSUE #11888: [Bug]: Quicksearch delivers invalid results with workspace conditions
### Expected behavior

The quicksearch should list all matches in documents/assets/objects according to the searchterm and considering current allowed user workspaces. 

### Actual behavior

If the current user or role has workspace definitions without the list permission, the quicksearch results are invalid because the corresponding SQL condition to consider the user workspaces is incorrect.

This bug is somehow related to https://github.com/pimcore/pimcore/issues/11822 but describes a seperate issue.

On Pimcore\Bundle\AdminBundle\Controller\Searchadmin\SearchController::quicksearchAction() the SQL query is extended for current user workspaces via "getForbiddenCondition()" function.
The function adds something like 
`(maintype = 'asset' AND fullpath not like " . $db->quote($forbiddenAssetPaths[$i] . '%') . ')`
to the query, e. g. for asset workspaces without list permission.

But this extended condition not only excludes the forbidden asset paths, it also excludes the entire maintypes "object" and "document".

That means, if you define a forbidden asset workspace you won't find anymore objects or documents although the user may be actually allowed to list them.

The condition in this case above should be something like:

`(maintype != 'asset' OR (maintype = 'asset' and fullpath NOT LIKE '%forbiddenpath%'))`

### Steps to reproduce

1. Fresh install skeleton Pimcore 10.3.4
2. Create test object class and test object to search for
3. Create different user and configure: 
 - grant objects, assets permission
 - define a forbidden asset workspace without list permission
 - define object workspace on root level with all permissions
5. Login as admin user
6. Seaerch for the created object => object is listed in quicksearch
7. Login as the different user
8. Search for the created object => object is not listed in quicksearch matches

kingjia90:
fixes #11888, #11822

tldr: search worked but wasn't efficiently filtering the results in the listing, quicksearch blocked allowed children due a folder with permission `list=0`

CHANGES: Prior this PR, it took one folder with list permission disabled to disallow all its children (because it worked on a `LIKE path%` condition, any longer path would be blocked), also did not follow the rule that allowed children should turn the parent folder to list=1 to allow path traversal and didn't properly follow the [precedence ](https://github.com/pimcore/pimcore/issues/1634#issuecomment-831885117) on user level permissions and the conflicting rules on roles.

The `getForbiddenCondition` was skipped in `findAction` because the `type` in the request payload is singular, while in the `getForbiddenCondition` worked on plural ones, so it was scanning the entire `search_backend_data` (after the other filter/conditions) and relied on `($element->isAllowed('list')`, which explains why it worked at some extend, but with this PR, there are surely great reduction of this isAllowed calls/queries.
Quicksearch wasn't affected but had problems with the permissions priority rules. 
After PR, both follow the same filtering rules and results should be exactly the same.

TESTS: re-enables search related tests which now successfully passes, fixes codecep tests workspace typos.

PERFOMANCE: I have not noticed any big bump from Admin times (which skips this query conditions building). 
Prior PR quicksearch returned wrong results, so cannot do a direct comparison.
Slightly improves quicksearchByIdAction (the preview frame) perfomances (doesn't need getForbiddenCondition since it search directly by `id`).

### DIFF
--- a/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
+++ b/bundles/AdminBundle/Controller/Searchadmin/SearchController.php
@@ -90,7 +90,7 @@ class SearchController extends AdminController
         $forbiddenConditions = $this->getForbiddenCondition($types);
 
         if ($forbiddenConditions) {
-            $conditionParts[] = '(' . implode(' AND ', $forbiddenConditions) . ')';
+            $conditionParts[] = '(' . implode(' OR ', $forbiddenConditions) . ')';
         }
 
         $queryCondition = '';
@@ -304,7 +304,7 @@ class SearchController extends AdminController
         $elements = [];
         foreach ($hits as $hit) {
             $element = Element\Service::getElementById($hit->getId()->getType(), $hit->getId()->getId());
-            if ($element->isAllowed('list')) {
+//            if ($element->isAllowed('list')) {
                 $data = null;
                 if ($element instanceof DataObject\AbstractObject) {
                     $data = DataObject\Service::gridObjectData($element, $fields);
@@ -317,10 +317,10 @@ class SearchController extends AdminController
                 if ($data) {
                     $elements[] = $data;
                 }
-            } else {
+//            } else {
                 //TODO: any message that view is blocked?
                 //$data = Element\Service::gridElementData($element);
-            }
+//            }
         }
 
         // only get the real total-count when the limit parameter is given otherwise use the default limit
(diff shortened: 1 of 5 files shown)


bundles/AdminBundle/Controller/Searchadmin/SearchController.php
```diff
@@ -453,7 +436,7 @@ public function quicksearchAction(Request $request, EventDispatcherInterface $ev
 
         $forbiddenConditions = $this->getForbiddenCondition();
         if ($forbiddenConditions) {
-            $conditionParts[] = '(' . implode(' AND ', $forbiddenConditions) . ')';
+            $conditionParts[] = '(' . implode(' OR ', $forbiddenConditions) . ')';
```
fashxp: why `OR`?

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
```diff
@@ -90,7 +90,7 @@ public function findAction(Request $request, EventDispatcherInterface $eventDisp
         $forbiddenConditions = $this->getForbiddenCondition($types);
 
         if ($forbiddenConditions) {
-            $conditionParts[] = '(' . implode(' AND ', $forbiddenConditions) . ')';
+            $conditionParts[] = '(' . implode(' OR ', $forbiddenConditions) . ')';
```
fashxp: why `OR`?

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
```diff
@@ -304,7 +304,7 @@ public function findAction(Request $request, EventDispatcherInterface $eventDisp
         $elements = [];
         foreach ($hits as $hit) {
             $element = Element\Service::getElementById($hit->getId()->getType(), $hit->getId()->getId());
-            if ($element->isAllowed('list')) {
+//            if ($element->isAllowed('list')) {
```
fashxp: can remove that completely, right?

models/Element/Service.php
```diff
@@ -718,31 +718,81 @@ public static function gridElementData(ElementInterface $element)
...
+                    $allowed[] = $path;
                 }
             }
         } else {
-            $forbidden[] = '/';
+            $forbidden['/'] = [];
         }
 
-        return $forbidden;
+        return array('forbidden' => $forbidden, 'allowed' => $allowed);
```
fashxp, line 792: change return value will break our advanced object search, see [here](https://github.com/pimcore/advanced-object-search/blob/e74ffaa8266414e97003966d6ae90f03c98e4a3c/src/Service.php#L792)
... so we need to add a BC layer in the advanced object search.

fashxp (changes requested):
it breaks quick search now ... added some tests to cover that.

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
kingjia90: cannot be removed beucase of `Pimcore\Event\AdminEvents::ELEMENT_PERMISSION_IS_ALLOWED` which need to be decoupled from isAllowed

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
kingjia90: nice catch, it seems that i was trying an alternate approach and forgot to revert it. changing this also fixes the quicksearch and the tests 😄

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
kingjia90: i would suggest to have all these changes in a separate PR that splits this `isAllowed` from `ELEMENT_PERMISSION_IS_ALLOWED` event and extends the tests to take in account of these permissions changed by events, wdyt?

models/Element/Service.php
kingjia90: mhh, i've missed that, what about keeping `findForbiddenPaths()` and instead put the new changes from the PR in a new method `getAllowedAndForbiddenPaths()`?

models/Element/Service.php
```diff
@@ -718,31 +718,81 @@ public static function gridElementData(ElementInterface $element)
...
+            WHERE userId IN (' . implode(',', $userIds) . ') AND cid NOT IN ('.implode(',', $userCid).')
+            GROUP BY cpath
+        ';
+        $roleWorkspaces = $db->fetchAll($roleWorkspacesSql);
+
+        $allWorkspaces = array_merge($userWorkspaces, $roleWorkspaces);
+
+        $uniquePaths = [];
+        foreach ($allWorkspaces as $workspace){
+            $uniquePaths[$workspace['cpath']] = $workspace['list'];
```
fashxp, line 757: doesn't here the role permission potentially overwrite the user permission with the same path?

models/Element/Service.php
fashxp: since this is internal, I'm fine with fixing the advanced object search ... at some point we also should completely fix the permissions there too. but that is a different task then.

fashxp (changes requested):

bundles/AdminBundle/Controller/Searchadmin/SearchController.php
fashxp: I reverted the isAllowed in quick search too.

models/Element/Service.php
fashxp: following issue needs to be fixed after this is merged: https://github.com/pimcore/advanced-object-search/issues/149

kingjia90: > doesn't here the role permission potentially overwrite the user permission with the same path?

No, it shouldn't, the role permissions query have this condition
`cid NOT IN ('.implode(',', $userCid).')`, which also means that the cpath would be excluded, if found in the first query (which gets the user level)

fashxp:
following issue needs to be fixed after this is merged: https://github.com/pimcore/advanced-object-search/issues/149

fashxp (approved):

PR #11953, opened 2022-04-24
[BUG] Add showAppLogger to imported properties on classdefinition import

asysrieke:
On import of classdefinitions the property "showAppLoggerTab" is missed out.


mcop1:
Thanks!

PR #11954, opened 2022-04-25
[Grid] Enhance performance to get data object classes for opened folder

BlackbitDevs:
Follow-up of #11856

I noticed that most of the time (actually on every project I checked) the query https://github.com/pimcore/pimcore/blob/75bbc5bbacbd57383ccb20e52bbb151397b81a78/models/DataObject/AbstractObject/Dao.php#L478 uses the index `type` - even if all entries in the `objects` table have `o_type='object'`. It would be better to use an index for `o_path`.

I saw 2 alternatives: 
1. Add `USE INDEX (fullpath)` but I actually do not like this because it complicates maintenance when the query gets changed in future
2. Create composite index on `o_type`, `o_path`

This PR implements 2.

Here is the comparison:

**Without this PR**:
```sql
explain SELECT o_classId FROM objects WHERE o_path LIKE '/Artikel/%' AND o_type = 'object' AND o_classId NOT IN ('Artikel') LIMIT 1
```
| id | select_type | table   | partitions | type | possible_keys                                        | key    | key_len | ref   | rows    | filtered | Extra                              |
|----|-------------|---------|------------|------|------------------------------------------------------|--------|---------|-------|---------|----------|------------------------------------|
| 1  | SIMPLE      | objects | NULL       | ref  | fullpath,o_classId,data_director_classId_path,type | type | 2       | const | 3080750 | 16.76    | Using index condition; Using where |

Execution Time: 24.858 s

**With this PR**:
| id | select_type | table   | partitions | type  | possible_keys                                           | key       | key_len | ref  | rows    | filtered | Extra                              |
|----|-------------|---------|------------|-------|---------------------------------------------------------|-----------|---------|------|---------|----------|------------------------------------|
| 1  | SIMPLE      | objects | NULL       | range | fullpath,o_classId,data_director_classId_path,type_path | type_path | 2300    | NULL | 1718724 | 50.01    | Using index condition; Using where |

Execution Time: 1.509 s

PS: As the index `type` previously contained only the column `o_type` and MySQL uses index prefixes, it is safe to extend the index by another column - should have no influence on other queries which use this index.

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20220425082914.php
+++ b/bundles/CoreBundle/Migrations/Version20220425082914.php
@@ -29,11 +29,11 @@ final class Version20220425082914 extends AbstractMigration
 
     public function up(Schema $schema): void
     {
-        $this->addSql('ALTER TABLE `objects` DROP INDEX `type`, ADD INDEX `type_path` (o_type, o_path)');
+        $this->addSql('ALTER TABLE `objects` DROP INDEX `type`, ADD INDEX `type_path_classId` (o_type, o_path, o_classId)');
     }
 
     public function down(Schema $schema): void
     {
-        $this->addSql('ALTER TABLE `objects` DROP INDEX `type_path`, ADD INDEX `type` (o_type)');
+        $this->addSql('ALTER TABLE `objects` DROP INDEX `type_path_classId`, ADD INDEX `type` (o_type)');
     }
 }
--- a/bundles/CoreBundle/Migrations/Version20220425082914.php
+++ b/bundles/CoreBundle/Migrations/Version20220425082914.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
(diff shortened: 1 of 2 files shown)


brusch:
Hmm, the benefit I see in option 1 is, that it keeps the index leaner, because with millions of objects the index could get quite bloated, not?

BlackbitDevs:
Hmm, but `fullpath` index also contains `o_key` and so would not be optimal for this query. 

Interesting: I just created another composite index on `o_path` and `o_classId` and this runs above query in 0.3s.
Here complete comparison:
* USE INDEX (fullpath): 1.396s
* composite index on `o_type` and `o_path`: 1.430s (this is what this PR currently implements)
* composite index on `o_path` and `o_classId`: 0.318s
* USE INDEX (o_type): 23.518s (current behaviour without this PR)

What do you think, @brusch ? I think https://github.com/pimcore/pimcore/issues/9727#issue-942073498 is right
> Expected behaviour: Pimcore optimizes queries inside admin interface to the fullest.

Which way would you prefer?

brusch:
Then let's go for the fastest option 😎 Plz don't forget about the `install.sql` 😉

BlackbitDevs:
Done. I also added `o_type` to the index, otherwise MySQL always uses the single-column index `type`. Alternatively we could only drop the `type` index and only use columns `o_path` and `o_classId` for the new index but I think there was a reason when the index in `o_type` got created, so I would keep this (as this is still covered with the expanded index).

brusch:
@BlackbitNeueMedien I'm just wondering whether we run into a max key length issue with the current key, we have: 
- `o_key` 255 * 3 = 765 bytes (as it is 3-byte utf8)
- `o_path` 765 * 3 = 2 295 bytes (as it is 3-byte utf8)
- `o_classId` 11 bytes
- `o_type` 5 bytes

= 3 077 bytes 

where the limit is as 3072 bytes

BlackbitDevs:
But `o_key` is not part of the index.
`EXPLAIN SELECT sql_no_cache o_classId FROM objects WHERE o_path LIKE '/Artikel/%' AND o_type = 'object' AND o_classId NOT IN ('Artikel') LIMIT 1;`
says `key_len`=2503

brusch:
@BlackbitNeueMedien omg - sorry, my fault 🤦‍♂️

PR #11957, opened 2022-04-25
[BUG] Fix for dependency table indexes.

tanaydin:
Fixed dependency queries for using indexes.

## Additional info  
There was composite indexes in dependency table we can use. But in queries we were using wrong order.


brusch:
🍒-picked into `10.3` as it is a bug fix. (see 9da078457a598be65d03c6148cf735483037fe25 )

PR #11959, opened 2022-04-25
[Bug]: uuids deadlock issue on delete, extra update query

hstjep:
When deleting objects, from multiple connections at the same time, deadlocks often happen on the "uuids" table. This happens if two or more users delete their own objects for their accounts. For example, it can also be reproduced if two or more admin users delete different folders from administration, with larger amount of objects, at the same time.  

uiids-deadlock-log

MySql version: 5.7.24 (also reproduced on 8.0.28)
Isolation level: REPEATABLE-READ
The uuids table has over 500000 records. 

In the query log, I noticed that the following query is generated:
START TRANSACTION
...
INSERT INTO `uuids` (`itemId`, `type`, `uuid`, `instanceIdentifier`) VALUES ('278219', 'object', '22f4d69d-b1e9-58e4-80d8-91eb0ca9f1d7', 'demo-ecommerce') ON DUPLICATE KEY UPDATE `itemId` = '278219', `type` = 'object', `uuid` = '22f4d69d-b1e9-58e4-80d8-91eb0ca9f1d7', `instanceIdentifier` = 'demo-ecommerce'
DELETE FROM uuids WHERE uuid = '22f4d69d-b1e9-58e4-80d8-91eb0ca9f1d7'
COMMIT

I think that "INSERT INTO ... ON DUPLICATE KEY UPDATE" increases the chance for deadlock as it adds exclusive gap lock (X). Then, DELETE fails when two other connections hold insert intention lock and gap lock at the same time.

Also, it seems like UPDATE is not really necessary to be executed every time when deleting uuid. I would suggest, when creating uiid, to first check if uuid exists. If not, skip save (as nothing needs to be changed). Otherwise, if does not exist, it will perform INSERT which doesn't use gap lock (X, REC_NOT_GAP). Please check PR code.

After this change, deadlocks are not happening anymore and both, delete and insert of uuid work fine.

### DIFF
--- a/models/Tool/UUID.php
+++ b/models/Tool/UUID.php
@@ -156,7 +156,7 @@ final class UUID extends Model\AbstractModel
         $uuid = Uid::v5($namespace, $this->getInstanceIdentifier() . '~' . $this->getType() . '~' . $this->getItemId());
         $this->uuid = $uuid->toRfc4122();
 
-        if (!$this->uuidExists($this->uuid)) {
+        if (!$this->getDao()->exists($this->uuid)) {
             $this->getDao()->create();
         }
 
@@ -225,18 +225,6 @@ final class UUID extends Model\AbstractModel
         $self = new self;
 
         return $self->getDao()->getByUuid($uuid);
-    }
-
-    /**
-     * @param string $uuid
-     *
-     * @return bool
-     */
-    public static function uuidExists($uuid)
-    {
-        $self = new self;
-
-        return $self->getDao()->exists($uuid);
     }
 
     /**
--- a/models/Tool/UUID.php
+++ b/models/Tool/UUID.php
@@ -156,7 +156,9 @@ final class UUID extends Model\AbstractModel
         $uuid = Uid::v5($namespace, $this->getInstanceIdentifier() . '~' . $this->getType() . '~' . $this->getItemId());
         $this->uuid = $uuid->toRfc4122();
 
-        $this->getDao()->save();
+        if (!$this->uuidExists($this->uuid)) {
+            $this->getDao()->create();
+        }
 
         return $this->uuid;
     }
@@ -223,6 +225,18 @@ final class UUID extends Model\AbstractModel
         $self = new self;
 
         return $self->getDao()->getByUuid($uuid);
+    }
+
+    /**
+     * @param string $uuid
+     *
+     * @return bool
+     */
+    public static function uuidExists($uuid)
+    {
+        $self = new self;
+
+        return $self->getDao()->exists($uuid);
     }
 
     /**
--- a/models/Tool/UUID/Dao.php
+++ b/models/Tool/UUID/Dao.php
@@ -28,19 +28,22 @@ class Dao extends Model\Dao\AbstractDao
 
     public function save()
     {
-        $data = $this->model->getObjectVars();
-
-        foreach ($data as $key => $value) {
-            if (!in_array($key, $this->getValidTableColumns(static::TABLE_NAME))) {
-                unset($data[$key]);
-            }
-        }
+        $data = $this->getValidObjectVars();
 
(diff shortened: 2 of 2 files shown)


models/Tool/UUID/Dao.php
```diff
@@ -64,4 +77,16 @@ public function getByUuid($uuid)
...
     }
+
+    /**
+     * @param string $uuid
+     *
+     * @return bool
+     */
+    public function exists($uuid)
+    {
+        $result = $this->db->fetchOne('SELECT EXISTS(SELECT uuid FROM ' . self::TABLE_NAME . ' where uuid = ?)', [$uuid]);
```
brusch:
```suggestion
        $result = $this->db->fetchOne('SELECT uuid FROM ' . self::TABLE_NAME . ' where uuid = ?', [$uuid]);
```
I think we can omit the `EXISTS()` here, as it makes no difference, right?

models/Tool/UUID.php
```diff
@@ -225,6 +227,18 @@ public static function getByUuid($uuid)
...
+     *
+     * @return bool
+     */
+    public static function uuidExists($uuid)
+    {
+        $self = new self;
+
+        return $self->getDao()->exists($uuid);
+    }
+
```
brusch:
```suggestion
```
We actually don't need that, we can just straight call `$this->getDao()->exists($uuid)` in `createUuid()`.

models/Tool/UUID.php
```diff
@@ -156,7 +156,9 @@ public function createUuid()
         $uuid = Uid::v5($namespace, $this->getInstanceIdentifier() . '~' . $this->getType() . '~' . $this->getItemId());
         $this->uuid = $uuid->toRfc4122();
 
-        $this->getDao()->save();
+        if (!$this->uuidExists($this->uuid)) {
```
brusch:
```suggestion
        if (!$this->getDao()->exists($this->uuid)) {
```

models/Tool/UUID/Dao.php
```diff
@@ -39,6 +39,19 @@ public function save()
...
 
+    public function create()
+    {
+        $data = $this->model->getObjectVars();
+
+        foreach ($data as $key => $value) {
+            if (!in_array($key, $this->getValidTableColumns(static::TABLE_NAME))) {
+                unset($data[$key]);
+            }
+        }
```
brusch: Maybe we can put that into a `private` helper function to avoid duplicate code.

brusch (changes requested):

models/Tool/UUID/Dao.php
```diff
@@ -64,4 +77,16 @@ public function getByUuid($uuid)
...
+    /**
+     * @param string $uuid
+     *
+     * @return bool
+     */
+    public function exists($uuid)
+    {
+        $result = $this->db->fetchOne('SELECT uuid FROM ' . self::TABLE_NAME . ' where uuid = ?', [$uuid]);
+
+        return (bool)$result;
```
brusch:
```suggestion
        return (bool) $this->db->fetchOne('SELECT uuid FROM ' . self::TABLE_NAME . ' where uuid = ?', [$uuid]);
```

brusch (changes requested):

models/Tool/UUID/Dao.php
brusch: just another small micro-optimization 😊

models/Tool/UUID/Dao.php
hstjep: Agree, added helper function.

brusch: @hstjep so that's the only remaining thing 😊 Thanks in advance!

brusch:
LGTM, thank you very much! 👌

PR #11960, opened 2022-04-25
Fix escaping quotes in dataobject keys for tree-childs-by-id

### LINKED ISSUE #11943: [Bug]: Failing permission check with double quotes in object-name
### Expected behavior

Show every allowed object in list view if the permission is given. A double quote should be escaped in the permission check.

### Actual behavior

If the user is not an admin a permission check is done in the DataObjectController (Line 175, `EXISTS(SELECT list FROM users_workspaces_object uwo WHERE userId IN (...`). If the `objects.o_path` or `objects.o_key` contains a double quote this results in a SQL error.
E.g.: 

```
An exception occurred while executing 'SELECT `list` FROM users_workspaces_object WHERE LOCATE(cpath,"/News Corporate Website/2022-04-07-Circular-Globe "text"")=1 AND userId IN (3,12) ORDER BY LENGTH(cpath) DESC, FIELD(userId, 12) DESC, `list` DESC LIMIT 1': SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '"")=1 AND userId IN (3,12) ORDER BY LENGTH(cpath) DESC, FIELD...' at line 1

```

### Steps to reproduce

Pimcore: 10.3.5
User with no admin rights

Create or rename an object with a name containing a double quote `"`, e.g. `Test with "quotes"`. Save the object. Close the parent folder. Re-open the parent folder.

It is important to close the folder, a refresh will not trigger the error.

tecbird:
Resolves problem if data object contains at leat one quote, sql command is ended

## Additional info  
image

### DIFF
--- a/models/Element/Dao.php
+++ b/models/Element/Dao.php
@@ -16,6 +16,7 @@
 namespace Pimcore\Model\Element;
 
 use Pimcore\Model;
+use function addslashes;
 
 /**
  * @internal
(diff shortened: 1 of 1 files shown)


dvesh3:
follow up to https://github.com/pimcore/pimcore/pull/11714

kingjia90:
Possibly already covered in #11946

fashxp:
Thx for your contribution!

Closing in favor of https://github.com/pimcore/pimcore/pull/11946

PR #11962, opened 2022-04-25
[PhpStan] setLocked expected string|null, got bool

blankse:


dvesh3 (approved):

dvesh3:
@blankse thanks!

PR #11965, opened 2022-04-26
Fix new error with PhpStan 1.6.0

blankse:
Fix a new error with new phpstan version 1.6.0
https://github.com/phpstan/phpstan/releases/tag/1.6.0


dvesh3:
@blankse thanks a lot!

PR #11966, opened 2022-04-26
Fix debug log message

jdreesen:
This is a followup of #11948 where the log message was moved outside the `if` and is now more general.


mcop1:
Thanks :)

PR #11967, opened 2022-04-26
Document::setChildren() should cache empty array

blankse:


dvesh3:
@blankse thanks!

PR #11410, opened 2022-02-09
[QuantityValue] Add convenience method for quantity value unit conversion

BlackbitDevs:
Currently you can [convert `QuantityValue` objects to another unit](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Data_Types/Number_Types.html#page_Quantity-Value-Unit-Conversion) with
```php
$originalValue = new QuantityValue(3, Unit::getByAbbreviation('m')->getId());
$converter = $this->container->get(\Pimcore\Model\DataObject\QuantityValue\UnitConversionService::class);
$convertedValue = $converter->convert($originalValue, Unit::getByAbbreviation('mm'));
// $convertedValue is a QuantityValue with value 3000 and unit mm
```

This PR adds a shorter convenience method to do conversion like this:
```php
$originalValue = new QuantityValue(3, Unit::getByAbbreviation('m')->getId());
$convertedValue = $originalValue->convertTo(Unit::getByAbbreviation('mm'));
// $convertedValue is a QuantityValue with value 3000 and unit mm
```

and it even supports
```php
$originalValue = new QuantityValue(3, Unit::getByAbbreviation('m')->getId());
$convertedValue = $originalValue->convertTo('mm');
// $convertedValue is a QuantityValue with value 3000 and unit mm
```

Additionally I changed `UnitConversionService::convert()` to also support `InputQuantityValue` objects. With a [unit conversion service](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Data_Types/Number_Types.html#page_Dynamic-unit-conversion) it might be possible that those can also be converted to another unit.

PR #11968, opened 2022-04-26
[PhpStan] convertTo Method not possible for InputQuantityValue

blankse:

Followup to https://github.com/pimcore/pimcore/pull/11410

The [QuantityValueConverterInterface](https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/QuantityValue/QuantityValueConverterInterface.php) allow only QuantityValue

### DIFF
--- a/models/DataObject/Data/AbstractQuantityValue.php
+++ b/models/DataObject/Data/AbstractQuantityValue.php
@@ -15,9 +15,11 @@
 
 namespace Pimcore\Model\DataObject\Data;
 
+use InvalidArgumentException;
 use Pimcore;
 use Pimcore\Model\DataObject\OwnerAwareFieldInterface;
 use Pimcore\Model\DataObject\QuantityValue\Unit;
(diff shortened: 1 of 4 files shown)


dvesh3 (approved):

dvesh3:
I think convert should work for `inputQuantityValue` as well but it would not be possible to support it without changing `QuantityValueConverterInterface::convert`, which is a BC break. I'll create an issue to support this on Pimcore 11.

dvesh3 (commented):
@blankse how about we throw exception from `InputQuantityValue::convertTo` instead of simply kicking it?

blankse:
@dvesh3 We can throw an "Not supported now" exception in UnitConversionService::convert(), so the param type can be hold there.

blankse:
@dvesh3 I changed it.

dvesh3 (approved):

dvesh3:
@blankse thanks!

PR #11971, opened 2022-04-26
Custom Reports: Send drilldownFilters to createCSVAction

AlexanderHeidrich:
Currently the `\Pimcore\Bundle\AdminBundle\Controller\Reports\CustomReportController::createCsvAction()` allows to provide
* `sort`
* `dir`
* `filter`
* `drillDownFilters`
* `headers`
* `offset`
* `exportFile`

While most of them are sent, `sort`, `dir` and `drillDownFilters` aren't.
The result will be an unfiltered list or at least a list which is not filtered by custom filters.

This PR allows at least the usage of `drillDownFilters` which are available in the JS-methods scope.


aryaantony92:
@AlexanderHeidrich Thanks

PR #11972, opened 2022-04-26
[Bug]: Index Field Selection - last change in the current dev branch introduced backward compatibility break

### LINKED ISSUE #11941: [Bug]: Index Field Selection - last change in the current dev branch introduced backward compatibility break
### Expected behavior

Follow up to #11861.

In the PR #11895, the NormalizerInterface is implemented for IndexFieldSelections.

I used an array to normalize the relevant properties from the data object. With the help of a backup from before a PCX update, I noticed that the `ObjectData\IndexFieldSelection` object was used before PCX for serialization (which is not a best practice I suppose).

So instead the ` DataObject\Traits\SimpleNormalizerTrait` could have been used instead.

We now have two options:
1. Either, keep it in an array (current state of the code) and break backward compatibility further (which was already broken since PCX) but do not use a dependency to a data object in the serialization OR
2. Implement NormalizationInterface with SimpleNormalizerTrait instead and keep it serialized in the data object. I can provide a new PR for that.

**What would you say?**

### Actual behavior

Serialized Index Field Selection from before PCX can not be loaded correctly.

### Steps to reproduce

Apply a serialized DB entry from before a PCX update to the current dev branch and it will not be loaded correctly.

mcop1:
Resolves #11941

### DIFF
--- a/bundles/EcommerceFrameworkBundle/CoreExtensions/ClassDefinition/IndexFieldSelection.php
+++ b/bundles/EcommerceFrameworkBundle/CoreExtensions/ClassDefinition/IndexFieldSelection.php
@@ -351,6 +351,9 @@ class IndexFieldSelection extends Data implements ResourcePersistenceAwareInterf
 
             return new ObjectData\IndexFieldSelection($tenant, $field, $preSelect);
         }
+        elseif ($value instanceof ObjectData\IndexFieldSelection) {
+            return $value;
+        }
 
         return null;
     }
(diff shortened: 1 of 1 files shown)


bundles/EcommerceFrameworkBundle/CoreExtensions/ClassDefinition/IndexFieldSelection.php
```diff
@@ -351,6 +351,9 @@ public function denormalize($value, $params = [])
 
             return new ObjectData\IndexFieldSelection($tenant, $field, $preSelect);
         }
+        elseif ($value instanceof ObjectData\IndexFieldSelection) {
```
jdreesen: The `else` is not necessary, as the `if` above always returns.
```suggestion
        if ($value instanceof ObjectData\IndexFieldSelection) {
```

PR #11980, opened 2022-04-27
[PhpStan] Fix AdminExceptionListener:recursiveAddValidationExceptionSubItems()

blankse:
Split from #11826 
```
------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   bundles/AdminBundle/EventListener/AdminExceptionListener.php                                                                                                                                                                
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  85     Parameter #1 $items of method Pimcore\Bundle\AdminBundle\EventListener\AdminExceptionListener::recursiveAddValidationExceptionSubItems() expects array<Pimcore\Model\Element\ValidationException>, array<Exception> given.  
  142    Parameter #1 $items of method Pimcore\Bundle\AdminBundle\EventListener\AdminExceptionListener::recursiveAddValidationExceptionSubItems() expects array<Pimcore\Model\Element\ValidationException>, array<Exception> given.  
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

### DIFF
--- a/bundles/AdminBundle/EventListener/AdminExceptionListener.php
+++ b/bundles/AdminBundle/EventListener/AdminExceptionListener.php
@@ -131,7 +131,9 @@ class AdminExceptionListener implements EventSubscriberInterface
         foreach ($items as $e) {
             if ($e->getMessage()) {
                 $message .= '<b>' . $e->getMessage() . '</b>';
-                $this->addContext($e, $message);
+                if ($e instanceof ValidationException) {
+                    $this->addContext($e, $message);
+                }
                 $message .= '<br>';
 
                 $detailedInfo .= '<br><b>Message:</b><br>';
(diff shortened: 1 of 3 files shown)


dvesh3 (commented):
@blankse why not just change the `AdminExceptionListener::recursiveAddValidationExceptionSubItems` phpdoc to accept `\Exception[]`?

I'm afraid if the `ValidationException` class is being used for aggregating `\Exception` then it will break. Also, please note this class is not internal.

blankse:
@dvesh3 Exception have no method getSubItems. So the recursiveAddValidationExceptionSubItems method is broken. So it was not possible to add Exceptions here I think. But we can check it in the recursiveAddValidationExceptionSubItems method if it is a ValidationException and only call it for this type, if you want.

dvesh3:
>  But we can check it in the recursiveAddValidationExceptionSubItems method if it is a ValidationException and only call it for this type, if you want.

yes, please give it a try. thanks!

blankse:
@dvesh3 I changed it.

dvesh3:
@blankse thanks!

PR #11981, opened 2022-04-27
FIX: Consider Listing Offset in Pagination

eliasfellner:
If Listing has an Offset, add it to the calculated Pagination Offset

### DIFF
--- a/lib/Model/Paginator/EventSubscriber/PaginateListingSubscriber.php
+++ b/lib/Model/Paginator/EventSubscriber/PaginateListingSubscriber.php
@@ -26,7 +26,7 @@ class PaginateListingSubscriber implements EventSubscriberInterface
         $paginationAdapter = $event->target;
 
         if ($paginationAdapter instanceof PaginateListingInterface) {
-            $items = $paginationAdapter->getItems($event->getOffset(), $event->getLimit());
+            $items = $paginationAdapter->getItems((int)$paginationAdapter->getOffset() + $event->getOffset(), $event->getLimit());
             $event->count = $paginationAdapter->count();
             $event->items = $items;
             $event->stopPropagation();


dvesh3 (commented):
offset is calculated and passed by Knp `$offset = ($page - 1) * $limit;` based on the page number passed to the `paginate()`. 

what is your use case?

eliasfellner:
My use case is an Objects Overview Page, where we have 3 special teasers at the start + 9 teasers with a specific styling
In Page 1 we want to load 12 Items, for page 2 or higher we want to load 9, therefore an offset of 3 has to be added to the paginator for page 2 or higher.

dvesh3:
@eliasfellner for your usecase, i would go for `getItems($offset, $itemCountPerPage)` call on listing rather than using knp paginator.

passioneight:
@dvesh3 wouldn't that imply that it is necessary to handle the pagination manually?

dvesh3:
@passioneight the use case mentioned is more about slicing the result-set than pagination. Or i got it wrong. So it is about the teasers showing on each page, for first page all 12 needs to be shown and for page 2 or higher only 9 to be shown...right?

passioneight:
@dvesh3 yes, that sounds about right. To put it in (pseudo-)code:

Instead of doing this:
```php
$listing->setOffset($offset);
$paginator = $paginatorInterface->paginate($listing, $page, $limit);
```

You'd do it like this:
```php
$listing->setOffset($offset);
$paginator = $paginatorInterface->paginate($listing->getItems($offset, null), $page, $limit); 
```

That's basically what you suggested, correct?

> I'm aware that you wouldn't use the paginator at all (as per your comment), but that seems a bit off when a pagination is needed, wouldn't you agree?

If that's correct, the latter example comes with the following drawbacks:

1. `$listing->setOffset($offset);` may still be needed for some code that is executed before passing the `$listing` to the paginator. Not really an issue, but seems a bit inconvenient to have to set `$offset` twice.
2. `$listing->getItems($offset, null)` is needed to load the objects and still have the the total count correct. However, this will load all objects starting from `$offset` which will eventually have bad performance.

If I misunderstood, please let me know - possibly in form of (pseudo-)code, to avoid misunderstandings :-)

dvesh3:
> That's basically what you suggested, correct?

@passioneight no, you got me wrong. I suggested to completely avoid using paginator for your case. 

so if you are using paginator to slice teaser items like this:
```
$listing->setOffset($offset);
$paginator = $paginatorInterface->paginate($listing, $page, $limit);

```
then i would suggest to do this:
```
$listing->setOffset($offset);
$items = $listing->load();
```

passioneight:
@dvesh3 I see, thanks for clarifying. Calling `load` instead of `setItems(...)` certainly resolves Nr. 1. Additionally, calling `setLimit` on the listing will then resolve Nr. 2.

So, I guess the result would be something like:

```php
$listing = new Example\Listing();

if($page === 1) {
   $listing->setLimit(12);
} else {
   $listing->setOffset(12);
   $lisinst->setLimit(9);
}
```

This PR can probably be closed then (unless @eliasfellner has some more input on this?). But I still wouldn't mind seeing support for the offset in the listing directly in the adapter as it was unexpected behaviour.

dvesh3:
Like i said before, when using paginator the offset is calculated based on page number from Knp listener. 

I'm closing it for now. I would be happy to reopen and merge this PR, if you have a valid use-case. thanks!

PR #11983, opened 2022-04-27
Fixed Deprecated error message in admin login

aryaantony92:

Fixed the deprecated error message with strip_tags() function, when admin does the login/logout from the admin interface.

### DIFF
--- a/bundles/AdminBundle/Security/Guard/AdminAuthenticator.php
+++ b/bundles/AdminBundle/Security/Guard/AdminAuthenticator.php
@@ -134,10 +134,7 @@ class AdminAuthenticator extends AbstractGuardAuthenticator implements LoggerAwa
             return $response;
         }
 
-        $perspective = $request->get('perspective');
-        $perspective = ($perspective !== null)? strip_tags($perspective): [];
-
-        $event = new LoginRedirectEvent('pimcore_admin_login', ['perspective' => $perspective ]);
+        $event = new LoginRedirectEvent('pimcore_admin_login', ['perspective' => strip_tags($request->get('perspective', ''))]);
         $this->dispatcher->dispatch($event, AdminEvents::LOGIN_REDIRECT);
 
         $url = $this->router->generate($event->getRouteName(), $event->getRouteParams());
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Security/Guard/AdminAuthenticator.php
```diff
@@ -134,7 +134,10 @@ public function start(Request $request, AuthenticationException $authException =
             return $response;
         }
 
-        $event = new LoginRedirectEvent('pimcore_admin_login', ['perspective' => strip_tags($request->get('perspective'))]);
+        $perspective = $request->get('perspective');
+        $perspective = ($perspective !== null)? strip_tags($perspective): [];
```
jdreesen: `strip_tags()` returns a `string`, so why is the other case an `array`?

bundles/AdminBundle/Security/Guard/AdminAuthenticator.php
```diff
@@ -134,7 +134,10 @@ public function start(Request $request, AuthenticationException $authException =
...
+
+        $event = new LoginRedirectEvent('pimcore_admin_login', ['perspective' => $perspective ]);
```
jdreesen: Shouldn't this be enough?

```suggestion
        $event = new LoginRedirectEvent('pimcore_admin_login', ['perspective' => strip_tags($request->get('perspective', ''))]);
```

blankse:
```suggestion
        $event = new LoginRedirectEvent('pimcore_admin_login', ['perspective' => strip_tags($request->get('perspective', ''))]);
```
This should be enough, right?

dvesh3:
```suggestion

        $event = new LoginRedirectEvent('pimcore_admin_login', ['perspective' => is_string($perspective) ? strip_tags($perspective) : $perspective]);
```

bundles/AdminBundle/Security/Guard/AdminAuthenticator.php
dvesh3: yes, using default string should solve it.

PR #11988, opened 2022-04-27
[Documents] Not possible to save documents

### LINKED ISSUE #11987: [Documents] Not possible to save documents
### Expected behavior

Documents should be saved

### Actual behavior

throwing exception Doctrine\DBAL\Connection::fetchOne(): Argument #2 ($params) must be of type array, null given

### Steps to reproduce

1. Open Document
2. Click save / save & publish.

blankse:
Fix https://github.com/pimcore/pimcore/issues/11987


dvesh3 (approved):
we can still merge this as just another way of checking the type and it will fix if you are passing wrong type the condition params.

PR #11416, opened 2022-02-10
Add filter for filename options to thumbnail commands

jremmurd:
Add options to filter by filename to commands:

- pimcore:image:low-quality-preview
- pimcore:thumbnails:image

**Use case:**
Pre-generate thumbnails for a productgrid while all product images have the same naming structure and the needed images all end with the same postfix.

PR #11995, opened 2022-04-28
[Asset] Thumbnails: exclude assets by paths from pimcore:thumbnails:image command

patkul0:
In Image Thumbnails configuration panel there is a collapsible Advanced Fieldset to which we added new GridPanel that allows user to drag&drop folders or images from tree. Thumbnail for assets to which paths were assign to that field will be NOT generated in background with pimcore:thumbnails:image command.

## Additional info  
We faced with a situation where client didn't want to generate specific thumbnail configuration for couple paths because of some reasons (ex. they use for product photos thumbnails with FRAME transformation and not with CONTAIN which on the other hand is perfect for dimensions picture). After that PR admin can limit that directly in Pimcore interface.


brusch:
Have you already seen #11416 ? That should cover your requirement as well. 
If not, we should extend the existing parameter solution and keep the thumbnail definition itself independent from this command. 
Looking forward to your feedback 🤗

PR #11999, opened 2022-04-28
Fix for QuickSearch regression for user accounts without cid

### LINKED ISSUE #11998: [Bug]: 10.3.6 QuickSearch exception for non-admin users
### Expected behavior

QuickSearch should work for all users.

### Actual behavior

Left side: Admin user
Right side: User with limited permissions
image

```trace
Timestamp: Thu Apr 28 2022 16:39:16 GMT+0200 (Central European Summer Time)
URL: /admin/search/search/quicksearch?_dc=1651156756409
Method: GET
Message: An exception occurred while executing '
            SELECT
                cpath, userid, max(list) as list
            FROM users_workspaces_asset
            WHERE userId IN (4,15) AND cid NOT IN ()
            GROUP BY cpath
        ':

SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')
            GROUP BY cpath' at line 4
Trace: 
in /var/www/pimcore/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php:98
#0 /var/www/pimcore/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php(182): Doctrine\DBAL\Driver\AbstractMySQLDriver->convertException('An exception oc...', Object(Doctrine\DBAL\Driver\PDO\Exception))
#1 /var/www/pimcore/vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php(159): Doctrine\DBAL\DBALException::wrapException(Object(Doctrine\DBAL\Driver\PDO\MySQL\Driver), Object(Doctrine\DBAL\Driver\PDO\Exception), 'An exception oc...')
#2 /var/www/pimcore/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(2226): Doctrine\DBAL\DBALException::driverExceptionDuringQuery(Object(Doctrine\DBAL\Driver\PDO\MySQL\Driver), Object(Doctrine\DBAL\Driver\PDO\Exception), '\n            SE...', Array)
#3 /var/www/pimcore/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(1313): Doctrine\DBAL\Connection->handleExceptionDuringQuery(Object(Doctrine\DBAL\Driver\PDO\Exception), '\n            SE...', Array, Array)
#4 /var/www/pimcore/vendor/pimcore/pimcore/lib/Db/PimcoreExtensionsTrait.php(99): Doctrine\DBAL\Connection->executeQuery('\n            SE...', Array, Array, NULL)
#5 /var/www/pimcore/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(1000): Pimcore\Db\Connection->executeQuery('\n            SE...', Array, Array)
#6 /var/www/pimcore/vendor/pimcore/pimcore/models/Element/Service.php(751): Doctrine\DBAL\Connection->fetchAll('\n            SE...')
#7 /var/www/pimcore/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Searchadmin/SearchController.php(367): Pimcore\Model\Element\Service::findForbiddenPaths('asset', Object(Pimcore\Model\User))
#8 /var/www/pimcore/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Searchadmin/SearchController.php(443): Pimcore\Bundle\AdminBundle\Controller\Searchadmin\SearchController->getForbiddenCondition()
#9 /var/www/pimcore/vendor/symfony/http-kernel/HttpKernel.php(152): Pimcore\Bundle\AdminBundle\Controller\Searchadmin\SearchController->quicksearchAction(Object(Symfony\Component\HttpFoundation\Request), Object(Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher))
#10 /var/www/pimcore/vendor/symfony/http-kernel/HttpKernel.php(74): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#11 /var/www/pimcore/vendor/symfony/http-kernel/Kernel.php(202): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#12 /var/www/pimcore/public/index.php(35): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#13 {main}
```

### Steps to reproduce

1. Search for something as admin
2. Log in as a different user
3. Search for anything as the limited user.

(demo.pimcore.fun is currently unavailable due to "maintenance downtime or capacity problems" so can not attempt to reproduce there)

NiklasBr:
Resolves #11998

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -725,28 +725,18 @@ class Service extends Model\AbstractModel
             return ['forbidden' => [], 'allowed' => []];
         }
 
-        $currentUserId = $user->getId();
-        $userIds = $user->getRoles();
-        $userIds[] = $currentUserId;
-
-        $userWorkspacesSql = '
-            SELECT cpath, cid, list
-            FROM users_workspaces_'.$type.'
-            WHERE userId = '.$currentUserId.'
-        ';
-        $userWorkspaces = $db->fetchAll($userWorkspacesSql);
-
+        $userWorkspaces = $db->fetchAll('SELECT cpath, cid, list FROM users_workspaces_'.$type.' WHERE userId = ?', [$user->getId()]);
         if (!empty($userWorkspaces)) {
             //this collects the array that are on user-level, which have top priority
-            $userCid = [];
+            $workspaceCids = [];
             foreach ($userWorkspaces as $userWorkspace) {
-                $userCid[] = $userWorkspace['cid'];
+                $workspaceCids[] = $userWorkspace['cid'];
             }
 
             $roleWorkspacesSql = 'SELECT
                     cpath, userid, max(list) as list
                 FROM users_workspaces_'.$type.'
-                WHERE userId IN (' . implode(',', $userIds) . ') AND cid NOT IN (' . implode(',', $userCid) . ')
+                WHERE userId IN (' . implode(',', $user->getRoles()) . ') AND cid NOT IN (' . implode(',', $workspaceCids) . ')
                 GROUP BY cpath';
 
             $roleWorkspaces = $db->fetchAll($roleWorkspacesSql);
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -742,13 +742,17 @@ class Service extends Model\AbstractModel
             $userCid[] = $userWorkspace['cid'];
         }
 
-        $roleWorkspacesSql = '
-            SELECT
+        $roleWorkspacesSql = 'SELECT
                 cpath, userid, max(list) as list
             FROM users_workspaces_'.$type.'
-            WHERE userId IN (' . implode(',', $userIds) . ') AND cid NOT IN ('.implode(',', $userCid).')
-            GROUP BY cpath
-        ';
+            WHERE userId IN (' . implode(',', $userIds) . ')';
+
+        if (!empty($userCid)) {
+            $roleWorkspacesSql .= ' AND cid NOT IN ('.implode(',', $userCid).')';
+        }
+
+        $roleWorkspacesSql .= ' GROUP BY cpath';
+
         $roleWorkspaces = $db->fetchAll($roleWorkspacesSql);
 
         $allWorkspaces = array_merge($userWorkspaces, $roleWorkspaces);
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -716,7 +716,7 @@ class Service extends Model\AbstractModel
      * @param string $type asset|object|document
      * @param Model\User $user
      *
-     * @return array
+     * @return array{forbidden: array, allowed: array}
      */
     public static function findForbiddenPaths($type, $user)
     {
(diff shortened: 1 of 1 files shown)


NiklasBr:
Static analysis failing on `lib/Helper/Mail.php` and not these changes

blankse:
@NiklasBr I created an PR for the static analysis errors: https://github.com/pimcore/pimcore/pull/12000

kingjia90:
Thank you @NiklasBr and @blankse !

@NiklasBr Now i have also noticed that it should skip `$roleWorkspacesSql` when the user doesn't have any role assigned (this query would be obsolete), could you also make a commit about this? Thank you in advance

brusch:
@NiklasBr #12000 got merged, would be nice if you could merge the latest `10.4` into your branch as well, then the tests can run through 😊

models/Element/Service.php
```diff
@@ -742,13 +742,17 @@ public static function findForbiddenPaths($type, $user)
...
-            SELECT
+        $roleWorkspacesSql = 'SELECT
                 cpath, userid, max(list) as list
             FROM users_workspaces_'.$type.'
-            WHERE userId IN (' . implode(',', $userIds) . ') AND cid NOT IN ('.implode(',', $userCid).')
-            GROUP BY cpath
-        ';
+            WHERE userId IN (' . implode(',', $userIds) . ')';
+
+        if (!empty($userCid)) {
```
blankse: The variable always exists. So this should be enough.
```suggestion
        if ($userCid) {
```

NiklasBr: I know what you mean, but I think this is slightly more readable as it is more explicit (and more in line with JavaScript and Python where `[]` is truthy).

blankse: I would avoid empty when its possible, empty is slower than boolean cast.
https://www.contextualcode.com/Blog/php-micro-optimization.-variable-boolean-cast-vs-!empty
But it doesn't matter yes. Better avoid the sql if no role is set :)

NiklasBr:
@kingjia90 did I understand your comment correctly here?

kingjia90:
@NiklasBr no, i am sorry, i wasn't clear. 

It should be more clearer when removing `$userIds[] = $currentUserId;` on line 730 (which is wrong, as those workspaces are already considered in the first query and excluded in the second one by `cid NOT IN`), and eventually rename `$userIds` to `$roleIds`. 
Now, for an user without any roles, it would use an IN condition on a imploded empty array like the first bug.

Line 730 should be removed, because it's unnecessary, no matter if roles are assigned, the query would look like `SELECT FROM workspaces WHERE userId IN (currentUser) AND cid NOT IN (cids of workspaces where userId = currentUser)`, so always empty.

NiklasBr:
I think I got it now :)

kingjia90:
@NiklasBr yes, but  `$roleWorkspacesSql` needs to be "independent", so outside the block
`if (!empty($userWorkspaces)) {`, and in a separate `if ($user->getRoles())`;  while the condition `cid NOT IN (' . implode(',', $workspaceCids) . ')` must be there/not there depending on the $workspaceCids, the rest of the query still need to be run to get the roles workspace (even if $userWorkspace is empty)

NiklasBr:
Wait.

So, if I understand how this is going to work there are multiple contexts:

1. A user account has no roles, but a workspace defined on itself.
2. A user account has one or more roles, with no workspaces defined.
3. A user account has one or more roles, with at least one defined workspace among those roles, but no workspace on itself.
4. A user account has one or more roles, with at least one defined workspace among those roles, and with a workspace on itself?

This seems like there should be logic available for this somewhere else because this is not the only place all these conditions are checked. Right?

kingjia90:
So at the moment, PR looks like:
```
$userWorkspaces
if (!empty($userWorkspaces)) {
        $roleWorkspacesSql
}
```
what is need is more like
```
$userWorkspaces
if (user has roles) {
        $roleWorkspacesSql
        if (found $userWorkspaces){ 
             add condition `cid NOT IN cids of $userWorkspaces` to roleWorkspacesSql
        }
}
```
1. this works, since there are no roles, the roleWorkspacesSql is never run
2. this would return nothing as no permitted workspace/element is found, it's covered in the else of `if ($totalPaths > 0) {`
3. userWorkspaces is empty, array_merge (former $allWorkspaces) would return basically just roleWorkspacesSql
4. again the array_merge, would just concatenate $userWorkspaces and $roleWorkspacesSql

maybe the missing thing is that the merge here is used ONLY to append the data, since there won't be any duplicates because of the `cid NOT IN $workspaceCids` in $roleWorkspacesSql

NiklasBr:
@kingjia90

PR #12000, opened 2022-04-28
Fix new PhpStan errors

blankse:
Fix new errors with PhpStan Version 1.6.3 https://github.com/phpstan/phpstan/releases/tag/1.6.3


brusch:
Thank you very much! 👌

Issue #12004, opened 2022-04-29
[Bug]: trim null in Document/Page 


ghost:
### Expected behavior

Service can perform Service::copyRecursive from Document/Page

### Actual behavior

Crash on perform Service::copyRecursive from Document/Page

Service.php#164
```
        if (method_exists($new, 'setPrettyUrl')) {
            $new->setPrettyUrl(null);
        }
```

Page.php#174
```
    /**
     * @param string $prettyUrl
     *
     * @return $this
     */
    public function setPrettyUrl($prettyUrl)
    {
        $this->prettyUrl = '/' . trim($prettyUrl, ' /');
        if (strlen($this->prettyUrl) < 2) {
            $this->prettyUrl = null;
        }

        return $this;
    }
```

and trim can not handle null values, also phpdoc say it can not receive null as prettyUrl.

### Steps to reproduce

We had this issue, by recreating database for testing szenarios.

Fix?: 
```
/**
 * @param string|null $prettyUrl
 *
 * @return $this
 */
public function setPrettyUrl($prettyUrl)
{
    if (null == $prettyUrl) {
        $this->prettyUrl = null;
        return $this;
    }
    $this->prettyUrl = '/' . trim($prettyUrl, ' /');
    if (strlen($this->prettyUrl) < 2) {
        $this->prettyUrl = null;
    }
    return $this;
}
```

ghost:
It looks like the Dao in Testszenario will also trigger this issue.

Affected
* prettyUrl
* targetGroupIds

```
#0 /var/www/html/vendor/pimcore/pimcore/lib/Model/AbstractModel.php(205): Pimcore\Model\Document\Page->setTargetGroupIds(NULL)\n
#1 /var/www/html/vendor/pimcore/pimcore/lib/Model/AbstractModel.php(187): Pimcore\Model\AbstractModel->setValue('targetGroupIds', NULL)\n
#2 /var/www/html/vendor/pimcore/pimcore/lib/Model/Dao/DaoTrait.php(45): Pimcore\Model\AbstractModel->setValues(Array)\n

```

blankse:
@galdp I created 2 PRs for this. Can you test it?

dvesh3:
Fixed by #12008 & #12009

ghost:
> @galdp I created 2 PRs for this. Can you test it?

@blankse Tests passing, thanks

PR #12008, opened 2022-05-02
Fix: Document/Page::setPrettyUrl should be nullable

blankse:
Partly fix https://github.com/pimcore/pimcore/issues/12004


dvesh3 (approved):

dvesh3:
@blankse thanks!

PR #12009, opened 2022-05-02
Fix: targetGroupIds Field in documents_page db table should not nullable

blankse:
Partly fix https://github.com/pimcore/pimcore/issues/12004


dvesh3 (approved):

dvesh3:
@blankse thanks!

PR #12011, opened 2022-05-02
SEO Document Editor - prevent of calling undefined methods

zerlpaMMT:
Prevent Exception of undefined method call -> /admin/document/seopanel-tree -> Call to undefined method .... in class Pimcore\Model\Document\Folder

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
@@ -1614,7 +1614,7 @@ class DocumentController extends ElementControllerBase implements KernelControll
     }
 
     /**
-     * @param Document\Page $document
+     * @param Document $document
      *
      * @return array
      */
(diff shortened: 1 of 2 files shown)


blankse:
@zerlpaMMT Can you change the param phpdoc to only Document and can you use "instanceof Page" instead of method_exists?

blankse:
@zerlpaMMT You fixed a ignored phpstan error. You can remove it from the baseline file now:
https://github.com/pimcore/pimcore/blob/b35761e749f42000170b778ef8c127ae2735d266/phpstan-baseline.neon#L8-L12

I think we don't need to set the array keys to an empty string if it isn't a Page. You can simply set the instanceof if around the existing code.

blankse (approved):
@zerlpaMMT Thanks for the changes :)

robertSt7:
Thank you very much for the PR :)

PR #12012, opened 2022-05-02
Fix: Areablock in Block editable works only after reloading

### LINKED ISSUE #11285: [Bug]: Nested Area Toolbar missing
### Expected behavior

If I have a Block Editable in a Document which contains an Areablock
I expect to be able to add a new Block Item and immediately add an Area in the Areablock, without the Area Toolbar missing

### Actual behavior

When adding a new Block Item which contains an Areablock, and immediately adding an Area
The Area is rendered but not the Toolbar.
You have to reload the Document to see the Toolbar (which is not good when using EditableDialogBoxConfiguration)

### Steps to reproduce

In a Document have this structure:
Block -> Areablock

1. Add a new Block Item
2. Add an Area in the Areablock
3. Check if the Area Toolbar is there

Pimcore Version: 10.2.9

blankse:
Fix https://github.com/pimcore/pimcore/issues/11285


dvesh3:
@blankse thanks!

Issue #11800, opened 2022-03-29
[Bug]: obsolete Presta / Sitemap bundle generates deprecation notices on PHP 8.1


dkarlovi:
### Expected behavior

No deprecations regarding Presta Sitemap Bundle.

### Actual behavior

```
Deprecated: Return type of Presta\SitemapBundle\Sitemap\XmlConstraint::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
```

### Steps to reproduce

Run Pimcore 10.3.3 on PHP 8.1

This issue seems to have been fixed in SitemapBundle ^3, Pimcore requires ^2.1

dvesh3:
@dkarlovi please see https://github.com/pimcore/pimcore/issues/11661

brusch:
Closed in favor of #11661

PR #12020, opened 2022-05-03
[Task:] Update presta/sitemap-bundle to version 3

### LINKED ISSUE #11661: Update presta/sitemap-bundle to version 3
### Improvement description

Version 3 Supports Symfony 4,5,6 so can be already supported on Pimcore X.

kingjia90:
Resolves #11661

in 3.0, only the folder structure has changed, the rest seems working
https://github.com/prestaconcept/PrestaSitemapBundle/releases/tag/v3.0.0

please see upgrade notes or Demo and related [PR](https://github.com/pimcore/demo/pull/313/files)

### DIFF
--- a/doc/Development_Documentation/18_Tools_and_Features/39_Sitemaps.md
+++ b/doc/Development_Documentation/18_Tools_and_Features/39_Sitemaps.md
@@ -44,7 +44,7 @@ For details see:
 
 * [Bundle Documentation](https://github.com/prestaconcept/PrestaSitemapBundle/blob/3.x/doc/2-configuration.md#configuring-your-application-base-url)
 * [Symfony Documentation on the Request Context](https://symfony.com/doc/5.2/routing.html#generating-urls-in-commands)
-* [`UrlGenerator`](https://github.com/pimcore/pimcore/blob/10.x/lib/Sitemap/UrlGenerator.php)
+* [`UrlGenerator`](https://github.com/pimcore/pimcore/blob/11.x/lib/Sitemap/UrlGenerator.php)
 
 
 ## Sitemap Generators
@@ -98,8 +98,8 @@ pimcore:
 For more advanced use cases involving Pimcore models, Pimcore defines an `AbstractElementGenerator` which is extendable
 via pluggable filters and processors. This makes it possible to define reusable behaviour in a filter/processor which can
 be used from multiple generators. A **filter** determines if an element can be added to the sitemap and if it is able to handle children (it's up to the
-generator to query for this information). For example the [PropertiesFilter](https://github.com/pimcore/pimcore/blob/10.x/lib/Sitemap/Element/Filter/PropertiesFilter.php)
-excludes elements with a property `sitemaps_exclude`. A **processor** can enhance an entry before it is added to the container. For example, the [ModificationDateProcessor](https://github.com/pimcore/pimcore/blob/10.x/lib/Sitemap/Element/Processor/ModificationDateProcessor.php)
+generator to query for this information). For example the [PropertiesFilter](https://github.com/pimcore/pimcore/blob/11.x/lib/Sitemap/Element/Filter/PropertiesFilter.php)
+excludes elements with a property `sitemaps_exclude`. A **processor** can enhance an entry before it is added to the container. For example, the [ModificationDateProcessor](https://github.com/pimcore/pimcore/blob/11.x/lib/Sitemap/Element/Processor/ModificationDateProcessor.php)
 adds the modification date of an element to the url.
 
 Which filters and processors to use can be defined on the generator level. For example, the [`DocumentTreeGenerator`](#page_The-DocumentTreeGenerator)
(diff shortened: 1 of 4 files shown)


jdreesen:
This needs an entry in the upgrade-docs, no?

jdreesen:
I'm just wondering if it would be possible to support both, `v2` and `v3`, leaving the decision when to upgrade to the users. It wouldn't be a BC break then.

kingjia90:
> I'm just wondering if it would be possible to support both, `v2` and `v3`, leaving the decision when to upgrade to the users. It wouldn't be a BC break then.

it has some problems with PHP8.1, more details here https://github.com/pimcore/pimcore/issues/11800 and it's required if we want to update to symfony 6

jdreesen:
But do you want to force-upgrade to Symfony 6 or do you just want to *allow* Symfony 6? If it's the latter, it would be fine to just allow `presta/sitemap-bundle v3`, too, and anyone who wants to use PHP 8.1 or Symfony 6 can upgrade to `presta/sitemap-bundle v3` themselves.

doc/Development_Documentation/18_Tools_and_Features/39_Sitemaps.md
```diff
@@ -11,12 +11,12 @@ an event handler and start to add entries.
...
 of your site (e.g. the size of the tree which needs to be processed). In general it's recommended to create static files
 as it reduces the overhead of creating the sitemap on every crawler request. If you want to serve the sitemap directly,
-you need to register the sitemaps routes in your routing config (see [PrestaSitemapBundle Documentation](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Resources/doc/1-installation.md)
+you need to register the sitemaps routes in your routing config (see [PrestaSitemapBundle Documentation](https://github.com/prestaconcept/PrestaSitemapBundle/blob/3.x/doc/1-installation.md)
 for details).
 
 ```yaml
 PrestaSitemapBundle:
-    resource: "@PrestaSitemapBundle/Resources/config/routing.yaml"
+    resource: "@PrestaSitemapBundle/config/routing.yaml"
```
NiklasBr, line 19: Also mentioned in https://github.com/pimcore/pimcore/blob/10.x/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md?

fashxp:
> But do you want to force-upgrade to Symfony 6 or do you just want to _allow_ Symfony 6?

Pimcore 11 will most likely require Symfony 6

jdreesen:
> Pimcore 11 will most likely require Pimcore 6

That's fine, as it is a new major version which allows BC breaks. But this PR is currently assigned to the `10.5` milestone, so a graceful update should be possible.

fashxp:
if BC break is needed, it will not go into 10.5 :-) 
we'll have a look...

jdreesen:
If the only BC break is the new folder structure, it should be fine to support both and let the users update as desired. If this is postponed to Pimcore v11, users will not be able to fix the deprecation notices it throws on PHP 8.

brusch:
@jdreesen but this would still need manual work before upgrading, because by using `"presta/sitemap-bundle": "^2.1 || ^3.0"` Composer would automatically install the latest release of v3.

jdreesen:
That's true (if the user doesn't have a `"presta/sitemap-bundle": "^2.1"` in their `composer.json`, which they probably haven't).

brusch:
@kingjia90 a possible solution I can think of is to override the config using this function: 
https://github.com/pimcore/pimcore/blob/6706dba38446ab23ca2ea1a704cae261eac45fbf/bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php#L448

We did this in the past and there's still an exampe in the code how to do that. 
So what I'm thinking of is to check there whether we are on v3 and if the config path is still `@PrestaSitemapBundle/Resources/config/routing.yml` and if so, we just rewrite the config to the new path. 

WDYT? Should we try that approach? Any concerns?

kingjia90:
@brusch that would be a good solution, i'll give it a try and update you

brusch:
@kingjia90 btw. even if that works, we should imho still allow v2, by using `"presta/sitemap-bundle": "^2.1 || ^3.0"`

brusch:
Unfortunately my approach described above using `prepend()` method, doesn't work, as `resource` loading get's evaluated right on loading the config and throws an exception if the linked resource doesn't exist.

So I've another approach using a custom route loader: 
https://symfony.com/doc/current/routing/custom_route_loader.html#creating-a-custom-loader

We could implement another custom route loader, that has the following in the `supports()` method: 
```php
public function supports($resource, string $type = null)
{
    if ($resource === '@PrestaSitemapBundle/Resources/config/routing.yml') {
        try {
            $loader->getLocator()->locate('@PrestaSitemapBundle/config/routing.yml');
            return true;
        } catch (FileLocatorFileNotFoundException $e) {
            return false;
        }
    }
}
```

And in the `load()` method it just rewrites the resource path and then delegates to the existing `YamlFileLoader`. 

That should actually work 😊

kingjia90:
@brusch I've give it a try, unfortunately it doesn't "detect" the Resource and breaks
image

but it seems could work if i manually add `type: annotation` below 
```
presta_sitemap:
    resource: "@PrestaSitemapBundle/Resources/config/routing.yml"
```
any tips?

brusch:
@kingjia90 please see ecedde6af003eacb42a67ee9f8d4503aef503177 - I've implemented a different approach, not super nice, but  works 😊
Can u plz double check if that is a proper bc layer? Thanks!

kingjia90:
The BC Layer works fine.
Changed version to [3.2](https://github.com/prestaconcept/PrestaSitemapBundle/releases/tag/v3.2.0)  as it has a better support for Symfony 6.

PR #12022, opened 2022-05-03
[Authentication] Do not close session on authenticateSession

### LINKED ISSUE #11927: [Bug]: Authentication is closing session when its not supposed to
### Expected behavior

`Authentication::authenticateSession($request);` should find the current Admin User in the Session

### Actual behavior

`Authentication::authenticateSession($request);` finds the current user, but it uses `AdminSessionHandler::useSessionAttributeBag` which runs `writeClose` closing the current session and therefore hindering Symfony from sending the right newly, generated and migrated, session ID to the user since the session is actually closed.

### Steps to reproduce

Enabled Pimcore Targeting:

```
pimcore:
    targeting:
        enabled: false
```

And authenticate to a firewall. In my case it happens with CoreShop Tests: https://github.com/coreshop/CoreShop/pull/1970/files#diff-2f8cab0f04dcf2454dbe1699f7f35fd8d169952ad38ba6baee510b20bde8ef6cR5

dvesh3:
Resolves #11927

### DIFF
--- a/bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
+++ b/bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
@@ -49,6 +49,11 @@ class AdminSessionHandler implements LoggerAwareInterface, AdminSessionHandlerIn
     protected $readOnlySessionBagsCache = [];
 
     /**
+     * @var bool|null
+     */
+    private $canWriteAndClose;
+
+    /**
      * @var RequestHelper
      */
     protected $requestHelper;
@@ -107,17 +112,7 @@ class AdminSessionHandler implements LoggerAwareInterface, AdminSessionHandlerIn
         $attributeBag = $this->loadAttributeBag($name, $session);
 
         $result = call_user_func_array($callable, [$attributeBag, $session]);
-
-        // write & close session when in admin context
-        // https://github.com/pimcore/pimcore/pull/12022#issuecomment-1119451897
-        $request = $this->requestHelper->getCurrentRequest();
-        if (
-            $this->matchesPimcoreContext($request, PimcoreContextResolver::CONTEXT_ADMIN) &&
-            $this->requestHelper->isFrontendRequestByAdmin($request)
-        ) {
-            $this->writeClose();
-        }
-
+        $this->writeClose();
         return $result;
     }
 
@@ -258,6 +253,10 @@ class AdminSessionHandler implements LoggerAwareInterface, AdminSessionHandlerIn
      */
     public function writeClose()
     {
+        if (!$this->shouldWriteAndClose()) {
+            return;
+        }
+
         $this->openedSessions--;
 
         if (0 === $this->openedSessions) {
@@ -272,5 +271,24 @@ class AdminSessionHandler implements LoggerAwareInterface, AdminSessionHandlerIn
                 'count' => $this->openedSessions,
             ]);
         }
+    }
+
+    /**
+     * @return bool
+     */
+    private function shouldWriteAndClose(): bool
+    {
+        if ($this->canWriteAndClose === null) {
+            $this->canWriteAndClose = false;
+            $request = $this->requestHelper->getMainRequest();
+            if (
+                $this->matchesPimcoreContext($request, PimcoreContextResolver::CONTEXT_ADMIN) ||
+                $this->requestHelper->isFrontendRequestByAdmin($request)
+            ) {
+                $this->canWriteAndClose = true;
+            }
+        }
+
+        return $this->canWriteAndClose;
     }
 }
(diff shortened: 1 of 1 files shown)


brusch:
Hmm, I think that has some unwanted side-effects as well, because actually we don't want to keep the session open here (it's about concurrency and session locks). So this fix would cause performance issues (especially in the admin UI) when using a session handler that uses locks (e.g. file handler).

brusch:
Hmm, I think that has some unwanted side-effects as well, because actually we don't want to keep the session open here (it's about concurrency and session locks). So this fix would cause performance issues (especially in the admin UI) when using a session handler that uses locks (e.g. file handler).

brusch:
Maybe a solution could be to change the behavior depending on the context, so if in admin context, we keep the current behavior, if outside, the AdminSessionHandler just does nothing in `writeClose()`?

dvesh3:
> Maybe a solution could be to change the behavior depending on the context, so if in admin context, we keep the current behavior, if outside, the AdminSessionHandler just does nothing in `writeClose()`?

@brusch I changed it to this solution now. thanks!

bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
```diff
@@ -98,7 +108,12 @@ public function useSessionAttributeBag(callable $callable, string $name = 'pimco
 
         $result = call_user_func_array($callable, [$attributeBag, $session]);
 
-        $this->writeClose();
+        // write & close session when in admin context
+        // https://github.com/pimcore/pimcore/pull/12022#issuecomment-1119451897
+        $request = $this->requestStack->getCurrentRequest();
+        if ($this->matchesPimcoreContext($request, PimcoreContextResolver::CONTEXT_ADMIN)) {
```
brusch: I think we also need to consider `$requestHelper->isFrontendRequestByAdmin($request)` here, so that the editmode is not blocking anything.

brusch (changes requested):

bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
dvesh3: done 😉

bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
```diff
@@ -98,7 +108,15 @@ public function useSessionAttributeBag(callable $callable, string $name = 'pimco
 
         $result = call_user_func_array($callable, [$attributeBag, $session]);
 
-        $this->writeClose();
+        // write & close session when in admin context
+        // https://github.com/pimcore/pimcore/pull/12022#issuecomment-1119451897
+        $request = $this->requestHelper->getCurrentRequest();
+        if (
+            $this->matchesPimcoreContext($request, PimcoreContextResolver::CONTEXT_ADMIN) &&
```
brusch:
```suggestion
            $this->matchesPimcoreContext($request, PimcoreContextResolver::CONTEXT_ADMIN) ||
```
This should be an `or` as the context it not admin in case of `isFrontendRequestByAdmin() === true`
See e.g. https://github.com/pimcore/pimcore/blob/d124b16a837567ed9b7ff010fdeff6e82e9bfaf3/bundles/AdminBundle/EventListener/AdminSecurityListener.php#L70

I'm also just thinking whether we should do the check on the main request and cache the result for further use? wdyt?

brusch (changes requested):

bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
dvesh3: Changed to `||` check and added `$canWriteAndClose` property for caching the result for subsequent requests.

bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
```diff
@@ -273,4 +272,23 @@ public function writeClose()
...
+            if (
+                $this->matchesPimcoreContext($request, PimcoreContextResolver::CONTEXT_ADMIN) ||
+                $this->requestHelper->isFrontendRequestByAdmin($request)
+            ) {
+                $this->canWriteAndClose = true;
+            }
+        }
+
+        return $this->canWriteAndClose;
+    }
```
jdreesen: What about writing it like this? 
For me this is more readable:

```suggestion
    private function shouldWriteAndClose(): bool
    {
        return $this->canWriteAndClose ??= $this->isAdminRequest($this->requestHelper->getMainRequest());
    }

    /**
     * @return bool
     */
    private function isAdminRequest(Request $request): bool
    {
        return $this->matchesPimcoreContext($request, PimcoreContextResolver::CONTEXT_ADMIN)
            || $this->requestHelper->isFrontendRequestByAdmin($request);
    }
```

dvesh3: and reusable as well 😉 thanks!

brusch:
Great - thanks guys! 👍

brusch (approved):

PR #12026, opened 2022-05-04
Fix: AbstractUser::getById() doesn't check the type

### LINKED ISSUE #12024: [Bug]: updateAction (UserController - AdminBundle)
### Expected behavior

Admins should be able to create and modify the Users Entities.

### Actual behavior

The updateAction in `bundles/AdminBundle/Controller/Admin/UserController.php` tries to retrieve the User by executing `$user = User\UserRole::getById((int)$request->get('id'));` This fails every time, since the request contains the UserId and not the userRoleId.

The bug seems to have been introduced with the 10.4 release [in this commit](https://github.com/pimcore/pimcore/commit/d5fd15b5a2fae2dba61c4f6bd9423429855834d3). All other methods in the UserController use `User::getById($request->get('id'));`

image

### Steps to reproduce

Try to set a password for an existing user. You will receive a successfull response (status 200, success: true) but nothing gets written in the DB.

blankse:
Fix https://github.com/pimcore/pimcore/issues/12024

The getById method of AbstractUser doesn't check the type.
It only load the correct type if it is loaded over AbstractUser class.

I changed this and optimize the Asset/Document Model in the same behavior (There it is loaded always over the Main Class and loaded again with the type). 

DataObjects load first only the typeInfo and than by the correct Type the data. There is no optimization possible.

### DIFF
--- a/models/User/Service.php
+++ b/models/User/Service.php
@@ -15,6 +15,8 @@
 
 namespace Pimcore\Model\User;
 
+use Pimcore\Model\User;
+
 /**
  * @internal
  */
@@ -29,13 +31,17 @@ class Service
      *
      * @return string|null
      */
-    public static function getClassNameForType($type)
+    public static function getClassNameForType($type): ?string
     {
         switch ($type) {
-            case 'user': return '\\Pimcore\\Model\\User';
-            case 'userfolder': return '\\Pimcore\\Model\\User\\Folder';
-            case 'role': return '\\Pimcore\\Model\\User\\Role';
-            case 'rolefolder': return '\\Pimcore\\Model\\User\\Role\\Folder';
+            case 'user':
+                return User::class;
+            case 'userfolder':
+                return User\Folder::class;
+            case 'role':
+                return User\Role::class;
+            case 'rolefolder':
+                return User\Role\Folder::class;
             default:
                 return null;
         }
--- a/models/User/Service.php
+++ b/models/User/Service.php
@@ -33,17 +33,12 @@ class Service
      */
     public static function getClassNameForType($type): ?string
     {
-        switch ($type) {
-            case 'user':
-                return User::class;
-            case 'userfolder':
-                return User\Folder::class;
-            case 'role':
-                return User\Role::class;
-            case 'rolefolder':
-                return User\Role\Folder::class;
-            default:
-                return null;
-        }
+        return match ($type) {
+            'user' => User::class,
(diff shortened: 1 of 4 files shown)


models/User/Service.php
```diff
@@ -29,13 +31,17 @@ class Service
...
+                return User::class;
+            case 'userfolder':
+                return User\Folder::class;
+            case 'role':
+                return User\Role::class;
+            case 'rolefolder':
+                return User\Role\Folder::class;
             default:
                 return null;
         }
```
jdreesen: This is a great use-case for PHP 8's new [`match` expressions](https://www.php.net/manual/en/control-structures.match.php):

```suggestion
        return match ($type) {
            'user' => User::class,
            'userfolder' => User\Folder::class,
            'role' => User\Role::class,
            'rolefolder' => User\Role\Folder::class,
            default => null,
        }
```

jdreesen: The suggestion that GitHub created is somehow broken O.o

blankse: @jdreesen I can't commit it :( I changed it here: https://github.com/pimcore/pimcore/pull/12026/commits/2d58443f5c8a1a01ca594234df6f5a0efeefad5c

antonio-masotti:
@blankse Thanks for the fixes, interesting implementation! Just one question: the source of the bug #12024 is - as far as I could see - this line (287) in the [UserController of the AdminBundle](https://github.com/pimcore/pimcore/blob/d5fd15b5a2fae2dba61c4f6bd9423429855834d3/bundles/AdminBundle/Controller/Admin/UserController.php):

~~~
        $user = User\UserRole::getById((int)$request->get('id'));
~~~

If this is not changed in 

~~~
        $user = User\AbstractUser::getById((int)$request->get('id'));
~~~

I don't see how the bug would be fixed. Or am I missing something?

blankse:
@antonio-masotti The bug is that the AbstractUser Model only load and check the type if it is called via AbstractUser. See:
https://github.com/pimcore/pimcore/blob/64bbdbbca14d98ad3ce53d2d8816e563351d888b/models/User/AbstractUser.php#L67

In other Models (Asset, Document, DataObject) it is possible to call the Model via a Subtype and the Model check the type and return null if it is wrong. In the UserController it can be a User or a Role. So it exist a UserRole class for the common methods. But the getById method doesn't load the correct Subtype class (User/Role) and leaves the UserRole class. So all instanceof ifs in the controller are false.

PR #12027, opened 2022-05-04
Bump doctrine/doctrine-bundle to 2.5

### LINKED ISSUE #11663: Bump doctrine/doctrine-bundle to 2.5
### Improvement description

Supports Symfony ^5.0|^6.0

mcop1:
Resolves #11663


composer.json
```diff
@@ -53,7 +53,7 @@
     "doctrine/collections": "^1.6.5",
     "doctrine/persistence": "^2.1.0",
     "doctrine/dbal": "^2.12",
-    "doctrine/doctrine-bundle": "^2.2.1",
+    "doctrine/doctrine-bundle": "^2.5.7",
```
dvesh3: There's even nothing to change from 2.5 to 2.6 https://github.com/doctrine/DoctrineBundle/blob/2.6.x/UPGRADE-2.6.md

So maybe we can bump it to 2.6?
```suggestion
    "doctrine/doctrine-bundle": "^2.6.3",
```

mcop1: Basically I would be fine with that, if there aren´t any other implications. I just selected the version based on the Epic in product management

PR #12028, opened 2022-05-04
Fix new PhpStan error

blankse:
Fix new PhpStan error with Version 1.6.5 https://github.com/phpstan/phpstan/releases/tag/1.6.5

### DIFF
--- a/models/DataObject/Traits/ElementWithMetadataComparisonTrait.php
+++ b/models/DataObject/Traits/ElementWithMetadataComparisonTrait.php
@@ -50,7 +50,7 @@ trait ElementWithMetadataComparisonTrait
             if ((!$container1 && $container2) || ($container1 && !$container2)) {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


models/DataObject/Traits/ElementWithMetadataComparisonTrait.php
```diff
@@ -50,7 +50,7 @@ public function isEqual($array1, $array2): bool
             if ((!$container1 && $container2) || ($container1 && !$container2)) {
                 return false;
             }
-            if (!$container1 && !$container2) {
+            if (!$container1) {
```
dvesh3: a bit cleaner:
```php
if (!$container1 || !$container2) {
    return !$container1 && !$container2;
}
```

blankse: @dvesh3 I changed it

dvesh3:
@blankse thanks!

PR #12030, opened 2022-05-04
[Localization] Fix fallback should not be stored in case of 0 or false value

### LINKED ISSUE #11984: [Bug]: When generating a localized view, the fallback language is used if its a checkbox and its value is false
### Expected behavior

Say we have an object with a localized checkbox field `enabled`. We also have a set of languages with fallbacks, e.g. `de_AT` with a fallback to `de`. 

When I set the `enabled` value to `false` (or `0`) within an objects `de_AT`, and to `true` (or `1`) within `de`, the returned value (from the database view) should be `0`. 
So when I run something like this:
```
        $employeeListing = new ContactPerson\Listing();
        $employeeListing->setLocale('de_AT');
        // only show employees that are enabled for this locale
        $employeeListing->setCondition("enabled = 1");
```
it should NOT return this object.

### Actual behavior

However, this object IS returned.

### Steps to reproduce

And this is the reason.
The generated SQL view would look something like this:
```
select `database`.`object_query_9`.`oo_id`                                                     AS `oo_id`,
       `database`.`object_query_9`.`oo_classId`                                                AS `oo_classId`,
       `database`.`object_query_9`.`oo_className`                                              AS `oo_className`,
       `database`.`objects`.`o_id`                                                             AS `o_id`,
       `database`.`objects`.`o_parentId`                                                       AS `o_parentId`,
       `database`.`objects`.`o_type`                                                           AS `o_type`,
       `database`.`objects`.`o_key`                                                            AS `o_key`,
       `database`.`objects`.`o_path`                                                           AS `o_path`,
       `database`.`objects`.`o_index`                                                          AS `o_index`,
       `database`.`objects`.`o_published`                                                      AS `o_published`,
       `database`.`objects`.`o_creationDate`                                                   AS `o_creationDate`,
       `database`.`objects`.`o_modificationDate`                                               AS `o_modificationDate`,
       `database`.`objects`.`o_userOwner`                                                      AS `o_userOwner`,
       `database`.`objects`.`o_userModification`                                               AS `o_userModification`,
       `database`.`objects`.`o_classId`                                                        AS `o_classId`,
       `database`.`objects`.`o_className`                                                      AS `o_className`,
       `database`.`objects`.`o_childrenSortBy`                                                 AS `o_childrenSortBy`,
       `database`.`objects`.`o_childrenSortOrder`                                              AS `o_childrenSortOrder`,
       `database`.`objects`.`o_versionCount`                                                   AS `o_versionCount`,
       `de_AT`.`ooo_id`                                                                                 AS `ooo_id`,
       if(((`de_AT`.`enabled` is null) or (`de_AT`.`enabled` = '')), `de`.`enabled`, `de_AT`.`enabled`) AS `enabled`
from (
    (
        (
            `database`.`object_query_9` 
            join `database`.`objects`
                on ((`database`.`objects`.`o_id` = `database`.`object_query_9`.`oo_id`))
        ) 
        left join `database`.`object_localized_query_9_de_AT` `de_AT`
            on (((0 <> 1) and (`database`.`object_query_9`.`oo_id` = `de_AT`.`ooo_id`)))
    ) 
    left join `database`.`object_localized_query_9_de` `de`
        on (((0 <> 1) and (`database`.`object_query_9`.`oo_id` = `de`.`ooo_id`)))
)
```
and the problem lies within this statement: 
```
if(((`de_AT`.`enabled` is null) or (`de_AT`.`enabled` = '')), `de`.`enabled`, `de_AT`.`enabled`) AS `enabled`
```
because if the value stored in `object_localized_query_9_de_AT`.`enabled` is false (or `0`) and thus
```
`de_AT`.`enabled` = ''
```
will be read as `true`, so the view will actually use the fallback language setting
```
`de`.`enabled`
```
which returns `1`

dvesh3:
Resolves #11984

### DIFF
--- a/tests/Model/DataType/LocalizedFieldTest.php
+++ b/tests/Model/DataType/LocalizedFieldTest.php
@@ -15,6 +15,7 @@
 
 namespace Pimcore\Tests\Model\DataType;
 
+use Pimcore\Config;
 use Pimcore\Model\DataObject;
 use Pimcore\Model\DataObject\Fieldcollection;
 use Pimcore\Model\DataObject\Localizedfield;
@@ -91,5 +92,49 @@ class LocalizedFieldTest extends ModelTestCase
 
         //new value (de): index 0
         $this->assertEquals('textDE', $loadedItem->getLinput('de'), 'New localized value inside fieldcollection not saved or loaded properly');
+    }
+
+    public function testLocalizedFieldFallback()
+    {
+        $configuration = Config::getSystemConfiguration();
+        $configuration['general']['fallback_languages']['de'] = 'en';
+        Config::setSystemConfiguration($configuration);
+
+        $object = TestHelper::createEmptyObject();
+
+        //en values
+        $object->setLinput('TestEN', 'en');
+        $object->setLcheckbox(true, 'en');
+        $object->setLnumber(123, 'en');
+
+        //de values
+        $object->setLinput('TestDE', 'de');
+        $object->setLcheckbox(true, 'de');
+        $object->setLnumber(0, 'de');
+
+        $object->save();
+
+        //empty de values check fallback
+        $object->setLinput('', 'de');
+        $object->setLcheckbox(false, 'de');
+        $object->save();
+
+        $object = DataObject\Unittest::getById($object->getId(), true);
+
+        $this->assertEquals('TestEN', $object->getLinput('de'));
+        $this->assertEquals(0, $object->getLnumber('de')); //fallback should not return, num 0 is set
+
+        //fallback for checkbox is not supported anymore
+        //$this->assertEquals(true, $object->getLcheckbox('de'));
+
+        //asset listing works with fallback value
+        $listing = new DataObject\Unittest\Listing();
+        $listing->setLocale('de');
+
+        $listing->setCondition("linput = 'TestEN'");
+        $this->assertEquals(1, count($listing->load()), 'Expected 1 item for fallback en condition');
+
+        $listing->setCondition("lcheckbox = '1'");
+        $this->assertEquals(0, count($listing->load()), 'Expected 0 item for fallback en condition');
     }
 }
--- a/tests/Model/DataType/LocalizedFieldTest.php
+++ b/tests/Model/DataType/LocalizedFieldTest.php
@@ -146,10 +146,10 @@ class LocalizedFieldTest extends ModelTestCase
         $this->assertEquals(1, count($listing->load()), 'Expected 1 item for fallback en condition');
 
         //special case checkbox: set value to false and test fallback should not work
-        $object->setLcheckbox(false, 'de'); //should not take the fallback
+        $object->setLcheckbox(false, 'de');
         $object->save();
 
-        //todo check getter
+        $this->assertEquals(false, $object->getLcheckbox('de')); //should not take the fallback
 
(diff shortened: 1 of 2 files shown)


tests/Model/DataType/LocalizedFieldTest.php
```diff
@@ -92,4 +93,48 @@ public function testLocalizedFieldInsideFieldCollection()
...
+        $object->setLcheckbox(true, 'en');
+        $object->setLnumber(123, 'en');
+
+        //de values
+        $object->setLinput('TestDE', 'de');
+        $object->setLcheckbox(true, 'de');
+        $object->setLnumber(0, 'de');
+
+        $object->save();
+
```
fashxp, line 116: I would add checks here to make sure getter and listings work properly

tests/Model/DataType/LocalizedFieldTest.php
```diff
@@ -92,4 +93,48 @@ public function testLocalizedFieldInsideFieldCollection()
...
+        //empty de values check fallback
+        $object->setLinput('', 'de');
+        $object->setLcheckbox(false, 'de');
```
fashxp: can't we set `null` here ... as it says we empty the values?

tests/Model/DataType/LocalizedFieldTest.php
```diff
@@ -92,4 +93,48 @@ public function testLocalizedFieldInsideFieldCollection()
...
+        $object->save();
+
+        $object = DataObject\Unittest::getById($object->getId(), true);
+
+        $this->assertEquals('TestEN', $object->getLinput('de'));
+        $this->assertEquals(0, $object->getLnumber('de')); //fallback should not return, num 0 is set
+
+        //fallback for checkbox is not supported anymore
+        //$this->assertEquals(true, $object->getLcheckbox('de'));
```
fashxp: when value is actively set to `false`, I would not expect that fallback value is taken ... only when value is set to `null`.

tests/Model/DataType/LocalizedFieldTest.php
fashxp: we also need to set number to null and check if fallback works properly

tests/Model/DataType/LocalizedFieldTest.php
```diff
@@ -92,4 +93,48 @@ public function testLocalizedFieldInsideFieldCollection()
...
+
+        //asset listing works with fallback value
+        $listing = new DataObject\Unittest\Listing();
+        $listing->setLocale('de');
+
+        $listing->setCondition("linput = 'TestEN'");
+        $this->assertEquals(1, count($listing->load()), 'Expected 1 item for fallback en condition');
+
+        $listing->setCondition("lcheckbox = '1'");
+        $this->assertEquals(0, count($listing->load()), 'Expected 0 item for fallback en condition');
```
fashxp: also test for number field filtering

fashxp (changes requested):

tests/Model/DataType/LocalizedFieldTest.php
dvesh3: I used the value possible to set from Admin UI

dvesh3: checking both cases now, when setting `null` & `false`(special case)

fashxp (approved):

PR #12034, opened 2022-05-05
Change migration to reflect install.sql for storageType column

mcop1:
See
https://github.com/pimcore/pimcore/pull/11458#discussion_r865689435


blankse:
@mcop1 Should we rename the migration, so it will also executed in update from 10.4.0 to 10.4.1?

mcop1:
> @mcop1 Should we rename the migration, so it will also executed in update from 10.4.0 to 10.4.1?

Good point, maybe I will just add a new migration to check if the column is already existing. And if it is existing, change it to "not null".

Issue #10952, opened 2021-12-03
[Bug]: TranslatorInterface created by Container is not supported anymore


alexanderkroneis:
### Expected behavior

Show a list of documents on the left side of the navigation.

### Actual behavior

Modal opens with following error message

```
The "Symfony\Contracts\Translation\TranslatorInterface" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.

Server threw exception - could not perform action. Please reload the admin interface and try again.
```

### Steps to reproduce

- Access your local installation on Pimcore v10.1.5
- Login with valid credentials
- See the error message

alexanderkroneis:
Meanwhile I figured out it's coming from this line:

```
if ($site instanceof Site) {
->  $translator = \Pimcore::getContainer()->get(TranslatorInterface::class);
    $this->elementQtipConfig['text'] .= '<br>' . $translator->trans('site_id', [], 'admin') . ': ' . $site->getId();
}
```

dvesh3:
@alexgaal sorry, it works for me:

![image](https://user-images.githubusercontent.com/5137917/144587748-a4b6813c-5206-412b-a0b6-3ab57e4d5db2.png)

dvesh3:
Closing as there's no response. Please feel free to come back with reproducible case.

kevinulrich:
This happens to me as soon as I convert anything directly under root to a site. Same line as OP has pointed out. Ridiculous, I'm not sure why we're still getting our container from a singleton with `->get()` instead of just using DI for this. Even if this Bug wasn't relevant that's just bad practise.

For anyone looking for a solution, it's a temporary fix to simply alias the Interface to Pimcore Translator. I didn't check this too deeply:

```
# config/services.yaml

Symfony\Contracts\Translation\TranslatorInterface: '@Pimcore\Translation\Translator'
```

You will also need to make that service us public by either setting it as default or especially targeting that Translator service.

dvesh3:
please see, if you are facing the issue https://github.com/pimcore/pimcore/pull/12041#discussion_r867312266

micru:
Frankly, the use of `\Pimcore::getContainer()` is such bad practice that it simply should not be used and should be completely removed from the core of Pimcore. Perhaps apart from special cases, such as CalculatedValue fields when the service is loaded from a configuration specified by the user via the Admin UI.

[...]

PR #12041, opened 2022-05-06
Exception "TranslationInterface service has been removed . . ." on multisite environment

mittererr:
After updating from 6.9.6 to 10.4.0 and Symfony Flex, Eception `The "Symfony\Contracts\Translation\TranslationInterface" service has been removed or inlined when the container was compiled.` was thrown:

Screen Shot 2022-05-05 at 19 46 41

This topic was also discussed in #10952. Hopefully this is the final solution.

### DIFF
--- a/models/Element/AdminStyle.php
+++ b/models/Element/AdminStyle.php
@@ -22,7 +22,7 @@ use Pimcore\Model\DataObject\Concrete;
 use Pimcore\Model\DataObject\Folder;
 use Pimcore\Model\Document;
 use Pimcore\Model\Site;
-use Symfony\Contracts\Translation\TranslatorInterface;
+use Pimcore\Translation\Translator;
 
 class AdminStyle
 {
@@ -49,7 +49,7 @@ class AdminStyle
     /**
      * @param AbstractObject|Asset|Document|ElementInterface $element
      */
-    public function __construct(ElementInterface $element)
+    public function __construct(ElementInterface $element, Translator $translator)
     {
         if ($element instanceof AbstractObject) {
             if ($element instanceof Folder) {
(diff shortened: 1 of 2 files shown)


models/Element/AdminStyle.php
```diff
@@ -49,7 +49,7 @@ class AdminStyle
     /**
      * @param AbstractObject|Asset|Document|ElementInterface $element
      */
-    public function __construct(ElementInterface $element)
+    public function __construct(ElementInterface $element, Translator $translator)
```
dvesh3, line 52: BC break as this can be extended in Custom DataObject classes https://pimcore.com/docs/pimcore/current/Development_Documentation/Tools_and_Features/Custom_Icons.html#page_Example-Custom-Icon-for-the-Car-DataObject

dvesh3: I'm wondering why it doesn't work in your case, as the alias is there https://github.com/pimcore/pimcore/blob/417b50202ceba7721311eb73f92c308f84236546/bundles/CoreBundle/Resources/config/aliases.yaml#L13

mittererr: I have seen the alias, but it does not seem to work in my case. I have commented/uncommted it, without any effect!?

dvesh3: please try this on your console and check the output:
```cmd
bin/console debug:container Symfony\Contracts\Translation\TranslatorInterface
```
It should say it is a public  alias for service `Pimcore\Translation\Translator`

mittererr: image

mittererr: My fault - removed `Symfony\Bundle\FrameworkBundle\FrameworkBundle::class => ['all' => true]` from `bundles.php` after Symfony flex installation and everything is fine. Otherwise `aliases.yaml` ist not loaded.

PR #12042, opened 2022-05-06
Fix PhpDocs

blankse:


fashxp:
@blankse what is the status of that? ready for review? we should be careful that the PR doesn't get too big...

blankse:
@fashxp Yes, it is ready for review.

lib/Messenger/Handler/SanityCheckHandler.php
```diff
@@ -59,12 +59,15 @@ private function process(array $jobs): void
...
 
     /**
-     * @param PageSnippet|Asset|Concrete $element
+     * @param ElementInterface $element
      *
      * @throws \Exception
      */
     private function performSanityCheck(ElementInterface $element)
     {
+        if (!$element instanceof PageSnippet && !$element instanceof Concrete) {
```
fashxp: isn't there the check for asset missing?

fashxp (changes requested):

lib/Messenger/Handler/SanityCheckHandler.php
fashxp: @blankse  actually this could be considered as a BC break anyway ... should we just get rid of that additional if?

blankse: @fashxp I add the missing Asset check. These are the only ElementTypes which have the getLatestVersion method. I thought about an Interface for this classes, but the method is on the Dao classes.

Where do you see a BC break? Shoud I remove the if and add the phpstan error to the baseline?

fashxp: I see, good point. 
The BC break idea is a more academic one ... I think, with the additional check we are good to go 👍
Thx for the efforts...

fashxp (approved):

dvesh3 (approved):

Issue #11997, opened 2022-04-28
[Bug]: getThumbnail() returns a path with the image id included in the path in staging server (live)


swina:
### Expected behavior

Using in local environment / staging environment :

`getImage()->getThumbnail()->getPath();`

returns a path like :

`_thumbnail_prefix_url/_thumbnail_path_/_thumbnail_file_name_`

for example: 

`https://myserver.com/assets/images/image-thumb__2953__slider_image/my_image.jpg`

### Actual behavior

In staging environment (live server) returns :

`_thumbnail_prefix_url_/_thumbnail_path_/image_id/_thumnail_file_name_generated_`

that results as a broken link (image_id added to the path) before the `_thumbnail_file_name_generated

`https://myserver.com/assets/images/2953/image-thumb__2953__slider_image/my_image.jpg`

instead of the correct url

`https://myserver.com/assets/images/image-thumb__2953__slider_image/my_image.jpg`

### Steps to reproduce

getThumbnail should return always the same url like : 

`_thumbnail_prefix_url/_thumbnail_path_/_thumbnail_file_name_`

brusch:
You have most probably different Pimcore Versions in use. 

See also upgrade notes: https://pimcore.com/docs/pimcore/current/Development_Documentation/Installation_and_Upgrade/Upgrade_Notes/index.html#page_10-4-0

GALCF:
@brusch I have the exact same problem with the current version (10.4.0). The problem only occurs when an asset ist located directly underneath the asset root and not in a separate folder. This is also reproducible in the demo. Just move a used image to the root folder and set a name without spaces / special characters. The image will not work in frontend.

brusch:
@GALCF Thanks for the detailed info, I've just reopened the issue. 
As you can reproduce the issue on your side, would it be possible for your to provide us a PR fixing the issue? 
Thanks a lot!

GALCF:
@brusch Done, but I'm not sure if there are any edge-cases that break because of this prefix replacement. I tested root and non-root image thumbnails, and they seem to work

PR #12047, opened 2022-05-06
Fix: Root asset thumbnails

GALCF:
Relates to #11997


brusch:
Thank you very much! 
I'm closing this there in favor of #12058, because it doesn't have the unneccessary escape in it. 🤗

Issue #12045, opened 2022-05-06
[Bug]: LQIP are not loaded when img tag has a srcset attribute


Brainshaker95:
### Expected behavior

The srcset attribute should not be set, when lowQualityPlaceholder is set

### Actual behavior

The srcset attribute is set and so the lowQualityPlaceholder in the src attribute is ignored.

### Steps to reproduce

Add an image with the thumbnail configuration `lowQualityPlaceholder: true`

fashxp:
@Brainshaker95  isn't that fixed already?

brusch:
Yes, that got fixed.

PR #12049, opened 2022-05-06
Fix: LQIP srcset

GALCF:
Relates to #12045


brusch:
Nice, thanks a lot! 👍

PR #12050, opened 2022-05-06
Fix: Save permissions keybindings

### LINKED ISSUE #11992: [Bug]: Keyboard shortcut to save causes error within folders
### Expected behavior

When pressing the shortcut to save (ctrl + s), it should save (here the folder).

However, a popup is then visible, saying, that save was not successful. If I manually click the save button, it works fine and no error message is visible.

image

### Actual behavior

image

It triggers a save action, but not the correct one.

This is what I get inside network traffic out of debugger console

image

### Steps to reproduce

- Create a folder in the document tree
- Press the shortcut to save when within the folder window (usually CTRL+S)
- You'll see an error message. If you click the save button by hand, you won't get it

GALCF:
Relates to #11992


mcop1:
Thanks 👍

PR #12056, opened 2022-05-06
fix: update asset cache on miss

### LINKED ISSUE #12055: [Bug]: Pimcore serving deferred thumbnail URL even though the thumbnail exists (Azure Storage used)
### Expected behavior

```yaml
flysystem:
    storages:
        pimcore.thumbnail.storage:
            visibility: public

pimcore:
    assets:
        frontend_prefixes:
```

### Actual behavior

### Steps to reproduce

dkarlovi:
Resolves #12055.

## Additional info  

This is actually quite convoluted:
1. deferred=true + cache miss will create a deferred URL
2. deferred URL (deferred = false implicit) will skip updating the asset cache
3. if your asset cache is empty, but your thumbnail modification time on storage is newer than the asset, the cache never gets updated and it *always* generates the deferred URL, even though the file exists on storage

### DIFF
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -184,6 +184,10 @@ class Processor
         $statusCacheEnabled = \Pimcore::getContainer()->getParameter('pimcore.config')['assets']['image']['thumbnails']['status_cache'];
         if ($statusCacheEnabled && $deferred) {
             $modificationDate = $asset->getDao()->getCachedThumbnailModificationDate($config->getName(), $filename);
+            if ($modificationDate === null && $storage->fileExists($storagePath)) {
+                $modificationDate = $storage->lastModified($storagePath);
+                $modificationDate = $asset->getDao()->addToThumbnailCache($config->getName(), $filename);
+            }
         } else {
             if ($storage->fileExists($storagePath)) {
                 $modificationDate = $storage->lastModified($storagePath);
--- a/models/Asset/Image/Thumbnail/Processor.php
+++ b/models/Asset/Image/Thumbnail/Processor.php
@@ -186,7 +186,7 @@ class Processor
             $modificationDate = $asset->getDao()->getCachedThumbnailModificationDate($config->getName(), $filename);
             if ($modificationDate === null && $storage->fileExists($storagePath)) {
                 $modificationDate = $storage->lastModified($storagePath);
-                $modificationDate = $asset->getDao()->addToThumbnailCache($config->getName(), $filename);
+                $asset->getDao()->addToThumbnailCache($config->getName(), $filename);
             }
         } else {
             if ($storage->fileExists($storagePath)) {


brusch:
@dkarlovi Thank you very much for your PR. However, we also had this discussion internally and we came the the conclusion that this implementation is making the cache less efficient, because every time you have a miss, it checks again the file-system. 

Consider the following example: 
```html
<picture >
    <source srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--f9fb8a03--query.avif 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--f9fb8a03--query@2x.avif 2x" media="(max-width: 1280px)" type="image/avif" />
    <source srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--f9fb8a03--query.webp 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--f9fb8a03--query@2x.webp 2x" media="(max-width: 1280px)" type="image/webp" />
    <source srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--f9fb8a03--query.png 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--f9fb8a03--query@2x.png 2x" media="(max-width: 1280px)" type="image/png" />
    <source srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--cac93fe4--query.avif 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--cac93fe4--query@2x.avif 2x" media="(max-width: 768px)" type="image/avif" />
    <source srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--cac93fe4--query.webp 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--cac93fe4--query@2x.webp 2x" media="(max-width: 768px)" type="image/webp" />
    <source srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--cac93fe4--query.png 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts~-~media--cac93fe4--query@2x.png 2x" media="(max-width: 768px)" type="image/png" />
    <source srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts.avif 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts@2x.avif 2x" type="image/avif" />
    <source srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts.webp 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts@2x.webp 2x" type="image/webp" />
    <source srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts.png 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts@2x.png 2x" type="image/png" />
    <img class="d-block w-100" src="/Sample/image-thumb__337__portalCarousel/car%20fronts.png" width="1920" height="600" alt="" loading="lazy" srcset="/Sample/image-thumb__337__portalCarousel/car%20fronts.png 1x, /Sample/image-thumb__337__portalCarousel/car%20fronts@2x.png 2x" />
</picture>
```

Here we have 21 unique files references, most of them are most probably not generated yet.
With your fix, in worst case, there will be 21 checks on the file-system ... 

So I think we should maybe think about a cache sync command or something similar, but not doing some implicit checks as it is implemented in this PR.

dkarlovi:
I see your point and agree. :+1: In that case, the existing thumbnail generator script could update the cache when you run it.

The idea might be to make the thumbnail generator idempotent:
1. it checks if the thumbnail exists
2. if not, creates it
3. either way, adds info about it to the cache

When you run the command, it should do anything it takes to get the thumbnails up to date. I wouldn't go with a separate command since it's quite niche.

dkarlovi:
> most of them are most probably not generated yet

IMO, Pimcore should optimize (or have a config) to always try and have all the images generated up front. It's much better for app stability and performance to generate images in workers than in web requests: worker request rates can be restricted and controlled, web requests are dependent on users.

brusch:
@dkarlovi I agree - can you update the command accordingly?

dkarlovi:
OK, maybe we can do that and #12057 in the same PR? I see you've tagged that one for 10.5, but those both seem like bugs to me.

brusch:
Yes, you can put them into the same PR - that's fine. 
Well, it's not really a bug, but we can discuss to add this to a bug fix release as well. 
Looking forward to your PR. 😊

PR #12064, opened 2022-05-09
Bugfix: Update User does not work anymore

lukas-schnieper:
The update action needs load a User object and not a UserRole object. Otherwise changes are not saved.


blankse:
Duplicate of https://github.com/pimcore/pimcore/pull/12026
This fix is false, because it can also be a Role

brusch:
@lukas-schnieper Thanks a lot for your PR, however, closing it in favor of #12026

PR #12069, opened 2022-05-09
Fix/dynamic multiselect options not loaded in grid

### LINKED ISSUE #12038: [Bug]: options provider multiselect object context missing
### Expected behavior

When retrieving options for a multiselect field from an options provider a data object is set in `$context['object']` when loading the field in a grid: `$context['purpose'] === 'gridconfig'`.

### Actual behavior

The `$context['object']` value is null. This prevents you from retrieving options dynamically based the value of another field. It does work for regular select fields and in the regular class layout.

### Steps to reproduce

- Create an [options provider](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Data_Types/Dynamic_Select_Types.html).
- Add a condition to return an option labeled 'no object' if no data object is available and some other options if it ís available.
- Configure the options provider on a multiselect field.
- Load the data object in a (folder) grid.
- Click the field to edit the options. The 'no object' option will pop up.

I attempted to debug this issue, but I got a bit lost in the code. There appear to be some differences between behaviour of the select and multiselect fields. In the DataObject\Service::gridObjectData() method:
- getDataForGrid() is called on the field definition. The select field implementation of this method is way more complex and seems to execute the options provider again. Attempting to copy this for the multiselect didn't work unfortunately.
- For regular select fields an '%options' property is added to the data, which can be seen in the 'grid-proxy' response. The select JavaScript also contains some related logic. I'm not sure what it's used for. Removing it doesn't seem to affect the regular select options in the grid and options are also available in the field itself next to the value (in the JSON response).

In addition, the multiselect is loaded in an ExtJS window, so that may have got something to do with it.

kjkooistra-youwe:
Resolves #12038

## Additional info  

When using an options provider to generate options depending on a value from the data object the options were not loaded in the grid, because the data object is not available when retrieving the general grid configuration. It does work for the regular select, because it has a separate grid data implementation for the grid-proxy setup. Fixed by adding the same implementation for the multiselect field and updating the JavaScript file accordingly.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
@@ -102,11 +102,18 @@ pimcore.object.helpers.grid = Class.create({
                     }
                 }
 
-                if (pimcore.object.tags[type] && pimcore.object.tags[type].prototype.allowBatchAppend) {
-                    batchAppendColumns.push(key);
-                }
-                if (pimcore.object.tags[type] && pimcore.object.tags[type].prototype.allowBatchRemove) {
-                    batchRemoveColumns.push(key);
+                // Don't add batch options if field has dynamic options
+                if (
+                    !fieldConfig.hasOwnProperty('layout')
+                    || !fieldConfig.layout.hasOwnProperty('dynamicOptions')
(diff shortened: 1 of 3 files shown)


robertSt7:
Thank you for the PR

Issue #8917, opened 2021-04-27
Move properties, children and siblings to ElementAbstract


brusch:
- [x] properties https://github.com/pimcore/pimcore/pull/12080
  - setProperties
  - setProperty
  - getProperties
  - getProperty	
- [x] children  --> doesn't make sense, as they are different for every type anyway
  - hasChildren
  - setChildrenSortBy
  - getChildrenSortBy
  -  setChildren
  -  getChildren
- [x] siblings  --> doesn't make sense, as they are different for every type anyway				
   - hasSiblings
   -  getSiblings
- [x] versionCount https://github.com/pimcore/pimcore/pull/12154	
- [x] parent https://github.com/pimcore/pimcore/pull/12110	
- [ ] parentId 
  -  getParentId
  -  setParentId
- [x] setUserOwner https://github.com/pimcore/pimcore/pull/12155
- [x] userModification https://github.com/pimcore/pimcore/pull/12157
- [x] id https://github.com/pimcore/pimcore/pull/12137
- [x] creationDate https://github.com/pimcore/pimcore/pull/12164
- [x] modificationDate https://github.com/pimcore/pimcore/pull/12170
- [x] path https://github.com/pimcore/pimcore/pull/12159
- [x] locked https://github.com/pimcore/pimcore/pull/12171
- [x] key https://github.com/pimcore/pimcore/pull/12158
- [x] Clone & Sleep method https://github.com/pimcore/pimcore/pull/12125

mcop1:
* Move variables to Abstractelement and include setter + getter methods
	* Attention: Properties have a "o_" prefix in AbstractObject class, e.g. https://github.com/pimcore/pimcore/blob/244af0cf550f85b360f243cd44e6771768e5b7f1/models/DataObject/AbstractObject.php#L173
	* properties
		* Be careful, methods are not exactly the same, because of the cache key. Could be passed as an argument, e.g.
			https://github.com/pimcore/pimcore/blob/244af0cf550f85b360f243cd44e6771768e5b7f1/models/DataObject/AbstractObject.php#L1267
		* There are 2 different getter/setter
			* setProperties
			* setProperty
			* getProperties
			* getProperty
		
	* children
		* hasChildren
		* setChildrenSortBy
		* getChildrenSortBy
		* setChildren
		* getChildren
		
	* siblings				
		* hasSiblings
		* getSiblings
		
	* versionCount		
	* parent	
	* parentId
[...]

PR #12070, opened 2022-05-09
[Task] Refactored common properties and getter/setter methods

aashan10:

Moved common properties and their respective getters/setters from AbstractObject to AbstractElement

Resolves #8917 

## Additional info  
- methods relating to children and siblings not moved because of the conflicting method signatures with the implementations in other classes


brusch:
splitted into separate PRs, like #12080

brusch:
splitted into separate PRs, like #12080

PR #12078, opened 2022-05-10
Added latestVersion to forbiddenNames

cuca24:
This commit adds "latestVersion" as forbiddenName to object fields. If the field is called "latestVersion"  opening existing or creating new object would stop working in the admin and exception will be thrown in bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php inside function getLatestVersion on line 2319

  
  

Resolves #


dvesh3:
🍒 picked https://github.com/pimcore/pimcore/commit/1acde505b24e917668ceb4f9e7f19a75967e6010 on 10.4 branch. thanks!

PR #12080, opened 2022-05-10
Removed duplicate properties

aashan10:
Migrating properties with `o_` prefix to the properties with removed prefixes

partially fixes #8917

### DIFF
--- a/models/DataObject/AbstractObject.php
+++ b/models/DataObject/AbstractObject.php
@@ -167,6 +167,7 @@ abstract class AbstractObject extends Model\Element\AbstractElement
 
     /**
      * @internal
+     * @deprecated
      *
      * @var array|null
      */
@@ -231,6 +232,11 @@ abstract class AbstractObject extends Model\Element\AbstractElement
      * @var int
      */
     protected $o_versionCount = 0;
+
+    public function __construct()
+    {
+        $this->o_properties = & $this->properties;
+    }
 
     /**
      * @static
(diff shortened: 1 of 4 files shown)


models/DataObject/AbstractObject.php
```diff
@@ -232,6 +233,11 @@ abstract class AbstractObject extends Model\Element\AbstractElement
      */
     protected $o_versionCount = 0;
 
+    public function __construct()
+    {
+        $this->o_properties = & $this->properties;
```
brusch: @aashan10 as `o_properties` is tagged as `@internal` we don't need this BC layer here and we can just remove the property.

brusch (changes requested):
Of course we also need to consider `Asset`  and `Document`. 😊

models/DataObject/AbstractObject.php
brusch: additionally we need to consider this change in `_sleep()` and `wakeup()`.

aashan10: @brusch I didn't see `o_properties` being used by `Asset` and `Document` classes so I don't know if anything needs to be changed there.

Also, I have updated the `__sleep` and `__wakeup` methods.

brusch: @aashan10 there, they are not prefixed 😉 The `o_` prefix was only on data objects for historical reasons, but actually most of them also exist in `Asset` and `Document` without the prefix, but the same purpose. 

See e.g.: 
https://github.com/pimcore/pimcore/blob/d05c900e6e40273ac80c345ef74b9a3c812de91a/models/Asset.php#L1413
https://github.com/pimcore/pimcore/blob/430706e64baa0b42112541a4379f63387061d0c2/models/Document.php#L1254

They share exactly the same code, that can be unified into `ElementAbstract`. 
Things like the cache tag, can be solved by e.g. using `Element\Service::getElementType()`.

aashan10: Ahh .. got it. Thanks. I'm having better understanding now. I'll have to remove these duplicate methods in Asset and Document as well. Got it!

models/Element/AbstractElement.php
```diff
@@ -47,6 +48,74 @@ abstract class AbstractElement extends Model\AbstractModel implements ElementInt
...
+     */
+    public function setProperty($name, $type, $data, $inherited = false, $inheritable = false)
+    {
+        $this->getProperties();
+
+        $property = new Model\Property();
+        $property->setType($type);
+        $property->setCid($this->getId());
+        $property->setName($name);
+        $property->setCtype('object');
```
brusch:
```suggestion
        $property->setCtype(Service::getElementType($this));
```

brusch (changes requested):

PR #12081, opened 2022-05-10
Fix PhpStan error after merge conflict

blankse:
The lines was removed in https://github.com/pimcore/pimcore/pull/12011 but was not removed after merging 10.4 to 10.x


brusch:
That was fast, thanks a lot 👍

PR #12084, opened 2022-05-10
Custom layout for object bricks

### LINKED ISSUE #10924: Custom Layouts for object bricks + field collections

BlackbitDevs:
This PR implements custom layouts for object bricks.

There are 2 features:
1. In class custom layout for an object brick container you can define which bricks should be addable / visible in this custom layout
2. For a certain object brick define the fields which shall be visible (in a specific custom layout)

As it is only possible to select one custom layout to be used for loading a certain object, it would not make sense to be able to create custom layouts on object brick level. Thus on the object bricks only *existing* custom layouts from the corresponding classes are shown. This way you can *extend* an existing class custom layout. 

As a minor change this PR removes the tree root for class definition tree in custom layout panel (the left tree) to save space - tree root does not make sense here imho (in contrast on custom layout tree where it does make sense as a drop target)

Partly resolves #10924

If this PR is ok, we can do the same for field collections.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
@@ -1345,11 +1345,7 @@ class ClassController extends AdminController implements KernelControllerEventIn
                     ];
             } else {
                 if ($forObjectEditor) {
-                    $itemLayoutDefinitions = null;
-
-                    if ($itemLayoutDefinitions === null) {
-                        $layout = $item->getLayoutDefinitions();
-                    }
+                    $layout = $item->getLayoutDefinitions();
 
                     $currentLayoutId = $request->get('layoutId', null);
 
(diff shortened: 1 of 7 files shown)


fashxp:
@BlackbitNeueMedien  is this ready for review?

BlackbitDevs:
@fashxp, Yes it is.

robertSt7:
image


When I select the Custom Layout for the Class (CLBrick - clDemo) everything works fine and when I try to select the Custom Layout for the Brick (CLBrick - clDemo Props) I see everything as no layout would be selected. 

What do you think about this? Maybe we could hide the Custom Layouts for the Bricks in the selection or something like that

BlackbitDevs:
@robertSt7 You are right. I adjusted this PR so that the brick custom layouts do not get shown in the user permission workspaces panel.

robertSt7:
Many thanks for the PR

PR #12092, opened 2022-05-11
Delete uuid - query update to reduce locking

hstjep:
In addition to the PR #11959, there is another thing that could be optimized when deleting uuid.

"uuids" table has index on primary key which consists of three columns - itemId, type and uuid. However, when deleting uuid only "uuid" value is used in the WHERE clause. As mysql searches the row using the non-indexed column, it often sets additional gap locks on uuids table as shown below:

image

After adding all three columns in the WHERE clause, delete seems to be a lot more efficient. It can find row faster using the index what requires less locking: 

image

Please check PR.

### DIFF
--- a/models/Tool/UUID/Dao.php
+++ b/models/Tool/UUID/Dao.php
@@ -65,7 +65,11 @@ class Dao extends Model\Dao\AbstractDao
         if (!$uuid) {
             throw new \Exception("Couldn't delete UUID - no UUID specified.");
         }
-        $this->db->delete(self::TABLE_NAME, ['uuid' => $uuid]);
+
+        $itemId = $this->model->getItemId();
+        $type = $this->model->getType();
+
+        $this->db->delete(self::TABLE_NAME, ['itemId' => $itemId, 'type' => $type, 'uuid' => $uuid]);
     }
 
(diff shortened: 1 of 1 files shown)


kingjia90:
Thank you for the PR! 
I've tested and i am confirming what you described.

This change should be likely considered an improvement. Could you please rebase and switch to 10.x branch?

hstjep:
Switched to 10.x branch.

jdreesen:
Why didn't the tests run? :raised_eyebrow: Seems like only the CLA was checked...

kingjia90:
> Why didn't the tests run? 🤨 Seems like only the CLA was checked...

@jdreesen that's a good question, unfortunately i have no idea.

For first-time contributor, there would be a button requiring manual run of the Actions but this wasn't the case, because there are recent/past PRs from her (eg. #11959). 

Anyway, the changes do not affect the core (not sure if tests would also cover UUID but that's another issue 🙈) and there would be a further check in the branch after merge, so i guess it's fine and "safe", but thank you for the heads-up. 
I have also mentioned this to colleagues as well, so if similar problems are happening again in future, we will check further.

Issue #12066, opened 2022-05-09
[Feature]: make configuration storage locations configurable via Pimcore's Symfony config tree


dkarlovi:
### Feature description

Currently, [the config storage location tweaks are available via env vars](https://pimcore.com/docs/pimcore/current/Development_Documentation/Deployment/Configuration_Environments.html#page_Configuration-Storage-Locations-Fallbacks).

This, to me, feels counter-intuitive because this configuration is quite central to Pimcore's behaviour, it should be manageable app-wide (meaning, the app should explicitly allow its own config to be environment-specific). This mostly means this shouldn't be random environment vars which do this from thin air, it should be explicitly stated in Pimcore's own config:

```yaml
pimcore:
   config:
       storage:
           image_thumbnails: settings-store
```

This allow the app to enforce where it expects the config to be, with the developer being able to explicitly allow the current behaviour

```yaml
pimcore:
   config:
       storage:
           image_thumbnails: '%env(PIMCORE_WRITE_TARGET_IMAGE_THUMBNAILS)%'
```

but being able to ship this decision with the app.

In general, I feel Pimcore is way too reliant on environment variables which should all actually be Symfony config, in which case the developer would be able to make them use env vars, hardcode, make them configurable per (Symfony) environment, etc.

Note, **this does not work**:

```yaml
parameters:
    env(PIMCORE_WRITE_TARGET_IMAGE_THUMBNAILS): settings-store
```

robertSt7:
- implement changes already in 10.6 with BC layers
- Add method `addWriteTargetNodes` to https://github.com/pimcore/pimcore/blob/b864c79c30abd8f0f741852ae45ac6352a78d540/bundles/CoreBundle/src/DependencyInjection/Configuration.php and set as default value `symfony-config` for every env-variable
- Replace the env-variables with the config values
- Adjusting the bundles because of the changes of https://github.com/pimcore/pimcore/blob/83142b53cfd5a791e3c6167470a66336fd03da22/lib/Model/Dao/PimcoreLocationAwareConfigDao.php 
- https://github.com/pimcore/pimcore/pull/14263#pullrequestreview-1297624842
- https://github.com/pimcore/pimcore/pull/14263#discussion_r1114131480 remove WRITE_TARGET in 11
- changes in `LocationAwareConfigRepository` 
  - replace argument `$writeTargetEnvVariableName` with `$writeTarget` to pass write target instead of env variable name
[...]

Issue #11411, opened 2022-02-09
[Bug]: \Pimcore\Bootstrap::prepareEnvVariables forces using DotEnv


dkarlovi:
### Expected behavior

When running in modern environments like Lambda or containers (local dev and prod), the env vars are not distributed via files but are **real environment variables**. In which case the DotEnv component creates some issues and noise, so it's generally a best practice to not use it there at all. This is known even to Symfony so [they support checking for it being installed or disabling it in their Runtime component](https://symfony.com/doc/current/components/runtime.html#using-options).

Pimcore currently hard requires the `DotEnv` component (which is IMO a mistake, as discussed in #11105), but the component itself can be removed in a hackish way by adding this to your own `composer.json`:

```json
    "provide": {
        "symfony/dotenv": "v5.4.0"
    },
```

The `DotEnv` component should be removed as the hard dependency and the loading using it of it should be made optional (either via config or at least checking with `class_exists` that it's there).

### Actual behavior

`DotEnv` component hardcoded into Pimcore 10.x

### Steps to reproduce

1. add to your `composer.json`:
```json
    "provide": {
        "symfony/dotenv": "v5.4.0"
    },
```
2. run `composer update`, `symfony/dotenv` gets removed
3. `\Pimcore\Bootstrap::prepareEnvVariables` complains about the class not being available

dpfaffenbauer:
Solution is: make `dotenv` optional and check if installed in `prepareEnvVariables`?

dkarlovi:
@dpfaffenbauer I think so, yes. AFAIK, it's only ever used there and the behaviour is optional anyway: if your `.env` file is empty, it does nothing. This would make the whole component optional (can be suggested in `composer.json`) and, if there, it loads the `.env` file.

Issue #10344, opened 2021-09-17
[Improvement]: Leverage the new Symfony Runtime component


brusch:
### Improvement description

See: 

https://symfony.com/doc/current/components/runtime.html
https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/5.3/public/index.php

dkarlovi:
Ideally, Pimcore should **not** provide its own custom runtime to solve architectural problems, even though I'd expect it will be the first direction here.

The reason is: you'll want to do that to **actually swap out runtimes for the intended purpose**, for example using [Bref](https://github.com/php-runtime/bref) to deploy Pimcore as AWS Lambda maybe? Having a Pimcore runtime + and then needing to integrate it somehow with Bref (or Swoole or other) runtime implementation will be a huge obstacle.

mattamon:
Using Symfony Runtime on demo and demo-enterprise now

PR #12099, opened 2022-05-11
Allow skipping the loading of the `.env*` file

jdreesen:

Pimcore wants to load `.env*` files when bootstrapping and fails when there's none. But sometimes you don't have an `.env*` file, because everything is already set as "real" env variables (e.g. in a prod environment). 

Currently, you have to create an empty `.env` file to keep Pimcore happy, but it would be nice to be able to skip this (overhead). Therefore, I added a `SKIP_DOTENV_FILE` env variable that, when set, skips the `.env*` file loading.

### DIFF
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -19,6 +19,7 @@ use Pimcore\Model\DataObject;
 use Pimcore\Model\Document;
 use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\ErrorHandler\Debug;
+use Symfony\Component\Finder\Finder;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\KernelInterface;
 
@@ -146,7 +147,12 @@ class Bootstrap
 
     private static function prepareEnvVariables()
     {
-        (new Dotenv())->bootEnv(PIMCORE_PROJECT_ROOT .'/.env');
+        $finder = new Finder();
+        $finder->files()->name('.env*')->in(PIMCORE_PROJECT_ROOT)->depth('== 0');
+
+        if ($finder->hasResults()) {
+            (new Dotenv())->bootEnv(PIMCORE_PROJECT_ROOT.'/.env');
+        }
     }
 
     public static function defineConstants()
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -19,7 +19,6 @@ use Pimcore\Model\DataObject;
 use Pimcore\Model\Document;
 use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\ErrorHandler\Debug;
-use Symfony\Component\Finder\Finder;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\KernelInterface;
 
@@ -147,11 +146,8 @@ class Bootstrap
 
     private static function prepareEnvVariables()
     {
-        $finder = new Finder();
-        $finder->files()->name('.env*')->in(PIMCORE_PROJECT_ROOT)->depth('== 0');
-
-        if ($finder->hasResults()) {
-            (new Dotenv())->bootEnv(PIMCORE_PROJECT_ROOT.'/.env');
+        if ($_SERVER['SKIP_DOTENV_FILE'] ?? false) {
+            (new Dotenv())->bootEnv(PIMCORE_PROJECT_ROOT .'/.env');
         }
     }
 
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -146,7 +146,7 @@ class Bootstrap
 
     private static function prepareEnvVariables()
     {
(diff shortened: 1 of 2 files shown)


jdreesen:
No idea what's wrong with PHPStan 😞

lib/Bootstrap.php
```diff
@@ -146,7 +146,9 @@ public static function bootstrap()
 
     private static function prepareEnvVariables()
     {
-        (new Dotenv())->bootEnv(PIMCORE_PROJECT_ROOT .'/.env');
+        if ($_SERVER['SKIP_DOTENV_FILE'] ?? false) {
```
blankse:
```suggestion
        if (!($_SERVER['SKIP_DOTENV_FILE'] ?? false)) {
```

jdreesen: Oh my... 🤦 thx!

brusch:
Should we maybe just check if the file exists?

jdreesen:
Yeah, that was my first intent (see first commit in this PR) but it's not soo easy, as it may not be exactly that file, as Symfony tries to load several derived files, too (like `.env.local.php` or `.env.dist`). Also, it has a (bigger) performance impact.

Thus, I thought maybe it's better to just add another env var that can be set on systems that configure all env vars system-wide anyway.

brusch:
@jdreesen Yep, makes sense, but `PIMCORE_PROJECT_ROOT .'/.env'` needs to exist in any case, right, so we won't need to check for all the derived files, right? 
But anyway, I'm fine with the env var, but maybe we should prefix it with `PIMCORE_` as we do with others and also we need to add this to the docs. 

Thanks in advance!

blankse:
It would also be possible to check after the Dotenv class. So we can make the symfony/dotenv optional in Pimcore 11 and you can remove it via the provide composer option now.
See https://github.com/pimcore/pimcore/issues/11411

jdreesen:
> `PIMCORE_PROJECT_ROOT .'/.env'` needs to exist in any case, right, so we won't need to check for all the derived files, right?

No, it does not. It's just the "base name" for a pattern of files that Symfony tries to load.

First, Symfony [looks for a `PIMCORE_PROJECT_ROOT.'/.env.local.php'`](https://github.com/symfony/dotenv/blob/5.4/Dotenv.php#L150-L155) and if it exists, **only that is loaded** (that's the "prod mode" - this file contains an optimized version of all the possible `.env.*` files).

Otherwise, it [looks for `PIMCORE_PROJECT_ROOT.'/.env'` and `PIMCORE_PROJECT_ROOT.'/.env.dist'`](https://github.com/symfony/dotenv/blob/5.4/Dotenv.php#L113-L117). If the first one exists, it'll be loaded, else the second one will be loaded. If neither exists, [an error is thrown](https://github.com/symfony/dotenv/blob/5.4/Dotenv.php#L566-L568).
(Thereafter, it loads [`.env.local`](https://github.com/symfony/dotenv/blob/5.4/Dotenv.php#L123-L126) and [`.env.{env}`](https://github.com/symfony/dotenv/blob/5.4/Dotenv.php#L132-L134) and [`.env.{env}.local`](https://github.com/symfony/dotenv/blob/5.4/Dotenv.php#L136-L138), if one of them exists (and maybe some other conditions are true), but that doesn't matter here, because one of the first three *has* to exist for these to be loaded at all.) 

So, we would have to check for `.env`, `.env.dist` and `.env.local.php` as far as I understand it.

brusch:
@jdreesen you're right 🤦Thanks for clarifying. 
But as I said, I'd anyway stay on the env variable 😊(just consider the prefix and the docs 😉) 
Thanks!

jdreesen:
Just a sidenote: for Pimcore 11 I'd prefer to make the dependency on the DotEnv component optional instead, as described in #11411.

brusch:
#11411 is already assigned to 11 milestone 😉

dkarlovi:
@jdreesen while I'm happy this issue is getting tackled, it seems Pimcore is relying on a forrest of env variables to do anything already, which is IMO a problem on its own, see #12066. IMO, the goal of Pimcore in the future should be to review all the environment variables and convert them all into Symfony Config instead.

_(The same thing should be done within the codebase with all the constants, but that's another story)_

Another direction is to consider doing it similarly to how [Symfony Runtime](https://github.com/symfony/runtime/blob/5.4/SymfonyRuntime.php) is doing it, making it more easy to align with Runtime in the future, which is a design goal #10344. Maybe we can even solve these two issues together in some way?

dvesh3 (approved):

dvesh3:
For simplicity, let's add the environment variable for now to skip loading the dotenv file. PR's are welcome to improve the approach 😊

jdreesen:
I think this should be added to the [10.4.3](https://github.com/pimcore/pimcore/milestone/166) milestone.

dvesh3:
> I think this should be added to the [10.4.3](https://github.com/pimcore/pimcore/milestone/166) milestone.

@jdreesen good catch! Done 😉

PR #12100, opened 2022-05-12
Add CalculatorClassInterface requirement to Pimcore X Update Notes

blankse:
See https://github.com/pimcore/pimcore/discussions/12059


dvesh3:
@blankse thank you 🙏

PR #12104, opened 2022-05-12
Fix: Cannot instantiate abstract class Pimcore\Model\Document\PrintAbstract

blankse:
Regression from https://github.com/pimcore/pimcore/pull/12026

```
In Document.php line 300:
Cannot instantiate abstract class Pimcore\Model\Document\PrintAbstract
```

The Document Model has abstract classes as child classes. There are getById calls of this abstract classes.

### DIFF
--- a/models/Document.php
+++ b/models/Document.php
@@ -244,7 +244,7 @@ class Document extends Element\AbstractElement
         }
 
         try {
-            $helperDoc = new static();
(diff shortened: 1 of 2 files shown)


Chris-Gett:
I can confim the Bug, fix works for me too

dvesh3:
@blankse thanks!

PR #12105, opened 2022-05-12
Update 06_Create_a_First_Project.md

7ossam7assan:
it's not working if we didn't wrap the content into block section

  
  

Resolves # product twig template content fix

## Additional info  
if we didn't wrap the content into block section it'll throw an exception


fashxp:
thx very much!

PR #12110, opened 2022-05-12
Refactored duplicate o_parent/parent properties

aashan10:
## Additional info  
Migrating o_parent and o_parentId

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1784,6 +1784,14 @@ class Asset extends Element\AbstractElement
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function getParent() : ?Asset
+    {
+        return parent::getParent();
+    }
+
+    /**
      * @param Asset|null $parent
      *
      * @return $this
@@ -1813,16 +1821,6 @@ class Asset extends Element\AbstractElement
 
         return array_diff($parentVars, $blockedVars);
     }
-
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getParent() : ?Asset
-    {
-        return parent::getParent();
-    }
-
 
     public function __wakeup()
     {
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1786,7 +1786,7 @@ class Asset extends Element\AbstractElement
     /**
      * @return Asset|null
      */
-    public function getParent() : ?Asset
+    public function getParent() /** : ?Asset */
     {
         $parent = parent::getParent();
         return $parent instanceof Asset ? $parent : null;
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -48,6 +48,7 @@ use Symfony\Component\Mime\MimeTypes;
  * @method bool __isBasedOnLatestData()
  * @method int getChildAmount($user = null)
  * @method string|null getCurrentFullPath()
+ *  * @method ?Asset getParent()
  */
 class Asset extends Element\AbstractElement
 {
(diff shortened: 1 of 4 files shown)


models/Asset.php
```diff
@@ -215,6 +209,20 @@ class Asset extends Element\AbstractElement
      */
     protected $versionCount = 0;
 
+    /**
+     * @param Asset|null $parent
+     *
+     * @return Asset
+     */
+    public function setParent($parent)
```
fashxp: please just restore the old implementation of `setParent` instead of changing it ...

models/Document.php
```diff
@@ -189,6 +183,20 @@ class Document extends Element\AbstractElement
      */
     protected $versionCount = 0;
 
+    /**
+     * @param Document|null $parent
+     *
+     * @return Document
+     */
+    public function setParent($parent)
```
fashxp: please just restore the old implementation of setParent instead of changing it ...

models/Element/ElementInterface.php
```diff
@@ -23,7 +23,7 @@
 use Pimcore\Model\Version;
 
 /**
- * @method static setParent(?ElementInterface $parent)
+ * @method ?ElementInterface setParent(?ElementInterface $parent)
```
fashxp: why change that?

models/Element/ElementInterface.php
```diff
@@ -209,7 +209,7 @@ public function getLocked();
     public function getParentId();
 
     /**
-     * @return self|null
+     * @return static|null
```
fashxp: why change that?

models/Element/AbstractElement.php
```diff
@@ -116,6 +123,21 @@ public function setProperty($name, $type, $data, $inherited = false, $inheritabl
         return $this;
     }
 
+    /**
+     * @return static|null
```
fashxp: shouldn't that be `self`?

fashxp (changes requested):
still some changes needed ... if changing return type of `getParent` to `self` doesn't make phpstan happy (or introduces new errors), I guess we have to introduce a `getParent` method to `Asset`, `Document` and `AbstractObject` as follows (or with even more precise return type definition...): 

```php
    /**
     * @return self|null
     */
    public function getParent() /*: self */
    {
        return parent::getParent();
    }
```

models/Element/AbstractElement.php
```diff
@@ -527,7 +529,7 @@ public function getVersions()
     public function __sleep()
     {
         $parentVars = parent::__sleep();
-        $blockedVars = ['dependencies', 'parent'];
+        $blockedVars = ['dependencies'];
```
fashxp: why was that removed again?
I guess that is also the reason for the failing tests....

fashxp (changes requested):

models/Element/AbstractElement.php
aashan10: Yes, I was working on the other `__sleep` and `__wakeup` method refactoring and I might have mixed up those two. Sorry, fixed it.

fashxp (approved):

models/Element/AbstractElement.php
```diff
@@ -82,6 +82,28 @@ public function setId($id)
...
+        if ($this->parent === null) {
+            $parent = Service::getElementById(Service::getElementType($this), $this->getParentId());
+            $this->setParent($parent);
+        }
+
+        return $this->parent;
+    }
+
+    public function setParentId($parentId) {
+        return $this;
```
fashxp: parent id is not set here ... or why add this method here at all?

fashxp (changes requested):

fashxp (approved):

PR #12113, opened 2022-05-12
Translate workflow transition notes title and titles of notes list of object

Pierre-Datasolution:
When using translation keys on workflow transition titles, translations are correctly displayed in workflow actions menu 
image

but not in transition note window title :
image

neither in the list of notes of the object :
image

Proposal in the MR is to translate transition note window title
image

and tile of each notes when browsing list of of notes of an object (translation key is still stored in database to be able to show an adapted translated title in each available language, no change)
image


aryaantony92:
@Pierre-Datasolution  Could you please rebase the PR to 10.4 since this is a bug ? Thanks.

PR #12116, opened 2022-05-13
Add CSP to upgrade notes

ultramegatom:
Might break some custom backend dashboards.


brusch:
@ultramegatom but you need to enable that feature, see: 
https://github.com/pimcore/pimcore/pull/11447/files#diff-426a4264cf4493c4fc1bb152fc2fcdeea5398e59e4

So actually the upgrade should be smooth and no need to change anything. 
Did you experience issues? 
Thanks for sharing.

ultramegatom:
ahh nvm then

Was testing something on the demo and its active there 

Didnt see the config: https://github.com/pimcore/demo/blob/f780cf17e0b294b8a70c7e2c69993329e74cd28f/config/config.yaml#L42 and assumed it was default sry

PR #12123, opened 2022-05-16
[Db] remove all usages of Pimcore\Db\ConnectionInterface and Pimcore\Db\PimcoreExtensionstrait

### LINKED ISSUE #12363: [Improvement]: Replace all deprecated \Doctrine\DBAL\Connection methods & constants usages
`Doctrine\DBAL\Connection` has some deprecated methods and constants, which shouldn't be used anymore.

dpfaffenbauer:
Version 2 of https://github.com/pimcore/pimcore/pull/10653

Resolves: #12363


brusch:
@dpfaffenbauer have you seen the issues reported by PhpStan?

dpfaffenbauer:
@brusch that is intentionally for now, have you seen my gitter message?

brusch:
@dpfaffenbauer 👍go ahead 😉

brusch:
So the approach is now to change the return types in `Pimcore\Db` first, so that we can then fix all issues (commit by commit) with the help of our lovely PhpStan. As soon as all issues are fixed, we change the return types back for compatibility.

brusch:
@dpfaffenbauer just a quick related info: @aashan10 will work on #12363 in the meantime - so that ones can be left out here. 

Another topic that came to my mind, how to deal with methods like `insertOrUpdate()` without making the application code more complex 🤔

dpfaffenbauer:
@brusch I had the inserOrUpdate in the old PR already done... With a Helper class basically.

dpfaffenbauer:
@brusch I think it would be good if @aashan10 makes changes in this PR, otherwise the merge could be quite difficult, what do you think?

brusch:
@dpfaffenbauer should be safe to do them in separate PRs and then after they got merged we merge 10.x back into this one. We need to keep the PRs easy to review, and as @aashan10 's tasks are fixing Doctrine deprecations  it's also not directly related to this PR 😉

dpfaffenbauer:
@brusch Changes made, tests are running. I'll revert the Db changes now to still return the Pimcore ConnectionInterface

bundles/CoreBundle/Migrations/Version20211103055110.php
```diff
@@ -50,7 +50,7 @@ public function down(Schema $schema): void
     {
         $db = \Pimcore\Db::get();
 
-        $classes = $db->fetchRow('SELECT id FROM classes');
+        $classes = $db->fetchAssociative('SELECT id FROM classes');
```
brusch, line 53: That's unfortunalely not returning the same result. 
`fetchAssociative()` is equal to `fetchAll()` but `fetchRow()` is only returning the first row using indices.

brusch (changes requested):

brusch:
@dpfaffenbauer cool thanks! 
I think we now need to add the reported missing methods on `ConnectionInterface` as `@method` annotations to it, to get rid of the PhpStan errors, right?

dpfaffenbauer:
@brusch on it ;)

dpfaffenbauer:
@brusch, all done, tests are green :)

aashan10 (approved):

lib/Db.php
```diff
@@ -39,11 +40,11 @@ public static function reset()
...
 
     /**
-     * @return Connection
+     * @return ConnectionInterface|Connection
      */
     public static function get()
     {
-        /** @var Connection $db */
+        /** @var ConnectionInterface|Connection $db */
         $db = \Pimcore::getContainer()->get('database_connection');
```
brusch:
```suggestion
        $db = \Pimcore::getContainer()->get('doctrine.dbal.default_connection');
```
We should use `doctrine.dbal.default_connection` here as well, to be consistent, right?

lib/Db.php
```diff
@@ -16,20 +16,21 @@
 namespace Pimcore;
 
 use Pimcore\Db\Connection;
```
brusch, line 18:
```suggestion
use Doctrine\DBAL\Connection;
```
Here we could use `Doctrine\DBAL\Connection` already, as it would be redundant by adding `ConnectionInterface` right?

bundles/AdminBundle/EventListener/ImportConfigListener.php
```diff
@@ -66,7 +66,7 @@ public function onUserDelete($event)
 
         $db = Db::get();
 
-        $importConfigIds = $db->fetchCol('select id from importconfigs where ownerId = ' . $userId);
+        $importConfigIds = $db->fetchFirstColumn('select id from importconfigs where ownerId = ' . $userId);
```
brusch:
```suggestion
        $importConfigIds = $db->fetchFirstColumn('select id from importconfigs where ownerId = ?', [$userId]);
```

brusch:
Seems that PhpStorm needs the `@deprecated` on every single method on the interface and trait ... maybe because he gets confused because of the way it's defined ... anyway, I've added the annotations and it seems fine now 😊

bundles/InstallBundle/Installer.php
```diff
@@ -26,7 +26,6 @@
 use Pimcore\Config;
 use Pimcore\Console\Style\PimcoreStyle;
 use Pimcore\Db\Connection;
```
brusch:
```suggestion
use Doctrine\DBAL\Connection;
```

lib/Model/Listing/AbstractListing.php
```diff
@@ -466,7 +467,7 @@ public function escapeLike(string $value): string
     {
         $db = Db::get();
```
brusch:
```suggestion
```
This is obsolete 😊

models/DataObject/AbstractObject/Dao.php
```diff
@@ -661,7 +662,7 @@ public function getChildPermissions($type, $user, $quote = true)
             }
 
             $cid = $this->model->getId();
-            $sql = 'SELECT ' . $type . ' FROM users_workspaces_object WHERE cid != ' . $cid . ' AND cpath LIKE ' . $this->db->quote($this->db->escapeLike($this->model->getRealFullPath()) . '%') . ' AND userId IN (' . implode(',', $userIds) . ') ORDER BY LENGTH(cpath) DESC';
+            $sql = 'SELECT ' . $type . ' FROM users_workspaces_object WHERE cid != ' . $cid . ' AND cpath LIKE ' . $this->db->quote(Helper::escapeLike($this->model->getRealFullPath()) . '%') . ' AND userId IN (' . implode(',', $userIds) . ') ORDER BY LENGTH(cpath) DESC';
             $permissions = $this->db->fetchAll($sql);
```
brusch:
```suggestion
            $permissions = $this->db->fetchAllAssociative($sql);
```

brusch (changes requested):

brusch:
@dpfaffenbauer we are nearly there 😊 Please see my feedback & quickly resolve the conflicts. Thanks again!

lib/Db.php
dpfaffenbauer: ConnectionInterface does not have all the methods as Connection has. For BC reasons we need both here.

dpfaffenbauer:
@brusch all done

lib/Db.php
brusch: But shouldn't we include `Doctrine\DBAL\Connection` in the `@return` as well, as it has additional methods which are not covered by `Pimcore\Db\ConnectionInterface` and `Pimcore\Db\Connection` ?

brusch: @dpfaffenbauer have you noticed my above comment? 😊

brusch:
@dpfaffenbauer #12462 caused another conflict - may I ask you again - Sorry & Thanks ☺️

dpfaffenbauer:
@brusch done

brusch:
@dpfaffenbauer Thanks a lot - great work! 🙋

dpfaffenbauer:
💪

brusch:
Seems this is causing an issue now on install: 
```
Couldn't establish connection to MySQL: The given 'wrapperClass' Doctrine\DBAL\Connection has to be a subtype of \Doctrine\DBAL\Connection.
```

Just for reference. Needs further investigation.

PR #12125, opened 2022-05-16
Refactor duplicate __sleep and __wakeup methods

aashan10:
partial fix for #8917

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1842,22 +1842,6 @@ class Asset extends Element\AbstractElement
         return $this;
     }
 
-    public function __sleep()
-    {
-        $parentVars = parent::__sleep();
-        $blockedVars = ['scheduledTasks', 'hasChildren', 'versions', 'parent', 'stream'];
-
-        if ($this->isInDumpState()) {
-            // this is if we want to make a full dump of the asset (eg. for a new version), including children for recyclebin
-            $this->removeInheritedProperties();
-        } else {
-            // this is if we want to cache the asset
-            $blockedVars = array_merge($blockedVars, ['children', 'properties']);
-        }
-
-        return array_diff($parentVars, $blockedVars);
-    }
-
     public function __wakeup()
     {
         if ($this->isInDumpState()) {
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -216,9 +216,21 @@ class Asset extends Element\AbstractElement
     protected $versionCount = 0;
 
     /**
-     * @var array
+     * @return array
      */
-    protected array $blockedVars = ['scheduledTasks', 'hasChildren', 'versions', 'parent', 'stream', 'children', 'properties'];
+    protected function getBlockedVars(): array
+    {
+        $blockedVars = ['scheduledTasks', 'hasChildren', 'versions', 'parent', 'stream', 'children', 'properties'];
+
+        if(!$this->isInDumpState()) {
+
+            // for caching asset
+            $blockedVars = array_merge($blockedVars, ['children', 'properties']);
+
+        }
+
+        return $blockedVars;
+    }
 
     /**
      *
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1885,4 +1885,26 @@ class Asset extends Element\AbstractElement
             new AssetUpdateTasksMessage($this->getId())
         );
     }
+
+    public function __wakeup()
+    {
+        if ($this->isInDumpState()) {
+            // set current key and path this is necessary because the serialized data can have a different path than the original element ( element was renamed or moved )
+            $originalElement = static::getById($this->getId());
+
+            if ($originalElement && !self::$doNotRestoreKeyAndPath) {
+
+                // set parentId and path for Asset
+                $this->setParentId($originalElement->getParentId());
+                $this->setPath($originalElement->getRealPath());
+
+            }
+        }
+
+        if ($this->isInDumpState() && $this->properties !== null) {
+            $this->renewInheritedProperties();
+        }
+
+        $this->setInDumpState(false);
+    }
 }
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -1716,6 +1716,24 @@ class Asset extends Element\AbstractElement
         return $this;
     }
 
+    public function __wakeup()
+    {
+        if ($this->isInDumpState()) {
+            // set current parent and path, this is necessary because the serialized data can have a different path than the original element (element was moved)
+            $originalElement = Asset::getById($this->getId());
+            if ($originalElement) {
+                $this->setParentId($originalElement->getParentId());
+                $this->setPath($originalElement->getRealPath());
+            }
+        }
+
+        if ($this->isInDumpState() && $this->properties !== null) {
+            $this->renewInheritedProperties();
+        }
+
+        $this->setInDumpState(false);
+    }
+
     public function __destruct()
     {
         // close open streams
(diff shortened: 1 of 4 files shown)


models/Element/AbstractElement.php
```diff
@@ -479,12 +487,51 @@ public function getVersions()
...
+
         $parentVars = parent::__sleep();
-        $blockedVars = ['dependencies'];
+
+        $type = Service::getElementType($this);
+
+        // this should be removed when the o_ prefix will be removed from data objects
+        $hasPrefix = in_array($type, ['document', 'asset']);
+
+        $blockedVars = match ($type) {
```
fashxp: I would add a property to the `AbstractElement` class which is overwritten for every subclass then ... 


In `AbstractElement`: 
```php
protected $blockedVars = [];
```

In `Document`: 
```php
protected $blockedVars = ['hasChildren', 'versions', 'scheduledTasks', 'parent', 'fullPathCache'];
```

models/Document.php
```diff
@@ -1272,22 +1272,6 @@ public function setParent($parent)
...
-            $this->removeInheritedProperties();
-        } else {
-            // this is if we want to cache the object
-            $blockedVars = array_merge($blockedVars, ['children', 'properties']);
-        }
-
-        return array_diff($parentVars, $blockedVars);
-    }
-
     public function __wakeup()
```
fashxp: is there a reason why this method is still here?

models/Element/AbstractElement.php
fashxp: or maybe we even should introduce a protected method `getBlockedVars` that also returns `(o_)children` etc.

fashxp (changes requested):

models/Element/AbstractElement.php
aashan10: I have added a protected variable `$blockedVars` and then added a method `getBlockedVars` for all the logic

models/Document.php
aashan10: no, had missed this one for some reason. I've fixed and updated

models/Asset.php
```diff
@@ -215,6 +215,11 @@ class Asset extends Element\AbstractElement
      */
     protected $versionCount = 0;
 
+    /**
+     * @var array
+     */
+    protected array $blockedVars = ['scheduledTasks', 'hasChildren', 'versions', 'parent', 'stream', 'children', 'properties'];
+
```
fashxp:
```suggestion
    /**
     * @return array
     */
    protected function getBlockedVars(): array
    {

        $blockedVars = ['scheduledTasks', 'hasChildren', 'versions', 'parent', 'stream'];

        if (! $this->isInDumpState()) {
            // this is if we want to cache the asset
            $blockedVars = array_merge($blockedVars, ['children', 'properties']);
        }
        
        return $blockedVars; 

    }
```

models/DataObject/AbstractObject.php
```diff
@@ -225,6 +218,11 @@ abstract class AbstractObject extends Model\Element\AbstractElement
      */
     protected $o_versionCount = 0;
 
+    /**
+     * @var array
+     */
+    protected array $blockedVars = ['o_hasChildren', 'o_parent', 'scheduledTasks', 'o_versions', 'omitMandatoryCheck', 'dependencies'];
+
```
fashxp:
```suggestion
    /**
     * @return array
     */
    protected function getBlockedVars(): array
    {
        $blockedVars = ['o_hasChildren', 'o_versions', 'o_class', 'scheduledTasks', 'o_parent', 'omitMandatoryCheck'];

        if ($this->isInDumpState()) {
            // this is if we want to make a full dump of the object (eg. for a new version), including children for recyclebin
            $blockedVars = array_merge($blockedVars, ['o_dirtyFields']);
        } else {
            // this is if we want to cache the object
            $blockedVars = array_merge($blockedVars, ['o_children', 'properties']);
        }
        
        return $blockedVars; 
    }

```

models/Document.php
```diff
@@ -189,6 +189,11 @@ class Document extends Element\AbstractElement
      */
     protected $versionCount = 0;
 
+    /**
+     * @var array
+     */
+    protected array $blockedVars = ['hasChildren', 'parent', 'scheduledTasks', 'versions', 'fullPathCache'];
+
```
fashxp:
```suggestion
    /**
     * @return array
     */
    protected function getBlockedVars(): array
    {
        $blockedVars = ['hasChildren', 'versions', 'scheduledTasks', 'parent', 'fullPathCache'];

        if (! $this->isInDumpState()) {
            // this is if we want to cache the object
            $blockedVars = array_merge($blockedVars, ['children', 'properties']);
        }
        
        return $blockedVars; 
    }

```

models/Element/AbstractElement.php
```diff
@@ -55,6 +56,19 @@ abstract class AbstractElement extends Model\AbstractModel implements ElementInt
      */
     protected ?array $properties = null;
 
+    /**
+     * @internal
+     * @var array
+     */
+    protected array $blockedVars = [];
```
fashxp:
```suggestion

```

models/Element/AbstractElement.php
```diff
@@ -474,15 +488,49 @@ public function getVersions()
...
+            $blockedVars = [...$this->blockedVars, $hasPrefix ? 'o_dirtyFields' : 'dirtyFields'];
+            $this->removeInheritedProperties();
+
+            return $blockedVars;
+        }
+
+        return [...$this->blockedVars, $hasPrefix ? 'o_children' : 'children', 'properties'];
+
+    }
+
```
fashxp:
```suggestion
```

models/Element/AbstractElement.php
```diff
@@ -474,15 +488,49 @@ public function getVersions()
...
     /**
      * {@inheritdoc}
      */
     public function __sleep()
     {
-        $parentVars = parent::__sleep();
-        $blockedVars = ['dependencies'];
+        return array_diff(parent::__sleep(), $this->getBlockedVars());
```
fashxp, line 687:
```suggestion

        if ($this->isInDumpState()) {
            // this is if we want to make a full dump of the object (eg. for a new version), including children for recyclebin
            $this->removeInheritedProperties();
        }

        return array_diff(parent::__sleep(), $this->getBlockedVars());
```

models/Asset.php
```diff
@@ -1842,40 +1847,6 @@ public function setParent($parent)
...
-            $this->removeInheritedProperties();
-        } else {
-            // this is if we want to cache the asset
-            $blockedVars = array_merge($blockedVars, ['children', 'properties']);
-        }
-
-        return array_diff($parentVars, $blockedVars);
-    }
-
-    public function __wakeup()
```
fashxp, line 1718: we cannot remove that one, because this is different to the one in the `AbstractElement`

fashxp (changes requested):
Hmm, the introduced changes were not exactly what I meant and also would change behavior ... please have a look at my suggestions below... 

please also resolve the conflicts. 

thx

models/Element/AbstractElement.php
aashan10: I have made changes according to the review

models/Asset.php
aashan10: This has been changed

models/Element/AbstractElement.php
```diff
@@ -662,15 +662,40 @@ public function getVersions()
         return [];
     }
 
+    protected function getBlockedVars(): array
+    {
+        return ['dependencies'];
```
fashxp: 'parent' got lost now again here?

fashxp (changes requested):

fashxp (approved):

PR #12127, opened 2022-05-16
Update treenodelocator.js

Daniel-Ateles:
"Show in Tree" in the Asset-tree having paginated folders doesn't work because the direction i calculated wrong.
Steps to reproduce: Have an assetfolder with 400 assets and a folder. Search for an asset that resides in the end of the asset folder containing 400 assets.
Press "Show in Tree" on the searched asset and the bug is that the asset is not selected and Pimcore stays on the first page.

  
  

Resolves #


dvesh3:
@Daniel-Ateles thanks for your PR! could you please rebase it on 10.4 branch? as it is a bug fix.

Daniel-Ateles:
> @Daniel-Ateles thanks for your PR! could you please rebase it on 10.4 branch? as it is a bug fix.

@dvesh3 Did I rebase correctly? The PR now seems to have lot's of commits in it

PR #12128, opened 2022-05-16
[Documents] Reset draft changes on applying Content Master Document

### LINKED ISSUE #11918: [Bug]: Preview doesn´t get updated, after reapplying Content Master Document
### Expected behavior

Update preview

### Actual behavior

Preview doesn´t get updated

### Steps to reproduce

1.) Create a new page
2.) Apply Content Master Document in settings
3.) Overwrite one of the inherited fields
4.) Check preview tab -> preview should be fine
5.) Reapply Content Master Document 
6.) Check and compare edit / preview tab

dvesh3:
Resolves #11918

## Additional info  
- On applying content master in Document Settings, the document is reloaded with dirty changes and if `autoSave` feature is turned on (default behavior), then it saves the overwrites the changes from content master. The issue fixed by calling resetChanges() to reset the dirty changes before calling reload().

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/settings_abstract.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/settings_abstract.js
@@ -140,6 +140,7 @@ pimcore.document.settings_abstract = Class.create({
                                                     "contentMasterDocumentPath_" + this.document.id).getValue()
                                             },
                                             success:function () {
+                                                this.document.resetChanges();
                                                 this.document.reload();
                                             }.bind(this)
                                         });


mcop1 (commented):
@dvesh3 The PR does reset the changes, this part works fine. But after resetting, the "Draft" button in the top right still remains there. And I have to click it twice to get rid of it. It seems that it somehow creates 2 draft versions?

dvesh3:
> @dvesh3 The PR does reset the changes, this part works fine. But after resetting, the "Draft" button in the top right still remains there. And I have to click it twice to get rid of it. It seems that it somehow creates 2 draft versions?

@mcop1 the draft button shows up because it saves a version, when you apply master document. please see https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Controller/Admin/Document/DocumentControllerBase.php#L355

mcop1:
> > @dvesh3 The PR does reset the changes, this part works fine. But after resetting, the "Draft" button in the top right still remains there. And I have to click it twice to get rid of it. It seems that it somehow creates 2 draft versions?
> 
> @mcop1 the draft button shows up because it saves a version, when you apply master document. please see https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Controller/Admin/Document/DocumentControllerBase.php#L355

Ahh ok, makes sense, thanks for pointing that out 👍

PR #12135, opened 2022-05-16
[Bug]: Custom View Classes not working if is not associative array

### LINKED ISSUE #12134: [Bug]: Custom Views are not working
### Expected behavior

custom view working

### Actual behavior

the folder show nothing even though it has a plus icon and content 
image

It is very similar to some permissions problem that i've recently fixed, but it is unrelated to that.

### Steps to reproduce

log into demo and use a perspective that uses custom view eg. `PIM` Cars

related to https://github.com/pimcore/pimcore/pull/7143

image
changing `$key` to `$cvClass` works, but it is not the actual fix. 
As the changes in the PR mentioned above, the $cvClasses (circled in screenshot) should match the "new" format

kingjia90:
fixes #12134

## Additional info  
the custom view's filter `classes` uses the array key of an associative array since #7143 and X, but in the demo, the `pimcore_custom_views` in the setting store table are using a sequential array and this change fixes it.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -202,8 +202,9 @@ class DataObjectController extends ElementControllerBase implements KernelContro
             if (!empty($cv['classes'])) {
                 $cvConditions = [];
                 $cvClasses = $cv['classes'];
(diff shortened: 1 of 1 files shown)


kingjia90:
@fashxp in case this would be reopen, we can get an associative array by checking the first element of the array has index/key is 0, because id:0, name:0 are not allowed in the object class creation.

PR #12146, opened 2022-05-18
Migrated parent property and setParent method

aashan10:
Continuation of previous #12110

### DIFF
--- a/models/DataObject/AbstractObject.php
+++ b/models/DataObject/AbstractObject.php
@@ -107,7 +107,7 @@ abstract class AbstractObject extends Model\Element\AbstractElement
     /**
      * @internal
      *
-     * @var self|null
+     * @deprecated
      */
     protected $o_parent;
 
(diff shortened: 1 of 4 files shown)


models/DataObject/AbstractObject.php
```diff
@@ -1230,11 +1231,7 @@ public function getParent()
      */
     public function setParent($o_parent)
     {
-        $newParentId = $o_parent instanceof self ? $o_parent->getId() : 0;
-        $this->setParentId($newParentId);
-        $this->o_parent = $o_parent;
-
-        return $this;
+        return parent::setParent($o_parent);
     }
```
jdreesen, line 1235: Shouldn't it be enough to annotate the class with `@method $this setParent(self|null $parent)`?

fashxp: not, when we start using php type hints, right?

PR #12150, opened 2022-05-18
[Bug]: Perspective Editor unexpected behavior 

### LINKED ISSUE #12149: [Bug]: Perspective Editor unexpected behavior
### Expected behavior

Ability to create perspectives and custom views, and assign different content to each view. 

### Actual behavior

When trying to create a new perspective and adding custom views to it: 
- Perspective is created, but the custom views in the new perspective are not shown as they should. 
- Sometimes the folder itself is not showing at all, sometimes the folder structure is showing, but the products are hidden even though there is no filter activated for classes. 
- First seen on a customer project, recreated on demo.pimcore.com
This is what the perspective looks like:
Bildschirmfoto 2022-05-12 um 09 22 33

Bildschirmfoto 2022-05-12 um 09 22 25

This is what the configuration looks like(using custom views included in demo): 

Bildschirmfoto 2022-05-12 um 09 22 48

Bildschirmfoto 2022-05-12 um 09 23 20

Bildschirmfoto 2022-05-12 um 09 23 14

### Steps to reproduce

Create a new perspective in the perspective editor and try to add custom views to it.
Go to the newly created view and see wrong behaviour. 
Happened on 10.3.5 and on 10.4.1.

mcop1:
Resolves #12149


lib/CustomView/Config.php
```diff
@@ -69,6 +69,25 @@ private static function getRepository()
...
+            return [];
+        }
+        else {
+            $flipArray = [];
+            $tempClasses = explode(',', $data['classes']);
+
+            foreach ($tempClasses as $tempClass) {
+                $flipArray[$tempClass] = null;
+            }
+            return $flipArray;
```
NiklasBr: I think [this one](https://www.php.net/manual/en/function.array-fill-keys.php) could work:

```suggestion
            $tempClasses = explode(',', $data['classes']);
            return array_fill_keys($tempClasses, null);
```

mcop1: It does, neat, thanks!

fashxp (approved):

PR #12158, opened 2022-05-18
Refactored duplicate o_key/key property and related methods

aashan10:
Partially Resolves #8917


fashxp:
lets skip that for now as assets don't have a key ... and I also don't want to add it as it won't be stored to the DB ... with assets it is called filename.

brusch:
@fashxp Should we think about renaming `filename` to `key` on assets as well? We can still offer `getFilename()` just the other way round as it is now?

PR #12161, opened 2022-05-19
[Asset upload] Prevent creation of __MACOS folder when uploading a zip file which got created with MacOS

BlackbitDevs:
Zip files which get generated in MacOS usually contain a folder `__MACOS`. When you upload such a zip to Pimcore assetsm you will end up with the actual files and also the (imho useless) `__MACOS` folder:
Bildschirmfoto 2022-05-19 um 07 54 58

With this PR only the real files will get imported, the `__MACOS` folder will get ignored.


dvesh3:
@BlackbitNeueMedien thanks!

PR #12183, opened 2022-05-23
Added getters specifically for DataObjects

aashan10:
Fixes reference bug discovered when migrating properties of DataObjects


aashan10:
Closing this because we can do this inside `__wakeup` method for less clutter

PR #12188, opened 2022-05-23
Fix: Localizedfield DAO improperly checks the DBMS error message

### LINKED ISSUE #12168: [Bug]: Localizedfield DAO improperly checks the DBMS error message
### Expected behavior

The missing table is properly detected, for example by inspecting the error code provided by MySQL.

### Actual behavior

When using the Localizedfield DAO, it will check for the error message, if it contains the string `exist`, it will assume that means the table doesn't exists and try to create it, possibly creating another error (because the table actually exists).

https://github.com/pimcore/pimcore/blob/6dec4d51e44886db546b49e923bae01d33878b78/models/DataObject/Localizedfield/Dao.php#L243

### Steps to reproduce

1. create a class with localized `Slug`, `Description`, make Slug unique
2. save one object with slug `aaa`
3. save another object with slug `aaa`, description `un proceso manual existente` (Spanish)
4. you get an error ` missing table created, start next run ... ;-)` (Pimcore is winking in all these error messages for some reason), and then a message related to no active transaction, etc
5. the underlying error is actually similar to
```
An exception occurred while executing 'INSERT INTO `object_localized_data_5... (the whole INSERT here, including the *values*)

SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'aaa' for key 'object_localized_data_5.u_index_Slug'
```
6. Pimcore finds`exist` in `existente` and assumes the table is missing.

blankse:
Fix https://github.com/pimcore/pimcore/issues/12168

### DIFF
--- a/models/DataObject/Localizedfield/Dao.php
+++ b/models/DataObject/Localizedfield/Dao.php
@@ -238,23 +238,19 @@ class Dao extends Model\Dao\AbstractDao
                     )) {
                     $this->db->insertOrUpdate($storeTable, $insertData);
                 }
-            } catch (\Exception $e) {
+            } catch (TableNotFoundException $e) {
                 // if the table doesn't exist -> create it! deferred creation for object bricks ...
-                if (strpos($e->getMessage(), 'exist')) {
-                    try {
-                        $this->db->rollBack();
-                    } catch (\Exception $er) {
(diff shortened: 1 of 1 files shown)


robertSt7:
Thanks for the PR

PR #12190, opened 2022-05-23
Refactored duplicate o_parentId/parentId properties and methods

aashan10:
partly fix of #8917


models/Element/AbstractElement.php
```diff
@@ -137,6 +137,40 @@ public function setPath($path)
...
+    /**
+     * @param int $parentId
+     * @return $this
+     */
+    public function setParentId($parentId)
+    {
+        $parentId = (int) $parentId;
+        $type = Service::getElementType($this);
+
+        if ($parentId != $this->parentId && $type === 'object') {
```
fashxp: please avoid that kind of ifs in the parent classes ... if there is specific functionality of a subclass, then it should go into the subclass...

fashxp (changes requested):

models/Element/AbstractElement.php
aashan10: Fixed by overriding the `setParentId` method in `AbstractObject` class

fashxp:
@aashan10  can you please merge 10.x once more?

fashxp (approved):

Issue #12151, opened 2022-05-18
[Bug]: Exception when going to navigation properties


GAL-GC:
### Expected behavior

When creating or editing a document, I want to set navigation settings, for example the document language or exclude it from navigation.

### Actual behavior

It happened, that an exception occured, when I went to Navigation & properties tab.

```
Status: 500 | Internal Server Error
URL: /admin/page/save?task=autoSave
Method: PUT
Message: Pimcore\Bundle\AdminBundle\Controller\Admin\Document\DocumentControllerBase::getPropertyInheritance(): Argument #3 ($propertyValue) must be of type string, null given, called in /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/Document/DocumentControllerBase.php on line 138
Trace: 
in /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/Document/DocumentControllerBase.php:306
#0 /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/Document/DocumentControllerBase.php(138): Pimcore\Bundle\AdminBundle\Controller\Admin\Document\DocumentControllerBase->getPropertyInheritance(Object(Pimcore\Model\Document\Page), 'language', NULL)
#1 /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/Document/PageController.php(446): Pimcore\Bundle\AdminBundle\Controller\Admin\Document\DocumentControllerBase->addPropertiesToDocument(Object(Symfony\Component\HttpFoundation\Request), Object(Pimcore\Model\Document\Page))
#2 /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/Document/DocumentControllerBase.php(445): Pimcore\Bundle\AdminBundle\Controller\Admin\Document\PageController->setValuesToDocument(Object(Symfony\Component\HttpFoundation\Request), Object(Pimcore\Model\Document\Page))
#3 /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/Document/PageController.php(155): Pimcore\Bundle\AdminBundle\Controller\Admin\Document\DocumentControllerBase->saveDocument(Object(Pimcore\Model\Document\Page), Object(Symfony\Component\HttpFoundation\Request))
[...]

PR #12195, opened 2022-05-24
Fix: Exception when going to navigation properties

blankse:
Fix: https://github.com/pimcore/pimcore/issues/12151


aryaantony92:
@blankse Thank you

PR #12200, opened 2022-05-24
[DataObject Tree] Allow expand and collapse on expandable property 

### LINKED ISSUE #12089: Do not show expand/collapse children on leaf element
### Expected behavior

This button is not supposed to appear on elements that have no children (leaf)
image
image

### Actual behavior

The button appears

### Steps to reproduce

1. Right click on leaf (no children) tree item.
2. Go to Advanced -> Collapse Children

Related to https://github.com/pimcore/pimcore/pull/11908

dvesh3:
Resolves #12089

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -303,6 +303,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
 
         if ($tmpObject['leaf']) {
             $tmpObject['expandable'] = false;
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -303,8 +303,8 @@ protected function getTreeNodeConfig($element): array
 
         if ($tmpObject['leaf']) {
             $tmpObject['expandable'] = false;
+            $tmpObject['leaf'] = false; //this is required to allow drag&drop
```
dvesh3, line 306: we have to use this due to ExtJS limitation, which doesn't allow drag&drop on leaf nodes.

PR #12202, opened 2022-05-24
Fix custom entrypoint login link

### LINKED ISSUE #12111: [Bug]: Login as user does not work with custom entrypoint
### Expected behavior

When copying the login as user link and opening the link in a new browser, you should be login as that user regardless of the admin entry point

### Actual behavior

If you have configured a custom admin entry point you get a 404 error as you cannot access the /admin route without being redirected from the custom entry point route

### Steps to reproduce

1. Configure a custom admin entry point as described here: https://pimcore.com/docs/pimcore/current/Development_Documentation/Development_Tools_and_Details/Custom_Admin_Login_Entry_Point.html
2. Create a second test user and click on "Login as this user in a different browser"
3. Open the copied URL in a new browser

It seems that replacing /admin/login/login with the custom entry point (i.e. /myadmin) works for custom entry points, but trying to do that by removing /login/login so you only have the default admin entry point (/admin) does not work.

matthiashamacher:
Resolves #12111


bundles/AdminBundle/Controller/Admin/UserController.php
```diff
@@ -1238,8 +1238,10 @@ protected function getUserId(Request $request)
     private function generateCustomUrl(array $params, $fallbackUrl = 'pimcore_admin_login_check', $referenceType = UrlGeneratorInterface::ABSOLUTE_URL): string
     {
         try {
+            $adminEntryPointRoute = $this->getParameter('pimcore_admin.custom_admin_route_name');
+
             //try to generate invitation link for custom admin point
-            $loginUrl = $this->generateUrl('my_custom_admin_entry_point', $params, $referenceType);
+            $loginUrl = $this->generateUrl($adminEntryPointRoute, $params, $referenceType);
```
dvesh3, line 1244: this will be a BC break if you have already defined a route with name `my_custom_admin_entry_point`

dvesh3: I would use `my_custom_admin_entry_point` as default value for BC reasons

matthiashamacher: Yeah you're right. Haven't thought about that. I've changed it back to `my_custom_admin_entry_point`

dvesh3 (approved):

dvesh3:
@matthiashamacher I just realized the base branch is wrong as this is a fix. please rebase it on 10.4 branch. thanks!

matthiashamacher:
@dvesh3 I've changed the base branch. I already developed it on 10.4 but forgot the change the branch while creating the PR...

dvesh3:
@matthiashamacher thanks a lot!

PR #12203, opened 2022-05-24
[Task] Add @param and @return php doc types to /lib/Cache

mcop1:
Resolves https://github.com/pimcore/planning/issues/69


mcop1:
Checked 

lib/Cache.php 
lib/Cache/Runtime.php 
lib/RuntimeCacheTrait.php

and there is nothing to do. Rest is marked as internal.

PR #12204, opened 2022-05-24
[Reports] Regard drilldown filters on export

BlackbitDevs:
Steps to reproduce bug:

1. Create 2 data object classes `Product` and `Category`
2. Create a custom SQL report with the query `SELECT id,name FROM classes`
    For field `name` set `Filter Drilldown`=`filter_and_show`
3. Open report, filter for `Name`=`Product` with drilldown filter -> only 1 row is shown
4. Click `Export as CSV` button

Expected behaviour: Export the filtered report
Actual behaviour: Unfiltered report gets exported

With this PR the drilldown filters get used also for export.


fashxp:
thx very much!

PR #12207, opened 2022-05-24
Added latestVersion to forbiddenNames

blankse:
Followup to https://github.com/pimcore/pimcore/pull/12078


dvesh3:
@blankse this was my first thought when I reviewed the other PR. This is not an issue with ClassDefinition name only the attribute name.

blankse:
@dvesh3 This is used for attribute name.

dvesh3:
> @dvesh3 This is used for attribute name.

Ohh, you are right. I have overseen that 😄 thanks!

dvesh3 (approved):

PR #12208, opened 2022-05-24
[User] Add tip for Disable unsaved content warning unused on auto save

### LINKED ISSUE #12176: [Bug]: Disable unsaved content warning no working
### Expected behavior

When the box is not checked it should alert the user if there are not saved changes when closing the tab.

### Actual behavior

It just saves the content and closes the tab.

### Steps to reproduce

- Uncheck "Disable unsaved content warning"
- Save user
- Change something in a document
- Close the tab without saving
- It just saves the content without asking first
image

dvesh3:
Resolves #12176

### DIFF
--- a/doc/Development_Documentation/18_Tools_and_Features/02_Auto_Save_Drafts.md
+++ b/doc/Development_Documentation/18_Tools_and_Features/02_Auto_Save_Drafts.md
@@ -12,3 +12,5 @@ pimcore:
     objects: 
         auto_save_interval: 60 # saving interval in seconds, default 60s, set to 0 to disable it
 ```
+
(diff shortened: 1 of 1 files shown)


aryaantony92:
@dvesh3  Suppose if we enable the auto_save option in case of the documents and we disable it for the objects, then we will not be able to change the changes/values.

dvesh3:
> @dvesh3 Suppose if we enable the auto_save option in case of the documents and we disable it for the objects, then we will not be able to change the changes/values.

Yes, you are right. 

The challenge is for auto save, we have individual configs and for unsaved warning, we just have one checkbox for both elements. I revert my changes as the current behavior is still in better shape as a user should be able to use the checkbox when auto save disabled for an element. I added a note in the docs section.

aryaantony92:
@dvesh3 Thanks

PR #12209, opened 2022-05-24
[Task] Add @param and @return php doc types to /models/Tool

kingjia90:
Resolves https://github.com/pimcore/planning/issues/64

### DIFF
--- a/models/Tool/CustomReport/Adapter/AbstractAdapter.php
+++ b/models/Tool/CustomReport/Adapter/AbstractAdapter.php
@@ -33,7 +33,7 @@ abstract class AbstractAdapter implements CustomReportAdapterInterface
      * @param \stdClass $config
      * @param Config|null $fullConfig
      */
-    public function __construct($config, $fullConfig = null)
+    public function __construct(\stdClass $config, ?Config $fullConfig = null)
     {
         $this->config = $config;
         $this->fullConfig = $fullConfig;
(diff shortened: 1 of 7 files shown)


kingjia90:
Closed because the task is just about fixing `@params` and there was nothing to do here

jdreesen:
Why don't you add the parameter types, too?

kingjia90:
> Why don't you add the parameter types, too?

We are currently only making some preparation work in 10.5, but they are going to be fully added in 11, as we noticed adding the type now can lead to some BC Break and because the type in the docblock is not always right

jdreesen:
Adding parameter types should not break BC... except if you have to change it, because the docblock was wrong, of course... I see.

fashxp:
It depends on the perspective ... symfony bc promise doesn't allow to add native php types to parameters of public methods. 
And there are really situations where adding types can break your code (even if docblock was correct already). 

That's, why we decided to revise docblocks now and add parameter and return types with Pimcore 11.

jdreesen:
Interesting. Out of curiosity, can you tell me in which cases it doesn't work when you add parameter types?

fashxp:
that all works ... if it does the proper thing is a different question, but your code runs. 

```php

  /**
  * @param int $bar
  **/
  public function foo($bar) {   
    var_dump($bar);   
  }

 foo(123);
 foo('123');
 foo('asd');
 foo('');
 foo(null);
```

changing to nativ php type hints results in following: 

```php

  public function foo(int $bar) {   
    var_dump($bar);   
  }

 foo(123); 
 foo('123'); 
 foo('asd'); //fatal
 foo('');    //fatal  
 foo(null);  //fatal
```

jdreesen:
Okay, so it only breaks if you use the API wrong...

fashxp:
Jup ... but it could break working code. 
I just don't want to get the blame and the blame :-)

PR #12218, opened 2022-05-25
Use DoctrineDbalAdapter cache adapter

blankse:
The PdoAdapter is [deprecated ](https://fossies.org/linux/symfony/UPGRADE-5.4.md) with DBAL Connection argument.


kingjia90:
Thank you for the PR! 
Looks good to me, only very minor things that i would add/change in this PR are documentation related: 
1) the commented config still referencing pdo [here ](https://github.com/pimcore/pimcore/tree/10.x/doc/Development_Documentation/19_Development_Tools_and_Details/09_Cache#configuring-the-cache) instead of `cache.adapter.doctrine_dbal`
2) A small description in the upgrade note of 10.5

blankse:
@kingjia90 I changed the docs.

PR #12248, opened 2022-05-25
[Task] Add @param and @return php doc types to /models/User

kingjia90:
Resolves https://github.com/pimcore/planning/issues/65

## Additional info  
Most setters in UserRole are `array|string` due $this->prepareArray() exploding string. 
It is also because from  `$user->getDao()->getById($id);`, empty values are not an array but an empty string.


models/User/AbstractUser.php
```diff
@@ -124,7 +124,7 @@ public function getId()
     }
 
     /**
-     * @param int $id
+     * @param int|null $id
```
fashxp: I think we should not allow `null` here (in the future) ... thus not adding `null` to the type hints.

models/User/AbstractUser.php
```diff
@@ -144,7 +144,7 @@ public function getParentId()
     }
 
     /**
-     * @param int $parentId
+     * @param int|null  $parentId
```
fashxp: I think we should not allow `null` here (in the future) ... thus not adding `null` to the type hints.

models/User/UserRole.php
```diff
@@ -150,7 +150,7 @@ public function getPermissions()
     }
 
     /**
-     * @param string $permissionName
+     * @param string|null $permissionName
```
fashxp: why should that we `null` here?

models/User/UserRole.php
```diff
@@ -187,7 +187,7 @@ public function generatePermissionList()
     }
 
     /**
-     * @param string|array $permissions
+     * @param string|array|null $permissions
```
fashxp: why should we allow `null` here?

fashxp: as if it is null, it won't change anthing which might be not expected behavior...

fashxp (changes requested):

fashxp (approved):

PR #12249, opened 2022-05-26
Added @param and @returns to docblocks in lib/Extension

### LINKED ISSUE #12223: [Task] Add @param and @return php doc types to /lib/Extension
### Improvement description

- [ ] Add @param php doc types for all public/protected classes/methods/properties in /lib/Extension (excl. Dao and @internal) 
- [ ] Add @return php doc types for all public/protected classes/methods/properties in /lib/Extension (excl. Dao and @internal) 

aashan10:
Resolves #12223


lib/Extension/Bundle/Installer/SettingsStoreAwareInstaller.php
```diff
@@ -142,18 +142,27 @@ public function uninstall()
...
     }
 
+    /**
+     * @return bool|float|int|string
+     */
     public function isInstalled()
     {
         $installSetting = SettingsStore::get($this->getSettingsStoreInstallationId(), 'pimcore');
 
         return $installSetting ? $installSetting->getData() : false;
```
fashxp: we should cast to bool here and just return bool values

fashxp (changes requested):

lib/Extension/Bundle/Installer/SettingsStoreAwareInstaller.php
```diff
@@ -142,18 +142,27 @@ public function uninstall()
...
+    /**
+     * @return bool
+     */
     public function canBeInstalled()
     {
         return !$this->isInstalled();
     }
 
+    /**
+     * @return bool|float|int|string
```
fashxp: sorry, missed that one before ... here we also should just allow bool as return value ...

lib/Extension/Bundle/Traits/PackageVersionTrait.php
```diff
@@ -35,7 +35,7 @@ trait PackageVersionTrait
     abstract protected function getComposerPackageName(): string;
 
     /**
-     * {@inheritdoc}
+     * @return array|string|string[]|null
```
fashxp: how can that be an array? Shouldn't that be just a string?

fashxp: I think not even null should be allowed there...

fashxp (changes requested):

lib/Extension/Bundle/Traits/PackageVersionTrait.php
```diff
@@ -49,7 +49,7 @@ public function getVersion()
     }
 
     /**
-     * {@inheritdoc}
+     * @return mixed|string
```
jdreesen: I think this should always be a `string`.

aashan10: Thanks! Changed that

lib/Extension/Bundle/Traits/PackageVersionTrait.php
aashan10: I saw that `preg_replace` has those return types that are possible. So, just used those. 
However, I realized that in this scenario, we might just be getting a string. So, changed.

Thanks.

lib/Extension/Bundle/Installer/SettingsStoreAwareInstaller.php
aashan10: Changed this one as well.

fashxp (approved):

Issue #10912, opened 2021-11-26
[Bug]: Areabrick autoloading is not compatible with the directory structure as described in Symfony's best practices for bundles


jdreesen:
### Expected behavior

Symfony proposes a directory structure as follows in its [Best Practices for Reusable Bundles](https://symfony.com/doc/current/bundles/best_practices.html#directory-structure):

```
/
├── config/
├── docs/
│   └─ index.md
├── public/
├── src/
│   ├── Controller/
│   ├── DependencyInjection/
│   └── AcmeBlogBundle.php
├── templates/
├── tests/
├── translations/
├── LICENSE
└── README.md
```

When creating a bundle with this structure, Pimcore should be able to autoload areabricks from this bundle (which would be located in `src/Document/Areabrick/...` with their templates in `templates/areas/...`).

### Actual behavior

With the above directory structure, the method `AcmeBlogBundle::getPath()` must be overridden to point to the actual bundle root (instead of `src`) [as described in the docs](https://symfony.com/doc/current/bundles/best_practices.html#directory-structure), otherwise Symfony/Twig is not able to locale the templates in the `templates/` folder.

But this means, that now Pimcore isn't able to locate the `Areabrick` classes anymore, because it looks in the wrong folder: it looks in `Document/Areabrick/` while the class is actually in `src/Document/Areabrick/`.

### Steps to reproduce

Create a bundle with a directory structure as described in the [Best Practices for Reusable Bundles](https://symfony.com/doc/current/bundles/best_practices.html#directory-structure) with an Areabrick named "Foo" in `src/Document/Areabrick/Foo.php` and its template in `templates/areas/foo.html.twig`.

Pimcore is not able to autoload the areabrick. 

It must be manually registered like this instead:

```yaml
services:
    MyBundle\Document\Areabrick\Foo:
        tags:
            - { name: pimcore.area.brick, id: foo }
```

jdreesen:
I'm working on this. Will provide a PR soon.

jdreesen:
There it is: #12356

PR #12256, opened 2022-05-27
[Improvement] use composer name for packagename

### LINKED ISSUE #6154: PackageVersionTrait getComposerPackageName default implementation
# Feature Request
When creating bundles, the extension manager may display the current version of the bundle (as per [documentation](https://pimcore.com/docs/5.x/Development_Documentation/Extending_Pimcore/Bundle_Developers_Guide/Pimcore_Bundles/index.html#page_Returning-the-composer-package-version-in-extension-manager)).

Returning the name in a harcoded fashion as shown in Pimcore's documentation (see above link) seems to be a _possible cause of error_ in case the bundle changes its name (or namespace).

For this reason, we propose a default implementation (see [below](#code-example)) of the `getComposerPackageName` method of the `PackageVersionTrait` within the `AbstractPimcoreBundle`.

### Code example

First, copy the `getVersion` method from `PackageVersionTrait` into `AbstractPimcoreBundle`. Additionally, change the first line to check whether a package-name is available at all.
```php
    public function getVersion()
    {
        $version = "";
        $packageName = $this->getComposerPackageName();

        if(!empty($packageName){
            $version = Versions::getVersion($packageName);

            // normalizes v2.3.0@9e016f4898c464f5c895c17993416c551f1697d3 to 2.3.0
            $version = preg_replace('/^v/', '', $version);
            $version = preg_replace('/@(.+)$/', '', $version);
        }
        return $version;
    }
```

Next, implement the `getComposerPackageName` method wihtin the `AbstractPimcoreBundle`, like so:

```php
    protected function getComposerPackageName(): string
    {
        $composerFile = __DIR__ . '/../composer.json';
        $property = "name";

        if(file_exists($composerFile)) {
            $composerConfig = file_get_contents($composerFile);
            $composerConfig = json_decode($composerConfig);
            if(property_exists($composerConfig, $property)) {
                return $composerConfig->{$property};
            }
        }

        return "";
    }
```
This will check if there is a `composer.json` in the root of the bundle. If so, it is parsed and the `name` defined in the file is returned.

Now, remove the `PackageVersionTrait`, as it is no longer needed, due to the default implementation.

> In case a different implementation is needed, one can still override the methods `getVersion`and `getComposerPackageName`.

Now, in case of renaming the bundle, the version will still be properly displayed.
> It also seems convenient to automatically display the version.

dpfahlbusch:
Use 'name' property of composer.json for packageVersionTrait

co-authored: @passioneight
  

Resolves #6154

### DIFF
--- a/lib/Extension/Bundle/Traits/PackageVersionTrait.php
+++ b/lib/Extension/Bundle/Traits/PackageVersionTrait.php
@@ -22,9 +22,8 @@ use PackageVersions\Versions;
 use Pimcore\Composer\PackageInfo;
 
 /**
- * Exposes a simple getVersion() implementation by looking up the installed versions via ocramius/package-versions
- * which is generated on composer install. This trait can be used by using it from a bundle class and implementing
- * getComposerPackageName() to return the name of the composer package to lookup.
+ * Exposes a simple getVersion() and getComposerPackageName() implementation by looking up the installed versions 
+ * via ocramius/package-versions which is generated on composer install.
  */
 trait PackageVersionTrait
 {
--- a/lib/Extension/Bundle/Traits/PackageVersionTrait.php
+++ b/lib/Extension/Bundle/Traits/PackageVersionTrait.php
@@ -27,12 +27,26 @@ use Pimcore\Composer\PackageInfo;
  */
 trait PackageVersionTrait
 {
+
     /**
      * Returns the composer package name used to resolve the version
      *
      * @return string
      */
-    abstract protected function getComposerPackageName(): string;
+    public function getComposerPackageName(): string {
+        $composerFile = __DIR__ . '/../composer.json';
+
+        if (file_exists($composerFile)) {
+            $composerConfig = file_get_contents($composerFile);
+            $composerConfig = json_decode($composerConfig);
+
+            if (property_exists($composerConfig, 'name')) {
+                return $composerConfig->name;
+            }
+        }
+
+        return '';
+    }
 
     /**
      * {@inheritdoc}
--- a/lib/Extension/Bundle/Traits/PackageVersionTrait.php
+++ b/lib/Extension/Bundle/Traits/PackageVersionTrait.php
@@ -34,7 +34,8 @@ trait PackageVersionTrait
      *
      * @return string
      */
-    public function getComposerPackageName(): string {
+    public function getComposerPackageName(): string
+    {
         foreach (InstalledVersions::getAllRawData() as $installed) {
             foreach ($installed['versions'] as $packageName => $packageInfo) {
                 if (!isset($packageInfo['install_path'])) {
--- a/lib/Extension/Bundle/Traits/PackageVersionTrait.php
+++ b/lib/Extension/Bundle/Traits/PackageVersionTrait.php
@@ -34,14 +34,13 @@ trait PackageVersionTrait
      * @return string
      */
     public function getComposerPackageName(): string {
-        $composerFile = __DIR__ . '/../composer.json';
-
-        if (file_exists($composerFile)) {
-            $composerConfig = file_get_contents($composerFile);
-            $composerConfig = json_decode($composerConfig);
-
-            if (property_exists($composerConfig, 'name')) {
-                return $composerConfig->name;
+        foreach([__DIR__ . '/../composer.json', __DIR__ . '/composer.json'] as $composerFile) {
+            if (file_exists($composerFile)) {
+                $composerConfig = json_decode(file_get_contents($composerFile));
+    
+                if (property_exists($composerConfig, 'name')) {
+                    return $composerConfig->name;
+                }
             }
         }
 
--- a/lib/Extension/Bundle/Traits/PackageVersionTrait.php
+++ b/lib/Extension/Bundle/Traits/PackageVersionTrait.php
@@ -35,7 +35,7 @@ trait PackageVersionTrait
      */
     public function getComposerPackageName(): string {
         foreach([__DIR__ . '/../composer.json', __DIR__ . '/composer.json'] as $composerFile) {
-            if (file_exists($composerFile)) {
+            if (is_file($composerFile)) {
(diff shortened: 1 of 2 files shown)


lib/Extension/Bundle/Traits/PackageVersionTrait.php
```diff
@@ -27,12 +27,26 @@
...
+        $composerFile = __DIR__ . '/../composer.json';
+
+        if (file_exists($composerFile)) {
+            $composerConfig = file_get_contents($composerFile);
+            $composerConfig = json_decode($composerConfig);
+
+            if (property_exists($composerConfig, 'name')) {
+                return $composerConfig->name;
+            }
+        }
```
jdreesen: While the `*Bundle` class in Symfony's [current bundle structure](https://symfony.com/doc/4.4/bundles/best_practices.html#directory-structure) (which is not supported by Pimcore at the moment, though, but [I'm on it](https://github.com/pimcore/pimcore/issues/10912)) is located in the `src/` directory, this was not necessarily the case in [the old one](https://symfony.com/doc/3.4/bundles/best_practices.html#directory-structure). Some people put their code [into an `src/` folder](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle) and [some don't](https://github.com/FriendsOfSymfony/FOSUserBundle). So I'd suggest supporting both:

```suggestion
        foreach([__DIR__ . '/../composer.json', __DIR__ . '/composer.json'] as $composerFile) {
            if (file_exists($composerFile)) {
                $composerConfig = json_decode(file_get_contents($composerFile));
    
                if (property_exists($composerConfig, 'name')) {
                    return $composerConfig->name;
                }
            }
        }
```

lib/Extension/Bundle/Traits/PackageVersionTrait.php
```diff
@@ -27,12 +27,25 @@
...
 {
+
     /**
      * Returns the composer package name used to resolve the version
      *
      * @return string
      */
-    abstract protected function getComposerPackageName(): string;
+    public function getComposerPackageName(): string {
+        foreach([__DIR__ . '/../composer.json', __DIR__ . '/composer.json'] as $composerFile) {
```
brusch: I'm wondering if this is really generic enough, as this assumes the bundle class in a specific directory, right?

brusch (changes requested):

lib/Extension/Bundle/Traits/PackageVersionTrait.php
jdreesen: Right. As explained in my previous comment, it may be in one of two places if you adhere to Symfony's "bundle best practices": in the bundle root, or in `src/`.

So, in my opinion, that should be enough, as this is only a default implementation for lazy developers ;) 
If you have some other bundle structure, you can still implement the method yourself, as before.

passioneight:
Much appreciated <3

lib/Extension/Bundle/Traits/PackageVersionTrait.php
dpfahlbusch: So with @jdreesen explantation, the solution is okay and the thread is resolved? @brusch :)

brusch: I was wondering whether we can additionally utilize`\Composer\InstalledVersions::getAllRawData()` - at least for installed packages this would be a safe approach ...

jdreesen: Well... it may be possible, by iterating over all installed bundles and comparing if the current `realpath` starts with one of those `realpath($install_path)`... but I'm not sure if that's worth it.

jdreesen (commented):
You should update the PhpDoc of the trait and [the section in the docs](https://github.com/pimcore/pimcore/blob/10.x/doc/Development_Documentation/20_Extending_Pimcore/13_Bundle_Developers_Guide/05_Pimcore_Bundles/README.md?plain=1#L120), too.

lib/Extension/Bundle/Traits/PackageVersionTrait.php
```diff
@@ -27,12 +27,25 @@
...
+            if (file_exists($composerFile)) {
```
jdreesen: Maybe use `is_file` instead of `file_exists` (which also returns true for directories)?

```suggestion
            if (is_file($composerFile)) {
```

lib/Extension/Bundle/Traits/PackageVersionTrait.php
jdreesen: Should be something like this, I guess:

> **Warning** 
> I didn't test it!

```php
    public function getComposerPackageName(): string
    {
        foreach(InstalledVersions::getAllRawData() as $installed) {
            foreach ($installed['versions'] as $packageName => $packageInfo) {
                if (!isset($packageInfo['install_path'])) {
                    // It's a replaced or provided (virtual) package
                    continue;
                }

                if (str_starts_with(__DIR__, realpath($packageInfo['install_path']))) {
                    return $packageName;
                }
            }
        }

        return '';
    }
```

dpfahlbusch: Tested it , works :) ty @jdreesen

aryaantony92:
@dpfahlbusch Could you please update the documentation also ?

aryaantony92:
@dpfahlbusch Could you please update the documentation also ?

jdreesen:
I think you should update the PhpDoc of the trait, too. It currently says:

> Exposes a simple getVersion() implementation by looking up the installed versions via ocramius/package-versions which is generated on composer install. This trait can be used by using it from a bundle class and implementing getComposerPackageName() to return the name of the composer package to lookup.

jdreesen (approved):

aryaantony92:
Thank you.

jdreesen:
@aryaantony92 maybe you wanna have a look at #12315 next, as it's related and does some cleanup.

aryaantony92:
Thanks, I will have a look.

PR #12257, opened 2022-05-27
Restore Info Block for definition files

blankse:
Restores the deleted info blocks in class/fieldcollection/objectbrick definition files.
See https://github.com/pimcore/pimcore/pull/11373#discussion_r883468253

### DIFF
--- a/models/DataObject/ClassDefinition.php
+++ b/models/DataObject/ClassDefinition.php
@@ -511,7 +511,7 @@ final class ClassDefinition extends Model\AbstractModel
             $data .= $this->getInfoDocBlock();
             $data .= "\n\n";
 
-            $data .= "\nreturn ".$exportedClass.";\n";
+            $data .= 'return '.$exportedClass.";\n";
 
             \Pimcore\File::putPhpFile($definitionFile, $data);
         }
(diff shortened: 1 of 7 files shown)


models/DataObject/ClassDefinition.php
```diff
@@ -508,13 +508,46 @@ public function generateClassFiles($generateDefinitionFile = true)
...
+        $cd = '/**' . "\n";
+        $cd .= '* Inheritance: '.($this->getAllowInherit() ? 'yes' : 'no')."\n";
+        $cd .= '* Variants: '.($this->getAllowVariants() ? 'yes' : 'no')."\n";
+
+        if ($this->getDescription()) {
+            $description = str_replace(['/**', '*/', '//'], '', $this->getDescription());
+            $description = str_replace("\n", "\n* ", $description);
+
+            $cd .= '* '.$description."\n";
+        }
```
jdreesen: Can we have it properly indented, maybe (at least the part that starts with a `*`)? :)

```suggestion
        $cd .= ' * Inheritance: '.($this->getAllowInherit() ? 'yes' : 'no')."\n";
        $cd .= ' * Variants: '.($this->getAllowVariants() ? 'yes' : 'no')."\n";

        if ($this->getDescription()) {
            $description = str_replace(['/**', '*/', '//'], '', $this->getDescription());
            $description = str_replace("\n", "\n * ", $description);

            $cd .= ' * '.$description."\n";
        }
```

blankse: I simple restored the old code. But yes we can optimize it :)

aryaantony92:
@blankse Thank you for the fix.

PR #12258, opened 2022-05-27
Annotate static getContainer method as deprecated

DrLuke:
Annotate discouraged method as deprecated to stop people from using it in new code

  
  

Relates to: https://github.com/pimcore/pimcore/issues/10952#issuecomment-1139347223

The use of `\Pimcore::getContainer()` is a bad practice, as services should be wired in instead of yoinking them out of the container. The comment for the method says that it only really exists for legacy compatibility, yet it's still being used in [new](https://github.com/pimcore/pimcore/blob/10.x/lib/Console/Traits/Parallelization.php#L138) [code](https://github.com/pimcore/data-hub/blob/1.4/src/GraphQL/Mutation/MutationType.php#L260) (thanks @micru).

Annotating it as deprecated will actually mark it as such in the IDE, hopefully resulting in no new code using this.

### DIFF
--- a/lib/Pimcore.php
+++ b/lib/Pimcore.php
@@ -162,6 +162,7 @@ class Pimcore
      * needing to access the container directly. This exists mainly for compatibility with legacy code.
      *
      * @internal
+     * @deprecated
      *
      * @return ContainerInterface|null
      */
--- a/lib/Pimcore.php
+++ b/lib/Pimcore.php
@@ -162,7 +162,7 @@ class Pimcore
      * needing to access the container directly. This exists mainly for compatibility with legacy code.
      *
      * @internal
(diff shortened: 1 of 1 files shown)


lib/Pimcore.php
```diff
@@ -162,6 +162,7 @@ public static function setKernel(KernelInterface $kernel)
      * needing to access the container directly. This exists mainly for compatibility with legacy code.
      *
      * @internal
+     * @deprecated
```
brusch:
```suggestion
     * @deprecated this method just exists for legacy reasons and shouldn't be used in new code
```

brusch:
Thanks a lot! 👍

PR #12261, opened 2022-05-27
[TASK] add possibility to add additional class to block in blockStart

### LINKED ISSUE #11310: [Improvement]: Ability to pass additional classes to block wrapper divs in edit mode, via blockConstruct
### Improvement description

When trying to match the front-end templates as closely as possible in edit mode, there are some circumstances where the wrapper divs which get added are preventing this from being possible.

These circumstances could be prevent if it was possible to pass additional css classes to those wrapper divs.

It might be a simple as being able to pass additional css classes to the [blockStart](https://github.com/pimcore/pimcore/blob/705751d2bcea04b11fab355980c686ef95575dc2/models/Document/Editable/Block.php#L278) function.

For example:

`{% do block.blockStart(false, false, "additional_class") %}`

would results in

```
<div class="pimcore_block_entry additional_class" key="1" data-name="content:1.imagesContainer" data-real-name="imagesContainer" id="ext-element-26">
</div>
```

dpfahlbusch:
Resolves #11310

### DIFF
--- a/doc/Development_Documentation/03_Documents/01_Editables/06_Block.md
+++ b/doc/Development_Documentation/03_Documents/01_Editables/06_Block.md
@@ -118,6 +118,23 @@ If you want to wrap buttons in a div or change the Position.
 {% do block.end() %}
 ```
 
+### Using Manual Mode with additional css class for element in editmode
+
+```twig
+{% set block = pimcore_block("gridblock", {"manual": true}).start() %}
+<div>
+    {% for b in block.iterator %}
+        {% do block.blockConstruct() %}
+            {% do block.blockStart(true, false, 'my-additional-class') %}
+                Add additional class 'my-addional-class' to editmode-div
+            {% do block.blockEnd() %}
+        {% do block.blockDestruct() %}
+    {% endfor %}
+</div>
+{% do block.end() %}
+```
+
+
 ### Accessing Data Within a Block Element
 
 Bricks and structure refer to the CMS demo (content/default template).
--- a/doc/Development_Documentation/03_Documents/01_Editables/06_Block.md
+++ b/doc/Development_Documentation/03_Documents/01_Editables/06_Block.md
@@ -125,7 +125,7 @@ If you want to wrap buttons in a div or change the Position.
 <div>
     {% for b in block.iterator %}
         {% do block.blockConstruct() %}
-            {% do block.blockStart(true, false, 'my-additional-class') %}
+            {% do block.blockStart(additionalClass='my-additional-class') %}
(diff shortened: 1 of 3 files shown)


fashxp:
thx very much for the contribution ... could you also please update the docs?

doc/Development_Documentation/03_Documents/01_Editables/06_Block.md
```diff
@@ -118,6 +118,23 @@ If you want to wrap buttons in a div or change the Position.
...
 ```
 
+### Using Manual Mode with additional css class for element in editmode
+
+```twig
+{% set block = pimcore_block("gridblock", {"manual": true}).start() %}
+<div>
+    {% for b in block.iterator %}
+        {% do block.blockConstruct() %}
+            {% do block.blockStart(true, false, 'my-additional-class') %}
```
jdreesen, line 128: It should be possible to skip the default parameters and [write it like this](https://twig.symfony.com/doc/3.x/templates.html#named-arguments), no?

```suggestion
            {% do block.blockStart(additionalClass='my-additional-class') %}
```

aryaantony92: An error is thrown when we try to skip the default parameters as mentioned. So, I request to revert the commit with the suggested change please.

aryaantony92:
@dpfahlbusch Thank you for your contribution. But, could you please revert this functionality for scheduled blocks ? Since manual mode is not supported for scheduled blocks, it doesn't make sense to add the additional class functionality to scheduled blocks.

models/Document/Editable/Scheduledblock.php
```diff
@@ -200,7 +200,7 @@ public function blockConstruct()
     /**
      * {@inheritdoc}
      */
-    public function blockStart($showControls = true, $return = false)
+    public function blockStart($showControls = true, $return = false, $additionalClass = '')
```
aryaantony92, line 203:
```suggestion
    public function blockStart($showControls = true, $return = false)
```

jdreesen: Not sure if this is possible, because `Scheduledblock extends Block` and while it's allowed to add additional parameters, I don't think it's allowed to remove any.

See: https://3v4l.org/YURPH

dpfahlbusch: If I remove this parameter I get an "Declaration must be compatible with ..." error.
So I just added it and don't use it just like `$return` and `$showControls`

aryaantony92: Yes, you are right. Thank you

PR #12286, opened 2022-05-31
Update grid.js

picman:
When trying to add a column to the grid I get an error `fieldConfig.layout is undefined`. The Apply button doesn't respond. I tried to debug the code and realized that that situation is as follows:
```
// Don't add batch options if field has dynamic options
console.log('hasOwnProperty: ' + fieldConfig.hasOwnProperty('layout')); // => true
console.log('undefined: ' + (fieldConfig.layout === undefined));    // => true
```
So, `hasOwnProperty` is true but `layout` === `undefined` is true too. The suggested solution is to add a condition to check `layout` for `undefined`.

image

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
@@ -104,6 +104,7 @@ pimcore.object.helpers.grid = Class.create({
 
                 // Don't add batch options if field has dynamic options
                 if (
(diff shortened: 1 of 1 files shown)


mcop1:
@picman could you please merge 10.4 into your pr/branch again? This should eliminate the phpstan errors. Thanks!

picman:
I've done that but nothing has changed.

kingjia90:
Closing as the fix is already merged in #12297

PR #12289, opened 2022-05-31
[Task] Add @param and @return php doc types to /lib/Log

### LINKED ISSUE #12227: [Task] Add @param and @return php doc types to /lib/Log
### Improvement description

- [ ] Add @param php doc types for all public/protected classes/methods/properties in /lib/Log (excl. Dao and @internal) 
- [ ] Add @return php doc types for all public/protected classes/methods/properties in /lib/Log (excl. Dao and @internal) 

mcop1:
Resolves #12227

### DIFF
--- a/lib/Log/Processor/ApplicationLoggerProcessor.php
+++ b/lib/Log/Processor/ApplicationLoggerProcessor.php
@@ -141,7 +141,7 @@ class ApplicationLoggerProcessor
 
     /**
      * @param string $filename
-     * @return array|string
+     * @return string
      */
     private function normalizeFilename($filename)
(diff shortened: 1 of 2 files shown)


lib/Log/Processor/ApplicationLoggerProcessor.php
```diff
@@ -138,6 +138,11 @@ private function processLoggingSource(array $record): array
         return $record;
     }
 
+
+    /**
+     * @param string $filename
+     * @return array|string
```
blankse:
```suggestion
     * @return string
```

PR #12297, opened 2022-05-31
Fix — grid options cant save

### LINKED ISSUE #12281: [Bug]: Grid Option can't save
### Expected behavior

If I customize the Grid Option in the admin panel, I should be able to Apply or Save Copy & Share. 

### Actual behavior

I try to add new data attributes into the Grid Option from the admin panel, and it can't Apply or Save Copy & Share, without any alert. 

微信图片_20220531095513

### Steps to reproduce

As above. The version is v10.4.2

andrew72ru:
Resolves #12281

## Additional info  

When I'm trying to apply grid options the `fieldConfig` in `bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js` has a `layout` but it's undefined. So when JS is trying to check `fieldConfig.layout.hasOwnProperty('dynamicOptions')` it's actually trying to call `hasOwnProperty` on undefined. 
I've fixed it by check type of `fieldConfig.layout`

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
@@ -104,7 +104,7 @@ pimcore.object.helpers.grid = Class.create({
 
                 // Don't add batch options if field has dynamic options
                 if (
-                    !fieldConfig.hasOwnProperty('layout')
+                    !fieldConfig.hasOwnProperty('layout') || typeof (fieldConfig.layout) === 'undefined'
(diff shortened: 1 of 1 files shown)


jdreesen:
@andrew72ru you need to rebase your branch onto `10.4` and skip all commits from `10.x`.

aashan10 (approved):
Verified that this fixes #12281

gjorgic:
I wonder why we didn't use the following expression here instead of chaining one more checks
```
 if (!fieldConfig.layout || fieldConfig.layout.dynamicOptions !== true) {
 ...
}
```

dvesh3:
@gjorgic could please provide a follow up PR? thanks!

PR #12302, opened 2022-05-31
[Docs] Remove description how to setup cronjob randomness

jdreesen:
This explanation somehow got missing in the newer docs. I just asked myself why someone would want to do this, so I [dug for the reason](https://github.com/pimcore/pimcore/commit/cbbd9adf63efe53138824cebc1ebe790493a35df#diff-d875910eed5015570bbe4577091b0fad971b2eba0f929c2809dcf81d8fdb3b8eR77). I'd suggest re-adding the explanation to the docs.

To be honest, I'm not entirely sure if this is actually needed in Pimcore X, since we've got the message queue now. What do you think?

### DIFF
--- a/doc/Development_Documentation/01_Getting_Started/01_Advanced_Installation_Topics.md
+++ b/doc/Development_Documentation/01_Getting_Started/01_Advanced_Installation_Topics.md
@@ -45,6 +45,9 @@ pimcore_install:
 
 
 ## Add some randomness to the maintenance cron job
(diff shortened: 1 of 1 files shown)


brusch:
Imho this is obsolete now because of the message queue.

jdreesen:
Nice, I deleted the paragraph.

brusch:
Thanks a lot!

PR #12304, opened 2022-05-31
[Symfony] Fix some deprecations from version 5

### LINKED ISSUE #11750: [Symfony] Fix some deprecations from version 5
### Feature description

- [x] `PreAuthenticatedAdminSessionFactory` should not implement deprecated `Symfony\Bundle\SecurityBundle\DependencyInjection\Security\Factory\SecurityFactoryInterface`
- [x] ~`UserChecker` should not extend from deprecated class `Symfony\Component\Security\Core\User\UserChecker`, instead use `InMemoryUserChecker`~ already fixed and merged in [12193](https://github.com/pimcore/pimcore/pull/12193/files#diff-cb3667f0b182bf474cbad5f8205fe71ff283c5d62c680b4d92cb03deb9c51e2aR22)
- [x] `UserProvider` & `ObjectUserProvider` should not throw deprecated exception UsernameNotFoundException
- [x] `IntlFormatter` should stop using internal class  `Symfony\Component\Intl\DateFormatter\IntlDateFormatter` use `symfony/polyfill-intl-icu` ^1.21 instead
- [x] `ExpressionService` should not utilize deprecated class `AnonymousToken`

kingjia90:
Resolves #11750


kingjia90:
~Mhh... It seems that i can't apply the changes of bundles/AdminBundle/PimcoreAdminBundle.php from `$extension->addSecurityListenerFactory` to `addAuthenticatorFactory` because this was introduced in [5.4](https://github.com/symfony/security-bundle/blob/4d5f4953969f136ed7fa6b4a4924cf0fa54a2cd4/DependencyInjection/SecurityExtension.php#L1098-L1119), that's is also why it only passes the tests with 5.4.x-dev~

brusch:
@kingjia90 it should be safe to do `"symfony/security-bundle": "^5.4.0"`  - we did the same for `symfony/cache` and `symfony/messanger` ... 😊

bundles/AdminBundle/Security/User/UserProvider.php
```diff
@@ -34,7 +34,7 @@ public function loadUserByUsername($username)
             return new User($pimcoreUser);
         }
 
-        throw new UsernameNotFoundException(sprintf('User %s was not found', $username));
+        throw new UserNotFoundException(sprintf('User %s was not found', $username));
```
dvesh3, line 37: I would add this to upgrade notes.

lib/Security/User/ObjectUserProvider.php
```diff
@@ -96,7 +96,7 @@ public function loadUserByUsername($username)
             return $user;
         }
 
-        throw new UsernameNotFoundException(sprintf('User %s was not found', $username));
+        throw new UserNotFoundException(sprintf('User %s was not found', $username));
```
dvesh3, line 99: I would add this to upgrade notes.

dvesh3 (approved):

PR #12305, opened 2022-05-31
Remove internal annotation from navigation builder

### LINKED ISSUE #12276: [Improvement]: Pimcore\Navigation\Builder marked as internal but mentioned in documentation
### Improvement description

`Pimcore/Navigation/Builder` was marked as internal on 29th April 2021 (https://github.com/pimcore/pimcore/commit/f85ba025c30aae29a78805f822002bf9fd96b39e#diff-0ae9dfd15f90eab5d95db3ca02c83501513ecb88b3700267e2f3fd7d83c0d16d). However, the documentation still mentions it as an alternative to building the navigation in the view:

> The building step does not necessarily need to happen in the view script. In fact the view helper just forwards the `build()` call to the `Pimcore\Navigation\Builder` service. You can also build the navigation in your controller or a service and pass the navigation object to the view. (https://pimcore.com/docs/pimcore/current/Development_Documentation/Documents/Navigation.html)

Either the class was mistakenly marked as internal or the documentation needs to be updated. Removing the notice from the documentation would mean that the navigation has to be build in the view and can not be build in a controller.

IAmMajo:
This pull request removes the internal annotation from the navigation builder and adds an internal annotation to all protected methods and properties of the navigation builder.

Resolves #12276

### DIFF
--- a/lib/Navigation/Builder.php
+++ b/lib/Navigation/Builder.php
@@ -24,9 +24,6 @@ use Pimcore\Navigation\Iterator\PrefixRecursiveFilterIterator;
 use Pimcore\Navigation\Page\Document as DocumentPage;
 use Pimcore\Navigation\Page\Url;
 
-/**
- * @internal
- */
 class Builder
 {
     /**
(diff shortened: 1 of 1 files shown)


brusch:
LGTM, thank you very much! 
Related: #12310 - maybe you can create a PR for that as well 😉

PR #12306, opened 2022-06-01
Added @param and @returns to docblocks for models/Asset

### LINKED ISSUE #12211: [Task] Add @param and @return php doc types to /models/Asset
### Improvement description

- [ ] Add @param php doc types for all public/protected classes/methods/properties in /models/Asset (excl. Dao and @internal) 
- [ ] Add @return php doc types for all public/protected classes/methods/properties in /models/Asset (excl. Dao and @internal) 

aashan10:
Resolves #12211

### DIFF
--- a/models/Asset/Image.php
+++ b/models/Asset/Image.php
@@ -282,7 +282,7 @@ EOT;
     /**
      * Returns a path to a given thumbnail or an thumbnail configuration.
      *
-     * @param string|array|Image\Thumbnail\Config $config
+     * @param null|string|array|Image\Thumbnail\Config $config
      * @param bool $deferred
      *
      * @return Image\Thumbnail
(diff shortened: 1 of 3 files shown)


kingjia90:
looks good, only thing i've noticed that might be missing is the `null` in $config `@param`
https://github.com/pimcore/pimcore/blob/6f220dab7e74840bc241ac60696df2e364c74119/models/Asset/Image.php#L294-L299

aashan10:
> looks good, only thing i've noticed that might be missing is the `null` in $config `@param`
> 
> https://github.com/pimcore/pimcore/blob/6f220dab7e74840bc241ac60696df2e364c74119/models/Asset/Image.php#L294-L299

@kingjia90 thanks. Fixed it!

PR #12315, opened 2022-06-01
Use composer's own package info instead of `ocramius/package-versions`

jdreesen:
Since Composer v2, `ocramius/package-versions` is not necessary anymore, as composer itself provides the info we need.

> **Note**
> The first commit contains the real changes. 
> The second and third one are only some refactoring/cleanup of code that use the updated code.

Related to #12256


brusch:
@jdreesen Totally makes sense - but since we remove a dependency someone could rely on, I'd at least mention it in the upgrade notes. Can you please add this info there? 

Thanks a lot!

jdreesen:
Done.

brusch:
Very nice, thank you very much!

BlackbitDevs:
@jdreesen https://github.com/pimcore/pimcore/pull/12315/files#diff-53c413bf6c2d11d248cb02f5366a7823ada5ddb0430c790dbb0318f595c6eb23R32 is a BC break. Before a version number like `10.5.0` got returned, now it is `v10.5.0.0`. Especially the `v` prefix is a problem when combining it with [`version_compare()`](https://www.php.net/manual/de/function.version-compare.php). WDYT, could we switch to `InstalledVersions::getVersion()`? This would return `10.5.0.0` - which is possible to use with `version_compare()`.

brusch:
@BlackbitNeueMedien `InstalledVersions::getVersion()` seems to do the trick - can you create a PR? 😊

jdreesen:
Well, it's an `@internal` class, so it's not really a BC break, right?

But sure, we can change it. `InstalledVersions::getVersion()` returns a longer version with an additional `.0` at the end, while `InstalledVersions::getPrettyVersion()` *may* start with a `v` (which we could remove as done in the [`PackageVersionTrait`](https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/lib/Extension/Bundle/Traits/PackageVersionTrait.php#L60)).

brusch:
Ah yes, totally missed the `@internal` 😊
For me both is fine, but `getVersion()` seems to be closer to the nameing and the expected return value, right?

jdreesen:
Note that while most of the time this is the only difference, when e.g. a package is installed as a dependency in a stable version:

```php
'pimcore/pimcore' => array(
    'pretty_version' => 'v10.5.0',
    'version' => '10.5.0.0',
    'type' => 'library',
    'install_path' => __DIR__ . '/../pimcore/pimcore',
    'aliases' => array(),
    'reference' => 'bf022270642f26cf957d06ba252dea515c7298b8',
    'dev_requirement' => false,
),
```
sometimes it's not that easy, like for dev versions:

```php
'pimcore/pimcore' => array(
    'pretty_version' => '10.x-dev',
    'version' => '10.9999999.9999999.9999999-dev',
    'type' => 'library',
    'install_path' => __DIR__ . '/../../',
    'aliases' => array(),
    'reference' => '3255bd9f8bd33b36a5ab26fe9941b751d496fdaa',
    'dev_requirement' => false,
),
```
So I'm not sure that it's safe to be used with `version_compare()` anyway.

As far as I know, `composer/semver` should be used for this: https://getcomposer.org/doc/07-runtime.md#knowing-whether-package-x-is-installed-in-version-y

jdreesen:
By the way, `ocramius/package-versions` [uses the same code](https://github.com/Ocramius/PackageVersions/blob/6738b33f71099cad889e5560f50477bf3f05bfd9/src/PackageVersions/Versions.php#L34-L38) as we do now, except that it merges `InstalledVersions::getPrettyVersion()` with `InstalledVersions::getReference()` and Pimcore [split it up again](https://github.com/pimcore/pimcore/blob/v10.4.6/lib/Version.php#L41), so I'm wondering why it now behaves differently.

BlackbitDevs:
> By the way, ocramius/package-versions [uses the same code](https://github.com/Ocramius/PackageVersions/blob/6738b33f71099cad889e5560f50477bf3f05bfd9/src/PackageVersions/Versions.php#L34-L38) as we do now, except that it merges InstalledVersions::getPrettyVersion() with InstalledVersions::getReference() and Pimcore [split it up again](https://github.com/pimcore/pimcore/blob/v10.4.6/lib/Version.php#L41), so I'm wondering why it now behaves differently.

Hmm, you are right. The BC break obviously happened between Composer 1 and 2 when also `ocramius/package-versions` used a [different approach](https://github.com/Ocramius/PackageVersions/blob/1.9.x/src/PackageVersions/Versions.php) (because at this time `Composer\InstalledVersions` did not exist yet). As Pimcore requires Composer 2 since `10.0`, I think current code can stay. Have created a fallback class for still supporting Pimcore 6 in a bundle:
```php
<?php
namespace Composer;

use Pimcore\Version;

class InstalledVersions
{
    public static function getVersion() {
        return Version::getVersion();
    }
}
```

Together with a fallback namespace this only gets loaded when Composer's `Composer\InstalledVersions` is not available.

PR #12325, opened 2022-06-02
[Grid exports] Initially enable inheritance

BlackbitDevs:
Currently due the non-existing ExtJs config `inputValue` in 
https://github.com/pimcore/pimcore/blob/52eb193be4307054a161d65bdd2ac0f7fb32227d/bundles/AdminBundle/Resources/public/js/pimcore/object/gridexport/csv.js#L28
inheritance is not enabled initially. Thus values may differ from what the user sees in the grid table. Imho in the default setting the export data should be the same as in the grid.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/gridexport/csv.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/gridexport/csv.js
@@ -26,6 +26,7 @@ pimcore.object.gridexport.csv = Class.create(pimcore.element.gridexport.abstract
             fieldLabel: t('enable_inheritance'),
             name: 'enableInheritance',
             value: true,
+            inputValue: true,
             labelWidth: 200
         });
 
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/gridexport/csv.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/gridexport/csv.js
@@ -25,7 +25,7 @@ pimcore.object.gridexport.csv = Class.create(pimcore.element.gridexport.abstract
         var enableInheritance = new Ext.form.Checkbox({
             fieldLabel: t('enable_inheritance'),
             name: 'enableInheritance',
-            inputValue: true,
+            value: true,
             labelWidth: 200
         });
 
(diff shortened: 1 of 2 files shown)


kingjia90:
Thank you! I also agree with you that it should be enabled initially, so it matches to what you see in the grid.

It works but i have one minor thing though, the `inputValue` is not necessary but would be nice to keep it: 
without it ExtJS [by default](https://docs.sencha.com/extjs/6.0.2/classic/Ext.form.field.Checkbox.html#cfg-inputValue) it passes `'on'` and while it still works because it is truthy in PHP, might be better to keep that `true` for the stricter params type/"phpstan" and consistency with other checkboxes (eg. 'share globally' and 'set as favourite').

BlackbitDevs:
Ok, added `inputValue: true` again. To me this is consistent now - but consistently wrong because ExtJS defines this config option as a string value.
I do not know what you mean with `it is truthy in PHP` and `phpstan` because this config is pure JS and never used in PHP afaik.

kingjia90:
> Ok, added `inputValue: true` again. To me this is consistent now - but consistently wrong because ExtJS defines this config option as a string value. I do not know what you mean with `it is truthy in PHP` and `phpstan` because this config is pure JS and never used in PHP afaik.

Thank you for the changes and followup.

It's passes the value to PHP here, i've put a breakpoint on `$settings['enableInheritance']` and it's a string `'on`' when inputValue is not set.
https://github.com/pimcore/pimcore/blob/90fdb3d6f99f022170d9cc1ea44f69f5542986da/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php#L1325-L1337

I see the consistently wrong point, my bad, the link i've provided in the previous comment was referring doc 6.0.2, but since extJS 7.0 (which is the version Pimcore currently uses) it supports both string and boolean.
https://docs.sencha.com/extjs/7.0.0/classic/Ext.form.field.Checkbox.html#cfg-inputValue

So all good, ready to merge.

BlackbitDevs:
Ah, I see, `inputValue` is the value which ExtJS sends when the checkbox is checked and the form gets submitted.

PR #12327, opened 2022-06-02
fix: UserController::updateUser does not update all properties

xlttj:
Using `User\UserRole` leads to the bug that not all properties of a user are updated. I suggest using `AbstractUser`instead.


xlttj:
errorneous pr

PR #12328, opened 2022-06-02
 fix: UserController::updateUser does not update all properties

xlttj:
Using User\UserRole leads to the bug that not all properties of a user are updated.
I suggest using AbstractUserinstead.


blankse:
@xlttj Do you use the latest 10.4.x Version? Is already fixed in 10.4.1, see https://github.com/pimcore/pimcore/issues/12024

xlttj:
@blankse You're right, I'm using 10.4.0. I'll update and close this PR.

Issue #12333, opened 2022-06-02
[Bug]: $currentFilter not correct?


elements-team:
https://github.com/pimcore/pimcore/blob/10.x/bundles/EcommerceFrameworkBundle/FilterService/FilterService.php <- line 84

### Expected behavior

I guess the initFilterService() should return a array of all currently set filters (as it is described in the Doc-Block).

### Actual behavior

Currently $currentFilter is always overwritten and filters are not added to the $currentFilter array .

### Steps to reproduce

I guess the the code should be changed from 
`$currentFilter = $this->addCondition($filter, $productList, $currentFilter, $params);`
to
`$currentFilter[] = $this->addCondition($filter, $productList, $currentFilter, $params);`

dvesh3:
@elements-team sorry, it works for me, filters are added up when selecting multiple on the frontend:
![image](https://user-images.githubusercontent.com/5137917/172340162-b443023b-f0b3-40ff-b939-988609904639.png)

FilterType `addCondition()` returns the appended array and set (overwrite) the $currentFilter value. please see https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/bundles/EcommerceFrameworkBundle/FilterService/FilterType/MultiSelect.php#L69

ctippler:
Hi,
sorry - you are right. I just figured out that project specific custom filter types caused the problem.
Regards,
Christian

PR #12336, opened 2022-06-02
Bugifx currentFilter #12333

ctippler:
https://github.com/pimcore/pimcore/issues/12333


dvesh3:
@ctippler please see https://github.com/pimcore/pimcore/issues/12333#issuecomment-1148395332

ctippler:
Sorry - you are right - the problem was caused by an app specific implementation

PR #12337, opened 2022-06-02
[Task] Add @param and @return php doc types to /lib/HttpKernel

### LINKED ISSUE #12225: [Task] Add @param and @return php doc types to /lib/HttpKernel
### Improvement description

- [ ] Add @param php doc types for all public/protected classes/methods/properties in /lib/HttpKernel (excl. Dao and @internal) 
- [ ] Add @return php doc types for all public/protected classes/methods/properties in /lib/HttpKernel (excl. Dao and @internal) 

aryaantony92:
Resolves #12225

### DIFF
--- a/lib/HttpKernel/BundleCollection/AbstractItem.php
+++ b/lib/HttpKernel/BundleCollection/AbstractItem.php
@@ -34,6 +34,11 @@ abstract class AbstractItem implements ItemInterface
      */
     private $source;
 
+    /**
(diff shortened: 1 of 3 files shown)


fashxp:
we don't need to update phpdocs when native php docs are already in place. 
and we also don't need to update phpdocs when `@inheritdoc` is in place and up-to-date.

thus, it seems there is nothing to change here :-)

PR #12339, opened 2022-06-02
[Task] Add @param and @return php doc types to /models/<REST>

### LINKED ISSUE #12215: [Task] Add @param and @return php doc types to /models/<REST>
### Improvement description

- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Exception (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Exception (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Glossary (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Glossary (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/GridConfig (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/GridConfig  (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/GridConfigFavourite (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/GridConfigFavourite (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Metadata (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Metadata (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Notification (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Notification (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Property (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Property (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Redirect (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Redirect (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Schedule (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Schedule (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Search (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Search (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Site (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Site (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Staticroute (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Staticroute (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Translation (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Translation (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/WebsiteSetting (excl. Dao and @internal) 
- [ ]Add @return php doc types for all public/protected classes/methods/properties in /models/WebsiteSetting (excl. Dao and @internal) 
- [x] Add @param php doc types for all public/protected classes/methods/properties in /models/Workflow (excl. Dao and @internal) 
- [x] Add @return php doc types for all public/protected classes/methods/properties in /models/Workflow (excl. Dao and @internal) 

- [x] Make sure phpstan tests still run through

dvesh3:
Resolves #12215

### DIFF
--- a/models/Glossary.php
+++ b/models/Glossary.php
@@ -27,7 +27,7 @@ class Glossary extends AbstractModel
     /**
      * @internal
      *
-     * @var int|null
+     * @var int
      */
     protected $id;
 
@@ -136,7 +136,7 @@ class Glossary extends AbstractModel
     }
 
     /**
-     * @return int|null
+     * @return int
      */
     public function getId()
     {
--- a/models/Glossary.php
+++ b/models/Glossary.php
@@ -27,7 +27,7 @@ class Glossary extends AbstractModel
     /**
      * @internal
      *
-     * @var int
+     * @var int|null
      */
     protected $id;
 
@@ -136,7 +136,7 @@ class Glossary extends AbstractModel
     }
 
     /**
-     * @return int
+     * @return int|null
      */
     public function getId()
     {
--- a/models/Notification.php
+++ b/models/Notification.php
@@ -61,14 +61,14 @@ class Notification extends AbstractModel
     /**
      * @internal
      *
-     * @var User|null
+     * @var User
      */
     protected $recipient;
 
     /**
      * @internal
      *
-     * @var string|null
+     * @var string
      */
     protected $title;
 
@@ -82,7 +82,7 @@ class Notification extends AbstractModel
     /**
      * @internal
      *
-     * @var string|null
+     * @var string
      */
     protected $message;
 
@@ -212,7 +212,7 @@ class Notification extends AbstractModel
     }
 
     /**
-     * @return User|null
+     * @return User
      */
     public function getRecipient(): ?User
     {
@@ -220,11 +220,11 @@ class Notification extends AbstractModel
     }
 
     /**
-     * @param User|null $recipient
+     * @param User $recipient
      *
      * @return Notification
      */
-    public function setRecipient(?User $recipient): self
+    public function setRecipient(User $recipient): self
     {
         $this->recipient = $recipient;
 
@@ -232,19 +232,19 @@ class Notification extends AbstractModel
     }
 
     /**
-     * @return string|null
+     * @return string
      */
-    public function getTitle(): ?string
+    public function getTitle(): string
     {
         return $this->title;
     }
 
     /**
-     * @param string|null $title
+     * @param string $title
      *
      * @return Notification
      */
-    public function setTitle(?string $title): self
+    public function setTitle(string $title): self
     {
         $this->title = $title;
 
@@ -272,19 +272,19 @@ class Notification extends AbstractModel
     }
 
     /**
-     * @return string|null
+     * @return string
      */
-    public function getMessage(): ?string
+    public function getMessage(): string
     {
         return $this->message;
     }
 
     /**
-     * @param string|null $message
+     * @param string $message
      *
      * @return Notification
      */
-    public function setMessage(?string $message): self
+    public function setMessage(string $message): self
     {
         $this->message = $message;
 
--- a/models/Notification.php
+++ b/models/Notification.php
@@ -61,7 +61,7 @@ class Notification extends AbstractModel
     /**
      * @internal
      *
-     * @var User
+     * @var User|null
      */
     protected $recipient;
 
@@ -82,7 +82,7 @@ class Notification extends AbstractModel
     /**
      * @internal
      *
-     * @var string
+     * @var string|null
      */
     protected $message;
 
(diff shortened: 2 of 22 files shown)


models/WebsiteSetting/Listing.php
```diff
@@ -16,31 +16,32 @@
 namespace Pimcore\Model\WebsiteSetting;
 
 use Pimcore\Model;
+use Pimcore\Model\WebsiteSetting;
 
 /**
  * @method \Pimcore\Model\WebsiteSetting\Listing\Dao getDao()
```
jdreesen: Could be changed, too, as `WebsiteSetting` is imported now.

```suggestion
 * @method WebsiteSetting\Listing\Dao getDao()
```

dvesh3: yes, thanks!

models/Property.php
```diff
@@ -129,6 +129,8 @@ public function getCid()
     }
 
     /**
+     * enum('document','asset','object')
+     *
      * @return string
```
jdreesen, line 134: What about using PHPStan's type annotations for this? [Like this:](https://phpstan.org/writing-php-code/phpdoc-types#literals-and-constants)
```suggestion
     * @return 'document'|'asset'|'object'
```

Or maybe even [with an alias](https://phpstan.org/writing-php-code/phpdoc-types#local-type-aliases), which can be used for the property and its getters/setters.

dvesh3: @jdreesen This needs more refactoring so not the priority for us. we would be happy to receive a PR for this improvement. thanks for your suggestion 😉

jdreesen: I'm afraid I won't have time for that in the near future :/

models/Notification.php
```diff
@@ -212,39 +212,39 @@ public function setSender(?User $sender): self
     }
 
     /**
-     * @return User|null
+     * @return User
```
blankse: Why it cannot be null? If I create a new Notification it is null, right?
```
$notification = new Notification();
$notification->getRecipient();
```
In addition, the return type hint is still nullable

aryaantony92: @blankse Thanks, I will change the typehint.  It cannot be null because it doesn't make sense to create/send a new notification with an empty recipient.

blankse: Yes, but I think that should check the save method. Until the setRecipient() method is called, the attribute is still null. So why shouldn't the user be able to use getRecipient to check whether it's already set?

models/Glossary.php
```diff
@@ -27,7 +27,7 @@ class Glossary extends AbstractModel
     /**
      * @internal
      *
-     * @var int|null
+     * @var int
      */
     protected $id;
```
blankse: Can be null for new unsaved models. Also for the getter and other models.
```
$glossary = new Glossary();
var_dump($glossary->getId());
```

models/Notification.php
aryaantony92: @blankse I have added the null check before the save() method. But, we decided to follow an approach where the  properties type would match with the db column types. That's why we made it to not null.

blankse: For setter typehints I agree. But getter/properties can be null for new unsaved models. When you really want to change this its a BC break I think, because no getter could be called now until the setter is called.

dvesh3: yes, you are right. Reverted the changes for BC reasons. thanks!

models/Glossary.php
dvesh3: yes, you are right. Reverted the changes for BC reasons. thanks!

models/Translation.php
```diff
@@ -93,7 +93,7 @@ public static function IsAValidLanguage(string $domain, string $locale): bool
     }
 
     /**
-     * @return string|null
```
aryaantony92: @dvesh3 Should n't we keep this getter also as nullable ?

dvesh3: good catch! changed it 😊

PR #12343, opened 2022-06-03
CMS Feature: Brick Preview Tooltip

GALCF:
Adds the possibility to show a preview on add-brick button hover.

### How to use

- The brick needs to implement `Pimcore\Extension\Document\Areabrick\PreviewAwareInterface`
- The html returned in `getPreviewHtml` will be shown as a tooltip on the add-brick button

### Examples

With text:
image

With image:
image


robertSt7:
Many thanks for the PR

PR #12348, opened 2022-06-07
Replace `Pimcore\Cache\Runtime` inside `models/*` with `Pimcore\Cache\RuntimeCacheTrait`

### LINKED ISSUE #11339: [Improvement]: Use `RuntimeCacheTrait` whenever possible instead of dealing with `Cache\Runtime` directly
### Improvement description

`RuntimeCacheTrait` was introduced by #11254 

There are multiple places in the code where `Pimcore\Cache\Runtime` is used directly in models, this should be changed to the new trait.

aashan10:
Resolves #11339

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -18,6 +18,7 @@ namespace Pimcore\Model;
 use Doctrine\DBAL\Exception\DeadlockException;
 use League\Flysystem\FilesystemOperator;
 use League\Flysystem\UnableToMoveFile;
+use Pimcore\Cache\RuntimeCacheTrait;
 use Pimcore\Event\AssetEvents;
 use Pimcore\Event\FrontendEvents;
 use Pimcore\Event\Model\AssetEvent;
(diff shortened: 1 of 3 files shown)


models/Asset.php
```diff
@@ -248,7 +251,7 @@ public static function getById($id, $force = false)
                     $asset->getDao()->getById($id);
                 }
 
-                \Pimcore\Cache\Runtime::set($cacheKey, $asset);
+                self::setCache($asset, $cacheKey);
```
jdreesen, line 254: The (non-runtime-)cache is now saved twice, no? `RuntimeCacheTrait::setCache()` saves both caches, and a few lines below (259) the non-runtime-cache is saved again.

models/Asset.php
```diff
@@ -258,7 +261,7 @@ public static function getById($id, $force = false)
                 return null;
             }
         } else {
-            \Pimcore\Cache\Runtime::set($cacheKey, $asset);
+            self::setCache($asset, $cacheKey);
```
jdreesen, line 264: This now saves the non-runtime-cache, too, which wasn't the case before. Is this correct/intended?

aashan10: No, it is not intended. This is still a work in progress.

aashan10:
Closing this pull request as discussed [here](https://github.com/pimcore/pimcore/issues/11339#issuecomment-1151152660)

PR #8759, opened 2021-04-12
Localized fields in block in object brick result in Pimcore error

weisswurstkanone:
resolves #5082

PR #12349, opened 2022-06-07
Objectbrick: Allow localized block fields again

blankse:
Since Pimcore 10.0 it is not allowed to create an objectbick with localized block fields. But this was working in Pimcore 6.9
Only localized fields in block doesn't work.

See https://github.com/pimcore/pimcore/pull/8759


dvesh3:
@blankse thanks!

PR #12354, opened 2022-06-07
ClassDefinition: maxItems property should be int or null

blankse:
It's not constantly the same behaviour at the moment.

Data types `Fieldcollections`,  `ManyToManyObjectRelation`, `ManyToManyRelation`, `Multiselect` and `Objectbricks` use the getAsIntegerCast method which return int or empty string (but the doc type says int or null).

Data type `Block` do not cast. So it can be int or null.

Data type `Classificationstore` always cast to int.

I constantly use the getAsIntegerCast method and correct the return to int or null.


mcop1:
Thanks!

PR #12361, opened 2022-06-08
Add configName to PERSPECTIVE_POST_GET_RUNTIME event

blankse:
Followup to https://github.com/pimcore/pimcore/pull/10815

In the PERSPECTIVE_POST_GET_RUNTIME event we didn't have the info which perspective is active.


brusch:
Totally makes sense, thanks!

PR #12368, opened 2022-06-08
[IMPROVEMENT] add min, max, decimalsize and -precision to quantityValue

### LINKED ISSUE #2035: Options precision and max-min for quantity value
Offer same options as for numeric data type.

https://github.com/pimcore/pimcore/pull/845 could be used as a guideline

dpfahlbusch:
Resolves #2035

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/quantityValue.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/quantityValue.js
@@ -132,7 +132,7 @@ pimcore.object.classes.data.quantityValue = Class.create(pimcore.object.classes.
                     xtype: "numberfield",
                     fieldLabel: t("decimal_precision"),
                     name: "decimalPrecision",
-                    maxValue: 65,
+                    maxValue: 30,
                     value: this.datax.decimalPrecision
                 },
                 {
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/quantityValue.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/quantityValue.js
@@ -67,13 +67,6 @@ pimcore.object.classes.data.quantityValue = Class.create(pimcore.object.classes.
             },
             {
                 xtype: "numberfield",
-                fieldLabel: t("decimal_precision"),
-                name: "decimalPrecision",
-                maxValue: 65,
-                value: this.datax.decimalPrecision
-            },
-            {
-                xtype: "numberfield",
                 fieldLabel: t("default_value"),
                 name: "defaultValue",
                 value: this.datax.defaultValue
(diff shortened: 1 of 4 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/quantityValue.js
```diff
@@ -126,6 +119,61 @@ pimcore.object.classes.data.quantityValue = Class.create(pimcore.object.classes.
...
+                    fieldLabel: t("decimal_size"),
+                    name: "decimalSize",
+                    maxValue: 65,
+                    value: this.datax.decimalSize
+                },
+                {
+                    xtype: "numberfield",
+                    fieldLabel: t("decimal_precision"),
+                    name: "decimalPrecision",
+                    maxValue: 65,
```
kingjia90:
```suggestion
                    maxValue: 30,
```

models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -278,7 +534,11 @@ public function getDataForResource($data, $object = null, $params = [])
     public function getDataFromResource($data, $object = null, $params = [])
     {
         if ($data[$this->getName() . '__value'] !== null || $data[$this->getName() . '__unit']) {
-            $value = $data[$this->getName() . '__value'];
+            if (is_numeric($data[$this->getName() . '__value'])) {
```
kingjia90:
```suggestion
            if (!is_numeric($data[$this->getName() . '__value'])) {
```
Shouldn't it be that If it is not already `numeric`, then must be converted with `$this->toNumeric()` ?

models/DataObject/ClassDefinition/Data/QuantityValue.php
```diff
@@ -402,12 +662,40 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
             throw new Model\Element\ValidationException('Empty mandatory field [ '.$this->getName().' ]');
         }
 
+        if ($data !== null && !$this->isEmpty($data->getValue()) && !is_numeric($data->getValue())) {
```
kingjia90, line 664: just wondering, would be there any drawbacks by moving this block inside the one line 676 since it shares 2 of the conditions?

kingjia90 (changes requested):
Thank you for the PR! 
It seems working fine to me, i've just found some minor issues.

models/DataObject/ClassDefinition/Data/QuantityValue.php
dpfahlbusch: Not sure. Numeric uses the same logic
https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/ClassDefinition/Data/Numeric.php#L511

kingjia90:
@dpfahlbusch Just a gentle reminder 🏓

models/DataObject/ClassDefinition/Data/QuantityValue.php
kingjia90: ah, i see, there's a difference on how omitMandatoryCheck is managed.
IMHO the best would be that both be almost the same, by even removing this

if ($omitMandatoryCheck) {
    return;
 }
and adding the omitMandatoryCheck to the IF block in the same way as numeric, otherwise it would be inconsistent.
Could you please do the adaptations?

dpfahlbusch: Done :)

kingjia90 (approved):

kaurov:
Thanks a lot for this change!

how about translations for units?

PR #12371, opened 2022-06-08
[Bug]: LocationAwareConfigRepository to check $_SERVER instead on $_ENV

### LINKED ISSUE #12335: [Bug]: LocationAwareConfigRepository doesn't check SERVER
### Expected behavior

`LocationAwareConfigRepository` checks both `$_ENV` and `$_SERVER`, same as Symfony's `\Symfony\Component\DependencyInjection\EnvVarProcessor` does.

### Actual behavior

`LocationAwareConfigRepository` only checks `$_ENV`, which is commonly disabled by setting PHP's `variables_order` to not include it. [`$_SERVER` is populated from underlying environment variables](https://www.php.net/manual/en/ini.core.php#ini.variables-order).

### Steps to reproduce

See https://github.com/pimcore/data-hub/issues/539

kingjia90:
Fixes #12335

### DIFF
--- a/bundles/EcommerceFrameworkBundle/VoucherService/DefaultService.php
+++ b/bundles/EcommerceFrameworkBundle/VoucherService/DefaultService.php
@@ -222,7 +222,7 @@ class DefaultService implements VoucherServiceInterface
 
         // calculate not applied rules with voucher conditions
         $notAppliedRules = array_udiff($validRules, $appliedRules, function ($rule1, $rule2) {
-            return strcmp($rule1->getId(), $rule2->getId());
+            return $rule1->getId() <=> $rule2->getId();
         });
         $notAppliedRulesWithVoucherCondition = [];
         foreach ($notAppliedRules as $notAppliedRule) {
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -147,7 +147,11 @@ class Bootstrap
     private static function prepareEnvVariables()
     {
         if (!($_SERVER['PIMCORE_SKIP_DOTENV_FILE'] ?? false)) {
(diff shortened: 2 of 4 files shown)


jdreesen:
Shouldn't it check both?

kingjia90:
> Shouldn't it check both?

I was wondering the same, but as what Bernie mentioned [here](https://github.com/pimcore/pimcore/issues/12335#issuecomment-1144791786) and if i understood correctly, the DotEnv symfony component already manages it, in a way that `$_SERVER` contains all the `$_ENV` values without overriding $_SERVER existing keys thanks to the [union `+` operator](https://github.com/symfony/symfony/blob/62c10e7b1fd743f77df5e4eb7db85975e649a057/src/Symfony/Component/Dotenv/Dotenv.php#L151), making it safe to directly use $_SERVER while still having precedence and fallback over/to any $_ENV values.

jdreesen:
But you can't rely on DotEnv being present, as it can be deactivated since #12099 and may [become entirely optional](https://github.com/pimcore/pimcore/issues/11411) in the future.

kingjia90:
> But you can't rely on DotEnv being present, as it can be deactivated since #12099 and may [become entirely optional](https://github.com/pimcore/pimcore/issues/11411) in the future.

ahh i see, i wasn't aware of that planned changes. Knowing that, then it won't hurt to have both. 
Thank you for the heads-up!

brusch:
@kingjia90 but maybe we can just do `$_SERVER += $_ENV;` in Pimcore's code, as it doesn't hurt and saves checks for both `$_SERVER` and `$_ENV` in several locations ....

kingjia90:
🙈 Messed up this PR, wrong base branch, opened a new PR #12384

PR #12376, opened 2022-06-09
Static Class Builder should not call "classSaved"

dpfaffenbauer:
"classSaved" should not be called during "static" class building where no database might be available. It should be called during the "ClassDefinition" save function


brusch:
Thanks a lot! 👍

PR #12384, opened 2022-06-09
[Bug]: LocationAwareConfigRepository to check $_SERVER instead on $_ENV

### LINKED ISSUE #12335: [Bug]: LocationAwareConfigRepository doesn't check SERVER

kingjia90:
Fixes https://github.com/pimcore/pimcore/issues/12335

### DIFF
--- a/lib/Bootstrap.php
+++ b/lib/Bootstrap.php
@@ -147,7 +147,11 @@ class Bootstrap
     private static function prepareEnvVariables()
     {
         if (!($_SERVER['PIMCORE_SKIP_DOTENV_FILE'] ?? false)) {
-            (new Dotenv())->bootEnv(PIMCORE_PROJECT_ROOT .'/.env');
+            if (class_exists('Symfony\Component\Dotenv\Dotenv')) {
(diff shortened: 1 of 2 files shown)


kingjia90:
The places that still use $_SERVER ?? $_ENV (not covered by `prepareEnvVariables()` because happening before or outside that) are like setProjectRoot (but is a BC layer) and in installation script, that do not need to be changed. 
The rest most already use $_SERVER without checking $_ENV, because of the array union, so it should be also used when DotEnv is not set, here i just add sort of "placeholder" for 11 when DotEnv would become optional

PR #12385, opened 2022-06-09
Deprecated `RuntimeCacheTrait` and replaced its usage

### LINKED ISSUE #11339: [Improvement]: Use `RuntimeCacheTrait` whenever possible instead of dealing with `Cache\Runtime` directly

aashan10:
Resolves #11339

## Additional info  
As per the discussions on the mentioned issue, `RuntimeCacheTrait` has been deprecated and its usage has been removed from the core.


lib/Cache/RuntimeCacheTrait.php
```diff
@@ -46,6 +52,7 @@ public static function getCacheEnabled(): bool
      *
      * @param mixed $config
      * @param string $cacheKey
+     * @deprecated
```
dvesh3, line 65: we should trigger a deprecation warning in methods like:
```
trigger_deprecation(
    'pimcore/pimcore',
    '10.5.0',
    sprintf('%s is deprecated and will be removed in Pimcore 11. Use \Pimcore\Cache\Runtime::set() instead.', __METHOD__)
);
```

aashan10: Thanks! Will change this accordingly!

aashan10: `RuntimeCacheTrait` has been deprecated with deprecation warnings

lib/Cache/RuntimeCacheTrait.php
```diff
@@ -18,26 +18,42 @@
...
     /**
      * @return bool
+     * @deprecated
      */
     public static function getCacheEnabled(): bool
     {
+        trigger_deprecation(
+            'pimcore/pimcore',
+            '10.5.0',
+            sprintf('%s is deprecated and will be removed in Pimcore 11. Use \Pimcore\Cache\Runtime::set() instead.', __METHOD__)
```
dvesh3: Please remove "Use \Pimcore\Cache\Runtime::set() instead."

lib/Cache/RuntimeCacheTrait.php
```diff
@@ -18,26 +18,42 @@
...
     /**
      * @param bool $cacheEnabled
+     * @deprecated
      */
     public static function setCacheEnabled(bool $cacheEnabled): void
     {
+        trigger_deprecation(
+            'pimcore/pimcore',
+            '10.5.0',
+            sprintf('%s is deprecated and will be removed in Pimcore 11. Use \Pimcore\Cache\Runtime::set() instead.', __METHOD__)
```
dvesh3: Please remove "Use \Pimcore\Cache\Runtime::set() instead."

lib/Cache/RuntimeCacheTrait.php
```diff
@@ -60,16 +82,23 @@ private static function setCache(mixed $config, string $cacheKey): void
...
      *
      * @param string $cacheKey
+     * @deprecated
      */
     private static function removeCache(string $cacheKey): void
     {
+        trigger_deprecation(
+            'pimcore/pimcore',
+            '10.5.0',
+            sprintf('%s is deprecated and will be removed in Pimcore 11. Use \Pimcore\Cache\Runtime::set() and \Pimcore\Cache::remove() instead.', __METHOD__)
```
dvesh3:
```suggestion
            sprintf('%s is deprecated and will be removed in Pimcore 11. Use \Pimcore\Cache\Runtime::set() with null value and \Pimcore\Cache::remove() instead.', __METHOD__)
```

lib/Cache/RuntimeCacheTrait.php
aashan10: Thanks! This is resolved now!

lib/Cache/RuntimeCacheTrait.php
aashan10: Thanks! This is resolved now!

dvesh3 (approved):

dvesh3:
@aashan10 one last thing.. please add it to the upgrade notes here https://github.com/pimcore/pimcore/blob/10.x/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md#1050

aashan10:
@dvesh3 thank you! I've added the changes to the upgrade notes as well!

PR #12389, opened 2022-06-10
Fix updating application_logs table entries in LogMailMaintenanceTask

michaelhader:
This PR fixes the update process of the rows in the `application_logs` table by also checking the `maintenanceChecked` column against NULL. Otherwise the rows in the `application_logs` table will never get their `maintenanceChecked` column set to `1` and will then be sent on every Application Logger mail update.


mcop1:
Thank you!

PR #12391, opened 2022-06-10
[Bug]: Cleanup FC and Bricks Stores on lower-case DB Servers

### LINKED ISSUE #12101: [Bug]: Cleanup FC and Bricks Stores on lower-case DB Servers
Pimcore's cleanup tasks checks for tables like 'object_collection_%' and 'object_brick_%',  extrudes the brick/collection name from that and checks if the brick/collection in Pimcore exists. That doesn't work with Database Servers that are lower-case by default. Pimcore then logs "Fieldcollection 'coreshoptaxitem' not found. Please check table object_collection_coreshoptaxitem_cs_order_item".

Pimcore should rather ignore the case for this task.

robertSt7:
resolves #12101

### DIFF
--- a/tests/Model/DataObject/DefinitionModifierTest.php
+++ b/tests/Model/DataObject/DefinitionModifierTest.php
@@ -489,8 +489,7 @@ class DefinitionModifierTest extends ModelTestCase
     public function testGetByIdIgnoreCaseWithValidId()
     {
         $id = 'Inheritance';
-        $classDef = new ClassDefinition();
-        $checkVal = $classDef->getByIdIgnoreCase($id);
+        $checkVal = ClassDefinition::getByIdIgnoreCase($id);
 
         $this->assertNotEmpty($checkVal);
     }
--- a/tests/Model/DataObject/DefinitionModifierTest.php
+++ b/tests/Model/DataObject/DefinitionModifierTest.php
@@ -480,8 +480,7 @@ class DefinitionModifierTest extends ModelTestCase
     public function testGetByIdIgnoreCaseWithoutValidId()
     {
             $id = '-9999';
-            $classDef = new ClassDefinition();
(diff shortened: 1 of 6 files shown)


tests/Model/DataObject/DefinitionModifierTest.php
```diff
@@ -476,4 +476,22 @@ private static function getMethod($class, $name)
...
+            $checkVal = $classDef->getByIdIgnoreCase($id);
+
+            $this->assertNull($checkVal);
+    }
+
+    public function testGetByIdIgnoreCaseWithValidId()
+    {
+        $id = 'Inheritance';
+        $classDef = new ClassDefinition();
+        $checkVal = $classDef->getByIdIgnoreCase($id);
```
mcop1, line 493:
```suggestion
        $checkVal = ClassDefinition::getByIdIgnoreCase($id);
```

mcop1:
```suggestion
        $checkVal = ClassDefinition::getByIdIgnoreCase($id);
```

tests/Model/DataObject/DefinitionModifierTest.php
```diff
@@ -476,4 +476,22 @@ private static function getMethod($class, $name)
 
         return $method;
     }
+
+    public function testGetByIdIgnoreCaseWithoutValidId()
+    {
+            $id = '-9999';
+            $classDef = new ClassDefinition();
+            $checkVal = $classDef->getByIdIgnoreCase($id);
```
mcop1:
```suggestion
            $checkVal = ClassDefinition::getByIdIgnoreCase($id);
```

mcop1 (commented):
Since the method is static I would call it that way. What do you think?

PR #12396, opened 2022-06-13
[Task] Add @param and @return php doc types to /lib/Twig

### LINKED ISSUE #12236: [Task] Add @param and @return php doc types to /lib/Twig
### Improvement description

- [ ] Add @param php doc types for all public/protected classes/methods/properties in /lib/Twig (excl. Dao and @internal) 
- [ ] Add @return php doc types for all public/protected classes/methods/properties in /lib/Twig (excl. Dao and @internal) 

kingjia90:
Resolves #12236

### DIFF
--- a/lib/Twig/Extension/Templating/HeadLink.php
+++ b/lib/Twig/Extension/Templating/HeadLink.php
@@ -123,7 +123,7 @@ class HeadLink extends CacheBusterAware
      *
      * @return HeadLink
      */
-    public function __invoke(array $attributes = null, $placement = Container::APPEND)
+    public function __invoke(?array $attributes = null, $placement = Container::APPEND)
     {
         if (null !== $attributes) {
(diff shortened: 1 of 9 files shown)


lib/Twig/Extension/Templating/HeadLink.php
```diff
@@ -123,7 +123,7 @@ public function __construct(
      *
      * @return HeadLink
      */
-    public function __invoke(array $attributes = null, $placement = Container::APPEND)
+    public function __invoke(?array $attributes = null, $placement = Container::APPEND)
```
mcop1, line 128: Maybe we should add @param tags for the parameters too, since $placement doesn´t have a type assigned yet? Wdyt?

kingjia90: thank you! nice catch, i've added the missing params

PR #12401, opened 2022-06-13
[Task] Add @param and @return php doc types to /lib/Workflow

### LINKED ISSUE #12237: [Task] Add @param and @return php doc types to /lib/Workflow
### Improvement description

- [ ] Add @param php doc types for all public/protected classes/methods/properties in /lib/Workflow  (excl. Dao and @internal) 
- [ ] Add @return php doc types for all public/protected classes/methods/properties in /lib/Workflow (excl. Dao and @internal) 

aryaantony92:
Resolves #12237

### DIFF
--- a/lib/Workflow/GlobalAction.php
+++ b/lib/Workflow/GlobalAction.php
@@ -108,11 +108,11 @@ class GlobalAction implements NotesAwareInterface
 
     /**
      * @param Workflow $workflow
-     * @param mixed $subject
+     * @param object $subject
      *
-     * @return bool
+     * @return mixed
      */
-    public function isGuardValid(Workflow $workflow, $subject): bool
+    public function isGuardValid(Workflow $workflow, $subject): mixed
     {
         if (empty($this->getGuard())) {
             return true;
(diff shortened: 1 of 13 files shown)


lib/Workflow/GlobalAction.php
```diff
@@ -108,11 +108,11 @@ public function getGuard(): ?string
...
     /**
      * @param Workflow $workflow
-     * @param mixed $subject
+     * @param object $subject
      *
-     * @return bool
+     * @return mixed
      */
-    public function isGuardValid(Workflow $workflow, $subject): bool
+    public function isGuardValid(Workflow $workflow, $subject): mixed
```
jdreesen: Should an "isser" really return `mixed`? It seems strange to me, I'd always assume it returns a `bool`.

aryaantony92: @jdreesen I changed it as "mixed", because evaluateExpression() returns "mixed"

https://github.com/pimcore/pimcore/pull/12401/files/4ce0015c82735579a17ea5648b973e3bd4cc847a#:~:text=%2D%3EgetGuard())%3B-,return%20%24this%2D%3EexpressionService%2D%3EevaluateExpression(%24workflow%2C%20%24subject%2C%20%24this%2D%3EgetGuard())%3B,-%7D

jdreesen: I know, but is this right? What does it evaluate? Maybe its result should be cast to `bool`?

aryaantony92: @jdreesen I will check it.

PR #12404, opened 2022-06-13
Transfer permissions as boolvars

spotbot2k:
The user permissions are transferded as an in converted to string, it causes many wrong positive checcks as "0" is evaluated as true by chrome and firefox. So a user gets menu entries to

Create objects
Sort objects
Move objects
etc.
without havin a permission for it. The permission is evaluated in Backend on action, so it does not give the user any more rights but it does cause exceptions and errors.

### DIFF
--- a/models/Element/AbstractElement.php
+++ b/models/Element/AbstractElement.php
@@ -213,7 +213,7 @@ abstract class AbstractElement extends Model\AbstractModel implements ElementInt
         $columns = array_diff(array_keys($vars), $ignored);
 
         foreach ($columns as $name) {
-            $permissions[$name] = 1;
+            $permissions[$name] = true;
         }
 
         if (null === $user) {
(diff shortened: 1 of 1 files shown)


models/Element/AbstractElement.php
```diff
@@ -229,7 +229,7 @@ public function getUserPermissions(User $user = null)
             $event = new ElementEvent($this, ['isAllowed' => $isAllowed, 'permissionType' => $type, 'user' => $user]);
             \Pimcore::getEventDispatcher()->dispatch($event, AdminEvents::ELEMENT_PERMISSION_IS_ALLOWED);
 
-            $permissions[$type] = $event->getArgument('isAllowed');
+            $permissions[$type] = boolval($event->getArgument('isAllowed'));
```
jdreesen, line 232:
```suggestion
            $permissions[$type] = (bool) $event->getArgument('isAllowed');
```

spotbot2k:
Will be resolved in #12400

PR #12405, opened 2022-06-13
Cookies should be auto-secure

### LINKED ISSUE #12377: [Improvement]: Allow Secure Cookies
### Improvement description

Currently, all cookies created in pimcore are configured as non-secure (6th argument set to `false`):

https://github.com/pimcore/pimcore/search?l=PHP&q=%22new+cookie%22

It should be possible to set this definition via configuration.

brusch:
resolves #12377

Unlike as requested in #12377 I think it would be better to just use Symfony's "auto secure", rather than introducing another config option.

### DIFF
--- a/bundles/AdminBundle/Security/Event/LogoutListener.php
+++ b/bundles/AdminBundle/Security/Event/LogoutListener.php
@@ -115,7 +115,8 @@ class LogoutListener implements EventSubscriberInterface, LoggerAwareInterface
 
         // cleanup pimcore-cookies => 315554400 => strtotime('1980-01-01')
         $response->headers->setCookie(new Cookie('pimcore_opentabs', null, 315554400));
-        $response->headers->clearCookie('pimcore_admin_sid', '/', null, false, true);
+        // clear cookie -> we can't use $response->headers->clearCookie() because it doesn't allow $secure = null
+        $response->headers->setCookie(new Cookie('pimcore_admin_sid', null, 1));
 
         if ($response instanceof RedirectResponse) {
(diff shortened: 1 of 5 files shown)


bundles/AdminBundle/Security/Event/LogoutListener.php
```diff
@@ -114,7 +114,7 @@ public function onLogoutSuccess(Request $request): RedirectResponse|Response
         }
 
         // cleanup pimcore-cookies => 315554400 => strtotime('1980-01-01')
-        $response->headers->setCookie(new Cookie('pimcore_opentabs', null, 315554400, '/'));
+        $response->headers->setCookie(new Cookie('pimcore_opentabs', null, 315554400));
         $response->headers->clearCookie('pimcore_admin_sid', '/', null, false, true);
```
aashan10: I'm not sure but does the auto-secure cookies get cleared when `$secure` parameter is passed `false`?
As far as I remember, it needs to be null in order to set an auto-secure cookie.

brusch: @aashan10 I'd remove it here as well and let the default value, as we do using `setCookie()`

brusch: Oh no ... tricky one ... `clearCookie()` has `$secure = false` as the default, so we need to set it to `null` and cannot rely on the defaults.

brusch: Fixed, good catch, thanks! 👍

aashan10 (approved):

PR #12414, opened 2022-06-14
perf: add file size, width and height to thumbnail cache

### LINKED ISSUE #12055: [Bug]: Pimcore serving deferred thumbnail URL even though the thumbnail exists (Azure Storage used)

### LINKED ISSUE #12057: [Bug]: \Pimcore\Model\Asset\Thumbnail\ImageThumbnailTrait::getDimensions is reading dimensions from file every time
### Expected behavior

With a remote storage, reading remote files just to find out the thumbnail width / height is suboptimal. The asset cache could cache these values.

### Actual behavior

The values are read from the file which is pulled from remote storage.

### Steps to reproduce

1. have a remote storage for thumbnails set up
2. call `\Pimcore\Twig\Extension\ImageThumbnailExtension::getImageThumbnailHtml`
3. `\Pimcore\Model\Asset\Thumbnail\ImageThumbnailTrait::readDimensionsFromFile` is triggered, which pulls the remote file and reads width / height directly from it

This also applies to filesize, maybe MIME type?

dkarlovi:

Closes #12057.
Closes #12055.

### DIFF
--- a/models/Asset/Dao.php
+++ b/models/Asset/Dao.php
@@ -546,20 +546,20 @@ class Dao extends Model\Element\Dao
     public function addToThumbnailCache(string $name, string $filename, int $filesize, int $width, int $height): void
     {
         $assetId = $this->model->getId();
-        $time = time();
-        $this->db->insertOrUpdate('assets_image_thumbnail_cache', [
+        $thumb = [
             'cid' => $assetId,
             'name' => $name,
             'filename' => $filename,
-            'modificationDate' => $time,
+            'modificationDate' => time(),
             'filesize' => $filesize,
             'width' => $width,
             'height' => $height,
-        ]);
+        ];
+        $this->db->insertOrUpdate('assets_image_thumbnail_cache', $thumb);
 
         if (isset(self::$thumbnailStatusCache[$assetId])) {
             $hash = $name . $filename;
-            self::$thumbnailStatusCache[$assetId][$hash] = $time;
+            self::$thumbnailStatusCache[$assetId][$hash] = $thumb;
         }
     }
 
--- a/models/Asset/Dao.php
+++ b/models/Asset/Dao.php
@@ -543,8 +543,14 @@ class Dao extends Model\Element\Dao
         return false;
     }
 
-    public function addToThumbnailCache(string $name, string $filename, int $filesize, int $width, int $height): void
+    public function addToThumbnailCache(string $name, string $filename, ?int $filesize = null, ?int $width = null, ?int $height = null): void
     {
+        if ($filesize === null || $width === null || $height === null) {
+            trigger_deprecation('pimcore/pimcore', '10.5',
+                sprintf('Not passing "filesize", "width" and "height" arguments to %s is deprecated and will be removed in Pimcore 11.', __METHOD__)
+            );
+        }
+
         $assetId = $this->model->getId();
         $thumb = [
             'cid' => $assetId,
--- a/models/Asset/Dao.php
+++ b/models/Asset/Dao.php
@@ -543,7 +543,7 @@ class Dao extends Model\Element\Dao
         return false;
     }
 
-    public function addToThumbnailCache(string $name, string $filename): void
+    public function addToThumbnailCache(string $name, string $filename, int $filesize, int $width, int $height): void
     {
         $assetId = $this->model->getId();
         $time = time();
@@ -552,6 +552,9 @@ class Dao extends Model\Element\Dao
             'name' => $name,
             'filename' => $filename,
             'modificationDate' => $time,
+            'filesize' => $filesize,
+            'width' => $width,
+            'height' => $height,
         ]);
 
         if (isset(self::$thumbnailStatusCache[$assetId])) {
@@ -561,6 +564,17 @@ class Dao extends Model\Element\Dao
     }
 
     public function getCachedThumbnailModificationDate(string $name, string $filename): ?int
+    {
+        $thumbnail = $this->getCachedThumbnail($name, $filename);
+
+        if ($thumbnail === null) {
+            return null;
+        }
+
+        return $thumbnail['modificationDate'];
+    }
+
+    public function getCachedThumbnail(string $name, string $filename): ?array
     {
         $assetId = $this->model->getId();
 
@@ -574,7 +588,7 @@ class Dao extends Model\Element\Dao
 
             foreach ($thumbs as $thumb) {
                 $hash = $thumb['name'] . $thumb['filename'];
-                self::$thumbnailStatusCache[$assetId][$hash] = $thumb['modificationDate'];
+                self::$thumbnailStatusCache[$assetId][$hash] = $thumb;
             }
         }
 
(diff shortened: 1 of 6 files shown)


dkarlovi:
@brusch can the test suite even run locally? I'm getting all sorts of weird Composer errors about missing Kernel.php

The feature works as expected from my testing.

fashxp:
does [that](https://pimcore.com/docs/pimcore/current/Development_Documentation/Development_Tools_and_Details/Testing/Core_Tests.html#page_Requirements) help?

dkarlovi:
@fashxp it does, thanks!

dkarlovi:
@fashxp could you check / review the PR and see what the cause of test failures might be?

brusch:
@dkarlovi see https://github.com/pimcore/pimcore/runs/6881340900?check_suite_focus=true#step:11:3260

dkarlovi:
@brusch I have the tests running locally, but do you have a suggestion / workflow how to most easily fix them? The test run takes a long time, it would take me days to figure out what's wrong here doing it like that. Even filtered by group, it's slow.

brusch:
@dkarlovi if you look into the log of the tests here on the failing GitHub Action, you can see that the tests are failing here: 
https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/tests/Model/Asset/AssetTest.php#L103
and here: 
https://github.com/pimcore/pimcore/blob/9826eafc6bf22b3e781ad6b20b870aa7f03e7020/tests/Model/Asset/AssetThumbnailCacheTest.php#L77

So I'd assume there's something wrong with the returned values for width. 
Should be easy to reproduce in a custom controller or command if you don't want to run all the tests (just copy the code from the tests) 😊

dkarlovi:
@brusch OK, I hoped there's some "Run this test fast" method in Pimcore's test setup I didn't know about. I'll work on this right away.

dkarlovi:
@brusch I've missed the `install.sql` file :see_no_evil: should pass now, it does for me locally.

dkarlovi:
OK, ready for review.

brusch:
@dkarlovi thanks man! 🙋

fashxp:
> @brusch OK, I hoped there's some "Run this test fast" method in Pimcore's test setup I didn't know about. I'll work on this right away.

just for the record. Codeception also allows you to run single tests: https://codeception.com/docs/02-GettingStarted

dkarlovi:
@fashxp I think the biggest issue here is something was swallowing the error since the test environment was missing the new fields in the table and nothing was complaining specifically about that, it only noticed the side-effects (still good, but not great).

dkarlovi:
@brusch with the latest fix it closes #12055.

models/Asset/Dao.php
```diff
@@ -543,24 +543,44 @@ public function __isBasedOnLatestData()
         return false;
     }
 
-    public function addToThumbnailCache(string $name, string $filename): void
+    public function addToThumbnailCache(string $name, string $filename, ?int $filesize = null, ?int $width = null, ?int $height = null): void
     {
+        if ($filesize === null || $width === null || $height === null) {
+            trigger_deprecation('pimcore/pimcore', '10.5',
```
brusch, line 549: As this is an internal function/class we don't need to issue a deprecation message, we can just make the new arguments required.

models/Asset/Image/Thumbnail/Processor.php
```diff
@@ -460,7 +462,11 @@ public static function process(Asset $asset, Config $config, $fileSystemPath = n
         }
 
         if ($statusCacheEnabled) {
-            $asset->getDao()->addToThumbnailCache($config->getName(), $filename);
+            if (isset($filesize) && isset($width) && isset($height)) {
+                $asset->getDao()->addToThumbnailCache($config->getName(), $filename, $filesize, $width, $height);
```
brusch, line 466: We should be able to read the dimensions from the `$image` instance, here: 
https://github.com/pimcore/pimcore/blob/5ac4a712f4aee61a3b8648f8370a8162e984b231/models/Asset/Image/Thumbnail/Processor.php#L426
So we should always have the dimensions available as well as the filesize.

models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -194,6 +194,13 @@ private function readDimensionsFromFile(): array
                             'width' => $info[0],
                             'height' => $info[1],
                         ];
+
+                        $filesize = filesize($localFile);
+                        $config = $this->getConfig();
+                        $asset = $this->getAsset();
+                        $filename = pathinfo($pathReference['storagePath'], PATHINFO_FILENAME);
+
+                        $asset->getDao()->addToThumbnailCache($config->getName(), $filename, $filesize, $dimensions['width'], $dimensions['height']);
```
brusch, line 203: This shouldn't be necessary anymore then, if we write that already in the processor.

brusch (changes requested):
Should we also consider the mime-type?

dkarlovi:
@brusch you've said the MIME type is not required since it's deducible from the filename.

models/Asset/Image/Thumbnail/Processor.php
dkarlovi: PHPStan warns that you might not reach the same code route. Meaning, your code might run into the lock which gets released.

models/Asset/Thumbnail/ImageThumbnailTrait.php
dkarlovi: This is required to retroactively populate the table, no?

brusch: You are right, that could be useful.

models/Asset/Image/Thumbnail/Processor.php
brusch: But it should be still possible when the image got written ... maybe it requires some more ifs for Stan, but in case the thumb was written to the local filesystem, there should be also a width, height and filesize, no?

dkarlovi: I think we need to figure out which paths actually write the image and which don't. The null check is basically that, but direct instead of indirect.

If you have suggestions how to fix this exactly, I'll be glad to do it.

dkarlovi: In other words, we need to be able to guarantee the variables are available when storing since we'll make them mandatory and non-nullable.

brusch: @dkarlovi see #12461 let me know what you think. Thanks!

dkarlovi: LGTM!

brusch:
Closed in favor of #12461

PR #12416, opened 2022-06-14
[FIX]: Filter for object brick attributes does not work

### LINKED ISSUE #12293: [Bug]: Filter for object brick attributes does not work
### Expected behavior

Sort or filter on an attribute of an object brick in the grid view.

### Actual behavior

An error message pops up:
image

### Steps to reproduce

1. Open a gridview for a class that contains objectbricks
2. Add a random attribute (from an objectbrick) to the grid view. (in my case this was a WYSIWYG-field)
3. Try to sort or filter on it.

Pimcore-version: 10.3.3

dturopoli:
Resolves #12293

Issue was caused by localized fields inside object brick

### DIFF
--- a/bundles/AdminBundle/Helper/GridHelperService.php
+++ b/bundles/AdminBundle/Helper/GridHelperService.php
@@ -295,7 +295,7 @@ class GridHelperService
                             foreach ($filter['value'] as $filterValue) {
                                 $brickCondition = '(' . $brickField->getFilterCondition($filterValue, $operator,
                                         ['brickPrefix' => $brickPrefix]
-                                    ) . ' AND ' . $brickPrefix . 'fieldname = ' . $db->quote($brickFilterField) . ')';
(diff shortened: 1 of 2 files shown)


kingjia90:
Thank you for the PR! 

I've review it, filtering works again, but Sorting is not fixed. It seems expecting a key with a question mark for localized Fields. 

I've also noticed with the classTree changes, the following case (highlighted in yellow) cannot be dragged in Grid Options anymore. Could you also double check this?
image

kingjia90:
Notes: might also need to take in consideration this https://github.com/pimcore/pimcore/pull/12349

kingjia90:
@dturopoli Just a gentle reminder 🏓

dturopoli:
@kingjia90 I think d&d is working now. I'm not sure about Sorting, I can't reproduce bug, it is working for me. Maybe I'm missing some test case

dvesh3:
@dturopoli thanks!

PR #12417, opened 2022-06-14
Children grid doesn't open when there are no select options

ultramegatom:
I had a class where field.layout.options was null. I got a js error and the Grid didnt open.
Im not exactly sure why it was null, maybe it has to do with a faulty OptionsProvider used for that select, but i think this additional check doesent hurt anyone.


dvesh3:
@ultramegatom a bug fix should be on 10.4 branch. thanks!

ultramegatom:
@dvesh3 https://github.com/pimcore/pimcore/pull/12455

PR #12418, opened 2022-06-14
[Bug]: Fix batch append/remove doesn't work on CS attributes

### LINKED ISSUE #12313: [Bug]: Batch append doesn't work on CS attributes
### Expected behavior

- Add multiselect classification store attributes to Data Object list via Grid Options
- Batch Append To all/select should keep existing maintenance and only add the new options

### Actual behavior

- Can't apply grid config changes after adding CS multiselect attribute:

image

- Batch Append To all/select replace the existing maintenance instead of only add the new options in case of CS store multiselect attributes

### Steps to reproduce

- Create CS store
- Create test multiselect attribute with few options
- Add it to Class
- Open Grid and try to add above attribute in grid
- Try to use Batch append option

robertSt7:
resolve #12313

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -1619,6 +1619,12 @@ class DataObjectHelperController extends AdminController
 
                                 /** @var DataObject\Classificationstore $classificationStoreData */
                                 $classificationStoreData = $object->$getter();
+                                if ($append) {
+                                    $oldValues = $classificationStoreData->getLocalizedKeyValue($groupId, $keyId);
+                                    if(is_array($oldValues)) {
+                                        $value = array_merge($oldValues, $value);
+                                    }
+                                }
                                 $classificationStoreData->setLocalizedKeyValue(
                                     $groupId,
                                     $keyId,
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
```diff
@@ -1619,6 +1619,12 @@ public function batchAction(Request $request)
 
                                 /** @var DataObject\Classificationstore $classificationStoreData */
                                 $classificationStoreData = $object->$getter();
+                                if ($append) {
+                                    $oldValues = $classificationStoreData->getLocalizedKeyValue($groupId, $keyId);
+                                    if(is_array($oldValues)) {
+                                        $value = array_merge($oldValues, $value);
+                                    }
+                                }
```
dvesh3: we should use `appendData` (to avoid duplicate values) and `removeData` (to remove specific value)
```suggestion
                                if ($append) {
                                    $oldValue = $classificationStoreData->getLocalizedKeyValue($groupId, $keyId);
                                    $value = $dataDefinition->appendData($oldValue, $value);
                                }
                                if ($remove) {
                                    $oldValue = $classificationStoreData->getLocalizedKeyValue($groupId, $keyId);
                                    $value = $dataDefinition->removeData($oldValue, $value);
                                }
```

PR #12420, opened 2022-06-14
[Improvement] add name to export / set name of custom layout on import if na…

### LINKED ISSUE #5023: Add CustomLayout "name" to CustomLayout  JSON export
Hello,

We are using CustomLayout export / import and i wonder why the "name" field is not exported : 

https://github.com/pimcore/pimcore/blob/b9cb305e60518a3b4968d9227fb1f8a11603b43d/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php#L631

It's would be easier instead of retrieving the name from the filename...

Thanks for considering this.

dpfahlbusch:
…me is available in custom layout json file

Resolves #5023


kingjia90 (commented):
Thank you for the PR!

It works, but there's one case: when you set the same name of an existing custom layout, the error thrown is unclear. 
image
Could you please add a check if the name is an existing or improve the error message a little bit mentioning the possible causes?

aryaantony92:
@dpfahlbusch Just a gentle reminder. Could you please add a check if the name already exists or improve the error message a little bit mentioning the possible causes?

bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/customLayoutEditor.js
```diff
@@ -997,7 +997,13 @@ pimcore.object.helpers.customLayoutEditor = Class.create({
 
             });
         }.bind(this), function () {
```
aryaantony92:
```suggestion
        }.bind(this), function (response) {
```

bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
```diff
@@ -706,6 +721,7 @@ public function exportCustomLayoutDefinitionAction(Request $request)
             if ($customLayout) {
                 $name = $customLayout->getName();
                 $customLayoutData = [
+                    'name' => $customLayout->getName(),
```
aryaantony92:
```suggestion
                    'name' => $name,
```

aryaantony92:
@dpfahlbusch Thank you for the PR

PR #12422, opened 2022-06-15
Fix project root determination case in `pimcore-install`

jdreesen:
Although I didn't run in this case, I'm pretty sure it was wrong.


brusch:
Thanks a lot! 👍

PR #12428, opened 2022-06-15
[Core] Migrate Custom Layout definition files to use LocationAwareConfigRepository 

### LINKED ISSUE #12321: [Improvement]: Migrate Custom Layout definition files to use LocationAwareConfigRepository
### Improvement description

Currently the ~~[PHP classes definitions](https://pimcore.com/docs/pimcore/current/Development_Documentation/Deployment/Deployment_Tools.html#page_Pimcore-Class-Definitions) and~~ custom layout are saved in `/var/classes` but it should be using `LocationAwareConfigRepository` (which is already in use for [perspective/custom view](https://github.com/pimcore/pimcore/pull/10564/files)) as it will provide more flexibility and legacy support until Pimcore 11.

See https://github.com/pimcore/pimcore/issues/12262 where the filesystem is read-only it, it may cause problems.

Also: `PIMCORE_CLASS_DEFINITION_WRITABLE` var should not have any effect on custom layout screens.

dvesh3:
Resolves  #12321


models/DataObject/ClassDefinition/CustomLayout/Listing/Dao.php
```diff
@@ -53,9 +51,41 @@ public function load()
...
+
+        return array_merge($list, array_keys($this->loadLegacyConfigs()));
+    }
+
+    /**
+     * @TODO remove in Pimcore 11
+     *
+     * @return array
+     */
+    private function loadLegacyConfigs(): array
```
kingjia90, line 77:
```suggestion
    private function loadLegacyConfigs(): array|null
```
should it be like this? in the case that there are no existing custom layouts, it returns `null`.
In case, also need to adapt the phpdoc block and add a null check `loadIdList`

bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php
```diff
@@ -259,6 +260,9 @@ public function getCustomLayoutAction(Request $request)
                     );
 
                     $customLayout->setId($request->get('id'));
+                    if (!$customLayout->isWriteable()) {
+                        throw new ConfigWriteException();
```
kingjia90, line 264: phpdocblock should have `* @throws ConfigWriteException|DataObject\Exception\DefinitionWriteException`

doc/Development_Documentation/21_Deployment/03_Configuration_Environments.md
```diff
@@ -81,6 +82,7 @@ PIMCORE_CONFIG_STORAGE_DIR_PREDEFINED_ASSET_METADATA=/var/www/html/var/config/pr
 PIMCORE_CONFIG_STORAGE_DIR_STATICROUTES=/var/www/html/var/config/staticroutes
 PIMCORE_CONFIG_STORAGE_DIR_PERSPECTIVES=/var/www/html/var/config/perspectives
 PIMCORE_CONFIG_STORAGE_DIR_CUSTOM_VIEWS=/var/www/html/var/config/custom-views
+PIMCORE_CONFIG_STORAGE_DIR_OBJECT_CUSTOM_LAYOUTS=/var/www/html/var/config/custom-layouts
```
kingjia90, line 86: works fine and looks good. Just need to add `PIMCORE_WRITE_TARGET_OBJECT_CUSTOM_LAYOUTS=settings-store` on line 67 in the list of Available environment variables and in tests [_boostrap.php](https://github.com/pimcore/pimcore/blob/10.4/tests/_bootstrap.php#L41)

models/DataObject/ClassDefinition/CustomLayout/Listing/Dao.php
dvesh3: I think the $layout value should be changed to `$layouts = [];` instead, as this is passed to array_keys..

kingjia90 (changes requested):

bundles/CoreBundle/Migrations/Version20220617145524.php
```diff
@@ -0,0 +1,38 @@
...
+final class Version20220617145524 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Drop custom_layouts table in favor of LocationAwareConfigRepository';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('DROP TABLE IF EXISTS custom_layouts');
```
kingjia90, line 19: As discussed in the daily, this needs to be changed from `drop` to rename table with prefix 'to_be_delete` or so, if it cannot be easily rebuild/restored as it stores extra data.

dvesh3: Added warning on downgrade to run command `pimcore:deployment:custom-layouts-rebuild -c` manually as complete data can be restored from def. files.

kingjia90 (approved):

PR #12434, opened 2022-06-17
Don't cache thumbnail urls with error image

blankse:
If a deferred thumbnail couldn't be generated a error image is returned. But browser cache this urls. It should be only redirected to the error image. So the browser can receive the correct thumbnail when it could be generated.


dkarlovi:
Option is also to just disable caching in this case.

brusch:
LGTM, thanks!

PR #12439, opened 2022-06-20
Fix: Error in GeneratePagePreviewHandler

blankse:
Error in GeneratePagePreviewHandler if HtmlToImage::convert() doesn't work and returns false.
```
08:39:37 ERROR     [pimcore] Unable to load image: /var/www/html/var/tmp/screenshot_tmp_2122.png
08:39:37 ERROR     [pimcore] unable to open image `/var/www/html/var/tmp/screenshot_tmp_2122.png': No such file or directory @ error/blob.c/OpenBlob/2924
08:39:37 CRITICAL  [messenger] Error thrown while handling message Pimcore\Messenger\GeneratePagePreviewMessage. Removing from transport after 3 retries. Error: "Handling "Pimcore\Messenger\GeneratePagePreviewMessage" failed: Call to a member function stripImage() on null" ["message" => Pimcore\Messenger\GeneratePagePreviewMessage^ { …},"class" => "Pimcore\Messenger\GeneratePagePreviewMessage","retryCount" => 3,"error" => "Handling "Pimcore\Messenger\GeneratePagePreviewMessage" failed: Call to a member function stripImage() on null","exception" => Symfony\Component\Messenger\Exception\HandlerFailedException^ { …}]
```

### DIFF
--- a/lib/Image/Chromium.php
+++ b/lib/Image/Chromium.php
@@ -17,6 +17,7 @@ namespace Pimcore\Image;
 
 use HeadlessChromium\BrowserFactory;
 use HeadlessChromium\Communication\Message;
+use Pimcore\Logger;
 use Pimcore\Tool\Console;
 use Pimcore\Tool\Session;
 use Symfony\Component\HttpFoundation\Session\Attribute\AttributeBagInterface;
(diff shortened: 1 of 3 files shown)


dvesh3:
@blankse thanks for your PR! could you please share steps to reproduce?

blankse:
@dvesh3 I think a missing "Settings" -> "System" -> "Website" -> "Domain" setting should be enough. Than the url has no Host.

If the convert method returns false, the output file doesn't exist and the Imagemagick save method throws a "Call to a member function stripImage() on null" error, because the resource attribute is null.

dvesh3:
@blankse thanks!

PR #12443, opened 2022-06-20
[Improvement]: Remove/deprecate some folder creation by MkdirCacheWarmer

### LINKED ISSUE #12322: [Improvement]: Remove/deprecate folder creation in var folder by MkdirCacheWarmer
### Improvement description

After #12321, any folder created during the Cache warming phase should be removed/deprecated when possible.  

https://github.com/pimcore/pimcore/blob/597b1629f8c990781763a455372715fb2ac9ce9b/lib/HttpKernel/CacheWarmer/MkdirCacheWarmer.php#L69

kingjia90:
Improves #https://github.com/pimcore/pimcore/issues/12322

## Notes
PIMCORE_LOG_MAIL_PERMANENT removed as followup of #11366
PIMCORE_CUSTOMLAYOUT_DIRECTORY should be not necessary after #12428
PIMCORE_LOG_FILEOBJECT_DIRECTORY because the folder is also created in the [constructor](https://github.com/pimcore/pimcore/blob/cae32ab5dea2e5237a798cdfb6b8f18266d9a72b/lib/Log/FileObject.php#L39), and "safely" deprecated after #12460
PIMCORE_CLASS_DEFINITION_DIRECTORY would be optional depending on PIMCORE_CLASS_DEFINITION_WRITABLE (if 0, it's not necessary to create the folder), even though the default folder is the same as PIMCORE_CLASS_DIRECTORY.
Added an `array_unique` so if the folder are the same, it saves from some folder exists checks.


lib/HttpKernel/CacheWarmer/MkdirCacheWarmer.php
```diff
@@ -54,17 +54,19 @@ public function warmUp($cacheDir): array
...
 
+        // Since #12392, PIMCORE_CLASS_DEFINITION_WRITABLE = 0 doesn't allow creation in var/classes but is allowed when not set or 1.
+        if (true == ($_SERVER['PIMCORE_CLASS_DEFINITION_WRITABLE'] ?? true)) {
+            $directories[] = PIMCORE_CLASS_DEFINITION_DIRECTORY;
+        }
+
         $fs = new Filesystem();
-        foreach ($directories as $directory) {
+        $uniqueDirectories = array_unique($directories);
+        foreach ($uniqueDirectories as $directory) {
```
dvesh3:
```suggestion
        foreach (array_unique($directories) as $directory) {
```

PR #12444, opened 2022-06-20
[Security] SQL Injection in Data Hub GraphQL

mcop1:
Resolves https://github.com/pimcore/planning/issues/93

### DIFF
--- a/lib/Model/Listing/AbstractListing.php
+++ b/lib/Model/Listing/AbstractListing.php
@@ -411,7 +411,7 @@ abstract class AbstractListing extends AbstractModel implements \Iterator, \Coun
         if ($groupBy) {
             $this->groupBy = $groupBy;
 
-          if ($qoute && strpos($groupBy, '`') !== 0) {
+          if ($qoute && !str_starts_with($groupBy, '`')) {
                 $quotedParts = [];
                 $parts = explode(",", $groupBy);
                 foreach($parts as $part) {
--- a/lib/Model/Listing/AbstractListing.php
+++ b/lib/Model/Listing/AbstractListing.php
@@ -235,7 +235,7 @@ abstract class AbstractListing extends AbstractModel implements \Iterator, \Coun
                 if ($quote === false) {
                     $this->orderKey[] = $o;
                 } elseif ($this->isValidOrderKey($o)) {
-                    $this->orderKey[] = '`' . $o . '`';
+                    $this->orderKey[] = $this->quoteIdentifier($o);
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


lib/Model/Listing/AbstractListing.php
```diff
@@ -411,8 +411,14 @@ public function setGroupBy($groupBy, $qoute = true)
         if ($groupBy) {
             $this->groupBy = $groupBy;
 
-            if ($qoute && strpos($groupBy, '`') !== 0) {
-                $this->groupBy = '`' . $this->groupBy . '`';
+          if ($qoute && strpos($groupBy, '`') !== 0) {
```
jdreesen:
```suggestion
          if ($qoute && !str_starts_with($groupBy, '`')) {
```

lib/Model/Listing/AbstractListing.php
```diff
@@ -411,8 +411,14 @@ public function setGroupBy($groupBy, $qoute = true)
         if ($groupBy) {
             $this->groupBy = $groupBy;
 
-            if ($qoute && strpos($groupBy, '`') !== 0) {
-                $this->groupBy = '`' . $this->groupBy . '`';
+          if ($qoute && !str_starts_with($groupBy, '`')) {
```
brusch: I'm wondering whether it would make sense to just replace all occurrences of  \`. 
Otherwise it would be still possible to add unquoted stuff, even when `$quote = true`. WDYT?

brusch (changes requested):

lib/Model/Listing/AbstractListing.php
mcop1: Replacing it with what? `quoteIdentifier` escapes the backtick signs, if there are any. So basically it already does a replace.

lib/Model/Listing/AbstractListing.php
```diff
@@ -411,8 +411,14 @@ public function setGroupBy($groupBy, $qoute = true)
...
+                $quotedParts = [];
+                $parts = explode(",", $groupBy);
```
brusch:
```suggestion
          if ($qoute) {
                $quotedParts = [];
                $parts = explode(",", str_replace('`', '', $groupBy));
```
I mean, quoting should be done in any case, no matter if `$groupBy` starts with a  \` character or not.

brusch (changes requested):

lib/Model/Listing/AbstractListing.php
mcop1: Aaah ok, sorry, now I get what you mean. Yes, that would be even better. I will do the same for setOrderKey then.

mcop1:
```suggestion
          if ($qoute) {
                $quotedParts = [];
                $parts = explode(",", trim($groupBy, '`'));
```
Hmm .. don´t want to be petty, but technically it is possible to create columns with a backtick in the name. If we do a str_replace, then we can´t order by these columns anymore. 
Maybe trimming the backticks would be better, since all other backticks will be escaped by `quoteIdentifier`?

brusch: Valid point 😊 LGTM 👍

brusch (approved):

PR #12445, opened 2022-06-20
Add an option "allowedTypes" to the video editable.

### LINKED ISSUE #2999: disable video providers in video editable
# Feature Request 

## Feature description 

It would be great if there was an way to configure the video providers for the video editable. At the moment you can select between an asset, youtube, vimeo and dailymotion. I want to disable dailymotion and vimeo so that only an asset or youtube video can be selected.

timwiechers:
Resolves #2999

This PR adds an "allowedTypes" option to the video editable to limit the selectable video types.

If this parameter is not set or is empty or its content is invalid, all available types are displayed.

### DIFF
--- a/models/Document/Editable/Video.php
+++ b/models/Document/Editable/Video.php
@@ -196,7 +196,6 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
         return 'video';
     }
 
-
     /**
      * @param array $allowedTypes
      *
@@ -1035,6 +1034,10 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
      */
     public function getVideoType()
     {
+        if (empty($this->type) === true) {
+            $this->type = $this->allowedTypes[0];
+        }
+
         return $this->type;
     }
 
@@ -1043,7 +1046,7 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
      */
     public function getVideoAsset()
     {
-        if ($this->getVideoType() == self::TYPE_ASSET) {
+        if ($this->getVideoType() === self::TYPE_ASSET) {
             return Asset\Video::getById($this->id);
         }
 
--- a/models/Document/Editable/Video.php
+++ b/models/Document/Editable/Video.php
@@ -26,6 +26,18 @@ use Pimcore\Tool;
  */
 class Video extends Model\Document\Editable implements IdRewriterInterface
 {
+    public const TYPE_ASSET = 'asset';
+    public const TYPE_YOUTUBE = 'youtube';
+    public const TYPE_VIMEO = 'vimeo';
+    public const TYPE_DAILYMOTION = 'dailymotion';
+
+    public const ALLOWED_TYPES = [
+        self::TYPE_ASSET,
+        self::TYPE_YOUTUBE,
+        self::TYPE_VIMEO,
+        self::TYPE_DAILYMOTION,
+    ];
+
     /**
      * contains depending on the type of the video the unique identifier eg. "http://www.youtube.com", "789", ...
      *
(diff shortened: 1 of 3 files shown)


models/Document/Editable/Video.php
```diff
@@ -36,13 +48,13 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
...
     /**
-     * one of asset, youtube, vimeo, dailymotion
+     * one of self::ALLOWED_TYPES
      *
      * @internal
      *
      * @var string|null
      */
-    protected $type = 'asset';
+    protected $type;
```
dvesh3, line 57: If i call `getVideoType()` before `getData()` this will now return `null` as oppose to returning value 'asset' before.. right? 

we should consider setting `$this->type` from `allowedTypes` when `null` in `getVideoType()`.

dvesh3 (commented):
@timwiechers The PR looks good 😉 and works like a charm. 

I just came across one use case, where a type which is allowed before and then removed from the `allowedTypes` list so what happens to the existing data of that particular type? 
1. Should we keep the existing data on documents (even though it is not allowed aymore & also not possible to update the data again of that type) or
2. we don't set the data of that type anymore as it is not allowed in the template (this could have adverse effect on the existing pages).

I personally want to go with the option 1, which is the case right now. wdyt?

dvesh3:
> @timwiechers The PR looks good 😉 and works like a charm.
> 
> I just came across one use case, where a type which is allowed before and then removed from the `allowedTypes` list so what happens to the existing data of that particular type?
> 
> 1. Should we keep the existing data on documents (even though it is not allowed aymore & also not possible to update the data again of that type) or
> 2. we don't set the data of that type anymore as it is not allowed in the template (this could have adverse effect on the existing pages).
> 
> I personally want to go with the option 1, which is the case right now. wdyt?

Just discussed it with the team and we decided to go with option 2, so we don't set the data anymore of the type **not** allowed in the template.

models/Document/Editable/Video.php
timwiechers: Good catch 👍 

You mean like so? 
```
public function getVideoType()
{
    if (empty($this->type) === true) {
        $this->type = $this->allowedTypes[0];
    }

    return $this->type;
}
```

dvesh3: exactly!!

timwiechers: Resolved with https://github.com/pimcore/pimcore/pull/12445/commits/11d1ab3bca09acdb87e9a587d03fafdf832c4e0d

dvesh3:
@timwiechers did you see https://github.com/pimcore/pimcore/pull/12445#issuecomment-1165528799?

timwiechers:
@dvesh3 Regarding https://github.com/pimcore/pimcore/pull/12445#issuecomment-1165528799 :
Does this mean that a change in the template (where the editable is used) should have a direct influence on the frontend display, even if the affected documents have not yet been saved in the admin area?

dvesh3:
> @dvesh3 Regarding [#12445 (comment)](https://github.com/pimcore/pimcore/pull/12445#issuecomment-1165528799) : Does this mean that a change in the template (where the editable is used) should have a direct influence on the frontend display, even if the affected documents have not yet been saved in the admin area?

@timwiechers yes, the template is the source of truth for Documents. Even direct editables and area bricks works like this, so if you change in the template then it will not be reflected on the documents anymore.

timwiechers:
@dvesh3 I have customized it so that if the configuration is invalid, the editable data is not displayed in the frontend or in the Documents section. See: https://github.com/pimcore/pimcore/pull/12445/commits/1723e42fa09493ae98d6363bd6e58d1745b8f51f

dvesh3 (approved):

dvesh3:
@timwiechers nice work 👏 thanks a lot!

PR #12449, opened 2022-06-20
Fix FFMPEG with durations like 00:00:11.04 (milliseconds)

dpfaffenbauer:
Ffmpeg adapter cannot parse durations where milliseconds are applied, using a better regex solves that problem

### DIFF
--- a/lib/Video/Adapter/Ffmpeg.php
+++ b/lib/Video/Adapter/Ffmpeg.php
@@ -253,7 +253,7 @@ class Ffmpeg extends Adapter
     }
 
     /**
-     * @return float
+     * @return float|null
      *
      * @throws \Exception
      */
--- a/lib/Video/Adapter/Ffmpeg.php
+++ b/lib/Video/Adapter/Ffmpeg.php
@@ -262,14 +262,16 @@ class Ffmpeg extends Adapter
         $output = $this->getVideoInfo();
 
         // get total video duration
-        preg_match("/Duration: ([0-9:\.]+),/", $output, $matches);
-        $durationRaw = $matches[1];
-        $durationParts = explode(':', $durationRaw);
+        $result = preg_match("/\s+Duration: ((\d\d):(\d\d):(\d\d)\.(\d+))/", $output, $matches);
 
-        // calculate duration in seconds
-        $duration = ((int)$durationParts[0] * 3600) + ((int)$durationParts[1] * 60) + (float)$durationParts[2];
+        if (false !== $result && 0 !== $result && count($matches) === 6) {
+            // calculate duration in seconds
+            $duration = ((int)$matches[2] * 3600) + ((int)$matches[3] * 60) + (float)$matches[4];
 
-        return $duration;
+            return $duration;
+        }
+
+        return null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


brusch:
@dpfaffenbauer can you please check the failing tests and rebase your PR onto `10.4`? Thanks!

dpfaffenbauer:
@brusch done

brusch:
@dpfaffenbauer the regex now always expects milliseconds, is this always returned by FFMPEG?

dpfaffenbauer:
@brusch yes, it was for me in multiple examples I tried

blankse:
After this PR, the milliseconds are missing in the return. Otherwise it's enough to return an integer.

lib/Video/Adapter/Ffmpeg.php
```diff
@@ -253,7 +253,7 @@ protected function getVideoInfo()
     }
 
     /**
-     * @return float
+     * @return float|null
```
blankse, line 256: Should also be changed in the abstract method of the parent class

lib/Video/Adapter/Ffmpeg.php
```diff
@@ -262,14 +262,16 @@ public function getDuration()
         $output = $this->getVideoInfo();
 
         // get total video duration
-        preg_match("/Duration: ([0-9:\.]+),/", $output, $matches);
-        $durationRaw = $matches[1];
-        $durationParts = explode(':', $durationRaw);
+        $result = preg_match("/\s+Duration: ((\d\d):(\d\d):(\d\d)\.(\d+))/", $output, $matches);
```
blankse, line 265: So the milliseconds should be in the return value again
```suggestion
        $result = preg_match("/\s+Duration: ((\d\d):(\d\d):(\d\d\.\d+))/", $output, $matches);
```

lib/Video/Adapter/Ffmpeg.php
```diff
@@ -262,14 +262,16 @@ public function getDuration()
...
+
+        if (false !== $result && 0 !== $result && count($matches) === 6) {
+            // calculate duration in seconds
+            $duration = ((int)$matches[2] * 3600) + ((int)$matches[3] * 60) + (float)$matches[4];
```
blankse: So the milliseconds should be in the return value again:
```suggestion
        $result = preg_match("/Duration: (\d\d):(\d\d):(\d\d\.\d+),/", $output, $matches);

        if ($result) {
            // calculate duration in seconds
            $duration = ((int)$matches[1] * 3600) + ((int)$matches[2] * 60) + (float)$matches[3];
```

dpfaffenbauer: it said already "calculate duration in seconds", thats why the milliseconds got ignored.

blankse: As you can see the 3rd value is a float. The milliseconds were there. This has worked before. The return is in seconds, the milliseconds are only after the decimal point.

dpfaffenbauer: @blankse alright, added your suggestion

lib/Video/Adapter/Ffmpeg.php
dpfaffenbauer: also done

blankse (approved):

brusch:
Thank you very much guys! 👍

PR #12452, opened 2022-06-21
[Improvement]: Refactor navigation builder to use param style arguments

### LINKED ISSUE #12310: [Improvement]: Refactor navigation builder to use param style arguments
### Improvement description

Currently `\Pimcore\Navigation\Builder::getNavigation()` is using extra arguments, this should be changed to a params array, similar to what we do in the twig helper: 

https://github.com/pimcore/pimcore/blob/2f18a0f26380c56b609d937b40d2050c8d1a3e1a/lib/Twig/Extension/Templating/Navigation.php#L72

BC layer should be possible and we should trigger a deprecation message.

aryaantony92:
Resolves #12310


lib/Navigation/Builder.php
```diff
@@ -68,21 +68,32 @@ public function __construct(RequestHelper $requestHelper, ?string $pageClass = n
...
     {
+        //TODO Pimcore 11: remove the if(count($params) > 1) block BC layer
+        if (count ($params) > 1) {
+            trigger_deprecation ('pimcore/pimcore', '10.5', 'Calling Pimcore\Navigation\Builder::getNavigation() using extra arguments is deprecated and will be removed in Pimcore 11. Instead, specify the arguments as an array
+            ');
+
+            list($activeDocument, $navigationRootDocument, $htmlMenuIdPrefix, $pageCallback, $cache, $maxDepth, $cacheLifetime) = $params;
+        } else {
+
+            $activeDocument = $params[0]['active'];
```
blankse: I would move the OptionsResolver from the Twig Extension to here. So some options can be optional.

lib/Navigation/Builder.php
```diff
@@ -68,21 +68,32 @@ public function __construct(RequestHelper $requestHelper, ?string $pageClass = n
...
      * @throws \Exception
      */
-    public function getNavigation($activeDocument = null, $navigationRootDocument = null, $htmlMenuIdPrefix = null, $pageCallback = null, $cache = true, ?int $maxDepth = null, ?int $cacheLifetime = null)
+
+    public function getNavigation (...$params)
     {
+        //TODO Pimcore 11: remove the if(count($params) > 1) block BC layer
+        if (count ($params) > 1) {
+            trigger_deprecation ('pimcore/pimcore', '10.5', 'Calling Pimcore\Navigation\Builder::getNavigation() using extra arguments is deprecated and will be removed in Pimcore 11. Instead, specify the arguments as an array
+            ');
```
jdreesen:
```suggestion
            trigger_deprecation ('pimcore/pimcore', '10.5', 'Calling Pimcore\Navigation\Builder::getNavigation() using extra arguments is deprecated and will be removed in Pimcore 11. Instead, specify the arguments as an array');
```

lib/Navigation/Builder.php
```diff
@@ -68,21 +68,32 @@ public function __construct(RequestHelper $requestHelper, ?string $pageClass = n
...
+
+            list($activeDocument, $navigationRootDocument, $htmlMenuIdPrefix, $pageCallback, $cache, $maxDepth, $cacheLifetime) = $params;
```
jdreesen: Just a note: this probably doesn't work with named params, does it?

lib/Navigation/Builder.php
dvesh3: Moved it. thanks!

lib/Navigation/Builder.php
dvesh3: Reverted the params for BC reasons.

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,6 +1,7 @@
 # Upgrade Notes
 
 ## 10.5.0
+- [Navigation Builder] Calling the method Pimcore\Navigation\Builder::getNavigation() using extra arguments is deprecated and will be removed in Pimcore 11. Instead of using the extra arguments, it is recommended to call the method using the params array. eg: Instead of calling getNavigation($activeDocument = null, $navigationRootDocument = null,..) use getNavigation($params)
```
```suggestion
- [Navigation Builder] Calling the method `Pimcore\Navigation\Builder::getNavigation()` using extra arguments is deprecated and will be removed in Pimcore 11. Instead of using the extra arguments, it is recommended to call the method using the params array. eg: Instead of calling `getNavigation($activeDocument = null, $navigationRootDocument = null,..)` use `getNavigation($params)`
```

lib/Navigation/Builder.php
```diff
@@ -66,10 +72,47 @@ public function __construct(RequestHelper $requestHelper, ?string $pageClass = n
...
+            'cache' => true,
+            'cacheLifetime' => null,
+            'maxDepth' => null,
+            'active' => null,
+        ]);
+
+        $options->setAllowedTypes('root', [Document::class, 'null']);
+        $options->setAllowedTypes('htmlMenuPrefix', ['string', 'null']);
+        $options->setAllowedTypes('pageCallback', ['callable', 'null']);
+        $options->setAllowedTypes('cache', 'bool');
```
blankse: Can also be a string

PR #12458, opened 2022-06-21
[PHP 8.1] fix Passing null to parameter #1 ($string) of type string is deprecate

dpfaffenbauer:
https://github.com/coreshop/CoreShop/runs/6981738951?check_suite_focus=true#step:17:93

### DIFF
--- a/models/Staticroute.php
+++ b/models/Staticroute.php
@@ -66,7 +66,7 @@ final class Staticroute extends AbstractModel
     /**
      * @var array
      */
-    protected $siteId;
+    protected $siteId = [];
 
     /**
      * @var array
@@ -398,12 +398,12 @@ final class Staticroute extends AbstractModel
      *
      * @return $this
      */
-    public function setSiteId($siteId = [])
+    public function setSiteId($siteId)
     {
         $result = [];
         
         if (null === $siteId) {
-            $siteId = [];
+            return;
         }
 
         if (!is_array($siteId)) {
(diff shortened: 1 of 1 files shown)


blankse:
The $siteId attribute should as default value have a empty array. So it cant be null in the future (Maybe the support can be dropped in Pimcore 11).

dvesh3:
> The $siteId attribute should as default value have a empty array. So it cant be null in the future (Maybe the support can be dropped in Pimcore 11).

@dpfaffenbauer could you please update it? thanks!

dpfaffenbauer:
@dvesh3 @blankse changed. But I don't know why it is null... Maybe cause how it is stored in Config Store or something like this... So IMO that will not really solve the problem, since it still can be null for now.

blankse:
@dpfaffenbauer I mean the attribute in line 69, not the param. So if you create a new static route and save it it can't be null and the load method set null to the setSiteId method.

dpfaffenbauer:
@blankse gotcha

models/Staticroute.php
```diff
@@ -394,13 +394,17 @@ public function getReverse()
...
+     * @param string|array|null $siteId
      *
      * @return $this
      */
     public function setSiteId($siteId)
     {
         $result = [];
+        
+        if (null === $siteId) {
+            return;
```
blankse: I think this could stay. Null should set an empty array.
```suggestion
            $siteId = [];
```

dvesh3:
@dpfaffenbauer @blankse thanks guys!!

PR #12460, opened 2022-06-21
[Improvement] Application Logger File Objects should be stored via flysystem storage

### LINKED ISSUE #11382: [Improvement] Application Logger File Objects should be stored via flysystem storage

kingjia90:
Resolves #11382

## Notes
There are some changes from #7164 : the archived file are not deleted between midnight and 4a.m. but just before the records are archived, because is  some type of storages, it does not support last modification timestamp.


BlackbitDevs:
@kingjia90 @mcop1 Is there a migration for existing `FileObject`s? Imho the `PIMCORE_LOG_FILEOBJECT_DIRECTORY` prefix needs to to stripped in `application_logs` table, e.g. 
previously: /var/application-logger/my-file-object
now: my-file-object

mcop1:
@BlackbitNeueMedien please see #12704, we added a fallback for those entries

PR #12461, opened 2022-06-21
[Thumbnails] Add width, height and filesize to the cache

### LINKED ISSUE #12055: [Bug]: Pimcore serving deferred thumbnail URL even though the thumbnail exists (Azure Storage used)

### LINKED ISSUE #12057: [Bug]: \Pimcore\Model\Asset\Thumbnail\ImageThumbnailTrait::getDimensions is reading dimensions from file every time

brusch:
follow up to #12414 from @dkarlovi (🍒-picked & squashed commits from there, keeping the author) 

resolves #12057 
resolves #12055

### DIFF
--- a/bundles/CoreBundle/Command/ThumbnailsImageCommand.php
+++ b/bundles/CoreBundle/Command/ThumbnailsImageCommand.php
@@ -182,12 +182,17 @@ class ThumbnailsImageCommand extends AbstractCommand
             $thumbnail = $image->getThumbnail($thumbnailConfig);
             $path = $thumbnail->getPath(false);
 
+            // triggers fetching the thumbnail info and updating the asset cache table
+            $dimensions = $thumbnail->getDimensions();
+
             if ($output->isVerbose()) {
                 $output->writeln(
                     sprintf(
-                        'generated thumbnail for image [%d] | file: %s',
+                        'generated thumbnail for image [%d] | file: %s | dimensions: %dx%d',
                         $image->getId(),
-                        $path
+                        $path,
+                        $dimensions['width'],
+                        $dimensions['height']
                     )
                 );
             }
--- a/bundles/CoreBundle/Command/ThumbnailsImageCommand.php
+++ b/bundles/CoreBundle/Command/ThumbnailsImageCommand.php
@@ -182,17 +182,15 @@ class ThumbnailsImageCommand extends AbstractCommand
             $thumbnail = $image->getThumbnail($thumbnailConfig);
             $path = $thumbnail->getPath(false);
 
-            // triggers fetching the thumbnail info and updating the asset cache table
-            $dimensions = $thumbnail->getDimensions();
+            // triggers fetching the thumbnail info and updating the asset cache table if width or height are not in the cache
+            $thumbnail->getDimensions();
 
             if ($output->isVerbose()) {
                 $output->writeln(
                     sprintf(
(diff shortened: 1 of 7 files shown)


bundles/CoreBundle/Command/ThumbnailsImageCommand.php
```diff
@@ -182,17 +182,12 @@ protected function runSingleCommand(string $item, InputInterface $input, OutputI
             $thumbnail = $image->getThumbnail($thumbnailConfig);
             $path = $thumbnail->getPath(false);
 
-            // triggers fetching the thumbnail info and updating the asset cache table
-            $dimensions = $thumbnail->getDimensions();
```
dkarlovi: Note: this updates the thumbnail cache when running the thumbnail generator, I'm not sure it's dead code. Updating the cache is required to fix #12055.

dkarlovi: See https://github.com/pimcore/pimcore/pull/12056#issuecomment-1120870522

brusch: Yeah, that could be useful in case you have inconsistent data in the cache (missing width, height, filesize)  ... it doesn't hurt, so yes, we can keep that.

dkarlovi: The issue with #12055 is that otherwise you can't fix it without clearing the cache table IIRC.

aashan10 (approved):

dkarlovi:
Yay! \o/ Thanks for your help with this @brusch.

BlackbitDevs:
@brusch The migration class / file name is wrong, it is for the year 2202. But we cannot simply rename it because this may have been already executed in some systems and then would lead to a doctrine notice. Perhaps keep the wrongly named migration but make it do nothing - and then create a renamed one which contains the actual logic. Otherwise, when we keep it as is, it will lead to wrong migration execution order as soon as there will be a migration which relies on the changes of this migration.

brusch:
@BlackbitNeueMedien Is there an actual problem if we just keep the wrong filename?

BlackbitDevs:
Not now but when a new migration will get created and for example be named `Version20220907000000`, this will be executed before the `Version22020614115124` if none of the migrations has been executed on the given Pimcore system before (because Doctrine executes them in ascending order). But it the `Version22020614115124` has already been executed, the new migration will be executed afterwards (because already executed migrations will not get executed again).
So the order of execution differs if you have installed Pimcore 10.5 before (which includes `Version22020614115124`) or not - and if the new migration relies on changes made in `Version22020614115124`, there will be errors.

PR #12488, opened 2022-06-22
fix: Ecommerce broken Index Field Selection

### LINKED ISSUE #12487: [Bug]: Ecommerce broken Index Field Selection
### Expected behavior

image

Expecting values for pre select

### Actual behavior

Not all values are necessarily displayed.

Bug was introduced with the following pull request
https://github.com/pimcore/pimcore/pull/9067/files

https://github.com/pimcore/pimcore/pull/9067/files#diff-d3963737de522fd60567f4e85b2021b1ea8e198acf42e65944af98fd7397c07bR277

$row  !== false is needed -> all values evaluating to null or break the loop, therefore pre select values are empty

### Steps to reproduce

Try to introduce ecommerce index field with nullable values, 0 values etc.

Bug reproducable in pimcore demo

mattamon:
 Resolves #12487
Check if row is false so nullable values do not brake the loop

### DIFF
--- a/lib/Db/PimcoreExtensionsTrait.php
+++ b/lib/Db/PimcoreExtensionsTrait.php
@@ -278,8 +278,10 @@ trait PimcoreExtensionsTrait
         $stmt = $this->executeQuery($sql, $params, $types);
         $data = [];
         if ($stmt instanceof Result) {
-            while (($row = $stmt->fetchOne()) || $row !== false) {
+            $row = $stmt->fetchOne();
+            while (false !== $row) {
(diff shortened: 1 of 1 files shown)


mysliwietzflorian:
I'd write it explicitly like:

```php
$row = $stmt->fetchOne();
while (false !== $row) {
    $data[] = $row;
    $row = $stmt->fetchOne();
}
```

When the return value of `fetchOne()` is not false they are added even when they are nullable.

aashan10 (approved):

PR #12489, opened 2022-06-22
Add getObjects to annotated class methods inside of generated listing…

Brainshaker95:
… classes

  
  

Generated Listing classes will now also have a typed method annotation for the `getObjects` function additionally to `getData` and `load`.


Brainshaker95:
Any thoughts?

brusch:
I'm wondering whether we should deprecate those aliases for `getData()` on listing classes and remove them in Pimcore 11. 
@dvesh3 @fashxp any thoughts?

Brainshaker95:
Speaking for myself I have to say that I like to use `getObjects()` more than `getData()`, it seems more explicit to me.

PR #12492, opened 2022-06-23
Fix for RequiredBy grid operator

Maksonis:
Resolves

image


questionmark78:
Nice that we ran into this at about the same time 👍 
My PR is on the 10.4 branch, just a few minutes later: https://github.com/pimcore/pimcore/pull/12493

dvesh3:
@Maksonis thanks! closed in favor of #12493 as it is on correct branch.

PR #12493, opened 2022-06-23
Fix 'number of bound variables does not match number of tokens' in Required By Operator

questionmark78:
Using the Required By Operator in the Grid Options results in an exception:

> An exception occurred while executing 'select * from dependencies where targettype = ? AND targetid = 7666 AND sourcetype = 'object'' with params ["object", 7666]: SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens

The code had both the element id in the string and in the parameters. Changed it to be a parameter and removed it from the query.


mcop1:
@questionmark78 Thank you!

PR #12499, opened 2022-06-23
[Bug]: Cannot filter on multiple CS keys on object grid 

### LINKED ISSUE #12375: [Bug]: Cannot filter on multiple CS keys on object grid
### Expected behavior

You can filter on multiple CS keys on object grid.

### Actual behavior

While trying to filter on multiple CS keys on object grid, en error occurs: 

```
*sql query* SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'value' in field list is ambiguous.
```

image

### Steps to reproduce

1. Open Object grid
2. Open grid options and select CS keys- key type is irrelevant.
3. Filter one key on a specific value.
4. FIlter another key on a specific value.

robertSt7:
resolve #12375

### DIFF
--- a/bundles/AdminBundle/Helper/GridHelperService.php
+++ b/bundles/AdminBundle/Helper/GridHelperService.php
@@ -113,7 +113,7 @@ class GridHelperService
                     if ($field instanceof Model\DataObject\ClassDefinition\Data) {
                         $mappedKey = 'cskey_' . $fieldName . '_' . $groupId . '_' . $keyid;
                         $featureJoins[] = ['fieldname' => $fieldName, 'groupId' => $groupId, 'keyId' => $keyid, 'language' => $language];
-                        if($filter['value']) {
+                        if(isset($filter['value'])) {
                             $featureCondition = $field->getFilterConditionExt(
                                 $filter['value'],
                                 $operator,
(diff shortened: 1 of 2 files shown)


mcop1:
@robertSt7  The error message regarding the sql statement is gone, but I still get the following error. I just simply activate the checkbox next to the filter option.

image

Could you please have a look?

bundles/AdminBundle/Helper/GridHelperService.php
```diff
@@ -113,14 +113,16 @@ public function getFeatureAndSlugFilters(string $filterJson, ClassDefinition $cl
...
                         $featureJoins[] = ['fieldname' => $fieldName, 'groupId' => $groupId, 'keyId' => $keyid, 'language' => $language];
-                        $featureCondition = $field->getFilterConditionExt(
-                            $filter['value'],
-                            $operator,
-                            [
-                                'name' => $mappedKey, ]
-                        );
-
-                        $featureConditions[$mappedKey] = $featureCondition;
+                        if($filter['value']) {
```
mcop1:
```suggestion
                        if(isset($filter['value'])) {
```

Otherwise the error still occurs, because we still try to access the array with an unknown key.

PR #12511, opened 2022-06-24
[Improvement]: dispatch an event after an element was loaded

jdreesen:
I'd like to propose dispatching an event after an element was loaded ([like Doctrine does](https://www.doctrine-project.org/projects/doctrine-orm/en/current/reference/events.html#postload)).

Our use case for this is recording which elements get loaded during a request, to be able to tag the response (via a header field) with this, which will later allow us to ban the affected pages from a reverse proxy cache (e.g. Varnish) when an element changes.

Manually recording which elements are used per request is tedious, and this event would greatly help us with it.

### DIFF
--- a/lib/Event/AssetEvents.php
+++ b/lib/Event/AssetEvents.php
@@ -117,6 +117,13 @@ final class AssetEvents
     const POST_DELETE_FAILURE = 'pimcore.asset.postDeleteFailure';
 
     /**
+     * @Event("Pimcore\Event\Model\AssetEvent")
+     *
+     * @var string
+     */
+    const POST_LOAD = 'pimcore.asset.postLoad';
+
+    /**
      * Arguments:
      *  - target_element | Pimcore\Model\Asset | contains the target asset used in copying process
      *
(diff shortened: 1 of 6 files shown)


brusch:
We should consider this here as well #12515 
I'm thinking of adding an optional `context` parameter for individual use and `$params` should be available as argument. WDYT?

brusch:
This would allow the following: 
```php
$image = Asset\Image::getById(360, [
    'context' => [
        'foo' => 'bar',
        'meow' => 'cat',
    ]
]);
```

jdreesen:
I don't really get it. What does it have to do with the event?

brusch:
Actually you would be then able to pass some contextual information to `getId()` which then can be read in the event listener. 

### Example: 
```php
$image = Asset\Image::getById(360, [
    'context' => [
        'foo' => 'bar',
        'meow' => 'cat',
    ]
]);
```

Listener: 
```php
\Pimcore::getEventDispatcher()->addListener(AssetEvents::POST_LOAD, function(AssetEvent $event) {
    if($event->hasArgument('params')) {
        if($event->getArgument('params')['context']['meow'] ?? null) {
            // do certain context specific stuff
        }
    }
});
```

jdreesen:
I added the `$params` as an argument to the events.

jdreesen:
PHPStan error in pipeline is unrelated.

aryaantony92:
@jdreesen Thank you!

PR #12513, opened 2022-06-27
[Improvement] add config data for area and image editable

MarcW09:
We are using a headless approach for the CMS part. For this to work we need some more config information in the backend for the area and image editable.
Area: The defined Brick type that is configured in the Twig Template. We can then serialize the brick configured in the area correctly.
Image: The defined Thumbnail Config that is configured in the Twig Template. We can then send the same Thumbnail of the image to the frontend.

Adding the configured brick type in the area editable.
Adding the configured thumbnail in the image editable.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/area.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/area.js
@@ -18,8 +18,12 @@ pimcore.document.editables.area = Class.create(pimcore.document.area_abstract, {
 
         this.id = id;
         this.name = name;
-        this.elements = [];
+        this.datax = {};
         this.config = this.parseConfig(config);
+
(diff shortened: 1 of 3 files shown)


MarcW09:
What is the due date of milestone 10.5.0? We need this PR in the next two weeks deployed, if that is possible 😄

dvesh3:
> What is the due date of milestone 10.5.0? We need this PR in the next two weeks deployed, if that is possible 😄

@MarcW09 the planned due date is July 13, 2022 for 10.5 release.

passioneight:
@MarcW09 you can always create patches for situations like these: https://github.com/cweagans/composer-patches . Just don't forget to remove the patches once the needed version was released.

dvesh3:
@MarcW09 thanks!

PR #12515, opened 2022-06-27
[Elements] change $force parameter to a generic $params array

brusch:
### OLD
```php
Asset::getById(123,true); 
```

### NEW
```php
Asset::getById(123, ['force' => true]); 
```

This allows more options in the future and more flexibility.


doc/Development_Documentation/26_Best_Practice/95_Working_With_Runtime_Cache.md
```diff
@@ -20,5 +20,5 @@ foreach ($list as $element) {
 //using force param to load latest data from database
 \Pimcore\Model\DataObject::getById(123) === \Pimcore\Model\DataObject::getById(123) => true
 
-\Pimcore\Model\DataObject::getById(123) === \Pimcore\Model\DataObject::getById(123, true) => true/false
+\Pimcore\Model\DataObject::getById(123) === \Pimcore\Model\DataObject::getById(123, ['force' =>> true]) => false
```
dvesh3:
```suggestion
\Pimcore\Model\DataObject::getById(123) === \Pimcore\Model\DataObject::getById(123, ['force' => true]) => false
```

models/Element/AbstractElement.php
```diff
@@ -867,4 +868,26 @@ protected function renewInheritedProperties()
...
+     * @internal
+     * @param bool|array $params
+     * @return array
+     */
+    protected static function prepareGetByIdParams(/*array */$params, string $method): array
+    {
+        if($params === true) {
+            trigger_deprecation('pimcore/pimcore', '10.5', 'Using $force=true on %s is deprecated, please use array-syntax [force=>true] instead.', $method);
+            $params = ['force' => true];
+        }
```
jdreesen: Maybe we should check for `bool` here? Passing `false` is deprecated as well, I suppose.

```suggestion
        if(is_bool($params)) {
            trigger_deprecation('pimcore/pimcore', '10.5', 'Using $force=%s on %s is deprecated, please use array-syntax [force=>true] instead.', $force ? 'true' : 'false', $method);
            $params = ['force' => $force];
        }
```

brusch: Hmm, the problem is that `false` is the default value, so when calling `getById(123)` without the second argument a deprecation message gets issued.

aryaantony92:
@brusch Could you please resolve the conflicts ? Thanks.

brusch:
@aryaantony92 done

aryaantony92:
Thanks 👍

PR #12536, opened 2022-06-27
[BackOffice] change PaginateListingSubscriber

Johanna-Maria:
I can't see more than 10 entries in the order backoffice. When I swap these two lines, it works. In 6.9 these two lines are swapped too. I've seen this was changed on branch 10.x recently with commit 4013e6c.

### DIFF
--- a/lib/Model/Paginator/EventSubscriber/PaginateListingSubscriber.php
+++ b/lib/Model/Paginator/EventSubscriber/PaginateListingSubscriber.php
@@ -26,8 +26,8 @@ class PaginateListingSubscriber implements EventSubscriberInterface
         $paginationAdapter = $event->target;
 
         if ($paginationAdapter instanceof PaginateListingInterface) {
-            $items = $paginationAdapter->getItems($event->getOffset(), $event->getLimit());
             $event->count = $paginationAdapter->count();
+            $items = $paginationAdapter->getItems($event->getOffset(), $event->getLimit());
             $event->items = $items;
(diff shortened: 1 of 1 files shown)


dvesh3:
@Johanna-Maria Thanks for your PR! this will introduce this problem again https://github.com/pimcore/pimcore/issues/11322 ..right?

I'd rather go for proper solution which also fixes the Index service product list load. please see https://github.com/pimcore/pimcore/pull/11324#issuecomment-1026751469

Johanna-Maria:
@dvesh3 Hello,

I had a look at the problem from #11322 and I can't reproduce the bug.

What I did:

1. I set up a local instance of the pimcore demo and swapped the two lines in the PaginateListingSubscriber, so that count is called before getItems.
2. I changed "Results per Page" of object 699 (the volkswagen filter definition) to something other than 10 (I tried 3 and 22)
3. I visited http://localhost/en/More-Stuff/Landingpages/Volkswagen?pimcore_preview=true&_dc=1657175953138 where the count of the items displayed always equaled the amount I had configured in the pimcore object.

So, my question is: did I do something wrong while reproducing the bug or am I right?

Our Backoffice (the reason I made this pull request) works fine when the count comes before the getItems, but otherwise it is rather useless as one can't see more than ten items (total item count shows 10 and the pagination to go to the next side is not visible)

dvesh3:
Closed in favor of #12626. thanks!

PR #12537, opened 2022-06-27
[Improvement]: automatically run migrations in correct order

jdreesen:
Doctrine Migrations compares migration versions [by their FQCN by default](https://github.com/doctrine/migrations/blob/3.5.x/lib/Doctrine/Migrations/Version/AlphabeticalComparator.php#L13), which means that they will most likely be run in the wrong order if you simply run `bin/console doctrine:migrations:migrate`, since `App\...` migrations will be executed before `Pimcore\Bundle\CoreBundle\...`. You can prevent this by running core migrations first, via `--prefix=Pimcore\\Bundle\\CoreBundle`. But if you've many bundles installed that have migrations themselves, it becomes tedious to manually maintain the order in which migrations should run.

But since Pimcore has a bundle registration priority and a bundle dependency system, we're able to automatically determine the correct order of migrations. Bundles with a higher priority or those that other bundles depend on are registered earlier, which means that their `config/pimcore/config.yaml` will be loaded first, resulting in the correct `doctrine_migrations: { migrations_paths: { ... }}` sequence. Now if we sort by the migration version class namespace first and then the class name, we should have the correct order.

### DIFF
--- a/bundles/CoreBundle/Resources/config/migrations.yaml
+++ b/bundles/CoreBundle/Resources/config/migrations.yaml
@@ -13,3 +13,6 @@ services:
     Doctrine\Migrations\DependencyFactory:
         alias: 'doctrine.migrations.dependency_factory'
 
+    Pimcore\Migrations\DirectoryAwareVersionComparator:
+        arguments:
+            $config: '@doctrine.migrations.configuration'
--- a/bundles/CoreBundle/Resources/config/pimcore/doctrine_migrations.yaml
+++ b/bundles/CoreBundle/Resources/config/pimcore/doctrine_migrations.yaml
@@ -1,7 +1,7 @@
 doctrine_migrations:
     connection: default
     migrations_paths:
-        'Pimcore\Bundle\CoreBundle\Migrations': '@PimcoreCoreBundle/Migrations'
+        Pimcore\Bundle\CoreBundle\Migrations: '@PimcoreCoreBundle/Migrations'
     storage:
         table_storage:
             table_name: 'migration_versions'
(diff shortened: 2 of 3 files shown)


dvesh3:
@jdreesen This should also work for bundles registered through `config/bundles.php`..right? where bundles are registered as per array list. Asking in regards to https://github.com/pimcore/pimcore/pull/12616 where `extensions.php` is deprecated and will be removed in Pimcore 11, so there will be no priority for bundles anymore.

jdreesen:
Yes, it should still work (because `\Pimcore\HttpKernel\BundleCollection\BundleCollection` is going to stay, right?). 

Bundles from `config/bundles.php` are all registered with priority `0` (because Symfony doesn't know a priority concept for bundles), which is probably fine most of the time, but there's still the possibility to register your bundles through `AppKernel::registerBundlesToCollection()` when you need a different priority or even `CustomBundle::registerDependentBundles()` (which comes from `\Pimcore\HttpKernel\Bundle\DependentBundleInterface`).

dvesh3:
Yes, `BundleCollection` and `PimcoreBundleManager` will stay as we need installer and load public files for bundles. Only the `Extensions/Config` will go as it registers the bundles through `extensions.php`, which will be replaced by `bundles.php` making Pimcore more compatible with Symfony flex. 

The conclusion is then that we don't need to adapt this PR after merging #12616. thanks!

dvesh3 (approved):

dvesh3:
@jdreesen nice work! 👏 thanks!

PR #12539, opened 2022-06-27
[Translations] Filter by languages

### LINKED ISSUE #12490: [Improvement]: GraphQL - Filter translations by language
### Improvement description

 Filter translations by language with language code (en, de ....)

dvesh3:
Resolves #12490

## Additional info  
```php
  $translations = new Translation\Listing();
  $translations->setLanguages(['en', 'de']);
  
  foreach ($translations as $trans) {
      // fetch en or de value with getTranslation() on translation object
  }
```

### DIFF
--- a/models/Translation.php
+++ b/models/Translation.php
@@ -263,7 +263,7 @@ final class Translation extends AbstractModel
      *
      * @throws \Exception
      */
-    public static function getByKey(string $id, $domain = self::DOMAIN_DEFAULT, $create = false, $returnIdIfEmpty = false)
+    public static function getByKey(string $id, $domain = self::DOMAIN_DEFAULT, $create = false, $returnIdIfEmpty = false, $languages = null)
     {
         $cacheKey = 'translation_' . $id . '_' . $domain;
         if (Runtime::isRegistered($cacheKey)) {
@@ -273,10 +273,10 @@ final class Translation extends AbstractModel
         $translation = new static();
         $translation->setDomain($domain);
         $idOriginal = $id;
-        $languages = static::getValidLanguages($domain);
+        $languages = $languages ? array_intersect(static::getValidLanguages($domain), $languages) : static::getValidLanguages($domain);
 
         try {
-            $translation->getDao()->getByKey($id);
+            $translation->getDao()->getByKey($id, $languages);
         } catch (\Exception $e) {
             if (!$create && !$returnIdIfEmpty) {
                 return null;
--- a/models/Translation.php
+++ b/models/Translation.php
@@ -258,6 +258,7 @@ final class Translation extends AbstractModel
      * @param string $domain
      * @param bool $create
      * @param bool $returnIdIfEmpty
+     * @param array|null $languages
      *
      * @return static|null
      *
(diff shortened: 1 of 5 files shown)


models/Translation/Dao.php
```diff
@@ -40,12 +40,20 @@ public function getDatabaseTableName(): string
...
      *
      * @throws NotFoundResourceException
      */
-    public function getByKey($key)
+    public function getByKey($key, $languages = null)
     {
-        $data = $this->db->fetchAll('SELECT * FROM ' . $this->getDatabaseTableName() . ' WHERE `key` = ? ORDER BY `creationDate` ', [$key]);
+        if (is_array($languages)) {
+            $sql = 'SELECT * FROM ' . $this->getDatabaseTableName() . ' WHERE `key` = ?
+            AND `language` IN (' . implode(',' , array_map([$this->db, 'quote'], $languages)) . ') ORDER BY `creationDate` ';
```
jdreesen: Can we please stop inlining parameters and use prepared statements instead?

dvesh3: that was my first try but it is not working with IN operator and this is a general improvement topic as we do this at lot of places.

jdreesen: Are you sure it doesn't work? I'm pretty sure I already used it. I think it has to be an array and needs a type, like: `fetchAll('...', ['key', ['de','en']], [1 => \Doctrine\DBAL\Connection::PARAM_STR_ARRAY])` (it would probably better to use named parameters instead of positional ones in this case). See: https://www.doctrine-project.org/projects/doctrine-dbal/en/2.13/reference/data-retrieval-and-manipulation.html#list-of-parameters-conversion

dvesh3: yes, you are right. I missed passing the type in first attempt. I'll give it a try. thanks!

dvesh3: @jdreesen works like a charm. thanks!

tests/Unit/Translation/TranslatorTest.php
```diff
@@ -219,5 +219,14 @@ public function testLoadingTranslationList()
...
         $this->assertCount(1, $translations);
+
+        //test: Filter by languages
+        $translations = new Translation\Listing();
+        $translations->setDomain('messages');
+        $translations->setLanguages(['en', 'de']);
+
+        $translations = $translations->getTranslations();
+        $translationValues = $translations[0]->getTranslations();
+        $this->assertArrayNotHasKey('fr', array_keys($translationValues));
```
robertSt7:
```suggestion
        $this->assertArrayNotHasKey('fr', $translationValues);
```

models/Translation.php
```diff
@@ -263,7 +263,7 @@ public static function clearDependentCache()
      *
      * @throws \Exception
      */
-    public static function getByKey(string $id, $domain = self::DOMAIN_DEFAULT, $create = false, $returnIdIfEmpty = false)
+    public static function getByKey(string $id, $domain = self::DOMAIN_DEFAULT, $create = false, $returnIdIfEmpty = false, $languages = null)
     {
         $cacheKey = 'translation_' . $id . '_' . $domain;
```
robertSt7, line 269:
```suggestion
        $cacheKey = 'translation_' . $id . '_' . $domain;
        if(is_array($languages)) {
            $cacheKey = 'translation_' . $id . '_' . $domain . '_' . implode("-", $languages);
        }
```

dvesh3: Changed here https://github.com/pimcore/pimcore/pull/12539/commits/6c0190a1bf45c7e8c0d850a95d24a24e65ef2091

PR #12483, opened 2022-06-22
Glossary: allow html entities like &nbsp; or &copy;

dpfaffenbauer:
Solves error 

`An exception has been thrown during the rendering of a template ("Warning: DOMDocumentFragment::appendXML(): Entity: line 1: parser error : Entity 'nbsp' not defined").`

PR #12541, opened 2022-06-28
Glossary fix for &lt; &gt; and &amp;

dpfaffenbauer:
Fix before didn't work with &amp;. this now does.


dvesh3:
follow up to https://github.com/pimcore/pimcore/pull/12483

dvesh3:
@dpfaffenbauer thanks!

PR #12542, opened 2022-06-28
Fix: toggle tag highlighting check if element is not null

dvelguru:
### Problem
In admin, while working with a custom area brick setup, I receive a javascript error when pressing the toggle tag highlighter to highlight all editable elements in a Document.

Bildschirmfoto 2022-06-28 um 11 18 22

### Fix
This fix checks whether the element exists, and the `hasCls` method can be called or not. If this fix is applied, then the editable highlighting works perfectly fine.


aryaantony92:
@dvelguru Thank you!

PR #12546, opened 2022-06-28
Change the function "getClosestParentOfClass" to from "private" to "p…

MarkTro:
I found this method always very useful if you need a specific parent type. Therefore I would like to have it "public" again. Otherwise I need to implement it myself, which would be mostly just copy and paste.

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -531,7 +531,7 @@ class Concrete extends DataObject implements LazyLoadedFieldsInterface
      *
      * @return self|null
      */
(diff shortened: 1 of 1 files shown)


brusch:
Hmm, not sure about this, as this could change in future ... let's keep that open for discussion 😊

NiklasBr:
+1 to this change. It wasn't clear why it originally happened, so we simply copied the `getNextParentForInheritance()` and `getClosestParentOfClass()` into a Trait which we use in most classes now because of the `@internal`/`private` changes.

Issue #11082, opened 2021-12-22
[Bug]: Timeout when using shared translations "import & merge"


rjankiew:
### Expected behavior

Large files are processed without error.

### Actual behavior

When using  "import & merge" with csv files that have a lot of records, en error occurs:
![obraz](https://user-images.githubusercontent.com/89988908/147116608-cbcb3cf7-671f-41d5-9450-6c6904b98230.png)

Details:
```
Status: 504 |

URL: /admin/translation/import?merge=1

Method: POST

Message:

 

504 Gateway Time-out
```

The import is continued in the background (on a cron container as I understand), and the file is processed to the end. If you reload the grid the number of keys goes up until all elements are processed.

### Steps to reproduce

Create a csv file with large amount of new translation entries (for my enviroinment 3,5k record is sufficient to trigger this behaviour). This should be new, not updated keys. Use "import & merge" on shared translations panel to import that file.

kingjia90:
I couldn't reproduce the same problem so far. 5000+ rows, 4 languages and still worked fine.

Could it be that the refactoring in this PR fixed something?
https://github.com/pimcore/pimcore/pull/10979
Specifically the commented `$this->getAllTranslations()`, before that it could also depending on how much existing data is there and filling up the memory.

kingjia90:
After trying same process with same import files ( truncate table, import 1 smaller csv (130 rows), then import the big csv 5000+ rows) and only when doing in Admin translation, it crashes. In Admin translation there are around 20 languages.
So the first guess is that the quantity of languages is a determining factor and that can cause a memory exhaustion crash.

![image](https://user-images.githubusercontent.com/6014195/175352786-5a6f3d04-76c8-4c06-a073-b4096a1d3f4f.png)
https://github.com/pimcore/pimcore/blob/2cf0a32eeea1bd0fe1d0272fa73ebe3aa2dd54ec/lib/Localization/LocaleService.php#L53-L59

kingjia90:
https://github.com/pimcore/pimcore/blob/62258f0b3e3dfce781872b694c0f1544fc8c1bed/models/Translation.php#L479
getByKey creates a new instance of the static class and uses around 0,05 mbytes (by putting `memory_get_usage();` before and after) and for 3500 records is around 175 mbytes. 
So it depending on how long the translation text,  it can reach quickly the limit of 268mbytes and throwing fatal error.

PR #12547, opened 2022-06-28
[Translations] Fix memory issue while importing large data

dvesh3:
Resolves partially #11082

## Additional info  
It only fixes the issue https://github.com/pimcore/pimcore/issues/11082#issuecomment-1165313306 not timeout

### DIFF
--- a/models/Translation.php
+++ b/models/Translation.php
@@ -515,7 +515,7 @@ final class Translation extends AbstractModel
                     }
 
                     // call the garbage collector if memory consumption is > 100MB
-                    if (memory_get_usage() > 100000000) {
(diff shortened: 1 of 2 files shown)


models/Translation.php
```diff
@@ -513,6 +513,11 @@ public static function importTranslationsFromFile(string $file, $domain = self::
                             $t->save();
                         }
                     }
+
+                    // call the garbage collector if memory consumption is > 100MB
+                    if (memory_get_usage() > 100000000) {
```
jdreesen: What about writing it like this for better readability?

```suggestion
                    if (memory_get_usage() > 100_000_000) {
```

PR #12548, opened 2022-06-28
[Improvement]: Rename Pimcore\Cache\Runtime to Pimcore\Cache\RuntimeCache

### LINKED ISSUE #12474: [Improvement]: Rename Pimcore\Cache\Runtime to Pimcore\Cache\RuntimeCache
### Improvement description

BC layer + deprecation messages. 

See following code how to make the BC layer properly: 
https://github.com/pimcore/pimcore/blob/6.9/models/DataObject/ClassDefinition/Data/Multihref.php
https://github.com/pimcore/pimcore/blob/6.9/models/DataObject/ClassDefinition/Data/ManyToManyRelation.php#L1117

aashan10:
Resolves #12474


lib/Cache/Runtime.php
```diff
@@ -15,65 +15,21 @@
 
 namespace Pimcore\Cache;
 
-final class Runtime extends \ArrayObject
+final class Runtime extends RuntimeCache
```
dvesh3, line 18:
```suggestion
/**
 * @deprecated 
 */
final class Runtime extends RuntimeCache
```

PR #12555, opened 2022-06-29
Disabled the option to drag and drop objects if the target is a paginated tree

### LINKED ISSUE #11097: [Bug]: Clearing siblings cache when moving object in tree causes timeout
### Expected behavior

When moving an object to a folder already containing a lot of objects, moved object should appear in the new folder without errors.

### Actual behavior

An error occurs:
obraz
if in `updateIndexesOfObjectSiblings` method the foreach clearing siblings cache operates for too long. 

### Steps to reproduce

Create a folder containing a large number (>35k) of objects. Create an object in a different part of the object tree. Move newly created object to the folder created before.

aryaantony92:
Resolves #11097 

Disabled the option to drag and drop objects if the target is a paginated tree

### DIFF
--- a/bundles/CoreBundle/Resources/translations/en.json
+++ b/bundles/CoreBundle/Resources/translations/en.json
@@ -167,6 +167,7 @@
     "locked": "Locked",
     "element_cannot_be_move_because_it_is_locked": "Element cannot be moved because it is locked.",
     "element_cannot_be_moved": "The element cannot be moved here",
+    "element_cannot_be_moved_because_target_is_paginated": "The element cannot be moved here because the target is a paginated tree/folder",
     "no_collections_allowed": "No Collections allowed",
     "filter_condition": "Filter Condition",
     "all_types": "All Types",
--- a/bundles/CoreBundle/Resources/translations/en.json
+++ b/bundles/CoreBundle/Resources/translations/en.json
@@ -167,7 +167,7 @@
     "locked": "Locked",
     "element_cannot_be_move_because_it_is_locked": "Element cannot be moved because it is locked.",
     "element_cannot_be_moved": "The element cannot be moved here",
-    "element_cannot_be_moved_because_target_is_paginated": "The element cannot be moved here because the target is a paginated tree/folder",
+    "element_cannot_be_moved_because_target_is_paginated": "The element cannot be moved here because the target is a paginated tree/folder. Please use \"Search & Move\" option under the \"Advanced\" menu for moving this.",
     "no_collections_allowed": "No Collections allowed",
     "filter_condition": "Filter Condition",
     "all_types": "All Types",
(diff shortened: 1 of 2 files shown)


bundles/CoreBundle/Resources/translations/en.json
```diff
@@ -167,6 +167,7 @@
     "locked": "Locked",
     "element_cannot_be_move_because_it_is_locked": "Element cannot be moved because it is locked.",
     "element_cannot_be_moved": "The element cannot be moved here",
+    "element_cannot_be_moved_because_target_is_paginated": "The element cannot be moved here because the target is a paginated tree/folder",
```
dvesh3: we should also suggest to use "Search & Move" option under Advance Menu for still required cases.

bundles/CoreBundle/Resources/translations/en.json
```diff
@@ -167,7 +167,7 @@
     "locked": "Locked",
     "element_cannot_be_move_because_it_is_locked": "Element cannot be moved because it is locked.",
     "element_cannot_be_moved": "The element cannot be moved here",
-    "element_cannot_be_moved_because_target_is_paginated": "The element cannot be moved here because the target is a paginated tree/folder",
+    "element_cannot_be_moved_because_target_is_paginated": "The element cannot be moved here because the target is a paginated tree/folder. Please use \"Search & Move\" option under the \"Advanced\" menu for moving this.",
```
dvesh3:
```suggestion
    "element_cannot_be_moved_because_target_is_paginated": "The element cannot be dragged here because the target is a paginated tree/folder. Please use \"Search & Move\" option under the \"Advanced\" context menu of the target to move the elements.",
```

brusch:
@aryaantony92 @dvesh3 is this also relevant for assets and documents? 
And wouldn't a reload of the tree (or parent node) solve the problem as well?

dvesh3:
@brusch This is about performance when dealing with large number of data as moving element on the tree re-indexes the all elements (manual index) under the same parent, so imagine reindexing 10-20k objects at the same time. please see more details here https://github.com/pimcore/pimcore/issues/11097#issuecomment-1162827694

Assets does not support manual indexes. For Documents, it is still relevant for navigation building.

brusch:
@dvesh3 ah ok, thanks for sharing the details. But then we could add a further check whether manual indexing is enabled on the node or not, right? Then dnd would still work for all nodes where manual index is not used.

dvesh3:
> @dvesh3 ah ok, thanks for sharing the details. But then we could add a further check whether manual indexing is enabled on the node or not, right? Then dnd would still work for all nodes where manual index is not used.

@brusch yes, we could improve it to check for manual indexing but first we need to decide if drag/drop or move really needed on a paginated list as you won't be able to move elements between pages. Only the elements outside the parent folder can be dropped on a page. If this is still relevant, I'd go for clear message that dnd not possible with manual indexes.

PR #12562, opened 2022-06-30
addCondition default $value value should be array

MercuryKojo:
if you don't give a value for addCondition the function calls setParameters with [null] which gives an "SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens" Error
So the default value should be [], since setParameters requires array anyway

### DIFF
--- a/bundles/EcommerceFrameworkBundle/OrderManager/Order/Listing.php
+++ b/bundles/EcommerceFrameworkBundle/OrderManager/Order/Listing.php
@@ -319,7 +319,7 @@ class Listing extends AbstractOrderList implements OrderListInterface
      *
      * @return $this
      */
-    public function addCondition($condition, $value = null)
+    public function addCondition($condition, $value = [])
     {
         if (!is_array($value)) {
             $value = [$value];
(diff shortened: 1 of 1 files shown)


bundles/EcommerceFrameworkBundle/OrderManager/Order/Listing.php
```diff
@@ -319,7 +319,7 @@ public function addFilter(OrderListFilterInterface $filter)
      *
      * @return $this
      */
-    public function addCondition($condition, $value = null)
+    public function addCondition($condition, $value = [])
     {
         if (!is_array($value)) {
```
brusch:
```suggestion
    public function addCondition($condition, $value = null)
    {
        if (null === $value) {
            $value = [];
        }
                
        if (!is_array($value)) {
```

brusch (changes requested):
Actually your change is a BC break, because the method signature changes and could break existing code. But my suggestion should fix the issue as well without breaking anything.

MercuryKojo:
Yes makes sense.
Already accepted the change

aryaantony92:
@MercuryKojo Thank you for the fix.

PR #12565, opened 2022-06-30
Improve navigation builder a bit (followup of #12452)

jdreesen:
I just wanted to suggest some small improvements for the navigation builder as a followup of #12452.

I've done it in two commits in case you don't like the options unpacking - I find it a lot cleaner.


aryaantony92:
@Thank you!

PR #12566, opened 2022-07-01
Change JS core events for bundles to native JS events

### LINKED ISSUE #7290: Change JS core events for bundles to native JS events

robertSt7:
resolve #7290


bundles/AdminBundle/Resources/public/js/pimcore/event-dispatcher.js
```diff
@@ -1,56 +0,0 @@
...
- * - Pimcore Commercial License (PCL)
- * Full copyright and license information is available in
- * LICENSE.md which is distributed with this source code.
- *
- * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
- * @license    http://www.pimcore.org/license     GPLv3 and PCL
- */
-
-pimcore.registerNS("pimcore.eventDispatcher");
-pimcore.eventDispatcher = {
```
dpfaffenbauer, line 15: removing this is a BC break

dpfaffenbauer: https://github.com/coreshop/CoreShop/pull/2046

brusch: Thanks for the hint, we didn't expect that someone is using that, because actually it was just intended to be used internally and was not documented.  
We probably need to make that clearer in 11 by using annotations in JS as well.

dpfaffenbauer: @brusch Berni, CoreShop is using everything Pimcore throws at it ;). You should know that by now :P

PR #12567, opened 2022-07-01
[Data Objects]: Deprecate "generate type declarations" in class definitions

### LINKED ISSUE #12467: [Data Objects]:  Deprecate "generate type declarations" in class definitions
### Improvement description

As type declarations will be added in Pimcore 11, it's time to deprecate this BC option in class definitions. 

image

We should make this deprecation also visible in the UI.

robertSt7:
resolve #12467


mcop1:
@robertSt7 Looks good, but I think we should add an entry in the upgrade notes. Something like "Option is deprecated, because type declarations will be added with Pimcore 11"

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -15,6 +15,7 @@
   Listing, grid, tree view are not severely affected as the main permission is checked on a Kernel event level that prevents the page to be shown and prevents any process that iterate isAllowed() calls. 
   The only cases could be affected are those where the workspace are set but master permissions are disallowed, before this change, it could lead to (not intended) false positive.
 - [Security/User] `UsernameNotFoundException` (deprecated since Symfony 5.3) occurences have been replaced with `UserNotFoundException`.
+- [Deprecated] Generate type declarations option in class definition is deprecated, because type declarations will be added with Pimcore 11
```
```suggestion
- [Deprecated] Generate type declarations option in class definition is deprecated, because type declarations will always be added with Pimcore 11
```

PR #12570, opened 2022-07-01
New video thumbnail filter - cut, mute, fps, colorchanelmixer

spotbot2k:
Adds a simple filter so that the video thumbnails can be cut to a certain duration. Can be useful to create short previews.

Works as follows:
 - Set the `start` param in seconds for the starting position
 - Set `duration` in seconds or set it to `0` for full length (start position -> end of the video)


spotbot2k:
Shall this one be of any interest then few more can be added - **"mute"** and **"fps"**. In combination it will allow the creation of smaller thumbnails sawing storage, bandwitch and computing power.

brusch:
@spotbot2k yes, i think adding more options as suggested would be a great addition 👍

spotbot2k:
@brusch - I've added them, as separating this in two or three PR's does not make mutch sense. I had an ideo of adding an intro to a video as a filter, basicly concatinate them. But it is a bit more complex then just adding a few arguments and nees a bit more of brainstorming to implement.

fashxp:
@spotbot2k can you please remove the translations other than english, as they are translated via poeditor by the community and then merged into the pimcore repository.
thx very much...

bundles/AdminBundle/Resources/public/js/pimcore/settings/videothumbnail/item.js
```diff
@@ -458,6 +458,118 @@ pimcore.settings.videothumbnail.items = {
...
+            id: myId,
+            style: "margin-top: 10px",
+            border: true,
+            bodyStyle: "padding: 10px;",
+            tbar: this.getTopBar(niceName, myId, panel),
+            items: [{
+                xtype: 'numberfield',
+                name: "fps",
+                fieldLabel: t("fps"),
+                minValue: 0,
```
kingjia90:
```suggestion
                minValue: 1,
```

lib/Video/Adapter/Ffmpeg.php
```diff
@@ -418,4 +426,29 @@ public function scaleByHeight($height)
         $height = ceil($height / 2) * 2;
         $this->addArgument('-filter:v', 'scale=trunc(oh/(ih/iw)/2)*2:'.$height);
     }
+
+    /**
+     * @param integer $start
+     * @param integer $duration
+     */
+    public function cut(int $start, ?int $duration): void
```
kingjia90: both `$start` and `$duration` seem nullable:

-may need to change `$start` param type to `?int`
-when `$start` is empty or null, should be passing `0`
-phpdocblock should be rendundant then, if not, they are missing the `|null`

lib/Video/Adapter/Ffmpeg.php
```diff
@@ -333,11 +333,19 @@ protected function getConversionLogFile()
      * @param string $key
      * @param string $value
      */
-    public function addArgument($key, $value)
+    public function addArgument(string $key, string $value): void
```
kingjia90, line 358: this stricter parameter type could be problematic, we planned to do such changes in 11, can you remove these changes for the moment?

kingjia90 (changes requested):
Thank you for the PR!

I've only noticed these small things, the rest looks good and works fine so far

lib/Video/Adapter/Ffmpeg.php
kingjia90: oh nevermind on this, it's an internal class

lib/Video/Adapter/Ffmpeg.php
spotbot2k: Thanks! The thing is - in thory `$start` is nullable, so you can cut the video to a certain length from the beginning. But then, you can not set the second parameter without setting the first one, so the `$duration` must be null as well... and it does not make any sense. But yes, in php 8 we have named arguments, so it is posible - fixed.

kingjia90 (approved):

kingjia90:
Thank you for the changes!

One possible further improvement that i had in mind is to change cut start field to be a more human-readable time field with format `hh:mm:ss`, since currently it is considering the seconds, anything above around X minutes it would take a calculator to convert hours and minutes to seconds by the user, with the timefield it also would avoid dealing with `int` and typecasting to string, but not sure if it would take too much effort.

Notes: some documentation about `-ss` https://trac.ffmpeg.org/wiki/Seeking valid values can be hh:mm:ss or even '3000ms'

spotbot2k:
@kingjia90 indeed. It needs to be changed to a string. Some kind of visual preview in backend might be useful.

kingjia90:
> @kingjia90 indeed. It needs to be changed to a string. Some kind of visual preview in backend might be useful.

@spotbot2k Would you be able to provide these implementations in this PR? 
If not, i would merge this first and open separate followup issues for that. Let me know

spotbot2k:
> Would you be able to provide these implementations in this PR?

I have changed the argument type to string, added regex to the extjs fields. The thing with visual preview will not work as there is no particular video assets to do so. I stil want to do two things - add duration check, as setting the start beyond will produce an empty file and add tooltips to extjs, as the format (H:M:S.MS) is not so self explainatory.

kingjia90:
Thank you, much better, just wanted to point out about the limiting the range `0-59` for seconds and minutes or it may affect(?) FFmpeg.

For video thumbnail preview, maybe we can add a dropdown to decide which thumbnail config to load when viewing the asset, that would be very nice. In other words, the equivalent of this download button for image format but playing the preview instead of download
image


Looking forward for the other 2 additions.

spotbot2k:
@kingjia90 One step at a time please) I had to change the way the video filter (including scale) is added as otherwise scale and fps would cancel each other. The seeking is revamped as well - now using "input seeking" and setting the `-ss` argument before `-i` to make it way faster. In Extjs the regex was changed to limit minutes and seconds to 0-59 and a placeholder was added.

lib/Video/Adapter/Ffmpeg.php
```diff
@@ -416,6 +446,34 @@ public function scaleByHeight($height)
...
+
+    /**
+     * @param string|null $inputSeeking
+     * @param string|null $targetDuration
+     */
+    public function cut(?string $inputSeeking = null, ?string $targetDuration = null): void
+    {
+        if (!empty($inputSeeking)) {
+            $parts = explode(':', $inputSeeking);
+            $this->inputSeeking = (float) ((int) $parts[0] * 3600 + (int) $parts[1] * 60 + (float) $parts[2]);
```
kingjia90: IMHO this explosion and float are not necessary, i would just declare `$inputSeeking` as string and use as is. 
Since `-ss` accepts string values as `00:01:59` as [valid](http://ffmpeg.org/ffmpeg-utils.html#time-duration-syntax) , that command line ultimately will be working with strings and as long as is properly validated in frontend, it would be simpler and more direct. 

wdyt?

spotbot2k: @kingjia90 the reason behind it is that this way we are able to do some calculations with the time, for example in this implementation it will reset the start position to 0 if it is beyond the duration of the video. In the future we can reset it not to 0 but recalculate the position as `start = sourceDuration - targetDuration`. All of above is easier to do if all the values are numerical.

kingjia90: ah i see, that could be a valid point. 
Can we do the similar "validation" and conversion as here https://github.com/pimcore/pimcore/pull/12449/files, so the values would be "compatible" with each other.

spotbot2k: Ok, I've changed the conversion, the "validation" might follow. The one for duration is limited to 99 hours, as it only checks two digits per group. For the same reason it needs seconds with leading zero. Not that it is a big issue. But hey, now there is a color chanel mixer like described in [https://hhsprings.bitbucket.io/docs/programming/examples/ffmpeg/manipulating_video_colors/colorchannelmixer.html](https://hhsprings.bitbucket.io/docs/programming/examples/ffmpeg/manipulating_video_colors/colorchannelmixer.html). With presets to choose from or, if you know what you are doing, custom string. At this point I need to update the docs.

kingjia90: Thank you, but for validation i meant the `preg_match` instead of the `explode` and eventually aborting the video thumbnail generation process when returning `false`.

About the color channel mixer, that is cool and can see it will come in handy, but it needs to be somehow extendable so devs can add more presets or manually define own custom values 🤔, and please rename the PR title accordigly. Looking forward for the docs.

spotbot2k: @kingjia90 - ok, preg_match it will be. The "effect" setting for the colormixer is in fact a combobox, so you can just write a custom value into it OR select a preset. The preview feature needs to be added for this one to be more usefull (like the one you mentioned with downloadable thumbnails). Lots to do indeed.

spotbot2k:
So, the next update.

1. The preview of the video assets can now switch between the thumbnails in the player. There is no "this one is downloadable, this one is not". Because they are all reachable via frontend, there are less of them as it is the case for images and a download button would be confusing, as the time needed to create the video the first time might be longer than the user would expect. The player on the other hand gives a visual hint of the current status of the task.
2. The "Use current player position as preview" is disabled for any custom thumbnail and only enabled for the original video.
Bildschirmfoto vom 2022-07-30 22-38-15
3. The "validation" is changed to regex, though slightly changed - unlike parsing the ffmpeg system input the milliseconds are not mandatory here, but will be accepted into the calculation if set.
4. If regex fails it will not cancel the task all together, it will just not apply the filter
5. Docs are updated

Please reopen the PR, as it wass closed after I renamed my branch. Or I can rename it and create a new one.

spotbot2k:
I think its enough for the start. Oviously there is a lot of effect that can be added and a fearute I personaly would like - adding an intro to a thumbnail, but they can be added in a separate PR. Anny suggestions for fixes/improvements?

kingjia90:
> I think its enough for the start. Oviously there is a lot of effect that can be added and a fearute I personaly would like - adding an intro to a thumbnail, but they can be added in a separate PR. Anny suggestions for fixes/improvements?

Thank you, that's quite a good amount of improvment and new features for the start. 
I will review the latest changes and let you know, but so far looks promising!

kingjia90:
merged, thank you again for all these features @spotbot2k and wish you a nice weekend!

PR #12571, opened 2022-07-01
fix: do not use application cache while building system cache

dkarlovi:

This change makes sure the application cache is not invoked when building system cache. This can cause some unexpected bugs, but can also outright fail because app cache might not be available at the time of system cache build at all.

### DIFF
--- a/lib/HttpKernel/CacheWarmer/PimcoreCoreCacheWarmer.php
+++ b/lib/HttpKernel/CacheWarmer/PimcoreCoreCacheWarmer.php
@@ -46,7 +46,12 @@ class PimcoreCoreCacheWarmer implements CacheWarmerInterface
 
         if (\Pimcore::isInstalled()) {
             try {
+                $cache = \Pimcore\Cache::isEnabled();
+                \Pimcore\Cache::disable();
                 $this->dataObjectClasses($classes);
+                if ($cache) {
+                    \Pimcore\Cache::enable();
+                }
             } catch (\Exception $exception) {
                 if (!$exception instanceof DriverException) {
                     throw $exception;
--- a/lib/HttpKernel/CacheWarmer/PimcoreCoreCacheWarmer.php
+++ b/lib/HttpKernel/CacheWarmer/PimcoreCoreCacheWarmer.php
@@ -45,19 +45,20 @@ class PimcoreCoreCacheWarmer implements CacheWarmerInterface
         $this->modelClasses($classes);
 
         if (\Pimcore::isInstalled()) {
+            $cache = \Pimcore\Cache::isEnabled();
             try {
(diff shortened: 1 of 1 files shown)


lib/HttpKernel/CacheWarmer/PimcoreCoreCacheWarmer.php
```diff
@@ -46,7 +46,12 @@ public function warmUp($cacheDir): array
 
         if (\Pimcore::isInstalled()) {
             try {
+                $cache = \Pimcore\Cache::isEnabled();
+                \Pimcore\Cache::disable();
                 $this->dataObjectClasses($classes);
+                if ($cache) {
+                    \Pimcore\Cache::enable();
+                }
```
NiklasBr: Will the cache be re-enabled if `$this->dataObjectClasses($classes);` throws? If not, maybe add a `finally` and move the `::enable()` there so it may be re-enabled regardless?

dkarlovi: Nice suggestion, thanks @NiklasBr!

brusch:
I'm wondering that the cache is used inside  `$this->dataObjectClasses($classes);` - can you maybe share some more details on that? 
I mean your change doesn't hurt but anyway, I'd like to document the reason for the change here 😊

I'm also thinking whether we should make the cache warmer independent from the database, just as we do it here: 
https://github.com/pimcore/pimcore/blob/a47cf15bb7121aca050f195b5f8dab27a866c0e7/bundles/CoreBundle/Command/ClassesDefinitionsBuildCommand.php#L55

dkarlovi:
@brusch I'm introducing Redis cache in my Dockerized Pimcore build and want to prebuild Symfony DIC (run the warmers) while building the image.

The change was introduced since Pimcore will try to use the app cache for stuff like `getList()`, IIRC. I don't have the project in front of me to test right now. The same type of issue is [also caused by Coreshop because it's injecting `@session`](https://github.com/coreshop/CoreShop/issues/2033).

IMO this warmer overall needs work because, as is, it's consuming a lot of memory to do what it does. But, the PR is just a bugfix for the obvious issue, app cache being used while system cache being constructed.

brusch:
@dkarlovi please check #12571 - that should fix the issue as well and additionally makes the cache warmer independet from the database. WDYT?

dkarlovi:
You've linked to this PR. 😉

dkarlovi:
I guess you mean https://github.com/pimcore/pimcore/pull/12577?

dkarlovi:
The other one looks good to me, closing here, thanks @brusch!

brusch:
@dkarlovi 🤦‍♂️ thanks for clarifying!

PR #12572, opened 2022-07-01
Allow disabling of marking the active trail when building the navigation

jdreesen:
Sometimes, it's not necessary to mark the active trail in the navigation, which is rather expensive for large navigation trees, and there's currently no way of disabling it, so I decided to add an option for it.

> **Note**:
> The first commit is the actual change, the second one refactors a bit.

### DIFF
--- a/lib/Navigation/Builder.php
+++ b/lib/Navigation/Builder.php
@@ -90,6 +90,7 @@ class Builder
             'cacheLifetime' => null,
             'maxDepth' => null,
             'active' => null,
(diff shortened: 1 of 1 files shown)


brusch:
Totally makes sense, thanks! 👍

jdreesen:
I'd love to see this in `v10.5`, too.

dvesh3 (approved):

dvesh3:
@jdreesen thanks!

PR #12585, opened 2022-07-05
getHttpData() throws exception without query params

MercuryKojo:
an isset validation is needed before merging the get params of the url to not throw an exception in php8.1

  
  

Resolves #


brusch:
Thank you very much 🙋

PR #12587, opened 2022-07-05
12569 - Fix. Ignore classes === null.

### LINKED ISSUE #12569: [Bug]: Exception when clicking on folder under dataobjects in Backend
### Expected behavior

Using pimcore actual version with coreshop.

When I try to click on folder coreshop, I expect, as on every other folder in dataobjects, a list of objects is listed, or a selection appears where I can choose which objects I like to see listed.
I see described Excption every time I click on the folder 'coreshop'

expected:
no exception, existing classes should be selectable

### Actual behavior

Exception is raised:
```
URL: /admin/object/get-folder
Message: Call to a member function getId() on null
Trace: 
in /var/www/dev/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php:777
#0 /var/www/dev/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php(729): Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectController->prepareChildClasses()
#1 /var/www/dev/vendor/symfony/http-kernel/HttpKernel.php(152): Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectController->getFolderAction()
#2 /var/www/dev/vendor/symfony/http-kernel/HttpKernel.php(74): Symfony\Component\HttpKernel\HttpKernel->handleRaw()
#3 /var/www/dev/vendor/symfony/http-kernel/Kernel.php(202): Symfony\Component\HttpKernel\HttpKernel->handle()
#4 /var/www/dev/public/index.php(35): Symfony\Component\HttpKernel\Kernel->handle()
#5 {main}
```

### Steps to reproduce

not sure how this can be reproduced. On my system every time.

I could provide a PR if wanted, that null-objects are ignored for building the selection list.

Stefan-JustMe69-SQ:
Resolves #12569

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -769,6 +769,10 @@ class DataObjectController extends ElementControllerBase implements KernelContro
     {
         $reduced = [];
         foreach ($classes as $class) {
+            if (null === $class) {
+                continue;
+            }
+
             $reduced[] = [
                 'id' => $class->getId(),
                 'name' => $class->getName(),
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -761,11 +761,11 @@ class DataObjectController extends ElementControllerBase implements KernelContro
     }
 
     /**
-     * @param DataObject\ClassDefinition[] $classes
+     * @param DataObject\ClassDefinition[]|null $classes
      *
      * @return array
      */
-    protected function prepareChildClasses($classes)
+    protected function prepareChildClasses(array $classes): array
     {
         $reduced = [];
         foreach ($classes as $class) {
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -761,7 +761,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
     }
 
     /**
-     * @param DataObject\ClassDefinition[]|null $classes
+     * @param array<DataObject\ClassDefinition[]|null> $classes
      *
      * @return array
      */
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -761,7 +761,7 @@ class DataObjectController extends ElementControllerBase implements KernelContro
     }
 
     /**
-     * @param array<DataObject\ClassDefinition[]|null> $classes
+     * @param array<DataObject\ClassDefinition|null> $classes
      *
      * @return array
      */


mcop1:
@JustMe69 could you please have a look at the failing action (Static Analysis)?

Stefan-JustMe69-SQ:
So please tell me how?

Exception ist telling that $class **is** null somehow:
```
URL: /admin/object/get-folder
Message: Call to a member function getId() on null
Trace: 
in /var/www/dev/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php:777
#0 /var/www/dev/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php(729): Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectController->prepareChildClasses()
#1 /var/www/dev/vendor/symfony/http-kernel/HttpKernel.php(152): Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectController->getFolderAction()
#2 /var/www/dev/vendor/symfony/http-kernel/HttpKernel.php(74): Symfony\Component\HttpKernel\HttpKernel->handleRaw()
#3 /var/www/dev/vendor/symfony/http-kernel/Kernel.php(202): Symfony\Component\HttpKernel\HttpKernel->handle()
#4 /var/www/dev/public/index.php(35): Symfony\Component\HttpKernel\Kernel->handle()
#5 {main}
```

Your code-tool is telling that it cannot be null?

So how should I fix this?

mcop1:
@JustMe69 just had a look and to be honest I don´t see a problem with your solution.
But I would suggest to change the check to 
```
if($class !== null) {
....
}
```
At least this was working when I tested it in my dev environment. I would have added a code suggestion, but somehow I can´t right now. Please forgive me, it is still kinda early :)

jdreesen:
@mcop1 your suggestion is semantically the same and doesn't change anything. Both is right. I think the problem is the `@param DataObject\ClassDefinition[] $classes` annotation of the method, which tells PHPStan, that the array only contains objects of type `ClassDefinition` and not `null`, which makes it think the `null` check is useless. Probably changing it to `@param array<DataObject\ClassDefinition|null> $classes` would help, but I'm not sure if it's the best way to fix this. Maybe @blankse can help here?

dpfaffenbauer:
@jdreesen `ClassDefinition::getById` can return null (https://github.com/pimcore/pimcore/blob/10.x/models/DataObject/ClassDefinition.php#L312), so your answer is actually the solution, the array param value should allow in the array too.

Stefan-JustMe69-SQ:
The @param is ok I think.
The problem ist, that the param array can **contain null-values**, which are leading to the exception.
So the problem occurs when ```DataObject\ClassDefinition[] $classes``` is not null but contains null-values in the array.
The fix is to check and jump over the null values.

jdreesen:
@JustMe69 yes, and that's why you have to change the `@param` annotation to tell PHPStan that the `$classes` array may contain `null` values. `DataObject\ClassDefinition[]` says it contains **only** objects of this type, nothing else, so checking for `null` would be useless. If you change it to `array<DataObject\ClassDefinition|null>` it says: this array may contain objects of this type **or** `null` values, which actually seems to be the case here.

mcop1:
> @mcop1 your suggestion is semantically the same and doesn't change anything. Both is right. I think the problem is the `@param DataObject\ClassDefinition[] $classes` annotation of the method, which tells PHPStan, that the array only contains objects of type `ClassDefinition` and not `null`, which makes it think the `null` check is useless. Probably changing it to `@param array<DataObject\ClassDefinition|null> $classes` would help, but I'm not sure if it's the best way to fix this. Maybe @blankse can help here?

I am aware that it is basically the same, that´s why I tested it locally. But your suggested solution is better anyways, tested it in my dev env and it is working. Thank you :)

Stefan-JustMe69-SQ:
Thanks @ all for your help with my first pimcore-PR

mcop1:
> Thanks @ all for your help with my first pimcore-PR

And we thank you for submitting the pr 👍

dpfaffenbauer:
@JustMe69 💪

blankse:
Why not fix the issue in the getClasses method of the DataObject model? Why should we have null entries in the return array?
https://github.com/pimcore/pimcore/blob/acd4512a0090bb3b748c34b25354c1f9fbd809ae/models/DataObject/AbstractObject/Dao.php#L488

jdreesen:
Done in #12608

PR #12588, opened 2022-07-05
Fixed object reference setting for objectbricks and fieldcollections …

### LINKED ISSUE #12290: [Bug]: Broken pimcore cache for relations inside objectbricks
### Expected behavior

Relation in Objectbricks should work without any problems.

### Actual behavior

If the field is a mandatory field, an validation exception is thrown. If it' not a mandatory field, the relation will be removed.

It can be reproduced in the official demo:
image
image

### Steps to reproduce

1. Create a mandatory relation field inside a ojectbrick
2. Create a data object and assigne any object to relation to the new field
3. Reload the object (or open an existing object with a relation) and save it without change the relation

<hr>

## More information

The bug occurs when the cache is fetched from the cache_items table while saving the object. By opening the object and clearing the Pimcore cache before saving, everything works.

fashxp:
…when loading data from cache

fixes #12290

### DIFF
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -800,9 +800,8 @@ class Concrete extends DataObject implements LazyLoadedFieldsInterface
 
         // renew object reference to other object aware fields
         foreach($this->__objectAwareFields as $objectAwareField => $exists) {
-            $brickContainer = $this->$objectAwareField;
-            if($brickContainer instanceof ObjectAwareFieldInterface) {
-                $brickContainer->setObject($this);
+            if(isset($this->$objectAwareField) && $this->$objectAwareField instanceof ObjectAwareFieldInterface) {
+                $this->$objectAwareField->setObject($this);
             }
         }
     }
--- a/models/DataObject/Concrete.php
+++ b/models/DataObject/Concrete.php
@@ -46,6 +46,15 @@ class Concrete extends DataObject implements LazyLoadedFieldsInterface
     /**
      * @internal
      *
+     * Necessary for assigning object reference to corresponding fields while wakeup
+     *
+     * @var array
+     */
+    public $__objectAwareFields = [];
+
+    /**
+     * @internal
+     *
      * @var array
      */
     public const SYSTEM_COLUMN_NAMES = ['id', 'fullpath', 'key', 'published', 'creationDate', 'modificationDate', 'filename', 'classname', 'index'];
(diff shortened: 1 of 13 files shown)


fashxp:
@skochTC  can you please check if that solves your issue (don't forget to clear the pimcore data cache)

skoch-twocream:
> @skochTC can you please check if that solves your issue (don't forget to clear the pimcore data cache)

I checked it several times now and compared it with the previous bug behavior. With your change it works fine!

Thanks a lot!

models/DataObject/Concrete.php
```diff
@@ -788,6 +797,14 @@ public function __wakeup()
         if (property_exists($this, 'localizedfields') && $this->localizedfields instanceof Localizedfield) {
             $this->localizedfields->setObject($this, false);
         }
+
+        // renew object reference to other object aware fields
+        foreach($this->__objectAwareFields as $objectAwareField => $exists) {
+            $brickContainer = $this->$objectAwareField;
+            if($brickContainer instanceof ObjectAwareFieldInterface) {
+                $brickContainer->setObject($this);
+            }
```
brusch:
```suggestion
            if(isset($this->$objectAwareField) && $this->$objectAwareField instanceof ObjectAwareFieldInterface) {
                $this->$objectAwareField->setObject($this);
            }
```
I'd add `isset($this->$objectAwareField)` here, just in case the class definition changed in the meantime and a field definition was removed.

brusch (changes requested):

models/DataObject/Concrete.php
fashxp: but `$brickContainer instanceof ObjectAwareFieldInterface` doesn't evaluate to true, if `$this->$objectAwareField` is not set, right?

brusch: But it would throw an `Warning: Undefined property: ....` error.

brusch: Just verified that: 
```php
        $image = Asset\Image::getById(360);

        // works
        if(isset($image->sdfdsgdfgfghfgh) && $image->sdfdsgdfgfghfgh instanceof Asset) {
            $foo = 'bar';
        }

        // throws Warning: Undefined property: Pimcore\Model\Asset\Image::$sdfdsgdfgfghfgh
        if($image->sdfdsgdfgfghfgh instanceof Asset) {
            $foo = 'bar';
        }
```

PR #12595, opened 2022-07-06
[Advanced relations] Do not check for type equality of meta fields

BlackbitDevs:
The `isEqual()` methods shall compare in non-strict mode - at least this is what most other data types do (see https://github.com/pimcore/pimcore/blob/e41f4fd1be7816fc2950bf3adc0d8e2fc5b174e8/models/DataObject/Traits/SimpleComparisonTrait.php#L31) - but currently the meta fields of advanced relations are compared strictly.

Steps to reproduce:
1. Create class `Category`
1. Create data object class `Product` with advanced many-to-many object relation `relation`, allow class `Category`, add meta field `test` with type `number`
3. Use following script:
```php
$category = new \Pimcore\Model\DataObject\Category;
$product = new \Pimcore\Model\DataObject\Product;
$metaObject1 = new \Pimcore\Model\DataObject\Data\ObjectMetadata('relation', ['test']);
$metaObject1->setObject($category);
$metaObject1->setTest(2);

$metaObject2 = new \Pimcore\Model\DataObject\Data\ObjectMetadata('relation', ['test']);
$metaObject2->setObject($category);
$metaObject2->setTest("2");

$fieldDefinition = $product->getClass()->getFieldDefinition('relation');
var_dump($fieldDefinition->isEqual([$metaObject1], [$metaObject2]); // returns false
```


mcop1:
@BlackbitNeueMedien PR looks good, but basically we are changing the behaviour. 
I would suggest to do it in 10.5, add an upgrade note and I think it is perfectly made for an automated test too. What do you think?

brusch:
Hmmm, sounds very risky, maybe even a topic for 11? 🤔
I'm also wondering why `serialize()` is used at the moment ... strange

mcop1:
Since we want to release 10.4.6 and 10.5. asap, I moved this issue to milestone 11.0.0

mcop1:
@BlackbitNeueMedien We would like to merge this pr into Pimcore 11, but since it is a change of behaviour, could you please add  something to the upgrade notes?

BlackbitDevs:
@mcop1 Done.

PR #12616, opened 2022-07-08
Deprecate Registering bundles through extensions.php

### LINKED ISSUE #12274: Deprecate Extension Manager
### Improvement description

Use Symfony Flex's `config/bundle.php` for bundles instead. 
For Areabricks I think we don't need a management (enable/disable), they are just getting loaded automatically. 

- [x] check if non-flex bundles are supported in `bundles.php` - yes, it works!
- [x] when there are extensions defined in the `bundles.php` then need to be activated by default.
- [x] issue deprecation warnings when there is content in the `extensions.php`
- [x] make `extensions.php` optional
- [x] deprecate enable/disable in the `pimcore:bundle` command
- [x] deprecate all related classes
- [x] add deprecation message for bundle entries in the UI

dvesh3:
Resolves #12274


bundles/CoreBundle/Command/Bundle/DisableCommand.php
```diff
@@ -46,6 +48,12 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        trigger_deprecation(
```
brusch, line 52: Should we maybe directly send this info to the output as well, so that you notice when manually executing it? 🤔

brusch (changes requested):
I think a lot of methods in `\Pimcore\Extension\Bundle\PimcoreBundleManager` can be deprecated as well, such as `getActiveBundle()` or `enable()` and a lot more, right? 

`\Pimcore\Extension\Bundle\PimcoreBundleInterface::getAdminIframePath()` is also obsolete as there's no UI anymore in 11, so should be deprecated as well.

dvesh3:
> I think a lot of methods in `\Pimcore\Extension\Bundle\PimcoreBundleManager` can be deprecated as well, such as `getActiveBundle()` or `enable()` and a lot more, right?
> 
> `\Pimcore\Extension\Bundle\PimcoreBundleInterface::getAdminIframePath()` is also obsolete as there's no UI anymore in 11, so should be deprecated as well.

@brusch review changes: 

- Deprecated methods related to `enable` \ `disable` states in `PimcoreBundleManager`. 
- `getActiveBundles()` & `getActiveBundle()` is not deprecated, since it also considers Kernel & manually registered bundles.
- Deprecated `PimcoreBundleInterface::getAdminIframePath()` method

bundles/CoreBundle/Command/Bundle/DisableCommand.php
dvesh3: Sending deprecation info to output with verbose check.

brusch:
@dvesh3 regarding `getActiveBundles()` & `getActiveBundle()`, I'm wondering if we should rename them (incl. BC layer), because there are no "active" bundles anymore since we remove the possibility to enable/disable them? WDYT?

dvesh3:
> @dvesh3 regarding `getActiveBundles()` & `getActiveBundle()`, I'm wondering if we should rename them (incl. BC layer), because there are no "active" bundles anymore since we remove the possibility to enable/disable them? WDYT?

@brusch As the phpdoc comment says "_List of currently active bundles from kernel_" and also there could be bundles manually registered on the kernel. The list is not based on enabled status.

BlackbitNeueMedien:
What is the goal of this PR? In the end it replaces `var/config/extensions.php` by `app/config/bundles.php`. 
In the past there were some changes for better cloud hosting support but here one file-based config gets exchanged for another file-based config.
Plus the addition that it is now more error-prone because you have to edit a real PHP file for installation instead of being able to execute `bin/console pimcore:bundle:enable MyBundle`.

PR #12624, opened 2022-07-11
fix: video file extension check should not be case-sensitive

rberneder:
File extensions of video editables are currently compared with case-sensitivity. However there are many reasons why editors might upload videos with all-caps namings or extensions like `.MP4` instead of `.mp4`. In these cases the video output fails because of the case-sensitive comparison and results in a broken-looking "asset not found" placeholder image in the frontend.

### DIFF
--- a/models/Document/Editable/Video.php
+++ b/models/Document/Editable/Video.php
@@ -374,7 +374,7 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
 
         // compatibility mode when FFMPEG is not present or no thumbnail config is given
         if (!\Pimcore\Video::isAvailable() || !$thumbnailConfig) {
-            if ($asset instanceof Asset && preg_match("/\.(f4v|flv|mp4)/", $asset->getFullPath())) {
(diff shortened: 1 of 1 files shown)


dvesh3:
@rberneder thanks for your PR! could you please check off the list here https://github.com/pimcore/pimcore/pull/12624#issuecomment-1180215914?

rberneder:
> @rberneder thanks for your PR! could you please check off the list here [#12624 (comment)](https://github.com/pimcore/pimcore/pull/12624#issuecomment-1180215914)?

Where can I check off this list @dvesh3 ? I cannot do so directly in the comment.

dvesh3:
> > @rberneder thanks for your PR! could you please check off the list here [#12624 (comment)](https://github.com/pimcore/pimcore/pull/12624#issuecomment-1180215914)?
> 
> Where can I check off this list @dvesh3 ? I cannot do so directly in the comment.

Oops, sorry! I thought the PR owner could also check off items in the list. Then it is my task to do 😄

dvesh3:
@rberneder thanks!

PR #12625, opened 2022-07-11
[Bug]: Prevent Document Link targeting itself causing redirection loop

### LINKED ISSUE #12599: [Bug]: Dragging a Document Link onto itself breaks the document tree
### Expected behavior

Pimcore continues to not be broken. It shouldn't be possible for a link to link to itself.

### Actual behavior

The Document tree won't expand the parent node of the link and gets stuck in an endless loop.

### Steps to reproduce

1. Right click on any node in the document tree and "Add Link". /testing-links in this example:
image
2. Name the link, open and "Save & Publish" it. /testing-links/test in this example:
image
3. Take the same link, drag it onto itself and "Save & Publish" again.
image
5. The document tree is now stuck loading and won't expand /testing-links. Eventually, it fails with an out of memory exception.
image

Deleting the parent node also does not work for the same reason, I assume. A quick and easy fix is to either manually delete the database entry of the link in the `documents` table or change its target ID in the `documents_link` table.

It's a pretty easy thing to mess up on accident, so I think there should be some protection.

Tested on two different installs of 10.4.4

kingjia90:
Resolves #12599

## Additional info  
In the Link model, it silently log an error if the link target is the same id and nullifies the `id`.
On Frontend, it doesn't show it's droppable when the same document is dragged over. 
When using the search, it throws an error if you select the link itself.

## To reproduce

Create some custom implementation that tries to change a link target path to be the same as itself (same as in the provided tests)
Create a link and try to link itself by drag and drop (from folder tree)
Create a link and use the search function to assign itself to the redirection target

### DIFF
--- a/models/Document/Link.php
+++ b/models/Document/Link.php
@@ -368,11 +368,11 @@ class Link extends Model\Document
         try {
             if ($this->internal) {
                 if ($this->internalType == 'document') {
-                    $this->object = Document::getById($this->internal);
 
-                    if ($this->getId() == $this->internal){
-                        throw new \Exception('Prevented infinite redirection loop: attempted to linking "'. $this->getKey().'" to itself. ');
+                    if ($this->getId() == $this->internal) {
+                        throw new \Exception('Prevented infinite redirection loop: attempted to linking "' . $this->getKey() . '" to itself. ');
                     }
+                    $this->object = Document::getById($this->internal);
 
                 } elseif ($this->internalType == 'asset') {
                     $this->object = Asset::getById($this->internal);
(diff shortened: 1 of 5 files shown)


kingjia90:
I have a problem, the test provided actually "cannot fail": it gets into a infinite loop until it eventually crashes/times-out. Any suggestion on improving this?

At ther moment, i've just added a debug line to give a hint of potential problem
image

models/Document/Link.php
```diff
@@ -369,6 +369,11 @@ private function setObjectFromId()
             if ($this->internal) {
                 if ($this->internalType == 'document') {
                     $this->object = Document::getById($this->internal);
+
+                    if ($this->getId() == $this->internal){
```
dvesh3: check and throw exception before loading the object

PR #12627, opened 2022-07-11
Added dedicated method to get thumbnails in specific format

### LINKED ISSUE #12484: [Improvement]: Add public method that returns thumbnails in a specific format
### Improvement description

Offer a public method that return thumbnails in the desired format, e.g.

`$thumbnail = $asset->getThumbnail('foo')->getAsFormat('avif');`

After that add a support for this method in datahub/graphql (https://github.com/pimcore/data-hub/issues/541).

aashan10:
Resolves #12484


models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -429,4 +429,22 @@ private function getFilename(): string
...
     }
+
+
+    /**
+     * @description Returns path for thumbnail image in a given file format
+     * @param string $format
+     * @param bool $deferred
+     * @return string
+     */
+    public function getThumbnailPathAs(string $format, $deferred = true)
```
brusch: @aashan10 I'd really go for `getAsFormat(string $format): self` 
So I think we should return a thumbnail instance again instead of a path, because then you have much more possibilities, such as: 
```php
$asset->getThumbnail('foo')->getAsFormat('avif')->getStream();
// or
$asset->getThumbnail('foo')->getAsFormat('avif')->getMimeType();
```

brusch (changes requested):

models/Asset/Thumbnail/ImageThumbnailTrait.php
aashan10: Modification has been made accordingly!

models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -429,4 +430,24 @@ private function getFilename(): string
...
+     * @param string $format
+     * @return self
+     */
+    public function getAsFormat(string $format): self
+    {
+        $config = $this->getConfig();
+
+        if (!$config) {
+            $config = new Config();
+        }
```
jdreesen:
```suggestion
        $config = $this->getConfig() ?? new Config();
```

models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -429,4 +430,24 @@ private function getFilename(): string
 
         return basename($pathReference['src']);
     }
+
+
+    /**
+     * @description Returns path for thumbnail image in a given file format
+     * @param string $format
+     * @return self
```
jdreesen:
```suggestion
     * Returns path for thumbnail image in a given file format
     *
     * @param string $format
     * @return $this
```

jdreesen: Are you sure you don't want to change this? 
The `@description` tag isn't used anywhere else in Pimcore and I think the `@return $this` is more correct if you `return $this;`.

models/Asset/Thumbnail/ImageThumbnailTrait.php
jdreesen: Don't you want to change this? The `@description` tag isn't used anywhere else in Pimcore and I think the `@return $this` is more correct if you `return $this;`.

models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -429,4 +430,20 @@ private function getFilename(): string
...
+     */
```
jdreesen: Are you sure you don't want to change this?

The `@description` tag isn't used anywhere else in Pimcore and I think the `@return $this` is more correct if you `return $this;`.

```suggestion
    /**
     * Returns path for thumbnail image in a given file format
     *
     * @param string $format
     * @return $this
     */
```

models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -429,4 +430,20 @@ private function getFilename(): string
...
+    public function getAsFormat(string $format): self
+    {
+        $config = $this->getConfig() ?? new Config();
+        $config->setFormat($format);
+
+        $this->config = $config;
+
+        return $this;
```
brusch:
```suggestion
        $thumb = clone $this;
        $config = clone $thumb->getConfig();
        $config->setFormat($format);
        $thumb->setConfig($config); 

        return $thumb ;
```
We need to clone the instance, otherwise this will change the behavior of the original object.

brusch (changes requested):

models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -429,4 +430,23 @@ private function getFilename(): string
 
         return basename($pathReference['src']);
     }
+
+
+    /**
+     * Returns path for thumbnail image in a given file format
+     *
+     * @param string $format
+     * @return $this
```
jdreesen: Well, now `$this` isn't correct anymore, as it doesn't actually return `$this` but a clone of it...

aashan10: Hmm.. fair point. What would you suggest we should use here @jdreesen?

dvesh3:
```suggestion
     * @return static
```

brusch (approved):

mcop1:
`getPath` method isn´t returning the path for **webp**, **avif**, ... which is what I would expect. I tried the following code:

```
  $asset = Asset::getById(67);
  $thumbnail = $asset->getThumbnail("exampleCorners");

  //Outputs jpeg path as expected
  print($thumbnail->getPath() . '<br/>');

  $format = $thumbnail->getAsFormat("webp");

  //Outputs html picture tag with webp and jpeg only
  print(htmlspecialchars($format->getHTML()) . '<br/>');

  //Outputs jpeg path, but webp expected
  print($format->getPath(false)  . '<br/>');

  //Outputs jpeg path, but webp expected
  print($format->getPath(true)  . '<br/>');
```

I think we need to clear the `pathReference` array as well. Maybe we could just call the `$thumb->reset()` in `getAsFormat` method? 

@aashan10 could you please have a look?

brusch:
Nice catch - LGTM now 👍

mcop1:
@aashan10 @brusch 
Working now, nice, thank you. 

Sorry, I could have mentioned that right away, but what about documentation? I think we should at least add an example somewhere here:
https://github.com/pimcore/pimcore/blob/748751efc4f246914e9649dc008614dca3f5a3ff/doc/Development_Documentation/04_Assets/03_Working_with_Thumbnails/01_Image_Thumbnails.md?plain=1#L194

And there are already automated tests for thumbnails, it would be a good thing if we add a test for `getAsFormat` too:
https://github.com/pimcore/pimcore/blob/748751efc4f246914e9649dc008614dca3f5a3ff/tests/Model/Asset/AssetTest.php#L102

brusch:
@mcop1 agree, should be an easy one to add
@aashan10 can you add the tests? Thanks 😊

PR #12630, opened 2022-07-12
fixed object/edit.js: test if type of addReferencedField is a functio…

sevarozh:
test if typeof addReferencedField is a function before call, otherwise can throw a exception by 'undefined


fashxp:
@sevarozh  can you please sign the CLA, thx

mcop1:
@sevarozh could you please sign the CLA?

blankse:
I created an alternative PR with other wrong tests: https://github.com/pimcore/pimcore/pull/12821

PR #12639, opened 2022-07-13
Fix docs to mention both translation lists for POEditor.

timwiechers:
This PR will update the documentation to mention both available translation lists for POEditor.


dvesh3:
@timwiechers thanks!

PR #12641, opened 2022-07-13
[CustomLayouts] Add filter option on listing

### LINKED ISSUE #12544: [CustomLayouts] Add filter option on listing
### Feature description

 - Add option to filter custom layout listing which should replace `setCondition` on Pimcore 11
 - Deprecate `setCondition`

mcop1:
Resolves #12544 

- Moved get/setFilter methods from `JsonListing` to a new trait. Created `CallableFilterListingInterface`.
- Moved get/setOrder methods from `JsonListing` to a new trait. Created `CallableOrderListingInterface`.
- Deprecated JsonListing class
- Added interfaces + traits to the classes that extended from `JsonListing`
- Tried to adapt `CustomLayout\Listing` to match the other classes that inherit from `LocationAwareConfigRepository` as close as possible without actually inheriting the class.
- Added update notes

For testing: 
- Please make sure to test custom layouts for bricks too.

### DIFF
--- a/lib/Model/Listing/FilterListingInterface.php
+++ b/lib/Model/Listing/FilterListingInterface.php
@@ -1,0 +1,32 @@
+<?php
+
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\Listing;
+
+interface FilterListingInterface
+{
+    /**
+     * @param callable|null $filter
+     */
+    public function setFilter($filter);
+
+    /**
+     * @return callable|null
+     */
+    public function getFilter();
+
+
+}
--- a/lib/Model/Listing/OrderListingInterface.php
+++ b/lib/Model/Listing/OrderListingInterface.php
@@ -1,0 +1,30 @@
+<?php
+
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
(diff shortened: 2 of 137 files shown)


models/Asset/Image/Thumbnail/Config/Listing.php
```diff
@@ -16,12 +16,20 @@
...
+use Pimcore\Model\Listing\FilterListingInterface;
+use Pimcore\Model\Listing\OrderListingInterface;
+use Pimcore\Model\Listing\Traits\FilterListingTrait;
+use Pimcore\Model\Listing\Traits\OrderListingTrait;
 
 /**
  * @method \Pimcore\Model\Asset\Image\Thumbnail\Config\Listing\Dao getDao()
  */
-class Listing extends Model\Listing\JsonListing
+class Listing extends AbstractModel implements FilterListingInterface, OrderListingInterface
```
dvesh3: I would still extend from `JsonListing` on 10.5 since this can lead to BC break, if there's an existing code:
```
$listing = new Asset\Image\Thumbnail\Config\Listing();
if ($listing instanceof JsonListing) {
  $listing ->setFilter(function () { ...
}
```

dvesh3: also this will not trigger the deprecations, since the methods are overridden by traits.

lib/Model/Listing/FilterListingInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\Listing;
+
+interface FilterListingInterface
```
dvesh3: How about naming it `CallableFilterListingInterface`?

lib/Model/Listing/OrderListingInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\Listing;
+
+interface OrderListingInterface
```
dvesh3: rename to `CallableOrderListingInterface`?

PR #12643, opened 2022-07-13
[Admin] Add delete confirm dialog boxes

### LINKED ISSUE #6259: [Improvement/Epic] More delete confirmations and giving element name about to be deleted
Hello,

Following https://github.com/pimcore/pimcore/issues/6240 which is really a great things, I went through all the Pimcore GUI to check if deleting elements are risky.
I found the followings which could be improved.

There is no confirmation, as soon clicked, as soon lost ; )
- tag & snippet management
- qr code
- targeting rules/groups
- custom reports
- thumbnails

There is a popin confirmation but it doesn't diplay the name of the element which is about to be deleted, which could be prone to error
- tags
- roles / users
- fieldcollection
- objectbricks

Those are a little different as it use the small red `x` icon in extjs grid, but I think they are valuable to a confirm popin:
- keydefinition from storecollections... this could be a pretty hard delete has there is no confirm and maybe data in bdd are lost ?
- groups from storecollections

In more generally there are other part in the GUI with red `x` icon without any confirmation (like static routes, etc.). They are less sensitive but can defenitly break an existing application if clicked by error.
I think maybe a global handler on those red "x" could be added.

This issue could be splitted in smallest issues. It could be left as it and planned later.
But in the end purpose is to achieve a better UI with good confirm message and when possible saying what is about to be deleted.

In our production project we have sometimes editors saying, it doesn't work no more, and we found that something has been deleted. Arh : (( We try to limit permissions for sure, but it is often a human wrong clic and having a good preventing system with message could really imporve this I think.

In the long run, it will prepare Pimcore for touch screen, as wrong click can be really easy this way, and deleting should be confirmed for sure.

Best regards.

dvesh3:
Resolves #6259

### DIFF
--- a/bundles/EcommerceFrameworkBundle/Resources/public/js/pricing/config/panel.js
+++ b/bundles/EcommerceFrameworkBundle/Resources/public/js/pricing/config/panel.js
@@ -295,16 +295,19 @@ pimcore.bundle.EcommerceFramework.pricing.config.panel = Class.create({
      * delete existing rule
      */
     deleteRule: function (tree, record) {
-        Ext.Ajax.request({
-            url: Routing.generate('pimcore_ecommerceframework_pricing_delete'),
-            method: 'DELETE',
-            params: {
-                id: record.id
-            },
-            success: function () {
-                this.refresh(this.tree.getRootNode());
-            }.bind(this)
-        });
+        pimcore.helpers.deleteConfirm(t('bundle_ecommerce_pricing_rule'), record.data.text, function () {
+            Ext.Ajax.request({
+                url: Routing.generate('pimcore_ecommerceframework_pricing_delete'),
+                method: 'DELETE',
+                params: {
+                    id: record.id
+                },
+                success: function () {
+                    this.refresh(this.tree.getRootNode());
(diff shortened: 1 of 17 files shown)


bundles/EcommerceFrameworkBundle/Resources/public/js/pricing/config/panel.js
```diff
@@ -295,16 +295,19 @@ pimcore.bundle.EcommerceFramework.pricing.config.panel = Class.create({
...
-            url: Routing.generate('pimcore_ecommerceframework_pricing_delete'),
-            method: 'DELETE',
-            params: {
-                id: record.id
-            },
-            success: function () {
-                this.refresh(this.tree.getRootNode());
-            }.bind(this)
-        });
+        pimcore.helpers.deleteConfirm(t('bundle_ecommerce_pricing_rule'), record.data.text, function () {
```
aryaantony92, line 298:
```suggestion
        pimcore.helpers.deleteConfirm(t('bundle_ecommerce_pricing_rules'), record.data.text, function () {
```

Issue #10614, opened 2021-10-15
[Improvement]: Use package-manager (npm, ...) to install frontend dependencies


brusch:
3rd party frontend dependencies should be installed using npm and packed by [Encore](https://symfony.com/doc/current/frontend/encore/installation.html). 

This includes all libraries **EXCEPT** ExtJS, because installing this via npm requires a registration at Sencha and is not very convenient. 

Examples: 
- Ace
- leaflet
- minipaint
- vrview
- Bootstrap (Ecommerce Bundle)

See also: #10606

HaaseIT:
NPM/Yarn is not actually needed for this. You could add this without BC breaking issues I guess with:

https://github.com/oomphinc/composer-installers-extender

All you need to do is include tha package and some config for it in the compoer.json:

```
"extra": {
  "installer-types": ["bower-asset", "npm-asset"],
  "installer-paths": {
    "public/static/js/{$type}/{$name}/": [
      "type:bower-asset",
      "type:npm-asset"
    ]
  }
}
```

On another note, not really fitting this issue, but why not bundle symfony encore for js/scss compression?

brusch:
Topics to consider: 
- https://github.com/eloquent/composer-npm-bridge
- how to deal with bundles, maybe something similar to [Ibexa](https://doc.ibexa.co/en/latest/administration/back_office/back_office_elements/importing_assets_from_bundle/#importing-assets-from-a-bundle)? ( [Example used in bundle](https://github.com/ezsystems/ezplatform-admin-ui/blob/2.3/src/bundle/Resources/encore/ez.config.js) )

brusch:
This is the actual implementation how Ibexa does it: 
https://github.com/ibexa/core/pull/120

robertSt7:
The current approach is to run `npm install` and the build manually in each repo and commit the assembled package with the respective versions. This has the advantage that the package versions are tested and no npm is needed to run Pimcore. 

Alternatively, `npm install` could also trigger with `composer update` and follow the approach of Ibexa and collect all the configs from the bundles and build them at once. This is very dynamic, as the config files from all bundles are added automatically, and everything is built from one place. 

Community input very much appreciated!!!

dpfaffenbauer:
I'd be fine with that every bundle says you need to include these styles in your webpack config and be done with it.

brusch:
@robertSt7 reopened, see [comment](https://github.com/pimcore/pimcore/pull/14182/files#r1111717827)

PR #12644, opened 2022-07-13
Minimize admin js scripts

### LINKED ISSUE #11505: [Bug]: scripts used for admin minimized on demand (even in prod env) instead of at built-time
### Expected behavior

The admin UI is a build-time requirement, it should be satisfied as such.

Currently, scripts listed in `vendor/pimcore/pimcore/bundles/AdminBundle/Resources/views/Admin/Index/index.html.twig` get processed and minified into `var/admin/minified_javascript_core_bf4669852c609bea6be38a41c0f8485e.js` on demand (meaning, in runtime), this file then gets served by `\Pimcore\Bundle\AdminBundle\Controller\Admin\MiscController::scriptProxyAction`. This means:
1. `var/admin` **must** be writable at runtime
2. if you have Pimcore deployed in a HA environment, each instance will need to build their own copy of this
3. you can't (easily) do this work upfront, say when building a Docker image

Proposal here is to:
- move these files into `var/cache/<env>`
- add a Symfony cache warmer which will generate these files upfront when doing `cache:warmup` or `cache:clear --warmup`

Ideally, I guess the best possible solution would be to implement Symfony Webpack Encore, but that's a much more complex thing, as discussed in https://github.com/pimcore/pimcore/issues/10614

### Actual behavior

Assets minimized and processed on demand instead of being pre-processed and placed in the correct place by default.

### Steps to reproduce

Install skeleton Pimcore 10.x, go to /admin/, the files get generated in `var/admin` or a permission error is raised.

aryaantony92:
Resolves #11505


bundles/AdminBundle/Controller/Admin/MiscController.php
```diff
@@ -127,9 +131,18 @@ public function jsonTranslationsSystemAction(Request $request, TranslatorInterfa
...
-            $scriptsContent = $storage->read($storageFile);
+            $fileExtension = \Pimcore\File::getFileExtension ($storageFile);
+
+            $cacheDir = \Pimcore::getKernel ()->getCacheDir ();
+            $jsCacheDir = $cacheDir . '/minifiedJs/';
+            if ($this->adminJsHelperService->isMinifiedScriptExists ($storageFile)) {
+                $scriptsContent = file_get_contents ($jsCacheDir . $storageFile);
+            }
+            else {
+                trigger_deprecation ('pimcore/pimcore', '10.5', 'saving the minified js scripts in the var/admin folder is deprecated and would be removed in Pimcore 11.');
```
jdreesen:
```suggestion
                trigger_deprecation ('pimcore/pimcore', '10.5', 'saving the minified js scripts in the var/admin folder is deprecated and will be removed in Pimcore 11.');
```

bundles/AdminBundle/Helper/AdminJsHelperService.php
```diff
@@ -0,0 +1,734 @@
...
+namespace Pimcore\Bundle\AdminBundle\Helper;
+
+use Pimcore\Localization\LocaleService;
+use Pimcore\Bundle\AdminBundle\Security\ContentSecurityPolicyHandler;
+use Pimcore\Extension\Bundle\PimcoreBundleManager;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\HttpKernel\CacheWarmer\WarmableInterface;
+use Pimcore\Version;
+
+class AdminJsHelperService implements WarmableInterface
```
brusch:
```suggestion
/**
 * @internal
 */
class AdminJsHelperService implements WarmableInterface
```

bundles/AdminBundle/Helper/AdminJsHelperService.php
```diff
@@ -0,0 +1,737 @@
...
+
+        $scriptPaths = ($scriptType == 'bundle') ? $this->bundleScriptPaths : $this->internalScriptPaths;
+
+
+        if ($this->isMinifiedScriptExists ($storageFile)) {
+            return ['storageFile' => basename ($storageFile),
+                '_dc' => \Pimcore\Version::getRevision ()
+            ];
+        }
+        return $this->minifyAndSaveJs ($scriptPaths, $storageFile);
```
dkarlovi: Could we have this run from a cache warmer so it's basically always prebuilt?

bundles/AdminBundle/Controller/Admin/MiscController.php
```diff
@@ -39,7 +40,10 @@
  */
 class MiscController extends AdminController
 {
-    /**
+    public function __construct (private AdminJsHelperService $adminJsHelperService)
```
dvesh3: I would inject the service in the action (where it is needed) instead of the constructor.

bundles/AdminBundle/Helper/AdminJsHelperService.php
aryaantony92: Thanks for the suggestion. And it's already in progress.

dkarlovi: @aryaantony92 thank you very much for your work!

dvesh3: Hi, @dkarlovi Unfortunately, we have to postpone these changes to Pimcore 11, since we have a blocker for making the [AdminJsHelperService](https://github.com/pimcore/pimcore/pull/12644/files/7d0233efae405f12283f218a58b5be852e518f16#diff-97dc22a14872d654fb90e5d557b0f40664ecfae1f921577cebc2aea69480ef98) cache warmable. 

The problem is that we also minify bundles JS files from `PimcoreBundleManager`, which loads the bundles from Kernel and `Pimcore\Extension\Config` by reading `extensions.php` on `Kernel::boot`. The `Pimcore\Extension\Config` is a synthetic service and added to the container after initializing the container [here ](https://github.com/pimcore/pimcore/blob/d924f8e6f6cc163409d9ef959a83cedf8e0104b7/lib/Kernel.php#L311) so when AdminJsHelper service tries to warmup the cache on [initializeContainer ](https://github.com/pimcore/pimcore/blob/d924f8e6f6cc163409d9ef959a83cedf8e0104b7/lib/Kernel.php#L305) it complains about `Extension\Config` service not found https://github.com/pimcore/pimcore/runs/7359155468?check_suite_focus=true#step:9:598

Now, `Pimcore\Extension\Config` service is deprecated on 10.5 version and will be removed in Pimcore 11, so then it will be feasible to make the AdminJsHelperService cache warmable without any issues.

dvesh3:
this might be obsolete in case we decide to implement a package manager for frontend scripts on v11. please see https://github.com/pimcore/pimcore/issues/10614

dkarlovi:
@dvesh3 that would be my preference too. Ideally, using Symfony Encore would completely remove the need to do anything here, it would be handled by Encore itself.

brusch:
Obsolete, we're moving to npm and Encore.

PR #12645, opened 2022-07-13
Improve generated types of many-to-one relation fields

jdreesen:
Many-to-one relation fields return single objects, which means we can specify them as return/parameter type. This is currently only the case in PhpDoc, but not for the real types, which is possible since PHP 8.

Currently, it is:
```php
* @return \Pimcore\Model\Document\Snippet | \Pimcore\Model\Document\Page | \Pimcore\Model\Document\Folder|null
*/
public function getMyField(): ?\Pimcore\Model\Element\AbstractElement
{
```

```php
* @param \Pimcore\Model\Document\Snippet | \Pimcore\Model\Document\Page | \Pimcore\Model\Document\Folder $myField
* @return \Pimcore\Model\DataObject\MyClass
*/
public function setMyField(?\Pimcore\Model\Element\AbstractElement $myField)
{
```

With this PR, it would be:

```php
* @return \Pimcore\Model\Document\Snippet | \Pimcore\Model\Document\Page | \Pimcore\Model\Document\Folder|null
*/
public function getMyField(): \Pimcore\Model\Document\Snippet | \Pimcore\Model\Document\Page | \Pimcore\Model\Document\Folder|null
{
```

```php
* @param \Pimcore\Model\Document\Snippet | \Pimcore\Model\Document\Page | \Pimcore\Model\Document\Folder $myField
* @return \Pimcore\Model\DataObject\MyClass
*/
public function setMyField(\Pimcore\Model\Document\Snippet | \Pimcore\Model\Document\Page | \Pimcore\Model\Document\Folder|null $myField)
{
```

> **Note**
I think this would be a BC break in case someone overrode the models [like described here](https://pimcore.com/docs/pimcore/current/Development_Documentation/Extending_Pimcore/Overriding_Models.html), though.

### DIFF
--- a/bundles/EcommerceFrameworkBundle/src/Model/AbstractOrder.php
+++ b/bundles/EcommerceFrameworkBundle/src/Model/AbstractOrder.php
@@ -18,8 +18,11 @@ namespace Pimcore\Bundle\EcommerceFrameworkBundle\Model;
 
 use Carbon\Carbon;
 use Pimcore\Model\DataObject\Concrete;
+use Pimcore\Model\DataObject\Customer;
 use Pimcore\Model\DataObject\Fieldcollection;
-use Pimcore\Model\Element\AbstractElement;
+use Pimcore\Model\DataObject\Objectbrick;
+use Pimcore\Model\DataObject\OnlineShopOrder;
+use Pimcore\Model\DataObject\OnlineShopVoucherToken;
 
 /**
  * Abstract base class for order pimcore objects
@@ -88,9 +91,9 @@ abstract class AbstractOrder extends Concrete
      */
     abstract public function setGiftItems(?array $giftItems): static;
 
-    abstract public function getCustomer(): ?AbstractElement;
+    abstract public function getCustomer(): ?Customer;
 
-    abstract public function setCustomer(?AbstractElement $customer): static;
+    abstract public function setCustomer(?Customer $customer): static;
 
     abstract public function getPriceModifications(): ?Fieldcollection;
 
@@ -106,9 +109,9 @@ abstract class AbstractOrder extends Concrete
 
     abstract public function getPaymentInfo(): ?Fieldcollection;
 
-    abstract public function setPaymentInfo(?\Pimcore\Model\DataObject\Fieldcollection $paymentInfo): static;
+    abstract public function setPaymentInfo(?Fieldcollection $paymentInfo): static;
 
-    abstract public function getPaymentProvider(): ?\Pimcore\Model\DataObject\Objectbrick;
+    abstract public function getPaymentProvider(): ?Objectbrick;
 
     /**
      * returns latest payment info entry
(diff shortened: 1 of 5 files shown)


mcop1:
@jdreesen I think it would be a good idea to narrow down the return types if possible. Since we are on 11.x now a bc break wouldn´t be a problem anymore. Would you mind to update your pr and maybe have a look at failing tests, please?

jdreesen:
Yes. I'm very busy until Christmas, but I'll try to get on it during the Christmas holidays, maybe.

jdreesen:
I'm a bit stuck at the moment. The problem is, that now precise types (e.g. for generated `DataObject`'s) are generated which doesn't work well with the more general types in e.g. the `EcommerceFrameworkBundle`.

While this is not a problem for return types, since they can be more *precise* than the parent ones, it **is** a problem for parameter types, which **cannot** be (they can only be more *general*). This means that it would not be possible to define an interface or an (abstract) parent class in a bundle and use that for parameter types, since it would no longer be compatible with the generated class.

I'm not really sure what we can do here. Maybe adding the parent/intercaface's (if defined) from the class definition would help (not sure if that would be enough though), but the type definitions would then get very large very quickly.

---

Apart from that, the tests are currently failing because there seem to be some issues with the redis server.

jdreesen:
I fixed it now by adding both the general types and (if defined) the specific ones as parameter types.

But there's another problem: when we add types for generated classes, they **must** be available, which is not always the case at the moment, as it seems that the `EcommerceFrameworkBundle` supports some type of "customer" (provided by the `CustomerManagementFrameworkBundle`, for example) that doesn't necessarily exist.

> Fatal error: Could not check compatibility between Pimcore\Model\DataObject\OnlineShopOrder::getCustomer(): ?Pimcore\Model\DataObject\Customer and Pimcore\Bundle\EcommerceFrameworkBundle\Model\AbstractOrder::getCustomer(): ?Pimcore\Model\DataObject\AbstractObject, because class Pimcore\Model\DataObject\Customer is not available in /home/runner/work/pimcore/pimcore/tests/_output/var/classes/DataObject/OnlineShopOrder.php on line 836

jdreesen:
A compromise could be to omit the generated classes in the parameter/return types (they're still present in the PhpDoc anyway) and only use the general asset/document/object types (as already shown in the description of this PR).

jdreesen:
Not sure why the Redis is always failing, but I don't think this is related.

mcop1:
> Not sure why the Redis is always failing, but I don't think this is related.

Thanks for having a look at this again. Failing redis tests are not related, we have them in other prs too. 
We will discuss the issue and I will come back to you, thanks!

mcop1:
@jdreesen I think we have to go with the more general approach. Because even then we would need to change the `getGroup` and `setGroup` method in the interface 
https://github.com/pimcore/customer-data-framework/blob/f028d4ce886814b9c1f8418c235fc08f710da38c/src/Model/CustomerSegmentInterface.php#L58
to support at least
```
\Pimcore\Model\Document\Page
\Pimcore\Model\Document\Snippet
\Pimcore\Model\Document
\Pimcore\Model\Asset
\Pimcore\Model\DataObject\AbstractObject
null
```

We could change it to `AbstractObject`, but as soon as someone is clicks on the wrong checkbox in the data object class editor (e.g. "Allow Assets") for class `CustomerSegment` you get an type error again. 

What do you think?

mcop1:
Just revisited this pr and unfortunately I don´t think we can implement the proposed changes at this point. 

Since we have implemented the class loader, we get a problem here. The class loader loads the existing class (with `AbstractElement` as return/parameter type) during the creation of the new class. But then there is already a conflict, since the abstract class has `AbstractObject` set as return/parameter type. The environment will then be broken and can only be fixed manually. 
Second problem would be, as already mentioned, you could click on `Allow assets` for in the UI and it will break again. We could add 
```
\Pimcore\Model\Document\Page
\Pimcore\Model\Document\Snippet
\Pimcore\Model\Document
\Pimcore\Model\Asset
\Pimcore\Model\DataObject\AbstractObject
null
```
But then we can keep `AbstractElement` as well. 

So right now, we would like to keep the `AbstractElement` element and therefore close the pr.

PR #12646, opened 2022-07-13
Fix generated PhpDoc parameter type of many-to-one relation field setters

jdreesen:
While creating #12645, I noticed that the PhpDoc parameter type of many-to-one relation field setter misses the `null` part.

While the parameter type of the setter is generated as `?\Pimcore\Model\Element\AbstractElement`, the PhpDoc type is only e.g. `@param \Pimcore\Model\Document\Snippet | \Pimcore\Model\Document\Page`

With this PR, it would be `@param \Pimcore\Model\Document\Snippet | \Pimcore\Model\Document\Page|null` instead.

> **Note**
The PhpDoc type of the getter is already correct: `@return \Pimcore\Model\Document\Snippet | \Pimcore\Model\Document\Page|null`


mcop1:
Thank you!

PR #12652, opened 2022-07-14
[Bug]: Fix document PHP API not saving versionNote

kingjia90:
Resolves https://github.com/pimcore/pimcore/discussions/12558 (discussion)

### DIFF
--- a/tests/Model/Asset/AssetTest.php
+++ b/tests/Model/Asset/AssetTest.php
@@ -96,7 +96,7 @@ class AssetTest extends ModelTestCase
         $this->testAsset = TestHelper::createImageAsset('', null, true, 'assets/images/image1.jpg');
         $this->testAsset->save($versionNote);
         $versions = $this->testAsset->getVersions();
-        $this->assertEquals(end($versions)->getNote(), $versionNote['versionNote']);
+        $this->assertEquals($this->testAsset->getLatestVersion(null, true)->getNote(), $versionNote['versionNote']);
     }
 
     public function testThumbnails()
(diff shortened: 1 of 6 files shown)


tests/Model/Asset/AssetTest.php
```diff
@@ -88,6 +88,17 @@ public function testCRUD()
...
     }
 
+    /**
+     * Verifies that asset PHP API version note is saved
+     */
+    public function testSavingVersionNotes(){
+        $versionNote = ["versionNote" => "a new version of this asset"];
+        $this->testAsset = TestHelper::createImageAsset('', null, true, 'assets/images/image1.jpg');
+        $this->testAsset->save($versionNote);
+        $versions = $this->testAsset->getVersions();
```
dvesh3: why not use `getLatestVersion()` here?

kingjia90: It was not working, not sure if it is a bug or intentional

https://github.com/pimcore/pimcore/blob/66afb7f27934abf7498663e620a11aeefdd6b282/models/Element/Traits/VersionDaoTrait.php#L33-L35
It used to have a `Sforce` parameter but somehow got removed in https://github.com/pimcore/pimcore/pull/8838 even though in the comments is still mentioning it
https://github.com/pimcore/pimcore/blob/66afb7f27934abf7498663e620a11aeefdd6b282/models/Element/Traits/VersionDaoTrait.php#L27

without the former `force` to also consider the current version, and because the condition `'date > :mdate OR versionCount > :versionCount`, it cannot get the current version (it should be greater or equal), and works only for scheduled version?

dvesh3: yes, I think we can add `$force` param again, which should do `'date >= :mdate` so that we also get the published version when calling the `getLatestVersion()` method.

dvesh3:
@kingjia90 please rebase to fix the tests

PR #12657, opened 2022-07-15
Added a method to return publicly available full url path of the thumbnail

### LINKED ISSUE #11692: [Improvement] Provide public available getFrontendPath method for assets and (image) thumbnails
### Improvement description

to have one method to be used when generating public frontend urls that consider eventually configured prefixes and consistently adds a domain to the path

aashan10:
Resolves #11692

### DIFF
--- a/models/Asset/Thumbnail/ImageThumbnailTrait.php
+++ b/models/Asset/Thumbnail/ImageThumbnailTrait.php
@@ -341,16 +341,12 @@ trait ImageThumbnailTrait
      */
     public function getFrontendPath(): string
     {
-        $pathReference = $this->getPathReference();
-        $type = $pathReference['type'] ?? null;
-
-        if ($type === 'data-uri') {
-            return $pathReference['src'];
+        $path = $this->getPath();
+        if(!\preg_match('@^(https?|data):@', $path)) {
+            $path = \Pimcore\Tool::getHostUrl() . $path;
         }
 
-        $path = $this->convertToWebPath($pathReference);
-
-        return Tool::getHostUrl() . $path;
+        return $path;
     }
 
     /**
--- a/models/Asset/Thumbnail/ImageThumbnailTrait.php
+++ b/models/Asset/Thumbnail/ImageThumbnailTrait.php
@@ -337,12 +337,20 @@ trait ImageThumbnailTrait
     }
 
     /**
-     * @param bool $deferred
      * @return string
      */
-    public function getFrontendPath(bool $deferred = true): string
+    public function getFrontendPath(): string
     {
-        return Tool::getHostUrl() . $this->getPath($deferred);
+        $pathReference = $this->getPathReference();
+        $type = $pathReference['type'] ?? null;
+
+        if ($type === 'data-uri') {
+            return $pathReference['src'];
+        }
+
+        $path = $this->convertToWebPath($pathReference);
+
+        return Tool::getHostUrl() . $path;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -336,6 +336,15 @@ protected function convertToWebPath(array $pathReference): ?string
         return $path;
     }
 
+    /**
+     * @param bool $deferred
+     * @return string
+     */
+    public function getFrontendPath(bool $deferred = true): string
+    {
+        return Tool::getHostUrl() . $this->getPath($deferred);
```
brusch: This is not entirely right, what about remote object storages such as S3? 
Those are handled here: https://github.com/pimcore/pimcore/blob/10.x/models/Asset/Thumbnail/ImageThumbnailTrait.php#L314

models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -336,6 +336,23 @@ protected function convertToWebPath(array $pathReference): ?string
...
+        $type = $pathReference['type'] ?? null;
+
+        if ($type === 'data-uri') {
+            return $pathReference['src'];
+        }
+
+        $path = $this->convertToWebPath($pathReference);
+
+        return Tool::getHostUrl() . $path;
+    }
```
jdreesen: Why does this, unlike [`getPath()`](https://github.com/pimcore/pimcore/blob/10.x/models/Asset/Document/ImageThumbnail.php#L64-L77) (or [this](https://github.com/pimcore/pimcore/blob/10.x/models/Asset/Video/ImageThumbnail.php#L70-L83), or [this](https://github.com/pimcore/pimcore/blob/10.x/models/Asset/Image/Thumbnail.php#L57-L91)), not dispatch an event?

brusch (changes requested):
This is still not right, see my last comment, because the return value of `convertToWebPath()` can already include an absolute path, e.g. when using `frontend_prefixes` config.

models/Asset/Thumbnail/ImageThumbnailTrait.php
brusch:
```suggestion
    /**
     * @return string
     */
    public function getFrontendPath(): string
    {
        $path = $this->getPath();
        if(!\preg_match('@^(https?|data):@', $path)) {
            $path = \Pimcore\Tool::getHostUrl() . $path;
        }

        return $path;
    }
```

This should do the trick, imho.

brusch (changes requested):

models/Asset/Thumbnail/ImageThumbnailTrait.php
brusch: If we change to the suggested approach, an additional event shouldn't be necessary, as `getPath()` gets called anyway 😊

dasraab:
i think that is still not correct because you are not able to get a frontend path (with frontend_prefixes) when calling getFrontendPath() in a console command.

[see ImageThumbnailTrait.php#L320](https://github.com/pimcore/pimcore/blob/b234d787fcc7908d3855c729a0f24ed2368788f2/models/Asset/Thumbnail/ImageThumbnailTrait.php#L320)

maybe you should also consider to use symfonys built in router-configuration to set the default uri for the application:
[Generating URLs in Commands](https://symfony.com/doc/5.3/routing.html#generating-urls-in-commands)

cancan101:
How does `getFrontendPath` compare to `getFrontendFullPath`? I can't tell from the docs what the difference is.

PR #12658, opened 2022-07-15
Fix link to Apache configuration file in the docs

jdreesen:
The Apache configuration was removed from the skeleton and is now in the docs. The docs for older version now point to a non-existing file in the skeleton. This fixes it.

See: pimcore/skeleton#105

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/01_Apache_Configuration.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/01_Apache_Configuration.md
@@ -4,7 +4,7 @@
 Make sure `Allowoverride All` is set for the `DocumentRoot`, which enables `.htaccess` support. 
 
 All the necessary rewrite rules, which are needed for Pimcore to work, are defined in the `.htaccess` of the install package, 
-see [https://github.com/pimcore/skeleton/blob/10.x/public/.htaccess](https://github.com/pimcore/skeleton/blob/10.x/public/.htaccess). 
+see [https://github.com/pimcore/skeleton/blob/10.1/public/.htaccess](https://github.com/pimcore/skeleton/blob/10.x/public/.htaccess). 
 
 #### Example 
 ```
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/01_Apache_Configuration.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/01_Apache_Configuration.md
@@ -4,7 +4,7 @@
 Make sure `Allowoverride All` is set for the `DocumentRoot`, which enables `.htaccess` support. 
 
 All the necessary rewrite rules, which are needed for Pimcore to work, are defined in the `.htaccess` of the install package, 
-see [https://github.com/pimcore/skeleton/blob/10.1/public/.htaccess](https://github.com/pimcore/skeleton/blob/10.x/public/.htaccess). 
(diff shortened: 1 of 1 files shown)


jdreesen:
Should this be fixed in older versions of the docs? It points to the wrong location in the docs for `10.0`, `10.1`, `10.2`, `10.3`, too.

jdreesen:
There are several other links in the docs that point to https://github.com/pimcore/skeleton/blob/10.x. Should I change them all to https://github.com/pimcore/skeleton/blob/10.1?

doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/01_Apache_Configuration.md
```diff
@@ -4,7 +4,7 @@
 Make sure `Allowoverride All` is set for the `DocumentRoot`, which enables `.htaccess` support. 
 
 All the necessary rewrite rules, which are needed for Pimcore to work, are defined in the `.htaccess` of the install package, 
-see [https://github.com/pimcore/skeleton/blob/10.x/public/.htaccess](https://github.com/pimcore/skeleton/blob/10.x/public/.htaccess). 
+see [https://github.com/pimcore/skeleton/blob/10.1/public/.htaccess](https://github.com/pimcore/skeleton/blob/10.x/public/.htaccess).
```
dvesh3:
```suggestion
see [https://github.com/pimcore/skeleton/blob/10.1/public/.htaccess](https://github.com/pimcore/skeleton/blob/10.1/public/.htaccess). 
```

dvesh3:
> There are several other links in the docs that point to https://github.com/pimcore/skeleton/blob/10.x. Should I change them all to https://github.com/pimcore/skeleton/blob/10.1?

I would just fix the links which are broken as not every file is removed on 10.x

doc/Development_Documentation/23_Installation_and_Upgrade/03_System_Setup_and_Hosting/01_Apache_Configuration.md
jdreesen: Oops 🙈

dvesh3: I'm just thinking if we should really point to specific release branch as this tends to be removed with next minor release e.g. 10.2.

dvesh3: @jdreesen please see 👆

dvesh3: we can 🍒 pick [this ](https://github.com/pimcore/pimcore/pull/12619/files#diff-b9c0eecf9bf7df32f4c2b79c1d207fc75acb51575abc87972ef0139e9684e63b)on 10.4 branch

brusch: > I'm just thinking if we should really point to specific release branch as this tends to be removed with next minor release e.g. 10.2.

Fair point, maybe we should just include the `.htaccess` in the docs for 10.4 as well? 
Older branches I won't fix, as they are unsupported anyway and just there for reference.

jdreesen: Oh, well... Then maybe cherry-picking the mentioned commit may be the best solution.

brusch:
🍒-picked the docs into `10.4`: 
https://github.com/pimcore/pimcore/commit/5437ac186db0af1e2f4e14b3ceb7d91b9a49035e

PR #12662, opened 2022-07-15
[Bug]: Fix missing $force param that includes published in GetLatestVersion() 

kingjia90:
Resolves / related [12652](https://github.com/pimcore/pimcore/pull/12652/files#r922016738) (comment)

## Additional info  
Mhhh.. It's a little borderline from being an Improvement 🤔 fixing the code so that the comment about `$force` is correct.
Renamed `$force` to `$includingPublished`

### DIFF
--- a/models/Element/Traits/VersionDaoTrait.php
+++ b/models/Element/Traits/VersionDaoTrait.php
@@ -24,15 +24,17 @@ use Pimcore\Model\Version;
 trait VersionDaoTrait
 {
     /**
-     * Get latest available version, using $force always returns a version no matter if it is the same as the published one
+     * Get latest available version, using $includeCurrent to also consider the published one
      *
      * @param int|null $userId
+     * @param bool $includeCurrent
      *
      * @return Version|null
      */
-    public function getLatestVersion($userId = null)
+    public function getLatestVersion($userId = null, $includeCurrent = false)
     {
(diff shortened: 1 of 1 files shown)


models/Element/Traits/VersionDaoTrait.php
```diff
@@ -24,16 +24,16 @@
 trait VersionDaoTrait
 {
     /**
-     * Get latest available version, using $includeCurrent to also consider the published one
+     * Get latest available version, using $includingPublished to also consider the published one
      *
      * @param int|null $userId
      * @param bool $includeCurrent
```
NiklasBr:
```suggestion
     * @param bool $includingPublished
```

models/Element/Traits/VersionDaoTrait.php
```diff
@@ -24,15 +24,17 @@
...
     /**
-     * Get latest available version, using $force always returns a version no matter if it is the same as the published one
+     * Get latest available version, using $includingPublished to also consider the published one
      *
      * @param int|null $userId
+     * @param bool $includingPublished
      *
      * @return Version|null
      */
-    public function getLatestVersion($userId = null)
```
cancan101, line 33: when this signature was updated, should this line have been updated as well: https://github.com/pimcore/pimcore/blob/e04935c09918a762572e6e17e0e8461c7b37ed33/models/DataObject/Concrete.php#L36

blankse: @cancan101 I created a PR: https://github.com/pimcore/pimcore/pull/17059

PR #12668, opened 2022-07-16
Only generate webv2print PDF file $path when required

NiklasBr:

Tiny optimisation. `$path` is only used in the return file path condition.


brusch:
Thank you very much! 👍

PR #12676, opened 2022-07-18
[Docs] Use map for static page config in nginx

jdreesen:
See: pimcore/skeleton#100
Followup of #9447 and #11211


dvesh3:
@jdreesen can we also add it to the 10.5 upgrade notes to update the NGINX config?

jdreesen:
Well, the old one should still work, but I can add it to the upgrade notes if you want.

dvesh3:
> Well, the old one should still work, but I can add it to the upgrade notes if you want.

yes, I mean adding a recommendation 😉

dvesh3:
@jdreesen could you please update it? thanks!

jdreesen:
I added something, is that okay?

dvesh3:
> I added something, is that okay?

yes, looks good. thanks! 🙏

PR #12680, opened 2022-07-19
Remove locks for folders in Asset, Data Object and Document controllers.

### LINKED ISSUE #12673: [Improvement]: Disable edit lock for folders in Asset management
### Improvement description

At the moment when a backend user views an Assets folder, an edit lock is automatically applied to it. This will then show a warning to anyone other user that tries to access this folder.

I suggest changing AssetController->getDataByIdAction() to exclude folders from the automatic edit lock.

I am happy to make a PR with this change.

Andreas-Koepke:
This pull request resolves issue #12673 - removes the edit lock for folders in the Asset, Data Object and Document controllers.


mcop1:
Thank you!

PR #12681, opened 2022-07-19
Disabled dragging and dropping assets from the Preview and List

### LINKED ISSUE #12554: [Improvement]: Drag & Drop in asset list not working
### Expected behavior

In the tree-view of the assets it is possible to drag&drop one or more elements and move them fron one folder to another.
The same should work if moving an asset from the main asset view (preview or list view) to the three view.
After selecting multiple assets in the list-view, the expected behavior should be, that all selected elements are moved to the new folder when dropping on the tree-list.

Testet in the public demo with Version 10.4.4

### Actual behavior

When dragging an asset from the asset-preview or list-view to the tree view, an empty message window opens and nothing happens. The window disappears after a few seconds.

image

### Steps to reproduce

1. open the "Assets"-view
2. drag & drop an asset from the preview- or list-view to the tree-list on the side.

image

aryaantony92:
Resolves #12554


dvesh3:
@aryaantony92 please merge 10.4 branch to fix the failing tests.

PR #12682, opened 2022-07-19
[Admin] Fix searching objects for specific class

dvesh3:
Resolves #

## Additional info  
1. Go to Search -> DataObject in Admin
2. Select a class type
3. Click Search => results are not class specific


kingjia90:
Closing in favor of https://github.com/pimcore/customer-data-framework/pull/307

The changes in this PR would cause problems because it would be changing something intentional: if you have 10 `objects` with specific class eg.`CustomerSegment` and 20 `folders`, selecting "All Types" as sub type filter (object/folder/variant), it should show 30 entries.  
That removed line is to making this happen, as a sort of `OR` conditon, in other words `((subtype = object AND class = CustomerSegment) OR subtype = folder)`

Issue #12656, opened 2022-07-14
[Bug]: LinkGenerators to not work with path/url twig extension


fashxp:
### Expected behavior

thinks like `{{ url(car) }}` work in template when car is instance of a dataobject with link generator defined ... as stated here in [docs](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Class_Settings/Link_Generator.html#page_path-url)

### Actual behavior

![image](https://user-images.githubusercontent.com/8792145/179012808-56bcb342-70ce-4f34-9231-406ee3130502.png)

### Steps to reproduce

create a template and call  `{{ url(car) }}`

blankse:
Maybe we should update the docs. I changed it here:
https://github.com/pimcore/pimcore/pull/9694

`{{ url('pimcore_element', {'element': car}) }}` should work

dvesh3:
Fixed by https://github.com/pimcore/pimcore/commit/31b495fc62e0c243c7f126a37c60a9f1c92311cf

PR #12683, opened 2022-07-19
Updated documentation for usage of url / path with link generators 

aashan10:
## Changes in this pull request  
Resolves #12656

PR #12687, opened 2022-07-19
[Doc] LinkGenerator: Fix path/url example

blankse:
See https://github.com/pimcore/pimcore/issues/12656


blankse:
Duplicate of https://github.com/pimcore/pimcore/pull/12683
I was to slow :D

PR #12689, opened 2022-07-19
[10.4] Fix 'Only direct children' checkbox in variants tab

### LINKED ISSUE #7726: Variants tab shows only direct childrens
# Bug Report

If I create a typical object tree with variants (2 levels):

grafik

Then I open the root product and navigate to the variants tab, it's only possible to show the direct child variants:

grafik

### Expected behavior

The checkbox "just direct children" implies that all variants below the this root product should be listet by default and can be restricted to just direct childrens.

### Actual behavior

Only direct child variants will be listed.

### Steps to reproduce

Create a simple Object variant structure and try to display all the variants.

Pimcore 6.8.4

DuckThom:
Resolves #7726

## Additional info  

`only_direct_children` was being hard-coded to 'true' breaking the checkbox in the Variants tab of a product


dvesh3:
@DuckThom thanks!

Issue #12666, opened 2022-07-15
[Bug]: Chrome Headless Web2Print omits backgrounds even when explicitly told to keep them


NiklasBr:
### Expected behavior

Background colours and images should be visible with these settings:
```json
{
  "displayHeaderFooter": false,
  "format": "A4",
  "height": "297mm",
  "landscape": false,
  "margin": {
    "top": "0mm",
    "bottom": "0mm"
  },
  "omitBackground": false,
  "preferCSSPageSize": true,
  "printBackground": true,
  "width": "210mm"
}
```

With this preview:
![image](https://user-images.githubusercontent.com/279826/179234421-4c1937ee-1aec-4553-94af-ec8c9f80223c.png)

### Actual behavior

Positioning works, but no backgrounds visible, not `background-color` nor `background-image`, nor `fill` in SVG:s:

![image](https://user-images.githubusercontent.com/279826/179234688-36fc4f9b-b44a-4893-acea-afa7ef6c0881.png)

### Steps to reproduce

Example HTML to send in a `printpage`:
```html
<!DOCTYPE html>
<html lang="nb">

<head>
	<meta charset="UTF-8">
	<title>Print</title>
	<link rel="stylesheet" type="text/css" media="all" href="https://localhost/static/ateles.1fcac.css"> </head>

<body class="pdf">
	<main class="page grid-wrapper custom-background">
		<div class="decoration grid-item-3-4-5-2 fill-primary"></div>
		<div class="decoration grid-item-1-4-7-2 fill-primary"></div>
		<div class="decoration grid-item-1-2-9-2 fill-primary"></div>
		<div class="decoration grid-item-3-6-1-4 fill-secondary"></div>
		<div class="logo grid-item-1-2-1-4 fill-primary logo-color">
			<svg viewBox="0 0 339.03 357.51" xmlns="http://www.w3.org/2000/svg" preserveAspectRatio="xMinYMin meet">
				<path d="M0,.14H10.46l21,47.78L52.11.14H62.63V73.48H51.51V27.75h-.25L35.92,63H26.66L11.32,27.75h-.2V73.48H0Z" />
				<path d="M148.58,0h46.83V10.5H159.69V31.43H190.2v9.86H159.69V62.82h35.72V73.33H148.58Z" />
				<path d="M292.2,0h11.12V62.82h35.71V73.33H292.2Z" />
[...]

PR #12691, opened 2022-07-19
Add visual test for background in testWeb2print.html.twig

NiklasBr:
Could be of help with #12666

image


kingjia90:
Thank you! 
The PR is fine and It works, which doesn't explain why #12666 happens, but at least narrows down to the processor.

~IMHO this PR looks like should be considered as an improvement, in case, please could you rebase with 10.x?~ 
changed to Task

PR #12693, opened 2022-07-19
[Task]: Change broken link to documentation of headless chrome options

kingjia90:
The link position is here:

image

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/web2print.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/web2print.js
@@ -217,7 +217,7 @@ pimcore.settings.web2print = Class.create({
                         value: t('web2print_headlesschrome_puppeteer_documentation'),
                         autoEl:{
                             tag: 'a',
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/settings/web2print.js
```diff
@@ -217,7 +217,7 @@ pimcore.settings.web2print = Class.create({
                         value: t('web2print_headlesschrome_puppeteer_documentation'),
                         autoEl:{
                             tag: 'a',
-                            href: "https://pptr.dev/#?product=Puppeteer&version=v5.2.1&show=api-pagepdfoptions",
+                            href: "https://pptr.dev/api/puppeteer.pdfoptions",
```
kingjia90:
```suggestion
                            href: "https://github.com/spiritix/php-chrome-html2pdf#options",
```
Leaving here as alternative option, feel free to commit suggestion

dvesh3: I would go for spiritix options since it adds more options like `pageWaitFor`, `cookies` and so on..

PR #12697, opened 2022-07-20
Fix: Missing Video Thumbnails won't be generated

blankse:
We have the problem that video thumbnails are missing but in the thumbnail cache. I don't know how the thumbnails are deleted.

The getHtml5Code() method will generate a auto thumbnail name. But when this is in the cache no config will stored in the TmpStore.

The PublicServiceController can so not load the thumbnail config. I would delete the thumbnail from cache in this case.

Maybe related: https://github.com/pimcore/pimcore/discussions/12549


dvesh3:
@blankse sorry, I'm not able to reproduce the issue. If i delete generated thumbnail files then these are re-generated on next load.

blankse:
@dvesh3 Do you use the pimcore_video Editable? Without "imagethumbnail" option?

blankse:
@dvesh3 In the backend it works. Its only in frontend.

dvesh3:
yes, I can reproduce the problem on frontend and your changes fixes the issue.

dvesh3:
@blankse thanks!

PR #12701, opened 2022-07-20
Avoid multiple Thumbnail::getPath() calls in Video::getHtml5Code()

blankse:
I splitted this from #12697 

The _toString() (and so the getPath()) method would call very often (Line 827, 828, 829, 869, 872).

The doc type of the param is also wrong. It can only be a Thumbnail object. See line 380/400.

### DIFF
--- a/models/Document/Editable/Video.php
+++ b/models/Document/Editable/Video.php
@@ -768,7 +768,7 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
 
     /**
      * @param array $urls
-     * @param string|null $thumbnail
+     * @param Asset\Image\Thumbnail|Asset\Video\ImageThumbnail|null $thumbnail
      *
      * @return string
      */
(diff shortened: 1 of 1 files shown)


models/Document/Editable/Video.php
```diff
@@ -824,9 +824,10 @@ private function getHtml5Code($urls = [], $thumbnail = null)
                 $jsonLd['contentUrl'] = Tool::getHostUrl() . $urls['mp4'];
             }
 
-            $jsonLd['thumbnailUrl'] = (string)$thumbnail;
-            if (!preg_match('@https?://@', (string)$thumbnail)) {
-                $jsonLd['thumbnailUrl'] = Tool::getHostUrl() . $thumbnail;
+            $thumbnailUrl = (string)$thumbnail;
+            $jsonLd['thumbnailUrl'] = $thumbnailUrl;
```
dvesh3, line 828: Should we call newly introduced method `getFrontendPath` here? please see https://github.com/pimcore/pimcore/pull/12657/files#diff-b48489e5be12fb4a4ec4f716c03e4690a799c39a87430cbb1a435d7badebf97c

dvesh3: Nevermind, this was introduced on 10.x

brusch: @dvesh3 don't think this is necessary, it just makes the path longer but doesn't improve anything

PR #12704, opened 2022-07-20
[Task]: Followup app logger flysystem

### LINKED ISSUE #12698: [Improvement]: Add a fallback for application logs that are not stored in flysystem
### Improvement description

Follow up to #12460
First check if the requested application log is available via flysystem, if not, then check the old local path.

kingjia90:
Resolves #12698
Related https://github.com/pimcore/pimcore/pull/12460 , https://github.com/pimcore/data-hub-productsup/issues/32

~Flysystem bumped to version 3
https://github.com/thephpleague/flysystem/releases/tag/3.0.0~


kingjia90:
mhh, test are failing because `directoryExists` is introduced in 3.0 while we are on 2.0
image

jdreesen:
I think you should upgrade. Afaik the changes are small and v2 isn't supported anymore or support ends soon.

composer.json
```diff
@@ -159,7 +159,7 @@
     "onnov/detect-encoding": "^2.0",
     "knplabs/knp-paginator-bundle": "^5.4.2",
     "masterminds/html5" : "^2.7",
-    "league/flysystem-bundle" : "^2.0",
+    "league/flysystem-bundle" : "^3.0",
```
blankse:
```suggestion
    "league/flysystem-bundle" : "^2.0",
    "league/flysystem": "^3.0",
```

kingjia90: ahh.. Thank you very much

kingjia90:
Unfortunately, it seems that Flysystem cannot be bumped right now as it will conflict with these 2 bundle requirements
https://github.com/pimcore/data-hub-file-export/blob/1.x/composer.json#L15
https://github.com/pimcore/data-importer/blob/88812a1fdc912eb9f456768cfbe283ed0a26e19a/composer.json#L19

brusch:
`pimcore/pimcore` is  only using `league/flysystem-bundle` which requires `league/flysystem: ^2.0|^3.0` so shouldn't be a Pimcore topic I guess, right? 
I think the bundles should just kick the requirement completely or also use the bundle and not directly `league/flysystem`

jdreesen:
Well, if we use methods that only exist in v3, we should require v3 of the lib.

brusch:
That's true - additionally to the bundle. 
Afterwards we should "fix" the bundles (by talking away the requirement from `composer.json`, then it should be fine.

lib/Maintenance/Tasks/LogArchiveTask.php
```diff
@@ -98,7 +98,20 @@ public function execute()
...
+                    $storage->delete($filePath);
+                } else {
+
+                    // Fallback, if is not found and deleted in the flysystem, tries to delete from local
+                    $fileRealPath = realpath($filePath);
+                    if (str_starts_with(realpath($fileRealPath), PIMCORE_LOG_FILEOBJECT_DIRECTORY)) {
+                        @unlink($fileRealPath);
+                    }
+
+                }
```
mcop1:
```suggestion
                if($filePath !== null) {
                    if ($storage->fileExists($filePath)) {
                        $storage->delete($filePath);
                    } else {
    
                        // Fallback, if is not found and deleted in the flysystem, tries to delete from local
                        $fileRealPath = realpath($filePath);
                        if (str_starts_with(realpath($fileRealPath), PIMCORE_LOG_FILEOBJECT_DIRECTORY)) {
                            @unlink($fileRealPath);
                        }  
                    }
                 }
```

`$storage->fileExists` will fail if `$filePath` is null

BlackbitDevs:
Wouldn't it be better to create a migration which takes the `FileObject` files which are currently linked in `application_logs` and send them to the flysystem storage? Then we would not have to keep both storage ways everywhere in the code.

dvesh3:
> Wouldn't it be better to create a migration which takes the `FileObject` files which are currently linked in `application_logs` and send them to the flysystem storage? Then we would not have to keep both storage ways everywhere in the code.

@BlackbitNeueMedien you could do this with `pimcore:migrate:storage` command by defining source & target locations.. right? https://pimcore.com/docs/pimcore/current/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/File_Storage_Setup.html#page_Storage-Migration

BlackbitDevs:
Probably yes. But you prefer to keep the flysystem code and the local file storage code everywhere to support both ways? 
In former times I often heard the argument to prevent the code from getting too complex.

Or in a more constuctive way: Why not use `new LocalFilesystemAdapter(PIMCORE_LOG_FILEOBJECT_DIRECTORY)` when `pimcore.application_logs.storage` is not set?

dvesh3:
The fallback is there to support bundles which still require `pimcore/pimcore: ^6.9 || ^10.0` otherwise it will be counted as a BC break.

BlackbitDevs:
Ah ok. The problem is that in old approach the `fileObject` column in `application_logs` contains the `PIMCORE_LOG_FILEOBJECT_DIRECTORY` path. With `new LocalFilesystemAdapter(PIMCORE_LOG_FILEOBJECT_DIRECTORY)` we would have to strip this everywhere. Ok, I am fine with this ;-)

BlackbitDevs:
@dvesh3, @kingjia90 THere is another BC break concerning `FileObject`s: In https://github.com/pimcore/pimcore/blob/b675696e8c3dff29e51d97e67fc79d3003c9ca3a/bundles/CoreBundle/Resources/config/pimcore/flysystem.yaml#L72-L77 you hardcode the default root directory for the `pimcore.application_log.storage` to `<Pimcore root>/var/application-logger`. This works as long as you let `PIMCORE_LOG_FILEOBJECT_DIRECTORY` and `PIMCORE_PRIVATE_VAR` untouched but it won't work if one of those points to a different directory.

kingjia90:
> @dvesh3, @kingjia90 THere is another BC break concerning `FileObject`s: In
> 
> https://github.com/pimcore/pimcore/blob/b675696e8c3dff29e51d97e67fc79d3003c9ca3a/bundles/CoreBundle/Resources/config/pimcore/flysystem.yaml#L72-L77
> 
> you hardcode the default root directory for the `pimcore.application_log.storage` to `<Pimcore root>/var/application-logger`. This works as long as you let `PIMCORE_LOG_FILEOBJECT_DIRECTORY` and `PIMCORE_PRIVATE_VAR` untouched but it won't work if one of those points to a different directory.

That's true, i've overlooked on that... 
What about something like this as default from 10.5.1 until 11? Would it be there any drawbacks?
```
parameters:
    log.fileobject: !php/const PIMCORE_LOG_FILEOBJECT_DIRECTORY
```
```
            options:
                directory: "%kernel.project_dir%%log.fileobject%"
```

BlackbitDevs:
@kingjia90 Hmm, is this even possible? Because `PIMCORE_LOG_FILEOBJECT_DIRECTORY` gets set dynamically during runtime. In https://symfony.com/blog/new-in-symfony-3-2-php-constants-in-yaml-files I only see usage of predefined constants (PHP or class constants).
If not, then we probably have to create a compiler pass for this... (like in https://stackoverflow.com/questions/31859005/overriding-parameters-yml-values-dynamically-after-running-composer-install-in-s)

brusch:
@BlackbitNeueMedien @kingjia90 we should create a dedicated ticket for that. 

@BlackbitNeueMedien `!php/const PIMCORE_LOG_FILEOBJECT_DIRECTORY` could work, because `PIMCORE_LOG_FILEOBJECT_DIRECTORY` should be set in startup.php anyway, which is loaded before the container gets build, right?

BlackbitDevs:
> PIMCORE_LOG_FILEOBJECT_DIRECTORY should be set in startup.php anyway, which is loaded before the container gets build, right?

You are right. Would be nice if you or @kingjia90 can try it out.

PR #12708, opened 2022-07-20
Sort Predefined Document Types

NiklasBr:

Sorts by priority first and name second


NiklasBr:
Reason for the JS change is this:

image

*1* is first priority
*10* is last priority as shown in the select options

It also makes sense, which comes first when sorting naturally. It is also [how others](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Weight.php/9.4.x) [do it](https://www.man7.org/linux/man-pages/man2/getpriority.2.html).

mattamon:
Hey @NiklasBr,

Thank you for your contribution.
It makes totally sense to sort it that way. I also checked the priority for the doc types and it appears the priority is never used anyways. 

Atm the priority is a select with 10 entries. Once you changed it, you cannot change it back to 0. 
I think it would make more sense if it was just a number input? Why would we restrict it to 10 anyways? Wdyt?
 
There is also an api involved that can read the docTypes via this route: `pimcore_admin_document_document_doctypesget`
It would make sense to change the sorting there too right?

NiklasBr:
Sounds good, will see what I can do

models/Document/DocType/Listing/Dao.php
```diff
@@ -35,6 +36,9 @@ public function loadList(): array
         }
         if ($this->model->getOrder()) {
             usort($docTypes, $this->model->getOrder());
+        } else {
+            // Default sort if nothing else has been set
+            usort($docTypes, [self::class, 'sortByPriority']);
```
jdreesen: PHP 8.1 :tada: 

```suggestion
            usort($docTypes, self::sortByPriority(...)]);
```

NiklasBr: Nice 👍

bundles/AdminBundle/public/js/pimcore/settings/docTypes.js
```diff
@@ -149,13 +149,15 @@ pimcore.settings.document.doctypes = Class.create({
...
             {
                 text: t("priority"),
-                flex: 50,
+                flex: 30,
                 sortable: true,
                 dataIndex: 'priority',
-                editor: new Ext.form.ComboBox({
-                    store: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
+                editor: new Ext.form.NumberField({
+                    allowBlank: true,
```
mattamon: If we allow "blank" or null here we have to change setPriority method to allow null. We should then also remove the int cast. If we do not allow null values we could still remove the int cast though :D 
Currently it looks like the following:
```php
   public function setPriority(int $priority): static
    {
        $this->priority = (int) $priority;

        return $this;
    }
```

mattamon (changes requested):
I noticed that I cannot update Document Types either with an empty Controller or an empty template due to the param types of the setter methods not being nullable. 

```php
   public function setController(string $controller): static
    {
        $this->controller = $controller;

        return $this;
    }
```
Maybe we can change this in this PR as well so it works properly again?

bundles/AdminBundle/public/js/pimcore/settings/docTypes.js
```diff
@@ -149,13 +149,14 @@ pimcore.settings.document.doctypes = Class.create({
...
+                flex: 30,
                 sortable: true,
                 dataIndex: 'priority',
-                editor: new Ext.form.ComboBox({
-                    store: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
+                editor: new Ext.form.NumberField({
                     mode: "local",
-                    editable: false,
+                    editable: true,
+                    minValue: 1,
```
mattamon: Sorry about this one, but I mentioned it in the review before :) 
If the min value is 1 and the initial value is 0, you cannot change back to the initial value once you set the priority. So I would keep it at 0 tbh.
```suggestion
                    minValue: 0,
```

mattamon (changes requested):
I will open a separate Issue for the Controller and the Template not being nullable. :)

mattamon (approved):

Issue #12719, opened 2022-07-21
Pass objectBrickLayout definition to CLASS_OBJECTBRICK_LIST_PRE_SEND_DATA event 


gjorgic:
### Improvement description

I wondering, why we do not pass object brick layout definition data into event object [here](https://github.com/pimcore/pimcore/blob/11.x/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php#L1414), and give possibility for developer to modify this layout information

Same approach is used on `..../object/get` route [here](https://github.com/pimcore/pimcore/blob/11.x/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L539)

fashxp:
fixed via https://github.com/pimcore/pimcore/pull/12723

PR #12723, opened 2022-07-21
Object Brick layout improvement; Relates #12719

gjorgic:
This will give ability to override brick layout definition
#12719


gjorgic:
Branch 10.x not exist, so i target 10.5

fashxp:
correct target would be 11.x, I changed it

robertSt7:
@gjorgic Thanks

PR #12727, opened 2022-07-21
Fix: Document\DocType::$staticGeneratorEnabled should be bool

### LINKED ISSUE #10819: [Bug]: Document\DocType::$staticGeneratorEnabled should be `bool`
### Expected behavior

Document\DocType::$staticGeneratorEnabled is boolean. 

### Actual behavior

Document\DocType::$staticGeneratorEnabled is integer

### Steps to reproduce

just needs to be fixed.

blankse:
Fix: https://github.com/pimcore/pimcore/issues/10819


aryaantony92:
@blankse Thank you for the PR. But, the field 'staticGeneratorEnabled' is nullable in db, so the default value will be 'null' for this field. Since the property type should match with the db column type, either we have to add a db migration or we should accept 'null' value also.

blankse:
@aryaantony92 Do you mean the `staticGeneratorEnabled` column at the `documents_page` table? This is not for the doc type. The page document attribute can be null. See https://github.com/pimcore/pimcore/blob/2646cc26d7aa7af9a069f11183c31306fe5de662/models/Document/PageSnippet.php#L95

aryaantony92:
> @aryaantony92 Do you mean the `staticGeneratorEnabled` column at the `documents_page` table? This is not for the doc type. The page document attribute can be null. See
> 
> https://github.com/pimcore/pimcore/blob/2646cc26d7aa7af9a069f11183c31306fe5de662/models/Document/PageSnippet.php#L95

Yes, you are right. But, I think the `staticGeneratorEnabled` property in `models/Document/DocType.php` also should accept null values. Because, in the settings_store table, there are some predefined document type configurations which have the  `staticGeneratorEnabled` value as null like `{"name":"Standard-Teaser",...."staticGeneratorEnabled":null}`

blankse:
@aryaantony92 The DocTypes in the settings_store Table can also be an integer staticGeneratorEnabled... Think we should migrate them anyway

blankse:
@aryaantony92 I added a migration

aryaantony92:
> @aryaantony92 I add a migration

Thanks 👍

aryaantony92:
@blankse But, I think still it should be nullable. Because, if the docTypes was saved in a symfony config (yaml file) instead of settings store, the value can be null there in case of existing entries.

blankse:
@aryaantony92 There also can be integer values. I added a BC layer so there is integer/null values still possible.

aryaantony92:
@blankse Thanks a lot!

bundles/CoreBundle/Migrations/Version20220809164000.php
```diff
@@ -0,0 +1,56 @@
...
+    public function up(Schema $schema): void
+    {
+        $db = Db::get();
+
+        $docTypeList = $db->fetchAllAssociative("SELECT id, data FROM `settings_store` WHERE scope = 'pimcore_document_types'");
+        foreach ($docTypeList as $docType) {
+            $dataArray = json_decode($docType['data'], true);
+            $dataArray['staticGeneratorEnabled'] = (bool) $dataArray['staticGeneratorEnabled'];
+            $docType['data'] = json_encode($dataArray);
+            $this->addSql('UPDATE `settings_store` SET data = :data WHERE id = :id', $docType);
```
mcop1, line 40: @blankse @aryaantony92  Seems like the update statement is missing the scope, right? 
`$this->addSql('UPDATE `settings_store` SET data = :data WHERE scope = 'pimcore_document_types' and id = :id', $docType);`

bundles/CoreBundle/Migrations/Version20220809164000.php
```diff
@@ -0,0 +1,56 @@
...
+    public function down(Schema $schema): void
+    {
+        $db = Db::get();
+
+        $docTypeList = $db->fetchAllAssociative("SELECT id, data FROM `settings_store` WHERE scope = 'pimcore_document_types'");
+        foreach ($docTypeList as $docType) {
+            $dataArray = json_decode($docType['data'], true);
+            $dataArray['staticGeneratorEnabled'] = (int) $dataArray['staticGeneratorEnabled'];
+            $docType['data'] = json_encode($dataArray);
+            $this->addSql('UPDATE `settings_store` SET data = :data WHERE id = :id', $docType);
```
mcop1, line 53: Same here

bundles/CoreBundle/Migrations/Version20220809164000.php
blankse: @mcop1 You are right. I created a followup PR: https://github.com/pimcore/pimcore/pull/12874

mcop1:
@blankse figured I could create a pr too, but you were a bit faster than me. I´ll close mine, thanks!

blankse:
@mcop1 Only few seconds 😄

PR #12733, opened 2022-07-21
Fix undefined allowedTypes in video editable

blankse:
After update from 10.4 to 10.5 we get following error:
image

Code in Areabrick:
```
$videoTag = $this->getDocumentEditable($info->getDocument(), 'video', 'video');
$videoAsset = $videoTag->getVideoAsset();
```

Regression from https://github.com/pimcore/pimcore/pull/12445


dvesh3 (approved):

dvesh3:
@blankse thanks for the fix 😃

PR #12734, opened 2022-07-21
Image Thumbnail: Fix undefined getName() method when no config is set

blankse:
After update from 10.4 to 10.5 we get following error:
image

Code:
```
{{ pimcore_image('image').frontend|raw }}
```

Regression from https://github.com/pimcore/pimcore/pull/12461


dvesh3:
@blankse thanks!

PR #9517, opened 2021-06-22
Fix: Avoid error logs with PPTX files and LibreOffice/Ghostscript

blankse:
- LibreOffice/Ghoscript available
- Upload a PPTX file
- Execute Maintenance Cronjob

Following logs in error output and in prod.log
```
08:58:06 ERROR     [pimcore] Couldn't load document /var/www/html/web/var/assets/test.pptx only PDF documents are currently supported
08:58:07 ERROR     [pimcore] Couldn't convert document to Text: /var/www/html/web/var/assets/test.pptx with the command: ''/usr/bin/soffice' '--headless' '--nologo' '--nofirststartwizard' '--norestore' '--convert-to' 'txt:Text' '--outdir' '/var/www/html/web/var/tmp' '/var/www/html/web/var/assets/test.pptx'' - now trying to get the text out of the PDF with ghostscript...
08:58:07 ERROR     [pimcore] Couldn't load document /var/www/html/web/var/assets/test.pptx only PDF documents are currently supported
```

Issue #12631, opened 2022-07-12
[Bug][Assets] Extracting text from assets does not work anymore


fashxp:
### Expected behavior

Text can be extracted from assets and added to the full-text search index after uploading/updating an asset

### Actual behavior

command fails because it tries to send a docx/pptx/etc to ghost script ... which does not work

most likely regression of that change: 
https://github.com/pimcore/pimcore/commit/2fabce94fa9a9a13a68a9013afe183ec302bdb24#diff-463c7c3a9e135e190743ca5a25b3eee85e94d97a4fb9db2fbe67ca964198f86bR180

### Steps to reproduce

just upload an document asset like docx or pptx and watch the logs of the symfony messenger queue

dvesh3:
Fixed by #12735 & #12770

PR #12735, opened 2022-07-21
extract text from pdfs via ghostscript

### LINKED ISSUE #6289: Converting PDF to text with libreoffice seems no more working (at least Debian)
Hello,

It seems conversion to txt using libre office do not more work, at least on Debian systems.

Code is started from here:
https://github.com/pimcore/pimcore/blob/ae6bf9bef67793dbb5a91ee91a37cbce614b1094/lib/Document/Adapter/LibreOffice.php#L189

Run manually on shell, using a pdf file (take a really light one in size please):
`/usr/bin/soffice --headless --nologo --nofirststartwizard --norestore --convert-to txt:Text mydoc.pdf`
Returns:
```
convert /var/www/html/git-excluded/mydoc.pdf -> /var/www/html/git-excluded/mydoc.txt using filter : "Text"
Error: Please verify input parameters... (SfxBaseModel::impl_store <file:///var/www/html/git-excluded/mydoc.txt> failed: 0x81a(Error Area:Io Class:Parameter Code:26))
```

If you take some more big PDF, the command line stay struggle and the cpu works hard... command **never finish** (i must do CTRL+C). **Here this is problematic as it can be CPU/memory hogger.**

I tried to search through internet (libreoffice github, etc.) and it seems text filter has been removed from some distribution:
Redhat: https://ask.libreoffice.org/en/question/82541/how-can-i-install-all-the-convert-to-filters-i-am-missing-txttext-but-have-pdf/
And debian (this is my test environment)

Going further, i upload a very simple PDF in the assets section, and just right after made a search with a string contained in the PDF... nothing is returned.
I made the test on the Pimcore online demo, and same. It uses Debain buster too, so same libreoffice version I imagine.
But from the Pimcore code if libreoffice fail it should use pdftotext (parent method ghostscript adapter).
**So there is this libreoffice pbl, but why it don't fall back to pdftotext after ? At the end no text is indexed from the PDF into the database.**

So it is not a bug with Pimcore, but the tool (libreoffice) and to me more exact the filter txt works no more.
This is a really great feature to be able to search through documents, and having it working is a really good feature for DAM I think.

I tried to update libre office using the backport on buster (really recent version of libreoffice), but same problem.

Feel free to ask any further questions to help.

EDIT: here the comment seems to say we should have a PDF ?? Is it a just a typo pbl https://github.com/pimcore/pimcore/blob/ae6bf9bef67793dbb5a91ee91a37cbce614b1094/lib/Document/Adapter/LibreOffice.php#L202 and should be `Couldn't convert document to TXT` i imagine.

Sorry if this post is a little confuse, i wanted to give as many insight as possible.

Thanks.

jheimbach:
Resolves #6289

## Additional info  
With the official Pimcore Docker image, the Text extraction from (large) PDF Assets is running into a timeout from libreoffice. 

```
pimcore.ERROR: Symfony\Component\Process\Exception\ProcessTimedOutException: The process "'/usr/bin/nice' '-n' '19' '/usr/bin/soffice' '--headless' '--nologo' '--nofirststartwizard' '--norestore' '--convert-to' 'txt:Text' '--outdir' '/var/www/pimcore/var/tmp' '/var/www/pimcore/var/tmp/temp-file-62d84df67e98b-a811104cd38790ea88e09bfa7869bc.pdf'" exceeded the timeout of 240 seconds.
```

In #6289 it is mentioned that ghostscript has precedence over libreoffice, but since [these changes](https://github.com/pimcore/pimcore/commit/f4c7b218) #9517, that is no longer true.

### DIFF
--- a/lib/Document/Adapter/LibreOffice.php
+++ b/lib/Document/Adapter/LibreOffice.php
@@ -186,7 +186,13 @@ class LibreOffice extends Ghostscript
             // for per page extraction we have to convert the document to PDF and extract the text via ghostscript
             return parent::getText($page, $asset);
         }
-        if (self::isFileTypeSupported($asset->getFilename())) {
+
+        // if asset is pdf extract via ghostscript
+        if (parent::isFileTypeSupported($asset->getFilename())) {
+            return parent::getText(null, $asset);
+        }
+
+        if ($this->isFileTypeSupported($asset->getFilename())) {
             // if we want to get the text of the whole document, we can use libreoffices text export feature
             $cmd = [self::getLibreOfficeCli(), '--headless', '--nologo', '--nofirststartwizard', '--norestore', '--convert-to', 'txt:Text', '--outdir',  PIMCORE_SYSTEM_TEMP_DIRECTORY, $asset->getLocalFile()];
             Console::addLowProcessPriority($cmd);


fashxp:
related https://github.com/pimcore/pimcore/issues/12631?

jheimbach:
I think #12631 is the same method but different bug. 
[This line](https://github.com/pimcore/pimcore/blob/11.x/lib/Document/Adapter/LibreOffice.php#L200) should work with the localTmpFile name but uses the Asset name.

I can add this fix to this PR or create a new one.

aashan10 (approved):

dvesh3:
@JHeimbach thanks!

PR #12267, opened 2022-05-30
[Bug]: CSV/XLSX Export temp file moved to Flysystem

kingjia90:
## Changes in this pull request  
Fixes #12174

PR #12736, opened 2022-07-21
downloadXlsxFileAction(): Use loadSpreadsheetFromString() to avoid the extra temp file

blankse:
Follow up to https://github.com/pimcore/pimcore/pull/12267


kingjia90:
Does this `phpoffice/phpspreadsheet` version bump need a line in the upgrade notes? 
As far as i see, this bundle is just used there for the export in xlsx

PR #12742, opened 2022-07-23
improvement: don't catch exceptions in ClassResolver

passioneight:
Since we had a little mix up in https://github.com/pimcore/pimcore/pull/11240, here's the new PR.

### DIFF
--- a/models/DataObject/ClassDefinition/Helper/ClassResolver.php
+++ b/models/DataObject/ClassDefinition/Helper/ClassResolver.php
@@ -37,20 +37,16 @@ abstract class ClassResolver
                 return self::$cache[$class];
             }
 
-            try {
-                if (strpos($class, '@') === 0) {
-                    $serviceName = substr($class, 1);
-                    $service = \Pimcore::getKernel()->getContainer()->get($serviceName);
-                } else {
-                    $service = new $class;
-                }
-
-                self::$cache[$class] = self::returnValidServiceOrNull($service, $validationCallback);
-
-                return self::$cache[$class];
-            } catch (\Throwable $e) {
-                Logger::error((string) $e);
+            if (strpos($class, '@') === 0) {
+                $serviceName = substr($class, 1);
+                $service = \Pimcore::getKernel()->getContainer()->get($serviceName);
+            } else {
+                $service = new $class;
             }
+
+            self::$cache[$class] = self::returnValidServiceOrNull($service, $validationCallback);
+
+            return self::$cache[$class];
         }
 
         return null;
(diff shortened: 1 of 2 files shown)


models/DataObject/ClassDefinition/Helper/ClassResolver.php
```diff
@@ -37,20 +37,16 @@ protected static function resolve($class, callable $validationCallback = null)
...
             }
 
-            try {
-                if (strpos($class, '@') === 0) {
-                    $serviceName = substr($class, 1);
-                    $service = \Pimcore::getKernel()->getContainer()->get($serviceName);
-                } else {
-                    $service = new $class;
-                }
+            if (strpos($class, '@') === 0) {
```
jdreesen:
```suggestion
            if (str_starts_with($class, '@')) {
```

jdreesen: Actually, the whole method could be written as:

```php
    protected static function resolve($class, callable $validationCallback = null)
    {
        if (!$class) {
            return null;
        }

        return self::$cache[$class] ??= self::returnValidServiceOrNull(
            str_starts_with($class, '@') ? \Pimcore::getContainer()->get(substr($class, 1)) : new $class, 
            $validationCallback,
        );
    }
```

passioneight: @jdreesen thank you for your feedback! I, too, was thinking about refactoring, but wasn't too sure about it, since the PR is only supposed to remove the `try-catch`.

Should we still use the new implementation? Maybe if we rename the PR to better reflect the change. WDYT?

brusch: @passioneight I'd go for the new implementation (as it's an improvement anyway) and rename the PR. 
Can you change the PR accordingly? Thanks a lot!

passioneight: Done.
Thanks & credits should go to @jdreesen :-)

dvesh3 (commented):
@passioneight Please add upgrade notes [here](https://github.com/pimcore/pimcore/tree/11.x/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes#1100) about the changed behavior. 

Should we also trigger a deprecation in 10.5? that you should cleanup unresolved classes otherwise it will throw exception in Pimcore 11. wdyt?

passioneight:
@dvesh3 sorry, I was on vacation, will have a look soon :-)

passioneight:
@dvesh3 I updated the upgrade notes as per your suggestion.
I think it's a good idea to add the deprecation warning to `10.5`.

passioneight:
@dvesh3 deprecation warning added in https://github.com/pimcore/pimcore/pull/13175

dvesh3:
@passioneight @jdreesen thanks guys!

PR #12294, opened 2022-05-31
[Session] Deprecate session configurators in favor of event listeners

dvesh3:
## Changes in this pull request  
Related to #12241

PR #12241, opened 2022-05-25
[Session] Do not use deprecated session service injections

dvesh3:
## Changes in this pull request  
Resolves #11042

## Additional info  
- [x] Registering a session bag through Session Configurator will not work anymore on Symfony 6 as Session service is removed. Use Event Listeners to register session bags before the session is started. Done here https://github.com/pimcore/pimcore/pull/12294

PR #12746, opened 2022-07-25
[Session] Fallback to session service when request or session in request is not available for BC

dvesh3:
Follow up to #12241 & #12294

## Additional info  
Steps to reproduce:
1. Execute a command with following code:
```
public function execute(InputInterface $input, OutputInterface $output)
    {
        $session = \Pimcore\Tool\Session::get();
        ...
```
throws exception `Request is not available`

### DIFF
--- a/bundles/EcommerceFrameworkBundle/EventListener/Frontend/TrackingCodeFlashMessageListener.php
+++ b/bundles/EcommerceFrameworkBundle/EventListener/Frontend/TrackingCodeFlashMessageListener.php
@@ -21,8 +21,10 @@ use Pimcore\Bundle\EcommerceFrameworkBundle\Tracking\TrackingManager;
 use Pimcore\Http\Request\Resolver\PimcoreContextResolver;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Cookie;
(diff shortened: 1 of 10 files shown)


bundles/EcommerceFrameworkBundle/EventListener/Frontend/TrackingCodeFlashMessageListener.php
```diff
@@ -90,14 +92,30 @@ public function onKernelRequest(RequestEvent $event)
...
+    /**
+     * @return SessionInterface
+     */
+    private function getSession()
+    {
+        try {
+            return $this->requestStack->getSession();
+        } catch (SessionNotFoundException $e) {
+        }
+
```
fashxp, line 105: we should not go for a fallback here but return null

fashxp (changes requested):

fashxp (approved):

bundles/EcommerceFrameworkBundle/EventListener/Frontend/TrackingCodeFlashMessageListener.php
dvesh3: fixed here https://github.com/pimcore/pimcore/pull/12746/commits/a5c037bad3ad318097cd68c6bd3dfe294e72c17d

PR #12753, opened 2022-07-25
fix: error when trying to access width or height

### LINKED ISSUE #12752: [Bug]: Update from 10.4 to 10.5 breaks code if thumbnail does not contain a config
### Expected behavior

Creating a thumbnail without configuration does not lead to an error.

### Actual behavior

image

### Steps to reproduce

1. Create thumbnail without configuration
2. Render thumbnail

Related code:

https://github.com/blankse/pimcore/blob/e72eb19f3332c6a1e2cfbfe236b14dfb5018fe0e/models/Asset/Thumbnail/ImageThumbnailTrait.php#L221

Proposed fix:
```php
if ($thumbnail && array_key_exists('width', $thumbnail) && array_key_exists('height', $thumbnail)) {
    // ...
}
```

passioneight:
resolves #12752

### DIFF
--- a/models/Asset/Thumbnail/ImageThumbnailTrait.php
+++ b/models/Asset/Thumbnail/ImageThumbnailTrait.php
@@ -218,7 +218,7 @@ trait ImageThumbnailTrait
 
             if ($config) {
                 $thumbnail = $asset->getDao()->getCachedThumbnail($config->getName(), $this->getFilename());
-                if ($thumbnail && $thumbnail['width'] && $thumbnail['height']) {
+                if ($thumbnail && array_key_exists('width', $thumbnail) && array_key_exists('height', $thumbnail)) {
                     $dimensions['width'] = $thumbnail['width'];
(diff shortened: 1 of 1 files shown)


models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -218,7 +218,7 @@ public function getDimensions()
 
             if ($config) {
                 $thumbnail = $asset->getDao()->getCachedThumbnail($config->getName(), $this->getFilename());
-                if ($thumbnail && $thumbnail['width'] && $thumbnail['height']) {
+                if ($thumbnail && array_key_exists('width', $thumbnail) && array_key_exists('height', $thumbnail)) {
```
jdreesen: Should be enough:

```suggestion
                if (isset($thumbnail['width'], $thumbnail['height'])) {
```

dvesh3:
@passioneight thanks!

PR #12756, opened 2022-07-25
[Translations] Fix permissions check and add userOwner & userModification columns

### LINKED ISSUE #12722: [Bug]: shared translations does not apply language permissions
### Expected behavior

When a user has view permissions for a language he should not be able to edit the shared translations for that language.

### Actual behavior

The user is able to edit the restricted language.

### Steps to reproduce

Tested on Pimcore 10.5. Unable to test this on the demo environment as it requires access to the users/roles.

Create a new user
- Permissions: enable 'Translations'
- Shared Translation Settings > Language Permissions: enable view access for one or more languages
- Save

Language_permissions

- Log in as the new user
- Go to shared translations
- Edit a translation for a restricted language

Shared_translations_edit

The same happens when setting the language permissions on a role and assigning the role to a user.

In Pimcore 6 the bug already existed partially. Directly editing the cell by double clicking was blocked, but you were still able to right click and edit the text/HTML.

dvesh3:
Resolves #12722

### DIFF
--- a/models/Translation.php
+++ b/models/Translation.php
@@ -68,6 +68,21 @@ final class Translation extends AbstractModel
      */
     protected $type = 'simple';
 
+
+    /**
+     * ID of the owner user
+     *
+     * @var int|null
+     */
+    protected ?int $userOwner = null;
+
+    /**
+     * ID of the user who make the latest changes
+     *
+     * @var int|null
+     */
+    protected $userModification;
+
     /**
      * @return string
      */
@@ -198,6 +213,38 @@ final class Translation extends AbstractModel
     public function setDomain(string $domain): void
     {
         $this->domain = !empty($domain) ? $domain : self::DOMAIN_DEFAULT;
+    }
+
+    /**
+     * @return int|null
+     */
+    public function getUserOwner(): ?int
+    {
+        return $this->userOwner;
+    }
+
+    /**
+     * @param int|null $userOwner
+     */
+    public function setUserOwner(?int $userOwner)
+    {
+        $this->userOwner = $userOwner;
+    }
+
+    /**
+     * @return int|null
+     */
+    public function getUserModification()
+    {
+        return $this->userModification;
+    }
+
+    /**
+     * @param int $userModification
+     */
+    public function setUserModification($userModification)
+    {
+        $this->userModification = $userModification;
     }
 
     /**
@@ -383,14 +430,6 @@ final class Translation extends AbstractModel
     public function save()
     {
         $this->dispatchEvent(new TranslationEvent($this), TranslationEvents::PRE_SAVE);
-
-        if (!$this->getCreationDate()) {
-            $this->setCreationDate(time());
-        }
-
-        if (!$this->getModificationDate()) {
-            $this->setModificationDate(time());
-        }
 
         $this->getDao()->save();
 
--- a/models/Translation.php
+++ b/models/Translation.php
@@ -226,7 +226,7 @@ final class Translation extends AbstractModel
     /**
      * @param int|null $userOwner
      */
-    public function setUserOwner(?int $userOwner)
+    public function setUserOwner(?int $userOwner): void
     {
         $this->userOwner = $userOwner;
     }
@@ -242,7 +242,7 @@ final class Translation extends AbstractModel
     /**
      * @param int|null $userModification
      */
-    public function setUserModification(?int $userModification)
+    public function setUserModification(?int $userModification): void
     {
         $this->userModification = $userModification;
     }
--- a/models/Translation.php
+++ b/models/Translation.php
@@ -81,7 +81,7 @@ final class Translation extends AbstractModel
      *
      * @var int|null
      */
-    protected $userModification;
+    protected ?int $userModification = null;
 
     /**
      * @return string
@@ -234,13 +234,13 @@ final class Translation extends AbstractModel
     /**
      * @return int|null
      */
-    public function getUserModification()
+    public function getUserModification(): ?int
     {
         return $this->userModification;
     }
 
     /**
-     * @param int $userModification
+     * @param int|null $userModification
      */
     public function setUserModification($userModification)
     {
--- a/models/Translation.php
+++ b/models/Translation.php
@@ -242,7 +242,7 @@ final class Translation extends AbstractModel
     /**
      * @param int|null $userModification
      */
-    public function setUserModification($userModification)
+    public function setUserModification(?int $userModification)
     {
         $this->userModification = $userModification;
     }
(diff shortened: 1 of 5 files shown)


models/Translation/Dao.php
```diff
@@ -173,8 +185,32 @@ public function createOrUpdateTable()
...
+        $updateTime = time();
+        $this->model->setModificationDate($updateTime);
+
+        if (!$this->model->getCreationDate()) {
+            $this->model->setCreationDate($updateTime);
+        }
+
+        // auto assign user if possible, if no user present, use ID=0 which represents the "system" user
+        $userId = 0;
+        $user = \Pimcore\Tool\Admin::getCurrentUser();
```
kjkooistra-youwe: Alternatively you could maybe do something like:
`$userId = \Pimcore\Tool\Admin::getCurrentUser()?->getId() ?? 0`

kjkooistra-youwe (approved):

kjkooistra-youwe:
Patched local environment. Langues are now blocked correctly.
Translations

models/Translation/Dao.php
dvesh3:
```suggestion
        $userId = \Pimcore\Tool\Admin::getCurrentUser()?->getId() ?? 0;
```

models/Translation/Dao.php
```diff
@@ -173,8 +185,31 @@ public function createOrUpdateTable()
...
+
+        if (!$this->model->getCreationDate()) {
+            $this->model->setCreationDate($updateTime);
+        }
+
+        // auto assign user if possible, if no user present, use ID=0 which represents the "system" user
+        $userId = \Pimcore\Tool\Admin::getCurrentUser()?->getId() ?? 0;
+        if ($user instanceof Model\User) {
+            $userId = $user->getId();
+        }
```
dvesh3:
```suggestion
```

dvesh3:
> Patched local environment. Langues are now blocked correctly. Translations

@kjkooistra-youwe thanks for the confirmation! this will be part of next bugfix release 10.5.1.

models/Translation.php
```diff
@@ -68,6 +68,21 @@ final class Translation extends AbstractModel
...
+     * @var int|null
+     */
+    protected ?int $userOwner = null;
+
+    /**
+     * ID of the user who make the latest changes
+     *
+     * @var int|null
+     */
+    protected $userModification;
```
blankse:
```suggestion
    protected ?int $userModification = null;
```

models/Translation.php
```diff
@@ -200,6 +215,38 @@ public function setDomain(string $domain): void
...
+     */
+    public function getUserOwner(): ?int
+    {
+        return $this->userOwner;
+    }
+
+    /**
+     * @param int|null $userOwner
+     */
+    public function setUserOwner(?int $userOwner)
```
blankse:
```suggestion
    public function setUserOwner(?int $userOwner): void
```

models/Translation.php
```diff
@@ -200,6 +215,38 @@ public function setDomain(string $domain): void
...
+    {
+        $this->userOwner = $userOwner;
+    }
+
+    /**
+     * @return int|null
+     */
+    public function getUserModification()
```
blankse:
```suggestion
    public function getUserModification(): ?int
```

models/Translation.php
```diff
@@ -200,6 +215,38 @@ public function setDomain(string $domain): void
...
+    {
+        return $this->userModification;
+    }
+
+    /**
+     * @param int $userModification
+     */
+    public function setUserModification($userModification)
```
blankse:
```suggestion
    public function setUserModification(?int $userModification)
```

models/Translation/Dao.php
```diff
@@ -173,8 +185,28 @@ public function createOrUpdateTable()
...
                           `creationDate` int(11) unsigned DEFAULT NULL,
                           `modificationDate` int(11) unsigned DEFAULT NULL,
+                          `userOwner` int(11) unsigned DEFAULT NULL,
+                          `userModification` int(11) unsigned DEFAULT NULL,
                           PRIMARY KEY (`key`,`language`),
                           KEY `language` (`language`)
                         ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;");
     }
+
+    protected function updateModificationInfos()
```
blankse:
```suggestion
    protected function updateModificationInfos(): void
```

models/Translation.php
```diff
@@ -200,6 +215,38 @@ public function setDomain(string $domain): void
...
+    /**
+     * @return int|null
+     */
+    public function getUserModification()
+    {
+        return $this->userModification;
+    }
+
+    /**
+     * @param int $userModification
```
blankse:
```suggestion
     * @param int|null $userModification
```

models/Translation.php
dvesh3: we don't add `void` return type-hints for simplicity.

models/Translation.php
dvesh3: please revert `void`

models/Translation/Dao.php
dvesh3: we don't add void return type-hints for simplicity.

models/Translation.php
blankse: I changed it

models/Translation.php
jdreesen: What? :confused:

dvesh3: just to be consistent with other models as we avoid the overhead of adding `void` return type to the existing methods in models.

blankse: @dvesh3 Than we can't get a higher PhpStan level. The next level 6 want to set return types. When there is no `: void` you have to set `@return void` in the php doc

dvesh3: Ok, that's a good point. @fashxp @brusch I think we to consider this change for our models. wdyt?

brusch: @dvesh3 yep, that's a valid point, better having proper ` : void` than `@return void` imho.

models/Translation.php
```diff
@@ -200,6 +215,38 @@ public function setDomain(string $domain): void
...
+     */
+    public function getUserModification(): ?int
+    {
+        return $this->userModification;
+    }
+
+    /**
+     * @param int|null $userModification
+     */
+    public function setUserModification(?int $userModification)
```
blankse:
```suggestion
    public function setUserModification(?int $userModification): void
```

models/Translation.php
dvesh3: We decided to add `:void` return typehints in new methods and then we can add `@return void` phpdoc to models on Pimcore 11 to upgrade phpstan level 6.

aashan10 (approved):

PR #12757, opened 2022-07-25
Marked traits in /lib/Extension/Bundle/Traits as @internal

### LINKED ISSUE #12684: Make traits in /lib/Extension/Bundle/Traits `@internal`
### Improvement description

Shouldn't be necessary to provide them publicly

aryaantony92:
Resolves #12684

### DIFF
--- a/doc/Development_Documentation/20_Extending_Pimcore/13_Bundle_Developers_Guide/05_Pimcore_Bundles/README.md
+++ b/doc/Development_Documentation/20_Extending_Pimcore/13_Bundle_Developers_Guide/05_Pimcore_Bundles/README.md
@@ -116,34 +116,3 @@ An example of a `composer.json` defining a Pimcore bundle:
     }
 }
 ```
-
-#### Returning the composer package version in extension manager
-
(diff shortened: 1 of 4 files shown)


kingjia90:
Please fix the upgrade note README merge conflict and remove this 
https://github.com/pimcore/pimcore/blob/484204406077891569373646ba8ade73a8f739c6/doc/Development_Documentation/20_Extending_Pimcore/13_Bundle_Developers_Guide/05_Pimcore_Bundles/README.md?plain=1#L120-L149

blankse:
Why it this internal now? It is still used by other repros and the methods are still used by the `pimcore:bundle:list` command. I get now following deprecation warnings:
`PHP Deprecated: The "Pimcore\Extension\Bundle\Traits\PackageVersionTrait" trait is considered internal. It may change without further notice. You should not use it from "CustomerManagementFrameworkBundle\PimcoreCustomerManagementFrameworkBundle". in /var/www/html/vendor/symfony/error-handler/DebugClassLoader.php on line 337`

PR #12759, opened 2022-07-26
Removed migration script from composer.json on post update

### LINKED ISSUE #12514: Composer: do not run migrations automatically on post update
### Improvement description

see: https://github.com/pimcore/skeleton/blob/469496120e95dd7755d7368c3ac168363ede20b0/composer.json#L34

aashan10:
Resolves #12514

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -221,8 +221,7 @@
     ],
     "post-update-cmd": [
       "Pimcore\\Composer::postUpdate",
-      "@pimcore-scripts",
-      "Pimcore\\Composer::executeMigrationsUp"
+      "@pimcore-scripts"
     ],
     "pimcore-scripts": [
       "Pimcore\\Composer::clearCache",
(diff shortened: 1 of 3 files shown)


brusch:
We should also change the docs accordingly, e.g. here https://pimcore.com/docs/pimcore/current/Development_Documentation/Installation_and_Upgrade/Updating_Pimcore/index.html

brusch:
@aashan10 I think we can also remove `Pimcore\Composer::executeMigrationsUp()` as we don't need it anywhere else ...

brusch:
@aashan10 note: we also need to remove that from skeleton and demos. 
E.g.: https://github.com/pimcore/skeleton/blob/11.x/composer.json#L34

aashan10:
@brusch I'll create the issues in skeleton and demo and remove the `executeMigrationsUp` method as well. Thanks

brusch:
@aashan10 you can straight create a PR as well, shouldn't be too much effort - think it's just a one-liner 😉

dvesh3:
for demo https://github.com/pimcore/demo/pull/341

Issue #12586, opened 2022-07-05
[Bug]: incorrect video dimensions


passioneight:
### Expected behavior

When using a `pimcore_video`, without setting a video, in frontend: either render `pimcore_editable_video_empty` without `width` and `height` or (in my opinion even better) do not render anything at all.

### Actual behavior

`pimcore_editable_video_empty` `div` is rendered with `style="width: 100%px; height: 300px`:

![image](https://user-images.githubusercontent.com/57256630/177339030-7a8b9a51-0f7b-4f0d-8380-a2d20f652516.png)

> So, the `width` has incorrect dimensions and since the video is empty an unwanted whitespace is added. 

### Steps to reproduce

1. Create a document
2. Add a `pimcore_video` editable (in our project, the editable is in an `area`)
3. Leave editable empty and save the document
4. Inspect frontend

aryaantony92:
Fixed by #12760

PR #12760, opened 2022-07-26
Fix: Incorrect dimension style in Video::getEmptyCode()

blankse:
Fix https://github.com/pimcore/pimcore/issues/12586


aashan10 (approved):

aryaantony92:
@blankse Thank you for the fix !

PR #12766, opened 2022-07-26
Fix: ImageThumbnailTrait::readDimensionsFromFile()

blankse:
Fix error reported in https://github.com/pimcore/pimcore/issues/12752#issuecomment-1195483836

Regression from https://github.com/pimcore/pimcore/pull/12461

### DIFF
--- a/models/Asset/Thumbnail/ImageThumbnailTrait.php
+++ b/models/Asset/Thumbnail/ImageThumbnailTrait.php
@@ -189,13 +189,19 @@ trait ImageThumbnailTrait
                 $localFile = $this->getLocalFile();
                 if (null !== $localFile) {
                     if ($imageInfo = @getimagesize($localFile)) {
-                        $this->getAsset()->getDao()->addToThumbnailCache(
-                            $this->getConfig()->getName(),
-                            basename($pathReference['storagePath']),
-                            filesize($localFile),
-                            $imageInfo[0],
-                            $imageInfo[1]
-                        );
+                        $dimensions = [
+                            'width' => $imageInfo[0],
+                            'height' => $imageInfo[1],
+                        ];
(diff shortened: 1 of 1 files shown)


kathangeorg (approved):
@blankse as far as I can say your changes work perfectly.

dvesh3:
@blankse thanks!

models/Asset/Thumbnail/ImageThumbnailTrait.php
```diff
@@ -189,13 +189,19 @@ private function readDimensionsFromFile(): array
...
                 if (null !== $localFile) {
                     if ($imageInfo = @getimagesize($localFile)) {
-                        $this->getAsset()->getDao()->addToThumbnailCache(
-                            $this->getConfig()->getName(),
-                            basename($pathReference['storagePath']),
-                            filesize($localFile),
-                            $imageInfo[0],
-                            $imageInfo[1]
-                        );
+                        $dimensions = [
```
brusch, line 192: @blankse what is the reason for defining `$dimensions`? 
Using `$imageInfo[0]` ... should be sufficient, right?

blankse: @brusch For the return value of readDimensionsFromFile. Otherwise it's always an empty array right?

dvesh3: Like it was before https://github.com/pimcore/pimcore/pull/12461/files#diff-b48489e5be12fb4a4ec4f716c03e4690a799c39a87430cbb1a435d7badebf97cL193

brusch: 🤦 thanks for clarifying 😊

PR #12767, opened 2022-07-26
Video Editable: Fix undefined array key "v"

vintagesucks:
Fixes `Warning: Undefined array key "v"` when using a YouTube URL like `https://youtu.be/ajqOIVOztwE?rel=0`.


aryaantony92:
@vintagesucks Thank you for the fix.

Issue #7341, opened 2020-10-16
[Data Objects] Remove `o_` prefixed of system properties and database columns


brusch:
Add (maybe an optional) compatibility layer to the data objects views (e.g. `object_MyClass`. 

Also consider this BC layer: #12769

brusch:
Done forget to include #7309

brusch:
Postponed.

mcop1:
@fashxp Bc layer would be great, but I think it would need some discussion. Whereas for javascript it wouldn´t be too big of a problem, what about the database? 

I had a look at the bundles and at least for them the database is the bigger problem, because in almost every bundle the properties are used in a direct sql statement or as condition for the listings, e.g. 

https://github.com/pimcore/data-hub/blob/6eda52aae09bd883340fd1f1f5f43b24aae9dec4/src/GraphQL/Resolver/QueryType.php#L287-L308

- Demo
Used as sql condition
- data-hub-simple-rest
Used for filtering/paging, used in WorkspaceResolver as sql condition
- data-importer
Used in sql statement
- data-hub
Used as sql condition, what about GraphQL?
- portal-engine
Used in sql statements
....

Thoughts:
- We can´t use generated columns, because they don´t work with auto_increment columns
- We could use a trigger to fill o_id/id, o_key/key, ... columns with the same values. But that would duplicate the amount of data, especially if we have to create another index for those columns
- We could use a view to select from, instead of selecting directly from objects table. This would give us the possibility to use column aliases and we would be more flexible in the future. But what about performance and what about other tables like assets and documents? 

Solution: 
https://github.com/pimcore/pimcore/issues/13512

PR #12769, opened 2022-07-26
Renew o_* references on cloning

### LINKED ISSUE #12768: [Bug]: 10.5: Cloning object keeps reference for object id
### Expected behavior

Cloning elements should deep-clone

### Actual behavior

Reference for id is kept in clone

### Steps to reproduce

Create data object class `Brand`.

Run this script:
```php
<?php
$itemMold = new \Pimcore\Model\DataObject\Brand();
$object1 = Pimcore\Model\Element\Service::cloneMe($itemMold);
$object2 = Pimcore\Model\Element\Service::cloneMe($itemMold);

$object1->setId(123);
var_dump($object2->getId());
```

Output is 123. This leads to the problem that when you call `save()` all the objects are tried to be saved with the same id!

The cause are the references in https://github.com/pimcore/pimcore/blob/b675696e8c3dff29e51d97e67fc79d3003c9ca3a/models/DataObject/AbstractObject.php#L248-L261 because even DeepCopy cannot copy those references but will keep it - it only deep-clones objects, see https://github.com/myclabs/DeepCopy/blob/14daed4296fae74d9e3201d2c4925d1acb7aa614/src/DeepCopy/DeepCopy.php#L121-L150
For non-object references we have to care in `__clone` method.

Example from above explained:
`$itemMold` contains reference `o_id` which points to `id`
$object1 is a clone of `$itemMold` but the references in all scalar reference fields are kept - as `$itemMold->o_id` points to `$itemMold->id` also `$object1->o_id` points to  `$itemMold->id`
Same for `$object2`: `$object2->o_id` points to  `$itemMold->id`

As soon as any `id` of those objects gets changed, the `o_id`s of all other objects will change as well.

BlackbitDevs:
Resolves #12768

Testable with this script:
```php
<?php
$itemMold = new \Pimcore\Model\DataObject\Brand();
$object1 = \Pimcore\Model\Element\Service::cloneMe($itemMold);
$object2 = \Pimcore\Model\Element\Service::cloneMe($itemMold);
$object1->setId(123);
echo '"'.$object2->getId().'" should be ""'.PHP_EOL;

$object3 = clone $object1;
echo '"'.$object3->getId().'" should be "123"'.PHP_EOL;
```

### DIFF
--- a/models/DataObject/AbstractObject.php
+++ b/models/DataObject/AbstractObject.php
@@ -1331,38 +1331,12 @@ abstract class AbstractObject extends Model\Element\AbstractElement
         parent::__clone();
 
         // renew references when cloning
-        unset($this->o_id);
-        $this->o_id = &$this->id;
-
-        unset($this->o_path);
-        $this->o_path = &$this->path;
-
-        unset($this->o_creationDate);
-        $this->o_creationDate = &$this->creationDate;
-
-        unset($this->o_userOwner);
-        $this->o_userOwner = &$this->userOwner;
-
-        unset($this->o_versionCount);
-        $this->o_versionCount = &$this->versionCount;
-
-        unset($this->o_modificationDate);
-        $this->o_modificationDate = &$this->modificationDate;
-
-        unset($this->o_locked);
-        $this->o_locked = &$this->locked;
-
-        unset($this->o_parent);
-        $this->o_parent = &$this->parent;
-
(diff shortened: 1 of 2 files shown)


mcop1:
Looks good, thank you!

brusch:
Needs to be considered here as well #7341 (when removing the BC layer).

PR #12771, opened 2022-07-27
Send a console error when using deprecated js functions

blankse:
When someone still use the ts() function it should be send a console error (so you have also a stack trace).


robertSt7:
@blankse  Thank you for the PR

PR #12773, opened 2022-07-27
Fix: Application Logger File Storage flysystem default path to be same as the Constant 

### LINKED ISSUE #12772: [Bug]: Application Logger File Object Flysystem default directory not considering constant causing BC break
### Expected behavior

When upgrading to 10.5, no BC break should happen with a custom PIMCORE_LOG_FILEOBJECT_DIRECTORY 

### Actual behavior

The default directory for application log file object in the flysystem is hardcoded to `/var/application-logger`, not properly considering any different definitions of PIMCORE_LOG_FILEOBJECT_DIRECTORY.

More details on https://github.com/pimcore/pimcore/pull/12704#issuecomment-1195873199

### Steps to reproduce

Have a defined PIMCORE_LOG_FILEOBJECT_DIRECTORY different than the default one.
Upgrade to 10.5 without configuring flysystem.yml to override the default directory path with the equivalent of PIMCORE_LOG_FILEOBJECT_DIRECTORY.

kingjia90:
Resolves #12772

### DIFF
--- a/bundles/CoreBundle/Resources/config/pimcore/flysystem.yaml
+++ b/bundles/CoreBundle/Resources/config/pimcore/flysystem.yaml
@@ -1,3 +1,7 @@
+parameters:
+    # TODO: remove this BC layer in 11 or as soon as PIMCORE_LOG_FILEOBJECT_DIRECTORY is deprecated
+    log.fileobject: !php/const PIMCORE_LOG_FILEOBJECT_DIRECTORY
+
 flysystem:
     storages:
         # TODO: remove private permissions once https://github.com/thephpleague/flysystem-bundle/pull/69 got merged
@@ -74,4 +78,6 @@ flysystem:
             adapter: 'local'
             visibility: private
             options:
-                directory: '%kernel.project_dir%/var/application-logger'
+                # TODO: revert default to `var/application-logger` when PIMCORE_LOG_FILEOBJECT_DIRECTORY is deprecated
+                directory: "%kernel.project_dir%%log.fileobject%"
+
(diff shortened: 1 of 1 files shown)


bundles/CoreBundle/Resources/config/pimcore/flysystem.yaml
```diff
@@ -74,4 +78,6 @@ flysystem:
             adapter: 'local'
             visibility: private
             options:
-                directory: '%kernel.project_dir%/var/application-logger'
+                # TODO: revert default to `var/application-logger` when PIMCORE_LOG_FILEOBJECT_DIRECTORY is deprecated
+                directory: "%kernel.project_dir%%log.fileobject%"
```
mcop1:
```suggestion
                directory: "%log.fileobject%"
```
Not 100% sure, but I think we don´t need `%kernel.project_dir%` here. I tried the following:

1.) Checked out 10.4.6 
2.) Added this line `PIMCORE_LOG_FILEOBJECT_DIRECTORY=/var/www/html/var/application-logger-custom/` to my .env file
3.) Started an import with data-importer
4.) Checked the logs tab and the file system, `fileObject` gets stored in `var/application-logger-custom/` as expected

After that:

5.) Checked out this pr
6.) Started the same import again
7.) Checked the filesystem, `fileObject` now gets stored in `var\www\html\var\application-logger-custom\` 

Info: The fallback for `fileObjects` stored before 10.4.6. is still working.

@kingjia90 What do you think? Could you please double check?

kingjia90: Just double-checked and you are right, in my local tests i've just used defined the constants as '/var/application-logger-custom', missing that it takes and is defined with a full path by looking at there
https://github.com/pimcore/pimcore/blob/b675696e8c3dff29e51d97e67fc79d3003c9ca3a/lib/Bootstrap.php#L206

PR #12776, opened 2022-07-27
[Grid] Enhance loading performance for non-admin users

BlackbitDevs:
When loading data for the grid for non-admin users an additional query condition is added for example in https://github.com/pimcore/pimcore/blob/b675696e8c3dff29e51d97e67fc79d3003c9ca3a/bundles/AdminBundle/Helper/GridHelperService.php#L589-L597

This condition is quite slow because the sub query has to be executed for every row individually.

This leads to the following query plan:
```sql
EXPLAIN 
SELECT COUNT(*) 
FROM object_12 
WHERE (o_path = '/Products' OR o_path LIKE '/Products/%') AND  (
  (select list from users_workspaces_object where userId in (302,290,287,623) and LOCATE(CONCAT(o_path,o_key),cpath)=1  ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
  OR
 (select list from users_workspaces_object where userId in (302,290,287,623) and LOCATE(cpath,CONCAT(o_path,o_key))=1  ORDER BY LENGTH(cpath) DESC LIMIT 1)=1
) AND  object_12.o_type IN ('object','folder')
```

1 | PRIMARY | objects | NULL | ALL | PRIMARY,fullpath,type,o_type_o_path_o_classId | NULL | NULL | NULL | 548601 | 25.01 | Using where
1 | PRIMARY | object_query_12 | NULL | eq_ref | PRIMARY | PRIMARY | 4 | pim.objects.o_id | 1 | 100.00 | Using where; Using index
3 | DEPENDENT SUBQUERY | users_workspaces_object | NULL | range | userId | userId | 4 | NULL | 18 | 100.00 | Using index condition; Using where; Using filesort
2 | DEPENDENT SUBQUERY | users_workspaces_object | NULL | range | userId | userId | 4 | NULL | 18 | 100.00 | Using index condition; Using where; Using filesort

Execution time: 32.5 seconds

---
With this PR the allowed paths are previously fetched and then a set of `OR CONCAT(o_path,o_key) LIKE '/allowed/workspace/path%'` conditions gets added. This should logically be the same as above query.

Example query:
```sql
EXPLAIN 
SELECT COUNT(*) FROM object_12 
WHERE (o_path = '/Products' OR o_path LIKE '/Products/%') AND (
  CONCAT(o_path,o_key) LIKE '/Products/Category A%' 
  OR CONCAT(o_path,o_key) LIKE '/Products/Category B%' 
  OR CONCAT(o_path,o_key) LIKE '/Products/Category C%' 
  ...
) AND  object_12.o_type IN ('object','folder')
```

1 | SIMPLE | objects | NULL | ALL | PRIMARY,fullpath,type,o_type_o_path_o_classId | NULL | NULL | NULL | 549066 | 25.01 | Using where
1 | SIMPLE | object_query_12 | NULL | eq_ref | PRIMARY | PRIMARY | 4 | pim.objects.o_id | 1 | 100.00 | Using where; Using index

Execution time: 2,9 seconds


BlackbitDevs:
Now also affects tree loading performance, this also contained a similarly slow query part in https://github.com/pimcore/pimcore/blob/b675696e8c3dff29e51d97e67fc79d3003c9ca3a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php#L215-L227
For above client instead of 3 seconds, now 0.8s for loading `/admin/object/tree-get-childs-by-id`.

Now you can advert Pimcore 11 as the "fastest Pimcore of all time" ;-)

kingjia90:
Thank you!

That's an interesting approach, but there might some missing points in `getAllowedPaths`: 
- when you have a user level permission `list` = 0 and the same cpath but on role level as `list` = 1, then it takes the user level value as valid, voiding the role level ones.
- only for listing, when you define the permission of a children, all its parents are granted list permissions even if not specified or 0, so the furthest path has precedence over (that was the "idea" behind `ORDER BY cpath`)

Tests are failing because Permissiontest3 has only `/permissionfoo/bars/userfolder/usertestobject` permission and should be having list=1 permissions on `/permissionfoo/` and all folders in between.
You can tag me if you have any questions, i've been through all that not so long time ago.

Maybe this could be of help
https://github.com/pimcore/pimcore/pull/11938/files#diff-f1c5d50aa07c3685f76a2947c0c3859e2582f8b683ebadf175679eb4aae78785

kingjia90:
@BlackbitNeueMedien just a hint on the latests codeception errors, when dealing with objects, the column name `path` is actually `o_path`, also note that for the concatenated `key` there is this differentiation:
- assets: `filename` 
- documents: `key`
- dataobjects: `o_key`

BlackbitDevs:
@kingjia90 Thanks. Will edit the PR...

kingjia90:
I don't know yet how can i help to fix the tests, but there's a question arousing: 
Will the `Element\Service::findForbiddenPaths...forbiddenAndAllowedSql` approach really be faster than the former `exists` 3 subqueries? 
`findForbiddenPaths` was intended to be used in the search, you can see the `LIKE` statements because it would be working with the `search_backend_data` table and it have some nested `for` loops and condition concatenation beforehand which made sense because it has to do a fulltext search in the column `fullpath` once. 

In `buildChildrenCondition` we work with the tree list in the sidebar, it would be more "shortcircuit" way: if you have folder structure like
```
a
└── b
    ├── bb1
    │   ├── bb11
    │   │   ├── bb111
    │   │   └── bb112
    │   └── bb12
    └── bb2
        ├── bb21
        └── bb22
```
from `a`,  to know if `b` is  should be visible(listable), it just takes ANY one of the permission of the children(grandchildren and so on) that is not voided by a `list=0` permission on user level; when nothing below the `b` is defined, then just takes the inheritation from parent.

If using `findForbiddenPaths` approach is better, then "brings back" the idea we had about a sort of cached `temp table` that merges the permissions conflicts (between roles and user level), so we would avoid those nested loops everytime.

BlackbitDevs:
As soon as the tests are green, I can do another performance comparison. Obviously my initial assumption that as soon as `list` is allowed for `/a` then it will also be allowed for all subfolders, was wrong. Thanks to your hint the `findForbiddenPaths()` already implements the correct logic. Of course we could implement a cache here (I think Pimcore\Cache would probably be enough, we do not need to save it in the database), but this will be a matter of another ticket. Imho any approach which does not use the subqueries of https://github.com/pimcore/pimcore/blob/b675696e8c3dff29e51d97e67fc79d3003c9ca3a/bundles/AdminBundle/Helper/GridHelperService.php#L589-L597 will be much faster because those are *dependent subqueries* which means that MySQL has to join those tables for each and every row. Also the set of `OR` conditions is not optimal because it requires full table scan but at least we do not need to do 2 subquery joins for each row.

Btw: @kingjia90 do you see the reason why the `Permissiontest3` test is failing?

BlackbitDevs:
@kingjia90 Imho the test is wrong. The tests which currently fails is https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L705-L709

`$this->permissionfoo` is defined in https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L159
So we have a folder `/permissionfoo`
It has one child element:
https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L162
So we have `/permissionfoo/bars`
This folder has 3 children:
https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L163-L167
So we have 
- `/permissionfoo/bars/userfolder`
- `/permissionfoo/bars/groupfolder`
- `/permissionfoo/bars/hugo`

The `/permissionfoo/bars/userfolder` has one child:
https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L168

So we end up with this folder structure:
- /
  - permissionfoo
    - bars
      - userfolder
        - usertestobject
      - groupfolder
      - hugo

`$this->userPermissionTest3` is defined in https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L268-L274

This user has only access to 
https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L168
which means he has access to `/permissionfoo/bars/userfolder/usertestobject`.

And so actually when fetching the children of `/permissionfoo` for a user who has only access to `/permissionfoo/bars/userfolder/usertestobject` should return `[]`, shouldn't it? Because he is not wllowed to `list` `/permissionfoo/bars`

BlackbitDevs:
Next failing test is https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L713-L717

This user has 2 roles:
https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L280

https://github.com/pimcore/pimcore/blob/bc483848fba6f3c937676361d634ecedb9d0247c/tests/Model/Permissions/ModelDataObjectPermissionsTest.php#L222-L234

The roles allow and disallow `list` for same folder. Now the question is what shall happen? Current implementation of `findForbiddenPaths()` disallows the folder as soon as it finds one workspace which disallows it. Current test assumes that the folder is allowed.

PS: But actually when I look at the folder structure again, we have:
- /
  - permissionfoo
    - bars 
      - groupfolder

And this user has list=1 and list=0 for `/permissionfoo/bars/groupfolder`. The test fetches the children of `/permissionfoo` for this user. I do not see any reason why anything else than `[]` should get returned because the user is not allowed to `list` `/permissionfoo/bars`

Now this continues also for the other tests. Before I continue changing all the tests, it would be good if @kingjia90 or @brusch can give some feedback about my 2 last comments.

kingjia90:
> he has access to `/permissionfoo/bars/userfolder/usertestobject`.
> And so actually when fetching the children of `/permissionfoo` for a user who has only access to `/permissionfoo/bars/userfolder/usertestobject` should return `[]`, shouldn't it?

No, at least only in the case of `list`, it implicitly grants permissions to `permissionfoo`,`bar`,`userfolder` , otherwise you wouldn't be able to navigate to the element `usertestobject` if you cannot see and open its parent folders in between.
 
So in this case the tests expect `bars` folder even though there are not any permission set.

kingjia90:
> And this user has list=1 and list=0 for `/permissionfoo/bars/groupfolder`. The test fetches the children of `/permissionfoo` for this user. I do not see any reason why anything else than `[]` should get returned because the user is not allowed to `list` `/permissionfoo/bars`

When conflicting permissions of the same element are present, the rule is that they get merged between roles but on user level is always taking precedence.
In other words, if you have multiple roles and if one of the conflicting permissions allows the same element, it ignores all the blocking permissions. If you customize on an user level though, either by allowing/disallowing, it has precedence over any rule defined in roles. So in this case should be allowed to `list`

BlackbitDevs:
Hmm, then `Service::findForbiddenPaths()` is wrong. This actually should only be a performance enhancement ticket - now it is about fixing broken permission logic...

> No, at least only in the case of list, it implicitly grants permissions to permissionfoo,bar,userfolder , otherwise you wouldn't be able to navigate to the element usertestobject if you cannot see and open its parent folders in between.
>
> So in this case the tests expect bars folder even though there are not any permission set.

Imho in your explanation there are too many special cases... is this documented somewhere how to configure permissions in those special cases - like colliding permissions for same folder or special treatment of `list` permission for `/a` when user is only allowed to `list` `/a/b/c`?

kingjia90:
> Hmm, then `Service::findForbiddenPaths()` is wrong. This actually should only be a performance enhancement ticket - now it is about fixing broken permission logic...
> 
> > No, at least only in the case of list, it implicitly grants permissions to permissionfoo,bar,userfolder , otherwise you wouldn't be able to navigate to the element usertestobject if you cannot see and open its parent folders in between.
> > So in this case the tests expect bars folder even though there are not any permission set.

Generally speaking, the logic and tests should be valid and consistent across the different areas (tree,search,quicksearch,grids,etc..). Fash extensively covered them and we even tested it against the before&after PRs to verify if there's any behaviour/expectation/perfomance change. 
Maybe `findForbiddenPaths` has problems, but it was not intended to be used there, it has more a global point of view and for search purposes, while the current working approach in tree view uses a more relative approach (checks in-common parent permission once).

> 
> Imho in your explanation there are too many special cases... is this documented somewhere how to configure permissions in those special cases - like colliding permissions for same folder or special treatment of `list` permission for `/a` when user is only allowed to `list` `/a/b/c`?

Unfortunately, at that time, i haven't found any documentation about this, everything about roles/merging rules was based on this  https://github.com/pimcore/pimcore/issues/1634#issuecomment-831885117 and treated the `a/b/c` listing cases as [bug ](https://github.com/pimcore/pimcore/issues/11603) as it didn't somehow meet the UX(?) expectations and logics. 
I've added lots of description and comments in the related [PRs](https://github.com/pimcore/pimcore/pulls?q=is%3Apr+is%3Aclosed+author%3Akingjia90+permissions) but, indeed, it lacks of entries in the offical documentation.

kingjia90:
Closing this PR for the moment due to inactivity. 
Feel free to reopen and/or tag me if you need further info and support about the permissions

BlackbitDevs:
Have to look if I find a client who pays for further development...
But perhaps you could also fix the remaining bug, @kingjia90 - I think we should not simply ignore the chance of a performance improvement of 1000%.

BlackbitDevs:
@kingjia90 This is still a big performance problem when there is a lot of data. Sadly current clients simply apply this PR via patching and are not willing to pay for getting this into Pimcore core. I think this PR is not far from being logically correct. Could you maybe take a look at it again? Because all mentioned problems are in `findForbiddenPaths` - if there are logic errors inside this, they still exist because this PR was not intended to change any logic but only to improve performance. And in the end there was a big discussion about permission logic problems - which existed before this PR and still exist. So maybe we find a way to solve this, otherwise Pimcore backend is difficult to use for big amounts of data (for non-admin users).

kingjia90:
Reopening for the moment.

It would take some time, needs to be refactored into a service, there is too much duplicated code (SonarCloud is not going to be happy 😄 ), may also need further tests targeting the grids and eventually some sort of benchmark/stress tests to see the actual performance improvement. 
Will try to do all that but in the meantime any help into getting closer to this is appreciated

IronSean:
Thanks @BlackbitDevs for starting down this path, and @kingjia90 for re-opening it. Here at Torq we've also started to run up against this. For some clients with millions of data objects it makes the data grids inoperable for non-admin users. An admin user can filter and sort a million rows in a second, but for a non-admin user it's taking 19 seconds to load 25 results, and an additional 19 seconds to get the total count. Since these both finish sequentially before the web request returns, the EXT.JS timeout of 30 seconds expires and the grid never loads. 

>  I think we should not simply ignore the chance of a performance improvement of 1000%.
I think it is more than that, beyond a certain point it's not just about faster performance, but about whether the system works. When queries take longer than 30 seconds the grid never loads and the system is unusable to users. Some of these clients are Enterprise subscribers, if so would opening an issue via enterprise support help prioritize this?

For clients making everyone an admin so the grids load quickly is not a valid option.

I do understand @kingjia90 the requirement to implicitly grant permissions to parent paths for listing. If you are on the default view, you must be able to see `/a` and `/a/b` in order to view and open `/a/b/c` manually. Though there is a workaround if you use a perspective which loads `/a/b/c` directly.

cancan101:
We are also running into this issue and is causing the asset view for our non admin users to take 20+ seconds rather than 1.5 seconds. The request ends up timing out and they get an error.

kingjia90:
Closing here for the moment in favor of a quicker fix of https://github.com/pimcore/admin-ui-classic-bundle/pull/767
Thank you again btw, your fix in the grid itself were good.

This topic would be still open to improvement, refactoring etc..

PR #12779, opened 2022-07-27
[Core] Migrate Custom Layout definition files to use LocationAwareConfigRepository

### LINKED ISSUE #12321: [Improvement]: Migrate Custom Layout definition files to use LocationAwareConfigRepository

dvesh3:
Resolves #12321


models/DataObject/ClassDefinition/CustomLayout/Dao.php
```diff
@@ -15,60 +15,106 @@
...
      * @var Model\DataObject\ClassDefinition\CustomLayout
      */
     protected $model;
 
+    public function configure()
+    {
+        $config = \Pimcore::getContainer()->getParameter('pimcore.config');
+
+        // @deprecated legacy will be removed in Pimcore 11
+        $loadLegacyConfigCallback = function ($legacyRepo, &$dataSource) {
```
dvesh3: This needs migration if legacy support is removed.

dvesh3: Added migration for migrating legacy config to current location aware repository.

lib/Model/DataObject/ClassDefinition/ClassLayoutDefinitionManager.php
```diff
@@ -26,12 +26,20 @@ class ClassLayoutDefinitionManager
...
+     *
      * Delete all custom layouts from db
      */
     public function cleanUpDeletedLayoutDefinitions(): array
     {
         $db = \Pimcore\Db::get();
-        $layouts = $db->fetchAllAssociative('SELECT * FROM custom_layouts');
+
+        try {
+            $layouts = $db->fetchAll('SELECT * FROM custom_layouts');
```
kingjia90:
```suggestion
            $layouts = $db->fetchAllAssociative('SELECT * FROM custom_layouts');
```
The `fetchAll` is deprecated and were replaced in https://github.com/pimcore/pimcore/issues/12363

kingjia90 (changes requested):
All looks good and works fine, just one small thing

bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -769,6 +769,28 @@ private function addObjectsNode(ArrayNodeDefinition $rootNode)
...
+                                            ->scalarNode('id')->end()
+                                            ->scalarNode('name')->end()
+                                            ->scalarNode('description')->end()
+                                            ->integerNode('creationDate')->end()
+                                            ->integerNode('modificationDate')->end()
+                                            ->integerNode('userOwner')->end()
+                                            ->integerNode('userModification')->end()
+                                            ->scalarNode('classId')->end()
+                                            ->integerNode('default')->end()
+                                            ->variableNode('layoutDefinitions')->end()
```
dpfaffenbauer, line 793: Sooner or later also Pimcore Class Definitions make sense to be moved to a Symfony Configuration. (not the container config, but a config though). Since Pimcore internally needs a lot of different fields within the layoutDefinitions, it makes sense to better describe what is allowed here

dvesh3: we can't define allowed types in layoutDefinitions as this stores serialized layout data for BC reasons. please see https://github.com/pimcore/pimcore/pull/12779/files#diff-7657572963ff89400760ba6fabcfe4ae198543d224132bc7ced6e562504bd799R208. 

yes, it absolutely make sense when we decide to move class definitions to Symfony config.

lib/Model/DataObject/ClassDefinition/ClassLayoutDefinitionManager.php
dvesh3: good catch! 😉

bundles/CoreBundle/DependencyInjection/Configuration.php
dpfaffenbauer: is it necessary to serialize it?

kingjia90 (approved):

bundles/CoreBundle/DependencyInjection/Configuration.php
dvesh3: yes, as the storage can also be settingsStore db table, where the data is stored in a single column.

dpfaffenbauer: shouldn't the settings-store then be responsible for Serialization and De-Serialization?

models/DataObject/ClassDefinition/CustomLayout.php
```diff
@@ -214,7 +240,7 @@ public function save($saveDefinitionFile = true)
             \Pimcore\File::mkdir(PIMCORE_CUSTOMLAYOUT_DIRECTORY);
         }
 
-        $this->getDao()->save($isUpdate);
+        $this->getDao()->save();
 
         $this->saveCustomLayoutFile($saveDefinitionFile);
```
kingjia90: This `saveCustomLayoutFile` and `getDefinitionFile` could be removed, if we don't support legacy anymore

models/DataObject/ClassDefinition/CustomLayout.php
```diff
@@ -214,7 +240,7 @@ public function save($saveDefinitionFile = true)
             \Pimcore\File::mkdir(PIMCORE_CUSTOMLAYOUT_DIRECTORY);
```
kingjia90: This would be obsolete

models/DataObject/ClassDefinition/CustomLayout.php
```diff
@@ -195,7 +221,7 @@ public static function create($values = [])
      */
     public function save($saveDefinitionFile = true)
```
kingjia90: `$saveDefinitionFile` should be `false`, or removed from 11, if we can trigger deprecation in 10.6.

lib/Model/DataObject/ClassDefinition/ClassLayoutDefinitionManager.php
```diff
@@ -26,12 +26,20 @@ class ClassLayoutDefinitionManager
     public const DELETED = 'deleted';
 
     /**
+     * @deprecated
```
kingjia90: what about if this should be deprecated in 10.6 and cleaned up in 11?

bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -769,6 +769,28 @@ private function addObjectsNode(ArrayNodeDefinition $rootNode)
...
+                                ->normalizeKeys(false)
+                                    ->prototype('array')
+                                        ->children()
+                                            ->scalarNode('id')->end()
+                                            ->scalarNode('name')->end()
+                                            ->scalarNode('description')->end()
+                                            ->integerNode('creationDate')->end()
+                                            ->integerNode('modificationDate')->end()
+                                            ->integerNode('userOwner')->end()
+                                            ->integerNode('userModification')->end()
```
kingjia90: I've encoureted a problem with this being intengerNode, since the custom layout is created in "2 steps" (first "add layout" and then you can edit and save), when first created, it has `null`on `UserModification`
https://github.com/pimcore/pimcore/blob/9cfb411ca31695c64fb164f1f00bf32f7f1df29a/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php#L322-L328
need to add UserModification default value here or change integerNode

kingjia90 (changes requested):
I've tried to take a look in why it doesn't load the custom layout list with `symfony-config` but haven't found the cause yet 😞 

In the meanwhile, i've found other things that could be improved

lib/Model/DataObject/ClassDefinition/ClassLayoutDefinitionManager.php
dvesh3: done here https://github.com/pimcore/pimcore/pull/12944

kingjia90 (approved):

bundles/CoreBundle/DependencyInjection/Configuration.php
dvesh3: yes, you are right. I fixed it now so symfony config doesn't store the serialize data however defining allowed types is a bigger task as we need to gather all possible properties on layout & data definitions. I think it makes sense to do this with migration of Class definitions to Symfony config.

PR #12786, opened 2022-07-27
Fix object bricks control menu for grouped bricks

### LINKED ISSUE #12791: [Bug]:  10.5.0 Add objectbrick to object not working when objectbrick groupname is not empty
### Expected behavior

Should be possible to add objectbrick when objectbrick has a groupname

image
image

### Actual behavior

Objectbricks with groupname not displayed in object dropdown

image
image

### Steps to reproduce

Set group name on objectbrick

ChrisB9:
Recently this feature got merged: https://github.com/pimcore/pimcore/pull/12084 .
While itself being a nice feature, it introduced a bug to the control menu for adding (a) new object brick(s) to an object.

Before:
image

After:
image

This does only affect grouped elements, as they don't have the `key`-property and their children do.


mcop1:
Thank you!

PR #12788, opened 2022-07-28
Metainfo user names

### LINKED ISSUE #12787: [Improvement]: Show user's name on metadata information panel
### Improvement description

When viewing asset metadata Pimcore will show the user ID of the owner and user who last modified the asset. Administrators have the ability to click on the user ID to view the user's details but this is not available to all users.

When there are multiple users of the DAM, it is nice to have the full name of the user shown in the panel, not just the ID.

Andreas-Koepke:
Add new endpoint for getting full name of user, supports multiple users.
Add full name of user to metadata panel.
  
Resolves #12787


Andreas-Koepke:
This ended up being a much more significant change than expected. 

I have updated all of the associated controllers to now include the user's name. 

All of the JS files that use pimcore.element.metainfo have also been updated.

bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -479,6 +480,24 @@ static function (Task $task) {
             $objectData['general']['versionDate'] = $objectFromDatabase->getModificationDate();
             $objectData['general']['versionCount'] = $objectFromDatabase->getVersionCount();
 
+            $userOwner = User::getById($objectData['general']['o_userOwner']);
+            if (empty($userOwner)) {
```
brusch: I think we should make this a bit more generic, what about adding a `getFullName()` method in the `User` object and maybe a trait that handles all the displaying related stuff?

brusch (changes requested):

aryaantony92:
@Andreas-Koepke Thank you for the PR. And just a gentle reminder about the generic approach. Please refer to the comment above.

Andreas-Koepke:
Sounds good to me. I can make these changes later this week.

aryaantony92:
> Sounds good to me. I can make these changes later this week.

Thank you 👍

Andreas-Koepke:
I have added in the new method for the User object and updated my code.

As for the trait for the display related code, I am not sure exactly what part of my changes you are referring to.

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -217,6 +218,24 @@ static function (Task $task) {
...
+
+        $userModification = ($asset->getUserOwner() == $asset->getUserModification()) ? $userOwner : User::getById($asset->getUserModification());
+        if (empty($userModification)) {
+            $data['userModificationUsername'] = '';
+            $data['userModificationFullname'] = $this->trans('user_unknown');
+        } else {
+            $data['userModificationUsername'] = $userModification->getName();
+            $data['userModificationFullname'] = $userModification->getFullName();
+        }
+
```
aryaantony92: We have similar code (for username and usermodification) in DataObjectController, DocumentController etc. 
If we add this code in a trait, then different controllers can access it from there. Thanks!

Andreas-Koepke: The new trait has been added, please review the updated code.

I have also made a fix to the metainfo panel so it handles the system account properly.

aryaantony92: Thank you! . Could you please resolve the conflicts also ?

bundles/AdminBundle/src/Controller/Traits/UserNameTrait.php
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @internal
+ */
+trait UserNameTrait
+{
+    /**
+     * @param int $user_id The User ID.
+     */
+    protected function getUserName($user_id) {
```
blankse:
```suggestion
    /**
     * @param int $user_id The User ID.
     *
     * @return array{username: string, fullname: string}
     */
    protected function getUserName(int $user_id): array
    {
```

We should use type hints for new methods

models/User.php
```diff
@@ -216,6 +216,14 @@ public function setLastname($lastname)
         return $this;
     }
 
+    /**
+     * @return string|null
+     */
+    public function getFullName()
```
blankse:
```suggestion
    public function getFullName(): string
```
We should use type hints for new methods

bundles/AdminBundle/src/Controller/Admin/Document/DocumentControllerBase.php
```diff
@@ -465,4 +468,18 @@ protected function saveDocument(Model\Document $document, Request $request, bool
 
         return [$task, $document, $version];
     }
+
+    /**
+     * @param Model\Document $document
+     * @param array $data
+     */
```
blankse:
```suggestion
```
We should avoid redundant php docs

Andreas-Koepke:
Merge conflicts have been resolved and some final modifications made.

aryaantony92:
> Merge conflicts have been resolved and some final modifications made.

Thank you.

bundles/AdminBundle/src/Controller/Traits/UserNameTrait.php
```diff
@@ -0,0 +1,45 @@
...
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\AdminBundle\Controller\Traits;
+
+use Pimcore\Model\User;
+
+/**
+ * @internal
+ */
```
NiklasBr, line 22: Why `internal`? Seems like a useful Trait which could be used elsewhere.

Andreas-Koepke: All of the other traits in that folder were also marked as internal so I just made it the same.

bundles/AdminBundle/src/Controller/Traits/UserNameTrait.php
```diff
@@ -0,0 +1,48 @@
...
+trait UserNameTrait
+{
+    /**
+     * @param int $user_id The User ID.
+     *
+     * @return array{username: string, fullname: string}
+     */
+    protected function getUserName(int $user_id): array
```
aryaantony92:
```suggestion
    protected function getUserName(int $userId): array
```

bundles/AdminBundle/src/Controller/Traits/UserNameTrait.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $user = User::getById($user_id);
```
aryaantony92:
```suggestion
        $user = User::getById($userId);
```

bundles/AdminBundle/src/Controller/Traits/UserNameTrait.php
```diff
@@ -0,0 +1,48 @@
...
+
+        if (empty($user)) {
+            $data = [
+                'username' => '',
+                'fullname' => $this->trans('user_unknown'),
```
aryaantony92:
```suggestion
                'userName' => '',
                'fullName' => $this->trans('user_unknown'),
```

aryaantony92: Could you please change the 'username' and 'fullname' also to Camel case like the suggestion ? Thanks

bundles/AdminBundle/src/Controller/Traits/UserNameTrait.php
brusch: Agree, we should keep that one `@internal`.

bundles/AdminBundle/src/Controller/Traits/UserNameTrait.php
Andreas-Koepke: I have made this change and the associate changes to the controllers.

aryaantony92:
@Andreas-Koepke I just tested the PR and it's working fine. But, what about moving the conditions like https://github.com/pimcore/pimcore/blob/a4fd44d71f805aa7e48f5475e76adcbcaff97d24/bundles/AdminBundle/public/js/pimcore/asset/asset.js#L495
https://github.com/pimcore/pimcore/blob/a4fd44d71f805aa7e48f5475e76adcbcaff97d24/bundles/AdminBundle/public/js/pimcore/asset/asset.js#L496 in Js files to the controller ? So, we need to pass only one variable from the controller (if full name exists, then full name otherwise the username). Since, we are using either full name/username in the js files, we don't need both of them in the js.

Andreas-Koepke:
I just made a change to the UserNameTrait that checks if the user's full name is available. I have removed the check from JS files as it is no longer necessary.

At the moment, I have left both variables in the data returned by the controllers.

aryaantony92:
@Andreas-Koepke Thanks a lot!

PR #12794, opened 2022-07-29
Fix: use the `RuntimeCache` instead of a static array to cache `UrlSlug`s

jdreesen:
Using a static array to cache `UrlSlug`s causes problems in integration/functional tests in combination with CoreShop, because the array is not reset between tests. So if you test an object with the same slug in multiple tests, the slug will change in each test case because CoreShop's [SluggableListener](https://github.com/coreshop/CoreShop/blob/aee62640ffe93dfc09980e0eb65ad09b35eed51b/src/CoreShop/Bundle/PimcoreBundle/EventListener/SluggableListener.php#L61-L73) will increment a number on the end of the slug if it already exists (in the cache).

Using the `RuntimeCache` should fix this, as it is cleared when the kernel is shut down.

### DIFF
--- a/models/DataObject/Data/UrlSlug.php
+++ b/models/DataObject/Data/UrlSlug.php
@@ -15,6 +15,7 @@
 
 namespace Pimcore\Model\DataObject\Data;
 
+use Pimcore\Cache\RuntimeCache;
 use Pimcore\Db;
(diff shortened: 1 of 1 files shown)


mcop1:
Thanks!

PR #12800, opened 2022-08-01
Refactored getMetadata method of Asset to include unused parameters

### LINKED ISSUE #12725: [Improvement]: Refactor Asset::getMetadata
### Improvement description

This method seems a bit inconsistent: https://github.com/pimcore/pimcore/blob/11.x/models/Asset.php#L1537

The first parameter `$name` decides if we receive all metadata or just the one with the name passed.

All other parameters `$language`, `$strictMatch`, `$raw` are only considered if `$name` is passed. But it would also make sense to consider them if we want to get all metadata.

aashan10:
Resolves #12725


models/Asset.php
```diff
@@ -1555,53 +1555,39 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
+        if ($language === null) {
+            $language = Pimcore::getContainer()->get(LocaleServiceInterface::class)->findLocale();
         }
 
         $metaData = $this->getObjectVar('metadata');
+
         $result = [];
+
         if (is_array($metaData)) {
             foreach ($metaData as $md) {
```
kingjia90: Shouldn't this be `$this->metadata` (after `AssetEvents::PRE_GET_METADATA`) instead of `$metaData` (from `getObjectVar`)? If yes, we should be also removing `$metaData = $this->getObjectVar('metadata');`

models/Asset.php
```diff
@@ -1555,53 +1555,39 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
-                if ($raw) {
-                    return $data;
-                }
-
-                return $convert($data);
-            }
-
-            return null;
+        if ($language === null) {
+            $language = Pimcore::getContainer()->get(LocaleServiceInterface::class)->findLocale();
```
kingjia90: *here

models/Asset.php
```diff
@@ -1555,53 +1555,39 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
+                if($name) {
+                    if ($md['name'] == $name) {
+                        if ($language == $md['language'] || (empty($md['language']) && !$strictMatch)) {
+                            if ($raw) {
+                                return $md;
+                            }
+                            return $convert($md);
+                        }
+                    }
+                } else if ($language == $md['language'] || (empty($md['language']) && !$strictMatch)) {
```
kingjia90: The logic is changed and wrong here:
When specific `$language` and `$name` are not set, it should return all the metadata in an array, but with this PR `$language` is now always set at line 1559* and as soon as encounters a metadata row that has a different language (eg. `$language = 'en'` and `$md['language'] = 'de'` not empty), it doesn't match the conditions of this `else if` and exits by returning `null` discarding every other metadata rows stored in `$result[]`

kingjia90 (changes requested):

kingjia90:
I would suggest to keep the current 2 loops (one when $name and $language are set and one when you want all the others in a array), but adding the feature to filter by language here to the latest loop https://github.com/pimcore/pimcore/blob/20e1ac63052877391ce640d8e19fd6c3b588149e/models/Asset.php#L1601-L1607
by adding something like `if ( (isset($language) && $language != $md['language']) || !empty($md['language']) && $strictMatch) { continue; }`

So when you set the `$name`, and either specific `$language` or not (in that case it will use the locale as fallback anyways) then it returns a `string`, which can be converted or not depending on the `$raw`.

When you don't set `$name`, you are not aiming for a specific metadata, so you can either set a specific $language or not:
- when $language is set, you get an array of all the metadatas of a specific language, (thanks to that `continue` you can skip the other languages)
- when $language is not set, then you get an array with all metadatas (as how it currently were)

In both cases, your data is converted accordingly by the `$raw` and it also considers the `$strictMatch`, which is only relevant when `$language` is set

aashan10:
@kingjia90 I have brought back the two loops and added filters for `$language` and `$strictMatch` on the case with no `$name`. Please have a look now!

models/Asset.php
```diff
@@ -1594,10 +1594,21 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
         if (is_array($metaData)) {
             foreach ($metaData as $md) {
                 $md = (array)$md;
-                if (!$raw) {
-                    $md['data'] = $convert($md);
+                if (empty($md['language']) && !$strictMatch) {
+                    if (!$raw) {
+                        $md['data'] = $convert($md);
+                    } else {
```
kingjia90: this `else` is wrong here, the converted `$md`  would be never pushed to `$result` array.

models/Asset.php
```diff
@@ -1594,10 +1594,21 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
+                        $result[] = $md;
+                    }
+                } else {
+                    if($md['language'] == $language) {
```
kingjia90: what about using `elseif` or moving this IF block as second condition with an OR on the first IF block since the code is the same?
eg. `if ((empty($md['language']) && !$strictMatch)  || ($md['language'] == $language)) {`

kingjia90 (changes requested):
Looks better but there are some issues.

Beside that, since we are here on refactoring, i have noticed that in current code this https://github.com/pimcore/pimcore/blob/abdaf98297b543b515e8f312bab136b9ed8031da/models/Asset.php#L1582-L1584 is never reached.

What about changing this part 
https://github.com/pimcore/pimcore/blob/abdaf98297b543b515e8f312bab136b9ed8031da/models/Asset.php#L1564-L1576
by removing `if ($raw)` and using `$data = $md; break;` instead of these returns, in this way we do this `$raw` and `$convert` once outside the loop.

kingjia90:
Another thing is that, by guessing the purpose of https://github.com/pimcore/pimcore/blob/abdaf98297b543b515e8f312bab136b9ed8031da/models/Asset.php#L1576
is to "temporary store" the metadata with no language set and keep looping through all the metadata to check if there is the same metadata name but with specific language (`if ($language == $md['language']) {`). 
As soon as it is found, we "discard" that `$data` and take in consideration the language-specificied metadata, otherwise it is just using $data as the fallback value at the end of the loop.

I would say these 2 ` $result[] = $md;` would be wrong in this case. We need something similar that uses the metadata with specific language when exists and fallbacks to the no-languages-specified one, but you can only know it at the end of the loop.

aashan10:
@kingjia90 I have made changes accordingly!

models/Asset.php
```diff
@@ -1562,28 +1562,15 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
-
-                        return $convert($md);
-                    }
-                    if (empty($md['language']) && !$strictMatch) {
-                        if ($raw) {
-                            return $md;
-                        }
+                    if ($language == $md['language'] || (empty($md['language']) && !$strictMatch)) {
                         $data = $md;
+                        break;
```
kingjia90: This cannot be breaking so early in the case of `empty($md['language']) && !$strictMatch`, see:
https://github.com/pimcore/pimcore/pull/12800#issuecomment-1232750617

models/Asset.php
```diff
@@ -1594,10 +1581,9 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
         if (is_array($metaData)) {
             foreach ($metaData as $md) {
                 $md = (array)$md;
-                if (!$raw) {
-                    $md['data'] = $convert($md);
+                if ((empty($md['language']) && !$strictMatch) || ($md['language'] == $language)) {
+                    $result[] = $raw ? $md : $convert($md);
```
kingjia90: also like above

kingjia90 (changes requested):

models/Asset.php
aashan10: @kingjia90 I have updated accordingly and now the tests are happy :)

models/Asset.php
```diff
@@ -1594,10 +1586,13 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
         if (is_array($metaData)) {
             foreach ($metaData as $md) {
                 $md = (array)$md;
-                if (!$raw) {
-                    $md['data'] = $convert($md);
+
+                if ((empty($language) && !$strictMatch) || ($language == $md['language'])) {
```
kingjia90:
```suggestion
                if ((empty($md['language']) && !$strictMatch) || ($language == $md['language'])) {
```
but still it doesn't work properly, the "default" case (what we see in admin interface, without any parameter passed), the result should be showing all of them as previously did.

models/Asset.php
```diff
@@ -1562,28 +1562,20 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
-                    if ($language == $md['language']) {
-                        if ($raw) {
-                            return $md;
-                        }
-
-                        return $convert($md);
-                    }
                     if (empty($md['language']) && !$strictMatch) {
                         if ($raw) {
                             return $md;
```
kingjia90: this part could be removed because it's converted below when $data is set at the end of the loop, but it looks like a bug unrelated to the change on this PR

kingjia90 (changes requested):
There are still some flaws. 
We also would need to extend the tests to support these changes, can we add them so we can more easily see and cover all these combinations?

aashan10:
@kingjia90 I have added the tests. Please have a look

tests/Model/Asset/Metadata/NormalizerTest.php
```diff
@@ -159,4 +159,54 @@ public function testSelectMetadata()
...
+        $savedMetaData = $asset->getMetadata(null, $language, true, true);
+        $this->assertCount(1, $savedMetaData);
+
+        // When no name given and language is not strictly matched, all metadata should be returned
+        $savedMetaData = $asset->getMetadata(null, $language, false, true);
+        $this->assertCount(3, $savedMetaData);
+
+        // When no parameters are passed, metadata with locale should get priority over the ones that don't
+        $savedMetaData = $asset->getMetadata();
+        $this->assertCount(2, $savedMetaData);
```
kingjia90: in this case, it should display all, otherwise when you open the metadata tab when viewing an asset in the admin edit mode, it wouldn't list all the available metadata
see: https://github.com/pimcore/pimcore/blob/ba16263e44f45a6986568b65ab730f540f301ef0/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php#L201

kingjia90 (changes requested):

tests/Model/Asset/Metadata/NormalizerTest.php
```diff
@@ -159,4 +159,54 @@ public function testSelectMetadata()
...
+        $this->assertEquals($savedMetaData['language'], $language);
+        $this->assertEquals($savedMetaData['data'], $localizedData);
+
+        // When name is not given and language is strictly matched, only metadata with specified locale should be returned
+        $savedMetaData = $asset->getMetadata(null, $language, true, true);
+        $this->assertCount(1, $savedMetaData);
+
+        // When no name given and language is not strictly matched, all metadata should be returned
+        $savedMetaData = $asset->getMetadata(null, $language, false, true);
+        $this->assertCount(3, $savedMetaData);
```
kingjia90: This case should be returning 2, `metadata with locale should get priority over the ones that don't` (like said in comment on line 207), otherwise there would be no difference when setting a `$language` or not , when strict match is false

tests/Model/Asset/Metadata/NormalizerTest.php
```diff
@@ -159,4 +159,54 @@ public function testSelectMetadata()
...
+
+        // When no parameters are passed, metadata with locale should get priority over the ones that don't
```
kingjia90:
```suggestion
        // When no parameters are passed, all metadata should be returned
```

doc/Development_Documentation/04_Assets/01_Working_with_PHP_API.md
```diff
@@ -64,7 +64,15 @@ therefore used on all output channels.
 
 ### Examples
 ##### Getting Data
- ```php
+Metadata for an asset can be fetched using `getMetadata` method of Asset class. It takes 4 parameters: 
+1. `$name` can be a nullable string which corresponds to the name of metadata. If null is passed, it tries to get all the metadata. The actual contents may be manipulated by other parameters.
```
kingjia90:
```suggestion
1. `$name` (string) when passed it returns the metadata with the given name, otherwise it lists all.
```

this should implicitly suggest that it can either return a single value or a set of values (array)

doc/Development_Documentation/04_Assets/01_Working_with_PHP_API.md
```diff
@@ -64,7 +64,15 @@ therefore used on all output channels.
...
+2. `$language` is a nullable string which corresponds to language of the metadata. If language is passed, it tries to fetch metadata in a specific language.
```
kingjia90:
```suggestion
2. `$language` (string) with this parameter you can filter out the metadata of a specific language.
```

doc/Development_Documentation/04_Assets/01_Working_with_PHP_API.md
```diff
@@ -64,7 +64,15 @@ therefore used on all output channels.
...
+3. `$strictMatch` is a boolean value which, if true tries to get only the metadata which exactly matches the requested language.
```
kingjia90:
```suggestion
3. `$strictMatch` (boolean) if true tries to get only the metadata which exactly matches the requested language without considering the fallback.
```

doc/Development_Documentation/04_Assets/01_Working_with_PHP_API.md
```diff
@@ -64,7 +64,15 @@ therefore used on all output channels.
 
 ### Examples
 ##### Getting Data
- ```php
+Metadata for an asset can be fetched using `getMetadata` method of Asset class. It takes 4 parameters:
```
kingjia90:
```suggestion
Metadata for an asset can be fetched using `getMetadata` method of Asset class. It can take the following optional parameters: 
```
saying `optional` should be sufficent to hint that all params can be `nullable`

doc/Development_Documentation/04_Assets/01_Working_with_PHP_API.md
```diff
@@ -64,7 +64,15 @@ therefore used on all output channels.
 
 ### Examples
 ##### Getting Data
- ```php
+Metadata for an asset can be fetched using `getMetadata` method of Asset class. It takes 4 parameters: 
+1. `$name` can be a nullable string which corresponds to the name of metadata. If null is passed, it tries to get all the metadata. The actual contents may be manipulated by other parameters.
+2. `$language` is a nullable string which corresponds to language of the metadata. If language is passed, it tries to fetch metadata in a specific language.
+3. `$strictMatch` is a boolean value which, if true tries to get only the metadata which exactly matches the requested language.
+4. `$raw` is a boolean value which, if true gets extra information like name, data, language and input type of the metadata.
```
kingjia90:
```suggestion
4. `$raw`  (boolean) if true, it will also return extra information like name, data, language and input type of the metadata.
```

doc/Development_Documentation/04_Assets/01_Working_with_PHP_API.md
```diff
@@ -73,6 +81,18 @@ $asset->getMetadata("title");
...
 $asset->getMetadata("title", "en");
 // if there's no title for "en" but one without a language this will be returned (fallback mechanism).
+
+// get all available metadata
+$asset->getMetadata();
+
+// get all available metadata for a specific language along with metadata which have no language assigned
+$asset->getMetadata(null, 'en');
+
+// get all available metadata for a specific language discarding any that does not have language set
```
kingjia90:
```suggestion
// get exclusively all the metadata that have a specific language
```

doc/Development_Documentation/04_Assets/01_Working_with_PHP_API.md
jdreesen: Maybe this parameter can be renamed to better reflect what is does, like `$strictLanguageMatch`.

models/Asset.php
```diff
@@ -1540,68 +1540,88 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
 
-            $data = null;
-            foreach ($this->metadata as $md) {
-                if ($md['name'] == $name) {
-                    if ($language == $md['language']) {
-                        if ($raw) {
-                            return $md;
-                        }
+        foreach($result as $key => &$item) {
+            if ($language && !$strictMatch) {
```
kingjia90:
```suggestion
        if ($language && !$strictMatch) {
            foreach($result as $key => &$item) {
```

models/Asset.php
```diff
@@ -1540,68 +1540,88 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
+                $lang = $item['language'];
 
-                        return $convert($md);
-                    }
-                    if (empty($md['language']) && !$strictMatch) {
-                        if ($raw) {
-                            return $md;
-                        }
-                        $data = $md;
+                if (!$lang && isset($metaDataWithLanguage[$language]) && isset($metaDataWithLanguage[$language][$item['name']])) {
```
kingjia90:
```suggestion
                if (!$item['language'] && isset($metaDataWithLanguage[$language][$item['name']])) {
```
and remove the assignment `$lang = $item['language'];`, since `$lang` not used elsewhere beside in this IF block.
In case it is better to keep it there, probably would rename `$lang` to something more descriptive.

The `isset` of the deeper array key would be enough

models/Asset.php
```diff
@@ -1540,68 +1540,88 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
-            if ($data) {
-                if ($raw) {
-                    return $data;
-                }
+        return $result;
+    }
 
-                return $convert($data);
-            }
+    protected function transformMetadata(array $metaData) {
```
kingjia90:
```suggestion
    private function transformMetadata(array $metaData) {
```
it was formerly an anonymous function `$convert`, probably it should be `private`, but i am not sure

models/Asset.php
```diff
@@ -1540,68 +1540,88 @@ public function getMetadata($name = null, $language = null, $strictMatch = false
...
-        $result = [];
-        if (is_array($metaData)) {
-            foreach ($metaData as $md) {
-                $md = (array)$md;
-                if (!$raw) {
-                    $md['data'] = $convert($md);
+        return $transformedData;
+    }
+
+    protected function getMetadataByName(string $name, ?string $language = null, bool $strictMatch = false, bool $raw = false) {
```
kingjia90: the phpdoc return type are missing, since it's a PR for 11, native return type would be also ok.

kingjia90 (changes requested):
Looks good and works, just having some minor changes to suggest.
It would be also nice to add an entry in the changelog.

doc/Development_Documentation/04_Assets/01_Working_with_PHP_API.md
kingjia90: I agree, without specifying language, it could be "confused" for some strict matching metadata name feature, like if it would be possible to fuzzy search eg. `copy*`.

aashan10: Renamed the flag to `$strictMatchLanguage`. Thanks @jdreesen for the feedback!

models/Asset.php
aashan10: Added param types and return types for the method

models/Asset.php
```diff
@@ -1525,83 +1525,95 @@ public function removeMetadata(string $name, ?string $language = null)
...
-            $data = null;
-            foreach ($this->metadata as $md) {
-                if ($md['name'] == $name) {
-                    if ($language == $md['language']) {
-                        if ($raw) {
-                            return $md;
-                        }
+        if ($language && !$strictMatchLanguage) {
+            foreach($result as $key => &$item) {
+                $lang = $item['language'];
```
kingjia90:
```suggestion

```
this line can be deleted.

Please also add a upgrade note about the improvement.

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,5 +1,6 @@
 # Upgrade Notes
 ## 11.0.0
+- [Assets] Refactored and added tests for `Pimcore\Model\Asset::getMetadata` method to include the parameters even when `$name` parameter is not provided. Added native type hints for this method.
```
kingjia90:
```suggestion
- [Assets] Refactored `Pimcore\Model\Asset::getMetadata` method to allow listing of all metadata entries filtered by a specific language. Prior this version, the language filter was only available when a specific metadata name was defined in the parameters. Added native type hints and related tests.
```
what about something like this? changed `parameters` to `language` because that would be a little too vague,  `$raw` was already working and is unchanged while `$strictMatch` is just something tied to $language anyways.

kingjia90: oh wait, i don't make mention of the `$raw` then.. i was simplyifing to language because the strictMatch is "linked" to it anyways

kingjia90 (approved):

PR #12812, opened 2022-08-02
[Bug]: Manual sorting of object does not work

### LINKED ISSUE #12628: [Bug]: Manual sorting of object does not work
### Steps to reproduce

Have an object tree with a number of objects (in our example five) under a parent object. You want to sort them manually. The are currently sorted like this: 

before

Notice that "Führungsschiene" is currently on 4th position. Now let's move the object "Sägetisch" to the top within that "branch" of the object tree.

### Actual behavior

When manually sorting the objects dragging an object to a different position within the same "branch" the object that is being moved gets the right position. All other objects are resorted as well (I haven't researched yet whether they're being sorted by ID or by name).

The result in our example now looks like this:
after

"Sägetische" is now on the first position, but "Führungsschiene" is now on second position.

### Expected behavior

The sorting of the other objects should not change. In our example this means "Führungsschiene" should not move to the second position but of the other objects, but should actually be on fifth position.

Obviously the index does have to change, but it should reflect the previous order.

Pimcore version in use: 10.4.4

robertSt7:
resolve #12628 

Tests: Please also test dropping new object(s) in the folder

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -1192,29 +1192,32 @@ class DataObjectController extends ElementControllerBase implements KernelContro
             $list = new DataObject\Listing();
             $updatedObject->saveIndex($newIndex);
 
+            /*
+             * The cte and the limit are needed to order the data before the newIndex is set
+             */
             Db::get()->executeStatement(
                 'UPDATE '.$list->getDao()->getTableName().' o,
                     (
-                        SELECT newIndex, o_id FROM (SELECT @n := IF(@n = ? - 1,@n + 2,@n + 1) AS newIndex, o_id
-                        FROM '.$list->getDao()->getTableName().',
-                        (SELECT @n := -1) variable
-                        WHERE o_id != ? AND o_parentId = ? AND o_type IN (\''.implode(
-                    "','", [
-                        DataObject::OBJECT_TYPE_OBJECT,
-                        DataObject::OBJECT_TYPE_VARIANT,
-                        DataObject::OBJECT_TYPE_FOLDER,
-                    ]
-                ).'\')
-                            ORDER BY o_index, o_id=?
+                        SELECT newIndex, o_id
+                        FROM (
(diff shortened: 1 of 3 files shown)


mcop1:
@robertSt7 there seems to be a problem when moving more than object to the folder, e.g. drag & drop "A", "B" & "C" to the new folder:
image

Results in
image

But after refresh
image

mcop1:
@robertSt7 there is one more thing
1.) Switch back to "A-Z" sort mode
2.) Try to drag & drop an object inside the folder (tried to drag E above A)
3.) Get corresponding error message, which is fine
image
4.) But d&d breaks after that and you can find an error in the console

```
Uncaught TypeError: this.treeNodeMoveParameter.oldParent.getOwnerTree is not a function
    at klass.onTreeNodesDrop (script-proxy?storageFile=minified_javascript_core_8253144a0ef4d0cf82111e9839e08bf0.js&_dc=1d203273aee11e253295e86d97eadb430bbcdbce:99329:57)
    at constructor.fire (ext-all.js?_dc=1d203273aee11e253295e86d97eadb430bbcdbce:2407:19)
    at constructor.doFireEvent (ext-all.js?_dc=1d203273aee11e253295e86d97eadb430bbcdbce:2464:104)
    at constructor.doFireEvent (ext-all.js?_dc=1d203273aee11e253295e86d97eadb430bbcdbce:4851:110)
    at prototype.doFireEvent (ext-all.js?_dc=1d203273aee11e253295e86d97eadb430bbcdbce:4452:267)
    at constructor.fireEventArgs (ext-all.js?_dc=1d203273aee11e253295e86d97eadb430bbcdbce:2457:223)
    at constructor.<anonymous> (ext-all.js?_dc=1d203273aee11e253295e86d97eadb430bbcdbce:2479:188)
    at constructor.fire (ext-all.js?_dc=1d203273aee11e253295e86d97eadb430bbcdbce:2407:19)
    at constructor.doFireEvent (ext-all.js?_dc=1d203273aee11e253295e86d97eadb430bbcdbce:2464:104)
    at constructor.doFireEvent (ext-all.js?_dc=1d203273aee11e253295e86d97eadb430bbcdbce:4851:110)
```

PR #12817, opened 2022-08-03
Update developer doc example HTML in layout.html.twig

falconsmilie:

Update the example code in the Create a First project documentation. 

Resolves #
Move style to head from disallowed location https://developer.mozilla.org/en-US/docs/web/html/element/style

Remove deprecated type attribute in style tag https://developer.mozilla.org/en-US/docs/web/html/element/style#attr-type


aryaantony92:
@falconsmilie Thank you !

PR #12820, opened 2022-08-03
Fix: explode_and_trim() returns empty array

### LINKED ISSUE #12819: [Bug]: explode_and_trim() returns empty array
### Expected behavior

The function ``explode_and_trim('/', '/my/test/path/)``should return a non empty array.

The expected result is:
```
array(
 'my',
'test',
'path'
)
```
.

### Actual behavior

In Pimcore 10.5 and above the function is returning:
```
array(
'/my/test/path'
)
```

This is because the ``limit``parameter is defaulted to zero and passed to the ``explode`` function as a third parameter within the helper method in helper-functions.php.

https://www.php.net/manual/en/function.explode.php:
> If the limit parameter is zero, then this is treated as 1.

see https://github.com/pimcore/pimcore/blob/11.x/lib/helper-functions.php#L376

### Steps to reproduce

See example provided in expected behavior section.
Error exists in 10.5 and ^11. 

### Solution
Defaulting parameter ``$limit = PHP_INT_MAX`` might do it.

blankse:
Fix #12819


jdreesen (approved):

lib/helper-functions.php
```diff
@@ -371,7 +371,7 @@ function rscandir($base = '', &$data = [])
  *
  * @phpstan-param non-empty-string $delimiter
  */
-function explode_and_trim($delimiter, $string = '', $limit = 0, $useArrayFilter = true)
+function explode_and_trim($delimiter, $string = '', $limit = PHP_INT_MAX, $useArrayFilter = true)
```
jdreesen: Maybe we should remove the default parameter for `$string`, too. It doesn't really make sense to call this with an empty string, does it?

```suggestion
function explode_and_trim($delimiter, $string, $limit = PHP_INT_MAX, $useArrayFilter = true)
```

PR #12824, opened 2022-08-04
Update 3rd-party-licenses.md

### LINKED ISSUE #12763: [Improvement][Docs]: Versions of 3rd party libraries
### Improvement description

Would it make sense to add the versions for the 3rd party libraries somewhere?  
I would propose it in this file https://github.com/pimcore/pimcore/blob/11.x/3rd-party-licenses.md

mysliwietzflorian:
- refer to `PIMCORE_MODIFICATIONS.md` to see the versions of 3rd party libraries

  

Resolves #12763

## Additional info  

I though some cross-referencing might help 😉

### DIFF
--- a/3rd-party-licenses.md
+++ b/3rd-party-licenses.md
@@ -6,6 +6,9 @@ The Pimcore project is licensed under the terms of the [GPLv3](LICENSE.md).
 However, Pimcore includes several third-party Open-Source libraries, 
 which are licensed under their own respective Open-Source licenses.
 
+The specific verions of a library are documented in
+[PIMCORE_MODIFICATIONS.md](bundles/AdminBundle/Resources/public/js/lib/PIMCORE_MODIFICATIONS.md).
+
 ## Libraries directly included in Pimcore
 
 ### ExtJS by Sencha Inc.


3rd-party-licenses.md
```diff
@@ -6,6 +6,9 @@ The Pimcore project is licensed under the terms of the [GPLv3](LICENSE.md).
 However, Pimcore includes several third-party Open-Source libraries, 
 which are licensed under their own respective Open-Source licenses.
 
+The specific verions of a library are documented in
+[PIMCORE_MODIFICATIONS.md](bundles/AdminBundle/Resources/public/js/lib/PIMCORE_MODIFICATIONS.md).
```
dvesh3, line 10:
```suggestion
[PIMCORE_MODIFICATIONS.md](https://github.com/pimcore/pimcore/tree/11.x/bundles/AdminBundle/Resources/public/js/lib/PIMCORE_MODIFICATIONS.md).
```

mysliwietzflorian: @dvesh3 Is this actually a good idea? We would need to update that for every future branching change.

The way I wrote it should work for future changes where the path has not changed and forked repos, etc. WDYT?

dvesh3: yes, you are right. I misunderstood it with docs so should be fine like this.

dvesh3:
@mysliwietzflorian thanks!

PR #12825, opened 2022-08-04
Fix link to the ExtJS docs

jdreesen:
> **Note**: 
Apart from that: the `Edit on GitHub` link is broken, because it points to the `10.x` branch, which doesn't exist anymore.


aryaantony92:
@jdreesen Thanks a lot for the fix.

PR #12826, opened 2022-08-04
Remove deprecated JS functions

blankse:
Follow up to https://github.com/pimcore/pimcore/pull/12771 for Pimcore 11


robertSt7:
@blankse Thanks for the PR

PR #12831, opened 2022-08-05
Fix VR video viewer

spotbot2k:
Fixes the VR Video preview by setting its source to the one of the normal player

## Additional info  
Pushing it to 11.x as nobody seems to use the feature anyway and it is not urgent in any way


bundles/AdminBundle/Resources/public/js/pimcore/asset/video.js
```diff
@@ -338,10 +338,10 @@ pimcore.asset.video = Class.create(pimcore.asset.asset, {
         var previewContainerId = 'pimcore_video_preview_vr_' + this.id;
         this.previewPanel.update('<div id="' + previewContainerId + '" class="pimcore_asset_image_preview"></div>');
         var vrView = new VRView.Player('#' + previewContainerId, {
-            video: this.data['videoInfo']['previewUrl'],
+            video: '/admin/asset/serve-video-preview?id=' + this.data.id,
```
brusch:
```suggestion
            video: Routing.generate('pimcore_admin_asset_servevideopreview', {id: this.data.id}),
```
We should use proper route generation here.

brusch (changes requested):

bundles/AdminBundle/Resources/public/js/pimcore/asset/video.js
spotbot2k: Indeed. And I've changed the size and aspect ratio of the preview to the more common 16:9.

brusch:
Thank you very much!

PR #12836, opened 2022-08-08
DocType/Staticroute: module and action are not allowed properties anymore

blankse:
Followup to https://github.com/pimcore/pimcore/pull/7536


aryaantony92:
@blankse Thank you !

PR #12839, opened 2022-08-08
[Version preview] Prevent E_DEPRECATED error when having input field = null

BlackbitDevs:
All fields get initialized with `null`. For text fields in object preview this causes an 
> E_DEPRECATED: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated 
in PHP 8.1 because https://github.com/pimcore/pimcore/blob/03508497d3fcb22d8c44f0d56e487136159912a2/models/DataObject/ClassDefinition/Data/Extension/Text.php#L61 then calls `htmlspecialchars(null)`.


mcop1:
Thanks!

PR #12841, opened 2022-08-08
Fix Bug: .m4a files are not handled as audio files

### LINKED ISSUE #12838: [Bug]: .m4a files are not handled as audio files
### Expected behavior

Handle as audio file
- show audio player in preview
- don't break the worker (messenger) process

### Actual behavior

It looks like it is trying to handle the file as a video file:
`Pimcore\Model\Asset\Video\ImageThumbnail::__toString(): Return value must be of type string, null returned`

It also create an infinite loop in the worker process:
`[pimcore] Image thumbnail not yet available, processing is done asynchronously.`

### Steps to reproduce

Upload a .m4a file to Pimcore Assets.

blankse:
Fix https://github.com/pimcore/pimcore/issues/12838

### DIFF
--- a/bundles/AdminBundle/Resources/public/css/icons.css
+++ b/bundles/AdminBundle/Resources/public/css/icons.css
@@ -654,7 +654,7 @@
 }
 
 .pimcore_icon_mp3, .pimcore_icon_aac, .pimcore_icon_flac, .pimcore_icon_ogg, .pimcore_icon_wav, .pimcore_icon_wma,
-.pimcore_icon_ogg {
+.pimcore_icon_ogg, .pimcore_icon_m4a {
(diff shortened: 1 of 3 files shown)


kingjia90:
PR is fine and ready to be merged. 

Can you actually play the `m4a` file? On my end it is returning a 404 error  `No route found for XYZ` on each m4a file, but is fine  with `mp3` files.

blankse:
Yes my Chrome browser do play the m4a files.

blankse:
@kingjia90 Do you use nginx? Seems there is a whitlist in the config. With apache all files that exists works: https://pimcore.com/docs/pimcore/10.3/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/Nginx_Configuration.html
I can add m4a to this whitelist

kingjia90:
@blankse ahh i see, that's right, i am on nginx. Thank you for the fix!

PR #12842, opened 2022-08-08
[Improvement]: Remove Wkhtmltopdf and Wkhtmltoimage integrations

### LINKED ISSUE #12432: [Improvement]: Remove Wkhtmltopdf integrations
### Improvement description

Follow up to #12431

kingjia90:
Resolves #12432


kingjia90:
TODO: Need to provide migration tool to remove 
`wkhtmltopdfBin`, `wkhtml2pdfOptions`, `wkhtml2pdfHostname` config params from `settings_store`:`web_to_print` and remove it from demos dump, eg. https://github.com/pimcore/demo/blob/58247933cd5e3812e29cd20e0883ecfca038b5ad/dump/data-1-settings_store.sql#L82

brusch:
@kingjia90 I'm not sure if we really need a migration for that ... it shouldn't hurt/break if the config is still there and as soon as you save it again, it will kick the legacy keys, right? 
But of course we should remove it from the demo data 😊

bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
```diff
@@ -1166,13 +1165,9 @@ public function diffVersionsAction(Request $request, $from, $to)
 
         $viewParams = [];
 
-        if (Chromium::isSupported()) {
-            $tool = Chromium::class;
-        } else {
-            $tool = HtmlToImage::class;
-        }
+        $tool = Chromium::class;
```
blankse:
```suggestion
```

bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
```diff
@@ -1166,13 +1165,9 @@ public function diffVersionsAction(Request $request, $from, $to)
...
 
-        /** @var Chromium|HtmlToImage $tool */
+        /** @var Chromium $tool */
         $tool::convert($fromUrl, $fromFile);
         $tool::convert($toUrl, $toFile);
```
blankse:
```suggestion
        Chromium::convert($fromUrl, $fromFile);
        Chromium::convert($toUrl, $toFile);
```

models/Document/Service.php
```diff
@@ -669,25 +668,18 @@ public static function generatePagePreview($id, $request = null, $hostUrl = null
 
         File::mkdir(dirname($file));
 
-        $tool = false;
-        if (Chromium::isSupported()) {
-            $tool = Chromium::class;
-        } elseif (HtmlToImage::isSupported()) {
-            $tool = HtmlToImage::class;
-        }
+        $tool = Chromium::class;
```
blankse:
```suggestion
```

models/Document/Service.php
```diff
@@ -669,25 +668,18 @@ public static function generatePagePreview($id, $request = null, $hostUrl = null
...
 
-        if ($tool) {
-            /** @var Chromium|HtmlToImage $tool */
-            if ($tool::convert($url, $tmpFile)) {
-                $im = \Pimcore\Image::getInstance();
-                $im->load($tmpFile);
-                $im->scaleByWidth(800);
-                $im->save($file, 'jpeg', 85);
+        /** @var Chromium $tool **/
+        if ($tool::convert($url, $tmpFile)) {
```
blankse:
```suggestion
        if (Chromium::convert($url, $tmpFile)) {
```

blankse (changes requested):
You don't need the $tool variable anymore. We can use the Chromium class directly

kingjia90:
@brusch ah ok, i've checked and confirmed 👍🏻 it doesn't hurt and cleans on save but it would be loaded like this 
image
which is the former value for `WkHtmlToPdf(Deprecated)` option and will stay like this until one of the valid options is picked, but i guess it would be up to the devs anyway to make sure the setup the right tool for pdf generation and the configs are properly save/set, right?

PR #12846, opened 2022-08-08
[Bug]: Fix Video Datatype not supporting allowedTypes

### LINKED ISSUE #12832: [Bug]: Asset video open link
### Expected behavior

Created new video attribute for class. On creating data object for the video attributes,

When i click on the video edit icon, Video panel pop up should come.
   

### Actual behavior

On the video attributes, When i click on the video edit icon, The Video editor panel not opened.
"Uncaught TypeError: Cannot read properties of undefined (reading 'includes')  at pimcore.helpers.editmode.openVideoEditPanel"

'if (allowedTypes.includes("asset")) {
        fieldPath.on("render", initDD);
        poster.on("render", initDD);
    }"

### Steps to reproduce

1.Created new video attribute for class. 
2.Create a new data object.,
3. click on the video edit icon on the video attribute

kingjia90:
Resolves #12832

## Additional info  

The problem was caused by the introduction of this feature `allowedTypes` in Video Editables introduced with
https://github.com/pimcore/pimcore/pull/12445 where `openVideoEditPanel` got changed along the editable
https://github.com/pimcore/pimcore/blob/1c3b20e530e9568f9d80bf8fa5c8b32b5400ac24/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/video.js#L58 
but didn't cover the [video datatype](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Data_Types/Video_Type.html) which uses the same helper https://github.com/pimcore/pimcore/blob/b675696e8c3dff29e51d97e67fc79d3003c9ca3a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/video.js#L138

even though it can look like an improvement by adding this
image
 it should be a bug fix 😄


kingjia90:
TODO: The constants need to be refactored to avoid duplicate code

aashan10 (approved):

models/DataObject/ClassDefinition/Data/Video.php
```diff
@@ -27,6 +27,18 @@ class Video extends Data implements ResourcePersistenceAwareInterface, QueryReso
     use Extension\ColumnType;
     use Extension\QueryColumnType;
 
+    public const TYPE_ASSET = 'asset';
+    public const TYPE_YOUTUBE = 'youtube';
+    public const TYPE_VIMEO = 'vimeo';
+    public const TYPE_DAILYMOTION = 'dailymotion';
+
+    public const ALLOWED_TYPES = [
```
brusch: Don't think we need this constant, because we're not reusing this anyway.

models/DataObject/ClassDefinition/Data/Video.php
```diff
@@ -70,6 +82,20 @@ class Video extends Data implements ResourcePersistenceAwareInterface, QueryReso
...
+     * @var array
+     */
+    public $allowedTypes;
+
+    /**
+     * @internal
+     *
+     * @var array
+     */
+    public $videoTypes = self::ALLOWED_TYPES;
```
brusch: This feels like a workaround, probably it's better to use `FieldDefinitionEnrichmentInterface`
See https://github.com/pimcore/pimcore/blob/6245efe3be3d64ca5f3369a8472533f98d73043e/models/DataObject/ClassDefinition/Data/Multiselect.php#L713 for an example.

brusch (dismissed):

PR #12851, opened 2022-08-09
[Admin] Image HotspotItem - Rewrite Markers/Hotspots relation data

### LINKED ISSUE #12835: [Bug]: Markers/Hotspots lose relation data
### Expected behavior

Data stays.

### Actual behavior

Data is lost.

### Steps to reproduce

I can confirm, that the dialog can now be opened. However, when using assets, the path is lost, instead the ID is set.

Reproduce like this:

1. Add Asset relation in marker/hotspot
2. Drag & Drop Asset onto relation field
image
3. Save everything an reload the document
4. Reopen the marker/hotspot and find the an ID is now displayed of the path
image

_Originally posted by @passioneight in https://github.com/pimcore/pimcore/issues/12777#issuecomment-1206113519_

dvesh3:
Resolves #12835

### DIFF
--- a/models/Document/Editable/Image.php
+++ b/models/Document/Editable/Image.php
@@ -176,6 +176,11 @@ class Image extends Model\Document\Editable implements IdRewriterInterface, Edit
                                 $metaData = get_object_vars($metaData);
                             }
 
+                            if (in_array($metaData['type'], ['object', 'asset', 'document'])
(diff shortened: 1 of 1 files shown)


passioneight:
@dvesh3 I can confirm that this fixes the issue.

PR #12854, opened 2022-08-09
[Bug] getTotalCount() on listings should return correct count on grouped queries

youwe-petervanderwal:
## Error:
When a `GROUP BY` was added to a listing via the `onCreateQueryBuilder` callback, the `$listing->getTotalCount()` returned an incorrect/unexpected row count.

## Steps to reproduce:
```
// A listing we have to retrieve all the assets that have the tag with id 2 and/or the tag with id 7:
$requireOneOfTags = [2, 7];

$listing = (new Asset\Listing())->setCondition('assets.type = ?', ['image']);
$listing->getDao()->onCreateQueryBuilder(function (QueryBuilder $queryBuilder) use ($requireOneOfTags) {
    $expressionBuilder = $queryBuilder->expr();

    // Require assets to have one of the tags
    $queryBuilder
        ->innerJoin(
            'assets',
            'tags_assignment',
            'ta',
            $expressionBuilder->and(
                $expressionBuilder->in('ta.tagid', $requireOneOfTags),
                $expressionBuilder->eq('ta.ctype', $expressionBuilder->literal('asset')),
                $expressionBuilder->eq('ta.cid', 'assets.id')
            )
        )
        ->groupBy('assets.id')
    ;
});

$totalCount = $listing->getTotalCount();
// Executed query:
//     SELECT COUNT(*) FROM assets INNER JOIN tags_assignment ta ON (ta.tagid IN (2, 7)) AND (ta.ctype = 'asset') AND (ta.cid = assets.id) WHERE assets.type = ? GROUP BY assets.id
// Returned data by query
//     For each asset, the amount of tags_assignment
// Returned data by method
//     For the first asset in the rowset, the amount of tags_assignment
// Expected data
//     The number of assets
dump($totalCount); // 1

$realCount = count($listing->getAssets());
dump($realCount); // 3

// After fix:
$totalCount = $listing->getTotalCount();
// Executed query:
//     SELECT COUNT(*) FROM (SELECT * FROM assets INNER JOIN tags_assignment ta ON (ta.tagid IN (2, 7)) AND (ta.ctype = 'asset') AND (ta.cid = assets.id) WHERE assets.type = ? GROUP BY assets.id) as XYZ
// Returned data by query
//     The number of rows that would be returned by the original query = the number of assets
// Returned data by method
//     The number of assets
dump($totalCount); // 3
```

### Background info:
In Pimcore <6.9 we could fix that within the `onCreateQueryBuilder` callback itself as the `COUNT(*)` expression was already added before executing the callback. So our callback method was a bit more complex, checking on the `COUNT(*)` and rewriting that to a `COUNT(DISTINCT assets.id)` whithin the callback itself. However in later versions the callback is executed first and the `COUNT(*)` is attached to the query after the callback - so it's more important to us to fix this within Pimcore itself.

### DIFF
--- a/lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
+++ b/lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
@@ -16,7 +16,6 @@
 namespace Pimcore\Model\Listing\Dao;
 
 use Doctrine\DBAL\Query\QueryBuilder;
-use Doctrine\DBAL\Query\QueryBuilder as DoctrineQueryBuilder;
 use Pimcore\Model\DataObject;
 
 /**
@@ -29,19 +28,12 @@ trait QueryBuilderHelperTrait
      */
     protected $onCreateQueryBuilderCallback;
 
-    /**
-     * @param callable|null $callback
-     */
     public function onCreateQueryBuilder(?callable $callback): void
     {
         $this->onCreateQueryBuilderCallback = $callback;
     }
 
-    /**
-     * @param DoctrineQueryBuilder $queryBuilder
-     * @param bool $join
-     */
-    protected function applyListingParametersToQueryBuilder(QueryBuilder $queryBuilder, bool $join = false): void
+    protected function applyListingParametersToQueryBuilder(QueryBuilder $queryBuilder): void
     {
         $this->applyConditionsToQueryBuilder($queryBuilder);
         $this->applyGroupByToQueryBuilder($queryBuilder);
@@ -120,16 +112,13 @@ trait QueryBuilderHelperTrait
         }
     }
 
-    /**
-     * @param DoctrineQueryBuilder $queryBuilder
-     */
     private function applyLimitToQueryBuilder(QueryBuilder $queryBuilder): void
     {
         $queryBuilder->setFirstResult($this->model->getOffset());
         $queryBuilder->setMaxResults($this->model->getLimit());
     }
 
-    protected function prepareQueryBuilderForTotalCount(DoctrineQueryBuilder &$queryBuilder): void
+    protected function prepareQueryBuilderForTotalCount(QueryBuilder &$queryBuilder): void
     {
         $originalSelect = $queryBuilder->getQueryPart('select');
         $queryBuilder->select('COUNT(*)');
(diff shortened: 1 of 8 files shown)


lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
```diff
@@ -120,45 +112,34 @@ private function applyOrderByToQueryBuilder(QueryBuilder $queryBuilder): void
...
-     * @param DoctrineQueryBuilder $queryBuilder
-     */
     private function applyLimitToQueryBuilder(QueryBuilder $queryBuilder): void
     {
         $queryBuilder->setFirstResult($this->model->getOffset());
         $queryBuilder->setMaxResults($this->model->getLimit());
     }
 
-    protected function prepareQueryBuilderForTotalCount(DoctrineQueryBuilder &$queryBuilder): void
+    protected function prepareQueryBuilderForTotalCount(QueryBuilder $queryBuilder, string $identifierColumn): void
```
youwe-petervanderwal, line 121: Note: this will create a backwards incompatibility when this trait is used within application landscape. If you prefer, I could rewrite this to a runtime error instead

```
protected function prepareQueryBuilderForTotalCount(QueryBuilder $queryBuilder, ?string $identifierColumn = null): void
{
    // ...
    } elseif ($this->isQueryBuilderPartInUse($queryBuilder, 'distinct')) {
        if ($identifierColumn === null) {
            throw new \InvalidArgumentException(...);
        }
        // ...
    }
}
```

brusch: `QueryBuilderHelperTrait` is `@internal`, so should be fine ...

youwe-petervanderwal: I suddenly notice I made a fuck up here anyway (`$queryBuilder` was passed as reference with a good reason), will have a look tomorrow with some fresh energy

youwe-petervanderwal: Updated my PR, added unit test as well so we have this part covered in the future

lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
```diff
@@ -29,19 +28,12 @@ trait QueryBuilderHelperTrait
      */
     protected $onCreateQueryBuilderCallback;
 
-    /**
-     * @param callable|null $callback
-     */
```
dvesh3, line 34: why removed the phpdocs?

lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
```diff
@@ -120,45 +112,41 @@ private function applyOrderByToQueryBuilder(QueryBuilder $queryBuilder): void
...
-     * @param DoctrineQueryBuilder $queryBuilder
-     */
     private function applyLimitToQueryBuilder(QueryBuilder $queryBuilder): void
     {
         $queryBuilder->setFirstResult($this->model->getOffset());
         $queryBuilder->setMaxResults($this->model->getLimit());
     }
 
-    protected function prepareQueryBuilderForTotalCount(DoctrineQueryBuilder &$queryBuilder): void
+    protected function prepareQueryBuilderForTotalCount(QueryBuilder $queryBuilder, string $identifierColumn): void
```
dvesh3, line 121: Does this work without passing as reference? as nothing is returned from the method.

dvesh3:
@youwe-petervanderwal please see my comments above. thanks!

lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
youwe-petervanderwal: Redundant as it contains exactly the same information as already stated in the method's signature. I'm actually used to remove such duplicate info (within our company it's best practice to do so), so probably did this from muscle memory. Happy to revert this if you prefer

lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
youwe-petervanderwal: Yes, the QueryBuilder object itself is updated (instead of updating the argument value to a string which was the previous approach)

```
$queryBuilder
                ->resetQueryParts()
                ->select('COUNT(*)')
                ->from('(' . $innerQuery . ')', 'XYZ')
            ;
```

youwe-petervanderwal:
> @youwe-petervanderwal please see my comments above. thanks!

I somehow (accidentally) added my replies as pending review so they probably weren't visible to you, I re-added them here

lib/Model/Listing/Dao/QueryBuilderHelperTrait.php
dvesh3: I got your point now. Let's keep it this way 😉

dvesh3 (approved):

PR #12862, opened 2022-08-10
Fix PhpDocs

blankse:


bundles/AdminBundle/Security/Authenticator/AdminAbstractAuthenticator.php
```diff
@@ -131,7 +132,7 @@ public function onAuthenticationSuccess(Request $request, TokenInterface $token,
     }
 
     /**
-     * @param $user
+     * @param User|null $user
```
dvesh3:
```suggestion
     * @param User $user
```
passing null has no effect so I would avoid that.

dvesh3:
```suggestion
     * @param User $user
```
passing null has no effect so I would avoid that. Better adjust the if check.

models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -46,35 +46,35 @@ class Block extends Data implements CustomResourcePersistingInterface, ResourceP
      *
      * @var bool
      */
-    public $lazyLoading;
+    public $lazyLoading = false;
```
dvesh3: BC Break? as now it is always `bool` as oppose to `null` on initialize before:

```
$block = new DataObject\ClassDefinition\Data\Block();
if ($block->getLazyLoading() === null) {
    //do something
}
```

models/DataObject/ClassDefinition/Data/FieldDefinitionEnrichmentInterface.php
```diff
@@ -27,7 +27,7 @@ interface FieldDefinitionEnrichmentInterface
      *
      * @param array $context
      *
-     * @return Data
+     * @return $this
      */
-    public function enrichFieldDefinition(/** array */ $context = []) /** : Data */;
+    public function enrichFieldDefinition(/** array */ $context = []) /** : static */;
```
dvesh3, line 32:
```suggestion
    public function enrichFieldDefinition(/** array */ $context = []) /** : self */;
```

models/DataObject/ClassDefinition/Data/Fieldcollections.php
```diff
@@ -46,7 +46,7 @@ class Fieldcollections extends Data implements CustomResourcePersistingInterface
      *
      * @var bool
      */
-    public $lazyLoading;
+    public $lazyLoading = false;
```
dvesh3: BC Break? see above statement

models/DataObject/ClassDefinition/Data/Input.php
```diff
@@ -103,14 +103,14 @@ class Input extends Data implements
      *
      * @var bool
      */
-    public $unique;
+    public $unique = false;
```
dvesh3: BC Break? see above statement

bundles/AdminBundle/Security/Authenticator/AdminAbstractAuthenticator.php
blankse: Do you also mean we can change it to this in Pimcore 11?
```
    protected function saveUserToSession(User $user): void
    {
        if (Authentication::isValidUser($user->getUser())) {
```

The parameter types should not be what has effect, but what types may be passed.

dvesh3: yes. this should require `User` type to be passed on Pimcore 11.

blankse: I changed it: https://github.com/pimcore/pimcore/pull/12862/commits/ddcad9d2c3b53614b169de3d18900c948285172a

models/DataObject/ClassDefinition/Data/FieldDefinitionEnrichmentInterface.php
blankse: Returns always `$this`, so it should be static return type

models/DataObject/ClassDefinition/Data/Block.php
blankse: I don't think so. The PhpDocs says it is a bool, so its a bug to return `null`. I also see no use case to do compare this with `null`.

dvesh3: Agree, the example is not correct. However, these phpdocs are not completely right at the moment and that's what we are fixing in this PR. 

Also keeping BC promise in mind, we should not change the property values in bugfix release.

I would simply move these changes to a separate PR for Pimcore 11, where it is allowed to introduce breaking changes. Could you please take care of it?

blankse: Okay I reverted it and create a PR for the 11.x branch: https://github.com/pimcore/pimcore/pull/12999

dvesh3 (approved):

dvesh3:
@blankse thanks a lot!

Issue #11012, opened 2021-12-14
[Ecommerce] Pricing Manager - Replace SessionInterface injection


dvesh3:
### Improvement description

Please see https://github.com/pimcore/pimcore/pull/10909/files#diff-9a89925675cb58ed54af98157193e012e7badaea0de66c9beae61d5cfdf9f162R51-R54

PR #12867, opened 2022-08-10
Removed session configurators

dvesh3:
## Changes in this pull request  
Resolves #12472

PR #12873, opened 2022-08-11
[Ecommerce] Pricing Manager - Remove  Token condition & session getter/setter from Environment

### LINKED ISSUE #12866: [Ecommerce] Pricing Manager - Remove Token condition
### Feature description

Follow up to https://github.com/pimcore/pimcore/pull/12294/files#diff-9b8cc03e549f3dfc10262b50715c44800ac96aca0d75ea518a50943c0bae3e1d

consider also https://github.com/pimcore/pimcore/pull/12294/files#diff-1db6ed94371e3fd8636c7769a4f20ff8a5b45c524dfa7c43c422bbfa249afcf2

dvesh3:
Resolves #12866 & #11012


kingjia90 (changes requested):
Should we also remove [Tools/SessionConfigurator.php](https://github.com/pimcore/pimcore/blob/ecommerce_pricingmanager_removals/bundles/EcommerceFrameworkBundle/Tools/SessionConfigurator.php) , the Interface and Service as well here or will it be covered in a separate PR?

There's a problem when you don't remove all the `Token` conditions before upgrading
image
that breaks the interface and you are not able to modify/edit anymore. In the DB is a pain to fix because it's serialized, so one must downgrade and fix or remember to do the task beforehand. 
Not a big deal but can we improve this? I would suggest in a way that it "self-cleans" Token condition on next save and generally speaking not crashing when included Class is not found

dvesh3:
> Should we also remove [Tools/SessionConfigurator.php](https://github.com/pimcore/pimcore/blob/ecommerce_pricingmanager_removals/bundles/EcommerceFrameworkBundle/Tools/SessionConfigurator.php) , the Interface and Service as well here or will it be covered in a separate PR?
> 
> There's a problem when you don't remove all the `Token` conditions before upgrading image that breaks the interface and you are not able to modify/edit anymore. In the DB is a pain to fix because it's serialized, so one must downgrade and fix or remember to do the task beforehand. Not a big deal but can we improve this? I would suggest in a way that it "self-cleans" Token condition on next save and generally speaking not crashing when included Class is not found

yes, `SessionConfigurator` is removed in a separate PR [here](https://github.com/pimcore/pimcore/pull/12867).

Well, this is part of the upgrade preparation so we trigger the [deprecations](https://github.com/pimcore/pimcore/pull/12873/files#diff-9b8cc03e549f3dfc10262b50715c44800ac96aca0d75ea518a50943c0bae3e1dL38) 10.5.x and you should get rid of the deprecations before upgrading to major version otherwise you will face these exceptions more often.

kingjia90 (approved):

PR #12876, opened 2022-08-11

dasraab:
This PR gives you the option to get the public image thumbnail path with included frontend_prefixes even when you are in a console command. 

Resolves #12326

### DIFF
--- a/models/Asset/Image/Thumbnail.php
+++ b/models/Asset/Image/Thumbnail.php
@@ -54,7 +54,7 @@ final class Thumbnail
      *
      * @return string
      */
-    public function getPath($deferredAllowed = true, $cacheBuster = false)
+    public function getPath($deferredAllowed = true, $cacheBuster = false, $forceFrontend = false)
     {
         $pathReference = null;
         if ($this->getConfig()) {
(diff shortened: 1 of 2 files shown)


kingjia90:
@dasraab a gentle reminder to sign the CLA

brusch:
This should be already covered by https://github.com/pimcore/pimcore/pull/12657
Thanks anyway 😊

dasraab:
@brusch its not because you won't ever get the frontend path with frontend_prefixes in a console command. 

See [here](https://github.com/aashan10/pimcore/blob/ccb97544263c5d30ecab7daa4dad255cf27f84bf/models/Asset/Thumbnail/ImageThumbnailTrait.php#L319)

brusch:
@dasraab then we should fix `getFrontendPath()` to return the right value, I'd say 😊Can you create another PR for that?

dasraab:
@brusch that's what i planned to do in the first place but sadly this [line](https://github.com/pimcore/pimcore/blob/0984d48c8f629fedfab0f560f8ad37d6ab0de819/models/Asset/Thumbnail/ImageThumbnailTrait.php#L328) in convertToWebPath() is preventing the generation of frontend_paths with frontend_prefixes for console commands (it works fine as long its called via browser request)

i would love to submit a pr but right now i don't see a way of fixing this without also modifying getPath() and convertToWebPath()...

if you give me a hint on how you would like to have this fixed i'll give my very best to submit a PR

dasraab:
@brusch i gave it another try #12895

PR #12880, opened 2022-08-11
[Bug]: Asset Folder Preview tab supporting audio files and other files formats

### LINKED ISSUE #12808: [Bug]: Asset "Preview" tab pagination breaks for the folder which has audio files and other files (jpg,mp4 etc)
### Expected behavior

Items equal to the total value in pagination should be listed in preview tab.  System can display a fixed thumbnail for audio files.
So that the pagination total matches with the items count.

### Actual behavior

Asset preview tab does not list the audio files, but it does list the other files.
Because of this, the total value in pagination toolbar does not match with the count of items shown on the page.

### Steps to reproduce

Create a new folder in Assets.
Add an audio file in folder
Add a jpg/mp4/png file in the folder
Click on folder
Notice the items displayed on the tab and the total count displayed on the pagination.

audio_preview_not_supported

kingjia90:
Resolves https://github.com/pimcore/pimcore/issues/12808

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -833,7 +833,9 @@ class AssetController extends ElementControllerBase implements KernelControllerE
             return $this->generateUrl('pimcore_admin_asset_getdocumentthumbnail', $params);
         }
 
-        return null;
+        if ($asset instanceof Asset) {
+            return $asset->getFallbackThumbnailUrl();
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -833,7 +833,9 @@ protected function getThumbnailUrl(Asset $asset, array $params = [])
             return $this->generateUrl('pimcore_admin_asset_getdocumentthumbnail', $params);
         }
 
-        return null;
+        if ($asset instanceof Asset) {
+            return $asset->getFallbackThumbnailUrl();
```
brusch: As this is entirely admin functionality, I'd just put the whole logic here. Imho the icon image shouldn't be defined in the model.

brusch (changes requested):

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -833,7 +833,13 @@ protected function getThumbnailUrl(Asset $asset, array $params = [])
             return $this->generateUrl('pimcore_admin_asset_getdocumentthumbnail', $params);
         }
 
-        return null;
+        if ($asset instanceof Asset\Folder) {
```
```suggestion
        if ($asset instanceof Asset\Audio) {
```

brusch (dismissed):

aashan10 (approved):

PR #12881, opened 2022-08-12
[Data Objects]: Remove "generate type declarations" in class definitions

### LINKED ISSUE #12468: [Data Objects]: Remove "generate type declarations" in class definitions
### Improvement description

Follow up to #12467

robertSt7:
resolve #12468


tests/Model/DataType/LinkTest.php
```diff
@@ -80,21 +80,4 @@ public function testSave()
...
-     * @throws \Exception
-     */
-    public function testCheckValidity()
-    {
-        $linkObject = $this->createLinkObject();
-        $linkObject->setTestlink('https://pimcore.com/');
-        $linkObject->setLtestlink('https://pimcore.com/');
-
-        $this->expectException(ValidationException::class);
-        $this->expectExceptionMessage('Expected DataObject\\Data\\Link or null');
```
blankse, line 96: We can test if the TypeError will correct thrown.
```
        try {
            $linkObject = $this->createLinkObject();
            $linkObject->setTestlink('https://pimcore.com/');
            $linkObject->setLtestlink('https://pimcore.com/');
            $this->fail('Expected a TypeError');
        } catch (\Throwable $e) {
            $this->assertInstanceOf(\TypeError::class, $e);
        }
```

robertSt7: @blankse  Thank you, I didn't knew this workaround, but do you think we should test php behavior? Because the TypeError is not thrown by our code

blankse: So we can test the correct typehint. The typehint is set by our code.

blankse: Is also used in RgbaColor
https://github.com/pimcore/pimcore/blob/2ce8f9a7e2694b3a5b3f1b7e370dff150fe7cb8c/tests/_support/Helper/DataType/TestDataHelper.php#L1280-L1284

robertSt7: Thanks for mentioning

PR #12884, opened 2022-08-12

dasraab:
This PR gives you the option to get the public image thumbnail path with included frontend_prefixes even when you are in a console command.

Resolves #12326


kingjia90:
Closing, see https://github.com/pimcore/pimcore/pull/12876#issuecomment-1216310327 
It is the same changes as #12876 but in a different branch.

PR #12888, opened 2022-08-16
Fixes missing class reference in built Listing classes

NiklasBr:
Resolves "Undefined class 'ElementInterface'" in class definition files at `var/classes/DataObject/ClassName/Listing.php`

image

## After this change

image


mcop1:
Thanks!

dvesh3:
we need a migration here to fix the existing classes.

NiklasBr:
@dvesh3 I suppose I could re-save them like here?

https://github.com/pimcore/pimcore/blob/11.x/bundles/CoreBundle/Migrations/Version20220201132131.php

dvesh3:
> @dvesh3 I suppose I could re-save them like here?
> 
> https://github.com/pimcore/pimcore/blob/11.x/bundles/CoreBundle/Migrations/Version20220201132131.php

yes, like this..

PR #12895, opened 2022-08-16

dasraab:
i've moved the decision if its a frontend request or not to the getPath()-method with an option to force a frontend request.

now you can get a full frontend path with $asset->getThumbnail("dummy")->getFrontendPath() even if you are in a console command.

Version 1 of this PR see [12876](https://github.com/pimcore/pimcore/pull/12876)
Resolves [12326](https://github.com/pimcore/pimcore/discussions/12326)


brusch:
@dasraab thanks for the updated PR, however I'm still not entirely happy with changing the signature of `getPath()` without prior notice in v10. 
Maybe we should anyway think about changing to `getPath(array $params = [])` to be more flexible in the future 🤔

brusch:
@dasraab btw. please sign the CLA 😉

dasraab:
@brusch thats what i thought ... :P i'm also not really happy with that implementation but at least it has no breaking changes... getPath(array $params = []) would be a way better idea imo but i will also come with a lot of refactoring work. We could also go with getPath(...$params) and check if the first param is an array and if not we can extract the old params like $deferredAllowed and $cachBuster to still be backwards compatible? What do you think?

btw i've already signed the CLA but it doesn't show up as signed... idk why

brusch:
Thanks for updating the PR. 
Regarding BC-break: changing to `$args` shouldn't be a bc break as well, as the class itself is `final` so should be fine.

brusch:
We'll discuss this internally maybe it still needs some more adjustments, and probably also updating the docs and some more usages to the new style.

dasraab:
@brusch any updates on this topic?

brusch:
@dasraab unfortunately not yet, sorry - we'll let you know as soon as we have a decision. Thanks for your patience. 😊

dasraab:
@brusch still patiently waiting for this :P do you have any updates for me?

kingjia90:
I've took a quick look, the PR itself looks okay, but feels a little bit like a "workaround" 🤔 

What about instead just flip
https://github.com/pimcore/pimcore/blob/1e9f6135c0d112f0ae4cb7fc247a12d778847b32/models/Asset/Thumbnail/ImageThumbnailTrait.php#L302
to `!isFrontendRequestByAdmin()`? since the purpose of that line added on https://github.com/pimcore/pimcore/pull/9768 is to avoid an admin interface related problem, by using this condition, we just exclude that but allow frontend and CLI. 

This may work, as long as CLI is not considered true on `isFrontendRequestByAdmin` and as long as we don't want to specify this behaviour based on the given asset and want to keep it as a default rule.

dasraab:
@kingjia90 that solution would be perfectly fine for me! how do we proceed. do you want me to push a PR?

kingjia90:
> @kingjia90 that solution would be perfectly fine for me! how do we proceed. do you want me to push a PR?

Yes, please, but let's first have a confirmation from @brusch if that would make sense or if is better the more "flexible" `forceFrontend` approach.

brusch:
@kingjia90 nice idea, but I'm afraid that `isFrontendRequestByAdmin()` would throw an exception in CLI, because of https://github.com/pimcore/pimcore/blob/b6b7ed75190b9825d248dadb47d5516968669dfe/lib/Http/RequestHelper.php#L47

I'm fine with the `..$args` approach, but 
- [x] we should avoid duplicate code (extract the bc-stuff into an internal protected method) 
- [x] rebase onto `10.6` branch
- [x] issue deprecation when using old approach
- [x] add to upgrade notes
- [x] create follow up for `11.x` to remove the BC layer

dasraab:
@brusch i've moved my pimcore fork to another account. will issue a new pull request shortly. how do you want the protected method to be called? in which class do you want me to place it?

bitbirddev:
@brusch i've moved my pimcore fork to this account. will issue a new pull request shortly. how do you want the protected method to be called? in which class do you want me to place it?

brusch:
@bitbirddev I guess `ImageThumbnailTrait` would be a good place for it, right? It's used by all afftected methods.

PR #12897, opened 2022-08-17
[Bug]: Fallback for parentId not working anymore

### LINKED ISSUE #12892: [Bug]: Fallback for parentId not working anymore
### Expected behavior

If no parentId is set it should fallback to parentId = 1

### Actual behavior

Fallback isn´t working anymore, because id and parentId are null. And therefore the same.
Check for dataObjects, assets and documents.

### Steps to reproduce

Create new DataObject without parentId

```
        $dataObject = new DataObject\Product();
        $dataObject->setKey(\Pimcore\Model\Element\Service::getValidKey('New Name', 'object'));
        $dataObject->setName("test");
        $dataObject->save();
```
results in 

image

mcop1:
Added "null" check for element id and/or parent id

Resolves #12892

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
@@ -93,6 +93,7 @@ Please use [event listener](../../20_Extending_Pimcore/13_Bundle_Developers_Guid
       ...
   }
   ```
+- [Elements] Deprecated setting the parent id to 1 (root node), when an element with a non-existing parent id gets created.
 
 ## 10.4.2
 - When maintenance mode is active, all commands are prevented from starting (not just commands inheriting from `AbstractCommand`).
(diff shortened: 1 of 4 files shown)


brusch:
Was this possible before? 
Should it be really possible to save an element without a parent? 
Maybe we should just throw an exception with a more meaningful message, like `Unable to save element without a valid parent` ?

mcop1:
I assumed it was possible because of the "else" part here:
https://github.com/pimcore/pimcore/pull/12897/files#diff-867a334e4cc9a097b215afcb2b1f4949910cb2f2103a2cd997d8699d4d414fc9L646

If there is no parent set, it would fallback to parent id 1. That means the data object/asset/document would be saved under the root (home) element in the tree. 

But maybe we can discuss it in a few minutes :)

mcop1:
> I assumed it was possible because of the "else" part here: https://github.com/pimcore/pimcore/pull/12897/files#diff-867a334e4cc9a097b215afcb2b1f4949910cb2f2103a2cd997d8699d4d414fc9L646
> 
> If there is no parent set, it would fallback to parent id 1. That means the data object/asset/document would be saved under the root (home) element in the tree.
> 
> But maybe we can discuss it in a few minutes :)

Thanks for the discussion, what we will do: 

- [x] Revert changes and add a deprecation message for the else part
- [x] Create new pr for Pimcore 11 (improve exception message, add a new check for empty parent id, remove mentioned else part) -> https://github.com/pimcore/pimcore/issues/12902

PR #12905, opened 2022-08-17
Changing tab-bar color to make scrolling arrows visible

ChrisB9:

Changing the color for the tab-bar scrolling arrows, as they are currently invisible.

The 'bug' can be reproduced in both 10.5.2 and 11.x and looks on the demo system like this:
image

## Additional info  

Before:
image

After:
image


blankse:
@ChrisB9 The target branch for bug fixes should be 10.5, can you change it?

ChrisB9:
Yes i change it, I have to reopen it then, as just changing the main branch does not work :(

PR #12907, opened 2022-08-18
[Feature][DataObject] add title to class definition

### LINKED ISSUE #12647: Possibility to add an title to classdefinitions
### Feature description

Currently there is only a name for classdefinitions. With the name the classfile is generated and the name is currently used for translating the class in the backend.
It would be nice to have an additional field (title) where you can add a translation key like app.classes.product which is used for translations in your translation file.
If no title is available, the behavior stays the same.

I would provide a PR for this :)

dpfahlbusch:
Resolves #12647


models/DataObject/ClassDefinition.php
```diff
@@ -51,6 +51,13 @@ final class ClassDefinition extends Model\AbstractModel
      */
     public $name;
 
+    /**
+     * @internal
+     *
+     * @var string
+     */
+    public $title = '';
```
blankse:
```suggestion
    public string $title = '';
```

models/DataObject/ClassDefinition.php
```diff
@@ -1255,6 +1266,26 @@ public function getDescription()
         return $this->description;
     }
 
+    /**
+     * @param string $title
+     *
+     * @return $this
+     */
+    public function setTitle($title)
```
blankse:
```suggestion
    public function setTitle(string $title): static
```

models/DataObject/ClassDefinition.php
```diff
@@ -1255,6 +1266,26 @@ public function getDescription()
...
+    {
+        $this->title = $title;
+
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getTitle()
```
blankse:
```suggestion
    public function getTitle(): string
```

blankse (commented):
We should add typehints for new attributes/methods

dpfahlbusch:
Wasn't sure if this is already a thing :)
Thanks @blankse

dpfahlbusch:
Ready for review @fashxp

robertSt7:
@dpfahlbusch  Thanks for the PR.  It looks good, but could you also please add some documentation like: Add "title" field to classdefinitions, where you can add a translation key like app.classes.product which is used for translations in your translation file. Thank you

And what do you think about the use of the title in the following menu?
Screenshot (4)

dpfahlbusch:
@robertSt7 I will add that :)

dpfahlbusch:
@robertSt7 added documentation and translated the dropdown, so please review again :)

robertSt7:
@dpfahlbusch Thanks a lot for the PR :)

PR #12909, opened 2022-08-18
Fix video iframe embed attributes (permissions, data)

NiklasBr:
`allow="fullscreen"` is the recommended way and `allowFullScreen` should be all lower case

## Additional info  

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-allowfullscreen

### DIFF
--- a/models/Document/Editable/Video.php
+++ b/models/Document/Editable/Video.php
@@ -776,7 +776,7 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
             }
 
             $code .= '<div id="pimcore_video_' . $this->getName() . '" class="pimcore_editable_video '. ($config['class'] ?? '') .'">
-                <iframe src="https://player.vimeo.com/video/' . $vimeoId . '?dnt=1&title=0&amp;byline=0&amp;portrait=0'. $additional_params .'" width="' . $width . '" height="' . $height . '" frameborder="0" webkitAllowFullScreen mozallowfullscreen allowFullScreen></iframe>
+                <iframe src="https://player.vimeo.com/video/' . $vimeoId . '?dnt=1&title=0&amp;byline=0&amp;portrait=0'. $additional_params .'" width="' . $width . '" height="' . $height . '" frameborder="0" webkitAllowFullScreen mozallowfullscreen allowfullscreen allow="fullscreen"></iframe>
             </div>';
 
             return $code;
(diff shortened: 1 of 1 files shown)


models/Document/Editable/Video.php
```diff
@@ -701,7 +701,7 @@ private function getYoutubeCode()
         }
 
         $code .= '<div id="pimcore_video_' . $this->getName() . '" class="pimcore_editable_video '. ($config['class'] ?? '') .'">
-            <iframe width="' . $width . '" height="' . $height . '" src="https://www.youtube-nocookie.com/embed/' . $seriesPrefix . $youtubeId . $wmode . $additional_params .'" frameborder="0" webkitAllowFullScreen mozallowfullscreen allowFullScreen data-type="pimcore_video_editable"></iframe>
+            <iframe width="' . $width . '" height="' . $height . '" src="https://www.youtube-nocookie.com/embed/' . $seriesPrefix . $youtubeId . $wmode . $additional_params .'" frameborder="0" webkitAllowFullScreen mozallowfullscreen allowfullscreen allow="fullscreen" data-type="pimcore_video_editable"></iframe>
```
aryaantony92, line 704: Is there a reason for keeping the "allowfullscreen" attribute also? Because, as per the docs, "allowfullscreen" is a legacy attribute and it's redefined as allow = "fullscreen". 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-allowfullscreen

NiklasBr: Better legacy browser compatibility e.g. Edge 12-18 does not understand `allow="fullscreen"`. Better to avoid breaking anything for now. Especially since these methods are all `private` for some reason, so it cannot be overridden by those who may need to. I can make a second PR against 11.x with only the modern attributes perhaps?

NiklasBr: @aryaantony92 I see that only the YouTube editable has `data-type="pimcore_video_editable"` but not the others. Is this expected or have they become lost somewhere? Should I add them where missing?

aryaantony92: @NiklasBr Let me check it

aryaantony92: @NiklasBr The data-type="pimcore_video_editable" was added to fix an issue with the editing of inherited video editable. So, Could you please add it to the others too. Thanks

aryaantony92:
@NiklasBr Thanks a lot!

PR #12911, opened 2022-08-18
[TASK]: Remove Pimcore\Config\Config

### LINKED ISSUE #12477: Remove Pimcore\Config\Config
### Improvement description

Follow up to #12476

mcop1:
Resolves #12477 

- [x] Create a new issue in demo, because of the use of websiteConfig - https://github.com/pimcore/demo/issues/343


lib/Config.php
```diff
@@ -282,12 +280,9 @@ public static function getWebsiteConfig($language = null)
                 }
 
                 //TODO resolve for all langs, current lang first, then no lang
-                $config = new PimcoreConfig($settingsArray, true);
-
-                Cache::save($config, $cacheKey, $cacheTags, null, 998);
-            } elseif ($config instanceof PimcoreConfig) {
-                $data = $config->toArray();
-                foreach ($data as $key => $setting) {
+                Cache::save($settingsArray, $cacheKey, $cacheTags, null, 998);
```
kingjia90, line 283: image
Isn't here missing a `$config = $settingsArray;` (to be replacing former `$config = new PimcoreConfig($settingsArray, true);`) ?
Currently `$config` at 295 `self::setWebsiteConfig($config, $language);` is crashing due being a `bool` (`Cache::load()` at 224 returns [false](https://github.com/pimcore/pimcore/blob/14f44bb56238429b98eb7aeaa81f2b46b57f2a1f/lib/Cache/Core/CoreCacheHandler.php#L316) when not found) and the `return $config` at line 298 would be returning `bool` instead of expected `array`, everytime cache is cleared.

brusch: @robertSt7 @mcop1

kingjia90: I've opened https://github.com/pimcore/pimcore/pull/13032 to better manage and see this potential issue (and leaving alone the merged pr/archived task 😄 )

mcop1: Thanks for creating the pr, I assigned it to myself and I will have a look right away.

PR #12914, opened 2022-08-19
[Improvement][SystemSettings] Drag and Drop asset to pimcore backend login screen background

### LINKED ISSUE #12783: [Improvement][SystemSettings] Drag and Drop asset to pimcore backend login screen background
### Improvement description

As an admin I would like to Drag and Drop the asset of the pimcore backend login screen background image isntead of inserting the path.

Settings --> System Settings --> Custom Login Background Image

image

robertSt7:
resolve #12783


mcop1:
@robertSt7 The input field for the URL isn´t editable anymore. Therefore you can´t add a custom url, only assets. I think we should allow both options, because it was possible before.

PR #12921, opened 2022-08-19
Fix: ContentMasterDocument breaks when the Master Document gets deleted

### LINKED ISSUE #12915: [Bug]: ContentMasterDocument breaks when the Master Document gets deleted
I got this error after updating to 10.5.3, i had 10.4.2 before where the error was not appearing.

### Expected behavior

After deleting a page (A) that is the Content Master Document of another page (B), the other page (B) should open anyway or - event better - the "Content Master Document" of (A) should be removed from the field in page (B).

### Actual behavior

After deleting a page that is the Content Master Document of another page, the page could not be opend anymore and the following error is returned:

URL: /admin/page/get-data-by-id
Message: Call to a member function getId() on null

Thats because the ID of the Content Master Document does not exist anymore.

### Steps to reproduce

Tested in the Demo-Pimcore.

1. create a new page "test" under /en/
2. copy the page
3. paste the page as a new language variant under /de/
4. remove the page /en/test/
5. try to open /de/test/ and get the error

blankse:
Fix https://github.com/pimcore/pimcore/issues/12915

### DIFF
--- a/models/Document/PageSnippet.php
+++ b/models/Document/PageSnippet.php
@@ -436,12 +436,17 @@ abstract class PageSnippet extends Model\Document
             $validate = \func_get_args()[1] ?? false;
             $maxDepth = 20;
             $currentContentMasterDocument = Document::getById($contentMasterDocumentId);
-            do {
-                if ($currentContentMasterDocument->getId() === $this->getId()) {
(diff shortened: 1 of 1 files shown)


dvesh3 (approved):

aryaantony92:
@blankse Thank you for the fix

PR #12925, opened 2022-08-22
[PhpStan] Fix: InMemory::plainText() expects non-empty-string, string given

blankse:
```
 ------ -----------------------------------------------------------------------------------------------
  Line   lib/Targeting/Storage/Cookie/JWTCookieSaveHandler.php
 ------ -----------------------------------------------------------------------------------------------
  54     Parameter #1 $contents of static method Lcobucci\JWT\Signer\Key\InMemory::plainText() expects
         non-empty-string, string given.
 ------ -----------------------------------------------------------------------------------------------
```


brusch:
👏was also just about fixing it - thanks a lot 😊

PR #12929, opened 2022-08-22
[Improvement]: Add exception if no parent id was set

### LINKED ISSUE #12902: [Improvement]: Add exception if no parent id was set
### Improvement description

- [x] Add a new exception if no parent id was set for dataObjects, assets and documents.
- [x] Improve existing exception message for identical parent id and id
- [x] Add corresponding tests

See #12897

mcop1:
Resolves #12902


models/Asset.php
```diff
@@ -640,15 +644,6 @@ public function correctPath()
                 // use the parent's path from the database here (getCurrentFullPath), to ensure the path really exists and does not rely on the path
                 // that is currently in the parent asset (in memory), because this might have changed but wasn't not saved
                 $this->setPath(str_replace('//', '/', $parent->getCurrentFullPath() . '/'));
-            } else {
-                trigger_deprecation(
```
blankse, line 644: I think we should also throw an exception if the parent with the given ID doesn't exist.

models/Asset.php
```diff
@@ -627,8 +627,12 @@ public function correctPath()
                 throw new Exception("invalid filename '" . $this->getKey() . "' for asset with id [ " . $this->getId() . ' ]');
             }
 
+            if($this->getParentId() === null || $this->getParentId() === 0) {
```
```suggestion
            if(!$this->getParentId()) {
```
Should be sufficient here 😊

models/DataObject/AbstractObject.php
```diff
@@ -852,8 +852,12 @@ protected function correctPath()
                 throw new \Exception('invalid key for object with id [ '.$this->getId().' ] key is: [' . $this->getKey() . ']');
             }
 
+            if($this->getParentId() === null || $this->getParentId() === 0) {
```
brusch:
```suggestion
            if(!$this->getParentId()) {
```

models/Document.php
```diff
@@ -490,24 +490,19 @@ private function correctPath()
                 throw new \Exception('invalid key for document with id [ ' . $this->getId() . ' ] key is: [' . $this->getKey() . ']');
             }
 
+            if($this->getParentId() === null || $this->getParentId() === 0) {
```
brusch:
```suggestion
            if(!$this->getParentId()) {
```

brusch (changes requested):

models/Asset.php
brusch: Agree, we need to throw an exception here instead 😊

brusch (changes requested):

models/Asset.php
mcop1: You are right, added the exception + tests

PR #12933, opened 2022-08-22
[Bug]: Adding a new variant of an object which is using ClassificationStore Selects crashes

BentBr:
## Bug
If a user in Pimcore admin is adding a new variant of an object which is using ClassificationStore Selects it crashes due to a conversion error.
- Parent (and therefore the variants) must use ClassificationStore
- At least one Select field must be used
- (select values are being set via custom data provider - this might be optional)

Changes the "isEmpty" check for Select type such as arrays are being treated correctly.

Resolves the issue of users in admin are not able to add new variants thought the interface.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Select.php
+++ b/models/DataObject/ClassDefinition/Data/Select.php
@@ -378,12 +378,16 @@ class Select extends Data implements
     }
 
     /**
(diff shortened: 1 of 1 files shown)


robertSt7:
@BentBr Thanks for the PR

PR #12941, opened 2022-08-23
[Bug]: Class.create with undefined parameter crashes js

### LINKED ISSUE #12939: [Bug]: Class.create with undefined parameter crashes js
### Expected behavior

Show console error and continue other scripts

### Actual behavior

crashes js

### Steps to reproduce

Checkout 11.x branch

robertSt7:
resolve #12939

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/lib/class.js
+++ b/bundles/AdminBundle/Resources/public/js/lib/class.js
@@ -43,9 +43,13 @@ var Class = (function() {
             parent.subclasses.push(klass);
         }
 
-        for ( var i = 0; i < properties.length; i++)
+        for (let i = 0; i < properties.length; i++) {
+            if(!properties[i]){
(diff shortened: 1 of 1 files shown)


brusch:
That's actually caused by bundles extending from legacy class `pimcore.plugin.admin` which was removed for `11.x`. 
e.g.: 
https://github.com/pimcore/object-merger/blob/3.x/src/Resources/public/js/plugin.js#L17

I think this should be rather fixed there (by requiring `pimcore/pimcore: ^10.5` and moving to native events), instead of "ignoring" the error here.

mcop1:
@brusch yes, we figured that out, and there are the corresponding issues in the bundles (https://github.com/pimcore/object-merger/issues/13). But without this PR we can´t use e.g. the data-hub right now, because it breaks the javascript execution. 

But I think we should add a bit more information to the console log.

brusch:
Yep, agree 👍

Issue #12948, opened 2022-08-23
[Bug]: Missing return for non-admin users when renaming objects


QWpreischl:
### Expected behavior

- Non-Admin user renames object in backend
- Object gets renamed

### Actual behavior

- Non-Admin user renames object in backend
- Error message pops up "There was an error while renaming the item." 
- After closing the error message you can see that the renaming on the object worked

### Steps to reproduce

- Login as Non-Admin user
- Rename any object

What fixed the problem for me on Version 10.5.2, but also is still missing in 11.x:
In file:
https://github.com/pimcore/pimcore/tree/11.x/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
on Line 1101 there is a "return" missing before the adminJson function so the "$success = true" of the renameObject function is passed through to the response of the request:

```
} elseif ($key && $object->isAllowed('rename')) {
            return $this->adminJson($this->renameObject($object, $key));
        } else {
```

blankse:
@QWpreischl I created a PR: https://github.com/pimcore/pimcore/pull/12949

dvesh3:
Fixed by #12949

PR #11915, opened 2022-04-14
[AdminBundle] Refactor DataObject controllers to share more code

dvesh3:
## Changes in this pull request  
Resolves #6906

PR #12949, opened 2022-08-23
Fix: Missing return for non-admin users when renaming objects

blankse:
Fix https://github.com/pimcore/pimcore/issues/12948

Regression from https://github.com/pimcore/pimcore/pull/11915


dvesh3:
@blankse we can't thank you enough 💜 thanks a lot for your contributions!

PR #12954, opened 2022-08-24
fix: video in object cannot be edited

passioneight:
Specifically, if the class definition does not restrict the video type, the `allowedTypes` variable is **not** defined. Yet, it is followed by the call `allowedTypes.includes...` somewhere further down. This causes a JS exception, which means the user cannot edit the video due to the edit dialog not opening.

Caused by https://github.com/pimcore/pimcore/pull/12445
Similar to https://github.com/pimcore/pimcore/pull/12733, but the JS equivalent

image

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
@@ -2001,7 +2001,7 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
 
 pimcore.helpers.editmode.openVideoEditPanel = function (data, callback) {
 
-    const allowedTypes = data.allowedTypes;
+    const allowedTypes = data.allowedTypes || [];
     var window = null;
     var form = null;
     var fieldPath = new Ext.form.TextField({


passioneight:
Turns out, this is only part of the fix. While the dialog will open, one can neither choose the asset type nor drag & drop assets on the path field.

Our patch is this atm instead of the suggested change:

```js
const allowedTypes = data.allowedTypes || [
        "asset",
        "youtube",
        "vimeo",
        "dailymotion"
    ];
```

@dvesh3 seems to me that there should be a better place to define the allowed types. Can you point me to the file that needs these default values so I can update the PR?

dvesh3:
@passioneight could you please check again with #12846 & #12947? the problem should be already resolved by these PRs. thanks!

passioneight:
Thanks, will check soon :-)

passioneight:
- https://github.com/pimcore/pimcore/pull/12846 was already applied and did not change anything.
- https://github.com/pimcore/pimcore/pull/12947 was not yet applied, so I applied the changes, but it didn't solve the issue.

I wanted to try on Pimcore Demo, but I can't even add the `Video` type to the class definition:

image

image

Please have a look :-)

dvesh3:
@passioneight It works for me with latest changes on 10.5 branch. please make sure to clear cache when applying changes from #12947

The reason for not working on demo.fun is the demo is still using v10.5.3.

passioneight:
@dvesh3 so, the video was in a field collection. After manually saving the field collection itself, the console error was resolved. So, I guess reinstalling classes should work for anyone else coming across this issue.

However, there's still one more thing. I did not select any video types and would expect all types to be available, but I can't select anything:

image

dvesh3:
@passioneight please see #13035

dvesh3:
Closed in favor of #13035. thank you!

Issue #8465, opened 2021-03-15
500 Internal Server Error on Versions tab for calculated localized field


moiashvin:
# Bug Report

### Expected behavior
Display of details in Version tab correctly (including localized calculated fields)

### Actual behavior
Selecting Version tab where there is calculated localized field displays
Oops! An Error Occurred
The server returned a "500 Internal Server Error".
Something is broken. Please let us know what you were doing when this error occurred. We will fix it as soon as possible. Sorry for any inconvenience caused. 

### Steps to reproduce

Steps to reproduce:
(1) Open a Product - e.g. /Product Data/Accessories/rims/alfa romeo-1900-rims-simple
(Here the Versions tab is empty)

(2) Do a Save and Publish to save new version of product.
(Here, a new Version is created and listed in Versions tab)

(3) Select the Version. On field Fullname (en) in the table, the Value is:
Oops! An Error Occurred
The server returned a "500 Internal Server Error".
Something is broken. Please let us know what you were doing when this error occurred. We will fix it as soon as possible. Sorry for any inconvenience caused.

### Code example
```php
// your code
```

# Feature Request

weisswurstkanone:
fixed via #8475

PR #12963, opened 2022-08-24
[Bug]: datahub_context runtime cache value is reset by Pimcore\Model\Version::loadData

### LINKED ISSUE #12960: [Bug]: datahub_context runtime cache value is reset by Pimcore\Model\Version::loadData
### Expected behavior

Either the runtime cache value `datahub_context` should not get reset by loading version data or the datahub implementation should not rely on the value being available (see pimcore/data-hub#465).

### Actual behavior

An exception `No entry is registered for key 'datahub_context'` is raised because of these lines in `Pimcore\Model\Version`:

```PHP
//clear runtime cache to avoid dealing with marshalled data
Runtime::clear();
```

A quick, but ugly workaround would be to replace the second line with this:

`Runtime::clear(['datahub_context']);`

### Steps to reproduce

In a custom resolver, call `Pimcore\Model\Version::loadData()` on a `Version` of any `Document`.

For a very quick repro this would do the trick, I think. But I couldn't verify for now:

```PHP
<?php
namespace App\EventSubscriber;

use GraphQL\Type\Definition\ResolveInfo;
use GraphQL\Type\Definition\Type;
use Pimcore\Bundle\DataHubBundle\Event\GraphQL\Model\QueryTypeEvent;
use Pimcore\Bundle\DataHubBundle\Event\GraphQL\QueryEvents;
use Pimcore\Bundle\DataHubBundle\GraphQL\Resolver\QueryType;
use Pimcore\Bundle\DataHubBundle\GraphQL\Service;
use Pimcore\Cache\Runtime;
use Symfony\Component\EventDispatcher\EventSubscriberInterface;
use Symfony\Component\EventDispatcher\EventDispatcherInterface;

class QueryEventSubscriber implements EventSubscriberInterface
{
    protected EventDispatcherInterface $eventDispatcher;

    public function __construct(EventDispatcherInterface $eventDispatcher)
    {
        $this->eventDispatcher = $eventDispatcher;
    }

    public static function getSubscribedEvents(): array
    {
        return [
            QueryEvents::POST_BUILD => 'onPostBuild'
        ];
    }

    public function onPostBuild(QueryTypeEvent $event)
    {
        $graphQlService = $event->getQueryType()->getGraphQlService();

        $documentResolver = new LatestVersionQueryType($this->eventDispatcher);
        $documentResolver->setGraphQlService($graphQlService);

        $config['fields']['getDocument']['resolve'] = function($value = null, $args = [], $context = [], ResolveInfo $resolveInfo = null) use($documentResolver) {
            $document = $documentResolver->resolveDocumentGetter();

            $version = $document?->getLatestVersion();
            if( $version instanceof Version ) {
                // This triggers the bug, no need to return the Version.
                // Runtime::get('datahub_context') fails from this point onwards (i.e. when loading children)
                $version->loadData();
            }
            
            return $document;
        };

        $event->setConfig($config);
    }
}
```

Now, running a `getDocument` query which, for example, loads the children, _should_ produce the error. Again, I've not yet had time to test this but I hope it's clear what the issue is.

```grqphql
query {
    getDocument(id: 1) {
        id
        children {
            __typename
        }
    }
}
```

Related issues: pimcore/data-hub#465 pimcore/data-hub#546,

robertSt7:
resolves #12960

test also: https://github.com/pimcore/pimcore/issues/8465

### DIFF
--- a/models/Version.php
+++ b/models/Version.php
@@ -367,7 +367,7 @@ final class Version extends AbstractModel
 
             $data = Serialize::unserialize($data);
             //clear runtime cache to avoid dealing with marshalled data
-            RuntimeCache::clear();
+            \Pimcore::collectGarbage();
             if ($data instanceof \__PHP_Incomplete_Class) {
                 Logger::err('Version: cannot read version data from file system because of incompatible class.');
 


ThisIsJustARandomGuy:
@robertSt7 Unfortunately, I don't think this resolves the bug. `\Pimcore::collectGarbage()` still eventually calls `RuntimeCache::clear($protectedItems)` via the `\Pimcore\Helper\LongRunningHelper`. If you could modify the call to be `\Pimcore::collectGarbage(['datahub_context'])` it would work perfectly.

mcop1:
> @robertSt7 Unfortunately, I don't think this resolves the bug. `\Pimcore::collectGarbage()` still eventually calls `RuntimeCache::clear($protectedItems)` via the `\Pimcore\Helper\LongRunningHelper`. If you could modify the call to be `\Pimcore::collectGarbage(['datahub_context'])` it would work perfectly.

As I see it `Pimcore::CollectGarbage` calls `LongRunningHelper::cleanUp`, which calls `LongRunningHelper::cleanupPimcoreRuntimeCache` that ultimately results in a call to `RuntimeCache::Clear`, but it should be respecting the protected items such as datahub_context:

https://github.com/pimcore/pimcore/blob/15f4c8eb41a3e36c4456b4dd44a3f7e999698047/lib/Helper/LongRunningHelper.php#L109-L115

So it should be fine?

PR #12965, opened 2022-08-24
[Data objects] Support exporting empty class

BlackbitDevs:
Steps to reproduce bug:

1. Create data object class without any layout or data elements. Save.
2. Export class

You will get the error `Uncaught PHP Exception TypeError: "method_exists(): Argument #1 ($object_or_class) must be of type object|string, null given" at /var/www/html/vendor/pimcore/pimcore/models/DataObject/ClassDefinition/Service.php line 77 `

With this PR the class definition JSON will get exported.


robertSt7:
@BlackbitNeueMedien Thanks for the PR

PR #12969, opened 2022-08-24
reset classId for class other than existing class

### LINKED ISSUE #12968: [Bug]: ClassDefinitionManager fails to create multiple classes if id of the multiple definition files are same
### Expected behavior

In the case of two class definitions files with the same id for different classes. for example (Customer and Product):
- `definition_Customer.php`
```php
return Pimcore\Model\DataObject\ClassDefinition::__set_state(array(
   'dao' => NULL,
   'id' => '6',
   'name' => 'Customer',
   ...
```
- `definition_Product.php`
```php
return Pimcore\Model\DataObject\ClassDefinition::__set_state(array(
   'dao' => NULL,
   'id' => '6',
   'name' => 'Product',
   ...
```
**Expected Behaviour**
`bin/console pimcore:deployment:classes-rebuild -c`  to create two Data Objects Class 
- Customer
- Product

### Actual behavior

`bin/console pimcore:deployment:classes-rebuild -c` will first create the `Customer` class and then will update the same `Customer` class definition with the `Product` class definition, seeming as if it never created the `Customer` class but only `Product` class.

**Verbose output of `bin/console pimcore:deployment:classes-rebuild -c -v`**
```
---------------------
Saving all classes
Customer [6] created
Product [6] created
---------------------
```

### Steps to reproduce

- create Data object *Customer* via admin panel
- `cp defintion_Customer.php defintion_Product.php` and change `name` attribute of ``defintion_Product.php`` to `product`
- `bin/console pimcore:deployment:classes-rebuild -c`

sndsabin:
Resolves #12968

### DIFF
--- a/lib/Model/DataObject/ClassDefinition/ClassDefinitionManager.php
+++ b/lib/Model/DataObject/ClassDefinition/ClassDefinitionManager.php
@@ -71,8 +71,9 @@ class ClassDefinitionManager
                     $changes[] = [$class->getName(), $class->getId(), self::SAVED];
                     $existingClass->save(false);
                 } else {
-                    $changes[] = [$class->getName(), $class->getId(), self::CREATED];
+                    $class->setId('');
                     $class->save(false);
+                    $changes[] = [$class->getName(), $class->getId(), self::CREATED];
                 }
             }
         }


brusch:
Obsolete, see https://github.com/pimcore/pimcore/issues/12968#issuecomment-1230234838

Thanks anyway for your PR 😊

PR #12980, opened 2022-08-25
[BUG][DataObjects]: Pass condition variables types in ClassDefinition listing DAO

passioneight:
Not passing the condition variable types will cause an SQL error if the `ClassDefinition` listing is filtered like so:
`$classDefinitions->addConditionParam("id IN (?)", [['my_class_id']]);`


mcop1:
Thanks!

PR #12982, opened 2022-08-25
Removed deprecated and legacy iframe attributes, added title, allow, and data-type

NiklasBr:
11.x follow-up on #12909

`frameborder="0"` is often not required when using frameworks anyway: 

https://github.com/twbs/bootstrap/blob/edf9c40956d19e6ab3f9151bfe0dfac6be06fa21/site/content/docs/5.2/helpers/ratio.md?plain=1#L16


dvesh3:
@NiklasBr thanks for your PR! could you please resolve the conflicts?

models/Document/Editable/Video.php
```diff
@@ -1036,7 +1036,7 @@ private function getEmptyCode(): string
             $height .= 'px';
         }
 
-        return '<div id="pimcore_video_' . $this->getName() . '" class="pimcore_editable_video"><div class="pimcore_editable_video_empty" id="' . $uid . '" style="width: ' . $width . '; height: ' . $height . ';"></div></div>';
+        return '<div id="pimcore_video_' . $this->getName() . '" class="pimcore_editable_video"><div class="pimcore_editable_video_empty" id="' . $uid . '" style="width: ' . $width . '; height: ' . $height . ';" data-type="pimcore_video_editable"></div></div>';
```
aryaantony92: Thank you for the PR. But, the ´data-type` attribute (data-type="pimcore_video_editable") is required for videos inside the iframe only.

```suggestion
        return '<div id="pimcore_video_' . $this->getName() . '" class="pimcore_editable_video"><div class="pimcore_editable_video_empty" id="' . $uid . '" style="width: ' . $width . '; height: ' . $height . ';"></div></div>';
```

NiklasBr: This is from #12909

aryaantony92: Yes, we added it only for the video inside the iframe

models/Document/Editable/Video.php
```diff
@@ -1052,6 +1052,14 @@ private function updateAllowedTypesFromConfig(array $config): void
         }
     }
 
+    /**
+     * @return string
+     */
```
aryaantony92:
```suggestion

```

aryaantony92: This phpDoc can be removed since we have already typed the return type here https://github.com/pimcore/pimcore/pull/12982/files#diff-9cba4e2ecaf34598897940fdff3416069ea2708f77e953bf87f08b2529651007R1055-R1057:~:text=function%20getUniqId()%3A-,string,-%7B

NiklasBr: I was just following the code style such as 

https://github.com/pimcore/pimcore/blob/d49f5a5bc9c62a4c75808d877187ee67d36190c1/models/Document/Editable/Video.php#L1022-L1025

dvesh3: @NiklasBr yes, we decided to get rid off duplicate info recently, so we can already fix the changed files and later on rest of things will be taken care of. thanks!

aryaantony92:
@NiklasBr Thank you!

PR #12985, opened 2022-08-26
[BUG][DataObjects]: Renaming DataObjects doesn´t work

eliasfellner:
Currently renaming an object is not possible
line 1080 checks if the array has an indices key, but then it is immediately used it anyway

fix:
moved the setting of the $indexUpdate variable down into the if condition, so no error is thrown and it's set for the functions that need it


mcop1:
Thank you!

PR #12991, opened 2022-08-27
Use fullpath to fix formatter on readonly

### LINKED ISSUE #11232: [Bug]: PathFormatter class result not taken/visible on Reverse Object Relation field if not editable
### Expected behavior

On the backend ExtJS datagrid of a Reverse Object Relation field, the custom formatted result processed by a custom Formater Service class is not shown (still have the default 'fullpath' of the related reverse object)

### Actual behavior

Show the default "fullpath" of the reverse related object

### Steps to reproduce

Create a class (eg. Contact) that will be the linked reverse object in the the relation field. (add a "store" field Many To One relation)
Create a class (eg. Store) with a Reverse Object Relation field (contacts) and set it to "not editable", linked to Contact via Contact::store
Also set it a custom Formatter Service class (implementing the PathFormatterInterface)
Eg : 
```php
<?php


namespace App\PathFormatter;


use Pimcore\Model\DataObject\ClassDefinition\PathFormatterInterface;
use Pimcore\Model\DataObject\Contact;

class ContactPathFormatter implements PathFormatterInterface
{
    {
        foreach ($targets as $id => $item) {
            if ($item['type'] === 'object') {
                $targetObject = Concrete::getById($item['id']);
                if ($targetObject instanceof Contact) {
                    $result[$id] = sprintf('%s %s / %s / %s', $targetObject->getFirstname(), strtoupper($targetObject->getLastname()), $targetObject->getPhone(), $targetObject->getEmail());
                }
            }
        }

        return $result;
    }
}
```

Create a "Store" and linked "Contact" object
In the Store object, the "contacts" datagrid field (the reverse relation) list the contacts object but displays the default fullpath instead of the custom formatter string.

matthiashamacher:
Resolves #11232

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyObjectRelation.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyObjectRelation.js
@@ -316,7 +316,7 @@ pimcore.object.tags.manyToManyObjectRelation = Class.create(pimcore.object.tags.
         if(visibleFields.length === 0) {
             columns.push(
                 {text: 'ID', dataIndex: 'id', width: 50},
-                {text: t("reference"), dataIndex: 'fullpath', flex: 200, renderer:this.fullPathRenderCheck.bind(this)},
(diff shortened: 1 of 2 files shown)


mcop1:
Thank you

PR #12993, opened 2022-08-29
Fix: "Save only scheduled tasks" Button doesn't work

blankse:
1. Open a document in demo
2. Change anything
3. Go to Schedule tab
4. Create a new entry
5. Click "Save only scheduled tasks" Button
6. Reload
-> Scheduled tasks tab is empty

Regression from https://github.com/pimcore/pimcore/pull/12344


dvesh3 (approved):

dvesh3:
@blankse thanks!

PR #12997, opened 2022-08-30
Update Doctrine DBAL to V3

### LINKED ISSUE #10613: [Improvement]: Doctrine  DBAL v3

blankse:
Resolve https://github.com/pimcore/pimcore/issues/10613


aashan10:
@blankse please fix the merge conflicts

lib/Model/Listing/AbstractListing.php
```diff
@@ -44,9 +44,9 @@ abstract class AbstractListing extends AbstractModel implements \Iterator, \Coun
     protected $limit;
 
     /**
-     * @var int|null
+     * @var int
      */
-    protected $offset;
+    protected $offset = 0;
```
aashan10, line 49: @brusch does this fall under the backwards compatibility promise that we discussed earlier or not? This seems to be for pimcore 11, so I am a bit confused!

aashan10 (approved):

lib/Model/Listing/AbstractListing.php
dvesh3: We can introduce breaking changes since this is for Pimcore 11.

dvesh3:
@blankse thank you very much!

NiklasBr:
Great work! In a few days you do it again for version 4 😉 

https://github.com/doctrine/dbal/milestone/3

brusch:
We should definitely prepare already for 4 and check that we're not using any deprecated methods: 
https://github.com/doctrine/dbal/blob/4.0.x/UPGRADE.md

brusch:
Created a follow-up ticket: #13015

PR #13001, opened 2022-08-30
[Installer] Fix creating "system" user

dvesh3:
Always create a system user with `id = 0`, irrespective of data dump provided.

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20220830105212.php
+++ b/bundles/CoreBundle/Migrations/Version20220830105212.php
@@ -16,7 +16,7 @@ final class Version20220830105212 extends AbstractMigration
 
     public function up(Schema $schema): void
     {
-        $this->addSql("INSERT IGNORE INTO users (`parentId`, `name`, `admin`, `active`) VALUES(0, 'system', 'admin', 'active');");
+        $this->addSql("INSERT IGNORE INTO users (`parentId`, `name`, `admin`, `active`) VALUES(0, 'system', 1, 1);");
         $this->addSql("UPDATE users set id = 0 where name = 'system'");
     }
 
--- a/bundles/CoreBundle/Migrations/Version20220830105212.php
+++ b/bundles/CoreBundle/Migrations/Version20220830105212.php
@@ -22,6 +22,6 @@ final class Version20220830105212 extends AbstractMigration
 
     public function down(Schema $schema): void
     {
-        $this->addSql("DELETE FROM users WHERE `id` = 0");
+        //no need to delete system user entry
     }
 }
--- a/bundles/CoreBundle/Migrations/Version20220830105212.php
+++ b/bundles/CoreBundle/Migrations/Version20220830105212.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pimcore\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20220830105212 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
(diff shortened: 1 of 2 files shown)


bundles/CoreBundle/Migrations/Version20220830105212.php
```diff
@@ -0,0 +1,27 @@
...
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql("INSERT IGNORE INTO users (`parentId`, `name`, `admin`, `active`) VALUES(0, 'system', 'admin', 'active');");
+        $this->addSql("UPDATE users set id = 0 where name = 'system'");
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql("DELETE FROM users WHERE `id` = 0");
```
brusch: I'm sure about this, because the entry could have been there before, so maybe we just kick that, as it doesn't hurt anyway. WDYT?

brusch (changes requested):

bundles/CoreBundle/Migrations/Version20220830105212.php
dvesh3: this was my first  thought too! kicked it 😉

bundles/InstallBundle/Installer.php
```diff
@@ -815,4 +809,15 @@ protected function insertDatabaseContents()
             $db->insert('users_permission_definitions', $up);
         }
     }
+
+    protected function insertSystemUser(Connection $db)
```
blankse:
```suggestion
    protected function insertSystemUser(Connection $db): void
```

bundles/CoreBundle/Migrations/Version20220830105212.php
```diff
@@ -0,0 +1,27 @@
...
+final class Version20220830105212 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql("INSERT IGNORE INTO users (`parentId`, `name`, `admin`, `active`) VALUES(0, 'system', 'admin', 'active');");
```
aryaantony92:
```suggestion
        $this->addSql("INSERT IGNORE INTO users (`parentId`, `name`, `admin`, `active`) VALUES(0, 'system', 1, 1);");
```

Fahl-Design:
@aryaantony92 @brusch @blankse Guess what happens when updating an existing install ;)

```
[notice] Migrating up to Pimcore\Bundle\CoreBundle\Migrations\Version22020614115124
[error] Migration Pimcore\Bundle\CoreBundle\Migrations\Version20220830105212 failed during Execution. Error: "An exception occurred while executing 'UPDATE users set id = 0 where name = 'system'':

SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '0' for key 'users.PRIMARY'"

In AbstractMySQLDriver.php line 74:
                                                                               
  An exception occurred while executing 'UPDATE users set id = 0 where name =  
   'system'':                                                                  
                                                                               
  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '0' f  
  or key 'users.PRIMARY'                                                       
                                                                               
```

dvesh3:
@Fahl-Design do you have a user with `id = 0` & name is not `system` in users table? if yes, could you please share some info of this entry? thanks!

Fahl-Design:
@dvesh3 my system user already had `id = 0` I did not reinstall, I just updated it (first installation was version <6.X, than updated over time)

dvesh3:
I wonder then why you are getting the duplicate entry error as we are updating the same row.

Fahl-Design:
hmmm looks like my sql server is not CaseSensetive

image

updating with `name = "system"` tries to update 2 entries and results in duplication error

dvesh3:
Ok, this is reproducible. we need to fix the migration with type in where condition.

PR #13018, opened 2022-08-31
[Workflows] Add option to define what to do with unsaved changes when applying a transition

### LINKED ISSUE #7660: [UX] Workflow Transition for modified DataObject
It is currently possible to trigger a Workflow Transition even though the Object (guess also Asset/Document) has modifications. That ends up in loosing all changed stuff and having an "empty" DataObject again.

What do you think if we just add a Popup message telling the User that there are unsaved changes?

BlackbitDevs:
As discussed in #7660 it is an inconvenient workflow when you want to do some changes to a data object and afterwards apply a workflow transition. Currently you have to save the object first and then apply the workflow. If you miss the saving, currently the transition is tried to be applied without the just made changes.

This PR adds a new option `unsavedChangesBehaviour` for each transition with the possible values:
- `warn`: Show warning message that the object has not been saved yet. Do not apply transition.
- `save`: Save the object implicitly and apply workflow transition
- `ignore`: Do not save the object and apply workflow transition (this is current behaviour)

Default is `warn`.

Resolves #7660.

### DIFF
--- a/lib/Workflow/ActionsButtonService.php
+++ b/lib/Workflow/ActionsButtonService.php
@@ -55,7 +55,7 @@ class ActionsButtonService
                 'iconCls' => $transition->getIconClass(),
                 'objectLayout' => $transition->getObjectLayout(),
                 'notes' => $notes,
-                'saveBehavior' => $transition->getOptions()['saveBehavior']
+                'saveBehavior' => $transition->getOptions()['unsavedChangesBehaviour']
             ];
         }
 
--- a/lib/Workflow/ActionsButtonService.php
+++ b/lib/Workflow/ActionsButtonService.php
@@ -55,6 +55,7 @@ class ActionsButtonService
                 'iconCls' => $transition->getIconClass(),
                 'objectLayout' => $transition->getObjectLayout(),
                 'notes' => $notes,
+                'saveBehavior' => $transition->getOptions()['saveBehavior']
             ];
         }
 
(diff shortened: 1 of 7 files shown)


lib/Workflow/ActionsButtonService.php
```diff
@@ -55,6 +55,7 @@ public function getAllowedTransitions(Workflow $workflow, ElementInterface $elem
                 'iconCls' => $transition->getIconClass(),
                 'objectLayout' => $transition->getObjectLayout(),
                 'notes' => $notes,
+                'saveBehavior' => $transition->getOptions()['saveBehavior']
```
mcop1:
```suggestion
                'saveBehavior' => $transition->getOptions()['unsavedChangesBehaviour']
```
Here it should be something like this, right? And I think it would be good if we spell "behaviour" always in the same way, either "behaviour" or "behavior". What do you think?

BlackbitDevs: You are right. Have fixed naming to "behaviour" in all the different places.

mcop1:
Thanks for the changes, pr looks great! 
One last thing though, could you please add the configuration + comments here too? 

https://github.com/pimcore/pimcore/tree/11.x/doc/Development_Documentation/07_Workflow_Management/01_Configuration_Details

BlackbitDevs:
@mcop1 Thanks. Have added `unsavedChangesBehaviour` to the docs.

PR #13023, opened 2022-08-31
Update DAO classmap

blankse:
There are missing DAOs in the classmap.
After executing `php bin/console internal:model-dao-mapping-generator` the order is not the same. So I add `ksort()`. So the order should be the same for the next update.


mcop1:
Thanks 👍

PR #13035, opened 2022-09-01
[DataObject] Video - Fix adding multiple video data types & container enrichment 

dvesh3:
Resolves https://github.com/pimcore/pimcore/pull/12954#issuecomment-1232975312

## Additional info  
Fixes issues:
- Adding video data type multiple times in a definition.
- Allowed type option empty in container fields like object brick & fieldcollection.


dvesh3:
Reverted the Migration as no field level enrichment will be done on save from non admin context. Instead, video type implements `LayoutDefinitionEnrichmentInterface` which enriches the containerized fields on load.

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/video.js
```diff
@@ -211,9 +211,11 @@ pimcore.object.tags.video = Class.create(pimcore.object.tags.abstract, {
     },
 
     empty: function () {
+        let allowedTypes = this.data.allowedTypes;
```
kingjia90: this doesn't work because of this
https://github.com/pimcore/pimcore/blob/a0dc084959c787a1d05ed942a5dc7b623f3be6cd/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/video.js#L170
 after confirming with `openEdit`
 
We could use this
```suggestion
        let allowedTypes = this.fieldConfig.allowedTypes;
```
or push `allowedTypes` also on openEdit save

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/video.js
```diff
@@ -211,9 +211,11 @@ pimcore.object.tags.video = Class.create(pimcore.object.tags.abstract, {
...
         this.data = {
             type: "asset",
```
kingjia90: the actual default currently is not type = `asset`, it shows an empty type & path field
image

https://github.com/pimcore/pimcore/blob/a0dc084959c787a1d05ed942a5dc7b623f3be6cd/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/video.js#L24 

so either we change to default "asset" on init as well or remove from there to keep consistency

dvesh3: I would kick it since it is possible now that asset type is not allowed.

dvesh3:
```suggestion
            type: "",
```

Issue #12504, opened 2022-06-24
[Bug]: AssetUpdateTasksMessage queues endless for Video with errors


dpfaffenbauer:
### Expected behavior

Pimcore should only try once to create run AssetUpdateTasksMessage for Video

### Actual behavior

it runs it endless due to how `AssetUpdateTaskHlandler` works and how `AssetUpdateTasksMessage` is queued.

### Steps to reproduce

Upload a broken Video and see how the messages get spammed.

dpfaffenbauer:
https://user-images.githubusercontent.com/5981845/175475775-a7301112-2e2f-4b0a-8b54-9b3efd98b7a5.mp4

like this MP4

aryaantony92:
@dpfaffenbauer Could you please share the broken video for testing. Thanks!

dpfaffenbauer:
https://github.com/pimcore/pimcore/issues/12504#issuecomment-1165244354

aryaantony92:
It was unable to download the video directly.  Thanks!

dpfaffenbauer:
[Screenshot 2022-06-24 at 09.15.11 copy 4.mp4.zip](https://github.com/pimcore/pimcore/files/9083876/Screenshot.2022-06-24.at.09.15.11.copy.4.mp4.zip)

It's basically an empty file with an mp4 suffix

dvesh3:
Fixed by #12692

PR #13039, opened 2022-09-01
Log error and return null when ffmpeg couldn't read dimensions/duration

### LINKED ISSUE #13038: [Bug]: Endless message loop in pimcore_maintenance
### Expected behavior

Maintenance task is running without errors

### Actual behavior

Maintenance task runs in endless queue with these errors:

15:04:39 INFO      [messenger] Received message Pimcore\Messenger\AssetUpdateTasksMessage ["class" => "Pimcore\Messenger\AssetUpdateTasksMessage"]
15:04:39 WARNING   [php] Warning: Undefined array key 1 ["exception" => ErrorException { …}]
15:04:39 WARNING   [php] Warning: Undefined array key 1 ["exception" => ErrorException { …}]

How we can find the ErrorException? It is not tracked in php.log?

We've updated to 10.5.3 but the error still exists.
We looked for unknown or empty images and videos too in our Assets.

### Steps to reproduce

Sadly we can't share the steps how to reproduce. 
Maybe there's a corrupt image or video but the customer has over 2'000 assets.

blankse:
See https://github.com/pimcore/pimcore/issues/13038

Exceptions are already catched and logged in the messanger handler:
https://github.com/pimcore/pimcore/blob/b3f3111ce0a781069f6289e03fcfe5e974a98c29/lib/Messenger/Handler/AssetUpdateTasksHandler.php#L80-L93

### DIFF
--- a/lib/Messenger/Handler/AssetUpdateTasksHandler.php
+++ b/lib/Messenger/Handler/AssetUpdateTasksHandler.php
@@ -69,27 +69,18 @@ class AssetUpdateTasksHandler
 
     private function processVideo(Asset\Video $asset): void
     {
-        try {
-            $asset->setCustomSetting('duration', $asset->getDurationFromBackend());
-        } catch (\Exception $e) {
-            Logger::err('Unable to get duration of video: ' . $asset->getId());
-
-            throw new UnrecoverableMessageHandlingException($e->getMessage(), 0, $e);
+        if ($duration = $asset->getDurationFromBackend()) {
+            $asset->setCustomSetting('duration', $duration);
+        } else {
+            $asset->removeCustomSetting('duration');
         }
 
-        try {
-            $dimensions = $asset->getDimensionsFromBackend();
-            if ($dimensions) {
-                $asset->setCustomSetting('videoWidth', $dimensions['width']);
-                $asset->setCustomSetting('videoHeight', $dimensions['height']);
-            } else {
-                $asset->removeCustomSetting('videoWidth');
-                $asset->removeCustomSetting('videoHeight');
-            }
-        } catch (\Exception $e) {
-            Logger::err('Unable to get dimensions of video: ' . $asset->getId());
-
-            throw new UnrecoverableMessageHandlingException($e->getMessage(), 0, $e);
+        if ($dimensions = $asset->getDimensionsFromBackend()) {
+            $asset->setCustomSetting('videoWidth', $dimensions['width']);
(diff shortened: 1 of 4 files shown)


blankse:
I think getDimensions() should also like getDuration() return null if the adapter could not read. This should only be logged direct in the adapter. Because the getDurationFromBackend/getDimensionsFromBackend methods are also used in the getDuration/getDimensions methods of the Video asset. There the exceptions wouldn't be catched.

See https://github.com/pimcore/pimcore/blob/b3f3111ce0a781069f6289e03fcfe5e974a98c29/models/Asset/Video.php#L184-L274

dvesh3:
@blankse The reason `UnrecoverableMessageHandlingException` is thrown is to stop the worker from retries if there's an exception while processing the video. When removing it, this [issue ](https://github.com/pimcore/pimcore/issues/12504) will resurface.

blankse:
@dvesh3 Where is a exception thrown in the methods when the video is corrupt? I can test it with the given video.

dvesh3:
yes, try uploading this video 

[corrupt.zip](https://github.com/pimcore/pimcore/files/9490040/corrupt.zip)

blankse:
@dvesh3 In my system this video file will created as Text Asset. Because the MimeTypeGuesser return "text/plain".

I deleted some lines from a correct video file. Than I get correct return null and a log entry. Can you test it on your system. Is there an exception?

blankse:
I think it can thrown a exception if the ffmpeg binary isn't available or the Symfony Process had a timeout. I add a try catch in the adapter to catch this.

blankse:
@dvesh3 I found the endless loop. The thumbnail generation add the asset to the queue again if the video has no dimensions. See https://github.com/pimcore/pimcore/pull/13039/commits/d0fa49a513b821cb150fe7541c68e5a8446978c5

dvesh3:
> @dvesh3 I found the endless loop. The thumbnail generation add the asset to the queue again if the video has no dimensions. See [d0fa49a](https://github.com/pimcore/pimcore/commit/d0fa49a513b821cb150fe7541c68e5a8446978c5)

good catch! you mean, it is added to the queue [here](https://github.com/pimcore/pimcore/blob/11.x/models/Asset/Video.php#L176) if `videoWidth` & `videoHeight` is not set.

blankse:
Right :)

dvesh3:
@blankse thanks!

Issue #8163, opened 2021-02-12
[Custom Reports] Adding custom SQL statements to SELECT field does not work anymore


fashxp:
![image](https://user-images.githubusercontent.com/8792145/107785089-743c8f00-6d4c-11eb-9d7f-d8ce983af435.png)

regression of #7805

dvesh3:
fixed by #8166

PR #13050, opened 2022-09-02
[Reports] Support selecting functions without alias

BlackbitDevs:
Steps to reproduce bug:
1. Create data object class with field `analyze` - or any other reserved keyword from https://dev.mysql.com/doc/refman/8.0/en/keywords.html - another often used example is `set`
2. Create report with 
    `Select`: ``IFNULL(`analyze`, '- empty -')``
    `From`: `object_<class id>`
3. Open report

There will be an SQL error:
```sql
SELECT `IFNULL(`analyze`, '- empty -')` FROM (SELECT IFNULL(`analyze`, '- empty -') FROM object_<class id>) AS somerandxyz WHERE 1 = 1 LIMIT 0,50
```

The cause is that in https://github.com/pimcore/pimcore/blob/e53d45e0e69883ebc86cc273a2dcca2faa0cf4c6/models/Tool/CustomReport/Adapter/Sql.php#L32-L40 the fields for selection get extracted which get quoted in backticks in https://github.com/pimcore/pimcore/blob/e53d45e0e69883ebc86cc273a2dcca2faa0cf4c6/models/Tool/CustomReport/Adapter/Sql.php#L231

The problem is that when a field already contains backticks, it cannot be wrapped in backticks again. There is also no way of escaping backticks in SQL (really strange that this does not exist).

It would work if we change the select field to ``IFNULL(`analyze`, '- empty -') AS `analyze` `` but it would be good if we could support all valid SQL statements. I wonder why in #8163 the listing of the selected `$fields` got introduced. Without this, it works perfectly fine (and so did in Pimcore <= 6.8) - even if you put ``IFNULL(`analyze`, '- empty -') FROM object_<class id>`` into the `Select`field. For this reason I removed this in this PR and am curious for the feedback.


models/Tool/CustomReport/Adapter/Sql.php
```diff
@@ -118,12 +108,10 @@ protected function buildQueryString($config, $ignoreSelectAndGroupBy = false, $d
 
         if (!empty($config['where'])) {
             $whereParts = [];
-            if (!empty($config['where'])) {
-                if (strpos(strtoupper(trim($config['where'])), 'WHERE') === 0) {
-                    $config['where'] = preg_replace('/^\s*WHERE\s*/', '', $config['where']);
-                }
-                $whereParts[] = '(' . str_replace("\n", ' ', $config['where']) . ')';
+            if (strpos(strtoupper(trim($config['where'])), 'WHERE') === 0) {
```
jdreesen:
```suggestion
            if (str_starts_with(strtoupper(trim($config['where'])), 'WHERE')) {
```

mcop1:
@BlackbitNeueMedien thanks, looks good, but could you please take a look at the conflict? Thanks!

BlackbitDevs:
@mcop1 Resolved the conflict.

PR #13061, opened 2022-09-05
[Improvement] Concrete::__callStatic should support variants

### LINKED ISSUE #9686: Concrete::__callStatic should support variants
### Feature description

Follow up to https://github.com/pimcore/pimcore/issues/9457
please see https://github.com/pimcore/pimcore/issues/9457#issuecomment-865844378

mcop1:
Resolves #9686


brusch:
If we support them in the `getByXXX()` methods, we should do that in all other places as well to be consistent. 
Thinking of things like this (no a complete list): 
https://github.com/pimcore/pimcore/blob/acba01ca5bba7c8e218b554d45bb48e396f113d2/models/DataObject/AbstractObject.php#L528
https://github.com/pimcore/pimcore/blob/fe3e22f6d8eb5c6d34f47d4bf66f65f87aa426a5/models/DataObject/Listing.php#L40

models/DataObject/AbstractObject.php
```diff
@@ -1409,18 +1409,18 @@ public static function __callStatic($method, $arguments)
     protected static function makeList(array $listConfig, mixed $objectTypes): Listing
```
jdreesen: I think `$objectTypes` should be changed to `array`, as it'll probably break when trying to `array_diff()` it with some other type.

mcop1: Good point, thanks

models/DataObject/AbstractObject.php
```diff
@@ -1399,28 +1399,32 @@ public static function __callStatic($method, $arguments)
...
-        if (!empty($objectTypes)) {
-            if (\array_diff($objectTypes, [static::OBJECT_TYPE_VARIANT, static::OBJECT_TYPE_OBJECT])) {
-                Logger::error('Class: DataObject\\AbstractObject => Unsupported object type in array ' . implode(',', $objectTypes));
+        if(!is_array($objectTypes)) {
+            Logger::error('Class: DataObject\\AbstractObject => Unsupported argument type ' . gettype($objectTypes)  . ' in function ' . __FUNCTION__);
 
-                throw new \Exception('Unsupported object type in array [' . implode(',', $objectTypes) . '] in class DataObject\\AbstractObject');
-            }
+            throw new \Exception('Argument $objectTypes must be of type array, ' .  gettype($objectTypes) . ' given.');
+        }
```
jdreesen: What's wrong with changing the parameter type? It's way easier than doing the error handling manually.

blankse:
@mcop1 It is 0 because of this lines
https://github.com/pimcore/pimcore/blob/a57140f4d0ed37bee8294b670e18ff5cfe56f404/models/DataObject/AbstractObject.php#L1362-L1363
We can change the type hint to array and change this lines.

mcop1:
> @mcop1 It is 0 because of this lines
> 
> https://github.com/pimcore/pimcore/blob/a57140f4d0ed37bee8294b670e18ff5cfe56f404/models/DataObject/AbstractObject.php#L1362-L1363
> 
> 
> We can change the type hint to array and change this lines.

I figured that one out, but I didn´t want to change it to "null", because that would include `$limit` and `$offset` variables too. And I didn´t want to do something like 
`$objectTypes = [DataObject::OBJECT_TYPE_OBJECT, DataObject::OBJECT_TYPE_FOLDER];` there, because we would need to do that everywhere we call the `makeList` function and want to support the default. 

We could remove the array padding and check for the arguments separately- that would work, I guess.

blankse:
@mcop1 Why exists this `array_pad`? I think we can simply remove this from AbstractObject.php and Concrete.php. The list will set the variables to `null` then. This should be okay. We can than use `?array` type hint.

`$limit` and `$offset` are checked by an if. So `0` or `null` make no different.

jdreesen:
> Edit: Why exists this array_pad? I think we can simply remove this from AbstractObject.php and Concrete.php. The list will set the variables to null then. This should be okay. We can than use ?array type hint.

The padding is needed to prevent `Undefined array key` warnings, see: https://3v4l.org/Fo0cc

blankse:
@jdreesen Can we pad it with `null`?

jdreesen:
I think so. Or do e.g. something like this: https://3v4l.org/9ts7u

jdreesen:
Or don't use array unpacking ;)

jdreesen:
Should work, too:

```diff
-$arguments = array_pad($arguments, 4, 0);
+$arguments = array_pad($arguments, 3, 0);
-[$value, $limit, $offset, $objectTypes] = $arguments;
+[$value, $limit, $offset] = $arguments;
...
-$list = static::makeList($listConfig, $objectTypes);
+$list = static::makeList($listConfig, $arguments[3] ?? []);
```

So many possibilities :cat2:

mcop1:
I would like to keep the individual values for (0 for offset and limit, null for objectTypes). 
But I would prefer something like that:
` $value = $arguments[0] ?? 0;
                $limit = $arguments[1] ?? 0;
                $offset = $arguments[2] ?? 0;
                $objectTypes = $arguments[3] ?? null;`

Not that compact, but I think it is easy to read and makes the intention on what we are doing here pretty clear. Any thoughts on that?

blankse:
The form is okay I think. But I think `$value` and `$limit` should also be `null` (when there wasn't a if for limit 0 will return nothing). Only `$offset` should be `0`. So we don't need the ifs for $limit/$offset.

mcop1:
> The form is okay I think. But I think `$value` and `$limit` should also be `null` (when there wasn't a if for limit 0 will return nothing). Only `$offset` should be `0`. So we don't need the ifs for $limit/$offset.

Good point, then I will change the class definition as well, so that the default value is actually representing what we are assigning.

lib/DataObject/ClassBuilder/ClassBuilder.php
```diff
@@ -77,7 +77,7 @@ public function buildClass(ClassDefinition $classDefinition): string
                     $classDefinition->getName()
                 ).'|null getBy'.ucfirst(
                     $fieldDefinition->getName()
-                ).'($field, $value, $locale = null, $limit = 0, $offset = 0, $objectTypes = null)'."\n";
+                ).'($field, $value, $locale = null, $limit = null, $offset = 0, $objectTypes = null)'."\n";
```
blankse:
```suggestion
                ).'(string $field, mixed $value, ?string $locale = null, ?int $limit = null, int $offset = 0, ?array $objectTypes = null)'."\n";
```

lib/DataObject/ClassBuilder/ClassBuilder.php
```diff
@@ -86,14 +86,14 @@ public function buildClass(ClassDefinition $classDefinition): string
                         $classDefinition->getName()
                     ).'|null getBy'.ucfirst(
                         $localizedFieldDefinition->getName()
-                    ).'($value, $locale = null, $limit = 0, $offset = 0, $objectTypes = null)'."\n";
+                    ).'($value, $locale = null, $limit = null, $offset = 0, $objectTypes = null)'."\n";
```
blankse:
```suggestion
                    ).'(mixed $value, ?string $locale = null, ?int $limit = null, int $offset = 0, ?array $objectTypes = null)'."\n";
```

lib/DataObject/ClassBuilder/ClassBuilder.php
```diff
@@ -86,14 +86,14 @@ public function buildClass(ClassDefinition $classDefinition): string
...
                 }
             } elseif ($fieldDefinition->isFilterable()) {
                 $cd .= '* @method static \\Pimcore\\Model\\DataObject\\'.ucfirst(
                     $classDefinition->getName()
                 ).'\Listing|\\Pimcore\\Model\\DataObject\\'.ucfirst(
                     $classDefinition->getName()
-                ).'|null getBy'.ucfirst($fieldDefinition->getName()).'($value, $limit = 0, $offset = 0, $objectTypes = null)'."\n";
+                ).'|null getBy'.ucfirst($fieldDefinition->getName()).'($value, $limit = null, $offset = 0, $objectTypes = null)'."\n";
```
blankse:
```suggestion
                ).'|null getBy'.ucfirst($fieldDefinition->getName()).'(mixed $value, ?int $limit = null, int $offset = 0, ?array $objectTypes = null)'."\n";
```

models/DataObject/AbstractObject.php
```diff
@@ -1359,8 +1359,10 @@ public static function __callStatic($method, $arguments)
         $db = \Pimcore\Db::get();
 
         if (in_array(strtolower($realPropertyName), self::$objectColumns)) {
-            $arguments = array_pad($arguments, 4, 0);
-            [$value, $limit, $offset, $objectTypes] = $arguments;
+            $value = $arguments[0] ?? null;
```
blankse: Should we throw an exception for required arguments instead of setting it to `null`?

blankse (changes requested):

models/DataObject/AbstractObject.php
mcop1: Like that, added the exceptions

models/DataObject/AbstractObject.php
```diff
@@ -1359,7 +1359,7 @@ public static function __callStatic($method, $arguments)
         $db = \Pimcore\Db::get();
 
         if (in_array(strtolower($realPropertyName), self::$objectColumns)) {
-            $value = $arguments[0] ?? null;
+            $value = empty($arguments[0]) ? throw new \InvalidArgumentException('Mandatory argument $value not set.') : $arguments[0];
```
blankse: null, '', 0 should be possible. Use `!array_key_exists()`

jdreesen: Should be possible to write it like this, no? Or do you explicitly want to use `empty()` here?
```suggestion
            $value = $arguments[0] ?? throw new \InvalidArgumentException('Mandatory argument $value not set.');
```

jdreesen: (only possible if `null` is invalid)

jdreesen: I'd write it like this, then: 
```suggestion
            $value = array_key_exists($arguments[0]) ? $arguments[0] : throw new \InvalidArgumentException('Mandatory argument $value not set.');
```

mcop1: I agree on using array_key_exists, things are a bit tricky here, because passing `''` or `null` doesn´t make a difference for strings, it will result in the same sql query. And with passing `''` you won´t find any newly created objects with textfields that haven´t been edited- because the corresponding database field is set to `NULL` and not `''`. I suspect this to be a defect and created an issue here: https://github.com/pimcore/pimcore/issues/13089 

Thanks for the help guys.

PR #13073, opened 2022-09-06
[Recycle bin] Add date index

BlackbitDevs:
By default the recycle bin is sorted by `date`, see https://github.com/pimcore/pimcore/blob/2ccdd54162f97b5e89caf1f25fee76bfadf8f972/bundles/AdminBundle/Controller/Admin/RecyclebinController.php#L56

Currently there is no database index covering this. With lots of recycle bin items this may take some time to load the list. This PR adds a database index.


mcop1:
Thank you

PR #13074, opened 2022-09-06
[Improvement] dispatch event before admin login render

kjkooistra-youwe:
Dispatch an event before the admin login view is rendered. This allows overriding the parameters and rendering controller actions or including templates.

The purpose of rendering controllers or including templates is to allow adding custom login buttons for Single Sign-on, for example.


brusch:
I'm wondering if using Twig blocks and the template override functionaility of Symfony is a more elegant way, because `{{ contentBelowLogin|raw }}` is not really best practice ... 

We can still keep the event to add additional template vars, but then you'd need to override the template/blocks: 
https://symfony.com/doc/current/bundles/override.html

kjkooistra-youwe:
@brusch I agree it would be preferable to use another method than the `raw` variable. Unfortunately it's rather tricky to override a template from another bundle (let alone multiple). It doesn't make sense to have to override the template for each project when you can do it from a standardized SSO bundle (like the Pimcore OpenID bundle).

Maybe an alternative could be to configure include paths, which can be configured via the event listener.

kjkooistra-youwe:
@brusch I pushed an improved version.

Updated commit message.

kjkooistra-youwe:
Fixed merge conflict in controller.

kjkooistra-youwe:
@brusch Any thoughts on the improved version? Is it OK this way?

brusch:
Looks good to me, maybe just 2 things: 
- I'd go for either `renderControllerActions` or `includeTemplates` but not both, I'd prefer `includeTemplates`
- I'd move the include part in `login.html.twig` more to the end of the template, maybe to here: https://github.com/pimcore/pimcore/blob/7f30871c61248a77a506c2b3f0d572f416b87d09/bundles/AdminBundle/templates/admin/login/login.html.twig#L105 (that gives you more possibilities)

kjkooistra-youwe:
@brusch Applied changes.

brusch:
LGTM, thanks!

PR #13093, opened 2022-09-07
[Data objects] Dynamically update select fields options when using dynamic options provider

BlackbitDevs:
Pimcore supports context-sensitive option providers for select fields since a long time. But the available option list got only updated when the object was loaded for editing. When the options depend on another object field, you had to save the object and reload it to get the updated select options.

With this PR the option list gets dynamically updated also with the yet unsaved information.

Imagine you have an OptionsProvider like this:
```php
class OptionsProvider implements SelectOptionsProviderInterface
{
    public function getOptions($context, $fieldDefinition) {
        $object = isset($context["object"]) ? $context["object"] : null;

        $objectFieldValue = $object ? $object->getSomeField() : 'no value';
        return [
            ["key" => $objectFieldValue.' 1', "value" => $objectFieldValue.'1'],
            ["key" => $objectFieldValue.' 2', "value" => $objectFieldValue.'2'],
            ["key" => $objectFieldValue.' 3', "value" => $objectFieldValue.'3'],
        ];
    }

    public function getDefaultValue($context, $fieldDefinition) {
        return $fieldDefinition->getDefaultValue();
    }

    public function hasStaticOptions($context, $fieldDefinition) {
        return false;
    }
}
```

You see that the available options depend on the field "someField". As this options provider returns `false` for `hasStaticOptions()` it means that the select field options depend on the current object's state.

**Workflow**

1. Options before changing data
Bildschirmfoto 2022-09-07 um 15 59 23

2. Changed value for field (but did not save yet) -> Options automatically get updated
Bildschirmfoto 2022-09-07 um 15 59 39


mcop1:
@BlackbitNeueMedien Basically the pr looks good to me. 
But there are 2 things I noticed: 

1.) When you have selected a value in the select and change the value of "Some Field" afterwards, then there is no notification about the changed values in the select. Therefore, after reloading the object, the default value will be set again. This is more a general thing, since the behavior is the same with or without the changes in the pr. But I think it would be a good idea to somehow bring this to the user´s attention while editing the object. What do you think? I would create an separate issue/improvement for that. 

2.) When `hasStaticOptions` returns true, then we could cut the `get-select-options` request, right? Maybe we could check for `this.fieldConfig.dynamicOptions` before sending the request?

BlackbitDevs:
Thx for your feedback @mcop1.

about 1.: Added a notification for select fields which have a value which is not valid anymore. This also helps for non-dynamic select fields, e.g. when you change the available options (in this moment in the database there still is the previously set option value). When a user opens the object again, he will get a notification that the set value is not valid anymore. 

about 2.: You are right, is fixed now.

And now also multiselect fields are supported.

mcop1:
@BlackbitNeueMedien thank you for your efforts!

brusch:
Does this really cover all cases, don't think so ... what about field-collections, object bricks just to name 2.
If we provide this feature it should be consistently working ... 
@mcop1 If my assumption is right, we should revert this PR and recreate it once it's really ready.

brusch:
Also this PR produces quite a lot of duplicate code, see: 
https://github.com/pimcore/pimcore/blob/70b1db360a2a4fa498c69f71a33aa1db5a3a42d1/bundles/AdminBundle/src/Controller/Admin/DataObject/DataObjectController.php#L1309 

We should extract this into a private function, as it's in the same controller class anyway.

brusch:
@BlackbitNeueMedien Can you maybe create a follow-up PR that cleans up the duplicate code? 😊 Thanks a lot in advance!

BlackbitDevs:
@brusch Yes, will do. About the object bricks: Will test if current implementation works with object bricks / field collections...

BlackbitDevs:
The small refactoring is in #13239
I think this feature is compatible with object bricks and field collections. Have just tried and `this.object` in
https://github.com/pimcore/pimcore/blob/da2265126f4b14153a433d876d9776353ca56da2/bundles/AdminBundle/public/js/pimcore/object/tags/select.js#L278-L279
 is available there too. Also applying the unsaved changes should work in https://github.com/pimcore/pimcore/blob/da2265126f4b14153a433d876d9776353ca56da2/bundles/AdminBundle/src/Controller/Admin/DataObject/DataObjectController.php#L597 (respectively the refactored code in #13239)

Anything I have missed, @brusch ?

brusch:
@BlackbitNeueMedien Thanks for your feedback - we should however do another manual test just to double check.

mcop1:
@brusch @BlackbitNeueMedien I tested it with object bricks and classification stores too, but I will do another quick check while testing #13239

PR #13094, opened 2022-09-07
Fix sending image data embedded in CSV

NiklasBr:
Sending image data in CSV didn't really make a lot of sense, fixing it to instead send the expected image path.

See #13079


mcop1:
@NiklasBr since this is a change of behaviour, could you please rebase to 11.x? 
Thank you.

NiklasBr:
image

Huh?

Was the behaviour correct to begin with?

NiklasBr:
@mcop1 why change target?

mcop1:
@NiklasBr I tried to change the target to 11.x, because I wanted to see if it works without rebasing, but it didn´t, so I changed it back to 10.5. 
image

But this pr changes the behaviour and I think it would be better to avoid this in a bugfix release. That´s why I have asked you to please rebase it to 11.x.

NiklasBr:
Isn't the entire point of bug fixes to resolve unintended behaviours?

mcop1:
The path makes more sense to me too, I agree with you on that. But the code is as it is since years. I can´t say if no one is using this feature, but if someone is, then we would break the behaviour for him.

NiklasBr:
See #13154

PR #13109, opened 2022-09-08
Added reset option for getChildren on objects. Solves the problem tha…

BentBr:
Hey Folks!

There are certain cases in which your're playing around with objects during one (long) reqest.
Storing new Objects and adding new requires to check whether recently created children of an object to clash with names. Therefore all recently (as well as old objects) must be get-able.

We can see here that $object->getChildAmount() increases as expected during creation of multiple children of one parent, but $object->getChildren() is always the same array.
This is caused due to the in-class caching via the $o_children[$cacheKey] variable.

To circumvent this problem I've added the option to manually reset this cache.

  

- New private method to reset the $o_children var
- New optional param to reset the cache during fetching of children

## Additional info  

Cheers!

### DIFF
--- a/models/DataObject/AbstractObject.php
+++ b/models/DataObject/AbstractObject.php
@@ -522,11 +522,16 @@ abstract class AbstractObject extends Model\Element\AbstractElement
     /**
      * @param array $objectTypes
      * @param bool $includingUnpublished
+     * @param bool $resetCache
      *
      * @return DataObject[]
      */
-    public function getChildren(array $objectTypes = [self::OBJECT_TYPE_OBJECT, self::OBJECT_TYPE_FOLDER], $includingUnpublished = false)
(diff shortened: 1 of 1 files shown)


blankse:
@BentBr You can already call `->setChildren(null)` to reset the children.

BentBr:
@blankse, thx for this pro tip. I didn't see that one coming. I bit counter intuitive, not?

I slightly enhanced my suggestion. What do you think?

BentBr:
Can some1 restart the tests? Seems like a docker issue :confused:

blankse:
@BentBr I think it is enough to reset it via `setChildren(null)`. This is also in documents and asset folders. The core use this on many places, see:
image
But that has to be decided by the pimcore core team.
I think we should add the `setChildren(null)` reset to the docs.

brusch:
I'd also prefer to use `setChildren(null)` and document it properly. 
This change cannot be merged anyway into `10.5` as it is a breaking change ...

brusch:
@BentBr thanks anyway for your PR ... can you maybe create a follow up for the documentation? ☺️

PR #13113, opened 2022-09-09
[Bug]: 10.5.2 language selection does not work on grid

### LINKED ISSUE #12958: [Bug]: 10.5.2 language selection does not work on grid
### Expected behavior

After choosing language in grid options, user can see translated values on the grid

### Actual behavior

After choosing language in grid options, user cannot see translated values on the grid. User has to change application language in order to see translated values.

### Steps to reproduce

1. open grid options
2. choose english language
3. choose desired attribute

Additional info:
my attribute (type:select) is translated in the translations "messages" and "admin"
admin

Grid view with polish language in the app:
grid with polish app language

Grid view with english language in the app:
grid with english app language

robertSt7:
resolves #12958

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/MiscController.php
+++ b/bundles/AdminBundle/Controller/Admin/MiscController.php
@@ -768,4 +768,29 @@ class MiscController extends AdminController
     {
         return new Response('done');
     }
+
+    /**
+     * @Route("/translate", name="pimcore_admin_misc_translate")
+     *
+     * @param Request $request
+     *
+     * @return JsonResponse
+     */
+    public function translate(Request $request)
+    {
+        $values = explode(',', $request->get('values'));
+        $language = $request->get('language') ?? null;
+        if($language === 'default') {
+            $language = null;
+        }
+        $translatedValues = [];
+        foreach ($values as $value) {
+            $translatedValues[$value] = $this->trans($value, locale:$language);
+        }
+
+        return new JsonResponse([
+            'success' => true,
+            'data' => $translatedValues
+        ]);
+    }
 }
--- a/bundles/AdminBundle/Controller/Admin/MiscController.php
+++ b/bundles/AdminBundle/Controller/Admin/MiscController.php
@@ -778,7 +778,7 @@ class MiscController extends AdminController
      */
     public function translate(Request $request)
     {
-        $values = explode(',', $request->get('values'));
+        $values = $this->decodeJson(rawurldecode($request->get('values')));
         $language = $request->get('language') ?? null;
         if($language === 'default') {
             $language = null;
--- a/lib/DataObject/GridColumnConfig/Operator/Alias.php
+++ b/lib/DataObject/GridColumnConfig/Operator/Alias.php
@@ -21,18 +21,16 @@ use Symfony\Contracts\Translation\TranslatorInterface;
 /**
  * @internal
  */
-final class Alias extends AbstractOperator
+final class Alias extends AbstractOperator implements TranslatorOperatorInterface
 {
     private TranslatorInterface $translator;
 
     /**
      * {@inheritdoc}
      */
-    public function __construct(\stdClass $config, $context = null, TranslatorInterface $translator)
+    public function __construct(\stdClass $config, $context = null)
     {
         parent::__construct($config, $context);
-
-        $this->translator = $translator;
     }
 
     /**
@@ -80,5 +78,10 @@ final class Alias extends AbstractOperator
         }
 
         return $result;
+    }
+
+    public function setTranslator(TranslatorInterface $translator): void
+    {
+        $this->translator = $translator;
     }
 }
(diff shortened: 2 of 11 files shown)


mcop1 (commented):
@robertSt7 seems like operators aren´t working anymore. E.g. this is working with 10.5 / 11.x, but not with the changes in this pr. Could you please have a look? 

image

bundles/AdminBundle/Controller/Admin/MiscController.php
```diff
@@ -768,4 +768,29 @@ public function testAction(Request $request)
...
+    /**
+     * @Route("/translate", name="pimcore_admin_misc_translate")
+     *
+     * @param Request $request
+     *
+     * @return JsonResponse
+     */
+    public function translate(Request $request)
+    {
+        $values = explode(',', $request->get('values'));
```
brusch: Could be an issue if the value contains a `,` - I'd prefer a more generic approach, e.g. using `values[]=value` or escaping (urlencoding) the values. WDYT?

brusch (changes requested):

bundles/AdminBundle/Controller/Admin/MiscController.php
robertSt7: I will use urlencoding

lib/DataObject/GridColumnConfig/Operator/Alias.php
```diff
@@ -40,6 +53,11 @@ public function getLabeledValue($element)
             $childResult = $c->getLabeledValue($element);
             $isArrayType = $childResult->isArrayType ?? null;
             $childValues = $childResult->value;
+
+            if($childResult->def instanceof Select) {
```
mcop1, line 57: I think we need to include Select and Multiselect here, right?

lib/DataObject/GridColumnConfig/Operator/Concatenator.php
```diff
@@ -61,6 +66,10 @@ public function getLabeledValue($element)
             $childResult = $c->getLabeledValue($element);
             $childValues = (array)($childResult->value ?? []);
 
+            if($childResult->def instanceof Select) {
```
mcop1, line 69: I think we need to include Select and Multiselect here too

lib/DataObject/GridColumnConfig/Operator/Merge.php
```diff
@@ -57,6 +62,10 @@ public function getLabeledValue($element)
             $childResult = $c->getLabeledValue($element);
             $childValues = $childResult->value ?? null;
 
+            if($childResult->def instanceof Select) {
```
mcop1, line 65: I think we need to include Select and Multiselect here too

brusch:
I'm not entirely sure if this is really the intended behavior ... 
Changing the language in the grid is ONLY for localized values. Translated labels of select data-types are not localized values, the value is the same and the label is shown in the language of the user-interface - that's imho the expected behavior. 

Simple example: 
- Select field with 3 values: `yes` , `no` and `none`. 
- There are translations for the 3 labels of the values
- Now you have also configured 2 languages, EN and FR and your class is using localized fields
- Your admin UI language is EN
- You switch to FR in the grid
- The values of the select are still  `yes` , `no` and `none` - why should you now translate them to FR if you're using EN in the UI?

kamilawiercinska:
hello @brusch  , I would like to respond to your last concern, as this is a fix to my original bug report.
Well, I cannot fully agree with you, that the behaviour described by you is correct.
I do not see a proper logic there- why should you see both english and french values?
For example, you are british, but you want to export data for your french customer. You should be able to see the object grid fully in french, without changing the UI language, in order to provide understandable data to your customer.
I am looking forward to hear from you, what are your thoughts on this one.

brusch:
The value is still right, the value is NOT the label and the export should export the actual data, not some translated labels. 
Different example, select is used as an enum, because you're using the CSV export for importing stuff into another system.

image

For me, it would be wrong to export the label and not the actual value of a field.

kamilawiercinska:
@brusch , okay so let's forget about the export- because it's another issue actually.
Can you please explain then  what is the use for translating only localized fields using the grid?
I cannot see any profit from that, let's suppose that I just want check some data in french but I am an english user. There is no point for me to switch the language of the app in order to see full data in french.
imho, seeing data both in english and french simultuanously is not an expected behvaiour.

brusch:
>  let's suppose that I just want check some data in french but I am an english user.

In case of translated labels from a select field, the value doesn't change, so no matter in which language the labels are shown, the value behind the label is still the same. While for localized fields the values are NOT the same, they are persisted for each language. So for a select field the English user will see the English labels, no matter which selection has been made for the localized fields. Please also keep in mind that there can be different languages for the user interface and localized fields. So the scenario you described won't work anyway in every case.

brusch:
Let's keep this open for discussion for now ...

robertSt7:
The PR is closed due to inactivity.

PR #13114, opened 2022-09-09
fix: cache warmer must not use the database or application cache

### LINKED ISSUE #12887: [Bug]: app cache still being read while the Symfony cache warmer is running
### Expected behavior

See https://github.com/pimcore/pimcore/pull/12571

### Actual behavior

See https://github.com/pimcore/pimcore/pull/12577

### Steps to reproduce

Enable Redis cache and try to build Pimcore into a Docker image.

It will try to trigger a cache connection from `\Pimcore\Tool::getSupportedLocales()`.

IMO the original PR still seems like the most correct solution. /cc @brusch

dkarlovi:
Resolves #12887

### DIFF
--- a/lib/HttpKernel/CacheWarmer/PimcoreCoreCacheWarmer.php
+++ b/lib/HttpKernel/CacheWarmer/PimcoreCoreCacheWarmer.php
@@ -98,20 +98,18 @@ class PimcoreCoreCacheWarmer implements CacheWarmerInterface
             $classes[] = $listingClass;
         }
 
-        $list = new DataObject\Objectbrick\Definition\Listing();
-        $list = $list->load();
-
-        foreach ($list as $brickDefinition) {
-            $className = 'Pimcore\\Model\\DataObject\\Objectbrick\\Data' . ucfirst($brickDefinition->getKey());
+        $objectBricksFolder = PIMCORE_CLASS_DEFINITION_DIRECTORY . '/objectbricks';
+        $files = glob($objectBricksFolder . '/*.php');
+        foreach ($files as $file) {
+            $className = 'Pimcore\\Model\\DataObject\\Objectbrick\\Data\\' . basename($file, '.php');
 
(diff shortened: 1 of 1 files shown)


kingjia90:
@dkarlovi Thank you for the PR! 
Just wondering, wasn't this PR supposed to be using `\Pimcore\Cache::disable();` apporach ? Or did you find an alternative approach by moving the object brick and fieldcollection `\Definition\Listing load()` ?
https://github.com/pimcore/pimcore/blob/cd6d886e530d0c92db4b03f1fd41413c093acdc2/models/DataObject/Objectbrick/Definition/Listing.php#L25-L36

dkarlovi:
The `include` will do some unwanted stuff like try to fetch available languages which then triggers database and/or cache calls. It's a whole soup of calls which are being forced here when using the existing feature as-is, preventing it from working in described circumstances.

kingjia90:
Ahhh.. that makes sense, nice catch.

Could we keep using `Definition/Listing.php` load() but refactoring it a little bit, in a way that there 2 methods(?) or with an extra optional parameter, where it can be used to loading the files with/without the `include`.

I see that it could benefit some other places that loading with `include` would be reduntant (and more expensive) since we only need the filename eg:
https://github.com/pimcore/pimcore/blob/cd6d886e530d0c92db4b03f1fd41413c093acdc2/bundles/AdminBundle/Controller/Admin/DataObject/ClassController.php#L1180-L1187
 and have less "duplicated" code

dkarlovi:
Sounds good to me, but I have no time to work on it in that more extended way, this was a simple bugfix for the currently broken feature. Maybe we can merge here and then refactor into a separate method or do it right away, it's up to you.

brusch:
The solution in the PR would be ok for me for `10.5`. 
For 11 we should think about changing the approach for `Listing` as suggested (maybe we even don't need a flag and just use the files).

PR #13117, opened 2022-09-09
Remove useless exception handling in Area(Block)::buildInfoObject()

jdreesen:
The exception handling in `Area(Block)::buildInfoObject()` was rather useless.

In `Area`, `$info` was set to `null` in case of an error, but later `$info->setParams(...)` gets called, which would crash.

In `Areablock`, an error was logged if one of the first three setters failed (which is rather unlikely), and I think it'll be better to fail hard in this case.

---

I also simplified the code to determine the `$params` a bit.


aryaantony92:
@jdreesen Thank you!

PR #13124, opened 2022-09-12
[Unit Conversion] Do not use ContainerInterface 

### LINKED ISSUE #12790: Fix Psr\Container\ContainerInterface injection to Symfony\Component\DependencyInjection\ContainerInterface Rest Deprecations...
### Improvement description

...

dvesh3:
Resolves #12790

### DIFF
--- a/models/DataObject/ClassDefinition/Helper/UnitConverterResolver.php
+++ b/models/DataObject/ClassDefinition/Helper/UnitConverterResolver.php
@@ -1,0 +1,37 @@
+<?php
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\DataObject\ClassDefinition\Helper;
+
+use Pimcore\Model\DataObject\QuantityValue\QuantityValueConverterInterface;
+
+/**
+ * @internal
+ */
+class UnitConverterResolver extends ClassResolver
+{
+
+    /**
+     * @param string $converterServiceName
+     *
+     * @return QuantityValueConverterInterface|null
+     */
+    public static function resolveUnitConverter($converterServiceName): ?QuantityValueConverterInterface
+    {
+        return self::resolve('@' . $converterServiceName, static function ($converterService) {
+            return $converterService instanceof QuantityValueConverterInterface;
+        });
+    }
+}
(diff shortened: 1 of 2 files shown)


aashan10 (approved):

aashan10:
@dvesh3 null is being returned from the class resolver when trying out a custom Converter

models/DataObject/ClassDefinition/Helper/UnitConverterResolver.php
```diff
@@ -0,0 +1,37 @@
...
+{
+
+    /**
+     * @param string $converterServiceName
+     *
+     * @return QuantityValueConverterInterface|null
+     */
+    public static function resolveUnitConverter($converterServiceName): ?QuantityValueConverterInterface
+    {
+        return self::resolve('@' . $converterServiceName, static function ($converterService) {
```
aashan10, line 33: `null` is being returned from the resolve method even if the class implements `QuantityValueConverterInterface` when using a custom converter

dvesh3: please make sure you define a public custom converter service.

aashan10: Thanks, there was an error in the name of service in services.yaml on my side. I have corrected it and now it works!

aashan10 (approved):

models/DataObject/QuantityValue/UnitConversionService.php
```diff
@@ -51,16 +47,12 @@ public function convert(AbstractQuantityValue $quantityValue, Unit $toUnit)
...
-        if ($converterServiceName) {
-            $converterService = $this->container->get($converterServiceName);
-        } else {
-            $converterService = $this->container->get(QuantityValueConverterInterface::class);
-        }
+        $converterServiceName = $baseUnit->getConverter();
+        $converterService = UnitConverterResolver::resolveUnitConverter($converterServiceName);
 
         if (!$converterService instanceof QuantityValueConverterInterface) {
-            throw new \Exception('Converter class needs to implement '.QuantityValueConverterInterface::class);
```
blankse: Why not throw an exception if the service name is wrong? I think the default service should only be used if no service name is set

dvesh3: yes, [this](https://github.com/pimcore/pimcore/pull/12742) should already throw exception. maybe we can refactor a bit to only resolve converter when the value is set.

blankse: @dvesh3 This will only throw an exception if the service doesn't exist. But returns null if only the interface is missing, right? I think we should hold the logic here as it was before.

```
        $converterServiceName = $baseUnit->getConverter();
        if ($converterServiceName) {
            $converterService = UnitConverterResolver::resolveUnitConverter($converterServiceName);
        } else {
            $converterService = $this->defaultConverter;
        }

        if (!$converterService instanceof QuantityValueConverterInterface) {
            throw new \Exception('Converter class needs to implement '.QuantityValueConverterInterface::class);
        }
```

dvesh3: I got your point now. yes, this changes behavior as it silently fallback to the default converter if the custom converter service is not implementing `QuantityValueConverterInterface`. 

Changed it as per suggestion. thanks!

aashan10 (approved):

PR #13131, opened 2022-09-12
[Relations] Filter allowed objects by SQL condition

### LINKED ISSUE #5969: Dynamic relation fields
When building a PIM for our clients the question is raised that they want to be able to filter specific relational fields depending on other relation fields in the same object.

For instance, when you create a new object the client sets a `mainCategory` which is a one-to-many relation. There is a second field `subCategory`, but when you're using the search it returns all the subCategory objects that are found. In this case i only want to be able to return the subCategories that are related to the selected `mainCategory`. 

Is something like this possible or would this be a new feature?

BlackbitDevs:
This PR adds an `Filter condition (SQL)` configuration to (advanced) many-to-many object relation fields. This can be used to filter the allowed objects. Until now you could only filter for allowed classes. With this PR you are able to narrow the allowed objects down also by SQL condition. 

Example:
You have 3 classes:
1. `Company`
  - name
2. `Department`
  - name
  - many-to-one relation to `Company`
3. `Employee`
  - name
  - many-to-one relation to `Company`
  - many-to-many relation to `Department` (the employee can work for different departments but only for those which belong to his `Company`
    - this configured `Filter condition (SQL)` is `company__id='{{ object.company ? object.company.id : "" }}'`
    - the SQL condition supports placeholders with Twig syntax. This currently opened object gets provided as variable `object`. This way you can access any data of the current object and so create dynamic relation conditions.

The data object tree looks like this:
- Home
  - Companies (folder)
    - Blackbit
      - Departments (folder)
        - Development
        - Sales
      - Employees (folder)
        - John Doe
    - Pimcore
      - Departments (folder)
        - Marketing
        - IT
 
Now we open `/Companies/Blackbit/Employees/John Doe`. At first in the field `Department` there will be no object allowed to be assigned because the configured `Filter condition (SQL)` will get resolved to `company__id=''`. As soon as the `Company` `/Companies/Blackbit` gets assigned you will be allowed to assign `/Companies/Blackbit/Departments/Development` and `/Companies/Blackbit/Departments/Sales`. The `Department` objects `/Companies/Pimcore/Departments/Marketing` and `/Companies/Pimcore/Departments/IT` will not be allowed.

Similar to #13093 this filtering works dynamically, so you do not have to reload the object to update the allowed objects. For example in above case if you opened `/Companies/Blackbit/Employees/John Doe` and change the assigned `Company` of `/Companies/Pimcore/Departments/Marketing` to `Blackbit`, you will be able to assign `/Companies/Pimcore/Departments/Marketing` to the Departments of `John Doe` (because the companies now match).

If an assigned object is not allowed anymore, you will get a validation exception when trying to save.

Resolves #5969


models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -73,6 +82,23 @@ public function setClasses($classes)
...
+     */
+    public function getSqlCondition(): ?string {
+        return $this->sqlCondition;
+    }
+
+    /**
+     * @param string|null $sqlCondition
+     * @return AbstractRelations
+     */
+    public function setSqlCondition(?string $sqlCondition): AbstractRelations {
```
jdreesen: I think this would be better:

```suggestion
     * @return $this
     */
    public function setSqlCondition(?string $sqlCondition): static {
```

brusch:
To be honest, even if this could be quite handy, it opens up a security issue.
I know, this should be used by privileged users only, but this won't count in case of security audits & certifications (we're currently struggling with that 😥)  
For that reason we're also removing the SQL filtering in the grid in `v11` as well as Adminer, SQL reports and other stuff, see #13224  
So basically there won't be any SQL possible anymore in the admin interface when using the default configuration. 
So, probably this is a topic for a bundle? 🤔

BlackbitDevs:
@brusch As this would really be handy and as I would like to have this feature in Pimcore rather than sending clients patch files, I thought again to find a way. I saw that other tools "invented" query languages (e.g. JIRA query language or filter languages by log analysis tools) - but they all share the problem that people would have to learn a new language. I wondered if we could use https://github.com/phpmyadmin/sql-parser to check for the query type (prevent write operations) and add some other checks to keep support for SQL querying from within backend. I mean you are now extracting the report bundle, remove the SQL button from the grid - this can all be solved with a bundle. But supporting restrictions on the allowed objects for a relation cannot because it is deeply integrated into Pimcore (search window, drag & drop etc.).

Or perhaps you could imagine a different way of getting this to work.
I could even imagine further developments of such restrictions like constraining under which parent elements it is allowed to create certain objects - because when you have 20 classes it becomes difficult to find the right one when manually creating a new object via data object tree.

At the Mountain Summit you mentioned a "query builder". Maybe we can support you with this or completely implement it. Maybe this is the solution for above features...

Looking forward to your response.

BlackbitDevs:
@brusch or @fashxp ?

brusch:
@BlackbitDevs thanks for sharing your thoughts. We had already similar discussions internally, but we haven't come to a conclusion yet how we want to deal with filters in the future. Since we're currently focusing in getting v11 done, we have postponed this topic for now.

BlackbitDevs:
@brusch, @jdreesen What do you think of a data option provider (similar to https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Data_Types/Dynamic_Select_Types.html)? The `getOptions()` would need to return a list of compatible `ElementInterface` objects.

cancan101:
Any updates on a replacement for the old "direct sql query"?

brusch:
@cancan101 yessss, awesome stuff: https://pimcore.com/docs/platform/Generic_Data_Index/Searching_For_Data_In_Index/Pimcore_Query_Language/

cancan101:
This looks very cool! Is it currently exposed to end users in the UI?

brusch:
@cancan101 see https://pimcore.com/docs/platform/Generic_Data_Index/Installation/ 😉

cancan101:
Ah yes, that makes sense, I hadn't noticed that the "Pimcore Query Language" was scoped under the "Generic Data Index".

Is there anything comparable to the old "direct SQL query" that runs against the MySQL DB and doesn't require OpenSearch?

brusch:
@cancan101 actually no 😬 Sorry

cancan101:
>> This looks very cool! Is it currently exposed to end users in the UI?

> @cancan101 see https://pimcore.com/docs/platform/Generic_Data_Index/Installation/ 😉

@brusch once the Bundle is installed, how would a user execute queries using PQL?

I don't see anything UI-wise documented at the ink you provided.

fashxp:
PQL will be supported in Studio UI. There are no plans for classic ui.

cancan101:
Do the security issues here not also apply to the custom reports: https://docs.pimcore.com/platform/Pimcore/Tools_and_Features/Custom_Reports/, especially with the comment:
>I know, this should be used by privileged users only, but this won't count in case of security audits & certifications (we're currently struggling with that 😥)

I personally would like to see a way to enable SQL (with sanitizing) even if PQL is coming soon.

People keep trying to find less good workarounds to this missing feature: https://github.com/pimcore/admin-ui-classic-bundle/pull/910 (CC @BlackbitDevs )

fashxp:
> Do the security issues here not also apply to the custom reports:

Yes they do, and we are already thinking about how to deal with that in custom reports. But, configuration of custom reports is limited to privileged uses only and they are an optional bundle.

cancan101:
If we were to use `phpmyadmin/sql-parser`, could we do something like:
```php
use SqlParser\Parser;
use SqlParser\Statements\SelectStatement;

function validateSelectQuery(string $sql): void {
    $parser = new Parser($sql);

    // 1. Ensure single query
    if (count($parser->statements) !== 1) {
        throw new Exception("Only a single query is allowed.");
    }

    $statement = $parser->statements[0];

    // 2. Ensure it's a SELECT query
    if (!$statement instanceof SelectStatement) {
        throw new Exception("Only SELECT queries are allowed.");
    }

    // 3. Optional: Check for subqueries
    foreach ($statement->where ?? [] as $expr) {
        if (stripos((string)$expr, 'SELECT') !== false) {
            throw new Exception("Subqueries in WHERE clause are not allowed.");
        }
    }
}
```

fashxp:
> phpmyadmin/sql-parser

sadly license is not compatible

cancan101:
Why do you say it's not compatible? That library is licensed under gpl v2 or later: https://github.com/phpmyadmin/sql-parser/blob/8fb3e83289abec9d13a6a3686b17495a2353bc28/composer.json#L4

https://www.gnu.org/licenses/gpl-faq.en.html#AllCompatibility

if GPL doesn't work, there is: https://github.com/greenlion/PHP-SQL-Parser which is BSD

brusch:
Yeah, GPL unfortunately doesn't work for us, since it's a copyleft license and Pimcore is not only licensed under the GPLv3.

cancan101:
Some acceptable options:
1. https://github.com/greenlion/PHP-SQL-Parser (BSD)
2. Doctrine AST parser: https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/cookbook/dql-custom-walkers.html
3. ANTLR AST: https://github.com/antlr/grammars-v4/blob/master/sql/mysql/Positive-Technologies/MySqlLexer.g4

PR #13136, opened 2022-09-12
Use hash_equals() rather than direct string comparison for password fields

NiklasBr:

A very small hardening using https://www.php.net/manual/en/function.hash-equals.php instead


mcop1:
Thanks!

PR #13140, opened 2022-09-13
Site event handling

### LINKED ISSUE #13139: [Improvement]: Add new post update event for site changes
### Improvement description

Add new SiteEvent to hook into post site updates. For example we would like to reindex our Elasticsearch after a Site is saved

- Add new SiteEvents class
- Add new SiteEvent event class
- Adjust Site Model to override save method
- Adjust Site Model to override delete method

TobiSW:
Resolves #13139 

## Additional info  
Add new SiteEvent to hook into post site updates. For example we would like to reindex our Elasticsearch after a Site is saved

### DIFF
--- a/lib/Event/Model/SiteEvent.php
+++ b/lib/Event/Model/SiteEvent.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Pimcore\Event\Model;
+
+use Pimcore\Event\Traits\ArgumentsAwareTrait;
+use Pimcore\Model\Element\ElementInterface;
+use Pimcore\Model\Site;
+use Symfony\Contracts\EventDispatcher\Event;
+
+class SiteEvent extends Event implements ElementEventInterface
+{
+    use ArgumentsAwareTrait;
+
+    /** @var Site */
+    protected $site;
+
+    /**
+     * @param Site $site
+     */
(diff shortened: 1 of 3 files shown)


brusch:
@TobiSW Thanks a lot for your PR, however, I have some change requests 😊 
- to make it consistent with other entities we should also provide additional events, just similar to redirects: https://github.com/pimcore/pimcore/blob/f97a4ec54dc5e276d0fcc96dd0c0a8a83d2f00e3/lib/Event/RedirectEvents.php#L40-L39 
- Since this is an improvement it should have `11.x` as the target branch. 

Thanks!

TobiSW:
> @TobiSW Thanks a lot for your PR, however, I have some change requests blush
> 
> * to make it consistent with other entities we should also provide additional events, just similar to redirects: https://github.com/pimcore/pimcore/blob/f97a4ec54dc5e276d0fcc96dd0c0a8a83d2f00e3/lib/Event/RedirectEvents.php#L40-L39
> * Since this is an improvement it should have `11.x` as the target branch.
> 
> Thanks!

Hi @brusch thanks for the quick reply. Understood i will adjust and add the additional events, i would close this PR as it is based on 10.5 and open another one created from 11.x to avoid merge conflicts

PR #13144, opened 2022-09-13
[Feature][DataObject] add confirm dialog to delete in manyToManyRelation

### LINKED ISSUE #13125: [Improvement]: Many-To-Many Relation - Ask for deletion
### Improvement description

When deleting a relation in a Many-to-Many Relation it should ask the user if he wants to delete the row.
Currently it is very easy to delete the row which can have a big impact.

Bildschirmfoto 2022-09-12 um 11 16 45

dpfahlbusch:
Resolves: #13125


mcop1:
@dpfahlbusch looks great, thank you. But could you please have a look at the conflicts?

dpfahlbusch:
Done :) @mcop1

brusch:
Hmmm, I'm wondering if this is always wanted ... because as long as you're not saving the data object nothing will happen. 
So I fully agree that there should be a confirmation dialog for actions which are immediately persisted, but in this case I'm not entirely sure if this is really helpful in all cases. Thoughts? @mcop1 @dpfahlbusch

dpfahlbusch:
With the autosave functionality some customers are sometimes confused, when they reload the object afterwards and the relation is still gone.
So I would say, we should ask for a confirmation and the user is safe :)

mcop1:
@brusch @dpfahlbusch I think it can happen that you remove an object from the relation without any intention. We would prevent that with the dialog. And there is, of course, the auto save feature. But at the very least I would say the dialog doesn´t hurt anyone.

brusch:
Thanks for your feedback - let's see how it works in practice and if we get any further feedback - if it's an issue, it could be removed easily anyway 😊

PR #13147, opened 2022-09-13
[Feature][DataObject] Add possibility to download asset in relations

### LINKED ISSUE #7633: [Feature request] add download button for asset in dataobject view
# Add download button for asset in dataobject view
## Current situation 

image

Currently, in the data object view, when you have a relation attribute to an asset, you can:
- search for an existing asset
- upload a new asset
- open a linked asset 
- delete the relation to the asset

I'm thinking of the following asset related attributes in data objects:
- Image
- Video
- ImageGallery
- Many-to-one relation (with 'allow assets' checked)
- Many-to-many relation (with 'allow assets' checked)

## Desired functionality
To add an asset download option for users who are primarily working in the data object views and who want to download an asset directly from the data object view.

dpfahlbusch:
Resolves #7633


brusch:
I'm wondering if that is really a very common use case ... maybe we should keep that configurable, or just provide the functionality in the context menu? Just thinking, because we actually have already quite some buttons there 😉

dpfahlbusch:
You mean adding a new checkbox in classconfig for relations ("allow inline download" or something like that)? :)

brusch:
Yeah, something like that. Or maybe we could even think about a more generic approach where you can configure the available buttons incl. adding your own buttons (or bundles) - which is then similar to use in all data-types? 🤔

dpfahlbusch:
So I would say:
In a first step I will make the buttons configurable (all of them or only a few? 🤔 )
And then in a followup we can make it generic :)

models/DataObject/ClassDefinition/Data/ManyToManyRelation.php
```diff
@@ -65,6 +65,11 @@ class ManyToManyRelation extends AbstractRelations implements QueryResourcePersi
      */
     public $maxItems;
 
+    /**
+     * @internal
+     */
+    public bool $assetInlineDownloadAllowed;
```
blankse:
```suggestion
    public bool $assetInlineDownloadAllowed = false;
```

models/DataObject/ClassDefinition/Data/ManyToManyRelation.php
```diff
@@ -663,6 +668,18 @@ public function getMaxItems()
         return $this->maxItems;
     }
 
+    public function setAssetInlineDownloadAllowed(bool $assetInlineDownloadAllowed): self
```
blankse:
```suggestion
    /**
     * @return $this
     */
    public function setAssetInlineDownloadAllowed(bool $assetInlineDownloadAllowed): static
```

blankse (changes requested):

models/DataObject/ClassDefinition/Data/ManyToManyRelation.php
dpfahlbusch: Done with last commit :)

models/DataObject/ClassDefinition/Data/ManyToManyRelation.php
dpfahlbusch: Done with last commit :)

bundles/AdminBundle/public/js/pimcore/object/tags/advancedManyToManyRelation.js
```diff
@@ -286,6 +286,27 @@ pimcore.object.tags.advancedManyToManyRelation = Class.create(pimcore.object.tag
...
+                xtype: 'actioncolumn',
+                menuText: t('download'),
+                width: 40,
+                sortable: false,
+                items: [
+                    {
+                        tooltip: t('download'),
+                        icon: "/bundles/pimcoreadmin/img/flat-color-icons/download-cloud.svg",
+                        handler: function (grid, rowIndex) {
+                            var data = grid.getStore().getAt(rowIndex);
```
robertSt7:
```suggestion
                            const data = grid.getStore().getAt(rowIndex);
```

bundles/AdminBundle/public/js/pimcore/object/tags/manyToManyRelation.js
```diff
@@ -185,8 +185,30 @@ pimcore.object.tags.manyToManyRelation = Class.create(pimcore.object.tags.abstra
...
+                xtype: 'actioncolumn',
+                menuText: t('download'),
+                width: 40,
+                sortable: false,
+                items: [
+                    {
+                        tooltip: t('download'),
+                        icon: "/bundles/pimcoreadmin/img/flat-color-icons/download-cloud.svg",
+                        handler: function (grid, rowIndex) {
+                            var data = grid.getStore().getAt(rowIndex);
```
robertSt7:
```suggestion
                            const data = grid.getStore().getAt(rowIndex);
```

robertSt7 (approved):
@dpfahlbusch When I try to download a folder I got an error. Could you please check that?
Could you also please add some docs. Thank you

robertSt7:
@dpfahlbusch Could you please fix the conflicts. Thanks

dpfahlbusch:
@robertSt7 done :)

robertSt7:
@dpfahlbusch Could you please merge 11.x again in your branch? Thanks :)

robertSt7:
@dpfahlbusch Thanks a lot :)

PR #13150, opened 2022-09-14
[Improvement] edit not editable advanced relation columns

kjkooistra-youwe:
Allow editing columns of advanced many to many (object) relation fields when not editable is enabled by providing a new option: Allow editing columns when not editable

This allows the columns to be used as configuration, while the user is unable to add or (re)move rows. Rows can be added by administrators or via event listeners, for example.

### DIFF
--- a/bundles/AdminBundle/public/js/pimcore/object/classes/data/advancedManyToManyObjectRelation.js
+++ b/bundles/AdminBundle/public/js/pimcore/object/classes/data/advancedManyToManyObjectRelation.js
@@ -184,6 +184,13 @@ pimcore.object.classes.data.advancedManyToManyObjectRelation = Class.create(pimc
 
         this.specificPanel.add({
             xtype: "checkbox",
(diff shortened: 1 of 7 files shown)


brusch:
Thank you very much for your PR. However, I think this is quite an edge case, that should be better solved by a custom data-type extending from the core type. 

Any other opinions? 
I'm closing this PR for now, please feel free to continue the discussion. 
Thanks for your understanding!

kjkooistra-youwe:
@brusch I agree extending the field would be a better approach. However, the sheer size of the createLayout function already makes that impossible, since you cannot override the conditions. Unless you copy the createLayout method and modify it, but that would mean you have to keep track of all updates in the original version of the field and apply them to the extended version, which isn't a very appealing idea.

An alternative might be to refactor the JavaScript to make it easier to extend. That would require quite a bit of effort though.

brusch:
Yep, refactoring the code absolutely makes sense anway 😊

PR #13151, opened 2022-09-14
Make editmodeImagePreview available for YouTube videos

bitfactory-jurgen-jansma:
The config `editmodeImagePreview` for the video editable was not working for YouTube videos. I made it working in this pull request. 

## Additional info  
I'm not sure if you see this as a bug fix or as a feature/improvement. But, in our case, this solves a bug in loading the editor in the admin panel, so that's why I submitted this pull request in 10.5.

### DIFF
--- a/models/Document/Editable/Video.php
+++ b/models/Document/Editable/Video.php
@@ -295,11 +295,11 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
         } elseif ($this->type === self::TYPE_ASSET) {
             return $this->getAssetCode($inAdmin);
         } elseif ($this->type === self::TYPE_YOUTUBE) {
-            return $this->getYoutubeCode();
+            return $this->getYoutubeCode($inAdmin);
         } elseif ($this->type === self::TYPE_VIMEO) {
-            return $this->getVimeoCode();
+            return $this->getVimeoCode($inAdmin);
         } elseif ($this->type === self::TYPE_DAILYMOTION) {
-            return $this->getDailymotionCode();
+            return $this->getDailymotionCode($inAdmin);
         } elseif ($this->type === 'url') {
             return $this->getUrlCode();
         }
@@ -615,7 +615,7 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
     /**
      * @return string
      */
-    private function getYoutubeCode()
+    private function getYoutubeCode($inAdmin = false)
     {
         if (!$this->id) {
             return $this->getEmptyCode();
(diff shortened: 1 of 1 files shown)


brusch:
@bitfactory-jurgen-jansma thanks a lot for your PR! 
2 thinks we should consider as well: 
1. We should also use `$inAdmin` argument for `getYoutubeCode()` same as for `getAssetCode()`, see: https://github.com/pimcore/pimcore/blob/b447bc02341a2eb097bc48c2f6e9f3f84d708829/models/Document/Editable/Video.php#L457
2. to be consisten we should try to fix that for Vimeo and Dailymotion as well

Thanks a lot!

bitfactory-jurgen-jansma:
@brusch thank you for your review. I have used the `$inAdmin` argument instead of `$this->editmode` now. I have also added a thumbnail for Vimeo and Dailymotion as well.

models/Document/Editable/Video.php
```diff
@@ -615,7 +615,7 @@ private function parseYoutubeId()
     /**
      * @return string
      */
-    private function getYoutubeCode()
+    private function getYoutubeCode($inAdmin = false)
```
blankse:
```suggestion
    private function getYoutubeCode(bool $inAdmin = false): string
```

models/Document/Editable/Video.php
```diff
@@ -710,7 +716,7 @@ private function getYoutubeCode()
     /**
      * @return string
      */
-    private function getVimeoCode()
+    private function getVimeoCode($inAdmin = false)
```
blankse:
```suggestion
    private function getVimeoCode(bool $inAdmin = false): string
```

models/Document/Editable/Video.php
```diff
@@ -789,7 +801,7 @@ private function getVimeoCode()
     /**
      * @return string
      */
-    private function getDailymotionCode()
+    private function getDailymotionCode($inAdmin = false)
```
blankse:
```suggestion
    private function getDailymotionCode(bool $inAdmin = false): string
```

models/Document/Editable/Video.php
```diff
@@ -789,7 +801,7 @@ private function getVimeoCode()
     /**
      * @return string
      */
```
blankse:
```suggestion
```

models/Document/Editable/Video.php
```diff
@@ -710,7 +716,7 @@ private function getYoutubeCode()
...
```
blankse:
```suggestion
```

models/Document/Editable/Video.php
```diff
@@ -615,7 +615,7 @@ private function parseYoutubeId()
...
```
blankse:
```suggestion
```

blankse (changes requested):
Please use type hints for new params

bitfactory-jurgen-jansma:
@blankse I didn't add the type hinting because it wasn't use yet, and I expected that it wasn't there for a reason. However, I have added the type hinting to all edited methods.

brusch:
Failing Codeception tests are not related to this PR ... caused by #13171 (fix)

PR #13157, opened 2022-09-15
[Docs / Wysiwyg fields] Fix wysiwyg editor example config

BlackbitDevs:
Since Pimcore 10 the example editor config in https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Data_Types/Text_Types.html#page_Editor-Configuration does not work anymore. The reason is that in https://github.com/pimcore/pimcore/blob/1be12b0493bc9153783c8d85210f9863637aec3c/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/wysiwyg.js#L193 `Ext.decode()` gets used - but the difference with Pimcore 10 is that in this function
```javascript
try {
            // check USE_NATIVE_JSON here so it can be changed if needed
            if (hasNative && Ext.USE_NATIVE_JSON) {
                return JSON.parse(json);
            }
            return doDecode(json);
        } catch (e) {
            if (safe) {
                return null;
            }
            Ext.raise({
                sourceClass: "Ext.JSON",
                sourceMethod: "decode",
                msg: "You're trying to decode an invalid JSON String: " + json
            });
        }
```

`Ext.USE_NATIVE_JSON` is `true` whereas in Pimcore 6 it was `false`. Thus in Pimcore 6 the `doDecode()` function was used which is much more tolerant. The native `JSON.parse()` is really picky. Back to the docs example:
```markdown
{
  toolbarGroups : [ { name: 'links' }],
  enterMode: CKEDITOR.CKEDITOR.ENTER_BR,
}
```
This does not work for the following reasons:
- JSON attributes need to be quoted
- trailing comma after `enterMode` not allowed
- `CKEDITOR.CKEDITOR.ENTER_BR` got renamed to `CKEDITOR.ENTER_BR`
- constants cannot be resolved(!)

Especially the last point is really bad because you cannot use the generated configurations from [the linked CKEditor configuration docs](https://ckeditor.com/docs/ckeditor4/latest/features/toolbar.html) or the [CKEditor configurator](https://ckeditor.com/latest/samples/toolbarconfigurator). Because of this this PR does not only fix the docs but disables `Ext.USE_NATIVE_JSON` for this.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/wysiwyg.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/wysiwyg.js
@@ -190,7 +190,7 @@ pimcore.object.tags.wysiwyg = Class.create(pimcore.object.tags.abstract, {
         }
 
         if(this.fieldConfig.toolbarConfig) {
-            var useNativeJson = Ext.USE_NATIVE_JSON;
+            const useNativeJson = Ext.USE_NATIVE_JSON;
             Ext.USE_NATIVE_JSON = false;
             var elementCustomConfig = Ext.decode(this.fieldConfig.toolbarConfig);
             Ext.USE_NATIVE_JSON = useNativeJson;
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/wysiwyg.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/wysiwyg.js
@@ -190,7 +190,10 @@ pimcore.object.tags.wysiwyg = Class.create(pimcore.object.tags.abstract, {
         }
 
         if(this.fieldConfig.toolbarConfig) {
+            var useNativeJson = Ext.USE_NATIVE_JSON;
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/tags/wysiwyg.js
```diff
@@ -190,7 +190,10 @@ pimcore.object.tags.wysiwyg = Class.create(pimcore.object.tags.abstract, {
         }
 
         if(this.fieldConfig.toolbarConfig) {
+            var useNativeJson = Ext.USE_NATIVE_JSON;
```
robertSt7:
```suggestion
            const useNativeJson = Ext.USE_NATIVE_JSON;
```

robertSt7 (changes requested):
Could you also please merge 10.5 in your branch to fix the tests? Thank you

BlackbitDevs:
@robertSt7 Done.

robertSt7:
@BlackbitNeueMedien Thanks a lot

PR #13160, opened 2022-09-15
[Calculated value fields] In grid view do not strip HTML tags if display type is "HTML"

BlackbitDevs:
Steps to reproduce:
1. Create calculator which implements `CalculatorClassInterface` and in `compute()` returns
   ```html
   <div style="width:100%">
     <div style="width:300px;background-color:#e0e0e0;padding:3px;border-radius:3px;box-shadow:inset 0 1px 3px rgba(0, 0, 0, .2);">
       <span class="progress-bar-fill" style="width:53%;display:block;height:22px;background-color:#659cef;border-radius:3px;transition:width 500ms ease-in-out;"></span>
     </div>
   </div>
   ```
2. Create data object class with calculated value field and `Type = HTML` and enable `Visible in grid view`
3. Create object of this class
4. open grid

Before this PR you will see:
Bildschirmfoto 2022-09-15 um 09 38 16

With this PR you will see:
Bildschirmfoto 2022-09-15 um 09 38 57

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/grid.js
@@ -380,9 +380,9 @@ pimcore.object.helpers.grid = Class.create({
 
                 } else {
                     var fieldType = fields[i].type;
-                    var tag = pimcore.object.tags[fieldType];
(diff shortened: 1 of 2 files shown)


robertSt7:
@BlackbitNeueMedien Thanks for the PR

PR #13164, opened 2022-09-15
Fieldcollection/Objectbrick/CustomLayout/Bulk Export: Cleanup BlockedVars and Dynamic Options like ClassDefinition

blankse:
In Fieldcollection/Objectbrick/CustomLayout/Bulk exports are the deprecated "childs" Attributes and all unnecessary attributes (like the dynamic options). This should be removed like in the class definition export.


robertSt7:
@blankse Thank you very much

PR #13172, opened 2022-09-15
Fix external link detection in link edit panel

jdreesen:
Until now, external links in the link edit panel are detected by checking if they start with `www.`, which is not sufficient, because nowadays, many websites don't have a `www` prefix anymore.

From now on, it looks if the value starts with a single `/` (as Pimcore assets/documents/objects are all relative to the host (as far as I know)). Everything else (including protocol-relative URLs like `//google.de`) are considered external.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
@@ -1794,7 +1794,7 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
                 // if it doesn't start with a single "/", we assume it's an external link
                 if (value && !value.startsWith('/') && !'http://'.startsWith(value) && !'https://'.startsWith(value)) {
                     if (!value.match(/^https?:\/\//)) {
-                        el.setValue("http://" + value.replace(/^\/+/, ''));
+                        el.setValue("https://" + value.replace(/^\/+/, ''));
                     }
                     internalTypeField.setValue(null);
                     linkTypeField.setValue("direct");
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
@@ -1786,12 +1786,13 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
         name: "path",
         width: 520,
         fieldCls: "pimcore_droptarget_input",
+        enableKeyEvents: true,
         listeners: {
-            blur: function (el) {
+            keyup: function (el) {
                 const value = el.getValue();
 
                 // if it doesn't start with a single "/", we assume it's an external link
-                if (!value.match(/^\/[^/]+/)) {
+                if (value && !value.startsWith('/') && !'http://'.startsWith(value) && !'https://'.startsWith(value)) {
                     if (!value.match(/^https?:\/\//)) {
                         el.setValue("http://" + value.replace(/^\/+/, ''));
                     }
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
@@ -1799,6 +1799,11 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
                     internalTypeField.setValue(null);
                     linkTypeField.setValue("direct");
                 }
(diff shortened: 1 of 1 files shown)


kingjia90:
The idea sounds good, but there is something that is not working properly at the moment, could you please check this?

I've tried to just type `/` and it automatically prepends, resulting in `http:///`.
Then, if you attempt to empty the field again using backspace, it would turn into `http://http:/`.
If you select all and delete, it would turn to `http://` again.

It only works by drag&dropping, search&select or pasting a string eg. `/abc/qwe`, but cannot manually type in it.

jdreesen:
Oops, you're totally right. I pushed a new version and hope it works now.

kingjia90:
> Oops, you're totally right. I pushed a new version and hope it works now.

Thank you for the update but it still doesn't work as expected, because for example it still tries to prepend http when the field is empty and it doesn't allow you to manually type `http://`, also i would like to avoid the `blur` event.

What about reverting to `keyup` and add this conditions:
`if (value && !value.startsWith('/') && 'http://'.indexOf(value) !== 0 && 'https://'.indexOf(value) !== 0) {` 
instead of  `if (!value.match(/^\/[^/]+/)) {`

The purpose of `'http://'.indexOf(value) !== 0` is to avoid the auto-prepend when the user may be trying to type `http://`, so it will not trigger when typing `htt`, but it will then prepend if you type `htta`.

wdyt?

jdreesen:
That seems to do the trick! Thx :)

May I ask what's wrong with the blur event?

/edit: this doesn't support protocol-relative URLs anymore, but maybe that isn't necessary, as I don't think they're very common.

kingjia90:
> May I ask what's wrong with the blur event?

IMHO the blur makes the UI feeling "sluggish".
It's also some kind of distracting: when you are editing and move on on next field/context, you see the former field flashing/changing due some script in a event requiring ones' attention again and eventually move cursor back there to double-check. 
In this case specifically, i'd prefer the keyup, it feels more fast and "active", like some of those real-time autocomplete 😄

kingjia90:
> /edit: this doesn't support protocol-relative URLs anymore, but maybe that isn't necessary, as I don't think they're very common.

Nice input, i haven't considered that. 
If these url are starting as `//pimcore.com` theoretically the first character also being `/`, it doesn't make the http protocol to be automatically prepended. 
We could add a further `if` block whenever we see the field value starting as `//` to set the link type as direct. 

Another input is that we are preferring `http` over `https` as default, while it should be likely the opposite.

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
```diff
@@ -1789,8 +1789,13 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
...
         listeners: {
             keyup: function (el) {
-                if (el.getValue().match(/^www\./)) {
-                    el.setValue("http://" + el.getValue());
+                const value = el.getValue();
+
+                // if it doesn't start with a single "/", we assume it's an external link
+                if (value && !value.startsWith('/') && !'http://'.startsWith(value) && !'https://'.startsWith(value)) {
+                    if (!value.match(/^https?:\/\//)) {
+                        el.setValue("http://" + value.replace(/^\/+/, ''));
```
kingjia90:
```suggestion
                        el.setValue("https://" + value.replace(/^\/+/, ''));
```

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
```diff
@@ -1789,8 +1789,13 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
...
+                    }
                     internalTypeField.setValue(null);
                     linkTypeField.setValue("direct");
                 }
```
kingjia90, line 1801: after the closing bracket at 1802, something like
```
// if it starts with "//", we assume it is a protocol-relative URL
if (value.startsWith('//'){
   internalTypeField.setValue(null);
   linkTypeField.setValue("direct");
}
```

kingjia90 (changes requested):
suggesting some minor changes as mentioned in the last comments

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
jdreesen: Done

jdreesen:
I added your suggestions. Thank you very much!

kingjia90 (approved):

aashan10 (approved):

PR #13173, opened 2022-09-15
Make link edit panel configurable

### LINKED ISSUE #4184: Allow restricting Link editable fields
# Feature Request

## Feature description

It seems that there's no way to restrict fields of the Link editable:

```twig
{%- set button = pimcore_link('button-' ~ column) -%}
```

I'm only reading out the link title and href with the provided `.getText()` and `.getHref()` methods, nothing else. I don't need "Properties" or "Advanced":

Users may be confused because they can still enter it, even though it's not used in the frontend. For me, even the implemented `noText` config option didn't have an impact on the editing itself – I could still see the "Text" field.

jdreesen:
This introduces three configuration options for the link edit panel:

1. `allowedTypes`, which may contain the values `asset`, `document` and/or `object`. 
   This allows you to restrict a link editable to certain types.

2. `allowedTargets`, which may contain the values ` `, `_blank`, `_self`, `_top` and/or `_parent`.
   This allows you to e.g. only allow ` ` and `_blank`, when you don't use frames on your page (because the other ones don't make sense then).

3. `disabledFields`, which may contain the values `text`, `target`, `parameters`, `anchor`, `title`, `accesskey`, `rel`, `tabindex`, `class` and/or `attributes`.
   This allows you to hide/remove certain fields when you don't need/support them and don't want to confuse the editors with useless fields.


kingjia90:
Thank you for the PR!

I've reviewed it and it works fine, just some feeedback and ideas:

- Could you please extend this features to `Link Data Type` too? It should be customizable in the UI under `Specific Settings`.
For reference, there was something similar done for a video data type https://github.com/pimcore/pimcore/pull/12846 and it was formerly only changing the  document editable.
- What about a sort `preset/alias/group` that allow you to hide the fields in `Advanced` tab (which is probably too much for the regular user anyway)? In other words, instead of typing all the 6 fields name individually, just one parameter value to indentify them.
- Would be good to add the "whitelisting" version of `disabledFields`, eg. `allowedFields`?

bundles/AdminBundle/public/js/pimcore/helpers.js
```diff
@@ -1788,41 +1794,145 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
...
+    }
+    if (!disabledFields.includes('attributes')) {
+        advancedFields.push({
+            fieldLabel: t('attributes') + ' (key="value")',
+            name: 'attributes',
+            width: 300,
+            value: data["attributes"]
+        });
+    }
+    const advancedTab = propertyFields.length === 0 ? null : {
```
kingjia90:
```suggestion
    const advancedTab = advancedFields.length === 0 ? null : {
```

jdreesen:
> - Could you please extend this features to Link Data Type too? It should be customizable in the UI under Specific Settings.
    For reference, there was something similar done for a video data type https://github.com/pimcore/pimcore/pull/12846 and it was formerly only changing the document editable.

Sure, will do.

> - What about a sort preset/alias/group that allow you to hide the fields in Advanced tab (which is probably too much for the regular user anyway)? In other words, instead of typing all the 6 fields name individually, just one parameter value to indentify them.

I thought about that, too, but wasn't sure what's the best way to do that. 
Maybe just two additional config options `disableProperties: bool` and `disableAdvanced: bool`?
Or something like: `allowedGroups: ['properties', 'advanced']`.

> - Would be good to add the "whitelisting" version of disabledFields, eg. allowedFields?

Do you mean `allowedFields` instead of `disabledFields`, or in addition to them?
We could do the first, if you think it's better. I used `disabledFields` in the assumption that you most of the time only want to disable a few fields. But maybe that's not the case, and we should better be consistent here.
Allowing both would be confusing, I think.

kingjia90:
The problem of having additional config options like `disableProperties` would be that there would be conflicting cases where you disable the group but allow a single field of it or so. 

My initial though was `allowedFields` accepting values like `advanced` as placeholder/keyword so it would look like `['title','href','properties','accessKey', 'advanced']` and a function to replace it with the specific fields, ignoring duplicates.
Not sure if it is bad practice here to mix the available html attributes with other textual stuff.

Even better would be just defining the panel content in a 3 dimension array, where the first level is the tab name, the second level are the fields as items, if one of the items in the second level is an array, it turns it into a fieldset with the item key as fieldsename and level 3 array as items. 
Whenever the key is found but sub-array is empty/undefined, it would be filled with default values eg. `['basic']` but may be an overkill.

kingjia90:
> > * Would be good to add the "whitelisting" version of disabledFields, eg. allowedFields?
> 
> Do you mean `allowedFields` instead of `disabledFields`, or in addition to them? We could do the first, if you think it's better. I used `disabledFields` in the assumption that you most of the time only want to disable a few fields. But maybe that's not the case, and we should better be consistent here. Allowing both would be confusing, I think.

Mhh, true, it can be also weird if anyone adds the same field in both config options (allowed/disabled) and can't tell which rule has more priority. 
So it depends on typical usage, if most of the time it's only used to hide few fields, then disabledFields is optimal, on the contrary, if you just want to use a few ones, maybe the most basic like title and href, then one would love allowedFields 🤔

kingjia90:
@jdreesen i've asked the team, `disabledFields` is preferred here, as it is more "close" to the idea of `empty/unset` is `all options selected`.
Looking forward for the changes to the Link Data Type

kingjia90:
@jdreesen Just writing for a reminder about the Link data type 😄

jdreesen:
Thx. I have it on my radar. Unfortunately, I'm very busy at the moment and probably won't get around to it in October.

kingjia90:
Ah ok, no problem, it's for 11.x anyways, so not that urgent, but thank you for the status update, that helps me to change its status in the planning board (in other words postpone to the right iteration).

fashxp:
@kingjia90  what is the state here?

jdreesen:
The link data type is still missing, and I'm still very busy. I try to find some time to implement it.

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -416,4 +420,37 @@ public function getFieldType(): string
     {
         return 'link';
     }
+
+    public function getAllowedTypes(): array
+    {
+        return $this->allowedTypes;
+    }
+
+    public function setAllowedTypes(array $allowedTypes): static
```
blankse:
```suggestion
    /**
     * @param string[] $allowedTypes
     *
     * @return $this
     */
    public function setAllowedTypes(array $allowedTypes): static
```

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -416,4 +420,37 @@ public function getFieldType(): string
...
+        $this->allowedTypes = $allowedTypes;
+        return $this;
+    }
+
+    public function getAllowedTargets(): array
+    {
+        return $this->allowedTargets;
+    }
+
+    public function setAllowedTargets(array $allowedTargets): static
```
blankse:
```suggestion
    /**
     * @param string[] $allowedTargets
     *
     * @return $this
     */
    public function setAllowedTargets(array $allowedTargets): static
```

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -416,4 +420,37 @@ public function getFieldType(): string
...
+        $this->allowedTargets = $allowedTargets;
+        return $this;
+    }
+
+    public function getDisabledFields(): array
+    {
+        return $this->disabledFields;
+    }
+
+    public function setDisabledFields(array $disabledFields): static
```
blankse:
```suggestion
    /**
     * @param string[] $disabledFields
     *
     * @return $this
     */
    public function setDisabledFields(array $disabledFields): static
```

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -30,6 +30,10 @@ class Link extends Data implements ResourcePersistenceAwareInterface, QueryResou
     use DataObject\Traits\SimpleComparisonTrait;
     use DataObject\Traits\ObjectVarTrait;
 
+    public array $allowedTypes;
+    public array $allowedTargets;
+    public array $disabledFields;
```
blankse:
```suggestion
    /**
     * @var string[]
     */
    public array $allowedTypes;

    /**
     * @var string[]
     */
    public array $allowedTargets;

    /**
     * @var string[]
     */
    public array $disabledFields;
```

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -416,4 +420,37 @@ public function getFieldType(): string
     {
         return 'link';
     }
+
+    public function getAllowedTypes(): array
```
blankse:
```suggestion
    /**
     * @return string[]
     */
    public function getAllowedTypes(): array
```

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -416,4 +420,37 @@ public function getFieldType(): string
...
+        return $this->allowedTypes;
+    }
+
+    public function setAllowedTypes(array $allowedTypes): static
+    {
+        $this->allowedTypes = $allowedTypes;
+        return $this;
+    }
+
+    public function getAllowedTargets(): array
```
blankse:
```suggestion
    /**
     * @return string[]
     */
    public function getAllowedTargets(): array
```

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -416,4 +420,37 @@ public function getFieldType(): string
...
+        return $this->allowedTargets;
+    }
+
+    public function setAllowedTargets(array $allowedTargets): static
+    {
+        $this->allowedTargets = $allowedTargets;
+        return $this;
+    }
+
+    public function getDisabledFields(): array
```
blankse:
```suggestion
    /**
     * @return string[]
     */
    public function getDisabledFields(): array
```

blankse (changes requested):
Please add php docs for `arrays` and `return $this`

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -30,6 +30,21 @@ class Link extends Data implements ResourcePersistenceAwareInterface, QueryResou
     use DataObject\Traits\SimpleComparisonTrait;
     use DataObject\Traits\ObjectVarTrait;
 
+    /**
+     * @var string[]
+     */
+    public array $allowedTypes;
```
blankse: Shouldn't the default be null if not set? So that everything is then allowed?
So it will throw an error for existing definitions or definition via php script.

kingjia90: mhh nice point, didn't consider that.. Can it be just a an empty array by default or does it have drawbacks compared to being null and nullable?

jdreesen:
Thank you so much for taking this over ❤️

kingjia90 (approved):

models/DataObject/ClassDefinition/Data/Link.php
```diff
@@ -416,4 +431,61 @@ public function getFieldType(): string
...
+    {
+        return $this->allowedTypes;
+    }
+
+    /**
+     * @param string[] $allowedTypes
+     *
+     * @return $this
+     */
+    public function setAllowedTypes(array $allowedTypes): static
```
blankse: Shouldn't be the setters also nullable?

kingjia90: technically yes, but iirc we do like the current code, that it is only `null` on initialization but afterwards (eg. saving in class definition) it would be an empty array anyways.

kingjia90: nevermind, pushed the changes as suggested, thank you!
I guess i was recalling something prior 11 because i can't find an example anymore and/or my brain is just a little fried since is `Frieday` 😄

blankse: But empty array will allow nothing, right?
In Js a empty array is true (different than in PHP)
console.log([] || ['test']); -> []

PR #13177, opened 2022-09-16
Remove the BC layer which allows the getNavigation() method call with extra arguments

### LINKED ISSUE #12453: [Improvement]: Remove the BC layer which allows the getNavigation() method call with extra arguments
### Improvement description

Remove the BC layer in Pimcore\Navigation\Builder::getNavigation() method which allows the call of getNavigation() method using extra arguments. From pimcore 11, the extra arguments are not allowed and instead, a param style array should be used as the argument.

Follow up to #12310 and #12572

aryaantony92:
Resolves #12453


lib/Navigation/Builder.php
```diff
@@ -114,38 +114,25 @@ protected function resolveOptions(array $options): array
...
-     * @param array|Document|null $activeDocument
-     * @param Document|null $navigationRootDocument
-     * @param string|null $htmlMenuIdPrefix
-     * @param \Closure|null $pageCallback
-     * @param bool|string $cache
-     * @param int|null $maxDepth
-     * @param int|null $cacheLifetime
-     *
+     * @param array $params
      * @return mixed|\Pimcore\Navigation\Container
```
blankse: Should only return `\Pimcore\Navigation\Container`, right? So we can add this as return type hint. Maybe we can optimize the part with `CacheManager::load()` from line 144 to 167

```
        $navigation = null;
        if ($cacheEnabled) {
            // the cache key consists out of the ID and the class name (eg. for hardlinks) of the root document and the optional html prefix
            $cacheKeys = ['root_id__' . $navigationRootDocument->getId(), $htmlMenuIdPrefix, get_class($navigationRootDocument)];
            if (Site::isSiteRequest()) {
                $site = Site::getCurrentSite();
                $cacheKeys[] = 'site__' . $site->getId();
            }
            if (is_string($cache)) {
                $cacheKeys[] = 'custom__' . $cache;
            }
            if ($pageCallback instanceof \Closure) {
                $cacheKeys[] = 'pageCallback_' . closureHash($pageCallback);
            }
            if ($maxDepth) {
                $cacheKeys[] = 'maxDepth_' . $maxDepth;
            }
            $cacheKey = 'nav_' . md5(serialize($cacheKeys));
            $navigation = CacheManager::load($cacheKey);
        }
        if (!$navigation instanceof \Pimcore\Navigation\Container) {
```

This can be optimized also in 10.5. I created an PR: https://github.com/pimcore/pimcore/pull/13183

aryaantony92: @blankse Thanks, you are right. It will return \Pimcore\Navigation\Container. Accepted the php Doc change.

lib/Navigation/Builder.php
```diff
@@ -114,38 +114,25 @@ protected function resolveOptions(array $options): array
...
-     * @param int|null $cacheLifetime
-     *
-     * @return mixed|\Pimcore\Navigation\Container
+     * @param array $params
+     * @return Container
      *
      * @throws \Exception
      */
-    public function getNavigation($activeDocument = null, $navigationRootDocument = null, $htmlMenuIdPrefix = null, $pageCallback = null, $cache = true, ?int $maxDepth = null, ?int $cacheLifetime = null)
+    public function getNavigation(array $params)
```
blankse: @aryaantony92 We should add it also as type hint
```suggestion
    public function getNavigation(array $params): Container
```

aryaantony92: But, we decided to get rid off duplicate info. So, in this case, I think the phpDoc is complete.

blankse: Yes if you add the type hint here you can remove it from php doc. But I think we should prefer type hints.

aryaantony92: https://github.com/pimcore/pimcore/blob/e0f4ec4e5f1bdece047aab175bd14e41b9f52053/lib/Navigation/Builder.php#L116 to https://github.com/pimcore/pimcore/blob/e0f4ec4e5f1bdece047aab175bd14e41b9f52053/lib/Navigation/Builder.php#L121

jdreesen: > But, we decided to get rid off duplicate info. So, in this case, I think the phpDoc is complete.

Then... why is there a parameter type? ;)

I'm with @blankse that generally real types should be preferred over PhpDoc.

blankse:
```
    /**
     * @param array{
     *     root?: ?Document,
     *     htmlMenuPrefix?: ?string,
     *     pageCallback?: ?callable,
     *     cache?: string|bool,
     *     cacheLifetime?: ?int,
     *     maxDepth?: ?int,
     *     active?: ?Document,
     *     markActiveTrail?: bool
     * } $params
     *
     * @throws \Exception
     */
    public function getNavigation(array $params): Container
```
I think so it would be perfect :) So the array keys are also documented. (It's also required for PhpStan level 6)

aryaantony92: @blankse Accepted the change. Thanks!

dvesh3 (approved):

PR #13183, opened 2022-09-16
Navigation Builder shouldn't load from Cache if cache is disabled

blankse:
Noticed in the review of https://github.com/pimcore/pimcore/pull/13177

### DIFF
--- a/lib/Navigation/Builder.php
+++ b/lib/Navigation/Builder.php
@@ -155,7 +155,7 @@ class Builder
         }
 
         $navigation = null;
-        $cacheKeys = [];
+        $cacheKey = null;
         if ($cacheEnabled) {
             // the cache key consists out of the ID and the class name (eg. for hardlinks) of the root document and the optional html prefix
             $cacheKeys = ['root_id__' . $navigationRootDocument->getId(), $htmlMenuIdPrefix, get_class($navigationRootDocument)];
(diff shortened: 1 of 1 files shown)


lib/Navigation/Builder.php
```diff
@@ -154,30 +154,33 @@ public function getNavigation($activeDocument = null, $navigationRootDocument =
             $navigationRootDocument = Document::getById(1);
         }
 
-        // the cache key consists out of the ID and the class name (eg. for hardlinks) of the root document and the optional html prefix
-        $cacheKeys = ['root_id__' . $navigationRootDocument->getId(), $htmlMenuIdPrefix, get_class($navigationRootDocument)];
+        $navigation = null;
+        $cacheKeys = [];
```
jdreesen: I don't think this is necessary.

blankse: I think also. But PhpStan dont :(

jdreesen: PhpStan complains about `$cacheKey`, not `$cacheKeys`. Apparently, it's not smart enough to understand that it gets set in `if ($cacheEnabled) ...` so it's available inside the next `if ($cacheEnabled) ...`, too.

blankse: Yes i mixed up the variable name in the rush. Should fit now: https://github.com/pimcore/pimcore/pull/13183/commits/b4db0cafa61992604e954216217945cb9424138d

jdreesen (approved):

aryaantony92:
@blankse Thanks!

Issue #12559, opened 2022-06-29
Use new bundle folder structure for core bundles


brusch:
### Improvement description

All bundles in `/bundles` should follow the new directory structure. 

See also #12356

- [x] AdminBundle #12885
- [x] CoreBundle #13163
- [x] EcommerceBundle #13184
- [x] InstallBundle #13165

PR #13184, opened 2022-09-16
[EcommerceFrameworkBundle] Refactored as per modern bundle structure

dvesh3:
Resolves partially #12559


aashan10:
@dvesh3 please fix the conflicts :)

kingjia90 (commented):
Confirming the problem that Arya found about add/update pricing rule, but it's not caused by this PR.
The root of the problem is `condition` being a [reserved word](https://mariadb.com/kb/en/reserved-words/) but not being automatically quoted. The weird thing is that on Demo it works.

Opening a new issue for it https://github.com/pimcore/pimcore/issues/13336

dvesh3:
> @dvesh3 please fix the conflicts :)

done.

kingjia90:
Update: I've found the culprit for the pricing rule not working on 11.x, https://github.com/pimcore/pimcore/pull/13373 so we can now safely merge this one.

PR #13185, opened 2022-09-18
[Improvement][Dataobject] Add possiblity to disable creation of new object in reverseRelation

### LINKED ISSUE #11679: [Feature]: Make Creating New Objects in Reverse-Many-To-Many-Object-Relations optional
### Feature description

Dear Pimcore Community,

As with "normal" Many-To-Many-relations. Would it not better to add an option to the Reverse Many-To-Many-Object-Relations that it would become optional whether there is this button to create a new Object. 
Currently, we have a use case where users should be able to see the relation but not being able to actually create new Data Objects.

I have looked into this, and all it would currently require would to add the code from the Many-To-Many-Relation  (which the Reverse M2MO-Relation inherits), since all the required code on the PHP-side is available to the [:reverseObjectRelation.js](https://github.com/pimcore/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/reverseObjectRelation.js)

This has to be added to `this.specificPanel.add()` in the `getLayout: function ($super)`
`{
                xtype: "checkbox",
                boxLabel: t("allow_to_create_new_object"),
                name: "allowToCreateNewObject",
                value: this.datax.allowToCreateNewObject
            }`

and in the end 
`allowToCreateNewObject: source.datax.allowToCreateNewObject`
has to be added to the `applySpecialData: function(source)`

However, there is a bug that I am facing that the owner class / owner field is not saved properly and seems to disappear, under a certain change to the "Allow to Create New Object"-Checkbox in the Class editor. Therefore, I have not created a pull request. I have added a video (turned .gif) below and can provide the corresponding change to the file.

allowtocreateobject_cut
[reverseObjectRelation.txt](https://github.com/pimcore/pimcore/files/8262146/reverseObjectRelation.txt)

dpfahlbusch:
Resolves #11679

### DIFF
--- a/bundles/AdminBundle/public/js/pimcore/object/classes/data/reverseObjectRelation.js
+++ b/bundles/AdminBundle/public/js/pimcore/object/classes/data/reverseObjectRelation.js
@@ -56,6 +56,12 @@ pimcore.object.classes.data.reverseObjectRelation = Class.create(pimcore.object.
         this.specificPanel.removeAll();
         this.specificPanel.add([
             {
+                xtype: "checkbox",
+                boxLabel: t("allow_to_create_new_object"),
+                name: "allowToCreateNewObject",
+                value: this.datax.allowToCreateNewObject
+            },
+            {
                 xtype: "textfield",
                 fieldLabel: t("width"),
                 name: "width",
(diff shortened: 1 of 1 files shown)


mcop1:
Works great, thanks!

PR #12999, opened 2022-08-30
Boolean Attributes should always be a boolean

blankse:
Splitted from https://github.com/pimcore/pimcore/pull/12862 for 11.x

PR #13187, opened 2022-09-19
AdvancedManyToManyObjectRelation: Cast to boolean

blankse:
Boolean attributes should be cast to boolean in the setters to support old fielddefinitions with null values.

Followup to https://github.com/pimcore/pimcore/pull/12999

See codeception error in https://github.com/pimcore/customer-data-framework/pull/336


robertSt7:
@blankse Thank you

PR #13194, opened 2022-09-19
Add type hints to private methods

blankse:
Add type hints to all private methods and removes redundant php docs.


bundles/AdminBundle/src/Controller/Admin/Document/PrintpageControllerBase.php
```diff
@@ -306,12 +306,7 @@ public function getProcessingOptionsAction(Request $request): JsonResponse
         return $this->adminJson(['options' => $returnValue]);
     }
 
-    /**
-     * @param int $documentId
-     *
-     * @return array|mixed
-     */
-    private function getStoredProcessingOptions($documentId): mixed
+    private function getStoredProcessingOptions(int $documentId): mixed
```
aashan10: Since we are already using php 8.0 wouldn't it be good if we used union type here instead of eliminating the array return type?

bundles/EcommerceFrameworkBundle/Type/Decimal.php
```diff
@@ -745,12 +731,8 @@ public function discountPercentageOf(Decimal $other)
 
     /**
      * Transforms operand into a numeric value used for calculations.
-     *
-     * @param mixed $operand
-     *
-     * @return float
      */
-    private function getScalarOperand($operand)
+    private function getScalarOperand(mixed $operand): float|int|string
```
aashan10: Can we make this consistent with previous docblock return type somehow? Maybe cast the value to float before returning?

lib/Workflow/MarkingStore/DataObjectSplittedStateMarkingStore.php
```diff
@@ -136,13 +136,7 @@ public function getMappedPlaces(string $fieldName)
...
     }
 
-    /**
-     *
-     * @param string|PropertyPathInterface $property
-     * @param mixed $places
-     *
-     */
-    private function setProperty(Concrete $subject, $property, $places)
+    private function setProperty(Concrete $subject, string $property, mixed $places): void
```
aashan10, line 138:
```suggestion
    private function setProperty(Concrete $subject, string | PropertyPathInterface $property, mixed $places): void
```

models/DataObject/ClassDefinition/Data/EncryptedField.php
```diff
@@ -142,15 +136,9 @@ private function encrypt($data, $object, $params = [])
...
-     * @param string|null $data
-     * @param Model\DataObject\Concrete $object
-     * @param array $params
-     *
-     * @return string|null
-     *
      * @throws \Exception
      */
-    private function decrypt($data, $object, $params = [])
+    private function decrypt(?string $data, Model\DataObject\Concrete $object, array $params)
```
aashan10:
```suggestion
    private function decrypt(?string $data, Model\DataObject\Concrete $object, array $params): ?string
```

models/DataObject/ClassDefinition/Data/Select.php
```diff
@@ -150,7 +150,7 @@ public function setColumnLength($columnLength)
      *
      * @param string $type
      */
-    private function correctColumnDefinition($type)
+    private function correctColumnDefinition(string $type)
```
aashan10:
```suggestion
    private function correctColumnDefinition(string $type): void
```

models/DataObject/Localizedfield.php
```diff
@@ -415,13 +415,6 @@ protected function getFieldDefinitions($context = [], $params = []): array
         return $container->getFieldDefinitions($params);
     }
 
-    /**
-     * @param ClassDefinition\Data $fieldDefinition
-     * @param string $name
-     * @param string $language
-     *
-     * @internal
```
aashan10, line 423: I'm not sure if removing `@internal` is a good idea here!

aashan10 (changes requested):

bundles/AdminBundle/src/Controller/Admin/Document/PrintpageControllerBase.php
blankse: @aashan10 array is also included in mixed. See https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.mixed

bundles/EcommerceFrameworkBundle/Type/Decimal.php
blankse: Done: https://github.com/pimcore/pimcore/pull/13194/commits/c0f52cc66e4340bc01d764c42c29536d5840daea

lib/Workflow/MarkingStore/DataObjectSplittedStateMarkingStore.php
blankse: @aashan10 Why? Is only called in line 118 there it should be a string. Also it is in the method used for getFieldDefinition() which only allowed string.

models/DataObject/Localizedfield.php
blankse: @aashan10 Private methods are always internal, right?

lib/Workflow/MarkingStore/DataObjectSplittedStateMarkingStore.php
aashan10: Oh I see! Thanks.

models/DataObject/Localizedfield.php
aashan10: Yeah.. you're right! Thanks

aashan10 (approved):

doc/Development_Documentation/26_Best_Practice/83_Showing_Custom_Layouts_based_on_Object_Data.md
```diff
@@ -44,63 +44,55 @@ namespace App\EventListener;
...
-        
-        if($customLayoutToSelect != null) {
-            //set current layout to subcategory layout
-            $data['currentLayoutId'] = $customLayoutToSelect;
-            $customLayout = CustomLayout::getById($customLayoutToSelect);
-            $data['layout'] = $customLayout->getLayoutDefinitions();
-            Service::enrichLayoutDefinition($data["layout"], $object);            
-        }
+
+    private function doModifyCustomLayouts(array $data, Product $object, int $customLayoutToSelect, array $layoutsToRemove): void
```
dvesh3:
```suggestion
    private function doModifyCustomLayouts(array $data, Product $object, int $customLayoutToSelect, array $layoutsToRemove): array
```

dvesh3 (approved):

dvesh3:
@blankse thank you very much 💜

PR #13197, opened 2022-09-19
[Improvement]: Safety measure preventing user from deleting parent instead of elements selected in grid

### LINKED ISSUE #12916: [Improvement]: Safety measure preventing user from deleting parent instead of elements selected in grid
### Improvement description

Ever since there have been several reported cases from different customers/users where the user selected elements in the grid view and intended to delete them (right-click context menu on grid & "Delete"). However, the user accidentally uses the delete/recycle bin icon in the toolbar and also "blindly" confirms the warning dialog for the deletion (expecting a regular "Do you really want to?", I guess) and thus deleted the parent instead of the selected elements in the grid.

Depending on the amount of children objects the non-cancellable delete operation skips the entries in the recycle bin (yes, this is also mentioned in the confirm dialog), which leads to unrecoverable data loss.

parent-delete

If you agree with me this being potentially error prone, there could be multiple ways to make the error more obvious:

1. When elements are selected in the grid, show another/different warning modal like "You have selected elements in the grid but this operation will delete the entire parent/folder. Are you sure?".
2. When elements are selected in the grid, disable the parent's delete button (and optionally add a tooltip to it).
3. ?

robertSt7:
resolve #12916


bundles/CoreBundle/Resources/translations/en.json
```diff
@@ -985,5 +985,6 @@
     "asset_upload_overwrite_all": "Overwrite all",
     "could_not_open_zip_file": "The uploaded zip file could not be opened.",
     "objects_sort_method": "Change data objects sort method",
-    "link_recursion_error": "The link is redirecting to itself, causing an infinite loop"
+    "link_recursion_error": "The link is redirecting to itself, causing an infinite loop",
+    "delete_entire_folder_question": "This operation will delete the entire parent/folder. Are you sure that you want to delete the folder?"
```
mcop1:
```suggestion
    "delete_entire_folder_question": "This operation will delete the entire parent/folder. You may not be able to fully undo this operation."
```

PR #13201, opened 2022-09-20
Composer: Make some static methods useable for Package Events too

storbahn:
The static methods hasDirectory, getConsoleDir, getOptions and executeCommand could now be used for PackageEvents too.

  
  

Resolves #

### DIFF
--- a/lib/Composer.php
+++ b/lib/Composer.php
@@ -34,7 +34,7 @@ class Composer
     ];
 
     /**
-     * @param Event $event
+     * @param Event|PackageEvent $event
      *
      * @return string
      */
@@ -161,7 +161,7 @@ class Composer
      *
      * @internal
      */
-    protected static function executeCommand(Event $event, $consoleDir, array $cmd, $timeout = 900, $writeBuffer = true)
+    protected static function executeCommand(Event|PackageEvent $event, $consoleDir, array $cmd, $timeout = 900, $writeBuffer = true)
     {
         $command = [static::getPhp(false)];
         $command = array_merge($command, static::getPhpArguments());
@@ -230,7 +230,7 @@ class Composer
         return $arguments;
     }
 
-    protected static function getOptions(Event $event)
+    protected static function getOptions(Event|PackageEvent $event)
     {
         $options = array_merge(static::$options, $event->getComposer()->getPackage()->getExtra());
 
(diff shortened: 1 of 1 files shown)


brusch:
@storbahn Can you please explain why this is necessary? Thanks for clarifying 😊

brusch:
@storbahn 🏓 ping 😊 Thanks in advance!

storbahn:
Hi @brusch,

the idea was to use the console commands like "bin/console data:import:definition-files" for composer package events (e.g. post-package-install or post-package-update) when we install or update composer packages.

At the moment it is not possible, because only Event as first parameter type is allowed.

jdreesen:
Have you tried it this way (not sure if it works, though, as I haven't tried it :wink:)?

```json
"post-package-install": [
    "@php bin/console data:import:definition-files"
],
```

storbahn:
Hi @jdreesen,

it would work, but in our project we wanted to execute the command only for one specific package.

brusch:
As this class is tagged as `@internal` anyway, imho this doesn't make sense as this functionality is not used by Pimcore itself. 
But please feel free to continue the conversation 😊 Thanks anyway for your input and thoughts.

jdreesen:
@storbahn I don't know what you mean by:

> execute the command only for one specific package

storbahn:
Hi @jdreesen,

I hope this example, makes it a little bit more understandable.

composer.json:
```
"scripts": {
        "post-package-install": [
            "App\\Composer\\DataObjectImporter::import"
        ],
        "post-package-update": [
            "App\\Composer\\DataObjectImporter::import"
        ],
        ...
    },
```
App\\Composer\\DataObjectImporter:
```
<?php

declare(strict_types=1);

namespace App\Composer;

use Composer\DependencyResolver\Operation\InstallOperation;
use Composer\DependencyResolver\Operation\UpdateOperation;
use Composer\Installer\PackageEvent;
use Pimcore\Composer;

class DataObjectImporter extends Composer
{
    public static function import(PackageEvent $event): void
    {
        $packageName = '';
        $operation = $event->getOperation();
        if ($operation instanceof UpdateOperation) {
            $packageName = $operation->getTargetPackage()->getName();
        } elseif ($operation instanceof InstallOperation) {
            $packageName = $operation->getPackage()->getName();
        }

        if ($packageName === 'vendor/data-model') {
            self::importDefinitionFiles($event);
        }
    }

    private static function importDefinitionFiles(PackageEvent $event): void
    {
        if (!$consoleDir = static::getConsoleDir($event, 'data import definition files')) {
            return;
        }

        $event->getIO()->write('  > bin/console data:import:definition-files --verbose', true);

        $command = [
            'data:import:definition-files',
            '--verbose',
        ];
        static::executeCommand($event, $consoleDir, $command);
    }
}
```

jdreesen:
@brusch it was tagged as `{@internal}`, which PhpStorm doesn't recognize. I fixed that in #13409

Issue #3396, opened 2018-09-29
Deleting Notes


solverat:
Currently it's not possible to delete notes. Are there any reasons for that? I could imagine that it's connected with important workflow / system / commerce notes which are not allowed to delete. But maybe we could allow the deletion of notes made by non-system users?

dpfaffenbauer:
Maybe with another parameter: `locked` or something like that

KatrinAckermannEk:
I would like to see this enabled/fixed, too. As well as the possibility of editing a note after it has been made by a non-system user.

brusch:
- only by admin users 
- possibility to `lock` a note -> this should be the default (BC) 
- notes which are created manually using the admin interface are set to `lock = false` so that they can be deleted

alexthunder3:
Is there at least some sort of cleaning job that is part of maintenance which I've missed? I haven't found a section for it in the system settings (as there is for objects/ documents versions), the php API doesn't give me any methods for deleting, the documentation doesn't even mention it. So the only way to keep these in check for now is to make your own query command?

dvesh3:
Fixed by #13206

PR #13206, opened 2022-09-20
Notes: Missing delete option

blankse:
In the admin interface the notes can't be deleted but in the PHP API is a delete method. I think it should be possible.


brusch:
Thanks @blankse !
See discussion here: 
https://github.com/pimcore/pimcore/issues/3396

Actually it would be something for `11.x` (not a bug) - but we should discuss the details how it should work first 😊

blankse:
@brusch Thanks for the reply. I'll take a look and report back.

aashan10:
@blankse any updates on this one?

blankse:
@brusch @aashan10 I'm back from vacation. I added the locked attribute. Think this would be a good solution. WDYT?

brusch:
@blankse Thanks, LGTM! 😊

aashan10:
@blankse The changes in this pr work fine, however I have a question. If a note is locked by default, shouldn't there be some option to "unlock" a note?  Or else, there will be no delete option and I don't think this feature will be known to anyone at all.. 

https://github.com/pimcore/pimcore/issues/3396#issuecomment-461086161
The third point in this list is missing.

blankse:
@aashan10 No the third point is not missing. See here:
https://github.com/pimcore/pimcore/blob/1d736e3acdf8c3c9f9c1356bc8485589c891e638/bundles/AdminBundle/src/Controller/Admin/ElementController.php#L324

I think it sould in the first step no possibility in the admin interface exist to unlock it, right @brusch?

aashan10:
In that case, everything else looks good to me! 
Thanks @blankse

aashan10 (approved):

dvesh3 (approved):

dvesh3:
@blankse thanks a lot!

PR #13213, opened 2022-09-21
[Assets] Support uploading files whose MIME-Type cannot be determined

BlackbitDevs:
Probably due to change of asset storage to Flysystem it is currently not possible anymore to upload asset files whose MIME-Type cannot be determined. For us this happened with an stp file. You will get
Bildschirmfoto 2022-09-21 um 09 48 42

To reproduce you can try to upload a file from https://www.eagleburgmann.com/en/company/download-center/stp-files - obviously this file extension is missing in https://github.com/thephpleague/mime-type-detection/blob/main/src/GeneratedExtensionToMimeTypeMap.php - have tackled this in https://github.com/thephpleague/mime-type-detection/pull/18 - but there will be other file extensions which are also not listed here. Those assets should be possbile to get uploaded as type `unknown` to Pimcore. With this PR this works.

### DIFF
--- a/models/Asset.php
+++ b/models/Asset.php
@@ -17,6 +17,7 @@ namespace Pimcore\Model;
 
 use Doctrine\DBAL\Exception\DeadlockException;
 use Exception;
+use League\Flysystem\UnableToRetrieveMetadata;
 use function is_array;
 use League\Flysystem\FilesystemException;
 use League\Flysystem\FilesystemOperator;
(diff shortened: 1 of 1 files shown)


jdreesen:
This sounds more like a bugfix to me. Shouldn't this target `10.5` instead?

BlackbitDevs:
Completely right @jdreesen - have changed target branch. Thx.

models/Asset.php
```diff
@@ -727,7 +728,11 @@ protected function update($params = [])
 
                 $this->closeStream(); // set stream to null, so that the source stream isn't used anymore after saving
 
-                $mimeType = $storage->mimeType($path);
+                try {
+                    $mimeType = $storage->mimeType($path);
+                } catch(UnableToRetrieveMetadata $e) {
+                    $mimeType = 'unknown';
```
```suggestion
                    $mimeType = 'application/octet-stream';
```
We should use `application/octet-stream` here instead, because `unknown` is not a valid mime-type, right?

brusch (changes requested):

PR #13214, opened 2022-09-21
Return $this should be type hint static instead of self

blankse:


bundles/AdminBundle/src/Security/ContentSecurityPolicyHandler.php
```diff
@@ -105,12 +105,9 @@ private function getAllowedUrls(string $key, bool $flatten = true): array|string
     }
 
     /**
-     * @param string $key
-     * @param array $value
-     *
      * @return $this
      */
```
mcop1, line 109: I think we can remove the whole php doc annotation for all occurrences where just `@return $this` remains, because it doesn´t really add any valuable information. What do you think?

blankse: There is a differenz between static and $this (Static can be another instance of the same class).
There is also a RFC to add $this to return types. See: https://wiki.php.net/rfc/this_return_type
Symfony also add this phpdoc to specify return $this instead only static.

jdreesen: I think it does. The return type says `static` which may actually be another object of the same type (perhaps cloned, for immutable objects), while the `return $this` annotation says it returns the *same* object. 

The use of the two types is actually different. While you can chain both, you should assign `static` returns to a variable, which is not necessary for `$this`.

As an example for `$this`, both are fine:
```php
$object->a()->b()->c();

// or

$object->a();
$object->b();
$object->c();
```

For `static`, on the other hand, you should write:
```php
$object = $object->a()->b()->c();

// or

$object = $object->a();
$object = $object->b();
$object = $object->c();
```
because you can't be sure that it returns the same object.

jdreesen: For this reason, there were once considerations to introduce this return type in PHP: https://wiki.php.net/rfc/this_return_type

mcop1: Then I think we are good, thanks to you both.

PR #13222, opened 2022-09-22
Changed ElementInterface::save() and implementations of the interface

### LINKED ISSUE #13207: [Improvement]: Implement `$args` parameter to `Concrete` & `AbstractObject::save()`
### Improvement description

The `save()` method on these classes actually accept arguments, but since it is by means of `func_get_args()` rather than method parameters it is not clear from the code overview given by docblock or editor inspections, e.g.

image

Nobody knows, unless reading every single line of code, that it is possible.

The code today:
https://github.com/pimcore/pimcore/blob/8746e8988922ad36560a202f83a99560c82e8275/models/DataObject/Concrete.php#L696-L715

But if re-written like this:
```php
public function save(...$params) 
```

Or like this, which is more in line with what #13177 is doing:
```php
public function save($params = []) 
```

I believe it would be much clearer and the need for calling `func_get_args()` would not be needed.

NiklasBr:
Resolves #13207


models/Element/ElementInterface.php
```diff
@@ -243,9 +243,13 @@ public function setVersionCount(?int $versionCount): self;
     public function getVersionCount(): int;
 
     /**
-     * @return $this
+     * @param array $parameters{
+     *      versionNote: string, # Used to set a note on the version
+     * } Associative array with optional parameters that are also available to Events
+     *
+     * @return static
```
blankse:
```suggestion
     * @return $this
```

models/Element/ElementInterface.php
```diff
@@ -243,9 +243,13 @@ public function setVersionCount(?int $versionCount): self;
     public function getVersionCount(): int;
 
     /**
-     * @return $this
+     * @param array $parameters{
+     *      versionNote: string, # Used to set a note on the version
+     * } Associative array with optional parameters that are also available to Events
```
blankse:
```suggestion
     * @param array{
     *      versionNote?: string, # Used to set a note on the version
     * } $parameters Associative array with optional parameters that are also available to Events
```

See https://phpstan.org/writing-php-code/phpdoc-types#array-shapes

NiklasBr: Sure about this? PHP Storm complains with the error "Argument type does not match the declared. Doc tag without variable name doesn't provide type information for any expression" if I commit this change

blankse: @NiklasBr Maybe the comment is not allowed in the type? Can we put it also to the description?

NiklasBr: That might be it.

blankse:
```suggestion
     * @param array{versionNote?: string} $parameters Associative array with optional parameters that are also available to Events
     *  - versionNote: Used to set a note on the version
```

aryaantony92:
Failing tests are not related to this PR.

aryaantony92:
@NiklasBr Thanks a lot!

SamyMP:
Do you think it would be possible to pass parameters from a javascript event listener in the admin UI interface ? 

My use case is : I want to be able to trigger a PHP EventListener only if necessary, for example if a field of my dataobject has changed. Currently, all the eventListeners are executed and there's no way to know if the field in question has changed or not from the PHP code. Being able to pass parameters to "save" function from the javascript event listener `pimcore.events.preSaveObject` would be great, we could for example send the field that have changed as an argument

What do you think ?

NiklasBr:
Sounds like something [DirtyIndicatorInterface ](https://github.com/pimcore/pimcore/blob/11.x/models/Element/DirtyIndicatorInterface.php)maybe can help with?

SamyMP:
I have tried to implement my EventListener with [DirtyIndicatorInterface](https://github.com/pimcore/pimcore/blob/11.x/models/Element/DirtyIndicatorInterface.php) and use trait [DirtyIndicatorTrait](https://github.com/pimcore/pimcore/blob/11.x/models/Element/Traits/DirtyIndicatorTrait.php) but the only field marked as dirty is "userModification". If I change something on my dataObject, the field doesn't appear in the dirtyFields array. 

Is there something I'm missing ? I'm still on 10.5.15 by the way

NiklasBr:
I was afraid so, there is no documentation on the intended or explicit usage and the only thing we have to go by is the naming convention. No idea if the current behaviour is just missing implementation of the interface or if I am misunderstanding the intention of the code.

SamyMP:
Ok, so I debugged the code a little bit and the "problem" is actually simple (or so it seems) :

For some DataTypes (`ManyToOneRelation`, `AdvancedManyToManyRelation`, `ManyToManyObjectRelation`... and surely others), the setters generated in the `var/classes/DataObject/MyObject.php` have this line : 
```php
if (!$isEqual) {
    $this->markFieldDirty("myFieldName", true);
}
```

From what I've seen so far, it only concerns Relation types, I gave a few examples above. For testing purposes, I tried to add `$this->markFieldDirty("myInputField", true);` to another field type (like Input), and it worked. I was able to retrieve the information that myInputField had changed.
```php
$fieldHasChanged = $object->isFieldDirty('myInputField'); // returns true if myInputField has changed
```

The only remaining problem is that as long as the object is not reloaded from the admin interface, the field will be marked as dirty. For example, modifying `myInputField`, then saving marks this field as dirty in the PHP code. Saving a second time in a row _without changing anything in the object_ also sets `myInputField` as dirty. I feel like at this point, it could be a JavaScript issue, but not sure. I'll keep investigating

Would there be a problem to implement `$this->markFieldDirty("myFieldName", true)` on every dataTypes setters and not only on Relations ?

SamyMP:
Would it be ok to modify setters from this :
```php
public function setMyField($myField)
{
    $this->myField = $myField;

    return $this;
}
```

to this :
```php
public function setMyField($myField)
{
    if ($this->getMyField() !== $myField) {
        $this->markFieldDirty('myField', true);
        
        $this->myField = $myField;
    }
    return $this;
}
```

What do you think ?

NiklasBr:
I don't know. Seems like a bit annoying to do it everywhere? But I can't think of anything else right now.

SamyMP:
We would only have to add this 
```php
if ($this instanceof DataObject\ClassDefinition\Data\PreSetDataInterface) {
            $code .= "\t" . '$this->' . $key . ' = ' . '$fd->preSetData($this, $' . $key . ');' . "\n";
        } else {
            $code .= "\t" . 'if($this->get' . ucfirst($this->getName()) . '() !== $' . $key . ') {' . "\n";
            $code .= "\t\t" . '$this->markFieldDirty("' . $key . '", true);' . "\n\n";
            $code .= "\t\t" . '$this->' . $key . ' = ' . '$' . $key . ";\n";
            $code .= "\t" . '}' . "\n";
        }
```

here : 
https://github.com/pimcore/pimcore/blob/c4d6cf43b0653588a34e8a29a92363887109be41/models/DataObject/ClassDefinition/Data.php#L552-L556

I don't know if there is a better way to do it...

NiklasBr:
I like it. Maybe you can create a new issue (or PR)?

SamyMP:
I will provide a PR, I'll try to see what I can do about the reload problem as well

Issue #11662, opened 2022-03-14
Update scheb/2fa-bundle to version 6


dvesh3:
### Improvement description

Version 6 supports Symfony 5.4 & 6 so can be already on Pimcore X.

- [ ] Support both versions "^5.3 || ^6.0", as version 6 requires [authenticator based security](https://github.com/pimcore/pimcore/pull/10967), which is enabled by the user. By default, old authentication system is used.
- [ ] also consider `scheb/2fa-google-authenticator`

dvesh3:
Seems like we have to wait until authenticator based security is used by default, as version 6 of `scheb/2fa` bundle will not work with old authentication system.

**Notes from the 2fa-bundle:** "_Symfony 5.4 will be supported as a migration path, but only authenticator security. If you try using Symfony 5.4 and the old security system, there will be an error_" 

Throws exception with old authentication security:
![image](https://user-images.githubusercontent.com/5137917/168049212-ffe3f114-a35c-4954-afea-2cfb765d30c5.png)

PR #13225, opened 2022-09-22
Add support for 2FA with new authenticator

### LINKED ISSUE #13220: [Bug] 2FA not trigger with new authenticator
After installing the latest Demo Package Version (10.5) on PHP 8.1, i tried to setup 2FA but it simply does not work. 

- Enabled the 2FA checkbox in the user profile 
- After reloading the backend no change in login form - 2FA does not trigger
- Fresh browser instance with cleared caches does not work either

I cleared all caches and temporary files, doesn't help. 

Setup: 
- Apache / PHP 8.1 FPM
- Redis Cache
- Maria DB 10.5

Thanks in advance!
Elmar

_Originally posted by @elmarputz in https://github.com/pimcore/pimcore/discussions/13219_

dvesh3:
Resolves #13220


composer.json
```diff
@@ -150,8 +150,8 @@
     "umpirsky/country-list": "^2.0.6",
     "vrana/adminer": "^4.8.0",
     "vrana/jush": "~2.0.2",
-    "scheb/2fa-bundle": "^5.3",
-    "scheb/2fa-google-authenticator": "^5.3",
+    "scheb/2fa-bundle": "^5.3 || ^6.0",
+    "scheb/2fa-google-authenticator": "^5.3 || ^6.0",
```
kingjia90: i was testing a little bit, it seems that we don't need to extend v6 at all.
The v5.x supports both legacy and new authenticator since https://github.com/scheb/2fa/releases/tag/v5.10.0, while [v6](
https://github.com/scheb/2fa/releases/tag/v6.0.0) is working ONLY with the new authenticator.

dvesh3: Ohh, you are right. we can just implement the FC layer for new authenticator and then upgrade it to version 6 on Pimcore 11 as described in this issue https://github.com/pimcore/pimcore/issues/11662

PR #13227, opened 2022-09-22
[Object Relations] Tag mode

BlackbitDevs:
In edit mode relations get currently displayed in a grid. Often this is useful but sometimes it wastes a lot of space and is complicated to handle as the assignment via drag & drop / search takes a lot of time. This PR adds a new configuration `Display mode` in the field definition of Many-to-many Object Relation and Advanced Many-to-many Object Relation fields. In `Tag mode` you can just type to get an auto-suggest list of matching elements, select one, search for the next and so on. This is much faster in daily data maintenance than looking for the object in the tree or using the search window.

In class definition you can select display mode option
MicrosoftTeams-image (23)

The configured visible fields will be shown as tag item label
Снимок экрана 2022-09-22 в 16 27 12

Снимок экрана 2022-09-22 в 16 37 29

This screenshot shows the auto-suggest list when the user typed something
Снимок экрана 2022-09-22 в 16 27 27

This PR got its inspiration from https://github.com/nambu-ch/pimcore-relation-select-fields - the problem with this bundle is that it creates new field types. Those are for example not supported as target field for the Data Importer bundle. This was the main issue why we decided to create this core PR. Another difference is that the configured `visible fields` get used while in the Nambu package always the `key` gets shown.


models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -42,6 +42,15 @@ abstract class AbstractRelations extends Data implements
...
     public $classes = [];
 
+    /**
+     * Optional display mode
+     *
+     * @internal
+     *
+     * @var null|string
+     */
+    public $displayMode;
```
blankse:
```suggestion
    /**
     * Optional display mode
     *
     * @internal
     */
    public ?string $displayMode = null;
```

models/DataObject/ClassDefinition/Data/Relations/AbstractRelations.php
```diff
@@ -73,6 +82,23 @@ public function setClasses($classes)
...
+
+    /**
+     * @param string|null $sqlCondition
+     * @return AbstractRelations
+     */
+    public function setDisplayMode(?string $displayMode): AbstractRelations {
+        $this->displayMode = $displayMode;
+
+        return $this;
+    }
```
blankse:
```suggestion
    public function getDisplayMode(): ?string
    {
        return $this->displayMode;
    }

    /**
     * @return $this
     */
    public function setDisplayMode(?string $displayMode): static
    {
        $this->displayMode = $displayMode;

        return $this;
    }
```

bundles/AdminBundle/src/Controller/Admin/DataObject/DataObjectController.php
```diff
@@ -547,6 +548,82 @@ static function (Task $task) {
         throw $this->createAccessDeniedHttpException();
     }
 
+    /**
+     * @Route("/relation-objects-list", name="relation_objects_list", methods={"GET"})
+     * @param Request $request
+     * @return JsonResponse
+     */
+    public function optionsAction(Request $request) {
```
blankse:
```suggestion
     */
    public function optionsAction(Request $request): JsonResponse
    {
```

blankse (changes requested):
Set type hints and no redundant php docs

BlackbitDevs:
Thx @blankse, have applied the changes.

bundles/AdminBundle/public/js/pimcore/object/classes/data/advancedManyToManyObjectRelation.js
```diff
@@ -123,6 +123,22 @@ pimcore.object.classes.data.advancedManyToManyObjectRelation = Class.create(pimc
 
         this.specificPanel.add(this.classCombo);
 
+        this.specificPanel.add(
+            {
+                xtype: "combo",
+                fieldLabel: t("display_mode"),
+                name: "displayMode",
+                value: this.datax.displayMode,
```
robertSt7:
```suggestion
                value: this.datax.displayMode ?? 'grid',
```

bundles/AdminBundle/public/js/pimcore/object/classes/data/manyToManyObjectRelation.js
```diff
@@ -150,6 +150,22 @@ pimcore.object.classes.data.manyToManyObjectRelation = Class.create(pimcore.obje
             }
         }));
 
+        this.specificPanel.add(
+            {
+                xtype: "combo",
+                fieldLabel: t("display_mode"),
+                name: "displayMode",
+                value: this.datax.displayMode,
```
robertSt7:
```suggestion
                value: this.datax.displayMode ?? 'grid',
```

robertSt7 (approved):
@BlackbitDevs  I think we should also display the default value in the ui. WDYT?

robertSt7:
When I open an object with a relation data component, I get the following error:

```
URL: /admin/object/relation-objects-list?_dc=1666262555655
Method: GET
Message: explode(): Argument #2 ($string) must be of type string, array given
Trace: 
in /var/www/html/dev/pimcore/pimcore/bundles/AdminBundle/src/Controller/Admin/DataObject/DataObjectController.php:566
#0 /var/www/html/dev/pimcore/pimcore/bundles/AdminBundle/src/Controller/Admin/DataObject/DataObjectController.php(566): explode(',', Array)
#1 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(153): Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectController->optionsAction(Object(Symfony\Component\HttpFoundation\Request))
#2 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(75): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#3 /var/www/html/vendor/symfony/http-kernel/Kernel.php(202): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#4 /var/www/html/public/index.php(36): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#5 {main}
```

Could you please take a look at this?

And could you please also avoid using var and use let and const instead? Thank you

robertSt7:
@BlackbitDevs Have you had time to look at my feedback?

BlackbitDevs:
@robertSt7 Have fixed it. This problem was caused when no visible fields have been selected. In this case in https://github.com/pimcore/pimcore/blob/2f46cce6e9f048ac6f68bcf7ca72f8d2829879e3/models/DataObject/ClassDefinition/Data/ManyToManyObjectRelation.php#L815-L823 do not get converted to a string.

robertSt7:
> @robertSt7 Have fixed it. This problem was caused when no visible fields have been selected. In this case in
> 
> https://github.com/pimcore/pimcore/blob/2f46cce6e9f048ac6f68bcf7ca72f8d2829879e3/models/DataObject/ClassDefinition/Data/ManyToManyObjectRelation.php#L815-L823
> 
> do not get converted to a string.

@BlackbitDevs Thanks for that. I get the following error when I select the display mode "Tag field"

```
URL: /admin/object/relation-objects-list?_dc=1667983544404
Method: GET
Message: Warning: Undefined array key "sqlCondition"
Trace: 
in /var/www/html/dev/pimcore/pimcore/bundles/AdminBundle/src/Controller/Admin/DataObject/DataObjectController.php:574
#0 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(163): Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectController->optionsAction(Object(Symfony\Component\HttpFoundation\Request))
#1 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(75): Symfony\Component\HttpKernel\HttpKernel->handleRaw(Object(Symfony\Component\HttpFoundation\Request), 1)
#2 /var/www/html/vendor/symfony/http-kernel/Kernel.php(202): Symfony\Component\HttpKernel\HttpKernel->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#3 /var/www/html/public/index.php(36): Symfony\Component\HttpKernel\Kernel->handle(Object(Symfony\Component\HttpFoundation\Request))
#4 {main}
```

Could you please also take a look at this? Thanks

BlackbitDevs:
@robertSt7 Fixed. This was because at this time we developed this PR and #13131 in parallel but as #13131 was declined, we do not have the `sqlCondition`.

robertSt7:
@BlackbitDevs Thank you. If I select no visible fields, then I see blank rows and empty elements in the multi-select. I think we should use the default fields id, fullpath and classname like in the grid view for this case. What do you think?

robertSt7:
@BlackbitDevs I am also just thinking about a clear-relation-button in the "Tag field" view like in the grid view. What's your opinion about that?

BlackbitDevs:
@robertSt7 I have added the fallback when no visible fields have been set - as suggested id, fullpath and classname get displayed in this case.

The clear button imho is not necessary because clearing is very simple by just using backspace key. But if you think this is necessary, I can add it.
But for example tag mode for multiselect fields also does not have a clear button.

robertSt7:
@BlackbitDevs You are right. We don't need a clear button. Everthing looks good now, except one thing. Could you please use 'let' and 'const' in the js files instead of 'var'? Thanks

BlackbitDevs:
@robertSt7 Now `let` gets used in the JS files.

I saw that most of the changed lines here at Github came from formatting - in PhpStorm you can ignore formatting changes, there it looks much cleaner. Of course the Github comparison distracts from the real changes. Is there any chance that you provide PhpStorm formatting rules or a bot which automatically formats the code?

robertSt7:
@BlackbitDevs Not at the moment, but we don't want to use PhPStorm formatting rules because some developers might use other IDEs. However we would like to format our JS code like for PhP with a PHP-CS-Fixer. Suggestions are very much appreciated :)

jdreesen:
For JavaScript, that would be https://prettier.io/

brusch:
Would be time then for https://github.com/marketplace/actions/prettier-action 😊

jdreesen:
It can format your Markdown docs, too ;)

robertSt7:

PR #13229, opened 2022-09-22
[Bug] listing getTotalCount() column already exists

kjkooistra-youwe:
Adding multiple joins to a listing via onCreateQueryBuilder causes PDOException: SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name 'cid'").

Caused by the subquery retrieving `*`, while join queries may contain the same column. Fixed by using the identifier column introduced in https://github.com/pimcore/pimcore/pull/12854


brusch:
Imho this should be safe and shouldn't have any side-effects. Thanks a lot 👍

PR #13232, opened 2022-09-22
Added event listener to update search indices

### LINKED ISSUE #13208: [Bug]: search_backend_data does not get updated when the folder changes
### Expected behavior

Backend Search Index (table: search_backend_data) should also be updated if the folder changes (rename)

### Actual behavior

Backend Search Index (table: search_backend_data) is not updated if the folder changes (rename)

### Steps to reproduce

- create a folder in assets
- upload a file to this folder
- open database and search for the ID of this document in the table "search_backend_data"
  - check "fullpath" and "data"
- rename the folder
- refresh database view
  - check "fullpath" and "data" again (its still the old value)

aashan10:
Resolves #13208


brusch:
That's probably not only the case for assets, but also data objects and documents, right? 
And there's already an existing listener for that, that uses the message queue for processing the stuff: 
https://github.com/pimcore/pimcore/blob/1aed6b7e85ebcb42d494cc892e0e42920c4dc809/bundles/CoreBundle/EventListener/SearchBackendListener.php#L75 

So I think what we need to do here is to check whether the full path has changed or not and add this as a flag to the message, so that we can then decide in the handler whether we process the children as well or not. 
You can use the `getCurrentFullPath()` method for that, which should be available in the DAOs of all 3 element types. 

Keep in mind that it could be the case that elements can have millions of children, the current implementation would timeout here 😊

PR #13237, opened 2022-09-23
Fix deeplink with custom entrypoint

matthiashamacher:
Currently, it is not possible to share a deep link with a custom admin entry point as the route `/login/deeplink` does not work if there is a custom admin entry point available. 

To fix this I implemented a new route that generates the deep link URL. This is necessary as the custom admin entry point is not exposed to the frontend. The new route tries to generate the deep link URL with the custom admin entry point; if that is not possible, the standard deep link URL is returned. If there is a custom admin entry point defined the URL will look like this `/mycustomlogin?deeplink=object_123_object` That way we have the necessary part for the deep link in the query params. In the custom admin entry point route, I've added a new case where, if the deep link query param is present it redirects to the real deep link URL after setting the cookie. That way the standard deep link logic applies.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/folder.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/folder.js
@@ -291,23 +291,19 @@ pimcore.asset.folder = Class.create(pimcore.asset.asset, {
                 });
             }
 
-            let user = pimcore.globalmanager.get("user");
+            this.setDeeplink("asset_" + this.data.id + "_" + this.data.type);
 
-            if (user.admin) {
-                this.setDeeplink("asset_" + this.data.id + "_" + this.data.type);
+            this.toolbarButtons.metainfo = new Ext.SplitButton(
+                {
+                    tooltip: t("show_metainfo"),
+                    iconCls: "pimcore_material_icon_info pimcore_material_icon",
+                    scale: "medium",
+                    handler: this.showMetaInfo.bind(this),
+                    menu: this.getMetaInfoMenuItems()
+                }
+            );
 
-                this.toolbarButtons.metainfo = new Ext.SplitButton(
-                    {
-                        tooltip: t("show_metainfo"),
-                        iconCls: "pimcore_material_icon_info pimcore_material_icon",
-                        scale: "medium",
-                        handler: this.showMetaInfo.bind(this),
-                        menu: this.getMetaInfoMenuItems()
-                    }
-                );
-
-                buttons.push(this.toolbarButtons.metainfo);
-            }
+            buttons.push(this.toolbarButtons.metainfo);
 
             buttons.push("-");
             buttons.push({
--- a/bundles/AdminBundle/Resources/public/js/pimcore/asset/folder.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/asset/folder.js
@@ -291,16 +291,22 @@ pimcore.asset.folder = Class.create(pimcore.asset.asset, {
                 });
             }
 
-            var user = pimcore.globalmanager.get("user");
+            let user = pimcore.globalmanager.get("user");
+
             if (user.admin) {
-                buttons.push({
-                    xtype: "splitbutton",
-                    tooltip: t("show_metainfo"),
-                    iconCls: "pimcore_material_icon_info pimcore_material_icon",
-                    scale: "medium",
-                    handler: this.showMetaInfo.bind(this),
-                    menu: this.getMetaInfoMenuItems()
-                });
+                this.setDeeplink();
+
+                this.toolbarButtons.metainfo = new Ext.SplitButton(
+                    {
+                        tooltip: t("show_metainfo"),
+                        iconCls: "pimcore_material_icon_info pimcore_material_icon",
+                        scale: "medium",
+                        handler: this.showMetaInfo.bind(this),
+                        menu: this.getMetaInfoMenuItems()
+                    }
+                );
+
+                buttons.push(this.toolbarButtons.metainfo);
             }
 
             buttons.push("-");
@@ -323,6 +329,37 @@ pimcore.asset.folder = Class.create(pimcore.asset.asset, {
         return this.toolbar;
     },
 
+    setDeeplink: function () {
+        let target   = "asset_" + this.data.id + "_" + this.data.type;
+        let that     = this;
+        let response = Ext.Ajax.request({
+            method: 'POST',
+            url: Routing.generate('pimcore_admin_element_getdeeplink'),
+            params: {
+                target: target
+            },
+            success: function (response) {
+                let data = Ext.decode(response.responseText);
+
+                if (data.success) {
+                    that.deeplink = data.url;
+
+                    that.toolbarButtons.metainfo.menu.add(
+                        {
+                            text: t("metainfo_copy_deeplink"),
+                            iconCls: "pimcore_icon_copy",
+                            handler: pimcore.helpers.copyStringToClipboard.bind(this, that.deeplink)
+                        }
+                    )
+                }
+            }
+        });
+    },
+
+    getDeeplink: function () {
+        return this.deeplink;
+    },
+
     getMetaInfo: function() {
         return {
             id: this.data.id,
@@ -332,7 +369,7 @@ pimcore.asset.folder = Class.create(pimcore.asset.asset, {
             creationdate: this.data.creationDate,
             usermodification: this.data.userModification,
             userowner: this.data.userOwner,
-            deeplink: pimcore.helpers.getDeeplink("asset", this.data.id, this.data.type)
+            deeplink: this.getDeeplink()
         };
     },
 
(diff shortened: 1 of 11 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
```diff
@@ -2737,7 +2737,25 @@ pimcore.helpers.isValidPassword = function (pass) {
 };
 
 pimcore.helpers.getDeeplink = function (type, id, subtype) {
-    return Routing.generate('pimcore_admin_login_deeplink', {}, true) + '?' + type + "_" + id + "_" + subtype;
+    let target = type + "_" + id + "_" + subtype;
+    let url = Routing.generate('pimcore_admin_login_deeplink', {}, true) + '?' + target;
+
+    let response = Ext.Ajax.request({
+        method: 'POST',
+        async: false,
```
brusch: We should avoid `async: false` here, as it is deprecated.

brusch (changes requested):

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
matthiashamacher: Just tried to implement it without `async: false` but I haven't found any solution. Any hint on what I can try?

kingjia90: weird, `async: false`  doesn't look `deprecated` but just discouraged, or am i missing something? 
https://docs.sencha.com/extjs/7.0.0/modern/Ext.Ajax.html#cfg-async
https://docs.sencha.com/extjs/7.6.0/modern/Ext.Ajax.html#cfg-async

brusch: It's not about ExtJS, but synchronous requests in general, see: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#synchronous_request

brusch: @matthiashamacher you just need to implement a `success` callback, just as we do in other places: 
https://github.com/pimcore/pimcore/blob/11.x/bundles/AdminBundle/public/js/pimcore/asset/asset.js#L389

brusch (changes requested):

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
matthiashamacher: I don't think it is possible to make this async. The Problem is that the deeplink is directly set to the button. Therefore we cannot update the value after it is initially set: https://github.com/pimcore/pimcore/blob/11.x/bundles/AdminBundle/public/js/pimcore/element/abstract.js#L296
At least I haven't found a way. Also tried to save the deeplink to the globalmanager and getting it from there but it still uses the initial value (false) where the result of the request is not there yet.

kingjia90: what about getting and storing the route name parameter in `globalmanager` once and change `getDeepLink` as following

```
pimcore.helpers.getDeeplink = function (type, id, subtype) {
    let target = type + "_" + id + "_" + subtype;
    let url;
    try {
        url = Routing.generate('pimcore.globalmanager.get('custom_admin_route_name')', {}, true) + '?deeplink=' + target;
    } catch(e){
        url = Routing.generate('pimcore_admin_login_deeplink', {}, true) + '?' + target;
    }

    return url;
};
```
There is some redundancy in the ajax approach, the `let url = Routing.generate('pimcore_admin_login_deeplink', {}, true) + '?' + target;` and the fallback in `getDeepLink()` will end up returning the same value. 

It affects user experience, as it requires multiple syncronous calls eg.if one has lots of extjs tabs open, and it does at least twice for each element, first when it loads the button with the dropdown to allow copy of id/fullpath/deeplink and again when showing the content after clicking the `Info` button
https://github.com/pimcore/pimcore/blob/59357ac83ca165f41a53536bae14f7c3fc70f9f3/bundles/AdminBundle/Resources/public/js/pimcore/element/abstract.js#L279-L280
https://github.com/pimcore/pimcore/blob/59357ac83ca165f41a53536bae14f7c3fc70f9f3/bundles/AdminBundle/Resources/public/js/pimcore/asset/asset.js#L488-L489

matthiashamacher: Storing the route name parameter makes no difference. The problem is, that the custom admin route is not exposed to the frontend. So it is not exported to js. As it makes the whole custom admin route logic useless.

The only other idea I have is to remove the deeplink entry point from the admin routes. So i.e. /deeplink?object_1189_object instead of /admin/login/deeplink?object_1189_object. Then we can redirect from there to either the custom admin route or the normal login. This would add an extra redirect but we could have the complete logic to which endpoint it should be redirect in Symfony and not in ExtJs.

kingjia90: ah, even by adding "expose" as true? like in the [documentation ](https://github.com/FriendsOfSymfony/FOSJsRoutingBundle/blob/master/Resources/doc/usage.rst) eg.
```
myCustomAdminRoute:
    options:
        expose: true
```
or in annotations
```
* @Route("/my-custom-login-page", options={"expose"=true}, name="myCustomAdminRoute") ?
```

matthiashamacher: Yes if you expose the route then it would work but as I have said, that would make the custom admin entry point logic useless. When the route is public it makes no sense to even configure it. At least that is my opinion.

kingjia90: Just pushed an experimental PR https://github.com/pimcore/pimcore/pull/13365 that could help in turning into asyncronous. In short, the button show meta info and the related dropdown are not appearing until the ajax call is succeded.
Hope it helps

matthiashamacher: Thanks for the idea. Changed a little bit so only the deeplink button appears after the ajax call succeded not the whole metainfo button.

bundles/AdminBundle/Resources/public/js/pimcore/asset/folder.js
```diff
@@ -291,16 +291,22 @@ pimcore.asset.folder = Class.create(pimcore.asset.asset, {
                 });
             }
 
-            var user = pimcore.globalmanager.get("user");
+            let user = pimcore.globalmanager.get("user");
+
             if (user.admin) {
```
matthiashamacher: This is the only place where it is checked if the user is admin. Why is that? I don't think that is necessary so I would remove it if it is ok? Or it should be implemented in the other places as well

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
```diff
@@ -2736,10 +2736,6 @@ pimcore.helpers.isValidPassword = function (pass) {
     return true;
 };
 
-pimcore.helpers.getDeeplink = function (type, id, subtype) {
-    return Routing.generate('pimcore_admin_login_deeplink', {}, true) + '?' + type + "_" + id + "_" + subtype;
```
kingjia90, line 2740: I am not sure if we can remove it right away, for BC purposes i would keep it there and just adding a comment/note that this won't work for custom entrypoints and/or add a deprecation trigger.

bundles/AdminBundle/Resources/public/js/pimcore/asset/folder.js
kingjia90: Yes, i've digged a little and it seems a "leftover" and this removal looks fine to me since it will be consistent with the other places. 

By looking at this commit message, it used to be admin only but then got changed to be available to all users.
https://github.com/pimcore/pimcore/commit/36916cdce1bc8f420c4c780ad78f0e2505953aaf

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
kingjia90: Or maybe provide the syncronous call but it would be only for the custom implementation out there.

kingjia90 (changes requested):
Looks good to me and works fine, just one doubt on the removal of the `pimcore.helpers.getDeeplink`

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
matthiashamacher: Added the deprecated flag, console.error and implemented it as synchronous call

kingjia90 (approved):

brusch:
What if we include the custom entry point just here: 
https://github.com/pimcore/pimcore/blob/38419b06c61565e62122e32684e37c102cbe25d2/bundles/AdminBundle/src/Controller/Admin/IndexController.php#L227

We can then simply access them using `pimcore.settings['custom_admin_entrypoint']` ... makes it a lot less complicated and no synchronous call. WDYT?

matthiashamacher:
@brusch 
`pimcore.settings` is only available if you are logged in, correct? Then I think that would be the best solution as the entrypoint is not exposed to the general site and we don't have to make an extra request to get the link.

brusch:
@matthiashamacher exactly 👍😎

matthiashamacher:
@brusch 
Changed it accordingly. I needed to add the whole custom admin url ('http://localhost/my-admin') to the settings as if the route is not exposed we cannot use `Route.genereate()` with the route name, but I still think this is the better solution.

brusch:
@matthiashamacher LGTM - a lot less code, thanks a lot 👍

PR #13238, opened 2022-09-23
Fix styling for inherited fields with label top

matthiashamacher:
With the current styling, the label is moved to the right and the field is not the full width (as seen in screenshot 1) if the label position is set to top. This is because of the `display: table-cell` styling. 

Old:
CleanShot 2022-09-23 at 16 37 29

New:
CleanShot 2022-09-23 at 16 42 56

### DIFF
--- a/bundles/AdminBundle/Resources/public/css/admin.css
+++ b/bundles/AdminBundle/Resources/public/css/admin.css
@@ -1143,6 +1143,10 @@ span.warning {
     margin-right: 10px;
 }
 
+.object_field > .x-form-item-label-top > .x-form-item-label-inner {
+    display: inline-block;
+}
+
 .object_field_panel {
     padding: 0;
 }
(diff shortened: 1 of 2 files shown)


brusch:
I'm afraid that this causes some unwanted side-effects because this style is coming directly from ExtJS. 
Probably we need to make a specific override in `admin.css`. 

But anyway, I'm not able to reproduce the described problem, see: 

image

jdreesen:
@brusch but you have the label on the left, not on the top ;)

brusch:
Aaaah, missed that little detail 😂 Thanks for clarifying. 
But still, I'm not sure if this fix causes some unintended side-effects ... as the style comes from the lib itself.

matthiashamacher:
Yeah, I can understand. I thought this would be ok because it's a static file with no dynamic styling...

How about adding it to the https://github.com/pimcore/pimcore/blob/11.x/bundles/AdminBundle/public/css/ext-modifications.css Seems that some other styles override the PimcoreApp-all_1.css

brusch:
Or just in `admin.css` ... but maybe we should make it specific to the top positioning, rather than changing the generic style, you know what I mean? I'm a bit worried about possible side effects because the selector is quite generic.

robertSt7:
@matthiashamacher Thanks for the PR

PR #13257, opened 2022-09-28
Object Brick layout improvement; Relates #12719

gjorgic:
This was temp improvement for 10.5, because same change was created for 11.x in PR #12723


brusch:
As this is not a bug fix, it won't go into `10.5` - so `11.x` should be fine then - thanks anyway 😊

PR #13259, opened 2022-09-28
Override getFilterConditionExt

sigere:

Values in queries sent to DB are quoted. That makes results invalid. Example:
```
HAVING
	cskey_technicalData_5_10.value > '952'
```
Should be:
```
HAVING
	cskey_technicalData_5_10.value > 952
```

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Numeric.php
+++ b/models/DataObject/ClassDefinition/Data/Numeric.php
@@ -35,7 +35,10 @@ class Numeric extends Data implements ResourcePersistenceAwareInterface, QueryRe
 
     const DECIMAL_PRECISION_DEFAULT = 0;
 
-    const VALID_FILTER_OPERATORS = [
+    /**
+     * @var array
+     */
+    public static $validFilterOperators = [
         '=',
         'IS',
         'IS NOT',
@@ -582,7 +585,7 @@ class Numeric extends Data implements ResourcePersistenceAwareInterface, QueryRe
             }
         }
 
-        if (is_numeric($value) && in_array($operator, self::VALID_FILTER_OPERATORS)) {
+        if (is_numeric($value) && in_array($operator, self::$validFilterOperators)) {
             return $key . ' ' . $operator . ' ' . $value . ' ';
         }
 
--- a/models/DataObject/ClassDefinition/Data/Numeric.php
+++ b/models/DataObject/ClassDefinition/Data/Numeric.php
@@ -36,6 +36,20 @@ class Numeric extends Data implements ResourcePersistenceAwareInterface, QueryRe
     const DECIMAL_PRECISION_DEFAULT = 0;
 
     /**
+     * @var array
+     */
+    public static $validFilterOperators = [
+        '=',
+        'IS',
+        'IS NOT',
+        '!=',
+        '<',
+        '>',
+        '>=',
+        '<=',
+    ];
+
+    /**
      * Static type of this element
      *
      * @internal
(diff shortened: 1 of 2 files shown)


sigere:
I have already signed CLA. @CLAassistant you are being wrong.

brusch:
Please ensure you have the local Git email address added to your GitHub account: 
1. You can check your used email with this link: https://patch-diff.githubusercontent.com/raw/pimcore/pimcore/pull/13259.patch (2nd line) 
2. Add this email here: https://github.com/settings/emails
3. Re-start the signing process

models/DataObject/ClassDefinition/Data/Numeric.php
```diff
@@ -35,6 +35,20 @@ class Numeric extends Data implements ResourcePersistenceAwareInterface, QueryRe
 
     const DECIMAL_PRECISION_DEFAULT = 0;
 
+    /**
+     * @var array
+     */
+    public static $validFilterOperators = [
```
jdreesen, line 41: Can this be a `const`?

models/DataObject/ClassDefinition/Data/Numeric.php
```diff
@@ -545,6 +559,39 @@ public function getForCsvExport($object, $params = [])
...
+
+        if ($value === 'NULL') {
+            if ($operator === '=') {
+                $operator = 'IS';
+            } elseif ($operator === '!=') {
+                $operator = 'IS NOT';
+            }
+        }
+
+        if (!is_array($value) && !is_object($value) && in_array($operator, self::$validFilterOperators)) {
```
jdreesen: Wouldn't this be better?

```suggestion
        if (is_scalar($value) && in_array($operator, self::$validFilterOperators, true)) {
```

Then `$value` may only be `int`, `string`, `float` or `bool`, but **not** `array`, `object`, `resource` or `null`.

models/DataObject/ClassDefinition/Data/Numeric.php
sigere: copied from https://github.com/pimcore/pimcore/blob/10.5/models/DataObject/ClassDefinition/Data.php#L104

models/DataObject/ClassDefinition/Data/Numeric.php
sigere: copied from https://github.com/pimcore/pimcore/blob/10.5/models/DataObject/ClassDefinition/Data.php#L581

sigere:
@jdreesen please confirm you want me to apply your suggestions or leave it similar to ones from `Data.php` file.

jdreesen:
I think that's up to @brusch or @dvesh3 or someone else from the core Pimcore team then.

models/DataObject/ClassDefinition/Data/Numeric.php
sigere: maybe we should only allow `int` and `float` since it is numeric data?

models/DataObject/ClassDefinition/Data/Numeric.php
brusch: I'd prefer `const` here as well, yes .... 
and probably we should change https://github.com/pimcore/pimcore/blob/10.5/models/DataObject/ClassDefinition/Data.php#L104 as well, as we are working on a major version anyway. 

@sigere can you update your PR accordingly, please 😊

models/DataObject/ClassDefinition/Data/Numeric.php
brusch: it probably can be a `string` as well, right? 
What about using `is_numeric()` that should cover all cases?

brusch (changes requested):

mcop1:
@sigere Hello, 
thank you for your contribution, but since this is a bug fix for 10.5 we can´t change `$validFilterOperators` from `static` to `const` - could you please revert this change? Sorry for the confusion and the inconvenience.

mcop1:
Thanks again 👍

PR #13263, opened 2022-09-28
[Task]: Bump league/flysystem-bundle to ^3.0

### LINKED ISSUE #13246: Bump league/flysystem-bundle to ^3.0
### Improvement description

Also consider clean-up task https://github.com/pimcore/pimcore/blob/ece8fb4941784bc6c0aab3f499ced300b14e3dd6/lib/Maintenance/Tasks/LogArchiveTask.php#L138-L139 follow-up of https://github.com/pimcore/pimcore/pull/12704#issuecomment-1190120082

kingjia90:
Resolves #13246
and improves the flysystem conf by using `directory_visibility` instead of unix permissions


kingjia90:
https://github.com/pimcore/pimcore/blob/28718e6101a3d603dec2e318d89f8866c2517364/bundles/CoreBundle/config/pimcore/flysystem.yaml#L7-L8

It seems 69 got replaced by alternative approach in https://github.com/thephpleague/flysystem-bundle/pull/100, but it is released in version 3.

brusch:
@kingjia90 I'd do it right here as it's related anyway, then we have it all together 😊 Thanks!

kingjia90:
There's a slight change in the permission, where it was formerly specified as `0644`, it is now `private` which is translated as `0600` , while for public from `0755` to `0744` but it should be fine.

source: https://flysystem.thephpleague.com/docs/visibility/#unix-style-visibility

dvesh3 (approved):

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -58,6 +58,7 @@ Please make sure to set your preferred storage location ***before*** migration.
 - [PhpArrayTable]: Removed PhpArrayTable class
 - [Navigation Builder] Calling the method `Pimcore\Navigation\Builder::getNavigation()` using extra arguments is
   removed. Instead, please pass the arguments as an associative array (eg.`getNavigation($args)`.For details, please see [#12310](https://github.com/pimcore/pimcore/issues/12310)
+- [Flysystem] Bumped `league/flysystem-bundle` minimum requirement to ^3.0 (which introduces `directoryExists()`,`has()` methods and fixes support for `directory_visibility` configuration option). NB: [Visibility](https://flysystem.thephpleague.com/docs/visibility/) permissions were specified as `0644` for files and `0755` for directories, are now using the more simplified `public/private` options which are translated respectively into `0600` and `0744`.
```
dvesh3: one last thing...if a user requires package for using third party adapters e.g. [aws s3](https://flysystem.thephpleague.com/v1/docs/adapter/aws-s3-v3/) then it should be upgraded to version 3 otherwise it will conflict on composer update. should we also describe it here?

PR #13270, opened 2022-09-29
Update the search indices of element and its children after element has been updated

### LINKED ISSUE #13208: [Bug]: search_backend_data does not get updated when the folder changes

aashan10:
Resolves #13208


bundles/CoreBundle/EventListener/SearchBackendListener.php
```diff
@@ -71,9 +71,19 @@ public function onPostAddUpdateElement(ElementEventInterface $e)
...
         $element = $e->getElement();
+        $this->dispatchSearchBackendMessage($element);
+    }
+
+    protected function dispatchSearchBackendMessage(ElementInterface $element) {
         $this->messengerBusPimcoreCore->dispatch(
             new SearchBackendMessage(Service::getElementType($element), $element->getId())
         );
+
+        if (count($element->getChildren()) > 0) {
```
NiklasBr:
```suggestion
        if ($element->hasChildren()) {
```

aashan10:
Thanks @NiklasBr 👍🏽

bundles/CoreBundle/EventListener/SearchBackendListener.php
```diff
@@ -71,9 +71,19 @@ public function onPostAddUpdateElement(ElementEventInterface $e)
...
+    }
+
+    protected function dispatchSearchBackendMessage(ElementInterface $element) {
         $this->messengerBusPimcoreCore->dispatch(
             new SearchBackendMessage(Service::getElementType($element), $element->getId())
         );
+
+        if ($element->hasChildren()) {
+            foreach ($element->getChildren() as $child) {
+                $this->dispatchSearchBackendMessage($child);
```
brusch: I'm not sure if this is really a good idea, as already shared in this [comment](https://github.com/pimcore/pimcore/pull/13232#issuecomment-1255866004), I think it's better to add another flag to `SearchBackendMessage` whether the children should be processed or not.
Because with the current approach we could possible generate millions of messages in the queue ...

brusch (changes requested):

brusch:
I'm afraid that there's still a misunderstanding here 😊 

What I mean is: 

1. we need to check whether the path has changed or not, because we don't need to update all the children if the path hasn't changed, right? We should use https://github.com/pimcore/pimcore/blob/70b1db360a2a4fa498c69f71a33aa1db5a3a42d1/models/Asset/Dao.php#L292 for that (which is available for all 3 element types) 
2. If the path changes, we need to set a new property here: https://github.com/pimcore/pimcore/blob/96dfb78d70ff6a3868e863246a913fd26d2145bb/lib/Messenger/SearchBackendMessage.php#L23 e.g. called `protected bool $updateChildren`. 
3. Then in the `SearchBackendHandler` we can check for this property, if `true` we update the path of the children (only the path, for performance reasons, ...). 

Thanks!

aashan10:
Thanks @brusch. I have made changes accordingly. Now there is a single message and everything else is handled by the message handler as you suggested!

bundles/CoreBundle/EventListener/SearchBackendListener.php
```diff
@@ -71,8 +72,11 @@ public function onPostAddUpdateElement(ElementEventInterface $e)
         }
 
         $element = $e->getElement();
+        $data = Data::getForElement($element);
+
+        $shouldChildrenBeUpdated = $element->getRealFullPath() == $data->getFullPath();
```
brusch:
```suggestion
        $shouldChildrenBeUpdated = $element->getRealFullPath() != $data->getFullPath();
```

lib/Messenger/Handler/SearchBackendHandler.php
```diff
@@ -35,6 +38,29 @@ public function __invoke(SearchBackendMessage $message, Acknowledger $ack = null
...
+        $searchEntry = Data::getForElement($element);
+        if ($searchEntry->getId() instanceof Data\Id) {
+            $searchEntry->setDataFromElement($element);
+        } else {
+            $searchEntry = new Data($element);
+        }
+        $searchEntry->save();
+
+        if ($updateChildren) {
+            if ($element instanceof Asset || $element instanceof AbstractObject || $element instanceof Document) {
```
brusch: This is redundant since we have already `Element\ElementInterface $element` in the arguments 😉

lib/Messenger/Handler/SearchBackendHandler.php
```diff
@@ -35,6 +38,29 @@ public function __invoke(SearchBackendMessage $message, Acknowledger $ack = null
...
+                foreach ($element->getChildren() as $child) {
+                    $data = Data::getForElement($child);
+
+                    $shouldChildrenBeUpdated = $child->getRealFullPath() == $data->getFullPath();
```
brusch:
```suggestion
```
This seems to be redundant as well.

lib/Messenger/Handler/SearchBackendHandler.php
```diff
@@ -35,6 +38,29 @@ public function __invoke(SearchBackendMessage $message, Acknowledger $ack = null
...
+        if ($searchEntry->getId() instanceof Data\Id) {
+            $searchEntry->setDataFromElement($element);
+        } else {
+            $searchEntry = new Data($element);
+        }
+        $searchEntry->save();
+
+        if ($updateChildren) {
+            if ($element instanceof Asset || $element instanceof AbstractObject || $element instanceof Document) {
+                foreach ($element->getChildren() as $child) {
```
brusch: We can't use `getChildren()` here, because this returns only direct children (1st level). 
I think we should anyway consider to use plain SQL to update all the children for performance reasons, just as we do e.g. here: 
https://github.com/pimcore/pimcore/blob/70b1db360a2a4fa498c69f71a33aa1db5a3a42d1/models/Asset/Dao.php#L213 

This should speed up things a lot and make it more scalable (when it comes to millions of children).

brusch (changes requested):

lib/Messenger/Handler/SearchBackendHandler.php
aashan10: This is because phpstan won't allow `$element->getChildren()` on `ElementInterface` since the methods `getChildren`, `hasChildren` etc are not present there.

bundles/CoreBundle/Migrations/Version20221005122653.php
```diff
@@ -0,0 +1,27 @@
...
+final class Version20221005122653 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add fulltext index to search_backend_data table';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql("CREATE FULLTEXT INDEX search_backend_data_fulltext_index ON search_backend_data(`data`, `properties`, `fullpath`);");
```
brusch: I think we should replace the existing index here, otherwise we're more or less doubling the size of the full-text index. So actually we need to drop the existing index here as well.

brusch (changes requested):

brusch:
Tests are failing as well ...

bundles/CoreBundle/Migrations/Version20221005122653.php
aashan10: @brusch I have added the changes, but the tests are still failing. I saw that it has something to do with the github actions itself. Could you please re-run the tests for this pr?

kingjia90:
I am facing some problem with the migration
image
image

bundles/InstallBundle/Resources/install.sql
```diff
@@ -464,7 +464,7 @@ CREATE TABLE `search_backend_data` (
   `id` int(11) NOT NULL,
   `key` varchar(255) CHARACTER SET utf8 COLLATE utf8_bin default '',
   `index` int(11) unsigned DEFAULT '0',
-  `fullpath` varchar(765) CHARACTER SET utf8 COLLATE utf8_general_ci DEFAULT NULL, /* path in utf8 (3-byte) using the full key length of 3072 bytes */
+  `fullpath` varchar(765) DEFAULT NULL,
```
kingjia90: The full text index creation is failing because the encoding is still `utf8` (while default is `utf8mb4`, therefore the other 2 fields are in this encoding), so in the migration it needs to be converted first for existing instances/projects, while with this change, it may only work for fresh installations.

kingjia90: I have some doubts, if we change the collation here, it should be changed elsewhere (`path`,`o_path`,`cpath`) for consistency eg.
https://github.com/pimcore/pimcore/blob/675ac0a18664e3d121d8ef89127cd582a95c5f5e/bundles/InstallBundle/dump/install.sql#L101

I am also not sure why it was `3-byte` and the reason behind that comment line (other than saving storage space purpose?), but maybe would try to ask if there is any problem drawback or shortcoming when changing it

brusch: `3-byte` is there because of the max key length of 3072 bytes, if we would go for 4-byte, we'd have to reduce  `varchar(765)` accordingly, which we can't (see `assets`, `documents` and `objects` table). 
Alsp we don't allow 4-bytes characters as keys, so that's fine and we must keep it that way.

brusch (changes requested):

aashan10:
@brusch mysql is not allowing to create fulltext index because of the charset and collation difference in between `fullpath`, `data` and `properties`. Can we instead change the collation of `data` and `properties` to match that of `fullpath` in order to create the fulltext search index?

brusch:
@aashan10 hmm, unfortunately we don't have functional indexes in MariaDB, that would have been the easiest solution. 
Changing the collation/charset of `data` and `properties` could have some side-effects, but I'm not really sure if they really matter, e.g. when searching for emojis 😎

kingjia90:
@aashan10 do we really need fullpath in the fulltext index? Was that change only to trigger a re-index? 
if yes, what about using 
https://github.com/pimcore/pimcore/blob/b675696e8c3dff29e51d97e67fc79d3003c9ca3a/lib/Maintenance/Tasks/FullTextIndexOptimizeTask.php#L43
along with `set GLOBAL innodb_optimize_fulltext_only=ON;` [source](https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_optimize_fulltext_only)

if not, doesn't work just fine with `OR LIKE string%`?

brusch:
Afaik the strings for the FTI get tokenized, so `OR LIKE string%` wouldn't give the same results as the current implementation. But I agree, I'm also not sure whether we even need the full path in the index and if that is a use-case in reality that someone is searching for parts of the path ... for me it would be fine to just leave it out.

aashan10:
@brusch @kingjia90 dropped the `fullpath` column from fulltext index and removed the relevant migrations. Also, Changed the bit which would add `Path: fullpath` to the `data` column.

models/Search/Backend/Data.php
```diff
@@ -534,7 +534,7 @@ public function setDataFromElement(Element\ElementInterface $element)
 
         $pathWords = str_replace([ '-', '_', '/', '.', '(', ')'], ' ', $this->getFullPath());
         $this->data .= ' ' . $pathWords;
-        $this->data = 'ID: ' . $element->getId() . "  \nPath: " . $this->getFullPath() . "  \n"  . $this->cleanupData($this->data);
+        $this->data = 'ID: ' . $element->getId() . "  \n"  . $this->cleanupData($this->data);
```
brusch, line 537: @aashan10 I think that's the only relevant change left, right? All the other changes have no more relevancy anymore and can be removed.

brusch (changes requested):

models/Search/Backend/Data.php
aashan10: Well, this and the message queue part I think. These two are the only things needed for resolving the issue.

brusch: Why the message queue stuff? No need to update children if there's no path in the data anymore 😉

brusch (changes requested):

models/Search/Backend/Data.php
aashan10: Oops .. forgot about that 😆 Thanks for reminding .. Ill make the changes accordingly!

hethehe: @brusch & @aashan10 Do we really don't need to update children, there is still the "fullpath" column which has to be updated, or am I wrong?

aashan10: @hethehe That is actually a good point! @brusch should we really get rid of the message queue stuff? We removed fullpath from `data` column, however there is still the `fullpath` column which needs to be updated when the user tries to rename the folder.

brusch: @hethehe totally right, good catch! 
I've created another PR #13417 that should cover this as well. wdyt?

brusch:
Closed in favor of #13417

models/Search/Backend/Data.php
hethehe: looks good, nice solution with the SQL query, thank you

PR #13274, opened 2022-09-29
verify path points to file before reading file

nehlsen:
make sure $fullPath points to a file prior to reading its contents, fixes error exceptions in case of broken configuration

  
  

fixes error exceptions in case of broken configuration


jdreesen (approved):

mcop1:
Thank you!

PR #13276, opened 2022-09-29
move config values into variables to easily change it

tecbird:
Resolves #
move config values into variables to easily change it


robertSt7:
@tecbird Could you please define maxRenderItems in https://github.com/pimcore/pimcore/blob/11.x/bundles/AdminBundle/public/js/pimcore/object/tags/abstractRelations.js, replace the maximum render items (10) in every realtion type with maxRenderItems and pass it as parameter with the default value 10, where necassary. Thanks a lot

robertSt7:
@tecbird Have you had time to look at this?

tecbird:
hey robert, i had some problems with rebasing in this mr, so i opened a new "clean" one https://github.com/pimcore/pimcore/pull/14078

PR #13279, opened 2022-09-29
Add exception when resource not exists for existing asset

### LINKED ISSUE #13260: [Bug]: AssetController uses fpassthru($asset->getStream()), but it can return null
### Expected behavior

Handle null streams correctly

### Actual behavior

If asset stream is null, it produces an error

```
fpassthru(): Argument #1 ($stream) must be of type resource, null given
```

### Steps to reproduce

Have an asset where you cannot obtain its stream and try to download it.

See 

https://github.com/pimcore/pimcore/blob/b2110bb3c416cbf4a7fce967fe98a0e031d642a0/bundles/AdminBundle/src/Controller/Admin/Asset/AssetController.php#L1299
https://github.com/pimcore/pimcore/blob/b2110bb3c416cbf4a7fce967fe98a0e031d642a0/bundles/AdminBundle/src/Controller/Admin/Asset/AssetController.php#L1147

Maybe more.

msoroka:
Resolves #13260

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -1149,6 +1149,10 @@ class AssetController extends ElementControllerBase implements KernelControllerE
 
         $stream = $asset->getStream();
 
+        if (!is_resource($stream)) {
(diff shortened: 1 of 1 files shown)


aryaantony92:
@msoroka Thank you!

PR #13282, opened 2022-09-30
Fix custom reports filtering

mhgtlc:
When apply filter on a certain column in custom reports, only data for filtered column is shown

https://user-images.githubusercontent.com/6807023/193236523-924e7c7a-7105-4607-98c1-26d403c6a954.mov

Resolves #
Filtering of custom reports


mcop1:
@mhgtlc Thank you for your contribution. Something I noticed during testing: Do you get the same error, when you activate the filter for the first time? 
grafik

Would be cool if we could add a fix for that too, e.g. something like 

`$value = $filter['value'] ?? null;`

mhgtlc:
@mcop1, indeed, I got the same error. Fixed it

PR #10423, opened 2021-09-27
Use Symfony Messenger for Search Backend Data

dvesh3:
## Changes in this pull request  
Resolves  partially https://github.com/pimcore/pimcore/issues/6371

PR #13285, opened 2022-09-30
Add option to reindex backend search data that has not been reindexed yet

### LINKED ISSUE #8077: Parallelize SearchBackendReindexCommand
Currently for big Pimcore installations the SearchBackendReindexCommand is un-usable due to doctrine memory leak problems as all Pimcore elements are loaded within one command. I tried it for an installation with 1 million assets and 4 million data objects.

As improvement it should perfectically be possible to parallelize this command using the Webmozarts console parallelization trait.

alyna:
The SearchBackendReindexCommand can't be used because of memory issues. Each time I run _bin/console pimcore:search-backend-reindex_ after a while the process stops and I get the error:
 _PHP Fatal error: Allowed memory size of 1048576000 bytes exhausted (tried to allocate 1769472 bytes) in /var/www/html/vendor/symfony/error-handler/Resources/views/logs.html.php on line 38_

I tested it on a data with 455000 assets, 3700 documents, 3900000 objects.

In this solution I added an option so that when the command is run the whole data is reindexed:  _bin/console pimcore:search-backend-reindex --full=1_
If the command is run without the _full_ parameter, then it will reindex only the data that has not been reindexed yet.

Resolves #8077

## Additional info  
I also found this merge request https://github.com/pimcore/pimcore/pull/10423 but I still get the out of memory error, even with this fix.

### DIFF
--- a/bundles/CoreBundle/src/Command/SearchBackendReindexCommand.php
+++ b/bundles/CoreBundle/src/Command/SearchBackendReindexCommand.php
@@ -23,6 +23,7 @@ use Pimcore\Model\Element\Service;
 use Pimcore\Model\Search;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Input\InputOption;
 
 /**
  * @internal
@@ -34,6 +35,13 @@ class SearchBackendReindexCommand extends AbstractCommand
         $this
             ->setName('pimcore:search-backend-reindex')
             ->setAliases(['search-backend-reindex'])
+            ->addOption(
+                'full',
+                'f',
+                InputOption::VALUE_OPTIONAL,
+                'If the option value is 1, then Pimcore will truncate the search backend data table and reindex the whole data.',
+                null
+            )
             ->setDescription('Re-indexes the backend search of pimcore');
     }
 
@@ -44,7 +52,10 @@ class SearchBackendReindexCommand extends AbstractCommand
     {
         // clear all data
         $db = \Pimcore\Db::get();
-        $db->executeQuery('TRUNCATE `search_backend_data`;');
+        $isFull = $input->getOption('full');
+        if ($isFull) {
+            $db->executeQuery('TRUNCATE `search_backend_data`;');
+        }
 
         $elementsPerLoop = 100;
         $types = ['asset', 'document', 'object'];
@@ -63,6 +74,10 @@ class SearchBackendReindexCommand extends AbstractCommand
                     DataObject::OBJECT_TYPE_FOLDER,
                     DataObject::OBJECT_TYPE_VARIANT,
                 ]);
+            }
+            // Load the data that has already been processed
+            if (!$isFull && $ids = $db->fetchFirstColumn("SELECT id FROM search_backend_data WHERE maintype = '". $type ."' ")) {
+                $list->setCondition($type === 'object' ? 'o_id' : 'id' .' NOT IN (:oids)', ['oids' => $ids]);
             }
 
             $elementsTotal = $list->getTotalCount();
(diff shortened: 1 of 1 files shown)


mcop1:
@alyna Thanks for the pr, but this is changing the default behavior, right? Default right now is to reindex everything, with your pr the default would be to reindex the missing ids only. We would like to keep the default behavior as it is. 

Could you please change the logic in your pr, so that you have to pass a parameter to index the missing ids only?

CodeBrauer:
By indexing only missing IDs with this solution, how would the table be optimized afterwards? (Running this in a cronjob)

bundles/CoreBundle/src/Command/SearchBackendReindexCommand.php
```diff
@@ -64,6 +75,10 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                     DataObject::OBJECT_TYPE_VARIANT,
                 ]);
             }
+            // Load the data that has already been processed
+            if (!$isFull && $ids = $db->fetchFirstColumn("SELECT id FROM search_backend_data WHERE maintype = '". $type ."' ")) {
+                $list->setCondition($type === 'object' ? 'o_id' : 'id' .' NOT IN (:oids)', ['oids' => $ids]);
```
Countryen, line 80: Maybe this is wrong?
Should this resolve to `'o_id'` or `'o_id NOT IN (:oids)'` for type 'object'?

I think it should be changed to:
`$list->setCondition(($type === 'object' ? 'o_id' : 'id') .' NOT IN (:oids)', ['oids' => $ids]);`
(see the extra braces)

We did this and without the proposed change he restarted all objects every time.

jdreesen: As this is for Pimcore 11... wasn't the `o_`-prefix recently removed?

/edit: oh, no, it's not ready yet: #13308

Countryen:
I've had a very similar issue when upgrading to 10.5. Thanks for your PR, I have created a separate command with this content until this is merged into pimcore.

Additionally to the memory issue, the command creates temporary files (var/tmp) and this was too much file size for our setup (~3 Million entries).

Maybe it would also help if the command could be run with something like --async, creating SearchBackendMessages instead of doing it synchronous.

mcop1:
> I've had a very similar issue when upgrading to 10.5. Thanks for your PR, I have created a separate command with this content until this is merged into pimcore.
> 
> Additionally to the memory issue, the command creates temporary files (var/tmp) and this was too much file size for our setup (~3 Million entries).
> 
> Maybe it would also help if the command could be run with something like --async, creating SearchBackendMessages instead of doing it synchronous.

@Countryen Please open a new issue/improvement for your suggestion, otherwise it can get overlooked easily

bundles/CoreBundle/src/Command/SearchBackendReindexCommand.php
```diff
@@ -64,6 +75,10 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                     DataObject::OBJECT_TYPE_VARIANT,
                 ]);
             }
+            // Load the data that has already been processed
+            if (!$isFull && $ids = $db->fetchFirstColumn("SELECT id FROM search_backend_data WHERE maintype = '". $type ."' ")) {
```
brusch, line 79: Using this query without a limit could potentially lead to memory issues, I'd suggest to do them in chunks/batches?

brusch (changes requested):

bundles/CoreBundle/src/Command/SearchBackendReindexCommand.php
```diff
@@ -34,6 +35,13 @@ protected function configure()
         $this
             ->setName('pimcore:search-backend-reindex')
             ->setAliases(['search-backend-reindex'])
+            ->addOption(
+                'full',
+                'f',
+                InputOption::VALUE_OPTIONAL,
```
BlackbitDevs, line 41: Wouldn't `VALUE_NONE` be better? The CLI call then would look like `bin/console pimcore:search-backend-reindex --full`.

fashxp:
Closing due to inactivity. Feel free to ping us for reopening when comments and merge conflicts are resolved.

PR #13289, opened 2022-09-30
Bump `psr/simple-cache` to ^3.0.0

### LINKED ISSUE #13245: Remove psr/simple-cache dependency
### Improvement description

~https://github.com/php-fig/simple-cache/releases/tag/3.0.0 we are 2 major behind, but in one it adds return type and the other one param native types~

aashan10:
Resolves #13245


kingjia90:
It doesn't break anything, because there are no usage at all😄 

As far as i can tell it was used in 6.x, somewhere `lib/Document/Tag/NamingStrategy/Migration/AbstractMigrationStrategy.php`,  that can be seen in [LTS](https://github.com/search?q=org%3Apimcore+SimpleCache&type=code).
It must have been added to composer for BC purposes, as symfony/cache removed simple-cache dependency in [4.3](
https://github.com/symfony/cache/blob/6.1/CHANGELOG.md#430)
and it could be that would break when using symfony 4
https://github.com/pimcore/pimcore/blob/50b169428384f2b988b7698ede90a8d85a2e5b8d/composer.json#L85
(this line is from the same composer.json on where the prs/simple-cache was added to fix some tests by looking at git blame)

So if i am not wrong, this package could actually be removed, could you please confirm @brusch ?

brusch:
@kingjia90 If it's not in use, we should remove it 👍 Thanks!

PR #13291, opened 2022-10-02
Update Video.php

mackrais:
Bug when you use aws s3 thumbnail storage. Method getFullPath return absolutely file url. Is wrong because after second save your video not showing. Because paths is wrong

  
  

Resolves #

### DIFF
--- a/models/Document/Editable/Video.php
+++ b/models/Document/Editable/Video.php
@@ -230,7 +230,7 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
     {
         $path = $this->id;
         if ($this->type === self::TYPE_ASSET && ($video = Asset::getById($this->id))) {
-            $path = $video->getFullPath();
+            $path = $video->getRealFullPath();
         }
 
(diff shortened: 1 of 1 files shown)


dvesh3:
@mackrais thanks for the PR! could you please sign the CLA since this is your first contribution?

mackrais:
> 

Ok. Yes first.

kingjia90:
@mackrais Can you please give a look to this PR https://github.com/pimcore/pimcore/pull/13311 ? 
I've tried this `realFullPath` change but didn't worked for me, instead i've found a possible problem that could be the same issue that you have encountered.

brusch:
I also don't think that this would fix the issue, because `getRealFullPath()` is also just returning a virtual / local path that doesn't exist. https://github.com/pimcore/pimcore/pull/13311 LGTM

kingjia90:
Closing in favor of #13311.

Thank you again for reporting, please feel free to re-open this if it is a different case and/or add any further feedback on the other PR to stay with the topic.

PR #13295, opened 2022-10-03
Added null/empty check for predefinedMetaDataGroups data array

amarjeet6694:
I have configured Azure Blob storage and able to add/update/delete in Azure but when I try to open the asset in Admin panel, I got an error regarding predefinedMetaDataGroups data array.
So, added a null check for the same.

_Would be great if it is merged ASAP.
Thanks_


mcop1:
@amarjeet6694 Sounds like a bug, could you please rebase the pr to 10.5? (10.x is for features, 10.5 for bug fixes)

dvesh3:
🍒 picked on branch 10.5 https://github.com/pimcore/pimcore/commit/39c7fb61cc741637ad2947403830aeca4f4ad905

PR #13305, opened 2022-10-04
[Bug]: RGBA Color doesn't work if user language is german

### LINKED ISSUE #13304: [Bug]: RGBA Color doesn't work if user language is german
### Expected behavior

You should select a alpha color if you have a user in german language.

### Actual behavior

The alpha channel is always FF. If a english user has created a color with alpha channel. The object can't be opend and there is a JS error:
image

After debugging I see the Ext.util.Format.number() function in ColorUtilis.js line 217 doesn't respect the alphaFormat (should be #.## but is #,## after)
image

Is this bug fixed already in Ext JS? Can we make a hotfix in the colorpicker-overrides.js?

### Steps to reproduce

1. Create a color field in the demo (for example in the news class)
2. Create a news object with alpha channel in the color field
3. Change language of the user to german
4. Reload the backend and try to open the object

blankse:
Fix https://github.com/pimcore/pimcore/issues/13304

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/colorpicker-overrides.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/colorpicker-overrides.js
@@ -93,6 +93,13 @@ pimcore.helpers.colorpicker = {
                 if (me.colorPicker) {
                     me.colorPicker.setColor(c);
                 }
+            },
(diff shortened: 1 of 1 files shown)


mcop1:
Thanks!

PR #13309, opened 2022-10-04
[Bug] Fix search bug in HTTP Errors log

perharstrom:
Steps to reproduce:
1. Go to "Marketing -> Search Engine Optimization -> HTTP Errors".
2. Enter anything in the "Filter/Search"-field and hit enter. 
3. Developer tools -> console will reveal that there is an error `Uncaught ReferenceError: filterField is not defined` due to a wrongly named variable.

This PR renames the wrongly named variable and also removes a redundant variable.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/settings/httpErrorLog.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/settings/httpErrorLog.js
@@ -102,7 +102,7 @@ pimcore.settings.httpErrorLog = Class.create({
             listeners: {
                 "keydown" : function (field, key) {
                     if (key.getKey() == key.ENTER) {
-                        var val = field.getValue();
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/settings/httpErrorLog.js
```diff
@@ -102,8 +102,7 @@ pimcore.settings.httpErrorLog = Class.create({
             listeners: {
                 "keydown" : function (field, key) {
                     if (key.getKey() == key.ENTER) {
-                        var input = filterField;
-                        var val = input.getValue();
+                        var val = field.getValue();
```
robertSt7:
```suggestion
                        const val = field.getValue();
```

robertSt7:
@perharstrom Thanks for the PR

PR #13311, opened 2022-10-04
[Bug]: Adds Video Thumbnail support in PublicServices when using Flysystem

kingjia90:
Alternative solution of a problem reported in https://github.com/pimcore/pimcore/pull/13291

## Additional info  
Having thumbnail in cloud storage with flysystem and a video editable on a public page, the video thumbnail won't work because the file eg. `/566/video-thumb__566__content/abcedf_10s_5MB.mp4` wouldn't be served by the Public Service Controller, throwing routing problems or 404 not found. 
This is because `video-thumb` was not covered, so it was just displaying the video image cover (which was working fine due the prefix as `image-thumb`) but couldn't play.

### DIFF
--- a/bundles/CoreBundle/Controller/PublicServicesController.php
+++ b/bundles/CoreBundle/Controller/PublicServicesController.php
@@ -106,6 +106,8 @@ class PublicServicesController extends Controller
 
                             if ($storage->fileExists($storagePath)) {
                                 $thumbnailStream = $storage->readStream($storagePath);
+                                $mime = $storage->mimeType($storagePath);
+                                $fileSize = $storage->fileSize($storagePath);
                             }
                         } else {
                             $time = 1;
@@ -163,9 +165,6 @@ class PublicServicesController extends Controller
                                 $requestedFile = preg_replace('/\.' . $actualFileExtension . '$/', '.' . $requestedFileExtension, $pathReference['src']);
                                 $storage->writeStream($requestedFile, $thumbnailStream);
                             }
-                        }else{
-                            $mime = $storage->mimeType($storagePath);
-                            $fileSize = $storage->fileSize($storagePath);
                         }
                         // set appropriate caching headers
                         // see also: https://github.com/pimcore/pimcore/blob/1931860f0aea27de57e79313b2eb212dcf69ef13/.htaccess#L86-L86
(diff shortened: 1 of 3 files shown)


bundles/CoreBundle/Controller/PublicServicesController.php
```diff
@@ -56,6 +61,37 @@ public function thumbnailAction(Request $request)
             if ($asset->getPath() === ('/' . $prefix)) {
                 // we need to check the path as well, this is important in the case you have restricted the public access to
                 // assets via rewrite rules
+
+                if ($asset instanceof Asset\Video && $request->get('type') == 'video') {
```
dvesh3: Isn't this already handled here https://github.com/pimcore/pimcore/pull/13311/files#diff-80c761bc7a7fb7998e853a73bfc7ddb1a083639cc26b42c314ffa845633d94acR132?

kingjia90: That's for `image-thumb` of a video, in other words the `cover` image, while what i am adding is serving `video-thumb` when `request ->get('type')` matches "video" ("video-thumb") in the path

dvesh3: Ok...then how about we incorporate this into already existing if else checks instead of repeating the code for returning the response?

bundles/CoreBundle/Controller/PublicServicesController.php
```diff
@@ -56,12 +59,15 @@ public function thumbnailAction(Request $request)
...
                 // assets via rewrite rules
+
                 try {
-                    $imageThumbnail = null;
+                    $thumbnail = null;
                     $thumbnailStream = null;
 
                     // just check if the thumbnail exists -> throws exception otherwise
-                    $thumbnailConfig = Asset\Image\Thumbnail\Config::getByName($thumbnailName);
+                    $thumbnailConfigClass = 'Pimcore\Model\Asset\\' . ucfirst($thumbnailType) . '\\Thumbnail\Config';
```
dvesh3:
```suggestion
                    $thumbnailConfigClass = 'Pimcore\\Model\\Asset\\' . ucfirst($thumbnailType) . '\\Thumbnail\Config';
```

bundles/CoreBundle/Controller/PublicServicesController.php
```diff
@@ -56,12 +59,15 @@ public function thumbnailAction(Request $request)
...
+                    $thumbnailConfig = new $thumbnailConfigClass();
```
dvesh3: Not necessary to initialize it for calling static method.

dvesh3:
```suggestion
```

PR #13314, opened 2022-10-05
[Ecommerce] Token Manager: do no silently ignore errors on generation

brusch:


bundles/EcommerceFrameworkBundle/VoucherService/TokenManager/Pattern.php
```diff
@@ -247,6 +248,7 @@ public function insertOrUpdateVoucherSeries()
 
             return $codeSets;
         } catch (\Exception $e) {
+            Logger::error($e);
```
blankse:
```suggestion
            Logger::error((string) $e);
```

bundles/EcommerceFrameworkBundle/VoucherService/TokenManager/Single.php
```diff
@@ -153,6 +154,7 @@ public function insertOrUpdateVoucherSeries()
 
             return trim($this->configuration->getToken());
         } catch (\Exception $e) {
+            Logger::error($e);
```
blankse:
```suggestion
            Logger::error((string) $e);
```

blankse (changes requested):
Exceptions should be cast to string

aashan10 (approved):

Issue #10245, opened 2021-09-06
[Feature]: Custom Asset Types


brusch:
### Feature description

It should be possible to add custom asset types, just as you could do for documents.

brusch:
See also #2270

PR #13319, opened 2022-10-06
[Improvement] Adding functionality to create preview images on InDesign Files

Alanin:
Adding functionality to render preview images from InDesign (indd) Files.

### DIFF
--- a/lib/Helper/InddPreviewGenerationHelper.php
+++ b/lib/Helper/InddPreviewGenerationHelper.php
@@ -26,28 +26,36 @@ class InddPreviewGenerationHelper
      */
     public static function generatePreviewImage(Asset $asset, string $storagePath): void
     {
-        $exiftool = \Pimcore\Tool\Console::getExecutable('exiftool');
-        $filePath = $asset->getLocalFile();
-        $process = new Process([$exiftool, '-b', '-json', $filePath]);
-        $process->run();
-        $output = $process->getOutput();
-        $embeddedMetaData = json_decode($output);
-        $pageImage = reset($embeddedMetaData)->PageImage;
+        try {
+            $exiftool = \Pimcore\Tool\Console::getExecutable('exiftool');
+            $filePath = $asset->getLocalFile();
+            $process = new Process([$exiftool, '-b', '-json', $filePath]);
+            $process->run();
+            $output = $process->getOutput();
+            $embeddedMetaData = json_decode($output);
+            $pageImage = reset($embeddedMetaData)->PageImage;
 
-        $tmpPath = sprintf('%s/var/tmp/indd_%s.jpg', PIMCORE_WEB_ROOT, $asset->getFilename());
-        $tmpPathPng = sprintf('%s/var/tmp/indd_%s.png', PIMCORE_WEB_ROOT, $asset->getFilename());
-        if (file_exists($tmpPath)) {
+            $tmpPath = sprintf('%s/var/tmp/indd_%s.jpg', PIMCORE_WEB_ROOT, $asset->getFilename());
+            $tmpPathPng = sprintf('%s/var/tmp/indd_%s.png', PIMCORE_WEB_ROOT, $asset->getFilename());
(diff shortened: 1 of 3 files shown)


brusch:
I'm not entirely happy with the current approach, for the following reasons: 
1. since we're just extracting the embedded thumbnail of the `indd` file, we should only use it for preview purposes in the admin UI
2. in the admin UI we should make clear that this is just a preview and that this file-type cannot be used for thumbnails (e.g. image editables, image data-types, ...) 
3. `indd` shouldn't be generally considered as an image imho

So I'm thinking of having this as an admin UI - only functionality, that can be handled in an action, wdyt?

brusch:
@Alanin ping 🏓😊
Please also see the failing PHPStan check: 

```
Error: Caught class Pimcore\Helper\Exception not found.
  Line   lib/Helper/InddPreviewGenerationHelper.php                           
  57     Caught class Pimcore\Helper\Exception not found.                     
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
```

brusch:
@Alanin 🏓 any updates on this? Thanks 😊

brusch:
@Alanin one more try - any chance to incorporate my feedback?

Alanin:
Hey @brusch,

sorry I am very busy at the moment. not able to address all issues here.

1. true
2. true :)  
3. indd files contain a lot of meta data. using just 'document' as type would prevent showing those data.

tbh. I don't really like this solution. I would like to see a way to have some kind of file handlers which helps community to add functions for yet unsupported filetypes. so we could add methods to generate thumbnails or define functions for exporting derivates when needed.

do you think something like that would be possible in future? I think a more generic way of handling assets would be great ;)

brusch:
@Alanin that's planned for `v11` , see https://github.com/pimcore/pimcore/issues/10245
You could help is in providing a PR for that 😉

brusch:
With the current state I'm still not happy, as it's an backend functionality it should probably covered in the `AdminBundle` and also the file headers need to be correct. 
If you don't have time for that, I'd close it for now and create a follow up issue for us.

SandraRGitHub:
Hi, we tried this code for getting previews on indesign files, but we only get it to work with multiple page indd files (it displays only the preview for first page) but it does not give a preview on single page indd files. 
Any suggestions or help? We are running pimcore portals on v2.3.1.

Alanin:
@SandraRGitHub is the preview integrated in the single page files as well? the code is only extracting the low-res preview images provided by InDesign.

those are saved in the EXIF Metadata in Field PageImage

SandraRGitHub:
@Alanin we actually did a test, we created a new indd file with just one single page. And the checked the box for "Always save preview images with documents". 
We checked the raw data for the file in page info and there is a tagg <xmpGImg:image> in the <xmp:PageInfo> of the file.

But when we upload the file to the pimcore DAM portal then we get no preview, and instead it just displays a blue icon a question mark. So something seems not right...

brusch:
@SandraRGitHub this feature is not yet implemented, so it can't work 😉

Alanin:
Tbh it's not a good idea to use that code here. Pimcore implemented an official way to add adapters to create previews for unsupported filetypes. 

It would be way better to make a bundle to extract the images using the official way

Alanin:
I'll have a look into it tomorrow, maybe I can refactor that code here to a standalone bundle using the new apis.

But that would be Pimcore 11 and not compatible to 10.x

Alanin:
@SandraRGitHub i have refactored the code to a Pimcore 11 compatible bundle. When you have a running Pimcore 11 instance, can you please get in touch with me. I would like you to test the bundle before I release that bundle to the public.

Alanin:
@brusch i have tried to implement the bundle, but seems the is still some more work in core to make it easy and convenient to implement own asset types... https://github.com/pimcore/pimcore/discussions/15707

brusch:
Closing this for now due to inactivity. 
Might  be a nice addition for Studio UI in future 😊

PR #13324, opened 2022-10-06
[Bug]: DocumentController::docTypesAction fixes 

### LINKED ISSUE #13313: [Bug]: Editing multiple Document Type rows causes issues
### Expected behavior

When editing Document Types they should all be updated with no error.

### Actual behavior

Sometimes ExtJs will send multiple edits simultaneously if the user makes many small edits after each other.

Because `bundles/AdminBundle/Controller/Admin/Document/DocumentController.php` expects only a single item rather than an array of items it breaks fatally in the method

https://github.com/pimcore/pimcore/blob/caab4a9186b302dfcadcb48b4bb414fbecabb350/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php#L708

Screen shot showing debug:
image

Screen showing two rows being edited, this causes `$data` in DocumentController to be an `array`:
image

### Steps to reproduce

1. Add a number of Document Types
2. Update them (e.g. the name) in quick succession, best way appears to quickly set focus on a second item quickly after pressing the `enter` key

NiklasBr:

- Routing name and behaviour
- Comparison operator
- data getter

Resolves #13313

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
+++ b/bundles/AdminBundle/Controller/Admin/Document/DocumentController.php
@@ -699,7 +699,7 @@ class DocumentController extends ElementControllerBase implements KernelControll
     }
 
     /**
-     * @Route("/doc-types", name="pimcore_admin_document_document_doctypes", methods={"PUT", "POST","DELETE"})
+     * @Route("/doc-types", name="pimcore_admin_document_document_doctypes", methods={"PUT", "POST", "DELETE"})
      *
      * @param Request $request
      *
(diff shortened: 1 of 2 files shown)


NiklasBr:
Codeception fails on:

> Error: Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_ab0b75b6-761b-4c22-b638-249feadcf859/24c4aa6d-dc26-4470-a350-82be1c07e072.tar.gz. return code: 2.

aryaantony92:
@NiklasBr Thank you!

PR #13328, opened 2022-10-06
[Documents] Editable Dialog Box - Add description support

### LINKED ISSUE #9656: Improvement: EditableDialogBox - Description
# Feature Request

## Feature description

It would be nice if there was a possibility to add a description text to editables inside of the EditableDialogBox.
Just a simple textbox that will be displayed over the editable kind of like the "description" option of the [ToolboxBundle](https://github.com/dachcom-digital/pimcore-toolbox).

This would be great for explainig more complex editables and the needed input to the user.

Brainshaker95:
Resolves #9656

This PR extends the editable dialog box config with a `description`.  

## Additional info  

Sometimes an editables label isn't enough to properly describe what it will be used for. With a description above a more detailed explanation can be achieved.

```php
(new EditableDialogBoxConfiguration())->setItems([
    [
        'type' => 'input',
        'name' => 'SEO Headline',
        'config' => [
            'description' => 'This headline is used for search engine blah blah.',
            // With custom HTML
            // 'description' => '<b>This is now bold</b>',
        ],
    ],
]);
```
image

### DIFF
--- a/bundles/AdminBundle/public/js/pimcore/document/editables/area_abstract.js
+++ b/bundles/AdminBundle/public/js/pimcore/document/editables/area_abstract.js
@@ -110,10 +110,20 @@ pimcore.document.area_abstract = Class.create(pimcore.document.editable, {
                         html: templateEl.innerHTML
                     };
                 } else {
+                    var templateHTML = templateEl.innerHTML;
+                    if (editablesInBox[config['name']]['config'] 
+                        && editablesInBox[config['name']]['config']['description']) {
+                        
+                        var descriptionHTML = '<div style="font-size: 14px; margin-bottom: 10px;">'
+                          + editablesInBox[config['name']]['config']['description']
+                          + '</div>';
+
+                        templateHTML = descriptionHTML + templateHTML;
+                    }
                     return {
                         xtype: 'fieldset',
                         title: config['label'] ?? config['name'],
-                        html: templateEl.innerHTML
+                        html: templateHTML
                     };
                 }
             }
--- a/bundles/AdminBundle/public/js/pimcore/document/editables/area_abstract.js
+++ b/bundles/AdminBundle/public/js/pimcore/document/editables/area_abstract.js
@@ -101,23 +101,25 @@ pimcore.document.area_abstract = Class.create(pimcore.document.editable, {
             let templateId = 'template__' + editablesInBox[config['name']].getId();
             var templateEl = document.getElementById(templateId);
             if(templateEl) {
+                var templateHTML = templateEl.innerHTML;
+
+                if (config['description']) {
+                    var descriptionHTML = '<div style="font-size: 14px; margin-bottom: 10px;">'
+                        + config['description']
+                        + '</div>';
+
+                    templateHTML = descriptionHTML + templateHTML;
+                }
+
                 if(typeof editablesInBox[config['name']]['renderInDialogBox'] === "function") {
                     if (editablesInBox[config['name']]['config']) {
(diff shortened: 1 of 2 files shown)


Brainshaker95:
@mcop1 Thank you for taking a look at my PR and adding it to the next Pimcore version! I can't count the times I needed a feature like this and now it will be added to the core, that's awesome.
Would you mind adding the "hacktoberfest-accepted" label to my PR? Please don't get me wrong. I don't only contribute since it's Hacktober but I wouldn't mind the fact that this PR counts towards my progress.

brusch:
@Brainshaker95 ✅ `hacktoberfest-accepted`

aryaantony92:
@Brainshaker95 Thank you for the PR and it looks good. I think it would be great if we include the description config in the docs also https://github.com/pimcore/pimcore/blob/ea407a2574ca8cab09e019e1d38fbac50e6a92dd/doc/Development_Documentation/03_Documents/01_Editables/02_Areablock/02_Bricks.md?plain=1#L266

Brainshaker95:
@aryaantony92 No problem, I can update the docs.  
But thinking about it now that I'm looking at it again:
```php
$config->setItems([
    [
        'type' => 'input',
        'label' => 'Some additional Text', // labels are optional
        'name' => 'myDialogInput'
    ],
// ...
```
Is it correct that these labels are only used for the dialog box as well?  
We could move the description one level up so that the `config` key is not needed. Would this break something or is this maybe something we want?  
To be honest I don't really understand why the label is not part of the config array. `type` and `name` of course are neccessary to construct an editable but since `label` is not it should be part of the config array shouldn't it?

I'm not even sure myself what the best solution would be, but I'm happy to provide changes to this PR if you have some feedback. Maybe you can explain the difference between top-level options and options underneath `config`.  
I will update the docs as soon as this is cleared up 👍

aryaantony92:
@Brainshaker95 Basically, the $config array is for specifying the additional configurations. For eg: in case of a select/multiselect editable, the $config array would be like this https://github.com/pimcore/pimcore/blob/ea407a2574ca8cab09e019e1d38fbac50e6a92dd/doc/Development_Documentation/03_Documents/01_Editables/02_Areablock/02_Bricks.md?plain=1#L346
In case of the new entry 'description', you can move it to the top level, but it should be optional.
Like 
$config->setItems([
    [
        'type' => 'input',
        'label' => 'Some additional Text', // labels are optional
        'name' => 'myDialogInput',
        'description' => 'This headline is used for search engine blah blah.' // description is optional
    ],

Brainshaker95:
@aryaantony92 Thank you for the feedback. So basically top level options apply to all editables and `config` options are editable specific is what I understood. 
Changes are added

aryaantony92:
@Brainshaker95 To add description to the editable like checkbox, we need to append the description in html tag to this if block also https://github.com/pimcore/pimcore/blob/ab450994de5bf7750430546c9be5cef22184a554/bundles/AdminBundle/public/js/pimcore/document/editables/area_abstract.js#L104  The description can be appended to the html variable (descriptionHTML) before the if/else block so that it would be applied to both parts.

Brainshaker95:
@aryaantony92 Ah sorry I missed that. I didn't fully understand what it was used for 😅
Anyway, I fixed it

aryaantony92:
@Brainshaker95 Thank you. The PR looks good and ready to be merged. Feel free to ask if you need more clarification. 

> @aryaantony92 Ah sorry I missed that. I didn't fully understand what it was used for 😅 Anyway, I fixed it

aryaantony92 (approved):

PR #13329, opened 2022-10-07
fixes the height of the object preview iframe

aweichler:
In the preview tab of an data object, the iframe height was not correct. This PR fixes this issue.

## Additional info  
You can see the problem in the pimcore demo (https://demo.pimcore.fun/admin/login/deeplink?object_81_object -> click to the preview tab). The footer is not fully shown. You can see that if you put a border to the footer element or if you watch the scollbars (they disappear for a few px).


mcop1:
Thank you!

PR #13077, opened 2022-09-06
[Improvement]: consistently use .yaml not .yml

robertSt7:
resolve #11014 

- [x] https://github.com/pimcore/demo/issues/354 
- [x] https://github.com/pimcore/data-hub/issues/604
- [x] https://github.com/pimcore/workflow-designer/issues/46
- [x] https://github.com/pimcore/statistics-explorer/issues/51
- [x] https://github.com/pimcore/saas-tools/issues/27
- [x] https://github.com/pimcore/rest-webservice-bundle/issues/23
- [x] https://github.com/pimcore/portal-engine/issues/234
- [x] https://github.com/pimcore/perspective-editor/issues/95
- [x] https://github.com/pimcore/data-importer/issues/252
- [x] https://github.com/pimcore/data-hub-simple-rest/issues/71
- [x] https://github.com/pimcore/data-hub-productsup/issues/43
- [x] https://github.com/pimcore/data-hub-file-export/issues/75
- [x] https://github.com/pimcore/data-hub-ci-hub/issues/15
- [x] https://github.com/pimcore/translations-provider-interfaces/issues/82
- [x] https://github.com/pimcore/openid-connect/issues/27
- [x] https://github.com/pimcore/headless-documents/issues/84
- [x] https://github.com/pimcore/customer-data-framework/issues/332

PR #13331, opened 2022-10-07
Renaming routing.yml to routing.yaml

blankse:
The routing.yml was changed in the demo to routing.yaml, but not in the core. 
https://github.com/pimcore/demo/pull/355
For skeleton it is also planned:
https://github.com/pimcore/skeleton/pull/118

I get following error in the demo:

`Unable to find file "@PimcoreCoreBundle/config/routing_dev.yaml" in @PimcoreCoreBundle/config/routing_dev.yaml (which is being imported from "/var/www/html/config/routes/dev/routes.yaml"). Make sure the "PimcoreCoreBundle" bundle is correctly registered and loaded in the application kernel class. If the bundle is registered, make sure the bundle path "@PimcoreCoreBundle/config/routing_dev.yaml" is not empty.`


bundles/CoreBundle/config/routing.yaml
```diff
@@ -26,7 +26,7 @@ _pimcore_service_thumbnail:
 _fos_js_routing:
     resource: "@FOSJsRoutingBundle/Resources/config/routing/routing-sf4.xml"
 
-# auto loads routes from all bundles in config/pimcore/routing.yml
+# auto loads routes from all bundles in config/pimcore/routing.yaml
```
kingjia90, line 29: Thank you for the PR! 
This is actually covered in https://github.com/pimcore/pimcore/pull/13077, it's a known issue with 11.x as we all have this problem until that PR get merged.
@robertSt7 Can you add this change in the comments, i have noticed that this change is missing in your PR

brusch: @kingjia90 so we can close this in favor of https://github.com/pimcore/pimcore/pull/13077?

blankse:
@kingjia90 Okay, thanks for clarifying.

PR #13334, opened 2022-10-07
Fix twig example code to actually use a comment as intended

danrot:
I am not absolutely if this change is correct, as I currently do not really understand what should be happening in this example. But my guess would be that the changed passage should be a twig comment. The way it is written now it was confusing to me.

  
  

Resolves #


danrot:
I do not think that any of the review checklist item applies here. Let me know if I am wrong and you need anything from me.

aryaantony92:
@danrot Thank you for the PR!

PR #13338, opened 2022-10-07
add `is_string()` before `strlen()` on possible `null` values in `Pimcore\Model\DataObject\ClassDefinition\Data\Time`

sdrewergutland:
Resolves deprecated use of `strlen()`on possible null values in `Pimcore\Model\DataObject\ClassDefinition\Data\Time`

### To reproduce
* Enable deprecation warnings
* Have a time field with any min/max value as `null`
* Run classes-rebuild
* See deprecations being logged

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Time.php
+++ b/models/DataObject/ClassDefinition/Data/Time.php
@@ -143,7 +143,7 @@ class Time extends Model\DataObject\ClassDefinition\Data\Input
      */
     public function isEmpty($data)
     {
-        return is_string($data) && strlen($data) !== 5;
+        return !is_string($data) || strlen($data) !== 5;
     }
 
     /**
--- a/models/DataObject/ClassDefinition/Data/Time.php
+++ b/models/DataObject/ClassDefinition/Data/Time.php
@@ -122,7 +122,7 @@ class Time extends Model\DataObject\ClassDefinition\Data\Input
                 throw new Model\Element\ValidationException('Value in field [ '.$this->getName().' ] is not at least ' . $this->getMinValue());
             }
 
-            if (is_string($this->getMaxValue()) && strlen($this->getMaxValue()) && $this->isLater($this->getMaxValue(), $data)) {
+            if ($this->getMaxValue() && $this->isLater($this->getMaxValue(), $data)) {
                 throw new Model\Element\ValidationException('Value in field [ ' . $this->getName() . ' ] is bigger than ' . $this->getMaxValue());
             }
         }
--- a/models/DataObject/ClassDefinition/Data/Time.php
+++ b/models/DataObject/ClassDefinition/Data/Time.php
@@ -143,7 +143,7 @@ class Time extends Model\DataObject\ClassDefinition\Data\Input
      */
     public function isEmpty($data)
     {
-        return !is_string($data) || strlen($data) !== 5;
+        return !is_string($data) || !preg_match('/^(2[0-3]|[01][0-9]):[0-5][0-9]$/', $data);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


brusch:
Should we maybe just cast to `string` for better readability?

models/DataObject/ClassDefinition/Data/Time.php
```diff
@@ -113,16 +113,16 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
             throw new Model\Element\ValidationException('Wrong time format given must be a 5 digit string (eg: 06:49) [ '.$this->getName().' ]');
         }
 
-        if (!$omitMandatoryCheck && strlen($data)) {
+        if (!$omitMandatoryCheck && is_string($data) && strlen($data)) {
```
blankse:
```suggestion
        if (!$omitMandatoryCheck && $data) {
```
Should be empty or a string at this position. Otherwise it should already thrown a exception at this position. So I think it is enough to check it without strlen()

models/DataObject/ClassDefinition/Data/Time.php
```diff
@@ -143,7 +143,7 @@ public function isDiffChangeAllowed($object, $params = [])
      */
     public function isEmpty($data)
     {
-        return strlen($data) !== 5;
+        return !is_string($data) || strlen($data) !== 5;
```
blankse:
```suggestion
        return !is_string($data) || !preg_match('/^(2[0-3]|[01][0-9]):[0-5][0-9]$/', $data);
```
Maybe we should also here make the preg_math and not only the length

models/DataObject/ClassDefinition/Data/Time.php
```diff
@@ -113,16 +113,16 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
...
         }
 
-        if (!$omitMandatoryCheck && strlen($data)) {
+        if (!$omitMandatoryCheck && is_string($data) && strlen($data)) {
             if (!$this->toTime($data)) {
                 throw new Model\Element\ValidationException('Wrong time format given must be a 5 digit string (eg: 06:49) [ '.$this->getName().' ]');
             }
 
-            if (strlen($this->getMinValue()) && $this->isEarlier($this->getMinValue(), $data)) {
+            if (is_string($this->getMinValue()) && strlen($this->getMinValue()) && $this->isEarlier($this->getMinValue(), $data)) {
```
blankse:
```suggestion
            if ($this->getMinValue() && $this->isEarlier($this->getMinValue(), $data)) {
```
Should be `string|null`, so I think it is enough to check it without `strlen()`

models/DataObject/ClassDefinition/Data/Time.php
```diff
@@ -113,16 +113,16 @@ public function checkValidity($data, $omitMandatoryCheck = false, $params = [])
...
                 throw new Model\Element\ValidationException('Value in field [ '.$this->getName().' ] is not at least ' . $this->getMinValue());
             }
 
-            if (strlen($this->getMaxValue()) && $this->isLater($this->getMaxValue(), $data)) {
+            if (is_string($this->getMaxValue()) && strlen($this->getMaxValue()) && $this->isLater($this->getMaxValue(), $data)) {
```
blankse:
```suggestion
            if ($this->getMaxValue() && $this->isLater($this->getMaxValue(), $data)) {
```
Should be `string|null`, so I think it is enough to check it without `strlen()`

blankse (changes requested):

models/DataObject/ClassDefinition/Data/Time.php
sdrewergutland: Wouldn't that cause a different behavior for falsy values like "0"?

blankse: @sdrewergutland minValue/maxValue should be the same string schema like the data. The setters always set a string with 5 digit (eg: 06:49). So I think "0" is not allowed.

brusch: Yep, agree 😊

sdrewergutland: Ah, I see,- didnt take the context into consideration; thanks for the explanation:)

blankse (approved):
LGTM now :)

robertSt7:
@sdrewergutland Thanks for the PR

PR #13339, opened 2022-10-10
[Bug]: Can't filter classification store booleanSelect type in grid

### LINKED ISSUE #13096: [Bug]: Can't filter classification store booleanSelect type in grid
### Expected behavior

Be able to filter grid list on classification store field of type booleanSelect.

### Actual behavior

Filter grid list on classification store field of type booleanSelect gives error.

### Steps to reproduce

- add CS key for booleanSelect
image

- add CS store field in Product class
- add CS key in Grid
- try to filter
image

- error
image

robertSt7:
resolve #13096

### DIFF
--- a/models/DataObject/ClassDefinition/Data.php
+++ b/models/DataObject/ClassDefinition/Data.php
@@ -580,7 +580,9 @@ abstract class Data implements DataObject\ClassDefinition\Data\TypeDeclarationSu
         }
 
         if (in_array($operator, DataObject\ClassDefinition\Data::$validFilterOperators)) {
+
             $trailer = '';
+            //the db interprets 0 as NULL -> if empty (0) is selected in the filter, we must also filter for NULL
             if($value === '\'0\'' || is_array($value) && in_array(0, $value)) {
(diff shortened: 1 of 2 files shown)


models/DataObject/ClassDefinition/Data.php
```diff
@@ -579,7 +580,19 @@ public function getFilterConditionExt($value, $operator, $params = [])
         }
 
         if (in_array($operator, DataObject\ClassDefinition\Data::$validFilterOperators)) {
-            return $key . ' ' . $operator . ' ' . $value . ' ';
+            $trailer = '';
+            if($value === '\'0\'' || is_array($value) && in_array(0, $value)) {
```
brusch, line 590: I think we should explain the following lines a bit :)

mcop1 (commented):
The filter itself works now, but there is still an exception. 
1.) Make sure the checkboxes for "yes", "no" and "empty" aren´t checked
2.) Klick on the checkbox next to the filter button
grafik

results in

grafik

PR #13343, opened 2022-10-10
[Feature][Notes] Make user-created notes deletable

### LINKED ISSUE #3396: Deleting Notes

GALCF:

Resolves #3396

With this pull request, user-created notes will be deletable.  
With `$note->setIsDeletable(true)` the note can also be changed to deletable via PHP API.  

DB-Changes: Adds a column `isDeletable TINYINT(1) DEFAULT 0` to the `notes` table.

### DIFF
--- a/bundles/AdminBundle/public/js/pimcore/element/notes.js
+++ b/bundles/AdminBundle/public/js/pimcore/element/notes.js
@@ -173,6 +173,22 @@ pimcore.element.notes = Class.create({
                 }]
             });
 
+            columns.push({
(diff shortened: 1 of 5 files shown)


GALCF:
Nevermind, duplicate of https://github.com/pimcore/pimcore/pull/13206

PR #13346, opened 2022-10-11
[Bug]: empty values in $context in CalculatorClassInterface

### LINKED ISSUE #13098: [Bug]: empty values in $context in CalculatorClassInterface
### Expected behavior

`$context` should have set according data in instances of CalculatorClassInterface 

### Actual behavior

$context has empty values for index, position and groupId in CalculatorClassInterface within object bricks

maybe related to #13028 ?

image

### Steps to reproduce

Create a calculated value field within an object brick and check the values of `$context`

`public function compute(Concrete $object, CalculatedValue $context): string`

mcop1:
Resolves #13098

### DIFF
--- a/models/DataObject/Data/CalculatedValue.php
+++ b/models/DataObject/Data/CalculatedValue.php
@@ -64,7 +64,7 @@ class CalculatedValue implements OwnerAwareFieldInterface
      *
      * @param string $ownerType
      * @param string|null $ownerName
-     * @param int|null $index
+     * @param int|string|null $index
      * @param string|null $position
      * @param int|null $groupId
      * @param int|null $keyId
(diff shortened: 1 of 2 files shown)


models/DataObject/Data/CalculatedValue.php
```diff
@@ -64,7 +64,7 @@ public function __construct($fieldname)
      *
      * @param string $ownerType
      * @param string|null $ownerName
-     * @param int|null $index
+     * @param int|string|null $index
```
blankse, line 67: Should be also changed for the attribute and the getter return

mcop1: You are way too fast 💯 
And I am really looking forward to having everything typed.

PR #13347, opened 2022-10-11
[Mail] Renderer email content twig templates in a sandbox

dvesh3:
Mail twig templates should be rendered in a sandbox with restricted Twig security policy.
Resolves https://github.com/pimcore/planning/issues/102
## Additional info  
please see https://twig.symfony.com/doc/2.x/api.html#sandbox-extension


lib/Mail.php
```diff
@@ -654,10 +654,14 @@ private function getDebugMailRecipients(array $recipients): array
...
-        $twig = $templatingEngine->getTwigEnvironment();
+        $twig = $templatingEngine->getTwigEnvironment(true);
         $template = $twig->createTemplate($string);
 
-        return $template->render($this->getParams());
+        $rendered = $template->render($this->getParams());
+
+        $templatingEngine->disableSandboxExtensionFromTwigEnvironment();
+
+        return $rendered;
```
jdreesen: What about writing it like this?

```suggestion
        try {
            $twig = $templatingEngine->getTwigEnvironment(true);
            $template = $twig->createTemplate($string);

            return $template->render($this->getParams());
        } finally {
            $templatingEngine->disableSandboxExtensionFromTwigEnvironment();
        }
```

fashxp (changes requested):
- I would suggest to extract the allowed `tags`, `filters` and `functions` to a Symfony config ... with quite restrictive default config values
- Also consider other places like text layout component in data objects

lib/Mail.php
dvesh3: yes, looks cleaner. thanks!

lib/Templating/TwigDefaultDelegatingEngine.php
```diff
@@ -100,14 +95,40 @@ public function isDelegate()
         return $this->delegate;
     }
 
-    /**
-     * @return Environment
-     */
-    public function getTwigEnvironment(): Environment
+    public function getTwigEnvironment($sandboxed = false): Environment
```
blankse:
```suggestion
    public function getTwigEnvironment(bool $sandboxed = false): Environment
```

blankse (changes requested):

bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -2276,4 +2277,44 @@ private function addGlossaryNode(ArrayNodeDefinition $rootNode): void
                             ->useAttributeAsKey('name')
                             ->prototype('scalar');
     }
+
+    /**
+     * @param ArrayNodeDefinition $rootNode
+     */
```
blankse:
```suggestion
```

lib/Templating/TwigDefaultDelegatingEngine.php
```diff
@@ -100,14 +95,40 @@ public function isDelegate()
...
+        }
+
         return $this->twig;
     }
 
+    public function disableSandboxExtensionFromTwigEnvironment(): void
+    {
+        if ($this->twig->hasExtension(SandboxExtension::class)) {
+            $sandbox = $this->twig->getExtension(SandboxExtension::class);
+            /** @var SandboxExtension $sandbox */
```
jdreesen:
```suggestion
            /** @var SandboxExtension $sandbox */
            $sandbox = $this->twig->getExtension(SandboxExtension::class);
```

bundles/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -2276,4 +2277,41 @@ private function addGlossaryNode(ArrayNodeDefinition $rootNode): void
...
+    {
+        $rootNode
+            ->children()
+            ->arrayNode('templating_engine')
+            ->addDefaultsIfNotSet()
+            ->children()
+                ->arrayNode('twig')
+                ->addDefaultsIfNotSet()
+                ->children()
+                    ->arrayNode('security_policy')
```
fashxp: we should add some inline description here

fashxp (changes requested):

fashxp (commented):
we should try to provide more reasonable error messages: 

right now when opening a data object with a bad template I get
image

and when opening preview in class editor, we get just a 500 at all 
image

bundles/CoreBundle/Resources/config/pimcore/default.yaml
```diff
@@ -293,6 +293,14 @@ pimcore:
               'abbr', 'option', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'
           ]
 
+    templating_engine:
+        twig:
+            security_policy:
```
brusch:
```suggestion
            sandbox_security_policy:
```
Maybe the naming should be a bit more specific here, otherwise someone could assume that this applies globally?

brusch (changes requested):

dvesh3:
> we should try to provide more reasonable error messages:

@fashxp done by rethrowing a message with context in mailer and setting the error message to the text layout html.

bundles/CoreBundle/Resources/config/pimcore/default.yaml
dvesh3: done

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,6 +1,22 @@
...
+         Please use following configuration to allow more tags, filters, properties, methods in template rendering:
+  ```yaml
+  pimcore:
+        templating_engine:
+            twig:
+              sandbox_security_policy:
+                tags: ['if']
+                filters: ['upper']
+                methods:
+                    Article: ['getTitle', 'getBody']
```
fashxp: Do we really need that?

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,6 +1,22 @@
...
+                properties:
+                    Article: ['title', 'body']
```
fashxp: Do we really need that?

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,6 +1,22 @@
 # Upgrade Notes
 
 ## 10.5.8
+- [Twig] Sending mails and Dataobject Text Layouts, which allow rendering user controlled twig templates are now executed in a sandbox with restrictive security policies for tags, filters, functions and son on.
```
fashxp: we need to add that information also to documentation it self... 

like [here](https://pimcore.com/docs/pimcore/current/Development_Documentation/Development_Tools_and_Details/Email_Framework/index.html) and [here](https://pimcore.com/docs/pimcore/current/Development_Documentation/Objects/Object_Classes/Layout_Elements/Dynamic_Text_Labels.html)

fashxp (changes requested):

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
dvesh3: yes, this is needed to whitelist the method and property calls on objects as the sandbox execution disallows everything by default.

So basically for sending emails with html body: `{{ pimcore_object(769).title }}`, I need to add the config:
```yaml
twig:
  sandbox_security_policy:
    methods:
        \Pimcore\Model\DataObject\News: ['getTitle']
    functions: ['pimcore_object']
```

dvesh3: This could be a BC break for existing implementations, so i would suggest that we add a config to enable sandbox execution for email templates rendering. wdyt?

dvesh3: I have re-used the Twig SecurityPolicy class to avoid performing checks for object methods and properties. Please see https://github.com/pimcore/pimcore/pull/13347/commits/aa5d1534784b99860da300390f672ff14d91299c

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
dvesh3: done

fashxp (approved):
looks good to me now. 
don't forget to create a security advisory as stated here...

lib/Templating/TwigDefaultDelegatingEngine.php
```diff
@@ -100,14 +95,38 @@ public function isDelegate()
...
+            if (!$this->twig->hasExtension(SandboxExtension::class)) {
+                $securityPolicy = $this->config['templating_engine']['twig']['sandbox_security_policy'];
+
+                $tags = $securityPolicy['tags'];
+                $filters = $securityPolicy['filters'];
+                $functions = $securityPolicy['functions'];
+
+                $policy = new SecurityPolicy($tags, $filters, $functions);
+                $sandbox = new SandboxExtension($policy);
+                $this->twig->addExtension($sandbox);
```
brusch: I think this dynamic approach could cause some issues in case the Twig environment was already initialized previously, then `addExtension()` will throw an exception. 
I'd propose to register `SandboxExtension` as a service here: https://github.com/pimcore/pimcore/blob/3a539e605e5bf4f4924c83366837c0b840933315/bundles/CoreBundle/config/templating_twig.yaml#L76 

It will the get automatically registered as a Twig extension and here, we can just use `enableSandbox()` and `disableSandbox()` depending on how the Twig env was requested. WDYT?

brusch (changes requested):

lib/Templating/TwigDefaultDelegatingEngine.php
dvesh3: yes, that's a good catch. I registered it as a service now.

brusch (approved):
LGTM 👍

PR #13349, opened 2022-10-11
[StaticPageGenerator] Fix generating static content for homepage

### LINKED ISSUE #13353: [Bug]: Statically-generated homepage content is replaced with contents from a different page
### Expected behavior

The Pimcore homepage should support static page generation.

### Actual behavior

Any document page access _other_ than the homepage will overwrite the statically-generated contents of the homepage. This causes nginx to return incorrect contents when the homepage uri is requested.

### Steps to reproduce

_All of this is reproduced on a fresh copy of `pimcore/demo`_

1. Configure the homepage to be statically generated
    image
1. Access any other document page on site, eg. `/en/Events`
1. Visit the homepage -- you'll see content returned of the last page you visited even though your URL is now `/`. Essentially, Pimcore is returning the wrong page contents for your homepage!
1. Look at `.html` to confirm that the statically-generated homepage file contains the output of a different page (or even JSON):
    image

dvesh3:
Resolves #13353

## Additional info  
Update nginx config to have URI mapping and restart the docker container to test the fix.

### DIFF
--- a/bundles/CoreBundle/EventListener/Frontend/StaticPageGeneratorListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/StaticPageGeneratorListener.php
@@ -143,6 +143,11 @@ class StaticPageGeneratorListener implements EventSubscriberInterface
             return;
         }
 
+        if (($contentType = $event->getResponse()->headers->get('Content-Type'))
+            && stripos($contentType, 'text/html') === false) {
+            return;
+        }
+
         //return if request is from StaticPageGenerator
         if ($request->attributes->has('pimcore_static_page_generator')) {
             return;
--- a/bundles/CoreBundle/EventListener/Frontend/StaticPageGeneratorListener.php
+++ b/bundles/CoreBundle/EventListener/Frontend/StaticPageGeneratorListener.php
@@ -30,6 +30,7 @@ use Pimcore\Model\Document\PageSnippet;
 use Pimcore\Model\Site;
 use Pimcore\Tool\Storage;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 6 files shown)


jdreesen:
The nginx config needs to be updated in the skeleton and demo, too.

bundles/CoreBundle/EventListener/Frontend/StaticPageGeneratorListener.php
```diff
@@ -143,6 +143,11 @@ public function onKernelResponse(ResponseEvent $event)
             return;
         }
 
+        if (($contentType = $event->getResponse()->headers->get('Content-Type'))
```
tmannherz: FWIW, I think both the `onKernelResponse` _and_ `onKernelRequest` methods in this class should further restrict the requests that are processed by this listener. The restrictions I've added to an overridden class that extends this are:
```php
if ($request->getMethod() !== 'GET' || !in_array('text/html', $request->getAcceptableContentTypes())) {
    return;
}
```
You wouldn't ever want to return statically-generated page contents for non-GET requests or if the client doesn't accept html.

kingjia90:
Tested and it works now, but in the case there's a page called `/home`, and both `root`(id:1) and `/home` do generate static pages, it wouldn't work for `/home` as it will be replaced by the content of `/`.

Probably it would help to have a note about this case in the documentation, or prepend it with a symbol that is stripped by Pimcore but it's safe to use in uri/nginx.

dvesh3:
@kingjia90 Good catch! I have adjusted the PR to use `/%home` uri in nginx mapping as this is not possible to create documents with this char prefix and also not possible in pretty url. I have also incorporated the changes suggested by @tmannherz. thanks!

kingjia90:
LGTM and works fine. 
Do we need to adapt the Apache .htaccess documentation as well?

dvesh3:
> LGTM and works fine. Do we need to adapt the Apache .htaccess documentation as well?

Done.

PR #13352, opened 2022-10-11
Add missing config params in facade docs

jdreesen:
This method allows the same config params as [`\Pimcore\Navigation\Builder`](https://github.com/pimcore/pimcore/blob/10.5/lib/Navigation/Builder.php#L83), but two were missing in the docs.


aryaantony92:
@jdreesen Thanks!

PR #13356, opened 2022-10-12
Bugfix: Fixed imageGallery mandatory check if at least one image is n…

### LINKED ISSUE #13355: [Bug] [DataObject]: ImageGallery mandatory check
### Expected behavior

Validation of ImageGallery should be successful if at least 1 image was uploaded.

### Actual behavior

Validation fails if the first Image is null and at least one of rest images is not null.

### Steps to reproduce

1. Add an imageGallery field in your class definition and set mandatory checkbox to `true`.
2. Update your image gallery: the first image should be null and the second one not (example screenshot)
Bildschirmfoto 2022-10-12 um 10 57 29
3. Save the object.

MyZik:
Resolves #13355


mcop1:
Thank you!

PR #13360, opened 2022-10-12
Define MySQL server and port

agendartobias:
Resolves #

### DIFF
--- a/doc/Development_Documentation/01_Getting_Started/01_Advanced_Installation_Topics.md
+++ b/doc/Development_Documentation/01_Getting_Started/01_Advanced_Installation_Topics.md
@@ -5,8 +5,9 @@ To fully automate the installation process, options can be passed in the CLI as 
 The `--no-interaction` flag will prevent any interactive prompts:
 
 ```
-./vendor/bin/pimcore-install --admin-username admin --admin-password admin \
-  --mysql-username username --mysql-password password --mysql-database pimcore --mysql-host-socket=MYSQL-HOST-SOCKET --mysql-port=MYSQL-PORT \
+./vendor/bin/pimcore-install --admin-username=admin --admin-password=admin \
+  --mysql-username=username --mysql-password=password --mysql-database=pimcore \
+  --mysql-host-socket=127.0.0.1 --mysql-port=3306 \
   --no-interaction
 ```
 
@@ -15,8 +16,8 @@ variable. See `./vendor/bin/pimcore-install` for details. Example:
 
 ```
 $ PIMCORE_INSTALL_MYSQL_USERNAME=username PIMCORE_INSTALL_MYSQL_PASSWORD=password ./vendor/bin/pimcore-install \
-  --admin-username admin --admin-password admin \
-  --mysql-database pimcore \
+  --admin-username=admin --admin-password=admin \
+  --mysql-database=pimcore \
   --no-interaction
 ```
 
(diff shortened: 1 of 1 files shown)


doc/Development_Documentation/01_Getting_Started/01_Advanced_Installation_Topics.md
```diff
@@ -6,7 +6,7 @@ The `--no-interaction` flag will prevent any interactive prompts:
 
 ```
 ./vendor/bin/pimcore-install --admin-username admin --admin-password admin \
-  --mysql-username username --mysql-password password --mysql-database pimcore \
+  --mysql-username username --mysql-password password --mysql-database pimcore --mysql-host-socket=MYSQL-HOST-SOCKET --mysql-port=MYSQL-PORT \
```
jdreesen: I think this should be consistent, so either: 
```suggestion
./vendor/bin/pimcore-install --admin-username admin --admin-password admin \
  --mysql-username username --mysql-password password --mysql-database pimcore \
  --mysql-host-socket 127.0.0.1 --mysql-port 3306 \
```

or: 
```suggestion
./vendor/bin/pimcore-install --admin-username=admin --admin-password=admin \
  --mysql-username=username --mysql-password=password --mysql-database=pimcore \
  --mysql-host-socket=127.0.0.1 --mysql-port=3306 \
```

dvesh3:
@agendartobias thanks for your PR 😊 could you please sign the CLA?

brusch:
@agendartobias can you also rebase onto `10.5` please? Thanks 😊

kingjia90:
@agendartobias any chance to fix the branch base? if you are facing difficulties when rebasing, i'd suggest to open a new PR since it's just a one-liner, but please also consider the feedback from @jdreesen about the format consistency

doc/Development_Documentation/01_Getting_Started/01_Advanced_Installation_Topics.md
kingjia90: @agendartobias I hope you don't mind that i have fixed the branch and force pushed. Switching the base branch on github wasn't enough.

We can now focus on this suggestion again 😄 


## consistency
I am fine with the `=` , at least, it will match with the suggested command line in the demo readme
https://github.com/pimcore/demo/blob/686e91e30c430d54487a66c8558cb6ea5c2051b5/README.md?plain=1#L41
but then the following example line need also the same changes
https://github.com/pimcore/pimcore/blob/e82da03347abab204d332088cfa2b82207b63929/doc/Development_Documentation/01_Getting_Started/01_Advanced_Installation_Topics.md?plain=1#L18-L19



## env variable

I was confused at first on why it was proposed to `127.0.0.1` but after a second thought, i am also in favor of this change, it makes sense to have the hardcoded value like the other values instead of a custom(?) `env` as placeholder.
Also, if someone is planning to use an env variable, probably it would be just cleaner to directly set and use `PIMCORE_INSTALL_MYSQL_HOST_SOCKET` , so that they can even avoid adding `--mysql-host-socket=MYSQL-HOST-SOCKET` to the command line.
https://github.com/pimcore/pimcore/blob/e82da03347abab204d332088cfa2b82207b63929/doc/Development_Documentation/01_Getting_Started/01_Advanced_Installation_Topics.md?plain=1#L13-L14
https://github.com/pimcore/pimcore/blob/e82da03347abab204d332088cfa2b82207b63929/bundles/InstallBundle/src/Command/InstallCommand.php#L129

kingjia90: couldn't apply the suggestions it from github, so i've just commited with the co-author for the credits
image

PR #13361, opened 2022-10-13
[Classification store search window] Support finding group collection / groups / keys by translation

BlackbitDevs:
Steps to reproduce:

1. Create classification store
2. Create classification store group collection "Group Collection ABC"
3. Create classification store group "Group ABC", assign it to "Group Collection ABC"
4. Create classification store key / field "Field ABC", assign it to "Field ABC"
5. In translations for domain `admin` add the following translations:
    - key: Group Collection ABC, translation DE: Gruppensammlung ABC
    - key: Group ABC, translation DE: Gruppe ABC
    - key: Field ABC, translation DE: Feld ABC
6. Create class with classification store container, assign the just created store
7. Change user language to DE
8. Create object, click the "+" icon to search for classification store items
9. Search for "gruppe" -> nothing found (although we initially saw an item with name=Gruppe ABC)
10. Search for "gruppensammlung" -> same problem
11. Search for "feld" -> same problem

When you have a classification store with thousands of groups, e.g. for ETIM classification it is very difficult for users to find a certain group.

With this PR also the translated names and descriptions of the group collection / groups / keys get searched. This way all of above search terms will return the corresponding items.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/ClassificationstoreController.php
@@ -21,6 +21,10 @@ use Pimcore\Db;
 use Pimcore\Model\DataObject;
 use Pimcore\Model\DataObject\ClassDefinition\Data\LayoutDefinitionEnrichmentInterface;
 use Pimcore\Model\DataObject\Classificationstore;
+use Pimcore\Model\Translation;
+use Pimcore\Model\Translation\Listing;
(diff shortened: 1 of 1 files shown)


brusch:
@BlackbitDevs Thanks a lot! 👍 Seems like quite some code could be shared in a `private` method to avoid duplicate code? WDYT?

BlackbitDevs:
@brusch Fixed.

robertSt7:

Issue #13344, opened 2022-10-10
[Bug]: Why is `QueryBuilderHelperTrait` marked as `@internal`?


jdreesen:
### Expected behavior

The docs suggest using the trait here: https://pimcore.com/docs/pimcore/current/Development_Documentation/Extending_Pimcore/Custom_Persistent_Models.html#page_Listing-Dao

So I'd assume it isn't `@internal`.

### Actual behavior

The trait is marked as `@internal`.

### Steps to reproduce

Try to create a new listing DAO according to the docs.

dvesh3:
Fixed by #13364

PR #13363, opened 2022-10-13
Remove `internal` tag from PimcoreEcommerceFrameworkBundle

NiklasBr:
Resolves issues with documentation, deprecations, and Symfony compatibility. This discrepancy between the docs and implementation is sort of similar to #13344.

The Pimcore docs says:

> To load a bundle with the application, it must first be enabled in `config/bundles.php` (see [bundles documentation](https://symfony.com/doc/5.2/bundles.html)).

But doing it the way the Symfony docs suggests (using `::class`) in the array keys will trigger a warning not to use `@internal` classes. 

Because there is no documentation as to why PimcoreEcommerceFrameworkBundle is tagged as `@internal` I guess it was unintentionally made?

Without this change this bundle is the odd one
image


jdreesen:
Right, it makes no sense tagging a `*Bundle` class as `@internal` because it obviously has to be used by the apps to enable the bundle. Maybe make it `final` instead.

dvesh3:
@NiklasBr yes, it was not intentionally marked as internal. thanks for the change!

PR #13297, opened 2022-10-03
[Improvement]: Move configurations to AdminBundle which are UI only

kingjia90:
## Changes in this pull request  
Resolves #13054

## Additional info  
left out authenticator related as it would covered by other PR

PR #13367, opened 2022-10-13
[Security] Make new authenticator default and remove support for old authentication system

### LINKED ISSUE #12896: [Security] Make New Authenticator default and remove guard authentication
### Feature description

- [x] Remove Password encoders
- [x] Remove BruteforceProtectionHandler
- [x] Remove Guard Authentication
- [x] Remove deprecated Authentication classes

dvesh3:
Resolves #12896


doc/Development_Documentation/19_Development_Tools_and_Details/10_Security_Authentication/01_Authenticate_Pimcore_Objects.md
```diff
@@ -169,8 +135,8 @@ pimcore:
             App\Model\DataObject\User: website_demo.security.password_hasher_factory
 ```
 
-When an encoder is loaded for a `App\Model\DataObject\User` object, the UserAwareEncoderFactory will build a dedicated
-instance of `PasswordFieldEncoder` instead of always returning the same instance for all users.
+When a password hasher is loaded for a `App\Model\DataObject\User` object, the UserAwareHasherFactory will build a dedicated
```
kingjia90:
```suggestion
When a password hasher is loaded for a `App\Model\DataObject\User` object, the UserAwarePasswordHasherFactory will build a dedicated
```

bundles/CoreBundle/config/services.yaml
```diff
@@ -7,23 +7,6 @@ services:
...
-    pimcore.security.encoder_factory:
-        class: Pimcore\Security\Encoder\EncoderFactory
-        decorates: security.encoder_factory
-        arguments: ['@pimcore.security.encoder_factory.inner', '@?']
-        public: false
-        deprecated:
-            message: 'The "%service_id%" service is deprecated, use "pimcore.security.user_password_hasher" instead.'
-            package: pimcore/pimcore
-            version: 10.1
-
```
kingjia90, line 26: not this line but there some change needed on the comment  for example
`The second argument (factory mapping) is read from config (pimcore.security.encoder_factories) and set by the extension.`
that need to be changed on password_hasher_factory (former line 34, current 17)

bundles/AdminBundle/src/EventListener/TwoFactorListener.php
```diff
@@ -51,7 +50,6 @@ public function __construct(TwoFactorProviderRegistry $providerRegistry, Prepara
     public function onAuthenticationComplete(TwoFactorAuthenticationEvent $event)
     {
         // this session flag is set in \Pimcore\Bundle\AdminBundle\Security\AdminAuthenticator
```
kingjia90:
```suggestion
        // this session flag is set in \Pimcore\Bundle\AdminBundle\Security\Authenticator\AdminAbstractAuthenticator
```
but wondering are we able to solve this TODO about changing the session to be less hacky?

doc/Development_Documentation/19_Development_Tools_and_Details/10_Security_Authentication/01_Authenticate_Pimcore_Objects.md
```diff
@@ -93,48 +93,24 @@ For more information see [How to Create a custom User Provider](https://symfony.
...
 
 To be able to integrate our user object, we need 2 integration points:
 
-* A `PasswordFieldhasher` (former. `PasswordFieldEncoder`) which has access to the user instance and delegates calculation and verification of the password
-  hash to the password field definition. The encoder needs to be configured with the name of the field it should operate 
+* A `PasswordFieldHasher` which has access to the user instance and delegates calculation and verification of the password
+  hash to the password field definition. The password hasher needs to be configured with the name of the field it should operate 
   on (`password` in our case).
-* A `UserAwareHasherFactory` (former. `UserAwareEncoderFactory`) which builds a dedicated instance of a `PasswordFieldhasher` per user object.
+* A `UserAwareHasherFactory` which builds a dedicated instance of a `PasswordFieldHasher` per user object.
```
kingjia90:
```suggestion
* A `UserAwarePasswordHasherFactory` which builds a dedicated instance of a `PasswordFieldHasher` per user object.
```

doc/Development_Documentation/19_Development_Tools_and_Details/10_Security_Authentication/01_Authenticate_Pimcore_Objects.md
```diff
@@ -93,48 +93,24 @@ For more information see [How to Create a custom User Provider](https://symfony.
...
-        arguments:
-            - Pimcore\Security\Encoder\PasswordFieldEncoder
-            - ['password']
-```
-
-new way:
+To achieve this, we define a factory service which builds `PasswordFieldHasher` instances as specified above:
 ```yaml
 # The password hasher factory is responsible for verifying the password hash for a given user. As we need some special
 # handling to be able to work with the password field, we use the UserAwareHasherFactory to build a dedicated
```
kingjia90:
```suggestion
# handling to be able to work with the password field, we use the UserAwarePasswordHasherFactory to build a dedicated
```

kingjia90 (changes requested):
Just found some thing to change in docs and comments.
Also reminding that we should move the configs to AdminBundle, which are not covered by https://github.com/pimcore/pimcore/pull/13297

bundles/AdminBundle/src/EventListener/TwoFactorListener.php
dvesh3: I can't think of a better solution than using a session attribute at the moment.

kingjia90: ah ok, then no problem, i also don't have a better idea.
Then it just missing this change in the comment and the rest looks good to me, ready to be merged

kingjia90 (approved):

PR #13368, opened 2022-10-13
Remove checking of non-existing request attribute

jdreesen:
This is a remainder of the `EventedControllerInterface` from Pimcore 6. It was replaced by `KernelControllerEventInterface` and `KernelResponseEventInterface` but this code part was apparently forgotten to remove.

---

By the way... are you sure the `KernelResponseEventInterface` is working correctly? As far as I can see it only gets called, when the controller also implements the `KernelControllerEventInterface` (which e.g. the `Pimcore\Bundle\AdminBundle\Controller\Admin\IndexController` does not), because only then the `_event_controller` request attribute gets set.

/edit: see #13369

### DIFF
--- a/bundles/CoreBundle/EventListener/EventedControllerListener.php
+++ b/bundles/CoreBundle/EventListener/EventedControllerListener.php
@@ -63,12 +63,7 @@ class EventedControllerListener implements EventSubscriberInterface
     public function onKernelResponse(ResponseEvent $event)
     {
         $request = $event->getRequest();
-        $eventedController = $request->attributes->get('_evented_controller');
         $eventController = $request->attributes->get('_event_controller');
-
-        if (!$eventedController && !$eventController) {
-            return;
-        }
 
         if ($eventController instanceof KernelResponseEventInterface) {
             $eventController->onKernelResponseEvent($event);


jdreesen:
Test failure is not related. It says: 
> Error starting userland proxy: listen tcp4 0.0.0.0:33006: bind: address already in use

dvesh3:
@jdreesen yes, `KernelResponseEventInterface` implementation should not be dependent on `KernelControllerEventInterface` like you pointed out in `indexController.php`. 

Should we also get rid of `_event_controller` request attribute as this is not used anymore to check the dependency between the two interfaces. Maybe reasonable to keep it for BC reasons but then i would add a note to remove it later. wdyt?

jdreesen:
I don't think the `ResponseEvent` has a controller attached, so we probably still need the attribute.

/edit: Or maybe it has, because Symfony adds an attribute for the controller to the request? In this case, we probably can use that and don't need our own.

/edit2: but it should probably be done in #13369, as this PR is just some cleanup of forgotten code.

dvesh3:
> because Symfony adds an attribute for the controller to the request?

@jdreesen I don't know of any attribute in the request. However, I would like to keep `_event_controller` for BC reasons.

dvesh3 (approved):

dvesh3:
@jdreesen thanks!

PR #13369, opened 2022-10-13
Fix usage of `KernelResponseEventInterface` without `KernelControllerEventInterface`

jdreesen:
Purely from looking at the code, I don't think that `KernelResponseEventInterface` is useable without using `KernelControllerEventInterface` too, because `_event_controller` only gets set when the controller implements `KernelControllerEventInterface`, and will be `null`  in `onKernelResponse()` otherwise.

> **Note**
> There's #13368 which cleans `onKernelResponse()` up.


dvesh3:
@jdreesen thanks!

PR #13370, opened 2022-10-13
fix: mailto not working anymore when editing links

passioneight:
We had a project on `10.5.3` and updated to `10.5.7`, in which https://github.com/pimcore/pimcore/pull/13172 was introduced. This resulted in our client creating a new ticket, because prio to the mentioned PR, it was possible to have `mailto`'s in the `Path` field. Now, with the mentioned PR merged, `https://` will get prepended.

This PR removes all of the magic, ultimately, checking if the `Path` value is _not_, in fact, a _path_ to set the required values.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
@@ -1790,17 +1790,9 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
         listeners: {
             keyup: function (el) {
                 const value = el.getValue();
+                const pathRegex = new RegExp('^((\\/)|((\\/[^\\/]+)+(\\/)?))$');
 
-                // if it doesn't start with a single "/", we assume it's an external link
-                if (value && !value.startsWith('/') && !'http://'.startsWith(value) && !'https://'.startsWith(value)) {
-                    if (!value.match(/^https?:\/\//)) {
-                        el.setValue("https://" + value.replace(/^\/+/, ''));
-                    }
-                    internalTypeField.setValue(null);
-                    linkTypeField.setValue("direct");
-                }
-                // if it starts with "//", we assume it is a protocol-relative URL
-                else if (value.startsWith('//')) {
+                if(value && !value.match(pathRegex)) {
                     internalTypeField.setValue(null);
                     linkTypeField.setValue("direct");
                 }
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
```diff
@@ -1790,17 +1790,9 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
         listeners: {
             keyup: function (el) {
                 const value = el.getValue();
+                const pathRegex = new RegExp('^((\\/)|((\\/[^\\/]+)+(\\/)?))$');
```
jdreesen: I don't think the escaping of the slashes is necessary.

```suggestion
                const pathRegex = new RegExp('^((/)|((/[^/]+)+(/)?))$');
```

jdreesen: Also, I don't get why you defined so many groups. It should be fine like this:

```suggestion
                const pathRegex = new RegExp('^(/|(/[^/]+)+/?)$');
```

passioneight: Ah yes, that was something PhpStorm did after copy and pasting the pattern. Nicely spotted, thanks!

bundles/AdminBundle/Resources/public/js/pimcore/helpers.js
```diff
@@ -1790,17 +1790,9 @@ pimcore.helpers.editmode.openLinkEditPanel = function (data, callback) {
         listeners: {
             keyup: function (el) {
                 const value = el.getValue();
+                const pathRegex = new RegExp('^((/)|((/[^/]+)+(/)?))$');
```
jdreesen: Are those groups really necessary? I don't thing we need them, because we don't use them.

```suggestion
                const pathRegex = new RegExp('^(/|(/[^/]+)+/?)$');
```

passioneight: @jdreesen sorry, didn't see the second comment before. Had an issue without the groups on https://regex101.com/ while testing. Couldn't figured out what's wrong, so I left it as is. Updated as per your suggestion :-)

kingjia90:
Thank you, this fixes the problem introduced by that PR, but the initial purpose to prepend `http://` was (likely) due the fact that if one types path eg. `pimcore.com`, the href would be relative as `localhost/en/pimcore.com`. 

This prepend doesn't necessarily need to be added in the typing and saving phase, but in the output: if the url is not internal and has no any colon symbol (the protocol part), it should then prepend the default `https://`? wdyt?

kingjia90:
Just a gentle reminder 🏓  hopefully we have a better solution before this week's release

kingjia90:
Update: merging this now so that regression would be fixed in today's bug fix release, in the meantime i will be thinking on how to improve that relative path in a new PR or open a related issue.

PR #13376, opened 2022-10-14
Fix explanation in availability system configuration

danrot:
I guess the comment is wrong, therefore I changed it in a way that would make sense for me.

  
  

Resolves #


brusch:
Thank you very much! 😊

brusch:
🍒-picked into `10.5` 4d05802e97f5f60182442c88356d01d3b22d1905

PR #13379, opened 2022-10-14
Workflow fix `docs/` to `doc/` for included and ignored paths

NiklasBr:

I think there was an error in previous attempt. Correct path should be [`/pimcore/pimcore/tree/11.x/doc`](https://github.com/pimcore/pimcore/tree/11.x/doc) 😄


brusch:
Totally makes sense, thanks a lot 😊

brusch:
🍒-picked into `10.5` as well 😊 654c47074258537372cdb931cb19209f1972d193

PR #13382, opened 2022-10-14
[Elements] Filter all control/unassigned and invalid characters as well as line breaks and tabs from keys

brusch:


models/Element/Service.php
```diff
@@ -1112,11 +1112,16 @@ public static function getValidKey($key, $type)
...
 
+        // replace all control/unassigned and invalid characters
+        $key = preg_replace('/[^\PCc^\PCn^\PCs]/u', '', $key);
         // replace all 4 byte unicode characters
         $key = preg_replace('/[\x{10000}-\x{10FFFF}]/u', '-', $key);
         // replace slashes with a hyphen
         $key = str_replace('/', '-', $key);
 
+        // replace some other special characters
+        $key = preg_replace('/[\t\n\r]/', '', $key);
```
jdreesen: What about the other whitespace characters `\f` (form-feed) and `\v` (vertical whitespace)

brusch: 👍 added - thanks for the hint!

brusch:
@dvesh3 rebased onto `11.x`

jdreesen:
I don't think this is right.

- the regex `/[^\PCc^\PCn^\PCs]/u` doesn't look right:
  - `[^` means: not one of the following, repeating the `^` doesn't make sense except you wanna match it
  - I don't think `\PCc` (etc.) works. I think it's `\PC` followed by `c`. It should be `\P{Cc}` instead.

  So this regex **should** match everything that's **not** a `\P{Cc}`, `\P{Cf}`, `\P{Co}`, `\P{Cs}`, `\P{Cn}` (that's all `\PC`'s, which is three times in the list), or `^` (which is two times in the list) or `c`, `n`, `s` and replace it with nothing.
But, as far as I can see, it's not doing that. It seems that there's some kind of "security mechanism" which reads this as just `[^\PC]` - which is good, but still means it matches more than expected (which is probably also good, because I'm not sure why we *don't* wanna remove the "invisible formatting indicators" (`\p{Cf}`) or "code points reserved for private use" (`\p{Co}`)).

- `\t\n\r\f\v` (or `\s` in short) should not be necessary, because they're already included in the first regex (they're "control" characters)

I'll provide a PR that fixes this.

PR #13384, opened 2022-10-14
[Fix] getting value.value if value is an object

### LINKED ISSUE #13341: [Bug]: country not displayed in advanced many to many object relation column
### Expected behavior

Display the country when a country select field is configured as visible field on an advanced many to many object relation.

### Actual behavior

The country is not displayed.

### Steps to reproduce

- Go to the [demo environment](https://demo.pimcore.com/admin/) and open the class editor.
- Add a Country class (identifier 'country') and import [class_country_export.json.zip](https://github.com/pimcore/pimcore/files/9745559/class_country_export.json.zip).
- Add a Test class (identifier 'test') and import [class_test_export.json.zip](https://github.com/pimcore/pimcore/files/9745564/class_test_export.json.zip).

- Edit the 'countries' field.
- Select 'Country' as 'Allowed Class' (doesn't appear to be imported correctly).
- Save the class.

Test_countries_allowed_class_empty

- Create a few Country data objects.

Country_AT
Country_NL

- Create a Test data object.
- Drag and drop the Country data objects into the countries relation field.
- Save and publish.
- Reload the data object.

Test_countries

mattamon:
Adding additional object check
Resolves #13341 

## Additional info  
value.value returns undefined if the original value is not an object. e.g. when loading an advanced many-to-many object relation, the value is the countryCode itself.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/country.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/country.js
@@ -24,7 +24,7 @@ pimcore.object.tags.country = Class.create(pimcore.object.tags.select, {
 
     getGridColumnConfig:function (field) {
         var renderer = function (key, value, metaData, record) {
-            if (value) {
+            if (value && Ext.isObject(value)) {
                 value = value.value;
             }
             this.applyPermissionStyle(key, value, metaData, record);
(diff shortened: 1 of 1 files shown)


robertSt7:
@mattamon Thanks a lot

PR #13394, opened 2022-10-17
[Docs] Performance: Mention command pimcore:cache:warming

BlackbitDevs:
In some situations it might be a good idea to warm-up the cache to get good performance even on first access of elements in Pimcore admin area.


dvesh3:
@BlackbitDevs yes, good point. thanks!

PR #13398, opened 2022-10-17
Workaround for ImageMagick >= 7.1.0-32 with transparent images/SVG

### LINKED ISSUE #13167: [Bug]: svg with transparent background becomes black background
### Expected behavior

background remains transparent

### Actual behavior

background becomes black

### Steps to reproduce

Upload a .svg image with transparent background (or without background at all)
Show the preview in admin mode or in frontend (with defined thumbnail)

Didn't found out where exactly the bug is, but with an older (about 3 month ago / Image) pimcore/pimcore:PHP8.1 image it worked as expected. The only difference I found is that on the old image Inkscape was NOT installed.
ImageMagick Version of the old (working) image was:
Version: ImageMagick 7.1.0-32 Q16-HDRI x86_64 20006

brusch:
fixes #13167

### DIFF
--- a/lib/Image/Adapter/Imagick.php
+++ b/lib/Image/Adapter/Imagick.php
@@ -105,6 +105,10 @@ class Imagick extends Adapter
                     // only for vector graphics
                     // the below causes problems with PSDs when target format is PNG32 (nobody knows why ;-))
                     $i->setBackgroundColor(new \ImagickPixel('transparent'));
+                    if(version_compare($this->getImageMagickVersion(),'7.1.0-32') >= 0) {
+                        // workaround for ImageMagick >= 7.1.0-32, see https://github.com/pimcore/pimcore/issues/13167
+                        $i->setBackgroundColor(new \ImagickPixel("rgba(255, 255, 255, 0.00000001)"));
+                    }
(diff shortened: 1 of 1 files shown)


NiklasBr:
Which one is the corresponding ImageMagick bug report? [Is it this one](https://github.com/ImageMagick/ImageMagick/issues/5264)?

brusch:
I'm not sure, I tried to find it but didn't find an exact match, that's why I didn't linked one.

NiklasBr:
I think it is fixed ImageMagick `7.1.0-40` or greater but still remains in Inkscape, building Pimcore Docker without Inkscape get a better background:

image

brusch:
Hmm, very interesting 🤔 But without Inkscape we're probably getting some other troubles, because the built-in SVG support is by far not as good as Inkscape, right?

NiklasBr:
Never had any problems without it really. Not against this fix either, but it should probably be reported to the ImageMagick repo as well to be fixed there.

brusch:
@NiklasBr Thanks for the feedback! I'll check with ImageMagick and I'll also create a ticket for pimcore/docker to check whether we still need Inkscape.

brusch:
Closing this for now ... should be fixed by Imagick/Inkscape.

PR #13400, opened 2022-10-18
Bumpped 2fa-bundle version to ^6.0

### LINKED ISSUE #11662: Update scheb/2fa-bundle to version 6

aashan10:
Resolves #11662


dvesh3:
Depends on https://github.com/pimcore/pimcore/pull/13367

PR #13409, opened 2022-10-18
Replace `{@internal}` annotations with `@internal`

jdreesen:
PhpStorm (at least) doesn't recognize `{@internal}` annotations and doesn't mark usages as such.
So developers extending the classes/calling the methods may not notice it.


storbahn (approved):

aryaantony92:
@jdreesen Thank you!

Issue #12094, opened 2022-05-11
[Task]: Move specific functionalities into bundles to keep the core leaner


brusch:
| Functionality     | Package Name                    | Notes                                          | Own Git Repo / Package |  
|-------------------|---------------------------------|------------------------------------------------| ---- | 
| Custom Reports  | pimcore/custom-reports-bundle   |                                                | ❌  
| Glossary          | pimcore/glossary-bundle         |                                                | ❌ 
| SEO               | pimcore/seo-bundle              | redirects, robots.txt, SEO Editor, HTTP Errors, Sitemap | ❌ 
| ~~System Info~~   | [pimcore/system-info-bundle](https://github.com/pimcore/system-info-bundle)      | PHP info, OpCache status                       | ✅ | 
| ~~File Explorer~~     | [pimcore/file-explorer-bundle](https://github.com/pimcore/file-explorer-bundle)    |                                                | ✅ | ✅ |
| Word Export    | pimcore/word-export-bundle      |                                                |❌ | 
| ~~Personalization~~ | [pimcore/personalization-bundle](https://github.com/pimcore/personalization-bundle)  | personalization and targeting                  |✅ | 
| ~~Google Marketing~~  | [pimcore/google-marketing-bundle](https://github.com/pimcore/google-marketing-bundle) | Everything in "Marketing Settings"             | ✅ | 
| ~~Web-to-print~~      | [pimcore/web-to-print-bundle](https://github.com/pimcore/web-to-print-bundle)     |                                                | ✅ |   
| ~~Ecommerce~~         | [pimcore/ecommerce-framework-bundle](https://github.com/pimcore/ecommerce-framework-bundle)        |                                                | ✅ | 
| Static Routes | pimcore/static-routes-bundle        |                                                |❌ | 
| Simple Backend Search | pimcore/simple-backend-search-bundle        |                                                |❌ | 
| XLIFF | pimcore/xliff-bundle        |                                                |❌ | 
| Application Logger | pimcore/application-logger-bundle        |                                                |❌ | 
| ~~Admin UI Classic~~ | [pimcore/admin-ui-classic-bundle](https://github.com/pimcore/admin-ui-classic-bundle)        |                                                | 
| UUID | pimcore/uuid-bundle | | ❌ | 
| ~~Newsletter~~ | [pimcore/newsletter-bundle](https://github.com/pimcore/newsletter-bundle) | | ✅ | 

Tasks: 
[...]

PR #13418, opened 2022-10-19
Allow usage of `twig/extra-bundle`

jdreesen:
The skeleton (and demo) have the `TwigExtraBundle` as an example in their [`bundles.php`](https://github.com/pimcore/skeleton/blob/11.x/config/bundles.php). But when you install the package and enable the bundle, you get an exception, because Pimcore manually registers the `StringExtension` from the `twig/string-extra`, which makes the `TwigExtraBundle` cry:

> Unable to register extension "Twig\Extra\String\StringExtension" as it is already registered.

To be able, to use the bundle, you currently have to disable it there, like:
```yaml
twig_extra:
    # Note: this is actually enabled. But because Pimcore does it, this bundle cannot do it, too.
    string: false
```

This PR changes it, so that the `StringExtension` is only registered, when the `TwigExtraBundle` is not loaded.

> **Note**
> I would consider this a bug, but it is also a BC break unless we find a way to fix it **and** tell the users who have the configuration `twig_extra: { string: false }` to change it to `true`, because otherwise Pimcore can't use it.

### DIFF
--- a/bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
+++ b/bundles/CoreBundle/DependencyInjection/PimcoreCoreExtension.php
@@ -39,8 +39,6 @@ use Symfony\Component\DependencyInjection\Loader\YamlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\DependencyInjection\ServiceLocator;
 use Symfony\Component\HttpKernel\DependencyInjection\ConfigurableExtension;
-use Twig\Extra\TwigExtraBundle\TwigExtraBundle;
-use Twig\Extra\String\StringExtension;
 
 /**
  * @internal
@@ -156,7 +154,6 @@ final class PimcoreCoreExtension extends ConfigurableExtension implements Prepen
         $this->configureGoogleAnalyticsFallbackServiceLocator($container);
         $this->configureSitemaps($container, $config['sitemaps']);
         $this->configureGlossary($container, $config['glossary']);
-        $this->configureTemplating($container);
 
         $container->setParameter('pimcore.workflow', $config['workflows']);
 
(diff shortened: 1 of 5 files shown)


dvesh3:
If there's a workaround to make it work with config:
```
twig_extra:
    string: false
```
then we can just introduce the change for Pimcore 11 with upgrade note, so that you don't need to disable it in the config anymore.

jdreesen:
Another possibility would be to require `twig/extra-bundle` directly in Pimcore. As far as I understand it, this bundle automatically registers `twig/*-extra` packages when they are available (installed via composer).

Then Pimcore would not have to register `twig/string-extra` manually and app developers would not have to disable it in the `twig_extra` configuration.

Another advantage would be that the `TwigExtraBundle` [gives hints on what to do](https://github.com/twigphp/twig-extra-bundle/blob/2.x/MissingExtensionSuggestor.php) if you try to use some methods that are listed in the official Twig extension but are not part of the core, like [`format_date`](https://twig.symfony.com/doc/3.x/filters/format_date.html).

dvesh3:
@jdreesen Sorry, for late response. yes, I would go with your last suggestion to include `twig/extra-bundle` directly in core since it adds extra features for Twig. Could you please adapt your PR? thanks!

jdreesen:
Done. 

I'm not sure what happens if the bundle is added/enabled in the project, too. If that results in an error, I'd still consider this as breaking. Otherwise, we may move this to `v10.5`.

dvesh3:
> I'm not sure what happens if the bundle is added/enabled in the project, too. 

I don't see a problem, if this is already required on project level. I would move it to 10.6 milestone, so you can already rebase it on 10.5 branch for now (since we don't have 10.6 at the moment).

Also, could you please add upgrade notes for 10.6? 

and remove this https://github.com/pimcore/pimcore/pull/13418/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R131 since this is already required in twig-extra package. thanks!

jdreesen:
> and remove this https://github.com/pimcore/pimcore/pull/13418/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R131
since this is already required in twig-extra package. thanks!

No, it isn't. `twig/extra-bundle` only does the setup for you and contains some stubs that tell you to install the packages if you try to use the filters/methods/... but you have to require the extras you want to use yourself, so we still need to require the `twig/string-extra`.

jdreesen:
I rebased it and added upgrade notes.

brusch:
Basically fine for me, but `twig/string-extra` will be removed from the core in v11, because it's only used by the targeting functionality which will be extracted into a bundle, so it shouldn't be a core requirement anymore. 

See also: #12094

brusch:
So can be merged into `10.6` as soon as it's available.

PR #13420, opened 2022-10-20
Add translation for "no_further_classes_allowed"

BlackbitDevs:
Is used in https://github.com/pimcore/pimcore/blob/0b5c3aa67ab0ce4197b89e8cdd6415c90e6196ad/bundles/AdminBundle/Resources/public/js/pimcore/object/objectbricks/field.js#L169


aryaantony92:
@BlackbitDevs Thank you!

PR #13422, opened 2022-10-20
[Bugfix]: KeyConfig cache issue

MPodravac:
There is an issue when you get KeyConfig by name, it is saved in Runtime Cache. When you do some action on that object and save it, Runtime Cache sets cache key value to null instead of deleting it. So when you want to get that KeyConfig by name again, you get null (database is not contacted again because value in cache exists, even though it is not correct).


robertSt7:
@MPodravac Thanks!

PR #13425, opened 2022-10-20
Fix realFullPath for Video Poster Images

dpfaffenbauer:
otherwise, if you have a prefix set, it will only work the first time

### DIFF
--- a/models/Document/Editable/Video.php
+++ b/models/Document/Editable/Video.php
@@ -256,7 +256,7 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
             'title'        => $this->title,
             'description'  => $this->description,
             'path'         => $path,
-            'poster'       => $poster ? $poster->getFullPath() : '',
+            'poster'       => $poster ? $poster->getRealFullPath() : '',
         ];
     }
 


brusch:
Hmm,, but what if the image is stored on a remote filesystem, where the prefix starts eg. with https://foo.bar/.... `getRealFullPath()` will always return the internal path of the asset, right?

kingjia90:
This works but may change some behaviour when using the PHP API and the result of the [getData ](https://pimcore.com/docs/pimcore/current/Development_Documentation/Documents/Editables/Video.html#page_Methods) method, where it may be more useful and consistent to return the full absolute path.

The issue seems happening because `getFullPath` prepends the prefix when calling `getFrontendFullPath` and causes the input to change going into some id dependency resolving mismatch(?) which end up turning it into `null` on save or on second reload.  
image

If the getData issue is a valid point, then i would suggest to add something like this instead:

```
    protected function getDataEditmode()
    {
        $data = $this->getData();

        $poster = Asset::getById($this->poster);
        $data['poster'] = $poster->getRealFullPath();

        if ($this->type === self::TYPE_ASSET && ($video = Asset::getById($this->id))) {
            $data['path'] = $video->getRealFullPath();
        }

        return $data;
    }
```

I've also changed the video path, which was actually covered in https://github.com/pimcore/pimcore/pull/13291 but i didn't understood the issue was related when using prefixes. 
For videos, it was a little more "permissive" and didn't clear the input (treating it as if it was an external url or so), but still may end up with the same rendering problem.

brusch:
@kingjia90 thanks for clarifying 👍 
I'd prefer your proposal with implementing `getDataEditmode()` as this is BC-safe and more specific anyway. 

Can you quickly put that in a PR? Thanks a lot!

kingjia90:
Closing in favor of #13465

Issue #189, opened 2015-07-17
Cannot filter on checkbox value "No" in Object Grid View


brusch:
There seems to be a problem when filter for checkbox values "No" in the object grid view if the checkbox value never has been set to true.
If I manually set a checkbox value to True, save the object and then set it to false again, it will appear in the filter results for "No".

I would expect filtering checkboxes for "No" values would return all objects where the checkbox is unchecked.

I was able to reproduce this behaviour on demo.pimcore.org on the "news" class, see attached images.

Regards.

![image](https://cloud.githubusercontent.com/assets/142037/8744396/2d434dbc-2c77-11e5-9de8-79717c49463e.png)

![image](https://cloud.githubusercontent.com/assets/142037/8744398/318367f4-2c77-11e5-8433-259067b58040.png)

![image](https://cloud.githubusercontent.com/assets/142037/8744399/3c6ef160-2c77-11e5-9164-f9b41bba156a.png)

![image](https://cloud.githubusercontent.com/assets/142037/8744417/605c6512-2c77-11e5-85b4-e26f508ee9e5.png)

![image](https://cloud.githubusercontent.com/assets/142037/8744425/6397f7a0-2c77-11e5-9f14-3d5d2254d9e9.png)

![image](https://cloud.githubusercontent.com/assets/142037/8744447/84799bea-2c77-11e5-8897-f22dc58db3a2.png)

ThomasKeil:
The reason is that checkboxes can have the value 0, 1 and NULL in the database, NULL if it has never been touched. You can't compare NULL to anything with SQL, except for IS NULL and IS NOT NULL, so checking for "checkbox = 0" or "checkbox != 1" won't render the expected result. 
See https://dev.mysql.com/doc/refman/5.0/en/working-with-null.html

Thus, to check this properly you'd have to use "checkbox = 0 OR checkbox IS NULL".

The proper place for this would be in pimcore/models/Object/Service.php:getFilterCondition(), although it seems to me to be a bigger change.

PR #13427, opened 2022-10-20
[Bug] Grid unable to filter by published null

kjkooistra-youwe:
When (programmatically) creating a data object without setting an explicit published value, the published value will be null. When filtering 'No' on the published column of a grid it won't return these data objects a results.

Similar to https://github.com/pimcore/pimcore/issues/189


robertSt7:
@kjkooistra-youwe Thanks for the PR

PR #13430, opened 2022-10-20
[Assets] "Apply to all" checkbox for asset upload (overwrite, keep both, skip)

### LINKED ISSUE #11315: [Improvement]: Improve dialog to overwrite existing assets
### Improvement description

Since we have an "Overwrite all" option (see #1020), the only thing missing is a similiar option for "keep both".

Maybe take a look at the screenshot mentioned here:
[#6893 (comment)](https://github.com/pimcore/pimcore/issues/6893#issuecomment-668529774)

I think it would be a good idea to remove the "overwrite all" button and add the "apply to all" checkbox to the asset dialog.

BlackbitDevs:
As proposed in https://github.com/pimcore/pimcore/issues/6893#issuecomment-668529774 this PR adjusts the behaviour when uploading files which already exist.

Снимок экрана 2022-10-20 в 18 48 18

The buttons explained:

- _Apply to all_: If checked, action from pressed button will be applied to all uploaded files.
- _Overwrite_: Use the uploaded file as data for the already existing asset
- _Keep both files_: Upload new file with suffix
- _Skip_: New uploaded file will be skipped (existing asset stays untouched)

Close window: Abort upload -> No files will be changed 
   
Resolves #11315


bundles/CoreBundle/translations/en.json
```diff
@@ -985,6 +985,8 @@
   "asset_upload_overwrite": "Overwrite",
   "asset_upload_keep_both": "Keep both files",
   "asset_upload_overwrite_all": "Overwrite all",
```
kingjia90:
```suggestion
```
this translation entry can be deleted

Corepex:
@BlackbitDevs thanks for the improvement.

Thers one small thing i found - the "apply to all" checkbox doesnt work if you choose "skip".

I would suggest to move the IF check (in line 320) a few lines down to only wrap the uploadFunction. 
That should do the job.

bundles/AdminBundle/public/js/pimcore/asset/tree.js
```diff
@@ -295,31 +295,46 @@
...
-                                 } else if (action === 'yes') {
-                                     uploadFunction(false);
-                                 } else if (action === 'no') {
-                                     Ext.each(overwriteConfirmMessageBoxes, function(messageBox) {
-                                         if (messageBox) {
-                                             messageBox.down('button[itemId=ok]').fireHandler();
-                                         }
-                                     });
-                                 } else if (action === 'cancel') {
+                                 if (action === 'ok' || action === 'yes') {
```
Corepex: @BlackbitDevs thanks for the improvement.

Thers one small thing i found - the "apply to all" checkbox doesnt work if you choose "skip".

I would suggest to move the IF check a few lines down to only wrap the uploadFunction.
That should do the job.

Corepex:
@BlackbitDevs are there any updates according to my comment?

bundles/AdminBundle/public/js/pimcore/asset/tree.js
BlackbitDevs: You are right, I applied your suggested solution.

BlackbitDevs:
Hi @Corepex, thx for your comment. Have applied your suggested solution. Now also works with "Skip".

bundles/CoreBundle/Resources/translations/en.json
```diff
@@ -992,7 +992,8 @@
     "asset_upload_want_to_overwrite": "Do you want to overwrite file %s?",
     "asset_upload_overwrite": "Overwrite",
     "asset_upload_keep_both": "Keep both files",
-    "asset_upload_overwrite_all": "Overwrite all",
+    "asset_upload_apply_to_all": "Apply to all",
+    "asset_upload_skip": "Skip",
```
Corepex: could you please move these translations to `CoreBundle/translations/en.json` since `CoreBundle/Resources/translations/en.json` will be removed soon. 

thanks

BlackbitDevs: @Corepex Done.

Corepex:
@BlackbitDevs, thanks for your PR. The new "apply to all" checkbox truly improves pimcore 🥇 💯

PR #13434, opened 2022-10-21
[Improvement] Pass additionalData to transition events

MarcoGuiducci:
If we pass additionalData to the workflow "apply" method, those data are passed to the various transition and place events in the "context" variable (as it is just working now for the "GlobalActionEvent") that is currently unused.

This is necessary to have the possibility to use additionalData in the various workflow events.

The $additionalData variable is passed to the "apply" method of the workflow as third argument, which is the "context"
On cascade, the context will be passed to all the workflow events (e.g the TransitionEvent) so that it's possible to use the additionalData selected by the user in those events to guide some decisions

### DIFF
--- a/doc/Development_Documentation/07_Workflow_Management/09_Working_with_PHP_API.md
+++ b/doc/Development_Documentation/07_Workflow_Management/09_Working_with_PHP_API.md
@@ -66,3 +66,70 @@ In addition to the Symfony events, Pimcore provides two additional events for gl
 - `pimcore.workflow.postGlobalAction`
 See [WorkflowEvents](https://github.com/pimcore/pimcore/blob/10.x/lib/Event/WorkflowEvents.php) for details. 
 
+### Using Additional Data in Events
+
+If additional data fields are defined in the transition configuration, it's possible to retrieve those data on event listener functions.
+
+See following example how to interact with additional data on transition events.
+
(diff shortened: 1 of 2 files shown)


robertSt7:
@MarcoGuiducci Could you please add some documentation and rebase the PR to 11.x? Thanks

MarcoGuiducci:
@robertSt7 sorry but it's just the second time that I open a PR on Pimcore... what do you mean for "add some documentation"? Where should I add that documentation? In this PR or in the development documentation files?

robertSt7:
@MarcoGuiducci No problem, I think it would be good to have a short description (as you have it in the PR description) in this file: https://pimcore.com/docs/pimcore/current/Development_Documentation/Workflow_Management/Working_with_PHP_API.html  and perhaps you could also insert an example there.

MarcoGuiducci:
@robertSt7 I've added an example in the documentation

robertSt7:
@MarcoGuiducci Thanks, could you also please rebase the PR to 11.x, because we merge improvements into the 11.x branch. Thank you

MarcoGuiducci:
@robertSt7 done ;) Will this change also be merged also on the 10.5 branch?

robertSt7:
@MarcoGuiducci Thanks for the PR :) Unfortunately, since this is an improvement, we will not merge it to 10.5.

PR #13442, opened 2022-10-24
FullPageCacheListener kernel.request handler needs to be executed aft…

brusch:
…er session was initialized

Follow-up to #12294


paulhaem:
@brusch Is there any special reason this as bug labled PR was merged into 11.x instead of 10.x? We recently ran into the same issue where pages that have been cached using the FullPageCache are not considering the active session. Only pages that have not been cached yet are rendered correctly with `X-Pimcore-Output-Cache-Disable-Reason: Session in use`.

Using latest Pimcore version 10.5.19.

PR #13443, opened 2022-10-24
Session: removed deprecated service fallback (session without request)

brusch:


brusch:
Closed in favor of #13475

PR #13446, opened 2022-10-25
Update 07_Logging.md

kaurov:
\Pimcore\Log\Simple is marked as @internal so this sample need to be replaced with updated version

  
  

Resolves #


dvesh3:
we need to update the example with Monolog implementation.

aashan10:
Appropriate example has been added in https://github.com/pimcore/pimcore/pull/13607. This pr can be closed.

PR #13449, opened 2022-10-25
[Improvement] Detect focal point when detecting faces

### LINKED ISSUE #12847: [Improvement]: Detect focal point when detecting faces
### Improvement description

When using the Detect Image Features feature on an image asset, would it make sense to set the focal point at the same time?

At the moment, AssetController only calls detectFaces() in the detectImageFeaturesAction() method. Adding in a call to detectFocalPoint() here would automatically set the focal point based on the detected faces.

mattamon:
Adding detectFocalPoint() after detectFaces() in detectImageFeaturesAction
Resolves #12847 

## Additional info  
Adding in a call to detectFocalPoint() here would automatically set the focal point based on the detected faces.


dvesh3:
@mattamon looks good. thanks!

PR #13452, opened 2022-10-25
fix: include documents without navigation name as invisible in navigation builder

youwe-petervanderwal:
Error: when overriding the navigation label via a pageCallback, documents without a configured navigation_name were still excluded from the navigation
Resolved by: including all documents, but marking the ones without a navigation_name as invisible by default

## Background info  
We used the NavigationBuilder to create a secondary menu/sitemap where the customer wanted to have a different label being displayed. Therefore we used the pageCallback, but documents/links without a configured navigation_name were still missing from the result.

Moving this "do I have a label?" check to the `setVisible(false)` gives the developer more flexibility using the pageCallback.

### DIFF
--- a/lib/Navigation/Builder.php
+++ b/lib/Navigation/Builder.php
@@ -407,7 +407,7 @@ class Builder
                 continue;
             }
 
-            if (($child instanceof Document\Folder || $child instanceof Document\Page || $child instanceof Document\Link) && $child->getProperty('navigation_name')) {
+            if ($child instanceof Document\Folder || $child instanceof Document\Page || $child instanceof Document\Link) {
                 $path = $child->getFullPath();
                 if ($child instanceof Document\Link) {
                     $path = $child->getHref();
@@ -429,7 +429,7 @@ class Builder
                 $page->setRelation($child->getProperty('navigation_relation'));
                 $page->setDocument($child);
 
-                if ($child->getProperty('navigation_exclude') || !$child->getPublished()) {
+                if (trim((string)$child->getProperty('navigation_name')) === '' || $child->getProperty('navigation_exclude') || !$child->getPublished()) {
                     $page->setVisible(false);
                 }
 


lib/Navigation/Builder.php
```diff
@@ -407,7 +407,7 @@ protected function buildNextLevel($parentDocument, $isRoot = false, $pageCallbac
                 continue;
             }
 
-            if (($child instanceof Document\Folder || $child instanceof Document\Page || $child instanceof Document\Link) && $child->getProperty('navigation_name')) {
+            if (($child instanceof Document\Folder || $child instanceof Document\Page || $child instanceof Document\Link)) {
```
youwe-petervanderwal: I just see a `if ((` here which could just be a single `if (` - fixing that

youwe-petervanderwal:
Can this PR have a review? I'd like to be able to inform our customer whether this is going to be resolved in a future update or that we need to change the implementation within the application.

lib/Navigation/Builder.php
```diff
@@ -429,7 +429,7 @@ protected function buildNextLevel($parentDocument, $isRoot = false, $pageCallbac
                 $page->setRelation($child->getProperty('navigation_relation'));
                 $page->setDocument($child);
 
-                if ($child->getProperty('navigation_exclude') || !$child->getPublished()) {
+                if (trim((string)$child->getProperty('navigation_name')) === '' || $child->getProperty('navigation_exclude') || !$child->getPublished()) {
```
NiklasBr, line 432:
```suggestion
                if (empty($child->getProperty('navigation_name')) || $child->getProperty('navigation_exclude') || !$child->getPublished()) {
```

[I believe `empty()` would work here](https://www.virendrachandak.com/techtalk/php-isset-vs-empty-vs-is_null/), no need to coerce the value to a string first.

youwe-petervanderwal: I always hate PHP's `empty()` treating the string value `"0"` as empty as well, so I always try to keep my conditions to a strict empty string. Although the chance someone configures the navigation label to `"0"` is pretty slim, I think it's (from a users point of view) unexpected that setting that label would hide it from the menu.

NiklasBr: What? I did not know about that quirk. 😠

youwe-petervanderwal: Yep, it's actually just [casting to boolean](https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting) and there it makes a bit more sense again to treat `"0"` as `false` (and `"1"` as `true`). But from a string perspective the function name `empty` is a bit unfortunate...

aryaantony92:
@youwe-petervanderwal Thank you for the PR!

PR #13457, opened 2022-10-26
Fix notification filtering

MarcoGuiducci:
Filtering of notifications by "title" don't work properly. The query don't provide any result because the condition is wrong. Fixed the parseString method of NotificationServiceFilterParser class to let the notification filtering works well

### DIFF
--- a/models/Notification/Service/NotificationServiceFilterParser.php
+++ b/models/Notification/Service/NotificationServiceFilterParser.php
@@ -110,7 +110,7 @@ class NotificationServiceFilterParser
     {
         $result = null;
         $property = $this->getDbProperty($item);
-        $value = $item[self::KEY_VALUE];
+        $value = array_key_exists(self::KEY_VALUE, $item) ? $item[self::KEY_VALUE] : "";
 
         switch ($item[self::KEY_OPERATOR]) {
             case self::OPERATOR_LIKE:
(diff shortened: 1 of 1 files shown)


mcop1:
@MarcoGuiducci I know, it is not really part of your pr, but when I activate the checkbox next to the filter, without text, I get an error message:
grafik
=>
grafik

Could you please add an check for the item in the array here https://github.com/pimcore/pimcore/pull/13457/files#diff-50f76d3198e6e87ad4707d5291996fef67e8c300a620ca31adf3fe85f098a031L113 ? 

Thank you for your efforts!

MarcoGuiducci:
@mcop1 done in this commit.
[https://github.com/pimcore/pimcore/pull/13457/commits/1f5037d74c05eed840495fc19aa53f1a8f44f31d](https://github.com/pimcore/pimcore/pull/13457/commits/1f5037d74c05eed840495fc19aa53f1a8f44f31d)

models/Notification/Service/NotificationServiceFilterParser.php
```diff
@@ -110,11 +110,11 @@ private function parseString(array $item): array
     {
         $result = null;
         $property = $this->getDbProperty($item);
-        $value = $item[self::KEY_VALUE];
+        $value = array_key_exists(self::KEY_VALUE, $item) ? $item[self::KEY_VALUE] : "";
```
jdreesen: We don't have to consider `null` here, right?
```suggestion
        $value = $item[self::KEY_VALUE] ?? '';
```

PR #13462, opened 2022-10-27
fix: always call Process::setTimeout()

dkarlovi:
Process has default timeout of 60s which can be omitted if setting it to null. Pimcore doesn't call setTimeout() unless and defaults to null. This means by default every process has a 60sec timeout and you can't unset it, you can only set it really high

### DIFF
--- a/lib/Tool/Console.php
+++ b/lib/Tool/Console.php
@@ -241,9 +241,7 @@ final class Console
         $cmd = self::buildPhpScriptCmd($script, $arguments);
         self::addLowProcessPriority($cmd);
         $process = new Process($cmd);
-        if ($timeout) {
-            $process->setTimeout($timeout);
-        }
+        $process->setTimeout($timeout);
         $process->start();
 
         if (!empty($outputFile)) {


lib/Tool/Console.php
```diff
@@ -241,9 +241,7 @@ public static function runPhpScript($script, $arguments = [], $outputFile = null
         $cmd = self::buildPhpScriptCmd($script, $arguments);
         self::addLowProcessPriority($cmd);
         $process = new Process($cmd);
-        if ($timeout) {
-            $process->setTimeout($timeout);
-        }
+        $process->setTimeout($timeout);
```
jdreesen, line 244: But now it has no timeout unless you explicitly set the parameter to something not `null`. 
Maybe the parameter's default value should be changed to `60` (as it is in the constructor of `Process`)?

dkarlovi: I'm not sure, TBH. The expected behaviour from the current signature is "no timeout", but the behaviour is "60s timeout", so either one is unexpected if you consider the other one.

dkarlovi: If we change the signature (which is a BC break), we're making the current behaviour "official". But if we keep this change as is (just the bugfix), we're making the current signature "official". IMO, the latter makes more sense since the behaviour comes from the underlying component which you don't interact with directly.

dvesh3: @dkarlovi how about we just pass the param to Process initialization and let Symfony set the value?
```
$process = new Process(command: $cmd, timeout: $timeout);
``` 
and then it will not be a BC break since we still have the value set to 60 if nothing passed in params.

dkarlovi: But current Pimcore's API timeout is `null`, meaning "no timeout". We can't keep both the behaviour and the API, one is incorrect (and IMO, it's the 60s timeout currently applied).

dvesh3: yes, I agree the API and the current behavior don't match. I would say changing behavior could be more impactful than changing the signature so i'd vote for setting the default value to 60 with upgrade notes.

dkarlovi: I'm the opposite, I'd vote to keep BC.

jdreesen: What about keeping BC in Pimcore 10 and changing it to `60` seconds in Pimcore 11? 
I think it's better to have a timeout by default and require unsetting it explicitly.

dkarlovi: I agree @jdreesen

jdreesen: Wait... currently, there already *is* a timeout of 60 seconds by default (even though the default value of the parameter suggests there is none).

So if we suddenly interpret the default value of `null` as such and *remove* the timeout, I think that would be a BC break, because it would be a change of behavior, to one that I consider dangerous (unless it is explicitly requested).

The real problem is, that it's currently not possible, to actually *remove* the timeout by passing `null`, right? So we should fix that.

dvesh3: yes, we can do something like this to fix the current behavior: 
```php
//@TODO: set default value to 60
if (array_key_exists(3, func_get_args())) {
    $process->setTimeout(func_get_arg(3));
}
```
so it is possible to remove the timeout by passing null.

dvesh3: @dkarlovi could you please adapt the PR with the fix suggested [here](https://github.com/pimcore/pimcore/pull/13462#discussion_r1015116973)? thanks!

dkarlovi: @dvesh3 that means the current API is still broken though? If I don't pass null, the timeout is 60s?

aryaantony92: Yes, but that should be fine since even in symfony, the default value is 60 s.

dkarlovi: This is Pimcore's API though, not Symfony's? It doesn't matter what Symfony's timeout is if Pimcore says `timeout=null`. Also, it's super confusing to have it behave differently if you explicitly say `timeout=null` or it defaults to `timeout=null`.

> - Hey Pimcore, run this command without timeout set, it's null, right?
> - Correct, the timeout is 60s.
> - What?

fashxp: Summing up the conversation, we vote for removing the `if` (as done already) and changing default value of `$timeout` to `60` in method signature (needs to be done). 

We don't see that as a BC break (class is final, so it cannot break any code; it is also not against [symfony BC promise](https://symfony.com/doc/current/contributing/code/bc.html)). 

We rather see it as a bugfix because:
- when value (especially `null` or `0`) for `$timeout` is provided, that value always will be set (was not possible up to now)
- when no value for `$timeout` is provided, it is 60 seconds (as it is now, so no processes are running endlessly all of a sudden)

@dkarlovi could you please update the PR accordingly, thank you.

fashxp (changes requested):

aryaantony92:
@dkarlovi Just a gentle reminder for https://github.com/pimcore/pimcore/pull/13462#discussion_r1017766386.

dkarlovi:
@aryaantony92 unfortunately I'll not have time to work on this for a bit, feel free to fix as suggested and close here, thank you. :+1:

aryaantony92:
> @aryaantony92 unfortunately I'll not have time to work on this for a bit, feel free to fix as suggested and close here, thank you. 👍

Sure, will update. Thank you 👍

dvesh3:
Closed in favor of #13624

PR #13464, opened 2022-10-27
[Fix] adding validation message for wrong upload type

### LINKED ISSUE #13000: [Bug]: Field Type Image Allowed PDF, txt etc... files
### Expected behavior

There should be a validation message.

### Actual behavior

When we use image type we can upload only images, here we can not use other files on object but when we select PDF or any other extension file into the image type field and is uploaded into the assets.

### Steps to reproduce

1. Add an image type field **[Upload Path: /Rushit]** into the class.
2. Create the object and select PDF, txt, or any other file instead of an image(png, jpeg, etc..).

->> Output is like the file that is not viewed into an object but we can see this file in the Assets.
image

mattamon:
Adding validation message for wrong upload type
Resolves #13000 

## Additional info  
See #13000 for description
The mime type wildcard check was taken from https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Validator/Constraints/FileValidator.php#L226

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -15,6 +15,7 @@
 
 namespace Pimcore\Bundle\AdminBundle\Controller\Admin\Asset;
 
+use GuzzleHttp\Psr7\MimeType;
 use Pimcore\Bundle\AdminBundle\Controller\Admin\ElementControllerBase;
 use Pimcore\Bundle\AdminBundle\Controller\Traits\AdminStyleTrait;
 use Pimcore\Bundle\AdminBundle\Controller\Traits\ApplySchedulerDataTrait;
@@ -49,6 +50,7 @@ use Symfony\Component\HttpKernel\Event\ControllerEvent;
 use Symfony\Component\Mime\MimeTypes;
 use Symfony\Component\Process\Process;
 use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Validator\Constraints\FileValidator;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
 
 /**
@@ -518,6 +520,15 @@ class AssetController extends ElementControllerBase implements KernelControllerE
             throw new \Exception('File is empty!');
         } elseif (!is_file($sourcePath)) {
             throw new \Exception('Something went wrong, please check upload_max_filesize and post_max_size in your php.ini as well as the write permissions of your temporary directories.');
+        }
+
+        if($request->get('accept')) {
+            $mimetype = MimeTypes::getDefault()->guessMimeType($sourcePath);
+            $accept = $request->get('accept');
+            $acceptedMimeTypes = explode(',', $accept);
+            if(!$this->_assetService->validateMimeType($mimetype, $acceptedMimeTypes)) {
+                throw new \Exception("Mime type does not match, accepted types: $accept");
+            }
         }
 
         if ($request->get('allowOverwrite') && Asset\Service::pathExists($parentAsset->getRealFullPath().'/'.$filename)) {
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -15,7 +15,6 @@
 
 namespace Pimcore\Bundle\AdminBundle\Controller\Admin\Asset;
 
-use GuzzleHttp\Psr7\MimeType;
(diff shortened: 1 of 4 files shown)


models/Asset/Service.php
```diff
@@ -499,4 +499,27 @@ public static function getUniqueKey($item, $nr = 0)
...
     }
+
+    /**
+     * @param string $mime
+     * @param array $acceptedMimeTypes
+     *
+     * @return bool
+     *
+     */
+    public function validateMimeType(string $mime, array $acceptedMimeTypes) {
```
blankse:
```suggestion
    public function validateMimeType(string $mime, array $acceptedMimeTypes): bool
    {
```

brusch:
@mattamon unfortunately this doesn't cover all use-cases, e.g. `.ai` files, where the mime-type doesn't start with `image/`. 
So I'd stick to the server-side only validation, but with a proper error message. 
I think we could use just https://github.com/pimcore/pimcore/blob/7668ed4814833dae9833bd86f3b29e6c1ab83179/models/Asset.php#L418 to get the asset type for the uploaded mime-type and then compare it with the current type of the asset.

mattamon:
@brusch, I totally agree with you, this does not cover all cases.

We would still need something like the accept param, since the addAsset function is used for every type of asset right?
Like we need to know from which tag e.g. tag/image, the upload was started.

brusch:
@mattamon don't think this is necessary, imho it's fine if we validate on server side and let the user pick whatever file type, because it's rather an edge case anyway that someone will try to upload a video as a replacement for an image or so.

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -520,6 +520,15 @@ protected function addAsset(Request $request, Config $config)
             throw new \Exception('Something went wrong, please check upload_max_filesize and post_max_size in your php.ini as well as the write permissions of your temporary directories.');
         }
 
+        if($request->get('accept')) {
+            $mimetype = MimeTypes::getDefault()->guessMimeType($sourcePath);
+            $accept = $request->get('accept');
+            $acceptedMimeTypes = explode(',', $accept);
+            if(!$this->_assetService->validateMimeType($mimetype, $acceptedMimeTypes)) {
```
dvesh3, line 527: @mattamon could you please adapt it as per https://github.com/pimcore/pimcore/pull/13464#issuecomment-1293313339? so we should just check the type here instead of validating mime type. I would get rid of accept property from upload dialog and just compare the type with the value from `getTypeFromMimeMapping()` on server side.

mattamon: I think the server side validation only is a bit lacking. 
As mention in the issue, it is possible to add e.g. a .docx to an image tag with no validation. The file is uploaded but will never be displayed in the object, because of the wrong type.

On the other hand, what do I compare the new file type to, since I the type is not specified anywhere?

If I upload a .jpg first and then a .docx, I don't find the file (the extension is part of the filename -> `Asset\Service::pathExists($parentAsset->getRealFullPath().'/'.$filename)`) and a new one will be created, or did I miss something?

dvesh3: @mattamon please find my answers below:
> The file is uploaded but will never be displayed in the object, because of the wrong type.

yes, we should refactor the calls here to cleanup the uploaded file even in case of exceptions https://github.com/pimcore/pimcore/blob/8d527233a22fe230d57d84fa6c7d7ae3e793cfa8/bundles/AdminBundle/src/Controller/Admin/Asset/AssetController.php#L371 

> On the other hand, what do I compare the new file type to, since I the type is not specified anywhere?

we can still pass this property (with correct value) to the server for comparison https://github.com/pimcore/pimcore/pull/13464/files#diff-fe453464481618fcfe5a3be91df93fed62f7f00b8053f32ebdd4c97dfdfbf6b2R19

> If I upload a .jpg first and then a .docx, I don't find the file (the extension is part of the filename -> Asset\Service::pathExists($parentAsset->getRealFullPath().'/'.$filename)) and a new one will be created, or did I miss something?

This is quite an edge case, if you try to upload files with same names and different extensions so i would not consider it.

The reason we'd like to just compare the type as you won't be able to validate all image types with one group (e.g. `image/*`) so it is better to use `getTypeFromMimeMapping()` for comparison.

dvesh3: @mattamon are you on it?

mattamon:
Closed in favor of #13726

PR #13475, opened 2022-10-28
Remove Session Fallback BC layer

dvesh3:
Follow up to #12746

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
@@ -95,7 +95,7 @@ Please make sure to set your preferred storage location ***before*** migration. 
 - [Codeception] Bumped `codeception/codeception` version to ^5.0. Now, Pimcore is using a new directory structure for tests (Codeception 5 directory structure). For details, please see [#13415](https://github.com/pimcore/pimcore/pull/13415)
 - [Session] 
   - `AdminSessionHandler` requires session from request stack.
-  - `EcommerceFrameworkBundle\SessionEnvironment` not loading from or storing to session in cli mode anymore.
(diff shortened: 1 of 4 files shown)


doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -93,6 +93,10 @@ Please make sure to set your preferred storage location ***before*** migration.
   - Removed commands: `pimcore:bundle:enable`, `pimcore:bundle:disable`.
   - Removed `dontCheckEnabled` config support from Areablock editable.
 - [Codeception] Bumped `codeception/codeception` version to ^5.0. Now, Pimcore is using a new directory structure for tests (Codeception 5 directory structure). For details, please see [#13415](https://github.com/pimcore/pimcore/pull/13415)
+- [Session] 
+  - `AdminSessionHandler` requires session from request stack.
+  - `EcommerceFrameworkBundle\SessionEnvironment` does not work in cli mode anymore.
```
fashxp: you still can use `SessionEnvironment` in cli mode, but it does not load/store any values from/to session - which does not make any sense in cli mode anyway. 
right? 
so 'does not work' is probably not right?

fashxp (changes requested):

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
dvesh3: I meant it does nothing in cli mode as we return the control from every method, if called from cli. I can change to 'not supported'.

fashxp: not supported is also not entirely correct, since you still can use it in cli mode too ... I changed the phrase a bit.

fashxp (approved):

PR #13478, opened 2022-10-28
Symfony 6

dvesh3:
## Changes in this pull request  
Resolves #13242
Resolves #11751

PR #13480, opened 2022-10-28
Bump webmozarts/console-parallelization version to v2

blankse:
Is needed for https://github.com/pimcore/pimcore/pull/13478

### DIFF
--- a/bundles/CoreBundle/src/Command/ThumbnailsImageCommand.php
+++ b/bundles/CoreBundle/src/Command/ThumbnailsImageCommand.php
@@ -240,8 +240,8 @@ class ThumbnailsImageCommand extends AbstractCommand
         return $thumbnailsToGenerate;
     }
 
-    protected function getItemName(int $count): string
+    protected function getItemName(?int $count): string
     {
-        return $count == 1 ? 'thumbnail' : 'thumbnails';
+        return $count === 1 ? 'thumbnail' : 'thumbnails';
     }
 }
--- a/bundles/CoreBundle/src/Command/ThumbnailsImageCommand.php
+++ b/bundles/CoreBundle/src/Command/ThumbnailsImageCommand.php
@@ -96,7 +96,7 @@ class ThumbnailsImageCommand extends AbstractCommand
         }
     }
 
-    protected function fetchItems(InputInterface $input): array
+    protected function fetchItems(InputInterface $input, OutputInterface $output): array
     {
         $list = new Asset\Listing();
 
--- a/lib/Console/Traits/Parallelization.php
+++ b/lib/Console/Traits/Parallelization.php
@@ -78,6 +78,7 @@ trait Parallelization
     ): ParallelExecutorFactory {
         return ParallelExecutorFactory::create(...func_get_args())
             ->withSegmentSize($this->getSegmentSize())
+            ->withScriptPath($this->getConsolePath())
             ->withRunBeforeFirstCommand($this->runBeforeFirstCommand(...))
             ->withRunAfterLastCommand($this->runAfterLastCommand(...))
             ->withRunAfterBatch($this->runAfterBatch(...));
@@ -156,5 +157,13 @@ trait Parallelization
             $this->lock->release();
             $this->lock = null;
         }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getConsolePath(): string
+    {
+        return PIMCORE_PROJECT_ROOT . '/bin/console';
     }
 }
--- a/lib/Console/Traits/Parallelization.php
+++ b/lib/Console/Traits/Parallelization.php
@@ -78,9 +78,9 @@ trait Parallelization
     ): ParallelExecutorFactory {
         return ParallelExecutorFactory::create(...func_get_args())
             ->withSegmentSize($this->getSegmentSize())
-            ->withRunBeforeFirstCommand(Closure::fromCallable([$this, 'runBeforeFirstCommand']))
-            ->withRunAfterLastCommand(Closure::fromCallable([$this, 'runAfterLastCommand']))
-            ->withRunAfterBatch(Closure::fromCallable([$this, 'runAfterBatch']));
+            ->withRunBeforeFirstCommand($this->runBeforeFirstCommand(...))
+            ->withRunAfterLastCommand($this->runAfterLastCommand(...))
+            ->withRunAfterBatch($this->runAfterBatch(...));
     }
 
     /**
--- a/lib/Console/Traits/Parallelization.php
+++ b/lib/Console/Traits/Parallelization.php
@@ -15,13 +15,17 @@
 
 namespace Pimcore\Console\Traits;
 
+use Closure;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputDefinition;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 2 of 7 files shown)


lib/Console/Traits/Parallelization.php
```diff
@@ -64,6 +68,21 @@ protected static function configureParallelization(Command $command): void
...
+        callable $getItemName,
+        string $commandName,
+        InputDefinition $commandDefinition,
+        ErrorHandler $errorHandler
+    ): ParallelExecutorFactory {
+        return ParallelExecutorFactory::create(...func_get_args())
+            ->withSegmentSize($this->getSegmentSize())
+            ->withRunBeforeFirstCommand(Closure::fromCallable([$this, 'runBeforeFirstCommand']))
+            ->withRunAfterLastCommand(Closure::fromCallable([$this, 'runAfterLastCommand']))
+            ->withRunAfterBatch(Closure::fromCallable([$this, 'runAfterBatch']));
```
jdreesen: Pimcore 11 is `PHP ^8.1`, so why not use [First-class Callable Syntax](https://www.php.net/manual/en/functions.first_class_callable_syntax.php) here:

```suggestion
            ->withRunBeforeFirstCommand($this->runBeforeFirstCommand(...))
            ->withRunAfterLastCommand($this->runAfterLastCommand(...))
            ->withRunAfterBatch($this->runAfterBatch(...));
```

dvesh3:
@blankse thanks a lot!

bundles/CoreBundle/src/Command/ThumbnailsImageCommand.php
```diff
@@ -240,8 +240,8 @@ private function fetchThumbnailConfigs(InputInterface $input, string $thumbnailC
         return $thumbnailsToGenerate;
     }
 
-    protected function getItemName(int $count): string
+    protected function getItemName(?int $count): string
     {
-        return $count == 1 ? 'thumbnail' : 'thumbnails';
+        return $count === 1 ? 'thumbnail' : 'thumbnails';
```
theofidry, line 245: you probably also want:

```php
if (null === $count) {
    return 'thumbnail(s)';
}
```

And the same in a couple of other places

lib/Console/Traits/Parallelization.php
```diff
@@ -64,6 +68,22 @@ protected static function configureParallelization(Command $command): void
...
+        callable $fetchItems,
+        callable $runSingleCommand,
+        callable $getItemName,
+        string $commandName,
+        InputDefinition $commandDefinition,
+        ErrorHandler $errorHandler
+    ): ParallelExecutorFactory {
+        return ParallelExecutorFactory::create(...func_get_args())
+            ->withSegmentSize($this->getSegmentSize())
+            ->withScriptPath($this->getConsolePath())
```
theofidry, line 81: is this (and the method `getConsolePath()`) really necessary? I think it should be that one picked by default and then if the user wants to use another one he can configure it like you are doing it, i.e. you can phase out that `getConsolePath()` method.

To double check with the last release there should be a more friendly upgrade path, at least it keeps most of the old parallelization trait API but points out what has been deprecated

lib/Console/Traits/Parallelization.php
```diff
@@ -15,13 +15,17 @@
...
+use Symfony\Component\Console\Input\InputDefinition;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Lock\LockFactory;
 use Symfony\Component\Lock\LockInterface;
+use Webmozarts\Console\Parallelization\ErrorHandler\ErrorHandler;
+use Webmozarts\Console\Parallelization\ParallelExecutorFactory;
 
 trait Parallelization
```
theofidry, line 30: I think the `::configureParallelization()` might need another review, a few arguments & options changed or at least were added

dvesh3: I created an issue https://github.com/pimcore/pimcore/issues/13724. thanks!

bundles/CoreBundle/src/Command/ThumbnailsImageCommand.php
mattamon: It would only be a cosmetical thing. I am not sure if we should more code just because of this.  if the count is null it will always be thumbnails which seems correct to me.

theofidry: Note necessarily, if it is `null` it just means `fetchItems()` returned a non-array `iterable`. Behind there could be 0, 1 or X items.

But yes ultimately it's only cosmetics, the same way pluralizing the name is :)

mattamon: You are totally right, but I think semantically it would not make a difference unless you really want to know if you got a non-array iterable, which should not be possible since fetchItems need to return an array? :)

theofidry: >which should not be possible since fetchItems need to return an array? :)

That's been changed in v2 to support the case where you don't know all the items beforehand, hence the change in signature 😄

lib/Console/Traits/Parallelization.php
mattamon: @dvesh3  Is there a specific reason, why the segment size is the batch size?
```php
    protected function getSegmentSize(): int
    {
        return (int)$this->input->getOption('batch-size');
    }
```
This is kind of confusing since the segment size would actually be an own option. Is it just because there is a default value of 50 for the batch size?

@theofidry would it be better if we add another option for the script path but by default use ```/bin/console```

theofidry: >Is there a specific reason, why the segment size is the batch size?

By default (coming from 1.x), the bach size was at `50` and the segment size was `getBatchSize()` (so `50` as well by default). I am not entirely sure of the reason, I think the intent was:

- assume the app is leaky (it was definitely the case internally when we introduced this trait before this was even a lib), so we could not really afford to have a segment size greater than the batch size at least in the common case (at that time)
- if you bump batch size the segment size should remain bigger than the batch size

So although not perfect, I reckon with that context this was "good enough" at the time.

Fast forward in 2.x:

- I kept those defaults although they are now in `ParallelExecutorFactory`
- Via the `ParallelExecutorFactory`, it can be configured directly at the class level, which is automatically done if you are overriding `getBatchSize()` or `getSegmentSize()` in the `Parallelization` trait (but you should get a deprecation notice warning you to configure the `ParallelExecutorFactory` directly instead)
- Via the input options

**Maybe the default could value of the segment size could be revisited however.**

>This is kind of confusing since the segment size would actually be an own option.

I believe the excerpt you gave is from your code not the parallelization library unless I remember wrong. So I agree this looks confusion, and no longer makes sense in 2.x at the very least. Could it be it was there in 1.x because at that time there was no such options?

>would it be better if we add another option for the script path but by default use /bin/console

I think the native extension point should be good enough (the `{get|configure}ParallelExecutorFactory()`).

Also note that now (in 2.x), the script path is actually detected automatically so I think the % of users that actually want to override it now is close to 0% (although still possible – and allowed).

mattamon: @theofidry thank you for the detailed explanation! 
I added the missing options and a deprecation notice for the configureParallelization.
I also removed the deprecated methods withSegmentSize and withScriptPath.
Maybe you could have a look at #13790 please :)

lib/Console/Traits/Parallelization.php
mattamon: Please checkout #13790 :)

PR #13484, opened 2022-10-28
Bugfix Migration:

ctippler:
Fix migration error if language key doesn't exist

[notice] Migrating up to Pimcore\Bundle\CoreBundle\Migrations\Version22020614115124 [error] Migration Pimcore\Bundle\CoreBundle\Migrations\Version20210323082921 failed during Execution. Error: "Warning: Undefined array key "language""

In Version20210323082921.php line 60:

  Warning: Undefined array key "language"

### DIFF
--- a/bundles/CoreBundle/Migrations/Version20210323082921.php
+++ b/bundles/CoreBundle/Migrations/Version20210323082921.php
@@ -57,7 +57,7 @@ final class Version20210323082921 extends AbstractMigration
 
             foreach ($data as $row) {
                 if (!empty($row['id'])) {
-                    if (!array_key_exists('language',$row) || $row['language'] === null) {
+                    if (!isset($row['language'])) {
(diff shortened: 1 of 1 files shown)


bundles/CoreBundle/Migrations/Version20210323082921.php
```diff
@@ -57,7 +57,7 @@ public function up(Schema $schema): void
 
             foreach ($data as $row) {
                 if (!empty($row['id'])) {
-                    if ($row['language'] === null) {
+                    if (!array_key_exists('language',$row) || $row['language'] === null) {
```
jdreesen: I would use `isset`, as it checks if the key exists and is not `null`.

```suggestion
                    if (!isset($row['language'])) {
```

aashan10 (approved):

PR #13490, opened 2022-10-31
[Task]: Add parameter types getGetter/getSetterCode

mcop1:
Add types for all parameters, including $language parameter
Removed setter functions for `CalculatedValue`, since they are not used, change return values for getPhpdocInputType/getParameterTypeDeclaration accordingly

### DIFF
--- a/models/DataObject/ClassDefinition/Data/CalculatedValue.php
+++ b/models/DataObject/ClassDefinition/Data/CalculatedValue.php
@@ -510,7 +510,7 @@ class CalculatedValue extends Data implements QueryResourcePersistenceAwareInter
         $code .= '* @param ' . $this->getPhpdocInputType() . ' $' . $key . "\n";
         $code .= '* @return $this' . "\n";
         $code .= '*/' . "\n";
-        $code .= 'public function set' . ucfirst($key) . '(' . '$' . $key . ', ?string $language = null): static' . "\n";
+        $code .= 'public function set' . ucfirst($key) . '(' . $this->getParameterTypeDeclaration() . ' $' . $key . ', ?string $language = null): static' . "\n";
         $code .= '{' . "\n";
         $code .= "\t" . 'return $this;' . "\n";
         $code .= "}\n\n";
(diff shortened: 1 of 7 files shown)


jdreesen:
The title is wrong, it should be *parameter type* ;)

mcop1:
> The title is wrong, it should be _parameter type_ ;)

Thanks

models/DataObject/ClassDefinition/Data/CalculatedValue.php
```diff
@@ -510,7 +510,7 @@ public function getSetterCodeLocalizedfields($class)
         $code .= '* @param ' . $this->getPhpdocInputType() . ' $' . $key . "\n";
         $code .= '* @return $this' . "\n";
         $code .= '*/' . "\n";
-        $code .= 'public function set' . ucfirst($key) . '(' . '$' . $key . ', $language = null): static' . "\n";
+        $code .= 'public function set' . ucfirst($key) . '(' . '$' . $key . ', ?string $language = null): static' . "\n";
```
blankse: Should we add also the missing parameter type here?
```suggestion
        $code .= 'public function set' . ucfirst($key) . '(' . $this->getParameterTypeDeclaration() . ' $' . $key . ', ?string $language = null): static' . "\n";
```

But this is missing also in the other getSetterCode methods...

mcop1: I was thinking about doing another pr, but I guess it is ok to do it in the same pr, so thanks for the suggestion!

blankse:
Why do empty setters exist for CalculatedValue? Should we remove them completely?

blankse:
The methods getParameterTypeDeclaration()/getPhpdocInputType() must return string "mixed"

mcop1:
> The methods getParameterTypeDeclaration()/getPhpdocInputType() must return string "mixed"

I changed the `getSetterCodeXXX` functions to return an empty string and adapted the testcases. I don´t think there is a point in keeping the setters, what do you think?

blankse:
I also think that we do not need them.

PR #13393, opened 2022-10-16
Removing bundle manager and registering bundles in extensions.php

aashan10:
## Changes in this pull request  
Resolves #11633

PR #13491, opened 2022-10-31
Remove unused translations

blankse:
Was removed also here https://github.com/pimcore/pimcore/pull/12881
But was added here again: https://github.com/pimcore/pimcore/commit/c1f1f832be30136bc920cedbfee1e687267d0650


blankse:
Also remove deprecated translation. Follow up to https://github.com/pimcore/pimcore/pull/13393

dvesh3:
> Also remove deprecated translation. Follow up to #13393

@blankse we intentionally kept the deprecation keys as it is generic and could be used going forward. could you please revert it? thanks!

blankse:
@dvesh3 Done

mcop1:
Thanks!

PR #13492, opened 2022-10-31
Fix missing semicolon in nginx config

luklewluk:
I copied and pasted nginx configuraton from the documentation and I noticed that it fails because of missing semicolon.


jdreesen (approved):

dvesh3:
@luklewluk thanks for the fix 👍

Issue #12607, opened 2022-07-06
[Improvement]: abstract web2print to support replacing it with alternative providers like Gotenberg


dkarlovi:
### Improvement description

See https://github.com/pimcore/pimcore/issues/11673#issuecomment-1165685397

A possible alternative (includes PHP client): https://gotenberg.dev/

Ideally, Pimcore would allow replacing the PDF / page screenshot tool to allow decorating.

brusch:
Actually I'm thinking of making Gotenberg even the default implementation and use it in our default `docker-compose.yml` files. This would shrink our `pimcore/pimcore` images massively and if you don't need it you can just kick it out.

dkarlovi:
@brusch I can see value in having it be bundled for simpler install, but for more advanced teams and deployments, the separation would be very beneficial. You could have say 10 or 100 Pimcore installs in a cluster all sharing a common print service.

brusch:
Sure, it's just about having a ready to use `docker-compose.yml` - of course you're still free to adapt it to your setup.

PR #13493, opened 2022-10-31
Headless Chrome web2print document-level config customizability

kingjia90:
Related https://github.com/pimcore/pimcore/discussions/13453

## Additional info  
As suggested in https://github.com/pimcore/pimcore/discussions/13453#discussioncomment-4018557

For headless chrome it would look like this
image

### DIFF
--- a/lib/Web2Print/Processor/HeadlessChrome.php
+++ b/lib/Web2Print/Processor/HeadlessChrome.php
@@ -39,6 +39,8 @@ class HeadlessChrome extends Processor
         $web2printConfig = $web2printConfig->get('headlessChromeSettings');
         $web2printConfig = json_decode($web2printConfig, true);
 
+        $config = (array) $config;
+
         $params = ['document' => $document];
         $this->updateStatus($document->getId(), 10, 'start_html_rendering');
         $html = $document->renderDocument($params);
@@ -46,19 +48,26 @@ class HeadlessChrome extends Processor
         $html = $this->processHtml($html, $params);
         $this->updateStatus($document->getId(), 40, 'finished_html_rendering');
 
-        if ($web2printConfig) {
+        if ($config) {
             foreach (['header', 'footer'] as $item) {
-                if (key_exists($item, $web2printConfig) && $web2printConfig[$item] &&
-                    $content = file_get_contents($web2printConfig[$item])) {
-                    $web2printConfig[$item . 'Template'] = $content;
+                if (key_exists($item, $config) && $config[$item] &&
+                    $content = file_get_contents($config[$item])) {
+                    $config[$item . 'Template'] = $content;
                 }
-                unset($web2printConfig[$item]);
+                unset($config[$item]);
+            }
+            foreach (['Top', 'Right', 'Bottom', 'Left'] as $item) {
+                $config['margin'][strtolower($item)] = $config['margin' . $item];
+                unset($config['margin' . $item]);
             }
         }
 
+        //Adds any other values that are not covered by the getProcessingOptions
+        $config = array_merge($config, array_diff_key($web2printConfig, $config));
+
         try {
             $this->updateStatus($document->getId(), 50, 'pdf_conversion');
-            $pdf = $this->getPdfFromString($html, $web2printConfig);
+            $pdf = $this->getPdfFromString($html, $config);
             $this->updateStatus($document->getId(), 100, 'saving_pdf_document');
         } catch (\Exception $e) {
             Logger::error((string) $e);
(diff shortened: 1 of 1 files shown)


NiklasBr:
~This setting is supposed to only show on Print Containers, right? So effectively the print editor need to create a Print container for each different layout?~ Edit: Forgot to look at the *Generate & Preview PDF* tab on Printpages

NiklasBr:
One problem, and I'm not sure it is a problem with this change or something else.

The settings appear to carried onto the `node` command as they ought to be:

image

But the set margins are not reflected in the end result:

image

kingjia90:
mhh weird... 
I have noticed that this library is quite "sensible", and at the same time error output is not that clear.
It could be that these empty fields like height and width are necessary on how the margin will be calculated, and there is a space between the value and the measure (margin right `50 mm`). 
Please retry with these changes, and then i will try to add some input sanitization and a `array_filter`, so that when empty value are there it would just use the library default instead of a defined empty

NiklasBr:
I figured out the issue above. I need to be specific with the height and width. That resolves it. However, there is still an issue with a single setting overriding every other setting. For example:

Container like this:
image

First page like this:
image

Second page like this:
image

Final page like this:
image

End result  shows the larger margins on all pages when generating for the container:
image

NiklasBr:
@kingjia90 looking closer it seems only one config is ever sent to `node` during the `convert()` method, so Chrome only ever gets one set of margins no matter how many specific settings there are. So the per-document configs still does not work.

kingjia90:
ahhh, that is not what  i was expecting from the printer container 🤔 i am not even sure if it's intended or is a bug or so.

What about an additional printer container generation method that basically merges single PDF files? It might be slower (if one doesn't have already all the pages/chapters generated) and probably require a pogress bar or so. Sounds something for 11.x though.

NiklasBr:
I'm not sure either, but it feels like such a basic thing with PDF Catalogs (such as [this example with a cover page](https://pimcore.com/en/customers/web-to-print-automates-creation-of-full-featured-product-catalog_c1052)) that it could be either a bug or an oversight.

In previous projects I have used [FPDI](https://manuals.setasign.com/fpdi-manual/v2/) to merge different PDF:s pages.

NiklasBr:
What's your conclusion @kingjia90 ?

kingjia90:
@NiklasBr i think that it may work, the extension you recommended looks also good.
The question is that this might be something for 11.x and therefore not going to be in the highest priority in terms of the whole planning and from a bigger picture perspective. 
For example, Gotemberg(https://github.com/pimcore/pimcore/issues/12607#issuecomment-1177197105) is taken in consideration, then we could just use its [merge](https://gotenberg.dev/docs/modules/pdf-engines#merge) command.

~Of course, if in the meantime you are able to provide a PR for this, i think it could be going into 10.6, but internally we are not working on adding improvement there.~ 🤦🏻 i've realized this phrase is misleading and sounds as BS, i wanted to say that if it turns out to be a bug that needs to be a minor (eg. due the requirement of FPDI), it could be potentially in 10.6. Sorry for that

NiklasBr:
@kingjia90 I have been trying out Gotenberg for a while, I think it is solvable in a decent way using that service, and can provide a PR for 10.6 if you would allow me.

brusch:
@NiklasBr Gotenberg would be a topic for `11.x` - a PR would be great 🥰

NiklasBr:
@brusch I am available to make a PR for 10.6 for Gotenberg compatibility.

fashxp:
Coming back to the original topic: 
The generate PDF settings are meant as settings for the PDF generation process as a whole (color spaces, adding bookmarks, activating printer marks, etc.), and not for page (or chapter) specific settings like different margins for different layout types. So this works as intended. 

Generating separate PDFs and merging them afterwards will cause issues, for example when generating TOCs or glossaries with Javascript and when linking inbetween the document. 

When we need to define different margings etc. on page level, we do that via CSS and eventually add properties or similar on the corresponding document. 
For the pdfreactor you can find different page layouts in our demo.

NiklasBr:
@fashxp I hear you. Let us leave aside the updated PDF settings currently in this pull request. Let us also leave aside the Web2Print settings which applies to all documents equally. Let's file that under that working as intended" for the moment. But even if we do, there are still unresolved idiosyncrasies/issues/bugs which absolutely ought to be fixed.

Unexpected or lost functionality is solidly in the _bug_ territory.  This is not new functionality, this is restoring functionality promised. And we offer this fix as a PR to Pimcore, You are of course 100% free to decline.

First one is how different Templates currently break regular CSS-based rules. And given what our customer understood, and I understood, and @kingjia90 [understood](https://github.com/pimcore/pimcore/pull/13493#issuecomment-1298551019) this was supposed to work the way it has been sold to us.

We assumed we could create "fully featured product catalogues" using Pimcore's Web To Print functionality, for example based on these quotes (my emphasis):

> [This makes the creation of **_full-featured product catalogs_** and price lists possible without using any desktop publishing software such as InDesign. Print-ready, high-quality PDFs can be assembled in a user-friendly interface. Either fully automated or based on flexible layout components. ](https://pimcore.com/en/digital-experience-platform/features/web-to-print)

> [This feature can be used for](https://pimcore.com/en/resources/blog/web-to-print-is-now-part-of-the-open-source-pimcore-distribution_a979)
> 
> - One-page flyers
> - Product Sheets
> - Price Lists
> - **_Full-featured Product Catalogs_**

> [Pimcore based Web-To-Print solution with a set of pre-configured templates, an integrated rendering engine](https://pimcore.com/en/customers/pollmann_c1052)

brusch:
@NiklasBr unfortunately we can't accept any further features for `10.x` (we need to focus here) - but a PR for `11.x`  would be great! 🥰

NiklasBr:
@brusch understood.

Do you have any idea how to resolve the CSS issues?

NiklasBr:
The offer to provide a PR on 10.x to fix this is still open, we have the code running locally and in a cloud environment.

kingjia90:
Closing in favor of https://github.com/pimcore/pimcore/pull/13806

NiklasBr:
@kingjia90 these two are not the same. The event listener will not resolve this issue (I tested it).

You can easily see the problem still persists after applying the #13806 patch by changing the *config* on template two or later and it will apply to all earlier templates.

kingjia90:
Yes, that's true, but i am closing this PR for the moment, because it was initially meant to fix the problem of lacking a personalization at document level beside global settings. 
 
This PR's approach attempts to fix that problem by adding a more user-friendly interface to customize settings, however, it may turn into a sort of "unrequested feature" (causing some ambiguity wheter it should be considered as bug or improvement), and also it doesn't fully list all the available options, so it's more "limited" in comparison with the event listener approach.

The event listener PR resolves the same initial problem without BC breaks nor adding new features, therefore, at least speaking for 10.5, would be the best.


The problem of not being able to define the settings on document level when exporting from an upper container level is still open but tha  still would be something that would be done and merged in 11.x

NiklasBr:
> on document level when exporting from an upper container

No, that's not what my initial bug was about. Sorry if that is what you took away from it. Sorry if I was unclear.

kingjia90:
> > on document level when exporting from an upper container
> 
> No, that's not what my initial bug was about. Sorry if that is what you took away from it. Sorry if I was unclear.

I have incorrectly "summerized" the initial issue, let me rephrase: "allowing and applying custom setting on a single page level when exporting a collection of pages"

NiklasBr:
Oh well. In any case, the offer to supply a non-breaking patch to 10.5 still stands if you will accept it.

PR #13498, opened 2022-11-02
[Improvement]: Being able to save filters on grid options

### LINKED ISSUE #11651: [Improvement]: Filters should be saved for table options too
### Improvement description

Currently we can save table options in dataobject folders.
But it should be possible to save the filters too. For example if someone wants to save a list of users which have newsletter enabled.

Bildschirmfoto 2022-03-11 um 14 42 57

SamyMP:
Resolves #11651

## Additional info  
Probably still WIP, but I managed to add a button in the grid options to save the filters applied on the columns :
image

The JSON in the `gridconfigs` table in the database now contains an array "filter" with the filter saved. Example : 
```json
"filter": [
    {
      "operator": "lt",
      "value": 13,
      "id": "x-gridfilter-id-lt",
      "property": "id",
      "root": "data",
      "serializer": null,
      "type": "numeric"
    }
  ],
```

To remove a saved filter from a grid config, you have to clear filters, then check the checkbox "Save filters", otherwise, the filters will stay on the grid.

What do you think ? As I said, there may still be some work to do on this


models/GridConfig.php
```diff
@@ -297,6 +299,22 @@ public function setSetAsFavourite($setAsFavourite)
...
+        return $this->saveFilters;
+    }
+
+    /**
+     * @param bool $saveFilters
+     */
+    public function setSaveFilters($saveFilters)
+    {
+        $this->saveFilters = (bool) $saveFilters;
+    }
```
blankse: Please use type hints
```suggestion
    public function isSaveFilters(): bool
    {
        return $this->saveFilters;
    }

    public function setSaveFilters(bool $saveFilters): void
    {
        $this->saveFilters = $saveFilters;
    }
```

robertSt7:
@SamyMP When I try to save a filter with a select data component, it does not work for me. Could you please take a look at this? Thanks

SamyMP:
No problem, I'll take a look

SamyMP:
@robertSt7 I've just committed the fix for the select data component

robertSt7:
@SamyMP Thanks

BigJabber:
This is a feature a lot of our users have been asking for. I've noticed that the option has only been made available for admin users but is there any reason that this shouldn't be managed via role/user permissions?

SamyMP:
I merely copied the way the other checkboxes worked (set as favorite, share globally), so I hadn't noticed that it was not available to non admin users. Maybe a PR can be submitted to address that

PR #13502, opened 2022-11-02
Deprecate PreAuthenticatedAdminSessionListener

dvesh3:
Follow up to https://github.com/pimcore/pimcore/pull/12606

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
@@ -52,7 +52,7 @@ location / {
     security:
         enable_authenticator_manager: true
     ```
-  `PreAuthenticatedAdminSessionListener` has been deprecated and will be removed in Pimcore 11.
+  `[Authentication] PreAuthenticatedAdminSessionListener` has been deprecated and will be removed in Pimcore 11.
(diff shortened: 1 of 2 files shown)


brusch:
@dvesh3 we should also add an upgrade note for that, or is it already there? 😊

dvesh3:
> @dvesh3 we should also add an upgrade note for that, or is it already there? 😊

The reason it is missing because I directly edited this file on Github 😄  (to avoid switching from 11.x branch to 10.5 on my local) but now is the time to have a separate instance for 10 😉

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -52,6 +52,7 @@ location / {
     security:
         enable_authenticator_manager: true
     ```
+  `PreAuthenticatedAdminSessionListener` has been deprecated and will be removed in Pimcore 11.
```
kingjia90:
```suggestion
  `[Authentication] PreAuthenticatedAdminSessionListener` has been deprecated and will be removed in Pimcore 11.
```

kingjia90 (changes requested):
Does this service also need a deprecation message?

https://github.com/pimcore/pimcore/blob/38419b06c61565e62122e32684e37c102cbe25d2/bundles/AdminBundle/config/security_abstract_services.yaml#L20-L21

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
dvesh3: > Does this service also need a deprecation message?

I would avoid that since this used internally and then it triggers the deprecations on compilation.

PR #13508, opened 2022-11-03
Concrete::getById() must return only Concrete objects

blankse:
Concrete::getById() returns also Folder objects. But this is wrong because the return type is static|null

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
@@ -1,7 +1,7 @@
 # Upgrade Notes
 
 ## 10.5.10
-- [DataObject] Deprecated: Loading none Concrete objects with the Concrete class will not possible in Pimcore 11.
+- [DataObject] Deprecated: Loading non-Concrete objects with the Concrete class will not be possible in Pimcore 11.
 
 ## 10.5.8
 - [Twig] Sending mails and Dataobject Text Layouts, which allow rendering user controlled twig templates are now executed in a sandbox with restrictive security policies for tags, filters, functions.
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
@@ -1,5 +1,8 @@
 # Upgrade Notes
 
+## 10.5.10
+- [DataObject] Deprecated: Loading none Concrete objects with the Concrete class will not possible in Pimcore 11.
+
 ## 10.5.8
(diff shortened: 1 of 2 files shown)


models/DataObject/AbstractObject.php
```diff
@@ -516,7 +516,7 @@ public static function getTotalCount($config = [])
      */
     protected static function typeMatch(AbstractObject $object)
     {
-        return in_array(static::class, [Concrete::class, __CLASS__], true) || $object instanceof static;
+        return static::class === __CLASS__ || $object instanceof static;
```
dvesh3: This change can break implementations relying on `Concrete::getById()` to load folder objects. Maybe we just trigger deprecation here and change the behavior in Pimcore 11.

blankse:
```suggestion
        if (static::class === Concrete::class && !$object instanceof static) {
            trigger_deprecation(
                'pimcore/pimcore',
                '10.5',
                'Loading none Concrete objects with via the Concrete class will not possible in Pimcore 11'
            );

            return true;
        }

        return static::class === __CLASS__ || $object instanceof static;
```

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -1,5 +1,8 @@
 # Upgrade Notes
 
+## 10.5.10
+- [DataObject] Deprecated: Loading none Concrete objects with the Concrete class will not possible in Pimcore 11.
```
```suggestion
- [DataObject] Deprecated: Loading non-Concrete objects with the Concrete class will not be possible in Pimcore 11.
```

models/DataObject/AbstractObject.php
```diff
@@ -516,7 +516,17 @@ public static function getTotalCount($config = [])
      */
     protected static function typeMatch(AbstractObject $object)
     {
-        return in_array(static::class, [Concrete::class, __CLASS__], true) || $object instanceof static;
+        if (static::class === Concrete::class && !$object instanceof static) {
+            trigger_deprecation(
+                'pimcore/pimcore',
+                '10.5',
+                'Loading none Concrete objects with the Concrete class will not possible in Pimcore 11'
```
jdreesen:
```suggestion
                'Loading non-Concrete objects with the Concrete class will not be possible in Pimcore 11'
```

models/DataObject/AbstractObject.php
```diff
@@ -516,7 +516,17 @@ public static function getTotalCount($config = [])
...
+            );
+
+            return true;
+        }
+
+        return static::class === __CLASS__ || $object instanceof static;
```
jdreesen: What do you think about this? It looks a bit more consistent.

```suggestion
        return static::class === self::class || $object instanceof static;
```

dvesh3:
@blankse thanks!

PR #13509, opened 2022-11-03
[Bug fix]: handle the case where the ObjectFieldGetter returns a list of objects

bgpack:
Adding check for return value of getter is an array otherwise the returned objects are not accessible in the grid selection by the next operation.


Corepex:
@bgpack thanks for helping us to improve pimcore.

btw. congrats to your first PR 🥇  - keep going

PR #13511, opened 2022-11-03
[Bug]: Fix video editable height configuration problem with percentage values and improving error rendering

### LINKED ISSUE #13472: [Bug]: A string and int operation problem in Video Editable error return html code
### Expected behavior

Using css [calc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc) for `%` and `px` values and consistency between `width` and `height`.

### Actual behavior

https://github.com/pimcore/pimcore/blob/c07e8ab75a89c859e4ea1bbbb9a6174bd9cd6fc9/models/Document/Editable/Video.php#L557-L559

`($this->getHeight() - 1)` in this line
https://github.com/pimcore/pimcore/blob/c07e8ab75a89c859e4ea1bbbb9a6174bd9cd6fc9/models/Document/Editable/Video.php#L568

### Steps to reproduce

I don't remember how i end up to see it, just had some problems with video editable and thumbnails. 
The height value was set to be `auto` somehow and with the `-1` in php caused a fatal error.

kingjia90:
Resolves #13472

Update: i was trying to reproduce the error again and found out that this twig template in Demo may be the culprit
https://github.com/pimcore/demo/blob/686e91e30c430d54487a66c8558cb6ea5c2051b5/templates/areas/video/view.html.twig#L9
image

The `auto` value in the twig template isn't complying with the valid values
https://github.com/pimcore/pimcore/blob/747b0d7761fdd8a933de5fdd8e3a7990336589e2/doc/Development_Documentation/03_Documents/01_Editables/38_Video.md?plain=1#L16

and it does not actually work with percentage as stated because of ` ($this->getHeight() - 1) . 'px`

### DIFF
--- a/models/Document/Editable/Video.php
+++ b/models/Document/Editable/Video.php
@@ -576,11 +576,17 @@ class Video extends Model\Document\Editable implements IdRewriterInterface
         // If contains at least one digit (0-9), then assume it is a value that can be calculated,
         // otherwise it is likely be `auto`,`inherit`,etc..
         if (preg_match('/[\d]/', $width)){
+
+            // when is numeric, therefore is without length units or %, then is px
+            if (is_numeric($width)) {
+                $width .= 'px';
+            }
             $width = 'calc(' . $width . ' - 1px)';
         }
 
         $height = $this->getHeight();
         if (preg_match('/[\d]/', $height)){
+
             if (is_numeric($height)) {
                 $height .= 'px';
             }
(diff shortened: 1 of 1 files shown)


kingjia90:
Update: i was trying to reproduce the error again and found out that this twig template in Demo may be the culprit
https://github.com/pimcore/demo/blob/686e91e30c430d54487a66c8558cb6ea5c2051b5/templates/areas/video/view.html.twig#L9
image

So this PR is just a nice to have to make it less error prone

PR #13518, opened 2022-11-04
JS Error in properties tab when we try to add property without type

croitoruion:
Added missing "check" in if condition before trying to access length, to avoid calling on null.
Attached screenshot to ensure everything is clear.

## Additional info  

image


Corepex:
@croitoruion thanks for this PR :) - have a nice day

PR #13438, opened 2022-10-24
Added exception handling in migrations

aashan10:
## Changes in this pull request  
Resolves #13372

PR #13526, opened 2022-11-08
Fix migration execution - Version20201008101817

dvesh3:
follow up to #13438


brusch:
@dvesh3 maybe we should use `SHOW FULL TABLES` to check whether the view exists or not and then use standard deferred `addSql()`. wdyt?

dpfaffenbauer:
@dvesh3 @brusch maybe `$this->connection->getSchemaManager()->listViews()` works?

dvesh3:
yes, I think it is better to check if view exists and then add the sql to avoid error message from the try catch. I'll give it a try with @dpfaffenbauer's suggestion. thanks!

PR #13533, opened 2022-11-08
change dialog rendering to use default overwrite logic when using inheritance

### LINKED ISSUE #13431: [Bug]: relations editable in editable dialog box for brick can not be overwritten with inheritance
### Expected behavior

I have a simple test brick with one "relations" editable in editable dialog box.
I created a English page, add the brick and add an data object to the relations editable.
Then I copied the English page with inheritance to german language. When I add an other data object to the relation editable in the german version (click override at the brick), save an reload, there are 2 data objects in the relations editable.

### Actual behavior

After overriding the brick in the German language an add an other data object, I save the page, reload it and then the relations editable only shows the data object from the English page. The newly added data object is not there.
With inheritance it is not possible to override a relations editable in the editable dialog box.
Without inheritance all works fine.

### Steps to reproduce

You can use the brick from the [zip](https://github.com/pimcore/pimcore/files/9836263/pimcore-test.zip) file.

I am on Pimcore 10.5.8.

hallekamp:
This PR fixes that inherited elements in dialogs never have their inherited property changed.

This behavior occurs when using translations with inheritance.

1. To reproduce, add any brick that uses `getEditableDialogBoxConfiguration()` to a page.
2. Copy and Paste (Inheritance) as a new language variant.
3. In the translated page, overwrite the element from step 1 and try to edit any setting in the dialog.
4. After this fix, the inherited property gets set to false and changes are persisted.

Resolves #13431

## Additional info  
I fixed this for a customer, so I haven't tested if it actually resolves the issue above, but it most likely will.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/area_abstract.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/area_abstract.js
@@ -45,6 +45,7 @@ pimcore.document.area_abstract = Class.create(pimcore.document.editable, {
                                 } else {
                                     editablesInBox[editableName].render();
                                 }
+                                editablesInBox[editableName].setInherited(editablesInBox[editableName].inherited);
                             });
                         }, 200);
                     }


aryaantony92:
@hallekamp Thank you!

PR #13534, opened 2022-11-08
[Predefined Asset Metadata] Prevent creation of predefined item with same name

BlackbitDevs:
Steps to reproduce:

1. Create predefined asset meta data (e.g. simply keep default name `New Definition`)
2. Create another predefined asset meta data with same name as for 1.
3. Close predefined meta data panel and reopen it

Expected result:
Only one item should exist because the names are unique according to https://github.com/pimcore/pimcore/blob/e8173ce9655519d57f3878804648d03fbe9a4c05/bundles/AdminBundle/Controller/Admin/SettingsController.php#L207-L210

When creating duplicate, an error message should be shown.

Actual result:
In https://github.com/pimcore/pimcore/blob/e8173ce9655519d57f3878804648d03fbe9a4c05/bundles/AdminBundle/Controller/Admin/SettingsController.php#L203 an empty predefined meta data item gets created and saved to the database
https://github.com/pimcore/pimcore/blob/e8173ce9655519d57f3878804648d03fbe9a4c05/models/Metadata/Predefined.php#L146

The actual check if there is another item with the same name is only done afterwards https://github.com/pimcore/pimcore/blob/e8173ce9655519d57f3878804648d03fbe9a4c05/bundles/AdminBundle/Controller/Admin/SettingsController.php#L207-L210

This then recognizes the existence of a duplicate and returns the `rule_violation`. But the user does not get notified of this error.
Moreover after reloading the predefined meta data panel, there is a PHP error in https://github.com/pimcore/pimcore/blob/e8173ce9655519d57f3878804648d03fbe9a4c05/models/Metadata/Predefined.php#L372 because the `type` has not been set for the duplicate. And thus you cannot even remove this empty duplicate via UI.

This PR adjusts `Predefined::create()` to not save the item because in this moment it is completely empty anyway and cannot be loaded. Maybe it was designed differently as the comment https://github.com/pimcore/pimcore/blob/e8173ce9655519d57f3878804648d03fbe9a4c05/bundles/AdminBundle/Controller/Admin/SettingsController.php#L202 says that the type gets saved - but in fact it does not.

Additionally this PR show an error window in case of a server-side error.

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers/generic-grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers/generic-grid.js
@@ -51,6 +51,7 @@ pimcore.helpers.grid.buildDefaultStore = function (url, fields, itemsPerPage, cu
         listeners: {
             exception: function(proxy, response, operation){
                 pimcore.helpers.showNotification(t("error"), t(operation.getError()), "error");
+                store.load();
             }
         }
     });
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers/generic-grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers/generic-grid.js
@@ -27,8 +27,7 @@ pimcore.helpers.grid.buildDefaultStore = function (url, fields, itemsPerPage, cu
         type: 'ajax',
         reader: {
             type: 'json',
-            rootProperty: 'data',
-            messageProperty: 'message'
+            rootProperty: 'data'
         },
         writer: {
             type: 'json',
@@ -47,13 +46,17 @@ pimcore.helpers.grid.buildDefaultStore = function (url, fields, itemsPerPage, cu
             read: 'POST',
             update: 'POST',
             destroy: 'POST'
-        },
+        }/*,
         listeners: {
             exception: function(proxy, response, operation){
-                pimcore.helpers.showNotification(t("error"), t(operation.getError()), "error");
-                store.load();
+                Ext.MessageBox.show({
+                    title: 'REMOTE EXCEPTION',
+                    msg: operation.getError(),
+                    icon: Ext.MessageBox.ERROR,
+                    buttons: Ext.Msg.OK
+                });
             }
-        }
+        }*/
     });
 
     var config = {
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers/generic-grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers/generic-grid.js
@@ -27,7 +27,8 @@ pimcore.helpers.grid.buildDefaultStore = function (url, fields, itemsPerPage, cu
         type: 'ajax',
         reader: {
             type: 'json',
-            rootProperty: 'data'
+            rootProperty: 'data',
+            messageProperty: 'message'
         },
         writer: {
             type: 'json',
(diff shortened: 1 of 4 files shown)


aashan10:
@BlackbitDevs I am still able to create duplicate predefined metadata while testing this pr. Also, I am not getting any errors and even when I try to manually trigger an error by throwing an exception in this part https://github.com/pimcore/pimcore/blob/e8173ce9655519d57f3878804648d03fbe9a4c05/bundles/AdminBundle/Controller/Admin/SettingsController.php#L207-L210

I got the error dialog box. But the error message  is not actually visible and is something like this
Screenshot 2022-11-14 at 20 00 07

BlackbitDevs:
@aashan10 What did you do to create the duplicate item? The same as described above?

> 1. Create predefined asset meta data (e.g. simply keep default name New Definition)
> 2. Create another predefined asset meta data with same name as for 1.
> 3. Close predefined meta data panel and reopen it

aashan10:
@BlackbitDevs yes, I did the same process. Just clicked the add button a bunch of times  and reloaded the page. Found multiple entries with the same name

BlackbitDevs:
@aashan10 I found the cause. There was another bug in https://github.com/pimcore/pimcore/blob/23b66280576675381569d7e588668c78d94d0c16/models/Metadata/Predefined/Listing.php#L108-L129
As `array_filter()` keeps the original array keys, it could be that it returns something like
```php
[ 1 => object(Pimcore\Model\Metadata\Predefined), ... ]
```
This happens when another than the first item matches.
In this case `return $definitions[0] ?? null;` returns `null` although a matching item has been found. 

I changed this method to a foreach loop with early return. This is more efficient - and even more important: it works ;-)

aashan10:
https://user-images.githubusercontent.com/18713900/202371748-0fb9226a-0b24-44b6-8cae-33eff4eac762.mov

@BlackbitDevs There are still some random errors that I encountered while testing this. Please have a look

Jan-Walther-Blackbit:
@aashan10 Ok, another try ;-)
The problem was that when there was a duplicate entry, the ids of those 2 duplicates were the same in ExtJS. Thus changing the name of a duplicate entry triggered an update with the wrong item (the one with the duplicate name and thus the `rule_violation` appeared again.
I have now changed it that in case of a server-side error the store of the grid gets reloaded so that the user sees the current database data.

BlackbitDevs:
@aashan10 Ok, another try ;-)
The problem was that when there was a duplicate entry, the ids of those 2 duplicates were the same in ExtJS. Thus changing the name of a duplicate entry triggered an update with the wrong item (the one with the duplicate name and thus the rule_violation appeared again.
I have now changed it that in case of a server-side error the store of the grid gets reloaded so that the user sees the current database data.

bundles/AdminBundle/Resources/public/js/pimcore/helpers/generic-grid.js
```diff
@@ -46,17 +47,13 @@ pimcore.helpers.grid.buildDefaultStore = function (url, fields, itemsPerPage, cu
             read: 'POST',
             update: 'POST',
             destroy: 'POST'
-        }/*,
+        },
         listeners: {
             exception: function(proxy, response, operation){
-                Ext.MessageBox.show({
```
aashan10: Since this is a generic grid, which might be used in multiple places, is it really a good idea to change the messagebox to a notification?

BlackbitDevs: Yes, I know. In my opinion it is good because otherwise the user will just see nothing in case of errors. But if you are afraid we can duplicate this code and use the notification window only for predefined metadata.

It is interesting that the Pimcore maintainers often ask to not duplicate code to have the logic in one place but then are afraid of changes for this one place ;-)

aashan10: @BlackbitDevs I am also confused about this one. Maybe @dvesh3 has an opinion about it?

dvesh3: I would go for adding this listener in predefined metadata and not sure if we should reload the store in case of exception. maybe something like https://github.com/pimcore/pimcore/blob/28718e6101a3d603dec2e318d89f8866c2517364/bundles/AdminBundle/public/js/pimcore/object/helpers/grid.js#L151 is enough.
 
 > It is interesting that the Pimcore maintainers often ask to not duplicate code to have the logic in one place but then are afraid of changes for this one place ;-)

The reason for being skeptical about this change is to analyze the impact as this generic helper is used in lot of places and testing every dependent feature requires efforts. We always appreciate the code which is maintainable however sometimes the fix is more important than refactoring 😉

aashan10 (approved):

PR #13537, opened 2022-11-08
[Bug] DataObject: add validation for asset upload in ManyToManyRelation fields

storbahn:
At the moment it is possible to upload a disallowed asset type to a ManyToManyRelation field of a DataObject.

Expected Behaviour:

Show an error message and prevent the upload of the disallowed asset.

Current Behaviour:

It's possible to upload the disallowed asset. The data object is validated only when the object is saved. But at this point the asset was already uploaded.

Resolves #

## Additional info  

Question:

I'm not sure, where is the best place of the implemented code of this PR, e.g. the classes AssetTypeHelper and the ManyToManyRelationValidator?


Corepex:
@storbahn awesome that you found (and fixed) this bug. I have a few suggestions to improve your code. Would be very nice if you could implement these 👍

bundles/AdminBundle/Event/ManyToManyRelationAssetUploadSubscriber.php
```diff
@@ -0,0 +1,71 @@
...
+use Pimcore\Bundle\AdminBundle\Helper\AssetTypeHelper;
+use Pimcore\Bundle\AdminBundle\Helper\ManyToManyRelationValidator;
+use Pimcore\Event\AssetEvents;
+use Pimcore\Event\Model\Asset\ResolveUploadTargetEvent;
+use Pimcore\Model\DataObject\ClassDefinition\Data\ManyToManyRelation;
+use Pimcore\Model\DataObject\Concrete;
+use Pimcore\Model\Element\ValidationException;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+final class ManyToManyRelationAssetUploadSubscriber implements EventSubscriberInterface
```
Corepex: Looks like you forgot to commit the `service.yml` file where you added the event listener as a service.

Furthermore, I would move the file into the `EventListener` folder and rename it according to the other Symfony naming conventions (so it would be `ManyToManyRelationAssetUploadListener.php`)

bundles/AdminBundle/Helper/AssetTypeHelper.php
```diff
@@ -0,0 +1,31 @@
...
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\AdminBundle\Helper;
+
+use Pimcore\Model\Asset;
+use Symfony\Component\Mime\MimeTypes;
+
+class AssetTypeHelper
```
Corepex: You should register this class as a service since autowire isn't working otherwise

bundles/AdminBundle/Helper/ManyToManyRelationValidator.php
```diff
@@ -0,0 +1,32 @@
...
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\AdminBundle\Helper;
+
+use Pimcore\Model\DataObject\ClassDefinition\Data\ManyToManyRelation;
+
+class ManyToManyRelationValidator
```
Corepex: same as above - register this class as service

bundles/AdminBundle/Helper/AssetTypeHelper.php
jdreesen: Are there any reasons to create separate classes for this instead of just inlining the methods into the listener?

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -491,7 +492,8 @@ protected function addAsset(Request $request, Config $config)
         if ($context) {
             $context = json_decode($context, true);
             $context = $context ? $context : [];
-            $event = new \Pimcore\Event\Model\Asset\ResolveUploadTargetEvent($parentId, $filename, $context);
+            $context['sourcePath'] = $sourcePath;
+            $event = new ResolveUploadTargetEvent($parentId, $filename, $context);
```
brusch, line 501: I think we could and should do the validation right here, I don't see any benefit of having this in an event ... 🤔

brusch (changes requested):

bundles/AdminBundle/Event/ManyToManyRelationAssetUploadSubscriber.php
storbahn: I used the Symfony naming convention @see https://symfony.com/doc/current/event_dispatcher.html or https://symfony.com/doc/current/components/event_dispatcher.html#using-event-subscribers

For EventSubscribers we don't need the service.yml (only implementing the EventSubscriberInterface).

In my opinion we should rename the EventListener folder to EventSubscriber or Event. All classes implement the EventSubscriberInterface interface (but not part of this PR).

bundles/AdminBundle/Helper/AssetTypeHelper.php
storbahn: My idea was to make writing unit tests more easier.

bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
storbahn: @brusch Updated the PR.

brusch: @storbahn Thanks!

Corepex:
@storbahn, thanks for fixing this bug and helping us to improve pimcore 💯

PR #13538, opened 2022-11-09
[Bug]: Objectbrick definition update fails when changing field of type text to type textarea

### LINKED ISSUE #13436: [Bug]: Objectbrick definition update fails when changing field of type text to type textarea
### Expected behavior

Save of objectbrick field definition to succeed.

### Actual behavior

Save fails with MySQL error resulting in the following exception:
`An exception occurred while executing 'ALTER TABLE `object_brick_query_catalog_product_description_1` CHANGE COLUMN `fieldValue` `fieldValue` longtext NULL;':
SQLSTATE[42000]: Syntax error or access violation: 1170 BLOB/TEXT column 'fieldValue' used in key specification without a key length.`
image

### Steps to reproduce

From admin panel:
- Create Class definition with objectbrick field.
- Create object brick definition with 1 text field linked to the class and save it.
- Change the type of the objectbrick field from text to textarea and save.
The exception occurs.
This same fieldtype change scenario works as expected when done in class definition but fails when done to objectbrick definition.
If we create the objectbrick field with type textarea the first time than the save works as expected as there is no such index. It seems that the index created for the text type field is not being dropped before changing the column type in MySQL.

mcop1:
Resolves #13436

### DIFF
--- a/models/DataObject/Objectbrick/Definition/Dao.php
+++ b/models/DataObject/Objectbrick/Definition/Dao.php
@@ -15,6 +15,7 @@
 
 namespace Pimcore\Model\DataObject\Objectbrick\Definition;
 
+use Pimcore\Db\Helper;
 use Pimcore\Model;
 use Pimcore\Model\DataObject;
 
@@ -111,6 +112,9 @@ class Dao extends Model\Dao\AbstractDao
 
         DataObject\ClassDefinition\Service::updateTableDefinitions($this->tableDefinitions, ([$tableStore, $tableQuery]));
 
+        $this->removeIndices($tableStore, $columnsToRemoveStore, $protectedColumnsStore);
+        $this->removeIndices($tableQuery, $columnsToRemoveQuery, $protectedColumnsQuery);
+
         foreach ($this->model->getFieldDefinitions() as $value) {
             $key = $value->getName();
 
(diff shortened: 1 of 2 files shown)


models/DataObject/Objectbrick/Definition/Dao.php
```diff
@@ -173,4 +177,22 @@ public function classSaved(DataObject\ClassDefinition $classDefinition)
...
+     * @param string $table
+     * @param array $columnsToRemove
+     * @param array $protectedColumns
+     */
+    protected function removeIndices($table, $columnsToRemove, $protectedColumns)
+    {
+        if (is_array($columnsToRemove) && count($columnsToRemove) > 0) {
+            $indexPrefix = str_starts_with($table, 'object_brick_query_') ? 'p_index_' : 'u_index_';
+            foreach ($columnsToRemove as $value) {
+                if (!in_array(strtolower($value), array_map('strtolower', $protectedColumns))) {
```
jdreesen: I don't think you should call this `array_map('strtolower', $protectedColumns)` over and over again in the loop.

mcop1: Good point, I copied this part of the method. I think we should do it for both then.

dvesh3 (approved):

PR #13543, opened 2022-11-09
Check for limit reach before uploading via upload dialog

croitoruion:
Resolves #
Perform frontend validation if we can upload additional object, as it is when we use filter dialog.
So this fix, will display same "limit reach error" when using upload dialog modal.

## Additional info  
image


Corepex:
@croitoruion please also add the limit check to `bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyRelation.js`.

croitoruion:
> @croitoruion please also add the limit check to `bundles/AdminBundle/Resources/public/js/pimcore/object/tags/manyToManyRelation.js`.

ohh sorry forgot about that. 
done thanks.

Corepex:
@croitoruion nice - thanks :)

PR #13544, opened 2022-11-09
[Fix] moving "allowToCreateNewObject" option out of class context scope

### LINKED ISSUE #13495: [Bug]: Unable to remove  shortcut add button  from many to many object relation field
### Expected behavior

image
image

If we add a many to many object relation field in class It should display  checkbox for  shortcut object create button

### Actual behavior

But its not showing if I  add many to many object relation field in class
image

### Steps to reproduce

1. Create Field Collection 
2. Add a many to many object relation

mattamon:
Moving the "allowToCreateNewObject" option outside of class scope.
Resolves #13495

## Additional info  
"allowToCreateNewObject" can now be disabled in fieldcollections or objectbricks

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/manyToManyObjectRelation.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/manyToManyObjectRelation.js
@@ -192,18 +192,19 @@ pimcore.object.classes.data.manyToManyObjectRelation = Class.create(pimcore.obje
 
         this.specificPanel.add({
             xtype: "checkbox",
+            boxLabel: t("enable_text_selection"),
+            name: "enableTextSelection",
+            value: this.datax.enableTextSelection
+        });
+
+        this.specificPanel.add({
+            xtype: "checkbox",
             boxLabel: t("allow_to_create_new_object"),
             name: "allowToCreateNewObject",
             value: this.datax.allowToCreateNewObject
         });
 
         if(this.context == 'class') {
-            this.specificPanel.add({
-                xtype: "checkbox",
-                boxLabel: t("enable_text_selection"),
-                name: "enableTextSelection",
-                value: this.datax.enableTextSelection
-            });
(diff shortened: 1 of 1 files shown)


bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/manyToManyObjectRelation.js
```diff
@@ -190,19 +190,20 @@ pimcore.object.classes.data.manyToManyObjectRelation = Class.create(pimcore.obje
...
 
+        this.specificPanel.add({
+            xtype: "checkbox",
+            boxLabel: t("allow_to_create_new_object"),
+            name: "allowToCreateNewObject",
+            value: this.datax.allowToCreateNewObject
+        });
+
         if(this.context == 'class') {
             this.specificPanel.add({
```
Corepex, line 201: maybe we should move the "enable text selection" out of the class scope as well

bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/manyToManyObjectRelation.js
```diff
@@ -190,19 +190,20 @@ pimcore.object.classes.data.manyToManyObjectRelation = Class.create(pimcore.obje
...
                 xtype: "checkbox",
                 boxLabel: t("enable_text_selection"),
```
Corepex: maybe we should move the "enable text selection" out of the class scope as well

mattamon: Yes, this would be closer to the advancedManyToManyObjectRelation then.
https://github.com/pimcore/pimcore/blob/10.5/bundles/AdminBundle/Resources/public/js/pimcore/object/classes/data/advancedManyToManyObjectRelation.js#L171

Corepex:
@mattamon thanks for your PR :) - this truly improves pimcore 👍

PR #13415, opened 2022-10-19
Upgrade codeception version to 5

aryaantony92:
## Changes in this pull request  
Resolves #11664

PR #13546, opened 2022-11-09
BC layer for tests support classes

dvesh3:
Follow up to #13415

## Additional info  
Tests support classes are moved to new namepaces with codeception 5 directory structure change, so adding BC layer for old namespaces.


dvesh3:
closed in favor of #13564

PR #13549, opened 2022-11-10
remove install and plugin url prefix check

### LINKED ISSUE #13545: [Bug]: Site Requests with `install` url prefix cannot generate links correctly
### Expected behavior

Document Page with pretty-url /installateursuche should not be classified as backend request

### Actual behavior

Document Page with pretty-url /installateursuche is always classified  as backend request, because the request url starts with /install ( source of bug: https://github.com/pimcore/pimcore/blob/11.x/lib/Http/RequestHelper.php#L127-L145)

### Steps to reproduce

1. Create a Site /example
2. Create a Page /example/installateursuche
3. (Optional create a different page in /example)
4. visit http://example.com/installateursuche
5. each internal link will be shown as document path /example/internal-link instead of the correct /internal-link route

jheimbach:
Resolves #13545

### DIFF
--- a/lib/Http/RequestHelper.php
+++ b/lib/Http/RequestHelper.php
@@ -166,7 +166,7 @@ class RequestHelper
             return false;
         }
 
-        if (preg_match("/^\/admin.*/", $request->getRequestUri())) {
+        if (preg_match("@^/admin.*@", $request->getRequestUri())) {
             return false;
         }
 
--- a/lib/Http/RequestHelper.php
+++ b/lib/Http/RequestHelper.php
@@ -166,16 +166,8 @@ class RequestHelper
             return false;
         }
 
-        $excludePatterns = [
-            "/^\/admin.*/",
-            "/^\/install.*/",
-            "/^\/plugin.*/",
-        ];
-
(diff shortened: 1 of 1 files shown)


lib/Http/RequestHelper.php
```diff
@@ -166,16 +166,8 @@ private function detectFrontendRequest(Request $request): bool
...
-            "/^\/admin.*/",
-            "/^\/install.*/",
-            "/^\/plugin.*/",
-        ];
-
-        foreach ($excludePatterns as $pattern) {
-            if (preg_match($pattern, $request->getRequestUri())) {
-                return false;
-            }
+        if (preg_match("/^\/admin.*/", $request->getRequestUri())) {
```
jdreesen: I think we should use another delimiter here, e.g.
```suggestion
        if (preg_match("#^/admin.*#", $request->getRequestUri())) {
```

jheimbach: Changed it to `@` because `isFrontendRequestByAdmin` is also using some regex with `@` as a delimiter, so we are consistend in the same Class

aashan10 (approved):

PR #13551, opened 2022-11-10
Fix counting items in store while filtering

croitoruion:
Resolves #
The problem consist in fact that we can filter items and then getCount() in elementAlreadyExists() function will return 0 so we can avoid validation and add multiple items.
In first screenshot I reproduced this bug to be more explicit. In the second screenshot, I did some testing and you can see returns of counting.
My solution is based on this stackoverflow thread https://stackoverflow.com/questions/18658105/extjs-get-all-store-records

## Additional info  
image

image

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyRelation.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/advancedManyToManyRelation.js
@@ -744,7 +744,7 @@ pimcore.object.tags.advancedManyToManyRelation = Class.create(pimcore.object.tag
 
         // check max amount in field
         if (this.fieldConfig["maxItems"] && this.fieldConfig["maxItems"] >= 1) {
-            if (this.store.getCount() >= this.fieldConfig.maxItems) {
+            if ((this.store.getData().getSource() || this.store.getData()).count() >= this.fieldConfig.maxItems) {
                 Ext.Msg.alert(t("error"), t("limit_reached"));
                 return true;
             }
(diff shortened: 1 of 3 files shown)


Corepex:
@croitoruion thanks for reporting this bug. Could you please apply your change to the other relations as well? It seems that this bug affects nearly all other relations too. 

Affected Relations:

- ManyToManyRelation
- ManyToManyObjectRelation
- AdvancedManyToManyRelation (you already fixed this one)

thanks

Corepex:
I just found another way to bypass the "max. items" limit! - the "upload asset" button. 
Could you please copy the whole IF condition that checks the limit to the `uploadDialog()` function as well?
Just insert it right at the beginning of the function.

croitoruion:
> I just found another way to bypass the "max. items" limit! - the "upload asset" button. Could you please copy the whole IF condition that checks the limit to the `uploadDialog()` function as well? Just insert it right at the beginning of the function.

I already created PR for this scenario, but I need to update it with correct way of counting store items.
https://github.com/pimcore/pimcore/pull/13543

Corepex:
> I already created PR for this scenario, but I need to update it with correct way of counting store items. #13543

sorry, I must have missed that. 
Please add the updated "getCount" method to #13543 as well. 
After that we can merge both PRs 👍🥇

croitoruion:
> > I already created PR for this scenario, but I need to update it with correct way of counting store items. #13543
> 
> sorry, I must have missed that. Please add the updated "getCount" method to #13543 as well. After that we can merge both PRs 👍🥇

Updated

Corepex:
@croitoruion, very nice PR thanks for helping us to improve pimcore 👍

PR #13552, opened 2022-11-10
[Data objects] Support fulltext search for multiselect fields with tag mode

BlackbitDevs:
Steps to reproduce:
1. Create class with multiselect field, set Render Type=Tags
2. Add options `This is a test` and `another test`
3. Create object
4. Enter `test` into multiselect field

Without this PR: Nothing will get found
With this PR: Both options will get found

Reason: Currently multiselect fields with tag mode only support prefix search, for this reason it works when you search for `this` in above example (you will get `This is a test`). 
With this PR you get a fulltext search.

PS: Don't know if this is a bug or a new feature. If you think it makes more sense to fix this already in 10.5, please notify me, then I will adjust the PR...


robertSt7:

PR #13553, opened 2022-11-10
[Bug]: ReadOnly layout for Float number classification store key cause javascript exception "Layout run failed"

### LINKED ISSUE #13377: [Bug]: ReadOnly layout for Float number classification store key cause javascript exception "Layout run failed"
### Expected behavior

Class fields (classification store keys) should be shown.

### Actual behavior

Javascript error and fields are not shown

### Steps to reproduce

Using pimcore demo (PIM/MDM):
- Select "PIM" perspective
- Settings -> Data Objects -> Classification Store -> ProductAttributes -> change key definition 'height' from 'Input' to 'Number'. Edit its detail: decimal size: 2, decimal precision: 2 then click 'Apply'
- Settings-> Data Objects -> Classes -> Product Data -> AccessoryPart -> Configure Custom Layouts -> Add new layout called 'ReadOnly'
- Drag the root 'Layout' from Class to Base of Custom Layout
- Select 'Technical Attributes' and check 'Not Editable'
- Save then close Custom Layout Definition dialog
- Open object /Product Data/Accessories/windows/wind screens/austin-healey-3000 mk ii-windscreens
- Tab 'Structured Data' we will see Technical Data with Dimensions group that has 'Height', 'Width' and 'Length'
- Click Reload button and select 'ReadOnly' (which was created before)
- Now 'Dimensions' group will be empty because of javascript error 'Layout run failed' in browser's developer console

robertSt7:
resolves #13377

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/numeric.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/numeric.js
@@ -139,6 +139,8 @@ pimcore.object.tags.numeric = Class.create(pimcore.object.tags.abstract, {
         var input = {
             fieldLabel: this.fieldConfig.title,
             name: this.fieldConfig.name,
+            labelWidth: 100,
+            width: 175,
             componentCls: this.getWrapperClassNames(),
         };
 


Corepex:
@robertSt7 I noticed that choosing "table" instead of "number" also causes layout crashes

Corepex:
> @robertSt7 I noticed that choosing "table" instead of "number" also causes layout crashes

the table issue seems not to be related to "read-only" - will open a separate issue to fix the table

PR #13558, opened 2022-11-11
Update Namespace for Support

aryaantony92:
Follow up to #13415


tests/_bootstrap.php
```diff
@@ -48,7 +48,7 @@
 
 Autoloader::addNamespace('Pimcore\Model\DataObject', __DIR__ . '/_output/var/classes/DataObject');
 Autoloader::addNamespace('Pimcore\Tests', __DIR__);
-Autoloader::addNamespace('Pimcore\Tests', __DIR__ . '/Support');
+Autoloader::addNamespace('Pimcore\Tests\Support', __DIR__ . '/Support');
```
dvesh3:
```suggestion
```
I think we can just git rid of it since the line above adds the namespace and support folder now follows the Psr standards.

PR #13559, opened 2022-11-11
Bugfix: Get image Thumbnail Admin Action

TobiSW:
- Call getById from Asset instead of Asset\Image to not break class model overrides

## Steps to reproduce
- Overwrite the Pimcore\Model\Asset\Image in your config.yaml models -> class_overrides as described here:
https://pimcore.com/docs/pimcore/current/Development_Documentation/Extending_Pimcore/Overriding_Models.html
- Upload a new Asset in the DAM and click directly on the uploaded image (No reload)
- getImageThumbnailAction gets the image
- The image is an instance of Pimcore\Model\Asset\Image instead of your own overwritten class

  
  

Change how a image will be loaded in the imageThumbnailAction

## Additional info  
The Pimcore\Model\Asset getById method which is being called in the Pimcore\Bundle\AdminBundle\Controller\Admin\Asset\AssetContoller getImageThumbnailAction creates a new static instance.

https://github.com/pimcore/pimcore/blob/87ec1423edc77b5c78a05709146ecc092c055ce8/models/Asset.php#L250-L261

This will break class model overrides in this case. To avoid this the method should be called from the Asset itself instead of Asset\Image

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -1331,7 +1331,7 @@ class AssetController extends ElementControllerBase implements KernelControllerE
     public function getImageThumbnailAction(Request $request)
     {
         $fileinfo = $request->get('fileinfo');
-        $image = Asset\Image::getById((int)$request->get('id'));
+        $image = Asset::getById((int)$request->get('id'));
 
         if (!$image) {
             throw $this->createNotFoundException('Asset not found');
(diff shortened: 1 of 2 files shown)


mcop1:
@TobiSW could you please have a look at the failing tests? Thanks!

blankse:
@TobiSW Can you test my fix https://github.com/pimcore/pimcore/pull/13561 ?
This should be a better fix for this.

TobiSW:
> @TobiSW Can you test my fix #13561 ? This should be a better fix for this.

thanks very cool. I tested it out, your fix is also working.
I was also already pushing an update , which seems also to work. Which solution do you prefer?

blankse:
@TobiSW I prefer my fix.
With your change the Dao::getById() would be called always twice, although the right class is given.
My PR fix also the Document::getById() method.

TobiSW:
@blankse okay nvm lets go ahead with your one, thanks for bringing this one in :)
I will close this one and reference to your one https://github.com/pimcore/pimcore/pull/13561.

PR #13563, opened 2022-11-11
Bugfix(ClassificationStore): KeyConfig

christianbasilicom:
This PR fixes an error that occurs when creating or adding a KeyConfig because the cache is cleared on save without "testing" if the cache key exists.

## Reproduction
```php
// get a KeyConfig by name
$keyConfig = Classificationstore\KeyConfig::getByName($name, $storeId);
// make changes and save
$keyConfig->save();
// exception will be thrown
```

## Additional info  
None.


robertSt7:
@christianbasilicom Thanks

PR #13569, opened 2022-11-14
[Reverse relation] Support configuration of visible fields

BlackbitDevs:
For (advanced) many-to-many object relations you can configure the fields to be shown for the relation field grid in the editing panel.
This PR adds the same for reverse relations:

Снимок экрана 2022-11-14 в 16 00 57
Снимок экрана 2022-11-15 в 15 06 37

### DIFF
--- a/models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
+++ b/models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
@@ -28,7 +28,6 @@ class ReverseObjectRelation extends ManyToManyObjectRelation
 {
     /**
      * @internal
-     *
      */
     public ?string $ownerClassName = null;
 
@@ -73,7 +72,9 @@ class ReverseObjectRelation extends ManyToManyObjectRelation
                     return null;
                 }
                 $class = DataObject\ClassDefinition::getById($this->ownerClassId);
-                $this->ownerClassName = $class->getName();
+                if($class instanceof DataObject\ClassDefinition) {
+                    $this->ownerClassName = $class->getName();
+                }
             } catch (\Exception $e) {
                 Logger::error($e->getMessage());
             }
@@ -178,11 +179,6 @@ class ReverseObjectRelation extends ManyToManyObjectRelation
         return [];
     }
 
-    public function isOptimizedAdminLoading(): bool
-    {
-        return true;
-    }
-
     public function preGetData(mixed $container, array $params = []): array
     {
         return $this->load($container);
@@ -199,5 +195,15 @@ class ReverseObjectRelation extends ManyToManyObjectRelation
     public function getFieldType(): string
     {
         return 'reverseObjectRelation';
+    }
+    /**
+     * @inheritdoc
+     */
+    public function getClasses(): array
+    {
+        if($this->ownerClassId) {
+            return Model\Element\Service::fixAllowedTypes([$this->ownerClassId], 'classes');
+        }
+        return [];
     }
 }
--- a/models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
+++ b/models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
@@ -202,7 +202,7 @@ class ReverseObjectRelation extends ManyToManyObjectRelation
     public function getClasses(): array
     {
         if($this->ownerClassId) {
-            return Model\Element\Service::fixAllowedTypes([$this->ownerClassId], 'classes');
+            return Model\Element\Service::fixAllowedTypes([$this->ownerClassName], 'classes');
         }
         return [];
     }


Corepex:
@BlackbitDevs could you please resolve the conflicts? :) thanks

BlackbitDevs:
@Corepex Done.

Corepex:
@BlackbitDevs, could you also please merge the 11.x branch into your branch again - this should fix the test errors - thanks

BlackbitDevs:
@Corepex Done.

Corepex:
@BlackbitDevs could you please fix the tests 😺

brusch:
@BlackbitDevs 🏓😉 Just a quick reminder  -  Thanks!

BlackbitDevs:
@Corepex, @brusch Done.

Corepex:
@BlackbitDevs, thanks for this really cool PR. I really like the idea of customizing the columns of the reverse relation.

One weird thing I found is that if I save the class multiple times (10 - 20 times), I sometimes get the following error: `Typed property Pimcore\Model\DataObject\ClassDefinition\Data\ReverseObjectRelation::$ownerClassName must not be accessed before initialization`

Seems like this is called in `ReverseObjectRelation::getClasses()` and throws the error at `ReverseObjectRelation::getOwnerClassId` -> (`$class = DataObject\ClassDefinition::getByName($this->ownerClassName)`)

image

BlackbitDevs:
@Corepex Is fixed. It was caused by wrong property type in 
https://github.com/pimcore/pimcore/blob/46c8fe5c31e5cd78ebd6e6d0afcea55b4ea374ae/models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php#L43

When this shall be initialized with `null` (what it does when no default value is given and the property does not get set in the __constructor) the type has to be `?string`. For more information see https://madewithlove.com/blog/software-engineering/typed-property-must-not-be-accessed-before-initialization/#uninitialized-fields

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -40,7 +40,7 @@ class ReverseObjectRelation extends ManyToManyObjectRelation
      *
      * @var string
```
blankse:
```suggestion
```

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -40,7 +40,7 @@ class ReverseObjectRelation extends ManyToManyObjectRelation
...
      */
-    public string $ownerClassName;
+    public ?string $ownerClassName;
```
blankse:
```suggestion
    public ?string $ownerClassName = null;
```

blankse (commented):
We should init it with null otherwise it will occure a error when you read it without writing

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
BlackbitDevs: Why? Isn't a defined property without value implicitly set to `null` automatically?

blankse: No it isn't

jdreesen: https://3v4l.org/tPdCJ

jdreesen: Untyped properties are `null` if not initialized, whereas typed properties are "not initialized".

BlackbitDevs: Ahhh, thanks for explanation, will adjust the code.

Corepex:
@BlackbitDevs is that PR ready for testing again?

BlackbitDevs:
@Corepex Yes

Corepex:
@BlackbitDevs now i get another error when saving the class multiple times. The following error message gets displayed: `Pimcore\Model\DataObject\ClassDefinition::getByName(): Argument #1 ($name) must be of type string, null given, called in /var/www/html/dev/pimcore/pimcore/models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php on line 99` (see screenshot).


image

It seems that the `owner_class` field gets cleared for some reason after saving ... 

image

Corepex:
@BlackbitDevs friendly reminder ⏰ 🍭

BlackbitDevs:
Ok, wrapped another if condition. Can you please try again, @Corepex?

Corepex:
@BlackbitDevs, I tested the PR again. The issue that occurs when saving multiple times seems to be fixed. 

I found another issue that throws an error. 

image


image


https://github.com/pimcore/pimcore/blob/19ef4b24b8bc8151fd000f07016da5269b54cb9a/bundles/SimpleBackendSearchBundle/src/Controller/DataObjectController.php#L99

**Steps to reproduce: (in the demo)**
1. go to the `car` data object and add the `cars` field in the `categories` relation to the visible fields
2. switch the `display mode` from `grid view` to `tag field`
3. save the class definition
4. open/reload a car (for example id 81) -> you should see the mentioned error now

Corepex:
> It seems that the `owner_class` field gets cleared for some reason after saving ...

@BlackbitDevs, this error doesn't seem to be fixed - the `owner_class` field is still getting cleared out sometimes.

Corepex:
@BlackbitDevs final reminder. ⏰  
Please respond latest by 24.03 - otherwise, I will close the ticket due to inactivity.

BlackbitDevs:
@Corepex Please check again. I cannot reproduce the issue. I clicked multiple times on save - with set owner class as well as without. Everything works as expected.

Corepex:
@BlackbitDevs I just set up a new pimcore 11.x instance. The `owner_class` issue seems to be fixed now - good job 🎉 

Sadly I found some new bugs that need to be fixed. 

1) Empty Columns
If I add a new entry to the relation it seems that some of the columns are empty ... maybe you could find a way to load the column data when adding the object to the relation.

image

2) Missing scrollbar
I also tested the option to add an image field to the grid. 
There we should add a scrollbar to the relation field because it only shows some (not all) entries.

image

I used the `Category` object from the demo and added the following fields:
- fullpath
- key
- published
- modificationDate
- filename
- name
- location
- (genericImages)

BlackbitDevs:
@Corepex Fixed loading the visible field data after adding via drag & drop or search window.

The height issue was caused by https://github.com/pimcore/pimcore/blob/24d8ca42c67b07b45f00bac2a9b0af13ee27af92/bundles/AdminBundle/public/js/pimcore/object/tags/manyToManyObjectRelation.js#L963-L967
This was introduced in #5378 (after I suggested a similar solution in #4337)
For fields which do not use `optimizedAdminLoading`, the field data already gets fetched by `loadObjectData`, for example  in https://github.com/pimcore/pimcore/blob/24d8ca42c67b07b45f00bac2a9b0af13ee27af92/bundles/AdminBundle/public/js/pimcore/object/tags/manyToManyObjectRelation.js#L618
For those fields `fields`stays empty in https://github.com/pimcore/pimcore/blob/24d8ca42c67b07b45f00bac2a9b0af13ee27af92/bundles/AdminBundle/public/js/pimcore/object/tags/manyToManyObjectRelation.js#L921-L934
and thus https://github.com/pimcore/pimcore/blob/24d8ca42c67b07b45f00bac2a9b0af13ee27af92/bundles/AdminBundle/public/js/pimcore/object/tags/manyToManyObjectRelation.js#L937 returns `nicePathRequested = undefined`
and thus we did not get to
https://github.com/pimcore/pimcore/blob/24d8ca42c67b07b45f00bac2a9b0af13ee27af92/bundles/AdminBundle/public/js/pimcore/object/tags/manyToManyObjectRelation.js#L963-L968
And for this reason the height was not adjusted. The same problem existed for normal many-to-many object relations with a visible `image` field: After adding something via drag & drop, the image is cut off.

Now it should work.

robertSt7:
@BlackbitDevs Could you please resolve the conflicts? Thanks

BlackbitDevs:
@robertSt7 As the JS files are now part of https://github.com/pimcore/admin-ui-classic-bundle I have splitted this PR into 2. This one contains the core changes and https://github.com/pimcore/admin-ui-classic-bundle/pull/290 contains the UI changes.

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -178,11 +179,6 @@ public function resolveDependencies(mixed $data): array
         return [];
     }
 
-    public function isOptimizedAdminLoading(): bool
-    {
-        return true;
-    }
```
robertSt7, line 184: @BlackbitDevs I think we cannot remove this, because it's not internal

BlackbitDevs: Actually it is not removed but is now gets inherited from the parent class (with default value `false`). If you want we can set this explicitly to `false` in `ReverseObjectRelation` but the whole purpose of this PR is to make reverse relations work the same as usual many-to-many object relations so I think inheriting from `ManyToManyObjectRelation` is better.

robertSt7: @BlackbitDevs No it's fine. Thanks for the explanation

models/DataObject/ClassDefinition/Data/ReverseObjectRelation.php
```diff
@@ -200,4 +196,14 @@ public function getFieldType(): string
     {
         return 'reverseObjectRelation';
     }
+    /**
+     * @inheritdoc
+     */
+    public function getClasses(): array
+    {
+        if($this->ownerClassId) {
+            return Model\Element\Service::fixAllowedTypes([$this->ownerClassId], 'classes');
```
dvesh3:
```suggestion
            return Model\Element\Service::fixAllowedTypes([$this->ownerClassName], 'classes');
```
Class IDs can be alphanumeric. Otherwise it will try to load the class with name by passing id https://github.com/pimcore/pimcore/blob/e22a8209fde57c953a5caba13292d60ae3b06979/models/DataObject/ClassDefinition/Data/ManyToManyObjectRelation.php#L438

dvesh3:
@robertSt7 ready for merge?

robertSt7:

PR #13573, opened 2022-11-14
[Search] Search within a word

BlackbitDevs:
The backend search uses MySQL's fulltext search in boolean mode. By default the search is done exactly as requested - assuming the user knows the technical details about [available search operators](https://dev.mysql.com/doc/refman/8.0/en/fulltext-boolean.html). Often this is not the case as users are domain experts but not necessarily IT experts.

This leads to the problem that when somebody has a product "Ring with 5 diamonds" and searches for "diamond", he will not find it because MySQL does an exact search on a word basis and the word "diamond" does not appear on the product (assuming it neither does appear on any other data object field of this object).

With this PR the search will automatically append `*` to search terms and thus above object will get found. To execute an exact search, you can use a quoted string like `"diamond"` to only find objects which contain the word `diamond` but not `diamonds`. This behaviour is more convenient to users and it works like on most other applications like Google (where partial matching is default and only if quotes get used, the exact search term gets used).

PS: There is no problem with performance as prefix searches are covered with the index in the same way as exact searches.

### DIFF
--- a/bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
+++ b/bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
@@ -421,22 +421,22 @@ class SearchController extends AdminController
      */
     protected function filterQueryParam(string $query)
     {
-        if ($query == '*') {
+        if ($query === '*') {
             $query = '';
         }
 
-        $query = str_replace('&quot;', '"', $query);
-        $query = str_replace('%', '*', $query);
-        $query = str_replace('@', '#', $query);
+        $query = str_replace(['&quot;', '%', '@'], ['"', '*', '#'], $query);
         $query = preg_replace("@([^ ])\-@", '$1 ', $query);
 
         $query = str_replace(['<', '>', '(', ')', '~'], ' ', $query);
 
         // it is not allowed to have * behind another *
-        $query = preg_replace('#[*]+#', '*', $query);
+        $query = preg_replace('#\*+#', '*', $query);
 
         // no boolean operators at the end of the query
         $query = rtrim($query, '+- ');
+
+        $query = preg_replace('/([\p{L}\p{Nd}_]+)([\s\)]|$)/u', '$1*$2', $query);
 
         return $query;
     }
--- a/bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
+++ b/bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
@@ -430,6 +430,7 @@ class SearchController extends AdminController
 
         $query = str_replace(['<', '>', '(', ')', '~'], ' ', $query);
 
+        // expand search terms with * to get prefix search. Terms within quotes do not get expanded to allow for exact searches
         $query = preg_replace('/(?!\B"[^"]*)([\p{L}\p{Nd}_]+)(?![^"]*"\B)/u', '$1*', $query);
 
         // it is not allowed to have * behind another *
--- a/bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
+++ b/bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
@@ -430,22 +430,13 @@ class SearchController extends AdminController
 
         $query = str_replace(['<', '>', '(', ')', '~'], ' ', $query);
 
+        $query = preg_replace('/(?!\B"[^"]*)([\p{L}\p{Nd}_]+)(?![^"]*"\B)/u', '$1*', $query);
+
         // it is not allowed to have * behind another *
         $query = preg_replace('#\*+#', '*', $query);
 
         // no boolean operators at the end of the query
         $query = rtrim($query, '+- ');
-
-        $terms = str_getcsv($query, ' ');
-        $fulltextSearchTerms = array_map(static function ($term) {
-            if(str_contains($term, ' ')) {
-                // term was wrapped in quotes -> do not change
-                return '"'.$term.'"';
-            }
-
-            return preg_replace('/([\p{L}\p{Nd}_]+)([\s\)]|$)/u', '$1*$2', $term);
-        }, $terms);
-        $query = implode(' ', $fulltextSearchTerms);
 
         return $query;
     }
--- a/bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
+++ b/bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
@@ -436,7 +436,16 @@ class SearchController extends AdminController
         // no boolean operators at the end of the query
         $query = rtrim($query, '+- ');
 
-        $query = preg_replace('/([\p{L}\p{Nd}_]+)([\s\)]|$)/u', '$1*$2', $query);
+        $terms = str_getcsv($query, ' ');
+        $fulltextSearchTerms = array_map(static function ($term) {
+            if(str_contains($term, ' ')) {
+                // term was wrapped in quotes -> do not change
+                return '"'.$term.'"';
+            }
+
+            return preg_replace('/([\p{L}\p{Nd}_]+)([\s\)]|$)/u', '$1*$2', $term);
+        }, $terms);
+        $query = implode(' ', $fulltextSearchTerms);
 
         return $query;
     }


fashxp:
I can see the use case, but I'm not so sure about that change. When I want to search JUST for "diamond", I will now get 'wrong' results and I have no option to avoid that.
On the other hand ... I always have the option to add a * manually.

BlackbitDevs:
@fashxp I adjusted the implementation so that a quotes string like `"diamond"` can be used for exact searches and a string without quotes `diamond` for prefix search.

bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
```diff
@@ -421,19 +421,19 @@ protected function getPermittedPaths($types = ['asset', 'document', 'object'])
...
 
-        $query = str_replace('&quot;', '"', $query);
-        $query = str_replace('%', '*', $query);
-        $query = str_replace('@', '#', $query);
+        $query = str_replace(['&quot;', '%', '@'], ['"', '*', '#'], $query);
         $query = preg_replace("@([^ ])\-@", '$1 ', $query);
 
         $query = str_replace(['<', '>', '(', ')', '~'], ' ', $query);
 
+        $query = preg_replace('/(?!\B"[^"]*)([\p{L}\p{Nd}_]+)(?![^"]*"\B)/u', '$1*', $query);
```
jdreesen, line 434: I think a comment on what this does would be nice, since it's not really self-explanatory :wink:

NiklasBr: https://www.youtube.com/watch?v=SWEphAZynYo

brusch:
I agree with @fashxp - I'm not sure if this leads to more questions than it helps. At the moment you have the full power of the binary search, without restrictions. I haven't "deconstructed" the regex yet, but you know it's most likely that the regex doesn't cover 100% of what was possible before. 

I'd leave it as it is to be honest.

bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php
BlackbitDevs: I added a comment ;-)

BlackbitDevs:
Please give one example of what is not possible with this approach. 
The backend search is really a problem for almost all users! The more experienced ones get used to the operators of course but the "average" people just wonder why a search for "computer" will not find the product "one of our best computers".

SamyMP:
I underdstand the reasons why @BlackbitDevs submitted this pull request, I think that from a user perspective, it can be confusing to see a search bar, to type a word (any word in singular for example), and not find a product which contains this word in plural. Most users will expect the search to work just like a google search, or something similar

Maybe something can be done to satisfy both sides ?

BlackbitDevs:
> Maybe something can be done to satisfy both sides ?

From @fashxp and @brusch there was the argument that some queries now might not be possible anymore. This refusing mindset is hard to tackle because I cannot prove that everything will work. Of course I can fix the tests and probably add some others but I have to know if the approach of changing the search term in general has a chance to succeed. Otherwise I will simply write an event listener for `AdminEvents::SEARCH_LIST_BEFORE_FILTER_PREPARE` in the app which does the same logic. Or maybe a plugin because I know of a lot of Pimcore users who are not satisfied with the backend search... but actually solving a problem at root would be the better thing. So, @fashxp  and @brusch please comment if you generally are against the approach of changing the search term.

SamyMP:
Another idea would be to let the user choose, like adding a checkbox next to the search bar (or admin via the admin configs), and put a condition in the code to switch the search mode accordingly, like here : 
https://github.com/pimcore/pimcore/blob/23b66280576675381569d7e588668c78d94d0c16/bundles/AdminBundle/src/Controller/Searchadmin/SearchController.php#L468

Changing `IN BOOLEAN MODE` to another search mode for example.

This way, those who prefer the boolean mode can still choose to use it, and the others can have a search mode more permissive. What do you think ?

brusch:
@SamyMP The user can choose already, you just need to use `* ` in your query, similar to many other search solutions incl. Google 😉

SamyMP:
I mean, they'd first have to know how the pimcore backend works and that it's using a boolean search mode, and then be aware that there are special search operators to get the result they want 😅
It's not said anywhere (appart from the code) that it's using such a search mode, and the operators can only be found in the MySQL/MariaDB doc, which is technical and shouldn't serve as a knowledge base for non-technical users... You'll say it's a documentation problem, but then again, I think this pull request is raising a point and it deserved to be discussed with proper pros/cons arguments, and maybe a solution that doesn't involve the users to manually add a `*` at the end of every query

There might still be a solution that solves this problem at the root and that satisfies everyone though. The solution @BlackbitDevs proposed seems pretty clean to me

brusch:
Let's leave this open for now and keep the discussion going - I've shared my concerns, but as always we're open for feedback and implement it accordingly 😊

brusch:
Converted to "Draft" to indicate it's dangling.

Jan-Walther-Blackbit:
@brusch What do you mean with "Draft"? What shall I do to finish it? In my opinion it is finished (except for the tests - but it does not make sense to adjust the tests if this concept will not be accepted anyway). Do you think the search currently works good? Or do you have something in the pipeline to enhance this anyway which would make this PR obsolete?

BlackbitDevs:
@brusch What do you mean with "Draft"? What shall I do to finish it? In my opinion it is finished (except for the tests - but it does not make sense to adjust the tests if this concept will not be accepted anyway). You and @fashxp only said that it may be useful but some queries might not be possible - without any example.
Do you think the search currently works good? Or do you have something in the pipeline to enhance this anyway which would make this PR obsolete?

Making this a draft does not make sense. Either say this is a good idea (if some modifications are necessary, I am perfectly fine with adjusting it) or say it is a bad idea, then I will create a plugin for this...

fashxp:
> I can see the use case, but I'm not so sure about that change. When I want to search JUST for "diamond", I will now get 'wrong' results and I have no option to avoid that. On the other hand ... I always have the option to add a * manually.

Here is one example. Another example would be searching for IDs (when I search for 123 it also will find 1231 etc.). As stated, I can see your use case, but I have respect of changing that behavior by default, since other users are used to the behavior right now. 

Besides that, to be honest, I don't understand that regex and I'm afraid of maintaining or even debugging it. And following the comments I'm not alone with that view. 

Maybe an approach would be to allow the user to choose what 'search-mode' it wants to have. Similar to what phpstorm does: 

image

Default behavior stays as it is (would be word mode in phpstorm), but user always can change to wildcard mode. And the users selection is also stored in local storage or something so that it remembers when starting another search. 

WDYT?

BlackbitDevs:
> Besides that, to be honest, I don't understand that regex and I'm afraid of maintaining or even debugging it. And following the comments I'm not alone with that view.

Me too, I just copied it from https://stackoverflow.com/a/21106122 ;-)
Of course we could also use a loop or a more readable regex.

> Another example would be searching for IDs (when I search for 123 it also will find 1231 etc.)

Right, of course we can change the logic to not apply this for numerical terms.

> Maybe an approach would be to allow the user to choose what 'search-mode' it wants to have

In general I like this suggestion. But we would have to add this button to all the search fields, e.g. searching for a relation. But this should be feasible.

NiklasBr:
> Default behavior stays as it is (would be word mode in phpstorm), but user always can change to wildcard mode. And the users selection is also stored in local storage or something so that it remembers when starting another search.

I am currently setting up a new computer out of the box today.

I just installed PHP Storm and the default behaviour does not match what you describe. The "W" is not active.
Default search in Firefox is also to allow searching for parts of words ("hole words" remains unchecked).
Default behaviour in MacOS System Search (Spotlight) allows searches by substrings.
In Outlook v16.67 the search also matches substrings by default.

fashxp:
> I just installed PHP Storm and the default behaviour does not match what you describe. The "W" is not active.
> Default search in Firefox is also to allow searching for parts of words ("hole words" remains unchecked).
> Default behaviour in MacOS System Search (Spotlight) allows searches by substrings.
> In Outlook v16.67 the search also matches substrings by default.

Never said it is default behavior in phpstorm. I just said, default behavior stays as it is, which would be word mode in phpstorm.

> But we would have to add this button to all the search fields, e.g. searching for a relation. But this should be feasible.

That PR also would change search behavior for all search dialogs.

Jan-Walther-Basilicom:
> That PR also would change search behavior for all search dialogs.

Of course (but the change is only in one place). Only wanted to say that we have to add this search mode button in multiple places - but as said, this no problem.

BlackbitDevs:
@fashxp Answering to https://github.com/pimcore/pimcore/issues/13689#issuecomment-1329185831 here to keep https://github.com/pimcore/pimcore/issues/13689 focusing on the new approach.
The problem with the "search mode" button is that it would have the same UX problems as search operators. Non-technical users simply do not know what a "wildcard" mode is. And even if they knew, there are places where an additional button is somehow strange - for example with the direct input relations which got introduced in #13227 and #13676.

Imho a search form needs exactly one input field. If the user does not find what he was looking for, it is bad UX.

Another problem with this PR's approach is chinese / japanese: I was not aware that those languages do not separate words with whitespaces or anything similar. And we are supposed to implement this feature for a large client whose Pimcore is mainly used for chinese, japanese and other eastern languages.

PR #13599, opened 2022-11-16
[BUG] Object Grid: Filter for Multiselect field with OptionProvider

storbahn:
Object Grid: Filter for MultiSelect field with OptionProvider (dynamicOptions = true) is not working.

Result: SQL Query Error 

Object_Grid_Error_MultiSelect_With_Provider

@kjkooistra-youwe: If I remove this code block of your PR https://github.com/pimcore/pimcore/pull/12069 then it will work (for MultiSelect field with OptionProvider) but which influence does it have on your implementation?

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/multiselect.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/multiselect.js
@@ -81,13 +81,6 @@ pimcore.object.tags.multiselect = Class.create(pimcore.object.tags.abstract, {
     },
 
     getGridColumnFilter: function(field) {
-        if (field.layout.dynamicOptions) {
-            return {
-                type: 'string',
-                dataIndex: field.key
-            };
-        }
-
         var storeData = this.prepareStoreDataAndFilterLabels(field.layout);
 
         var store = Ext.create('Ext.data.JsonStore', {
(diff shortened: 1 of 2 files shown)


kjkooistra-youwe:
> @kjkooistra-youwe: If I remove this code block of your PR #12069 then it will work (for MultiSelect field with OptionProvider) but which influence does it have on your implementation?

Because it was broken for multiselect and working for the regular select I think I copied it from https://github.com/kjkooistra-youwe/pimcore/blob/10.x/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/select.js#L206
Possibly the query generation logic changed? Since I don't recall any issue like this.

I don't think it should be possible to filter dynamic selects, since the system can never detect the available options to filter by. Or is it a plain text filter?

storbahn:
It depends on the option provider. In our case we use an option provider which read the options from a csv file. For this case it should be possible.

storbahn:
Hi @kjkooistra-youwe,
I updated the PR, could you please check it too. Now it should work for both use cases or?

kjkooistra-youwe:
The options are still loaded and can be edited. I can also use the filter without an error.

Of course, you can only filter by 1 option value this way, but I guess that's better than nothing.

robertSt7:
@storbahn Thanks for the PR

PR #13605, opened 2022-11-17
[Users / Roles] Add permissions defining for field collections, object bricks, classification stores and quantity values

BlackbitDevs:
Currently there is the permission `Classes`. Users who have this are allowed to change almost anything for the data model: classes, object bricks, field collections, classification store, quantity value units etc.

We often put technical data of products into object bricks or classification stores. From time to time the product experts want to add a new technical data field. In this case currently there are two options:
1. Ask administrator to add the field -> may cause some delay
2. Give product experts permission for `Classes` -> risk of field changes at class level

So in our opinion it would a good idea to have more granular control over the data model permissions. For this reason this PR introduces the permission `Object bricks`. Users who have this permission, are allowed to change object brick fields but no classes.

There is a migration so that all users who had the `Classes` permission before, afterwards get `Classes` and `Object bricks` permission.


mattamon (changes requested):
Thank you for your contribution.
We do appreciate your effort making the permissions more granular.

Could you please extend  these permissions for all types like field-collections, quantity value, and classification store since it would make more sense than just doing it for the object bricks. 

It would also be nice to make a category out of these data object permissions so the main permissions category gets a bit decluttered as well.

You just need to add the category in the user_permission_definitons table like e.g. for the Datahub.

bundles/CoreBundle/src/Migrations/Version20221116115427.php
```diff
@@ -0,0 +1,33 @@
...
+    public function getDescription(): string
+    {
+        return 'Add "object bricks" permission';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql("INSERT INTO `users_permission_definitions` (`key`, `category`) VALUES ('objectbricks', '')");
+
+        $this->addSql('UPDATE `users` SET `permissions`=CONCAT(`permissions`, \',objectbricks\') WHERE `permissions` REGEXP \'(?:^|,)classes(?:^|,)\'');
```
mattamon: Unfortunately this regex does not cover all cases.
Since the user could only have the classes permissions, the permissions text would look like this
`classes` without any commas.
It could also be the first permission or the last permission:
`classes,translations` first
`assets,classes` last

bundles/CoreBundle/src/Migrations/Version20221116115427.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)objectbricks(?:^|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)objectbricks(?:^|,)\'');
```
mattamon: Same with removing the permissions again.
```suggestion
        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:|,)objectbricks(?:|,)\', \'\') WHERE `permissions` REGEXP \'(?:|,)objectbricks(?:|,)\'');
```

BlackbitDevs: The `(?:^|,)` at the end should have been `(?:$|,)` which means "end of string or comma". Have fixed it.

bundles/CoreBundle/src/Migrations/Version20221116115427.php
BlackbitDevs: The (?:^|,) at the end should have been (?:$|,) which means "end of string or comma". Have fixed it.

mattamon:
Hey @BlackbitDevs did you have a chance to review my comment? :)

BlackbitDevs:
@mattamon Now there are separate permissions for object bricks, field collections, quantity values and classification store.

models/DataObject/Classificationstore/KeyConfig.php
```diff
@@ -54,7 +54,7 @@ final class KeyConfig extends Model\AbstractModel
      *
      * @var string
```
blankse:
```suggestion
```

models/DataObject/Classificationstore/KeyConfig.php
```diff
@@ -173,7 +173,7 @@ public function getName(): string
      *
      * @return string
```
blankse:
```suggestion
```

models/DataObject/Classificationstore/KeyConfig.php
```diff
@@ -185,7 +185,7 @@ public function getDescription(): string
      *
      * @return Model\DataObject\Classificationstore\KeyConfig
```
blankse:
```suggestion
     * @return $this
```

blankse (commented):
Fix return php docs

bundles/AdminBundle/src/Controller/Admin/DataObject/ClassController.php
```diff
@@ -1723,47 +1739,53 @@ public function bulkExportAction(Request $request): JsonResponse
             ];
         }
 
-        $classes = new DataObject\ClassDefinition\Listing();
-        $classes->setOrder('ASC');
-        $classes->setOrderKey('id');
-        $classes = $classes->load();
+        if($this->getAdminUser()->isAllowed('classes')) {
```
mattamon, line 1744: Why would we need this check?
Admin User always has all permissions?

What about the Fieldcollections above? Why would the check not be there but for classes and object bricks?

bundles/AdminBundle/src/Controller/Admin/DataObject/ClassController.php
```diff
@@ -1784,25 +1806,25 @@ public function doBulkExportAction(Request $request): Response
         $result = [];
 
         foreach ($list as $item) {
-            if ($item['type'] == 'fieldcollection') {
+            if ($item['type'] === 'fieldcollection' && $this->getAdminUser()->isAllowed('fieldcollections')) {
```
mattamon, line 1811: Like above, the admin user always has the permission

bundles/CoreBundle/src/Migrations/Version20221116115427.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql("INSERT INTO `users_permission_definitions` (`key`, `category`) VALUES ('objectbricks', 'Data Objects')");
+
+
+        $this->addSql("INSERT INTO `users_permission_definitions` (`key`, `category`) VALUES ('fieldcollections', 'Data Objects')");
+
+
+        $this->addSql("INSERT INTO `users_permission_definitions` (`key`, `category`) VALUES ('quantityValue', 'Data Objects')");
```
mattamon: Maybe make it more conistent, its always the plural but for the quantityValue it is singular.
Then we have to change the Translation key too.

mattamon (changes requested):

models/DataObject/Classificationstore/KeyConfig.php
```diff
@@ -183,9 +179,9 @@ public function getDescription(): string
      *
      * @param string $description
      *
```
blankse:
```suggestion
```

bundles/AdminBundle/src/Controller/Admin/DataObject/ClassController.php
BlackbitDevs: ? `$this->getAdminUser()` has nothing to do with admin users. It just returns the `\Pimcore\Model\User` object of the currently logged user - and this can or cannot have the permission `fieldcollections`.

bundles/AdminBundle/src/Controller/Admin/DataObject/ClassController.php
BlackbitDevs: ? `$this->getAdminUser()` has nothing to do with admin users. It just returns the `\Pimcore\Model\User` object of the currently logged user - and this can or cannot have the permission `classes`.

mattamon (approved):
LGTM

Corepex:
@BlackbitDevs, could you please resolve the merge conflicts 👍 thanks

BlackbitDevs:
@Corepex Done.

bundles/CoreBundle/src/Migrations/Version20221116115427.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $this->addSql("INSERT INTO `users_permission_definitions` (`key`, `category`) VALUES ('classificationstore', 'Data Objects')");
+
+
+        $this->addSql('UPDATE `users` SET `permissions`=CONCAT(`permissions`, \',objectbricks,fieldcollections,quantityValueUnits,classificationstore\') WHERE `permissions` REGEXP \'(?:^|,)classes(?:$|,)\'');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)objectbricks(?:^|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)objectbricks(?:$|,)\'');
```
mattamon:
```suggestion
        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)objectbricks(?:$|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)objectbricks(?:$|,)\'');
```
Shouldn't this be the same as the WHERE condition?

BlackbitDevs: Completely right!

bundles/CoreBundle/src/Migrations/Version20221116115427.php
```diff
@@ -0,0 +1,49 @@
...
+        $this->addSql("INSERT INTO `users_permission_definitions` (`key`, `category`) VALUES ('classificationstore', 'Data Objects')");
+
+
+        $this->addSql('UPDATE `users` SET `permissions`=CONCAT(`permissions`, \',objectbricks,fieldcollections,quantityValueUnits,classificationstore\') WHERE `permissions` REGEXP \'(?:^|,)classes(?:$|,)\'');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)objectbricks(?:$|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)objectbricks(?:$|,)\'');
+        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)fieldcollections(?:^|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)fieldcollections(?:$|,)\'');
```
mattamon:
```suggestion
        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)fieldcollections(?:$|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)fieldcollections(?:$|,)\'');
```

bundles/CoreBundle/src/Migrations/Version20221116115427.php
```diff
@@ -0,0 +1,49 @@
...
+        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)quantityValueUnits(?:^|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)quantityValueUnits(?:$|,)\'');
```
mattamon:
```suggestion
        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)quantityValueUnits(?:$|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)quantityValueUnits(?:$|,)\'');
```

bundles/CoreBundle/src/Migrations/Version20221116115427.php
```diff
@@ -0,0 +1,49 @@
...
+        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)classificationstore(?:^|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)classificationstore(?:$|,)\'');
```
mattamon:
```suggestion
        $this->addSql('UPDATE `users` SET `permissions`=REGEXP_REPLACE(`permissions`, \'(?:^|,)classificationstore(?:$|,)\', \'\') WHERE `permissions` REGEXP \'(?:^|,)classificationstore(?:$|,)\'');
```

bundles/CoreBundle/src/Migrations/Version20221116115427.php
mattamon: Sorry I didn't mark all the places, but there are some more. I add new comments.

bundles/AdminBundle/public/js/pimcore/object/classes/data/classificationstore.js
```diff
@@ -40,7 +40,7 @@ pimcore.object.classes.data.classificationstore = Class.create(pimcore.object.cl
     },
 
     getTypeName: function () {
-        return t("classificationstore");
+        return t("classification_store");
```
Corepex: why is this line changed?

BlackbitDevs: Because this translation is used in https://github.com/pimcore/pimcore/blob/0c15abf9691589c77fc488a76091ac163fe6d705/bundles/AdminBundle/public/js/pimcore/layout/toolbar.js#L626 and https://github.com/pimcore/pimcore/blob/0c15abf9691589c77fc488a76091ac163fe6d705/bundles/AdminBundle/public/js/pimcore/object/classificationstore/storeTree.js#L32

We do not need both imho.

Corepex: It seems that both versions exist ...

https://github.com/pimcore/pimcore/blob/c8f29107a6a39f29165bd909248b32a5a83e7ee6/bundles/AdminBundle/translations/admin_ext.en.yaml#L665

and 

https://github.com/pimcore/pimcore/blob/c8f29107a6a39f29165bd909248b32a5a83e7ee6/bundles/AdminBundle/translations/admin_ext.en.yaml#L732

I checked the occurences in the code and I would suggest going with `classificationstore` because this version is used more often. According to that, we could also delete the `classification_store` translations.

Corepex: WDYT?

BlackbitDevs: Ok, reverted to `classification_store`.
Changing the other translations which use `classification_store` should be done in separate PR.

Corepex:
@BlackbitDevs, thanks for your PR. Nice improvement. I think there are many people out there who will love the more specific permissions. 

It would be great if you could create a follow-up PR for the translations (as mentioned here: https://github.com/pimcore/pimcore/pull/13605#discussion_r1115946221).

PR #13607, opened 2022-11-17
[Logger] Added examples and reference to logging in the docs

aashan10:

### DIFF
--- a/doc/Development_Documentation/19_Development_Tools_and_Details/07_Logging.md
+++ b/doc/Development_Documentation/19_Development_Tools_and_Details/07_Logging.md
@@ -48,10 +48,6 @@ In this log you can see every request where a redirect takes action.
 ```
 
 ## Writing your own log files
-You can add your own logging functionality using Pimcore's log writer. You can call a static 
-function like this:
-
-##### Custom log entry
 To create a custom log entry, we need to create the monolog log channels and log handlers configuration. Here is an example on how to log in a custom filename called `custom.log`
 
 ```yaml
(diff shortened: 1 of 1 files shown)


kingjia90:
Could you please rebase for 10.5?

aashan10:
@kingjia90 done! :)

doc/Development_Documentation/19_Development_Tools_and_Details/07_Logging.md
```diff
@@ -52,14 +52,30 @@ You can add your own logging functionality using Pimcore's log writer. You can c
...
+    handlers:
+        custom_handler:
+            level:    debug
+            type:     stream
+            path:     '%kernel.logs_dir%/custom.log'
+            channels: [custom_log]
+    channels: [custom_log, some_other_channel]
+
 ```
+Now, when injecting the `Psr\Log\LoggerInterface` instead of using the name `$logger` like in previous examples, something like `$customLogLogger` needs to be used. Symfony will automatically figure out which channel to log into by the name of the variable.
```
kingjia90:
```suggestion
It is possible to inject the `Psr\Log\LoggerInterface` by changing the variable name eg. `$customLogLogger` (camel case channel name + `Logger`) and Symfony will automatically wire the specified channel.
```

suggesting some small changes, there's no other "previous examples" on the same page, it could be that it "landed" on this page first etc...

kingjia90 (changes requested):
This part needs to be removed as it still mentions about a static method that got removed in this PR
https://github.com/pimcore/pimcore/blob/64b636f1e989e52c096de635e4bb30f095eaf53a/doc/Development_Documentation/19_Development_Tools_and_Details/07_Logging.md?plain=1#L51-L54

PR #13609, opened 2022-11-17
[Pimcore admin] Prevent error if saved object does not exist anymore

BlackbitDevs:
Actually I do not know how to reproduce the error but we saw some entries like this in the logs:
> request.CRITICAL: Uncaught PHP Exception TypeError: "Pimcore\Bundle\AdminBundle\Controller\Admin\DataObject\DataObjectController::getLatestVersion(): Argument pimcore/pimcore#1 ($object) must be of type Pimcore\Model\DataObject\Concrete, null given, called in /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php on line 1280" at /var/www/html/vendor/pimcore/pimcore/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php line 2090

This PR prevents this error.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectController.php
@@ -1276,6 +1276,10 @@ class DataObjectController extends ElementControllerBase implements KernelContro
     {
         $objectFromDatabase = DataObject\Concrete::getById((int) $request->get('id'));
 
+        if(!$objectFromDatabase instanceof DataObject\Concrete) {
+            return $this->adminJson(['success' => false, 'message' => 'Could not find object']);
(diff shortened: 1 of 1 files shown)


Corepex:
@BlackbitDevs, thanks for creating this PR. 

To reproduce this error I opened a data object in the admin UI and removed it via the API - after that you can try to save the data object - now you will get the `DataObjectController::getLatestVersion()` error.

I created a follow-up issue to improve the version reload behavior. 
follow-up: pimcore/admin-ui-classic-bundle#272

BlackbitDevs:
@Corepex Perfect - but I think you linked the wrong follow-up. pimcore/pimcore#3669 is years old ;-)

PR #13613, opened 2022-11-17
Introduced Db\Helper::upsert() which avoids deadlocks

### LINKED ISSUE #13507: [Bug]: backend search update fails in high concurrency situations
### Expected behavior

Search backend updates work.

### Actual behavior

We're seeing hundreds of deadlock messages and our workers constantly crashing.

### Steps to reproduce

Have a high rate of update (say, import 30k objects) and multiple workers (we use 4). The rate of deadlocks is very high which ends up crashing our workers (which rely on DBAL based messenger).

Looking at the code which saves, I can see it's doing a lot of weird work, catching deadlock and then trying to do it at random (?).

The issue is the query doing it is doing
```sql
INSERT ... ON DUPLICATE KEY UPDATE
```

which is [acquiring different types of locks](https://medium.com/@mingwho/mysql-deadlock-caused-by-insert-on-duplicate-key-update-893ef30993a8) than just a plain `INSERT`.

The solution might be to do it yourself by trying an INSERT and, if it fails, do an UPDATE. This wouldn't acquire [a gap lock](https://dev.mysql.com/doc/refman/8.0/en/innodb-locking.html#innodb-gap-locks) and should reduce the number of deadlocks significantly, probably making all the current deadlock-avoidance related code in the search DAO obsolete.

dkarlovi:
`INSERT ON DUPLICATE KEYS` is quite problematic in specific use cases and situations, see #13507 for details.

  
  
Initial work to fix #13507.

## Additional info  

Requires explicitly setting the expected keys.


brusch:
Added type declarations and fixed merge conflict.

PR #13622, opened 2022-11-18
[Bug]: Read only style for Quantity Value

### LINKED ISSUE #13554: [Bug]: Read only style for Quantity Value
### Expected behavior

Quantity Value should have the same not-editable-style (grey label) as Input

### Actual behavior

image

### Steps to reproduce

Create a class with Quantity Value and Input and set both to Not editable

Corepex:
Resolves #13554

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/quantityValue.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/quantityValue.js
@@ -362,6 +362,7 @@ pimcore.object.tags.quantityValue = Class.create(pimcore.object.tags.abstract, {
 
     getLayoutShow: function () {
         this.getLayoutEdit();
+        this.component.setDisabled(true);
         this.unitField.setReadOnly(true);
         this.inputField.setReadOnly(true);
 


Corepex:
we use `Ext.form.FieldContainer` for the quantityValue fields, which have no `readOnly` property. To achieve a "disabled" state we have to use `disabled`.

https://docs.sencha.com/extjs/6.0.2/classic/Ext.form.FieldContainer.html#cfg-disabled

bundles/AdminBundle/Resources/public/js/pimcore/object/tags/quantityValue.js
```diff
@@ -362,6 +362,7 @@ pimcore.object.tags.quantityValue = Class.create(pimcore.object.tags.abstract, {
 
     getLayoutShow: function () {
         this.getLayoutEdit();
+        this.component.setDisabled(true);
         this.unitField.setReadOnly(true);
         this.inputField.setReadOnly(true);
```
Corepex, line 367: these two `setReadOnly` lines are obsolete since we have the `disabled` option.

PR #13623, opened 2022-11-18
make custom reports scrollable when needed

### LINKED ISSUE #11779: [Bug]: Pimcore 10 pagination issue on dashboard custom reports
### Expected behavior

First page and previous page of the custom reports pagination should be visible when user open the dashboard after the load of the application 
expected

### Actual behavior

When clicked on last page of the custom report pagination , First page and previous page at the bottom of the custom reports on the dashboard are not visible only for the first time after the hard referesh of the application .
error2

### Steps to reproduce

1) Click on file menu on left panel 
2) Click on dashboards 
3) Click on welcome 
4) Click on last Page of the custom report.
dashboard-custom-report-1

torqdev:
Resolves #11779

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/helpers/generic-grid.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/helpers/generic-grid.js
@@ -92,7 +92,8 @@ pimcore.helpers.grid.buildDefaultPagingToolbar = function (store, options) {
         store: store,
         displayInfo: true,
         displayMsg: '{0} - {1} / {2}',
-        emptyMsg: t("no_items_found")
+        emptyMsg: t("no_items_found"),
+        scrollable: true
     };
     if (typeof options !== "undefined") {
         config = Ext.applyIf(options, config);


Corepex:
@torqdev, thanks for the PR. I tested the fix you provided and it basically works, but every time I resize the browser window or even the custom report widget itself, the scrollbar gets lost  ... maybe because of a wrong `overflow: hidden` or something like that.

Would be great if you could investigate this issue.

cameronfromtorq:
@Corepex, Thank you for the reply. Unfortunately I am unable to reproduce what I think you are describing. For me, on multiple browsers, the scrollbar remains visible unless the custom report pagination bar is large enough to be fully displayed. If instead the desired functionality is that the scrollbar is always visible, even when unneeded please let me know.

if you are still able to replicate the error, would you be able to post a screen recording?

Corepex:
@cameronfromtorq @torqdev 

Here is a small video showing the resize "issue" I mentioned in my first comment.
I hope this will help to understand what the problem is :)


https://user-images.githubusercontent.com/16717695/205581623-380b1d77-0fb0-4ad3-82a1-52d1963be1ab.mp4

cameronfromtorq:
@Corepex thank you for the video, it was very helpful. Unfortunately me and others here at Torq are unable to produce the results you are demonstrating. Are you able to produce this error on multiple browsers? 

for reference, here is a recording of what we are seeing:
https://gyazo.com/6f379ce12f3152a72949ae0bedf9be7e

Corepex:
@torqdev, I did a fresh install and it works now. It works perfectly! 🍾 Sometimes it just doesn't work right away ... 🤣 

Thanks for this awesome PR. 🥇

PR #13624, opened 2022-11-18
Fix the default timeout value  in process

aryaantony92:
Resolves #13462


doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -2,6 +2,7 @@
 
 ## 10.5.10
 - [DataObject] Deprecated: Loading non-Concrete objects with the Concrete class will not be possible in Pimcore 11.
+- [Console] Default value of the timeout for the process is set to 60 seconds.
```
jdreesen: It already was 60s by default before. It was just not possible to unset (`null`) it.

PR #13626, opened 2022-11-18
Bugfix/scrollable brick list

### LINKED ISSUE #12005: [Bug]: Grouped Area Brick List not scrollable
### Expected behavior

Be able to scroll the grouped area brick list and access all bricks without scrolling the whole page

### Actual behavior

When having a grouped brick list, the brick lis is not scrollable under all conditions. 
Looks like if the interaction with the list happens far enough at the top of the current page, the list is scrollable
However this is almost always not the case and list is not scrollable. 

Img: cut of brick list on editor edges
brick list cut off on screen editor edges

Img: carefully scrolled page up to see top of brick list
carfully scrolled page up

### Steps to reproduce

Have a long list of bricks in a brick group
access the content of brick group while somewhere down on the page you want to edit

torqdev:
Resolves #12005

### DIFF
--- a/bundles/AdminBundle/Resources/public/css/editmode.css
+++ b/bundles/AdminBundle/Resources/public/css/editmode.css
@@ -481,7 +481,7 @@ input.cke_dialog_ui_input_tel {
     opacity: .9;
     border-color: #000;
     overflow-y: auto;
-    max-height: 90%;
+    max-height: calc(100vh - 15px);
     color-scheme: dark;
 }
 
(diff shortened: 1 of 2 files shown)


bundles/AdminBundle/Resources/public/css/editmode.css
```diff
@@ -480,6 +480,9 @@ input.cke_dialog_ui_input_tel {
 .pimcore_areablock_toolbar {
     opacity: .9;
     border-color: #000;
+    overflow-y: auto;
+    max-height: 90%;
```
kingjia90: what about
```suggestion
    max-height: calc(100vh - 15px);
```
just to not change from the former values defined in js?

evanjamesjackson: Thanks, verified this works and will commit

Corepex:
@torqdev, thanks for your PR - nice fix :)

PR #13634, opened 2022-11-21
[Bug]: Serverside relation limit validation not working

### LINKED ISSUE #13575: [Bug]: Serverside relation limit validation not working
Belongs to: #13551

### Expected behavior

You get a popup showing an error and the dataObject will NOT be saved

image

### Actual behavior

You can save the data object and the corresponding relation field. You will get an error popup, but the fields are saved anyway

Affected Relations:

- ManyToManyRelation
- ManyToManyObjectRelation
- AdvancedManyToManyRelation

### Steps to reproduce

1. create a class and put all many-to-many relations in it (or use the class export)
2. set the "max. items" limit of all relations to 1
3. create a from the created class (obv. in the data objects tab)
4. add **one** item to the relation and hit "save"
5. choose one relation where you want to add a second item -> apply a filert (something like "abcde") 
6. drag a item (asset, dataobject or document) to the relation 
7. remove the filter (you should now see 2 items in the relation) -> the will be fixed by (#13551)
8. hit save -> you should now see a "error" popup - reload the object -> now you should see the 2 items in the relation that was saved!
9. if step 8 doesn't work, try to open the dataobject in a private browser tab - hit save!

hint: also try to hit "save as draft"

[class_TestObject_export.zip](https://github.com/pimcore/pimcore/files/10004258/class_TestObject_export.zip)

Corepex:
Resolves #13575

### DIFF
--- a/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
+++ b/models/DataObject/ClassDefinition/Data/AdvancedManyToManyObjectRelation.php
@@ -416,6 +416,10 @@ class AdvancedManyToManyObjectRelation extends ManyToManyObjectRelation implemen
                     throw new Element\ValidationException('Invalid object relation to object [' . $id . '] in field ' . $this->getName() . ' , tried to assign ' . $o->getId());
                 }
             }
+
+            if ($this->getMaxItems() && count($data) > $this->getMaxItems()) {
+                throw new Element\ValidationException('Number of allowed relations in field `' . $this->getName() . '` exceeded (max. ' . $this->getMaxItems() . ')');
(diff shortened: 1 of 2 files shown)


Corepex:
It turns out that only `AdvancedManyToManyRelation` and `AdvancedManyToManyObjectRelation` are affected by the mentioned problem

Note: its only reproducible through the API ...

Use something like this:
```php
$testObject = \Pimcore\Model\DataObject\TestObject::getById(1234);
$testObject->setAdvancedManyToManyObjectRelation([
    $metaDataObject1, $metaDataObject2
]);
$testObject->save();
```

PR #13637, opened 2022-11-21
[Session] Fix AdminSessionHandler write close in case of no main request

### LINKED ISSUE #11927: [Bug]: Authentication is closing session when its not supposed to

dvesh3:
Resolves #11927

### DIFF
--- a/bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
+++ b/bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
@@ -297,7 +297,8 @@ class AdminSessionHandler implements LoggerAwareInterface, AdminSessionHandlerIn
      */
     private function shouldWriteAndClose(): bool
     {
-        return $this->canWriteAndClose ??= $this->isAdminRequest($this->requestHelper->getMainRequest());
+        return $this->canWriteAndClose ??= $this->requestHelper->hasMainRequest()
+            && $this->isAdminRequest($this->requestHelper->getMainRequest());
     }
 
     /**
--- a/bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
+++ b/bundles/AdminBundle/Session/Handler/AdminSessionHandler.php
@@ -297,8 +297,10 @@ class AdminSessionHandler implements LoggerAwareInterface, AdminSessionHandlerIn
      */
     private function shouldWriteAndClose(): bool
     {
-        return $this->canWriteAndClose ??= $this->requestHelper->hasMainRequest()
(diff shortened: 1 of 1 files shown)


solverat:
@dvesh3 This won't work, because it would not close the session if no request is available.

This should work: 

```php

if (!$this->requestHelper->hasMainRequest()) {
    return true;
}

return $this->canWriteAndClose ??= $this->isAdminRequest($this->requestHelper->getMainRequest());

dvesh3:
> @dvesh3 This won't work, because it would not close the session if no request is available.
> 
> This should work:
> 
> ```
> if (!$this->requestHelper->hasMainRequest()) {
>     return true;
> }
> 
> return $this->canWriteAndClose ??= $this->isAdminRequest($this->requestHelper->getMainRequest());
> ```

@solverat I'm wondering, how it loads session in first place, when request is not available 🤔 
image

solverat:
That's why I suggested the "In a perfect World, the request should reach `AdminSessionHandler`  https://github.com/pimcore/pimcore/issues/11927#issuecomment-1320510099"

That's why I think that the method `shouldWriteAndClose` is quite dangerous. It should receive the original request (which also holds the initiated session, started in `Authentication::authenticateSession($request)`). Otherwise, we **can't be 100%** sure, that we're dealing with the right request object.

To answer your question: I'm not sure why this is happening, that's why I also attached the stack trace in the linked comment above. But it's also a fact, that functional tests (Symfony Module) are running into `Failed to start the session: already started by PHP.` because of this `There is no main request available` exception.

dvesh3:
@solverat This is happening due to session service [fallback](https://github.com/pimcore/pimcore/blob/98f11e7f066e06ba8a5762e542475efa39b381a3/bundles/AdminBundle/Session/Handler/AdminSessionHandler.php#L97). I have now adapted the PR to return true if request is not available.

Issue #9329, opened 2021-06-04
Variant classificationstore fields are not marked as inherited unless one field is set.


Daniel-Ateles:
# Bug Report

### Expected behavior
When creating a variant with empty classificationstores all fields from the parent object should be visible (as now) but also grayed out and marked with inheritance-triangles.

### Actual behavior
Fields are not grayed out and no triangles is visible - until you change one field in the variant classificationstore.
Inheritance still works though - if you change the parent the changed value will be displayed.

### Steps to reproduce
Create object with classificationstore.
Create a variant.

### Faulty code
DataObjectController.php
```php
if ($fielddefinition->isEmpty($fieldData) && !empty($parent)) { 
    $this->getDataForField($parent, $key, $fielddefinition, $objectFromVersion, $level + 1);
} else {
```
As you can see a recursion with the parent will occur if the cs is empty (no items) because of $fielddefinition->isEmpty($fieldData)

The recursion will throw away all the inheritance metadata collected in the variant.

# Feature Request

The collected inheritance metadata from the variant should be merged with the parents metadata instead of resetting it to only contain the parents metadata.

dvesh3:
Fixed by #9683

PR #13638, opened 2022-11-21
fix: from https://github.com/pimcore/pimcore/issues/9329

tszustak:
Resolves #


brusch:
@tszustak thanks for your PR, however, `6.9` is already EOL - sorry for that 😊

PR #13648, opened 2022-11-22
Add info about naming strategy to update guide

Cruiser13:
docs only


brusch:
@Cruiser13 Thanks for your PR, however, this should go into `10.5` as it's not relevant for v11 anymore. Thanks again :)

PR #13651, opened 2022-11-22
[Data objects] Version preview for blocks

BlackbitDevs:
Currently blocks do not have a version preview. It just looks like this:
Bildschirmfoto 2022-11-22 um 11 25 48

This PR adds this feature, version preview looks like this:

Block:
Bildschirmfoto 2022-11-22 um 10 30 02

It is actually based on https://github.com/pimcore/pimcore/blob/8d831bac31c3a6d97e7c563c4322ae87841a8698/models/DataObject/ClassDefinition/Data/Fieldcollections.php#L552-L583

In fact `Fieldcollections::getDiffVersionPreview()` is not used for version comparison in Pimcore backend but to provide a uniform interface to get the version preview, I have applied this to `Block::getVersionPreview()`, `Block::getDiffVersionPreview()` and `Fieldcollections::getVersionPreview()`.

An alternative would be to put all version display logic from https://github.com/pimcore/pimcore/blob/11.x/bundles/AdminBundle/templates/admin/data_object/data_object/preview_version.html.twig into the corresponding `getVersionPreview()` methods? But this is for another PR...

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Block.php
+++ b/models/DataObject/ClassDefinition/Data/Block.php
@@ -479,7 +479,7 @@ class Block extends Data implements CustomResourcePersistingInterface, ResourceP
      */
     public function getVersionPreview(mixed $data, DataObject\Concrete $object = null, array $params = []): string
     {
-        return 'not supported';
+        return $this->getDiffVersionPreview($data, $object, $params)['html'];
     }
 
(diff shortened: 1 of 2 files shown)


robertSt7:
@BlackbitDevs Thanks a lot

PR #13474, opened 2022-10-28
[Task] Apply native parameter and return type hints according to php docs

mcop1:
Resolves #12240

Draft, tests not working yet.

PR #13653, opened 2022-11-22
[Task] Follow up native params and return type hints

### LINKED ISSUE #13656: [Task]: Fix possible errors after applying php native types
### Improvement description

Follow up to #12240

kingjia90:
Follow-up https://github.com/pimcore/pimcore/pull/13474

- [x] PdfReactor processor
https://github.com/pimcore/pimcore/blob/8d831bac31c3a6d97e7c563c4322ae87841a8698/lib/Web2Print/Processor.php#L153
- [x] Ecommerce Environment
I see `useGuestCart` is [nullable ](https://github.com/pimcore/pimcore/commit/3df0381dc02275d6dbde96e8246983c1d90ea56f) now, then it should be `null` like the other properties as well when clearing, also [this change ](https://github.com/pimcore/demo/blob/922243c3c9acb7f6b39985553120a2bb52c0cd26/src/Ecommerce/Overrides/Environment.php#L40-L41)would be missing after merging https://github.com/pimcore/demo/pull/376 
- [x] this line got fixed by Blankse in the [phpStan repo ](https://github.com/phpstan/phpstan-src/pull/2020) and could be removed again https://github.com/pimcore/pimcore/blob/631a52073541777720403653f987bd62f1ac881f/lib/Image/Adapter/Imagick.php#L328 - will be resolved with a follow up task: https://github.com/pimcore/pimcore/issues/13688

### DIFF
--- a/bundles/EcommerceFrameworkBundle/src/Tracking/AbstractProductData.php
+++ b/bundles/EcommerceFrameworkBundle/src/Tracking/AbstractProductData.php
@@ -26,9 +26,9 @@ abstract class AbstractProductData extends AbstractData
 
     protected array $categories;
 
-    protected string $variant;
+    protected ?string $variant = null;
 
-    protected int $position;
+    protected int $position = 0;
 
     protected float $price;
 
@@ -89,7 +89,7 @@ abstract class AbstractProductData extends AbstractData
         return $this;
     }
 
-    public function getVariant(): string
+    public function getVariant(): ?string
     {
         return $this->variant;
     }
--- a/bundles/EcommerceFrameworkBundle/src/Tracking/AbstractProductData.php
+++ b/bundles/EcommerceFrameworkBundle/src/Tracking/AbstractProductData.php
@@ -61,6 +61,9 @@ abstract class AbstractProductData extends AbstractData
         return $this->brand;
     }
 
+    /**
+     * @return $this
+     */
     public function setBrand(?string $brand = null): static
     {
         $this->brand = $brand;
--- a/bundles/EcommerceFrameworkBundle/src/Tracking/AbstractProductData.php
+++ b/bundles/EcommerceFrameworkBundle/src/Tracking/AbstractProductData.php
@@ -22,7 +22,7 @@ abstract class AbstractProductData extends AbstractData
 
     protected string $name;
 
-    protected string $brand;
+    protected ?string $brand = null;
 
     protected array $categories;
 
@@ -56,12 +56,12 @@ abstract class AbstractProductData extends AbstractData
         return $this;
     }
 
-    public function getBrand(): string
+    public function getBrand(): ?string
     {
         return $this->brand;
     }
 
-    public function setBrand(string $brand): static
+    public function setBrand(?string $brand = null): static
     {
         $this->brand = $brand;
 
--- a/lib/Image/Adapter/Imagick.php
+++ b/lib/Image/Adapter/Imagick.php
@@ -325,7 +325,6 @@ class Imagick extends Adapter
             for ($i = 0; $i < $width; $i++) {
                 for ($j = 0; $j < $height; $j++) {
                     $pixel = $this->resource->getImagePixelColor($i, $j);
-                    // @phpstan-ignore-next-line - phpstan expects bool, but actually doc says int
                     $color = $pixel->getColor(1); // get the real alpha not just 1/0
                     if ($color['a'] < 1) { // if there's an alpha pixel, return true
                         return true;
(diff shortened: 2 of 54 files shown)


lib/Navigation/Renderer/Breadcrumbs.php
```diff
@@ -110,12 +110,12 @@ public function getLinkLast(): bool
...
     }
 
-    public function getTemplate(): array|string
+    public function getTemplate(): array|string|null
     {
         return $this->_template;
     }
 
-    public function setTemplate(array|string $template): static
+    public function setTemplate(array|string|null $template): static
```
blankse, line 121:
```suggestion
    /**
     * @return $this
     */
    public function setTemplate(array|string|null $template): static
```

bundles/EcommerceFrameworkBundle/src/Tracking/AbstractProductData.php
```diff
@@ -56,12 +56,12 @@ public function setName(string $name): static
...
     }
 
-    public function getBrand(): string
+    public function getBrand(): ?string
     {
         return $this->brand;
     }
 
-    public function setBrand(string $brand): static
+    public function setBrand(?string $brand = null): static
```
blankse, line 67:
```suggestion
    /**
     * @return $this
     */
    public function setBrand(?string $brand = null): static
```

models/Document/Editable/Select.php
```diff
@@ -66,7 +66,9 @@ public function frontend()
      */
     public function setDataFromResource(mixed $data): static
     {
-        $this->text = $data;
+        if(is_string($data)) {
```
fashxp: what if data is a integer value (or float) ... should we not just cast at this place?

fashxp (changes requested):

lib/Image/Adapter/Imagick.php
```diff
@@ -325,7 +325,6 @@ private function hasAlphaChannel(): bool
             for ($i = 0; $i < $width; $i++) {
                 for ($j = 0; $j < $height; $j++) {
                     $pixel = $this->resource->getImagePixelColor($i, $j);
-                    // @phpstan-ignore-next-line - phpstan expects bool, but actually doc says int
```
kingjia90: tests failing because of  the [unreleased ](https://github.com/phpstan/phpstan-src/pull/2020)changes, so waiting for 1.9.3

blankse: @kingjia90 We can set the phpstan version to 1.9.2 in the composer.json till the version 1.9.3 is released.

kingjia90: do you mean 1.9.x-dev?

blankse: @kingjia90 I meant set to version 1.9.2 and leave the comment. But without comment with 1.9.x-dev could also work :)

models/Document/Editable/Select.php
mcop1: Removed the check, but added the string cast

fashxp (approved):

models/DataObject/Concrete.php
```diff
@@ -75,16 +75,16 @@ class Concrete extends DataObject implements LazyLoadedFieldsInterface
     /**
      * @internal
      *
-     * @var string
+     * @var string|null
      */
-    protected $o_classId;
+    protected ?string $o_classId = null;
```
dvesh3: this change requires migration for the implementations on minor version, which should add the property typehints here: https://github.com/pimcore/pimcore/blob/fa3a9090ebc3fc826067c818edccb9a8ec7a80eb/lib/DataObject/ClassBuilder/FieldDefinitionPropertiesBuilder.php#L34
https://github.com/pimcore/pimcore/blob/83142b53cfd5a791e3c6167470a66336fd03da22/lib/DataObject/ClassBuilder/FieldCollectionClassBuilder.php#L68
https://github.com/pimcore/pimcore/blob/83142b53cfd5a791e3c6167470a66336fd03da22/lib/DataObject/ClassBuilder/ObjectBrickClassBuilder.php#L72

fashxp: Can we even provide a migration for that? 
As soon as we add the type to either (Concrete or real implementation, e.g. `Pimcore\Model\DataObject\Car`), it will break. 
So I think the only option is to add this manually in the course of upgrading to Pimcore 11. 
Or we remove that property type again.

fashxp: agreed solution: remove types for `$class_id` and `$class_name`.

mcop1: Removed types for both properties, removed the type in the build as well.

PR #13658, opened 2022-11-22
Add info about naming strategy to update guide

Cruiser13:
docs only


aryaantony92:
@Cruiser13 Thank you!

BlackbitDevs:
@Cruiser13 Did the migration of the editable naming strategy work for you? We had serious problems. At first the strategy services were not public and thus could not be loaded in the command class. Ok, was easy to fix.

Then we tried migrating a single document. Some did not work at all (afterwards the documents were simply empty (yes, we changed the naming strategy in the YML file after the migration). For some others it worked. 
When we tested the command with single documents when a document includes a snippet in the footer, this does also get migrated (so imho the `-d` parameter is not strict enough) - and then with the next document it complained about that the footer document was already migrated. 

Ok, so we thought we have to migrate all documents at once. This also did not work because the command aborted on the first error it gets - even if other documents worked perfectly when we converted them as single document with `-d` parameter.

So all in all, we had to give up and copy the contents manually. Maybe this was because of the project which has been touched by a lot of people over the last decade. But I am curious if you had a better experience...

Cruiser13:
@BlackbitDevs we did the migration using the render strategy on 6.9.6 and have been able to update smoothly to pimcore X afterwards, yes. We had some errors thrown on rendering first (warnings like "var not defined") so we fixed them and retried. It worked fine for us but I guess you might run into errors depending on the structure of your project (for things like the footer). We did not have editables in includes for example, we used snippets here.

PR #13663, opened 2022-11-23
fix sql condition being skipped when exporting objects in grid

### LINKED ISSUE #13662: [Bug]: Sql condition is skipped when exporting to csv in grid
### Expected behavior

When applying sql condition only data objects that match sql conditions are exported when export to csv or xlsx is called.

### Actual behavior

When applying sql condition all data objects are exporting when export to csv or xlsc is called skipping sql condition.

### Steps to reproduce

1. Create some sql condition that will limit data objects, for example "length(color) <= 5"
2. Click "export to csv". All records will be exported including records that not match sql condition.

pfilipkowskilemon:
Resolves #
#13662 

## Additional info  
Resolves issue with skipping sql condition when exporting to csv in grid

### DIFF
--- a/bundles/AdminBundle/Resources/public/js/pimcore/element/helpers/gridColumnConfig.js
+++ b/bundles/AdminBundle/Resources/public/js/pimcore/element/helpers/gridColumnConfig.js
@@ -820,7 +820,7 @@ pimcore.element.helpers.gridColumnConfig = {
         var condition = "";
         var searchQuery = this.searchField ? this.searchField.getValue() : "";
 
-        if (this.sqlFilter) {
+        if (this.sqlFilter || (this.sqlButton && this.sqlButton.pressed)) {
             condition = this.sqlEditor.getValue();
         } else {
             var filterData = this.store.getFilters().items;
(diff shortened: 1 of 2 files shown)


Corepex:
@pfilipkowskilemon, thanks for your PR. I tested your solution, and it worked correctly, but I maybe found another way that fixes the `this.sqlFilter`. I think it would be a cleaner solution to add `this.sqlFilter = field.getValue()` to `gridTabAbstract.js#L181`. In this case, `this.sqlFilter` will be set on "submitting" the filter input.

https://github.com/pimcore/pimcore/blob/10.6/bundles/AdminBundle/Resources/public/js/pimcore/object/helpers/gridTabAbstract.js#L181

Corepex:
@pfilipkowskilemon friendly reminder - if you provide the changes, we could add them to today's release

pfilipkowskilemon:
@Corepex thanks for reminder

Corepex:
@pfilipkowskilemon, thanks for helping us to improve pimcore 👍

PR #13672, opened 2022-11-24
Fix max height of grouped brick menu

blankse:
Followup to https://github.com/pimcore/pimcore/pull/13626

Still not working after this PR unfortunately. It only works at the beginning of a document. But if the document is long and I want to insert something at the end there are no scroll buttons. 

I think the fix only works for the brick menu without groups. We have all bricks in groups, so I found this solution to fix this.


Corepex:
@blankse, thanks for your PR 👍

PR #13676, opened 2022-11-24
[Data Objects] Add select mode to many-to-one relation field

BlackbitDevs:
Similar to #13227 this PR implements a *select* display mode for many-to-one relational fields. With this display mode the user does not need to open the search window but can directly type to search for the object to be assigned:

Снимок экрана 2022-11-28 в 12 36 03

Of course all other functionality like the search window, drag & drop, asset uploading also still work.

Currently the visible fields are not configurable for many-to-one relations (think this will change shortly). For now the `fullpath` is displayed as it currently is the default. In near future we may make visible fields also configurable for many-to-one-relation fields...


BlackbitDevs:
The PhpStan error comes from https://github.com/pimcore/pimcore/blob/631a52073541777720403653f987bd62f1ac881f/lib/Image/Adapter/Imagick.php#L328-L329
I have not changed anything there. Tried to merge `11.x` but to no success. Any suggestions?

mcop1:
Hello @BlackbitDevs ,
when I check out your pr, I get the following error:
`Cannot access property starting with "\0"` and because of that the DataObject tab is missing. 
Could you please try to merge 11.x into your pr once again? I noticed that it isn´t based on the most actual version of the branch.

BlackbitDevs:
@mcop1 Have merged 11.x again. Is it better now?

mcop1:
@BlackbitDevs Thanks, way better now. But unfortunately you were a few minutes too early. We just merged the pr that should fix the failing tests. Could you please merge 11.x once more? Then I can test & merge the changes tomorrow.

BlackbitDevs:
@mcop1 Merged 11.x again. Looking forward to your review ;-)

models/DataObject/ClassDefinition/Data/ManyToManyRelation.php
```diff
@@ -29,7 +29,6 @@
 
 class ManyToManyRelation extends AbstractRelations implements QueryResourcePersistenceAwareInterface, OptimizedAdminLoadingInterface, TypeDeclarationSupportInterface, VarExporterInterface, NormalizerInterface, IdRewriterInterface, PreGetDataInterface, PreSetDataInterface
```
mcop1:
```suggestion
class ManyToManyRelation extends AbstractRelations implements QueryResourcePersistenceAwareInterface, OptimizedAdminLoadingInterface, VarExporterInterface, NormalizerInterface, PreGetDataInterface, PreSetDataInterface
```
Can´t we remove these 2 interfaces here too?

BlackbitDevs:
@mcop1 You are right, I committed your suggested change.

mcop1:
@BlackbitDevs Thanks, looks good. 
Could you have a look at the failing tests please?

BlackbitDevs:
@mcop1 Have fixed the test. I removed the additional `is_array` checks and some others in `AllowAssetRelationTrait` similar to https://github.com/pimcore/pimcore/blob/554e67a4061854221deb9a6b8aba8f7c10a7d354/models/DataObject/ClassDefinition/Data/Relations/AllowObjectRelationTrait.php#L42-L66
Now PhpStan is satisfied.

mcop1:
Hello @BlackbitDevs ,
thanks for the heads up, but there still seems to be something wrong with phpstan.

BlackbitDevs:
@mcop1 Do you have an idea what PhpStan wants? I am missing the line number for 
```
------ ----------------------------------------------------------------------- 
  Line   models/DataObject/ClassDefinition/Data/ManyToManyRelation.php          
 ------ ----------------------------------------------------------------------- 
         Ignored error pattern #^Left side of && is always true\.$# in path     
         /home/runner/work/pimcore/pimcore/models/DataObject/ClassDefinition/D  
         ata/ManyToManyRelation.php was not matched in reported errors.         
 ------ -----------------------------------------------------------------------
```

mcop1:
> @mcop1 Do you have an idea what PhpStan wants? I am missing the line number for
> 
> ```
> ------ ----------------------------------------------------------------------- 
>   Line   models/DataObject/ClassDefinition/Data/ManyToManyRelation.php          
>  ------ ----------------------------------------------------------------------- 
>          Ignored error pattern #^Left side of && is always true\.$# in path     
>          /home/runner/work/pimcore/pimcore/models/DataObject/ClassDefinition/D  
>          ata/ManyToManyRelation.php was not matched in reported errors.         
>  ------ -----------------------------------------------------------------------
> ```

@BlackbitDevs seems like you resolved errors that where previously ignored. I think it should be fine to remove them from the `phpstan-baseline.neon` file: 
https://github.com/pimcore/pimcore/blob/6dd1f96a6800aed8551762e33286a3edeebcab4e/phpstan-baseline.neon#L143-L151

BlackbitDevs:
IT'S GREEN!!! 

@mcop1 Please have a close look if my changes to AllowAssetRelation are correct, especially if this part, which I removed, is still necessary:
https://github.com/pimcore/pimcore/blob/03a07aa932286b786d8f7904c31db98476cab2ca/models/DataObject/ClassDefinition/Data/Relations/AllowAssetRelationTrait.php#L57-L63

mcop1:
@BlackbitDevs looks good to me, we are already doing something similar here https://github.com/pimcore/pimcore/blob/4e4fae5ca778b69727f6b4a5e3582e42f48fdbc3/bundles/AdminBundle/src/Controller/Admin/Asset/AssetController.php#L2845

PR #13696, opened 2022-11-28
Bug fix: handle thumbnail path with comma

bgpack:
Encode comma in thumbnail path as html srcset is a comma separated list and any getHtml will fail as the path will not be correctly extracted from the srcset.


dvesh3:
please check for srcset in other places

aashan10 (approved):

PR #13697, opened 2022-11-28
Bug fix: handle the case where the 'containerType' of context is not set

bgpack:
handle null pointer in the case where the 'containerType' of context is not set


Corepex:
@bgpack, thanks for helping us improving pimcore 👍

PR #13701, opened 2022-11-29
[Task]: Implement method to get the version dependent system column name for data objects

### LINKED ISSUE #13512: [Task]: Implement method to get the version dependent system column name for data objects
### Improvement description

We want to remove the "o_" prefix for system properties in pimcore 11. But those properties (or the corresponding db columns) are often used in direct (sql-) statement. For bundles we want to have the possibility to support Pimcore 10.x and 11.x, therefore this method is needed. You can call the method to get the proper name for the property/column in the used version.

E.g. `DataObject\Service::getSystemColumnName('id')` returns `o_id` for 10.x and `id` for 11.x

Please see https://github.com/pimcore/pimcore/issues/7341

mcop1:
Resolves #13512

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -62,6 +62,15 @@ class Service extends Model\Element\Service
     protected static $systemFields = ['o_path', 'o_key', 'o_id', 'o_published', 'o_creationDate', 'o_modificationDate', 'o_fullpath'];
 
     /**
+     * TODO Bc layer for bundles to support both Pimcore 10 & 11, remove with Pimcore 12
+     * @var string[]
+     */
+    protected static array $versionDependentSystemFields = ['id', 'parentId', 'type', 'key', 'path', 'index', 'published',
+                                                            'creationDate', 'modificationDate', 'userOwner', 'userModification',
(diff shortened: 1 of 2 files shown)


models/DataObject/Service.php
```diff
@@ -61,6 +61,15 @@ class Service extends Model\Element\Service
      */
     protected static $systemFields = ['o_path', 'o_key', 'o_id', 'o_published', 'o_creationDate', 'o_modificationDate', 'o_fullpath'];
 
+    /**
+     * TODO Bc layer for bundles to support both Pimcore 10 & 11, remove with Pimcore 12
+     * @var string[]
+     */
+    protected static array $versionDependentSystemFields = ['id', 'parentId', 'type', 'key', 'path', 'index', 'published',
```
brusch:
```suggestion
    private const array BC_VERSION_DEPENDENT_DATABASE_COLUMNS = ['id', 'parentId', 'type', 'key', 'path', 'index', 'published',
```
WDYT about that? 
And also in general the naming schema.

brusch (changes requested):

models/DataObject/Service.php
mcop1: Fine with me, changed name and type (const) of the property and the method name

PR #13703, opened 2022-11-29
[BUG] Asset Upload of advancedManyToManyRelation and manyToManyRelation

storbahn:
Fix validation of asset upload of advancedManyToManyRelation and manyToManyRelation (https://github.com/pimcore/pimcore/pull/13543)

If this.fieldConfig.maxItems is not defined (null) then 0 >= null is always true and we get the error message "Limit Reached".

asset-upload_limit_reached_error


Corepex:
@storbahn, thanks for your PR. We must have missed that case while testing the upload limit. 👍

Issue #13627, opened 2022-11-18
[Improvement] Revalidate phpdoc types


fashxp:
### Improvement description

Double check and compoare with 11.x branch phpdoc types for following files (approx line in (xx) for every entry) : 

- [x] bundles/AdminBundle/src/Controller/Admin/DataObject/DataObjectController.php:getTreeNodeConfig() (234)
- [x] bundles/AdminBundle/src/Controller/Admin/DataObject/DataObjectController.php:getTreeNodeConfig() (234)
- [x] bundles/EcommerceFrameworkBundle/src/CoreExtensions/ClassDefinition/IndexFieldSelection.php:setMultiPreSelect (95)
- [x] bundles/EcommerceFrameworkBundle/src/CoreExtensions/ClassDefinition/IndexFieldSelection.php:getMultiPreSelect (100)
- [x] bundles/EcommerceFrameworkBundle/src/CoreExtensions/ClassDefinition/IndexFieldSelection.php:getDataForResource (124)
- [x] bundles/EcommerceFrameworkBundle/src/CoreExtensions/ClassDefinition/IndexFieldSelection.php:getDataFromResource (150)
- [x] bundles/EcommerceFrameworkBundle/src/CoreExtensions/ClassDefinition/IndexFieldSelection.php:getDataForQueryResource (168)
- [x] bundles/EcommerceFrameworkBundle/src/CoreExtensions/ClassDefinition/IndexFieldSelection.php:getDataForEditmode (182)
- [x] models/DataObject/Data/QuantityValue.php:__construct() (33)
- [x] models/DataObject/Data/QuantityValue.php:setValue() (39)
- [x] models/DataObject/Data/QuantityValue.php:getValue() (45)
- [x] models/DataObject/Data/InputQuantityValue.php:setValue() (35)
- [x] models/DataObject/Data/InputQuantityValue.php:getValue() (41)
- [x] models/DataObject/Data/ImageGallery.php:current (40) 
- [x] lib/Image/Adapter.php:$sourceImageFormat (41)
- [x] models/DataObject/Data/Video.php:$data (31)
- [x] models/DataObject/Data/Video.php:$poster (33)
- [x] models/Document/Editable/Image.php:$image (54)
- [x] lib/Model/AbstractModel.php:getDao()  (44)
- [x] lib/Model/ModelInterface.php:getDao() (23) 
- [x] models/Tool/SettingsStore.php:$scope (41)
[...]

PR #13705, opened 2022-11-29
Further type fixes

### LINKED ISSUE #13656: [Task]: Fix possible errors after applying php native types

fashxp:
followup to #13653

### TODO
- [x] add relevant method signature changes to that list: https://github.com/pimcore/pimcore/issues/13627

### DIFF
--- a/lib/Http/ResponseHelper.php
+++ b/lib/Http/ResponseHelper.php
@@ -61,7 +61,7 @@ class ResponseHelper
             return false;
         }
 
-        if (false !== strpos($response->getContent(), '<html')) {
+        if (false !== strpos((string)$response->getContent(), '<html')) {
             return true;
         }
 
--- a/models/DataObject/ClassDefinition/Data/Block.php
+++ b/models/DataObject/ClassDefinition/Data/Block.php
@@ -555,6 +555,9 @@ class Block extends Data implements CustomResourcePersistingInterface, ResourceP
     }
 
     /**
+     * typehint "mixed" is required for asset-metadata-definitions bundle
+     * since it doesn't extend Core Data Types
+     *
      * @param Data|Layout $child
      */
     public function addChild(mixed $child)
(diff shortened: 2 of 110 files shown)


models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -554,7 +554,10 @@ public function hasChildren(): bool
         return false;
     }
 
-    public function addChild(Data|Layout $child)
+    /**
+     * @param Data|Layout $child
+     */
+    public function addChild(mixed $child)
```
fashxp, line 563: @dvesh3 should we add a comment here (and all the other places) why `mixed` is necessary?

dvesh3: yes, this is required for assets metadata definitions bundle. please see https://github.com/pimcore/asset-metadata-class-definitions/actions/runs/3576806963/jobs/6015056604#step:10:47

I'll add comments here once the approach is finalized.

lib/Http/ResponseHelper.php
```diff
@@ -61,7 +61,7 @@ public function isHtmlResponse(Response $response): bool
             return false;
         }
 
-        if (false !== strpos($response->getContent(), '<html')) {
+        if (false !== strpos((string)$response->getContent(), '<html')) {
```
brusch, line 64:
```suggestion
        if (false !== strpos($response->getContent(), '<html')) {
```

See #13710

brusch (dismissed):

lib/Http/ResponseHelper.php
fashxp: lets keep the cast too ...

models/DataObject/ClassDefinition/Data/Block.php
dvesh3: done here https://github.com/pimcore/pimcore/pull/13705/commits/cbbbe8150108d3b2a1387eae678e1bd103998dc4

PR #13713, opened 2022-11-30
[Docs] Mention migrating document editable naming strategy

BlackbitDevs:
Pimcore 6 supported the Pimcore 5 (or 4?) document editable naming strategy with the config
```yaml
pimcore:
    documents:
        editables:
            naming_strategy: legacy
```

Pimcore 10 does not. So let's advise people to at least check if they already have migrated - because afterwards the `pimcore:documents:migrate-naming-strategy` command does not exist anymore.

PS: As https://github.com/pimcore/pimcore/pull/13713#issuecomment-1332100284 already added this to the docs, this PR now only adds the command to check if the project uses legacy naming strategy.

### DIFF
--- a/doc/Development_Documentation/23_Installation_and_Upgrade/07_Updating_Pimcore/10_V6_to_V10.md
+++ b/doc/Development_Documentation/23_Installation_and_Upgrade/07_Updating_Pimcore/10_V6_to_V10.md
@@ -28,7 +28,7 @@ Update your project folder structure as per [Symfony Flex](https://symfony.com/d
 You can use `./bin/console migration:controller-reference` to migrate your existing Documents, Static routes and Document Types to the new controller references in the format: `AppBundle\Controller\FooController::barAction`.
 
 ### Check document editable naming strategy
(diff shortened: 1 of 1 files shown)


jdreesen:
This relates to #13658 which was merged recently.

BlackbitDevs:
Thx @jdreesen. Fixed by https://github.com/pimcore/pimcore/pull/13658

BlackbitDevs:
I will add how to check of the project uses legacy naming strategy.

aryaantony92:
@BlackbitDevs Thank you!

BlackbitDevs:
@aryaantony92 ~~How can it be that this change is not visible on https://pimcore.com/docs/pimcore/current/Development_Documentation/Installation_and_Upgrade/Updating_Pimcore/V6_to_V10.html ? How often do the docs get updated? Or is some configuration wrong (because the link `Edit on Github` on the top right does also not work)?~~

Sorry, browser search only did not find it (in Firefox). Everything ok.

PR #13714, opened 2022-11-30
[Data objects] Advanced relations: Render "number" meta fields with correct decimal separator

BlackbitDevs:
Steps to reproduce:

1. Assure current user's language is `English`
2. Create class with advanced many-to-many (object) relation field. To this add meta column `test` of type `number`
3. Create object of this class, assign the object itself to the relation, in meta column `test` write `1.23`
4. Change language of currently logged in user to `German`
5. Open object again

Expected behaviour: In column `test` the value should be `1,23` (because in Germany comma is used as decimal separator).

Actual behaviour: You will see `1.23` (with dot as decimal separator).

This PR applies the correct locale renderer for number fields.


Corepex:
@BlackbitDevs, thanks for the PR. Now the formatting changes with the language :)

PR #13715, opened 2022-11-30
Fix maintenance mode no session error

### LINKED ISSUE #13711: [Bug] [Pimcore 11] Activate Maintenance mode destroys instance
### Expected behavior

maintenance message is shown

### Actual behavior

image

also it is not possible to deactivate maintenance mode in backend anymore

### Steps to reproduce

activate maintenance mode in backend

dvesh3:
Resolves #13711

### DIFF
--- a/bundles/CoreBundle/config/event_listeners.yaml
+++ b/bundles/CoreBundle/config/event_listeners.yaml
@@ -135,8 +135,6 @@ services:
     Pimcore\Bundle\CoreBundle\EventListener\MaintenancePageListener:
         calls:
             - [loadTemplateFromResource, ['@@PimcoreCoreBundle/public/misc/maintenance.html']]
-        tags:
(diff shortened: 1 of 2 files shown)


brusch:
Looks good to me so far, there's just one potential issue, it's about static pages, which will still get served by `StaticPageGeneratorListener` when maintenance mode is on - but imho that's not relevant, because normally they are getting served anyway if configured correctly on the webserver (Nginx, Apache, ...), as the listener is just a fallback anyway.

dvesh3:
> Looks good to me so far, there's just one potential issue, it's about static pages, which will still get served by `StaticPageGeneratorListener` when maintenance mode is on - but imho that's not relevant, because normally they are getting served anyway if configured correctly on the webserver (Nginx, Apache, ...), as the listener is just a fallback anyway.

@brusch yes, we can't really do anything about static pages as they are served directly from the server.

PR #13722, opened 2022-12-01
[Elements] Remove non-printable characters from element keys

BlackbitDevs:
To reproduce, run command
```php
var_dump(json_encode(\Pimcore\Model\Element\Service::getValidKey("Nivellier\xC2\xADmasse.pdf", 'asset')));
```
It returns `"Nivellier\u00admasse.pdf"`. `\xC2\xAD` is a [soft-hyphen](https://www.compart.com/de/unicode/U+00AD)

When trying to create an asset with this name, there even is a Flysystem error:
`League\Flysystem\CorruptedPathDetected: Corrupted path detected: Nivellier­­masse.pdf in <Pimcore root>vendor/league/flysystem/src/CorruptedPathDetected.php:11` (the soft-hyphen is not visible here but when you copy the filename to an editor, you will see it)

For other element types there is no error but there may be unexpected behaviour.

This PR removes such control characters and other non-printable characters from element keys.

### DIFF
--- a/models/Element/Service.php
+++ b/models/Element/Service.php
@@ -1114,6 +1114,10 @@ class Service extends Model\AbstractModel
 
         // replace all 4 byte unicode characters
         $key = preg_replace('/[\x{10000}-\x{10FFFF}]/u', '-', $key);
+
+        // replace non-printable characters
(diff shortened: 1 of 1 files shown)


brusch:
@BlackbitDevs #13382 should fix that as well, right and it's easier to read? 
We merged #13382 into `11.x` because we were afraid of possible side effects, so I'm not sure whether we really should merge that into `105` as well - this would probably need further testing.

BlackbitDevs:
You are right, https://github.com/pimcore/pimcore/pull/13382 is perfectly fine.

PR #13726, opened 2022-12-01
[Fix] adding validation message for wrong upload type

### LINKED ISSUE #13000: [Bug]: Field Type Image Allowed PDF, txt etc... files

mattamon:
Adding validation message for wrong upload type
Resolves #13000 

## Additional info  
See #13000 for description

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -534,7 +534,7 @@ class AssetController extends ElementControllerBase implements KernelControllerE
             $assetType = Asset::getTypeFromMimeMapping($mimetype, $filename);
 
             if($type !== $assetType) {
-                throw new \Exception("Mime type does not match asset type: $type");
+                throw new \Exception("Mime type does not match with asset type: $type");
             }
         }
 
--- a/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
+++ b/bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
@@ -527,6 +527,17 @@ class AssetController extends ElementControllerBase implements KernelControllerE
             throw new \Exception('Something went wrong, please check upload_max_filesize and post_max_size in your php.ini as well as the write permissions of your temporary directories.');
         }
 
+        // check if there is a requested type and if matches the asset type of the uploaded file
+        $type = $request->get('type');
+        if($type) {
+            $mimetype = MimeTypes::getDefault()->guessMimeType($sourcePath);
+            $assetType = Asset::getTypeFromMimeMapping($mimetype, $filename);
+
+            if($type !== $assetType) {
+                throw new \Exception("Mime type does not match asset type: $type");
+            }
+        }
+
         if ($request->get('allowOverwrite') && Asset\Service::pathExists($parentAsset->getRealFullPath().'/'.$filename)) {
             $asset = Asset::getByPath($parentAsset->getRealFullPath().'/'.$filename);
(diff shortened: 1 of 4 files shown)


bundles/AdminBundle/Controller/Admin/Asset/AssetController.php
```diff
@@ -527,6 +527,17 @@ protected function addAsset(Request $request, Config $config)
...
         }
 
+        // check if there is a requested type and if matches the asset type of the uploaded file
+        $type = $request->get('type');
+        if($type) {
+            $mimetype = MimeTypes::getDefault()->guessMimeType($sourcePath);
+            $assetType = Asset::getTypeFromMimeMapping($mimetype, $filename);
+
+            if($type !== $assetType) {
+                throw new \Exception("Mime type does not match asset type: $type");
```
dvesh3:
```suggestion
                throw new \Exception("Mime type does not match with asset type: $type");
```

dvesh3 (changes requested):
It works like a charm on DataObject Image Data type. We should also consider this for document image editable:
https://github.com/pimcore/pimcore/blob/5bf7298f7a7601da7839e647831030a5f656fbd5/bundles/AdminBundle/Resources/public/js/pimcore/document/editables/image.js#L231

bundles/AdminBundle/Resources/public/js/pimcore/document/editables/image.js
```diff
@@ -242,7 +244,17 @@ pimcore.document.editables.image = Class.create(pimcore.document.editable, {
...
+        function (res) {
+            const response = Ext.decode(res.response.responseText);
+            if (response && response.success === false) {
+                pimcore.helpers.showNotification(t("error"), response.message, "error",
+                    res.response.responseText);
+            } else {
+                pimcore.helpers.showNotification(t("error"), res, "error",
+                    res.response.responseText);
+            }
+        }.bind(this), [] ,this.type);
```
dvesh3:
```suggestion
        }.bind(this), [] ,this.getType());
```
we can kick the newly introduced property 😉

bundles/AdminBundle/Resources/public/js/pimcore/document/editables/image.js
```diff
@@ -14,6 +14,8 @@
 pimcore.registerNS("pimcore.document.editables.image");
 pimcore.document.editables.image = Class.create(pimcore.document.editable, {
 
+    type: "image",
+
```
dvesh3:
```suggestion
```

bundles/AdminBundle/Resources/public/js/pimcore/document/editables/image.js
mattamon: Ah yes, I wanted to do it exactly like in the tag/image.js
Shouldn't we stick to one pattern? We could also use this property for the getType method?

dvesh3: yes, I understand you tried to follow one pattern but then we have to introduce this property for abstract and other editables which requires more refactoring. we can just use the existing method here with less code changes.

dvesh3 (approved):

PR #13729, opened 2022-12-02
Fix xliff import

### LINKED ISSUE #13557: [Bug]: XLIFF Import - document target language not checked
### Expected behavior

When importing a xliff file with documents in it there should be a check if the target language of the document is the same of the current document. If the language doesn't match the entry should either be skipped totally, or the corresponding translation of the document should be resolved, or if the no translation exists at all a new document should be created . 

The behavior could/should be changed via a checkbox or something similar (e.g. _Overwrite if language doesn't match_). Default behavior should be IMO the skipping/resolving of the entries.

### Actual behavior

There is no check if the target-language matches the language of the page. This results in an overwriting of pages in the source language.

### Steps to reproduce

Tested on the pimcore demo
Pimcore 10.5.10

Create new language tree with one of the following options:

- Paste as language variant
- Paste as language variant (recursive)
- Paste language (recursive) ...

image

Select language

image

Export language with "Relations"-Option checked, otherwise no original documents won't be in the export!

image

The new created tree starts with ID 255

image

But there are entries lower than this ID e.g. ID54.

image

If everything is translated, the documents with IDs lower 255 will be overwritten with the translated language.

aryaantony92:
Resolves #13557

### DIFF
--- a/lib/Translation/ImporterService/Importer/DocumentImporter.php
+++ b/lib/Translation/ImporterService/Importer/DocumentImporter.php
@@ -26,6 +26,10 @@ class DocumentImporter extends AbstractElementImporter
      */
     protected function importAttribute(Element\ElementInterface $element, string $targetLanguage, Attribute $attribute)
     {
+        if ($targetLanguage != $element->getProperty ("language")) {
+            return;
+        }
+
         parent::importAttribute($element, $targetLanguage, $attribute);
 
         if ($attribute->getType() === Attribute::TYPE_TAG && $element instanceof Document\PageSnippet) {


kingjia90:
Clearing the milestone, because It would take some more time and work to provide better feedbacks (eg. throw expections, report on which lines got imported) of what's went wrong , since the PR in the current state is just silently skipping.

dvesh3:
> Clearing the milestone, because It would take some more time and work to provide better feedbacks (eg. throw expections, report on which lines got imported) of what's went wrong , since the PR in the current state is just silently skipping.

The PR looks fine as we also skip the entries when attribute is empty or element doesn't exist anymore [here](https://github.com/pimcore/pimcore/blob/30fe7ba313c6ccb82b6bcd94a38785b1386aba25/lib/Translation/ImporterService/Importer/AbstractElementImporter.php#L32). In case of documents, only the target language documents should be imported and skip the relations which are not target specific.

dvesh3 (approved):

PR #13746, opened 2022-12-06
[Bug] Fixed undefined array key issue when exporting localized objectbricks

### LINKED ISSUE #13673: [Bug]: Grid csv-export with localized brick fields causes error 'undefined array key "language"'
### Expected behavior

If the Brick's localized fields are selected in Grid-Config, they must also be exported to CSV.

### Actual behavior

The following error message appears 'undefined array key "language"' and export is broken.

Bildschirmfoto 2022-11-24 um 09 27 49

### Steps to reproduce

- In Grid option add the localized brick fields
- Do not mark any rows in the grid
- Press CSV export

in \Pimcore\Model\DataObject\Service::getCsvFieldData it is not checked whether Brick-Field is localized and no 'language' is added in params-array.

Bildschirmfoto 2022-11-24 um 09 25 40

then the error happens in \Pimcore\Model\DataObject\ClassDefinition\Data::getDataFromObjectParam when calling 
`$data = $classificationStoreData->getLocalizedKeyValue($groupId, $keyId, $language, true, true);`

Corepex:
resolves #13673

### DIFF
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -2072,6 +2072,10 @@ class Service extends Model\Element\Service
                                 if ($brickDescriptor) {
                                     $innerContainer = $brickDescriptor['innerContainer'] ?? 'localizedfields';
                                     $value = $brick->{'get' . ucfirst($innerContainer)}();
+
+                                    if($value instanceof Localizedfield){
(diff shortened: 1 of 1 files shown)


blankse:
@Corepex The target branch should be changed to `10.5`

PR #13748, opened 2022-12-06
[Documents] Fix non-reloading areabricks editable to have correct translation

### LINKED ISSUE #13695: [Bug]: Editable loads static translated messages in wrong language
### Expected behavior

Add a brick (which has needsReload: false) to a document will load all of its translated messages in the correct language set in the current document.

### Actual behavior

The brick does not reflect the document language right after a brick is newly added to the document.

### Steps to reproduce

Set at least 2 languages in the system settings.
Add a new document where bricks can be added and set the language of the document e.g. to German.
Create a brick with a static text that will be translated, e.g.
```twig
<div>{{ 'myTranslation'|trans }}</div>
```
Go to the translation messages and add translations for `myTranslation` so you can tell the translated languages apart from each other.
Go to the document and add the brick => wrong language for the translated text
Reload the document => correct language for the translated text

Sometimes changing the admin language and/or fallback language does the trick if this is not reproducable at first sight.

dvesh3:
Resolves #13695


mattamon (approved):
Works as intended now!

PR #13756, opened 2022-12-07
[Bug]: strip_tags needs empty string as default value, cannot handle null

### LINKED ISSUE #13723: [Bug] Null Perspective when logging in as some other user or in deeplinks
### Expected behavior

https://github.com/pimcore/pimcore/blob/906f47135946b1ea598d268bb957926797126b10/bundles/AdminBundle/src/Security/Authenticator/AdminAbstractAuthenticator.php#L124

maybe `$request->get('perspective') ?? ''` or just append `&perspective=` to the generated url to fix the problem

### Actual behavior

image

Deeplink also affected

### Steps to reproduce

Create a new user
set a password

then press this and copy&paste the generated temporary link in a new browser
image

image

#AND

When using a deeplink eg. `http://localhost/admin/login/deeplink?document_136_page`

mattamon:
Resolves #13723 

## Additional info  
strip_tags cannot handle null values

### DIFF
--- a/bundles/AdminBundle/src/Controller/Admin/LoginController.php
+++ b/bundles/AdminBundle/src/Controller/Admin/LoginController.php
@@ -233,7 +233,7 @@ class LoginController extends AdminController implements KernelControllerEventIn
 
         if (preg_match('/(document|asset|object)_([0-9]+)_([a-z]+)/', $queryString, $deeplink)) {
             $deeplink = $deeplink[0];
-            $perspective = strip_tags($request->get('perspective'));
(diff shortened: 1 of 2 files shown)


blankse:
I think this should be also changed in `10.5` branch

blankse:
@mattamon I created a PR for `10.5` https://github.com/pimcore/pimcore/pull/13757

mattamon:
@blankse You're totally right. Thank you for your PR!

mattamon:
Closing in favor of #13757

Issue #13089, opened 2022-09-07
[Improvement][DataObjects]: Consistent way of storing null/empty textfields to the database


mcop1:
### Improvement description

Empty textfields are stored as `NULL` in the database, when an object gets created. But after editing, saving and removing the value again it will be stored as `''`.

For example: 
1.) Create a new "car"
2.) Use this code to get all cars without description

`$carList = DataObject\Car::getByDescription('', "de");` 

This will result in the following sql:
```
SELECT xxx  FROM object_localized_CAR_de WHERE (`description` = ''  AND  object_localized_CAR_de.o_type IN ('variant','object')) AND object_localized_CAR_de.o_published = 1
```

3.) Check if the newly created car is in the result list
4.) It won´t be, because the description field is set to NULL in the database

5.) Add some text to the description field, save and remove the text again
6.) Description filed in the database then gets set to `''` 
7.) Try the example code from above again, the car should be in the result set now

Noticed during implementation of https://github.com/pimcore/pimcore/pull/13061

jremmurd:
Related to https://github.com/pimcore/pimcore/issues/4369

AndrewKMoses:
We are having issues with this also in the case of using Unique on an attribute. We want to be able to have no value (null) for the attribute on some objects and in those cases we can not update the object after creating it if there are multiple objects with no value for the attribute (because null turns into '' on update).

We had to turn disable the unique check.

dkarlovi:
This seems like a thing which should be solved at client level maybe? The fix in #13760 seems like a direction, but it means it's guessing at what the intent was: API was invoked with an empty string, we know that means `NULL`. Seems messy because what if at some point the empty string needs to be kept as-is?

The client should get to decide when to store with `NULL` and when with an empty string. It should be fixed client-side.

mcop1:
I tried to change the `getDataFromResource` method (because here we can return null according to the return type) to:

`return empty($data) ? null : $data;` 
in
https://github.com/pimcore/pimcore/blob/815ec901d89af3f57f458012993357f269c39eef/models/DataObject/ClassDefinition/Data/Input.php#L126 

Besides that sometimes you may want to save the empty string (like mentioned in the comment above), there is another problem here: 

https://github.com/pimcore/pimcore/blob/4d93f1f93eb8a5544ea4ae8c4345d5fec181b0f3/lib/Model/AbstractModel.php#L187

[...]

PR #13760, opened 2022-12-07
fix: allow textareas to be nullable from admin

dkarlovi:

Makes textareas nullable. Currently, once you edit it, it can never go back to `null`.

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Textarea.php
+++ b/models/DataObject/ClassDefinition/Data/Textarea.php
@@ -254,6 +254,10 @@ class Textarea extends Data implements ResourcePersistenceAwareInterface, QueryR
      */
     public function getDataFromEditmode($data, $object = null, $params = [])
     {
+        if ($data === '') {
+            return null;
+        }
+        
         return $data;
     }
 
--- a/models/DataObject/ClassDefinition/Data/Textarea.php
+++ b/models/DataObject/ClassDefinition/Data/Textarea.php
@@ -250,7 +250,7 @@ class Textarea extends Data implements ResourcePersistenceAwareInterface, QueryR
      * @param null|Model\DataObject\Concrete $object
      * @param mixed $params
      *
-     * @return string
+     * @return string|null
      */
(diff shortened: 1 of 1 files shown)


dkarlovi:
The expectation this always returns a string is weird since the parent is nullable. Is there a reason for this or is it a type error?

Corepex (changes requested):
@dkarlovi could you please change the doctype from `string` to `string|null` - this should fix the phpstan issues 👍 

https://github.com/pimcore/pimcore/blob/10.5/models/DataObject/ClassDefinition/Data/Textarea.php#L253

dkarlovi:
@Corepex 

> @dkarlovi could you please change the doctype from `string` to `string|null` - this should fix the phpstan issues +1

Yes, it's very clear how we can fix PHPStan complaining. the question was more why the type was as narrow as it was? If it's expected to be this narrow, we cannot just widen it since it's not invariant.

mcop1:
Related to https://github.com/pimcore/pimcore/issues/13089

Corepex:
> @Corepex
> 
> > @dkarlovi could you please change the doctype from `string` to `string|null` - this should fix the phpstan issues +1
> 
> Yes, it's very clear how we can fix PHPStan complaining. the question was more why the type was as narrow as it was? If it's expected to be this narrow, we cannot just widen it since it's not invariant.

I think adding `null` shouldn't be a problem since the base class (`Data`) returns `mixed` ...

Corepex:
@dkarlovi friendly reminder ⏰

brusch:
@dkarlovi 🏓😊 Thanks!

brusch:
@dkarlovi nevermind, just pushed the fix directly to your PR 😉

dkarlovi:
thank you @brusch! \o/

mcop1:
@dkarlovi Thank you for contributing. We discussed about this pr internally and we see it as kind of a change in behavior. That´s why we would like to do it in 11.x with this pr #13963.

PR #13763, opened 2022-12-07
[Bug]: Fix problem in grid configuration with invisble block

frithjof:
If you create a class with a block that is invisible and has visible fields inside the grid configuration fails because it tries to add the subfields to non existing invisible block node


robertSt7:
@frithjof Could you please rebase the PR to 10.5? Thanks

frithjof:
done :)

robertSt7:
@frithjof Thanks for the PR

PR #13778, opened 2022-12-12
Avoid unecessary cast to string in Translator::trans()

NiklasBr:
Incoming parameter to method is typed to `string` and the `trim()` function is [guaranteed to return a string](https://www.php.net/manual/en/function.trim) anyway.


brusch:
Thanks a lot! 👍

PR #13781, opened 2022-12-12
Extracted file explorer to a separate bundle

### LINKED ISSUE #12521: [Task] Move File Explorer to repository pimcore/file-explorer-bundle
### Improvement description

Move `File Explorer` to repository `pimcore/file-explorer-bundle`

aashan10:
Resolves #12521


bundles/FileExplorerBundle/config/pimcore/routing.yaml
```diff
@@ -0,0 +1,6 @@
+pimcore_admin_file_explorer:
+    resource: "../../src/Controller/Admin/"
+    type:     annotation
+    prefix:   /admin
```
mattamon: We should add the bundle name to the prefix eg. /admin/fileexplorer to make it more consistent with the other bundle extractions.

bundles/FileExplorerBundle/config/pimcore/routing.yaml
```diff
@@ -0,0 +1,6 @@
+pimcore_admin_file_explorer:
```
mattamon: For consistency add a _ before the name

bundles/FileExplorerBundle/public/js/explorer.js
```diff
@@ -11,8 +11,8 @@
  * @license    http://www.pimcore.org/license     GPLv3 and PCL
  */
 
-pimcore.registerNS("pimcore.settings.fileexplorer.explorer");
-pimcore.settings.fileexplorer.explorer = Class.create({
+pimcore.registerNS("pimcore.fileexplorer.explorer");
+pimcore.fileexplorer.explorer = Class.create({
```
mattamon: In the glossary bundle we try to keep the settings, maybe we can do this here too?

bundles/FileExplorerBundle/public/js/file.js
```diff
@@ -11,8 +11,8 @@
  * @license    http://www.pimcore.org/license     GPLv3 and PCL
  */
 
-pimcore.registerNS("pimcore.settings.fileexplorer.file");
-pimcore.settings.fileexplorer.file = Class.create({
+pimcore.registerNS("pimcore.fileexplorer.file");
+pimcore.fileexplorer.file = Class.create({
```
mattamon: Same here we should keep the settings

bundles/FileExplorerBundle/public/js/startup.js
```diff
@@ -0,0 +1,43 @@
...
+            perspectiveCfg.inToolbar("extras.systemtools") &&
+            perspectiveCfg.inToolbar("extras.systemtools.fileexplorer")
+        ) {
+            const index = toolbar.extrasMenu.items.keys.indexOf('pimcore_menu_extras_system_info');
+            const systemInfoMenu  = toolbar.extrasMenu.items.items[index];
+
+            systemInfoMenu.getMenu().add(
+                {
+                    text: t("server_fileexplorer"),
+                    iconCls: "pimcore_nav_icon_fileexplorer",
```
mattamon: We should move the css parts to the bundle too, so maybe move the icons to its own icons.css in the bundle

bundles/FileExplorerBundle/src/Controller/Admin/AdminController.php
```diff
@@ -0,0 +1,261 @@
...
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @Route("/file-explorer")
+ *
+ * @internal
+ */
+class AdminController extends \Pimcore\Bundle\AdminBundle\Controller\AdminController
```
mattamon: I am not sure if we should put all in one controller. Maybe we should keep the original structure a little more and move it to a MiscController instead? At least that is what I did e.g. for the glossary. Wdyt?

bundles/FileExplorerBundle/src/FileExplorerBundle.php
```diff
@@ -0,0 +1,44 @@
...
+use Pimcore\Extension\Bundle\Traits\PackageVersionTrait;
+use Pimcore\Bundle\FileExplorerBundle\DependencyInjection\FileExplorerExtension;
+use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
+
+class FileExplorerBundle extends AbstractPimcoreBundle
+{
+
+    use PackageVersionTrait;
+
+    public function getComposerPackageName(): string
```
mattamon: I guess this is for future references right? 
Are there plans to move it to its own repo or will it be kept under bundles?

mattamon (changes requested):
For reference: https://github.com/pimcore/pimcore/pull/13811/files
We need to add the bundle in .github/ci/files/config/bundle so it gets installed in the tests

Move css parts into the bundle as well.

We do need a migration so the bundle will be marked as installed. It should also have a warning that you have to enable the bundle in the bundles.php -> see glossary bundle pr

It also needs and Installer, to add or remove the permissions I think.

bundles/FileExplorerBundle/src/FileExplorerBundle.php
aashan10: Yes! File Explorer is supposed to be a separate composer package.

bundles/FileExplorerBundle/src/Controller/Admin/AdminController.php
aashan10: The methods inside the AdminController serve a single purpose of handling the actions for creating/updating/deleting the files inside the file-explorer tree. I think it makes more sense to keep them together and not split into multiple places as debugging in future can be painful because of that

bundles/FileExplorerBundle/config/pimcore/routing.yaml
```diff
@@ -0,0 +1,6 @@
+_pimcore_admin_file_explorer:
+    resource: "../../src/Controller/Admin/"
+    type:     annotation
+    prefix:   /admin/fileexplorer
```
mattamon: Add the prefix to the routes e.g.  prefix:   /admin/bundle/glossary in the [routing.yaml](https://github.com/pimcore/pimcore/blob/4d3b5dd592472b9e02fa1fec4ddb09d81b3e9afb/bundles/GlossaryBundle/config/pimcore/routing.yaml)

bundles/FileExplorerBundle/public/js/startup.js
```diff
@@ -0,0 +1,44 @@
+pimcore.registerNS("pimcore.settings.fileexplorer");
+
+pimcore.settings.fileexplorer = Class.create({
+    initialize: function () {
+        document.addEventListener(pimcore.events.pimcoreReady, this.pimcoreReady.bind(this));
+    },
+
+
+    pimcoreReady: function(e) {
```
mattamon: Please merge the latest changes from 11.x with the Glossarybundle and update the event listeners accordingly.
Checkout the startup.js from the glossary bundle https://github.com/pimcore/pimcore/blob/11.x/bundles/GlossaryBundle/public/js/startup.js

Please also make sure to update the namespaces of the js files like `pimcore.registerNS("pimcore.bundle.glossary.startup");` and `pimcore.bundle.glossary.startup = Class.create({`

Please also mark every js file as private as you can see in the startup.js from glossary.

A priority for the menu items should also be provided. 

See also the settings.js from the glossary for more info https://github.com/pimcore/pimcore/blob/11.x/bundles/GlossaryBundle/public/js/settings.js

bundles/FileExplorerBundle/src/Controller/Admin/AdminController.php
```diff
@@ -0,0 +1,261 @@
...
+{
+    /**
+     * @Route("/tree", name="pimcore_file_explorer_tree", methods={"GET"})
```
mattamon: Change the route names accordingly e.g. pimcore_bundle_glossary_settings_glossary. See [SettingsController](https://github.com/pimcore/pimcore/blob/4d3b5dd592472b9e02fa1fec4ddb09d81b3e9afb/bundles/GlossaryBundle/src/Controller/SettingsController.php)

bundles/FileExplorerBundle/src/Installer.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        $db = \Pimcore\Db::get();
+        foreach (self::USER_PERMISSIONS as $permission) {
+            $db->insert('users_permission_definitions', [
+                $db->quoteIdentifier('key') => $permission,
+            ]);
+        }
+    }
+
+    protected function revokePermissions()
```
mattamon:
```suggestion
    protected function removePermissions()
```

bundles/FileExplorerBundle/src/Installer.php
```diff
@@ -0,0 +1,43 @@
...
+    ];
+
+    public function install()
+    {
+        $this->addPermissions();
+    }
+
+    public function uninstall()
+    {
+        $this->revokePermissions();
```
mattamon:
```suggestion
        $this->removePermissions();
```

bundles/FileExplorerBundle/src/Installer.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Pimcore\Bundle\FileExplorerBundle;
+
+use Pimcore\Extension\Bundle\Installer\SettingsStoreAwareInstaller;
+
+class Installer extends SettingsStoreAwareInstaller
```
mattamon, line 7: The installer must be defined in the services.yaml https://github.com/pimcore/pimcore/blob/11.x/bundles/GlossaryBundle/config/services.yaml#L10

mattamon (changes requested):
If the bundle stay in the core (not have its own repo) then do not extract docs & test files (discuss if changes are big). Only extract en translations (en.json and en.extended.json) to the bundle and keep the rest in the core, as these will be cleaned up by POEditor sync job. See extraction [here](https://github.com/pimcore/pimcore/blob/b325af277864154d99deafad7b15afe85c7ee4c8/bundles/GlossaryBundle/translations/admin.en.json)

Bundle specific config should be extracted into own node e.g. pimcore_glossary or pimcore_admin

Add the moved bundle to the Upgrade Notes under [Bundles](https://github.com/pimcore/pimcore/blob/333f0de246f9bd023ec2caa91e0b22061451ccf2/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md)

brusch:
@mattamon I wouldn't enable this bundle in the demo, as it isn't really useful for demo purposes. 
Also the bundle will be extracted into it's own repo, see https://github.com/pimcore/pimcore/issues/12094

mattamon:
@brusch I updated my comment. 
What about the migration then? Should it be marked as installed by default or should we discard the migration?

brusch:
@mattamon yep, no migration for that.

mattamon:
@brusch what about upgrading from 10 to 11? everyone who has the fileexplorer permission won't be able to use it anymore without a migration? so maybe we should mark the bundle as installed if someone has the fileexplorer permission?

brusch:
@mattamon that's true, but imho that's fine because it doesn't break anything, it's UI only. 😊

brusch:
@mattamon additionally it would be very complicated and error prone, because the bundle will be in it's own repository, therefore it would require a `composer require pimcore/file-explorer-bundle` for the migration ... I wouldn't do migration for all bundles which will be moved to their own repo.

mattamon:
@brusch yes I agree that it won't be as much of a problem and very complicated. 
I will update the refactoring tips.

bundles/FileExplorerBundle/config/services.yaml
```diff
@@ -0,0 +1,15 @@
...
+    _defaults:
+        autowire: true
+        autoconfigure: true
+        public: false
+
+    Pimcore\Bundle\FileExplorerBundle\Controller\Admin\:
+        resource: '../src/Controller/Admin'
+        public: true
+        tags: [ 'controller.service_arguments' ]
+    Pimcore\Bundle\FileExplorerBundle\Installer:
```
mattamon, line 13:
```suggestion

    Pimcore\Bundle\FileExplorerBundle\Installer:
```

We should add a space here

bundles/FileExplorerBundle/public/js/startup.js
```diff
@@ -0,0 +1,97 @@
...
+                iconCls: "pimcore_nav_icon_fileexplorer",
+                itemId: 'pimcore_menu_extras_system_info_server_fileexplorer',
+                handler: this.showFileExplorer
+            });
+        }
+
+        return items;
+    },
+
+    pimcoreReady: function(e) {
```
mattamon: We don't need the pimcoreReady function anymore.

bundles/FileExplorerBundle/public/js/startup.js
```diff
@@ -0,0 +1,97 @@
...
+
+        if (
+            user.admin &&
+            perspectiveCfg.inToolbar("extras") &&
+            perspectiveCfg.inToolbar("extras.systemtools") &&
+            perspectiveCfg.inToolbar("extras.systemtools.fileexplorer")
+        ) {
+            items.push({
+                text: t("server_fileexplorer"),
+                iconCls: "pimcore_nav_icon_fileexplorer",
```
mattamon, line 40: Should we add the priority for the menu here or should it always be shown at the bottom?

bundles/FileExplorerBundle/public/js/startup.js
```diff
@@ -0,0 +1,97 @@
...
+
+        if (filteredMenu.length > 0) {
+            const systemInfoMenu = filteredMenu.shift();
+            systemInfoMenuItems.map(function(item) {
+                systemInfoMenu.menu.items.push(item);
+            });
+        } else {
+            menu.extras.items.push({
+                text: t("system_infos_and_tools"),
+                iconCls: "pimcore_nav_icon_info",
```
mattamon: Same for the priority here right? It should just be added at the bottom?

bundles/FileExplorerBundle/src/Installer.php
```diff
@@ -0,0 +1,60 @@
...
+
+    protected function addPermissions()
+    {
+        $db = \Pimcore\Db::get();
+
+        /**
+         * The following code is here for backwards compatibility reason.
+         * If the permissions are already installed, the installer should not generate
+         * any further errors and only add permissions which
+         */
```
mattamon: @brusch this is a good point. If we do not mark the bundle as installed via a migration, and someone upgrades from 10 to 11 and tries to install the file explorer bundle, but the permission is already there the installation of the bundle will fail.

Maybe we should remove this permission via a migration for all bundles that will get its own repo?

bundles/FileExplorerBundle/public/js/file.js
```diff
@@ -11,11 +11,11 @@
  * @license    http://www.pimcore.org/license     GPLv3 and PCL
  */
 
-pimcore.registerNS("pimcore.settings.fileexplorer.file");
+pimcore.registerNS("pimcore.bundle.file_explorer.settings.file");
```
mattamon: @brusch how should we tackle namespaces like these? 
Should we keep the underscore or replace it with a dash or dot?

mattamon (changes requested):
Please check the tests. Those are still failing because of the installer. :)

bundles/FileExplorerBundle/src/Installer.php
brusch: @mattamon @aashan10 yes, we should do so! 👍

bundles/FileExplorerBundle/public/js/file.js
brusch: @aashan10 @mattamon dash won't work (not allowed for js objects), I'd keep the underscore here.

bundles/FileExplorerBundle/src/Installer.php
mattamon: @brusch We should update the refactoring tips once again. Should this be the standard, to remove permissions via migrations if the bundle gets its own repo?

@aashan10 Could you please create a Migration in the Core to remove the permission. I think we then do not need all the compatibility code in the installer.

bundles/CoreBundle/src/Migrations/Version20230112203416.php
```diff
@@ -0,0 +1,41 @@
...
+    public function up(Schema $schema): void
+    {
+        if (SettingsStore::get('BUNDLE_INSTALLED__Pimcore\\Bundle\\FileExplorerBundle\\PimcoreFileExplorerBundle', 'pimcore')) {
+            SettingsStore::set('BUNDLE_INSTALLED__Pimcore\\Bundle\\FileExplorerBundle\\PimcoreFileExplorerBundle', false, 'bool', 'pimcore');
+        }
+
+        $db = \Pimcore\Db::get();
+        $permissionsTable = `users_permission_definitions`;
+
+        $query = sprintf('DELETE FROM `%s` WHERE `key` IN (%s);', $permissionsTable, implode(',', ['fileexplorer']));
```
mattamon: I did check the permissions table and there is no extra permission for the fileexplorer so we do not need that migration at all.

bundles/FileExplorerBundle/composer.json
```diff
@@ -0,0 +1,27 @@
...
+  "type": "pimcore-bundle",
+  "description": "Pimcore File Explorer Bundle that allows to read/write files on the server from within the Pimcore Admin.",
+  "config": {
+    "sort-packages": true
+  },
+  "prefer-stable": true,
+  "minimum-stability": "dev",
+  "require": {
+    "php": "^8.0 || ^8.1",
+    "pimcore/pimcore": "^10.6 || ^11.0"
```
mattamon: I think we can just require 11 here, right?
@dvesh3 wdyt?

bundles/FileExplorerBundle/public/js/startup.js
mattamon: Can we still add the priority here please?

bundles/FileExplorerBundle/src/PimcoreFileExplorerBundle.php
```diff
@@ -0,0 +1,54 @@
...
+
+    public function getPath(): string
+    {
+        return dirname(__DIR__);
+    }
+
+    public function getInstaller(): Installer
+    {
+        /** @var Installer $installer */
+        $installer = $this->container->get(Installer::class);
```
mattamon: We can remove this `$installer` variable

mattamon (changes requested):

bundles/CoreBundle/src/Migrations/Version20230112203416.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
mattamon: Like I wrote in my comment before, we do not need the Migration at all. You can remove the whole file.

mattamon (changes requested):

bundles/FileExplorerBundle/public/js/startup.js
```diff
@@ -44,7 +44,7 @@ pimcore.bundle.file_explorer.startup = Class.create({
         var perspectiveCfg = pimcore.globalmanager.get("perspective");
 
         if (
-            user.admin &&
+            (user.isAllowed('fileexplorer') || user.admin )&&
```
mattamon: The new permission was fine. Please add it back again, since we introduce the new permission in the installer.

bundles/FileExplorerBundle/src/Installer.php
```diff
@@ -10,19 +10,19 @@ class Installer extends SettingsStoreAwareInstaller
         'fileexplorer'
     ];
 
-    public function install()
+    public function install(): void
     {
         $this->addPermissions();
```
mattamon, line 17: Can you please also give the permissions a category like here? https://github.com/pimcore/pimcore/pull/14039/files#diff-2828c0270e1810c7d1d830a4a8598641b7d2cc717ac45500babdba3ef9a88d20R23

mattamon (changes requested):

mattamon (changes requested):
Cann you please add the moved bundle to the Upgrade Notes under [Bundles](https://github.com/pimcore/pimcore/blob/333f0de246f9bd023ec2caa91e0b22061451ccf2/doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md)

aashan10:
@mattamon done! Thanks :)

mattamon (approved):
LGTM now. We should maybe wait for the tests to run through again before merging.

mattamon (changes requested):
Please resolve the conflicts and then it is ready to merge IMHO

bundles/FileExplorerBundle/composer.json
```diff
@@ -0,0 +1,27 @@
...
+  "license": "GPL-3.0+",
+  "type": "pimcore-bundle",
+  "description": "Pimcore File Explorer Bundle that allows to read/write files on the server from within the Pimcore Admin.",
+  "config": {
+    "sort-packages": true
+  },
+  "prefer-stable": true,
+  "minimum-stability": "dev",
+  "require": {
+    "php": "^8.0 || ^8.1",
```
dvesh3:
```suggestion
    "php": "^8.1",
```

bundles/FileExplorerBundle/public/js/startup.js
```diff
@@ -0,0 +1,73 @@
...
+        const menu = event.detail.menu;
+        this.user = pimcore.globalmanager.get('user');
+        this.toolbar = pimcore.globalmanager.get('layout_toolbar');
+        const systemInfoMenuItems = this.getSystemInfoMenu();
+
+        const filteredMenu = menu.extras.items.filter(function (item) {
+            return item.itemId === 'pimcore_menu_extras_system_info';
+        });
+
+        if (filteredMenu.length > 0) {
```
dvesh3: No need to check and create main menu anymore as the with new approach the menu is always there. please see https://github.com/pimcore/pimcore/pull/13887/commits/a7a6083f190e384fa0cd1b1a97c85d25b5fb6c09#diff-594df434516f523fb7936f5c0652aaa376b5b4e4a10f410e423aa7926a1cdad6

bundles/FileExplorerBundle/src/Controller/Admin/AdminController.php
```diff
@@ -0,0 +1,261 @@
...
+
+namespace Pimcore\Bundle\FileExplorerBundle\Controller\Admin;
+
+use Pimcore\File;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @Route("/")
```
dvesh3:
```suggestion
 * @Route("/admin")
```
Also change the route names.

bundles/FileExplorerBundle/public/js/file.js
dvesh3: Either move the js file under `/settings` folder or drop `settings` from namespace.

bundles/FileExplorerBundle/public/js/explorer.js
```diff
@@ -11,19 +11,19 @@
  * @license    http://www.pimcore.org/license     GPLv3 and PCL
  */
 
-pimcore.registerNS("pimcore.settings.fileexplorer.explorer");
+pimcore.registerNS("pimcore.bundle.file_explorer.settings.explorer");
```
dvesh3: Either move the js file under /settings folder or drop settings from namespace.

bundles/FileExplorerBundle/src/Controller/Admin/AdminController.php
mattamon: The full route then looks something like /admin/bundle/file-explorer/admin/tree
Do we really need to call it admin controller? Can't we just rename it to SettingsController?

bundles/FileExplorerBundle/public/js/startup.js
mattamon: We still have to do it because System Settings & Tools is now it's own bundle, so if it is disabled, we still need to add the menu item.

bundles/FileExplorerBundle/src/Controller/Admin/AdminController.php
dvesh3: Good point. I also vote for `SettingsController` and move it out of `/Admin` folder

bundles/FileExplorerBundle/public/js/startup.js
dvesh3: main menu is still in the core https://github.com/pimcore/pimcore/blob/3cde75bc431119461c302a4b22b6808b21c34192/bundles/AdminBundle/public/js/pimcore/layout/toolbar.js#L411

mattamon: Ah yes sorry you are right :) 
Removed it.

dvesh3 (approved):

PR #13789, opened 2022-12-13
Some type fixes

fashxp:
Part of https://github.com/pimcore/pimcore/issues/13656

### DIFF
--- a/models/DataObject/Data/Video.php
+++ b/models/DataObject/Data/Video.php
@@ -31,7 +31,7 @@ class Video implements OwnerAwareFieldInterface
 
     protected string|int|ElementInterface|Asset|\Pimcore\Model\Element\ElementDescriptor $data;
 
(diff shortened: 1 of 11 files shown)


models/DataObject/Data/Video.php
```diff
@@ -31,7 +31,7 @@ class Video implements OwnerAwareFieldInterface
 
     protected string|int|ElementInterface|Asset|\Pimcore\Model\Element\ElementDescriptor $data;
 
-    protected string|int|Asset|\Pimcore\Model\Element\ElementDescriptor $poster;
+    protected string|int|Asset|\Pimcore\Model\Element\ElementDescriptor|null $poster;
```
blankse:
```suggestion
    protected string|int|Asset|\Pimcore\Model\Element\ElementDescriptor|null $poster = null;
```

mcop1:
Looks good, but we should check the corresponding PhpDocs in 10.6 and adapt them if necessary.

PR #13791, opened 2022-12-14
[Task] Substitute `method_exists` with interface check in fieldDefinitions

### LINKED ISSUE #12861: [Task]: Substitute "ClassSavedTrait" with interface
### Improvement description

Follow up to #9439 and #9571

Remove ClassSavedTrait and add a corresponding interface. 
Then remove "method_exists" call and check for the interface instead, e.g.
https://github.com/pimcore/pimcore/blob/fcd078eccbb0600922db3857be6c01bd2d32be65/models/DataObject/ClassDefinition/Data/Fieldcollections.php#L681

Corepex:
resolves #12861


models/DataObject/ClassDefinition/Data/Fieldcollections.php
```diff
@@ -686,8 +686,7 @@ public function classSaved($class, $params = [])
                     $fieldDefinition = $definition->getFieldDefinitions();
 
                     foreach ($fieldDefinition as $fd) {
-                        //TODO Pimcore 11 remove method_exists call
-                        if (!$fd instanceof DataContainerAwareInterface && method_exists($fd, 'classSaved')) {
+                        if (!$fd instanceof DataContainerAwareInterface && $fd instanceof ClassSavedInterface) {
```
blankse: BC Break, please add a BC layer
```suggestion
                        if (!$fd instanceof DataContainerAwareInterface && ($fd instanceof ClassSavedInterface || method_exists($fd, 'classSaved'))) {
                            if (!$fd instanceof ClassSavedInterface) {
                                trigger_deprecation('pimcore/pimcore', '10.6',
                                    sprintf('Usage of method_exists is deprecated since version 10.6 and will be removed in Pimcore 11.' .
                                    'Implement the %s interface instead.', ClassSavedInterface::class));
                            }
```
Also please add a Update Note.

Corepex: Why should this be a BC break? each class implements the interface and the corresponding method ...

blankse: Every class in this repro. But devs can create own data types.

Corepex: Your right - will provide a BC layer for that 👍

blankse:
@Corepex The same change must be here:
https://github.com/pimcore/pimcore/blob/e6ca020773eff957ac14dd459e168a2a808d82ba/models/DataObject/ClassDefinition/Data/Objectbricks.php#L905-L911

and here:
https://github.com/pimcore/pimcore/blob/e6ca020773eff957ac14dd459e168a2a808d82ba/models/DataObject/ClassDefinition/Data/Localizedfields.php#L574-L577

The params are only set in Localizedfields. Is it right not set in Fieldcollections/Objectbricks?

models/DataObject/ClassDefinition/Data/Fieldcollections.php
brusch: @Corepex Please also create a follow-up issue for `11.x` to remove the BC layer 😉

Corepex:
follow-up: #13798

models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php
```diff
@@ -774,8 +774,9 @@ public function getColumnKeys()
 
     /**
      * @param DataObject\ClassDefinition $class
+     * @param array $params
      */
-    public function classSaved($class)
+    public function classSaved($class, $params = [])
```
jdreesen: Technically, this is a BC break, too, as this class/method is neither `@internal` nor `final`, so someone *could* extend it, which would break after this change.

models/DataObject/ClassDefinition/Data/Localizedfields.php
```diff
@@ -572,7 +572,7 @@ public function classSaved($class, $params = [])
 
         foreach ($this->getFieldDefinitions() as $fd) {
             //TODO Pimcore 11 remove method_exists call
-            if (!$fd instanceof DataContainerAwareInterface && method_exists($fd, 'classSaved')) {
+            if (!$fd instanceof DataContainerAwareInterface && ($fd instanceof ClassSavedInterface || method_exists($fd, 'classSaved'))) {
```
blankse:
```suggestion
            if (!$fd instanceof DataContainerAwareInterface && ($fd instanceof ClassSavedInterface || method_exists($fd, 'classSaved'))) {
                if (!$fd instanceof ClassSavedInterface) {
                    trigger_deprecation('pimcore/pimcore', '10.6',
                        sprintf('Usage of method_exists is deprecated since version 10.6 and will be removed in Pimcore 11.' .
                        'Implement the %s interface instead.', ClassSavedInterface::class));
                }
```

models/DataObject/ClassDefinition/Data/Objectbricks.php
```diff
@@ -903,7 +903,7 @@ public function classSaved($class, $params = [])
 
                     foreach ($fieldDefinition as $fd) {
                         //TODO Pimcore 11 remove method_exists call
-                        if (method_exists($fd, 'classSaved')) {
+                        if ($fd instanceof ClassSavedInterface || method_exists($fd, 'classSaved')) {
```
blankse:
```suggestion
                        if ($fd instanceof ClassSavedInterface || method_exists($fd, 'classSaved')) {
                            if (!$fd instanceof ClassSavedInterface) {
                                trigger_deprecation('pimcore/pimcore', '10.6',
                                    sprintf('Usage of method_exists is deprecated since version 10.6 and will be removed in Pimcore 11.' .
                                    'Implement the %s interface instead.', ClassSavedInterface::class));
                            }
```

blankse (changes requested):
Also add the deprecation warnings to Localizedfields and Objectbricks

models/DataObject/ClassDefinition/Data/Block.php
```diff
@@ -26,7 +26,7 @@
 use Pimcore\Normalizer\NormalizerInterface;
 use Pimcore\Tool\Serialize;
 
-class Block extends Data implements CustomResourcePersistingInterface, ResourcePersistenceAwareInterface, LazyLoadingSupportInterface, TypeDeclarationSupportInterface, VarExporterInterface, NormalizerInterface, DataContainerAwareInterface, PreGetDataInterface, PreSetDataInterface
+class Block extends Data implements CustomResourcePersistingInterface, ResourcePersistenceAwareInterface, LazyLoadingSupportInterface, TypeDeclarationSupportInterface, VarExporterInterface, NormalizerInterface, DataContainerAwareInterface, PreGetDataInterface, PreSetDataInterface, ClassSavedInterface
```
blankse: @Corepex Make it sense to add `DataContainerAwareInterface` and `ClassSavedInterface` in one class? I think we should only set once. So we can remove the `!DataContainerAwareInterface` check in 11.x

models/DataObject/ClassDefinition/Data/Objectbricks.php
```diff
@@ -903,10 +903,15 @@ public function classSaved($class, $params = [])
...
+                            if (!$fd instanceof ClassSavedInterface) {
+                                trigger_deprecation('pimcore/pimcore', '10.6',
+                                    sprintf('Usage of method_exists is deprecated since version 10.6 and will be removed in Pimcore 11.' .
+                                    'Implement the %s interface instead.', ClassSavedInterface::class));
+                            }
                             // defer creation
                             if (!$fd instanceof DataContainerAwareInterface) {
```
blankse: We should add this check before `$fd instanceof ClassSavedInterface || method_exists($fd, 'classSaved')` like in the other classes. So we get no deprecation warnings for `DataContainerAwareInterface` classes without `ClassSavedInterface`

models/DataObject/ClassDefinition/Data/Block.php
Corepex: Do you mean to merge the `DataContainerAwareInterface` and the `ClassSavedInterface` into one file or to remove the `DataContainerAwareInterface` completely from that class?

`DataContainerAwareInterface` implements `preSave` and `postSave`
`ClassSavedInterface` implements `classSaved`

blankse: But the classSaved method of `ClassSavedInterface` will only executed if the class is not a instance of `DataContainerAwareInterface`. So I would not add it to this classes because it have no effect. We can remove the `!DataContainerAwareInterface` in 11.x when we only add it to non `DataContainerAwareInterface` classes.

Corepex: But for example, here we call the `preSave` method on each $fielddefinition and therefore, we check for the `DataContainerAwareInterface` to verify the method - this $fielddefinition could also be a `block` or?

https://github.com/pimcore/pimcore/blob/c8e3a921b3dca7779a6ae951c65185a922b763ed/models/DataObject/ClassDefinition.php#L426-L435

blankse: You should not remove the `DataContainerAwareInterface` interface but the `ClassSavedInterface` interface.

The ClassSavedInterface interface should only added for:
- AdvancedManyToManyRelation
- Classificationstore

Following classes should NOT have the ClassSavedInterface interface:
- Block
- Fieldcollections
- Localizedfields
- Objectbricks

Corepex: Ah, sorry now I get it - that makes sense true 👍

Corepex: > So we can remove the `!DataContainerAwareInterface` check in 11.x

updated the follow-up ticket to also remove the `!DataContainerAwareInterface` check

models/DataObject/ClassDefinition/Data/Fieldcollections.php
```diff
@@ -687,13 +687,18 @@ public function classSaved($class, $params = [])
...
+                                    'Implement the %s interface instead.', ClassSavedInterface::class));
+                            }
                             // defer creation
-                            $fd->classSaved($class);
+                            $fd->classSaved($class, $params);
                         }
                     }
 
-                    $definition->getDao()->classSaved($class);
+                    $definition->getDao()->classSaved($class, $params);
```
blankse: This method have no params argument
```suggestion
                    $definition->getDao()->classSaved($class);
```

models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php
brusch:
```suggestion
    public function classSaved($class/**, $params = [] **/)
```
We should keep the definition commented out and introduce it in v11. 
For compatibility we can use `func_get_args()` to fill `$params` in v10.6.

brusch (changes requested):
Please also don't forget about the upgrade notes and docs 😊

models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php
blankse: But then we can not use the interface and get deprecation warnings with this data type, right?

Corepex: The `ClassSavedInterface` interface requires the `$params`

brusch: Why not? I mean the interface should also have the same signature (with the commented `$params`)  - then it shouldn't be a problem. Then in v11, we change both the interface and the implementations.

blankse: But then the interface is not compatible with the other field types, or can a child method have more parameters than the interface?

blankse: I think we can make the interface without methods in 10.6 (only a `@method` annotation). This should work.

brusch: @blankse right, this is what I meant 😉

jdreesen: > But then the interface is not compatible with the other field types, or can a child method have more parameters than the interface?

Yes, that's possible, if the additional parameters are at the end and have a default value (which is the case here). 
See: https://3v4l.org/F2hKa

models/DataObject/ClassDefinition/Data/ClassSavedInterface.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Pimcore\Model\DataObject\ClassDefinition\Data;
+
+use Pimcore\Model\DataObject;
+
+interface ClassSavedInterface
+{
+    /**
+     * @param $class
```
jdreesen: This is missing a type.

Corepex: true that thx 👍

models/DataObject/ClassDefinition/Data/AdvancedManyToManyRelation.php
Corepex: None of our classes (that implement the interface) really uses the `$param` parameter - a custom implementation must therefore use the default value (as seen in your demo) or don't use the `$params` parameter at all ....

models/DataObject/ClassDefinition/Data/Objectbricks.php
Corepex: hope this is what you meant https://github.com/pimcore/pimcore/pull/13791/commits/3ff72c476dfd034e4e0af952fbdd5e73e9dac692

blankse: Yes 👍🏼

robertSt7:
closed by https://github.com/pimcore/pimcore/pull/13846

PR #13792, opened 2022-12-14
replace lrm characters with - for getValidKey

frithjof:
If you have a string with left to right marker character you get an invalid key

reproduce
create string with lrm character $imagename

$prod_image_folder = new Asset\Folder();
$img_key = Service::getValidKey($imagename , 'asset');
$prod_image_folder->setKey(Service::getValidKey(img_key , 'asset'));
$prod_image_folder->setParentId(1);
$prod_image_folder->save();

you get a flysystem error because key is not valid in file system


dvesh3:
thanks @frithjof

PR #13811, opened 2022-12-15
12517 move glossary to bundle

### LINKED ISSUE #12517: [Task] Move Glossary to GlossaryBundle within bundles folder in pimcore/pimcore repository
### Improvement description

Move Glossary to `GlossaryBundle` within bundles folder in `pimcore/pimcore` repository

mattamon:
Resolves #12517

## Additional info  
- Moving glossary functionality into GlossaryBundle
- Able to install and uninstall via Installer. Permissions are in the Bundle
- User Keybindings are now configurable and can be added from anywhere via event
- Menu Items are now added via Eventlistener pimcore.events.preMenuBuild.
- Menus do now have a priority to figure out the sorting. Standard pimcore menus priority increases by 10 to leave enough space for additonal menu items.


mattamon:
What should we do with the documentation?
https://pimcore.com/docs/pimcore/11.0/Development_Documentation/Tools_and_Features/Glossary.html
https://github.com/pimcore/pimcore/blob/11.x/doc/Development_Documentation/18_Tools_and_Features/21_Glossary.md

Is there a way to we can sync this in the documentation after we move it?

brusch:
@mattamon since the glossary bundle will stay in this repo, we keep the docs for this in place.
For bundles that will be separated out into their own repo, we move the docs as well, just as we did with other bundles (DataHub, ...)

bundles/GlossaryBundle/public/css/glossary.css
```diff
@@ -0,0 +1,7 @@
+.pimcore_icon_glossary {
```
dvesh3: filename `icons.css`?

bundles/GlossaryBundle/public/js/glossary.js
```diff
@@ -11,8 +11,8 @@
  * @license    http://www.pimcore.org/license     GPLv3 and PCL
  */
 
-pimcore.registerNS("pimcore.settings.glossary");
-pimcore.settings.glossary = Class.create({
+pimcore.registerNS("pimcore.glossary.glossary");
```
dvesh3: can we still use `settings`?

bundles/GlossaryBundle/src/Controller/SettingsController.php
```diff
@@ -0,0 +1,145 @@
...
+use Pimcore\Bundle\AdminBundle\Controller\AdminController;
+use Pimcore\Bundle\GlossaryBundle\Model\Glossary;
+use Pimcore\Cache;
+use Pimcore\Model\Document;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @Route("/settings")
```
dvesh3, line 29: we are using prefix in the routing.yaml, so the complete route will be `/admin/glossary/settings/glossary` and route name changed to `pimcore_settings_glossary`. Either we keep using the admin prefix in the name or remove it from the actual route prefix. wdyt?

bundles/GlossaryBundle/src/DependencyInjection/GlossaryExtension.php
```diff
@@ -0,0 +1,47 @@
...
+
+        $loader = new YamlFileLoader(
+            $container,
+            new FileLocator(__DIR__ . '/../../config')
+        );
+
+        $loader->load('services.yaml');
+    }
+
+    public function prepend(ContainerBuilder $container): void
```
dvesh3: If this is not implemented, then we can kick `PrependExtensionInterface`

bundles/GlossaryBundle/src/Migrations/Version20221215071650.php
```diff
@@ -0,0 +1,37 @@
...
+        return 'In case the glossary table already exists, it marks the GlossaryBundle as installed';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $tableExists = Db::get()->fetchOne('SHOW TABLES LIKE "glossary"');
+
+        $installed = !empty($tableExists);
+
+        if ($installed) {
```
dvesh3:
```suggestion
        if ($schema->hasTable('glossary')) {
```

dvesh3: we also need a warning here that the bundle should be enabled manually in `config/bundles.php`.

dvesh3 (changes requested):

bundles/GlossaryBundle/src/Controller/SettingsController.php
mattamon: I think we cannot remove the admin prefix from the routing, because then we cannot use the AdminController anymore because of the checks. We can simply add the admin back into the name, but I guess we should make it  ```pimcore_admin_glossary_settings_glossary``` then?

bundles/GlossaryBundle/src/Migrations/Version20221215071650.php
mattamon: If the glossary table exists:
image

bundles/GlossaryBundle/public/js/glossary.js
mattamon: Yes we can still use the settings. Just wanted to make it a bit clearer that it is now under glossary instead of settings.

bundles/GlossaryBundle/src/Migrations/Version20221215071650.php
dvesh3: looks good.. can we also add the bundle name to the warning?

bundles/GlossaryBundle/src/Controller/SettingsController.php
dvesh3: Ok, I just scanned `Ecommercebundle` and there we use the admin prefix for the route and not in the name so keeping it consistent then we can go without admin prefix in the name. could you please revert it? sorry for rework 😞

bundles/GlossaryBundle/src/Migrations/Version20221215071650.php
mattamon: Should we add the whole name ```Pimcore\Bundle\GlossaryBundle\GlossaryBundle```  or just ```GlossaryBundle```?

mattamon: We can do it like this
```php
  $this->warnIf(
          $schema->hasTable('glossary'),
          sprintf('Please make sure to enable the %s manually in config/bundles.php', GlossaryBundle::class)
  );
```

bundles/GlossaryBundle/src/Migrations/Version20221215071650.php
```diff
@@ -0,0 +1,34 @@
...
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+use Pimcore\Bundle\GlossaryBundle\GlossaryBundle;
+use Pimcore\Db;
+use Pimcore\Model\Tool\SettingsStore;
+
+/**
+ * Checking if glossary tables already exist
+ */
+final class Version20221215071650 extends AbstractMigration
```
dvesh3: I guess this needs to be moved to Core migrations otherwise we get into chicken-egg problem as the bundle needs to be installed to run the migration and the migration needs to be executed to install the bundle.

dvesh3 (changes requested):
we should update docs to notify the bundle should be enabled and installed to use the feature https://github.com/pimcore/pimcore/blob/bd8b764859c2a28be11280aab44d20f761c4cb72/doc/Development_Documentation/18_Tools_and_Features/21_Glossary.md#L2-L1

tests/Unit/Document/Glossary/GlossaryTest.php
```diff
@@ -31,6 +33,9 @@ class GlossaryTest extends TestCase
      */
     protected function setUp(): void
     {
+        if(!GlossaryBundle::isInstalled()) {
```
dvesh3: Glossary tests are [skipped](https://github.com/pimcore/pimcore/actions/runs/3730712672/jobs/6328082354#step:11:3218)at the moment. The bundle needs to be installed before running tests like we do it for ecommerce bundle here https://github.com/pimcore/pimcore/actions/runs/3730712672/jobs/6328082354#step:11:3218

mattamon: Is it enough to write a Helper like this one?
https://github.com/pimcore/pimcore/blob/11.x/tests/Support/Helper/Ecommerce.php

dvesh3: we already have a helper for Unit suite https://github.com/pimcore/pimcore/blob/56d742deb8e61b33d34040714458460fa774f7fb/tests/Support/Helper/Unit.php#L23

dvesh3: maybe we don't need the skip check anymore since the bundle is always installed.

mattamon:
Updating docs with new examples -> https://pimcore.com/docs/pimcore/11.0/Development_Documentation/Best_Practice/Open_By-External_Id.html

doc/Development_Documentation/23_Installation_and_Upgrade/09_Upgrade_Notes/README.md
```diff
@@ -55,7 +55,10 @@ Please make sure to set your preferred storage location ***before*** migration.
 - [Email] Removed the deprecated methods setBodyHtml(), setBodyText(), createAttachment() and setSubject(). Use html(),
   text(), attach() and subject() instead.
 - [Application Logger] Table names of archive tables are now named with year-month rather than month-year see [#8237](https://github.com/pimcore/pimcore/issues/8237).
-- [Glossary] `pimcoreglossary()` tag has been removed, please use the `pimcore_glossary` Twig filter.
+- [Glossary] 
+  - `pimcoreglossary()` tag has been removed, please use the `pimcore_glossary` Twig filter.
+  -  Glossary was moved into PimcoreGlossaryBundle
```
dvesh3:
```suggestion
  -  Glossary has been moved into PimcoreGlossaryBundle
```

dvesh3 (approved):

bundles/AdminBundle/config/pimcore/default.yaml
```diff
@@ -19,6 +19,135 @@ pimcore_admin:
...
+                key: 't'
+            openAsset:
+                key: 'A'
+                ctrl: true
+                shift: true
+            openObject:
+                key: 'O'
+                ctrl: true
+                shift: true
+            openDocument:
```
brusch:
```suggestion
            open_document:
```
I think we should use snake-case here for all the keys, as this is Symfony bast practice and we should be consistent here.

bundles/AdminBundle/public/js/pimcore/helpers.js
```diff
@@ -3241,3 +3236,37 @@ pimcore.helpers.deleteConfirm = function (title, name, deleteCallback) {
             }
         }.bind(this))
 };
+
+/**
+ * Building menu with priority
+ * @param items
+ */
+pimcore.helpers.buildMenu = function(items) {
```
brusch, line 3244: Do we really still need that, as we're doing the ordering in `pimcore.menu.menu` anyway?

bundles/AdminBundle/public/js/pimcore/helpers.js
```diff
@@ -3241,3 +3236,37 @@ pimcore.helpers.deleteConfirm = function (title, name, deleteCallback) {
...
+    }
+};
+
+/**
+ * Creating menu with sorting function
+ * @param menu
+ * @returns {pimcore.menu.menu}
+ */
+pimcore.helpers.createMenu = function(menu) {
+    return Ext.create('pimcore.menu.menu', menu);
```
brusch: I think we don't need a helper for this, let's just use `Ext.create('pimcore.menu.menu', menu)` directly 😊

bundles/AdminBundle/public/js/pimcore/layout/menu.js
```diff
@@ -0,0 +1,43 @@
...
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+/**
+ * Adding a priority sorting function for menus
```
brusch, line 16:
```suggestion
 * @internal
 * Adding a priority sorting function for menus
```

bundles/GlossaryBundle/config/pimcore/default.yaml
```diff
@@ -0,0 +1,15 @@
+# Pimcore
+pimcore:
+    glossary:
```
brusch:
```suggestion
pimcore_glossary:
```
To be consistent, we probably should use a dedicated root key here, right? 
The config definition needs to be moved to the bundle as well. 
An we need to consider this in the upgrade notes.

bundles/GlossaryBundle/public/js/glossary.js
```diff
@@ -57,7 +57,7 @@ pimcore.settings.glossary = Class.create({
 
         var itemsPerPage = pimcore.helpers.grid.getDefaultPageSize();
         this.store = pimcore.helpers.grid.buildDefaultStore(
-            Routing.generate('pimcore_admin_settings_glossary'),
+            Routing.generate('pimcore_glossary_settings_glossary'),
```
brusch, line 60: maybe we should move that to the `pimcore.glossary.*` namespace as well, to be consistent.

brusch (dismissed):

mattamon:
Added a section in the Upgrade Notes for all bundle extractions

bundles/AdminBundle/config/pimcore/default.yaml
mattamon: Yes this seems to be the right way. I Just did it to match the method names. Maybe we should add the method then as its  own property? I don't want to like fiddle around with all the keys to make proper function calls automatically. I would then set the key and the action as required
```yaml
            open_document:
                key: 'D'
                action: openDocument
                ctrl: true
                shift: true
```

Something like this?

bundles/AdminBundle/public/js/pimcore/helpers.js
mattamon: Yes the buildMenu runs through all submenus since every submenu needs to be a pimcore.menu.menu, as well. Otherwise it  won't get sorted.

bundles/GlossaryBundle/public/js/glossary.js
mattamon: Well strange, this is already in the glossary namespace in my file. :) 
Otherwise the routing would not work because I changed the route in the Settingscontroller too.

bundles/AdminBundle/config/pimcore/default.yaml
dvesh3: fixed it with adding `action` property

PR #13766, opened 2022-12-07
Bump symfony/contracts to 3.2 & knplabs/knp-paginator-bundle to 6.0.0

dvesh3:
## Changes in this pull request  
Resolves #13579
Resolves #13580

PR #13812, opened 2022-12-15
Revert symfony/contracts version 2 support

dvesh3:
Follow up to #13766

## Additional info  
There's a `psr/cache` requirement conflict between `symfony/contracts:3` and `matomo/device-detector`:

- symfony/contracts requires `psr/cache: ^3.0`
- matomo/device-detector requires `psr/cache: ^1.0.1`


dvesh3:
Conflicts coming from enterprise bundles only.

PR #13821, opened 2022-12-16
[Task]: Fix possible errors after applying php native types 

### LINKED ISSUE #13656: [Task]: Fix possible errors after applying php native types

mcop1:
Part of #13656

### DIFF
--- a/lib/Workflow/SupportStrategy/ExpressionSupportStrategy.php
+++ b/lib/Workflow/SupportStrategy/ExpressionSupportStrategy.php
@@ -54,7 +54,8 @@ class ExpressionSupportStrategy implements WorkflowSupportStrategyInterface
             return false;
         }
 
-        return $this->expressionService->evaluateExpression($workflow, $subject, $this->expression);
+        $ret = $this->expressionService->evaluateExpression($workflow, $subject, $this->expression);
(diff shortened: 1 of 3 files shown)


lib/Workflow/SupportStrategy/ExpressionSupportStrategy.php
```diff
@@ -54,7 +54,8 @@ public function supports(WorkflowInterface $workflow, $subject): bool
             return false;
         }
 
-        return $this->expressionService->evaluateExpression($workflow, $subject, $this->expression);
+        $ret = $this->expressionService->evaluateExpression($workflow, $subject, $this->expression);
+        return is_bool($ret) ? (bool)$ret : false;
```
jdreesen: Why cast it to bool when it already is bool?

mcop1: To be honest, since `evaluateExpression` returns mixed, I think we should just go for

`return (bool)$this->expressionService->evaluateExpression($workflow, $subject, $this->expression);`

mcop1: Or use `filter_var` instead of `is_bool`

Issue #6747, opened 2020-07-07
Don't invalidate the complete session on logout


elements-team:
# Feature Request
When logging out, [the complete session is invalidated](https://github.com/pimcore/pimcore/blob/master/bundles/AdminBundle/Security/LogoutSuccessHandler.php#L99).

### Expected behavior
Only clear Pimcore-related session values.

#### Suggested Solution
Only use the attribute bags defined [here](https://github.com/pimcore/pimcore/blob/master/bundles/CoreBundle/Resources/config/pimcore/default.yml#L236).

elements-team:
### Use Case
1. User clicks on logout
2. Event is triggered
3. New RedirectResponse is set using `$event->setResponse(new RedirectResponse($url))`
4. `Session::invalidate();` is called.
5. Redirect is triggered, calling the new action, which wants to access values from the session, which is now empty.

psuhane:
This is the critical issue, below are few scenerios from Pimcore Demo - https://demo.pimcore.com

- If user is logged-in as frontend webshop user and try logging-in as pimcore admin user then frontend user session is invalidated. This also happens vice-a-versa scenario.

- I doubt if below code works after the payment completion in Checkout controller, because session is already invalidated. This is not working in my webshop
`/**
     * @Route("/checkout-completed", name="shop-checkout-completed")
     *
     * @param SessionInterface $session
     * @param Factory $ecommerceFactory
     *
     * @return Response
     */
    public function checkoutCompletedAction(SessionInterface $session, Factory $ecommerceFactory)
    {
        $orderId = $session->get('last_order_id');

        $order = OnlineShopOrder::getById($orderId);

        $trackingManager = $ecommerceFactory->getTrackingManager();
        $trackingManager->trackCheckoutComplete($order);

        return $this->render('checkout/checkout_completed.html.twig', [
            'order' => $order,
            'hideBreadcrumbs' => true
        ]);
    }`

- Because of this GA tracking is also being affected.

Can't we take this issue on priority and fix this on next minor release? @brusch @dvesh3 @fashxp

brusch:
This is on the roadmap for v11.

mlumia:
Temporary solution until v11 is to add a logout listener and call `$event->stopPropagation();` if uri path contains admin.

dvesh3:
Fixed by #13825

PR #13825, opened 2022-12-16
Logout from custom firewall always deletes complete session

jheimbach:
## Steps to reproduce

1. create custom symfony firewall
2. login to custom firewall
3. login to admin
4. trigger logout for custom firewall
5. Check login status in pimcore admin (you are logged out)

## Additional info  
The LogoutEventListener is called on each LogoutEvent, not only pimcore_admin related events. Additionally the session is cleared entirely but it should only remove the admin attributes.

I also fixed the event handler, it should set the response to the given event object and not return it.


fashxp:
Related to https://github.com/pimcore/pimcore/issues/6747 ... 
We don't wanna do that in 10.5, can you please rebase to 11.x?

jheimbach:
Hey @dvesh3, 

sorry didn't searched the open issues enough.

I rebased onto 11.x branch

dpfaffenbauer:
@fashxp what are the side-effects of this? Why not having it in 10.x?

brusch:
@JHeimbach if we use tags now, we can remove the `EventSubscriberInterface` from `LogoutListener` and these lines, right: 
https://github.com/pimcore/pimcore/blob/83142b53cfd5a791e3c6167470a66336fd03da22/bundles/AdminBundle/src/Security/Event/LogoutListener.php#L49-L54

@dpfaffenbauer we are just worried about potential security related side-effects, and it has been always like this 😉 So let's merge this into `11.x` and keep the behavior for the `10` branch.

jheimbach:
@brusch we use an event_subscriber tag, the tag changes only the dispatcher.  If you want we can use an event_listener tag with all settings combined (event, method, dispatcher)

brusch:
@JHeimbach All good, thanks for clarifying! 👍

dvesh3:
@JHeimbach please rebase on 11.x branch to have this change https://github.com/pimcore/pimcore/commit/03822b93c4aefdd08e218b7758353c58996dfee4 for fixing failing tests. thanks!

jheimbach:
done

dvesh3 (approved):

dvesh3:
@JHeimbach works like a charm 💯 thank you very much!

PR #13828, opened 2022-12-19
[PHP] Prevent deprecated error "Passing null to parameter #1 ($string) of type string is deprecated

BlackbitDevs:
In `getVersionPreview()` method of some field types `htmlspecialchars()` gets called. Nowadays this triggers the deprecation error `Passing null to parameter #1 ($string) of type string is deprecated` if called with `null`. This PR casts the provided `$data` to a string to prevent this error, in analogy to https://github.com/pimcore/pimcore/blob/a837b3375ac230c537520db1b4fc79fcbabb1e49/models/DataObject/ClassDefinition/Data/Extension/Text.php#L61


mcop1:
Thank you

PR #13835, opened 2022-12-19
[Task]: Deprecate event "pimcore.element.note.postAdd"

### LINKED ISSUE #13461: [Task]: Deprecate event "pimcore.element.note.postAdd"
Add "pimcore.note.postAdd" as event and trigger both events for 10.6.

Remove deprecated event in Pimcore 11 later

mcop1:
resolves #13461


models/Element/Note.php
```diff
@@ -156,6 +157,7 @@ public function save()
 
         if (!$isUpdate) {
             \Pimcore::getEventDispatcher()->dispatch(new ElementEvent($this), ElementEvents::POST_ADD);
+            \Pimcore::getEventDispatcher()->dispatch(new ElementEvent($this), NoteEvents::POST_ADD);
```
brusch, line 167: Should we add a check here if there's a listener for the deprecated event and issue a deprecation message? 
E.g. 
```php
if(!empty(\Pimcore::getEventDispatcher()->getListeners(ElementEvents::POST_ADD)) { ...
```

brusch (changes requested):

models/Element/Note.php
mcop1: Sure, done -> https://github.com/pimcore/pimcore/pull/13835/commits/9ac080c19d180305e7254cf743301c12f8c04adf

PR #13837, opened 2022-12-19
Fix preGetData missing language param for Data\Wysiwyg

jremmurd:
Without this PR a link generator won't be able to know for which language to generate the link in the WYSIWYG for.

Only a `context` is passed, but it does not contain the current language/index from the LocalizedField.
image

Maybe for future versions, it would be better to use `_locale` for links as recommended by Symfony.

Also see https://github.com/pimcore/pimcore/blob/11.x/models/DataObject/Localizedfield.php#L502

### DIFF
--- a/models/DataObject/ClassDefinition/Data/Wysiwyg.php
+++ b/models/DataObject/ClassDefinition/Data/Wysiwyg.php
@@ -343,6 +343,7 @@ class Wysiwyg extends Data implements ResourcePersistenceAwareInterface, QueryRe
         return Text::wysiwygText($data, [
                 'object' => $container,
                 'context' => $this,
+                'language' => $params['language'],
             ]);
     }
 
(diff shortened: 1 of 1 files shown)


models/DataObject/ClassDefinition/Data/Wysiwyg.php
```diff
@@ -343,6 +343,7 @@ public function preGetData(/** mixed */ $container, /** array */ $params = []) /
         return Text::wysiwygText($data, [
                 'object' => $container,
                 'context' => $this,
+                'language' => $params['language'] ?? null,
```
brusch, line 346:
```suggestion
                'params' => $params ?? [],
```
Probably even better to pass all params, right?

brusch (changes requested):

models/DataObject/ClassDefinition/Data/Wysiwyg.php
jremmurd: I am not sure. This params are directly passed to the link generator and i am not sure if the other params are  useful in the generator at all. 

Example:
image

brusch: Hmm, that's true, seems to be redundant information 👍 Thanks for clarifying! 😊

PR #13844, opened 2022-12-20
[Tags] Enhance tag UI

BlackbitDevs:
This PR brings:
- resizable "assigned tags" panel when assigning tags to elements (otherwise long tag paths may not be visible)
- use full panel width for assigned element tags (before the tag path column was hard-coded to 400px)
- add context menu item to "All Tags" in tag tree to reload tag tree

Tag assignment panel before:
Bildschirmfoto 2022-12-20 um 07 52 57

Tag assignment panel after:
Bildschirmfoto 2022-12-20 um 07 52 09
(I manually adjusted tag tree panel size - initially it is equally wide as before)


robertSt7:

PR #13845, opened 2022-12-20
Translation::getByKeyLocalized(): Remove unnecessary func_get_args()

blankse:
Was introduced by https://github.com/pimcore/pimcore/pull/7800 but the params now uncommented by https://github.com/pimcore/pimcore/pull/13474


dvesh3:
@blankse Thanks 😊 nice catch!

PR #13851, opened 2022-12-20
12518 move seo to bundle

### LINKED ISSUE #12518: [Task] Move SEO to SeoBundle within bundles folder in pimcore/pimcore repository
### Improvement description

Move `SEO` to `SeoBundle` within bundles folder in `pimcore/pimcore` repository

mattamon:
Resolves #12518 

## Additional info  
Moving seo functionality into SeoBundle
Able to install and uninstall via Installer. Permissions are in the Bundle
User Keybindings will be added after glossary bundle is merged
Moved translations into the bundle


bundles/AdminBundle/src/Controller/Admin/SettingsController.php
```diff
@@ -1553,49 +1553,6 @@ public function videoThumbnailUpdateAction(Request $request): JsonResponse
...
-    {
-        $this->checkPermission('robots.txt');
-
-        $values = $request->get('data');
-        if (!is_array($values)) {
-            $values = [];
-        }
-
-        foreach ($values as $siteId => $robotsContent) {
-            SettingsStore::set('robots.txt-' . $siteId, $robotsContent, 'string', 'robots.txt');
```
kingjia90, line 1478: the following import can be removed
https://github.com/pimcore/pimcore/blob/ecd4a3f54bfa521916c5603f649d5131e34a54d0/bundles/AdminBundle/src/Controller/Admin/SettingsController.php#L38

bundles/SeoBundle/public/js/seopanel.js
```diff
@@ -21,6 +21,7 @@ pimcore.document.seopanel = Class.create({
 
     activate: function () {
         var tabPanel = Ext.getCmp("pimcore_panel_tabs");
+        console.log(tabPanel);
```
kingjia90:
```suggestion

```

bundles/SeoBundle/src/Installer.php
```diff
@@ -0,0 +1,63 @@
...
+        $this->removeUserPermission();
+        parent::uninstall();
+    }
+
+    private function addUserPermission(): void
+    {
+        $db = \Pimcore\Db::get();
+
+        foreach (self::USER_PERMISSIONS as $permission) {
+            $db->insert('users_permission_definitions', [
```
kingjia90, line 47: in my local instance, this would break because these permissions definitions are already there
Can we change it to something that check if already exists? and/or a migration to remove the existing permissions

bundles/SeoBundle/public/js/startup.js
```diff
@@ -0,0 +1,96 @@
...
+                    itemId: 'pimcore_menu_marketing_seo_http_errors',
+                    handler: this.showHttpErrorLog
+                });
+            }
+
+            // get index of marketing.targeting
+            if (seoMenu.length > 0) {
+                const toolbar = pimcore.globalmanager.get('layout_toolbar');
+
+                toolbar.marketingMenu.add({
```
kingjia90: For a non-admin user with only SEO bundle permissions, the `marketingMenu` doesn't exists (due the IF in the following snippet), and thus won't work/appear.
 
https://github.com/pimcore/pimcore/blob/ecd4a3f54bfa521916c5603f649d5131e34a54d0/bundles/AdminBundle/public/js/pimcore/layout/toolbar.js#L580-L581

kingjia90 (changes requested):
Overall looks and works fine, just found some minor issues

bundles/SeoBundle/src/Installer.php
mattamon: I moved the migration to the CoreBundle. It should by default get marked as installed, so the permission should not be installed again. If you do a fresh install, this problem should not occur, because I removed the permissions in the InstallerBundle. So basically when you upgrade, the bundle should be marked as installed by default. 

Wdyt?

bundles/SeoBundle/public/js/startup.js
mattamon: You are right, maybe we should discuss, what should happen here then. 
There are some changes with the PimcoreGlossaryBundle but they still would not solve this problem.
https://github.com/pimcore/pimcore/pull/13811/files#diff-0bd635bc68ae1ed97161f850140cef5707f9be71d5dd982c1f978c6aab2255c7R604

Maybe we should add those menus even though they are empty and check before creating the menu again.
https://github.com/pimcore/pimcore/pull/13811/files#diff-0bd635bc68ae1ed97161f850140cef5707f9be71d5dd982c1f978c6aab2255c7R1156 -> with this event you can still add items. Maybe we have a check after that if some menu items would be empty right? 

Wdyt?

bundles/SeoBundle/src/Installer.php
kingjia90: sounds good to me, at least for a seamless upgrade from 10.x, ~but not sure about from a skeleton point of view, if anything "optional" should be deactiveted by default. Let's ask in the daily~ confrimed that by default in fresh install it should be deactived

bundles/SeoBundle/public/js/startup.js
mattamon: Like we discussed in the daily:
https://github.com/pimcore/pimcore/pull/13811/files#diff-0bd635bc68ae1ed97161f850140cef5707f9be71d5dd982c1f978c6aab2255c7R482 -> i removed the if condition.
Everyone who does the refactoring should do this, so we do it one by one.

Unfortunately it is not possible to simply hide the menu item so I do not create it if items are empty.
https://github.com/pimcore/pimcore/pull/13811/files#diff-0bd635bc68ae1ed97161f850140cef5707f9be71d5dd982c1f978c6aab2255c7R1164

kingjia90: what about moving all the code in a separate method 
eg. 
```
createMarketingMenu: function(items){
   this.marketingMenu = new Ext.menu.Menu({
   ...
}
```
and from the bundle, check if `typeof toolbar.marketingMenu == 'undefined'` , then `toolbar.createMarketingMenu()` first, and finally `toolbar.marketingMenu.add();`

mattamon: Like I mentioned, I refactored everything in the GlossaryBundle, so maybe we should wait for those changes to be merged. :) This add method won't work anymore because we create every menu at the very end, to avoid sorting the menu after it is already rendered.

bundles/SeoBundle/src/Controller/Document/DocumentController.php
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * @Route("/document")
+ *
+ * @internal
+ */
+class DocumentController extends ElementControllerBase
+{
+    use DocumentTreeConfigTrait;
+
+    final const DOCUMENT_ROOT_ID = 1;
```
brusch:
```suggestion
    final private const DOCUMENT_ROOT_ID = 1;
```

brusch (changes requested):

brusch:
@mattamon there are some new conflicts that need to be resolved 😊

bundles/SeoBundle/src/Controller/Document/DocumentController.php
mattamon: I made it private only, final private is not possible, since private is not visible to others.

bundles/SeoBundle/public/js/startup.js
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+    },
+
+    showDocumentSeo: function () {
+        try {
+            pimcore.globalmanager.get("bundle_seo_seo_seopanel").activate();
+        }
+        catch (e) {
+            pimcore.globalmanager.add("bundle_seo_seo_seopanel", new pimcore.bundle.seo.seopanel());
```
kingjia90: Looks like there's an extra `seo` 
```suggestion
            pimcore.globalmanager.add("bundle_seo_seopanel", new pimcore.bundle.seo.seopanel());
```

bundles/SeoBundle/src/Controller/SettingsController.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        $this->checkPermission('robots.txt');
+
+        $values = $request->get('data');
+        if (!is_array($values)) {
+            $values = [];
+        }
+
+        foreach ($values as $siteId => $robotsContent) {
+            SettingsStore::set('robots.txt-' . $siteId, $robotsContent, 'string', 'robots.txt');
```
kingjia90, line 63: Not sure if something changed in this PR, but i cannot see the `robots.txt`
On demo.pimcore.fun, in the settingstore table, id column is `robots.txt-default`, while on this is `robots.txt-0`. 
Could you please also have a look.

kingjia90 (changes requested):
LGTM, just found these 2 things, beside the keybidings which will be fixed in https://github.com/pimcore/pimcore/issues/13994

bundles/SeoBundle/src/Controller/SettingsController.php
mattamon: I just took a deeper look and it this was part of a type fix PR
https://github.com/pimcore/pimcore/commit/92da8b98e56ef80ddf4e5cefa5fdb78b7af999f9#diff-d8a9d98ddf01d40976e2f54e4ed4c36bdd81218f67107385e81f3947911b7828R1044
There it changed from "default" to 0. 
Maybe we should bring it up?

PR #13853, opened 2022-12-20
[Bug] Fixed string comparison issue to detect changed while saving localized fields

### LINKED ISSUE #13767: [Bug]: If autosave is disabled, input fields in localizedfields are not saved correctly.
### Expected behavior

If I change the value of an inputField under localizedFields in an object from 3.0 to 3.00 and save it, this value should also be set after reloading the object.
image

### Actual behavior

After reloading the object the value jumps back to 3.0.

### Steps to reproduce

1. Create a class with the following elements (all default settings):
- Panel
  - Localized Fields
    - Input

image

2. Set auto_save_interval to 0 for objects in config/config.yaml:
```
    objects:
        auto_save_interval: 0
```

3. Create object and set value of input field to 3.0
4. Save object
5. Change value of input field to 3.00 and save object
6. Reload Object

Corepex:
Resolves #13767


Corepex:
We should discuss if this is correct: 

https://github.com/pimcore/pimcore/blob/3ccc154dbef71d3baebfef8bf65c0580ad2d7fb3/models/DataObject/Localizedfield.php#L767-L783

Corepex:
TODO: 
- [ ] add upgrade notes (if needed)

mcop1:
> We should discuss if this is correct:
> 
> https://github.com/pimcore/pimcore/blob/3ccc154dbef71d3baebfef8bf65c0580ad2d7fb3/models/DataObject/Localizedfield.php#L767-L783

@Corepex Please open a new ticket for that

PR #13854, opened 2022-12-20
Fix `property_exists(): Passing null to parameter #2 ($property) of type string is deprecated`

jdreesen:
Fixes:

> Deprecated: property_exists(): Passing null to parameter #2 ($property) of type string is deprecated in /var/www/pimcore/vendor/pimcore/pimcore/models/DataObject/Traits/ObjectVarTrait.php on line 53


blankse:
Should we also change the php doc to `@param string|null $var` if $var can be null?

mcop1:
As always, thank you guys 👍

PR #13858, opened 2022-12-20
[Assets] Support document (PDF) version preview

BlackbitDevs:
For PDF assets you currently cannot preview the single versions:
Bildschirmfoto 2022-12-20 um 17 06 48

With this PR, preview for PDF documents gets available - via iframe similar to https://github.com/pimcore/pimcore/blob/1dccd93c925c961e49a21a6fef8753f6ee743e0a/bundles/AdminBundle/public/js/pimcore/asset/document.js#L108-L110

Version preview for PDF now looks like this:
Bildschirmfoto 2022-12-20 um 17 11 49


lib/Twig/Extension/HelpersExtension.php
```diff
@@ -104,6 +105,21 @@ public function getImageVersionPreview(string $file): string
...
+    /**
+     * @param string $file
+     *
+     * @return string
+     *
+     * @throws \Exception
+     */
+    public function getAssetVersionPreview($file)
+    {
+        $dataUri = 'data:application/pdf;base64,'.base64_encode(file_get_contents($file));
```
kingjia90: For PDFs it works fine, but doesn't work for any other file format that would end up as "document" type.

image

Could you please narrow down by checking if the file is a pdf format or extend the support to every other formats and remove this hardcoded `application/pdf` here?

kingjia90 (changes requested):

lib/Twig/Extension/HelpersExtension.php
brusch: @BlackbitDevs 🏓😉 - Thanks!

BlackbitDevs: @kingjia90 , @brusch I have added a MIME type instead of the hard-coded `application/pdf`. 
And I have also changed the preview for non-displayable file types: Those versions now get provided as downloads:
Bildschirmfoto 2023-01-09 um 08 44 28

bundles/AdminBundle/templates/admin/asset/show_version_document.html.twig
```diff
@@ -1 +1,10 @@
-{% include '@PimcoreAdmin/admin/asset/show_version_unknown.html.twig' %}
+{% if asset.getMimeType() == 'pdf' %}
```
kingjia90:
```suggestion
{% if asset.getMimeType() == 'application/pdf' %}
```
See also https://github.com/pimcore/pimcore/blob/455d6743e0558f00e6ba7754f9933ed0f42848e6/bundles/AdminBundle/src/Controller/Admin/Asset/AssetController.php#L1647

bundles/AdminBundle/templates/admin/asset/show_version_unknown.html.twig
```diff
@@ -30,7 +30,11 @@
 <table id="wrapper" cellpadding="0" cellspacing="0" border="0">
     <tr>
         <td>
-            Sorry, no preview available
+            Sorry, no preview available<br>
+            {% set tempFile = asset.getTemporaryFile() %}
+            {% set dataUri = pimcore_asset_version_preview(tempFile) %}
+
+            <a href="{{ dataUri }}" download="{{ asset.getFilename() }}">{{ 'download'|trans([],'admin') }}</a>
```
kingjia90: This is cool, might be quite handy, so users do not have to re-publish an older file to download it.

Just an input, would it be good to add a prefix (eg. date or version id?) to the filename (under the attribute `download`) to differentiate between historical versions?

kingjia90 (changes requested):
Looks better and works fine, just these 2 inputs

bundles/AdminBundle/templates/admin/asset/show_version_unknown.html.twig
BlackbitDevs: Thanks. I now added the version date as file name suffix, e.g. when asset file name is abc.doc then for a version of date 2023-01-10 the download file name will be abc-2023-01-10.doc.

kingjia90: better, but might be a "problem" if one uploaded multiple files on the same date, eg.
image

BlackbitDevs: Right, so I added also the time

kingjia90: Thank you! 
Just one last thing, the files in the screenshot gets suffixed as `2023-01-10-08-59-26` (1 hour difference), somewhere the timezone isn't applied, causing a mismatch with the datetime we see in the column

BlackbitDevs: This was the UNIX timestamp (which is by definition in GMT / UTC). Now I converted it it to the server timezone.
The displayed timezone in the versions panel is a bit different: Here the UNIX timestamp from the version gets transfered and then converted to a human readable date by ExtJS (which also does timezone adjustment).

kingjia90: Mhhh, still don't match, for example, i haven't changed the server timezone (and by default is UTC), while in frontend ExtJS "automatically" knows it should convert to CET.

BlackbitDevs: This is not "automatically", this is because your browser knows which timezone it is in. Of course we could also do the conversion of the UNIX timestamp with JS ~~but is it really worth it~~? Yes, it is because it may lead to confusion which version got downloaded. Ok, will adjust it

kingjia90: mhhh, what about manage the date suffix in frontend ExtJs, since the purpose is to help end-user to distinguish between files and they likely would expect that the filename date matches the one that one sees in the versions grid. 

All these conversion could be too troublesome, eg. if server timezone is something UTC+8 and user is browsing from UTC-3, still would end up in a mismatch.

BlackbitDevs: Ok, now it uses browser / user timezone.

kingjia90: It works!
Maybe a little too vanilla JS, what about the following change?
```
<script>
    document.addEventListener('DOMContentLoaded', function () {

        {% set filenamePrefix = asset.getFilename()|split('.')|slice(0,-1)|join('.') %}
        {% set filenameSuffix = asset.getFilename()|split('.')|last %}
        let versionDate = Ext.Date.format( new Date({{ version.getDate() * 1000 }}), 'Y-m-d-H-i-s');
        document.getElementById('download-version-link').setAttribute('download', "{{ filenamePrefix }}-"+ versionDate+".{{ filenameSuffix }}");
    }, false);
</script>

<!-- library scripts -->
<script src="/bundles/pimcoreadmin/js/../extjs/js/ext-all.js" {{ pimcore_csp.getNonceHtmlAttribute()|raw }}></script>
```

BlackbitDevs: To me this would be ok. Others may say that loading complete ExtJS only to format a date would be like taking a sledgehammer to crack a nut - but actually it should get loaded from the browser cache anyway becasue this page only gets displayed in Pimcore backend.
Have changed it ;-)

kingjia90: Thank you!
I had the exact thoughts 😄
In addition, thought that to avoid that, it could be solved by replacing the DOM content instead of using iframes, but that may require too much refactoring for just this minor drawback of loading extjs from cache again

BlackbitDevs: With DOM content instead of iframes for version preview I think the PDF preview would not work. But even if it would, let's finish this PR for now and maybe enhance the behaviour later...

bundles/AdminBundle/src/Controller/Admin/Asset/AssetController.php
```diff
@@ -15,6 +15,8 @@
 
 namespace Pimcore\Bundle\AdminBundle\Controller\Admin\Asset;
 
+use DateTimeImmutable;
+use DateTimeZone;
```
kingjia90: LGTM, just need to remove these lines, plus `$fileExtension = File::getFileExtension($asset->getFilename());` and merge 11.x to fix and pass tests.

BlackbitDevs: Done.

PR #13861, opened 2022-12-21
[Actions] PHP 8.2 Tests & Checks

brusch:


jdreesen:
Will Pimcore 10 also be compatible to PHP 8.2?

brusch:
@jdreesen currently we don't have this on our list, no 😬

jdreesen:
I'm just asking because [the `composer.json` suggests it is compatible](https://github.com/pimcore/pimcore/blob/10.5/composer.json#L31) (I haven't tried it, though).

brusch:
@jdreesen yep, noticed that as well 👍😉  We need to check and see, if it's easily possible I'd say it would be ok for 10.6.

models/DataObject/ClassDefinition/Data/Extension/ColumnType.php
```diff
@@ -18,16 +18,14 @@
 
 trait ColumnType
 {
+    public $columnType = null;
```
blankse:
```suggestion
    public array|string|null $columnType = null;
```

models/DataObject/ClassDefinition/Data/Extension/ColumnType.php
```diff
@@ -37,9 +35,7 @@ public function getColumnType(): array|string|null
      */
     public function setColumnType($columnType): static
```
blankse:
```suggestion
    public function setColumnType(array|string $columnType): static
```

models/DataObject/ClassDefinition/Data/Extension/QueryColumnType.php
```diff
@@ -18,16 +18,14 @@
 
 trait QueryColumnType
 {
+    public $queryColumnType = null;
```
blankse:
```suggestion
    public array|string|null $queryColumnType = null;
```

models/DataObject/ClassDefinition/Data/Extension/QueryColumnType.php
```diff
@@ -37,9 +35,7 @@ public function getQueryColumnType(): array|string|null
      */
     public function setQueryColumnType($queryColumnType): static
```
blankse:
```suggestion
    public function setQueryColumnType(array|string $queryColumnType): static
```

blankse (changes requested):
Add type hints

brusch:
Depends on #13888

kingjia90:
Closing this PR in favor of https://github.com/pimcore/pimcore/pull/14080
As requested here https://github.com/pimcore/pimcore/pull/14080#issuecomment-1398099768

PR #13862, opened 2022-12-21
Remove PreAuthenticatedAdminSessionFactory::create()

blankse:
This method was from [SecurityFactoryInterface](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/SecurityFactoryInterface.php) and is not a part of [AuthenticatorFactoryInterface](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AuthenticatorFactoryInterface.php)


blankse:
PhpStan error is fixed with https://github.com/pimcore/pimcore/pull/13870

brusch:
#13909 makes this PR obsolete, @blankse thanks anyway 😊

PR #13866, opened 2022-12-21
Fix adding existing translation results in error in Pimcore admin

### LINKED ISSUE #13860: [Bug]: Adding existing translation results in error in Pimcore admin
### Expected behavior

Adding an existing translation does not result in an error.

### Actual behavior

Error is shown:

image

### Steps to reproduce

1. Open Pimcore Demo /admin
2. Trie to add an already existing translation key

dvesh3:
Resolves #13860


mattamon (changes requested):
Exception is still thrown.
image
Shouldn't it be a more friendly user message?

dvesh3:
>  Shouldn't it be a more friendly user message?

@mattamon fixed with user friendly message. Also, there was an issue with predefined meta data existing identifier exception handler, which is fixed in this PR.

mattamon (approved):

PR #13874, opened 2022-12-21
[Task] Implemented `ObjectAwareFieldInterface` into localizedfield

### LINKED ISSUE #12598: Localized fields should implement `ObjectAwareFieldInterface`
### Improvement description

Follow up to #12588

Currently there's a special solution for localized fields which is covered by using `ObjectAwareFieldInterface`
See: https://github.com/pimcore/pimcore/blob/6774f998a83b3f09cf4f5d1374a5f7406e9ca0e4/models/DataObject/Concrete.php#L789

The approach is then similar to the object bricks and field collections, see: 
https://github.com/pimcore/pimcore/pull/12588/files

Corepex:
Resolves #12598


models/DataObject/Service.php
```diff
@@ -774,6 +774,7 @@ public static function loadAllObjectFields(AbstractObject $object)
                     $value = $object->$getter();
                     if ($value instanceof Localizedfield) {
                         $value->loadLazyData();
+                        $object->__objectAwareFields['localizedfields'] = true;
```
fashxp: why is that needed?

models/DataObject/Concrete.php
```diff
@@ -745,17 +745,14 @@ public function __wakeup()
...
-        // no clue why this happens
-        if (property_exists($this, 'localizedfields') && $this->localizedfields instanceof Localizedfield) {
-            $this->localizedfields->setObject($this, false);
-        }
-
         // renew object reference to other object aware fields
         foreach ($this->__objectAwareFields as $objectAwareField => $exists) {
             if (isset($this->$objectAwareField) && $this->$objectAwareField instanceof ObjectAwareFieldInterface) {
-                $this->$objectAwareField->setObject($this);
+                if($this->$objectAwareField instanceof Localizedfield){
```
fashxp: lets create another interface `ObjectAwareFieldWithMarkAsDirtyInterface` that extends from the `ObjectAwareFieldInterface` and use the interface check in that `if` instead of the `Localizedfield` check

`Localizedfield` then should implement the `ObjectAwareFieldWithMarkAsDirtyInterface`

fashxp (changes requested):

models/DataObject/Concrete.php
Corepex: done 👍

models/DataObject/ObjectAwareFieldWithMarkAsDirtyInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\DataObject;
+
+interface ObjectAwareFieldWithMarkAsDirtyInterface
+{
+    public function setObject(?Concrete $object, bool $markAsDirty): static;
```
blankse:
```suggestion
    /**
     * @return $this
     */
    public function setObject(?Concrete $object, bool $markAsDirty): static;
```

models/DataObject/ObjectAwareFieldWithMarkAsDirtyInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Model\DataObject;
+
+interface ObjectAwareFieldWithMarkAsDirtyInterface
```
brusch: We don't need a dedicated interface here, we just use `ObjectAwareFieldInterface` and change the method signature accordingly. 
The `$markAsDirty` parameter is not needed. In case it's necessary we can call `markAllLanguagesAsDirty()` directly instead. 

Additionally we need to check why `$markAsDirty` does even exists, because in theory, changing the owner of the localized field should make any difference in the dirty status. So I guess this doesn't have a valid ready at all (need to be checked). 

See also this comment: https://github.com/pimcore/pimcore/pull/14089#discussion_r1092977900

brusch (changes requested):

models/DataObject/Fieldcollection/Data/AbstractData.php
```diff
@@ -81,7 +81,16 @@ public function setObject(?Concrete $object): static
         $this->object = $object;
 
         if (property_exists($this, 'localizedfields') && $this->localizedfields instanceof Localizedfield) {
-            $this->localizedfields->setObject($object, false);
+            $dirtyLanguages = $this->localizedfields->getDirtyLanguages();
+            $this->localizedfields->setObject($object);
+            if(is_array($dirtyLanguages)) {
```
brusch, line 86: @Corepex since we need this piece of code several time, we should put that into a separate method, e.g. `markLanguagesAsDirty(array $languages)`

brusch (changes requested):

models/DataObject/Fieldcollection/Data/AbstractData.php
Corepex: added within cbfd9110f99a011123c82b1309579c2d3ea1ab8a

PR #13880, opened 2022-12-22
[Routing] Do not invoke dynamic routing for non-default context requests

### LINKED ISSUE #13587: [Bug]: Pimcore dynamic routing (and other routing-related listeners) invoked even for system routes
### Expected behavior

System routes should be resolved ASAP, without invoking `Pimcore\Routing\DynamicRouteProvider::getRouteCollectionForRequest`, `\Pimcore\Model\Document\Service::getNearestDocumentByPath`, etc.

### Actual behavior

System routes like `/_fragment` used for [Symfony's ESI implementation](https://symfony.com/doc/current/http_cache/esi.html) still triggers Pimcore's routing even though it's not required in this case.

### Steps to reproduce

1. generate a fragment request which triggers a controller like so

```php
class Controller
{
    public function invokeAction(): Reponse
    {
        return new Response('AAA');
    }
}
```
2. generate a new fragment URL (for example, using `render_esi` Twig function), it will looks something like
```
/_fragment?_hash=ABC&_controller%3DController%253A%253AinvokeAction
```
3. access the URL and notice it's behaving as if this route is dynamic: loading routes from the database, checking against possible documents named like that (which shouldn't even be allowed since it would break this feature).

Optimizing this endpoint specifically is crucial to be able to use Pimcore efficiently with ESI since there could be many ESI requests required to fulfil the response in certain scenarios, every ms saved here makes a difference.

dvesh3:
Resolves #13587

### DIFF
--- a/lib/Routing/DynamicRouteProvider.php
+++ b/lib/Routing/DynamicRouteProvider.php
@@ -68,6 +68,11 @@ final class DynamicRouteProvider implements RouteProviderInterface
     public function getRouteCollectionForRequest(Request $request): RouteCollection
     {
         $collection = new RouteCollection();
+
+        if($request->attributes->has('_controller')) {
+            return $collection;
+        }
+
         $path = $originalPath = urldecode($request->getPathInfo());
 
         // site path handled by FrontendRoutingListener which runs before routing is started


brusch:
@dvesh3 just pushed a more generic approach, can you plz check? 

@dkarlovi maybe you can double check this fix as well, thanks (it's fixing #13587)

lib/Routing/DynamicRouteProvider.php
```diff
@@ -68,6 +68,11 @@ public function addHandler(DynamicRouteHandlerInterface $handler)
     public function getRouteCollectionForRequest(Request $request): RouteCollection
     {
         $collection = new RouteCollection();
+
+        if($request->attributes->has('_controller')) {
```
dvesh3, line 72:
```suggestion
        if($request->get('_controller', false)) {
```
the param is in the query:
image

brusch: @dvesh3 For me it was in `attributes`, it's added to the request here: 
https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/HttpKernel/EventListener/FragmentListener.php#L73

Not sure if `$request->get()` could have some possible side-effects, as you can then use `?_controller=foo` to skip dynamic routing on any request 🤔

dkarlovi: It will be in the query for fragment endpoint (ESI), but that's a special case since the request is signed.

dvesh3: @brusch the fragment request does not reach at line 73 because of this check here https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/HttpKernel/EventListener/FragmentListener.php#L61

dkarlovi: @dvesh3 the goal here is to avoid doing dynamic routing if Symfony's routing already sucessfully triggered (or wil

brusch: @dvesh3 really? Why should that be set already, as it's in the `attributes` bag, that cannot be set by request params (e.g. `GET` ...). I think only the `_path` query param is set in the first place, right?

dkarlovi: Need to pay very close attention here because it's easy to introduce RCE here on accident.

dvesh3: yes, this could lead to RCE so reverted the commit.

PR #12048, opened 2022-05-06
[Task]: change $this->get() in controllers to use injection

kingjia90:
## Changes in this pull request  
Resolves #11745

## Additional info  
the `get` shortcut method in controller is deprecating in symfony 5.4
https://symfony.com/blog/new-in-symfony-5-4-controller-changes#deprecated-some-controller-shortcuts

PR #13881, opened 2022-12-22
AdminController: Use dependency injection

blankse:
The attributes were introduced here https://github.com/pimcore/pimcore/pull/12048 but never used.

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/IndexController.php
+++ b/bundles/AdminBundle/Controller/Admin/IndexController.php
@@ -184,8 +184,8 @@ class IndexController extends AdminController implements KernelResponseEventInte
      */
     protected function addPluginAssets(array &$templateParams)
     {
-        $templateParams['pluginJsPaths'] = $this->getBundleManager()->getJsPaths();
-        $templateParams['pluginCssPaths'] = $this->getBundleManager()->getCssPaths();
+        $templateParams['pluginJsPaths'] = $this->bundleManager->getJsPaths();
+        $templateParams['pluginCssPaths'] = $this->bundleManager->getCssPaths();
 
         return $this;
     }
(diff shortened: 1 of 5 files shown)


bundles/AdminBundle/src/Controller/AdminController.php
```diff
@@ -38,6 +38,31 @@ abstract class AdminController extends Controller implements AdminControllerInte
 
     protected PimcoreBundleManager $bundleManager;
 
+
+    /**
+     * @required
+     */
```
jdreesen: I think it's time to use attributes instead of annotations, see: https://symfony.com/doc/6.1/service_container/autowiring.html#autowiring-other-methods-e-g-setters-and-public-typed-properties

blankse: You are right. I changed it and create a PR for the other annotations: https://github.com/pimcore/pimcore/pull/13885

kingjia90:
Thank you for the PR!

Just one related thing, It would bc break the output data config toolkit bundle
https://github.com/pimcore/output-data-config-toolkit/blob/5476dbc13b001d61bb55bd1d7d3bf2f41454525f/src/Controller/AdminController.php#L98

Since the 5.x would support [10.6 & 11](https://github.com/pimcore/output-data-config-toolkit/blob/5476dbc13b001d61bb55bd1d7d3bf2f41454525f/composer.json#L8)
Should we keep the `getTranslator()` to `return $this->translator` (eventually switching base to 10.6 and add deprecation triggers) or is better to have a BC layer in the bundle?

jdreesen:
Since the `AdminController` is not `@internal`, the getters should be deprecated in `10.6`.

blankse:
Should I rebase the PR to 10.6 and keep the getters with deprecation warning? Then we can remove them in 11.0.

kingjia90:
> Should I rebase the PR to 10.6 and keep the getters with deprecation warning? Then we can remove them in 11.0.

Yes, please. That would be ideal, then in the bundle 5.x branch could direclty use `$this->translator`

blankse:
@kingjia90 Done :)

PR #13882, opened 2022-12-22
Unit::setBaseunit() must allow string and null

blankse:
You get errors if you try to load a unit with baseunit.


mcop1:
... and thank you again!

PR #13888, opened 2022-12-22
[Data Objects] Remove unnecessary properties from data-type classes

### LINKED ISSUE #13864: `ColumnType::setColumnType()` obsolete?
### Improvement description

Related to #13888
https://github.com/pimcore/pimcore/blob/83142b53cfd5a791e3c6167470a66336fd03da22/models/DataObject/ClassDefinition/Data/Extension/ColumnType.php#L38 seems to be obsolete - we should remove it if so. 

This would need a deprecation in `10.6`.

brusch:
Resolves #13864


models/DataObject/ClassDefinition/Data.php
```diff
@@ -1308,4 +1291,14 @@ public function isForbiddenName(): bool
...
+
+    public function jsonSerialize(): mixed
+    {
+        $tempObj = clone $this; // we clone, so that we only get public properties of the object
+        $data = get_object_vars($tempObj);
+        $data['fieldtype'] = $this->getFieldType();
+        $data['datatype'] = 'data';
+        unset($data['blockedVarsForExport']);
+        return $data;
+    }
```
jdreesen: Cloning doesn't help here to only get the `public` properties, because it's still the same object scope, which means you can still access `protected`/`private` properties.

See: https://3v4l.org/p9FOu

brusch: @jdreesen 👍 thanks for the hint, I was hoping this is a quick win 😉

brusch: @jdreesen should be fixed now -> I tried to avoid using reflection for performance reasons, don't know if this actually matters, but just to explain my approach 😉

models/DataObject/ClassDefinition/Data/EncryptedField.php
```diff
@@ -291,7 +271,7 @@ public function checkValidity(mixed $data, bool $omitMandatoryCheck = false, arr
      */
     public function synchronizeWithMasterDefinition(Model\DataObject\ClassDefinition\Data $masterDefinition)
     {
-        $this->datatype = $masterDefinition->datatype;
+
```
mcop1: Just double checking .. was the method left empty by intent? If yes, can we just remove it?

models/DataObject/ClassDefinition/Data/EncryptedField.php
```diff
@@ -513,4 +493,14 @@ public function denormalize(mixed $value, array $params = []): Model\DataObject\
 
         return $value;
     }
+
+    public function getColumnType(): array|string|null
```
mcop1:
```suggestion
    public function getColumnType(): string
```
We could narrow down the return types to either string or array for all of the `getColumnType` and `getQueryColumnType` methods. Like it is already the case here https://github.com/pimcore/pimcore/blob/a2bc2e66e92b521d2bd7ef3cc6765c8eec715389/models/DataObject/ClassDefinition/Data/NumericRange.php#L166 (we can even omit the `null`)

I think we should be consistent at least.

models/DataObject/ClassDefinition/Data/InputQuantityValue.php
```diff
@@ -161,4 +127,22 @@ public function denormalize(mixed $value, array $params = []): ?InputQuantityVal
...
+    {
+        return [
+            'value' => 'varchar(255)',
+            'unit' => 'varchar(50)',
+        ];
+    }
+
+    public function getQueryColumnType(): array|string|null
+    {
+        return $this->getColumnType();
```
mcop1, line 135: I noticed the call to `getColumnType` here, which is fine, because column types and query column types are the same. But at other instances we just copied the definition, e.g. https://github.com/pimcore/pimcore/pull/13888/files#diff-f9e8614c6fe54e4873af1506ee66172b2c8aad0ee906e85e93b07c8905382c53R439 and/or https://github.com/pimcore/pimcore/pull/13888/files#diff-f9e8614c6fe54e4873af1506ee66172b2c8aad0ee906e85e93b07c8905382c53R444

Maybe we should be consistent here?

mcop1 (commented):
Had no problems while testing the pr, but I noticed some really small things we could adapt, if we want to.

brusch:
@mcop1 Ready 4 Review 😊

fashxp:
@brusch @mcop1 deprecations in 10.6 for removed methods are missing

PR #13889, opened 2022-12-22
Bug fix - allow export to contain duplicate columns

### LINKED ISSUE #11493: [Bug]: colums missing in CSV Export from GRID with multiple operators of the same type
### Expected behavior

When exporting a CSV from the grid I would expect to get a file with all the columns that are shown in Pimcore.
This should also be the case, when I use the same operator/formatter multiple times.

No columns should be missing.

### Actual behavior

Currently when you use the same operator/formatter multiple times, the csv export removes all of them except the first one.
This is the case, when you use the grid options to define your columns and use the same type multiple times.

It seems to be the case with all operators, where you can not explicitly set a name for the column head.
The export removes duplicate column-names.

### Steps to reproduce

1. Open a grid view for data objects
2. Set Grid Options with operators of the same type
image

2a) Grid View is fine
image

3. Export the CSV File -> colum missing
image

torqdev:
Resolves #11493

### DIFF
--- a/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
+++ b/bundles/AdminBundle/Controller/Admin/DataObject/DataObjectHelperController.php
@@ -1441,6 +1441,7 @@ class DataObjectHelperController extends AdminController
 
     public function encodeFunc($value)
     {
+        $value = $value['data'];
         $value = str_replace('"', '""', $value);
         //force wrap value in quotes and return
         return '"' . $value . '"';
--- a/models/DataObject/Service.php
+++ b/models/DataObject/Service.php
@@ -1837,7 +1837,6 @@ class Service extends Model\Element\Service
         }
 
         if ($returnMappedFieldNames) {
-            $tmp = [];
             foreach ($mappedFieldnames as $key => $value) {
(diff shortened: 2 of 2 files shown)


dvesh3:
@torqdev please rebase it on 10.5 branch, since this is a fix. thanks!

torqdev:
Hi @dvesh3 - we were going with `10.6` because of this line https://github.com/pimcore/pimcore/blob/bf8d956c5e7ab85ee85ea79a59348c58625d53d6/CONTRIBUTING.md?plain=1#L11 in `CONTRIBUTING.md` - please advise

robertSt7:
> please advise

@torqdev  Thanks for pointing that out, we'll change that right away. The correct branch for bug fixes is 10.5 at the moment

torqdev:
Hi @dvesh3 - we've updated to use base `10.5`.

dvesh3:
@torqdev thanks! we'll review the changes soon 😉

robertSt7:
@torqdev Thanks

mcop1:
Unfortunately we decided to revert the changes of this pr, because it affects several other bundles, see  #14009

PR #13896, opened 2022-12-23
Improve version checks with Version::MAJOR_VERSION constant

dvesh3:
Follow up to #13701

### DIFF
--- a/lib/Version.php
+++ b/lib/Version.php
@@ -24,6 +24,8 @@ final class Version
 {
     const PACKAGE_NAME = 'pimcore/pimcore';
 
+    const MAJOR_VERSION = 10;
+
     /**
      * @return string
      */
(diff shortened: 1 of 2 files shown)


lib/Version.php
```diff
@@ -24,6 +24,8 @@ final class Version
 {
     const PACKAGE_NAME = 'pimcore/pimcore';
 
+    const MAJOR_VERSION = 10;
```
jdreesen, line 27: This isn't used anywhere?

dvesh3: this is added to support Pimcore bundles on both versions 10 & 11 with BC layer, which checks the major version from this class and implement version specific code.

dvesh3: e.g. https://github.com/pimcore/customer-data-framework/blob/d0d20f591d2f8625e3a6732b33bbcf041c10abbd/src/Model/ActivityList/ActivityListInterface.php#L23

PR #13899, opened 2022-12-23
[DataObject\Document] getChildren & getSiblings shouldn't load the listing

### LINKED ISSUE #11737: `getChildren` shouldn't load the listing
[The line `$this->children[$cacheKey] = $list->load();`](https://github.com/pimcore/pimcore/blob/master/models/Document.php#L666) loads the complete list. Seems like this should be done only when needed, not when `getChildren` is called.

Returning the `Listing` instead would further allow for more flexibility; one could:
- alter the `Listing`s conditions or
- sort the result or
- limit the result (e.g. for pagination)
- ...

> Note that this is only one of the many `getChildren` methods. Changing this is not trivial, as one would probably expect all of these methods to return the same data type. It's also a potential BC for any application calling `getChildren` if they access the data directly - [as it is done here](https://github.com/pimcore/pimcore/blob/master/tests/Model/Document/DocumentTest.php#L244).

---
Currently implemented code from the link above:
```
/**
 * Get a list of the children (not recursivly)
 *
 * @param bool $includingUnpublished
 *
 * @return self[]
 */
public function getChildren($includingUnpublished = false)
{
    $cacheKey = $this->getListingCacheKey(func_get_args());

    if (!isset($this->children[$cacheKey])) {
        $list = new Document\Listing();
        $list->setUnpublished($includingUnpublished);
        $list->setCondition('parentId = ?', $this->getId());
        $list->setOrderKey('index');
        $list->setOrder('asc');
        $this->children[$cacheKey] = $list->load();
    }

    return $this->children[$cacheKey];
}
```

Suggested changes:
```
/**
 * Get a list of the children (not recursivly)
 *
 * @param bool $includingUnpublished
 *
 * @return Document\Listing
 */
public function getChildren($includingUnpublished = false)
{
    $list = new Document\Listing();
    $list->setUnpublished($includingUnpublished);
    $list->setCondition('parentId = ?', $this->getId());
    $list->setOrderKey('index');
    $list->setOrder('asc');

    return $list;
}
```

> Note that the suggested changes do not consider any kind of caching for brevity.

---

What's your take on this?

_Originally posted by @passioneight in https://github.com/pimcore/pimcore/discussions/8439_

dvesh3:
Resolves #11737


blankse:
@dvesh3 Why is only the getChildren() method changed of the Document Model and not the getSiblings() method?

blankse:
And why isn't changed the Asset model to be consistent?

models/Document.php
```diff
@@ -597,7 +594,7 @@ public function getChildren(bool $includingUnpublished = false): array
                 $list->setCondition('parentId = ?', $this->getId());
                 $list->setOrderKey('index');
                 $list->setOrder('asc');
-                $this->children[$cacheKey] = $list->load();
+                $this->children[$cacheKey] = $list;
             } else {
                 $this->children[$cacheKey] = [];
```
blankse, line 599: @dvesh3 The return type doesn't allow an array. Should be also a listing. Or should we return null if no Listing is possible?

models/Element/Service.php
```diff
@@ -671,7 +671,9 @@ protected function updateChildren(DataObject|Document|Asset\Folder $target, Elem
         }
         if (!$found) {
             $newElement = Element\Service::getElementById($new->getType(), $new->getId());
-            $target->setChildren(array_merge($target->getChildren(), [$newElement]));
+            $listing = $target->getChildren();
+            $listing->setData(array_merge($listing->getData(), [$newElement]));
+            $target->setChildren($listing);
```
blankse, line 676: This is also used for Assets. So we must change it also there.

dvesh3:
@blankse thanks for the review. yes, we missed few points here so i'll create a follow up PR. thanks!

PR #13901, opened 2022-12-23
[Bug]: Avoid error of protected property starting with "\0" when removing reference loops

Cruiser13:
There are events if the propvalue is 0 (for never-published documents of Pimcore 6 after upgrading to Pimcore X for example). Prevent errors like 'Cannot access property starting with "\0"' here.

Resolves recent comment in #11801

## Additional info  
To reproduce the bug create a pimcore 6.9.6, create a document without ever publishing it, update to Pimcore 10.5.9+ and try to open it.

### DIFF
--- a/lib/Tool/Serialize.php
+++ b/lib/Tool/Serialize.php
@@ -102,8 +102,9 @@ final class Serialize
             $propCollection = get_object_vars($clone);
 
             foreach ($propCollection as $name => $propValue) {
-                $name = trim($name);
-                $clone->$name = self::loopFilterCycles($propValue);
+                if (!str_starts_with($name, "\0")) {
+                    $clone->$name = self::loopFilterCycles($propValue);
+                }
             }
 
             array_splice(self::$loopFilterProcessedObjects, array_search($element, self::$loopFilterProcessedObjects, true), 1);
--- a/lib/Tool/Serialize.php
+++ b/lib/Tool/Serialize.php
@@ -102,7 +102,9 @@ final class Serialize
             $propCollection = get_object_vars($clone);
 
             foreach ($propCollection as $name => $propValue) {
-                $clone->$name = self::loopFilterCycles($propValue);
+                if(is_array($propValue) || (is_object($propValue))){
+                    $clone->$name = self::loopFilterCycles($propValue);
+                }
             }
 
             array_splice(self::$loopFilterProcessedObjects, array_search($element, self::$loopFilterProcessedObjects, true), 1);
--- a/lib/Tool/Serialize.php
+++ b/lib/Tool/Serialize.php
@@ -102,9 +102,8 @@ final class Serialize
             $propCollection = get_object_vars($clone);
 
             foreach ($propCollection as $name => $propValue) {
-                if(is_array($propValue) || (is_object($propValue))){
-                    $clone->$name = self::loopFilterCycles($propValue);
-                }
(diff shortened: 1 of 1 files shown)


lib/Tool/Serialize.php
```diff
@@ -102,7 +102,9 @@ protected static function loopFilterCycles($element)
             $propCollection = get_object_vars($clone);
 
             foreach ($propCollection as $name => $propValue) {
-                $clone->$name = self::loopFilterCycles($propValue);
+                if(is_array($propValue) || (is_object($propValue))){
+                    $clone->$name = self::loopFilterCycles($propValue);
+                }
```
kingjia90: It doesn't look the right solution, because looking at `loopFilterCycles` could and should return the same `$propValue` when it's not object nor array, and in your case, it may have worked because it skipped the second-level recursion but looking at the error `'Cannot access property starting with "\0"'` that you have mentioned, it looks quite similar to this https://github.com/deliciousbrains/better-search-replace/issues/71, therefore the suggested solution there could also helpful here

```suggestion
                $name = trim($name);
                $clone->$name = self::loopFilterCycles($propValue);
```


Would you be able to retry with this approach?

kingjia90 (changes requested):

lib/Tool/Serialize.php
Cruiser13: @kingjia90 your solution seems to work too 👍

lib/Tool/Serialize.php
```diff
@@ -102,6 +102,7 @@ protected static function loopFilterCycles($element)
             $propCollection = get_object_vars($clone);
 
             foreach ($propCollection as $name => $propValue) {
+                $name = trim($name);
                 $clone->$name = self::loopFilterCycles($propValue);
```
kingjia90: I've took a second look and found out that the right approach is by using a `try` `catch` as found [here](https://github.com/pimcore/pimcore/issues/11801#issuecomment-1082784978)

```suggestion
                try{
                    $clone->$name = self::loopFilterCycles($propValue);
                } catch (\Throwable $e) {
                    continue;
                }
```

The `trim` works but is a false positive, it creates extra properties (5 instead of the expected 3)

image

Live example 
https://3v4l.org/U1u9p#v8.1.11  try catch
https://3v4l.org/52MFj#v8.1.11  trim

kingjia90 (changes requested):

lib/Tool/Serialize.php
jdreesen: This should work, too:

```suggestion
                if (!str_starts_with($name, "\0")) {
                    $clone->$name = self::loopFilterCycles($propValue);
                }
```

https://3v4l.org/H65s7

---

To explain: when you cast an object to an array, the keys of protected/private properties start with an `\0`.

Just trimming the `\0` will result in different array keys, which would then be set as new public properties (by the way: setting these would be [deprecated in PHP 8.2](https://www.php.net/releases/8.2/en.php#deprecate_dynamic_properties))

Cruiser13: Fine with me too :-)

kingjia90 (approved):

blankse:
@kingjia90 I think the Milestone is wrong. Is merged to branch 10.6.

kingjia90:
@blankse ahh, thank you! i'll fix it

